PR #583, opened 2015-02-18
Service interface stubbing

kevinhartman:
* In the event that a service interface is not implemented in Citra, a stubbed (empty) service is returned. The old behavior was to report to the caller that GetServiceHandle itself was unimplemented (which games seem to dislike).
* Adds an unimplemented service interface for what is presumably the Download Play Server service.
* Fixes the nwm::UDS service interface's port name.


src/core/hle/service/dlp_srvr.cpp
```diff
@@ -0,0 +1,15 @@
+#include "core/hle/hle.h"
+#include "core/hle/service/dlp_srvr.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Namespace DLP_SRVR
+
+namespace DLP_SRVR {
+    
+    ////////////////////////////////////////////////////////////////////////////////////////////////////
```
purpasmart96: Alignemnt

src/core/hle/service/dlp_srvr.h
```diff
@@ -0,0 +1,22 @@
...
+// Refer to the license.txt file included.
+
+#include "core/hle/service/service.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Namespace DLP_SRVR
+
+namespace DLP_SRVR {
+
+    /// Interface to "dlp:" service
```
purpasmart96: Alignment

src/core/hle/service/nwm_uds.h
```diff
@@ -18,7 +18,7 @@ class Interface : public Service::Interface {
     Interface();
 
     std::string GetPortName() const override {
-        return "nwm:UDS";
+        return "nwm::UDS";
```
purpasmart96: Is this verified?

src/core/hle/service/dlp_srvr.h
```diff
@@ -0,0 +1,22 @@
...
+    class Interface : public Service::Interface {
+    public:
+        Interface();
+
+        std::string GetPortName() const override {
+        return "dlp:SRVR";
+    }
+};
+
+} // namespace
```
kevinhartman: Forgot newline at the end :walking:

src/core/hle/service/dlp_srvr.cpp
```diff
@@ -0,0 +1,15 @@
...
+    // Interface class
+
```
kevinhartman: I could probably put an empty function table in here.

src/core/hle/service/dlp_srvr.h
```diff
@@ -0,0 +1,22 @@
+// Copyright 2014 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
```
kevinhartman: #pragma once

src/core/hle/service/dlp_srvr.cpp
```diff
@@ -0,0 +1,22 @@
+// Copyright 2014 Citra Emulator Project
```
Subv: Stop living in the past yo!

src/core/hle/service/dlp_srvr.h
Subv: It's already there?

Subv:
I think its better to let the call fail than to return a dummy service.

src/core/hle/service/dlp_srvr.h
kevinhartman: It wasn't when I'd made that comment ;)

kevinhartman:
@Subv The reason that stubbing it seems like a good idea to me is that when this call fails, games seem to trigger a fatal error (and then they hang). I haven't verified that returning an error other than "unimplemented" causes the same kind of hang, but I assume it would. It's probably some library common code that triggers the error. If we defer the errors to the subsequent stubbed service calls, then it's more likely the code will press on for users (which is really the whole point of this change).

purpasmart96:
@kevinhartman My only question is will there still be an error log of the unimplemented function?

kevinhartman:
@purpasmart96 Yep. Here's the flow: The missing port is logged (as before), the stubbed service is returned to the caller, and then the subsequent calls made on that service will be logged as unimplemented, but will return "success" to the caller (as before). We could theoretically delete all of the stubbed service interfaces that have no implementation and we'd have the same behavior as we do now. The point is to get user's games the farthest and to give developers the choice of continuing execution (there's a debug assertion when a service doesn't exist), instead of hitting a fatal error.

src/core/hle/service/nwm_uds.h
purpasmart96: Edit: nevermind

src/core/hle/service/srv.cpp
```diff
@@ -42,8 +42,14 @@ static void GetServiceHandle(Service::Interface* self) {
         LOG_TRACE(Service_SRV, "called port=%s, handle=0x%08X", port_name.c_str(), cmd_buff[3]);
     } else {
         LOG_ERROR(Service_SRV, "(UNIMPLEMENTED) called port=%s", port_name.c_str());
-        res = UnimplementedFunction(ErrorModule::SRV);
+
+        // Note: developers should add the corresponding service interface implementation when this is hit
+        UNIMPLEMENTED();
+
+        // Return the stubbed (empty) service so that applications can continue
+        cmd_buff[3] = Kernel::g_handle_table.Create(&Service::g_stub_service).MoveFrom();
```
Subv: I'm not well-versed in how the handle stuff works now, but I think you should be duplicating this handle instead of creating a new one each time
@yuriks

yuriks: Creating handles is fine, it's essentially what DuplicateHandle does anyway.

However, as it this, this code is unsafe. You cannot create kernel objects except with dynamic allocation. This is because they will try to free themselves using `delete` when their refcount goes to 0. This will cause a crash. `g_stub_service` should be a SmartPtr, and the service should be created during initialization.

kevinhartman:
Made g_stub_service a SharedPtr.

src/core/hle/service/srv.cpp
```diff
@@ -42,8 +42,14 @@ static void GetServiceHandle(Service::Interface* self) {
         LOG_TRACE(Service_SRV, "called port=%s, handle=0x%08X", port_name.c_str(), cmd_buff[3]);
     } else {
         LOG_ERROR(Service_SRV, "(UNIMPLEMENTED) called port=%s", port_name.c_str());
-        res = UnimplementedFunction(ErrorModule::SRV);
+
+        // Note: developers should add the corresponding service interface implementation when this is hit
+        UNIMPLEMENTED();
```
yuriks: Wait, what?

kevinhartman: The reason for UNIMPLEMENTED() here is to pause the debugger for developers so that it's clear they should add a new interface for the discovered service, but that for the current run, they may continue. Maybe that's too invasive. This could just be logged. This also assumes we should continue to have a bunch of stubbed interfaces just for book-keeping.

yuriks: I'm marking the assertions as noreturn in another PR, so they wouldn't be continuable, so yeah, I don't think this should be done. Logging is fine, imho.

purpasmart96:
LGTM after the copyright dates are updated

bunnei:

kevinhartman:
This would probably be more useful if we didn't already have most of the services stubbed out. I'm fine with this being closed for now. I can always tweak it and bring it back if need be.

PR #584, opened 2015-02-18
Asserts: Use lambdas to keep assertion code away from the main code path

yuriks:
I'd love if someone could check disassemblies for gcc and clang to see if it works fine there too.

cc @neobrain :)

### DIFF
--- a/src/common/assert.h
+++ b/src/common/assert.h
@@ -4,24 +4,43 @@
 
 #pragma once
 
+#include <cstdlib>
+
 #include "common/common_funcs.h"
+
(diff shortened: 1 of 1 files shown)


bunnei:
LGTM. I'll hold off on merging until @yuriks gets his GCC/Clang confirmation.

neobrain:
Is there some "easy" translation unit to check how the disassembly for this looks like?

yuriks:
I just checked myself (I don't know why I didn't do this before...) and both gcc and clang pass with flying colors, they emit just a single call.

PR #585, opened 2015-02-18
Rasterizer: Fixed a warning in GetWrappedTexCoord.

Subv:
Redeclaring the variable inside the switch was causing weird behavior.

### DIFF
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -266,10 +266,10 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
 
                         case Regs::TextureConfig::MirroredRepeat:
                         {
-                            int val = (int)((unsigned)val % (2 * size));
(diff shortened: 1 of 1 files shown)


src/video_core/rasterizer.cpp
```diff
@@ -266,9 +266,9 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
 
                         case Regs::TextureConfig::MirroredRepeat:
                         {
-                            int val = (int)((unsigned)val % (2 * size));
-                            if (val >= size)
-                                val = 2 * size - 1 - val;
+                            int coord = (int)((unsigned)val % (2 * size));
+                            if (coord >= size)
+                                coord = 2 * size - 1 - coord;
                             return val;
```
neobrain: You'll want to return `coord` here :p

Subv:
Fixed

PR #588, opened 2015-02-19
Sweeping cleanup of Common

archshift:
- Removed several redundant or useless files that were left over from Dolphin
- Removed some other duplication and redundancy elsewhere in Common


bunnei:
LGTM

PR #596, opened 2015-02-22
Clean up unaligned 32-bit memory reads

kevinhartman:
SkyEye was apparently written against a memory system that would return unaligned 32-bit reads in a "shifted way". Our Read32 implementation would do the opposite shift, so that the interpreter would end up with the correct value. This assumed that Read32 would only be called with an aligned address, or that if it were called with an unaligned address, the necessary shift would be performed in the interpreter. These changes remove both shifts so that no special behavior is needed for unaligned reads.

Note that the SkyEye interpreter (our interpreter in its current state?) only did this shifting for 32-bit reads. Therefore, we don't need to do anything special for 16-bit reads, and hence the comment was simply removed.


lioncash:
Looks good to me. Since we only care about ARMv6K the CP15 check isn't necessary.

bunnei:
LGTM

PR #594, opened 2015-02-21
GPU: Fixed the RGBA8 input format and RGB8 output format

Subv:
in Display Transfers, tested with hwtests.
Verified with https://gist.github.com/Subv/ae2dedc1236b21f9fdbe I'll PR this to the hwtests repo later today after I speak with @Normmatt

PR #597, opened 2015-02-22
Rasterize with the correct color component order.

bunnei:
- Fixes a regression with #594.

### DIFF
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -20,7 +20,7 @@ namespace Rasterizer {
 
 static void DrawPixel(int x, int y, const Math::Vec4<u8>& color) {
     const PAddr addr = registers.framebuffer.GetColorBufferPhysicalAddress();
-    u32* color_buffer = reinterpret_cast<u32*>(Memory::GetPointer(PAddrToVAddr(addr)));
+    u8* color_buffer = Memory::GetPointer(PAddrToVAddr(addr));
 
     // Similarly to textures, the render framebuffer is laid out from bottom to top, too.
     // NOTE: The framebuffer height register contains the actual FB height minus one.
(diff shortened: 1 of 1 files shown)


src/video_core/rasterizer.cpp
```diff
@@ -48,10 +48,10 @@ static const Math::Vec4<u8> GetPixel(int x, int y) {
 
     u32 value = *(color_buffer_u32 + x + y * registers.framebuffer.GetWidth());
     Math::Vec4<u8> ret;
```
Subv: We'll need to add a switch to this function too, but imo that's better left for another PR.

Subv:
LGTM

src/video_core/rasterizer.cpp
```diff
@@ -29,7 +29,7 @@ static void DrawPixel(int x, int y, const Math::Vec4<u8>& color) {
     switch (registers.framebuffer.color_format) {
     case registers.framebuffer.RGBA8:
     {
-        u32 value = (color.a() << 24) | (color.r() << 16) | (color.g() << 8) | color.b();
+        u32 value = (color.r() << 24) | (color.g() << 16) | (color.b() << 8) | color.a();
```
neobrain: I think this code only will work correctly on little-endian CPUs. For better portability, each byte should be initialized individually.

src/video_core/rasterizer.cpp
```diff
@@ -48,10 +48,10 @@ static const Math::Vec4<u8> GetPixel(int x, int y) {
 
     u32 value = *(color_buffer_u32 + x + y * registers.framebuffer.GetWidth());
     Math::Vec4<u8> ret;
-    ret.a() = value >> 24;
-    ret.r() = (value >> 16) & 0xFF;
-    ret.g() = (value >> 8) & 0xFF;
-    ret.b() = value & 0xFF;
+    ret.r() = value >> 24;
```
neobrain: Likewise, the color components here should not be extracted from a u32 here but from a u8 data stream.

PR #600, opened 2015-02-24
Add WriteHWRegRepeat

MoochMcGee:
This hasn't been tested at all. I figured it was low-hanging fruit tho.


Subv:
Isn't this supposed to "repeatedly write the register"? I'm not sure what that means though

MoochMcGee:
OH WHOOPS

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -167,6 +167,29 @@ static void WriteHWRegsWithMask(Service::Interface* self) {
...
+    //Taking size to mean amount of times the reg is hammered.
+    /*// size should be word-aligned
+    if ((size % 4) != 0) {
+        LOG_ERROR(Service_GSP, "Invalid size 0x%08x", size);
+        return;
+    }*/
+    
+    u32 data = cmd_buff[4];
+    
+    for(int i = 0;i<size;i++) GPU::Write<u32>(reg_addr,data);
```
lioncash:
```cpp
for (int i = 0; i < size; i++) 
    GPU::Write<u32>(reg_addr, data);
```

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -167,6 +167,29 @@ static void WriteHWRegsWithMask(Service::Interface* self) {
...
+        LOG_ERROR(Service_GSP, "Write address out of range! (address=0x%08x, size=0x%08x)", reg_addr, size);
+        return;
+    }
+    
+    //Taking size to mean amount of times the reg is hammered.
+    /*// size should be word-aligned
+    if ((size % 4) != 0) {
+        LOG_ERROR(Service_GSP, "Invalid size 0x%08x", size);
+        return;
+    }*/
```
lioncash: If this isn't gong to be used, then just remove it entirely. Don't leave stuff like this around without an explanation.

Subv:
I don't think it should work like this, this needs a hwtest

MoochMcGee:
Yeah, good point.

PR #610, opened 2015-02-25
Add logic ops according to OpenGL ES 1.1 spec and 3dbrew.

MoochMcGee:
This has not been tested in any way, shape, or form. It should be tho.

### DIFF
--- a/src/video_core/math.h
+++ b/src/video_core/math.h
@@ -423,6 +423,27 @@ public:
     {
         return x*x + y*y + z*z + w*w;
     }
+    
+    // Needed for logic ops
+    Vec4<decltype(~T{})> operator ~() const
+    {
+        return MakeVec(~x,~y,~z,~w);
+    }
+    
+    Vec4<decltype(T{}&T{})> operator &(const Vec4& other) const
+    {
+        return MakeVec(x&other.x, y&other.y, z&other.z, w&other.w);
+    }
+    
+    Vec4<decltype(T{}|T{})> operator |(const Vec4& other) const
+    {
+        return MakeVec(x|other.x, y|other.y, z|other.z, w|other.w);
+    }
+    
+    Vec4<decltype(T{}^T{})> operator ^(const Vec4& other) const
+    {
+        return MakeVec(x^other.x, y^other.y, z^other.z, w^other.w);
+    }
 
     // Only implemented for T=float
     float Length() const;
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -374,7 +374,22 @@ struct Regs {
 
         union {
             enum Op {
+                Clear = 0,
+                And = 1,
+                AndReverse = 2,
+                Copy = 3,
                 Set = 4,
+                CopyInverted = 5,
+                NoOp = 6,
+                Invert = 7,
+                Nand = 8,
+                Or = 9,
+                Nor = 10,
+                Xor = 11,
+                Equiv = 12,
+                AndInverted = 13,
+                OrReverse = 14,
+                OrInverted = 15,
             };
 
             BitField<0, 4, Op> op;
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -778,8 +778,87 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                 blend_output     = EvaluateBlendEquation(combiner_output, srcfactor, dest, dstfactor, params.blend_equation_rgb);
                 blend_output.a() = EvaluateBlendEquation(combiner_output, srcfactor, dest, dstfactor, params.blend_equation_a).a();
             } else {
-                LOG_CRITICAL(HW_GPU, "logic op: %x", registers.output_merger.logic_op);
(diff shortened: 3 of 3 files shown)


neobrain:
I think we've had enough examples recently to show that untested code should not be submitted as a PR, yet even be merged.

bunnei:
The implementation is likely right, rather than close the PR for @neobrain's concern and then rewrite these anyway as we stumble upon them in the wild, I'd rather just see you write some hwtests.

src/video_core/rasterizer.cpp
```diff
@@ -778,8 +778,87 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                    case LogicOp::AndReverse:
+                        result = src & ~dest;
+                        break;
+                    
+                    case LogicOp::Copy:
+                        result = src;
+                        break;
+                    
+                    case LogicOp::Set:
+                        result = Math::Vec4<u8>(1,1,1,1);
```
neobrain: Should be 255

src/video_core/rasterizer.cpp
```diff
@@ -778,8 +778,87 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                        break;
+                    
+                    case LogicOp::OrReverse:
+                        result = src | ~dest;
+                        break;
+                    
+                    case LogicOp::OrInverted:
+                        result = ~src | dest;
+                        break;
+                    default:
```
neobrain: insert blank line here

src/video_core/rasterizer.cpp
```diff
@@ -778,8 +778,87 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                 blend_output     = EvaluateBlendEquation(combiner_output, srcfactor, dest, dstfactor, params.blend_equation_rgb);
                 blend_output.a() = EvaluateBlendEquation(combiner_output, srcfactor, dest, dstfactor, params.blend_equation_a).a();
             } else {
-                LOG_CRITICAL(HW_GPU, "logic op: %x", registers.output_merger.logic_op);
-                UNIMPLEMENTED();
+                auto params = registers.output_merger.logic_op;
+
+                using LogicOp = decltype(params)::Op;
+
+                auto EvaluateLogicOp = [&](const Math::Vec4<u8>& src, const Math::Vec4<u8>& dest
```
neobrain: this can be made a static lambda.

src/video_core/pica.h
```diff
@@ -374,7 +374,22 @@ struct Regs {
...
+                NoOp = 6,
+                Invert = 7,
+                Nand = 8,
+                Or = 9,
+                Nor = 10,
+                Xor = 11,
+                Equiv = 12,
+                AndInverted = 13,
+                OrReverse = 14,
+                OrInverted = 15,
```
neobrain: Please align those at `=`

src/video_core/pica.h
```diff
@@ -374,7 +374,22 @@ struct Regs {
 
         union {
             enum Op {
```
neobrain: Should be `enum Op : u32 {`, can you please fix that while you're changing the code?

src/video_core/math.h
```diff
@@ -423,6 +423,27 @@ class Vec4
     {
         return x*x + y*y + z*z + w*w;
     }
+    
+    // Needed for logic ops
```
neobrain: Don't put comments like this in generic utility libraries. In general, I don't think these are appropriate to have as members of Vec at all, and should be defined locally instead (e.g. via some "OrVec" lambda).

src/video_core/rasterizer.cpp
```diff
@@ -778,8 +778,87 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                    
+                    case LogicOp::OrReverse:
+                        result = src | ~dest;
+                        break;
+                    
+                    case LogicOp::OrInverted:
+                        result = ~src | dest;
+                        break;
+                    default:
+                        LOG_CRITICAL(HW_GPU, "Unknown logic op %x", op);
```
neobrain: This can never happen iff all of the logic ops are implemented, so just remove the default branch.

src/video_core/rasterizer.cpp
```diff
@@ -778,8 +778,87 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                        UNIMPLEMENTED();
+                    }
+                    
+                    return result;
+                };
+                
+                blend_output = EvaluateLogicOp(combiner_output, dest, params.op);
```
neobrain: There's not really any need to have EvaluateLogicOp be a lambda since you're only using it once anyway.

src/video_core/rasterizer.cpp
lioncash: Linking onto what neo said, doing things like this (switching off an enum type and having a default branch) will cause warnings on OSX builds.

neobrain:
@MoochMcGee What's the status of this PR?

MoochMcGee:
Dead, I think. I think I lost the branch.

PR #611, opened 2015-02-26
Video core: Fix pixelation/blockiness in textures.

yuriks:
This was caused during morton decoding by me not masking the bits of
each coordinate before merging them, so the bits from x could set bits
in y if it was >255.


Subv:
LGTM

PR #612, opened 2015-02-26
Video core: Fix A4 texture decoding

yuriks:
It was trying to take the LSB from `coarse_x`, which would always be 0
and thus would always return the same texel from each byte. To add
insult to the injury, the conditional was actually the wrong way around
too.

Fixes blocky text in OoT.


bunnei:
LGTM

PR #613, opened 2015-02-26
Changes to improve Windows MSVC CMake support

jroweboy:
I recently downloaded and tried to build citra on windows MSVC 2013 only to be greeted with 151 errors (all of them were just msvc over reacting to linker errors though). I figure this would be quite a deterrent to people that want to get started on Windows, so I tried to streamline the build process by eliminating any unexpected steps. This doesn't change anything about the setup instructions, but rather should just make the instructions work properly. 

The first change was a simple modification to support the following MSVC configurations: Debug, Release, RelWithDebInfo. I tried to choose good compiler flags for each of them, but the main thing that needed to happen was it needed to link against the Multithreaded Runtime Debug DLL instead of the non debug version. In a similar vein, I also added the `/DEBUG` linker flag which fixed some other bug that I don't recall.

The other big change was copying the Qt5 dlls to the right folders when building so you don't get the Qt5Core(d).dll is missing errors. It looks a little hackish, but it seems there really isn't a better way to handle this use case. 

Please let me know if this works for other people. I tried really hard to avoid path issues, but I can't guarantee that I got it correct for everyone's system configuration :smile:


CMakeLists.txt
```diff
@@ -12,25 +12,31 @@ else()
...
+    set(optimization_flags "/MP /Ox /Oy- /GR- /GS- /EHsc")
     # /Zi - Output debugging information
     # /Zo - enahnced debug info for optimized builds
-    set(CMAKE_C_FLAGS_RELEASE   "${optimization_flags} /Zi" CACHE STRING "" FORCE)
-    set(CMAKE_CXX_FLAGS_RELEASE "${optimization_flags} /Zi" CACHE STRING "" FORCE)
-    set(CMAKE_C_FLAGS_RELWITHDEBINFO   "${optimization_flags} /Zi /Zo" CACHE STRING "" FORCE)
-    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${optimization_flags} /Zi /Zo" CACHE STRING "" FORCE)
+    # /MDd - Multi-threaded Debug Runtime DLL
+    set(CMAKE_C_FLAGS_DEBUG   "${optimization_flags} /MDd /Zi /Zo" CACHE STRING "" FORCE)
+    set(CMAKE_CXX_FLAGS_DEBUG "${optimization_flags} /MDd /Zi /Zo" CACHE STRING "" FORCE)
```
yuriks: It's pointless to optimize a debug build so much. It won't be a debug build anymore.

src/citra_qt/CMakeLists.txt
```diff
@@ -67,3 +67,47 @@ target_link_libraries(citra-qt ${OPENGL_gl_LIBRARY} ${CITRA_QT_LIBS})
 target_link_libraries(citra-qt ${PLATFORM_LIBRARIES})
 
 #install(TARGETS citra-qt RUNTIME DESTINATION ${bindir})
+
+# setup a post build to copy the Qt Dlls to the right place on Windows MSVC
+# I only have Qt 5 so if someone wants to add this for Qt 4 as well that would be great
+if (Qt5_FOUND)
```
yuriks: This is windows specific and shouldn't be done on other platforms.

src/citra_qt/CMakeLists.txt
```diff
@@ -67,3 +67,47 @@ target_link_libraries(citra-qt ${OPENGL_gl_LIBRARY} ${CITRA_QT_LIBS})
 target_link_libraries(citra-qt ${PLATFORM_LIBRARIES})
 
 #install(TARGETS citra-qt RUNTIME DESTINATION ${bindir})
+
+# setup a post build to copy the Qt Dlls to the right place on Windows MSVC
+# I only have Qt 5 so if someone wants to add this for Qt 4 as well that would be great
```
neobrain: Who is this ominous "I"?

neobrain: Actually, just change this line to "TODO: Port to Qt4". Although actually actually, I don't think we really intend to support Qt4 on Windows anyway.

src/citra_qt/CMakeLists.txt
```diff
@@ -67,3 +67,47 @@ target_link_libraries(citra-qt ${OPENGL_gl_LIBRARY} ${CITRA_QT_LIBS})
...
+        "${Qt5_DLL_DIR}/icudt*.dll"
+        "${Qt5_DLL_DIR}/icuin*.dll"
+        "${Qt5_DLL_DIR}/icuuc*.dll"
+        "${Qt5_DLL_DIR}/Qt5Cored.*"
+        "${Qt5_DLL_DIR}/Qt5Guid.*"
+        "${Qt5_DLL_DIR}/Qt5OpenGLd.*"
+        "${Qt5_DLL_DIR}/Qt5Widgetsd.*"
+    )
+    foreach(Dll ${Qt5_DEBUG_DLLS})
+        # the if not exist skips this copy if the build directory doesn't exist so it doesn't error out
```
neobrain: => `"if not exists"`

neobrain: Why don't we instead just create the build directory if it doesn't exist?

src/citra_qt/CMakeLists.txt
```diff
@@ -67,3 +67,47 @@ target_link_libraries(citra-qt ${OPENGL_gl_LIBRARY} ${CITRA_QT_LIBS})
...
+        add_custom_command(TARGET citra-qt POST_BUILD
+                           COMMAND if not exist "${CMAKE_BINARY_DIR}/bin/Debug/" goto :cmEnd
+                           COMMAND ${CMAKE_COMMAND} -E
+                             copy ${Dll} "${CMAKE_BINARY_DIR}/bin/Debug/")
+    endforeach()
+    # sooo here's a fun struggle. I can't set a custom command for a single build target,
```
neobrain: Please don't make pointless comments like "sooo here's a fun struggle". It provides no value.

neobrain: Also again, don't talk about some "I" without including context (i.e. a name), or preferably don't phrase it in first person at all.

CMakeLists.txt
```diff
@@ -12,25 +12,31 @@ else()
     add_definitions(/D_CRT_SECURE_NO_WARNINGS)
     # set up output paths for executable binaries (.exe-files, and .dll-files on DLL-capable platforms)
     set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
-
+    set(CMAKE_CONFIGURATION_TYPES Debug Release RelWithDebInfo CACHE TYPE INTERNAL FORCE)
```
neobrain: After 2 minutes of internet research I couldn't find any info on what the later parameters (starting with CACHE) actually do. Maybe a comment would be appropriate?

neobrain: Actually, nevermind, I should've checked `set` instead of the variable name.

src/citra_qt/CMakeLists.txt
```diff
@@ -67,3 +67,47 @@ target_link_libraries(citra-qt ${OPENGL_gl_LIBRARY} ${CITRA_QT_LIBS})
...
+        "${Qt5_DLL_DIR}/Qt5Guid.*"
+        "${Qt5_DLL_DIR}/Qt5OpenGLd.*"
+        "${Qt5_DLL_DIR}/Qt5Widgetsd.*"
+    )
+    foreach(Dll ${Qt5_DEBUG_DLLS})
+        # the if not exist skips this copy if the build directory doesn't exist so it doesn't error out
+        add_custom_command(TARGET citra-qt POST_BUILD
+                           COMMAND if not exist "${CMAKE_BINARY_DIR}/bin/Debug/" goto :cmEnd
+                           COMMAND ${CMAKE_COMMAND} -E
+                             copy ${Dll} "${CMAKE_BINARY_DIR}/bin/Debug/")
```
neobrain: Why not use `file(COPY)` ?

jroweboy: Copying the files there would be done at configuration time, whereas adding it as a custom command means its done at build time (ie the post build step in MSVC). If that is acceptable, then I can change it to `file(COPY)` instead. I just checked and using Clean solution in MSVC didn't remove the dlls from the folder so it seems to be safe to copy them at configuration time.

CMakeLists.txt
jroweboy: So for Debug, it should only have the following flags:

    /MDd /Zi 

Everything else doesn't seem to be needed.

jroweboy:
Ok, I removed extraneous comments and changed the file copying to happen at configure time instead of build time. Also I made the debug build a true debug build in that there are no optimizations that are run.

CMakeLists.txt
yuriks: It should also have `/MP` to enabled parallel builds, which is currently missing.

Subv:
LGTM

bunnei:
LGTM. Because it's a change to our build files, I'd like feedback from other team members before merging. @yuriks @neobrain ?

CMakeLists.txt
```diff
@@ -12,25 +12,32 @@ else()
...
     message(STATUS "Cache compiler flags ignored, please edit CMakeLists.txt to change the flags.")
     # /MP - Multi-threaded compilation
-    # /MD - Multi-threaded runtime
     # /Ox - Full optimization
     # /Oy- - Don't omit frame pointer
     # /GR- - Disable RTTI
     # /GS- - No stack buffer overflow checks
     # /EHsc - C++-only exception handling semantics
-    set(optimization_flags "/MP /MD /Ox /Oy- /GR- /GS- /EHsc")
+    set(optimization_flags "/MP /Ox /Oy- /GR- /GS- /EHsc")
```
yuriks: Since you moved `/MD` out I think you should also move `/MP` since it's not really an opt flag either. Since you're modifying this, also replace `/Ox` with `/O2`, which someone had previously pointed out optimizes more. (And which I didn't change earlier. I think I disagreed, but I was wrong, sorry.)

yuriks:
Flag changes LGTM. I'm not so sure on the DLL copying though. This will pre-copy the DLLs to the directories even if you never build those configurations, right? Look into using [add_custom_command](http://www.cmake.org/cmake/help/v3.0/command/add_custom_command.html) with `POST_BUILD` to copy them after building instead.

jroweboy:
@yuriks i originally used `add_custom_command` but I ended up swapping this out since it was a kind of hacky implementation. Maybe I can look again with fresh eyes and find something better, but this was the best that I've gotten so far https://github.com/jroweboy/citra/blob/a9620cec41e17e730874665c9965a60a484a3639/src/citra_qt/CMakeLists.txt#L84

I'm fine with either way though. I'll address these in a bit and update the PR

yuriks:
You should be able to use [generator expressions](http://www.cmake.org/cmake/help/v3.0/manual/cmake-generator-expressions.7.html) to avoid the issue where you have to always copy all files. This should also reduce the amount of duplication and remove the need to do the directory check, which should greatly improve things.

To remove the "nasty" of having to do multiple `add_custom_command` calls, you can use a cmd for loop: `for %I in (file1.txt file2.txt file3.txt) do copy %I c:\somedir\`, but using CMake variable expansions to expand the list of DLLs to copy instead.

neobrain:
To be honest, copying these files around is misplaced in any build step other than the INSTALL target (which would be done on Linux if it were relevant). However, I guess our Windows developers aren't interested in such a workflow, hence my opinion is mostly "do whatever you want" :p

jroweboy:
I'm not sure why it added in those extra commits by other people, so to make things cleaner, I'm just going to close this PR and open another that I just rebased on master.

yuriks:
That's because you're merging on top of other people's commits instead of rebasing. In the future just force-push instead of opening a new PR.

PR #615, opened 2015-02-27
 Services: Moved the PTM and APT services to their own folder

Subv:
This coincidentally fixes an issue about the PTM service failing to create its SharedExtSaveData archive due to the FS service not being initialized by the time the creating code runs.

Ideally I'd like to move each process to its own folder, and have a single file per process that registers the service classes, which would be in their own files inside that folder. Then each service class would just call functions from the process to complete the commands.


purpasmart96:
LGTM, although I'm not sure removing comments was really warranted. @bunnei ?

src/core/hle/service/apt/apt_a.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "core/hle/hle.h"
+#include "core/hle/service/apt/apt.h"
+#include "core/hle/service/apt/apt_a.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Namespace APT_A
+
+namespace APT_A {
+
+using namespace Service::APT;
```
purpasmart96: I don't think it should be done like this, "using namespace X" is generally to avoided. same applies to the other files which use this.

yuriks: Nah, it's fine, the only thing in here is the function table anyway.

src/core/hle/service/apt/apt_a.cpp
```diff
@@ -0,0 +1,39 @@
...
+// Refer to the license.txt file included.
+
+#include "core/hle/hle.h"
+#include "core/hle/service/apt/apt.h"
+#include "core/hle/service/apt/apt_a.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Namespace APT_A
+
+namespace APT_A {
```
yuriks: Could you put these inside the proper sub-namespace instead of leaving them in the root? `Service::Interfaces` or something.

src/core/hle/service/apt/apt_s.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include "common/common.h"
+#include "common/file_util.h"
+
+#include "core/hle/hle.h"
+#include "core/hle/service/apt/apt.h"
+#include "core/hle/service/apt/apt_s.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Namespace APT_S
```
yuriks: Get rid of these silly headers please, they're pretty annoying.

src/core/hle/service/fs/archive.cpp
```diff
@@ -419,6 +421,8 @@ ResultCode CreateExtSaveData(u32 high, u32 low) {
 void ArchiveInit() {
     next_handle = 1;
 
+    AddService(new FS::FSUserInterface);
```
yuriks: This is inconsistent with the others, that would have `FS_User::Interface` instead.

src/core/hle/service/service.h
```diff
@@ -120,4 +120,7 @@ extern std::unordered_map<std::string, Kernel::SharedPtr<Interface>> g_kernel_na
 /// Map of services registered with the "srv:" service, retrieved using GetServiceHandle.
 extern std::unordered_map<std::string, Kernel::SharedPtr<Interface>> g_srv_services;
 
+/// Adds a service to the services table
+void AddService(Interface* iface);
```
yuriks: `iface` ur mom. (Write the entire word.)

Subv: It was giving errors with Visual Studio, apparently "interface" is a reserved keyword that acts as a more restrictive "struct".

Subv:
@purpasmart96 I removed the "// no error" comments because they were redundant now that RESULT_SUCCESS is being used

src/core/hle/service/apt/apt_a.cpp
Subv: Service::Interfaces::APT::A or something? I didn't put these in the Service::APT namespace because the FunctionTable name would be repeated in each (Or does that not matter because they are static?)

yuriks: It doesn't matter, because they're static, so they're only visible from inside the same cpp.

src/core/hle/service/service.h
yuriks: But `interface` is being used fine in the definition of the function. You should ignore that VS highlights `interface`, it only applies to C++/CLI.

Subv:
Updated

Subv:
Updated again

src/core/hle/service/apt/apt_a.cpp
```diff
@@ -0,0 +1,36 @@
...
+    {0x000E0080, GlanceParameter,    "GlanceParameter?"},
+    {0x003B0040, nullptr,            "CancelLibraryApplet?"},
+    {0x00430040, nullptr,            "NotifyToWait?"},
+    {0x00440000, GetSharedFont,      "GetSharedFont?"},
+    {0x004B00C2, nullptr,            "AppletUtility?"},
+    {0x00550040, nullptr,            "WriteInputToNsState?"},
+};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Interface class
```
yuriks: Remove header pls

src/core/hle/service/service.h
Subv: It's not just the highlight, there are actual errors if i use the entire "interface" word

src/core/hle/service/apt/apt_s.cpp
```diff
@@ -0,0 +1,107 @@
...
+    {0x004B00C2, AppletUtility,         "AppletUtility"},
+    {0x004C0000, nullptr,               "SetFatalErrDispMode"},
+    {0x004D0080, nullptr,               "GetAppletProgramInfo"},
+    {0x004E0000, nullptr,               "HardwareResetAsync"},
+    {0x004F0080, nullptr,               "SetApplicationCpuTimeLimit"},
+    {0x00500040, nullptr,               "GetApplicationCpuTimeLimit"},
+};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Interface class
```
yuriks: Annoy-o-banner

src/core/hle/service/apt/apt_u.cpp
```diff
@@ -0,0 +1,106 @@
...
+    {0x004B00C2, AppletUtility,                   "AppletUtility"},
+    {0x004C0000, nullptr,                         "SetFatalErrDispMode"},
+    {0x004D0080, nullptr,                         "GetAppletProgramInfo"},
+    {0x004E0000, nullptr,                         "HardwareResetAsync"},
+    {0x004F0080, SetAppCpuTimeLimit,              "SetAppCpuTimeLimit"},
+    {0x00500040, GetAppCpuTimeLimit,              "GetAppCpuTimeLimit"},
+};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Interface class
```
yuriks: banner

yuriks:
LGTM

bunnei:
Needs a rebase

Subv:
Fixed the merge conflicts

bunnei:
Squash plz

Subv:
Squashed

bunnei:
LGTM

PR #599, opened 2015-02-23
GPU: Implemented bits 3 and 1 from the display transfer flags.

Subv:
Bit 3 is used to specify a raw copy, where no processing is done to the data, seems to behave exactly as a DMA.
Bit 1 is used to specify whether to convert from a tiled format to a linear format or viceversa.

Verified with hwtests

PR #616, opened 2015-02-27
Added RGBA5551 compatibility in the rasterizer

archshift:
This allows Virtual Console games to display properly.

### DIFF
--- a/src/video_core/color.h
+++ b/src/video_core/color.h
@@ -28,5 +28,24 @@ static inline u8 Convert6To8(u8 value) {
     return (value << 2) | (value >> 4);
 }
 
+/// Convert a 8-bit color component to 1 bit
+static inline u8 Convert8To1(u8 value) {
+    return value >> 7;
+}
+
+/// Convert a 8-bit color component to 4 bit
+static inline u8 Convert8To4(u8 value) {
+    return value >> 4;
+}
+
+/// Convert a 8-bit color component to 5 bit
+static inline u8 Convert8To5(u8 value) {
+    return value >> 3;
+}
+
+/// Convert a 8-bit color component to 6 bit
+static inline u8 Convert8To6(u8 value) {
+    return value >> 2;
+}
 
 } // namespace
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -51,6 +51,16 @@ static void DrawPixel(int x, int y, const Math::Vec4<u8>& color) {
         break;
     }
 
+    case registers.framebuffer.RGBA5551:
+    {
+        u16_le* pixel = (u16_le*)(color_buffer + dst_offset);
+        *pixel = (Color::Convert8To5(color.r()) << 11) |
+                 (Color::Convert8To5(color.g()) << 6)  |
+                 (Color::Convert8To5(color.b()) << 1)  |
+                 Color::Convert8To1(color.a());
+        break;
+    }
+
     default:
         LOG_CRITICAL(Render_Software, "Unknown framebuffer color format %x", registers.framebuffer.color_format.Value());
         UNIMPLEMENTED();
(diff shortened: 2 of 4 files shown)


archshift:
Note: this depends on #599.

archshift:
Rebased onto master.

src/video_core/rasterizer.cpp
```diff
@@ -51,6 +51,16 @@ static void DrawPixel(int x, int y, const Math::Vec4<u8>& color) {
         break;
     }
 
+    case registers.framebuffer.RGBA5551:
+    {
+        u16* pixel = (u16*)(color_buffer + dst_offset);
+        *pixel = (Color::Convert8To5(color.r()) << 11)   |
+                    (Color::Convert8To5(color.g()) << 6) |
+                    (Color::Convert8To5(color.b()) << 1) |
+                    Color::Convert8To1(color.a());
```
bunnei: Can you indent these lines so that they all line up?

bunnei:
Mostly LGTM

src/video_core/rasterizer.cpp
archshift: But they do all line up :P

src/video_core/color.h
```diff
@@ -28,5 +28,14 @@ static inline u8 Convert6To8(u8 value) {
     return (value << 2) | (value >> 4);
 }
 
+/// Convert a 8-bit color component to 5 bit
+static inline u8 Convert8To5(u8 value) {
+    return value * 0x1F / 0xFF;
```
neobrain: Use `>> 3` instead for consistency with the other functions.

src/video_core/color.h
```diff
@@ -28,5 +28,14 @@ static inline u8 Convert6To8(u8 value) {
...
+}
+
+/// Convert a 8-bit color component to 1 bit
+static inline u8 Convert8To1(u8 value) {
+    return value > (0xFF / 2) ? 1 : 0;
```
neobrain: `>> 7` here.

src/video_core/rasterizer.cpp
```diff
@@ -51,6 +51,16 @@ static void DrawPixel(int x, int y, const Math::Vec4<u8>& color) {
         break;
     }
 
+    case registers.framebuffer.RGBA5551:
+    {
+        u16* pixel = (u16*)(color_buffer + dst_offset);
```
neobrain: This is not big-endian compatible. Please fill the individual bytes instead.

src/video_core/rasterizer.cpp
```diff
@@ -90,6 +100,17 @@ static const Math::Vec4<u8> GetPixel(int x, int y) {
         return ret;
     }
 
+    case registers.framebuffer.RGBA5551:
+    {
+        Math::Vec4<u8> ret;
+        u16 pixel = *(u16*)(color_buffer + src_offset);
```
neobrain: Again, not big-endian compatible.

src/video_core/rasterizer.cpp
```diff
@@ -90,6 +100,17 @@ static const Math::Vec4<u8> GetPixel(int x, int y) {
         return ret;
     }
 
+    case registers.framebuffer.RGBA5551:
+    {
+        Math::Vec4<u8> ret;
```
neobrain: There's actually not much of a reason to redeclare this for each switch case, is there?

src/video_core/rasterizer.cpp
neobrain: (Also, similarly to below, this needn't be redeclared for each switch case)

neobrain:
LGTM once big-endian compatibility and consistency issues are addressed.

neobrain:
Actually, repost from #617, which apparently is meant to be merged after this PR:

Since it's now more relevant than ever, We should explain something like this explicitly:

// Components are laid out in reverse byte order, most significant bits first.

Could you add this comment to pica.h (ColorFormat) and replace the one in gpu.h (PixelFormat) with something like this?

src/video_core/color.h
archshift: That would change the meaning of this code, though.

src/video_core/color.h
archshift: Again, that would change the meaning of the code.

archshift:
Should be good to go.

src/video_core/color.h
```diff
@@ -28,5 +28,24 @@ static inline u8 Convert6To8(u8 value) {
...
 }
 
+/// Convert a 8-bit color component to 1 bit
+static inline u8 Convert8To1(u8 value) {
+    return value >> 7;
+}
+
+/// Convert a 8-bit color component to 4 bit
+static inline u8 Convert8To4(u8 value) {
+    return value >> 7;
```
neobrain: typo here?

archshift: Indeed.

yuriks:
LGTM

neobrain:
LGTM

PR #617, opened 2015-02-27
GPU: Added RGB565/RGB8 framebuffer support and various cleanups.

bunnei:
Implements RGB565 and RGB8 framebuffer support. I also refactored framebuffer format encoding/decoding to use shared code that properly handles endianness. Verified correct behavior with multiple games that exercise different framebuffer formats:

### DIFF
--- a/src/core/hw/gpu.cpp
+++ b/src/core/hw/gpu.cpp
@@ -113,8 +113,8 @@ inline void Write(u32 addr, const T data) {
     {
         const auto& config = g_regs.display_transfer_config;
         if (config.trigger & 1) {
-            u8* source_pointer = Memory::GetPointer(Memory::PhysicalToVirtualAddress(config.GetPhysicalInputAddress()));
-            u8* dest_pointer = Memory::GetPointer(Memory::PhysicalToVirtualAddress(config.GetPhysicalOutputAddress()));
+            u8* src_pointer = Memory::GetPointer(Memory::PhysicalToVirtualAddress(config.GetPhysicalInputAddress()));
+            u8* dst_pointer = Memory::GetPointer(Memory::PhysicalToVirtualAddress(config.GetPhysicalOutputAddress()));
 
             unsigned horizontal_scale = (config.scale_horizontally != 0) ? 2 : 1;
             unsigned vertical_scale = (config.scale_vertically != 0) ? 2 : 1;
@@ -125,7 +125,7 @@ inline void Write(u32 addr, const T data) {
             if (config.raw_copy) {
                 // Raw copies do not perform color conversion nor tiled->linear / linear->tiled conversions
                 // TODO(Subv): Verify if raw copies perform scaling
-                memcpy(dest_pointer, source_pointer, config.output_width * config.output_height * 
+                memcpy(dst_pointer, src_pointer, config.output_width * config.output_height * 
                         GPU::Regs::BytesPerPixel(config.output_format));
                 
                 LOG_TRACE(HW_GPU, "DisplayTriggerTransfer: 0x%08x bytes from 0x%08x(%ux%u)-> 0x%08x(%ux%u), flags 0x%08X, Raw copy",
@@ -142,9 +142,7 @@ inline void Write(u32 addr, const T data) {
             // right now we're just skipping the extra pixels.
             for (u32 y = 0; y < output_height; ++y) {
                 for (u32 x = 0; x < output_width; ++x) {
-                    struct {
-                        int r, g, b, a;
-                    } source_color = { 0, 0, 0, 0 };
+                    Math::Vec4<u8> src_color = { 0, 0, 0, 0 };
 
                     u32 scaled_x = x * horizontal_scale;
                     u32 scaled_y = y * vertical_scale;
@@ -170,77 +168,54 @@ inline void Write(u32 addr, const T data) {
                         dst_offset = (x + y * output_width) * dst_bytes_per_pixel;
                     }
 
+                    const u8* src_pixel = src_pointer + src_offset;
                     switch (config.input_format) {
                     case Regs::PixelFormat::RGBA8:
-                    {
-                        u8* srcptr = source_pointer + src_offset;
-                        source_color.r = srcptr[3]; // red
-                        source_color.g = srcptr[2]; // green
-                        source_color.b = srcptr[1]; // blue
-                        source_color.a = srcptr[0]; // alpha
+                        src_color = Color::DecodeRGBA8(src_pixel);
                         break;
-                    }
+
+                    case Regs::PixelFormat::RGB8:
+                        src_color = Color::DecodeRGB8(src_pixel);
+                        break;
+
+                    case Regs::PixelFormat::RGB565:
+                        src_color = Color::DecodeRGB565(src_pixel);
+                        break;
 
                     case Regs::PixelFormat::RGB5A1:
-                    {
-                        u16 srcval = *(u16*)(source_pointer + src_offset);
-                        source_color.r = Color::Convert5To8((srcval >> 11) & 0x1F); // red
-                        source_color.g = Color::Convert5To8((srcval >>  6) & 0x1F); // green
-                        source_color.b = Color::Convert5To8((srcval >>  1) & 0x1F); // blue
-                        source_color.a = Color::Convert1To8(srcval & 0x1);          // alpha
+                        src_color = Color::DecodeRGB5A1(src_pixel);
                         break;
-                    }
 
                     case Regs::PixelFormat::RGBA4:
-                    {
-                        u16 srcval = *(u16*)(source_pointer + src_offset);
-                        source_color.r = Color::Convert4To8((srcval >> 12) & 0xF); // red
-                        source_color.g = Color::Convert4To8((srcval >>  8) & 0xF); // green
-                        source_color.b = Color::Convert4To8((srcval >>  4) & 0xF); // blue
-                        source_color.a = Color::Convert4To8( srcval        & 0xF); // alpha
+                        src_color = Color::DecodeRGBA4(src_pixel);
                         break;
-                    }
 
                     default:
                         LOG_ERROR(HW_GPU, "Unknown source framebuffer format %x", config.input_format.Value());
                         break;
                     }
 
+                    u8* dst_pixel = dst_pointer + dst_offset;
                     switch (config.output_format) {
                     case Regs::PixelFormat::RGBA8:
-                    {
-                        u8* dstptr = dest_pointer + dst_offset;
-                        dstptr[3] = source_color.r;
-                        dstptr[2] = source_color.g;
-                        dstptr[1] = source_color.b;
-                        dstptr[0] = source_color.a;
+                        Color::EncodeRGBA8(src_color, dst_pixel);
                         break;
-                    }
 
                     case Regs::PixelFormat::RGB8:
-                    {
-                        u8* dstptr = dest_pointer + dst_offset;
-                        dstptr[2] = source_color.r; // red
-                        dstptr[1] = source_color.g; // green
-                        dstptr[0] = source_color.b; // blue
+                        Color::EncodeRGB8(src_color, dst_pixel);
                         break;
-                    }
+
+                    case Regs::PixelFormat::RGB565:
+                        Color::EncodeRGB565(src_color, dst_pixel);
+                        break;
 
                     case Regs::PixelFormat::RGB5A1:
-                    {
-                        u16* dstptr = (u16*)(dest_pointer + dst_offset);
-                        *dstptr = ((source_color.r >> 3) << 11) | ((source_color.g >> 3) << 6)
-                                | ((source_color.b >> 3) <<  1) | ( source_color.a >> 7);
+                        Color::EncodeRGB5A1(src_color, dst_pixel);
                         break;
-                    }
 
                     case Regs::PixelFormat::RGBA4:
-                    {
-                        u16* dstptr = (u16*)(dest_pointer + dst_offset);
-                        *dstptr = ((source_color.r >> 4) << 12) | ((source_color.g >> 4) << 8)
-                                | ((source_color.b >> 4) <<  4) | ( source_color.a >> 4);
+                        Color::EncodeRGBA4(src_color, dst_pixel);
                         break;
-                    }
 
                     default:
                         LOG_ERROR(HW_GPU, "Unknown destination framebuffer format %x", config.output_format.Value());
--- a/src/video_core/color.h
+++ b/src/video_core/color.h
@@ -5,47 +5,152 @@
 #pragma once
 
 #include "common/common_types.h"
+#include "video_core/math.h"
 
 namespace Color {
 
 /// Convert a 1-bit color component to 8 bit
-static inline u8 Convert1To8(u8 value) {
+inline u8 Convert1To8(u8 value) {
     return value * 255;
 }
 
 /// Convert a 4-bit color component to 8 bit
-static inline u8 Convert4To8(u8 value) {
+inline u8 Convert4To8(u8 value) {
     return (value << 4) | value;
 }
 
 /// Convert a 5-bit color component to 8 bit
-static inline u8 Convert5To8(u8 value) {
+inline u8 Convert5To8(u8 value) {
     return (value << 3) | (value >> 2);
 }
 
 /// Convert a 6-bit color component to 8 bit
-static inline u8 Convert6To8(u8 value) {
+inline u8 Convert6To8(u8 value) {
     return (value << 2) | (value >> 4);
 }
 
 /// Convert a 8-bit color component to 1 bit
-static inline u8 Convert8To1(u8 value) {
+inline u8 Convert8To1(u8 value) {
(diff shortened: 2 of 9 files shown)


bunnei:
Note: This PR needs to use Convert8To5 and Convert8To6 functions. I'll add these once #616 is merged.

src/core/hw/gpu.cpp
```diff
@@ -181,6 +181,16 @@ inline void Write(u32 addr, const T data) {
                         break;
                     }
 
+                    case Regs::PixelFormat::RGB565:
+                    {
+                        u16 srcval = *(u16*)(source_pointer + src_offset);
```
neobrain: not big-endian compatible.

src/core/hw/gpu.cpp
```diff
@@ -181,6 +181,16 @@ inline void Write(u32 addr, const T data) {
...
+                        source_color.r = Color::Convert5To8((srcval >> 11) & 0x1F); // red
```
neobrain: Those comments aren't necessary.

src/core/hw/gpu.cpp
```diff
@@ -181,6 +181,16 @@ inline void Write(u32 addr, const T data) {
...
+                        source_color.g = Color::Convert6To8((srcval >> 5) & 0x3F);  // green
+                        source_color.b = Color::Convert5To8(srcval & 0x1F);         // blue
+                        source_color.a = 255;                                       // alpha
```
neobrain: Has this been tested?

src/core/hw/gpu.cpp
```diff
@@ -226,6 +236,14 @@ inline void Write(u32 addr, const T data) {
                         break;
                     }
 
+                    case Regs::PixelFormat::RGB565:
+                    {
+                        u16* dstptr = (u16*)(dest_pointer + dst_offset);
```
neobrain: Again, not big-endian compatible.

src/video_core/rasterizer.cpp
```diff
@@ -79,6 +87,17 @@ static const Math::Vec4<u8> GetPixel(int x, int y) {
         return ret;
     }
 
+    case registers.framebuffer.RGB565:
+    {
+        Math::Vec4<u8> ret;
```
neobrain: There's no need to duplicate this declaration in each switch case

src/video_core/rasterizer.cpp
```diff
@@ -43,6 +43,14 @@ static void DrawPixel(int x, int y, const Math::Vec4<u8>& color) {
         break;
     }
 
+    case registers.framebuffer.RGB565:
+    {
+        u8* pixel = color_buffer + dst_offset;
```
neobrain: There's no need to duplicate this declaration in each switch case.

src/video_core/rasterizer.cpp
```diff
@@ -43,6 +43,14 @@ static void DrawPixel(int x, int y, const Math::Vec4<u8>& color) {
...
+        pixel[1] = (color.r() & 0xF8) | (color.g() >> 5);
```
neobrain: Since it's now more relevant than ever, We should explain something like this explicitly:

    // Components are laid out in reverse byte order, most significant bits first.

Could you add this comment to pica.h (ColorFormat) and replace the one in gpu.h (PixelFormat) with something like this?

neobrain:
LGTM other than the endianness and documentation remarks.

yuriks:
Does this work with the ctrulib console?

yuriks:
Also, for reference, a good way of making this endian safe is changing the casts to `u16*` to `u16_le*` instead.

src/core/hw/gpu.cpp
```diff
@@ -226,6 +236,14 @@ inline void Write(u32 addr, const T data) {
...
+                        *dstptr = ((source_color.r >> 3) << 11) | ((source_color.g >> 2) << 5)
+                            | (source_color.b >> 3);
```
yuriks: The wrapping here looks weird and arbitrary. Wrap the part for each component in its own line.

src/core/hw/gpu.cpp
bunnei: yes

src/core/hw/gpu.cpp
```diff
@@ -170,34 +170,41 @@ inline void Write(u32 addr, const T data) {
...
+                        source_color.b = srcptr[1];
+                        source_color.a = srcptr[0];
+                        break;
+                    }
+
+                    case Regs::PixelFormat::RGB565:
+                    {
+                        source_color.r = Color::Convert5To8(srcptr[1] >> 3);
+                        source_color.g = Color::Convert6To8(((srcptr[1] & 0x7) << 3) | (srcptr[0] >> 5));
+                        source_color.b = Color::Convert5To8(srcptr[0] & 0x1F);
```
yuriks: Load from `u16_le` here for less error-prone bit manipulation.

src/core/hw/gpu.cpp
```diff
@@ -170,34 +170,41 @@ inline void Write(u32 addr, const T data) {
...
                     {
-                        u16 srcval = *(u16*)(source_pointer + src_offset);
-                        source_color.r = Color::Convert5To8((srcval >> 11) & 0x1F); // red
-                        source_color.g = Color::Convert5To8((srcval >>  6) & 0x1F); // green
-                        source_color.b = Color::Convert5To8((srcval >>  1) & 0x1F); // blue
-                        source_color.a = Color::Convert1To8(srcval & 0x1);          // alpha
+                        source_color.r = Color::Convert5To8(srcptr[1] >> 3);
+                        source_color.g = Color::Convert5To8(((srcptr[1] & 0x7) << 2) | (srcptr[0] >> 6));
+                        source_color.b = Color::Convert5To8((srcptr[0] >> 1) & 0x1F);
+                        source_color.a = Color::Convert1To8(srcptr[0] & 0x1);
```
yuriks: Load from `u16_le` here.

src/video_core/rasterizer.cpp
```diff
@@ -79,9 +90,17 @@ static const Math::Vec4<u8> GetPixel(int x, int y) {
         return ret;
     }
 
+    case registers.framebuffer.RGB565:
+    {
+        u8* pixel = color_buffer + src_offset;
+        ret.r() = Color::Convert5To8(pixel[1] >> 3);
+        ret.g() = Color::Convert6To8(((pixel[1] & 0x7) << 3) | (pixel[0] >> 5));
+        ret.b() = Color::Convert5To8(pixel[0] & 0x1F);
```
yuriks: Load from `u16_le` here.

yuriks:
LGTM after the u16 changes.

neobrain:
I'm not entirely happy about using `u16_le`, for two reasons:

* it's trying to solve a portability issue, but isn't portable in the strict sense (non-BE or non-LE platforms will break it, and it also by design can't adapt to switching endianness). That said, it's portable enough for virtually anything we'll ever want to run Citra on, so this is just an ideological issue.
* Decoding things from a raw byte buffer makes data serialization points much more explicit than the somewhat subtle `_le` suffix.

On the other hand, data serialization via raw byte buffers is less straightforward to read and might make it harder for compilers to produce optimized code.

So uh.. While I'd prefer the raw byte buffer approach, I guess for the sake of "getting things done" I won't disagree with converting things to `u16_le`.

neobrain:
Oh yeah, LGTM btw, once some uniform way of data serialization is used...

src/video_core/color.h
```diff
@@ -28,5 +28,28 @@ static inline u8 Convert6To8(u8 value) {
     return (value << 2) | (value >> 4);
 }
 
+/// Convert a 8-bit color component to 1 bit
+static inline u8 Convert8To1(u8 value) {
+    return value >> 7;
```
Subv: I don't think this is OK.
See these alpha tests: https://github.com/citra-emu/hwtests/blob/master/source/tests/gpu/displaytransfer.cpp#L238

src/video_core/color.h
```diff
@@ -28,5 +28,28 @@ static inline u8 Convert6To8(u8 value) {
...
+/// Convert a 8-bit color component to 4 bit
+static inline u8 Convert8To4(u8 value) {
+    return value >> 4;
+    
+}
+
+/// Convert a 8-bit color component to 5 bit
+static inline u8 Convert8To5(u8 value) {
+    return value >> 3;
+
```
Subv: Extra line

src/video_core/color.h
yuriks: What's wrong with it? It seems like it would do ok on the test.

bunnei:
Ready for another review folks

src/video_core/color.h
```diff
@@ -48,4 +49,63 @@ static inline u8 Convert8To6(u8 value) {
...
+static inline const Math::Vec4<u8> DecodeRGBA5551(const u8 bytes[]) {
+    return { Convert5To8(bytes[1] >> 3),
+             Convert5To8(((bytes[1] & 0x7) << 2) | (bytes[0] >> 6)),
+             Convert5To8((bytes[0] >> 1) & 0x1F),
+             Convert1To8(bytes[0] & 0x1) };
+}
+
+static inline const Math::Vec4<u8> DecodeRGBA4(const u8 bytes[]) {
+    return{ Convert4To8(bytes[1] >> 4),
+        Convert4To8(bytes[1] & 0xF),
```
neobrain: alignment

src/video_core/color.h
```diff
@@ -48,4 +49,63 @@ static inline u8 Convert8To6(u8 value) {
...
+
+static inline const Math::Vec4<u8> DecodeRGBA5551(const u8 bytes[]) {
+    return { Convert5To8(bytes[1] >> 3),
+             Convert5To8(((bytes[1] & 0x7) << 2) | (bytes[0] >> 6)),
+             Convert5To8((bytes[0] >> 1) & 0x1F),
+             Convert1To8(bytes[0] & 0x1) };
+}
+
+static inline const Math::Vec4<u8> DecodeRGBA4(const u8 bytes[]) {
+    return{ Convert4To8(bytes[1] >> 4),
```
neobrain: space after return

neobrain:
I love it! :D 

Haven't reviewed the actual color en/decoding code thoroughly, since I assume it has changed little since my last review and it has been tested anyway.

If you're bored, debug_utils.cpp still has its own color conversion code, but that can be addressed in a later PR too. Other than the minor style issues, LGTM.

src/core/hw/gpu.cpp
```diff
@@ -172,34 +170,24 @@ inline void Write(u32 addr, const T data) {
 
                     switch (config.input_format) {
                     case Regs::PixelFormat::RGBA8:
-                    {
-                        u8* srcptr = source_pointer + src_offset;
-                        source_color.r = srcptr[3]; // red
-                        source_color.g = srcptr[2]; // green
-                        source_color.b = srcptr[1]; // blue
-                        source_color.a = srcptr[0]; // alpha
+                        src_color = Color::DecodeRGBA8(src_pointer + src_offset);
```
yuriks: This `src_pointer + src_offset` expression is repeated in all of the cases, it should be factored into the calculation above.

src/core/hw/gpu.cpp
```diff
@@ -208,39 +196,24 @@ inline void Write(u32 addr, const T data) {
 
                     switch (config.output_format) {
                     case Regs::PixelFormat::RGBA8:
-                    {
-                        u8* dstptr = dest_pointer + dst_offset;
-                        dstptr[3] = source_color.r;
-                        dstptr[2] = source_color.g;
-                        dstptr[1] = source_color.b;
-                        dstptr[0] = source_color.a;
+                        Color::EncodeRGBA8(src_color, dst_pointer + dst_offset);
```
yuriks: Same comment about the repeated expression.

src/core/hw/gpu.h
```diff
@@ -48,11 +48,11 @@ struct Regs {
...
     enum class PixelFormat : u32 {
-        RGBA8  = 0,
-        RGB8   = 1,
-        RGB565 = 2,
-        RGB5A1 = 3,
-        RGBA4  = 4,
+        RGBA8    = 0,
+        RGB8     = 1,
+        RGB565   = 2,
+        RGBA5551 = 3,
```
yuriks: OpenGL uses `RGB5_A1`, and we also don't follow this convention for the other enum cases (it's not `RGBA8888`, for example) so I think it should be left as it was before.

src/video_core/color.h
```diff
@@ -48,4 +49,63 @@ static inline u8 Convert8To6(u8 value) {
     return value >> 2;
 }
 
+static inline const Math::Vec4<u8> DecodeRGBA8(const u8 bytes[]) {
```
yuriks: Why isn't the param `const u8*`? It's pretty unusual to use this syntax for unsized arrays.

src/video_core/color.h
```diff
@@ -48,4 +49,63 @@ static inline u8 Convert8To6(u8 value) {
...
+    return { bytes[3], bytes[2], bytes[1], bytes[0] };
```
yuriks: This whole file: Why aren't you using `u*_le`? :|

src/video_core/color.h
```diff
@@ -48,4 +49,63 @@ static inline u8 Convert8To6(u8 value) {
...
+}
+
+static inline const Math::Vec4<u8> DecodeRGB8(const u8 bytes[]) {
```
yuriks: Why are these functions marked `static`? They're in a shared header file.

src/video_core/rasterizer.cpp
```diff
@@ -34,32 +34,24 @@ static void DrawPixel(int x, int y, const Math::Vec4<u8>& color) {
 
     switch (registers.framebuffer.color_format) {
     case registers.framebuffer.RGBA8:
-    {
-        u8* pixel = color_buffer + dst_offset;
-        pixel[3] = color.r();
-        pixel[2] = color.g();
-        pixel[1] = color.b();
-        pixel[0] = color.a();
+        Color::EncodeRGBA8(color, color_buffer + dst_offset);
```
yuriks: `color_buffer + dst_offset`, `color_buffer + dst_offset`, `color_buffer + dst_offset`, `color_buffer + dst_offset`

src/video_core/rasterizer.cpp
```diff
@@ -80,34 +72,19 @@ static const Math::Vec4<u8> GetPixel(int x, int y) {
...
     case registers.framebuffer.RGBA8:
-    {
-        u8* pixel = color_buffer + src_offset;
-        ret.r() = pixel[3];
-        ret.g() = pixel[2];
-        ret.b() = pixel[1];
-        ret.a() = pixel[0];
-        return ret;
-    }
+        return Color::DecodeRGBA8(color_buffer + src_offset);
```
yuriks: `color_buffer + src_offset`, `color_buffer + src_offset`, `color_buffer + src_offset`, `color_buffer + src_offset`, `color_buffer + src_offset`.

yuriks:
I kind of changed my mind of ok-ing this without using `u*_le`. I'll summarize my points below:
 - Explicit loading of bytes makes the code much more complicated and harder to read and verify.
 - It's also slower compared to something using bswap in the compilers I've looked at. It's also certainly slower than a plain load which is what `u*_le` will compile down to on a LE arch.
 - There don't exist any platforms in use that aren't either little-endian or big-endian, and it's safe to say that none will be created.
 - For CPUs that in theory support runtime endianess switching, runtime switches aren't supported by any OS that I could find, and would certainly break a great deal of application. There would be no reason to do it within Citra either.
 - Even ignoring the two previous points, manual loading of bytes can be implemented inside `u*_le`, centralizing this concern instead of spreading it over the rest of the codebase, and also easily allowing other individual platforms to take advantage of the performance improvement.

yuriks:
LGTM

neobrain:
Whatever :p

src/video_core/color.h
```diff
@@ -5,47 +5,152 @@
...
+inline const Math::Vec4<u8> DecodeRGBA8(const u8* bytes) {
+    return { bytes[3], bytes[2], bytes[1], bytes[0] };
+}
+
+/**
+ * Decode a color stored in RGB8 format
+ * @param bytes Pointer to encoded source color
+ * @return Result color decoded as Math::Vec4<u8>
+ */
+inline const Math::Vec4<u8> DecodeRGB8(const u8* bytes) {
```
Subv: Are you sure that the alpha component should be 0?

Subv:
It'd be awesome if you could replace the code in the debug framebuffer widget with this

src/video_core/color.h
neobrain: Should probably be 255 indeed. In any case should have a TODO, unless you actually tested it.

Subv: https://github.com/subv/hwtests/blob/master/source/tests/gpu/displaytransfer.cpp#L370

https://github.com/subv/hwtests/blob/master/source/tests/gpu/displaytransfer.cpp#L399

I haven't PR'd them because something is causing a random freeze with the display transfers but these tests are correct

bunnei:
All comments should be addressed

PR #618, opened 2015-02-28
result: Make comparison operators take references

lioncash:
It's unnecessary to make copies for simple comparisons like this.


yuriks:
`ResultCode` is a simple u32, so taking references is potentially more expensive. (It doesn't matter here though since they'll always get inlined.)

bunnei:
LGTM, but yeah, agreed with @yuriks

PR #619, opened 2015-02-28
arm_disasm: Remove unused variable

lioncash:
Also declared an array as static, as it's only used in this translation unit.


yuriks:
LGTM

bunnei:
LGTM

PR #620, opened 2015-02-28
arm_disasm: Show conditional code for BKPT instructions.

lioncash:
Changed cond_to_str to take a uint32, since unsigned numbers are only ever passed to it, and this can be a source of warnings for some compilers (also indexing an array without bounds checking a signed number is kind of iffy).


yuriks:
LGTM

bunnei:
LGTM

PR #623, opened 2015-03-01
Services/FS: Stubbed CardSlotIsInserted to always return false

Subv:
We won't be emulating this for the foreseeable future and it is needed for Home Menu to boot further


yuriks:
Won't this produce interesting results, like if some game expects to be running from the card and calls this?

yuriks:
LGTM

archshift:
LGTM

bunnei:
LGTM

PR #625, opened 2015-03-02
vfp: Get rid of warnings

lioncash:
Silences some warnings when compiling with VS.

### DIFF
--- a/src/core/arm/skyeye_common/vfp/vfp_helper.h
+++ b/src/core/arm/skyeye_common/vfp/vfp_helper.h
@@ -144,8 +144,8 @@ static inline void mul64to128(u64* resh, u64* resl, u64 n, u64 m)
     u32 nh, nl, mh, ml;
     u64 rh, rma, rmb, rl;
 
-    nl = n;
-    ml = m;
+    nl = static_cast<u32>(n);
+    ml = static_cast<u32>(m);
     rl = (u64)nl * ml;
 
     nh = n >> 32;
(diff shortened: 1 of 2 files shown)


src/core/arm/skyeye_common/vfp/vfp_helper.h
```diff
@@ -144,8 +144,8 @@ static inline void mul64to128(u64* resh, u64* resl, u64 n, u64 m)
     u32 nh, nl, mh, ml;
     u64 rh, rma, rmb, rl;
 
-    nl = n;
-    ml = m;
+    nl = static_cast<u32>(n & 0xFFFFFFFF);
+    ml = static_cast<u32>(m & 0xFFFFFFFF);
```
yuriks: Why the ands?

lioncash:
Those were added just to make intent explicit. I can remove them if they aren't wanted.

src/core/arm/skyeye_common/vfp/vfp_helper.h
lioncash: Those were added just to make intent explicit. I can remove them if they aren't wanted.

yuriks: With both the cast and the and, it seemed to me that the and was supposed to be achieving something else, so I was a bit perplexed by it for a bit. It's fine either way to me.

bunnei:
LGTM

PR #634, opened 2015-03-06
Apply the logging filter before sending the message to the queue

linkmauve:
This puts back 3dscraft to its performance level of before logging.


Subv:
LGTM

linkmauve:
Now with actual performance measures, this PR makes 3dscraft render 20× faster, as shown here: http://imgur.com/a/0imIP

bunnei:
LGTM

PR #636, opened 2015-03-07
Set framebuffer layout from EmuWindow.

bunnei:
Precursor PR to touchpad support. This moves the logic for setting the render window framebuffer layout (e.g. width, height, top screen boundaries, bottom screen boundaries) from the renderer to EmuWindow the implementation. Because for now our two EmuWindow implementations use the same basic layout (top screen + bottom screen scaled to the window size), I made a default function for computing the layout part of EmuWindow so that it can be shared. But this design is more flexible and should allow for more creative things in the future (e.g. making the lower screen a thumbnail, or putting it on a separate window).

### DIFF
--- a/src/common/emu_window.h
+++ b/src/common/emu_window.h
@@ -8,6 +8,7 @@
 #include "common/scm_rev.h"
 #include "common/string_util.h"
 #include "common/key_map.h"
+#include "common/math_util.h"
 
 /**
  * Abstraction class used to provide an interface between emulation code and the frontend
@@ -36,6 +37,23 @@ public:
         int     res_width;
         int     res_height;
         std::pair<unsigned,unsigned> min_client_area_size;
+    };
+
+    /// Describes the layout of the window framebuffer (size and top/bottom screen positions)
+    struct FramebufferLayout {
+
+        /**
+         * Factory method for constructing a default FramebufferLayout
+         * @param width Window framebuffer width in pixels
+         * @param height Window framebuffer height in pixels
+         * @return Newly created FramebufferLayout object with default screen regions initialized
+         */
+        static FramebufferLayout DefaultScreenLayout(int width, int height);
+
+        unsigned width;
+        unsigned height;
+        MathUtil::Rectangle<unsigned> top_screen;
+        MathUtil::Rectangle<unsigned> bottom_screen;
     };
 
     /// Swap buffers to display the next frame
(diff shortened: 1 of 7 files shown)


purpasmart96:
Yeah this is good, no since duplicating code.

src/common/emu_window.h
```diff
@@ -38,6 +39,17 @@ class EmuWindow
         std::pair<unsigned,unsigned> min_client_area_size;
     };
 
+    /// Describes the layout of the window framebuffer (size and top/bottom screen positions)
+    struct FramebufferLayout {
+        FramebufferLayout() : width(0), height(0), top_screen({}), bottom_screen({}) { }
+        FramebufferLayout(int width, int height);
+
+        int width;
```
neobrain: should be unsigned

src/citra_qt/main.cpp
```diff
@@ -193,7 +196,10 @@ void GMainWindow::OnDisplayTitleBars(bool show)
 void GMainWindow::BootGame(std::string filename)
 {
     LOG_INFO(Frontend, "Citra starting...\n");
-    System::Init(render_window);
+    System::Init();
+
+    VideoCore::g_renderer->SetWindow(render_window);
```
neobrain: Why was this moved out of System::Init?

src/common/emu_window.h
```diff
@@ -38,6 +39,17 @@ class EmuWindow
         std::pair<unsigned,unsigned> min_client_area_size;
     };
 
+    /// Describes the layout of the window framebuffer (size and top/bottom screen positions)
+    struct FramebufferLayout {
+        FramebufferLayout() : width(0), height(0), top_screen({}), bottom_screen({}) { }
+        FramebufferLayout(int width, int height);
```
neobrain: Please make this constructor a factory method instead, i.e. "static FramebufferLayout DefaultScreenLayout(int width, int height)". That way, it's more clear that the frontend is still free to chose its own screen layout.

src/common/emu_window.h
```diff
@@ -59,6 +68,14 @@ class EmuWindow
     static void KeyReleased(KeyMap::HostDeviceKey key);
 
     /**
+     * Factory method for constructing a default FramebufferLayout
+     * @param width Window framebuffer width in pixels
+     * @param height Window framebuffer height in pixels
+     * @return Newly created FramebufferLayout object with the default screen regions initialized
+     */
+    static FramebufferLayout DefaultScreenLayout(int width, int height);
```
neobrain: Sorry for not being clear, I was meaning to have that static function as part of FramebufferLayout. It doesn't really make a lot of sense to have a factory method as part of the parent class.

neobrain:
Added one minor nitpick, other than that LGTM.

PR #637, opened 2015-03-07
GPU/Textures: Fixed ETC texture decoding.

Subv:
The problem was a typo, coordinates with x < 2 should apparently use the first lookup table whereas our code was using the second table. The fix was taken from 3dmoo's code https://github.com/plutooo/3dmoo/commit/a57d27cabbb598110b518a6c6637e1265e1811b0#diff-db756bba03e234eb3ad60cf015e7d0adR802


archshift:
Ship it!

neobrain:
Nice catch ;)

PR #639, opened 2015-03-08
Build app bundles on OS X. Fixes the issue where the menubar would not appear.

archshift:
Previously, one would have to change focus to another program and then back to Citra in order for the menubar to appear. 

This also makes it easier to distribute builds in the future.


lioncash:
Looks good to me.

PR #640, opened 2015-03-08
default_ini.h: Put comments on their own lines

archshift:
Apparently inline comments is not necessarily standard in the INI format, and our parser was erroneously parsing the comments as values.


lioncash:
looks good to me

bunnei:
LGTM

PR #641, opened 2015-03-08
Services: Stubs and minor changes

purpasmart96:


src/core/hle/service/apt/apt.cpp
```diff
@@ -35,9 +35,14 @@ static Kernel::SharedPtr<Kernel::Event> notification_event; ///< APT notificatio
...
 static std::vector<u8> shared_font;
 
+static u32 percent = 0;
+
 void Initialize(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 app_id = cmd_buff[1];
+    u32 flags  = cmd_buff[2];
 
+    cmd_buff[2] = 0x04000000;
```
Subv: What is this value?

src/core/hle/service/apt/apt.cpp
```diff
@@ -35,9 +35,14 @@ static Kernel::SharedPtr<Kernel::Event> notification_event; ///< APT notificatio
 static Kernel::SharedPtr<Kernel::Event> pause_event = 0;    ///< APT pause event
 static std::vector<u8> shared_font;
 
+static u32 percent = 0;
```
Subv: Use a more descriptive name and add a comment for this please

src/core/hle/service/apt/apt.cpp
purpasmart96: See http://3dbrew.org/wiki/APTU:Initialize

Subv: Could you add a comment here please? it looks random at first glance

purpasmart96: Well there is nothing really to provide the comment since there is so little info on it, would it be fine if it just said "// From 3dbrew" ?

src/core/hle/service/ac_u.cpp
```diff
@@ -23,16 +48,118 @@ static void GetWifiStatus(Service::Interface* self) {
...
+ */
+static void GetInfraPriority(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    for (int i = 0; i < 0x200; i++) {
+        Memory::Read8(i + cmd_buff[2]);
+    }
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    cmd_buff[2] = 0; // ?
```
bunnei: Why are you setting this to zero if the purpose is unknown? If there is some reason, please put a well written TODO in here.

src/core/hle/service/ac_u.cpp
```diff
@@ -23,16 +48,118 @@ static void GetWifiStatus(Service::Interface* self) {
...
+ *  Outputs:
+ *      0 : Return header
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : Unknown
+ */
+static void GetInfraPriority(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    for (int i = 0; i < 0x200; i++) {
+        Memory::Read8(i + cmd_buff[2]);
```
bunnei: What is the purpose of this function? Has this behavior been tested (if not, where did you get this impl from?)

src/core/hle/service/ac_u.cpp
```diff
@@ -23,16 +48,118 @@ static void GetWifiStatus(Service::Interface* self) {
...
+ *      0 : Return header
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void SetRequestEulaVersion(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 unk1 = cmd_buff[1];
+    u32 unk2 = cmd_buff[2];
+
+    for (int i = 0; i < 0x200; i++) {
+        Memory::Read8(i + cmd_buff[4]);
```
bunnei: What is the purpose of this function? Has this behavior been tested (if not, where did you get this impl from?)

src/core/hle/service/ac_u.cpp
bunnei: NOTE: If it wasn't clear, please add a description of the purpose of this function to the docstring.

src/core/hle/service/ac_u.cpp
```diff
@@ -23,16 +48,118 @@ static void GetWifiStatus(Service::Interface* self) {
...
+ *      3 : Unknown
+ *  Outputs:
+ *      0 : Return header
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void SetClientVersion(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 unk1 = cmd_buff[1];
+    u32 unk2 = cmd_buff[2];
+    u32 unk3 = cmd_buff[3];
```
bunnei: Same comment as above, where was this functionality taken from? What does it mean? Please document in the docstring.

src/core/hle/service/apt/apt.cpp
```diff
@@ -35,9 +35,14 @@ static Kernel::SharedPtr<Kernel::Event> notification_event; ///< APT notificatio
 static Kernel::SharedPtr<Kernel::Event> pause_event = 0;    ///< APT pause event
 static std::vector<u8> shared_font;
 
+static u32 cpu_percent = 0; ///< CPU time available to the running application
```
bunnei: Is this really the default value at run time? Can we check this?
Also, variables like this should be initialized at runtime (e.g. in the `Interface::Interface` constructor). We need to be able to support running multiple apps in a row in the same instance of the emulator without carrying over state from the previous run.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -84,6 +84,30 @@ static void GetSemaphoreEventHandle(Service::Interface* self) {
...
+ *      3 : Value 0, some descriptor for the KProcess Handle
+ *      4 : KProcess handle
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void FlushDataCache(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 address = cmd_buff[1];
+    u32 size    = cmd_buff[2];
+    u32 process = cmd_buff[4];
```
bunnei: Should we be logging these arguments (probably as a debug or trace so there typically isn't spam)? Otherwise, this is just going to generate a compiler warning because they are unused local variables.

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -520,6 +522,51 @@ static void CardSlotIsInserted(Service::Interface* self) {
     LOG_WARNING(Service_FS, "(STUBBED) called");
 }
 
+static void InitializeWithSdkVersion(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 unk1 = cmd_buff[1];
+    u32 unk2 = cmd_buff[2];
+    u32 unk3 = cmd_buff[3];
```
bunnei: Same comment as above, please log these below.

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -20,6 +20,8 @@ using Kernel::Session;
 namespace Service {
 namespace FS {
 
+static u32 priority = 0;
```
bunnei: Same comment as with `cpu_percent`. 1) Is this default value verified? 2) Please initialize in the appropriate place.

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -520,6 +522,51 @@ static void CardSlotIsInserted(Service::Interface* self) {
...
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void SetPriority(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    priority = cmd_buff[1];
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+
+    LOG_DEBUG(Service_FS, "called priority=0x%08X", priority);
```
bunnei: What is this setting the priority of? Is this actually still a stub given the current functionality? (Guessing probably, since it doesn't seem like we do anything with priority other than pass it back to the app)

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -520,6 +522,51 @@ static void CardSlotIsInserted(Service::Interface* self) {
...
+ *      0 : 0x08630000
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : priority
+ */
+static void GetPriority(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    cmd_buff[2] = priority;
```
bunnei: What is this getting the priority of? Is this actually still a stub given the current functionality? (Guessing probably, since it doesn't seem like we do anything with priority other than pass it back to the app)

NOTE: These sorts of things should probably be documented in the docstrings if possible.

src/core/hle/service/nwm_uds.cpp
```diff
@@ -10,21 +11,112 @@
...
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : Value 0
+ *      3 : Output handle
+ */
+static void Initialize(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 unk1   = cmd_buff[1];
+    u32 unk2   = cmd_buff[12];
+    u32 value  = cmd_buff[13];
+    u32 handle = cmd_buff[14];
```
bunnei: Please log these below

src/core/hle/service/nwm_uds.cpp
```diff
@@ -10,21 +11,112 @@
...
+    u32 unk2            = cmd_buff[3];
+    u32 max_buffer_size = cmd_buff[4];
+    u32 mac_address     = cmd_buff[5];
+    u32 unk3            = cmd_buff[6];
+
+    u32 wlan_comm_id    = cmd_buff[15];
+    u32 ctr_gen_id      = cmd_buff[16];
+    u32 value           = cmd_buff[17];
+    u32 input_handle    = cmd_buff[18];
+    u32 new_buffer_size = cmd_buff[19];
```
bunnei: Please log these below

src/core/hle/service/fs/fs_user.cpp
purpasmart96: If I Knew what it did exactly it would be in the doc string. Same applies to the other functions

bunnei:
Left a lot of comments. Nothing seriously flawed with this PR, but you tried to do so much in a single PR (and commit even) that it'll probably take you a bit of time to address all of these. In the future, please break this sort of thing up into smaller pieces. That way, there is less burden on me to review, and less burden on you to have to address so much all at once :) (and therefore we can get some of these things merged in sooner). Bonus points if you feel like splitting this PR up, but I won't make you so long as my feedback is addressed.

src/core/hle/service/ac_u.cpp
purpasmart96: 3dmoo

src/core/hle/service/ac_u.cpp
purpasmart96: It's from 3dmoo

src/core/hle/service/fs/fs_user.cpp
purpasmart96: There isn't really a default value I know of, it seems the priority is 0 most of the times I see in citra for the service function.

src/core/hle/service/apt/apt.cpp
purpasmart96: There is no default value, Values are 5 to 80, but there really isn't a default value, the zero is just there, because any number would be fine since it will always be changed the service function is called during runtime

src/core/hle/service/fs/fs_user.cpp
purpasmart96: If I knew what they did I would of put it in the doc header

purpasmart96:
@bunnei I can move the AC_U stuff to another PR  :)

src/core/hle/service/fs/fs_user.cpp
bunnei: Where did you get this info from? 3dbrew, or 3dmoo? Please add that to the comment. Also,

src/core/hle/service/nwm_uds.cpp
```diff
@@ -10,21 +11,117 @@
 
 namespace NWM_UDS {
 
+static Kernel::SharedPtr<Kernel::Event> handle_event;
```
bunnei: Initialize as nullptr on emulator boot

src/core/hle/service/apt/apt.cpp
```diff
@@ -35,9 +35,14 @@ static Kernel::SharedPtr<Kernel::Event> notification_event; ///< APT notificatio
 static Kernel::SharedPtr<Kernel::Event> pause_event = 0;    ///< APT pause event
 static std::vector<u8> shared_font;
 
+static u32 cpu_percent; ///< CPU time available to the running application
```
bunnei: Doesn't seem like you addressed my comment from the last review, this should be initialized at emulator boot

src/core/hle/service/fs/fs_user.cpp
purpasmart96: 3dmoo

src/core/hle/service/nwm_uds.cpp
purpasmart96: addressed

src/core/hle/service/nwm_uds.cpp
purpasmart96: addressed

src/core/hle/service/nwm_uds.cpp
purpasmart96: How heck would you do that with this?

Subv: = nullptr inside the Interface constructor, at least for now

purpasmart96: Oh, its more simple then I thought...

src/core/hle/service/apt/apt.cpp
bunnei: Please initialize it to whatever value is "typical" in Init(). E.g. what is this set to if you call GetAppCpuTimeLimit without calling SetAppCpuTimeLimit? Surely it's some non-zero result that is valid...

src/core/hle/service/fs/fs_user.cpp
bunnei: Please document that this behavior is from 3dmoo (in the code) and that it's unverified on hardware.

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -606,15 +654,17 @@ const Interface::FunctionInfo FunctionTable[] = {
...
+    {0x08620040, SetPriority,           "SetPriority"},
+    {0x08630000, GetPriority,           "GetPriority"},
 };
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Interface class
 
 Interface::Interface() {
+
+    priority = 0;
```
bunnei: As with `cpu_percent` above, can we read out what the default priority is without setting it?

src/core/hle/service/apt/apt.cpp
```diff
@@ -28,16 +28,21 @@ namespace APT {
...
 /// Handle to shared memory region designated to for shared system font
-static Kernel::SharedPtr<Kernel::SharedMemory> shared_font_mem;
+static Kernel::SharedPtr<Kernel::SharedMemory> shared_font_mem = nullptr;
 
-static Kernel::SharedPtr<Kernel::Mutex> lock;
-static Kernel::SharedPtr<Kernel::Event> notification_event; ///< APT notification event
-static Kernel::SharedPtr<Kernel::Event> pause_event = 0;    ///< APT pause event
+static Kernel::SharedPtr<Kernel::Mutex> lock = nullptr;
+static Kernel::SharedPtr<Kernel::Event> notification_event = nullptr; ///< APT notification event
+static Kernel::SharedPtr<Kernel::Event> pause_event = nullptr;    ///< APT pause event
```
bunnei: Fix comment alignment (sorry to nitpick, but they were lined up before. It's OK to align them, and it's OK to just make them a single ` ` from the `;`, but right now it just looks silly)

bunnei:
Added a few more comments. The biggest issue is some unverified default values IMO. Some tests are pending before we can merge this PR.

src/core/hle/service/service.cpp
```diff
@@ -69,7 +70,7 @@ void Init() {
 
     Service::FS::ArchiveInit();
     Service::CFG::CFGInit();
-    Service::APT::APTInit();
+    Service::APT::Init();
```
bunnei: When I asked you to make this change, I meant to change all of them because of the redundancy. E.g. `Service::CFG::CFGInit()` -> `Service::CFG::Init()`, `Service::PTM::PTMInit()` -> `Service::PTM::Init()`, etc. The same goes for the Shutdown methods below. The only one that shouldn't be renamed is `ArchiveInit`/`ArchiveShutdown`, although if you feel like it, it would probably make sense to add a fs.h/fs.cpp to hle/service/fs/ and add the functions for `Service::FS::Init()` and `Service::FS::Shutdown()` that call `ArchiveInit` and `ArchiveShutdown`, respectively.

bunnei:
Added one more comment... I promise, after that it's all set :)

purpasmart96:
ok

bunnei:
Woo LGTM

PR #642, opened 2015-03-08
Touchpad support

bunnei:
Adds (preliminary) touchpad support. Tested with homebrew, as well as OoT with both frontends:

I'm calling it "preliminary" because I just implemented the basics to make it work, there are some things in HID shared memory that I'm not emulating (pointed out in the TODOs), and furthermore there might be some behavioral subtleties that I missed (test feedback is appreciated!)

### DIFF
--- a/src/citra/emu_window/emu_window_glfw.cpp
+++ b/src/citra/emu_window/emu_window_glfw.cpp
@@ -16,35 +16,36 @@ EmuWindow_GLFW* EmuWindow_GLFW::GetEmuWindow(GLFWwindow* win) {
     return static_cast<EmuWindow_GLFW*>(glfwGetWindowUserPointer(win));
 }
 
-void EmuWindow_GLFW::OnMouseButtonEvent(GLFWwindow* window, int button, int action, int mods) {
+void EmuWindow_GLFW::OnMouseButtonEvent(GLFWwindow* win, int button, int action, int mods) {
     if (button == GLFW_MOUSE_BUTTON_LEFT) {
-        auto layout = GetEmuWindow(window)->GetFramebufferLayout();
+        auto emu_window = GetEmuWindow(win);
+        auto layout = emu_window->GetFramebufferLayout();
         double x, y;
-        glfwGetCursorPos(window, &x, &y);
+        glfwGetCursorPos(win, &x, &y);
 
         if (action == GLFW_PRESS) {
-            EmuWindow::TouchPressed(layout, static_cast<u16>(x), static_cast<u16>(y));
+            emu_window->TouchPressed(layout, static_cast<u16>(x), static_cast<u16>(y));
         } else if (action == GLFW_RELEASE) {
-            EmuWindow::TouchReleased(layout, static_cast<u16>(x), static_cast<u16>(y));
+            emu_window->TouchReleased(layout, static_cast<u16>(x), static_cast<u16>(y));
         }
     }
 }
 
-void EmuWindow_GLFW::OnCursorPosEvent(GLFWwindow* window, double x, double y) {
-
-    auto layout = GetEmuWindow(window)->GetFramebufferLayout();
-    EmuWindow::TouchMoved(layout, static_cast<u16>(x), static_cast<u16>(y));
+void EmuWindow_GLFW::OnCursorPosEvent(GLFWwindow* win, double x, double y) {
+    auto emu_window = GetEmuWindow(win);
+    auto layout = emu_window->GetFramebufferLayout();
+    emu_window->TouchMoved(layout, static_cast<u16>(x), static_cast<u16>(y));
 }
 
 /// Called by GLFW when a key event occurs
 void EmuWindow_GLFW::OnKeyEvent(GLFWwindow* win, int key, int scancode, int action, int mods) {
-
-    int keyboard_id = GetEmuWindow(win)->keyboard_id;
+    auto emu_window = GetEmuWindow(win);
+    int keyboard_id = emu_window->keyboard_id;
 
     if (action == GLFW_PRESS) {
-        EmuWindow::KeyPressed({key, keyboard_id});
+        emu_window->KeyPressed({key, keyboard_id});
     } else if (action == GLFW_RELEASE) {
-        EmuWindow::KeyReleased({key, keyboard_id});
+        emu_window->KeyReleased({ key, keyboard_id });
     }
 
     Service::HID::PadUpdateComplete();
--- a/src/citra/emu_window/emu_window_glfw.cpp
+++ b/src/citra/emu_window/emu_window_glfw.cpp
@@ -16,6 +16,26 @@ EmuWindow_GLFW* EmuWindow_GLFW::GetEmuWindow(GLFWwindow* win) {
     return static_cast<EmuWindow_GLFW*>(glfwGetWindowUserPointer(win));
 }
 
+void EmuWindow_GLFW::OnMouseButtonEvent(GLFWwindow* window, int button, int action, int mods) {
+    if (button == GLFW_MOUSE_BUTTON_LEFT) {
+        auto layout = GetEmuWindow(window)->GetFramebufferLayout();
+        double x, y;
+        glfwGetCursorPos(window, &x, &y);
+
+        if (action == GLFW_PRESS) {
+            EmuWindow::TouchPressed(layout, static_cast<u16>(x), static_cast<u16>(y));
+        } else if (action == GLFW_RELEASE) {
+            EmuWindow::TouchReleased(layout, static_cast<u16>(x), static_cast<u16>(y));
+        }
+    }
+}
+
+void EmuWindow_GLFW::OnCursorPosEvent(GLFWwindow* window, double x, double y) {
+
+    auto layout = GetEmuWindow(window)->GetFramebufferLayout();
+    EmuWindow::TouchMoved(layout, static_cast<u16>(x), static_cast<u16>(y));
+}
+
 /// Called by GLFW when a key event occurs
 void EmuWindow_GLFW::OnKeyEvent(GLFWwindow* win, int key, int scancode, int action, int mods) {
 
@@ -88,6 +108,8 @@ EmuWindow_GLFW::EmuWindow_GLFW() {
 
     // Setup callbacks
     glfwSetKeyCallback(m_render_window, OnKeyEvent);
+    glfwSetMouseButtonCallback(m_render_window, OnMouseButtonEvent);
+    glfwSetCursorPosCallback(m_render_window, OnCursorPosEvent);
     glfwSetFramebufferSizeCallback(m_render_window, OnFramebufferResizeEvent);
     glfwSetWindowSizeCallback(m_render_window, OnClientAreaResizeEvent);
 
--- a/src/citra/emu_window/emu_window_glfw.cpp
+++ b/src/citra/emu_window/emu_window_glfw.cpp
@@ -23,18 +23,15 @@ void EmuWindow_GLFW::OnMouseButtonEvent(GLFWwindow* win, int button, int action,
         double x, y;
         glfwGetCursorPos(win, &x, &y);
 
-        if (action == GLFW_PRESS) {
-            emu_window->TouchPressed(layout, static_cast<u16>(x), static_cast<u16>(y));
-        } else if (action == GLFW_RELEASE) {
-            emu_window->TouchReleased(layout, static_cast<u16>(x), static_cast<u16>(y));
-        }
+        if (action == GLFW_PRESS)
+            emu_window->TouchPressed(static_cast<unsigned>(x), static_cast<unsigned>(y));
+        else if (action == GLFW_RELEASE)
+            emu_window->TouchReleased();
     }
 }
 
 void EmuWindow_GLFW::OnCursorPosEvent(GLFWwindow* win, double x, double y) {
-    auto emu_window = GetEmuWindow(win);
-    auto layout = emu_window->GetFramebufferLayout();
-    emu_window->TouchMoved(layout, static_cast<u16>(x), static_cast<u16>(y));
+    GetEmuWindow(win)->TouchMoved(static_cast<unsigned>(x), static_cast<unsigned>(y));
 }
 
 /// Called by GLFW when a key event occurs
@@ -45,10 +42,8 @@ void EmuWindow_GLFW::OnKeyEvent(GLFWwindow* win, int key, int scancode, int acti
     if (action == GLFW_PRESS) {
         emu_window->KeyPressed({key, keyboard_id});
     } else if (action == GLFW_RELEASE) {
-        emu_window->KeyReleased({ key, keyboard_id });
+        emu_window->KeyReleased({key, keyboard_id});
     }
-
-    Service::HID::PadUpdateComplete();
 }
 
 /// Whether the window is still open, and a close request hasn't yet been sent
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -278,6 +278,31 @@ void GRenderWindow::keyReleaseEvent(QKeyEvent* event)
     Service::HID::PadUpdateComplete();
 }
 
+void GRenderWindow::mousePressEvent(QMouseEvent *event)
+{
+    if (event->button() == Qt::LeftButton)  {
+        auto pos = event->pos();
+        EmuWindow::TouchPressed(GetFramebufferLayout(), static_cast<u16>(pos.x()),
+            static_cast<u16>(pos.y()));
+    }
+}
+
+void GRenderWindow::mouseMoveEvent(QMouseEvent *event)
+{
+    auto pos = event->pos();
+    EmuWindow::TouchMoved(GetFramebufferLayout(), static_cast<u16>(pos.x()),
+        static_cast<u16>(pos.y()));
+}
+
+void GRenderWindow::mouseReleaseEvent(QMouseEvent *event)
+{
+    if (event->button() == Qt::LeftButton)  {
+        auto pos = event->pos();
+        EmuWindow::TouchReleased(GetFramebufferLayout(), static_cast<u16>(pos.x()),
+            static_cast<u16>(pos.y()));
+    }
+}
+
 void GRenderWindow::ReloadSetKeymaps()
 {
     KeyMap::SetKeyMapping({Settings::values.pad_a_key,      keyboard_id}, Service::HID::PAD_A);
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -268,39 +268,33 @@ QByteArray GRenderWindow::saveGeometry()
 
 void GRenderWindow::keyPressEvent(QKeyEvent* event)
 {
-    EmuWindow::KeyPressed({event->key(), keyboard_id});
-    Service::HID::PadUpdateComplete();
+    this->KeyPressed({event->key(), keyboard_id});
 }
 
 void GRenderWindow::keyReleaseEvent(QKeyEvent* event)
 {
-    EmuWindow::KeyReleased({event->key(), keyboard_id});
-    Service::HID::PadUpdateComplete();
+    this->KeyReleased({event->key(), keyboard_id});
 }
 
 void GRenderWindow::mousePressEvent(QMouseEvent *event)
 {
-    if (event->button() == Qt::LeftButton)  {
+    if (event->button() == Qt::LeftButton)
+    {
         auto pos = event->pos();
-        EmuWindow::TouchPressed(GetFramebufferLayout(), static_cast<u16>(pos.x()),
-            static_cast<u16>(pos.y()));
+        this->TouchPressed(static_cast<unsigned>(pos.x()), static_cast<unsigned>(pos.y()));
     }
 }
 
 void GRenderWindow::mouseMoveEvent(QMouseEvent *event)
 {
     auto pos = event->pos();
-    EmuWindow::TouchMoved(GetFramebufferLayout(), static_cast<u16>(pos.x()),
-        static_cast<u16>(pos.y()));
+    this->TouchMoved(static_cast<unsigned>(pos.x()), static_cast<unsigned>(pos.y()));
 }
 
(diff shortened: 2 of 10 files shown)


src/common/emu_window.h
```diff
@@ -77,6 +77,33 @@ class EmuWindow
     static void KeyReleased(KeyMap::HostDeviceKey key);
 
     /**
+     * Signal that a touch pressed event has occurred (e.g. mouse click pressed)
+     * @param layout FramebufferLayout object describing the framebuffer size and screen positions
+     * @param framebuffer_x Framebuffer x-coordinate that was pressed
+     * @param framebuffer_y Framebuffer y-coordinate that was pressed
+     */
+    static void TouchPressed(const FramebufferLayout& layout, unsigned framebuffer_x,
+        unsigned framebuffer_y);
```
neobrain: ... btw, I'd prefer if we changed our coding style such that those parameters are to be aligned properly, i.e.

    static void TouchPressed(const FramebufferLayout& layout, unsigned framebuffer_x,
                             unsigned framebuffer_y

src/common/emu_window.h
```diff
@@ -77,6 +77,33 @@ class EmuWindow
...
+
+    /**
+     * Signal that a touch released event has occurred (e.g. mouse click released)
+     * @param layout FramebufferLayout object describing the framebuffer size and screen positions
+     * @param framebuffer_x Framebuffer x-coordinate that was released
+     * @param framebuffer_y Framebuffer y-coordinate that was released
+     */
+    static void TouchReleased(const FramebufferLayout& layout, unsigned framebuffer_x,
```
neobrain: Since this function performs memory access across threads (called from GUI thread, accesses emu thread data), it needs documentation on thread-safety (or lack of such, depending on what you intended for that).

src/core/hle/service/hid/hid.cpp
```diff
@@ -123,13 +115,57 @@ void PadUpdateComplete() {
...
+        shared_mem->pad.index_reset_ticks = (s64)Core::g_app_core->GetTicks();
+    }
+
+    // Signal both handles when there's an update to Pad or touch
+    g_event_pad_or_touch_1->Signal();
+    g_event_pad_or_touch_2->Signal();
+}
+
+void TouchPress(u16 x, u16 y) {
+    next_touch_x = x;
```
neobrain: Could we change the HID design to query the current touch state from EmuWindow, instead of introducing ugly state changing functions which are called on the wrong thread? (TouchPress changes emu thread state, but is called from the GUI thread)

src/core/hle/service/hid/hid.h
```diff
@@ -140,11 +162,37 @@ const PadState PAD_CIRCLE_DOWN  = {{1u << 31}};
...
+ * key changes and analog circle Pad changes.
+ */
 void PadUpdateComplete();
 
+/**
+ * Signal that the touchpad has been pressed
+ * @param x Touchpad x-coordinate in bottom screen pixels (between 0 and 320)
+ * @param y Touchpad y-coordinate in bottom screen pixels (between 0 and 240)
+ */
+void TouchPress(u16 x, u16 y);
```
neobrain: Needs documentation on thread-safety (or lack thereof)

src/core/hle/service/hid/hid.h
```diff
@@ -77,22 +77,44 @@ struct PadDataEntry {
...
+ * Structure of a single entry of touch state history within HID shared memory
  */
-struct PadData {
-    s64 index_reset_ticks;
-    s64 index_reset_ticks_previous;
-    u32 index; // the index of the last updated Pad state history element
+struct TouchDataEntry {
+    u16 x;
+    u16 y;
+    u32 data_valid;
```
neobrain: Under what circumstances would this field indicate "invalid data"? Maybe needs a comment.

Why is `data_valid` not a BitField?

src/core/hle/service/hid/hid.h
```diff
@@ -77,22 +77,44 @@ struct PadDataEntry {
...
+    struct {
+        s64 index_reset_ticks;
+        s64 index_reset_ticks_previous;
+        u32 index; // Index of the last updated pad state history element
+
+        INSERT_PADDING_BYTES(0x8);
 
-    u32 pad1;
-    u32 pad2;
+        PadState current_state; // Same as entries[index].current_state
```
neobrain: How specifically is it the "same", even though they are stored at different locations in memory?

src/core/hle/service/hid/hid.h
```diff
@@ -77,22 +77,44 @@ struct PadDataEntry {
...
+
+/**
+ * Structure of data stored in HID shared memory
+ */
+struct SharedMem {
+    // Offset 0x0 : "PAD" data, this is used for buttons and the circle pad
+    struct {
+        s64 index_reset_ticks;
+        s64 index_reset_ticks_previous;
+        u32 index; // Index of the last updated pad state history element
```
neobrain: Could you clarify what "pad state history element" is referring to? I.e. an actual member variable name?

src/core/hle/service/hid/hid.h
```diff
@@ -77,22 +77,44 @@ struct PadDataEntry {
...
+    u32 data_valid;
+};
+
+/**
+ * Structure of data stored in HID shared memory
+ */
+struct SharedMem {
+    // Offset 0x0 : "PAD" data, this is used for buttons and the circle pad
+    struct {
+        s64 index_reset_ticks;
```
neobrain: Could you add a comment explaining what this does?

src/core/hle/service/hid/hid.h
```diff
@@ -77,22 +77,44 @@ struct PadDataEntry {
...
+        s64 index_reset_ticks_previous;
+        u32 index; // Index of the last updated pad state history element
+
+        INSERT_PADDING_BYTES(0x8);
 
-    u32 pad1;
-    u32 pad2;
+        PadState current_state; // Same as entries[index].current_state
+        u32 raw_circle_pad_data;
```
neobrain: How is it "raw"? This sounds like it should be put in a union with a bunch of BitFields. Either that, or I simply don't get why this is called "raw".

src/core/hle/service/hid/hid.h
```diff
@@ -77,22 +77,44 @@ struct PadDataEntry {
...
-    PadState current_state; // same as entries[index].current_state
-    u32 raw_circle_pad_data;
+        INSERT_PADDING_BYTES(0x4);
 
-    u32 pad3;
+        std::array<PadDataEntry, 8> entries; // Pad state history
+    } pad;
 
-    std::array<PadDataEntry, 8> entries; // Pad state history
+    // Offset 0xA8 : Touchpad data, this is used for touchpad input
```
neobrain: Please use `static_assert` (cf. pica.h) when documenting offsets explicitly. There's little value in documenting offsets at all, and a `static_assert` will at least make sure that it's actually the correct offset.

src/core/hle/service/hid/hid.cpp
```diff
@@ -123,13 +115,57 @@ void PadUpdateComplete() {
...
+}
+
+void TouchUpdateComplete() {
+    SharedMem* shared_mem = GetSharedMem();
+
+    if (shared_mem == nullptr)
+        return;
+
+    shared_mem->touch.index = next_touch_index;
+    next_touch_index = (next_touch_index + 1) % shared_mem->touch.entries.size();
```
neobrain: `++next_touch_index %= shared_mem->touch.entries.size()`

src/core/hle/service/hid/hid.cpp
```diff
@@ -123,13 +115,57 @@ void PadUpdateComplete() {
...
+
+void TouchRelease() {
+    next_touch_x = 0;
+    next_touch_y = 0;
+}
+
+void TouchUpdateComplete() {
+    SharedMem* shared_mem = GetSharedMem();
+
+    if (shared_mem == nullptr)
```
neobrain: Ugh.

src/core/hle/service/hid/hid.cpp
```diff
@@ -25,17 +25,21 @@ Kernel::SharedPtr<Kernel::Event> g_event_debug_pad;
 
 // Next Pad state update information
 static PadState next_state = {{0}};
-static u32 next_index = 0;
-static s16 next_circle_x = 0;
-static s16 next_circle_y = 0;
+static u32 next_pad_index = 0;
```
neobrain: If we keep adding statically managed memory all over the place like this, it'll become more and more annoying to actually fix up the emulator to support things like state saving, or even more crucial things like being able to run two different applications without having to restart Citra.

src/common/emu_window.h
```diff
@@ -168,4 +195,6 @@ class EmuWindow
 
     WindowConfig config;         ///< Internal configuration (changes pending for being applied in ProcessConfigurationChanges)
     WindowConfig active_config;  ///< Internal active configuration
+
+    static bool touch_pressed;   ///< True if touchpad area is currently pressed, otherwise false
```
neobrain: EmuWindow shall not use any static state.

src/core/hle/service/hid/hid.h
bunnei: Not sure, not my code, I just moved it into a union.

src/core/hle/service/hid/hid.h
bunnei: I'm not sure what this comment means, I did not write it.

src/core/hle/service/hid/hid.h
bunnei: Not sure, again, this isn't really anything to do with touchpad and was code that was already here (just indented with the union), hence the diff

src/core/hle/service/hid/hid.cpp
Subv: Is that valid code?

neobrain: Sure, why wouldn't it be? :)

Subv: You are right, for some reason I thought the preincrement converted next_touch_index into an rvalue

src/core/hle/service/hid/hid.cpp
```diff
@@ -51,85 +50,61 @@ static inline PadData* GetPadData() {
...
-    current_pad_entry->delta_additions.hex = additions.hex;
-    current_pad_entry->delta_removals.hex = removals.hex;
+    pad_entry->current_state.hex = state.hex;
+    pad_entry->delta_additions.hex = changed.hex & state.hex;
+    pad_entry->delta_removals.hex = changed.hex & old_state.hex;;
 
     // Set circle Pad
-    current_pad_entry->circle_pad_x = next_circle_x;
-    current_pad_entry->circle_pad_y = next_circle_y;
+    pad_entry->circle_pad_x = state.circle_left ? -0x9C : state.circle_right ? 0x9C : 0x0;
```
Subv: Can we have a constant for these values? (0x9C)

src/core/hle/service/hid/hid.h
```diff
@@ -77,24 +77,60 @@ struct PadDataEntry {
...
+ * Structure of a single entry of touch state history within HID shared memory
  */
-struct PadData {
-    s64 index_reset_ticks;
-    s64 index_reset_ticks_previous;
-    u32 index; // the index of the last updated Pad state history element
+struct TouchDataEntry {
+    u16 x;                   ///< Y-coordinate of a touchpad press on the lower screen
+    u16 y;                   ///< X-coordinate of a touchpad press on the lower screen
+    BitField<0,7,u32> valid; ///< Set to 1 when this entry contains actual X/Y data, otherwise 0
```
Subv: Space after comma :)

src/core/hle/service/hid/hid.h
```diff
@@ -77,24 +77,60 @@ struct PadDataEntry {
...
+};
+
+/**
+ * Structure of data stored in HID shared memory
+ */
+struct SharedMem {
+    // "Pad data, this is used for buttons and the circle pad
```
Subv: Extra "

Subv:
Mostly LGTM, just a few minor nitpicks here and there

bunnei:
Addressed comments, should be good to go.

purpasmart96:
LGTM

src/citra/emu_window/emu_window_glfw.cpp
```diff
@@ -16,18 +16,36 @@ EmuWindow_GLFW* EmuWindow_GLFW::GetEmuWindow(GLFWwindow* win) {
...
-    int keyboard_id = GetEmuWindow(win)->keyboard_id;
+    auto emu_window = GetEmuWindow(win);
+    int keyboard_id = emu_window->keyboard_id;
 
     if (action == GLFW_PRESS) {
-        EmuWindow::KeyPressed({key, keyboard_id});
+        emu_window->KeyPressed({key, keyboard_id});
     } else if (action == GLFW_RELEASE) {
-        EmuWindow::KeyReleased({key, keyboard_id});
+        emu_window->KeyReleased({ key, keyboard_id });
```
linkmauve: The added spaces in {} are incoherent with the same construct two lines earlier.

src/citra_qt/bootmanager.cpp
```diff
@@ -268,14 +268,34 @@ QByteArray GRenderWindow::saveGeometry()
 
 void GRenderWindow::keyPressEvent(QKeyEvent* event)
 {
-    EmuWindow::KeyPressed({event->key(), keyboard_id});
-    Service::HID::PadUpdateComplete();
+    this->KeyPressed({ event->key(), keyboard_id });
```
linkmauve: Same here, and at other places in this file.

linkmauve:
You should also set the touch bit in PadState (src/core/hle/service/hid/hid.h line 44) when the touchscreen is pressed. Aside from that, LGTM.

bunnei:
Addressed @linkmauve's comments.

src/citra_qt/bootmanager.cpp
```diff
@@ -268,14 +268,34 @@ QByteArray GRenderWindow::saveGeometry()
...
+    this->KeyReleased({event->key(), keyboard_id});
+}
+
+void GRenderWindow::mousePressEvent(QMouseEvent *event)
+{
+    if (event->button() == Qt::LeftButton)
+    {
+        auto pos = event->pos();
+        auto layout = GetFramebufferLayout();
+        this->TouchPressed(layout, static_cast<u16>(pos.x()), static_cast<u16>(pos.y()));
```
neobrain: Those casts should be to `unsigned`, not `u16`.

src/core/hle/service/hid/hid.cpp
```diff
@@ -51,90 +45,69 @@ static inline PadData* GetPadData() {
...
- */
-static void UpdateNextCirclePadState() {
-    static const s16 max_value = 0x9C;
-    next_circle_x = next_state.circle_left ? -max_value : 0x0;
-    next_circle_x += next_state.circle_right ? max_value : 0x0;
-    next_circle_y = next_state.circle_down ? -max_value : 0x0;
-    next_circle_y += next_state.circle_up ? max_value : 0x0;
-}
+void HIDUpdate() {
+    SharedMem* mem = reinterpret_cast<SharedMem*>(shared_mem->GetPointer().ValueOr(nullptr));
```
yuriks: Wtf? Nothing about this line makes any sense.

src/common/emu_window.cpp
```diff
@@ -6,18 +6,60 @@
...
+void EmuWindow::TouchReleased() {
+    touch_pressed = false;
+    touch_x = 0;
+    touch_y = 0;
+    pad_state.touch = 0;
+}
+
+void EmuWindow::TouchMoved(const FramebufferLayout& layout, unsigned framebuffer_x,
+                            unsigned framebuffer_y) {
+    if (touch_pressed) {
```
neobrain: try to keep indentation level low: `if(!touch_pressed) return;`

src/common/emu_window.h
```diff
@@ -71,10 +71,46 @@ class EmuWindow
...
+    /// Signal that a touch released event has occurred (e.g. mouse click released)
+    void TouchReleased();
+
+    /**
+     * Signal that a touch movement event has occurred (e.g. mouse was moved over the emu window)
+     * @param layout FramebufferLayout object describing the framebuffer size and screen positions
+     * @param framebuffer_x Framebuffer x-coordinate
+     * @param framebuffer_y Framebuffer y-coordinate
+     */
+    void TouchMoved(const FramebufferLayout& layout, unsigned framebuffer_x, unsigned framebuffer_y);
```
neobrain: Why does this take a reference to some FramebufferLayout?

src/common/emu_window.cpp
neobrain: Nevermind, I misunderstood the point of this function.

neobrain: Actually yes, you can still keep indentation level low by using `if (!touch_pressed) return;`

src/common/emu_window.h
```diff
@@ -71,10 +71,46 @@ class EmuWindow
...
+
+    /**
+     * Gets the current pad state (which buttons are pressed and the circle pad direction)
+     * @return PadState object indicating the current pad state
```
neobrain: Needs a note about intended use across threads / thread-safety (or lack thereof).

src/common/emu_window.h
```diff
@@ -168,4 +198,11 @@ class EmuWindow
 
     WindowConfig config;         ///< Internal configuration (changes pending for being applied in ProcessConfigurationChanges)
     WindowConfig active_config;  ///< Internal active configuration
+
+    bool touch_pressed;          ///< True if touchpad area is currently pressed, otherwise false
+
+    u16 touch_x;
```
neobrain: These should be documented to be the *native* touch coordinates (rather than ones in a stretched window or the like).

src/common/emu_window.cpp
```diff
@@ -6,18 +6,60 @@
...
+            (layout.bottom_screen.bottom - layout.bottom_screen.top);
+        touch_pressed = true;
+        pad_state.touch = 1;
+    }
 }
 
-EmuWindow::FramebufferLayout EmuWindow::FramebufferLayout::DefaultScreenLayout(unsigned width, unsigned height) {
+void EmuWindow::TouchReleased() {
+    touch_pressed = false;
+    touch_x = 0;
```
neobrain: Is there any reason to zero these out?

src/common/emu_window.h
```diff
@@ -71,10 +71,46 @@ class EmuWindow
...
+    const Service::HID::PadState& GetPadState() const {
+        return pad_state;
+    }
+
+    /**
+     * Gets the current touch screen state (touch X/Y coordinates and whether or not it is pressed)
+     * @return std::tuple of (x, y, pressed) where `x` and `y` are the touch coordinates and
+     *         `pressed` is true if the touch screen is currently being pressed
+     */
+    std::tuple<u16, u16, bool> GetTouchState() const {
```
neobrain: Should be a const tuple.

src/common/emu_window.cpp
Subv: Keeping the return in the same line as the if? I don't agree with that, what would be the purpose of this?

src/core/hle/service/hid/hid.cpp
```diff
@@ -51,90 +45,69 @@ static inline PadData* GetPadData() {
...
+    const PadState& state = VideoCore::g_emu_window->GetPadState();
```
neobrain: PadState is essentially just a u32, hence we shouldn't use a reference here. Given that we perform no synchronization at all, this is safer anyway since it guarantees that the data read throughout this function is consistent.

src/common/emu_window.h
```diff
@@ -71,10 +71,46 @@ class EmuWindow
...
-    static void KeyReleased(KeyMap::HostDeviceKey key);
+    void KeyReleased(KeyMap::HostDeviceKey key);
+
+    /**
+     * Signal that a touch pressed event has occurred (e.g. mouse click pressed)
+     * @param layout FramebufferLayout object describing the framebuffer size and screen positions
+     * @param framebuffer_x Framebuffer x-coordinate that was pressed
+     * @param framebuffer_y Framebuffer y-coordinate that was pressed
+     */
+    void TouchPressed(const FramebufferLayout& layout, unsigned framebuffer_x, unsigned framebuffer_y);
```
yuriks: What's the point of passing the `FramebufferLayout` here, if it can simply be gotten from the EmuWindow itself? The 3DS isn't likely to get itself an extra touchscreen anytime soon.

neobrain:
Mostly LGTM

src/common/emu_window.cpp
neobrain: nah, it's about returning early such that the rest of this function needs to be intended by one level less.

src/common/emu_window.cpp
bunnei: Just to initialize them so that they are not invalid... I don't see the harm?

bunnei: It's consistent with the system behavior, if the touch screen is not being pressed, the X and Y coordinates are 0

src/core/hle/service/hid/hid.h
```diff
@@ -77,24 +67,65 @@ struct PadDataEntry {
...
+        // TODO(bunnei): Implement `raw_entry` field
+        TouchDataEntry raw_entry; ///< Raw (analog) touch data, before being converted
+
+        std::array<TouchDataEntry, 8> entries; ///< Last 8 touch entries, in pixel coordinates
+    } touch;
 };
 
+// TODO: MSVC does not support using offsetof() on non-static data members even though this
+//       is technically allowed since C++11. This macro should be enabled once MSVC adds
+//       support for that.
```
lioncash: This should be fixed in VS 2015. So you can extend this saying it can be removed once we move over to it.

bunnei: I duplicated this comment from other areas in the codebase... I think it's fine to just leave it as-is for consistency. So we don't miss things like this when VS2015 is released, I opened #651.

bunnei:
Addressed all additional comments, and all builds passed (for some reason TravisCI isn't showing, see https://travis-ci.org/citra-emu/citra/builds/53869114, and AppVeyor failed after the build completed successfully due to a wget issue)

PR #643, opened 2015-03-08
GPU: Implemented more depth buffer formats.

Subv:
This fixes the horizontal lines in Picross E, Cubic Ninja, Cave Story 3D and possibly others.

### DIFF
--- a/src/citra_qt/debugger/graphics_framebuffer.cpp
+++ b/src/citra_qt/debugger/graphics_framebuffer.cpp
@@ -176,7 +176,7 @@ void GraphicsFramebufferWidget::OnUpdate()
     {
         // TODO: Store a reference to the registers in the debug context instead of accessing them directly...
 
-        auto framebuffer = Pica::registers.framebuffer;
+        const auto& framebuffer = Pica::registers.framebuffer;
 
         framebuffer_address = framebuffer.GetColorBufferPhysicalAddress();
         framebuffer_width = framebuffer.GetWidth();
@@ -189,7 +189,7 @@ void GraphicsFramebufferWidget::OnUpdate()
 
     case Source::DepthBuffer:
     {
-        auto framebuffer = Pica::registers.framebuffer;
+        const auto& framebuffer = Pica::registers.framebuffer;
 
         framebuffer_address = framebuffer.GetDepthBufferPhysicalAddress();
         framebuffer_width = framebuffer.GetWidth();
--- a/src/citra_qt/debugger/graphics_framebuffer.cpp
+++ b/src/citra_qt/debugger/graphics_framebuffer.cpp
@@ -27,6 +27,7 @@ GraphicsFramebufferWidget::GraphicsFramebufferWidget(std::shared_ptr<Pica::Debug
 
     framebuffer_source_list = new QComboBox;
     framebuffer_source_list->addItem(tr("Active Render Target"));
+    framebuffer_source_list->addItem(tr("Active Depth Buffer"));
     framebuffer_source_list->addItem(tr("Custom"));
     framebuffer_source_list->setCurrentIndex(static_cast<int>(framebuffer_source));
 
@@ -49,6 +50,9 @@ GraphicsFramebufferWidget::GraphicsFramebufferWidget(std::shared_ptr<Pica::Debug
     framebuffer_format_control->addItem(tr("RGB5A1"));
     framebuffer_format_control->addItem(tr("RGB565"));
     framebuffer_format_control->addItem(tr("RGBA4"));
+    framebuffer_format_control->addItem(tr("D16"));
+    framebuffer_format_control->addItem(tr("D24"));
+    framebuffer_format_control->addItem(tr("D24S8"));
 
     // TODO: This QLabel should shrink the image to the available space rather than just expanding...
     framebuffer_picture_label = new QLabel;
@@ -173,13 +177,24 @@ void GraphicsFramebufferWidget::OnUpdate()
         // TODO: Store a reference to the registers in the debug context instead of accessing them directly...
 
         auto framebuffer = Pica::registers.framebuffer;
-        using Framebuffer = decltype(framebuffer);
 
         framebuffer_address = framebuffer.GetColorBufferPhysicalAddress();
         framebuffer_width = framebuffer.GetWidth();
         framebuffer_height = framebuffer.GetHeight();
         // TODO: It's unknown how this format is actually specified
         framebuffer_format = Format::RGBA8;
+
+        break;
+    }
+
+    case Source::DepthBuffer:
+    {
+        auto framebuffer = Pica::registers.framebuffer;
+
+        framebuffer_address = framebuffer.GetDepthBufferPhysicalAddress();
+        framebuffer_width = framebuffer.GetWidth();
+        framebuffer_height = framebuffer.GetHeight();
+        framebuffer_format = Format::D16;
 
         break;
     }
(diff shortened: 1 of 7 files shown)


Subv:
This behavior seems to be wrong according to some hwtests, but the hwtests may be wrong, I wasn't able to get the second fill engine to work, and some things just don't make sense at all.
https://gist.github.com/Subv/0ccbf7e48f87554e346f

Subv:
Found the real problem, it was due to these games using the DEPTH24_STENCIL8 depth buffer format, which was not implemented

src/video_core/pica.h
```diff
@@ -418,6 +418,13 @@ struct Regs {
             RGBA4    = 4,
         };
 
+        enum DepthFormat : u32 {
+            DEPTH16 = 0,
```
purpasmart96: Names should be like "Depth16"  in enum"s with no underslashes

src/video_core/pica.h
```diff
@@ -418,6 +418,13 @@ struct Regs {
             RGBA4    = 4,
         };
 
+        enum DepthFormat : u32 {
+            D16 = 0,
+        
+            D24 = 2,
```
neobrain: Please align these at the `=`

src/video_core/rasterizer.cpp
```diff
@@ -100,23 +100,66 @@ static u32 GetDepth(int x, int y) {
...
+        case registers.framebuffer.D16:
+        {
+            u32 stride = registers.framebuffer.width * 2;
+            return *(u16_le*)(depth_buffer + VideoCore::GetMortonOffset(x, y, 2) + coarse_y * stride);
+        }
+        case registers.framebuffer.D24:
+        {
+            u32 stride = registers.framebuffer.width * 3;
+            u8* address = depth_buffer + VideoCore::GetMortonOffset(x, y, 3) + coarse_y * stride;
+            return (address[2] << 16) | (address[1] << 8) | address[0];
```
neobrain: Functions in color.h akin to how color formats are en-/decoded should be introduced for this.

neobrain:
If you feel like it, it would also be nice to have depth preview implemented in the framebuffer widget. For this purpose, what needs to be done is:
* `framebuffer_source_list` should have a new option called `Active depth buffer`
* `framebuffer_format_control` needs to have the depth formats added
* `GraphicsFramebufferWidget::Format` needs to have the depth formats added
* `GraphicsFramebufferWidget::OnUpdate` needs to implement the decoding logic appropriately (using the helper functions which should be added in color.h)

Since UI code only supports 8 bits per color component when displaying images, it would either be possible to show the depth buffer as 8-bit greyscale (i.e. sacrifizing the lower 16 bits of precision) OR to just visualize it by copying the 24 bits of depth into the RGB components directly.

Subv:
Updated to include the viewing code for the framebuffer widget.

src/citra_qt/debugger/graphics_framebuffer.cpp
```diff
@@ -184,6 +187,18 @@ void GraphicsFramebufferWidget::OnUpdate()
         break;
     }
 
+    case Source::DepthBuffer:
+    {
+        auto framebuffer = Pica::registers.framebuffer;
```
neobrain: should be a const reference (not sure why I went with a non-reference above)

src/citra_qt/debugger/graphics_framebuffer.cpp
```diff
@@ -240,3 +279,19 @@ void GraphicsFramebufferWidget::OnUpdate()
...
+        case Format::RGB8:
+        case Format::D24:
+            return 3;
+        case Format::RGB5A1:
+        case Format::RGB565:
+        case Format::RGBA4:
+        case Format::D16:
+            return 2;
+    }
+}
```
neobrain: missing newline at EOF

src/citra_qt/debugger/graphics_framebuffer.h
```diff
@@ -21,7 +21,8 @@ class GraphicsFramebufferWidget : public BreakPointObserverDock {
 
     enum class Source {
         PicaTarget = 0,
-        Custom = 1,
+        DepthBuffer = 1,
+        Custom = 2,
```
neobrain: Please align at `=`

src/core/hw/gpu.cpp
```diff
@@ -81,9 +81,9 @@ inline void Write(u32 addr, const T data) {
             if (config.fill_24bit) {
                 // fill with 24-bit values
                 for (u8* ptr = start; ptr < end; ptr += 3) {
-                    ptr[0] = config.value_24bit_b;
+                    ptr[0] = config.value_24bit_r;
```
neobrain: This is unrelated to the PR description, please move it to a separate PR.

src/video_core/color.h
```diff
@@ -101,6 +101,33 @@ inline const Math::Vec4<u8> DecodeRGBA4(const u8* bytes) {
...
+inline const u32 DecodeD16(const u8* bytes) {
+    return *reinterpret_cast<const u16_le*>(bytes);
+}
+
+/**
+ * Decode a depth value stored in D24 format
+ * @param bytes Pointer to encoded source value
+ * @return Depth value as an u32
+ */
+inline const u32 DecodeD24(const u8* bytes) {
```
neobrain: Primitive types need not be marged `const` in return values.

(the rationale in returning `const Math::Vec2<u32>` is that the caller might be thinking that the return value is a reference and treat it as such, e.g. doing stupid stuff like `DecodeD24S8(..).Normalize()`)

neobrain:
Added a bunch of comments, other than that LGTM

Subv:
Updated

src/video_core/color.h
```diff
@@ -153,4 +180,34 @@ inline void EncodeRGBA4(const Math::Vec4<u8>& color, u8* bytes) {
         (Convert8To4(color.g()) << 8) | (Convert8To4(color.b()) << 4) | Convert8To4(color.a());
 }
 
+/**
+ * Encode a depth value as D16 format
+ * @param value Source depth value to encode
+ * @param bytes Pointer where to store the encoded value
+ */
+inline void EncodeD16(u32 value, u8* bytes) {
+    *reinterpret_cast<u16_le*>(bytes) = value & 0xFFFF;
```
bunnei: What is the format of value? If its a 32-bit depth value, shouldn't it be value >> 16?

src/video_core/color.h
```diff
@@ -153,4 +180,34 @@ inline void EncodeRGBA4(const Math::Vec4<u8>& color, u8* bytes) {
...
+}
+
+/**
+ * Encode a depth value as D24 format
+ * @param value Source depth value to encode
+ * @param bytes Pointer where to store the encoded value
+ */
+inline void EncodeD24(u32 value, u8* bytes) {
```
bunnei: Again, assuming value here is probably already a 24-bit depth value? Please document if so

src/video_core/color.h
Subv: The value can be anything, but only 16 bits are encoded, this is only called with values with less or equal than 16 bits

src/video_core/color.h
Subv: Yes, the value doesn't change formats between Getdepth and Setdepth calls

src/video_core/color.h
bunnei: Then why not make value a u16 and document in the docstring that it expects a unsigned 16-bit depth component?

src/video_core/color.h
bunnei: Again, same comment as above. Please put it in the docstring. When I see a function declaration like `EncodeD24(u32 value, u8* bytes)`, I automatically assume value is a 32-bit component.

src/video_core/color.h
```diff
@@ -153,4 +180,34 @@ inline void EncodeRGBA4(const Math::Vec4<u8>& color, u8* bytes) {
...
+    bytes[0] = value & 0xFF;
+    bytes[1] = (value >> 8) & 0xFF;
+    bytes[2] = (value >> 16) & 0xFF;
+}
+
+/**
+ * Encode depth and stencil values as D24S8 format
+ * @param depth Source depth values to encode
```
bunnei: `24-bit source depth component to encode`

src/video_core/rasterizer.cpp
```diff
@@ -100,23 +100,65 @@ static u32 GetDepth(int x, int y) {
...
     
     const u32 coarse_y = y & ~7;
-    u32 stride = registers.framebuffer.width * 2;
 
-    // Assuming 16-bit depth buffer format until actual format handling is implemented
-    return *(u16*)(depth_buffer + VideoCore::GetMortonOffset(x, y, 2) + coarse_y * stride);
+    switch (registers.framebuffer.depth_format) {
+        case registers.framebuffer.D16:
+        {
+            u32 stride = registers.framebuffer.width * 2;
```
bunnei: Can we add a `bytes_per_pixel` variable like we did with setting/getting the color values?

src/video_core/color.h
Subv: Because then we'd need a cast in Setdepth and Getdepth for it, would that be OK?

Subv:
Updated

bunnei:
Squash please

Subv:
Squashed

bunnei:
LGTM

PR #645, opened 2015-03-09
Minor bugfixes to LDM/STM.

lioncash:
These variants are privileged instructions ("LDM/STM (User Register)" variants) so this doesn't affect anything a game could do. May as well be correct about it regardless.


bunnei:
LGTM

PR #648, opened 2015-03-09
GPU: Use the correct position for the finished bit in memory fills

Subv:
It was using the same bit position as the "trigger" bit.


neobrain:
LGTM

PR #652, opened 2015-03-12
Pica/VertexShader: Fix a bug caused due to incorrect assumptions of consecutive output register tables.

neobrain:
Credits go to @RHY3756547 for spotting this issue (I rewrote his fix, though, since some ideas came to my mind for cleaning up the shader emulation code).

This should still fix some issue in Super Monkey Ball, hopefully.

### DIFF
--- a/src/video_core/vertex_shader.cpp
+++ b/src/video_core/vertex_shader.cpp
@@ -72,7 +72,7 @@ struct VertexShaderState {
     u32* program_counter;
 
     const float24* input_register_table[16];
-    float24* output_register_table[7*4];
+    Math::Vec4<float24> output_registers[16];
 
     Math::Vec4<float24> temporary_registers[16];
     bool conditional_code[2];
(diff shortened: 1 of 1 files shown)


RHY3756547:
Yeah, this is definitely a lot cleaner than occasionally modifying vec4s of pointers instead of values for the destination. I can confirm that monkey ball still works, so long as you also include the change from command_processor.cpp:

https://github.com/RHY3756547/citra/commit/e91e9afe385bf6eaea88510873fa763b2edbe7e5#commitcomment-10171159

Obligatory screenshot:
image

neobrain:
Thanks for verifying the new patch :)

As for zeroing out the input attributes, I don't think this is something we want to do since it's more likely to hide issues. At least currently, when uninitialized vertex attributes are read, it's obvious by the immediate brokenness. However, Subv is looking into an issue, the fix for which will probably make that workaround obsolete hopefully.

RHY3756547:
I'll hold out for that, but I'm not sure keeping the same vertex attributes as the last submitted vertex is the correct behaviour either. Might be worth investigating in the future.

bunnei:
LGTM

PR #654, opened 2015-03-13
Port Dot3RGB and Dot3RGBA from zmm-progress

MoochMcGee:
This ports these two operations in a less broken way from bunnei's zmm-progress branch. I'm guessing they were needed for Majora's Mask. Anyway, they have been redone to better suit the GL spec.


src/video_core/rasterizer.cpp
```diff
@@ -502,6 +502,17 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         return result.Cast<u8>();
                     }
 
+                    case Operation::Dot3RGB:
+                    case Operation::Dot3RGBA:
```
neobrain: Where is the difference between these two?

Subv: In DOT3RGB the result is stored in the RGB components of the output vector, in DOT3RGBA the result is stored in the RGBA components

neobrain: But why is there no difference in the code?

neobrain: Nevermind, I think it's clear to me now.

Could we still have a commend explaining
* What the difference is between the RGB and the RGBA version
* What each of them is supposed to be doing in the first place
* Why they are not treated differently in this particular place.

src/video_core/rasterizer.cpp
```diff
@@ -502,6 +502,20 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                    //This computes the dot product of the RGB components according to GLES spec.
+                    //Dot3RGB just stores the dot product in the RGB components.
+                    //Dot3RGBA also stores it in the alpha component.
+                    case Operation::Dot3RGB:
+                    case Operation::Dot3RGBA:
+                    {
+                        int a[] = {input[0].r(), input[0].g(), input[0].b()};
+                        int b[] = {input[1].r(), input[1].g(), input[1].b()};
+
+                        auto res = std::inner_product(a, a + sizeof(a) / sizeof(a[0]), b, 0);
```
Subv: std::inner_product(a, std::end(a), b, 0);
Or better yet, use std::array for a and b instead of C arrays and call a.begin() and a.end()

src/video_core/rasterizer.cpp
```diff
@@ -545,6 +559,16 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                     case Operation::Subtract:
                         return std::max(0, (int)input[0] - (int)input[1]);
 
+                    case Operation::Dot3RGBA:
+                    {
+                        int a[] = {input[0].r(), input[0].g(), input[0].b(), input[0].a()};
```
Subv: Why is this also taking the alpha into account for the dot product?

src/video_core/rasterizer.cpp
Subv: Also why isn't there any distinction between Dot3RGB and Dot3RGBA?

src/video_core/rasterizer.cpp
```diff
@@ -502,6 +502,20 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         return result.Cast<u8>();
                     }
 
+                    //This computes the dot product of the RGB components according to GLES spec.
```
Subv: Space after each //

MoochMcGee:
All comments addressed.

src/video_core/rasterizer.cpp
MoochMcGee: Because they do the same thing for the RGB components. It's the treatment of the alpha component that differs.

purpasmart96:
Build failing on travis... you might want to something about that

MoochMcGee:
Closing because I don't see a way to fix this without invoking major changes to the TEV combiners.

neobrain:
@MoochMcGee Mind explaining what you're talking about?

PR #655, opened 2015-03-13
 HID: Proper Signal Interrupts for EnableAccelerometer & EnableGyroscopeLow along  with a stub for GetSoundVolume

purpasmart96:
This fixes the freeze when pressing the ocarina on the touch pad in OoT.

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -114,6 +114,7 @@ void GetIPCHandles(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
     cmd_buff[1] = 0; // No error
+    cmd_buff[2] = 0x14000000; // IPC Command Structure translate-header
     // TODO(yuriks): Return error from SendSyncRequest is this fails (part of IPC marshalling)
     cmd_buff[3] = Kernel::g_handle_table.Create(Service::HID::shared_mem).MoveFrom();
     cmd_buff[4] = Kernel::g_handle_table.Create(Service::HID::event_pad_or_touch_1).MoveFrom();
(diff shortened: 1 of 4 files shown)


archshift:
Exactly which part of this fixes the bug?

purpasmart96:
For some reason OoT calls the service function  EnableAccelerometer, which has a signal interrupt in it, before that interrupt wasn't being triggered. Also I added bunnei's Gyroscope interrupt event on Init, that should fix issues in other games such as Majora's mask which would let it boot further.

src/core/hle/service/hid/hid.cpp
```diff
@@ -123,6 +124,31 @@ void GetIPCHandles(Service::Interface* self) {
...
+    event_gyroscope->Signal();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+}
+
+void GetSoundVolume(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    cmd_buff[2] = 0x3F; // TODO(purpasmart): Find out if this is the max value for the volume
```
Subv: Please make a variable to hold the current volume level

purpasmart96:
Errrr... I think the name is good enough.

src/core/hle/service/hid/hid.cpp
```diff
@@ -123,6 +124,37 @@ void GetIPCHandles(Service::Interface* self) {
...
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+
+    LOG_WARNING(Service_HID, "(STUBBED) called");
+}
+
+void GetSoundVolume(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 volume = 0x3F; // TODO(purpasmart): Find out if this is the max value for the volume
```
bunnei: Should be const

bunnei:
LGTM

PR #657, opened 2015-03-14
GPU: Implemented the flip_data (bit 0) bit in display transfers.

Subv:
This is based on information on 3dbrew.

Tested with blargSNES, and this is the output, the top screen is now correctly flipped: http://i.imgur.com/yaoAzLj.png

### DIFF
--- a/src/core/hw/gpu.cpp
+++ b/src/core/hw/gpu.cpp
@@ -146,8 +146,17 @@ inline void Write(u32 addr, const T data) {
                 for (u32 x = 0; x < output_width; ++x) {
                     Math::Vec4<u8> src_color = { 0, 0, 0, 0 };
 
-                    u32 scaled_x = x * horizontal_scale;
-                    u32 scaled_y = y * vertical_scale;
+                    // Calculate the [x,y] position of the input image 
+                    // based on the current output position and the scale
+                    u32 input_x = x * horizontal_scale;
+                    u32 input_y = y * vertical_scale;
+
+                    if (config.flip_vertically) {
+                        // Flip the y value of the output data, 
+                        // we do this after calculating the [x,y] position of the input image 
+                        // to account for the scaling options.
+                        y = output_height - y - 1;
+                    }
 
                     u32 dst_bytes_per_pixel = GPU::Regs::BytesPerPixel(config.output_format);
                     u32 src_bytes_per_pixel = GPU::Regs::BytesPerPixel(config.input_format);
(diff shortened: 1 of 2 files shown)


src/core/hw/gpu.cpp
```diff
@@ -149,6 +149,13 @@ inline void Write(u32 addr, const T data) {
                     u32 scaled_x = x * horizontal_scale;
                     u32 scaled_y = y * vertical_scale;
 
+                    if (config.flip_data) {
+                        // Flip the y value of the output data, 
+                        // we do it after calculating the [x,y] position of the input image 
+                        // to account for the scaling options.
+                        y = output_height - y;
```
neobrain: Shouldn't this be `output_height - y - 1`?

src/core/hw/gpu.cpp
```diff
@@ -149,6 +149,13 @@ inline void Write(u32 addr, const T data) {
                     u32 scaled_x = x * horizontal_scale;
                     u32 scaled_y = y * vertical_scale;
 
+                    if (config.flip_data) {
+                        // Flip the y value of the output data, 
+                        // we do it after calculating the [x,y] position of the input image
```
neobrain: it => this

neobrain: Which variable stores the [x,y] pos?

src/core/hw/gpu.h
```diff
@@ -191,7 +191,7 @@ struct Regs {
         union {
             u32 flags;
 
-            BitField< 0, 1, u32> flip_data;        // flips input data horizontally (TODO) if true
+            BitField< 0, 1, u32> flip_data;        // flips input data vertically
```
neobrain: => `flip_vertically`

src/core/hw/gpu.cpp
neobrain: It's not obvious that "scaled_x" is the position of the input image (btw, is "of" even correct? What would that offset be relative to?). Should have a comment on top of the declaration of that variable.

Subv:
Updated

src/core/hw/gpu.h
```diff
@@ -191,7 +191,7 @@ struct Regs {
         union {
             u32 flags;
 
-            BitField< 0, 1, u32> flip_data;        // flips input data horizontally (TODO) if true
+            BitField< 0, 1, u32> flip_vertically;        // flips input data vertically
```
neobrain: comment alignment messed up

neobrain:
LGTM other than the misaligned comment.

Subv:
Updated

bunnei:
LGTM

PR #659, opened 2015-03-15
Implement SETEND.

lioncash:


src/core/arm/skyeye_common/armmmu.h
```diff
@@ -54,3 +58,82 @@ enum
 	XSCALE_CP15_AUX_CONTROL = 1,
 	XSCALE_CP15_COPRO_ACCESS = 15,
 };
+
+// Reads data in big/little endian format based on the
+// state of the E (endian) bit in the emulated CPU's APSR.
+template<typename T>
+inline T ReadMemory(ARMul_State* cpu, u32 address)
```
Subv: I think template specializations are better than the soup of std::is_same in this function

lioncash: Also should I just use Common::swap32, etc here, instead of the *_be variants? Doing that would eliminate the need for any of the explicit instantiations added in this PR, but I'm not sure what the preferred way of handling endianness would be.

lioncash:
Updated it to simply swap on big endian mode, instead of juggling instantiations.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -4537,8 +4552,10 @@ unsigned InterpreterMainLoop(ARMul_State* state) {
             // Should check if RD is even-numbered, Rd != 14, addr[0:1] == 0, (CP15_reg1_U == 1 || addr[2] == 0)
             inst_cream->get_addr(cpu, inst_cream->inst, addr, 1);
 
-            cpu->Reg[BITS(inst_cream->inst, 12, 15)] = Memory::Read32(addr);
-            cpu->Reg[BITS(inst_cream->inst, 12, 15) + 1] = Memory::Read32(addr + 4);
+            // The 3DS doesn't have LPAE (Large Physical Access Extension), so it
+            // wouldn't do this as a single read.
```
yuriks: Are CPUs supporting LPAE required to do atomic 64-bit reads?

src/core/arm/skyeye_common/armmmu.h
yuriks: I think you shouldn't even use templates here, since they don't buy you anything, you're manually specifying the type everywhere you use them. So just make 4 functions named ReadMemory8/16/32/64.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
lioncash: Yes, with the one exception being the LDREXD/STREXD pair.

lioncash: Page 128 of the ARMARMv7 explains most of this.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -6173,8 +6209,14 @@ unsigned InterpreterMainLoop(ARMul_State* state) {
...
                 cpu->exclusive_state = 0;
 
-                Memory::Write32(write_addr, cpu->Reg[inst_cream->Rm]);
-                Memory::Write32(write_addr + 4, cpu->Reg[inst_cream->Rm + 1]);
+                u64 value;
+
+                if (InBigEndianMode(cpu))
+                    value = (((u64)cpu->Reg[inst_cream->Rm + 0] << 32) | cpu->Reg[inst_cream->Rm + 1]);
+                else
+                    value = (((u64)cpu->Reg[inst_cream->Rm + 1] << 32) | cpu->Reg[inst_cream->Rm + 0]);
```
yuriks: You could make the test a ternary deciding between `is_be ? 0 : 1` in the array offsets to reduce the duplication. It's debatable if that helps readability, but these are quite heavy statements so I think it would be worth it. (Or at least factor the register reads into separate variables instead.)

yuriks: It looks like you already do the latter in [several places](https://github.com/citra-emu/citra/pull/659/files#diff-24745854bccf9f1bb2386290e37b846fR1395) so should probably go with that.

lioncash: Updated it with the ARM notation for them.

yuriks:
LGTM

bunnei:
LGTM

PR #661, opened 2015-03-16
Fix two minor understandability issues in common

linkmauve:
This also adds back the ability to change the name of the default main directory name at compile-time on Windows.


bunnei:
LGTM

PR #662, opened 2015-03-16
Add static_cast around expressions where the compiler doesn’t deduce the right type

linkmauve:
This removes a few annoying warnings in VideoCore.


bunnei:
LGTM

PR #666, opened 2015-03-19
Pica/Rasterizer: Add SecondaryFragmentColor

MoochMcGee:
Verified with the Dragon Quest 7 warning screen.

dq7_screen_1

### DIFF
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -384,10 +384,13 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
 
                 auto GetSource = [&](Source source) -> Math::Vec4<u8> {
                     switch (source) {
-                    // TODO: What's the difference between these two?
+                    // TODO: Implement fragment lighting. The following is a bunch of hacks.
                     case Source::PrimaryColor:
                     case Source::PrimaryFragmentColor:
                         return primary_color;
+                    
+                    case Source::SecondaryFragmentColor:
+                        return 0;
 
                     case Source::Texture0:
                         return texture_color[0];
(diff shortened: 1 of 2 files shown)


src/video_core/rasterizer.cpp
```diff
@@ -384,10 +384,13 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
 
                 auto GetSource = [&](Source source) -> Math::Vec4<u8> {
                     switch (source) {
-                    // TODO: What's the difference between these two?
+                    // TODO: Implement fragment lighting. The following is a bunch of hacks.
```
bunnei: Not the best TODO, but meh, I think we all understand the issue here

bunnei:
LGTM

src/video_core/rasterizer.cpp
```diff
@@ -384,10 +384,13 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
                     case Source::PrimaryColor:
                     case Source::PrimaryFragmentColor:
                         return primary_color;
+                    
+                    case Source::SecondaryFragmentColor:
+                        return 0;
```
Normmatt: shouldn't this return {} like in the default case?

neobrain: Indeed. This probably does the right thing, but tbh I'm not sure how it even compiles at all.

bunnei:
@MoochMcGee please address the above comments

neobrain:
@MoochMcGee Bump?

MoochMcGee:
This PR is dead. Not sure if I have the code anymore.

neobrain:
@MoochMcGee Please take more responsibility about this in the future instead of making us chase after you. Thanks for your contribution regardless.

PR #668, opened 2015-03-20
PICA: Minor updates to the pica registers

purpasmart96:


purpasmart96:
got fix this

src/video_core/clipper.cpp
```diff
@@ -62,8 +62,8 @@ static void InitScreenCoordinates(OutputVertex& vtx)
     viewport.halfsize_y = float24::FromRawFloat24(registers.viewport_size_y);
     viewport.offset_x   = float24::FromFloat32(static_cast<float>(registers.viewport_corner.x));
     viewport.offset_y   = float24::FromFloat32(static_cast<float>(registers.viewport_corner.y));
-    viewport.zscale     = float24::FromRawFloat24(registers.viewport_depth_range);
-    viewport.offset_z   = float24::FromRawFloat24(registers.viewport_depth_far_plane);
+    viewport.zscale     = float24::FromRawFloat24(registers.viewport_depth_scale);
```
neobrain: What's the reason for renaming these?

neobrain:
Sorry, but -1 on this whole change. Blindly adding such definitions without actually implementing the corresponding functionality is a bad idea, for various reasons:

* It's easy to mess it up when it's not tested, and hard to debug when someone implements the functionality and doesn't double-check that the register offsets are correct.
* It makes it impossible to see which functionality is not yet implemented without explicitly searching through the code base.

Also, in this particular patch:
* Some of the renaming/rephrasing changes seem odd (no rationale was given for them, so I don't know why they are here to begin with).
* Some fields are added as full u32 values, even though they should be BitFields.

PR #669, opened 2015-03-20
armmmu: Remove unnecessary enum values

lioncash:
We don't need to care about XScale or Intel specific ARM stuff. Especially since all of the other stuff related to these two was removed.


bunnei:
LGTM

PR #672, opened 2015-03-23
ConfigMem: Set the app memory to be 96MB instead of the default 64MB

purpasmart96:
Since we aren't emulating the actual firmware, games that would use more then 64MB would crash. Because on the real 3ds, the firmware would change its Config Memory to comply with the game's memory resources.
Pokemon Gates to Affinity no longer crashes at boot, and you can almost get in game.

### DIFF
--- a/src/core/hle/config_mem.cpp
+++ b/src/core/hle/config_mem.cpp
@@ -65,9 +65,9 @@ void Init() {
     config_mem.sys_core_ver = 0x2;
     config_mem.unit_info = 0x1; // Bit 0 set for Retail
     config_mem.prev_firm = 0;
-    config_mem.app_mem_type = 0; // Defualt app mem type
+    config_mem.app_mem_type = 0x2; // Default app mem type is 0
     config_mem.unit_info = 0x1; // Bit 0 set for Retail
-    config_mem.app_mem_alloc = 0x04000000; // Default app memory size is 64MB
+    config_mem.app_mem_alloc = 0x06000000; // Set to 96MB, since some games use more than the default (64MB)
     config_mem.base_mem_alloc = 0x01400000; // Default base memory is 20MB
     config_mem.sys_mem_alloc = Memory::FCRAM_SIZE - (config_mem.app_mem_alloc + config_mem.base_mem_alloc);
     config_mem.firm_unk = 0;


src/core/hle/config_mem.cpp
```diff
@@ -65,9 +65,9 @@ void Init() {
     config_mem.sys_core_ver = 0x2;
     config_mem.unit_info = 0x1; // Bit 0 set for Retail
     config_mem.prev_firm = 0;
-    config_mem.app_mem_type = 0; // Defualt app mem type
+    config_mem.app_mem_type = 0x2; // Defualt app mem type is 0
```
Subv: Could you fix that typo please? "Default"

src/core/hle/config_mem.cpp
```diff
@@ -65,9 +65,9 @@ void Init() {
...
     config_mem.unit_info = 0x1; // Bit 0 set for Retail
-    config_mem.app_mem_alloc = 0x04000000; // Default app memory size is 64MB
+    config_mem.app_mem_alloc = 0x06000000; // Set to 96MB, since some games use more then the default (64MB)
```
Subv: "use more than"

Subv:
Nice catch, LGTM

src/core/hle/config_mem.cpp
```diff
@@ -65,9 +65,9 @@ void Init() {
     config_mem.sys_core_ver = 0x2;
     config_mem.unit_info = 0x1; // Bit 0 set for Retail
     config_mem.prev_firm = 0;
-    config_mem.app_mem_type = 0; // Defualt app mem type
+    config_mem.app_mem_type = 0x2; // Default app mem type is 0
```
Subv: Why are you changing the app_mem_type? what does 2 mean? what does 0 mean? what is this value used for in games?

yuriks: I believe it corresponds to the rows in this table: http://3dbrew.org/wiki/Memory_layout#FCRAM_memory-regions_layout

purpasmart96: That's, correct, for 96MB, the app memory type is value 2

emmauss:
why not 124MB, the default of the New 3ds. thus, games for only the n3DS can have some support.

purpasmart96:
@emmauss Well right now we aren't really doing that. Plus we would have to change some of our memory mappings  to support that change, which is out of scope of this PR, this could change could happen in the sometime in the future though, but for right now were doing mostly old3ds stuff.

yuriks:
LGTM

bunnei:
LGTM

PR #673, opened 2015-03-24
dyncom: Remove unused/unnecessary macros and macro constants

lioncash:
Should also fix warnings related to macro redefinitions.


bunnei:
LGTM

PR #674, opened 2015-03-24
dyncom: Implement RFE and SRS.

lioncash:
These are system instructions, but they can be easily implemented. Since the dyncom handles banked registers in the background upon mode switching, we don't need to care about handling them explicitly.


bunnei:
LGTM

PR #675, opened 2015-03-26
Windows build fixes

jroweboy:
Changes to make the cmake build on windows simpler. This sets proper flags for each of the build types that are supported in addition to adding a post build event that copies the necessary dll files to the build directory to facilitate simpler development. This PR is a continuation of https://github.com/citra-emu/citra/pull/613


yuriks:
You have a bug. See the appveyor build results.

jroweboy:
The issue is with this line in the appveyor build script https://github.com/citra-emu/citra/blob/master/appveyor.yml#L36 where it doesn't check for the folders existance before trying to build it. I can resolve this one of a few ways, add a `if not exist PATH` before that `mkdir`, or I can stop copying the platforms to the build directory in post build, or I can just remove those copy steps from appveyor altogether now that the post build step already does that.

yuriks:
If CMake already correctly copies the files, then I think removing that from the appveyor script is fine.

yuriks:
Despite Travis randomly erroring this time, the same commit had passed on it before, so I'll go ahead and LGTM this.

src/citra_qt/CMakeLists.txt
```diff
@@ -74,3 +74,39 @@ target_link_libraries(citra-qt ${OPENGL_gl_LIBRARY} ${CITRA_QT_LIBS})
...
+    # windows commandline expects the / to be \ so switch them
+    string(REPLACE "/" "\\" Qt5_DLL_DIR ${Qt5_DLL_DIR})
+    string(REPLACE "/" "\\" Qt5_PLATFORMS_DIR ${Qt5_PLATFORMS_DIR})
+    string(REPLACE "/" "\\" DLL_DEST ${DLL_DEST})
+    string(REPLACE "/" "\\" PLATFORMS ${PLATFORMS})
+    
+    # /NJH /NJS /NDL /NFL /NC /NS /NP - Silence any output
+    # cmake adds an extra check for command success which doesn't work too well with robocopy 
+    # so trick it into thinking the command was successful with the || cmd /c "exit /b 0"
+    add_custom_command(TARGET citra-qt POST_BUILD
```
Subv: If i run cmake with Debug configuration but later want to compile a Release build in Visual Studio, this will copy the wrong DLLs to the wrong folder, won't it? (Unless I re-run cmake with CONFIG = Release)

yuriks: It shouldn't. The point of generator expressions is that semantically they are evaluated during building, not during the CMake run.

jroweboy:
Should I just remove the parts dealing with post build moving the dlls to the directory and just reduce this PR to the debug config options? If thats whats holding up merging this, I have no problem with removing things to make it less troublesome.

yuriks:
Can someone else review and merge this?

archshift:
Will the release upload code still work after this change? Are you sure they will still be found in `.\build\bin\release\`?

yuriks:
On Windows? I believe so, since it was actually complaining that the files were *already* there when the AppVeyor script was still copying them.

Subv:
LGTM

PR #676, opened 2015-03-28
IR: Move The IR services to their own folder and implement "GetHandles"

purpasmart96:


src/core/hle/service/service.cpp
```diff
@@ -114,6 +112,7 @@ void Shutdown() {
     Service::PTM::PTMShutdown();
     Service::APT::APTShutdown();
     Service::CFG::CFGShutdown();
+    Service::IR::Shutdown();
```
Subv: You initialized this service last in line 74, please destroy it first to maintain the order

src/core/hle/service/ir/ir.h
```diff
@@ -0,0 +1,35 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "core/hle/kernel/kernel.h"
+#include "core/hle/service/service.h"
+
+namespace Kernel {
```
Subv: Why do you need this?

src/core/hle/service/service.cpp
purpasmart96: alright, will fix

src/core/hle/service/ir/ir.h
purpasmart96: It only includes the stuff I'm using  in the namespace, also it was  also used in https://github.com/citra-emu/citra/blob/master/src/core/hle/service/hid/hid.h#L13

src/core/hle/service/ir/ir.cpp
```diff
@@ -0,0 +1,48 @@
...
+namespace IR {
+
+static Kernel::SharedPtr<Kernel::Event> handle_event = nullptr;
+static Kernel::SharedPtr<Kernel::SharedMemory> shared_memory = nullptr;
+
+void GetHandles(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    cmd_buff[2] = 0x4000000;
```
yuriks: Any idea what this magic number is?

Subv: I'm guessing it's a translation descriptor, @purpasmart96 Could you confirm?

purpasmart96: Yes, I can confirm

yuriks: If that were the case, wouldn't you need one for the other handle too?

purpasmart96: Well its not really a handle though, its a translate-header used by the  ARM11-kernel.

yuriks: Uh, I think you misunderstood my question.

purpasmart96: I'm not sure what  exactly you're asking for.

yuriks: My question should make perfect sense if you know what a translation descriptor is and how it's used when transmitting handles via IPC. But either way, 0x40000000 doesn't fit the format of any of the documented descriptor types in 3dbrew, so I doubt that's what it actually is.

And now I noticed that these are actually lacking descriptors, which means the format for this response is probably wrong.

yuriks:
```
[14:32:07] <plutoo> 0x40000000 means two handles
[14:32:49] <plutoo> 0x60000000 means three handles
[14:32:55] <plutoo> 0x80000000 means four handles
[14:33:19] <plutoo> not sure why that's not on the wiki
[14:41:29] <plutoo> yuriks: updated http://3dbrew.org/wiki/IPC_Command_Structure
```

So it looks like it's legit.

yuriks:
LGTM

src/core/hle/service/ir/ir.h
Subv: But you're not using them in this file, there's no need to have these forward declarations here

purpasmart96: I'm using the kernel namespace in the Init() though, so I am using them.

yuriks: Not in the header though, so there's no point in putting the declarations here. And in the implementation file you're including the corresponding headers anyway.

purpasmart96: Well... it compiles fine without it so I guess I can remove it.

src/core/hle/service/ir/ir.cpp
bunnei: (outside of the scope of this PR): @yuriks didn't we plan on doing these for all the service routines? I remember discussing doing it in a more automated way

bunnei:
LGTM

PR #677, opened 2015-03-30
dyncom: Isolate CP15 register reading and writing

lioncash:
This moves CP15 read/write functionality into separate functions outside of the main dyncom interpreter. This PR also adds all of the missing registers supported by the MPCore and adds the correct checking for privileged modes before allowing access to some of those registers. I've preserved the same formatting used by the ARMv7 manual to make looking at this much easier.


src/core/arm/interpreter/armsupp.cpp
```diff
@@ -207,3 +209,429 @@ bool InBigEndianMode(ARMul_State* cpu)
...
+// are not implemented.
+u32 ReadCP15Register(ARMul_State* cpu, u32 crn, u32 opcode_1, u32 crm, u32 opcode_2)
+{
+    // Unprivileged registers
+    if (crn == 13 && opcode_1 == 0 && crm == 0)
+    {
+        if (opcode_2 == 2)
+            return cpu->CP15[CP15(CP15_THREAD_UPRW)];
+
+        if (opcode_2 == 3)
```
Subv: Please place a todo here until we implement the TLS

src/core/arm/interpreter/armsupp.cpp
```diff
@@ -207,3 +209,429 @@ bool InBigEndianMode(ARMul_State* cpu)
...
+                else if (opcode_2 == 2)
+                    cpu->CP15[CP15(CP15_INVALIDATE_DATA_CACHE_LINE_USING_INDEX)] = value;
+            }
+            else if (crm == 7 && opcode_2 == 0)
+            {
+                cpu->CP15[CP15(CP15_INVALIDATE_DATA_AND_INSTR_CACHE)] = value;
+            }
+            else if (crm == 10)
+            {
+                LOG_WARNING(Core_ARM11, "Cache operations not fully implemented.");
```
Subv: Is this duplicate of https://github.com/citra-emu/citra/pull/677/files#diff-62ac4b40bc99e91fff5efc11c6353246R455 intended?

src/core/arm/interpreter/armsupp.cpp
lioncash: Sure thing, however it should actually be `return cpu->CP15[CP15(CP15_THREAD_URO)];` since it's a user read-only register, but a system read/write register. So said address should be stored into that register when needed. This was carry-over from the original code.

src/core/arm/interpreter/armsupp.cpp
lioncash: Yes, that was intended to go in the TLB operations section. I've amended the commit.

src/core/arm/interpreter/armsupp.cpp
```diff
@@ -207,3 +209,432 @@ bool InBigEndianMode(ARMul_State* cpu)
...
+    if (crn == 13 && opcode_1 == 0 && crm == 0)
+    {
+        if (opcode_2 == 2)
+            return cpu->CP15[CP15(CP15_THREAD_UPRW)];
+
+        // TODO: Whenever TLS is implemented, this should return
+        // "cpu->CP15[CP15(CP15_THREAD_URO)];"
+        // which contains the address of the 0x200-byte TLS
+        if (opcode_2 == 3)
+            return Memory::KERNEL_MEMORY_VADDR;
```
yuriks: Can you do it correctly right now, and then simply have the initialization code set the register to `KERNEL_MEMORY_VADDR`?

src/core/arm/interpreter/armsupp.cpp
```diff
@@ -207,3 +209,432 @@ bool InBigEndianMode(ARMul_State* cpu)
...
+           cpu->CP15[CP15(CP15_DATA_SYNC_BARRIER)] = value;
+       else if (opcode_2 == 5)
+           cpu->CP15[CP15(CP15_DATA_MEMORY_BARRIER)] = value;
+           
+    }
+    else if (crn == 13 && opcode_1 == 0 && crm == 0 && opcode_2 == 2)
+    {
+        cpu->CP15[CP15(CP15_THREAD_UPRW)] = value;
+    }
+}
```
yuriks: Shouldn't you add a warning if you don't manage to write anything?

src/core/arm/interpreter/armsupp.cpp
lioncash: In a follow up PR, I was planning to set all of the register's documented reset values on initialization, so I can do it alongside that.

yuriks: Sounds reasonable.

src/core/arm/interpreter/armsupp.cpp
lioncash: Added it, good catch.

yuriks:
LGTM

bunnei:
LGTM

PR #678, opened 2015-03-30
callstack: Remove unnecessary disassembler instantiation

lioncash:
Decode is a static function. There's no need to allocate a disassembler instance.

Technically this function would also leak every time it's called because of this.


Subv:
LGTM

bunnei:
Lgtm

PR #680, opened 2015-04-03
Allow the user to set the background clear color during emulation

archshift:
The background color can be seen at the sides of the bottom screen or when the window is wider than normal.


Subv:
LGTM

Subv:
Sorry about that, misclicked

bunnei:
LGTM

PR #683, opened 2015-04-06
Thread priority and scheduler improvements

bunnei:
Adds support for priority boosting for starved threads, priority inheritance for mutexes, some additional (verified) error codes on svcCreateThread, and some additional bug fixes/improvements. I've written some hardware tests that I'll publish at a later time that verify these changes. Some outstanding things for my next PR:
* Identify any other scenarios (other than a mutex wait) that might cause priority inheritance (and implement)
* Implement kernel reschedule on other SVCs as appropriate
* RE other error codes that are used and implement

### DIFF
--- a/src/common/thread_queue_list.h
+++ b/src/common/thread_queue_list.h
@@ -40,6 +40,18 @@ struct ThreadQueueList {
         return -1;
     }
 
+    T get_first() {
+        Queue *cur = first;
+        while (cur != nullptr) {
+            if (!cur->data.empty()) {
+                return cur->data.front();
+            }
+            cur = cur->next_nonempty;
+        }
+
+        return T();
+    }
+
     T pop_first() {
         Queue *cur = first;
         while (cur != nullptr) {
@@ -77,6 +89,12 @@ struct ThreadQueueList {
     void push_back(Priority priority, const T& thread_id) {
         Queue *cur = &queues[priority];
         cur->data.push_back(thread_id);
+    }
+
+    void move(const T& thread_id, Priority old_priority, Priority new_priority) {
+        remove(old_priority, thread_id);
+        prepare(new_priority);
+        push_back(new_priority, thread_id);
     }
 
     void remove(Priority priority, const T& thread_id) {
--- a/src/core/hle/function_wrappers.h
+++ b/src/core/hle/function_wrappers.h
@@ -46,6 +46,13 @@ template<ResultCode func(u32*, u32, u32, u32, u32, u32)> void Wrap(){
     FuncReturn(retval);
 }
 
+template<ResultCode func(u32*, s32, u32, u32, u32, s32)> void Wrap() {
+    u32 param_1 = 0;
+    u32 retval = func(&param_1, PARAM(0), PARAM(1), PARAM(2), PARAM(3), PARAM(4)).raw;
+    Core::g_app_core->SetReg(1, param_1);
+    FuncReturn(retval);
+}
+
 template<ResultCode func(s32*, u32*, s32, bool, s64)> void Wrap() {
     s32 param_1 = 0;
     s32 retval = func(&param_1, (Handle*)Memory::GetPointer(PARAM(1)), (s32)PARAM(2),
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -154,7 +154,7 @@ void Shutdown() {
  */
 bool LoadExec(u32 entry_point) {
     // 0x30 is the typical main thread priority I've seen used so far
-    g_main_thread = Kernel::SetupMainThread(Kernel::DEFAULT_STACK_SIZE, entry_point, 0x30);
+    g_main_thread = Kernel::SetupMainThread(Kernel::DEFAULT_STACK_SIZE, entry_point, THREADPRIO_DEFAULT);
 
     return true;
 }
--- a/src/core/hle/kernel/mutex.cpp
+++ b/src/core/hle/kernel/mutex.cpp
@@ -56,7 +56,15 @@ SharedPtr<Mutex> Mutex::Create(bool initial_locked, std::string name) {
 }
 
 bool Mutex::ShouldWait() {
-    return lock_count > 0 && holding_thread != GetCurrentThread();;
+    auto thread = GetCurrentThread();
+    bool wait = lock_count > 0 && holding_thread != thread;
+
+    // If the holding thread of the mutex is lower priority than this thread, that thread should
+    // temporarily inherit this thread's priority
+    if (wait && thread->current_priority < holding_thread->current_priority)
+        holding_thread->BoostPriority(thread->current_priority);
+
+    return wait;
 }
 
 void Mutex::Acquire() {
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -140,6 +140,29 @@ void ArbitrateAllThreads(u32 address) {
     }
 }
 
+/// Boost low priority threads (temporarily) that have been starved
+static void PriorityBoostStarvedThreads() {
+    u64 current_ticks = CoreTiming::GetTicks();
+
+    for (auto& thread : thread_list) {
+        // TODO(bunnei): Threads that have been waiting to be scheduled for `boost_ticks` (or
+        // longer) will have their priority temporarily adjusted to 1 higher than the highest
+        // priority thread to prevent thread starvation. This general behavior has been verified
+        // on hardware. However, this is almost certainly not perfect, and the real CTR OS scheduler
+        // should probably be reversed to verify this.
+
+        const u64 boost_timeout = 2000000;  // Boost threads that have been ready for > this long
+
+        u64 delta = current_ticks - thread->last_running_ticks;
+
+        if (thread->status == THREADSTATUS_READY && delta > boost_timeout && !thread->idle) {
+            const s32 boost_priority = std::max(ready_queue.get_first()->current_priority - 1, 0);
+            ready_queue.move(thread, thread->current_priority, boost_priority);
+            thread->current_priority = boost_priority;
+        }
+    }
+}
+
 /** 
  * Switches the CPU's active thread context to that of the specified thread
  * @param new_thread The thread to switch to
@@ -151,6 +174,7 @@ static void SwitchContext(Thread* new_thread) {
 
     // Save context for previous thread
     if (previous_thread) {
+        previous_thread->last_running_ticks = CoreTiming::GetTicks();
         Core::g_app_core->SaveContext(previous_thread->context);
 
         if (previous_thread->status == THREADSTATUS_RUNNING) {
@@ -167,6 +191,9 @@ static void SwitchContext(Thread* new_thread) {
 
         ready_queue.remove(new_thread->current_priority, new_thread);
         new_thread->status = THREADSTATUS_RUNNING;
+
+        // Restores thread to its nominal priority if it has been temporarily changed
+        new_thread->current_priority = new_thread->nominal_priority;
 
         Core::g_app_core->LoadContext(new_thread->context);
     } else {
@@ -364,7 +391,8 @@ ResultVal<SharedPtr<Thread>> Thread::Create(std::string name, VAddr entry_point,
     thread->status = THREADSTATUS_DORMANT;
     thread->entry_point = entry_point;
     thread->stack_top = stack_top;
-    thread->initial_priority = thread->current_priority = priority;
+    thread->nominal_priority = thread->current_priority = priority;
+    thread->last_running_ticks = CoreTiming::GetTicks();
     thread->processor_id = processor_id;
     thread->wait_set_output = false;
     thread->wait_all = false;
@@ -400,18 +428,11 @@ static void ClampPriority(const Thread* thread, s32* priority) {
 void Thread::SetPriority(s32 priority) {
     ClampPriority(this, &priority);
 
-    if (current_priority == priority) {
-        return;
-    }
+    // If thread was ready, adjust queues
+    if (status == THREADSTATUS_READY)
+        ready_queue.move(this, current_priority, priority);
 
-    if (status == THREADSTATUS_READY) {
-        // If thread was ready, adjust queues
-        ready_queue.remove(current_priority, this);
-        ready_queue.prepare(priority);
-        ready_queue.push_back(priority, this);
-    }
-    
-    current_priority = priority;
+    nominal_priority = current_priority = priority;
 }
 
 SharedPtr<Thread> SetupIdleThread() {
@@ -440,6 +461,9 @@ SharedPtr<Thread> SetupMainThread(u32 stack_size, u32 entry_point, s32 priority)
 
 void Reschedule() {
     Thread* prev = GetCurrentThread();
+
+    PriorityBoostStarvedThreads();
+
     Thread* next = PopNextReadyThread();
     HLE::g_reschedule = false;
 
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -156,9 +156,8 @@ static void PriorityBoostStarvedThreads() {
         u64 delta = current_ticks - thread->last_running_ticks;
 
         if (thread->status == THREADSTATUS_READY && delta > boost_timeout && !thread->idle) {
-            const s32 boost_priority = std::max(ready_queue.get_first()->current_priority - 1, 0);
-            ready_queue.move(thread, thread->current_priority, boost_priority);
-            thread->current_priority = boost_priority;
+            const s32 priority = std::max(ready_queue.get_first()->current_priority - 1, 0);
+            thread->BoostPriority(priority);
         }
     }
 }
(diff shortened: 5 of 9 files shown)


Subv:
This fixes the freeze in 1001 Spikes

src/core/hle/function_wrappers.h
```diff
@@ -46,6 +46,13 @@ template<ResultCode func(u32*, u32, u32, u32, u32, u32)> void Wrap(){
     FuncReturn(retval);
 }
 
+template<ResultCode func(u32*, s32, u32, u32, u32, s32)> void Wrap(){
```
lioncash: space between ) and {

src/core/hle/service/apt/apt.cpp
```diff
@@ -312,7 +313,7 @@ void Init() {
 
     // TODO(bunnei): Check if these are created in Initialize or on APT process startup.
     notification_event = Kernel::Event::Create(RESETTYPE_ONESHOT, "APT_U:Notification");
-    pause_event = Kernel::Event::Create(RESETTYPE_ONESHOT, "APT_U:Pause");
+    start_event = Kernel::Event::Create(RESETTYPE_ONESHOT, "APT_U:Pause");
```
Subv: Change the name string if you're changing the variable name

Subv:
LGTM other than the comment i left

Subv:
Do semaphores also use priority-inheritance?

archshift:
10/10 would give :cake:

src/core/hle/kernel/kernel.cpp
```diff
@@ -154,7 +154,7 @@ void Shutdown() {
  */
 bool LoadExec(u32 entry_point) {
     // 0x30 is the typical main thread priority I've seen used so far
-    g_main_thread = Kernel::SetupMainThread(Kernel::DEFAULT_STACK_SIZE, entry_point, 0x30);
+    g_main_thread = Kernel::SetupMainThread(Kernel::DEFAULT_STACK_SIZE, entry_point, THREADPRIO_DEFAULT);
```
purpasmart96: You could just use the raw priority value from the exheader instead of using a default value, which is located here https://github.com/citra-emu/citra/blob/master/src/core/loader/ncch.h#L109

yuriks:
> [23:33:27] <bunnei> Did bravely default do this before -> http://i.imgur.com/7qOWHI9.png ?
720 no-scope insta-merge

archshift:
HAVE ALL MY CAKE: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake:

src/core/hle/kernel/kernel.cpp
Subv: I believe that's out of the scope of this PR, however it should be in @yuriks's next PR :)

src/core/hle/svc.cpp
```diff
@@ -283,8 +283,13 @@ static ResultCode ArbitrateAddress(Handle handle, u32 address, u32 type, u32 val
     if (arbiter == nullptr)
         return ERR_INVALID_HANDLE;
 
-    return arbiter->ArbitrateAddress(static_cast<Kernel::ArbitrationType>(type),
-            address, value, nanoseconds);
+    auto res = arbiter->ArbitrateAddress(static_cast<Kernel::ArbitrationType>(type),
+                                         address, value, nanoseconds);
+
+    if (res == RESULT_SUCCESS)
+        HLE::Reschedule(__func__);
```
Subv: You're now rescheduling twice when this is called, i suggest removing the Reschedule calls from inside ArbitrateAddress

src/common/thread_queue_list.h
```diff
@@ -40,6 +40,18 @@ struct ThreadQueueList {
         return -1;
     }
 
+    T get_first() {
+        Queue *cur = first;
+        while (cur != nullptr) {
+            if (!cur->data.empty()) {
+                return std::move(cur->data.front());
```
yuriks: It's wrong to std::move here since this might invalidate the T inside the structure.

src/core/hle/kernel/kernel.cpp
yuriks: Word.

src/core/hle/kernel/mutex.cpp
```diff
@@ -56,7 +56,15 @@ SharedPtr<Mutex> Mutex::Create(bool initial_locked, std::string name) {
...
 
 bool Mutex::ShouldWait() {
-    return lock_count > 0 && holding_thread != GetCurrentThread();;
+    auto thread = GetCurrentThread();
+    bool wait = lock_count > 0 && holding_thread != thread;
+
+    // If the holding thread of the mutex is lower priority than this thread, that thread should
+    // temporarily inherit this thread's priority
+    if (wait && thread->current_priority < holding_thread->current_priority)
+        holding_thread->BoostPriority(thread->current_priority);
```
yuriks: I really dislike having this code here, especially since `ShouldWait` sounds like something that should be an observer that doesn't modify anything. I thought about it for a moment and can't find any better place to stick this though, since this really doesn't seem to belong here. I'll ruminate on this a bit more.

src/core/hle/kernel/thread.cpp
```diff
@@ -140,6 +140,29 @@ void ArbitrateAllThreads(u32 address) {
     }
 }
 
+/// Boost low priority threads (temporarily) that have been starved
+static void PriorityBoostStarvedThreads() {
+    u64 current_ticks = CoreTiming::GetTicks();
+
+    for (auto& handle : thread_list) {
```
yuriks: s/handle/thread/

src/core/hle/kernel/thread.cpp
```diff
@@ -140,6 +140,29 @@ void ArbitrateAllThreads(u32 address) {
...
+        // TODO(bunnei): Threads that have been waiting to be scheduled for `boost_ticks` (or
+        // longer) will have their priority temporarily adjusted to 1 higher than the highest
+        // priority thread to prevent thread starvation. This general behavior has been verified
+        // on hardware. However, this is almost certainly not perfect, and the real CTR OS scheduler
+        // should probably be reversed to verify this.
+
+        const u64 boost_timeout = 2000000;  // Boost threads that have been ready for > this long
+
+        auto thread = handle.get();
```
yuriks: This is unnecessary, you can directly use the SharedPtr (it overloads `*` and `->`)

src/core/hle/kernel/thread.cpp
```diff
@@ -167,6 +191,7 @@ static void SwitchContext(Thread* new_thread) {
 
         ready_queue.remove(new_thread->current_priority, new_thread);
         new_thread->status = THREADSTATUS_RUNNING;
+        new_thread->current_priority = new_thread->initial_priority;
```
yuriks: Add a comment saying that this is unboosting the priority.

src/core/hle/kernel/thread.cpp
```diff
@@ -400,17 +426,18 @@ static void ClampPriority(const Thread* thread, s32* priority) {
...
-        ready_queue.remove(current_priority, this);
-        ready_queue.prepare(priority);
-        ready_queue.push_back(priority, this);
-    }
-    
+    // If thread was ready, adjust queues
+    if (status == THREADSTATUS_READY)
+        ready_queue.move(this, current_priority, priority);
+
+    initial_priority = current_priority = priority;
```
yuriks: `initial_priority` should be renamed to something more adequate like `configured_priority` or `nominal_priority`. Also add a comment on the struct explaining the roles of each field.

src/core/hle/kernel/thread.h
```diff
@@ -17,17 +17,18 @@
...
 };
 
-enum ThreadProcessorId {
-    THREADPROCESSORID_0     = 0xFFFFFFFE,   ///< Enables core appcode
-    THREADPROCESSORID_1     = 0xFFFFFFFD,   ///< Enables core syscore
-    THREADPROCESSORID_ALL   = 0xFFFFFFFC,   ///< Enables both cores
+enum ThreadProcessorId : s32 {
+    THREADPROCESSORID_0     = -2, ///< Enables core appcode
+    THREADPROCESSORID_1     = -3, ///< Enables core syscore
+    THREADPROCESSORID_ALL   = -4, ///< Enables both cores
```
yuriks: I think these are better represented as ~1, ~2 and ~(1 | 2).

src/core/hle/kernel/thread.h
```diff
@@ -17,17 +17,18 @@
...
+    THREADPROCESSORID_MAX   =  2, ///< Max thread processor ID
```
yuriks: This feels out of place here. It's not a bitmask like the other ones. Should be move out to a standaline const named `NUM_PROCESSORS` or `PROCESSORID_MAX`.

src/core/hle/kernel/thread.h
```diff
@@ -88,6 +89,12 @@ class Thread final : public WaitObject {
     void SetPriority(s32 priority);
 
     /**
+     * Boosts the thread's current priority (temporarily)
+     * @param priority The new priority
+     */
+    void BoostPriority(s32 priority);
```
yuriks: "Temporarily boosts the thread's priority until the next time it is scheduled."

src/core/hle/kernel/thread.h
```diff
@@ -138,6 +145,8 @@ class Thread final : public WaitObject {
     s32 initial_priority;
     s32 current_priority;
 
+    u64 last_running_ticks; ///< CPU tick count when thread was last running
```
yuriks: "tick count" sounds weird, I think just "CPU tick", meaning a timestamp, is fine. "tick count" sounds like an interval.

src/core/hle/svc.cpp
```diff
@@ -323,6 +328,22 @@ static ResultCode CreateThread(u32* out_handle, u32 priority, u32 entry_point, u
         name = Common::StringFromFormat("unknown-%08x", entry_point);
     }
 
+    // TODO(bunnei): Implement resource limits to correctly support the below test
+    //if (priority < THREADPRIO_HIGHEST_USERLAND) {
```
yuriks: Why is this commented out? Since this is the max permissible value, shouldn't this test work anyway, even though it might not catch all instances of this error? (That is: it will only error on invalid values, even though it might not error on all invalid values yet.)

src/core/hle/svc.cpp
```diff
@@ -331,10 +352,9 @@ static ResultCode CreateThread(u32* out_handle, u32 priority, u32 entry_point, u
         "threadpriority=0x%08X, processorid=0x%08X : created handle=0x%08X", entry_point,
         name.c_str(), arg, stack_top, priority, processor_id, *out_handle);
 
-    if (THREADPROCESSORID_1 == processor_id) {
-        LOG_WARNING(Kernel_SVC,
-            "thread designated for system CPU core (UNIMPLEMENTED) will be run with app core scheduling");
-    }
+    ASSERT_MSG(processor_id != THREADPROCESSORID_1, "Citra does not support SysCore!");
```
yuriks: Some notes about this, might not be relevant for you:
  - Application initial threads have Preferred Proc = 0, affinity mask = 01b.
  - Application user threads also have Preferred Proc = 0 afaik, but can have affinity != 01b.
  - Service initial threads have Preferred Proc = 1, but affinity 11b. That is, they can run on the AppCore.

src/core/hle/kernel/thread.h
bunnei: Perhaps it's a bad name... it's not a bitfield, however CreateThread checks that processor ID < 2, and if that check fails, it returns an error.

bunnei: So rather, it's the max "value" a thread processor ID can have (although ofc 1, 0, -1... are undefined)

src/core/hle/svc.cpp
bunnei: It's commented out because we don't know what THREADPRIO_HIGHEST_USERLAND is without svcGetResourceLimit being implemented

src/core/hle/svc.cpp
bunnei: This is really just meant to be a catch-all to find something that actually tries to use the syscore. I'm worried that just "passively" running syscore threads on the appcore would have consequences, so I'd rather know about it first

src/core/hle/kernel/thread.h
yuriks: But that's exactly what I'm talking about, not the name. The other enum items are bitmasks, this one is not. It makes no sense to me to cram it in the same enum.

yuriks: (As an aside, "PROCESSORID" is also a crappy name for the other ones since these are masks not ids...)

src/core/hle/svc.cpp
yuriks: But we can know what the maximum possible limit is an check against that.

src/core/hle/svc.cpp
yuriks: So this check is insufficient, since a thread might be set to run on either.

src/core/hle/kernel/thread.h
bunnei: It's called "THREADPROCESSORID" because this is just an enumeration of valid values we've seen passed into the "processor_id" field of CreateThread. That is the commonality between all of them, it has nothing to do with some being bitmasks.

src/core/hle/svc.cpp
bunnei: Set to run on either is OK, since we are not doing anything WRONG by running it on the appcore. Without this assert, we were running threads that may have been only supposed to run on the syscore on the appcore instead. So no, this check is not insufficient for what I am trying to do, but perhaps the comment could be better if it was not clear to you. More specifically, I just want to find games that actually try to use the syscore via processor ID.

src/core/hle/svc.cpp
bunnei: 0? The error code will be the same. I'd rather just implement it correctly with the known resource limit. The actual CreateThread does not check the "maximum possible limit", it checks the resource limit.

src/core/hle/kernel/thread.h
yuriks: So after some discussions in `#3dsdev` I/we are pretty sure the info currently on 3dbrew is wrong. Rather than a being at bitmask, the meaning of the parameter is:
 - -2: Run on default core specified in exheader. 0 for apps, 1 for system services.
 - -1: Run on all cores
 - 0: Run on core 0 (AppCore)
 - 1: Run on core 1 (SysCore)

Apart from fixing yet another Completely Wrong Thing™ this information obsoletes this discussion, since this enum just turned into a neat actual enumeration without any bitmasks involved.

src/core/hle/svc.cpp
yuriks: See comment above about the `processor_id` parameter. I think now we can simply check if it's == 0. (-2 should be converted to 0 for now, so it doesn't need explicitly checking. Leave a TODO to copy it from the exheader properly once we have that info.) I haven't encountered any game that actually uses -1, and it would be interesting to know of any that does, so I don't think it's a problem to fail the assert in that case.

src/core/hle/svc.cpp
yuriks: It seemed to me that you determined that 24 was the highest permissible priority in the default resource limit configuration. I am suggesting we test against that for now until we can read the actual limit, whereupon we can simply substitute the variable.

src/core/hle/svc.cpp
```diff
@@ -331,10 +346,10 @@ static ResultCode CreateThread(u32* out_handle, u32 priority, u32 entry_point, u
         "threadpriority=0x%08X, processorid=0x%08X : created handle=0x%08X", entry_point,
         name.c_str(), arg, stack_top, priority, processor_id, *out_handle);
 
-    if (THREADPROCESSORID_1 == processor_id) {
-        LOG_WARNING(Kernel_SVC,
-            "thread designated for system CPU core (UNIMPLEMENTED) will be run with app core scheduling");
-    }
+    // TODO(bunnei): Implement resource limits to return an error code instead of the below assert
+    ASSERT_MSG(priority >= THREADPRIO_USERLAND_MAX, "Unexpected thread priority!");
```
yuriks: Since you know the correct error code already, might as well leave a note for future implementors so that they don't need to dig it up.

yuriks:
LGTM once @Subv's comments are addressed.

src/core/hle/kernel/mutex.cpp
yuriks: Meh, there's not really a better place to put this I don't think, so this is fine.

src/core/hle/svc.cpp
```diff
@@ -323,6 +328,21 @@ static ResultCode CreateThread(u32* out_handle, u32 priority, u32 entry_point, u
...
+    // The error code should be: Description::NotAuthorized, Module::OS, Summary::WrongArgument,
+    // Level::Permanent
+    ASSERT_MSG(priority >= THREADPRIO_USERLAND_MAX, "Unexpected thread priority!");
+
+    if (priority > THREADPRIO_LOWEST) {
+        return ResultCode(ErrorDescription::OutOfRange, ErrorModule::OS,
+                          ErrorSummary::InvalidArgument, ErrorLevel::Usage);
+    }
+
+    if (processor_id > THREADPROCESSORID_MAX) {
```
yuriks: Shouldn't this be `>=`?

PR #684, opened 2015-04-06
dyncom: Suppress uninitialized variable warnings

lioncash:
The switch cases will always be hit, but this makes compilers stop complaining.


Subv:
LGTM

yuriks:
LGTM

PR #685, opened 2015-04-06
dyncom: Set the MPCore CP15 register reset values on initialization.

lioncash:


Subv:
LGTM

src/core/arm/interpreter/arminit.cpp
```diff
@@ -80,6 +139,16 @@ void ARMul_Reset(ARMul_State* state)
...
 
+    // Reset CP15
+    ResetMPCoreCP15Registers(state);
+
+    // This is separate from the CP15 register reset function, as
+    // this isn't an ARM-defined reset value; it's set by the 3DS.
+    //
+    // TODO: Whenever TLS is implemented, this should contain
+    // the address of the 0x200-byte TLS
+    state->CP15[CP15(CP15_THREAD_URO)] = Memory::KERNEL_MEMORY_VADDR;
```
yuriks: I think this should be moved to `Core::Init`.

yuriks: Hm, I thought you had already added functions to `ARMInterface` to change CP15 registers from C++. We're going to need that in the future anyway so might as well do it now imo.

lioncash:
@yuriks Updated accordingly. If there's more to do, I'll finish it later today; I have a test in approx 30 minutes.

yuriks:
LGTM. I think both sets should go through the same path, in case we add some actual code to be executed when one of the registers is set (for a mode change, cache flushing, etc.) but this should be fine for now.

bunnei:
Lgtm

PR #689, opened 2015-04-07
gpu: Fix a missing format specifier

lioncash:


Subv:
The output format fits in a byte, using a 32-bit-hex format to display it seems like a waste, I suggest using %x like the other display transfer log

neobrain:
> The output format fits in a byte, using a 32-bit-hex format to display it seems like a waste, I suggest using %x like the other display transfer log

It actually even fits in a nibble, so yeah - %x ;)

lioncash:
Amended it.

PR #690, opened 2015-04-09
Shared_memory: Initialize base address to 0x0

Zaneo:
Initialize base_address to 0x0. Instead of leaving it as uninitialized (0xcdcdcdcd) which is out of memory range. This should now identify the unmapped block, rather than reporting bad pointer.


Subv:
LGTM

yuriks:
The hex literal seems a bit pointless, but sure, LGTM.

bunnei:
LGTM

PR #691, opened 2015-04-09
Silence some -Wsign-compare warnings.

rohit-n:


src/citra_qt/debugger/profiler.cpp
```diff
@@ -26,7 +26,7 @@ static QVariant GetDataForColumn(int col, const AggregatedDuration& duration)
 static const TimingCategoryInfo* GetCategoryInfo(int id)
 {
     const auto& categories = GetProfilingManager().GetTimingCategoriesInfo();
-    if (id >= categories.size()) {
+    if (id >= (int)categories.size()) {
```
yuriks: Should cast id to unsigned instead, since that'll also make the comparison more robust.

src/video_core/rasterizer.cpp
```diff
@@ -343,7 +343,7 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         case Regs::TextureConfig::MirroredRepeat:
                         {
                             int coord = (int)((unsigned)val % (2 * size));
-                            if (coord >= size)
+                            if (coord >= (int)size)
```
yuriks: Change `coord` to unsigned int, this also removes the need for the cast on the above line.

rohit-n:
Changing coord to unsigned int causes a "inconsistent types ‘int’ and ‘unsigned int’ deduced for lambda return type" - should I change val to unsigned as well?

yuriks:
In that case just cast the return expression to int.

bunnei:
LGTM

PR #694, opened 2015-04-10
Added stubbed APT::Uknown_0x01010000, and PTM::CheckNew3Ds

Zaneo:
APT:Uknown_0x01010000 does not have a name on 3D Brew, and the description didn't seem concise enough to give it one.

Apparently there is some reliance on an "internal NS state field" which can be found in .data/.bss. The location changes for different titles.


src/core/hle/service/apt/apt_u.cpp
```diff
@@ -93,6 +93,7 @@ const Interface::FunctionInfo FunctionTable[] = {
     {0x004E0000, nullptr,                         "HardwareResetAsync"},
     {0x004F0080, SetAppCpuTimeLimit,              "SetAppCpuTimeLimit"},
     {0x00500040, GetAppCpuTimeLimit,              "GetAppCpuTimeLimit"},
+	{0x01010000, Unknown_0x01010000,			  "Unknown_0x01010000"},
```
archshift: Spaces, not tabs

src/core/hle/service/apt/apt_s.cpp
```diff
@@ -94,6 +94,7 @@ const Interface::FunctionInfo FunctionTable[] = {
     {0x004E0000, nullptr,               "HardwareResetAsync"},
     {0x004F0080, nullptr,               "SetApplicationCpuTimeLimit"},
     {0x00500040, nullptr,               "GetApplicationCpuTimeLimit"},
+	{0x01010000, Unknown_0x01010000,	"Unknown_0x01010000" },
```
archshift: We use spaces, not tabs

src/core/hle/service/ptm/ptm.cpp
```diff
@@ -38,6 +38,10 @@ ChargeLevels GetBatteryLevel() {
     return ChargeLevels::CompletelyFull; // Set to a completely full battery
 }
 
+u8 CheckNew3Ds(){
```
archshift: Should be `CheckNew3DS` <- no lowercase 's' in '3DS'

src/core/hle/service/ptm/ptm_sysm.h
```diff
@@ -9,6 +9,12 @@
 namespace Service {
 namespace PTM {
 
+/// Physical version of the 3Ds
+enum class PhysicalVersion : u8 {
+	Old3Ds = 0,
```
archshift: Again with the lowercase 's' in '3DS'

src/core/hle/service/ptm/ptm_sysm.cpp
```diff
@@ -32,7 +46,7 @@ const Interface::FunctionInfo FunctionTable[] = {
     {0x040700C0, nullptr, "ShutdownAsync"},
     {0x04080000, nullptr, "Awake"},
     {0x04090080, nullptr, "RebootAsync"},
-    {0x040A0000, nullptr, "CheckNew3DS"},
+    {0x040A0000, CheckNew3Ds, "CheckNew3DS"},
```
archshift: Please align these!

src/core/hle/service/apt/apt.h
```diff
@@ -249,6 +249,15 @@ void SetAppCpuTimeLimit(Service::Interface* self);
  */
 void GetAppCpuTimeLimit(Service::Interface* self);
 
+/**
+* From 3Ds Brew:
```
archshift: "3dbrew"

src/core/hle/service/apt/apt.cpp
```diff
@@ -279,6 +281,22 @@ void GetAppCpuTimeLimit(Service::Interface* self) {
     LOG_WARNING(Service_APT, "(STUBBED) called value=%u", value);
 }
 
+void Unknown_0x01010000(Service::Interface* self){
+	u32* cmd_buff = Kernel::GetCommandBuffer();
+
+	cmd_buff[1] = RESULT_SUCCESS.raw;
+
+	// When a certain NS state field (internal NS state in .data/.bss.) is non-zero, the output value is zero
+	if (false){
```
archshift: wat

src/core/hle/service/apt/apt.cpp
```diff
@@ -279,6 +281,22 @@ void GetAppCpuTimeLimit(Service::Interface* self) {
...
+		cmd_buff[2] = 0;
+	}
+	else{
```
archshift: Assuming you actually intended for the condition to be possible, it'd probably be better for you to use the syntax:

```cpp
cmdbuff[2] = some_condition ? 0 : PTM::CheckNew3ds();
```

src/core/hle/service/ptm/ptm_sysm.cpp
```diff
@@ -22,6 +23,19 @@ void IsLegacyPowerOff(Service::Interface* self) {
     cmd_buff[2] = 0;
 }
 
+/**
+* Checks the physical version of the 3Ds
+*  Outputs:
+*      0 : Return header
+*      1 : 0 if old 3Ds, 1 if new 3DS
```
archshift: Again with uniform casing of "3ds"

src/core/hle/service/apt/apt.cpp
Zaneo: The location of the value that needs to be checked for the conditional is not known. It was left like this to convey that ideally a check should be made to this "internal NS state field" which is apparently title specific.

I can switch it to a ternary operator, with the false condition still there. Or I can directly set it to PTM::CheckNew3ds, and leave a comment that it should be doing a check. Is there a preference?

src/core/hle/service/apt/apt.cpp
Zaneo: From the other functions in apt.cpp, It looked like this is how you wrote a response, is this incorrect? Or is it the name, because there currently isn't a name on 3dbrew for this function. The function name contains the command header to identify it.

MoochMcGee: if(false) never executes. Do you not know basic C?

lioncash:
Can you put spaces after your braces; it just maintains consistency.

e.g. changing:

```cpp
if (cond){
    <code here>
}

void SomeFunction(){
    <code here>
}
```

to

```cpp
if (cond) {
    <code here>
}

void SomeFunction() {
    <code here>
}
```

Zaneo:
Yeah that's not a problem. Ive previous missed some things in your formatting document, so I'll review that as well.

src/core/hle/service/apt/apt.cpp
yuriks: You totally missed the point of what Yellows8 said. This internal state from NS is an implementation detail of the service. It is not some game value that needs to be checked or anything like that. This simply means there is some condition that needs to be checked, likely some functionality to put the system into an Old3DS compatibility mode.

yuriks:
As an aside, is this necessary for any games? If not, then there's no point in doing this.

src/core/hle/service/apt/apt.cpp
lioncash: Anything in the `if (false)` branch will never be executed.

Zaneo:
I can't attest to if it is needed, but I know that 3dscraft calls it.

PR #695, opened 2015-04-11
GPU: Implemented default vertex shader attributes.

Subv:
Fixes the crash in some games like Picross E and Cubic Ninja.

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -27,6 +27,10 @@ static int float_regs_counter = 0;
 
 static u32 uniform_write_buffer[4];
 
+static int default_attr_counter = 0;
+
+static u32 default_attr_write_buffer[3];
+
 Common::Profiling::TimingCategory category_drawing("Drawing");
 
 static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
@@ -71,12 +75,9 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
             u32 vertex_attribute_sources[16];
             boost::fill(vertex_attribute_sources, 0xdeadbeef);
             u32 vertex_attribute_strides[16];
-            u32 vertex_attribute_formats[16];
+            Regs::VertexAttributeFormat vertex_attribute_formats[16];
 
-            // HACK: Initialize vertex_attribute_elements to zero to prevent infinite loops below.
-            // This is one of the hacks required to deal with uninitalized vertex attributes.
-            // TODO: Fix this properly.
-            u32 vertex_attribute_elements[16] = {};
+            u32 vertex_attribute_elements[16];
             u32 vertex_attribute_element_size[16];
 
             // Setup attribute data from loaders
@@ -90,7 +91,7 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
                     u32 attribute_index = loader_config.GetComponent(component);
                     vertex_attribute_sources[attribute_index] = load_address;
                     vertex_attribute_strides[attribute_index] = static_cast<u32>(loader_config.byte_count);
-                    vertex_attribute_formats[attribute_index] = static_cast<u32>(attribute_config.GetFormat(attribute_index));
+                    vertex_attribute_formats[attribute_index] = attribute_config.GetFormat(attribute_index);
                     vertex_attribute_elements[attribute_index] = attribute_config.GetNumElements(attribute_index);
                     vertex_attribute_element_size[attribute_index] = attribute_config.GetElementSizeInBytes(attribute_index);
                     load_address += attribute_config.GetStride(attribute_index);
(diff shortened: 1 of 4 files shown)


src/video_core/command_processor.cpp
```diff
@@ -259,6 +266,47 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
             }
             break;
         }
+        
+        // Load default VS attributes
```
neobrain: *vertex attributes

neobrain: *vertex input attributes

src/video_core/pica.h
```diff
@@ -625,7 +629,19 @@ struct Regs {
...
-    INSERT_PADDING_WORDS(0x2e);
+    INSERT_PADDING_WORDS(0x2);
+
+    // These registers are used to setup the default "fall-back" vertex shader attributes
+    struct {
+        // Index of the current default attribute
+        u32 index;
+        
+        // Writing to these registers sets the "current" default attribute.
+        // TODO: It's not clear how the hardware stores what the "current" default attribute is.
```
neobrain: It is. That's what the index variable is for :p

src/video_core/vertex_shader.cpp
```diff
@@ -35,6 +35,8 @@ static struct {
     std::array<Math::Vec4<u8>,4> i;
 } shader_uniforms;
 
+Math::Vec4<float24> vs_default_attributes[16];
```
neobrain: static

src/video_core/command_processor.cpp
```diff
@@ -126,26 +130,29 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                                  i, vertex, index,
+                                  input.attr[i][0].ToFloat32(), input.attr[i][1].ToFloat32(),
+                                  input.attr[i][2].ToFloat32(), input.attr[i][3].ToFloat32());
+                    } else {
+                        for (unsigned int comp = 0; comp < vertex_attribute_elements[i]; ++comp) {
+                            const u8* srcdata = Memory::GetPointer(PAddrToVAddr(vertex_attribute_sources[i] + vertex_attribute_strides[i] * vertex + comp * vertex_attribute_element_size[i]));
+
+                            const float srcval = (vertex_attribute_formats[i] == 0) ? *(s8*)srcdata :
+                                (vertex_attribute_formats[i] == 1) ? *(u8*)srcdata :
+                                (vertex_attribute_formats[i] == 2) ? *(s16*)srcdata :
```
bunnei: Could make the format an enum, esp if it's used elsewhere

neobrain: There actually is already an enum for this: https://github.com/citra-emu/citra/blob/master/src/video_core/pica.h#L493 . Not sure why I didn't use it.

src/video_core/command_processor.cpp
```diff
@@ -259,6 +266,47 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+
+                if (setup.index >= 16) {
+                    LOG_ERROR(HW_GPU, "Invalid VS default attribute index %d", (int)setup.index);
+                    break;
+                }
+
+                Math::Vec4<float24>& attribute = VertexShader::GetDefaultAttribute(setup.index);
+                
+                // NOTE: The destination component order indeed is "backwards"
+                // TODO: Untested
```
bunnei: How can we test this?

neobrain: Copy-pasta leftover.

Really, as soon as there is one application which uses different values for the x,y,z, and w components (which obviously is the case for nontrivial users of the default attribute feature), the component order has been properly "tested".

src/video_core/pica.h
```diff
@@ -489,14 +489,14 @@ struct Regs {
...
     INSERT_PADDING_WORDS(0xe0);
 
-    struct {
-        enum class Format : u64 {
-            BYTE = 0,
-            UBYTE = 1,
-            SHORT = 2,
-            FLOAT = 3,
-        };
+	enum class AttributeFormat : u64 {
```
neobrain: *VertexAttributeFormat

(that's what you get for moving it out of the structure! :p)

Subv:
Updated

bunnei:
LGTM

src/video_core/command_processor.cpp
yuriks: Hm, fix this comment then?

Subv:
Squashed

yuriks:
I'm still getting crashes in Bravely Default, OoT and Sub Wars with this branch.

src/video_core/pica.h
```diff
@@ -489,14 +489,14 @@ struct Regs {
...
-            UBYTE = 1,
-            SHORT = 2,
-            FLOAT = 3,
-        };
+	enum class VertexAttributeFormat : u64 {
+		BYTE = 0,
+		UBYTE = 1,
+		SHORT = 2,
+		FLOAT = 3,
+	};
```
yuriks: Spaces instead of tabs.

src/video_core/pica.h
```diff
@@ -555,14 +555,18 @@ struct Regs {
...
+            return (GetFormat(n) == VertexAttributeFormat::FLOAT) ? 4 :
+                (GetFormat(n) == VertexAttributeFormat::SHORT) ? 2 : 1;
         }
 
         inline int GetStride(int n) const {
             return GetNumElements(n) * GetElementSizeInBytes(n);
         }
 
+        inline bool IsDefaultAttribute(int id) const {
+            return (id >= 12) || (attribute_mask & (1 << id)) == 1;
```
yuriks: This condition is incorrect and will always return false. Remove the `== 1`.

yuriks:
The OoT logo also seems to be completely dark:
OoT logo

As far as I know, this wasn't supposed to be happening.

src/video_core/vertex_shader.cpp
```diff
@@ -568,22 +574,11 @@ OutputVertex RunShader(const InputVertex& input, int num_attributes) {
...
-    if(num_attributes > 8) state.input_register_table[attribute_register_map.attribute8_register] = &input.attr[8].x;
-    if(num_attributes > 9) state.input_register_table[attribute_register_map.attribute9_register] = &input.attr[9].x;
-    if(num_attributes > 10) state.input_register_table[attribute_register_map.attribute10_register] = &input.attr[10].x;
-    if(num_attributes > 11) state.input_register_table[attribute_register_map.attribute11_register] = &input.attr[11].x;
-    if(num_attributes > 12) state.input_register_table[attribute_register_map.attribute12_register] = &input.attr[12].x;
-    if(num_attributes > 13) state.input_register_table[attribute_register_map.attribute13_register] = &input.attr[13].x;
-    if(num_attributes > 14) state.input_register_table[attribute_register_map.attribute14_register] = &input.attr[14].x;
-    if(num_attributes > 15) state.input_register_table[attribute_register_map.attribute15_register] = &input.attr[15].x;
+    
+    for (int attribute = 0; attribute < num_attributes; ++attribute) {
```
neobrain: Please don't submit unrelated changes in a PR just because they affect the same code. There's `git add -p` to deal with such scenarios.

neobrain:
By the way, please choose more descriptive branch names in the future, as they do show up in the git logs.

src/video_core/pica.h
```diff
@@ -555,14 +555,18 @@ struct Regs {
...
+        }
+
```
neobrain: Trailing whitespace.

I shouldn't have to point this out this often, especially since I miss out on most of these and they're annoying regardless. Can you /please/ setup your editor to highlight them and/or setup your own linter locally?

Subv:
We are currently investigating a platform-specific bug in master that manifests itself only on Windows, but not on Linux or OS X. This PR seems to make it show up on all platforms.

The bug makes Steel Diver (and possibly other games?) render incorrectly.

bunnei:
What is the status of this PR? Are there plans to fix the bug before merging, and if so, has any progress been made or could you use any additional help? Otherwise, should we just create an issue for it and move forward?

bunnei:
I've looked into the bug with Steel Diver (cf #714) and determined that it's unrelated to this change. Please squash and we can merge so long as @neobrain is OK with it.

neobrain:
Yeah, this LGTM. Thanks for looking into the other bug so extensively!

PR #697, opened 2015-04-14
Fix a few warnings as reported by clang 3.6

linkmauve:
This was just at its default error reporting level.

The remaining warnings are mostly due to UNREACHABLE() not telling it really is unreachable to the compiler.


yuriks:
LGTM. `override` is always good

lioncash:
Looks good to me.

bunnei:
Lgtm

PR #698, opened 2015-04-16
Clip mouse input touchpad area

Zaneo:
Instead of ending the TouchMove when the mouse is dragged outside the touchpad area, clip the coordinates and allow the TouchMove to continue.

### DIFF
--- a/src/citra/emu_window/emu_window_glfw.cpp
+++ b/src/citra/emu_window/emu_window_glfw.cpp
@@ -31,7 +31,7 @@ void EmuWindow_GLFW::OnMouseButtonEvent(GLFWwindow* win, int button, int action,
 }
 
 void EmuWindow_GLFW::OnCursorPosEvent(GLFWwindow* win, double x, double y) {
-    GetEmuWindow(win)->TouchMoved(static_cast<unsigned>(x), static_cast<unsigned>(y));
+    GetEmuWindow(win)->TouchMoved(static_cast<unsigned>(std::max(x, 0.0)), static_cast<unsigned>(std::max(y, 0.0)));
 }
 
 /// Called by GLFW when a key event occurs
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -288,7 +288,7 @@ void GRenderWindow::mousePressEvent(QMouseEvent *event)
 void GRenderWindow::mouseMoveEvent(QMouseEvent *event)
 {
     auto pos = event->pos();
-    this->TouchMoved(static_cast<unsigned>(pos.x()), static_cast<unsigned>(pos.y()));
+    this->TouchMoved(static_cast<unsigned>(std::max(pos.x(), 0)), static_cast<unsigned>(std::max(pos.y(), 0)));
 }
 
 void GRenderWindow::mouseReleaseEvent(QMouseEvent *event)
--- a/src/common/emu_window.cpp
+++ b/src/common/emu_window.cpp
@@ -28,6 +28,17 @@ static bool IsWithinTouchscreen(const EmuWindow::FramebufferLayout& layout, unsi
             framebuffer_x <  layout.bottom_screen.right);
 }
 
+std::tuple<unsigned,unsigned> EmuWindow::ClipToTouchScreen(unsigned new_x, unsigned new_y) {
+
+    new_x = std::max(new_x, framebuffer_layout.bottom_screen.left);
+    new_x = std::min(new_x, framebuffer_layout.bottom_screen.right-1);
+    
+    new_y = std::max(new_y, framebuffer_layout.bottom_screen.top);
+    new_y = std::min(new_y, framebuffer_layout.bottom_screen.bottom-1);
+
+    return std::make_tuple(new_x, new_y);
+}
+
 void EmuWindow::TouchPressed(unsigned framebuffer_x, unsigned framebuffer_y) {
     if (!IsWithinTouchscreen(framebuffer_layout, framebuffer_x, framebuffer_y))
         return;
@@ -52,14 +63,13 @@ void EmuWindow::TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y) {
     if (!touch_pressed)
         return;
 
-    if (IsWithinTouchscreen(framebuffer_layout, framebuffer_x, framebuffer_y))
-        TouchPressed(framebuffer_x, framebuffer_y);
-    else
-        TouchReleased();
+    if (!IsWithinTouchscreen(framebuffer_layout, framebuffer_x, framebuffer_y))
+        std::tie(framebuffer_x, framebuffer_y) = ClipToTouchScreen(framebuffer_x, framebuffer_y);
+
+    TouchPressed(framebuffer_x, framebuffer_y);
 }
 
-EmuWindow::FramebufferLayout EmuWindow::FramebufferLayout::DefaultScreenLayout(unsigned width,
-    unsigned height) {
+EmuWindow::FramebufferLayout EmuWindow::FramebufferLayout::DefaultScreenLayout(unsigned width, unsigned height) {
 
     ASSERT(width > 0);
     ASSERT(height > 0);
--- a/src/common/emu_window.h
+++ b/src/common/emu_window.h
@@ -206,5 +206,10 @@ private:
     u16 touch_x;    ///< Touchpad X-position in native 3DS pixel coordinates (0-320)
     u16 touch_y;    ///< Touchpad Y-position in native 3DS pixel coordinates (0-240)
 
+   /**
+    * Clip the provided coordinates to be inside the touchscreen area.
+    */
+    std::tuple<unsigned,unsigned> ClipToTouchScreen(unsigned new_x, unsigned new_y);
+
     Service::HID::PadState pad_state;
 };


src/citra_qt/bootmanager.cpp
```diff
@@ -288,7 +288,7 @@ void GRenderWindow::mousePressEvent(QMouseEvent *event)
 void GRenderWindow::mouseMoveEvent(QMouseEvent *event)
 {
     auto pos = event->pos();
-    this->TouchMoved(static_cast<unsigned>(pos.x()), static_cast<unsigned>(pos.y()));
+    this->TouchMoved(static_cast<signed>(pos.x()), static_cast<signed>(pos.y()));
```
neobrain: Please use int instead of signed

src/common/emu_window.cpp
```diff
@@ -20,12 +20,26 @@ void EmuWindow::KeyReleased(KeyMap::HostDeviceKey key) {
...
-            framebuffer_x <  layout.bottom_screen.right);
+static bool IsWithinTouchscreen(const EmuWindow::FramebufferLayout& layout, signed framebuffer_x,
+                                signed framebuffer_y) {
+    return (framebuffer_y >= (signed)layout.bottom_screen.top    &&
+            framebuffer_y <  (signed)layout.bottom_screen.bottom &&
+            framebuffer_x >= (signed)layout.bottom_screen.left   &&
+            framebuffer_x <  (signed)layout.bottom_screen.right);
+}
+
+void EmuWindow::ClipToTouchScreen(signed * const new_x, signed * const new_y) {
```
neobrain: Please follow coding style more thoroughly. To stay in line with existing code, return a tuple to the clipped coordinates here instead of taking reference parameters.

src/common/emu_window.cpp
```diff
@@ -20,12 +20,26 @@ void EmuWindow::KeyReleased(KeyMap::HostDeviceKey key) {
...
+        *new_y = (signed)framebuffer_layout.bottom_screen.top;
+    else if (*new_y > (signed)framebuffer_layout.bottom_screen.bottom)
+        *new_y = (signed)framebuffer_layout.bottom_screen.bottom-1;
+
+    if (*new_x < (signed)framebuffer_layout.bottom_screen.left)
+        *new_x = (signed)framebuffer_layout.bottom_screen.left;
+    else if (*new_x >(signed)framebuffer_layout.bottom_screen.right)
+        *new_x = (signed)framebuffer_layout.bottom_screen.right - 1;
+
+    return;
```
neobrain: no need for explicit return

src/common/emu_window.h
```diff
@@ -54,6 +54,7 @@ class EmuWindow
         unsigned height;
         MathUtil::Rectangle<unsigned> top_screen;
         MathUtil::Rectangle<unsigned> bottom_screen;
+
```
neobrain: redundant space

Zaneo:
They are changed to int, because GLFW will return a negative coordinate position if the mouse is to the left, or above the window origin. If left as unsigned, dragging off screen would be a unsigned overflow. Previously this wasn't an issue since the touch was released anyways (x = 4 billion) is still outside the window.

src/common/emu_window.h
Zaneo: Sorry about that I missed it when I was cleaning test code.

src/common/emu_window.cpp
```diff
@@ -20,12 +20,26 @@ void EmuWindow::KeyReleased(KeyMap::HostDeviceKey key) {
  * @param framebuffer_y Framebuffer y-coordinate to check
  * @return True if the coordinates are within the touchpad, otherwise false
  */
-static bool IsWithinTouchscreen(const EmuWindow::FramebufferLayout& layout, unsigned framebuffer_x,
-                                unsigned framebuffer_y) {
-    return (framebuffer_y >= layout.bottom_screen.top    &&
-            framebuffer_y <  layout.bottom_screen.bottom &&
-            framebuffer_x >= layout.bottom_screen.left   &&
-            framebuffer_x <  layout.bottom_screen.right);
+static bool IsWithinTouchscreen(const EmuWindow::FramebufferLayout& layout, signed framebuffer_x,
```
lioncash: You can change `signed` to `int`

src/citra/emu_window/emu_window_glfw.cpp
```diff
@@ -31,7 +31,13 @@ void EmuWindow_GLFW::OnMouseButtonEvent(GLFWwindow* win, int button, int action,
...
 
 void EmuWindow_GLFW::OnCursorPosEvent(GLFWwindow* win, double x, double y) {
-    GetEmuWindow(win)->TouchMoved(static_cast<unsigned>(x), static_cast<unsigned>(y));
+
+    /* From GLFW documentations:
+    * Casting directly to an integer type works for positive coordinates, but fails for negative ones.
+    *
+    * The testing I did for this seems to indicate that you can directly cast without issue (Zaneo)
+    */
+    GetEmuWindow(win)->TouchMoved(x > 0 ? static_cast<signed>(x) : -static_cast<signed>(-x), y > 0 ? static_cast<signed>(y) : -static_cast<signed>(-y));
```
lioncash: Use single-line comments for this:

```cpp
// From GLFW documentations:
//  Casting directly to an integer type works for positive coordinates, but fails for negative ones.
//
// The testing I did for this seems to indicate that you can directly cast without issue (Zaneo)
```

src/common/emu_window.h
```diff
@@ -206,5 +207,10 @@ class EmuWindow
     u16 touch_x;    ///< Touchpad X-position in native 3DS pixel coordinates (0-320)
     u16 touch_y;    ///< Touchpad Y-position in native 3DS pixel coordinates (0-240)
 
+    /**
+    * Clip the provided coordinates to be inside the touchscreen area.
+    */
+    void ClipToTouchScreen(signed * new_x, signed *new_y);
```
lioncash: Situate the asterisks so that they vertically align equally with one another.
```cpp
/**
 * Clip the provided coordinates to be inside the touchscreen area.
 */
void ClipToTouchScreen(signed * new_x, signed *new_y);
```

src/citra/emu_window/emu_window_glfw.cpp
```diff
@@ -31,7 +31,13 @@ void EmuWindow_GLFW::OnMouseButtonEvent(GLFWwindow* win, int button, int action,
 }
 
 void EmuWindow_GLFW::OnCursorPosEvent(GLFWwindow* win, double x, double y) {
-    GetEmuWindow(win)->TouchMoved(static_cast<unsigned>(x), static_cast<unsigned>(y));
+
+    /* From GLFW documentations:
+    * Casting directly to an integer type works for positive coordinates, but fails for negative ones.
```
neobrain: Why?

Zaneo: You are correct, it looks like I read the comment for, glfwGetCursorPos(double, double).

src/common/emu_window.cpp
Zaneo: @neobrain Orignally I opted to make them int, because I was taking the slope into consideration. I suppose now that it i just clipped to the region they could be made unsigned, and handled in the GLFW code.

I don't know how the QT code interacts, I haven't built that. So I don't know if it also returns negatives.

src/common/emu_window.cpp
```diff
@@ -52,14 +66,17 @@ void EmuWindow::TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y) {
...
-    else
-        TouchReleased();
+    if (!IsWithinTouchscreen(framebuffer_layout, framebuffer_x, framebuffer_y)) {
+        auto clippedTuple = ClipToTouchScreen(framebuffer_x, framebuffer_y);
+
+        framebuffer_x = std::get<0>(clippedTuple);
+        framebuffer_y = std::get<1>(clippedTuple);
+    }
+
+    TouchPressed(static_cast<unsigned>(framebuffer_x), static_cast<unsigned>(framebuffer_y));
```
neobrain: Redundant casts

src/citra/emu_window/emu_window_glfw.cpp
```diff
@@ -31,7 +31,7 @@ void EmuWindow_GLFW::OnMouseButtonEvent(GLFWwindow* win, int button, int action,
 }
 
 void EmuWindow_GLFW::OnCursorPosEvent(GLFWwindow* win, double x, double y) {
-    GetEmuWindow(win)->TouchMoved(static_cast<unsigned>(x), static_cast<unsigned>(y));
```
neobrain: You can just use `static_cast<unsigned>(std::max(x, 0.0))` here. I don't think the explicit cast is necessary here, but chances are it fixes some compiler warnings.

neobrain:
@Zaneo Could you double-check if Qt requires a similar fix?

Zaneo:
Yes it looks like it affects Qt5 as well, and the same fix works.

Zaneo:
There are a number of commits that aren't relevant any more, is there a way I can squish them?

purpasmart96:
@Zaneo yeah

neobrain:
Could you clarify your commit message a bit? The "these are clipped to zero" sounded out-of-place to me until I read it 4 times. Also while you're at it, the usual format we use for the summary line is "Module: <Description>", with the description using active forms of a verb, i.e. "EmuWindow: Clip mouse input coordinates to emulated screen dimensions."

neobrain:
Other than that, LGTM

PR #699, opened 2015-04-16
Common: thread.h cleanups

yuriks:
The helper classes are rendered obsolete by C++11 lambdas. Also made
formatting conform to our code style.

Note: Because these aren't used anywhere in the codebase I couldn't really test them.

### DIFF
--- a/src/common/thread.h
+++ b/src/common/thread.h
@@ -51,109 +51,60 @@ int CurrentThreadId();
 void SetThreadAffinity(std::thread::native_handle_type thread, u32 mask);
 void SetCurrentThreadAffinity(u32 mask);
 
-class Event
-{
+class Event {
 public:
-    Event()
-        : is_set(false)
-    {}
+    Event() : is_set(false) {}
 
-    void Set()
-    {
+    void Set() {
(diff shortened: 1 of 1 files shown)


src/common/thread.h
```diff
@@ -51,109 +51,60 @@ int CurrentThreadId();
...
-        const Event* const m_event;
-    };
-
-    volatile bool is_set;
+    bool is_set;
     std::condition_variable m_condvar;
     std::mutex m_mutex;
 };
 
-// TODO: doesn't work on windows with (count > 2)
```
lioncash: I'm assuming this isn't a concern anymore?

lioncash:
Looks fine to me.

src/common/thread.h
yuriks: Examining it, I couldn't find any reason for this to be the case, so I assume it's a leftover comment from the old implementation. (This was re-written at some point to use the c++11 threading primitives instead of the custom stuff we had on common.)

lioncash: Ah OK, fair enough.

bunnei:
LGTM

PR #702, opened 2015-04-17
Add CiTrace recording support

neobrain:
Adds support for recording CiTraces within Citra (tl;dr: record all GPU commands sent to a game, so that https://github.com/citra/citrace-player can replay them arbitrarily often). The functionality is exposed in a new debugging widget called "CiTrace Recorder".

NOTE: CiTracing needs to be started and halted at GPU breakpoints. The GUI mostly takes care of not allowing the user to start/stop at any other time anyway, but there are still some rough edges.

### DIFF
--- a/src/citra_qt/debugger/graphics_tracing.cpp
+++ b/src/citra_qt/debugger/graphics_tracing.cpp
@@ -12,6 +12,8 @@
 #include <QPushButton>
 #include <QSpinBox>
 
+#include <boost/range/algorithm/copy.hpp>
+
 #include "core/hw/gpu.h"
 #include "core/hw/lcd.h"
 
@@ -76,15 +78,19 @@ void GraphicsTracingWidget::StartRecording() {
         for (unsigned comp = 0; comp < 3; ++comp)
             vs_float_uniforms[4 * i + comp] = nihstro::to_float24(Pica::g_state.vs.uniforms.f[i][comp].ToFloat32());
 
-    auto recorder = new CiTrace::Recorder((u32*)&GPU::g_regs, sizeof(GPU::g_regs) / sizeof(u32),
-                                          (u32*)&LCD::g_regs, sizeof(LCD::g_regs) / sizeof(u32),
-                                          (u32*)&Pica::g_state.regs, sizeof(Pica::g_state.regs) / sizeof(u32),
-                                          default_attributes.data(), default_attributes.size(),
-                                          shader_binary.data(), shader_binary.size(),
-                                          swizzle_data.data(), swizzle_data.size(),
-                                          vs_float_uniforms.data(), vs_float_uniforms.size(),
-                                          nullptr, 0, nullptr, 0, nullptr, 0 // Geometry shader is not implemented yet, so submit dummy data for now
-                                          );
+    CiTrace::Recorder::InitialState state;
+    std::copy_n((u32*)&GPU::g_regs, sizeof(GPU::g_regs) / sizeof(u32), std::back_inserter(state.gpu_registers));
+    std::copy_n((u32*)&LCD::g_regs, sizeof(LCD::g_regs) / sizeof(u32), std::back_inserter(state.lcd_registers));
+    std::copy_n((u32*)&Pica::g_state.regs, sizeof(Pica::g_state.regs) / sizeof(u32), std::back_inserter(state.pica_registers));
+    boost::copy(default_attributes, std::back_inserter(state.default_attributes));
+    boost::copy(shader_binary, std::back_inserter(state.vs_program_binary));
+    boost::copy(swizzle_data, std::back_inserter(state.vs_swizzle_data));
+    boost::copy(vs_float_uniforms, std::back_inserter(state.vs_float_uniforms));
+    //boost::copy(TODO: Not implemented, std::back_inserter(state.gs_program_binary));
+    //boost::copy(TODO: Not implemented, std::back_inserter(state.gs_swizzle_data));
+    //boost::copy(TODO: Not implemented, std::back_inserter(state.gs_float_uniforms));
+
+    auto recorder = new CiTrace::Recorder(state);
     context->recorder = std::shared_ptr<CiTrace::Recorder>(recorder);
 
     emit SetStartTracingButtonEnabled(false);
--- a/src/citra_qt/debugger/graphics_tracing.cpp
+++ b/src/citra_qt/debugger/graphics_tracing.cpp
@@ -8,6 +8,7 @@
 #include <QComboBox>
 #include <QFileDialog>
 #include <QLabel>
+#include <QMessageBox>
 #include <QPushButton>
 #include <QSpinBox>
 
@@ -49,8 +50,6 @@ GraphicsTracingWidget::GraphicsTracingWidget(std::shared_ptr<Pica::DebugContext>
     }
     main_widget->setLayout(main_layout);
     setWidget(main_widget);
-
-    // TODO: Make sure to have this widget disabled as soon as emulation is started!
 }
 
 void GraphicsTracingWidget::StartRecording() {
@@ -120,4 +119,34 @@ void GraphicsTracingWidget::OnBreakPointHit(Pica::DebugContext::Event event, voi
 
 void GraphicsTracingWidget::OnResumed() {
     widget()->setEnabled(false);
+}
+
+void GraphicsTracingWidget::OnEmulationStarting(EmuThread* emu_thread) {
+    // Disable tracing starting/stopping until a GPU breakpoint is reached
+    widget()->setEnabled(false);
+}
+
+void GraphicsTracingWidget::OnEmulationStopping() {
+    // TODO: Is it safe to access the context here?
+
+    auto context = context_weak.lock();
+    if (!context)
+        return;
+
+
+    if (context->recorder) {
+        auto reply = QMessageBox::question(this, tr("CiTracing still active"),
+                tr("A CiTrace is still being recorded. Do you want to save it? If not, all recorded data will be discarded."),
+                QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
+
+        if (reply == QMessageBox::Yes) {
+            StopRecording();
+        } else {
+            AbortRecording();
+        }
+    }
+
+    // If the widget was disabled before, enable it now to allow starting
+    // tracing before starting the next emulation session
+    widget()->setEnabled(true);
 }
--- a/src/citra_qt/debugger/graphics_tracing.cpp
+++ b/src/citra_qt/debugger/graphics_tracing.cpp
@@ -13,6 +13,8 @@
 #include <QSpinBox>
 
 #include "core/hw/gpu.h"
+#include "core/hw/lcd.h"
+
 #include "video_core/pica.h"
 
 #include "nihstro/float24.h"
@@ -62,23 +64,25 @@ void GraphicsTracingWidget::StartRecording() {
 
     // Encode floating point numbers to 24-bit values
     // TODO: Drop this explicit conversion once we store float24 values bit-correctly internally.
-    std::array<Math::Vec4<uint32_t>, 16> default_attributes;
+    std::array<uint32_t, 4 * 16> default_attributes;
     for (unsigned i = 0; i < 16; ++i) {
         for (unsigned comp = 0; comp < 3; ++comp) {
-            default_attributes[i][comp] = nihstro::to_float24(Pica::g_state.vs.default_attributes[i][comp].ToFloat32());
+            default_attributes[4 * i + comp] = nihstro::to_float24(Pica::g_state.vs.default_attributes[i][comp].ToFloat32());
         }
     }
 
-    std::array<Math::Vec4<uint32_t>, 96> vs_float_uniforms;
+    std::array<uint32_t, 4 * 96> vs_float_uniforms;
     for (unsigned i = 0; i < 96; ++i)
         for (unsigned comp = 0; comp < 3; ++comp)
-            vs_float_uniforms[i][comp] = nihstro::to_float24(Pica::g_state.vs.uniforms.f[i][comp].ToFloat32());
+            vs_float_uniforms[4 * i + comp] = nihstro::to_float24(Pica::g_state.vs.uniforms.f[i][comp].ToFloat32());
 
-    auto recorder = new CiTrace::Recorder((u32*)&GPU::g_regs, 0x700, nullptr, 0, (u32*)&Pica::g_state.regs, 0x300,
-                                          (u32*)default_attributes.data(), default_attributes.size() * 4,
+    auto recorder = new CiTrace::Recorder((u32*)&GPU::g_regs, sizeof(GPU::g_regs) / sizeof(u32),
+                                          (u32*)&LCD::g_regs, sizeof(LCD::g_regs) / sizeof(u32),
+                                          (u32*)&Pica::g_state.regs, sizeof(Pica::g_state.regs) / sizeof(u32),
+                                          default_attributes.data(), default_attributes.size(),
                                           shader_binary.data(), shader_binary.size(),
                                           swizzle_data.data(), swizzle_data.size(),
-                                          (u32*)vs_float_uniforms.data(), vs_float_uniforms.size() * 4,
+                                          vs_float_uniforms.data(), vs_float_uniforms.size(),
                                           nullptr, 0, nullptr, 0, nullptr, 0 // Geometry shader is not implemented yet, so submit dummy data for now
                                           );
     context->recorder = std::shared_ptr<CiTrace::Recorder>(recorder);
--- a/src/citra_qt/debugger/graphics_tracing.cpp
+++ b/src/citra_qt/debugger/graphics_tracing.cpp
@@ -1,0 +1,123 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <memory>
+
+#include <QBoxLayout>
+#include <QComboBox>
+#include <QFileDialog>
+#include <QLabel>
+#include <QPushButton>
+#include <QSpinBox>
+
+#include "core/hw/gpu.h"
+#include "video_core/pica.h"
+
+#include "nihstro/float24.h"
+
+#include "graphics_tracing.h"
+
+GraphicsTracingWidget::GraphicsTracingWidget(std::shared_ptr<Pica::DebugContext> debug_context,
+                                             QWidget* parent)
+    : BreakPointObserverDock(debug_context, tr("CiTrace Recorder"), parent) {
+
+    setObjectName("CiTracing");
+
+    QPushButton* start_recording = new QPushButton(tr("Start Recording"));
+    QPushButton* stop_recording = new QPushButton(QIcon::fromTheme("document-save"), tr("Stop and Save"));
+    QPushButton* abort_recording = new QPushButton(tr("Abort Recording"));
+
+    connect(this, SIGNAL(SetStartTracingButtonEnabled(bool)), start_recording, SLOT(setVisible(bool)));
+    connect(this, SIGNAL(SetStopTracingButtonEnabled(bool)), stop_recording, SLOT(setVisible(bool)));
+    connect(this, SIGNAL(SetAbortTracingButtonEnabled(bool)), abort_recording, SLOT(setVisible(bool)));
+    connect(start_recording, SIGNAL(clicked()), this, SLOT(StartRecording()));
+    connect(stop_recording, SIGNAL(clicked()), this, SLOT(StopRecording()));
+    connect(abort_recording, SIGNAL(clicked()), this, SLOT(AbortRecording()));
+
+    stop_recording->setVisible(false);
+    abort_recording->setVisible(false);
+
+    auto main_widget = new QWidget;
+    auto main_layout = new QVBoxLayout;
+    {
+        auto sub_layout = new QHBoxLayout;
+        sub_layout->addWidget(start_recording);
+        sub_layout->addWidget(stop_recording);
+        sub_layout->addWidget(abort_recording);
+        main_layout->addLayout(sub_layout);
+    }
+    main_widget->setLayout(main_layout);
+    setWidget(main_widget);
+
+    // TODO: Make sure to have this widget disabled as soon as emulation is started!
+}
+
+void GraphicsTracingWidget::StartRecording() {
+    auto context = context_weak.lock();
+    if (!context)
+        return;
+
+    auto shader_binary = Pica::g_state.vs.program_code;
+    auto swizzle_data = Pica::g_state.vs.swizzle_data;
+
+    // Encode floating point numbers to 24-bit values
+    // TODO: Drop this explicit conversion once we store float24 values bit-correctly internally.
+    std::array<Math::Vec4<uint32_t>, 96> vs_float_uniforms;
+    for (unsigned i = 0; i < 96; ++i)
+        for (unsigned comp = 0; comp < 3; ++comp)
+            vs_float_uniforms[i][comp] = nihstro::to_float24(Pica::g_state.vs.uniforms.f[i][comp].ToFloat32());
+
+    auto recorder = new CiTrace::Recorder((u32*)&GPU::g_regs, 0x700, nullptr, 0, (u32*)&Pica::g_state.regs, 0x300,
+                                          shader_binary.data(), shader_binary.size(),
+                                          swizzle_data.data(), swizzle_data.size(),
+                                          (u32*)vs_float_uniforms.data(), vs_float_uniforms.size() * 4,
+                                          nullptr, 0, nullptr, 0, nullptr, 0 // Geometry shader is not implemented yet, so submit dummy data for now
+                                          );
+    context->recorder = std::shared_ptr<CiTrace::Recorder>(recorder);
+
+    emit SetStartTracingButtonEnabled(false);
+    emit SetStopTracingButtonEnabled(true);
+    emit SetAbortTracingButtonEnabled(true);
+}
+
+void GraphicsTracingWidget::StopRecording() {
+    auto context = context_weak.lock();
+    if (!context)
+        return;
+
+    QString filename = QFileDialog::getSaveFileName(this, tr("Save CiTrace"), "citrace.ctf",
+                                                    tr("CiTrace File (*.ctf)"));
+
+    if (filename.isEmpty()) {
+        // If the user canceled the dialog, keep recording
+        return;
+    }
+
+    context->recorder->Finish(filename.toStdString());
+    context->recorder = nullptr;
+
+    emit SetStopTracingButtonEnabled(false);
+    emit SetAbortTracingButtonEnabled(false);
+    emit SetStartTracingButtonEnabled(true);
+}
+
+void GraphicsTracingWidget::AbortRecording() {
+    auto context = context_weak.lock();
+    if (!context)
+        return;
+
+    context->recorder = nullptr;
+
+    emit SetStopTracingButtonEnabled(false);
+    emit SetAbortTracingButtonEnabled(false);
+    emit SetStartTracingButtonEnabled(true);
+}
+
+void GraphicsTracingWidget::OnBreakPointHit(Pica::DebugContext::Event event, void* data) {
+    widget()->setEnabled(true);
+}
+
+void GraphicsTracingWidget::OnResumed() {
+    widget()->setEnabled(false);
+}
--- a/src/citra_qt/debugger/graphics_tracing.cpp
+++ b/src/citra_qt/debugger/graphics_tracing.cpp
@@ -62,12 +62,20 @@ void GraphicsTracingWidget::StartRecording() {
 
     // Encode floating point numbers to 24-bit values
     // TODO: Drop this explicit conversion once we store float24 values bit-correctly internally.
+    std::array<Math::Vec4<uint32_t>, 16> default_attributes;
+    for (unsigned i = 0; i < 16; ++i) {
+        for (unsigned comp = 0; comp < 3; ++comp) {
+            default_attributes[i][comp] = nihstro::to_float24(Pica::g_state.vs.default_attributes[i][comp].ToFloat32());
+        }
+    }
+
     std::array<Math::Vec4<uint32_t>, 96> vs_float_uniforms;
     for (unsigned i = 0; i < 96; ++i)
         for (unsigned comp = 0; comp < 3; ++comp)
             vs_float_uniforms[i][comp] = nihstro::to_float24(Pica::g_state.vs.uniforms.f[i][comp].ToFloat32());
 
     auto recorder = new CiTrace::Recorder((u32*)&GPU::g_regs, 0x700, nullptr, 0, (u32*)&Pica::g_state.regs, 0x300,
+                                          (u32*)default_attributes.data(), default_attributes.size() * 4,
                                           shader_binary.data(), shader_binary.size(),
                                           swizzle_data.data(), swizzle_data.size(),
                                           (u32*)vs_float_uniforms.data(), vs_float_uniforms.size() * 4,
--- a/src/citra_qt/debugger/graphics_tracing.h
+++ b/src/citra_qt/debugger/graphics_tracing.h
@@ -6,6 +6,8 @@
 
 #include "graphics_breakpoint_observer.h"
 
+class EmuThread;
+
 class GraphicsTracingWidget : public BreakPointObserverDock {
     Q_OBJECT
 
@@ -19,6 +21,9 @@ private slots:
 
     void OnBreakPointHit(Pica::DebugContext::Event event, void* data) override;
     void OnResumed() override;
+
+    void OnEmulationStarting(EmuThread* emu_thread);
+    void OnEmulationStopping();
 
 signals:
     void SetStartTracingButtonEnabled(bool enable);
--- a/src/citra_qt/debugger/graphics_tracing.h
+++ b/src/citra_qt/debugger/graphics_tracing.h
@@ -1,0 +1,27 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "graphics_breakpoint_observer.h"
+
+class GraphicsTracingWidget : public BreakPointObserverDock {
+    Q_OBJECT
+
+public:
+    GraphicsTracingWidget(std::shared_ptr<Pica::DebugContext> debug_context, QWidget* parent = nullptr);
+
+private slots:
+    void StartRecording();
+    void StopRecording();
+    void AbortRecording();
+
+    void OnBreakPointHit(Pica::DebugContext::Event event, void* data) override;
+    void OnResumed() override;
+
+signals:
+    void SetStartTracingButtonEnabled(bool enable);
+    void SetStopTracingButtonEnabled(bool enable);
+    void SetAbortTracingButtonEnabled(bool enable);
+};
--- a/src/common/file_util.h
+++ b/src/common/file_util.h
@@ -181,6 +181,10 @@ public:
     template <typename T>
     size_t WriteArray(const T* data, size_t length)
     {
+        static_assert(std::is_standard_layout<T>::value, "Given array does not consist of standard layout objects");
+        // TODO: gcc 4.8 does not support is_trivially_copyable, but we really should check for it here.
+        //static_assert(std::is_trivially_copyable<T>::value, "Given array does not consist of trivially copyable objects");
+
         if (!IsOpen()) {
             m_good = false;
             return -1;
@@ -201,6 +205,12 @@ public:
     size_t WriteBytes(const void* data, size_t length)
     {
         return WriteArray(reinterpret_cast<const char*>(data), length);
+    }
+
+    template<typename T>
+    size_t WriteObject(const T& object) {
+        static_assert(!std::is_pointer<T>::value, "Given object is a pointer");
+        return WriteArray(&object, 1);
     }
 
     bool IsOpen() { return nullptr != m_file; }
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -245,7 +245,7 @@ set(HEADERS
             loader/loader.h
             loader/ncch.h
             tracer/recorder.h
-            tracer/tracer.h
+            tracer/citrace.h
             mem_map.h
             memory.h
             memory_setup.h
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -115,6 +115,7 @@ set(SRCS
             loader/elf.cpp
             loader/loader.cpp
             loader/ncch.cpp
+            tracer/recorder.cpp
             mem_map.cpp
             memory.cpp
             settings.cpp
@@ -243,6 +244,8 @@ set(HEADERS
             loader/elf.h
             loader/loader.h
             loader/ncch.h
+            tracer/recorder.h
+            tracer/tracer.h
             mem_map.h
             memory.h
             memory_setup.h
--- a/src/core/hw/gpu.cpp
+++ b/src/core/hw/gpu.cpp
@@ -101,39 +101,43 @@ inline void Write(u32 addr, const T data) {
         const bool is_second_filler = (index != GPU_REG_INDEX(memory_fill_config[0].trigger));
         auto& config = g_regs.memory_fill_config[is_second_filler];
 
-        if (config.address_start && config.trigger) {
-            u8* start = Memory::GetPhysicalPointer(config.GetStartAddress());
-            u8* end = Memory::GetPhysicalPointer(config.GetEndAddress());
+        if (config.trigger) {
+            if (config.address_start) { // Some games pass invalid values here
+                u8* start = Memory::GetPhysicalPointer(config.GetStartAddress());
+                u8* end = Memory::GetPhysicalPointer(config.GetEndAddress());
 
-            if (config.fill_24bit) {
-                // fill with 24-bit values
-                for (u8* ptr = start; ptr < end; ptr += 3) {
-                    ptr[0] = config.value_24bit_r;
-                    ptr[1] = config.value_24bit_g;
-                    ptr[2] = config.value_24bit_b;
+                if (config.fill_24bit) {
+                    // fill with 24-bit values
+                    for (u8* ptr = start; ptr < end; ptr += 3) {
+                        ptr[0] = config.value_24bit_r;
+                        ptr[1] = config.value_24bit_g;
+                        ptr[2] = config.value_24bit_b;
+                    }
+                } else if (config.fill_32bit) {
+                    // fill with 32-bit values
+                    for (u32* ptr = (u32*)start; ptr < (u32*)end; ++ptr)
+                        *ptr = config.value_32bit;
+                } else {
+                    // fill with 16-bit values
+                    for (u16* ptr = (u16*)start; ptr < (u16*)end; ++ptr)
+                        *ptr = config.value_16bit;
                 }
-            } else if (config.fill_32bit) {
-                // fill with 32-bit values
-                for (u32* ptr = (u32*)start; ptr < (u32*)end; ++ptr)
-                    *ptr = config.value_32bit;
-            } else {
-                // fill with 16-bit values
-                for (u16* ptr = (u16*)start; ptr < (u16*)end; ++ptr)
-                    *ptr = config.value_16bit;
+
+                LOG_TRACE(HW_GPU, "MemoryFill from 0x%08x to 0x%08x", config.GetStartAddress(), config.GetEndAddress());
+
+                if (!is_second_filler) {
+                    GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::PSC0);
+                } else {
+                    GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::PSC1);
+                }
+
+                VideoCore::g_renderer->hw_rasterizer->NotifyFlush(config.GetStartAddress(), config.GetEndAddress() - config.GetStartAddress());
             }
 
-            LOG_TRACE(HW_GPU, "MemoryFill from 0x%08x to 0x%08x", config.GetStartAddress(), config.GetEndAddress());
-
+            // Reset "trigger" flag and set the "finish" flag
+            // NOTE: This was confirmed to happen on hardware even if "address_start" is zero.
             config.trigger = 0;
             config.finished = 1;
-
-            if (!is_second_filler) {
-                GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::PSC0);
-            } else {
-                GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::PSC1);
-            }
-
-            VideoCore::g_renderer->hw_rasterizer->NotifyFlush(config.GetStartAddress(), config.GetEndAddress() - config.GetStartAddress());
         }
         break;
     }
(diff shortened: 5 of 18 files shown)


src/core/hle/service/soc_u.cpp
```diff
@@ -320,6 +320,9 @@ static void Socket(Service::Interface* self) {
 
     u32 socket_handle = static_cast<u32>(::socket(domain, type, protocol));
 
+    const int sock_opt = 1;
+    ::setsockopt(socket_handle, SOL_SOCKET, SO_REUSEADDR, &sock_opt, sizeof(sock_opt));
```
archshift: Unrelated change?

neobrain: It's required to use the CiTrace player without changing the port after each session, because otherwise creating the listening socket will fail after N runs.

src/common/file_util.h
```diff
@@ -204,6 +204,15 @@ class IOFile : public NonCopyable
         return WriteArray(reinterpret_cast<const char*>(data), length);
     }
 
+    template<typename T>
+    size_t WriteObject(const T& object) {
+        static_assert(std::is_standard_layout<T>::value, "Given object is not of standard layout");
+        // TODO: Not supported by gcc 4.8...
+        //static_assert(std::is_trivially_copyable<T>::value, "Given object is not trivially copyable");
+        static_assert(!std::is_pointer<T>::value, "Given object is a pointer");
+        return WriteBytes(&object, sizeof(object)) / sizeof(object);
```
linkmauve: WriteArray would be better suited here.

neobrain: Indeed. I'll change it, and move the static_asserts to that functions instead.

src/core/hle/service/soc_u.cpp
yuriks: Why? And why does SO_RESUEADDR fix this?

neobrain: I guess this should instead be set in the citrace-player, so nevermind (albeit I think it would still be sensible to set this option by default in Citra).

neobrain: The actual is that if an application creates a connection on a particular port and is then forcefully killed and the application is restarted and another attempt for connection is made, there will be an "Address already in use" error in one of the socket setup calls.

yuriks: What @neobrain seems to be hitting is trying to rebind ports when a socket is still in TIME_WAIT (waiting for a clean connection close). SO_REUSEADDR works around this and seems to have no significant disvantages, so it's fine. I think a comment is warranted.

src/core/hw/gpu.cpp
```diff
@@ -259,7 +277,16 @@ inline void Write(u32 addr, const T data) {
         if (config.trigger & 1)
         {
             u32* buffer = (u32*)Memory::GetPointer(Memory::PhysicalToVirtualAddress(config.GetPhysicalAddress()));
+            LOG_ERROR(HW_GPU, "Loading list from paddr %x, i.e. vaddr %x",
```
neobrain: Remove this.

src/core/hw/gpu.cpp
```diff
@@ -59,6 +63,8 @@ inline void Write(u32 addr, const T data) {
     addr -= HW::VADDR_GPU;
     u32 index = addr / 4;
 
+    LOG_ERROR(HW_GPU, "Writing %x to %x", (u32)data, addr);
```
neobrain: Remove this.

src/core/hw/gpu.cpp
```diff
@@ -130,12 +138,21 @@ inline void Write(u32 addr, const T data) {
             u32 output_width = config.output_width / horizontal_scale;
             u32 output_height = config.output_height / vertical_scale;
 
+            if (Pica::g_debug_context && Pica::g_debug_context->recorder) {
+                // Only record memory access if this display transfer happens from non-VRAM and goes to VRAM
+                // TODO: Figure out something better than this to distinguish between renderer-generated data and user-provided data.
+                if ((config.GetPhysicalInputAddress() + /* input_size */1 <= 0x18000000 || config.GetPhysicalInputAddress() > 0x18600000) &&
+                    (config.GetPhysicalOutputAddress() >= 0x18000000 && config.GetPhysicalOutputAddress() < 0x18600000)) {
+//                    Pica::g_debug_context->recorder->MemoryAccessed(src_pointer, config.input_width * config.input_height * GPU::Regs::BytesPerPixel(config.input_format), config.GetPhysicalInputAddress());
```
bunnei: ?

Subv:
What's the status of this PR?

neobrain:
Addressed review comments. Reverted the SOC_U change because it's out of place in citra (I'll add it in the actual 3DS application instead).

Please perform your *final* review so that I can rebase and make sure it still works.

bunnei:
Please rebase latest master so that CI can build

src/video_core/command_processor.cpp
```diff
@@ -182,10 +228,16 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
                 // Send to triangle clipper
                 clipper_primitive_assembler.SubmitVertex(output, Clipper::ProcessTriangle);
             }
+            int i = 0;
```
bunnei: Unused?

neobrain:
Rebased (old version at neobrain@d313426973a155244b6136ff65fd4181b6e3ecde)

src/video_core/command_processor.cpp
```diff
@@ -110,12 +110,53 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                    if (!texture.enabled)
+                        continue;
+
+                    u8* texture_data = Memory::GetPhysicalPointer(texture.config.GetPhysicalAddress());
+                    if (g_debug_context && Pica::g_debug_context->recorder)
+                        g_debug_context->recorder->MemoryAccessed(texture_data, Pica::Regs::NibblesPerPixel(texture.format) * texture.config.width / 2 * texture.config.height, texture.config.GetPhysicalAddress());
+                }
+            }
+
+            // map physical start address to size
```
bunnei: *Map

src/video_core/command_processor.cpp
```diff
@@ -110,12 +110,53 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+            std::map<u32, u32> accessed_ranges;
+            static auto SimplifyRanges = [](std::map<u32, u32>& ranges) {
+                for (auto it = ranges.begin(); it != ranges.end(); ++it) {
+
+                    // Combine overlapping ranges ... artificially extend first range by 32 bytes to merge "close" ranges
+                    auto it2 = std::next(it);
+                    while (it2 != ranges.end() && it->first + it->second + 32 >= it2->first) {
+                        if (it2->first < it->first)
+                            exit(2);
```
bunnei: Please don't do this.... assert or something

src/video_core/clipper.cpp
```diff
@@ -161,6 +161,7 @@ void ProcessTriangle(OutputVertex &v0, OutputVertex &v1, OutputVertex &v2) {
                   vtx2.screenpos.x.ToFloat32(), vtx2.screenpos.y.ToFloat32(), vtx2.screenpos.z.ToFloat32());
 
         Rasterizer::ProcessTriangle(vtx0, vtx1, vtx2);
+        Rasterizer::ProcessTriangle(vtx0, vtx2, vtx1);
```
bunnei: What?

src/core/tracer/tracer.h
```diff
@@ -0,0 +1,98 @@
...
+    FrameMarker   = 0xE1,
+    MemoryLoad    = 0xE2,
+    RegisterWrite = 0xE3,
+};
+
+struct CTMemoryLoad {
+    uint32_t file_offset;
+    uint32_t size;
+    uint32_t physical_address;
+	uint32_t pad;
```
bunnei: Tab

src/core/tracer/tracer.h
```diff
@@ -0,0 +1,98 @@
...
+// NOTE: Things are stored in little-endian
+
+#pragma pack(1)
+
+struct CTHeader {
+    static const char* ExpectedMagicWord() {
+        return "CiTr";
+    }
+
+    static uint32_t ExpectedVersion() {
```
bunnei: Why are you using uint types in this file? u32/u64 please

src/core/tracer/recorder.cpp
```diff
@@ -0,0 +1,198 @@
...
+        if (file.Tell() != header.stream_offset)
+            throw "Unexpected end of extra data";
+
+        // Write actual stream elements
+        for (const auto& stream_element : stream) {
+            if (1 != file.WriteObject(stream_element.data))
+                throw "Failed to write stream element";
+        }
+    } catch(const char* str) {
+        LOG_ERROR(HW_GPU, "Writing CiTrace file failed: %s", str);
```
bunnei: Should this be a critical?

src/core/CMakeLists.txt
```diff
@@ -98,6 +98,10 @@ set(SRCS
             loader/elf.cpp
             loader/loader.cpp
             loader/ncch.cpp
+            loader/3dsx.cpp
+            tracer/recorder.cpp
+            core.cpp
+            core_timing.cpp
```
bunnei: Why are you adding `core.cpp` and `core_timing.cpp` twice?

src/common/file_util.h
```diff
@@ -204,6 +208,12 @@ class IOFile : public NonCopyable
         return WriteArray(reinterpret_cast<const char*>(data), length);
     }
 
+    template<typename T>
+    size_t WriteObject(const T& object) {
+        static_assert(!std::is_pointer<T>::value, "Given object is a pointer");
```
bunnei: Should be "Given object is not a pointer" ?

bunnei:
Finished reviewing... still a few mistakes, overall change looks fine though

Subv:
I feel like there are already too many lambdas in the code, is there any reason why you're adding these as lambdas instead of normal functions?

src/video_core/command_processor.cpp
neobrain: Oh my god, I totally forgot about this.. yeah, definitely.

src/core/tracer/recorder.cpp
neobrain: Nope, if this fails, the citrace will not be dumped, but emulation should be able to continue just fine. IMO critical logs should be reserved for errors which are unrecoverable.

src/core/tracer/tracer.h
neobrain: This header is used in citrace_player too, for which I'd rather not use the ugly `uX` types.

src/core/CMakeLists.txt
neobrain: Incorrectly fixed merge conflict, sorry :/

src/video_core/command_processor.cpp
```diff
@@ -110,12 +110,53 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+
+                    u8* texture_data = Memory::GetPhysicalPointer(texture.config.GetPhysicalAddress());
+                    if (g_debug_context && Pica::g_debug_context->recorder)
+                        g_debug_context->recorder->MemoryAccessed(texture_data, Pica::Regs::NibblesPerPixel(texture.format) * texture.config.width / 2 * texture.config.height, texture.config.GetPhysicalAddress());
+                }
+            }
+
+            // map physical start address to size
+            std::map<u32, u32> accessed_ranges;
+            static auto SimplifyRanges = [](std::map<u32, u32>& ranges) {
```
neobrain: Consider promoting `accessed_ranges` to a class, of which SimplifyRanges and AddMemoryAccess can be simply made member functions.

neobrain:
@Subv as for me personally, it makes no sense to introduce new top-level functions for functionality which is only used in one function anyway. Making it a lambda makes it dead-obvious that the code will only be used in the current function, and nowhere else. It also allows for a more linear code reading, since there's no need to jump between different functions all the time when trying to understand the code.

That said, I think in this particular case accessed_ranges and the two lambdas operating on it would be better suited as a locally implemented utility class.

src/citra_qt/debugger/graphics_tracing.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+void GraphicsTracingWidget::StartRecording() {
+    auto context = context_weak.lock();
+    auto shader_binary = Pica::VertexShader::GetShaderBinary();
+    auto swizzle_data = Pica::VertexShader::GetSwizzlePatterns();
+
+    // Encode floating point numbers to 24-bit values
+    // TODO: Drop this explicit conversion once we store float24 values bit-correctly internally.
+    std::array<Math::Vec4<uint32_t>, 96> vs_float_uniforms;
+    for (unsigned i = 0; i < 96; ++i)
```
archshift: Braces around this? (`for` spans more than one line, and gets a bit messy otherwise)

src/core/tracer/recorder.h
```diff
@@ -0,0 +1,92 @@
...
+        CTStreamElement data;
+
+        /**
+          * Extra data to store along "core" data.
+          * This is e.g. used for data used in MemoryUpdates.
+          */
+        std::vector<u8> extra_data;
+
+        /// Optional CRC hash (e.g. for hashing memory regions)
+        boost::crc_32_type::value_type hash;
```
archshift: Is there any way you can forward-declare this to avoid adding in the entire boost header in all the files that include this one?

src/citra_qt/debugger/graphics_tracing.cpp
```diff
@@ -0,0 +1,116 @@
...
+    connect(start_recording, SIGNAL(clicked()), this, SLOT(StartRecording()));
+    connect(stop_recording, SIGNAL(clicked()), this, SLOT(StopRecording()));
+    connect(abort_recording, SIGNAL(clicked()), this, SLOT(AbortRecording()));
+
+    stop_recording->setVisible(false);
+    abort_recording->setVisible(false);
+
+    auto main_widget = new QWidget;
+    auto main_layout = new QVBoxLayout;
+    {
```
archshift: Why this extra subscope?

src/core/tracer/recorder.h
neobrain: I don't think so.

src/citra_qt/debugger/graphics_tracing.cpp
neobrain: We use this kind of style commonly in Qt code, so that each "sub_layout" goes into its own scope. it makes adding multiple sublayouts easier on the eyes, and clearly structures the code.

archshift:
All in all, LGTM after you address the outstanding comments.

src/video_core/command_processor.cpp
```diff
@@ -110,12 +110,53 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+
             for (unsigned int index = 0; index < registers.num_vertices; ++index)
             {
                 unsigned int vertex = is_indexed ? (index_u16 ? index_address_16[index] : index_address_8[index]) : index;
 
                 if (is_indexed) {
                     // TODO: Implement some sort of vertex cache!
+                    if (g_debug_context && Pica::g_debug_context->recorder) {
+                        int size = index_u16 ? 2 : 1;
+                        AddMemoryAccess(accessed_ranges, base_address + index_info.offset + size*index, size);
```
Subv: How much does this call affect performance?

src/citra_qt/debugger/graphics_tracing.cpp
```diff
@@ -0,0 +1,116 @@
...
+    auto swizzle_data = Pica::VertexShader::GetSwizzlePatterns();
+
+    // Encode floating point numbers to 24-bit values
+    // TODO: Drop this explicit conversion once we store float24 values bit-correctly internally.
+    std::array<Math::Vec4<uint32_t>, 96> vs_float_uniforms;
+    for (unsigned i = 0; i < 96; ++i)
+        for (unsigned comp = 0; comp < 3; ++comp)
+            vs_float_uniforms[i][comp] = nihstro::to_float24(Pica::VertexShader::GetFloatUniform(i)[comp].ToFloat32());
+
+    auto recorder = new CiTrace::Recorder((u32*)&GPU::g_regs, 0x700, nullptr, 0, (u32*)&Pica::registers, 0x300,
```
Subv: It doesn't seem like you're saving the default attributes to the CiTrace

neobrain: Good point!

src/video_core/command_processor.cpp
neobrain: Not at all, unless citracing is active (in which case we don't care, I would say).

neobrain:
All comments should've been addressed by now.

Some new code was added to 1) record default vertex attributes 2) properly enable/disable the citrace widget upon stopping/starting emulation.

neobrain:
There's some potential for merge conflicts with regards to the pull requests  #776 and  #789. I suggest merging #776 first, then this PR, then #789, and deal with any merge conflicts in that same order.

neobrain:
Rebased.

(old version was f8d8e5ff620)

neobrain:
Chances are the hardware renderer needs changes for this to work. CC @tfarley.

neobrain:
Chances are the hardware renderer needs changes for this to work. CC @tfarley.

In any case, I don't consider this blocking this PR.

src/citra_qt/debugger/graphics_tracing.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+void GraphicsTracingWidget::StartRecording() {
+    auto context = context_weak.lock();
+    auto shader_binary = Pica::g_state.vs.program_code;
+    auto swizzle_data = Pica::g_state.vs.swizzle_data;
+
+    // Encode floating point numbers to 24-bit values
+    // TODO: Drop this explicit conversion once we store float24 values bit-correctly internally.
+    std::array<Math::Vec4<uint32_t>, 16> default_attributes;
+    for (unsigned i = 0; i < 16; ++i) {
```
Subv: Why do these loops have brackets and the vs_float_uniforms loops don't?

src/citra_qt/debugger/graphics_tracing.cpp
```diff
@@ -0,0 +1,148 @@
...
+            default_attributes[i][comp] = nihstro::to_float24(Pica::g_state.vs.default_attributes[i][comp].ToFloat32());
+        }
+    }
+
+    std::array<Math::Vec4<uint32_t>, 96> vs_float_uniforms;
+    for (unsigned i = 0; i < 96; ++i)
+        for (unsigned comp = 0; comp < 3; ++comp)
+            vs_float_uniforms[i][comp] = nihstro::to_float24(Pica::g_state.vs.uniforms.f[i][comp].ToFloat32());
+
+    auto recorder = new CiTrace::Recorder((u32*)&GPU::g_regs, 0x700, nullptr, 0, (u32*)&Pica::g_state.regs, 0x300,
```
Subv: What do 0x700 and 0x300 mean?

src/citra_qt/debugger/graphics_tracing.cpp
```diff
@@ -0,0 +1,148 @@
...
+                QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
+
+        if (reply == QMessageBox::Yes) {
+            StopRecording();
+        } else {
+            AbortRecording();
+        }
+    }
+
+    // Allow starting tracing before emulation is started
```
Subv: What does this comment mean? We're in OnEmulationStopping

src/common/file_util.h
```diff
@@ -182,6 +182,10 @@ class IOFile : public NonCopyable
     template <typename T>
     size_t WriteArray(const T* data, size_t length)
     {
+        static_assert(std::is_standard_layout<T>::value, "Given array does not consist of standard layout objects");
+        // TODO: gcc 4.8 does not support is_trivially_copyable, but we really should check for it here.
+        //static_assert(std::is_trivially_copyable<T>::value, "Given array does not consist of trivially copyable objects");
```
Subv: Dolphin has a workaround for this, see https://github.com/dolphin-emu/dolphin/pull/2440/files

src/video_core/command_processor.cpp
```diff
@@ -111,12 +111,55 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                    if (!texture.enabled)
+                        continue;
+
+                    u8* texture_data = Memory::GetPhysicalPointer(texture.config.GetPhysicalAddress());
+                    if (g_debug_context && Pica::g_debug_context->recorder)
+                        g_debug_context->recorder->MemoryAccessed(texture_data, Pica::Regs::NibblesPerPixel(texture.format) * texture.config.width / 2 * texture.config.height, texture.config.GetPhysicalAddress());
+                }
+            }
+
+            class {
```
Subv: What?

src/video_core/command_processor.cpp
```diff
@@ -111,12 +111,55 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+
             for (unsigned int index = 0; index < regs.num_vertices; ++index)
             {
                 unsigned int vertex = is_indexed ? (index_u16 ? index_address_16[index] : index_address_8[index]) : index;
 
                 if (is_indexed) {
                     // TODO: Implement some sort of vertex cache!
+                    if (g_debug_context && Pica::g_debug_context->recorder) {
+                        int size = index_u16 ? 2 : 1;
+                        memory_accesses.AddAccess(base_address + index_info.offset + size * index, size);
```
Subv: How much does the SimplifyRanges operation impact the running time of this loop?

src/video_core/command_processor.cpp
neobrain: The case in the switch of this function is literally the only place where this class is used. Moving it outside this function would make it appear like a more generic class and hence isn't any less confusing than having a local class.

src/video_core/command_processor.cpp
neobrain: Not at all unless citracing is active.

src/common/file_util.h
neobrain: I guess we could use that, but at the same time I think it's not particularly critical to address it in this PR, is it?

src/citra_qt/debugger/graphics_tracing.cpp
neobrain: I'll clarify it.

src/citra_qt/debugger/graphics_tracing.cpp
neobrain: static_cast is not a legal method to cast between unrelated pointer types. I guess I should use reinterpret_cast, though, even though that's not valid either, but at least compiles with strict aliasing rules disabled.

Subv: Why are you not passing the LCD registers?

neobrain: Heh, I was about to push a fix for that ;)

At the time I wrote this code, LCD hadn't been implemented.

neobrain:
Addressed another round of comments.

* It was suggested to use a C++-style cast in a situation where I'm casting a Regs* pointer to a u32* pointer. This is not actually possible without violating strict aliasing rules, and while a `reinterpret_cast` compiles and behaves perfectly fine for that purpose, it seems wrong to me to use a cast implying a particular intention when using that cast is not actually valid.
* It was suggested to move a local helper class into the global scope of the translation unit. I still think it makes more sense to have the class be a locally defined one because its usage is highly localized (to a case within a switch statement in a function).

yuriks:
Build is failing on all platforms...

neobrain:
.. fixed :/

src/citra_qt/debugger/graphics_tracing.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+#include "video_core/pica.h"
+
+#include "nihstro/float24.h"
+
+#include "graphics_tracing.h"
+
+GraphicsTracingWidget::GraphicsTracingWidget(std::shared_ptr<Pica::DebugContext> debug_context,
+                                             QWidget* parent)
+    : BreakPointObserverDock(debug_context, tr("CiTrace Recorder"), parent) {
```
yuriks: Missing "™" after "CiTrace".

src/citra_qt/debugger/graphics_tracing.h
```diff
@@ -0,0 +1,32 @@
...
+    void OnBreakPointHit(Pica::DebugContext::Event event, void* data) override;
+    void OnResumed() override;
+
+    void OnEmulationStarting(EmuThread* emu_thread);
+    void OnEmulationStopping();
+
+signals:
+    void SetStartTracingButtonEnabled(bool enable);
+    void SetStopTracingButtonEnabled(bool enable);
+    void SetAbortTracingButtonEnabled(bool enable);
```
yuriks: Wtf is the point of these signals? Why aren't they just a function???

src/citra_qt/debugger/graphics_tracing.cpp
```diff
@@ -0,0 +1,153 @@
...
+        sub_layout->addWidget(stop_recording);
+        sub_layout->addWidget(abort_recording);
+        main_layout->addLayout(sub_layout);
+    }
+    main_widget->setLayout(main_layout);
+    setWidget(main_widget);
+}
+
+void GraphicsTracingWidget::StartRecording() {
+    auto context = context_weak.lock();
```
yuriks: `context` isn't checked for being null before being used, in case acquiring the `weak_ptr` fails. The same error is made in all other uses in the PR.

src/core/tracer/recorder.cpp
```diff
@@ -0,0 +1,206 @@
...
+Recorder::Recorder(u32* gpu_registers,      u32 gpu_registers_size,
+                   u32* lcd_registers,      u32 lcd_registers_size,
+                   u32* pica_registers,     u32 pica_registers_size,
+                   u32* default_attributes, u32 default_attributes_size,
+                   u32* vs_program_binary,  u32 vs_program_binary_size,
+                   u32* vs_swizzle_data,    u32 vs_swizzle_data_size,
+                   u32* vs_float_uniforms,  u32 vs_float_uniforms_size,
+                   u32* gs_program_binary,  u32 gs_program_binary_size,
+                   u32* gs_swizzle_data,    u32 gs_swizzle_data_size,
+                   u32* gs_float_uniforms,  u32 gs_float_uniforms_size)
```
yuriks: This is simply terrible. A function with a zillion parameters is very error prone to use and confusing to read, especially seeing as they all have the same type, there's no way to know if you got the order right or to modify the parameter list without breaking users unknowingly. Define a parameters struct that you fill in and pass to the function instead.

yuriks: Also, considering that the treatment of all this data seems to be uniform, consider if you can't simply use an array or something.

src/citra_qt/debugger/graphics_tracing.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+void GraphicsTracingWidget::StopRecording() {
+    QString filename = QFileDialog::getSaveFileName(this, tr("Save CiTrace"), "citrace.ctf",
+                                                    tr("CiTrace File (*.ctf)"));
+
+    auto context = context_weak.lock();
+
+    if (filename.isEmpty()) {
+        // If no filename was given, keep recording
```
yuriks: "If the user cancelled the dialog, keep recording."

src/common/file_util.h
yuriks: Instead of this static assert, wouldn't it be more explicit to simply have the function accept a `T*` in the first place?

src/core/hw/gpu.cpp
```diff
@@ -75,25 +79,27 @@ inline void Write(u32 addr, const T data) {
...
-            u8* end = Memory::GetPhysicalPointer(config.GetEndAddress());
-
-            if (config.fill_24bit) {
-                // fill with 24-bit values
-                for (u8* ptr = start; ptr < end; ptr += 3) {
-                    ptr[0] = config.value_24bit_r;
-                    ptr[1] = config.value_24bit_g;
-                    ptr[2] = config.value_24bit_b;
+        if (config.trigger) {
+            if (config.address_start) { // Some games pass invalid values here
```
yuriks: Why split this?

neobrain:
There is an annoying trend in our PR review process, which I would like to clarify here. Case in point, the review made by @yuriks is IMO beyond any reasonable timeline. It's been more than a month since I opened this PR. No offense, but I don't feel like going through major iterations over something which was already public for weeks.

Apparently there was a misunderstanding as to how I'm addressing review comments - people seem to think that if one person gave his review on a PR, those review comments should be addressed before the next person can jump in to give his review. This is a really inefficent process though, since even in the better cases it requires multiple iterations on the PR (instead of just having the PR author address all comments in one go).

Let me make this clear for once: When opening PRs, my expectation is that everyone who wants to have a say on things will do so immediately without waiting for other events (delaying review is what we have the RFC tagfor). I'm not willing to put up with an inefficient review process which by design requires me to go through multiple iterations beyond the initial PRing phase. Sorry if this comes across a bit demand-ish, but it simply will not work any other way.

I will look over the "late review comments" which have been added earlier today and consider fixing them, but in the future my tendency will go towards ignoring such late comments unless their background has a critical impact on software quality.

src/common/file_util.h
Subv: The point of this assert is that you should not pass a pointer to WriteObject

yuriks: Yes, but I perceive it as trying to prevent people from accidentally passing a pointer when they meant to pass the object. I think this could be more directly done by just accepting a pointer in the first place.

neobrain: I can't think of any scenario where anyone would actually want to write a pointer to a file. Hence it's explicitly disallowed.

src/core/hw/hw.cpp
```diff
@@ -13,7 +13,7 @@ namespace HW {
 
 template <typename T>
 inline void Read(T &var, const u32 addr) {
-    switch (addr & 0xFFFFF000) {
+    switch (addr & 0xFFFF0000) {
```
yuriks: I assume you changed the mask because the GPU register range spans more than a page (0x1000). However all other registers (see http://3dbrew.org/wiki/IO) are on individual pages, so masking like this will cause writes to other ranges to fail. You should instead just add multiple cases for the GPU.

src/core/tracer/citrace.h
```diff
@@ -0,0 +1,101 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <cstdint>
+
+namespace CiTrace {
+
+// NOTE: Things are stored in little-endian
```
yuriks: Then shouldn't you be using `u32_le` types?

src/core/tracer/citrace.h
```diff
@@ -0,0 +1,101 @@
...
+
+#pragma pack(1)
```
yuriks: This doesn't seem necessary at all, since the entire struct has natural padding already.

src/core/tracer/citrace.h
```diff
@@ -0,0 +1,101 @@
...
+        SIZE_32 = 0xD3,
+        SIZE_64 = 0xD4
+    } size;
+
+    // TODO: Make it clearer which bits of this member are used for sizes other than 32 bits
+    uint64_t value;
+};
+
+struct CTStreamElement {
+    CTStreamElementType type;
```
yuriks: Do we also want to store timing info, in case real-time trace replays are desirable?

src/core/tracer/citrace.h
neobrain: Are any of these "natural alignment"s guaranteed by spec? If not, I'd rather keep the `#pragma pack(1)` to indeed guarantee it.

src/core/tracer/recorder.cpp
```diff
@@ -0,0 +1,206 @@
...
+      vs_swizzle_data(vs_swizzle_data, vs_swizzle_data + vs_swizzle_data_size),
+      vs_float_uniforms(vs_float_uniforms, vs_float_uniforms + vs_float_uniforms_size),
+      gs_program_binary(gs_program_binary, gs_program_binary + gs_program_binary_size),
+      gs_swizzle_data(gs_swizzle_data, gs_swizzle_data + gs_swizzle_data_size),
+      gs_float_uniforms(gs_float_uniforms, gs_float_uniforms + gs_float_uniforms_size) {
+
+}
+
+void Recorder::Finish(const std::string& filename) {
+    // TODO: Make sure this is only called when the gpu thread is paused!
```
yuriks: Have you made sure? Alternatively, should we try to assert that somehow?

src/core/tracer/citrace.h
neobrain: I don't think there's a reason to store this sort of information. In any case, it can be added in a later version if it's deemed useful after all.

src/core/tracer/recorder.cpp
neobrain: Indeed, I have made sure it's only called when the GPU thread is paused. This TODO should be removed.

I don't think there's a good way to assert for this condition in general, though.

src/core/tracer/recorder.h
yuriks: You can just use `u32`.

src/core/tracer/recorder.cpp
```diff
@@ -0,0 +1,206 @@
...
+    try {
+        // Open file and write header
+        FileUtil::IOFile file(filename, "wb");
+        size_t written = file.WriteObject(header);
+        if (written != 1 || file.Tell() != initial.gpu_registers)
+            throw "Failed to write header";
+
+        // Write initial state
+        written = file.WriteArray(gpu_registers.data(), gpu_registers.size());
+        if (written != gpu_registers.size() || file.Tell() != initial.lcd_registers)
```
yuriks: Doesn't `IOFile` already keep track of a "failed" flag that you can simply check at the end to know if something went wrong when writing? Additionally, asserting the cursor position after every read is redundant with the code above which calculates the offsets.

src/core/tracer/citrace.h
```diff
@@ -0,0 +1,101 @@
...
+    FrameMarker   = 0xE1,
+    MemoryLoad    = 0xE2,
+    RegisterWrite = 0xE3,
+};
+
+struct CTMemoryLoad {
+    uint32_t file_offset;
+    uint32_t size;
+    uint32_t physical_address;
+    uint32_t pad;
```
yuriks: Why the manual pad?

src/core/tracer/recorder.cpp
yuriks: (I you want to ensure you're at the right place I'd rather you seek beforehand so that at least the calculation of the offsets above can be changed without affecting this whole chain of tell checks.)

src/core/tracer/recorder.cpp
yuriks: Agree with neobrain

src/core/tracer/recorder.cpp
```diff
@@ -0,0 +1,206 @@
...
+                                     : (sizeof(T) == 2) ? CTRegisterWrite::SIZE_16
+                                     : (sizeof(T) == 4) ? CTRegisterWrite::SIZE_32
+                                     :                    CTRegisterWrite::SIZE_64;
+    element.data.register_write.physical_address = physical_address;
+    element.data.register_write.value = value;
+
+    stream.push_back(element);
+}
+
+template void Recorder::RegisterWritten(u32,u8);
```
yuriks: Space after the comma

src/core/tracer/recorder.cpp
```diff
@@ -0,0 +1,206 @@
...
+    // Compute hash over given memory region to check if the contents are already stored internally
+    boost::crc_32_type result;
+    result.process_bytes(data, size);
+    element.hash = result.checksum();
+
+    element.uses_existing_data = (memory_regions.find(result.checksum()) != memory_regions.end());
+    if (!element.uses_existing_data) {
+        element.extra_data.resize(size);
+        memcpy(element.extra_data.data(), data, size);
+        memory_regions.insert({result.checksum(), 0}); // file offset will be initialized in Finish()
```
yuriks: Why not simply use `element.hash` instead of `result.checksum()`? (Less redundancy, potentially faster if `checksum()` does a non-trivial amount of work.)

src/core/tracer/recorder.cpp
```diff
@@ -0,0 +1,206 @@
...
+    StreamElement element = { MemoryLoad };
+    element.data.memory_load.size = size;
+    element.data.memory_load.physical_address = physical_address;
+
+    // Compute hash over given memory region to check if the contents are already stored internally
+    boost::crc_32_type result;
+    result.process_bytes(data, size);
+    element.hash = result.checksum();
+
+    element.uses_existing_data = (memory_regions.find(result.checksum()) != memory_regions.end());
```
yuriks: Same comment as below.

src/core/tracer/recorder.cpp
yuriks: Is there any point in inserting the dummy zero here? Nothing seem to actually depends on this. (Be aware that operator[] automatically inserts a blank element in the map if the key isn't found.)

yuriks: Also, instead of `.insert({foo, bar})` you can just use `.emplace(foo, bar)`.

src/core/tracer/recorder.h
```diff
@@ -0,0 +1,95 @@
...
+        bool uses_existing_data;
+    };
+
+    std::vector<StreamElement> stream;
+
+    /**
+     * Internal cache which maps hashes of memory contents to file offsets at which those memory
+     * contents are stored.
+     */
+    std::map<boost::crc_32_type::value_type /*hash*/, u32 /*file_offset*/> memory_regions;
```
yuriks: Unless you need the data in sorted order, use `std::unordered_map`.

src/core/tracer/recorder.h
```diff
@@ -0,0 +1,95 @@
...
+
+    /// Mark end of a frame
+    void FrameFinished();
+
+    /**
+     * Store a copy of the given memory range in the recording.
+     * @note Use this whenever the GPU is about to access a particular memory region.
+     * @note The implementation will make sure to minimize redundant memory updates.
+     */
+    void MemoryAccessed(const u8* data, u32 size, u32 physical_address);
```
yuriks: The name of this func just sounds really weird and vague to me, especially when the documentation explicitly says what it actually does.

src/video_core/command_processor.cpp
yuriks: Agreed with @Subv. This breaks up the reading flow for something that is only used by tracing, which is an incidental feature that you'll almost never be paying attention to when reading this class.

src/video_core/command_processor.cpp
yuriks: Isn't there a way to know the size of the vertex buffer beforehand, and then just uploading the whole thing, instead of bothering with all this?

yuriks: (Even if there isn't, the simplification logic can be greatly simplified by just keep tracking of a single min,max address range for the whole draw operation. This will be efficient since vertices already have locality anyway.)

src/video_core/command_processor.cpp
```diff
@@ -136,10 +179,17 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
                     }
-                    
+
                     // Load per-vertex data from the loader arrays
                     for (unsigned int comp = 0; comp < vertex_attribute_elements[i]; ++comp) {
-                        const u8* srcdata = Memory::GetPhysicalPointer(vertex_attribute_sources[i] + vertex_attribute_strides[i] * vertex + comp * vertex_attribute_element_size[i]);
+                        u32 source_addr = vertex_attribute_sources[i] + vertex_attribute_strides[i] * vertex + comp * vertex_attribute_element_size[i];
+                        const u8* srcdata = Memory::GetPhysicalPointer(source_addr);
+
+                        if (g_debug_context && Pica::g_debug_context->recorder) {
```
yuriks: `g_debug_context` and then `Pica::g_debug_context`. Make it consistent.

src/video_core/command_processor.cpp
```diff
@@ -111,12 +111,55 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                std::map<u32, u32> ranges;
+            } memory_accesses;
+
             for (unsigned int index = 0; index < regs.num_vertices; ++index)
             {
                 unsigned int vertex = is_indexed ? (index_u16 ? index_address_16[index] : index_address_8[index]) : index;
 
                 if (is_indexed) {
                     // TODO: Implement some sort of vertex cache!
+                    if (g_debug_context && Pica::g_debug_context->recorder) {
```
yuriks: `Pica::` only used on the second access.

src/video_core/command_processor.cpp
```diff
@@ -136,10 +179,17 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
                     // Load per-vertex data from the loader arrays
                     for (unsigned int comp = 0; comp < vertex_attribute_elements[i]; ++comp) {
-                        const u8* srcdata = Memory::GetPhysicalPointer(vertex_attribute_sources[i] + vertex_attribute_strides[i] * vertex + comp * vertex_attribute_element_size[i]);
+                        u32 source_addr = vertex_attribute_sources[i] + vertex_attribute_strides[i] * vertex + comp * vertex_attribute_element_size[i];
+                        const u8* srcdata = Memory::GetPhysicalPointer(source_addr);
+
+                        if (g_debug_context && Pica::g_debug_context->recorder) {
+                            memory_accesses.AddAccess(source_addr,
+                                    (vertex_attribute_formats[i] == Regs::VertexAttributeFormat::FLOAT) ? 4
+                                    : (vertex_attribute_formats[i] == Regs::VertexAttributeFormat::SHORT) ? 2 : 1);
```
yuriks: Just add the whole vertex instead of using fine-grained detection like this?

src/video_core/command_processor.cpp
yuriks: (Even if there isn't, the logic can be greatly simplified by just keep tracking of a single min,max address range for the each loader and then merging them if they overlap. This handles the any combination of a single interleaved or many partially interleaved or dense buffers. This will be efficient since even indexed vertices already have locality inside their own buffers anyway.)

src/common/file_util.h
yuriks: Meh, nevermind.

src/core/tracer/citrace.h
yuriks: They're not, but neither is `#pragma pack` or that unaligned accesses should work, so it comes down to the platforms you're supporting.

yuriks:
Pass finished.

src/core/tracer/recorder.h
neobrain: @yuriks [citation needed] ?

src/core/tracer/citrace.h
neobrain: To make sure the sizes of the CTMemoryLoad, CTRegisterWrite, and CTStreamElement structs are equal.

src/core/tracer/recorder.h
neobrain: Could you suggest a better name?

src/core/tracer/citrace.h
neobrain: Why should the two things you mentioned (the latter which I'm not sure what it's referring to exactly) not work?

hrydgard: If things are naturally aligned, packing is indeed guaranteed and no pragma is necessary.

src/core/tracer/recorder.h
hrydgard: The type of a CRC32 result must be u32, otherwise it's not a CRC32.

neobrain: @hrydgard While unlikely, future boost versions could decide to change value_type to something like `std::array<uint8_t, 4>`.

I guess I could hardcode u32 in the header, and place a static_assert in the .cpp file to make sure the value_type indeed is u32, but why bother? recorder.h is a rarely included header, hence it's not like compile times are going to explode over including a (comparatively) dependency-lightweight boost header.

src/citra_qt/debugger/graphics_tracing.h
neobrain: Making them functions requires keeping around pointers to the respective buttons, which is bad style IMO.

neobrain: Also, it actually requires implementing three further functions instead of just calling connect once for each of them.

src/core/hw/gpu.cpp
neobrain: Because some program code should be executed even if address_start is zero, cf below.

src/core/tracer/recorder.cpp
neobrain: The point of inserting the dummy zero here is to make sure the checksum can be found via `.find()` in successive calls to AccessedMemory.

src/core/tracer/recorder.h
yuriks: They could, but they won't, because it wouldn't make any sense.

And it's never the case that doing something "just this once" will explode compilation times, but after you do something "just this once" several times it all adds up.

src/citra_qt/debugger/graphics_tracing.h
yuriks: Uh, no it does not require implementing three functions. It requires you to move three pointers variables from being function local to being class members and that's it. And it's not bad style, it's literally what the Qt UI designer does.

src/core/hw/gpu.cpp
yuriks: Which code? None of the code that was already there should be executed if the transfer isn't configured, and you didn't add any new code in this scope.

src/core/tracer/citrace.h
yuriks: But why is that important? The union below will already take the max size of all it's members.

src/core/hw/gpu.cpp
neobrain: Just read the code? I'm not sure where the issue is, really. if config.trigger and config.address_start are both true, all code is being enabled, and if only config.trigger is true, only part of it is enabled. That's what happens on hardware. No idea why they use such odd configurations as trigger=true and address_start=0.

src/core/tracer/recorder.cpp
yuriks: Ah, ok.

src/core/tracer/recorder.h
yuriks: Something like `CaptureMemoryRegion`, `RecordRegisterWrite` and `RecordFrameFinished`.

src/core/tracer/citrace.h
neobrain: I don't remember because it's been ages since I wrote this code, but I did have a good reason for it.

src/core/hw/gpu.cpp
yuriks: If no copy happened we should neither log (just turns into a misleading red herring) or signal a hw renderer flush. I don't think an interrupt is fired in that case either, from my earlier experiments (and @Subv also getting similar results.). Resetting the config register might be valid though.

neobrain: Fair enough. Resetting the config register is the only thing I actually verified explicitly.

neobrain:
Updated again. I'll merge this branch as soon as I have tested it once again to make sure it still works.

src/core/tracer/citrace.h
neobrain: FYI, I tried removing this and it completely broke citrace recordings. I guess this is just due to some member of the code breaking natural alignment and I'm open to suggestions as to where that member is, but for now I don't really feel like finding the culprit and will just keep the alignment specifier.

src/core/tracer/citrace.h
```diff
@@ -0,0 +1,101 @@
...
+    } size;
+
+    // TODO: Make it clearer which bits of this member are used for sizes other than 32 bits
+    uint64_t value;
+};
+
+struct CTStreamElement {
+    CTStreamElementType type;
+
+    union {
```
yuriks: Missing padding is before this union. The union has 64-bit alignment due to CTRegisterWrite, so you need to add a u32 of padding here.

PR #705, opened 2015-04-19
Modularize and refactor software rasterizer to allow use by other renderers.

tfarley:
Modularizes and refactors the renderer pipeline a little bit by moving vertex processing into its own function/file, and passing geometry to the renderer rather than processing the vertices directly in the command processor. Structured so that hardware renderers can be integrated later that can use the software rasterizer.


Subv:
LGTM

archshift:
LGTM

purpasmart96:
LGTM

src/video_core/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ set(SRCS
             primitive_assembly.cpp
             rasterizer.cpp
             utils.cpp
+            vertex_processor.cpp
```
neobrain: Please rename to "vertex_loader". It doesn't actually perform any vertex processing.

neobrain:
As for shuffling the code around, LGTM.

As for the renderer changes: First off, the current renderer class is crap, and we're better off starting a class from scratch. Re-using a class "just because it's there and the new functionality kinda fits" is rarely a good idea, and leads to unstructured mega-classes with no clear purpose.

More importantly: I'm not sure it's constructive to handle this via virtual methods at all. I was thinking it should be possible to optionally use a flexible combination of vertex loading, vertex shader processing, and primitive processing (clipping) of the software renderer for the hardware renderer and let the latter deal with rasterization, only. So there should be a total of four configurations possible for things which are covered by hw rendering, namely rast-only, rast+clipping, rast+clipping+shader, rast+clipping+shader+loading. We should probably rethink the particular division of stages, but that's the general idea. Virtual methods can only provide this by having a renderer class for each configuration, which will lead to combinatoric explosion once someone tries to add more "interesting" code paths.

Furthermore, virtual methods lead to sloppy design choices, where we really just end up with separate code that is sold under the same name rather than actually embedding the hw renderer into the current framework.

bunnei:
> As for the renderer changes: First off, the current renderer class is crap, and we're better off starting a class from scratch. Re-using a class "just because it's there and the new functionality kinda fits" is rarely a good idea, and leads to unstructured mega-classes with no clear purpose.

@neobrain so what exactly does your LGTM mean? Are you saying this PR can be merged, despite your aforementioned concerns?

I agree with your points to some extent, and that this should be well thought through now rather than later, but I'm not sure that throwing this redesign on @tfarley is the right way to go. You seem to have a vision for what the architecture of this will be. Asking someone else to implement it just based on a few comments/discussions is setting ourselves up for wasting a lot of time with that person totally missing the mark on the first few iterations (especially because @tfarley is new to the project).

Do you have time to provide a more thorough document for how you would like the different components of the render framework to be laid out, and how they might interconnect? Or, alternatively, could you propose some of the code refactors to support his work? I'm sure that you have many other things that you'd like to work on, but considering how invested you are in the design and implementation of the GPU code, I think you need to be willing/able to work with other people to see your vision through. There is enough interest/motivation for a HW renderer that it's going to happen - and if you want it done "right" by you, you're probably going to need to be more heavily involved (even if you're not the guy writing OpenGL or DirectX code).

Personally, I have no problem using an abstract base class with virtual methods, so long as it's done with reasonable deliberation. What you're proposing will likely need much more oversight. My concern here is that we might be (again) going down the path of repeatedly telling a new developer "you're doing this wrong, here is how _we_ would do it", without effective oversight, to the point where a new developer might get scared off, when its quite obvious that they have a lot of potential to contribute to the project.

neobrain:
@bunnei I discussed things with tfarley on IRC. The plan is to focus on the pixel processing part, and (for now) leave all vertex processing to the software renderer. This will make review easier (less code, less invasive changes, etc) and ultimately yield a better integrated OpenGL renderer. I'm sure that even the latter is in everyone's interest, since we don't want to end up with software-only debugging features (or well, software-only features in general).

That said, there are multiple issues with the particular interfaces introduced in this PR:
* it wasn't clear why the renderer abstraction was put at exactly this particular point
* it gave me the impression that it would lead to two completely distinct code paths between the sw and hw renderers
* I would've broken up this new interface soon enough anyway as part of the upcoming geometry shader support. This would have added a nontrivial workload for me, for something which isn't crucial at this point (since it's touching vertex processing, cf. first paragraph)

neobrain:
Closing as per reasons mentioned above. Things were discussed with @tfarley and he has a good understanding of how to proceed.

PR #707, opened 2015-04-22
Install a .desktop file to make citra-qt launchable from DE menus

linkmauve:
This makes Citra launchable from DE menus, applications, or by clicking on a 3DS file in a file manager.

For now those files are only installed on Linux, FreeBSD, OpenBSD and NetBSD, as I’m not sure which other platforms would use them, but this list can easily be extended.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,3 +193,16 @@ if(ENABLE_QT)
     add_subdirectory(externals/qhexedit)
 endif()
 add_subdirectory(src)
+
+# Install freedesktop.org metadata files, following those specifications:
+# http://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html
+# http://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html
+# http://standards.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux|FreeBSD|OpenBSD|NetBSD")
+    install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.desktop"
+            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/applications")
+    install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.svg"
+            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/pixmaps")
+    install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.xml"
+            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/mime/packages")
+endif()
--- a/dist/citra.svg
+++ b/dist/citra.svg
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright 2014 Citra Emulator Project
+    Licensed under GPLv2 or any later version
+    Refer to the license.txt file included.
+-->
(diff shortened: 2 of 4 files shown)


dist/citra.xml
```diff
@@ -0,0 +1,54 @@
...
+
+    <mime-type type="application/x-ctr-cia">
+        <comment>3DS importable archive</comment>
+        <comment xml:lang="fr">Archive importable 3DS</comment>
+        <acronym>CIA</acronym>
+        <expanded-acronym>CTR Importable Archive</expanded-acronym>
+        <glob pattern="*.cia"/>
+    </mime-type>
+
+    <mime-type type="application/x-ctr-smdh">
```
neobrain: What does this actually do? Does it give a hint to open .smdh files with Citra? Would that actually make any sense?

CMakeLists.txt
```diff
@@ -190,3 +190,13 @@ if(ENABLE_QT)
     add_subdirectory(externals/qhexedit)
 endif()
 add_subdirectory(src)
+
+# Install freedesktop.org files to make Citra user-friendly.
```
neobrain: "to make Citra user-friendly" is a pointless comment. Rather use something like "Install metadata files according to the <specname> specification"

dist/citra.svg
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
neobrain: I guess there should be some license information on this file?

dist/citra.xml
linkmauve: It only assignates a MIME type to SMDH files, without doing any application association (this is done in the .desktop file).
That way, DEs will tell “there is no program associated with ‘3DS icons’” instead of “with ‘unknown file type’”.

dist/citra.svg
linkmauve: I took and modified a bit the icon from our website: https://github.com/citra-emu/citra-emu.github.io/blob/master/resources/citra-icns-multicolor/citra.svg
If anything, its original author should set the license.

CMakeLists.txt
linkmauve: Fixed in the newer revision, with added links to the specifications.

dist/citra.svg
bunnei: The current icon was created by @archshift. I don't see any reason why this wouldn't hold the same license as the rest of the project?

bunnei:
After the license comment is addressed, LGTM

linkmauve:
In the latest version, I added the expanded acronym for SMDH, and fixed the case of the expanded CXI acronym.

dist/citra.svg
neobrain: @bunnei Upon some consideration, there indeed is no reason not to license this under the GPL, too. I was under the impression licensing artworks under the GPL is not actually possible, but after having read http://archive09.linux.com/feature/119212 it seems like it is (it brings up the concern that the concept of "source code" might not be well-defined, but that should be a non-issue for SVGs).

archshift: I'm fine licensing it under any version of the GPL or MIT.

Subv:
Can we do something similar for Windows?

linkmauve:
I think it involves some manipulation of the regedit, I’d rather not bloat this PR with that kind of black magic I don’t understand.

archshift:
LGTM

linkmauve:
Added our usual GPLv2+ header in the SVG file, and made executable/archive files display our icon by default (until a further PR adds thumbnail generation).

PR #713, opened 2015-04-28
Fix emulation state resetting to support multiple emulation sessions

bunnei:
This PR adds support for multiple emulation sessions (e.g. starting a game, stopping it, then starting another) by properly initializing/shutting down state throughout the emulator. To support this, I restructured how the emu thread was handled in the Qt frontend. The biggest change was moving ownership of the emu thread from the render window to the main window, and furthermore making the emu thread short-lived for just the emulation session.

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -10,6 +10,7 @@
 
 #include "common/common.h"
 #include "bootmanager.h"
+#include "main.h"
 
 #include "core/core.h"
 #include "core/settings.h"
@@ -30,6 +31,7 @@ EmuThread::EmuThread(GRenderWindow* render_window) :
     filename(""), exec_cpu_step(false), cpu_running(false),
     stop_run(false), render_window(render_window)
 {
+    connect(this, SIGNAL(started()), render_window, SLOT(moveContext()));
 }
 
 void EmuThread::SetFilename(std::string filename)
@@ -133,13 +135,9 @@ private:
     GRenderWindow* parent;
 };
 
-EmuThread& GRenderWindow::GetEmuThread()
-{
-    return emu_thread;
-}
+GRenderWindow::GRenderWindow(QWidget* parent, GMainWindow& main_window) :
+    QWidget(parent), main_window(main_window), keyboard_id(0) {
 
-GRenderWindow::GRenderWindow(QWidget* parent) : QWidget(parent), emu_thread(this), keyboard_id(0)
-{
     std::string window_title = Common::StringFromFormat("Citra | %s-%s", Common::g_scm_branch, Common::g_scm_desc);
     setWindowTitle(QString::fromStdString(window_title));
 
@@ -160,7 +158,6 @@ GRenderWindow::GRenderWindow(QWidget* parent) : QWidget(parent), emu_thread(this
     layout->addWidget(child);
     layout->setMargin(0);
     setLayout(layout);
-    connect(&emu_thread, SIGNAL(started()), this, SLOT(moveContext()));
 
     OnMinimalClientAreaChangeRequest(GetActiveConfig().min_client_area_size);
 
@@ -180,27 +177,15 @@ void GRenderWindow::moveContext()
     // We need to move GL context to the swapping thread in Qt5
 #if QT_VERSION > QT_VERSION_CHECK(5, 0, 0)
     // If the thread started running, move the GL Context to the new thread. Otherwise, move it back.
-    child->context()->moveToThread((QThread::currentThread() == qApp->thread()) ? &emu_thread : qApp->thread());
+    auto thread = QThread::currentThread() == qApp->thread() ? main_window.GetEmuThread() : qApp->thread();
+    child->context()->moveToThread(thread);
 #endif
-}
-
-GRenderWindow::~GRenderWindow()
-{
-    if (emu_thread.isRunning())
-        emu_thread.Stop();
 }
 
 void GRenderWindow::SwapBuffers()
 {
     // MakeCurrent is already called in renderer_opengl
     child->swapBuffers();
-}
-
-void GRenderWindow::closeEvent(QCloseEvent* event)
-{
-    if (emu_thread.isRunning())
-        emu_thread.Stop();
-    QWidget::closeEvent(event);
 }
 
 void GRenderWindow::MakeCurrent()
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -30,21 +30,19 @@
 EmuThread::EmuThread(GRenderWindow* render_window) :
     exec_cpu_step(false), cpu_running(false), stop_run(false), render_window(render_window) {
 
+    shutdown_event.Reset();
     connect(this, SIGNAL(started()), render_window, SLOT(moveContext()));
 }
 
-void EmuThread::run()
-{
+void EmuThread::run() {
     stop_run = false;
 
     // holds whether the cpu was running during the last iteration,
     // so that the DebugModeLeft signal can be emitted before the
     // next execution step
     bool was_active = false;
-    while (!stop_run)
-    {
-        if (cpu_running)
-        {
+    while (!stop_run) {
+        if (cpu_running) {
             if (!was_active)
                 emit DebugModeLeft();
 
@@ -53,9 +51,7 @@ void EmuThread::run()
             was_active = cpu_running || exec_cpu_step;
             if (!was_active)
                 emit DebugModeEntered();
-        }
-        else if (exec_cpu_step)
-        {
+        } else if (exec_cpu_step) {
             if (!was_active)
                 emit DebugModeLeft();
 
@@ -67,15 +63,14 @@ void EmuThread::run()
             was_active = false;
         }
     }
+
     render_window->moveContext();
 
-    Core::Stop();
+    shutdown_event.Set();
 }
 
-void EmuThread::Stop()
-{
-    if (!isRunning())
-    {
+void EmuThread::Stop() {
+    if (!isRunning()) {
         LOG_WARNING(Frontend, "EmuThread::Stop called while emu thread wasn't running, returning...");
         return;
     }
@@ -88,23 +83,19 @@ void EmuThread::Stop()
 
     // TODO: Waiting here is just a bad workaround for retarded shutdown logic.
     wait(1000);
-    if (isRunning())
-    {
+    if (isRunning()) {
         LOG_WARNING(Frontend, "EmuThread still running, terminating...");
         quit();
 
         // TODO: Waiting 50 seconds can be necessary if the logging subsystem has a lot of spam
         // queued... This should be fixed.
         wait(50000);
-        if (isRunning())
-        {
+        if (isRunning()) {
             LOG_CRITICAL(Frontend, "EmuThread STILL running, something is wrong here...");
             terminate();
         }
     }
     LOG_INFO(Frontend, "EmuThread stopped");
-
-    System::Shutdown();
 }
 
 
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -48,7 +48,7 @@ void EmuThread::run() {
             Core::RunLoop();
 
             was_active = running || exec_step;
-            if (!was_active)
+            if (!was_active && !stop_run)
                 emit DebugModeEntered();
         } else if (exec_step) {
             if (!was_active)
@@ -273,10 +273,10 @@ void GRenderWindow::OnMinimalClientAreaChangeRequest(const std::pair<unsigned,un
     setMinimumSize(minimal_size.first, minimal_size.second);
 }
 
-void GRenderWindow::OnEmulationStarted(EmuThread* emu_thread) {
+void GRenderWindow::OnEmulationStarting(EmuThread* emu_thread) {
     this->emu_thread = emu_thread;
 }
 
-void GRenderWindow::OnEmulationStopped() {
+void GRenderWindow::OnEmulationStopping() {
     emu_thread = nullptr;
 }
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -28,9 +28,8 @@
 #define COPYRIGHT       "Copyright (C) 2013-2014 Citra Team"
 
 EmuThread::EmuThread(GRenderWindow* render_window) :
-    exec_cpu_step(false), cpu_running(false), stop_run(false), render_window(render_window) {
+    exec_step(false), running(false), stop_run(false), render_window(render_window) {
 
-    shutdown_event.Reset();
     connect(this, SIGNAL(started()), render_window, SLOT(moveContext()));
 }
 
@@ -42,20 +41,20 @@ void EmuThread::run() {
     // next execution step
     bool was_active = false;
     while (!stop_run) {
-        if (cpu_running) {
+        if (running) {
             if (!was_active)
                 emit DebugModeLeft();
 
             Core::RunLoop();
 
-            was_active = cpu_running || exec_cpu_step;
+            was_active = running || exec_step;
             if (!was_active)
                 emit DebugModeEntered();
-        } else if (exec_cpu_step) {
+        } else if (exec_step) {
             if (!was_active)
                 emit DebugModeLeft();
 
-            exec_cpu_step = false;
+            exec_step = false;
             Core::SingleStep();
             emit DebugModeEntered();
             yieldCurrentThread();
@@ -65,39 +64,7 @@ void EmuThread::run() {
     }
 
     render_window->moveContext();
-
-    shutdown_event.Set();
 }
-
-void EmuThread::Stop() {
-    if (!isRunning()) {
-        LOG_WARNING(Frontend, "EmuThread::Stop called while emu thread wasn't running, returning...");
-        return;
-    }
-    stop_run = true;
-
-    // Release emu threads from any breakpoints, so that this doesn't hang forever.
-    Pica::g_debug_context->ClearBreakpoints();
-
-    //core::g_state = core::SYS_DIE;
-
-    // TODO: Waiting here is just a bad workaround for retarded shutdown logic.
-    wait(1000);
-    if (isRunning()) {
-        LOG_WARNING(Frontend, "EmuThread still running, terminating...");
-        quit();
-
-        // TODO: Waiting 50 seconds can be necessary if the logging subsystem has a lot of spam
-        // queued... This should be fixed.
-        wait(50000);
-        if (isRunning()) {
-            LOG_CRITICAL(Frontend, "EmuThread STILL running, something is wrong here...");
-            terminate();
-        }
-    }
-    LOG_INFO(Frontend, "EmuThread stopped");
-}
-
 
 // This class overrides paintEvent and resizeEvent to prevent the GUI thread from stealing GL context.
 // The corresponding functionality is handled in EmuThread instead
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -87,8 +87,8 @@ private:
     GRenderWindow* parent;
 };
 
-GRenderWindow::GRenderWindow(QWidget* parent, GMainWindow& main_window) :
-    QWidget(parent), main_window(main_window), keyboard_id(0) {
+GRenderWindow::GRenderWindow(QWidget* parent, EmuThread* emu_thread) :
+    QWidget(parent), emu_thread(emu_thread), keyboard_id(0) {
 
     std::string window_title = Common::StringFromFormat("Citra | %s-%s", Common::g_scm_branch, Common::g_scm_desc);
     setWindowTitle(QString::fromStdString(window_title));
@@ -129,7 +129,7 @@ void GRenderWindow::moveContext()
     // We need to move GL context to the swapping thread in Qt5
 #if QT_VERSION > QT_VERSION_CHECK(5, 0, 0)
     // If the thread started running, move the GL Context to the new thread. Otherwise, move it back.
-    auto thread = QThread::currentThread() == qApp->thread() ? main_window.GetEmuThread() : qApp->thread();
+    auto thread = (QThread::currentThread() == qApp->thread() && emu_thread != nullptr) ? emu_thread : qApp->thread();
     child->context()->moveToThread(thread);
 #endif
 }
@@ -271,4 +271,12 @@ void GRenderWindow::OnClientAreaResized(unsigned width, unsigned height)
 
 void GRenderWindow::OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>& minimal_size) {
     setMinimumSize(minimal_size.first, minimal_size.second);
+}
+
+void GRenderWindow::OnEmulationStarted(EmuThread* emu_thread) {
+    this->emu_thread = emu_thread;
+}
+
+void GRenderWindow::OnEmulationStopped() {
+    emu_thread = nullptr;
 }
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -28,15 +28,9 @@
 #define COPYRIGHT       "Copyright (C) 2013-2014 Citra Team"
 
 EmuThread::EmuThread(GRenderWindow* render_window) :
-    filename(""), exec_cpu_step(false), cpu_running(false),
-    stop_run(false), render_window(render_window)
-{
-    connect(this, SIGNAL(started()), render_window, SLOT(moveContext()));
-}
+    exec_cpu_step(false), cpu_running(false), stop_run(false), render_window(render_window) {
 
-void EmuThread::SetFilename(std::string filename)
-{
-    this->filename = filename;
+    connect(this, SIGNAL(started()), render_window, SLOT(moveContext()));
 }
 
 void EmuThread::run()
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -51,9 +51,9 @@ public:
     bool IsRunning() { return running; }
 
     /**
-     * Requests for the emulation thread to stop running and shutdown emulation
+     * Requests for the emulation thread to stop running
      */
-    void RequestShutdown() {
+    void RequestStop() {
         stop_run = true;
         running = false;
     };
@@ -115,8 +115,8 @@ public:
 public slots:
     void moveContext();  // overridden
 
-    void OnEmulationStarted(EmuThread* emu_thread);
-    void OnEmulationStopped();
+    void OnEmulationStarting(EmuThread* emu_thread);
+    void OnEmulationStopping();
 
 private:
     void OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>& minimal_size) override;
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -21,13 +21,6 @@ class EmuThread : public QThread
     Q_OBJECT
 
 public:
-    /**
-     * Set image filename
-     *
-     * @param filename
-     * @warning Only call when not running!
-     */
-    void SetFilename(std::string filename);
 
     /**
      * Start emulation (on new thread)
@@ -71,8 +64,6 @@ private:
     friend class GMainWindow;
 
     EmuThread(GRenderWindow* render_window);
-
-    std::string filename;
 
     bool exec_cpu_step;
     bool cpu_running;
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -25,65 +25,45 @@ public:
 
     /**
      * Start emulation (on new thread)
-     *
      * @warning Only call when not running!
      */
     void run() override;
 
     /**
-     * Allow the CPU to process a single instruction (if cpu is not running)
-     *
+     * Steps the emulation thread by a single CPU instruction (if the CPU is not already running)
      * @note This function is thread-safe
      */
-    void ExecStep() { exec_cpu_step = true; }
+    void ExecStep() { exec_step = true; }
 
     /**
-     * Sets whether the CPU is running 
-     *
+     * Sets whether the emulation thread is running or not
+     * @param running Boolean value, set the emulation thread to running if true
      * @note This function is thread-safe
      */
-    void SetCpuRunning(bool running) { cpu_running = running; }
+    void SetRunning(bool running) { this->running = running; }
 
     /**
-     * Allow the CPU to continue processing instructions without interruption
-     *
+     * Check if the emulation thread is running or not
+     * @return True if the emulation thread is running, otherwise false
      * @note This function is thread-safe
      */
-    bool IsCpuRunning() { return cpu_running; }
-
-
-    /**
-     * Shutdown (permantently stops) the CPU
-     */
-    void ShutdownCpu() { stop_run = true; };
+    bool IsRunning() { return running; }
 
     /**
-     * Waits for the CPU shutdown to complete
+     * Shutdown (permanently stops) the emulation thread
      */
-    void WaitForCpuShutdown() { shutdown_event.Wait(); }
-
-
-public slots:
-    /**
-     * Stop emulation and wait for the thread to finish.
-     *
-     * @details: This function will wait a second for the thread to finish; if it hasn't finished until then, we'll terminate() it and wait another second, hoping that it will be terminated by then.
-     * @note: This function is thread-safe.
-     */
-    void Stop();
+    void Shutdown() { stop_run = true; };
 
 private:
     friend class GMainWindow;
 
     EmuThread(GRenderWindow* render_window);
 
-    bool exec_cpu_step;
-    bool cpu_running;
+    bool exec_step;
+    bool running;
     std::atomic<bool> stop_run;
 
     GRenderWindow* render_window;
-
-    Common::Event shutdown_event;
 
 signals:
     /**
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -22,6 +22,7 @@ class EmuThread : public QThread
     Q_OBJECT
 
 public:
+    EmuThread(GRenderWindow* render_window);
 
     /**
      * Start emulation (on new thread)
@@ -50,15 +51,14 @@ public:
     bool IsRunning() { return running; }
 
     /**
-     * Shutdown (permanently stops) the emulation thread
+     * Requests for the emulation thread to stop running and shutdown emulation
      */
-    void Shutdown() { stop_run = true; };
+    void RequestShutdown() {
+        stop_run = true;
+        running = false;
+    };
 
 private:
-    friend class GMainWindow;
-
-    EmuThread(GRenderWindow* render_window);
-
     bool exec_step;
     bool running;
     std::atomic<bool> stop_run;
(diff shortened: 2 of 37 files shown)


src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -4001,7 +3983,11 @@ unsigned InterpreterMainLoop(ARMul_State* state) {
 
         phys_addr = cpu->Reg[15];
 
-        if (find_bb(cpu->Reg[15], ptr) == -1)
+        // Find the cached instruction cream, otherwise translate it...
+        auto itr = cpu->cream_cache.find(cpu->Reg[15]);
+        if (itr != cpu->cream_cache.end())
+            ptr = itr->second;
+        else
             if (InterpreterTranslate(cpu, ptr, cpu->Reg[15]) == FETCH_EXCEPTION)
```
lioncash: Can you brace this else section?

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -4001,9 +3983,14 @@ unsigned InterpreterMainLoop(ARMul_State* state) {
 
         phys_addr = cpu->Reg[15];
 
-        if (find_bb(cpu->Reg[15], ptr) == -1)
+        // Find the cached instruction cream, otherwise translate it...
+        auto itr = cpu->cream_cache.find(cpu->Reg[15]);
+        if (itr != cpu->cream_cache.end()) {
+            ptr = itr->second;
+        } else {
             if (InterpreterTranslate(cpu, ptr, cpu->Reg[15]) == FETCH_EXCEPTION)
```
archshift: `else if`?

src/citra_qt/bootmanager.cpp
```diff
@@ -27,28 +28,21 @@
 #define COPYRIGHT       "Copyright (C) 2013-2014 Citra Team"
```
Subv: Not related to your PR, but we should probably change the copyright year in this define already

src/citra_qt/main.cpp
```diff
@@ -328,7 +343,7 @@ int __cdecl main(int argc, char* argv[])
 
     GMainWindow main_window;
     // After settings have been loaded by GMainWindow, apply the filter
-    log_filter.ParseFilterString(Settings::values.log_filter);
+    log_filter.ParseFilterString("*:Warning");
```
Subv: Huh?

bunnei: Derp... commit error

src/citra_qt/bootmanager.cpp
```diff
@@ -180,14 +162,13 @@ void GRenderWindow::moveContext()
     // We need to move GL context to the swapping thread in Qt5
 #if QT_VERSION > QT_VERSION_CHECK(5, 0, 0)
     // If the thread started running, move the GL Context to the new thread. Otherwise, move it back.
-    child->context()->moveToThread((QThread::currentThread() == qApp->thread()) ? &emu_thread : qApp->thread());
+    auto thread = QThread::currentThread() == qApp->thread() ? main_window.GetEmuThread() : qApp->thread();
+    child->context()->moveToThread(thread);
 #endif
 }
 
 GRenderWindow::~GRenderWindow()
```
yuriks: If the destructor is empty, just remove it.

src/citra_qt/bootmanager.cpp
```diff
@@ -198,8 +179,6 @@ void GRenderWindow::SwapBuffers()
 
 void GRenderWindow::closeEvent(QCloseEvent* event)
```
yuriks: Since there's no custom action taken in thi handler anymore, simply remove it.

src/citra_qt/bootmanager.h
```diff
@@ -43,20 +38,31 @@ class EmuThread : public QThread
...
      * @note This function is thread-safe
      */
     void SetCpuRunning(bool running) { cpu_running = running; }
 
     /**
-    * Allow the CPU to continue processing instructions without interruption
-    *
-    * @note This function is thread-safe
-    */
+     * Allow the CPU to continue processing instructions without interruption
```
yuriks: Dat gibberish doc-comment

src/citra_qt/main.cpp
```diff
@@ -210,13 +214,32 @@ void GMainWindow::BootGame(std::string filename)
...
     render_window->show();
     OnStartGame();
 }
 
+void GMainWindow::ShutdownGame() {
+    emu_thread->SetCpuRunning(false);
+
+    emu_thread->ShutdownCpu();
+    emu_thread->WaitForCpuShutdown();
+    emu_thread->Stop();
```
yuriks: Since you call `WaitForCpuShutdown` right before this, I believe the `Stop` method is now redundant, since this is the only place where it's still called (I think?) and by then the thread will be stopped already anyway.

src/citra_qt/main.cpp
```diff
@@ -210,13 +214,32 @@ void GMainWindow::BootGame(std::string filename)
...
 
     render_window->show();
     OnStartGame();
 }
 
+void GMainWindow::ShutdownGame() {
+    emu_thread->SetCpuRunning(false);
+
+    emu_thread->ShutdownCpu();
+    emu_thread->WaitForCpuShutdown();
```
yuriks: Since this thread is signalled at the same time as the thread function exits, the correct way to do it is to forego the even and simply join on the thread instead using [`wait`](http://doc.qt.io/qt-4.8/qthread.html#wait).

Another way of handling this termination is to simply subscribe to the `finished` signal in the QThread and then run the rest of the cleanup that's in this function in that signal handler. This should in theory improve UI fluidity (since the UI thread won't block) but probably doesn't matter. (In fact, maybe `QThread::wait` already pumps the event loop while waiting, I wouldn't know.)

src/citra_qt/main.cpp
```diff
@@ -265,24 +282,22 @@ void GMainWindow::OnOpenHotkeysDialog()
...
+void GMainWindow::ToggleWindowMode() {
+    // Render in the main window...
+    if (ui.action_Single_Window_Mode->isChecked()) {
         render_window->BackupGeometry();
         ui.horizontalLayout->addWidget(render_window);
         render_window->setVisible(true);
         render_window->setFocusPolicy(Qt::ClickFocus);
         render_window->setFocus();
+
+    // Render in a separate window...
```
yuriks: Put this comment *inside* the braced block?

src/citra_qt/main.cpp
```diff
@@ -265,24 +282,22 @@ void GMainWindow::OnOpenHotkeysDialog()
...
-        ui.horizontalLayout->removeWidget(render_window);
-        render_window->setParent(nullptr);
-        render_window->setVisible(true);
-        render_window->RestoreGeometry();
-        render_window->setFocusPolicy(Qt::NoFocus);
-    }
-    else if (enable && render_window->parent() == nullptr)
-    {
+void GMainWindow::ToggleWindowMode() {
+    // Render in the main window...
```
yuriks: Likewise, if you change the other.

src/citra_qt/main.cpp
neobrain: +1

src/core/hle/service/apt/apt.cpp
```diff
@@ -317,7 +318,6 @@ void Init() {
 }
 
 void Shutdown() {
```
yuriks: You should zero the corresponding pointers on shutdown, otherwise we'll leave a bunch of dead objects hanging around until the next time the emu is initialized. It's probably not a big deal, but it's sloppy code imo. (Also makes it harder to find initialization bugs.)

src/core/mem_map.cpp
```diff
@@ -11,30 +11,30 @@
...
-static u8* physical_system_mem  = nullptr;   ///< System physical memory
-static u8* physical_fcram       = nullptr;   ///< Main physical memory (FCRAM)
-static u8* physical_heap_gsp    = nullptr;   ///< GSP heap physical memory
-static u8* physical_vram        = nullptr;   ///< Video physical memory (VRAM)
-static u8* physical_shared_mem  = nullptr;   ///< Physical shared memory
-static u8* physical_dsp_mem     = nullptr;   ///< Physical DSP memory
-static u8* physical_kernel_mem;              ///< Kernel memory
+u8* g_base;                     ///< The base pointer to the auto-mirrored arena.
+
+static MemArena* arena;         ///< The MemArena class
```
yuriks: `unique_ptr` pls. (Alternatively, isn't there a way to simply clear out the MemArena?)

src/core/hle/service/apt/apt.cpp
neobrain: Instead of trying to enforce the rule of explicitly resetting variables in Shutdown functions, I think we should strive to introduce a better system for handling initialization. Explicit uninitialization is annoying, ugly, and hard to enforce consistently. It's basically just an ad-hoc solution for the lack of an actual initialization system.

src/citra_qt/bootmanager.h
```diff
@@ -43,20 +38,31 @@ class EmuThread : public QThread
     void ExecStep() { exec_cpu_step = true; }
 
     /**
-     * Allow the CPU to continue processing instructions without interruption
+     * Sets whether the CPU is running
```
neobrain: What?

src/citra_qt/bootmanager.h
neobrain: What does that doc-comment try to say? What kind of interruption are you talking about? What does "allow to" specifically mean?

src/citra_qt/bootmanager.h
```diff
@@ -43,20 +38,31 @@ class EmuThread : public QThread
...
-    */
+     * Allow the CPU to continue processing instructions without interruption
+     *
+     * @note This function is thread-safe
+     */
     bool IsCpuRunning() { return cpu_running; }
 
 
+    /**
+     * Shutdown (permantently stops) the CPU
```
neobrain: spelling mistake. rewrite this comment plz

src/citra_qt/bootmanager.h
```diff
@@ -43,20 +38,31 @@ class EmuThread : public QThread
...
+     */
+    void ShutdownCpu() { stop_run = true; };
+
+    /**
+     * Waits for the CPU shutdown to complete
```
neobrain: rewrite comment plz

neobrain: In particular, this docstring sounds like something else is expected to be called before WaitForCpuShutdown(). Should hence mention what that is.

src/citra_qt/bootmanager.h
```diff
@@ -43,20 +38,31 @@ class EmuThread : public QThread
...
+     *
+     * @note This function is thread-safe
+     */
     bool IsCpuRunning() { return cpu_running; }
 
 
+    /**
+     * Shutdown (permantently stops) the CPU
+     */
+    void ShutdownCpu() { stop_run = true; };
```
neobrain: Why does this specifically talk about the "CPU", suggesting that other parts of the hardware keep running?

src/citra_qt/bootmanager.h
```diff
@@ -139,10 +143,10 @@ public slots:
 
     QGLWidget* child;
 
-    EmuThread emu_thread;
-
     QByteArray geometry;
 
+    GMainWindow& main_window;
```
neobrain: Should be a pointer as per common Qt style. Also please consider adding one or two const qualifiers.

src/citra_qt/debugger/disassembler.cpp
```diff
@@ -209,13 +211,13 @@ void DisassemblerWidget::OnStep()
 
 void DisassemblerWidget::OnStepInto()
 {
-    emu_thread.SetCpuRunning(false);
-    emu_thread.ExecStep();
+    main_window.GetEmuThread()->SetCpuRunning(false);
```
neobrain: You're only using `main_window`'s `GetEmuThread` member function here, and the whole widget expects the returned EmuThread to never change throughout the widget's existence. Should just store a pointer to the EmuThread instead, hence (and also because that's cleaner than just calling GetEmuThread() all the time).

neobrain: Actually, looks like the return value of GetEmuThread may change between invocations. Whenever that happens, DisassemblerWidget should be notified about it (and reset its own state).

src/citra_qt/main.h
```diff
@@ -54,6 +60,7 @@ private slots:
     Ui::MainWindow ui;
 
     GRenderWindow* render_window;
+    EmuThread* emu_thread;
```
neobrain: Should be a unique_ptr since it's a manually managed pointer.

src/citra_qt/main.cpp
```diff
@@ -210,13 +214,32 @@ void GMainWindow::BootGame(std::string filename)
...
 }
 
+void GMainWindow::ShutdownGame() {
+    emu_thread->SetCpuRunning(false);
+
+    emu_thread->ShutdownCpu();
+    emu_thread->WaitForCpuShutdown();
+    emu_thread->Stop();
+
+    delete emu_thread;
```
neobrain: just fyi, I'd say whenever you're performing a delete followed by a re-initialization (or just a nullptr assignment), you should use `unique_ptr` instead.

src/citra_qt/main.cpp
```diff
@@ -210,13 +214,32 @@ void GMainWindow::BootGame(std::string filename)
...
+    emu_thread->start();
 
     render_window->show();
     OnStartGame();
 }
 
+void GMainWindow::ShutdownGame() {
+    emu_thread->SetCpuRunning(false);
+
+    emu_thread->ShutdownCpu();
```
neobrain: Is there any reason not to merge WaitForCpuShutdown and ShutdownCpu instead?

src/core/arm/skyeye_common/armdefs.h
```diff
@@ -152,6 +154,10 @@ So, if lateabtSig=1, then it means Late Abort Model(Base Updated Abort Model)
 
     // Added by ksh in 2005-10-1
     cpu_config_t* cpu;
+
+    // TODO(bunnei): Move this cache to a better place - it should be per codeset (likely per
+    // process for our purposes), not per ARMul_State (which tracks CPU core state).
+    std::unordered_map<u32, int> cream_cache;
```
neobrain: What is "cream cache" supposed to tell me? Is there a better name?

neobrain: Also, should mention what the key/value pair represents.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -11,7 +11,7 @@
 
 namespace DSP_DSP {
 
-static u32 read_pipe_count    = 0;
+static u32 read_pipe_count;
```
neobrain: The old initial value doesn't seem to be assigned anymore at all - is this intended?

neobrain:
The straightforward "Introduce Initialize/Shutdown functions everywhere" approach works and gets the job done for now, but this will get us into big trouble in the long run. I'd suggest figuring out a better system for state management ASAP. Possibly it should be considered a blocker for introducing a state saving/loading framework, because state restoration is essentially the same functionality as module initialization.

Other than summoning the apocalypse, this LGTM modulo the comments I added.

src/citra_qt/main.h
yuriks: I thought of suggesting this, but it goes at odds with the Qt style, which encourages manual pointers that use Qt's ownership model to be automatically destroyed when their parent is deleted. `EmuThread` itself is a `QObject` so I made a mental note to check this and then forgot...

What I'm saying is +1, I guess xD

yuriks:
Re. another approach: The only sensible alternative I can think of is wrapping everything in structs, which is certainly a much larger code change in several respects. To avoid having that expose all internal variables of each module in the header file, this struct can be private to the module, and the Init/Shutdown routines would simply manage a global instances of it.

src/citra_qt/bootmanager.h
bunnei: @neobrain lol I'm guessing you should ask yourself, I just fixed the docstring indentation :) I'll fix it anyway

src/citra_qt/debugger/disassembler.cpp
bunnei: Right, the emu thread will change between emulation sessions. Please elaborate on what the more elegant solution with Qt would be?

neobrain: I suppose you'll want to add `EmuThreadDestroyed()`, `EmuThreadCreated(EmuThread*)` slots in the widget, add corresponding signals in the main window class, then connect the two in the render window and emit the signals whenever something changes about the EmuThread.

NOTE: Not sure if "created" and "destroyed" are the proper notions we want to report here. Maybe started+stopped are better ones. I'll leave that up to you ;)

src/core/arm/skyeye_common/armdefs.h
Subv: instruction_cache sounds like a better name

src/citra_qt/bootmanager.h
```diff
@@ -9,72 +9,58 @@
...
-public slots:
     /**
-     * Stop emulation and wait for the thread to finish.
-     *
-     * @details: This function will wait a second for the thread to finish; if it hasn't finished until then, we'll terminate() it and wait another second, hoping that it will be terminated by then.
-     * @note: This function is thread-safe.
+     * Shutdown (permanently stops) the emulation thread
      */
-    void Stop();
+    void Shutdown() { stop_run = true; };
```
yuriks: I think this could automatically set `running` to false, to avoid needing to also call `SetRunning`. (Which afaik would be necessary in any instance this function is used.)

src/citra_qt/main.cpp
```diff
@@ -196,32 +202,61 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
+    // Shutdown the emulation thread and wait for it to complete
+    emit EmuThreadStopped();
+    emu_thread->SetRunning(false);
+    emu_thread->Shutdown();
+    emu_thread->wait();
+    delete emu_thread;
+    emu_thread = nullptr;
+
+    // Release emu threads from any breakpoints
+    Pica::g_debug_context->ClearBreakpoints();
```
yuriks: This needs to go after `Shutdown()` but before `wait()`: The issue adding this originally fixed is that if the emulation stops on a GPU breakpoint after (or before) `Shutdown` is called, the emulation would never be able to continue out to the main loop and terminate. Thus `wait()` would hang forever.

src/citra_qt/bootmanager.h
yuriks: Also, I think `RequestShutdown` is a better name.

src/citra_qt/bootmanager.h
```diff
@@ -134,15 +115,18 @@ class GRenderWindow : public QWidget, public EmuWindow
 public slots:
     void moveContext();  // overridden
 
+    void OnEmuThreadStarted(EmuThread* emu_thread);
+    void OnEmuThreadStopped();
```
yuriks: Can you use more "generally applicable" names for this like "OnEmulationStopped" so that they don't become misnomers in the future?

src/citra_qt/main.cpp
```diff
@@ -196,32 +202,61 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
+void GMainWindow::BootGame(std::string filename) {
     LOG_INFO(Frontend, "Citra starting...\n");
+
+    // Initialize the core emulation
     System::Init(render_window);
 
-    // Load a game or die...
+    // Load the game
     if (Loader::ResultStatus::Success != Loader::LoadFile(filename)) {
         LOG_CRITICAL(Frontend, "Failed to load ROM!");
```
yuriks: I have a suspicion this error should actually be aborting startup lol

src/citra_qt/main.cpp
```diff
@@ -196,32 +202,61 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
-    render_window->GetEmuThread().start();
-
     render_window->show();
+
     OnStartGame();
 }
 
+void GMainWindow::ShutdownGame() {
+    // Shutdown the emulation thread and wait for it to complete
+    emit EmuThreadStopped();
```
yuriks: Doesn't really make a difference, but this can be called later, just before deleting the thread.

src/core/hle/service/dsp_dsp.cpp
bunnei: It's set on boot, https://github.com/bunnei/citra/blob/qt-emuthread-fixes/src/core/hle/service/dsp_dsp.cpp#L292

bunnei:
Ready for another pass

yuriks:
LQGTM

src/citra_qt/debugger/disassembler.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include "disassembler.h"
 
+#include "../main.h"
```
neobrain: This doesn't need main.h anymore, does it?

src/citra_qt/debugger/disassembler.cpp
```diff
@@ -199,7 +193,10 @@ void DisassemblerWidget::Init()
 
 void DisassemblerWidget::OnContinue()
 {
-    emu_thread.SetCpuRunning(true);
+    if (emu_thread == nullptr)
```
neobrain: Instead of nullptr checks, disable the widget when the emu_thread is set to `nullptr` using something like `setDisabled()`.

src/citra_qt/debugger/disassembler.cpp
```diff
@@ -255,3 +264,22 @@ int DisassemblerWidget::SelectedRow()
...
+    connect(disasm_ui.button_breakpoint, SIGNAL(clicked()), model, SLOT(OnSetOrUnsetBreakpoint()));
+    connect(GetHotkey("Disassembler", "Set Breakpoint", this), SIGNAL(activated()), model, SLOT(OnSetOrUnsetBreakpoint()));
+
+    Init();
+}
+
+void DisassemblerWidget::OnEmulationStopped() {
+    delete model;
+    emu_thread = nullptr;
+}
```
neobrain: Should additionally clear the widget contents using something like `disasm_ui.treeView.clear()`

src/citra_qt/debugger/disassembler.h
```diff
@@ -13,6 +13,7 @@
 #include "common/break_points.h"
 
 class QAction;
+class GMainWindow;
```
neobrain: not necessary anymore

src/citra_qt/debugger/registers.cpp
```diff
@@ -41,10 +39,12 @@ RegistersWidget::RegistersWidget(QWidget* parent) : QDockWidget(parent)
     CSPR->addChild(new QTreeWidgetItem(QStringList("N")));
 }
 
-void RegistersWidget::OnDebugModeEntered()
-{
+void RegistersWidget::OnDebugModeEntered() {
     ARM_Interface* app_core = Core::g_app_core;
 
+    if (app_core == nullptr)
```
neobrain: Again, instead of nullptr checks, disable the widget entirely if emulation is not running.

neobrain: Actually, this works just as well and some sort of check is necessary to not access invalid memory here.

However, the widget should still be disabled on emulation start/stop.

src/citra_qt/main.cpp
```diff
@@ -196,32 +195,84 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
     if (Loader::ResultStatus::Success != Loader::LoadFile(filename)) {
         LOG_CRITICAL(Frontend, "Failed to load ROM!");
+        System::Shutdown();
+        return;
     }
 
-    disasmWidget->Init();
+    // Create and start the emulation thread
+    emu_thread = Common::make_unique<EmuThread>(render_window);
+    emit EmulationStarted(emu_thread.get());
```
neobrain: Confusing name: Why would you be signalizing that emulation has started *before* you actually start it? Maybe instead call it something like EmulatorReady? Do we actually have all relevant state initialized up to this point, or will emu_thread.start further initialize other state? If the latter is the case, that additional initialization should be moved to before EmulationStarted, to prevent incorrect assumptions about memory validity.

src/citra_qt/main.cpp
```diff
@@ -196,32 +195,84 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
+
     OnStartGame();
 }
 
+void GMainWindow::ShutdownGame() {
+    // Shutdown the emulation thread
+    emu_thread->RequestShutdown();
+
+    // Disconnect signals that are attached to the current emulation thread
+    disconnect(emu_thread.get(), SIGNAL(DebugModeEntered()), disasmWidget, SLOT(OnDebugModeEntered()));
```
neobrain: You don't need (read: shouldn't need) to explicitly disconnect signals/slots. That's done when the `emu_thread` object is being deleted. We should wait for the shutdown to complete here though, to make sure no race conditions are triggered.

src/citra_qt/main.cpp
```diff
@@ -196,32 +195,84 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
+    disconnect(emu_thread.get(), SIGNAL(DebugModeEntered()), callstackWidget, SLOT(OnDebugModeEntered()));
+    disconnect(emu_thread.get(), SIGNAL(DebugModeLeft()), disasmWidget, SLOT(OnDebugModeLeft()));
+    disconnect(emu_thread.get(), SIGNAL(DebugModeLeft()), registersWidget, SLOT(OnDebugModeLeft()));
+    disconnect(emu_thread.get(), SIGNAL(DebugModeLeft()), callstackWidget, SLOT(OnDebugModeLeft()));
+
+    // Release emu threads from any breakpoints
+    // This belongs after RequestShutdown() and before wait() because if emulation stops on a GPU
+    // breakpoint after (or before) RequestShutdown() is called, the emulation would never be able
+    // to continue out to the main loop and terminate. Thus wait() would hang forever.
+    Pica::g_debug_context->ClearBreakpoints();
```
neobrain: This function is not thread-safe and may only be called when we are at a GPU breakpoint. Right now, I'm not sure how to resolve this properly.

src/citra_qt/main.cpp
```diff
@@ -196,32 +195,84 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
+
+    emit EmulationStopped();
```
neobrain: I suppose this signals that emulator resources have become invalid, since it's called after shutdown. It should instead signal when emulator resources are *about to become* invalid, though. Hence it should be called before shutdown, and renamed to "EmulatorStopping" or something like that.

src/citra_qt/main.cpp
```diff
@@ -196,32 +195,84 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
+
+    // Wait for emulation thread to complete and delete it
+    emu_thread->wait();
```
neobrain: What if the emulation thread is stuck in an infinite loop? can we do anything about that situation, i.e. possibly spawn some "emu thread not responding, kill?" message after waiting for N seconds.

src/citra_qt/main.cpp
```diff
@@ -196,32 +195,84 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
-    render_window->GetEmuThread().SetFilename(filename);
-    render_window->GetEmuThread().start();
-
     render_window->show();
+
     OnStartGame();
 }
 
+void GMainWindow::ShutdownGame() {
+    // Shutdown the emulation thread
```
neobrain: err, misleading comment - this does not shutdown anything, it *requests* it. As such, having a comment here is completely pointless anyway. Don't write comments for code which is already self-documenting.

src/citra_qt/bootmanager.h
```diff
@@ -9,72 +9,58 @@
...
-public slots:
     /**
-     * Stop emulation and wait for the thread to finish.
-     *
-     * @details: This function will wait a second for the thread to finish; if it hasn't finished until then, we'll terminate() it and wait another second, hoping that it will be terminated by then.
-     * @note: This function is thread-safe.
+     * Requests for the emulation thread to stop running and shutdown emulation
      */
-    void Stop();
+    void RequestShutdown() {
```
neobrain: Given that you still need to call `System::Shutdown` manually afterwards, this sounds like a misnomer. If it doesn't actually request the shutdown to be performed, what does it actually do?

src/citra_qt/main.cpp
```diff
@@ -196,32 +195,84 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
+}
+
 void GMainWindow::OnMenuLoadFile()
 {
     QString filename = QFileDialog::getOpenFileName(this, tr("Load File"), QString(), tr("3DS executable (*.3ds *.3dsx *.elf *.axf *.bin *.cci *.cxi)"));
-    if (filename.size())
-       BootGame(filename.toLatin1().data());
+    if (filename.size()) {
+        // Shutdown previous session if the emu thread is still active...
+        if (emu_thread != nullptr)
```
neobrain: Nah, please don't do this. This function should never be callable if a previous session is running. If you really want to make sure, make this an assert instead.

What should instead happen is that the menu option should be greyed out if emulation is running.

src/citra_qt/main.h
```diff
@@ -55,6 +62,8 @@ private slots:
 
     GRenderWindow* render_window;
 
+    std::unique_ptr<EmuThread> emu_thread;
```
neobrain: This should actually be a `shared_ptr`, since references to it are stored in multiple places, sorry!

src/core/hle/service/hid/hid.cpp
```diff
@@ -175,6 +175,12 @@ void Init() {
 }
 
 void Shutdown() {
+    shared_mem = nullptr;
```
neobrain: My personal preference for resetting `shared_ptr`s is to use `.reset()` to make more explicit why you are explicitly reinitializing these variables here.

neobrain:
Qt changes look mostly fine, however I'd like to give this a stress test tomorrow to make sure all of it works like expected.

src/citra_qt/debugger/disassembler.cpp
yuriks: Just realized this deletes model while it's being used by the tree view. Does anyone know if the model unsets itself from it's parent when it's destructed? I know Qt object auto de-parent themselves, but not sure if it would unset the model in this case. To be safe you should first set the model to null before deleting it.

src/citra_qt/main.cpp
yuriks: `emu_thread->start()` only starts the emulation thread. All initialization is done in `System::Init` above. I think the name is fine, but if we rename it then I think it should be `EmulationStarting` which is the usual name I've seen for these "state transition" kinds of events.

src/citra_qt/main.cpp
yuriks: We were already calling it an unsafe manner before then, it only moved from `Stop` to here, but the conditions in which it's called remain the same.

src/citra_qt/main.cpp
yuriks: This is the same situation we had before with those delays in `Stop`, and as we've seen, it's messy and doesn't work anyway. Forcefully terminating the thread often just causes a random crash or deadlock later on.

If the emulation thread is stuck in a loop then it's a bug which we should address in some other way.

src/citra_qt/main.cpp
yuriks: I can't think of a single application or emulator that requires you to explicitly choose a "Close" menu option before letting you load a new file/game. This should totally work.

src/citra_qt/main.h
yuriks: If possible (and it looks like it we're managing it) it's better to keep the lifetime of the object explicit by making it a unique_ptr. I don't think we need a shared_ptr in this scenario.

src/citra_qt/main.cpp
bunnei: Ditto. It works fine, and it's specifically to prevent the need of "Stopping" the emulation session before starting a new session.

src/core/hle/service/hid/hid.cpp
bunnei: I'm following the convention of how we reset unique_ptr variables (after they've been freed) throughout the emulator. I don't see the point in changing just the ones I touched with this PR, and I think it's out of scope to change all of them.

src/citra_qt/debugger/disassembler.cpp
bunnei: There's no point to deleting the contents... We recreate the model each emulation session

src/citra_qt/main.cpp
bunnei: Going to leave this here for now...

src/citra_qt/main.cpp
bunnei: It is called before shutdown, so what you are suggesting is actually how it works. I think you're getting caught up on naming. `emu_thread->RequestShutdown()` just requests that the emulation thread be stopped. Shutdown does not happen until after this signal, with `System::Shutdown()`. I'll rename `RequestShutdown` to `RequestStop` and `EmulationStopped` to `EmulationStopping`.

src/citra_qt/main.h
bunnei: Keeping as `unique_ptr` per @yuriks suggestion.

src/citra_qt/debugger/registers.cpp
bunnei: Addressed

bunnei:
All additional comments addressed

src/citra_qt/main.cpp
neobrain: meh, fair enough, even though trying to load an application while one is already running seems like the most borderline scenario ever to me.

src/core/hle/service/hid/hid.cpp
neobrain: Alright, it doesn't matter too much in this particular instance anyway.

src/citra_qt/debugger/registers.cpp
neobrain: The nullptr check is still there...

neobrain:
When closing citra while an application is being emulated, I'm getting a log message `QThread: Destroyed while thread is still running`. This does not happen when I just stop emulation without closing citra. Are we still missing some cleanup in the former case?

src/citra_qt/main.cpp
```diff
@@ -196,32 +195,75 @@ void GMainWindow::OnDisplayTitleBars(bool show)
...
 }
 
+void GMainWindow::ShutdownGame() {
+    emu_thread->RequestStop();
+
+    // Release emu threads from any breakpoints
+    // This belongs after RequestStop() and before wait() because if emulation stops on a GPU
+    // breakpoint after (or before) RequestStop() is called, the emulation would never be able
+    // to continue out to the main loop and terminate. Thus wait() would hang forever.
+    Pica::g_debug_context->ClearBreakpoints();
```
neobrain: Should have a TODO mentioning that this not thread-safe but being used as if it was.

src/citra_qt/main.h
```diff
@@ -34,8 +36,13 @@ class GMainWindow : public QMainWindow
     GMainWindow();
     ~GMainWindow();
 
+signals:
+    void EmulationStarting(EmuThread* emu_thread);
```
neobrain: These two should have actual documentation on what you mean with "Starting" and "Stopping", in particular with regards to resource readyness/validity/etc.

neobrain:
LGTM.

PR #716, opened 2015-04-28
Services: Stubbed many low level services

purpasmart96:
I also made the PTM services have the same format as the  HID, APT, IR services as well.
These low level services are used by the the by the core service modules, which can  be dumped from the 3ds and can be run as ncch's. Not useful to games, but useful for getting low level service command headers and stuff that utilized in the future.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -60,12 +60,32 @@ set(SRCS
             hle/service/hid/hid_spvr.cpp
             hle/service/gsp_lcd.cpp
             hle/service/http_c.cpp
+            hle/service/i2c/i2c.cpp
+            hle/service/i2c/i2c_cam.cpp
+            hle/service/i2c/i2c_deb.cpp
+            hle/service/i2c/i2c_eep.cpp
+            hle/service/i2c/i2c_hid.cpp
+            hle/service/i2c/i2c_ir.cpp
+            hle/service/i2c/i2c_lcd.cpp
+            hle/service/i2c/i2c_mcu.cpp
+            hle/service/i2c/i2c_nfc.cpp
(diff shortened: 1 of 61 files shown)


neobrain:
-1 on stubbing even more things:

* It just pollutes the code base with unused stuff, which likely will not be implemented for another while
* It adds significant overhead to potential restructuring works of service interfaces
* It increases the risk of copy-pasting over incorrect info from 3dbrew, and it's likely that any fixes made to 3dbrew in the future might be "forgotten" to be ported over to Citra, effectively requiring a full review of the existing dummy functions as soon as someone actually wants to seriously tackle an implementation.

purpasmart96:
@neobrain 
"It increases the risk of copy-pasting over incorrect info from 3dbrew, and it's likely that any fixes made to 3dbrew in the future might be "forgotten" to be ported over to Citra, effectively requiring a full review of the existing dummy functions as soon as someone actually wants to seriously tackle an implementation."
Most of these services were from applications that called these services, the applications in question were not liking the fact that we did Not have these service ports. They weren't just copied and pasted from 3dbrew is that's what your saying.

yuriks:
I have my doubts that games are really directly using `i2c` and `mcu` services, but I'd love some logs proving me wrong.

purpasmart96:
@yuriks They are not used by games, BUT they are used by the core service modules, which are ncch's which can be run on citra, and I currently have them in my possession, in which they do call these services.

yuriks:
We're not about to turn Citra into an LLE emulator (which is what this means) so I'd say that running those is out of scope for now.

purpasmart96:
Alright... I guess it was kinda premature, but I also did since we could then get their command headers. since there is like none in 3dbrew.

PR #722, opened 2015-05-06
[RFC] A binary translator for citra

Syrco:
Asking for comments on how to get this integrated


src/binary_translation/BlockColors.h
```diff
@@ -0,0 +1,50 @@
...
+	BlockColors(ModuleGen *module);
+	~BlockColors();
+
+	void AddBlock(InstructionBlock *block);
+	// Generates a function for each color
+	void GenerateFunctions();
+
+	llvm::FunctionType *GetFunctionType() { return function_type; }
+	size_t GetColorCount() { return colors.size(); }
+	size_t GetColorInstructionCount(size_t color) { return colors[color].instructions.size(); }
```
lioncash: GetColorInstructionCount and GetColorCount can both be const.

src/binary_translation/CodeGen.h
```diff
@@ -0,0 +1,37 @@
...
+ * Holds alls the basic llvm structures
+ */
+class CodeGen
+{
+public:
+	CodeGen(const char *output_object_filename, const char *output_debug_filename, bool verify);
+	~CodeGen();
+
+    void Run();
+    void IntializeLLVM();
```
lioncash: InitializeLLVM

src/binary_translation/InstructionBlock.cpp
```diff
@@ -0,0 +1,60 @@
+#include "InstructionBlock.h"
+#include "ModuleGen.h"
+#include "Instructions/Instruction.h"
+#include <sstream>
+#include <iomanip>
+#include "MachineState.h"
+
```
lioncash: Standard library includes first. Preferably in alphabetical order.

src/binary_translation/InstructionBlock.h
```diff
@@ -0,0 +1,85 @@
...
+
+    u32 Address();
+    ModuleGen *Module() { return module; }
+    llvm::IRBuilder<> *IrBuilder() { return module->IrBuilder(); }
+
+    llvm::BasicBlock *GetEntryBasicBlock() { return entry_basic_block; }
+
+	bool HasColor() { return has_color; }
+	void SetColor(size_t color) { this->color = color; has_color = true; }
+	size_t GetColor() { return color; }
```
lioncash: HasColor and GetColor can be const.

src/binary_translation/InstructionBlock.h
```diff
@@ -0,0 +1,85 @@
...
+    /*
+     * Creates a basic block for use by instructions
+     */
+    llvm::BasicBlock *CreateBasicBlock(const char *name);
+    /*
+	 * Links two instructions, adding to prev and next lists
+	 */
+	static void Link(InstructionBlock *prev, InstructionBlock *next);
+
+    u32 Address();
```
lioncash: Address can be const.

src/binary_translation/InstructionBlock.h
```diff
@@ -0,0 +1,85 @@
...
+    llvm::IRBuilder<> *IrBuilder() { return module->IrBuilder(); }
+
+    llvm::BasicBlock *GetEntryBasicBlock() { return entry_basic_block; }
+
+	bool HasColor() { return has_color; }
+	void SetColor(size_t color) { this->color = color; has_color = true; }
+	size_t GetColor() { return color; }
+
+	std::list<InstructionBlock *> GetNexts() { return nexts; }
+	std::list<InstructionBlock *> GetPrevs() { return prevs; }
```
lioncash: These can both be const.

src/binary_translation/Instructions/Arithmetic.h
```diff
@@ -0,0 +1,36 @@
...
+    };
+    enum class Op
+    {
+        BitwiseAnd = 0, BitwiseXor, Subtract, RevSubtract, Add, AddWithCarry, SubtractWithCarry, ReverseSubtractWithCarry,
+        // Compare, Test, Misc
+        BitwiseOr = 12, MoveAndShifts, BitwiseBitClear, BitwiseNot
+    };
+
+public:
+    virtual bool Decode() override;
```
lioncash: unnecessary `virtual`

src/binary_translation/Instructions/Instruction.h
```diff
@@ -0,0 +1,111 @@
...
+     * Returns true on success, or false otherwise
+     */
+    bool Read(u32 instruction, u32 address);
+
+    /*
+     * Generates non instruction specific code, and then calls GenerateInstructionCode
+     */
+    void GenerateCode(InstructionBlock *instruction_block);
+
+    u32 Address() { return address; }
```
lioncash: Address can be const.

src/binary_translation/Instructions/MovShift.h
```diff
@@ -0,0 +1,34 @@
...
+    {
+        MoveAndLSL, LSR, ASR, RRXAndROR
+    };
+    enum class Form
+    {
+        Register, ImmediateA1, ImmediateA2
+    };
+
+public:
+    virtual bool Decode() override;
```
lioncash: unnecessary `virtual`

src/binary_translation/Instructions/Types.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+/*
+ * A register in a broad sense: R0-R15, and flags
+ */
+enum class Register
+{
+    R0, R1, R2, R3, R4, R5, R6, R7,
+    R8, R9, R10, R11, R12, SP, LR, PC,
+    N, Z, C, V,
```
lioncash: `Register` is pretty misleading here as a name if it includes flags.

Syrco:
How to compile:

* Get llvm (I have ce14bcb58cb8a67740138d07869d423489082c1f)
* Compile LLVM ALL_BUILD
  * Make sure to target only native
  * Using LLVM_TARGETS_TO_BUILD=X86 in cmake
* In citra's cmake
  * ENABLE_BINARY_TRANSLATION
  * LLVM_DIR <llvm build>/share/llvm/cmake

How to run:

* Run `binary_translate.exe ROM ROM.obj`
  * like `binary_translate.exe "PIX3D.3ds" "PIX3D.3ds.obj"`
* Run citra, it should load it automatically and print `Loader <Info>: ... Binary translation enabled`

src/binary_translation/ARMFuncs.h
```diff
@@ -0,0 +1,52 @@
...
+    {
+        llvm::Value *result, *carry;
+    };
+    struct ResultCarryOverflow
+    {
+        llvm::Value *result, *carry, *overflow;
+    };
+
+    static ShiftTN DecodeImmShift(InstructionBlock *instruction, u32 type, u32 imm5);
+    static SRType DecodeRegShift(u32 type);
```
neobrain: This function looks like it's never used.

src/binary_translation/Instructions/Arithmetic.cpp
```diff
@@ -0,0 +1,128 @@
...
+    return false;
+}
+
+void Arithmetic::GenerateInstructionCode(InstructionBlock* instruction_block)
+{
+    auto ir_builder = instruction_block->Module()->IrBuilder();
+
+    llvm::Value *left, *right, *carry_in;
+    ARMFuncs::ResultCarryOverflow result = { nullptr, nullptr, nullptr };
+    auto bitwise = IsBitwise(op);
```
neobrain: Please don't overly use `auto` for simple types like this.

src/binary_translation/Instructions/Arithmetic.cpp
```diff
@@ -0,0 +1,128 @@
...
+    case Op::SubtractWithCarry: result = ARMFuncs::AddWithCarry(instruction_block, left, ir_builder->CreateNot(right), carry_in); break;
+    case Op::ReverseSubtractWithCarry: result = ARMFuncs::AddWithCarry(instruction_block, ir_builder->CreateNot(left), right, carry_in); break;
+    case Op::BitwiseOr: result.result = ir_builder->CreateOr(left, right); break;
+    case Op::BitwiseBitClear: result.result = ir_builder->CreateAnd(left, ir_builder->CreateNot(right)); break;
+    default: break;
+    }
+
+    instruction_block->Write(rd, result.result);
+
+    if (s)
```
neobrain: Variable names should scale with the length of their scope, to make it easier to grasp what a particular variable represents. The name "s" is way too indescriptive.

src/binary_translation/Instructions/Instruction.cpp
```diff
@@ -0,0 +1,108 @@
+#include "Instruction.h"
+#include "common/logging/log.h"
+#include <cassert>
+#include "InstructionBlock.h"
+#include "ModuleGen.h"
+#include "MachineState.h"
+#include "BinarySearch.h"
+
+Instruction::Instruction()
```
neobrain: There's no need for constructors/destructors if they are empty anyway.

yuriks: They might be useful if they're on the .cpp file since they prevent re-instantiation in all uses of the class.

src/binary_translation/Instructions/Instruction.cpp
```diff
@@ -0,0 +1,108 @@
...
+    }
+    // If the basic block is terminated there has been a jump
+    // If not, jump to the next instruction
+    if (!ir_builder->GetInsertBlock()->getTerminator())
+    {
+		instruction_block->Module()->BranchWritePCConst(instruction_block, Address() + 4);
+    }
+}
+
+bool Instruction::ReadFields(const std::initializer_list<FieldDefObject> &fields)
```
neobrain: Haven't looked too deeply into it, but the functionality provided by FieldDefObject seems to be covered by the BitField template in common already. Is this the case indeed, or does this do anything special that's not covered by BitField?

src/binary_translation/Instructions/Instruction.h
```diff
@@ -0,0 +1,111 @@
...
+Instruction::FieldDefObject Instruction::FieldDef(Type* field)
+{
+    return FieldDefObject(BitCount, field, &Instruction::WriteFunction<Type>);
+}
+
+template <typename Type>
+void Instruction::WriteFunction(u32 value, void *field_address)
+{
+    *static_cast<Type *>(field_address) = Type(value);
+}
```
neobrain: Missing newlines at end of the files.

src/binary_translation/Instructions/Ldr.h
```diff
@@ -0,0 +1,26 @@
+#include "Instruction.h"
+#include "Types.h"
+#include <bitset>
+
+class Ldr : public Instruction
```
neobrain: What does Ldr stand for? If it's "Loader", please just write that instead of using potentially confusing abbreviations which safe only 3 letters.

src/binary_translation/Instructions/MovShift.cpp
```diff
@@ -0,0 +1,101 @@
...
+        if (rd == Register::PC && s) return false; // SEE SUBS PC, LR and related instructions;
+        return true;
+    }
+	if (ReadFields({ CondDef(), FieldDef<7>(0x1d), FieldDef<1>(&s), FieldDef<4>(0),
+					 FieldDef<4>(&rd), FieldDef<12>(&imm12) }))
+	{
+		form = Form::ImmediateA1;
+		return true;
+	}
+	if (ReadFields({ CondDef(), FieldDef<8>(0x30), FieldDef<4>(&imm4),
```
neobrain: Use tabs instead of spaces for indentation, please.

src/binary_translation/Instructions/MovShift.h
```diff
@@ -0,0 +1,34 @@
...
+    void GenerateInstructionCode(InstructionBlock* instruction_block) override;
+private:
+    Form form;
+    bool s;
```
neobrain: Please choose more descriptive variable names, and write comments to explain a particular variable's meaning.

src/binary_translation/Instructions/Str.cpp
```diff
@@ -0,0 +1,70 @@
...
+    return false;
+}
+
+void Str::GenerateInstructionCode(InstructionBlock* instruction_block)
+{
+    auto ir_builder = instruction_block->IrBuilder();
+
+    if (form == Form::Immediate)
+    {
+        auto add = U == 1;
```
neobrain: excessive use of auto.

src/binary_translation/Instructions/Types.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+/*
+ * A register in a broad sense: R0-R15, and flags
+ */
+enum class Register
```
neobrain: Please enclose things in a namespace.

src/binary_translation/MachineState.cpp
```diff
@@ -0,0 +1,131 @@
+#include "MachineState.h"
+#include "ModuleGen.h"
+#include "Instructions/Types.h"
+#include <llvm/IR/GlobalValue.h>
+#include <llvm/IR/Constants.h>
+#include <llvm/IR/LLVMContext.h>
+#include <llvm/IR/GlobalVariable.h>
+#include "TBAA.h"
+
+using namespace llvm;
```
neobrain: I'd suggest against `using namespace`, especially if it just saves 6 letters. Explicitly writing `llvm::` makes it easier for people unfamiliar with llvm to read and comprehend the program logic, since it's easier to see which parts come from llvm and which parts are citra-internal.

src/binary_translation/MachineState.cpp
```diff
@@ -0,0 +1,131 @@
...
+    {
+        global = flags_global;
+        index = (static_cast<unsigned>(reg)-static_cast<unsigned>(Register::N)) * 4;
+    }
+    auto base = module->IrBuilder()->CreateAlignedLoad(global, 4);
+    module->GetTBAA()->TagConst(base);
+    return module->IrBuilder()->CreateConstInBoundsGEP1_32(base, index);
+}
+
+Value* MachineState::ReadRegiser(Register reg, bool allow_pc)
```
neobrain: typo

src/binary_translation/MachineState.h
```diff
@@ -0,0 +1,58 @@
+
```
neobrain: no #pragma once

src/binary_translation/MachineState.h
```diff
@@ -0,0 +1,58 @@
...
+enum class Condition;
+enum class Register;
```
neobrain: Also, all files should have the standard citra license header at the top

src/core/loader/loader.cpp
```diff
@@ -13,10 +13,19 @@
...
 
+#if ENABLE_BINARY_TRANSLATION
+#include "core/binary_translation/BinaryTranslationLoader.h"
+#endif
+
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
 namespace Loader {
 
+u32 ROMCodeStart = 0;
```
neobrain: Please don't rely on global initialization. This will cause trouble when trying to run more than one game in a single citra session.

src/binary_translation/Instructions/Instruction.cpp
neobrain: @yuriks does this have any real-world effects other than a marginally smaller program binary?

src/core/loader/loader.cpp
```diff
@@ -110,15 +119,18 @@ ResultStatus LoadFile(const std::string& filename) {
 
     LOG_INFO(Loader, "Loading file %s as %s...", filename.c_str(), GetFileTypeString(type));
 
+    ResultStatus status = ResultStatus::Error;
     switch (type) {
 
     //3DSX file format...
     case FileType::THREEDSX:
-        return AppLoader_THREEDSX(std::move(file)).Load();
+        status = AppLoader_THREEDSX(std::move(file)).Load();
```
purpasmart96: I think plain old "return" is better.

src/binary_translation/ModuleGen.h
```diff
@@ -0,0 +1,113 @@
+#pragma once
```
purpasmart96: lower case names with under slashes for the code files, for  example, should be "module_gen.h"

src/core/binary_translation/BinaryTranslationLoader.h
```diff
@@ -0,0 +1,23 @@
...
+    static bool CanRun(bool specific_address);
+    // Checks whether the cpu state can run the specific address at the specific mode
+    static bool CanRun(u32 pc, bool tflag);
+    // Runs the state provided at SetCpuState.
+    // Returns instruction_count + number of instructions executed
+    static uint32_t Run(uint32_t instruction_count);
+    // Link between Run and VerifyCallback
+    static uint32_t RunInternal(uint32_t instruction_count);
+    static void VerifyCallback();
+};
```
purpasmart96: No new line at EOF

src/core/loader/ncch.cpp
```diff
@@ -116,7 +116,14 @@ ResultStatus AppLoader_NCCH::LoadExec() const {
     std::vector<u8> code;
     if (ResultStatus::Success == ReadCode(code)) {
         Memory::WriteBlock(entry_point, &code[0], code.size());
-        Kernel::LoadExec(entry_point);
+		Kernel::LoadExec(entry_point);
+
+		Loader::ROMCodeStart = entry_point;
+		Loader::ROMCodeSize = entry_point + code.size();
+		// TODO: Don't know where these are
+		Loader::ROMReadOnlyDataStart = 0;
```
purpasmart96: https://github.com/citra-emu/citra/blob/master/src/core/loader/ncch.h#L83
https://github.com/citra-emu/citra/blob/master/src/core/loader/ncch.h#L85

archshift:
I think the binary translation loader would better belong under `loader/` and not `core/binary_translation`.

src/binary_translation/Instructions/Ldr.h
lioncash: LDR is an ARM pseudoinstruction.

src/binary_translation/BinarySearch.h
```diff
@@ -0,0 +1,22 @@
...
+
+struct BinarySearch
+{
+    size_t min;
+    size_t mid;
+    size_t max;
+    BinarySearch(size_t max) : min(0), mid(max / 2), max(max) { }
+    BinarySearch(size_t min, size_t max) : min(min), mid((min + max) / 2), max(max) { }
+    BinarySearch l() { return BinarySearch(min, mid); }
+    BinarySearch r() { return BinarySearch(mid, max); }
```
lioncash: r and l can be const.

src/binary_translation/Instructions/Arithmetic.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    switch (op)
+    {
+    case Arithmetic::Op::BitwiseAnd: return true;
+    case Arithmetic::Op::BitwiseXor: return true;
+    case Arithmetic::Op::BitwiseOr: return true;
+    case Arithmetic::Op::BitwiseBitClear: return true;
+    default: break;
+    }
+}
```
lioncash:
```cpp

bool IsBitwise(Arithmetic::Op op)
{
    return op == Arithmetic::Op::BitwiseAnd ||
           op == Arithmetic::Op::BitwiseXor ||
           op == Arithmetic::Op::BitwiseOr  ||
           op == Arithmetic::Op::BitwiseBitClear;
}
```

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3669,6 +3674,10 @@ static int InterpreterTranslate(ARMul_State* cpu, int& bb_start, u32 addr) {
 translated:
         phys_addr += inst_size;
 
+        if (BinaryTranslationLoader::CanRun(phys_addr, cpu->TFlag))
```
lioncash: Should have an if preprocesser construct around it.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3727,6 +3740,7 @@ unsigned InterpreterMainLoop(ARMul_State* state) {
 #define GOTO_NEXT_INST \
     if (num_instrs >= cpu->NumInstrsToExecute) goto END; \
     num_instrs++; \
+    BinaryTranslationLoader::VerifyCallback(); \
```
lioncash: Here as well.

src/binary_translation/Instructions/Branch.h
```diff
@@ -0,0 +1,20 @@
...
+
+class Branch : public Instruction
+{
+public:
+    enum class Form
+    {
+        Immediate, Register
+    };
+
+public:
```
lioncash: unnecessary public

src/binary_translation/Instructions/Arithmetic.h
```diff
@@ -0,0 +1,36 @@
...
+        Register, RegisterShiftedRegister, Immediate
+    };
+    enum class Op
+    {
+        BitwiseAnd = 0, BitwiseXor, Subtract, RevSubtract, Add, AddWithCarry, SubtractWithCarry, ReverseSubtractWithCarry,
+        // Compare, Test, Misc
+        BitwiseOr = 12, MoveAndShifts, BitwiseBitClear, BitwiseNot
+    };
+
+public:
```
lioncash: unnecessary public

src/binary_translation/Instructions/Instruction.cpp
yuriks: It doesn't really save binary size, but can have a dramatic impact on compile times sometimes.

src/binary_translation/Instructions/MovShift.cpp
Zaneo: Seems to be conflicting info, 

>Also spaces, not tabs.

purpasmart96

archshift: I think that's what neobrain meant, given that you're already using tabs and we use spaces everywhere else.

neobrain: Erm yes, sorry - I meant spaces instead of tabs of course.

src/binary_translation/Instructions/Instruction.cpp
neobrain: Is this one such particular instance, and why would we care about it in this particular case but nowhere else?

src/binary_translation/Instructions/Instruction.cpp
Syrco: FieldDefObject is not supposed to be a bit field. It is supposed to be a field of an Instruction.
Thus it supports either constant value which is verified or a target field to write to.
It does not care about the bit count of the field, but it stores it for the disassembler.

src/core/loader/loader.cpp
Syrco: This is to allow binary translation loading only on success later.

ILOVEPIE:
You're missing a lot of includes, you should try compiling it to figgure out where the errors are so you can clean them up, apart from that, i've currently been recompiling a rom for the last 30 min, so you may want to consider changing the optimization level to -O2 instead of -O3.

Syrco:
@ILOVEPIE what includes? It compiles fine in VS2013. What is the size of the ROM?

ILOVEPIE:
Syrco, you have some visual studio specific things in there, like multi-parameter asserts which prevent it from compiling on linux pcs and macs. Also you should rebase this PR to the latest master to see what won't work, you're missing includes to <list> and <cstdef> in quite a few of your files in the binary_translator target. Also i found a way greatly reduce the time required to compile, it removes some of the optimization passes but makes the compile time shorter by about a factor of 24.

ILOVEPIE:
also at one point you use a variable called "not", which is a reserved word, idk why visual studio compiles that but gcc and clang both throw a hissy fit over it.

src/binary_translation/CodeGen.cpp
```diff
@@ -0,0 +1,163 @@
...
+    pass_manager.add(new DataLayoutPass());
+    target_machine->addAnalysisPasses(pass_manager);
+    function_pass_manager.add(new DataLayoutPass());
+    target_machine->addAnalysisPasses(function_pass_manager);
+
+    pass_manager_builder.OptLevel = 3;
+    pass_manager_builder.SizeLevel = 0;
+    pass_manager_builder.Inliner = createFunctionInliningPass(3, 0);
+    pass_manager_builder.LoopVectorize = true;
+    pass_manager_builder.SLPVectorize = true;
```
ILOVEPIE: pass_manager_builder.OptLevel = 3;
    pass_manager_builder.SizeLevel = 0;
    pass_manager_builder.Inliner = createFunctionInliningPass(3, 0);
    pass_manager_builder.LoopVectorize = true;
    pass_manager_builder.SLPVectorize = true;

    pass_manager_builder.populateFunctionPassManager(function_pass_manager);
	
	pass_manager_builder.OptLevel = 1;
    pass_manager_builder.SizeLevel = 0;
    pass_manager_builder.LoopVectorize = false;
    pass_manager_builder.SLPVectorize = false;

replacing lines 127 through 133 with the code above results in a decrease in compilation time by a factor of 24, reducing recompiling Ocarina of Time 3D from 4 hours to 10 min.

Syrco:
@ILOVEPIE: What passes did you remove?

ILOVEPIE:
I changed it to apply the -O3 passes to the functions, and the -O1 passes to the overall binary

Syrco:
@ILOVEPIE I'm not sure what the effect on optimization level will be. I do not expect cross function optimizations to work, as functions don't call each other. But I assume there are optimization that are not run in the function passes but will improve each function on it's own.

ILOVEPIE:
another option would be to allow the user to specifiy the optimization level with an -O flag

src/binary_translation/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+			Instructions/Arithmetic.h
+			Instructions/Ldr.h
+			Instructions/Str.h
+            )
+			
+create_directory_groups(${SRCS} ${HEADERS})
+
+include_directories(.)
+add_executable(binary_translate ${SRCS} ${HEADERS})
+target_link_libraries(binary_translate ${llvm_libs})
```
ILOVEPIE: In order to get this target to compile with gcc and clang line 43 must be moved to the end of the file so that it is the last target_link_librarys call in the file.

Subv:
Maybe the recompilation can be performed offline (with an external program) instead of doing it when the ROM is loaded?

ILOVEPIE:
@Subv that's what it does, there's an external program called binary_translate which translates the rom.

src/binary_translation/MachineState.cpp
```diff
@@ -0,0 +1,131 @@
...
+    {
+    case Condition::NE: case Condition::CC: case Condition::PL: case Condition::VC:
+    case Condition::LS: case Condition::LT: case Condition::LE:
+        not = true;
+        cond = (Condition)((int)cond - 1);
+    }
+
+    switch (cond)
+    {
+    case Condition::EQ: pred = ReadRegiser(Register::Z); break;
```
ILOVEPIE: still not fixed

CMakeLists.txt
```diff
@@ -192,4 +192,13 @@ if(ENABLE_QT)
...
     add_subdirectory(externals/qhexedit)
 endif()
+
+option(ENABLE_BINARY_TRANSLATION "Enable binary translation. Requires LLVM" OFF)
+if(ENABLE_BINARY_TRANSLATION)
+    add_definitions(-DENABLE_BINARY_TRANSLATION)
+    find_package(LLVM REQUIRED CONFIG)
+    include_directories(${LLVM_INCLUDE_DIRS})
+    add_definitions(${LLVM_DEFINITIONS})
+    llvm_map_components_to_libnames(llvm_libs Core Support X86 ExecutionEngine MCJIT BitWriter ipo)
```
ILOVEPIE: change `X86` to `native` to allow support for ports to platforms other than x86, also will allow x86_64 builds to utilize x86_64 code.

src/binary_translation/CodeGen.cpp
```diff
@@ -0,0 +1,164 @@
...
+    raw_os_ostream stream(file);
+    if (target_machine->addPassesToEmitMC(pass_manager, context, stream, false))
+    {
+        LOG_CRITICAL(BinaryTranslator, "Target does not support MC emission!");
+        return;
+    }
+    LOG_INFO(BinaryTranslator, "Generating code");
+    pass_manager.run(*module);
+    stream.flush();
+    file.close();
```
lioncash: close isn't necessary. File streams close at the end of scope.

src/binary_translation/Instructions/Instruction.h
```diff
@@ -0,0 +1,111 @@
...
+
+    Condition cond;
+};
+
+/*
+ * Object produced by FieldDef
+ */
+class Instruction::FieldDefObject
+{
+public:
```
lioncash: this can be merged with the public below.

i30817:
If i understand this correctly, it's a ahead of time compiler (and hopefully optimizer) for the N3DS executables inside rooms. Assuming that the 'real' executable format (i assume not cci) doesn't pack huge amounts of rom data, there is no danger of when using this suddenly a 5 gigabytes file gets duplicated on disk? Maybe a bad assumption.

Another thing is for major bug fixes/optimizations, it maybe a good idea to ask for automatic recompilation of the executable? Does the PR have any kind of header of the version of the AoT recompiler was used on the produced files?

ILOVEPIE:
@i30817 The recompiler only recompiles the code, the actual data that gets used is from the .cci file and the better thing to do would be to include a gui for the recompiler. As for savestates it shouldn't impact the recompiler significantly, the one issue with the recompiler is that it takes a long time to process. This isn't jit compilation (JIT means Just In Time, i.e. it recompiles a section of the code before executing it) this is AOT recompilation (meaning it recompiles the whole entirety of the code before executing it)

JMC47:
I think there are a few reasons why this shouldn't really be the path that Citra goes.  Feel free to disagree with me if you wish, but, here's my experiences with it.

1:  Considering how weak the 3DS is the interpreter isn't even close to the bottleneck.  There's really no performance to be gained currently.
2:  The compile times are ridiculous.  I've been waiting four hours on MGS3-3D and I have no idea how much longer it's going to take before I decide screw it.
3:  Again on the performance subject, would it really make a difference over a JIT when we are likely to be limited on the renderers anyway?
4:  Having two separate programs is rather annoying, plus it adds dependencies for those who wish to compile Citra (LLVM)
5:  It has bugs that aren't in the interpreter.  Broken controls in Super Monkey Ball 3D, for instance.

Overall, it's a neat proof of concept, but the usability side just isn't up to snuff for me.  If things could be ironed out, at least my concerns could be quelled, but I think it has a long way to go personally.

ILOVEPIE:
JMC it's an RFC, which means it's not ready to have people make judgements on weither or not it should be pulled yet.

JMC47:
I was discussing the merits of it on the whole, not whether it should be pulled in its current state.  I don't know if an AoT compiler will ever be worth the headaches it brings.

neobrain:
I agree with what @JMC47 said. This code surely was a huge amount of work, but in its current form I don't think it cannot be considered more than a nice experiment. I'm not sure if the kinks could be worked out - the bugs maybe can, the "two separate programs" one as well. On the other hand, the LLVM dependence really isn't too nice, and the "4 hours compile time" issue kills it all for any reasonable use. Maybe there is a fancy way to only partially compile things and then have the AoT compilation run in parallel to interpreter emulation, though.

Anyway, I think you understand what I'm getting at.

ILOVEPIE:
It actually takes 10 min with the modifications i suggested.

ILOVEPIE:
This has promise but the optimization settings must be changed also needs extensive testing and bugfix work. We'll need llvm anyways for a jit. This also needs some thought about how it will handle CRO files.

JMC47:
Two things I want to add real quick.

1:  I don't believe Lioncash's JIT will use LLVM.
2:  The other huge issue is the filesizes of the files created by the translator.  We can't ask user to deal with a significant wait (10 minutes or hours, let's just say 10 mins for the sake of conversation) and then bloat the rom to 3x as big on top of it.  This is really another huge issue with this that a JIT will not run into.

yuriks:
What's this about bloating the ROM 3x as big? The .text of a 3DS app rarely exceeds 10MB at most (in fact the largest of my games is only ~6MB) so there's no way this should be making even a *dent* on disk usage compared to the RomFS of the original cxi unless there's something very wrong with the approach used here.

Anyway, while I don't disagree in principle with AoT compilation, I've seen signs that this implementation is quite flawed, so it would need to see some improvement. Apart from the file size issue reported by @JMC47, the very long compilation times make me thing that the entire program graph is being compiled as a single unit, which is certainly non-ideal.

Others have expressed that the LLVM dependency is quite cumbersome, and I'm somewhat inclined to agree. if we were emulating a different system, I would probably have stronger feelings about this due to increased optimization opportunities, but given that the 3DS has relatively weak CPUs (Where you can get pretty decent performance even with an interpreter like ours.) makes this level of optimization non-essential.

ILOVEPIE:
Since the original author seems inactive should I attempt to remedy some of
these issues?

ILOVEPIE:
We can use some of the emscripten code to remedy the single unit compilation issue.

ILOVEPIE:
ill pick this up since the author isn't active.

bunnei:
I think this has been given sufficient time for comments, and we heard anything back from the author in over a month. Feel free to open another PR in the future if you plan on pursuing more work with this.

ILOVEPIE:
I've updated the recompiler to the latest citra and added 64bit support in my RecompilerLatest branch, i'll try working on some of the things we discussed here, llvm does however provide a nice framework for doing the recompilation that supports multiple CPU architectures, which means this recompiler should have very little problems being ported to say PowerPC. I'll rewrite it without llvm if necessary but for now i'll stick with llvm.

src/binary_translation/CodeGen.cpp
ILOVEPIE: it's necessary to get llvm to work, i tried this without it and it wouldn't write the files,

PR #725, opened 2015-05-07
Remove unused hash and mem_arena from common

yuriks:
Notably, this gets rid of mem_arena, which should fix 32-bit builds crashing on bootup. (Haven't tested though.)


bunnei:
LGTM

purpasmart96:
Thank you, mem_arena needed to go, badly...

yuriks:
Just tested, and this *does* fix 32-bit builds. Wee!

linkmauve:
Also tested on Linux, the i686 build even seem to work better than the amd64!

lioncash:
Looks fine to me.

PR #726, opened 2015-05-07
GPU improvements

bunnei:
This is a bunch of small GPU emulation fixes/improvements initially started to fix Steel Diver. Commits are all separate and should probably be individually reviewed. Summary:
* Added MADI and FLR vertex shader instructions
* Added combiner buffer (i.e. "previous combiner buffer" color source)
* Added "AddSigned" combiner operation
* Added combiner output scaling
* Fixed a bug with depth testing (via @Subv)
* Fixed a bug with invalid vectors being returned for unimplemented features

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -226,7 +226,8 @@ struct Regs {
             Texture1               = 0x4,
             Texture2               = 0x5,
             Texture3               = 0x6,
-            // 0x7-0xc = primary color??
+
+            PreviousBuffer         = 0xd,
             Constant               = 0xe,
             Previous               = 0xf,
         };
@@ -309,11 +310,36 @@ struct Regs {
     TevStageConfig tev_stage2;
     INSERT_PADDING_WORDS(0x3);
     TevStageConfig tev_stage3;
-    INSERT_PADDING_WORDS(0x13);
+    INSERT_PADDING_WORDS(0x3);
+
+    union {
+        // Tev stages 0-3 write their output to the combiner buffer if the corresponding bit in
+        // these masks are set
+        BitField< 8, 4, u32> update_mask_rgb;
+        BitField<12, 4, u32> update_mask_a;
+
+        bool TevStageUpdatesCombinerBufferColor(unsigned stage_index) const {
+            return (stage_index < 4) && (update_mask_rgb & (1 << stage_index));
+        }
+
+        bool TevStageUpdatesCombinerBufferAlpha(unsigned stage_index) const {
+            return (stage_index < 4) && (update_mask_a & (1 << stage_index));
+        }
+    } tev_combiner_buffer_input;
+    
+    INSERT_PADDING_WORDS(0xf);
     TevStageConfig tev_stage4;
     INSERT_PADDING_WORDS(0x3);
     TevStageConfig tev_stage5;
-    INSERT_PADDING_WORDS(0x3);
+
+    union {
+        BitField< 0, 8, u32> r;
+        BitField< 8, 8, u32> g;
+        BitField<16, 8, u32> b;
+        BitField<24, 8, u32> a;
+    } tev_combiner_buffer_color;
+
+    INSERT_PADDING_WORDS(0x2);
 
     const std::array<Regs::TevStageConfig,6> GetTevStages() const {
         return { tev_stage0, tev_stage1,
@@ -784,8 +810,10 @@ struct Regs {
         ADD_FIELD(tev_stage1);
         ADD_FIELD(tev_stage2);
         ADD_FIELD(tev_stage3);
+        ADD_FIELD(tev_combiner_buffer_input);
         ADD_FIELD(tev_stage4);
         ADD_FIELD(tev_stage5);
+        ADD_FIELD(tev_combiner_buffer_color);
         ADD_FIELD(output_merger);
         ADD_FIELD(framebuffer);
         ADD_FIELD(vertex_attributes);
@@ -859,8 +887,10 @@ ASSERT_REG_POSITION(tev_stage0, 0xc0);
 ASSERT_REG_POSITION(tev_stage1, 0xc8);
 ASSERT_REG_POSITION(tev_stage2, 0xd0);
 ASSERT_REG_POSITION(tev_stage3, 0xd8);
+ASSERT_REG_POSITION(tev_combiner_buffer_input, 0xe0);
 ASSERT_REG_POSITION(tev_stage4, 0xf0);
 ASSERT_REG_POSITION(tev_stage5, 0xf8);
+ASSERT_REG_POSITION(tev_combiner_buffer_color, 0xfd);
 ASSERT_REG_POSITION(output_merger, 0x100);
 ASSERT_REG_POSITION(framebuffer, 0x110);
 ASSERT_REG_POSITION(vertex_attributes, 0x200);
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -452,9 +452,7 @@ struct Regs {
         D24S8  = 3
     };
 
-    /*
-     * Returns the number of bytes in the specified depth format
-     */
+    // Returns the number of bytes in the specified depth format
     static u32 BytesPerDepthPixel(DepthFormat format) {
         switch (format) {
         case DepthFormat::D16:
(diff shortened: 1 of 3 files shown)


src/video_core/pica.h
```diff
@@ -306,11 +320,35 @@ struct Regs {
...
-    INSERT_PADDING_WORDS(0x13);
+    INSERT_PADDING_WORDS(0x3);
+
+    union {
+        // Tev stages 0-3 write their output to the combiner buffer if the corresponding bit in
+        // these masks are set
+        BitField< 8, 4, u32> update_mask_rgb;
+        BitField<12, 4, u32> update_mask_a;
+
+        bool UpdateCombinerBufferColor(unsigned stage_index) const {
```
neobrain: Confusing name (this sounds like a command, rather than like a getter function).

src/video_core/pica.h
```diff
@@ -440,6 +476,20 @@ struct Regs {
         }
     }
 
+    // Returns the number of bits per depth component of the specified depth format
+    static u32 BitsPerDepthFormat(DepthFormat format) {
```
neobrain: Imprecise name: You're not returning the bits per format, but the depth bits with respect to a particular format.

src/video_core/vertex_shader.cpp
```diff
@@ -424,6 +437,7 @@ static void ProcessShaderCode(VertexShaderState& state) {
 
                     dest[i] = src1[i] * src2[i] + src3[i];
                 }
+
```
neobrain: pointless empty line

purpasmart96:
Could you also add the MIN shader opcode, it's the same as MAX except you just use std::min

src/video_core/pica.h
```diff
@@ -306,11 +320,35 @@ struct Regs {
...
+
+    union {
+        // Tev stages 0-3 write their output to the combiner buffer if the corresponding bit in
+        // these masks are set
+        BitField< 8, 4, u32> update_mask_rgb;
+        BitField<12, 4, u32> update_mask_a;
+
+        bool UpdateCombinerBufferColor(unsigned stage_index) const {
+            return (stage_index < 4) && (update_mask_rgb & (1 << stage_index));
+        }
```
Subv: Can you add a line between these two functions?

src/video_core/pica.h
Subv: I'd be OK with "UpdatesCombinerBufferColor" (note the s)

bunnei:
@purpasmart96 I'll add it in another PR once I hit a game that uses it

MoochMcGee:
pls squash your commits.

MoochMcGee:
Other than that, this PR LGTM

src/video_core/pica.h
```diff
@@ -440,6 +477,20 @@ struct Regs {
         }
     }
 
+    // Returns the number of bits per depth component of the specified depth format
+    static u32 BitsPerDepthComponent(DepthFormat format) {
```
yuriks: Still a bad name. `DepthComponentBits` or something.

Subv:
LGTM

src/video_core/pica.h
MoochMcGee: I think it's fine.

bunnei:
@Subv I think he just meant the autosquash

src/video_core/pica.h
bunnei: I was just trying to be consistent with the function immediately above, `BytesPerDepthPixel`
IMO it's fine as well, but if you insist I can change it.

neobrain: BitsOfDepthPerPixel, maybe?

yuriks: `DepthBitsPerPixel`?

MoochMcGee:
still LGTM

yuriks:
LGTM

src/video_core/pica.h
```diff
@@ -296,7 +297,20 @@ struct Regs {
...
         };
 
-        INSERT_PADDING_WORDS(0x1);
+        union {
+            BitField< 0, 2, u32> color_scale;
+            BitField<16, 2, u32> alpha_scale;
+        };
+
+        inline int GetColorMultiplier() const {
+            ASSERT(color_scale < 3);
```
neobrain: Why is this case ruled out? If its behavior is unknown, add a TODO for it.

src/video_core/pica.h
```diff
@@ -296,7 +297,20 @@ struct Regs {
             BitField<24, 8, u32> const_a;
         };
 
-        INSERT_PADDING_WORDS(0x1);
+        union {
+            BitField< 0, 2, u32> color_scale;
+            BitField<16, 2, u32> alpha_scale;
+        };
+
+        inline int GetColorMultiplier() const {
```
neobrain: should return unsigned

src/video_core/rasterizer.cpp
```diff
@@ -376,7 +376,13 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
             // with some basic arithmetic. Alpha combiners can be configured separately but work
             // analogously.
             Math::Vec4<u8> combiner_output;
-            for (const auto& tev_stage : tev_stages) {
+            Math::Vec4<u8> combiner_buffer = {
+                registers.tev_combiner_buffer_color.r, registers.tev_combiner_buffer_color.g,
+                registers.tev_combiner_buffer_color.b, registers.tev_combiner_buffer_color.a
+            };
+
+            for (unsigned tev_stage_index = 0; tev_stage_index < tev_stages.size(); ++tev_stage_index) {
```
neobrain: Bleh. I suggest instead using boost.range:

    for (const auto& indexed_tev_stage : tev_stages | boost::adators::indexed()) {
        const auto& tev_stage = indexed_tev_stage.value();
        size_t tev_stage_index = indexed_tev_stage.index();

neobrain: Actually, a better alternative is to remove the TevStageUpdatesCombinerBuffer functions and instead add a boolean member in TevStage with that info. We can just keep the range based for-loop then.

src/video_core/rasterizer.cpp
```diff
@@ -490,6 +499,15 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         return result.Cast<u8>();
                     }
 
+                    case Operation::AddSigned:
+                    {
+                        auto result = input[0].Cast<int>() + input[1].Cast<int>() - Math::MakeVec<int>(128, 128, 128);
```
neobrain: Where does the value 128 come from (rather than e.g. 127) ?

neobrain: In any case this should have some documentation defining what "signed addition" is supposed to mean.

src/video_core/rasterizer.cpp
```diff
@@ -624,9 +655,10 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
 
             // TODO: Does depth indeed only get written even if depth testing is enabled?
             if (registers.output_merger.depth_test_enable) {
-                u16 z = (u16)((v0.screenpos[2].ToFloat32() * w0 +
-                            v1.screenpos[2].ToFloat32() * w1 +
-                            v2.screenpos[2].ToFloat32() * w2) * 65535.f / wsum);
+                int num_bits = Pica::Regs::DepthBitsPerPixel(registers.framebuffer.depth_format);
```
neobrain: unsigned

neobrain:
Gpu is an abbreviation and hence should be spelled "GPU".

Other than that and my other comments, LGTM.

src/video_core/rasterizer.cpp
bunnei: I was under the impression 0.5f -> 128

bunnei:
Thats just GitHub naming the PR from the branch name.........

neobrain:
@bunnei you can rename the PR title. do eet.

src/video_core/rasterizer.cpp
neobrain: It could also be 127, or 127.5. Should have a TODO if you're not sure and it's not verified.

src/video_core/rasterizer.cpp
bunnei: To your first comment: Sorry, but I don't see any real reason why using boost.range makes this any "better". IMO it's just less readable to anyone who does not know Boost that well, when the latter is perfectly readable, functional, and sufficient.

To your second comment: I don't see an obvious way to determine the stage from within `TevStageConfig`.

So I'm going to leave this as-is and continue onward... If you come up with a good solution to #2, I'll PR it.

yuriks: I agree with bunnei on both counts.

PR #728, opened 2015-05-08
dyncom: Remove an unnecessary variable in the interpreter

lioncash:
All this was doing was needlessly aliasing a variable.


archshift:
LGTM

yuriks:
LGTM. I'm curious if this causes any performance change. (Probably doesn't, but I suspect that was the original reason for the extra variable.)

PR #731, opened 2015-05-08
Kernel: Process class and ExHeader caps parsing

yuriks:
This adds a `Process` object to the kernel, which for now only holds kernel process settings. It also adds support for reading those from the ExHeader in NCCHs, or providing default values for other file types.

### DIFF
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -85,7 +85,7 @@ void Process::ParseKernelCaps(const u32* kernel_caps, size_t len) {
 }
 
 void Process::Run(VAddr entry_point, s32 main_thread_priority, u32 stack_size) {
-    g_main_thread = Kernel::SetupMainThread(stack_size, entry_point, main_thread_priority);
+    Kernel::SetupMainThread(stack_size, entry_point, main_thread_priority);
 }
 
 Kernel::Process::Process() {}
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -1,0 +1,35 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/assert.h"
+
+#include "core/hle/kernel/process.h"
+#include "core/hle/kernel/thread.h"
+
+namespace Kernel {
+
+SharedPtr<Process> Process::Create(std::string name, u64 program_id) {
+    SharedPtr<Process> process(new Process);
+
+    process->svc_access_mask.set();
+    process->name = std::move(name);
+    process->program_id = program_id;
+
+    return process;
+}
+
+void Process::ParseKernelCaps(const u32 * kernel_caps, size_t len) {
+    //UNIMPLEMENTED();
+}
+
+void Process::Run(VAddr entry_point, s32 main_thread_priority, u32 stack_size) {
+    g_main_thread = Kernel::SetupMainThread(stack_size, entry_point, main_thread_priority);
+}
+
+Kernel::Process::Process() {}
+Kernel::Process::~Process() {}
+
+SharedPtr<Process> g_current_process;
+
+}
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -59,16 +59,16 @@ void Process::ParseKernelCaps(const u32* kernel_caps, size_t len) {
             u32 end_desc = kernel_caps[i+1];
             ++i; // Skip over the second descriptor on the next iteration
 
-            StaticAddressMapping mapping;
+            AddressMapping mapping;
             mapping.address = descriptor << 12;
             mapping.size = (end_desc << 12) - mapping.address;
             mapping.writable = descriptor & BIT(20);
             mapping.unk_flag = end_desc & BIT(20);
 
-            static_address_mappings.push_back(mapping);
+            address_mappings.push_back(mapping);
         } else if ((type & 0xFFF) == 0xFFE) { // 0x000F
             // Mapped memory page
-            StaticAddressMapping mapping;
+            AddressMapping mapping;
             mapping.address = descriptor << 12;
             mapping.size = Memory::PAGE_SIZE;
             mapping.writable = true; // TODO: Not sure if correct
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -18,6 +18,9 @@ SharedPtr<Process> Process::Create(std::string name, u64 program_id) {
     process->name = std::move(name);
     process->program_id = program_id;
 
+    process->flags.raw = 0;
+    process->flags.memory_region = MemoryRegion::APPLICATION;
+
     return process;
 }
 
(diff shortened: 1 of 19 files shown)


src/core/hle/kernel/process.cpp
```diff
@@ -0,0 +1,100 @@
...
+    return process;
+}
+
+void Process::ParseKernelCaps(const u32* kernel_caps, size_t len) {
+    for (int i = 0; i < len; ++i) {
+        u32 descriptor = kernel_caps[i];
+        u32 type = descriptor >> 20;
+
+        if (descriptor == 0xFFFFFFFF) {
+            continue;
```
bunnei: What does this mean? Please add a comment or define `0xFFFFFFFF` somewhere.

src/core/hle/kernel/process.cpp
```diff
@@ -0,0 +1,100 @@
...
+        } else if ((type & 0xFF8) == 0xFF0) { // 0x007F
+            // Misc. flags
+            bool allow_debug       = descriptor & BIT(0);
+            bool force_debug       = descriptor & BIT(1);
+            bool allow_nonalphanum = descriptor & BIT(2);
+            shared_page_writable   = descriptor & BIT(3);
+            privileged_priority    = descriptor & BIT(4);
+            bool allow_main_args   = descriptor & BIT(5);
+            bool shared_device_mem = descriptor & BIT(6);
+            bool runnable_on_sleep = descriptor & BIT(7);
```
bunnei: BitField?

src/core/loader/loader.cpp
```diff
@@ -42,19 +49,11 @@ static FileType IdentifyFile(FileUtil::IOFile& file) {
...
-    if (filename.size() == 0) {
-        LOG_ERROR(Loader, "invalid filename %s", filename.c_str());
-        return FileType::Error;
-    }
-
-    size_t extension_loc = filename.find_last_of('.');
-    if (extension_loc == std::string::npos)
-        return FileType::Unknown;
-    std::string extension = Common::ToLower(filename.substr(extension_loc));
+static FileType GuessFromFilename(const std::string& extension_) {
```
linkmauve: Rename that to GuessFromExtension, to better reflect this change.

src/core/loader/loader.cpp
```diff
@@ -115,11 +117,11 @@ ResultStatus LoadFile(const std::string& filename) {
 
     //3DSX file format...
     case FileType::THREEDSX:
-        return AppLoader_THREEDSX(std::move(file)).Load();
+        return AppLoader_THREEDSX(std::move(file), filename_filename).Load();
```
linkmauve: Why do you pass the filename without extension here, instead of the entire filename?

src/core/hle/kernel/process.cpp
yuriks: All descriptor types have different fields, and they even overlap they bits that identify the type for some fields. (For example, the interrupt and SVC mask descriptors have a very short prefix identifying the type and then use the rest of the word for the bits.) Defining this struct as a union with BitField would end up with a lot of overlapping fields which I think would be quite confusing, so I decided to explicitly not use BitField.

linkmauve:
Overall LGTM, but why do you create the process in each loader? Wouldn’t it be better for the loader to only wrap each different file type, and then return the loader itself for the caller to create a process and load the binary the loader would provide?

src/core/loader/loader.cpp
yuriks: This filename is only used to identify the process (since 3dsx doesn't have a name in the header.) The idea is that if you f.ex. open `3dscraft.3dsx` the process name will end up as `3dscraft`.

yuriks:
@linkmauve: Sharing more code between the loaders would be ideal, but this gets harder as more correct process loading is implemented: Each loader will have to setup the process memory map, which is something that will change depending on the file format. I suppose I could get around that by assuming a (less flexible) fixed memory map structure of single .text/.rodata/.data/.bss sections for all executable types, and then the loaders would return a struct with this information for the kernel to setup the memory map in that way. Here I am mirroring the implementation on the system, where this is all handled by the loader to my knowledge.

src/core/loader/ncch.h
```diff
@@ -109,7 +111,13 @@ struct ExHeader_StorageInfo{
 struct ExHeader_ARM11_SystemLocalCaps{
```
bunnei: These structs are missing a ` ` before the `{`

src/core/hle/kernel/process.cpp
bunnei: Also, are these local variables just defined for documentation since they are not being used anywhere...?

src/core/loader/ncch.h
yuriks: They were already like that. Should I got and reformat all of them?

src/core/hle/kernel/process.cpp
```diff
@@ -0,0 +1,100 @@
...
+            mapping.size = (end_desc << 12) - mapping.address;
+            mapping.writable = descriptor & BIT(20);
+            mapping.unk_flag = end_desc & BIT(20);
+
+            static_address_mappings.push_back(mapping);
+        } else if ((type & 0xFFF) == 0xFFE) { // 0x000F
+            // Mapped memory page
+            StaticAddressMapping mapping;
+            mapping.address = descriptor << 12;
+            mapping.size = 0x1000;
```
bunnei: Define a constant for this?

src/core/hle/kernel/process.cpp
yuriks: Yeah, most of those aren't relevant to us or I'm not exactly sure what they do, instead of leaving them out entirely I left them as dummy local variables that we can easily change to store on the class later.

src/core/hle/kernel/process.h
```diff
@@ -0,0 +1,62 @@
...
+public:
+    static SharedPtr<Process> Create(std::string name, u64 program_id);
+
+    std::string GetTypeName() const override { return "Process"; }
+    std::string GetName() const override { return name; }
+
+    static const HandleType HANDLE_TYPE = HandleType::Process;
+    HandleType GetHandleType() const override { return HANDLE_TYPE; }
+
+    std::string name; ///< Name of the process
```
bunnei: Inconsistent use of `///<` here and `//` elsewhere in this class

src/core/hle/kernel/process.h
```diff
@@ -0,0 +1,62 @@
...
+    void Run(VAddr entry_point, s32 main_thread_priority, u32 stack_size);
+
+private:
+    Process();
+    ~Process() override;
+};
+
+extern SharedPtr<Process> g_current_process;
+
+}
```
bunnei: This whole header could use more doxygen documentation

src/core/loader/loader.cpp
```diff
@@ -139,11 +140,15 @@ ResultStatus LoadFile(const std::string& filename) {
     // Raw BIN file format...
     case FileType::BIN:
```
bunnei: I originally added loading BIN a while ago when I was just getting started with the project and wanted to just load/debug the code section of a NCCH. Does this still make sense to keep around anymore? Does it even work?

src/core/loader/loader.h
```diff
@@ -105,10 +107,18 @@ class AppLoader : NonCopyable {
 };
 
 /**
+ * Common address mappings found in most games, used for binary format that don't have this
```
bunnei: *formats

src/core/loader/loader.h
```diff
@@ -105,10 +107,18 @@ class AppLoader : NonCopyable {
...
+extern const std::initializer_list<Kernel::StaticAddressMapping> default_address_mappings;
+
+/**
  * Identifies and loads a bootable file
  * @param filename String filename of bootable file
  * @return ResultStatus result of function
  */
 ResultStatus LoadFile(const std::string& filename);
 
+
```
bunnei: Unnecessary extra whitespace

src/core/hle/kernel/process.cpp
bunnei: Doesn't this create compile time warnings?

src/core/loader/ncch.h
bunnei: I realized they were already like that... just figured a quick search+replace would be easy enough since you are already updating the file

src/core/hle/kernel/process.cpp
yuriks: Not on my warning level at least. At any rate, I replaced these with a BitField union.

src/core/loader/loader.cpp
yuriks: It still works... I think. But only because we map all the address space to something. As discussed with @archshift on IRC, I'm gonna remove it soon.

bunnei:
This is good

Subv:
Is the g_main_thread variable used anywhere else? If not, then it can be removed

bunnei:
Builds on Windows, I won't make you wait @yuriks

purpasmart96:
Nice

yuriks:
@Subv Did you have any comments on this PR that I can address "post-mergeously"?

Subv:
LGTM

PR #213, opened 2014-11-19
[RFC] Make the OpenGL renderer work on GL 2.0+ and GLES 2.0+

linkmauve:
In order to do that, I introduced an OpenGL section in both config files, allowing the user to select the OpenGL version they want to use.
And to get the function corresponding to the context we got, move from glLoadGen to libepoxy.
This libepoxy change isn’t ready for merge yet, I didn’t change the CMakeLists.txt file to support MSVC or OSX, only pkg-config is currently supported.

PR #732, opened 2015-05-08
OpenGL 2.1 support

kitling:
With a replacement of the generated files that were being used, and downgrading the shader version (not quite to 1.2, the shaders for GL 2.1, but to 1.3, the shaders for GL 3.0), you can make citra use 2.1 instead of 3.2. No code was changed besides the includes and the versions (and the generated files, of course).


yuriks:
Since OpenGL 2.1 has many more entrypoints than 3 Core does, can you slim down the loader files so that only functions we need are included? This should make things somewhat faster overall. I think removing only the now-deprecated functions and leaving the ones still in 3.2 should be fine, so as to not remove things we might potentially want in the future.

What platforms did you test this on?

kitling:
I tested on my 'bad' laptop, running Ubuntu Studio, which glxinfo says only supports OpenGL 2.1. Which still leaves the mystery of GLSL 1.3 (GL 3.0) shaders working. *shrugs* So, I've tested on a Debian linux based system, but it seems to compile on the other platforms, too. I don't have any way to test the others...

Hmm. Slimming down the loader files. I can look into that, though we would need to change the readme for the generated files to say that it's a slimmed down version with only the things we need. I'll look into it.

For future notice: What I'm looking at to determine deprecated functions is here: https://www.opengl.org/registry/oldspecs//gl.spec
Which, itself, is deprecated. Funny how things like that work.

yuriks:
@linkmauve eventually plans on moving to Citra to libepoxy (See #213) so I think hand-slimming the files is fine as an interim solution. I'll try to test here on Windows and on Virtualbox/Linux, but we really need someone to test on OS X as that's usually the problematic platform.

archshift:
Don't worry, I'll test it on OS X.

archshift:
Turns out, it segfaults upon running a game:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007fff97557152 in __block_descriptor_tmp () from /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit
(gdb) bt
#0  0x00007fff97557152 in __block_descriptor_tmp () from /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit
#1  0x00007fff5fbff730 in ?? ()
#2  0x0000000100122f5c in ProcExtsFromExtString (strExtList=0x0) at /tmp/citra/src/video_core/renderer_opengl/generated/gl_2_1_core.c:1847
Backtrace stopped: frame did not save the PC
```

archshift:
Got a little farther with the newest push, still segfaulting after failing to compile shaders.

```
[   1.196579] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadShaders:38: Error compiling vertex shader:
ERROR: 0:2: '' :  version '130' is not supported
ERROR: 0:13: 'modelview_matrix' : syntax error: syntax error
 
[   1.196766] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadShaders:58: Error compiling fragment shader:
ERROR: 0:2: '' :  version '130' is not supported
 
[   1.196824] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadShaders:80: Error linking shader:
ERROR: One or more attached shaders not successfully compiled
 
 
Program received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x000000010012ad8f in RendererOpenGL::InitOpenGLObjects (this=0x1083ba860) at /tmp/citra/src/video_core/renderer_opengl/renderer_opengl.cpp:191
#2  0x000000010012b312 in RendererOpenGL::Init (this=0x1083ba860) at /tmp/citra/src/video_core/renderer_opengl/renderer_opengl.cpp:340
#3  0x0000000100164e3f in VideoCore::Init (emu_window=0x108300ad0) at /tmp/citra/src/video_core/video_core.cpp:27
#4  0x00000001000db586 in System::Init (emu_window=0x108300ad0) at /tmp/citra/src/core/system.cpp:24
#5  0x0000000100005ac8 in main (argc=2, argv=0x7fff5fbff988) at /tmp/citra/src/citra/citra.cpp:43
```

kitling:
So, finally. I hope this works for Mac OS. I changed texture() to texture2D(). And changed variables to varying where necessary. This is officially 2.1. I hope. I'll come back tomorrow if it still doesn't work.

archshift:
Unfortunately not:

```
[   1.235032] Render.OpenGL <Info> video_core/renderer_opengl/renderer_opengl.cpp:Init:339: GL_VERSION: 2.1 INTEL-10.6.20
[   1.241834] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadShaders:38: Error compiling vertex shader:
ERROR: 0:3: Invalid storage qualifiers 'in' in global variable context
ERROR: 0:4: Invalid storage qualifiers 'in' in global variable context
ERROR: 0:5: Invalid storage qualifiers 'out varying' in global variable context
ERROR: 0:16: Use of undeclared identifier 'vert_position'
ERROR: 0:17: Use of undeclared identifier 'frag_tex_coord'
ERROR: 0:17: Use of undeclared identifier 'vert_tex_coord'

[   1.242334] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadShaders:58: Error compiling fragment shader:
ERROR: 0:3: Invalid storage qualifiers 'in varying' in global variable context
ERROR: 0:4: Invalid storage qualifiers 'out varying' in global variable context
ERROR: 0:7: Use of undeclared identifier 'color'
ERROR: 0:7: Use of undeclared identifier 'frag_tex_coord'

[   1.242428] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadShaders:80: Error linking shader:
ERROR: One or more attached shaders not successfully compiled
```

src/citra/emu_window/emu_window_glfw.cpp
```diff
@@ -76,11 +76,11 @@ EmuWindow_GLFW::EmuWindow_GLFW() {
...
     }
-    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
-    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 2);
+    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 2);
+    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 1);
     // GLFW on OSX requires these window hints to be set to create a 3.2+ GL context.
-    glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GL_TRUE);
-    glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
+    //glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GL_TRUE); // We should no longer be creating a 3.2 GL context.
+    //glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
```
lioncash: Just remove this code if it's not used anymore. There's no point in keeping dead code around.

kitling: You're right, just didn't want to *remove* anything yet when I was still trying to figure it out.
I will do that. Now, I guess.

kitling:
I don't. I don't even. Well, at least I can say that I expected this. Lets see what exactly is needed of me. Google~!...

yuriks:
You have to get rid of the `in` and `out`.

kitling:
Haha, just found that, too. Well, that's the thing, I guess.

kitling:
Long story short, mesa is too lenient, it looks like. Now, assuming I didn't break everyone because of this assumption...

It looks like, in GLSL 1.2, fragment shaders have no custom output. So I changed the shader to use the built in variable. But I have yet to investigate elsewhere, and will be back tomorrow. In the mean time...

purpasmart96:
I'm not so sure about this...

kitling:
So, appearantly, it works perfectly fine, and I can't find any refrences to the "color" output variable. Though, I could be not finding it.

I agree, @purpasmart96 , if you're referring to how tenative I'm being about this, and such things. I am messing with things that I have had little experience with. I hate being so "Uhh, did I break it?" but I don't know. That said, I love working on this, because I'm learning.

Does anyone spot any problems? Questions? I'm not sure what to look for next...

purpasmart96:
Loosing vertex array objects... bleah... Is this really worth it just so a couple more toasters can run citra? At this point we might as well use the deprecated fixed function stuff in open gl since it was so much more easy work with... but I honestly  think 3.2 is fine as it is.

archshift:
Confirmed to work on OS X now.

kitling:
@purpasmart96 This effort is only needed once, and I'm clumsy here, which is why it's taking so much effort. And yes, it is worth it for me to be able to run citra. If my laptop had support for OpenGL 3.2, then there would be a slightly different story, but I'd still probably do it. Yes, it's worth it, because we don't even need anything from 3.x. VAOs were only even there because 3.whatever required it. So in the end, some code was simplified, and more people can run it.

chinhodado:
Thanks @kitling for doing this. linkmauve attempted something similar in #213 but didn't finish it.

purpasmart96:
@chinhodado  Except this is only limited to one open gl version, the other one wasn't. So everyone is going to be forced to run through the old 2.1 profile...

kitling:
@purpasmart96 I'd like to know what the disadvantage of running everything through the OpenGL 2.1 profile is. I can try to add an option to change what OpenGL context is used, but it seems to me that the only change is what functions are available. So, there will be no difference unless we take advantage of newer features. What disadvantage are you seeing by only using the version that we really need? Until we make use of newer features, I see no point in requiring a newer profile/context. Unless there is a real disadvantage to do so...?

kitling:
So, it appears that this works everywhere, besides some randomness for @yuriks linux guest in VBox.

So... Does it look good?

yuriks:
LGTM

purpasmart96:
This is also going to make tfarely have to redo his entire pr if this is merged...

kitling:
@purpasmart96 I think that it's fair that if I break a nice PR with something else, I should help out with fixing it. And as for the entire PR... Nah. Unless there are functions that he uses that require OpenGL 3.0 or up, the only thing that would need to be changed in his PR is to make the HW shaders compatible with GLSL 1.2. There may need to be a conflict that'll need to be reconciled, but other than that...?

That is, unless I've gotten something horribly wrong. In which case: please correct me.

yuriks:
@kitling: Actually, while I don't think the GL calls would be a big challenge, the shader might prove quite hard to make compatible with 2.1, since it's very complex.

neobrain:
What's the status of this PR?

kitling:
Status? Old. I stopped working on it when, after (trying) to port tfarley's shaders to 2.1, my computer complained about too many instructions. And then I haven't taken a look in a good while, unfortunately. I'm going to close this PR for now, and open a new one when I get around to thinking about it... :/

linkmauve:
I’d recommend you to keep the hw rasterizer a 3.2-exclusive, and only port the quad blitting part to 2.1.

It’s what I did in my GLES 2.0 branch (actually starting from before the hw rasterizer got merged).

PR #733, opened 2015-05-09
Loader: Remove .bin file support

yuriks:
It is of very limited practical utility currently, and will soon be
impossible to support due to more accurate memory map emulation.


bunnei:
LGTM. Feel free to merge once builds pass.

PR #736, opened 2015-05-09
Common: Remove the BIT macro

yuriks:
When the macro was introduced in 326ec51261299e48de97592631c02523da9c8118
it wasn't noticed that it conflicted in name with a heavily used macro
inside of dyncom. This causes some compiler warnings. Since it's only
lightly used, it was opted to simply remove the new macro.


archshift:
LGTM

linkmauve:
LGTM.

PR #738, opened 2015-05-10
Implementation of http:C service.

archshift:
Early implementation, but it's enough to run DownloadMii and get Nintendo Zone farther (although I suspect it depends on SSL, which will be a later PR).

~~Note that the cURL submodule does not currently work and as such the CMake code for it is commented out. I still have to figure out a good way to integrate that.~~

~~Also, lots of hwtests to be done to verify errors or certain behaviors.~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 # CMake 2.8.11 required for Qt5 settings to be applied automatically on
 # dependent libraries.
 cmake_minimum_required(VERSION 2.8.11)
+include(ExternalProject)
 
 project(citra)
 
@@ -123,7 +124,7 @@ IF (APPLE)
     FIND_LIBRARY(IOKIT_LIBRARY IOKit)           # GLFW dependency
     FIND_LIBRARY(COREVIDEO_LIBRARY CoreVideo)   # GLFW dependency
     set(PLATFORM_LIBRARIES iconv ${COCOA_LIBRARY} ${IOKIT_LIBRARY} ${COREVIDEO_LIBRARY})
-    
+
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -stdlib=libc++")
 ELSEIF(MINGW)
@@ -184,6 +185,26 @@ git_branch_name(GIT_BRANCH)
 set(INI_PREFIX "${CMAKE_CURRENT_SOURCE_DIR}/externals/inih")
 include_directories(${INI_PREFIX})
 add_subdirectory(${INI_PREFIX})
+
+find_package(CURL 7.28.0)
+if (NOT CURL_FOUND)
+    set(CURL_SOURCE_PREFIX "${CMAKE_CURRENT_SOURCE_DIR}/externals/curl")
+    set(CURL_BUILT_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/externals/curl")
+    externalproject_add(
+        PREFIX SOURCE_DIR ${CURL_SOURCE_PREFIX}
+        CMAKE_CACHE_ARGS
+          -DBUILD_CURL_EXE:string=off
+          -DBUILD_CURL_TESTS:string=off
+          -DHTTP_ONLY:string=on
+          -DENABLE_MANUAL:string=off
+          -DCMAKE_INSTALL_PREFIX:string=${CURL_BUILT_PREFIX}
+        )
+    include_directories("${CURL_BUILT_PREFIX}/include")
+    link_directories("${CURL_BUILT_PREFIX}/lib")
+    set(CURL_LIBRARIES "curl")
+else()
+    include_directories(${CURL_INCLUDE_DIRS})
+endif()
 
 include_directories(externals/nihstro/include)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,7 +190,7 @@ find_package(CURL 7.28.0)
 if (NOT CURL_FOUND)
     set(CURL_SOURCE_PREFIX "${CMAKE_CURRENT_SOURCE_DIR}/externals/curl")
     set(CURL_BUILT_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/externals/curl")
-    externalproject_add(
+    externalproject_add(curl
         PREFIX SOURCE_DIR ${CURL_SOURCE_PREFIX}
         CMAKE_CACHE_ARGS
           -DBUILD_CURL_EXE:string=off
@@ -201,7 +201,11 @@ if (NOT CURL_FOUND)
         )
     include_directories("${CURL_BUILT_PREFIX}/include")
     link_directories("${CURL_BUILT_PREFIX}/lib")
-    set(CURL_LIBRARIES "curl")
+    if (NOT MSVC)
+        set(CURL_LIBRARIES "curl")
+    else()
+        set(CURL_LIBRARIES "libcurl_imp")
+    endif()
 else()
     include_directories(${CURL_INCLUDE_DIRS})
 endif()
--- a/src/citra/CMakeLists.txt
+++ b/src/citra/CMakeLists.txt
@@ -16,6 +16,6 @@ create_directory_groups(${SRCS} ${HEADERS})
 add_executable(citra ${SRCS} ${HEADERS})
 target_link_libraries(citra core common video_core)
 target_link_libraries(citra ${GLFW_LIBRARIES} ${OPENGL_gl_LIBRARY} inih)
-target_link_libraries(citra ${PLATFORM_LIBRARIES})
+target_link_libraries(citra ${PLATFORM_LIBRARIES} ${CURL_LIBRARIES})
 
 #install(TARGETS citra RUNTIME DESTINATION ${bindir})
--- a/src/core/hle/service/http/http.cpp
+++ b/src/core/hle/service/http/http.cpp
@@ -39,8 +39,7 @@ static CURLcode SetConnectionType(CURL* connection, RequestType type) {
     case REQUEST_TYPE_PUT_:
         return curl_easy_setopt(connection, CURLOPT_UPLOAD, 1);
     case REQUEST_TYPE_DELETE:
-        // TODO
-        break;
+        break; // TODO
     case REQUEST_TYPE_HEAD:
         return curl_easy_setopt(connection, CURLOPT_NOBODY, 1);
     }
@@ -90,7 +89,7 @@ void MakeRequest(HttpContext* context) {
 
         {
             std::lock_guard<std::mutex> lock(context->mutex);
-            if (context->should_quit)
+            if (context->cancel_request)
                 break;
         }
 
@@ -127,7 +126,7 @@ void Init() {
 
 void ClearInstance() {
     for (auto& pair : context_map) {
-        pair.second->should_quit = true;
+        pair.second->cancel_request = true;
         if (pair.second->req_thread != nullptr)
             pair.second->req_thread->join();
     }
--- a/src/core/hle/service/http/http.cpp
+++ b/src/core/hle/service/http/http.cpp
@@ -1,0 +1,119 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <curl/curl.h>
+
+#include "common/string_util.h"
+
+#include "core/hle/service/http/http.h"
+#include "core/hle/service/http/http_c.h"
+#include "core/hle/service/service.h"
+
+namespace Service {
+namespace HTTP {
+
+std::unordered_map<ContextHandle, std::unique_ptr<HttpContext>> context_map;
+ContextHandle next_handle;
+
+static int WriteBuf(u8 *data, size_t size, size_t nmemb, std::mutex *mutex, std::vector<u8> *buffer) {
+    std::lock_guard<std::mutex> lock(*mutex);
+    buffer->reserve(buffer->size() + size * nmemb);
+    buffer->insert(buffer->end(), data, data + size * nmemb);
+    return static_cast<int>(size * nmemb);
+}
+
+static int HdrWriter(u8 *data, size_t size, size_t nmemb, HttpContext *context) {
+    if (context == nullptr || context->should_quit)
+        return -1;
+    return WriteBuf(data, size, nmemb, &context->mutex, &context->response_hdrs);
+}
+
+static int DataWriter(u8 *data, size_t size, size_t nmemb, HttpContext *context) {
+    if (context == nullptr || context->should_quit)
+        return -1;
+    return WriteBuf(data, size, nmemb, &context->mutex, &context->response_data);
+}
+
+HttpContext::~HttpContext() {
+    curl_slist_free_all(request_hdrs);
+}
+
+void MakeRequest(HttpContext* context) {
+    CURL* connection = curl_easy_init();
+    CURLcode res;
+    {
+        std::lock_guard<std::mutex> lock(context->mutex);
+
+        context->state = REQUEST_STATE_IN_PROGRESS;
+        res = curl_easy_setopt(connection, CURLOPT_URL, context->url.c_str());
+        switch (context->req_type) {
+            case REQUEST_TYPE_GET:
+                res = curl_easy_setopt(connection, CURLOPT_HTTPGET, 1);
+                break;
+            case REQUEST_TYPE_POST:
+            case REQUEST_TYPE_POST_:
+                res = curl_easy_setopt(connection, CURLOPT_POST, 1);
+                break;
+            case REQUEST_TYPE_PUT:
+            case REQUEST_TYPE_PUT_:
+                res = curl_easy_setopt(connection, CURLOPT_UPLOAD, 1);
+                break;
+            case REQUEST_TYPE_DELETE:
+                // TODO
+                break;
+            case REQUEST_TYPE_HEAD:
+                res = curl_easy_setopt(connection, CURLOPT_NOBODY, 1);
+                break;
+        }
+
+        res = curl_easy_setopt(connection, CURLOPT_HEADERFUNCTION, HdrWriter);
+        res = curl_easy_setopt(connection, CURLOPT_WRITEFUNCTION, DataWriter);
+        res = curl_easy_setopt(connection, CURLOPT_HEADERDATA, context);
+        res = curl_easy_setopt(connection, CURLOPT_WRITEDATA, context);
+        res = curl_easy_setopt(connection, CURLOPT_HTTPHEADER, context->request_hdrs);
+    }
+
+    res = curl_easy_perform(connection);
+    {
+        std::lock_guard<std::mutex> lock(context->mutex);
+
+        res = curl_easy_getinfo(connection, CURLINFO_RESPONSE_CODE, &context->response_code);
+        res = curl_easy_getinfo(connection, CURLINFO_SIZE_DOWNLOAD, &context->downloaded_size);
+        res = curl_easy_getinfo(connection, CURLINFO_CONTENT_LENGTH_DOWNLOAD, &context->download_size);
+
+        context->state = REQUEST_STATE_READY;
+    }
+
+    curl_easy_cleanup(connection);
+}
+
+void AddRequestHeader(std::string name, std::string value, curl_slist** hdr_list) {
+    std::string header = Common::StringFromFormat("%s: %s", name.c_str(), value.c_str());
+    *hdr_list = curl_slist_append(*hdr_list, header.c_str());
+}
+
+void Init() {
+    AddService(new HTTP_C_Interface);
+
+    curl_global_init(CURL_GLOBAL_DEFAULT);
+}
+
+void ClearInstance() {
+    for (auto& pair : context_map) {
+        pair.second->should_quit = true;
+        if (pair.second->req_thread != nullptr)
+            pair.second->req_thread->join();
+    }
+
+    context_map.clear();
+    next_handle = 0;
+}
+
+void Shutdown() {
+    curl_global_cleanup();
+    ClearInstance();
+}
+
+} // namespace HTTP
+} // namespace Service
--- a/src/core/hle/service/http/http.cpp
+++ b/src/core/hle/service/http/http.cpp
@@ -24,6 +24,16 @@ static int BufWriter(u8 *data, size_t size, size_t nmemb, std::vector<u8>* out_b
     return static_cast<int>(size * nmemb);
 }
 
+HttpContext::HttpContext() {
+    state = REQUEST_STATE_NONE;
+    cancel_request = false;
+    req_type = REQUEST_TYPE_NONE;
+    request_hdrs = nullptr;
+    response_code = 0;
+    content_length = 0.0;
+    downloaded_size = 0.0;
+}
+
 HttpContext::~HttpContext() {
     curl_slist_free_all(request_hdrs);
 }
@@ -68,8 +78,8 @@ void MakeRequest(HttpContext* context) {
 
     mres = curl_multi_add_handle(manager, connection);
 
-    int still_running;
-    int repeats;
+    int still_running = 0;
+    int repeats = 0;
     mres = curl_multi_perform(manager, &still_running);
 
     do {
--- a/src/core/hle/service/http/http.cpp
+++ b/src/core/hle/service/http/http.cpp
@@ -25,7 +25,7 @@ static int BufWriter(u8 *data, size_t size, size_t nmemb, std::vector<u8>* out_b
 }
 
 HttpContext::HttpContext() {
-    state = REQUEST_STATE_NONE;
+    state = REQUEST_STATE_NOT_STARTED;
     cancel_request = false;
     req_type = REQUEST_TYPE_NONE;
     request_hdrs = nullptr;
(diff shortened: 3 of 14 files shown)


archshift:
This has gone through notable revisions, so it's best to look at the entire thing versus the individual commits.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,119 @@
...
+        res = curl_easy_getinfo(connection, CURLINFO_SIZE_DOWNLOAD, &context->downloaded_size);
+        res = curl_easy_getinfo(connection, CURLINFO_CONTENT_LENGTH_DOWNLOAD, &context->download_size);
+
+        context->state = REQUEST_STATE_READY;
+    }
+
+    curl_easy_cleanup(connection);
+}
+
+void AddRequestHeader(std::string name, std::string value, curl_slist** hdr_list) {
```
lioncash: const std::string&

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,277 @@
+// Copyright 2014 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <stdexcept>
```
lioncash: This isn't necessary as far as I can tell.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,119 @@
...
+    curl_slist_free_all(request_hdrs);
+}
+
+void MakeRequest(HttpContext* context) {
+    CURL* connection = curl_easy_init();
+    CURLcode res;
+    {
+        std::lock_guard<std::mutex> lock(context->mutex);
+
+        context->state = REQUEST_STATE_IN_PROGRESS;
```
archshift: Should probably be moved to BeginRequest to avoid race conditions (however unlikely).

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,74 @@
...
+    REQUEST_TYPE_PUT_   = 7
+};
+
+enum RequestState : u32 {
+    REQUEST_STATE_DEFAULT, // TODO
+    REQUEST_STATE_IN_PROGRESS = 5,
+    REQUEST_STATE_READY = 8,
+};
+
+struct HttpContext {
```
neobrain: Is it clear what a "http context" actually is? if not, could we have some documentation explaining what this struct's purpose is?

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,74 @@
...
+    std::mutex mutex;
+    bool should_quit;
```
neobrain: Could we have documentation on what exactly "should quit" here? Also, consider using `std::atomic<bool>` if this is accessed across threads.

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,277 @@
...
+}
+
+static void CreateContext(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 url_len = cmd_buff[1];
+    RequestType req_type = static_cast<RequestType>(cmd_buff[2]);
+    u32 url_ptr = cmd_buff[4];
+
+    std::string url(reinterpret_cast<const char*>(Memory::GetPointer(url_ptr)), url_len);
```
neobrain: Is there any need to add constness here?

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+}
+
+static void CreateContext(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 url_len = cmd_buff[1];
+    RequestType req_type = static_cast<RequestType>(cmd_buff[2]);
+    u32 url_ptr = cmd_buff[4];
```
neobrain: bad variable name: this is an address, not a pointer.

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    std::string url(reinterpret_cast<const char*>(Memory::GetPointer(url_ptr)), url_len);
+
+    LOG_DEBUG(Service, "request url=%s req_type=%u",
+              url.c_str(), req_type);
+
+    // TODO: give HttpContext a proper constructor
+    std::unique_ptr<HttpContext> context(new HttpContext{});
```
neobrain: Don't we have a make_unique implementation in common?

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,277 @@
...
+    HttpContext* response = map_it->second.get();
+    std::lock_guard<std::mutex> lock(response->mutex);
+
+    // TODO: Request not done yet
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    cmd_buff[2] = (u32) response->downloaded_size;
+    // TODO: invalid content-length?
+    cmd_buff[3] = (u32) response->download_size;
+    return;
```
neobrain: Unnecessary

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,277 @@
...
+    auto map_it = context_map.find(handle);
+    if (map_it == context_map.end()) {
+        cmd_buff[1] = -1; // TODO: Find proper result code for invalid handle
+        return;
+    }
+
+    // TODO: Verify behavior
+    while (true) {
+        std::lock_guard<std::mutex> lock(map_it->second->mutex);
+        if (map_it->second->state == REQUEST_STATE_READY)
```
neobrain: This has potential for introducing uninterruptable lookups in the whole program :/

archshift: Indeed, which is why I was baffled to learn that apparently `GetResponseStatusCode` waits on the request to finish. I don't know if there's an intrinsic timeout or something, though, which is why the behavior needs to be verified.

Subv: Pasting this from IRC, just a thought
```
<@Subv> archshift: https://github.com/citra-emu/citra/pull/738#discussion_r30001147 schedule a CoreTiming event to check the response (or can you specify a callback function within curl?) and just sleep the calling ctr thread and reschedule
```

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,74 @@
...
+
+    std::string url;
+    RequestType req_type;
+    curl_slist* request_hdrs;
+
+    std::vector<u8> response_hdrs;
+    std::vector<u8> response_data;
+    long response_code;
+    double download_size;
+    double downloaded_size;
```
bunnei: I don't like how similarly named these two variables are. Also, please consider adding some more documentation on what all the members of this struct mean.

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,277 @@
...
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 url_len = cmd_buff[1];
+    RequestType req_type = static_cast<RequestType>(cmd_buff[2]);
+    u32 url_ptr = cmd_buff[4];
+
+    std::string url(reinterpret_cast<const char*>(Memory::GetPointer(url_ptr)), url_len);
+
+    LOG_DEBUG(Service, "request url=%s req_type=%u",
+              url.c_str(), req_type);
```
bunnei: Unnecessary line break

src/citra/CMakeLists.txt
```diff
@@ -16,6 +16,6 @@ create_directory_groups(${SRCS} ${HEADERS})
 add_executable(citra ${SRCS} ${HEADERS})
 target_link_libraries(citra core common video_core)
 target_link_libraries(citra ${GLFW_LIBRARIES} ${OPENGL_gl_LIBRARY} inih)
-target_link_libraries(citra ${PLATFORM_LIBRARIES})
+target_link_libraries(citra ${PLATFORM_LIBRARIES} curl)
```
yuriks: What's with the somewhat arbitrary separation of these lines? Why does curl go here and not above?

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+namespace Service {
+namespace HTTP {
+
+std::unordered_map<ContextHandle, std::unique_ptr<HttpContext>> context_map;
+ContextHandle next_handle;
+
+static int WriteBuf(u8 *data, size_t size, size_t nmemb, std::mutex *mutex, std::vector<u8> *buffer) {
+    std::lock_guard<std::mutex> lock(*mutex);
+    buffer->reserve(buffer->size() + size * nmemb);
```
yuriks: You don't need this reserve.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,119 @@
...
+    buffer->insert(buffer->end(), data, data + size * nmemb);
+    return static_cast<int>(size * nmemb);
+}
+
+static int HdrWriter(u8 *data, size_t size, size_t nmemb, HttpContext *context) {
+    if (context == nullptr || context->should_quit)
+        return -1;
+    return WriteBuf(data, size, nmemb, &context->mutex, &context->response_hdrs);
```
yuriks: Extra function is unnecessary abstraction imo. Just put the one line in here.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+static int WriteBuf(u8 *data, size_t size, size_t nmemb, std::mutex *mutex, std::vector<u8> *buffer) {
+    std::lock_guard<std::mutex> lock(*mutex);
+    buffer->reserve(buffer->size() + size * nmemb);
+    buffer->insert(buffer->end(), data, data + size * nmemb);
+    return static_cast<int>(size * nmemb);
+}
+
+static int HdrWriter(u8 *data, size_t size, size_t nmemb, HttpContext *context) {
+    if (context == nullptr || context->should_quit)
```
yuriks: You're accessing `should_quit` (which is not an atomic variable) without locking context.

src/core/hle/service/http/http.h
yuriks: I think any cross access of this is a bug. There's already a mutex protecting the whole struct, so this shouldn't be atomic.

src/core/hle/service/http/http.h
yuriks: Agreed. Maybe "total_size" for the other one.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,144 @@
...
+    do {
+        // Number of file descriptors. If this is zero, nothing happened to
+        // the active connection.
+        int numfds;
+
+        if (curl_multi_wait(manager, NULL, 0, 1000, &numfds) != CURLM_OK)
+            break;
+
+        if (numfds == 0) {
+            if (repeats++ > 1)
```
darkf: I think you meant to initialize `repeats` here (otherwise this would be a garbage value)?

archshift: Initialized at the declaration.

archshift:
At this point I'm considering this PR out of RFC. As such, can I get more thorough reviews?

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    ContextHandle handle = static_cast<ContextHandle>(cmd_buff[1]);
+
+    auto map_it = context_map.find(handle);
+    if (map_it == context_map.end()) {
+        cmd_buff[1] = 0xD8E007F7;
+        return;
+    }
+
+    // TODO: Verify behavior
```
archshift: Remove this comment; the behavior has been verified.

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,274 @@
...
+}
+
+static void CreateContext(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 url_len = cmd_buff[1];
+    RequestType req_type = static_cast<RequestType>(cmd_buff[2]);
+    char* url_ptr = reinterpret_cast<char*>(Memory::GetPointer(cmd_buff[4]));
+
+    LOG_DEBUG(Service, "request url=%s req_type=%u", url_ptr, req_type);
```
archshift: Don't use the `Service` log module; add the proper one for HTTP.

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,94 @@
...
+    std::string url;                  //< URL to the target server.
+    RequestType req_type;             //< Type of request that will be performed
+    curl_slist* request_hdrs;         //< Buffer for request headers to be sent to the server.
+
+    //--- Response data
+    std::vector<u8> response_hdrs;    //< Buffer for response headers returned by the server.
+    std::vector<u8> response_data;    //< Buffer for response body returned by the server.
+    long response_code;               //< The three-digit HTTP response code returned by the server.
+    double content_length;            //< The total size in bytes that will be downloaded this request.
+    double downloaded_size;           //< The amount in bytes that has been downloaded so far.
```
yuriks: "double"? wtf?

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "core/hle/service/http/http_c.h"
+#include "core/hle/service/service.h"
+
+namespace Service {
+namespace HTTP {
+
+std::unordered_map<ContextHandle, std::unique_ptr<HttpContext>> context_map;
+ContextHandle next_handle;
+
+static int BufWriter(u8 *data, size_t size, size_t nmemb, std::vector<u8>* out_buffer) {
```
yuriks: This name should be a verb

src/core/hle/service/http/http.h
archshift: They're `double`s because that's the type cURL gives you when you ask for them.

yuriks: wtfwtfwtfwtfwtfwtfwtfwtfwtf
At least leave a comment here to warn future readers that this is a curl induced wtf.
wtfwtfwtfwtfwtfwtfwtfwtfwtf

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,94 @@
...
+/// HTTP operation that will be performed by the request (API-exposed).
+enum RequestType : u32 {
+    REQUEST_TYPE_NONE   = 0,
+    REQUEST_TYPE_GET    = 1,
+    REQUEST_TYPE_POST   = 2,
+    REQUEST_TYPE_HEAD   = 3,
+    REQUEST_TYPE_PUT    = 4,
+    REQUEST_TYPE_DELETE = 5,
+    REQUEST_TYPE_POST_  = 6,
+    REQUEST_TYPE_PUT_   = 7
```
yuriks: What's with the underscored members?

archshift: This enum is API-exposed, and the 3DS API treats various values the same. These duplicates are denoted with an underscore at the end.

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,94 @@
...
+enum RequestState : u32 {
+    REQUEST_STATE_NOT_STARTED = 1,
+    REQUEST_STATE_IN_PROGRESS = 5,
+    REQUEST_STATE_READY = 8,
+    REQUEST_STATE_TIMEOUT = 10,
+};
+
+/** A helper struct that contains all the required information for a HTTP request.
+ * This struct, as it can be accessed by multiple threads, contains a mutex that
+ * must be locked whenever any of the struct's members are being accessed. This is
```
yuriks: As far as I could tell, there are many members which can be used without locking the mutex under some circumstances. (The download buffers for example.)

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,94 @@
...
+ * to prevent any possible race conditions.
+ */
+struct HttpContext {
+    std::mutex mutex;
+
+    //--- Ongoing request management
+    RequestState state;               //< API-exposed current state of the HTTP request
+    std::unique_ptr<std::thread> req_thread;
```
yuriks: Why the unique_ptr?

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,94 @@
...
+    bool cancel_request;              //< True if the request's thread should be canceled ASAP
```
yuriks: "request thread"*

src/core/hle/service/http/http.h
archshift: How? The download buffers are used by both the curl output and the `ReceiveData` function, potentially at the same time.

src/core/hle/service/http/http.h
yuriks: But I'd describe it was "Signals the request thread to exit as soon as possible."

src/core/hle/service/http/http.h
archshift: You're right, I should probably use `std::move` instead.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,159 @@
...
+void AddRequestHeader(const std::string& name, const std::string& value, curl_slist** hdr_list) {
+    // TODO: header value is empty
+    std::string header = Common::StringFromFormat("%s: %s", name.c_str(), value.c_str());
+    *hdr_list = curl_slist_append(*hdr_list, header.c_str());
+}
+
+void Init() {
+    AddService(new HTTP_C_Interface);
+
+    curl_global_init(CURL_GLOBAL_DEFAULT);
```
yuriks: Why not do this inside the Initialize call like you used to?

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+void ClearInstance() {
+    for (auto& pair : context_map) {
+        pair.second->cancel_request = true;
+        if (pair.second->req_thread != nullptr)
+            pair.second->req_thread->join();
+    }
```
yuriks: You should split this loop in two: one that sets the flags for all threads, and then another one that joins on all threads. This should speed up shutdown somewhat if you have several connections.

src/core/hle/service/http/http.cpp
archshift: I think it'd make more sense for the library itself to remain initialized independent of the client program.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    context_map.clear();
+    next_handle = 0;
+}
+
+void Shutdown() {
+    curl_global_cleanup();
+    ClearInstance();
```
yuriks: The order of these two calls should be inverted, otherwise you might shutdown curl while there are still active threads.

src/core/hle/service/http/http.cpp
yuriks: And for me it seems to be a bit pointless for the HTTP library to always be initialized even if the application's never going to use it.

src/core/hle/service/http/http.h
yuriks: Hmm. Add a comment explaining this on these enums. I'd also use a less subtle name like an `_ALT` suffix perhaps.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+void Init() {
+    AddService(new HTTP_C_Interface);
+
+    curl_global_init(CURL_GLOBAL_DEFAULT);
+}
+
+void ClearInstance() {
+    for (auto& pair : context_map) {
+        pair.second->cancel_request = true;
```
yuriks: I think you're supposed to lock the struct before accesing this.

src/core/hle/service/http/http.h
yuriks: You use this in a lot of places without locking the struct first. Either ensure you lock the struct around all accesses or make this atomic.

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,274 @@
...
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    ContextHandle handle = static_cast<ContextHandle>(cmd_buff[1]);
+    auto map_it = context_map.find(handle);
+    if (map_it == context_map.end()) {
+        cmd_buff[1] = 0xD8E007F7;
+        return;
+    }
+
+    map_it->second->cancel_request = true;
```
yuriks: Didn't lock first.

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,274 @@
...
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    ContextHandle handle = static_cast<ContextHandle>(cmd_buff[1]);
+
+    auto map_it = context_map.find(handle);
+    if (map_it == context_map.end()) {
+        cmd_buff[1] = 0xD8E007F7;
+        return;
+    }
+    map_it->second->cancel_request = true;
```
yuriks: Didn't lock first.

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,274 @@
...
+    u8* buffer = Memory::GetPointer(cmd_buff[4]);
+
+    auto map_it = context_map.find(handle);
+    if (map_it == context_map.end()) {
+        cmd_buff[1] = 0xD8E007F7;
+        return;
+    }
+
+    std::lock_guard<std::mutex> lock(map_it->second->mutex);
+    const std::vector<u8>& data = map_it->second->response_data;
```
yuriks: Use `auto&` here? I don't think explicitly spelling out the type adds anything.

src/core/hle/service/http/http_c.cpp
yuriks: Agreed with @Subv, this should reschedule to another thread while it waits for the connection to finish, which is what would happen in a real system, not a complete program block.

yuriks: Also, this is a busy wait. NEVER DO THIS. Use a condvar instead.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+        if (numfds == 0) {
+            // We allow one repeat try--after which, we start to sleep the
+            // thread until the timeout runs out.
+            if (repeats++ > 1)
+                std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        } else {
+            // While the file descriptor is responding, there's no need to check for repeats.
+            repeats = 0;
+        }
```
yuriks: I understood what this loop was doing, but it took me a while, you should add a comment explaining the general objective here.

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+        // Set the timeout to 1sec.
+        if (curl_multi_wait(manager, NULL, 0, 1000, &numfds) != CURLM_OK)
+            break;
+
+        if (numfds == 0) {
+            // We allow one repeat try--after which, we start to sleep the
+            // thread until the timeout runs out.
+            if (repeats++ > 1)
+                std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
yuriks: Is there any necessity for this sleep? Isn't it better to just go straight back to curl? Is there anything this achieves that you can't more easily do by setting the timeout in `curl_multi_wait` to 2100?

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+namespace Service {
+namespace HTTP {
+
+std::unordered_map<ContextHandle, std::unique_ptr<HttpContext>> context_map;
+ContextHandle next_handle;
+
+static int BufWriter(u8 *data, size_t size, size_t nmemb, std::vector<u8>* out_buffer) {
+    if (out_buffer == nullptr)
+        return -1;
```
yuriks: Will `out_buffer` ever be null? Shouldn't you assert this instead?

yuriks:
I think that's enough for a pass. One general comment I have is that, if you're going through the trouble of using curl's "multi" API, then why not simply handle all connections in a single thread instead of spreading them over multiple threads like currently? In addition to creating less threads, this should ease the blocking concerns, since now the program can only block waiting for a thread shutdown in the cleanup function (in other instances you'll simply remove a connection from the working set).

CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 # CMake 2.8.11 required for Qt5 settings to be applied automatically on
 # dependent libraries.
 cmake_minimum_required(VERSION 2.8.11)
+include(ExternalProject)
```
neobrain: Does this actually exist on 2.8.11?

src/core/hle/service/http/http.cpp
```diff
@@ -0,0 +1,159 @@
...
+    context_map.clear();
+    next_handle = 0;
+}
+
+void Shutdown() {
+    curl_global_cleanup();
+    ClearInstance();
+}
+
+}; // namespace HTTP
```
neobrain: no need for ";"

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,94 @@
...
+    REQUEST_TYPE_POST   = 2,
+    REQUEST_TYPE_HEAD   = 3,
+    REQUEST_TYPE_PUT    = 4,
+    REQUEST_TYPE_DELETE = 5,
+    REQUEST_TYPE_POST_  = 6,
+    REQUEST_TYPE_PUT_   = 7
+};
+
+/// Current state of the HTTP request (API-exposed).
+enum RequestState : u32 {
```
neobrain: Instead of prefixing all enum values with the enum name, just use an enum class.

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,94 @@
...
+ */
+struct HttpContext {
+    std::mutex mutex;
+
+    //--- Ongoing request management
+    RequestState state;               //< API-exposed current state of the HTTP request
+    std::unique_ptr<std::thread> req_thread;
+    bool cancel_request;              //< True if the request's thread should be canceled ASAP
+
+    //--- Request data
```
neobrain: Just curious, is `//---` recognized by doxygen in some special way?

src/core/hle/service/http/http.h
```diff
@@ -0,0 +1,94 @@
...
+    std::string url;                  //< URL to the target server.
+    RequestType req_type;             //< Type of request that will be performed
+    curl_slist* request_hdrs;         //< Buffer for request headers to be sent to the server.
+
+    //--- Response data
+    std::vector<u8> response_hdrs;    //< Buffer for response headers returned by the server.
```
neobrain: Is it really necessary to save 3 letters here by not saying "headers"?

src/core/hle/service/http/http_c.cpp
```diff
@@ -0,0 +1,274 @@
...
+    if (map_it == context_map.end()) {
+        cmd_buff[1] = 0xD8E007F7;
+        return;
+    }
+
+    HttpContext* response = map_it->second.get();
+    std::lock_guard<std::mutex> lock(response->mutex);
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    cmd_buff[2] = (u32) response->downloaded_size;
```
neobrain: Prefer `static_cast` over C-style casts, please.

Subv:
What's the status of this PR?

archshift:
Waiting on me to find an opportunity to finish it, I suppose.

src/core/hle/service/http/http_c.cpp
archshift: Invalidated by new style guide, although there shouldn't be a space there

src/core/hle/service/http/http.h
archshift: Nope, I just used it as a delimiter to separate the struct into more logical pieces.

CMakeLists.txt
archshift: [Yes](http://www.kitware.com/media/html/BuildingExternalProjectsWithCMake2.8.html).

archshift:
Superseded by #1071.

PR #739, opened 2015-05-10
[RFC] Identify games that use procedural textures

bunnei:
I'm looking for some assistance finding games that use procedural textures for reverse engineering. Developers, please ignore this, I'm not looking for a review. Instead, I'd like for testers that have a lot of games (and/or time to test) to run this branch on their game collections to help me find games that use procedural textures.

If a game tries to use a procedural texture, you'll immediately see the following message box (only once per game):

image

**This will only work with with the Qt GUI.** Furthermore, any triangles that use this texture should appear bright magenta-colored in-game.

Any help is greatly appreciated. Furthermore, any screenshots of where the procedural textures occur (and/or how to reproduce the results) would be helpful. Thanks!

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -44,8 +44,12 @@ void EmuThread::run() {
             if (!was_active)
                 emit DebugModeLeft();
 
-            Core::RunLoop();
-
+            try {
+                Core::RunLoop();
+            } catch (const char* str) {
+                emit UserError(str);
+            }
+            
             was_active = running || exec_step;
             if (!was_active && !stop_run)
                 emit DebugModeEntered();
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -78,6 +78,8 @@ signals:
      * @warning When connecting to this signal from other threads, make sure to specify either Qt::QueuedConnection (invoke slot within the destination object's message thread) or even Qt::BlockingQueuedConnection (additionally block source thread until slot returns)
      */
     void DebugModeLeft();
+
+    void UserError(const char* str);
 };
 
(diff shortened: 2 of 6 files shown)


src/citra_qt/bootmanager.cpp
```diff
@@ -44,8 +44,12 @@ void EmuThread::run() {
             if (!was_active)
                 emit DebugModeLeft();
 
-            Core::RunLoop();
-
+            try {
+                Core::RunLoop();
+            } catch (const char* str) {
```
Subv: I love it

Subv:
LGTM, insta merge pls

src/citra_qt/bootmanager.cpp
neobrain: @Subv See, **this** is why we need exceptions! :D

bunnei: I actually do really like exceptions as a way to abort emulation and shutdown safely, or to notify the user with a GUI box (as I am here..). So yeah, we should do this.

bunnei:
NOTE: This will NOT work in conjunction with tfarley's OGL branches. Those will short circuit the rasterizer code with the procedural texture test exists. Please test with the software renderer in this branch.

bunnei:
We didn't find any games, and I'm not really working on this right now. Will resume at another time.

Subv:
#2213

JayFoxRox:
More importantly, we have a label for that now: https://github.com/citra-emu/citra/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Agpu%3Aproctex%20

Wether #2213 is a false positive will have to be tested. I could imagine it being a bug in the CPU JIT or something which accidentely calculates a wrong texture unit. As op claimed that it worked fine before.

PR #749, opened 2015-05-11
Thread: Correctly set main thread initial stack position

yuriks:
Apparently "top" really means top, and not "bottom" like I seemed to be thinking at the time.


Subv:
LGTM

bunnei:
LGTM

PR #750, opened 2015-05-11
Core/HLE: Implemented the SVCs GetProcessId and GetProcessIdOfThread

Subv:
I suggest we create a helper function called GetCurrentProcess, analogous to our GetCurrentThread.

### DIFF
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -115,8 +115,7 @@ SharedPtr<Object> HandleTable::GetGeneric(Handle handle) const {
     if (handle == CurrentThread) {
         return GetCurrentThread();
     } else if (handle == CurrentProcess) {
-        LOG_ERROR(Kernel, "Current process (%08X) pseudo-handle not supported", CurrentProcess);
-        return nullptr;
+        return g_current_process;
     }
 
     if (!IsValid(handle)) {
@@ -138,6 +137,7 @@ void Init() {
     Kernel::ThreadingInit();
     Kernel::TimersInit();
 
+    Process::next_process_id = 0;
     Object::next_object_id = 0;
 }
 
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -137,8 +137,10 @@ void Init() {
     Kernel::ThreadingInit();
     Kernel::TimersInit();
 
-    Process::next_process_id = 0;
     Object::next_object_id = 0;
+    // TODO(Subv): Start the process ids from 10 for now, as lower PIDs are
+    // reserved for low-level services
+    Process::next_process_id = 10;
 }
 
 /// Shutdown the kernel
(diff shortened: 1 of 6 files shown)


bunnei:
LGTM

src/core/hle/svc.cpp
```diff
@@ -424,6 +425,34 @@ static ResultCode ReleaseMutex(Handle handle) {
...
+    const SharedPtr<Kernel::Process> process = Kernel::g_handle_table.Get<Kernel::Process>(handle);
+    if (process == nullptr)
+        return ERR_INVALID_HANDLE;
+
+    *process_id = process->GetProcessId();
+    return RESULT_SUCCESS;
+}
+
+/// Get the ID of the process that owns the specified thread
+static ResultCode GetProcessIdOfThread(u32* process_id, Handle handle) {
```
yuriks: `handle` -> `thread_handle`

src/core/hle/svc.cpp
```diff
@@ -424,6 +425,34 @@ static ResultCode ReleaseMutex(Handle handle) {
     return RESULT_SUCCESS;
 }
 
+/// Get the ID of the specified process
+static ResultCode GetProcessId(u32* process_id, Handle handle) {
```
yuriks: `handle` -> `process_handle`

src/core/hle/svc.cpp
```diff
@@ -424,6 +425,34 @@ static ResultCode ReleaseMutex(Handle handle) {
...
+static ResultCode GetProcessIdOfThread(u32* process_id, Handle handle) {
+    LOG_TRACE(Kernel_SVC, "called thread=0x%08X", handle);
+
+    const SharedPtr<Kernel::Thread> thread = Kernel::g_handle_table.Get<Kernel::Thread>(handle);
+    if (thread == nullptr)
+        return ERR_INVALID_HANDLE;
+
+    const SharedPtr<Kernel::Process> process = thread->owner_process;
+    if (process == nullptr)
+        return ERR_INVALID_HANDLE;
```
yuriks: Does it make sense to return this? IMO process being null should be an assert.

src/core/hle/svc.cpp
```diff
@@ -424,6 +425,34 @@ static ResultCode ReleaseMutex(Handle handle) {
...
+
+    *process_id = process->process_id;
```
yuriks: `GetProcessId`

src/core/hle/kernel/process.h
```diff
@@ -55,6 +55,14 @@ class Process final : public Object {
     static const HandleType HANDLE_TYPE = HandleType::Process;
     HandleType GetHandleType() const override { return HANDLE_TYPE; }
 
+    static u32 next_process_id;
+
+    /*
+     * Gets the process' id
+     * @returns The process' id
+     */
+    u32 GetProcessId() const { return process_id; }
```
yuriks: This getter function doesn't really add any value.

src/core/hle/kernel/kernel.cpp
```diff
@@ -138,6 +137,7 @@ void Init() {
     Kernel::ThreadingInit();
     Kernel::TimersInit();
 
+    Process::next_process_id = 0;
```
yuriks: I think this should start at a higher number for now (like, say, 10) since the first few pids have hardcoded special meanings in several parts of the system.

Subv:
Updated

yuriks:
LGTM

PR #748, opened 2015-05-11
Core/Memory: Add TLS support for creating up to 300 threads

Subv:
The real kernel autogrows the TLS area as more threads are created, up to an unknown limit, but for our purposes this will suffice for now.
This fixes #743

PR #751, opened 2015-05-11
Thread: Remove the idle thread

yuriks:
This felt too easy, but I couldn't trigger any nullptr or assert crashes after running things in a debug build for a while, so I'm gonna call it good.

This preemptively fixes the crash issue @Subv would have created in #748

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -24,9 +24,9 @@ ARM_Interface*     g_sys_core = nullptr;  ///< ARM11 system (OS) core
 
 /// Run the core CPU loop
 void RunLoop(int tight_loop) {
-    // If the current thread is an idle thread, then don't execute instructions,
+    // If we don't have a currently active thread then don't execute instructions,
     // instead advance to the next event and try to yield to the next thread
-    if (Kernel::GetCurrentThread()->IsIdle()) {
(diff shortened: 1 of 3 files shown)


src/core/core.cpp
```diff
@@ -26,7 +26,7 @@ ARM_Interface*     g_sys_core = nullptr;  ///< ARM11 system (OS) core
 void RunLoop(int tight_loop) {
     // If the current thread is an idle thread, then don't execute instructions,
```
bunnei: Update comment?

yuriks:
Looks like this makes OoT get stuck on the title screen, so hold on.

Subv:
This patch fixes it, there are some cases when we'll request a reschedule but the current thread is still able to continue running, if this happens and there's no higher-priority thread available, we should just let the current thread continue
https://gist.github.com/Subv/8c1ce6c8866f8a90931c

yuriks:
Update with a fix based on @Subv's

Subv:
LGTM after @bunnei's comment is addressed

yuriks:
@Subv: way ahead of ya

Subv:
LGTM

bunnei:
LGTM

PR #752, opened 2015-05-11
vfp: Handle flush-to-zero mode.

lioncash:
A few games use this instead of being fully IEEE compliant (e.g. Cubic Ninja), along with Default NaN mode (which is already supported).

### DIFF
--- a/src/core/arm/skyeye_common/vfp/vfp_helper.h
+++ b/src/core/arm/skyeye_common/vfp/vfp_helper.h
@@ -35,6 +35,7 @@
 #include <cstdio>
 #include "common/common_types.h"
 #include "core/arm/skyeye_common/armdefs.h"
+#include "core/arm/skyeye_common/vfp/asm_vfp.h"
 
 #define do_div(n, base) {n/=base;}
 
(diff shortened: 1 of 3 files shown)


src/core/arm/skyeye_common/vfp/vfp_helper.h
```diff
@@ -294,6 +268,42 @@ static inline int vfp_single_type(vfp_single* s)
...
+// of the single-precision float mantissa with the 1. if necessary,
+// aligned to bit 30.
+static inline void vfp_single_unpack(vfp_single* s, s32 val, u32* fpscr)
+{
+    u32 significand;
+
+    s->sign = vfp_single_packed_sign(val) >> 16,
+    s->exponent = vfp_single_packed_exponent(val);
+
+    significand = (u32) val;
```
yuriks: Why not declare the variable here too?

src/core/arm/skyeye_common/vfp/vfp_helper.h
```diff
@@ -294,6 +268,42 @@ static inline int vfp_single_type(vfp_single* s)
...
+    significand = (significand << (32 - VFP_SINGLE_MANTISSA_BITS)) >> 2;
```
yuriks: And actually just roll it into this expression.

yuriks:
LGTM

bunnei:
LGTM

PR #754, opened 2015-05-11
NWM_UDS: Fix a typo in the nwm service port name

purpasmart96:


archshift:
Verified.

purpasmart96:
Nintendo isn't very consistent with their service names...

PR #755, opened 2015-05-11
dyncom: Stub MCRR and MRRC

lioncash:
There's no other coprocessor outside the VFP (which has its own VMOV variants) in which the MPCore can send/retrieve data from.
Stubbed so citra won't crash and burn on the odd chance someone actually tries to use these.

### DIFF
--- a/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
@@ -992,6 +992,14 @@ typedef struct _mcr_inst {
     unsigned int inst;
 } mcr_inst;
 
+typedef struct mcrr_inst {
+    unsigned int opcode_1;
+    unsigned int cp_num;
+    unsigned int crm;
+    unsigned int rt;
+    unsigned int rt2;
+} mcrr_inst;
+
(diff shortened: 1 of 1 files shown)


src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -1261,11 +1269,6 @@ static get_addr_fp_t get_calc_addr_op(unsigned int inst) {
 #define CHECK_RM            (inst_cream->Rm == 15)
 #define CHECK_RS            (inst_cream->Rs == 15)
 
-#define UNIMPLEMENTED_INSTRUCTION(mnemonic) \
```
Subv: Was this the last unimplemented ARM instruction in the dyncom? :O

lioncash: Yep

purpasmart96: Good now you can work on JITing the branch instructions :)

yuriks:
LGTM

Subv:
LGTM

bunnei:
LGTM

PR #756, opened 2015-05-11
PTM: Changed the ptm services  to be like the IR, HID, and APT services.

purpasmart96:
I hope Subv doesn't mind... Because I really like how for the HID services have all their stuff in one area, so you don't have to duplicate the same functions all over the place in the related services to it.

### DIFF
--- a/src/core/hle/service/ptm/ptm.h
+++ b/src/core/hle/service/ptm/ptm.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <array>
+#include "core/hle/service/service.h"
 #include "core/hle/result.h"
 
 namespace Service {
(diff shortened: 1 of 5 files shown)


Subv:
It's OK with me, but you'll have to change the doxygen comments in some of the functions you changed.

src/core/hle/service/ptm/ptm.h
```diff
@@ -41,19 +42,38 @@ struct GameCoin {
  * that controls another part of the HW.
  * @returns 1 if the battery is charging, and 0 otherwise.
```
Subv: Here, for example, this function doesn't return anything anymore.

purpasmart96: Will do.

purpasmart96:
Done

bunnei:
LGTM

PR #757, opened 2015-05-12
Core/Scheduling: Prepare the new priority in the thread queue when svcSetPriority is called

Subv:
Fixes a crash in swkbd, and could probably fix many other threading bugs

(Sadly it doesn't fix the MoFlex playback bug :( )


yuriks:
LGTM

purpasmart96:
LGTM

bunnei:
LGTM

PR #758, opened 2015-05-12
Common: Remove async logging

yuriks:
It provided a large increase in complexity of the logging system while
having a negligible performance impact: the usage patterns of the ring
buffer meant that each log contended with the logging thread, causing
it to effectively act as a synchronous extra buffering.

Also removed some broken code related to filtering of subclasses which
was broken since it was introduced. (Which means no one ever used that
feature anyway, since, 8 months later, no one ever complained.)


yuriks:
Updated with the assert change.

src/common/assert.h
```diff
@@ -28,19 +29,14 @@ static void assert_noinline_call(const Fn& fn) {
     exit(1); // Keeps GCC's mouth shut about this actually returning
 }
 
-// TODO (yuriks) allow synchronous logging so we don't need printf
 #define ASSERT(_a_) \
     do if (!(_a_)) { assert_noinline_call([] { \
-        fprintf(stderr, "Assertion Failed!\n\n  Line: %d\n  File: %s\n  Time: %s\n", \
-                     __LINE__, __FILE__, __TIME__); \
+        LOG_CRITICAL(Debug, "Assertion Failed!"); \
```
bunnei: Why are you getting rid of line/file/..? I find that stuff useful.

yuriks: All that info is already included in the message by the log system.

bunnei:
Woo lots of deletions!
Looks fine, except my one comment. Fwiw, is it worth checking if the performance impact is still negligible even without the SW renderer bottleneck?

src/common/assert.h
bunnei: Ah right...

linkmauve:
You should also remove the now-unused concurrent_ring_buffer.h

yuriks:
Updated

archshift:
LGTM if CI passes.

yuriks:
360 noscope merge

PR #760, opened 2015-05-12
GPU: add more fine grained profiling for vertex shader and rasterization

yuriks:


src/video_core/rasterizer.cpp
```diff
@@ -186,6 +187,8 @@ static int SignedArea (const Math::Vec2<Fix12P4>& vtx1,
     return Math::Cross(vec1, vec2).z;
 };
 
+Common::Profiling::TimingCategory rasterization_category("Rasterization");
```
neobrain: Should this be static?

Subv:
LGTM

src/video_core/rasterizer.cpp
yuriks: Good point

yuriks:
Updated

neobrain:
LGTM

PR #761, opened 2015-05-12
Core/ResourceLimits: Implemented the basic structure of ResourceLimits.

Subv:
Implemented svcs GetResourceLimit, GetResourceLimitCurrentValues and GetResourceLimitLimitValues.

Note that the resource limits do not currently keep track of used objects, since we have no way to distinguish between an object created by the application, and an object created by some HLE module once we're inside Kernel::T::Create.

If you see some place where we can start using resource limits, let me know and I'll implement it either here or in a future PR.

### DIFF
--- a/src/core/hle/kernel/kernel.h
+++ b/src/core/hle/kernel/kernel.h
@@ -46,7 +46,8 @@ enum class HandleType : u32 {
     Process         = 8,
     AddressArbiter  = 9,
     Semaphore       = 10,
-    Timer           = 11
+    Timer           = 11,
+    ResourceLimit   = 12,
 };
 
 enum {
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -7,6 +7,7 @@
 #include "common/logging/log.h"
 
 #include "core/hle/kernel/process.h"
+#include "core/hle/kernel/resource_limit.h"
 #include "core/hle/kernel/thread.h"
 #include "core/memory.h"
 
--- a/src/core/hle/kernel/process.h
+++ b/src/core/hle/kernel/process.h
@@ -45,6 +45,8 @@ union ProcessFlags {
     BitField<12, 1, u16> loaded_high; ///< Application loaded high (not at 0x00100000).
 };
 
+class ResourceLimit;
+
 class Process final : public Object {
 public:
     static SharedPtr<Process> Create(std::string name, u64 program_id);
@@ -61,6 +63,8 @@ public:
     std::string name;
     /// Title ID corresponding to the process
     u64 program_id;
+    /// Resource limit descriptor for this process
+    SharedPtr<ResourceLimit> resource_limit;
 
     /// The process may only call SVCs which have the corresponding bit set.
     std::bitset<0x80> svc_access_mask;
--- a/src/core/hle/kernel/resource_limit.cpp
+++ b/src/core/hle/kernel/resource_limit.cpp
@@ -1,0 +1,157 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <cstring>
+
+#include "common/logging/log.h"
+
+#include "core/mem_map.h"
+#include "core/hle/kernel/resource_limit.h"
+
+namespace Kernel {
+
+static SharedPtr<ResourceLimit> resource_limits[4];
+
+ResourceLimit::ResourceLimit() {}
+ResourceLimit::~ResourceLimit() {}
+
+SharedPtr<ResourceLimit> ResourceLimit::Create(std::string name) {
+    SharedPtr<ResourceLimit> resource_limit(new ResourceLimit);
+
+    resource_limit->name = std::move(name);
+    return resource_limit;
+}
+
+SharedPtr<ResourceLimit> ResourceLimit::GetForCategory(ResourceLimitCategory category) {
+    switch (category)
+    {
+        case ResourceLimitCategory::APPLICATION:
+        case ResourceLimitCategory::SYS_APPLET:
+        case ResourceLimitCategory::LIB_APPLET:
+        case ResourceLimitCategory::OTHER:
+            return resource_limits[static_cast<u8>(category)];
+        default:
+            LOG_CRITICAL(Kernel, "Unknown resource limit category");
+            UNREACHABLE();
+    }
+}
+
+s32 ResourceLimit::GetCurrentResourceValue(u32 resource) const {
+    switch (resource) {
+        case COMMIT:
+            return current_commit;
+        case THREAD:
+            return current_threads;
+        case EVENT:
+            return current_events;
+        case MUTEX:
+            return current_mutexes;
+        case SEMAPHORE:
+            return current_semaphores;
+        case TIMER:
+            return current_timers;
+        case SHARED_MEMORY:
+            return current_shared_mems;
+        case ADDRESS_ARBITER:
+            return current_address_arbiters;
+        case CPU_TIME:
+            return current_cpu_time;
+        default:
+            LOG_ERROR(Kernel, "Unknown resource type=%08X", resource);
+            UNIMPLEMENTED();
+            return 0;
+    }
+}
+
+s32 ResourceLimit::GetMaxResourceValue(u32 resource) const {
+    switch (resource) {
+        case COMMIT:
+            return max_commit;
+        case THREAD:
+            return max_threads;
+        case EVENT:
+            return max_events;
+        case MUTEX:
+            return max_mutexes;
+        case SEMAPHORE:
+            return max_semaphores;
+        case TIMER:
+            return max_timers;
+        case SHARED_MEMORY:
+            return max_shared_mems;
+        case ADDRESS_ARBITER:
+            return max_address_arbiters;
+        case CPU_TIME:
+            return max_cpu_time;
+        default:
+            LOG_ERROR(Kernel, "Unknown resource type=%08X", resource);
+            UNIMPLEMENTED();
+            return 0;
+    }
+}
+
+void ResourceLimitsInit() {
+    // Create the four resource limits that the system uses
+    // Create the APPLICATION resource limit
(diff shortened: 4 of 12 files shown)


src/core/hle/kernel/kernel.h
```diff
@@ -46,7 +46,8 @@ enum class HandleType : u32 {
     Process         = 8,
     AddressArbiter  = 9,
     Semaphore       = 10,
-    Timer           = 11
+    Timer           = 11,
+    ResourceLimit   = 12
```
yuriks: Leave a trailing `,` here so future additions don't have to change the previous line again.

src/core/hle/kernel/process.cpp
```diff
@@ -14,12 +15,13 @@ namespace Kernel {
 
 u32 Process::next_process_id;
 
-SharedPtr<Process> Process::Create(std::string name, u64 program_id) {
+SharedPtr<Process> Process::Create(std::string name, u64 program_id, u8 resource_limit_category) {
```
yuriks: -1. The loader should attach a resource limit to the process after creating it.

src/core/hle/kernel/process.h
```diff
@@ -61,6 +63,10 @@ class Process final : public Object {
     std::string name;
     /// Title ID corresponding to the process
     u64 program_id;
+    // Resource limit descriptor for this process
+    SharedPtr<ResourceLimit> resource_limit;
+    // Resource limit category of this process
+    u8 resource_limit_category;
```
yuriks: I don't think you need to store this number.

src/core/hle/kernel/resource_limit.cpp
```diff
@@ -0,0 +1,158 @@
...
+            LOG_ERROR(Kernel, "Unknown resource type=%08X", resource);
+            UNIMPLEMENTED();
+            return 0;
+    }
+}
+
+void ResourceLimitsInit() {
+    // Create the four resource limits that the system uses
+    // Create the APPLICATION resource limit
+    SharedPtr<ResourceLimit> resource_limit = ResourceLimit::Create("Application Resource Limit");
```
yuriks: I think adding the "Resource Limit" at the end of each name is kinda redundant.

yuriks:
I didn't look at details for the whole thing yet, but a few general comments:

The way the resource limit categories are handled in the real system is quite detached from the kernel: the loader handles all that. It creates the 4 KResourceLimit objects and stores handles to them. Then when loading a process it uses the exheader to decide which limit to attach to the process after creating it. (See SVCs 0x77, 0x78 and 0x79: SetProcessResourceLimits, CreateResourceLimit and SetResourceLimitValues.) So the kernel itself doesn't know anything about the hardcoded limit ids and what the limits are.

Since we don't have a real loader service I think this is fine, but you might want to move all the stuff related to the 4 default limit categories to the loaders to make this separation more explicit.

Subv:
Updated.

I don't think our loader infrastructure is ready to handle the resource limits creation.

src/core/hle/kernel/process.h
```diff
@@ -61,6 +63,8 @@ class Process final : public Object {
     std::string name;
     /// Title ID corresponding to the process
     u64 program_id;
+    // Resource limit descriptor for this process
```
yuriks: Use triple slash

src/core/hle/kernel/resource_limit.h
```diff
@@ -0,0 +1,105 @@
...
+class ResourceLimit final : public Object {
+public:
+    /**
+     * Creates a resource limit object
+     */
+    static SharedPtr<ResourceLimit> Create(std::string name = "Unknown");
+
+    /**
+     * Retrieves the resource limit associated with the specified resource
+     * limit category
```
yuriks: Missing full stop.

src/core/hle/kernel/resource_limit.cpp
```diff
@@ -0,0 +1,167 @@
...
+static SharedPtr<ResourceLimit> resource_limits[4];
+
+ResourceLimit::ResourceLimit() {}
+ResourceLimit::~ResourceLimit() {}
+
+SharedPtr<ResourceLimit> ResourceLimit::Create(std::string name) {
+    SharedPtr<ResourceLimit> resource_limit(new ResourceLimit);
+
+    resource_limit->name = std::move(name);
+    resource_limit->max_priority = 0;
```
yuriks: You could initialize these inline in the class, to avoid duplicating all the fields here.

src/core/hle/kernel/resource_limit.h
```diff
@@ -0,0 +1,105 @@
...
+
+    s32 max_priority; ///< Max thread priority that a process in this category can create
+    s32 max_commit; ///< Max memory that processes in this category can use
+    s32 max_threads; ///< Max number of threads that can be collectively created by the processes in this category
+    s32 max_events; ///< Max number of events that can be collectively created by the processes in this category
+    s32 max_mutexes; ///< Max number of mutexes that can be collectively created by the processes in this category
+    s32 max_semaphores; ///< Max number of semaphores that can be collectively created by the processes in this category
+    s32 max_timers; ///< Max number of timers that can be collectively created by the processes in this category
+    s32 max_shared_mems; ///< Max number of shared memories that can be collectively created by the processes in this category
+    s32 max_address_arbiters; ///< Max number of address arbiters that can be collectively created by the processes in this category
```
yuriks: Did you forget the CPU time fields?

src/core/hle/svc.cpp
```diff
@@ -301,21 +302,45 @@ static void OutputDebugString(const char* string) {
...
-    // 0xFFFF8001 is a handle alias for the current KProcess, and 0xFFFF8000 is a handle alias for
-    // the current KThread.
-    *resource_limit = 0xDEADBEEF;
-    LOG_ERROR(Kernel_SVC, "(UNIMPLEMENTED) called process=0x%08X", process);
+static ResultCode GetResourceLimit(Handle* resource_limit, Handle process_handle) {
+    SharedPtr<Kernel::Process> process = Kernel::g_handle_table.Get<Kernel::Process>(process_handle);
+    if (process == nullptr)
+        return ERR_INVALID_HANDLE;
+
+    *resource_limit = Kernel::g_handle_table.Create(process->resource_limit).MoveFrom();
```
yuriks: Use `CASCADE_RESULT` here like in the other SVCs. (Ditto fro the other two)

src/core/hle/svc.cpp
```diff
@@ -301,21 +302,45 @@ static void OutputDebugString(const char* string) {
...
+    LOG_TRACE(Kernel_SVC, "called process=0x%08X", process_handle);
```
yuriks: You should move this log to the top to be consistent with the other two calls.

Subv:
Updated

yuriks:
```
Give me an L
Give me a  G
Give me a  T
Give me an M!
```

src/core/hle/kernel/resource_limit.h
```diff
@@ -0,0 +1,107 @@
...
+class ResourceLimit final : public Object {
+public:
+    /**
+     * Creates a resource limit object.
+     */
+    static SharedPtr<ResourceLimit> Create(std::string name = "Unknown");
+
+    /**
+     * Retrieves the resource limit associated with the specified resource
+     * limit category.
```
bunnei: unnecessary line break... this line would still be < 100 characters

src/core/hle/kernel/resource_limit.h
```diff
@@ -0,0 +1,107 @@
...
+    s32 max_threads = 0; ///< Max number of threads that can be collectively created by the processes in this category
+    s32 max_events = 0; ///< Max number of events that can be collectively created by the processes in this category
+    s32 max_mutexes = 0; ///< Max number of mutexes that can be collectively created by the processes in this category
+    s32 max_semaphores = 0; ///< Max number of semaphores that can be collectively created by the processes in this category
+    s32 max_timers = 0; ///< Max number of timers that can be collectively created by the processes in this category
+    s32 max_shared_mems = 0; ///< Max number of shared memories that can be collectively created by the processes in this category
+    s32 max_address_arbiters = 0; ///< Max number of address arbiters that can be collectively created by the processes in this category
+    s32 max_cpu_time = 0; ///< Max CPU time that the processes in this category can utilize
+
+    // TODO(Subv): Increment these in their respective Kernel::T::Create functions, keeping in mind that APPLICATION resource limits
```
bunnei: Can you line break this comment block @ 100 characters per line?

src/core/hle/kernel/resource_limit.h
```diff
@@ -0,0 +1,107 @@
...
+    s32 max_priority = 0; ///< Max thread priority that a process in this category can create
+    s32 max_commit = 0; ///< Max memory that processes in this category can use
+    s32 max_threads = 0; ///< Max number of threads that can be collectively created by the processes in this category
+    s32 max_events = 0; ///< Max number of events that can be collectively created by the processes in this category
+    s32 max_mutexes = 0; ///< Max number of mutexes that can be collectively created by the processes in this category
+    s32 max_semaphores = 0; ///< Max number of semaphores that can be collectively created by the processes in this category
+    s32 max_timers = 0; ///< Max number of timers that can be collectively created by the processes in this category
+    s32 max_shared_mems = 0; ///< Max number of shared memories that can be collectively created by the processes in this category
+    s32 max_address_arbiters = 0; ///< Max number of address arbiters that can be collectively created by the processes in this category
+    s32 max_cpu_time = 0; ///< Max CPU time that the processes in this category can utilize
```
bunnei: A lot of comment duplication, and a wall of text... could do:
```
// Max thread priority that a process in this category can create
s32 max_priority = 0;

// Max memory that processes in this category can use
s32 max_commit = 0;

// Max number of each respective object that can be collectively created by the processes in this category
s32 max_threads = 0;
s32 max_events = 0;
s32 max_mutexes = 0;
..

// Max CPU time that the processes in this category can utilize
s32 max_cpu_time = 0;
```
IMO this is more readable. Ditto for the block below.

bunnei:
LGTM

phase:
LGTM

PR #765, opened 2015-05-13
dyncom: Fix decoding of BKPT's immediate

lioncash:
A shift here is intended since the representation is imm12:imm4.

No actual behavior is affected by this (the immediate is purely for assembly/disassembly purposes), but it's more correct.


bunnei:
LGTM

PR #767, opened 2015-05-14
DOT3_RGB combiner

Lectem:
Implemented missing DOT3_RGB texture combiner. However it is not 100% accurate.

The hardware tests are  available at https://github.com/Lectem/3DS_gpu_tests/tree/DOT3_RGB

Sources : 
https://github.com/smealum/ctrulib/blob/2fed2f42411e0cf69e7c61bcae657139ca5cf1b8/libctru/include/3ds/gpu/gpu.h#L193
http://chimera.labs.oreilly.com/books/1234000001814/ch08.html#TextureCombiners

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -265,7 +265,7 @@ struct Regs {
             AddSigned       = 3,
             Lerp            = 4,
             Subtract        = 5,
-			Dot3_RGB		= 6,
+            Dot3_RGB		= 6,
             MultiplyThenAdd = 8,
             AddThenMultiply = 9,
         };
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -265,7 +265,7 @@ struct Regs {
             AddSigned       = 3,
             Lerp            = 4,
             Subtract        = 5,
-            Dot3_RGB		= 6,
+            Dot3_RGB        = 6,
             MultiplyThenAdd = 8,
             AddThenMultiply = 9,
         };
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -266,6 +266,7 @@ struct Regs {
             Lerp            = 4,
             Subtract        = 5,
             Dot3_RGB        = 6,
+
             MultiplyThenAdd = 8,
             AddThenMultiply = 9,
         };
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -265,7 +265,7 @@ struct Regs {
             AddSigned       = 3,
             Lerp            = 4,
             Subtract        = 5,
-
+			Dot3_RGB		= 6,
             MultiplyThenAdd = 8,
             AddThenMultiply = 9,
         };
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -546,8 +546,7 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                     }
                     case Operation::Dot3_RGB:
                     {
-                        //Not fully accurate
-                        //For hardware tests please check : https://github.com/Lectem/3DS_gpu_tests/tree/DOT3_RGB/
+                        // Not fully accurate
                         int result = ((input[0].r() * 2 - 255) * (input[1].r() * 2 - 255)+128) / 256 +
                                      ((input[0].g() * 2 - 255) * (input[1].g() * 2 - 255)+128) / 256 +
                                      ((input[0].b() * 2 - 255) * (input[1].b() * 2 - 255)+128) / 256;
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -546,12 +546,15 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                     }
                     case Operation::Dot3_RGB:
                     {
-                        // Not fully accurate
-                        int result = ((input[0].r() * 2 - 255) * (input[1].r() * 2 - 255)+128) / 256 +
-                                     ((input[0].g() * 2 - 255) * (input[1].g() * 2 - 255)+128) / 256 +
-                                     ((input[0].b() * 2 - 255) * (input[1].b() * 2 - 255)+128) / 256;
-                        result = std::max(0,std::min(255,result));
-                        return{ (u8)result, (u8)result, (u8)result };
+                        // Not fully accurate.
+                        // Worst case scenario seems to yield a +/-3 error
+                        // Some HW results indicate that the per-component computation can't have a higher precision than 1/256,
+                        // while dot3_rgb( (0x80,g0,b0),(0x7F,g1,b1) ) and dot3_rgb( (0x80,g0,b0),(0x80,g1,b1) ) give different results
+                        int result = ((input[0].r() * 2 - 255) * (input[1].r() * 2 - 255) + 128) / 256 +
+                                     ((input[0].g() * 2 - 255) * (input[1].g() * 2 - 255) + 128) / 256 +
+                                     ((input[0].b() * 2 - 255) * (input[1].b() * 2 - 255) + 128) / 256;
+                        result = std::max(0, std::min(255, result));
+                        return { (u8)result, (u8)result, (u8)result };
                     }
                     default:
                         LOG_ERROR(HW_GPU, "Unknown color combiner operation %d\n", (int)op);
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -544,12 +544,11 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         result = (result * input[2].Cast<int>()) / 255;
                         return result.Cast<u8>();
                     }
-					case Operation::Dot3_RGB:
-					{
-						auto result = Math::Dot(input[0], input[1]);
-						return{ result, result, result };
-					}
-
+                    case Operation::Dot3_RGB:
+                    {
+                        auto result = 4 * Math::Dot(input[0] - Math::MakeVec<u8>(127, 127, 127), input[1] - Math::MakeVec<u8>(127, 127, 127));
+                        return{ result, result, result };
+                    }
                     default:
                         LOG_ERROR(HW_GPU, "Unknown color combiner operation %d\n", (int)op);
                         UNIMPLEMENTED();
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -548,9 +548,9 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                     {
                         //Not fully accurate
                         //For hardware tests please check : https://github.com/Lectem/3DS_gpu_tests/tree/DOT3_RGB/
-                        int result = ((input[0].r() * 2 - 255) * (input[1].r() * 2 - 255)) / 256 +
-                                     ((input[0].g() * 2 - 255) * (input[1].g() * 2 - 255)) / 256 +
-                                     ((input[0].b() * 2 - 255) * (input[1].b() * 2 - 255)) / 256;
+                        int result = ((input[0].r() * 2 - 255) * (input[1].r() * 2 - 255)+128) / 256 +
+                                     ((input[0].g() * 2 - 255) * (input[1].g() * 2 - 255)+128) / 256 +
+                                     ((input[0].b() * 2 - 255) * (input[1].b() * 2 - 255)+128) / 256;
                         result = std::max(0,std::min(255,result));
                         return{ (u8)result, (u8)result, (u8)result };
                     }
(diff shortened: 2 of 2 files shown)


src/video_core/pica.h
```diff
@@ -265,7 +265,7 @@ struct Regs {
             AddSigned       = 3,
             Lerp            = 4,
             Subtract        = 5,
-
+			Dot3_RGB		= 6,
```
neobrain: spaces instead tabs for indentation

linkmauve: We use spaces for indent, not tabs.

src/video_core/rasterizer.cpp
```diff
@@ -544,6 +544,11 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         result = (result * input[2].Cast<int>()) / 255;
                         return result.Cast<u8>();
                     }
+					case Operation::Dot3_RGB:
```
linkmauve: Same here.

neobrain: Insert a padding line between case and }

src/video_core/rasterizer.cpp
```diff
@@ -544,6 +544,11 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+					{
+						auto result = Math::Dot(input[0], input[1]);
+						return{ result, result, result };
```
neobrain: change to `return { result, result, result };`

src/video_core/rasterizer.cpp
```diff
@@ -544,6 +544,11 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         result = (result * input[2].Cast<int>()) / 255;
                         return result.Cast<u8>();
                     }
+					case Operation::Dot3_RGB:
+					{
+						auto result = Math::Dot(input[0], input[1]);
```
linkmauve: For trivial types, the actual type (here u8 I guess) should be prefered to auto.

neobrain:
Your second link says:

`GL_DOT3_RGB: OutputColor = 4 * Dot(Arg0 - H, Arg1 - H) where H = (0.5, 0.5, 0.5)`

but your code doesn't actually subtract the H vector as it should be by documentation. Why?

neobrain:
Also: What was this change tested against?

src/video_core/rasterizer.cpp
linkmauve: Look around for our whitespace use, no space inside of the {} and one after return.

Lectem:
Oh damn, I got the wrong commit, lemme update it.
Sorry for the indentation, I don't use VS a lot

src/video_core/rasterizer.cpp
```diff
@@ -544,7 +544,11 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         result = (result * input[2].Cast<int>()) / 255;
                         return result.Cast<u8>();
                     }
-
+                    case Operation::Dot3_RGB:
+                    {
+                        auto result = 4 * Math::Dot(input[0] - Math::MakeVec<u8>(127, 127, 127), input[1] - Math::MakeVec<u8>(127, 127, 127));
```
bunnei: You use 127 for 0.5, and above we use 128 for 0.5. If this is unverified, can we find a way to check, or at least be consistent? (if its unverified, please put a TODO to verify)

Lectem: I didn't check it, I'm also unsure as to what we should do if the value is negative or over 255

src/video_core/pica.h
```diff
@@ -265,7 +265,7 @@ struct Regs {
             AddSigned       = 3,
             Lerp            = 4,
             Subtract        = 5,
-
+            Dot3_RGB		= 6,
```
neobrain: spaces for alignment...

neobrain:
There are some outstanding issues with the actual arithmetic being performed. E.g. there are lots of overflow/underflow issues. Even if those were fixed up, this PR seems to involve too much guessing to be merged without being tested properly.

purpasmart96:
It could just be 127.5

Lectem:
Here is a gist with the HW tests. After a few hours of trial and error, still couldn't reproduce the exact same result as the hardware. It probably has something to do with he way the PICA200 handles floats internally.
https://gist.github.com/Lectem/ecb65f4b1c36d0f146d4
I will give the f12.4 representation another shot tomorrow.

hrydgard:
dot products if done in hardware (DPPS on x86, vdot/vtfm/vmmul on PSP) is sometimes done with "align and add" for speed, which creates results that you can't replicate exactly with sequences of mul and add, instead you need to perform the same steps as the hardware in software to get it exact.

However as this is a fragment operation on PICA it's likely integer math so probably not impossible to replicate, but I would guess that 1-bit errors in DOT3 are relatively unlikely to have bad consequences for graphics, even though it's of course nice to be spot-on.

yuriks:
@hrydgard There is actually a case in @Lectem's tests that diverges quite widely and which we cannot explain: for (128, 128, 128) dot3 (x, 128, 128), which should be roughly equivalent to (0, 0, 0) dot3 (x, 0, 0), the result should give zero or very near zero for all values of x. Instead it seems that the value is 0 until x=128, at which point it increases linearly to reach 255 when x=255. This is the rightmost test-case on the spreadsheet. It's possible that this is simply a bug in the test-case however.

Lectem:
The problem seemed to come from a wrong RGBA macro and that the blending was still on. I'll reupload tests as soon as I confirm that the HW test is clear of any interferences

Lectem:
The new HW tests are available here : https://github.com/Lectem/3DS_gpu_tests/tree/DOT3_RGB
The best general formula I managed to get is 

```c++
int result = ((src0_r*2-255)*(src1_r*2-255)+0x80)/255+
((src0_g*2-255)*(src1_g*2-255)+0x80)/255+
((src0_b*2-255)*(src1_b*2-255)+0x80)/255;
uint8_t output = std::min(255u,std::max(0u,result )) ;
```

For the (x,x,x) dot3_rgb (x,x,x) I have 
```c++
int mult = ((color*4 - 510) );
if(mult<0) mult++;
mult*=mult;
if(color*4-510 > 0)mult = (mult+510)/(255*4);
else mult = (mult)/(255*4);
uint32_t result = (mult +mult +mult);
uint8_t output = std::min(255u,std::max(0u,result )) ;
```
which is obviously wrong but closer.

My guess is that there must be some rounding happening somewhere, but I'll have to spend a bit more time on it to figure it out

The factor 4 is embedded in the multiplication, and can't be a factor of the dot product. The reason is that by using as input (x,x,x) and (x,x,x) you always get a multiple of 3.

Lectem:
This is the best I managed to get, I will stop investigating this. Accuracy is in worst case +/-3, and in most cases it has only a 1bit error.

src/video_core/rasterizer.cpp
```diff
@@ -544,7 +544,16 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         result = (result * input[2].Cast<int>()) / 255;
                         return result.Cast<u8>();
                     }
-
+                    case Operation::Dot3_RGB:
+                    {
+                        //Not fully accurate
```
archshift: Leave a space between the comment syntax and message: `// Not fully accurate`.

Could you add `TODO` to this comment?

src/video_core/rasterizer.cpp
```diff
@@ -544,7 +544,16 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                        //For hardware tests please check : https://github.com/Lectem/3DS_gpu_tests/tree/DOT3_RGB/
```
archshift: Any way you could integrate these with our current hwtest system at https://github.com/citra-emu/hwtests ?

neobrain: Yeah, please don't refer to random-ish places in the Citra source code.

neobrain: Also, space after `//`

neobrain: Also, no space before `:`. Typography plz ;)

src/video_core/rasterizer.cpp
```diff
@@ -544,7 +544,16 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                        int result = ((input[0].r() * 2 - 255) * (input[1].r() * 2 - 255)+128) / 256 +
```
neobrain: This can be simplified to

    auto result_vec = (input[0].rgb() * 2 - Math::MakeVec(255,255,255)) * (input[1].rgb() * 2 - Math::MakeVec(255,255,255)),
    result_vec = (result_vec + Math::MakeVec(128,128,128)) / 256;
    int result = result_vec.r() + result_vec.g() + result_vec.b();

neobrain: Also, that +128 in the original code looks really confusing without spaces.

src/video_core/rasterizer.cpp
Lectem: I will try to integrate those tests on citra's hwtest system when I got some free time.
@neobrain sorry for the typography, I am used to using french typography. Will fix it too.

src/video_core/rasterizer.cpp
Lectem: I find this vec3 * vec3 operator overloading kinda confusing personally. I agree for the missing spaces though

neobrain: What specifically do you find confusing about it? It's a per-component product, just like + is a per-component sum. It's quite common among vector libraries, in fact, and at least if you store it in an explicitly typed variable there should be no confusion about that.

Lectem: I always find it ambiguous when people overload * on vectors, sometimes you'll see it as a per component product, sometimes a dot product, and sometimes even as a scalar product. But well I don't really mind either way, if you think it's better...

src/video_core/rasterizer.cpp
Lectem: This hwtest system isn't really suited for gpu tests, and I don't really have the time to port my tests to it (the gfx init and stuff would be hard to maintain in the long term anyway).
I'll fix the typos, but that'll be all. If you guys want my hw tests, feel free to copy/fork/submodule my repo.

src/video_core/rasterizer.cpp
```diff
@@ -544,7 +544,15 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                         result = (result * input[2].Cast<int>()) / 255;
                         return result.Cast<u8>();
                     }
-
+                    case Operation::Dot3_RGB:
+                    {
+                        // Not fully accurate
```
neobrain: Please be more specific in this comment - this is not really helpful without any context.

src/video_core/rasterizer.cpp
```diff
@@ -544,7 +544,15 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                        int result = ((input[0].r() * 2 - 255) * (input[1].r() * 2 - 255)+128) / 256 +
+                                     ((input[0].g() * 2 - 255) * (input[1].g() * 2 - 255)+128) / 256 +
```
neobrain: Spaces before/after +

src/video_core/rasterizer.cpp
```diff
@@ -544,7 +544,15 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                                     ((input[0].b() * 2 - 255) * (input[1].b() * 2 - 255)+128) / 256;
+                        result = std::max(0,std::min(255,result));
+                        return{ (u8)result, (u8)result, (u8)result };
```
neobrain: Spaces before {

src/video_core/pica.h
```diff
@@ -265,7 +265,7 @@ struct Regs {
             AddSigned       = 3,
             Lerp            = 4,
             Subtract        = 5,
-
+            Dot3_RGB        = 6,
```
neobrain: Insert a blank line after this to make the "gap" obvious.

yuriks:
Is there anything holding this up?

neobrain:
@lectem Could you set up a full (compileable) hardware test instead of just a gist file, so that I can eventually port the test easily to the hwtests framework?

Lectem:
@neobrain I already did, but I think you missed it (see https://github.com/citra-emu/citra/pull/767#issuecomment-103187122 )

https://github.com/Lectem/3DS_gpu_tests/blob/DOT3_RGB/source/main.c

neobrain:
I missed it indeed (I only saw the gist link). Mind putting that in the description box for easier discoverability?

Other than that, this PR can probably be merged now.

Lectem:
done.

archshift:
LGTM

yuriks:
LGTM

yuriks:
Ended up being merged as 04cb7ae6019402f1b53cea1beb5117d67628bce7 after a squash.

PR #772, opened 2015-05-14
core/video_core: Fix a few warnings when compiling on MSVC.

lioncash:


archshift:
LGTM

src/core/arm/skyeye_common/vfp/vfpsingle.cpp
```diff
@@ -388,7 +388,7 @@ static u32 vfp_single_fsqrt(ARMul_State* state, int sd, int unused, s32 m, u32 f
         } else {
             u64 term;
             s64 rem;
-            vsm.significand <<= !(vsm.exponent & 1);
+            vsm.significand <<= static_cast<u32>((vsm.exponent & 1) == 0);
```
yuriks: Why the cast?

lioncash: Because it causes a warning otherwise (using a bool in an arithmetic expression directly)

Subv:
Why is this PR named "core/video_core" ?

archshift:
Because it involves changes to core and video_core...

bunnei:
LGTM

PR #773, opened 2015-05-14
GPU: HILO8 support

ILOVEPIE:
This PR adds support for texture format 6 (HILO8).


lioncash:
You're using tabs, while the project uses spaces for indentation.

ILOVEPIE:
i intentionally used spaces... maybe my git changed them?

ILOVEPIE:
there we go tabs should be fixed

src/video_core/color.h
```diff
@@ -70,6 +70,15 @@ inline const Math::Vec4<u8> DecodeRGB8(const u8* bytes) {
 }
 
 /**
+ * Decode a "color" stored in HILO8 format
```
bunnei: Why is "color" in quotes in your docstrings? I'd get rid of the quotes

bunnei:
Changes mostly look fine. Please address my one comment and squash this to one commit. Also, have you tested this? (if not, please show that is correct with homebrew or a retail game)

src/video_core/color.h
ILOVEPIE: color is in quotes because this format isn't intended to store colors, it's intended to be used for normal maps.

ILOVEPIE:
squashed

neobrain:
Agreed that this should be tested in some way before merging.

src/video_core/color.h
neobrain: I guess this is more confusing than if the quotes weren't there. Just say `color` instead of `"color"`.

src/video_core/pica.h
```diff
@@ -152,7 +152,7 @@ struct Regs {
         RGB565       =  3,
         RGBA4        =  4,
         IA8          =  5,
-
+        HILO8        =  6,
```
neobrain: Is there any reason to actually call this HILO8? It's just an RG8 format, just that it's usually used in a very particular use case, so we can just call it RG8.

src/video_core/pica.h
```diff
@@ -174,6 +174,7 @@ struct Regs {
         case TextureFormat::RGB565:
         case TextureFormat::RGBA4:
         case TextureFormat::IA8:
+        case TextureFormat::HILO8:
```
neobrain: Why exactly does this return 4? I suppose this format should be documented somewhere in the source, if this is actually correct.

src/video_core/pica.h
ILOVEPIE: HILO8 is the standard OpenGL name

src/video_core/pica.h
ILOVEPIE: Format is 4 nibbles, 1 byte for the Hi part, 1 byte for the Lo part

src/video_core/color.h
lioncash: Might want to consider leaving a remark (using the `@remark` tag) that HILO8 is intended for normal maps as well.

src/video_core/pica.h
neobrain: Can you show me where standard OpenGL refers to this name? The closest that I could find was some NV extension.

IMO RG8 is both more consistent and self-explanatory. Also, we don't really follow any of the OpenGL conventions in this particuar place, so even if HILO is standard OpenGL terminology, that need not mean we should follow it in this place.

ILOVEPIE: 3dbrew also uses HILO8 as the name, this was named to be consistent with the Nvidia extension and 3dbrew. The main reason for the HILO8 name though is the fact that the format has a strict endianness specification. the format is strictly big endian with the High byte being R and the Low byte being G, wheras if we called it RG8 theoretically it could be interpereted as R is Low byte and G is High byte.

neobrain: Citra has an internal, also well-defined endianness specification. It doesn't seem to be mentioned in this enum, but it is explained in one of the framebuffer format enumerations. If RG8 doesn't match that specification, it would be GR8, of course.

Also fyi, bringing up the consistency with a source that you added yourself along with this PR is somewhat brazen.

I still think a non-standard extension is not the best reference, especially if it has such a cryptic name and when we have an internally more consistent alternative and definitely more clear name for it.

yuriks: @neobrain: References to HILO8 were already scattered through 3DBrew with that name, he just copied the relevant entry to the texture format list page.

yuriks:
Until an actual test case is found (Either a homebrew program, or a game. No, "this game uses it but I don't even know where" is not a valid test.) discussing this PR is pointless since it will not be merged, as there's no way to verify that the implementation is even remotely correct.

src/video_core/pica.h
ILOVEPIE: As yuriks said i only copyed the name to the texture format page from another part of 3dbrew

neobrain: Oh well. Then apparently it's NV and 3dbrew, the latter which rarely is a good source for naming things. My points stands for the same reasons: I don't think HILO is a good naming choice.

yuriks:
Please resume enum naming bikeshedding after this is tested.

yuriks:
So I finally got around to writing a test for this...

HILO8 test

Do you want to fix and revive this? If not, can I use your branch as a base for my own PR?

FWIW, since it seems to be just an unremarkable red-green texture format, I'd also prefer naming this RG8 to follow OpenGL 3 convention.

ILOVEPIE:
It's named that way in my patch hangon, ill look into why the texture is
flipped

yuriks:
It's flipped because the order of the R and G channels is wrong, should be a simple fix.

Anyway, I don't think this PR can be re-opened because you pushed to the branch at some point, so please open a new one.

ILOVEPIE:
can you send me the test application?

yuriks:
https://github.com/yuriks/3DS_gpu_tests/tree/hilo8-test

I changed it slightly since I took that picture, so the output will come out different. You should compile it and run the same binary on both Citra and HW to compare.

ILOVEPIE:
can you test it with the new version i put up i can't find my 3ds

PR #774, opened 2015-05-14
dyncom: Add ARMv6K NOP and hint instructions to the interpreter.

lioncash:
Previously these would be decoded as an MSR instruction, which is (obviously) wrong (and can possibly trash the CPSR as a result). With this, the interpreter fully handles all of the ARMv6K instruction set.


bunnei:
LGTM

ILOVEPIE:
LGTM

yuriks:
Without having verified the bits in the instruction encodings... LGTM

PR #776, opened 2015-05-14
GPU: Consolidate Pica state

bunnei:
This PR creates a new struct, `Pica::State`, and moves `Pica::Regs` and vertex shader memory there.  The purpose of this struct is to consolidate all Pica memory, primarily to make state management easier, but also cleanliness - as I introduce new features that require additional memory/state, I don't want it sprinkled as static objects all throughout various modules. A few cleanups/refactors were necessary for this change.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -16,6 +16,8 @@
 #include "common/common_types.h"
 #include "common/logging/log.h"
 
+#include "math.h" 
+
 namespace Pica {
 
 // Returns index corresponding to the Regs member labeled by field_name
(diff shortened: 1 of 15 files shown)


src/video_core/pica.h
```diff
@@ -1051,4 +1049,43 @@ union CommandHeader {
...
+    Regs regs;
+
+    struct {
+        struct {
+            Math::Vec4<float24> f[96];
+            std::array<bool, 16> b;
+            std::array<Math::Vec4<u8>, 4> i;
+        } uniforms;
+
+        Math::Vec4<float24>& GetFloatUniform(u32 index) {
```
neobrain: There's no reason to have these functions anymore.

src/video_core/pica.h
```diff
@@ -1051,4 +1049,31 @@ union CommandHeader {
...
+
+        Math::Vec4<float24> default_attributes[16];
+
+        std::array<u32, 1024> program_code;
+        std::array<u32, 1024> swizzle_data;
+    } vs;
+};
+
+/// Gets the current state of Pica (registers, memory, etc.)
+State& GetState();
```
yuriks: Is there any compelling reason for this function over a simple global var?

bunnei: Nope... I can just make it a global

src/video_core/video_core.cpp
```diff
@@ -31,7 +35,10 @@ void Init(EmuWindow* emu_window) {
 
 /// Shutdown the video core
 void Shutdown() {
+    Pica::Shutdown();
```
neobrain: Resources and modules should be uninitialized in reverse-order compared to initialization order.

neobrain:
LGTM other than my comment.

src/video_core/pica.h
```diff
@@ -18,6 +18,8 @@
 
 #include "core/mem_map.h"
 
+#include "math.h"
```
neobrain: Trailing whitespace

src/video_core/vertex_shader.cpp
```diff
@@ -135,12 +91,12 @@ static void ProcessShaderCode(VertexShaderState& state) {
         const Instruction& instr = *(const Instruction*)state.program_counter;
         const SwizzlePattern& swizzle = *(SwizzlePattern*)&swizzle_data[instr.common.operand_desc_id];
 
-        static auto call = [](VertexShaderState& state, u32 offset, u32 num_instructions,
+        static auto call = [&program_code](VertexShaderState& state, u32 offset, u32 num_instructions,
```
neobrain: You can't reliably capture non-static variables in static lambda functions. Please submit a follow-up PR removing this capture or un-static-fying this lambda.

PR #782, opened 2015-05-16
vertex_shader.cpp: Convert giant if block to a switch block

archshift:
GCC/Clang's optimizer isn't smart enough to realize that the two options have identical outcomes, but the switch variant has much smaller generated code size and as such has a significantly smaller performance footprint when called in large numbers.

[Code generation between the if-block variant and the switch-block variant](
https://goo.gl/6hYXHP)

[A short benchmark of performance of these two variants compared](http://ideone.com/5PnzFg). As you can see, the switch variant takes about 86% of the execution time of the if variant.


Subv:
This should be a loop instead of a bunch of ifs, imo, wouldn't that lead to even smaller code?

neobrain:
Fyi, none of this code actually needs to be there. All of this setup in command_processor.cpp instead, IIRC without the need for any "if"s.

Also btw, does this have any real-world benefit? If it doesn't show up in a profiler, chances are there's little point in optimizing it.

yuriks:
@Subv: It's hard to make this into a loop due to the use of the BitFields. They would have to be converted into an array + manual bitshifting instead.

@archshift: Did you actually benchmark this *inside* Citra? Artificial benchmarks aren't a suitable substitute.

archshift:
@neobrain are you saying that this code can just be removed without any detriment?

Subv:
@yuriks There's a GetRegisterForAttribute function

yuriks:
@Subv which will be much more inneficient than doing this due to how it's coded.

archshift:
Anyway, if there are performance gains in Citra, I wasn't able to measure them. Merge or close this PR at your discretion.

yuriks:
@archshift: You can just remove the `case 0`. Do that an I vote merge.

archshift:
I can't remove the `case 0`, because then zero would fall under the default case and set all the attributes, which it didn't do originally.

yuriks:
Then... don't add a `default` case?

yuriks:
To clarify: I think the range this variable can take is bounded, correct? So the `default` label isn't needed. Though in that case you can keep the `case 0` label, just don't add a break to it.

archshift:
Well the original code just had `if (num_attributes > 15)`, which theoretically could be anything from 16 to INT_MAX (necessitating the `default` label).

kitling:
Unless there's a rework of the underlying class this is accessing to allow you to index arrays, it would be mush easier (cleaner?) to do the switch statement. But if it were setup like this:
```
attribute_register_map.attribute[x]
```
Or similar, it would make more sense and would allow using a loop for this. Anyway...

@archshift: Yeah, that seems to be a bit confusing. I understand *why* it's setup that way, but I think, how that code was setup, this would more clearly convey the intent:
```
...
if (num_attributes >= 14) // do something
if (num_attributes >= 15) // do something
if (num_attributes == 16) // do something
...
```
The code makes me feel like you are never going to make more than 16 registers, so perhaps a case for 0 that instantly returns, and a LOG_whatever for the default case that falls through to the rest would be pertinent.

neobrain:
@archshift num_attributes cannot be greater than 16, yes.

archshift:
Removed the zero case after some clarification by @neobrain.

lioncash:
Shouldn't the default just break from the switch instead of falling through? Seems more safe IMO, in the future if there's a bug that somehow (incredibly) affects this, it won't trash values.

yuriks:
Just put the default label at the end.

archshift:
I can tell by the varying interpretations of this code that the micro-optimization leads to a visible loss in code readability. Not really worth it.

PR #783, opened 2015-05-16
Use condition var to properly pause the CPU thread

### LINKED ISSUE #779: Emu thread spins at 100% CPU when emulation is paused
Even if you pause emulation, the emu thread continues to spin checking for the unpause condition. It should do a non-busy wait on a condvar instead.

jroweboy:
Adds support for threaded pausing so citra doesn't spin wait on pause. Fixes #779

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -59,6 +59,9 @@ void EmuThread::run() {
             yieldCurrentThread();
             
             was_active = false;
+        } else {
+            std::unique_lock<std::mutex> lock(running_mutex);
+            running_cv.wait(lock, [this]{ return IsRunning() || stop_run; });
         }
     }
 
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -3,6 +3,8 @@
 // Refer to the license.txt file included.
 
 #include <atomic>
+#include <condition_variable>
+#include <mutex>
 
 #include <QThread>
 #include <QGLWidget>
@@ -40,7 +42,12 @@ public:
      * @param running Boolean value, set the emulation thread to running if true
      * @note This function is thread-safe
      */
-    void SetRunning(bool running) { this->running = running; }
+    void SetRunning(bool running) {
+        std::unique_lock<std::mutex> lock(running_mutex);
+        this->running = running;
+        lock.unlock();
+        running_cv.notify_all();
+    }
 
     /**
      * Check if the emulation thread is running or not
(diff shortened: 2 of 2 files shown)


src/citra_qt/bootmanager.cpp
```diff
@@ -60,6 +59,10 @@ void EmuThread::run() {
             
             was_active = false;
         }
+        else {
+            std::unique_lock<std::mutex> lock(running_mutex);
+            running_cv.wait(lock, [this]{return IsRunning();});
```
lioncash: `running_cv.wait(lock, [this]{ return IsRunning(); });`

yuriks:
This does not handle the case when Stop is selected while in a paused state, causing the program to hang when you do that or close Citra.

jroweboy:
Ah good catch, lemme look into that

yuriks:
Last version fixes the issue. LGTM

src/citra_qt/bootmanager.cpp
```diff
@@ -34,7 +34,6 @@ EmuThread::EmuThread(GRenderWindow* render_window) :
 
 void EmuThread::run() {
     stop_run = false;
-
```
yuriks: Why remove this line?

src/citra_qt/bootmanager.cpp
```diff
@@ -60,6 +59,10 @@ void EmuThread::run() {
             
             was_active = false;
         }
+        else {
```
yuriks: `else` goes in the same line as the closing brace

src/citra_qt/bootmanager.h
```diff
@@ -7,6 +7,9 @@
 #include <QThread>
 #include <QGLWidget>
 
+#include <mutex>
+#include <condition_variable>
```
yuriks: Include these headers at the top with the rest of the standard library headers. Also sort them alphabetically.

src/citra_qt/bootmanager.h
```diff
@@ -54,13 +61,15 @@ class EmuThread : public QThread
      */
     void RequestStop() {
         stop_run = true;
-        running = false;
+        this->SetRunning(false);
```
lioncash: There's no need for the `this->` here.

src/citra_qt/bootmanager.cpp
```diff
@@ -59,6 +59,9 @@ void EmuThread::run() {
             yieldCurrentThread();
             
             was_active = false;
+        } else {
+            std::unique_lock<std::mutex> lock(running_mutex);
+            running_cv.wait(lock, [this]{ return this->IsRunning() || this->stop_run; });
```
lioncash: `this->` isn't necessary here, it's implied in the capture.

archshift:
LGTM

bunnei:
Looks fine. Pleasr squash to a single commit.

jroweboy:
I've squashed the commits. Thanks for the feedback so far :)

PR #785, opened 2015-05-17
Implement svcBreak

archshift:


Subv:
Is it ok to let it continue execution after this?

archshift:
Ostensibly no, but I'm not sure what the best way to break execution would be from here. I don't think an assert would be appropriate.

yuriks:
Breaking execution here would cause many titles that currently boot to stop booting. I think we should hold off on that for a while longer.

yuriks:
LGTM btw

Subv:
LGTM

bunnei:
LGTM

PR #753, opened 2015-05-11
OpenGL Hardware Renderer

tfarley:
This implements an OpenGL renderer that currently only performs the drawing and rasterizing stages using output vertices from the existing vertex shader interpreter.

The components of this with primary source files are:
* OpenGL rasterizer (gl_rasterizer, command_processor)
  * Receives geometry and draws it, manages OpenGL state, syncs with 3ds framebuffers
* Rasterizer fragment shader (gl_shaders)
  * Emulates texture environment stages, alpha testing, etc. with dummy vertex shader
* Rasterizer cache (gl_rasterizer_cache)
  * Loads and holds 3ds textures in OGL VRAM, receives flushes
* Caching/flush management (gsp_gpu + gpu)
  * Notifies rasterizer and cache when 3ds memory changes so they may flush caches
* OpenGL state management (gl_state)
  * Encasulates OpenGL state values, applies when requested with minimal state changes
* OpenGL resource management (gl_resource_manager)
  * Auto-deletion of resources when out of scope, copy prevention
* INI setting for sw/hw rasterizer toggle

I'd like comments on any/all of these!
In particular, how reasonable the state and resource management parts are.

PR #789, opened 2015-05-19
OpenGL Hardware Renderer

tfarley:
This implements an OpenGL renderer that currently only performs the drawing and rasterizing stages using output vertices from the existing vertex shader interpreter.

The components of this with primary source files are:
* OpenGL rasterizer (gl_rasterizer, command_processor)
  * Receives geometry and draws it, manages OpenGL state, syncs with PICA state and 3DS in-memory framebuffers
* Rasterizer fragment shader (gl_shaders)
  * Emulates texture environment stages, alpha testing, etc. with dummy vertex shader
* Rasterizer cache (gl_rasterizer_cache)
  * Loads and holds 3DS textures in OpenGL VRAM, receives flushes
* Caching/flush management (gsp_gpu, gpu)
  * Notifies rasterizer and resource cache when 3DS memory changes so they may flush caches
* OpenGL state management (gl_state)
  * Encasulates OpenGL state values, applies when requested with minimal state changes
* OpenGL resource management (gl_resource_manager)
  * Auto-deletion of resources when out of scope, copy prevention
* INI setting for sw/hw rasterizer toggle
* MakeCurrent race condition fix (bootmanager, main)
  * Makes the OpenGL context current before the emu thread processing begins

The initial RFC for reference: #753

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -28,11 +28,11 @@
 
 EmuThread::EmuThread(GRenderWindow* render_window) :
     exec_step(false), running(false), stop_run(false), render_window(render_window) {
-
-    connect(this, SIGNAL(started()), render_window, SLOT(moveContext()));
 }
 
 void EmuThread::run() {
+    render_window->MakeCurrent();
+
     stop_run = false;
 
     // holds whether the cpu was running during the last iteration,
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -44,6 +44,8 @@
 #include "core/arm/disassembler/load_symbol_map.h"
 #include "citra_qt/config.h"
 
+#include "video_core/video_core.h"
+
 #include "version.h"
 
 GMainWindow::GMainWindow() : emu_thread(nullptr)
@@ -123,6 +125,9 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     restoreState(settings.value("state").toByteArray());
     render_window->restoreGeometry(settings.value("geometryRenderWindow").toByteArray());
 
+    ui.action_Use_Hardware_Renderer->setChecked(Settings::values.use_hw_renderer);
+    SetHardwareRendererEnabled(ui.action_Use_Hardware_Renderer->isChecked());
+    
     ui.action_Single_Window_Mode->setChecked(settings.value("singleWindowMode", true).toBool());
     ToggleWindowMode();
 
@@ -135,6 +140,7 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     connect(ui.action_Start, SIGNAL(triggered()), this, SLOT(OnStartGame()));
     connect(ui.action_Pause, SIGNAL(triggered()), this, SLOT(OnPauseGame()));
     connect(ui.action_Stop, SIGNAL(triggered()), this, SLOT(OnStopGame()));
+    connect(ui.action_Use_Hardware_Renderer, SIGNAL(triggered(bool)), this, SLOT(SetHardwareRendererEnabled(bool)));
     connect(ui.action_Single_Window_Mode, SIGNAL(triggered(bool)), this, SLOT(ToggleWindowMode()));
     connect(ui.action_Hotkeys, SIGNAL(triggered()), this, SLOT(OnOpenHotkeysDialog()));
 
@@ -300,6 +306,9 @@ void GMainWindow::OnOpenHotkeysDialog()
     dialog.exec();
 }
 
+void GMainWindow::SetHardwareRendererEnabled(bool enabled) {
+    VideoCore::g_hw_renderer_enabled = enabled;
+}
 
 void GMainWindow::ToggleWindowMode() {
     if (ui.action_Single_Window_Mode->isChecked()) {
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -216,6 +216,7 @@ void GMainWindow::BootGame(std::string filename) {
     // Create and start the emulation thread
     emu_thread = Common::make_unique<EmuThread>(render_window);
     emit EmulationStarting(emu_thread.get());
+    render_window->moveContext();
     emu_thread->start();
 
     // BlockingQueuedConnection is important here, it makes sure we've finished refreshing our views before the CPU continues
--- a/src/core/hle/service/gsp_gpu.cpp
+++ b/src/core/hle/service/gsp_gpu.cpp
@@ -15,6 +15,7 @@
 #include "core/hw/lcd.h"
 
 #include "video_core/gpu_debugger.h"
+#include "video_core/video_core.h"
 
 // Main graphics debugger object - TODO: Here is probably not the best place for this
 GraphicsDebugger g_debugger;
@@ -264,6 +265,8 @@ static void FlushDataCache(Service::Interface* self) {
     u32 size    = cmd_buff[2];
     u32 process = cmd_buff[4];
 
+    VideoCore::g_renderer->hw_rasterizer->NotifyFlush(Memory::VirtualToPhysicalAddress(address), size);
+
     // TODO(purpasmart96): Verify return header on HW
 
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
@@ -352,10 +355,16 @@ static void ExecuteCommand(const Command& command, u32 thread_id) {
 
     // GX request DMA - typically used for copying memory from GSP heap to VRAM
     case CommandId::REQUEST_DMA:
+        VideoCore::g_renderer->hw_rasterizer->NotifyPreRead(Memory::VirtualToPhysicalAddress(command.dma_request.source_address),
+                                                            command.dma_request.size);
+
         memcpy(Memory::GetPointer(command.dma_request.dest_address),
                Memory::GetPointer(command.dma_request.source_address),
                command.dma_request.size);
         SignalInterrupt(InterruptId::DMA);
+
+        VideoCore::g_renderer->hw_rasterizer->NotifyFlush(Memory::VirtualToPhysicalAddress(command.dma_request.dest_address),
+                                                          command.dma_request.size);
         break;
 
     // ctrulib homebrew sends all relevant command list data with this command,
--- a/src/core/hw/gpu.cpp
+++ b/src/core/hw/gpu.cpp
@@ -106,6 +106,8 @@ inline void Write(u32 addr, const T data) {
             } else {
                 GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::PSC1);
             }
+
+            VideoCore::g_renderer->hw_rasterizer->NotifyFlush(config.GetStartAddress(), config.GetEndAddress() - config.GetStartAddress());
         }
         break;
     }
@@ -129,19 +131,25 @@ inline void Write(u32 addr, const T data) {
             u32 output_width = config.output_width / horizontal_scale;
             u32 output_height = config.output_height / vertical_scale;
 
+            u32 input_size = config.input_width * config.input_height * GPU::Regs::BytesPerPixel(config.input_format);
+            u32 output_size = output_width * output_height * GPU::Regs::BytesPerPixel(config.output_format);
+
+            VideoCore::g_renderer->hw_rasterizer->NotifyPreRead(config.GetPhysicalInputAddress(), input_size);
+
             if (config.raw_copy) {
                 // Raw copies do not perform color conversion nor tiled->linear / linear->tiled conversions
                 // TODO(Subv): Verify if raw copies perform scaling
-                memcpy(dst_pointer, src_pointer, config.output_width * config.output_height * 
-                        GPU::Regs::BytesPerPixel(config.output_format));
+                memcpy(dst_pointer, src_pointer, output_size);
                 
                 LOG_TRACE(HW_GPU, "DisplayTriggerTransfer: 0x%08x bytes from 0x%08x(%ux%u)-> 0x%08x(%ux%u), output format: %x, flags 0x%08X, Raw copy",
-                    config.output_height * output_width * GPU::Regs::BytesPerPixel(config.output_format),
+                    output_size,
                     config.GetPhysicalInputAddress(), config.input_width.Value(), config.input_height.Value(),
                     config.GetPhysicalOutputAddress(), config.output_width.Value(), config.output_height.Value(),
                     config.output_format.Value(), config.flags);
 
                 GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::PPF);
+
+                VideoCore::g_renderer->hw_rasterizer->NotifyFlush(config.GetPhysicalOutputAddress(), output_size);
                 break;
             }
 
@@ -247,6 +255,8 @@ inline void Write(u32 addr, const T data) {
                       config.output_format.Value(), config.flags);
 
             GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::PPF);
+
+            VideoCore::g_renderer->hw_rasterizer->NotifyFlush(config.GetPhysicalOutputAddress(), output_size);
         }
         break;
     }
--- a/src/video_core/CMakeLists.txt
+++ b/src/video_core/CMakeLists.txt
@@ -1,7 +1,11 @@
 set(SRCS
             renderer_opengl/generated/gl_3_2_core.c
-            renderer_opengl/renderer_opengl.cpp
+            renderer_opengl/gl_rasterizer.cpp
+            renderer_opengl/gl_rasterizer_cache.cpp
+            renderer_opengl/gl_resource_manager.cpp
             renderer_opengl/gl_shader_util.cpp
+            renderer_opengl/gl_state.cpp
+            renderer_opengl/renderer_opengl.cpp
             debug_utils/debug_utils.cpp
             clipper.cpp
             command_processor.cpp
@@ -15,13 +19,19 @@ set(SRCS
 set(HEADERS
             debug_utils/debug_utils.h
             renderer_opengl/generated/gl_3_2_core.h
+            renderer_opengl/gl_rasterizer.h
+            renderer_opengl/gl_rasterizer_cache.h
+            renderer_opengl/gl_resource_manager.h
             renderer_opengl/gl_shader_util.h
             renderer_opengl/gl_shaders.h
+            renderer_opengl/gl_state.h
+            renderer_opengl/pica_to_gl.h
             renderer_opengl/renderer_opengl.h
             clipper.h
             color.h
             command_processor.h
             gpu_debugger.h
+            hwrasterizer_base.h
             math.h
             pica.h
             primitive_assembly.h
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -1,0 +1,879 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/settings.h"
+#include "core/hw/gpu.h"
+
+#include "video_core/color.h"
+#include "video_core/pica.h"
+#include "video_core/utils.h"
+#include "video_core/renderer_opengl/gl_rasterizer.h"
+#include "video_core/renderer_opengl/gl_shaders.h"
+#include "video_core/renderer_opengl/gl_shader_util.h"
+#include "video_core/renderer_opengl/pica_to_gl.h"
+
+#include "generated/gl_3_2_core.h"
+
+#include <memory>
+
+static bool IsPassThroughTevStage(const Pica::Regs::TevStageConfig& stage) {
+    return (stage.color_op == Pica::Regs::TevStageConfig::Operation::Replace &&
+            stage.alpha_op == Pica::Regs::TevStageConfig::Operation::Replace &&
+            stage.color_source1 == Pica::Regs::TevStageConfig::Source::Previous &&
+            stage.alpha_source1 == Pica::Regs::TevStageConfig::Source::Previous &&
+            stage.color_modifier1 == Pica::Regs::TevStageConfig::ColorModifier::SourceColor &&
+            stage.alpha_modifier1 == Pica::Regs::TevStageConfig::AlphaModifier::SourceAlpha &&
+            stage.GetColorMultiplier() == 1 &&
+            stage.GetAlphaMultiplier() == 1);
+}
+
+RasterizerOpenGL::RasterizerOpenGL() : last_fb_color_addr(0), last_fb_depth_addr(0) { }
+RasterizerOpenGL::~RasterizerOpenGL() { }
+
+void RasterizerOpenGL::InitObjects() {
+    // Create the hardware shader program and get attrib/uniform locations
+    shader.Create(GLShaders::g_vertex_shader_hw, GLShaders::g_fragment_shader_hw);
+    attrib_position = glGetAttribLocation(shader.handle, "vert_position");
+    attrib_color = glGetAttribLocation(shader.handle, "vert_color");
+    attrib_texcoords = glGetAttribLocation(shader.handle, "vert_texcoords");
+
+    uniform_alphatest_enabled = glGetUniformLocation(shader.handle, "alphatest_enabled");
+    uniform_alphatest_func = glGetUniformLocation(shader.handle, "alphatest_func");
+    uniform_alphatest_ref = glGetUniformLocation(shader.handle, "alphatest_ref");
+
+    uniform_tex = glGetUniformLocation(shader.handle, "tex");
+
+    uniform_tev_combiner_buffer_color = glGetUniformLocation(shader.handle, "tev_combiner_buffer_color");
+
+    const auto tev_stages = Pica::registers.GetTevStages();
+    for (unsigned tev_stage_index = 0; tev_stage_index < tev_stages.size(); ++tev_stage_index) {
+        auto& uniform_tev_cfg = uniform_tev_cfgs[tev_stage_index];
+
+        std::string tev_ref_str = "tev_cfgs[" + std::to_string(tev_stage_index) + "]";
+        uniform_tev_cfg.enabled = glGetUniformLocation(shader.handle, (tev_ref_str + ".enabled").c_str());
+        uniform_tev_cfg.color_sources = glGetUniformLocation(shader.handle, (tev_ref_str + ".color_sources").c_str());
+        uniform_tev_cfg.alpha_sources = glGetUniformLocation(shader.handle, (tev_ref_str + ".alpha_sources").c_str());
+        uniform_tev_cfg.color_modifiers = glGetUniformLocation(shader.handle, (tev_ref_str + ".color_modifiers").c_str());
+        uniform_tev_cfg.alpha_modifiers = glGetUniformLocation(shader.handle, (tev_ref_str + ".alpha_modifiers").c_str());
+        uniform_tev_cfg.color_alpha_op = glGetUniformLocation(shader.handle, (tev_ref_str + ".color_alpha_op").c_str());
+        uniform_tev_cfg.color_alpha_multiplier = glGetUniformLocation(shader.handle, (tev_ref_str + ".color_alpha_multiplier").c_str());
+        uniform_tev_cfg.const_color = glGetUniformLocation(shader.handle, (tev_ref_str + ".const_color").c_str());
+        uniform_tev_cfg.updates_combiner_buffer_color_alpha = glGetUniformLocation(shader.handle, (tev_ref_str + ".updates_combiner_buffer_color_alpha").c_str());
+    }
+
+    // Generate VBO and VAO
+    vertex_buffer.Create();
+    vertex_array.Create();
+
+    // Update OpenGL state
+    state.draw.vertex_array = vertex_array.handle;
+    state.draw.vertex_buffer = vertex_buffer.handle;
+    state.draw.shader_program = shader.handle;
+
+    state.Apply();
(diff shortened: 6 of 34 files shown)


src/core/hw/gpu.cpp
```diff
@@ -142,6 +149,9 @@ inline void Write(u32 addr, const T data) {
                     config.output_format.Value(), config.flags);
 
                 GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::PPF);
+
+                VideoCore::g_renderer->hw_rasterizer->NotifyFlush(config.GetPhysicalOutputAddress(), output_size);
```
bunnei: I don't think this is necessary, won't https://github.com/citra-emu/citra/pull/789/files#diff-ab95d061e69303f8a795c12b7fdb1208R261 still get called right after this `case` statement?

tfarley: They're part of the same case statement :) and the break right after this line would skip the second flush.

bunnei: Right... I misread that

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+        state.blend.src_a_func = PicaToGL::BlendFunc(Pica::registers.output_merger.alpha_blending.factor_source_a);
+        state.blend.dst_a_func = PicaToGL::BlendFunc(Pica::registers.output_merger.alpha_blending.factor_dest_a);
+        break;
+    }
+    case PICA_REG_INDEX(output_merger.blend_const):
+    {
+        state.blend.color.red = (GLclampf)Pica::registers.output_merger.blend_const.r / 255.0f;
+        state.blend.color.green = (GLclampf)Pica::registers.output_merger.blend_const.g / 255.0f;
+        state.blend.color.blue = (GLclampf)Pica::registers.output_merger.blend_const.b / 255.0f;
+        state.blend.color.alpha = (GLclampf)Pica::registers.output_merger.blend_const.a / 255.0f;
```
bunnei: Utility functions for GL <-> Pica color format conversions would probably be nice (e.g. like https://github.com/citra-emu/citra/blob/master/src/video_core/color.h).

bunnei: Followup: Probably belongs in `PicaToGL` namespace

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+    }
+
+    // Depth test
+    case PICA_REG_INDEX(output_merger.depth_test_enable):
+    {
+        state.depth.test_enabled = Pica::registers.output_merger.depth_test_enable;
+        state.depth.test_func = PicaToGL::CompareFunc(Pica::registers.output_merger.depth_test_func);
+
+        if (Pica::registers.output_merger.depth_write_enable) {
+            state.depth.write_mask = GL_TRUE;
```
bunnei: `state.depth.write_mask = Pica::registers.output_merger.depth_write_enable ? GL_TRUE : GL_FALSE;`

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+        break;
+    case PICA_REG_INDEX(tev_stage5.color_op):
+        SyncTevOps(5, Pica::registers.tev_stage5);
+        break;
+    case PICA_REG_INDEX(tev_stage5.const_r):
+        SyncTevColor(5, Pica::registers.tev_stage5);
+        break;
+    case PICA_REG_INDEX(tev_stage5.color_scale):
+        SyncTevMultipliers(5, Pica::registers.tev_stage5);
+        break;
```
bunnei: I can't help but think that all of these case's for the 6 TEV stages could be collapsed some how, but it's fine as-is for this PR

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+                                      Pica::registers.tev_combiner_buffer_color.a / 255.0f };
+
+        glUniform4fv(uniform_tev_combiner_buffer_color, 1, combiner_color);
+        break;
+    }
+
+    // TEV combiner buffer write flags
+    case PICA_REG_INDEX(tev_combiner_buffer_input):
+    {
+        for (int tev_stage_idx = 0; tev_stage_idx < 6; ++tev_stage_idx) {
```
bunnei: This should be that constant (6) we have somewhere for the number of TEV stages, not the actual number 6

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+            glUniform2i(uniform_tev_cfgs[tev_stage_idx].updates_combiner_buffer_color_alpha,
+                        Pica::registers.tev_combiner_buffer_input.TevStageUpdatesCombinerBufferColor(tev_stage_idx),
+                        Pica::registers.tev_combiner_buffer_input.TevStageUpdatesCombinerBufferAlpha(tev_stage_idx));
+        }
+        break;
+    }
+
+    default:
+        break;
```
bunnei: Is this `default` necessary if we aren't going to do anything?

JMC47:
Tested the following games on the latest build to confirm there were no regressions:

- Cave Story 3D
- Super Monkey Ball 3D
- VVVVVV
- The Legend of Zelda: Ocarina of Time 3D
- Rollercoaster Tycoon 3D
- ShovelKnight
- Cubic Ninja

Everything seems to be fine and match the software renderer for the most part.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+    // TODO: Ensure viewport_corner.x should not be negated or origin flipped
+    glViewport((GLsizei)static_cast<float>(Pica::registers.viewport_corner.x),
+                -(GLsizei)static_cast<float>(Pica::registers.viewport_corner.y)
+                    + Pica::registers.framebuffer.GetHeight() - viewport_height,
+                viewport_width, viewport_height);
+
+    // Sync bound texture(s), upload if not cached
+    const auto pica_textures = Pica::registers.GetTextures();
+
+    for (int i = 0; i < 3; ++i) {
```
bunnei: `3` could be a constant defined somewhere (for better readability)

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -53,17 +53,21 @@ static std::array<GLfloat, 3*2> MakeOrthographicMatrix(const float width, const
...
 /// RendererOpenGL constructor
 RendererOpenGL::RendererOpenGL() {
+    hw_rasterizer.reset(new RasterizerOpenGL());
     resolution_width  = std::max(VideoCore::kScreenTopWidth, VideoCore::kScreenBottomWidth);
     resolution_height = VideoCore::kScreenTopHeight + VideoCore::kScreenBottomHeight;
 }
 
 /// RendererOpenGL destructor
 RendererOpenGL::~RendererOpenGL() {
+
```
bunnei: Unnecessary space

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -0,0 +1,159 @@
...
+            blend.src_a_func != cur_state.blend.src_a_func ||
+            blend.dst_a_func != cur_state.blend.dst_a_func) {
+            glBlendFuncSeparate(blend.src_rgb_func, blend.dst_rgb_func, blend.src_a_func, blend.dst_a_func);
+        }
+    } else if (blend.enabled != cur_state.blend.enabled) {
+        glDisable(GL_BLEND);
+    }
+
+    // Textures
+    for (int i = 0; i < 3; ++i) {
```
bunnei: Ditto to the comment above on using a constant for this

src/video_core/renderer_opengl/gl_resource_manager.cpp
```diff
@@ -0,0 +1,116 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "video_core/renderer_opengl/gl_resource_manager.h"
+#include "video_core/renderer_opengl/gl_shader_util.h"
+
+// Textures
+OGLTexture::OGLTexture() : handle(0) {
+
```
bunnei: I made a comment about the extra space below for empty functions... I don't necessarily really care what you do, I'm guessing this is just what you're used to, but I think the codebase more commonly does not have a blank line.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -0,0 +1,174 @@
...
+    /// Syncs the specified TEV stage's constant color to match the PICA register
+    void SyncTevColor(int stage_index, const Pica::Regs::TevStageConfig& config);
+
+    /// Syncs the specified TEV stage's color and alpha multipliers to match the PICA register
+    void SyncTevMultipliers(int stage_index, const Pica::Regs::TevStageConfig& config);
+
+    /// Syncs the remaining OpenGL drawing state to match the current PICA state
+    void SyncDrawState();
+
+    /// Copies the 3ds color framebuffer into the OpenGL color framebuffer texture
```
bunnei: Search+replace `3ds` for `3DS`

bunnei:
Gave this another full review pass. Seems like you hit my major concerns from the RFC, just some minor nitpicks. I'm all set with getting this merged once those are addressed.

src/citra_qt/bootmanager.cpp
```diff
@@ -29,10 +29,11 @@
 EmuThread::EmuThread(GRenderWindow* render_window) :
     exec_step(false), running(false), stop_run(false), render_window(render_window) {
```
yuriks: Remove this other empty line too.

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -352,10 +355,16 @@ static void ExecuteCommand(const Command& command, u32 thread_id) {
...
     case CommandId::REQUEST_DMA:
+        VideoCore::g_renderer->hw_rasterizer->NotifyPreRead(Memory::VirtualToPhysicalAddress(command.dma_request.source_address),
+                                                            command.dma_request.size);
+
         memcpy(Memory::GetPointer(command.dma_request.dest_address),
                Memory::GetPointer(command.dma_request.source_address),
                command.dma_request.size);
         SignalInterrupt(InterruptId::DMA);
+
+        VideoCore::g_renderer->hw_rasterizer->NotifyFlush(Memory::VirtualToPhysicalAddress(command.dma_request.dest_address), command.dma_request.size);
```
yuriks: Huge line.

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -352,10 +355,16 @@ static void ExecuteCommand(const Command& command, u32 thread_id) {
...
+
```
yuriks: And I think you can skip this blank line before the break too.

src/core/hw/gpu.cpp
```diff
@@ -129,6 +131,11 @@ inline void Write(u32 addr, const T data) {
             u32 output_width = config.output_width / horizontal_scale;
             u32 output_height = config.output_height / vertical_scale;
 
+            u32 input_size = config.input_height * config.input_width * GPU::Regs::BytesPerPixel(config.input_format);
+            u32 output_size = output_height * output_width * GPU::Regs::BytesPerPixel(config.output_format);
```
yuriks: Since you're computing these vars, reuse them in the code below.

src/core/hw/gpu.cpp
```diff
@@ -142,6 +149,9 @@ inline void Write(u32 addr, const T data) {
                     config.output_format.Value(), config.flags);
 
                 GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::PPF);
+
+                VideoCore::g_renderer->hw_rasterizer->NotifyFlush(config.GetPhysicalOutputAddress(), output_size);
+
```
yuriks: Remove the newline.

src/video_core/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
 set(SRCS
             renderer_opengl/generated/gl_3_2_core.c
+            renderer_opengl/gl_rasterizer.cpp
+            renderer_opengl/gl_rasterizer_cache.cpp
             renderer_opengl/renderer_opengl.cpp
+            renderer_opengl/gl_resource_manager.cpp
             renderer_opengl/gl_shader_util.cpp
+            renderer_opengl/gl_state.cpp
```
yuriks: Sort these.

src/video_core/CMakeLists.txt
```diff
@@ -15,13 +19,19 @@ set(SRCS
...
             debug_utils/debug_utils.h
             renderer_opengl/generated/gl_3_2_core.h
+            renderer_opengl/gl_pica_to_gl.h
+            renderer_opengl/gl_rasterizer.h
+            renderer_opengl/gl_rasterizer_cache.h
+            renderer_opengl/gl_resource_manager.h
             renderer_opengl/gl_shader_util.h
             renderer_opengl/gl_shaders.h
+            renderer_opengl/gl_state.h
             renderer_opengl/renderer_opengl.h
```
yuriks: Likewise

src/video_core/CMakeLists.txt
```diff
@@ -15,13 +19,19 @@ set(SRCS
 set(HEADERS
             debug_utils/debug_utils.h
             renderer_opengl/generated/gl_3_2_core.h
+            renderer_opengl/gl_pica_to_gl.h
```
yuriks: This filename is confusing. Renamed to just `pica_to_gl.h`

src/video_core/CMakeLists.txt
```diff
@@ -15,13 +19,19 @@ set(SRCS
...
+            renderer_opengl/gl_rasterizer.h
+            renderer_opengl/gl_rasterizer_cache.h
+            renderer_opengl/gl_resource_manager.h
             renderer_opengl/gl_shader_util.h
             renderer_opengl/gl_shaders.h
+            renderer_opengl/gl_state.h
```
yuriks: Also, why do all these have a `gl_` prefix if they're already inside the `renderer_opengl/` directory?

src/video_core/renderer_opengl/gl_pica_to_gl.h
```diff
@@ -0,0 +1,97 @@
...
+inline GLenum WrapMode(Pica::Regs::TextureConfig::WrapMode mode) {
+    static const GLenum wrap_mode_table[] = {
+        GL_CLAMP_TO_EDGE,  // WrapMode::ClampToEdge
+        0,                 // Unknown
+        GL_REPEAT,         // WrapMode::Repeat
+        GL_MIRRORED_REPEAT // WrapMode::MirroredRepeat
+    };
+
+    // Range check table for input
+    if (mode >= sizeof(wrap_mode_table) / sizeof(GLenum)) {
```
yuriks: Use the `ARRAY_SIZE` macro in common.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+#include "video_core/renderer_opengl/gl_pica_to_gl.h"
+#include "video_core/renderer_opengl/gl_rasterizer.h"
+#include "video_core/renderer_opengl/gl_shaders.h"
+#include "video_core/renderer_opengl/gl_shader_util.h"
+
+#include "generated/gl_3_2_core.h"
+
+#include <memory>
+
+static bool IsNotPassThroughTevStage(const Pica::Regs::TevStageConfig& stage) {
```
yuriks: It's silly to have this function check for *Not* being a passthrough, and this tends to lead to confusing double-negatives in expressions.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+            stage.alpha_source1 != Pica::Regs::TevStageConfig::Source::Previous ||
+            stage.color_modifier1 != Pica::Regs::TevStageConfig::ColorModifier::SourceColor ||
+            stage.alpha_modifier1 != Pica::Regs::TevStageConfig::AlphaModifier::SourceAlpha ||
+            stage.GetColorMultiplier() != 1 ||
+            stage.GetAlphaMultiplier() != 1);
+}
+
+RasterizerOpenGL::RasterizerOpenGL() : last_fb_color_addr(0), last_fb_depth_addr(0) {
+
+}
```
yuriks: Closing brace on the same line and condense these two functions in two adjacent lines since they're empty.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+
+    if (color_buffer == nullptr)
+        return;
+
+    u32 bytes_per_pixel = Pica::registers.framebuffer.BytesPerColorPixel(fb_color_texture.format);
+
+    std::unique_ptr<u8[]> temp_fb_color_buffer(new u8[fb_color_texture.width * fb_color_texture.height * bytes_per_pixel]);
+
+    // Directly copy pixels. Internal OpenGL color formats are consistent so no conversion is necessary.
+    // TODO: Evaluate whether u16/memcpy/u32 is faster for 2/3/4 bpp versus memcpy for all
```
yuriks: memcpy automatically uses larger registers if possible (even SSE, etc.) and manual for loop copies are optimized to memcpies, so I can pretty confidently say that it won't make any difference.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+    PAddr cur_fb_depth_addr = Pica::registers.framebuffer.GetDepthBufferPhysicalAddress();
+    Pica::Regs::DepthFormat new_fb_depth_format = Pica::registers.framebuffer.depth_format;
+
+    bool fb_prop_changed = (fb_color_texture.format != new_fb_color_format ||
+                            fb_depth_texture.format != new_fb_depth_format ||
+                            fb_color_texture.width != Pica::registers.framebuffer.GetWidth() ||
+                            fb_color_texture.height != Pica::registers.framebuffer.GetHeight());
+    bool fb_modified = (last_fb_color_addr != cur_fb_color_addr ||
+                        last_fb_depth_addr != cur_fb_depth_addr ||
+                        fb_prop_changed);
```
yuriks: No need for the parenthesis.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -0,0 +1,174 @@
...
+
+        GLfloat position[4];
+        GLfloat color[4];
+        GLfloat tex_coord0[2];
+        GLfloat tex_coord1[2];
+        GLfloat tex_coord2[2];
+    };
+
+    /// Reconfigure the OpenGL color texture to use the given format and dimensions
+    void ReconfigureColorTexture(TextureInfo& texture, u32 format, u32 width, u32 height);
```
yuriks: Should this `format` also be an enum?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,818 @@
...
+
+void RasterizerOpenGL::DrawTriangles() {
+    SyncFramebuffer();
+    SyncDrawState();
+
+    glBufferData(GL_ARRAY_BUFFER, vertex_batch.size() * sizeof(HardwareVertex), vertex_batch.data(), GL_STREAM_DRAW);
+    glDrawArrays(GL_TRIANGLES, 0, (GLsizei)vertex_batch.size());
+
+    vertex_batch.clear();
+}
```
yuriks: I think you're missing a way to invalidate cached textures that have been rendered to, which is required for render-to-texture to work right. (Unless this is handled in some other way, but I couldn't find anything.)

src/video_core/renderer_opengl/gl_resource_manager.h
```diff
@@ -0,0 +1,109 @@
...
+
+class OGLTexture : public NonCopyable {
+public:
+    OGLTexture();
+    ~OGLTexture();
+
+    /// Returns the internal OpenGL resource handle for this resource
+    inline GLuint GetHandle() {
+        return handle;
+    }
```
yuriks: Make `handle` public and eliminate all these getters.

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -0,0 +1,159 @@
...
+    draw.framebuffer = 0;
+    draw.vertex_array = 0;
+    draw.vertex_buffer = 0;
+    draw.shader_program = 0;
+}
+
+const void OpenGLState::Apply() {
+    // Culling
+    if (cull.enabled) {
+        if (cull.enabled != cur_state.cull.enabled) {
```
yuriks: Pull this test to the outside so that you don't need to duplicated it.

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -0,0 +1,159 @@
...
+        if (cull.mode != cur_state.cull.mode) {
+            glCullFace(cull.mode);
+        }
+    } else if (cull.enabled != cur_state.cull.enabled) {
+        glDisable(GL_CULL_FACE);
+    }
+
+    // Depth test
+    if (depth.test_enabled) {
+        if (depth.test_enabled != cur_state.depth.test_enabled) {
```
yuriks: Same.

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    // Depth mask
+    if (depth.write_mask != cur_state.depth.write_mask) {
+        glDepthMask(depth.write_mask);
+    }
+
+    // Stencil test
+    if (stencil.test_enabled) {
+        if (stencil.test_enabled != cur_state.stencil.test_enabled) {
```
yuriks: Same.

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -175,7 +186,8 @@ void RendererOpenGL::LoadColorToActiveGLTexture(u8 color_r, u8 color_g, u8 color
  */
 void RendererOpenGL::InitOpenGLObjects() {
     glClearColor(Settings::values.bg_red, Settings::values.bg_green, Settings::values.bg_blue, 0.0f);
-    glDisable(GL_DEPTH_TEST);
+    state.depth.test_enabled = false;
```
yuriks: This doesn't do anything here. It was only originally used to disable depth tests for the duration of the program.

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -175,7 +186,8 @@ void RendererOpenGL::LoadColorToActiveGLTexture(u8 color_r, u8 color_g, u8 color
...
+    state.texture_units[0].enabled_2d = true;
```
yuriks: Why do you enable this here?

src/video_core/renderer_opengl/renderer_opengl.h
```diff
@@ -49,17 +51,17 @@ class RendererOpenGL : public RendererBase {
     };
 
     void InitOpenGLObjects();
-    static void ConfigureFramebufferTexture(TextureInfo& texture,
+    static void ConfigureFramebufferTexture(OpenGLState &state, TextureInfo& texture,
```
yuriks: IMO at this point there's no reason for these to be static since they use the `state` variable anyway.

yuriks:
Done with a full pass. Mostly trivial things. The only one I'm wondering about is the one to do with render-to-texture. I hadn't thought about it during RFC.

neobrain:
For reference, here's my list of things which need to be tested before merging this PR again:

- EmuWindow: Exit on close without introducing Qt runtime warnings; switching between single-window-mode and standalone render window; resizing behavior
- debugger: texture viewing, framebuffer viewing, CiTraces, geometry dumping
- citrace_player
- hotswitching gl/sw rasterizers
- multiple emulation runs within a single Citra session
- interaction with other pending or recently merged PRs

Most of these might have a nontrivial impact on the actual renderer interfaces, and hence should be considered early-on.

JMC47:
Multiple Emulation Runs within one session works.
Interaction with PRs like the fix-zalbw works
It's beyond me how to merge this with CiTrace player.
I also don't know how any of the debugging stuff works.

hrydgard:
neobrain:"Most of these might have a nontrivial impact on the actual renderer interfaces, and hence should be considered early-on"

As Citra, like all other sane emulators, is not plugin-based, renderer interfaces can be changed anytime as needed. This will inevitably happen no matter how much you plan ahead as you discover new things about the hardware :) So I wouldn't consider such things blockers, just things to keep in mind. 

Also, interaction with pending pull requests is nearly always out of scope for individual pull requests in my opinion. Merge one, then fix other conflicting ones to match.

bunnei:
Ditto with ector on the interoperability with other outstanding PRs... (CiTrace included). Renderer hot switching I'd also say is out of scope, since AFAIK hot switching of any of the config settings is unsupported. Debug widgets, however, is a valid concern.

neobrain:
@bunnei @ector: As for interaction with outstanding PRs, I still think it makes more sense to merge #702 and #776 first, especially since at least some nontrivial work is remaining for this PR and I'd furthermore give this one a test, too. In any case things will have to be patched up in the PRs which gets merged later. Also, adding CiTrace recording support is not a big deal, maybe it actually works already modulo merge conflicts.

I don't see any point in rushing a merge. Yes, some points are just small things and can be addressed easily later - but if it's really that small why not just take care of it right now and be done with it, instead of adding more PR review cycles than necessary?

yuriks:
Hot-switching renderers and breakpoints work flawlessly and overall the output of the HW renderer is identical or better than the SW one (no glitches as far as I could test). LGTM

JMC47:
Tested as well, everything working here.  Only difference between renderers I could notice is some odd zfighting in some games on HW renderer.  Nothing serious.

purpasmart96:
For me OoT instead of link being completely black (as in the entire character is just back) and the tilte screen being black, they are now white, along with trees in Tarmina Field being white.

src/citra_qt/main.cpp
```diff
@@ -300,6 +306,14 @@ void GMainWindow::OnOpenHotkeysDialog()
     dialog.exec();
 }
 
+void GMainWindow::ToggleHardwareRenderer() {
+    // TODO: Save new setting value to qt config file for next startup
+    Settings::values.use_hw_renderer = ui.action_Use_Hardware_Renderer->isChecked();
+
+    if (Settings::values.use_hw_renderer && VideoCore::g_renderer != nullptr) {
+        VideoCore::g_renderer->hw_rasterizer->Reset();
```
neobrain: Uhm, this is looking far from thread-safe. ToggleHardwareRenderer is called from the GUI thread, while Reset() affects EmuThread data.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,882 @@
...
+#include "video_core/renderer_opengl/gl_rasterizer.h"
+#include "video_core/renderer_opengl/gl_shaders.h"
+#include "video_core/renderer_opengl/gl_shader_util.h"
+#include "video_core/renderer_opengl/pica_to_gl.h"
+
+#include "generated/gl_3_2_core.h"
+
+#include <memory>
+
+// TODO: When a PICA breakpoint is hit, commit the current framebuffer (color + depth) to get per-batch results in framebuffer debug widget
```
neobrain: Doesn't this already work?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,882 @@
...
+// TODO: Use floating-point viewports for accuracy if supported
```
neobrain: Why not mention this at the place where it's actually relevant, rather than at some random place in the source?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -0,0 +1,882 @@
...
+// TODO: Add a qt menu item to hot-swap between sw/hw renderers (reload all PICA state into OGL and full resource flush when swapping to hw)
```
neobrain: This is already done, isn't it?

neobrain:
* There are three "odd" TODOs in the source
* sw/hw hotswitching is implemented in what looks like highly unsafe code. The GUI code should instead atomically set some flag, which the renderer checks at some "safe" point and then reacts appropriately.
* the GUI option for hw rendering would be a lot more suitably stored in the "Emulation" menu rather than in the "View" menu.

Other than that, I tried out the renderer and it seems to be working fine on my computer, nice :)

Anyway, that said...

    < bunnei_> neobrain: can we get a final sign off on https://github.com/citra-emu/citra/pull/789 today?
    < bunnei_> neobrain: I'd feel better if you get it a reluctant "fine"

Unless others feel like my outstanding comments are worth addressing, "fine".

tfarley:
Addressed neobrain's comments.

bunnei:
This is as good as it gets guys... Lgtm would merge twice

purpasmart96:
YES

yuriks:
:fireworks::tada::sparkles::balloon::balloon::balloon:

neobrain:
This thing is too fast, can't have the haters emulate at reasonable speed.

Revert plz.

yuriks:
Oh, that's right! We forgot the "slow emulation" anti-piracy mandate.

formatme:
nice wow

PR #792, opened 2015-05-20
Thread reschedule and MOFLEX fixes

bunnei:
This is a few fixes to get Zelda: A Link Between Worlds (boots to the menus, but lots of issues), Zelda: Majora's Mask (shows some graphics, still totally broken), and MOFLEX games (Aero Porter, Brunswick Pro Bowling, etc.) to boot (no video output, but it won't hang). The MOFLEX videos will still try to play, but they will appear as a black screen (and could take a while to compete). 

Obligatory Zelda screenshot:

image

Summary of changes:
* Cause a threading reschedule on underlying rescheduling mechanisms, not within the SVCs themselves (e.g. if Citra calls them, but not the game, they should still rechedule)
* Trigger the interrupt signal for Y2R conversions being completed (used for MOFLEX videos), even though they are unimplemented at this time

### DIFF
--- a/src/core/hle/kernel/address_arbiter.cpp
+++ b/src/core/hle/kernel/address_arbiter.cpp
@@ -79,6 +79,9 @@ ResultCode AddressArbiter::ArbitrateAddress(ArbitrationType type, VAddr address,
         LOG_ERROR(Kernel, "unknown type=%d", type);
         return ResultCode(ErrorDescription::InvalidEnumValue, ErrorModule::Kernel, ErrorSummary::WrongArgument, ErrorLevel::Usage);
     }
+
+    HLE::Reschedule(__func__);
+
     return RESULT_SUCCESS;
 }
 
--- a/src/core/hle/kernel/event.cpp
+++ b/src/core/hle/kernel/event.cpp
@@ -41,7 +41,10 @@ void Event::Acquire() {
 
 void Event::Signal() {
     signaled = true;
+
     WakeupAllWaitingThreads();
+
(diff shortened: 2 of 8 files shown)


archshift:
LGTM

src/video_core/command_processor.cpp
```diff
@@ -52,11 +52,6 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
     DebugUtils::OnPicaRegWrite(id, registers[id]);
 
     switch(id) {
-        // Trigger IRQ
```
Subv: Is this actually correct? what does trigger_irq do then?

purpasmart96:
So were moving the reschedules back to where they were before? Okay.

bunnei:
@purpasmart96 No they were never like this... could be true for perhaps one of them, but overall, rescheduling was mostly done in the SVC functions.

src/video_core/command_processor.cpp
bunnei: Here's my crappy answer: Just based on empirical evidence, more games seem to be fixed by this, while none that I know of were broken by it.
I suppose we should device a test for it, but I'm not sure right now how to go about that.

yuriks: I have the same question, what's up with this? This seems wrong and should definitely be tested.

yuriks:
LGTM

lioncash:
LGTM

Subv:
LGTM

PR #793, opened 2015-05-21
Implement stencil testing

neobrain:
Implements basic stencil testing functionality. Also adds two new framebuffer viewing formats: D24X8 and X24S8, each of which spans 32 bits per pixel but only shows the depth and/or stencil values.

The code in this PR was tested against portal3DS, which uses stencil testing to implement "looking through portals". Here's a screenshot showing it in action, as well as the X24S8 framebuffer viewing format (showing that the area covered by a portal is setting a mask for rendering anything beyond the portal).

stencil testing in action 

Revision history:
* ~~fa770594a29d123ca5612c91603aaf1e13f932be~~
* 5e79706db23a956ea0d6f3e5cc5a57b3a5d86b5e

### DIFF
--- a/src/citra_qt/debugger/graphics_framebuffer.cpp
+++ b/src/citra_qt/debugger/graphics_framebuffer.cpp
@@ -55,7 +55,9 @@ GraphicsFramebufferWidget::GraphicsFramebufferWidget(std::shared_ptr<Pica::Debug
     framebuffer_format_control->addItem(tr("RGBA4"));
     framebuffer_format_control->addItem(tr("D16"));
     framebuffer_format_control->addItem(tr("D24"));
-    framebuffer_format_control->addItem(tr("D24S8"));
+    framebuffer_format_control->addItem(tr("D24X8"));
+    framebuffer_format_control->addItem(tr("X24S8"));
+    framebuffer_format_control->addItem(tr("(unknown)"));
 
     // TODO: This QLabel should shrink the image to the available space rather than just expanding...
     framebuffer_picture_label = new QLabel;
@@ -221,7 +223,24 @@ void GraphicsFramebufferWidget::OnUpdate()
         framebuffer_address = framebuffer.GetDepthBufferPhysicalAddress();
         framebuffer_width = framebuffer.GetWidth();
         framebuffer_height = framebuffer.GetHeight();
-        framebuffer_format = Format::D16;
+
+        switch (framebuffer.depth_format) {
+        case Pica::Regs::DepthFormat::D16:
+            framebuffer_format = Format::D16;
+            break;
+
+        case Pica::Regs::DepthFormat::D24:
+            framebuffer_format = Format::D24;
+            break;
+
+        case Pica::Regs::DepthFormat::D24S8:
+            framebuffer_format = Format::D24X8;
+            break;
+
+        default:
+            framebuffer_format = Format::Unknown;
+            break;
+        }
 
         break;
     }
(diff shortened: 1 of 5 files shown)


src/citra_qt/debugger/graphics_framebuffer.cpp
```diff
@@ -265,6 +309,20 @@ void GraphicsFramebufferWidget::OnUpdate()
                 color.b() = (data.x >> 16) & 0xFF;
```
neobrain: this should take data.y into account, now that it's been replaced by D24X8.

src/video_core/pica.h
```diff
@@ -431,15 +431,26 @@ struct Regs {
...
+            union {
+                BitField< 0, 1, u32> enable;
+                BitField< 4, 3, CompareFunc> func;
+                BitField< 8, 8, u32> replacement_value;
+                BitField<16, 8, u32> reference_value;
+                BitField<24, 8, u32> mask;
+            };
+
+            union {
+                BitField< 0, 3, StencilAction> action_sfail;
```
neobrain: pick better variable names for these

src/video_core/pica.h
```diff
@@ -431,15 +431,26 @@ struct Regs {
...
-            BitField<24, 8, u32> stencil_mask;
-        } stencil_test;
+        struct {
+            enum StencilAction : u32 {
+                Keep = 0,
+                Xor  = 5,
+            };
 
-        INSERT_PADDING_WORDS(0x1);
+            union {
```
neobrain: some documentation would be helpful here

src/video_core/rasterizer.cpp
```diff
@@ -123,6 +123,29 @@ static u32 GetDepth(int x, int y) {
...
+
+    u32 src_offset = VideoCore::GetMortonOffset(x, y, bytes_per_pixel) + coarse_y * stride;
+    u8* src_pixel = depth_buffer + src_offset;
+
+    switch (registers.framebuffer.depth_format) {
+        case Pica::Regs::DepthFormat::D24S8:
+            return Color::DecodeD24S8(src_pixel).y;
+
+        default:
+            LOG_CRITICAL(HW_GPU, "GetStencil called for function which doesn't have a stencil component (format %u)", registers.framebuffer.depth_format);
```
neobrain: This is far from critical.

src/video_core/rasterizer.cpp
```diff
@@ -312,6 +370,132 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                 return interpolated_attr_over_w * interpolated_w_inverse;
             };
 
+            // TODO: Should the stencil mask be applied for these? Most likely not!
```
neobrain: be more specific: What's "these"?

Subv:
I don't like how this is introducing yet more inline lambdas into the code, but hey, that's just me

src/video_core/rasterizer.cpp
```diff
@@ -312,6 +370,132 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+
+                if (!pass) {
+                    u8 new_stencil = PerformStencilAction(stencil_test.action_sfail, old_stencil, stencil_test.replacement_value);
+                    SetStencil(x >> 4, y >> 4, new_stencil);
+                    continue;
+                }
+            }
+
+            // TODO: Does depth indeed only get written even if depth testing is enabled?
+            if (registers.output_merger.depth_test_enable) {
```
neobrain: This was incorrectly moved to happen before alpha testing. Hence, this might modify the depthstencil buffer even when the alpha test fails.

src/citra_qt/debugger/graphics_framebuffer.cpp
Subv: I think it doesn't make much sense to keep this case, is there any situation where you'd want to visualize the Depth and Stencil buffers at the same time?

Subv:
What's the status of this PR?

src/citra_qt/debugger/graphics_framebuffer.cpp
neobrain: > I think it doesn't make much sense to keep this case, is there any situation where you'd want to visualize the Depth and Stencil buffers at the same time?

Not really. I'll remove it.

src/video_core/pica.h
```diff
@@ -431,15 +431,26 @@ struct Regs {
...
 
-        union {
-            BitField< 0, 1, u32> stencil_test_enable;
-            BitField< 4, 3, CompareFunc> stencil_test_func;
-            BitField< 8, 8, u32> stencil_replacement_value;
-            BitField<16, 8, u32> stencil_reference_value;
-            BitField<24, 8, u32> stencil_mask;
-        } stencil_test;
+        struct {
+            enum StencilAction : u32 {
```
neobrain: Move this out of the struct.

neobrain:
@Subv  It's done.

PR #795, opened 2015-05-21
dyncom: Eliminate clang warnings

lioncash:
Gets rid of a whole load of missing brace initialization warnings.

200+ warnings out the window.


yuriks:
LGTM

PR #797, opened 2015-05-21
Implement blended downscaling for display transfers

linkmauve:
And a bit of cleanup before that.

I only tested it on the according test in hwtests, which is now passing: https://github.com/citra-emu/hwtests/blob/master/source/tests/gpu/displaytransfer.cpp#L497

### DIFF
--- a/src/core/hw/gpu.cpp
+++ b/src/core/hw/gpu.cpp
@@ -53,6 +53,29 @@ inline void Read(T &var, const u32 raw_addr) {
     var = g_regs[addr / 4];
 }
 
+static Math::Vec4<u8> DecodePixel(Regs::PixelFormat input_format, const u8* src_pixel) {
+    switch (input_format) {
+    case Regs::PixelFormat::RGBA8:
+        return Color::DecodeRGBA8(src_pixel);
+
+    case Regs::PixelFormat::RGB8:
+        return Color::DecodeRGB8(src_pixel);
+
+    case Regs::PixelFormat::RGB565:
+        return Color::DecodeRGB565(src_pixel);
+
+    case Regs::PixelFormat::RGB5A1:
+        return Color::DecodeRGB5A1(src_pixel);
+
+    case Regs::PixelFormat::RGBA4:
+        return Color::DecodeRGBA4(src_pixel);
+
+    default:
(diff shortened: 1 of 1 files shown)


src/core/hw/gpu.cpp
```diff
@@ -185,30 +206,15 @@ inline void Write(u32 addr, const T data) {
...
-
-                    default:
-                        LOG_ERROR(HW_GPU, "Unknown source framebuffer format %x", config.input_format.Value());
-                        break;
+                    src_color = DecodePixel(config.input_format, src_pixel);
+                    if (config.scaling == config.ScaleX) {
+                        Math::Vec4<u8> pixel = DecodePixel(config.input_format, src_pixel + src_bytes_per_pixel);
+                        src_color = ((src_color + pixel) / 2).Cast<u8>();
+                    } else if (config.scaling == config.ScaleXY) {
+                        Math::Vec4<u8> pixel1 = DecodePixel(config.input_format, src_pixel + 1 * src_bytes_per_pixel);
```
Subv: This works for the tiled (Morton) format, but what about the normal format, where all four blended pixels aren't horizontally contiguous?

linkmauve: I added a guard against that case, since I can’t imagine the GPU doing vertical scaling on a non-tiled format, but maybe under certain conditions it could.

Subv: We should write a hwtest for that

yuriks:
Agree that we need a hwtest for the linear 2x2 downscaling case.

Subv:
What's the status of this PR?

src/core/hw/gpu.cpp
linkmauve: I just wrote a test for that, the GPU is then downscaling from four horizontal pixels instead of a square.

PR incoming to hwtests.

linkmauve: Actually it crashes the GPU, see https://github.com/citra-emu/hwtests/pull/35

linkmauve:
I added a LOG_CRITICAL for the case where a downscale operation is done on a linear input, which crashes the PICA.

This PR is ready to go.

Subv:
LGTM

bunnei:
Lgtm

PR #798, opened 2015-05-22
Service::Y2R: Support for grayscale decoding of specific formats

yuriks:
Implements unrotated planar YUV 4:2:0 -> RGB24 conversions in Y2R.
Currently only the Y (luma) channel is used, so the results don't
contain color. This will be added in a later PR at some point.

This is enough to get all currently know Moflex videos to decode. (Some
don't display on-screen due to seemingly unrelated reasons.)

Thanks to @archshift for doing the initial implementation which I
cleaned up and then fixed the 8x8 block mode.

### DIFF
--- a/src/core/hle/service/y2r_u.cpp
+++ b/src/core/hle/service/y2r_u.cpp
@@ -1,85 +1,314 @@
-// Copyright 2014 Citra Emulator Project
+// Copyright 2015 Citra Emulator Project
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
+
+#include <cstring>
 
 #include "common/logging/log.h"
 
 #include "core/hle/hle.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/service/y2r_u.h"
+#include "video_core/utils.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Namespace Y2R_U
 
 namespace Y2R_U {
 
+enum class InputFormat {
+    /// 8-bit input, with YUV components in separate planes and using 4:2:2 subsampling.
+    YUV422_Indiv8 = 0,
+    /// 8-bit input, with YUV components in separate planes and using 4:2:0 subsampling.
+    YUV420_Indiv8 = 1,
+
+    YUV422_INDIV_16 = 2,
+    YUV420_INDIV_16 = 3,
+    YUV422_BATCH = 4,
+};
+
+enum class OutputFormat {
+    Rgb32 = 0,
+    Rgb24 = 1,
+    Rgb16_555 = 2,
+    Rgb16_565 = 3,
+};
+
+enum class Rotation {
+    None = 0,
+    Clockwise_90 = 1,
+    Clockwise_180 = 2,
+    Clockwise_270 = 3,
+};
+
+enum class BlockAlignment {
+    /// Image is output in linear format suitable for use as a framebuffer.
+    Linear = 0,
+    /// Image is output in tiled PICA format, suitable for use as a texture.
+    Block8x8 = 1,
+};
+
+enum class StandardCoefficient {
+    ITU_Rec601 = 0,
+    ITU_Rec709 = 1,
+    ITU_Rec601_Scaling = 2,
+    ITU_Rec709_Scaling = 3,
+};
+
 static Kernel::SharedPtr<Kernel::Event> completion_event;
 
-/**
- * Y2R_U::IsBusyConversion service function
- *  Outputs:
- *      1 : Result of function, 0 on success, otherwise error code
- *      2 : Whether the current conversion is of type busy conversion (?)
- */
-static void IsBusyConversion(Service::Interface* self) {
+struct ConversionParameters {
+    InputFormat input_format;
+    OutputFormat output_format;
+    Rotation rotation;
+    BlockAlignment alignment;
+    u16 input_line_width;
+    u16 input_lines;
+
+    // Input parameters for the Y (luma) plane
+    VAddr srcY_address;
+    u32 srcY_image_size;
+    u16 srcY_transfer_unit;
+    u16 srcY_stride;
+
+    // Output parameters for the conversion results
+    VAddr dst_address;
+    u32 dst_image_size;
+    u16 dst_transfer_unit;
+    u16 dst_stride;
+};
+
+static ConversionParameters conversion_params;
+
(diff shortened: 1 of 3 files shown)


Subv:
I think you should reset those variables to some default value when the module is initialized

yuriks:
As far as I've seen, all games set all the registers before a conversion, but fair point.

yuriks:
I moved all the conversion parameters into a struct so that it can be easily memsetted. This also gives me an excuse to move all params used in the conversion to local variables, which should help compile analysis and help generate faster code. While at it I also changed all variables to `size_t` which also helps generate better code in x64 (due to `int` being different from the native register size).

src/core/hle/service/y2r_u.cpp
```diff
@@ -2,84 +2,307 @@
 // Licensed under GPLv2 or any later version
```
Subv: I'd say it's fair to change the copyright year to 2015 now

bunnei: +1

src/core/hle/service/y2r_u.cpp
```diff
@@ -2,84 +2,307 @@
...
-    // otherwise hang on trying to play moflex videos, which uses the Y2R service.
+    const u8* srcY_buffer = Memory::GetPointer(conversion_params.srcY_address);
+    u8* dst_buffer = Memory::GetPointer(conversion_params.dst_address);
+
+    ASSERT(conversion_params.input_format == InputFormat::INPUT_YUV420_INDIV_8
+        || conversion_params.input_format == InputFormat::INPUT_YUV422_INDIV_8);
+    ASSERT(conversion_params.output_format == OutputFormat::OUTPUT_RGB_24);
+    ASSERT(conversion_params.rotation == Rotation::ROTATION_NONE);
+
+    const int bpp = 3;
```
archshift: Add a TODO here

yuriks:
Updated.

archshift:
LGTM

Subv:
LGTM

src/core/hle/service/y2r_u.cpp
```diff
@@ -1,85 +1,308 @@
...
 #include "core/hle/kernel/event.h"
 #include "core/hle/service/y2r_u.h"
+#include "video_core/utils.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Namespace Y2R_U
 
 namespace Y2R_U {
 
+enum InputFormat {
```
bunnei: Can you change all of these to `enum class ClassName : u32 {`, then get rid of the duplicate prefixes (e.g. `INPUT_`, `OUTPUT_`, `ROTATION_`, ...)

bunnei: Also, some documentation would be nice (and/or in the struct below)

src/core/hle/service/y2r_u.cpp
```diff
@@ -1,85 +1,308 @@
...
 
-    LOG_WARNING(Service, "(STUBBED) called");
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+}
+
+static void SetOutputFormat(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    conversion_params.output_format = static_cast<OutputFormat>(cmd_buff[1]);
+    LOG_DEBUG(Service, "Y2R OutputFormat: %u", conversion_params.output_format);
```
bunnei: Can we change this `LOG_DEBUG` (and the one's below to be consistent) with how we log service routines? E.g. replace the `Y2R` (we know what service this called from..) with `called`.

src/core/hle/service/y2r_u.cpp
```diff
@@ -1,85 +1,308 @@
...
+
+static void SetSendingY(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    conversion_params.srcY_address = cmd_buff[1];
+    conversion_params.srcY_image_size = cmd_buff[2];
+    conversion_params.srcY_transfer_unit = cmd_buff[3];
+    conversion_params.srcY_stride = cmd_buff[4];
+    u32 src_process_handle = cmd_buff[6];
+    LOG_DEBUG(Service, "called! image_size: 0x%08X, transfer_unit: %hu, transfer_stride: %hu, "
```
bunnei: `called!`? Ditto format with above^

src/core/hle/service/y2r_u.cpp
```diff
@@ -1,85 +1,308 @@
...
+    ASSERT(conversion_params.output_format == OutputFormat::OUTPUT_RGB_24);
+    ASSERT(conversion_params.rotation == Rotation::ROTATION_NONE);
+    const int bpp = 3;
+
+    switch (conversion_params.alignment) {
+    case BLOCK_LINE:
+    {
+        const size_t input_lines = conversion_params.input_lines;
+        const size_t input_line_width = conversion_params.input_line_width;
+        const size_t srcY_stride = conversion_params.srcY_stride;
```
bunnei: `srcY_stride` ?

src/core/hle/service/y2r_u.cpp
```diff
@@ -1,85 +1,308 @@
...
+    OutputFormat output_format;
+    Rotation rotation;
+    BlockAlignment alignment;
+    u16 input_line_width;
+    u16 input_lines;
+
+    VAddr srcY_address;
+    u32 srcY_image_size;
+    u16 srcY_transfer_unit;
+    u16 srcY_stride;
```
bunnei: `srcY` "is not the preferred nomenclature, dude"

bunnei:
Just come stylistic comments, please address & squash

src/core/hle/service/y2r_u.cpp
yuriks: These enums were taken from the BPB symbol table. Still change them?

src/core/hle/service/y2r_u.cpp
archshift: I also just noticed this, but `Service` should instead be some specific service log format, like `Service_Y2R`

src/core/hle/service/y2r_u.cpp
yuriks: "Y" here stands for the luma channel in YUV, which is always written like that. I'll keep it and add comments.

yuriks:
Updated

bunnei:
LGTM

PR #805, opened 2015-05-23
video_core/core: Get rid of more warnings.

lioncash:


src/core/hle/service/y2r_u.cpp
```diff
@@ -236,7 +236,6 @@ static void StartConversion(Service::Interface* self) {
         const size_t tile_size = 8 * 8 * bpp;
 
         for (size_t line = 0; line < input_lines;) {
-            size_t tile_y = line / 8;
```
archshift: @yuriks was there any reason you added this?

yuriks:
```
[01:44:31] <Lioncash> yuriks: Should tile_y be used here somewhere, or is it removable: https://github.com/citra-emu/citra/blob/master/src/core/hle/service/y2r_u.cpp#L239 ?
[01:45:23] <yuriks> Lioncash: no, it's fine
[01:45:35] <yuriks> I was using it before but switched to incrementing the offset instead
```

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -43,7 +43,7 @@ OpenGLState::OpenGLState() {
     draw.shader_program = 0;
 }
 
-const void OpenGLState::Apply() {
```
yuriks: What warning was this?

yuriks:
LGTM

archshift:
LGTM

PR #806, opened 2015-05-23
Qt: Silence a bogus warning printed when using the debug runtime

yuriks:
The Qt debug runtime prints a bogus warning on the console if you
haven't called makeCurrent since the last time you called swapBuffers.
This presumably means something if you're using QGLWidget the "regular"
way, but in our multi-threaded use case is harmless since we never call
doneCurrent in the rendering thread.


Subv:
LGTM

lioncash:
LGTM

PR #811, opened 2015-05-24
Commonify video_core utility headers

archshift:
These two headers (`color.h` and `math.h`) both are very useful outside of video_core, and don't assume anything inherent about the 3DS or any of its internals. Therefore, they should be in common instead. 

I renamed `math.h` to `vector_math.h` because it only contained code for doing vector math anyway.


src/video_core/debug_utils/debug_utils.h
```diff
@@ -12,7 +12,7 @@
 #include <mutex>
 #include <vector>
 
-#include "video_core/math.h"
+#include "common/vector_math.h"
```
linkmauve: Add a blank line here, as it’s done elsewhere to separate common from video_core.

linkmauve:
LGTM.

src/video_core/debug_utils/debug_utils.h
archshift: done

src/video_core/pica.h
```diff
@@ -14,10 +14,9 @@
 #include "common/bit_field.h"
 #include "common/common_funcs.h"
 #include "common/common_types.h"
+#include "common/vector_math.h"
```
linkmauve: Wrong order of include, vector comes after logging.

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -11,9 +11,9 @@
 #include <QSpinBox>
 #include <QComboBox>
 
-#include "video_core/pica.h"
-#include "video_core/math.h"
+#include "common/vector_math.h"
 
+#include "video_core/pica.h"
 #include "video_core/debug_utils/debug_utils.h"
```
yuriks: These two lines aren't sorted.

src/video_core/pica.h
yuriks: +1

yuriks:
LGTM

neobrain:
LGTM

PR #816, opened 2015-05-25
QT: Updated the QMenuBar to follow Windows 10's design guidelines

filfat:
The tabs in Citra are currently gray which makes it feel quite out of place when using it on Windows 10 Insider Preview, this is easily fixed by simply making the background white.


MaJoRoesch:
Pics? Most people are not using Windows 10 yet...

filfat:
@MaJoRoesch My Surface just died (don't have a charger at work) :/, just imagine the tabs with white background instead of gray. ;)

filfat:
The Travis build failed because of timing out when build for OSX, but works fine for Linux. ignore that.

neobrain:
Uhm, this looks wrong. Are you sure this is not actually a Qt issue? Manually specifying style properties rarely is a good idea and seems like the wrong solution here. Seems pretty likely that it's a Qt issue, since Windows 10 is not even final yet, and hence Qt support for that platform is expected to be buggy.

What effects would this have on other platforms? Which Qt version did you compile against?

filfat:
I used the QT designer, and all builds passed on appveyor I'm sure if a Mac owner tries to compile this fork it'll work.

Currently it changes the tab background on all platforms that uses it, I'm not sure if Mac uses a Windows Styled tab menu for QT Applications or if it places the tabs at the top of the screen (if that's the case Mac users won't notice anything). Linux uses will also get the white background on tabs.

yuriks:
Screenshot:
white-menubar

I'm sorry, but this simply looks wrong to me. I've never seen an application using this style and it simply looks out of place.

neobrain:
Yeah, I don't think this is working.

Let's wait until official Qt support for the final Windows 10 release materializes.

PR #817, opened 2015-05-25
Move src/assets/citra.ico and doc-icon.png to dist

linkmauve:
It was confusing to have a binary file alongside actual source files.

I tested the doxygen change, but I’d like a Windows user to confirm I didn’t mess anything up with the rc files.


yuriks:
LGTM

PR #820, opened 2015-05-25
Core/SVC: Map the shared memory created in CreateMemoryBlock to the specified address

Subv:
This SharedMemory can be passed to service functions (Which should map the memory into their own address space).

Since our services all share the same address space as the running application, all we need to do is assign the base_address of the SharedMemory.


yuriks:
LGTM

lioncash:
LGTM

bunnei:
LGTM

PR #821, opened 2015-05-25
Service/GSP: Implemented ImportDisplayCaptureInfo.

Subv:
Used by applications when preparing to launch an applet, leaving this unimplemented would lead to the application reading garbage from the buffers

### DIFF
--- a/src/core/hle/service/gsp_gpu.cpp
+++ b/src/core/hle/service/gsp_gpu.cpp
@@ -496,6 +496,52 @@ static void TriggerCmdReqQueue(Service::Interface* self) {
     cmd_buff[1] = 0; // No error
 }
 
+/**
+ * GSP_GPU::ImportDisplayCaptureInfo service function
+ *
+ * Returns information about the current framebuffer state
+ *
+ *  Inputs:
+ *      0: Header 0x00180000
+ *  Outputs:
+ *      1: Result code
+ *      2: Left framebuffer virtual address for the main screen
+ *      3: Right framebuffer virtual address for the main screen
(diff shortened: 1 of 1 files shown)


archshift:
LGTM

lioncash:
LGTM

yuriks:
I'm reversing the GSP module to verify that this is correct, hold on.

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -496,6 +496,52 @@ static void TriggerCmdReqQueue(Service::Interface* self) {
...
+ *      9: Bottom screen framebuffer width
+ */
+
+static void ImportDisplayCaptureInfo(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    
+    // TODO(Subv): We're always returning the framebuffer structures for thread_id = 0, 
+    // because we only support a single call to RegisterInterruptRelayQueue. 
+    // This should always return the framebuffer data for the thread_id of the caller process, 
+    // which was set by the GSP during the call to RegisterInterruptRelayQueue.
```
neobrain: IMO it would be good to explicitly introduce a local variable called `thread_id` here, so that once we address this TODO we only need to insert the proper value, instead of figuring out which of the code below needs to be fixed.

yuriks:
The data read to assemble the command response is read from a global struct, so I suspect this simply returns the info about the framebuffer that's currently being displayed on-screen.

Subv:
Updated, LGTM, instamergepls

bunnei:
Lgtm

PR #822, opened 2015-05-25
Pica improvements

bunnei:
This is a combination of several Pica emulation improvements. All changes have been implemented for both the SW rasterizer and OpenGL rasterizer. I recommend reviewing each commit separately, as they are well separated. This PR implements:
* Pica command buffer execution registers
* Vertex shader MIN, SLT, and SLTI instructions
* Color "logic operation" (same as `glLogicOp`)
* ~~RG8 texture format~~
* AddSigned combiner function for the alpha channel

Here are some results of these features (although I'm sure there are many more, testing welcome!):

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -56,7 +56,17 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
         // Trigger IRQ
         case PICA_REG_INDEX(trigger_irq):
             GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::P3D);
-            return;
+            break;
+
+        case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[0], 0x23c):
+        case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[1], 0x23d):
+        {
+            unsigned index = id - PICA_REG_INDEX(command_buffer.trigger[0]);
+            u32* head_ptr = (u32*)Memory::GetPhysicalPointer(regs.command_buffer.GetPhysicalAddress(index));
+            g_state.cmd_list.head_ptr = g_state.cmd_list.current_ptr = head_ptr;
+            g_state.cmd_list.length = regs.command_buffer.GetSize(index) / sizeof(u32);
+            break;
+        }
 
         // It seems like these trigger vertex rendering
         case PICA_REG_INDEX(trigger_draw):
@@ -363,38 +373,34 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
         g_debug_context->OnEvent(DebugContext::Event::CommandProcessed, reinterpret_cast<void*>(&id));
 }
 
-static std::ptrdiff_t ExecuteCommandBlock(const u32* first_command_word) {
-    const CommandHeader& header = *(const CommandHeader*)(&first_command_word[1]);
-
-    u32* read_pointer = (u32*)first_command_word;
-
-    const u32 write_mask = ((header.parameter_mask & 0x1) ? (0xFFu <<  0) : 0u) |
-                           ((header.parameter_mask & 0x2) ? (0xFFu <<  8) : 0u) |
-                           ((header.parameter_mask & 0x4) ? (0xFFu << 16) : 0u) |
-                           ((header.parameter_mask & 0x8) ? (0xFFu << 24) : 0u);
-
-    WritePicaReg(header.cmd_id, *read_pointer, write_mask);
-    read_pointer += 2;
-
-    for (unsigned int i = 1; i < 1+header.extra_data_length; ++i) {
-        u32 cmd = header.cmd_id + ((header.group_commands) ? i : 0);
-        WritePicaReg(cmd, *read_pointer, write_mask);
-        ++read_pointer;
-    }
-
-    // align read pointer to 8 bytes
-    if ((first_command_word - read_pointer) % 2)
-        ++read_pointer;
-
-    return read_pointer - first_command_word;
-}
-
 void ProcessCommandList(const u32* list, u32 size) {
-    u32* read_pointer = (u32*)list;
-    u32 list_length = size / sizeof(u32);
+    g_state.cmd_list.head_ptr = g_state.cmd_list.current_ptr = list;
+    g_state.cmd_list.length = size / sizeof(u32);
 
-    while (read_pointer < list + list_length) {
-        read_pointer += ExecuteCommandBlock(read_pointer);
+    while (g_state.cmd_list.current_ptr < g_state.cmd_list.head_ptr + g_state.cmd_list.length) {
+        // Expand a 4-bit mask to 4-byte mask, e.g. 0b0101 -> 0x00FF00FF
+        static const u32 expand_bits_to_bytes[] = {
+            0x00000000, 0x000000ff, 0x0000ff00, 0x0000ffff,
+            0x00ff0000, 0x00ff00ff, 0x00ffff00, 0x00ffffff,
+            0xff000000, 0xff0000ff, 0xff00ff00, 0xff00ffff,
+            0xffff0000, 0xffff00ff, 0xffffff00, 0xffffffff
+        };
+
+        // Align read pointer to 8 bytes
+        if ((g_state.cmd_list.head_ptr - g_state.cmd_list.current_ptr) % 2 != 0)
+            ++g_state.cmd_list.current_ptr;
+
+        u32 value = *g_state.cmd_list.current_ptr++;
+        const CommandHeader header = { *g_state.cmd_list.current_ptr++ };
+        const u32 write_mask = expand_bits_to_bytes[header.parameter_mask];
+        u32 cmd = header.cmd_id;
+
+        WritePicaReg(cmd, value, write_mask);
+
+        for (unsigned i = 0; i < header.extra_data_length; ++i) {
+            u32 cmd = header.cmd_id + (header.group_commands ? i + 1 : 0);
+            WritePicaReg(cmd, *g_state.cmd_list.current_ptr++, write_mask);
+         }
     }
 }
 
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -162,6 +162,25 @@ struct Regs {
         ETC1A4       = 13,  // compressed
     };
 
+    enum class LogicOp : u32 {
+        Clear        =  0,
+        And          =  1,
+        AndReverse   =  2,
+        Copy         =  3,
+        Set          =  4,
+        CopyInverted =  5,
+        NoOp         =  6,
+        Invert       =  7,
+        Nand         =  8,
+        Or           =  9,
+        Nor          = 10,
+        Xor          = 11,
+        Equiv        = 12,
+        AndInverted  = 13,
+        OrReverse    = 14,
+        OrInverted   = 15,
+    };
+
     static unsigned NibblesPerPixel(TextureFormat format) {
         switch (format) {
         case TextureFormat::RGBA8:
@@ -413,12 +432,8 @@ struct Regs {
         } alpha_blending;
 
         union {
-            enum Op {
-                Set = 4,
-            };
-
-            BitField<0, 4, Op> op;
-        } logic_op;
+            BitField<0, 4, LogicOp> logic_op;
+        };
 
         union {
             BitField< 0, 8, u32> r;
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -708,7 +708,33 @@ struct Regs {
         u32 set_value[3];
     } vs_default_attributes_setup;
 
-    INSERT_PADDING_WORDS(0x28);
+    INSERT_PADDING_WORDS(0x2);
+
+    struct {
+        // There are two channels that can be used to configure the next command buffer, which
+        // can be then executed by writing to the "trigger" registers. There are two reasons why a
+        // game might use this feature:
+        //  1) With this, an arbitrary number of additional command buffers may be executed in
+        //     sequence without requiring any intervention of the CPU after the initial one is
+        //     kicked off.
+        //  2) Games can configure these registers to provide a command list subroutine mechanism.
+
+        BitField< 0, 20, u32> size[2]; ///< Size (in bytes / 8) of each channel's command buffer
+        BitField< 0, 28, u32> addr[2]; ///< Physical address / 8 of each channel's command buffer
+        u32 trigger[2]; ///< Triggers execution of the channel's command buffer when written to
+
+        unsigned GetSize(unsigned index) const {
+            ASSERT(index < 2);
+            return 8 * size[index];
+        }
+
+        PAddr GetPhysicalAddress(unsigned index) const {
+            ASSERT(index < 2);
+            return (PAddr)(8 * addr[index]);
+        }
+    } command_buffer;
+
+    INSERT_PADDING_WORDS(0x20);
 
     enum class TriangleTopology : u32 {
         List        = 0,
@@ -861,6 +887,7 @@ struct Regs {
         ADD_FIELD(trigger_draw);
         ADD_FIELD(trigger_draw_indexed);
         ADD_FIELD(vs_default_attributes_setup);
+        ADD_FIELD(command_buffer);
         ADD_FIELD(triangle_topology);
         ADD_FIELD(vs_bool_uniforms);
         ADD_FIELD(vs_int_uniforms);
@@ -938,6 +965,7 @@ ASSERT_REG_POSITION(num_vertices, 0x228);
 ASSERT_REG_POSITION(trigger_draw, 0x22e);
 ASSERT_REG_POSITION(trigger_draw_indexed, 0x22f);
 ASSERT_REG_POSITION(vs_default_attributes_setup, 0x232);
+ASSERT_REG_POSITION(command_buffer, 0x238);
 ASSERT_REG_POSITION(triangle_topology, 0x25e);
 ASSERT_REG_POSITION(vs_bool_uniforms, 0x2b0);
 ASSERT_REG_POSITION(vs_int_uniforms, 0x2b1);
@@ -1053,21 +1081,12 @@ private:
     float value;
 };
 
-union CommandHeader {
-    CommandHeader(u32 h) : hex(h) {}
-
-    u32 hex;
-
-    BitField< 0, 16, u32> cmd_id;
-    BitField<16,  4, u32> parameter_mask;
-    BitField<20, 11, u32> extra_data_length;
-    BitField<31,  1, u32> group_commands;
-};
-
 /// Struct used to describe current Pica state
 struct State {
+    /// Pica registers
     Regs regs;
 
+    /// Vertex shader memory
     struct {
         struct {
             Math::Vec4<float24> f[96];
@@ -1080,6 +1099,13 @@ struct State {
         std::array<u32, 1024> program_code;
         std::array<u32, 1024> swizzle_data;
     } vs;
+
+    /// Current Pica command list
+    struct {
+        const u32* head_ptr;
+        const u32* current_ptr;
+        u32 length;
+    } cmd_list;
 };
 
 /// Initialize Pica state
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -240,6 +240,7 @@ struct Regs {
         enum class Source : u32 {
             PrimaryColor           = 0x0,
             PrimaryFragmentColor   = 0x1,
+            SecondaryFragmentColor = 0x2,
 
             Texture0               = 0x3,
             Texture1               = 0x4,
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -873,8 +873,63 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                 blend_output     = EvaluateBlendEquation(combiner_output, srcfactor, dest, dstfactor, params.blend_equation_rgb);
                 blend_output.a() = EvaluateBlendEquation(combiner_output, srcfactor, dest, dstfactor, params.blend_equation_a).a();
             } else {
-                LOG_CRITICAL(HW_GPU, "logic op: %x", output_merger.logic_op);
-                UNIMPLEMENTED();
+                static auto LogicOp = [](u8 src, u8 dest, Regs::LogicOp op) -> u8 {
+                    switch (op) {
+                    case Regs::LogicOp::Clear:
+                        return 0;
+
+                    case Regs::LogicOp::And:
+                        return src & dest;
+
+                    case Regs::LogicOp::AndReverse:
+                        return src & ~dest;
+
+                    case Regs::LogicOp::Copy:
+                        return src;
+
+                    case Regs::LogicOp::Set:
+                        return 255;
+
+                    case Regs::LogicOp::CopyInverted:
+                        return ~src;
+
+                    case Regs::LogicOp::NoOp:
+                        return dest;
+
+                    case Regs::LogicOp::Invert:
+                        return ~dest;
+
+                    case Regs::LogicOp::Nand:
+                        return ~(src & dest);
+
+                    case Regs::LogicOp::Or:
+                        return src | dest;
+
+                    case Regs::LogicOp::Nor:
+                        return ~(src | dest);
+
+                    case Regs::LogicOp::Xor:
+                        return src ^ dest;
+
+                    case Regs::LogicOp::Equiv:
+                        return ~(src ^ dest);
+
+                    case Regs::LogicOp::AndInverted:
+                        return ~src & dest;
+
+                    case Regs::LogicOp::OrReverse:
+                        return src | ~dest;
+
+                    case Regs::LogicOp::OrInverted:
+                        return ~src | dest;
+                    }
+                };
+
+                blend_output = Math::MakeVec(
+                    LogicOp(combiner_output.r(), dest.r(), output_merger.logic_op),
+                    LogicOp(combiner_output.g(), dest.g(), output_merger.logic_op),
+                    LogicOp(combiner_output.b(), dest.b(), output_merger.logic_op),
+                    LogicOp(combiner_output.a(), dest.a(), output_merger.logic_op));
             }
 
             const Math::Vec4<u8> result = {
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -570,6 +570,13 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                     case Operation::Add:
                         return std::min(255, input[0] + input[1]);
 
+                    case Operation::AddSigned:
+                    {
+                        // TODO(bunnei): Verify that the color conversion from (float) 0.5f to (byte) 128 is correct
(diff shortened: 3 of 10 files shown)


src/video_core/vertex_shader.cpp
```diff
@@ -119,17 +119,13 @@ static void ProcessShaderCode(VertexShaderState& state) {
...
-            //       For now, we just abort in this situation.
-            ASSERT_MSG(!is_inverted, "Bad condition...");
+            const bool is_inverted = 0 != (instr.opcode.Value().GetInfo().subtype & OpCode::Info::SrcInversed);
 
             const int address_offset = (instr.common.address_register_index == 0)
                                        ? 0 : state.address_registers[instr.common.address_register_index - 1];
 
-            const float24* src1_ = LookupSourceRegister(instr.common.GetSrc1(is_inverted) + address_offset);
-            const float24* src2_ = LookupSourceRegister(instr.common.GetSrc2(is_inverted));
+            const float24* src1_ = LookupSourceRegister(instr.common.GetSrc1(is_inverted) + (!is_inverted ? address_offset : 0));
```
Subv: Is this really all that was needed to implement the inverted instruction format?

src/video_core/color.h
```diff
@@ -70,6 +70,17 @@ inline const Math::Vec4<u8> DecodeRGB8(const u8* bytes) {
 }
 
 /**
+ * Decode a color stored in RG8 format
+ * @remark RG8 is not intended to store colors, the word color here is used just for consistency,
+ *  this a format intended for normal maps.
+ * @param bytes Pointer to encoded source color
+ * @return Result color decoded as Math::Vec4<u8>
+ */
+inline const Math::Vec4<u8> DecodeRG8(const u8* bytes) {
```
neobrain: Sure that this is RG, rather than GR?

src/video_core/pica.h
```diff
@@ -707,8 +723,27 @@ struct Regs {
...
     } vs_default_attributes_setup;
+
+    INSERT_PADDING_WORDS(0x2);
+
+    struct {
+        BitField< 0, 20, u32> size_0;
+        BitField< 0, 20, u32> size_1;
+        BitField< 0, 28, u32> addr_0;
+        BitField< 0, 28, u32> addr_1;
+        u32 trigger_0;
```
neobrain: Why not make this an array?

neobrain: *these, I guess.

src/video_core/pica.h
```diff
@@ -707,8 +723,27 @@ struct Regs {
...
+        u32 trigger_1;
+
+        int GetSize(int index) {
+            return 8 * ((index == 0) ? size_0 : size_1);
+        }
+
+        PAddr GetPhysicalAddress(int index) {
+            return (PAddr)(8 * ((index == 0) ? addr_0 : addr_1));
+        }
+    } cmd_buffer;
```
neobrain: No abbreviations plz. => `command_buffer`

src/video_core/rasterizer.cpp
```diff
@@ -402,9 +402,11 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
 
                 auto GetSource = [&](Source source) -> Math::Vec4<u8> {
                     switch (source) {
-                    // TODO: What's the difference between these two?
                     case Source::PrimaryColor:
+
+                    // HACK: Until we implement fragment lighting, just use primary_color
                     case Source::PrimaryFragmentColor:
+                    case Source::SecondaryFragmentColor:
```
neobrain: Return 0 instead for this branch. primary_color makes no sense at all.

src/video_core/rasterizer.cpp
```diff
@@ -570,6 +572,10 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                     case Operation::Add:
                         return std::min(255, input[0] + input[1]);
 
+                    case Operation::AddSigned:
+                        // TODO(bunnei): Verify that the color conversion from (float) 0.5f to (byte) 128 is correct
+                        return MathUtil::Clamp<u8>(int(input[0]) + int(input[1]) - 128, 0, 255);
```
neobrain: is the int cast actually necessary there? if it is, use static_cast instead.

Also note that this clamp is buggy: casting the `Clamp` inputs to u8 will over-/underflow in all cases where the Clamp would have any effect.

src/video_core/rasterizer.cpp
```diff
@@ -866,8 +871,69 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                 blend_output     = EvaluateBlendEquation(combiner_output, srcfactor, dest, dstfactor, params.blend_equation_rgb);
                 blend_output.a() = EvaluateBlendEquation(combiner_output, srcfactor, dest, dstfactor, params.blend_equation_a).a();
             } else {
-                LOG_CRITICAL(HW_GPU, "logic op: %x", output_merger.logic_op);
-                UNIMPLEMENTED();
+                static auto LogicOp = [](const Math::Vec4<u8>& src, const Math::Vec4<u8>& dest, Regs::LogicOp op) -> Math::Vec4 < u8 > {
+                    switch (op) {
+                    case Regs::LogicOp::Clear:
```
neobrain: Don't submit untested code. Just implement what you actually found used, and of which you asserted that it works in a more sophisticated way than "I didn't see any glitches on the screen". Alternatively, write an automatable hardware test for it.

src/video_core/rasterizer.cpp
neobrain: Actually, given that this was submitted in a broken way without the author noticing it (suggesting a lack of proper test case), I'd suggest dropping it from the change set until a hardware test is written.

src/video_core/rasterizer.cpp
```diff
@@ -866,8 +871,69 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                    case Regs::LogicOp::AndInverted:
+                        return Math::MakeVec<u8>(~src.r() & dest.r(), ~src.g() & dest.g(),
+                                                 ~src.b() & dest.b(), ~src.a() & dest.a());
+
+                    case Regs::LogicOp::OrReverse:
+                        return Math::MakeVec<u8>(src.r() | ~dest.r(), src.g() | ~dest.g(),
+                                                 src.b() | ~dest.b(), src.a() | ~dest.a());
+
+                    case Regs::LogicOp::OrInverted:
+                        return Math::MakeVec<u8>(~src.r() | dest.r(), ~src.g() | dest.g(),
```
neobrain: What's the result of bitwise operators on `uint8_t`? If it's u8 again, there's no need to explicitly specify the template parameter for MakeVec.

src/video_core/pica.h
```diff
@@ -707,8 +723,27 @@ struct Regs {
...
+        BitField< 0, 28, u32> addr_0;
+        BitField< 0, 28, u32> addr_1;
+        u32 trigger_0;
+        u32 trigger_1;
+
+        int GetSize(int index) {
+            return 8 * ((index == 0) ? size_0 : size_1);
+        }
+
+        PAddr GetPhysicalAddress(int index) {
```
lioncash: const function.

src/video_core/pica.h
```diff
@@ -707,8 +723,27 @@ struct Regs {
...
+
+    struct {
+        BitField< 0, 20, u32> size_0;
+        BitField< 0, 20, u32> size_1;
+        BitField< 0, 28, u32> addr_0;
+        BitField< 0, 28, u32> addr_1;
+        u32 trigger_0;
+        u32 trigger_1;
+
+        int GetSize(int index) {
```
lioncash: const function

src/video_core/vertex_shader.cpp
neobrain: Suggestion: => `!is_inverted * address_offset`

src/video_core/vertex_shader.cpp
```diff
@@ -119,17 +119,13 @@ static void ProcessShaderCode(VertexShaderState& state) {
         switch (instr.opcode.Value().GetInfo().type) {
         case OpCode::Type::Arithmetic:
         {
-            bool is_inverted = 0 != (instr.opcode.Value().GetInfo().subtype & OpCode::Info::SrcInversed);
-            // TODO: We don't really support this properly: For instance, the address register
-            //       offset needs to be applied to SRC2 instead, etc.
-            //       For now, we just abort in this situation.
-            ASSERT_MSG(!is_inverted, "Bad condition...");
+            const bool is_inverted = 0 != (instr.opcode.Value().GetInfo().subtype & OpCode::Info::SrcInversed);
```
neobrain: Please put parentheses around comparisions when their result is being assigned to a variable.

src/video_core/pica.h
```diff
@@ -707,8 +723,27 @@ struct Regs {
         // Writing to these registers sets the "current" default attribute.
         u32 set_value[3];
     } vs_default_attributes_setup;
+
+    INSERT_PADDING_WORDS(0x2);
+
+    struct {
+        BitField< 0, 20, u32> size_0;
```
neobrain: should have documentation wrt units.

src/video_core/command_processor.cpp
```diff
@@ -58,6 +58,16 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
             return;
 
+        case PICA_REG_INDEX(cmd_buffer.trigger_0):
+        case PICA_REG_INDEX(cmd_buffer.trigger_1):
+        {
+            unsigned index = id - PICA_REG_INDEX(cmd_buffer.trigger_0);
+            u32* head_ptr = (u32*)Memory::GetPhysicalPointer(g_state.regs.cmd_buffer.GetPhysicalAddress(index));
+            g_state.cmd_list.head_ptr = g_state.cmd_list.current_ptr = head_ptr;
+            g_state.cmd_list.length = g_state.regs.cmd_buffer.GetSize(index) / sizeof(u32);
+            return;
```
neobrain: Why does this "return" instead of "break" (same for the case above, actually) ?

neobrain: If this has an actual reason, please add a comment about it.

src/video_core/vertex_shader.cpp
neobrain: @Subv yup.

src/video_core/pica.h
neobrain: Also the types involved here should all be unsigned.

src/video_core/pica.h
neobrain: unsigned

SeannyM:
Tested these changes, the intro of Phoenix Wright Dual Destinies is working better now, this effect previously didn't work:
capcom
...but I'm getting a black screen from Ocarina of Time on the OpenGL renderer. It still works on software.

src/video_core/rasterizer.cpp
bunnei: It's tested

src/video_core/color.h
ILOVEPIE: Did you really just copy my #773 pull request? that was closed for lack of having a good test case yet so this shouldn't be merged until we have a test case.

src/video_core/rasterizer.cpp
ILOVEPIE: A good test for this is Animal Crossing.

src/video_core/rasterizer.cpp
ILOVEPIE: result is int

bunnei: It's `int`, so they are indeed necessary

bunnei:
Addressed comments. There was a regression that broke some games (pointed out by @SeannyM) that should be fixed now. I also removed the RG8 texture format decoding; while Zelda: A Link Between Worlds used this format, it wasn't using it in a way that was trivial to verify correctness.

src/video_core/rasterizer.cpp
bunnei: @neobrain I'm going to leave it in. Mistakes happen, that's the point of the PR process. It's fixed and is correct now, with the except of the 0.5f -> u8 conversion, which has a TODO to verify (as it does in other places in the code).

SeannyM:
@bunnei everything working as it should now on my end.

JohnGodgames:
Some of regression:
-White color instead of textures which worked before...


Harvest Moon: A New Beginning
harvest moon


Pokémon Rumble Blast
pokemon

src/video_core/command_processor.cpp
```diff
@@ -363,38 +373,34 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
-                           ((header.parameter_mask & 0x8) ? (0xFFu << 24) : 0u);
+void ProcessCommandList(const u32* list, u32 size) {
+    g_state.cmd_list.head_ptr = g_state.cmd_list.current_ptr = list;
+    g_state.cmd_list.length = size / sizeof(u32);
 
-    WritePicaReg(header.cmd_id, *read_pointer, write_mask);
-    read_pointer += 2;
+    while (g_state.cmd_list.current_ptr < g_state.cmd_list.head_ptr + g_state.cmd_list.length) {
+        const u32* first_command_word = g_state.cmd_list.current_ptr;
+        const CommandHeader header = *(const CommandHeader*)&first_command_word[1];
```
Subv: static_cast please

Subv:
Did you test the logic ops? if so LGTM.

src/video_core/command_processor.cpp
yuriks: I'd prefer this instead: `const CommandHeader header = { first_command_word[1] };`

yuriks: Actually, `CommandHeader` has a constructor, so just `const CommandHeader header = first_command_word[1];` will already work.

src/video_core/command_processor.cpp
```diff
@@ -363,38 +373,34 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
-    // align read pointer to 8 bytes
-    if ((first_command_word - read_pointer) % 2)
-        ++read_pointer;
+        u32 cmd = header.cmd_id;
+        u32 value = *g_state.cmd_list.current_ptr++;
 
-    return read_pointer - first_command_word;
-}
+        for (unsigned i = 0; i < 1 + header.extra_data_length; ++i) {
+            ++g_state.cmd_list.current_ptr;
```
yuriks: `+= 1` instead? I always find standalone pre/post increments a bit off-putting.

src/video_core/command_processor.cpp
```diff
@@ -363,38 +373,34 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+            WritePicaReg(cmd, value, write_mask);
 
-    while (read_pointer < list + list_length) {
-        read_pointer += ExecuteCommandBlock(read_pointer);
+            cmd = header.cmd_id + ((header.group_commands) ? i + 1 : 0);
+            value = *g_state.cmd_list.current_ptr;
+        }
+        
+        // align read pointer to 8 bytes
+        if ((first_command_word - g_state.cmd_list.current_ptr) % 2)
```
yuriks: Shouldn't this use `head_ptr` instead of `first_command_word` to prevent a bogus alignment on a buffer change? (This will never cause problems since both buffers will most likely have 16-byte alignment in the first place, but is still sloppy.)

src/video_core/command_processor.cpp
```diff
@@ -363,38 +373,34 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
 
-void ProcessCommandList(const u32* list, u32 size) {
-    u32* read_pointer = (u32*)list;
-    u32 list_length = size / sizeof(u32);
+            WritePicaReg(cmd, value, write_mask);
 
-    while (read_pointer < list + list_length) {
-        read_pointer += ExecuteCommandBlock(read_pointer);
+            cmd = header.cmd_id + ((header.group_commands) ? i + 1 : 0);
+            value = *g_state.cmd_list.current_ptr;
```
yuriks: This part is pretty confusing, imo. I tried to make it a bit more clearer: https://gist.github.com/yuriks/84a676187ebbe5d6c06e

src/video_core/command_processor.cpp
```diff
@@ -363,38 +373,34 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
 
-    for (unsigned int i = 1; i < 1+header.extra_data_length; ++i) {
-        u32 cmd = header.cmd_id + ((header.group_commands) ? i : 0);
-        WritePicaReg(cmd, *read_pointer, write_mask);
-        ++read_pointer;
-    }
+        const u32 write_mask = ((header.parameter_mask & 0x1) ? (0xFFu << 0) : 0u) |
+            ((header.parameter_mask & 0x2) ? (0xFFu << 8) : 0u) |
+            ((header.parameter_mask & 0x4) ? (0xFFu << 16) : 0u) |
+            ((header.parameter_mask & 0x8) ? (0xFFu << 24) : 0u);
```
yuriks: http://ideone.com/hcFibk *hint* *hint* *nudge* *nudge*

src/video_core/rasterizer.cpp
```diff
@@ -570,6 +575,13 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
                     case Operation::Add:
                         return std::min(255, input[0] + input[1]);
 
+                    case Operation::AddSigned:
+                    {
+                        // TODO(bunnei): Verify that the color conversion from (float) 0.5f to (byte) 128 is correct
+                        auto result = static_cast<int>(input[0]) + static_cast<int>(input[1]) - 128;
+                        return static_cast<u8>(MathUtil::Clamp<int>(result, 0, 255));
```
yuriks: Is this static_cast necessay?

src/video_core/rasterizer.cpp
```diff
@@ -866,8 +877,63 @@ static void ProcessTriangleInternal(const VertexShader::OutputVertex& v0,
...
+                    case Regs::LogicOp::OrInverted:
+                        return ~src | dest;
+                    }
+                };
+
+                blend_output = Math::MakeVec(
+                    LogicOp(combiner_output.r(), dest.r(), output_merger.logic_op),
+                    LogicOp(combiner_output.g(), dest.g(), output_merger.logic_op),
+                    LogicOp(combiner_output.b(), dest.b(), output_merger.logic_op),
+                    LogicOp(combiner_output.a(), dest.a(), output_merger.logic_op));
```
yuriks: You can make `LogicOp` operate on a `u32` and do all of these in one go.

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -99,8 +101,11 @@ void OpenGLState::Apply() {
     if (blend.enabled != cur_state.blend.enabled) {
         if (blend.enabled) {
             glEnable(GL_BLEND);
+            glDisable(GL_COLOR_LOGIC_OP);
+            glLogicOp(GL_COPY);
```
yuriks: This should also set the corresponding flags in the state so that they don't go out of sync.

yuriks:
Pass done.

yuriks:
You might also want to look at @JohnGodgames's regressions and see if they're simply caused by the stubbed implementation of fragment lighting.

src/video_core/command_processor.cpp
neobrain: Sounds like a matter of personal taste. We do standalone increments all over the place, and IMO they're perfectly fine :/

src/video_core/command_processor.cpp
neobrain: There's no reason to have a constructor in CommandHeader. Instead, the constructor should be removed and `const CommandHeader header = { first_command_word[1] };` be used instead.

src/video_core/rasterizer.cpp
neobrain: ... but please don't actually do that. There's little point and it just makes the code less obvious to understand IMO.

src/video_core/command_processor.cpp
yuriks: I looked into this and it turns out that for some reason we have *two* definitions of `CommandHeader` in the codebase, one in `video_core/pica.h` and another in `command_processor.h`. The difference is one is documented and the other has a constructor defined. One of those should be removed.

src/video_core/renderer_opengl/gl_state.h
```diff
@@ -42,6 +42,8 @@ class OpenGLState {
         } color; // GL_BLEND_COLOR
     } blend;
 
+    GLenum logic_op; // GL_COPY
```
tfarley: This comment should be "// GL_LOGIC_OP_MODE" to correspond to its glGet*() like the others (since the GL_COPY default is already specified in constructor).

yuriks:
LGTM

JohnGodgames:
Just tested an update and there is a result:

pros:

Fixed
pica pokemon rb

Fixed
pica harvest moon


cons:

Stiil not perfect
pica sonic

src/video_core/command_processor.cpp
```diff
@@ -363,38 +373,34 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
-        ++read_pointer;
-    }
-
-    // align read pointer to 8 bytes
-    if ((first_command_word - read_pointer) % 2)
-        ++read_pointer;
-
-    return read_pointer - first_command_word;
+/// Expands a 4-bit mask to 4-byte mask, e.g. 0b0101 -> 0x00FF00FF
+static u32 ExpandBitsToBytes(u32 x) {
```
neobrain: Wouldn't this be easier to grasp if we used the old code (i.e. using ternaries) in this function instead of this gibberish? The old code was obviously correct, while this new one is pretty confusing (I couldn't verify its correctness without pencil+paper ...)

src/video_core/command_processor.cpp
```diff
@@ -363,38 +373,34 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
-        read_pointer += ExecuteCommandBlock(read_pointer);
+    g_state.cmd_list.head_ptr = g_state.cmd_list.current_ptr = list;
+    g_state.cmd_list.length = size / sizeof(u32);
+
+    while (g_state.cmd_list.current_ptr < g_state.cmd_list.head_ptr + g_state.cmd_list.length) {
+        // Align read pointer to 8 bytes
+        if ((g_state.cmd_list.head_ptr - g_state.cmd_list.current_ptr) % 2 != 0)
+            ++g_state.cmd_list.current_ptr;
+
+        const CommandHeader header = { g_state.cmd_list.current_ptr[1] };
```
neobrain: This can be reordered and rewritten to:

    u32 value = *g_state.cmd_list.current_ptr++;
    const CommandHeader header = { *g_state.cmd_list.current_ptr++ };

which doesn't require a standalone "g_state.cmd_list.current_ptr += 2;" line and in general is easier to read IMO.

src/video_core/pica.h
```diff
@@ -707,8 +723,31 @@ struct Regs {
         // Writing to these registers sets the "current" default attribute.
         u32 set_value[3];
     } vs_default_attributes_setup;
+
+    INSERT_PADDING_WORDS(0x2);
+
+    struct {
+        // There are two channels that can be used to configure the next command buffer, which
+        // can be then executed (without a CPU interrupt) by writing to the "trigger" registers
```
neobrain: The "regular" method to submit GPU commands via GSP doesn't require CPU interrupts either, or does it? Seems like it's not really made clear what the actual point here is.

src/video_core/pica.h
```diff
@@ -707,8 +723,31 @@ struct Regs {
...
+        // below. An arbitrary number of additional command buffers may be executed in a row. Games
+        // often configure these registers to provide a command list "jump" mechanism.
```
neobrain: "Games often configure these registers to provide a command list "jump" mechanism." ... isn't this exactly the point? This phrasing suggests that it could be configured in a way which doesn't "jump".

src/video_core/vertex_shader.cpp
```diff
@@ -279,6 +284,18 @@ static void ProcessShaderCode(VertexShaderState& state) {
                 break;
             }
 
+            case OpCode::Id::SLT:
+            case OpCode::Id::SLTI:
+            {
```
neobrain: redundant braces

src/video_core/pica.h
bunnei: @neobrain yes a normal command list from GSP always ends with an interrupt

src/video_core/pica.h
bunnei: What I was trying to say is not clear, by "jump" I mean "subroutine" (as in, jump to a command list, then return). I'll fix the comment

src/video_core/pica.h
neobrain: @bunnei that doesn't have anything to do with launching other command lists, though...

src/video_core/pica.h
neobrain: yeah, subroutine is a lot clearer indeed.

src/video_core/command_processor.cpp
yuriks: You can easily verify it's correctness just by executing the function for all the 16 possible inputs. (Which is what I did.)

The function is documented on what it does, and this code is never going to change, so the ability for it to be easily modified isn't important.

neobrain: easy != obvious.

I'm not seeing how this voodoo (no offense) is better than plain obvious code.

hrydgard: This one is probably faster to do with a lookup-table anyway actually, as it's used quite a bit so the tiny 16-entry table will sit in L1 or L0

src/video_core/command_processor.cpp
hrydgard: Either way you really should move current_ptr into a local variable before the start of the loop, and put it back afterwards. Walking pointers stored in global variables is actually really slow.

yuriks: current_ptr is modified inside WritePicaReg, unfortunately, so it has to be read at least once per loop iteration.

src/video_core/command_processor.cpp
yuriks: Good point about the table.

neobrain: @hrydgard I don't think performance considerations matter here (since this is far from being a hotspot), but I guess a LUT-based approach would be both performant and readable ;)

src/video_core/pica.h
bunnei: I still think it's relevant information, because that's why these registers are used :)

src/video_core/pica.h
```diff
@@ -707,8 +723,34 @@ struct Regs {
...
         u32 set_value[3];
     } vs_default_attributes_setup;
+
+    INSERT_PADDING_WORDS(0x2);
+
+    struct {
+        // There are two channels that can be used to configure the next command buffer, which
+        // can be then executed by writing to the "trigger" registers. There are two reasons why a
+        // game might use this feature:
+        //  1) With this, an arbitrary number of additional command buffers may be executed without
```
neobrain: suggested alternative: "An arbitrary number of additional command buffers may be queued without explicitly having the CPI wait for completion of the current command buffer execution"

yuriks: Meh, I think the current wording for the first part is more accurate, though it meanders at the end.

How about this?
> With this, an arbitrary number of additional command buffers may be executed in sequence without requiring intervention of the CPU after the initial one is kicked off.

PR #823, opened 2015-05-26
Library applet support (swkbd for now)

### LINKED ISSUE #718: Hanging when on-screen keyboard should appear
Any ideas how to fix?
Games that it probably also needs on-screen keyboard support:
- Acrylic Palette: Irodori Cafe - Cheers
- Animal Hospital
- A-Train 3D
- Beyblade Evolution (log - http://pastie.org/10100354)
- Dead or Alive: Dimensions (log - http://pastie.org/10100510)
- ESSE Rakuraku Kakeibo
- Gundam Try Age SP
- My Pet Puppy 3D
- Paws & Claws: Pampered Pets Resort 3D
- Pokémon Art Academy
- Theatrhythm Final Fantasy  (log - http://pastie.org/10126152)
- Zoo Resort 3D
- Yumi's Odd Odyssey (Sayonara Umihara Kawase)

and propably more

CITRA HANGS HERE (Theatrhythm Final Fantasy):
citra keyboard 1

Function on-screen keyboard:
citra keyboard 2

Subv:
These changes add a new interface for implementing library applets, and adds an implementation of it for the SoftwareKeyboard (swkbd).

An applet's lifetime goes like this
* PrepareToLaunchLibraryApplet creates the applet.
* Application sends signal 2 to the applet, and the applet responds with signal 3.
* StartLibraryApplet signals the applet to start doing its thing (Drawing to the screen, etc).
* Once the applet finishes, it sends signal 0xA to the application.
* Application now restores the framebuffers and continues execution.

In this HLE implementation, the drawing and handling of touch events / keys should be done in the Applet::Update method.
The implementation of the SoftwareKeyboard currently only returnrs "Citra" as the text, without asking the user for input, but this can change in the future by means of just drawing the keyboard to the framebuffer inside SoftwareKeyboard::DrawScreenKeyboard, and handling the input events + updating the framebuffer in SoftwareKeyboard::Update. I didn't do it because I'm not good at drawing stuff.

Furthermore, this PR should fix #718


ILOVEPIE:
wouldn't it be simple to just pop up a window asking the user to type in text?

Subv:
We have no way of doing that in a frontend-agnostic way, and it doesn't scale to the other Applets that we'll have to support in the future

src/core/hle/service/apt/apt.cpp
```diff
@@ -143,50 +158,81 @@ void InquireNotification(Service::Interface* self) {
...
-    u32 dst_app_id          = cmd_buff[2];
-    u32 signal_type         = cmd_buff[3];
-    u32 buffer_size         = cmd_buff[4];
-    u32 value               = cmd_buff[5];
-    u32 handle              = cmd_buff[6];
-    u32 size                = cmd_buff[7];
-    u32 in_param_buffer_ptr = cmd_buff[8];
-    
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    u32 src_app_id = cmd_buff[1];
```
purpasmart96: Why remove the indenting?

src/core/hle/service/apt/apt.cpp
```diff
@@ -279,6 +325,28 @@ void GetAppCpuTimeLimit(Service::Interface* self) {
     LOG_WARNING(Service_APT, "(STUBBED) called value=%u", value);
 }
 
+void PrepareToStartLibraryApplet(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    cmd_buff[1] = HLE::Applets::Applet::Create(static_cast<AppletId>(cmd_buff[1])).raw;
```
purpasmart96: Is this a result value? Since cmd_buff[1] is usually zero or the error code. Unless this is tested.

src/core/hle/service/apt/apt.cpp
Subv: Side effect from Visual Studio's automatic formatting, i'll change it back.

src/core/hle/service/apt/apt.cpp
Subv: HLE::Applets::Applet::Create returns a ResultCode

src/core/hle/service/apt/apt.cpp
purpasmart96: That's odd, Visual studio really doesn't do that to me, And if it does you can ctrl-z and Visual studio will undo it's automatic formatting.

src/core/hle/service/apt/apt.cpp
purpasmart96: Ok, just making sure.

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -203,7 +203,7 @@ static void ReadHWRegs(Service::Interface* self) {
     }
 }
 
-static void SetBufferSwap(u32 screen_id, const FrameBufferInfo& info) {
```
lioncash: Why are the static and inline keywords commented out?

yuriks: He's using them from another file (with manual prototypes) but I think they should be just removed instead of commented out then.

JohnGodgames:
Problematic games debug logs:

Harvest moon 3d a new beginning - http://pasted.co/0a18d5ce
Zoo Resort 3D (debug log) - http://pasted.co/4c13e999

src/core/hle/service/gsp_gpu.cpp
Subv: I need to access these from another file, I'll probably just end up moving their declarations to the header

purpasmart96:
@JohnGodgames Not related to the software keyboard though
"unimplemented SVC function UnmapMemoryBlock(..)"

src/core/hle/service/apt/apt.cpp
```diff
@@ -47,18 +57,18 @@ void Initialize(Service::Interface* self) {
...
-    start_event->Clear();
+    parameter_event->Clear();
 
     ASSERT_MSG((nullptr != lock), "Cannot initialize without lock");
     lock->Release();
 
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
 
-    LOG_TRACE(Service_APT, "called app_id=0x%08X, flags=0x%08X", app_id, flags);
+    LOG_WARNING(Service_APT, "called app_id=0x%08X, flags=0x%08X", app_id, flags);
```
purpasmart96: A debug log instead of a warning log makes more sense.

Subv:
@purpasmart96 That's normal, and has nothing to do with the issue

src/core/hle/applets/applet.cpp
```diff
@@ -0,0 +1,77 @@
...
+        }
+    };
+}
+
+namespace HLE {
+namespace Applets {
+
+static std::unordered_map<Service::APT::AppletId, std::shared_ptr<Applet>> applets;
+static u32 applet_update_event = -1; ///< The CoreTiming event identifier for the Applet update callback.
+static const u64 applet_update_interval = 1000000; ///< The interval at which the Applet update callback will be called.
```
yuriks: Add a suffix to the variable to make clear that this is in `us`.

yuriks: Also, an update rate of once a second? This is too slow for any sort of realtime interaction.

src/core/hle/applets/swkbd.cpp
```diff
@@ -0,0 +1,106 @@
...
+    message.buffer_size = sizeof(SoftwareKeyboardConfig);
+    message.data = reinterpret_cast<u8*>(&config);
+    message.signal = static_cast<u32>(Service::APT::SignalType::LibAppClosed);
+    message.destination_id = static_cast<u32>(Service::APT::AppletId::Application);
+    message.sender_id = static_cast<u32>(id);
+    Service::APT::SendParameter(message);
+
+    // Unset the current applet, we are not running anymore
+    g_current_applet = nullptr;
+    started = false;
```
neobrain: Won't setting g_current_applet potentially access invalid memory, since setting `g_current_applet` might deallocate `this` ?

Subv: There's always at least one instance of the Applet alive at any time after it's first created by PrepareToStartLibraryApplet, the one in the "applets" map.

neobrain: Is there any particular reason to not use the "safe" way of switching around the order of these two statements? Seems more logical to me to change internal applet state and then release it, instead of releasing but then still accessing it.

Subv: No particular reason, I'll change it in a moment

src/core/hle/applets/applet.cpp
```diff
@@ -0,0 +1,77 @@
...
+    struct hash<Service::APT::AppletId> {
+        typedef Service::APT::AppletId argument_type;
+        typedef std::size_t result_type;
+
+        result_type operator()(const argument_type& id_code) const {
+            typedef std::underlying_type<argument_type>::type Type;
+            return std::hash<Type>()(static_cast<Type>(id_code));
+        }
+    };
+}
```
yuriks: We might want to move this workaround to base `EnumHashWorkaround` class in common and then just do private inheritance from it to invoke the workaround.

src/core/hle/applets/applet.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace HLE {
+namespace Applets {
+
+static std::unordered_map<Service::APT::AppletId, std::shared_ptr<Applet>> applets;
+static u32 applet_update_event = -1; ///< The CoreTiming event identifier for the Applet update callback.
+static const u64 applet_update_interval = 1000000; ///< The interval at which the Applet update callback will be called.
+std::shared_ptr<Applet> g_current_applet = nullptr; ///< The applet that is currently executing
```
yuriks: Why do both this and the `applets` map exist? Can more than one applet run at once or can they not? (In the real system they can not, afaik.)

src/core/hle/applets/applet.cpp
```diff
@@ -0,0 +1,77 @@
...
+    if (g_current_applet && g_current_applet->IsRunning())
+        g_current_applet->Update();
+
+    CoreTiming::ScheduleEvent(usToCycles(applet_update_interval) - cycles_late,
+        applet_update_event);
+}
+
+void Init() {
+    applet_update_event = CoreTiming::RegisterEvent("HLE Applet Update Event", AppletUpdateEvent);
+    CoreTiming::ScheduleEvent(usToCycles(applet_update_interval), applet_update_event);
```
yuriks: The timing event should only be active while the applet is running.

src/core/hle/applets/applet.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include "common/common_types.h"
+#include "core/hle/kernel/kernel.h"
+#include "core/hle/kernel/shared_memory.h"
+#include "core/hle/service/apt/apt.h"
+
+namespace HLE {
+namespace Applets {
+
+class Applet : public std::enable_shared_from_this<Applet> {
```
yuriks: Giant red flag. This shouldn't be necessary.

src/core/hle/applets/applet.cpp
Subv: Depends, an application can start a system applet, which can start a library applet. I'll remove the g_current_applet variable and go with another approach

src/core/hle/applets/applet.cpp
```diff
@@ -0,0 +1,78 @@
...
+    };
+}
+
+namespace HLE {
+namespace Applets {
+
+static std::unordered_map<Service::APT::AppletId, std::shared_ptr<Applet>> applets;
+static u32 applet_update_event = -1; ///< The CoreTiming event identifier for the Applet update callback.
+/// The interval at which the Applet update callback will be called.
+static const u64 applet_update_interval_microseconds = 16666;
```
yuriks: I think `us` suffices as the suffix... also add a comment saying "16.6ms".

src/core/hle/applets/swkbd.cpp
```diff
@@ -0,0 +1,106 @@
...
+    text_memory = boost::static_pointer_cast<Kernel::SharedMemory, Kernel::Object>(parameter.object);
+
+    // TODO(Subv): Verify if this is the correct behavior
+    memset(text_memory->GetPointer(), 0, text_memory->size);
+
+    DrawScreenKeyboard();
+
+    // Update the current applet so we can get update events
+    started = true;
+    g_current_applet = shared_from_this();
```
yuriks: Shouldn't happen here. See other comment below.

src/core/hle/applets/swkbd.cpp
```diff
@@ -0,0 +1,106 @@
...
+    message.buffer_size = sizeof(SoftwareKeyboardConfig);
+    message.data = reinterpret_cast<u8*>(&config);
+    message.signal = static_cast<u32>(Service::APT::SignalType::LibAppClosed);
+    message.destination_id = static_cast<u32>(Service::APT::AppletId::Application);
+    message.sender_id = static_cast<u32>(id);
+    Service::APT::SendParameter(message);
+
+    started = false;
+    // Unset the current applet, we are not running anymore
+    g_current_applet = nullptr;
```
yuriks: This shouldn't be being done here. Instead, the variable should be set when the applet is launching, and it should be unset by the update callback, which should check if the applet has finished already and the "reap" it. This way you avoid the memory management issues.

yuriks:
Haven't looked at all the actual applety stuff, but the memory management stuff here is pretty sub-par. Should be fixable though.

Subv:
Updated

Subv:
Anybody feels like writing a nice keyboard UI for this?

Subv:
I've removed the RFC tag from this PR, I think it's complete-enough for merging. The actual drawing of the keyboard can be achieved in a future PR.

src/core/hle/applets/applet.h
```diff
@@ -0,0 +1,77 @@
...
+#include "core/hle/kernel/kernel.h"
+#include "core/hle/kernel/shared_memory.h"
+#include "core/hle/service/apt/apt.h"
+
+namespace HLE {
+namespace Applets {
+
+class Applet {
+public:
+    virtual ~Applet() {};
```
lioncash: Semicolon isn't necessary here.

src/core/hle/applets/swkbd.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    started = true;
+    return RESULT_SUCCESS;
+}
+
+void SoftwareKeyboard::Update() {
+    // TODO(Subv): Handle input using the touch events from the HID module
+
+    // TODO(Subv): Remove this hardcoded text
+    std::u16string text = Common::UTF8ToUTF16("Citra");
```
archshift: `u"Citra"`?

Subv: VS2013 doesn't support that

mailwl: L"zyxw";

Subv: That creates a wchar_t* and is not necessarily an UTF-16 string

src/core/hle/applets/applet.cpp
bunnei: Ditto... I actually had something else recently that I could use this on as well

src/core/hle/applets/applet.h
bunnei: It also bothers me that the destructor has a space but this does not :)

src/core/hle/applets/swkbd.cpp
```diff
@@ -0,0 +1,102 @@
...
+    result.destination_id = static_cast<u32>(Service::APT::AppletId::Application);
+    result.sender_id = static_cast<u32>(id);
+    result.object = framebuffer_memory;
+
+    Service::APT::SendParameter(result);
+    return RESULT_SUCCESS;
+}
+
+ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter const& parameter) {
+    memcpy(&config, parameter.data, parameter.buffer_size);
```
neobrain: Is there any chance that buffer_size is specified by the emulated application, or will be specified by the emulated application in the future? If so, there should be a safety check here to make sure we perform no out-of-bounds memcpy.

src/core/hle/applets/swkbd.h
```diff
@@ -0,0 +1,83 @@
...
+    INSERT_PADDING_WORDS(0x1);
+
+    u32 return_code; ///< Return code of the SoftwareKeyboard, usually 2, other values are unknown
+
+    INSERT_PADDING_WORDS(0x2);
+
+    u32 text_offset; ///< Offset in the SharedMemory where the output text starts
+    u16 text_length; ///< Length in characters of the output text
+
+    INSERT_PADDING_BYTES(0x2B6);
```
neobrain: Where does this size come from?

Subv: From inspecting the disassembled binary of a game that uses swkbd. This size (0x400) is also passed as the 'buffer_size' parameter to StartLibraryApplet

neobrain: Could you document this in the code?

Subv: I don't see the point, there's a static_assert right below this code that implies that this structure's size is correct.

neobrain: The code doesn't mention anywhere where that size is taken from. For all one knows when reading the source, it could just be a blind guess or a dumb "just to be be safe and have this struct large enough" hack.

Subv:
Addressed the comments by neobrain

bunnei:
LgTm

PR #827, opened 2015-05-27
QT: Keep the render window invisible until emulation starts

### LINKED ISSUE #174: Qt: Rendering area never cleared by GL while games aren't actively rendering.
example

SeannyM:
Fixes #174. Calling ToggleWindowMode makes the render window visible, and since it's called before emulation starts it sets the window to visible before it should be.


src/citra_qt/main.cpp
```diff
@@ -316,7 +316,8 @@ void GMainWindow::ToggleWindowMode() {
         // Render in the main window...
         render_window->BackupGeometry();
         ui.horizontalLayout->addWidget(render_window);
-        render_window->setVisible(true);
+        if (emu_thread.get())
```
archshift: This should be `if (emu_thread != nullptr)`.

yuriks:
Tested and seems to work ok, but unfortunately this removes the minimum size restriction on the window until you start emulation. Is there an easy way to fix this?

SeannyM:
@yuriks sorry I didn't have a chance to reply sooner. FYI, master has the same issue; if you start then stop emulation it hides the render window and the size restriction is lifted. AFAIK there is no simple way to make the program account for hidden widget space. If I have time I can look into the best way of doing this.

SeannyM:
I added a placeholder QSpacerItem to account for the render window's space when it's hidden. This is probably the easiest way to account for hidden widget space, at least that I know of.

src/citra_qt/main.cpp
```diff
@@ -57,6 +57,8 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     ui.setupUi(this);
     statusBar()->hide();
 
+    render_placeholder = new QSpacerItem(400, 480, QSizePolicy::Fixed, QSizePolicy::Fixed);
```
yuriks: This will leak if the window is destroyed while the spacer isn't inserted into the layout, right? Can you add something to ensure it is destroyed when the windows is closed. (Also make sure not to do a double-free in the cases where it's already destroyed by Qt.)

yuriks:
I hadn't actually realized we were already hidding the widget when you clicked stop. :) Your approach seems to work well. LGTM apart from the minor memory management issue.

SeannyM:
@yuriks destroying the placeholder crashes Citra when closed. It turns out QT should already destroying it. See: http://doc.qt.io/qt-5/objecttrees.html

yuriks:
Yeah, I just realized that since emulation is stopped before closing the window, that will automatically insert the placeholder, causing it to always be destroyed by Qt as part of the layout.

src/citra_qt/main.h
```diff
@@ -75,6 +75,7 @@ private slots:
 private:
     Ui::MainWindow ui;
 
+    QSpacerItem* render_placeholder;
```
neobrain: Please rename to "render_window_placeholder".

neobrain:
Isn't there a better way to solve this without the issues @yuriks mentioned? (i.e. ownership management on the placeholder widget and minimum render window size)

Maybe you could implement something in the render code widget itself?

SeannyM:
@neobrain The placeholder was put in to solve the minimum window size issue. There should be no ownership issues, since the placeholder is added to the layout when emulation is stopped (right before the window closes), the layout should have ownership of the placeholder whenever the window is closed, so it will be destroyed. It would be possible to implement something in the render widget itself, and I did try a few things (which didn't work as desired) but I settled on using a placeholder as it seems to be the easiest solution. The way the render widget works complicates things when it comes to fully hiding it without removing it from the layout.

yuriks:
Anything holding this up? Otherwise I think it can be merged.

neobrain:
Sorry, this approach is ugly, and the area clearing should be done in the emu window itself.

yuriks:
@neobrain I disagree. We're going to have to do some kind of widget switching anyway when we add a background splash or something of the sort later on. Also, there is no clean way to match the system's toolkit background color if we're clearing the window ourselves (not to mention the context ownership mess you have to deal with if you go that way of drawing on it while emulation isn't running).

@SeannyM: Maybe to appease neobrain, keep track of if the spacer is currently inserted into the layout, and then do an explicit check to see if you need to delete it during the destruction of the window.

PR #828, opened 2015-05-27
hid: Get rid of undefined behavior

lioncash:
Modifying a variable twice across a sequence point.


yuriks:
LGTM (oh my god who wrote this)

PR #830, opened 2015-05-28
QT: Remove border around widgets

### LINKED ISSUE #815: QT: Remove gray border from rendering canvas
Currently there's a ~10px gray border between the rendered 3DS screen and window border, this is not only a waste of screen space but also doesn't fit the style of the modern operating systems.

(Left: Now. Right: How it should be.)

SeannyM:
Fixes #815.
Without any extra widgets, it looks like this:
nowidgets

And with extra widgets, there is still visible space between different widgets because of the empty space that's part of them:
widgets

### DIFF
--- a/src/citra_qt/main.ui
+++ b/src/citra_qt/main.ui
@@ -24,7 +24,20 @@
    <bool>true</bool>
   </property>
   <widget class="QWidget" name="centralwidget">
-   <layout class="QHBoxLayout" name="horizontalLayout"/>
+   <layout class="QHBoxLayout" name="horizontalLayout">
+    <property name="leftMargin">
+     <number>0</number>
+    </property>
+    <property name="topMargin">
+     <number>0</number>
+    </property>
+    <property name="rightMargin">
+     <number>0</number>
+    </property>
+    <property name="bottomMargin">
+     <number>0</number>
+    </property>
+   </layout>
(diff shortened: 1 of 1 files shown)


archshift:
How does this look with the widget header bars enabled?

Other than seeing that, LGTM.

SeannyM:
@archshift With widget headers:
headers

neobrain:
Couldn't this be adjusted in the .ui file itself?

filfat:
LATM (Looks awesome to me)

neobrain:
Also note that setMargin is a deprecated member: http://doc.qt.io/qt-4.8/qlayout-obsolete.html

SeannyM:
@neobrain I didn't see anything about margins in the QT Creator, though I could've missed something. I switched to setContentsMargins() which isn't decrypted.

neobrain:
@SeannyM selecting "centralWidget" in the designer should give you a number of Layout options. Are these equivalent to this change?

SeannyM:
@neobrain Ah, I found it. I've never worked with QT before, so I don't fully know my way around yet. Also, it looks like the QT Editor auto-fixed indentation on another line.

yuriks:
LGTM

bunnei:
LGTM

PR #832, opened 2015-05-29
Remove gpu_refresh_rate configuration option

yuriks:
Changing it makes emulation inherently inaccurate. It also had a wrong
default value (30, whereas the real system has a refresh rate of 60 Hz)
which, even if changed, would continue to be used unless people manually
removed it from their config files.


purpasmart96:
I'm not sure about this, some games with this no longer quite work.

yuriks:
@purpasmart96 Then we can fix this. As is it's completely wrong.

archshift:
LGTM

Subv:
LGTM

bunnei:
Lgtm

Issue #829, opened 2015-05-28
Hardware renderer bugs - Black screens & similar issues


JohnGodgames:
It would be good that everyone knows about these problems.If I or someone else as @JMC47 will find more bugs there will be a place for them.

Update:
#### Games fixed by https://github.com/citra-emu/citra/pull/835
CRUSH 3D
Farming Simulator 14
Labyrinth no Kanata
Castle Clout 3D
Robot Rescue
Tales of the World - Reve Unitia

#### Games fixed by https://github.com/citra-emu/citra/pull/1023
Planet Crashers

# Games not fixed yet:

#### Monster 4x4 3D:
SW
![monster 4x4 sw](https://cloud.githubusercontent.com/assets/11453507/7871567/009f3f12-0591-11e5-8e5c-5d65aec7057d.PNG)

HW
![monster 4x4 hw](https://cloud.githubusercontent.com/assets/11453507/7871568/044fc41a-0591-11e5-8053-aafe14295071.PNG)

#### Rise of the Guardians:
SW
![rise of the guardians sw](https://cloud.githubusercontent.com/assets/11453507/7871602/38f3ff92-0591-11e5-9afc-76ac0179b00d.PNG)

HW
![rise of the guardians hw](https://cloud.githubusercontent.com/assets/11453507/7871604/3ca1aa18-0591-11e5-9b9e-f25ed082b63d.PNG)

#### Turbo - Super Stunt Squad:
SW
![turbo sw](https://cloud.githubusercontent.com/assets/11453507/7871615/4e49dfa6-0591-11e5-8aef-ce2e670edf54.PNG)

HW
![barbie hw](https://cloud.githubusercontent.com/assets/11453507/7871673/b527b716-0591-11e5-924b-79c9d8041e46.PNG)

#### Madagascar 3:
SW
![madagascar 3 sw](https://cloud.githubusercontent.com/assets/11453507/7900607/91656b48-0762-11e5-8c5e-b3257fbd6946.PNG)

HW
![madagascar 3 hw](https://cloud.githubusercontent.com/assets/11453507/7900608/97678bf2-0762-11e5-9f65-476cd6e32c50.PNG)

RHY3756547:
I dunno, I think I prefer the barbie game on HW.

JMC47:
Can confirm more games.  

New Super Mario Bros. 2 - No graphics on HW Renderer
Donkey Kong Country Returns 3D - No Graphics  on HW Renderer
[...]

Issue #814, opened 2015-05-24
Hardware Renderer renders nothing in Donkey Kong Country Returns 3D


JMC47:
Simple enough: try to start up Donkey Kong Country Returns 3D in the Hardware Renderer or switch to it from SW Renderer and you will get a black screen.  Software renderer definitely renders stuff.  Proof below.

![dkcr3d](https://cloud.githubusercontent.com/assets/6598209/7789531/eba41950-0231-11e5-81fd-d7ebfbdea56b.png)

neobrain:
Why was this closed?

Syphurith:
835 is said to cover this issue and 829. I think that's enough expect someone "fixed in XXX".

Subv:
This was closed by #835

PR #835, opened 2015-05-30
HW Renderer Screen Fixes

tfarley:
A batch of smaller fixes to the hardware renderer.
Addresses issue #814 and some bugs in issue #829
* blargSnes rom rendering fixed
  * glColorMask implemented
* Fewer black screens
  * Depth format handling fixes
  * Textures unbound after done using them
  * Render-to-texture flush
  * Interval intersection math fix

### DIFF
--- a/src/common/math_util.h
+++ b/src/common/math_util.h
@@ -12,7 +12,7 @@ namespace MathUtil
 {
 
 inline bool IntervalsIntersect(unsigned start0, unsigned length0, unsigned start1, unsigned length1) {
-    return (std::max(start0, start1) <= std::min(start0 + length0, start1 + length1));
+    return (std::max(start0, start1) < std::min(start0 + length0, start1 + length1));
 }
 
 template<typename T>
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -491,7 +491,7 @@ void RasterizerOpenGL::ReconfigureDepthTexture(DepthTextureInfo& texture, Pica::
     case Pica::Regs::DepthFormat::D24:
         internal_format = GL_DEPTH_COMPONENT24;
         texture.gl_format = GL_DEPTH_COMPONENT;
-        texture.gl_type = GL_UNSIGNED_INT_24_8;
+        texture.gl_type = GL_UNSIGNED_INT;
         break;
 
     case Pica::Regs::DepthFormat::D24S8:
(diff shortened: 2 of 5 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -484,7 +511,7 @@ void RasterizerOpenGL::ReconfigureDepthTexture(DepthTextureInfo& texture, Pica::
     case Pica::Regs::DepthFormat::D24:
         internal_format = GL_DEPTH_COMPONENT24;
         texture.gl_format = GL_DEPTH_COMPONENT;
-        texture.gl_type = GL_UNSIGNED_INT_24_8;
+        texture.gl_type = GL_UNSIGNED_INT;
```
yuriks: What's with this change?

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -75,6 +80,14 @@ void OpenGLState::Apply() {
         glDepthMask(depth.write_mask);
     }
 
+    // Color mask
+    if (color_mask.red_enabled != cur_state.color_mask.red_enabled ||
+        color_mask.green_enabled != cur_state.color_mask.green_enabled ||
+        color_mask.blue_enabled != cur_state.color_mask.blue_enabled ||
+        color_mask.alpha_enabled != cur_state.color_mask.alpha_enabled) {
+        glColorMask(color_mask.red_enabled, color_mask.green_enabled, color_mask.blue_enabled, color_mask.alpha_enabled);
```
yuriks: Can you do something so that the if condition's indentation doesn't sit flush with the contents of the code block?

src/video_core/renderer_opengl/gl_rasterizer.cpp
tfarley: GL_UNSIGNED_INT_24_8 seems to be finicky on some GPU/driver combos and didn't really reflect the usage here any better than just a plain 32-bit type. This is basically the same behavior but treats the uploaded values as full 32-bit before being chopped down to 24 internally instead of uploading 24+8 bit values smashed together.

MoochMcGee: I don't think this is accurate.

src/common/math_util.h
```diff
@@ -12,7 +12,7 @@ namespace MathUtil
 {
 
 inline bool IntervalsIntersect(unsigned start0, unsigned length0, unsigned start1, unsigned length1) {
-    return (std::max(start0, start1) <= std::min(start0 + length0, start1 + length1));
+    return (std::max(start0, start1) <= std::min(start0 + length0 - 1, start1 + length1 - 1));
```
neobrain: Uhm, why not just change to `<`?

tfarley:
Addressed all comments.

bunnei:
Looks mostly fine... please squash

PR #836, opened 2015-05-30
[WIP] InputCommon

shonumi:
The purpose of InputCommon is to centralize input handling from a single interface. It also abstracts the process to allow different backends (SDL, DInput, XInput) to write their own implementations. Currently only SDL has an implementation written here, but others can be added.

Please keep in mind that this represents basic work done towards getting InputCommon towards where it ideally needs to be. Emu_window still handles keyboard input for GLFW and Qt (this is something InputCommon should do), and there is no configuration just yet. For those brave enough to test controller support, the default is set to work with an XBOX 360 controller (Pretty much 1:1 A on 360 = A on the 3DS and so on).


src/input_common/sdl_input/sdl_input.h
```diff
@@ -0,0 +1,59 @@
...
+    bool Init();
+
+    ///Grabs a list of devices supported by the backend
+    void DiscoverDevices();
+
+    ///Grab and process input via SDL2
+    void Poll();
+
+    ///Shuts down all joysticks opened by SDL2
+    void ShutDown();
```
lioncash: Init, DiscoverDevice, Poll, and ShutDown should have an override specifier on them.

src/input_common/sdl_input/sdl_input.h
```diff
@@ -0,0 +1,59 @@
...
+
+#pragma once
+
+#include <SDL2/SDL.h>
+#include <iostream>
+#include "input_common/input_common.h"
+
+namespace InputCommon {
+
+class SDLController  : virtual public ControllerBase {
```
lioncash: `class SDLController final : public ControllerBase`

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+    //Attempt to close joystick with SDL
+    if(jpad != nullptr) {
+        SDL_JoystickClose(jpad);
+    }
+}
+
+void SDLController::Poll() {
+    if(SDL_PollEvent(&input_event)) {
+        if((input_event.type == SDL_JOYBUTTONDOWN) || (input_event.type == SDL_JOYBUTTONUP)
+        || (input_event.type == SDL_JOYAXISMOTION) || (input_event.type == SDL_JOYHATMOTION)) {
```
lioncash: the `==` conditions don't need to be parenthesized.

src/input_common/input_common.cpp
```diff
@@ -0,0 +1,81 @@
...
+        case NO_JOY:
+            break;
+
+        //What?
+        default:
+            break;
+    }
+
+    //Shutdown immediately if backend failed to initialize
+    if((g_user_input != nullptr) && (!g_user_input->activated)) {
```
lioncash: The conditions here don't need to be internally parenthesized.

src/input_common/input_common.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    //Setup CoreTiming
+    frame_ticks = 268123480 / Settings::values.gpu_refresh_rate / 16;
+
+    input_event = CoreTiming::RegisterEvent("InputCommon::InputCallback", InputCallback);
+    CoreTiming::ScheduleEvent(frame_ticks, input_event);
+}
+
+void ShutDown() {
+    if((g_user_input != nullptr)
```
lioncash: extra parenthesis?

lioncash:
`if` and `switch` should also have a space after them.

src/input_common/input_common.cpp
shonumi: Whoops :p

src/input_common/input_common.h
```diff
@@ -0,0 +1,62 @@
...
+    std::string device_name;
+
+    ///Internal view this specific object will have of 3DS input states
+    ControllerState controller;
+};
+
+///Initialize the user input system
+void Init(ControllerBackends backend);
+
+///Deactive the user input system
```
purpasmart96: Space after the "///"

shonumi:
Sorry about the stylistic mess, I'm still stuck on other styles I've used for other projects :p 

Let me know about anything else that catches your eyes. Thanks!

src/input_common/sdl_input/sdl_input.h
```diff
@@ -0,0 +1,57 @@
...
+    void ProcessInput();
+
+    /// Sets up keymaps from input configuration
+    void SetupKeyMaps();
+
+    /// Sets analog deadzones
+    void SetDeadZone(int range);
+
+    /// Checks if analog input is within the dead zone
+    bool CheckDeadZone(int range);
```
lioncash: CheckDeadZone can be const.

src/input_common/input_common.h
```diff
@@ -0,0 +1,60 @@
...
+    virtual void Poll() = 0;
+
+    /// Shuts down all backend related data
+    virtual void Shutdown() = 0;
+
+    /// Returns internal name of currently selected device. Expose this to the UI
+    std::string GetDeviceName() const;
+
+    /// Returns internal pad state
+    const Service::HID::PadState GetPadState() const;
```
lioncash: The first const isn't necessary, since it isn't returning a pointer.

src/core/hle/service/hid/hid.cpp
```diff
@@ -49,7 +51,7 @@ static u32 next_touch_index;
 
 void Update() {
     SharedMem* mem = reinterpret_cast<SharedMem*>(shared_mem->GetPointer());
-    const PadState state = VideoCore::g_emu_window->GetPadState();
+    const PadState state = (InputCommon::g_user_input == nullptr) ? VideoCore::g_emu_window->GetPadState() : InputCommon::g_user_input->GetPadState();
```
bunnei: Line width should be < 100 characters

src/core/system.cpp
```diff
@@ -22,9 +24,11 @@ void Init(EmuWindow* emu_window) {
     Kernel::Init();
     HLE::Init();
     VideoCore::Init(emu_window);
+    InputCommon::Init(InputCommon::SDL2_JOY);
```
bunnei: Input backend should be a configuration setting

src/input_common/controller_state.h
```diff
@@ -0,0 +1,20 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+#include "core/hle/service/hid/hid.h"
+
+namespace InputCommon {
+
+    struct ControllerState {
```
bunnei: Don't indent things within a namespace

src/input_common/input_common.cpp
```diff
@@ -0,0 +1,85 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "input_common/input_common.h"
+
+#ifdef HAS_SDL
```
bunnei: It's nice that you made SDL optional, but having multiple dependency paths (long term) will probably be a pain. I think I'd prefer that we nix GLFW (what we use for the CLI window) and replace with SDL2, and make it a mandatory dependency.


Perhaps OK as an incremen

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+void SDLController::Poll() {
+    if (SDL_PollEvent(&input_event)) {
+        if (input_event.type == SDL_JOYBUTTONDOWN || input_event.type == SDL_JOYBUTTONUP
+        || input_event.type == SDL_JOYAXISMOTION || input_event.type == SDL_JOYHATMOTION) {
+            ProcessInput();
+        }
+    }
+}
+
+void SDLController::ProcessInput() {
```
linkmauve: Instead of defining input_event as an object variable, pass it as argument here.

src/input_common/input_common.cpp
```diff
@@ -0,0 +1,85 @@
...
+    #include "input_common/sdl_input/sdl_input.h"
+#endif
+
+#include "core/core_timing.h"
+#include "core/settings.h"
+
+namespace InputCommon {
+
+//Set to nullptr until initialized by a backend
```
bunnei: Space between the `//` and the comment

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+void SDLController::Shutdown() {
+    //Attempt to close joystick with SDL
+    if (jpad != nullptr) {
+        SDL_JoystickClose(jpad);
+    }
+}
+
+void SDLController::Poll() {
+    if (SDL_PollEvent(&input_event)) {
```
linkmauve: s/if/while/, because there could be multiple events in queue.

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+    KeyMap::SetKeyMapping({202, joypad_id}, Service::HID::PAD_CIRCLE_UP);
+    KeyMap::SetKeyMapping({203, joypad_id}, Service::HID::PAD_CIRCLE_DOWN);
+}
+
+void SDLController::SetDeadZone(int range) {
+    dead_zone = range;
+}
+
+bool SDLController::CheckDeadZone(int range) const {
+    if ((range < -dead_zone) || (range > dead_zone)) {
```
linkmauve: Check against std::abs(dead_zone) instead.

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+        return true;
+    } else if ((range > -dead_zone) && (range < dead_zone)) {
```
linkmauve: Drop the else as it happens after a return, and in every other case you want to return false.
Also no brackets since the if contains only one line.

src/input_common/input_common.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return device_name;
+}
+
+Service::HID::PadState ControllerBase::GetPadState() const {
+    return controller.pad_state;
+}
+
+void Init(ControllerBackends backend) {
+    switch (backend) {
+        #ifdef HAS_SDL
```
bunnei: Ugly... Seeing this reaffirms to me that this PR should be contingent on SDL2 being a mandatory dependency, and it should probably just replace GLFW. Could you submit such  a PR as a prereq to this one? It should be quite trivial.

src/input_common/input_common.cpp
```diff
@@ -0,0 +1,85 @@
...
+            //SDL2 backend selected
+            case SDL2_JOY:
+                g_user_input = new SDLController();
+                break;
+        #endif
+
+        //No backend selected
```
bunnei: FWIW re: the comment above, please search+replace all your comments to be `// comment` (with the space)

src/input_common/input_common.cpp
```diff
@@ -0,0 +1,85 @@
...
+        case NO_JOY:
+            break;
+
+        //What?
```
bunnei: Bad comment, probably unnecessary. Default case statement should log an error (probably at the "critical" level)

src/input_common/input_common.cpp
```diff
@@ -0,0 +1,85 @@
...
+static int input_event;
+
+//Ticks for CoreTiming event, same as VBlank in GPU
+static u64 frame_ticks;
+
+/// Callback for CoreTiming
+static void InputCallback(u64 userdata, int cycles_late) {
+
+    //Call user input plugin Poll()
+    if (g_user_input != nullptr) {
```
bunnei: nullptr check here is not very elegant... I would initialize `g_user_input` at runtime, then add an `IsValid()` (or something similar) method to do this check.

src/input_common/input_common.cpp
linkmauve: I think this can be done later, there should be no issue with using both at once.

I’m not sure SDL2 should be mandatory, we could just make input_common expose no joystick at all, right?

src/input_common/input_common.cpp
bunnei: ditto with the other checks below...

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+            controller.pad_state.hex &= ~KeyMap::GetPadKey({pad_id, joypad_id}).hex;
+            break;
+
+        //Joystick axis motion events
+        case SDL_JOYAXISMOTION:
+            pad_id = 200;
+            pad_id += (input_event.jaxis.axis * 2);
+            val = input_event.jaxis.value;
+
+            if (val > 0) {
```
linkmauve: Same here, and in many other places, we prefer not to have unnecessary brackets when there is only one line in the then (*and* else) part(s).

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+}
+
+void SDLController::DiscoverDevices() {}
+
+void SDLController::Shutdown() {
+    //Attempt to close joystick with SDL
+    if (jpad != nullptr) {
+        SDL_JoystickClose(jpad);
+    }
+}
```
linkmauve: You should also call SDL_Quit() here.

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+    if (SDL_Init(SDL_INIT_JOYSTICK) != 0) {
+        return false;
+    }
+
+    //Attempt to open joystick with SDL
+    //TODO - Use DiscoverDevice to generate a list of available joysticks, then open the correct index
+    jpad = SDL_JoystickOpen(index);
+
+    if (jpad == nullptr) {
+        return false;
```
linkmauve: Call Shutdown() here, probably.

src/input_common/input_common.h
```diff
@@ -0,0 +1,60 @@
...
+#pragma once
+
+#include "common/common_types.h"
+#include "controller_state.h"
+
+namespace InputCommon {
+
+/// Enum defining available backends
+enum ControllerBackends {
+    NO_JOY,
```
bunnei: `JOY` naming is _so_ GameCube scene :D How about making this an `enum class`, then renaming `NO_JOY` to `NONE` and `SDL2_JOY` to `SDL2`? That way, these enumerations will be explicitly referenced (e.g. `ControllerBackend::SDL2`), so the "JOY" part is now redundant.

src/input_common/input_common.h
```diff
@@ -0,0 +1,60 @@
...
+
+#pragma once
+
+#include "common/common_types.h"
+#include "controller_state.h"
+
+namespace InputCommon {
+
+/// Enum defining available backends
+enum ControllerBackends {
```
bunnei: `InputBackend` is a more consistent name IMO

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+    return true;
+}
+
+void SDLController::DiscoverDevices() {}
+
+void SDLController::Shutdown() {
+    //Attempt to close joystick with SDL
+    if (jpad != nullptr) {
+        SDL_JoystickClose(jpad);
+    }
```
linkmauve: Set back jpad to nullptr, so other checks like that won’t take the wrong branch.

src/input_common/input_common.h
```diff
@@ -0,0 +1,60 @@
...
+    NO_JOY,
+    SDL2_JOY
+};
+
+class ControllerBase {
```
bunnei: `InputBase` ?

bunnei: Also, this class should inherit from `NonCopyable`

src/core/hle/service/hid/hid.cpp
linkmauve: Also add a comment saying it’s only a stop-gap solution until both keyboard backends are ported to InputCommon.

externals/cmake-modules/FindSDL2.cmake
```diff
@@ -0,0 +1,180 @@
...
+# see accompanying file Copyright.txt for details.
+#
+# This software is distributed WITHOUT ANY WARRANTY; without even the
+# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the License for more information.
+#=============================================================================
+# (To distribute this file outside of CMake, substitute the full
+#  License text for the above reference.)
+
+FIND_PATH(SDL2_INCLUDE_DIR SDL.h
```
linkmauve: I find this solution very poor, at least on UNIX-like the path to the include directory shouldn’t be hardcoded, you should instead use pkg-config.

externals/cmake-modules/FindSDL2.cmake
```diff
@@ -0,0 +1,180 @@
...
+  /usr/local/include/SDL2
+  /usr/include/SDL2
+  /sw # Fink
+  /opt/local # DarwinPorts
+  /opt/csw # Blastwave
+  /opt
+)
+#MESSAGE("SDL2_INCLUDE_DIR is ${SDL2_INCLUDE_DIR}")
+
+FIND_LIBRARY(SDL2_LIBRARY_TEMP
```
linkmauve: Same here.

src/input_common/sdl_input/sdl_input.h
```diff
@@ -0,0 +1,57 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <SDL2/SDL.h>
```
linkmauve: #include <SDL.h>, so that users who put it elsewhere than in a SDL2 directory can still include it.

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+                    controller.pad_state.hex &= ~KeyMap::GetPadKey({pad_id+2, joypad_id}).hex;
+                    controller.pad_state.hex |= KeyMap::GetPadKey({pad_id+3, joypad_id}).hex;
+                    break;
+            }
+
+            break;
+    }
+}
+
+void SDLController::SetupKeyMaps() {
```
linkmauve: It would be good to add configurability right now, you just have to use Settings::values.pad_*_key instead of hardcoding them.

linkmauve: Maybe you should use the SDL_GameController API instead of the SDL_Joystick one, and depend on such a database: https://github.com/gabomdq/SDL_GameControllerDB

src/input_common/input_common.h
```diff
@@ -0,0 +1,60 @@
...
+};
+
+/// Initialize the user input system
+void Init(ControllerBackends backend);
+
+/// Deactive the user input system
+void Shutdown();
+
+/// InputCommon 'plugin'
+extern ControllerBase* g_user_input;
```
neobrain: use unique_ptr instead

src/input_common/input_common.h
neobrain: +1.

A more modern terminology would be "gamepad" rather than "joy", btw.

src/input_common/input_common.h
```diff
@@ -0,0 +1,60 @@
...
+    virtual void DiscoverDevices() = 0;
+
+    /// Grab and process input
+    virtual void Poll() = 0;
+
+    /// Shuts down all backend related data
+    virtual void Shutdown() = 0;
+
+    /// Returns internal name of currently selected device. Expose this to the UI
+    std::string GetDeviceName() const;
```
neobrain: When returning non-trivial objects from functions, make them const so that people don't accidentally use them as references.

neobrain: Actually, this can even return a const reference.

src/input_common/sdl_input/sdl_input.h
```diff
@@ -0,0 +1,57 @@
...
+    void SetupKeyMaps();
+
+    /// Sets analog deadzones
+    void SetDeadZone(int range);
+
+    /// Checks if analog input is within the dead zone
+    bool CheckDeadZone(int range) const;
+
+private:
+    SDL_Joystick* jpad;
```
neobrain: unique_ptr

src/input_common/sdl_input/sdl_input.h
```diff
@@ -0,0 +1,57 @@
...
+#include <SDL2/SDL.h>
+#include <iostream>
+#include "input_common/input_common.h"
+
+namespace InputCommon {
+
+class SDLController final : public ControllerBase {
+public:
+    SDLController();
+    ~SDLController();
```
neobrain: shouldn't this be marked override, too?

src/input_common/sdl_input/sdl_input.h
```diff
@@ -0,0 +1,57 @@
...
+    bool CheckDeadZone(int range) const;
+
+private:
+    SDL_Joystick* jpad;
+
+    /// SDL event used for polling
+    SDL_Event input_event;
+
+    /// Index of joystick, used for opening/closing a joystick
+    int index;
```
neobrain: Use unsigned for values which aren't ever intended to have a sign.

src/input_common/sdl_input/sdl_input.h
```diff
@@ -0,0 +1,57 @@
...
+    void Shutdown() override;
+
+    /// Processes input when polling
+    void ProcessInput();
+
+    /// Sets up keymaps from input configuration
+    void SetupKeyMaps();
+
+    /// Sets analog deadzones
+    void SetDeadZone(int range);
```
neobrain: Needs documentation on the "range" parameter.

src/input_common/input_common.h
```diff
@@ -0,0 +1,60 @@
...
+
+    /// Returns internal pad state
+    Service::HID::PadState GetPadState() const;
+
+    bool activated;
```
neobrain: public members *need* to be documented properly.

src/input_common/sdl_input/sdl_input.cpp
neobrain: Agreeing that configurability should be taken care of in the initial PR rather than in a followup one.

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+    KeyMap::SetKeyMapping({100, joypad_id}, Service::HID::PAD_A);
+    KeyMap::SetKeyMapping({101, joypad_id}, Service::HID::PAD_B);
+    KeyMap::SetKeyMapping({102, joypad_id}, Service::HID::PAD_X);
+    KeyMap::SetKeyMapping({103, joypad_id}, Service::HID::PAD_Y);
+    KeyMap::SetKeyMapping({104, joypad_id}, Service::HID::PAD_R);
+    KeyMap::SetKeyMapping({105, joypad_id}, Service::HID::PAD_L);
+    KeyMap::SetKeyMapping({0, joypad_id}, Service::HID::PAD_ZL);
+    KeyMap::SetKeyMapping({0, joypad_id}, Service::HID::PAD_ZR);
+
+    KeyMap::SetKeyMapping({107, joypad_id}, Service::HID::PAD_START);
```
neobrain: What's with all these hardcoded numbers?

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+}
+
+SDLController::~SDLController() {
+    Shutdown();
+}
+
+bool SDLController::Init() {
+
+    //Attempt to initialize SDL with joystick only
+    if (SDL_Init(SDL_INIT_JOYSTICK) != 0) {
```
neobrain: Wow, I didn't know SDL actually still calls this stuff joystick, too...

src/input_common/input_common.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    //Reschedule recurrent event
+    CoreTiming::ScheduleEvent(frame_ticks - cycles_late, input_event);
+}
+
+std::string ControllerBase::GetDeviceName() const {
+    return device_name;
+}
+
+Service::HID::PadState ControllerBase::GetPadState() const {
```
neobrain: should return a const reference, probably.

src/core/system.cpp
neobrain: Will this code actually work if SDL2 is not installed on the system?

src/core/hle/service/hid/hid.cpp
neobrain: This looks really fishy, and by design prevents combining e.g. gamepad button input and mouse touchscreen input until InputCommon is in some way merged with EmuWindow. Note that I don't think with the current design it is sensible to move keyboard+mouse input from EmuWindow to InputCommon (I'll probably write a comment on that later on).

src/input_common/input_common.h
```diff
@@ -0,0 +1,60 @@
...
+
+protected:
+    /// Internal name of currently selected device
+    std::string device_name;
+
+    /// Internal view this specific object will have of 3DS input states
+    ControllerState controller;
```
neobrain: Why are the touch screen and circlepad inputs not exposed publically in this interface?

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+void SDLController::ProcessInput() {
+    //Use a unique id for joystick input
+    int pad_id = 0;
+    int val = 0;
+
+    //Generate the id based on input from an event
+    switch (input_event.type) {
+        //Joystick button press events
+        case SDL_JOYBUTTONDOWN:
+            pad_id = 100;
```
neobrain: Where do these constants come from?

src/input_common/sdl_input/sdl_input.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+            break;
+
+        //Joystick hat motion events
+        case SDL_JOYHATMOTION:
+            pad_id = 300;
+            pad_id += input_event.jhat.hat * 4;
+
+            switch (input_event.jhat.value) {
+                case SDL_HAT_CENTERED:
```
neobrain: This whole switch case looks like voodoo. What's going on there?

neobrain:
Made an initial pass over the code. Looking mostly good modulo coding style issues ;)

There's one design change I would like to see implemented: Input grabbing shouldn't be core emulator work but instead frontend work, which is one of the reasons it's part of EmuWindow to begin with. Note that keyboard and mouse handling should in any case stay in EmuWindow since they are handled by the GUI toolkit used by the frontend (not getting those inputs from the GUI is a mess). Moving things to EmuWindow is hence consistent and beneficial overall, since it gets rid of the ugly `g_user_input ? InputCommon::... : EmuWindow:... ` checks in core code. InputCommon interface itself need not be changed (much) for this, but the place where Poll() gets called should be moved to a thread managed by the EmuWindow implementations and the PadState/ControllerState should be publicly exposed only by EmuWindow.

src/input_common/sdl_input/sdl_input.cpp
shonumi: I guess I should better commented about pad_ids. They're a way I came up with for SDL to assign unique numbers  for controller input. It was designed to be simple and easy to remember (unlike using horrible hashes like Mednafen iirc). 

Essentially the first analog input is assigned values 200, 201, 202, 203, corresponding for left, right, up, and down movements. Whenever the user moves the stick to the left, it's physically impossible to activate any input that depends on right movement. With pad_ids, we can always tell which id is "opposite" by looking at whether the id is odd or even (modulo 2). If it's even, we need to release the next highest pad_id (200 -> 201); if it's odd we need to release the next the next lowest pad_id (201 -> 200)

So better commenting? Hope that clears up the "voodoo". I was going to make a gist properly detailing how pads worked, since I didn't feel it quite right to make it a wiki item on citra (but I could have made it a wiki page on the branch >_<)

src/input_common/sdl_input/sdl_input.cpp
shonumi: They come from SDL.h (which pulls SDL_joystick.h)

src/input_common/sdl_input/sdl_input.cpp
shonumi: Initial work until they can be read from a file. Does Citra have some generic file parsing set up already? I would prefer not to have to come up with a custom solution.

src/input_common/sdl_input/sdl_input.cpp
shonumi: Yeah, you're right. Looks like I need to switch to the SDL GameController API.

externals/cmake-modules/FindSDL2.cmake
shonumi: Fwiw, this was a copy+paste job from Dolphin's CMake module :p I'll change it soon in this PR.

src/input_common/input_common.cpp
shonumi: I'm fine with making SDL2 mandatory. I was just concerned that we might want to use other APIs in the future. Is this a valid concern? I remember MaJoR explaining why SDL2 was removed as an input backend for Dolphin on Windows (something about webcams getting detected as controllers) but I'm sure we can avoid or workaround any pitfalls.

src/core/system.cpp
shonumi: @bunnei - Agreed. Sorry, as you can see this PR is more a Preliminary Review rather than a Pull Request :/ It's rough, I apologize for stuff like this.

@neobrain - It compiles just fine with Travis CI. The CMake log shows it can't find SDL but builds properly.

src/input_common/input_common.h
shonumi: Honestly overlooked that :/ Wow

src/core/hle/service/hid/hid.cpp
shonumi: @neobrain - Actually touchscreen input + gamepad input works. Touchscreen input (from EmuWindow) does not come from it's PadState. That's handled by separate members. I tested this from my end on a few games. It does prevent keyboard input from EmuWindow and gamepad input from occurring at the same time. But yes, the code here is an ugly temp solution...

shonumi:
@neobrain - I kind of struggled finding a place for Poll() initially :p I'm still not sure where everyone stands on EmuWindow though. Some don't think it should handle input (at all), so I'd like to have a definitive direction in regards to that. 

I'm totally all for Polling in EmuWindow, but it rubs me the wrong way having to directly process "raw" input in EmuWindow ([like this](https://github.com/citra-emu/citra/blob/master/src/common/emu_window.cpp#L9)). At least, processing directly seems weird to do when that's InputCommon's job. It would be necessary to write out code for each specific GUI we have, but if bunnei says we can eliminate GLFW, that means just Qt for the time being.

So in short, how involved should EmuWindow be with input (Polling, Pass "raw" input vs. processing it directly, none at all)?

src/input_common/sdl_input/sdl_input.cpp
shonumi: If it were passed it as an argument to SDLController::Poll(), what would be the best way to call it from [here](https://github.com/shonumi/citra/blob/input_common/src/input_common/input_common.cpp#L30)?

src/input_common/sdl_input/sdl_input.cpp
shonumi: About the else parts, like [this](https://gist.github.com/shonumi/a4725247ebaa3c6d9a7b)?

src/input_common/input_common.cpp
shonumi: How about using unique pointer (as neobrain suggested) and just doing a simple check like `if (g_user_input)`, or would you still prefer my making an `IsValid()` function?

src/input_common/sdl_input/sdl_input.h
shonumi: When using unique_ptr, how does one set the stored pointer when that pointer comes from a function. See [here](https://github.com/shonumi/citra/blob/input_common/src/input_common/sdl_input/sdl_input.cpp#L37), I'm having trouble figuring out how to do it properly.

Subv: You cant use unique_ptr here, this code is not the owner of the pointer, and it should not be released by calling delete on it (You should use SDL_JoystickClose on it, and you're already doing that as far as i can see).

neobrain:
@shonumi My opinion is that all of it should be done in EmuWindow one way or the other. You're free to just have the EmuWindow implementation refer to InputCommon however. That way we can have a single InputCommon framework for all EmuWindow implementations.

neobrain:
Unfortunately, it seems like @shonumi will not be able to work on this any further. (correct me if I'm wrong!)

shonumi:
Sorry, I'm still dealing with personal matters atm; it leaves very little time for other projects. Hopefully that will change in the future.

PR #839, opened 2015-05-31
Enforce the "no tab, use spaces" policy with a pre-commit hook.

Lectem:
Enforce the "no tab, use spaces" policy with a pre-commit hook.
This hook is installed through cmake if the user didn't set up any pre-commit hook before (so that we don't override it).

This will also check for trailing spaces

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,2 +1,0 @@
-[core]
-    whitespace = blank-at-eol,trailing-space,tab-in-indent
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,2 @@
+[core]
+    whitespace = blank-at-eol,trailing-space,tab-in-indent
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,13 @@ cmake_minimum_required(VERSION 2.8.11)
 
 project(citra)
 
+if(NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git/hooks/pre-commit)
+    message(STATUS "Copying pre-commit hook")
+    file(COPY hooks/pre-commit
+        DESTINATION ${CMAKE_CURRENT_SOURCE_DIR}/.git/hooks
+        FILE_PERMISSIONS WORLD_EXECUTE )
+endif()
+
 if (NOT MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -pthread")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+#check the config, in case the user really wants to allow tabs...
+allowtabs=$(git config hooks.allowtabs)
+if [ "$allowtabs" != "true" ] &&
+   git diff --cached | egrep '^\+.*	'
+then
+   cat<<END;
+Error: This commit would contain a tab, which is against this repo's policy.
+END
+  exit 1
+fi
+
+# If there are whitespace errors, print the offending file names and fail.
+if
+# Use git built-in checks for trailing whitespaces
+    ! git diff --check --cached
+then
+   cat<<END;
+Error: This commit would contain trailing spaces, which is against this repo's policy.
+END
+  exit 1
+fi
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,23 +1,24 @@
 #!/bin/sh
 
-#check the config, in case the user really wants to allow tabs...
-allowtabs=$(git config hooks.allowtabs)
-if [ "$allowtabs" != "true" ] &&
-   git diff --cached | egrep '^\+.*	'
-then
-   cat<<END;
-Error: This commit would contain a tab, which is against this repo's policy.
-END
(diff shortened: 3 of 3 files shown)


hooks/pre-commit
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+if git rev-parse --verify HEAD >/dev/null 2>&1
```
neobrain: More comments please: explain what you're doing throughout this file.

CMakeLists.txt
```diff
@@ -4,6 +4,13 @@ cmake_minimum_required(VERSION 2.8.11)
 
 project(citra)
 
+if(NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git/hooks/pre-commit)
```
neobrain: Will this actually work on Windows?

Lectem: I just tested it, it's working for me (even without msys and stuff). cmake usually does the hard job by converting paths.

hooks/pre-commit
Lectem: I didn't comment it because it's from the hooks samples of git. (you can check your .git/hooks directory)
This file is actually the same as the pre-commit.sample except for the tab check

CMakeLists.txt
neobrain: I was more concerned about running the sh script on Windows... :p

Lectem: If git is installed, then hooks will work.
Git has an embedded version of msys on Windows.

archshift:
Could you please change the commit message to something more descriptive?

.gitconfig
```diff
@@ -0,0 +1,2 @@
+[core]
+    whitespace = blank-at-eol,trailing-space,tab-in-indent
```
linkmauve: There is no \n at the end of this line.

CMakeLists.txt
```diff
@@ -4,6 +4,13 @@ cmake_minimum_required(VERSION 2.8.11)
...
+message(STATUS "Copying pre-commit hook")
```
linkmauve: Wrong indentation here, and for the three next lines.

.gitconfig
Lectem: Does it matter somehow ?

archshift: Yes, it messes with UNIX utilities. Please fix that.

archshift:
LGTM once you squash.

hooks/pre-commit
```diff
@@ -0,0 +1,23 @@
...
+
+#check the config, in case the user really wants to allow tabs...
+allowtabs=$(git config hooks.allowtabs)
+if [ "$allowtabs" != "true" ] &&
+   git diff --cached | egrep '^\+.*	'
+then
+   cat<<END;
+Error: This commit would contain a tab, which is against this repo's policy.
+END
+  exit 1
```
yuriks: This file is intended 3 spaces on the rest of the file but two in this line. Indent everything using 4 spaces like the project standard.

hooks/pre-commit
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+#check the config, in case the user really wants to allow tabs...
+allowtabs=$(git config hooks.allowtabs)
+if [ "$allowtabs" != "true" ] &&
+   git diff --cached | egrep '^\+.*	'
```
yuriks: Doesn't `tab-in-indent` in the git check below already check for this?

.gitconfig
yuriks: `trailing-space` already include `blank-at-eol` so you don't need to add it. Suggest also including `space-before-tab` in the list.

CMakeLists.txt
yuriks: Git requires sh to work (some of its commands are implemented as scripts) so yes, this is fine.

Lectem:
I made a mistake with the config before, I thought that .gitconfig would work, while it is only for a user config, not a local config.
Instead I set it up so that the hook would config the whitespace and got rid of grep.

hooks/pre-commit
Lectem: fixed

hooks/pre-commit
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+# Enforce citra's whitespace policy
+git config --local core.whitespace tab-in-indent,trailing-space
+
+# If there are whitespace errors, print the offending file names and fail.
+if
+    ! git diff --cached --check
```
yuriks: Why is this on a separate line?

Lectem: fixed

Lectem:
I put back the grep since tab-in-indent only checks for tabs at the beginning of a line

hooks/pre-commit
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+# Enforce citra's whitespace policy
+git config --local core.whitespace tab-in-indent,trailing-space
+
+# If there are whitespace errors, print the offending file names and fail.
+if ! git diff --cached --check
+then	//test
```
yuriks: `// test`???

Also: `if ! git diff --cached --check; then` please.

Lectem: Aw sorry, got some of my tests mixed up here, will ammend

yuriks:
LGTM

neobrain:
@archshift @yuriks if you think this is still okay, please merge it.

PR #841, opened 2015-05-31
Have SDL2 as a project dependency

shonumi:
This PR is a prerequisite to [PR #836](https://github.com/citra-emu/citra/pull/836) The goal is to eventually eliminate GLFW in place of SDL2 and use SDL2 for input.


shonumi:
As noted [here](https://github.com/citra-emu/citra/pull/836#discussion-diff-31388185) the CMake module from Dolphin appears to be suboptimal. CMake is not my particularly strong point, so pointers would be appreciated on how to deal with pkg-config.

ILOVEPIE:
glfw is faster and simpler I don't see the benefit to using SDL2

lioncash:
Thanks for providing or linking to benchmarks that prove otherwise.

ILOVEPIE:
SDL is a large library that has higher overhead than glfw. also glfw is
preinstalled on practically every system. the only benifit I see to sdl is
it's cross platform audio support.

yuriks:
> also glfw is preinstalled on practically every system

Sorry, but that's just patently false. Stop talking out of your ass.
https://qa.debian.org/popcon.php?package=libsdl2
https://qa.debian.org/popcon.php?package=glfw3

ILOVEPIE:
that only shows debian

Lectem:
Does it even matter if one or the other is more widespread anyway ? I mean for the users it won't change much, and I don't think most devs would worry about installing one or the other...

lioncash:
@Lectem It doesn't really matter, I agree.

@ILOVEPIE Stop making claims without anything backing them up. You do this quite often and you seem surprised people actually don't respond positively when it's quite obvious what the reason is.

yuriks:
@ILOVEPIE You can check the Ubuntu (by far the most popular Linux distro) popcon result, and the difference is even more massive: http://popcon.ubuntu.com/

@Lectem @lioncash It doesn't *really* matter, especially seeing as both libraries have pretty similar APIs (except for GLFW lacking audio), but given the choice between two roughly equivalent libraries I see no reason for going with the less popular one.

lioncash:
Yeah, what I had meant is that I don't mind whichever one we go with. SDL2 is equally fine.

shonumi:
For all of those who don't know yet, this is a change bunnei wants to see, so it didn't just come out of the blue. Personally, I favor SDL2 being mandatory. but the upcoming InputCommon is designed to be agnostic; it doesn't bother me if SDL2 is optional or not chosen at all. Requiring SDL2 was one solution to get rid of [nasty checks like this](https://github.com/shonumi/citra/blob/input_common/src/input_common/input_common.cpp#L33) If SDL2 is to be optional, I would like suggestions on how to better, more cleanly detect the presence of the library in the codebase. 

One concern of retaining glfw is gamepad input across different front-ends. With SDL2, Citra can grab gamepad input pretty much independently from the front-end. So whether Citra uses Qt, glfw, or something else, gamepads can be supported by the emulator. I'm not certain how well glfw can grab gamepad input unless it's tied to the glfw front-end.

@ILOVEPIE - How well can glfw handle gamepads outside of glfw itself? That is to say, if the user chooses, citra-qt, how can glfw be called to handle gamepads? Will it be messy, simple, convoluted, elegant?

bunnei:
Ditto that we will for sure be getting rid of GLFW if we're adding SDL for other things (e.g. input, and/or sound in the future). For this change, I also had in mind a few other things: 

* Windows libraries/headers for SDL2 put in externals, e.g. like -> https://github.com/citra-emu/citra/tree/master/externals/glfw-3.1.1.bin
* GLFW removed, and use of it replaced with SDL2, e.g. here -> https://github.com/citra-emu/citra/blob/master/src/citra/emu_window/emu_window_glfw.cpp (Note: this should be very trivial, since there are 1:1 equivalent SDL2 calls to everything we are using in GLFW).
This should mean that you no longer need any conditionally built SDL code in #836.

shonumi:
Thank you for the details bunnei :) I guess I'll start tackling that tonight.

Subv:
Keep in mind that SDL2 does not support recording audio, so we'll need yet another library to handle that

linkmauve:
@bunnei I’m very much against bloating the repository with yet another binary files, if you want external dlls it would be better to use a subrepository or some script to retrieve them from http://libsdl.org/ imo.

shonumi:
@bunnei, @linkmauve - What about building it from externals before compiling Citra? That is to say, build it locally, then statically link to it? Dolphin does this iirc for some libraries. SDL is pretty minimal to compile anyway. This would eliminate binaries files in the repo and simplify dependency checking in CMake. 

Assuming we ignore whatever version the user has installed, this would give us a definite location of the headers with no ambiguity and allow us greater control (say SDL 2.0.whatever does something we don't like, we can force a version more suitable for Citra). Getting all of this sorted out is necessary to proceed, imo, unless someone tells me otherwise.

Currently getting the SDL port up to speed to what the GLFW version was. Expect some commits here in a bit.

yuriks:
> Assuming we ignore whatever version the user has installed

I think this is quite undesirable, especially so on Linux, where I'd even call it a showstopper. Even on Windows, the current setup allows you to override the GLFW location if you so desire.

I think @linkmauve's suggestion of having a separate submodule repository for binaries is a good solution, it prevents repo bloat while still being seamless to use and simple to integrate.

shonumi:
Just throwing ideas out there. If @bunnei agrees, I'd like to go the submodule route too.

Lectem:
Let me be the advocate of http://www.cmake.org/cmake/help/v3.0/module/ExternalProject.html instead of just using a submodule. It can be used with binaries too.
This way we would all have the good version of the dependencies without having to git init/update.
The downside is that you have to redownload the binaries if you change your build dir.

shonumi:
Well, as it stands we have at least 3 potential solutions. I still need confirmation on a single direction, so whenever that comes down I'll implement whatever we choose.

neobrain:
Could you provide a tl;dr on these 3 options so that people can jump in without reading the whole comment section?

shonumi:
1) Add SDL2 to externals, 2) Grab SDL2 from a git sub-module or, 3) use CMake External Projects

Fwiw, bunnei and I discussed this a while ago on IRC, and the decision was to go with option 1. In the end, we actually have less binaries floating around in the project (since GLFW has at least 4 different sets depending on the compiler). Haven't gotten around to it so far because of RL backlog stuff.

bunnei:
@shonumi, what is the status of this? I think we agreed to just submodule the SDL headers and binaries for Windows users. (Fwiw I wouldn't mind doing a minimal Qt submodule like this too since the full Windows install is monstrous).

shonumi:
@bunnei - Sorry, the past few weeks have been rough for me (personally, emotionally), so I didn't have any motivation to do much. I think my head's straight now, so I'll work on the submodule method soon.

yuriks:
@bunnei For Qt even just the DLLs we use are 40 megs, so I'd honestly pass on that idea.

hrydgard:
Our FFMPEG submodule is bigger than that (counting all archs) with no issues :P  It works fine, Git can do binaries (although it's best not to have them in your main repo indeed)

bunnei:
@yuriks I'm not sure I see the issue with 40mb... You still have the option to use the 500mb+ official installer if you'd like. Maybe one of the Dolphin devs can comment on how well https://github.com/dolphin-emu/ext-win-qt works out? (Which is a bit bigger, ~75mb)

shonumi:
I'm actually unable to work on Citra for the foreseeable future, so if anyone want to pick this up, feel free to do so.

PR #842, opened 2015-06-01
Filter xcodebuild's output

chinhodado:
xcodebuild's output is incredibly verbose. This will grep out only lines with error/warning.

Before: [~15,000 lines](https://s3.amazonaws.com/archive.travis-ci.org/jobs/64956564/log.txt)
After:  [~500 lines](https://api.travis-ci.org/jobs/64957057/log.txt?deansi=true)


.travis-build.sh
```diff
@@ -18,5 +18,5 @@ elif [ "$TRAVIS_OS_NAME" = "osx" ]; then
     export Qt5_DIR=$(brew --prefix)/opt/qt5
     mkdir build && cd build
     cmake .. -GXcode
-    xcodebuild -configuration Release
+    xcodebuild -configuration Release | egrep '^(/.+:[0-9+:[0-9]+:.(error|warning):|fatal|===)'
```
neobrain: comment plz.

chinhodado: You mean explaining the regex, what it matches etc.?

archshift:
I'm not sure this is the best approach. How about one of the many tools to prettify xcodebuild's output?

chinhodado:
I only know xcpretty, and it doesn't support displaying warnings.

archshift:
What about using `xctool` instead of `xcodebuild`?

archshift:
Also apparently xcpretty [supports warnings now](https://github.com/supermarin/xcpretty/issues/108#issuecomment-107734863).

chinhodado:
Cool, I'll close this one then.

PR #843, opened 2015-06-01
Use correct style name: PascalCase instead of CamelCase

chinhodado:


yuriks:
LGTM

neobrain:
Uhm, is PascalCase actually the more common thing to say? I never heared about that and find it pretty confusing. After some research, it seems that only some microsoft documentation makes a distinction between CamelCase and PascalCase...

yuriks:
As far as I know camel case is widely understood to mean `camelCaseLikeThis`. `ThisConvention` is usually called camel case qualified with something else.

chinhodado:
When we say PascalCase, there's no ambiguity. Camel case usually means `camelCaseLikeThis`, but can also mean `CamelCaseLikeThis` sometimes too, which is not something we would want.

yuriks:
Can we just call it "capitalized letters joined without separators" and call it a day?

```
[18:10:20] <yuriks> I need a device that will automatically launch nuclear weapons to anihilate all humanity in case someone starts having a bikeshedding discussion
[18:10:53] <yuriks> this will either eliminate bikeshedding or it won't. In either case the outcome is probably positive
```

neobrain:
Fair enough.

PR #848, opened 2015-06-04
arm_dyncom_thumb: Fix handling of writeback for thumb LDMIA

lioncash:
Prior to this, it'd assume writeback always occurred. However, it should only occur if Rn is not in the register list. Writeback always occurs with STMIA in thumb mode, so it's fine.


bunnei:
LGTM

PR #849, opened 2015-06-08
Fix svcWaitSynch to correctly acquire wait objects

### LINKED ISSUE #432: Thread scheduling bugs
@Normmatt wrote a threading demo (that of which I have not seen the sourcecode for) that tests scheduling, and ran it against Citra, 3dmoo, and hardware. The results are here -> https://gist.github.com/Normmatt/b0a0440c9f55d1b54664

At first glance Citra looks very wrong... But I'm not sure that is true. It looks like there is simply a bug with new threads being immediately switched to, instead of only being switched to on the next qualifying event. If I move the "begins" to happen first for the Citra log, you get:

<pre>Thread 1 begins
Thread 2 begins
Thread 3 begins
Thread 4 begins
Thread 1, count = 0
Thread 2, count = 0
Thread 1, count = 1
Thread 2, count = 1
Thread 1, count = 2
Thread 1 ends
Thread 3, count = 0
Thread 2, count = 2
Thread 2 ends
Thread 3, count = 1
Thread 4, count = 0
Thread 3, count = 2
Thread 3 ends
Thread 4, count = 1</pre>

Which is actually pretty darn close to hardware - so it's probably just a simple fix. I'm going to take the action to go through thread scheduling after @Subv makes his upcoming timing changes, and make it more identical to hardware.

bunnei:
This changes how we emulate svcWaitSynch to ensure that wait objects are always acquired by the right thread. My solution with this PR is to always call the SVC when a thread is awoken until conditions are satisfied for it to continue. A side-effect of this is that threads will often be scheduled that are not actually able to run, resulting in unnecessary context switches (this could result in a hopefully minor performance hit). This is because when a wait_object becomes available, all waiting threads will be awoken, so whichever don't acquire the resource will still be scheduled then rescheduled. However, this actually makes our code much simpler/cleaner, as we can get rid of all the state management cruft trying to "predict" the next best thread to run. This fixes #432. I suspect it will fix a lot of games, although I don't own enough games to verify this claim.

Mario Kart 7 now will go in game (NOTE: needs a few other patches that aren't mainline yet):

image

### DIFF
--- a/src/core/hle/function_wrappers.h
+++ b/src/core/hle/function_wrappers.h
@@ -9,10 +9,14 @@
 #include "core/arm/arm_interface.h"
 #include "core/memory.h"
 #include "core/hle/hle.h"
+#include "core/hle/result.h"
 
 namespace HLE {
 
 #define PARAM(n)    Core::g_app_core->GetReg(n)
+
+/// An invalid result code that is meant to be overwritten when a thread resumes from waiting
+static const ResultCode RESULT_INVALID(0xDEADC0DE);
 
 /**
  * HLE a function return from the current ARM11 userland process
@@ -57,8 +61,11 @@ template<ResultCode func(s32*, u32*, s32, bool, s64)> void Wrap() {
     s32 param_1 = 0;
     s32 retval = func(&param_1, (Handle*)Memory::GetPointer(PARAM(1)), (s32)PARAM(2),
         (PARAM(3) != 0), (((s64)PARAM(4) << 32) | PARAM(0))).raw;
-    Core::g_app_core->SetReg(1, (u32)param_1);
-    FuncReturn(retval);
+
+    if (retval != RESULT_INVALID.raw) {
+        Core::g_app_core->SetReg(1, (u32)param_1);
+        FuncReturn(retval);
+    }
 }
 
 template<ResultCode func(u32, u32, u32, u32, s64)> void Wrap() {
@@ -73,7 +80,11 @@ template<ResultCode func(u32*)> void Wrap(){
 }
 
 template<ResultCode func(u32, s64)> void Wrap() {
-    FuncReturn(func(PARAM(0), (((s64)PARAM(3) << 32) | PARAM(2))).raw);
+    s32 retval = func(PARAM(0), (((s64)PARAM(3) << 32) | PARAM(2))).raw;
+
+    if (retval != RESULT_INVALID.raw) {
+        FuncReturn(retval);
+    }
 }
 
 template<ResultCode func(void*, void*, u32)> void Wrap(){
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -32,27 +32,13 @@ void WaitObject::RemoveWaitingThread(Thread* thread) {
         waiting_threads.erase(itr);
 }
 
-SharedPtr<Thread> WaitObject::WakeupNextThread() {
-    if (waiting_threads.empty())
-        return nullptr;
-
-    auto next_thread = std::move(waiting_threads.front());
-    waiting_threads.erase(waiting_threads.begin());
-
-    next_thread->ReleaseWaitObject(this);
-
-    return next_thread;
-}
-
 void WaitObject::WakeupAllWaitingThreads() {
-    auto waiting_threads_copy = waiting_threads;
+    for (auto thread : waiting_threads)
+        thread->ResumeFromWait();
 
-    // We use a copy because ReleaseWaitObject will remove the thread from this object's
-    // waiting_threads list
-    for (auto thread : waiting_threads_copy)
-        thread->ReleaseWaitObject(this);
+    waiting_threads.clear();
 
-    ASSERT_MSG(waiting_threads.empty(), "failed to awaken all waiting threads!");
+    HLE::Reschedule(__func__);
 }
 
 HandleTable::HandleTable() {
--- a/src/core/hle/kernel/semaphore.cpp
+++ b/src/core/hle/kernel/semaphore.cpp
@@ -48,13 +48,7 @@ ResultVal<s32> Semaphore::Release(s32 release_count) {
     s32 previous_count = available_count;
     available_count += release_count;
 
-    // Notify some of the threads that the semaphore has been released
-    // stop once the semaphore is full again or there are no more waiting threads
-    while (!ShouldWait() && WakeupNextThread() != nullptr) {
-        Acquire();
-    }
-
-    HLE::Reschedule(__func__);
+    WakeupAllWaitingThreads();
 
     return MakeResult<s32>(previous_count);
 }
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -13,6 +13,7 @@
 #include "common/thread_queue_list.h"
 
 #include "core/arm/arm_interface.h"
+#include "core/arm/skyeye_common/armdefs.h"
 #include "core/core.h"
 #include "core/core_timing.h"
 #include "core/hle/hle.h"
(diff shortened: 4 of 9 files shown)


src/core/hle/kernel/thread.cpp
```diff
@@ -193,8 +192,17 @@ static void SwitchContext(Thread* new_thread) {
...
+        // Cancel any outstanding wakeup events for this thread
+        CoreTiming::UnscheduleEvent(ThreadWakeupEventType, new_thread->callback_handle);
+
         current_thread = new_thread;
 
+        // If the thread was waited by a svcWaitSynch call, step back PC by one instruction to rerun
+        // the SVC when the thread wakes up. This is necessary to ensure that the thread can acquire
+        // the requested wait object(s) before continuing.
+        if (new_thread->waitsynch_waited)
+            new_thread->context.pc -= 4;
```
Subv: This troubles me, can SVCs be called from Thumb mode? and if not, could you add a comment clarifying that, please?

src/core/hle/kernel/semaphore.cpp
```diff
@@ -48,13 +48,7 @@ ResultVal<s32> Semaphore::Release(s32 release_count) {
...
     available_count += release_count;
 
-    // Notify some of the threads that the semaphore has been released
-    // stop once the semaphore is full again or there are no more waiting threads
-    while (!ShouldWait() && WakeupNextThread() != nullptr) {
-        Acquire();
-    }
-
-    HLE::Reschedule(__func__);
+    WakeupAllWaitingThreads();
```
Subv: Note to self, take a closer look to this tomorrow.(@Subv)

src/core/hle/kernel/thread.cpp
Normmatt: Yes they can and this doesn't take that into account.

MoochMcGee:
360 noscope instamerge

src/core/hle/kernel/semaphore.cpp
Subv: Done, this is OK

src/core/hle/kernel/thread.cpp
```diff
@@ -193,8 +192,18 @@ static void SwitchContext(Thread* new_thread) {
...
+        CoreTiming::UnscheduleEvent(ThreadWakeupEventType, new_thread->callback_handle);
+
         current_thread = new_thread;
 
+        // If the thread was waited by a svcWaitSynch call, step back PC by one instruction to rerun
+        // the SVC when the thread wakes up. This is necessary to ensure that the thread can acquire
+        // the requested wait object(s) before continuing.
+        if (new_thread->waitsynch_waited)
+            // CPSR flag indicates CPU mode: SVC instruction is 2 bytes for THUMB, 4 bytes for ARM.
+            new_thread->context.pc -= ((new_thread->context.cpsr & (1 << 5)) != 0) ? 2 : 4;
```
yuriks: Wouldn't kill you to add a separate `bool thumb_mode` variable and would make it much more instantly readable.

src/core/hle/kernel/kernel.cpp
```diff
@@ -26,33 +26,19 @@ void WaitObject::AddWaitingThread(SharedPtr<Thread> thread) {
         waiting_threads.push_back(std::move(thread));
 }
 
-void WaitObject::RemoveWaitingThread(Thread* thread) {
+void WaitObject::RemoveWaitingThread(SharedPtr<Thread> thread) {
```
yuriks: What's up with this change? Pointers to kernel objects can be trivially converted to a SharedPtr, so I usually reserved them to places that actually held ownership.

src/core/hle/kernel/thread.cpp
```diff
@@ -163,8 +163,7 @@ static void PriorityBoostStarvedThreads() {
         u64 delta = current_ticks - thread->last_running_ticks;
 
         if (thread->status == THREADSTATUS_READY && delta > boost_timeout) {
-            const s32 priority = std::max(ready_queue.get_first()->current_priority - 1, 0);
-            thread->BoostPriority(priority);
+            thread->BoostPriority(THREADPRIO_USERLAND_MAX);
```
yuriks: Rationale?

src/core/hle/result.h
```diff
@@ -222,6 +222,9 @@ inline bool operator!=(const ResultCode& a, const ResultCode& b) {
 /// The default success `ResultCode`.
 const ResultCode RESULT_SUCCESS(0);
 
+/// An invalid result code that is meant to be overwritten when a thread resumes from waiting
+const ResultCode RESULT_INVALID(0xDEADC0DE);
```
yuriks: We'd discussed this in the PR this was introduced, and I still think it doesn't belong here and should go in one of the kernel-related headers instead.

bunnei:
Suggestion then? (Since I now use it in function_wrapper.h)

src/core/hle/kernel/thread.cpp
lioncash: there's also a TBITvariable in armdefs.h that represents `(1 << 5)`

src/core/hle/result.h
yuriks: I think I'd put it in function_wrappers.h, which is where you're using it anyway.

src/core/hle/function_wrappers.h
```diff
@@ -9,11 +9,15 @@
...
 #include "core/memory.h"
 #include "core/hle/hle.h"
+#include "core/hle/result.h"
 
 namespace HLE {
 
 #define PARAM(n)    Core::g_app_core->GetReg(n)
 
+/// An invalid result code that is meant to be overwritten when a thread resumes from waiting
+const ResultCode RESULT_INVALID(0xDEADC0DE);
```
lioncash: Shouldn't this be static?

src/core/hle/kernel/thread.cpp
```diff
@@ -193,8 +194,22 @@ static void SwitchContext(Thread* new_thread) {
...
 
+        // Cancel any outstanding wakeup events for this thread
+        CoreTiming::UnscheduleEvent(ThreadWakeupEventType, new_thread->callback_handle);
+
         current_thread = new_thread;
 
+        // If the thread was waited by a svcWaitSynch call, step back PC by one instruction to rerun
+        // the SVC when the thread wakes up. This is necessary to ensure that the thread can acquire
+        // the requested wait object(s) before continuing.
+		if (new_thread->waitsynch_waited) {
```
yuriks: TAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABS!

src/core/hle/svc.cpp
```diff
@@ -241,7 +249,7 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
         Kernel::GetCurrentThread()->WakeAfterDelay(nano_seconds);
 
         // NOTE: output of this SVC will be set later depending on how the thread resumes
-        return RESULT_INVALID;
+		return HLE::RESULT_INVALID;
```
yuriks: TAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABS!

src/core/hle/svc.cpp
```diff
@@ -261,7 +269,7 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
 
     // TODO(bunnei): If 'wait_all' is true, this is probably wrong. However, real hardware does
     // not seem to set it to any meaningful value.
-    *out = wait_all ? 0 : handle_index;
+    *out = handle_count != 0 ? (wait_all ? -1 : handle_index) : 0;
```
yuriks: If handle_count is 0 it sets the index to 0 (aka, an out of bounds index)? Is that right?

yuriks:
LGTM. I think that moving the wait-satisfied checks into the scheduler would be a more flexible and efficient solution, but this is fine for a first implementation which fixes some glaring issues.

src/core/hle/svc.cpp
bunnei: Yes, we in fact have a test for this: https://github.com/citra-emu/hwtests/blob/master/source/tests/kernel/waitsynch.cpp#L371

bunnei:
<Subv> alright, LGTM
<Subv> instamergepls

PR #851, opened 2015-06-09
Robocopy doesn't like trailing slashes

clienthax:


yuriks:
LGTM

(For those wondering: If a path has spaces, then CMake will automatically quote it. Since some of the paths end with a trailing `\` this ends up with a path like `"C:\Foo\Bar\"`. The windows shell interprets the `\"` as a literal `"` instead of closing the quote, and robocopy fails. Doubling the slashes escapes them and prevents this from happening.)

bunnei:
Looks fine

PR #857, opened 2015-06-12
vfp: Handle accesses to the VFP media feature registers

lioncash:
These are able to be accessed in any privilege mode.


Subv:
LGTM

PR #859, opened 2015-06-13
build with visual studio 2015

Apology11:
this prepares citra to be build with Visual Studio 2015.

There is still [a linker error](https://gist.githubusercontent.com/Apology11/c99fe8a3f51cbf587503/raw/76d66f8befa1ec24a48a69cf92bdb0ddd620d2bb/gistfile1.txt), but as far as I´ve read this is a problem [tracked by microsoft](http://stackoverflow.com/questions/30765256/linker-error-using-vs-2015-rc-cant-find-symbol-related-to-stdcodecvt).

### DIFF
--- a/src/common/common_funcs.h
+++ b/src/common/common_funcs.h
@@ -69,8 +69,10 @@ inline u64 _rotr64(u64 x, unsigned int shift){
 }
 
 #else // _MSC_VER
-    // Function Cross-Compatibility
-    #define snprintf _snprintf
+    #if (_MSC_VER < 1900)
+        // Function Cross-Compatibility
+        #define snprintf _snprintf
+    #endif
 
     // Locale Cross-Compatibility
     #define locale_t _locale_t


src/common/common_funcs.h
```diff
@@ -80,8 +80,10 @@ inline u64 _rotr64(u64 x, unsigned int shift){
 }
 
 #else // _MSC_VER
+#if (_MSC_VER != 1900)
```
yuriks: Use `<` instead of `!=`.

yuriks:
"It is not yet fixed, and may not be fixed before RTM." Leave it to MS to ship broken toolchains again.

Is the glfw library an official binary or did you build it yourself from source? We might be switching to SDL2 (which is independent of the compiler version since it only uses the C ABI) soon so maybe we want to hold on merging this. Even if that doesn't happen I think we want to move these library files to externals to avoid the repo bloat.

Apology11:
I build the library myself. Sounds like a good idea, as CMake would throw an error and you can´t even generate the project without the change I made. 

So should I close this and make a new pr for the snprintf define only?

neobrain:
What's up with this PR?

@yuriks mind handling this appropriately?

yuriks:
@Apology11: VS 2015 RTM is coming in a few days, and will require a recompile of the library, so I think we can skip that for the moment. Feel free to update this PR with only the snprintf define thing.

mailwl:
do you know, there is will be Community version of vs2015?

Apology11:
updated

yuriks:
LGTM

PR #860, opened 2015-06-14
Color support for Y2R

yuriks:
This implements support for all combinations of conversion formats to the y2r implementation. There are still some unknown hardware that haven't been RE'd, but they don't seem to affect the output on my tests.

### DIFF
--- a/src/core/hle/service/y2r_u.cpp
+++ b/src/core/hle/service/y2r_u.cpp
@@ -20,7 +20,7 @@
 
 namespace Y2R_U {
 
-enum class InputFormat {
+enum class InputFormat : u8 {
     /// 8-bit input, with YUV components in separate planes and using 4:2:2 subsampling.
     YUV422_Indiv8 = 0,
     /// 8-bit input, with YUV components in separate planes and using 4:2:0 subsampling.
@@ -31,28 +31,28 @@ enum class InputFormat {
     YUV422_BATCH = 4,
 };
 
-enum class OutputFormat {
+enum class OutputFormat : u8 {
     Rgb32 = 0,
     Rgb24 = 1,
     Rgb16_555 = 2,
     Rgb16_565 = 3,
 };
 
-enum class Rotation {
+enum class Rotation : u8 {
     None = 0,
     Clockwise_90 = 1,
     Clockwise_180 = 2,
     Clockwise_270 = 3,
 };
 
-enum class BlockAlignment {
+enum class BlockAlignment : u8 {
     /// Image is output in linear format suitable for use as a framebuffer.
     Linear = 0,
     /// Image is output in tiled PICA format, suitable for use as a texture.
     Block8x8 = 1,
 };
 
-enum class StandardCoefficient {
+enum class StandardCoefficient : u8 {
     ITU_Rec601 = 0,
     ITU_Rec709 = 1,
     ITU_Rec601_Scaling = 2,
@@ -65,30 +65,37 @@ struct ConversionParameters {
     InputFormat input_format;
     OutputFormat output_format;
     Rotation rotation;
-    BlockAlignment alignment;
+    BlockAlignment block_alignment;
     u16 input_line_width;
     u16 input_lines;
+    StandardCoefficient standard_coefficient;
+    u8 reserved;
+    u16 alpha;
+};
+static_assert(sizeof(ConversionParameters) == 12, "ConversionParameters struct has incorrect size");
 
-    // Input parameters for the Y (luma) plane
-    VAddr srcY_address;
-    u32 srcY_image_size;
-    u16 srcY_transfer_unit;
-    u16 srcY_stride;
-
-    // Output parameters for the conversion results
-    VAddr dst_address;
-    u32 dst_image_size;
-    u16 dst_transfer_unit;
-    u16 dst_stride;
+struct ConversionBuffer {
+    VAddr address;
+    u32 image_size;
+    u16 transfer_unit;
+    u16 stride;
 };
 
-static ConversionParameters conversion_params;
+struct ConversionData {
+    ConversionParameters params;
+    /// Input parameters for the Y (luma) plane
+    ConversionBuffer src_Y;
+    /// Output parameters for the conversion results
+    ConversionBuffer dst;
+};
+
+static ConversionData conversion;
 
 static void SetInputFormat(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    conversion_params.input_format = static_cast<InputFormat>(cmd_buff[1]);
-    LOG_DEBUG(Service_Y2R, "called input_format=%u", conversion_params.input_format);
+    conversion.params.input_format = static_cast<InputFormat>(cmd_buff[1]);
+    LOG_DEBUG(Service_Y2R, "called input_format=%u", conversion.params.input_format);
 
     cmd_buff[1] = RESULT_SUCCESS.raw;
 }
@@ -96,8 +103,8 @@ static void SetInputFormat(Service::Interface* self) {
 static void SetOutputFormat(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    conversion_params.output_format = static_cast<OutputFormat>(cmd_buff[1]);
-    LOG_DEBUG(Service_Y2R, "called output_format=%u", conversion_params.output_format);
+    conversion.params.output_format = static_cast<OutputFormat>(cmd_buff[1]);
+    LOG_DEBUG(Service_Y2R, "called output_format=%u", conversion.params.output_format);
 
     cmd_buff[1] = RESULT_SUCCESS.raw;
 }
@@ -105,8 +112,8 @@ static void SetOutputFormat(Service::Interface* self) {
 static void SetRotation(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    conversion_params.rotation = static_cast<Rotation>(cmd_buff[1]);
-    LOG_DEBUG(Service_Y2R, "called rotation=%u", conversion_params.rotation);
+    conversion.params.rotation = static_cast<Rotation>(cmd_buff[1]);
+    LOG_DEBUG(Service_Y2R, "called rotation=%u", conversion.params.rotation);
 
     cmd_buff[1] = RESULT_SUCCESS.raw;
 }
@@ -114,18 +121,18 @@ static void SetRotation(Service::Interface* self) {
 static void SetBlockAlignment(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    conversion_params.alignment = static_cast<BlockAlignment>(cmd_buff[1]);
-    LOG_DEBUG(Service_Y2R, "called alignment=%u", conversion_params.alignment);
+    conversion.params.block_alignment = static_cast<BlockAlignment>(cmd_buff[1]);
+    LOG_DEBUG(Service_Y2R, "called alignment=%u", conversion.params.block_alignment);
 
     cmd_buff[1] = RESULT_SUCCESS.raw;
 }
 
 /**
-* Y2R_U::GetTransferEndEvent service function
-*  Outputs:
-*      1 : Result of function, 0 on success, otherwise error code
-*      3 : The handle of the completion event
-*/
+ * Y2R_U::GetTransferEndEvent service function
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      3 : The handle of the completion event
+ */
 static void GetTransferEndEvent(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
@@ -137,14 +144,14 @@ static void GetTransferEndEvent(Service::Interface* self) {
 static void SetSendingY(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    conversion_params.srcY_address = cmd_buff[1];
-    conversion_params.srcY_image_size = cmd_buff[2];
-    conversion_params.srcY_transfer_unit = cmd_buff[3];
-    conversion_params.srcY_stride = cmd_buff[4];
+    conversion.src_Y.address = cmd_buff[1];
+    conversion.src_Y.image_size = cmd_buff[2];
+    conversion.src_Y.transfer_unit = cmd_buff[3];
+    conversion.src_Y.stride = cmd_buff[4];
     u32 src_process_handle = cmd_buff[6];
     LOG_DEBUG(Service_Y2R, "called image_size=0x%08X, transfer_unit=%hu, transfer_stride=%hu, "
-        "src_process_handle=0x%08X", conversion_params.srcY_image_size,
-        conversion_params.srcY_transfer_unit, conversion_params.srcY_stride, src_process_handle);
+        "src_process_handle=0x%08X", conversion.src_Y.image_size,
+        conversion.src_Y.transfer_unit, conversion.src_Y.stride, src_process_handle);
 
     cmd_buff[1] = RESULT_SUCCESS.raw;
 }
@@ -152,14 +159,14 @@ static void SetSendingY(Service::Interface* self) {
 static void SetReceiving(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    conversion_params.dst_address = cmd_buff[1];
-    conversion_params.dst_image_size = cmd_buff[2];
-    conversion_params.dst_transfer_unit = cmd_buff[3];
-    conversion_params.dst_stride = cmd_buff[4];
+    conversion.dst.address = cmd_buff[1];
+    conversion.dst.image_size = cmd_buff[2];
+    conversion.dst.transfer_unit = cmd_buff[3];
+    conversion.dst.stride = cmd_buff[4];
     u32 dst_process_handle = cmd_buff[6];
     LOG_DEBUG(Service_Y2R, "called image_size=0x%08X, transfer_unit=%hu, transfer_stride=%hu, "
-        "dst_process_handle=0x%08X", conversion_params.dst_image_size,
-        conversion_params.dst_transfer_unit, conversion_params.dst_stride,
+        "dst_process_handle=0x%08X", conversion.dst.image_size,
+        conversion.dst.transfer_unit, conversion.dst.stride,
         dst_process_handle);
 
     cmd_buff[1] = RESULT_SUCCESS.raw;
@@ -168,8 +175,8 @@ static void SetReceiving(Service::Interface* self) {
 static void SetInputLineWidth(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    conversion_params.input_line_width = cmd_buff[1];
-    LOG_DEBUG(Service_Y2R, "input_line_width=%u", conversion_params.input_line_width);
+    conversion.params.input_line_width = cmd_buff[1];
+    LOG_DEBUG(Service_Y2R, "input_line_width=%u", conversion.params.input_line_width);
 
     cmd_buff[1] = RESULT_SUCCESS.raw;
 }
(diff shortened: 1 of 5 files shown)


src/core/hw/y2r.cpp
```diff
@@ -0,0 +1,379 @@
...
+    switch (cvt.block_alignment) {
+    case BlockAlignment::Linear:
+        tile_remap = linear_lut; break;
+    case BlockAlignment::Block8x8:
+        tile_remap = morton_lut; break;
+    }
+
+    for (unsigned int y = 0; y < cvt.input_lines; y += 8) {
+        unsigned int row_height = cvt.input_lines - y;
+        if (row_height >= 8) {
```
archshift: `row_height = std::min(row_height, 8)`

src/core/hw/y2r.cpp
```diff
@@ -0,0 +1,377 @@
...
+static void WriteTileToOutput(u32* output, const ImageTile& tile, int height, int line_stride) {
+    for (int y = 0; y < height; ++y) {
+        for (int x = 0; x < 8; ++x) {
+            output[y * line_stride + x] = tile[y * 8 + x];
+        }
+    }
+}
+
+/**
+ * Performs a Y2R colorspace conversion.
```
purpasmart96: Is this wall of text really necessary? Maybe you could summarize it?

archshift: I think it's fine to have this here. It contains a lot of relevant information for developers hacking on y2r, and it's pretty brief IMO.

yuriks: I didn't really know where to put it either. In the end I figured this is where people would be most likely to look and need it.

src/core/hw/y2r.cpp
```diff
@@ -0,0 +1,377 @@
...
+namespace HW {
+namespace Y2R {
+
+using namespace Y2R_U;
+
+static const size_t MAX_TILES = 1024 / 8;
+static const size_t TILE_SIZE = 8 * 8;
+using ImageTile = std::array<u32, TILE_SIZE>;
+
+/// Converts a image strip from the source YUV format into individual 8x8 RGB32 tiles.
```
turbedi: *an
~nitpick~

src/core/hw/y2r.cpp
lioncash: I think it's quite fine. It's documentation which is always useful.

src/core/hle/service/y2r_u.cpp
```diff
@@ -20,75 +20,73 @@
...
-static ConversionParameters conversion_params;
+ResultCode ConversionConfiguration::SetInputLineWidth(u16 width) {
+    if (width == 0 || width > 1024 || width % 8 != 0) {
+        return ResultCode(ErrorDescription::OutOfRange, ErrorModule::CAM,
+            ErrorSummary::InvalidArgument, ErrorLevel::Usage); // 0xE0E053FD
+    }
+
+    // Note: The hardware uses the register value 0 to represent a width of 1024, so for a width of
+    // 1024 the `camera` module would set the value 0 here, but we don't need to emulate this
+    // internal detail.
```
bunnei: Why don't you need to emulate this? Is the error on `width == 0` correct then if that really means 1024?

src/core/hle/service/y2r_u.cpp
```diff
@@ -114,18 +112,26 @@ static void SetRotation(Service::Interface* self) {
...
+    conversion.block_alignment = static_cast<BlockAlignment>(cmd_buff[1]);
+    LOG_DEBUG(Service_Y2R, "called alignment=%hhu", conversion.block_alignment);
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+}
+
+static void SetTransferEndInterrupt(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
 
+    cmd_buff[0] = 0x000D0040;
```
bunnei: Is there a reason for this?

src/core/hle/service/y2r_u.cpp
```diff
@@ -114,18 +112,26 @@ static void SetRotation(Service::Interface* self) {
...
     cmd_buff[1] = RESULT_SUCCESS.raw;
+    LOG_DEBUG(Service_Y2R, "called (STUBBED)");
```
bunnei: Should be `LOG_WARNING(Service_Y2R, "(STUBBED) called` (to be consistent with similar logs in other services).

src/core/hle/service/y2r_u.cpp
```diff
@@ -137,138 +143,132 @@ static void GetTransferEndEvent(Service::Interface* self) {
...
+}
 
-        size_t srcY_offset = 0;
-        size_t dst_tile_line_offs = 0;
+static void SetInputLineWidth(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
 
-        const size_t tile_size = 8 * 8 * bpp;
+    cmd_buff[1] = conversion.SetInputLineWidth(cmd_buff[1]).raw;
+    LOG_DEBUG(Service_Y2R, "called input_line_width=%u", cmd_buff[1]);
```
bunnei: This is going to print a result code, not the width (which is probably what you want to log)

src/core/hle/service/y2r_u.cpp
```diff
@@ -137,138 +143,132 @@ static void GetTransferEndEvent(Service::Interface* self) {
...
-        for (size_t line = 0; line < input_lines;) {
-            size_t max_line = line + 8;
+static void SetInputLines(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
 
-            for (; line < max_line; ++line) {
-                for (size_t x = 0; x < input_line_width; ++x) {
-                    size_t tile_x = x / 8;
+    cmd_buff[1] = conversion.SetInputLines(cmd_buff[1]).raw;
+    LOG_DEBUG(Service_Y2R, "called input_line_number=%u", cmd_buff[1]);
```
bunnei: Ditto

src/core/hle/service/y2r_u.cpp
```diff
@@ -276,12 +276,20 @@ static void StartConversion(Service::Interface* self) {
     cmd_buff[1] = RESULT_SUCCESS.raw;
 }
 
+static void StopConversion(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[0] = 0x00270040;
```
bunnei: Is this necessary? Curious if there is a reason why you set some but not others.

src/core/hle/service/y2r_u.cpp
```diff
@@ -290,6 +298,40 @@ static void IsBusyConversion(Service::Interface* self) {
...
+        params->reserved, params->alpha);
+
+    ResultCode result = RESULT_SUCCESS;
+
+    conversion.input_format = params->input_format;
+    conversion.output_format = params->output_format;
+    conversion.rotation = params->rotation;
+    conversion.block_alignment = params->block_alignment;
+    result = conversion.SetInputLineWidth(params->input_line_width);
+    if (result.IsError()) goto error;
```
bunnei: But but but... I thought we said "absolutely no goto blah blah blah"

src/core/hle/service/y2r_u.cpp
```diff
@@ -290,6 +298,40 @@ static void IsBusyConversion(Service::Interface* self) {
...
+    auto params = reinterpret_cast<const ConversionParameters*>(&cmd_buff[1]);
+    LOG_DEBUG(Service_Y2R,
+        "called input_format=%hhu output_format=%hhu rotation=%hhu block_alignment=%hhu "
+        "input_line_width=%hu input_lines=%hu standard_coefficient=%hhu "
+        "reserved=%hhu alpha=%hX",
+        params->input_format, params->output_format, params->rotation, params->block_alignment,
+        params->input_line_width, params->input_lines, params->standard_coefficient,
+        params->reserved, params->alpha);
+
+    ResultCode result = RESULT_SUCCESS;
```
bunnei: You never actually set `cmd_buff[1]` to success..

src/core/hle/service/y2r_u.h
```diff
@@ -11,6 +15,98 @@
...
+    YUV422_Indiv16 = 2,
+    /// 16-bit input (only LSB used), with YUV components in separate planes and 4:2:0 subsampling.
+    YUV420_Indiv16 = 3,
+
+    /// 8-bit input, with a single interleaved stream in YUYV format and 4:2:2 subsampling.
+    YUYV422_Interleaved = 4,
+};
+
+enum class OutputFormat : u8 {
+    Rgb32 = 0,
```
bunnei: Is there are a reason why these are named differently than, say, https://github.com/citra-emu/citra/blob/9adb2dfb2ed49900c894a445145fe616e3426b6f/src/core/hw/gpu.h#L53 ? (e.g. RGBA8, RGB, RGB5A1, RGB565)

src/core/hle/service/y2r_u.h
```diff
@@ -11,6 +15,98 @@
...
+ */
+using CoefficientSet = std::array<s16, 8>;
+
+struct ConversionBuffer {
+    /// Current reading/writing address of this buffer.
+    VAddr address;
+    /// Remaining amount of bytes to be DMAed, does not include the inter-trasfer gap.
+    u32 image_size;
+    /// Size of a single DMA transfer.
+    u16 transfer_unit;
```
bunnei: Is this really `Size of a single DMA transfer.` ? The name `transfer_unit` does not imply that at all to me.

src/core/hle/service/y2r_u.h
```diff
@@ -11,6 +15,98 @@
...
+ * B = trunc((c_0 * Y + c_4 * U          ) + c_7 + 0.75)
+ * ```
+ */
+using CoefficientSet = std::array<s16, 8>;
+
+struct ConversionBuffer {
+    /// Current reading/writing address of this buffer.
+    VAddr address;
+    /// Remaining amount of bytes to be DMAed, does not include the inter-trasfer gap.
+    u32 image_size;
```
bunnei: Why is this called `image_size` if it's the "remaining number of bytes..." ?

src/core/hw/y2r.cpp
```diff
@@ -0,0 +1,377 @@
...
+        output += output_unit;
+        input += buf.transfer_unit + buf.gap;
+
+        buf.address += buf.transfer_unit + buf.gap;
+        buf.image_size -= buf.transfer_unit;
+        amount_of_data -= output_unit;
+    }
+}
+
+/// Convert intermediate RGB32 format to the final output format while simulating a outgoing CDMA transfer.
```
bunnei: *an outgoing CDMA transfer.

src/core/hw/y2r.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    while (amount_of_data > 0) {
+        u8* unit_end = output + buf.transfer_unit;
+        while (output < unit_end) {
+            u32 color = *input++;
+            u32 r = (color >> 24) & 0xFF;
+            u32 g = (color >> 16) & 0xFF;
+            u32 b = (color >>  8) & 0xFF;
+
+            switch (output_format) {
```
bunnei: Can we have these conversion functions use those in https://github.com/citra-emu/citra/blob/9adb2dfb2ed49900c894a445145fe616e3426b6f/src/common/color.h instead of duplicating code?

src/core/hle/service/y2r_u.cpp
yuriks: There's two separate things:
The hardware register: width field is 10-bit, allows values 0-1023. 0 is interpreted as a width of 1024.
The y2r interface: Accepts values 1-1024, does not allow 0. When it receives 1024 is sets the hardware register to 0 accordingly.

It doesn't need to be emulated since I haven't bothered to use the register-level interface for y2r, I just pass the HLE struct to the conversion function. (This is mostly because the CDMA mechanism is mostly unknown and completely unimplemented in citra, otherwise I'd have made it more low-level.)

src/core/hw/y2r.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+/// Converts a image strip from the source YUV format into individual 8x8 RGB32 tiles.
+static void ConvertYUVToRGB(InputFormat input_format,
+        const u8* input_Y, const u8* input_U, const u8* input_V, ImageTile output[],
+        unsigned int width, unsigned int height, const CoefficientSet& coefficients) {
+
+    for (unsigned int y = 0; y < height; ++y) {
+        for (unsigned int x = 0; x < width; ++x) {
+            s32 Y, U, V;
+            switch (input_format) {
```
bunnei: Might be nice to have these color conversion functions in https://github.com/citra-emu/citra/blob/9adb2dfb2ed49900c894a445145fe616e3426b6f/src/common/color.h

src/core/hle/service/y2r_u.cpp
yuriks: Needed for real IPC. Most of our calls lack this, which is going to be a *major pain in the butt* later on when we emulate IPC more accurately. So I add them when I remember/it's convenient.

yuriks: Though I should probably add some macros for this instead of using the hex every time. Should be a easy PR.

src/core/hw/y2r.cpp
```diff
@@ -0,0 +1,377 @@
...
+        for (int x = 0; x < 8; ++x) {
+            output[y * line_stride + x] = tile[y * 8 + x];
+        }
+    }
+}
+
+/**
+ * Performs a Y2R colorspace conversion.
+ *
+ * The Y2R hardware implements hardware-accelerated YUV to RGB colorspace conversions. It is mostly
```
bunnei: *most commonly used for

src/core/hle/service/y2r_u.cpp
yuriks: ditto. Please point it out if I missed it in any newly added calls.

src/core/hle/service/y2r_u.cpp
yuriks: *I* don't remember saying that. :) Anyway, if you can suggest a cleaner way of doing it I'll take it.

src/core/hle/service/y2r_u.h
yuriks: Just following the naming standard used by most enums, but I just realized it's not even internally consistent with this PR. Gonna change it.

src/core/hw/y2r.cpp
```diff
@@ -0,0 +1,377 @@
...
+ * - The YUV tuple is converted, using fixed point calculations, to RGB. This step can be configured
+ *   using a set of coefficients to support different colorspace standards. See `CoefficientSet`.
+ * - The strip can be optionally rotated 90, 180 or 270 degrees. Since each strip is processed
+ *   independently, this notably rotates each *strip*, not the entire image. This means that for 90
+ *   or 270 degree rotations, the output will be in terms of several 8 x height images, and for any
+ *   non-zero rotation the strips will have to be re-arranged so that the parts of the image will
+ *   not be shuffled together. This limitation makes this a feature of somewhat dubious utility. 90
+ *   or 270 degree rotations in images with non-even height don't seem to work properly.
+ * - The data is converted to the output RGB format. See the `OutputFormat` enum.
+ * - The data can be output either linearly row-by-row or in the swizzled 8x8 tile format used by
```
bunnei: I'm assuming you mean "pixel row-by-row" and not "strip row-by-row", since you use just "row" in other parts of this description that I assume to mean just the strip. Please clarify when you mean "pixel row" and "current strip" throughout.

src/core/hle/service/y2r_u.h
yuriks: The CDMA engine will do several transfers in one go. It'll transfer `transfer_unit` bytes, then advance the pointer by `gap`. It'll do this until `image_size` bytes have been copied. I guess `image_size` is an easy one to rename, can you suggest a better name for this though?

src/core/hle/service/y2r_u.h
yuriks: IIRC this name came from BPB symbols. I'm welcome to renaming it.

src/core/hw/y2r.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+            using MathUtil::Clamp;
+            *out = ((u32)Clamp(r >> 5, 0, 0xFF) << 24) |
+                   ((u32)Clamp(g >> 5, 0, 0xFF) << 16) |
+                   ((u32)Clamp(b >> 5, 0, 0xFF) << 8);
+        }
+    }
+}
+
+/// Simulates an incoming CDMA transfer. The N parameter is used to automatically convert 16-bit formats to 8-bit.
```
bunnei: What is CDMA and how does it differ from a normal DMA? Don't think you actually define this anywhere. I thought perhaps it was a term I should have known, but a quick Google search didn't yield anything specifically for "CDMA" with regards to direct memory access, so it's probably worth documenting in your docstring(s) somewhere.

src/core/hw/y2r.cpp
yuriks: I'd have to write them out to an array before reading them back, which seemed like a very akward thing to do at the time. I guess I can just turn the `r`,`g`,`b` variables into an array and it's not so bad.

src/core/hw/y2r.cpp
yuriks: What conversion functions? The YUV component reading is closely tied to the input format. In any case, I don't think any of this code will come up elsewhere in the codebase.

src/core/hw/y2r.cpp
yuriks: CDMA is the ARM11 DMA engine: http://3dbrew.org/wiki/Corelink_DMA_Engines

bunnei: I still think it's worth mentioning that in a parenthesis somewhere if it doesn't already exist in the source base

src/core/hw/y2r.cpp
bunnei: I'd rather reuse the code that we already have that has been verified to be correct. At one point we had like 3-4 implementations of these conversions in our source. I'm sure your impl. is correct, but in general (if possible) I'd rather not encourage the duplication.

src/core/hle/service/y2r_u.h
bunnei: `transfer_unit_size`? When something is just called "unit" I think of some functioning component

src/core/hw/y2r.cpp
yuriks: Nevermind, I was on drugs. Yeah, this is an obvious win, will do.

src/core/hle/service/y2r_u.cpp
jroweboy: This isn't much cleaner, but another way could be to use `do while false`

    do {
        result = conversion.SetInputLineWidth(params->input_line_width);
        if (result.IsError()) break;
        ...
    } while (0);

src/core/hw/y2r.cpp
bunnei: I assumed these were just standard format conversions... If what you said is the case, I guess that's fine. Probably the only other use for these (which is really a stretch) would be some sort of debug feature (like dumping the frames or showing them in a debugger widget).

src/core/hle/service/y2r_u.cpp
yuriks: Yes I do, at the end of the function.

src/core/hle/service/y2r_u.h
yuriks: > unit: an individual thing or person regarded as single and complete but which can also form an individual component of a larger or more complex whole

"Transfer unit", as in one piece of the whole transfer operation. I don't think `transfer_unit_size` clears that up at all.

src/core/hw/y2r.cpp
yuriks: I added a link to the 3dbrew page in the wall of text.

yuriks:
Updated.

bunnei:
LGTM. Ship it.

yuriks:
Rebased

src/core/hle/service/y2r_u.cpp
```diff
@@ -20,75 +20,73 @@
...
+    }
+    return RESULT_SUCCESS;
+}
+
+ResultCode ConversionConfiguration::SetStandardCoefficient(StandardCoefficient standard_coefficient) {
+    size_t index = static_cast<size_t>(standard_coefficient);
+    if (index < 4) {
+        std::memcpy(coefficients.data(), standard_coefficients[index].data(), sizeof(coefficients));
+        return RESULT_SUCCESS;
+    } else {
```
linkmauve: Unneeded else here, but I’d inverse the branches to return early on error, and have the expected branch outside of any if.

archshift:
LGTM

PR #861, opened 2015-06-14
video_core: add extra braces around initializer

yuriks:
Trivial change and fixes several warnings in the clang build.


lioncash:
LGTM

PR #863, opened 2015-06-15
Create a second Windows archive with only the binaries

linkmauve:
This will help people who don’t have much bandwidth to download only what they want, and not the Qt dll again and again.

It is put in an hidden directory by default, in order to not confuse normal users who do need the Qt dll.


archshift:
I don't think confusing the user is a problem, as normal users probably wouldn't browse the build directories like that. I'd prefer to keep it a normal, non-hidden folder.

Other than that, LGTM.

linkmauve:
On Mon, Jun 15, 2015 at 10:46:38AM -0700, archshift wrote:
> I don't think confusing the user is a problem, as normal users probably wouldn't browse the build directories like that. I'd prefer to keep it a normal, non-hidden folder.

Ok, I removed the dot in front of the directory.

> 
> Other than that, LGTM.

PR #864, opened 2015-06-15
Log the GL driver’s vendor and renderer

linkmauve:
This helps differentiate the various Mesa drivers, especially hardware and software ones, or in multi-GPU setups.


archshift:
LGTM

Issue #373, opened 2014-12-30
Drop EMU_PLATFORM usage


neobrain:
It provides virtually no advantage over checking for the platform-specific preprocessor definitions, maybe other than being consistent, which is kind of useless because it trades that consistency for being de facto standardized.

I like having few project-specific oddities, so I suggest dropping and replacing it with the corresponding preprocessor definitions.

lioncash:
This is likely fixed now by @linkmauve's cleanups

yuriks:
Yep, it is. "Syntax" of the commit message was wrong, so it didn't auto-close.

PR #876, opened 2015-06-21
Cleanup includes, mostly in common

linkmauve:
Using [include-what-you-use](https://code.google.com/p/include-what-you-use/) to drive the cleanup, I removed unused includes, added some that were required but already included somewhere else, converted a few includes in headers into class/struct declaration, etc.

I limited the scope of my changes to common, citra, citra-qt, file_sys and the root of core, as it already felt like a pretty big PR.

0a0a2c7ac98cee0e107329bff4e477d6f6687b28 also fixes issue #373.


src/citra/emu_window/emu_window_glfw.cpp
```diff
@@ -2,13 +2,23 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
+#include <cstdlib>
+#include <string>
+
+// Here we assume one of the GL headers has been included by GLFW, for GL_TRUE to be defined.
```
neobrain: What?

linkmauve: GLFW will almost always include one of GL/gl.h, GLES2/gl2.h, etc., depending on the system we are on and the GLFW_INCLUDE_* defines that happened before its inclusion.

The only case it doesn’t is when GLFW_INCLUDE_NONE is defined prior to its inclusion, in which case it assumes a specific OpenGL header has already been included, so we can assume that too.

src/common/profiler.cpp
```diff
@@ -2,15 +2,20 @@
...
 
-#include "common/profiler.h"
-#include "common/profiler_reporting.h"
-#include "common/assert.h"
-
 #if defined(_MSC_VER) && _MSC_VER <= 1800 // MSVC 2013.
-#define WIN32_LEAN_AND_MEAN
-#include <Windows.h> // For QueryPerformanceCounter/Frequency
+    #define WIN32_LEAN_AND_MEAN
+    #include <Windows.h> // For QueryPerformanceCounter/Frequency
```
yuriks: Keep the Windows.h include at the end, since Windows.h defines a *lot* of macros which can pretty easily cause compilation errors in unrelated areas of the codebase.

yuriks:
Thank you thank you thank you thank you for doing this. I've wanted to run YWYU on the codebase for a while but couldn't get it to compile on Arch.

src/common/profiler.cpp
linkmauve: Fixed, thanks.

src/citra/emu_window/emu_window_glfw.cpp
neobrain: My point was that this is documented in a really confusing way. It's not clear how an `#include` can assume anything, or how it is assumed. It just states the mere dependency without mentioning any reasoning for it. Could you clarify this?

linkmauve: Actually, it makes more sense to include our own OpenGL wrapper than to let GLFW include the system one.

In this case it doesn’t change anything though, since there is no actual GL call being made.

Subv:
LGTM

PR #877, opened 2015-06-22
Add helpers to create IPC command buffer headers and descriptors

yuriks:
This should hopefully encourage people to add these (without using magic numbers) to service returns in the future.

### DIFF
--- a/src/core/hle/kernel/session.h
+++ b/src/core/hle/kernel/session.h
@@ -8,6 +8,40 @@
 #include "core/hle/kernel/thread.h"
 #include "core/memory.h"
 
+namespace IPC {
+
+inline u32 MakeHeader(u16 command_id, unsigned int regular_params, unsigned int translate_params) {
+    return ((u32)command_id << 16) | (((u32)regular_params & 0x3F) << 6) | (((u32)translate_params & 0x3F) << 0);
+}
+
+inline u32 MoveHandleDesc(unsigned int num_handles = 1) {
+    return 0x0 | ((num_handles - 1) << 26);
+}
+
+inline u32 CopyHandleDesc(unsigned int num_handles = 1) {
+    return 0x10 | ((num_handles - 1) << 26);
+}
+
+inline u32 CallingPidDesc() {
+    return 0x20;
+}
+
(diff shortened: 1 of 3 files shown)


yuriks:
I was additionally thinking of moving `GetCommandBuffer` to the `Ipc` namespace too, but that would make ripple changes through all the service functions, and I'm not sure what people think of it.

Subv:
LGTM

archshift:
LGTM

src/core/hle/kernel/session.h
```diff
@@ -8,6 +8,40 @@
 #include "core/hle/kernel/thread.h"
 #include "core/memory.h"
 
+namespace Ipc {
```
neobrain: *IPC ?

neobrain:
LGTM other than my comment.

src/core/hle/kernel/session.h
```diff
@@ -8,6 +8,40 @@
...
+}
+
+inline u32 StaticBufferDesc(u32 size, unsigned int buffer_id) {
+    return 0x2 | (size << 14) | ((buffer_id & 0xF) << 10);
+}
+
+enum MappedBufferPermissions {
+    R = 2,
+    W = 4,
+    RW = 6,
```
linkmauve: Replace 6 with R | W, which describes better what it is.

linkmauve:
Other than that, LGTM.

Subv:
@mailwl I think it's better to be explicit in this case, lest the person using it just forgets about it

yuriks:
Yeah, I did consider default parameters, but it's indeed better to make it explicit.

linkmauve:
Btw, I’ve been thinking about making every service call return a std::pair with the header and the return code, and to have the wrapper do the actual assignment to cmd_buffer[0] and cmd_buffer[1], how practical would that be?

yuriks:
I think that should be done using a wrapper for reading/writing to the command buffer, not by tacking it onto the IPC system, since, for example, the return code is just a convention used by most of the calls, it's not a requirement of the IPC system. Doing that would just further distance the emulation from how the console works.

yuriks:
Updated

PR #881, opened 2015-06-23
Switch argument parsing to use getopt

gwicks:


archshift:
`getopt` needs to definitely go in `externals` and not `src/citra/`.

src/citra/citra.cpp
```diff
@@ -20,12 +20,42 @@
...
+            optind++;
+        }
+    }
+
+
     Log::Filter log_filter(Log::Level::Debug);
     Log::SetFilter(&log_filter);
 
-    if (argc < 2) {
+    if (boot_filename == "") {
```
archshift: `if (boot_filename.empty())`

src/citra/citra.cpp
```diff
@@ -33,7 +63,7 @@ int main(int argc, char **argv) {
     Config config;
     log_filter.ParseFilterString(Settings::values.log_filter);
 
-    std::string boot_filename = argv[1];
+
```
archshift: Redundant blank line.

archshift:
Note that we use brackets on the same line as the statement that precede them. See the style guide for details.

archshift:
Because POSIX already has getopt, you'll need to use an `#ifdef` for MSVC (note: MSVC is not the same thing as Windows), including `"getopt.h"` for MSVC and `"unistd.h"` for other platforms.

src/citra/citra.cpp
```diff
@@ -20,12 +20,42 @@
 
 #include "video_core/video_core.h"
 
+#include "getopt.h"
```
neobrain: include library headers before citra-internal headers.

src/citra/citra.cpp
```diff
@@ -20,12 +20,42 @@
...
+
+
+void printHelp()
```
neobrain: PrintHelp

src/citra/citra.cpp
```diff
@@ -20,12 +20,42 @@
...
+    std::string boot_filename = "";
+
+    char arg;
+    while (optind < argc)
+    {
+        if ((arg = getopt(argc, argv, ":h")) != -1)
+        {
+            switch (arg)
+            {
+            case 'h':
```
neobrain: Is there a good way to also support long hand versions of these, i.e. `--help`?

src/citra/citra.cpp
```diff
@@ -20,12 +20,42 @@
 
 #include "video_core/video_core.h"
 
+#include "getopt.h"
+
+
+void printHelp()
+{
+    printf("Usage: citra <filname> \n");
```
neobrain: trailing whitespace.

neobrain: Is there a better name to print the application name without hardcoding it? (e.g. I suppose on Windows it would be better to print citra.exe, or whatever the application binary is actually called)

src/citra/citra.cpp
```diff
@@ -20,12 +20,42 @@
...
+            switch (arg)
+            {
+            case 'h':
+                printHelp();
+                break;
+            }
+        }
+        else
+        {
+            boot_filename = argv[optind];
```
neobrain: If more than one such argument is given, the help should probably be printed and program execution be aborted.

src/citra/getopt.c
```diff
@@ -0,0 +1,961 @@
+/* Getopt for Microsoft C
```
neobrain: I'd rather have the system-widely installed version of this library be used if it's available.

src/citra/citra.cpp
archshift: Yes, that's where he'd have to use `getopt_long()`.

src/citra/citra.cpp
```diff
@@ -20,12 +26,39 @@
...
 int main(int argc, char **argv) {
+    std::string boot_filename = "";
+
+    char arg;
+    while (optind < argc)
+    {
+        if ((arg = getopt(argc, argv, ":h")) != -1)
+        {
+            switch (arg)
+            {
```
archshift: Bracket on same line please.

gwicks: Ah, sorry about that, will fix that

src/citra/citra.cpp
```diff
@@ -20,12 +37,38 @@
 
 #include "video_core/video_core.h"
 
+
+
+
+void PrintHelp()
```
neobrain: static

src/citra/citra.cpp
```diff
@@ -4,6 +4,23 @@
 
 #include <thread>
 
+#ifdef _MSC_VER
+#include "getopt/getopt.h"
+extern char *optarg;
```
neobrain: Shouldn't this stuff be extern C?

src/citra/citra.cpp
```diff
@@ -20,12 +37,38 @@
 
 #include "video_core/video_core.h"
 
+
+
+
+void PrintHelp()
+{
+    printf("Usage: citra <filname>\n");
```
neobrain: Please use std::cout instead of printf.

src/citra/citra.cpp
```diff
@@ -20,12 +37,38 @@
...
+    std::string boot_filename = "";
+    static struct option long_options[] = { { "help", no_argument, 0, 'h' }, { 0, 0, 0, 0 } };
+
+    char arg;
+    while (optind < argc) {
+        if ((arg = getopt_long(argc, argv, ":h",long_options,&option_index)) != -1) {
+            switch (arg) {
+            case 'h':
+                PrintHelp();
+                exit(1);
```
neobrain: Is there any reason to actually use exit(1) instead of return 1 here?

src/citra/citra.cpp
```diff
@@ -20,12 +37,38 @@
...
+    printf("Usage: citra <filname>\n");
+}
+
 /// Application entry point
 int main(int argc, char **argv) {
+    int option_index = 0;
+    std::string boot_filename = "";
+    static struct option long_options[] = { { "help", no_argument, 0, 'h' }, { 0, 0, 0, 0 } };
+
+    char arg;
```
neobrain: This declaration should be moved into the while loop.

src/citra/citra.cpp
```diff
@@ -20,12 +37,38 @@
...
+
+void PrintHelp()
+{
+    printf("Usage: citra <filname>\n");
+}
+
 /// Application entry point
 int main(int argc, char **argv) {
+    int option_index = 0;
+    std::string boot_filename = "";
```
neobrain: There's no need to initialize string objects with an empty string

src/citra/citra.cpp
neobrain: Actually, this should return 0, since no failure happened.

mailwl:
May I ask silly question?
Why not `boost::program_options`?

neobrain:
@mailwl I'll happily accept any PR which adds BPO to our ext-boost repository (along with Windows binaries which are suitable for use in Citra).

src/citra/citra.cpp
gwicks: Not sure, but it didn't link properly unless I inclded those declarations

gwicks:
@mailwl The issue with using BPO is that the binaries that they distribute have RTTI enabled, which conflicts with the settings for Release, and it requires adding a significant number of extra dependencies.

CMakeLists.txt
```diff
@@ -173,6 +173,8 @@ if (ENABLE_QT)
     endif()
 endif()
 
+
+
```
archshift: Random newlines?

src/citra/citra.cpp
```diff
@@ -20,12 +38,37 @@
...
+static void PrintHelp()
+{
+    std::cout << "Usage: citra <filname>" << std::endl;
+}
+
 /// Application entry point
 int main(int argc, char **argv) {
+    int option_index = 0;
+    std::string boot_filename;
+    static struct option long_options[] = { { "help", no_argument, 0, 'h' }, { 0, 0, 0, 0 } };
```
archshift: Can you put the individual options each on new lines? Like so:
```cpp
static struct option long_options[] = {
    { "help", no_argument, 0, 'h' },
    { 0, 0, 0, 0 },
};
```

src/citra/citra.cpp
```diff
@@ -20,12 +38,37 @@
...
+
+    while (optind < argc) {
+        char arg;
+        if ((arg = getopt_long(argc, argv, ":h",long_options,&option_index)) != -1) {
```
archshift: Spaces after each argument

archshift:
LGTM once you fix those little nitpicks!

gwicks:
Made the tweaks, should be good to go

src/citra/CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ create_directory_groups(${SRCS} ${HEADERS})
 add_executable(citra ${SRCS} ${HEADERS})
 target_link_libraries(citra core common video_core)
 target_link_libraries(citra ${GLFW_LIBRARIES} ${OPENGL_gl_LIBRARY} inih)
+if (MSVC)
+target_link_libraries(citra getopt)
```
neobrain: indentation

src/citra/citra.cpp
```diff
@@ -20,12 +38,40 @@
 
 #include "video_core/video_core.h"
 
+
```
neobrain: two newlines are more than enough

src/citra/citra.cpp
```diff
@@ -20,12 +38,40 @@
...
+
+static void PrintHelp()
+{
+    std::cout << "Usage: citra <filname>" << std::endl;
```
neobrain: filename

src/citra/citra.cpp
```diff
@@ -3,6 +3,24 @@
 // Refer to the license.txt file included.
 
 #include <thread>
+#include <iostream>
+#ifdef _MSC_VER
```
neobrain: Insert a newline here to make clear that what follows is stuff dealing with getopt..

src/citra/citra.cpp
```diff
@@ -3,6 +3,24 @@
...
+#include "getopt/getopt.h"
+extern char *optarg;
```
neobrain: Actually, none of these `extern` declarations should be necessary, since all of that is defined in getopt.h already, and even then it should all be wrapped around an `extern "C" ` block... Are you sure getopt.c is actually being compiled as C code?

src/citra/citra.cpp
```diff
@@ -20,12 +38,40 @@
...
 int main(int argc, char **argv) {
+    int option_index = 0;
+    std::string boot_filename;
+    static struct option long_options[] = {
+        { "help", no_argument, 0, 'h' },
+        { 0, 0, 0, 0 }
+    };
+
+    while (optind < argc) {
+        char arg;
```
neobrain: Please move the getopt_long call into this line. That should make it a lot easier to read.

src/citra/citra.cpp
gwicks: I agree they shouldn't be necessary, I just couldn't get it to work without them. Maybe it's a result of how getopt itself is written?

neobrain: Probably need. Either way, this should definitely be cleared up before being merged...

Have you tried looking at the build log or the generated project file settings to check whether getopt.c indeed is compiled as a C target rather than a C++ one?

gwicks:
Done and done

src/citra/citra.cpp
gwicks: Yeah, getopt.c is being compiled as C code for sure, according to C/C++->Advanced

src/citra/citra.cpp
```diff
@@ -20,12 +39,39 @@
...
+    while (optind < argc) {
+        char arg = getopt_long(argc, argv, ":h", long_options, &option_index);
+        if (arg != -1) {
+            switch (arg) {
+            case 'h':
+                PrintHelp();
+                return 0;
+            }
+        } else {
+            boot_filename = argv[optind];
```
archshift: What happens if `citra` is supplied no arguments?

archshift:
Pretty much LGTM

src/citra/citra.cpp
gwicks: Nothing happens, just complains about not being able to load a rom

neobrain:
LGTM other than the `extern` mess in main.cpp, which shouldn't be necessary, since all of those are done in getopt.h already. If it can't be helped, we should at least figure out *why* it's required.

gwicks:
My thought on the extern mess is that it may have to do with how its declared in getopt.h, possibly those _GETOPT_API declarations in the extern lines are messing with it. Honestly, it's a bit beyond my abilities to figure out why exactly it's necessary.

src/citra/citra.cpp
Syphurith: @gwicks Or else it could be designed to display the help info for command line.

externals/getopt/getopt/getopt.h
```diff
@@ -0,0 +1,136 @@
...
+#elif defined(STATIC_GETOPT)
+#pragma message("Warning static builds of getopt violate the Lesser GNU Public License")
+#define _GETOPT_API
+#elif defined(EXPORTS_GETOPT)
+#pragma message("Exporting getopt library")
+#define _GETOPT_API __declspec(dllexport)
+#else
+#pragma message("Importing getopt library")
+#define _GETOPT_API __declspec(dllimport)
+#endif
```
Syphurith: @gwicks Citra build try to link against `getopt` parts in static way. However the default option is an export way with DLL. You may add the macro `STATIC_GETOPT` to see if there is defintion now (if you keep the extern definitions for ANSI, it would be re-definition)

bunnei:
Changes look mostly fine to me... Please squash your commits, though!

gwicks:
Not sure how I messed that up, tried squashing my commits 12 commits into the most recent one.

gwicks:
Think it's good to go, dunno why AppVeyor failed, seemed to mess up when cloining in nihistro

chinhodado:
Your commits don't seem to be squashed, though.

gwicks:
Hmm, it worked on my local repo, I might have to delete this PR, recreate my fork, and open a new one with the full changes in one commit.

chinhodado:
Just do a force push (`git push -f`) and you'll be fine

neobrain:
@gwicks: yeah, please try to figure it out instead of cluttering the repository by creating redundant PRs. What @chinhodado said should already help you, if not then you can always ask around on IRC.

gwicks:
Alright, I'll do it the right way, turns out I did the rebase wrong, so I reverted it and am trying again

neobrain:
For reference, c084606f701372 is the current version of this branch at the time of writing this comment.

gwicks:
There, think I got it that time, sorry about that, never had to use rebase before.

neobrain:
Looks fine, please also fixup the commit message though. Currently it's just a merger of all previous commit messages, which of course doesn't make a lot of sense.

gwicks:
I ammended my commit message using git commit --amend, then I force pushed, why did it break I wonder?

chinhodado:
A lot of other PRs have been merged recently so there's a conflict. You'll need to do a rebase.

gwicks:
So basically just pull in from upstream and fix it locally on my repo?

chinhodado:
http://stackoverflow.com/questions/7929369/how-to-rebase-local-branch-with-remote-master

gwicks:
Ah, thanks, didn't know exactly what you meant by use rebase, thanks for all your help!

gwicks:
Got it all updated now, should be good to go

neobrain:
Referring back to my previous comment:

    LGTM other than the extern mess in main.cpp, which shouldn't be necessary, since all of those are done in getopt.h already. If it can't be helped, we should at least figure out why it's required.

Could *someone* look into this, please?

externals/getopt/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+set(SRCS
+            getopt/getopt.c
+            )
+set(HEADERS
+            getopt/getopt.h
+            )
+
+create_directory_groups(${SRCS} ${HEADERS})
+add_library(getopt ${SRCS} ${HEADERS})
```
neobrain: `target_compile_definitions(getopt INTERFACE STATIC_GETOPT)` should define STATIC_GETOPT for all targets linking against our getopt library. If getopt itself needs to be compiled with that flag too, use PUBLIC instead of INTERFACE.

CMakeLists.txt
```diff
@@ -203,6 +203,11 @@ add_subdirectory(${INI_PREFIX})
 
 include_directories(externals/nihstro/include)
 
+if (MSVC)
+    include_directories(externals/getopt)
```
neobrain: Try removing this command and instead specifying `target_include_directories(getopt INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})` in the getopt CMakeLists.txt. This should add the proper include directories for all projects linking against the getopt target.

src/citra/CMakeLists.txt
```diff
@@ -16,8 +16,14 @@ create_directory_groups(${SRCS} ${HEADERS})
...
+    target_link_libraries(citra getopt)
+endif()
 target_link_libraries(citra ${PLATFORM_LIBRARIES})
 
+
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux|FreeBSD|OpenBSD|NetBSD")
     install(TARGETS citra RUNTIME DESTINATION "${CMAKE_INSTALL_PREFIX}/bin")
 endif()
+
+#install(TARGETS citra RUNTIME DESTINATION ${bindir})
```
neobrain: Merge conflict here. This line was replaced since the initial version of this PR ;)

src/citra/citra.cpp
```diff
@@ -2,7 +2,17 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+
```
neobrain: Redundant newline

src/citra/CMakeLists.txt
```diff
@@ -16,8 +16,14 @@ create_directory_groups(${SRCS} ${HEADERS})
 add_executable(citra ${SRCS} ${HEADERS})
 target_link_libraries(citra core common video_core)
 target_link_libraries(citra ${GLFW_LIBRARIES} ${OPENGL_gl_LIBRARY} inih)
+if (MSVC)
+    target_link_libraries(citra getopt)
+endif()
 target_link_libraries(citra ${PLATFORM_LIBRARIES})
 
+
```
neobrain: Redundant newline

src/citra/citra.cpp
```diff
@@ -2,7 +2,17 @@
...
 #include <string>
+#include <thread>
+#include <iostream>
+
+#ifdef _MSC_VER
+#include "getopt/getopt.h"
```
neobrain: I think "#include <getopt.h>" should work fine on MSVC now, too.

src/citra/citra.cpp
```diff
@@ -2,7 +2,17 @@
...
+#else
+#include <unistd.h>
+#include <getopt.h>
```
neobrain: Most documentation I could find says that `unistd.h` already has all getopt defines. Hence I guess you can drop the explicit getopt.h include?

gwicks:
Looks removing getopt.h from the includes for other platforms didnt work, ill revert that

src/citra/citra.cpp
```diff
@@ -3,6 +3,15 @@
 // Refer to the license.txt file included.
 
 #include <string>
+#include <thread>
```
neobrain: why?

neobrain:
LGTM other than the `<thread>` include in main.cpp.

gwicks:
I don't recall adding the thread, that's odd

neobrain:
I guess we can worry about the thread include another day.

PR #888, opened 2015-06-27
Core\HLE : Fix Warning

zawata:
"signed/unsigned mismatch"

I am unsure of exactly the best way to deal with this but to me it seemed ideal to make the comparison with two variables of the same type.

Of course this does present the unlikely scenario that the variable 'i' could be negative because its no longer unsigned.

Whats the ideal or recommended way to fix this?

### DIFF
--- a/src/core/hle/function_wrappers.h
+++ b/src/core/hle/function_wrappers.h
@@ -113,7 +113,7 @@ template<ResultCode func(u32)> void Wrap() {
     FuncReturn(func(PARAM(0)).raw);
 }
 
-template<ResultCode func(s64*, u32, u32*, s32)> void Wrap(){
+template<ResultCode func(s64*, u32, u32*, u32)> void Wrap(){
     FuncReturn(func((s64*)Memory::GetPointer(PARAM(0)), PARAM(1), (u32*)Memory::GetPointer(PARAM(2)),
         (s32)PARAM(3)).raw);
 }
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -333,7 +333,7 @@ static ResultCode GetResourceLimit(Handle* resource_limit, Handle process_handle
 
 /// Get resource limit current values
 static ResultCode GetResourceLimitCurrentValues(s64* values, Handle resource_limit_handle, u32* names,
-    s32 name_count) {
+    u32 name_count) {
     LOG_TRACE(Kernel_SVC, "called resource_limit=%08X, names=%p, name_count=%d",
         resource_limit_handle, names, name_count);
 
@@ -349,7 +349,7 @@ static ResultCode GetResourceLimitCurrentValues(s64* values, Handle resource_lim
 
 /// Get resource limit max values
 static ResultCode GetResourceLimitLimitValues(s64* values, Handle resource_limit_handle, u32* names,
-    s32 name_count) {
+    u32 name_count) {
     LOG_TRACE(Kernel_SVC, "called resource_limit=%08X, names=%p, name_count=%d",
         resource_limit_handle, names, name_count);
 


mailwl:
count can't be negative (i have had 2 apples, eat 3 apples, now i have -1 apples), so, in this case signed and unsigned as index is equal. i think,  s32 is ok

archshift:
`s32` should only be used for types where the size needs to be fixed, which is not the case. You should use `int` instead.

mailwl:
@archshift, good point
anyway, all this eliminating compile warnings is just cosmetic changes, nothing real new in emulator itself

zawata:
@archshift will do, i only went with my gut feeling
@mailwl of course it is, but its the only way i can help as im not good enough to do graphic programming let alone hardware/software reverse engineering
plus warning fixing actually can lead to bug uncovering so...until I find some other way to help ill keep trying to do this

neobrain:
Btw, is there any actual reason as to why this function was designed to take `name_count` as a signed integer argument?

For what it's worth, I generally try to make variables which are not intended to ever become negative to be unsigned. Hence I'd rather leave `i` unsigned and cast `name_count` to an unsigned value here. Ultimately it's probably not too important since the code will likely compile to the same binary anyway.

neobrain:
@mailwl Warnings often hint at actual software bugs. When they don't, accumulated warning spam hides the actually useful warnings. As such, it's very much always beneficial to fix warnings.

mailwl:
@zawata , the emulator itself is very good, it's like a miracle when changing one or two line of code lets the whole game run.
the main problem, at my point of view - poor HLE emulating, most service functions even not stubbed.
good stubbing HLE with description from 3dbrew.org - good way to help citra (imho! i'm even not collaborator)

mailwl:
@neobrain did you try to check static analize with PVS Studio? http://www.viva64.com/ru/pvs-studio/
i saw the issue about static analize, but pvs is real monster in code checking

neobrain:
@mailwl Actually, I'm really not a fan of stubbing functions. It's not very useful per se unless it's proven to actually help get a specific game further, and by the time someone actually implements stubbed functions, bugs about the function signature may have been found and fixed on 3dbrew but not in Citra.

Also no, I haven't used PVS studio.

mailwl:
@neobrain well, when unimplemented functions should return result error, not success, at least in debug version, because result may affect on further code, far from real problem

zawata:
@neobrain 
```C++
for (unsigned int i = 0; i < static_cast<unsigned>(name_count); ++i)
```
?

neobrain:
yes.

zawata:
@ neobrain yeah i know it isnt, most of the stuff im doing isnt though lol
i can still push it though to make you more comfortable

should i add more fixes to this PR to make it less almost useless?

Subv:
I'm more comfortable with making name_count an u32.

zawata:
@subv it would make sense but it need not affect the rest of the function
which seeing as its name_"count" it probably wont

zawata:
this should probably be either concluded(tell me exactly what is wanted) or closed as its seems to be, literally, one of the lowest priority changes possible and clogs up the PR list

I cast my vote(if i have one) to subv's suggestion because it seems the cleanest

yuriks:
We're unlikely to hit this corner case either way (Who will allocate a >2GB buffer to hold the output? It's not like they even can.) so making name_count u32 sounds good to me.

neobrain:
I wonder whether `name_count` was actually documented to be an s32 value for a reason. Maybe applications can pass -1 for the parameter value to trigger special behavior? If that were the case, the proposed code would fail horribly.

Subv:
@neobrain Everything in the SVC page of 3dbrew is marked s32, I think they just did it to be safe, since you (usually) can't differentiate between signed and unsigned in the disassembly.
I'm sure if any game passes 0xFFFFFFFF as name_count, we'll notice, as it will start spamming memory errors like crazy when trying to write the output.

zawata:
Alright so I'll change the code in both cases

Subv:
LGTM, please squash it

yuriks:
This doesn't build, I believe you need to change the wrappers too.

neobrain:
@zawata Please add a followup comment whenever updating a PR, otherwise people won't get any email notifications about it.

zawata:
@neobrain of course, will do next time, sorry

neobrain:
Please squash the three commits into a single one. Should be good to merge, then.

zawata:
squashed it a few days ago but forgot to push 
done
also what happened with appveyor?

yuriks:
CI has been kinda flaky lately. Will test locally and merge if ok.

yuriks:
97 -> 95 warnings on MSVC, LGTM

(correct branch this time -_-)

zawata:
@yuriks hope this isnt like against the rules(commenting on a closed PR) but thanks for all of that work you did a few hours ago, in both PRs you merged. got a few more warnings when turning up the warning level, i hope i can get that number down to 0!

lioncash:
@zawata It's fine to comment on closed PRs, don't worry about it :)

PR #892, opened 2015-06-28
Yet More Warning Fixes

zawata:
Pretty sure I got all of them

### DIFF
--- a/src/common/string_util.cpp
+++ b/src/common/string_util.cpp
@@ -313,7 +313,7 @@ static std::string UTF16ToUTF8(const std::wstring& input)
     std::string output;
     output.resize(size);
 
-    if (size == 0 || size != WideCharToMultiByte(CP_UTF8, 0, input.data(), input.size(), &output[0], output.size(), nullptr, nullptr))
+    if (size == 0 || size != WideCharToMultiByte(CP_UTF8, 0, input.data(), static_cast<int>(input.size()), &output[0], static_cast<int>(output.size()), nullptr, nullptr))
         output.clear();
 
     return output;
--- a/src/core/loader/3dsx.cpp
+++ b/src/core/loader/3dsx.cpp
@@ -191,7 +191,7 @@ static THREEDSX_Error Load3DSXFile(FileUtil::IOFile& file, u32 base_addr, Shared
                             *pos = (addr);
                             break;
                         case 1:
-                            *pos = (addr - in_addr);
+                            *pos = static_cast<u32>(addr - in_addr);
                             break;
                         default:
                             break; //this should never happen
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -103,7 +103,7 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
         case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[0], 0x23c):
         case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[1], 0x23d):
         {
-            unsigned index = static_cast<int>(id - PICA_REG_INDEX(command_buffer.trigger[0]));
+            unsigned index = static_cast<unsigned>(id - PICA_REG_INDEX(command_buffer.trigger[0]));
             u32* head_ptr = (u32*)Memory::GetPhysicalPointer(regs.command_buffer.GetPhysicalAddress(index));
             g_state.cmd_list.head_ptr = g_state.cmd_list.current_ptr = head_ptr;
             g_state.cmd_list.length = regs.command_buffer.GetSize(index) / sizeof(u32);
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -103,7 +103,7 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
         case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[0], 0x23c):
         case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[1], 0x23d):
         {
-            unsigned index = id - PICA_REG_INDEX(command_buffer.trigger[0]);
+            unsigned index = static_cast<int>(id - PICA_REG_INDEX(command_buffer.trigger[0]));
             u32* head_ptr = (u32*)Memory::GetPhysicalPointer(regs.command_buffer.GetPhysicalAddress(index));
             g_state.cmd_list.head_ptr = g_state.cmd_list.current_ptr = head_ptr;
             g_state.cmd_list.length = regs.command_buffer.GetSize(index) / sizeof(u32);
--- a/src/video_core/debug_utils/debug_utils.cpp
+++ b/src/video_core/debug_utils/debug_utils.cpp
@@ -85,6 +85,7 @@ void GeometryDumper::AddTriangle(Vertex& v0, Vertex& v1, Vertex& v2) {
     vertices.push_back(v1);
     vertices.push_back(v2);
 
+    int num_vertices = static_cast<int>(vertices.size());
     faces.push_back({ num_vertices-3, num_vertices-2, num_vertices-1 });
 }
 
@@ -240,6 +241,8 @@ void DumpShader(const u32* binary_data, u32 binary_size, const u32* swizzle_data
 
     dvle.main_offset_words = main_offset;
     dvle.output_register_table_offset = write_offset - dvlb.dvle_offset;
+    dvle.output_register_table_size = static_cast<uint32_t>(output_info_table.size());
+    QueueForWriting((u8*)output_info_table.data(), static_cast<u32>(output_info_table.size() * sizeof(OutputRegisterInfo)));
 
     // TODO: Create a label table for "main"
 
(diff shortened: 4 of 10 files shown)


mailwl:
@zawata, you on windows? may you check what says http://www.viva64.com/en/pvs-studio ?
@neobrain told he do not used it

zawata:
Finally caught this warning : 
```
5>C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\utility(144): warning C4244: 'initializing' : conversion from 'unsigned __int64' to 'const u32', possible loss of data
5>          C:\Users\Zach\Documents\GitHub\citra\src\video_core/pica.h(875) : see reference to function template instantiation 'std::pair<const _Kty,_Ty>::pair<unsigned __int64,const char(&)[12],void>(_Other1 &&,_Other2)' being compiled
5>          with
5>          [
5>              _Kty=u32
5>  ,            _Ty=std::string
5>  ,            _Other1=unsigned __int64
5>  ,            _Other2=const char (&)[12]
5>          ]
5>          C:\Users\Zach\Documents\GitHub\citra\src\video_core/pica.h(875) : see reference to function template instantiation 'std::pair<const _Kty,_Ty>::pair<unsigned __int64,const char(&)[12],void>(_Other1 &&,_Other2)' being compiled
5>          with
5>          [
5>              _Kty=u32
5>  ,            _Ty=std::string
5>  ,            _Other1=unsigned __int64
5>  ,            _Other2=const char (&)[12]
5>          ]
```
This pesky warning popped up about 28 times in the build log and frankly I'm a little ashamed of how long it took me to find it...

src/core/loader/3dsx.cpp
```diff
@@ -179,7 +179,7 @@ static THREEDSX_Error Load3DSXFile(FileUtil::IOFile& file, u32 base_addr)
                     pos += table.skip;
                     s32 num_patches = table.patch;
                     while (0 < num_patches && pos < end_pos) {
-                        u32 in_addr = (char*)pos - (char*)&all_mem[0];
+                        u32 in_addr = static_cast<u32>((char*)pos - (char*)&all_mem[0]);
```
neobrain: The variable type should be changed to ptrdiff_t instead.

src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -84,7 +84,7 @@ void GeometryDumper::AddTriangle(Vertex& v0, Vertex& v1, Vertex& v2) {
     vertices.push_back(v1);
     vertices.push_back(v2);
 
-    int num_vertices = vertices.size();
+    int num_vertices = static_cast<int>(vertices.size());
```
neobrain: Change the underlying types to the proper types instead, please.

src/video_core/vertex_shader.cpp
```diff
@@ -96,7 +96,7 @@ static void ProcessShaderCode(VertexShaderState& state) {
             state.program_counter = &program_code[offset] - 1; // -1 to make sure when incrementing the PC we end up at the correct offset
             state.call_stack.push({ offset + num_instructions, return_offset, repeat_count, loop_increment, offset });
         };
-        u32 binary_offset = state.program_counter - program_code.data();
+        u32 binary_offset = static_cast<u32>(state.program_counter - program_code.data());
```
neobrain: This should probably be ptrdiff_t.

zawata: that throws more warnings than it fixes

src/video_core/debug_utils/debug_utils.cpp
zawata: what exactly would be the proper type?

Subv: size_t

zawata: oh...well now i feel dumb...

mailwl: @Subv, size_t - 64 bits on x64 systems

src/video_core/command_processor.cpp
```diff
@@ -61,7 +61,7 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
         case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[0], 0x23c):
         case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[1], 0x23d):
         {
-            unsigned index = id - PICA_REG_INDEX(command_buffer.trigger[0]);
+            unsigned index = static_cast<int>(id - PICA_REG_INDEX(command_buffer.trigger[0]));
```
Subv: static_cast<unsigned>

src/video_core/vertex_shader.cpp
lioncash: What are the warnings in particular?

yuriks: It's not like the rest of the code doesn't assumed this is unsigned, so this is IMO fine. Perhaps assert that `program_counter <= program_code.data()` but that would add overhead to something which is an inner loop.

neobrain: > that throws more warnings than it fixes

@zawata This argument is kind of worrying. You shouldn't be concerned to minimize the warning cound, but instead on fixing the causes of the warnings properly. Blindly adding static_casts everywhere is eventually going to cause a stupid bug eventually, and in most cases will miss out on a chance for clarifying the code by switching to different types rather than forcing a cast.

That said, in this particular case both u32 and `ptrdiff_t` probably work well, as yuriks pointed out (... although it's not that the code "assumes" unsignedness, but it's rather an inherent property of the calculations performed). Using `ptrdiff_t` is the cleaner way to fix this particular warning, however.

src/common/string_util.cpp
```diff
@@ -308,7 +308,7 @@ static std::string UTF16ToUTF8(const std::wstring& input)
     std::string output;
     output.resize(size);
 
-    if (size == 0 || size != WideCharToMultiByte(CP_UTF8, 0, input.data(), input.size(), &output[0], output.size(), nullptr, nullptr))
+    if (size == 0 || size != WideCharToMultiByte(CP_UTF8, 0, input.data(), static_cast<int>(input.size()), &output[0], static_cast<int>(output.size()), nullptr, nullptr))
```
yuriks: I have a feeling we can simply get rid of a lot of these functions instead. (They all seem to deal with Windows "native codepage" stuff or windows UCS-2, when we should just be using utf-8 internally everywhere.)

src/video_core/vertex_shader.cpp
Subv: I suggest we stop using this 'binary_offset' variable and start using the program_counter not as a pointer directly, but as an offset to the program_code.

zawata: @neobrain you're completely right, my mistake
@lioncash 
warning C4244: 'argument' : conversion from 'ptrdiff_t' to 'const u32', possible loss of data
warning C4244: 'argument' : conversion from 'ptrdiff_t' to 'u32', possible loss of data
warning C4244: 'argument' : conversion from 'ptrdiff_t' to 'u32', possible loss of data
warning C4244: 'argument' : conversion from 'ptrdiff_t' to 'u32', possible loss of data
warning C4244: 'argument' : conversion from 'ptrdiff_t' to 'u32', possible loss of data
warning C4244: 'argument' : conversion from 'ptrdiff_t' to 'u32', possible loss of data
warning C4244: 'argument' : conversion from 'ptrdiff_t' to 'u32', possible loss of data

this makes me feel like only more of them would propagate if we use my method of fixing it...

zawata:
I'v read the IRC and comments and if people dislike the way i did this(I don't blame you it isn't the best way to fix them) then I can abandon this and let other people who know whats going on do it instead.
I just read the warning messages from the build log and adjusted things accordingly

src/video_core/vertex_shader.cpp
zawata: conclusions?

changing the type throws warnings but only dealing with arguments for offsets in the function 
'''
static auto call = [&program_code](VertexShaderState& state, u32 offset, u32 num_instructions,
                              u32 return_offset, u8 repeat_count, u8 loop_increment) {
'''
more secifially the arguments "offset and return_offset

src/common/string_util.cpp
zawata: removing them seems logical but im going to leave it for the purposes of this PR

zawata:
any other lines that need changing?

src/video_core/vertex_shader.cpp
neobrain: I guess stop worrying about it for now since it turns out this code will likely be replaced soon anyway. Just use the u32 cast if that's the easiest way to do, or drop the change from this PR altogether.

zawata: @neobrain since bunnei is going to replacing the code I'll just drop it

yuriks: What? bunnei isn't replacing this code afaik.

zawata: im probably mistaken then, i thought he was rewriting the vertex shader?

yuriks: He's writing a vertex shader JIT, but this interpreter will stay.

zawata: oh i see so then should i keep the cast or ditch it?
theres no benefit to either so i think for purely asthetic reasons ill keep it(silencing the warning)

yuriks: Is there any reason "program_counter" is a pointer in the first place? Looking at the code, it's only used as an actual pointer *once*, and in all other cases is being derived from an index or being subtracted from vector.data() to get the offset. It should be just replaced with an index.

neobrain: ... maybe let's stop overanalyzing this and postponing this particular change to a later stage, for the sake of merging the other gazillion warning fixes in this PR.

neobrain:
LGTM other than the vertex shader change, which should just be dropped from this PR for the sake of getting everything else merged.

src/video_core/vertex_shader.cpp
zawata: @neobrain change dropped

zawata:
@neobrain merge conflicts?

archshift:
You have to rebase your branch on the latest master.

zawata:
Rebased

Subv:
LGTM

neobrain:
Does anyone see what's up with the Travis OS X build?

archshift:
Yeah, he's narrowing types inside of a list-initialized type, which definitely needs to be fixed.

zawata:
@archshift how should i go about fixing that?
ive got a few ideas but i dont know which would be best

src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -85,7 +85,7 @@ void GeometryDumper::AddTriangle(Vertex& v0, Vertex& v1, Vertex& v2) {
     vertices.push_back(v1);
     vertices.push_back(v2);
 
-    int num_vertices = vertices.size();
+    size_t num_vertices = vertices.size();
     faces.push_back({ num_vertices-3, num_vertices-2, num_vertices-1 });
```
yuriks: Narrowing is happening here. Instead of adding casts to all 3 member of the initializer, I think you can just keep the int there instead (and cast `vertices.size()`)

yuriks:
Trying to get momentum back on this before it degrades too much. Can you fix that error and rebase/squash it so we can merge?

yuriks:
97 -> 95 warnings on MSVC, LGTM

yuriks:
I accidentally merged the wrong branch just now (meant to merge your other one), so I'll just keep this merged. I addressed the narrowing error myself.

PR #893, opened 2015-06-28
Replace standard uint*_t and int*_t with CommonTypes’ u* and s* types

linkmauve:
The ARM core was ridden with those, but there was also an occurence in the CitraQt debugger.

This commit is purely stylistic, there is no change in behaviour.

### DIFF
--- a/src/citra_qt/debugger/callstack.cpp
+++ b/src/citra_qt/debugger/callstack.cpp
@@ -4,12 +4,14 @@
 
 #include <QStandardItemModel>
 
+#include "common/common_types.h"
+#include "common/symbols.h"
+
 #include "callstack.h"
 
 #include "core/core.h"
 #include "core/arm/arm_interface.h"
 #include "core/memory.h"
-#include "common/symbols.h"
 #include "core/arm/disassembler/arm_disasm.h"
 
 CallstackWidget::CallstackWidget(QWidget* parent): QDockWidget(parent)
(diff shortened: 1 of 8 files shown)


zawata:
i came across this several times during my warning fixes and was unsure of why it was that way
i just left it because i figured there was a reason

Subv:
Is there any reason for this? the way I see it, it's just adding more includes and making the dyncom harder for other projects to use for no gain

linkmauve:
I was thinking about unification of the codebase.

My personal opinion would be to drop those ugly u32, s16, etc. and use the proper uint32_t, int16_t types instead, everywhere, and drop common_types.

neobrain:
I'm not opposed to this change. I'd also prefer dropping u32 & pals, though.

mailwl:
please dont do it!
i like u32 s32 and friends

lioncash:
likewise

bunnei:
LGTM. I don't really have a strong opinion in keeping them or changing all to uint*... but for now I think this change makes sense.

yuriks:
I like the u* overloads, but I think we should keep dyncom as is. It's not a significant cleanup to change it, generates diff churn and also makes it harder to use it in other projects.

purpasmart96:
I agree with yurkis, I like the u* typedefs too myself.

bunnei:
So can this be merged...?

neobrain:
@linkmauve Please decide whether you want this merged or not.

linkmauve:
I rebased it as is, as I think uniformity is better in the case of Citra (and dyncom was already ridden with our own integer types anyway).

bunnei:
@linkmauve I'm not sure why this wasn't merged after you rebased.... I guess no one noticed it was good to go. Can you rebase again to address the current conflicts? Sorry :-/

linkmauve:
Here, rebased and included the newer files that also added some standard int types. :)

PR #895, opened 2015-06-29
vfp: Handle accesses to FPINST/FPINST2 system registers.

lioncash:
Also has a side-benefit of correcting access to the FPEXC register. Only privileged mode stuff. No userland changes.


bunnei:
LGTM

PR #897, opened 2015-06-29
vfp: Change return type of VFPInit from unsigned int to void.

lioncash:
This is technically cruft from armemu code, where it took specific function pointer types (where an init routine with a return code was one of them).


Subv:
LGTM

PR #904, opened 2015-07-07
Y2R: Fix narrowing warning

aroulin:
Fix initializer-list narrowing warning (u32 -> u8)

### DIFF
--- a/src/core/hw/y2r.cpp
+++ b/src/core/hw/y2r.cpp
@@ -111,7 +111,7 @@ static void SendData(const u32* input, ConversionBuffer& buf, int amount_of_data
         while (output < unit_end) {
             u32 color = *input++;
             Math::Vec4<u8> col_vec{
-                (color >> 24) & 0xFF, (color >> 16) & 0xFF, (color >>  8) & 0xFF, alpha,
+                (u8)(color >> 24), (u8)(color >> 16), (u8)(color >> 8), alpha
             };
 
             switch (output_format) {


yuriks:
LGTM

archshift:
(Oops)

Wht not keep the brace initialization but just change the masks to casts?

aroulin:
Changed masks to casts. It's nice that integer conversions to smaller unsigned types is well-defined (unlike conversions to smaller signed types which are implementation-defined).

neobrain:
I find those ugly casts both needlessly verbose, yet much less clear than the manually masking. Why not just use constructor initialization by changing `{` and `}` to `(` and `(`, as discussed on IRC earlier today?

yuriks:
One advantage of the casts is that I think MSVC will still generate narrowing warnings with the function call.

neobrain:
@yuriks: Could you verify that claim? I'd rather not have such ugly code used based on preventing hypothetical warning messages (also, if MSVC is really that dumb about narrowing warnings, we might actually want to consider disabling that one).

yuriks:
@neobrain MSVC emits warnings for any implicit narrowing conversions (assignment, parameters passing, etc.). We could disable them, but I think they're valuable in identifying mistakes in type usage. Instead I'd rather we simply allow C-style casts for simple integer conversions, since `static_cast` is verbose and distracting.

yuriks:
As per #919, change it to C-style casts and then LGTM.

aroulin:
Changed to C-style casts. Masks with 0xFF are not necessary but if that's better for readability, I can put them back.

yuriks:
LGTM

purpasmart96:
LGTM

PR #905, opened 2015-07-07
AppVeyor: Remove submodule checkout depth

yuriks:
This avoid intermittent build failures due to a commit not being
reachable using a fixed depth, at the expense of longer checkout times.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ configuration:
   - Release
 
 install:
-  - git submodule update --init --recursive --depth 20
+  - git submodule update --init --recursive
 
 before_build:
   - mkdir build


chinhodado:
Under what circumstance would this happen? I.e., why would a depth > 20 be needed?

yuriks:
It seems that `git submodule` clones the submodule repo starting from `master`, and then checks out the specific revision. If the commit we're pointing to in the submodule isn't within 20 commits of the current master, then it'll fail to checkout.

This is happening here for example: https://ci.appveyor.com/project/bunnei/citra/build/1.0.1366

bunnei:
LGTM

PR #906, opened 2015-07-07
Loader: Fix format specifier warning

aroulin:
Another possibility that keeps uppercase hex:

     "0x%" PRIXPTR, (uintptr_t) exheader_header.arm11_system_local_caps.resource_limit_descriptor

but it requires a cast and importing inttypes.h


Subv:
This doesn't look right, resource_limit_descriptor is an array of arrays, I don't know what the original intent of this code was but it's not OK.

yuriks:
Yeah, looks like it was just wrong to begin with. It's too much crap to just dump the whole thing in logs, so I'd just remove that log.

purpasmart96:
Yeah just get rid of it.

yuriks:
If someone wants to have logging for this then it can be added during parsing at https://github.com/citra-emu/citra/blob/master/src/core/hle/kernel/process.cpp#L30

aroulin:
Done. That line seemed also weird to me. At least ```%X``` for pointers was plain wrong.

yuriks:
LGTM

PR #908, opened 2015-07-09
[RFC] SIMD shaders

bunnei:
Adds an optimized SIMD-based vertex shader interpreter for x86. This should be as accurate as the older vertex shader interpreter, except much faster. That being said, it's only really faster by design - I've not done any serious profiling, and I'm sure we can get even more performance with more fine-tuned optimization (any related suggestions are welcome...)

### DIFF
--- a/src/common/cpu_detect_x86.cpp
+++ b/src/common/cpu_detect_x86.cpp
@@ -1,0 +1,229 @@
+// Copyright 2008 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <cstring>
+#include <string>
+
+#include "common_types.h"
+#include "cpu_detect.h"
+#include "intrinsics.h"
+
+#ifndef _WIN32
+
+#ifdef __FreeBSD__
+#include <sys/types.h>
+#include <machine/cpufunc.h>
+#endif
+
+static inline void __cpuidex(int info[4], int function_id, int subfunction_id)
+{
+#ifdef __FreeBSD__
+    // Despite the name, this is just do_cpuid() with ECX as second input.
+    cpuid_count((u_int)function_id, (u_int)subfunction_id, (u_int*)info);
+#else
+    info[0] = function_id;    // eax
+    info[2] = subfunction_id; // ecx
+    __asm__(
+        "cpuid"
+        : "=a" (info[0]),
+        "=b" (info[1]),
+        "=c" (info[2]),
+        "=d" (info[3])
+        : "a" (function_id),
+        "c" (subfunction_id)
+        );
+#endif
+}
+
+static inline void __cpuid(int info[4], int function_id)
+{
+    return __cpuidex(info, function_id, 0);
+}
+
+#define _XCR_XFEATURE_ENABLED_MASK 0
+static u64 _xgetbv(u32 index)
+{
+    u32 eax, edx;
+    __asm__ __volatile__("xgetbv" : "=a"(eax), "=d"(edx) : "c"(index));
+    return ((u64)edx << 32) | eax;
+}
+
+#endif // ifndef _WIN32
+
+namespace Common {
+
+CPUInfo cpu_info;
+
+CPUInfo::CPUInfo() {
+    Detect();
+}
+
+// Detects the various CPU features
+void CPUInfo::Detect() {
+    memset(this, 0, sizeof(*this));
+#ifdef _M_X86_64
+    Mode64bit = true;
+    OS64bit = true;
+#endif
+    num_cores = 1;
+
+    // Set obvious defaults, for extra safety
+    if (Mode64bit) {
+        bSSE = true;
+        bSSE2 = true;
+        bLongMode = true;
+    }
+
+    // Assume CPU supports the CPUID instruction. Those that don't can barely
+    // boot modern OS:es anyway.
+    int cpu_id[4];
+    memset(brand_string, 0, sizeof(brand_string));
+
+    // Detect CPU's CPUID capabilities, and grab CPU string
+    __cpuid(cpu_id, 0x00000000);
+    u32 max_std_fn = cpu_id[0];  // EAX
+    *((int *)brand_string) = cpu_id[1];
+    *((int *)(brand_string + 4)) = cpu_id[3];
+    *((int *)(brand_string + 8)) = cpu_id[2];
+    __cpuid(cpu_id, 0x80000000);
+    u32 max_ex_fn = cpu_id[0];
+    if (!strcmp(brand_string, "GenuineIntel"))
+        vendor = VENDOR_INTEL;
+    else if (!strcmp(brand_string, "AuthenticAMD"))
+        vendor = VENDOR_AMD;
+    else
+        vendor = VENDOR_OTHER;
+
+    // Set reasonable default brand string even if brand string not available.
+    strcpy(cpu_string, brand_string);
+
+    // Detect family and other misc stuff.
+    bool ht = false;
+    HTT = ht;
+    logical_cpu_count = 1;
+    if (max_std_fn >= 1) {
+        __cpuid(cpu_id, 0x00000001);
+        int family = ((cpu_id[0] >> 8) & 0xf) + ((cpu_id[0] >> 20) & 0xff);
+        int model = ((cpu_id[0] >> 4) & 0xf) + ((cpu_id[0] >> 12) & 0xf0);
+        // Detect people unfortunate enough to be running Dolphin on an Atom
+        if (family == 6 && (model == 0x1C || model == 0x26 || model == 0x27 || model == 0x35 || model == 0x36 ||
+            model == 0x37 || model == 0x4A || model == 0x4D || model == 0x5A || model == 0x5D))
+            bAtom = true;
+        logical_cpu_count = (cpu_id[1] >> 16) & 0xFF;
+        ht = (cpu_id[3] >> 28) & 1;
+
+        if ((cpu_id[3] >> 25) & 1) bSSE = true;
+        if ((cpu_id[3] >> 26) & 1) bSSE2 = true;
+        if ((cpu_id[2]) & 1) bSSE3 = true;
+        if ((cpu_id[2] >> 9) & 1) bSSSE3 = true;
+        if ((cpu_id[2] >> 19) & 1) bSSE4_1 = true;
+        if ((cpu_id[2] >> 20) & 1) bSSE4_2 = true;
+        if ((cpu_id[2] >> 22) & 1) bMOVBE = true;
+        if ((cpu_id[2] >> 25) & 1) bAES = true;
+
+        if ((cpu_id[3] >> 24) & 1)
+        {
+            // We can use FXSAVE.
+            bFXSR = true;
+        }
+
+        // AVX support requires 3 separate checks:
+        //  - Is the AVX bit set in CPUID?
+        //  - Is the XSAVE bit set in CPUID?
+        //  - XGETBV result has the XCR bit set.
+        if (((cpu_id[2] >> 28) & 1) && ((cpu_id[2] >> 27) & 1)) {
+            if ((_xgetbv(_XCR_XFEATURE_ENABLED_MASK) & 0x6) == 0x6) {
+                bAVX = true;
+                if ((cpu_id[2] >> 12) & 1)
+                    bFMA = true;
+            }
+        }
+
+        if (max_std_fn >= 7) {
+            __cpuidex(cpu_id, 0x00000007, 0x00000000);
+            // careful; we can't enable AVX2 unless the XSAVE/XGETBV checks above passed
+            if ((cpu_id[1] >> 5) & 1)
+                bAVX2 = bAVX;
+            if ((cpu_id[1] >> 3) & 1)
+                bBMI1 = true;
+            if ((cpu_id[1] >> 8) & 1)
+                bBMI2 = true;
+        }
+    }
+
+    bFlushToZero = bSSE;
+
+    if (max_ex_fn >= 0x80000004) {
+        // Extract CPU model string
+        __cpuid(cpu_id, 0x80000002);
+        memcpy(cpu_string, cpu_id, sizeof(cpu_id));
+        __cpuid(cpu_id, 0x80000003);
+        memcpy(cpu_string + 16, cpu_id, sizeof(cpu_id));
+        __cpuid(cpu_id, 0x80000004);
+        memcpy(cpu_string + 32, cpu_id, sizeof(cpu_id));
+    }
+    if (max_ex_fn >= 0x80000001) {
+        // Check for more features.
+        __cpuid(cpu_id, 0x80000001);
+        if (cpu_id[2] & 1) bLAHFSAHF64 = true;
+        if ((cpu_id[2] >> 5) & 1) bLZCNT = true;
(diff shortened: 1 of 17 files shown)


src/common/cpu_detect_x86.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+    // Detect family and other misc stuff.
+    bool ht = false;
+    HTT = ht;
+    logical_cpu_count = 1;
+    if (max_std_fn >= 1) {
+        __cpuid(cpu_id, 0x00000001);
+        int family = ((cpu_id[0] >> 8) & 0xf) + ((cpu_id[0] >> 20) & 0xff);
+        int model = ((cpu_id[0] >> 4) & 0xf) + ((cpu_id[0] >> 12) & 0xf0);
+        // Detect people unfortunate enough to be running Dolphin on an Atom
```
Subv: Dolphin?

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,528 @@
...
+        reg.y = input.attr[i].y.ToFloat32();
+        reg.z = input.attr[i].z.ToFloat32();
+        reg.w = input.attr[i].w.ToFloat32();
+    }
+
+#if _M_SSE >= 0x401
+    while (true) {
+        if (!state.call_stack.empty()) {
+            auto& top = state.call_stack.back();
+            if (&program_code[state.pc] - program_code.data() == top.final_address) {
```
Subv: if (state.pc == top.final_address)

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,528 @@
...
+    for (int i = 0; i < 7; ++i) {
+        const auto& output_register_map = regs.vs_output_attributes[i];
+
+        u32 semantics[4] = {
+            output_register_map.map_x, output_register_map.map_y,
+            output_register_map.map_z, output_register_map.map_w
+        };
+
+        for (int comp = 0; comp < 4; ++comp) {
+            f32* out = ((f32*)&ret) + semantics[comp];
```
Subv: This is still very evil.

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+            src1.raw_i = NEGATE(SWIZZLE(instr.mad.src1, swizzle.src1_selector), swizzle.negate_src1); \
+            src2.raw_i = NEGATE(SWIZZLE(instr.mad.src2, swizzle.src2_selector), swizzle.negate_src2); \
+            src3.raw_i = NEGATE(SWIZZLE(instr.mad.src3, swizzle.src3_selector), swizzle.negate_src3); \
+            temp.raw_f = operation_; \
+            __m128i dst_mask = swizzle_dst_mask[swizzle.dest_mask]; \
+            auto& dest = state.OutputReg(instr.mad.dest); \
+            dest.raw_i = _mm_or_si128(_mm_andnot_si128(dst_mask, dest.raw_i), _mm_and_si128(dst_mask, temp.raw_i)); \
+        }
+
+        #define FORMAT5I(operation_) { \
```
neobrain: These blocks should be wrapped around `do`-`while(false)`.

src/video_core/vertex_shader_simd.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include "vertex_shader.h"
+
+namespace Pica {
+
+namespace VertexShaderSIMD {
+
+struct Reg {
+    union {
+        struct {
```
neobrain: As far as I can see, this blatantly violates strict aliasing rules. We use -fno-strict-aliasing, but it's bad style regardless and throws out potentially useful optimizations that we could gain with -f-strict-aliasing. Read the `x`, `y`, `z`, and `w` members using getter methods instead by decoding from raw_f/raw_i.

Furthermore, putting the floating and integer portions together like this is ugly, since (as far as I can see) it's already clear at compile-time anyway which of the two fields will be used for a particular variable.

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+};
+static_assert(sizeof(Instruction) == 0x4, "Incorrect structure size");
+static_assert(std::is_standard_layout<Instruction>::value, "Structure does not have standard layout");
+
+static inline void Call(CoreState& state, u32 offset, u32 num_instructions, u32 return_offset, u8 repeat_count, u8 loop_increment) {
+    state.pc = offset;
+    state.call_stack.push_back({ offset + num_instructions, return_offset, repeat_count, loop_increment, offset });
+};
+
+static inline bool Compare(Instruction::CompareOp op, f32 src1, f32 src2) {
```
neobrain: Why `static inline` as opposed to `static`?

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+        BitField<0x05, 0x7, u32> src3i;
+        BitField<0x0c, 0x5, u32> src2i;
+
+        BitField<0x18, 0x5, u32> dest;
+    } mad;
+};
+static_assert(sizeof(Instruction) == 0x4, "Incorrect structure size");
+static_assert(std::is_standard_layout<Instruction>::value, "Structure does not have standard layout");
+
+static inline void Call(CoreState& state, u32 offset, u32 num_instructions, u32 return_offset, u8 repeat_count, u8 loop_increment) {
```
neobrain: Now that this is a free function, give it a more descriptive name.

src/video_core/vertex_shader_simd.cpp
neobrain: Now that this is a global function, give it a more descriptive name, e.g. CompareFloatRegisters.

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+        return results[0];
+
+    case flow_control.JustY:
+        return results[1];
+    }
+
+    return false;
+};
+
+void InitCore(CoreState& state) {
```
neobrain: Make this return a `CoreState` object instead.

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+
+namespace VertexShaderSIMD {
+
+#if _M_SSE >= 0x401
+static __m128i swizzle_dst_mask[16];
+static __m128i swizzle_src_selector[256];
+#endif
+
+void Init() {
+#if _M_SSE >= 0x401
```
neobrain: This is already checked in all instances where Init() is called, when it really should only be checked once. Please decide how to address this.

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+
+using nihstro::OpCode;
+using nihstro::SwizzlePattern;
+
+namespace Pica {
+
+namespace VertexShaderSIMD {
+
+#if _M_SSE >= 0x401
+static __m128i swizzle_dst_mask[16];
```
neobrain: These never change, yet aren't marked as const and are initialized in a somewhat pointless Init() function. Instead, you could do

    struct SwizzleDestMask : std::array<__m128i, 16> {
        SwizzleDestMask() {
            // Initialization here
            // ...
        }
    } static const swizzle_dest_mask;

(cf. http://ideone.com/eJh7rq ;  alternatively to inheriting `std::array`, you could just keep the array as a member and provide a [] operator for syntactic sugar ; alternatively to that, you could initialize the array via a call to a lambda function, but you probably wouldn't like that)

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+void Init() {
+#if _M_SSE >= 0x401
+    for (int i = 0; i < 16; ++i) {
+        swizzle_dst_mask[i].m128i_u32[0] = ((i >> 3) & 1) * 0xffffffff;
+        swizzle_dst_mask[i].m128i_u32[1] = ((i >> 2) & 1) * 0xffffffff;
+        swizzle_dst_mask[i].m128i_u32[2] = ((i >> 1) & 1) * 0xffffffff;
+        swizzle_dst_mask[i].m128i_u32[3] = (i & 1) * 0xffffffff;
+    }
+
+#define SELECT(n) (((((n) << 2) + 3) << 24) | ((((n) << 2) + 2) << 16) | ((((n) << 2) + 1) << 8) | ((n) << 2))
```
neobrain: A comment explaining what's going on here would be nice.

neobrain: Same for the line above, actually. Don't assume people already know how `swizzle_dst_mask` and  `swizzle_src_selector` are used later on.

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+    for (int i = 0; i < num_attributes; ++i) {
+        Reg& reg = state.input[reg_map.GetRegisterForAttribute(i)];
+        reg.x = input.attr[i].x.ToFloat32();
+        reg.y = input.attr[i].y.ToFloat32();
+        reg.z = input.attr[i].z.ToFloat32();
+        reg.w = input.attr[i].w.ToFloat32();
+    }
+
+#if _M_SSE >= 0x401
+    while (true) {
```
neobrain: `while (!exit_loop)`

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+
+    return false;
+};
+
+void InitCore(CoreState& state) {
+    std::memcpy(&state.uniform[0].x, g_state.vs.uniforms.f, sizeof(state.uniform));
+}
+
+VertexShader::OutputVertex RunShader(CoreState& state, const VertexShader::InputVertex& input, int num_attributes) {
+    const auto& regs = g_state.regs;
```
neobrain: You forgot copy-pasting the ScopedTimer here, I guess.

neobrain:
This implementation looks like quite the copy-pasta from the original vertex_shader.cpp. Most (if not everything) other than the state object and the actual instruction-implementations is shared between the SIMD and non-SIMD code path (and then there are bits and pieces that you rewrote to fit your coding style). IMO it would be easy to encapsulate those differences in a boolean template parameter. This will also make adding geometry shaders a lot easier because I won't have to bother updating the same code twice.

neobrain:
Also, it's "SIMD", not "Simd".

src/video_core/vertex_shader_simd.cpp
Subv: Why?

src/video_core/vertex_shader_simd.cpp
Subv: If that's the alternative, I prefer this remains without the const modifier

src/video_core/vertex_shader_simd.cpp
neobrain: It's a common idiom to use when writing macros like that, to prevent misuse when e.g. used within if-else blocks: stackoverflow.com/questions/257418/do-while-0-what-is-it-good-for

src/video_core/vertex_shader_simd.cpp
neobrain: @Subv: It's not only about the const modifier, but also about the artificial Init() function, which adds a dependency between the `pica.cpp` and the `vertex_shader_simd.cpp` translation units. If people are scared of inheriting std::array like above, I don't see what's so terrible about the suggested (first) alternative.

src/video_core/vertex_shader_simd.cpp
Subv: Agreed

src/common/cpu_detect_x86.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+#endif // ifndef _WIN32
+
+namespace Common {
+
+CPUInfo cpu_info;
+
+CPUInfo::CPUInfo() {
+    Detect();
```
jroweboy: Is there any reason to have a private Detect method that is only caused at construction? It looks like it can be safely placed in the constructor.

src/common/cpu_detect_x86.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// Detects the various CPU features
+void CPUInfo::Detect() {
+    memset(this, 0, sizeof(*this));
```
lioncash: lol

src/common/intrinsics.h
```diff
@@ -0,0 +1,31 @@
+// Copyright 2015 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#ifdef _M_X86
+
+#ifdef _MSC_VER
+#include <intrin.h>
```
yuriks: You want either `xmmintrin.h` or `emmintrin.h` instead, (which will also be available in gcc based systems too.)

src/video_core/vertex_shader_simd.h
```diff
@@ -0,0 +1,78 @@
...
+            f32 x, y, z, w;
```
yuriks: `__m128` in `xmmintrin.h` is already a union containing arrays for each possible datatype. You should just use that instead.

src/common/intrinsics.h
```diff
@@ -0,0 +1,31 @@
...
+#else
+#include <x86intrin.h>
+#endif
+
+#if defined _M_GENERIC
+#  define _M_SSE 0
+#elif _MSC_VER || __INTEL_COMPILER
+#  define _M_SSE 0x402
```
yuriks: What's up with these defines? They seems to be a Dolphin/PCSX2 only thing, in which case they shouldn't use this naming convention. But in what situations do we care about it?

src/video_core/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ set(SRCS
             rasterizer.cpp
             utils.cpp
             vertex_shader.cpp
+            vertex_shader_simd.cpp
```
yuriks: Should be `vertex_shader_sse` instead, imo. Also, shouldn't these also be conditional on the architecture like the CPU detection thing.

src/video_core/vertex_shader_simd.cpp
```diff
@@ -0,0 +1,529 @@
...
+        LessEqual = 3,
+        GreaterThan = 4,
+        GreaterEqual = 5,
+        Unk6 = 6,
+        Unk7 = 7
+    };
+
+    // Format used e.g. by arithmetic instructions and comparisons
+    union Common { // TODO: Remove name
+        BitField<0x00, 0x7, u32> operand_desc_id;
```
yuriks: Hex for bit amounts? Thanks but no thanks.

src/video_core/vertex_shader_simd.cpp
yuriks: Split this macro into several lines at least. An alternative suggestion that may or may not be more readable is: `#define SELECT(n) ((n) * 0x04040404u + 0x03020100u)`

src/video_core/vertex_shader_simd.cpp
yuriks: Since both should same the same timer, what I did instead was move that to command processor instead. I think that's better than adding the timer to a header so it can be used in other TUs.

bunnei:
This is about to be superseded by a JIT...

PR #909, opened 2015-07-09
CMake: Fix Debug build configuration in MSVC

yuriks:
Debug was missing compiler flags, causing MSVC to default it to building
with optimizations enabled (making for a not very useful binary for
actual debugging...). Additionally, the variables were re-organized to
remove some redundancy, the old Release build type was removed, and
RelWithDebInfo was renamed to take its place instead.


Subv:
LGTM

bunnei:
LGTM

PR #910, opened 2015-07-09
Tell cmake to install the compiled binaries on Linux

linkmauve:
This will help packaging tremendously, as a `make DESTDIR=… install` will now put every file at their place.

Ideally we should do that on other systems as well, but I don’t know them enough to enable this command there.


yuriks:
LGTM

archshift:
LGTM

PR #911, opened 2015-07-09
Do not load entire RomFS to memory, read from the file as needed instead.

condut:
Changes the loader/ivfs code to read from the file as needed instead of loading the entire RomFS to memory (which can often take 4+ seconds for large RomFSes)

Feel free to change the code to be how you like it, if needed.

### DIFF
--- a/src/core/file_sys/archive_romfs.h
+++ b/src/core/file_sys/archive_romfs.h
@@ -27,9 +27,9 @@ public:
     ResultCode Format(const Path& path) override;
 
 private:
-    std::shared_ptr<FileUtil::IOFile> m_romfs_file;
-    u64 m_offset;
-    u64 m_size;
+    std::shared_ptr<FileUtil::IOFile> romfs_file;
+    u64 data_offset;
+    u64 data_size;
 };
 
 } // namespace FileSys
--- a/src/core/file_sys/archive_romfs.h
+++ b/src/core/file_sys/archive_romfs.h
@@ -27,7 +27,9 @@ public:
     ResultCode Format(const Path& path) override;
 
 private:
-    std::shared_ptr<std::vector<u8>> romfs_data;
+    std::shared_ptr<FileUtil::IOFile> m_romfs_file;
+    u64 m_offset;
+    u64 m_size;
 };
 
 } // namespace FileSys
--- a/src/core/file_sys/ivfc_archive.cpp
+++ b/src/core/file_sys/ivfc_archive.cpp
@@ -75,7 +75,7 @@ size_t IVFCFile::Write(const u64 offset, const u32 length, const u32 flush, cons
 }
 
 size_t IVFCFile::GetSize() const {
-    return sizeof(u8) * data_size;
+    return data_size; // TODO: return value will overflow on 32-bit machines
 }
 
 bool IVFCFile::SetSize(const u64 size) const {
--- a/src/core/file_sys/ivfc_archive.cpp
+++ b/src/core/file_sys/ivfc_archive.cpp
@@ -21,7 +21,7 @@ std::string IVFCArchive::GetName() const {
 }
 
 std::unique_ptr<FileBackend> IVFCArchive::OpenFile(const Path& path, const Mode mode) const {
-    return Common::make_unique<IVFCFile>(m_romfs_file, m_offset, m_size);
+    return Common::make_unique<IVFCFile>(romfs_file, data_offset, data_size);
 }
 
 bool IVFCArchive::DeleteFile(const Path& path) const {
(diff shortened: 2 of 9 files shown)


src/core/file_sys/ivfc_archive.cpp
```diff
@@ -66,8 +63,10 @@ std::unique_ptr<DirectoryBackend> IVFCArchive::OpenDirectory(const Path& path) c
 
 size_t IVFCFile::Read(const u64 offset, const u32 length, u8* buffer) const {
     LOG_TRACE(Service_FS, "called offset=%llu, length=%d", offset, length);
-    memcpy(buffer, data->data() + offset, length);
-    return length;
+    m_romfs_file->Seek(m_offset + offset, SEEK_SET);
+    u32 read_length = (u32) std::min((u64) length, m_size - offset);
```
Subv: Why are you casting to u64 then back to u32?

src/core/file_sys/ivfc_archive.h
```diff
@@ -56,7 +60,9 @@ class IVFCFile : public FileBackend {
     void Flush() const override { }
 
 private:
-    std::shared_ptr<const std::vector<u8>> data;
+    std::shared_ptr<FileUtil::IOFile> m_romfs_file;
+    u64 m_offset;
```
Subv: Please don't use m_*.
Something like data_offset and data_size

src/core/loader/loader.h
```diff
@@ -93,10 +93,13 @@ class AppLoader : NonCopyable {
 
     /**
      * Get the RomFS of the application
-     * @param buffer Reference to buffer to store data
+     * Since the RomFS is huge, we return a file reference instead of copying to a buffer
```
Subv: can be huge*

src/core/loader/ncch.h
```diff
@@ -163,7 +163,8 @@ namespace Loader {
 /// Loads an NCCH file (e.g. from a CCI, or the first NCCH in a CXI)
 class AppLoader_NCCH final : public AppLoader {
 public:
-    AppLoader_NCCH(std::unique_ptr<FileUtil::IOFile>&& file) : AppLoader(std::move(file)) { }
+    AppLoader_NCCH(std::unique_ptr<FileUtil::IOFile>&& file, std::string filepath)
```
lioncash: filepath can be a const reference.

src/core/loader/ncch.cpp
```diff
@@ -286,12 +286,16 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::vector<u8>& buffer) const {
...
         LOG_DEBUG(Loader, "RomFS size:             0x%08X", romfs_size);
 
-        buffer.resize(romfs_size);
-
-        file->Seek(romfs_offset, SEEK_SET);
-        if (file->ReadBytes(&buffer[0], romfs_size) != romfs_size)
+        if (file->GetSize () < romfs_offset + romfs_size)
             return ResultStatus::Error;
 
+        // We reopen the file, to avoid reuse of the file offset
```
Subv: This comment is a bit confusing, could you clarify it please?

src/core/loader/ncch.cpp
```diff
@@ -286,12 +286,16 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::vector<u8>& buffer) const {
...
+        romfs_file.reset(new FileUtil::IOFile(filepath, "rb"));
+        offset = romfs_offset;
+        size = romfs_size;
+
+        if (!romfs_file->IsOpen())
```
Subv: IMO this check should be before the assignment of offset and size

src/core/file_sys/archive_romfs.h
```diff
@@ -27,7 +27,9 @@ class ArchiveFactory_RomFS final : public ArchiveFactory {
     ResultCode Format(const Path& path) override;
 
 private:
-    std::shared_ptr<std::vector<u8>> romfs_data;
+    std::shared_ptr<FileUtil::IOFile> m_romfs_file;
+    u64 m_offset;
```
Subv: Same here

src/core/file_sys/ivfc_archive.cpp
condut: m_size - offset is 64 bit, while length is 32 bit.
Doing the cast I did is safer than casting (m_size - offset) to (u32) since that could truncate.
My (u32) cast is safe since min(u32, u64) -> u32 (but sadly, std::min isn't that smart)

Subv: I see, well then, please remove the whitespaces after the casts

src/core/loader/ncch.cpp
condut: I'll change to "// We reopen the file, to allow its position to be independent from file's" in my coming update, hopefully that's clearer.

src/core/file_sys/ivfc_archive.cpp
```diff
@@ -76,7 +75,7 @@ size_t IVFCFile::Write(const u64 offset, const u32 length, const u32 flush, cons
 }
 
 size_t IVFCFile::GetSize() const {
-    return sizeof(u8) * data->size();
+    return sizeof(u8) * data_size;
```
yuriks: Not sure what this was doing here but.. you can get rid of the `sizeof(u8)` if you want. Also, add a TODO saying that the return value might overflow on 32-bit machines. (Or fix it if you feel like it, otherwise I'll do that as soon as this is merged.)

yuriks:
Not perfect, but good enough, and we really need this. LGTM after you add that TODO and squash.

src/core/file_sys/ivfc_archive.cpp
archshift: Can it overflow if ROMs are never greater than 4GB?

yuriks: Right now, no. It's still wrong. :)

condut:
@yuriks  - updated per your request.

And I was aiming for good enough ;)
Feel free to make it perfect later.

yuriks:
This needs to be rebased.

yuriks:
Author seems to have ragequit on IRC over git yesterday, so I'll assume I can pick this up...

yuriks:
Deprecating in favor of #918

PR #912, opened 2015-07-10
Core: Properly configure address space during binary loading

yuriks:
The code now properly configures the process image to match the loaded
binary segments (code, rodata, data) instead of just blindly allocating
a large chunk of dummy memory.

### DIFF
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -5,24 +5,39 @@
 #include "common/assert.h"
 #include "common/common_funcs.h"
 #include "common/logging/log.h"
+#include "common/make_unique.h"
 
 #include "core/hle/kernel/process.h"
 #include "core/hle/kernel/resource_limit.h"
 #include "core/hle/kernel/thread.h"
+#include "core/hle/kernel/vm_manager.h"
+#include "core/mem_map.h"
 #include "core/memory.h"
 
 namespace Kernel {
 
+SharedPtr<CodeSet> CodeSet::Create(std::string name, u64 program_id) {
+    SharedPtr<CodeSet> codeset(new CodeSet);
+
+    codeset->name = std::move(name);
+    codeset->program_id = program_id;
+
+    return codeset;
+}
+
+CodeSet::CodeSet() {}
+CodeSet::~CodeSet() {}
+
 u32 Process::next_process_id;
 
-SharedPtr<Process> Process::Create(std::string name, u64 program_id) {
+SharedPtr<Process> Process::Create(SharedPtr<CodeSet> code_set) {
     SharedPtr<Process> process(new Process);
 
-    process->name = std::move(name);
-    process->program_id = program_id;
-
+    process->codeset = std::move(code_set);
     process->flags.raw = 0;
     process->flags.memory_region = MemoryRegion::APPLICATION;
+    process->address_space = Common::make_unique<VMManager>();
+    Memory::InitLegacyAddressSpace(*process->address_space);
 
     return process;
 }
@@ -87,8 +102,19 @@ void Process::ParseKernelCaps(const u32* kernel_caps, size_t len) {
     }
 }
 
-void Process::Run(VAddr entry_point, s32 main_thread_priority, u32 stack_size) {
-    Kernel::SetupMainThread(entry_point, main_thread_priority);
+void Process::Run(s32 main_thread_priority, u32 stack_size) {
+    auto MapSegment = [&](CodeSet::Segment& segment, VMAPermission permissions, MemoryState memory_state) {
+        auto vma = address_space->MapMemoryBlock(segment.addr, codeset->memory,
+                segment.offset, segment.size, memory_state).Unwrap();
+        address_space->Reprotect(vma, permissions);
+    };
+
+    MapSegment(codeset->code,   VMAPermission::ReadExecute, MemoryState::Code);
+    MapSegment(codeset->rodata, VMAPermission::Read,        MemoryState::Code);
+    MapSegment(codeset->data,   VMAPermission::ReadWrite,   MemoryState::Private);
+
+    address_space->LogLayout();
+    Kernel::SetupMainThread(codeset->entrypoint, main_thread_priority);
 }
 
 Kernel::Process::Process() {}
--- a/src/core/loader/3dsx.cpp
+++ b/src/core/loader/3dsx.cpp
@@ -19,7 +19,7 @@
 
 namespace Loader {
 
-/**
+/*
  * File layout:
  * - File header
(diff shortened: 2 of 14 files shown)


CMakeLists.txt
```diff
@@ -22,32 +22,34 @@ else()
 
     # set up output paths for executable binaries (.exe-files, and .dll-files on DLL-capable platforms)
     set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
-    set(CMAKE_CONFIGURATION_TYPES Debug Release RelWithDebInfo CACHE STRING "" FORCE)
```
Subv: Wat

yuriks: Oops, looks like some stuff leaked from #909

src/core/hle/kernel/process.cpp
```diff
@@ -5,24 +5,39 @@
...
 #include "core/hle/kernel/process.h"
 #include "core/hle/kernel/resource_limit.h"
 #include "core/hle/kernel/thread.h"
+#include "core/hle/kernel/vm_manager.h"
+#include "core/mem_map.h"
 #include "core/memory.h"
 
 namespace Kernel {
 
+SharedPtr<CodeSet> CodeSet::Create(std::string name, u64 program_id) {
```
Subv: const reference for the string

src/core/loader/3dsx.cpp
```diff
@@ -202,13 +205,29 @@ static THREEDSX_Error Load3DSXFile(FileUtil::IOFile& file, u32 base_addr)
     }
 
     // Write the data
-    memcpy(Memory::GetPointer(base_addr), &all_mem[0], loadinfo.seg_sizes[0] + loadinfo.seg_sizes[1] + loadinfo.seg_sizes[2]);
+    //memcpy(Memory::GetPointer(base_addr), program_image.data(), program_image.size());
```
Subv: Please don't leave dead code in comments

src/core/loader/3dsx.cpp
```diff
@@ -202,13 +205,29 @@ static THREEDSX_Error Load3DSXFile(FileUtil::IOFile& file, u32 base_addr)
...
+    SharedPtr<CodeSet> code_set = CodeSet::Create("", 0);
```
Subv: Can't we give this CodeSet a name? something like "Homebrew" or something

src/core/hle/kernel/process.cpp
yuriks: The function moves out of the string, so this might potentially elide an unecessary copy. See [Want Speed? Pass by Value.](https://web.archive.org/web/20140205194657/http://cpp-next.com/archive/2009/08/want-speed-pass-by-value/)

src/core/loader/3dsx.cpp
yuriks: Oops, it was supposed to be set from the filename in `Load()`, guess I must've forgotten to while moving code.

src/core/loader/3dsx.cpp
```diff
@@ -19,7 +19,7 @@
 
 namespace Loader {
 
-/**
```
archshift: Why?

yuriks: This wasn't really an appropriate doc-comment for the `THREEDSX_Error` enum that followed it.

src/core/loader/elf.cpp
```diff
@@ -266,19 +275,56 @@ void ElfReader::LoadInto(u32 vaddr) {
...
         if (p->p_type == PT_LOAD) {
-            segment_addr[i] = base_addr + p->p_vaddr;
-            memcpy(Memory::GetPointer(segment_addr[i]), GetSegmentPtr(i), p->p_filesz);
-            LOG_DEBUG(Loader, "Loadable Segment Copied to %08x, size %08x", segment_addr[i],
-                      p->p_memsz);
+            CodeSet::Segment* codeset_segment;
+            u32 permission_flags = p->p_flags & (PF_R | PF_W | PF_X);
+            if (permission_flags == (PF_R | PF_X)) {
+                codeset_segment = &codeset->code;
+            } else if (permission_flags == (PF_R)) {
```
archshift: Perhaps use a `switch` block here?

yuriks: I tried both after you posted this and I think prefer the if/else tree overall, so I'll keep it like this unless there are strong objections.

src/core/mem_map.h
```diff
@@ -6,9 +6,14 @@
 
 #include "common/common_types.h"
 
+namespace Kernel {
+    class VMManager;
```
bunnei: Indentation

src/core/hle/kernel/process.cpp
```diff
@@ -5,24 +5,39 @@
...
+    SharedPtr<CodeSet> codeset(new CodeSet);
+
+    codeset->name = std::move(name);
+    codeset->program_id = program_id;
```
Subv: Is the program_id stored in the CodeSet in the 3DS's Kernel?

src/core/loader/elf.cpp
```diff
@@ -267,19 +276,56 @@ void ElfReader::LoadInto(u32 vaddr) {
...
                   p->p_filesz, p->p_memsz);
 
         if (p->p_type == PT_LOAD) {
-            segment_addr[i] = base_addr + p->p_vaddr;
-            memcpy(Memory::GetPointer(segment_addr[i]), GetSegmentPtr(i), p->p_filesz);
-            LOG_DEBUG(Loader, "Loadable Segment Copied to %08x, size %08x", segment_addr[i],
-                      p->p_memsz);
+            CodeSet::Segment* codeset_segment;
+            u32 permission_flags = p->p_flags & (PF_R | PF_W | PF_X);
+            if (permission_flags == (PF_R | PF_X)) {
```
Subv: Is there any possibility of there being multiple segments with these flags and not just one (R, RX, RW)?

src/core/hle/kernel/process.cpp
yuriks: Yes: http://3dbrew.org/wiki/KCodeSet
It's also part of the info struct passed to the codeset creation syscall: http://3dbrew.org/wiki/SVC#struct_CodeSetInfo

Subv:
LGTM

src/core/loader/elf.cpp
yuriks: Yes, but it would be unusual, and there wouldn't be a good way of handling them either way.

yuriks: I added a check and a message in case this situation occurs.

PR #914, opened 2015-07-10
Common: Fix mask generation in BitField

### LINKED ISSUE #913: BitField does not work for raw types smaller than 32-bit
Currently, the BitField::GetMask code is:
return ((~(StorageTypeU)0) >> (8 * sizeof(T)-bits)) << position;

It's intended that the result of the "~" will be of type StorageTypeU which will then be shifted right logically by some amount.
However, if StorageTypeU is u16 or u8, the result of the "~" will be of type int, will contain ones in the upper bytes, and will be shifted right arithmetically.

To see this, you can step into the GetMask of a u16/u8 bitfield, or run this stand-alone program which simulates the result of BitField<0, 1, u16>::GetMask:
int result = ((~(unsigned short)  0) >> (8 * sizeof (unsigned short) - 1)) << 0;
printf ("%x\n", result);
It will be 0xffffffff instead of the "expected" 1.

One possible solution may be:
return (((StorageTypeU)~(StorageTypeU)0) >> (8 * sizeof(T)-bits)) << position;

yuriks:
Fixes #913

cc @neobrain @condut

### DIFF
--- a/src/common/bit_field.h
+++ b/src/common/bit_field.h
@@ -188,7 +188,7 @@ private:
 
     __forceinline StorageType GetMask() const
     {
-        return ((~(StorageTypeU)0) >> (8 * sizeof(T)-bits)) << position;
+        return (((StorageTypeU)~0) >> (8 * sizeof(T)-bits)) << position;
     }
 
     StorageType storage;
--- a/src/common/bit_field.h
+++ b/src/common/bit_field.h
@@ -160,7 +160,7 @@ public:
         if (std::numeric_limits<T>::is_signed)
         {
             std::size_t shift = 8 * sizeof(T)-bits;
-            return (T)(((storage & GetMask()) << (shift - position)) >> shift);
+            return (T)((storage << (shift - position)) >> shift);
         }
         else
         {


src/common/bit_field.h
```diff
@@ -188,7 +188,7 @@ struct BitField
 
     __forceinline StorageType GetMask() const
     {
-        return ((~(StorageTypeU)0) >> (8 * sizeof(T)-bits)) << position;
+        return ((StorageTypeU)~0) >> (8 * sizeof(T)-bits) << position;
```
neobrain: This will break the `StorageTypeU=u64` case, won't it? (that case was the original motivation for casting first, then inverting)

yuriks: No: `~0` will be sign extended:
```
[17:23:40] <yuriks> { uint64_t x = (uint64_t)~0; cout << hex << x; }
[17:23:41] <geordi> ffffffffffffffff
[17:23:46] <yuriks> << sizeof(int)
[17:23:47] <geordi> 4
```

src/common/bit_field.h
```diff
@@ -160,7 +160,7 @@ struct BitField
         if (std::numeric_limits<T>::is_signed)
         {
             std::size_t shift = 8 * sizeof(T)-bits;
-            return (T)(((storage & GetMask()) << (shift - position)) >> shift);
+            return (T)(storage << (shift - position) >> shift);
```
lioncash: Maybe parenthesize this so the shift precedence/intent is immediately clear.

neobrain:
Would be great if this PR could be forwarded to dolphin-emu/dolphin , along with unit tests to back this change. Other than that, it's probably good to go either way.

yuriks:
@lioncash said he's going to forward it to Dolphin.

I added explicit parenthesis around the shifts.

PR #915, opened 2015-07-10
Travis: Remove use of sudo to enable use container-based builds

yuriks:
The expectation was that this would improve build times but they seem to have stayed about the same.

"Most projects see an improvement in build times. We’d love to hear from you if you don’t.", if someone cares enough to contact them about it.

Branch build is green: https://travis-ci.org/citra-emu/citra/builds/70465190


archshift:
LGTM

bunnei:
LGTM

PR #919, opened 2015-07-11
Update style rules with to better match current practice.

yuriks:


CONTRIBUTING.md
```diff
@@ -7,19 +7,14 @@ Citra is a brand new project, so we have a great opportunity to keep things clea
 * Don't ever introduce new external dependencies into Core
 * Don't use any platform specific code in Core
 * Use namespaces often
+* Avoid the use of C-style casts and instead prefer C++-style `static_cast` and `reinterpret_cast`. Never use `const_cast` or `dynamic_cast` (we build with RTTI disabled). The only exception to this rule is for casting between two numeric types, where C-style casts are encouraged for brevity and readability.
```
neobrain: Why not const_cast ?

yuriks: I don't think I've ever seen a case where it wasn't either just a lazy hack or didn't cause UB.

CONTRIBUTING.md
```diff
@@ -101,11 +99,7 @@ void FooBar() {
         break;
     }
 
-    std::vector
-        you_can_declare,
-        a_few,
-        variables,
-        like_this;
+    std::vector<T> you_can_declare, a_few, variables, like_this;
```
neobrain: Is there any point in mentioning this explicitly anymore?

CONTRIBUTING.md
neobrain: Mhm I don't think it quite makes sense to overgeneralize this to be a coding style rule, though. If it ends up being used incorrectly, that should be pointed out on a case-by-case basis.

neobrain:
LGTM, although I don't think the const_cast rule is necessary or sensible.

CONTRIBUTING.md
yuriks: Well, it's explicitly mentioned at the top that there are always exceptions. Point is, if you're using const_cast, chances are high that you're not on the right track.

CONTRIBUTING.md
yuriks: There isn't a multiple-variable declaration example anywhere else.

lioncash:
LGTM

bunnei:
LGTM. Agree with @yuriks on the subject of const_cast .

PR #920, opened 2015-07-12
Remove unused Copy, CopyDir and GetThemeDir functions from FileUtil

linkmauve:


yuriks:
Is there any chance we might want to use these to implement some functions in FS?

archshift:
`Copy` and `CopyDir` could definitely come in handy in the future, IMO. I don't think they should be removed.

mailwl:
Agreed. emulator must be similar to real hardware

lioncash:
That has nothing to do with emulation itself lol

mailwl:
Ah! sorry. i think the talk about fs service

neobrain:
Ew, the implementation of these looks kind of ugly to begin with. But they will sooner or later be needed, and if they are deemed too ugly they can still be rewritten at that point.

neobrain:
That said, feel free to reopen this PR with only GetThemeDir removed.

PR #921, opened 2015-07-12
Fix applet includes using iwyu

linkmauve:
`std::shared_ptr` wasn’t included from `<memory>` as it should have.

I also fixed the other includes related to the applet.


bunnei:
LGTM

PR #924, opened 2015-07-13
Qt: Fix disassembly widget stepping

### LINKED ISSUE #923: citra-qt: cannot step instructions in the Disassembly widget
**Issue**: 
- Load a simple .3dsx file, e.g. the ```read-controls.3dsx``` example from the ctrulib library
- Open the Disassembly widget
- Pause the game using the ```Pause``` button
- Press the ```Step``` button

**What goes wrong**
- The disassembler does not step the instruction, nothing changes in the Disassembly widget

**What should happen**
- The disassembler should step the instruction and the GUI should highlight the next instruction

aroulin:
Should Fix #923.

Not sure if I need to lock on ```running_mutex``` and even why this lock is needed in first place.

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -65,7 +65,7 @@ void EmuThread::run() {
             was_active = false;
         } else {
             std::unique_lock<std::mutex> lock(running_mutex);
-            running_cv.wait(lock, [this]{ return IsRunning() || stop_run; });
+            running_cv.wait(lock, [this]{ return IsRunning() || exec_step || stop_run; });
         }
     }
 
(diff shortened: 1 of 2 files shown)


Subv:
Where is this variable reset to false?

aroulin:
Here:
https://github.com/aroulin/citra/blob/qt-disassembly-step/src/citra_qt/bootmanager.cpp#L60

Subv:
LGTM

jroweboy:
I added the lock originally as part of #783 to fix the spin wait that used to happen when you would pause the emulator. 

According to http://en.cppreference.com/w/cpp/thread/condition_variable

> Any thread that intends to wait on std::condition_variable has to acquire a std::unique_lock first. The wait operations atomically release the mutex and suspend the execution of the thread. When the condition variable is notified, the thread is awakened, and the mutex is reacquired.

aroulin:
@jroweboy Thanks for the reference! It is not clear for me whether it is needed to lock the mutex when setting ```exec_step```.

yuriks:
LGTM

PR #927, opened 2015-07-13
getopt: Fix compilation settings

yuriks:
INTERFACE doesn't define the symbol when compiling the library itself.
PUBLIC should be used when the definition is needed both by the library
and by the users.


archshift:
LGTM

Issue #926, opened 2015-07-13
Pre-commit hook causes CMake error


neobrain:
The pre-commit hook introduced in  772ef097ead8d6be77cf00d753e8380fcac96710 is causing issues for me. Specifically, every time the CMakeLists.txt is processed with a pre-commit file already existing (e.g. because CMakeLists.txt had already been processed before, and is being re-read as part of calling `make`), I get the following error:

    CMake Error at CMakeLists.txt:9 (file):
      file COPY cannot set permissions on
      "~/citra/.git/hooks/pre-commit"

Further info: Upon trying to `rm` the file, a confirmation question is asked saying `rm: remove write-protected regular file ‘/home/tony/citra4/.git/hooks/pre-commit’?`. Upon answering `y`, removing the file works, of course.

neobrain:
CC @Lectem

yuriks:
What's your CMake version?

neobrain:
2.8.11

neobrain:
Apparently this is caused because improper permissions are being set on the target file. The file ends up having no permission flags set at all other than the "executable" flag.

Lectem:
Windows or Linux ? It is a bit weird because it should be ignored if a file is already present : 

    if(NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git/hooks/pre-commit)

yuriks:
My guess is that CMake checks for existence by trying to read the file (which fails since it has no read permissions.)
```
[18:25:59] <yuriks> COPY preserves source permissions, so I think what should be done is add +x to the template script and simply leave out the permissions part of the command
```

Lectem:
Could you try adding  WORLD_READ WORLD_WRITE  to FILE_PERMISSIONS on https://github.com/citra-emu/citra/blob/master/CMakeLists.txt#L11 ?

yuriks:
@Lectem: This will still leave the file with wonky permissions. I think just letting COPY copy the perms from the original file is a better choice.

Lectem:
@yuriks Thing is, git only sets the +x flag, or we could use SETUID.

neobrain:
[...]

PR #928, opened 2015-07-13
Fix permissions in pre-commit hook

JSFernandes:
This PR is related with the following issue: https://github.com/citra-emu/citra/issues/926
Since the pre-commit hook was being copied with the incorrect permissions, any attempt to run `cmake` beyond the first would fail, as it would be unable to read the pre-commit hook.


yuriks:
:-1: We should just add +x to the file in Git and then let CMake copy it without altering permissions.

JSFernandes:
@yuriks That sounds good too, since it means that changing permissions on the file won't be a "trap" overriden by CMake, but can you explain any other advantages you see?

neobrain:
Thanks!

PR #929, opened 2015-07-15
Pica/Shader: Add geometry shader definitions.

neobrain:
Just register definitions, nothing else.

VS and GS registers share the same layout, so this is mostly restructuring and renaming. Furthermore, VertexShader interfaces were changed to make less assumptions about the shader pipeline configuration (i.e. VS<->GS).

If you agree that GitHub is doing a terrible job at diffing this, try `git diff --word-diff=color --ignore-all-space` instead.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -784,112 +784,119 @@ struct Regs {
     INSERT_PADDING_WORDS(0x20);
 
     enum class TriangleTopology : u32 {
-        List        = 0,
-        Strip       = 1,
-        Fan         = 2,
-        ListIndexed = 3, // TODO: No idea if this is correct
+        List   = 0,
+        Strip  = 1,
+        Fan    = 2,
+        Shader = 3, // Programmable setup unit implemented in a geometry shader
     };
 
     BitField<8, 2, TriangleTopology> triangle_topology;
 
-    INSERT_PADDING_WORDS(0x51);
+    INSERT_PADDING_WORDS(0x21);
 
-    BitField<0, 16, u32> vs_bool_uniforms;
-    union {
-        BitField< 0, 8, u32> x;
-        BitField< 8, 8, u32> y;
-        BitField<16, 8, u32> z;
-        BitField<24, 8, u32> w;
-    } vs_int_uniforms[4];
-
-    INSERT_PADDING_WORDS(0x5);
-
-    // Offset to shader program entry point (in words)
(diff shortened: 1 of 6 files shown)


neobrain:
Formal request for more reviews. If nothing comes up by later today, I'll assume it's fine and hit the merge button.

src/video_core/pica.h
```diff
@@ -784,112 +784,119 @@ struct Regs {
...
+            BitField<36, 4, u64> attribute9_register;
+            BitField<40, 4, u64> attribute10_register;
+            BitField<44, 4, u64> attribute11_register;
+            BitField<48, 4, u64> attribute12_register;
+            BitField<52, 4, u64> attribute13_register;
+            BitField<56, 4, u64> attribute14_register;
+            BitField<60, 4, u64> attribute15_register;
+
+            int GetRegisterForAttribute(int attribute_index) const {
+                u64 fields[] = {
```
Subv: If we change this (and the other GetRegisterForAttribute) to manipulate the bits in the u64 directly using shifts, we could get rid of the if-tables without any performance hit.

neobrain: Sorry, but what are you talking about specifically? What if-tables? How do they relate to using shifts here instead?

Subv: I'm talking about https://github.com/citra-emu/citra/pull/933/files#diff-14f7afb9f55448d7a4f62057813d32f4R619
As @yuriks once pointed out, the current GetRegisterForAttribute does hurt performance because of the local array construction. You can get rid of that performance hit by manipulating the bits directly from the "hex" member of the union, and then just use GetRegisterForAttribute without any worries instead of that cascade of ifs

yuriks:
LGTM. What @Subv said is true but not *really* relevant to the PR, so I think it can just come in a follow-up by someone.

bunnei:
LGTM

mailwl:
as all registers described at 3dbrew, why not replace INSERT_PADDING_WORDS with real names for all?
i started do it, look http://pastebin.com/EgQayTKP, but now see no point any more
i may help with debugging purposes, to see, what exactly writes in unused right now registers

neobrain:
@mailwl Keeping placeholders makes it clear what's still not emulated, and official register names aren't necessarily good names (and sometimes rather misleading in fact).

neobrain:
Agreeing that @Subv's comment is not crucial to this PR and can be considered for a future PR instead.

PR #930, opened 2015-07-15
citra-qt: Add support for copying the command list contents to clipboard.

neobrain:
Title says it all. This is useful for analyzing command lists more efficiently, e.g. by searching for specific commands.

### DIFF
--- a/src/citra_qt/debugger/graphics_cmdlists.cpp
+++ b/src/citra_qt/debugger/graphics_cmdlists.cpp
@@ -2,6 +2,8 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <QApplication>
+#include <QClipboard>
 #include <QLabel>
 #include <QListView>
 #include <QMainWindow>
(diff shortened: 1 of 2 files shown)


src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -330,3 +340,21 @@ void GPUCommandListWidget::OnToggleTracing() {
         toggle_tracing->setText(tr("Start Tracing"));
     }
 }
+
+void GPUCommandListWidget::CopyAllToClipboard() {
+    QClipboard* clipboard = QApplication::clipboard();
+    QString text;
+
+    QAbstractItemModel* model = (QAbstractListModel*)list_widget->model();
```
Subv: static_cast

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -330,3 +340,21 @@ void GPUCommandListWidget::OnToggleTracing() {
...
+
+    for (int row = 0; row < model->rowCount({}); ++row) {
+        for (int col = 0; col < model->columnCount({}); ++col) {
+            QModelIndex index = model->index(row, col);
+            text += model->data(index).value<QString>();
+            text += '\t';
+        }
+        text += '\n';
```
Subv: Does this work on Windows, where the default newline is \r\n?

neobrain: It probably does. Anyone mind double-checking?

archshift: OS X uses just a simple line feed, no carriage return.

neobrain: @archshift That's the theory, does Qt actually follow that standard for texts in the GUI, though?

neobrain: Erm, nevermind. I was mixing up stuff here. This will need to be taken care of indeed.

neobrain: Oh well. Could any of you guys still confirm that this is actually broken?

Subv: I'll do it later today and report back

archshift:
Is there a way to also allow copying through Ctrl/Cmd+C?

neobrain:
@archshift You would need to inherit a custom widget from QTreeView and reimplement it's keyPressEvent, in which you can then use `QKeySequence(QKeySequence::Copy)` to check for such key presses. Not sure yet if I'll implement that in this PR or keep it for a followup one.

src/citra_qt/debugger/graphics_cmdlists.cpp
Subv: It works

neobrain:
Formal request for more reviews. If nothing comes up by later today, I'll assume it's fine as is, address the outstanding comments, and hit the merge button.

src/citra_qt/debugger/graphics_cmdlists.cpp
neobrain: @Subv Thanks for verifying. I'll just assume OSX works too until reported otherwise.

Subv:
LGTM after the comments are addressed

bunnei:
LGTM

PR #931, opened 2015-07-15
Pica/CommandProcessor: Move default attribute setup to the proper position.

neobrain:
The handlers are sorted by register index for improved readability.


archshift:
LGTM

PR #933, opened 2015-07-15
Shader debugger improvements

neobrain:
A number of functional improvements to the shader disassembler (which is now an actual debugger):

* When the emulator halts at a VertexLoaded breakpoint, the vertex attributes are shown in the shader debugger
* Unreached instructions are colored yellow in the shader disassembly
* Shaders can be dumped from the GUI
* More output semantics than before are recognized when dumping shaders
* Each shader interpreter cycle records its inputs and outputs now, so that they can be inspected in the GUI (think of this as stepping through the shader and inspecting data, just that it's all done automatically and you just select a particular "iteration" to analyze in the GUI)

Obligatory screenshot:
Screenshot of shader debugger GUI

This PR pends on #929 .

### DIFF
--- a/src/citra_qt/debugger/graphics_vertex_shader.cpp
+++ b/src/citra_qt/debugger/graphics_vertex_shader.cpp
@@ -272,10 +272,9 @@ void GraphicsVertexShaderModel::DumpShader() {
     auto& setup  = Pica::g_state.vs;
     auto& config = Pica::g_state.regs.vs;
 
-    Pica::DebugUtils::DumpShader(setup.program_code.data(), setup.program_code.size(),
-                                 setup.swizzle_data.data(), setup.swizzle_data.size(),
-                                 config.main_offset, Pica::g_state.regs.vs_output_attributes);
+    Pica::DebugUtils::DumpShader(config, setup, Pica::g_state.regs.vs_output_attributes);
 }
+
 GraphicsVertexShaderWidget::GraphicsVertexShaderWidget(std::shared_ptr< Pica::DebugContext > debug_context,
                                                        QWidget* parent)
         : BreakPointObserverDock(debug_context, "Pica Vertex Shader", parent) {
--- a/src/citra_qt/debugger/graphics_vertex_shader.cpp
+++ b/src/citra_qt/debugger/graphics_vertex_shader.cpp
@@ -6,11 +6,16 @@
 #include <sstream>
 
 #include <QBoxLayout>
+#include <QFileDialog>
+#include <QGroupBox>
 #include <QLabel>
+#include <QLineEdit>
 #include <QPushButton>
+#include <QSignalMapper>
+#include <QSpinBox>
 #include <QTreeView>
 
-#include "video_core/shader/shader_interpreter.h"
+#include "video_core/shader/shader.h"
 
 #include "graphics_vertex_shader.h"
 
@@ -19,7 +24,7 @@ using nihstro::Instruction;
 using nihstro::SourceRegister;
 using nihstro::SwizzlePattern;
 
-GraphicsVertexShaderModel::GraphicsVertexShaderModel(QObject* parent): QAbstractItemModel(parent) {
+GraphicsVertexShaderModel::GraphicsVertexShaderModel(GraphicsVertexShaderWidget* parent): QAbstractItemModel(parent), par(parent) {
 
 }
 
@@ -36,7 +41,7 @@ int GraphicsVertexShaderModel::columnCount(const QModelIndex& parent) const {
 }
 
 int GraphicsVertexShaderModel::rowCount(const QModelIndex& parent) const {
-    return static_cast<int>(info.code.size());
+    return static_cast<int>(par->info.code.size());
 }
 
 QVariant GraphicsVertexShaderModel::headerData(int section, Qt::Orientation orientation, int role) const {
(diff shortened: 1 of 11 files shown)


Subv:
Could we get some kind of indication about which attributes came from defaults and which ones came from buffer loaders please?

Subv:
Does this allow modifying the shader code on the fly?

neobrain:
@Subv Shader modification will not be possible too soon. As for the default attribute suggestion, that sounds like a good idea and I'll check whether it's feasible to implement or not.

Subv:
Adding to that, from your screenshot i can't see how one knows how many iterations a particular instruction has been through, does it say that anywhere?

neobrain:
@Subv Changing the iteration count selects (i.e. as if the user clicked on it) the instruction processed at that iteration. In the future, the disassembly viewer could be modified to add a small icon next to the "current" instruction. An earlier prototype of the shader debugger just displayed information for the currently selected instruction, but unfortunately that's ambiguous when the instruction might be processed more than once (e.g. in loops or helper functions). Hence the somewhat artificial distinction between "iterations" and "instructions".

Instruction offset refers to the "offset" column of the shader disassembly view. It's the offset of a particular instruction from the start of the shader.

DEST_IN and DEST_OUT account for swizzling and negation.

I'm open for concrete proposals for improvement on the GUI, but for now I consider the added functionality more important than fancy appearance. I'm personally not entirely sure how one would expose the information given in the description text in a nicer way.

src/citra_qt/debugger/graphics_vertex_shader.cpp
```diff
@@ -17,7 +24,7 @@ using nihstro::Instruction;
 using nihstro::SourceRegister;
 using nihstro::SwizzlePattern;
 
-GraphicsVertexShaderModel::GraphicsVertexShaderModel(QObject* parent): QAbstractItemModel(parent) {
+GraphicsVertexShaderModel::GraphicsVertexShaderModel(GraphicsVertexShaderWidget* parent): QAbstractItemModel(parent), par(parent) {
```
Subv: Any reason why the par variable isn't called parent?

src/citra_qt/debugger/graphics_vertex_shader.h
```diff
@@ -43,9 +49,41 @@ private slots:
     void OnBreakPointHit(Pica::DebugContext::Event event, void* data) override;
     void OnResumed() override;
 
+    void OnInputAttributeChanged(int index);
+
+    void OnIterationIndexChanged(int index);
+
+    void DumpShader();
+
+    /** Reload widget based on the current PICA200 state
```
Subv: Comments should begin on the next line

src/citra_qt/debugger/graphics_vertex_shader.h
```diff
@@ -43,9 +49,41 @@ private slots:
...
+    // Call this to change the current command selection in the disassembly view
+    void SelectCommand(const QModelIndex&, QItemSelectionModel::SelectionFlags);
 
 private:
+    QLabel* instruction_description;
+    QTreeView* binary_list;
+    GraphicsVertexShaderModel* model;
+
+    // TODO: Move these into a single struct
+    std::array<QLineEdit*, 4*16> input_data;  // A text box for each of the 4 components of up to 16 vertex attributes
```
Subv: ///

src/citra_qt/debugger/graphics_vertex_shader.cpp
neobrain: The base class already has a member function called parent(), and I couldn't find a different good name :/

Subv: any reason why you can't use the member of the base class?

neobrain: That would involve `static_cast` for each use of that member, since the base class only returns a pointer of type `QObject*`.

src/citra_qt/debugger/graphics_vertex_shader.cpp
```diff
@@ -34,7 +41,7 @@ int GraphicsVertexShaderModel::columnCount(const QModelIndex& parent) const {
 }
 
 int GraphicsVertexShaderModel::rowCount(const QModelIndex& parent) const {
```
chinhodado: RowCount?

src/citra_qt/debugger/graphics_vertex_shader.cpp
```diff
@@ -34,7 +41,7 @@ int GraphicsVertexShaderModel::columnCount(const QModelIndex& parent) const {
...
-    return info.code.size();
+    return par->info.code.size();
 }
 
 QVariant GraphicsVertexShaderModel::headerData(int section, Qt::Orientation orientation, int role) const {
```
chinhodado: HeaderData?

src/citra_qt/debugger/graphics_vertex_shader.cpp
```diff
@@ -247,53 +266,231 @@ QVariant GraphicsVertexShaderModel::data(const QModelIndex& index, int role) con
...
+
+    for (unsigned i = 0; i < ARRAY_SIZE(input_data); ++i) {
+        connect(input_data[i], SIGNAL(textEdited(const QString&)), input_data_mapper, SLOT(map()));
+        input_data_mapper->setMapping(input_data[i], i);
+    }
+    connect(input_data_mapper, SIGNAL(mapped(int)), this, SLOT(OnInputAttributeChanged(int)));
 
     auto main_widget = new QWidget;
     auto main_layout = new QVBoxLayout;
     {
```
chinhodado: Why are these blocks needed?

src/citra_qt/debugger/graphics_vertex_shader.h
```diff
@@ -43,9 +49,41 @@ private slots:
     void OnBreakPointHit(Pica::DebugContext::Event event, void* data) override;
     void OnResumed() override;
 
+    void OnInputAttributeChanged(int index);
+
```
chinhodado: Why is there a new line after each of these?

src/citra_qt/debugger/graphics_vertex_shader.cpp
neobrain: This is the implementation of a (Qt) base class method.

src/citra_qt/debugger/graphics_vertex_shader.cpp
neobrain: This is the implementation of a (Qt) base class method.

src/citra_qt/debugger/graphics_vertex_shader.cpp
neobrain: They are not needed, but they logically separate the individual sub-layouts from each other, making sure that widgets are attached to the intended sublayout. We use this pattern throughout the Qt code base.

With any alternative method, it might be easy to accidently put a widget into the wrong sublayout.

src/citra_qt/debugger/graphics_vertex_shader.h
neobrain: There is no contextual connection between these three functions (contrary to OnBreakPointHit and OnResumed, both of which deal with GPU breakpoints).

src/citra_qt/debugger/graphics_vertex_shader.cpp
Subv: These are overrides from some Qt base class

neobrain:
Formal request for more reviews. If nothing comes up by later today, I'll assume it's fine as is, and will merge it after addressing the outstanding comments.

neobrain:
For reference: It might be a good idea to make the input vertex data collapsable in some way, and to display the output vertex data in a similar fashion.

neobrain:
Addressed comments (the new patches don't need squashing), and rebased on latest master.

QToolBox doesn't support the UI I had in mind, hence none of the contents are expandable, unfortunately.

Issue #840, opened 2015-05-31
MH4U: Unsupported thread processor ID


mhvuze:
Only experienced this in Monster Hunter 4 Ultimate and the demo version of the game so far, outright crashes Citra. If you let me know how I can provide more info, I'll gladly do so.

Here's the log: http://pastie.org/10216479

Subv:
This happens because MH4U believes that Citra is a New Nintendo 3DS. This can be fixed by stubbing the APT command 0x01010000

ignika98:
Is that something we can fix ourselves when compiling, or does it need to be done by you guys?

yuriks:
It's fixable if you know programming, but it's not trivial. You need to add the missing function as indicated by @Subv.

ignika98:
Oh, well I guess I'll just have to wait then.

Zaneo:
@ignika98 If you get a chance could you try this? [Zaneo/citra/tree/apt_ptm_checknew3ds](https://github.com/Zaneo/citra/tree/apt_ptm_checknew3ds)

ignika98:
![capture](https://cloud.githubusercontent.com/assets/2106615/8667687/b3de7f22-2a3d-11e5-9ba2-b20e3a794a6c.JPG)
It still doesn't work, but I get a different error this time. Unless I did something wrong...

yuriks:
Apparently it's still not quite working, since it's trying to create a thread for one of the n3DS's new CPU cores.

yuriks:
Hm, looking again at the screenshot, it doesn't look like the patch worked correctly, since it's giving unknown function error.

ignika98:
Guess I'll just have to wait again then...

yuriks:
Should be a matter of copying this line: https://github.com/Zaneo/citra/commit/982687f70d97a07b861b8b844c20ec4591c15fe6#diff-1534728c6bee9cbcd50cbf1ba054ab63R94
and also adding it to the table in apt_a.cpp

ignika98:
Hmmmm, doing that gives me build errors, but I'm probably doing it wrong. How and where exactly would it be copied?

mhvuze:
Citra still crashes but I guess that's not related to the CPU anymore, huh?
![4e2c64e225](https://cloud.githubusercontent.com/assets/4907001/8668815/45480642-2a0f-11e5-97b6-0c48d4ecea96.png)

ignika98:
[...]

PR #934, opened 2015-07-16
Services: PTM::Added CheckNew3DS and ATP::Uknown_0x01010000 to help with #840

Zaneo:
Added stubbed ATP::Uknown_0x01010000
Added PTM::CheckNew3Ds


src/core/hle/service/apt/apt.cpp
```diff
@@ -327,6 +329,23 @@ void GetAppCpuTimeLimit(Service::Interface* self) {
...
 }
 
+void Unknown_0x01010000(Service::Interface* self){
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+
+    // When a certain NS state field (internal NS state in .data/.bss.) is non-zero, the output value is zero
+    // If/Else is to indicate that a check should occur here.
+    if (false){
```
lioncash: tbh I'm not really for half-implementing service functions like this.

src/core/hle/service/apt/apt.h
```diff
@@ -299,6 +299,15 @@ void PrepareToStartLibraryApplet(Service::Interface* self);
  */
 void StartLibraryApplet(Service::Interface* self);
 
+/**
+* From 3Ds Brew:
+*	This writes an output u8 to cmdreply indexword[2].
+*	This uses PTMSYSM:CheckNew3DS.
+*	When a certain NS state field (internal NS state in .data/.bss.) is non-zero, the output value is zero, otherwise the output is from PTMSYSM:CheckNew3DS. 
+*	Normally this NS state field is zero, however this state field is set to 1 when APT:PrepareToStartApplication is used with flags bit8 is set.
```
Zaneo: The pre-commit hook may have missed the trailing whitespcae?

src/core/hle/service/apt/apt.cpp
Zaneo: This was mostly to help move #840 along. It looked like this allowed that to progress further. I don't have any hardware to track down this state field, which is apparently title specific.

bunnei:
I'm curious if making this function say Citra is a N3DS helps ZMM along... I guess I'll test that later.

Subv:
I'm against this, stubbing these functions like this brings nothing useful. The real behavior should be investigated

neobrain:
Seconding @Subv's comment, especially since the PR description doesn't even describe why this change was deemed useful.

Also, it's "3DBrew", not "3Ds Brew".

Zaneo:
Not disagreeing, in this case is there much risk?

"When a certain NS state field is non-zero, the output value is zero, otherwise the output is from PTMSYSM:CheckNew3DS."

Currently Old3DS (Value = 0) is returned always, which means that the condition is always satisfied.

Maybe a comment to say that until the NS statefield is discovered it should return 0 instead of CheckNew3DS?

Subv:
I also don't like how this introduces yet another "CheckNew3DS" function, we ought to have a centralized point for all of this (see the CFG functions that also query the 3DS type)

Zaneo:
No rush.

yuriks:
For me it would be a safe bet to say that this "internal field" that's being talked about is some sort of flag that says that the process is a Old3DS compatibility-mode process. I think we can just implement this as returning false until we investigate more comprehensive n3DS support.

purpasmart96:
Plus all the formatting was all wrong.

PR #935, opened 2015-07-17
arm_dyncom_interpreter: Simplify assignment in SMLAW

lioncash:
Also a side-benefit of not having implementation-defined behavior.


Subv:
LGTM

archshift:
LGTM

PR #936, opened 2015-07-17
Update homebrew before building, which no longer uses SourceForge

archshift:
Also removes the `brew tap` command, which is irrelevant now that Travis taps `homebrew/versions` by default.


yuriks:
What about the old tap command? What did that do?

archshift:
The tap command was unnecessary because Travis now taps `homebrew/versions` itself. Unfortunately, this isn't working out the way I had hoped anyway :|

archshift:
There we go, now it works fine and doesn't depend on SourceForge at all!

PR #937, opened 2015-07-17
Ensure all kernel objects are released during shutdown

yuriks:
This commit fixes several kernel object leaks. The most severe of them
was threads not being removed from the private handle table used for
CoreTiming events. This resulted in Threads never being released, which
in turn held references to Process, causing CodeSets to never be freed
when loading other applications.


Subv:
LGTM instamerge

bunnei:
LGTM

PR #938, opened 2015-07-17
Kernel/SVC: Implemented svcQueryMemory.

Subv:


src/core/hle/svc.cpp
```diff
@@ -530,8 +531,19 @@ static ResultCode ReleaseSemaphore(s32* count, Handle handle, s32 release_count)
 }
 
 /// Query memory
-static ResultCode QueryMemory(void* info, void* out, u32 addr) {
-    LOG_ERROR(Kernel_SVC, "(UNIMPLEMENTED) called addr=0x%08X", addr);
+static ResultCode QueryMemory(MemoryInfo* memory_info, PageInfo* page_info, u32 addr) {
+    auto vma = Kernel::g_current_process->address_space->FindVMA(addr);
+
+    if (vma == Kernel::g_current_process->address_space->vma_map.end())
+        return ResultCode(-1); // TODO(Subv): Find right error code
```
yuriks: You have a demo already, should take almost no time to find this one. :P

yuriks:
Btw, this was tested with: https://github.com/yuriks/3ds-demos

Demo Run

yuriks:
Other than the error, LGTM

mailwl:
looks like games doesn't check actual result, so i think, the error result may be somethink like "1013 	Invalid address", or "1021 	Out of range "

bunnei:
LGTM

Subv:
Updated with the correct error code

PR #939, opened 2015-07-17
Kernel/SVC: Implemented svcQueryProcessMemory

Subv:


yuriks:
Meh, instead of duplicating the code, just have one func call the other with the process pseudo-handle.

yuriks:
LGTM

mailwl:
and another question, is it good idea to change R4-R5 registers, if SVC call returns error?
i understand the syscall not usual function, but i... don't know, i don't like it

yuriks:
@mailwl Since the caller can't know if the function will fail, it'll have saved the registers anyway, so there shouldn't be any harm.

bunnei:
LGTM

PR #940, opened 2015-07-18
Kernel: Properly implement ControlMemory FREE and COMMIT

yuriks:
This gets rid of another hard-coded part of the address space.


src/core/hle/kernel/process.cpp
```diff
@@ -28,6 +28,109 @@ SharedPtr<CodeSet> CodeSet::Create(std::string name, u64 program_id) {
...
+
+    CASCADE_RESULT(auto vma, vm_manager.MapMemoryBlock(target, heap_memory, target - heap_start, size, MemoryState::Private));
+    vm_manager.Reprotect(vma, perms);
+
+    return MakeResult<VAddr>(heap_end - size);
+}
+
+ResultCode ProcessMemoryManager::HeapFree(VAddr target, u32 size) {
+    if (target < Memory::HEAP_VADDR || target + size > Memory::HEAP_VADDR_END || target + size < target) {
+        return ERR_INVALID_ADDRESS;
```
archshift: Maybe add a log if this happens so we can know (if the game, for whatever reason, doesn't error-check this)?

src/core/hle/kernel/process.cpp
```diff
@@ -28,6 +28,109 @@ SharedPtr<CodeSet> CodeSet::Create(std::string name, u64 program_id) {
...
+
+    VAddr heap_end = Memory::LINEAR_HEAP_VADDR + (u32)linear_heap_memory->size();
+    // Games and homebrew only ever seem to pass 0 here (which lets the kernel decide the address),
+    // but explicit addresses are also accepted and respected.
+    if (target == 0) target = heap_end;
+
+    if (target < Memory::LINEAR_HEAP_VADDR || target + size > Memory::LINEAR_HEAP_VADDR_END ||
+        target > heap_end || target + size < target) {
+
+        return ERR_INVALID_ADDRESS;
```
archshift: Ditto.

src/core/hle/kernel/process.cpp
```diff
@@ -28,6 +28,109 @@ SharedPtr<CodeSet> CodeSet::Create(std::string name, u64 program_id) {
 CodeSet::CodeSet() {}
 CodeSet::~CodeSet() {}
 
+ResultVal<VAddr> ProcessMemoryManager::HeapAllocate(VAddr target, u32 size, VMAPermission perms) {
+    if (target < Memory::HEAP_VADDR || target + size > Memory::HEAP_VADDR_END || target + size < target) {
+        return ERR_INVALID_ADDRESS;
```
archshift: Ditto.

Subv:
I'm not sure if it's related to these changes, but any game that calls GetSharedFont will crash citra due to GetPointer returning nullptr

yuriks:
It sure is. `SHARED_FONT_VADDR` is 0x18000000, which is somewhere in the middle of the linear heap area, which now does not have a dummy mapping anymore. I was expecting the font to be mapped in the shared memory area at 0x10000000-0x14000000. We need to find out why it's positioned where it currently is.

yuriks:
I need to do some testing around this and the patch will need some impovements to properly support whatever's going on here, so I'm going to close it for now.

PR #941, opened 2015-07-18
Dyncom: Support for a new ARMv6 Thumb MOV encoding

yuriks:
cc @lioncash 

Ref. ARM sections A6.2.3 and A8.8.103.

The change to the `Rs` variable is cosmetic. Both versions of the code are equivalent.


yuriks:
Updated with a more general version.

lioncash:
Looks fine

yuriks:
@Normmatt also said it looks fine, so merging

PR #943, opened 2015-07-19
Rasterizer/Textures: Fixed a bug where the I4 format would get twice the real stride

Subv:
Also added its name to the texture viewer widget.

Fixes some artifacts in games that use this format, like Crashmo


yuriks:
LGTM

PR #947, opened 2015-07-19
Change trace/unimplemented service call logs to use hex

archshift:
Changes the log to use hex in the parameter list instead of decimal.


Subv:
LGTM

lioncash:
Looks good

yuriks:
decimal is for scrubs

PR #948, opened 2015-07-19
Pica: Fix DP3 instruction, which wasn't assigning to the w component

yuriks:
This fixes the OoT NaN/white/black triangle glitch.


neobrain:
s/DP3 instruction/dot product instructions/g

yuriks:
@neobrain For DP4 "num_components" is 4, so it was working correctly.

neobrain:
Derp. Let's forget I ever said anything about this ;p

PR #949, opened 2015-07-19
Pica: Correct switched S/T texture wrapping registers

yuriks:
This was found and hwtested by Lectem in https://github.com/Lectem/3DS_gpu_tests/tree/wrap_modes

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -132,8 +132,8 @@ struct Regs {
         };
 
         union {
-            BitField< 8, 2, WrapMode> wrap_s;
-            BitField<12, 2, WrapMode> wrap_t;
+            BitField< 8, 2, WrapMode> wrap_t;
+            BitField<12, 2, WrapMode> wrap_s;
         };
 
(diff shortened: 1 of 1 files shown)


neobrain:
Once ready, please provide a link to a hardware test homebrew app.

yuriks:
Added a link to Lectem's test.

neobrain:
@Lectem What revision of ctrulib was that test written against?

Lectem:
Well, it was built against https://github.com/smealum/ctrulib/commit/096bbdf40155fda036ba517a40f59c5796a8013a but https://github.com/smealum/ctrulib/commit/4109bf5b66d1d29ad816f327be49c0b82ec28f30 fixed it.

neobrain:
So if anyone wants to build the test, they should use smealum/ctrulib@096bbdf. Thanks.

yuriks:
While that's the case for verifying things against the future, any version if fine to just compare Citra vs. HW since the results will have to be consistent on both.

PR #957, opened 2015-07-20
Kernel/Scheduling: Clean up a thread's wait_objects when its scheduled.

Subv:
They'll be reset if needed during the next svcWaitSynchronization call (if there's any pending)


yuriks:
LGTM

bunnei:
LGTM

Issue #961, opened 2015-07-21
Harvest Moon The Lost Valley hangs at title screen


JohnGodgames:
![hmtlv](https://cloud.githubusercontent.com/assets/11453507/8792581/6eb22472-2f69-11e5-8b97-0425f00b43cb.PNG)

SHORTENED LOG: http://pastie.org/10303511

Zaneo:
Speaking with @Subv on IRC he mentioned this might be related to DLC. As am:app deals with that, and a similar issue has been seen in Steel Diver. Below you can see the menu as it should be rendered, with the DLC option.
![dlc](https://cloud.githubusercontent.com/assets/295949/8793353/7a82e398-2f43-11e5-9f6e-63cb291b0131.png)

JohnGodgames:
#962 Not working for me.Before menu it should be a Touch the touch screen

reference: https://www.youtube.com/watch?v=Yvr4fpvR0Ho

JohnGodgames:
Fixed with some recent updates (checking dlc content)

amirlou:
its got Fixed with new master but got buggy texture with jit enable and after i sleep for first time in game i get black screen

![xx2 2016-05-01 01-26-27-41](https://cloud.githubusercontent.com/assets/16171668/14938547/5f5b4964-0f3c-11e6-874d-8e20305fd8e5.jpg)
![xx2 2016-05-01 01-27-25-71](https://cloud.githubusercontent.com/assets/16171668/14938548/5f657bdc-0f3c-11e6-9133-c209386a355d.jpg)
![xx2 2016-05-01 01-27-35-33](https://cloud.githubusercontent.com/assets/16171668/14938549/5f6c0a24-0f3c-11e6-845c-20ea8d6387b7.jpg)

JayFoxRox:
@amirlou: If this is fixed we should close it, if this game hangs elsewhere now, it might be a different issue. If that's the case remove your post please and create a new issue.
If this is only a JIT issue it's important to know that - does it also have problems without JIT?

amirlou:
no it doesn't have texture problem without jit but if jit is enable game get texture problem with snow but game still freeze after first time sleeping i open a new issue with log

JohnGodgames:
Already fixed in recent master

PR #962, opened 2015-07-21
Services/AM: Stubbed am:app::GetNumContentInfos to return 0 results.

Subv:
Named the service functions in am:app as per 3dbrew.

This fixes an illegal read loop in Steel Diver

Related to #961

Pic for reference
Steel Diver

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -38,6 +38,15 @@ void GetTitleIDList(Service::Interface* self) {
     LOG_WARNING(Service_AM, "(STUBBED) Requested %u titles from media type %u", num_titles, media_type);
 }
 
+void GetNumContentInfos(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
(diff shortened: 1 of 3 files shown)


yuriks:
LGTM. Can confirm that it now boots properly.

src/core/hle/service/am/am.cpp
```diff
@@ -38,6 +38,15 @@ void GetTitleIDList(Service::Interface* self) {
     LOG_WARNING(Service_AM, "(STUBBED) Requested %u titles from media type %u", num_titles, media_type);
 }
 
+void GetNumContentInfos(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    cmd_buff[2] = 1; // Number of content infos plus one
```
bunnei: Can we document here what the heck a "content info" is?

Subv: I would if i knew

bunnei:
LGTM

PR #963, opened 2015-07-21
Misc. GPU vertex loading fixes

yuriks:
Two small GPU fixes. See individual commit messages for more info.

Notably, one of them fixes the UI in Fire Emblem, which is now playable:

fire_emblem

Test homebrew at:
 - https://github.com/yuriks/3DS_gpu_tests/tree/color-clamping
 - https://github.com/yuriks/3DS_gpu_tests/tree/vertex-loading

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -217,56 +217,52 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
                 // Initialize data for the current vertex
                 VertexShader::InputVertex input;
 
-                // Load a debugging token to check whether this gets loaded by the running
-                // application or not.
-                static const float24 debug_token = float24::FromRawFloat24(0x00abcdef);
-                input.attr[0].w = debug_token;
-
                 for (int i = 0; i < attribute_config.GetNumTotalAttributes(); ++i) {
-                    // Load the default attribute if we're configured to do so, this data will be overwritten by the loader data if it's set
-                    if (attribute_config.IsDefaultAttribute(i)) {
+                    if (vertex_attribute_elements[i] != 0) {
+                        // Default attribute values set if array elements have < 4 components. This
+                        // is *not* carried over from the default attribute settings even if they're
+                        // enabled for this attribute.
+                        static const float24 zero = float24::FromFloat32(0.0f);
+                        static const float24 one = float24::FromFloat32(1.0f);
+                        input.attr[i] = Math::Vec4<float24>(zero, zero, zero, one);
+
+                        // Load per-vertex data from the loader arrays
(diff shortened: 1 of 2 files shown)


Subv:
LGTM

yuriks:
Added test links to PR post.

src/video_core/command_processor.cpp
```diff
@@ -284,6 +280,12 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
                 // Send to vertex shader
                 VertexShader::OutputVertex output = VertexShader::RunShader(input, attribute_config.GetNumTotalAttributes());
 
+                // The hardware saturates vertex colors like this, *before* doing interpolation
```
neobrain: Should explicitly mention that the sign is dropped.

neobrain:
Please wait for my LGTM on this one.

neobrain:
Could you add some comments to the hardware tests explaining what you're doing, and why? It's very hard to follow what the point of everything is currently.

It's already pretty worrying how the current trend is towards having all tests as single applications with the same boilerplate copied over and over again in no standardized format. A year from now, all those tests will be impossible to verify due to ctrulib changes and the whatnot. At least make it so that people trying to come up with a better solution don't need to spend an hour for each test case to understand what's going on.

bunnei:
Agreed with @neobrain on the tests.... At least in that we should start thinking about collecting our GPU tests somewhere so they don't get lost in time.

Changes LGTM.

src/video_core/command_processor.cpp
```diff
@@ -284,6 +280,12 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                for (int i = 0; i < 4; ++i) {
+                    output.color[i] = float24::FromFloat32(
+                            std::fmin(std::fabs(output.color[i].ToFloat32()), 1.0f));
+                }
```
yuriks: This should probably be moved to inside `RunShader` instead.

yuriks:
Color clamping test updated with cleanups and documentation.

neobrain:
@yuriks as for the vertex-loading test, could you check what happens if you add another render pass (before the existing two passes) where you actually render the triangles using the default attributes for the color? A theory of mine is that the default attributes might only actually be "loaded" when being used, rather than when being set through the command list.

yuriks:
"existing two passes"? There is only one draw call per frame, I have no idea what you're talking about. Your comment makes no sense to me, the fallback attribute *is* enabled for rendering. It's still not used.

bunnei:
Straight from the mouth of The Brain<sup>TM</sup>:
<pre>[10:51] <@neobrain> bunnei_: ogogo merge</pre>

Subv:
fingers*

PR #964, opened 2015-07-21
dyncom: Pass SVC immediates directly.

lioncash:
Previously it would just re-read the already decoded instruction and extract the immediate value.


yuriks:
LGTM

PR #966, opened 2015-07-21
Services/Logging: Log more useful information when some operations fail.

Subv:
Namely OpenFileDirectly, OpenDirectory and OpenArchive


bunnei:
LGTM. OS X build got stuck.

PR #971, opened 2015-07-22
Implement stencil action 2 in SW renderer

MoochMcGee:
I referenced 3dbrew for this. This PR solely exists as a test to see if it fixes Kingdom Hearts: Dream Drop Distance. I haven't implemented this in the hardware renderer, because the PICA200's stencil operations are really different from the standard in OpenGL and OpenGL ES, and they seem to be completely incompatible.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -431,6 +431,7 @@ struct Regs {
 
     enum class StencilAction : u32 {
         Keep = 0,
+        AndReverse = 2,
         Xor  = 5,
     };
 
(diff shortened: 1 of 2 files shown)


Subv:
I'm pretty sure this (and all the related information in 3dbrew) is wrong.
We'll need to write hwtests to find the real behavior of those operations

lioncash:
Closing because no hw tests.

yuriks:
Btw, in the future, if you just want to share code with people, just push it to a branch on your fork, no need to create a PR for it.

hrydgard:
Those issues in Dream Drop Distance do not look much like stencil errors, rather some broken vertex format conversion / vertex loading.

PR #977, opened 2015-07-22
GL Renderer: Remove erroneous glEnable(GL_TEXTURE_2D) calls

yuriks:
In OpenGL 3, texturing is always enabled, and this call is invalid.
While it produced no effect in the rest of the execution, it wouldn't
have the intended effect of disabling texturing for that unit. Instead
bind a null texture to the unit.


Subv:
Could be unrelated but, shouldn't we be calling glActiveTexture in the gl_cache before setting the wrap_s/t and min/mag parameters?

yuriks:
Indeed, I think we do. But the glTexImage call is already there and *seems* to work, so maybe there's something that guarantees that it'll have the correct unit active at the time.

bunnei:
LGTM

PR #980, opened 2015-07-22
Qt/GPU Breakpoints: Added three more breakpoint types

Subv:
* IncomingDisplayTransfer: Triggered just before a display transfer is performed.
* GSPCommandProcessed: Triggered right after a GSP command is processed.
* BufferSwap: Triggered when the buffers flip

As a side note, I suggest dropping the Pica naming from these classes and having them be used for general GPU breakpoints, and then get rid of the GraphicsDebugger class.

### DIFF
--- a/src/core/hle/service/gsp_gpu.cpp
+++ b/src/core/hle/service/gsp_gpu.cpp
@@ -14,6 +14,7 @@
 #include "core/hw/lcd.h"
 
 #include "video_core/gpu_debugger.h"
+#include "video_core/debug_utils/debug_utils.h"
 #include "video_core/renderer_base.h"
 #include "video_core/video_core.h"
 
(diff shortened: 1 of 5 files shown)


src/core/hle/service/gsp_gpu.cpp
```diff
@@ -448,6 +449,9 @@ static void ExecuteCommand(const Command& command, u32 thread_id) {
     default:
         LOG_ERROR(Service_GSP, "unknown command 0x%08X", (int)command.id.Value());
     }
+
+    if (Pica::g_debug_context)
+        Pica::g_debug_context->OnEvent(Pica::DebugContext::Event::GSPCommandProcessed, (void*)&command);
```
Subv: IMO Pica::g_debug_context::OnEvent's second argument should be const.

src/video_core/debug_utils/debug_utils.h
```diff
@@ -25,11 +25,14 @@ class DebugContext {
...
-        CommandLoaded = FirstEvent,
-        CommandProcessed,
+        PicaCommandLoaded = FirstEvent,
+        PicaCommandProcessed,
         IncomingPrimitiveBatch,
         FinishedPrimitiveBatch,
         VertexLoaded,
+        IncomingDisplayTransfer,
+        GSPCommandProcessed,
+        BufferSwap,
```
neobrain: *BufferSwapped

Subv:
Updated

yuriks:
LGTM

bunnei:
LGTM

src/core/hle/service/gsp_gpu.cpp
neobrain: Yeah. Feel free to make a PR about that ;)

PR #981, opened 2015-07-23
Qt/GPU Breakpoints: Changed the widget to have a checkbox next to each bp type

Subv:
Changed it so that we don't have to select and click the Enable button when enabling/disabling a breakpoint, now it is done via a checkbox next to the breakpoint's name.

Removed the now redundant Enable button and second column of the list.
The QTreeView's headers are now hidden, seeing how we don't need them anymore.

Pic
Pic

### DIFF
--- a/src/citra_qt/debugger/graphics_breakpoints.cpp
+++ b/src/citra_qt/debugger/graphics_breakpoints.cpp
@@ -4,7 +4,7 @@
 
 #include <QMetaType>
 #include <QPushButton>
-#include <QTreeWidget>
+#include <QTreeView>
 #include <QVBoxLayout>
 #include <QLabel>
 
@@ -23,7 +23,7 @@ BreakPointModel::BreakPointModel(std::shared_ptr<Pica::DebugContext> debug_conte
 
 int BreakPointModel::columnCount(const QModelIndex& parent) const
 {
-    return 2;
+    return 1;
 }
 
 int BreakPointModel::rowCount(const QModelIndex& parent) const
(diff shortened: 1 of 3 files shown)


src/citra_qt/debugger/graphics_breakpoints_p.h
```diff
@@ -35,4 +35,4 @@ public slots:
     std::weak_ptr<Pica::DebugContext> context_weak;
     bool at_breakpoint;
     Pica::DebugContext::Event active_breakpoint;
-};
+};
```
archshift: Newline at EOF

src/citra_qt/debugger/graphics_breakpoints.cpp
```diff
@@ -84,36 +81,33 @@ QVariant BreakPointModel::data(const QModelIndex& index, int role) const
...
 bool BreakPointModel::setData(const QModelIndex& index, const QVariant& value, int role)
 {
     const auto event = static_cast<Pica::DebugContext::Event>(index.row());
 
     switch (role) {
-    case Role_IsEnabled:
+    case Qt::CheckStateRole:
     {
+        if (index.column() != 0)
+            return false;
```
neobrain: Any particular reason to return false instead of QVariant() ?

neobrain:
Looks okay given the current feature set, but do you have a plan on how to introduce a "Break every n-th occurrence" feature with this design?

src/citra_qt/debugger/graphics_breakpoints.cpp
Subv: It doesn't make much sense to me to return a QVariant, the function returns a bool.

Should i change it to QVariant?

Subv:
If that's on the roadmap perhaps we could keep the QTreeView but hide the headers?, they provide no useful information as of right now

src/citra_qt/debugger/graphics_breakpoints.cpp
neobrain: Ah, I was mixing this up with data(). Nevermind, sorry for the confusion.

Subv:
Updated

neobrain:
Looks okay to me, albeit this will add considerable amount of extra work to implement any GUI for a "break every n-th time"-like feature, so consider it a +0, only.

yuriks:
Could you make it so that it toggles the checkbox if you double-click the item? Other than that, looks good.

Subv:
Updated

yuriks:
LGTM 10/10

PR #982, opened 2015-07-23
Service/APT: Return proper parameters in GetLockHandle.

Subv:
Documented some APT functions
This allows applets to boot.


src/core/hle/service/apt/apt.cpp
```diff
@@ -101,18 +101,19 @@ void NotifyToWait(Service::Interface* self) {
 
 void GetLockHandle(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 flags = cmd_buff[1]; // TODO(bunnei): Figure out the purpose of the flag field
+    // Bits [0, 2] are the applet type (System, Library, etc)
```
yuriks: Is this "bit 0 and 2" or "bits 0 to 2"? If the latter, then use "[0:2]" notation which is pretty common for that sort of thing.

yuriks:
LGTM other than the comment

Subv:
Updated

Subv:
Fixed a regression, updated

yuriks:
LGTM (lol forgot to press submit)

bunnei:
LGTM

PR #983, opened 2015-07-24
GSP: Don't try to write memory fill registers if start address is 0

yuriks:
Verified to be what GSP does via REing. Fixes invalid virt->phys
translation error spam in some games.


Subv:
LGTM

PR #986, opened 2015-07-25
citra-qt: Improve pica command list widget

Lectem:
I seperated the data column into "Register" and "New Value" for the sake of clarity.
I also switched to using a QTableView because it is godamn faster, scrolling with a QTreeView with a lot of elements is extremely slow.

### DIFF
--- a/src/citra_qt/debugger/graphics_cmdlists.cpp
+++ b/src/citra_qt/debugger/graphics_cmdlists.cpp
@@ -174,7 +174,7 @@ int GPUCommandListModel::rowCount(const QModelIndex& parent) const {
 }
 
 int GPUCommandListModel::columnCount(const QModelIndex& parent) const {
-    return 2;
+    return 3;
 }
 
 QVariant GPUCommandListModel::data(const QModelIndex& index, int role) const {
@@ -187,14 +187,13 @@ QVariant GPUCommandListModel::data(const QModelIndex& index, int role) const {
 
     if (role == Qt::DisplayRole) {
         QString content;
-        if (index.column() == 0) {
-            QString content = QString::fromLatin1(Pica::Regs::GetCommandName(cmd.cmd_id).c_str());
-            content.append(" ");
-            return content;
-        } else if (index.column() == 1) {
-            QString content = QString("%1 ").arg(cmd.hex, 8, 16, QLatin1Char('0'));
-            content.append(QString("%1 ").arg(val, 8, 16, QLatin1Char('0')));
-            return content;
+        switch ( index.column() ) {
+        case 0:
+            return QString::fromLatin1(Pica::Regs::GetCommandName(cmd.cmd_id).c_str());
+        case 1:
+            return QString("%1").arg(cmd.cmd_id, 3, 16, QLatin1Char('0'));
+        case 2:
+            return QString("%1").arg(val, 8, 16, QLatin1Char('0'));
         }
     } else if (role == CommandIdRole) {
         return QVariant::fromValue<int>(cmd.cmd_id.Value());
@@ -207,10 +206,13 @@ QVariant GPUCommandListModel::headerData(int section, Qt::Orientation orientatio
     switch(role) {
     case Qt::DisplayRole:
     {
-        if (section == 0) {
+        switch (section) {
+        case 0:
             return tr("Command Name");
-        } else if (section == 1) {
-            return tr("Data");
+        case 1:
+            return tr("Register");
+        case 2:
+            return tr("New Value");
         }
 
         break;
--- a/src/citra_qt/debugger/graphics_cmdlists.cpp
+++ b/src/citra_qt/debugger/graphics_cmdlists.cpp
@@ -10,6 +10,7 @@
 #include <QPushButton>
 #include <QVBoxLayout>
 #include <QTreeView>
+#include <QHeaderView>
 #include <QSpinBox>
 #include <QComboBox>
 
@@ -301,6 +302,13 @@ GPUCommandListWidget::GPUCommandListWidget(QWidget* parent) : QDockWidget(tr("Pi
     list_widget->setModel(model);
     list_widget->setFont(QFont("monospace"));
     list_widget->setRootIsDecorated(false);
+    list_widget->setUniformRowHeights(true);
+
+#if QT_VERSION >= QT_VERSION_CHECK(5,0,0)
+    list_widget->header()->setSectionResizeMode(QHeaderView::ResizeToContents);
+#else
+    list_widget->header()->setResizeMode(QHeaderView::ResizeToContents);
+#endif
 
     connect(list_widget->selectionModel(), SIGNAL(currentChanged(const QModelIndex&,const QModelIndex&)),
             this, SLOT(SetCommandInfo(const QModelIndex&)));


neobrain:
QTableView is the wrong widget to use here, since we want a list-like look rather than a table-like one. Have you considered enabling http://doc.qt.io/qt-4.8/qtreeview.html#uniformRowHeights-prop instead, though?

Lectem:
It seems that the speedup was indeed caused by the rows being the same size, unifomRowHeights did the trick for QTreeView

neobrain:
Please improve your commit messages.

neobrain:
Is there any actual point in displaying the raw command id? It can be looked up easily in pica.h and hence only makes for one redundant column. Instead, I suggest displaying the raw command id in the first column iff `Pica::Regs::GetCommandName(cmd.cmd_id).isEmpty()`.

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -298,7 +301,13 @@ GPUCommandListWidget::GPUCommandListWidget(QWidget* parent) : QDockWidget(tr("Pi
     list_widget = new QTreeView;
     list_widget->setModel(model);
     list_widget->setFont(QFont("monospace"));
-    list_widget->setRootIsDecorated(false);
+    list_widget->setUniformRowHeights(true);
+
+#if QT_VERSION >= QT_VERSION_CHECK(5,0,0)
+    list_widget->header()->setSectionResizeMode(QHeaderView::ResizeToContents);
```
neobrain: What is the difference to setResizeMode?

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -298,7 +301,13 @@ GPUCommandListWidget::GPUCommandListWidget(QWidget* parent) : QDockWidget(tr("Pi
     list_widget = new QTreeView;
     list_widget->setModel(model);
     list_widget->setFont(QFont("monospace"));
-    list_widget->setRootIsDecorated(false);
```
neobrain: Why was this removed?

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -207,10 +207,13 @@ QVariant GPUCommandListModel::headerData(int section, Qt::Orientation orientatio
...
     {
-        if (section == 0) {
+        switch (section) {
+        case 0 :
             return tr("Command Name");
-        } else if (section == 1) {
-            return tr("Data");
+        case 1 :
+            return tr("Register");
+        case 2 :
```
neobrain: redundant space before `:`

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -187,14 +188,13 @@ QVariant GPUCommandListModel::data(const QModelIndex& index, int role) const {
...
-            return content;
-        } else if (index.column() == 1) {
-            QString content = QString("%1 ").arg(cmd.hex, 8, 16, QLatin1Char('0'));
-            content.append(QString("%1 ").arg(val, 8, 16, QLatin1Char('0')));
-            return content;
+        switch ( index.column() ) {
+        case 0 :
+            return QString::fromLatin1(Pica::Regs::GetCommandName(cmd.cmd_id).c_str()).append(" ");
+        case 1 :
+            return QString("%1 ").arg(cmd.cmd_id, 4, 16, QLatin1Char('0'));
```
neobrain: Redundant space

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -187,14 +188,13 @@ QVariant GPUCommandListModel::data(const QModelIndex& index, int role) const {
...
-            QString content = QString::fromLatin1(Pica::Regs::GetCommandName(cmd.cmd_id).c_str());
-            content.append(" ");
-            return content;
-        } else if (index.column() == 1) {
-            QString content = QString("%1 ").arg(cmd.hex, 8, 16, QLatin1Char('0'));
-            content.append(QString("%1 ").arg(val, 8, 16, QLatin1Char('0')));
-            return content;
+        switch ( index.column() ) {
+        case 0 :
+            return QString::fromLatin1(Pica::Regs::GetCommandName(cmd.cmd_id).c_str()).append(" ");
```
neobrain: why append a space?

Lectem:
Well it's faster to look for those with the ID on 3dbrew, and not all users have citra's source under their hands. Plus it assumes that all the names are correct, which might not always be the case.

src/citra_qt/debugger/graphics_cmdlists.cpp
Lectem: Well, it was there so I didnt remove it, I guessed their was a good reason for why it was there

src/citra_qt/debugger/graphics_cmdlists.cpp
Lectem: My bad, forgot to add it back when using QTreeView again

src/citra_qt/debugger/graphics_cmdlists.cpp
Lectem: setResizeMode was renamed to setSectionResizeMode in Qt 5.0

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -187,14 +188,13 @@ QVariant GPUCommandListModel::data(const QModelIndex& index, int role) const {
...
-            content.append(QString("%1 ").arg(val, 8, 16, QLatin1Char('0')));
-            return content;
+        switch ( index.column() ) {
+        case 0:
+            return QString::fromLatin1(Pica::Regs::GetCommandName(cmd.cmd_id).c_str());
+        case 1:
+            return QString("%1").arg(cmd.cmd_id, 4, 16, QLatin1Char('0'));
+        case 2:
+            return QString("%1").arg(val, 8, 16, QLatin1Char('0'));
         }
```
neobrain: Will using a `switch` here introduce warnings about missing default case?

neobrain:
LGTM other than the potential introduction of two warnings related to unhandled `switch` cases.

Lectem:
Visual studio 2013 didn't give me any warning (I think most compilers only do so on enums, you are not warned when using an integer in a switch)

Subv:
```
Is there any actual point in displaying the raw command id? It can be looked up easily in pica.h and hence only makes for one redundant column. Instead, I suggest displaying the raw command id in the first column iff Pica::Regs::GetCommandName(cmd.cmd_id).empty().
```

I disagree, searching for the command id is not trivial in Pica.h, you have to look for a point of reference at the end of the file then backtrack until you find what you're looking for. Displaying the command id along with the name is better in my opinion.

neobrain:
> I disagree, searching for the command id is not trivial in Pica.h, you have to look for a point of reference at the end of the file then backtrack until you find what you're looking for. Displaying the command id along with the name is better in my opinion.

Are you thinking of something else maybe? You literally just have to look up the name shown in the widget in the list at https://github.com/citra-emu/citra/blob/master/src/video_core/pica.h#L1005 (and then do some basic arithmetic if it e.g. shows "framebuffer+4").

yuriks:
I would also prefer to keep the raw addresses always. The last thing I need when trying to RE something is to have to keep looking up crap.

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -187,14 +188,13 @@ QVariant GPUCommandListModel::data(const QModelIndex& index, int role) const {
...
-            return content;
-        } else if (index.column() == 1) {
-            QString content = QString("%1 ").arg(cmd.hex, 8, 16, QLatin1Char('0'));
-            content.append(QString("%1 ").arg(val, 8, 16, QLatin1Char('0')));
-            return content;
+        switch ( index.column() ) {
+        case 0:
+            return QString::fromLatin1(Pica::Regs::GetCommandName(cmd.cmd_id).c_str());
+        case 1:
+            return QString("%1").arg(cmd.cmd_id, 4, 16, QLatin1Char('0'));
```
yuriks: Set the width to 3, since regs only go to ~300.

src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -299,6 +302,13 @@ GPUCommandListWidget::GPUCommandListWidget(QWidget* parent) : QDockWidget(tr("Pi
     list_widget->setModel(model);
     list_widget->setFont(QFont("monospace"));
     list_widget->setRootIsDecorated(false);
+    list_widget->setUniformRowHeights(true);
+
+#if QT_VERSION >= QT_VERSION_CHECK(5,0,0)
+    list_widget->header()->setSectionResizeMode(QHeaderView::ResizeToContents);
+#else
+    list_widget->header()->setResizeMode(QHeaderView::ResizeToContents);
+#endif
```
yuriks: "setResizeMode" is deprecated, but afaik still present in Qt 5. I'd just keep using it rather than having an ifdef.

Lectem: It isn't available (at least for 5.5)

yuriks: Hm, it's still listed in the documentation: http://doc.qt.io/qt-5/qheaderview-obsolete.html#setResizeMode

But well, it's not a big issue either way.

yuriks:
LGTM

PR #987, opened 2015-07-25
Videocore: Don't reinitialize register name map on every query

yuriks:
This greatly speeds up the command list debug widget.

### DIFF
--- a/src/video_core/pica.cpp
+++ b/src/video_core/pica.cpp
@@ -2,13 +2,83 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include <string.h>
+#include <cstring>
+#include <unordered_map>
 
 #include "pica.h"
 
 namespace Pica {
 
 State g_state;
+
+std::string Regs::GetCommandName(int index) {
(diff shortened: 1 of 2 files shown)


src/video_core/pica.cpp
```diff
@@ -10,6 +11,70 @@ namespace Pica {
 
 State g_state;
 
+std::string Regs::GetCommandName(int index) {
+    static std::unordered_map<u32, std::string> map;
```
neobrain: Why don't you just change this to a plain array?

yuriks: The map is (at least right now) very sparse, so the array would waste a lot of memory with empty strings.

neobrain: Mind you, it's already doing this right now (to a lesser extent) since the command list viewer will call GetCommandName for lots of "unknown" commands, hence inserting many dummy elements into the map.

yuriks: Ugh, I hadn't spotted that. Let me fix it.

yuriks:
Updated

PR #988, opened 2015-07-25
loader: Remove unnecessary else usages

lioncash:


yuriks:
LGTM

mailwl:
Ah! I have a question, is it hard to implement loader for .cia-format?

PR #991, opened 2015-07-26
OpenGL: Make OpenGL object resource wrappers fully inline

yuriks:
The functions are so simple that having them separate only bloats the
code and hinders optimization. Also added move constructor which will
come in handy.


Subv:
LGTM

yuriks:
Oops, found a bug. Don't merge.

yuriks:
Ok, done. (Was missing returns on the operator='s)

bunnei:
LGTM

PR #992, opened 2015-07-26
VideoCore: #ifdef out some debugging routines

yuriks:
Some disabled debugging functionality was being called from rendering
routines in VideoCore. Although disabled, many of them still allocated
memory or did some extra work that was enough to show up in a profiler.
Gives a slight (~2ms) speedup.


archshift:
How about having these exposed via cmake?

neobrain:
Tbh both of these aren't really relevant long-term. The Tev setup will eventually be exposed in the GUI and the limited functionality available through the geo dumper will be rendered useless by the vertex shader debugger. So yeah, it's fine as is and doesn't really need a CMake option IMO.

bunnei:
LGTM

PR #994, opened 2015-07-26
Optimisations and cleanups in the shader interpreter

yuriks:
- Replaces `std::stack` with `boost::static_vector` (much faster)
 - Replaces two references to u32 types with values (simpler, faster)
 - Makes program counter an index instead of a pointer (simpler, faster)

### DIFF
--- a/src/video_core/vertex_shader.cpp
+++ b/src/video_core/vertex_shader.cpp
@@ -2,8 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include <stack>
-
+#include <boost/container/static_vector.hpp>
 #include <boost/range/algorithm.hpp>
 
 #include <common/file_util.h>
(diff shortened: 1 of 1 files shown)


archshift:
LGTM

src/video_core/vertex_shader.cpp
```diff
@@ -53,7 +52,7 @@ struct VertexShaderState {
     };
 
     // TODO: Is there a maximal size for this?
-    std::stack<CallStackElement> call_stack;
+    boost::container::static_vector<CallStackElement, 16> call_stack;
```
neobrain: This is limited to 16 elements, right? If so, I'm not seeing any bounds checks...

src/video_core/vertex_shader.cpp
```diff
@@ -88,17 +87,15 @@ static void ProcessShaderCode(VertexShaderState& state) {
         }
 
         bool exit_loop = false;
-        const Instruction& instr = *(const Instruction*)state.program_counter;
-        const SwizzlePattern& swizzle = *(SwizzlePattern*)&swizzle_data[instr.common.operand_desc_id];
+        const Instruction instr = { program_code[state.program_counter] };
```
neobrain: How is a copy faster here?

yuriks: It's a u32. The pointer/reference is actually larger than that. :P

neobrain:
LGTM

PR #997, opened 2015-07-26
citra-qt: Improve pica command list widget (add mask, fix some issues)

Lectem:
The monospace font was not being used on Windows, this has been fixed.
I added the mask and moved its expansion into WritePicaReg.
A widget was created to replace the previous one every single time a row was selected, which caused some slowdowns. No widget is now created if not needed.

### DIFF
--- a/src/citra_qt/debugger/graphics_cmdlists.cpp
+++ b/src/citra_qt/debugger/graphics_cmdlists.cpp
@@ -300,7 +300,9 @@ GPUCommandListWidget::GPUCommandListWidget(QWidget* parent) : QDockWidget(tr("Pi
 
     list_widget = new QTreeView;
     list_widget->setModel(model);
-    list_widget->setFont(QFont("monospace"));
+    QFont font("monospace");
+    font.setStyleHint(QFont::Monospace); // Automatic fallback to a monospace font on on platforms without a font called "monospace"
+    list_widget->setFont(font);
     list_widget->setRootIsDecorated(false);
     list_widget->setUniformRowHeights(true);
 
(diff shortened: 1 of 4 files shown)


src/citra_qt/debugger/graphics_cmdlists.cpp
```diff
@@ -300,7 +303,9 @@ GPUCommandListWidget::GPUCommandListWidget(QWidget* parent) : QDockWidget(tr("Pi
 
     list_widget = new QTreeView;
     list_widget->setModel(model);
-    list_widget->setFont(QFont("monospace"));
+    QFont font("monospace");
+    font.setStyleHint(QFont::Monospace); // Automatic fallback to a monospace font on windows
```
neobrain: *on platforms without a font called "monospace"

(also as a general note, it's not "windows" but "Windows" ..)

neobrain:
Please rework your patches to not include changes outside of their denoted summary.

Lectem:
I addressed the previous comments and fixed the build error https://travis-ci.org/citra-emu/citra/jobs/72684273#L143 by casting the values to u16.

src/video_core/command_processor.cpp
```diff
@@ -47,13 +55,16 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
 
     // TODO: Figure out how register masking acts on e.g. vs.uniform_setup.set_value
     u32 old_value = regs[id];
-    regs[id] = (old_value & ~mask) | (value & mask);
+
+    const u32 write_mask = expand_bits_to_bytes[mask];
+
+    regs[id] = (old_value & ~write_mask) | (value & write_mask);
+
+    DebugUtils::OnPicaRegWrite({ (u16)id, (u16)mask, regs[id] });
```
bunnei: static_cast instead of C-style cast

yuriks: We changed the style guide a while back to encourage C-casts for numeric-numeric conversions, since the static_cast was too verbose. (static_cast remains required for all other kinds of casts)

neobrain:
LGTM.

PR #998, opened 2015-07-26
Update Start menu text to match with the real state of the emulator.

LittleWhite-tb:
Small change to update the start menu. With this commit, the start menu is renamed to "Continue" while the emulator is running. In stop function, the start menu text is set back to "Start".

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -263,6 +263,7 @@ void GMainWindow::ShutdownGame() {
 
     // Update the GUI
     ui.action_Start->setEnabled(false);
+    ui.action_Start->setText(tr("Start"));
     ui.action_Pause->setEnabled(false);
     ui.action_Stop->setEnabled(false);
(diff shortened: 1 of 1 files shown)


src/citra_qt/main.cpp
```diff
@@ -305,6 +307,7 @@ void GMainWindow::OnPauseGame()
 }
 
 void GMainWindow::OnStopGame() {
+    ui.action_Start->setText(tr("Start"));
```
neobrain: This should be done in ShutdownGame instead

bunnei:
Please squash to a single commit. Otherwise, LGTM.

LittleWhite-tb:
Done. Hope everything is correct now.

yuriks:
Still isn't squashed.

archshift:
LGTM

yuriks:
LGTM

KitatusStudios:
LGTM

PR #999, opened 2015-07-26
Save the path leading where the last file have been loaded

LittleWhite-tb:
This commit adds the save of the path where the last file have been loaded.

However, I don't know if it is a good to keep the path in main.h.

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -122,9 +122,8 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     y = (screenRect.y() + screenRect.height()) / 2 - h * 55 / 100;
     setGeometry(x, y, w, h);
 
-
     // Restore UI state
-    QSettings settings(QSettings::IniFormat, QSettings::UserScope, "Citra team", "Citra");
+    QSettings settings;
     restoreGeometry(settings.value("geometry").toByteArray());
     restoreState(settings.value("state").toByteArray());
     render_window->restoreGeometry(settings.value("geometryRenderWindow").toByteArray());
@@ -271,8 +270,13 @@ void GMainWindow::ShutdownGame() {
 
 void GMainWindow::OnMenuLoadFile()
 {
-    QString filename = QFileDialog::getOpenFileName(this, tr("Load File"), QString(), tr("3DS executable (*.3ds *.3dsx *.elf *.axf *.cci *.cxi)"));
+    QSettings settings;
+    QString rom_path = settings.value("romsPath", QString()).toString();
+
+    QString filename = QFileDialog::getOpenFileName(this, tr("Load File"), rom_path, tr("3DS executable (*.3ds *.3dsx *.elf *.axf *.cci *.cxi)"));
     if (filename.size()) {
+        settings.setValue("romsPath", QFileInfo(filename).path());
+
         // Shutdown previous session if the emu thread is still active...
         if (emu_thread != nullptr)
             ShutdownGame();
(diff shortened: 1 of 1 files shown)


src/citra_qt/main.h
```diff
@@ -85,6 +85,8 @@ private slots:
     CallstackWidget* callstackWidget;
     GPUCommandStreamWidget* graphicsWidget;
     GPUCommandListWidget* graphicsCommandsWidget;
+
+    QString romPath;
```
yuriks: Is there any compelling reason to cache this as a member of the class? I think it can be comfortably read/set around the dialog showing code. For this, it's probably a good idea to make the `QSettings` object a member to avoid recreating it all the time though.

src/citra_qt/main.cpp
```diff
@@ -281,9 +285,12 @@ void GMainWindow::OnMenuLoadFile()
 }
 
 void GMainWindow::OnMenuLoadSymbolMap() {
-    QString filename = QFileDialog::getOpenFileName(this, tr("Load Symbol Map"), QString(), tr("Symbol map (*)"));
-    if (filename.size())
+    QString filename = QFileDialog::getOpenFileName(this, tr("Load Symbol Map"), romPath, tr("Symbol map (*)"));
+    if (filename.size()) {
+        romPath = QFileInfo(filename).path();
```
yuriks: While I think it's ok to default this to the rom path (it's reasonably likely that you'll be keeping binaries and map files in the same directory) I think saving it is a more shakey proposition, since if a user *does* navigate away, it would change the rom path and then they'll probably have to navigate *back* when they load a rom next time.

src/citra_qt/main.h
LittleWhite-tb: The reason was that I need the setting in multiple functions. Of course, we can replace that by putting the QSetting as member of the class.
Moreover, the documentation of the class, does not seem to be really bothered about reconstructing the QSetting multiple times.
"
If you use QSettings from many places in your application, you might want to specify the organization name and the application name using QCoreApplication::setOrganizationName() and QCoreApplication::setApplicationName(), and then use the default QSettings constructor
"
But for the moment, I will put it as class member. I think it's the best since we don't need it elsewhere (apart in hotkeys, but we are passing the QSettings as param).

yuriks: From the wording in the Qt docs:

>QSettings objects can be created either on the stack or on the heap (i.e. using new). Constructing and destroying a QSettings object is very fast.

It seems like they expect/recommend you to re-create the object as needed, and not store it persistently. With that in mind, I think we should go that route and just create a QSettings every time the file dialog is invoked. (Using `setOrganizationName` and `setApplicationName` to avoid duplicating those, of course.)

LittleWhite-tb: Ok to me. Will do it soon.

LittleWhite-tb:
I have applied everything mentioned before. I hope this is looking better.
I have squashed my commit. Maybe I should have wait for your request before squashing it ?

bunnei:
LGTM. I'll let @yuriks merge this one to ensure his comments are adequately addressed.

src/citra_qt/main.cpp
```diff
@@ -270,8 +275,13 @@ void GMainWindow::ShutdownGame() {
 
 void GMainWindow::OnMenuLoadFile()
 {
-    QString filename = QFileDialog::getOpenFileName(this, tr("Load File"), QString(), tr("3DS executable (*.3ds *.3dsx *.elf *.axf *.cci *.cxi)"));
+    QSettings settings;
+    QString romsPath = settings.value("romsPath", QString()).toString();
```
yuriks: Use "rom_path" naming style for variables.

src/citra_qt/main.cpp
```diff
@@ -281,9 +291,15 @@ void GMainWindow::OnMenuLoadFile()
 }
 
 void GMainWindow::OnMenuLoadSymbolMap() {
-    QString filename = QFileDialog::getOpenFileName(this, tr("Load Symbol Map"), QString(), tr("Symbol map (*)"));
-    if (filename.size())
+    QSettings settings;
+    QString symbolsPath = settings.value("symbolsPath", QString()).toString();
```
yuriks: Same

src/citra_qt/main.cpp
```diff
@@ -123,8 +123,13 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     setGeometry(x, y, w, h);
 
 
+    // Init settings params
+    QSettings::setDefaultFormat(QSettings::IniFormat);
+    QCoreApplication::setOrganizationName("Citra team");
+    QCoreApplication::setApplicationName("Citra");
```
yuriks: I think you should move these to the `main` function instead of the window constructor, but it doesn't matter much either way.

yuriks:
Squash and LGTM

LittleWhite-tb:
Squashed

PR #1002, opened 2015-07-27
Vertex Shader JIT for X86-64

bunnei:
This adds Just-In-Time (JIT) compilation for the 3DS’ vertex shaders to 64-bit x86 machine code. This requires SSE2, but there are some separate instruction implementations for SSE4 (if it is supported by the host hardware). I have no plans to add support for 32-bit x86. The JIT shader core should be as accurate as the interpreter, and several orders of magnitude faster. Due to the JIT being both faster than and as accurate as the interpreter, I have opted to make it the default configuration. I have not yet tested this on Unix, non-x86, nor non-64-bit platforms, so I’d appreciate anyone using these to test. Furthermore, because this will be the default shader core, I’d appreciate if as many people as possible can test their game libraries, and I will fix any deviations from the shader interpreter.

The commits in this PR should be nicely separated, so I recommend reviewing them individually. 

PR history:
* https://github.com/bunnei/citra/compare/82036694d8dc42efafc3d972a266270da0eb9046...bunnei:265a11a99fe4ed406fff89a91a00c953a59427b5


src/common/cpu_detect_x86.cpp
```diff
@@ -0,0 +1,230 @@
...
+CPUInfo cpu_info;
+
+CPUInfo::CPUInfo() {
+    Detect();
+    SetHash64Function();
+}
+
+// Detects the various CPU features
+void CPUInfo::Detect() {
+    memset(this, 0, sizeof(*this));
```
lioncash: simply epic

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    // Negate
+    const unsigned negate[] = { swiz.negate_src1, swiz.negate_src2, swiz.negate_src3 };
+    if (negate[src_num - 1] != 0) {
+        PXOR(dest, R(NEGBIT));
+    }
+}
+
+void Jit::Comp_DestEnable(Instruction instr,X64Reg src) {
```
lioncash: missing space

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -0,0 +1,620 @@
...
+    SwizzlePattern& swiz = *(SwizzlePattern*)&g_state.vs.swizzle_data[operand_desc_id];
+
+    // Is everything enabled?
+    if (swiz.dest_mask == 0xf) {
+        // Store dest back to memory
+        MOVAPS(MDisp(STATE, UnitState::OutputOffset(dest)), src);
+
+    // Nope, mask destination...
+    } else {
+
```
lioncash: unnecessary newline heere and in the if case

hrydgard:
Love seeing my JIT infrastructure spread to new projects!

PSHUFB is actually an SSSE3 instruction (note: 3 'S'), and it's not available in the very oldest x64 cores. However, you would probably be hard pressed to find one that has issues with this today, but having a fallback to a combination of SHUFPS instructions would probably be a good thing.

Oh, and "order of magnitude" usually means ~10x, so this is just a single order of magnitude faster, if your 6x number is right :)

src/video_core/shader/shader.h
```diff
@@ -0,0 +1,153 @@
...
+
+    struct {
+        u32 max_offset; // maximum program counter ever reached
+        u32 max_opdesc_id; // maximum swizzle pattern index ever used
+    } debug;
+
+    static size_t InputOffset(const SourceRegister& reg) {
+        switch (reg.GetRegisterType()) {
+        case RegisterType::Input:
+            return offsetof(UnitState, input_registers[reg.GetIndex()]);
```
darkf: g++ complains about this:

	citra-jit/src/./video_core/shader/shader.h: In static member function 'static size_t Pica::Shader::UnitState::InputOffset(const Pica::Shader::SourceRegister&)':
	citra-jit/src/./video_core/shader/shader.h:109:40: warning: invalid access to non-static data member 'Pica::Shader::UnitState::input_registers'  of NULL object [-Winvalid-offsetof]
	             return offsetof(UnitState, input_registers[reg.GetIndex()]);
	                                        ^
	citra-jit/src/./video_core/shader/shader.h:109:40: warning: (perhaps the 'offsetof' macro was used incorrectly) [-Winvalid-offsetof]
	citra-jit/src/./video_core/shader/shader.h:109:69: error: call to non-constexpr function 'int Pica::Shader::SourceRegister::GetIndex() const'
	             return offsetof(UnitState, input_registers[reg.GetIndex()]);

src/common/hash.cpp
```diff
@@ -0,0 +1,417 @@
...
+        samples = std::max(Step, 1u);
+    Step = Step / samples;
+    if (Step < 1)
+        Step = 1;
+#endif
+
+#ifdef _M_X86
+    while (data < end - Step * 3)
+    {
+        h[0] = _mm_crc32_u64(h[0], data[Step * 0]);
```
darkf: `_mm_crc32_u64` is SSE4 but it's not under a guard.

src/video_core/shader/shader.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+namespace Pica {
+
+namespace Shader {
+
+static std::unordered_map<u64, std::unique_ptr<JitShader>> shader_map;
+static Jit jit;
+static JitShader* jit_shader;
+
+void Setup(UnitState& state) {
```
Subv: What's this parameter for?

src/video_core/shader/shader.cpp
```diff
@@ -0,0 +1,126 @@
...
+    const u64 cache_key = Common::GetHash64((const u8*)&g_state.vs.program_code, 8192, 1024) ^
```
Subv: Is this guaranteed to have no collisions?

src/video_core/shader/shader.h
```diff
@@ -0,0 +1,153 @@
...
+    // TODO: Is there a maximal size for this?
+    boost::container::static_vector<CallStackElement, 16> call_stack;
+
+    struct {
+        u32 max_offset; // maximum program counter ever reached
+        u32 max_opdesc_id; // maximum swizzle pattern index ever used
+    } debug;
+
+    static size_t InputOffset(const SourceRegister& reg) {
+        switch (reg.GetRegisterType()) {
```
Subv: This code is not optimal, there are extra switches inside GetRegisterType that can be avoided

src/video_core/shader/shader_bytecode.h
```diff
@@ -0,0 +1,783 @@
...
+            return Id::MAD;
+        else if (static_cast<Id>(op & ~0x7) == Id::MADI)
+            return Id::MADI;
+        else if (static_cast<Id>(op & ~0x1) == Id::CMP)
+            return Id::CMP;
+        else
+            return static_cast<Id>(value);
+    }
+
+    const Info& GetInfo() const {
```
Subv: This file seems to be from an old version of nihstro, this function is not optimal, please use the file from the latest nihstro master

src/video_core/shader/shader_bytecode.h
```diff
@@ -0,0 +1,783 @@
...
+        };
+        std::string ret;
+        for (int i = 0; i < 4; ++i) {
+            ret += map.at(src2 ? GetSelectorSrc2(i) : GetSelectorSrc1(i));
+        }
+        return ret;
+    }
+
+    bool DestComponentEnabled(unsigned int i) const {
+        if (i >= 4)
```
Subv: Unneeded exception

src/video_core/shader/shader_interpreter.cpp
```diff
@@ -90,7 +46,7 @@ static void ProcessShaderCode(VertexShaderState& state) {
         const Instruction instr = { program_code[state.program_counter] };
         const SwizzlePattern swizzle = { swizzle_data[instr.common.operand_desc_id] };
 
-        static auto call = [](VertexShaderState& state, u32 offset, u32 num_instructions,
+        static auto call = [&program_code](UnitState& state, u32 offset, u32 num_instructions,
```
Subv: You don't need this capture, rebase error?

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -0,0 +1,620 @@
...
+struct SwizzleDstMask : std::array<__m128i, 16> {
+    SwizzleDstMask() {
+        for (int i = 0; i < 16; ++i) {
+            (*this)[i].m128i_u32[0] = ((i >> 3) & 1) * 0xffffffff;
+            (*this)[i].m128i_u32[1] = ((i >> 2) & 1) * 0xffffffff;
+            (*this)[i].m128i_u32[2] = ((i >> 1) & 1) * 0xffffffff;
+            (*this)[i].m128i_u32[3] = (i & 1) * 0xffffffff;
+        }
+    }
+} static swizzle_dst_mask;
```
Subv: Should this be const?

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -0,0 +1,620 @@
...
+#define SELECT(n) ((n) * 0x04040404u + 0x03020100u)
+        for (int i = 0; i < 256; ++i) {
+            (*this)[i].m128i_u32[0] = SELECT((i >> 6) & 0x3);
+            (*this)[i].m128i_u32[1] = SELECT((i >> 4) & 0x3);
+            (*this)[i].m128i_u32[2] = SELECT((i >> 2) & 0x3);
+            (*this)[i].m128i_u32[3] = SELECT(i & 0x3);
+        }
+#undef SELECT
+    }
+} static swizzle_src_selector;
```
Subv: Same

src/video_core/shader/shader.cpp
hrydgard: No hash is guaranteed but you'll never see any in practice with a good 64-bit hash, given how few shaders are used per game the probabilities are absolutely tiny.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -0,0 +1,620 @@
...
+
+void Jit::Comp_NextInstr(unsigned* offset) {
+    offset_ptr = offset;
+
+    Instruction instr = *(Instruction*)&g_state.vs.program_code[(*offset_ptr)++];
+    auto instr_func = instr_table[instr.opcode.Value()];
+
+    // JIT it!
+    if (instr_func) {
+        ((*this).*instr_func)(instr);
```
Subv: this->*instr_func(instr)

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -0,0 +1,620 @@
...
+
+    // Restore current offset pointer
+    offset_ptr = prev_offset_ptr;
+    *offset_ptr = offset;
+}
+
+void Jit::Comp_NextInstr(unsigned* offset) {
+    offset_ptr = offset;
+
+    Instruction instr = *(Instruction*)&g_state.vs.program_code[(*offset_ptr)++];
```
Subv: static_ or reinterpret_cast please

src/video_core/shader/shader.cpp
Subv: Is there any possibility (however small it may be) that we end up executing the wrong cached shader? This code doesn't seem to do any further checking on the returned map element, and since we're doing the hashing ourselves we won't benefit from the linear search unordered_map performs in its buckets if it finds a collision.

hrydgard: Unless you want to use a copy of the whole shader code as a key (or on the side to compare with), you can't _completely_ rule out collisions. But like I said, in practice, it's just not going to happen.

neobrain:
For reference, I edited the initial comment of this PR to include a history so that comments on old commits can still be retrieved. @bunnei Please update that history whenever you add new commits and/or force-push onto your branch.

neobrain:
GitHub flagged me as a bot while I was reviewing this PR, lol. Anyway, make sure you look at the link in the history to catch all my comments instead of relying on email notifications.

src/video_core/shader/shader.cpp
neobrain: Doesn't `std::unordered_map` resolve hash collisions internally?

src/common/abi.cpp
```diff
@@ -0,0 +1,680 @@
...
+#include "abi.h"
+
+using namespace Gen;
+
+// Shared code between Win64 and Unix64
+
+// Sets up a __cdecl function.
+void XEmitter::ABI_EmitPrologue(int maxCallParams)
+{
+#ifdef _M_IX86
```
archshift: This macro is Windows-only.

src/common/abi.cpp
```diff
@@ -0,0 +1,680 @@
...
+    // Don't really need to do anything
+#elif defined(_M_X64)
```
archshift: Ditto.

src/common/abi.cpp
```diff
@@ -0,0 +1,680 @@
...
+    SUB(64, R(RSP), Imm8(stacksize));
+#endif
+#else
+#error Arch not supported
+#endif
+}
+
+void XEmitter::ABI_EmitEpilogue(int maxCallParams)
+{
+#ifdef _M_IX86
```
archshift: Also Windows-only.

src/common/abi.cpp
```diff
@@ -0,0 +1,680 @@
...
+    RET();
+#elif defined(_M_X64)
```
archshift: Also Windows-only. I won't point this out anymore, please use something cross-platform.

src/video_core/shader/shader.cpp
yuriks: This should also be hashing the swizzle table, since they're baked into the generated code. (I doubt they'd ever change while staying with the same program, but you never know.)

yuriks:
Done with pass. Not perfect, but it's not too shabby. :)

I think my biggest complain is how you're not suppording backwards jumps, and more importantly, that you're inlining them. This can easily cause infinite loops in the JIT if the shader uses them in interesting ways.

Also I dislike how there's an entire "dummy emitter" class and "dummy jit". I feel like those should be just #ifdef'ed out completely when they're not available in the architecture.

src/video_core/shader/shader.cpp
hrydgard: @neobrain well, sure, but only if you use the whole shader as the key to an unordered_map (which will then proceed to hash them internally, and falling back to full compare).

neobrain: D'oh, right.

Subv: Can't we use unordered_map directly and just provide our own implementation of std::hash for the shader + swizzle?

src/video_core/shader/shader_interpreter.cpp
yuriks: Still here. Fix pls

src/video_core/shader/shader.cpp
```diff
@@ -0,0 +1,126 @@
...
+                *out = state.output_registers[i][comp];
+            }
+            else {
+                // Zero output so that attributes which aren't output won't have denormals in them,
+                // which would slow us down later.
+                memset(out, 0, sizeof(*out));
+            }
+        }
+    }
+
```
yuriks: Looks like you nuked the piece of code that was introduced here by #963

src/common/abi.cpp
```diff
@@ -0,0 +1,680 @@
...
+    // Don't really need to do anything
+#elif defined(_M_X86_64)
+#if _WIN32
+    int stacksize = ((maxCallParams + 1) & ~1) * 8 + 8;
+    // Set up a stack frame so that we can call functions
+    // TODO: use maxCallParams
+    SUB(64, R(RSP), Imm8(stacksize));
+#endif
+#else
+#error Arch not supported
```
linkmauve: Move that #error to the top of the file, and drop the x86 check everywhere else in that file.

CMakeLists.txt
```diff
@@ -51,6 +51,35 @@ else()
     set(CMAKE_EXE_LINKER_FLAGS_RELEASE "/DEBUG /INCREMENTAL:NO /OPT:REF,ICF" CACHE STRING "" FORCE)
 endif()
 
+# Detect 64bit or 32bit
+# CMake doesn't provide a simple way to determine 32bit or 64bit
+# If we ever support a architecture that is 64bit with 32bit pointers then this'll break
+# Of course the chances of that are slim(x32?) so who cares
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+    set(_ARCH_64 1)
+    add_definitions(-D_ARCH_64=1)
```
linkmauve: As per section 17.6.4.3.2 of the C++11 specification (and IIRC older specs as well), “Each name that contains a double underscore __ or begins with an underscore followed by an uppercase letter is reserved to the implementation for any use.” so you shouldn’t define any such variable.

archshift:
Segfaults on OS X because the memory region you use for JIT'd code is marked `rw-`...non-executable.

archshift:
Segfaults on OS X because jitted code `movaps (%r15), %xmm1`'s destination address is not aligned on a 16-byte boundary, which is required.

src/common/abi.cpp
```diff
@@ -0,0 +1,680 @@
...
+void XEmitter::ABI_AlignStack(unsigned int frameSize) {
+// Mac OS X requires the stack to be 16-byte aligned before every call.
+// Linux requires the stack to be 16-byte aligned before calls that put SSE
+// vectors on the stack, but since we do not keep track of which calls do that,
+// it is effectively every call as well.
+// Windows binaries compiled with MSVC do not have such a restriction*, but I
+// expect that GCC on Windows acts the same as GCC on Linux in this respect.
+// It would be nice if someone could verify this.
+// *However, the MSVC optimizing compiler assumes a 4-byte-aligned stack at times.
+    unsigned int fillSize =
```
linkmauve: The type of this seems wrong, since Imm8 will & 0xFF it, maybe make it an u8, or maybe add an assert for bigger values.

src/common/abi.cpp
```diff
@@ -0,0 +1,680 @@
...
+        ABI_GetAlignedFrameSize(frameSize) - (frameSize + 4);
+    if (fillSize != 0) {
```
linkmauve: We don’t use parens with single line content.

src/common/abi.cpp
yuriks: The fill size will never be larger than 32 (for AVX), so this isn't a problem.

src/common/abi.cpp
```diff
@@ -0,0 +1,680 @@
...
+    }
+}
+
+#else //64bit
+
+// Common functions
+void XEmitter::ABI_CallFunction(const void *func) {
+    u64 distance = u64(func) - (u64(code) + 5);
+    if (distance >= 0x0000000080000000ULL
+     && distance <  0xFFFFFFFF80000000ULL) {
```
linkmauve: As it is checked at many places, make that check a function (it will be more readable for someone who doesn’t know the memory layout of x86 as well).

src/common/code_block.h
```diff
@@ -0,0 +1,87 @@
...
+    void ClearCodeSpace()
+    {
+        PoisonMemory();
+        ResetCodePtr();
+    }
+
+    // Call this when shutting down. Don't rely on the destructor, even though it'll do the job.
+    void FreeCodeSpace()
+    {
+#ifdef __SYMBIAN32__
```
linkmauve: We won’t ever support Symbian.

src/common/hash.cpp
```diff
@@ -0,0 +1,270 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
```
linkmauve: The original code was written by Austin Appleby, and is placed in the public domain. We should probably reflect that here.

linkmauve: Oops, I should have read a bit lower, nvm.

archshift:
Confirmed to work on Linux now.

src/common/x64_emitter.cpp
```diff
@@ -0,0 +1,1989 @@
...
+
+void XEmitter::WriteSIB(int scale, int index, int base)
+{
+    Write8((u8)((scale << 6) | ((index & 7) << 3) | (base & 7)));
+}
+
+void OpArg::WriteRex(XEmitter *emit, int opBits, int bits, int customOp) const
+{
+    if (customOp == -1)       customOp = operandReg;
+#ifdef _M_X86_64
```
linkmauve: I don’t think we want to support anything but amd64 here, maybe removing any else code on that define and doing the check at the beginning of the file would be better.

yuriks: I don't think there's any reason to gut the existing support in the emitter if there's no good reason to do it. The code is there and it's been tested in PPSSPP.

src/common/code_block.h
bunnei: The JIT boilerplate code was taken from Dolphin/PPSSPP and supports several platforms. As mentioned in other comments, I'm not sure there is a point in gutting out support for other platforms just yet, even if we aren't immediately using it. It can always be done in a followup PR.

shinyquagsire23:
Compiles and runs fine on Arch Linux x86-64.

Works.

src/video_core/shader/shader.h
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * This structure contains the state information that needs to be unique for a shader unit. The 3DS
+ * has four shader units that process shaders in parallel. At the present, Citra only implements a
+ * single shader unit that processes all shaders serially. Putting the state information in a struct
+ * here will make it easier for us to parallelize the shader processing later.
+ */
+struct UnitState {
+    // The registers are accessed by the shader JIT using SSE instructions, and are therefore
+    // required to be 16-byte aligned.
+    Math::Vec4<float24> MEMORY_ALIGNED16(input_registers[16]);
```
neobrain: This code is making me uncomfortable because it's performing pointer arithmetic across different members of a struct in an undocumented matter (i.e. by depending on the order and types of these register fields). This code has avoidable potential to break whenever someone (for whatever reason) changes the number of registers per array, order of register blocks, or the types of these registers.

I suggest using this instead:

    union {
         struct {
             Math::Vec4<float24> MEMORY_ALIGNED16(input[16]);
             Math::Vec4<float24> MEMORY_ALIGNED16(output[16]);
             Math::Vec4<float24> MEMORY_ALIGNED16(tempoorary[16]);
        } registers;
        Math::Vec4<float24> MEMORY_ALIGNED16(register_block[48]);
    };
    static_assert(sizeof(registers) == sizeof(register_block), "The compiler unexpectedly misaligned our register arrays");

If `registers.*` are never actually accessed directly, they might as well be dropped and replaced with constexpr indexes pointing into `register_block` (which could then simply be called `registers`, btw).

src/video_core/shader/shader.h
```diff
@@ -0,0 +1,167 @@
...
+        u32 max_offset; // maximum program counter ever reached
+        u32 max_opdesc_id; // maximum swizzle pattern index ever used
+    } debug;
+
+    static size_t InputOffset(const SourceRegister& reg) {
+        UnitState inst; // Sample instance used for calculating register offset for JIT
+
+        switch (reg.GetRegisterType()) {
+        case RegisterType::Input:
+            return (std::size_t)&inst.input_registers[0] - (std::size_t)&inst + reg.GetIndex()*sizeof(inst.input_registers[0]);
```
neobrain: Huh? Didn't this use offsetof before? Why was it dropped?

src/video_core/shader/shader_jit_x64.h
```diff
@@ -0,0 +1,89 @@
...
+    void Compile_LOOP(Instruction instr);
+    void Compile_JMP(Instruction instr);
+    void Compile_CMP(Instruction instr);
+    void Compile_MAD(Instruction instr);
+
+private:
+    void Compile_Block(unsigned stop);
+    void Compile_NextInstr(unsigned* offset);
+
+#if defined(ARCHITECTURE_X64)
```
neobrain: We should just #error out if this file is compiled without this macro being present and instead not include this file on other architectures at all (similarly, _x64 files shouldn't be compiled on other architectures).


Also, `#if defined` = `#ifdef` ;)

src/video_core/shader/shader_jit_x64.h
```diff
@@ -0,0 +1,89 @@
...
+    void Compile_SwizzleSrc(Instruction instr, unsigned src_num, SourceRegister src_reg, Gen::X64Reg dest);
+    void Compile_DestEnable(Instruction instr, Gen::X64Reg dest);
+
+    void Compile_EvaluateCondition(Instruction instr);
+    void Compile_UniformCondition(Instruction instr);
+#endif
+
+    unsigned* offset_ptr = nullptr;
```
neobrain: Can we have some documentation on those members?

src/video_core/shader/shader_jit_x64.h
```diff
@@ -0,0 +1,89 @@
...
+
+    bool done = false;
+    bool looping = false;
+};
+
+class JitShader {
```
neobrain: Documentation?

src/video_core/shader/shader.h
bunnei: Re: https://github.com/bunnei/citra/commit/b6b4537013cc98ee3f742d375c055c82e463aad9#diff-f3410608ab38e242c823da61aa3558a0R109

That does not compile on VS2013 because `InputOffset`/`OutputOffset` are static functions and input_registers, etc. are member variables. Perhaps this would work on VS2015?

src/video_core/shader/shader.h
bunnei: I'm not sure I understand what you are talking about... These registers are only ever accessed via a pointer to a `UnitState` struct + an exact offset within each of the `input_registers`/`output_registers`/`temporary_registers` arrays via `InputOffset`/`OutputOffset`. It does not matter where these are within `UnitState` or what size they are, this should be perfectly safe.

src/video_core/shader/shader.h
neobrain: Is that remark referring to @yuriks's suggestion of using `offsetof(UnitState, input_registers[0]) + index*sizeof(input_registers[0])`? PICA_REG_INDEX (*) is able to compute member offsets in global scope too (and it compiles fine on all compilers that we support), so I'm not sure where the difference would be. For what it's worth, chances are you need to use `offsetof(UnitState, input_registers)` instead of `offsetof(UnitState, input_registers[0])`.

(*) https://github.com/citra-emu/citra/blob/master/src/video_core/pica.h#L29

src/video_core/shader/shader_jit_x64.h
bunnei: Oops yeah... This file is only compiled if ARCHITECTURE_X64 is defined, so I'm not even sure that we need the `#error`... I just forgot to take these out here.

src/video_core/shader/shader.h
neobrain: You are right. I thought I had seen InputOffset used differently before, but in any case none of my concerns anymore in the current revision.

Sorry for the confusion!

bunnei:
Whew :sweat:

I think I've hit all the comments. Should be good for a final review. Recently tested on all 3 supported platforms.

src/video_core/command_processor.cpp
```diff
@@ -210,11 +210,14 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
             // The size has been tuned for optimal balance between hit-rate and the cost of lookup
             const size_t VERTEX_CACHE_SIZE = 32;
             std::array<u16, VERTEX_CACHE_SIZE> vertex_cache_ids;
-            std::array<VertexShader::OutputVertex, VERTEX_CACHE_SIZE> vertex_cache;
+            std::array<Shader::OutputVertex, VERTEX_CACHE_SIZE> vertex_cache;
 
             unsigned int vertex_cache_pos = 0;
             vertex_cache_ids.fill(-1);
 
+            Shader::UnitState shader_unit;
```
Subv: Have a TODO here or something so we don't forget that we should implement the other 3 shader units

Subv:
LGTM after all comments are addressed

src/common/x64/cpu_detect.h
```diff
@@ -0,0 +1,63 @@
...
+    OTHER,
+};
+
+/// x86/x64 CPU capabilities that may be detected by this module
+struct CPUCaps {
+    CPUVendor vendor;
+    char cpu_string[0x21];
+    char brand_string[0x41];
+    bool mode_64_bit;
+    bool htt;
```
yuriks: `hyperthreading`?

src/common/x64/cpu_detect.h
```diff
@@ -0,0 +1,63 @@
...
+    AMD,
+    OTHER,
+};
+
+/// x86/x64 CPU capabilities that may be detected by this module
+struct CPUCaps {
+    CPUVendor vendor;
+    char cpu_string[0x21];
+    char brand_string[0x41];
+    bool mode_64_bit;
```
yuriks: I think this one deserves a comment.

src/common/x64/cpu_detect.h
```diff
@@ -0,0 +1,63 @@
...
+    bool fma;
+    bool fma4;
+    bool aes;
+    // FXSAVE/FXRSTOR
+    bool fxsr;
+    bool movbe;
+    // This flag indicates that the hardware supports some mode in which denormal inputs and outputs
+    // are automatically set to (signed) zero.
+    bool flush_to_zero;
+    bool lahfsahf64;
```
yuriks: u wot m8

src/common/x64/cpu_detect.h
```diff
@@ -0,0 +1,63 @@
...
+    bool long_mode;
```
yuriks: How does this differ from `mode_64_bit`?

src/common/x64/cpu_detect.cpp
```diff
@@ -0,0 +1,225 @@
...
+    else if (!strcmp(cpu_caps.brand_string, "AuthenticAMD"))
+        cpu_caps.vendor = CPUVendor::AMD;
+    else
+        cpu_caps.vendor = CPUVendor::OTHER;
+
+    // Set reasonable default brand string even if brand string not available
+    strcpy(cpu_caps.cpu_string, cpu_caps.brand_string);
+
+    // Detect family and other miscellaneous features
+    bool ht = false;
```
yuriks: This var isn't needed.

src/common/x64/cpu_detect.cpp
```diff
@@ -0,0 +1,225 @@
...
+    cpu_caps.htt = ht;
+    cpu_caps.logical_cpu_count = 1;
+    if (max_std_fn >= 1) {
+        __cpuid(cpu_id, 0x00000001);
+        int family = ((cpu_id[0] >> 8) & 0xf) + ((cpu_id[0] >> 20) & 0xff);
+        int model = ((cpu_id[0] >> 4) & 0xf) + ((cpu_id[0] >> 12) & 0xf0);
+
+        cpu_caps.logical_cpu_count = (cpu_id[1] >> 16) & 0xFF;
+        ht = (cpu_id[3] >> 28) & 1;
```
yuriks: This should set `cpu_caps.htt` instead. Right now this result is unused.

src/common/x64/cpu_detect.cpp
```diff
@@ -0,0 +1,225 @@
...
+        // Check for more features
+        __cpuid(cpu_id, 0x80000001);
+        if (cpu_id[2] & 1) cpu_caps.lahfsahf64 = true;
+        if ((cpu_id[2] >> 5) & 1) cpu_caps.lzcnt = true;
+        if ((cpu_id[2] >> 16) & 1) cpu_caps.fma4 = true;
+        if ((cpu_id[3] >> 29) & 1) cpu_caps.long_mode = true;
+    }
+
+    if (max_ex_fn >= 0x80000008) {
+        // Detect hyper threading
```
yuriks: Isn't the `ht` var above detecting hyper-threading already?

src/common/x64/cpu_detect.cpp
```diff
@@ -0,0 +1,225 @@
...
+    if (max_ex_fn >= 0x80000001) {
+        // Check for more features
+        __cpuid(cpu_id, 0x80000001);
+        if (cpu_id[2] & 1) cpu_caps.lahfsahf64 = true;
+        if ((cpu_id[2] >> 5) & 1) cpu_caps.lzcnt = true;
+        if ((cpu_id[2] >> 16) & 1) cpu_caps.fma4 = true;
+        if ((cpu_id[3] >> 29) & 1) cpu_caps.long_mode = true;
+    }
+
+    if (max_ex_fn >= 0x80000008) {
```
yuriks: Apparently this code doesn't work properly on all CPUs anymore. In addition, it queries if the CPU *supports* hyper-threading, and not if it's currently enabled in the BIOS/OS. So we should be querying the OS for this information instead.

yuriks: Ok, after some research, how to read core and thread topology on each OS:

 - Windows: Use [GetLogicalProcessorInformation](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683194%28v=vs.85%29.aspx). See the example in the MSDN page. From the example, we want `processorCoreCount` for physical cores and then the `logicalProcessorCount` of any one of the cores to get the HT count.
 - Linux: Read files in `/sys/devices/system/cpu/cpu0/topology/`. `core_siblings` contains a hex number which is a bit mask. Count the bits in the mask to get the number of logical cores (HTs). `thread_siblings` has the same format. Count the bits to get the number of HTs in a physical core. Divide both to get the number of physical cores.
 - OS X: Uses the [sysctlbyname](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man3/sysctl.3.html) with `hw.physicalcpu` and `hw.logicalcpu` which contain the respective counts.

src/common/x64/cpu_detect.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include <string>
+#include <thread>
+
+#include "common/common_types.h"
+
+#include "cpu_detect.h"
+
+namespace Common {
+
+CPUCaps cpu_caps = {};
```
yuriks: `static`

src/common/x64/cpu_detect.cpp
```diff
@@ -0,0 +1,225 @@
...
+static u64 _xgetbv(u32 index) {
+    u32 eax, edx;
+    __asm__ __volatile__("xgetbv" : "=a"(eax), "=d"(edx) : "c"(index));
+    return ((u64)edx << 32) | eax;
+}
+
+#endif // ifndef _MSC_VER
+
+// Detects the various CPU features
+static void Detect() {
```
yuriks: Make this return a CPUCaps struct instead of just modifying the global.

src/common/x64/cpu_detect.cpp
yuriks: Alternatively, feel free to just throw out the hyper-threading crap for now, since I doubt we'll be using it anytime soon.

src/common/x64/cpu_detect.cpp
```diff
@@ -0,0 +1,225 @@
...
+                cpu_caps.logical_cpu_count /= cores_x_package;
+            }
+        }
+    }
+}
+
+// Automatically detect CPU capabilities at initialization time
+struct DetectCPU_ {
+    DetectCPU_() { Detect(); }
+} static detect_cpu_;
```
yuriks: Replace this with a static member in `GetCPUCaps` which is initialized with `Detect()`.

src/common/x64/cpu_detect.cpp
yuriks: (If you do the change mentioned below you don't need this global at all.)

src/video_core/shader/shader.cpp
```diff
@@ -0,0 +1,143 @@
...
+#ifdef ARCHITECTURE_X64
+static std::unordered_map<u64, std::unique_ptr<JitShader>> shader_map;
+static JitCompiler jit;
+static JitShader* jit_shader;
+#endif // ARCHITECTURE_X64
+
+void Setup(UnitState& state) {
+#ifdef ARCHITECTURE_X64
+    const u64 cache_key = Common::ComputeHash64(&g_state.vs.program_code, 8192) ^
+            ((u64)g_state.regs.vs.main_offset << 32);
```
yuriks: For correctness, you need to also include the swizzle table in the hash.

To combine the various values, either:
 - Put the program hash, swizzles hash and entrypoint in a std::tuple and use that as the key.
 - Import versions of [boost's `hash_combine_impl`](https://github.com/boostorg/functional/blob/master/include/boost/functional/hash/hash.hpp#L215-L250) and a make a helper func to hash a value and then combine it. Then combine the program hash, swizzle hash and the hash of the entrypoint to get a single u64.

The hash combination functions will also come in handy for other uses, so I hope you do that. :)

For maximum efficiency in the second case, you can make a simple wrapper struct (just `struct HashKey128 { u64 hash[2]; }`) and specialize `std::hash<HashKey>` so that it just folds the words of the hash using xor. This will prevent the standard library from re-hashing the key unnecessarily (since it's already a pseudo-random hash.)

src/video_core/shader/shader.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+namespace Pica {
+
+namespace Shader {
+
+#ifdef ARCHITECTURE_X64
+static std::unordered_map<u64, std::unique_ptr<JitShader>> shader_map;
+static JitCompiler jit;
+static JitShader* jit_shader;
+#endif // ARCHITECTURE_X64
```
yuriks: It would be good if the shader caching infrastructure and the shader JIT interfaces could be unified. This would also remove explicit checks and ifdefs from various parts of the code. I don't think it's a big mark against this PR, so I'm willing to do it myself as a follow up PR so this isn't stalled even more.

src/video_core/shader/shader.cpp
```diff
@@ -0,0 +1,143 @@
...
+#ifdef ARCHITECTURE_X64
+    const u64 cache_key = Common::ComputeHash64(&g_state.vs.program_code, 8192) ^
+            ((u64)g_state.regs.vs.main_offset << 32);
+
+    auto iter = shader_map.find(cache_key);
+
+    if (iter != shader_map.end()) {
+        jit_shader = iter->second.get();
+    } else {
+        std::unique_ptr<JitShader> new_shader = Common::make_unique<JitShader>();
```
yuriks: You should directly store the `JitShader` (as opposed to a `unique_ptr` to it) to remove one level of indirection.

src/video_core/shader/shader_jit_x64.h
```diff
@@ -0,0 +1,96 @@
...
+};
+
+/**
+ * This class is used to store cached x86_x64 code for a given shader program.
+ */
+class JitShader {
+public:
+    JitShader();
+
+    void DoJit(JitCompiler& jit);
```
yuriks: Why doesn't `JitCompiler` simply return an instance of `JitShader` instead of having this odd indirection?

yuriks: (In fact, there doesn't seem to be a point to this wrapper struct at all.)

src/common/platform.h
```diff
@@ -27,7 +27,7 @@
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Platform detection
 
-#if defined(__x86_64__) || defined(_M_X64) || defined(__aarch64__)
+#if defined(__x86_64__) || defined(ARCHITECTURE_X64) || defined(__aarch64__)
```
yuriks: Either change this one to `_M_X64` which is the macro actually set by MSVC (so it's coherent with what the code was attempting to do), or remove the `__x86_64__` test which is now redundant.

src/common/x64/abi.cpp
```diff
@@ -0,0 +1,680 @@
...
+#include "emitter.h"
+
+using namespace Gen;
+
+// Shared code between Win64 and Unix64
+
+// Sets up a __cdecl function.
+void XEmitter::ABI_EmitPrologue(int maxCallParams)
+{
+#ifdef _M_IX86
```
yuriks: Acoording to predef.sf.net, this isn't set by gcc, so compiling this will fail in x86. Should replace these with `ARCHITECTURE_x86` I suppose.

CMakeLists.txt
```diff
@@ -11,8 +11,8 @@ if(NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git/hooks/pre-commit)
 endif()
 
 if (NOT MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -pthread")
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -pthread -msse4.2")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread -msse4.2")
```
yuriks: `-msse4.2` should only be passed if we're indeed on x86/x86_64.

Note to yourself: "this can be -msse4.1 (I think) now that we removed Dolphin's hash functions"

CMakeLists.txt
```diff
@@ -51,6 +51,17 @@ else()
     set(CMAKE_EXE_LINKER_FLAGS_RELEASE "/DEBUG /INCREMENTAL:NO /OPT:REF,ICF" CACHE STRING "" FORCE)
 endif()
 
+# Platform-agnostic definition to check if we are on x86_64
+if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "^x86" OR
+   ${CMAKE_SYSTEM_PROCESSOR} MATCHES "i.86" OR
+   ${CMAKE_SYSTEM_PROCESSOR} MATCHES "[aA][mM][dD]64" OR
+   APPLE)
```
yuriks: Most of these are checking for x86, not x86_64, which makes the check below seem somewhat redundant if you could just directly detect only x86_64. And what's up with `APPLE`? Why can't we detect using the CPU string on OS X too?

On Windows, this is `AMD64` and on Linux and OS X it's `x86_64`, so you can just match against those two and know you're on x86-64.
For x86, Windows has `x86`, Linux and OS X `i.86` (where `.` is any of 3, 4, 5 or 6) so likewise for detecting x86.

CMakeLists.txt
```diff
@@ -51,6 +51,17 @@ else()
...
+    if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+        set(ARCHITECTURE_X64 1)
+        add_definitions(-DARCHITECTURE_X64=1)
```
yuriks: `ARCHITECTURE_x86_64`*

src/common/common_funcs.h
```diff
@@ -35,7 +35,7 @@
 
 #ifndef _MSC_VER
 
-#if defined(__x86_64__) || defined(_M_X64)
+#if defined(__x86_64__) || defined(ARCHITECTURE_X64)
```
yuriks: The check for `__x86_64__` is redundant.

src/video_core/shader/shader_jit_x64.h
```diff
@@ -0,0 +1,96 @@
...
+    void Compile_Block(unsigned stop);
+    void Compile_NextInstr(unsigned* offset);
+
+    void Compile_SwizzleSrc(Instruction instr, unsigned src_num, SourceRegister src_reg, Gen::X64Reg dest);
+    void Compile_DestEnable(Instruction instr, Gen::X64Reg dest);
+
+    void Compile_EvaluateCondition(Instruction instr);
+    void Compile_UniformCondition(Instruction instr);
+
+    /// Pointer to the variable that stores the current code offset to handle nested code blocks
```
yuriks: "Pointer to the variable that stores the current Pica code offset. Used to handle nested code blocks."

src/video_core/shader/shader_jit_x64.h
```diff
@@ -0,0 +1,96 @@
...
+using nihstro::SwizzlePattern;
+
+namespace Pica {
+
+namespace Shader {
+
+using CompiledShader = void(void* state);
+
+/**
+ * This class implements the shader Just-In-Time (JIT) compiler. It decodes the current Pica vertex
```
yuriks: Feels kinda academic to expand "JIT" here. :)

"This class implements the shader JIT compiler. It recompiles a Pica shader program into x86_64 code that can be executed on the host machine directly."

src/common/hash.h
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+
+namespace Common {
+
+void MurmurHash3_128(const void* key, int len, u32 seed, void* out);
```
yuriks: IMO you should keep the original `MurmurHash3_x64_128` name, since they're all distinct algorithms.

src/video_core/shader/shader_jit_x64.h
```diff
@@ -0,0 +1,96 @@
...
+ * shader and leverages the x86_x64 emitter within our Common code to generate equivalent x86_x64
```
yuriks: either way, x86_64*

src/video_core/shader/shader.cpp
bunnei: This already _is_ hashing `swizzle_data`, in a non-obvious way :D

src/common/x64/abi.cpp
bunnei: It's not, since we don't support x86.... It's just part of the original code. These files aren't even compiled if you're not on x64, so the x86 build should still be fine.

So I don't think it makes sense trying to fix it for x86, but I also don't see a point of gutting out any x86 code from these modules (just yet) just in case we have a good reason to support 32-bit anytime soon.

src/common/hash.h
bunnei: Given that this is *the* hash algorithm in Citra, and it works on all platforms (not just x86_64), and that we'll likely never have the x86 variant... I don't see the point. Having x64 in the name makes it seem like a platform-specific function, when for our purposes it's not... It's just the best choice.

hrydgard:
Nitpicking CPUDetect.cpp/h for a pull request like this that really is nothing about detecting CPUs seems pointless and irrelevant, it's wholesale imported from Dolphin/PPSSPP (I take some responsibility for its general shoddiness, but it's been adequate for my needs). 

I would recommend to either separately merge CPUDetect and bring it up to your standards first, if it's so important, or simply merge it and clean it up afterwards. Blocking a major feature pull request on side stuff seems odd.

src/common/hash.h
yuriks: Then at least leave a comment indicating which hash function this function implements.

src/common/x64/abi.cpp
yuriks: Ah, good point about these not being compiled under x86 in the first place.

src/video_core/command_processor.cpp
yuriks: I separating the shaders units as an implementation detail really important? I imagine we'll end up having one for the geometry shader (since that does use separate parameters and all) but for the vertex shader we'll just be splitting them according to any threading we do.

src/video_core/shader/shader.cpp
yuriks: Oh. :| You should do 2 separate hashes because there's really no guarantee the two arrays will be always adjacent on the struct (and indeed there's not even a comment there telling people not to separate them) and it's very non-obvious that's going on.

src/video_core/shader/shader.cpp
bunnei: Yeah I'll think I defer this to a followup PR

src/common/hash.h
bunnei: Added it to the .cpp

bunnei:
can I get an 'L'
can I get a 'G'
can I get a 'T'
can I get an 'M'
What does it spell?

src/common/x64/cpu_detect.cpp
yuriks: Looks like you forgot to remove this var.

src/video_core/shader/shader.cpp
```diff
@@ -0,0 +1,149 @@
...
+        u64 cache_key = (Common::ComputeHash64(&g_state.vs.program_code, sizeof(g_state.vs.program_code)) ^
+            Common::ComputeHash64(&g_state.vs.swizzle_data, sizeof(g_state.vs.swizzle_data)) ^
+            g_state.regs.vs.main_offset);
+
+        auto iter = shader_map.find(cache_key);
+        if (iter != shader_map.end()) {
+            jit_shader = iter->second;
+        } else {
+            jit_shader = jit.Compile();
+            shader_map.emplace(cache_key, std::move(jit_shader));
```
yuriks: You don't want to `std::move`.

yuriks:
**M**agic **T**he **G**athering: **L**ive?

LGTM except for the *two* comments.

bunnei:
Final call for comments on this. Otherwise I plan on hitting the last few comments above, rebasing, and merging within the next day.

neobrain:
```
src/./video_core/shader/shader.h: In static member function ‘static int Pica::Shader::UnitState::InputOffset(const nihstro::SourceRegister&)’:
src/./video_core/shader/shader.h:119:45: warning: invalid access to non-static data member ‘Pica::Shader::UnitState::input_registers’  of NULL object [-Winvalid-offsetof]
             return (int)offsetof(UnitState, input_registers) + reg.GetIndex()*sizeof(Math::Vec4<float24>);
                                             ^
src/./video_core/shader/shader.h:119:45: warning: (perhaps the ‘offsetof’ macro was used incorrectly) [-Winvalid-offsetof]
src/./video_core/shader/shader.h:122:45: warning: invalid access to non-static data member ‘Pica::Shader::UnitState::temporary_registers’  of NULL object [-Winvalid-offsetof]
             return (int)offsetof(UnitState, temporary_registers) + reg.GetIndex()*sizeof(Math::Vec4<float24>);
                                             ^
src/./video_core/shader/shader.h:122:45: warning: (perhaps the ‘offsetof’ macro was used incorrectly) [-Winvalid-offsetof]
src/./video_core/shader/shader.h: In static member function ‘static int Pica::Shader::UnitState::OutputOffset(const nihstro::DestRegister&)’:
src/./video_core/shader/shader.h:133:45: warning: invalid access to non-static data member ‘Pica::Shader::UnitState::output_registers’  of NULL object [-Winvalid-offsetof]
             return (int)offsetof(UnitState, output_registers) + reg.GetIndex()*sizeof(Math::Vec4<float24>);
                                             ^
src/./video_core/shader/shader.h:133:45: warning: (perhaps the ‘offsetof’ macro was used incorrectly) [-Winvalid-offsetof]
src/./video_core/shader/shader.h:136:45: warning: invalid access to non-static data member ‘Pica::Shader::UnitState::temporary_registers’  of NULL object [-Winvalid-offsetof]
             return (int)offsetof(UnitState, temporary_registers) + reg.GetIndex()*sizeof(Math::Vec4<float24>);
                                             ^
src/./video_core/shader/shader.h:136:45: warning: (perhaps the ‘offsetof’ macro was used incorrectly) [-Winvalid-offsetof]
```

neobrain:
When running 3dscraft with this, I'm getting this output:
```
[   3.535764] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x20 (0x8006c368)
[   3.535798] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x20 (0x8006c364)
[   3.535809] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x20 (0x8006c362)
[   3.535820] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x20 (0x8006c361)
[   3.535831] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x20 (0x8006caae)
[   3.535841] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x20 (0x8006c36f)
[   3.535852] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x20 (0x8006c36e)
[   3.535862] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x20 (0x8006caef)
[   3.535872] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x20 (0x8006dfee)
[   3.535883] HW.GPU <Critical> video_core/shader/shader_jit_x64.cpp:Compile_NextInstr:627: Unhandled instruction: 0x11 (0x454c5644)
```

One of these is NOP, the other is indeed an unknown instruction. These errors do not show up in the shader interpreter.

yuriks:
re-LGTM if @neobrain confirms the warnings are gone

bunnei:
:boom: :star2: :heart_eyes_cat: :fireworks: :cake:

shinyquagsire23:
10/10

yuriks:
At long last, it's fast.

MoochMcGee:
Welp, now the slowest part of the emulator is the CPU interpreter, if this video of an extremely glitchy Super Mario 3D Land is to be believed: https://www.youtube.com/watch?v=JkXrWcAsBd4

yuriks:
@MoochMcGee Depends on the game. Some are very CPU heavy (like CN) but for most I've seen the bottlenecks are still in the drawing part of things. Both are things that should definitely be worked on independently, though.

PR #1004, opened 2015-07-28
Enable linker optimizations in MSVC Release builds

yuriks:
Apparently /DEBUG implicitly disables linker optimizations. This
explicitly re-enables them, giving a 40% reduction in binary sizes and
a very slight (not even 1ms) runtime speed improvement.


bunnei:
LGTM

PR #873, opened 2015-06-20
Move input values into an array (#870)

jroweboy:
As @neobrain suggested in #870 I moved the input mappings into an array and use an enum of Key Values to index the array.

PR #1005, opened 2015-07-28
Settings: Fix saving wrong values for input configuration

chinhodado:
There was a mistake in #873 where the values that get saved are just the index of the buttons, not the key mapped to the buttons.


yuriks:
LGTM

PR #1007, opened 2015-07-29
dyncom: Handle left-operand PC correctly for data-processing ops

lioncash:
This is considered deprecated in the ARM manual (using PC as an operand), however, this is still able to be executed on the MPCore (which I'm quite sure would be rare to begin with).


bunnei:
LGTM

PR #1008, opened 2015-07-29
dyncom: Handle the case where PC is the source register for STR/VSTM/VLDM

lioncash:
More corner-case handling of rare possible operations.


bunnei:
LGTM

PR #1015, opened 2015-07-31
Videocore: Implement simple vertex caching

yuriks:
This gives a ~2/3 reduction in the amount of vertices that need to be
processed through the vertex loaders and the vertex shader, yielding a
good speedup.

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -206,88 +206,115 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
                 std::map<u32, u32> ranges;
             } memory_accesses;
 
+            // Simple circular-replacement vertex cache
+            // The size has been tuned for optimal balance between hit-rate and the cost of lookup
+            const size_t VERTEX_CACHE_SIZE = 32;
+            std::array<u16, VERTEX_CACHE_SIZE> vertex_cache_ids;
+            std::array<VertexShader::OutputVertex, VERTEX_CACHE_SIZE> vertex_cache;
+
+            unsigned int vertex_cache_pos = 0;
+            vertex_cache_ids.fill(-1);
+
             for (unsigned int index = 0; index < regs.num_vertices; ++index)
             {
                 unsigned int vertex = is_indexed ? (index_u16 ? index_address_16[index] : index_address_8[index]) : index;
 
+                // -1 is a common special value used for primitive restart. Since it's unknown if
+                // the PICA supports it, and it would mess up the caching, guard against it here.
+                ASSERT(vertex != -1);
+
+                bool vertex_cache_hit = false;
+                VertexShader::OutputVertex output;
+
                 if (is_indexed) {
-                    // TODO: Implement some sort of vertex cache!
                     if (g_debug_context && Pica::g_debug_context->recorder) {
                         int size = index_u16 ? 2 : 1;
                         memory_accesses.AddAccess(base_address + index_info.offset + size * index, size);
                     }
-                }
 
-                // Initialize data for the current vertex
-                VertexShader::InputVertex input;
-
-                for (int i = 0; i < attribute_config.GetNumTotalAttributes(); ++i) {
-                    if (vertex_attribute_elements[i] != 0) {
-                        // Default attribute values set if array elements have < 4 components. This
-                        // is *not* carried over from the default attribute settings even if they're
-                        // enabled for this attribute.
-                        static const float24 zero = float24::FromFloat32(0.0f);
-                        static const float24 one = float24::FromFloat32(1.0f);
-                        input.attr[i] = Math::Vec4<float24>(zero, zero, zero, one);
-
-                        // Load per-vertex data from the loader arrays
-                        for (unsigned int comp = 0; comp < vertex_attribute_elements[i]; ++comp) {
-                            u32 source_addr = vertex_attribute_sources[i] + vertex_attribute_strides[i] * vertex + comp * vertex_attribute_element_size[i];
-                            const u8* srcdata = Memory::GetPhysicalPointer(source_addr);
-
-                            if (g_debug_context && Pica::g_debug_context->recorder) {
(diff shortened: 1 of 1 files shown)


hrydgard:
Ah right, this is to avoid reprocessing vertices that appear multiple times in like a strip or an index triangle list, I presume. But, couldn't you simply first process all the the vertices and then apply indexing afterwards instead, either through hardware or manually? No need for a cache then.

src/video_core/command_processor.cpp
```diff
@@ -206,88 +206,110 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
                 std::map<u32, u32> ranges;
             } memory_accesses;
 
+            // Simple circular-replacement vertex cache
+            const size_t VERTEX_CACHE_SIZE = 32;
+            std::array<u16, VERTEX_CACHE_SIZE> vertex_cache_ids;
+            std::array<VertexShader::OutputVertex, VERTEX_CACHE_SIZE> vertex_cache;
+
+            unsigned int vertex_cache_pos = 0;
+            vertex_cache_ids.fill(-1);
```
neobrain: -1 is a common index value to use for primitive restart. This is not OGL ES 1.x or 2.x functionality, but it might be supported by PICA200 regardless. Maybe add some asserts about that to make sure the index never becomes 65535?

src/video_core/command_processor.cpp
```diff
@@ -206,88 +206,110 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+
             for (unsigned int index = 0; index < regs.num_vertices; ++index)
             {
                 unsigned int vertex = is_indexed ? (index_u16 ? index_address_16[index] : index_address_8[index]) : index;
 
+                bool cache_hit = false;
```
neobrain: *vertex_cache_hit

yuriks:
@hrydgard My experimentation with different vertex cache types and sizes indicates that the cache hit rate for this scheme is almost optimal already. Huge cache sizes didn't improve hit rate significantly, indicating that we wouldn't gain much from guaranteeing that we only process each vertex once. At the same time, that would open the possibility of transforming vertices that won't be used at all (if a game puts several different meshes inside the same buffer), so I think it's not a good trade-off.

yuriks:
Updated

src/video_core/command_processor.cpp
```diff
@@ -206,88 +206,114 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+            unsigned int vertex_cache_pos = 0;
+            vertex_cache_ids.fill(-1);
+
             for (unsigned int index = 0; index < regs.num_vertices; ++index)
             {
                 unsigned int vertex = is_indexed ? (index_u16 ? index_address_16[index] : index_address_8[index]) : index;
 
+                // -1 is a common special value used for primitive restart. Since it's unknown if
+                // the PICA supports it, and it would mess up the caching, guard against it here.
+                ASSERT(index != -1);
```
Subv: How could this assert be hit? regs.num_vertices would have to be Max_u32 + 1 for index to ever reach (u32)(-1)

yuriks: Oops, wrong variable, good catch.

Subv:
LGTM instamergepls

src/video_core/command_processor.cpp
```diff
@@ -206,88 +206,114 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
                 std::map<u32, u32> ranges;
             } memory_accesses;
 
+            // Simple circular-replacement vertex cache
+            const size_t VERTEX_CACHE_SIZE = 32;
```
bunnei: Perhaps add a comment explaining why 32... IIRC you did some experimentation on this and somehow determined that 32 was somewhat ideal

src/video_core/command_processor.cpp
```diff
@@ -206,88 +206,114 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                            // Load the default attribute if we're configured to do so
+                            input.attr[i] = g_state.vs.default_attributes[i];
+                            LOG_TRACE(HW_GPU, "Loaded default attribute %x for vertex %x (index %x): (%f, %f, %f, %f)",
+                                      i, vertex, index,
+                                      input.attr[i][0].ToFloat32(), input.attr[i][1].ToFloat32(),
+                                      input.attr[i][2].ToFloat32(), input.attr[i][3].ToFloat32());
+                        } else {
+                            // TODO(yuriks): In this case, no data gets loaded and the vertex remains
+                            //              with the last value it had. This isn't currently maintained
+                            //              as global state, however, and so won't work in Cita yet.
```
bunnei: *Citra

:unamused:

src/video_core/command_processor.cpp
```diff
@@ -206,88 +206,114 @@ static inline void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                        } else if (attribute_config.IsDefaultAttribute(i)) {
+                            // Load the default attribute if we're configured to do so
+                            input.attr[i] = g_state.vs.default_attributes[i];
+                            LOG_TRACE(HW_GPU, "Loaded default attribute %x for vertex %x (index %x): (%f, %f, %f, %f)",
+                                      i, vertex, index,
+                                      input.attr[i][0].ToFloat32(), input.attr[i][1].ToFloat32(),
+                                      input.attr[i][2].ToFloat32(), input.attr[i][3].ToFloat32());
+                        } else {
+                            // TODO(yuriks): In this case, no data gets loaded and the vertex remains
+                            //              with the last value it had. This isn't currently maintained
```
bunnei: Align this text with 'In' (above), or don't tab it over at all, IMO

bunnei:
A couple of comments on comments, other than that this LGTM after a squash.

yuriks:
Updated, rebased and squashed.

PR #1018, opened 2015-08-02
Handle invalid `Log::Level::Count`

bbarenblat:
Add a case of `Log::Level::Count` to all switch statements that dispatch on `Log::Level`.  The case simply asserts `false` and notes the invalid log level.

This removes a couple warnings under Clang.

### DIFF
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -6,6 +6,7 @@
 #include <array>
 #include <cstdio>
 
+#include "common/assert.h"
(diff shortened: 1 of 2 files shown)


yuriks:
We have a `UNREACHABLE` macro which I think would be better to use for this.

bbarenblat:
I agree.  Normally I’d think of `UNREACHABLE` as indicating code that actually cannot be reached (because, for example, an `exit` occurs immediately prior), but it looks like `UNREACHABLE`

bunnei:
LGTM

PR #1019, opened 2015-08-02
Common: Work around bug in MSVC2015 standard library

yuriks:
The char16_t/char32_t implementations aren't present in the library and
cause linker errors. This is a known issue that wasn't fixed in VS2015
RTM.


lioncash:
Should leave comments indicating this (and to switch over whenever it's fixed)

yuriks:
Updated

Subv:
lgtm

bunnei:
LGTM

PR #1020, opened 2015-08-03
CMake: Add option to download Qt binaries

yuriks:
This adds an option in CMake that'll download a binary package of Qt binaries. With this (and the #1019 workaround also included in this branch) it's possible to compile Citra with VS2015.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,23 @@
 # dependent libraries.
 cmake_minimum_required(VERSION 2.8.11)
 
+include(CheckSymbolExists)
+function(detect_architecture symbol arch)
+    if (NOT DEFINED ARCHITECTURE)
+        set(CMAKE_REQUIRED_QUIET 1)
+        check_symbol_exists("${symbol}" "" ARCHITECTURE_${arch})
+        unset(CMAKE_REQUIRED_QUIET)
+
+        # The output variable needs to be unique across invocations otherwise
+        # CMake's crazy scope rules will keep it defined
+        if (ARCHITECTURE_${arch})
+            set(ARCHITECTURE "${arch}" PARENT_SCOPE)
+            set(ARCHITECTURE_${arch} 1 PARENT_SCOPE)
+            add_definitions(-DARCHITECTURE_${arch}=1)
+        endif()
+    endif()
+endfunction()
+
 project(citra)
 
 if(NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git/hooks/pre-commit)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/externals/cmake-modules")
 find_package(OpenGL REQUIRED)
 include_directories(${OPENGL_INCLUDE_DIR})
 
-option(ENABLE_GLFW "Enable the GLFW frontend" ON)
+option(ENABLE_GLFW "Enable the GLFW frontend" OFF)
```
jroweboy: Does the GLFW build just not work on VS2015? Just wondering why you are disabling it by default.

yuriks: It also needs a separate set of binaries, and I haven't built those yet.

CMakeLists.txt
```diff
@@ -153,12 +153,31 @@ ELSE()
...
+            set(QTVER qt5.5-msvc2015_64)
+        else()
+            message(FATAL_ERROR "No bundled Qt binaries for your toolchain. Disable CITRA_USE_BUNDLED_QT and provide your own.")
+        endif()
+
+        set(QTPATH "${CMAKE_BINARY_DIR}/externals/${QTVER}")
+        if (NOT EXISTS "${QTPATH}")
+            message(STATUS "Downloading Qt binaries for ${QTVER}...")
+            file(DOWNLOAD
+                https://github.com/yuriks/citra-depends/raw/master/${QTVER}.7z
```
bunnei: Submodule? I've kind of wanted to do this for a while anyway to alleviate the burden of Windows users having to download/install a 600MB binary. FWIW, also, what's the point of having this be nested zips?

yuriks: Nested zips?

Not a submodule so that people don't need to download all binaries for all platforms ever. Just the one they're going to use.

neobrain: If this stays as is, the repo should be hosted in the citra-emu organization.

yuriks:
Btw, if we go forward with this approach, then I plan on using the same system for GLFW or SDL2 binaries. (This decision is in fact blocking #841 right now.)

Subv:
What's the status of this RFC? I've been using the patch for a while now and it seems to be the best way to support VS2015 without having to wait for Qt to update their binaries.

yuriks:
@Subv: Was just waiting for more of a reaction. Seems like we agree to go with this, so I'll clean up the code and add a GLFW build.

yuriks:
Updated.
  - Added GLFW libraries for 2015 and 2013
  - Extracted downloading and extracting code to a CMake function
  - Fixed architecture detection in CMake (will split up in another commit later)

yuriks:
I think this is pretty much done. Just need to review and then create an official repo for the files before merging.

yuriks:
Updated fixing CI failures.

bunnei:
Finally got around to testing this after struggling a bit with my environment. Everything checks out fine here. Let's move forward with setting up the repo for the binaries in the Citra org.

yuriks:
Updated. Let's just check that the AppVeyor build is working fine.

bunnei:
LGTM

PR #1023, opened 2015-08-06
OpenGL: Fix state tracking in situations with reused object handles

yuriks:
If an OpenGL object is created, bound to a binding using the state
tracker, and then destroyed, a newly created object can be assigned the
same numeric handle by OpenGL. However, even though it is a new object,
and thus needs to be bound to the binding again, the state tracker
compared the current and previous handles and concluded that no change
needed to be made, leading to failure to bind objects in certain cases.

This manifested as broken text in VVVVVV, which this commit fixes along
with similar texturing problems in other games.


bunnei:
LGTM

bunnei:
If no one has any more feedback on this, I'm going to merge it by the end of the day.

Subv:
LGTM

PR #1025, opened 2015-08-06
Kernel: Correct(er) handling of Heap and Linear Heap allocations

yuriks:
Revival of #940.

This PR is a grab bag of revamps to the memory handling system implementing more proper tracking of memory allocations on the heap. Along the way some more future-facing infrastructure changes got added in.

In addition to making basic svcControlMemory COMMIT and FREE operations work correctly, this also adds support for different system memory layouts (it currently still always uses a default as before however) and for the expanded linear heap location used by newer kernels. (Required for games that dual-target the o3DS/n3DS.)

The APT shared font was still made to work via a hack, but I now understand what is required for a correct implementation and believe my current direction does not conflict with that.


src/core/hle/kernel/memory.h
```diff
@@ -4,10 +4,27 @@
 
 #pragma once
 
+#include <memory>
+
```
lioncash: should include vector as well.

Subv:
lgtminstamerge.

More seriously, I'll review this tomorrow (Please ping me if i forget).

src/common/logging/log.h
```diff
@@ -91,17 +91,16 @@ void LogMessage(Class log_class, Level log_level,
 } // namespace Log
 
 #define LOG_GENERIC(log_class, log_level, ...) \
-    ::Log::LogMessage(::Log::Class::log_class, ::Log::Level::log_level, \
-        __FILE__, __LINE__, __func__, __VA_ARGS__)
+    ::Log::LogMessage(log_class, log_level, __FILE__, __LINE__, __func__, __VA_ARGS__)
```
Subv: What's up with these changes?

src/core/hle/kernel/memory.cpp
```diff
@@ -0,0 +1,134 @@
...
+    {0x0B200000, 0x02E00000, 0x02000000}, // 7
+};
+
+void MemoryInit(u32 mem_type) {
+    // TODO(yuriks): On the n3DS, all o3DS configurations (<=5) are forced to 6 instead.
+    ASSERT_MSG(mem_type <= 5, "New 3DS memory configuration aren't supported yet!");
+    ASSERT(mem_type != 1);
+
+    VAddr base = 0;
+    for (int i = 0; i < 3; ++i) {
```
Subv: Can you add a comment that says "There are 3 regions: data, rodata and text", either here or in the declaration of memory_regions?

src/core/hle/kernel/memory.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+namespace {
+
+struct MemoryArea {
+    u32 base;
+    u32 size;
+    const char* name;
+};
+
+// We don't declare the IO regions in here since its handled by other means.
```
Subv: What other means?

src/core/hle/kernel/process.cpp
```diff
@@ -103,20 +104,151 @@ void Process::ParseKernelCaps(const u32* kernel_caps, size_t len) {
...
+    return RESULT_SUCCESS;
+}
+
+ResultVal<VAddr> Process::LinearAllocate(VAddr target, u32 size, VMAPermission perms) {
+    auto& linheap_memory = memory_region->linear_heap_memory;
+
+    VAddr heap_end = GetLinearHeapBase() + (u32)linheap_memory->size();
+    // Games and homebrew only ever seem to pass 0 here (which lets the kernel decide the address),
+    // but explicit addresses are also accepted and respected.
+    if (target == 0) target = heap_end;
```
Subv: newline

src/core/hle/kernel/process.cpp
```diff
@@ -103,20 +104,151 @@ void Process::ParseKernelCaps(const u32* kernel_caps, size_t len) {
...
+
+    return MakeResult<VAddr>(target);
+}
+
+ResultCode Process::LinearFree(VAddr target, u32 size) {
+    auto& linheap_memory = memory_region->linear_heap_memory;
+
+    if (target < GetLinearHeapBase() || target + size > GetLinearHeapLimit() ||
+        target + size < target) {
+
```
Subv: empty line

src/core/hle/kernel/memory.cpp
yuriks: But that's not what this is about at all. These are for the APPLICATION, SYSTEM and BASE kernel allocation regions. I'll point that out more explicitly somewhere.

src/core/hle/kernel/process.cpp
yuriks: That's because the if condition was continued on the next line.

src/core/hle/kernel/memory.cpp
yuriks: Actually none, lol, since we don't support MMIO since it's not used by games. I just moved this from the other file, so I'm going to keep it around for now. This array should eventually disappear anyway.

src/common/logging/log.h
yuriks: They serve to allow more control over logging in LogLayout. (So that it can be controlled via a parameter.) See the associated commit.

yuriks:
Addressed comments so far. (I think, tell me if I missed any, working with both bunnei's per-commit comments and the full-diff comments was kinda weird.)

src/core/hle/kernel/thread.cpp
```diff
@@ -117,6 +117,7 @@ void Thread::Stop() {
     wait_objects.clear();
 
     Kernel::g_current_process->used_tls_slots[tls_index] = false;
+    g_current_process->misc_memory_used -= Memory::TLS_ENTRY_SIZE;
```
Subv: You had 0x1000 here before, but TLS_ENTRY_SIZE is 0x200 (Which is the correct value according to 3dbrew), is that ok?

Subv:
Left a comment, other than that LGTM

src/core/hle/kernel/thread.cpp
yuriks: Yes, that was a mistake.

bunnei:
LGTM I think. Not sure the easiest way to access my old comments to verify there were all addressed... But I reread it and didn't see anything glaring.

PR #1026, opened 2015-08-06
arm_disasm: Remove unnecessary code

lioncash:
This part of disassembly only determines the opcode, there's no need for offset calculation here.

Also gets rid of higher level compiler warnings on OSX.


aroulin:
LGTM

PR #1027, opened 2015-08-07
debugger: Add the ability to view VFP register contents

lioncash:


neobrain:
I suppose all of this should be done with a custom QTreeModel instead, but well.. until it explodes, this is okay, I guess.

Also, remind me to add a static member function to BitField to extract a BitField value from its raw storage value via `BitField<pos, size, uint32_t>::GetFrom(value)`.

neobrain:
Other than that, LGTM.

bunnei:
LGTM

PR #1030, opened 2015-08-12
Stop defining GCC always_inline attributes as __forceinline

archshift:
`__forceinline` is a MSVC extension, which may confuse some people working on the codebase.
Furthermore, the C++ standard dictates that all names which contain adjacent underscores are reserved.


yuriks:
LGTM

bunnei:
LGTM

PR #1031, opened 2015-08-15
Handle invalid `Log::Class`

bbarenblat:
Add a case of `Log::Class::Count` to the switch statement that dispatches on `Log::Class`.  The case simply calls the `UNREACHABLE` macro.

Like PR #1018, this removes a warning under Clang.


yuriks:
LGTM

PR #1033, opened 2015-08-16
Handle `FileType::CIA` in `switch` statements

bbarenblat:
Give `FileType::CIA` a string representation and indicate that CIA loading support is unimplemented.

This removes a few warnings under Clang.


yuriks:
LGTM

PR #985, opened 2015-07-24
GPU: HILO8/RG8

ILOVEPIE:
New PR for HILO8/RG8 see PR #773
Has been tested with this: https://github.com/yuriks/3DS_gpu_tests/tree/hilo8-test

PR #1034, opened 2015-08-16
videocore: Added RG8 texture support

yuriks:
Manual address and rebase of #985.

Verified correct against hardware via hwtest: https://github.com/yuriks/3DS_gpu_tests/tree/hilo8-test


neobrain:
LGTM

Just curious though, have you verified that the alpha component which gets read indeed is 255?

Subv:
LGTM

yuriks:
@neobrain Yeah, I had when. (Using the TEVs to broadcast alpha to RGB.)

bunnei:
Thanks for PRing this. For the record, it is HILO8, but I don't want to hold this PR up any longer over semantics.

PR #1035, opened 2015-08-16
Fix building under MinGW

darkf:
Fixes a few minor issues to let Citra compile under MinGW.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,7 +155,8 @@ IF (APPLE)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -stdlib=libc++")
 ELSEIF(MINGW)
     # GCC does not support codecvt, so use iconv instead
-    set(PLATFORM_LIBRARIES winmm ws2_32 iconv)
+    # PSAPI is the Process Status API
+    set(PLATFORM_LIBRARIES winmm ws2_32 psapi iconv)
 
     # WSAPoll functionality doesn't exist before WinNT 6.x (Vista and up)
     add_definitions(-D_WIN32_WINNT=0x0600)
--- a/src/common/common_funcs.h
+++ b/src/common/common_funcs.h
@@ -45,14 +45,20 @@
 
 // GCC 4.8 defines all the rotate functions now
 // Small issue with GCC's lrotl/lrotr intrinsics is they are still 32bit while we require 64bit
-#ifndef _rotl
-inline u32 _rotl(u32 x, int shift) {
+#ifdef _rotl
+#define rotl _rotl
+#else
+inline u32 rotl(u32 x, int shift) {
     shift &= 31;
     if (!shift) return x;
     return (x << shift) | (x >> (32 - shift));
 }
+#endif
 
-inline u32 _rotr(u32 x, int shift) {
+#ifdef _rotr
+#define rotr _rotr
+#else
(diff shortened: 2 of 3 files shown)


mailwl:
how about CLion? Tryed under mac os, it compiles, but don't stop at breakpoints while debug.
And CLion/OsX don't understand new c++ standart

lioncash:
I think he's fine with using MinGW if he took the time to make this PR, lol.

CMakeLists.txt
```diff
@@ -155,7 +155,7 @@ IF (APPLE)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -stdlib=libc++")
 ELSEIF(MINGW)
     # GCC does not support codecvt, so use iconv instead
-    set(PLATFORM_LIBRARIES winmm ws2_32 iconv)
+    set(PLATFORM_LIBRARIES winmm ws2_32 psapi iconv)
```
neobrain: Mind adding a comment explaining why this is needed?

src/common/common_funcs.h
```diff
@@ -45,18 +45,23 @@
...
+// MinGW also ships with _rotl and _rotr defined, not under a macro, so we redefine it here.
 #ifndef _rotl
-inline u32 _rotl(u32 x, int shift) {
+inline u32 rotl(u32 x, int shift) {
     shift &= 31;
     if (!shift) return x;
     return (x << shift) | (x >> (32 - shift));
 }
+#define _rotl rotl
+#endif
```
yuriks: What I had in mind was more along the lines of `#define rotl _rotl` if available and otherwise fall back to just defining `rotl`. Isn't mingw rotl/rotr and intrinsic?

CMakeLists.txt
darkf: @neobrain It's to link the `#pragma comment(lib, "psapi")` that's in `common/memory_util.cpp` -- how would you suggest I explain that in a comment, or should I link it in CMake instead and remove that pragma?

src/common/common_funcs.h
darkf: @yuriks Yeah, `_rotl` and friends should be intrinsic if defined. Since we define `_rotl64`, etc., I figured it was alright to just macro `_rotl` to stick with the convention. Since people might be reaching for `_rotl` and might not know of the existence of `rotl` I figured that is a better name as well.

In addition, it doesn't seem like `_rotl`/`_rotr` are used in the codebase right now (at least from a `git grep _rotl`.)

CMakeLists.txt
neobrain: Something like "PSAPI = Process Status API" is sufficient.

darkf:
Addressed @neobrain's comment

darkf:
Addressed @yuriks 's comment as well. I think that's everything.

yuriks:
LGTM

ckdesu:
I successfully compiled Citra under 32bit Windows with Mingw using your patch, however for 32bit i had to edit the following line:
Line 53 of citra/src/video_core/shader/shader.cpp
```c++
void Shutdown() {
    shader_map.clear();
}
```
To
```c++
void Shutdown() {
    #ifdef ARCHITECTURE_x86_64
    shader_map.clear();
    #endif // ARCHITECTURE_x86_64
}
```

bunnei:
LGTM

PR #1036, opened 2015-08-16
Fix Linux GCC 4.9 build (complaining about undeclared memset)

LittleWhite-tb:


lioncash:
Please organize the header alphabetically

LittleWhite-tb:
Ok. Also squashed because it's just details :P

PR #1037, opened 2015-08-16
Shader: Implement EX2 and LG2 in interpreter/JIT

### LINKED ISSUE #1021: Rollercoaster Tycoon 3D: Missing Shader Opcodes 0x05 and 0x06
You should run into them just booting the game and navigating the menus to get into any of the modes.  I haven't found a way to avoid them.

According to @yuriks Fire Emblem Awakening also uses them and they're log2 and exp2

Dunno what else to list.

aroulin:
Implements the missing EX2 (exponential base 2) and LG2 (logarithm base 2) shader instructions both in the interpreter and in the JIT.

Using std::exp2 and std::log2 is probably imprecise w.r.t. the hardware. It depends on how the GPU computes floating-points operations internally and if it is still IEEE-754 compliant (despite using very likely 24-bit floating-points internally).

Note that these instructions seem to perform the operation only on the first component and assign the result to all enabled destination's components. Very simple hardware tests to verify this behavior are available here: https://github.com/aroulin/3ds-gpu-tests.

Fixes #1021

### DIFF
--- a/src/video_core/shader/shader_interpreter.cpp
+++ b/src/video_core/shader/shader_interpreter.cpp
@@ -334,6 +334,42 @@ void RunInterpreter(UnitState<Debug>& state) {
                 Record<DebugDataRecord::CMP_RESULT>(state.debug, iteration, state.conditional_code);
                 break;
 
+            case OpCode::Id::EX2:
+            {
+                Record<DebugDataRecord::SRC1>(state.debug, iteration, src1);
+                Record<DebugDataRecord::DEST_IN>(state.debug, iteration, dest);
+
+                // EX2 only takes first component exp2 and writes it to all dest components
+                float24 ex2_res = float24::FromFloat32(std::exp2(src1[0].ToFloat32()));
+                for (int i = 0; i < 4; ++i) {
+                    if (!swizzle.DestComponentEnabled(i))
+                        continue;
+
+                    dest[i] = ex2_res;
+                }
+
+                Record<DebugDataRecord::DEST_OUT>(state.debug, iteration, dest);
+                break;
+            }
+
+            case OpCode::Id::LG2:
+            {
+                Record<DebugDataRecord::SRC1>(state.debug, iteration, src1);
+                Record<DebugDataRecord::DEST_IN>(state.debug, iteration, dest);
+
+                // LG2 only takes the first component log2 and writes it to all dest components
+                float24 lg2_res = float24::FromFloat32(std::log2(src1[0].ToFloat32()));
+                for (int i = 0; i < 4; ++i) {
+                    if (!swizzle.DestComponentEnabled(i))
+                        continue;
+
+                    dest[i] = lg2_res;
+                }
+
+                Record<DebugDataRecord::DEST_OUT>(state.debug, iteration, dest);
+                break;
+            }
+
             default:
                 LOG_ERROR(HW_GPU, "Unhandled arithmetic instruction: 0x%02x (%s): 0x%08x",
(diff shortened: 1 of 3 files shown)


src/video_core/shader/shader_interpreter.cpp
```diff
@@ -334,6 +334,40 @@ void RunInterpreter(UnitState<Debug>& state) {
                 Record<DebugDataRecord::CMP_RESULT>(state.debug, iteration, state.conditional_code);
                 break;
 
+            case OpCode::Id::EX2:
+            {
+                Record<DebugDataRecord::SRC1>(state.debug, iteration, src1);
+                Record<DebugDataRecord::DEST_IN>(state.debug, iteration, dest);
+
+                float24 ex2_res = float24::FromFloat32(std::exp2(src1[0].ToFloat32()));
```
Subv: Does it really take only the exp of the first component and assign it everywhere? I thought it took a component-wise exp and assigned it to the respective component in the dest

aroulin: oh, I forgot to make a note about that in the PR summary. It seems that yes, it only take the exp of the first component. I recently tested it on hardware and updated 3dbrew. Other people confirming this behavior could be great though.

neobrain: Make sure to note it in the source, too ;)

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -523,6 +523,22 @@ void JitCompiler::Compile_MAD(Instruction instr) {
     Compile_DestEnable(instr, SRC1);
 }
 
+void JitCompiler::Compile_EX2(Instruction instr) {
+    Compile_SwizzleSrc(instr, 1, instr.common.src1, SRC1);
+    MOVSS(XMM0, R(SRC1));
+    ABI_CallFunction((const void*)(&exp2f));
```
Subv: shouldn't functions automatically decay to pointers? what's the need for this cast?

aroulin: with gcc-4.9.2, I get 

    error: invalid conversion from ‘float (*)(float) throw ()’ to ‘const void*’ [-fpermissive]

Subv: you should use a static_cast then, not a C cast

aroulin: Changed to reinterpret_cast, static_cast does not work here

Subv:
LGTM

src/video_core/shader/shader_jit_x64.cpp
yuriks: Function pointers aren't convertible to data pointers. (And certainly not implicitly.) So this explicit cast is necessary.

src/video_core/shader/shader_jit_x64.h
```diff
@@ -56,6 +56,8 @@ class JitCompiler : public Gen::XCodeBlock {
     void Compile_JMP(Instruction instr);
     void Compile_CMP(Instruction instr);
     void Compile_MAD(Instruction instr);
+    void Compile_EX2(Instruction instr);
+    void Compile_LG2(Instruction instr);
```
yuriks: Hmm, add these in their proper sorted positions? (Between DP4 and MUL.)

yuriks:
LGTM other than the nitpick

src/video_core/shader/shader_jit_x64.h
aroulin: Moved both declarations and definitions to their proper positions

bunnei:
LGTM

PR #1042, opened 2015-08-16
Rasterizer/HW: Improved the speed of the swizzling during fb readbacks.

Subv:
This reduces readback and reload time by a few milliseconds.

Original function by @yuriks

### DIFF
--- a/src/common/common_types.h
+++ b/src/common/common_types.h
@@ -50,7 +50,10 @@ typedef double  f64; ///< 64-bit floating point
 typedef u32 VAddr; ///< Represents a pointer in the userspace virtual address space.
 typedef u32 PAddr; ///< Represents a pointer in the ARM11 physical address space.
 
-// A 24-bit storage datatype to make working with data of this size (e.g. RGB8 textures) easier. Not intended for usage in arithmetic.
+/**
+ * A 24-bit storage datatype to make working with data of this size (e.g. RGB8 textures) easier.
+ * Not intended for usage in arithmetic.
+ */
 struct u24_be {
     u8 components[3];
 };
--- a/src/common/common_types.h
+++ b/src/common/common_types.h
@@ -50,6 +50,11 @@ typedef double  f64; ///< 64-bit floating point
 typedef u32 VAddr; ///< Represents a pointer in the userspace virtual address space.
 typedef u32 PAddr; ///< Represents a pointer in the ARM11 physical address space.
 
+// A 24-bit storage datatype to make working with data of this size (e.g. RGB8 textures) easier. Not intended for usage in arithmetic.
+struct u24_be {
+    u8 components[3];
+};
+
 // An inheritable class to disallow the copy constructor and operator= functions
 class NonCopyable {
 protected:
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -849,21 +849,25 @@ void RasterizerOpenGL::ReloadColorBuffer() {
     std::unique_ptr<u8[]> temp_fb_color_buffer(new u8[fb_color_texture.width * fb_color_texture.height * bytes_per_pixel]);
 
     // Directly copy pixels. The OpenGL internal formats match the 3DS ones, so no conversion is necessary.
+    using VideoCore::CopyTextureAndUntile;
     switch (bytes_per_pixel) {
     case 4:
-        VideoCore::CopyTextureAndTile<u32>((u32*)temp_fb_color_buffer.get(), (u32*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        CopyTextureAndUntile<u32>(reinterpret_cast<u32*>(temp_fb_color_buffer.get()),
+                reinterpret_cast<u32*>(color_buffer),
+                fb_color_texture.width, fb_color_texture.height);
         break;
     case 3:
-        VideoCore::CopyTextureAndTile<u24_be>((u24_be*)temp_fb_color_buffer.get(), (u24_be*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        CopyTextureAndUntile<u24_be>(reinterpret_cast<u24_be*>(temp_fb_color_buffer.get()),
+                reinterpret_cast<u24_be*>(color_buffer),
+                fb_color_texture.width, fb_color_texture.height);
         break;
     case 2:
-        VideoCore::CopyTextureAndTile<u16>((u16*)temp_fb_color_buffer.get(), (u16*)color_buffer, fb_color_texture.width, fb_color_texture.height);
-        break;
-    case 1:
-        VideoCore::CopyTextureAndTile<u8>(temp_fb_color_buffer.get(), color_buffer, fb_color_texture.width, fb_color_texture.height);
+        CopyTextureAndUntile<u16>(reinterpret_cast<u16*>(temp_fb_color_buffer.get()),
+                reinterpret_cast<u16*>(color_buffer),
+                fb_color_texture.width, fb_color_texture.height);
         break;
     default:
-        LOG_ERROR(Render_OpenGL, "Unimplemented pixel size %u bytes per pixel", bytes_per_pixel);
+        UNREACHABLE();
     }
 
     state.texture_units[0].texture_2d = fb_color_texture.texture.handle;
@@ -968,21 +972,25 @@ void RasterizerOpenGL::CommitColorBuffer() {
             state.Apply();
 
             // Directly copy pixels. The OpenGL internal formats match the 3DS ones, so no conversion is necessary.
+            using VideoCore::CopyTextureAndTile;
             switch (bytes_per_pixel) {
             case 4:
-                VideoCore::CopyTextureAndTile<u32>((u32*)color_buffer, (u32*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                CopyTextureAndTile<u32>(reinterpret_cast<u32*>(color_buffer),
+                        reinterpret_cast<u32*>(temp_gl_color_buffer.get()),
+                        fb_color_texture.width, fb_color_texture.height);
                 break;
             case 3:
-                VideoCore::CopyTextureAndTile<u24_be>((u24_be*)color_buffer, (u24_be*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                CopyTextureAndTile<u24_be>(reinterpret_cast<u24_be*>(color_buffer),
+                        reinterpret_cast<u24_be*>(temp_gl_color_buffer.get()),
+                        fb_color_texture.width, fb_color_texture.height);
                 break;
             case 2:
-                VideoCore::CopyTextureAndTile<u16>((u16*)color_buffer, (u16*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
-                break;
-            case 1:
-                VideoCore::CopyTextureAndTile<u8>(color_buffer, temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                CopyTextureAndTile<u16>(reinterpret_cast<u16*>(color_buffer),
+                        reinterpret_cast<u16*>(temp_gl_color_buffer.get()),
+                        fb_color_texture.width, fb_color_texture.height);
                 break;
             default:
-                LOG_ERROR(Render_OpenGL, "Unimplemented pixel size %u bytes per pixel", bytes_per_pixel);
+                UNREACHABLE();
             }
         }
     }
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -102,9 +102,10 @@ void RasterizerOpenGL::InitObjects() {
     glEnableVertexAttribArray(attrib_texcoords + 1);
     glEnableVertexAttribArray(attrib_texcoords + 2);
 
-    // Create textures for OGL framebuffer that will be rendered to, initially 1x1 to succeed in framebuffer creation
+    // Create textures for OGL framebuffer that will be rendered to, initially 8x8 to succeed in framebuffer creation
+    // 3DS textures' width and height must be multiples of 8
     fb_color_texture.texture.Create();
-    ReconfigureColorTexture(fb_color_texture, Pica::Regs::ColorFormat::RGBA8, 1, 1);
+    ReconfigureColorTexture(fb_color_texture, Pica::Regs::ColorFormat::RGBA8, 8, 8);
 
     state.texture_units[0].texture_2d = fb_color_texture.texture.handle;
     state.Apply();
@@ -848,15 +849,21 @@ void RasterizerOpenGL::ReloadColorBuffer() {
     std::unique_ptr<u8[]> temp_fb_color_buffer(new u8[fb_color_texture.width * fb_color_texture.height * bytes_per_pixel]);
 
     // Directly copy pixels. Internal OpenGL color formats are consistent so no conversion is necessary.
-    for (int y = 0; y < fb_color_texture.height; ++y) {
-        for (int x = 0; x < fb_color_texture.width; ++x) {
-            const u32 coarse_y = y & ~7;
-            u32 dst_offset = VideoCore::GetMortonOffset(x, y, bytes_per_pixel) + coarse_y * fb_color_texture.width * bytes_per_pixel;
-            u32 gl_pixel_index = (x + y * fb_color_texture.width) * bytes_per_pixel;
-
-            u8* pixel = color_buffer + dst_offset;
-            memcpy(&temp_fb_color_buffer[gl_pixel_index], pixel, bytes_per_pixel);
-        }
+    switch (bytes_per_pixel) {
+    case 4:
+        VideoCore::CopyTextureAndTile<u32>((u32*)temp_fb_color_buffer.get(), (u32*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    case 3:
+        VideoCore::CopyTextureAndTile<u24_be>((u24_be*)temp_fb_color_buffer.get(), (u24_be*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    case 2:
+        VideoCore::CopyTextureAndTile<u16>((u16*)temp_fb_color_buffer.get(), (u16*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    case 1:
+        VideoCore::CopyTextureAndTile<u8>(temp_fb_color_buffer.get(), color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    default:
+        LOG_ERROR(Render_OpenGL, "Unimplemented pixel size %u bytes per pixel", bytes_per_pixel);
     }
 
     state.texture_units[0].texture_2d = fb_color_texture.texture.handle;
@@ -961,15 +968,21 @@ void RasterizerOpenGL::CommitColorBuffer() {
             state.Apply();
 
             // Directly copy pixels. Internal OpenGL color formats are consistent so no conversion is necessary.
-            for (int y = 0; y < fb_color_texture.height; ++y) {
-                for (int x = 0; x < fb_color_texture.width; ++x) {
-                    const u32 coarse_y = y & ~7;
-                    u32 dst_offset = VideoCore::GetMortonOffset(x, y, bytes_per_pixel) + coarse_y * fb_color_texture.width * bytes_per_pixel;
-                    u32 gl_pixel_index = x * bytes_per_pixel + y * fb_color_texture.width * bytes_per_pixel;
-
-                    u8* pixel = color_buffer + dst_offset;
-                    memcpy(pixel, &temp_gl_color_buffer[gl_pixel_index], bytes_per_pixel);
-                }
+            switch (bytes_per_pixel) {
+            case 4:
+                VideoCore::CopyTextureAndTile<u32>((u32*)color_buffer, (u32*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            case 3:
+                VideoCore::CopyTextureAndTile<u24_be>((u24_be*)color_buffer, (u24_be*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            case 2:
+                VideoCore::CopyTextureAndTile<u16>((u16*)color_buffer, (u16*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            case 1:
+                VideoCore::CopyTextureAndTile<u8>(color_buffer, temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            default:
+                LOG_ERROR(Render_OpenGL, "Unimplemented pixel size %u bytes per pixel", bytes_per_pixel);
             }
         }
     }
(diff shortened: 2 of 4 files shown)


src/video_core/utils.h
```diff
@@ -35,17 +35,51 @@ struct TGAHeader {
...
+ * @param dst Pointer to which the texture will be copied
+ * @param src Pointer to the source texture data
+ * @param width Width of the texture
+ * @param height Height of the texture
+ * @param T Type of the source and destination pointers, the swizzling process depends on the size of this parameter.
+ */
+template<typename T>
+static inline void CopyTextureAndTile(T* dst, const T* src, unsigned int width, unsigned int height) {
+    for (unsigned int y = 0; y < height; y += 8) {
+        for (unsigned int x = 0; x < width; x += 8) {
```
yuriks: You should modify the boundary conditions here so that it rounds the size *down* instead of up, making this at least safe to use for non-mod8 dimensions.

src/common/common_types.h
```diff
@@ -50,6 +50,11 @@ typedef double  f64; ///< 64-bit floating point
 typedef u32 VAddr; ///< Represents a pointer in the userspace virtual address space.
 typedef u32 PAddr; ///< Represents a pointer in the ARM11 physical address space.
 
+// A 24-bit storage datatype to make working with data of this size (e.g. RGB8 textures) easier. Not intended for usage in arithmetic.
+struct u24 {
```
neobrain: This might be mixed up for a host-endianess u24 type. Maybe call it u24_be instead, reflecting the endianness which it would technically have if arithmetic operands were defined on it?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -788,15 +789,21 @@ void RasterizerOpenGL::ReloadColorBuffer() {
...
-        for (int x = 0; x < fb_color_texture.width; ++x) {
-            const u32 coarse_y = y & ~7;
-            u32 dst_offset = VideoCore::GetMortonOffset(x, y, bytes_per_pixel) + coarse_y * fb_color_texture.width * bytes_per_pixel;
-            u32 gl_pixel_index = (x + y * fb_color_texture.width) * bytes_per_pixel;
-
-            u8* pixel = color_buffer + dst_offset;
-            memcpy(&temp_fb_color_buffer[gl_pixel_index], pixel, bytes_per_pixel);
-        }
+    switch (bytes_per_pixel) {
+        case 4:
```
neobrain: switch cases don't increase the indentation level.

src/video_core/utils.h
```diff
@@ -35,17 +35,51 @@ struct TGAHeader {
  */
 void DumpTGA(std::string filename, short width, short height, u8* raw_data);
 
+static const u8 morton_lut[64] = {
```
neobrain: Since this is part of the public interface now, it should be documented, or put into a `namespace detail`.

src/video_core/utils.h
```diff
@@ -35,17 +35,51 @@ struct TGAHeader {
...
+ * applying a Morton-order transformation while copying.
+ * @param dst Pointer to which the texture will be copied
+ * @param src Pointer to the source texture data
+ * @param width Width of the texture
+ * @param height Height of the texture
+ * @param T Type of the source and destination pointers, the swizzling process depends on the size of this parameter.
+ */
+template<typename T>
+static inline void CopyTextureAndTile(T* dst, const T* src, unsigned int width, unsigned int height) {
+    for (unsigned int y = 0; y < height; y += 8) {
```
neobrain: Sounds like this requires height to be a multiple of 8, which should be documented. Same for width.

src/video_core/utils.h
```diff
@@ -35,17 +35,51 @@ struct TGAHeader {
...
+ * @param dst Pointer to which the texture will be copied
+ * @param src Pointer to the source texture data
+ * @param width Width of the texture
+ * @param height Height of the texture
+ * @param T Type of the source and destination pointers, the swizzling process depends on the size of this parameter.
+ */
+template<typename T>
+static inline void CopyTextureAndTile(T* dst, const T* src, unsigned int width, unsigned int height) {
+    for (unsigned int y = 0; y + 8 <= height; y += 8) {
+        for (unsigned int x = 0; x + 8 <= width; x += 8) {
```
neobrain: ... don't force-push without keeping a PR, history, please, since you just dropped my comment on this.

Anyway, check your mail inbox for what I said before.

Subv:
I always check the hidden comments.

Updated

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -788,15 +789,21 @@ void RasterizerOpenGL::ReloadColorBuffer() {
     std::unique_ptr<u8[]> temp_fb_color_buffer(new u8[fb_color_texture.width * fb_color_texture.height * bytes_per_pixel]);
 
     // Directly copy pixels. Internal OpenGL color formats are consistent so no conversion is necessary.
```
yuriks: While you're at it, can you clarify this comment? "Directly copy pixels. The OpenGL internal formats match the 3DS ones, so no conversion is necessary."

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -788,15 +789,21 @@ void RasterizerOpenGL::ReloadColorBuffer() {
...
+    case 4:
+        VideoCore::CopyTextureAndTile<u32>((u32*)temp_fb_color_buffer.get(), (u32*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    case 3:
+        VideoCore::CopyTextureAndTile<u24_be>((u24_be*)temp_fb_color_buffer.get(), (u24_be*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    case 2:
+        VideoCore::CopyTextureAndTile<u16>((u16*)temp_fb_color_buffer.get(), (u16*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    case 1:
```
yuriks: Is this case even possible? There are no 8bpp framebuffer formats.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -890,15 +897,21 @@ void RasterizerOpenGL::CommitColorBuffer() {
...
+            case 4:
+                VideoCore::CopyTextureAndTile<u32>((u32*)color_buffer, (u32*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            case 3:
+                VideoCore::CopyTextureAndTile<u24_be>((u24_be*)color_buffer, (u24_be*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            case 2:
+                VideoCore::CopyTextureAndTile<u16>((u16*)color_buffer, (u16*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            case 1:
```
yuriks: Likewise.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -788,15 +789,21 @@ void RasterizerOpenGL::ReloadColorBuffer() {
...
+        VideoCore::CopyTextureAndTile<u24_be>((u24_be*)temp_fb_color_buffer.get(), (u24_be*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    case 2:
+        VideoCore::CopyTextureAndTile<u16>((u16*)temp_fb_color_buffer.get(), (u16*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    case 1:
+        VideoCore::CopyTextureAndTile<u8>(temp_fb_color_buffer.get(), color_buffer, fb_color_texture.width, fb_color_texture.height);
+        break;
+    default:
+        LOG_ERROR(Render_OpenGL, "Unimplemented pixel size %u bytes per pixel", bytes_per_pixel);
```
yuriks: This should never happen, so assert UNREACHABLE.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -890,15 +897,21 @@ void RasterizerOpenGL::CommitColorBuffer() {
...
+                VideoCore::CopyTextureAndTile<u24_be>((u24_be*)color_buffer, (u24_be*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            case 2:
+                VideoCore::CopyTextureAndTile<u16>((u16*)color_buffer, (u16*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            case 1:
+                VideoCore::CopyTextureAndTile<u8>(color_buffer, temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                break;
+            default:
+                LOG_ERROR(Render_OpenGL, "Unimplemented pixel size %u bytes per pixel", bytes_per_pixel);
```
yuriks: Likewise

src/video_core/utils.h
```diff
@@ -35,17 +35,52 @@ struct TGAHeader {
  */
 void DumpTGA(std::string filename, short width, short height, u8* raw_data);
 
+/// Lookup table for the offsets used to convert an image to Morton order.
+static const u8 morton_lut[64] = {
+    0,  1,  4,  5, 16, 17, 20, 21,
+    2,  3,  6,  7, 18, 19, 22, 23,
+    8,  9, 12, 13, 24, 25, 28, 29,
```
yuriks: Align the columns nicely? :<

src/common/common_types.h
```diff
@@ -50,6 +50,11 @@ typedef double  f64; ///< 64-bit floating point
 typedef u32 VAddr; ///< Represents a pointer in the userspace virtual address space.
 typedef u32 PAddr; ///< Represents a pointer in the ARM11 physical address space.
 
+// A 24-bit storage datatype to make working with data of this size (e.g. RGB8 textures) easier. Not intended for usage in arithmetic.
```
yuriks: Long comment line. Also make it a doc-comment.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -95,9 +95,10 @@ void RasterizerOpenGL::InitObjects() {
     glEnableVertexAttribArray(attrib_texcoords + 1);
     glEnableVertexAttribArray(attrib_texcoords + 2);
 
-    // Create textures for OGL framebuffer that will be rendered to, initially 1x1 to succeed in framebuffer creation
+    // Create textures for OGL framebuffer that will be rendered to, initially 8x8 to succeed in framebuffer creation
+    // 3DS textures' width and height must be multiples of 8
     fb_color_texture.texture.Create();
-    ReconfigureColorTexture(fb_color_texture, Pica::Regs::ColorFormat::RGBA8, 1, 1);
+    ReconfigureColorTexture(fb_color_texture, Pica::Regs::ColorFormat::RGBA8, 8, 8);
```
yuriks: This isn't necessary anymore after fixing the for loop, right?

src/video_core/utils.h
```diff
@@ -35,17 +35,52 @@ struct TGAHeader {
...
-    i = (i ^ (i << 2)) & 0x1313;      // ---2 --10
-    i = (i ^ (i << 1)) & 0x1515;      // ---2 -1-0
-    i = (i | (i >> 7)) & 0x3F;
-    return i;
+    return morton_lut[y * 8 + x];
+}
+
+/**
+ * Copies the texture data from the source address to the destination address,
+ * applying a Morton-order transformation while copying.
```
yuriks: Weird word-wrapping is inconsistent with other lines in the same comment which are way longer.

yuriks:
This broken the Bravely Default title-screen:

broken-swizzle

yuriks:
Ah, the problem is that you're using the same function for both `CommitColorBuffer` and `ReloadColorBuffer`. The swizzle unfortunately isn't reflective, so for Commit you need to apply the morton LUT on the reads instead.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -889,16 +896,22 @@ void RasterizerOpenGL::CommitColorBuffer() {
...
-                    u32 dst_offset = VideoCore::GetMortonOffset(x, y, bytes_per_pixel) + coarse_y * fb_color_texture.width * bytes_per_pixel;
-                    u32 gl_pixel_index = x * bytes_per_pixel + y * fb_color_texture.width * bytes_per_pixel;
-
-                    u8* pixel = color_buffer + dst_offset;
-                    memcpy(pixel, &temp_gl_color_buffer[gl_pixel_index], bytes_per_pixel);
-                }
+            // Directly copy pixels. The OpenGL internal formats match the 3DS ones, so no conversion is necessary.
+            switch (bytes_per_pixel) {
+            case 4:
+                VideoCore::CopyTextureAndTile<u32>((u32*)color_buffer, (u32*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
```
yuriks: Use C++ casts

yuriks: I recommend doing `using VideoCore::CopyTextureAndTile` before the switch to remove a bit of the repetition.

yuriks:
Proposed patch:

```patch
diff --git a/src/video_core/renderer_opengl/gl_rasterizer.cpp b/src/video_core/renderer_opengl/gl_rasterizer.cpp
index c6b264d..a21535f 100644
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -789,21 +789,25 @@ void RasterizerOpenGL::ReloadColorBuffer() {
     std::unique_ptr<u8[]> temp_fb_color_buffer(new u8[fb_color_texture.width * fb_color_texture.height * bytes_per_pixel]);
 
     // Directly copy pixels. The OpenGL internal formats match the 3DS ones, so no conversion is necessary.
+    using VideoCore::CopyTextureAndUntile;
     switch (bytes_per_pixel) {
     case 4:
-        VideoCore::CopyTextureAndTile<u32>((u32*)temp_fb_color_buffer.get(), (u32*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        CopyTextureAndUntile<u32>(reinterpret_cast<u32*>(temp_fb_color_buffer.get()),
+                reinterpret_cast<u32*>(color_buffer),
+                fb_color_texture.width, fb_color_texture.height);
         break;
     case 3:
-        VideoCore::CopyTextureAndTile<u24_be>((u24_be*)temp_fb_color_buffer.get(), (u24_be*)color_buffer, fb_color_texture.width, fb_color_texture.height);
+        CopyTextureAndUntile<u24_be>(reinterpret_cast<u24_be*>(temp_fb_color_buffer.get()),
+                reinterpret_cast<u24_be*>(color_buffer),
+                fb_color_texture.width, fb_color_texture.height);
         break;
     case 2:
-        VideoCore::CopyTextureAndTile<u16>((u16*)temp_fb_color_buffer.get(), (u16*)color_buffer, fb_color_texture.width, fb_color_texture.height);
-        break;
-    case 1:
-        VideoCore::CopyTextureAndTile<u8>(temp_fb_color_buffer.get(), color_buffer, fb_color_texture.width, fb_color_texture.height);
+        CopyTextureAndUntile<u16>(reinterpret_cast<u16*>(temp_fb_color_buffer.get()),
+                reinterpret_cast<u16*>(color_buffer),
+                fb_color_texture.width, fb_color_texture.height);
         break;
     default:
-        LOG_ERROR(Render_OpenGL, "Unimplemented pixel size %u bytes per pixel", bytes_per_pixel);
+        UNREACHABLE();
     }
 
     state.texture_units[0].texture_2d = fb_color_texture.texture.handle;
@@ -897,21 +901,25 @@ void RasterizerOpenGL::CommitColorBuffer() {
             state.Apply();
 
             // Directly copy pixels. The OpenGL internal formats match the 3DS ones, so no conversion is necessary.
+            using VideoCore::CopyTextureAndTile;
             switch (bytes_per_pixel) {
             case 4:
-                VideoCore::CopyTextureAndTile<u32>((u32*)color_buffer, (u32*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                CopyTextureAndTile<u32>(reinterpret_cast<u32*>(color_buffer),
+                        reinterpret_cast<u32*>(temp_gl_color_buffer.get()),
+                        fb_color_texture.width, fb_color_texture.height);
                 break;
             case 3:
-                VideoCore::CopyTextureAndTile<u24_be>((u24_be*)color_buffer, (u24_be*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                CopyTextureAndTile<u24_be>(reinterpret_cast<u24_be*>(color_buffer),
+                        reinterpret_cast<u24_be*>(temp_gl_color_buffer.get()),
+                        fb_color_texture.width, fb_color_texture.height);
                 break;
             case 2:
-                VideoCore::CopyTextureAndTile<u16>((u16*)color_buffer, (u16*)temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
-                break;
-            case 1:
-                VideoCore::CopyTextureAndTile<u8>(color_buffer, temp_gl_color_buffer.get(), fb_color_texture.width, fb_color_texture.height);
+                CopyTextureAndTile<u16>(reinterpret_cast<u16*>(color_buffer),
+                        reinterpret_cast<u16*>(temp_gl_color_buffer.get()),
+                        fb_color_texture.width, fb_color_texture.height);
                 break;
             default:
-                LOG_ERROR(Render_OpenGL, "Unimplemented pixel size %u bytes per pixel", bytes_per_pixel);
+                UNREACHABLE();
             }
         }
     }
diff --git a/src/video_core/utils.h b/src/video_core/utils.h
index ba61b2a..887601d 100644
--- a/src/video_core/utils.h
+++ b/src/video_core/utils.h
@@ -37,9 +37,9 @@ void DumpTGA(std::string filename, short width, short height, u8* raw_data);
 
 /// Lookup table for the offsets used to convert an image to Morton order.
 static const u8 morton_lut[64] = {
-    0,  1,  4,  5, 16, 17, 20, 21,
-    2,  3,  6,  7, 18, 19, 22, 23,
-    8,  9, 12, 13, 24, 25, 28, 29,
+     0,  1,  4,  5, 16, 17, 20, 21,
+     2,  3,  6,  7, 18, 19, 22, 23,
+     8,  9, 12, 13, 24, 25, 28, 29,
     10, 11, 14, 15, 26, 27, 30, 31,
     32, 33, 36, 37, 48, 49, 52, 53,
     34, 35, 38, 39, 50, 51, 54, 55,
@@ -57,13 +57,15 @@ static inline u32 MortonInterleave(u32 x, u32 y) {
 }
 
 /**
- * Copies the texture data from the source address to the destination address,
- * applying a Morton-order transformation while copying.
+ * Copies the texture data from the source address to the destination address, applying a
+ * Morton-order transformation while copying.
+ *
+ * @param T Type of the source and destination pointers, the swizzling process depends on the size
+ *        of this parameter.
  * @param dst Pointer to which the texture will be copied.
  * @param src Pointer to the source texture data.
  * @param width Width of the texture, should be a multiple of 8.
  * @param height Height of the texture, should be a multiple of 8.
- * @param T Type of the source and destination pointers, the swizzling process depends on the size of this parameter.
  */
 template<typename T>
 static inline void CopyTextureAndTile(T* dst, const T* src, unsigned int width, unsigned int height) {
@@ -84,6 +86,34 @@ static inline void CopyTextureAndTile(T* dst, const T* src, unsigned int width,
 }
 
 /**
+ * Copies texture data while undoing the transformation applied by `CopyTextureAndTile`.
+ *
+ * @param T Type of the source and destination pointers, the swizzling process depends on the size
+ *        of this parameter.
+ * @param dst Pointer to which the texture will be copied.
+ * @param src Pointer to the source texture data.
+ * @param width Width of the texture, should be a multiple of 8.
+ * @param height Height of the texture, should be a multiple of 8.
+ */
+template<typename T>
+static inline void CopyTextureAndUntile(T* dst, const T* src, unsigned int width, unsigned int height) {
+    for (unsigned int y = 0; y + 8 <= height; y += 8) {
+        for (unsigned int x = 0; x + 8 <= width; x += 8) {
+            T* line = &dst[y * width + x];
+
+            for (unsigned int yy = 0; yy < 8; ++yy) {
+                for (unsigned int xx = 0; xx < 8; ++xx) {
+                    line[xx] = src[morton_lut[yy * 8 + xx]];
+                }
+                line += width;
+            }
+
+            src += 8 * 8;
+        }
+    }
+}
+
+/**
  * Calculates the offset of the position of the pixel in Morton order
  */
 static inline u32 GetMortonOffset(u32 x, u32 y, u32 bytes_per_pixel) {
```

yuriks:
Also, please reverse the order of the src/dst arguments so they follow the order used by the C++ standard library.

src/video_core/utils.h
```diff
@@ -35,17 +35,82 @@ struct TGAHeader {
...
+ *
+ * @param T Type of the source and destination pointers, the swizzling process depends on the size
+ *        of this parameter.
+ * @param dst Pointer to which the texture will be copied.
+ * @param src Pointer to the source texture data.
+ * @param width Width of the texture, should be a multiple of 8.
+ * @param height Height of the texture, should be a multiple of 8.
+ */
+template<typename T>
+static inline void CopyTextureAndTile(T* dst, const T* src, unsigned int width, unsigned int height) {
```
yuriks: I think these two functions shouldn't go in the header. Rather, just have their declaration here, then the definition in a .cpp file with explicit instantiations for u32/u24/u16.

Subv:
Updated

Subv:
Updated again

src/video_core/utils.cpp
```diff
@@ -33,4 +33,48 @@ void DumpTGA(std::string filename, short width, short height, u8* raw_data) {
...
+            }
+
+            src += 8 * 8;
+        }
+    }
+}
+
+template void CopyTextureAndUntile<u16>(const u16* src, u16* dst, unsigned int width, unsigned int height);
+template void CopyTextureAndUntile<u24_be>(const u24_be* src, u24_be* dst, unsigned int width, unsigned int height);
+template void CopyTextureAndUntile<u32>(const u32* src, u32* dst, unsigned int width, unsigned int height);
```
yuriks: Missing newline after this

Subv:
Updated, should be ready for a final review

yuriks:
LGTM!

archshift:
Instead of adding that ugly `u24_be type (which is pretty misleading too IMO), and the ugly `reinterpret_cast`ing, why not add a numerical template argument for the bytes per pixel to the `CopyTextureAndTile`/`CopyTextureAndUntile` functions? Then the compiler would be able to properly optimize a `memcpy`.

purpasmart96:
Personally, I'm against the unnecessary usage of templates and the usage of reinterpret_cast just so you can get a few extra milliseconds in speed.

neobrain:
CopyTextureAndUntile should take u8* pointers instead of T* ones. There's no way to implement that interface without violating strict aliasing rules otherwise (in fact, prefer strict-aliasing conforming implementations over non-conforming ones where possible in general).

Subv:
Just as a reminder, I still plan on finishing this, I've just been busy lately

src/common/common_types.h
```diff
@@ -50,6 +50,14 @@ typedef double  f64; ///< 64-bit floating point
 typedef u32 VAddr; ///< Represents a pointer in the userspace virtual address space.
 typedef u32 PAddr; ///< Represents a pointer in the ARM11 physical address space.
 
+/**
+ * A 24-bit storage datatype to make working with data of this size (e.g. RGB8 textures) easier.
+ * Not intended for usage in arithmetic.
+ */
+struct u24_be {
+    u8 components[3];
+};
```
yuriks: You forgot to take out the trash.

bunnei:
<pre>Subv: at this point that's a stale PR, i think someone was tinkering with compute shaders instead</pre>

PR #1045, opened 2015-08-17
Improvements for MRU

LittleWhite-tb:
* avoid duplicates
* always put the last file loaded to top of the list

I tried to add tooltips in order to show the fullpath when the mouse is hovering the item, but in fact, tooltips for QAction in QMenu are not shown at all.

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -287,6 +287,17 @@ void GMainWindow::ShutdownGame() {
     render_window->hide();
 }
 
+void GMainWindow::StoreRecentFile(const QString& filename)
+{
+    QSettings settings;
+    QStringList recent_files = settings.value("recentFiles").toStringList();
+    recent_files.prepend(filename);
+    recent_files.removeDuplicates();
+    settings.setValue("recentFiles", recent_files);
+
+    UpdateRecentFiles();
+}
+
 void GMainWindow::UpdateRecentFiles() {
     QSettings settings;
     QStringList recent_files = settings.value("recentFiles").toStringList();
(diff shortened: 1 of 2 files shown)


yuriks:
LGTM

src/citra_qt/main.cpp
```diff
@@ -357,13 +366,8 @@ void GMainWindow::OnMenuRecentFile() {
...
         recent_files.removeOne(filename);
         settings.setValue("recentFiles", recent_files);
-
-        action->setVisible(false);
-        // Grey out the recent files menu if the list is empty
-        if (ui.menu_recent_files->isEmpty()) {
-            ui.menu_recent_files->setEnabled(false);
-        }
     }
+    UpdateRecentFiles();
```
yuriks: Just noticed this, I think you can just fold this function call into `StoreRecentFile`, since they're always called together.

LittleWhite-tb:
You are right. Change done.

src/citra_qt/main.h
```diff
@@ -60,6 +60,7 @@ class GMainWindow : public QMainWindow
     void BootGame(const std::string& filename);
     void ShutdownGame();
 
+    void StoreRecentFile(const QString& filename);
```
neobrain: docstrings, please.

neobrain:
LGTM other than docstrings.

LittleWhite-tb:
Added docstrings. Maybe I made these too longs and you were expecting just one line ?

neobrain:
One line would've been good enough, but that's okay too. Do note that we usually use a different order for the docstring contents, but that doesn't matter too much, I guess.

LGTM.

LittleWhite-tb:
Removed brief annotations since auto brief is enabled on the project.

yuriks:
I also think the docstrings are a bit on the verbose side. (If you want a step-by-step description of what the function does you could just look in the source. :) )

But good enough.

PR #1047, opened 2015-08-18
Shader: Save caller-saved registers in JIT before a CALL

aroulin:
Caller-saved registers were not saved when calling functions from the shader JIT. We must also save some XMM registers depending on their uses in the JIT and the ABI.


yuriks:
LGTM

bunnei:
I'm merging this as a fix... But I'd like to see these instructions properly implemented in the JIT.

Issue #1041, opened 2015-08-16
Paper Mario Sticker Star - Graphical Glitches


JohnGodgames:
![paper 3](https://cloud.githubusercontent.com/assets/11453507/9294785/f64528dc-4457-11e5-9c48-c1f9e7a6c78f.PNG)

![paper 1](https://cloud.githubusercontent.com/assets/11453507/9294766/66b77418-4457-11e5-9849-d7ad40589831.PNG)

![paper 2](https://cloud.githubusercontent.com/assets/11453507/9294768/6f514e3c-4457-11e5-9390-a164c9bf25f7.PNG)

![paper 4](https://cloud.githubusercontent.com/assets/11453507/9294789/2834c78a-4458-11e5-89fc-fb6bd13ff873.PNG)

CITRACE LOG: https://goo.gl/Tff5vU

Subv:
Could you try with #996 please?

JohnGodgames:
#996 not helping...the same :(

yuriks:
What about the software renderer?

JohnGodgames:
screens are taken at software renderer...

JohnGodgames:
#1049 not helped

wwylele:
Fixed by #1462 , should be closed

PR #1049, opened 2015-08-19
Rasterizer: Corrected the stencil implementation.

### LINKED ISSUE #1039: Super Mario 3D Land DEMO - Graphics glitches / missing textures
Spamming error:
HW.GPU <Error> video_core\debug_utils\debug_utils.cpp:Pica::DebugUtils::LookupTexture:592: Unknown texture format: 6

sm3dl 1

sm3dl 2

CITRACE LOG: https://goo.gl/5auSyA

Subv:
And implemented stencil in the hw renderer.
Implemented more stencil functions.

This was verified using hw tests, some of which can be found [here](https://gist.github.com/Subv/692aaf09d4593018bbcc).

I'll update the information on 3dbrew later.

Fixes #1039 and _maybe_ some of the problems with #1041

hwtests: https://github.com/Subv/3ds-homebrew/tree/master/gpu_stencil_tests

There's currently a bug related to changing the stencil buffer with the OpenGL renderer, it seems to only happen on Intel and is apparently a driver bug, a report has been filed [here](https://communities.intel.com/message/324464)

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -441,12 +441,14 @@ struct Regs {
     };
 
     enum class StencilAction : u32 {
-        Keep      = 0,
-        Zero      = 1,
-        Replace   = 2,
-        Increment = 3,
-        Decrement = 4,
-        Invert    = 5
+        Keep           = 0,
+        Zero           = 1,
+        Replace        = 2,
+        Increment      = 3,
+        Decrement      = 4,
+        Invert         = 5,
+        IncrementWrap  = 6,
+        DecrementWrap  = 7
     };
 
     struct {
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -441,8 +441,12 @@ struct Regs {
     };
 
     enum class StencilAction : u32 {
-        Keep = 0,
-        Xor  = 5,
+        Keep      = 0,
+
+        Replace   = 2,
+        Increment = 3,
+        Decrement = 4,
+        Invert    = 5
     };
 
     struct {
@@ -481,23 +485,29 @@ struct Regs {
 
         struct {
             union {
+                // Raw value of this register
+                u32 raw_func;
+
                 // If true, enable stencil testing
                 BitField< 0, 1, u32> enable;
 
                 // Comparison operation for stencil testing
                 BitField< 4, 3, CompareFunc> func;
 
-                // Value to calculate the new stencil value from
-                BitField< 8, 8, u32> replacement_value;
+                // Mask used to control writing to the stencil buffer
+                BitField< 8, 8, u32> write_mask;
 
                 // Value to compare against for stencil testing
                 BitField<16, 8, u32> reference_value;
 
                 // Mask to apply on stencil test inputs
-                BitField<24, 8, u32> mask;
+                BitField<24, 8, u32> input_mask;
             };
 
             union {
+                // Raw value of this register
+                u32 raw_op;
+
                 // Action to perform when the stencil test fails
                 BitField< 0, 3, StencilAction> action_stencil_fail;
 
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -442,7 +442,7 @@ struct Regs {
 
     enum class StencilAction : u32 {
         Keep      = 0,
-
+        Zero      = 1,
         Replace   = 2,
         Increment = 3,
         Decrement = 4,
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -237,6 +237,12 @@ static u8 PerformStencilAction(Regs::StencilAction action, u8 old_stencil, u8 re
     case Regs::StencilAction::Invert:
         return ~old_stencil;
 
+    case Regs::StencilAction::IncrementWrap:
+        return old_stencil + 1;
+
+    case Regs::StencilAction::DecrementWrap:
+        return old_stencil - 1;
+
     default:
         LOG_CRITICAL(HW_GPU, "Unknown stencil action %x", (int)action);
         UNIMPLEMENTED();
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -791,6 +791,12 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             }
 
             u8 old_stencil = 0;
+
+            auto UpdateStencil = [stencil_test, x, y, &old_stencil](Pica::Regs::StencilAction action) {
+                u8 new_stencil = PerformStencilAction(action, old_stencil, stencil_test.reference_value);
+                SetStencil(x >> 4, y >> 4, (new_stencil & stencil_test.write_mask) | (old_stencil & ~stencil_test.write_mask));
+            };
+
             if (stencil_action_enable) {
                 old_stencil = GetStencil(x >> 4, y >> 4);
                 u8 dest = old_stencil & stencil_test.input_mask;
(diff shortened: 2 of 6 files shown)


neobrain:
Please change the hwtest to a readily compilable project.

neobrain:
Please divide the hw renderer implementation and the fix into separate commits. There's no reason to have them in a single patch.

src/video_core/pica.h
```diff
@@ -441,8 +441,12 @@ struct Regs {
     };
 
     enum class StencilAction : u32 {
-        Keep = 0,
-        Xor  = 5,
+        Keep      = 0,
+
+        Replace   = 2,
+        Incr      = 3,
```
neobrain: ..ement

src/video_core/pica.h
```diff
@@ -506,7 +510,7 @@ struct Regs {
 
                 // Action to perform when both stencil and depth testing pass
                 BitField< 8, 3, StencilAction> action_depth_pass;
-            };
+            } op;
```
neobrain: these names aren't really needed.

src/video_core/pica.h
```diff
@@ -487,15 +491,15 @@ struct Regs {
...
+                // Mask used to control writing to the stencil buffer
+                BitField< 8, 8, u32> write_mask;
 
                 // Value to compare against for stencil testing
                 BitField<16, 8, u32> reference_value;
 
                 // Mask to apply on stencil test inputs
                 BitField<24, 8, u32> mask;
-            };
+            } func;
```
neobrain: name not needed

src/video_core/pica.h
```diff
@@ -487,15 +491,15 @@ struct Regs {
...
-                // Value to calculate the new stencil value from
-                BitField< 8, 8, u32> replacement_value;
+                // Mask used to control writing to the stencil buffer
+                BitField< 8, 8, u32> write_mask;
 
                 // Value to compare against for stencil testing
                 BitField<16, 8, u32> reference_value;
 
                 // Mask to apply on stencil test inputs
                 BitField<24, 8, u32> mask;
```
neobrain: Rename this to "input_mask"?

src/video_core/pica.h
Subv: But they are, unless we start performing arithmetic on the NotifyPicaRegisterChanged switch (which I'm not sure is even allowed).

Subv: ... Can't edit on mobile. 

We need to listen for writes in the two words that compose this register, the most sensible way to do that was to name the inner structures and switch on their pica reg index

neobrain: I see.

Would adding a "u32 raw" member have worked too, though? If so, I think that would be preferable because it allows for cleaner usage of these fields in all other places. Having to look up whether `action_*` is stored in `op` or `func` all the time can be pretty annoying.

Subv: That doesn't solve the issue, there are two registers here.

neobrain: You can add the "raw" field for both op and func though.

Subv:
Addressed all comments, i'll upload the tests to my repo later.

src/video_core/rasterizer.cpp
```diff
@@ -215,14 +215,25 @@ static void SetStencil(int x, int y, u8 value) {
     }
 }
 
-// TODO: Should the stencil mask be applied to the "dest" or "ref" operands? Most likely not!
-static u8 PerformStencilAction(Regs::StencilAction action, u8 dest, u8 ref) {
+// TODO: Should the stencil mask be applied to the "old_stencil" or "ref" operands? Most likely not!
```
yuriks: Does this comment still apply?

Subv: I actually did not test that in the hwtests, I assumed it was correct.

yuriks: Is it easy to test? Just might as well get rid of the todo while at it.

yuriks:
LGTM

src/video_core/rasterizer.cpp
```diff
@@ -780,10 +791,16 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             }
 
             u8 old_stencil = 0;
+
+            auto UpdateStencil = [&stencil_test, &x, &y, &old_stencil](Pica::Regs::StencilAction action) {
```
yuriks: You can capture `stencil_test`, `x` and `y` by value.

src/video_core/rasterizer.cpp
Subv: I suppose, I'll get rid of another Todo while at it.

bunnei: This PR has my blessing after you get rid of that TODO

Subv:
Removed the two stencil-related TODOs in the code. I'll upload the hwtests after i clean them up.

Subv:
Uploaded the hwtests here: https://github.com/Subv/3ds-homebrew/tree/master/gpu_stencil_tests
I'm still working on this PR, please don't merge it yet.

Subv:
There's currently a bug with the hardware renderer.
It's apparently not clearing the stencil buffer when the guest application zeroes it out. The sw renderer doesn't show this problem. See the "stencil write mask" test for a test case.

Subv:
After some research, the bug happens because the glTexSubImage2D call in ReloadDepthBuffer isn't changing the stencil buffer (In this particular case it should be writing 0 to it), only the depth buffer.
Changing glTexSubImage2D to glTexImage2D fixes the issue. 
nVidia drivers don't show this problem however they crash with heap corruption when trying to load a new game.

Subv:
A bug report has been made to Intel (https://communities.intel.com/message/324464) so i guess this is good for another review pass.

yuriks:
@Subv: In the meantime, this issue can be worked around by checking if the texture type is GL_DEPTH_STENCIL and if so using `glTexImage2D` instead of `glTexSubImage2D`.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -860,10 +869,15 @@ void RasterizerOpenGL::ReloadDepthBuffer() {
...
-    state.texture_units[0].texture_2d = 0;
+    if (fb_depth_texture.format == Pica::Regs::DepthFormat::D24S8) {
+        // TODO(Subv): There is a bug with Intel Windows drivers that makes glTexSubImage2D not change the stencil buffer.
+        // The bug has been reported to Intel (https://communities.intel.com/message/324464)
+        glTexImage2D(GL_TEXTURE_2D, 0, GL_DEPTH24_STENCIL8, fb_depth_texture.width, fb_depth_texture.height, 0,
+            GL_DEPTH_STENCIL, GL_UNSIGNED_INT_24_8, temp_fb_depth_buffer.get());
+    } else {
+        glTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, fb_depth_texture.width, fb_depth_texture.height,
+            fb_depth_texture.gl_format, fb_depth_texture.gl_type, temp_fb_depth_buffer.get());
+    }    state.texture_units[0].texture_2d = 0;
```
yuriks: You ate a newline here

Subv:
Updated

yuriks:
LGTM

src/video_core/rasterizer.cpp
```diff
@@ -780,10 +799,16 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             }
 
             u8 old_stencil = 0;
+
+            auto UpdateStencil = [stencil_test, x, y, &old_stencil](Pica::Regs::StencilAction action) {
```
neobrain: There is no apparent difference between "UpdateStencil" and "SetStencil". Suggestion: "ComputeAndSetNewStencil"

Subv: I disagree, SetStencil does a bunch of other stuff related to the format and swizzling, id prefer if this code remains separate

neobrain: Hm? I'm just suggestion a rename of this lambda.

yuriks: "UpdateStencil" sounds like a fine enough name to me, fwiw. (Maybe something like "MergeStencil" if you think "Update" isn't explicit enough? But for me there's a clear distinction between what "set" (set value, disregard what was there) and "update" (refresh or merge information) mean.)

neobrain: MergeStencil sounds like a good alternative.

bunnei:
LGTM. I'm OK with the aforementioned change of "UpdateStencil" -> "MergeStencil".

PR #1048, opened 2015-08-19
Integrate the MicroProfile profiling library

yuriks:
This brings goodies such as a configurable user interface and multi-threaded timeline view.

Screenshot:

Detailed View

If this is well received then I plan to remove the old profiler and replace it with a dedicated frametime and FPS display on the status bar on a later PR.

(**Note:** Initially the profiler starts out not displaying anything. To see stuff, enable the groups via the `Groups` menu and then some timers in the `Timer` menu.)

PR #1052, opened 2015-08-19
Shader Debugger Improvements

yuriks:
A grab bag of improvements to the shader debugger. See individual commits.

cc @neobrain

Branch history:
 - yuriks/citra@8275ad6cef470c1ab91536de66506c9086770a2e

### DIFF
--- a/src/citra_qt/debugger/graphics_vertex_shader.cpp
+++ b/src/citra_qt/debugger/graphics_vertex_shader.cpp
@@ -65,6 +65,15 @@ QVariant GraphicsVertexShaderModel::headerData(int section, Qt::Orientation orie
     return QVariant();
 }
 
+static std::string SelectorToString(u32 selector) {
+    std::string ret;
+    for (int i = 0; i < 4; ++i) {
+        int component = (selector >> ((3 - i) * 2)) & 3;
+        ret += "xyzw"[component];
+    }
+    return ret;
+}
+
 // e.g. "-c92[a0.x].xyzw"
 static void print_input(std::ostringstream& output, const SourceRegister& input,
                         bool negate, const std::string& swizzle_mask, bool align = true,
@@ -133,6 +142,7 @@ QVariant GraphicsVertexShaderModel::data(const QModelIndex& index, int role) con
                 break;
 
             case OpCode::Type::Arithmetic:
+            case OpCode::Type::MultiplyAdd:
             {
                 // Use custom code for special instructions
                 switch (opcode.EffectiveOpCode()) {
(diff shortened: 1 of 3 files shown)


yuriks:
With #1048 fixing the monospace font issue. (The widget didn't use a monospace font only on Windows and I hadn't noticed it.) I'm going to probably revert the change where I use tabs to align stuff in the disassembly view and go back to spaces.

bunnei:
Tested this on Windows, and added some minor comments.. otherwise, LGTM.

Subv:
Does this also reduce the spacing between attribute rows? IMO it's too big.

neobrain:
(reserved; waiting for reply to my comments)

yuriks:
@Subv: That's what the `setContentMargins` call does.

@neobrain: What?

yuriks:
Here's a screenshot of how it looks now, btw: (compare against the one in the comment above)

Informative picture which I should have posted earlier

yuriks:
Rebased and re-worked disassembler part of the PR. Also added more fixes to the debugger. See individual commits (but *please* try to comment on the main diff).

shader-debugger2

src/citra_qt/debugger/graphics_vertex_shader.cpp
```diff
@@ -65,6 +58,28 @@ QVariant GraphicsVertexShaderModel::headerData(int section, Qt::Orientation orie
     return QVariant();
 }
 
+std::string SelectorToString(u32 selector) {
```
neobrain: static

src/citra_qt/debugger/graphics_vertex_shader.cpp
```diff
@@ -250,12 +292,23 @@ QVariant GraphicsVertexShaderModel::data(const QModelIndex& index, int role) con
...
+        // Highlight current instruction
+        int current_record_index = par->cycle_index->value();
+        if (current_record_index < par->debug_data.records.size()) {
+            const auto& current_record = par->debug_data.records[current_record_index];
+            if (index.row() == current_record.instruction_offset) {
+                return QColor(255, 255, 63);
+            }
+        }
+
         // Highlight instructions which have no debug data associated to them
```
neobrain: This isn't really a "highlight" anymore, I guess.

src/citra_qt/debugger/graphics_vertex_shader.cpp
```diff
@@ -288,6 +341,8 @@ GraphicsVertexShaderWidget::GraphicsVertexShaderWidget(std::shared_ptr< Pica::De
         : BreakPointObserverDock(debug_context, "Pica Vertex Shader", parent) {
     setObjectName("PicaVertexShader");
 
+    std::memset(&input_vertex, 0, sizeof(input_vertex));
```
neobrain: ?

If this is about the "unset inputs" thing that you reported before, then I have to say this is just a bad workaround. Instead, shader debug data shouldn't ever be gathered in the first place until a VertexLoaded breakpoint is hit.

src/citra_qt/debugger/graphics_vertex_shader.cpp
```diff
@@ -339,6 +391,8 @@ GraphicsVertexShaderWidget::GraphicsVertexShaderWidget(std::shared_ptr< Pica::De
             // Create an HBoxLayout to store the widgets used to specify a particular attribute
             // and store it in a QWidget to allow for easy hiding and unhiding.
             auto row_layout = new QHBoxLayout;
+            row_layout->setContentsMargins(0, 0, 0, 0);
```
neobrain: comment?

src/citra_qt/debugger/graphics_vertex_shader.cpp
```diff
@@ -358,20 +412,24 @@ GraphicsVertexShaderWidget::GraphicsVertexShaderWidget(std::shared_ptr< Pica::De
         input_data_group->setLayout(sub_layout);
         main_layout->addWidget(input_data_group);
     }
-    {
-        auto sub_layout = new QHBoxLayout;
-        sub_layout->addWidget(binary_list);
-        main_layout->addLayout(sub_layout);
-    }
+
+    binary_list->setSizePolicy(QSizePolicy::Preferred, QSizePolicy::MinimumExpanding);
```
neobrain: comment?

src/citra_qt/debugger/graphics_vertex_shader.cpp
yuriks: Yes, it's a bad workaround, but it stops Citra from crashing every time it stops on a breakpoint while the debugger is on-screen. I don't want to spend the time redesigning the code to avoid this while still allowing the user to edit the input if they want.

yuriks: (And either way I suspect the solution on this side would be exactly this memset to default the fields to 0. To make it absolutely crash proof the shader interpreter needs to be fixed instead.)

yuriks:
Addressed

src/citra_qt/debugger/graphics_vertex_shader.cpp
neobrain: There's no need to redesign anything - just don't run ProduceDebugInfo until the first VertexLoaded breakpoint has been reached. That's a sensible thing to do and doesn't require bad workarounds either.

yuriks: That causes crashes elsewhere in the interface because it expects to always have valid debug info when the UI is active. I'd done all this before, but didn't notice the input editing feature, and my rework couldn't deal with it, so I had to scrap it and didn't do it again. It's future work.

neobrain: Then a TODO should be added explaining that this memset is just a bad workaround.

neobrain:
Still thinking a TODO should be added on the memset.

yuriks:
Will do. That was before that discussion ended.

yuriks:
Updated

yuriks:
Ping?

PR #1055, opened 2015-08-19
Shader: Implement SGE, SGEI and SLT in interpreter/JIT

aroulin:
Implements SGE, SGEI and SLT shader instructions in the interpreter and the JIT.
Simple hardware tests are available here: https://github.com/aroulin/3ds-gpu-tests.

### DIFF
--- a/src/video_core/shader/shader_interpreter.cpp
+++ b/src/video_core/shader/shader_interpreter.cpp
@@ -278,6 +278,20 @@ void RunInterpreter(UnitState<Debug>& state) {
                 break;
             }
 
+            case OpCode::Id::SGE:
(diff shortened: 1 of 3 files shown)


yuriks:
LGTM

neobrain:
Please document the tests better. In particular it should include:

a) The tested hypothesis
b) How the hypothesis is tested
c) The expected output if the hypothesis is true
d) The expected output if the hypothesis is false

Currently, it seems like there's zero documentation and it's basically just a homebrew which uses SGE/SGEI/SLT.

bunnei:
LGTM. Could you move your tests to a shader module within https://github.com/citra-emu/hwtests/?

PR #1056, opened 2015-08-20
emitter: Minor cleanup

lioncash:
The last four commits can be squashed down to one if necessary.


bunnei:
LGTM. I don't see a point in squashing.

PR #1058, opened 2015-08-21
emitter: Remove pointer casts

lioncash:
This should also technically silence a few ubsan warnings if used when the shader JIT is running.


Subv:
LGTM

archshift:
LGTM

PR #1059, opened 2015-08-23
GPU: Implemented register 0x22A PICA_REG_DRAW_VERTEX_OFFSET

Subv:
This is the equivalent of the "first" parameter in glDrawArrays, it tells the GPU the vertex index at which to start rendering.

Tested with https://github.com/Subv/3ds-homebrew/tree/master/gpu_draw_arrays_offset

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -231,7 +231,8 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
 
             for (unsigned int index = 0; index < regs.num_vertices; ++index)
             {
-                unsigned int vertex = is_indexed ? (index_u16 ? index_address_16[index] : index_address_8[index]) : index;
+                // Indexed rendering doesn't use the start offset
+                unsigned int vertex = is_indexed ? (index_u16 ? index_address_16[index] : index_address_8[index]) : (index + regs.vertex_offset);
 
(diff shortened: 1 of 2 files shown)


neobrain:
How does this register behave during indexed rendering?

src/video_core/command_processor.cpp
```diff
@@ -229,7 +229,13 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
             Shader::UnitState<false> shader_unit;
             Shader::Setup(shader_unit);
 
-            for (unsigned int index = 0; index < regs.num_vertices; ++index)
+            u32 start_offset = regs.vertex_offset;
```
neobrain: Eh, this should just be merged into the `vertex` initialization: `vertex = ... : index + regs.vertex_offset`.

bunnei:
LGTM

yuriks:
LGTM

src/video_core/pica.h
```diff
@@ -757,7 +757,12 @@ struct Regs {
     // Number of vertices to render
     u32 num_vertices;
 
-    INSERT_PADDING_WORDS(0x5);
+    INSERT_PADDING_WORDS(0x1);
+
+    // The index of the first vertex to render
+    u32 vertex_offset;
```
neobrain: This should be added to the two registers lists below. Poke @bunnei for details.

Please submit this as a followup PR.

PR #1062, opened 2015-08-23
Shader: RCP and RSQ computes only the 1st component

aroulin:
RCP (RSQ) computes only the source vector's first component reciprocal (reciprocal of square root). It is not surprising as these are transcendental functions typically executed on SFUs.

Confirmed by tests available here:
- https://github.com/aroulin/3ds-gpu-tests/blob/master/rcp-tests/source/main.c
- https://github.com/aroulin/3ds-gpu-tests/blob/master/rsq-tests/source/main.c

### DIFF
--- a/src/video_core/shader/shader_interpreter.cpp
+++ b/src/video_core/shader/shader_interpreter.cpp
@@ -237,7 +237,7 @@ void RunInterpreter(UnitState<Debug>& state) {
             {
                 Record<DebugDataRecord::SRC1>(state.debug, iteration, src1);
                 Record<DebugDataRecord::DEST_IN>(state.debug, iteration, dest);
-                float24 rsq_res = float24::FromFloat32(1.0f / sqrt(src1[0].ToFloat32()));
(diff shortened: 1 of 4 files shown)


src/video_core/shader/shader_interpreter.cpp
```diff
@@ -221,13 +221,13 @@ void RunInterpreter(UnitState<Debug>& state) {
             {
                 Record<DebugDataRecord::SRC1>(state.debug, iteration, src1);
                 Record<DebugDataRecord::DEST_IN>(state.debug, iteration, dest);
+                // TODO: Be stable against division by zero!
```
yuriks: fyi, this is correct behavior for division by zero: it should yield infinite, like this will

src/video_core/shader/shader_interpreter.cpp
```diff
@@ -238,13 +238,13 @@ void RunInterpreter(UnitState<Debug>& state) {
...
```
yuriks: Same here.

yuriks:
LGTM (I had also independently implemented the exact same changes.)

bunnei:
LGTM

PR #1063, opened 2015-08-23
HWRenderer: Only reload the framebuffer from gpu memory if the hw renderer is in use during a breakpoint

Subv:
This should prevent some weird behaviors when the sw renderer is being used during a breakpoint


bunnei:
LGTM

PR #1065, opened 2015-08-24
Shader FP compliance fixes

yuriks:
Fixes or verifies some floating-point edges cases in the shader interpreter and JIT. The most visible effect of this PR is that OoT doesn't have missing interface elements anymore.

oot

- Automated test suite at aroulin/3ds-gpu-tests#1
- Some info gathered during the making of this PR can be found in 3DBrew: [Shader_Instruction_Set#Floating-Point_Behavior](http://www.3dbrew.org/wiki/Shader_Instruction_Set#Floating-Point_Behavior)
- This has a modest perf impact in the JIT: 2.21ms -> 2.62ms (+18%) in the OoT inventory screen.

Thanks @FioraAeterna for help and ideas when doing hwtests and implementing this in the JIT.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -1021,12 +1021,20 @@ struct float24 {
         return ret;
     }
 
+    static float24 Zero() {
+        return FromFloat32(0.f);
+    }
+
     // Not recommended for anything but logging
     float ToFloat32() const {
         return value;
     }
 
     float24 operator * (const float24& flt) const {
+        if ((this->value == 0.f && flt.value == flt.value) ||
+            (flt.value == 0.f && this->value == this->value))
+            // PICA gives 0 instead of NaN when multiplying by inf
+            return Zero();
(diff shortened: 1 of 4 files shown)


src/video_core/pica.h
```diff
@@ -1021,12 +1021,20 @@ struct float24 {
...
+        return FromFloat32(0.f);
+    }
+
     // Not recommended for anything but logging
     float ToFloat32() const {
         return value;
     }
 
     float24 operator * (const float24& flt) const {
+        if ((this->value == 0.f && flt.value == flt.value) ||
```
FioraAeterna: maybe isnan would be more appropriate here?

yuriks: I was testing using isinf earlier, but shied away due to performance concerns (this class is widely used in all parts of the graphics engine.) I'll do some actual benchmarking though.

aroulin:
LGTM (for test suite and gathered results too)

bunnei:
LGTM

PR #1072, opened 2015-08-26
SVC: Advance time when calling GetSystemTick to escape busy-wait loops

yuriks:
Cubic Ninja waited for the frame to end by spinning on a loop calling
GetSystemTick while doing nothing else. Since GetSystemTick doesn't
cause a reschedule (which advances time), this meant that very little
emulated time would pass inside that loop, causing the game to spend
most of the frame burning away CPU.

### DIFF
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -755,7 +755,10 @@ static void SleepThread(s64 nanoseconds) {
 
 /// This returns the total CPU ticks elapsed since the CPU was powered-on
 static s64 GetSystemTick() {
-    return (s64)CoreTiming::GetTicks();
+    s64 result = CoreTiming::GetTicks();
+    // Advance time to defeat dumb games (like Cubic Ninja) that busy-wait for the frame to end.
+    Core::g_app_core->AddTicks(150); // Measured time between two calls on a 9.2 o3DS with Ninjhax 1.1b
+    return result;
 }
 
 /// Creates a memory block at the specified address with the specified permissions and size


src/core/hle/svc.cpp
```diff
@@ -755,7 +755,10 @@ static void SleepThread(s64 nanoseconds) {
 
 /// This returns the total CPU ticks elapsed since the CPU was powered-on
 static s64 GetSystemTick() {
-    return (s64)CoreTiming::GetTicks();
+    s64 result = CoreTiming::GetTicks();
+    // Advance time to defeat dumb games (like Cubic Ninja) that busy-wait for the frame to end.
+    Core::g_app_core->AddTicks(4000);
```
bunnei: What was the thought process in choosing 4000?

yuriks: It's what we add during a reschedule. (It also gave a balanced perf boost. Smaller values increase CPU usage proportionally.)

archshift:
Shouldn't we mention in a TODO that this is essentially a hack?

hrydgard:
We do this too for similar functions in PPSSPP, but usually after measuring approximately how much time the function actually takes on real hardware first, and advancing that number of cycles. 

Skipping some cycles is definitely OK but 4000 might be quite some bit into "hack" territory.

yuriks:
@hrydgard Good point. I had measured a loop doing the call (+ the small overhead of the ctrulib wrapper function) and ended up with 150 cycles per call. I left it at 4000 because that would counter-act the "fastness" of the rest of the emulator somewhat, but now that you mention it, it makes no sense to make a call slower to work around timing issues in the rest of the emu, so I'll change it to 150.

The CPU usage should go down eventually when we make other things take more time, and with 150 it's even not that bad. (Citra was doing 4 cycles per loop vs. 154 on hw, so it's already about 37x less iterations.)

yuriks:
Updated

bunnei:
LGTM

PR #1078, opened 2015-08-26
Shader JIT: ifdef out reference to ifdef'd out shader_map

archshift:
`shader_map` was only defined on x86 architectures, but was cleared on shutdown
with no `#ifdef` protection. `#ifdef` this out so non-x86_64 architectures can be built.


Subv:
Didn't we have a PR for this already?

bunnei:
LGTM

archshift:
@Subv evidently not

PR #1079, opened 2015-08-27
Shader JIT: Fix float to integer rounding in MOVA

aroulin:
MOVA converts new address register values from floats to integers **using truncation**.
- specified here http://3dbrew.org/wiki/Shader_Instruction_Set#Instructions
- I validated this behavior on hardware: https://github.com/aroulin/3ds-gpu-tests/blob/master/mova-tests/source/main.c

This fixes an issue in The Legend of Zelda: A Link Between Worlds where the UI boxes were not drawn
in the main menu screen.

Before (with the shader JIT): 
albw_before

After (with the shader JIT and the shader interpreter):

albw_after


bunnei:
LGTM

PR #1083, opened 2015-08-28
Common: Fix MicroProfile compilation in MSVC2015

yuriks:
This typedef happened to be defined through a include path in the standard libraries in 2013, but isn't anymore in 2015. Ideally I'll upstream the fix, but this will do until then.

### DIFF
--- a/src/common/microprofile.h
+++ b/src/common/microprofile.h
@@ -11,6 +11,11 @@
 #define MICROPROFILE_CONTEXT_SWITCH_TRACE 0
 #define MICROPROFILE_PER_THREAD_BUFFER_SIZE (2048<<12) // 8 MB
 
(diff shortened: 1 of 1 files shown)


hrydgard:
I would

    #if defined(_MSC_VER) && _MSC_VER >= 1900

yuriks:
@hrydgard The symbol is used conditionally on _WIN32, and it wasn't ever explicitly included. It only got in because of an earlier stdlib include which eventually included concrt.h, which has the same typedef line. Since (I didn't know this) identical duplicated typedefs silently work, this will work fine in any case.

Subv:
Tested, LGTM

bunnei:
LGTM

PR #1084, opened 2015-08-28
[citra-qt] Fix refresh issue

LittleWhite-tb:
This fix the refresh issue and add a nice citra logo on main window.

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -11,17 +11,16 @@
 #include "bootmanager.h"
 #include "main.h"
 
-#include "common/string_util.h"
-#include "common/scm_rev.h"
 #include "common/key_map.h"
+#include "common/scm_rev.h"
 #include "common/microprofile.h"
+#include "common/string_util.h"
 
 #include "core/core.h"
 #include "core/settings.h"
 #include "core/system.h"
 
 #include "video_core/debug_utils/debug_utils.h"
-
 #include "video_core/video_core.h"
 
 #include "citra_qt/version.h"
@@ -87,6 +86,9 @@ public:
     }
 
     void paintEvent(QPaintEvent* ev) override {
+        if (do_painting) {
+            QPainter painter(this);
+        }
     }
 
     void resizeEvent(QResizeEvent* ev) override {
@@ -94,8 +96,12 @@ public:
         parent->OnFramebufferSizeChanged();
     }
 
+    void DisablePainting() { do_painting = false; }
+    void EnablePainting() { do_painting = true; }
+
 private:
     GRenderWindow* parent;
+    bool do_painting;
 };
 
 GRenderWindow::GRenderWindow(QWidget* parent, EmuThread* emu_thread) :
@@ -269,8 +275,10 @@ void GRenderWindow::OnMinimalClientAreaChangeRequest(const std::pair<unsigned,un
 
 void GRenderWindow::OnEmulationStarting(EmuThread* emu_thread) {
     this->emu_thread = emu_thread;
+    child->DisablePainting();
 }
 
 void GRenderWindow::OnEmulationStopping() {
     emu_thread = nullptr;
+    child->EnablePainting();
 }
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -82,21 +82,10 @@ class GGLWidgetInternal : public QGLWidget
 {
 public:
     GGLWidgetInternal(QGLFormat fmt, GRenderWindow* parent)
-                     : QGLWidget(fmt, parent), splash(":/data/citra.svg"), parent(parent) {
+                     : QGLWidget(fmt, parent), parent(parent) {
     }
 
     void paintEvent(QPaintEvent* ev) override {
-        if (do_painting) {
-            QPainter painter(this);
-
-            int scaled_width = width()/3;
-            int scaled_height = width()/3;
-
-            painter.drawImage(QRect(width()/2-scaled_width/2,
-                                    height()/2-scaled_height/2,
-                                    scaled_width,
-                                    scaled_height), splash);
-        }
     }
 
     void resizeEvent(QResizeEvent* ev) override {
@@ -104,13 +93,31 @@ public:
         parent->OnFramebufferSizeChanged();
     }
 
-    void DisablePainting() { do_painting = false; }
-    void EnablePainting() { do_painting = true; }
-
 private:
     GRenderWindow* parent;
+};
+
+class GSplashWidgetInternal : public QWidget
+{
+public:
+    GSplashWidgetInternal(QWidget* parent)
+                     : QWidget(parent), splash(":/data/citra.svg") {
+    }
+
+    void paintEvent(QPaintEvent* ev) override {
+        QPainter painter(this);
+
+        int scaled_width = width()/3;
+        int scaled_height = width()/3;
+
+        painter.drawImage(QRect(width()/2-scaled_width/2,
+                                height()/2-scaled_height/2,
+                                scaled_width,
+                                scaled_height), splash);
+    }
+
+private:
     QImage splash;
-    bool do_painting;
 };
 
 GRenderWindow::GRenderWindow(QWidget* parent, EmuThread* emu_thread) :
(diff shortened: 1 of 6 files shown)


LittleWhite-tb:
This is my attempt to fix lack of refresh when no emulation is done (and so, no rendering to the window).

I have splitted my work in three commits:
* First commit fixes the issue. Adding a QPainter let Qt to repaint the background with default UI background colour.
* I wanted to add citra logo to have a nicer program.
* Using drawImage() from Qt caused several disagreement (crashes) because the OpenGL context is stolen by the emulator. I did not found any proper way to make it no crash. So my last solution was to have two widget, one for emulation with OpenGL context that is given to the core, one for the logo.

Here, my thoughts around this work and how it is done:
* I used a resource file to embedded the logo in the final executable. Would you prefer to keep it as an external file ?
* In second commit, to get drawImage() to work, I had to remove QGLFormat::CoreProfile. It is back in third commit.
* I would be happy to find a proper solution that having two widgets. I did not investigated enough about getting back the OpenGL context ownership to the GUI. Do you have any ideas why after stopping the emulation, drawImage() still crash ? (I have tried once, moveContext() in OnEmulationStopping(), but no effect.
* Another solution would be to implement the logo rendering in OpenGL and avoid any use of Qt rendering stuff.

src/citra_qt/bootmanager.cpp
```diff
@@ -98,6 +97,29 @@ class GGLWidgetInternal : public QGLWidget
     GRenderWindow* parent;
 };
 
+class GSplashWidgetInternal : public QWidget
```
lioncash: `class GSplashWidgetInternal final : public QWidget`

bunnei:
Can you show a screenshot?

src/citra_qt/bootmanager.cpp
bunnei: `SplashWidgetInternal final : public QWidget`

LittleWhite-tb: I will rename all class with this 'G' prefix later, in another pull request. I have kept it this time just for consistency.

LittleWhite-tb:
Sure :)

citra-splash

archshift:
Honestly I don't think having a Citra logo there looks that great. It's probably better if you just leave it blank, ideally we wouldn't show the rendering area at all if no emulation is occurring. 

Also, what happens if emulation is paused?

LittleWhite-tb:
At the moment, when paused, we don't have refresh, so the last picture stay here, put can be overpainted by over windows and make it "unclean" (no refresh bug).

neobrain:
The Citra logo wlll look quite horrible on non-standard color themes there, fwiw. Think of the logo on orange background... ;)

Could just always paint a plain background. Or make a frame around the citra logo, with well-defined inner-frame background.

src/citra_qt/bootmanager.cpp
```diff
@@ -269,8 +295,22 @@ void GRenderWindow::OnMinimalClientAreaChangeRequest(const std::pair<unsigned,un
 
 void GRenderWindow::OnEmulationStarting(EmuThread* emu_thread) {
     this->emu_thread = emu_thread;
+
+    // Swaps to put in front the render widget
```
neobrain: Don't describe what code does; instead describe what it's intended to do. In particular, use imperatives instead of describing phrases like "Swaps...".

=> "Put render widget in front". Similarly below.

src/citra_qt/main.cpp
```diff
@@ -59,7 +59,6 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     statusBar()->hide();
 
     render_window = new GRenderWindow(this, emu_thread.get());
-    render_window->hide();
```
neobrain: Will this spawn a render window even if single-window-mode is disabled? I don't think we would want that.

RHY3756547:
image
Having it as a black semitransparent watermark might work a little better for unusual background colours, and is a bit less jarring. (the above is just an edit of the previous screenshot)

RavenHome1:
just a quick notice wouldn't it be cool if it had some changing color rainbow effects on the balloon shaped things in the center

archshift:
@RHY3756547 I really like this. It's much more fitting than the normal Citra icon.

neobrain:
@RHY3756547 That looks really great indeed :) Seconding what @archshift said, it's better fitting than the normal icon even for your default color theme.

Would be cool if someone could test this with some nonstandard themes.

yuriks:
I think the centered watermark still looks somewhat weird, but I agree it's better than the colored logo.

RHY3756547:
Here's PNG and SVG versions of the watermark logo to try it out:
https://dl.dropboxusercontent.com/u/12239448/citrawater.png
https://dl.dropboxusercontent.com/u/12239448/citrawater.svg
It's 10% opaque right now but you can change that easily enough in the svg, line 9.

neobrain:
Uhm. Can't you just change your color theme temporarily to try it out?

RHY3756547:
Don't have build setup on this PC, and other one is out of commission. Was thinking maybe the pull request owner could give it a shot.

src/citra_qt/main.cpp
LittleWhite-tb: With, or without this one, the render_window still show up (even on master). I would like to worry on everything related to this one/multiple window mode in another PR/Bug report.

neobrain: Then what specifically does removing these hide() calls achieve?

LittleWhite-tb: The point to remove the hide() in ShutdownGame() is to let appear the logo again.
Of course, this is not the best solution since the windows will still open in "multiple window mode" and that's not what we want. So, the hide call should be put in a condition checking in which mode we are.

LittleWhite-tb:
I have done some tests.
@RHY3756547 : your SVG gave me this result : 
weird-colours
I don't know why the top slice was darker :(

I have tried some other color scheme (the one given from Qt examples) :
coffee
pagefold

I will try more, since here, I did not tested weird coloring scheme.

LittleWhite-tb:
Here some more screens. I hope nobody uses such "ugly" themes. In such case, I guess all logo will fail to give nice output.
brown
darkkaki
pink
red

(But, these are not even themes. It's just me playing with the background color of the widget.)
Yesterday, I tried all themes given by Qt in their examples, and apart of the pictures above (in previous message), the grey seems to be often used as background color. Also, themes depends on your OS.

Finally, I have tried the background color that we can see here : http://blog.qt.io/blog/2012/10/30/cleaning-up-styles-in-qt5-and-adding-fusion/

fusion2
fusion3
old-motif
plastique

yuriks:
Hm... isn't a easier way to do this simply to have the logo be a pure-black stencil with alpha, then draw it with something like 10% alpha on the background?

RHY3756547:
Where's the circle of white under it coming from? Both the png and svg I linked blend correctly onto those backgrounds under normal circumstances...

LittleWhite-tb:
@RHY3756547 Since your SVG was giving me incorrect output (darker top slice) (not your fault, I suspect something wrong with Qt SVG support), I have certainly badly redone one myself.

@yuriks With transparent SVG :
red-alpha


(Again, I have removed some details from the original logo due to my bad use of Inkscape. If someone can make a logo witch is transparent (for this test, I have set the slice to 75/255. 10 % was not enough to let the logo appear).

LittleWhite-tb:
With better SVG, and using pure black with 10 % alpha
black
default
red

LittleWhite-tb:
There, new logo in place. Also, I have added small test to hide the render window when stopping the emulation in multiple window mode for neobrain :)

bunnei:
This PR is getting kind of stale... I'd suggest resubmitting this PR with just the refresh issue fixed (so we can get that merged), the splashscreen component of this should probably wait until we have the new artwork finalized.

LittleWhite-tb:
Here, the PR for the refresh issue only  -> #1171

bunnei:
@LittleWhite-tb I merged #1171, and the splash screen will probably be superseded by #1095. I'm going to close this since it is now outdated in almost every way, but feel free to resubmit a new PR related to the splash screen if you have any idea(s) on how it might tie in with the new game list view.

LittleWhite-tb:
Completely agree with you and I expected that. I will wait for gamelist (#1095), to see if I can come with nice ideas and make interesting PR.

PR #1086, opened 2015-08-30
Only check src/ directory for whitespace errors in the pre-commit hook

yuriks:
This allows importing of external libraries into externals/ without
having to reformat them. Unfortunately it also allows whitespace to be
introduced in files like the root CMakeLists.txt, but that is a small
downside compared to the tradeoff.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -3,8 +3,10 @@
 # Enforce citra's whitespace policy
 git config --local core.whitespace tab-in-indent,trailing-space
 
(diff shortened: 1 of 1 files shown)


archshift:
LGTM

linkmauve:
You should also explicitly specify CMakeLists.txt on the same line, after which LGTM.

yuriks:
Updated

bunnei:
LGTM

PR #1087, opened 2015-08-30
Replace the previous OpenGL loader with a glad-generated 3.3 one

yuriks:
The main advantage of switching to glad from glLoadGen is that, apart from being actively maintained, it supports a customizable entrypoint loader function, which makes it possible to also support OpenGL ES.

I originally revived #213, since libepoxy now apparently supports Windows. However, it still seems to have issues in Windows (namely, it requires being built as a DLL, and doesn't work in Release builds) so I opted to just update the loader for now. This new loader can support OpenGL ES if desired. (But I'll rebase that support on top of this on a later PR.)


archshift:
LGTM except for the travis build on Linux, where the error looks to be using a feature not available in its version of CMake.

bunnei:
Looks fine to me (I didn't actually test this, FWIW)

yuriks:
Just noticed glad seems to not spam the OS X CI build with warnings too, which is nice.

Subv:
Lgtm

yuriks:
@linkmauve says he's ok with this as an improvement for now, but would like to transition to libepoxy (or similar lib with support for liasing) anyway in the future.

PR #1088, opened 2015-08-30
x64: Proper stack alignment in shader JIT function calls

aroulin:
Fixes misaligned stack when calling external functions from the shader JIT. Both the Win64 and the System V ABIs require a 16-byte aligned stack when calling a function.

[Agner's Calling Conventions 5. Stack Alignment](http://www.agner.org/optimize/calling_conventions.pdf)
    
    The 64 bit systems keep the stack aligned by 16. The stack word size
    is 8 bytes, but the stack must be aligned by 16 before any call instruction.
    Consequently, the value of the stack pointer is always 8 modulo 16 at
    the entry of a procedure.

[System V ABI 3.2.2 The Stack Frame](http://www.x86-64.org/documentation/abi.pdf):

    […], the value (%rsp + 8) is always a multiple of 16 when control 
    is transferred to the function entry point.

[Win64 Software Conventions](https://msdn.microsoft.com/en-us/library/ew5tede7.aspx):

    The stack will always be maintained 16-byte aligned, except 
    within the prolog (for example, after the return address is pushed)

External functions are for the time being only used to emulate EX2 and LG2 instructions and these might be replaced in a near future with inline approximations. Nevertheless, it is certainly still useful to have a clean stack alignment system (and maybe geometry shader instructions will use external interpreter functions as a first step).

The list of registers to save is for now static. The next enhancement would be to have a register caching system that, additionally to cache registers, would keep track of which registers are currently in use and have to be saved.

The problem is fixed by using Dolphin's `ABI_{Push,Pop}RegistersAndAdjustStack` instead of `ABI_{Push,Pop}All{Caller,Callee}SavedRegsAndAdjustStack`. Dolphin's code is x64-only therefore the remaining x86 bits are removed from the ABI files.

Previous functions `ABI_{Push,Pop}All{Caller,Callee}SavedRegsAndAdjustStack` are removed because of multiple problems. One is that they don't make sure the stack is aligned as they just push **a hard-coded odd number of registers**. This can be misleading, for example the following will break the stack alignment:

    // Function entry: stack pointer misaligned
    ABI_PushAllCalleeSavedRegsAndAdjustStack();
    // OK stack pointer is aligned
    
    // [...]
    
    // later we want to call a function so need to save caller-saved register
    ABI_PushAllCallerSavedRegsAndAdjustStack();
    // stack pointer is misaligned now !
    CALL(myfunc); // calling with misaligned stack

Example of a misaligned stack when emulating the EX2 shader instruction by calling `std::exp2` 
misaligned_ex2_call

### DIFF
--- a/src/common/bit_set.h
+++ b/src/common/bit_set.h
@@ -1,0 +1,189 @@
+// This file is under the public domain.
+
+#pragma once
+
+#include <cstddef>
+#ifdef _WIN32
+#include <intrin.h>
+#endif
+#include <initializer_list>
+#include <type_traits>
+#include "common/common_types.h"
+
+// namespace avoids conflict with OS X Carbon; don't use BitSet<T> directly
+namespace Common {
+
+// Helper functions:
+
+#ifdef _WIN32
+template <typename T>
+static inline int CountSetBits(T v)
+{
+    // from https://graphics.stanford.edu/~seander/bithacks.html
+    // GCC has this built in, but MSVC's intrinsic will only emit the actual
+    // POPCNT instruction, which we're not depending on
+    v = v - ((v >> 1) & (T)~(T)0/3);
+    v = (v & (T)~(T)0/15*3) + ((v >> 2) & (T)~(T)0/15*3);
+    v = (v + (v >> 4)) & (T)~(T)0/255*15;
+    return (T)(v * ((T)~(T)0/255)) >> (sizeof(T) - 1) * 8;
+}
+static inline int LeastSignificantSetBit(u8 val)
+{
+    unsigned long index;
+    _BitScanForward(&index, val);
+    return (int)index;
+}
+static inline int LeastSignificantSetBit(u16 val)
+{
+    unsigned long index;
+    _BitScanForward(&index, val);
+    return (int)index;
+}
+static inline int LeastSignificantSetBit(u32 val)
+{
+    unsigned long index;
+    _BitScanForward(&index, val);
+    return (int)index;
+}
+static inline int LeastSignificantSetBit(u64 val)
+{
+    unsigned long index;
+    _BitScanForward64(&index, val);
+    return (int)index;
+}
+#else
+static inline int CountSetBits(u8 val) { return __builtin_popcount(val); }
+static inline int CountSetBits(u16 val) { return __builtin_popcount(val); }
+static inline int CountSetBits(u32 val) { return __builtin_popcount(val); }
+static inline int CountSetBits(u64 val) { return __builtin_popcountll(val); }
+static inline int LeastSignificantSetBit(u8 val) { return __builtin_ctz(val); }
+static inline int LeastSignificantSetBit(u16 val) { return __builtin_ctz(val); }
+static inline int LeastSignificantSetBit(u32 val) { return __builtin_ctz(val); }
+static inline int LeastSignificantSetBit(u64 val) { return __builtin_ctzll(val); }
+#endif
+
+// Similar to std::bitset, this is a class which encapsulates a bitset, i.e.
+// using the set bits of an integer to represent a set of integers.  Like that
+// class, it acts like an array of bools:
+//     BitSet32 bs;
+//     bs[1] = true;
+// but also like the underlying integer ([0] = least significant bit):
+//     BitSet32 bs2 = ...;
+//     bs = (bs ^ bs2) & BitSet32(0xffff);
+// The following additional functionality is provided:
+// - Construction using an initializer list.
+//     BitSet bs { 1, 2, 4, 8 };
+// - Efficiently iterating through the set bits:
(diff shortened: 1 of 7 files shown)


src/common/bit_set.h
```diff
@@ -0,0 +1,190 @@
+// This file is under the public domain.
```
yuriks: What do we want to do about licensing here?

src/common/bit_set.h
```diff
@@ -0,0 +1,190 @@
...
+static inline int CountSetBits(u32 val) { return __builtin_popcount(val); }
+static inline int CountSetBits(u64 val) { return __builtin_popcountll(val); }
+static inline int LeastSignificantSetBit(u8 val) { return __builtin_ctz(val); }
+static inline int LeastSignificantSetBit(u16 val) { return __builtin_ctz(val); }
+static inline int LeastSignificantSetBit(u32 val) { return __builtin_ctz(val); }
+static inline int LeastSignificantSetBit(u64 val) { return __builtin_ctzll(val); }
+#endif
+
+// namespace avoids conflict with OS X Carbon; don't use BitSet<T> directly
+namespace BS
```
yuriks: I think this namespace can be removed and the whole file be put in the `Common` namespace instead.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -560,7 +550,7 @@ void JitCompiler::Compile_NOP(Instruction instr) {
 }
 
 void JitCompiler::Compile_END(Instruction instr) {
-    ABI_PopAllCalleeSavedRegsAndAdjustStack();
+    ABI_PopRegistersAndAdjustStack(ABI_ALL_CALLEE_SAVED, 8);
```
yuriks: Just to confirm, these are 8 because they're coming directly from a CALL, which leaves us aligned on 8 mod 16, right? And then the others used 0 because this function has already aligned the stack on 0 mod 16?

yuriks:
ABI and BitSet stuff was imported from Dolphin, so I just gave it a quick read, but mostly it looks good apart from the small comments and deciding if we want to add a license header to that other file.

src/video_core/shader/shader_jit_x64.cpp
aroulin: Correct.

src/common/bit_set.h
bunnei: Unless we modify it, shouldn't this be kept as-is?

src/common/x64/abi.h
```diff
@@ -73,6 +47,13 @@
 #define ABI_PARAM5 R8
 #define ABI_PARAM6 R9
 
+// FIXME: avoid pushing all 16 XMM registers when possible? most functions we call probably
```
bunnei: Nitpick, but the format for this type of stuff is:
<pre>
TODO(aroulin): Avoid pushing all 16 XMM registers when possible. Most functions we call probably
don't actually clobber them.
</pre>
Notice I fixed your sentences... we're sort of sticklers for proper writing :)

src/common/x64/abi.h
```diff
@@ -64,7 +34,11 @@
 #define ABI_PARAM3 R8
 #define ABI_PARAM4 R9
 
-#else  //64-bit Unix (hopefully MacOSX too)
+// xmm0-xmm15 use the upper 16 bits in the functions that push/pop registers.
+#define ABI_ALL_CALLER_SAVED \
+    (BitSet32 { RAX, RCX, RDX, R8, R9, R10, R11, \
+                XMM0+16, XMM1+16, XMM2+16, XMM3+16, XMM4+16, XMM5+16 })
+#else  //64-bit Unix / OS X
```
bunnei: Two spaces between `else` and the comment looks weird... Typo?

bunnei:
Ditto with the minor nits.
Otherwise, looks good. Feel free to squash.

src/common/bit_set.h
aroulin: The file has a slight modification, but the decision was taken to leave it as-is until we get legal clarification.

src/common/x64/abi.h
aroulin: This comment is from Dolphin. I fixed the writing style nevertheless.

bunnei:
LGTM

dracojoaquin1:
SORRY IF NO CAN POST HERE FOR HELP BUT NOT FIND WHERE CAN SEND PRIVATE MENSAGE AND SORRY BAD ENGLISH
can help me?  i am download your emulator to play (Animal.Crossing.New.Leaf..EUR.3DS.Multi.Spa.3ds)
but not work have this in screen and not load nothing can help me?
i am have win 10 64 bits and me test too in win 7 64
nvidea gtx 760 2gb
16 gb ram
problem

PR #1091, opened 2015-08-31
Shader JIT: Fix SGE/SGEI NaN behavior

aroulin:
SGE is incorrectly emulated w.r.t. NaN behavior as the CMPSS SSE instruction is using NLT.
`a >= b` is not equivalent to `!(a < b)` because of NaN.

Behavior checked on hardware: 
https://github.com/aroulin/3ds-gpu-tests/tree/master/fp-tests


yuriks:
Ah, I'd missed SGE when I fixed CMP for this, thanks. LGTM

bunnei:
LGTM

PR #1095, opened 2015-09-01
Initial implementation of a game list

archshift:
To open a game, double click on a menu item. Uses the config keys `gameListRootDir` and `gameListDeepScan`. The former is, rather obviously, the root directory to scan for games in. The latter is whether you would like the game list to recurse through subdirectories to find games.

Looking for testers on different platforms!

#### Still needs to be done (in priority order):
- [x] Hide game list while a game is currently being played
- [x] Saving to `QSettings`:
  - [x] Sort
  - [x] Column sizes
- [x] Menu item (under File) for a folder selection dialog

#### For future PRs:
- SMDH integration for game icons, proper game names and descriptions, etc.
- Search/filter bar, using regex
- Different game list layouts, e.g. grid-style layout.
- Pop up window while the game list is populating (a la Dolphin)
- Option for caching game list (no re-scanning on startup)

### DIFF
--- a/src/citra_qt/CMakeLists.txt
+++ b/src/citra_qt/CMakeLists.txt
@@ -17,6 +17,7 @@ set(SRCS
             debugger/profiler.cpp
             debugger/ramview.cpp
             debugger/registers.cpp
+            game_list.cpp
             util/spinbox.cpp
             util/util.cpp
             bootmanager.cpp
@@ -42,6 +43,7 @@ set(HEADERS
             debugger/profiler.h
             debugger/ramview.h
             debugger/registers.h
+            game_list.h
             util/spinbox.h
             util/util.h
             bootmanager.h
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -1,0 +1,154 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <QHeaderView>
+#include <QThreadPool>
+#include <QVBoxLayout>
+
+#include "game_list.h"
+#include "game_list_p.h"
+
+#include "core/loader/loader.h"
+
+#include "common/common_paths.h"
+#include "common/logging/log.h"
+#include "common/string_util.h"
+
+GameList::GameList(QWidget* parent)
+{
+    QVBoxLayout* layout = new QVBoxLayout;
+
+    tree_view = new QTreeView;
+    item_model = new QStandardItemModel(tree_view);
+    tree_view->setModel(item_model);
+
+    tree_view->setAlternatingRowColors(true);
+    tree_view->setSelectionMode(QHeaderView::SingleSelection);
+    tree_view->setSelectionBehavior(QHeaderView::SelectRows);
+    tree_view->setVerticalScrollMode(QHeaderView::ScrollPerPixel);
+    tree_view->setHorizontalScrollMode(QHeaderView::ScrollPerPixel);
+    tree_view->setSortingEnabled(true);
+    tree_view->setEditTriggers(QHeaderView::NoEditTriggers);
+    tree_view->setUniformRowHeights(true);
+
+    item_model->insertColumns(0, COLUMN_COUNT);
+    item_model->setHeaderData(COLUMN_FILE_TYPE, Qt::Horizontal, "File type");
+    item_model->setHeaderData(COLUMN_NAME, Qt::Horizontal, "Name");
+    item_model->setHeaderData(COLUMN_SIZE, Qt::Horizontal, "Size");
+
+    connect(tree_view, SIGNAL(activated(const QModelIndex&)), this, SLOT(ValidateEntry(const QModelIndex&)));
+
+    // We must register all custom types with the Qt Automoc system so that we are able to use it with
+    // signals/slots. In this case, QList falls under the umbrells of custom types.
+    qRegisterMetaType<QList<QStandardItem*>>("QList<QStandardItem*>");
+
+    layout->addWidget(tree_view);
+    setLayout(layout);
+}
+
+GameList::~GameList()
+{
+    emit ShouldCancelWorker();
+}
+
+void GameList::AddEntry(QList<QStandardItem*> entry_items)
+{
+    item_model->invisibleRootItem()->appendRow(entry_items);
+}
+
+void GameList::ValidateEntry(const QModelIndex& item)
+{
+    // We don't care about the individual QStandardItem that was selected, but its row.
+    int row = item_model->itemFromIndex(item)->row();
+    QStandardItem* child_file = item_model->invisibleRootItem()->child(row, COLUMN_NAME);
+    QString file_path = child_file->data(GameListItemPath::FullPathRole).toString();
+
+    if (file_path.isEmpty())
+        return;
+    std::string std_file_path = file_path.toStdString();
+    if (!FileUtil::Exists(std_file_path) || FileUtil::IsDirectory(std_file_path))
+        return;
+    emit GameChosen(file_path);
+}
+
+void GameList::DonePopulating()
+{
+    tree_view->setEnabled(true);
+}
+
+void GameList::PopulateAsync(const QString& dir_path, bool deep_scan)
+{
+    if (!FileUtil::Exists(dir_path.toStdString()) || !FileUtil::IsDirectory(dir_path.toStdString())) {
+        LOG_ERROR(Frontend, "Could not find game list folder at %s", dir_path.toLatin1().data());
+        return;
+    }
+
+    tree_view->setEnabled(false);
+    // Delete any rows that might already exist if we're repopulating
+    item_model->removeRows(0, item_model->rowCount());
+
+    emit ShouldCancelWorker();
+    GameListWorker* worker = new GameListWorker(dir_path, deep_scan);
+
+    connect(worker, SIGNAL(EntryReady(QList<QStandardItem*>)), this, SLOT(AddEntry(QList<QStandardItem*>)), Qt::QueuedConnection);
+    connect(worker, SIGNAL(Finished()), this, SLOT(DonePopulating()), Qt::QueuedConnection);
+    // Use DirectConnection here because worker->Cancel() is thread-safe and we want it to cancel without delay.
+    connect(this, SIGNAL(ShouldCancelWorker()), worker, SLOT(Cancel()), Qt::DirectConnection);
+
+    QThreadPool::globalInstance()->start(worker);
+    current_worker = std::move(worker);
+}
+
+void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, bool deep_scan)
+{
+    const auto callback = [&](const std::string& directory,
+                                     const std::string& virtual_name) -> int {
+
+        std::string physical_name = directory + DIR_SEP + virtual_name;
+
+        if (stop_processing)
+            return -1; // A negative return value breaks the callback loop.
+
+        if (deep_scan && FileUtil::IsDirectory(physical_name)) {
+            AddFstEntriesToGameList(physical_name, true);
+        } else {
+            std::string filename_filename, filename_extension;
+            Common::SplitPath(physical_name, nullptr, &filename_filename, &filename_extension);
+
+            Loader::FileType guessed_filetype = Loader::GuessFromExtension(filename_extension);
+            if (guessed_filetype == Loader::FileType::Unknown)
+                return 0;
+            Loader::FileType filetype = Loader::IdentifyFile(physical_name);
+            if (filetype == Loader::FileType::Unknown) {
+                LOG_WARNING(Frontend, "File %s is of indeterminate type and is possibly corrupted.", physical_name.c_str());
+                return 0;
+            }
+            if (guessed_filetype != filetype) {
+                LOG_WARNING(Frontend, "Filetype and extension of file %s do not match.", physical_name.c_str());
+            }
+
+            emit EntryReady({
+                new GameListItem(QString::fromStdString(Loader::GetFileTypeString(filetype))),
+                new GameListItemPath(QString::fromStdString(physical_name)),
+                new GameListItemSize(FileUtil::GetSize(physical_name)),
+            });
+        }
+
+        return 0; // We don't care about the found entries
+    };
+    FileUtil::ScanDirectoryTreeAndCallback(dir_path, callback);
+}
+
+void GameListWorker::run()
+{
+    stop_processing = false;
+    AddFstEntriesToGameList(dir_path.toStdString(), deep_scan);
+    emit Finished();
+}
+
+void GameListWorker::Cancel()
+{
+    disconnect(this, 0, 0, 0);
+    stop_processing = true;
+}
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -100,6 +100,23 @@ void GameList::PopulateAsync(const QString& dir_path, bool deep_scan)
     current_worker = std::move(worker);
 }
 
+void GameList::SaveInterfaceLayout(QSettings& settings)
+{
+    settings.beginGroup("UILayout");
+    settings.setValue("gameListHeaderState", tree_view->header()->saveState());
+    settings.endGroup();
+}
+
+void GameList::LoadInterfaceLayout(QSettings& settings)
+{
+    auto header = tree_view->header();
+    settings.beginGroup("UILayout");
+    header->restoreState(settings.value("gameListHeaderState").toByteArray());
+    settings.endGroup();
+
+    item_model->sort(header->sortIndicatorSection(), header->sortIndicatorOrder());
+}
+
 void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, bool deep_scan)
 {
     const auto callback = [&](const std::string& directory,
--- a/src/citra_qt/game_list.h
+++ b/src/citra_qt/game_list.h
@@ -1,0 +1,48 @@
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <QModelIndex>
+#include <QStandardItem>
+#include <QStandardItemModel>
+#include <QString>
+#include <QTreeView>
+#include <QWidget>
+
+class GameListWorker;
+
+
+class GameList : public QWidget {
+    Q_OBJECT
+
+public:
+    enum {
+        COLUMN_FILE_TYPE,
+        COLUMN_NAME,
+        COLUMN_SIZE,
+        COLUMN_COUNT, // Number of columns
+    };
+
+    GameList(QWidget* parent = nullptr);
+    ~GameList() override;
+
+    void PopulateAsync(const QString& dir_path, bool deep_scan);
+
+public slots:
+    void AddEntry(QList<QStandardItem*> entry_items);
+
+private slots:
+    void ValidateEntry(const QModelIndex& item);
+    void DonePopulating();
+
+signals:
+    void GameChosen(QString game_path);
+    void ShouldCancelWorker();
+
+private:
+    QTreeView* tree_view = nullptr;
+    QStandardItemModel* item_model = nullptr;
+    GameListWorker* current_worker = nullptr;
+};
--- a/src/citra_qt/game_list.h
+++ b/src/citra_qt/game_list.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <QModelIndex>
+#include <QSettings>
 #include <QStandardItem>
 #include <QStandardItemModel>
 #include <QString>
@@ -29,6 +30,9 @@ public:
     ~GameList() override;
 
     void PopulateAsync(const QString& dir_path, bool deep_scan);
+
+    void SaveInterfaceLayout(QSettings& settings);
+    void LoadInterfaceLayout(QSettings& settings);
 
 public slots:
     void AddEntry(QList<QStandardItem*> entry_items);
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -141,6 +141,8 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     microProfileDialog->setVisible(settings.value("microProfileDialogVisible").toBool());
     settings.endGroup();
 
+    game_list->LoadInterfaceLayout(settings);
+
     ui.action_Use_Hardware_Renderer->setChecked(Settings::values.use_hw_renderer);
     SetHardwareRendererEnabled(ui.action_Use_Hardware_Renderer->isChecked());
 
@@ -490,6 +492,7 @@ void GMainWindow::closeEvent(QCloseEvent* event) {
     settings.setValue("singleWindowMode", ui.action_Single_Window_Mode->isChecked());
     settings.setValue("displayTitleBars", ui.actionDisplay_widget_title_bars->isChecked());
     settings.setValue("firstStart", false);
+    game_list->SaveInterfaceLayout(settings);
     SaveHotkeys(settings);
 
     // Shutdown session if the emu thread is active...
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -266,6 +266,7 @@ void GMainWindow::BootGame(const std::string& filename) {
     callstackWidget->OnDebugModeEntered();
     render_window->show();
 
+    emulation_running = true;
     OnStartGame();
 }
 
@@ -294,6 +295,8 @@ void GMainWindow::ShutdownGame() {
     ui.action_Pause->setEnabled(false);
     ui.action_Stop->setEnabled(false);
     render_window->hide();
+
+    emulation_running = false;
 }
 
 void GMainWindow::StoreRecentFile(const QString& filename)
@@ -423,17 +426,21 @@ void GMainWindow::ToggleWindowMode() {
         // Render in the main window...
         render_window->BackupGeometry();
         ui.horizontalLayout->addWidget(render_window);
-        render_window->setVisible(true);
         render_window->setFocusPolicy(Qt::ClickFocus);
-        render_window->setFocus();
+        if (emulation_running) {
+            render_window->setVisible(true);
+            render_window->setFocus();
+        }
 
     } else {
         // Render in a separate window...
         ui.horizontalLayout->removeWidget(render_window);
         render_window->setParent(nullptr);
-        render_window->setVisible(true);
-        render_window->RestoreGeometry();
         render_window->setFocusPolicy(Qt::NoFocus);
+        if (emulation_running) {
+            render_window->setVisible(true);
+            render_window->RestoreGeometry();
+        }
     }
 }
 
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -167,6 +167,7 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     connect(game_list, SIGNAL(GameChosen(QString)), this, SLOT(OnGameListLoadFile(QString)));
     connect(ui.action_Load_File, SIGNAL(triggered()), this, SLOT(OnMenuLoadFile()));
     connect(ui.action_Load_Symbol_Map, SIGNAL(triggered()), this, SLOT(OnMenuLoadSymbolMap()));
+    connect(ui.action_Select_Game_List_Root, SIGNAL(triggered()), this, SLOT(OnMenuSelectGameListRoot()));
     connect(ui.action_Start, SIGNAL(triggered()), this, SLOT(OnStartGame()));
     connect(ui.action_Pause, SIGNAL(triggered()), this, SLOT(OnPauseGame()));
     connect(ui.action_Stop, SIGNAL(triggered()), this, SLOT(OnStopGame()));
@@ -377,6 +378,16 @@ void GMainWindow::OnMenuLoadSymbolMap() {
         settings.setValue("symbolsPath", QFileInfo(filename).path());
 
         LoadSymbolMap(filename.toLatin1().data());
+    }
+}
+
+void GMainWindow::OnMenuSelectGameListRoot() {
+    QSettings settings;
+
+    QString dir_path = QFileDialog::getExistingDirectory(this, tr("Select Directory"));
+    if (!dir_path.isEmpty()) {
+        settings.setValue("gameListRootDir", dir_path);
+        game_list->PopulateAsync(dir_path, settings.value("gameListDeepScan").toBool());
     }
 }
 
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -12,6 +12,7 @@
 
 #include "citra_qt/bootmanager.h"
 #include "citra_qt/config.h"
+#include "citra_qt/game_list.h"
 #include "citra_qt/hotkeys.h"
 #include "citra_qt/main.h"
 
@@ -58,6 +59,9 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
 
     render_window = new GRenderWindow(this, emu_thread.get());
     render_window->hide();
+
+    game_list = new GameList();
+    ui.horizontalLayout->addWidget(game_list);
 
     profilerWidget = new ProfilerWidget(this);
     addDockWidget(Qt::BottomDockWidgetArea, profilerWidget);
@@ -160,6 +164,7 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     UpdateRecentFiles();
 
     // Setup connections
+    connect(game_list, SIGNAL(GameChosen(QString)), this, SLOT(OnGameListLoadFile(QString)));
     connect(ui.action_Load_File, SIGNAL(triggered()), this, SLOT(OnMenuLoadFile()));
     connect(ui.action_Load_Symbol_Map, SIGNAL(triggered()), this, SLOT(OnMenuLoadSymbolMap()));
     connect(ui.action_Start, SIGNAL(triggered()), this, SLOT(OnStartGame()));
@@ -192,6 +197,8 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     setWindowTitle(window_title.c_str());
 
     show();
+
+    game_list->PopulateAsync(settings.value("gameListRootDir").toString(), settings.value("gameListDeepScan").toBool());
 
     QStringList args = QApplication::arguments();
     if (args.length() >= 2) {
@@ -264,6 +271,9 @@ void GMainWindow::BootGame(const std::string& filename) {
     // Update the GUI
     registersWidget->OnDebugModeEntered();
     callstackWidget->OnDebugModeEntered();
+    if (ui.action_Single_Window_Mode->isChecked()) {
+        game_list->hide();
+    }
     render_window->show();
 
     emulation_running = true;
@@ -295,6 +305,7 @@ void GMainWindow::ShutdownGame() {
     ui.action_Pause->setEnabled(false);
     ui.action_Stop->setEnabled(false);
     render_window->hide();
+    game_list->show();
 
     emulation_running = false;
 }
@@ -340,12 +351,16 @@ void GMainWindow::UpdateRecentFiles() {
     }
 }
 
+void GMainWindow::OnGameListLoadFile(QString game_path) {
+    BootGame(game_path.toLatin1().data());
+}
+
 void GMainWindow::OnMenuLoadFile() {
     QSettings settings;
     QString rom_path = settings.value("romsPath", QString()).toString();
 
     QString filename = QFileDialog::getOpenFileName(this, tr("Load File"), rom_path, tr("3DS executable (*.3ds *.3dsx *.elf *.axf *.cci *.cxi)"));
-    if (filename.size()) {
+    if (!filename.isEmpty()) {
         settings.setValue("romsPath", QFileInfo(filename).path());
         StoreRecentFile(filename);
 
@@ -358,7 +373,7 @@ void GMainWindow::OnMenuLoadSymbolMap() {
     QString symbol_path = settings.value("symbolsPath", QString()).toString();
 
     QString filename = QFileDialog::getOpenFileName(this, tr("Load Symbol Map"), symbol_path, tr("Symbol map (*)"));
-    if (filename.size()) {
+    if (!filename.isEmpty()) {
         settings.setValue("symbolsPath", QFileInfo(filename).path());
 
         LoadSymbolMap(filename.toLatin1().data());
@@ -440,6 +455,7 @@ void GMainWindow::ToggleWindowMode() {
         if (emulation_running) {
             render_window->setVisible(true);
             render_window->RestoreGeometry();
+            game_list->show();
         }
     }
 }
--- a/src/citra_qt/main.h
+++ b/src/citra_qt/main.h
@@ -102,6 +102,8 @@ private:
 
     GRenderWindow* render_window;
 
+    // Whether emulation is currently running in Citra.
+    bool emulation_running = false;
     std::unique_ptr<EmuThread> emu_thread;
 
     ProfilerWidget* profilerWidget;
--- a/src/citra_qt/main.h
+++ b/src/citra_qt/main.h
@@ -92,6 +92,8 @@ private slots:
     void OnGameListLoadFile(QString game_path);
     void OnMenuLoadFile();
     void OnMenuLoadSymbolMap();
+    /// Called whenever a user selects the "File->Select Game List Root" menu item
+    void OnMenuSelectGameListRoot();
     void OnMenuRecentFile();
     void OnOpenHotkeysDialog();
     void OnConfigure();
--- a/src/citra_qt/main.h
+++ b/src/citra_qt/main.h
@@ -10,6 +10,7 @@
 
 #include "ui_main.h"
 
+class GameList;
 class GImageInfo;
 class GRenderWindow;
 class EmuThread;
@@ -87,6 +88,8 @@ private slots:
     void OnStartGame();
     void OnPauseGame();
     void OnStopGame();
+    /// Called whenever a user selects a game in the game list widget.
+    void OnGameListLoadFile(QString game_path);
     void OnMenuLoadFile();
     void OnMenuLoadSymbolMap();
     void OnMenuRecentFile();
@@ -101,6 +104,7 @@ private:
     Ui::MainWindow ui;
 
     GRenderWindow* render_window;
+    GameList* game_list;
 
     // Whether emulation is currently running in Citra.
     bool emulation_running = false;
--- a/src/citra_qt/main.ui
+++ b/src/citra_qt/main.ui
@@ -45,7 +45,7 @@
      <x>0</x>
      <y>0</y>
      <width>1081</width>
-     <height>21</height>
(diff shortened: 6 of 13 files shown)


src/core/loader/loader.h
```diff
@@ -33,6 +33,23 @@ enum class FileType {
     THREEDSX, //3DSX
 };
 
+/**
+ * Identifies the type of a bootable file
```
neobrain: based on what?

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,114 @@
+#include "game_list.h"
```
neobrain: license header

neobrain: this should be included after standard headers

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include <QThread>
+#include <QTreeWidgetItem>
+
+#include "core/loader/loader.h"
+
+#include "common/common_paths.h"
+#include "common/logging/log.h"
+#include "common/string_util.h"
+
+GameList::GameList(QWidget* parent)
```
neobrain: def constructor is not called

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,114 @@
...
+    setEditTriggers(NoEditTriggers);
+
+    item_model->insertColumns(0, COLUMN_COUNT);
+    item_model->setHeaderData(COLUMN_FILE_TYPE, Qt::Horizontal, "File type");
+    item_model->setHeaderData(COLUMN_NAME, Qt::Horizontal, "Name");
+    item_model->setHeaderData(COLUMN_SIZE, Qt::Horizontal, "Size");
+
+    connect(this, SIGNAL(activated(const QModelIndex&)), this, SLOT(ValidateEntry(const QModelIndex&)));
+
+    // Workaround for not being able to use QList in signals/slots
```
neobrain: Explain the issue and the workaround instead of mentioning the workaround and expecting people to be able to read your mind.

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,114 @@
...
+    qRegisterMetaType<StandardItemList>("StandardItemList");
+}
+
+void GameList::AddEntry(StandardItemList entry_items)
```
neobrain: Pass by reference instead?

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+    QThread* thread = new QThread;
+    GameListWorker* worker = new GameListWorker(dir_path, deep_scan);
+    worker->moveToThread(thread);
+    connect(thread, SIGNAL(started()), worker, SLOT(BeginProcessing()));
+    connect(worker, SIGNAL(EntryReady(StandardItemList)), this, SLOT(AddEntry(StandardItemList)));
+    connect(worker, SIGNAL(finished()), thread, SLOT(quit()));
+    connect(worker, SIGNAL(finished()), worker, SLOT(deleteLater()));
+    connect(thread, SIGNAL(finished()), thread, SLOT(deleteLater()));
+    thread->start();
```
neobrain: Did you make sure there are no race conditions with regards to starting citra and closing it while populating the game list?

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    if (file_path.isEmpty())
+        return;
+    std::string std_file_path = file_path.toStdString();
+    if (!FileUtil::Exists(std_file_path) || FileUtil::IsDirectory(std_file_path))
+        return;
+    emit GameChosen(file_path);
+}
+
+void GameList::PopulateGameListAsync(const QString& dir_path, bool deep_scan)
```
neobrain: I suppose while this happens, the game list should be disabled.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+#include <QStandardItem>
+#include <QStandardItemModel>
+#include <QString>
+#include <QTreeView>
+#include <QTreeWidgetItem>
+
+#include "common/file_util.h"
+#include "common/string_util.h"
+
+enum : int {
```
neobrain: Why the hardcoded underlying type?

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+    COLUMN_FILE_TYPE,
+    COLUMN_NAME,
+    COLUMN_SIZE,
```
neobrain: This should be an enum class instead of using prefixes.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+    COLUMN_COUNT, // Number of columns
+};
+
+class GameListItem: public QStandardItem {
+public:
+    GameListItem():
```
neobrain: Why the odd coding style?

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+#include "common/string_util.h"
+
+enum : int {
+    COLUMN_FILE_TYPE,
+    COLUMN_NAME,
+    COLUMN_SIZE,
+    COLUMN_COUNT, // Number of columns
+};
+
+class GameListItem: public QStandardItem {
```
neobrain: space before :

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+public:
+    GameListItem():
+        QStandardItem() {}
+    GameListItem(const QString& string):
+        QStandardItem(string) {}
+};
+
+class GameListItemPath: public GameListItem {
```
neobrain: docstrings

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+            Common::SplitPath(value.toString().toStdString(), nullptr, &filename, nullptr);
+            GameListItem::setData(QString::fromStdString(filename), Qt::DisplayRole);
+            GameListItem::setData(value, FullPathRole);
+        } else {
+            GameListItem::setData(value, role);
+        }
+    }
+};
+
+class GameListItemSize: public GameListItem {
```
neobrain: docstrings

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+        if (role == SizeRole) {
+            qulonglong size_bytes = value.toULongLong();
+            GameListItem::setData(QString::fromStdString(Common::ReadableByteSize(size_bytes)), Qt::DisplayRole);
+            GameListItem::setData(value, SizeRole);
+        } else {
+            GameListItem::setData(value, role);
+        }
+    }
+
+    virtual bool operator<(const QStandardItem& other) const override {
```
neobrain: why is this needed? => docstring

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+        return data(SizeRole).toULongLong() < other.data(SizeRole).toULongLong();
+    }
+};
+
+
+// Workaround for not being able to use QList in signals/slots
```
neobrain: workarounds shouldn't just be mentioned, but actually be explained.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+private slots:
+    void ValidateEntry(const QModelIndex& item);
+signals:
+    void GameChosen(QString game_path);
+private:
+    QStandardItemModel* item_model;
+};
+
+
+class GameListWorker: public QObject {
```
neobrain: docstring. Also, consider moving this kind of stuff to a private header.

src/citra_qt/main.cpp
```diff
@@ -104,6 +105,9 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     addDockWidget(Qt::RightDockWidgetArea, graphicsTracingWidget);
     graphicsTracingWidget->hide();
 
+    auto gameList = new GameList(this);
```
neobrain: Don't explicitly denote parents for widgets which get implicitly reparented within the same scope anyway.

src/citra_qt/main.cpp
```diff
@@ -143,7 +147,16 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     SetShaderJITEnabled(ui.action_Use_Shader_JIT->isChecked());
 
     ui.action_Single_Window_Mode->setChecked(settings.value("singleWindowMode", true).toBool());
-    ToggleWindowMode();
+    if (ui.action_Single_Window_Mode->isChecked()) {
+        // Render in the main window...
```
neobrain: What are these comments trying to tell me? They're just repeating the branch condition and hence serve no purpose.

Also, please explain what you're actually doing in each branch, and why.

src/citra_qt/main.cpp
```diff
@@ -192,6 +206,8 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
 
     show();
 
+    gameList->PopulateGameListAsync(settings.value("gameListRootDir").toString(), settings.value("gameListDeepScan").toBool());
```
neobrain: Use QSettings instead.

src/citra_qt/main.cpp
```diff
@@ -331,12 +347,16 @@ void GMainWindow::UpdateRecentFiles() {
     }
 }
 
+void GMainWindow::OnGameListLoadFile(QString game_path) {
```
neobrain: const& parameter ?

src/citra_qt/main.h
```diff
@@ -87,6 +87,7 @@ private slots:
     void OnStartGame();
     void OnPauseGame();
     void OnStopGame();
+    void OnGameListLoadFile(QString game_path);
```
neobrain: docstring (yes, none of the other slots have one, but gotta start somewhere.. )

src/common/file_util.h
```diff
@@ -96,6 +97,9 @@ bool Copy(const std::string &srcFilename, const std::string &destFilename);
 // creates an empty file filename, returns true on success
 bool CreateEmptyFile(const std::string &filename);
 
+// Internal function for use by templated ScanDirectoryTreeAndCallback
```
neobrain: If it's only internal, why expose it as part of the interface?

neobrain: ScanDirectoryTreeAndCallback isn't templated, unless you meant something else than it being a template function.

src/common/file_util.h
```diff
@@ -96,6 +97,9 @@ bool Copy(const std::string &srcFilename, const std::string &destFilename);
...
+u32 ScanDirectoryTreeAndCallback(const std::string &directory, std::function<u32(const std::string&, const std::string&)> callback);
+
 // Scans the directory tree gets, starting from _Directory and adds the
```
neobrain: There is no `_Directiory`

`Scans the directory tree gets` doesn't seem to be valid grammar.

Use @return instead of adding "returns..." sentences.

src/core/loader/loader.h
```diff
@@ -33,6 +33,23 @@ enum class FileType {
...
+ * @param file open file
```
neobrain: `@param file_name` is not explained.

src/core/loader/loader.h
```diff
@@ -33,6 +33,23 @@ enum class FileType {
...
+ * @return FileType of file
+ */
+FileType IdentifyFile(FileUtil::IOFile& file);
+FileType IdentifyFile(const std::string& file_name);
+
+/**
+ * Guess the type of a bootable file from its extension
+ * @param extension String extension of bootable file
+ * @return FileType of file
```
neobrain: If this may return FileType::Error,  it should be denoted as part of the interface description.

src/core/loader/loader.h
```diff
@@ -33,6 +33,23 @@ enum class FileType {
     THREEDSX, //3DSX
 };
 
+/**
+ * Identifies the type of a bootable file
+ * @param file open file
+ * @return FileType of file
```
neobrain: If this may return FileType::Error, it should be denoted as part of the interface description.

src/core/loader/loader.h
```diff
@@ -33,6 +33,23 @@ enum class FileType {
...
+FileType IdentifyFile(const std::string& file_name);
+
+/**
+ * Guess the type of a bootable file from its extension
+ * @param extension String extension of bootable file
+ * @return FileType of file
+ */
+FileType GuessFromExtension(const std::string& extension_);
+
+const char* GetFileTypeString(FileType type);
```
neobrain: docstring

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,114 @@
...
+    dir_path(dir_path), deep_scan(deep_scan) {}
+
+void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, bool deep_scan)
+{
+    const static auto callback = [&](const std::string& directory,
+                                     const std::string& virtualName) -> u32 {
+
+        std::string physicalName = directory + DIR_SEP + virtualName;
+
+        if (FileUtil::IsDirectory(physicalName.c_str())) {
```
lioncash: c_str isn't necessary here.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,116 @@
...
+
+    GameListItemSize():
+        GameListItem() {}
+    GameListItemSize(const qulonglong size_bytes):
+        GameListItem()
+    {
+        setData(size_bytes, SizeRole);
+    }
+
+    virtual void setData(const QVariant& value, int role = Qt::UserRole + 1) override {
```
lioncash: virtual isn't necessary with override specified.

src/common/string_util.cpp
```diff
@@ -35,6 +36,14 @@ std::string ToUpper(std::string str) {
     return str;
 }
 
+std::string ReadableByteSize(unsigned long long size) {
+    if (size == 0)
+        return "0";
+    const char* units[] { "B", "kiB", "MiB", "GiB", "TiB" };
```
lioncash: this can be static

src/core/loader/loader.h
archshift: This does not ever return `FileType::Error`.

src/core/loader/loader.h
archshift: This does not ever return `FileType::Error`.

src/common/string_util.cpp
```diff
@@ -35,6 +36,14 @@ std::string ToUpper(std::string str) {
...
+    int digit_groups = (int)(std::log10(size) / std::log10(1024));
+    return StringFromFormat("%.1f %s", size / std::pow(1024, digit_groups), units[digit_groups]);
```
archshift: This has the potential to read out-of-bounds from the `units` array.

src/citra_qt/game_list.cpp
archshift: Since this slot is meant to be used as a complement to the `GameListWorker::EntryReady` signal, that would require `emit`ting a reference, which could possibly be dangerous (especially in this multithreaded scenario).

src/citra_qt/game_list.cpp
archshift: Added a note to my to-do list that I would like some kind of loading animation or designation somewhere on the GUI as it is populating.

src/citra_qt/game_list.h
archshift: This is supposed to be used as a replacement for manually specifying a column `int` to various Qt functions.

src/citra_qt/game_list.h
archshift: Because this is supposed to be used as a replacement for manually specifying a column `int` to various Qt functions, and `enum class`es have to be manually casted to their underlying type, I decided against using them.

Any suggestions?

src/citra_qt/main.cpp
archshift: This is using `QSettings`...

src/citra_qt/main.cpp
archshift: This is a slot, which would mean its corresponding signal would need to `emit` a reference, which I don't think is necessarily a good idea.

yuriks: `QString`s are COW anyway, so passing them by-val is cheap.

src/citra_qt/game_list.cpp
archshift: For the record, both comments have been addressed.

archshift:
That latest fixup *should* have made things (as far as I can tell) thread-safe. I'd welcome some proofreading on that to make sure I didn't make any dumb mistakes though.

src/core/loader/loader.h
neobrain: Sorry for being unclear. That should be documented too, since it's highly relevant for error checking.

src/citra_qt/game_list.h
neobrain: You can provide a casting operator to int instead. Still ugly, but at least it doesn't pollute the global namespace with such generic names.

src/citra_qt/main.cpp
neobrain: Derp, you're right :)

src/citra_qt/game_list.h
neobrain: The underlying type doesn't matter for that, though. Any enum will always be castable to int, regardless of underlying type.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,157 @@
...
+public slots:
+    void AddEntry(StandardItemList entry_items);
+private slots:
+    void ValidateEntry(const QModelIndex& item);
+signals:
+    void GameChosen(QString game_path);
+    void StopWorker();
+private:
+    QStandardItemModel* item_model;
+    QThread* worker_thread;
```
neobrain: `unique_ptr` ?

src/citra_qt/game_list.cpp
neobrain: For the record, we'll still want to disable the game list even with a progress indicator, since randomly popping up items would make it impossible to properly select a particular game anyway.

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, bool deep_scan)
+{
+    const static auto callback = [&](const std::string& directory,
+                                     const std::string& virtualName) -> int {
+
+        std::string physicalName = directory + DIR_SEP + virtualName;
+
+        if (stop_processing)
+            return -1; // A negative return value breaks the callback loop.
```
neobrain: Sounds like a boolean return value is perfectly sufficient here?

neobrain: .. nevermind, we have to follow the ScanDirectoryTreeAndCallback interface.

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,148 @@
...
+    worker_thread->start();
+}
+
+GameListWorker::GameListWorker(const QString& dir_path, bool deep_scan):
+    QObject(), dir_path(dir_path), deep_scan(deep_scan) {}
+
+void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, bool deep_scan)
+{
+    const static auto callback = [&](const std::string& directory,
+                                     const std::string& virtualName) -> int {
```
neobrain: virtual_name

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+        std::string physicalName = directory + DIR_SEP + virtualName;
```
neobrain: physical_name

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,157 @@
...
+enum : int {
+    COLUMN_FILE_TYPE,
+    COLUMN_NAME,
+    COLUMN_SIZE,
+    COLUMN_COUNT, // Number of columns
+};
+
+class GameListItem : public QStandardItem {
+public:
+    GameListItem(): QStandardItem() {}
```
neobrain: space before :

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,157 @@
...
+    GameListItem(const QString& string): QStandardItem(string) {}
+};
+
+/**
+ * A specialization of GameListItem for path values.
+ * This class ensures that for every full path value it holds, a correct string representation
+ * of just the filename (with no extension) will be displayed to the user.
+ */
+class GameListItemPath : public GameListItem {
```
neobrain: All of these special-tailored classes look like they are more effort than it really should have been. Have you considered using a custom item model instead of trying to fit into QStandardItem ?

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,157 @@
...
+public:
+    static const int FullPathRole = Qt::UserRole + 1;
+
+    GameListItemPath(): GameListItem() {}
+    GameListItemPath(const QString& game_path): GameListItem()
+    {
+        setData(game_path, FullPathRole);
+    }
+
+    void setData(const QVariant& value, int role = Qt::UserRole + 1) override
```
neobrain: Why use UserRole+1 as the default argument here? (either you want to explicitly set it to FullPathRole, or you don't want to set it at all)

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,157 @@
...
+typedef QList<QStandardItem*> StandardItemList;
+
+class GameList : public QTreeView {
+    Q_OBJECT
+public:
+    GameList(QWidget* parent = nullptr);
+    ~GameList() override;
+
+    void PopulateGameListAsync(const QString& dir_path, bool deep_scan);
+public slots:
```
neobrain: blank lines before access specifiers.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,157 @@
...
+signals:
+    /**
+     * The `EntryReady` signal is emitted once an entry has been prepared and is ready
+     * to be added to the game list.
+     * @param entry_items a list with `QStandardItem`s that make up the columns of the new entry.
+     */
+    void EntryReady(StandardItemList entry_items);
+
+    /// The `finished` signal is emitted once the worker has finished all processing.
+    void finished();
```
neobrain: Why lowercase?

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,148 @@
...
+    worker->moveToThread(worker_thread);
+    connect(worker_thread, SIGNAL(started()), worker, SLOT(BeginProcessing()));
+    connect(this, SIGNAL(StopWorker()), worker, SLOT(StopProcessing()), Qt::DirectConnection);
+    connect(worker, SIGNAL(EntryReady(StandardItemList)), this, SLOT(AddEntry(StandardItemList)));
+    connect(worker, SIGNAL(finished()), worker_thread, SLOT(quit()));
+    connect(worker, SIGNAL(finished()), worker, SLOT(deleteLater()));
+    worker_thread->start();
+}
+
+GameListWorker::GameListWorker(const QString& dir_path, bool deep_scan):
```
neobrain: I don't think there is a need to explicitly call the parent constructor without arguments, is there?

src/common/file_util.cpp
```diff
@@ -490,6 +470,35 @@ u32 ScanDirectoryTree(const std::string &directory, FSTEntry& parentEntry)
     return foundEntries;
 }
 
+// Scans the directory tree gets, starting from _Directory and adds the
```
neobrain: directory

src/citra_qt/game_list.cpp
archshift: That's not an issue anyway, new items are appended to the bottom.

archshift:
Pushed new changes.

src/citra_qt/game_list.cpp
neobrain: We almost certainly will want some sort of sorting feature in the future, at which point it will become an issue which we could easily address at this stage already.

archshift: Okay, I've disabled the game list while it's populating. I still haven't figured out a good way to overlay some kind of loading animation or a place to stick some "Loading game list..." text.

src/citra_qt/main.cpp
```diff
@@ -292,6 +310,7 @@ void GMainWindow::ShutdownGame() {
     ui.action_Pause->setEnabled(false);
     ui.action_Stop->setEnabled(false);
     render_window->hide();
+    game_list->show();
```
LittleWhite-tb: I would put this in a test to let it appear when the multiple window mode is enabled. Thus, the game list would be still displayed in the main window, even after the emulation has started.

src/citra_qt/main.cpp
```diff
@@ -143,7 +147,17 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     SetShaderJITEnabled(ui.action_Use_Shader_JIT->isChecked());
 
     ui.action_Single_Window_Mode->setChecked(settings.value("singleWindowMode", true).toBool());
-    ToggleWindowMode();
+    // Similar to ToggleWindowMode, but here we set up the render widget without showing or focusing it.
```
neobrain: pointless reference: "This works like TWM, but it's different, it actually does..." ... if anything, put the reference at the back of the comment:

"// Set up the render widget (without showing or focussing it[ - contrary to ToggleWindowMode])"

src/citra_qt/main.cpp
archshift: Indeed :)

archshift:
A loading overlay on the game list widget would add a good deal of complexity for too little payoff, in my opinion. I think it might be better left to a future PR, perhaps placing the info on a status bar at the bottom of the window.

src/citra_qt/game_list.h
archshift: You can't provide casting operators to `enum class` because casting operators can only be provided within the type they apply to. You can't define any method within an `enum class`.

src/citra_qt/game_list.h
archshift: I figured trying to create my own item model from `QAbstractItemModel` would involve a lot of reimplementation of what is available with `QStandardItemModel`, with little need since in theory all I'm trying to do is display a simple tree view.

While certain parts of this code could be made somewhat more elegant that way, I didn't think it was worth the tradeoff. I could be wrong, though.

src/common/string_util.cpp
```diff
@@ -35,6 +37,16 @@ std::string ToUpper(std::string str) {
     return str;
 }
 
+std::string ReadableByteSize(unsigned long long size) {
+    static const std::array<const char*, 6> units = { "B", "kiB", "MiB", "GiB", "TiB", "PiB" };
```
linkmauve: KiB, not kiB, it’s an uppercase K unlike in base-1000 units where it’s a lowercase k.

archshift:
I think this is ready for review now.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,177 @@
...
+     * Override it so that it will correctly sort by numerical value instead of by string representation.
+     */
+    bool operator<(const QStandardItem& other) const override
+    {
+        return data(SizeRole).toULongLong() < other.data(SizeRole).toULongLong();
+    }
+};
+
+
+class GameList : public QTreeView {
```
yuriks: Inheritance for reuse. *belch* A GameList isn't substitutible by a TreeView and so you shouldn't inherit from it.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,177 @@
...
+    QThread* worker_thread;
+    GameListWorker* worker;
+};
+
+
+/**
+ * Asynchronous worker object for populating the game list.
+ * Communicates with other threads through Qt's signal/slot system.
+ */
+class GameListWorker : public QObject {
```
yuriks: This class can be private to the .cpp file.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,177 @@
...
+    }
+};
+
+
+/**
+ * A specialization of GameListItem for size values.
+ * This class ensures that for every numerical size value it holds (in bytes), a correct
+ * human-readable string representation will be displayed to the user.
+ */
+class GameListItemSize : public GameListItem {
```
yuriks: This class can be private to the .cpp file.

src/citra_qt/game_list.h
yuriks: This class can be private to the .cpp file.

src/citra_qt/game_list.h
```diff
@@ -0,0 +1,177 @@
...
+
+enum {
+    COLUMN_FILE_TYPE,
+    COLUMN_NAME,
+    COLUMN_SIZE,
+    COLUMN_COUNT, // Number of columns
+};
+
+
+class GameListItem : public QStandardItem {
```
yuriks: This class can be private to the .cpp file.

src/citra_qt/game_list.h
archshift: Could you clarify? Should I instead just subclass QWidget and have a
QTreeView member?

src/citra_qt/game_list.h
archshift: The Qt MOC only parses header files, so any Q_OBJECT class definitions need
to be in headers.

src/citra_qt/game_list.h
archshift: Ditto.

src/citra_qt/game_list.h
archshift: Ditto.

src/citra_qt/game_list.h
archshift: Ditto.

src/citra_qt/game_list.h
yuriks: Then create a separate header file for these (as done by some of the debugging widgets, which have headers suffixed with `_p`.)

src/citra_qt/CMakeLists.txt
```diff
@@ -42,6 +43,7 @@ set(HEADERS
             debugger/profiler.h
             debugger/ramview.h
             debugger/registers.h
+            game_list.h
```
yuriks: You forgot to add game_list_p.h

src/common/file_util.h
```diff
@@ -96,9 +97,28 @@ bool Copy(const std::string &srcFilename, const std::string &destFilename);
...
+ * Scans the directory tree, calling the callback for each file/directory found.
+ * @param directory the parent directory to start scanning from
+ * @param callback The callback which will be called for each file/directory. It is called
+ * with the arguments (const std::string& directory, const std::string& virtualName).
+ * The `directory `parameter is the path to the directory which contains the file/directory, and
+ * the `virtualName` parameter is the incomplete file path, without any directory info.
+ * The callback must return the number of files and directories which the provided path contains.
+ * If the callback's return value is -1, the callback loop is broken immediately. If the callback's
+ * return value is otherwise negative, the callback loop is broken immediately and the callback's
+ * return value is returned from this function (to allow for error handling).
```
yuriks: You should indent the body of the parameter description, or otherwise separate it from neighboring entries. However, I think this is too long for a parameter description, and should be moved to the main body of the function documentation instead.

src/common/string_util.cpp
```diff
@@ -35,6 +37,16 @@ std::string ToUpper(std::string str) {
     return str;
 }
 
+std::string ReadableByteSize(unsigned long long size) {
+    static const std::array<const char*, 6> units = { "B", "KiB", "MiB", "GiB", "TiB", "PiB" };
+
+    if (size == 0)
+        return "0";
+    int digit_groups = (int)(std::log10(size) / std::log10(1024));
+    const char* unit = digit_groups < units.size() ? units[digit_groups] : "0";
```
yuriks: Wat. If the value exceeds the largest unit just stop subdividing and return, e.g., "10000 PiB".

src/common/string_util.cpp
```diff
@@ -35,6 +37,16 @@ std::string ToUpper(std::string str) {
     return str;
 }
 
+std::string ReadableByteSize(unsigned long long size) {
```
yuriks: IMO this function should be in citra-qt so it can use Qt formatting tools to achieve proper internationalization.

src/common/string_util.h
```diff
@@ -21,6 +21,9 @@ std::string ToLower(std::string str);
 /// Make a string uppercase
 std::string ToUpper(std::string str);
 
+/// Convert a size in bytes into a readable format (kiB, MiB, etc.)
```
yuriks: "KiB"*

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    connect(worker_thread, SIGNAL(started()), worker, SLOT(BeginProcessing()), Qt::QueuedConnection);
+    connect(worker, SIGNAL(EntryReady(QList<QStandardItem*>)), this, SLOT(AddEntry(QList<QStandardItem*>)), Qt::QueuedConnection);
+    connect(worker, SIGNAL(FinishedProcessing()), this, SLOT(DonePopulating()), Qt::QueuedConnection);
+}
+
+GameList::~GameList()
+{
+    worker->StopProcessing();
+    worker_thread->quit();
```
yuriks: What's the point of calling `quit()`? (You're doing this in other places too.)

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,174 @@
...
+    item_model->setHeaderData(COLUMN_NAME, Qt::Horizontal, "Name");
+    item_model->setHeaderData(COLUMN_SIZE, Qt::Horizontal, "Size");
+
+    connect(this, SIGNAL(activated(const QModelIndex&)), this, SLOT(ValidateEntry(const QModelIndex&)));
+
+    // We must register all custom types with the Qt Automoc system so that we are able to use it with
+    // signals/slots. In this case, QList falls under the umbrells of custom types.
+    qRegisterMetaType<QList<QStandardItem*>>("QList<QStandardItem*>");
+
+    worker_thread = new QThread;
```
yuriks: Instead of creating a dedicated QThread, I think you can make GameListWorker into a [QRunnable](http://doc.qt.io/qt-5/qrunnable.html) and launch it using [QThreadPool](http://doc.qt.io/qt-5/qthreadpool.html).

archshift: I see that `QThreadPool` is designed to recycle individual `QThread`s. Since I already only use a single `QThread` per `GameList` instance, is there much of a point to using `QThreadPool` instead?

yuriks: QThreadPool recycles threads across the *entire application*. Since you only use the thread to do a (relatively) short-lived operation, it makes sense to use the thread pool to just grab a thread that's lying around if you can avoid spawning a new one. This makes it play nicely with any other concurrent operations added to the UI in the future that also use the thread pool.

yuriks: It also has the additional benefit of saving you from having to manage the lifetime of the thread yourself.

src/citra_qt/main.ui
```diff
@@ -182,6 +183,14 @@
     <string>Display Dock Widget Headers</string>
    </property>
   </action>
+  <action name="action_Select_Folder">
+   <property name="text">
+    <string>Select Folder...</string>
```
yuriks: "Select game list root..." or something, the current name has no context indicating it's talking about the game list.

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,174 @@
...
+    item_model->removeRows(0, item_model->rowCount());
+
+    if (worker_thread->isRunning()) {
+        worker->StopProcessing();
+        worker_thread->quit();
+        worker_thread->wait();
+    }
+
+    QMetaObject::invokeMethod(worker, "SetData", Qt::QueuedConnection,
+                              Q_ARG(QString, dir_path), Q_ARG(bool, deep_scan));
```
yuriks: Why do you need to use this weird method of setting the task parameters?

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,174 @@
...
+            AddFstEntriesToGameList(physical_name, true);
+        } else {
+            std::string filename_filename, filename_extension;
+            Common::SplitPath(physical_name, nullptr, &filename_filename, &filename_extension);
+
+            Loader::FileType guessed_filetype = Loader::GuessFromExtension(filename_extension);
+            if (guessed_filetype == Loader::FileType::Unknown)
+                return 0;
+            Loader::FileType filetype = Loader::IdentifyFile(physical_name);
+            if (filetype == Loader::FileType::Unknown || guessed_filetype != filetype) {
```
yuriks: I get skipping the file entirely if the guessed type is unknown (since you can then skip over non-game files) but why do you consider it corrupted if the guessed type differs from the identified type? This causes .cxi files incorrectly named as .cci, etc. to be skipped even though we know what type they are and you can load them fine via the menu.

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,174 @@
...
+                LOG_WARNING(Frontend, "File %s is of indeterminate type and is possibly corrupted.", physical_name.c_str());
+                return 0;
+            }
+
+            QList<QStandardItem*> row_entry {
+                new GameListItem(QString::fromStdString(Loader::GetFileTypeString(filetype))),
+                new GameListItemPath(QString::fromStdString(physical_name)),
+                new GameListItemSize(FileUtil::GetSize(physical_name)),
+            };
+            emit EntryReady(row_entry);
```
yuriks: It feels unnecessary to me to update the list in real-time, since you're going to keep it disabled anyway. I think you should just return a list of found games upon completion of the task, and otherwise just give periodic progress updates. (This also enables you to add a fake item to the view to serve as a loading indicator.)

src/citra_qt/main.cpp
```diff
@@ -61,6 +65,9 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     render_window = new GRenderWindow(this, emu_thread.get());
     render_window->hide();
 
+    game_list = new GameList(this);
```
neobrain: Don't explicitly set a parent for widgets which are implicitly reparented (to the same parent) by addWidget anyway.

src/citra_qt/CMakeLists.txt
archshift: Added it.

src/citra_qt/game_list.cpp
archshift: In a followup PR, I plan to add a popup window like Dolphin's when the game list is populating. I personally like the realtime updating, and it'll be an indicator to the user that it's still loading (as well as the fact that it's disabled).

archshift:
I've addressed all comments.

src/common/file_util.h
```diff
@@ -96,9 +97,28 @@ bool Copy(const std::string &srcFilename, const std::string &destFilename);
...
+ */
+int ScanDirectoryTreeAndCallback(const std::string &directory, std::function<int(const std::string&, const std::string&)> callback);
+
+/**
+ * Scans the directory tree, storing the results.
+ * @param directory the parent directory to start scanning from
+ * @param parentEntry FSTEntry where the filesystem tree results will be stored.
+ * @return the total number of files/directories found
+ */
+int ScanDirectoryTree(const std::string &directory, FSTEntry& parentEntry);
```
bunnei: Should be `parent_entry`, not `parentEntry`

src/common/file_util.h
```diff
@@ -96,9 +97,28 @@ bool Copy(const std::string &srcFilename, const std::string &destFilename);
...
-u32 ScanDirectoryTree(const std::string &directory, FSTEntry& parentEntry);
+/**
+ * Scans the directory tree, calling the callback for each file/directory found.
+ * The callback must return the number of files and directories which the provided path contains.
+ * If the callback's return value is -1, the callback loop is broken immediately.
+ * If the callback's return value is otherwise negative, the callback loop is broken immediately
+ * and the callback's return value is returned from this function (to allow for error handling).
+ * @param directory the parent directory to start scanning from
+ * @param callback The callback which will be called for each file/directory. It is called
+ *     with the arguments (const std::string& directory, const std::string& virtualName).
```
bunnei: Should be `virtual_name`, not `virtualName`

src/common/file_util.cpp
```diff
@@ -420,27 +420,22 @@ bool CreateEmptyFile(const std::string &filename)
...
 
-// Scans the directory tree gets, starting from _Directory and adds the
-// results into parentEntry. Returns the number of files+directories found
-u32 ScanDirectoryTree(const std::string &directory, FSTEntry& parentEntry)
+int ScanDirectoryTreeAndCallback(const std::string &directory, std::function<int(const std::string&, const std::string&)> callback)
 {
     LOG_TRACE(Common_Filesystem, "directory %s", directory.c_str());
     // How many files + directories we found
-    u32 foundEntries = 0;
+    int foundEntries = 0;
```
bunnei: You're refactoring this anyway, might as well fix the naming. Should be `found_entries`, not `foundEntries`.

src/common/file_util.cpp
```diff
@@ -420,27 +420,22 @@ bool CreateEmptyFile(const std::string &filename)
...
 #ifdef _WIN32
     // Find the first file in the directory.
     WIN32_FIND_DATA ffd;
 
     HANDLE hFind = FindFirstFile(Common::UTF8ToTStr(directory + "\\*").c_str(), &ffd);
```
bunnei: Same deal with naming.. `handle`?

src/citra_qt/main.cpp
```diff
@@ -47,6 +48,9 @@
 
 #include "video_core/video_core.h"
 
+// Whether emulation is currently running in Citra.
+static bool emulation_running = false;
```
bunnei: Wow, do we really not have something that serves this purpose already?

src/citra_qt/game_list.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+        if (deep_scan && FileUtil::IsDirectory(physical_name)) {
+            AddFstEntriesToGameList(physical_name, true);
+        } else {
+            std::string filename_filename, filename_extension;
+            Common::SplitPath(physical_name, nullptr, &filename_filename, &filename_extension);
+
+            Loader::FileType guessed_filetype = Loader::GuessFromExtension(filename_extension);
+            if (guessed_filetype == Loader::FileType::Unknown)
+                return 0;
```
bunnei: Should this log a warning?

bunnei:
Couple of nits, but looks good. Tested this on Windows, and it works well.
I'd like to get this merged within the next few days, so I'd say this is a final call to the team to review this if you haven't already.

src/common/file_util.cpp
archshift: HANDLE is a Win32 type name..

src/citra_qt/game_list.cpp
archshift: Nope, it's fine if the user keeps other types of files in the same folder.

src/common/file_util.cpp
bunnei: Yes, what I am saying is `hFind` is not to our naming convention, so if you're changing `foundEntries` might as well refactor that one too.

src/citra_qt/main.cpp
archshift: Not as far as I know!

neobrain: Do not introduce global(-ish) state into the GUI code. Please find a way to have this status flag implemented as a class member instead, or better yet don't use it at all and localize the functionality to EmuStarted/EmuStopped events instead if possible.

bunnei:
:cake:

PR #1102, opened 2015-09-03
Service/CAM: Dummy implementation of some functions

mailwl:


RavenHome1:
dose that mean that citra will support the 3ds camera functionality?

mailwl:
@RavenHome1 
no, it's just to passthrough the Fatal Error in game "Heavy Fire: The Chosen Few 3D"
it even doesn't provide any data to game, just garbage, already in buffer.
I don't think, the full functionality of camera is needed for emulation...

RavenHome1:
not sure i have not seen a game that full rely on the camera, however some parts of a certain games do require the camera to be played

mailwl:
may you name some games? just to test if the PR not broke something

RavenHome1:
as i have seen on youtube part of pokemon require you to put on a silly face for the cam in order to play with the pokemons not sure if others also the gome dose not boot atm so it is hard to test it

RavenHome1:
oh wait here is a game Spirit Camera: The Cursed Memoir

yuriks:
Seems like this unfortunately isn't enough to get Bravely Default past the AR intro yet.

mailwl:
@yuriks, yes, you are right, dummy implementation is unsuitable for AR-powered games.
but i can't find link between cam and y2r buffers, to feed a game good picture

yuriks:
@mailwl Once you *do* get the camera on, the game allows you to hit Y to skip the AR part, so it shouldn't be necessary to actually hook the camera up. Unfortunately it seems like it just deadlocks.

mailwl:
maybe citra owners merges my PR in this variant, because the game have strange graphics bugs, need to be patched?

src/core/hle/service/cam/cam.cpp
```diff
@@ -18,6 +18,219 @@
 namespace Service {
 namespace CAM {
 
+static const u32 TRANSFER_BYTES = 5 * 1024;
```
bunnei: Needs a comment explaining why this value was chosen

src/core/hle/service/cam/cam.cpp
```diff
@@ -18,6 +18,219 @@
...
+
+static Kernel::SharedPtr<Kernel::Event> completion_event;
+static Kernel::SharedPtr<Kernel::Event> interrupt_error_event;
+static Kernel::SharedPtr<Kernel::Event> vsync_interrupt_error_event;
+
+void DriverInitialize(Service::Interface* self) {
```
bunnei: This function (and the below ones) need docstrings explaining the known information - e.g. purpose, inputs/outputs, and any stubbed functionality. See https://github.com/citra-emu/citra/blob/master/src/core/hle/service/gsp_gpu.cpp#L259 for an example of what we usually do.

bunnei:
@mailwl can you add the above documentation so that we can get this merged? Thanks.

mailwl:
yes, i am in process, 3dbrew describes many enums, i want to add this info into sources, too

src/core/hle/service/cam/cam.h
```diff
@@ -10,6 +10,25 @@
 namespace Service {
 namespace CAM {
 
+void DriverInitialize(Service::Interface* self);
```
purpasmart96: There should be docstrings for these functions in the header as bunnei pointed out

mailwl:
may you help me? i don't know, how to fix appveyor errors

src/core/hle/service/cam/cam.h
```diff
@@ -6,10 +6,167 @@
...
+    LENS_CORRECTION_OFF    = 0,
+    LENS_CORRECTION_ON_70  = 1,
+    LENS_CORRECTION_ON_90  = 2,
+    LENS_CORRECTION_DARK   = LENS_CORRECTION_OFF,
+    LENS_CORRECTION_NORMAL = LENS_CORRECTION_ON_70,
+    LENS_CORRECTION_BRIGHT = LENS_CORRECTION_ON_90
+};
+
+struct PackageParameterCameraSelect {
+    enum CameraSelect camera;
```
bunnei: This line (and ones like it) break the build on AppVeyor. I think you just want to get rid of the 'enum' here.

src/core/hle/service/cam/cam.h
```diff
@@ -6,10 +6,167 @@
...
+    u8 padding;
+    s16 autoExposureWindowX;
+    s16 autoExposureWindowY;
+    s16 autoExposureWindowWidth;
+    s16 autoExposureWindowHeight;
+    s16 autoWhiteBalanceWindowX;
+    s16 autoWhiteBalanceWindowY;
+    s16 autoWhiteBalanceWindowWidth;
+    s16 autoWhiteBalanceWindowHeight;
+};
```
bunnei: Add a static_assert to verify that this struct is the correct size

src/core/hle/kernel/timer.cpp
```diff
@@ -80,10 +80,10 @@ static void TimerCallback(u64 timer_handle, int cycles_late) {
     timer->signaled = true;
 
     // Resume all waiting threads
-    timer->WakeupAllWaitingThreads();
```
Subv: Uhm, what?

src/core/hle/service/cam/cam.h
```diff
@@ -6,10 +6,167 @@
 
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/service/service.h"
+#include "core/hle/service/y2r_u.h"
 
 namespace Service {
 namespace CAM {
 
+enum Port: u8 {
```
Subv: add a whitespace before each :

src/core/hle/service/cam/cam.h
Subv: Remove all the "enum" specifiers from the variables

Subv:
So, i'm pretty sure this PR is not mergeable anymore due to the amount of code pollution it has. I'm going to close it and add the enums myself on another PR, maybe i'll take on implementing camera stuff.

src/core/hle/kernel/timer.cpp
mailwl: i don't remember why, but not signaling timers itself leads for troubles in some games, for example ToTA. waiting threads just wait signal forever.
anyway, this is should not be in cam pr, just my mistake.
i will create issue with full description later, as i remember the exact reason

yuriks: There is still some bug in our WaitSynch implementation I think. BPB
video playback broke after bunnei did his last round of changes a long
time ago. (There's an issue for it somewhere.) This is probably related.

--yuriks


On Mon, Nov 23, 2015, at 08:56 PM, mailwl wrote:
> In src/core/hle/kernel/timer.cpp[1]:


> > @@ -80,10 +80,10 @@ static void TimerCallback(u64 timer_handle, int
> > cycles_late) {
>      timer->signaled = true;
>
>      // Resume all waiting threads
> -    timer->WakeupAllWaitingThreads(); i don't remember why, but not
>      signaling timers itself leads for troubles in some games, for
>      example ToTA. waiting threads just wait signal forever.
>
anyway, this is should not be in cam pr, just my mistake.
>
i will create issue with full description later, as i remember the
exact reason


>



Links:

  1. https://github.com/citra-emu/citra/pull/1102#discussion_r45673730
  2. https://github.com/citra-emu/citra/pull/1102/files#r45673730

PR #1104, opened 2015-09-03
OpenGL: Use Sampler Objects to decouple sampler config from textures

### LINKED ISSUE #978: Outdated wrap modes and tex filter modes in the hw renderer
The hardware renderer is currently binding the wrap_s, wrap_t, min_filter and mag_filter values to the textures only when they're first created, this leads to graphic artifacts if a game changes these values without flushing the texture.

@yuriks suggested using OpenGL Sampler objects but those are available only in OpenGL 3.3+

Pic of the bug
HW renderer
hw

SW renderer
sw

yuriks:
plus two small fixes/cleanups

Fixes #978

### DIFF
--- a/src/video_core/renderer_opengl/gl_resource_manager.h
+++ b/src/video_core/renderer_opengl/gl_resource_manager.h
@@ -37,6 +37,30 @@ public:
     GLuint handle = 0;
 };
 
+class OGLSampler : private NonCopyable {
+public:
+    OGLSampler() = default;
+    OGLSampler(OGLSampler&& o) { std::swap(handle, o.handle); }
+    ~OGLSampler() { Release(); }
+    OGLSampler& operator=(OGLSampler&& o) { std::swap(handle, o.handle); return *this; }
+
+    /// Creates a new internal OpenGL resource and stores the handle
+    void Create() {
+        if (handle != 0) return;
+        glGenSamplers(1, &handle);
+    }
+
+    /// Deletes the internal OpenGL resource
+    void Release() {
+        if (handle == 0) return;
+        glDeleteSamplers(1, &handle);
+        OpenGLState::ResetSampler(handle);
+        handle = 0;
+    }
+
+    GLuint handle = 0;
+};
+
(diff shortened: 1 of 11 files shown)


Subv:
LGTM

src/video_core/renderer_opengl/gl_resource_manager.h
```diff
@@ -37,6 +37,30 @@ class OGLTexture : private NonCopyable {
     GLuint handle = 0;
 };
 
+class OGLSampler : private NonCopyable {
+public:
+    OGLSampler() = default;
+    OGLSampler(OGLSampler&& o) { std::swap(handle, o.handle); }
+    ~OGLSampler() { Release(); }
+    OGLSampler& operator=(OGLSampler&& o) { std::swap(handle, o.handle); return *this; }
```
archshift: Split this into multiple lines?

yuriks: This is very repetitive code simply copy&pasted several times (check the rest of the file). I tried to remove the redundancy using templates last time but couldn't find a good way of doing it. I don't think splitting this in several lines will improve anything.

src/video_core/renderer_opengl/gl_resource_manager.h
```diff
@@ -37,6 +37,30 @@ class OGLTexture : private NonCopyable {
     GLuint handle = 0;
 };
 
+class OGLSampler : private NonCopyable {
+public:
+    OGLSampler() = default;
+    OGLSampler(OGLSampler&& o) { std::swap(handle, o.handle); }
```
lioncash: couldn't this be `OGLSampler(OGLSampler&& o) = default;`? The default move constructor will always do a member-wise move.

yuriks: VS2013 can't into auto-generated move constructors ( :( ) but this wouldn't work anyway, since the compiler can't know what what the semantics for moving an OpenGL handle are.

bunnei:
LGTM

PR #671, opened 2015-03-22
Fix for Issue #449 Qt errors on startup.

Ramuh:
windowHandle() is not initialized until the Window is shown. Moved the slot connect to main.cpp file after show(). Had to change the OnFramebufferSizeChanged(); to a slot.

PR #1106, opened 2015-09-04
citra-qt: Fix connect error on startup (#449)

### LINKED ISSUE #449: Qt errors on startup
```
QObject::connect: Cannot connect (null)::screenChanged(QScreen*) to GRenderWindow::OnFramebufferSizeChanged()
QObject::connect: No such slot GraphicsFramebufferWidget::OnFramebufferWidthChanged(int) in ..\..\..\src\citra_qt\debugger\graphics_framebuffer.cpp:87
QObject::connect:  (receiver name: 'PicaFramebuffer')
QObject::connect: No such slot GraphicsFramebufferWidget::OnFramebufferHeightChanged(int) in ..\..\..\src\citra_qt\debugger\graphics_framebuffer.cpp:88 
QObject::connect:  (receiver name: 'PicaFramebuffer')
```
They've been happening for a while, I'm not sure when it started

Kloen:
Finished #671 
Fixes #449

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -128,9 +128,6 @@ GRenderWindow::GRenderWindow(QWidget* parent, EmuThread* emu_thread) :
 
     BackupGeometry();
 
-#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
-    connect(this->windowHandle(), SIGNAL(screenChanged(QScreen*)), this, SLOT(OnFramebufferSizeChanged()));
-#endif
 }
 
 void GRenderWindow::moveContext()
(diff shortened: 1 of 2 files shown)


src/citra_qt/bootmanager.cpp
```diff
@@ -274,3 +271,12 @@ void GRenderWindow::OnEmulationStarting(EmuThread* emu_thread) {
 void GRenderWindow::OnEmulationStopping() {
     emu_thread = nullptr;
 }
+
+void GRenderWindow::showEvent(QShowEvent * event) {
+    QWidget::showEvent(event);
+
+    // windowHandle() is not initialized until the Window is shown, so we connect it here.
+    #if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+        connect(this->windowHandle(), SIGNAL(screenChanged(QScreen*)), this, SLOT(OnFramebufferSizeChanged()));
```
LittleWhite-tb: I think there is a risk that the connect will be done multiple times during the execution of citra. Mainly because in multiple wndow mode, we are showing/hiding the window.

mattvail: You can specify the type in this case: Qt::UniqueConnection

"If you pass the Qt::UniqueConnection type, the connection will only be made if it is not a duplicate. If there is already a duplicate (exact same signal to the exact same slot on the same objects), the connection will fail and connect will return false."

LittleWhite-tb: That would be a perfect solution :)

Kloen:
Addressed

yuriks:
LGTM

bunnei:
LGTM

PR #1108, opened 2015-09-05
Add missing -ldl for glad in CMakeLists.txt

linkmauve:
Fixes a compilation failure introduced in a1a5570e9735cb3ead0149b26d00eb4d6ade744f.

Before merging that, somebody should check if the same issue happens on MinGW, I only reproduced it on gcc and clang on glibc.


yuriks:
LGTM (cc @darkf )

kitling:
I can confirm that this works for me, but the linux build on travis seems to still be having issues...

LGTM, though.

externals/glad/CMakeLists.txt
```diff
@@ -9,3 +9,7 @@ set(HEADERS
 create_directory_groups(${SRCS} ${HEADERS})
 add_library(glad STATIC ${SRCS} ${HEADERS})
 target_include_directories(glad PUBLIC "include/")
+
+if (LINUX)
```
yuriks: Turns out `LINUX` isn't an actual variable, sorry. :| `${CMAKE_SYSTEM_NAME} MATCHES "Linux"` is our best bet I think.

PR #1109, opened 2015-09-05
Also fixes missing -ldl for glad.

kitling:
Ahem... #1108 

My take on this, adding a conditionally set LINUX option.


yuriks:
I think this is overkill, I'd rather just do the matches in the glad condition. I'd prefer to avoid having the root CMakeLists.txt accumulate more cruft.

kitling:
*nods*

I was just keeping in mind the DRY principle.

PR #1111, opened 2015-09-05
Stop emulation when render window is closed

### LINKED ISSUE #1110: Citra continues emulation even if the render window is closed.
#### Reproduction
- Open a game with citra (Qt)
- Make sure you are rendering to a separate window
- Close the render window

#### What I expected
In this situation, Citra should either halt emulation, or disable render-to-main-window.

#### What happens
Citra continues emulation, which can either be halted with the menu stop button or returned to rendering to main window.

LittleWhite-tb:
Fixes #1110

### DIFF
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -110,6 +110,8 @@ public:
     void restoreGeometry(const QByteArray& geometry); // overridden
     QByteArray saveGeometry();  // overridden
 
+    void closeEvent(QCloseEvent* event) override;
+
     void keyPressEvent(QKeyEvent* event) override;
     void keyReleaseEvent(QKeyEvent* event) override;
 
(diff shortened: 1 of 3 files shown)


src/citra_qt/main.cpp
```diff
@@ -268,6 +270,9 @@ void GMainWindow::BootGame(const std::string& filename) {
 }
 
 void GMainWindow::ShutdownGame() {
+    if (!emu_thread)
```
neobrain: Needs a comment explaining when this wouldn't be the case. If this is just a safety check to workaround broken design, please drop it; instead fix the true issue, or add a TODO about it.

src/citra_qt/bootmanager.h
```diff
@@ -129,6 +131,10 @@ public slots:
     void OnEmulationStarting(EmuThread* emu_thread);
     void OnEmulationStopping();
 
+signals:
+    /// Emitted when the Windows is closed
+    void WindowClosed();
```
neobrain: Why not simply "Closed()"?

LittleWhite-tb:
@neobrain I have rethink my code to remove the bad code. Now, I monitor the render window closing only when the emulation is started.

bunnei:
LGTM

PR #1112, opened 2015-09-05
Have the ARM interpreter generate callstacks for running applications

### LINKED ISSUE #586: Qt: Callstack widget seems to have totally stopped working
Where before it occasionally used to decide to show the callstack information, now its output has always (in my experience) been totally blank. 

EDIT: More details

archshift:
This fixes the callstack widget (closes #586), although it removes some of its past "features" (which didn't work anyway).

I also changed `svcBreak` to print out a backtrace whenever it is triggered, for easier debugging when the emulated program panics.

CAVEAT: this does incur a minor performance cost, increasing the time spent in dyncom by ~1-2ms. If desired, I can add a CMake variable for disabling populating the callstack during execution.

### DIFF
--- a/src/core/arm/arm_interface.h
+++ b/src/core/arm/arm_interface.h
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <vector>
+
 #include "common/common_types.h"
 #include "core/arm/skyeye_common/arm_regformat.h"
 
@@ -110,6 +112,8 @@ public:
      * @param value The value to store into the CP15 register.
      */
     virtual void SetCP15Register(CP15Register reg, u32 value) = 0;
+
+    virtual std::vector<u32> GetStackTrace() const = 0;
 
     /**
      * Advance the CPU core by the specified number of ticks (e.g. to simulate CPU execution time)
--- a/src/core/arm/dyncom/arm_dyncom.cpp
+++ b/src/core/arm/dyncom/arm_dyncom.cpp
@@ -72,6 +72,10 @@ void ARM_DynCom::SetCP15Register(CP15Register reg, u32 value) {
     state->CP15[reg] = value;
 }
 
+std::vector<u32> ARM_DynCom::GetStackTrace() const {
+    return state->stack_trace;
+}
+
 void ARM_DynCom::AddTicks(u64 ticks) {
     down_count -= ticks;
     if (down_count < 0)
@@ -89,12 +93,15 @@ void ARM_DynCom::ExecuteInstructions(int num_instructions) {
 }
 
 void ARM_DynCom::ResetContext(Core::ThreadContext& context, u32 stack_top, u32 entry_point, u32 arg) {
-    memset(&context, 0, sizeof(Core::ThreadContext));
+    context = Core::ThreadContext();
 
     context.cpu_registers[0] = arg;
     context.pc = entry_point;
     context.sp = stack_top;
     context.cpsr = 0x1F; // Usermode
+
+    context.stack_trace.clear();
+    context.stack_trace.reserve(128);
 }
 
 void ARM_DynCom::SaveContext(Core::ThreadContext& ctx) {
@@ -108,6 +115,8 @@ void ARM_DynCom::SaveContext(Core::ThreadContext& ctx) {
 
     ctx.fpscr = state->VFP[1];
     ctx.fpexc = state->VFP[2];
+
+    ctx.stack_trace = state->stack_trace;
 }
 
 void ARM_DynCom::LoadContext(const Core::ThreadContext& ctx) {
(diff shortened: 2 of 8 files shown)


src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -4104,14 +4116,13 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
...
         //
         // This is sufficient for citra, as the CPU for the 3DS does not implement Jazelle.
-
         if (inst_base->cond == 0xE || CondPassed(cpu, inst_base->cond)) {
             bx_inst* const inst_cream = (bx_inst*)inst_base->component;
 
             u32 address = RM;
 
             if (inst_cream->Rm == 15)
-                address += 2 * cpu->GetInstructionSize();
```
archshift: Oops!

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -1,3 +1,4 @@
+
```
lioncash: unnecessary newline

src/core/arm/arm_interface.h
```diff
@@ -115,6 +117,8 @@ class ARM_Interface : NonCopyable {
      */
     virtual void SetCP15Register(CP15Register reg, u32 value) = 0;
 
+    virtual std::vector<u32> GetStackTrace() = 0;
```
lioncash: const function

src/core/arm/arm_interface.h
```diff
@@ -115,6 +117,8 @@ class ARM_Interface : NonCopyable {
      */
     virtual void SetCP15Register(CP15Register reg, u32 value) = 0;
 
+    virtual std::vector<u32> GetStackTrace() const = 0;
```
yuriks: Why not `const std::vector<u32>&`? Also, `u32` -> `VAddr`.

src/core/arm/dyncom/arm_dyncom.cpp
```diff
@@ -108,6 +115,8 @@ void ARM_DynCom::SaveContext(Core::ThreadContext& ctx) {
 
     ctx.fpscr = state->VFP[1];
     ctx.fpexc = state->VFP[2];
+
+    ctx.stack_trace = state->stack_trace;
```
yuriks: I bet this is where the bulk of the perf-hit comes from. You should move or swap the vector here and on load instead.

src/core/arm/arm_interface.h
archshift: I figured that:
- This is an interface, and perhaps a future backend might like to create a stack trace vector on the spot, and it wouldn't be possible to return a const reference to a local vector.
- This realistically would never be called in a performance-dependent scenario

As for `VAddr`, good point.

archshift:
We've discovered a pitfall where this will fail to generate proper stack traces for certain recursive code scenarios.

Any ideas for how we can work around this issue?

src/core/arm/arm_interface.h
yuriks: Ok.

yuriks:
As seen on #citra-dev™:
```
[23:53:56] <yuriks> archshift: idea for the stack trace
[23:54:45] <yuriks> archshift: do PUSH_STACK_TRACE every time LR is saved somehow
[23:55:17] <yuriks> archshift: PUSH_STACK_TRACE pushes the stack position of the save along with the value of LR
[23:55:44] <yuriks> it also compares the current stack position and discards any entries that are below it
[23:56:17] <yuriks> GetStackTrace can also do the same
[23:56:36] <yuriks> that way you only need to update the stack trace on pushes, and don't need to do anything on return
[00:20:41] <yuriks> archshift: does that sound like it can work?
[00:54:14] <archshift> That's what I was planning originally, it didn't work out but only because I ended up typoing somewhere (which I later discovered)
```

bunnei:
1-2ms is somewhat significant... I'd prefer that be part of a "debugmode" option (which I'm sure will be useful for other things)

bunnei:

archshift:
Sort of forgot about it, but I'll work on it this week.

bunnei:
@archshift Reminder to wrap this up...

bunnei:
@archshift, please finish this PR if you have any intention of doing so, otherwise close it.

RavenHome1:
wow this is an old PR, i guess i will ping it for a reminder XD

archshift:
To be quite honest, I couldn't find a good way to implement this such that it a) worked, and b) didn't spread the same code all over different dyncom instruction implementations.

I take responsibility: it was dumb to keep this open for so many months with little to no feedback.

PR #1114, opened 2015-09-06
DynCom: Converted all magic 0xE condition code checks to ConditionCode::AL

archshift:


lioncash:
There's also ones in vfpinstr.cpp as well.

archshift:
Updated.

lioncash:
looks good 2 me

Issue #1115, opened 2015-09-06
ScanDirectoryTree will infinitely recurse upon meeting a symlink to its parent directory.


archshift:

LittleWhite-tb:
Well, using a counter was a way to solve it, of course, but why not just checking if the file is a symlink, and in such case, we would ignore it ?
When I was reading code, I was completely wondering why "recursion" was an int and not a bool.

JayFoxRox:
It used to be a boolean and was only recently changed. The PR / review phase had been there for a long time btw.
Also disabling symlinks is a really bad idea imho.

I also sometimes symlink some of my games as I don't have enough storage on my SSD to keep them all in one folder + I can manually sort them with symlinks  (and especially my hardware tests for use in a dropbox are symlinked).

However, it shouldn't be a GUI option either (Confusing to users). Having the option in the ini/config file is probably good though.

LittleWhite-tb:
I was thinking about disabling the symlink. Even if it can be useful, since you can precise the input folder, the need of symlink is decreased. Otherwise, maybe it is possible to detect symlink recursion ?
By not letting the user define this value will cause two things :
* no advantages to have a counter, since there is high chance it is too high (no performance gain, we just avoid the infinite loop)
* if the counter is too low, users will report a bug
Of course, letting the user choose the value is very difficult for non technical users.

(Sorry to have missed the review phase ...)

PR #1116, opened 2015-09-06
Replace ScanDirectoryTree with a non-recursive implementation

linkmauve:
This makes it do only what we need for our readdir implementation, thus
improving performances a lot in heavy trees and fixing bugs like #1115.

### DIFF
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -420,9 +420,7 @@ bool CreateEmptyFile(const std::string &filename)
 }
 
 
-// Scans the directory tree gets, starting from _Directory and adds the
-// results into parentEntry. Returns the number of files+directories found
-u32 ScanDirectoryTree(const std::string &directory, FSTEntry& parentEntry)
+u32 ReadDirectory(const std::string& directory, FSTEntry& parentEntry, int recursion)
 {
     LOG_TRACE(Common_Filesystem, "directory %s", directory.c_str());
     // How many files + directories we found
@@ -467,12 +465,14 @@ u32 ScanDirectoryTree(const std::string &directory, FSTEntry& parentEntry)
         if (IsDirectory(entry.physicalName.c_str()))
         {
             entry.isDirectory = true;
-            // is a directory, lets go inside
-            entry.size = ScanDirectoryTree(entry.physicalName, entry);
-            foundEntries += (u32)entry.size;
+            if (recursion > 0)
+            {
+                entry.size = ReadDirectory(entry.physicalName, entry, recursion - 1);
+                foundEntries += (u32)entry.size;
+            }
         }
         else
-        { // is a file
+        {
             entry.isDirectory = false;
             entry.size = GetSize(entry.physicalName.c_str());
         }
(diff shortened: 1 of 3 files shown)


yuriks:
LGTM, if we need recursive scanning we'll need to implement it in a better way later anyway.

archshift:
I wouldn't exactly call this **fixing** #1115. `ScanDirectoryTree` is still useful (in fact, I'm using it in #1095). Can we add a `recursive_scan` flag instead?

yuriks:
@archshift: I thought the same thing about the flag, but then realized the function would still be unusable in the existence of symlinks anyway.

linkmauve:
We would need some tagging in order not to recurse on symlinks, which makes this function way more complicated.

We could still assume the user won’t symlink to a parent, or have a maximum number before which we abandon the descent in the tree.

linkmauve:
Here is a new version with a recursion limit, defaulting to 0.

src/common/file_util.cpp
```diff
@@ -420,9 +420,10 @@ bool CreateEmptyFile(const std::string &filename)
 }
 
 
-// Scans the directory tree gets, starting from _Directory and adds the
-// results into parentEntry. Returns the number of files+directories found
-u32 ScanDirectoryTree(const std::string &directory, FSTEntry& parentEntry)
+// Reads the contents of a directory, and adds the results into
+// parentEntry, optionally going down up to recursion levels.  Returns
+// the number of files+directories found
```
yuriks: Could take the opportunity to convert this to a proper doxy-comment.

src/common/file_util.h
```diff
@@ -96,9 +96,10 @@ bool Copy(const std::string &srcFilename, const std::string &destFilename);
 // creates an empty file filename, returns true on success
 bool CreateEmptyFile(const std::string &filename);
 
-// Scans the directory tree gets, starting from _Directory and adds the
-// results into parentEntry. Returns the number of files+directories found
-u32 ScanDirectoryTree(const std::string &directory, FSTEntry& parentEntry);
+// Reads the contents of a directory, and adds the results into
+// parentEntry, optionally going down up to recursion levels.  Returns
+// the number of files+directories found
+u32 ReadDirectory(const std::string &directory, FSTEntry& parentEntry, int recursion=0);
```
yuriks: Spaces around the `=`.

yuriks: `&` attached to the type, not argument name.

bunnei:
@linkmauve what's the status of this PR? Can we get it wrapped up? @archshift, will this be OK with #1095?

archshift:
If this is merged I can just remove the recursive scanning option from #1095.

bunnei:
@archshift how critical of a feature is this for #1095? I'd like to get both of these PRs closed, but recursive scanning seems useful (esp. for those of us on Windows who aren't using symlinks).

linkmauve:
Here, rebased including all comments.

archshift:
It's not exactly critical, just a convenient feature that people might appreciate. I don't think it's much of a problem to just remove it though.

bunnei:
@linkmauve Please resolve merge conflicts or close this PR.

bunnei:
Final reminder - I plan on closing this PR if I do not hear any response by the end of the week.

kitling:
@bunnei If this is closed, would it be okay for someone else to fork it and resolve the merge conflicts, and open a new PR?

bunnei:
@kitling Of course that would be welcome!

PR #156, opened 2014-10-27
Changed debugger font to Courier, added customization option

Gareth422:
A monospace font would allow for adresses and other debugging information to be properly aligned. Fixes issue #67.

Issue #67, opened 2014-08-21
Debugger: Use monospace font


archshift:
A monospace font (such as Source Code Pro) would provide for better alignment and better readability with the debugger tools.

Zaneo:
Can this be closed? It looks like a commit addressing the issue has been merged.

yuriks:
Yup

PR #1118, opened 2015-09-07
citra-qt: Use monospace font on Disassembler and ARM Registers

Kloen:
Picked up #156 if order to finish #67. If any other debugger tool should use it, just say it and I'll add it.

### DIFF
--- a/src/citra_qt/debugger/disassembler.cpp
+++ b/src/citra_qt/debugger/disassembler.cpp
@@ -9,6 +9,8 @@
 #include "../bootmanager.h"
 #include "../hotkeys.h"
 
+#include "../util/util.h"
+
 #include "core/memory.h"
 
 #include "core/core.h"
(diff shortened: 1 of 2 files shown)


src/citra_qt/main.cpp
```diff
@@ -58,6 +58,10 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     ui.setupUi(this);
     statusBar()->hide();
 
+    QSettings settings;
+    if (settings.value("debugger_font").isNull())
+        settings.setValue("debugger_font", QVariant("Courier"));
```
archshift: Courier isn't available on all platforms.

yuriks:
We have a function on util/util.cpp which I created to avoid duplicating monospace font policy through the codebase, please use that function and then if desired implement the option to allow the user to choose a font based using that too.

Kloen:
@yuriks wops, didn't saw it, definitely Notepad++ it's not as good as VS to help -.-'

yuriks:
LGTM (please squash)
I'd like to see the font configuration thing you before too, but you can probably do that in a different PR.

Kloen:
Squashed. 

I'd like to do the font configuration if I could right now. But QT doesn't like vs2015 at all -.-' so I'll look into it once I can build citra on my PC, or on my freetime at work.

bunnei:
LGTM

PR #531, opened 2015-02-02
citra-qt: Add Input Setting dialog

chinhodado:
image

PR #1119, opened 2015-09-07
citra-qt: Add input settings configuration

Kloen:
Based on #531 

Tried to address every comment the best I could. Requesting for comments and thoughts before mark it as finished.

Mandatory screenshot:
image

To set a key just:
- Click on one of the QPushButtons, it will be marked as checked.
- As long as it remains checked, you can simply change the key by hitting the desired one on your keyboard.
- You can't set a key that it's already in use by any other input.

If a QPushButton if checked and you clic another one, the first will back to normal and you will work with the one you just click.


src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,166 @@
...
+        pushButtons[current_button_id]->setChecked(false);
+
+    pushButtons[i]->setChecked(true);
+    current_button_id = i;
+
+}
+
+void GInputsDialog::UpdateValue(int key) {
+
+    bool update = true;
```
neobrain:
```
bool update = std::none_of(std::begin(Settings::NativeInput::All), std::end(Settings::NativeInput::All),
                          [key](Settings::NativeInput::Values val) { return key == val }); 
```

neobrain:
Can you describe how specifically keys are set? (what specifically does the user need to do, how do widgets change state, etc -- please don't expect people to actually try this out themselves just to get an understanding about that)

src/citra_qt/config/inputs.cpp
Subv: std::none_of instead of boost::none_of, please

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,166 @@
...
+    temp_settings = Settings::Values(Settings::values);
+    signalMapper = new QSignalMapper(this);
+
+    // create the lineEdits
+    for (int i = 0; i < Settings::NativeInput::NUM_INPUTS; ++i) {
+        pushButtons[i] = new QPushButton(GetKeyName(temp_settings.input_mappings[Settings::NativeInput::All[i]]), this);
+        pushButtons[i]->setCheckable(true);
+    }
+
+    std::string labelArray[] = {
```
Subv: static const

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,166 @@
...
+        "A:", "B:", "X:", "Y:",
+        "L:", "R:", "ZL:", "ZR:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Start:", "Select:", "Home:"
+    };
+
+    std::string groups_names[] = {
```
Subv: static const

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,166 @@
...
+        QGroupBox* group = new QGroupBox(tr(groups_names[i].c_str()));
+        QGridLayout* box = new QGridLayout;
+        for (int x = 0; x < 4; ++x) {
+            QLabel* label = new QLabel(tr(labelArray[id].c_str()));
+            box->addWidget(label, (int) x % 2 * 2, (int) x / 2 * 2);
+            box->addWidget(pushButtons[id], (int) x % 2 * 2 + 1, (int) x / 2 * 2);
+            signalMapper->setMapping(pushButtons[id], id);
+            connect(pushButtons[id], SIGNAL(clicked()), signalMapper, SLOT(map()));
+            ++id;
+            if (id > 22) break;
```
Subv: break in newline

src/citra_qt/config/inputs.cpp
neobrain: @Subv There's literally no reason to fear std::none_of over boost::none_of, and in this instance the latter considerably improves readability of that line.

Kloen:
Addressed. Also explained the use on the PR description, If more information is needed, just tell me, personally I think it's good as is.

src/citra/config.cpp
```diff
@@ -43,10 +43,10 @@ bool Config::LoadINI(INIReader* config, const char* location, const std::string&
 static const std::array<int, Settings::NativeInput::NUM_INPUTS> defaults = {
     GLFW_KEY_A, GLFW_KEY_S, GLFW_KEY_Z, GLFW_KEY_X,
     GLFW_KEY_Q, GLFW_KEY_W, GLFW_KEY_1, GLFW_KEY_2,
-    GLFW_KEY_M, GLFW_KEY_N, GLFW_KEY_B,
     GLFW_KEY_T, GLFW_KEY_G, GLFW_KEY_F, GLFW_KEY_H,
     GLFW_KEY_UP, GLFW_KEY_DOWN, GLFW_KEY_LEFT, GLFW_KEY_RIGHT,
-    GLFW_KEY_I, GLFW_KEY_K, GLFW_KEY_J, GLFW_KEY_L
+    GLFW_KEY_I, GLFW_KEY_K, GLFW_KEY_J, GLFW_KEY_L,
+    GLFW_KEY_M, GLFW_KEY_N, GLFW_KEY_B
```
linkmauve: Here and in many other places, why do you move that line?

Kloen: It was moved on the original PR, and I kept it like that cause I thought that it was the easiest way to order and group the QPushButtons in a way I think it's coherent. And only by looping over an array.

neobrain: Not sure if I'm understanding you correctly, but the order of the elements in this array *really* shouldn't have any effect on the GUI code. If it does, then that sounds like there's a false dependency on core interfaces.

Kloen: They were reorderer on ```core/settings.h```, and that ones are used on ```citra_qt/config/input.cpp``` to iterate over it assigning the position of the ```QPushButton``` on the array, which it's used later to draw the widgets. 

These in particular, are used a few lines below, depending on the ones of ```core/settings.h```

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,161 @@
+// Copyright 2014 Citra Emulator Project
```
bunnei: 2015?

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include "inputs.h"
+
+const int NUM_SECONDS = 5;
+
+GInputsDialog::GInputsDialog(QWidget* parent) : QDialog(parent) {
+    // create a copy of the current settings
+    temp_settings = Settings::Values(Settings::values);
+    signalMapper = new QSignalMapper(this);
+
+    // create the lineEdits
```
bunnei: Why are these called 'lineEdits'?

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,161 @@
...
+        "A:", "B:", "X:", "Y:",
+        "L:", "R:", "ZL:", "ZR:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Start:", "Select:", "Home:"
+    };
+
+    static const std::string groups_names[] = {
+        "Basic controls", "Extended controls", "Direcctional pad", "Circle pad", "C-Stick", "System buttons"
```
bunnei: Typo with 'Direcctional'

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,161 @@
...
+        QGroupBox* group = new QGroupBox(tr(groups_names[i].c_str()));
+        QGridLayout* box = new QGridLayout;
+        for (int x = 0; x < 4; ++x) {
+            QLabel* label = new QLabel(tr(labelArray[id].c_str()));
+            box->addWidget(label, (int) x % 2 * 2, (int) x / 2 * 2);
+            box->addWidget(pushButtons[id], (int) x % 2 * 2 + 1, (int) x / 2 * 2);
+            signalMapper->setMapping(pushButtons[id], id);
+            connect(pushButtons[id], SIGNAL(clicked()), signalMapper, SLOT(map()));
+            ++id;
+            if (id > 22)
```
bunnei: Why 22? If this is corresponding to the `pushButtons` size, then use that directly.

bunnei: Or `NUM_INPUTS`?

bunnei:
Thanks for resurrecting this. I'll test it tonight. There is no need for this to be 'RFC', I think we should go ahead and try to get this merged this time around.

Kloen:
Addressed all @bunnei comments. Rebased and squashed.

New: If double clicked on a button, it becomes unchecked.
New: Prevent default behavior on input buttons (As an Enter/Backspace key user, I want to be able to use them).

bunnei:
Can you fix the merge conflicts?

src/citra_qt/config.cpp
```diff
@@ -21,13 +20,13 @@ Config::Config() {
     Reload();
 }
 
-static const std::array<QVariant, Settings::NativeInput::NUM_INPUTS> defaults = {
+const std::array<QVariant, Settings::NativeInput::NUM_INPUTS> Config::defaults = {
```
yuriks: Why is this an array of QVariant?

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include <QSettings>
+#include <QLabel>
+#include <QGridLayout>
+#include <QDialogButtonBox>
+#include <QGroupBox>
+#include <QVBoxLayout>
+
+#include "common/logging/log.h"
+#include "config.h"
+#include "inputs.h"
```
lioncash: These should be relative paths

```cpp
#include "citra_qt/config/config.h"
```

etc.

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,166 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <QKeySequence>
+#include <QSettings>
+#include <QLabel>
+#include <QGridLayout>
+#include <QDialogButtonBox>
+#include <QGroupBox>
+#include <QVBoxLayout>
```
lioncash: These should be alphabetically sorted.

src/citra_qt/config/inputs.h
```diff
@@ -0,0 +1,49 @@
...
+#pragma once
+
+#include <array>
+
+#include <QDialog>
+#include <QPushButton>
+#include <QSignalMapper>
+#include <QKeyEvent>
+
+#include "../core/settings.h"
```
lioncash:
```cpp
#include "core/settings.h"
```

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,166 @@
...
+        "A:", "B:", "X:", "Y:",
+        "L:", "R:", "ZL:", "ZR:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Start:", "Select:", "Home:"
+    };
+
+    static const std::string groups_names[] = {
+        "Basic controls", "Extended controls", "Directional pad", "Circle pad", "C-Stick", "System buttons"
```
yuriks: - "Basic controls" -> "Face buttons"
 - "Extended controls" -> "Shoulder buttons"
 - "Directional pad" -> "D-Pad"

yuriks:
Going by the screenshot (Please update it if it has changed, since the branch has conflicts I cannot test it.) the order of the buttons in the layout seems quite random:

 - The YXBA buttons should be in some order at least vaguely resembling the console layout, I'd recomment "YX / BA"
 - Same for the Circle/D-pad/C-stick. Either make it "Left Right / Up Down" or in clockwise order or something.
 - L and R as well as ZL and ZR should be in the same line.
 - Put Circle-Pad and D-Pad on the left side of the dialog, and face buttons and C-Stick on the right.
 - This one doesn't matter as much, but I think "Start, Select, Home" would be a better ordering for that group.

archshift:
Why did you close this?

Kloen:
Uh, that was git just being stupid and overriding all my local changes when pushed. Maybe I did something wrong... well nevermind, will recover my changes and fix this.

Kloen:
Well, all should be fixed.

@yuriks updated screenshot after reordering buttons. About the D-Pad, well, I feel it better as "Directional pad", but that's just an end user opinion.

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,164 @@
...
+        "Y:", "X:", "B:", "A:",
+        "L:", "R:", "ZL:", "ZR:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Left:", "Right:", "Up:", "Down:",
+        "Start:", "Select:", "Home:"
+    };
+
+    static const std::string groups_names[] = {
+        "Face buttons", "Shoulder buttons", "C-Stick", "Directional pad", "Circle pad", "System buttons"
```
neobrain: Nouns should be uppercased in the GUI by convention.

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,164 @@
...
+    // create the QPushButtons
+    for (int i = 0; i < Settings::NativeInput::NUM_INPUTS; ++i) {
+        pushButtons[i] = new QPushButton(GetKeyName(Settings::NativeInput::All[i]), this);
+        pushButtons[i]->setCheckable(true);
+        pushButtons[i]->setDefault(false);
+        pushButtons[i]->setAutoDefault(false);
+    }
+
+    static const std::string labelArray[] = {
+        "Y:", "X:", "B:", "A:",
```
neobrain: Have you considered making these texts part of the QPushButton label (written before the key binding) instead of using a standalone label?

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,164 @@
...
+        "Start:", "Select:", "Home:"
+    };
+
+    static const std::string groups_names[] = {
+        "Face buttons", "Shoulder buttons", "C-Stick", "Directional pad", "Circle pad", "System buttons"
+    };
+
+    QGridLayout* grid = new QGridLayout();
+    int id = 0;
+    for (int i = 0; i < 6; ++i) {
```
neobrain: Too many magic values here. I suppose you meant to write something like `sizeof(groups_names)/sizeof(groups_names[0])` here? (I think we have a macro for that somewhere.)

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,164 @@
...
+        QGroupBox* group = new QGroupBox(tr(groups_names[i].c_str()));
+        QGridLayout* box = new QGridLayout;
+        for (int x = 0; x < 4; ++x) {
```
neobrain: same thing about magic values

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    setLayout(verticalLayout);
+    setWindowTitle(tr("Input Settings"));
+
+    // set up event handlers for the buttons
+    QPushButton* defaultButton = buttonBox->button(QDialogButtonBox::RestoreDefaults);
+    connect(defaultButton, SIGNAL(clicked()), this, SLOT(RestoreDefaultSettings()));
+
+    QPushButton* okButton = buttonBox->button(QDialogButtonBox::Ok);
+    connect(okButton, SIGNAL(clicked()), this, SLOT(SaveSettings()));
```
neobrain: You'll want to connect the QDialogButtonBox's accepted() signal here, instead.

src/citra_qt/config/inputs.h
```diff
@@ -0,0 +1,49 @@
...
+    /// Display the button settings on the QPushButtons from the given values
+    void UpdateKeyLabels();
+
+    /// Temporary settings used when configuration is changed but not saved yet
+    Settings::Values temp_settings;
+
+    /// Store all the QPushButtons signals to use the right parameters on ChangeValue
+    QSignalMapper* signalMapper;
+
+    /// Current QPushButton id
```
neobrain: -1 is not a valid "id". the meaning of that value should be explained in the docstring.

src/citra_qt/config/inputs.h
```diff
@@ -0,0 +1,49 @@
...
+    /// An array of all the QPushButtons for easy iterating though them
+    std::array<QPushButton*, 23> pushButtons;
+
+    /// Display the button settings on the QPushButtons from the given values
+    void UpdateKeyLabels();
+
+    /// Temporary settings used when configuration is changed but not saved yet
+    Settings::Values temp_settings;
+
+    /// Store all the QPushButtons signals to use the right parameters on ChangeValue
```
neobrain: This docstring describes implementation behavior (note the "<active verb>..." phrasing) instead of documenting the member variable that is being referred to (signalMapper).

Also, `signal_mapper`.

src/citra_qt/config/inputs.h
```diff
@@ -0,0 +1,49 @@
...
+    /// Given a key code, return the key name. In addition to QKeySequence(key_code).toString(); this will translate modifier keys properly.
+    static QString GetKeyName(int key_code);
+
+    void keyPressEvent(QKeyEvent* event) override;
+
+private:
+    /// An array of all the QPushButtons for easy iterating though them
+    std::array<QPushButton*, 23> pushButtons;
+
+    /// Display the button settings on the QPushButtons from the given values
```
neobrain: This doesn't actually "display" them, but changes the labels.

Suggested alternative: "Update QPushButton labels to reflect current button mapping."

neobrain: Put private methods above member variables.

src/citra_qt/config/inputs.h
```diff
@@ -0,0 +1,49 @@
...
+public:
+    GInputsDialog(QWidget* parent = nullptr);
+
+    /// Given a key code, return the key name. In addition to QKeySequence(key_code).toString(); this will translate modifier keys properly.
+    static QString GetKeyName(int key_code);
+
+    void keyPressEvent(QKeyEvent* event) override;
+
+private:
+    /// An array of all the QPushButtons for easy iterating though them
```
neobrain: *easy iteration over them

src/citra_qt/config/inputs.h
```diff
@@ -0,0 +1,49 @@
...
+#include "core/settings.h"
+
+/// The input configuration dialog
+class GInputsDialog : public QDialog {
+    Q_OBJECT
+
+public:
+    GInputsDialog(QWidget* parent = nullptr);
+
+    /// Given a key code, return the key name. In addition to QKeySequence(key_code).toString(); this will translate modifier keys properly.
```
neobrain: Can you be more specific what "properly" means, and where the QKeySequence thing fails?
(also, the ";" is redundant)

src/citra_qt/config/inputs.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <QDialog>
+#include <QKeyEvent>
+#include <QPushButton>
+#include <QSignalMapper>
+
+#include "core/settings.h"
+
+/// The input configuration dialog
+class GInputsDialog : public QDialog {
```
neobrain: No need for "G" prefixes.

src/citra_qt/config/inputs.h
```diff
@@ -0,0 +1,49 @@
...
+    /// Store all the QPushButtons signals to use the right parameters on ChangeValue
+    QSignalMapper* signalMapper;
+
+    /// Current QPushButton id
+    int current_button_id = -1;
+
+private slots:
+    void SaveSettings();
+    void RestoreDefaultSettings();
+    void ChangeValue(int i);
```
neobrain: Needs a docstring

src/citra_qt/config/inputs.h
```diff
@@ -0,0 +1,49 @@
...
+    void UpdateValue(int key);
```
neobrain: docstring

src/citra_qt/config/inputs.cpp
Kloen: If I understand you right, I think it  would be quite weird...

src/citra_qt/config/inputs.cpp
lioncash: It's present in `common/commonfuncs.h`

src/citra_qt/config/inputs.cpp
Kloen: QPushButton doesn't have accepted() signal.

neobrain: QDialogButtonBox's does.

Kloen: Already connected QDialogButtonBox on line 73:
```connect(buttonBox, SIGNAL(accepted()), this, SLOT(accept()));```

... Uh... I just realized that there are some issues with dialog behavior, so I'll look into them and maybe I could find a best approach to this part too.

Comment (neobrain) [inline src/citra_qt/config/inputs.cpp]: Heh yeah, connecting both shouldn't be necessary ;)

**Comment by Kloen** [comment]
All comments should be addressed by now.

**Comment by bunnei** [comment]
Please squash down to a single commit.
@yuriks @neobrain please review that your comments were addressed.

File: src/citra_qt/config.cpp [hunk 28]
```diff
@@ -21,13 +21,13 @@ Config::Config() {
     Reload();
 }
 
-static const std::array<QVariant, Settings::NativeInput::NUM_INPUTS> defaults = {
-    Qt::Key_A, Qt::Key_S, Qt::Key_Z, Qt::Key_X,
+const std::array<QVariant, Settings::NativeInput::NUM_INPUTS> Config::defaults = {
```
yuriks: This is still a QVariant array.

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,170 @@
...
+    };
+
+    QGridLayout* grid = new QGridLayout();
+    int id = 0;
+    for (int i = 0; i < ARRAY_SIZE(group_names); ++i) {
+        QGroupBox* group = new QGroupBox(tr(group_names[i].c_str()));
+        QGridLayout* box = new QGridLayout;
+        for (int x = 0; x <= (Settings::NativeInput::NUM_INPUTS / ARRAY_SIZE(group_names)); ++x) {
+            QLabel* label = new QLabel(tr(label_array[id].c_str()));
+            box->addWidget(label, (int) x / 2 * 2, (int) x % 2 * 2);
```
bunnei: IMO `x & ~1` is a little more clear for the second parameter, at first glance I didn't get why you were dividing by 2 then multiplying by 2

bunnei: Also, instead of `% 2`, I find `& 1` to be clearer

yuriks: Eh, I think I prefer current versions of both expressions. However, cast needs to stick to the `x` instead of having a space between them (makes it confusing to read).

src/citra_qt/config/inputs.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <QVBoxLayout>
+
+#include "citra_qt/config.h"
+#include "citra_qt/config/inputs.h"
+
+#include "common/common_funcs.h"
+#include "common/logging/log.h"
+
+InputsDialog::InputsDialog(QWidget* parent) : QDialog(parent) {
+    // create a copy of the current settings
```
bunnei: Total nitpick, but our general style is to capitalize the first letter of a comment, and not to formulate a full sentence unless it's a multi-line comment block. E.g.:
<pre>
// Create a copy of the current settings
</pre>
Or:
<pre>
/**
 * This is a multiline comment block. In this case, we use full sentences. We start a new line at
 * 100 characters. Blah blah blah.
 */
</pre>

bunnei:
Just tested this on Windows 8, x64, with Visual Studio 2015. I had several issues when testing OoT:
* With the default config, the left arrow makes Link go right, the right arrow has no effect, the down arrow makes Link go left, and the up arrow makes Link go backward. I'd appreciate it if something can test & reproduce these results.
* I can't actually configure anything, when I click a button it highlights, but then entering a key has no effect
* The '?' icon in the titlebar looks goofy and has no purpose... Can you remove it?
* Having 'Configure...' (which does nothing), and 'Configure Controls' menu options is weird... We could probably model after what Dolphin does, and just have one tabbed 'Citra Configuration' dialog that comes up when you click `Configure...`, with the only tab being 'Input Controls' (or something like that) for the time being.

src/citra_qt/config/inputs.cpp
neobrain: @bunnei Bitwise operations shouldn't be used for this purpose, since they convey the wrong intent (bit manipulation != doing something for every second element).

Kloen:
@bunnei :
1. Should be fixed, now the buttons are being set correctly.
2. I could only reproduce it with arrow keys (To be fixed Soon™), also, I'll add a error message when you try to set a key that it's already in use.
3. Soon™.
4. I see the point... I will implement it then as a base ```QDialog``` with a  ```QTabWidget``` with modular tabs if noone it's against it.

Also, could we get rid of ```citra_qt/config/controller_config*``` files?

bunnei:
I could get some keys to work, e.g. Shift/Control/Caps Lock,  but letter keys and arrows would not work. I will retest tonight. I would definitely suggest making the layout change to use a general tabbed configure dialog, as it sets the precedent for future config UI work (some of which I plan on doing next).

yuriks:
- Like bunnei, I can't bind the arrow keys, Tab, or Space
 - Trying to bind a key that is already bound gives no feedback. It should either double-bind or automatically clear the other assignment. (If double-binding is allowed then you should probably mark the two buttons somehow.)
 - The current assignment "flow" is *very* confusing. There is no feedback to indicate that it's in assignment mode, what you should do after clicking a button, and no feedback to tell you why an assignment failed, no nothing.

yuriks:
Also, there should be a way to remap all the keys in sequence somehow, instead of tediously clicking each button one by one.

neobrain:
@yuriks what use case do you have in mind for having a quick way to remap large parts of the inputs? IMO, that will happen at most once (upon initial Citra setup) and after that people will only remap single buttons individually.

neobrain:
@yuriks what use case do you have in mind for having a quick way to remap large parts of the inputs? IMO, that will happen at most once (upon initial Citra setup) and after that people will only remap single buttons individually, so I wouldn't really consider that feature crucial, especially if it conflicts with other design criteria.

yuriks:
@neobrain I often switch input layouts completely when playing different games in a the same emulator. Furthermore, the initial user experience is as, if not more, important than the one for repeat actions. Many emulators and games have this feature and I almost always use it when applicable.

neobrain:
Wouldn't some kind of "profile" feature be a more natural solution for that?

yuriks:
@neobrain: No, because implementing that is much more complicated, setting up a profile will still be a chore, and useless for me since I don't usually replay games, switching back and forth between them often.

Kloen:
So, as recall:

- [ ] Rework ```Configure...``` as a ```QTableWidget``` implementation.
- [ ] Add Input Configuration as a tab in ```Configure...```
- [ ] Fix input assignment bugs
- [ ] Improve feedback on user actions while mapping inputs.
- [ ] Add a way to remap both all the inputs at once or only input groups.
- [ ] Add input profiles.

Add input profiles it's not that hard, and quite a good option. Also, since the inputs are grouped, add a way to remap a group it's easier than make a full remap just for that, or simply remap key by key.

If you are all fine with these changes ( @bunnei @yuriks @neobrain ) I'll take a look on it during this weekend.

yuriks:
If you want to do profiles/mass reassignment as a separate PR, that's fine with me.

bunnei:
Ditto on potentially making input profiles a separate PR.

neobrain:
What specifically do you mean with " Rework Configure... as a QTabWidget implementation."? If it's about introducing a single configuration dialog which in the future will hold all "subdialogs" for emulation settings, then please postpone that to a separate PR too - it's far from critical to this PR and just sidetracks review from the relevant stuff.

Please focus efforts on getting a proper dialog finished while respecting basic usability rules, the latter which seems to have been the biggest unsolved issue in all approaches that were brought up so far.

bunnei:

bunnei:
It's been a month since my last comment. @Kloen, do you plan on addressing any of the aforementioned concerns? I plan on closing the PR if I do not hear anything within the next few days.

bunnei:
Closing, per my previous comments. Feel free to submit a new PR if you'd like to further pursue this feature.

PR #1121, opened 2015-09-07
Shader: Use constants and proper type casts

aroulin:
- Change return type of register offsets functions from `int` to `size_t` to decouple them from the signed x64 displacement and adds checks to make sure any needed cast is well-defined.
- Use emitter's SCALE_1 constant.

I think it is clearer to do the distinction between an unsigned register offset a base register block and the x64 displacement which can be signed although this can be argued.

### DIFF
--- a/src/video_core/shader/shader.h
+++ b/src/video_core/shader/shader.h
@@ -289,13 +289,13 @@ struct UnitState {
 
     DebugData<Debug> debug;
 
-    static int InputOffset(const SourceRegister& reg) {
+    static size_t InputOffset(const SourceRegister& reg) {
         switch (reg.GetRegisterType()) {
         case RegisterType::Input:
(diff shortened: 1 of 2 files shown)


src/video_core/shader/shader.h
```diff
@@ -289,13 +289,13 @@ struct UnitState {
 
     DebugData<Debug> debug;
 
-    static int InputOffset(const SourceRegister& reg) {
+    static size_t InputOffset(const SourceRegister& reg) {
         switch (reg.GetRegisterType()) {
         case RegisterType::Input:
-            return (int)offsetof(UnitState::Registers, input) + reg.GetIndex()*sizeof(Math::Vec4<float24>);
```
aroulin: In any case, these casts seems useless so it should be fixed on way or another (return type `int` or `size_t`)

bunnei:
LGTM

PR #1122, opened 2015-09-07
gdbstub implementation

polaris-:
I used Dolphin's gdbstub for this one. As per the guidelines, I changed the C-style casts to C++-style casts and put everything I planned to expose behind the Gdbstub namespace, among other things. In addition, I changed the way breakpoints are stored and accessed from using arrays to unordered_maps as per Subv's suggestion on IRC, which should open up the possibility for better access times and more than 10 breakpoints per type.

Features:
- Read, write, access (r/w), and execution breakpoints
- Read/modify registers
- Read/modify memory
- Stepping
- Suspending

Using IDA Pro to connect to the gdbstub also works well. Things like step over and set IP also work. 

The default port is 24689 which can be modified in the config ini files. Set the USE_GDBSTUB preprocessor flag in citra, citra-qt, and core to compile. If you don't set it in at least citra or citra-qt then it won't be able to read the port from the config and will default to 24689.

Since I was asked on IRC why everything is behind ifdefs, I will offer my justification here so everyone can read it. There is definitely going to be a performance hit when using the gdbstub since it is checking for breakpoints every instruction and memory read/write, so I don't think would be appropriate to include the code in any kind of build unless specifically wanted. Additionally, since it's on the same thread as the CPU, it will block while waiting for a gdb client to connect and send the start execution command when you load a ROM file. The good thing is that it allows you to start debugging from the very beginning of execution, but the blocking thing just isn't acceptable for normal builds.

Another change I had to make is to how CPSR is handled in the interpreter. During normal execution, the T flag (bit 5) of CPSR is never set even if cpu->TFlag is 1 and it's executing Thumb code. IDA struggles to determine if it should be reading code as ARM or Thumb without the T flag being set and the disassembly becomes a mess as a result. In order to fix that, I added code at the beginning of the GDB_BP_CHECK macro to clear and set the T flag of CPSR. According to Lioncash, the T flag not being set in CPSR is intended behavior and a leftover from the initial SkyEye import so I didn't open an issue about it, but I can if that's something that should be fixed independently of the gdbstub.

### DIFF
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -30,6 +30,8 @@
 
 #include "video_core/video_core.h"
 
+#include "core/gdbstub/gdbstub.h"
+
 
 static void PrintHelp()
 {
@@ -72,6 +74,7 @@ int main(int argc, char **argv) {
     Config config;
     log_filter.ParseFilterString(Settings::values.log_filter);
 
+    GDBStub::SetServerPort(static_cast<u32>(Settings::values.gdbstub_port));
 
     EmuWindow_GLFW* emu_window = new EmuWindow_GLFW;
 
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -23,14 +23,13 @@
 #include "core/settings.h"
 #include "core/system.h"
 #include "core/core.h"
+#include "core/gdbstub/gdbstub.h"
 #include "core/loader/loader.h"
 
 #include "citra/config.h"
 #include "citra/emu_window/emu_window_glfw.h"
 
 #include "video_core/video_core.h"
-
-#include "core/gdbstub/gdbstub.h"
 
 
 static void PrintHelp()
--- a/src/citra/config.cpp
+++ b/src/citra/config.cpp
@@ -76,7 +76,7 @@ void Config::ReadValues() {
     // Miscellaneous
     Settings::values.log_filter = glfw_config->Get("Miscellaneous", "log_filter", "*:Info");
 
-    // GDBStubebugging
+    // Debugging
     Settings::values.use_gdbstub = glfw_config->GetBoolean("Debugging", "use_gdbstub", false);
     Settings::values.gdbstub_port = glfw_config->GetInteger("Debugging", "gdbstub_port", 24689);
 }
--- a/src/citra/config.cpp
+++ b/src/citra/config.cpp
@@ -75,6 +75,10 @@ void Config::ReadValues() {
 
     // Miscellaneous
     Settings::values.log_filter = glfw_config->Get("Miscellaneous", "log_filter", "*:Info");
+
+    // GDBStubebugging
+    Settings::values.use_gdbstub = glfw_config->GetBoolean("Debugging", "use_gdbstub", false);
+    Settings::values.gdbstub_port = glfw_config->GetInteger("Debugging", "gdbstub_port", 24689);
 }
 
 void Config::Reload() {
--- a/src/citra/default_ini.h
+++ b/src/citra/default_ini.h
@@ -66,6 +66,11 @@ region_value =
 # A filter which removes logs below a certain logging level.
 # Examples: *:Debug Kernel.SVC:Trace Service.*:Critical
 log_filter = *:Info
+
+[Debugging]
+# Port for listening to GDB connections.
+use_gdbstub=false
+gdbstub_port=24689
 )";
 
 }
--- a/src/citra_qt/config.cpp
+++ b/src/citra_qt/config.cpp
@@ -62,6 +62,11 @@ void Config::ReadValues() {
     qt_config->beginGroup("Miscellaneous");
     Settings::values.log_filter = qt_config->value("log_filter", "*:Info").toString().toStdString();
     qt_config->endGroup();
+
+    qt_config->beginGroup("Debugging");
+    Settings::values.use_gdbstub = qt_config->value("use_gdbstub", false).toBool();
+    Settings::values.gdbstub_port = qt_config->value("gdbstub_port", 24689).toInt();
+    qt_config->endGroup();
 }
 
 void Config::SaveValues() {
@@ -96,6 +101,11 @@ void Config::SaveValues() {
 
     qt_config->beginGroup("Miscellaneous");
     qt_config->setValue("log_filter", QString::fromStdString(Settings::values.log_filter));
+    qt_config->endGroup();
+
+    qt_config->beginGroup("Debugging");
+    qt_config->setValue("use_gdbstub", Settings::values.use_gdbstub);
+    qt_config->setValue("gdbstub_port", Settings::values.gdbstub_port);
     qt_config->endGroup();
 }
 
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -48,6 +48,8 @@
 
 #include "video_core/video_core.h"
 
+#include "core/gdbstub/gdbstub.h"
+
 GMainWindow::GMainWindow() : emu_thread(nullptr)
 {
     Pica::g_debug_context = Pica::DebugContext::Construct();
@@ -143,6 +145,11 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
 
     game_list->LoadInterfaceLayout(settings);
 
+    ui.action_Use_Gdbstub->setChecked(Settings::values.use_gdbstub);
+    SetGdbstubEnabled(ui.action_Use_Gdbstub->isChecked());
+
+    GDBStub::SetServerPort(static_cast<u32>(Settings::values.gdbstub_port));
+
     ui.action_Use_Hardware_Renderer->setChecked(Settings::values.use_hw_renderer);
     SetHardwareRendererEnabled(ui.action_Use_Hardware_Renderer->isChecked());
 
@@ -175,6 +182,7 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     connect(ui.action_Stop, SIGNAL(triggered()), this, SLOT(OnStopGame()));
     connect(ui.action_Use_Hardware_Renderer, SIGNAL(triggered(bool)), this, SLOT(SetHardwareRendererEnabled(bool)));
     connect(ui.action_Use_Shader_JIT, SIGNAL(triggered(bool)), this, SLOT(SetShaderJITEnabled(bool)));
+    connect(ui.action_Use_Gdbstub, SIGNAL(triggered(bool)), this, SLOT(SetGdbstubEnabled(bool)));
     connect(ui.action_Single_Window_Mode, SIGNAL(triggered(bool)), this, SLOT(ToggleWindowMode()));
     connect(ui.action_Hotkeys, SIGNAL(triggered()), this, SLOT(OnOpenHotkeysDialog()));
 
@@ -443,6 +451,10 @@ void GMainWindow::OnOpenHotkeysDialog() {
 
 void GMainWindow::SetHardwareRendererEnabled(bool enabled) {
     VideoCore::g_hw_renderer_enabled = enabled;
+}
+
+void GMainWindow::SetGdbstubEnabled(bool enabled) {
+    GDBStub::ToggleServer(enabled);
 }
 
 void GMainWindow::SetShaderJITEnabled(bool enabled) {
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -44,11 +44,10 @@
 #include "core/settings.h"
 #include "core/system.h"
 #include "core/arm/disassembler/load_symbol_map.h"
+#include "core/gdbstub/gdbstub.h"
 #include "core/loader/loader.h"
 
 #include "video_core/video_core.h"
-
-#include "core/gdbstub/gdbstub.h"
 
 GMainWindow::GMainWindow() : emu_thread(nullptr)
 {
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -29,6 +29,7 @@ namespace Log {
         SUB(Debug, Emulated) \
         SUB(Debug, GPU) \
         SUB(Debug, Breakpoint) \
+        SUB(Debug, GDBStub) \
         CLS(Kernel) \
         SUB(Kernel, SVC) \
         CLS(Service) \
--- a/src/common/logging/log.h
+++ b/src/common/logging/log.h
@@ -43,6 +43,7 @@ enum class Class : ClassType {
     Debug_Emulated,             ///< Debug messages from the emulated programs
     Debug_GPU,                  ///< GPU debugging tools
     Debug_Breakpoint,           ///< Logging breakpoints and watchpoints
+    Debug_GDBStub,              ///< GDB Stub
     Kernel,                     ///< The HLE implementation of the CTR kernel
     Kernel_SVC,                 ///< Kernel system calls
     Service,                    ///< HLE implementation of system services. Each major service
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -22,6 +22,7 @@ set(SRCS
             file_sys/archive_systemsavedata.cpp
             file_sys/disk_archive.cpp
             file_sys/ivfc_archive.cpp
+            gdbstub/gdbstub.cpp
             hle/config_mem.cpp
             hle/hle.cpp
             hle/applets/applet.cpp
@@ -149,6 +150,7 @@ set(HEADERS
             file_sys/disk_archive.h
             file_sys/file_backend.h
             file_sys/ivfc_archive.h
+            gdbstub/gdbstub.h
             hle/config_mem.h
             hle/function_wrappers.h
             hle/hle.h
--- a/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
@@ -23,6 +23,8 @@
 #include "core/arm/skyeye_common/armsupp.h"
 #include "core/arm/skyeye_common/vfp/vfp.h"
 
+#include "core/gdbstub/gdbstub.h"
+
 Common::Profiling::TimingCategory profile_execute("DynCom::Execute");
 Common::Profiling::TimingCategory profile_decode("DynCom::Decode");
 
@@ -3548,6 +3550,7 @@ static int InterpreterTranslate(ARMul_State* cpu, int& bb_start, u32 addr) {
             CITRA_IGNORE_EXIT(-1);
         }
         inst_base = arm_instruction_trans[idx](inst, idx);
+
 translated:
         phys_addr += inst_size;
 
@@ -3580,6 +3583,8 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     Common::Profiling::ScopeTimer timer_execute(profile_execute);
     MICROPROFILE_SCOPE(DynCom_Execute);
 
+    int breakpoint_offset = -1;
+
     #undef RM
     #undef RS
 
@@ -3604,15 +3609,27 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     #define INC_PC(l)   ptr += sizeof(arm_inst) + l
     #define INC_PC_STUB ptr += sizeof(arm_inst)
 
+#define GDB_BP_CHECK \
+    cpu->Cpsr &= ~(1 << 5); \
+    cpu->Cpsr |= cpu->TFlag << 5; \
+    if (GDBStub::g_server_enabled) { \
+        if (GDBStub::IsMemoryBreak() || PC == breakpoint_offset) { \
+            GDBStub::Break(); \
+            goto END; \
+        } \
+    }
+
 // GCC and Clang have a C++ extension to support a lookup table of labels. Otherwise, fallback to a
 // clunky switch statement.
 #if defined __GNUC__ || defined __clang__
 #define GOTO_NEXT_INST \
+    GDB_BP_CHECK; \
     if (num_instrs >= cpu->NumInstrsToExecute) goto END; \
     num_instrs++; \
     goto *InstLabel[inst_base->idx]
 #else
 #define GOTO_NEXT_INST \
+    GDB_BP_CHECK; \
     if (num_instrs >= cpu->NumInstrsToExecute) goto END; \
     num_instrs++; \
     switch(inst_base->idx) { \
@@ -3878,6 +3895,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     unsigned int addr;
     unsigned int num_instrs = 0;
 
+
     int ptr;
 
     LOAD_NZCVT;
@@ -3901,6 +3919,11 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
         } else {
             if (InterpreterTranslate(cpu, ptr, cpu->Reg[15]) == FETCH_EXCEPTION)
                 goto END;
+        }
+
+        // Find breakpoint if one exists within the block
+        if (GDBStub::g_server_enabled && GDBStub::IsConnected()) {
+            breakpoint_offset = GDBStub::GetNextBreakpointFromAddress(cpu->Reg[15], GDBStub::BreakpointType::Execute);
         }
 
         inst_base = (arm_inst *)&inst_buf[ptr];
@@ -4454,7 +4477,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
             ldst_inst* inst_cream = (ldst_inst*)inst_base->component;
             inst_cream->get_addr(cpu, inst_cream->inst, addr);
 
-            cpu->Reg[BITS(inst_cream->inst, 12, 15)] = Memory::Read8(addr);
+            cpu->Reg[BITS(inst_cream->inst, 12, 15)] = cpu->ReadMemory8(addr);
 
             if (BITS(inst_cream->inst, 12, 15) == 15) {
                 INC_PC(sizeof(ldst_inst));
@@ -4472,7 +4495,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
             ldst_inst* inst_cream = (ldst_inst*)inst_base->component;
             inst_cream->get_addr(cpu, inst_cream->inst, addr);
 
-            cpu->Reg[BITS(inst_cream->inst, 12, 15)] = Memory::Read8(addr);
+            cpu->Reg[BITS(inst_cream->inst, 12, 15)] = cpu->ReadMemory8(addr);
 
             if (BITS(inst_cream->inst, 12, 15) == 15) {
                 INC_PC(sizeof(ldst_inst));
@@ -4531,7 +4554,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
 
             cpu->SetExclusiveMemoryAddress(read_addr);
 
-            RD = Memory::Read8(read_addr);
+            RD = cpu->ReadMemory8(read_addr);
             if (inst_cream->Rd == 15) {
                 INC_PC(sizeof(generic_arm_inst));
                 goto DISPATCH;
@@ -4604,7 +4627,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
         if (inst_base->cond == ConditionCode::AL || CondPassed(cpu, inst_base->cond)) {
             ldst_inst* inst_cream = (ldst_inst*)inst_base->component;
             inst_cream->get_addr(cpu, inst_cream->inst, addr);
-            unsigned int value = Memory::Read8(addr);
+            unsigned int value = cpu->ReadMemory8(addr);
             if (BIT(value, 7)) {
                 value |= 0xffffff00;
             }
@@ -6027,7 +6050,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
             ldst_inst* inst_cream = (ldst_inst*)inst_base->component;
             inst_cream->get_addr(cpu, inst_cream->inst, addr);
             unsigned int value = cpu->Reg[BITS(inst_cream->inst, 12, 15)] & 0xff;
-            Memory::Write8(addr, value);
+            cpu->WriteMemory8(addr, value);
         }
         cpu->Reg[15] += cpu->GetInstructionSize();
         INC_PC(sizeof(ldst_inst));
@@ -6040,7 +6063,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
             ldst_inst* inst_cream = (ldst_inst*)inst_base->component;
             inst_cream->get_addr(cpu, inst_cream->inst, addr);
             unsigned int value = cpu->Reg[BITS(inst_cream->inst, 12, 15)] & 0xff;
-            Memory::Write8(addr, value);
+            cpu->WriteMemory8(addr, value);
         }
         cpu->Reg[15] += cpu->GetInstructionSize();
         INC_PC(sizeof(ldst_inst));
@@ -6091,7 +6114,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
 
             if (cpu->IsExclusiveMemoryAccess(write_addr)) {
                 cpu->UnsetExclusiveMemoryAddress();
-                Memory::Write8(write_addr, cpu->Reg[inst_cream->Rm]);
+                cpu->WriteMemory8(write_addr, cpu->Reg[inst_cream->Rm]);
                 RD = 0;
             } else {
                 // Failed to write due to mutex access
@@ -6250,8 +6273,8 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
         if (inst_base->cond == ConditionCode::AL || CondPassed(cpu, inst_base->cond)) {
             swp_inst* inst_cream = (swp_inst*)inst_base->component;
             addr = RN;
-            unsigned int value = Memory::Read8(addr);
-            Memory::Write8(addr, (RM & 0xFF));
+            unsigned int value = cpu->ReadMemory8(addr);
+            cpu->WriteMemory8(addr, (RM & 0xFF));
             RD = value;
         }
         cpu->Reg[15] += cpu->GetInstructionSize();
--- a/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
@@ -3895,7 +3895,6 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     unsigned int addr;
     unsigned int num_instrs = 0;
 
-
     int ptr;
 
     LOAD_NZCVT;
--- a/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
@@ -3583,7 +3583,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     Common::Profiling::ScopeTimer timer_execute(profile_execute);
     MICROPROFILE_SCOPE(DynCom_Execute);
 
-    int breakpoint_offset = -1;
+    GDBStub::BreakpointAddress breakpoint_data;
 
     #undef RM
     #undef RS
@@ -3613,7 +3613,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     cpu->Cpsr &= ~(1 << 5); \
     cpu->Cpsr |= cpu->TFlag << 5; \
     if (GDBStub::g_server_enabled) { \
-        if (GDBStub::IsMemoryBreak() || PC == breakpoint_offset) { \
+        if (GDBStub::IsMemoryBreak() || (breakpoint_data.type != GDBStub::BreakpointType::None && PC == breakpoint_data.address)) { \
             GDBStub::Break(); \
             goto END; \
         } \
@@ -3923,7 +3923,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
 
         // Find breakpoint if one exists within the block
         if (GDBStub::g_server_enabled && GDBStub::IsConnected()) {
-            breakpoint_offset = GDBStub::GetNextBreakpointFromAddress(cpu->Reg[15], GDBStub::BreakpointType::Execute);
+            breakpoint_data = GDBStub::GetNextBreakpointFromAddress(cpu->Reg[15], GDBStub::BreakpointType::Execute);
         }
 
         inst_base = (arm_inst *)&inst_buf[ptr];
--- a/src/core/arm/skyeye_common/armstate.cpp
+++ b/src/core/arm/skyeye_common/armstate.cpp
@@ -7,6 +7,7 @@
 #include "core/memory.h"
 #include "core/arm/skyeye_common/armstate.h"
 #include "core/arm/skyeye_common/vfp/vfp.h"
+#include "core/gdbstub/gdbstub.h"
 
 ARMul_State::ARMul_State(PrivilegeMode initial_mode)
 {
@@ -185,8 +186,25 @@ void ARMul_State::ResetMPCoreCP15Registers()
     CP15[CP15_TLB_DEBUG_CONTROL] = 0x00000000;
 }
 
+static void CheckMemoryBreakpoint(u32 address, GDBStub::BreakpointType type)
+{
+    if (GDBStub::g_server_enabled && GDBStub::CheckBreakpoint(address, type)) {
+        LOG_DEBUG(Debug, "Found memory breakpoint @ %08x", address);
+        GDBStub::Break(true);
+    }
+}
+
+u8 ARMul_State::ReadMemory8(u32 address) const
+{
+    CheckMemoryBreakpoint(address, GDBStub::BreakpointType::Read);
+
+    return Memory::Read8(address);
+}
+
 u16 ARMul_State::ReadMemory16(u32 address) const
 {
+    CheckMemoryBreakpoint(address, GDBStub::BreakpointType::Read);
+
     u16 data = Memory::Read16(address);
 
     if (InBigEndianMode())
@@ -197,6 +215,8 @@ u16 ARMul_State::ReadMemory16(u32 address) const
 
 u32 ARMul_State::ReadMemory32(u32 address) const
 {
+    CheckMemoryBreakpoint(address, GDBStub::BreakpointType::Read);
+
     u32 data = Memory::Read32(address);
 
     if (InBigEndianMode())
@@ -207,6 +227,8 @@ u32 ARMul_State::ReadMemory32(u32 address) const
 
 u64 ARMul_State::ReadMemory64(u32 address) const
 {
+    CheckMemoryBreakpoint(address, GDBStub::BreakpointType::Read);
+
     u64 data = Memory::Read64(address);
 
     if (InBigEndianMode())
@@ -215,8 +237,17 @@ u64 ARMul_State::ReadMemory64(u32 address) const
     return data;
 }
 
+void ARMul_State::WriteMemory8(u32 address, u8 data)
+{
+    CheckMemoryBreakpoint(address, GDBStub::BreakpointType::Write);
+
+    Memory::Write8(address, data);
+}
+
 void ARMul_State::WriteMemory16(u32 address, u16 data)
 {
+    CheckMemoryBreakpoint(address, GDBStub::BreakpointType::Write);
+
     if (InBigEndianMode())
         data = Common::swap16(data);
 
@@ -225,6 +256,8 @@ void ARMul_State::WriteMemory16(u32 address, u16 data)
 
 void ARMul_State::WriteMemory32(u32 address, u32 data)
 {
+    CheckMemoryBreakpoint(address, GDBStub::BreakpointType::Write);
+
     if (InBigEndianMode())
         data = Common::swap32(data);
 
@@ -233,6 +266,8 @@ void ARMul_State::WriteMemory32(u32 address, u32 data)
 
 void ARMul_State::WriteMemory64(u32 address, u64 data)
 {
+    CheckMemoryBreakpoint(address, GDBStub::BreakpointType::Write);
+
     if (InBigEndianMode())
         data = Common::swap64(data);
 
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -13,6 +13,8 @@
 #include "core/hle/kernel/thread.h"
 #include "core/hw/hw.h"
 
+#include "core/gdbstub/gdbstub.h"
+
 namespace Core {
 
 ARM_Interface*     g_app_core = nullptr;  ///< ARM11 application core
@@ -20,6 +22,21 @@ ARM_Interface*     g_sys_core = nullptr;  ///< ARM11 system (OS) core
 
 /// Run the core CPU loop
 void RunLoop(int tight_loop) {
+    if (GDBStub::g_server_enabled) {
+        GDBStub::HandlePacket();
+
+        // If the loop is halted and we want to step, use a tiny (1) number of instructions to execute.
+        // Otherwise get out of the loop function.
+        if (GDBStub::GetCpuHaltFlag()) {
+            if (GDBStub::GetCpuStepFlag()) {
+                GDBStub::SetCpuStepFlag(false);
+                tight_loop = 1;
+            } else {
+                return;
+            }
+        }
+    }
+
     // If we don't have a currently active thread then don't execute instructions,
     // instead advance to the next event and try to yield to the next thread
     if (Kernel::GetCurrentThread() == nullptr) {
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -5,12 +5,12 @@
 // Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
 
 #include <algorithm>
+#include <climits>
 #include <csignal>
 #include <cstdarg>
 #include <cstdio>
 #include <cstring>
 #include <fcntl.h>
-#include <limits.h>
 #include <map>
 #include <numeric>
 
@@ -34,7 +34,7 @@
 #include "core/core.h"
 #include "core/memory.h"
 #include "core/arm/arm_interface.h"
-#include "gdbstub.h"
+#include "core/gdbstub/gdbstub.h"
 
 const int GDB_BUFFER_SIZE = 10000;
 
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -473,6 +473,7 @@ static void ReadRegister() {
         IntToHex(reply, Core::g_app_core->GetVFPReg(id - CSPR_REGISTER - 1)); // VFP registers should start at 26, so one after CSPR_REGISTER
     } else if (id == FPSCR_REGISTER) {
         IntToHex(reply, Core::g_app_core->GetVFPSystemReg(VFP_FPSCR)); // Get FPSCR
+        IntToHex(reply + 8, 0);
     } else {
         return SendReply("E01");
     }
@@ -497,6 +498,8 @@ static void ReadRegisters() {
             i++; // These registers seem to be all 64bit instead of 32bit, so skip two instead of one
         } else if (i > CSPR_REGISTER && i < FPSCR_REGISTER) {
             IntToHex(bufptr + i * 8, Core::g_app_core->GetVFPReg(i - CSPR_REGISTER - 1));
+            IntToHex(bufptr + (i + 1) * 8, 0);
+            i++;
         } else if (i == FPSCR_REGISTER) {
             IntToHex(bufptr + i * 8, Core::g_app_core->GetVFPSystemReg(VFP_FPSCR));
         }
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -231,13 +231,18 @@ static void RemoveBreakpoint(BreakpointType type, PAddr addr) {
     }
 }
 
-PAddr GetNextBreakpointFromAddress(PAddr addr, BreakpointType type) {
+BreakpointAddress GetNextBreakpointFromAddress(PAddr addr, BreakpointType type) {
     std::map<u32, Breakpoint>& p = GetBreakpointList(type);
     auto next_breakpoint = p.lower_bound(addr);
-    u32 breakpoint = -1;
+    BreakpointAddress breakpoint;
 
-    if (next_breakpoint != p.end())
-        breakpoint = next_breakpoint->first;
+    if (next_breakpoint != p.end()) {
+        breakpoint.address = next_breakpoint->first;
+        breakpoint.type = type;
+    } else {
+        breakpoint.address = 0;
+        breakpoint.type = BreakpointType::None;
+    }
 
     return breakpoint;
 }
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -476,20 +476,20 @@ static void ReadRegisters() {
     u8* bufptr = buffer;
     for (int i = 0, reg = 0; i <= MAX_REGISTERS; i++, reg++) {
         if (i <= R15_REGISTER) {
-            IntToHex(bufptr + i * 8, Core::g_app_core->GetReg(reg));
+            IntToHex(bufptr + i * CHAR_BIT, Core::g_app_core->GetReg(reg));
         } else if (i == CSPR_REGISTER) {
-            IntToHex(bufptr + i * 8, Core::g_app_core->GetCPSR());
+            IntToHex(bufptr + i * CHAR_BIT, Core::g_app_core->GetCPSR());
         } else if (i < CSPR_REGISTER) {
-            IntToHex(bufptr + i * 8, 0);
-            IntToHex(bufptr + (i + 1) * 8, 0);
+            IntToHex(bufptr + i * CHAR_BIT, 0);
+            IntToHex(bufptr + (i + 1) * CHAR_BIT, 0);
             i++; // These registers seem to be all 64bit instead of 32bit, so skip two instead of one
             reg++;
         } else if (i > CSPR_REGISTER && i < MAX_REGISTERS) {
-            IntToHex(bufptr + i * 8, Core::g_app_core->GetVFPReg(reg - CSPR_REGISTER - 1));
-            IntToHex(bufptr + (i + 1) * 8, 0);
+            IntToHex(bufptr + i * CHAR_BIT, Core::g_app_core->GetVFPReg(reg - CSPR_REGISTER - 1));
+            IntToHex(bufptr + (i + 1) * CHAR_BIT, 0);
             i++;
         } else if (i == MAX_REGISTERS) {
-            IntToHex(bufptr + i * 8, Core::g_app_core->GetVFPSystemReg(VFP_FPSCR));
+            IntToHex(bufptr + i * CHAR_BIT, Core::g_app_core->GetVFPSystemReg(VFP_FPSCR));
         }
     }
 
@@ -531,17 +531,17 @@ static void WriteRegisters() {
 
     for (int i = 0, reg = 0; i <= MAX_REGISTERS; i++, reg++) {
         if (i <= R15_REGISTER) {
-            Core::g_app_core->SetReg(reg, HexToInt(buffer_ptr + i * 8));
+            Core::g_app_core->SetReg(reg, HexToInt(buffer_ptr + i * CHAR_BIT));
         } else if (i == CSPR_REGISTER) {
-            Core::g_app_core->SetCPSR(HexToInt(buffer_ptr + i * 8));
+            Core::g_app_core->SetCPSR(HexToInt(buffer_ptr + i * CHAR_BIT));
         } else if (i < CSPR_REGISTER) {
             i++; // These registers seem to be all 64bit instead of 32bit, so skip two instead of one
             reg++;
         } else if (i > CSPR_REGISTER && i < MAX_REGISTERS) {
-            Core::g_app_core->SetVFPReg(reg - CSPR_REGISTER - 1, HexToInt(buffer_ptr + i * 8));
+            Core::g_app_core->SetVFPReg(reg - CSPR_REGISTER - 1, HexToInt(buffer_ptr + i * CHAR_BIT));
             i++; // Skip padding
         } else if (i == MAX_REGISTERS) {
-            Core::g_app_core->SetVFPSystemReg(VFP_FPSCR, HexToInt(buffer_ptr + i * 8));
+            Core::g_app_core->SetVFPSystemReg(VFP_FPSCR, HexToInt(buffer_ptr + i * CHAR_BIT));
         }
     }
 
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -186,7 +186,7 @@ static u8 ReadByte() {
     size_t received_size = recv(gdbserver_socket, reinterpret_cast<char*>(&c), 1, MSG_WAITALL);
     if (received_size != 1) {
         LOG_ERROR(Debug_GDBStub, "recv failed : %ld", received_size);
-        Deinit();
+        Shutdown();
     }
 
     return c;
@@ -322,7 +322,7 @@ static void SendReply(const char* reply) {
         int sent_size = send(gdbserver_socket, reinterpret_cast<char*>(ptr), left, 0);
         if (sent_size < 0) {
             LOG_ERROR(Debug_GDBStub, "gdb: send failed");
-            return Deinit();
+            return Shutdown();
         }
 
         left -= sent_size;
@@ -773,7 +773,7 @@ void HandlePacket() {
         HandleSignal();
         break;
     case 'k':
-        Deinit();
+        Shutdown();
         LOG_INFO(Debug_GDBStub, "killed by gdb");
         return;
     case 'g':
@@ -829,7 +829,7 @@ void ToggleServer(bool status) {
     else {
         // Stop server
         if (IsConnected()) {
-            Deinit();
+            Shutdown();
         }
 
         g_server_enabled = status;
@@ -908,7 +908,7 @@ void Init() {
     Init(gdbstub_port);
 }
 
-void Deinit() {
+void Shutdown() {
     if (!g_server_enabled) {
         return;
     }
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -4,11 +4,13 @@
 
 // Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
 
+#include <algorithm>
 #include <csignal>
 #include <cstdarg>
 #include <cstdio>
 #include <cstring>
 #include <fcntl.h>
+#include <limits.h>
 #include <map>
 #include <numeric>
 
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -488,17 +488,18 @@ static void ReadRegisters() {
     memset(buffer, 0, sizeof(buffer));
 
     u8* bufptr = buffer;
-    for (int i = 0; i <= MAX_REGISTERS; i++) {
+    for (int i = 0, reg = 0; i <= MAX_REGISTERS; i++, reg++) {
         if (i <= R15_REGISTER) {
-            IntToHex(bufptr + i * 8, Core::g_app_core->GetReg(i));
+            IntToHex(bufptr + i * 8, Core::g_app_core->GetReg(reg));
         } else if (i == CSPR_REGISTER) {
             IntToHex(bufptr + i * 8, Core::g_app_core->GetCPSR());
         } else if (i < CSPR_REGISTER) {
             IntToHex(bufptr + i * 8, 0);
             IntToHex(bufptr + (i + 1) * 8, 0);
             i++; // These registers seem to be all 64bit instead of 32bit, so skip two instead of one
+            reg++;
         } else if (i > CSPR_REGISTER && i < MAX_REGISTERS) {
-            IntToHex(bufptr + i * 8, Core::g_app_core->GetVFPReg(i - CSPR_REGISTER - 1));
+            IntToHex(bufptr + i * 8, Core::g_app_core->GetVFPReg(reg - CSPR_REGISTER - 1));
             IntToHex(bufptr + (i + 1) * 8, 0);
             i++;
         } else if (i == MAX_REGISTERS) {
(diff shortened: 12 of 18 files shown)


src/citra/default_ini.h
```diff
@@ -66,6 +66,9 @@ region_value =
 # A filter which removes logs below a certain logging level.
 # Examples: *:Debug Kernel.SVC:Trace Service.*:Critical
 log_filter = *:Info
+
+[Gdbstub]
+gdbstub_port=24689
```
Subv: This needs a comment

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3878,10 +3907,11 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     unsigned int addr;
     unsigned int num_instrs = 0;
 
+
     int ptr;
 
     LOAD_NZCVT;
-    DISPATCH:
+DISPATCH:
```
Subv: indenting

src/core/core.cpp
```diff
@@ -20,6 +24,27 @@ ARM_Interface*     g_sys_core = nullptr;  ///< ARM11 system (OS) core
...
+    }
+
+    // If the loop is halted and we want to step, use a tiny (1) number of instructions to execute.
+    // Otherwise get out of the loop function.
+    if (Gdbstub::halt_loop == true)
+    {
+        if (Gdbstub::step_loop == true) {
+            Gdbstub::step_loop = false;
+            tight_loop = 1;
+        }
```
Subv: else on the same line as }

src/core/core.cpp
```diff
@@ -20,6 +24,27 @@ ARM_Interface*     g_sys_core = nullptr;  ///< ARM11 system (OS) core
...
+        else {
+            //Gdbstub::halt_loop = false;
```
Subv: Please don't leave dead code in comments

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    u8 *ptr = cmd_bfr;
+    u8 c = 0;
+
+    while (len-- > 0)
+        c += *ptr++;
+
+    return c;
+}
+
+static std::unordered_map<u32, gdb_bp_t> &GetBreakpointList(u32 type)
```
Subv: Place the & and *s next to the type, not the function/variable name

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    case Gdbstub::Access:
+        return breakpoints_access;
+    default:
+        return breakpoints_read;
+    }
+}
+
+static void RemoveBreakpoint(u32 type, u32 addr, u32 len)
+{
+    std::unordered_map<u32, gdb_bp_t> &p = GetBreakpointList(type);
```
Subv: Same, please change this everywhere else.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+static int CheckBreakpoint(u32 addr, u32 type)
+{
+    std::unordered_map<u32, gdb_bp_t> &p = GetBreakpointList(type);
+
+    std::unordered_map<u32, gdb_bp_t>::const_iterator bp = p.find(addr);
+    if (bp != p.end()) {
+        u32 len = bp->second.len;
+
+        if (type == Gdbstub::Execute) {
+            len = 1;
```
Subv: Please add a comment explaining why this happens

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    std::unordered_map<u32, gdb_bp_t> &p = GetBreakpointList(type);
+
+    std::unordered_map<u32, gdb_bp_t>::const_iterator bp = p.find(addr);
+    if (bp != p.end()) {
+        LOG_DEBUG(GdbStub, "gdb: removed a breakpoint: %08x bytes at %08x\n", len, addr);
+        p.erase(addr);
+    }
+}
+
+static int CheckBreakpoint(u32 addr, u32 type)
```
Subv: Shouldn't this return bool?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    SendReply("E01");
+}
+
+static void HandleSignal()
+{
+    char bfr[128];
+
+    memset(bfr, 0, sizeof bfr);
+
+    sprintf(bfr, "T%02x%02x:%08x;%02x:%08x;", sig, 15, htonl(Core::g_app_core->GetPC()), 13, htonl(Core::g_app_core->GetReg(13)));
```
Subv: "The htonl function converts a u_long from host to TCP/IP network byte order (which is big-endian)." Huh?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    u8 * bufptr = bfr;
+    u32 i;
+
+    memset(bfr, 0, sizeof bfr);
+
+    for (i = 0; i < 42; i++)
+    {
+        if (i < 16) {
+            Wbe32Hex(bufptr + i * 8, Core::g_app_core->GetReg(i));
+        }
```
Subv: elses on the same line as }

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    bp.len = len;
+
+    p.insert({ addr, bp });
+
+    LOG_DEBUG(GdbStub, "gdb: added %d breakpoint: %08x bytes at %08x\n", type, bp.len, bp.addr);
+
+    return true;
+}
+
+static void _AddBreakpoint()
```
Subv: AddBreakpointInternal?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+{
+    Gdbstub::Init(gdbstub_port);
+}
+
+void Gdbstub::Init(u32 port)
+{
+    sockaddr_in saddr_server = {};
+    sockaddr_in saddr_client;
+
+
```
Subv: extra empty line

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    InitGeneric(PF_INET,
+        reinterpret_cast<const sockaddr *>(&saddr_server), sizeof(saddr_server),
+        reinterpret_cast<sockaddr *>(&saddr_client), &client_addrlen);
+
+
+    LOG_INFO(GdbStub, "Made connection on GDB server on port %d...", port);
+
+    saddr_client.sin_addr.s_addr = ntohl(saddr_client.sin_addr.s_addr);
+    /*if (((saddr_client.sin_addr.s_addr >> 24) & 0xff) != 127 ||
```
Subv: dead code?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+
+    saddr_server.sin_family = AF_INET;
+    saddr_server.sin_port = htons(port);
+    saddr_server.sin_addr.s_addr = INADDR_ANY;
+
+    socklen_t client_addrlen = sizeof(saddr_client);
+
+    InitGeneric(PF_INET,
+        reinterpret_cast<const sockaddr *>(&saddr_server), sizeof(saddr_server),
+        reinterpret_cast<sockaddr *>(&saddr_client), &client_addrlen);
```
Subv: Place the * next to the type

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+        LOG_ERROR(GdbStub, "Failed to listen to gdb socket");
+
+    LOG_INFO(GdbStub, "Waiting for gdb to connect...\n");
+
+    sock = accept(tmpsock, client_addr, client_addrlen);
+    if (sock < 0)
+        LOG_ERROR(GdbStub, "Failed to accept gdb client");
+    LOG_INFO(GdbStub, "Client connected.\n");
+
+    //close(tmpsock);
```
Subv: dead code?

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+// Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
+
+#ifdef USE_GDBSTUB
+#pragma once
+
+namespace Gdbstub {
+    typedef enum {
```
Subv: use an enum class here

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,40 @@
...
+    void Deinit();
+    bool IsActive();
+    void Break();
+    void MemoryBreak();
+
+    int IsMemoryBreak();
+
+    void HandlePacket();
+
+    int  CheckBreakpointExecute(u32 addr);
```
Subv: don't align the function names like this

src/core/memory.cpp
```diff
@@ -92,7 +96,18 @@ void UnmapRegion(VAddr base, u32 size) {
 }
 
 template <typename T>
-T Read(const VAddr vaddr) {
+T Read(const VAddr vaddr, bool bp_check) {
+#ifdef USE_GDBSTUB
+    if (bp_check) {
+        if (Gdbstub::CheckBreakpointRead(vaddr) || Gdbstub::CheckBreakpointAccess(vaddr))
+        {
```
Subv: { in the previous line

src/core/memory.cpp
```diff
@@ -114,7 +129,18 @@ T Read(const VAddr vaddr) {
 }
 
 template <typename T>
-void Write(const VAddr vaddr, const T data) {
+void Write(const VAddr vaddr, const T data, bool bp_check) {
+#ifdef USE_GDBSTUB
+    if (bp_check) {
+        if (Gdbstub::CheckBreakpointWrite(vaddr) || Gdbstub::CheckBreakpointAccess(vaddr))
+        {
```
Subv: same

Subv:
How does this play with multiple emulation sessions?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+
+static u8 cmd_bfr[GDB_BFR_MAX];
+static u32 cmd_len;
+
+static u32 sig = 0;
+static u32 send_signal = 0;
+static u32 step_break = 0;
+static int memory_break = 0;
+
+// Default to 24689 so it won't crash if USE_GDBSTUB is not defined in citra or citra-qt
```
neobrain: Why would anything crash just because something is not defined? There's no way for the reader to follow how any of that relates to defaulting some u32 value.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    u32 len;
+} gdb_bp_t;
+
+static std::unordered_map<u32, gdb_bp_t> breakpoints_execute;
+static std::unordered_map<u32, gdb_bp_t> breakpoints_read;
+static std::unordered_map<u32, gdb_bp_t> breakpoints_write;
+static std::unordered_map<u32, gdb_bp_t> breakpoints_access;
+
+// private helpers
+static u8 Hex2Char(u8 hex)
```
neobrain: Don't use dumb abbreviations like "2", especially when they just safe one character.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+}
+
+static void Hex2Mem(u8 *dst, u8 *src, u32 len)
+{
+    while (len-- > 0) {
+        *dst++ = (Hex2Char(*src) << 4) | Hex2Char(*(src+1));
+        src += 2;
+    }
+}
+static void Wbe32Hex(u8 *p, u32 v)
```
neobrain: docstring

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+{
+    u32 i;
+    for (i = 0; i < 8; i += 2) {
+        p[i + 1] = Nibble2Hex(v >> (4 * i));
+        p[i] = Nibble2Hex(v >> (4 * (i + 1)));
+    }
+}
+
+static void Wbe64Hex(u8 *p, u64 v)
```
neobrain: docstring

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+{
+    u32 i;
+    for (i = 0; i < 16; i += 2) {
+        p[i + 1] = Nibble2Hex(v >> (4 * i));
+        p[i] = Nibble2Hex(v >> (4 * (i + 1)));
+    }
+}
+
+static u32 Re32Hex(u8 *p)
```
neobrain: docstring

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+
+    for (i = 0; i < 8; i += 2) {
+        res = (res << 4) | Hex2Char(p[7 - i - 1]);
+        res = (res << 4) | Hex2Char(p[7 - i]);
+    }
+
+    return res;
+}
+
+static u64 Re64Hex(u8 *p)
```
neobrain: docstring

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+}
+
+static u8 CalculateChecksum()
+{
+    u32 len = cmd_len;
+    u8 *ptr = cmd_bfr;
+    u8 c = 0;
+
+    while (len-- > 0)
+        c += *ptr++;
```
neobrain: u8 c = static_cast<u8>(std::accumulate(ptr, ptr + len));

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    size_t res;
+
+    res = send(sock, &ack, 1, 0);
+    if (res != 1)
+        LOG_ERROR(GdbStub, "send failed");
+}
+
+static void SendReply(const char *reply)
+{
+    u8 chk;
```
neobrain: Use more descriptive variable names, and declare variables as close as possible to their first use (as opposed to declaring all of them at the top of the function).

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    u8 *ptr;
+    int n;
+
+    if (!Gdbstub::IsActive())
+        return;
+
+    memset(cmd_bfr, 0, sizeof cmd_bfr);
+
+    cmd_len = strlen(reply);
+    if (cmd_len + 4 > sizeof cmd_bfr)
```
neobrain: use sizeof() instead of sizeof without parentheses.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+
+    cmd_len++;
+    chk = CalculateChecksum();
+    cmd_len--;
+    cmd_bfr[0] = GDB_STUB_START;
+    cmd_bfr[cmd_len + 1] = GDB_STUB_END;
+    cmd_bfr[cmd_len + 2] = Nibble2Hex(chk >> 4);
+    cmd_bfr[cmd_len + 3] = Nibble2Hex(chk);
+
+    //LOG_DEBUG(GdbStub, "gdb: reply (len: %d): %s\n", cmd_len, cmd_bfr);
```
neobrain: don't introduce code that is commented out.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+
+    memset(bfr, 0, sizeof bfr);
+
+    sprintf(bfr, "T%02x%02x:%08x;%02x:%08x;", sig, 15, htonl(Core::g_app_core->GetPC()), 13, htonl(Core::g_app_core->GetReg(13)));
+    LOG_DEBUG(GdbStub, "Response: %s", bfr);
+
+    SendReply(bfr);
+}
+
+int SendSignal(u32 s)
```
neobrain: docstring

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+    memset(reply, 0, sizeof reply);
+    id = Hex2Char(cmd_bfr[1]);
+    if (cmd_bfr[2] != '\0')
+    {
+        id <<= 4;
+        id |= Hex2Char(cmd_bfr[2]);
+    }
+
+    switch (id) {
+    case 0: case 1: case 2: case 3:
```
neobrain: eh. prefer "if-else-if-else" over this stuff.

src/core/gdbstub/gdbstub.cpp
neobrain: don't use identifiers starting with underscores. in fact, don't introduce two functions with almost the same name to begin with..

src/core/gdbstub/gdbstub.cpp
polaris-: IDA Pro expected the register data to be in big endian format. Regardless of what platform citra is being compiled with, htonl should guarantee that the data comes out in big endian order. I figured it would be better to use a well tested and known function instead of writing my own implementation.

src/core/gdbstub/gdbstub.cpp
polaris-: Last night I was told that the gdbstub was crashing when trying to bind on port 0 in particular when the port wasn't properly initialized from the configuration. It doesn't seem to be crashing for me when I try it now, but I still think defaulting to a value at least outside of 0-1023 is the right thing to do, so I will update the comment with a better explanation.

neobrain: That sounds a lot like you're not actually understanding what's going wrong. Is that right? If so, please try to figure out an actual solution rather than adding workarounds.

archshift: Well, it makes sense if it is unable to bind to the reserved ports on the system.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,942 @@
...
+
+int Gdbstub::CheckBreakpointAccess(u32 addr)
+{
+    if (sock == -1)
+        return 0;
+
+    return CheckBreakpoint(addr, Gdbstub::Access);
+}
+
+#endif
```
archshift: Newline at EOF

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,40 @@
...
+    int  CheckBreakpointRead(u32 addr);
+    int  CheckBreakpointWrite(u32 addr);
+    int  CheckBreakpointAccess(u32 addr);
+}
+
+#endif
```
archshift: Newline at EOF

src/core/gdbstub/gdbstub.cpp
neobrain: @archshift None of that should matter if USE_GDBSTUB is not defined, though.

polaris-: @neobrain The issue was when trying to use the gdbstub without defining USE_GDBSTUB in either citra or citra-qt (so it was not getting the port from the configuration and was defaulting to port 0). If you aren't compiling with the gdbstub enabled at all then the issue won't happen. After the person added the flags and read the port number from the config file, everything worked fine.

archshift: It is indicative of the issue that the GDB stub does not fail gracefully upon failing to bind to a port, though.

polaris-: @archshift That's one thing to think about, definitely. I'm not sure if it actually crash or not so I'll ask him about that later, but I just tried on a port that's already in use and it could use some better handling in the case it can't start the server. No crash for me, though.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,895 @@
...
+
+// Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
+
+#ifdef USE_GDBSTUB
+
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <signal.h>
+#include <stdarg.h>
```
lioncash: cstdio
cstring
csignal
cstdarg

polaris-:
@Subv You will need to use different ports for each session. As of the changes in the latest commits, the default behavior is to continue executing the game if the server cannot be started for any reason.

polaris-:
I think everything mentioned in the comments should be fixed now as of the latest commit.

src/citra_qt/config.cpp
```diff
@@ -62,6 +62,10 @@ void Config::ReadValues() {
     qt_config->beginGroup("Miscellaneous");
     Settings::values.log_filter = qt_config->value("log_filter", "*:Info").toString().toStdString();
     qt_config->endGroup();
+
+    qt_config->beginGroup("Gdbstub");
```
yuriks: I think you should rename this section to "Debugging".

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3580,39 +3589,59 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
...
+#define RD              cpu->Reg[inst_cream->Rd]
+#define RD2             cpu->Reg[inst_cream->Rd + 1]
+#define RN              cpu->Reg[inst_cream->Rn]
+#define RM              cpu->Reg[inst_cream->Rm]
+#define RS              cpu->Reg[inst_cream->Rs]
+#define RDHI            cpu->Reg[inst_cream->RdHi]
+#define RDLO            cpu->Reg[inst_cream->RdLo]
+#define LINK_RTN_ADDR   (cpu->Reg[14] = cpu->Reg[15] + 4)
+#define SET_PC          (cpu->Reg[15] = cpu->Reg[15] + 8 + inst_cream->signed_immed_24)
+#define SHIFTER_OPERAND inst_cream->shtop_func(cpu, inst_cream->shifter_operand)
```
yuriks: You shouldn't reformat this code, it doesn't look like you changed it and it confuses the diff.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3580,39 +3589,59 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
...
-    #define INC_PC(l)   ptr += sizeof(arm_inst) + l
-    #define INC_PC_STUB ptr += sizeof(arm_inst)
+#define INC_PC(l)   ptr += sizeof(arm_inst) + l
+#define INC_PC_STUB ptr += sizeof(arm_inst)
+
+    // GCC and Clang have a C++ extension to support a lookup table of labels. Otherwise, fallback to a
+    // clunky switch statement.
+
+
+#ifdef USE_GDBSTUB
```
yuriks: Why did you insert this code in between the define and the comment describing it?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,956 @@
+// Copyright 2013 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+// Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
+
+#ifdef USE_GDBSTUB
```
yuriks: This define should be added as an option you can toggle in CMake. I can help with that if you need it.

yuriks:
I tried this out, and unfortunately, I think the performance level is not acceptable. With the stub enabled, the emulator runs slower than even a Citra Debug build, making it unusable for anything but booting title screens.

I believe that this can be implemented with no speed hit compared to not using the stub, at least when memory breakpoints aren't in use. I'd rather discuss this on IRC however, so hopefully we can talk about it tomorrow or when you have some time.

hrydgard:
Especially when you have the JIT later, you can entirely eliminate the overhead when not enabled.

polaris-:
@yuriks Adding it to CMake would be nice. I won't be free tomorrow or the day after, but I'll be on IRC tonight after class if you want to discuss things.

I did some tests and the majority of the slowdown isn't coming from the breakpoint checks I think. It seems like everything is slowing down because of the HandlePacket() function. In particular, IsDataAvailable() since it usually doesn't get past that code unless there's data found. Taking a closer look at that code, it looks like there's a 20ms timeout on the select(). Try setting t.tv_usec to 0 just above that and it should immediately return instead of blocking until timeout. Even 1ms has a significant performance hit it seems, but setting it to 0 makes things nice and fast.

Since I did all of my testing using the default timeout from when I copied the Dolphin code, I don't know if this will cause any bugs. I will do more testing tonight.

neobrain:
Taken all of this into consideration, this is a nice PoC, but ultimately I feel that it should be rewritten from scratch without relying on any from Dolphin's code. As a matter of fact, Dolphin's code originated in the (presumably quickly hacked-together) code of some PS3 SPU emulator written by hobbyist reverse-engineers, who very likely did not really care about overall software quality at all. The fact that it was quickly adapted to suit Dolphin and then even more twisted to fit into Citra isn't helping.

Anyway, just my opinion. For what it's worth, writing a GDB stub is a matter of less than a day once you know how it works, so it's not like rewriting from scratch would take a too large amount of time.

polaris-:
@neobrain What would you consider a passable runtime when active? After I changed the timeout mentioned in my previous code, the runtime isn't too bad in comparison to before (before, it could be in the 200-500 time range according to MicroProfile, while now it's usually in the 15-50 range). I'd still like to do some more testing now that the majority of the slowdown is gone since I think it might be possible to add some kind of option to toggle it and have it always compiled without having too much overhead involved now.

neobrain:
No specific numbers in mind. Actually, the overhead when the GDB stub is active actually isn't really all that important if it's possible to emutime-enable/disable the GDB stub.

polaris-:
@neobrain Alright. I'll work on that this week then. I think I can get this gdbstub to be always compiled with minimal/no overhead when disabled and acceptable performance when enabled without a few tweaks.

mailwl:
ida pro will be able to debug 3ds games? i want it!

src/core/arm/skyeye_common/armstate.cpp
```diff
@@ -185,8 +186,27 @@ void ARMul_State::ResetMPCoreCP15Registers()
...
+    }
+}
+
+u8 ARMul_State::ReadMemory8(u32 address) const
+{
+    CheckMemoryBreakpoint(address, Gdbstub::BreakpointType::Read);
+
+    u8 data = Memory::Read8(address);
+
+    return data;
```
lioncash: `return Memory::Read8(address);`

src/core/arm/skyeye_common/armstate.cpp
```diff
@@ -185,8 +186,27 @@ void ARMul_State::ResetMPCoreCP15Registers()
     CP15[CP15_TLB_DEBUG_CONTROL] = 0x00000000;
 }
 
+static inline void CheckMemoryBreakpoint(u32 address, Gdbstub::BreakpointType type)
```
lioncash: Did you check if inline actually did anything here (it probably doesn't).

src/core/core.cpp
```diff
@@ -20,6 +22,21 @@ ARM_Interface*     g_sys_core = nullptr;  ///< ARM11 system (OS) core
 
 /// Run the core CPU loop
 void RunLoop(int tight_loop) {
+    if (Gdbstub::server_enabled) {
+        Gdbstub::HandlePacket();
+
+        // If the loop is halted and we want to step, use a tiny (1) number of instructions to execute.
+        // Otherwise get out of the loop function.
+        if (Gdbstub::halt_loop == true) {
+            if (Gdbstub::step_loop == true) {
```
lioncash: `if (Gdbstub::step_loop)`

src/core/core.cpp
```diff
@@ -20,6 +22,21 @@ ARM_Interface*     g_sys_core = nullptr;  ///< ARM11 system (OS) core
 
 /// Run the core CPU loop
 void RunLoop(int tight_loop) {
+    if (Gdbstub::server_enabled) {
+        Gdbstub::HandlePacket();
+
+        // If the loop is halted and we want to step, use a tiny (1) number of instructions to execute.
+        // Otherwise get out of the loop function.
+        if (Gdbstub::halt_loop == true) {
```
lioncash: `if (Gdbstub::halt_loop)`

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,973 @@
...
+
+// Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
+
+#include <fcntl.h>
+#include <cstdio>
+#include <cstring>
+#include <csignal>
+#include <cstdarg>
+#include <numeric>
+#include <map>
```
lioncash: These should be sorted alphabetically

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,973 @@
...
+
+#ifdef _WIN32
+WSADATA InitData;
+#endif
+
+typedef struct {
+    u32 active;
+    u32 addr;
+    u32 len;
+} Breakpoint;
```
lioncash: typedefing structs like this isn't necessary in C++. You can just do `struct Breakpoint` and it's fine.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,973 @@
...
+    }
+
+    return c;
+}
+
+/// Calculate the checksum of the current command buffer.
+static inline u8 CalculateChecksum()
+{
+    u8 checksum = static_cast<u8>(std::accumulate(&command_buffer[0], &command_buffer[0] + command_length, 0, std::plus<u8>()));
+    return checksum;
```
lioncash: you can just return directly here

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,86 @@
...
+// Refer to the license.txt file included.
+
+// Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
+
+#pragma once
+
+namespace Gdbstub {
+    /// Breakpoint Method
+    enum class BreakpointType {
+        None = 0, ///< None
```
lioncash: = 0 isn't necessary (it's zero by default)

src/core/memory.cpp
```diff
@@ -143,6 +143,7 @@ u8* GetPointer(const VAddr vaddr) {
     }
 
     LOG_ERROR(HW_Memory, "unknown GetPointer @ 0x%08x", vaddr);
+
     return nullptr;
```
lioncash: Unnecessary newline

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,973 @@
...
+    SendReply("E01");
+}
+
+/// Create and send signal packet if signal flag is set using SendSignal.
+static void HandleSignal()
+{
+    char buffer[128];
+    memset(buffer, 0, sizeof(buffer));
+
+    sprintf(buffer, "T%02x%02x:%08x;%02x:%08x;", latest_signal, 15, htonl(Core::g_app_core->GetPC()), 13, htonl(Core::g_app_core->GetReg(13)));
```
lioncash: Consider using [StringFromFormat](https://github.com/citra-emu/citra/blob/master/src/common/string_util.cpp#L99) instead of a manual buffer

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,86 @@
...
+    void ToggleServer(bool status);
+
+    /// Start the gdbstub server.
+    void Init();
+
+    /**
+    * Start the gdbstub server on specific port.
+    *
+    * @param port Port to listen for connection
+    */
```
lioncash: Asterisks are off alignment. Same goes for the ones below this comment too.

src/core/arm/skyeye_common/armstate.cpp
polaris-: I just checked the assembly again without the inline keyword and you're right. I'll remove it.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+#define GDB_STUB_ACK   '+'
+#define GDB_STUB_NAK   '-'
+
+#ifdef _WIN32
+#define SIGTRAP      5
+#define SIGTERM     15
+#define MSG_WAITALL  8
+#endif
+
+static int sock = -1;
```
neobrain: use descriptive variable names for global variables.

src/common/logging/log.h
```diff
@@ -72,6 +72,7 @@ enum class Class : ClassType {
     Render_Software,            ///< Software renderer backend
     Render_OpenGL,              ///< OpenGL backend
     Loader,                     ///< ROM loader
+    GdbStub,                    ///< GDB Stub
```
neobrain: *GDBStub

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,86 @@
+// Copyright 2013 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+// Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
+
+#pragma once
+
+namespace Gdbstub {
```
neobrain: *GDBStub

src/core/settings.h
```diff
@@ -60,6 +60,10 @@ struct Values {
     float bg_blue;
 
     std::string log_filter;
+
+    // Debugging
+    int gdbstub_port;
+    int use_gdbstub;
```
neobrain: why int? either make it bool or add a comment.

src/core/settings.h
```diff
@@ -60,6 +60,10 @@ struct Values {
     float bg_blue;
 
     std::string log_filter;
+
+    // Debugging
+    int gdbstub_port;
```
neobrain: why not unsigned? actually, shouldn't it even be fixed-size, i.e. u16?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+static u32 latest_signal = 0;
+static u32 send_signal = 0;
+static u32 step_break = 0;
+static bool memory_break = false;
+
+// Binding to a port within the reserved ports range (0-1023) requires root permissions,
+// so default to a port outside of that range.
+static u32 gdbstub_port = 24689;
+
+bool Gdbstub::halt_loop = true;
```
neobrain: Why don't you just wrap the contents of this file within `namespace Gdbstub {}` ?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+#endif
+
+#include "common/logging/log.h"
+#include "common/string_util.h"
+#include <core/arm/arm_interface.h>
+#include "core/core.h"
+#include "core/memory.h"
+#include "gdbstub.h"
+
+#define GDB_BFR_MAX  10000
```
neobrain: GDB_BUFFER_MAX

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+
+#define GDB_STUB_START '$'
+#define GDB_STUB_END   '#'
+#define GDB_STUB_ACK   '+'
+#define GDB_STUB_NAK   '-'
```
neobrain: NACK

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+
+#include <csignal>
+#include <cstdarg>
+#include <cstdio>
+#include <cstring>
+#include <fcntl.h>
+#include <map>
+#include <numeric>
+
+#ifdef _WIN32
```
neobrain: Is this actually the proper check? maybe you want MSC_VER here, instead?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+        output = (output << 4) | HexToChar(src[7 - i]);
+    }
+
+    return output;
+}
+
+/// Read a byte from the gdb client.
+static u8 ReadByte()
+{
+    u8 c = '+';
```
neobrain: Why '+'?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+* Remove the breakpoint from the given address of the specified type.
+*
+* @param type Type of breakpoint.
+* @param addr Address of breakpoint.
+*/
+static void RemoveBreakpoint(Gdbstub::BreakpointType type, u32 addr)
+{
+    std::map<u32, Breakpoint>& p = GetBreakpointList(type);
+
+    std::map<u32, Breakpoint>::const_iterator bp = p.find(addr);
```
neobrain: auto

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+
+u32 Gdbstub::GetNextBreakpointFromAddress(u32 addr, Gdbstub::BreakpointType type)
+{
+    if (!Gdbstub::IsActive()) {
+        return -1;
+    }
+
+    std::map<u32, Breakpoint>& p = GetBreakpointList(type);
+    auto next_breakpoint = p.lower_bound(addr);
+    u32 breakpoint = -1;
```
neobrain: The docstring doesn't say anything about a return value of -1 in error case. Is this a mistake in the documentation? Please also consider if assigning -1 here is just a bad workaround.

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,86 @@
...
+    /// Read and handle packet from gdb client.
+    void HandlePacket();
+
+    /**
+     * Get the nearest breakpoint of the specified type at the given address.
+     *
+     * @param addr Address to search from.
+     * @param type Type of breakpoint.
+     */
+    u32 GetNextBreakpointFromAddress(u32 addr, Gdbstub::BreakpointType type);
```
neobrain: Please use the VAddr and PAddr typedefs to make clear what kind of address you're talking about.

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,86 @@
...
+    enum class BreakpointType {
+        None,     ///< None
+        Execute,  ///< Execution Breakpoint
+        Read,     ///< Read Breakpoint
+        Write,    ///< Write Breakpoint
+        Access    ///< Access (R/W) Breakpoint
+    };
+
+    /// If halt_loop is true, the CPU loop will stop executing.
+    extern bool halt_loop;
```
neobrain: use g_ prefixes for global variables (or better yet, don't intoduce global variables).

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+
+    // Clean up temporary socket if it's still alive at this point.
+    if (tmpsock != -1) {
+        shutdown(tmpsock, SHUT_RDWR);
+    }
+}
+
+void Gdbstub::Deinit()
+{
+    Gdbstub::halt_loop = false;
```
neobrain: How does this contribute to "deinitialization"? Don't reset variables in deinitializers "just to make sure". If this is a guarantee that needs to be given in Init(), put it there instead.

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,86 @@
...
+
+    /// If halt_loop is true and step_loop is true, the CPU loop will continue for 1 instruction.
+    extern bool step_loop;
```
neobrain: Please specify on which threads these variables may safely be read/written, because clearly access to them is not thread-safe.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+    if (bind(tmpsock, server_addr, server_addrlen) < 0) {
+        LOG_ERROR(GdbStub, "Failed to bind gdb socket");
+    }
+
+    if (listen(tmpsock, 1) < 0) {
+        LOG_ERROR(GdbStub, "Failed to listen to gdb socket");
+    }
+
+    // Wait for gdb to connect
+    LOG_INFO(GdbStub, "Waiting for gdb to connect...\n");
```
neobrain: No need for explicit line break there.

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,86 @@
...
+
+    /// Start the gdbstub server.
+    void Init();
+
+    /**
+     * Start the gdbstub server on specific port.
+     *
+     * @param port Port to listen for connection
+     */
+    void Init(u32 port);
```
neobrain: This isn't used outside of the implementation of Init(), hence there's no need to have it be part of the public interface.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+            type = Gdbstub::BreakpointType::Read;
+            break;
+        case 4:
+            type = Gdbstub::BreakpointType::Access;
+            break;
+        default:
+            return SendReply("E01");
+    }
+
+    u32 i = 3;
```
neobrain: don't use fixed-size types unless there's a reason to.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+        LOG_ERROR(GdbStub, "send failed");
+    }
+}
+
+/**
+* Send reply to gdb client.
+*
+* @param reply Reply to be sent to client.
+*/
+static void SendReply(const char* reply)
```
neobrain: How is the length of the reply deduced? Is it hardcoded to a fixed number? Is it null-terminated?

As is, this looks fishy. You're passing buffers with seemingly arbitrary length to this function (some function below uses a u8[64] for a two-byte reply...). Consider adding a length parameter to this function.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+#ifdef _WIN32
+#define SIGTRAP      5
+#define SIGTERM     15
+#define MSG_WAITALL  8
+#endif
+
+static int sock = -1;
+
+static u8 command_buffer[GDB_BFR_MAX];
+static u32 command_length;
```
neobrain: Why u32?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+
+static u32 latest_signal = 0;
```
neobrain: Aren't POSIX signals of type int?

src/core/gdbstub/gdbstub.h
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    /// If halt_loop is true, the CPU loop will stop executing.
+    extern bool halt_loop;
+
+    /// If halt_loop is true and step_loop is true, the CPU loop will continue for 1 instruction.
+    extern bool step_loop;
+
+    /// If set to false, the server will never be started and no gdbstub-related functions will be executed.
+    extern bool server_enabled;
```
neobrain: This is not thread-safe, yet is written to from the GUI thread and read from in the CPU thread. At least change the type to std::atomic<bool>, but please reflect whether your design is safe at all or whether it'll run into race conditions.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+    if (received_size != 1) {
+        LOG_ERROR(GdbStub, "recv failed : %ld", received_size);
+        Gdbstub::Deinit();
+    }
+
+    return c;
+}
+
+/// Calculate the checksum of the current command buffer.
+static u8 CalculateChecksum()
```
neobrain: The way some code calls this function below is really confusing ("increment cmd buffer length, call CalculateChecksum, decrement cmd buffer length"). Consider explicitly passing the command_buffer pointer and a length parameter to this function instead.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+    }
+
+    std::map<u32, Breakpoint>& p = GetBreakpointList(type);
+
+    std::map<u32, Breakpoint>::const_iterator bp = p.find(addr);
+    if (bp != p.end()) {
+        u32 len = bp->second.len;
+
+        // IDA Pro defaults to 4-byte breakpoints for all non-hardware breakpoints
+        // no matter if it's a 4-byte or 2-byte instructions. When you execute a
```
neobrain: instruction

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+        // Thumb instruction with a 4-byte breakpoint set, it will breakpoint for
```
neobrain: it will break (?)

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+        // two instructions instead of the single instruction you placed the breakpoint
+        // on. So, as a way to make sure that execution breakpoints are only breaking
+        // on the instruction that was breakpointed on, set the length of an execution
```
neobrain: "to breakpoint" is not a verb.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+    }
+}
+
+/// Handle query command from gdb client.
+static void HandleQuery()
+{
+    LOG_DEBUG(GdbStub, "gdb: query '%s'\n", command_buffer + 1);
+
+    if (!strcmp(reinterpret_cast<const char*>(command_buffer + 1), "TStatus")) {
+        return SendReply("T0");
```
neobrain: Uhm, is this syntax actually legit C++? (return X from a function returning void)

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+    SendPacket(GDB_STUB_ACK);
+}
+
+/// Check if there is data to be read from the gdb client.
+static bool IsDataAvailable() {
+    if (!Gdbstub::IsActive()) {
+        return false;
+    }
+
+    fd_set _fds;
```
neobrain: don't use underscore prefixes. In fact, use more distinguishing names for distinct variables.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+    struct timeval t;
+    t.tv_sec = 0;
+    t.tv_usec = 0;
+
+    if (select(sock + 1, fds, nullptr, nullptr, &t) < 0) {
+        LOG_ERROR(GdbStub, "select failed");
+        return false;
+    }
+
+    if (FD_ISSET(sock, fds)) {
```
neobrain: return FD_ISSET(sock, fds)

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,969 @@
...
+}
+
+/// Send all registers to the gdb client.
+static void ReadRegisters()
+{
+    static u8 buffer[GDB_BFR_MAX - 4];
+    memset(buffer, 0, sizeof(buffer));
+
+    u8* bufptr = buffer;
+    for (u32 i = 0; i < 42; i++) {
```
neobrain: magic numbers?

yuriks:
This is still a RFC. I'm working with @polaris- to change parts of the design as required to enable the stub to be always enabled without a performance penalty if possible. Until we finalized the design I think everyone should hold off on doing detailed code reviews.

src/citra_qt/config.cpp
```diff
@@ -97,6 +102,11 @@ void Config::SaveValues() {
     qt_config->beginGroup("Miscellaneous");
     qt_config->setValue("log_filter", QString::fromStdString(Settings::values.log_filter));
     qt_config->endGroup();
+
+    qt_config->beginGroup("Debugging");
+    qt_config->setValue("use_gdbstub", Settings::values.use_gdbstub);
```
yuriks: `gdbstub_enabled`? (to be prefix-consistent with the other option)

src/common/logging/backend.cpp
```diff
@@ -56,7 +56,8 @@ namespace Log {
         CLS(Render) \
         SUB(Render, Software) \
         SUB(Render, OpenGL) \
-        CLS(Loader)
+        CLS(Loader) \
+        CLS(GDBStub)
```
yuriks: put it under the Debug class: `SUB(Debug, GDBStub)`

src/core/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ set(SRCS
             file_sys/archive_systemsavedata.cpp
             file_sys/disk_archive.cpp
             file_sys/ivfc_archive.cpp
+            gdbstub/gdbstub.cpp
```
yuriks: IMO these two files can simply go inside the `arm/` directory.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3604,15 +3609,27 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     #define INC_PC(l)   ptr += sizeof(arm_inst) + l
     #define INC_PC_STUB ptr += sizeof(arm_inst)
 
+#define GDB_BP_CHECK \
+    cpu->Cpsr &= ~(1 << 5); \
+    cpu->Cpsr |= cpu->TFlag << 5; \
+    if (GDBStub::g_server_enabled) { \
+        if (GDBStub::IsMemoryBreak() || PC == breakpoint_offset) { \
+            GDBStub::Break(); \
+            return num_instrs; \
```
yuriks: I think you have to use `goto END;` here instead so that the cached flags are properly updated in the ARM state before exiting the loop.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3903,6 +3921,11 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
                 goto END;
         }
 
+        // Find breakpoint if one exists within the block
+        if (GDBStub::g_server_enabled && GDBStub::IsActive()) {
```
yuriks: You should just a single variable/function to call which will be true when the server is both enabled *and* connected. This way you cut down on redundant checks. (What does it mean for the server to be enabled but not active anyway? It doesn't seem like this is something that should ever happen.)

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3903,6 +3921,11 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
...
+            breakpoint_offset = GDBStub::GetNextBreakpointFromAddress(cpu->Reg[15], GDBStub::BreakpointType::Execute);
```
yuriks: Do you need to handle the case where, after a BP is hit, you need to avoid re-triggering the same BP when you conitnue? Or is that handled by the client?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+#define SIGTERM     15
+#define MSG_WAITALL  8
+#endif
+
+#define R0_REGISTER 0
+#define R15_REGISTER 15
+#define CSPR_REGISTER 25
+
+namespace GDBStub {
+    static int gdbserver_socket = -1;
```
yuriks: Don't indent namespace contents.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+#endif
+
+#include "common/logging/log.h"
+#include "common/string_util.h"
+#include <core/arm/arm_interface.h>
+#include "core/core.h"
+#include "core/memory.h"
+#include "gdbstub.h"
+
+#define GDB_BUFFER_SIZE  10000
```
yuriks: Use `const`s for this and the other constants instead of defines.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+    static u32 step_break = 0;
+    static bool memory_break = false;
+
+    // Binding to a port within the reserved ports range (0-1023) requires root permissions,
+    // so default to a port outside of that range.
+    static u16 gdbstub_port = 24689;
+
+    bool halt_loop = true;
+    bool step_loop = false;
+    std::atomic<bool> g_server_enabled(false);
```
yuriks: You can still use the `= false` initialization syntax here I think. (Looks better IMO for a type like this.)

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+    static u32 send_signal = 0;
+    static u32 step_break = 0;
+    static bool memory_break = false;
+
+    // Binding to a port within the reserved ports range (0-1023) requires root permissions,
+    // so default to a port outside of that range.
+    static u16 gdbstub_port = 24689;
+
+    bool halt_loop = true;
+    bool step_loop = false;
```
yuriks: These two vars should also have the `g_` prefix or should be made static.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+    static std::map<u32, Breakpoint> breakpoints_execute;
+    static std::map<u32, Breakpoint> breakpoints_read;
+    static std::map<u32, Breakpoint> breakpoints_write;
+
+    /**
+     * Turns hex string character into the equivalent byte.
+     *
+     * @param hex Input hex character to be turned into byte.
+     */
+    static u8 HexToChar(u8 hex)
```
yuriks: The name of this function implies it does exactly the opposite of what it actually does. `HexCharToValue` maybe?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+    {
+        if (hex >= '0' && hex <= '9') {
+            return hex - '0';
+        }
+        else if (hex >= 'a' && hex <= 'f') {
+            return hex - 'a' + 0xa;
```
yuriks: Uppercase hex literals: `0xA` (same on other lines here)

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+     * Turns hex string character into the equivalent byte.
+     *
+     * @param hex Input hex character to be turned into byte.
+     */
+    static u8 HexToChar(u8 hex)
+    {
+        if (hex >= '0' && hex <= '9') {
+            return hex - '0';
+        }
+        else if (hex >= 'a' && hex <= 'f') {
```
yuriks: `else if ...` should go in the same line as the closing brace. (likewise elsewhere)

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+    static std::map<u32, Breakpoint> breakpoints_read;
+    static std::map<u32, Breakpoint> breakpoints_write;
+
+    /**
+     * Turns hex string character into the equivalent byte.
+     *
+     * @param hex Input hex character to be turned into byte.
+     */
+    static u8 HexToChar(u8 hex)
+    {
```
yuriks: Function opening brace does not go in a separate line. Likewise for the rest of the file

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+    * Converts input hex string characters into an array of equivalent of u8 bytes.
+    *
+    * @param dest Pointer to buffer to store u8 bytes.
+    * @param src Pointer to array of output hex string characters.
+    * @param len Length of src array.
+    */
+    static void HexToMem(u8* dest, u8* src, u32 len)
+    {
+        while (len-- > 0) {
+            *dest++ = (HexToChar(*src) << 4) | HexToChar(*(src + 1));
```
yuriks: `src[0]` and `src[1]`? I think it looks a bit better.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+            src += 2;
+        }
+    }
+
+    /**
+     * Convert a u32 into a hex string.
+     *
+     * @param dest Pointer to buffer to store output hex string characters.
+     */
+    static void IntToHex(u8* dest, u32 v)
```
yuriks: Can't you just use snprintf instead of this function?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+            dest[i] = NibbleToHex(v >> (4 * (i + 1)));
+        }
+    }
+
+    /**
+    * Convert a hex string into a u32.
+    *
+    * @param src Pointer to hex string.
+    */
+    static u32 HexToInt(u8* src)
```
yuriks: Likewise, I think you can just use [strtoul](http://en.cppreference.com/w/c/string/byte/strtoul) for this?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+            Deinit();
+        }
+
+        return c;
+    }
+
+    /// Calculate the checksum of the current command buffer.
+    static u8 CalculateChecksum(u8 *buffer, u32 length)
+    {
+        return static_cast<u8>(std::accumulate(buffer, buffer + length + 1, 0, std::plus<u8>()));
```
yuriks: This would probably be clearer as an explicit loop. Either way, what's with the `+ 1` on the end iterator?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,977 @@
...
+        std::map<u32, Breakpoint>& p = GetBreakpointList(type);
+
+        auto bp = p.find(addr);
+        if (bp != p.end()) {
+            LOG_DEBUG(GDBStub, "gdb: removed a breakpoint: %08x bytes at %08x\n", bp->second.len, bp->second.addr);
+            p.erase(addr);
+        }
+    }
+
+    PAddr GetNextBreakpointFromAddress(PAddr addr, BreakpointType type)
```
yuriks: Is there any point in the `type` parameter? This is only used with execution breakpoints, right?

src/core/gdbstub/gdbstub.cpp
polaris-: I originally had it was `= false` before, but it broke TravisCI: https://travis-ci.org/citra-emu/citra/jobs/79774738

src/core/gdbstub/gdbstub.cpp
polaris-: For now it's only used for execution breakpoints, but I thought it wouldn't hurt make it explicit what kind of breakpoint you are searching for when you call the function, in case in the future someone wants to find a breakpoint based on address for another purpose. So instead of having "GetNextExecuteBreakpointFromAddress", "GetNextReadBreakpointFromAddress", etc, just having a "GetNextBreakpointFromAddress" would be better is what I was thinking.

src/citra_qt/config.cpp
polaris-: Would that be prefix-consistent with the other options? I saw things like use_hw_renderer and use_shader_jit, which is what I was using as examples when creating all of the option-related code.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
polaris-: I guess a better change here would be to IsActive to IsConnected then.

The reason I did this is because I wanted to reduce the number of function calls required to check if the server was enabled and running when it comes to normal execution when the gdbstub is disabled. The g_server_enabled flag just compiles to a comparison, but the IsActive function doesn't get inlined. It might be a little bit of a premature optimization(?), but I wanted to reduce the impact as much as possible when the gdbstub wasn't enabled.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
polaris-: That is handled by the client. I just verified with gdb (normal) and IDA Pro, and when you trigger a breakpoint it will remove it, issue a step, and add the breakpoint back.

src/core/gdbstub/gdbstub.cpp
polaris-: There's a specific format that the data needs to be output, so I don't think snprintf here or strtoul for HexToInt would work. For example, if you were to call IntToHex with the value 0x12345678, the output would be "78563412" while snprintf would just give you "12345678".

yuriks:
For informational purposes, here are benchmarks I took from an (unspecified, but without major revisions) earlier version of this PR:

 - master: 55.0ms
 - stub disabled: 58.0ms
 - IDA attached: 69.5ms

It's certainly possible to make the overhead smaller, but that should wait for a follow-up.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3580,6 +3583,8 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     Common::Profiling::ScopeTimer timer_execute(profile_execute);
     MICROPROFILE_SCOPE(DynCom_Execute);
 
+    int breakpoint_offset = -1;
```
neobrain: Don't use signed integers for values which are unsigned offsets. Use an optional type instead of the placeholder value "-1".

src/citra/config.cpp
```diff
@@ -75,6 +75,10 @@ void Config::ReadValues() {
 
     // Miscellaneous
     Settings::values.log_filter = glfw_config->Get("Miscellaneous", "log_filter", "*:Info");
+
+    // GDBStubebugging
```
bunnei: Huh? `GDB stub debugging` ?

src/core/CMakeLists.txt
bunnei: +1

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3878,6 +3895,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     unsigned int addr;
     unsigned int num_instrs = 0;
 
+
```
bunnei: Unnecessary extra space

src/core/system.cpp
```diff
@@ -22,9 +24,13 @@ void Init(EmuWindow* emu_window) {
     Kernel::Init();
     HLE::Init();
     VideoCore::Init(emu_window);
+
+    GDBStub::Init();
 }
 
 void Shutdown() {
+    GDBStub::Deinit();
```
bunnei: `Shutdown`? (To be consistent)... Also, no need for the extra space IMO

src/core/system.cpp
```diff
@@ -22,9 +24,13 @@ void Init(EmuWindow* emu_window) {
     Kernel::Init();
     HLE::Init();
     VideoCore::Init(emu_window);
+
```
bunnei: No need for the extra space IMO

bunnei:
I removed the [RFC] tag from this and would like to get it merged this week. Final call for reviews. If I don't hear anything, I'm going to assume everyone is OK with this change.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,945 @@
...
+static void SendReply(const char* reply) {
+    if (!IsConnected()) {
+        return;
+    }
+
+    memset(command_buffer, 0, sizeof(command_buffer));
+
+    command_length = strlen(reply);
+    if (command_length + 4 > sizeof(command_buffer)) {
+        LOG_ERROR(Debug_GDBStub, "command_buffer overflow in SendReply");
```
Subv: Shouldn't you return here? otherwise you'll be overruning the buffer below

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,945 @@
...
+        SendSignal(SIGTRAP);
+        return;
+    } else if (c != GDB_STUB_START) {
+        LOG_DEBUG(Debug_GDBStub, "gdb: read invalid byte %02x\n", c);
+        return;
+    }
+
+    while ((c = ReadByte()) != GDB_STUB_END) {
+        command_buffer[command_length++] = c;
+        if (command_length == sizeof(command_buffer)) {
```
Subv: if command_length is command_buffer - 1 before the increment (no overflow), this check will pass and act as if there were an error.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,945 @@
...
+    SendReply(reinterpret_cast<char*>(reply));
+}
+
+/// Send all registers to the gdb client.
+static void ReadRegisters() {
+    static u8 buffer[GDB_BUFFER_SIZE - 4];
+    memset(buffer, 0, sizeof(buffer));
+
+    u8* bufptr = buffer;
+    for (int i = 0; i <= CSPR_REGISTER; i++) {
```
Subv: Does this also transmit the FPSCR?

polaris-: @Subv It doesn't, but I tried to add support for VFP registers in my latest commit. I could never get gdb to display the VFP registers so I have no way of testing it that I know of. At the very least, I used the arm-with-vfpv3.dat register mapping file from the gdb source code to figure out the order of the registers so that should be right at least. One thing that worries me is that I'm not exactly sure if gdb is expecting a 32bit value or 64bit value for the VFP regs. From examples I could find online, and judging by the register map file, I believe it's expecting 64bit values so I padded the VFP registers.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+
+#ifndef MSG_WAITALL
+const u32 MSG_WAITALL = 8;
+#endif
+
+const u32 R0_REGISTER = 0;
+const u32 R15_REGISTER = 15;
+const u32 CSPR_REGISTER = 25;
+const u32 FPSCR_REGISTER = 58;
+const u32 MAX_REGISTERS = 90;
```
lioncash: These constants should be static

src/citra/citra.cpp
```diff
@@ -30,6 +30,8 @@
 
 #include "video_core/video_core.h"
 
+#include "core/gdbstub/gdbstub.h"
```
lioncash: This should be with the other core headers (above loader includes)

src/citra_qt/main.cpp
```diff
@@ -48,6 +48,8 @@
 
 #include "video_core/video_core.h"
 
+#include "core/gdbstub/gdbstub.h"
```
lioncash: Should be with other core includes (above loader)

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+#include <common/x64/abi.h>
+#include <io.h>
+#include <iphlpapi.h>
+#define SHUT_RDWR 2
+#else
+#include <sys/select.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <netinet/in.h>
+#include <unistd.h>
```
lioncash: unistd would be included first

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+#endif
+
+#include "common/logging/log.h"
+#include "common/string_util.h"
+#include <core/arm/arm_interface.h>
```
lioncash: This should be quotations for the include

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+#include "core/core.h"
+#include "core/memory.h"
+#include "gdbstub.h"
```
lioncash: `core/core.h` and `core/memory` go above the arm interface include

The include for gdbstub.h should have the full relative path.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+            IntToHex(bufptr + i * 8, 0);
+            IntToHex(bufptr + (i + 1) * 8, 0);
+            i++; // These registers seem to be all 64bit instead of 32bit, so skip two instead of one
+            reg++;
+        } else if (i > CSPR_REGISTER && i < MAX_REGISTERS) {
+            IntToHex(bufptr + i * 8, Core::g_app_core->GetVFPReg(reg - CSPR_REGISTER - 1));
+            IntToHex(bufptr + (i + 1) * 8, 0);
+            i++;
+        } else if (i == MAX_REGISTERS) {
+            IntToHex(bufptr + i * 8, Core::g_app_core->GetVFPSystemReg(VFP_FPSCR));
```
lioncash: What is `* 8` supposed to signify? If it's supposed to signify the amount of bits in a byte it should be replaced with `* CHAR_BIT` to be explict about intent.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+    int i = 3;
+    PAddr addr = 0;
+    while (command_buffer[i] != ',') {
+        addr = (addr << 4) | HexCharToValue(command_buffer[i++]);
+    }
+    i++;
+
+    u32 len = 0;
+    while (i < command_length) {
+        len = (len << 4) | HexCharToValue(command_buffer[i++]);
```
lioncash: This operation can be made into it's own function to be more explicit about what it's doing.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+
+        type = BreakpointType::Write;
+    }
+
+    RemoveBreakpoint(type, addr);
+    SendReply("OK");
+}
+
+void HandlePacket() {
+    if (!IsConnected()) {
```
lioncash:
```c
if (!IsConnected() || !IsDataAvailable())
    return;
```

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+ * @param signal Signal to be sent to client.
+ */
+int SendSignal(u32 signal) {
+    if (gdbserver_socket == -1) {
+        return 1;
+    }
+
+    latest_signal = signal;
+
+    if (send_signal) {
```
lioncash: send_signal is always set to 1 before this is called.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+
+    SendReply(buffer.c_str());
+}
+
+/**
+ * Set signal and send packet to client through HandleSignal if signal flag is set using SendSignal.
+ *
+ * @param signal Signal to be sent to client.
+ */
+int SendSignal(u32 signal) {
```
lioncash: If this returns basic success/fail, it should be a bool

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,970 @@
...
+namespace GDBStub {
+
+static int gdbserver_socket = -1;
+
+static u8 command_buffer[GDB_BUFFER_SIZE];
+static u32 command_length;
+
+static u32 latest_signal = 0;
+static u32 send_signal = 0;
+static u32 step_break = 0;
```
lioncash: `send_signal` and `step_break` can be made `bool`s

src/core/gdbstub/gdbstub.cpp
neobrain: Also, `<algorithm>` is your friend:

```
const uint32_t len = std::accumulate(&command_buffer[i], &command_buffer[i+command_length], (uint32_t)0,
                                     [](uint32_t length, uint8_t val) {
                                         return (length << 4) | HexCharToValue(val);
                                     });
```

lioncash: True, but I'd also say that doesn't really explain why that operation is necessary to get the length. (I do like the use of accumulate though)

src/core/gdbstub/gdbstub.cpp
yuriks: All global `const` are implicitly static by default.

bunnei:
Final final call for reviews. Anyone else have anything to add here? If not, I plan on merging this week.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,945 @@
...
+
+// Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
+
+#include <algorithm>
+#include <csignal>
+#include <cstdarg>
+#include <cstdio>
+#include <cstring>
+#include <fcntl.h>
+#include <limits.h>
```
lioncash: climits

src/core/gdbstub/gdbstub.cpp
```diff
@@ -0,0 +1,945 @@
...
+#include <sys/un.h>
+#include <netinet/in.h>
+#endif
+
+#include "common/logging/log.h"
+#include "common/string_util.h"
+#include "core/core.h"
+#include "core/memory.h"
+#include "core/arm/arm_interface.h"
+#include "gdbstub.h"
```
lioncash: Should use a relative path

PR #1123, opened 2015-09-07
GSP: Implement command 0x05, used for flushing caches

yuriks:
May fix additional texture caching issues. (Though mostly in homebrew, I haven't seen any commercial software use this to flush anything but command lists.)

### DIFF
--- a/src/core/hle/service/gsp_gpu.cpp
+++ b/src/core/hle/service/gsp_gpu.cpp
@@ -377,12 +377,16 @@ static void ExecuteCommand(const Command& command, u32 thread_id) {
                                                           command.dma_request.size);
         break;
 
-    // ctrulib homebrew sends all relevant command list data with this command,
-    // hence we do all "interesting" stuff here and do nothing in SET_COMMAND_LIST_FIRST.
-    // TODO: This will need some rework in the future.
-    case CommandId::SET_COMMAND_LIST_LAST:
+    // TODO: This will need some rework in the future. (why?)
+    case CommandId::SUBMIT_GPU_CMDLIST:
     {
-        auto& params = command.set_command_list_last;
+        auto& params = command.submit_gpu_cmdlist;
+
+        if (params.do_flush) {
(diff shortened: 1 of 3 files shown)


src/core/hle/service/gsp_gpu.cpp
```diff
@@ -377,12 +377,19 @@ static void ExecuteCommand(const Command& command, u32 thread_id) {
...
+    // TODO: This will need some rework in the future. (why?)
+    case CommandId::SUBMIT_GPU_CMDLIST:
     {
-        auto& params = command.set_command_list_last;
+        auto& params = command.submit_gpu_cmdlist;
+
+        if (params.do_flush) {
+            // Command lists are not processed by the hardware renderer, so we don't need to
+            // actually flush them.
+#if 0
```
lioncash: Wouldn't the comment itself suffice instead of leaving in dead code?

yuriks: I left it there as some sort of documentation to indicate what the function *would* do, because otherwise there's no indication of what it would actually flush. I could replace it with a text explanation instead?

lioncash: Text explanation might be nicer, in the end it's up to you.

yuriks:
Updated

src/core/hle/service/gsp_gpu.h
```diff
@@ -138,6 +142,13 @@ struct Command {
             u32 flags;
         } texture_copy;
 
+        struct {
+            struct {
+                u32 address;
+                u32 size;
+            } regions[3];
+        } cache_flush;
```
neobrain: Why not just cache_flush[3] or something similar? Not much point in nesting structs here.

yuriks: I think I'd prefer to keep them consistent with 1 struct per message type.

Subv: I agree with yuriks

bunnei:
LGTM. Feel free to squash and merge.

PR #1124, opened 2015-09-08
citra-qt: Trim recently used files list to size when insterting new item

yuriks:
Even though they weren't visible in the UI, old entries would never be removed from the list and would be stored in the config file across sessions.


archshift:
More complexity in `main.cpp` :(

LGTM

PR #1128, opened 2015-09-08
CMake: Make all cache options appear even in case of errors

yuriks:
The `option` commands have been moved to the top of the file, so that
the relevant options are registered in the CMake cache even if one of
the required libraries is not found. This solves an ergonomic problem
when using bundled libraries where you have to first download GLFW
before being able to select the option to also download Qt.


bunnei:
LGTM

PR #1130, opened 2015-09-10
memory: Get rid of pointer casts

lioncash:


yuriks:
I just realized this but, I don't think this optimization the code does of using larger writes works in non-LE systems. IMO you should just get rid of those, barely anything uses these functions anyway I may re-write them to be faster (do less page table lookups) later.

yuriks:
LGTM

PR #1131, opened 2015-09-10
y2r: Give local variables an initial value

lioncash:
Keeps compilers/static analyzers quiet.


yuriks:
LGTM *mumbles something about C's lack of a strict `match` expression*

PR #1134, opened 2015-09-10
gl_rasterizer: Get rid of pointer casts

lioncash:


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -1010,8 +1014,12 @@ void RasterizerOpenGL::CommitDepthBuffer() {
                         u32 gl_pixel_index = (x + y * fb_depth_texture.width);
 
                         u8* pixel = depth_buffer + dst_offset;
-                        u32 depth_stencil = ((u32*)temp_gl_depth_data)[gl_pixel_index];
-                        *(u32*)pixel = (depth_stencil >> 8) | (depth_stencil << 24);
+
+                        u32 depth_stencil;
+                        std::memcpy(&depth_stencil, &temp_gl_depth_data[gl_pixel_index], sizeof(u32));
```
yuriks: This isn't equivalent to the old code (different array element size when indexing). That said, I think this pointer can be casted (it starts out as a uchar and isn't aliased) before the two nested loops and then simply used directly without the memcpy.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -901,8 +901,12 @@ void RasterizerOpenGL::ReloadDepthBuffer() {
...
 
                 u8* pixel = depth_buffer + dst_offset;
-                u32 depth_stencil = *(u32*)pixel;
-                ((u32*)temp_fb_depth_data)[gl_pixel_index] = (depth_stencil << 8) | (depth_stencil >> 24);
+
+                u32 depth_stencil;
+                std::memcpy(&depth_stencil, pixel, sizeof(u32));
+
+                u32 pixel_value = (depth_stencil << 8) | (depth_stencil >> 24);
+                std::memcpy(&temp_fb_depth_data[gl_pixel_index], &pixel_value, sizeof(u32));
```
yuriks: Check if you can't also do a similar thing here.

PR #1141, opened 2015-09-11
video_core: Remove unnecessary includes from headers

lioncash:


yuriks:
What does this green button do?

PR #1144, opened 2015-09-11
common: Get rid of debug_interface.h

lioncash:
This is technically unused. Also removes TMemChecks because it relies on this. Whenever memory breakpoints are implemented for real, it should be designed to match the codebase debugging mechanisms.


bunnei:
LGTM

PR #1164, opened 2015-09-19
citra-qt: high-DPI fixes and Retina on OS X

kemenaran:
This PR brings two things:

* **Fix the coordinates of mouse events in high-DPI modes (on all platforms).** Mouse events didn't account for the window DPI.

* **Enable high-DPI UI on OS X.** The OS will render the widgets using the system screen DPI (instead of being locked at @1x resolution).

    For this we have to customize the app manifest file (named `Info.plist` on OS X) — instead of using the default cmake-generated one. This custom app manifest will also enable greater control of the app metadatas on OS X : I already have some code to fix the app icon and improve the app name displayed by the system.

    Apart from the OS X app manifest, no changes are made to the existing high-DPI rendering code already present in Citra.

standarddpi
highdpi

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -181,16 +181,9 @@ void GRenderWindow::PollEvents() {
 void GRenderWindow::OnFramebufferSizeChanged()
 {
     // Screen changes potentially incur a change in screen DPI, hence we should update the framebuffer size
-#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
-    // windowHandle() might not be accessible until the window is displayed to screen.
-    auto pixel_ratio = windowHandle() ? (windowHandle()->screen()->devicePixelRatio()) : 1.0;
-
-    unsigned width = child->QPaintDevice::width() * pixel_ratio;
-    unsigned height = child->QPaintDevice::height() * pixel_ratio;
-#else
-    unsigned width = child->QPaintDevice::width();
-    unsigned height = child->QPaintDevice::height();
-#endif
+    qreal pixelRatio = windowPixelRatio();
+    unsigned width = child->QPaintDevice::width() * pixelRatio;
+    unsigned height = child->QPaintDevice::height() * pixelRatio;
 
     NotifyFramebufferLayoutChanged(EmuWindow::FramebufferLayout::DefaultScreenLayout(width, height));
 }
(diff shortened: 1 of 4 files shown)


src/citra_qt/bootmanager.cpp
```diff
@@ -236,17 +236,27 @@ void GRenderWindow::keyReleaseEvent(QKeyEvent* event)
 
 void GRenderWindow::mousePressEvent(QMouseEvent *event)
 {
+    auto pixel_ratio = 1.0f;
```
neobrain: don't use auto for "trivial" types.

kemenaran: Fixed.

src/citra_qt/bootmanager.cpp
```diff
@@ -236,17 +236,27 @@ void GRenderWindow::keyReleaseEvent(QKeyEvent* event)
 
 void GRenderWindow::mousePressEvent(QMouseEvent *event)
 {
+    int pixel_ratio = 1.0;
```
neobrain: Why int? devicePixelRatio returns a float...

src/citra_qt/bootmanager.cpp
```diff
@@ -236,17 +236,27 @@ void GRenderWindow::keyReleaseEvent(QKeyEvent* event)
...
-        this->TouchPressed(static_cast<unsigned>(pos.x()), static_cast<unsigned>(pos.y()));
+        this->TouchPressed(static_cast<unsigned>(pos.x() * pixel_ratio), static_cast<unsigned>(pos.y() * pixel_ratio));
     }
 }
 
 void GRenderWindow::mouseMoveEvent(QMouseEvent *event)
 {
+    int pixel_ratio = 1.0;
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+    pixel_ratio = windowHandle()->screen()->devicePixelRatio();
```
neobrain: Given that this code is duplicated across a *lot* of places now, please move it into a GRenderWindow method.

src/citra_qt/bootmanager.cpp
```diff
@@ -179,7 +179,7 @@ void GRenderWindow::OnFramebufferSizeChanged()
     // Screen changes potentially incur a change in screen DPI, hence we should update the framebuffer size
 #if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
     // windowHandle() might not be accessible until the window is displayed to screen.
-    auto pixel_ratio = windowHandle() ? (windowHandle()->screen()->devicePixelRatio()) : 1.0;
+    int pixel_ratio = windowHandle() ? (windowHandle()->screen()->devicePixelRatio()) : 1.0;
```
neobrain: Why?

src/citra_qt/bootmanager.cpp
neobrain: Also uh, I mean.. you don't just assign "1.0" to an integer.. :p

kemenaran: Erf, indeed. Fixed :)

src/citra_qt/bootmanager.cpp
kemenaran: Done.

kemenaran:
@neobrain Thanks for the review. Comments addressed; I also rebased the branch.

bunnei:
LGTM. Tested on Windows.

Issue #874, opened 2015-06-20
Audit usage of `bool` in command buffers


yuriks:
@Lectem discovered an interesting pitfall when reading `bool`-type values from service command buffers: Nintendo's implementations use the `ldrb`/`strb` instruction for reading/writing bool values, leaving the 3 unused bytes unchanged. In contrast we usually just read the entire u32 and cast it to bool. This means that if the three unused bytes have non-zero values the field will be incorrectly interpreted as `true`. This was causing spurious failures when using the Y2R service.

We should audit all of our service handling code to ensure we aren't making this mistake anywhere.

Subv:
Here's how i did it for the AM::TitleIDListGetTotal function https://github.com/citra-emu/citra/blob/master/src/core/hle/service/am_sys.cpp#L23

mailwl:
short values often contains garbage in hiword, too

yuriks:
@mailwl Those aren't as much of a problem since they're usually truncated to a `u16`. Doesn't hurt to watch out for those too, though, true.

B3n30:
looking at our ipc helpers we do that there properly. Since we  use ipc helpers now everywhere I think this can be closed...

(If I missed something feel free to reopen it again)

PR #1165, opened 2015-09-19
Added CSND_ExecuteType0Commands stub.

esoteric-programmer:
The stub sets byte 0x4 of CSND command buffer to 1.
csndPlaySound of ctrulib waits until that byte becomes non-zero.
Therefore, homebrew playing sound by CSND does not freeze any more.

### DIFF
--- a/src/core/hle/service/csnd_snd.cpp
+++ b/src/core/hle/service/csnd_snd.cpp
@@ -3,6 +3,8 @@
 // Refer to the license.txt file included.
 
 #include "core/hle/hle.h"
+#include "core/hle/kernel/mutex.h"
+#include "core/hle/kernel/shared_memory.h"
 #include "core/hle/service/csnd_snd.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -11,11 +13,11 @@
 namespace CSND_SND {
 
 const Interface::FunctionInfo FunctionTable[] = {
-    {0x00010140, nullptr,               "Initialize"},
-    {0x00020000, nullptr,               "Shutdown"},
-    {0x00030040, nullptr,               "ExecuteType0Commands"},
+    {0x00010140, Initialize,            "Initialize"},
+    {0x00020000, Shutdown,              "Shutdown"},
+    {0x00030040, ExecuteType0Commands,  "ExecuteType0Commands"},
     {0x00040080, nullptr,               "ExecuteType1Commands"},
-    {0x00050000, nullptr,               "AcquireSoundChannels"},
+    {0x00050000, AcquireSoundChannels,  "AcquireSoundChannels"},
     {0x00060000, nullptr,               "ReleaseSoundChannels"},
     {0x00070000, nullptr,               "AcquireCaptureDevice"},
     {0x00080040, nullptr,               "ReleaseCaptureDevice"},
@@ -29,6 +31,53 @@ const Interface::FunctionInfo FunctionTable[] = {
 
 Interface::Interface() {
     Register(FunctionTable);
+}
+
+static Kernel::SharedPtr<Kernel::SharedMemory> shared_memory = nullptr;
+static Kernel::SharedPtr<Kernel::Mutex> mutex = nullptr;
+
+void Initialize(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    shared_memory = Kernel::SharedMemory::Create(cmd_buff[1],
+            Kernel::MemoryPermission::ReadWrite,
+            Kernel::MemoryPermission::ReadWrite, "CSNDSharedMem");
+
+    mutex = Kernel::Mutex::Create(false);
+
+    cmd_buff[1] = 0;
+    cmd_buff[2] = 0x4000000;
+    cmd_buff[3] = Kernel::g_handle_table.Create(mutex).MoveFrom();
+    cmd_buff[4] = Kernel::g_handle_table.Create(shared_memory).MoveFrom();
+}
+
+void ExecuteType0Commands(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    if (shared_memory != nullptr) {
+        struct Type0Command* command = reinterpret_cast<struct Type0Command*>(
+                shared_memory->GetPointer(cmd_buff[1]));
+        if (command == nullptr) {
+            cmd_buff[1] = 1;
+        }else{
+            LOG_WARNING(Service, "(STUBBED) CSND_SND::ExecuteType0Commands");
+            command->finished |= 1;
+            cmd_buff[1] = 0;
+        }
+    }else{
+        cmd_buff[1] = 1;
+    }
+}
+
+void AcquireSoundChannels(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    cmd_buff[1] = 0;
+    cmd_buff[2] = 0xFFFFFF00;
+}
+
+void Shutdown(Service::Interface* self) {
+    shared_memory = nullptr;
+    mutex = nullptr;
 }
 
 } // namespace
--- a/src/core/hle/service/csnd_snd.h
+++ b/src/core/hle/service/csnd_snd.h
@@ -20,4 +20,17 @@ public:
     }
 };
 
+struct Type0Command {
+    // command id and next command offset
+    u32 command_id;
+    u32 finished;
+    u32 flags;
+    u8  parameters[20];
+};
+
+void Initialize(Service::Interface* self);
+void ExecuteType0Commands(Service::Interface* self);
+void AcquireSoundChannels(Service::Interface* self);
+void Shutdown(Service::Interface* self);
+
 } // namespace


purpasmart96:
The  function "csnd_type0command" should be "ExecuteType0Commands" since that's the name in the table.

src/core/hle/service/csnd_snd.cpp
```diff
@@ -31,4 +31,10 @@ Interface::Interface() {
     Register(FunctionTable);
 }
 
+void csnd_type0command(Service::Interface* self) {
```
purpasmart96: Should be "void ExecuteType0Commands(Service::Interface* self)"
As it is the name used for the string in the table

src/core/hle/service/csnd_snd.h
```diff
@@ -20,4 +20,6 @@ class Interface : public Service::Interface {
     }
 };
 
+void csnd_type0command(Service::Interface* self);
```
purpasmart96: should be void ExecuteType0Commands(Service::Interface* self);

src/core/hle/service/csnd_snd.cpp
```diff
@@ -13,7 +13,7 @@ namespace CSND_SND {
 const Interface::FunctionInfo FunctionTable[] = {
     {0x00010140, nullptr,               "Initialize"},
     {0x00020000, nullptr,               "Shutdown"},
-    {0x00030040, nullptr,               "ExecuteType0Commands"},
+    {0x00030040, csnd_type0command,     "ExecuteType0Commands"},
```
purpasmart96: Function names should be the same as the string name

src/core/hle/service/csnd_snd.cpp
```diff
@@ -31,4 +31,10 @@ Interface::Interface() {
     Register(FunctionTable);
 }
 
+void csnd_type0command(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    reinterpret_cast<char*>(cmd_buff)[4] = 1;
```
bunnei: Needs a comment explaining why you are doing this. Also, the reinterpret_cast here is fugly... why not just `cmd_buff[1] = 1` ?

archshift:
Keep in mind #874 - the 3DS uses `ldrb`/`strb` so technically `(char*)cmd_buff[4] = val` is correct.

Subv:
Then we need a better way to assign byte values to the command buffer without corrupting the rest of the buffer.

yuriks:
That is only important when reading from the buffer, in which case we ahould simply mask the value. The calling convention for services always pads smaller individual items (i.e. not structs) to a word in the cmdbuf, so I think we should just do the u32 assignment.

mailwl:
i don't understand. cmd_buf[1] - is Result type, 32 bit, and zero - success, why 1 is correct?

src/core/hle/service/csnd_snd.cpp
esoteric-programmer: @mailwl The byte must be set to 1 to indicate success. See http://www.3dbrew.org/wiki/CSND_Shared_Memory#Type0_Command_Structure

src/core/hle/service/csnd_snd.cpp
```diff
@@ -13,7 +13,7 @@ namespace CSND_SND {
 const Interface::FunctionInfo FunctionTable[] = {
     {0x00010140, nullptr,               "Initialize"},
     {0x00020000, nullptr,               "Shutdown"},
-    {0x00030040, nullptr,               "ExecuteType0Commands"},
+    {0x00030040, ExecuteType0Commands,   "ExecuteType0Commands"},
```
purpasmart96: This is bit nit picky, but could realign the string with other ones?

src/core/hle/service/csnd_snd.cpp
Subv: This should be done in the SharedMemory though, not in the command buffer

src/core/hle/service/csnd_snd.cpp
esoteric-programmer: You mean, change "ExecuteType0Commands" to something else?
What would you prefer? The old functon name "csnd_type0command" from commit 54a280e, but this time for the function and the string?

yuriks: No, he means that the 3rd column is misaligned.

yuriks:
This implementation is still, afaik, wrong. I couldn't find any indication that it should be writing 1, as a single byte, there.

esoteric-programmer:
I'm open to change it to cmd_buff[1] = 1.
I haven't changed it yet, because @archshift said it would be correct.

src/core/hle/service/csnd_snd.cpp
esoteric-programmer: I see!

yuriks:
In the context of the original discussion. I don't think it should be set to 1 at all, since that indicates an error result.

esoteric-programmer:
But set to 0, homebrew using CSND hangs. Set to 1, it runs fine...

yuriks:
Probably because of this check exiting out of the function early in case of an error: https://github.com/smealum/ctrulib/blob/master/libctru/source/services/csnd.c#L230-L235
Still, wrong way to fix this.

esoteric-programmer:
Okay. I think I understood the problem.

src/core/hle/service/csnd_snd.cpp
```diff
@@ -31,4 +33,51 @@ Interface::Interface() {
...
+    cmd_buff[2] = 0x4000000;
+    cmd_buff[3] = Kernel::g_handle_table.Create(mutex).MoveFrom();
+    cmd_buff[4] = Kernel::g_handle_table.Create(shared_memory).MoveFrom();
+}
+
+void ExecuteType0Commands(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    if (shared_memory != nullptr) {
+        u8* shared_mem_success = shared_memory->GetPointer(cmd_buff[1]+4);
```
yuriks: Instead of offsetting the address here, I think it would be better to create a struct type with the layout of commands, and then cast the pointer to it's type instead. That way it's also easier to continue the implementation of a non-stub version of this call later.

src/core/hle/service/csnd_snd.h
```diff
@@ -20,4 +20,9 @@ class Interface : public Service::Interface {
     }
 };
 
+void Initialize(Service::Interface* self);
+void ExecuteType0Commands(Service::Interface* self);
+void AcquireSoundChannels(Service::Interface* self);
+void Shutdown(Service::Interface* self);
```
yuriks: The way we usually do this in other services is iirc by declaring all handlers above the command table, making them static, and then not having declarations in a header.

src/core/hle/service/csnd_snd.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include "core/hle/hle.h"
 #include "core/hle/service/csnd_snd.h"
+#include "core/hle/kernel/shared_memory.h"
+#include "core/hle/kernel/mutex.h"
```
yuriks: Sort include in alphabetical order.

yuriks:
The new implementation looks much more reasonable. I haven't verified since 3dbrew docs are poor and I don't have the time to dig through ctrulib source at the moment, but it looks like it's correct.

I left a few code organization and style comments.

src/core/hle/service/csnd_snd.cpp
esoteric-programmer: The struct would contain a mixture of at least u16 and u32 types. To be safe, it would need __attribute__ ((packed)), which is not supported by MSVC. Does citra have/need MSVC support?

src/core/hle/service/csnd_snd.h
esoteric-programmer: I have seen splitting of the command table and the functions in different cpp files with functions declared in a header.
E.g. https://github.com/citra-emu/citra/blob/master/src/core/hle/service/ir/ir_rst.cpp#L13 with service functions defined in https://github.com/citra-emu/citra/blob/master/src/core/hle/service/ir/ir.h#L21
I haven't seen handlers defined above the command table. Shall I split the declaration into two files, similar to the ir service?

src/core/hle/service/csnd_snd.cpp
lioncash: Yes

src/core/hle/service/csnd_snd.cpp
```diff
@@ -31,4 +33,51 @@ Interface::Interface() {
...
+static Kernel::SharedPtr<Kernel::Mutex> mutex = nullptr;
+
+void Initialize(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    using Kernel::MemoryPermission;
+    shared_memory = Kernel::SharedMemory::Create(cmd_buff[1], MemoryPermission::ReadWrite,
+            MemoryPermission::ReadWrite, "CSNDSharedMem");
+
+    using Kernel::Mutex;
```
lioncash: There's no need for a using directive for something that adequately fits on a single line.

src/core/hle/service/csnd_snd.h
```diff
@@ -20,4 +20,14 @@ class Interface : public Service::Interface {
...
 };
 
+struct Type0Command {
+    u16 next_command_offset;
+    u16 command_id;
+    u8  finished;
+    u8  padding[3];
+    u32 flags;
+    u8  parameters[20];
+} __attribute__ ((packed));
```
yuriks: `packed` is not necessary on structs with natural alignment (which is the case for all 3DS structs I've ever seen) and I do not recommend using it here either.

esoteric-programmer: Sorry, I didn't want to commit that version. I forgot to add the current state before commit.

bunnei:
LGTM. The history is wonky though - can you squash please? Thank you

src/core/hle/service/csnd_snd.cpp
```diff
@@ -10,12 +12,17 @@
 
 namespace CSND_SND {
 
+void Initialize(Service::Interface* self);
```
purpasmart96: these should be in the header

Issue #1166, opened 2015-09-19
3dsx RomFS support


Cruel:
Currently gives error:

    Service.FS <Error> core/hle/service/fs/fs_user.cpp:OpenFileDirectly:119: failed to get a handle for archive archive_id=0x00000003 archive_path=[Empty]

Presumably from this location in ctrulib: https://github.com/smealum/ctrulib/blob/master/libctru/source/romfs_dev.c#L142

And of course it fails to init romfs and continues loading attempts from sdmc.

fincs:
From IRC today:
```
<fincs> It just takes exposing the embedded layer3 romFS inside it as the RomFS FS archive
<fincs> 3dsx romFS is defined by extended fields in the header that launchers are supposed to ignore
<profi200> I think Citra should follow the 3dsx format and also should provide argc + argv[]. If it only would provide arg[0] apps with RomFS should work.
<fincs> It cannot provide argv[0] if the 3dsx is outside the filesystem folder
<fincs> And in order for it to try using argv[0] it needs to provide the service override list
<fincs> The presence of which is the official way to detect a homebrew loader
<fincs> So, it's simply easier and troublefree to just emulate the RomFS archive
```
[Extended header reference](https://github.com/smealum/ctrulib/blob/master/libctru/source/romfs_dev.c#L97)
[How to detect the presence of 3dsx-romFS](https://github.com/smealum/ctrulib/blob/master/libctru/source/romfs_dev.c#L145)

Cruel:
Okay thanks, think I got it working, but gotta work. I'll test it and do PR tonight.

vgturtle127:
We can close this now right? The PR was merged...

Cruel:
Don't know. May want to keep it for sentimental reasons... :sparkles:

PR #1167, opened 2015-09-21
Implement 3dsx RomFS

Cruel:
Addresses #1166 

@fincs I'm not sure if the size is right. I kinda just assumed that `filesize - offset = romfs_size`

So if the RomFS isn't appended like that, it needs to be changed.

That being said, I tested it with a few RomFS-enabled 3dsx files, and also a few 3dsx files that didn't use it. Works as expected, and gracefully fails to open RomFS when not there too.

### DIFF
--- a/src/core/loader/3dsx.cpp
+++ b/src/core/loader/3dsx.cpp
@@ -62,6 +62,10 @@ struct THREEDSX_Header
     // Sizes of the code, rodata and data segments +
     // size of the BSS section (uninitialized latter half of the data segment)
     u32 code_seg_size, rodata_seg_size, data_seg_size, bss_size;
+    // offset and size of smdh
+    u32 smdh_offset, smdh_size;
+    // offset to filesystem
+    u32 fs_offset;
 };
 
 // Relocation header: all fields (even extra unknown fields) are guaranteed to be relocation counts.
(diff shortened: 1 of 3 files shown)


src/core/loader/3dsx.cpp
```diff
@@ -267,4 +271,37 @@ ResultStatus AppLoader_THREEDSX::Load() {
...
+
+    // Reset read pointer in case this file has been read before.
+    file.Seek(0, SEEK_SET);
+
+    THREEDSX_Header hdr;
+    if (file.ReadBytes(&hdr, sizeof(THREEDSX_Header)) != sizeof(THREEDSX_Header))
+        return ResultStatus::Error;
+
+    // Check if the 3DSX has a RomFS...
+    if (hdr.fsOffset) {
```
archshift: `if (hdr.fsOffset != 0)`

src/core/loader/3dsx.cpp
```diff
@@ -267,4 +271,37 @@ ResultStatus AppLoader_THREEDSX::Load() {
...
+        u32 romfs_offset = hdr.fsOffset;
+        u32 romfs_size = ((file.GetSize() - hdr.fsOffset));
```
archshift: These two sets of parentheses are unnecessary.

archshift:
LGTM

Cruel:
Updated.

From what I can tell, the RomFS offset was added to 3dsx header for current 3dsxtool. So should we be concerned about possible undefined behavior when running 3dsx files built with old tool?

The header has a `header_size` field. Should we just check to see if the header size field matches the expected header?

fincs:
> The header has a  header_size  field. Should we just check to see if the header size field matches the expected header to weed out old headers?

Yes.

src/core/loader/3dsx.cpp
```diff
@@ -62,6 +62,10 @@ struct THREEDSX_Header
     // Sizes of the code, rodata and data segments +
     // size of the BSS section (uninitialized latter half of the data segment)
     u32 code_seg_size, rodata_seg_size, data_seg_size, bss_size;
+    // offset and size of smdh
+    u32 smdhOffset, smdhSize;
```
mattvail: u32 smdh_offset, smdh_size;

src/core/loader/3dsx.cpp
```diff
@@ -62,6 +62,10 @@ struct THREEDSX_Header
...
+    // offset to filesystem
+    u32 fsOffset;
```
mattvail: u32 fs_offset;

bunnei:
LGTM

PR #1172, opened 2015-09-25
Fix some xcode 7 (llvm) warnings

martinlindhe:
Also tested compiling under Ubuntu with these changes, couldn't spot any issues.

### DIFF
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -967,6 +967,8 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
                         UNIMPLEMENTED();
                         break;
                     }
+
+                    return {};
                 };
 
                 auto LookupFactorA = [&](Regs::BlendFactor factor) -> u8 {
@@ -1000,6 +1002,8 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
                         UNIMPLEMENTED();
                         break;
                     }
+
+                    return {};
                 };
 
                 static auto EvaluateBlendEquation = [](const Math::Vec4<u8>& src, const Math::Vec4<u8>& srcfactor,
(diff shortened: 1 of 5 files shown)


src/video_core/rasterizer.cpp
```diff
@@ -967,6 +967,8 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
                         UNIMPLEMENTED();
                         break;
                     }
+
+                    return NULL;
```
yuriks: The value being returned isn't a pointer, so this makes no sense. Return a zeroed vector instead or something.

src/video_core/rasterizer.cpp
```diff
@@ -1000,6 +1002,8 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
```
yuriks: Likewise.

martinlindhe:
@yuriks: thanks, i have updated the patch with your feedback

src/video_core/rasterizer.cpp
```diff
@@ -967,6 +967,8 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
                         UNIMPLEMENTED();
                         break;
                     }
+
+                    return 0;
```
bunnei: This should return `Math::Vec3<u8>(0, 0, 0);`

bunnei: Actually, just `{}` should be fine, I think

bunnei:
LGTM after my one comment is addressed. Please squash down to a single commit.

martinlindhe:
@bunnei: updated the patch from your feedback and squashed into a single commit

bunnei:
@martinlindhe did you screw up the rebase? You have some of @lioncash's commits in there now...

martinlindhe:
@bunnei: jeez, sorry for the trouble. should be ok now

bunnei:
LGTM

PR #1175, opened 2015-09-30
CMakeLists: Compile as C++14 on non-windows targets

lioncash:
MSVC 2015 supports some aspects of C++14, such as constexpr std::min/std::max, etc.


bunnei:
LGTM

lioncash:
Updated to also remove make_unique.h as it will not be needed after this (since MSVC 2015 supports make_unique)

yuriks:
VS has always supported make_unique, the reason for that one is gcc

Kloen:
@yuriks wasn't it added on [GCC 4.9](https://gcc.gnu.org/gcc-4.9/changes.html) ?

lioncash:
Close enough :^)

yuriks:
@Kloen: I think travis (or maybe neobrain) still used a distro with 4.8.

yuriks:
Shouldn't you also remove make_unique.h from common if you're getting rid of it?

lioncash:
@yuriks Whoops, amended that

neobrain:
For reference, indeed the latest opensuse release (which will be superseded soon though) is still shipping with gcc 4.8 by default.

archshift:
@neobrain, can't you upgrade to 4.9 or even 5.0 though?

bunnei:
What is the status of this PR? Can we not merge this because certain mainstream Linux distros don't ship with a new enough version of GCC?

neobrain:
Yeah, let's not do this, at least not for another few months.

lioncash:
I never thought I'd see the day Linux distros drop the ball compared to Windows + MSVC – incredible

Lectem:
I'd suggest using [cmake compile features](https://cmake.org/cmake/help/v3.1/manual/cmake-compile-features.7.html#manual:cmake-compile-features) instead of raw flags, or use the [CXX_STANDARD property](https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html)

bunnei:
Closing this PR for now, we'll revisit in 6 months or so. I've created #1245 so that this does not fall off the table.

PR #1176, opened 2015-09-30
Obligatory "Throw out workarounds VS2013 once limited us to" PR

lioncash:


src/common/bit_field.h
```diff
@@ -125,21 +125,15 @@ struct BitField
...
     // default copy assignment would copy the full storage value, rather than
     // just the bits relevant to this particular bit field.
     // Ideally, we would just implement the copy assignment to copy only the
     // relevant bits, but this requires compiler support for unrestricted
     // unions.
-    // MSVC 2013 has no support for this, hence we disable this code on
-    // Windows (so that the default copy assignment operator will be used).
-    // For any C++11 conformant compiler we delete the operator to make sure
-    // we never use this inappropriate operator to begin with.
     // TODO: Implement this operator properly once all target compilers
```
Subv: Do all of our target compilers now support unrestricted unions?

yuriks: VC2015 does, at least. (Dunno how well.)

src/video_core/command_processor.cpp
```diff
@@ -74,9 +74,9 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
             break;
 
         // Load default vertex input attributes
-        case PICA_REG_INDEX_WORKAROUND(vs_default_attributes_setup.set_value[0], 0x233):
-        case PICA_REG_INDEX_WORKAROUND(vs_default_attributes_setup.set_value[1], 0x234):
-        case PICA_REG_INDEX_WORKAROUND(vs_default_attributes_setup.set_value[2], 0x235):
+        case PICA_REG_INDEX(vs_default_attributes_setup.set_value[0]):
+        case PICA_REG_INDEX(vs_default_attributes_setup.set_value[1]):
+        case PICA_REG_INDEX(vs_default_attributes_setup.set_value[2]):
```
yuriks: RIP convenient reg numbers :(

lioncash: Nope, MSVC still can't do it.

src/common/bit_field.h
neobrain: FYI, after having used BitField for a while I realized it's a very bad idea to implement this operator as suggested in this comment. It would just cause more trouble than it's worth.

lioncash: I've updated the commit to remove that text as well.

bunnei:
LGTM

PR #1177, opened 2015-09-30
Use a constexpr function for country initialization in service/cfg

linkmauve:
This fixes a TODO left over from when we supported MSVC 2013.


lioncash:
LGTM

bunnei:
LGTM

PR #1178, opened 2015-09-30
ivfc_archive: Fix a printf specifier

lioncash:
Now that MSVC 2015 supports the %z family of specifiers, this can be used.


bunnei:
LGTM

Issue #1183, opened 2015-10-02
Fix linux shader (compilation?)


Cpasjuste:
Hi,

Citra does hang for a few seconds/minutes on linux at start. This render the whole thing unusable for developing. I think it's related to the shader compilation code. Unfortunately i don't think i have the knowledge to resolve this bug ... 

Would be much, much appreciated if someone would take a look :)

Edit: the problem is located in OGLShader.Create, disabling the "ShaderUtil::LoadShaders" function does get homebrew to work normaly (i guess homebrews who doesnt request gpu access).

Edit2: Ubuntu 15.04 - x64

lybxlpsv:
you should tell which mesa driver version and graphics card you are using, might also want to upload glxinfo output somewhere.

yuriks:
Dupe of #819

Issue #819, opened 2015-05-25
OpenGL rasterizer’s fragment shader doesn’t compile on i965 on Sandy Bridge


linkmauve:
This bug is tracking the following Mesa bug: https://bugs.freedesktop.org/show_bug.cgi?id=90631

linkmauve:
This is fixed in the next version of Mesa (see https://bugs.freedesktop.org/show_bug.cgi?id=86469#c16) and we are also dropping this mega-shader in PR #1187.

yuriks:
We haven't merged #1187 yet, and most people don't have the updated mesa versions, so I think we should leave this unclosed for now.

Subv:
#1187 has been merged

PR #1187, opened 2015-10-06
GLSL Fragment Shader Generation

bunnei:
This PR adds an initial implementation of GLSL fragment shader generation for the OpenGL HW renderer. It is based off of an initial implementation from @Subv, and some improvements and fixes made by @yuriks. I fixed the remaining (known) bugs and cleaned it up a lot. There are two motivating factors for this PR:
* We previously used a single GLSL shader to emulate TEV/fragment operations, and it was quite gnarly. If you run Citra on an IGP, this may have been a performance bottleneck for you. Instead, these changes now generate fragment shaders (and cache them) as-needed for various Pica configurations, greatly reducing the size of most shader programs and basically removing all branches from the shader code (which are slow). This should improve speed pretty significantly on low-end GPUs, and less significantly (but still noticeable) on discrete GPUs.
* More advanced Pica features (e.g. per-fragment lighting, bump mapping, fog, etc.) are sufficiently complex that it would simply be impractical to keep the "one shader fits all" approach. As I've been dabbling with implementing some of these features in parallel, it was important to me to get this groundwork in place.

EDIT: I recommend reviewing this PR as one large diff, not commit-by-commit. There are several large cleanups/refactors from the initial implementation.


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -592,6 +461,48 @@ void RasterizerOpenGL::ReconfigureDepthTexture(DepthTextureInfo& texture, Pica::
...
+
+    // Update global state
+    state.draw.shader_program = current_shader->shader.handle;
+    state.Apply();
+
+    // Update uniforms
+    SyncAlphaTest();
+    SyncCombinerColor();
+    auto& tev_stages = Pica::g_state.regs.GetTevStages();
+    for (int index = 0; index < tev_stages.size(); ++index)
```
lioncash: index should probably be size_t

bunnei: Fixed

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -592,6 +461,48 @@ void RasterizerOpenGL::ReconfigureDepthTexture(DepthTextureInfo& texture, Pica::
...
 
+void RasterizerOpenGL::SetShader() {
+    ShaderCacheKey config = ShaderCacheKey::CurrentConfig();
+
+    // Find (or generate) the GLSL shader for the current TEV state
+    auto cached_shader = shader_cache.find(config);
+    if (cached_shader != shader_cache.end()) {
+        current_shader = &cached_shader->second;
+    }
+    else {
```
tfarley: Else on same line as closing brace?

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+    if (config.TevStageUpdatesCombinerBufferAlpha(index))
+        out += "g_combiner_buffer.a = g_last_tex_env_out.a;\n";
+}
+
+std::string GenerateFragmentShader(const ShaderCacheKey& config) {
+    std::string out = R"(
+#version 150 core
+
+#define NUM_VTX_ATTR 7
+#define NUM_stageS 6
```
tfarley: Find+replace error? Used to be NUM_TEV_STAGES.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+in vec4 o[NUM_VTX_ATTR];
+out vec4 color;
+
+uniform int alphatest_ref;
+uniform vec4 const_color[NUM_stageS];
```
tfarley: Same here for NUM_TEV_STAGES.

src/video_core/renderer_opengl/gl_shader_util.cpp
```diff
@@ -65,6 +65,13 @@ GLuint LoadShaders(const char* vertex_shader, const char* fragment_shader) {
     GLuint program_id = glCreateProgram();
     glAttachShader(program_id, vertex_shader_id);
     glAttachShader(program_id, fragment_shader_id);
+
+    glBindAttribLocation(program_id, Attributes::ATTRIBUTE_POSITION, "vert_position");
+    glBindAttribLocation(program_id, Attributes::ATTRIBUTE_COLOR, "vert_color");
+    glBindAttribLocation(program_id, Attributes::ATTRIBUTE_TEXCOORDS + 0, "vert_texcoords0");
```
tfarley: Smallest nitpick ever: + 0 is used here but absent in rest of references to ATTRIBUTE_TEXCOORDS.

bunnei: Addressed the other instances to use `+ 0`

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -272,6 +201,7 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
     // Alpha test
     case PICA_REG_INDEX(output_merger.alpha_test):
         SyncAlphaTest();
+        state.draw.shader_dirty = true;
```
Subv: There's no point in using an uniform for the alpha test value if you're going to regenerate the shader anyways

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+            stage.GetAlphaMultiplier() == 1);
+}
+
+/// Writes the specified TEV stage source component(s)
+static void AppendSource(std::string& out, TevStageConfig::Source source,
+        const std::string& index_name) {
+    using Source = TevStageConfig::Source;
+    switch (source) {
+    case Source::PrimaryColor:
+        out += "o[2]";
```
Subv: Maybe it'd be a good idea to rename this "o" to something more meaningful

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+uniform vec4 tev_combiner_buffer_color;
+
+void main(void) {
+vec4 g_combiner_buffer = tev_combiner_buffer_color;
+vec4 g_last_tex_env_out = vec4(0.0, 0.0, 0.0, 0.0);
+)";
+
+    // Do not do any sort of processing if it's obvious we're not going to pass the alpha test
+    if (config.alpha_test_func == Regs::CompareFunc::Never) {
+        out += "discard;";
```
Subv: missing the closing ';' for the main() function

src/video_core/renderer_opengl/gl_rasterizer.cpp
bunnei: <bunnei> Subv, "There's no point in using an uniform for the alpha test value if you're going to regenerate the shader anyways" -> the uniform is just the reference, we still care if the enable and alpha test function change
<Subv> we're not checking for that, are we?
<Subv> you're just setting the uniform, then regenerating the shader and setting the uniform again immediately after
<bunnei> any time that register is written too, we automatically upload the uniform (for the alpha ref) and signal to check if the shader needs to be regenerated (because parts of that register are used in shader generation)
<bunnei> setting the dirty flag causes a cache lookup, not necessarily a new shader to be generated
<Subv> ah, right
<Subv> nevermind then

src/common/common_funcs.h
```diff
@@ -95,3 +95,21 @@ inline u64 _rotr64(u64 x, unsigned int shift){
 // This function might change the error code.
 // Defined in Misc.cpp.
 const char* GetLastErrorMsg();
+
+#include <cstddef>
```
neobrain: lack of docstrings... also, don't place includes at the top of a file

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -4,15 +4,123 @@
...
+        const auto& regs = Pica::g_state.regs;
+        ShaderCacheKey config;
+
+        config.alpha_test_func = regs.output_merger.alpha_test.enable ?
+            regs.output_merger.alpha_test.func.Value() : Pica::Regs::CompareFunc::Always;
+
+        // Copy relevant TevStageConfig fields only. We're doing this manually (instead of calling
+        // the GetTevStages() function) because BitField explicitly disables copies.
+
+        config.tev_stages[0].source_raw = regs.tev_stage0.source_raw;
```
neobrain: You should store source_raw & 0x0FFF0FFF instead, since there are a number of unused and unknown fields there. (that masking should be done as a member function in the register definition)

src/common/common_funcs.h
bunnei: include is not at the top of the file?

src/video_core/renderer_opengl/gl_rasterizer.h
bunnei: I think we are 99% sure these are unused fields in the register... IMO the extra function just adds cruft (and it's not like it would be a bug in cache key generation if these were used, worst case scenario, an extra shader is unnecessarily generated)

src/common/common_funcs.h
neobrain: ?

bunnei: The include is placed where it is used, which is not at the top of the file. So I don't understand why you said "don't place includes at the top of a file"

neobrain: Woops. Meant to say the opposite.

Includes shouldn't happen somewhere in the middle of a file.

bunnei:
Call for more reviews on this. If I don't hear anything else by the weekend, I'm going to assume everyone is good with it.

Fredabill:
Runs at same speed as master branch when playing Animal Crossing: New Leaf
Also has more texture glitches, strange black spots on the ground...
Running on AMD A8-4500M APU with Radeon HD Graphics 1.90 GHz
acnl

Subv:
lgtm

bunnei:
Looks fine on my GeForce GTX 860M:

Anyone else with AMD care to test?

bunnei:
@Fredabill are your drivers up to date, and you have no local mods? Unfortunately there's not a whole lot that I can do without someone else duplicating the issue and giving me more debug information

bunnei:
@Normmatt tested this on his AMD Radeon HD7970, and the results are the same:

bunnei:
On second thought, the water is broken on AMD there

bunnei:
<pre>[20:28] <@Subv> bunnei: consider it lgtmd</pre>

bunnei:
<pre>[20:03] <Fredabill> bunnei, fixed!</pre>

linkmauve:
After some quite extensive testing on i965 on Ivy Bridge and Sandy Bridge, here are my impressions:
- The biggest benefit is that #819 is fixed (on our end), as there is no humongous fragment shader anymore, the Sandy Bridge emitter in i965 can work properly.
- On Ivy Bridge, the frame duration is reduced by about 20~25% compared to master, which is pretty nice. :)
- At first I thought there was a regression in MM3D’s intro, instead of the scene of the Stock Pot Inn interior the top screen is black, and every glDrawArrays will emit GL_INVALID_FRAMEBUFFER_OPERATION.  But this one is present on master as well, only on the OpenGL renderer.
- Overall I haven’t noticed any regression, and instead only improvements compared to master.
- I haven’t looked at the code at all yet, I’ll try to do it tomorrow morning, but if I don’t consider this my blind LGTM.

bunnei:
This should be good to go. Final call for reviews, otherwise I will be merging this weekend.

mailwl:
@bunnei, yes, please merge. I want to see, how "Time Travelers" will look

Fredabill:
@bunnei My issue is now fixed, no water issue or black spots

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -592,6 +462,49 @@ void RasterizerOpenGL::ReconfigureDepthTexture(DepthTextureInfo& texture, Pica::
     state.Apply();
 }
 
+void RasterizerOpenGL::SetShader() {
+    std::unique_ptr<PicaShader> shader = Common::make_unique<PicaShader>();
```
archshift: Didn't we get rid of `Common::make_unique` after the move to VS2015?

lioncash: #1175 will need to be merged

src/video_core/pica.h
```diff
@@ -317,6 +317,7 @@ struct Regs {
         };
 
         union {
+            u32 source_raw;
```
yuriks: "sources_raw", maybe? Same for the rest, since each reg has several source fields.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -78,29 +50,27 @@ void RasterizerOpenGL::InitObjects() {
...
     state.draw.vertex_buffer = vertex_buffer.handle;
-    state.draw.shader_program = shader.handle;
-
     state.Apply();
 
-    // Set the texture samplers to correspond to different texture units
-    glUniform1i(uniform_tex, 0);
-    glUniform1i(uniform_tex + 1, 1);
-    glUniform1i(uniform_tex + 2, 2);
+    glBufferData(GL_ARRAY_BUFFER, 0, nullptr, GL_STREAM_DRAW);
```
yuriks: What's the purpose of this?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -592,6 +462,49 @@ void RasterizerOpenGL::ReconfigureDepthTexture(DepthTextureInfo& texture, Pica::
...
 }
 
+void RasterizerOpenGL::SetShader() {
+    std::unique_ptr<PicaShader> shader = Common::make_unique<PicaShader>();
+    u64 hash = shader->config.GetHash();
+
+    // Find (or generate) the GLSL shader for the current TEV state
+    auto cached_shader = shader_cache.find(hash);
+    if (cached_shader != shader_cache.end()) {
+        ASSERT_MSG(shader->config == cached_shader->second->config, "Hash collision occurred with shader generation!");
```
yuriks: wtf?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -592,6 +462,49 @@ void RasterizerOpenGL::ReconfigureDepthTexture(DepthTextureInfo& texture, Pica::
     state.Apply();
 }
 
+void RasterizerOpenGL::SetShader() {
+    std::unique_ptr<PicaShader> shader = Common::make_unique<PicaShader>();
+    u64 hash = shader->config.GetHash();
+
+    // Find (or generate) the GLSL shader for the current TEV state
+    auto cached_shader = shader_cache.find(hash);
```
yuriks: You're only using the hash to look up the shader, you should use the entire key.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -592,6 +462,49 @@ void RasterizerOpenGL::ReconfigureDepthTexture(DepthTextureInfo& texture, Pica::
...
+        state.draw.shader_program = current_shader->shader.handle;
+        state.Apply();
+    } else {
+        LOG_CRITICAL(Render_OpenGL, "Creating new shader: %llx", hash);
+
+        shader->shader.Create(GLShader::GenerateVertexShader().c_str(), GLShader::GenerateFragmentShader(shader->config).c_str());
+        shader->uniform_alphatest_ref = 0;
+        shader->uniform_tev_const_colors = 1;
+        shader->uniform_tex = 7;
+        shader->uniform_tev_combiner_buffer_color = 10;
```
yuriks: What's the point of these members if you're just hard-coding the values?

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -4,15 +4,94 @@
...
+ * This struct contains all state used to generate the GLSL shader program that emulates the current
+ * Pica register configuration. This struct is used as a cache key for generated GLSL shader
+ * programs. The functions in gl_shader_gen.cpp should retrieve state from this struct only, not by
+ * directly accessing Pica registers. This should reduce the risk of bugs in shader generation where
+ * Pica state is not being captured in the shader cache key, thereby resulting in (what should be)
+ * two separate shaders sharing the same key.
+ */
+struct PicaShaderConfig {
+    /// Construct a PicaShaderConfig with the current Pica register configuration.
+    PicaShaderConfig() {
```
yuriks: I don't think the default constructor should do all this work. The default constructor should be just empty, or clear the struct to zero or similar. Then make a separate static function "FromPicaRegs" that takes regs as a parameter and loads the settings from it, returning an instance of the struct.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -4,15 +4,94 @@
...
+
+    bool TevStageUpdatesCombinerBufferColor(unsigned stage_index) const {
+        return (stage_index < 4) && (combiner_buffer_input & (1 << stage_index));
+    }
+
+    bool TevStageUpdatesCombinerBufferAlpha(unsigned stage_index) const {
+        return (stage_index < 4) && ((combiner_buffer_input >> 4) & (1 << stage_index));
+    }
+
+    u64 GetHash() const {
```
yuriks: Instead (or in addition) to this, specialize the std hash stuff to be able to hash this type. This will make it usable automatically with the std containers.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -4,15 +4,94 @@
...
+        tev_stages[3].scale_raw = regs.tev_stage3.scale_raw;
+        tev_stages[4].scale_raw = regs.tev_stage4.scale_raw;
+        tev_stages[5].scale_raw = regs.tev_stage5.scale_raw;
+
+        combiner_buffer_input =
+            regs.tev_combiner_buffer_input.update_mask_rgb.Value() |
+            regs.tev_combiner_buffer_input.update_mask_a.Value() << 4;
+    }
+
+    bool TevStageUpdatesCombinerBufferColor(unsigned stage_index) const {
```
yuriks: This function (and the one below it) are duplicating some functionality from elsewhere in the source that was marked private. For my prototype I had simply re-implemented the function, but I don't think we should have this duplication in the tree since it'll eventually lead to bugs.

yuriks: (In case this isn't the same thing, then disregard since I must be thinking of something else. In any case, this will probably need some modification later to fix the combiner buffer update stage thing.)

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -4,15 +4,94 @@
...
+    }
+
+    bool operator ==(const PicaShaderConfig& o) const {
+        return std::memcmp(this, &o, sizeof(PicaShaderConfig)) == 0;
+    };
+
+    Pica::Regs::CompareFunc alpha_test_func;
+    std::array<Pica::Regs::TevStageConfig, 6> tev_stages = {};
+    u8 combiner_buffer_input;
+};
```
yuriks: Should perhaps add some static_asserts to verify it's POD/standard layout/whatever as necessary.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -213,21 +295,11 @@ class RasterizerOpenGL : public HWRasterizer {
     std::array<SamplerInfo, 3> texture_samplers;
     TextureInfo fb_color_texture;
     DepthTextureInfo fb_depth_texture;
-    OGLShader shader;
+
+    std::unordered_map<u64, std::unique_ptr<PicaShader>> shader_cache;
```
yuriks: This map should be keyed by the actual key, not a hash of the key... otherwise you throw away collision resolution.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+std::string GenerateFragmentShader(const PicaShaderConfig& config) {
+    std::string out = R"(
+#version 330
+#extension GL_ARB_explicit_uniform_location : require
+
+#define NUM_VTX_ATTR 7
+#define NUM_TEV_STAGES 6
+
+in vec4 attr[NUM_VTX_ATTR];
```
yuriks: Is there any advantage to having this as an array instead of as discrete variables with actual names (since the meaning of each attribute is hardcoded)? Otherwise, please add #defines for symbolic names for each attribute.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        out += "g_combiner_buffer.rgb = g_last_tex_env_out.rgb;\n";
+
+    if (config.TevStageUpdatesCombinerBufferAlpha(index))
+        out += "g_combiner_buffer.a = g_last_tex_env_out.a;\n";
+}
+
+std::string GenerateFragmentShader(const PicaShaderConfig& config) {
+    std::string out = R"(
+#version 330
+#extension GL_ARB_explicit_uniform_location : require
```
yuriks: How widespread is this extension?

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+#define NUM_VTX_ATTR 7
+#define NUM_TEV_STAGES 6
+
+in vec4 attr[NUM_VTX_ATTR];
+out vec4 color;
+
+layout(location = 0) uniform int alphatest_ref;
```
yuriks: What was the problem that required introducing the explicit uniform locations? (I know I suggested this on IRC, but I was only following the conversation lightly and was talking about the input vertex attributes.) Why does it fix it?

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+layout(location = 1) uniform vec4 const_color[NUM_TEV_STAGES];
+layout(location = 7) uniform sampler2D tex[3];
+layout(location = 10) uniform vec4 tev_combiner_buffer_color;
+
+void main(void) {
```
yuriks: The second void is redundant.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+vec4 g_last_tex_env_out = vec4(0.0, 0.0, 0.0, 0.0);
+)";
+
+    // Do not do any sort of processing if it's obvious we're not going to pass the alpha test
+    if (config.alpha_test_func == Regs::CompareFunc::Never) {
+        out += "discard; }";
+        return out;
+    }
+
+    for (std::size_t index = 0; index < config.tev_stages.size(); ++index)
```
yuriks: Do we really need the `std::` here? :|

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        WriteTevStage(out, config, (unsigned)index);
+
+    if (config.alpha_test_func != Regs::CompareFunc::Always) {
+        out += "if (";
+        AppendAlphaTestCondition(out, config.alpha_test_func);
+        out += ") {\n discard;\n }\n";
```
yuriks: Since you're not indenting the code correctly anyway, might as well drop the `\n`s.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+out vec4 color;
+
+layout(location = 0) uniform int alphatest_ref;
+layout(location = 1) uniform vec4 const_color[NUM_TEV_STAGES];
+layout(location = 7) uniform sampler2D tex[3];
+layout(location = 10) uniform vec4 tev_combiner_buffer_color;
+
+void main(void) {
+vec4 g_combiner_buffer = tev_combiner_buffer_color;
+vec4 g_last_tex_env_out = vec4(0.0, 0.0, 0.0, 0.0);
```
yuriks: Remove the `g_` prefix from both of these since they're not globals anymore.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        out += ") {\n discard;\n }\n";
+    }
+
+    out += "color = g_last_tex_env_out;\n}";
+
+    return out;
+}
+
+std::string GenerateVertexShader() {
+    static const std::string out = R"(
```
yuriks: This can just be a const char*, and the function can just return that too.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+in vec2 vert_texcoords0;
+in vec2 vert_texcoords1;
+in vec2 vert_texcoords2;
+
+out vec4 attr[NUM_VTX_ATTR];
+
+void main() {
+    attr[2] = vert_color;
+    attr[3] = vec4(vert_texcoords0.xy, vert_texcoords1.xy);
+    attr[5] = vec4(0.0, 0.0, vert_texcoords2.xy);
```
yuriks: Again, I don't see why we have this attr array. Like, what's the point of adding the padding elements at the start? O_o (I'm aware this was already there before this PR, but it's pointless to maintain it.)

src/video_core/renderer_opengl/gl_shader_util.cpp
yuriks: Shouldn't we just do `ATTRIBUTE_TEXCOORD0`, etc. instead then? They seem to always be separate.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+    case Source::PrimaryColor:
+        out += "attr[2]";
+        break;
+    case Source::PrimaryFragmentColor:
+        // HACK: Until we implement fragment lighting, use primary_color
+        out += "attr[2]";
+        break;
+    case Source::SecondaryFragmentColor:
+        // HACK: Until we implement fragment lighting, use zero
+        out += "vec4(0.0, 0.0, 0.0, 0.0)";
```
yuriks: `vec4(0.0)`

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        out += "texture(tex[1], attr[3].zw)";
+        break;
+    case Source::Texture2: // TODO: Unverified
+        out += "texture(tex[2], attr[5].zw)";
+        break;
+    case Source::PreviousBuffer:
+        out += "g_combiner_buffer";
+        break;
+    case Source::Constant:
+        out += "const_color[" + index_name + "]";
```
yuriks:
```
out += "const_color["; out += index_name; out += "]";
```
(or if you want to be clever: `((out += "const_color[") += index_name) += ']';`)

This prevents creation and allocating of temporary string objects

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+    switch (operation) {
+    case Operation::Replace:
+        out += variable_name + "[0]";
+        break;
+    case Operation::Modulate:
+        out += variable_name + "[0] * " + variable_name + "[1]";
+        break;
+    case Operation::Add:
+        out += "min(" + variable_name + "[0] + " + variable_name + "[1], vec3(1.0))";
```
yuriks: Use `clamp(..., 0.0, 1.0)` here (as per IRC discussion about saturate)

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        break;
+    case Operation::AddSigned:
+        out += "clamp(" + variable_name + "[0] + " + variable_name + "[1] - vec3(0.5), vec3(0.0), vec3(1.0))";
+        break;
+    case Operation::Lerp:
+        out += variable_name + "[0] * " + variable_name + "[2] + " + variable_name + "[1] * (vec3(1.0) - " + variable_name + "[2])";
+        break;
+    case Operation::Subtract:
+        out += "max(" + variable_name + "[0] - " + variable_name + "[1], vec3(0.0))";
```
yuriks: clamp

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        out += variable_name + "[0] * " + variable_name + "[1]";
+        break;
+    case Operation::Add:
+        out += "min(" + variable_name + "[0] + " + variable_name + "[1], vec3(1.0))";
+        break;
+    case Operation::AddSigned:
+        out += "clamp(" + variable_name + "[0] + " + variable_name + "[1] - vec3(0.5), vec3(0.0), vec3(1.0))";
+        break;
+    case Operation::Lerp:
+        out += variable_name + "[0] * " + variable_name + "[2] + " + variable_name + "[1] * (vec3(1.0) - " + variable_name + "[2])";
```
yuriks: Can't you just use the `lerp` builtin?

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        break;
+    case Operation::Subtract:
+        out += "max(" + variable_name + "[0] - " + variable_name + "[1], vec3(0.0))";
+        break;
+    case Operation::MultiplyThenAdd:
+        out += "min(" + variable_name + "[0] * " + variable_name + "[1] + " + variable_name + "[2], vec3(1.0))";
```
yuriks: clamp

src/video_core/renderer_opengl/gl_shader_gen.cpp
yuriks: Also, you don't need to explicitly construct the `vec3(1.0)`, just `1.0` will do. This is done through most of the shader generator and should be fixed everywhere it appears. (I have verified that this is legal GLSL using glslValidator, which should be spec accurate.)

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        out += "clamp(" + variable_name + "[0] + " + variable_name + "[1] - 0.5, 0.0, 1.0)";
+        break;
+    case Operation::Lerp:
+        out += variable_name + "[0] * " + variable_name + "[2] + " + variable_name + "[1] * (1.0 - " + variable_name + "[2])";
+        break;
+    case Operation::Subtract:
+        out += "max(" + variable_name + "[0] - " + variable_name + "[1], 0.0)";
+        break;
+    case Operation::MultiplyThenAdd:
+        out += "min(" + variable_name + "[0] * " + variable_name + "[1] + " + variable_name + "[2], 1.0)";
```
yuriks: clamp

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        break;
+    case Operation::AddThenMultiply:
+        out += "min(" + variable_name + "[0] + " + variable_name + "[1], 1.0) * " + variable_name + "[2]";
```
yuriks: clamp

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        AppendAlphaModifier(out, stage.alpha_modifier3, stage.alpha_source3, index_name);
+        out += ");\n";
+
+        out += "float alpha_output_" + index_name + " = ";
+        AppendAlphaCombiner(out, stage.alpha_op, "alpha_results_" + index_name);
+        out += ";\n";
+
+        out += "g_last_tex_env_out = vec4("
+            "clamp(color_output_" + index_name + " * " + std::to_string(stage.GetColorMultiplier()) + ".0, 0.0, 1.0),"
+            "clamp(alpha_output_" + index_name + " * " + std::to_string(stage.GetAlphaMultiplier()) + ".0, 0.0, 1.0));\n";
```
yuriks: With the amount that `clamp(..., 0.0, 1.0)` is used in the shaders, it probably makes sense to define `saturate` as a helper function in the header and then use that throughout.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -744,55 +656,17 @@ void RasterizerOpenGL::SyncDepthTest() {
...
-void RasterizerOpenGL::SyncTevMultipliers(unsigned stage_index, const Pica::Regs::TevStageConfig& config) {
-    glUniform2i(uniform_tev_cfgs[stage_index].color_alpha_multiplier, config.GetColorMultiplier(), config.GetAlphaMultiplier());
-}
-
 void RasterizerOpenGL::SyncCombinerColor() {
-    auto combiner_color = PicaToGL::ColorRGBA8(Pica::g_state.regs.tev_combiner_buffer_color.raw);
-    glUniform4fv(uniform_tev_combiner_buffer_color, 1, combiner_color.data());
+    if (current_shader->uniform_tev_combiner_buffer_color != -1) {
+        auto combiner_color = PicaToGL::ColorRGBA8(Pica::g_state.regs.tev_combiner_buffer_color.raw);
+        glUniform4fv(current_shader->uniform_tev_combiner_buffer_color, 1, combiner_color.data());
```
yuriks: One thing you have to watch out in glUniform calls is that they apply to the currently bound shader, so you may have to ensure `state.Apply` has been called since the last time. (These kinds of situations are why our state tracker is very poor when it comes to actually reducing redundant state changes.)

yuriks:
Review pass. The main thing is that the cache is still being keyed wrong. There are also several minor things in shader generation. I quite dislike the workaround of specifying fixed uniform locations, unless there is an actual rationale for it, since it requires careful management of the indices so that uniforms that take more than one index don't overlap each other. (And also spread a bunch of magic number through the source.)

src/video_core/renderer_opengl/gl_shader_gen.cpp
hrydgard: Good question but explicit uniform locations really is the way to go. Implicit locations is one of the worst misfeatures 3DLabs added to GLSL in OpenGL 2.0, and it's essentially deprecated. None of the new APIs have anything like it.

The main advantage with explicit locations is that you don't have to track a bunch of redundant numbers (the queried locations) along with each shader that uses a particular setup. You will have many shaders sharing essentially the same set of uniforms, and you will want to move to uniform buffers anyway eventually.

yuriks: I'm mostly hesitant about it because:
 - They introduce a dependency on a 4.x extension
 - The way the explicit locations have been integrated in the rest of the source is really shoddy
 - I think it's error prone to make sure you left enough slots for matrices or arrays when specifying the locations

If anything, I agree we should move to a std140 uniform buffer instead, which can be just represented on the client-side as a C struct.

hrydgard: Yeah, fair enough about the version requirement. On the other hand, if citra will move to UBOs anyway, having explicit uniform locations can be a stopgap until then. It's not like you already have to maintain compatibility with thousands of strange mobile devices with Mali 400 GPUs like *cough* someone here :)

Cpasjuste:
Hi,

I confirm that this also fix #1183 and that's great for linux users.

GL_VERSION: 3.3 (Core Profile) Mesa 10.5.9
GL_VENDOR: Intel Open Source Technology Center
GL_RENDERER: Mesa DRI Intel(R) HD Graphics 5500 (Broadwell GT2)

src/video_core/renderer_opengl/gl_shader_gen.cpp
bunnei: This fixes a bug in ZMM on my Intel HD 4600

hrydgard: I think vec3(1.0) is far clearer than just passing in 1.0 in a vec3 argument, and better practice, and safer against buggy shader compilers which are VERY common. Writing the shortest possible code should not be a goal.

src/video_core/renderer_opengl/gl_rasterizer.h
bunnei: This was moved, but not duplicated... Feel free to correct if I'm wrong

src/video_core/renderer_opengl/gl_shader_gen.cpp
bunnei: Does this really matter? I'd rather the function be defined the same as `GenerateFragmentShader`.

src/video_core/renderer_opengl/gl_rasterizer.h
bunnei: Hmm.. I'm not sure what to do about this one, because BitField fails the `is_pod` test (and therefore so does TevStageConfig). Any suggestions?

bunnei:
@yuriks I think I addressed all of your comments. Regarding the explicit layout of uniforms, per our discussions on IRC, this was added because it seems to fix an issue with some AMD drivers optimizing out uniforms (for some unknown reason). I'd like to move to UBOs anyways, but would prefer to do that with my next PR (fragment lighting), as an overhaul to uniform management makes more sense there. So while it's not a great solution, I'd like to keep it as-is in this PR (other than cleaning up @tfarley's patch a bit). FWIW, the extension seems to be pretty much supported everywhere.

src/video_core/renderer_opengl/gl_rasterizer.h
neobrain: It's standard_layout, which is almost good enough for the memcmp.

lioncash: Is there any reason just directly comparing members (either directly, or using `std::tie` to just do one == comparison) isn't possible? At least that way you're guaranteed it'll work and won't need to care about struct padding or any other unnecessary mental overhead.

yuriks: That doesn't work either due to the use of BitField. If you propose
comparing each subfield individually, that will be much slower. I think
we should either fix BitField, or use memcmp anyway under the
"reasonable implementation" assumption. ;P

--yuriks


On Tue, Oct 13, 2015, at 10:58 AM, Lioncash wrote:
> In src/video_core/renderer_opengl/gl_rasterizer.h[1]:


> > +    bool TevStageUpdatesCombinerBufferAlpha(unsigned stage_index)
> >      const {
> +        return (stage_index < 4) && ((combiner_buffer_input >> 4) &
>          (1 << stage_index));
> +    } +
> +    u64 GetHash() const {
> +        return Common::ComputeHash64(this, sizeof(PicaShaderConfig));
> +    } +
> +    bool operator ==(const PicaShaderConfig& o) const {
> +        return std::memcmp(this, &o, sizeof(PicaShaderConfig)) == 0;
> +    }; +
> +    Pica::Regs::CompareFunc alpha_test_func;
> +    std::array<Pica::Regs::TevStageConfig, 6> tev_stages = {};
> +    u8 combiner_buffer_input; +}; Is there any reason just directly
>      comparing members (either directly, or using std::tie to just do
>      one == comparison) isn't possible? At least that way you're
>      guaranteed it'll work and won't need to care about struct padding
>      or any other unnecessary mental overhead.


>



Links:

  1. https://github.com/citra-emu/citra/pull/1187#discussion_r41868178
  2. https://github.com/citra-emu/citra/pull/1187/files#r41868178

src/video_core/renderer_opengl/gl_shader_gen.cpp
Zaneo: @yuriks 

From wildfire Games, If I am reading this correctly:

 53% (s) GL_ARB_explicit_uniform_location

http://feedback.wildfiregames.com/report/opengl/
http://feedback.wildfiregames.com/report/opengl/feature/GL_ARB_explicit_uniform_location

src/video_core/renderer_opengl/gl_shader_gen.cpp
yuriks: It's useless. You're maintaining a bunch of useless operations that can be trivially eliminated, for the sake of some fuzzy "aesthetic" notion. All users of this function are just in turn calling c_str() on it.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,393 @@
...
+        break;
+    case Operation::AddThenMultiply:
+        out += "min(" + variable_name + "[0] + " + variable_name + "[1], vec3(1.0)) * " + variable_name + "[2]";
+        break;
+    default:
+        out += "vec3(0.0)";
+        LOG_CRITICAL(Render_OpenGL, "Unknown color combiner operation: %u", operation);
+        break;
+    }
+    out += ", vec3(0.0), vec3(1.0))";
```
yuriks: Add a comment saying that these are the parameters for the clamp above.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,393 @@
...
+    using CompareFunc = Regs::CompareFunc;
+    switch (func) {
+    case CompareFunc::Never:
+        out += "true";
+        break;
+    case CompareFunc::Always:
+        out += "false";
+        break;
+    case CompareFunc::Equal:
+        out += "int(last_tex_env_out.a * 255.0f) != alphatest_ref";
```
yuriks: You could factor this out so that you don't repeat everything other than the operator in each case.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -45,20 +134,24 @@ class RasterizerOpenGL : public HWRasterizer {
...
-        GLuint color_alpha_multiplier;
-        GLuint const_color;
-        GLuint updates_combiner_buffer_color_alpha;
+    /// OpenGL shader generated for a given Pica register state
+    struct PicaShader {
+        /// OpenGL shader resource
+        OGLShader shader;
+
+        /// Fragment shader uniforms
+        enum Uniform : GLuint {
```
yuriks: `enum class`? It seems like you're already using members scoped anyway.

src/video_core/renderer_opengl/gl_shader_gen.cpp
yuriks: Just for the record, I don't necessarily think this is a huge deal, but this is the kind of thing that I think you do not want to get into the habit of doing.

src/video_core/renderer_opengl/gl_shader_gen.cpp
yuriks: Did you look into this at all?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -78,29 +50,25 @@ void RasterizerOpenGL::InitObjects() {
...
-    glEnableVertexAttribArray(attrib_texcoords);
-    glEnableVertexAttribArray(attrib_texcoords + 1);
-    glEnableVertexAttribArray(attrib_texcoords + 2);
+    glVertexAttribPointer(GLShader::ATTRIBUTE_POSITION, 4, GL_FLOAT, GL_FALSE, sizeof(HardwareVertex), (GLvoid*)offsetof(HardwareVertex, position));
+    glEnableVertexAttribArray(GLShader::ATTRIBUTE_POSITION);
+
+    glVertexAttribPointer(GLShader::ATTRIBUTE_COLOR, 4, GL_FLOAT, GL_FALSE, sizeof(HardwareVertex), (GLvoid*)offsetof(HardwareVertex, color));
+    glEnableVertexAttribArray(GLShader::ATTRIBUTE_COLOR);
+
+    glVertexAttribPointer(GLShader::ATTRIBUTE_TEXCOORDS + 0, 2, GL_FLOAT, GL_FALSE, sizeof(HardwareVertex), (GLvoid*)offsetof(HardwareVertex, tex_coord0));
```
yuriks: I think you should turn this into 3 separate enum members for each texcoord, the current code really doesn't take any advantage of the implicit sequence.

(I though I had commented about this on the last pass, but I couldn't find anything. O_o)

src/video_core/renderer_opengl/gl_rasterizer.cpp
yuriks: Did you really? ;)

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -0,0 +1,388 @@
...
+        out += "false";
+        break;
+    case CompareFunc::Equal:
+    case CompareFunc::NotEqual:
+    case CompareFunc::LessThan:
+    case CompareFunc::LessThanOrEqual:
+    case CompareFunc::GreaterThan:
+    case CompareFunc::GreaterThanOrEqual:
+    {
+        static const std::string op[] = { "!=", "==", ">=", ">", "<=", "<", };
```
yuriks: ~`static const char*`~

VIRGINKLM:
Breaks (trying to spot which one of those commits) Kingdom Hearts Dream Drop Distance in Hardware Renderer mode.

buggiebuggie

On a contrary note, it made the game run in fullspeed! Hahaha!

kemenaran:
This seems to break the hardware renderer on my machine (OS X 10.11, late 2013, Intel Iris 1536 Mo). I get several errors like this:

```
[0m[1;31m[  11.006859] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadProgram:58: Error compiling fragment shader:
ERROR: 0:3: '' :  extension 'GL_ARB_explicit_uniform_location' is not supported

[0m[1;31m[  11.006883] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadProgram:81: Error linking shader:
ERROR: One or more attached shaders not successfully compiled
```

And of course not much is rendered (this is Zelda Ocarina 3DS) :)

capture d ecran 2015-10-25 a 10 57 52

yuriks:
@VIRGINKLM Try updating your drivers.

@kemenaran Since you're on OS X and can't update drivers it looks like there's no real way of fixing this for you yet. I'd encourage posting a separate issue about it. (Lastest OS X drivers not supporting the extension) since that's a pretty big deal.

VIRGINKLM:
OK found out what's going on. Some NVidia models starting with the driver 358.50 fail to save the setting of preferred GPU in NVidia settings panel, just whatever you'll do it will default again (under some circumstances I am yet to discover) to the IGP one. The only remedy I found to that issue is to right click citra (for example) and select it to run using the NVidia GPU. While it's still open, open the NVidia panel and set it to run all applications with the NVidia GPU. Apply the setting and then you can close the application you are running. Silly bug and if I'm not mistaken, it's not the first time this happened, I had that in a different PC like 2-3 years ago. Meh.

SeannyM:
Testing this using the Intel HD 4000 on both Windows with the latest drivers (15.33.39.4276) and OS X, they both reproduce the issue @kemenaran is experiencing. It seems it's not just the OS X drivers that have the problem.

yuriks:
>people in charge of researching extension coverage

VIRGINKLM:
Man if you have never commited that pull request, I would have never found out that those NVidia drivers were broken. I feel sooo dumb. Thanx!

SeannyM:
A quick workaround to the Intel problem is to use the 4.3 shaders instead of the 3.3 shaders. The 4.3 shaders have the extension in core, but the problem with this is it would raise the OpenGL requirement to 4.3, which we wouldn't want.

VIRGINKLM:
@SeannyM Even with the conditional this would render all pre-Haswell iGPUs incompatible. :/

PR #1196, opened 2015-10-09
Add optional GL_KHR_debug support

linkmauve:
This extension allows the driver to communicate errors, warnings and improvement suggestions about our usage of the API, which we then log using our standard mechanism.

In the future we will also likely be using it to insert debug information in our GL stream, to help us understand better what is going on.

I’m not satisfied with the first commit, I regenerated our Glad files from http://glad.dav1d.de/ using the exact same options we were already using plus GL_KHR_debug, and it included a crapload of other imports at the same time…

I haven’t tested on a driver that doesn’t expose the extension, but the guard should be enough anyway.

### DIFF
--- a/src/video_core/renderer_opengl/renderer_opengl.cpp
+++ b/src/video_core/renderer_opengl/renderer_opengl.cpp
@@ -369,6 +369,58 @@ void RendererOpenGL::SetWindow(EmuWindow* window) {
     render_window = window;
 }
 
+static const char* GetSource(GLenum source) {
+#define RET(s) case GL_DEBUG_SOURCE_##s: return #s
+    switch (source) {
+    RET(API);
+    RET(WINDOW_SYSTEM);
+    RET(SHADER_COMPILER);
+    RET(THIRD_PARTY);
+    RET(APPLICATION);
+    RET(OTHER);
+    default:
+        UNREACHABLE();
+    }
+#undef RET
+}
+
+static const char* GetType(GLenum type) {
+#define RET(t) case GL_DEBUG_TYPE_##t: return #t
+    switch (type) {
+    RET(ERROR);
+    RET(DEPRECATED_BEHAVIOR);
+    RET(UNDEFINED_BEHAVIOR);
+    RET(PORTABILITY);
+    RET(PERFORMANCE);
+    RET(OTHER);
+    RET(MARKER);
+    default:
+        UNREACHABLE();
+    }
+#undef RET
+}
+
+static void DebugHandler(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length,
+                         const GLchar* message, const void* user_param) {
+    Log::Level level;
+    switch (severity) {
+    case GL_DEBUG_SEVERITY_HIGH:
+        level = Log::Level::Error;
+        break;
+    case GL_DEBUG_SEVERITY_MEDIUM:
+        level = Log::Level::Warning;
+        break;
+    case GL_DEBUG_SEVERITY_NOTIFICATION:
+        level = Log::Level::Info;
+        break;
+    case GL_DEBUG_SEVERITY_LOW:
+        level = Log::Level::Debug;
+        break;
+    }
+    LOG_GENERIC(Log::Class::Render_OpenGL, level, "%s %s %d: %s",
+                GetSource(source), GetType(type), id, message);
+}
+
 /// Initialize the renderer
 void RendererOpenGL::Init() {
     render_window->MakeCurrent();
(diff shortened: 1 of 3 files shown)


linkmauve:
I just tested nchery’s Mesa branch (http://cgit.freedesktop.org/~nchery/mesa/log/?h=ext5) which allows to disable GL_KHR_debug using the `MESA_EXTENSION_OVERRIDE=-GL_KHR_debug` environment variable, as expected this worked properly.

bunnei:
Looks fine to me. I don't really have much insight on if/how this pollutes our Glad files... perhaps @yuriks should comment.

linkmauve:
I asked dav1d, Glad’s author, and I apparently forgot to select a core profile…

The churn added by the first commit is now fixed. :)

yuriks:
I have a few comments, hold on.

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -369,6 +369,58 @@ void RendererOpenGL::SetWindow(EmuWindow* window) {
     render_window = window;
 }
 
+#ifdef GL_KHR_debug
+static std::string GetSource(GLenum source) {
```
yuriks: Should just return `const char*`

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -369,6 +369,58 @@ void RendererOpenGL::SetWindow(EmuWindow* window) {
...
+    RET(THIRD_PARTY);
+    RET(APPLICATION);
+    RET(OTHER);
+    default:
+        return "";
+    }
+#undef RET
+}
+
+static std::string GetType(GLenum type) {
```
yuriks: Should just return `const char*`

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -369,6 +369,58 @@ void RendererOpenGL::SetWindow(EmuWindow* window) {
...
+#define RET(s) case GL_DEBUG_SOURCE_##s: return #s
+    switch (source) {
+    RET(API);
+    RET(WINDOW_SYSTEM);
+    RET(SHADER_COMPILER);
+    RET(THIRD_PARTY);
+    RET(APPLICATION);
+    RET(OTHER);
+    default:
+        return "";
```
yuriks: "Unknown"?

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -369,6 +369,58 @@ void RendererOpenGL::SetWindow(EmuWindow* window) {
...
+
+static void DebugHandler(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length,
+                         const GLchar* message, const void* user_param) {
+    Log::Level level;
+    switch (severity) {
+    case GL_DEBUG_SEVERITY_HIGH:
+        level = Log::Level::Error;
+        break;
+    case GL_DEBUG_SEVERITY_MEDIUM:
+    case GL_DEBUG_SEVERITY_LOW:
```
yuriks: This will spam a lot of crap on the logs. Drivers often use low severity messages to emit status or informational messages which don't matter to users and rarely matter to developers. I think this should go as debug.

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -379,6 +431,13 @@ void RendererOpenGL::Init() {
         exit(-1);
     }
 
+#ifdef GL_KHR_debug
```
yuriks: Is there any point in the ifdef? I think we should just rely on the symbols existing in our header.

linkmauve:
Addressed all of the comments, but replaced `return "Unknown";` with `UNREACHABLE();`, which gives the intent in a better way.

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -369,6 +369,58 @@ void RendererOpenGL::SetWindow(EmuWindow* window) {
...
+    RET(MARKER);
+    default:
+        UNREACHABLE();
+    }
+#undef RET
+}
+
+static void DebugHandler(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length,
+                         const GLchar* message, const void* user_param) {
+    Log::Level level;
```
bunnei: Should this have a default if the below switch for some reason doesn't set it?

yuriks:
LGTM

Issue #708, opened 2015-04-24
Warn the user their ROM is probably encrypted


linkmauve:
Encrypted ROMs are the number one cause of issues people have with Citra, and it doesn’t display any meaningful error.

The best way to handle that would be to error when the encrypted bit is set, but many decrypters apparently don’t clear that bit, which means a ROM can be both decrypted and have the encrypted bit set.

Thus, a warning that would tell the user their ROM is probably encrypted, while allowing them to still run it if they are sure it isn’t would be the best compromise we can do.

To prevent the user from seeing that warning again and again we could also have some way to unset the bit on successful run.

i30817:
Why should you care about buggy decrypters - why not let the users dump the games again with something better?

yuriks:
As far as I know, the usual heuristic used for this is to verify fields that are present in both the main NCCH header (which is always plain text) with the one in the inner exheader (which can be encrypted). If both match, then the NCCH is not encrypted.

PR #1199, opened 2015-10-11
Loader: Implement NCCH encryption check

Gareth422:
I've implemented a simple encryption check which compares the program IDs in the NCCH header and the extended header to see if the ROM is encrypted. I changed all of the header values to be explicitly little-endian to prevent any surprises if Citra gets ported to a big-endian microwave oven. I also changed the program ID from a uint8 array to a u64_le.

Issue #708

### DIFF
--- a/src/core/loader/loader.h
+++ b/src/core/loader/loader.h
@@ -71,6 +71,7 @@ enum class ResultStatus {
     ErrorNotUsed,
     ErrorAlreadyLoaded,
     ErrorMemoryAllocationFailed,
+    ErrorEncrypted,
 };
 
 static inline u32 MakeMagic(char a, char b, char c, char d) {
--- a/src/core/loader/ncch.cpp
+++ b/src/core/loader/ncch.cpp
@@ -266,6 +266,11 @@ ResultStatus AppLoader_NCCH::Load() {
     LOG_DEBUG(Loader, "Thread priority:             0x%X"  , priority);
     LOG_DEBUG(Loader, "Resource limit category:     %d"    , resource_limit_category);
 
+    if (exheader_header.arm11_system_local_caps.program_id != ncch_header.program_id) {
+        LOG_ERROR(Loader, "ExHeader Program ID mismatch: the ROM is probably encrypted.");
+        return ResultStatus::ErrorEncrypted;
+    }
+
     // Read ExeFS...
 
     exefs_offset = ncch_header.exefs_offset * kBlockSize;
--- a/src/core/loader/ncch.cpp
+++ b/src/core/loader/ncch.cpp
@@ -128,9 +128,8 @@ ResultStatus AppLoader_NCCH::LoadExec() {
     if (ResultStatus::Success == ReadCode(code)) {
         std::string process_name = Common::StringFromFixedZeroTerminatedBuffer(
                 (const char*)exheader_header.codeset_info.name, 8);
-        u64 program_id = *reinterpret_cast<u64_le const*>(&ncch_header.program_id[0]);
 
-        SharedPtr<CodeSet> codeset = CodeSet::Create(process_name, program_id);
+        SharedPtr<CodeSet> codeset = CodeSet::Create(process_name, ncch_header.program_id);
 
         codeset->code.offset = 0;
         codeset->code.addr = exheader_header.codeset_info.text.address;
--- a/src/core/loader/ncch.h
+++ b/src/core/loader/ncch.h
@@ -17,31 +17,31 @@
 
 struct NCCH_Header {
     u8 signature[0x100];
-    u32 magic;
-    u32 content_size;
+    u32_le magic;
+    u32_le content_size;
     u8 partition_id[8];
-    u16 maker_code;
-    u16 version;
+    u16_le maker_code;
+    u16_le version;
     u8 reserved_0[4];
     u64_le program_id;
     u8 reserved_1[0x10];
     u8 logo_region_hash[0x20];
     u8 product_code[0x10];
     u8 extended_header_hash[0x20];
-    u32 extended_header_size;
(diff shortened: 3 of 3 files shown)


yuriks:
Micronit: Write commit messages in the imperative ("implement" instead of "implemented"), also, refer to the module you're modifying: "Loader: Implement NCCH encryption check".

src/core/loader/loader.h
```diff
@@ -14,7 +14,7 @@
 #include "common/file_util.h"
 
 namespace Kernel {
-struct AddressMapping;
+    struct AddressMapping;
```
yuriks: We don't indent namespace contents. Also, spurious unrelated change.

src/core/loader/ncch.cpp
```diff
@@ -128,7 +128,7 @@ ResultStatus AppLoader_NCCH::LoadExec() {
     if (ResultStatus::Success == ReadCode(code)) {
         std::string process_name = Common::StringFromFixedZeroTerminatedBuffer(
                 (const char*)exheader_header.codeset_info.name, 8);
-        u64 program_id = *reinterpret_cast<u64_le const*>(&ncch_header.program_id[0]);
+        u64 program_id = ncch_header.program_id;
```
yuriks: I don't think this variable is necessary anymore, since now you can just read the program_id field directly. Just substitute it's uses below.

src/core/loader/loader.h
lioncash: this isn't related to the change

src/core/loader/ncch.cpp
```diff
@@ -278,6 +283,7 @@ ResultStatus AppLoader_NCCH::Load() {
     if (file.ReadBytes(&exefs_header, sizeof(ExeFs_Header)) != sizeof(ExeFs_Header))
         return ResultStatus::Error;
 
+
```
lioncash: unnecessary newline

src/core/loader/ncch.cpp
```diff
@@ -266,6 +266,11 @@ ResultStatus AppLoader_NCCH::Load() {
     LOG_DEBUG(Loader, "Thread priority:             0x%X"  , priority);
     LOG_DEBUG(Loader, "Resource limit category:     %d"    , resource_limit_category);
 
+    if (exheader_header.arm11_system_local_caps.program_id != ncch_header.program_id) {
+        LOG_ERROR(Loader, "Program ID mismatch: the ROM is probably encrypted.");
```
yuriks: "ExHeader program id mismatch: [...]"

src/core/loader/ncch.cpp
yuriks: Unnecessary extra newline

src/core/loader/ncch.h
```diff
@@ -17,31 +17,31 @@
 
 struct NCCH_Header {
     u8 signature[0x100];
-    u32 magic;
-    u32 content_size;
+    u32_le magic;
```
yuriks: Split these changes into a separate commit.

src/core/loader/ncch.h
```diff
@@ -109,7 +109,7 @@ struct ExHeader_StorageInfo {
 };
 
 struct ExHeader_ARM11_SystemLocalCaps {
-    u8 program_id[8];
+    u64 program_id;
```
yuriks: This should also be u64_le.

Gareth422:
I've reorganised all of my changes into two commits.

mailwl:
i like it

yuriks:
The commit messages need to have the "Loader: " or "NCCH:" prefix to indicate what they're modifying. Other than that, LGTM

Gareth422:
I've changed the commit messages.

Road-Drum:
good idea. helps to reduce "failed to load rom" issues (mainly because encrypted rom)

Gareth422:
> good idea. helps to reduce "failed to load rom" issues (mainly because encrypted rom)

That's the idea.

src/core/loader/ncch.cpp
```diff
@@ -266,6 +265,11 @@ ResultStatus AppLoader_NCCH::Load() {
     LOG_DEBUG(Loader, "Thread priority:             0x%X"  , priority);
     LOG_DEBUG(Loader, "Resource limit category:     %d"    , resource_limit_category);
 
+    if (exheader_header.arm11_system_local_caps.program_id != ncch_header.program_id) {
+        LOG_ERROR(Loader, "ExHeader Program ID mismatch: the ROM is probably encrypted.");
```
jroweboy: Maybe you should include a link to the wiki article on decryption or maybe say something like "Please do not create an issue on Github if you have not decrypted this game. You can decrypt your game by following these instructions: [wiki link]" My wording is poor, but saying something to that effect here might help a bit with the issue spam (I know they probably won't know how to copy paste the url, but its better than not having it IMO)

Gareth422: Having that in the log might not be that useful, as I'm not sure how many people would read it. I think it would make more sense to put that in the GUI pop-up when it gets implemented.

jroweboy: I agree with your sentiment, but I don't see how it could hurt to have it here until a GUI pop up is ready. Its not log spam that developers ever encounter, so, to me at least, it can only have a positive effect in the slim chance a user reads the message and doesn't make a issue about it. (Just my opinion though)

bunnei: I think the wording as-is is fine. I agree with the sentiment, but it really belongs in a pop up dialog and not in the debug log; and because we don't really have a nice cross-platform/cross-frontend way of doing this from core code, it's out of scope for this PR I think.

bunnei:
LGTM

PR #1200, opened 2015-10-11
Qt: Rename configuration folder (Breaks configuration, see description)

Gareth422:
The folder in %appdata% or ~/.local/share with Citra.ini in it is currently called "Citra team". This is inconsistent with the naming scheme generally used in these locations. I have renamed the folder to "Citra" to better reflect its purpose.

Instructions for migration:

You'll have to rename the "Citra team" directory to "Citra". This can be found in %appdata% on Windows, or in ~/.local/share on Linux.

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -480,7 +480,7 @@ void GMainWindow::OnConfigure() {
 
 void GMainWindow::closeEvent(QCloseEvent* event) {
     // Save window layout
-    QSettings settings(QSettings::IniFormat, QSettings::UserScope, "Citra team", "Citra");
+    QSettings settings(QSettings::IniFormat, QSettings::UserScope, "Citra", "Citra");
 
     settings.beginGroup("UILayout");
     settings.setValue("geometry", saveGeometry());
@@ -520,7 +520,7 @@ int main(int argc, char* argv[]) {
 
     // Init settings params
     QSettings::setDefaultFormat(QSettings::IniFormat);
-    QCoreApplication::setOrganizationName("Citra team");
+    QCoreApplication::setOrganizationName("Citra");
     QCoreApplication::setApplicationName("Citra");
 
     QApplication::setAttribute(Qt::AA_X11InitThreads);


neobrain:
Fwiw that particular name should refer to some kind of organization. Not sure if we actually want to use "Citra" for that, or "Citra Team"/"Citra Developers" instead.

linkmauve:
There is another issue related to this, on ¬Windows there is the normal `$XDG_CONFIG_HOME/citra-emu/qt-config.ini` and another one that imo shouldn’t be there, in `$XDG_CONFIG_HOME/Citra Team/Citra.ini`.

It would make sense to merge the second one in the first one.

Gareth422:
> It would make sense to merge the second one in the first one.

Great idea. I'd keep them as separate files, though, just in the same folder. Having more than one configuration file helps keeps things organised, in my opinion.

src/citra_qt/main.cpp
```diff
@@ -480,7 +480,7 @@ void GMainWindow::OnConfigure() {
 
 void GMainWindow::closeEvent(QCloseEvent* event) {
     // Save window layout
-    QSettings settings(QSettings::IniFormat, QSettings::UserScope, "Citra team", "Citra");
+    QSettings settings(QSettings::IniFormat, QSettings::UserScope, "Citra", "Citra");
```
yuriks: iirc there's a version of this constructor that takes just the product name, so just use that instead.

src/citra_qt/main.cpp
```diff
@@ -520,7 +520,7 @@ int main(int argc, char* argv[]) {
 
     // Init settings params
     QSettings::setDefaultFormat(QSettings::IniFormat);
-    QCoreApplication::setOrganizationName("Citra team");
+    QCoreApplication::setOrganizationName("Citra");
```
yuriks: And just drop this line I think.

neobrain:
Btw, since this will end up requiring people to migrate their configuration directory: please provide instructions (i.e. config path before + config path after) for doing so in both the PR description and the commit message.

Gareth422:
On it.

Gareth422:
I've been busy with school. I'll try to see if I can finish this tomorrow.

Gareth422:
How do I write more than one line in a commit message?

JayFoxRox:
Depends on your platform.
Normally you should be able to simply commit without any message (= `git commit`). `git` will start a text editor for you then.

It's been a long time - Is this still active aside from that? Otherwise close this please.

JayFoxRox:
I'll close this for now due to how little it changes and because there isn't any progress towards getting this merged.
Feel free to PR a new version of this.

PR #1204, opened 2015-10-14
citra-qt: Add icon to the OS X app

kemenaran:
Previously the OS X app didn't have any icon. The icon used is the same than for the Windows build.


bunnei:
LGTM

PR #1207, opened 2015-10-18
citra-qt: save hardware-rendering and shaders-jit settings

kemenaran:
Currently the "Use hardware renderer" or "Use Shader JIT" settings aren't persisted across launches when set from the Qt GUI. This PR makes those settings persistent.


Subv:
Are you sure this is the case? I could swear they get saved for me, though i haven't checked in a while

yuriks:
Yes, none of these settings are saved currently.

bunnei:
LGTM

PR #1208, opened 2015-10-18
Implement FS_User::GetFreeBytes

archshift:
Fixes SSB demo showing "no free space" error screen on boot.

### DIFF
--- a/src/core/file_sys/archive_backend.h
+++ b/src/core/file_sys/archive_backend.h
@@ -131,6 +131,12 @@ public:
      * @return Opened directory, or nullptr
      */
     virtual std::unique_ptr<DirectoryBackend> OpenDirectory(const Path& path) const = 0;
+
+    /**
+     * Get the free space
+     * @return The number of free bytes in the archive
+     */
+    virtual u64 GetFreeBytes() const = 0;
 };
 
 class ArchiveFactory : NonCopyable {
--- a/src/core/file_sys/disk_archive.cpp
+++ b/src/core/file_sys/disk_archive.cpp
@@ -74,6 +74,11 @@ std::unique_ptr<DirectoryBackend> DiskArchive::OpenDirectory(const Path& path) c
     return std::move(directory);
 }
 
+u64 DiskArchive::GetFreeBytes() const {
+    // TODO: Stubbed to return 1GiB
+    return 1024 * 1024 * 1024;
+}
+
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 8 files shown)


neobrain:
Just my 2c: I don't like leaking host state into guest state. Doing so has multiple issues:
* Applications may expect certain free-disk-space trends upon executing a number of filesystem commands. For example, a twisted way of ensuring a file has been written properly may be to call GetFreeBytes before and after writing a file, computing the difference of the return values, and comparing that difference to the intended size of the file. Obtaining the value from the host will not faithfully return the expected result here, since host disk usage will influence the result.
* Applications may (implicitly) expect the return value of GetFreeBytes to have an upper bound (i.e. the total NAND size) by ignoring the upper bits of the return value. This will cause issues when the free host disk space is larger than the actual 3DS NAND size. In particular cases, the free disk space may be incorrectly sensed to be zero bytes.
* It's wrong by principle, since the emulator should provide a sandboxed environment. The two issues mentioned above, which are maybe not really immediate, illustrate why that is generally a good rule to follow.

The intent of using the host disk space was probably along the line "let's provide more space than the actual 3DS", but that should be considered carefully after implementing faithful behavior beforehand.

yuriks:
I think that reporting host free space is unnecessary/bad because:
 - It's a more complicated solution which requires us to have platform-specific code paths, and sometimes has I'll defined semantics. Given that it's not strictly necessary it would be best to skip it and avoid this complexity.
 - Of all the issues @neobrain outlined above, I think applications expecting the available space to be "small enough" (that is, reasonable given a save archive or a SD card) is the only one likely to present problems in practice. (e.g. if you have a 4TB hard drive it'll very likely at least cause bad formatting etc. on games.)

I think for a first approximation we should just return a large enough value (a gig/a few hundred megs) for the SD. Saves (which are IVTC archives, though iirc we don't treat them as such) should ideally report something consistent with the size the game used when formatting the save, since in the 3DS they're stored in a self-contained IVTC image with a fixed size.

If some program uses free space to drive some kind of IO logic then we can in the future boot the emulator with some arbitrary initial free-size and adjust that value as appropriate during runtime as IO actions are executed. Calculating the size of the entire directory tree of files on disk each time would be prohibitively slow. This is a big if in the first place though, so it should only be done if this is identified as an actual need.

bunnei:
@archshift can you finish up this PR per the above suggestions? I think the above suggestion (to just make it a fixed value for now) is sufficient.

archshift:
Done.

bunnei:
LGTM

PR #1192, opened 2015-10-08
citra-qt: Change file path encoding

wwylele:
Tried changing the file path encoding method to support non-Latin characters (issue #730)

Issue #730, opened 2015-05-08
Japanese characters not working in file names


JanetMerai:
Just as it says, when I boot up a game in Citra that has Japanese characters in its file name, it won't load at all.

Renaming the game with English makes it work again.

kemenaran:
Accentuated characters in path (like `/usr/john/Émulation/demo.3dsx`) don't work either.

mailwl:
russian too, because of latin encoding

JanetMerai:
I've tested Japanese characters in some of my dumps and it still doesn't work.
When I open Citra, I see boxes instead of proper Japanese characters.

Any idea or estimation when this will be implemented?

yuriks:
I think I'm the only one interested in this, but I haven't had much time to work on Citra in a while, so this has been getting very low priority.

LittleWhite-tb:
This bug does not appear in Linux. I will take some time to set my environment on Windows and see which code does cause this problem.

LittleWhite-tb:
After investigation, I don't think there is much to do.
With just need to compile using UNICODE to handle this case. All the code handling UNICODE and wstring is already implemented and it should work on the go.
*BUT*, by enabling UNICODE, it will not compile perfectly, there is small defect in string_util.cpp/.h . Moreover, in Qt code, there will be no need to convert in Latin/Local8bit anywmore. Just pass the QString as std::string (toStdString()) and the internal code will handle it perfectly.

I will bring the minor changes soon and see how to make CMake enabling UNICODE.

yuriks:
We use std::string to store and process filenames, I don't think just enabling UNICODE will work. The proper way to do this would be to either use QString throughout the codebase (bad because we don't want a hard depedency), transcode qString into std::wstring (bad because it doesn't support all unicode characters anyway and UTF-16 is terrible) or transcode QString into a UTF-8 internally and and then go through a compat layer when opening files. I think the later is preferable, see http://www.utf8everywhere.org/ for more details and rationale for that.

yuriks:
(That said, we should totally be compiling with UNICODE, it's just that iirc we're doing that already, the issues come from C++ standard library io stuff.)

neobrain:
Adding on a detail of what @yuriks said:

> The proper way to do this would be to either use QString throughout the codebase (bad because we don't want a hard depedency),

[...]

PR #1209, opened 2015-10-20
citra-qt: Change file path encoding

wwylele:
change file path encoding method to support non-Latin characters (issue #730)

#1192 rework


yuriks:
Just as a note: The proper way of fixing this, especially on windows, is to use unicode, but this will do for now, since that requires non-trivial refactoring inside the emulator.

bunnei:
LGTM

PR #1225, opened 2015-10-28
csnd_snd: Get rid of type punning

lioncash:


yuriks:
What's the point of this? The memory isn't being aliased by anything.

lioncash:
For one, that violates alignment requirements.

yuriks:
How? The structure will be aligned on the 3DS, so it will be aligned on the host too.

lioncash:
You may run ubsan, and discover that no, no that isn't a guarantee.

yuriks:
That's a bogus warning. Ubsan can't detect that a pointer has alignment due to external factors (in this case, being aligned on the proper 3DS alignment relative to the page, which is allocated with max type alignment by malloc on the host.)

lioncash:
Do you have a source that verifies or confirms that inability?

bunnei:
I don't really have an opinion either way on this. @lioncash and @yuriks please resolve your discussion.

purpasmart96:
Looks a lot better imo

bunnei:
Time's up...
Merging this one. While I agree with @yuriks in that I don't believe this matters at all, it certainly does not break anything and if it fixes some warning by some tool than ¯\_(ツ)_/¯ .. I'm tired of seeing this in our PR queue

Issue #1170, opened 2015-09-23
Buffer overflow


mailwl:
here:
https://github.com/citra-emu/citra/blob/master/src/video_core/command_processor.cpp#L160

attribute_index max value = 11, but in some games it may be 13 (For example, some LEGO games, Tales of the Abyss...). It leads to getting garbage in Get... functions, and, for example, addresses become bigger, than max available memory address

may value 13 mean something unusual?

Subv:
I don't understand, attribute_index is used to index the various vertex_attribute_* arrays, which have 16 elements each, a attribute_index value of 13 will not cause an overflow.

Perhaps you are talking about the "component" loop variable?

mailwl:
sorry Sebastian, a don't remember the real trouble already, something like 12 elements in array, but game ask for 13

PR #1236, opened 2015-11-10
GPU/Loaders: Log an error when a loader tries to load from a component beyond the available ones

Subv:
Related to #1170 
None of my games do this but I figured it'd be a good thing to know if it does happen.


bunnei:
Seems reasonable

PR #1237, opened 2015-11-11
Shaders: Use UBOs instead of individual uniforms in the generated frag shaders

Subv:
Uniform Buffer Objects allow us to upload all our uniforms in a single OpenGL call instead of multiple ones.

This increases performance a little bit.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.h
+++ b/src/video_core/renderer_opengl/gl_rasterizer.h
@@ -141,12 +141,9 @@ public:
 
         /// Fragment shader uniforms
         enum Uniform : GLuint {
-            AlphaTestRef = 0,
-            TevConstColors = 1,
-            Texture0 = 7,
-            Texture1 = 8,
-            Texture2 = 9,
-            TevCombinerBufferColor = 10,
+            Texture0 = 0,
+            Texture1 = 1,
+            Texture2 = 2,
         };
     };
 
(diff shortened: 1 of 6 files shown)


bunnei:
LGTM.
Call for additional reviews or it's getting merged.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -216,6 +213,15 @@ class RasterizerOpenGL : public HWRasterizer {
...
     };
 
+    /// Uniform structure for the Uniform Buffer Object, all members must be 16-byte aligned
+    struct UniformData {
+        // A vec4 color for each of the six tev stages
+        std::array<GLfloat, 4> const_color[6];
+        std::array<GLfloat, 4> tev_combiner_buffer_color;
+        GLint alphatest_ref;
+        INSERT_PADDING_BYTES(12);
+    };
```
yuriks: Add a sizeof static assert to be sure no unwanted gets introduced by accident in the future?

yuriks:
What HIVs did you test this on? Anyway, LGTM

src/video_core/renderer_opengl/gl_rasterizer.h
bunnei: And/or that it is less than 16kb ("minimum maximum" size required by OpenGL for a UBO)

bunnei:
I've tested it on an Nvidia GTX 860M and Intel HD 4000...

bunnei:
@Normmatt tested on AMD.

Issue #1054, opened 2015-08-19
Enable RTTI


neobrain:
RTTI (RunTime Type Information) is a C++ language feature used to identify and cast dynamic objects (i.e. instances of classes with virtual members). Of course, this technically incurs /some/ overhead, since information about the "active" type of an object needs to be stored somewhere in memory. This overhead is the reason why many projects (including Citra) outright disable RTTI.

My proposal is to re-enable RTTI for Citra.

This might be somewhat controversial, but let me clarify that I'm not proposing active usage of RTTI all over the place. If need be, we can actually leave the active usage of RTTI prohibited. The proposal is just about enabling RTTI in the compiler flags.

With that settled, reasons for this proposal are:
* there is literally no overhead from using RTTI for non-virtual classes (since type information will always be known at compile-time), which is what virtually all of Citra's performance/memory-critical data structures should be.
* RTTI overhead for virtual classes seems to be neglegible, at least on GCC as per http://stackoverflow.com/questions/579887/how-expensive-is-rtti/4334421#4334421 . I don't know how things look like for MSVC or clang.
* RTTI might be useful in a few edge cases. GUI code comes to mind, albeit Qt's `moc` provides an alternative anyway (effectively reimplementing RTTI without relying on the language feature).
* Enabling RTTI allows us to safely link against the official boost binaries. There seems to be a workaround for linking them against non-RTTI (http://www.boost.org/doc/libs/1_59_0/doc/html/boost_typeindex/mixing_sources_with_rtti_on_and_.html), but it's not guaranteed to work. Linking boost libraries would be beneficial for including some non-header-only libraries (in particular, test, filesystem, and program_options are of interest for Citra).

However, I'm also genuinely curious if there are indeed any actual drawbacks of RTTI which would affect Citra considerably. Hence feel free to correct me if I'm missing something ;)

tl;dr: I suggest re-enabling RTTI because there are a few minor reasons to do so and there seem to be no actual reasons against it. If people agree, I will make the following changes:
* re-enable RTTI in the build settings
* add a line to the coding style saying that usage of RTTI is strongly discouraged.

hrydgard:
In PPSSPP, RTTI is only disabled because early versions of the Android NDK did not support it, and also Symbian, I believe, which you guys don't support, has broken support if any. The overhead is indeed negligible and I would like to eventually enable it. I don't see any reason why you guys wouldn't.

neobrain:
[...]

PR #1238, opened 2015-11-12
Enable RTTI on Windows builds.

neobrain:
As per https://github.com/citra-emu/citra/issues/1054.

If I'm not mistaken, RTTI was only disabled on Windows. Please correct me if I missed a gcc/clang compiler switch which disables RTTI currently.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,9 +92,8 @@ else()
     # /Zi - Output debugging information
     # /Zo - enahnced debug info for optimized builds
     set(CMAKE_C_FLAGS   "/W3 /MP /Zi /Zo" CACHE STRING "" FORCE)
-    # /GR- - Disable RTTI
     # /EHsc - C++-only exception handling semantics
-    set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} /GR- /EHsc" CACHE STRING "" FORCE)
+    set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} /EHsc" CACHE STRING "" FORCE)
 
     # /MDd - Multi-threaded Debug Runtime DLL
     set(CMAKE_C_FLAGS_DEBUG   "/Od /MDd" CACHE STRING "" FORCE)


yuriks:
- What is the performance and binary size impact of this change?
 - Is there any actual reason why you're enabling them now?

neobrain:
@yuriks Please read #1054 about both of these questions. Also, please mention your concerns earlier in the future.

yuriks:
@neobrain I had already read #1054, and it doesn't answer either of these questions.

neobrain:
@yuriks It does. Please read it more carefully.

yuriks:
@neobrain It does not. "Some" and "negligible" are not numbers, and it lists some "reasons" (which mostly boil down to "doesn't hurt", except for the linking thing) but you still haven't said why you're enabling it right now.

neobrain:
@yuriks I have given this far enough consideration already, especially after having only few people respond to the issue tracker ticket. I'm not willing to spend any time arguing over this, since people already had a chance at asking for details.

That said, all evidence points towards this being a zero-impact change, and it does enable use of RTTI were appropriate and makes Citra compatible with libraries relying on RTTI. That makes "no disadvantages" versus "potential advantages". If you feel the need to have hard numbers, please measure them yourself. I don't see any point in doing so myself until we demand performance measurements for any other PR with equally small chance of real-world performance regression.

yuriks:
What the fuck. Glad to know I'm also allowed to arbitrarily self-merge changes if I feel like it. I was only asking for data and a justification. (cc @bunnei )

hrydgard:
While doing that may have been a little bit brash, I think neobrain is right here. It's already enabled on all your other builds, and it's a standard feature in C++ that was disabled (with no justification!)

It will not affect emulation performance.

neobrain:
@hrydgard Thanks for the nice summary.

@yuriks Just for reference, I have mentioned it multiple times before - if people can't keep up with repository activity, reviving discussions about matters that date back by months (!) after they come up initially is an unacceptable way to catch up. I stand by the opinion that the information given in the issue report is more than sufficient to understand the motivation for this change and to justify it.

yuriks:
Btw, for posteriority, the size effect of this in Release was a 1.5-2% increase.

PR #1246, opened 2015-11-21
Fix read and write register blocks in gdbstub

polaris-:
Previously, the padding wasn't correctly accounted for which caused the gdbstub to read and write everything after R15 (starting with the dummy FPA registers) incorrectly, which caused CPSR to not be handled correctly. Everything appears to be working as expected with this change.

### DIFF
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -57,9 +57,8 @@ const u32 MSG_WAITALL = 8;
 
 const u32 R0_REGISTER = 0;
 const u32 R15_REGISTER = 15;
-const u32 CSPR_REGISTER = 25;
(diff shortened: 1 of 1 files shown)


yuriks:
LGTM, I'll let @Normmatt verify it works before merging.

(Btw, in the future try to keep the rename in a separate commit from the other logic changes. It makes it easier to check exactly what changed without clutter in the diff.)

lioncash:
Seems like there's tab/space mismatches

Normmatt:
This seems to be correct now :+1:

yuriks:
@polaris- Can you sqush? Then I'll merge.

bunnei:
LGTM

polaris-:
@yuriks Done.

Subv:
:shipit:

PR #1247, opened 2015-11-22
Add Initialize and GenerateRandomData function names to ssl:C

polaris-:
Added function names for Initialize and GenerateRandomData in ssl:C (names came from 3dsbrew). Found these functions while testing the game SPEC ~KAN~.


bunnei:
LGTM

PR #1248, opened 2015-11-22
Add stub functions for Initialize and GenerateRandomData in ssl:C

polaris-:
After doing some more research and determining exactly what was causing SPEC ~KAN~ to freeze during save data initialization, The game calls GenerateRandomData three times while it's initializing the save data on the first run, and if the data returned on the third call is the same as the data returned on the first call are not different then it will continue looping until it gets something different. The game would assume that everything was initialized properly if you closed the game and ran it again. This fixes the issue of it looping on the first run of the game.

I thought it was logical that the RNG gets seeded when the SSL service is initialized, but if there's a better way to do this then I'd be up for changing it.

### DIFF
--- a/src/core/hle/service/ssl_c.cpp
+++ b/src/core/hle/service/ssl_c.cpp
@@ -2,6 +2,8 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <random>
+
 #include "core/hle/hle.h"
 #include "core/hle/service/ssl_c.h"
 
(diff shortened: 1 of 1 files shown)


yuriks:
This is fine as a workaround for the problem, but add a TODO saying to make use of a proper CSPRNG whenever we implement actual SSL. Just in case someone is foolish enough to actually try to send sensitive data through Citra.

polaris-:
@yuriks I agree. I probably should've made a mention of that somewhere, but I wasn't intending for any of this to be used in any context where security was needed. I've added a comment mentioning it.

Subv:
LGTM

bunnei:
LGTM

Issue #1222, opened 2015-10-26
Shaders compilation fails on Intel drivers (missing 'GL_ARB_explicit_uniform_location' extension)


kemenaran:
On systems with Intel integrated graphics, shaders compilation fails with several of these errors:

```
Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadProgram:58:
Error compiling fragment shader:
extension 'GL_ARB_explicit_uniform_location' is not supported

Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadProgram:81:
Error linking shader:
ERROR: One or more attached shaders not successfully compiled
```

This error appeared after the merge of the [GSLS Fragment Shaders Generation PR](https://github.com/citra-emu/citra/pull/1187). Reproduced by @SeannyM on Intel HD 4000 on both Windows with the latest drivers (15.33.39.4276) and OS X.

Results in almost nothing being rendered:

<img width="512" alt="8c78abd6-7b07-11e5-94a4-d426b9b893bf" src="https://cloud.githubusercontent.com/assets/179923/10723223/f34f7838-7bb8-11e5-9a29-5a7b851c1113.png">

/cc @yuriks

Subv:
It works correctly on Intel HD 4600

Subv:
Fixed by #1253

PR #1253, opened 2015-11-25
Use regular uniform location

kemenaran:
The support for GL_ARB_explicit_uniform_location is decent, but not that good (53% according to http://feedback.wildfiregames.com/report/opengl/feature/GL_ARB_explicit_uniform_location).

Luckily, only a single place uses the explicit uniform location now: the textures uniform. So we can use regular location without much changes.

This fix the shader compilation on Intel HD 4000 (#1222 ; cc @bunnei).

_Disclaimer: I have no idea what I'm doing. But I'm eager for pointers and feedback :)_

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -492,9 +492,10 @@ void RasterizerOpenGL::SetShader() {
         state.Apply();
 
         // Set the texture samplers to correspond to different texture units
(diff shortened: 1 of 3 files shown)


Subv:
LGTM

bunnei:
Thanks, I was planning on taking this out now that we have the UBO, but hadn't gotten around to it yet. Looks fine, I'll test it later today just to be sure.

bunnei:
LGTM

PR #1255, opened 2015-11-26
[Common] Return error from ScanDirectoryTreeAndCallback for missing dir

yuriks:
This fixes a regression introduced in #1095 that prevented Fire Emblem:
Awakening from getting past its "Extdata needs to be recreated" screen.
However, this is very likely just a bad workaround, and other things in
the FS emulation need to be fixed for this to work as intended. (Such as
DeleteExtSaveData returning an error instead of "nothing happened" if
the data does not exist as in this case. The extdata doesn't even get
created, which is what causes this prompt to appear in the first place.)


archshift:
I'll try to figure out a way to clean up ScanDirectoryTreeAndCallback in the meantime so that we don't need these kinds of workarounds.

yuriks:
To clarify what I mean: I don't think that this PR is a workaround. DeleteDirRecursively not returning an error when the directory doesn't exist is a bug.

The "workaround" I'm referring to is that DeleteExtSaveData shouldn't have to fail in order for booting the game to work.

yuriks:
Superceded by #1256

PR #1257, opened 2015-11-27
Kernel: Implement svcGetSystemInfo

yuriks:
This makes smealum/ctrulib@b96dd51d3349961189d4ab1bc2a5c45deff21c09 work with Citra.

cc @fincs @Steveice10

### DIFF
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -778,6 +778,51 @@ static ResultCode CreateMemoryBlock(Handle* out_handle, u32 addr, u32 size, u32 
     return RESULT_SUCCESS;
 }
 
+static ResultCode GetSystemInfo(s64* out, u32 type, s32 param) {
+    using Kernel::MemoryRegion;
+
+    LOG_TRACE(Kernel_SVC, "called process=0x%08X type=%u param=%d", process_handle, type, param);
+
+    switch ((SystemInfoType)type) {
+    case SystemInfoType::REGION_MEMORY_USAGE:
+        switch ((SystemInfoMemUsageRegion)param) {
+        case SystemInfoMemUsageRegion::ALL:
+            *out = Kernel::GetMemoryRegion(Kernel::MemoryRegion::APPLICATION)->used
+                 + Kernel::GetMemoryRegion(Kernel::MemoryRegion::SYSTEM)->used
+                 + Kernel::GetMemoryRegion(Kernel::MemoryRegion::BASE)->used;
+            break;
+        case SystemInfoMemUsageRegion::APPLICATION:
+            *out = Kernel::GetMemoryRegion(Kernel::MemoryRegion::APPLICATION)->used;
+            break;
+        case SystemInfoMemUsageRegion::SYSTEM:
+            *out = Kernel::GetMemoryRegion(Kernel::MemoryRegion::SYSTEM)->used;
+            break;
+        case SystemInfoMemUsageRegion::BASE:
+            *out = Kernel::GetMemoryRegion(Kernel::MemoryRegion::BASE)->used;
+            break;
+        default:
+            LOG_ERROR(Kernel_SVC, "unknown GetSystemInfo type=0 region: param=%d", param);
+            *out = 0;
+            break;
+        }
+        break;
+    case SystemInfoType::KERNEL_ALLOCATED_PAGES:
+        LOG_ERROR(Kernel_SVC, "unimplemented GetSystemInfo type=2 param=%d", type, param);
+        *out = 0;
+        break;
+    case SystemInfoType::KERNEL_SPAWNED_PIDS:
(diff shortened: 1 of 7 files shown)


src/core/hle/svc.cpp
```diff
@@ -778,6 +778,56 @@ static ResultCode CreateMemoryBlock(Handle* out_handle, u32 addr, u32 size, u32
...
+            *out = Kernel::GetMemoryRegion(Kernel::MemoryRegion::APPLICATION)->used;
+            break;
+        case 2:
+            *out = Kernel::GetMemoryRegion(Kernel::MemoryRegion::SYSTEM)->used;
+            break;
+        case 3:
+            *out = Kernel::GetMemoryRegion(Kernel::MemoryRegion::BASE)->used;
+            break;
+        default:
+            LOG_ERROR(Kernel_SVC, "unknown GetSystemInfo type=2 region: param=%d", param);
```
archshift: Shouldn't this be `type=0`?

yuriks: Well spotted. Copy&paste fail.

src/core/hle/svc.cpp
```diff
@@ -778,6 +778,56 @@ static ResultCode CreateMemoryBlock(Handle* out_handle, u32 addr, u32 size, u32
...
 
+static ResultCode GetSystemInfo(s64* out, u32 type, s32 param) {
+    using Kernel::MemoryRegion;
+
+    LOG_TRACE(Kernel_SVC, "called process=0x%08X type=%u param=%d", process_handle, type, param);
+
+    switch (type) {
+    case 0:
+        // Reports total used memory for all regions, or for a specific process region.
+        switch (param) {
```
archshift: Can we make an enum out of this?

src/core/hle/svc.cpp
```diff
@@ -778,6 +778,56 @@ static ResultCode CreateMemoryBlock(Handle* out_handle, u32 addr, u32 size, u32
     return RESULT_SUCCESS;
 }
 
+static ResultCode GetSystemInfo(s64* out, u32 type, s32 param) {
+    using Kernel::MemoryRegion;
+
+    LOG_TRACE(Kernel_SVC, "called process=0x%08X type=%u param=%d", process_handle, type, param);
+
+    switch (type) {
```
archshift: Can we make an enum out of this, or is there not enough information RE'd?

yuriks: The cases handled here are the only ones my kernel supports. Since they're not contiguous, are only used here, exactly once, and there's no accepted naming scheme for them, I think we should skip enums here.

src/core/hle/svc.cpp
yuriks: Same as previous response. We could reuse the Kernel::MemoryRegion enum, but that would be conflating the internal implementation with the SVC interface. These values are only used once here, so an enum doesn't buy us much.

Subv:
lgtm

src/core/hle/svc.cpp
bunnei: Ditto with @archshift... I'm not sure why 'only being used once' is an argument against using an enum. It's not obvious what each of these case statements means, at the very least add some comments.

bunnei:
Looks good, other than adding the enum and/or comments (I see no need for comments when an enum self-documents the code a bit better, but I'm fine with either)

yuriks:
I already have comments for each of the major query types. I really don't what extra useful information I can add to the subtypes that won't be a straight duplication. Each case has a very visible MemoryRegion:APPLICATION, etc. enum in the sole line of code that comprises the case. The comments would just be "Gets the used memory for the APPLICATION region.", "Gets the used memory for the SYSTEM region.", "Gets the used memory for the BASE region.", and the comment for the type already says that: "Reports total used memory for all regions, or for a specific process region.". (Likewise, an enum would also just be APPLICATION, SYSTEM, BASE, except now with more confusion since we have two identical similar enums with the same items, in the same context and used together.)

yuriks:
Updated.

src/core/hle/svc.h
```diff
@@ -41,6 +41,35 @@ enum ArbitrationType {
...
+    KERNEL_ALLOCATED_PAGES = 2,
+    /**
+    * "This returns the total number of processes which were launched directly by the kernel.
+    * For the ARM11 NATIVE_FIRM kernel, this is 5, for processes sm, fs, pm, loader, and pxi."
+    */
+    KERNEL_SPAWNED_PIDS = 26,
+};
+
+/**
+* Accepted by svcGetSystemInfo param with REGION_MEMORY_USAGE type. Selects a region to query
```
bunnei: Autoformat on the docstrings in this file?

bunnei:
Other than the above comment, LGTM.

PR #1259, opened 2015-11-29
armstate: Zero out the registers on creation

lioncash:
std::array isn't always guaranteed to explicitly zero out it's contents without an initializer list.


Subv:
LGTM

PR #1260, opened 2015-11-30
Fix compilation warnings (part 1)

Dettorer:
I'm compiling with `-Werror`, `-Weverything` and some warnings disabled (see https://github.com/Dettorer/citra/commit/f47d8b846873d760de71d06d8e7eed98cc18cf1b, not in this pull request).

I find `static_cast` are more clear and less dangerous than C-style even on numeric types (they can hide a `const_cast`) but I would understand if you want me to keep them. I would then add `-Wno-old-style-cast` to my compile flags.

I expect this PR to be far from perfect as this is my first one, I will try to fix it and take the remarks into account for the next ones the best as I can!

I took a different, more progressive approach with PR https://github.com/citra-emu/citra/pull/1284. The fixes in this PR are still useful but I will continue developing with the new direction for future PRs.

### DIFF
--- a/src/common/emu_window.cpp
+++ b/src/common/emu_window.cpp
@@ -125,7 +125,6 @@ EmuWindow::FramebufferLayout EmuWindow::FramebufferLayout::DefaultScreenLayout(u
     return res;
 }
 
-void EmuWindow::OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>& minimal_size) {
+void EmuWindow::OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>&) {
     // By default, ignore this request and do nothing.
-    (void)minimal_size;
 }
--- a/src/common/emu_window.cpp
+++ b/src/common/emu_window.cpp
@@ -124,3 +124,8 @@ EmuWindow::FramebufferLayout EmuWindow::FramebufferLayout::DefaultScreenLayout(u
 
     return res;
 }
+
+void EmuWindow::OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>& minimal_size) {
+    // By default, ignore this request and do nothing.
+    (void)minimal_size;
+}
--- a/src/common/emu_window.h
+++ b/src/common/emu_window.h
@@ -195,7 +195,7 @@ private:
      * Handler called when the minimal client area was requested to be changed via SetConfig.
      * For the request to be honored, EmuWindow implementations will usually reimplement this function.
      */
-    virtual void OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>& minimal_size);
+    virtual void OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>&);
 
     FramebufferLayout framebuffer_layout; ///< Current framebuffer layout
 
--- a/src/common/emu_window.h
+++ b/src/common/emu_window.h
@@ -195,9 +195,7 @@ private:
      * Handler called when the minimal client area was requested to be changed via SetConfig.
      * For the request to be honored, EmuWindow implementations will usually reimplement this function.
      */
-    virtual void OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>& minimal_size) {
-        // By default, ignore this request and do nothing.
-    }
+    virtual void OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>& minimal_size);
 
     FramebufferLayout framebuffer_layout; ///< Current framebuffer layout
 
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -506,7 +506,7 @@ unsigned ScanDirectoryTree(const std::string &directory, FSTEntry& parent_entry)
 
 bool DeleteDirRecursively(const std::string &directory)
 {
-    const static auto callback = [](unsigned* num_entries_out,
+    const static auto callback = [](unsigned*,
                                     const std::string& _directory,
                                     const std::string& virtual_name) -> bool {
         std::string new_path = _directory + DIR_SEP_CHR + virtual_name;
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -470,6 +470,8 @@ bool ForeachDirectoryEntry(unsigned* num_entries_out, const std::string &directo
     // num_entries_out is allowed to be specified nullptr, in which case we shouldn't try to set it
     if (num_entries_out != nullptr)
         *num_entries_out = found_entries;
+
+    return true;
 }
 
 unsigned ScanDirectoryTree(const std::string &directory, FSTEntry& parent_entry)
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -487,7 +487,7 @@ unsigned ScanDirectoryTree(const std::string &directory, FSTEntry& parent_entry)
             entry.isDirectory = true;
             // is a directory, lets go inside
             entry.size = ScanDirectoryTree(entry.physicalName, entry);
-            *num_entries_out += (int)entry.size;
+            *num_entries_out += entry.size;
         } else { // is a file
             entry.isDirectory = false;
             entry.size = GetSize(entry.physicalName);
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -475,11 +475,11 @@ bool ForeachDirectoryEntry(unsigned* num_entries_out, const std::string &directo
 unsigned ScanDirectoryTree(const std::string &directory, FSTEntry& parent_entry)
 {
     const auto callback = [&parent_entry](unsigned* num_entries_out,
-                                          const std::string& directory,
+                                          const std::string& _directory,
                                           const std::string& virtual_name) -> bool {
         FSTEntry entry;
         entry.virtualName = virtual_name;
-        entry.physicalName = directory + DIR_SEP + virtual_name;
+        entry.physicalName = _directory + DIR_SEP + virtual_name;
 
         if (IsDirectory(entry.physicalName)) {
             entry.isDirectory = true;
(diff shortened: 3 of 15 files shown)


src/common/string_util.cpp
```diff
@@ -182,7 +182,7 @@ bool TryParse(const std::string &str, u32 *const output)
     if (errno == ERANGE)
         return false;
 
-#if ULONG_MAX > UINT_MAX
```
wwylele: `ULONG_MAX` and `UINT_MAX` are defined in stantard header `<climits>`

src/common/file_util.h
```diff
@@ -248,7 +248,7 @@ class IOFile : public NonCopyable
     void SetHandle(std::FILE* file);
 
     bool Seek(s64 off, int origin);
-    u64 Tell();
+    int64_t Tell();
```
archshift: We have a `s64` type.

src/common/emu_window.cpp
```diff
@@ -124,3 +124,7 @@ EmuWindow::FramebufferLayout EmuWindow::FramebufferLayout::DefaultScreenLayout(u
 
     return res;
 }
+
+void EmuWindow::OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>&) {
+        // By default, ignore this request and do nothing.
```
Subv: Extra indenting

src/common/hash.cpp
```diff
@@ -36,7 +36,7 @@ static FORCE_INLINE u64 fmix64(u64 k) {
 // platforms (MurmurHash3_x64_128). It was taken from:
 // https://code.google.com/p/smhasher/source/browse/trunk/MurmurHash3.cpp
 void MurmurHash3_128(const void* key, int len, u32 seed, void* out) {
-    const u8 * data = (const u8*)key;
+    const u8 * data = static_cast<const u8 *>(key);
```
Subv: no whitespace after the type

src/common/hash.cpp
```diff
@@ -47,7 +47,7 @@ void MurmurHash3_128(const void* key, int len, u32 seed, void* out) {
 
     // Body
 
-    const u64 * blocks = (const u64 *)(data);
+    const u64 * blocks = reinterpret_cast<const u64 *>(data);
```
Subv: same

src/common/logging/backend.cpp
```diff
@@ -109,7 +109,7 @@ Entry CreateEntry(Class log_class, Level log_level,
     vsnprintf(formatting_buffer.data(), formatting_buffer.size(), format, args);
     entry.message = std::string(formatting_buffer.data());
 
-    return std::move(entry);
+    return entry;
```
Subv: why?

src/common/string_util.cpp
```diff
@@ -390,7 +391,11 @@ static std::string CodeToUTF8(const char* fromcode, const std::basic_string<T>&
 
     while (0 != src_bytes)
     {
-        size_t const iconv_result = iconv(conv_desc, (char**)(&src_buffer), &src_bytes,
+        // This const_cast is here because iconv declares its second argument as
+        // "char **" where it should be "const char **" because the only thing
+        // actually modified is the pointer, hence the const_cast.
+        char **iconv_inbuff = const_cast<char **>(reinterpret_cast<const char **>(&src_buffer));
```
Subv: ** next to the type, not the variable name, no whitespace after the type in the cast

src/common/hash.cpp
Dettorer: Old habit from a different coding style, I'll fix it.

src/common/logging/backend.cpp
Dettorer: "Entry" has a move constructor, standard-compliant compilers *must* use it in this case (http://en.cppreference.com/w/cpp/language/move_constructor), even without std::move.
Plus, adding std::move here prevents some compilers from doing NRVO which is even faster.

yuriks: Oh, that's good to know.

src/common/emu_window.h
```diff
@@ -195,9 +195,7 @@ class EmuWindow
      * Handler called when the minimal client area was requested to be changed via SetConfig.
      * For the request to be honored, EmuWindow implementations will usually reimplement this function.
      */
-    virtual void OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>& minimal_size) {
-        // By default, ignore this request and do nothing.
-    }
+    virtual void OnMinimalClientAreaChangeRequest(const std::pair<unsigned,unsigned>&);
```
yuriks: You removed the parameter name here.

src/common/file_util.cpp
```diff
@@ -365,7 +365,7 @@ u64 GetSize(const std::string &filename)
     {
         LOG_TRACE(Common_Filesystem, "%s: %lld",
                 filename.c_str(), (long long)buf.st_size);
-        return buf.st_size;
+        return static_cast<u64>(buf.st_size);
```
yuriks: You can just use `(u64)` if you want for integer casts, as per the style guide. Not a required change though.

src/common/file_util.cpp
```diff
@@ -470,22 +470,24 @@ bool ForeachDirectoryEntry(unsigned* num_entries_out, const std::string &directo
...
         *num_entries_out = found_entries;
+
+    return true;
 }
 
 unsigned ScanDirectoryTree(const std::string &directory, FSTEntry& parent_entry)
 {
     const auto callback = [&parent_entry](unsigned* num_entries_out,
-                                          const std::string& directory,
+                                          const std::string& _directory,
```
yuriks: Use `directory_` instead.

src/common/file_util.cpp
```diff
@@ -504,10 +506,10 @@ unsigned ScanDirectoryTree(const std::string &directory, FSTEntry& parent_entry)
 
 bool DeleteDirRecursively(const std::string &directory)
 {
-    const static auto callback = [](unsigned* num_entries_out,
-                                    const std::string& directory,
+    const static auto callback = [](unsigned*,
+                                    const std::string& _directory,
```
yuriks: Same, `directory_`.

src/common/file_util.cpp
```diff
@@ -504,10 +506,10 @@ unsigned ScanDirectoryTree(const std::string &directory, FSTEntry& parent_entry)
 
 bool DeleteDirRecursively(const std::string &directory)
 {
-    const static auto callback = [](unsigned* num_entries_out,
-                                    const std::string& directory,
+    const static auto callback = [](unsigned*,
```
yuriks: Can you do something like `unsigned* /* num_entries_out */` instead?

src/common/hash.cpp
```diff
@@ -36,7 +36,7 @@ static FORCE_INLINE u64 fmix64(u64 k) {
 // platforms (MurmurHash3_x64_128). It was taken from:
 // https://code.google.com/p/smhasher/source/browse/trunk/MurmurHash3.cpp
 void MurmurHash3_128(const void* key, int len, u32 seed, void* out) {
-    const u8 * data = (const u8*)key;
+    const u8* data = static_cast<const u8*>(key);
```
yuriks: Are these actually fixing any compilation warnings? If not then I think we should keep the implementation intact.

src/common/memory_util.cpp
```diff
@@ -70,7 +70,7 @@ void* AllocateExecutableMemory(size_t size, bool low)
     }
 #endif
 
-#if EMU_ARCH_BITS == 64
+#if defined(EMU_ARCH_BITS) && EMU_ARCH_BITS == 64
```
yuriks: Please don't. This should always be defined, if it's not then the appropriate header should be included instead.

src/common/misc.cpp
```diff
@@ -12,7 +13,7 @@
 #endif
 
 // Neither Android nor OS X support TLS
-#if  defined(__APPLE__) || (ANDROID && __clang__)
+#if  defined(__APPLE__) || (defined(ANDROID) && ANDROID && __clang__)
```
yuriks: I think you're supposed to just use `defined` for everything here.

src/common/profiler.cpp
```diff
@@ -59,7 +59,7 @@ unsigned int ProfilingManager::RegisterTimingCategory(TimingCategory* category,
     info.name = name;
     info.parent = TimingCategoryInfo::NO_PARENT;
 
-    unsigned int id = (unsigned int)timing_categories.size();
+    unsigned int id = static_cast<unsigned int>(timing_categories.size());
```
yuriks: Meh.

src/common/profiler.cpp
```diff
@@ -90,10 +90,10 @@ void ProfilingManager::FinishFrame() {
     last_frame_end = now;
 }
 
-TimingResultsAggregator::TimingResultsAggregator(size_t window_size)
-        : max_window_size(window_size), window_size(0) {
-    interframe_times.resize(window_size, Duration::zero());
-    frame_times.resize(window_size, Duration::zero());
+TimingResultsAggregator::TimingResultsAggregator(size_t _window_size)
```
yuriks: Again, trailing underscore instead of leading.

src/common/profiler.cpp
```diff
@@ -140,7 +140,7 @@ static AggregatedDuration AggregateField(const std::vector<Duration>& v, size_t
         result.max = std::max(result.max, value);
     }
     if (len != 0)
-        result.avg /= len;
+        result.avg /= static_cast<const long>(len);
```
yuriks: Meh. (Especially "*const* long"???)

src/common/profiler.cpp
```diff
@@ -177,7 +177,7 @@ ProfilingManager& GetProfilingManager() {
 }
 
 SynchronizedRef<TimingResultsAggregator> GetTimingResultsAggregator() {
-    static SynchronizedWrapper<TimingResultsAggregator> aggregator(30);
+    static SynchronizedWrapper<TimingResultsAggregator> aggregator(static_cast<size_t>(30));
```
yuriks: This makes no sense.

yuriks:
I took a pass (didn't read to the end) but I think these warning settings are way too verbose. Some of the fixes there don't seem to make any sense to me, either (like casting literals).

Dettorer:
I agree, I wanted to add `-Werror` to the build system and was told that it would be good to use `-Weverything` too, so I started to fix all the warnings I got with `-Weverything`.
I started a different approach with #1284 (just adding `-Werror` and fixing the big warnings). I wanted to close this PR when #1284 gets merged, but it's still under review and I forgot about this one.

PR #1261, opened 2015-12-01
PICA: Properly emulate 1-stage delay in the combiner buffer

### LINKED ISSUE #1140: TEV buffer implementation is incorrect
@fincs pointed out that the Citra implementation of TEV buffers is incorrect. (Each stage should access the previous buffer instead of the current.)

I plan on at some point writing hwtests and fixing this.

yuriks:
This was discovered and verified by @fincs. The tev combiner buffer
actually lags behind by one stage, meaning stage 1 reads the initial
color, stage 2 reads stage 0's output, and so on.

Fixes character portraits in Fire Emblem: Awakening and world textures
in Zelda: ALBW.

Fixes #1140

### DIFF
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -498,7 +498,8 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             // with some basic arithmetic. Alpha combiners can be configured separately but work
             // analogously.
             Math::Vec4<u8> combiner_output;
-            Math::Vec4<u8> combiner_buffer = {
+            Math::Vec4<u8> combiner_buffer = {0, 0, 0, 0};
+            Math::Vec4<u8> next_combiner_buffer = {
                 regs.tev_combiner_buffer_color.r, regs.tev_combiner_buffer_color.g,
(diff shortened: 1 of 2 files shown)


src/video_core/rasterizer.cpp
```diff
@@ -498,7 +498,8 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             // with some basic arithmetic. Alpha combiners can be configured separately but work
             // analogously.
             Math::Vec4<u8> combiner_output;
-            Math::Vec4<u8> combiner_buffer = {
+            Math::Vec4<u8> combiner_buffer = {0, 0, 0, 0};
```
bunnei: Can't this just be `{}`?

bunnei:
LGTM

src/video_core/rasterizer.cpp
yuriks: No: https://github.com/citra-emu/citra/blob/master/src/common/vector_math.h#L358

bunnei: Ah right.. I recall now we had some previous bugs in the SW rasterizer due to the above assumption.

bunnei:
Tested, works well!

PR #1263, opened 2015-12-02
Pica/Scissor: Implemented the scissor test.

Subv:
This is based on an earlier work by @neobrain.

It is worth mentioning that the way ctrulib handles scissoring is incorrect, the last two registers are not the size of the scissor rectangle but instead are x+width-1 and y+height-1. This has been verified with homebrew. I will be making a PR to ctrulib later on.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -112,7 +112,36 @@ struct Regs {
         BitField<24, 5, Semantic> map_w;
     } vs_output_attributes[7];
 
-    INSERT_PADDING_WORDS(0x11);
+    INSERT_PADDING_WORDS(0xe);
+
+    enum class ScissorMode : u32 {
+        Disabled = 0,
+        Exclude = 1, // Exclude pixels inside the scissor box
+
+        Include = 3 // Exclude pixels outside the scissor box
+    };
+
+    struct {
+        BitField<0, 2, ScissorMode> mode;
+
+        union {
+            BitField< 0, 16, u32> right;
+            BitField<16, 16, u32> bottom;
+        };
+
+        union {
+            BitField< 0, 16, u32> left;
+            BitField<16, 16, u32> top;
+        };
+
+        u32 GetWidth() const {
+            return left - right + 1;
+        }
+
+        u32 GetHeight() const {
+            return top - bottom + 1;
+        }
+    } scissor_test;
 
     union {
         BitField< 0, 16, u32> x;
@@ -986,6 +1015,7 @@ ASSERT_REG_POSITION(viewport_depth_range, 0x4d);
 ASSERT_REG_POSITION(viewport_depth_far_plane, 0x4e);
 ASSERT_REG_POSITION(vs_output_attributes[0], 0x50);
 ASSERT_REG_POSITION(vs_output_attributes[1], 0x51);
+ASSERT_REG_POSITION(scissor_test, 0x65);
 ASSERT_REG_POSITION(viewport_corner, 0x68);
 ASSERT_REG_POSITION(texture0_enable, 0x80);
 ASSERT_REG_POSITION(texture0, 0x81);
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -130,16 +130,16 @@ struct Regs {
         };
 
         union {
-            BitField< 0, 16, u32> left;
-            BitField<16, 16, u32> top;
+            BitField< 0, 16, u32> left_minus_1;
+            BitField<16, 16, u32> top_minus_1;
         };
 
-        u32 GetWidth() const {
-            return left - right + 1;
+        u32 GetTop() const {
+            return top_minus_1 + 1;
         }
 
-        u32 GetHeight() const {
-            return top - bottom + 1;
+        u32 GetLeft() const {
+            return left_minus_1 + 1;
         }
     } scissor_test;
 
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -334,11 +334,24 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             return;
     }
 
-    // TODO: Proper scissor rect test!
     u16 min_x = std::min({vtxpos[0].x, vtxpos[1].x, vtxpos[2].x});
     u16 min_y = std::min({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
     u16 max_x = std::max({vtxpos[0].x, vtxpos[1].x, vtxpos[2].x});
     u16 max_y = std::max({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
+
+    // Convert the scissor box coordinates to 12.4 fixed point
+    u16 scissor_width = (u16)(regs.scissor_test.GetWidth() << 4);
+    u16 scissor_height = (u16)(regs.scissor_test.GetHeight() << 4);
+    u16 scissor_x = (u16)(regs.scissor_test.right << 4);
+    u16 scissor_y = (u16)(regs.scissor_test.bottom << 4);
+
+    if (regs.scissor_test.mode == Regs::ScissorMode::Include) {
+        // Calculate the new bounds
+        min_x = std::max(min_x, scissor_x);
+        min_y = std::max(min_y, scissor_y);
+        max_x = std::min(max_x, (u16)(scissor_x + scissor_width));
+        max_y = std::min(max_y, (u16)(scissor_y + scissor_height));
+    }
 
     min_x &= Fix12P4::IntMask();
     min_y &= Fix12P4::IntMask();
@@ -378,6 +391,13 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
     // TODO: Not sure if looping through x first might be faster
     for (u16 y = min_y + 8; y < max_y; y += 0x10) {
         for (u16 x = min_x + 8; x < max_x; x += 0x10) {
+
+            // Do not process the pixel if it's inside the scissor box and the scissor mode is set to Exclude
+            if (regs.scissor_test.mode == Regs::ScissorMode::Exclude) {
+                if (x >= scissor_x && x <= scissor_x + scissor_width &&
+                    y >= scissor_y && y <= scissor_y + scissor_height)
+                    continue;
+            }
 
             // Calculate the barycentric coordinates w0, w1 and w2
             int w0 = bias0 + SignedArea(vtxpos[1].xy(), vtxpos[2].xy(), {x, y});
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -340,17 +340,17 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
     u16 max_y = std::max({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
 
     // Convert the scissor box coordinates to 12.4 fixed point
-    u16 scissor_width = (u16)(regs.scissor_test.GetWidth() << 4);
-    u16 scissor_height = (u16)(regs.scissor_test.GetHeight() << 4);
-    u16 scissor_x = (u16)(regs.scissor_test.right << 4);
-    u16 scissor_y = (u16)(regs.scissor_test.bottom << 4);
+    u16 scissor_left = (u16)(regs.scissor_test.GetLeft() << 4);
+    u16 scissor_top = (u16)(regs.scissor_test.GetTop() << 4);
+    u16 scissor_right = (u16)(regs.scissor_test.right << 4);
+    u16 scissor_bottom = (u16)(regs.scissor_test.bottom << 4);
 
     if (regs.scissor_test.mode == Regs::ScissorMode::Include) {
         // Calculate the new bounds
-        min_x = std::max(min_x, scissor_x);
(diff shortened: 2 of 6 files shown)


MoochMcGee:
@Subv You could try implementing the weird properties of the scissor test with either stencil buffers or fragment shaders.

MoochMcGee:
Oh, whoops.

Subv:
Please do not merge this yet, I have found a case where the hw renderer implementation doesn't produce the same output as the 3DS (the sw renderer is fine), this is caused by the different coordinates system that the 3DS uses

Subv:
Updated, this should be ready now

MoochMcGee:
Other than that one comment, LGTM.

yuriks:
Calling dibs on reviewing this (probably/hopefully tonight) before it's
merged, please.

--yuriks

src/video_core/rasterizer.cpp
```diff
@@ -379,6 +392,13 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
     for (u16 y = min_y + 8; y < max_y; y += 0x10) {
         for (u16 x = min_x + 8; x < max_x; x += 0x10) {
 
+            // Do not process the pixel if it's inside the scissor box and the scissor mode is set to Exclude
+            if (regs.scissor_test.mode == Regs::ScissorMode::Exclude) {
+                if (x >= scissor_x && x <= scissor_x + scissor_width &&
```
yuriks: can't you just merge these two ifs?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -666,6 +676,26 @@ void RasterizerOpenGL::SyncDepthTest() {
     state.depth.write_mask = regs.output_merger.depth_write_enable ? GL_TRUE : GL_FALSE;
 }
 
+void RasterizerOpenGL::SyncScissorTest() {
+    const auto& regs = Pica::g_state.regs;
+
+    auto viewport_y = (GLsizei)Pica::float24::FromRawFloat24(regs.viewport_size_y).ToFloat32() * 2;
+
+    // OpenGL uses different y coordinates, so negate corner offset and flip origin
```
yuriks: Messy coords used by the hw renderer strike again. (There's a sign error elsewhere that makes most of these flips necessary.)

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -666,6 +676,26 @@ void RasterizerOpenGL::SyncDepthTest() {
...
+    GLint scissor_y = viewport_y - regs.scissor_test.bottom - regs.scissor_test.GetHeight();
+
+    if (scissor_y != uniform_block_data.data.scissor_y ||
+        regs.scissor_test.right != uniform_block_data.data.scissor_x ||
+        regs.scissor_test.GetWidth() != uniform_block_data.data.scissor_width ||
+        regs.scissor_test.GetHeight() != uniform_block_data.data.scissor_height) {
```
yuriks: Is this test actually necessary? If there's barely or no difference if you just aggressively overwrite values and mark the UBO dirty then I'd rather not have the test.

yuriks: (My thinking is that chances are almost 100% that uniforms are going to be changed between two drawcalls anyway.)

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -341,6 +346,16 @@ uniform sampler2D tex[3];
 void main() {
 )";
 
+    // Append the scissor test
+    if (config.scissor_test_mode == Regs::ScissorMode::Include || config.scissor_test_mode == Regs::ScissorMode::Exclude) {
+        out += "if (scissor_height <= 0 || scissor_width <= 0) discard;\n";
+        out += "if (";
+        // Negate the condition if we have to keep only the pixels outside the scissor box
+        if (config.scissor_test_mode == Regs::ScissorMode::Include)
```
yuriks: You should do the Include test using OpenGL's standard scissoring mechanism, since it's much faster than discarding in the shader. Use the shader approach only for exclude.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -341,6 +346,16 @@ uniform sampler2D tex[3];
 void main() {
 )";
 
+    // Append the scissor test
+    if (config.scissor_test_mode == Regs::ScissorMode::Include || config.scissor_test_mode == Regs::ScissorMode::Exclude) {
```
yuriks: These tests should go after the alpha_test == never test below.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -327,6 +327,7 @@ std::string GenerateFragmentShader(const PicaShaderConfig& config) {
 
 in vec4 primary_color;
 in vec2 texcoord[3];
+layout(pixel_center_integer) in vec4 gl_FragCoord;
```
yuriks: Why are you specifying `pixel_center_integer`?

src/video_core/renderer_opengl/gl_shader_gen.cpp
Subv: I originally had it like that but changed it to handle both cases in the shader so that they're both together, i don't think segmenting the functionality like that is worth the possible performance improvements we'll get out of it.

src/video_core/renderer_opengl/gl_shader_gen.cpp
Subv: Isn't the scissor test performed before the alpha test?

src/video_core/renderer_opengl/gl_shader_gen.cpp
Subv: By default gl_FragCoord will return the pixels as (x.5, y.5) and that'd give us some problems when comparing against the scissor box bounds

neobrain: Please don't just include code for such subtle details without any sort of remark about it.

src/video_core/renderer_opengl/gl_shader_gen.cpp
yuriks: Both orders are equivalent, and the Never alpha test will short-circuit further generation of the shader.

src/video_core/renderer_opengl/gl_shader_gen.cpp
yuriks: What sorts of problems?

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -334,6 +335,10 @@ layout (std140) uniform shader_data {
     vec4 const_color[NUM_TEV_STAGES];
     vec4 tev_combiner_buffer_color;
     int alphatest_ref;
+    int scissor_x;
+    int scissor_y;
+    int scissor_width;
+    int scissor_height;
```
yuriks: You should just use `scissor_left`, `scissor_right` here as they simplify the calculations everywhere. There's no advantage to doing width/height.

src/video_core/renderer_opengl/gl_shader_gen.cpp
yuriks: It feels to me that you should actually use pixel centers, but use appropriate boundary conditions instead. (`+ 1` the right side if necessary, use `<` on the right side instead of `<=`.)

src/video_core/renderer_opengl/gl_shader_gen.cpp
yuriks: I'm unconvinced. "Include" scissoring is probably much more common, and is natively supported by the GPU. Doing discard on a shader when not necessary disables several GPU feauters like Z-buffer optimizations.

src/video_core/renderer_opengl/gl_rasterizer.cpp
hrydgard: My thinking based on experience with two emulators is that it's extremely common that two drawcalls happen without any state changes in between - for example, when using triangle stripping, drawing each strip as a drawcall. However if the 3DS has support for a restart_primitive index, maybe that isn't happening so much...

Subv:
Fantastic, this apparently breaks ALBW http://i.imgur.com/mPowT8D.png
Working on it

yuriks:
Also, the description of this PR still doesn't mention OpenGL support.

Subv:
Fixed the bottom screen bug in ALBW, on a side note, is there really a need to perform the coordinate conversions every time? Can't we do something about that?

yuriks:
What coordinate conversions?

Subv:
@yuriks See https://github.com/citra-emu/citra/pull/1263/files#diff-d1b91c65951bd9733c0e7d924161e57dR686

yuriks:
It is, but requires fixing several places at once. I've done it in an
unreleased branch (tex-forwarding) but it's tangled with and inter-
dependent with a lot of unrelated changes. I'll try to unravel and
submit the cleanup tonight if possible.

--yuriks


On Fri, Dec 4, 2015, at 11:06 AM, Sebastian Valle wrote:
> @yuriks[1] See
> https://github.com/citra-emu/citra/pull/1263/files#diff-d1b91c65951bd9733c0e7d924161e57dR686


>



Links:

  1. https://github.com/yuriks
  2. https://github.com/citra-emu/citra/pull/1263#issuecomment-162054594

Subv:
Updated to work with #1267

src/video_core/rasterizer.cpp
```diff
@@ -334,12 +334,25 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
 
-    // TODO: Proper scissor rect test!
     u16 min_x = std::min({vtxpos[0].x, vtxpos[1].x, vtxpos[2].x});
     u16 min_y = std::min({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
     u16 max_x = std::max({vtxpos[0].x, vtxpos[1].x, vtxpos[2].x});
     u16 max_y = std::max({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
 
+    // Convert the scissor box coordinates to 12.4 fixed point
+    u16 scissor_width = (u16)(regs.scissor_test.GetWidth() << 4);
+    u16 scissor_height = (u16)(regs.scissor_test.GetHeight() << 4);
```
yuriks: No reason to use width/height instead of the sides directly. This then also lets you get rid of the acessor methods entirely I think.

tfarley:
I feel like the names for anything referencing "right" and "left" should be switched across the board, assuming a -1 left, 1 right, -1 bottom, 1 top coordinate system for both the sw and hw rasterizers and the way the conditionals are written.

yuriks:
-1 is top in the PICA, it doesn't follow GL convention

tfarley:
Perhaps those names could be switched too to match the PICA. At any rate, if the horizontal axis is consistent then it is more intuitive for left to be on the left and right to be on the right imo.

src/video_core/rasterizer.cpp
```diff
@@ -334,12 +334,25 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
 
+    // Convert the scissor box coordinates to 12.4 fixed point
+    u16 scissor_left = (u16)(regs.scissor_test.GetLeft() << 4);
+    u16 scissor_top = (u16)(regs.scissor_test.GetTop() << 4);
+    u16 scissor_right = (u16)(regs.scissor_test.right << 4);
+    u16 scissor_bottom = (u16)(regs.scissor_test.bottom << 4);
+
+    if (regs.scissor_test.mode == Regs::ScissorMode::Include) {
+        // Calculate the new bounds
+        min_x = std::max(min_x, scissor_right);
```
yuriks: It feels like these are flipped? min_x should be at the left, and min_y at the top. Though I think we should actually just rename the reg fields to match and avoid confusion.

yuriks:
Can you coordinate with @fincs to fix the register names at http://3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_SCISSORTEST_MODE and in ctrulib to reflect the fact that the regs represent bounds, not origin+dimensions?

src/video_core/pica.h
```diff
@@ -112,7 +112,36 @@ struct Regs {
...
+            BitField<16, 16, u32> top_minus_1;
+        };
+
+        u32 GetTop() const {
+            return top_minus_1 + 1;
+        }
+
+        u32 GetLeft() const {
+            return left_minus_1 + 1;
+        }
```
yuriks: Should just define these as half-open intervals instead of having these functions imo.

yuriks:
I have made several cleanups and a few fixes in yuriks@13b8b6d62c7542dee6e92ed5934d59df478baa1f
It (imo) really simplifies the bounds and also fixes a off-by-one error with the GL renderer. I've fully hwtested both normal and inverted cases and with both renderers.

yuriks:
@tfarley Also, I was wrong about the PICA coordinate origin above, just fyi.

Subv:
@yuriks said he'd take over this one in the future.

yuriks:
Well, you can still do it if you're back to updating PRs.

Subv:
@JayFoxRox has showed interest in taking over this PR with @yuriks's changes.

PR #1264, opened 2015-12-03
Fragment lighting support in the HW renderer

bunnei:
* This adds support for most features of Pica's fragment lighting to our OpenGL HW renderer:
 * Per-fragment normals via quaternions or normal maps
 * 8 positional or directional lights
 * Diffuse, specular 0, specular 1, and reflectance components
 * Fresnel
 * Distance attenuation
* ~~This is RFC for now.~~ I don't know of any bugs in my implementation, but I'm not a graphics expert, and furthermore there are limitless Pica lighting configurations; that of which I'm not able to fully test myself (and I do not actually own that many games that use fragment lighting) - testing of more games is very much appreciated.
* ~~There's an issue with incorrect lighting in certain games (e.g. Super Mario 3D Land and The Legend of Zelda: A Link Between Worlds). This is due to constant normal quaternions being outputted by the vertex shader for certain 3D models. I'm not sure yet if this is a bug in our vertex shader emulation, or if it is some lighting mode that I'm not handling correctly (another reason why this is RFC).~~ These issues were fixed with #1367 and #1369.
* I have implemented most components of this in the SW renderer, but I am not submitting this due to the fact that my interpolation for quaternions and view vectors is incorrect (this is something OpenGL can do for free), and at least for the time being I don't plan on working on it anymore.


src/common/vector_math.h
```diff
@@ -429,12 +429,12 @@ class Vec4
     }
 
     // Only implemented for T=float
-    float Length() const;
+    T Length() const;
```
neobrain: Why is this relevant here? At least SNIFAE this out so that it will only be enabled for `std::is_floating_point<T>::value || std::is_same<T,float24>::value`. It definitely shouldn't be available for integers.

Subv: He only specialized it for float, it will throw a linker error if you try to use this method with any other T

bunnei: Not relevant in the PR anymore... I was using this in the SW renderer, I'll rebase it out.

src/video_core/command_processor.cpp
```diff
@@ -471,6 +471,21 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+        case PICA_REG_INDEX_WORKAROUND(lighting.lut_data[2], 0x1ca):
+        case PICA_REG_INDEX_WORKAROUND(lighting.lut_data[3], 0x1cb):
+        case PICA_REG_INDEX_WORKAROUND(lighting.lut_data[4], 0x1cc):
+        case PICA_REG_INDEX_WORKAROUND(lighting.lut_data[5], 0x1cd):
+        case PICA_REG_INDEX_WORKAROUND(lighting.lut_data[6], 0x1ce):
+        case PICA_REG_INDEX_WORKAROUND(lighting.lut_data[7], 0x1cf):
+        {
+            auto& lut_config = regs.lighting.lut_config;
+            g_state.lighting.luts[lut_config.type][lut_config.index].raw = value;
+            lut_config.index = lut_config.index + 1;
```
Subv: What happens if i just keep on writing to these registers after reaching the max index value?

src/video_core/pica.h
```diff
@@ -1163,6 +1303,25 @@ struct State {
     ShaderSetup vs;
     ShaderSetup gs;
 
+    struct {
+        union LutEntry {
+            // Used for raw access
+            u32 raw;
+
+            // LUT value, encoded as 12-bit fixed point, with 12 fraction bits
```
Subv: 12 fraction bits?

src/video_core/pica_types.h
```diff
@@ -0,0 +1,182 @@
...
+static_assert(sizeof(float24) == sizeof(float), "Shader JIT assumes float24 is implemented as a 32-bit float");
+
+struct float16 {
+    // 10 bit mantissa, 5 bit exponent, 1 bit sign
+    // TODO: No idea if this works as intended
+    static float16 FromRawFloat16(u32 hex) {
+        float16 ret;
+        if ((hex & 0xFFFF) == 0) {
+            ret.value = 0;
+        }
```
Subv: line break here

src/video_core/pica_types.h
```diff
@@ -0,0 +1,182 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+
+namespace Pica {
+
+struct float24 {
```
Subv: I'm starting to think we should use a single templated class for the various float and fixpoint formats in the PICA.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -682,6 +888,76 @@ void RasterizerOpenGL::SyncTevConstColor(int stage_index, const Pica::Regs::TevS
...
+    if (color != uniform_block_data.data.lighting_global_ambient) {
+        uniform_block_data.data.lighting_global_ambient = color;
+        uniform_block_data.dirty = true;
+    }
+}
+
+void RasterizerOpenGL::SyncLightingLUT(unsigned lut_index) {
+    std::array<std::array<GLfloat, 4>, 256> new_data;
+
+    unsigned lut_group_num = lut_index;
```
Subv: what's the point of this assignment? you aren't using lut_index anywhere else

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -71,6 +71,59 @@ struct PicaShaderConfig {
...
+        res.lighting.lut_fr.scale = regs.lighting.lut_scale.GetScale(regs.lighting.lut_scale.fr);
+
+        res.lighting.lut_rr.enable = regs.lighting.lut_enable_rr == 0;
+        res.lighting.lut_rr.abs_input = regs.lighting.abs_lut_input.rr == 0;
+        res.lighting.lut_rr.type = (Pica::Regs::LightingLutInput)regs.lighting.lut_input.rr.Value();
+        res.lighting.lut_rr.scale = regs.lighting.lut_scale.GetScale(regs.lighting.lut_scale.rr);
+
+        res.lighting.lut_rg.enable = regs.lighting.lut_enable_rg == 0;
+        res.lighting.lut_rg.abs_input = regs.lighting.abs_lut_input.rg == 0;
+        res.lighting.lut_rg.type = (Pica::Regs::LightingLutInput)regs.lighting.lut_input.rg.Value();
```
Subv: Is this not automatically converted to Pica::Regs::LightingLutInput?

Punk34:
Any chance to see it in the release version?

src/video_core/pica.h
bunnei: Yes, that's right, it's a 12-bit unsigned fixed-point number where all 12 bits are the fraction component (therefore, 0x000 is 0.0 and 0xfff is 1.0)

src/video_core/command_processor.cpp
bunnei: Good thought; I've yet to see this happen. I'll add an assertion here, and if we come across something that does this, I'll investigate further.

bunnei:
All comments have been addressed. Due to the limited feedback received, and that I believe this PR to be sufficiently complete as an initial implementation, I have removed the 'RFC' tag.

I plan on getting this merged soon, so please review this PR!

vgturtle127:
@bunnei Exciting stuff! :smile:

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -320,27 +318,233 @@ static void WriteTevStage(std::string& out, const PicaShaderConfig& config, unsi
...
     int alphatest_ref;
+    vec3 lighting_global_ambient;
+    LightSrc light_src[NUM_LIGHTS];
 };
 
 uniform sampler2D tex[3];
+uniform sampler1D lut[6];
+
+// Rotate a vector by a quaternion
+vec3 quaternion_rotate(vec4 quatertion, vec3 vector) {
```
yuriks: Typo in "quaternion". Also, I think this function reads better with just `q` and `v` as parameters.

purpasmart96: What about "quat" Instead, as for the vector I'm not sure about "vec".

yuriks: The function is a one-liner. I don't think there's any danger of any confusion with the single-variable names. Maybe adjust the comment to specify which is which is that's deemed necessary:
"// Rotate the vector `v` by the quaternion `q`."

yuriks:
To note: I'm currently looking into this and trying to iron out #1293 which causes bugs with lighting in some shaders.

citrauser:
is this updated in the source code?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -120,6 +126,19 @@ void RasterizerOpenGL::InitObjects() {
...
     glFramebufferTexture2D(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, fb_depth_texture.texture.handle, 0);
 
+    for (size_t i = 0; i < lighting_lut.size(); ++i) {
+        lighting_lut[i].Create();
+        state.lighting_lut[i].texture_1d = lighting_lut[i].handle;
+
+        glActiveTexture(GL_TEXTURE3 + i);
+        glBindTexture(GL_TEXTURE_1D, state.lighting_lut[i].texture_1d);
+
+        glTexImage1D(GL_TEXTURE_1D, 0, GL_RGBA32F, 256, 0, GL_RGBA, GL_FLOAT, lighting_lut_data[0].data());
```
tfarley: You can make the data parameter here nullptr to allocate the texture without initializing, or lighting_lut_data[0] should be initialized prior to this if 3ds actually expects some default LUT values.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -673,12 +879,81 @@ void RasterizerOpenGL::SyncTevConstColor(int stage_index, const Pica::Regs::TevS
...
+        new_data[offset][0] = Pica::g_state.lighting.luts[(lut_index * 4) + 0][offset].ToFloat();
+        new_data[offset][1] = Pica::g_state.lighting.luts[(lut_index * 4) + 1][offset].ToFloat();
+        new_data[offset][2] = Pica::g_state.lighting.luts[(lut_index * 4) + 2][offset].ToFloat();
+        new_data[offset][3] = Pica::g_state.lighting.luts[(lut_index * 4) + 3][offset].ToFloat();
+    }
+
+    if (new_data != lighting_lut_data[lut_index]) {
+        lighting_lut_data[lut_index] = new_data;
+        glActiveTexture(GL_TEXTURE3 + lut_index);
+        glTexImage1D(GL_TEXTURE_1D, 0, GL_RGBA32F, 256, 0, GL_RGBA, GL_FLOAT, lighting_lut_data[lut_index].data());
```
tfarley: Since you already allocate the LUT textures during init with glTexImage1D(), this can be glTexSubImage1D() (note that the parameters are different though) so the texture doesn't have to be completely re-created.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -490,18 +675,39 @@ void RasterizerOpenGL::SetShader() {
...
         glUniformBlockBinding(current_shader->shader.handle, block_index, 0);
-    }
 
-    // Update uniforms
-    SyncAlphaTest();
-    SyncCombinerColor();
-    auto& tev_stages = Pica::g_state.regs.GetTevStages();
-    for (int index = 0; index < tev_stages.size(); ++index)
-        SyncTevConstColor(index, tev_stages[index]);
+        // Update uniforms
```
tfarley: The following uniform updates can't just be done only for new shaders; rather need to be force-checked when this SetShader() function is called from Reset(). Otherwise the registers could change while a cached hw shader is loaded but you are in swrast mode, and not be caught/updated when switching back to hwrast until they change again. I suggest making it SetShader(bool force_uniform_check) which would be true from Reset() and false in the other calls. Then if either a new shader is generated as checked currently, OR that bool is true, do all of the sync calls below.

yuriks: Latest master recreates the entire HW rasterizer state when switching
between backends, including any shaders. There was a bug relating to
this which I remember fixing, but I'm not sure if it's only locally or
if I submitted it.

--yuriks


On Thu, Jan 7, 2016, at 05:03 PM, tfarley wrote:
> In src/video_core/renderer_opengl/gl_rasterizer.cpp[1]:


> >
> -    // Update uniforms
> -    SyncAlphaTest();
> -    SyncCombinerColor();
> -    auto& tev_stages = Pica::g_state.regs.GetTevStages();
> -    for (int index = 0; index < tev_stages.size(); ++index)
> -        SyncTevConstColor(index, tev_stages[index]);
> +        // Update uniforms The following uniform updates can't just
>          be done only for new shaders; rather need to be force-checked
>          when this SetShader() function is called from Reset().
>          Otherwise the registers could change while a cached hw shader
>          is loaded but you are in swrast mode, and not be
>          caught/updated when switching back to hwrast until they
>          change again. I suggest making it SetShader(bool
>          force_uniform_check) which would be true from Reset() and
>          false in the other calls. Then if either a new shader is
>          generated as checked currently, OR that bool is true, do all
>          of the sync calls below.


>



Links:

  1. https://github.com/citra-emu/citra/pull/1264#discussion_r49148277
  2. https://github.com/citra-emu/citra/pull/1264/files#r49148277

yuriks:
In addition to the shader work, I'm currently trying to track down bugs in both Bravely Default (title screen) and ALBW where some shading effects don't happen. BraDef has a normal pointing into the screen, so the highlight effect in the title isn't visible. ALBW looks very flat overall, but notably the shading on the beveled triangle edges in the title is also flat.

neobrain:

2016-01-30T00:08:23Z: https://github.com/citra-emu/citra/compare/3ab6d77c206c33e5036fcde1de0663c2f72777ea...bunnei:7f65c06b4930835229a8847a8649aa127579002d

bunnei:
This should be ready to ship. Call for final reviews.

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+
+    enum class LightingConfig {
+        Config0 = 0,
+        Config1 = 1,
+        Config2 = 2,
+        Config3 = 3,
+        Config4 = 4,
+        Config5 = 5,
+        Config6 = 6,
+        Config7 = 8,
```
yuriks: If this is intentional you should probably leave a comment. Gah.

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+        Both = PrimaryAlpha | SecondaryAlpha, ///< Both primary and secondary lighting alphas are affected by fresnel
+    };
+
+    enum class LightingScale {
+        Scale1 = 0,
+        Scale2 = 1,
+        Scale4 = 2,
+        Scale8 = 3,
+        Scale1_4 = 6,
+        Scale1_2 = 7,
```
yuriks: Comment with 1/4 and 1/2.

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+        TangentMap = 2,
+    };
+
+    union LightColor {
+        BitField< 0, 10, u32> b;
+        BitField<10, 10, u32> g;
+        BitField<20, 10, u32> r;
+
+        Math::Vec3f ToVec3f() const {
+            return Math::MakeVec((f32)r / 255.f, (f32)g / 255.f, (f32)b / 255.f);
```
yuriks: The fields are 10 bit, but this divides by 255. Is this correct? Add something explaining.

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+                };
+                union {
+                    BitField< 0, 16, u32> z;
+                };
+
+                INSERT_PADDING_WORDS(0x3);
+
+                union {
+                    BitField<0, 1, u32> directional;
+                    BitField<1, 1, u32> two_sided_diffuse; // 1: GL_TRUE, 0: GL_FALSE; when disabled, clamp dot-product to 0
```
yuriks: GL_TRUE/FALSE is quite unnecessary here. Just say which one enables

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+            BitField<0, 20, u32> dist_atten_scale;
+
+            INSERT_PADDING_WORDS(0x4);
+        };
+        static_assert(sizeof(LightSrc) == 0x10 * sizeof(u32), "LightSrc structure must be 0x10 words");
+
+        LightSrc light[8];
+        LightColor global_ambient; // emission + (material.ambient * lighting.ambient)
+        INSERT_PADDING_WORDS(0x1);
+        BitField<0, 3, u32> src_num; // number of enabled lights - 1
```
yuriks: Rename to num_lights, or something?

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+
+        union {
+            BitField< 2, 2, LightingFresnelSelector> fresnel_selector;
+            BitField< 4, 4, LightingConfig> config;
+            BitField<22, 2, u32> bump_selector; // 0: Texture 0, 1: Texture 1, 2: Texture 2
+            BitField<27, 1, u32> clamp_highlights; // 1: GL_TRUE, 0: GL_FALSE
```
yuriks: GL_TRUE/GL_FALSE etc.

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+            BitField<30, 1, u32> bump_renorm; // 0: GL_TRUE, 1: GL_FALSE
+        };
+
+        union {
+            BitField<16, 1, u32> lut_enable_d0; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<17, 1, u32> lut_enable_d1; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<19, 1, u32> lut_enable_fr; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<20, 1, u32> lut_enable_rr; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<21, 1, u32> lut_enable_rg; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<22, 1, u32> lut_enable_rb; // 0: GL_TRUE, 1: GL_FALSE
```
yuriks: These could be `lut_disable_`* instead because they're inverted

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+            // corresponding light
+
+            BitField<24, 1, u32> dist_atten_enable_light_0; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<25, 1, u32> dist_atten_enable_light_1; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<26, 1, u32> dist_atten_enable_light_2; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<27, 1, u32> dist_atten_enable_light_3; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<28, 1, u32> dist_atten_enable_light_4; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<29, 1, u32> dist_atten_enable_light_5; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<30, 1, u32> dist_atten_enable_light_6; // 0: GL_TRUE, 1: GL_FALSE
+            BitField<31, 1, u32> dist_atten_enable_light_7; // 0: GL_TRUE, 1: GL_FALSE
```
yuriks: Can't help but feel like we should just make this a 8-bit bitmask instead...

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+                                        dist_atten_enable_light_6, dist_atten_enable_light_7 };
+            return enable[index] == 0;
+        }
+
+        union {
+            BitField<0, 8, u32> index;      ///< Index at which to set data in the LUT
+            BitField<8, 5, u32> type;       ///< Type of LUT for which to set data
+        } lut_config;
+
+        BitField<0, 1, u32> disable;
```
yuriks: "disable"?

src/video_core/pica.h
```diff
@@ -641,7 +644,254 @@ struct Regs {
...
+        INSERT_PADDING_WORDS(0x1);
+
+        // When data is written to any of these registers, it gets written to the lookup table of
+        // the selected type at the selected index, specified above in the `lut_config` register.
+        // With each write, `lut_config.index` is incremented. It does not matter which of these
+        // registers is written to, the behavior will be the same.
+        u32 lut_data[8];
+
+        union {
+            BitField< 1, 1, u32> d0; // 0: GL_TRUE, 1: GL_FALSE
```
yuriks: Comment about what these do

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -189,6 +283,21 @@ class RasterizerOpenGL : public VideoCore::RasterizerInterface {
         GLfloat tex_coord0[2];
         GLfloat tex_coord1[2];
         GLfloat tex_coord2[2];
+        GLfloat normquat[4];
+        GLfloat view[3];
+    };
+
+    struct LightSrc {
+        std::array<GLfloat, 3> specular_0;
+        INSERT_PADDING_WORDS(1);
```
yuriks: Use `alignas(16) std::array<GLfloat, 3> specular_0;`.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -198,11 +307,14 @@ class RasterizerOpenGL : public VideoCore::RasterizerInterface {
         std::array<GLfloat, 4> tev_combiner_buffer_color;
         GLint alphatest_ref;
         GLfloat depth_offset;
-        INSERT_PADDING_BYTES(8);
+        INSERT_PADDING_WORDS(2);
+        std::array<GLfloat, 3> lighting_global_ambient;
```
yuriks: Use alignas

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -298,4 +432,7 @@ class RasterizerOpenGL : public VideoCore::RasterizerInterface {
     OGLBuffer vertex_buffer;
     OGLBuffer uniform_buffer;
     OGLFramebuffer framebuffer;
+
+    std::array<OGLTexture, 6> lighting_lut;
+    std::array<std::array<std::array<GLfloat, 4>, 256>, 6> lighting_lut_data;
```
yuriks: General comment: At this point, I think it's useful to alias `std::array<GLfloat, n>` to GLvec*n* to improve readability.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -320,28 +318,234 @@ static void WriteTevStage(std::string& out, const PicaShaderConfig& config, unsi
         out += "next_combiner_buffer.a = last_tex_env_out.a;\n";
 }
 
+/// Writes the code to emulate fragment lighting
+static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
+    // Define lighting globals
+    out += "vec4 diffuse_sum = vec4(0.0, 0.0, 0.0, 1.0);\n";
+    out += "vec4 specular_sum = vec4(0.0, 0.0, 0.0, 1.0);\n";
+    out += "vec3 light_vector = vec3(0.0);\n";
+    out += "vec3 refl_value = vec3(0.0);\n";
```
yuriks: You can merge these into one statement:
```
out += "vec4 diffuse_sum = vec4(0.0, 0.0, 0.0, 1.0);\n"
       "vec4 specular_sum = vec4(0.0, 0.0, 0.0, 1.0);\n"
       ...;

File: src/video_core/renderer_opengl/gl_shader_gen.cpp [hunk 26]
```diff
@@ -320,28 +318,234 @@ static void WriteTevStage(std::string& out, const PicaShaderConfig& config, unsi
...
+    // Compute fragment normals
+    if (config.lighting.bump_mode == Pica::Regs::LightingBumpMode::NormalMap) {
+        // Bump mapping is enabled using a normal map, read perturbation vector from the selected texture
+        std::string bump_selector = std::to_string(config.lighting.bump_selector);
+        out += "vec3 surface_normal = 2.0 * texture(tex[" + bump_selector + "], texcoord[" + bump_selector + "]).rgb - 1.0;\n";
+
+        // Recompute Z-component of perturbation if 'renorm' is enabled, this provides a higher precision result
+        if (config.lighting.bump_renorm) {
+            std::string val = "1.0 - (surface_normal.x*surface_normal.x + surface_normal.y*surface_normal.y)";
+            out += "surface_normal.z = (" + val + ") > 0 ? sqrt(" + val + ") : 0;\n";
```
yuriks: `sqrt(max(val, 0.0))`

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -320,28 +318,234 @@ static void WriteTevStage(std::string& out, const PicaShaderConfig& config, unsi
...
+            std::string scale = std::to_string(light_config.dist_atten_scale);
+            std::string bias = std::to_string(light_config.dist_atten_bias);
+            std::string index = "(" + scale + " * length(-view - " + light_src + ".position) + " + bias + ")";
+            index = "((clamp(" + index + ", 0.0, FLOAT_255)))";
+            const unsigned lut_num = ((unsigned)Regs::LightingSampler::DistanceAttenuation + light_config.num);
+            dist_atten = GetLutValue((Regs::LightingSampler)lut_num, index);
+        }
+
+        // If enabled, clamp specular component if lighting result is negative
+        std::string clamp_highlights = config.lighting.clamp_highlights ? "(dot(light_vector, normal) <= 0.0 ? 0.0 : 1.0)" : "1.0";
```
yuriks: `sign(...)`

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -320,28 +318,234 @@ static void WriteTevStage(std::string& out, const PicaShaderConfig& config, unsi
...
+}
 
 void main() {
+vec4 primary_fragment_color = vec4(0.0);
+vec4 secondary_fragment_color = vec4(0.0);
 )";
 
+    if (config.lighting.enable)
+        WriteLighting(out, config);
+
```
yuriks: Shouldn't this be below the alpha test discard?

src/video_core/shader/shader.cpp
```diff
@@ -134,11 +134,12 @@ OutputVertex Run(UnitState<false>& state, const InputVertex& input, int num_attr
...
     }
 
-    LOG_TRACE(Render_Software, "Output vertex: pos (%.2f, %.2f, %.2f, %.2f), quat (%.2f, %.2f, %.2f, %.2f), col(%.2f, %.2f, %.2f, %.2f), tc0(%.2f, %.2f)",
+    LOG_TRACE(Render_Software, "Output vertex: pos(%.2f, %.2f, %.2f, %.2f), quat(%.2f, %.2f, %.2f, %.2f), "
+        "col(%.2f, %.2f, %.2f, %.2f), tc0(%.2f, %.2f), view(%.2f, %.2f, %.2f)",
         ret.pos.x.ToFloat32(), ret.pos.y.ToFloat32(), ret.pos.z.ToFloat32(), ret.pos.w.ToFloat32(),
         ret.quat.x.ToFloat32(), ret.quat.y.ToFloat32(), ret.quat.z.ToFloat32(), ret.quat.w.ToFloat32(),
         ret.color.x.ToFloat32(), ret.color.y.ToFloat32(), ret.color.z.ToFloat32(), ret.color.w.ToFloat32(),
-        ret.tc0.u().ToFloat32(), ret.tc0.v().ToFloat32());
+        ret.tc0.u().ToFloat32(), ret.tc0.v().ToFloat32(), ret.view.x.ToFloat32(), ret.view.y.ToFloat32(), ret.view.z.ToFloat32());
```
yuriks: Add this on a separate line.

src/video_core/pica_types.h
```diff
@@ -0,0 +1,144 @@
...
+        const int bias = 128 - (1 << (E - 1));
+        const int exponent = (hex >> M) & ((1 << E) - 1);
+        const unsigned mantissa = hex & ((1 << M) - 1);
+
+        if (hex & ((1 << (width - 1)) - 1))
+            hex = ((hex >> (E + M)) << 31) | (mantissa << (23 - M)) | ((exponent + bias) << 23);
+        else
+            hex = ((hex >> (E + M)) << 31);
+
+        res.value = *(float*)&hex;
```
lioncash: this should probably be a memcpy

src/video_core/command_processor.cpp
```diff
@@ -418,10 +418,10 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                         uniform[3 - i] = float24::FromFloat32(*(float*)(&uniform_write_buffer[i]));
                 } else {
                     // TODO: Untested
-                    uniform.w = float24::FromRawFloat24(uniform_write_buffer[0] >> 8);
-                    uniform.z = float24::FromRawFloat24(((uniform_write_buffer[0] & 0xFF)<<16) | ((uniform_write_buffer[1] >> 16) & 0xFFFF));
-                    uniform.y = float24::FromRawFloat24(((uniform_write_buffer[1] & 0xFFFF)<<8) | ((uniform_write_buffer[2] >> 24) & 0xFF));
-                    uniform.x = float24::FromRawFloat24(uniform_write_buffer[2] & 0xFFFFFF);
+                    uniform.w = float24::FromRaw(uniform_write_buffer[0] >> 8);
+                    uniform.z = float24::FromRaw(((uniform_write_buffer[0] & 0xFF)<<16) | ((uniform_write_buffer[1] >> 16) & 0xFFFF));
```
lioncash: spaces before and after <<

bunnei:
Fixes #695

neobrain:

2016-02-05T02:51:59Z: https://github.com/citra-emu/citra/compare/3ab6d77c206c33e5036fcde1de0663c2f72777ea...bunnei:1697fe1e765e59c133e5f23008fcb1bc254c8b89

neobrain:

2016-02-05T03:01:31Z: https://github.com/citra-emu/citra/compare/3ab6d77c206c33e5036fcde1de0663c2f72777ea...bunnei:a80f84bbb292eea3756e0e0d05f48f4bc2baffbe

yuriks:
LGTM other than the nit

tfarley:
LGTM

neobrain:

2016-02-05T22:23:52Z: https://github.com/citra-emu/citra/compare/69e2453649b30962ac68ef54ffbf30ab4f1a9842...bunnei:19557aaab3434a9a6e9b6730a76923de053084cd

purpasmart96:
It's finally in.

vgturtle127:
Yay! Now we just need the audio pull requests to be merged!

PR #1266, opened 2015-12-04
HLE/Applets: Implemented a dummy Mii Selector applet.

Subv:
This prevents some games (like Super Mario 3D Land) from freezing when trying to launch it, however, it's not complete and won't let you go past Mii selection as the parameter structure hasn't been reverse engineered yet.

### DIFF
--- a/src/core/hle/applets/mii_selector.cpp
+++ b/src/core/hle/applets/mii_selector.cpp
@@ -1,0 +1,75 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <cstring>
+#include <string>
+
+#include "common/assert.h"
+#include "common/logging/log.h"
+#include "common/string_util.h"
+
+#include "core/hle/applets/mii_selector.h"
+#include "core/hle/kernel/kernel.h"
+#include "core/hle/kernel/shared_memory.h"
+#include "core/hle/result.h"
+
+#include "video_core/video_core.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+namespace HLE {
+namespace Applets {
+
+MiiSelector::MiiSelector(Service::APT::AppletId id) : Applet(id), started(false) {
+    // Create the SharedMemory that will hold the framebuffer data
+    // TODO(Subv): What size should we use here?
+    using Kernel::MemoryPermission;
+    framebuffer_memory = Kernel::SharedMemory::Create(0x1000, MemoryPermission::ReadWrite, MemoryPermission::ReadWrite, "MiiSelector Memory");
+}
+
+ResultCode MiiSelector::ReceiveParameter(const Service::APT::MessageParameter& parameter) {
+    if (parameter.signal != static_cast<u32>(Service::APT::SignalType::LibAppJustStarted)) {
+        LOG_ERROR(Service_APT, "unsupported signal %u", parameter.signal);
+        UNIMPLEMENTED();
+        // TODO(Subv): Find the right error code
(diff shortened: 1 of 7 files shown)


purpasmart96:
LGTM

src/core/hle/applets/mii_selector.h
```diff
@@ -0,0 +1,38 @@
+// Copyright 2015 Citra Emulator Project
```
bunnei: *2016

src/core/hle/applets/mii_selector.cpp
```diff
@@ -0,0 +1,75 @@
+// Copyright 2015 Citra Emulator Project
```
bunnei: *2016

bunnei:
LGTM

src/core/hle/applets/mii_selector.h
bunnei: ...Feel free to ignore these comments if you feel like it, since this _was_ written in 2015

src/core/hle/applets/mii_selector.cpp
```diff
@@ -0,0 +1,75 @@
...
+namespace Applets {
+
+MiiSelector::MiiSelector(Service::APT::AppletId id) : Applet(id), started(false) {
+    // Create the SharedMemory that will hold the framebuffer data
+    // TODO(Subv): What size should we use here?
+    using Kernel::MemoryPermission;
+    framebuffer_memory = Kernel::SharedMemory::Create(0x1000, MemoryPermission::ReadWrite, MemoryPermission::ReadWrite, "MiiSelector Memory");
+}
+
+ResultCode MiiSelector::ReceiveParameter(Service::APT::MessageParameter const& parameter) {
```
yuriks: Weird const positioning. Also happening in the rest of the file

src/core/hle/applets/mii_selector.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    Service::APT::SendParameter(result);
+    return RESULT_SUCCESS;
+}
+
+ResultCode MiiSelector::StartImpl(Service::APT::AppletStartupParameter const& parameter) {
+    started = true;
+
+    // TODO(Subv): Set the expected fields in the response buffer before resending it to the application.
+    // TODO(Subv): Reverse the parameter format for the Mii Selector
```
yuriks: Where is this buffer? What data are we returning right now? Can we at least return a deterministic zeroed buffer instead of random stuff? (If applicable)

yuriks: Right now it looks like you're just ping-ponging the buffer the app sent back, which doesn't make much sense to me.

src/core/hle/applets/mii_selector.h
```diff
@@ -0,0 +1,38 @@
...
+#include "core/hle/result.h"
+#include "core/hle/service/apt/apt.h"
+
+namespace HLE {
+namespace Applets {
+
+class MiiSelector final : public Applet {
+public:
+    MiiSelector(Service::APT::AppletId id);
+    ~MiiSelector() {}
```
yuriks: I think you don't need this.

yuriks:
わかりませんＬＯＬ LGTM ¯\_(ツ)_/¯

src/core/hle/applets/mii_selector.cpp
Subv: That's what's supposed to happen, the Mii Applet only fills out some fields and returns the structure back to the application (just like the swkbd). This will need some more reversing in the future though

Subv:
Updated, this should be ready to go.

PR #1267, opened 2015-12-05
OpenGL: Flip framebuffers during transfer rather than when rendering

yuriks:
This simplifies the coordinate system used by the OpenGL renderer so that screen-based coords don't need to be flipped by taking into account the framebuffer size.

cc @Subv

### DIFF
--- a/src/video_core/renderer_opengl/gl_state.cpp
+++ b/src/video_core/renderer_opengl/gl_state.cpp
@@ -11,6 +11,7 @@ OpenGLState::OpenGLState() {
     // These all match default OpenGL values
     cull.enabled = false;
     cull.mode = GL_BACK;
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -67,6 +68,10 @@ void OpenGLState::Apply() {
         glCullFace(cull.mode);
```
Subv: Are you setting cull.mode anywhere?

yuriks: Look a few lines up in the GH diff. (No, I never change it from the default though.)

Subv:
LGTM

bunnei:
LGTM

PR #1269, opened 2015-12-06
GPU/PrimitiveAssembler: Fixed drawing triangle fans.

Subv:
It was skipping the second vertex assignment and using uninitialized garbage when assembling the corresponding triangle.

Fixes the sf2dlib sample.


yuriks:
LGTM

xerpi:
The bug was also affecting the sf2dlib's draw_circle function :')

bunnei:
LGTM

PR #1270, opened 2015-12-06
HWRenderer/TextureCache: Flush the texture cache when bit 16 of register 0x80 is set to 1

Subv:
Taken from 3dbrew, this fixes a very old texture cache issue with Castle Conqueror (and probably more)


neobrain:
What does our texture cache have to do with the PICA200's texture cache?

yuriks:
Yeah, this seems like a misfix. Our texture cache doesn't mean the same thing as the PICA one. (We don't implement or need to worry about the PICA one. It has to be flushed every time you change textures after rendering.)

Subv:
Castle Conqueror isn't issuing flush commands after changing the textures used for rendering, however it is writing to this register multiple times before triggering a draw again

yuriks:
*Every* game will issue a write to this register after *every* texture pointer change. Flushing our texture cache here makes it essentially pointless.

PR #1281, opened 2015-12-12
svc: Fix compilation with LOG_TRACE enabled

lioncash:
another reason why conditional macros outside of determining system specifics is a bad thing.


Subv:
LGTM

PR #1283, opened 2015-12-15
HLE/Sockets: Fixed the buffer offset in recvfrom.

### LINKED ISSUE #1277: In homebrew recvfrom doesn't work if src_addr isn't NULL
When trying to use recvfrom with src_addr, Citra throws unmapped write8 0x00000000 @ errors.
Example program
https://gist.github.com/StraDaMa/ea404fddc59bcef3a821
This example works as expected with real 3DSs and outputs "Received: XXXXXXXX from XXX.XXX.XXX.XXX" but does not work when Citra is server. Citra can be a client to 3DS server.

Subv:
Closes #1277


yuriks:
Can we turn this into a struct or something so we don't have to sling magic offsets around?

Subv:
Updated to use a structure

bunnei:
LGTM

Issue #136, opened 2014-10-12
Enable -Wall and -Wpedantic and fix all compiler warnings


neobrain:
`-Wall` and `-Wpedantic` are gcc compiler options that specify the amount of warnings to spam during compilation. I think it's beneficial for code quality to catch as many warnings as possible (warnings which turn out to be useless can be disabled explicitly later on).

Additionally, we should make an effort to get the build warning-free. This allows us to enable `-Werror` (warnings are treated as errors) on the travis build bot, so that new code is guaranteed not to introduce new warnings to the code base.

GPUCode:
Impmented in https://github.com/citra-emu/citra/pull/6456 and https://github.com/citra-emu/citra/pull/6602 for msvc and https://github.com/citra-emu/citra/pull/6602 for gcc/clang

PR #1284, opened 2015-12-16
Compile with -Werror

Dettorer:
This is a different approach to resolve issue https://github.com/citra-emu/citra/issues/136 than PR https://github.com/citra-emu/citra/pull/1260.

In this PR I add -Werror to the compile flags and fix the warnings I get without more warnings enabled than the default ones. In future PRs I will add more warnings (like -Wall, -Wextra, maybe -Wpedantic and others).

It won't compile until my other PR on nihstro (https://github.com/neobrain/nihstro/pull/46) gets merged, I will add a commit that updates the submodule when (and if) it does.

I tested on an up-to-date archlinux against clang and gcc. I don't know MSVC well so I didn't touch it's flags, but I tested the build in a Windows 8 VM (I don't own a physical Windows machine with a GPU).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,8 +64,15 @@ if (NOT DEFINED ARCHITECTURE)
 endif()
 message(STATUS "Target architecture: ${ARCHITECTURE}")
 
+if (NOT CMAKE_GENERATOR STREQUAL Xcode)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
+endif()
+
 if (NOT MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1y -Wno-attributes -pthread")
+    # Currently, clang incorrectly throws this warning with enums, see https://llvm.org/bugs/show_bug.cgi?id=16154
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-tautological-constant-out-of-range-compare")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")
 
     if (ARCHITECTURE_x86_64)
@@ -233,8 +240,8 @@ include_directories(${INI_PREFIX})
 add_subdirectory(${INI_PREFIX})
 
 add_subdirectory(externals/glad)
-include_directories(externals/microprofile)
-include_directories(externals/nihstro/include)
+include_directories(SYSTEM externals/microprofile) # TODO: remove "SYSTEM" and actually fix the warnings upstream
+include_directories(SYSTEM externals/nihstro/include) # TODO: remove "SYSTEM" and actually fix the warnings upstream
 
 if (MSVC)
     add_subdirectory(externals/getopt)
--- a/src/citra_qt/debugger/graphics_framebuffer.cpp
+++ b/src/citra_qt/debugger/graphics_framebuffer.cpp
@@ -346,5 +346,8 @@ u32 GraphicsFramebufferWidget::BytesPerPixel(GraphicsFramebufferWidget::Format f
         case Format::RGBA4:
         case Format::D16:
             return 2;
+        case Format::Unknown:
+            ASSERT_MSG(false, "Unknown frame buffer format.");
+            return 0;
     }
 }
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -55,10 +55,12 @@ const u32 SIGTERM = 15;
 const u32 MSG_WAITALL = 8;
 #endif
 
-const u32 R0_REGISTER = 0;
-const u32 R15_REGISTER = 15;
-const u32 CPSR_REGISTER = 25;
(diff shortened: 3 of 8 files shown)


src/video_core/renderer_opengl/pica_to_gl.h
```diff
@@ -12,20 +12,12 @@
...
 
-inline GLenum TextureFilterMode(Pica::Regs::TextureConfig::TextureFilter mode) {
+inline GLenum TextureFilterMode(const Pica::Regs::TextureConfig::TextureFilter& mode) {
     static const GLenum filter_mode_table[] = {
         GL_NEAREST,  // TextureFilter::Nearest
         GL_LINEAR    // TextureFilter::Linear
     };
 
-    // Range check table for input
-    if (mode >= ARRAY_SIZE(filter_mode_table)) {
```
Subv: Why do you remove these bounds checks?

Subv:
These are all the warnings? huh

src/video_core/renderer_opengl/pica_to_gl.h
Dettorer: I explained it in the commit message, mode is an enum that can only be 0 or 1, so it will always be strictly inferior to 2.
It would make sense to keep the check if the enum was likely to be updated with new features, but looking at the texture register on 3dbrew told me that it can only have these value and no more.

But right, 3dbrew may still be wrong, I'll look into making this check a static assert rather a runtime one.

Dettorer:
Yes, these are the only warnings I get when just adding -Werror and compiling with gcc 5.2 and clang 3.7.

CMakeLists.txt
```diff
@@ -65,8 +65,8 @@ endif()
 message(STATUS "Target architecture: ${ARCHITECTURE}")
 
 if (NOT MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -pthread")
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-attributes -Wno-tautological-constant-out-of-range-compare -pthread -Werror")
```
lioncash: There should be an indication that -Wtautological-constant-out-of-range-compare is disabled because of a bug (and reference the bug number), so that it can be enabled again in the future.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -55,10 +55,12 @@ const u32 SIGTERM = 15;
 const u32 MSG_WAITALL = 8;
 #endif
 
-const u32 R0_REGISTER = 0;
-const u32 R15_REGISTER = 15;
-const u32 CPSR_REGISTER = 25;
-const u32 FPSCR_REGISTER = 58;
+enum gdb_register {
```
lioncash: GDBRegister or just remove the name and leave a comment indicating what the meaning of them is.

Dettorer: I think I prefer giving it a real name (GDBRegister), it documents the enum as much as a comment and appears more clearly in the documentation.

Subv:
There are conflicts in the PR, please fix them.

I'd like to get this merged as soon as possible

Dettorer:
I rebased my branch, but it won't compile until https://github.com/neobrain/nihstro/pull/46 gets merged to nihstro.

Dettorer:
It seems there is some problem with my PR on XCode, I don't have the time to look into it right now but I'll get it fixed later today.

neobrain:

2016-01-10T12:59:18Z: https://github.com/citra-emu/citra/compare/0144b8e1dafeb25d9e784f78639e2f7c18ce39cc...Dettorer:6ced6fd918e4eb25f2ea91c28cc1b589313071b2

neobrain:

2016-01-10T13:11:50Z: https://github.com/citra-emu/citra/compare/0144b8e1dafeb25d9e784f78639e2f7c18ce39cc...Dettorer:26cf0d86a9e11377c63bf189faf0ce26b94aa536

src/citra_qt/debugger/graphics_framebuffer.cpp
```diff
@@ -345,5 +345,7 @@ u32 GraphicsFramebufferWidget::BytesPerPixel(GraphicsFramebufferWidget::Format f
         case Format::RGBA4:
         case Format::D16:
             return 2;
+        case Format::Unknown:
```
Subv: IMO this should be an assert or an UNREACHABLE

JayFoxRox:
Is this dead?
If so this should be closed.
If not this should be updated.

Dettorer:
Sorry for making this linger, I was waiting for a PR on nihstro to get merged but it will require more work than I thought (I will comment on it in a few minutes).

I rebased the PR and disabled -Werror on nihstro, I suggest two choices:

1. If the other devs really want -Werror, we can merge this in it's current state, where -Werror is enabled for citra itself but warnings are completely disabled on nihstro and microprofile.
2. If you prefer this PR to be more complete, it's probably better to close it while I'm working on nihstro, as it may take some time.

I have a really bad feeling about silencing the warnings on nihstro and microprofile, so I would recommend the second option.

bunnei:
@Dettorer taking your advice and closing this one. Please resubmit once the other issues are resolved!

PR #1287, opened 2015-12-20
dyncom: Minor changes

lioncash:


src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -4477,11 +4477,6 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
             inst_cream->get_addr(cpu, inst_cream->inst, addr);
 
             cpu->Reg[BITS(inst_cream->inst, 12, 15)] = cpu->ReadMemory8(addr);
-
-            if (BITS(inst_cream->inst, 12, 15) == 15) {
```
Subv: are the old code and the new code equivalent? cpu->Reg[15] will increase by cpu->GetInstructionSize() now.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -6196,8 +6162,16 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
             ldst_inst* inst_cream = (ldst_inst*)inst_base->component;
             inst_cream->get_addr(cpu, inst_cream->inst, addr);
 
-            unsigned int value = cpu->Reg[BITS(inst_cream->inst, 12, 15)];
+            const u32 previous_mode = cpu->Mode;
+            const u32 rt_index = BITS(inst_cream->inst, 12, 15);
+
+            u32 value = cpu->Reg[rt_index];
+            if (rt_index == 15)
```
Subv: So this was a bug?

lioncash: yes

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
lioncash: LDRB doesn't allow storing to the PC.

lioncash:
Wait before merging this. I need to add something to this (just waiting for the alcohol to wear off :^), friend's birthday party)

lioncash:
All good now

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
yuriks: Does that actually get explicitly checked and forbidden, or will executing that just cause execution to silently take a bogus codepath here?

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
yuriks: Nothing actually uses these instructions though (since this isn't a kernel) so it doesn't make much of a difference.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
lioncash: Bogus codepath, but since ARM disallows that in the manual, allowing it to continue as if nothing is wrong can hide bugs and other issues. Usually you'd catch this by signaling an undefined instruction, but that isn't implemented yet.

lioncash:
`[11:41] <@yuriks> Lioncash: looks pretty ok, I guess, that's as much as I can say :D`

PR #1288, opened 2015-12-20
[RFC] AOT Recompiler, Take 2. 

ILOVEPIE:
Updated Syrco's AOT Recompiler ( RFC #722 ) to the latest codebase, need feedback on what needs to be fixed and how to go about it.

Please note that there is no benefit to a JIT over an AOT in the case of the 3ds due to the fact that all executable memory must be available at the start of program execution, which means that an AOT is perfect for the job. (the only caviat is CRO files but those can be extracted and recompiled beforehand too).

TODO:
- fix overuse of auto.
- fix typos.
- add licence headers.
- think ahead about CRO files.

DONE:
- prove effectiveness (loot gets a significant speedup (close to 50%) over interpereted when hardware renderer and shader jit is enabled).
- fix compilation time (loot takes 1 min now, as does mm).
- add 64bit support.
- fix file-naming scheme of recompiler.
- update to latest LLVM release 3.7.X

Please feel free to post any reasonable issues with the code below and i will add that to do the TODO list, that way we can keep this organized.


lioncash:
Have the instructions been hardware verified for correctness? If not, now's a good time to start.

Also I believe significant discussion in the past was that AOT was not the preferred method to go in, surely you have a good reason why all of this discussion was seemingly ignored (which you were a part of).

hrydgard:
JIT does have one major benefit over AOT: instant startup time!

popalous:
@ILOVEPIE 
How do I compile this? I installed LLVM, i am not sure why cmake gives me the errors:

Could not find a package configuration file provided by "LLVM" with any of
  the following names:

    LLVMConfig.cmake
    llvm-config.cmake

RHY3756547:
I needed to manually compile LLVM from source with CMake+msvc, then set LLVM_DIR to llvm-3.7.0.src/build/share/llvm/cmake/. Note that LLVM takes a long time to compile...

popalous:
Thanks

shinyquagsire23:
CRO files can't be easily extracted and recompiled beforehand with a lot of games since you can get games like Sm4sh which do funky packaging and actually unload and reload CROs constantly. Pokemon in particular unloads and reloads CROs at almost every screen transisition.

ILOVEPIE:
How does the CRO loader load the CROs if they are packaged funny?

shinyquagsire23:
With Smash they're in a SARC archive labeled only by a CRC I believe, and the CRO will be read into RAM to be loaded by RO services. Most games will have CROs just in romFS but my point was that searching romFS to load all CROs early might not be doable. You'd have to recompile as they're loaded most likely.

purpasmart96:
Citra's implementation of the cro loader service is pretty much only stubbed, but that's all it needed for most games, usually the game takes care of all the files it uses, so with cro files, they would be need to be separately loaded into memory and then they can

mailwl:
@purpasmart96 .cro is not crypted? I mean, in decrypted roms, like pokemon

purpasmart96:
@mailwl It uses rsa-2048 if I'm not mistaken.

popalous:
@ILOVEPIE @RHY3756547 @shinyquagsire23 
Can one of you guys upload the LLVM that works?
the binary on the official site is missing certain needed folders. (i get errors when trying to compile from source)
If the whole LLVM folder is too big to upload, just upload the llvm-3.7.0.src/build/share/llvm/cmake/
folder. Thanks!

ILOVEPIE:
@lioncash we never came to a consensus on it, and this does provide a significant speedup with the latest versions of citra, and it would be less resource intensive and faster than a JIT, I'd rather get into the debate as to whither or not we should go this route in an actual PR. An RFC isn't a good place for debating that. Let's see which route produces the best results and follow the one that does.

ILOVEPIE:
@popalous you don't use a binary, you have to use the llvm static librarys, which means you need to compile llvm yourself. from the 3.7.X sourcecode.

lioncash:
@ILOVEPIE A RFC is actually a very good place to do that. If it's done in a PR and the work gets rejected for one reason or another, you'd have done more work for nothing.

popalous:
@ILOVEPIE 
I am getting multiple errors of which i have been trying to fix all day.
Do you have this citra branch compiled for windows?
Also, i remember you from the #citra freenode, you helped me with a dll problem i had through join.me screenshare, then you showed me OOT on your twitch stream.

neobrain:
@ILOVEPIE How do you know this approach actually works properly in the first place?

ILOVEPIE:
@neobrain Static recompilation works correctly in situations where the machine code is immutable and all machine code is available at start of execution, exactly the circumstances that are true for the 3ds, even the CROs must load fully into memory before executing, and they can't self extract etc.

neobrain:
@ILOVEPIE: Could you please provide a source for that claim?

yuriks:
Can we please stop beating around the bush here? This PR won't be accepted because we don't want to pay the cost incurred by this approach:
 - LLVM is a very hefty dependency and a significant obstacle to building Citra if included
 - AOT compilation as a separate step has been decided to be too user-unfriendly due to the necessary overhead required for the compilation
 - And more importantly, AOT isn't required to get good CPU performance for the 3DS, so it's supposed advantages aren't very valuable in this scope. A JIT should be much simpler design-wise than this AOT approach, will have none of the startup delay and will provide adequate performance.

Feel free to help with a JIT implementation, but there's no interest from any of the maintainers in an AOT recompiler.

neobrain:
Also, neither this nor the original PR describe how any of this code works on a high level at all. Please provide a brief description of
* How this uses LLVM to implement AOT
* Why you use LLVM for the task
* How we can be sure it actually works and will not be a deadend in the future
* Anything else we should know and of which you currently seem to assume we should just figure out ourselves by reading the diff.

ILOVEPIE:
It's an RFC not a PR. I'm trying to get feedback don't close this.

ILOVEPIE:
let me get an explanation put togeather then i'll add it to the thread.

yuriks:
You can get all the feedback you want, but this won't be merged anyway. This has already been discussed to death. Being stubborn won't help anything.

ILOVEPIE:
*sigh* this is the problem, you shut it down before they even get a chance to explain what's going on.

PR #1289, opened 2015-12-21
VideoCore: Sync state after changing rasterizers

yuriks:
This fixes various bugs that appear in the HW rasterizer after switching
between it and the SW one during emulation.


Subv:
lg

Subv:
tm

lioncash:
sa

lioncash:
me

lioncash:
memes

PR #1256, opened 2015-11-26
Refactor ScanDirectoryTreeAndCallback to separate errors and retvals

archshift:
ScanDirectoryTreeAndCallback, before this change, coupled error/return codes and actual return values (number of entries found). This caused confusion and difficulty interpreting the precise way the function worked.

Supersedes, and closes #1255.

PR #1290, opened 2015-12-21
Add a return value in ForeachDirectoryEntry 

LFsWang:
I find there is a mistake in #1256. Line of return was deleted and not rewrite it.

### DIFF
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -427,6 +427,9 @@ bool ForeachDirectoryEntry(unsigned* num_entries_out, const std::string &directo
     // How many files + directories we found
     unsigned found_entries = 0;
 
+    // Save the status of callback function
+    bool callback_error = false;
+
 #ifdef _WIN32
     // Find the first file in the directory.
     WIN32_FIND_DATA ffd;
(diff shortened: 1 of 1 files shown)


src/common/file_util.cpp
```diff
@@ -427,6 +427,9 @@ bool ForeachDirectoryEntry(unsigned* num_entries_out, const std::string &directo
     // How many files + directories we found
     unsigned found_entries = 0;
 
+    // save the status of callback function
+    bool is_all_callback_successful = true;
```
archshift: Can you rename this to something more succinct, like `callback_error` or something, and adjust its usage as appropriate?

archshift:
Nice catch!

Pretty much LGTM, although could you get rid of those two merge commits by rebasing please?

LFsWang:
Ok. I renamed var and remove unnecessary merges

LFsWang:
AppVeyor is still pending..

kitling:
@LFsWang The AppVeyor build was canceled...? So it's never gonna complete. No idea why it was canceled.

Does anybody know of a way to get AppVeyor to re-check something?

Subv:
Force-push to the branch

LFsWang:
I pushed again. It works!

bunnei:
LGTM

PR #1299, opened 2015-12-27
DSP/Interrupts: Keep track of the registered interrupts

Subv:
This is still a hack, but it's a bit better than just ignoring the previous interrupt events. It should allow some games to not freeze randomly.

### DIFF
--- a/src/core/hle/service/dsp_dsp.cpp
+++ b/src/core/hle/service/dsp_dsp.cpp
@@ -8,14 +8,24 @@
 #include "core/hle/kernel/event.h"
 #include "core/hle/service/dsp_dsp.h"
 
+#include <unordered_map>
+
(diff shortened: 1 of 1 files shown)


neobrain:

2015-12-27T17:54:49Z: https://github.com/citra-emu/citra/compare/031443199bf5df2fc1f840793b36b975a61a5596...Subv:36a0d5c66c86a2c8dc55dfcb9b34b62d960ffd0e

Normmatt:
LGTM

yuriks:
Ah, so this is where that map comes from. @MerryMage, is this PR completely deprecated by #1386? If not, can you include any relevant changes in it?

MerryMage:
Updated #1386 to incorporate all changes.

Subv:
As of #1386 being merged, this is no longer valid

PR #1316, opened 2015-12-30
arm_dyncom_dec: Fix decoding of VMLS

lioncash:
Good 'ol "present since the import" Skyeye code.

Previously, all VMLS variants would misdecode as CDP (which isn't necessarily wrong in itself, however
VMLS has it's own label of execution).


bunnei:
LGTM

Issue #1150, opened 2015-09-11
Rollercoaster Tycoon 3D: Zfighting in HW renderer.


JMC47:
Happens in latest as of this post.  Look at the grass overpowering the sidewalk stuff on the hardware renderer, vs the clear lines on the SW renderer.  Note that lighting is messed up making it kind of hard to see.

Software Renderer

![zfightingsw](https://cloud.githubusercontent.com/assets/6598209/9828520/96f5f054-58bc-11e5-8b27-7407d53e043f.jpg)

Hardware Renderer

![zfightinghw](https://cloud.githubusercontent.com/assets/6598209/9828524/a3162e4e-58bc-11e5-8b2e-2417c919a0e5.jpg)

purpasmart96:
Zfighting can also  be seen in the OoT using the HW  renderer. but I've only see in one small place though.

Subv:
ALBW has the same issue, it's because we're not using viewport_depth_range nor viewport_depth_far_plane anywhere in the hardware renderer

JMC47:
I think this was merged.

PR #1334, opened 2016-01-03
hwrasterizer: Use depth offset

tfarley:
Uses offsets in the hardware renderer fragment shader to emulate depth near plane + polygon offsets. Should fix z-fighting as in #1150 including ALBW along with making the "inner" sub in the Sub Wars title screen consistent with the software renderer.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -126,6 +126,7 @@ void RasterizerOpenGL::InitObjects() {
 
 void RasterizerOpenGL::Reset() {
     SyncCullMode();
+    SyncDepthModifiers();
     SyncBlendEnabled();
     SyncBlendFuncs();
     SyncBlendColor();
@@ -192,6 +193,12 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
     // Culling
     case PICA_REG_INDEX(cull_mode):
         SyncCullMode();
+        break;
+
+    // Depth modifiers
+    case PICA_REG_INDEX(viewport_depth_range):
+    case PICA_REG_INDEX(viewport_depth_far_plane):
+        SyncDepthModifiers();
         break;
 
     // Blending
@@ -600,6 +607,15 @@ void RasterizerOpenGL::SyncCullMode() {
         UNIMPLEMENTED();
         break;
     }
+}
+
+void RasterizerOpenGL::SyncDepthModifiers() {
+    float depth_scale = -Pica::float24::FromRawFloat24(Pica::g_state.regs.viewport_depth_range).ToFloat32();
+    float depth_offset = Pica::float24::FromRawFloat24(Pica::g_state.regs.viewport_depth_far_plane).ToFloat32() / 2.0f;
+
+    // TODO: Implement scale modifier
+    uniform_block_data.data.depth_offset = depth_offset;
+    uniform_block_data.dirty = true;
 }
 
 void RasterizerOpenGL::SyncBlendEnabled() {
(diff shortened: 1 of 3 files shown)


bunnei:
Tested with ZALBW:

LGTM

Subv:
Closes #1345

neobrain:

2016-01-12T04:55:26Z: https://github.com/citra-emu/citra/compare/0ca8354464c04f5a6a236edb7cba5741006af511...tfarley:2c53b07c243b32025a8151254992292a4d002546

tfarley:
The internals of this PR have changed significantly, and should be squashed before merge. Would appreciate re-testing with ZALBW, Rune Factory 4 (#1345), and Rollercoaster Tycoon (#1150).

bunnei:
Retested ZALBW, looks fine (not, the above screenshot was with other PRs, this one is just this branch, hence the obvious differences)

bunnei:
Formal request for additional code reviews. I plan on having this one merged by the end of the week if I do not hear from anyone else.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -126,6 +126,8 @@ void RasterizerOpenGL::InitObjects() {
 
 void RasterizerOpenGL::Reset() {
     SyncCullMode();
+    SyncDepthScale();
+    SyncDepthOffset();
```
yuriks: IMO you don't need two separate functions for this

yuriks:
LGTM except for the comment. Have you checked if this affects performance much/at all?

src/video_core/renderer_opengl/gl_rasterizer.cpp
tfarley: I've been trying to keep these handlers 1:1 with registers as much as possible to avoid extraneous branches/checking of unmodified registers when called from NotifyPicaRegisterChanged(). But I'm not against merging them if that would make it significantly cleaner or more maintainable in the long run.

tfarley:
I checked ALBW on my machine with both integrated graphics and discrete GPU and haven't found any measurable performance difference in the profiling metrics or framerate before vs. after the changes.

src/video_core/renderer_opengl/gl_rasterizer.cpp
yuriks: I think it's unnecessary because, in practice, groups of registers will be modified together. In addition, we will be re-uploading the uniform block in the vast majority of draw calls anyway, so I think that even bothering to check if the value changed is overkill. If you remove the branch, then the functions become too trivial to have a reason to stay separate.

I don't think it's a *huge* code quality issue, but this proliferation of one-liners increases code size in both the cpp and the header to very little benefit.

Ideally you could add a profiler counter to check how many times per frame the uniform block is being uploaded when dirty, and then seeing if playing around with these approaches makes any appreciable difference.

yuriks: (This is all incidental to the PR though, so feel free to not do any of that.)

neobrain:

2016-01-16T03:39:23Z: https://github.com/citra-emu/citra/compare/0ca8354464c04f5a6a236edb7cba5741006af511...tfarley:55046c28751e12a194d5e976b55ad8a7834fefc6

yuriks:
LGTM

bunnei:
Please squash

bunnei:
Also, there was a build error with OS X in Travis CI

archshift:
clang segfaulted trying to link citra-qt o_O

neobrain:

2016-01-21T03:05:42Z: https://github.com/citra-emu/citra/compare/1b9012aea00efc4afa4c2e22acb335de9c31ed4f...tfarley:f53dbafdae6b6ca8ab718d2299437df469441389

PR #1363, opened 2016-01-19
DisassembleMemHalf: actually use width in determining opcode name

robturn:
Uhh, I'm like 80% sure that this change is correct? I'm just trying to get my foot in the door, I don't really know what I'm doing yet. Before this pull request, width was not being used at all, and since this function is used for different size opcodes, it wouldn't make sense to just always append h.


neobrain:

2016-01-19T17:53:15Z: https://github.com/citra-emu/citra/compare/2c663fbc3e59444c8f2bf2c111630a713aca8e2c...robturn:d29578d467a64af8a54210a091f641c543a9212c

PR #1367, opened 2016-01-24
Shader JIT: Fix off-by-one error when compiling JMPs

yuriks:
There was a mistake in the JMP code which meant that one instruction at
the destination would be skipped when the jump was taken. This commit
also changes the meaning of the culprit parameter to make it less
confusing and avoid similar mistakes in the future.


neobrain:

2016-01-24T10:16:42Z: https://github.com/citra-emu/citra/compare/0b6cc0592dec5584d1a2211e9d777f0b2640faea...yuriks:c1071c1ff7701e2e31dd8909a1cc58fee34c9db9

bunnei:
LGTM

PR #1369, opened 2016-01-25
Shader: Implement "invert condition" feature of IFU instruction

yuriks:
If the bit 0 of the JMPU instruction is set, then the jump condition
will be inverted. That is, a jump will happen when the boolean is false
instead of when it is true.

Fixes normals in lots of stuff for #1264


neobrain:

2016-01-25T04:31:12Z: https://github.com/citra-emu/citra/compare/0b6cc0592dec5584d1a2211e9d777f0b2640faea...yuriks:083d2d89a5ef0d5f61de60bd241e65f43289a381

bunnei:
LGTM. I tested this on a bunch of games.
Very obvious improvements with #1264:

VIRGINKLM:
Fixes Kingdom Hearts Dream Drop Distance rendering with JIT enabled. Good job!

PR #1371, opened 2016-01-25
archive_backend: Remove unnecessary const from return types

lioncash:
This doesn't return by reference so const isn't really necessary

### DIFF
--- a/src/core/file_sys/archive_backend.cpp
+++ b/src/core/file_sys/archive_backend.cpp
@@ -43,7 +43,7 @@ Path::Path(LowPathType type, u32 size, u32 pointer) : type(type) {
     }
 }
 
-const std::string Path::DebugStr() const {
+std::string Path::DebugStr() const {
     switch (GetType()) {
     case Invalid:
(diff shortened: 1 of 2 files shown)


neobrain:

2016-01-25T05:15:36Z: https://github.com/citra-emu/citra/compare/0b6cc0592dec5584d1a2211e9d777f0b2640faea...lioncash:3ed5ecd67ab36fb8de81d8da85245d3e1361dda8

yuriks:
I think the bug here is that it *isn't* returning by reference, isn't it? Seems wasteful to keep copying all those vectors around.

lioncash:
Returning a local object by value is undefined behavior (when used); considering vector has move constructors (and copy-elision can alternatively be done in many cases), I hardly believe it would be a concern (especially for something that puts a path into a vector).

yuriks:
Oh, derp, I misremembered what these methods did and didn't bother expanding the code, you're right. LGTM

archshift:
LGTM

PR #1374, opened 2016-01-25
RFC: Audio

MerryMage:
Implementation raw audio output. DSP effects not yet implemented.

First contribution to a large C++ project.

Here are a few things I would like to do before a merge:
- [ ] Proper cmake for OpenAL-Soft. [Binary package (Windows)](http://kcat.strangesoft.net/openal-binaries/openal-soft-1.17.2-bin.zip), [Source repo (Linux / OS X)](https://github.com/kcat/openal-soft/tree/openal-soft-1.17.1).
- [ ] Implement audio mixing (stereo only).

Things I would like to have comments on:
- Is everything in the right place?
- Is OpenAL-Soft acceptable?
- Should multple possible audio outputs be possible in the future? (NullAudio?)
- Should there be a config option to disable audio?

Some things I haven't done and might do in the future (note to self):
- Wav dumper.
- Figure out why some audio clips don't play.
- Implement IIR filter.
- Implement aux and feedback.
- Implement full audio mixing.
- Implement rate.
- Figure out everything else.


neobrain:

2016-01-25T13:37:54Z: https://github.com/citra-emu/citra/compare/8b3994e9e4629423960f9ba22e5df01afd5b724b...MerryMage:444cf8b1ae16b360e21e5b363e5dfdfdd5ccb630

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
+
+#include "AL/al.h"
+#include "AL/alc.h"
+#include "AL/alext.h"
```
lioncash: OpenAL includes should have angle brackets.

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+        int coef2 = adpcm_coeff[idx * 2 + 1];
+
+        auto process_nybble = [&](int nybble) -> s16 {
+            int xn = nybble * scale;
+            // We first transform everything into 11 bit fixed point, perform the second order digital filter, then transform back.
+            // 0x400 == 0.5 in 11 bit fixed point.
+            // Filter: y[n] = x[n] + 0.5 + c1 * y[n-1] + c2 * y[n-2]
+            int val = ((xn << 11) + 0x400 + coef1 * yn1 + coef2 * yn2) >> 11;
+            // Clamp to output range.
+            if (val >= 32767) val = 32767;
```
lioncash: math_util.h in common has a clamp function.

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+    std::priority_queue<Buffer> queue; ///< Things we have gotten from userland we haven't queued onto `source` yet.
+    std::queue<Buffer> playing;        ///< Things we have queued onto `source`.
+    u16 last_bufid;                    ///< Userland wants us to report back what was the thing we last played.
+
+    // For ADPCM decoding use.
+    std::array<s16, 16> adpcm_coeffs;
+    AdpcmState adpcm_state;
+};
+
+OutputChannel chans[24];
```
lioncash: static, also std::array

src/core/audio/audio.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "AL/al.h"
+#include "AL/alc.h"
+#include "AL/alext.h"
```
lioncash: These don't need to be here. Nothing in the source file relies on them.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+static constexpr VAddr DspAddrToVAddr(VAddr base, DspRegion dsp_addr) {
+    return (VAddr(dsp_addr) << 1) + base;
+}
+
+/**
+ * dsp_u32:
+ *     Care must be taken when reading/writing 32-bit values as the words are not in the expected order.
+ */
+struct dsp_u32 {
+    operator u32() {
```
lioncash: this can be const.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+        return Convert(storage);
+    }
+    void operator=(u32 newvalue) {
```
lioncash: new_value

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+};
+ASSERT_STRUCT(ChannelStatus, 12);
+
+struct AdpcmCoefficients {
+    s16 coeff[16];
+};
+ASSERT_STRUCT(AdpcmCoefficients, 32);
+
+template <typename T>
+static inline bool TestAndUnsetBit(T& value, size_t bitno) {
```
lioncash: bit_number

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+        } else if (id0 - id1 > 10000 && id1 < 10) {
+            current_base = BASE_ADDR_1;
+        } else if (id1 > id0) {
+            current_base = BASE_ADDR_1;
+        } else {
+            current_base = BASE_ADDR_0;
+        }
+    }
+
+    auto channel_contexes = (ChannelContext*) Memory::GetPointer(DspAddrToVAddr(current_base, DSPADDR1));
```
lioncash: reinterpret_cast for pointer types.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
-void SignalInterrupt() {
-    // TODO(bunnei): This is just a stub, it does not do anything other than signal to the emulated
-    // application that a DSP interrupt occurred, without specifying which one. Since we do not
-    // emulate the DSP yet (and how it works is largely unknown), this is a work around to get games
-    // that check the DSP interrupt signal event to run. We should figure out the different types of
-    // DSP interrupts, and trigger them at the appropriate times.
+static void AudioTick(u64, int cycles_late) {
+    VAddr current_base;
+
+    {
```
lioncash: This whole scope should be its own function, `GetBaseAddress` or something.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+        std::tie(playing, status0.current_buffer_id, status0.buffer_position) = Audio::GetStatus(chanid);
+        if (playing) {
+            status0.is_playing |= 1;
+        } else {
+            status0.is_playing = 0;
+        }
+        status1 = status0;
+    }
+
+    for (auto interrupt_event : interrupt_events)
```
lioncash: auto&

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+        }
 
-    if (interrupt_event != 0)
-        interrupt_event->Signal();
+        // TODO: Detect any change to the structures without a dirty flag update to identify what the other bits do.
+
+        Audio::Tick(chanid);
+
+        // Update channel status
+        bool playing = false;
```
lioncash: Should be its own function, `UpdateChannelStatus`

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+    DSPADDR14 = 0xAC5C
+};
+
+static constexpr VAddr DspAddrToVAddr(VAddr base, DspRegion dsp_addr) {
+    return (VAddr(dsp_addr) << 1) + base;
+}
+
+/**
+ * dsp_u32:
+ *     Care must be taken when reading/writing 32-bit values as the words are not in the expected order.
```
neobrain: What order (assuming you mean byte order)? What's the expected order? What order are they actually in?

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+    for (int frameno = 0; frameno < NUM_FRAMES; frameno++) {
+        int frame_header = data[frameno * FRAME_LEN];
+        int scale = 1 << (frame_header & 0xF);
+        int idx = (frame_header >> 4) & 0x7;
+
+        // Coefficients are fixed point with 11 bits fractional part.
+        int coef1 = adpcm_coeff[idx * 2 + 0];
+        int coef2 = adpcm_coeff[idx * 2 + 1];
+
+        auto process_nybble = [&](int nybble) -> s16 {
```
neobrain: What does this do on a high level? needs a comment

neobrain:
Added some comments. Please do not only reply to the questions on github, but also address them by adding comments in the source code.

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+    // For ADPCM decoding use.
+    std::array<s16, 16> adpcm_coeffs;
+    AdpcmState adpcm_state;
+};
+
+OutputChannel chans[24];
+
+int InitAL() {
+    ALCdevice *device = alcOpenDevice(nullptr);
```
lioncash: You may want to consider defining custom unique_ptr aliases for OpenAL to make resource management a little more automated.
```cpp
using ALCDevicePointer  = std::unique_ptr<ALCdevice, decltype(&alcCloseDevice)>;
using ALCContextPointer = std::unique_ptr<ALCcontext, decltype(&alcDestroyContext)>;

e.g.

auto ctx = ALCContextPointer(alcCreateContext(device, nullptr), alcDestroyContext);
```

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+        std::vector<s16> decoded = DecodeADPCM((u8*)data, sample_count, chans[chanid].adpcm_coeffs, chans[chanid].adpcm_state);
+        alBufferData(b, AL_FORMAT_STEREO16, decoded.data(), decoded.size() * 2, BASE_SAMPLE_RATE);
+
+        if (alGetError() != AL_NO_ERROR) goto do_silence;
+
+        break;
+    }
+    default:
+        LOG_ERROR(Audio, "Channel %i: Buffer %i: Unrecognised audio format (size: %i samples)", chanid, buffer_id, sample_count);
+    do_silence:
```
neobrain: C++ has exceptions for this stuff, yo.

src/core/audio/audio.cpp
neobrain: +1

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+    }
+
+    auto channel_contexes = (ChannelContext*) Memory::GetPointer(DspAddrToVAddr(current_base, DSPADDR1));
+    auto channel_contex0 = (ChannelContext*)Memory::GetPointer(DspAddrToVAddr(BASE_ADDR_0, DSPADDR1));
+    auto channel_contex1 = (ChannelContext*)Memory::GetPointer(DspAddrToVAddr(BASE_ADDR_1, DSPADDR1));
+    auto channel_status0 = (ChannelStatus*)Memory::GetPointer(DspAddrToVAddr(BASE_ADDR_0, DSPADDR2));
+    auto channel_status1 = (ChannelStatus*)Memory::GetPointer(DspAddrToVAddr(BASE_ADDR_1, DSPADDR2));
+    auto channel_adpcm_coeffs = (AdpcmCoefficients*) Memory::GetPointer(DspAddrToVAddr(current_base, DSPADDR3));
+
+    for (int chanid=0; chanid<NUM_CHANNELS; chanid++) {
```
lioncash: `for (int chanid = 0; chanid < NUM_CHANNELS; chanid++)`

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+    LOG_INFO(Audio, "Device Frequency: %i", dev_rate);
+
+    for (int i = 0; i < 24; i++) {
+        alGenSources(1, &chans[i].source);
+        if (alGetError() != AL_NO_ERROR) {
+            LOG_CRITICAL(Audio, "Channel %i: Failed to setup sound source", i);
+        }
+    }
+
+    silence.fill(0);
```
neobrain: You should just initialize the global variable to 0 and make it const instead of filling it manually here. cc @lioncash who probably knows the best way to do this (chances are it's already done by default, or just requires {}).

src/core/audio/audio.cpp
hrydgard: It takes one nybble of data and runs it through the ADPCM filter, which is pretty obvious from the context IMHO...

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+        return Convert(storage);
+    }
+    void operator=(u32 newvalue) {
+        storage = Convert(newvalue);
+    }
+private:
+    static constexpr u32 Convert(u32 value) {
+        return ((value & 0x0000FFFF) << 16) | ((value & 0xFFFF0000) >> 16);
+    }
+    u32 storage;
```
lioncash: this should likely have an initial value like zero or something so it has a valid initial state.

src/core/hle/service/dsp_dsp.cpp
lioncash: inline isn't really necessary for TU-local functions marked static.

src/core/audio/audio.cpp
MerryMage: Will rename to decode_sample and add comment: // Decodes an audio sample. One nybble produces one s16 sample.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+    auto channel_status0 = (ChannelStatus*)Memory::GetPointer(DspAddrToVAddr(BASE_ADDR_0, DSPADDR2));
+    auto channel_status1 = (ChannelStatus*)Memory::GetPointer(DspAddrToVAddr(BASE_ADDR_1, DSPADDR2));
+    auto channel_adpcm_coeffs = (AdpcmCoefficients*) Memory::GetPointer(DspAddrToVAddr(current_base, DSPADDR3));
+
+    for (int chanid=0; chanid<NUM_CHANNELS; chanid++) {
+        ChannelContext& ctx = channel_contexes[chanid];
+        ChannelStatus& status0 = channel_status0[chanid];
+        ChannelStatus& status1 = channel_status1[chanid];
+
+        if (ctx.dirty) {
```
lioncash: The contents in this if statement should likely be its own function, or at the very least, have the handling for the individual bits in their own respective function. As more stuff gets implemented, this function will become very large.

src/core/hle/service/dsp_dsp.cpp
MerryMage: Byte order is middle endian from the ARM11's point of view which surprised me, but would make sense if the dsp is big endian with 16-bit words.

src/core/hle/service/dsp_dsp.cpp
neobrain: This is not valid C++ code because it violates strict aliasing rules. Do not abuse pointer casting for serialization.

src/core/hle/service/dsp_dsp.cpp
neobrain: All usages of this functions used it with constexpr values of bitno. Hence it should instead just take a BitField argument.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -271,6 +575,46 @@ static void GetHeadphoneStatus(Service::Interface* self) {
...
+*  Inputs:
+*      1 : Register Number
+*  Outputs:
+*      1 : Result of function, 0 on success, otherwise error code
+*      2 : Value in the register
+*/
+static void RecvData(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 registerNo = cmd_buff[1];
```
lioncash: register_number

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -271,6 +575,46 @@ static void GetHeadphoneStatus(Service::Interface* self) {
...
+*  Outputs:
+*      1 : Result of function, 0 on success, otherwise error code
+*      2 : non-zero == ready
+*  Notes:
+*      Seems to be mainly called when going into sleep mode.
+*/
+static void RecvDataIsReady(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 registerNo = cmd_buff[1];
```
lioncash: register_number

src/core/hle/service/dsp_dsp.cpp
lioncash: Suggestion: have another specialization where the bit number is a template argument so it can be used with constant bit positions to test and unset. so you can static_assert at compile-time if a bit number larger than the underlying type is specified. This is also particularly useful for catching unintended shifts into the sign-bit for signed types. Just gives a little compile-time guarding against small but subtle mistakes. That, or provide a regular assert, but that might be a little inferior, since you wait until runtime for it to blow up instead.

src/core/hle/service/dsp_dsp.cpp
MerryMage: @neobrain Ah; I was unaware of that, thanks. However I am uncertain of what else to do. `memcpy`? Would've been nice if GetPointer returned a char* then strict aliasing wouldn't apply.

I suppose that having MMIO or something implemented would be ideal.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,309 @@
...
+ASSERT_STRUCT(ChannelStatus, 12);
+
+struct AdpcmCoefficients {
+    s16 coeff[16];
+};
+ASSERT_STRUCT(AdpcmCoefficients, 32);
+
+template <typename T>
+static inline bool TestAndUnsetBit(T& value, size_t bitno) {
+    T mask = 1 << bitno;
```
lioncash: This should likely be:

```cpp
T mask = static_cast<T>(1) << bitno;
```

to prevent signed overflow if the sign-bit position is ever checked

hrydgard:
Curious - is this DSP performing all the mixing on the real hardware, or does it have hardwired channel mixing and then just using the DSP instruction set for effects?

If the former, this code should be considered a HLE DSP implementation, and maybe named as such :)

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+    s16 yn2; ///< y[n-2]
+};
+
+// GC-ADPCM with scale factor and variable coefficients.
+// Frames are 8 bytes long containing 14 samples each.
+// Samples are 4 bits (one nybble) long.
+std::vector<s16> DecodeADPCM(const u8 * const data, const size_t sample_count, const std::array<s16, 16>& adpcm_coeff, AdpcmState& state) {
+    const size_t FRAME_LEN = 8;
+    const size_t SAMPLES_PER_FRAME = 14;
+    const static int SIGNED_NYBBLES[16] = { 0,1,2,3,4,5,6,7,-8,-7,-6,-5,-4,-3,-2,-1 };
```
lioncash: These can all be constexpr, rather than const.

MerryMage:
This code should probably be named a HLE DSP implementation regardless; the DSP itself handles the buffer management this code performs (I think. I've been mainly poking things from userland.). On actual hardware, most communication with the DSP from userland is via a shared memory region.

src/core/hle/service/dsp_dsp.cpp
lioncash: @MerryMage It would still violate strict aliasing if the struct/entity is larger than char's bitsize, since it increases the alignment requirement.

But yes, memcpy would resolve this. It would probably be beneficial to define a template function in the Memory namespace called ExtractFromMemory, like

```cpp
template <typename T>
T ExtractFromMemory(u32 address)
{
    static_assert(std::is_trivially_copyable<T>::value, "Type must be trivially copyable");
    // Add other static_asserts here

    T entity;
    std::memcpy(&entity, GetPointer(address), sizeof(T));
    return entity;
}
```

It can alternatively be changed to use an out parameter and return a bool for memory validity checking (or some form of a pseudo std::optional can be used, which I think we have in common)

src/core/audio/audio.cpp
lioncash: Yeah, 

```cpp
std::array<u8, 10000> silence{};
```
would be fine.

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+    if (!c.playing.empty()) {
+        c.last_bufid = c.playing.front().id;
+    }
+}
+
+std::tuple<bool, u16, u32> GetStatus(int chanid) {
+    auto& c = chans[chanid];
+
+    bool isplaying = c.enabled;
```
lioncash: is_playing

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+    u16 bufid = c.last_bufid;
```
lioncash: buffer_id

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,339 @@
...
+    u32 pos;
```
lioncash: This can be joined with its assignment.

src/core/audio/audio.h
```diff
@@ -0,0 +1,37 @@
...
+
+void EnqueueBuffer(int chanid, u16 buffer_id, void* data, int sample_count, bool is_looping);
+
+void Tick(int chanid);
+
+// Return values:
+// <1>: is_enabled
+// <2>: prev buffer_id
+// <3>: current sample position in current buffer
+std::tuple<bool, u16, u32> GetStatus(int chanid);
```
lioncash: channel_id

src/core/audio/audio.h
```diff
@@ -0,0 +1,37 @@
...
+};
+
+void UpdateFormat(int chanid, int mono_or_stereo, Format format);
+void UpdateAdpcm(int chanid, s16 coeffs[16]);
+
+void Play(int chanid, bool play);
+
+void EnqueueBuffer(int chanid, u16 buffer_id, void* data, int sample_count, bool is_looping);
+
+void Tick(int chanid);
```
lioncash: here and above should use `channel_id`

src/core/audio/audio.h
```diff
@@ -0,0 +1,37 @@
...
+#include "common/common_types.h"
+
+#include <tuple>
+
+namespace Audio {
+
+void Init();
+void Shutdown();
+
+enum Format : u16 {
```
lioncash: `enum class Format`

src/core/hle/service/dsp_dsp.cpp
neobrain: Fwiw, if you want to keep the spirit of this function without changing to BitField, you can also use
```
template <typename T, size_t bitno>
static inline bool TestAndUnsetBit(T& value) {
    auto& field = reinterpret_cast<BitField<bitno, 1, T>&>(value).Value();
    bool ret = field;
    field = 0;
    return ret;
```
.. this will perform the assertions suggested by @lioncash while at least not NIHing them.

Do note however that I would personally prefer taking a BitField member directly, since doing so requires introducing a named field, which provides for more self-documenting code.

MerryMage: I think I'll go with that for now. Might switch to `BitField`s when I understand all the flags better later.

src/core/hle/service/dsp_dsp.cpp
MerryMage: Unfortunately couldn't find an `optional<T>` type in common. Going with `bool ExtractFromMemory(u32 address, T& object)`, and also implementing `InjectIntoMemory` with similar function signature.

neobrain: @MerryMage Feel free to use boost's optional.hpp.

yuriks: I wouldn't blindly use boost.optional without inspecting optimization output first. I suspect it's sufficient to make the function opaque for the compile and downgrade the memcpy from "cheap way to tell the compiler to do safe loads" to "actual memcpy". Though the same may apply to the extra function.

neobrain: @yuriks How does optimization output matter here? What evil things could optional do here?

yuriks: If this function is to be a generic utility and used everywhere we load structs from memory, then it should be at least reasonably efficient, otherwise we wouldn't be able to use it in any hot loops.

yuriks: Also, boost.optional can be fairly slow if you're not careful to make sure it's detecting your operations as noexcept, otherwise it'll allocate memory every time you assign to it.

MerryMage: Alright, T& + bool it is then.

Also as a side-note I can't static_assert for std::is_trivially_copyable<T> because BitField is not trivially copyable, so I won't.

lioncash: Crude comparison using the optional provided with GCC 5.3.0; I'd imagine it's similar: http://pastebin.com/Be1FnH4V

Paste that into http://gcc.godbolt.org, with the flags `--std=c++17 -O2`

IMO, I'd much prefer optional being used, considering it enforces validation better, and doesn't require an out parameter.

yuriks: You didn't write any code that actually used it the return values. This should always be inlined. http://goo.gl/G45GRO

GCC + std::optional does pretty well, and I'd totally go for std::optional if we could, but I highly doubt boost::optional will fare as well as it did.

lioncash: > but I highly doubt

I'd really prefer if we didn't throw discussions out because of unconfirmed preconceived notions. Especially ones about performance. As far as it is, I disagree with not using optional unless there's a valid test showing it generates suboptimal code to the extent it warrants using alternatives.

yuriks: I can't run a proper test with the real boost right now since I'm at work, but I'll do that once I'm home.

MerryMage: Compiled @lioncash's test w/ boost::optional: https://gist.github.com/MerryMage/f62e00d5bebff1104596

Output is similar to std::optional.

yuriks: Can you compile with the two user function I added in my link?

MerryMage: https://gist.github.com/MerryMage/e548b357bcdfb45e66e6#file-a-s-L76

yuriks: Seems to be doing fine with that class then. Knock yourself out. Thanks for running the tests, too.

MerryMage: np, thanks!~

neobrain:

2016-01-26T00:04:44Z: https://github.com/citra-emu/citra/compare/8b3994e9e4629423960f9ba22e5df01afd5b724b...MerryMage:3e75de6006e4ff1a70b1a0e187d81089fe7f0cdf

src/core/memory.h
```diff
@@ -131,6 +132,39 @@ void WriteBlock(VAddr addr, const u8* data, size_t size);
 u8* GetPointer(VAddr virtual_address);
 
 /**
+* Extracts a POD type from memory. Returns false if address is invalid.
```
lioncash: This returns an optional with no set value on failure.

MerryMage: Derp. Sorry.

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+    ALint state, samples;
+    alGetSourcei(c.source, AL_SOURCE_STATE, &state);
+    alGetSourcei(c.source, AL_SAMPLE_OFFSET, &samples);
+    ret.sample_position = samples;
+
+    return ret;
+}
+
+};
```
purpasmart96: No newline at end of file.

src/core/audio/audio.h
```diff
@@ -0,0 +1,40 @@
...
+struct ChannelStatus {
+    bool is_enabled;
+    bool was_fed_data;   ///< Have we been fed data since being enabled?
+    u16 most_recent_buffer_id;
+    u32 sample_position; ///< Play position in current buffer
+};
+
+ChannelStatus GetStatus(int channel_id);
+
+};
```
purpasmart96: No newline at end of file.

neobrain:

2016-01-26T12:45:47Z: https://github.com/citra-emu/citra/compare/8b3994e9e4629423960f9ba22e5df01afd5b724b...MerryMage:37658584bbb388daee4b3c3223e8c113e1342f01

neobrain:

2016-01-26T12:53:32Z: https://github.com/citra-emu/citra/compare/8b3994e9e4629423960f9ba22e5df01afd5b724b...MerryMage:726c7d93c15336690fe7e32a7a7a66b861176aa7

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,343 @@
...
+#include <AL/al.h>
+#include <AL/alc.h>
+#include <AL/alext.h>
+
+#include "common/logging/log.h"
+#include "common/math_util.h"
+
+#include "core/audio/audio.h"
+
+#include <algorithm>
```
lioncash: Forgot to mention this on my first pass,(sorry!), but standard library includes should come before all other includes. Usually it's in the order:

- Standard library
- Other external libraries
- Citra headers

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,343 @@
...
+int InitAL() {
+    device = ALCDevicePointer(alcOpenDevice(nullptr), &alcCloseDevice);
+    if (!device) {
+        LOG_CRITICAL(Audio, "Could not open a device!");
+        return 1;
+    }
+
+    context = ALCContextPointer(alcCreateContext(device.get(), nullptr), &alcDestroyContext);
+    if (context == nullptr || alcMakeContextCurrent(context.get()) == ALC_FALSE) {
+        if (context != nullptr) {
```
lioncash: This can just be

```cpp
context.reset();
device.reset();
```

if order of deallocation is important (reset doesn't call the unique_ptr's deleter if it contains nullptr), otherwise they can just be removed entirely

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,344 @@
...
 
+#include <unordered_map>
+
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Namespace DSP_DSP
 
 namespace DSP_DSP {
 
+struct PairHash {
+public:
```
lioncash: structs are already public by default, so this doesn't need to be specified

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,343 @@
...
+void UpdateAdpcm(int channel_id, s16 coeffs[16]) {
+    LOG_DEBUG(Audio, "Channel %i: ADPCM Coeffs updated", channel_id);
+    std::copy(coeffs, coeffs+16, std::begin(chans[channel_id].adpcm_coeffs));
+}
+
+void EnqueueBuffer(int channel_id, u16 buffer_id, void* data, int sample_count, bool is_looping) {
+    LOG_DEBUG(Audio, "Channel %i: Buffer %i: Enqueued (size %i)", channel_id, buffer_id, sample_count);
+
+    if (is_looping) {
+        LOG_WARNING(Audio, "Channel %i: Buffer %i: Looped buffers are unimplemented", channel_id, buffer_id);
```
lioncash: the formatter for buffer_id should be %u

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+        if (alGetError() != AL_NO_ERROR) goto do_silence;
+
+        break;
+    case Format::ADPCM: {
+        if (c.mono_or_stereo != 1) {
+            LOG_ERROR(Audio, "Channel %i: Buffer %i: Being fed non-mono ADPCM (size: %i samples)", channel_id, buffer_id, sample_count);
+        }
+
+        std::vector<s16> decoded = DecodeADPCM((u8*)data, sample_count, c.adpcm_coeffs, c.adpcm_state);
```
lioncash: static_cast

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+void Tick(int channel_id) {
+    auto& c = chans[channel_id];
+
+    if (!c.queue.empty()) {
+        while (!c.queue.empty()) {
+            alSourceQueueBuffers(c.source, 1, &c.queue.top().buffer);
+            if (alGetError() != AL_NO_ERROR) {
+                alDeleteBuffers(1, &c.queue.top().buffer);
+                LOG_CRITICAL(Audio, "Channel %i: Buffer %i: Failed to enqueue : %s", channel_id, c.queue.top().id, alGetString(alGetError()));
```
lioncash: %u for buffer

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,343 @@
...
+            ALint state;
+            alGetSourcei(c.source, AL_SOURCE_STATE, &state);
+            if (state != AL_PLAYING) {
+                alSourcePlay(c.source);
+            }
+        }
+    }
+
+    if (chans[channel_id].playing.size() > 10) {
+        LOG_ERROR(Audio, "Channel %i: We have far far too many buffers enqueued (%i of them)", channel_id, chans[channel_id].playing.size());
```
lioncash: %zu for `chans[channel_id].playing.size()`'s formatter

src/core/audio/audio.cpp
```diff
@@ -0,0 +1,343 @@
...
+    while (processed > 0) {
+        ALuint buf;
+        alSourceUnqueueBuffers(c.source, 1, &buf);
+        processed--;
+
+        if (!c.playing.empty()) {
+            if (c.playing.front().buffer != buf) {
+                LOG_CRITICAL(Audio, "Channel %i: Play queue desynced with OpenAL queue. (buf???)", channel_id);
+            } else {
+                LOG_DEBUG(Audio, "Channel %i: Buffer %i: Finished playing", channel_id, c.playing.front().id);
```
lioncash: %u for buffer

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,344 @@
...
+
+        // Embedded Buffer Changed
+        Audio::EnqueueBuffer(channel_id, ctx.buffer_id, Memory::GetPhysicalPointer(ctx.physical_address), ctx.sample_count, ctx.is_looping);
+    }
+
+    if (TestAndUnsetBit<19>(ctx.dirty)) {
+        // Buffer queue
+        for (int i = 0; i < 4; i++) {
+            if (ctx.buffers_dirty & (1 << i)) {
+                auto& b = ctx.buffers[i];
```
lioncash:
```cpp
const auto& buffer
```
since it's not being directly modified.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,344 @@
...
+                Audio::EnqueueBuffer(channel_id, b.buffer_id, Memory::GetPhysicalPointer(b.physical_address), b.sample_count, b.is_looping);
+            }
+        }
+
+        if (ctx.buffers_dirty & ~(u32)0xF) {
```
lioncash:
```cpp
if (ctx.buffers_dirty & ~0xFUL)
```
or
```cpp
if (ctx.buffers_dirty & ~UINT32_C(0xF))
```

RavenHome1:
sorry for the inconvenient and out of place question but will this finally add sound emulation?

bunnei:
Awesome progress! I tested this with many games, and to my surprise many actually played audio! (e.g. Super Monkey Ball, Ocarina of Time, Super Mario 3D Land, Pokemon X, and others). I’d love to see this move forward. I haven’t reviewed the code to much yet, but some general comments on my experience thus far:
* It didn’t build right out of the box (as evidenced by all of the build bots failing). I’m sure you know that, though – I see you’ve approached the dependency the same way we use GLFW, that seems reasonable to me.
* On a similar note, we were considering switching from GLFW to SDL 2.x at one time for our CLI render window and potential joypad input. Would we be able to sufficiently emulate 3DS audio with SDL? If so, this might be a good time to make that switch to reduce our dependencies.
* My experience testing this ranged from barely intelligible sounds (e.g. Zelda: Ocarina of Time), where maybe a single musical note would repeat on loop with various other popping/glitching noises, to full audio streams playing (e.g. Super Monkey Ball), but even then there would be lots of cutting out, synchronization issues, and glitches. Can you elaborate on why some of these issues are happening? Is it bugs (or hacks) in your implementation? Unimplemented features? Speed/timing issues with the core emulator?
* You should add a configuration option (and in the Qt GUI) to enable/disable this feature. As a partial implementation, I suspect this will remain “experimental” and work-in-progress for at least a few cycles of PRs.
...Expect more comments when I actually review this, which will hopefully be tonight.

hrydgard:
I would recommend rewriting to not use alBuffers for each channel of the 3DS, and instead writing your own custom mixer, mixing to a single continuous output stream. This is what pretty much all other emulators do, including Dolphin's HLE audio. That way you could run on any backend (SDL2, WASAPI, whatever) that lets you provide a plain audio stream, plus you get more control. It seems this is in your plans though?

lioncash:
It may also be beneficial to set up (after this is merged, or before, whichever is convenient) an interface for implementing sound backend support for different APIs. It'd keep the core code more library-agnostic

MerryMage:
@hrydgard Planning to do that; I would have to do that anyway in order to properly implement the loopback buffer (currently unimplemented).

@lioncash Yes! I also feel like audio.h/.cpp shouldn't /really/ be in core since it has library dependencies. Perhaps an audio_core?

I mainly chose OpenAL for it's cross-platformness so it would make a reasonable first implementation before having system-specific libraries. Once mixing is implemented all it'll need is something to push bytes through much like other emulators.

Thanks tremenously for your help reviewing by the way!!

@bunnei 
> didn’t build right out of the box 
Indeed, would be nice to pop that zip file along with the other bundles perhaps, might push something that should fix the buildbot builds this evening. (Would like that to be checked since I'm unfamiliar with cmake.)
>  Would we be able to sufficiently emulate 3DS audio with SDL?
Yes.
> Can you elaborate on why some of these issues are happening? 
> Is it bugs (or hacks) in your implementation? 
> Unimplemented features? 
* Audio speeds and interpolation: Unimplemented.
* IIR filter: Used to implement some effects. Unimplemented. A creative developer might use this for audio obfuscation but I doubt this.
* Looped buffers: Unimplemented.
* ChannelContext::is_playing: A hack. I haven't managed to figure out exacrly when to turn certain bits off. Could explain audio loops and pops.
* Padding in the middle of Buffer: Some of the variables here are involved in decoding ADPCM *I think* which could explain some strange audio. Meaning TBC. Could result in distorted audio and pops.
* Stopping audio when channel is disabled: Unimplemented. Could explain audio loops.
* I've also just made the assumption that the DSP firmware is the same across all games. Perhaps a hash-check should be done on the firmware and at least print an error to the console.
> As a partial implementation, I suspect this will remain “experimental” and work-in-progress for at least a few cycles of PRs.
I agree. Thinking about this I should implement that as part of this PR.

> Speed/timing issues with the core emulator?
As far as I can tell the primary interrupt occurs very consistently at DSP_DSP::frame_time cycles.

I haven't tested this on much more than just FE:A at the moment. Even there the audio isn't perfect (mainly missing minor sound effects in the menus and occasional pops). As for my RE work I've mainly worked out mostly what regions 0-5 do, but that's still rather patchy. Only the first two are semi-implemented. Most of the other regions don't seem to be used for much, though region 9 seems to be used for debugging(?). Maybe.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -42,9 +356,7 @@ static void ConvertProcessAddressFromDspDram(Service::Interface* self) {
     u32 addr = cmd_buff[1];
 
     cmd_buff[1] = 0; // No error
-    cmd_buff[2] = (addr << 1) + (Memory::DSP_RAM_VADDR + 0x40000);
-
-    LOG_WARNING(Service_DSP, "(STUBBED) called with address 0x%08X", addr);
+    cmd_buff[2] = DspAddrToVAddr(BASE_ADDR_0, (DspRegion)addr);
```
lioncash: static_cast

lioncash:
> Perhaps an audio_core?

Agreed, separating hard dependencies off core is a good idea.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -2,30 +2,344 @@
...
+}
+
+static void ReadChannelContext(VAddr current_base, int channel_id) {
+    ChannelContext ctx = GetChannelContext(current_base, channel_id);
+
+    if (!ctx.dirty) {
+        return;
+    }
+
+    if (TestAndUnsetBit<29>(ctx.dirty)) {
```
neobrain: Maybe I missed an explanation before, but... why not `TestAndUnsetBitField(ctx.dirty.first_time_init)`, with `first_time_init` of type `BitField<29, 1, x>` ?

neobrain:
Just my 2c, review on this would be a *lot* easier if the PR focussed on getting the hardware emulation set in stone without bothering about making it output sound yet. Currently, discussion seems to be focused on the comparatively unimportant topics of the audio API choice and of what's missing, rather than actually figuring out whether the foundation is "good enough for now" or not.

Having frontend code outputting sound is good and all, but it should at best be considered a prototype illustrating that the base code works reasonably well. Including it for review is just drawing away attention from important stuff (especially since we'll want to have multiple audio frontends in the future).

hrydgard:
@MerryMage Great! And really nice work. This feels a lot like my initial async mostly-guesswork-based AX emulation for the Gamecube :)

I agree with neobrain that the important part is the DSP communication stuff and the fundamental ucode emulation (might as well call the DSP program(s) ucodes like on the GC because it seems to be the same thing, just a different DSP processor) - but, even a rough improvised mixing solution like what you got here is important to "feel" things out in the beginning, so it's not necessarily a bad idea to include it from the start, with the perspective that it will be replaced with something more accurate soon.

hrydgard:
Oh yeah, like I mentioned above, when you do your own mixing you will find that the "slowing the audio down" becomes very easy - simply set up a periodic task to mix the right amount of audio samples for the emulated time passed between the task invocations, and have that mixer enqueue the mixed samples for an audio playback thread to pick up - and you'll automatically get the correct amount of annoying audio skipping when the emulation can't keep up :)

Subv:
I wonder if this fixes the freezes with Pokemon X

MerryMage:
Logging decisions made on IRC:

Implement this as a series of atomic PRs that build on top of each other instead of one huge PR. Use much of the code here as is possible that would fit with that.

MerryMage:
Closed as #1386 and subsequent PRs will reimplement functionality in this PR.

To-do list for reference: https://gist.github.com/MerryMage/9381b330b59cbb23feda

CMakeLists.txt
```diff
@@ -159,6 +161,26 @@ if (ENABLE_GLFW)
     endif()
 endif()
 
+if (CITRA_USE_BUNDLED_OPENAL)
+    if (WIN32)
+        set(OPENAL_VER "openal-soft-1.17.1-bin")
+    else()
+        message(FATAL_ERROR "No bundled OpenAL binaries for your toolchain. Disable CITRA_USER_BUNDLED_OPENAL and provide your own.")
```
byteandahalf: CITRA_USE_BUNDLED_OPENAL?

PR #1377, opened 2016-01-26
Memory: Implemented MMIO

MerryMage:
Intention is for devices that require memory mapped I/O to implement the `MMIORegion` interface which would handle the reads/writes for that memory region.

Intention is for `MapMMIO` calls to go into `InitLegacyAddressSpace`.

Ping @yuriks as I seemed to be implementing something that was on his TODO list.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -13,6 +13,7 @@
 #include "core/hle/kernel/process.h"
 #include "core/memory.h"
 #include "core/memory_setup.h"
+#include "core/mmio.h"
 
 namespace Memory {
 
@@ -23,6 +24,12 @@ enum class PageType {
     Memory,
     /// Page is mapped to a I/O region. Writing and reading to this page is handled by functions.
     Special,
+};
+
+struct SpecialRegion {
+    VAddr base;
+    u32 size;
+    MMIORegionPointer handler;
 };
 
 /**
@@ -41,8 +48,13 @@ struct PageTable {
     std::array<u8*, NUM_ENTRIES> pointers;
 
     /**
+     * Contains MMIO handlers that back memory regions whose entries in the `attribute` array is of type `Special`.
+     */
+    std::vector<SpecialRegion> special_regions;
+
+    /**
      * Array of fine grained page attributes. If it is set to any value other than `Memory`, then
-     * the corresponding entry in `pointer` MUST be set to null.
+     * the corresponding entry in `pointers` MUST be set to null.
      */
     std::array<PageType, NUM_ENTRIES> attributes;
 };
@@ -80,10 +92,12 @@ void MapMemoryRegion(VAddr base, u32 size, u8* target) {
     MapPages(base / PAGE_SIZE, size / PAGE_SIZE, target, PageType::Memory);
 }
 
-void MapIoRegion(VAddr base, u32 size) {
+void MapIoRegion(VAddr base, u32 size, MMIORegionPointer mmio_handler) {
     ASSERT_MSG((size & PAGE_MASK) == 0, "non-page aligned size: %08X", size);
     ASSERT_MSG((base & PAGE_MASK) == 0, "non-page aligned base: %08X", base);
     MapPages(base / PAGE_SIZE, size / PAGE_SIZE, nullptr, PageType::Special);
+
+    current_page_table->special_regions.emplace_back(SpecialRegion{base, size, mmio_handler});
 }
 
 void UnmapRegion(VAddr base, u32 size) {
@@ -91,6 +105,22 @@ void UnmapRegion(VAddr base, u32 size) {
     ASSERT_MSG((base & PAGE_MASK) == 0, "non-page aligned base: %08X", base);
     MapPages(base / PAGE_SIZE, size / PAGE_SIZE, nullptr, PageType::Unmapped);
 }
+
+/**
+ * This function should only be called for virtual addreses with attribute `PageType::Special`.
+ */
+static MMIORegionPointer GetMMIOHandler(VAddr vaddr) {
+    for (const auto& region : current_page_table->special_regions) {
+        if (vaddr >= region.base && vaddr < (region.base + region.size)) {
+            return region.handler;
+        }
+    }
+    ASSERT_MSG(false, "Mapped IO page without a handler @ %08X", vaddr);
+    return nullptr; // Should never happen
+}
+
+template<typename T>
+T ReadMMIO(MMIORegionPointer mmio_handler, VAddr addr);
 
 template <typename T>
 T Read(const VAddr vaddr) {
@@ -108,13 +138,16 @@ T Read(const VAddr vaddr) {
         return 0;
     case PageType::Memory:
         ASSERT_MSG(false, "Mapped memory page without a pointer @ %08X", vaddr);
+        break;
     case PageType::Special:
-        LOG_ERROR(HW_Memory, "I/O reads aren't implemented yet @ %08X", vaddr);
-        return 0;
+        return ReadMMIO<T>(GetMMIOHandler(vaddr), vaddr);
     default:
         UNREACHABLE();
     }
 }
+
+template<typename T>
+void WriteMMIO(MMIORegionPointer mmio_handler, VAddr addr, const T data);
 
 template <typename T>
 void Write(const VAddr vaddr, const T data) {
@@ -131,9 +164,10 @@ void Write(const VAddr vaddr, const T data) {
         return;
     case PageType::Memory:
         ASSERT_MSG(false, "Mapped memory page without a pointer @ %08X", vaddr);
+        break;
     case PageType::Special:
-        LOG_ERROR(HW_Memory, "I/O writes aren't implemented yet @ %08X", vaddr);
-        return;
+        WriteMMIO<T>(GetMMIOHandler(vaddr), vaddr, data);
+        break;
     default:
         UNREACHABLE();
     }
@@ -189,6 +223,46 @@ void WriteBlock(const VAddr addr, const u8* data, const size_t size) {
     for (u32 offset = 0; offset < size; offset++) {
         Write8(addr + offset, data[offset]);
     }
+}
+
+template<>
+u8 ReadMMIO<u8>(MMIORegionPointer mmio_handler, VAddr addr) {
+    return mmio_handler->Read8(addr);
+}
+
+template<>
+u16 ReadMMIO<u16>(MMIORegionPointer mmio_handler, VAddr addr) {
+    return mmio_handler->Read16(addr);
+}
+
+template<>
+u32 ReadMMIO<u32>(MMIORegionPointer mmio_handler, VAddr addr) {
+    return mmio_handler->Read32(addr);
+}
+
+template<>
+u64 ReadMMIO<u64>(MMIORegionPointer mmio_handler, VAddr addr) {
+    return mmio_handler->Read64(addr);
+}
+
+template<>
+void WriteMMIO<u8>(MMIORegionPointer mmio_handler, VAddr addr, const u8 data) {
+    mmio_handler->Write8(addr, data);
+}
+
+template<>
+void WriteMMIO<u16>(MMIORegionPointer mmio_handler, VAddr addr, const u16 data) {
+    mmio_handler->Write16(addr, data);
+}
+
+template<>
+void WriteMMIO<u32>(MMIORegionPointer mmio_handler, VAddr addr, const u32 data) {
+    mmio_handler->Write32(addr, data);
+}
+
+template<>
+void WriteMMIO<u64>(MMIORegionPointer mmio_handler, VAddr addr, const u64 data) {
+    mmio_handler->Write64(addr, data);
 }
 
 PAddr VirtualToPhysicalAddress(const VAddr addr) {
--- a/src/core/mmio.h
+++ b/src/core/mmio.h
@@ -1,0 +1,34 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
+
+#include "common/common_types.h"
+
+namespace Memory {
+
+/**
+ * Represents a device with memory mapped IO.
+ * A device may be mapped to multiple regions of memory.
+ */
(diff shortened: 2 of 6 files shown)


neobrain:

2016-01-26T23:29:39Z: https://github.com/citra-emu/citra/compare/c407b6ce2fbd6f438e5258bc14dc3def1a8c33e5...MerryMage:9312132b86d06256cedfc84a63cad8413a9dbba5

src/core/mmio.h
```diff
@@ -0,0 +1,35 @@
...
+    virtual u32 Read32(VAddr addr) = 0;
+    virtual u64 Read64(VAddr addr) = 0;
+
+    virtual void Write8(VAddr addr, u8 data) = 0;
+    virtual void Write16(VAddr addr, u16 data) = 0;
+    virtual void Write32(VAddr addr, u32 data) = 0;
+    virtual void Write64(VAddr addr, u64 data) = 0;
+};
+
+typedef std::shared_ptr<MMIORegion> MMIORegionPointer;
```
lioncash: using MMIORegionPointer = std::shared_ptr<MMIORegion>

src/core/mmio.h
```diff
@@ -0,0 +1,35 @@
...
+#pragma once
+
+#include <memory>
+
+#include "common/common_types.h"
+
+namespace Memory {
+
+/**
+* MMIORegion:
```
lioncash: asterisks should align

src/core/hle/kernel/memory.cpp
```diff
@@ -133,6 +133,8 @@ void InitLegacyAddressSpace(Kernel::VMManager& address_space) {
     auto shared_page_vma = address_space.MapBackingMemory(SHARED_PAGE_VADDR,
             (u8*)&SharedPage::shared_page, SHARED_PAGE_SIZE, MemoryState::Shared).MoveFrom();
     address_space.Reprotect(shared_page_vma, VMAPermission::Read);
+
+    // MapMMIO calls will go here.
```
lioncash: This should likely be prefixed with TODO or NOTE

neobrain:

2016-01-27T00:09:56Z: https://github.com/citra-emu/citra/compare/c407b6ce2fbd6f438e5258bc14dc3def1a8c33e5...MerryMage:4f213c24eafb198baa4fd474cbd0a606e06f5b17

src/core/memory.cpp
```diff
@@ -41,8 +41,14 @@ struct PageTable {
     std::array<u8*, NUM_ENTRIES> pointers;
 
     /**
+    * Array of MMIO handlers backing each page. An entry can only be non-null if the
+    * corresponding entry in the `attributes` array is of type `Special`.
+    */
+    std::array<MMIORegionPointer, NUM_ENTRIES> mmio_handlers;
```
yuriks: Instead of this, I think MMIO handlers can be kept in a small list (array) which just has a memory range and the handler. Then you just scan/binary search the table looking for the correct handler. This table contains 2^24 entries, shared_ptr is about 2 pointers in size, on x64 a pointer is 8 bytes. So just this array will end up using 2*8*1024*1024 = 16MB of memory, most of which will be unused.

The reason `pointers` uses this is for speed, since it can be done in just one lookup. MMIO access otoh is very rare (I think DSP is the only thing that uses it.) so there's no reason to optimize it for speed instead of size.

Another advantage is that then MMIO handlers could be installed with smaller than page granularity, which is something we'll probably want.

An alternative would be to restrict the size of this table to cover just the small (0x1E80~0x1FF8 0000) window of address space that can be actually used to map IO registers. This wouldn't have the aforementioned flexibility advantage though.

yuriks:
Other than commented, looks good. I'm curious though, what do you plan on using this for? Afaik this shouldn't be necessary for the DSP. The mapped region for the DSP is just shared memory, there's nothing that's triggered-on-write there, the DSP just "polls" it every audio frame (~5ms).

src/core/hle/kernel/memory.cpp
```diff
@@ -133,6 +133,8 @@ void InitLegacyAddressSpace(Kernel::VMManager& address_space) {
     auto shared_page_vma = address_space.MapBackingMemory(SHARED_PAGE_VADDR,
             (u8*)&SharedPage::shared_page, SHARED_PAGE_SIZE, MemoryState::Shared).MoveFrom();
     address_space.Reprotect(shared_page_vma, VMAPermission::Read);
+
+    // TODO: MapMMIO calls will go here.
```
yuriks: This wouldn't be the right place for these. They would probably go into individual modules' setup functions, inside hw/. This function is only for transitioning and I plan on removing it once I finish moving the areas it currently setups to the new system. (Our memory system used to be hardcoded, and I need this for compatibility while the kernel code isn't done yet.)

MerryMage:
1.  Implementing it this way would make it easier to understand what is going on, since it would be easier to trace writes and reads from the memory region.
2.  If I poll svcQueryMemory over the DSP memory region, 0x1ff50000-0x1ff57fff and 0x1ff70000-0x1ff77fff are marked as IO. So if I wanted to accurately reflect this, I would have to do something regarding this. 
3. Implementing output capture would be more efficient if I only had to give userland stuff on Read. (In my mind at least, *untested* *assumption*, otherwise it'll just be memcpying audio data around every audio frame (~256 bytes? probably not terrible.). I also do not understand how output capture is implemented as I haven't really tested that bit properly yet. If it would be possible for audio data to be modified by the ARM11 inflight by writing to this region I would like to know about it.)
4. Basically I do not undersatnd enough about the DSP to say with confidence that triggering-on-write would not be required.

As far as I can see my options were:

1. Do it as shared memory and implement something in the memory system that backs an IO-marked region with memory, and continue implementing DSP as per original PR.
2. (Just thought of this:) Do it as shared memory and implement something in the memory system that backs an IO-marked region with memory, but instead implement DSP similar to how SharedPage is.
3. Implement MMIO as was probably intended and implement DSP in terms of this.

Would 2 be preferable to 3, then moving to 3 if necessary?

src/core/memory.cpp
MerryMage: Okay, I'll implement a sorted std::vector<> and a scan over it (since it would only have two entries).

MerryMage:
> The state of the system might change more often than an audio tick and only writing to memory every audio tick would be inaccurate. Giving most up to date state on a Read would be nice.

To give an example of that, the current play-position in the buffer is exposed to userland via the DSP. I would prefer to implement this by querying the audio frontend then returning a value based on that, rather than updating a memory address at ~22kHz which seems inefficient.

neobrain:

2016-01-27T22:14:25Z: https://github.com/citra-emu/citra/compare/c407b6ce2fbd6f438e5258bc14dc3def1a8c33e5...MerryMage:1b38f5eb2b4cfb99866e620841614dc11f6924b7

src/core/memory.cpp
```diff
@@ -80,10 +92,12 @@ void MapMemoryRegion(VAddr base, u32 size, u8* target) {
     MapPages(base / PAGE_SIZE, size / PAGE_SIZE, target, PageType::Memory);
 }
 
-void MapIoRegion(VAddr base, u32 size) {
+void MapIoRegion(VAddr base, u32 size, MMIORegionPointer mmio_handler) {
     ASSERT_MSG((size & PAGE_MASK) == 0, "non-page aligned size: %08X", size);
     ASSERT_MSG((base & PAGE_MASK) == 0, "non-page aligned base: %08X", base);
     MapPages(base / PAGE_SIZE, size / PAGE_SIZE, nullptr, PageType::Special);
+
+    current_page_table->special_regions.emplace_back(SpecialRegion{base, size, mmio_handler});
```
lioncash: It may be beneficial to define a trivial default constructor (possibly giving regions a valid initial state along side it), and a constructor that assigns to its members for SpecialRegion which allows for forwarding. This line can then be changed to:

```cpp
current_page_table->special_regions.emplace_back(base, size, mmio_handler);
```

yuriks: This is only constructed once. Doing that would be larger than just assigning the members. I think doing this push_back is fine as is.

lioncash: IMO that makes for a less flexible struct, but I don't particularly mind. Was more of a "hey, possibly consider this", than anything else.

MerryMage: Oooh, good to know you can do that.

yuriks:
Note that none of this means that I have a problem with this PR, just about how applicable it is to the DSP. Unfortunately actual MMIO regions are not used by any games, only by system modules, so anything we're running on Citra so far never interacts with this. (The regions mapped for each binary are controlled by the (signed) exheader.)

> 1.  Implementing it this way would make it easier to understand what is going on, since it would be easier to trace writes and reads from the memory region.

Fair point. You could also easily hack this in as an extra check in the read/write functions in memory.cpp, for development (which is what I did a while back.)

> 2.  If I poll svcQueryMemory over the DSP memory region, 0x1ff50000-0x1ff57fff and 0x1ff70000-0x1ff77fff are marked as IO. So if I wanted to accurately reflect this, I would have to do something regarding this. 

The type returned by svcQueryMemory is just a tag provided by the kernel, it doesn't directly correspond to any implementation detail. Regions get the MMIO tag if they're mapped via the extra memory ranges specified in the exheader (this is also used for VRAM by system modules, though I haven't gotten around to adding support for this to our kernel yet.) You can change the tag type reported by Citra with the `MemoryState state` param to MapMMIO/MapBackingMemory/etc. so the svcQueryMemory result is accurate.

> 3. Implementing output capture would be more efficient if I only had to give userland stuff on Read. (In my mind at least, *untested* *assumption*, otherwise it'll just be memcpying audio data around every audio frame (~256 bytes? probably not terrible.). I also do not understand how output capture is implemented as I haven't really tested that bit properly yet. If it would be possible for audio data to be modified by the ARM11 inflight by writing to this region I would like to know about it.)

memcpying a frame of audio data shouldn't be a problem, especially since presumably you will have calculated this during the final mixing already (when doing software mixing instead of OpenAL), so it's cheaper to do that than recalculate when the game tries to read the buffer.

> 4. Basically I do not undersatnd enough about the DSP to say with confidence that triggering-on-write would not be required.

Based on what I know about it (double-buffering, with the frame counters to detect missed frames, etc.) all signs point to it being just memory. You can always pretty easily change it to the MMIO if needed.

> 5. The state of the system might change more often than an audio tick and only writing to memory every audio tick would be inaccurate. Giving most up to date state on a Read would be nice.

The reason the memory area is double-buffered is so that the DSP can read and update one page while the game writes to the other without them clobbering each-other. Based on RE work I've done games always obey the correct page when reading.

> 
> As far as I can see my options were:
> 
> 1. Do it as shared memory and implement something in the memory system that backs an IO-marked region with memory, and continue implementing DSP as per original PR.

This is what you should do, IMO. See my comment on 2 above.

> 2. (Just thought of this:) Do it as shared memory and implement something in the memory system that backs an IO-marked region with memory, but instead implement DSP similar to how SharedPage is.

Unfortunately I don't really understand what you're proposing here. :( Could you clear it up? Are you referring to setting the MemoryState tag in the VMA? If so, then yes, you should do this.

> 3. Implement MMIO as was probably intended and implement DSP in terms of this.
> 
> Would 2 be preferable to 3, then moving to 3 if necessary?

yuriks:
> To give an example of that, the current play-position in the buffer is exposed to userland via the DSP. I would prefer to implement this by querying the audio frontend then returning a value based on that, rather than updating a memory address at ~22kHz which seems inefficient.

Is this a field in the shared memory area? In which struct? My impression of how sync worked is that the DSP "owns" the page it's working with, mixes ~5ms of audio at a time, and then increments the frame counter in the shmem and signals an interrupt.

src/core/memory.cpp
```diff
@@ -92,6 +106,22 @@ void UnmapRegion(VAddr base, u32 size) {
     MapPages(base / PAGE_SIZE, size / PAGE_SIZE, nullptr, PageType::Unmapped);
 }
 
+/**
+ * This function should only be called for virtual addreses with attribute `PageType::Special`.
+ */
+static MMIORegionPointer GetMMIOHandler(VAddr vaddr) {
+    for (const auto& region : current_page_table->special_regions) {
+        if (region.base <= vaddr && vaddr < (region.base + region.size)) {
```
yuriks: `vaddr >= region.base`? I think it's kind confusing when it's inverted compared to the other half.

yuriks:
LGTM other than the one comment.

MerryMage:
> You can change the tag type reported by Citra with the MemoryState state param to MapMMIO/MapBackingMemory/etc. so the svcQueryMemory result is accurate.

Okay! This is what I'll do.

> Unfortunately I don't really understand what you're proposing here. :( Could you clear it up? Are you referring to setting the MemoryState tag in the VMA? If so, then yes, you should do this.

I was just referring to "doing it properly". 

i.e.: Create a structure that has layout equivalent to that in the 3DS's memory, pass a pointer of that into MapBackingMemory.

Also I was blind and didn't notice MapBackingMemory had a MemoryState parameter, I was proposing adding that which already exists.

neobrain:

2016-01-28T07:38:56Z: https://github.com/citra-emu/citra/compare/c407b6ce2fbd6f438e5258bc14dc3def1a8c33e5...MerryMage:1e15cca9587f70adee2ac8f212d77cf6f31c2fbe

MerryMage:
> Is this a field in the shared memory area? In which struct? 

Yes, in the third memory region by Pipe 2's output, fourth if you go by how it's actually laid out in memory. 

https://github.com/MerryMage/citra/blob/dsp_hle_save/src/core/hle/service/dsp_dsp.cpp#L150

Userspace however doesn't seem to read the status field more than once per audiotick afaik though so not an actual concern.

> My impression of how sync worked is that the DSP "owns" the page it's working with, mixes ~5ms of audio at a time, and then increments the frame counter in the shmem and signals an interrupt.

Userspace does the incrementation, but otherwise yes, exactly Double-buffering to prevent conflict.

yuriks:
LGTM

neobrain:

2016-01-30T18:42:06Z: https://github.com/citra-emu/citra/compare/3ab6d77c206c33e5036fcde1de0663c2f72777ea...MerryMage:2b9331334884349dc38cb9a447018dca0e5b0d9d

bunnei:
LGTM

PR #1383, opened 2016-01-30
GDB stub fixes

WinterMute:
These changes allow the gdb stub to be enabled and configured from the command line which allows for easier integration with IDEs that do source level debugging.

GDB complains Remote 'g' packet reply is too long with this setup because citra is attempting to send float registers it's not expecting, even if I strip out the legacy fpa registers which are no longer used. I'm still attempting to figure out how to get gdb to accept vfp registers.

The Qt build also appears to have problems with gdb communication but these patches don't change that behaviour.

### DIFF
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -34,25 +34,40 @@
 
 static void PrintHelp()
 {
-    std::cout << "Usage: citra <filename>" << std::endl;
+    std::cout << "Usage: citra [options] <filename>" << std::endl;
+    std::cout << "--help, -h            Display this information" << std::endl;
+    std::cout << "--gdbport, -g number  Enable gdb stub on port number" << std::endl;
 }
 
 /// Application entry point
 int main(int argc, char **argv) {
     int option_index = 0;
+    u32 gdb_port = 0;
+    char *endarg;
     std::string boot_filename;
+
     static struct option long_options[] = {
         { "help", no_argument, 0, 'h' },
+        { "gdbport", required_argument, 0, 'g' },
         { 0, 0, 0, 0 }
     };
 
     while (optind < argc) {
-        char arg = getopt_long(argc, argv, ":h", long_options, &option_index);
+        char arg = getopt_long(argc, argv, ":hg:", long_options, &option_index);
         if (arg != -1) {
             switch (arg) {
             case 'h':
                 PrintHelp();
                 return 0;
+            case 'g':
+                errno = 0;
+                gdb_port = strtoul(optarg, &endarg, 0);
+                if (endarg == optarg) errno = EINVAL;
+                if (errno != 0) {
+                    perror("--gdbport");
+                    exit(1);
+                }
+                break;
             }
         } else {
             boot_filename = argv[optind];
@@ -73,8 +88,10 @@ int main(int argc, char **argv) {
     Config config;
     log_filter.ParseFilterString(Settings::values.log_filter);
 
-    GDBStub::ToggleServer(Settings::values.use_gdbstub);
-    GDBStub::SetServerPort(static_cast<u32>(Settings::values.gdbstub_port));
+    if (gdb_port != 0) {
+        GDBStub::ToggleServer(true);
+        GDBStub::SetServerPort(gdb_port);
+    }
 
     EmuWindow_GLFW* emu_window = new EmuWindow_GLFW;
 
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -60,6 +60,59 @@ const u32 R15_REGISTER = 15;
 const u32 CPSR_REGISTER = 25;
 const u32 FPSCR_REGISTER = 58;
 
+// for sample xmls see the gdb source /gdb/features
+// gdb also wants the l character at the start
+// this XML defines what the registers are for this specific ARM device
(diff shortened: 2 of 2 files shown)


neobrain:

2016-01-30T14:32:12Z: https://github.com/citra-emu/citra/compare/3ab6d77c206c33e5036fcde1de0663c2f72777ea...WinterMute:9c4d9f4aa5c406d57c4f7ffb6438ab8d47de7d52

jroweboy:
@WinterMute You can tell GDB the list of registers as one of the query parameters. For example, I used this target description https://github.com/jroweboy/citra/blob/gdb-stub/src/core/arm/gdb_stub.cpp#L41 which should get sent in response to gdb's query packet. https://github.com/jroweboy/citra/blob/gdb-stub/src/core/arm/gdb_stub.cpp#L251 for where I sent it, and it should match up to citra's gdb stub here https://github.com/citra-emu/citra/blob/master/src/core/gdbstub/gdbstub.cpp#L353

Note that according to the documentation, one should be able to just tell GDB to use the target description from its source, but I could never seem to get that to work https://sourceware.org/gdb/onlinedocs/gdb/Target-Descriptions.html#Target-Descriptions

neobrain:

2016-02-01T13:10:29Z: https://github.com/citra-emu/citra/compare/3ab6d77c206c33e5036fcde1de0663c2f72777ea...WinterMute:c99a4079980822255f7c6ee5e7ba99957c741aaa

src/citra/citra.cpp
```diff
@@ -34,25 +34,40 @@
...
 /// Application entry point
 int main(int argc, char **argv) {
     int option_index = 0;
+    u32 gdb_port = 0;
+    char *endarg;
     std::string boot_filename;
+
     static struct option long_options[] = {
         { "help", no_argument, 0, 'h' },
+        { "gdbport", required_argument, 0, 'g' },
```
Subv: Why is this required?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -353,8 +406,15 @@ static void SendReply(const char* reply) {
...
 
-    if (!strcmp(reinterpret_cast<const char*>(command_buffer + 1), "TStatus")) {
+    const char *query = reinterpret_cast<const char*>(command_buffer + 1);
+
+    if (strcmp(query, "TStatus") == 0 ) {
         SendReply("T0");
+    } else if (strncmp(query, "Supported:",strlen("Supported:")) == 0) {
+        /// PacketSize needs to be large enough for target xml
+        SendReply("PacketSize=800;qXfer:features:read+");
+    } else if (strncmp(query, "Xfer:features:read:target.xml:",strlen("Xfer:features:read:target.xml:")) == 0) {
```
Subv: Whitespace before each parameter

neobrain:
@WinterMute While you're at fixing up stuff, I also suggest setting SO_REUSEADDR on the stub socket, since otherwise connecting to the stub will fail every second Citra run or so.

bunnei:
@WinterMute do you plan on responding to the above comments?

WinterMute:
@bunnei yeah, sorry. Haven't got around to looking at this again yet. Will do soon

src/core/gdbstub/gdbstub.cpp
```diff
@@ -353,8 +406,15 @@ static void SendReply(const char* reply) {
 static void HandleQuery() {
     LOG_DEBUG(Debug_GDBStub, "gdb: query '%s'\n", command_buffer + 1);
 
-    if (!strcmp(reinterpret_cast<const char*>(command_buffer + 1), "TStatus")) {
+    const char *query = reinterpret_cast<const char*>(command_buffer + 1);
```
bunnei: `const char* query` (minor whitespace nitpick)

src/core/gdbstub/gdbstub.cpp
```diff
@@ -353,8 +406,15 @@ static void SendReply(const char* reply) {
...
+
+    if (strcmp(query, "TStatus") == 0 ) {
         SendReply("T0");
+    } else if (strncmp(query, "Supported:",strlen("Supported:")) == 0) {
```
bunnei: whitepace

src/core/gdbstub/gdbstub.cpp
```diff
@@ -353,8 +406,15 @@ static void SendReply(const char* reply) {
...
+        /// PacketSize needs to be large enough for target xml
```
bunnei: `//`

bunnei: XML

src/core/gdbstub/gdbstub.cpp
```diff
@@ -60,6 +60,59 @@ const u32 R15_REGISTER = 15;
 const u32 CPSR_REGISTER = 25;
 const u32 FPSCR_REGISTER = 58;
 
+// for sample xmls see the gdb source /gdb/features
```
bunnei: Super huge pedantic nit, but I'm going to be _that guy_ anyways:
<pre>
// For sample XML fikes see the GDB source /gdb/features
// GDB also wants the l character at the start
// This XML defines what the registers are for this specific ARM device
</pre>

bunnei:
Superceded by #1644, @polaris- offered to wrap this one up!

PR #1388, opened 2016-02-03
OpenGL: Downgrade GL_DEBUG_SEVERITY_NOTIFICATION to Debug logging level

yuriks:
The nVidia driver is *extremely* spammy on this category, sending a
message on every buffer or texture upload, slowing down the emulator and
making the log useless.

### DIFF
--- a/src/video_core/renderer_opengl/renderer_opengl.cpp
+++ b/src/video_core/renderer_opengl/renderer_opengl.cpp
@@ -436,8 +436,6 @@ static void DebugHandler(GLenum source, GLenum type, GLuint id, GLenum severity,
         level = Log::Level::Warning;
         break;
     case GL_DEBUG_SEVERITY_NOTIFICATION:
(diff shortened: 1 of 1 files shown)


neobrain:

2016-02-03T06:46:24Z: https://github.com/citra-emu/citra/compare/cd0b31fd7334d574477a3c92bcbd392e8c3eb5e1...yuriks:05356543d9f8c6af1fbf7b1d498b23eb54f9b850

lioncash:
PR of 2016 right here

hrydgard:
Generally, it's not a great idea to enable GL debug logging by default anyway, unless some debug option is set in the emulator. Turning on debug logging can enable expensive extra validation in the driver.

Of course as Citra is still in an early state this might not be that important, but keep it in mind as it gets faster and better :)

yuriks:
Duly noted. This was really just a stopgap solution to make Citra usable on my new system. :)

PR #1394, opened 2016-02-05
Add immediate mode vertex submission

ds84182:
This adds support for Immediate Mode Vertex Submission. I have tested this with Mii Maker and [Citro3D's Immediate Mode example](https://github.com/devkitPro/3ds-examples/tree/master/graphics/gpu/immediate).

Triangles are drawn when the PICA GPU Mode register is reset to configuration mode.

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -15,6 +15,7 @@
 #include "video_core/clipper.h"
 #include "video_core/command_processor.h"
 #include "video_core/pica.h"
+#include "video_core/pica_state.h"
 #include "video_core/primitive_assembly.h"
 #include "video_core/renderer_base.h"
 #include "video_core/video_core.h"
@@ -73,6 +74,14 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
             GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::P3D);
             break;
 
+        case PICA_REG_INDEX_WORKAROUND(vs_default_attributes_setup.index, 0x232):
+            if (regs.vs_default_attributes_setup.index == 15) {
+                // Reset immediate primitive state
+                g_state.immediate.primitive_assembler.Reconfigure(regs.triangle_topology);
+                g_state.immediate.attribute_id = 0;
+            }
+            break;
+
         // Load default vertex input attributes
         case PICA_REG_INDEX_WORKAROUND(vs_default_attributes_setup.set_value[0], 0x233):
         case PICA_REG_INDEX_WORKAROUND(vs_default_attributes_setup.set_value[1], 0x234):
@@ -108,10 +117,47 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                           attribute.w.ToFloat32());
 
                 // TODO: Verify that this actually modifies the register!
-                setup.index = setup.index + 1;
+                if (setup.index < 15) {
+                    setup.index++;
+                } else {
+                    // Put each attribute into an immediate input buffer.
+                    // When all specified immediate attributes are present, the Vertex Shader is invoked and everything is
+                    // sent to the primitive assembler.
+
+                    auto& immediate_input = g_state.immediate.input;
+                    auto& immediate_attribute_id = g_state.immediate.attribute_id;
+                    const auto& attribute_config = regs.vertex_attributes;
+
+                    immediate_input.attr[immediate_attribute_id++] = attribute;
+
+                    if (immediate_attribute_id >= attribute_config.GetNumTotalAttributes()) {
+                        immediate_attribute_id = 0;
+
+                        Shader::UnitState<false> shader_unit;
+                        Shader::Setup(shader_unit);
+
+                        // Send to vertex shader
+                        Shader::OutputVertex output = Shader::Run(shader_unit, immediate_input, attribute_config.GetNumTotalAttributes());
+
+                        // Send to renderer
+                        using Pica::Shader::OutputVertex;
+                        auto AddTriangle = [](const OutputVertex& v0, const OutputVertex& v1, const OutputVertex& v2) {
+                            VideoCore::g_renderer->rasterizer->AddTriangle(v0, v1, v2);
+                        };
+
+                        g_state.immediate.primitive_assembler.SubmitVertex(output, AddTriangle);
+                    }
+                }
             }
             break;
         }
+
+        case PICA_REG_INDEX(gpu_mode):
+            if (regs.gpu_mode == Regs::GPUMode::Configuring && regs.vs_default_attributes_setup.index == 15) {
+                // Draw immediate mode triangles when GPU Mode is set to GPUMode::Configuring
+                VideoCore::g_renderer->rasterizer->DrawTriangles();
+            }
+            break;
 
         case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[0], 0x23c):
         case PICA_REG_INDEX_WORKAROUND(command_buffer.trigger[1], 0x23d):
--- a/src/video_core/primitive_assembly.h
+++ b/src/video_core/primitive_assembly.h
@@ -30,6 +30,16 @@ struct PrimitiveAssembler {
      */
     void SubmitVertex(VertexType& vtx, TriangleHandler triangle_handler);
 
+    /**
+     * Resets the internal state of the PrimitiveAssembler.
+     */
+    void Reset();
+
+    /**
+     * Reconfigures the PrimitiveAssembler to use a different triangle topology.
+     */
+    void Reconfigure(Regs::TriangleTopology topology);
+
 private:
     Regs::TriangleTopology topology;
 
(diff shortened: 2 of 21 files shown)


neobrain:

2016-02-05T22:42:33Z: https://github.com/citra-emu/citra/compare/69e2453649b30962ac68ef54ffbf30ab4f1a9842...ds84182:e26cef097dbbf51b2d4b55fab26335e9a76d5c98

puntywalker:
Wow, i just made an issue being that Mii Maker crashes...
 is this a fix for
 SVC::CreateThread::<lambda_3758cb5c4f75f5f350d285b6986e0bdf>::operator ():481: Assertion Failed!

Subv:
>Triangles are drawn when the PICA GPU Mode register is reset to configuration mode.
Is there any way we can verify this is actually what happens on hardware?

yuriks:
> Is there any way we can verify this is actually what happens on hardware?
It's not, hardware draws triangles as soon as a complete one arrives, but that makes it impossible to batch triangles for the HW rasterizer. The configuration mode register is a good one to use since you're required to flip it before and after drawing things.

src/video_core/primitive_assembly.h
```diff
@@ -30,6 +30,16 @@ struct PrimitiveAssembler {
      */
     void SubmitVertex(VertexType& vtx, TriangleHandler triangle_handler);
 
+    /*
```
bunnei: This line should be `/**` for the proper docstring format

src/video_core/primitive_assembly.h
```diff
@@ -30,6 +30,16 @@ struct PrimitiveAssembler {
...
+     * Resets the internal state of the PrimitiveAssembler.
+     */
+    void Reset();
+
+    /*
```
bunnei: Ditto

src/video_core/command_processor.cpp
```diff
@@ -108,11 +120,48 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                        immediate_primitive_assembler.SubmitVertex(output, AddTriangle);
+                    }
+                }
             }
             break;
         }
 
+        case PICA_REG_INDEX(gpu_mode):
+            if (regs.gpu_mode == Regs::GPUMode::Configuring && regs.vs_default_attributes_setup.index == 15) {
+                // Draw immediate mode triangles when GPU Mode is set to configuration
```
bunnei: More clear wording:
`// Draw immediate mode triangles when GPUMode is Configuring`

src/video_core/command_processor.cpp
```diff
@@ -108,11 +120,48 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                           attribute.w.ToFloat32());
 
                 // TODO: Verify that this actually modifies the register!
-                setup.index = setup.index + 1;
+                if (setup.index < 15) setup.index++;
```
bunnei:
```c++
if (setup.index < 15) {
     setup.index++;
} else {
   ...
```

src/video_core/command_processor.cpp
```diff
@@ -108,11 +120,48 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                else {
+                    // Put each attribute into an immediate input buffer.
+                    // When all specified immediate attributes are present, the Vertex Shader is invoked and everything is
+                    // sent to the primitive assembler.
+
+                    static Shader::InputVertex immediate_input;
```
bunnei: Don't use static local variables like this. It makes proper state management of the emulator very difficult. (e.g. imagine how you would implement save states with this... short answer is, you wouldn't)

src/video_core/command_processor.cpp
```diff
@@ -108,11 +120,48 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+
+                        Shader::UnitState<false> shader_unit;
+                        Shader::Setup(shader_unit);
+
+                        // Send to vertex shader
+                        Shader::OutputVertex output = Shader::Run(shader_unit, immediate_input, attribute_config.GetNumTotalAttributes());
+
+                        // Send to renderer
+                        using Pica::Shader::OutputVertex;
+                        auto AddTriangle = [](
```
bunnei: This line break hurts my eyes :<
I'd say get rid of it

src/video_core/command_processor.cpp
```diff
@@ -73,6 +77,14 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
             GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::P3D);
             break;
 
+        case PICA_REG_INDEX_WORKAROUND(vs_default_attributes_setup.index, 0x232):
+            if (regs.vs_default_attributes_setup.index == 15) {
+                // Reset immediate primitive state
+                immediate_primitive_assembler.Reconfigure(regs.triangle_topology.Value());
```
bunnei: Is this .Value() necessary? I don't think it is...

src/video_core/command_processor.cpp
```diff
@@ -45,6 +45,10 @@ static const u32 expand_bits_to_bytes[] = {
 
 MICROPROFILE_DEFINE(GPU_Drawing, "GPU", "Drawing", MP_RGB(50, 50, 240));
 
+// This is constructed with a dummy triangle topology
+static PrimitiveAssembler<Shader::OutputVertex> immediate_primitive_assembler(Regs::TriangleTopology::List);
```
bunnei: Do these need to be properly reinitialized when the emulator is reset? (I'm guessing yes; if so it should be in the relevant Init() method for this module)

yuriks: They should probably be. I think this should be move to the "pica state" struct we have in pica.h. However, this iirc introduces circular dependencies, so that struct needs to be moved out of pica.h and into a new file.

src/video_core/command_processor.cpp
yuriks: +1

src/video_core/command_processor.cpp
yuriks: I think the current wording is better, since it's the actual edge that triggers drawing, not the level.

yuriks: (Though using the actual identifiers is better, I agree.)

src/video_core/command_processor.cpp
```diff
@@ -45,6 +45,10 @@ static const u32 expand_bits_to_bytes[] = {
...
+static int immediate_attribute_id = 0;
```
yuriks: Would also encourage moving this one to the state struct.

src/video_core/command_processor.cpp
yuriks: Also, this should then probably also be used by the batch submission path.

neobrain:

2016-02-13T21:43:54Z: https://github.com/citra-emu/citra/compare/f1d1049c4fdf51fd382f9afb6b5cd5f104e71b0e...ds84182:4369c7bde06bf89dbc6ff9295b2603fa26b598fc

src/video_core/immediate.h
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+
+#include "video_core/pica.h"
+#include "video_core/primitive_assembly.h"
+#include "video_core/shader/shader.h"
+
+namespace Pica {
+
+/// Struct used to describe immediate mode rendering state
+struct ImmediateModeState {
```
bunnei: This should be a member of `Pica::State` (https://github.com/citra-emu/citra/blob/master/src/video_core/pica.h#L1296). Thats the point of that struct, to consolidate all Pica state information to the `g_state` instance. Then you can get rid of this file and the .cpp.

bunnei:
LGTM after the above change is made.

ds84182:
I can't make it part of Pica::State because of the circular dependencies with primitive_assembly.h, unless I move the entire State struct to another file. At this point I'm thinking it would be the best if I moved the enums from Pica::Regs to it's own file (pica_constants.h or something?), which would resolve the circular dependency in primitive_assembly.h.

yuriks:
@ds84182 I think moving the state struct would be best. Moving the enums would be at best a stop gap before you run into new problems where you have some other type-related circular dependency.

neobrain:

2016-02-21T01:11:58Z: https://github.com/citra-emu/citra/compare/f1d1049c4fdf51fd382f9afb6b5cd5f104e71b0e...ds84182:a50f6dd85d96cf170520cfce9b1c996635514ee5

mailwl:
add the issue have to be closed after merge
https://github.com/citra-emu/citra/issues/1170

yuriks:
LGTM (tested with the citro3d example)

Unfortunately Etrian Odyssey 4 still doesn't display anything this time, either :(

yuriks:
(Also, in the future please leave a comment when you update the PR, I hadn't noticed it.)

ds84182:
I'll dump Etrian Odyssey 4 later today so I can do some testing with it

yuriks:
Oh I didn't mean to imply that was a bug in the PR, sorry. Afaict it fails to rasterize due to unrelated depth problems, the vertices are fine.

ds84182:
Sounds like some more work for tfarley :P

bunnei:
LGTM

Issue #974, opened 2015-07-22
Glitchy meshes in several games


JohnGodgames:
#### Turbo - Super Stunt Squad:

![turbo](https://cloud.githubusercontent.com/assets/11453507/8816783/895055b6-302b-11e5-84e2-7f56d065cb3f.PNG)

![turbo 2](https://cloud.githubusercontent.com/assets/11453507/8816786/908ee996-302b-11e5-812e-a4488505b0fa.PNG)

CITRACE: https://goo.gl/QmSKrB

#### Barbie Dreamhouse Party:
SW
![barbie sw](https://cloud.githubusercontent.com/assets/11453507/7871664/a6417502-0591-11e5-9f2c-1fe17a5c3aa5.PNG)

#### 101 Dino Pets 3D:
![101 dino pets 1](https://cloud.githubusercontent.com/assets/11453507/13590246/8784e676-e4de-11e5-9139-2df45aeb4387.JPG)

#### 101 Penguin Pets 3D:
![101 penguins](https://cloud.githubusercontent.com/assets/11453507/13590268/af7fea7c-e4de-11e5-8f19-e1bde76f66ff.JPG)

#### 101 Pony Pets 3D:
![101 pony](https://cloud.githubusercontent.com/assets/11453507/13590271/b36ec716-e4de-11e5-9884-c1f83ceed89c.JPG)

#### The Sims 3:
![the sims 3](https://cloud.githubusercontent.com/assets/11453507/13590288/d2e5fcd6-e4de-11e5-9045-b8e96b7f1236.JPG)

#### The Sims 3 - Pets:
![sims pets](https://cloud.githubusercontent.com/assets/11453507/13590291/d6e37cf0-e4de-11e5-87c0-9f54956fc78d.JPG)

#### Outdoors Unleashed Africa 3D:
![outdoors unleashed africa 3d](https://cloud.githubusercontent.com/assets/11453507/13590294/de035bc2-e4de-11e5-88f7-2b9195d8d571.JPG)

yuriks:
Very likely a duplicate of #970

TheSoraHD:
Since this have the same problem as #970, can anyone try if #1059 solves it?

Subv:
Any news regarding this?

JohnGodgames:
still the same...

Subv:
Could you see if #1269 fixes it please?

JohnGodgames:
hmm nope :( 

[...]

PR #1395, opened 2016-02-06
Add support for padding vertex attributes

ds84182:
This adds support for padding attributes, as defined on [3dbrew](https://www.3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_ATTRIBBUFFERi_CONFIG1). I created a small example (which runs on real hardware) [here](https://github.com/ds84182/3ds-gpu-stuff/tree/master/attribute_padding).

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -160,12 +160,19 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     if (component >= 12)
                         LOG_ERROR(HW_GPU, "Overflow in the vertex attribute loader %u trying to load component %u", loader, component);
                     u32 attribute_index = loader_config.GetComponent(component);
-                    vertex_attribute_sources[attribute_index] = load_address;
-                    vertex_attribute_strides[attribute_index] = static_cast<u32>(loader_config.byte_count);
-                    vertex_attribute_formats[attribute_index] = attribute_config.GetFormat(attribute_index);
-                    vertex_attribute_elements[attribute_index] = attribute_config.GetNumElements(attribute_index);
-                    vertex_attribute_element_size[attribute_index] = attribute_config.GetElementSizeInBytes(attribute_index);
-                    load_address += attribute_config.GetStride(attribute_index);
+                    if (attribute_index < 12) {
+                        vertex_attribute_sources[attribute_index] = load_address;
+                        vertex_attribute_strides[attribute_index] = static_cast<u32>(loader_config.byte_count);
+                        vertex_attribute_formats[attribute_index] = attribute_config.GetFormat(attribute_index);
+                        vertex_attribute_elements[attribute_index] = attribute_config.GetNumElements(attribute_index);
+                        vertex_attribute_element_size[attribute_index] = attribute_config.GetElementSizeInBytes(attribute_index);
+                        load_address += attribute_config.GetStride(attribute_index);
+                    } else if (attribute_index < 16) {
+                        // Attribute ids 12, 13, 14 and 15 signify 4, 8, 12 and 16-byte paddings, respectively
+                        load_address += (attribute_index - 11) * 4;
+                    } else {
+                        UNREACHABLE(); // This is truly unreachable due to the number of bits for each component
+                    }
                 }
             }
 
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -157,8 +157,11 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
 
                 // TODO: What happens if a loader overwrites a previous one's data?
                 for (unsigned component = 0; component < loader_config.component_count; ++component) {
-                    if (component >= 12)
+                    if (component >= 12) {
                         LOG_ERROR(HW_GPU, "Overflow in the vertex attribute loader %u trying to load component %u", loader, component);
+                        continue;
+                    }
+
                     u32 attribute_index = loader_config.GetComponent(component);
                     if (attribute_index < 12) {
                         vertex_attribute_sources[attribute_index] = load_address;


neobrain:

2016-02-06T00:13:49Z: https://github.com/citra-emu/citra/compare/69e2453649b30962ac68ef54ffbf30ab4f1a9842...ds84182:dc8d5617099f6e26ded5bdec47dbe7c460e2ba34

src/video_core/command_processor.cpp
```diff
@@ -160,12 +160,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
                     u32 attribute_index = loader_config.GetComponent(component);
-                    vertex_attribute_sources[attribute_index] = load_address;
-                    vertex_attribute_strides[attribute_index] = static_cast<u32>(loader_config.byte_count);
-                    vertex_attribute_formats[attribute_index] = attribute_config.GetFormat(attribute_index);
-                    vertex_attribute_elements[attribute_index] = attribute_config.GetNumElements(attribute_index);
-                    vertex_attribute_element_size[attribute_index] = attribute_config.GetElementSizeInBytes(attribute_index);
-                    load_address += attribute_config.GetStride(attribute_index);
+                    if (attribute_index >= 12) {
+                        // Padding attribute
+                        load_address += (attribute_index-11)*4;
```
lioncash: space before and after `-` and `*`

yuriks:
@JohnGodgames: Does this fix #974?

JohnGodgames:
nope...

neobrain:

2016-02-06T15:03:13Z: https://github.com/citra-emu/citra/compare/f1d1049c4fdf51fd382f9afb6b5cd5f104e71b0e...ds84182:04ab188d64ed5c667c780197d7a42a8465f3e02a

src/video_core/command_processor.cpp
```diff
@@ -160,12 +160,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
                     u32 attribute_index = loader_config.GetComponent(component);
-                    vertex_attribute_sources[attribute_index] = load_address;
-                    vertex_attribute_strides[attribute_index] = static_cast<u32>(loader_config.byte_count);
-                    vertex_attribute_formats[attribute_index] = attribute_config.GetFormat(attribute_index);
-                    vertex_attribute_elements[attribute_index] = attribute_config.GetNumElements(attribute_index);
-                    vertex_attribute_element_size[attribute_index] = attribute_config.GetElementSizeInBytes(attribute_index);
-                    load_address += attribute_config.GetStride(attribute_index);
+                    if (attribute_index >= 12) {
+                        // Padding attribute
+                        load_address += (attribute_index - 11) * 4;
```
bunnei: Can you change this to be a little more self-documenting? E.g. something like:
<pre>
switch (attribute_index) {
case 0: // Valid vertex attribute indices are 0-11
case 1:
case 2:
case 3:
case 4:
case 5:
case 6:
case 7:
case 8:
case 9:
case 10:
case 11:
    vertex_attribute_sources[attribute_index] = load_address;
    vertex_attribute_strides[attribute_index] = static_cast<u32>(loader_config.byte_count);
    vertex_attribute_formats[attribute_index] = attribute_config.GetFormat(attribute_index);
    vertex_attribute_elements[attribute_index] = attribute_config.GetNumElements(attribute_index);
    vertex_attribute_element_size[attribute_index] = attribute_config.GetElementSizeInBytes(attribute_index);
    load_address += attribute_config.GetStride(attribute_index);
	break;
	
case 12: // 4-byte padding
case 13: // 8-byte padding
case 14: // 12-byte padding
case 15: // 16-byte padding
	load_address += (attribute_index - 11) * 4;
	break;
	
default:
	LOG_CRITICAL(HW_GPU, "Unknown vertex attribute index %d", attribute_index);
    UNREACHABLE();
	break;
}
</pre>
A little verbose, but this makes the intended functionality much more clear :)

bunnei:
LGTM after the above comment is addressed

src/video_core/command_processor.cpp
```diff
@@ -160,12 +160,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     if (component >= 12)
```
mailwl: i think this check not need anymore

src/video_core/command_processor.cpp
yuriks: Eh... I'm not too amazed by this. It explodes the code with verbosity and redundancy, and, to me, isn't even an improvement. I'd just expand on the comment at the top of the if to note "// Attribute ids 12, 13, 14 and 15 signify 4, 8, 12 and 16-byte paddings, respectively"

neobrain:

2016-02-13T21:47:08Z: https://github.com/citra-emu/citra/compare/f1d1049c4fdf51fd382f9afb6b5cd5f104e71b0e...ds84182:630a24cd89462657eea0c7330d6f329b5f928ca7

src/video_core/command_processor.cpp
bunnei: Ditto... I noticed you updated this PR without addressing this comment?

src/video_core/command_processor.cpp
```diff
@@ -160,12 +160,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
                         LOG_ERROR(HW_GPU, "Overflow in the vertex attribute loader %u trying to load component %u", loader, component);
                     u32 attribute_index = loader_config.GetComponent(component);
-                    vertex_attribute_sources[attribute_index] = load_address;
-                    vertex_attribute_strides[attribute_index] = static_cast<u32>(loader_config.byte_count);
-                    vertex_attribute_formats[attribute_index] = attribute_config.GetFormat(attribute_index);
-                    vertex_attribute_elements[attribute_index] = attribute_config.GetNumElements(attribute_index);
-                    vertex_attribute_element_size[attribute_index] = attribute_config.GetElementSizeInBytes(attribute_index);
-                    load_address += attribute_config.GetStride(attribute_index);
+                    if (attribute_index >= 12) {
```
bunnei: I think the solution I provided was more explicit, but meh, the comment below at least documents what going on here. I think this should still be changes to 3 if conditions, though:
</pre>
if >= 0 and < 12: Normal vertex attribute
elsif >= 12 and < 16: Padding
else: Critical error / Unreachable
</pre>

src/video_core/command_processor.cpp
yuriks: This check is unrelated to the target attribute index, they just both happen to bound at 12.

src/video_core/command_processor.cpp
```diff
@@ -160,12 +160,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     if (component >= 12)
                         LOG_ERROR(HW_GPU, "Overflow in the vertex attribute loader %u trying to load component %u", loader, component);
```
yuriks: Just realized. This should `continue` when this message is printed so that the remaining code doesn't make out of bounds array accesses.

yuriks: (This is not introduced by your changes, but might as well fix it here. I'd encourage making it a separate commit in the same PR.)

src/video_core/command_processor.cpp
yuriks: I agree about treating the unreachable case, though most of the tests there are redundant (and >= 0 will even warn, for example). I suggest:
```
if (attribute_index < 12) {
  // Load attribute
} else if (attribute_index < 16) {
  // Padding
} else {
  UNREACHABLE(); // This is truly unreachable due to the number of bits for each component, so I don't think a log is needed.
}
```

neobrain:

2016-02-21T00:03:35Z: https://github.com/citra-emu/citra/compare/0d086616d1af239b8e41b246bea3c1ef85fc907f...ds84182:ed8072b48bf6eacb0eea161f87746086401b5c08

PR #1403, opened 2016-02-09
Dependencies: Remove GLFW, Add SDL2

MerryMage:
Windows binary package used: [SDL2-devel-2.0.4-VC.zip](http://libsdl.org/release/SDL2-devel-2.0.4-VC.zip)

This commit removes GLFW and adds SDL2. EmuWindow_SDL2 is also implemented.

Notes:
* I have only tested this on Windows and Linux as my Mac is out of commission for the forseeable future.
* SDL doesn't seem to support transfering GL contexts between threads. (`EmuWindow_SDL2::MakeCurrent`, `EmuWindow_SDL2::DoneCurrent`)


neobrain:

2016-02-09T22:20:52Z: https://github.com/citra-emu/citra/compare/f1d1049c4fdf51fd382f9afb6b5cd5f104e71b0e...MerryMage:50114eadcca7d9a68caf51e7a8a13d6f5d6b7df2

src/citra/emu_window/emu_window_sdl2.h
```diff
@@ -0,0 +1,59 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <utility>
+
+#include "common/emu_window.h"
+
+struct SDL_Window;
+typedef void *SDL_GLContext;
```
bunnei: `using SDL_GLContext = void*;`, and this can be defined closer to where it's used (as it's only used once)

src/citra/emu_window/emu_window_sdl2.h
```diff
@@ -0,0 +1,59 @@
...
+    void PollEvents() override;
+
+    /// Makes the graphics context current for the caller thread
+    void MakeCurrent() override;
+
+    /// Releases (dunno if this is the "right" word) the GL context from the caller thread
+    void DoneCurrent() override;
+
+    /// Whether the window is still open, and a close request hasn't yet been sent
+    const bool IsOpen();
```
bunnei: `bool IsOpen() const;`

src/citra/emu_window/emu_window_sdl2.h
```diff
@@ -0,0 +1,59 @@
...
+
+    void ReloadSetKeymaps() override;
+
+    void SetTitle(std::string title);
```
bunnei: Is this even used anywhere?

src/citra/emu_window/emu_window_sdl2.h
```diff
@@ -0,0 +1,59 @@
...
+
+private:
+    void OnKeyEvent(int key, u8 state);
```
bunnei: I know that the GLFW was missing docstrings for these, but they'd be good to add

src/citra/emu_window/emu_window_sdl2.h
```diff
@@ -0,0 +1,59 @@
...
+
+    void OnMinimalClientAreaChangeRequest(const std::pair<unsigned, unsigned>& minimal_size) override;
+
+    void OnResize();
+
+    /// Is the window still open?
+    bool is_open = true;
+
+    /// Internal SDL2 render window
+    SDL_Window* m_render_window;
```
bunnei: The `m_` prefix isn't part of our style guide anymore, so these can be removed.

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <algorithm>
+#include <cstdlib>
+#include <string>
+
+#include <glad/glad.h>
+
+#define SDL_MAIN_HANDLED
+#include <SDL.h>
+
+#include "common/assert.h"
```
bunnei: Unused import

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,179 @@
...
+void EmuWindow_SDL2::OnKeyEvent(int key, u8 state) {
+    if (state == SDL_PRESSED) {
+        KeyPressed({ key, keyboard_id });
+    }
+    else if (state == SDL_RELEASED) {
+        KeyReleased({ key, keyboard_id });
+    }
+}
+
+/// Whether the window is still open, and a close request hasn't yet been sent
```
bunnei: Docstrings should just be in the class definition, and shouldn't be duplicated in the implementation

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+void EmuWindow_SDL2::OnResize() {
+    int width, height;
+
+    SDL_GetWindowSize(m_render_window, &width, &height);
+
+    NotifyFramebufferLayoutChanged(EmuWindow::FramebufferLayout::DefaultScreenLayout(width, height));
+}
+
+/// EmuWindow_SDL2 constructor
```
bunnei: Ditto with the above, but also not necessary to have this comment at all IMO

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,179 @@
...
+            TouchReleased();
+        }
+    }
+}
+
+void EmuWindow_SDL2::OnKeyEvent(int key, u8 state) {
+    if (state == SDL_PRESSED) {
+        KeyPressed({ key, keyboard_id });
+    }
+    else if (state == SDL_RELEASED) {
```
bunnei: This statement should be on the previous line with the `}`

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include "core/hle/service/hid/hid.h"
+
+#include "citra/emu_window/emu_window_sdl2.h"
+
+void EmuWindow_SDL2::OnMouseMotion(s32 x, s32 y) {
+    TouchMoved(static_cast<unsigned>(std::max(x, 0)), static_cast<unsigned>(std::max(y, 0)));
+}
+
+void EmuWindow_SDL2::OnMouseButton(u32 button, u8 state, s32 x, s32 y) {
+    if (button == SDL_BUTTON_LEFT) {
```
bunnei: <pre>if (button != SDL_BUTTON_LEFT)
    return;</pre>

bunnei: (I guess I should have added, so that the stuff below doesn't need an extra indentation. It's not so severe here, but a good habit to get into anyways to make code more readable)

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,179 @@
...
+        if (state == SDL_PRESSED) {
+            TouchPressed(static_cast<unsigned>(std::max(x, 0)), static_cast<unsigned>(std::max(y, 0)));
```
bunnei: IIRC our current best practices is to use a C-style cast for simple casts like these (to make the code shorter and more readable)

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,179 @@
...
+        case SDL_QUIT:
+            is_open = false;
+            break;
+        }
+    }
+}
+
+/// Makes the OpenGL context current for the caller thread
+void EmuWindow_SDL2::MakeCurrent() {
+    // There is no ability to do this in SDL as far as I can tell.
```
bunnei: https://wiki.libsdl.org/SDL_GL_MakeCurrent?highlight=%28%5CbCategoryAPI%5Cb%29%7C%28SDLFunctionTemplate%29 ?

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/// Releases (dunno if this is the "right" word) the context from the caller thread
+void EmuWindow_SDL2::DoneCurrent() {
+    // There is no ability to do this in SDL as far as I can tell.
```
bunnei: I think this is just:
<pre>SDL_GL_MakeCurrent(mainWindow, nullptr); </pre>
But you should double check to ensure this is what we want.

src/citra/emu_window/emu_window_sdl2.cpp
MerryMage: Just [checked](https://github.com/spurious/SDL-mirror/blob/dfd007d5c4c0a69db890c3b15aa87f9c6b0c1489/src/video/SDL_video.c#L3282). Yes, that has the desired behavior.

src/citra/emu_window/emu_window_sdl2.cpp
MerryMage: ✓

bunnei:
Discussion time!
I motivated this change from the perspective of a Windows user, but it has come to my attention that some folks oppose nixing GLFW in favor of SDL2. So here are my 2 cents. To give some background, we started using GLFW at a time when it was still quite 'beta' because it allowed us to have the ability to share a GL context between threads (which is actually something that we don't currently rely on for CLI mode, but probably still want), while SDL1.2, SFML (and other similar libraries) did not. Beyond this, our uses for GLFW are quite minimal: We need to create a GL window/context and get some user input via the keyboard and mouse. Now, it's 2016, SDL2 has been a thing for a few years, and it supports everything we need, with essentially a 1:1 API to GLFW. But why switch now?
* @MerryMage is working toward audio support in Citra, and SDL2 supports audio (GLFW does not)
* Citra CLI only relies on a single "hard dependency: right now for input/output - Keeping it this way, while not strictly necessary, will  probably make future porting efforts easier
* SDL is a more widely used/supported library than GLFW. As it gets adapted to other platforms, Citra porting efforts could become essentially free (and often times, SDL gets ported very early on to new platforms, as so many prior homebrew/emulators/games use it in the same way we would)
* Even if we didn't care about all of the above points, I don't see another alternative that wouldn't result in yet another fairly sizeable dependency, perhaps requiring us to write our own build files to work with VS2015 and submodule a lot of source code, or checkin prebuilt binaries as we do for GLFW and Qt.

I'd like to hear opinions from others on this... Particularly if you are against using SDL2 for some reason (for user input, windowing, and audio), and if so, what that reason is, and what your suggested alternative would be to address the issues I discussed above.

src/citra/emu_window/emu_window_sdl2.h
MerryMage: No, it was not. I think my mind was still thinking about that suggestion to add the name of the current application running in citra to the titlebar when I typed that.

lioncash: the title string can also be a const reference.

neobrain:

2016-02-10T12:43:45Z: https://github.com/citra-emu/citra/compare/f1d1049c4fdf51fd382f9afb6b5cd5f104e71b0e...MerryMage:f71cd3f23b293ca42bd0fc73ecabdb7170cbd316

MerryMage:
Cleaning up and pushed just so it's there if anyone needs it.

---

After thinking about it and listening to various voices, this is the summary of my thoughts.

For:

* Reduction in dependencies in the short-term.
* Likely to be easier to port to more obscure platforms.

Against:

* SDL2 does not yet provide audio recording/input. We would probably require another library for that / Submit patches upstream for that when we want to support the microphone.

Notes:

I intend write the AudioCore frontend to support multiple libraries if necessary. (Hence the [Sink](https://github.com/MerryMage/citra/blob/audio-core/src/audio_core/sink.h#L17) class.)

---

Regarding the worry about needing to write buildfiles / port to VC, I found a small library ([libsoundio](http://libsound.io/)) that provides prebuilt binaries for Windows. It's available on macbrew and AUR though it isn't yet in stable on Debian.

yuriks:
> SDL doesn't seem to support transfering GL contexts between threads. 

I assumed this just worked, but looking into the doc reference it seems like it's really not the case. If so, this is a major roadblock. Someone should clarify with the SDL developer.

bunnei:
>  assumed this just worked, but looking into the doc reference it seems like it's really not the case. If so, this is a major roadblock. Someone should clarify with the SDL developer.

```
SDL_GL_MakeCurrent(window, context); // Acquire the context
// do some stuff
SDL_GL_MakeCurrent(window, nullptr); // Release the context
```

bunnei:
There has been minimal response to my suggestion on moving to SDL :< Does anyone have any more comment on this? FWIW not supporting audio input doesn't seem like a deal breaker to me, that could be added to SDL, and it's also not a feature that I'd expect us to care about any time soon.

@MerryMage, can you look into confirming whether the solution I proposed for transferring GL contexts between threads is correct?

MerryMage:
> @MerryMage, can you look into confirming whether the solution I proposed for transferring GL contexts between threads is correct?

Looking at SDL's code, it *looks* correct, but I'll going to ask a SDL dev.

MerryMage:
Couldn't get a reply from an actual dev on IRC, and no replies from their mailing list.

They do [the exact same thing in their SDL_video.c example program](https://github.com/spurious/SDL-mirror/blob/dfd007d5c4c0a69db890c3b15aa87f9c6b0c1489/src/video/SDL_video.c#L2562) so I'm going with 'yeah it's supported but undocumented'.

MerryMage:
Blocking on citra-emu/ext-windows-bin#1.

MerryMage:
Ready.

src/citra/emu_window/emu_window_sdl2.h
```diff
@@ -0,0 +1,64 @@
...
+    /// Swap buffers to display the next frame
+    void SwapBuffers() override;
+
+    /// Polls window events
+    void PollEvents() override;
+
+    /// Makes the graphics context current for the caller thread
+    void MakeCurrent() override;
+
+    /// Releases (dunno if this is the "right" word) the GL context from the caller thread
```
bunnei: It's a good enough word! Feel free to take out the "(dunno if this is the "right" word)"

src/citra/emu_window/emu_window_sdl2.h
```diff
@@ -0,0 +1,64 @@
...
+    void ReloadSetKeymaps() override;
+
+private:
+    /// Called by PollEvents when a key is pressed or released.
+    void OnKeyEvent(int key, u8 state);
+
+    /// Called by PollEvents when the mouse moves.
+    void OnMouseMotion(s32 x, s32 y);
+
+    /// Called by PollEvents when a mouose button is pressed or released
```
bunnei: *mouse

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include "core/hle/service/hid/hid.h"
+
+#include "citra/emu_window/emu_window_sdl2.h"
+
+void EmuWindow_SDL2::OnMouseMotion(s32 x, s32 y) {
+    TouchMoved((unsigned)std::max(x, 0), (unsigned)std::max(y, 0));
+}
+
+void EmuWindow_SDL2::OnMouseButton(u32 button, u8 state, s32 x, s32 y) {
+    if (button == SDL_BUTTON_LEFT)
```
bunnei: Is this supposed to be `button != SDL_BUTTON_LEFT`? At any rate, this doesn't seem to do anything. Left mouse button works (when this seems like it should not), as well as center and right (which probably shouldn't do anything)

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -0,0 +1,169 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include <cstdlib>
+#include <string>
+
+#include <glad/glad.h>
```
bunnei: Is this actually used in here? Maybe I missed something

MerryMage:
Pushed corrections. `glad.h` wasn't used, that was a left-over from GLFW, apologies.

bunnei:
LGTM. Feel free to squash.

.travis-deps.sh
```diff
@@ -5,6 +5,7 @@ set -x
 
 #if OS is linux or is not set
 if [ "$TRAVIS_OS_NAME" = "linux" -o -z "$TRAVIS_OS_NAME" ]; then
+
```
yuriks: A wild newline appears! (Are these intended additions?)

src/citra/CMakeLists.txt
```diff
@@ -27,3 +26,19 @@ target_link_libraries(citra ${PLATFORM_LIBRARIES})
...
 endif()
+
+if (MSVC)
+    set(DLL_DEST "${CMAKE_BINARY_DIR}/bin/$<CONFIG>/")
+
+    # windows commandline expects the / to be \ so switch them
+    string(REPLACE "/" "\\\\" SDL2_DLL_DIR ${SDL2_DLL_DIR})
+    string(REPLACE "/" "\\\\" DLL_DEST ${DLL_DEST})
+
+    # /NJH /NJS /NDL /NFL /NC /NS /NP - Silence any output
```
yuriks: Instead of duplicating, this and the Qt instance should be refactored into a macro or function, otherwise we'll end up updating one copy command and forgetting the other.

archshift:
It works on OS X!

One minor comment, though, is that while Citra is starting up the window now doesn't show any decorations or anything besides a blank white rectangle in the middle of the screen.

externals/cmake-modules/FindSDL2.cmake
```diff
@@ -0,0 +1,224 @@
...
+#    Visualization Initiative.
+# 
+#  * National Alliance for Medical Image Computing (NAMIC) is funded by the
+#    National Institutes of Health through the NIH Roadmap for Medical Research,
+#    Grant U54 EB005149.
+# 
+#  * Kitware, Inc.
+#
+
+message("<FindSDL2.cmake>")
```
yuriks: Is this from the original file or some forgotten debug thing?

src/citra/config.cpp
```diff
@@ -41,51 +41,51 @@ bool Config::LoadINI(INIReader* config, const char* location, const std::string&
...
-    GLFW_KEY_M, GLFW_KEY_N, GLFW_KEY_B,
-    GLFW_KEY_T, GLFW_KEY_G, GLFW_KEY_F, GLFW_KEY_H,
-    GLFW_KEY_UP, GLFW_KEY_DOWN, GLFW_KEY_LEFT, GLFW_KEY_RIGHT,
-    GLFW_KEY_I, GLFW_KEY_K, GLFW_KEY_J, GLFW_KEY_L
+    SDL_SCANCODE_A, SDL_SCANCODE_S, SDL_SCANCODE_Z, SDL_SCANCODE_X,
+    SDL_SCANCODE_Q, SDL_SCANCODE_W, SDL_SCANCODE_1, SDL_SCANCODE_2,
+    SDL_SCANCODE_M, SDL_SCANCODE_N, SDL_SCANCODE_B,
+    SDL_SCANCODE_T, SDL_SCANCODE_G, SDL_SCANCODE_F, SDL_SCANCODE_H,
+    SDL_SCANCODE_UP, SDL_SCANCODE_DOWN, SDL_SCANCODE_LEFT, SDL_SCANCODE_RIGHT,
+    SDL_SCANCODE_I, SDL_SCANCODE_K, SDL_SCANCODE_J, SDL_SCANCODE_L
```
yuriks: The change in key mnemonics will probably break existing config files and the user is going to have to figure out that this is the cause of the problem. I think a good work-around is to rename the INI section that stores the controls so that it's separate from the previous one.

src/citra/config.cpp
```diff
@@ -17,8 +17,8 @@
 
 Config::Config() {
     // TODO: Don't hardcode the path; let the frontend decide where to put the config files.
-    glfw_config_loc = FileUtil::GetUserPath(D_CONFIG_IDX) + "glfw-config.ini";
-    glfw_config = new INIReader(glfw_config_loc);
+    sdl2_config_loc = FileUtil::GetUserPath(D_CONFIG_IDX) + "sdl2-config.ini";
+    sdl2_config = new INIReader(sdl2_config_loc);
```
yuriks: *hiss* Take the opportunity to make this a unique_ptr.

src/citra/emu_window/emu_window_sdl2.h
```diff
@@ -0,0 +1,64 @@
...
+    void OnKeyEvent(int key, u8 state);
+
+    /// Called by PollEvents when the mouse moves.
+    void OnMouseMotion(s32 x, s32 y);
+
+    /// Called by PollEvents when a mouse button is pressed or released
+    void OnMouseButton(u32 button, u8 state, s32 x, s32 y);
+
+    /// Called by PollEvents when any event that may cause the window to be resized occurs
+    void OnResize();
```
yuriks: There's no point in making the private functions above this member of the class, as far as I can see.

mailwl:
Ehhh... may be it helps to port Citra to Android (for Tegra K1/X1 with OpenGL4.5 support)

yuriks:
Works fine here on Windows. Mostly LGTM

externals/cmake-modules/FindSDL2.cmake
MerryMage: From the original file.

src/citra/config.cpp
MerryMage: The file config itself has changed name: s/glfw/sdl2/. Would renaming the section as well be necessary?

src/citra/config.cpp
MerryMage: Yessssssss.

src/citra/config.cpp
yuriks: Oh, nevermind then, I thought it was called simply citra.ini to go along with the exe name.

MerryMage:
@archshift I'll fix that by running the message pump once at startup.

MerryMage:
Updated and squashed into 2 commits.

ProfessorKaos64:
Thanks for the consideration on SDL2. I build 3rd party packages for SteamOS, and i've been wanting to build this for some time to tinker and keep in my testing dist. I am not a full time dev, or pretend to be one, but SDL2's portability has always been easy to work with. My 2 cents is to add an additional library for mic input. Sorry if I derail any talks here. I merged Mage's PR into my fork to test around in the mean time.

Subv:
LGTM

archshift:
@MerryMage yep, that fixed it!

purpasmart96:
LGTM

bunnei:
Thanks for the feedback everyone! Let make it official :cake:

yuriks:
Did anyone test this on Linux? I thought we were going to wait until @linkmauve tested.

mailwl:
glfw still needed it externals?

ProfessorKaos64:
@yuriks , I build a Debian package last night with a manual PR, seemed to work OK, but by "OK" at this stage is up in the air. It did build fine.

bunnei:
@mailwl oops yes! I'll get rid of that tonight if no one else has submitted a PR by then.

@yuriks @MerryMage said on IRC that she tested at least an earlier version of this on a Ubuntu VM, which was good enough for me. I did ping @linkmauve to test it (well rather, tried to give him the opportunity to test it anyway), but did not receive any response.

LittleWhite-tb:
Everything fine on my linux machine. Compiled and started the homebrew launcher properly.

MerryMage:
Tested on an ArchLinux VM (finally got round to setting that up). No issues.

PR #1420, opened 2016-02-23
[RFC] Current Vulkan progress.

ILOVEPIE:
Not in in a complete state, won't run, simply looking to reduce the number of problems blocking the final PR from being merged when we get there. Please keep this PR open as I update the support so we can continue discussion and feedback.


neobrain:

2016-02-23T01:48:46Z: https://github.com/citra-emu/citra/compare/0d086616d1af239b8e41b246bea3c1ef85fc907f...ILOVEPIE:1ac50c17e8d6aceae198ec1ed3cea12875554828

lioncash:
Please open PRs like this when code is at least in a working state and can actually run (so it can be tested). There's not much point in commenting on very incomplete code that is almost inevitably going to be different from the intended final revision.

Issue #1428, opened 2016-02-28
Project X Zone 2 freeze


Punk34:
The game freezes on the first window

![- -1](https://cloud.githubusercontent.com/assets/13983158/13378834/48fd260e-de24-11e5-9aa2-7c43efa2a2fc.jpg)

mailwl:
@Punk34, may u test with latest master?
there is no text shown at all, but in other the game fully playable.

Punk34:
Same thing with "HEAD-b8e9af9"

mhvuze:
Just to add: The same issue is present in b25605e for "MonHun Nikki: Poka Poka Airu Mura DX" when the game tries to write savedata or load savedata.

Subv:
Does this still happen? StoreDataCache is most definitely not the cause of this issue.

Dragios:
This does not happen anymore and can be safely close. It already managed to create save file long time ago.

Issue #1157, opened 2015-09-15
Heroes of the ruin - Hangs at title screen


JohnGodgames:
![heroes of the ruin](https://cloud.githubusercontent.com/assets/11453507/9866752/b2620ad8-5b66-11e5-9cde-47634950755d.PNG)

LOG:
http://pastie.org/10420479

mailwl:
easy to fix, just implement AC:U function "IsConnected", for example
```
static void IsConnected(Service::Interface* self) {
    u32* cmd_buff = Kernel::GetCommandBuffer();

    cmd_buff[1] = 0; // No error
    cmd_buff[2] = 0; // Connection type set to none

    LOG_WARNING(Service_AC, "(STUBBED) called");
}
```

woody2014:
could you submit this as a pull request pls

JohnGodgames:
Fixed by #1429

PR #1429, opened 2016-02-28
ac:u IsConnected implemented

mailwl:
code from: Heroes of the ruin - Hangs at title screen #1157
also fix: Project X Zone 2 - Unimplemented StoreDataCache #1428

### DIFF
--- a/src/core/hle/service/ac_u.cpp
+++ b/src/core/hle/service/ac_u.cpp
@@ -22,8 +22,23 @@ static void GetWifiStatus(Service::Interface* self) {
     // TODO(purpasmart96): This function is only a stub,
     // it returns a valid result without implementing full functionality.
 
-    cmd_buff[1] = 0; // No error
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
     cmd_buff[2] = 0; // Connection type set to none
+
+    LOG_WARNING(Service_AC, "(STUBBED) called");
+}
+
+/**
+ * AC_U::IsConnected service function
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : bool, is connected
+ */
+static void IsConnected(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    cmd_buff[2] = false; // Not connected to ac:u service
 
     LOG_WARNING(Service_AC, "(STUBBED) called");
 }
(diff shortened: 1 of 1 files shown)


src/core/hle/service/ac_u.cpp
```diff
@@ -28,6 +28,21 @@ static void GetWifiStatus(Service::Interface* self) {
     LOG_WARNING(Service_AC, "(STUBBED) called");
 }
 
+/**
+ * AC_U::IsConnected service function
```
bunnei: What is this actually used for?

mailwl: i think, if we don't set cmd_buff[2] to 0, then it contains garbage (not zero), and a game think, the service is connected, and wait something from it forever.
What exactly AC service is - i have no idea

yuriks: Pretty obviously Wifi access point stuff: https://www.3dbrew.org/wiki/AC_Services

src/core/hle/service/ac_u.cpp
```diff
@@ -28,6 +28,21 @@ static void GetWifiStatus(Service::Interface* self) {
...
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : bool, is connected
+ */
+static void IsConnected(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = 0; // No error
```
yuriks: use RESULT_SUCCESS.raw instead of 0.

yuriks:
Change commit message to "ac:u: Stub IsConnected"

LGTM apart from the two nits, though I haven't verified that the actual buffer format is correct.

mailwl:
about buffer format. How it call in Heroes of Ruins: parameters zero and pointer to local var in stack:
```
.text:002BEAA8                 MOV             R0, #0
.text:002BEAAC                 MOV             R1, SP
.text:002BEAB0                 STRB            R0, [SP,#0x10+var_10]
.text:002BEAB4                 BL              IsConnected
.text:002BEAB8                 LDRSB           R0, [SP,#0x10+var_10
```
the function itself:
```c++
int __fastcall IsConnected(int a1, _BYTE *a2)
{
  char *v2; // r4@1
  _DWORD *cmd_buff; // r4@1
  int result; // r0@1

  v2 = (char *)__mrc(15, 0, 13, 0, 3);
  *((_DWORD *)v2 + 0x20) = 0x3E0042;
  cmd_buff = v2 + 0x80;
  cmd_buff[1] = a1;
  cmd_buff[2] = 0x20;
  result = dword_58484C;
  __asm { SVC             0x32 }
  if ( (dword_58484C & 0x80000000 & 0x80000000) == 0 )
  {
    *a2 = *((_BYTE *)cmd_buff + 8); // cmd_buff[2] low byte
    result = cmd_buff[1];
  }
  return result;
}
```

Issue #1129, opened 2015-09-09
Digimon World Re: Digitize Decode - Unable to boot both HW and SW


Kloen:
This game it's most likely unable to boot.

On HW and SW, it only shows a dark, then white, and then dark again screen, and print these:
![image](https://cloud.githubusercontent.com/assets/9112818/9768846/a940d234-5726-11e5-9cf4-87d0e375c78a.png)

mailwl:
i think its easy to cure

Kloen:
Updated with the crash error on HW.

Subv:
@yuriks, It appears to be some y2r problem?

Kloen:
Update: The HW issue it's only on Intel cards, should check the drivers tomorrow, on my main PC the result it's the same as in the software render.

Kloen:
Update: blindly implementing y2r:u 0x00040000 which if I'm right it's GetOutputFormat().
![image](https://cloud.githubusercontent.com/assets/9112818/9796493/bd4c40ee-57f6-11e5-9604-06d5bd36c221.png)

It seems that there are a problem with all the graphic resources of the game, using GLIntercept I found some resources that should had been printed on screen before.

Kloen:
Managed to navigate throught menus idk how, both HW/SW (includes openl-tex-forwarding, will check without it once I reach home, cause HW render does not work on my Intel HD 4600 without that branch):
![image](https://cloud.githubusercontent.com/assets/9112818/10014374/1e92ccda-6119-11e5-85dd-a584d3cfa4a4.png)

Romsstar:
Using the most recent Nightly Build and a custom build where I used GetFormatInfo from Subv's branch result in the same thing:
- Enabling the Hardware Renderer with the game will immediately make Citra crash.
As Kloen already reported this issue on Intel Cards, in my case an Intel HD 530.
- Even after #1097 the game still requires config saveblock data or it won't boot at all.
- The game font is not being rendered. Actually most graphics are being rendered just wrong, but
the font is not being rendered at all.
-When you try to load the savedata it crashes, unfortunately even with debugging on it failed to  provide a reason. 
I assume it has something to do with this part. I attached the debuglog.
[debuglog.txt](https://github.com/citra-emu/citra/files/150222/debuglog.txt)

"[ 107.490148] Service.FS <Debug> core\hle\service\fs\fs_user.cpp:Service::FS::GetFormatInfo:762: archive_path=[Empty]"

[...]

PR #1433, opened 2016-03-01
Service/CFG: Add block 0x000A0000 (username) to default config file

yuriks:
Fixes missing block issue in #1129.

### DIFF
--- a/src/core/hle/service/cfg/cfg.cpp
+++ b/src/core/hle/service/cfg/cfg.cpp
@@ -4,9 +4,10 @@
 
 #include <algorithm>
 
+#include "common/file_util.h"
 #include "common/logging/log.h"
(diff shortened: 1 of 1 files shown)


yuriks:
@Romsstar: Check if this PR fixes that specific issue with Digimon. Make sure to delete the existing config savefile so that Citra re-creates it.

Romsstar:
This PR fixes the config savefile issue in #1129 completely. 
The other problems still persist.

Subv:
:shipit: LGTM

JamePeng:
LGTM

bunnei:
LGTM

wwylele:
I get confused... apparently there is already a name block, what's the point to create another one?

PR #1441, opened 2016-03-02
AudioCore: Implement Pipe 2

MerryMage:
Pipe 2 is a DSP pipe that is used to initialize both the DSP hardware (the application signals to the DSP to initialize) and the application (the DSP provides the memory location of structures in the shared memory region).

This PR implements that pipe, in addition to several service functions in the dsp::DSP service.

### DIFF
--- a/src/audio_core/hle/pipe.cpp
+++ b/src/audio_core/hle/pipe.cpp
@@ -5,50 +5,154 @@
 #include <array>
 #include <vector>
 
+#include "audio_core/hle/dsp.h"
 #include "audio_core/hle/pipe.h"
 
+#include "common/assert.h"
 #include "common/common_types.h"
 #include "common/logging/log.h"
 
 namespace DSP {
 namespace HLE {
 
-static size_t pipe2position = 0;
+static DspState dsp_state = DspState::Off;
+
+static std::array<std::vector<u8>, static_cast<size_t>(DspPipe::DspPipe_MAX)> pipe_data;
 
 void ResetPipes() {
-    pipe2position = 0;
+    for (auto& data : pipe_data) {
+        data.clear();
+    }
+    dsp_state = DspState::Off;
 }
 
-std::vector<u8> PipeRead(u32 pipe_number, u32 length) {
-    if (pipe_number != 2) {
-        LOG_WARNING(Audio_DSP, "pipe_number = %u (!= 2), unimplemented", pipe_number);
-        return {}; // We currently don't handle anything other than the audio pipe.
-    }
-
-    // Canned DSP responses that games expect. These were taken from HW by 3dmoo team.
-    // TODO: Our implementation will actually use a slightly different response than this one.
-    // TODO: Use offsetof on DSP structures instead for a proper response.
-    static const std::array<u8, 32> canned_response {{
-        0x0F, 0x00, 0xFF, 0xBF, 0x8E, 0x9E, 0x80, 0x86, 0x8E, 0xA7, 0x30, 0x94, 0x00, 0x84, 0x40, 0x85,
-        0x8E, 0x94, 0x10, 0x87, 0x10, 0x84, 0x0E, 0xA9, 0x0E, 0xAA, 0xCE, 0xAA, 0x4E, 0xAC, 0x58, 0xAC
-    }};
-
-    // TODO: Move this into dsp::DSP service since it happens on the service side.
-    // Hardware observation: No data is returned if requested length reads beyond the end of the data in-pipe.
-    if (pipe2position + length > canned_response.size()) {
+std::vector<u8> PipeRead(DspPipe pipe_number, u32 length) {
+    if (pipe_number >= DspPipe::DspPipe_MAX) {
+        LOG_ERROR(Audio_DSP, "pipe_number = %u invalid", pipe_number);
         return {};
     }
 
-    std::vector<u8> ret;
-    for (size_t i = 0; i < length; i++, pipe2position++) {
-        ret.emplace_back(canned_response[pipe2position]);
+    std::vector<u8>& data = pipe_data[static_cast<size_t>(pipe_number)];
+
+    if (length > data.size()) {
+        LOG_WARNING(Audio_DSP, "pipe_number = %u is out of data, application requested read of %u but %zu remain",
+                    pipe_number, length, data.size());
+        length = data.size();
     }
 
+    if (length == 0)
+        return {};
+
+    std::vector<u8> ret(data.begin(), data.begin() + length);
+    data.erase(data.begin(), data.begin() + length);
     return ret;
 }
 
-void PipeWrite(u32 pipe_number, const std::vector<u8>& buffer) {
-    // TODO: proper pipe behaviour
+size_t GetPipeReadableSize(DspPipe pipe_number) {
+    if (pipe_number >= DspPipe::DspPipe_MAX) {
+        LOG_ERROR(Audio_DSP, "pipe_number = %u invalid", pipe_number);
+        return 0;
+    }
+
+    return pipe_data[static_cast<size_t>(pipe_number)].size();
+}
+
+static void WriteU16(DspPipe pipe_number, u16 value) {
+    std::vector<u8>& data = pipe_data[static_cast<size_t>(pipe_number)];
+    // Little endian
+    data.emplace_back(value & 0xFF);
+    data.emplace_back(value >> 8);
+}
+
+static void AudioPipeWriteStructAddresses() {
+    // These struct addresses are DSP dram addresses.
+    // See also: DSP_DSP::ConvertProcessAddressFromDspDram
+    static const std::array<u16, 15> struct_addresses = {
+        0x8000 + offsetof(SharedMemory, frame_counter) / 2,
+        0x8000 + offsetof(SharedMemory, source_configurations) / 2,
+        0x8000 + offsetof(SharedMemory, source_statuses) / 2,
+        0x8000 + offsetof(SharedMemory, adpcm_coefficients) / 2,
+        0x8000 + offsetof(SharedMemory, dsp_configuration) / 2,
+        0x8000 + offsetof(SharedMemory, dsp_status) / 2,
+        0x8000 + offsetof(SharedMemory, final_samples) / 2,
+        0x8000 + offsetof(SharedMemory, intermediate_mix_samples) / 2,
+        0x8000 + offsetof(SharedMemory, compressor) / 2,
+        0x8000 + offsetof(SharedMemory, dsp_debug) / 2,
+        0x8000 + offsetof(SharedMemory, unknown10) / 2,
+        0x8000 + offsetof(SharedMemory, unknown11) / 2,
+        0x8000 + offsetof(SharedMemory, unknown12) / 2,
+        0x8000 + offsetof(SharedMemory, unknown13) / 2,
+        0x8000 + offsetof(SharedMemory, unknown14) / 2
+    };
+
+    // Begin with a u16 denoting the number of structs.
+    WriteU16(DspPipe::Audio, struct_addresses.size());
+    // Then write the struct addresses.
+    for (u16 addr : struct_addresses) {
+        WriteU16(DspPipe::Audio, addr);
+    }
+}
+
+void PipeWrite(DspPipe pipe_number, const std::vector<u8>& buffer) {
+    switch (pipe_number) {
+    case DspPipe::Audio: {
+        if (buffer.size() != 4) {
+            LOG_ERROR(Audio_DSP, "DspPipe::Audio: Unexpected buffer length %zu was written", buffer.size());
+            return;
+        }
+
+        enum class StateChange {
+            Initalize = 0,
+            Shutdown = 1,
+            Wakeup = 2,
+            Sleep = 3
+        };
+
+        // The difference between Initialize and Wakeup is that Input state is maintained
+        // when sleeping but isn't when turning it off and on again. (TODO: Implement this.)
+        // Waking up from sleep garbles some of the structs in the memory region. (TODO:
+        // Implement this.) Applications store away the state of these structs before
+        // sleeping and reset it back after wakeup on behalf of the DSP.
+
+        switch (static_cast<StateChange>(buffer[0])) {
+        case StateChange::Initalize:
+            LOG_INFO(Audio_DSP, "Application has requested initialization of DSP hardware");
+            ResetPipes();
+            AudioPipeWriteStructAddresses();
+            dsp_state = DspState::On;
+            break;
+        case StateChange::Shutdown:
+            LOG_INFO(Audio_DSP, "Application has requested shutdown of DSP hardware");
+            dsp_state = DspState::Off;
+            break;
+        case StateChange::Wakeup:
+            LOG_INFO(Audio_DSP, "Application has requested wakeup of DSP hardware");
+            ResetPipes();
+            AudioPipeWriteStructAddresses();
+            dsp_state = DspState::On;
+            break;
+        case StateChange::Sleep:
+            LOG_INFO(Audio_DSP, "Application has requested sleep of DSP hardware");
+            UNIMPLEMENTED();
+            dsp_state = DspState::Sleeping;
+            break;
+        default:
+            LOG_ERROR(Audio_DSP, "Application has requested unknown state transition of DSP hardware %hhu", buffer[0]);
+            dsp_state = DspState::Off;
+            break;
+        }
+
+        return;
+    }
+    default:
+        LOG_CRITICAL(Audio_DSP, "pipe_number = %u unimplemented", pipe_number);
+        UNIMPLEMENTED();
+        return;
+    }
+}
+
+DspState GetDspState() {
+    return dsp_state;
 }
 
 } // namespace HLE
--- a/src/audio_core/hle/pipe.h
+++ b/src/audio_core/hle/pipe.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <cstddef>
 #include <vector>
 
 #include "common/common_types.h"
@@ -14,25 +15,43 @@ namespace HLE {
 /// Reset the pipes by setting pipe positions back to the beginning.
 void ResetPipes();
 
+enum class DspPipe {
+    Debug = 0,
+    Dma = 1,
+    Audio = 2,
+    Binary = 3,
+    DspPipe_MAX
+};
+
 /**
  * Read a DSP pipe.
- * Pipe IDs:
- *   pipe_number = 0: Debug
- *   pipe_number = 1: P-DMA
- *   pipe_number = 2: Audio
- *   pipe_number = 3: Binary
  * @param pipe_number The Pipe ID
  * @param length How much data to request.
  * @return The data read from the pipe. The size of this vector can be less than the length requested.
  */
-std::vector<u8> PipeRead(u32 pipe_number, u32 length);
+std::vector<u8> PipeRead(DspPipe pipe_number, u32 length);
+
+/**
+ * How much data is left in pipe
+ * @param pipe_number The Pipe ID
+ * @return The amount of data remaning in the pipe. This is the maximum length PipeRead will return.
+ */
+size_t GetPipeReadableSize(DspPipe pipe_number);
 
 /**
  * Write to a DSP pipe.
  * @param pipe_number The Pipe ID
  * @param buffer The data to write to the pipe.
  */
-void PipeWrite(u32 pipe_number, const std::vector<u8>& buffer);
+void PipeWrite(DspPipe pipe_number, const std::vector<u8>& buffer);
+
+enum class DspState {
+    Off,
+    On,
+    Sleeping
+};
+/// Get the state of the DSP
+DspState GetDspState();
 
 } // namespace HLE
 } // namespace DSP
--- a/src/core/hle/service/dsp_dsp.cpp
+++ b/src/core/hle/service/dsp_dsp.cpp
@@ -58,10 +58,10 @@ static void ConvertProcessAddressFromDspDram(Service::Interface* self) {
 
     u32 addr = cmd_buff[1];
 
-    cmd_buff[1] = 0; // No error
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
     cmd_buff[2] = (addr << 1) + (Memory::DSP_RAM_VADDR + 0x40000);
 
-    LOG_TRACE(Service_DSP, "addr=0x%08X", addr);
+    LOG_DEBUG(Service_DSP, "addr=0x%08X", addr);
 }
 
 /**
@@ -142,8 +142,7 @@ static void FlushDataCache(Service::Interface* self) {
 
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
 
-    LOG_DEBUG(Service_DSP, "(STUBBED) called address=0x%08X, size=0x%X, process=0x%08X",
-              address, size, process);
+    LOG_TRACE(Service_DSP, "called address=0x%08X, size=0x%X, process=0x%08X", address, size, process);
 }
 
 /**
@@ -167,14 +166,14 @@ static void RegisterInterruptEvents(Service::Interface* self) {
         if (evt) {
             interrupt_events[std::make_pair(interrupt, channel)] = evt;
             cmd_buff[1] = RESULT_SUCCESS.raw;
-            LOG_WARNING(Service_DSP, "Registered interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
+            LOG_INFO(Service_DSP, "Registered interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
         } else {
-            cmd_buff[1] = -1;
-            LOG_ERROR(Service_DSP, "Invalid event handle! interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
+            LOG_CRITICAL(Service_DSP, "Invalid event handle! interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
+            ASSERT(false); // This should really be handled at a IPC translation layer.
         }
     } else {
         interrupt_events.erase(std::make_pair(interrupt, channel));
-        LOG_WARNING(Service_DSP, "Unregistered interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
+        LOG_INFO(Service_DSP, "Unregistered interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
     }
 }
 
@@ -188,7 +187,7 @@ static void RegisterInterruptEvents(Service::Interface* self) {
 static void SetSemaphore(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    cmd_buff[1] = 0; // No error
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
 
     LOG_WARNING(Service_DSP, "(STUBBED) called");
 }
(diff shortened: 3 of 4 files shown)


src/core/hle/service/dsp_dsp.cpp
```diff
@@ -246,19 +246,86 @@ static void ReadPipeIfPossible(Service::Interface* self) {
...
+        return;
+    }
+
+    if (DSP::HLE::GetPipeReadableSize(pipe) >= size) {
+        std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
+
+        Memory::WriteBlock(addr, response.data(), response.size());
+
+        cmd_buff[1] = 0; // No error
+        cmd_buff[2] = (u32)response.size();
```
lioncash: static_cast

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -246,19 +246,86 @@ static void ReadPipeIfPossible(Service::Interface* self) {
...
     }
 
-    std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
+    if (DSP::HLE::GetPipeReadableSize(pipe) >= size) {
+        std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
+
+        Memory::WriteBlock(addr, response.data(), response.size());
+
+        cmd_buff[1] = 0; // No error
+        cmd_buff[2] = (u32)response.size();
```
lioncash: static_cast

src/audio_core/hle/pipe.h
```diff
@@ -28,11 +28,21 @@ void ResetPipes();
 std::vector<u8> PipeRead(u32 pipe_number, u32 length);
 
 /**
+ * How much data is left in pipe
+ * @param pipe_number The Pipe ID
+ * @return The amoutn of data remaning in the pipe. This is the maximum length PipeRead will return.
```
lioncash: `amount`

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -246,19 +246,86 @@ static void ReadPipeIfPossible(Service::Interface* self) {
...
+        cmd_buff[1] = -1;
         return;
     }
 
-    std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
+    if (DSP::HLE::GetPipeReadableSize(pipe) >= size) {
+        std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
+
+        Memory::WriteBlock(addr, response.data(), response.size());
+
```
Subv: Use RESULT_SUCCESS

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -293,12 +360,66 @@ static void GetHeadphoneStatus(Service::Interface* self) {
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
     cmd_buff[2] = 0; // Not using headphones?
 
-    LOG_DEBUG(Service_DSP, "(STUBBED) called");
+    LOG_WARNING(Service_DSP, "(STUBBED) called");
+}
+
+/**
+ * DSP_DSP::RecvData service function
+ *      This function reads a value out of a register.
```
Subv: "Out of a DSP register"

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -293,12 +360,66 @@ static void GetHeadphoneStatus(Service::Interface* self) {
...
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : non-zero == ready
+ */
+static void RecvDataIsReady(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 register_number = cmd_buff[1];
+
+    if (register_number == 0) {
```
Subv: Leave a comment here about this zero

Subv:
LGTM after my comments are addressed

MerryMage:
Updated to address comments above.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -246,19 +246,86 @@ static void ReadPipeIfPossible(Service::Interface* self) {
...
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 pipe = cmd_buff[1];
+    u32 unk2 = cmd_buff[2];
+    u32 size = cmd_buff[3] & 0xFFFF;// Lower 16 bits are size
+    VAddr addr = cmd_buff[0x41];
+
+    if (!Memory::GetPointer(addr)) {
+        LOG_CRITICAL(Service_DSP, "Invalid addr: pipe=0x%08X, unk2=0x%08X, size=0x%X, buffer=0x%08X", pipe, unk2, size, addr);
+        cmd_buff[1] = -1;
```
Subv: What does this return on real hardware?

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -246,19 +246,86 @@ static void ReadPipeIfPossible(Service::Interface* self) {
...
+    if (DSP::HLE::GetPipeReadableSize(pipe) >= size) {
+        std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
+
+        Memory::WriteBlock(addr, response.data(), response.size());
+
+        cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+        cmd_buff[2] = static_cast<u32>(response.size());
+    } else {
+        LOG_CRITICAL(Service_DSP, "No more data is in pipe. (Hardware hangs here.)");
+        cmd_buff[1] = -1;
```
Subv: Add an assert or something here

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -293,12 +360,67 @@ static void GetHeadphoneStatus(Service::Interface* self) {
...
+
+    if (register_number == 0) {
+        // This is read to verify DSP shutdown.
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+        if (DSP::HLE::IsActivated())
+            cmd_buff[2] = 0;
+        else
+            cmd_buff[2] = 1;
+    } else {
+        LOG_ERROR(Service_DSP, "Unimplemented register_number=%u", register_number);
```
Subv: You should set the error code here

Subv:
Sorry! I added some more comments :x

src/core/hle/service/dsp_dsp.cpp
MerryMage: _Goes of to verify._ Hangs on hardware. Should probably be an assert instead.

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,117 @@
...
 #include "common/logging/log.h"
 
 namespace DSP {
 namespace HLE {
 
+static bool is_active = false;
+
 static size_t pipe2position = 0;
 
+/*
```
bunnei: This isn't a docstring, so just use `//` format. Total nitpick, but might as well be consistent!

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,117 @@
...
+    0x8000 + offsetof(SharedMemory, compressor) / 2,
+    0x8000 + offsetof(SharedMemory, dsp_debug) / 2,
+    0x8000 + offsetof(SharedMemory, unknown10) / 2,
+    0x8000 + offsetof(SharedMemory, unknown11) / 2,
+    0x8000 + offsetof(SharedMemory, unknown12) / 2,
+    0x8000 + offsetof(SharedMemory, unknown13) / 2,
+    0x8000 + offsetof(SharedMemory, unknown14) / 2
+};
+
 void ResetPipes() {
```
bunnei: Should this also be called at initialization to ensure no state leaks from a previous emulation session?

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,117 @@
...
+    is_active = false;
 }
 
 std::vector<u8> PipeRead(u32 pipe_number, u32 length) {
-    if (pipe_number != 2) {
-        LOG_WARNING(Audio_DSP, "pipe_number = %u (!= 2), unimplemented", pipe_number);
-        return {}; // We currently don't handle anything other than the audio pipe.
-    }
+    switch (pipe_number) {
+    case 2: {
```
bunnei: Is there any known significant to pipe 2? If so, I feel like enumeration makes code like this seem less arbitrary

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,117 @@
...
+        const static size_t max_end = sizeof(pipe2payload); ///< Length of the entire response in bytes
+        const static u8* entire_response_bytes = reinterpret_cast<const u8*>(pipe2payload.data());
 
-    // Canned DSP responses that games expect. These were taken from HW by 3dmoo team.
-    // TODO: Our implementation will actually use a slightly different response than this one.
-    // TODO: Use offsetof on DSP structures instead for a proper response.
-    static const std::array<u8, 32> canned_response {{
```
bunnei: :cake::fireworks: 
This is an old one that served us well! Nonetheless nice to see this hack go!

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,117 @@
...
+        size_t end = pipe2position + length;
+        if (end > max_end) end = max_end;
+
+        pipe2position = end;
+        ASSERT(pipe2position <= max_end);
+
+        return std::vector<u8>(entire_response_bytes + start, entire_response_bytes + end);
+    }
+    default:
+        LOG_ERROR(Audio_DSP, "pipe_number = %u unimplemented", pipe_number);
```
bunnei: <pre>
LOG_CRITICAL(...);
UNIMPLEMENTED();
</pre>

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,117 @@
...
 
-    std::vector<u8> ret;
-    for (size_t i = 0; i < length; i++, pipe2position++) {
-        ret.emplace_back(canned_response[pipe2position]);
+size_t GetPipeReadableSize(u32 pipe_number) {
+    switch (pipe_number) {
+    case 2:
+        return sizeof(pipe2payload) - pipe2position;
+    default:
+        LOG_ERROR(Audio_DSP, "pipe_number = %u unimplemented", pipe_number);
```
bunnei: Ditto w/ the above

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,117 @@
...
+        default:
+            LOG_ERROR(Audio_DSP, "Application has requested transition of DSP hardware into unknown state %hhu", buffer[0]);
+            is_active = false;
+            break;
+        }
+
+        return;
+    }
+    default:
+        LOG_ERROR(Audio_DSP, "pipe_number = %u unimplemented", pipe_number);
```
bunnei: Ditto w/ the above

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,117 @@
...
 void PipeWrite(u32 pipe_number, const std::vector<u8>& buffer) {
-    // TODO: proper pipe behaviour
+    switch (pipe_number) {
+    case 2: {
+        if (buffer.size() != 4) {
+            LOG_ERROR(Audio_DSP, "Pipe 2: Unexpected buffer length %zu was written", buffer.size());
+            return;
+        }
+
+        switch (buffer[0]) {
```
bunnei: Use an enum

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -129,7 +129,7 @@ static void FlushDataCache(Service::Interface* self) {
 
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
 
-    LOG_DEBUG(Service_DSP, "(STUBBED) called address=0x%08X, size=0x%X, process=0x%08X",
+    LOG_TRACE(Service_DSP, "(STUBBED) called address=0x%08X, size=0x%X, process=0x%08X",
```
bunnei: Hmm? Stubbed functions should log at the `LOG_WARNING` level

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -293,12 +348,69 @@ static void GetHeadphoneStatus(Service::Interface* self) {
...
+    if (register_number == 0) {
+        // This is read to verify DSP shutdown.
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+        if (DSP::HLE::IsActivated())
+            cmd_buff[2] = 0;
+        else
+            cmd_buff[2] = 1;
+    } else {
+        LOG_ERROR(Service_DSP, "Unimplemented register_number=%u", register_number);
+        cmd_buff[1] = -1;
```
bunnei: Is this the correct error code?

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -245,20 +241,79 @@ static void ReadPipeIfPossible(Service::Interface* self) {
...
+    if (DSP::HLE::GetPipeReadableSize(pipe) >= size) {
+        std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
 
-    LOG_TRACE(Service_DSP, "pipe=0x%08X, unk2=0x%08X, size=0x%X, buffer=0x%08X", pipe, unk2, size, addr);
+        Memory::WriteBlock(addr, response.data(), response.size());
+
+        cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+        cmd_buff[2] = static_cast<u32>(response.size());
+    } else {
+        ASSERT_MSG(0, "No more data is in pipe. (Hardware hangs here.)");
```
bunnei: If this is something that should never happen, you can just replace these three lines with `UNREACHABLE();` (and put the text in a comment)

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -245,20 +241,79 @@ static void ReadPipeIfPossible(Service::Interface* self) {
...
+}
 
-    Memory::WriteBlock(addr, response.data(), response.size());
+/**
+ * DSP_DSP::ReadPipe service function
+ *  Inputs:
+ *      1 : Pipe Number
+ *      2 : Unknown
+ *      3 : Size in bytes of read (observed only lower half word used)
+ *      0x41 : Virtual address to read from DSP pipe to in memory
```
bunnei: *Virtual address to read from DSP pipe in memory

vgturtle127: @bunnei Wouldn't it be "Virtual address to read from DSP pipe into memory"?

src/audio_core/hle/pipe.cpp
JamePeng: @bunnei 
http://3dbrew.org/wiki/DSP_Memory_Region

src/core/hle/service/dsp_dsp.cpp
MerryMage: FlushDataCache is called quite frequently, it would likely overwhelm the logs.

src/audio_core/hle/pipe.cpp
MerryMage: @bunnei [3dbrew](https://www.3dbrew.org/wiki/DSP:ReadPipeIfPossible) claims there are different purposes for each pipe, though I've not actually been able to read from them in my tests (maybe I need better tests).

In any case, I should probably have an `enum class DspPipe` somewhere.

src/audio_core/hle/pipe.cpp
MerryMage: Yes, [it is](https://github.com/citra-emu/citra/blob/af7282b5ea8ed3ebfb47f35f6ac8212cfcb4c4fd/src/audio_core/hle/dsp.cpp#L14).

src/audio_core/hle/pipe.cpp
Subv: Yay!

src/core/hle/service/dsp_dsp.cpp
bunnei: Oops yeah it seems so. Either way, the wording is kind of awkward

MerryMage: ... if I wrote that I was probably half-asleep. "Virtual address of memory buffer to write pipe contents to"?

src/audio_core/hle/pipe.cpp
MerryMage: Nitpicking is good!

MerryMage:
Updated to incorporate bunnei's comments.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -240,25 +230,83 @@ static void WriteProcessPipe(Service::Interface* self) {
...
+
+    if (DSP::HLE::GetPipeReadableSize(pipe) >= size) {
+        std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
+
+        Memory::WriteBlock(addr, response.data(), response.size());
+
+        cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+        cmd_buff[2] = static_cast<u32>(response.size());
+    } else {
+        cmd_buff[1] = RESULT_SUCCESS.raw; // No error
```
bunnei: This can be set outside the if-else, since you're doing the same thing

bunnei:
LGTM

src/core/hle/service/dsp_dsp.cpp
JamePeng: `if (DSP::HLE::GetPipeReadableSize(pipe) >= size) {
+        std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
+
+        Memory::WriteBlock(addr, response.data(), response.size());
+        cmd_buff[2] = static_cast<u32>(response.size());
+    } 
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error`

MerryMage: Thanks, @JamePeng.

MerryMage:
Squashed.

JamePeng:
LGTM , I think it can fix the METAL MAX 4 dsp problem!

src/core/hle/service/dsp_dsp.cpp
vgturtle127: @MerryMage @bunnei Yeah, I was just making sure I was understanding it the same way everybody else was. :laughing:

yuriks:
I don't understand why the pipes were implemented the way they were. Is there any behavior that prevents them from being implemented as a generic FIFO (using a vector or array to hold buffered data) which you can just write and read data to? Seems like that should simplify the code as a whole.

MerryMage:
Simplified code by implementing as FIFO as @yuriks suggested.

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,146 @@
...
 #include "common/common_types.h"
 #include "common/logging/log.h"
 
 namespace DSP {
 namespace HLE {
 
-static size_t pipe2position = 0;
+static bool is_active = false;
+
+static std::array<std::vector<u8>, static_cast<size_t>(DspPipe::DspPipe_MAX)> pipe_data;
```
Subv: This line looks weird, is the cast really needed?

JamePeng:
qq 20160306053020

test by METAL MAX 4 , look like in a dead roll !

src/audio_core/hle/pipe.cpp
MerryMage: Yes, a cast is required. An enum class does not have an implicit conversion to an integer. Type safety.

MerryMage:
@JamePeng, a complete log would be more helpful, but I suspect it's running a different version of the firmware.

On that note, printing a firmware hash to the console would be helpful. I'll implement that in a later PR.

JamePeng:
@MerryMage 
qq 20160306151034

log.txt:
[log.txt](https://github.com/citra-emu/citra/files/160214/log.txt)

MerryMage:
Rebased against master and partially implement sleep and wakeup.

Can't fully implement because the actual functionality of the DSP isn't implemented yet. The difference between an initialization and a sleep transtition is whether or not the internal state of the DSP is mantained across the transition. With sleep, (some) state is maintained.

MerryMage:
@JamePeng I believe that's more likely to be a issue with `PrepareToDoApplicationJump` not being implemented. I have however partially implemented the sleep state change which the application seems to expect, above. I don't expect this to help much.

Subv:
```
[ 107.599952] Service.FS <Error> core\hle\service\fs\fs_user.cpp:Service::FS::OpenFileDirectly:119: failed to get a handle for archive archive_id=0x2345678A archive_path=[Binary: 00fd0a008c00040001ffffff01000000]
```

That's most likely the problem with that game, PrepareToDoApplicationJump is usually called when trying to go back to the Home Menu (The game has crashed)

JamePeng:
The pipe look like have a good sleep! :p

src/audio_core/hle/pipe.h
```diff
@@ -14,25 +14,43 @@ namespace HLE {
...
  */
-std::vector<u8> PipeRead(u32 pipe_number, u32 length);
+std::vector<u8> PipeRead(DspPipe pipe_number, u32 length);
+
+/**
+ * How much data is left in pipe
+ * @param pipe_number The Pipe ID
+ * @return The amount of data remaning in the pipe. This is the maximum length PipeRead will return.
+ */
+std::size_t GetPipeReadableSize(DspPipe pipe_number);
```
yuriks: We use just `size_t` without the std prefix most of the time. (And so did you in the function implementation, so just, consistency I guess :D)

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -167,14 +166,14 @@ static void RegisterInterruptEvents(Service::Interface* self) {
         if (evt) {
             interrupt_events[std::make_pair(interrupt, channel)] = evt;
             cmd_buff[1] = RESULT_SUCCESS.raw;
-            LOG_WARNING(Service_DSP, "Registered interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
+            LOG_INFO(Service_DSP, "Registered interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
         } else {
-            cmd_buff[1] = -1;
-            LOG_ERROR(Service_DSP, "Invalid event handle! interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
+            LOG_CRITICAL(Service_DSP, "Invalid event handle! interrupt=%u, channel=%u, event_handle=0x%08X", interrupt, channel, event_handle);
+            UNREACHABLE(); // This should really be handled at a IPC translation layer.
```
yuriks: This isn't really unreachable, we use that macro for things where reaching the branch means a bug somewhere (not simply the application sending us garbage cmdbufs) so I think I'd changed this to an assert(false) or similar.

yuriks: (Though yes, the IPC layer should catch this.)

yuriks:
L~G~T~M

src/audio_core/hle/pipe.h
MerryMage: Bad habit of mine~

MerryMage:
Squashed!

src/audio_core/hle/pipe.cpp
```diff
@@ -5,50 +5,154 @@
...
+void PipeWrite(DspPipe pipe_number, const std::vector<u8>& buffer) {
+    switch (pipe_number) {
+    case DspPipe::Audio: {
+        if (buffer.size() != 4) {
+            LOG_ERROR(Audio_DSP, "DspPipe::Audio: Unexpected buffer length %zu was written", buffer.size());
+            return;
+        }
+
+        enum class StateChange {
+            Initalize = 0,
```
lioncash: I thing {0-3} can be omitted if it's an `enum class`, since the value isn't directly important.

yuriks: These interoperate with the 3DS though, so we do care about their specific values.

MerryMage: Not in this case as the values come from the application. Should it be an `enum` instead?

lioncash: @yuriks the point is since it's strongly typed, it doesn't matter if it's explicit. These are handled by regular enum semantics.

mailwl:
@MerryMage, some game waits the data with exact size from the dsp in function
`static void ReadPipeIfPossible(Service::Interface* self)`
in other way they just fall in endless loop waiting the data (i saw some, but remember only one - my lovely Tales of The Abyss). What way you see to fix it?

MerryMage:
@mailwl Some logs would be nice, but if it's failing on pipe 2, do note I currently only return the correct information for one version of the DSP firmware at the moment.
If it is that issue the way to fix that is to hash the firmware and change the response based on that.

If it's not pipe 2, then I would love to know what titles they are, because I really want to get some good info on the other pipes.

mailwl:
no, it for binary (3). I think, it's related sound too, but crossplatform way
the game ToTA is really weird, because it opens many hidden errors in Citra, because uses not standart 3ds api usages (from it's not 3ds nature).
Merry, did you think about actual loading dsp firm in `static void LoadComponent(Service::Interface* self)`? Maybe it contains some useful data? file format: https://www.3dbrew.org/wiki/DSP_Binary
 long tine ago i write such fucntion, but now it not working

```
struct SegmentRecord {
    u32 offset_data;
    u32 start_address;
    u32 size;
    u8  pad[3];
    u8  memory_type;
    u8  sha_hash[0x20];
};

struct DSP_Header {
    u8  rsa[0x100];
    u32 magic; // 'DSP1'
    u32 dsp_binary_size;
    u16 memory_layout;
    u16 pad;
    u8  unk1;
    u8  special_type;
    u8  num_segment;
    u8  flags;
    u32 special_segment_start;
    u32 special_segment_size;
    u64 zero;
    SegmentRecord segments[10];
};

static bool LoadDSPComponent(u8* address, u32 size) {
    DSP_Header* header = reinterpret_cast<DSP_Header*>(address);
    if (header->magic != 0x31505344) {
        return false;
    }
    for (u8 i = 0; i<header->num_segment; ++i) {
        const SegmentRecord& sr = header->segments[i];
        switch (sr.memory_type) {
        case 0:
        case 1:
            //memcpy(Memory::GetPhysicalPointer(Memory::DSP_RAM_VADDR) + sr.start_address * 2,
            //    address + sr.offset_data, sr.size);
            break;
        case 2:
            //memcpy(Memory::GetPhysicalPointer(Memory::DSP_RAM_VADDR) + sr.start_address * 2 + 0x40000,
            //    address + sr.offset_data, sr.size);
            break;
        }
    }
    FILE* dsp = fopen("dsp.bin", "wb");
    if(dsp) {
        fwrite(address, 1, size, dsp);
        fclose(dsp);
    }
    return true;
}
/**
 * DSP_DSP::LoadComponent service function
 *  Inputs:
 *      1 : Size
 *      2 : Program mask (observed only half word used)
 *      3 : Data mask (observed only half word used)
 *      4 : (size << 4) | 0xA
 *      5 : Buffer address
 *  Outputs:
 *      1 : Result of function, 0 on success, otherwise error code
 *      2 : Component loaded, 0 on not loaded, 1 on loaded
 */
static void LoadComponent(Service::Interface* self) {

    u32 size       = cmd_buff[1];
    u32 prog_mask  = cmd_buff[2];
    u32 data_mask  = cmd_buff[3];
    u32 desc       = cmd_buff[4];
    u32 buffer     = cmd_buff[5];

    cmd_buff[0] = IPC::MakeHeader(0x11, 2, 2);
    cmd_buff[2] = 1; // Pretend that we actually loaded the DSP firmware
    cmd_buff[3] = desc;
    cmd_buff[4] = buffer;

    // TODO(bunnei): Implement real DSP firmware loading

    ASSERT(Memory::GetPointer(buffer) != nullptr);
    ASSERT(size > 0x37C);

    LoadDSPComponent(Memory::GetPointer(buffer), size);

    LOG_INFO(Service_DSP, "Firmware hash: %#" PRIx64, Common::ComputeHash64(Memory::GetPointer(buffer), size));
    // Some versions of the firmware have the location of DSP structures listed here.
    LOG_INFO(Service_DSP, "Structures hash: %#" PRIx64, Common::ComputeHash64(Memory::GetPointer(buffer) + 0x340, 60));

    LOG_WARNING(Service_DSP, "(STUBBED) called size=0x%X, prog_mask=0x%08X, data_mask=0x%08X, buffer=0x%08X",
                size, prog_mask, data_mask, buffer);
}
```

MerryMage:
@mailwl Ooh, binary pipe. I'll take a closer look, thanks!

MerryMage:
@mailwl Had a look. Here's something which implements the binary pipe's behaviour on ToTA: https://github.com/MerryMage/citra/tree/pipe3

I feel this is out of the scope of this PR and have not added this commit to this PR.

I have also observed that the behavior of this pipe varies depending on DSP firmware version. On that note, I shall focus on fully implementing one version of the firmware before I take any differences into account, thus will not include this in the near future.

mailwl:
@MerryMage you my hero :)
now it's possible to debug the game in IDAPro

PR #1446, opened 2016-03-03
Update Building Links

vitorsilverio:
Links on README.md don't exist.


lioncash:
Whoops, thanks for catching those.

Issue #1449, opened 2016-03-04
Monster Hunter 4G  crash error :Unsupported thread processor ID: 2


JamePeng:
![qq 20160304155141](https://cloud.githubusercontent.com/assets/17095606/13521085/ee9bb620-e220-11e5-8d62-7feb2a256e46.png)

I think we should add the THREADPROCESSORID_MAX into SVC::CreateThread

PTM::CheckNew3DS will return the machine type(u8 output: 0 = Old3DS, 1 = New3DS.) 
so, if you select the New3DS,it will cause a assert failed!

JamePeng:
#1520

subzerohurley:
wait so can someone help me out? explain where i i do what?

Andrix44:
Shouldn't this be closed? Same with #840.

JayFoxRox:
Sounds okay to me, we'll reopen if this is still a problem. Closed.

PR #1450, opened 2016-03-04
Add THREADPROCESSORID_MAX on SVC::CreateThread

JamePeng:
#1449 
PTM::CheckNew3DS will return the machine type(u8 output: 0 = Old3DS, 1 = New3DS.) 
so, if you select the New3DS,it will cause a assert failed!


Subv:
:-1: The ProcessorId enum should be changed to reflect the new CPUs in the n3DS

JamePeng:
@

JamePeng:
@Subv No,the crash rely on "CheckNew3DSApp"，if you select New3DS status. 
The crash still happen!

Subv:
What? How is that related to what i said?

bunnei:
@JamePeng I'm going to close this one... This is not the right way to fix this issue. THREADPROCESSORID_MAX is meant for doing bounds checking on valid processor IDs for O3DS, it itself was not meant to be used as a valid ID.

PR #1455, opened 2016-03-05
core: Use unrestricted union to hold storage of ResultVal value

yuriks:
Yay for VS2015!


Subv:
What's the difference between the previous code and this one?

yuriks:
Functionally, none, it's just a little neater.

Subv:
I have come back after reading up a bit on unrestricted unions.
 LGTM

PR #1459, opened 2016-03-05
DSP: Print hash of firmware to console

MerryMage:
It would be useful to know what version of the DSP firmware is used by an application when receiving a bug report.

### DIFF
--- a/src/core/hle/service/dsp_dsp.cpp
+++ b/src/core/hle/service/dsp_dsp.cpp
@@ -2,8 +2,11 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <cinttypes>
+
 #include "audio_core/hle/pipe.h"
 
+#include "common/hash.h"
 #include "common/logging/log.h"
 
 #include "core/hle/kernel/event.h"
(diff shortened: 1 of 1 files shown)


src/core/hle/service/dsp_dsp.cpp
```diff
@@ -77,18 +80,25 @@ static void LoadComponent(Service::Interface* self) {
...
     u32 size       = cmd_buff[1];
-    u32 unk1       = cmd_buff[2];
-    u32 unk2       = cmd_buff[3];
-    u32 new_size   = cmd_buff[4];
+    u32 prog_mask  = cmd_buff[2];
+    u32 data_mask  = cmd_buff[3];
     u32 buffer     = cmd_buff[5];
 
-    cmd_buff[1] = 0; // No error
+    cmd_buff[0] = 0x00110082;
```
Subv: Please use IPC::MakeHeader

MerryMage: Thanks! Corrected.

Subv:
LGTM

yuriks:
LGTM

PR #1460, opened 2016-03-06
OpenGL: Don't attempt to draw empty triangle batches

### LINKED ISSUE #1251: Intel (and nvidia) driver crash when loading Steel Diver Sub Wars
As of 78b0d9c8b0279cc107c4b82e78e1e6b82a6edcb3 citra crashes the Intel driver a few moments after Steel Diver Sub Wars is loaded up, right after one of the glDrawArrays calls.
The nvidia driver keeps going but ultimately crashes citra when trying to exit ("A heap has been corrupted")

Tested on Intel HD 4600
Intel Driver Version: 10.18.15.4256

Attached is an Apitrace of the problem, running this until the last frame will also cause apiretrace to crash
https://drive.google.com/file/d/0B8Y3NfpjZ7WqS0pNZXlWTm1Ya2c/view?usp=sharing

yuriks:
Our code did not handle this well, causing random crashes inside drivers. Fixes #1251.

@Subv please test and confirm.


yuriks:
Does not actually fix #1251 (though it did fix other problems in a different game), so I'll roll this up into another PR instead.

PR #1461, opened 2016-03-06
Immediate-mode improvements

yuriks:
This fixes some issues I missed when reviewing #1394 and also forward-ports some extra features from my old branch:
  - Unifies some duplicated code paths between immediate mode and batch vertex draws.
  - Fixes initialization of the PicaState struct to not memset over PrimitiveAssembler.
  - Adds support for primitive restart.
  - Tweaks and moves a few things around to more closely match behavior from the hardware with unusual register values and command sequences.

cc @ds84182

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -75,12 +75,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
             GSP_GPU::SignalInterrupt(GSP_GPU::InterruptId::P3D);
             break;
 
+        case PICA_REG_INDEX_WORKAROUND(triangle_topology, 0x25E):
+            g_state.primitive_assembler.Reconfigure(regs.triangle_topology);
+            break;
+
+        case PICA_REG_INDEX_WORKAROUND(restart_primitive, 0x25F):
+            g_state.primitive_assembler.Reset();
+            break;
+
         case PICA_REG_INDEX_WORKAROUND(vs_default_attributes_setup.index, 0x232):
-            if (regs.vs_default_attributes_setup.index == 15) {
-                // Reset immediate primitive state
-                g_state.immediate.primitive_assembler.Reconfigure(regs.triangle_topology);
-                g_state.immediate.attribute_id = 0;
-            }
+            g_state.immediate.current_attribute = 0;
+            default_attr_counter = 0;
             break;
 
         // Load default vertex input attributes
@@ -105,7 +110,7 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     break;
                 }
 
-                Math::Vec4<float24>& attribute = g_state.vs.default_attributes[setup.index];
+                Math::Vec4<float24> attribute;
 
                 // NOTE: The destination component order indeed is "backwards"
                 attribute.w = float24::FromRaw(default_attr_write_buffer[0] >> 8);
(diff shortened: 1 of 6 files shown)


yuriks:
This partially fixes Etrian Odyssey 4 (as in, it submits correct vertices) but it still doesn't draw anything on-screen due to an unrelated depth bug.

src/video_core/pica.cpp
```diff
@@ -493,12 +493,25 @@ std::string Regs::GetCommandName(int index) {
...
+    g_state.Reset();
 }
 
 void Shutdown() {
     Shader::Shutdown();
+}
+
+template <typename T>
+void Zero(T& o) {
+    memset(&o, 0, sizeof(o));
```
lioncash: This should also check if T is trivially copyable. Otherwise it's undefined behavior.

src/video_core/command_processor.cpp
```diff
@@ -118,26 +123,29 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
 
-                    if (immediate_attribute_id >= attribute_config.GetNumTotalAttributes()) {
+                    if (immediate_attribute_id >= regs.vs.num_input_attributes+1) {
                         immediate_attribute_id = 0;
 
                         Shader::UnitState<false> shader_unit;
                         Shader::Setup(shader_unit);
 
+                        if (g_debug_context)
+                            g_debug_context->OnEvent(DebugContext::Event::VertexLoaded, (void*)&immediate_input);
```
lioncash: static_cast

src/video_core/pica.cpp
yuriks: I had added a check at first, but couldn't get it to pass with the `regs` struct for some reason. If you can get it working somehow I'll add it.

src/video_core/command_processor.cpp
yuriks: I just moved this line from elsewhere, but ok...

src/video_core/pica.cpp
lioncash: I believe that means the assert is working as intended and the regs struct should be amended.

yuriks:
Also, this PR changes the place where triangle drawing happens in a way that seems to be correct to me but could have unforeseen problems, so I would appreciate some testing in random games to see if any are missing any geometry.

Subv:
cc @JMC47 @JohnGodgames

src/video_core/command_processor.cpp
```diff
@@ -118,26 +123,29 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                    auto& immediate_input = g_state.immediate.input_vertex;
+                    auto& immediate_attribute_id = g_state.immediate.current_attribute;
 
                     immediate_input.attr[immediate_attribute_id++] = attribute;
 
-                    if (immediate_attribute_id >= attribute_config.GetNumTotalAttributes()) {
+                    if (immediate_attribute_id >= regs.vs.num_input_attributes+1) {
                         immediate_attribute_id = 0;
 
                         Shader::UnitState<false> shader_unit;
```
Subv: Shouldn't this shader_unit be an array of 3/4 and inside the Pica::State struct?

yuriks: Later, probably. For now it doesn't make any difference either way since we re-initialized everything before each shader instance.

src/video_core/pica.cpp
MerryMage: Same issue as before, GCC and clang have no issue with trivially copyable in this instance, but MSVC balks.

    template <typename T>
    void Zero(T& o) {
    #if (__GNUC__ >= 5) || defined(__clang__)
        static_assert(std::is_trivially_copyable<T>::value, "T is not trivially copyable");
    #endif
        memset(&o, 0, sizeof(o));
    }

I feel this should be a macro somewhere.

lioncash: @MerryMage Have you checked that clang and GCC's implementations are not buggy? MSVC's is_trivially_copyable implementation adheres more to the the standard in many cases. For example: [volatile](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=63959) would get flagged as trivially copyable by GCC/clang but MSVC would correctly reject it.

Just because one set of compilers accepts it doesn't mean it's correct behavior.

JohnGodgames:
### This PR fixes Crash Time 3D (visible ingame instead of solid color only)
crash time 3d imm2 2

### This PR fixes The Trash Pack
trash imm2

and propably more...

mailwl:
@JohnGodgames, may you test Sims 3? or Barbie Dreamhouse Party?
i remember about graphics problen in this games

JohnGodgames:
Already tested...it's the same.

mailwl:
i see...
just for info for graphics guru citra devs, in leaked sdk there is file named "$CTR_SDK\tools\CommandLineTools\ctr_GxCommandAnalyzer\detail.csv", contains all info about pica registers (i suppose this letters does not belong Nintendo, don't want to violate nintendo rights)

lioncash:
@mailwl [In reference to the removed message] Last warning. You've been told before to not post stuff about the SDK.

mailwl:
@lioncash i don't use it, so i have clear conscience.
but know it has useful info for devs.
well, i will stop talking about it

src/video_core/pica.cpp
MerryMage: By a strict reading of the C++11 standard I believe MSVC's implementation is incorrect. See also [N4148](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4148.html), which proposes to close this loophole.

lioncash: That proposal would actually make BitField non-trivially-copyable because the copy-assignment operator is private (coincidentally, making the copy-assignment operator public will make the static assert pass successfully in MSVC)

JohnGodgames:
Also fixed solid color issues at these games:

Crazy kangaroo
Fireman Sam to The Rescue
Gardenscapes
Chibi Devi
Jewel Match 3 EUR
Penguins of Madagascar
Tiny Games - Knights & Dragons

LGTM

bunnei:
LGTM. Can we get this one wrapped up and merged soon?

yuriks:
Will try to wrap this up tonight.

yuriks:
Rebased and addressed @lioncash's comment.

PR #1463, opened 2016-03-06
Memory: Do correct Phys->Virt address translation for non-APP linheap

yuriks:
cc @Subv


Subv:
LGTM instamerge pls :shipit:

PR #1467, opened 2016-03-06
Set the appropriate locale to get float conversion working using to_string

LittleWhite-tb:
This fixes a bug on Linux (maybe Windows too), on systems where the locale does separate float numbers with ",". Qt through QApplication changes the locale. The new locale affects std::to_string behavior in shaders generation.
The string will be "0,0000" instead of "0.0000" causing numerous problem such as invalid shaders.
My solution is to set back an appropriate locale after QApplication instantiation.

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <clocale>
 #include <thread>
 
 #include <QDesktopWidget>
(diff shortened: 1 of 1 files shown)


JamePeng:
@LittleWhite-tb  
missing    #include <locale.h> or #include <locale>

LittleWhite-tb:
Fixed !

src/citra_qt/main.cpp
```diff
@@ -563,6 +564,9 @@ int main(int argc, char* argv[]) {
     QApplication::setAttribute(Qt::AA_X11InitThreads);
     QApplication app(argc, argv);
 
+    // Qt changes the locale and causes issues in float conversion using std::to_string() when generating shaders
+    setlocale(LC_ALL,"C");
```
lioncash: space after the comma

LittleWhite-tb:
Fixed

mailwl:
don't know, is it good or evil for russian locale? what the difference?

lioncash:
@mailwl Makes sure different locales don't mangle floating point output (and other edge cases). e.g. In some locales `1.444` will print as `1,444`, which will break shader compilation.

mailwl:
Oh! nice when

bunnei:
LGTM

PR #1469, opened 2016-03-06
SDL2: Use pkgconfig

MerryMage:
Instead of using FindSDL2.cmake, we remove that and use pkgconfig instead.

This change also links to SDL2 statically on OS X.


Lectem:
As mentioned on #citra-dev this should be an addition to the current FindSDL2.cmake instead of a replacement, because pkg-config isn't cross-platform.
Your current patch makes it bothersome for Windows users (and those using SDL2DIR envvar).

MerryMage:
Noted! I'll look into alternative solutions for OS X users as well.

MerryMage:
Closing as per above. As for OS X delopyment I'm going to write a deploy script that `install_name_tool`s the dylib appropriately.

PR #1475, opened 2016-03-09
Common: Get rid of alignment macros

lioncash:
The GL rasterizer already uses alignas, so we may as well move everything over.


yuriks:
LGTM

JamePeng:
LGTM , I like the clean code！:p

Subv:
LGTM

PR #1476, opened 2016-03-09
emitter: constexpr/misc changes

lioncash:
Derived partially from an existing pull request that's up for Dolphin. Just figured I'd put it up here as well in case it's desirable.


src/common/x64/emitter.h
```diff
@@ -238,56 +229,60 @@ struct OpArg
...
+constexpr OpArg UImmAuto(u32 imm) {
     return OpArg(imm, imm >= 128 ? SCALE_IMM32 : SCALE_IMM8);
 }
-inline OpArg SImmAuto(s32 imm) {
+constexpr OpArg SImmAuto(s32 imm) {
     return OpArg(imm, (imm >= 128 || imm < -128) ? SCALE_IMM32 : SCALE_IMM8);
 }
 
+template <typename T>
+OpArg ImmPtr(const T* imm)
```
MerryMage: ImmPtr used to be inline, but here it is no longer. Was that intentional?

lioncash: Yes. Inline isn't necessary on function templates that aren't a full specialization.

MerryMage:
LGTM

bunnei:
LGTM

Issue #1504, opened 2016-03-12
Digimon World Re: Digitize Decode - Missing 2D overlays


Romsstar:
Using  fba547b with additional merges

#1302 
#1436 
#1447

I've also got a branch called digimon in my repository with those merged already.

Most 2D graphics of the game are missing/not being rendered correclty. This also includes the font.

The game requires dumped savegames, otherwise it will randomly hang when clicking to skip the video.(Sometimes it will work though)
Attached it here just in case it will be needen:

[000afc00.zip](https://github.com/citra-emu/citra/files/170515/000afc00.zip)

3DS
![top_0008](https://cloud.githubusercontent.com/assets/17484361/13724323/fd1983dc-e881-11e5-9f0f-3c323e225d0f.png)

Citra
![topcitra](https://cloud.githubusercontent.com/assets/17484361/13724325/010ee87e-e882-11e5-8ab7-65ab42af41fa.png)

3DS
![digimon_1_1](https://cloud.githubusercontent.com/assets/17484361/13724263/c04b9658-e880-11e5-9963-e3fc4207421f.png)

Citra
![digimon1](https://cloud.githubusercontent.com/assets/17484361/13724257/b665b3da-e880-11e5-9ff1-2dbdd1bc20bc.png)

3DS
![3ds2](https://cloud.githubusercontent.com/assets/17484361/13724282/34d4bb62-e881-11e5-8be7-956c0aad237b.png)

Citra
![citra2](https://cloud.githubusercontent.com/assets/17484361/13724283/385ee5e6-e881-11e5-9e00-d9f9fe277f23.png)

LOG
http://pastebin.com/7byN7mdc

PR #1496, opened 2016-03-11
Align attribute components

JayFoxRox:
Fixes #1488 and possible a good chunk of others.

Not tested against real hw, not documented on 3dbrew but very very very likely.
Not sure yet wether each component or just each attribute vec is aligned.

Example (Super Smash Bros Demo Characters) of problematic uploads in current master:

Outline:

```
v0.z [Vertex 206], format 0x2, base 0x18201B00, stride 18, element size: 2: 3192.000000
v1.x [Vertex 206], format 0x0, base 0x18201B06, stride 18, element size: 1: -43.000000
v1.y [Vertex 206], format 0x0, base 0x18201B06, stride 18, element size: 1: -118.000000
v1.z [Vertex 206], format 0x0, base 0x18201B06, stride 18, element size: 1: -6.000000
v2.x [Vertex 206], format 0x1, base 0x18201B09, stride 18, element size: 1: 141.000000
v2.y [Vertex 206], format 0x1, base 0x18201B09, stride 18, element size: 1: 83.000000
v2.z [Vertex 206], format 0x1, base 0x18201B09, stride 18, element size: 1: 46.000000
v2.w [Vertex 206], format 0x1, base 0x18201B09, stride 18, element size: 1: 255.000000
v3.x [Vertex 206], format 0x1, base 0x18201B0D, stride 18, element size: 1: 4.000000
v3.y [Vertex 206], format 0x1, base 0x18201B0D, stride 18, element size: 1: 8.000000
v4.x [Vertex 206], format 0x1, base 0x18201B0F, stride 18, element size: 1: 30.000000
v4.y [Vertex 206], format 0x1, base 0x18201B0F, stride 18, element size: 1: 70.000000
```

(v3 are bone indices, v4 are bone weights (30+70 = 100))

Actual model:

```
v0.z [Vertex 234], format 0x2, base 0x1820CCA0, stride 18, element size: 2: 3192.000000
v1.x [Vertex 234], format 0x0, base 0x1820CCA6, stride 18, element size: 1: 43.000000
v1.y [Vertex 234], format 0x0, base 0x1820CCA6, stride 18, element size: 1: 118.000000
v1.z [Vertex 234], format 0x0, base 0x1820CCA6, stride 18, element size: 1: 6.000000
v2.x [Vertex 234], format 0x2, base 0x1820CCA9, stride 18, element size: 2: 18432.000000 << alignment breaks here
[...]

PR #1508, opened 2016-03-13
Respect vs output map

JayFoxRox:
Something along these lines fixes most of #1504

What happens in Digimon?
* vs outmap_total is 5 = 5 output registers used
* vs outmap_mask is 0x2F (you might see where this is going..)
* shader wrote to o0, o1, o2, o3 and o5
* you know where the output semantics for o5 are configured? right.. vs outmap_o4
* said differently: outmap is packed tightly

Therefore I added a second index to count the actual outmap (output_attributes in citra) index.

As the file was already opened I switched the loop indices to unsigned.

I'm not sure about vs_output_total early out and it should be tested on actual HW.
I felt a little safer adding it because I wouldn't want possibly disabled (cause they are not included in output_total) regs overwriting outputs.

Logo
Push my buttons
Pushed me hard
Digimon stuff
Ash and pikachu
The dog still has broken eyes

I didn't know how to walk in the game as no button seemed to work so I didn't test if the 2D characters appear while talking.
There still seem to be some texturing issues but that's probably a seperate issue.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -71,7 +71,7 @@ struct Regs {
     BitField<0, 24, u32> viewport_depth_range; // float24
     BitField<0, 24, u32> viewport_depth_far_plane; // float24
 
-    INSERT_PADDING_WORDS(0x1);
+    BitField<0, 3, u32> vs_output_total;
 
     union VSOutputAttributes {
         // Maps components of output vertex attributes to semantics
@@ -1157,8 +1157,10 @@ struct Regs {
             }
         } input_register_map;
 
-        // OUTMAP_MASK, 0x28E, CODETRANSFER_END
-        INSERT_PADDING_WORDS(0x3);
+        BitField<0, 16, u32> output_mask;
+
+        // 0x28E, CODETRANSFER_END
+        INSERT_PADDING_WORDS(0x2);
 
         struct {
             enum Format : u32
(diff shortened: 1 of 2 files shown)


JohnGodgames:
### Also fixed:

Barbie And Her Sisters Puppy Rescue

barbie puppy

barbie puppy 2

Romsstar:
Confirming, this fixes almost all graphical issues of #1504, except for some textures like that Agumon above and also the 3D Model is not being shown on top screen during the save, it's supposed to.

Reference from 3DS again
34d4bb62-e881-11e5-8be7-956c0aad237b

The 2D Images during talking also show. 
fixefd

Instamerge please :D

MoochMcGee:
LGTM. Fixes stuff and it makes sense! Nice work!

JohnGodgames:
Also fixed solid color issued at these games:

Fireman Sam to The Rescue
Gardenscapes
Chibi Devi
Jewel Match 3 EUR
Penguins of Madagascar
Tiny Games - Knights & Dragons

LGTM

JayFoxRox:
Rebased to current master [a2024d749790f94979ad1f7a02363ab53f7de518]. Looks like #1496 also fixed Agumon texture in Digimon. Let's blame the menu rendering issue on depth issues or something which are not actually caused by this. (Also: The menu logo is rendered using sw rasterizer, which crashes on the save load screen - seperate issue)

Fixed textures in master

@citra-emu: We can merge this or check the outmap_total behaviour on hardware first (which is hard for me to do without direct access to hw). What do you prefer?

Romsstar:
Just tested this with #1496. It indeed fixes the agumon texture, but it now flickers sometimes.
I don't know though if it's my system or if that happens to others well.
Some 2D graphics still remain unfixed after this, but I believe this might be a separate issue.

Subv:
LGTM

JamePeng:
LGTM

bunnei:
LGTM

yuriks:
Did we run any hwtests on this?

JayFoxRox:
FTR: no.

PR #1509, opened 2016-03-13
common: Minor changes to NonCopyable

lioncash:
The signature for the copy constructor and copy assignment operators generated by the compiler contain a const reference parameter, so we may as well specify that for consistency.

Also make the default constructor constexpr so it doesn't hamper objects that use it from being constexpr.


MerryMage:
LGTM

Subv:
LGTM

bunnei:
LGTM

PR #1517, opened 2016-03-14
fix SDL2 detection for Visual

Lectem:
This fixes the default installation hierarchy of SDL2, so that users only have to extract and set SDL2DIR.
This was inspired by the following official cmake patch for FindSDL.cmake https://cmake.org/gitweb?p=cmake.git;a=commitdiff;h=96197198


MerryMage:
I hadn't considered Windows users not using the bundled libraries.

LGTM

externals/cmake-modules/FindSDL2.cmake
```diff
@@ -134,11 +134,17 @@ SET(SDL2_SEARCH_PATHS
...
+else()
+    set(VC_LIB_PATH_SUFFIX lib/x86)
+endif()
+
 FIND_LIBRARY(SDL2_LIBRARY_TEMP
     NAMES SDL2
     HINTS
     $ENV{SDL2DIR}
-    PATH_SUFFIXES lib64 lib
+    PATH_SUFFIXES lib64 lib  ${VC_LIB_PATH_SUFFIX}
```
MerryMage: Is the double space necessary?

Lectem: nope, it's a typo, I'll update the PR

bunnei:
LGTM

PR #1518, opened 2016-03-14
SWRasterizer: Don't draw outside of framebuffer

JayFoxRox:
Fixes a crash / sigsegv in Digimon load / save file menu with the software renderer when:

```
(gdb) print g_state.regs.viewport_size_x => 4579328 (BitField [0:23]) = 120.0
(gdb) print g_state.regs.viewport_size_y => 4654080 (BitField [0:23]) = 260.0

g_state.regs.viewport_corner = 54591488
(gdb) print g_state.regs.viewport_corner.x => 0 (BitField [0:15])
(gdb) print g_state.regs.viewport_corner.y => 833 (BitField [16:31])

(gdb) print g_state.regs.framebuffer.width => 480 (BitField [0:10])
(gdb) print g_state.regs.framebuffer.height => 399 (BitField [12:21])

(gdb) print min_y >> 4 => 1086
(gdb) print max_y >> 4 => 1090
```

Crashes here:
```
(gdb) bt
#0  0x000000000063beb9 in Color::DecodeD24S8 (bytes=0x8000d78bc464 <error: Cannot access memory at address 0x8000d78bc464>)
    at /home/fox/Data/Projects/citra/src/./common/color.h:137
#1  0x000000000078c39e in Pica::Rasterizer::GetDepth (x=103, y=-688) at /home/fox/Data/Projects/citra/src/video_core/rasterizer.cpp:123
#2  0x0000000000792274 in Pica::Rasterizer::ProcessTriangleInternal (v0=..., v1=..., v2=..., reversed=false)
    at /home/fox/Data/Projects/citra/src/video_core/rasterizer.cpp:867
#3  0x0000000000792d02 in Pica::Rasterizer::ProcessTriangle (v0=..., v1=..., v2=...) at /home/fox/Data/Projects/citra/src/video_core/rasterizer.cpp:1144
#4  0x0000000000788ff3 in Pica::Clipper::ProcessTriangle (v0=..., v1=..., v2=...) at /home/fox/Data/Projects/citra/src/video_core/clipper.cpp:167
#5  0x000000000077253d in VideoCore::SWRasterizer::AddTriangle (this=0x7fffcc030b30, v0=..., v1=..., v2=...)
```

This patch avoids a crash when the viewport is larger / outside of the the framebuffer.
I *think* there could be another problem if the viewport corner is not inside the framebuffer because the pixel coords could underflow then (with loop counters being u16..)

Doesn't crash anymore with this patch but some mesh for Agumon is still missing ~~- might be related to this (g_state.regs.viewport_corner?)?~~ (Screen based on #1508 + proposed fix):

Menu


src/video_core/rasterizer.cpp
```diff
@@ -380,6 +380,12 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
     for (u16 y = min_y + 8; y < max_y; y += 0x10) {
         for (u16 x = min_x + 8; x < max_x; x += 0x10) {
 
+            // Don't draw pixels outside of framebuffer
+            // framebuffer.height is actually "framebuffer height - 1"
+            const auto& framebuffer = g_state.regs.framebuffer;
+            if (x >> 4 >= framebuffer.width || y >> 4 > framebuffer.height)
```
lioncash: This can likely be moved to a bool function called IsOutsideFrameBuffer(x, y), which is more self-documenting.

JayFoxRox:
Fixed differently in #1519 . This issue still remains if viewport and framebuffer don't match up.
We should check if this kind of behaviour would be possible on hardware and possibly throw an error instead of silently continuing if writes / reads outside the framebuffer happen using routines like GetDepth() etc.

Subv:
LGTM after @lioncash's comment is addressed

JayFoxRox:
I'll close this for now. I have no intention of fixing the commit because this the original issues were solved indirectly by the viewport fix.
If the viewport on hardware can be larger than the framebuffer (= needs hw testing) this problem will still resurface at some point and we can reopen this issue then.

JayFoxRox:
I'm re-opening this one.

* I've changed the commit for better performance and tried to make the code a bit cleaner.
* I also noticed that there could be an underflow when viewport_halfsize is larger than viewport_offset so I added a std::max to avoid casting signed numbers to unsigned short.
* Furthermore I split the original commit into 2 parts: ASSERT'ing on the problem and fixing it.

PR #1519, opened 2016-03-14
PICA: Fix viewport offset

JayFoxRox:
Turns out the viewport offset is 10 bit signed.

Game is Digimon World ReDigtize Decode. Previously didn't render Agumon in the load / save menu.

3DS (Image by Romsstar)
On 3DS

Currently crashes on Citra (Addressed in #1518 ) or doesn't draw Agumon with hw renderer.

Fixed (SW Renderer only):
This PR

HW rendering has other issues with this title so only SW rendering was used to test this.
3dbrew already said that it was 10 bit but it was still wrong about signedness. I've already changed it there to reflect these findings from Digimon.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -117,8 +117,8 @@ struct Regs {
     INSERT_PADDING_WORDS(0x11);
 
     union {
-        BitField< 0, 16, u32> x;
-        BitField<16, 16, u32> y;
+        BitField< 0, 10, s32> x;
+        BitField<16, 10, s32> y;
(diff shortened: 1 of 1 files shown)


Romsstar:
Complimentary screenshot on Software Rendere. Doesn't draw on HWRenderer. 
software

lioncash:
LGTM. Will merge this if there's no objections when I get off of work.

bunnei:
LGTM. Sorry for the delay, I confused this one with your other PRs that we still wanted to do more testing with.

PR #1520, opened 2016-03-14
Implement CheckNew3DS and CheckNew3DSApp

JamePeng:
Append an item[is_new3ds] to config file that switch the machine mode easily, 

Implement APT::SetNSStateField,it will update the NS_state_field

qq 20160315162825
qq 20160316231830


JamePeng:
@JayFoxRox like this?
qq 20160315162825

src/citra_qt/config.cpp
```diff
@@ -70,6 +74,10 @@ void Config::ReadValues() {
 }
 
 void Config::SaveValues() {
+    qt_config->beginGroup("CheckNew3DS");
```
JayFoxRox: Maybe merge this into a "System" group with "region_value" and "new_3ds"?

src/core/hle/service/apt/apt.cpp
```diff
@@ -397,6 +398,16 @@ void GetAppletInfo(Service::Interface* self) {
     LOG_WARNING(Service_APT, "(stubbed) called appid=%u", app_id);
 }
 
+void CheckNew3DSApp(Interface * self)
+{
+    PTM::CheckNew3DS(self);
```
JayFoxRox: This has to do more: https://3dbrew.org/wiki/APT:CheckNew3DSApp

src/core/hle/service/apt/apt.h
```diff
@@ -344,6 +344,24 @@ void PreloadLibraryApplet(Service::Interface* self);
  */
 void StartLibraryApplet(Service::Interface* self);
 
+/**
+ * Wrapper for PTMSYSM:CheckNew3DS
+ * APT::CheckNew3DSApp service function
```
JayFoxRox: More than a wrapper. Maybe create a new short description from https://3dbrew.org/wiki/APT:CheckNew3DSApp ?

src/core/hle/service/ptm/ptm.cpp
```diff
@@ -89,6 +89,19 @@ void IsLegacyPowerOff(Service::Interface* self) {
     LOG_WARNING(Service_PTM, "(STUBBED) called");
 }
 
+void CheckNew3DS(Interface * self)
+{
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    ASSERT_MSG(Settings::values.is_new3ds != true,"Citra was not completely supported New3DS yet.",
+              " Please check the config file item[CheckNew3DS] whether it is false");
```
JayFoxRox: Find a native speaker who can improve this.

Also error instead? (Not sure how citra handles these cases normally but an ASSERT feels like a development tool, not a way to report errors to the user)

JamePeng: @JayFoxRox yes, it is a development tool .
If citra have completely support New3DS , this sentense should be removed.
Because the New3DS have different memory operation from old 3DS.
It is very dangerous to tranfer a illegal value(Presently)  to command buffer!

mhvuze:
Can confirm that this solves #1449 if is_new3ds=true.

JamePeng:
@mhvuze Actually,the is_new3ds==true was not supported now!
But before that I think there some logical error with NS(the shell)!

JamePeng:
append a APT:SetShellState implement ,  it will update the shell_open value!
qq 20160316142908

src/core/hle/service/ptm/ptm_sysm.cpp
```diff
@@ -35,7 +35,7 @@ const Interface::FunctionInfo FunctionTable[] = {
     {0x080E0140, nullptr,             "NotifyPlayEvent"},
     {0x080F0000, IsLegacyPowerOff,    "IsLegacyPowerOff"},
     {0x08100000, nullptr,             "ClearLegacyPowerOff"},
-    {0x08110000, nullptr,             "GetShellStatus"},
+    {0x08110000, GetShellState,       "GetShellStatus"},
```
JayFoxRox: Rename "GetShellStatus" to "GetShellState" to match the function name? (also used on 3dbrew)

src/core/hle/service/apt/apt_a.cpp
```diff
@@ -27,7 +27,9 @@ const Interface::FunctionInfo FunctionTable[] = {
     {0x00430040, NotifyToWait,                 "NotifyToWait?"},
     {0x00440000, GetSharedFont,                "GetSharedFont?"},
     {0x004B00C2, AppletUtility,                "AppletUtility?"},
-    {0x00550040, nullptr,                      "WriteInputToNsState?"},
+    {0x00550040, SetShellState,                "SetShellState"},
```
JayFoxRox: Nope. This looks very wrong.
I think you don't understand what the ["shell state" refers to in english - click here to see an image of a "shell"](http://thumbs.dreamstime.com/x/shell-pearl-beach-53415110.jpg).
The shell state therefore means if the 3DS is in [this position (Opened!)](https://www.avforums.com/image.php?imageparameters=editorial/products/6571f5daab91ad960958a2517d5435da_9186.jpg), or [this position (Closed!)](https://static.avforums.com/styles/avf/editorial/products/6e7aa-closed-angle.JPG).

As this is a hardware setting it can not be manipulated by the software. Therefore your function name doesn't make any sense (unless it would only be used on 2DS which has an unknown shell state - but I don't think so)

To my understanding 0x00550040 gives write access to unknown 8 bit variable which is part of the NS state but nobody knows what each bit means or if it's a value.
The 8 bit variable can be set using 0x00550040 and read back using 0x00560000.

src/core/hle/service/apt/apt.cpp
```diff
@@ -397,6 +407,23 @@ void GetAppletInfo(Service::Interface* self) {
     LOG_WARNING(Service_APT, "(stubbed) called appid=%u", app_id);
 }
 
+void CheckNew3DSApp(Interface * self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    if (PTM::Get_ShellState()) {
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+        cmd_buff[2] = 0;
```
JayFoxRox: This does not check Get_ShellState or anything related to the 8 bits NS State field controlled by 0x00550040 (probably).

The description on 3dbrew probably means that there is a 8 bit field in the NS State which indicates wether APT:PrepareToStartApplication had the bit8 set in it's flags parameter (if it did the value of the field value will be set to 0x01). [Starting with version 8.0.0-18]
(Those 8 bits might but it might not or might be the 8 bits from 0x00550040!)
If the 8 bit field has the value 0x00 this returns the same as PTMSYSM:CheckNew3DS.
If the 8 bit field has a value which is not 0x00 the output of this function must be 0.

src/core/hle/service/ptm/ptm.cpp
```diff
@@ -89,6 +97,21 @@ void IsLegacyPowerOff(Service::Interface* self) {
...
 }
 
+bool Check_New3DS() {
+    return Settings::values.is_new3ds;
+}
+
+void CheckNew3DS(Interface * self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    ASSERT_MSG(Check_New3DS() != true,"Citra was not completely supported New3DS yet. Please check the config file item[is_new3ds] whether it is false");
```
JayFoxRox: Instead of having this assert, how about adding a LOG_WARNING in the Qt config code whenever the value is loaded or maybe on startup of the rom (if there is any good place to put it).

The "is_new3ds" config will probably be used elsewhere too in the future and the problem is not in this function. Games might depend on new3ds functionality without even checking this.

Also (Warning) "'is_new3ds' is enabled as part of the 'System' settings. Citra does not fully support New 3DS emulation yet!" is clearer imho. However, that's for others to decided and I'm not a native speaker.

src/core/hle/service/apt/apt_a.cpp
JamePeng: http://3dbrew.org/wiki/NS_and_APT_Services

Are you sure the shell?

JayFoxRox: Yes, very likely.
http://3dbrew.org/wiki/PTM:GetShellState Indicates that it's untested for 2DS (because that isn't a shell-case and nobody bothered to try it). Subv also gave this impression when he responded on IRC: "Subv: because you usually play your 3DS with the shell open" and "Subv: what does that have to do with the shell state?".
I also think it's a bit confusing and a misnomer (I'd prefer "GetShellCaseState" or something but we shouldn't change it either or it will cause more confusion).

JamePeng: ya, there are a lot of 3dbrew's APIs  make me confusing! :(

JamePeng:
@JayFoxRox how about this time?

JayFoxRox:
The names for 0x00550040 and 0x00560000 are still wrong and their functionality is unknown at this point (afaik).
Again: There is **no way to control the ShellState from software**.

0x00550040 and 0x00560000 are **NOT** ShellState; They are related to the NS State (NS is the module name and includes Shell in its name, however, the word Shell in NS (as in computer shell) and Shell in ShellState (as in clamshell case) are probably entirely unrelated).

0x00550040 and 0x00560000 are either some generic NS State field OR it's a common NS State field user for multiple states/flags (including the one that influences the CheckNew3DSApp call) OR something entirely different.
**You should not implement 0x00550040 and 0x00560000** unless you know what they are doing, at least don't rename them until then. It will cause confusion.

If you need 0x00550040 and 0x00560000 you could debug this on hardware to figure out where the data from APT:PrepareToStartApplication ends up or how 0x00550040 and 0x00560000 work. Maybe the people in \#3dsdev have some idea?

JamePeng:
look like there are no way to implement checknew3dsapp without the 0x00550040 and 0x00560000

src/core/hle/service/apt/apt.cpp
```diff
@@ -397,6 +420,23 @@ void GetAppletInfo(Service::Interface* self) {
...
+void CheckNew3DSApp(Interface * self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    if (ns_state_field) {
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+        cmd_buff[2] = 0;
+    } else {
+        PTM::CheckNew3DS(self);
+    }
+    LOG_WARNING(Service_APT, "called isNew3DS = 0x%08x", PTM::Check_New3DS());
```
wwylele: Should this line be within the `else` block? Also `Check_New3DS` -> `CheckNew3DS`?

src/core/hle/service/ptm/ptm.cpp
```diff
@@ -89,6 +89,23 @@ void IsLegacyPowerOff(Service::Interface* self) {
     LOG_WARNING(Service_PTM, "(STUBBED) called");
 }
 
+bool Check_New3DS() {
```
wwylele: I really don't like a name like this. You can just use `CheckNew3DS` as an overload, or make a more specified name like `CheckNew3DSFromSettings`, imo.

src/core/hle/service/apt/apt.cpp
JayFoxRox: I believe this is supposed to be

```c++
cmd_buff[2] = ns_state_field ? 0 : PTM::CheckNew3DS(self);
LOG_WARNING(Service_APT, "called ns_state_field = 0x%02x isNew3DS = 0x%08x", ns_state_field, PTM::Check_New3DS());
```

src/core/hle/service/apt/apt.cpp
```diff
@@ -41,6 +42,8 @@ static std::shared_ptr<std::vector<u8>> shared_font;
 
 static u32 cpu_percent; ///< CPU time available to the running application
 
+static u8 ns_state_field;
```
JayFoxRox: This should probably be elsewhere? I have no idea where the other NS fields are stored.
Also the name should indicate that this is an **unknown** state field: maybe "unknown_ns_state_field" ?

I'm not sure where and if it should be initialized either. 3dbrew says the state field value is zero "normally".

Xcedf:
Merged this to test, can't run the game even with is_new3ds=true
what did i miss

src/core/hle/service/apt/apt.cpp
JamePeng: " Processes launch applications via home-menu, not directly with APT:PrepareToStartApplication and APT:StartApplication. Regular applications can't directly launch applications since APT:StartApplication launches the process without terminating the currently running application."
from   http://3dbrew.org/wiki/NS_and_APT_Services

I think the state field is zero "normally",because the citra(without the HOME menu)didn't use the APT:PrepareToStartApplication and APT:StartApplication to launch applications!

JamePeng:
@Xcedf 
#1450

subzerohurley:
ok.....how do i implement this? i'm completely new, i just want to play monster hunter 4 using citra, and i guess this fixes the issue

Xcedf:
This was helpfull, thanks

JamePeng:
Actually I am not recommend use the #1450 now. 
With my warning say that "Citra does not fully support New3DS emulation yet!"

JamePeng:
@bunnei Is there anything should I pay attention to work?

src/core/hle/service/apt/apt.cpp
```diff
@@ -379,6 +386,22 @@ void StartLibraryApplet(Service::Interface* self) {
     cmd_buff[1] = applet->Start(parameter).raw;
 }
 
+void SetNSStateField(Service::Interface * self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    ns_state_field = cmd_buff[1];
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    LOG_WARNING(Service_APT, "(stubbed) called SetNSStateField=%u", ns_state_field);
```
bunnei: Ditto on your inconsistent logging..
* stubbed -> STUBBED
* SetNSStateField -> ns_state_field

src/core/hle/service/apt/apt.cpp
```diff
@@ -379,6 +386,22 @@ void StartLibraryApplet(Service::Interface* self) {
...
+}
+
+void GetNSStateField(Service::Interface * self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    cmd_buff[8] = ns_state_field;
+    LOG_WARNING(Service_APT, "(stubbed) called GetNSStateField=%u", ns_state_field);
```
bunnei: We don't usually print return values.. I'd get rid of the argument

JamePeng: But we can see what value was returned?

src/core/hle/service/apt/apt.cpp
```diff
@@ -264,6 +267,10 @@ void PrepareToStartApplication(Service::Interface* self) {
     u32 title_info4  = cmd_buff[4];
     u32 flags        = cmd_buff[5];
 
+    if ((flags >> 8) & 0x00000001) {
```
bunnei: What is this? Please add a comment what the meaning of this is...

Also, why not just `if (flags & 0x00000100)` ?

src/core/hle/service/apt/apt.cpp
```diff
@@ -379,6 +386,22 @@ void StartLibraryApplet(Service::Interface* self) {
     cmd_buff[1] = applet->Start(parameter).raw;
 }
 
+void SetNSStateField(Service::Interface * self) {
```
bunnei: `void SetNSStateField(Service::Interface* self) {`

(position of the `*`)

src/core/hle/service/apt/apt.cpp
```diff
@@ -379,6 +386,22 @@ void StartLibraryApplet(Service::Interface* self) {
...
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    ns_state_field = cmd_buff[1];
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+    LOG_WARNING(Service_APT, "(STUBBED) ns_state_field=%u", ns_state_field);
+}
+
+void GetNSStateField(Service::Interface * self) {
```
bunnei: Ditto, will stop pointing out, please pay more attention to this

src/core/hle/service/apt/apt.cpp
```diff
@@ -414,6 +437,23 @@ void GetStartupArgument(Service::Interface* self) {
...
+void CheckNew3DSApp(Interface * self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    if (ns_state_field) {
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+        cmd_buff[2] = 0;
+    } else {
+        PTM::CheckNew3DS(self);
+    }
+    LOG_WARNING(Service_APT, "(STUBBED) isNew3DS = 0x%08x", PTM::CheckNew3DSFromSettings());
```
bunnei: As mentioned in other PRs, only log function inputs, therefore this can just be:
<pre>
LOG_WARNING(Service_APT, "(STUBBED) called");
</pre>

src/core/hle/service/apt/apt.cpp
```diff
@@ -414,6 +437,23 @@ void GetStartupArgument(Service::Interface* self) {
...
+}
+
+void CheckNew3DS(Interface * self) {
+    PTM::CheckNew3DS(self);
+    LOG_WARNING(Service_APT, "(STUBBED) isNew3DS = 0x%08x", PTM::CheckNew3DSFromSettings());
```
bunnei: ditto

src/core/hle/service/apt/apt.h
```diff
@@ -361,6 +361,41 @@ void StartLibraryApplet(Service::Interface* self);
  */
 void GetStartupArgument(Service::Interface* self);
 
+/**
+ * APT::SetNSStateField service function
+ * This writes the input u8 to a NS state field.
```
bunnei: Missing inputs/outputs section in docstring

src/core/hle/service/apt/apt.h
```diff
@@ -361,6 +361,41 @@ void StartLibraryApplet(Service::Interface* self);
...
+ */
+void SetNSStateField(Service::Interface* self);
+
+/**
+ * APT::GetNSStateField service function
```
bunnei: Missing inputs/outputs section in docstring

src/core/hle/service/apt/apt.h
```diff
@@ -361,6 +361,41 @@ void StartLibraryApplet(Service::Interface* self);
...
+ * This returns an u8 NS state field(which can be set by cmd 0x00550040), at cmdreply+8.
```
bunnei: *a u8

JamePeng:
done!

JamePeng:
:(  I really really really(....) need the config "is_new3ds" to do something ! ! !

src/core/hle/service/apt/apt.cpp
lioncash: This should likely have a comment indicating what it's current use is for

src/core/hle/service/ptm/ptm.h
```diff
@@ -88,6 +88,17 @@ void GetTotalStepCount(Interface* self);
  */
 void IsLegacyPowerOff(Interface* self);
 
+// Return the 3DS machine mode from settings
+bool CheckNew3DSFromSettings();
```
lioncash: This could likely just be named `IsNew3DS()`

lioncash: This function isn't particularly needed, since the main body of CheckNew3DS can just be:

```cpp
void CheckNew3DS(Service::Interface* self) {
    const bool is_new_3ds = Settings::values.is_new3ds;

    if (is_new_3ds) {
        LOG_CRITICAL(Service_PTM, "The option 'is_new3ds' is enabled as part of the 'System' settings. Citra does not fully support New3DS emulation yet!");
    }

    cmd_buff[2] = is_new_3ds ? 1 : 0;

    LOG_WARNING(Service_PTM, "(STUBBED) called isNew3DS = 0x%08x", static_cast<u32>(is_new_3ds));
}
```

JamePeng: Thanks， lioncash

JasonSome:
How is implementation here going?

bunnei:
LGTM

PR #1522, opened 2016-03-14
OS X: Use dylibbundler to include dylibs required by citra in nightly build

MerryMage:
Our nightly builds currently do not include a dylib of SDL2 along in the archive. This requires the user to hunt down and build/install the appropate version of SDL2.

This fixes that by adding the appropriate libraries to the archive using `dylibbundler`. `dylibbundler` also runs `install_name_tool` as appropriate to update the dylib references in the `citra` executable.

This is likely a better solution than #1468 which I originally proposed. Would like someone who actually develops programs for OS X to comment.


archshift:
Tested, and it works! LGTM!

bunnei:
Thanks!

PR #1527, opened 2016-03-15
Keep track of the active texture

linkmauve:
The main benefit of this is that the apitrace is much less cluttered.

I obviously couldn’t find any performance difference on Mesa, but I’d expect some drivers might behave in a different way.

Another state that would be useful to track is the texture bound to the current unit, so we could stop setting it to 0 and putting it back.


JayFoxRox:
I'm not sure about this.. it looks kind of redundant. Most drivers are probably smart enough to check this before sending data to the hardware or they are lazy and only check it on the upload / only set a pointer.

apitrace logs can be filtered so I see no issue there.

It would probably more benefical to use the debug extension and put all GL related functions into debug groups, then label each texture with:

* It's physical RAM address
* Texture type / format (to spot conversion errors)
* ... ?

I did something similar for xqemu where I was able to fix a lot of texture related bugs:

xqemu GL debug use

Something similar can be done for framebuffers. Shaders should be documented properly too (maybe only in debug builds?)
It also makes sense to debug print the attribute type / sources in the log.

---

I haven't tested this PR yet but the code looks fine. However I feel like it's harder to read the code this way.

bunnei:
I don't really have an opinion on this change, but I'd like to get this wrapped up either way.
@linkmauve do you have any followup to the above comment?

MerryMage:
> GL related functions into debug groups

IIRC KHR_debug is OpenGL 4.3 minimum.

LittleWhite-tb:
Isn't available through extension ? ( https://www.opengl.org/registry/specs/KHR/debug.txt )

linkmauve:
Superseded by #1724, which does a much better job at it by doing everything instead of just the active texture.

PR #1535, opened 2016-03-17
PICA: Alignment happens locally in vertex

JayFoxRox:
pcmaker informed me about a [regression with "Thor - Thunder of Gods"](http://i.imgur.com/IcbBtoj.jpg) since #1496.
As @yuriks and me hw-tested the aligment this came as a surprise.

I had a lot of trouble debugging this issue because the alignment looked fine while debugging it. Then I noticed that my debug code tested the alignment by aligning locally in each vertex (load_address = 0 for testing), while Citra was doing it globally because load_address can be unaligned.
The hw-test we did also only ever used aligned buffers so we never hit this case. (Sorry!)

This PR makes the alignment local to each vertex. It fixes the Thor regression, Super Smash Bros also still works. I didn't test any other games yet.

With the assumption that alignment happens to avoid accessing multiple words from RAM it would be insanely weird if PICA works as in this PR.. (as each local vertex could still be unaligned in RAM, so the align is not really necessary for the hw and only fills up bandwidth).

I could also imagine the Thor regression being caused by a memory allocator bug in Citra which might not return aligned addresses to the game. So this should be tested on hardware to be certain, however, even if that is the case this code will still work (but it wouldn't be as clear that alignment happens globally due to RAM access, instead it would suggest local alignment).

So try this PR and if it breaks even more I'll have to rethink the whole alignment thing (again).
\- If it works fine we'll just make another cross in the long list we shall call "PITA200 is being weird again" and merge this.

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -200,7 +200,7 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
             for (int loader = 0; loader < 12; ++loader) {
                 const auto& loader_config = attribute_config.attribute_loaders[loader];
 
-                u32 load_address = base_address + loader_config.data_offset;
+                u32 offset = 0;
 
                 // TODO: What happens if a loader overwrites a previous one's data?
                 for (unsigned component = 0; component < loader_config.component_count; ++component) {
@@ -212,17 +212,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     u32 attribute_index = loader_config.GetComponent(component);
                     if (attribute_index < 12) {
                         int element_size = attribute_config.GetElementSizeInBytes(attribute_index);
-                        load_address = Common::AlignUp(load_address, element_size);
-                        vertex_attribute_sources[attribute_index] = load_address;
+                        offset = Common::AlignUp(offset, element_size);
+                        vertex_attribute_sources[attribute_index] = base_address + loader_config.data_offset + offset;
                         vertex_attribute_strides[attribute_index] = static_cast<u32>(loader_config.byte_count);
                         vertex_attribute_formats[attribute_index] = attribute_config.GetFormat(attribute_index);
                         vertex_attribute_elements[attribute_index] = attribute_config.GetNumElements(attribute_index);
                         vertex_attribute_element_size[attribute_index] = element_size;
-                        load_address += attribute_config.GetStride(attribute_index);
+                        offset += attribute_config.GetStride(attribute_index);
                     } else if (attribute_index < 16) {
                         // Attribute ids 12, 13, 14 and 15 signify 4, 8, 12 and 16-byte paddings, respectively
-                        load_address = Common::AlignUp(load_address, 4);
-                        load_address += (attribute_index - 11) * 4;
+                        offset = Common::AlignUp(offset, 4);
+                        offset += (attribute_index - 11) * 4;
                     } else {
                         UNREACHABLE(); // This is truly unreachable due to the number of bits for each component
                     }


yuriks:
This would indeed be extremely weird. Mind adding a test to your hb app that does this so I can check?

JayFoxRox:
**tl;dr**: confirmed on HW.

---

@MerryMage tested this on HW using https://github.com/JayFoxRox/3ds-tests/commit/c3512473e85b266fc5309be33886118d8e42c3ab 

If the buffer base + offset passed to the GPU is pointing to vertex data it will be drawn. No alignment happens at this point. All bits of the offset are used, even the lower bits = the address is not aligned in any way.
However, the fields within the vertex data must still be aligned (= Alignment happens locally in vertex).

---

The version of the hwtest tests this by giving the user the option to press A to move the offset where vertex data is written on the CPU (= write data elsewhere inside VBO). B can be pressed to move the offset where the data is read by the GPU (= modify offset into VBO passed to GPU).
Each field of the vertex is still aligned relative to the vertex start (= locally).

Here is the hw-test discussion from IRC (which lead to the above summary):

```
JayFoxRox: what happens when A and B are equal? (so passed offset = base offset)
merry: triangle appears (1)
JayFoxRox: even if both are like 3, 5, 6, 7 or something?
merry: yes (2)
JayFoxRox: does it also still work when you press START a couple of times?
merry: okay, what I did was press start, then press A and B alternately about 10 times then press start again, repeatedly; triangle appears every time you expect (3)
merry: i.e.: when the offsets are equal (4)
JayFoxRox: but as soon as you press only A or B once it disappears again?
merry: yeah (5)
```

(1): if data is written where the GPU is told it will work..
(2): ..even if data is passed at an unaligned address!
(3): It works for different element types.
(4): merry concludes what (1) and (2) mean.
(5): the passed offsets are not aligned, otherwise off-by-one would be "handled" by alignment.

Kitlith also ran the test and came to the conclusion that it works for all alignment modes the test supports as long as passed offset == actual offset.
(He also looked for a combination where it would draw if one of the offsets was different from the other [implying alignment of addresses] but couldn't find any combination)

Unless there are any further issues with my hw-test code (review welcome, also check underlying ctrulib and citro3d) I'm assuming this PR is valid and can be merged.

(Also thanks to @wwylele for investing time into testing my first broken version of the HW test)

yuriks:
LGTM, then. Did you get a chance to check if the padding elements align at 4 or 2 bytes?

JayFoxRox:
Yes. @MerryMage just tested some more.

When you and me tested the stride for the previous PR it froze. My first version of a hw-test for this PR also froze and it turned out to be related to CPU alignment. So I had him retest that as well.
However, changing the passed stride still freezes and I don't think it can be CPU related this time.
= Stride is fine, Citra can trust what is passed to the GPU and doesn't have to do any alignment + it can't be larger than the vertex size either.

I also added a test mode 6:

```c++
struct {
  float x,y; // +0
  uint8_t pad1[3]; // +8
  uint8_t pad2[3]; // +11
  // expected align to 4 byte offset; +14
  uint32_t padding_attribute; // +16
  uint16_t r,g,b; // +20
  // implicit align to next 4 byte offset (for next vertex); +26
};
```

If the padding attribute would be aligned a 2 byte offset it would be at offset +14 and r,g,b would start at +18. The passed stride for this mode must be 28.
= Padding attributes are aligned to 4 byte offsets and that was correctly implemented in the last PR

I guess the last remaining case which could be checked would be wether the stride alignment might be related to the largest element size contained in the vertex (instead of the first field of the next vertex). However, that would be a super rare case and we have done enough hardware testing for now. If problems arise this is something which might be responsible though (highly unlikely).

bunnei:
LGTM

PR #1536, opened 2016-03-17
CMakeLists: Use list appending to reduce boilerplate

lioncash:
Self-explanatory stuff. Gets rid of the need to re-specify the variable name.


yuriks:
The new and old code don't do the same thing. CMake lists are `;` separated, not space separated. I don't know if CMake handles that in the flags variables but I think it doesn't. I remember already trying to do this before.

lioncash:
I'll take your word for it. I tried it on OSX and it worked fine, but maybe that was just a situational case.

PR #1538, opened 2016-03-17
shader_interpreter: use std::inner_product for the dot product

lioncash:
Same thing, less code.

### DIFF
--- a/src/video_core/shader/shader_interpreter.cpp
+++ b/src/video_core/shader/shader_interpreter.cpp
@@ -2,10 +2,10 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include <common/file_util.h>
-
+#include <numeric>
 #include <nihstro/shader_bytecode.h>
 
+#include "common/file_util.h"
 #include "video_core/pica.h"
 #include "video_core/pica_state.h"
 #include "video_core/shader/shader.h"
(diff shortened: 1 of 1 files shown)


JayFoxRox:
I've mentioned this on IRC but I'm not sure if anybody was reading:

As inner_product supports an initial value, dph can be "optimized" (= written differently). In case of dph: only do 3 components and use src2[3] as initial value
So num_components only needs to check for DP4 then, initial value is depending on check for DPH / DPHI then.

something like (untested code):

```c++ 
int num_components = (opcode == OpCode::Id::DP4) ? 4 : 3;
float24 add_value = (opcode == OpCode::Id::DPH || opcode == OpCode::Id::DPHI) ? src2[3] : float24::FromFloat32(0.f);
float24 dot = std::inner_product(src1, src1 + num_components, src2, add_value);
```

It shouldn't make any difference in perf as we are messing with constants here.
I don't like how DP3/DPH/DP4 are mixed anyway and I don't think my solution is nicer or worse.

MerryMage:
I don't really like the new suggestion as it is unreadable. `std::inner_product` alone is clear and self-documenting since a dot product is the inner product of two vectors.

MerryMage:
LGTM

bunnei:
LGTM, ditto on @MerryMage's comment

PR #1368, opened 2016-01-24
Implementation for a configure widget

LittleWhite-tb:
I have worked on a new widget to configure Citra:

citra-configure-dialog

There is nothing in "Input" tab since we don't have the input widget yet.

citra-debug-configuration

Additionally, I have centralized all the configuration in Config class. Now, we should have one configure file only.

PR #1540, opened 2016-03-17
Citra-Qt: Avoid warnings when loading the configuration

Kloen:
I think it's weird the fact that when you start citra-qt it spams the log with warnings. So... That's the reason of this PR.

In addition, when you set HW or JIT to enable/disable, it writes a warning in the log, which I think it's weird too, so... why not a LOG_INFO?.


src/citra_qt/main.cpp
```diff
@@ -151,10 +151,10 @@ GMainWindow::GMainWindow() : emu_thread(nullptr)
     GDBStub::SetServerPort(static_cast<u32>(Settings::values.gdbstub_port));
 
     ui.action_Use_Hardware_Renderer->setChecked(Settings::values.use_hw_renderer);
-    SetHardwareRendererEnabled(ui.action_Use_Hardware_Renderer->isChecked());
+    VideoCore::g_hw_renderer_enabled = ui.action_Use_Hardware_Renderer->isChecked();
```
yuriks: This reeeeeallly isn't how you fix this, since it'll disable saving of the settings.

The right way of doing this is to just get rid of the warnings logs in FileUtil and replace it with proper returning of errors instead.

Kloen: In fact, actually, we are:

Loading settings.
SetHardwareRendererEnabled (Loading settings. + Saving settings).
SetShaderJITEnabled (Loading settings. + Saving settings).

Just by starting citra-qt, cause when we call SetHardwareRendererEnabled or SetShaderJITEnabled we are loading and saving them again.

Replacing these lines only afects on first start, the behavior after that should be, and in fact I tested it to be the same. So there isn't no save disabling or something like that.

These lines only changes the behavior when we start citra-qt and instead of calling the SetSomethingEnabled it just set it directly to VideoCore as the method do, but without overwriting the config on a totally unnecesary moment.

Kloen:
I'll close this in favor of #1368 cause the warning spam on start that I'm trying to avoid here, isn't an issue anymore once #1368 gets eventually merged.

PR #1541, opened 2016-03-17
Fix Qt string encode problem on Windows

LFsWang:
On Windows10 with codepage 950

Before (can not open that game)
image

After (it is fine)
image

**But it just a temporary fix! **

There are more problems with [here](https://github.com/citra-emu/citra/blob/master/src/common/file_util.cpp#L444)

Citra use FindFirstFileA / FindNextFileA as default ,but these functions can not correctly handle filename. for example : 

There are some file names
```
ゲーム.cci (Japanese)
경기.cci (Korean)
游戏.cci  (Simplified Chinese)
遊戲.cci  (Traditional Chinese) (local environment)
game.cci (English)
```

image
image

Only 遊戲.cci & game.cci are correctly loaded. 
`ゲーム.cci`  with UTF-8 
`e3 82 b2 e3 83 bc e3 83 a0 2e 63 63 69`

dump ffd.cFileName got
`3f 3f 3f 2e 63 63 69`

and it cause error and output `core\loader\loader.cpp:Loader::IdentifyFile:49: Failed to load file ./???.cci`
image

I can keep trying to solve it.

### DIFF
--- a/src/citra_qt/config.cpp
+++ b/src/citra_qt/config.cpp
@@ -16,7 +16,7 @@ Config::Config() {
     // TODO: Don't hardcode the path; let the frontend decide where to put the config files.
     qt_config_loc = FileUtil::GetUserPath(D_CONFIG_IDX) + "qt-config.ini";
     FileUtil::CreateFullPath(qt_config_loc);
-    qt_config = new QSettings(QString::fromStdString(qt_config_loc), QSettings::IniFormat);
+    qt_config = new QSettings(QString::fromLocal8Bit(qt_config_loc.c_str()), QSettings::IniFormat);
 
     Reload();
 }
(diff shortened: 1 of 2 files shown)


hrydgard:
A better fix would be to replace the usage of FindFirstFileA etc with FindFirstFileW and handle the wide character strings correctly, would make it work even if the codepage is not that particular one. That will probably be a bit more work though.

yuriks:
Not sure how I feel about this. On one hand it does fix the problem for your native locale, and that's at least something I suppose. On the other, it's still really fragile like ector said. Code gets an LGTM but it would be really cool if someone could go all the way on this.

bunnei:
Yeah, I see no reason not to merge this one... Opened #1584 as a reminder to address the broader issue.

wwylele:
this seems breaking the compilation on master.

LittleWhite-tb:
It does fail to compile on Linux.
@LFsWang (or someone else) : could you check that #1589 does not cause the bug to come back on Windows.

PR #1543, opened 2016-03-18
vector_math: Add missing member in Vec4's SetZero function

lioncash:
Previously Vec4's SetZero would only reset three of its four elements.

### DIFF
--- a/src/common/vector_math.h
+++ b/src/common/vector_math.h
@@ -447,7 +447,10 @@ public:
 
     void SetZero()
     {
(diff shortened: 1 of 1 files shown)


archshift:
LGTM

yuriks:
Was this function even used? I wonder if it causes any bugs.

lioncash:
@yuriks Luckily, no

JayFoxRox:
LGTM

bunnei:
LGTM

Issue #1350, opened 2016-01-08
Software renderer relies on OpenGL 3.3


LittleWhite-tb:
Hello,

While going through the code, I have noticed that it is not possible to use the software renderer without having OpenGL 3.3.

It is natural to think that the software renderer relies only on software, so we (I include myself on this one) have to think a way to render the buffer without OpenGL 3.3.

yuriks:
It should be possible to do this by simply painting the surface using Qt, but there are several issues regarding making OpenGL mode also work (without slowing it down) at the same time. I expect this to require large refactors of the rendering frontend and UI code.

benderscruffy:
can be closed

benderscruffy:
Is this still an issue?.

SleepingSnakezzz:
Fixed in https://github.com/citra-emu/citra/pull/6347

PR #1544, opened 2016-03-19
Move glad initialisation to the frontend

linkmauve:
On SDL2 this allows it to use SDL_GL_GetProcAddress() instead of the
default function loader, and fixes a crash when using apitrace with an
EGL context.

On Qt we will need to migrate from QGLWidget to QOpenGLWidget and
QOpenGLContext before we can use gladLoadGLLoader() instead of
gladLoadGL(), since the former doesn’t expose a function loader.

### DIFF
--- a/src/citra/emu_window/emu_window_sdl2.cpp
+++ b/src/citra/emu_window/emu_window_sdl2.cpp
@@ -9,6 +9,8 @@
 #define SDL_MAIN_HANDLED
 #include <SDL.h>
 
+#include <glad/glad.h>
+
 #include "common/key_map.h"
 #include "common/logging/log.h"
 #include "common/scm_rev.h"
@@ -95,6 +97,11 @@ EmuWindow_SDL2::EmuWindow_SDL2() {
 
     if (gl_context == nullptr) {
         LOG_CRITICAL(Frontend, "Failed to create SDL2 GL context! Exiting...");
+        exit(1);
+    }
+
+    if (!gladLoadGLLoader(static_cast<GLADloadproc>(SDL_GL_GetProcAddress))) {
+        LOG_CRITICAL(Frontend, "Failed to initialize GL functions! Exiting...");
         exit(1);
     }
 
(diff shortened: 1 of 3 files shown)


bunnei:
@linkmauve do you plan on fixing the build error?

bunnei:
@linkmauve ping again. Please address (or close this PR if you don't think you'll get to it).

JayFoxRox:
Also, will this possibly affect something like #1350 ? (I'd imagine the goal is to avoid initializing GL/glad until it's really necessary)

src/citra_qt/main.cpp
```diff
@@ -6,6 +6,9 @@
 #include <memory>
 #include <thread>
 
+#include <glad/glad.h>
+
+#define QT_NO_OPENGL
```
linkmauve: This was the issue, some of the Qt headers was including `qopengl.h` which was including the system GL headers, and we don’t want that.

linkmauve:
JayFoxRox, this won’t affect that issue, the only thing changed in this PR is that we are loading the symbols using the platform loader, instead of always using `dlopen()` on libGL.so.1.  It does open the possibility of a GLES 2.0 renderer though, if the platform supports that.

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -98,6 +100,11 @@ EmuWindow_SDL2::EmuWindow_SDL2() {
         exit(1);
     }
 
+    if (!gladLoadGLLoader((GLADloadproc)SDL_GL_GetProcAddress)) {
```
JayFoxRox: This cast looks.. bad?
1. We don't use C Style casts anymore
2. Casting function pointers is a really bad idea because it's possible to break the stack etc. by passing a callback with wrong signature (without a warning!).

linkmauve: 1. Oops, right, I don’t know why I didn’t use static_cast here…
2. At least with gcc 6.1 and clang 3.7, I get a proper error on invalid static_cast from a function whose signature doesn’t match.

bunnei:
LGTM

PR #1546, opened 2016-03-20
Shader JIT Part 2

### LINKED ISSUE #1293: JMP compilation in the Shader JIT is fundamentally broken
The implementation of the JMP family instructions in the vertex shader JIT is broken. For certain programs, the generated control flow will jump to a random branch point in the program, causing certain parts of shaders to be completely skipped.

This happens when a jmp instruction jump over another jmp which itself jumps to a label beyond the target of the original jmp, like so:

```
    jmp label_1
    nop
    jmp label_2
label_1:
    nop
label_2:
    end
```

The first JMP will emit a branch, then initiate a nested loop compiling instructions until label_1. However, when it tries to compile `jmp label_2`, that will initiate another nested loop compiling until `label_2`. When this loop eventually finishes, it will return to the outer jmp compilation loop, which will notice it has passed `label_1` already and set the jump target. The end result is that both jmp instructions will jump to `label_2`.

### LINKED ISSUE #1402: Super Pokemon Rumble graphic issues.
I think that a picture talks on his own...

image

image

image

image

Specs:
image
image

bunnei:
Specifically, this is a new and improved implementation of flow control for the shader JIT. Previously, the shader JIT inlined CALL and JMP instructions. This had several issues: 1) Inlining all subroutines resulted in bloated shaders being generated, and 2) JMP to arbitrary addresses (and nested jumps) couldn't be supported.

With these changes, any arbitrary CALL/JMP will now work. Furthermore, this results in pretty constant compiled shader sizes of around ~40kb, so I changed allocation to happen on the fly (preventing cache clearing and recompiling of previously compiled shaders).

The changes in this patch could also be leveraged to support arbitrary (read: wacky) IF and LOOP instructions... But I didn't feel like doing this, particularly because these instructions in the JIT are a bit finicky and are already well vetted, and also AFAIK no game uses IF/LOOP in a way that the JIT didn't already support (mainly: backwards)

Fixes #1293
Fixes #1402

### DIFF
--- a/src/common/x64/emitter.cpp
+++ b/src/common/x64/emitter.cpp
@@ -531,6 +531,22 @@ void XEmitter::SetJumpTarget(const FixupBranch& branch)
     }
 }
 
+void XEmitter::SetJumpTarget(const FixupBranch& branch, const u8* target)
+{
+    if (branch.type == 0)
+    {
+        s64 distance = (s64)(target - branch.ptr);
+        ASSERT_MSG(distance >= -0x80 && distance < 0x80, "Jump target too far away, needs force5Bytes = true");
+        branch.ptr[-1] = (u8)(s8)distance;
+    }
+    else if (branch.type == 1)
+    {
+        s64 distance = (s64)(target - branch.ptr);
+        ASSERT_MSG(distance >= -0x80000000LL && distance < 0x80000000LL, "Jump target too far away, needs indirect register");
+        ((s32*)branch.ptr)[-1] = (s32)distance;
+    }
+}
+
 //Single byte opcodes
 //There is no PUSHAD/POPAD in 64-bit mode.
 void XEmitter::INT3() {Write8(0xCC);}
--- a/src/common/x64/emitter.cpp
+++ b/src/common/x64/emitter.cpp
@@ -455,6 +455,18 @@ void XEmitter::CALL(const void* fnptr)
     Write32(u32(distance));
 }
 
+FixupBranch XEmitter::CALL()
+{
+    FixupBranch branch;
+    branch.type = 1;
+    branch.ptr = code + 5;
+
+    Write8(0xE8);
+    Write32(0);
+
+    return branch;
+}
+
 FixupBranch XEmitter::J(bool force5bytes)
 {
     FixupBranch branch;
--- a/src/common/x64/emitter.h
+++ b/src/common/x64/emitter.h
@@ -425,6 +425,7 @@ public:
 #undef CALL
 #endif
     void CALL(const void* fnptr);
+    FixupBranch CALL();
     void CALLptr(OpArg arg);
 
     FixupBranch J_CC(CCFlags conditionCode, bool force5bytes = false);
--- a/src/common/x64/emitter.h
+++ b/src/common/x64/emitter.h
@@ -431,6 +431,7 @@ public:
     void J_CC(CCFlags conditionCode, const u8* addr, bool force5Bytes = false);
 
     void SetJumpTarget(const FixupBranch& branch);
+    void SetJumpTarget(const FixupBranch& branch, const u8* target);
 
     void SETcc(CCFlags flag, OpArg dest);
     // Note: CMOV brings small if any benefit on current cpus.
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -28,15 +28,8 @@ namespace Pica {
 namespace Shader {
 
 #ifdef ARCHITECTURE_x86_64
-static std::unordered_map<u64, CompiledShader*> shader_map;
-static JitCompiler jit;
-static CompiledShader* jit_shader;
-
-static void ClearCache() {
-    shader_map.clear();
-    jit.Clear();
-    LOG_INFO(HW_GPU, "Shader JIT cache cleared");
-}
+static std::unordered_map<u64, std::unique_ptr<JitCompiler>> shader_map;
+static const JitCompiler* jit_shader;
 #endif // ARCHITECTURE_x86_64
 
 void Setup(UnitState<false>& state) {
@@ -48,16 +41,12 @@ void Setup(UnitState<false>& state) {
 
         auto iter = shader_map.find(cache_key);
         if (iter != shader_map.end()) {
-            jit_shader = iter->second;
+            jit_shader = iter->second.get();
         } else {
-            // Check if remaining JIT code space is enough for at least one more (massive) shader
-            if (jit.GetSpaceLeft() < jit_shader_size) {
-                // If not, clear the cache of all previously compiled shaders
-                ClearCache();
-            }
-
-            jit_shader = jit.Compile();
-            shader_map.emplace(cache_key, jit_shader);
+            auto shader = std::make_unique<JitCompiler>();
+            shader->Compile();
+            jit_shader = shader.get();
+            shader_map[cache_key] = std::move(shader);
         }
     }
 #endif // ARCHITECTURE_x86_64
@@ -65,7 +54,7 @@ void Setup(UnitState<false>& state) {
 
 void Shutdown() {
 #ifdef ARCHITECTURE_x86_64
-    ClearCache();
+    shader_map.clear();
 #endif // ARCHITECTURE_x86_64
 }
 
@@ -109,7 +98,7 @@ OutputVertex Run(UnitState<false>& state, const InputVertex& input, int num_attr
 
 #ifdef ARCHITECTURE_x86_64
     if (VideoCore::g_shader_jit_enabled)
-        jit_shader(&state.registers);
+        jit_shader->Run(&state.registers);
     else
         RunInterpreter(state);
 #else
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -36,8 +36,7 @@ void Setup(UnitState<false>& state) {
 #ifdef ARCHITECTURE_x86_64
     if (VideoCore::g_shader_jit_enabled) {
         u64 cache_key = (Common::ComputeHash64(&g_state.vs.program_code, sizeof(g_state.vs.program_code)) ^
-            Common::ComputeHash64(&g_state.vs.swizzle_data, sizeof(g_state.vs.swizzle_data)) ^
-            g_state.regs.vs.main_offset);
+            Common::ComputeHash64(&g_state.vs.swizzle_data, sizeof(g_state.vs.swizzle_data)));
 
         auto iter = shader_map.find(cache_key);
         if (iter != shader_map.end()) {
@@ -98,7 +97,7 @@ OutputVertex Run(UnitState<false>& state, const InputVertex& input, int num_attr
 
 #ifdef ARCHITECTURE_x86_64
     if (VideoCore::g_shader_jit_enabled)
-        jit_shader->Run(&state.registers);
+        jit_shader->Run(&state.registers, g_state.regs.vs.main_offset);
     else
         RunInterpreter(state);
 #else
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -32,7 +32,7 @@ static std::unordered_map<u64, std::unique_ptr<JitCompiler>> shader_map;
 static const JitCompiler* jit_shader;
 #endif // ARCHITECTURE_x86_64
 
-void Setup(UnitState<false>& state) {
+void Setup() {
 #ifdef ARCHITECTURE_x86_64
     if (VideoCore::g_shader_jit_enabled) {
         u64 cache_key = (Common::ComputeHash64(&g_state.vs.program_code, sizeof(g_state.vs.program_code)) ^
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -28,8 +28,8 @@ namespace Pica {
 namespace Shader {
 
 #ifdef ARCHITECTURE_x86_64
-static std::unordered_map<u64, std::unique_ptr<JitCompiler>> shader_map;
-static const JitCompiler* jit_shader;
+static std::unordered_map<u64, std::unique_ptr<JitShader>> shader_map;
+static const JitShader* jit_shader;
 #endif // ARCHITECTURE_x86_64
 
 void Setup() {
@@ -42,7 +42,7 @@ void Setup() {
         if (iter != shader_map.end()) {
             jit_shader = iter->second.get();
         } else {
-            auto shader = std::make_unique<JitCompiler>();
+            auto shader = std::make_unique<JitShader>();
             shader->Compile();
             jit_shader = shader.get();
             shader_map[cache_key] = std::move(shader);
--- a/src/video_core/shader/shader_jit_x64.cpp
+++ b/src/video_core/shader/shader_jit_x64.cpp
@@ -741,7 +741,9 @@ void JitCompiler::Compile_Block(unsigned end) {
 void JitCompiler::Compile_NextInstr(unsigned* offset) {
     offset_ptr = offset;
 
-    Instruction instr = *(Instruction*)&g_state.vs.program_code[(*offset_ptr)++];
+    Instruction instr;
+    std::memcpy(&instr, &g_state.vs.program_code[(*offset_ptr)++], sizeof(Instruction));
+
     OpCode::Id opcode = instr.opcode.Value();
     auto instr_func = instr_table[static_cast<unsigned>(opcode)];
 
--- a/src/video_core/shader/shader_jit_x64.cpp
+++ b/src/video_core/shader/shader_jit_x64.cpp
@@ -845,6 +845,12 @@ void JitCompiler::Compile() {
         SetJumpTarget(branch.first, code_ptr[branch.second]);
     }
 
+    // Free memory that's no longer needed
+    return_offsets.clear();
+    return_offsets.shrink_to_fit();
+    fixup_branches.clear();
+    fixup_branches.shrink_to_fit();
+
     uintptr_t size = reinterpret_cast<uintptr_t>(GetCodePtr()) - reinterpret_cast<uintptr_t>(program);
     ASSERT_MSG(size <= MAX_SHADER_SIZE, "Compiled a shader that exceeds the allocated size!");
 
--- a/src/video_core/shader/shader_jit_x64.cpp
+++ b/src/video_core/shader/shader_jit_x64.cpp
@@ -810,7 +810,15 @@ void JitCompiler::FindReturnOffsets() {
 }
 
 void JitCompiler::Compile() {
+    // Reset flow control state
     program = (CompiledShader*)GetCodePtr();
+    program_counter = 0;
+    looping = false;
+    code_ptr.fill(nullptr);
+    fixup_branches.clear();
+
+    // Find all `CALL` instructions and identify return locations
+    FindReturnOffsets();
 
     // The stack pointer is 8 modulo 16 at the entry of a procedure
     ABI_PushRegistersAndAdjustStack(ABI_ALL_CALLEE_SAVED, 8);
@@ -832,15 +840,6 @@ void JitCompiler::Compile() {
     static const __m128 neg = { -0.f, -0.f, -0.f, -0.f };
     MOV(PTRBITS, R(RAX), ImmPtr(&neg));
     MOVAPS(NEGBIT, MatR(RAX));
-
-    // Find all `CALL` instructions and identify return locations
-    FindReturnOffsets();
-
-    // Reset flow control state
-    program_counter = 0;
-    looping = false;
-    code_ptr.fill(nullptr);
-    fixup_branches.clear();
 
     // Jump to start of the shader program
     JMPptr(R(ABI_PARAM2));
--- a/src/video_core/shader/shader_jit_x64.cpp
+++ b/src/video_core/shader/shader_jit_x64.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
 #include <smmintrin.h>
 
 #include "common/x64/abi.h"
@@ -760,8 +761,7 @@ void JitCompiler::Compile_Return() {
 }
 
 void JitCompiler::Compile_NextInstr() {
-    auto search = return_offsets.find(program_counter);
-    if (search != return_offsets.end()) {
+    if (std::binary_search(return_offsets.begin(), return_offsets.end(), program_counter)) {
         Compile_Return();
     }
 
@@ -793,10 +793,13 @@ void JitCompiler::FindReturnOffsets() {
         case OpCode::Id::CALL:
         case OpCode::Id::CALLC:
         case OpCode::Id::CALLU:
-            return_offsets.insert(instr.flow_control.dest_offset + instr.flow_control.num_instructions);
+            return_offsets.push_back(instr.flow_control.dest_offset + instr.flow_control.num_instructions);
             break;
         }
     }
+
+    // Sort for efficient binary search later
+    std::sort(return_offsets.begin(), return_offsets.end());
 }
 
 void JitCompiler::Compile() {
--- a/src/video_core/shader/shader_jit_x64.cpp
+++ b/src/video_core/shader/shader_jit_x64.cpp
@@ -138,6 +138,15 @@ static const u8 NO_SRC_REG_SWIZZLE = 0x1b;
 static const u8 NO_DEST_REG_MASK = 0xf;
 
 /**
+ * Get the vertex shader instruction for a given offset in the current shader program
+ * @param offset Offset in the current shader program of the instruction
+ * @return Instruction at the specified offset
+ */
+static Instruction GetVertexShaderInstruction(size_t offset) {
+    return { g_state.vs.program_code[offset] };
+}
+
+/**
  * Loads and swizzles a source register into the specified XMM register.
  * @param instr VS instruction, used for determining how to load the source register
  * @param src_num Number indicating which source register to load (1 = src1, 2 = src2, 3 = src3)
@@ -564,10 +573,23 @@ void JitCompiler::Compile_END(Instruction instr) {
 }
 
 void JitCompiler::Compile_CALL(Instruction instr) {
-    unsigned offset = instr.flow_control.dest_offset;
-    while (offset < (instr.flow_control.dest_offset + instr.flow_control.num_instructions)) {
-        Compile_NextInstr(&offset);
-    }
+    // Need to advance the return address past the proceeding instructions, this is the number of bytes to skip
+    constexpr unsigned SKIP = 21;
+    const uintptr_t start = reinterpret_cast<uintptr_t>(GetCodePtr());
+
+    // Push return address - not using CALL because we also want to push the offset of the return before jumping
+    MOV(64, R(RAX), ImmPtr(GetCodePtr() + SKIP));
+    PUSH(RAX);
+
+    // Push offset of the return
+    PUSH(32, Imm32(instr.flow_control.dest_offset + instr.flow_control.num_instructions));
+
+    // Jump
+    FixupBranch b = J(true);
+    fixup_branches.push_back({ b, instr.flow_control.dest_offset });
+
+    // Make sure that if the above code changes, SKIP gets updated
+    ASSERT(reinterpret_cast<uintptr_t>(GetCodePtr()) - start == SKIP);
 }
 
 void JitCompiler::Compile_CALLC(Instruction instr) {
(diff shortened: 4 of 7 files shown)


src/video_core/shader/shader_jit_x64.h
```diff
@@ -81,13 +85,33 @@ class JitCompiler : public Gen::XCodeBlock {
...
+     */
+    void AnalyzeCalls();
+
+    /// Mapping of Pica VS instructions to pointers in the emitted code
+    std::array<const u8*, 1024> code_ptr;
+
+    /// Offsets in code where a return needs to be inserted
+    std::set<unsigned> return_offsets;
+
+    unsigned last_program_counter;  ///< Offset of the most recent instruction decoded
```
lioncash: Should these be given default values of zero? Same as `program` below?

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -755,9 +777,25 @@ void JitCompiler::Compile_NextInstr(unsigned* offset) {
     }
 }
 
-CompiledShader* JitCompiler::Compile() {
-    const u8* start = GetCodePtr();
-    unsigned offset = g_state.regs.vs.main_offset;
+void JitCompiler::AnalyzeCalls() {
+    return_offsets.clear();
+
+    for (unsigned offset = 0; offset < g_state.vs.program_code.size(); ++offset) {
```
lioncash: `offset` should be a size_t

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -718,30 +727,43 @@ void JitCompiler::Compile_JMP(Instruction instr) {
...
+    auto search = return_offsets.find(program_counter);
+    if (search != return_offsets.end()) {
+        Compile_Return();
+    }
+
+    ASSERT_MSG(code_ptr[program_counter] == nullptr, "Compiled instruction already!");
+    code_ptr[program_counter] = GetCodePtr();
+
+    Instruction instr;
+    std::memcpy(&instr, &g_state.vs.program_code[program_counter++], sizeof(Instruction));
```
lioncash: Perhaps this can be extracted to its own function, `GetVertexShaderInstruction` which can be something like:

```cpp
Instruction GetVertexShaderInstruction(size_t offset)
{
    Instruction instruction;
    std::memcpy(&instruction, &g_state.vs.program_code[offset], sizeof(Instruction));
    return instruction;
}
```

since a similar piece of code is used in AnalyzeCalls as well.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -780,21 +818,36 @@ CompiledShader* JitCompiler::Compile() {
...
+    Compile_Block(static_cast<unsigned>(g_state.vs.program_code.size()));
 
-JitCompiler::JitCompiler() {
-    AllocCodeSpace(jit_cache_size);
+    // Set the target for any incomplete branches now that the entire shader program has been emitted
+    for (const auto& branch : fixup_branches) {
+        SetJumpTarget(branch.first, code_ptr[branch.second]);
+    }
+
+    uintptr_t size = reinterpret_cast<uintptr_t>(GetCodePtr()) - reinterpret_cast<uintptr_t>(program);
```
archshift: `ptrdiff_t size`?

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -755,9 +777,25 @@ void JitCompiler::Compile_NextInstr(unsigned* offset) {
...
+        case OpCode::Id::CALLC:
+        case OpCode::Id::CALLU:
+            return_offsets.insert(instr.flow_control.dest_offset + instr.flow_control.num_instructions);
+            break;
+        }
+    }
+}
+
+void JitCompiler::Compile() {
+    program = reinterpret_cast<CompiledShader*>(GetCodePtr());
```
archshift: `reinterpret_cast<const CompiledShader*>`

Don't want to cast away that `const`!

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -755,9 +787,24 @@ void JitCompiler::Compile_NextInstr(unsigned* offset) {
     }
 }
 
-CompiledShader* JitCompiler::Compile() {
-    const u8* start = GetCodePtr();
-    unsigned offset = g_state.regs.vs.main_offset;
+void JitCompiler::AnalyzeCalls() {
```
JayFoxRox: I'd prefer something like "CollectReturnOffsets" or something.
"AnalyzeCalls" sounds too generic to me (It even sounds like something performance related to me).

(Also if "IF" or "LOOP" are added later we'd have to rename this)

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -718,30 +738,42 @@ void JitCompiler::Compile_JMP(Instruction instr) {
 
     bool inverted_condition = (instr.opcode.Value() == OpCode::Id::JMPU) &&
         (instr.flow_control.num_instructions & 1);
-    FixupBranch b = J_CC(inverted_condition ? CC_Z : CC_NZ, true);
-
-    Compile_Block(instr.flow_control.dest_offset);
 
-    SetJumpTarget(b);
+    fixup_branches.push_back({ J_CC(inverted_condition ? CC_Z : CC_NZ, true), instr.flow_control.dest_offset });
```
JayFoxRox: I'd prefer if the emitter call would be seperated from the push_back (create a variable) so it's more obvious that this line emits code.
Or maybe something like Compile_Branch(condition, destination) ?

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -564,10 +575,22 @@ void JitCompiler::Compile_END(Instruction instr) {
...
+
+    // Push return address - not using CALL because we also want to push the offset of the return before jumping
+    MOV(64, R(RAX), ImmPtr(GetCodePtr() + SKIP));
+    PUSH(RAX);
+
+    // Push offset of the return
+    PUSH(32, Imm32(instr.flow_control.dest_offset + instr.flow_control.num_instructions));
+
+    // Jump
+    fixup_branches.push_back({ J(true), instr.flow_control.dest_offset });
```
JayFoxRox: Same as below. (Make it more obvious that this emits code).

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -780,21 +827,36 @@ CompiledShader* JitCompiler::Compile() {
...
+    program_counter = 0;
     looping = false;
+    code_ptr.fill(nullptr);
+    fixup_branches.clear();
 
-    while (offset < g_state.vs.program_code.size()) {
-        Compile_NextInstr(&offset);
+    // Jump to start of the shader program
+    if (g_state.regs.vs.main_offset != 0) {
+        fixup_branches.push_back({ J(true),  g_state.regs.vs.main_offset });
```
JayFoxRox: Can't we use code_ptr to pass the main_offset to the shader at runtime?

Some games have ubershaders (and most games I've seen only have around 4-10 shaders anyway) but the entry point will control which shader to run.
I believe VS and GS are sometimes sharing code too.
So we could probably reduce a lot of compilations by having a dynamic entry point (and code_ptr seems to make it possible?!)

bunnei:
Addressed all comments.

src/video_core/shader/shader_jit_x64.h
```diff
@@ -19,12 +22,10 @@ namespace Pica {
...
 
-/// Memory needed to be available to compile the next shader (otherwise, clear the cache)
-constexpr size_t jit_shader_size = 1024 * 512;
-/// Memory allocated for the JIT code space cache
-constexpr size_t jit_cache_size = 1024 * 1024 * 8;
+/// Memory allocated for each compiled shader (64Kb)
+constexpr size_t MAX_SHADER_SIZE = 1024 * 64;
 
-using CompiledShader = void(void* registers);
+using CompiledShader = void(void* registers, const u8* start_addr);
```
MerryMage: `CompiledShader` is now internal to `JitCompiler`, maybe bring it into the class. [minor]

MerryMage:
LGTM. Skimmed the logic a bit as I'm not 100% on it.

src/video_core/shader/shader.cpp
```diff
@@ -29,36 +29,24 @@ namespace Pica {
...
-static void ClearCache() {
-    shader_map.clear();
-    jit.Clear();
-    LOG_INFO(HW_GPU, "Shader JIT cache cleared");
-}
+static std::unordered_map<u64, std::unique_ptr<JitCompiler>> shader_map;
+static const JitCompiler* jit_shader;
 #endif // ARCHITECTURE_x86_64
 
 void Setup(UnitState<false>& state) {
```
JayFoxRox: `state` isn't used anymore here. Parameter can be removed.

src/video_core/shader/shader_jit_x64.h
JayFoxRox: Agreed. Bring `CompiledShader` into JitCompiler.

JayFoxRox:
(Thanks for waiting for my review and for putting so much effort into Citra in general)

I merged this with GS and tested:

* The Legend Of Korra: works
* WWE All Stars: works (only works with interpreter in current master)
* Ecco The Dolphin 3D: works
* Alien on the run: works
* Return to PopoCrolois: works (had glitches with interpreter in current master)

I've run some other games too but nothing weird happened.

---

Me and bunnei discussed replacing the compile time asserts by ABI function calls where the called function would `LOG_ERROR`. This sounds safer because the new approach might process dead code  which could trigger assertions otherwise.
In particular it was [this assert](https://github.com/citra-emu/citra/pull/1546/files#diff-57ec0b2a1b7ff70f4fd20bb37399827bR672).

So if that issue is addressed + what I recently diff-commented this LGTM.

src/common/x64/emitter.h
```diff
@@ -430,7 +430,7 @@ class XEmitter
     FixupBranch J_CC(CCFlags conditionCode, bool force5bytes = false);
     void J_CC(CCFlags conditionCode, const u8* addr, bool force5Bytes = false);
 
-    void SetJumpTarget(const FixupBranch& branch);
+    void SetJumpTarget(const FixupBranch& branch, const u8* target = nullptr);
```
yuriks: I'd drop the default param and add a small inline overload instead. This gets rid of the if, and then you can set jumps to null to your heart's content too!

src/video_core/shader/shader.cpp
```diff
@@ -29,36 +29,24 @@ namespace Pica {
...
-static std::unordered_map<u64, CompiledShader*> shader_map;
-static JitCompiler jit;
-static CompiledShader* jit_shader;
-
-static void ClearCache() {
-    shader_map.clear();
-    jit.Clear();
-    LOG_INFO(HW_GPU, "Shader JIT cache cleared");
-}
+static std::unordered_map<u64, std::unique_ptr<JitCompiler>> shader_map;
```
yuriks: ~~Stuff about thes being confusing and separating the essential members into a non-compiler struct~~

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -138,6 +138,27 @@ static const u8 NO_SRC_REG_SWIZZLE = 0x1b;
...
+    return instruction;
+}
+
+static void LogCritical(const char* msg) {
+    LOG_CRITICAL(HW_GPU, msg);
+}
+
+void JitCompiler::RuntimeAssert(bool condition, const char* msg) {
+    if (!condition) {
+        ABI_CallFunctionP(LogCritical, (void*)(msg));
```
lioncash: const_cast

src/video_core/shader/shader_jit_x64.h
```diff
@@ -81,13 +83,40 @@ class JitCompiler : public Gen::XCodeBlock {
...
+
     BitSet32 PersistentCallerSavedRegs();
 
-    /// Pointer to the variable that stores the current Pica code offset. Used to handle nested code blocks.
-    unsigned* offset_ptr = nullptr;
+    /**
+     * Assertion evaluated at compile-time, but only triggered if executed at runtime.
+     * @param msg Message to be logged if the assertion fails.
+     */
+    void RuntimeAssert(bool condition, char* msg);
```
lioncash: Missing const? (the implementation specifies it)

src/video_core/shader/shader_jit_x64.cpp
yuriks: This is more of a reinterpret_cast

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -138,6 +138,27 @@ static const u8 NO_SRC_REG_SWIZZLE = 0x1b;
 static const u8 NO_DEST_REG_MASK = 0xf;
 
 /**
+ * Get the vertex shader instruction for a given offset in the current shader program
+ * @param offset Offset in the current shader program of the instruction
+ * @return Instruction at the specified offset
+ */
+static Instruction GetVertexShaderInstruction(size_t offset) {
+    Instruction instruction;
+    std::memcpy(&instruction, &g_state.vs.program_code[offset], sizeof(Instruction));
```
yuriks: You can just do `Instruct instruction{ &g_state.vs.program_code[offset] };`

src/video_core/shader/shader_jit_x64.cpp
lioncash: Nope. This is a const_cast. It's well-defined.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -157,7 +178,7 @@ void JitCompiler::Compile_SwizzleSrc(Instruction instr, unsigned src_num, Source
     }
 
     int src_offset_disp = (int)src_offset;
-    ASSERT_MSG(src_offset == src_offset_disp, "Source register offset too large for int type");
+    RuntimeAssert(src_offset == src_offset_disp, "Source register offset too large for int type");
```
yuriks: What is this assert even checking against? It's kinda unclear, but I think this would be fine as a regular assert.

src/video_core/shader/shader_jit_x64.h
bunnei: addressed

JohnGodgames:
Ok so **Pokémon Rumble Blast** seems to pretty stable with this PR [video](https://www.youtube.com/watch?v=SstTsFoFnvo)

**Next tests:**
- Final Fantasy Explorers seems to stable too - played up to first quest

src/video_core/shader/shader_jit_x64.cpp
JayFoxRox: True, this shouldn't be a runtime assert. Actually it shouldn't even be possible to happen at all (because the offset is based on the struct size which is known at compile time and all members should be reachable or we have a huge problem anyway).
I'd remove this assert altogether.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -242,7 +252,7 @@ void JitCompiler::Compile_DestEnable(Instruction instr,X64Reg src) {
     SwizzlePattern swiz = { g_state.vs.swizzle_data[operand_desc_id] };
 
     int dest_offset_disp = (int)UnitState<false>::OutputOffset(dest);
-    ASSERT_MSG(dest_offset_disp == UnitState<false>::OutputOffset(dest), "Destinaton offset too large for int type");
+    RuntimeAssert(dest_offset_disp == UnitState<false>::OutputOffset(dest), "Destinaton offset too large for int type");
```
JayFoxRox: dito (can't happen and if it happens we have a different problem). Maybe we should static assert the size of the accessable data in UnitState?

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -564,10 +583,23 @@ void JitCompiler::Compile_END(Instruction instr) {
 }
 
 void JitCompiler::Compile_CALL(Instruction instr) {
-    unsigned offset = instr.flow_control.dest_offset;
-    while (offset < (instr.flow_control.dest_offset + instr.flow_control.num_instructions)) {
-        Compile_NextInstr(&offset);
-    }
+    // Need to advance the return address past the proceeding instructions, this is the number of bytes to skip
+    constexpr unsigned SKIP = 21;
+    const ptrdiff_t start = reinterpret_cast<ptrdiff_t>(GetCodePtr());
```
yuriks: I you want `uintptr_t` (or maybe `intptr_t`? I never used this one) instead here. Since the computation doesn't actually underflow uintptr_t is probably fine.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -564,10 +583,23 @@ void JitCompiler::Compile_END(Instruction instr) {
...
+
+    // Push return address - not using CALL because we also want to push the offset of the return before jumping
```
yuriks: Mismatching CALL/RET by replacing with JMP/RET is a bad idea since it messes with the return predictor. See [this blog post](https://blogs.msdn.microsoft.com/oldnewthing/20041216-00/?p=36973) and section 3.15 of [Agner's manual](http://www.agner.org/optimize/microarchitecture.pdf). Try to either use a matching CALL/RET pair or two JMPs.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -718,30 +745,43 @@ void JitCompiler::Compile_JMP(Instruction instr) {
...
+    auto search = return_offsets.find(program_counter);
+    if (search != return_offsets.end()) {
+        Compile_Return();
+    }
 
-    // Restore current offset pointer
-    offset_ptr = prev_offset_ptr;
-    *offset_ptr = offset;
-}
+    ASSERT_MSG(code_ptr[program_counter] == nullptr, "Compiled instruction already!");
```
yuriks: "Tried to compile already compiled shader location" or something to that effect.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -718,30 +745,43 @@ void JitCompiler::Compile_JMP(Instruction instr) {
...
-void JitCompiler::Compile_Block(unsigned end) {
-    // Save current offset pointer
-    unsigned* prev_offset_ptr = offset_ptr;
-    unsigned offset = *prev_offset_ptr;
+void JitCompiler::Compile_NextInstr() {
+    last_program_counter = program_counter;
 
-    while (offset < end)
-        Compile_NextInstr(&offset);
+    auto search = return_offsets.find(program_counter);
```
yuriks: To appease my O(n*m) concerns a bit: what's the maximum size this vector reaches usually? You could use sort it once after finding all addresses and then use std::binary_search, or maybe use a std::bitset.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -718,30 +745,43 @@ void JitCompiler::Compile_JMP(Instruction instr) {
...
-    Compile_Block(instr.flow_control.dest_offset);
+void JitCompiler::Compile_Block(unsigned end) {
+    while (program_counter < end) {
+        Compile_NextInstr();
+    }
+}
+
+void JitCompiler::Compile_Return() {
+    // Pop return offset off the stack and check if we're at that offset
+    POP(RAX);
```
yuriks: The POP, check, PUSH sequence is a bit weird. Should just check against the value in memory directly using SP-relative addressing.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -780,21 +835,34 @@ CompiledShader* JitCompiler::Compile() {
...
 
+    // Find all `CALL` instructions and identify return locations
+    FindReturnOffsets();
+
+    // Reset flow control state
+    last_program_counter = 0;
+    program_counter = 0;
     looping = false;
+    code_ptr.fill(nullptr);
+    fixup_branches.clear();
```
yuriks: You should do all this initialization before any code generation. I find it somewhat confusing/arbitrary to interleave it with other JIT (deferred) instructions.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -780,21 +835,34 @@ CompiledShader* JitCompiler::Compile() {
...
+
+    // Set the target for any incomplete branches now that the entire shader program has been emitted
+    for (const auto& branch : fixup_branches) {
+        SetJumpTarget(branch.first, code_ptr[branch.second]);
     }
 
-    return (CompiledShader*)start;
+    ptrdiff_t size = reinterpret_cast<ptrdiff_t>(GetCodePtr()) - reinterpret_cast<ptrdiff_t>(program);
+    ASSERT(size <= MAX_SHADER_SIZE);
+    LOG_INFO(HW_GPU, "Compiled shader size=%d", size);
```
yuriks: Is this really important enough to be INFO?

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -780,21 +835,34 @@ CompiledShader* JitCompiler::Compile() {
...
+    Compile_Block(static_cast<unsigned>(g_state.vs.program_code.size()));
+
+    // Set the target for any incomplete branches now that the entire shader program has been emitted
+    for (const auto& branch : fixup_branches) {
+        SetJumpTarget(branch.first, code_ptr[branch.second]);
     }
 
-    return (CompiledShader*)start;
+    ptrdiff_t size = reinterpret_cast<ptrdiff_t>(GetCodePtr()) - reinterpret_cast<ptrdiff_t>(program);
+    ASSERT(size <= MAX_SHADER_SIZE);
```
yuriks: Could be an ASSERT_MSG

src/video_core/shader/shader_jit_x64.h
```diff
@@ -81,13 +83,40 @@ class JitCompiler : public Gen::XCodeBlock {
...
+     */
+    void FindReturnOffsets();
+
+    /// Mapping of Pica VS instructions to pointers in the emitted code
+    std::array<const u8*, 1024> code_ptr;
+
+    /// Offsets in code where a return needs to be inserted
+    std::set<unsigned> return_offsets;
+
+    unsigned last_program_counter = 0;  ///< Offset of the most recent instruction decoded
```
yuriks: It's unclear why you introduced this variable. Is there any specific reason?

src/video_core/shader/shader_jit_x64.cpp
yuriks: Oops, didn't realize this was a std::map. In any case, I think a sorted vector or a bitset as I mentioned above are better options.

yuriks:
Review pass done.

src/video_core/shader/shader_jit_x64.cpp
bunnei: Does this really matter..? Seems to me that using the map here is concise and fine.

bunnei:
All comments addressed. Ready for another review pass.

src/video_core/shader/shader.cpp
JayFoxRox: @yuriks that's true, but it should be a seperate issue.
My GS branch is based on this PR and it would suck to refactor any of them again before a merge to fix this general design issue.
This PR certainly improves what's contained in master and what you are saying is not really related to this PR.

src/common/x64/emitter.h
```diff
@@ -425,12 +425,14 @@ class XEmitter
...
 #endif
     void CALL(const void* fnptr);
+    FixupBranch CALL();
     void CALLptr(OpArg arg);
 
     FixupBranch J_CC(CCFlags conditionCode, bool force5bytes = false);
     void J_CC(CCFlags conditionCode, const u8* addr, bool force5Bytes = false);
 
     void SetJumpTarget(const FixupBranch& branch);
+    void SetJumpTarget(const FixupBranch& branch, const u8* target);
```
JayFoxRox: Might be worth another (nit) issue: why does this take a `const u8*` called `target` while it's called `addr` for `J_CC`?
And worse: why is it a `const void*` called `fnptr` for `CALL`?
I think the argument for this should be called `addr` and `CALL` should also be updated with the new name at some point.
\- I don't have any type preference.

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -138,6 +138,25 @@ static const u8 NO_SRC_REG_SWIZZLE = 0x1b;
...
+ */
+static Instruction GetVertexShaderInstruction(size_t offset) {
+    return { g_state.vs.program_code[offset] };
+}
+
+static void LogCritical(const char* msg) {
+    LOG_CRITICAL(HW_GPU, msg);
+}
+
+void JitCompiler::RuntimeAssert(bool condition, const char* msg) {
```
JayFoxRox: Maybe call this: "Compile_Assert" or something?
It should be clear that this actually generates code imho.

(Also a comment wouldn't hurt maybe)

JayFoxRox:
Aside from what I mentioned LGTM.
Still works fine with my GS branch.

I think next time you change something you should squash some changes together, the list of fixups is pretty long by now (also git thought you force pushed some of them when I rebased so I had to remove them to avoid duplicates?!)

src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -583,23 +583,15 @@ void JitCompiler::Compile_END(Instruction instr) {
...
-    // Need to advance the return address past the proceeding instructions, this is the number of bytes to skip
-    constexpr unsigned SKIP = 21;
-    const uintptr_t start = reinterpret_cast<uintptr_t>(GetCodePtr());
-
-    // Push return address - not using CALL because we also want to push the offset of the return before jumping
-    MOV(64, R(RAX), ImmPtr(GetCodePtr() + SKIP));
-    PUSH(RAX);
-
     // Push offset of the return
     PUSH(32, Imm32(instr.flow_control.dest_offset + instr.flow_control.num_instructions));
```
yuriks: This is actually a 64-bit PUSH, but the bits param is ignored by our emitter (and then the 32-bit immediate is sign-extended in 64-bit mode). Would be better to make it match just to make it less confusing.

src/common/x64/emitter.h
yuriks: IMO `target` is a much, much clearer name, it's even in the function name. They're all pretty obvious in context but I don't see any reason for making the name worse for consistency when the other functions aren't even consistent in the first place.

src/video_core/shader/shader.cpp
yuriks: These are all things that changed in this PR in the first place. I don't really understand what you're saying, but maybe you misunderstood what I was referring to.

yuriks: After investigating a bit, I don't think we can actually split the compilation from the memory management of the block with the code as-is, so I'm willing to concede this since it becomes out of scope. You can either make the emitter be able to decouple from the codeblock and return it, so we can transfer it to a new object, or create a dedicated holder for the block and then call emitter functions on it, but those are both large-ish changes.

However, you can still make sure to clear and shrink-to-fit any data structures that aren't used after compilation so we leave less garbage hanging off the object.

bunnei:
Second pass of comments addressed. Ready for final review.

src/video_core/shader/shader_jit_x64.cpp
yuriks: In case anyone ever checks this again: (We already resolved this on IRC)
 - std::set does one separate allocation for each unsigned, which adds execution and memory overhead
 - std::unordered_set is the same
 - The std::vector solution is only one extra line of code

yuriks:
I bestow upon thee the holy LGTM

src/video_core/shader/shader.cpp
JayFoxRox: Considering this is mostly a bugfix PR (+ tiny new features) I'm against this change.
I like the general idea though. But please split it into a seperate issue / PR.

src/common/x64/emitter.cpp
```diff
@@ -531,6 +543,22 @@ void XEmitter::SetJumpTarget(const FixupBranch& branch)
...
+    {
+        s64 distance = (s64)(target - branch.ptr);
+        ASSERT_MSG(distance >= -0x80 && distance < 0x80, "Jump target too far away, needs force5Bytes = true");
+        branch.ptr[-1] = (u8)(s8)distance;
+    }
+    else if (branch.type == 1)
+    {
+        s64 distance = (s64)(target - branch.ptr);
+        ASSERT_MSG(distance >= -0x80000000LL && distance < 0x80000000LL, "Jump target too far away, needs indirect register");
+        ((s32*)branch.ptr)[-1] = (s32)distance;
```
lioncash: memcpy

bunnei: this is just a copy-pasta of code prevalent throughout the rest of the emitter. I think it makes sense to just clean up the emitter in a subsequent PR.

PR #1550, opened 2016-03-20
Fix missing header

LittleWhite-tb:
gdbstub.h could not be used without including common_types.h as noticed in PR #1368

### DIFF
--- a/src/core/gdbstub/gdbstub.h
+++ b/src/core/gdbstub/gdbstub.h
@@ -7,6 +7,8 @@
 #pragma once
 #include <atomic>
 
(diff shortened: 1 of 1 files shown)


src/core/gdbstub/gdbstub.h
```diff
@@ -7,6 +7,8 @@
 #pragma once
 #include <atomic>
 
+#include <common/common_types.h>
```
lioncash: `"common/common_type.h"`

LittleWhite-tb:
Fixed

bunnei:
LGTM

PR #1551, opened 2016-03-20
PICA: Use VS config to get immediate mode attribute count

JayFoxRox:
Fixes Crazy Kangaroo ([Looked like this](http://i.imgur.com/KPT8ixe.png)) - and hopefully more

Needs regression testing (Only with games which use immediate mode!)

**Source of problem:**
`attribute_config.GetNumTotalAttributes()` returns 1 for Crazy Kangaroo (meaning the register has a value of 0). The game uses 3 attributes in the VS (o0, o1, o2), but the game still uploads 4 attributes (= 4 * 3 registers). This was easily observed by looking at the CMD list.

**Possible reason**
I believe `attribute_config.GetNumTotalAttributes()` only indicates how many vertex attributes are loaded from the vertex buffer (= excluding fixed attributes and only when vertex buffers are in use).
\- The newly introduced shader config however should tells us how many attributes in total (= fixed + from buffer) are loaded by the shader. The game configures the VS input register to 3 (= total attributes 4).
There might be another register which sets the immediate attribute output count but I couldn't find it yet (and it would most likely match the VS input value).

Chances are the `attribute_config.GetNumTotalAttributes()` also breaks games which use vertex buffers together with fixed attributes. This PR only changes it for immediate mode as no problems could be confirmed in games using vertex buffers (yet).

This all came from observation. Though [3dbrew also has the shader config register](https://www.3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_SH_INPUTBUFFER_CONFIG).

**Fixed**
Menu
[More screenshots](http://imgur.com/a/iZEj6) (HW renderer has some issues with this game, unrelated to PR)

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -131,14 +131,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
 
                     immediate_input.attr[immediate_attribute_id++] = attribute;
 
-                    if (immediate_attribute_id >= attribute_config.GetNumTotalAttributes()) {
+                    if (immediate_attribute_id >= g_state.regs.vs.GetNumTotalAttributes()) {
                         immediate_attribute_id = 0;
 
                         Shader::UnitState<false> shader_unit;
                         Shader::Setup(shader_unit);
 
+                        if (g_debug_context)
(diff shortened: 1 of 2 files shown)


JayFoxRox:
Duplicate of #1461

PR #1554, opened 2016-03-20
[RFC] Very incomplete CPU JIT

MerryMage:
Refactored the very quick educational JIT I wrote into something more presentable and much less rushed.

Things in this PR:
* [Catch](https://github.com/philsquared/Catch) for unit tests
* Unit test that fuzzes the JIT against the interpreter for data processing instructions.
* New generic ARM decoder that can be adapted for code analysis / optimization in the future. (I stole the original tables from @lioncash then manged them beyond all recognition, apologies.)
* A JIT that falls back on the interpreter when it can't compile an instruction.
* ClearCache. (CROs require this, but I also found it useful for flushing after tests.)

Things not yet in this PR:
* [x] Non-immediate data processing instructions.
* [x] Branch instructions.
* [x] Porting other instructions implemented in the other prototype over.
* [x] Unit tests for the Thumb decoder. It might currently be complete garbage for all I know.

Things that won't be in this PR:
* A way to actually use switch to the JIT to run games.
* A decent register allocator.

---

### Design

* Current design inspired by something @lioncash was working on but probably not as good as theirs.
* Each instruction is handled by it's own function. 
* Everything to do with register allocation and assignment is handled by RegAlloc.
* JitState has an ARMul_State embedded in it for ease of calling the interpreter.
* CallInterpreter can be adjusted to interpret more than 1 instruction at a time.
* I'm tired and I originally had an idea for a design writeup in mind but can't recall what I wanted to write at the moment.
* Possibly point out places where there is a lack of comments please / Where something is unclear.
* Design comments welcome.

### Known issues

* ~~Interpreter cache pollution.~~ (Fixed by #1618)


src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,404 @@
...
+            std::array<u32, N> values;
+            for (int i = 0; i < N; i++) {
+                values[i] = (inst & masks[i]) >> shifts[i];
+            }
+            Call<Function, N, decltype(values)>::call(v, fn, values);
+        }
+    };
+}
+
+template<size_t N, typename Function>
```
Subv: What does the N represent?

src/core/arm/jit_x64/instructions/data_processing.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+void JitX64::CompileDataProcessingHelper(ArmReg Rn_index, ArmReg Rd_index, std::function<void(X64Reg)> body) {
+    // The major consideration is if Rn and/or Rd == R15.
+
+    if (Rd_index == 15) {
+        X64Reg Rd = reg_alloc.AllocAndLockTemp();
+        reg_alloc.LockArm(Rn_index);
+
+        if (Rn_index == 15) {
+            // TODO: In this case we can actually calculat the final result.
```
Subv: calculate :D

src/core/arm/jit_x64/interpret.cpp
```diff
@@ -0,0 +1,74 @@
...
+    ASSERT(reg_alloc.JitStateReg() != RSP);
+    code->MOV(64, R(RSP), MJitStateHostReturnRSP());
+    code->MOV(64, R(Gen::ABI_PARAM1), R(reg_alloc.JitStateReg()));
+    code->MOV(64, R(Gen::ABI_PARAM2), Imm64(current.arm_pc));
+    code->MOV(64, R(Gen::ABI_PARAM3), Imm64(current.TFlag));
+    code->MOV(64, R(Gen::ABI_PARAM4), Imm64(current.EFlag));
+
+    const void *const fn = &CallInterpreter;
+
+    const u64 distance = reinterpret_cast<u64>(fn) - (reinterpret_cast<u64>(code->GetCodePtr()) + 5);
```
Subv: shouldn't this be ptrdiff_t?

src/core/arm/decoder/arm.cpp
MerryMage: `NumberOfFunctionArguments`

Good point, this could be documented better.

`cccc0000100Snnnnddddssss0rr1mmmm` for example has c (cond), S (S), n (Rn), d (Rd), s (Rs), r (shift type) and m (Rm), so the function supplied must have 7 arguments.

src/core/arm/decoder/decoder.h
```diff
@@ -0,0 +1,327 @@
...
+    }
+    virtual void visit(Visitor* v, u32 inst) = 0;
+};
+
+class Instruction {
+private:
+    const std::unique_ptr<Matcher> matcher;
+
+public:
+    Instruction(char* name, std::unique_ptr<Matcher> matcher) : name(name), matcher(std::move(matcher)) {}
```
archshift: `const char* name`

src/core/arm/jit_x64/common.h
```diff
@@ -0,0 +1,39 @@
...
+
+struct JitState {
+    JitState() : cpu_state(PrivilegeMode::USER32MODE) {}
+    void Reset() {
+        cpu_state.Reset();
+    }
+
+    ARMul_State cpu_state;
+
+    void* bb;
```
MerryMage: This no longer serves a purpose and will be removed.

src/core/arm/jit_x64/interface.cpp
```diff
@@ -0,0 +1,231 @@
...
+    down_count -= ticks;
+    if (down_count < 0)
+        CoreTiming::Advance();
+}
+
+void ARM_Jit::ExecuteInstructions(int num_instructions) {
+    reschedule = false;
+
+    do {
+        //state->page_table = reinterpret_cast<void*>(Memory::current_page_table->pointers.data());
```
MerryMage: Commented code left over from porting a previous version that had memory load/stores: Will remove.

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+#include "core/arm/decoder/decoder.h"
+#include "core/arm/jit_x64/common.h"
+#include "core/arm/jit_x64/reg_alloc.h"
+
+namespace JitX64 {
+
+using CodePtr = u8*;
+
+struct LocationDescriptor {
+    u32 arm_pc;
```
lioncash: these should likely have default values just in case, if the comparison operator is implemented.

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+
+    /// Returns a pointer to the compiled basic block.
+    CodePtr Compile(u32 pc, bool TFlag, bool EFlag);
+
+private:
+    LocationDescriptor current;
+    unsigned instructions_compiled;
+    bool stop_compilation;
+
+    size_t GetInstSize() { return current.TFlag ? 2 : 4; }
```
lioncash: this can be const

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+    Gen::XEmitter* code;
+
+    RegAlloc reg_alloc;
+
+    /// ARM pc -> x64 code block
+    std::unordered_map<LocationDescriptor, CodePtr, LocationDescriptorHash> basic_blocks;
+
+public:
+    JitX64() = delete;
+    JitX64(Gen::XEmitter* code_);
```
lioncash:
```cpp
explicit JitX64(Gen::XEmitter* code_)
```

src/core/arm/jit_x64/reg_alloc.cpp
```diff
@@ -0,0 +1,361 @@
...
+    { Gen::R10, 10 },
+    { Gen::R11, 11 },
+    { Gen::R12, 12 },
+    { Gen::R13, 13 },
+    { Gen::R14, 14 },
+};
+
+constexpr Gen::X64Reg jit_state_reg = Gen::R15;
+
+Gen::X64Reg RegAlloc::JitStateReg() {
```
lioncash: this can be a const qualified function

src/core/arm/jit_x64/reg_alloc.h
```diff
@@ -0,0 +1,105 @@
...
+namespace JitX64 {
+
+// This is the register allocator
+// TODO: Better algorithm required, ideally something that does reigster usage lookahead.
+//       (Designed so it should be simple to implement later.)
+
+class RegAlloc final {
+private:
+    struct ArmState {
+        Gen::OpArg location;
```
lioncash: Should these have default values? Same with X64State.

src/core/arm/jit_x64/reg_alloc.h
```diff
@@ -0,0 +1,105 @@
...
+            MemoryMap,
+            UserManuallyLocked
+        };
+
+        bool locked;
+        State state;
+        ArmReg arm_reg;
+    };
+
+    std::array<ArmState, 15> arm_gpr;
```
lioncash: Mainly asking above, because the data in these arrays on initialization are in an inconsistent state.

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+        void Always();
+        void FlagsDirty();
+        ConditionCode CurrentCond();
+    } cond_manager;
+
+private:
+    void CompileInterpretInstruction();
+
+    // Barrier instructions
+    virtual void DMB() override;
```
lioncash: You don't need to specify virtual on funcs with override. The virtual is implicit.

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+    virtual void TEQ_reg(Cond cond,         ArmReg Rn,            ArmImm5 imm5, ShiftType shift, ArmReg Rm) override;
+    virtual void TEQ_rsr(Cond cond,         ArmReg Rn,            ArmReg Rs, ShiftType shift, ArmReg Rm) override;
+    virtual void TST_imm(Cond cond,         ArmReg Rn,            int rotate, ArmImm8 imm8) override;
+    virtual void TST_reg(Cond cond,         ArmReg Rn,            ArmImm5 imm5, ShiftType shift, ArmReg Rm) override;
+    virtual void TST_rsr(Cond cond,         ArmReg Rn,            ArmReg Rs, ShiftType shift, ArmReg Rm) override;
+
+    // Exception generation instructions
+    virtual void BKPT() override;
+    virtual void HVC() override;
+    virtual void SMC() override;
```
lioncash: SMC and HVC aren't present in ARMv6K

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+    virtual void UXTAB16() override;
+    virtual void UXTAH() override;
+    virtual void UXTB() override;
+    virtual void UXTB16() override;
+    virtual void UXTH() override;
+
+    // Hint instructions
+    virtual void DBG() override;
+    virtual void PLD() override;
+    virtual void PLI() override;
```
lioncash: PLI and DBG are ARMv7 only

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+
+    // Saturation instructions
+    virtual void SSAT() override;
+    virtual void SSAT16() override;
+    virtual void USAT() override;
+    virtual void USAT16() override;
+
+    // Multiply (Normal) instructions
+    virtual void MLA() override;
+    virtual void MLS() override;
```
lioncash: MLS isn't in ARMv6K

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+
+    // Unsigned sum of absolute difference functions
+    virtual void USAD8() override;
+    virtual void USADA8() override;
+
+    // Packing instructions
+    virtual void PKH() override;
+
+    // Reversal instructions
+    virtual void RBIT() override;
```
lioncash: RBIT isn't in ARMv6K

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+    virtual void STRHT() override;
+    virtual void STRT() override;
+
+    // Load/Store multiple instructions
+    virtual void LDM() override;
+    virtual void STM() override;
+
+    // Miscellaneous instructions
+    virtual void CLZ() override;
+    virtual void ERET() override;
```
lioncash: ERET isn't in ARMv6K

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,404 @@
...
+    { "MSR (reg)",           MakeMatcher<0>("----00010010--00111100000000----", &Visitor::MSR) },
+    { "MSR (banked)",        MakeMatcher<0>("----00010-10----1111001-0000----", &Visitor::MSR) },
+    { "MSR (imm special)",   MakeMatcher<0>("----00110-10----1111------------", &Visitor::MSR) },
+    { "MSR (reg special)",   MakeMatcher<0>("----00010-10----111100000000----", &Visitor::MSR) },
+    { "RFE",                 MakeMatcher<0>("----0001101-0000---------110----", &Visitor::RFE) },
+    { "SRS",                 MakeMatcher<0>("0000011--0-00000000000000001----", &Visitor::SRS) },
+}};
+
+const Instruction& DecodeArm(u32 i) {
+    return *std::find_if(arm_instruction_table.cbegin(), arm_instruction_table.cend(), [i](const auto& instruction) {
```
lioncash: This should likely return the iterator, which the caller should check after. If you ever hit the end of the table, this'll be an invalid dereference.

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,404 @@
...
+#include "common/make_unique.h"
+
+#include "core/arm/decoder/decoder.h"
+
+namespace ArmDecoder {
+
+namespace Impl {
+    template<typename Function, size_t i, typename Container, typename... Args>
+    struct Call {
+        FORCE_INLINE static void call(Visitor* v, Function fn, const Container& list, Args&&... args) {
```
lioncash: Were these force inlines profiled?

MerryMage: No, just wanted to ensure the compiler didn't needlessly nest N function calls.

http://goo.gl/tt1sdF (godbolt link) GCC on -O3 actually optimizes this really well.

src/core/arm/decoder/arm.cpp
MerryMage: perhaps an `boost::optional<const Instruction&>`. (boost::optional supports references)

src/core/arm/jit_x64/reg_alloc.h
MerryMage: Hmm. Perhaps there should be a constructor and it should call `Init` with a `nullptr` argument.

src/core/arm/decoder/arm.cpp
lioncash: An iterator effectively acts like that though. You just check if it's the end of the table, then dereference if it isn't at the usage-site.

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,402 @@
...
+    Gen::OpArg MJitStateHostReturnRIP();
+    Gen::OpArg MJitStateHostReturnRSP();
+    Gen::OpArg MJitStateZFlag();
+    Gen::OpArg MJitStateCFlag();
+    Gen::OpArg MJitStateNFlag();
+    Gen::OpArg MJitStateVFlag();
+    Gen::OpArg MJitStateExclusiveTag();
+    Gen::OpArg MJitStateExclusiveState();
+
+    FORCE_INLINE u32 GetReg15Value() { return (current.arm_pc & ~0x1) + GetInstSize() * 2; }
```
lioncash: This can also be const qualified

src/tests/core/arm/jit_x64/fuzz_arm_data_processing.cpp
```diff
@@ -0,0 +1,201 @@
...
+    // Prepare random numbers
+    std::random_device rd;
+    std::mt19937 mt(rd());
+    auto rand_int = [&mt](u32 min, u32 max) -> u32 {
+        std::uniform_int<u32> rand(min, max);
+        return rand(mt);
+    };
+
+    // Prepare memory
+    u8* test_mem = new u8[4096 * 2];
```
MerryMage: `std::vector`

MerryMage:
Opinions on submoduling Catch instead of including it?

src/core/arm/decoder/arm.cpp
MerryMage: Is that possible even though `arm_instruction_table` is `static` to this file in this case?

lioncash: You would need to provide a second function in that case. I guess use an optional if that works.

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,404 @@
...
+    };
+
+    template<size_t N, typename Function>
+    struct MatcherImpl : Matcher {
+        std::array<u32, N> masks;
+        std::array<size_t, N> shifts;
+        Function fn;
+        virtual void visit(Visitor *v, u32 inst) override {
+            std::array<u32, N> values;
+            for (int i = 0; i < N; i++) {
```
lioncash: `i` should be a size_t, since `N` is as well.

MerryMage:
Pushed improved documentation for `JitX64::RegAlloc`.

src/tests/core/arm/jit_x64/fuzz_arm_data_processing.cpp
```diff
@@ -0,0 +1,251 @@
...
+            break;
+        default:
+            // Do nothing
+            break;
+        }
+    }
+    return { bits, mask };
+}
+
+u32 RandInt(u32 min, u32 max) {
```
lioncash: this can likely be:

```cpp
template <typename T>
T RandInt(T min, T max) {
    static_cast(std::is_same<T, signed char>() || std::is_same<T, unsigned char>(),
                "Using char with uniform_int_distribution is undefined behavior");

    static std::random_device rd;
    static std::mt19937 mt(rd());
    std::uniform_int_distribution<T> rand(min, max);
    return rand(mt);
}
```

since you pass in size_t at some places

src/tests/core/arm/jit_x64/fuzz_arm_data_processing.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+void FuzzJit(const int instruction_count, const int run_count, const std::function<u32()> instruction_generator) {
+    // Init core
+    Core::Init();
+    SCOPE_EXIT({ Core::Shutdown(); });
+
+    // Prepare memory
+    constexpr size_t MEMORY_SIZE = 4096 * 2;
+    std::array<u8, MEMORY_SIZE> test_mem;
+    std::memset(test_mem.data(), 0, MEMORY_SIZE);
```
lioncash: this can just be:

```cpp
std::array<u8, MEMORY_SIZE> test_mem{};
```

allowing removal of the memcpy.

MerryMage:
Just woke up and am thinking of renaming the functions in `RegAlloc` to make it obvious when to use them:

Functions that just lock the ARM registers, they may either be in memory (in ARMul_State) or loaded in a x64 register:
* `Gen::OpArg LockReadArm(ArmReg)`: We're going to only read from this register.
* `Gen::OpArg LockReadWriteArm(ArmReg)`: We're going to read and/or write to this register.
* `Gen::OpArg LockWriteArm(ArmReg)`: We're going to write to this register only (i..e: we do not need to the read value, previous value is discarded).
After use of the register, `void UnlockArm(ArmReg)` must be called.

Functions that ensure the ARM register is in an x64 register (i.e.: bound to an x64 register):
* `Gen::X64Reg BindReadArm(ArmReg)`
* `Gen::X64Reg BindReadWriteArm(ArmReg)`
* `Gen::X64Reg BindWriteArm(ArmReg)`
After use of the register, `void UnlockArm(ArmReg)` must be called.

Then I'll also get rid of the `ArmR`, `DirtyArm` and other assistance functions.

MerryMage:
Pushed an implementation for branch instructions.

src/core/arm/disassembler/arm_disasm.cpp
```diff
@@ -513,6 +514,21 @@ std::string ARM_Disasm::DisassembleBX(u32 insn)
     return Common::StringFromFormat("bx%s\tr%d", cond_to_str(cond), rn);
 }
 
+std::string ARM_Disasm::DisassembleBLX(u32 insn)
+{
+    if ((insn & 0xFE000000) == 0xFA000000) {
+        u32 imm24 = insn & 0xFFFFFF;
+        u32 H = (insn >> 24) & 1;
+        s32 offset = MathUtil::SignExtend<26, s32>(imm24 << 2 + H << 1);
```
MerryMage: Mistake in implementation. Missing a `+ 4` here.

MerryMage:
Ported most of the instructions over. Missing some (multiplies, parallel and saturated operations) but I'm happy. Leaving it as-is as I'll be really busy for a few weeks, though I'd be around to do corrections based on reviews.

If one wished to implement instructions in the meantime I wouldn't mind. (VFP is a big thing that's missing.)

Tested and works quite well for non-floating point heavy code.

src/core/arm/decoder/decoder.h
```diff
@@ -0,0 +1,332 @@
...
+class Instruction;
+class Visitor;
+
+boost::optional<const Instruction&> DecodeArm(u32 instruction);
+boost::optional<const Instruction&> DecodeThumb(u16 instruction);
+
+struct Matcher {
+    u32 bit_mask;
+    u32 expected;
+    FORCE_INLINE bool Match(u32 x) const {
```
lioncash: I don't think force inlining these are particularly necessary. (same with the funcs in Instruction)

src/core/arm/decoder/thumb.cpp
```diff
@@ -0,0 +1,455 @@
...
+#include "common/assert.h"
+#include "common/common_types.h"
+#include "common/make_unique.h"
+
+#include "core/arm/decoder/decoder.h"
+
+namespace ArmDecoder {
+
+namespace Impl {
+    struct MatcherImpl : Matcher {
```
lioncash: should this be final?

src/core/arm/decoder/thumb.cpp
```diff
@@ -0,0 +1,455 @@
...
+        MatcherImpl(u32 mask, u32 expect, std::function<void(Visitor* v, u32 instruction)> fn) : fn(fn) {
+            bit_mask = mask;
+            expected = expect;
+        }
+        std::function<void(Visitor* v, u32 instruction)> fn;
```
lioncash: Does it make sense for this to be in the public interface when `visit()` exists?

src/core/arm/decoder/thumb.cpp
```diff
@@ -0,0 +1,455 @@
...
+            expect |= 1;
+            break;
+        default:
+            mask |= 0;
+            expect |= 0;
+            break;
+        }
+    }
+
+    return Common::make_unique<Impl::MatcherImpl>(mask, expect, fn);
```
lioncash: There's likely not that much of a reason to allocate this on the heap, considering the matcher is only a few bytes in size.

src/core/arm/decoder/thumb.cpp
```diff
@@ -0,0 +1,455 @@
...
+}
+
+template<size_t a, size_t b, typename T>
+static constexpr T bits(T s){
+    return ((s << ((sizeof(s) * 8 - 1) - b)) >> (sizeof(s) * 8 - b + a - 1));
```
lioncash: this should have more descriptive template paramters (such as `begin` and `end`). static_asserting if begin is larger than end may also be a good idea.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3856,6 +3860,8 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     LOAD_NZCVT;
     DISPATCH:
     {
+        if (num_instrs >= cpu->NumInstrsToExecute) goto END;
```
lioncash: the goto should likely go on the next line to stay consistent with the other one below.

src/core/arm/jit_x64/common.h
```diff
@@ -0,0 +1,40 @@
...
+    void Reset() {
+        cpu_state.Reset();
+    }
+
+    ARMul_State cpu_state;
+
+    u64 save_host_RSP;
+    u64 return_RIP;
+
+    s32 cycles_remaining;
```
lioncash: These primitives should likely have initial default values.

src/core/arm/jit_x64/common.h
```diff
@@ -0,0 +1,40 @@
...
+using ArmImm5 = ArmDecoder::Imm5;
+using ArmImm8 = ArmDecoder::Imm8;
+using ArmImm11 = ArmDecoder::Imm11;
+using ArmImm12 = ArmDecoder::Imm12;
+using ArmImm24 = ArmDecoder::Imm24;
+using Cond = ArmDecoder::Cond;
+using ShiftType = ArmDecoder::ShiftType;
+using SignExtendRotation = ArmDecoder::SignExtendRotation;
+
+struct JitState {
```
lioncash: likely should be `final`

src/core/arm/jit_x64/cond.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+void JitX64::CondManager::Always() {
+    CompileCond(ConditionCode::AL);
+}
+
+void JitX64::CondManager::FlagsDirty() {
+    flags_dirty = true;
+}
+
+ConditionCode JitX64::CondManager::CurrentCond() {
```
lioncash: This can be a const function

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,398 @@
...
+static std::unique_ptr<Matcher> MakeMatcher(const char format[32], Function fn) {
+    auto ret = Common::make_unique<Impl::MatcherImpl<NumArgs, Function>>();
+    ret->fn = fn;
+    ret->masks.fill(0);
+    ret->shifts.fill(0);
+
+    char ch = 0;
+    int arg = -1;
+
+    for (int i = 0; i < 32; i++) {
```
lioncash: `i` should likely be a u32.

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,398 @@
...
+            for (size_t i = 0; i < NumArgs; i++) {
+                values[i] = (inst & masks[i]) >> shifts[i];
+            }
+            call<NumArgs>(v, fn, values);
+        }
+    };
+}
+
+template<size_t NumArgs, typename Function>
+static std::unique_ptr<Matcher> MakeMatcher(const char format[32], Function fn) {
```
lioncash: format as an array isn't really enforced here (also your string arguments in the table are technically off-by-one since the null-terminator is present, so everything passed actually has a size of 33.

You can technically do:
```cpp
template<size_t NumArgs, typename Function, size_t N>
static std::unique_ptr<Matcher> MakeMatcher(const char format[N], Function fn)
{
    // Ensure proper representation of the ARM instruction's 32 bits, with null terminator 
    static_assert(N == 33, "Format strings must be 33 characters in size.");
    ...
}
```

src/core/arm/decoder/decoder.h
```diff
@@ -0,0 +1,332 @@
...
+};
+
+using Cond = u8;
+using Imm4 = u32;
+using Imm5 = u32;
+using Imm8 = u32;
+using Imm11 = u32;
+using Imm12 = u32;
+using Imm24 = u32;
+using Register = int;
```
lioncash: representing registers as a signed int doesn't particularly make sense (and will likely cause signedness warnings)

src/core/arm/decoder/decoder.h
```diff
@@ -0,0 +1,332 @@
...
+using RegisterList = u16;
+using ShiftType = int;
```
lioncash: This is better represented as an enum, since shift types only have 4 valid representations.

src/core/memory.cpp
```diff
@@ -104,6 +104,13 @@ void UnmapRegion(VAddr base, u32 size) {
     ASSERT_MSG((size & PAGE_MASK) == 0, "non-page aligned size: %08X", size);
     ASSERT_MSG((base & PAGE_MASK) == 0, "non-page aligned base: %08X", base);
     MapPages(base / PAGE_SIZE, size / PAGE_SIZE, nullptr, PageType::Unmapped);
+
+    for (auto iter = current_page_table->special_regions.begin(); iter != current_page_table->special_regions.end(); iter++) {
```
lioncash: this can just be a std::find_if

src/core/arm/jit_x64/reg_alloc.h
```diff
@@ -0,0 +1,189 @@
...
+#include "core/arm/jit_x64/common.h"
+
+namespace JitX64 {
+
+// This is the register allocator
+// TODO: Better algorithm required, ideally something that does reigster usage lookahead.
+//       (Designed so it should be simple to implement later.)
+
+class RegAlloc final {
+private:
```
lioncash: You should likely put the public interface first if possible. That's what people are going to look at most of the time.

src/core/arm/jit_x64/reg_alloc.cpp
```diff
@@ -0,0 +1,377 @@
...
+        X64State& x64_state = x64_gpr[i.second];
+        ASSERT(!x64_state.locked);
+    }
+}
+
+Gen::X64Reg RegAlloc::AllocReg() {
+    // TODO: Improve with an actual register allocator as this is terrible.
+
+    // First check to see if there anything free.
+    for (auto i : x64_reg_to_index) {
```
lioncash: This can likely be its own function `FindFreeRegister` or something and can consist of a single find_if call, since you can return the iterator.

src/core/arm/jit_x64/reg_alloc.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+        ASSERT(x64_state.state != X64State::State::Temp); // This can never happen.
+
+        if (x64_state.state == X64State::State::Free) {
+            return i.first;
+        }
+    }
+
+    // Otherwise flush something.
+    for (auto i : x64_reg_to_index) {
```
lioncash: This can also be a std::find_if

src/core/arm/jit_x64/reg_alloc.cpp
```diff
@@ -0,0 +1,377 @@
...
+        ArmState& arm_state = arm_gpr[arm_reg];
+        ASSERT(!arm_state.locked);
+        if (arm_state.location.IsSimpleReg()) {
+            X64State& x64_state = x64_gpr[x64_reg_to_index.at(arm_state.location.GetSimpleReg())];
+            ASSERT(x64_state.state == X64State::State::CleanArmReg || x64_state.state == X64State::State::DirtyArmReg);
+            ASSERT(x64_state.arm_reg == arm_reg);
+        }
+    }
+
+    for (auto i : x64_reg_to_index) {
```
lioncash: std::all_of

src/core/arm/jit_x64/reg_alloc.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    ASSERT(x64_state.locked);
+    ASSERT(x64_state.state == X64State::State::Temp);
+
+    x64_state.locked = false;
+    x64_state.state = X64State::State::Free;
+}
+
+void RegAlloc::AssertNoLocked() {
+    for (ArmReg arm_reg = 0; arm_reg < arm_gpr.size(); arm_reg++) {
```
lioncash: std::all_of

src/tests/core/arm/jit_x64/fuzz_arm_common.cpp
```diff
@@ -0,0 +1,202 @@
...
+            initial_regs[i] = val;
+        }
+
+        interp.SetCPSR(0x000001d0);
+        jit.SetCPSR(0x000001d0);
+
+        interp.SetPC(0);
+        jit.SetPC(0);
+
+        for (int i = 0; i < instruction_count; i++) {
```
lioncash:
```cpp
std::generate_n(test_mem->code_mem.begin(), instruction_count, instruction_generator);
```

src/tests/core/arm/jit_x64/fuzz_arm_common.cpp
```diff
@@ -0,0 +1,202 @@
...
+        jit.ExecuteInstructions(instructions_to_execute_count);
+        auto jit_mem_recording = test_mem->recording;
+
+        bool pass = true;
+
+        if (interp.GetCPSR() != jit.GetCPSR()) pass = false;
+        for (int i = 0; i <= 15; i++) {
+            if (interp.GetReg(i) != jit.GetReg(i)) pass = false;
+        }
+        if (interp_mem_recording != jit_mem_recording) pass = false;
```
lioncash: This pass determining code can be its own function.

src/core/arm/jit_x64/reg_alloc.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    return x64_reg;
+}
+
+Gen::X64Reg RegAlloc::BindArmForRead(ArmReg arm_reg) {
+    ASSERT(arm_reg >= 0 && arm_reg <= 14); // Not valid for R15 (cannot read from it)
+
+    const Gen::X64Reg x64_reg = BindArmToX64(arm_reg, true);
+
+    return x64_reg;
```
lioncash: the return can just be merged with the function call.

src/core/arm/jit_x64/reg_alloc.cpp
```diff
@@ -0,0 +1,377 @@
...
+    const Gen::X64Reg x64_reg = arm_state.location.GetSimpleReg();
+    const X64State& x64_state = x64_gpr[x64_reg_to_index.at(x64_reg)];
+
+    ASSERT(x64_state.state == X64State::State::CleanArmReg || x64_state.state == X64State::State::DirtyArmReg);
+    ASSERT(x64_state.arm_reg == arm_reg);
+
+    return x64_reg;
+}
+
+bool RegAlloc::IsBoundToX64(ArmReg arm_reg) {
```
lioncash: this ca be a const qualified function.

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,432 @@
...
+    void SETEND(bool E) override;
+    void SRS() override;
+
+    // Thumb specific instructions
+    void thumb_B(Cond cond, ArmImm8 imm8) override;
+    void thumb_B(ArmImm11 imm11) override;
+
+    ArmImm11 thumb_BLX_prefix_imm11 = 0;
+    bool thumb_BLX_prefix_executed = false;
+    bool thumb_BLX_suffix_executed = false;
```
lioncash: These var declarations should probably be under the functions  so all the functions are in order.

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,432 @@
...
+    bool EFlag; ///< Big / Little Endian
+
+    bool operator == (const LocationDescriptor& o) const {
+        return std::tie(arm_pc, TFlag, EFlag) == std::tie(o.arm_pc, o.TFlag, o.EFlag);
+    }
+};
+
+struct LocationDescriptorHash {
+    size_t operator()(const LocationDescriptor& x) const {
+        return std::hash<u64>()((u64)x.arm_pc ^ ((u64)x.TFlag << 32) ^ ((u64)x.EFlag << 33));
```
lioncash: static_cast

src/core/arm/jit_x64/jit_x64.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+void JitX64::CompileCallHost(const void* const fn) {
+    // There is no need to setup the stack as the stored RSP has already been properly aligned.
+
+    reg_alloc.FlushABICallerSaved();
+
+    ASSERT(reg_alloc.JitStateReg() != RSP);
+    code->MOV(64, R(RSP), MJitStateHostReturnRSP());
+
+    const u64 distance = reinterpret_cast<const u64>(fn) - (reinterpret_cast<const u64>(code->GetCodePtr()) + 5);
```
lioncash: the const within the reinterpret_casts aren't needed.

src/core/arm/decoder/thumb.cpp
MerryMage: Is it possible to have polymorphism without heap allocation in C++?

src/core/arm/decoder/thumb.cpp
MerryMage: Actually, should something like this be moved to common?

src/core/arm/decoder/arm.cpp
MerryMage: Ah, I had that in not really as a size check, but as a misremembered C99 optimization (should be `const char format[static 32]` which tells the compiler `format` is at least 32 elements long).

Having a compile-time size check is probably a better idea.

src/core/arm/decoder/decoder.h
MerryMage: Should probably be an `enum class`

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,398 @@
...
+#include "common/assert.h"
+#include "common/common_types.h"
+#include "common/make_unique.h"
+
+#include "core/arm/decoder/decoder.h"
+
+namespace ArmDecoder {
+
+namespace Impl {
+    // std::integer_sequence and std::make_integer_sequence are only available in C++14
```
lioncash: Now the std lib can just be used, since citra compiles as C++14 now.

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,398 @@
...
+        (v->*fn)(std::get<seq>(list)...);
+    }
+
+    /// Function has NumArgs arguments
+    template<size_t NumArgs, typename Function>
+    struct MatcherImpl : Matcher {
+        std::array<u32, NumArgs> masks = {};
+        std::array<size_t, NumArgs> shifts = {};
+        Function fn = nullptr;
+        virtual void visit(Visitor *v, u32 inst) override {
```
lioncash: `virtual` isn't needed here, since override implies it

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,398 @@
...
+            std::array<u32, NumArgs> values;
+            for (size_t i = 0; i < NumArgs; i++) {
```
lioncash: This can be a transform

```cpp
std::transform(masks.begin(), masks.begin() + NumArgs, shifts.begin(), values.begin(), [inst](u32 mask, size_t shift) {
    return (inst & mask) >> shift;
});
```

src/core/arm/decoder/thumb.cpp
MerryMage: Probably going to separate thumb and ARM instructions so they don't share the same `Visitor` interface; there's no reason for them to anyway.

src/core/arm/jit_x64/instructions/branch.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include "common/math_util.h"
+
+#include "core/arm/jit_x64/jit_x64.h"
+
+namespace JitX64 {
+
+using namespace Gen;
+
+void JitX64::B(Cond cond, ArmImm24 imm24) {
+    cond_manager.CompileCond((ConditionCode)cond);
```
lioncash: static_cast

src/core/arm/jit_x64/jit_x64.h
```diff
@@ -0,0 +1,432 @@
...
+
+    CodePtr GetBB(u32 pc, bool TFlag, bool EFlag);
+
+    /// Returns a pointer to the compiled basic block.
+    CodePtr Compile(u32 pc, bool TFlag, bool EFlag);
+
+private:
+    LocationDescriptor current;
+    unsigned instructions_compiled;
+    bool stop_compilation;
```
lioncash: These primitives should likely have default values.

src/core/arm/decoder/thumb.cpp
lioncash: Probably, but within its own header like bit_utils.h or something.

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,398 @@
...
+            ch = 0;
+            continue;
+        }
+
+        // Ban some characters
+        ASSERT(format[i] != 'I');
+        ASSERT(format[i] != 'l');
+        ASSERT(format[i] != 'O');
+
+        if (format[i] != ch){
```
lioncash: space before `{`

src/core/arm/decoder/decoder.h
```diff
@@ -0,0 +1,332 @@
...
+private:
+    const std::unique_ptr<Matcher> matcher;
+
+public:
+    Instruction(const char* const name, std::unique_ptr<Matcher> matcher) : matcher(std::move(matcher)), name(name) {}
+
+    const char* const name;
+
+    FORCE_INLINE bool Match(u32 instruction) const {
+        return (instruction & matcher->bit_mask) == matcher->expected;
```
lioncash: could this not be `return matcher->Match(instruction)`?

src/core/arm/decoder/decoder.h
```diff
@@ -0,0 +1,332 @@
...
+};
+
+class Instruction {
+private:
+    const std::unique_ptr<Matcher> matcher;
+
+public:
+    Instruction(const char* const name, std::unique_ptr<Matcher> matcher) : matcher(std::move(matcher)), name(name) {}
+
+    const char* const name;
```
lioncash: Should this be directly accessible?

MerryMage:
Addressed @lioncash's comments above. Going to rebase as I now use C++14 features.

src/core/arm/decoder/decoder.h
```diff
@@ -0,0 +1,417 @@
...
+
+enum class Register {
+    R0, R1, R2, R3, R4, R5, R6, R7, R8, R9, R10, R11, R12, R13, R14, R15,
+    SP = R13,
+    LR = R14,
+    PC = R15,
+    INVALID_REG = 99
+};
+
+static Register operator+ (Register arm_reg, int number) {
```
lioncash: this should likely be `inline` instead of `static`, since it's in the header

KitatusStudios: To my knowledge, there's nothing wrong with using static in a header but I agree that it's better to be safe rather than sorry.

lioncash: > To my knowledge, there's nothing wrong with using static in a header

static causes every translation unit that uses the function to have a copy of it, which is unnecessary.

MerryMage: Note to future self: `src/core/arm/jit_x64/common.h` is another instance of `static` in a header

src/core/arm/jit_x64/interface.cpp
```diff
@@ -0,0 +1,238 @@
...
+        auto cpu = &jit_state->cpu_state;
+
+        cpu->NFlag = (cpu->Cpsr >> 31);
+        cpu->ZFlag = (cpu->Cpsr >> 30) & 1;
+        cpu->CFlag = (cpu->Cpsr >> 29) & 1;
+        cpu->VFlag = (cpu->Cpsr >> 28) & 1;
+        cpu->TFlag = (cpu->Cpsr >> 5) & 1;
+
+        jit_state->cycles_remaining = cycles_to_run;
+        jit_state->return_RIP = (u64)CallCodeReturnAddress();
```
lioncash: reinterpret_cast

src/core/arm/decoder/arm.cpp
```diff
@@ -0,0 +1,392 @@
...
+            call<NumArgs>(v, fn, values);
+        }
+    };
+}
+
+template<size_t NumArgs, typename Function>
+static std::unique_ptr<ArmMatcher> MakeMatcher(const char* const format, Function fn) {
+    ASSERT(strlen(format) == 32);
+
+    auto ret = Common::make_unique<Impl::MatcherImpl<NumArgs, Function>>();
```
MerryMage: `std::make_unique`

src/core/arm/jit_x64/interface.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+namespace JitX64 {
+
+struct ARM_Jit::Impl {
+    Gen::RunJittedCode run_jit = {};
+    Gen::BlockOfCode block_of_code = {};
+    JitX64 compiler{ &block_of_code };
+};
+
+ARM_Jit::ARM_Jit(PrivilegeMode initial_mode) : impl(std::make_unique<Impl>()), state(Common::make_unique<JitState>()) {
```
MerryMage: `std::make_unique`

src/core/arm/jit_x64/instructions/load_store.cpp
```diff
@@ -0,0 +1,1034 @@
...
+    if (W)
+        ASSERT_MSG(Rn_index != Rd_index, "UNPREDICTABLE");
+
+    LoadStoreCommon_AddrMode2(this, reg_alloc, P, W,
+        &JitX64::LoadAndStoreWordOrUnsignedByte_ImmediateOffset,
+        &JitX64::LoadAndStoreWordOrUnsignedByte_ImmediatePreIndexed,
+        &JitX64::LoadAndStoreWordOrUnsignedByte_ImmediatePostIndexed,
+        U, Rn_index, imm12);
+
+    CompileCallHost(reinterpret_cast<const void* const>(!current.EFlag ? &Load32LE : &Load32BE));
```
lioncash: the second const after the `void*` isn't necessary, ditto for other similar occurrences.

src/common/bit_util.h
```diff
@@ -0,0 +1,32 @@
...
+// Refer to the license.txt file included.
+
+#include <cstddef>
+#include <type_traits>
+
+namespace BitUtil {
+
+/// Extract bits [begin_bit, end_bit] inclusive from s of type T.
+template<size_t begin_bit, size_t end_bit, typename T>
+constexpr T Bits(T s) {
```
lioncash: `s` -> `value` (or `val`)

src/common/bit_util.h
```diff
@@ -0,0 +1,32 @@
...
+    static_assert(begin_bit <= end_bit, "bit range must begin before it ends");
+    static_assert(begin_bit < sizeof(s) * 8, "begin_bit must be smaller than size of T");
```
lioncash: you can probably just define

```cpp
template <typename T>
constexpr size_t BitSize() {
    return sizeof(T) * CHAR_BIT;
}
```
to get rid of this common pattern

src/common/bit_util.h
```diff
@@ -0,0 +1,32 @@
...
+    static_assert(end_bit < sizeof(s) * 8, "begin_bit must be smaller than size of T");
+
+    return (s >> begin_bit) & ((1 << (end_bit - begin_bit + 1)) - 1);
+}
+
+/// Sign-extends a val of type T that has NBits bits.
+template<size_t NBits, typename T>
+inline T SignExtend(const T val) {
+    static_assert(std::is_signed<T>(), "T must be signed");
```
lioncash: This requirement will actually cause warnings when working with unsigned types.

MerryMage: Ah, I'll take a look at yours.

src/core/arm/decoder/decoder.h
lioncash: similarly in `src/core/arm/jit_x64/instructions/helper/load_store.h` (these should likely have their implementation in a .cpp file to avoid pulling in the memory header into anything)

MerryMage:
Addressed comments above + a few other things. Apologies for another rebase to master.

src/common/bit_util.h
```diff
@@ -0,0 +1,45 @@
...
+    return (value >> begin_bit) & ((1 << (end_bit - begin_bit + 1)) - 1);
+}
+
+/// Extracts a single bit at bit_position from val of type T.
+template<size_t bit_position, typename T>
+constexpr T Bit(const T value) {
+    return Bits<bit_position, bit_position, T>(value);
+}
+
+/// Sign-extends a val that has NBits bits to the full bitwidth of type T.
```
lioncash: value

src/common/bit_util.h
```diff
@@ -0,0 +1,45 @@
...
+template<size_t NBits, typename T>
+inline T SignExtend(const T value) {
+    static_assert(NBits <= BitSize<T>(), "NBits larger that bitsize of T");
```
lioncash: than

src/common/bit_util.h
```diff
@@ -0,0 +1,45 @@
...
+    static_assert(begin_bit < BitSize<T>(), "begin_bit must be smaller than size of T");
+    static_assert(end_bit < BitSize<T>(), "begin_bit must be smaller than size of T");
+
+    return (value >> begin_bit) & ((1 << (end_bit - begin_bit + 1)) - 1);
+}
+
+/// Extracts a single bit at bit_position from val of type T.
+template<size_t bit_position, typename T>
+constexpr T Bit(const T value) {
+    return Bits<bit_position, bit_position, T>(value);
```
lioncash: This isn't something you typically want to allow with `Bits` considering it *is* plural, yet it's being utilized to only extract a single bit. This allows for off-by-one errors.

src/common/bit_util.h
```diff
@@ -0,0 +1,45 @@
...
+/// The size of a type in terms of bits
+template <typename T>
+constexpr size_t BitSize() {
+    return sizeof(T) * CHAR_BIT;
+}
+
+/// Extract bits [begin_bit, end_bit] inclusive from val of type T.
+template<size_t begin_bit, size_t end_bit, typename T>
+constexpr T Bits(const T value) {
+    static_assert(begin_bit <= end_bit, "bit range must begin before it ends");
```
lioncash: This is kind of an ambiguous error message, something like `beginning bit specified cannot be greater than the ending bit` or something more along those lines would likely be better.

src/common/bit_util.h
MerryMage: Hm, that makes sense.

MerryMage:
Addressed comments above about BitUtil.

MerryMage:
Closing this for now. Will likely break this down into smaller PRs or something else when it's ready.

archshift:
@MerryMage, any news?

kreedz1:
@MerryMage Any updates on this ?

PR #1561, opened 2016-03-21
hle: Get rid of direct global access to g_reschedule

lioncash:
This shouldn't be directly exposed if there's already a partial API that operates on it.
We can just provide the rest of said API.

### DIFF
--- a/src/core/hle/hle.cpp
+++ b/src/core/hle/hle.cpp
@@ -14,9 +14,13 @@
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-namespace HLE {
+namespace {
 
-bool g_reschedule; ///< If true, immediately reschedules the CPU to a new thread
+bool reschedule; ///< If true, immediately reschedules the CPU to a new thread
+
+}
+
+namespace HLE {
 
 void Reschedule(const char *reason) {
     DEBUG_ASSERT_MSG(reason != nullptr && strlen(reason) < 256, "Reschedule: Invalid or too long reason.");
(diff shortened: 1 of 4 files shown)


Subv:
LGTM

src/core/hle/hle.cpp
```diff
@@ -29,13 +33,21 @@ void Reschedule(const char *reason) {
 
     Core::g_app_core->PrepareReschedule();
 
-    g_reschedule = true;
+    reschedule = true;
+}
+
+bool RescheduleIsPending() {
```
yuriks: "IsReschedulePending" is the style I've always seen everyone else use for APIs returning bool.

src/core/hle/kernel/thread.cpp
```diff
@@ -483,7 +483,8 @@ void Reschedule() {
 
     Thread* cur = GetCurrentThread();
     Thread* next = PopNextReadyThread();
-    HLE::g_reschedule = false;
+
+    HLE::DoneRescheduling();
```
yuriks: This feels weird. Why doesn't "Reschedule" have access to its internals? I think the variable maybe should be moved to thread.cpp

bunnei:
@lioncash do you plan on addressing the above comments? Seems pretty trivial and it'd be nice to get this one merged. Thanks.

lioncash:
I will once I take care of hardware testing #1568 and take care of issues with the pica config.

bunnei:
@lioncash Ping again to wrap this up... Sorry to nag, but we've been getting an absurd amount of PRs lately and this seems like really low hanging fruit to get merged!

lioncash:
@yuriks Where should I place the contents of HLE::Reschedule if I move the global into thread.cpp?

yuriks:
@lioncash nevermind that, after looking at it again there's no real good solution for where to move it. (Making random things call kernel thread functions also sounds bad.)

bunnei:
Superceded by #1762

PR #1562, opened 2016-03-21
core: Mark some helper functions as constexpr

lioncash:
Left the respective commits separate for review. Will squash them if preferred.


bunnei:
LGTM

PR #1563, opened 2016-03-21
armstate: Correct FIQ register banking

lioncash:
FIQ mode has seven banked registers (R8 to R14), not two.

### DIFF
--- a/src/core/arm/skyeye_common/armstate.cpp
+++ b/src/core/arm/skyeye_common/armstate.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
 #include "common/swap.h"
 #include "common/logging/log.h"
 #include "core/memory.h"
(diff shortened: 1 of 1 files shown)


lioncash:
Goddamnit, I just realized I totally blew the one chance I had to make my branch name arm-fanfiqtion

src/core/arm/skyeye_common/armstate.cpp
```diff
@@ -48,8 +49,7 @@ void ARMul_State::ChangePrivilegeMode(u32 new_mode)
             Spsr[UNDEFBANK] = Spsr_copy;
             break;
         case FIQ32MODE:
-            Reg_firq[0] = Reg[13];
-            Reg_firq[1] = Reg[14];
+            std::copy(Reg.begin() + 8, Reg.end(), Reg_firq.begin());
```
MerryMage: `Reg` is `std::array<u32, 16>`, so this would copy 8 elements but `Reg_firq` has 7 elements.

lioncash: Note to self to not write code 5 minutes before leaving for work.

lioncash:
amended

MerryMage:
LGTM

bunnei:
LGTM

PR #1564, opened 2016-03-21
tests: Infrastructure for unit tests

MerryMage:
Extracted the test infrastructure commit from #1554.

Does not execute tests on the buildbots yet. Should this be done?


Subv:
>Does not execute tests on the buildbots yet. Should this be done?
IMO yes

appveyor.yml
```diff
@@ -22,6 +22,10 @@ before_build:
   - cmake -G "Visual Studio 14 2015 Win64" -DCITRA_USE_BUNDLED_QT=1 -DCITRA_USE_BUNDLED_SDL2=1 ..
   - cd ..
 
+build:
+  project: build/citra.sln
+  parallel: true
+
```
JayFoxRox: How is this related to testing?

MerryMage: Appveyor fails to properly autodetect the correct `.sln` file as the catch submodule itself has a few `.sln` files, hence it has to be explicitly specified.

JayFoxRox:
LGTM so far BUT I'm afraid nobody will use it if there is not a single test defined.
So I'd prefer if there was an issue or even a PR (based on this one) which already implements some tests which could serve as example. I might even start adding unit tests myself then.

MerryMage:
> a PR (based on this one) which already implements some tests which could serve as example

#1554 has a number of tests (for example [this one](https://github.com/MerryMage/citra/blob/jit2/src/tests/core/arm/jit_x64/fuzz_arm_branch.cpp)), although those aren't exactly your typical unit tests. I would suggest reading the [Catch tutorial](https://github.com/philsquared/Catch/blob/master/docs/tutorial.md) before use. I would also love to write some audio tests, I've actually been itching to.

A suggsted folder structure for the `src/tests/` directory is to reflect the project structure (for example, a test for `core/arm/jit_x64/instructions/branch.cpp` might appear in `src/tests/core/arm/jit_x64/fuzz_arm_branch.cpp`).

Subv:
Could you rebase this so we can merge it once and for all? thanks!

MerryMage:
rebased

JayFoxRox:
Again please?

MerryMage:
@JayFoxRox rebased.

MerryMage:
Tests are now run on CI.

CMakeLists.txt
```diff
@@ -255,6 +255,8 @@ endif()
 
 add_subdirectory(externals/soundtouch)
 
+enable_testing()
```
MerryMage: This line needs to be in the root `CMakeLists.txt` and be before `add_subdirectory(src)`.

bunnei:
LGTM. hopefully people will write tests for this :p

MerryMage:
\tests/

PR #1568, opened 2016-03-24
Fix ftoi and disable VFPv3

### LINKED ISSUE #1218: Homebrew causes Citra to freeze while trying to print out double values with printf
Not really sure what's going on here but I've been getting freezes seemingly randomly while trying to print out double values with printf and format strings
example code: https://gist.github.com/iamevn/b6163562f03ed9068ac4 ([.3dsx](http://puu.sh/kVcXN/35c6061f40.3dsx))

I haven't managed to reproduce this freeze on the actual hardware so I'm pretty sure it isn't a bug in ctrulib or something like that.

### LINKED ISSUE #1483: Samurai Warriors Chronicles - Broken Character Select Screen
Using current master:

Broken

intended (from actual 3DS):

From 3DS

---

I already analyzed the vertexshader in question. There are no loops and no GS (0x2A / 0x2B) instructions.

v0 = pos
v1 = normal
v4 = uv
v7 = bone index
v8 = bone weight [0;100]

The function at 0x0030 will transform vertex [r15] and normal [r14] using matrices from c25[].
The skinning result is accumulated in r7 [vertex] and r12 [normal].

c8.w is scale for the bone weight to [0;1], r2 is used to calculate matrix indices from bone indices (bug? normal and vertex matrix is the same in the char select screen)

c93 is a { 1.0, 2.0, 3.0, 4.0 } LUT.

The buggy part seems to be the matrix stack at c25[]. The data is uploaded as 32 bit float but seems to contain garbage. I'd have expected mostly rotation values [-1;+1] but instead the data range is huge.

```
GPUREG_VSH_FLOATUNIFORM_INDEX	2c0	1111	80000019	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f1c116ce	(Example: 3.9106684E33)
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	65223bc6	(Example: 4.7882875E22)
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6e276ff8	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7005ae43	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	71324421	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	e4c3ebc2	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	ed9a874a	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	ef9e759c	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f22e32cb	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	65925c8d	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6e970e6a	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	70713451	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	672de673	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6d627847	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	e3ed2075	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6eb418d8	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	e6ab4765	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	ed07abd4	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	63536bbd	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	ee2092cf	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	679ce308	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6dcc5039	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	e455ed80	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6f227a39	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f740cf93	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7415e944	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	73b5df2b	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f2a2f8ef	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	76e88144	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f3b70621	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f35b5056	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7244860f	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f7adf285	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	74873ea8	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	74241417	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f3130733	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f2434009	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	70379025	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6eb82c32	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6da50850	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	71b6d7c9	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	efdedfc0	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	ee2c7827	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	ed1a8b80	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	f2b025e5	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	70a59ab6	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6f262780	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	6e14e2fd	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
GPUREG_VSH_FLOATUNIFORM_DATA0	2c1	1111	7fc00000	
```

I've tried injecting identity matrices but it still gave problems (I'm assuming some other matrices are also broken).
I'm guessing this bug is related to some broken CPU instruction / memory overflow / missing coprocessor or some sync issue.

Worth noting might be the following debug output just before the broken 3D part starts.

```
[ 120.800437] Service <Error> core/hle/service/service.cpp:SyncRequest:68: unknown / unimplemented function 'InvalidateDCache': port=dsp::DSP, cmd_buff[1]=0x521B40, cmd_buff[2]=0x4000, cmd_buff[3]=0x0, cmd_buff[4]=0xFFFF8001
[ 120.800702] Service <Error> core/hle/service/service.cpp:SyncRequest:68: unknown / unimplemented function 'InvalidateDCache': port=dsp::DSP, cmd_buff[1]=0x521B40, cmd_buff[2]=0x4000, cmd_buff[3]=0x0, cmd_buff[4]=0xFFFF8001
[ 120.800938] Service <Error> core/hle/service/service.cpp:SyncRequest:68: unknown / unimplemented function 'InvalidateDCache': port=dsp::DSP, cmd_buff[1]=0x521B40, cmd_buff[2]=0x4000, cmd_buff[3]=0x0, cmd_buff[4]=0xFFFF8001
[ 120.801171] Service <Error> core/hle/service/service.cpp:SyncRequest:68: unknown / unimplemented function 'InvalidateDCache': port=dsp::DSP, cmd_buff[1]=0x521B40, cmd_buff[2]=0x4000, cmd_buff[3]=0x0, cmd_buff[4]=0xFFFF8001
```

[Full Log](https://cdn.rawgit.com/JayFoxRox/45acbeb5fe4626686754/raw/043e481a9abed832af208e6024618899b195f34a/index.html)

(Screenshots by @JohnGodgames)

JayFoxRox:
This fixes the printf bug (which is actually a bug in "dtoa")
The buggy code is:
```C
ds = (d2.d - 1.5) * 0.289529654602168 + 0.1760912590558 + i * 0.301029995663981;
k = (int) ds;
```
Example values were 0x3ff999999999999a for d2.d (double) and int i = -3.
When ds is -0.6, k would currently be 0x80000000 when 0 is expected (suggested by running dtoa on my x64).

Changing the 2 lines from `if (vdm.exponent >= 1023 - 1)` to just 1023 makes printf work fine (I think?) but I didn't really test it much.
You can find [similar code in qemu](https://github.com/qemu/qemu/blob/master/fpu/softfloat.c#L3126) which also compares using `if ( aExp < 0x3FF )` = `if (!(aExp >= 0x3FF))` = `if (!(aExp >= 1023))`.
There might be some shifting of the mantissa which makes this change incorrect so I'd prefer if someone could look over it.
However, the [single-precision code does it also like this PR suggests](https://github.com/citra-emu/citra/blob/master/src/core/arm/skyeye_common/vfp/vfpsingle.cpp#L587) already.

Fixes #1218 

---

Furthermore the interpreter contains some VFP3 instructions which cause #1483 . This is an entirely different problem but it was found by accident while searching for the other bug.
The idea for this change came from @MerryMage who said it was important that some instructions are ambigious. So I figured that VFP3 might "overwrite" some other instructions.
Information that [3DS only supports VFP2 came from @lioncash ](http://lioncash.github.io/ARMBook/vfp/README.html)
Is this correct or is there any reason for VFP3 code to exist?

Fixes #1483 

---

Tiny change but I'd be happy to split this into 2 seperate commits and give appropiate commit names. Suggestions?

### DIFF
--- a/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
+++ b/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
@@ -615,7 +615,7 @@ static u32 vfp_double_ftoui(ARMul_State* state, int sd, int unused, int dm, u32 
 static u32 vfp_double_ftouiz(ARMul_State* state, int sd, int unused, int dm, u32 fpscr)
 {
     LOG_TRACE(Core_ARM11, "In %s", __FUNCTION__);
-    return vfp_double_ftoui(state, sd, unused, dm, FPSCR_ROUND_TOZERO);
+    return vfp_double_ftoui(state, sd, unused, dm, (fpscr & ~FPSCR_RMODE_MASK) | FPSCR_ROUND_TOZERO);
 }
 
 static u32 vfp_double_ftosi(ARMul_State* state, int sd, int unused, int dm, u32 fpscr)
@@ -692,7 +692,7 @@ static u32 vfp_double_ftosi(ARMul_State* state, int sd, int unused, int dm, u32 
 static u32 vfp_double_ftosiz(ARMul_State* state, int dd, int unused, int dm, u32 fpscr)
 {
     LOG_TRACE(Core_ARM11, "In %s", __FUNCTION__);
-    return vfp_double_ftosi(state, dd, unused, dm, FPSCR_ROUND_TOZERO);
+    return vfp_double_ftosi(state, dd, unused, dm, (fpscr & ~FPSCR_RMODE_MASK) | FPSCR_ROUND_TOZERO);
 }
 
 static struct op fops_ext[] = {
--- a/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
+++ b/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
@@ -560,7 +560,7 @@ static u32 vfp_double_ftoui(ARMul_State* state, int sd, int unused, int dm, u32 
     if (vdm.exponent >= 1023 + 32) {
         d = vdm.sign ? 0 : 0xffffffff;
         exceptions = FPSCR_IOC;
-    } else if (vdm.exponent >= 1023 - 1) {
+    } else if (vdm.exponent >= 1023) {
         int shift = 1023 + 63 - vdm.exponent;
         u64 rem, incr = 0;
 
@@ -595,12 +595,20 @@ static u32 vfp_double_ftoui(ARMul_State* state, int sd, int unused, int dm, u32 
     } else {
         d = 0;
         if (vdm.exponent | vdm.significand) {
-            exceptions |= FPSCR_IXC;
-            if (rmode == FPSCR_ROUND_PLUSINF && vdm.sign == 0)
+            if (rmode == FPSCR_ROUND_NEAREST) {
+                if (vdm.exponent >= 1022) {
+                    d = vdm.sign ? 0 : 1;
+                    exceptions |= vdm.sign ? FPSCR_IOC : FPSCR_IXC;
+                } else {
+                    exceptions |= FPSCR_IXC;
+                }
+            } else if (rmode == FPSCR_ROUND_PLUSINF && vdm.sign == 0) {
                 d = 1;
-            else if (rmode == FPSCR_ROUND_MINUSINF && vdm.sign) {
-                d = 0;
-                exceptions |= FPSCR_IOC;
+                exceptions |= FPSCR_IXC;
+            } else if (rmode == FPSCR_ROUND_MINUSINF) {
+                exceptions |= vdm.sign ? FPSCR_IOC : FPSCR_IXC;
+            } else {
+                exceptions |= FPSCR_IXC;
             }
         }
     }
@@ -639,12 +647,12 @@ static u32 vfp_double_ftosi(ARMul_State* state, int sd, int unused, int dm, u32 
     if (tm & VFP_NAN) {
         d = 0;
         exceptions |= FPSCR_IOC;
-    } else if (vdm.exponent >= 1023 + 32) {
+    } else if (vdm.exponent >= 1023 + 31) {
         d = 0x7fffffff;
         if (vdm.sign)
             d = ~d;
         exceptions |= FPSCR_IOC;
-    } else if (vdm.exponent >= 1023 - 1) {
+    } else if (vdm.exponent >= 1023) {
         int shift = 1023 + 63 - vdm.exponent;	/* 58 */
         u64 rem, incr = 0;
 
@@ -675,10 +683,17 @@ static u32 vfp_double_ftosi(ARMul_State* state, int sd, int unused, int dm, u32 
         d = 0;
         if (vdm.exponent | vdm.significand) {
             exceptions |= FPSCR_IXC;
-            if (rmode == FPSCR_ROUND_PLUSINF && vdm.sign == 0)
+            if (rmode == FPSCR_ROUND_NEAREST) {
+                if (vdm.exponent >= 1022) {
+                    d = vdm.sign ? 0xffffffff : 1;
+                } else {
+                    d = 0;
+                }
+            } else if (rmode == FPSCR_ROUND_PLUSINF && vdm.sign == 0) {
                 d = 1;
-            else if (rmode == FPSCR_ROUND_MINUSINF && vdm.sign)
-                d = -1;
+            } else if (rmode == FPSCR_ROUND_MINUSINF && vdm.sign) {
+                d = 0xffffffff;
+            }
         }
     }
 
--- a/src/core/arm/dyncom/arm_dyncom_dec.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_dec.cpp
@@ -422,6 +422,10 @@ ARMDecodeStatus DecodeARMInstruction(u32 instr, s32* idx) {
         n = arm_instruction[i].attribute_value;
         base = 0;
 
+        // 3DS has no VFP3 support
+        if (arm_instruction[i].version == ARMVFP3)
+            continue;
+
         while (n) {
             if (arm_instruction[i].content[base + 1] == 31 && arm_instruction[i].content[base] == 0) {
                 // clrex
--- a/src/core/arm/skyeye_common/vfp/vfpsingle.cpp
+++ b/src/core/arm/skyeye_common/vfp/vfpsingle.cpp
@@ -592,7 +592,11 @@ static u32 vfp_single_ftoui(ARMul_State* state, int sd, int unused, s32 m, u32 f
          * 2^0 <= m < 2^32-2^8
          */
         d = (vsm.significand << 1) >> shift;
-        rem = vsm.significand << (33 - shift);
+        if (shift > 0) {
+            rem = (vsm.significand << 1) << (32 - shift);
+        } else {
+            rem = 0;
+        }
 
         if (rmode == FPSCR_ROUND_NEAREST) {
             incr = 0x80000000;
@@ -619,12 +623,20 @@ static u32 vfp_single_ftoui(ARMul_State* state, int sd, int unused, s32 m, u32 f
     } else {
         d = 0;
         if (vsm.exponent | vsm.significand) {
-            exceptions |= FPSCR_IXC;
-            if (rmode == FPSCR_ROUND_PLUSINF && vsm.sign == 0)
+            if (rmode == FPSCR_ROUND_NEAREST) {
+                if (vsm.exponent >= 126) {
+                    d = vsm.sign ? 0 : 1;
+                    exceptions |= vsm.sign ? FPSCR_IOC : FPSCR_IXC;
+                } else {
+                    exceptions |= FPSCR_IXC;
+                }
+            } else if (rmode == FPSCR_ROUND_PLUSINF && vsm.sign == 0) {
                 d = 1;
-            else if (rmode == FPSCR_ROUND_MINUSINF && vsm.sign) {
-                d = 0;
-                exceptions |= FPSCR_IOC;
+                exceptions |= FPSCR_IXC;
+            } else if (rmode == FPSCR_ROUND_MINUSINF) {
+                exceptions |= vsm.sign ? FPSCR_IOC : FPSCR_IXC;
+            } else {
+                exceptions |= FPSCR_IXC;
             }
         }
     }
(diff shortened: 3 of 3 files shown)


lioncash:
Assigning myself to hardware test duty

JamePeng:
hmm,the ARM11(ARMv6) was not support the VFPv3, it implemented on most Cortex-A8 and A9 ARMv7 processors.
https://en.wikipedia.org/wiki/ARM_architecture#Floating-point_.28VFP.29

src/core/arm/skyeye_common/vfp/vfpdouble.cpp
```diff
@@ -560,7 +560,7 @@ static u32 vfp_double_ftoui(ARMul_State* state, int sd, int unused, int dm, u32
     if (vdm.exponent >= 1023 + 32) {
```
JayFoxRox: Chances are this is supposed to be 1023 + 64 ?

src/core/arm/skyeye_common/vfp/vfpdouble.cpp
```diff
@@ -644,7 +644,7 @@ static u32 vfp_double_ftosi(ARMul_State* state, int sd, int unused, int dm, u32
         if (vdm.sign)
```
JayFoxRox: (Slightly above where GitHub doesn't allow commenting)

Same: maybe should be 1023 + 64

JayFoxRox:
It's almost been a month. Progress update?
I need this for most HW tests (unless I want people to use compilers with softfp etc. or I change my HW tests to all int math).

lioncash:
Will make time for it tonight when I get home. Can you see if this issue still exists (can't do it myself since it's all at home). Curious if #1653 could have fixed it somehow, since it's possible a jump address was getting trashed.

JayFoxRox:
Nope, as I said before already: it's exactly that instruction + vfp3 bug.

The ftoi code looks fishy compared to the other similar code pieces (I linked to above) and that instruction produces absolutly ridiculous results (iirc: ftoi 0.3 ~~> (int)<giant value> instead of the expected 0(?))
(also see the dtoa code snippet above, I have actually followed the disassembly of it and traced it down to a handful of instructions - and the most promising candidate was ftoi where a quick comparison to other code pieces hinted what could be wrong)

The bug in Samurai Warriors (VFP3 issue) is also still present.

lioncash:
I'm somewhat skeptical this is a correct fix particularly because the Linux kernel emulates the soft-float routines with the same code, as seen [here](https://github.com/torvalds/linux/blob/master/arch/arm/vfp/vfpdouble.c#L528) and [here](https://github.com/torvalds/linux/blob/master/arch/arm/vfp/vfpdouble.c#L610). But anyway, I'll still test it when I get home.

Subv:
Reminder that this is still a thing

JayFoxRox:
I've written [a hardware test for this](https://github.com/JayFoxRox/3ds-tests/commit/a424220e82c8a7aa656877fa25b940003a5e73ab) now and could confirm my changes are better than master.
However, there were 2 issues:

1. The 3DS hangs after around 4000 randomly (some heuristics included) chosen doubles on some denormal [if not both, DN and FZ, are set in fpscr (click for doc)](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0360f/CHDFAFIA.html). The mode works fine in Citra and I can't see how this should hang the hardware. I might contact ARM about this because it looks like a hw issue to me.
2. With my fix only 3/4 rounding modes worked. The round-to-nearest mode (which was also broken in master) needed a special fix and I just looked at some numbers to figure out the current solution.

I have updated this PR with a fix for 2. and split the original commit into 2 commits, one for each feature.

I did not check the fpscr bits (inexact bits etc.) after running ftoi-instructions yet but I intend to do so. I also plan to test more numbers. I also hope @lioncash can help with 1. and I also hope he can do some extra tests due to how fishy this function seems to be.

JayFoxRox:
**Update:**

I've added NaN checking and denormal checking to my test now. Also I've run 100k tests per fpscr value (5 values), one run for ftoi32, ftoi64 (2 variants). => 5 * 2 * 100k tests.
I tried to run other modes but the hardware crash prevents me from testing without FZ or DN bit.
For the heuristics of input numbers you can [check this comment](https://github.com/JayFoxRox/3ds-tests/blob/cdb520a2ca0cc7651900ea250b31a931adb106dd/ftoi/source/main.c#L19)
Each test consists of 2 instructions: ftoiu + ftois, the fpscr is set before every instruction and read back afterwards. The output is one line per test, containing the inputer, result of the 2 ftoi and the 2 fpscr after execution.

The following is a comparison of output generated by @wwylele using [this version of the hw-test](https://github.com/JayFoxRox/3ds-tests/blob/cdb520a2ca0cc7651900ea250b31a931adb106dd/ftoi/source/main.c)

This is the number of lines output by diff for comparing wwyleles files against Citra `master` and this PR ("smdc" in the list) output.
A single line difference in the test output will cause 4 diff output lines.
Unfortunaly there is a bug in exception propagation in Citra (seperate PR coming soon). My test has 1/7 denormals => 57142.8571429 denormal lines which would cause the denormal flag in fpscr.
(As there are also totally random inputs in the test we get a couple more errors which are also denormals)

**However, with the exception propagation PR coming soon this will cause 0 errors.**
You can also see that a fpscr of 0x0300009F will produce 0 errors because the denormal flag is set anyway, so the missing exception propagation won't have any effect.

```
Result for Hardware (wwylele)/out-ftoi32-0x03000000.txt:
  master: 60356
  sdmc: 57340
Result for Hardware (wwylele)/out-ftoi32-0x0300009F.txt:
  master: 2716
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi32-0x03400000.txt:
  master: 59420
  sdmc: 57340
Result for Hardware (wwylele)/out-ftoi32-0x03800000.txt:
  master: 74766
  sdmc: 57340
Result for Hardware (wwylele)/out-ftoi32-0x03C00000.txt:
  master: 59420
  sdmc: 57340
Result for Hardware (wwylele)/out-ftoi64-0x03000000.txt:
  master: 57850
  sdmc: 57164
Result for Hardware (wwylele)/out-ftoi64-0x0300009F.txt:
  master: 1356
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi64-0x03400000.txt:
  master: 57850
  sdmc: 57164
Result for Hardware (wwylele)/out-ftoi64-0x03800000.txt:
  master: 72900
  sdmc: 57164
Result for Hardware (wwylele)/out-ftoi64-0x03C00000.txt:
  master: 57850
  sdmc: 57164
```

I didn't write this based on mathematical principles but just by observing the remaining errors and adding more code paths to work around issues.
The code quality is still pretty bad and I'm open to suggestions.

However, the hw-test shows that this is FAR better than this works correctly under lab-conditions and unless the tests are chosen poorly this confirms the functionality.
I've also tried to run a couple of games and they still work.

bunnei:
Code looks fine to me. Is this waiting on any additional testing or anything?

PR #1571, opened 2016-03-26
RasterizerOpenGL: Initialise uniform_block_data

MerryMage:
When I was doing some debugging with valgrind, this was something that was showing up as being read from when uninitialised. Not sure how important this is (or if this needs to be fixed / there's a better way) but putting this here anyway before I forget.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.h
+++ b/src/video_core/renderer_opengl/gl_rasterizer.h
@@ -421,7 +421,7 @@ private:
         UniformData data;
         bool lut_dirty[6];
         bool dirty;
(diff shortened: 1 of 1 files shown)


MerryMage:
Probably irrelevant.

LittleWhite-tb:
Even if it does not cause any errors, should not fix it ?
I mean, first, it will remove the error from valgrind and makes the reading easier. Secondly, we never know if a GPU driver will crash of silly things like this, or not.

MerryMage:
@LittleWhite-tb I might take another look later when I've time.

PR #1572, opened 2016-03-26
DSP: Implement audio filters (simple, biquad)

MerryMage:
In implementing this I went for ease of understanding. Modern compilers probably do a good job optimizing things like this anyway.

Matches firmware output exactly.

### DIFF
--- a/src/audio_core/hle/common.h
+++ b/src/audio_core/hle/common.h
@@ -1,0 +1,35 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <algorithm>
+#include <array>
+
+#include "audio_core/audio_core.h"
+
+#include "common/common_types.h"
+
+namespace DSP {
+namespace HLE {
+
+/// The final output to the speakers is stereo. Preprocessing output in Source is also stereo.
+using StereoFrame16 = std::array<std::array<s16, 2>, AudioCore::samples_per_frame>;
+
+/// The DSP is quadraphonic internally.
+using QuadFrame32   = std::array<std::array<s32, 4>, AudioCore::samples_per_frame>;
+
+/**
+ * This performs the filter operation defined by FilterT::ProcessSample on the frame in-place.
+ * FilterT::ProcessSample is called sequentially on the samples.
+ */
+template<typename FrameT, typename FilterT>
+void FilterFrame(FrameT& frame, FilterT& filter) {
+    std::transform(frame.begin(), frame.end(), frame.begin(), [&filter](const typename FrameT::value_type& sample) {
+        return filter.ProcessSample(sample);
+    });
+}
+
+} // namespace HLE
+} // namespace DSP
--- a/src/audio_core/hle/filter.cpp
+++ b/src/audio_core/hle/filter.cpp
@@ -1,0 +1,115 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <array>
+#include <cstddef>
+
+#include "audio_core/hle/common.h"
+#include "audio_core/hle/dsp.h"
+#include "audio_core/hle/filter.h"
+
+#include "common/common_types.h"
+#include "common/math_util.h"
+
+namespace DSP {
+namespace HLE {
+
+void SourceFilters::Reset() {
+    Enable(false, false);
+}
+
+void SourceFilters::Enable(bool simple, bool biquad) {
+    simple_filter_enabled = simple;
+    biquad_filter_enabled = biquad;
+
+    if (!simple)
+        simple_filter.Reset();
+    if (!biquad)
+        biquad_filter.Reset();
+}
+
+void SourceFilters::Configure(SourceConfiguration::Configuration::SimpleFilter config) {
+    simple_filter.Configure(config);
+}
+
+void SourceFilters::Configure(SourceConfiguration::Configuration::BiquadFilter config) {
+    biquad_filter.Configure(config);
+}
+
+void SourceFilters::ProcessFrame(StereoFrame16& frame) {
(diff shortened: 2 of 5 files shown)


src/audio_core/hle/filter.cpp
```diff
@@ -0,0 +1,114 @@
...
+void SourceFilters::Configure(SourceConfiguration::Configuration::BiquadFilter config) {
+    biquad_filter.Configure(config);
+}
+
+void SourceFilters::ProcessFrame(StereoFrame16& frame) {
+    if (!simple_filter_enabled && !biquad_filter_enabled)
+        return;
+
+    for (auto& x : frame) {
+        if (simple_filter_enabled)
```
lioncash: you can probably make this

```cpp
if (simple_filter_enabled) {
    std::transform(frame.begin(), frame.end(), frame.begin(),
                   [this](const auto& data) { return simple_filter.ProcessSample(data); });
}

if (biquad_filter_enabled) {
    std::transform(frame.begin(), frame.end(), frame.begin(),
                   [this](const auto& data) { return biquad_filter.ProcessSample(data); });
}
```
(possibly extracting the inner code out into a helper somehow) which avoids doing redundant compares in a loop.

src/audio_core/hle/filter.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+// BiquadFilter
+
+void SourceFilters::BiquadFilter::Reset() {
+    x1.fill(0);
+    x2.fill(0);
+    y1.fill(0);
+    y2.fill(0);
+    // Configure as passthrough.
+    a1 = a2 = b1 = b2 = 0;
```
lioncash: Is there any particular reason aX and bX aren't just arrays?

MerryMage: b0 doesn't exist.

src/audio_core/hle/filter.cpp
MerryMage: I'll actually be doing this in a few more places so I'll pop that in common.h, good idea.

src/audio_core/hle/common.h
```diff
@@ -0,0 +1,35 @@
...
+/// The DSP is quadraphonic internally.
+using QuadFrame32   = std::array<std::array<s32, 4>, AudioCore::samples_per_frame>;
+
+/**
+ * This performs the filter operation defined by FilterT::ProcessSample on the frame in-place.
+ * FilterT::ProcessSample is called sequentially on the samples.
+ */
+template<typename FrameT, typename FilterT>
+void FilterFrame(FrameT& frame, FilterT& filter) {
+    std::transform(frame.begin(), frame.end(), frame.begin(), [&filter](const typename FrameT::value_type& sample) {
```
lioncash: is there any particular reacson `const auto& sample` isn't just used? It's effectively the same thing.

JayFoxRox:
I didn't try compiling/running but the code looks fine (Format and algorithmicly).

One question though: Is the `<< 14` / `<< 15` from RE'ing the firmware or did you just choose these for the appropiate precision? (and do those work for all input values if you chose them yourself?)

Other than that, code LGTM

MerryMage:
Shifts are from observation of firmware behaviour, verified ([here are some simplified tests](https://github.com/MerryMage/3ds-audio/)) and produces results exactly equivalent for all values on the firmware revisions I've tested.

Background trivia: When writing DSP firmware it is typical to use fixed point math for anything that touches high frequency stuff because floating point (used to be) too slow. This is one of those things people [love optimizing](http://www.infineon.com/dgdl/sj2004284[1].pdf?fileId=db3a304412b407950112b40f7c7e13fd).

src/audio_core/hle/filter.h
```diff
@@ -0,0 +1,87 @@
...
+
+    /// Reset internal state.
+    void Reset();
+
+    /**
+     * Enable/Disable filters
+     * See also: SourceConfiguration::Configuration::simple_filter_enabled,
+     *           SourceConfiguration::Configuration::biquad_filter_enabled.
+     */
+    void Enable(bool simple, bool biquad);
```
bunnei: Missing docstring descriptions for arguments (ditto to the below methods)

src/audio_core/hle/filter.h
```diff
@@ -0,0 +1,87 @@
...
+        s32 a1, a2, b0, b1, b2;
+        // Internal state
+        std::array<s16, 2> x1;
+        std::array<s16, 2> x2;
+        std::array<s16, 2> y1;
+        std::array<s16, 2> y2;
+    } biquad_filter;
+};
+
+}
```
bunnei: Nitpick, but it's nice to add a `} // namespace blah` comment these for readability

src/audio_core/hle/filter.h
```diff
@@ -0,0 +1,87 @@
...
+}
```
bunnei: Ditto

src/audio_core/hle/filter.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    x2 = x1;
+    x1 = x0;
+    y2 = y1;
+    y1 = y0;
+
+    return y0;
+}
+
+}
```
bunnei: Nitpick, but it's nice to add a `} // namespace blah` comment for readability

src/audio_core/hle/filter.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
```
bunnei: Ditto

bunnei:
I don't really know how this stuff works, so I assume you tested it and it is right. From just a general code review, LGTM, other than the minor comment nits.

MerryMage:
Updated as per @bunnei's comments above.

JamePeng:
LGTM

src/audio_core/hle/common.h
lioncash: Since we compile as C++14 now, the argument can just be `const auto& sample`

MerryMage: Noted. I'll include that in my next PR.

PR #1549, opened 2016-03-20
hid: implement accelerometer and gyroscope back-end

wwylele:
This is just a back-end implement, so that front-end can be implemented with convenience in the future. I made two stubbed function left in `EmuWindow` for front-end.

Most behavior are confirmed by this [hardware test](https://github.com/wwylele/ctrhwtest/tree/master/acc_and_gyro)  

The axis system for gyroscope that I used (with x-y-z successively in each entry) is different from which in 3dbrew and ctrulib (with y and z swapped as x-z-y). I did this to make it consistent with accelerometer's axis. The axis explanation is documented in code.

The behavior of `GetGyroscopeLowCalibrateParam` is unclear, so I stubbed it with the output from my 3ds. Please help find the right way to implement this.

I test these with Kirby Triple Deluxe, by editing the return value of two functions in `EmuWindow` at runtime. The game response as expected.

PR #1573, opened 2016-03-26
Remove unnecessary const & replace pointer with reference

wwylele:
Found these when doing #1549 . Fix them to correspond with new-added one.


bunnei:
LGTM

PR #1576, opened 2016-03-27
Fix various build errors encountered on Clang 3.9 on OS X

smspillaz:
This commit fixes various build errors I encountered when building with clang 3.9

There are also some build errors in nihstro, which I will propose a separate PR for.

### DIFF
--- a/externals/microprofile/microprofileui.h
+++ b/externals/microprofile/microprofileui.h
@@ -1230,7 +1230,12 @@ void MicroProfileDrawDetailedBars(uint32_t nWidth, uint32_t nHeight, int nBaseY,
                 char ThreadName[MicroProfileThreadLog::THREAD_MAX_LEN + 16];
                 const char* cLocal = MicroProfileIsLocalThread(nThreadId) ? "*": " ";
 
+#if defined(WIN32)
+                // nThreadId is 32-bit on Windows
                 int nStrLen = snprintf(ThreadName, sizeof(ThreadName)-1, "%04x: %s%s", nThreadId, cLocal, i < nNumThreadsBase ? &S.Pool[i]->ThreadName[0] : MICROPROFILE_THREAD_NAME_FROM_ID(nThreadId) );
+#else
+                int nStrLen = snprintf(ThreadName, sizeof(ThreadName)-1, "%04llx: %s%s", nThreadId, cLocal, i < nNumThreadsBase ? &S.Pool[i]->ThreadName[0] : MICROPROFILE_THREAD_NAME_FROM_ID(nThreadId) );
+#endif
                 uint32_t nThreadColor = -1;
                 if(nThreadId == nContextSwitchHoverThreadAfter || nThreadId == nContextSwitchHoverThreadBefore)
                     nThreadColor = UI.nHoverColorShared|0x906060;
--- a/externals/microprofile/microprofileui.h
+++ b/externals/microprofile/microprofileui.h
@@ -879,7 +879,7 @@ void MicroProfileDrawDetailedBars(uint32_t nWidth, uint32_t nHeight, int nBaseY,
     static int64_t nRefCpu = 0, nRefGpu = 0;
     if(MicroProfileGetGpuTickReference(&nTickReferenceCpu, &nTickReferenceGpu))
     {
-        if(0 == nRefCpu || abs(nRefCpu-nBaseTicksCpu) > abs(nTickReferenceCpu-nBaseTicksCpu))
+        if(0 == nRefCpu || std::abs(nRefCpu-nBaseTicksCpu) > std::abs(nTickReferenceCpu-nBaseTicksCpu))
         {
             nRefCpu = nTickReferenceCpu;
             nRefGpu = nTickReferenceGpu;
--- a/src/citra_qt/debugger/graphics_framebuffer.cpp
+++ b/src/citra_qt/debugger/graphics_framebuffer.cpp
@@ -346,5 +346,11 @@ u32 GraphicsFramebufferWidget::BytesPerPixel(GraphicsFramebufferWidget::Format f
         case Format::RGBA4:
         case Format::D16:
             return 2;
+        default:
+            UNREACHABLE_MSG("GraphicsFramebufferWidget::BytesPerPixel: this "
+                            "should not be reached as this function should "
+                            "be given a format which is in "
+                            "GraphicsFramebufferWidget::Format. Instead got %i",
+                            static_cast<int>(format));
     }
 }
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -529,7 +529,7 @@ static void ReadRegister() {
         id |= HexCharToValue(command_buffer[2]);
     }
 
-    if (id >= R0_REGISTER && id <= R15_REGISTER) {
+    if (id <= R15_REGISTER) {
         IntToGdbHex(reply, Core::g_app_core->GetReg(id));
     } else if (id == CPSR_REGISTER) {
         IntToGdbHex(reply, Core::g_app_core->GetCPSR());
@@ -584,7 +584,7 @@ static void WriteRegister() {
         id |= HexCharToValue(command_buffer[2]);
     }
 
-    if (id >= R0_REGISTER && id <= R15_REGISTER) {
+    if (id <= R15_REGISTER) {
         Core::g_app_core->SetReg(id, GdbHexToInt(buffer_ptr));
     } else if (id == CPSR_REGISTER) {
         Core::g_app_core->SetCPSR(GdbHexToInt(buffer_ptr));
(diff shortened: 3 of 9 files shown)


src/video_core/shader/shader_interpreter.cpp
```diff
@@ -411,8 +411,8 @@ void RunInterpreter(UnitState<Debug>& state) {
 
                 bool is_inverted = (instr.opcode.Value().EffectiveOpCode() == OpCode::Id::MADI);
 
-                const int address_offset = (instr.mad.address_register_index == 0)
-                                           ? 0 : state.address_registers[instr.mad.address_register_index - 1];
+                const int address_offset = (instr.common.address_register_index == 0)
+                                           ? 0 : state.address_registers[instr.common.address_register_index - 1];
```
wwylele: Why is this?

src/video_core/renderer_opengl/pica_to_gl.h
```diff
@@ -22,7 +22,7 @@ inline GLenum TextureFilterMode(Pica::Regs::TextureConfig::TextureFilter mode) {
     };
 
     // Range check table for input
-    if (mode >= ARRAY_SIZE(filter_mode_table)) {
+    if (static_cast<u32>(mode) >= ARRAY_SIZE(filter_mode_table)) {
```
lioncash: ARRAY_SIZE actually returns a size_t

src/video_core/shader/shader_interpreter.cpp
smspillaz: My compiler complains that `address_register_index` isn't a member of `mad`.

src/video_core/renderer_opengl/pica_to_gl.h
smspillaz: Makes sense, I've changed the type to `size_t`

src/video_core/shader/shader_interpreter.cpp
neobrain: This is not what you want.

Maybe your nihstro submodule is out of sync?

src/video_core/renderer_opengl/pica_to_gl.h
lioncash: Thanks!

src/video_core/shader/shader_interpreter.cpp
smspillaz: It could be that - I did think it was odd that there was an actual build error. I'll update the submodule and check again.

bunnei: This PR cannot me merged until this is addressed

smspillaz: No worries. I've been intending to have a look into this week, but other
commitments have gotten the better of me. I'll address it tonight.

smspillaz:
See also https://github.com/neobrain/nihstro/pull/51

bunnei:
Code LGTM. Just needs the merge conflicts resolved. Thanks!

src/citra_qt/debugger/graphics_framebuffer.cpp
```diff
@@ -346,5 +346,8 @@ u32 GraphicsFramebufferWidget::BytesPerPixel(GraphicsFramebufferWidget::Format f
         case Format::RGBA4:
         case Format::D16:
             return 2;
+        default:
+            UNREACHABLE();
+            break;
```
yuriks: Should also be a return (of some dummy value) for consistency (and to stop possible warnings). Unless the compiler warns about UNREACHABLE I'd skip the break entirely;

yuriks:
Mostly LGTM. Nice fixes, thanks!

src/citra_qt/debugger/graphics_framebuffer.cpp
JayFoxRox: I think we should change `UNREACHABLE` so it is a `[[noreturn]]` function. Then we don't need the `break` or `return`.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -469,7 +469,7 @@ static void ReadRegister() {
         id |= HexCharToValue(command_buffer[2]);
     }
 
-    if (id >= R0_REGISTER && id <= R15_REGISTER) {
+    if (id <= R15_REGISTER) {
```
JayFoxRox: I actually think this change is not good.
I'm not aware of the context but I think the explicit comparison is better in case enums will be changed.

Maybe move this into it's own function: `GetRegisterByIndex()` -> {.type=(GP, CSP, VPF, VFPSystem), .index=()} ?
(Similar code exists in the Pica shader code)

src/core/loader/ncch.cpp
```diff
@@ -256,7 +256,7 @@ ResultStatus AppLoader_NCCH::Load() {
     resource_limit_category = exheader_header.arm11_system_local_caps.resource_limit_category;
 
     LOG_INFO(Loader,  "Name:                        %s"    , exheader_header.codeset_info.name);
-    LOG_INFO(Loader,  "Program ID:                  %016X" , ncch_header.program_id);
+    LOG_INFO(Loader,  "Program ID:                  %016llX" , ncch_header.program_id);
```
JayFoxRox: (nit) I'd prefer a change to `<cinttypes>` (`PRIX64` etc.) but that might be only me. It's a seperate issue though. Just wanted to point it out.

src/video_core/pica.h
```diff
@@ -737,8 +737,9 @@ struct Regs {
         case LightingSampler::ReflectGreen:
         case LightingSampler::ReflectBlue:
             return (config == LightingConfig::Config4) || (config == LightingConfig::Config5) || (config == LightingConfig::Config7);
+        default:
+            return false;
         }
-        return false;
```
JayFoxRox: `UNREACHABLE()` ?

Can we have an unreachable with reason? Something like `UNREACHABLE_MSG("switch(sampler) did not return")`?

(I think your way makes it easy to remove a `return` by accident to cause stupid errors because I don't think anybody will care about the warnings this would cause [until all of them would be silent normally])

src/core/gdbstub/gdbstub.cpp
smspillaz: I generally agree and thought that removing the check for `id >= R0_REGISTER` seemed to be an odd solution. It is a sanity check after all.

The reason I had to remove it was because `id` is a `uint`, so its always greater than zero, or at least always greater than what `R0_REGISTER` is defined to be.

I think it should be possible to suppress the warning if there's a function like

    GetRegisterByIndex(type, index);

which is called as follows:

    if (id >= GetRegisterByIndex(RegisterType::GP, 0) && id <= GetRegisterByIndex(RegisterType::GP, 15)) {
        ...
    }

I'll have to have a look into this later today or this week though.

src/video_core/pica.h
smspillaz: Added `UNREACHABLE_MSG` as suggested.

src/core/gdbstub/gdbstub.cpp
JayFoxRox: I actually meant passing `id` as a paramter and return a struct with register type and index.
The missing check for the lower bound of R0 would only  be missing in one easy to find place then. Yours is fine too but probably complicates the code too much.

Eitherway, you are the more experienced coder so do whatever feels right to you (and is fine with the citra maintainers)

src/core/loader/ncch.cpp
Subv: I agree

kemenaran:
Tested on OS X 10.11 with latest clang, works for me 👍

bunnei:
@smspillaz Can you please squash the fixups? Otherwise, this is good to go.

I don't like to squash on my end, because it destroys all commit history.

externals/microprofile/microprofileui.h
```diff
@@ -1230,7 +1230,7 @@ void MicroProfileDrawDetailedBars(uint32_t nWidth, uint32_t nHeight, int nBaseY,
                 char ThreadName[MicroProfileThreadLog::THREAD_MAX_LEN + 16];
                 const char* cLocal = MicroProfileIsLocalThread(nThreadId) ? "*": " ";
 
-                int nStrLen = snprintf(ThreadName, sizeof(ThreadName)-1, "%04x: %s%s", nThreadId, cLocal, i < nNumThreadsBase ? &S.Pool[i]->ThreadName[0] : MICROPROFILE_THREAD_NAME_FROM_ID(nThreadId) );
+                int nStrLen = snprintf(ThreadName, sizeof(ThreadName)-1, "%04llx: %s%s", nThreadId, cLocal, i < nNumThreadsBase ? &S.Pool[i]->ThreadName[0] : MICROPROFILE_THREAD_NAME_FROM_ID(nThreadId) );
```
jcowgill: `nThreadId` is 32-bit on windows.

https://github.com/citra-emu/citra/blob/master/externals/microprofile/microprofile.h#L207

smspillaz: I've added an `#ifdef` to handle that case now.

bunnei:
Thanks! LGTM

PR #1589, opened 2016-03-28
Compilation fix

LittleWhite-tb:
Attempt to fix compilation issue from #1541
It needs additional testing, to see if my modification does not cause a regression with #1541.

### DIFF
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -66,7 +66,7 @@ void GameList::ValidateEntry(const QModelIndex& item)
 
     if (file_path.isEmpty())
         return;
-    std::string std_file_path = file_path.toLocal8Bit();
+    std::string std_file_path(file_path.toLocal8Bit());
     if (!FileUtil::Exists(std_file_path) || FileUtil::IsDirectory(std_file_path))
         return;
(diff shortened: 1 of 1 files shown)


src/citra_qt/game_list.cpp
```diff
@@ -66,7 +66,7 @@ void GameList::ValidateEntry(const QModelIndex& item)
 
     if (file_path.isEmpty())
         return;
-    std::string std_file_path = file_path.toLocal8Bit();
+    std::string std_file_path = file_path.toLocal8Bit().toStdString();
```
JayFoxRox: There is no such function in Qt 4.x hence the travis build failing. I'm not so sure about this one. I'm not sure which version of Qt citra is targeting, but this seems like a kludge on kludge anyway.

I'd revert #1541 if it really breaks compilation on Linux. (I'll try it later and feedback in the original PR)

LittleWhite-tb:
I review the problem one more time, and found a better solution (I hope).

LFsWang:
it's woks!
I still try on windows. Some C functions (fopen) have same problem like FindFirstFileA on Windows...
and I have mid-exam this week...

bunnei:
LGTM

PR #1611, opened 2016-03-29
Add common methods to all cfg:* ports

ObsidianX:
According to 3dbrew all of the cfg ports share a common set of methods which were previously only implemented in cfg:u.  ctrulib will attempt to connect to the highest privilege port available with the assumption that these shared methods exist across all ports.


MerryMage:
LGTM

bunnei:
LGTM

PR #1612, opened 2016-03-29
SOC:U GetSockOpt/SetSockOpt

ObsidianX:
-Implement `GetSockOpt`/`SetSockOpt`
-Fix bug in `RecvFrom` where sending from localhost does not fill in `src_addr`/`src_addr_len` on Linux

`SOC_U::RecvFrom` was crashing due to odd behavior in Linux where `src_addr` was not being filled in, and `src_addr_len` was being set to `0`.  This caused `src_addr.sa_family` to become invalid, failing `CTRSockAddr::FromPlatform`'s assertion that `sa_family` must be `AF_INET`.

### DIFF
--- a/src/core/hle/service/soc_u.cpp
+++ b/src/core/hle/service/soc_u.cpp
@@ -568,7 +568,7 @@ static void RecvFrom(Service::Interface* self) {
     socklen_t src_addr_len = sizeof(src_addr);
     int ret = ::recvfrom(socket_handle, (char*)output_buff, len, flags, &src_addr, &src_addr_len);
 
-    if (buffer_parameters.output_src_address_buffer != 0) {
+    if (buffer_parameters.output_src_address_buffer != 0 && src_addr_len > 0) {
         CTRSockAddr* ctr_src_addr = reinterpret_cast<CTRSockAddr*>(Memory::GetPointer(buffer_parameters.output_src_address_buffer));
         *ctr_src_addr = CTRSockAddr::FromPlatform(src_addr);
     }
@@ -724,6 +724,49 @@ static void ShutdownSockets(Service::Interface* self) {
     cmd_buffer[1] = 0;
 }
 
+static void GetSockOpt(Service::Interface* self) {
+    u32* cmd_buffer = Kernel::GetCommandBuffer();
+    u32 socket_handle = cmd_buffer[1];
+    u32 level = cmd_buffer[2];
+    u32 optname = cmd_buffer[3];
+    u32 optlen = cmd_buffer[4];
+
+    // 0x100 = static buffer offset (bytes)
+    // + 0x4 = 2nd pointer (u32) position
+    // >> 2  = convert to u32 offset instead of byte offset (cmd_buffer = u32*)
+    u8* optval = Memory::GetPointer(cmd_buffer[0x104 >> 2]);
+
+    int ret = ::getsockopt(socket_handle, level, optname, &optval, &optlen);
+    int err = 0;
+    if(ret == SOCKET_ERROR_VALUE) {
+        err = TranslateError(GET_ERRNO);
+    }
+
+    cmd_buffer[0] = IPC::MakeHeader(0x11, 4, 2);
+    cmd_buffer[1] = ret;
+    cmd_buffer[2] = err;
+    cmd_buffer[3] = optlen;
+}
+
+static void SetSockOpt(Service::Interface* self) {
+    u32* cmd_buffer = Kernel::GetCommandBuffer();
+    u32 socket_handle = cmd_buffer[1];
+    u32 level = cmd_buffer[2];
+    u32 optname = cmd_buffer[3];
+    socklen_t optlen = static_cast<socklen_t>(cmd_buffer[4]);
+    void *optval = Memory::GetPointer(cmd_buffer[8]);
+
+    int ret = static_cast<u32>(::setsockopt(socket_handle, level, optname, optval, optlen));
+    int err = 0;
+    if(ret == SOCKET_ERROR_VALUE) {
+        err = TranslateError(GET_ERRNO);
+    }
+
+    cmd_buffer[0] = IPC::MakeHeader(0x12, 4, 4);
+    cmd_buffer[1] = ret;
+    cmd_buffer[2] = err;
+}
+
 const Interface::FunctionInfo FunctionTable[] = {
     {0x00010044, InitializeSockets,             "InitializeSockets"},
     {0x000200C2, Socket,                        "Socket"},
@@ -741,8 +784,8 @@ const Interface::FunctionInfo FunctionTable[] = {
     {0x000E00C2, nullptr,                       "GetHostByAddr"},
     {0x000F0106, nullptr,                       "GetAddrInfo"},
     {0x00100102, nullptr,                       "GetNameInfo"},
-    {0x00110102, nullptr,                       "GetSockOpt"},
-    {0x00120104, nullptr,                       "SetSockOpt"},
+    {0x00110102, GetSockOpt,                    "GetSockOpt"},
+    {0x00120104, SetSockOpt,                    "SetSockOpt"},
     {0x001300C2, Fcntl,                         "Fcntl"},
     {0x00140084, Poll,                          "Poll"},
     {0x00150042, nullptr,                       "SockAtMark"},
--- a/src/core/hle/service/soc_u.cpp
+++ b/src/core/hle/service/soc_u.cpp
@@ -729,11 +729,7 @@ static void GetSockOpt(Service::Interface* self) {
     u32 socket_handle = cmd_buffer[1];
     u32 level = cmd_buffer[2];
     u32 optname = cmd_buffer[3];
-#ifdef _WIN32
-    int optlen = (int)cmd_buffer[4];
-#else
     socklen_t optlen = (socklen_t)cmd_buffer[4];
-#endif
 
     // 0x100 = static buffer offset (bytes)
     // + 0x4 = 2nd pointer (u32) position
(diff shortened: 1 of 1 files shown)


src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +724,49 @@ static void ShutdownSockets(Service::Interface* self) {
...
+    u32* cmd_buffer = Kernel::GetCommandBuffer();
+    u32 socket_handle = cmd_buffer[1];
+    u32 level = cmd_buffer[2];
+    u32 optname = cmd_buffer[3];
+    socklen_t optlen = static_cast<socklen_t>(cmd_buffer[4]);
+    void *optval = Memory::GetPointer(cmd_buffer[8]);
+
+    int ret = static_cast<u32>(::setsockopt(socket_handle, level, optname, optval, optlen));
+    int err = 0;
+    if(ret == SOCKET_ERROR_VALUE) {
```
lioncash: space after the `if`

src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +724,49 @@ static void ShutdownSockets(Service::Interface* self) {
...
+    u32 optlen = cmd_buffer[4];
+
+    // 0x100 = static buffer offset (bytes)
+    // + 0x4 = 2nd pointer (u32) position
+    // >> 2  = convert to u32 offset instead of byte offset (cmd_buffer = u32*)
+    u8* optval = Memory::GetPointer(cmd_buffer[0x104 >> 2]);
+
+    int ret = ::getsockopt(socket_handle, level, optname, &optval, &optlen);
+    int err = 0;
+    if(ret == SOCKET_ERROR_VALUE) {
```
lioncash: space after the `if`

src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +724,49 @@ static void ShutdownSockets(Service::Interface* self) {
...
+    cmd_buffer[3] = optlen;
+}
+
+static void SetSockOpt(Service::Interface* self) {
+    u32* cmd_buffer = Kernel::GetCommandBuffer();
+    u32 socket_handle = cmd_buffer[1];
+    u32 level = cmd_buffer[2];
+    u32 optname = cmd_buffer[3];
+    socklen_t optlen = static_cast<socklen_t>(cmd_buffer[4]);
+    void *optval = Memory::GetPointer(cmd_buffer[8]);
```
lioncash: Properly cast the pointer down to a char pointer. void* does not mean `wildcard`.

src/core/hle/service/soc_u.cpp
```diff
@@ -568,7 +568,7 @@ static void RecvFrom(Service::Interface* self) {
     socklen_t src_addr_len = sizeof(src_addr);
     int ret = ::recvfrom(socket_handle, (char*)output_buff, len, flags, &src_addr, &src_addr_len);
 
-    if (buffer_parameters.output_src_address_buffer != 0) {
+    if (buffer_parameters.output_src_address_buffer != 0 && src_addr_len > 0) {
```
MerryMage: Is it possible for `src_addr_len <= 0`?

src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +724,49 @@ static void ShutdownSockets(Service::Interface* self) {
...
+    u32 level = cmd_buffer[2];
+    u32 optname = cmd_buffer[3];
+    u32 optlen = cmd_buffer[4];
+
+    // 0x100 = static buffer offset (bytes)
+    // + 0x4 = 2nd pointer (u32) position
+    // >> 2  = convert to u32 offset instead of byte offset (cmd_buffer = u32*)
+    u8* optval = Memory::GetPointer(cmd_buffer[0x104 >> 2]);
+
+    int ret = ::getsockopt(socket_handle, level, optname, &optval, &optlen);
```
MerryMage: `&optval` doesn't look correct.

src/core/hle/service/soc_u.cpp
ObsidianX: Another check has been added to this to ensure that `ret >= 0`.  I ran into another edge case where `recvfrom` failed, but `src_addr_len` hadn't yet been modified, passing the other check.

src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +724,49 @@ static void ShutdownSockets(Service::Interface* self) {
...
+
+static void SetSockOpt(Service::Interface* self) {
+    u32* cmd_buffer = Kernel::GetCommandBuffer();
+    u32 socket_handle = cmd_buffer[1];
+    u32 level = cmd_buffer[2];
+    u32 optname = cmd_buffer[3];
+    socklen_t optlen = static_cast<socklen_t>(cmd_buffer[4]);
+    u8 *optval = Memory::GetPointer(cmd_buffer[8]);
+
+    int ret = static_cast<u32>(::setsockopt(socket_handle, level, optname, optval, optlen));
```
lioncash: opt_val needs to be reinterpret_cast-ed into char* (like I previously mentioned). This will break on Windows (which has this parameter as `char*`, see [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ms740476%28v=vs.85%29.aspx))

src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +724,49 @@ static void ShutdownSockets(Service::Interface* self) {
...
+    u32 level = cmd_buffer[2];
+    u32 optname = cmd_buffer[3];
+    socklen_t optlen = (socklen_t)cmd_buffer[4];
+
+    // 0x100 = static buffer offset (bytes)
+    // + 0x4 = 2nd pointer (u32) position
+    // >> 2  = convert to u32 offset instead of byte offset (cmd_buffer = u32*)
+    char* optval = reinterpret_cast<char*>(Memory::GetPointer(cmd_buffer[0x104 >> 2]));
+
+    int ret = ::getsockopt(socket_handle, level, optname, optval, &optlen);
```
Lectem: optname doesn't map 1-1 from 3DS to other platforms, it even differs from linux to windows.
Also a lot of options are not supported by the 3DS.
I'd recommend to have a whitelist based on ctrulib headers, and convert the optname values accordingly.

src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +744,72 @@ static void ShutdownSockets(Service::Interface* self) {
...
+#ifdef _WIN32
+        err = WSAEINVAL;
+#else
+        err = EINVAL;
+#endif
+    } else {
+        // 0x100 = static buffer offset (bytes)
+        // + 0x4 = 2nd pointer (u32) position
+        // >> 2  = convert to u32 offset instead of byte offset (cmd_buffer = u32*)
+        char *optval = reinterpret_cast<char *>(Memory::GetPointer(cmd_buffer[0x104 >> 2]));
```
lioncash: Asterisks go against the type

src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +744,72 @@ static void ShutdownSockets(Service::Interface* self) {
...
+
+    if(optname < 0) {
+#ifdef _WIN32
+        err = WSAEINVAL;
+#else
+        err = EINVAL;
+#endif
+    } else {
+        socklen_t optlen = static_cast<socklen_t>(cmd_buffer[4]);
+        const char *optval = reinterpret_cast<const char *>(Memory::GetPointer(cmd_buffer[8]));
```
lioncash: Asterisks against type

ObsidianX:
Bump. All previous comments have been addressed.

Lectem:
Filtering is good, but there is still a conversion missing, the values of SO_* will differ from a platform to another from the ones of the 3DS. For example SO_ERROR is 0x1009 on 3ds but 0x1007 on linux

ObsidianX:
@Lectem Ahh, now I see. I wasn't able to see the values of the SO_* options on MSDN and didn't think Microsoft would be lame enough to change the values of those constants ;)  As an interesting side note: it turns out the 3DS uses values that almost entirely match WinSock's definitions (SO_ERROR is different).

Lectem:
@ObsidianX yeah, most of the constants look very much like winsock implementation, and sometimes differ more from Unix than Windows. In a general way, the 3DS OS looks based on an old Windows.

ObsidianX:
Is there anything else blocking this PR?

src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +752,72 @@ static void ShutdownSockets(Service::Interface* self) {
...
+        }
+    }
+
+    cmd_buffer[0] = IPC::MakeHeader(0x11, 4, 2);
+    cmd_buffer[1] = ret;
+    cmd_buffer[2] = err;
+    cmd_buffer[3] = optlen;
+}
+
+static void SetSockOpt(Service::Interface* self) {
```
bunnei: Missing docstring

src/core/hle/service/soc_u.cpp
```diff
@@ -724,6 +752,72 @@ static void ShutdownSockets(Service::Interface* self) {
     cmd_buffer[1] = 0;
 }
 
+static void GetSockOpt(Service::Interface* self) {
```
bunnei: Missing docstring

bunnei:
Can you squash this to one commit?

src/core/hle/service/soc_u.cpp
ObsidianX: There aren't any docstrings for any of the 1:1 methods.  See: `Bind`, `Socket`, `Fcntl`, etc.

ObsidianX:
@bunnei I believe Github now has that as an option when merging a PR.

Issue #1610, opened 2016-03-28
Batman Arkham Origins Blackgate - Won't boot


JohnGodgames:
[LOG](http://pasted.co/872a0556)

RavenHome1:
maybe a cro issue?

wwylele:
@RavenHome1 If it were about CRO, the log would say something about 'LoadExeCro'

KillzXGaming:
Isn't this same as #1547? Both have same error where GSP event cannot be created.

mailwl:
@JohnGodgames , try #1613
it shows intro now, but later back screen, maybe geometry shaders, like MH3U and others, don't know...

JohnGodgames:
Yep #1613 helps

![batman](https://cloud.githubusercontent.com/assets/11453507/14126620/97c0e180-f612-11e5-9b32-2b9c85959cd9.JPG)

bunnei:
Should this be closed?

PR #1613, opened 2016-03-29
Set Kernel config "Hardware Inited" to 1 (true)

mailwl:
pass through endless loop:
Batman Arkham Origins Blackgate - Won't boot #1610
Anpanman to Touch de Waku Waku Training

### DIFF
--- a/src/core/hle/shared_page.cpp
+++ b/src/core/hle/shared_page.cpp
@@ -16,6 +16,9 @@ void Init() {
     std::memset(&shared_page, 0, sizeof(shared_page));
 
     shared_page.running_hw = 0x1; // product
+
+    // Some games wait until this value becomes 0x1, before asking running_hw
+    shared_page.unknown_value = 0x1;
 }
 
 } // namespace
--- a/src/core/hle/shared_page.h
+++ b/src/core/hle/shared_page.h
@@ -39,12 +39,14 @@ struct SharedPageDef {
     DateTime date_time_0;                // 20
     DateTime date_time_1;                // 40
     u8       wifi_macaddr[6];            // 60
-    u8       wifi_unknown1;              // 66
+    u8       wifi_link_level;            // 66
     u8       wifi_unknown2;              // 67
     INSERT_PADDING_BYTES(0x80 - 0x68);   // 68
     float_le sliderstate_3d;             // 80
     u8       ledstate_3d;                // 84
-    INSERT_PADDING_BYTES(0xA0 - 0x85);   // 85
+    INSERT_PADDING_BYTES(1);             // 85
+    u8       unknown_value;              // 86
+    INSERT_PADDING_BYTES(0xA0 - 0x87);   // 87
     u64_le   menu_title_id;              // A0
     u64_le   active_menu_title_id;       // A8
     INSERT_PADDING_BYTES(0x1000 - 0xB0); // B0


KillzXGaming:
Some games i've tested with this with infinite loops now get fatal errors. 
Luigi's Mansion 2
[http://pastebin.com/aAaY72nT](LOG)

Mario Party Island Tour
[http://pastebin.com/RZUvVC1j](LOG)

mailwl:
ic. well, i hope someone adds "dlp:SRVR" service to Citra

KillzXGaming:
Wouldn't that be the download play client? Unsure why games would rely on this service.
https://www.3dbrew.org/wiki/DLP_Services

mailwl:
yes, it is. but there is no sources in Citra for this service, so games, which try to access it, immediately breaks. but it's other issue, not related with this PR

KillzXGaming:
Ok, feel free to delete those comments as they do not relate to the PR. I may add this as an issue.

JohnGodgames:
This PR also helps with:

Bugs vs. Tanks!

KillzXGaming:
This and
https://github.com/exhalatio/citra/tree/dlp_dummy
Fixed Luigi's Mansion 2
ediargz

JohnGodgames:
@exhalatio can you make a PR?

I'm confirming https://github.com/exhalatio/citra/tree/dlp_dummy + this PR

marion island

mansion

Leo626:
Rhythm Heaven The best+ now boots for me when additionally merged with exhalatio's [Dummy implementation dlp: SRVR Service. Only result success](https://github.com/exhalatio/citra/commit/8a689c9e730b19ed4808dd0028fe9d79813debc4) commit.

citra-qt_2016-03-29_20-43-47

exhalatio:
@JohnGodgames
Hi, I make a PR it.

MoochMcGee:
HACKS
A
C
K
S

mailwl:
rename the variable to running_hw2

JamePeng:
http://3dbrew.org/wiki/Configuration_Memory
I think it is totally wrong about the "running_hw2"

mailwl:
hm... well, you right. it looks more like "Hardware Initialized"
```c++
int Anpanman_stuck()
{
  while ( !(v1FF81086 & 1) )
    sleep(1000000, 0);
  return v1FF81004;
}
```

yuriks:
Where exactly did this "hardware inited" interpretation come from? There's absolutely nothing in the 3dbrew page that would suggest it, as far as I can tell.

mailwl:
it's just my guess, from the code, where game Anpanman is stuck.
before get type of running hardware at 0x1FF81004, it waits until 0x1FF81086 become 1.

bunnei:
Yeah, not a fan of the name "hardware_inited" if we don't actually know the purpose of this field... I'm concerned it could be taken as a 'known' and in fact be something totally different.

IMO, I'm okay with just calling this "unknown_foo" and adding a well-written comment in the code explaining what we know about it. I think this still falls in line with "stubbing" functionality rather than "hacking". Also might not hurt to open an issue about this (to do more testing and try to figure out what it is, or go back and update if/when 3dbrew documents it) and reference it in the comment.

JayFoxRox:
There should be a hardware test which does the same code with a very short sleep (to force context switches if that is necessary) and logs the values (first to memory, later to file) of both registers.

This could indeed be a field which reports wether the OS already knows what kind of hardware it is running on (which is still not what `hardware_initialized` would suggest..). Maybe the value of v1FF81004 is bad while v1FF81086 is low. We should also try forcing these regs to different values to find out when the OS controls them (or static analysis - whatever floats your boat)

But I agree with bunnei - create a new issue for this and call the variable `unknown_...` until then.
Then get this merged asap unless it causes regressions.

yuriks:
@bunnei Seems Good To Me

bunnei:
@mailwl Please address the above comment and I think this can be merged!

mailwl:
done

src/core/hle/shared_page.cpp
```diff
@@ -16,6 +16,9 @@ void Init() {
     std::memset(&shared_page, 0, sizeof(shared_page));
 
     shared_page.running_hw = 0x1; // product
+
+    // Some games waits until this value become 0x1, before asking running_hw
```
JayFoxRox: (nit) ~ "// Some games wait until this value becomes 0x1, before reading 'running_hw'"

LGTM other than that

mailwl:
done

bunnei:
LGTM

PR #1618, opened 2016-03-30
Prevent cache overflow when single stepping

MerryMage:
Calling `Core::SingleStep()` at the start of a 20 instruction basic block causes the translation of 20 instructions followed by the execution of one instruction. Calling `Core::SingleStep()` again causes the translation of 19 instructions followed by the execution of one instruction.

This is a O(n^2) use of the cache. The cache can be quickly overwhelmed when single-stepping a lot (for example, if you're writing a JIT that single-steps the interpreter on unknown instructions).

This PR fixes that. It also improves the speed of single-stepping in the JIT so that's a bonus.


bunnei:
LGTM

JamePeng:
LGTM

PR #1620, opened 2016-03-31
Fix filename&path encode problem on Windows

### LINKED ISSUE #1584: Qt: Handle wide character strings
Certain string locales are incompatible with the Qt GUI. See #1541 for more information. This is an incomplete fix. Instead, usage of FindFirstFileA etc. should be replaced with FindFirstFileW and handle the wide character strings correctly.

LFsWang:
fix #1584 #730 
Use wide version functions for Windows. It fix file name & dir path encode problem.

### Screenshot
image

image

### Booting with unicode dir path.
`SetConsoleOutputCP(CP_UTF8);` can fix console

image

image

with `SetConsoleOutputCP(CP_UTF8);`

(Consolas)
image

(default font style)
image


lioncash:
There's no need to update copyright years. The year establishes how far back the copyright claim was established.

LFsWang:
reverted

bunnei:
LGTM

PR #1622, opened 2016-04-01
Avoid warnings by casting to size_t for ARRAY_SIZE() comparisons

JayFoxRox:
Just something I've been annoyed with lately: clang complains about these (`unsigned` would be fine but ARRAY_SIZE() gives a `size_t` - so I prefer that).
There are other warnings I'd like to be silent but this will do for now.


lioncash:
static_cast (using this will also get rid of a few other warnings when compiling with -Weverything)

PR #1462, opened 2016-03-06
Pica: Write depth value even when depth test is disabled

yuriks:
This has been confirmed on hardware. Together with #1461, fixes Etrian Odyssey IV.

PR #1624, opened 2016-04-01
Implement buffer-write allow registers

JayFoxRox:
This PR implements 2 of the 4 "allow"-registers which are probably used for powergating (as suggested by @yuriks).
\- Some games such as Tappingo still depend on them for disabling depth/stencil writes.

I've chosen the name "allow" myself and I'd be happy to change it to something more fitting.
Essentially those are read/write-enable registers for the buffers.
I use the term "buffer" when it refers to color-, depth- and/or stencil-buffer.

While the RGBA register field is 4 bits it will still control all channels at once, it's not a bitmask.
Depth and stencil share 2 bit. It also controls access to both buffers at once and is not a bitmask.
I did not yet test wether there are more bits which are tested in either of those registers.

This PR implements the write-permission fully in the SW-Rasterizer and the HW-Renderer.
The read-permission will be handled in a seperate PR because the implementation would be somewhat more tricky.

This should fix most of the bugs caused by #1462 .

There is also another commit which fixes a bug which is also present in `master` where the GL stencil test could run out of sync with the Pica stencil test:
The stencil test sync function also checks wether the current Pica framebuffer has a stencil buffer.
The call to the sync function was missing in case the framebuffer is changed after stencil testing was configured (Potentially causing GL to enable stencil with no stencil buffer available - I'm not sure if this could cause problems).

---

This PR was tested [using a HW-Test](https://github.com/JayFoxRox/3ds-tests/tree/master/buffer-allow) ([Results](https://docs.google.com/spreadsheets/d/113T3Grv5HXEBrXbtdSjFQXuXmb0IO7gETu55ka7djNA/edit#gid=0)).
There are no known regressions while this fixes some games.

(I'd like to thank @JohnGodGames for testing)

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -271,6 +271,7 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
     // Stencil test
     case PICA_REG_INDEX(output_merger.stencil_test.raw_func):
     case PICA_REG_INDEX(output_merger.stencil_test.raw_op):
+    case PICA_REG_INDEX(framebuffer.depth_format):
         SyncStencilTest();
         break;
 
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -153,6 +153,9 @@ void RasterizerOpenGL::Reset() {
     SyncLogicOp();
     SyncStencilTest();
     SyncDepthTest();
+    SyncColorWriteMask();
+    SyncStencilWriteMask();
+    SyncDepthWriteMask();
 
     SetShader();
 
@@ -268,16 +271,23 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
         state.draw.shader_dirty = true;
         break;
 
-    // Stencil test
+    // Sync GL stencil test + stencil write mask
+    // (Pica stencil test function register also contains a stencil write mask)
     case PICA_REG_INDEX(output_merger.stencil_test.raw_func):
+        SyncStencilTest();
+        SyncStencilWriteMask();
+        break;
     case PICA_REG_INDEX(output_merger.stencil_test.raw_op):
     case PICA_REG_INDEX(framebuffer.depth_format):
         SyncStencilTest();
         break;
 
-    // Depth test
+    // Sync GL depth test + depth and color write mask
+    // (Pica depth test function register also contains a depth and color write mask)
     case PICA_REG_INDEX(output_merger.depth_test_enable):
         SyncDepthTest();
+        SyncDepthWriteMask();
+        SyncColorWriteMask();
         break;
 
     // Logic op
@@ -881,13 +891,30 @@ void RasterizerOpenGL::SyncLogicOp() {
     state.logic_op = PicaToGL::LogicOp(Pica::g_state.regs.output_merger.logic_op);
 }
 
+void RasterizerOpenGL::SyncColorWriteMask() {
+    const auto& regs = Pica::g_state.regs;
+    state.color_mask.red_enabled = regs.output_merger.red_enable;
+    state.color_mask.green_enabled = regs.output_merger.green_enable;
+    state.color_mask.blue_enabled = regs.output_merger.blue_enable;
+    state.color_mask.alpha_enabled = regs.output_merger.alpha_enable;
+}
+
+void RasterizerOpenGL::SyncStencilWriteMask() {
+    const auto& regs = Pica::g_state.regs;
+    state.stencil.write_mask = regs.output_merger.stencil_test.write_mask;
+}
+
+void RasterizerOpenGL::SyncDepthWriteMask() {
+    const auto& regs = Pica::g_state.regs;
+    state.depth.write_mask = regs.output_merger.depth_write_enable ? GL_TRUE : GL_FALSE;
+}
+
 void RasterizerOpenGL::SyncStencilTest() {
     const auto& regs = Pica::g_state.regs;
     state.stencil.test_enabled = regs.output_merger.stencil_test.enable && regs.framebuffer.depth_format == Pica::Regs::DepthFormat::D24S8;
     state.stencil.test_func = PicaToGL::CompareFunc(regs.output_merger.stencil_test.func);
     state.stencil.test_ref = regs.output_merger.stencil_test.reference_value;
     state.stencil.test_mask = regs.output_merger.stencil_test.input_mask;
-    state.stencil.write_mask = regs.output_merger.stencil_test.write_mask;
     state.stencil.action_stencil_fail = PicaToGL::StencilOp(regs.output_merger.stencil_test.action_stencil_fail);
     state.stencil.action_depth_fail = PicaToGL::StencilOp(regs.output_merger.stencil_test.action_depth_fail);
     state.stencil.action_depth_pass = PicaToGL::StencilOp(regs.output_merger.stencil_test.action_depth_pass);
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -876,13 +896,31 @@ void RasterizerOpenGL::SyncLogicOp() {
...
+    bool allow_color_write = regs.framebuffer.allow_color_write != 0;
+    state.color_mask.red_enabled = allow_color_write && regs.output_merger.red_enable ? GL_TRUE : GL_FALSE;
+    state.color_mask.green_enabled = allow_color_write && regs.output_merger.green_enable ? GL_TRUE : GL_FALSE;
+    state.color_mask.blue_enabled = allow_color_write && regs.output_merger.blue_enable ? GL_TRUE : GL_FALSE;
+    state.color_mask.alpha_enabled = allow_color_write && regs.output_merger.alpha_enable ? GL_TRUE : GL_FALSE;
+}
+
+void RasterizerOpenGL::SyncStencilWriteMask() {
+    const auto& regs = Pica::g_state.regs;
+    state.stencil.write_mask = regs.framebuffer.allow_depth_stencil_write != 0 ? (GLuint)regs.output_merger.stencil_test.write_mask : 0;
```
lioncash:
```cpp
state.stencil.write_mask = regs.framebuffer.allow_depth_stencil_write != 0
                         ? static_cast<GLuint>(regs.output_merger.stencil_test.write_mask)
                         : 0;
```

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -344,6 +344,15 @@ class RasterizerOpenGL : public VideoCore::RasterizerInterface {
     /// Syncs the logic op states to match the PICA register
     void SyncLogicOp();
 
+    /// Syncs the color write mask to match the PICA register
+    void SyncColorWriteMask();
+
+    /// Syncs the stencil write mask to match the PICA register
+    void SyncStencilWriteMask();
+
+    /// Syncs the depth write mask to match the PICA register
```
JayFoxRox: Either these comments should read "PICA registers" (indicating the dependency on different registers) or "PICA state" (indicating roughly the same)
Preferences?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -269,14 +272,31 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
...
         break;
 
     // Depth test
     case PICA_REG_INDEX(output_merger.depth_test_enable):
         SyncDepthTest();
+        SyncDepthWriteMask();
+        SyncColorWriteMask();
+        break;
+
+    // Stencil and depth write mask
```
JayFoxRox: (nit) will change to `// Depth and stencil write mask`

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -876,13 +896,31 @@ void RasterizerOpenGL::SyncLogicOp() {
     state.logic_op = PicaToGL::LogicOp(Pica::g_state.regs.output_merger.logic_op);
 }
 
+void RasterizerOpenGL::SyncColorWriteMask() {
```
lioncash:
```cpp
void RasterizerOpenGL::SyncColorWriteMask() {
    const auto& regs = Pica::g_state.regs;
    bool allow_color_write = regs.framebuffer.allow_color_write != 0;

    auto write_enabled = [allow_color_write](u32 value) {
        return (allow_color_write && value != 0) ? GL_TRUE : GL_FALSE;
    }

    state.color_mask.red_enabled = write_enabled(regs.output_merger.red_enable);
    ... etc
}
```

wwylele:
Great! This fixes several graphical glitches in Kirby Triple Deluxe ( keychain didn't show in upper screen, kirby didn't show in fighters mode when selecting, kirby disappeared when entering stage, etc.).

JohnGodgames:
Recent master vs buffer-allow-write

**Tappingo**
tappingo citra vs buffer-allow

**Disney Planes**
disney planes citra vs buffer-allow

**Thorium Wars**
thorium wars citra vs buffer-allow

**The Legend of Zelda - A Link Between Worlds DEMO**
tlzlbw demo citra vs buffer-allow

and propably more...

KillzXGaming:
Smash Bros Demo
Master without PR
http://i.imgur.com/QkVpKHc.png

Merged With PR
http://i.imgur.com/Bpm9tUf.png

KillzXGaming:
Paper Mario Sticker Star seems to break with this PR. Anyone else tested this game?
[http://i.imgur.com/2AUZ7Zy.png](url)

JayFoxRox:
That looks like clipping gone wrong and could be a seperate issue. Needs investigation

bunnei:
LGTM

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -876,13 +896,39 @@ void RasterizerOpenGL::SyncLogicOp() {
...
+void RasterizerOpenGL::SyncStencilWriteMask() {
+    const auto& regs = Pica::g_state.regs;
+    state.stencil.write_mask = regs.framebuffer.allow_depth_stencil_write != 0
+                             ? static_cast<GLuint>(regs.output_merger.stencil_test.write_mask)
+                             : 0;
+}
+
+void RasterizerOpenGL::SyncDepthWriteMask() {
+    const auto& regs = Pica::g_state.regs;
+    state.depth.write_mask = regs.framebuffer.allow_depth_stencil_write != 0 && regs.output_merger.depth_write_enable
```
linkmauve: I would put the condition in parentheses here, not that it would change anything but it’s more readable.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -876,13 +896,39 @@ void RasterizerOpenGL::SyncLogicOp() {
...
+
+    state.color_mask.red_enabled = IsColorWriteEnabled(regs.output_merger.red_enable);
+    state.color_mask.green_enabled = IsColorWriteEnabled(regs.output_merger.green_enable);
+    state.color_mask.blue_enabled = IsColorWriteEnabled(regs.output_merger.blue_enable);
+    state.color_mask.alpha_enabled = IsColorWriteEnabled(regs.output_merger.alpha_enable);
+}
+
+void RasterizerOpenGL::SyncStencilWriteMask() {
+    const auto& regs = Pica::g_state.regs;
+    state.stencil.write_mask = regs.framebuffer.allow_depth_stencil_write != 0
```
linkmauve: And here too.

linkmauve:
Aside from that, LGTM.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -269,14 +272,31 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
         break;
 
     // Stencil test
+    case PICA_REG_INDEX(framebuffer.depth_format):
+        SyncStencilTest();
+        break;
```
bunnei: Missing space after this line

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -269,14 +272,31 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
         break;
 
     // Stencil test
```
bunnei: Get rid of this comment and replace with something more specific to why your're syncing here, as well as something more specific to the proceeding case statements (?)

src/video_core/renderer_opengl/gl_rasterizer.cpp
JayFoxRox: Unfortunaly this is just how this function works..: Comment about GL feature in question, then a handful of Pica regs (or fields in such regs) which hopefully describe everything that is read in sync function, list of all sync functions using said register. Then the next register which controls the same GL feature *without empty line*.
Then finally an empty line and the comment for the next GL feature block.

As this is a giant switch case it's impossible to list regs twice for different featues so we have to merge different GL features whenever different GL features use the same reg as input.

I think adding the space would be even more confusing because it breaks how this function is structured at this point. I could still change the comment for this GL feature block to include "Stencil write mask" (which would also be confusing because there is an extra block below for that too).
(I think this entire function should be removed in the future and some state setters / getters in the sync functions should keep track of used registers instead which make this way clearer and easier to maintain)

As said in the original PR description I noticed that `depth_format` should update stencil test (which is an existing bug in `master`, not something for the feature I implement) as the sync uses that reg (so this line should have been in the code in the first place - also without the space).

However I also just noticed that `raw_op` doesn't influence the `StencilWriteMask` so I'll have to group these cases differently (stencil test only: { `raw_op`, `depth_format`} / stencil test and stencil write mask: { `raw_func` }). That is indeed a "bug" in code for the new feature.

JayFoxRox:
All cleaned up and rebased (Hopefully I didn't add any new bugs..).
I addressed all comments.

(I also noticed that this will be incompatible with #1634 - I'll just fix whichever is merged later)

bunnei:
I tested pretty much all of  my games on this as well to make sure there weren't any regressions.

PR #1625, opened 2016-04-02
Rasterizer: Allow all blend factors for alpha blend-func

JayFoxRox:
When using the SW-rasterizer Crash Time 3D spams the log with:
`HW.GPU <Critical> video_core/rasterizer.cpp:operator():1005: Unknown alpha blend factor 4`

This PR attempts to cause the same behaviour as with the HW renderer ([GL man page](https://www.opengl.org/sdk/docs/man/html/glBlendFuncSeparate.xhtml)) by implementing all remaining modes for the alpha factor.

Essentially this means:
* Color factor for R,G,B: R,G,B
* Alpha factor for R,G,B: A,A,A
* Color factor for A: A = same as equivalent alpha factor \[missing from the SW-Rast. in `master`\]
* Alpha factor for A: A

~~This was not HW-tested in any way.~~ The only commerical game I tested this on is Crash Time 3D.
~~\- [Subv suggested this is not what hardware does](https://github.com/citra-emu/citra/issues/1039#issuecomment-131869274) but he also talks about DestSource which doesn't even exist. I think this should be tagged `needs-testing` instead because clearly the mode is being used.~~

### DIFF
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -923,66 +923,34 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             if (output_merger.alphablend_enable) {
                 auto params = output_merger.alpha_blending;
 
-                auto LookupFactorRGB = [&](Regs::BlendFactor factor) -> Math::Vec3<u8> {
-                    switch (factor) {
-                    case Regs::BlendFactor::Zero :
-                        return Math::Vec3<u8>(0, 0, 0);
+                auto LookupFactor = [&](unsigned channel, Regs::BlendFactor factor) -> u8 {
+                    DEBUG_ASSERT(channel < 4);
 
-                    case Regs::BlendFactor::One :
-                        return Math::Vec3<u8>(255, 255, 255);
+                    const Math::Vec4<u8> blend_const = {
+                        static_cast<u8>(output_merger.blend_const.r),
+                        static_cast<u8>(output_merger.blend_const.g),
+                        static_cast<u8>(output_merger.blend_const.b),
+                        static_cast<u8>(output_merger.blend_const.a)
+                    };
 
-                    case Regs::BlendFactor::SourceColor:
-                        return combiner_output.rgb();
-
-                    case Regs::BlendFactor::OneMinusSourceColor:
-                        return Math::Vec3<u8>(255 - combiner_output.r(), 255 - combiner_output.g(), 255 - combiner_output.b());
-
-                    case Regs::BlendFactor::DestColor:
-                        return dest.rgb();
-
-                    case Regs::BlendFactor::OneMinusDestColor:
-                        return Math::Vec3<u8>(255 - dest.r(), 255 - dest.g(), 255 - dest.b());
-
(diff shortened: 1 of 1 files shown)


JayFoxRox:
Aside from some off-by-ones (not caused by this PR): confirmed using [HW-Test](https://github.com/JayFoxRox/3ds-tests/tree/master/alpha-blend) ([Results](https://docs.google.com/spreadsheets/d/1iieAbRtNkToFuvD15Au3i8FzKm8tVT63UdLFCSr7WRE/edit#gid=1098575531) by some anonymous animals which are probably @yuriks and @LokiX)

The test also hints that mode 15 is the same as `SRC_COLOR` (which is the same as disabling blending).
\- I did not implement this yet but would do so if requested.

JayFoxRox:
I've force pushed a new version which fallbacks to not blending  instead of returning "{}" (though that shouldn't be "reachable" with the current `default:` case).
Also added `BlendFactor::SourceAlphaSaturate` which is missing from master. I've implemented it according to GL spec and the results look good (HW renderer already implemented it the GL way).

src/video_core/rasterizer.cpp
```diff
@@ -923,92 +923,79 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             if (output_merger.alphablend_enable) {
                 auto params = output_merger.alpha_blending;
 
-                auto LookupFactorRGB = [&](Regs::BlendFactor factor) -> Math::Vec3<u8> {
+                auto LookupFactor = [&](unsigned channel, Regs::BlendFactor factor) -> u8 {
+                    ASSERT_MSG(channel < 4, "Channel index is too high");
```
yuriks: Don't think this needs a message, and can be a DEBUG_ASSERT too.

src/video_core/rasterizer.cpp
```diff
@@ -923,92 +923,79 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
                         break;
                     }
 
-                    return {};
+                    return combiner_output[channel];
                 };
 
+                auto LookupFactorRGB = [&](Regs::BlendFactor factor) -> Math::Vec3<u8> {
+                    return Math::Vec3<u8>(LookupFactor(0, factor), LookupFactor(1, factor), LookupFactor(2, factor));
+                };
```
yuriks: This is only used twice right below, I think I'd just skip the lambda since it doesn't add much.

src/video_core/rasterizer.cpp
```diff
@@ -923,92 +923,79 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
-                        return 255 - output_merger.blend_const.a;
-
-                    default:
-                        LOG_CRITICAL(HW_GPU, "Unknown alpha blend factor %x", factor);
-                        UNIMPLEMENTED();
-                        break;
-                    }
-
-                    return {};
+                    return LookupFactor(3, factor);
```
yuriks: Double so for this one, which is really just one function call.

yuriks:
LGTM after comments

bunnei:
LGTM

PR #1630, opened 2016-04-02
ARM:Disassembler Better differentiate TEQ from MSR

Zaneo:
Reference: ARM DDI 0100I which applies to ARMv4, ARMv4T, ARMv5T, (ARMv5TExP), ARMv5TE, ARMv5TEJ, and ARMv6

TEQ is defined on page 116
MSR is defined on page 226

As defined by the document to differentiate MSR from TEQ, bits 15:12 should be one, bits 11:8 should be zero, and bits 7:4 are zero. (The specification indicates should be as different than fixed).


src/core/arm/disassembler/arm_disasm.cpp
```diff
@@ -1500,6 +1500,9 @@ Opcode ARM_Disasm::DecodeALU(u32 insn) {
     u8 is_immed = (insn >> 25) & 0x1;
     u8 opcode = (insn >> 21) & 0xf;
     u8 bit_s = (insn >> 20) & 1;
+    u8 msr_15_12 = (insn >> 12) & 0xF;
+    u8 msr_11_8 = (insn >> 8) & 0xF;
+    u8 msr_7_4 = (insn >> 4) & 0xF;
```
JayFoxRox: There is a BITS() macro used elsewhere in this file. I think it looks cleaner so more recent code could probably use it. (maybe another PR though?)

I'd also give them more generic names (bits_15_to_12 maybe?) or more specific names: (msr_sbo etc.) (However, I'll leave that decision to somebody else. Personally I prefer more generic names.)
Maybe even drop the whole bit extracting and just mask all bits (in the case itself). At least it's clear that those bits will be used for instruction matching then.

src/core/arm/disassembler/arm_disasm.cpp
```diff
@@ -1529,9 +1532,9 @@ Opcode ARM_Disasm::DecodeALU(u32 insn) {
                 return OP_TST;
             return OP_MRS;
         case 0x9:
-            if (bit_s)
```
lioncash: The S bit is always set for TEQ and unset for MSR, so this is actually correct.

Zaneo:
Lioncash is correct, I was incorrectly reading specification

PR #1634, opened 2016-04-02
Rasterizer: Don't draw outside of framebuffer (re-open)

### LINKED ISSUE #1633: Gummy Bears Magical Medallion - Crash with SW-Rasterizer
Tested using master fc6f898e0bab7446df330bca3b0807e871520d32

Fixed by closed-no-merge #1518 . SW-Rasterizer crashes in logos + menu + ingame at `GetDepth`
Needs further investigation.

HW renderer works fine for logos + menus.
HW Renderer only draws sky for me (Different results for somebody else: #1532 ).

JayFoxRox:
I'm re-opening this one as a new PR because apparently GitHub can't re-open force pushed PRs.. Find the original PR here #1518 

This fixes #1633 and probably more.
- I've changed the commit for better performance and tried to make the code a bit cleaner.
- Now uses s16 instead of u16 to move viewport near center of our integer range
  (u16 came from PPSSPP but the PSP uses a 4096 x 4096 coordinate space and the viewport is usually located near the center (2048, 2048) whereas 3DS viewport offsets are usually near (0,0). This meant the slightest negative offset would underflow the rasterizer coords)

Not much testing was done with `s16` but the math makes sense to me. I also changed all `>> 4` to `/ 16` because the C spec (didn't actually verify with C++) says shifting negative numbers is implementation defined.
(The current framebuffer check makes it impossible to have negative x or y in the rasterizer loop but I still thought it would be a good idea to keep the same coordinate system (and types))

As the framebuffer ownership test required `regs.framebuffer` I decided to put it in a variable (like SetDepth etc.) and also changed the rasterizer function elsewhere to reflect this. Tell me if I should leave the old code instead or even remove the variable altogether (from the new code).

### DIFF
--- a/src/video_core/pica_types.h
+++ b/src/video_core/pica_types.h
@@ -5,11 +5,153 @@
 #pragma once
 
 #include <cmath>
+#include <cstdint>
 #include <cstring>
 
 #include "common/common_types.h"
 
 namespace Pica {
+
+/**
+ * Class for storing and operating on Fixed point types using:
+ *
+ * - 12 integer bits
+ * - 4 fraction bits
+ * - Negative values in two's complement
+ *
+ * @todo Verify on HW where and how this is used.
+ * @todo Create a template for more fixed point types and use this elsewhere too.
+ */
+class Fix12P4 final {
+public:
+    constexpr Fix12P4() = default;
+    constexpr Fix12P4(const Fix12P4&) = default;
+    constexpr Fix12P4(Fix12P4&&) = default;
+    constexpr explicit Fix12P4(int16_t raw) noexcept : value(raw) {}
+
+    Fix12P4& operator=(const Fix12P4&) = default;
+    Fix12P4& operator=(Fix12P4&&) = default;
+
+    static constexpr Fix12P4 FromInt(int16_t int_val, uint16_t frac_val = 0) noexcept {
+        return Fix12P4{static_cast<int16_t>(((int_val * 16) & IntMask()) | (frac_val & FracMask()))};
+    }
+
+    static Fix12P4 FromFloat(float float_val) noexcept {
+        return Fix12P4{static_cast<int16_t>(::round(float_val * 16.0f))};
+    }
+
+    static constexpr Fix12P4 Zero() noexcept {
+        return Fix12P4{static_cast<int16_t>(0)};
+    }
+
+    static constexpr int16_t FracMask() noexcept {
+        return static_cast<int16_t>(0xF);
+    }
+
+    static constexpr int16_t IntMask() noexcept {
+        return static_cast<int16_t>(~0xF);
+    }
+
+    constexpr int16_t Int() const noexcept {
+        return static_cast<int16_t>((value & IntMask()) / 16);
+    }
+
+    constexpr uint16_t Frac() const noexcept {
+        return static_cast<uint16_t>(value & FracMask());
+    }
+
+    constexpr Fix12P4 Ceil() const noexcept {
+        return Fix12P4{static_cast<int16_t>(value + FracMask())}.Floor();
+    }
+
+    constexpr Fix12P4 Floor() const noexcept {
+        return Fix12P4{static_cast<int16_t>(value & IntMask())};
+    }
+
+    constexpr explicit operator int16_t() const noexcept {
+        return value;
+    }
+
+    constexpr Fix12P4 operator+() const noexcept {
+        return Fix12P4{static_cast<int16_t>(+value)};
+    }
+
+    constexpr Fix12P4 operator-() const noexcept {
+        return Fix12P4{static_cast<int16_t>(-value)};
+    }
+
+    constexpr Fix12P4 operator+(const Fix12P4& other) const noexcept {
+        return Fix12P4{static_cast<int16_t>(value + other.value)};
+    }
+
+    constexpr Fix12P4 operator-(const Fix12P4& other) const noexcept {
+        return Fix12P4{static_cast<int16_t>(value - other.value)};
+    }
+
+    constexpr Fix12P4 operator*(const Fix12P4& other) const noexcept {
+        return Fix12P4{Multiply(value, other.value)};
+    }
+
+    constexpr Fix12P4 operator/(const Fix12P4& other) const noexcept {
+        return Fix12P4{Divide(value, other.value)};
+    }
+
+    Fix12P4& operator+=(const Fix12P4& other) noexcept {
+        value += other.value;
+        return *this;
+    }
+
+    Fix12P4& operator-=(const Fix12P4& other) noexcept {
+        value -= other.value;
+        return *this;
+    }
+
+    Fix12P4& operator*=(const Fix12P4& other) noexcept {
+        value = Multiply(value, other.value);
+        return *this;
+    }
+
+    Fix12P4& operator/=(const Fix12P4& other) noexcept {
+        value = Divide(value, other.value);
+        return *this;
+    }
+
+    friend constexpr bool operator<(const Fix12P4& left, const Fix12P4& right) noexcept {
+        return left.value < right.value;
+    }
+
+    friend constexpr bool operator>(const Fix12P4& left, const Fix12P4& right) noexcept {
+        return operator<(right, left);
+    }
+
+    friend constexpr bool operator>=(const Fix12P4& left, const Fix12P4& right) noexcept {
+        return !operator<(left, right);
+    }
+
+    friend constexpr bool operator<=(const Fix12P4& left, const Fix12P4& right) noexcept {
+        return !operator<(right, left);
+    }
+
+    friend constexpr bool operator==(const Fix12P4& left, const Fix12P4& right) noexcept {
+        return left.value == right.value;
+    }
+
+    friend constexpr bool operator!=(const Fix12P4& left, const Fix12P4& right) noexcept {
+        return !operator==(left, right);
+    }
+
+private:
+    static constexpr int16_t Multiply(int16_t left, int16_t right) noexcept {
+        return static_cast<int16_t>((left * right) / 16);
+    }
+
+    static constexpr int16_t Divide(int16_t left, int16_t right) noexcept {
+        return static_cast<int16_t>((left * 16) / right);
+    }
+
+    int16_t value = 0;
+};
+
 
 /**
  * Template class for converting arbitrary Pica float types to IEEE 754 32-bit single-precision
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -257,73 +257,6 @@ static u8 PerformStencilAction(Regs::StencilAction action, u8 old_stencil, u8 re
     }
 }
 
-// NOTE: Assuming that rasterizer coordinates are 12.4 fixed-point values
-struct Fix12P4 {
-    Fix12P4() {}
-
-    static Fix12P4 FromRaw(s16 val) {
-        Fix12P4 res;
-        res.val = val;
-        return res;
-    }
-
-    static Fix12P4 FromInt(s16 intVal, u16 fracVal = 0) {
-        return FromRaw(static_cast<s16>(((intVal * 16) & IntMask()) | (fracVal & FracMask())));
-    }
-
-    static Fix12P4 FromFloat(float fltVal) {
-        return FromRaw(static_cast<s16>(round(fltVal * 16.0f)));
-    }
-
-    static Fix12P4 Zero() {
-        return FromRaw(0);
-    }
-
-    static u16 FracMask() { return 0xF; }
-    static u16 IntMask() { return (u16)~0xF; }
-
-    s16 Int() const {
-        return static_cast<s16>((val & IntMask()) / 16);
-    }
-
-    u16 Frac() const {
-        return static_cast<u16>(val & FracMask());
-    }
-
-    Fix12P4 Ceil() const {
-        return FromRaw(static_cast<s16>(val + FracMask())).Floor();
-    }
-
-    Fix12P4 Floor() const {
-        return FromRaw(static_cast<s16>(val & IntMask()));
-    }
-
-    operator s16() const {
-        return val;
-    }
-
-    bool operator < (const Fix12P4& oth) const {
-        return (s16)*this < (s16)oth;
-    }
-
-    Fix12P4 operator + (const Fix12P4& oth) const {
-        return FromRaw(val + oth.val);
-    }
-
-    Fix12P4 operator / (const Fix12P4& oth) const {
-        return FromRaw(static_cast<s16>((int)val * 16 / (int)oth.val));
-    }
-
-    Fix12P4& operator += (const Fix12P4& oth) {
-        val += oth.val;
-        return *this;
-    }
-
-
-private:
-    s16 val;
-};
-
 /**
  * Calculate signed area of the triangle spanned by the three argument vertices.
  * The sign denotes an orientation.
@@ -333,8 +266,12 @@ private:
 static int SignedArea (const Math::Vec2<Fix12P4>& vtx1,
                        const Math::Vec2<Fix12P4>& vtx2,
                        const Math::Vec2<Fix12P4>& vtx3) {
-    const auto vec1 = Math::MakeVec(vtx2 - vtx1, 0);
-    const auto vec2 = Math::MakeVec(vtx3 - vtx1, 0);
+    const auto vec1 = Math::MakeVec<int>(static_cast<s16>(vtx2.x) - static_cast<s16>(vtx1.x),
+                                         static_cast<s16>(vtx2.y) - static_cast<s16>(vtx1.y),
+                                         0);
+    const auto vec2 = Math::MakeVec<int>(static_cast<s16>(vtx3.x) - static_cast<s16>(vtx1.x),
+                                         static_cast<s16>(vtx3.y) - static_cast<s16>(vtx1.y),
+                                         0);
     // TODO: There is a very small chance this will overflow for sizeof(int) == 4
     return Math::Cross(vec1, vec2).z;
 };
@@ -354,6 +291,9 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
     const auto& framebuffer = regs.framebuffer;
     const auto& output_merger = regs.output_merger;
     MICROPROFILE_SCOPE(GPU_Rasterization);
+
+    // NOTE: Assuming that rasterizer coordinates are signed 12.4 fixed-point values.
+    //       This type is what PSP uses, we haven't tested what the 3DS uses.
 
     // vertex positions in rasterizer coordinates
     static auto FloatToFix = [](float24 flt) {
@@ -420,7 +360,11 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
         } else {
             // check if vertex is on our left => right side
             // TODO: Not sure how likely this is to overflow
-            return (int)vtx.x < (int)line1.x + ((int)line2.x - (int)line1.x) * ((int)vtx.y - (int)line1.y) / ((int)line2.y - (int)line1.y);
+            int line_dx = static_cast<s16>(line2.x) - static_cast<s16>(line1.x);
+            int line_dy = static_cast<s16>(line2.y) - static_cast<s16>(line1.y);
+            int line_vtx_dx = static_cast<s16>(vtx.x) - static_cast<s16>(line1.x);
+            int line_vtx_dy = static_cast<s16>(vtx.y) - static_cast<s16>(line1.y);
+            return line_vtx_dx < line_vtx_dy * line_dx / line_dy;
         }
     };
     int bias0 = IsRightSideOrFlatBottomEdge(vtxpos[0].xy(), vtxpos[1].xy(), vtxpos[2].xy()) ? -1 : 0;
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -260,21 +260,60 @@ static u8 PerformStencilAction(Regs::StencilAction action, u8 old_stencil, u8 re
 // NOTE: Assuming that rasterizer coordinates are 12.4 fixed-point values
 struct Fix12P4 {
     Fix12P4() {}
-    Fix12P4(u16 val) : val(val) {}
+
+    static Fix12P4 FromRaw(s16 val) {
+        Fix12P4 res;
+        res.val = val;
+        return res;
+    }
+
+    static Fix12P4 FromInt(s16 intVal, u16 fracVal = 0) {
+        return FromRaw(static_cast<s16>(((intVal * 16) & IntMask()) | (fracVal & FracMask())));
+    }
+
+    static Fix12P4 FromFloat(float fltVal) {
+        return FromRaw(static_cast<s16>(round(fltVal * 16.0f)));
+    }
+
+    static Fix12P4 Zero() {
+        return FromRaw(0);
+    }
 
     static u16 FracMask() { return 0xF; }
     static u16 IntMask() { return (u16)~0xF; }
 
-    operator u16() const {
+    s16 Int() const {
+        return static_cast<s16>((val & IntMask()) / 16);
+    }
+
+    u16 Frac() const {
+        return static_cast<u16>(val & FracMask());
+    }
+
+    operator s16() const {
         return val;
     }
 
     bool operator < (const Fix12P4& oth) const {
-        return (u16)*this < (u16)oth;
+        return (s16)*this < (s16)oth;
     }
 
+    Fix12P4 operator + (const Fix12P4& oth) const {
+        return FromRaw(val + oth.val);
+    }
+
+    Fix12P4 operator / (const Fix12P4& oth) const {
+        return FromRaw(static_cast<s16>((int)val * 16 / (int)oth.val));
+    }
+
+    Fix12P4& operator += (const Fix12P4& oth) {
+        val += oth.val;
+        return *this;
+    }
+
+
 private:
-    u16 val;
+    s16 val;
 };
 
 /**
@@ -308,9 +347,9 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
 
     // vertex positions in rasterizer coordinates
     static auto FloatToFix = [](float24 flt) {
-        // TODO: Rounding here is necessary to prevent garbage pixels at
+        // TODO: Rounding in Fix12P4::FromFloat is necessary to prevent garbage pixels at
         //       triangle borders. Is it that the correct solution, though?
-        return Fix12P4(static_cast<unsigned short>(round(flt.ToFloat32() * 16.0f)));
+        return Fix12P4::FromFloat(flt.ToFloat32());
     };
     static auto ScreenToRasterizerCoordinates = [](const Math::Vec3<float24>& vec) {
         return Math::Vec3<Fix12P4>{FloatToFix(vec.x), FloatToFix(vec.y), FloatToFix(vec.z)};
@@ -338,16 +377,19 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             return;
     }
 
-    // TODO: Proper scissor rect test!
-    u16 min_x = std::min({vtxpos[0].x, vtxpos[1].x, vtxpos[2].x});
-    u16 min_y = std::min({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
-    u16 max_x = std::max({vtxpos[0].x, vtxpos[1].x, vtxpos[2].x});
-    u16 max_y = std::max({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
+    // Get bounding box for triangle
+    Fix12P4 min_x = std::min({vtxpos[0].x, vtxpos[1].x, vtxpos[2].x});
+    Fix12P4 min_y = std::min({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
+    Fix12P4 max_x = std::max({vtxpos[0].x, vtxpos[1].x, vtxpos[2].x});
+    Fix12P4 max_y = std::max({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
 
-    min_x &= Fix12P4::IntMask();
-    min_y &= Fix12P4::IntMask();
-    max_x = ((max_x + Fix12P4::FracMask()) & Fix12P4::IntMask());
-    max_y = ((max_y + Fix12P4::FracMask()) & Fix12P4::IntMask());
+    // TODO: Proper scissor rect test!
+
+    Fix12P4 frac = Fix12P4::FromInt(0, Fix12P4::FracMask());
+    min_x = Fix12P4::FromInt(min_x.Int(), 0);
+    min_y = Fix12P4::FromInt(min_y.Int(), 0);
+    max_x = Fix12P4::FromInt((max_x + frac).Int());
+    max_y = Fix12P4::FromInt((max_y + frac).Int());
 
     // Triangle filling rules: Pixels on the right-sided edge or on flat bottom edges are not
     // drawn. Pixels on any other triangle border are drawn. This is implemented with three bias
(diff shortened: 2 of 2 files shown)


src/video_core/rasterizer.cpp
```diff
@@ -335,17 +336,25 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
     min_x &= Fix12P4::IntMask();
     min_y &= Fix12P4::IntMask();
     max_x = ((max_x + Fix12P4::FracMask()) & Fix12P4::IntMask());
     max_y = ((max_y + Fix12P4::FracMask()) & Fix12P4::IntMask());
 
+    // Keep pixels inside framebuffer
+    min_x = std::max({min_x, (s16)0});
+    min_y = std::max({min_y, (s16)0});
+    max_x = std::min({max_x, (s16)(framebuffer.GetWidth() << 4)});
+    max_y = std::min({max_y, (s16)(framebuffer.GetHeight() << 4)});
```
JayFoxRox: I should probably do `* 16` here so everyone gets it

bunnei: Yes please

bunnei:
Other than the one comment, LGTM

JamePeng:
LGTM

src/video_core/rasterizer.cpp
```diff
@@ -335,17 +336,25 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
     min_x &= Fix12P4::IntMask();
     min_y &= Fix12P4::IntMask();
     max_x = ((max_x + Fix12P4::FracMask()) & Fix12P4::IntMask());
     max_y = ((max_y + Fix12P4::FracMask()) & Fix12P4::IntMask());
 
+    // Keep pixels inside framebuffer
+    min_x = std::max({min_x, (s16)0});
+    min_y = std::max({min_y, (s16)0});
+    max_x = std::min({max_x, (s16)(framebuffer.GetWidth() * 16)});
+    max_y = std::min({max_y, (s16)(framebuffer.GetHeight() * 16)});
```
yuriks: Why are you using the initializer_list version of min/max here instead of the regular 2-param ones?

src/video_core/rasterizer.cpp
```diff
@@ -809,11 +820,11 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
 
             auto UpdateStencil = [stencil_test, x, y, &old_stencil](Pica::Regs::StencilAction action) {
                 u8 new_stencil = PerformStencilAction(action, old_stencil, stencil_test.reference_value);
-                SetStencil(x >> 4, y >> 4, (new_stencil & stencil_test.write_mask) | (old_stencil & ~stencil_test.write_mask));
+                SetStencil(x / 16, y / 16, (new_stencil & stencil_test.write_mask) | (old_stencil & ~stencil_test.write_mask));
```
yuriks: Would be good to make this a local constant (with a *short* variable name) in case we decide to increase/decrease the subpixel precision later.

src/video_core/rasterizer.cpp
JayFoxRox: Will be changed. Copied the min/max from above and didn't even question it ;)

JayFoxRox:
I've shifted some stuff around and it's using a more "complete" Fix12P4 class which also gets rid of the nasty shifts.

I had some rebase / reorder conflicts so it'd be nice if you could re-test / review this again.

src/video_core/rasterizer.cpp
```diff
@@ -255,21 +255,60 @@ static u8 PerformStencilAction(Regs::StencilAction action, u8 old_stencil, u8 re
...
     static u16 FracMask() { return 0xF; }
     static u16 IntMask() { return (u16)~0xF; }
 
-    operator u16() const {
+    s16 Int() const {
+        return (val & IntMask()) / 16;
+    }
+
+    u16 Frac() const {
+        return val & FracMask();
```
lioncash: This needs a static_cast, otherwise you'll get truncation warnings (the & promotes to int)

src/video_core/rasterizer.cpp
```diff
@@ -255,21 +255,60 @@ static u8 PerformStencilAction(Regs::StencilAction action, u8 old_stencil, u8 re
...
+    static Fix12P4 Zero() {
+        return FromRaw(0);
+    }
 
     static u16 FracMask() { return 0xF; }
     static u16 IntMask() { return (u16)~0xF; }
 
-    operator u16() const {
+    s16 Int() const {
+        return (val & IntMask()) / 16;
```
lioncash: Ditto wrt my comment below

src/video_core/rasterizer.cpp
```diff
@@ -335,16 +375,25 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
 
-    min_x &= Fix12P4::IntMask();
-    min_y &= Fix12P4::IntMask();
-    max_x = ((max_x + Fix12P4::FracMask()) & Fix12P4::IntMask());
-    max_y = ((max_y + Fix12P4::FracMask()) & Fix12P4::IntMask());
+    Fix12P4 frac = Fix12P4::FromInt(0, Fix12P4::FracMask());
+    min_x = Fix12P4::FromInt(min_x.Int(), 0);
+    min_y = Fix12P4::FromInt(min_y.Int(), 0);
+    max_x = Fix12P4::FromInt((max_x + frac).Int());
+    max_y = Fix12P4::FromInt((max_y + frac).Int());
```
yuriks: I don't understand what's the purpose of doing this rounding here.

src/video_core/rasterizer.cpp
```diff
@@ -372,13 +421,17 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
     auto textures = regs.GetTextures();
     auto tev_stages = regs.GetTevStages();
 
-    bool stencil_action_enable = g_state.regs.output_merger.stencil_test.enable && g_state.regs.framebuffer.depth_format == Regs::DepthFormat::D24S8;
+    bool stencil_action_enable = g_state.regs.output_merger.stencil_test.enable && framebuffer.depth_format == Regs::DepthFormat::D24S8;
```
yuriks: Can use just `regs` instead of `g_state.regs`, same for other nearby instance here.

src/video_core/rasterizer.cpp
```diff
@@ -335,16 +375,25 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
-    u16 max_x = std::max({vtxpos[0].x, vtxpos[1].x, vtxpos[2].x});
-    u16 max_y = std::max({vtxpos[0].y, vtxpos[1].y, vtxpos[2].y});
 
-    min_x &= Fix12P4::IntMask();
-    min_y &= Fix12P4::IntMask();
-    max_x = ((max_x + Fix12P4::FracMask()) & Fix12P4::IntMask());
-    max_y = ((max_y + Fix12P4::FracMask()) & Fix12P4::IntMask());
+    Fix12P4 frac = Fix12P4::FromInt(0, Fix12P4::FracMask());
+    min_x = Fix12P4::FromInt(min_x.Int(), 0);
+    min_y = Fix12P4::FromInt(min_y.Int(), 0);
```
yuriks: The 0 is an implied default.

yuriks: I think it would be clearer to add `Floor` and `Ceil` members to `Fix12P4` instead of doing these directly here. The extracting and repacking of the values is noisy and directly manipulating the raw value would be better.

src/video_core/rasterizer.cpp
```diff
@@ -372,13 +421,17 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
     // Enter rasterization loop, starting at the center of the topleft bounding box corner.
     // TODO: Not sure if looping through x first might be faster
-    for (u16 y = min_y + 8; y < max_y; y += 0x10) {
-        for (u16 x = min_x + 8; x < max_x; x += 0x10) {
+    Fix12P4 pixel = Fix12P4::FromInt(1);
+    Fix12P4 half_pixel = pixel / Fix12P4::FromInt(2);
+    for (Fix12P4 y = min_y + half_pixel; y < max_y; y += pixel) {
+        for (Fix12P4 x = min_x + half_pixel; x < max_x; x += pixel) {
+
+            ASSERT_MSG(x.Int() < framebuffer.GetWidth() && y.Int() < framebuffer.GetHeight(), "Pixel not inside framebuffer");
```
yuriks: Don't think this assert is necessary. If you feel strongly about keeping it then at least cache the fb width/height in a variable because unpacking them from the register every time is slow. (And compilers don't optimize it away because of aliasing.)

src/video_core/rasterizer.cpp
```diff
@@ -328,8 +370,8 @@ struct Fix12P4 {
 static int SignedArea (const Math::Vec2<Fix12P4>& vtx1,
                        const Math::Vec2<Fix12P4>& vtx2,
                        const Math::Vec2<Fix12P4>& vtx3) {
-    const auto vec1 = Math::MakeVec(vtx2 - vtx1, 0);
-    const auto vec2 = Math::MakeVec(vtx3 - vtx1, 0);
+    const auto vec1 = Math::MakeVec<int>(vtx2.x - vtx1.x, vtx2.y - vtx1.y, 0);
+    const auto vec2 = Math::MakeVec<int>(vtx3.x - vtx1.x, vtx3.y - vtx1.y, 0);
```
JayFoxRox: Doesn't work with the old code and the new "operator -". I wasn't sure how to solve this so I used this (which is really ugly).
- Ideas?

src/video_core/rasterizer.cpp
```diff
@@ -297,23 +297,65 @@ struct Fix12P4 {
...
+        return *this;
     }
 
     Fix12P4& operator += (const Fix12P4& oth) {
         val += oth.val;
         return *this;
     }
 
+    Fix12P4& operator -= (const Fix12P4& oth) {
+        val -= oth.val;
```
JayFoxRox: Does this integer truncate? If so, how to improve this?

src/video_core/rasterizer.cpp
```diff
@@ -297,23 +297,65 @@ struct Fix12P4 {
...
+        return FromRaw(static_cast<s16>(val - oth.val));
+    }
+
+    Fix12P4& operator *= (const Fix12P4& oth) {
+        val = (*this * oth).val;
+        return *this;
+    }
+
+    Fix12P4& operator /= (const Fix12P4& oth) {
+        val = (*this / oth).val;
```
JayFoxRox: I don't like this, but didn't want to copy the div code (same argument for `operator *=`).  Ideas?

lioncash: `val /= oth.val;` They're already defined for built-in types :)

JayFoxRox: That wouldn't be the correct operation. Fixed point division would be: `val = val * 16 / oth.val` which requires a `static_cast` again. (So the same as `operator /` which is why I decided to use it instead of copying the code)

lioncash: Note that it creates a new unnecessary copy that way.

src/video_core/rasterizer.cpp
```diff
@@ -255,21 +255,110 @@ static u8 PerformStencilAction(Regs::StencilAction action, u8 old_stencil, u8 re
...
+        return val == oth.val;
+    }
+
+    bool operator != (const Fix12P4& oth) const {
+        return val != oth.val;
     }
 
 private:
-    u16 val;
+    s16 val;
```
lioncash: This should likely be `s16 val = 0` or some other preferable default. If you declare a new instance of the class and then do arithmetic on it, it'll be operating on uninitialized memory.

src/video_core/rasterizer.cpp
lioncash: fwiw, this is how I'd have implemented it, assuming it was specifically meant to not be generified, if interested (the class can actually be adapted to handle any fixed-point length, but that's not relevant): [code](https://gist.github.com/lioncash/f47ecedaba714b250cad67997b2f2583)

JayFoxRox:
I've rebased to master and replaced a lot of Fix12P4 with code from [lioncashs fixed point class](https://gist.github.com/lioncash/f47ecedaba714b250cad67997b2f2583) (thanks!).

I had some minor rebase problems and also had to some code for the Fix12P4 class (casts to int etc.).
Tested fine in Zelda OoT, Tappingo and Gummy Bears Magical Medallion and found no issues.

I've added some more casts than there are in the current code (so more int math than s16 math). We are not likely to overflow in most of those places, but better safe than sorry, right?

I didn't want to add the numeric_limits stuff yet because I felt it would take some time to look up the precise definition for everything as I never worked with it + I don't think anybody would use it at this point.
Also we still don't know if Fix12P4 is actually necessary or if we need another type (for the rasterizer coords).
However, whenever we turn Fix12P4 into a template we should take this into consideration again.

---

Review and test (SW renderer) please :)

bunnei:
@JayFoxRox can you please post an update here on the status of this change?

JayFoxRox:
Waiting for reviews and tests, code should be complete, only needs to be squashed (which I'll do after reviews).

src/video_core/rasterizer.cpp
```diff
@@ -304,13 +288,18 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
                                     bool reversed = false)
 {
     const auto& regs = g_state.regs;
+    const auto& framebuffer = regs.framebuffer;
+    const auto& output_merger = regs.output_merger;
     MICROPROFILE_SCOPE(GPU_Rasterization);
 
+    // NOTE: Assuming that rasterizer coordinates are signed 12.4 fixed-point values.
+    //       This type is what PSP uses, we haven't tested what the 3DS uses.
```
Subv: As an aside, how can we know what the 3DS uses internally?

JayFoxRox:
Refactored a bit so it's less code to review.

JayFoxRox:
@citra-emu Bump..

LFsWang:
@JayFoxRox C++ Std 5.8.3 Shift operators
E1 >> E2
If E1 has a signed type and a negative value, the resulting value is implementation-defined.

lioncash:
@LFsWang What part of the code are you referencing this about?

LFsWang:
I simply answer this statement.
> Not much testing was done with s16 but the math makes sense to me. I also changed all >> 4 to / 16 because the C spec (didn't actually verify with C++) says shifting negative numbers is implementation defined.

src/video_core/rasterizer.cpp
```diff
@@ -375,13 +376,15 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
+    bool stencil_action_enable = output_merger.stencil_test.enable && framebuffer.depth_format == Regs::DepthFormat::D24S8;
+    const auto stencil_test = output_merger.stencil_test;
 
     // Enter rasterization loop, starting at the center of the topleft bounding box corner.
     // TODO: Not sure if looping through x first might be faster
-    for (u16 y = min_y + 8; y < max_y; y += 0x10) {
-        for (u16 x = min_x + 8; x < max_x; x += 0x10) {
+    Fix12P4 pixel = Fix12P4::FromInt(1);
+    Fix12P4 half_pixel = pixel / Fix12P4::FromInt(2);
+    for (Fix12P4 y = min_y + half_pixel; y < max_y; y += pixel) {
```
Subv: This is much, much easier to understand than the previous code, thanks!

Subv:
LGTM, can we please get this merged already?

bunnei:
Doesn't compile for me on VS15... anyone else see similar errors?
<pre>4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(36): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(51): note: see declaration of 'Pica::Fix12P4::IntMask' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(36): error C2134: 'Pica::Fix12P4::FracMask': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(36): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(47): note: see declaration of 'Pica::Fix12P4::FracMask' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(36): error C2134: 'Pica::Fix12P4::IntMask': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(36): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(51): note: see declaration of 'Pica::Fix12P4::IntMask' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(64): error C2134: 'Pica::Fix12P4::Floor': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(36): error C2134: 'Pica::Fix12P4::FracMask': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(64): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(67): note: see declaration of 'Pica::Fix12P4::Floor' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(36): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(47): note: see declaration of 'Pica::Fix12P4::FracMask' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(92): error C2134: 'Pica::Fix12P4::Multiply': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(92): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(144): note: see declaration of 'Pica::Fix12P4::Multiply' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(64): error C2134: 'Pica::Fix12P4::Floor': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(96): error C2134: 'Pica::Fix12P4::Divide': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(64): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(67): note: see declaration of 'Pica::Fix12P4::Floor' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(96): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(148): note: see declaration of 'Pica::Fix12P4::Divide' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\primitive_assembly.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(92): error C2134: 'Pica::Fix12P4::Multiply': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(92): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(144): note: see declaration of 'Pica::Fix12P4::Multiply' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(96): error C2134: 'Pica::Fix12P4::Divide': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(96): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(148): note: see declaration of 'Pica::Fix12P4::Divide' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\clipper.cpp)
4>D:\dev\3ds\citra-bunnei\externals\nihstro\include\nihstro/shader_bytecode.h(764): warning C4800: 'unsigned int': forcing value to bool 'true' or 'false' (performance warning) (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\pica.cpp)
4>D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(36): error C2134: 'Pica::Fix12P4::IntMask': call does not result in a constant expression (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\pica.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(36): note: failure was caused by call of undefined function or one not declared 'constexpr' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\pica.cpp)
4>  D:\dev\3ds\citra-bunnei\src\video_core/pica_types.h(51): note: see declaration of 'Pica::Fix12P4::IntMask' (compiling source file D:\dev\3ds\citra-bunnei\src\video_core\pica.cpp)</pre>

src/video_core/pica_types.h
```diff
@@ -5,13 +5,155 @@
...
+        return left.value == right.value;
+    }
+
+    friend constexpr bool operator!=(const Fix12P4& left, const Fix12P4& right) noexcept {
+        return !operator==(left, right);
+    }
+
+private:
+    static constexpr int16_t Multiply(int16_t left, int16_t right) noexcept {
+        return static_cast<int16_t>((left * right) / 16);
```
yuriks: I think you could put a comment here indicating that `left * right` promotes both to int, and so overflow doesn't occur. (Or maybe cast to int32_t manually if you're worried about a platform with 16-bit ints :P)

src/video_core/pica_types.h
```diff
@@ -5,13 +5,155 @@
...
+ */
+class Fix12P4 final {
+public:
+    constexpr Fix12P4() = default;
+    constexpr Fix12P4(const Fix12P4&) = default;
+    constexpr Fix12P4(Fix12P4&&) = default;
+    constexpr explicit Fix12P4(int16_t raw) noexcept : value(raw) {}
+
+    Fix12P4& operator=(const Fix12P4&) = default;
+    Fix12P4& operator=(Fix12P4&&) = default;
```
yuriks: Why is this one necessary?

src/video_core/pica_types.h
```diff
@@ -5,13 +5,155 @@
...
+
+    static constexpr Fix12P4 FromInt(int16_t int_val, uint16_t frac_val = 0) noexcept {
```
yuriks: Passing negative values to `int_val` won't work properly, so note in a comment that you shouldn't do that and should do `-Fix12P4::FromInt(x)` instead.

src/video_core/pica_types.h
```diff
@@ -5,13 +5,155 @@
...
+        return Fix12P4{static_cast<int16_t>(((int_val * 16) & IntMask()) | (frac_val & FracMask()))};
+    }
+
+    static Fix12P4 FromFloat(float float_val) noexcept {
+        return Fix12P4{static_cast<int16_t>(::round(float_val * 16.0f))};
+    }
+
+    static constexpr Fix12P4 Zero() noexcept {
```
yuriks: Can this and the next two be constants instead of functions?

src/video_core/pica_types.h
```diff
@@ -5,13 +5,155 @@
...
+        value = Divide(value, other.value);
+        return *this;
+    }
+
+    friend constexpr bool operator<(const Fix12P4& left, const Fix12P4& right) noexcept {
+        return left.value < right.value;
+    }
+
+    friend constexpr bool operator>(const Fix12P4& left, const Fix12P4& right) noexcept {
+        return operator<(right, left);
```
yuriks: Reusing `operator<` is interesting in terms of reducing redundancy, but I think it gains us too little compared to how less obvious it is.

src/video_core/rasterizer.cpp
```diff
@@ -286,8 +266,12 @@ struct Fix12P4 {
 static int SignedArea (const Math::Vec2<Fix12P4>& vtx1,
                        const Math::Vec2<Fix12P4>& vtx2,
                        const Math::Vec2<Fix12P4>& vtx3) {
-    const auto vec1 = Math::MakeVec(vtx2 - vtx1, 0);
-    const auto vec2 = Math::MakeVec(vtx3 - vtx1, 0);
+    const auto vec1 = Math::MakeVec<int>(static_cast<s16>(vtx2.x) - static_cast<s16>(vtx1.x),
```
yuriks: Can't you subtract and then cast?

src/video_core/rasterizer.cpp
```diff
@@ -286,8 +266,12 @@ struct Fix12P4 {
...
+                                         static_cast<s16>(vtx2.y) - static_cast<s16>(vtx1.y),
+                                         0);
+    const auto vec2 = Math::MakeVec<int>(static_cast<s16>(vtx3.x) - static_cast<s16>(vtx1.x),
```
yuriks: Specifying the type kind of defeats the purpose of MakeVec, just use Vec3<int>{ ... } instead.

src/video_core/rasterizer.cpp
```diff
@@ -304,13 +288,18 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
     MICROPROFILE_SCOPE(GPU_Rasterization);
 
+    // NOTE: Assuming that rasterizer coordinates are signed 12.4 fixed-point values.
+    //       This type is what PSP uses, we haven't tested what the 3DS uses.
+
     // vertex positions in rasterizer coordinates
     static auto FloatToFix = [](float24 flt) {
-        // TODO: Rounding here is necessary to prevent garbage pixels at
+        // TODO: Rounding in Fix12P4::FromFloat is necessary to prevent garbage pixels at
         //       triangle borders. Is it that the correct solution, though?
```
yuriks: "Is that the"*

src/video_core/rasterizer.cpp
```diff
@@ -363,7 +360,11 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
         } else {
             // check if vertex is on our left => right side
             // TODO: Not sure how likely this is to overflow
-            return (int)vtx.x < (int)line1.x + ((int)line2.x - (int)line1.x) * ((int)vtx.y - (int)line1.y) / ((int)line2.y - (int)line1.y);
+            int line_dx = static_cast<s16>(line2.x) - static_cast<s16>(line1.x);
```
yuriks: Same with subtracting before casting

src/video_core/rasterizer.cpp
```diff
@@ -838,11 +840,11 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             auto UpdateStencil = [stencil_test, x, y, &old_stencil](Pica::Regs::StencilAction action) {
                 u8 new_stencil = PerformStencilAction(action, old_stencil, stencil_test.reference_value);
                 if (g_state.regs.framebuffer.allow_depth_stencil_write != 0)
-                    SetStencil(x >> 4, y >> 4, (new_stencil & stencil_test.write_mask) | (old_stencil & ~stencil_test.write_mask));
+                    SetStencil(x.Int(), y.Int(), (new_stencil & stencil_test.write_mask) | (old_stencil & ~stencil_test.write_mask));
```
yuriks: `x.Int()` and `y.Int()` are used so many times here that you could probably introduce variables for them

yuriks:
Point out if I'm missing something, but it doesn't look like you're *actually* adding any framebuffer bounds testing anymore, right? It's just a numeric overflow fix? But then I don't understand how changing from u16 to s16 would make overflow *better*, so I think I must be missing something.

src/video_core/rasterizer.cpp
```diff
@@ -338,16 +327,24 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
             return;
     }
 
+    // Get bounding box for triangle
```
yuriks: I think the scissor TODO still belongs with this block of code (since it's how it'd be implemented.) If nothing else it makes the diff slightly less noisy. :)

src/video_core/pica_types.h
JayFoxRox: Ask @lioncash who wrote this class (because he didn't like mine)

src/video_core/pica_types.h
JayFoxRox: why not? Looks fine to me. (except for the frac bits which should probably be negated here?)

src/video_core/pica_types.h
JayFoxRox: This started out as a 1:1 copy as a short modification of the existing Fix12P4 class, then there were complaints that this wasn't good code (despite only changing the data type) so I copied the Float24 class and turned it into Fix12P4. So all design decisions of Float24 / Fix12P4 have been used.
In the end @lioncash still wasn't happy with my class and wrote his own. Not sure if this addition was by him or by me, but it's what the Float24 class does probably.

---

Aside from that, yes. Could be a constant

src/video_core/pica_types.h
JayFoxRox: Exactly my thought, my class did it seperately too. @lioncash

src/video_core/rasterizer.cpp
JayFoxRox: I can't remember why this was done, but iirc there were some problems when doing it differently.
Would have to be tested again.

src/video_core/rasterizer.cpp
JayFoxRox: Can you give an example? I don't see what you mean

src/video_core/rasterizer.cpp
JayFoxRox: why? We are working with a subpixel-coordinate type here.
There won't be any perf difference and creating a variable from it will only make it possible to work with outdated values if one is not careful enough.

JayFoxRox:
As I've mentioned on IRC earlier: this PR was nothing but frustrating.
The more recent experience of trying to get GS merged wasn't very nice either so I'll take a pause from Citra for now.

I'll leave this PR opened until someone takes this over.

src/video_core/pica_types.h
lioncash: I wrote mine as an example of how I'd generally structure the class, I don't know where you got that I was completely unhappy with it. My only main complaint was that the original class only provided `operator+` and `operator*` arithmetic operators.

hrydgard:
I know it's not my project, but the amount of nit-picking here for such a simple change baffles me. There's not much that is made clearer by the introduction of the special 12.4 fixed point type either - a regular int16_t or int32_t, properly named and commented, will do the trick just fine.

JayFoxRox:
I've added this to my personal TODO list.
I'll close this one as we have too many open PRs.

Issue #1636, opened 2016-04-03
ffmpeg fails to render VP8 videos properly


linkmauve:
Probably related to #862 (which is about mpeg4 instead), there are green and magenta blocks everywhere, and a fully cyan frame once in a while on i965 (see https://bugs.freedesktop.org/show_bug.cgi?id=94148). I-frames seem rendered correctly, but all predicted ones with changes compared to the keyframe render wrong.

Here is a sample homebrew: https://linkmauve.fr/files/3Damnesic.3dsx.xz
Here is a WebM sample containing a VP8 video track, which you must put in `$XDG_DATA_HOME/citra-emu/sdmc/ending.webm`: https://linkmauve.fr/files/ending.webm

I tested it on hardware, it renders properly.

Source code is at https://github.com/Lectem/3Damnesic except I compiled ffmpeg with nothing but VP8 support.

linkmauve:
When ffmpeg is compiled with --disable-asm, this video renders fine, so this issue comes from one of the the hand-written assembly parts.

linkmauve:
This was on ffmpeg 2.8.5, using `--enable-decoder=vp8,vorbis --enable-demuxer=matroska` to replace these two flags listed at https://github.com/Lectem/3Damnesic/blob/master/ffmpeg-configure3ds

JayFoxRox:
Just a quick update:

```
av_cold void ff_vp78dsp_init_armv6(VP8DSPContext *dsp)
{
    dsp->put_vp8_epel_pixels_tab[0][0][0] = ff_put_vp8_pixels16_armv6;
#if 0
// Broken instant
    dsp->put_vp8_epel_pixels_tab[0][0][2] = ff_put_vp8_epel16_h6_armv6;
#endif
#if 0
// Broken instant
    dsp->put_vp8_epel_pixels_tab[0][2][0] = ff_put_vp8_epel16_v6_armv6;
#endif
#if 1
// Broken instant
    dsp->put_vp8_epel_pixels_tab[0][2][2] = ff_put_vp8_epel16_h6v6_armv6;
#endif
    dsp->put_vp8_epel_pixels_tab[1][0][0] = ff_put_vp8_pixels8_armv6;
#if 0
// Broken: soon
    dsp->put_vp8_epel_pixels_tab[1][0][1] = ff_put_vp8_epel8_h4_armv6;
    dsp->put_vp8_epel_pixels_tab[1][0][2] = ff_put_vp8_epel8_h6_armv6;
[...]

PR #1638, opened 2016-04-03
OpenGL: Check for framebuffer completeness

JayFoxRox:
Work around [this mesa bug](https://bugs.freedesktop.org/show_bug.cgi?id=94148). It's also the sane thing to do anyway.

I'm not sure if ASSERT_MSG() is meant to catch errors which should be presentable to the user or wether it should be a debugging aid (absent from release builds). - Probably worth another issue.
This line however was copied from elsewhere in the same file so.. insta-merge?

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -808,6 +808,9 @@ void RasterizerOpenGL::SyncFramebuffer() {
 
         ReloadDepthBuffer();
     }
+
+    ASSERT_MSG(glCheckFramebufferStatus(GL_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE,
+               "OpenGL rasterizer framebuffer setup failed, status %X", glCheckFramebufferStatus(GL_FRAMEBUFFER));
 }
 
 void RasterizerOpenGL::SyncCullMode() {


linkmauve:
This fixes the cyan frames in #1636, and makes the Stock Pot Inn scene render in MM3D, on i965.

I’d rather use a temporary variable to do only one call in the error case, but it’s fine to do it afterwards in both places.

So LGTM.

MerryMage:
> I'm not sure if ASSERT_MSG() is meant to catch errors which should be presentable to the user or wether it should be a debugging aid (absent from release builds)

We have `DEBUG_ASSERT_MSG` for the latter.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -808,6 +808,9 @@ void RasterizerOpenGL::SyncFramebuffer() {
 
         ReloadDepthBuffer();
     }
+
+    ASSERT_MSG(glCheckFramebufferStatus(GL_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE,
```
hrydgard: glCheckFramebufferStatus should be called once during creation of a framebuffer object, not all the time.

Besides, you shouldn't modify the attachments of a framebuffer object during a frame - much better to create separate framebuffer objects for each "setup" and switch between them, they can share attachements.

PR #1640, opened 2016-04-03
CMakeFiles: Compile as C++14

lioncash:
Quite overdue

As Travis CI devs are continually dedicated to providing only the most up-to-date compiler tools, they use vitrual environments containing Mac OS X 10.9 with XCode 6.1 (LLVM 3.5) and Ubuntu Trusty with the blazingly up-to-date GCC 4.9. So we cannot use the regular `-std=c++14` switch. So we need to use `-std=c++1y` (the name of C++14 when it didn't have a concrete date).


yuriks:
LGTM

PR #1641, opened 2016-04-04
Implement actual system time

ObsidianX:
libctru: osGetTime();


src/core/hle/shared_page.cpp
```diff
@@ -15,7 +16,16 @@ SharedPageDef shared_page;
 void Init() {
     std::memset(&shared_page, 0, sizeof(shared_page));
 
+    // 3DS uses 1/1/1900 for Epoch
+    time_t plat_time = time(NULL);
```
wwylele: std::

lioncash: also nullptr should be used instead of NULL

src/core/hle/shared_page.cpp
```diff
@@ -15,7 +16,16 @@ SharedPageDef shared_page;
...
+    struct tm console_epoch = {0};
```
lioncash: `struct` can be omitted

src/core/hle/shared_page.cpp
```diff
@@ -15,7 +16,16 @@ SharedPageDef shared_page;
...
+    console_epoch.tm_mday = 1;
+    u64_le console_time = (u64_le) difftime(plat_time, mktime(&console_epoch)) * 1000L;
+
     shared_page.running_hw = 0x1; // product
+    shared_page.date_time_0 = {
+            .date_time = console_time,
```
lioncash: Designated initializers are a C99 feature which MSVC does not support.

src/core/hle/shared_page.cpp
```diff
@@ -15,7 +16,16 @@ SharedPageDef shared_page;
 void Init() {
     std::memset(&shared_page, 0, sizeof(shared_page));
 
+    // 3DS uses 1/1/1900 for Epoch
+    time_t plat_time = time(NULL);
+    struct tm console_epoch = {0};
+    console_epoch.tm_mday = 1;
+    u64_le console_time = (u64_le) difftime(plat_time, mktime(&console_epoch)) * 1000L;
```
lioncash: static_cast

Subv:
So this is a fixed time?
IMO the "current time" should be set in a configuration variable and incremented automatically with a CoreTiming event.
Also remember that there are 2 date_time structures in this memory region (I wonder why, though)

ObsidianX:
@Subv:

```
struct DATETIME_0 {
    u64 MillisecondsSinceFirstJan1900,
    u64 SystemTickLastTimeThisWasUpdated,
    ...
}
```

ctrulib calculates the time based on `MillisecondsSinceFirstJan1900 + (currentTick - SystemTickLastTimeThisWasUpdated)`.  Either the RTC sets `MillisecondsSinceFirstJan1900` once at boot or it continues to at some regular interval, with the delta calculation only being used to keep millisecond accuracy between RTC polls.  The implementation provided in this PR matches the time I get on my physical 3DS and keeps pace, though I haven't run it for a prolonged period of time to see if there is drift.

wwylele:
I am not sure if this is a MSVC bug, but `std::mktime(&console_epoch)` always return -1, which make the time calculation fail.

wwylele:
OK found the problem. the epoch for `time_t` is 1970, and `std::mktime(&console_epoch)` should result a negative value, which cannot be handled. A work-around can be set epoch to 2000:
```
time_t plat_time = std::time(nullptr);
tm console_epoch = {};
console_epoch.tm_year = 100; // 2000
u64_le console_time = (static_cast<u64_le>(std::difftime(plat_time, std::mktime(&console_epoch)))
        +3155673600L/*seconds from 1900 to 2000*/) * 1000L;
```

Also, there is an unknown entry in the shared page that affects timing, and commercial games can't work without this: 
```
struct DateTime {
    u64_le date_time;                  // 0
    u64_le update_tick;                // 8
    u32 unk;                           // 10  <- Here
    INSERT_PADDING_BYTES(0x20 - 0x14); // 14
};
```
I feed this entry with a value get from real console:
```
shared_page.date_time_0.unk = 0x0FFB0FF0;
```
With all of this, I test the game 2048, and the date time is shown and walk correctly.

yuriks:
The system periodically updates this value from some background process. Would be good to run a hwtest to figure out exactly how often.

There are two copies of the struct and a flag that selects between them. This is to make the update atomic: The process can write to one of the structs while the app uses the other one, and then flip the flag. Otherwise if the app read the struct at the same time as it is being updated, it would see partial updates.

yuriks:
Re. the epoch issue. Instead of 200 can't you just set the epoch to 1970 then (eliminating the need to call mktime entirely for the epoch) and then apply the correct offset?

wwylele:
Re. updating period. I am doing this [hwtest](https://github.com/wwylele/ctrhwtest/tree/master/sharedpage), which continues fetching and printing out the whole shared page. The system time got update **1 hour** after booting.

The update behaves like this:
Since booting, until update:
```
*(u8*)0x1FF81000 = 1
date_time_0 = all zero
date_time_1 = {(u64)date_time, (u64)update_tick, (u64) 0x000000000FFB0FF0, (u64) 0}
```
After update:
```
*(u8*)0x1FF81000 = 2
date_time_0 = {(u64)(new)date_time, (u64)(new)update_tick, (u64) 0x000000000FFB0FF0, (u64) 0xFFFFFFFFFFFFFFBC}
date_time_1 = (remain the same)
```

I am still running this hwtest to see what happens at 2 hour.

wwylele:
About how to implement system time citra, I come up with following different ways:

* Update system time each (real time) hour, as how it works on real console. This may be a good solution in long term, but it currently cause issues due to the fact that we can't emulate game at full speed. There are two more options:
  * assign `tick_to_second_coefficient` with system tick rate (268MHz), as it does on real console. This makes the system time synchronized with game itself, but generate a time "jump" each hour if we don't run the game at full speed
  * assign `tick_to_second_coefficient` with calculated tick rate on PC. This makes he system time synchronized with real time, but will fail if the emulating speed is unstable.
* Only update system time once when initializing (as this PR did) and assign `tick_to_second_coefficient` with 268MHz. This totally desynchronizes the system time with real time while synchronizes it with game.
* Instead of updating system time every hours, update it every seconds (or in even higher frequency). This makes he system time synchronized with real time, but can cause the game unstable.

ObsidianX:
Is the RTC on the 3DS a stand-alone IC that has its own tick?  If so, I wouldn't think that games would become unstable with time jumps since there isn't a guarantee that the time has ticked at the CPU's rate anyway.  I'm far from familiar with the inner hardware workings of the 3DS, so someone more qualified should make the decision, but I vote for the once per real-time hour, sync with system time approach.

As far as games triggering the "you changed the time!" error, that appears to be determined by "GetUserTimeOffset()" as described here: http://3dbrew.org/wiki/3DS_Development_Unit_Software#Config (see: Raw RTC Setting)

ObsidianX:
@yuriks:

Re: epoch

I didn't want to assume that epoch would always be 1/1/1970 in the event there's some weird platform that doesn't adhere to that.  It's highly unlikely, but I felt it was safer to know precisely what time we're starting from.  Since this PR only runs once, and the approach the hardware uses is only once per hour the extra calculations are negligible towards performance but I certainly wouldn't care if it was switched to adding a constant to fill in the 70-year gap.

yuriks:
@ObsidianX: The C std doesn't specify that time_t holds seconds either, then. In practice everyone uses POSIX time which starts in 1970-01-01.

However, you could instead use the newer C++ [chrono](http://en.cppreference.com/w/cpp/chrono) library which has explicit support for durations, etc.

yuriks:
It's interesting that the "clock rate factor" changes over time on the 3DS. I wonder if it's doing that to prevent longer-term clock skew while at the same time ensuring that the clock is monotonic across updates. That would probably be the best solution to implement here, but would require some thought to work correctly.

Alternatively, (and probably a good enough solution for now) we can get the current date at Citra startup, and from then on update the shared memory every emulated hour. This skews game time versus real time, but is simpler to implement, while still making sure we at least preserve some of the original behavior.

yuriks:
In any case this is all valuable research and I'd include it in the code as a comment.

bunnei:
@ObsidianX what is the plan with this PR?

ObsidianX:
@bunnei 

I implemented a simple hourly update using `CoreTiming`'s event system.  Since `::mktime` is inconsistent -- as @wwylele reported -- I've switched to a simple offset, assuming a 1970 epoch from `::time``.

I also updated the names of some variables to match @wwylele's findings.

src/core/hle/shared_page.cpp
```diff
@@ -3,22 +3,48 @@
...
 namespace SharedPage {
 
 SharedPageDef shared_page;
 
+static int update_time_event;   ///< Time is updated/swapped every hour
+
 void Init() {
     std::memset(&shared_page, 0, sizeof(shared_page));
 
-    shared_page.running_hw = 0x1; // product
```
wwylele: shouldn't be removed

src/core/hle/shared_page.cpp
```diff
@@ -3,22 +3,48 @@
...
+    u64_le console_time = (static_cast<u64_le>(plat_time) + _3DS_EPOCH_OFFSET) * 1000L;
+
+    DateTime* current_time = (shared_page.date_time_update_counter & 1) ? &shared_page.date_time_1 : &shared_page.date_time_0;
+
+    current_time->date_time = console_time;
+    current_time->tick_rate = (u64_le)CoreTiming::GetClockFrequency();
+    current_time->update_tick = CoreTiming::GetTicks();
+
+    // run again in an hour
+    CoreTiming::ScheduleEvent((u64_le)CoreTiming::GetClockFrequency() * 360ULL, update_time_event);
```
wwylele: ` * 3600ULL`

src/core/hle/shared_page.cpp
```diff
@@ -3,22 +3,50 @@
...
     // Some games wait until this value becomes 0x1, before asking running_hw
     shared_page.unknown_value = 0x1;
+
+    update_time_event = CoreTiming::RegisterEvent("SharedPage::update_time_event", UpdateTimeCallback);
+    // update now
+    CoreTiming::ScheduleEvent(0, update_time_event);
+}
+
+static void UpdateTimeCallback(u64 /*userdata*/, int /*cycles_late*/) {
+    shared_page.date_time_update_counter++;
```
wwylele: Actually this should be placed after updating date time to make it atomic... well since we currently emulate in single thread so it doesn't matter.

src/core/hle/shared_page.cpp
```diff
@@ -3,22 +3,53 @@
 // Refer to the license.txt file included.
 
 #include <cstring>
+#include <ctime>
+#include <core/core_timing.h>
```
lioncash: `This include should be using quotations like the ones below.

src/core/hle/shared_page.cpp
```diff
@@ -3,22 +3,53 @@
...
 
 #include "core/hle/shared_page.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// 3DS Uses 1900 for Epoch instead of 1970
+#define _3DS_EPOCH_OFFSET 2208988800L
```
lioncash: This should be a concrete `constexpr` variable. Also consider naming it `THREEDS_EPOCH_OFFSET`, as underscores followed by a capital letter are reserved by the language standard.

MerryMage:
`time_t` is ~usually~ in UTC; does this not take timezones into account?

wwylele:
As MerryMage said, `time_t` doesn't take timezone into account on my computer. I also watched a citra's Pokemon livestream (an unofficial build, with this PR included) that shows the time was off by one day. So we can't rely on the epoch of `time_t`. As yuriks suggested, C++'s chrono library may be a better choice, but I don't know much about it, though.

JohnGodgames:
Also helps to F1 2014 to get ingame

src/core/hle/shared_page.cpp
```diff
@@ -3,22 +3,53 @@
...
+}
+
+static void UpdateTimeCallback(u64 /*userdata*/, int /*cycles_late*/) {
+    u32_le next_count = shared_page.date_time_update_counter + 1;
+
+    // 3DS uses 1/1/1900 for Epoch
+    time_t plat_time = std::time(nullptr);
+    u64_le console_time = (static_cast<u64_le>(plat_time) + console_epoch_offset) * 1000L;
+
+    DateTime* current_time = (next_count & 1) ? &shared_page.date_time_1 : &shared_page.date_time_0;
```
Subv: Is there a reason why you're using a pointer instead of a reference here?

src/core/hle/shared_page.cpp
```diff
@@ -3,22 +3,53 @@
...
+
+    current_time->date_time = console_time;
+    current_time->tick_rate = (u64_le)CoreTiming::GetClockFrequency();
+    current_time->update_tick = CoreTiming::GetTicks();
+
+    // update after changing the opposite time structure to keep updates atomic
+    shared_page.date_time_update_counter = next_count;
+
+    // run again in an hour
+    CoreTiming::ScheduleEvent((u64_le)CoreTiming::GetClockFrequency() * 3600ULL, update_time_event);
```
Subv: There is a msToCycles (And probably a secondsToCycles) function, you should use that.

JohnGodgames:
**Also helps with games:**
F1 2011
Fairune
New Love Plus
New Love Plus Plus
Witch & Hero

wwylele:
Any news on this?

linkmauve:
Implemented in #1963.

PR #1644, opened 2016-04-06
Adopted WinterMute's gdbstub changes

polaris-:
This includes all changes from WinterMute's PR #1383. Since there hasn't been any updates to the PR, bunnei suggested I take it over since he wants to get these changes merged.

For the most part the code should be the same as WinterMute's except it contains the changes proposed by Subv, neobrain, and bunnei in the last PR.

### DIFF
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -36,25 +36,40 @@
 
 static void PrintHelp()
 {
-    std::cout << "Usage: citra <filename>" << std::endl;
+    std::cout << "Usage: citra [options] <filename>" << std::endl;
+    std::cout << "--help, -h            Display this information" << std::endl;
+    std::cout << "--gdbport, -g number  Enable gdb stub on port number" << std::endl;
 }
 
 /// Application entry point
 int main(int argc, char **argv) {
     int option_index = 0;
+    u32 gdb_port = 0;
+    char *endarg;
     std::string boot_filename;
+
     static struct option long_options[] = {
         { "help", no_argument, 0, 'h' },
+        { "gdbport", required_argument, 0, 'g' },
         { 0, 0, 0, 0 }
     };
 
     while (optind < argc) {
-        char arg = getopt_long(argc, argv, ":h", long_options, &option_index);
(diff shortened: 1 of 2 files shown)


src/citra/citra.cpp
```diff
@@ -76,8 +91,10 @@ int main(int argc, char **argv) {
     Config config;
     log_filter.ParseFilterString(Settings::values.log_filter);
 
-    GDBStub::ToggleServer(Settings::values.use_gdbstub);
-    GDBStub::SetServerPort(static_cast<u32>(Settings::values.gdbstub_port));
+    if (gdb_port != 0) {
```
Subv: This change overrides the config values, maybe we should use them as a fallback if the gdb option wasn't specified in the command line?

polaris-: Good point. A fallback would probably be more preferable, especially if the user sets true in the ini and expects it to always start the server. I will fix that up.

mailwl:
Sometimes IDA freezes, trying to read memory from unmapped memory (0x000000o-*), and Citra logs error "unknown GetPointer @ 0x..." permanently.
is there any way to fix it? For example, fill unaligned buffer with zeroes, without spam to log?

polaris-:
@mailwl I think that might be an issue with IDA and not exactly the gdbstub's fault. IDA tries to read *all* memory addresses byte by byte starting at 0x00000000 when you first attach even if the address isn't valid address for the hardware, which is why those errors get spammed in the log. I don't know if there's a way to stop that from happening. Since I don't think completely removing the error message would be a good idea, maybe adding an optional quiet parameter to GetPointer that would not display an error would be an acceptable fix? Would need feedback from some of the other devs on that one.

bunnei:
@mailwl @polaris- I'd like to get this PR merged things have been getting a bit unwieldy lately, and it's not really related to the changes in this patch (other than the fact that its related to the GDB stub).. Feel free to address it in a separate change, or open an issue.

mailwl:
sure, i am just interesting, and have no against this PR

PR #1645, opened 2016-04-06
Fix thumb ADR instruction alignment

mailwl:
Only one instruction fixed, not sure if another pc-based instructions must be aligned, too.

PR fixes the game **Conception 2: Children of the Seven Stars**

### DIFF
--- a/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
@@ -3955,9 +3955,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
         if (inst_base->cond == ConditionCode::AL || CondPassed(cpu, inst_base->cond)) {
             add_inst* const inst_cream = (add_inst*)inst_base->component;
 
-            u32 rn_val = RN;
-            if (inst_cream->Rn == 15)
-                rn_val += 2 * cpu->GetInstructionSize();
+            u32 rn_val = CHECK_READ_REG15_WA(cpu, inst_cream->Rn);
 
             bool carry;
             bool overflow;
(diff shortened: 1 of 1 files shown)


src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -3956,8 +3956,11 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
             add_inst* const inst_cream = (add_inst*)inst_base->component;
 
             u32 rn_val = RN;
-            if (inst_cream->Rn == 15)
+            if (inst_cream->Rn == 15) {
                 rn_val += 2 * cpu->GetInstructionSize();
+                if(cpu->TFlag)
```
lioncash: space after the if

lioncash: This if statement is unnecessary. You can just leave the AND in place.

lioncash: Something is wrong if code is reaching here and the T flag is set. Thumb mode is not capable of encoding PC as register.

lioncash: The branch here isn't necessary, you can just leave the AND

JayFoxRox:
(ARM DDI 0406) says (about ADR):
```
if ConditionPassed() then
  EncodingSpecificOperations();
  result = if add then (Align(PC,4) + imm32) else (Align(PC,4) - imm32);
  if d == 15 then // Can only occur for ARM encodings
    ALUWritePC(result);
  else
    R[d] = result;
```

So the align seems to happen before the addition and it seems to happen regardless of ARM / Thumb mode.
Most (all?) other places probably use [CHECK_READ_REG15_WA](https://github.com/citra-emu/citra/blob/master/src/core/arm/dyncom/arm_dyncom_run.h#L33) already which does the align.

mailwl:
changed to macro

lioncash:
@mailwl This should also be done with SUB as well.

@JayFoxRox ADR is a pseudoinstruction which the assembler translates to an ADD or a SUB depending on the offset.

lioncash:
Commit message should contain `alignment` instead of `align`. LGTM after that.

wwylele:
I don't know about ARM instructions, but I saw some other instructions such as ADC, CMN have similar pattern - should them also be fixed?

lioncash:
No, the other data-processing instructions don't require a fixed alignment. ADD and SUB are a special case, because ADR generates a PC-relative offset which needs to be word-aligned when the assembler decides which one the two instructions to generate.

MerryMage:
The alignment should only happen on ADD/SUB R15 with an immediate.
ADD/SUB with a register operand do not perform the align.
(And with a regsiter-shifted register it's unpredictable.)

Issue #1601, opened 2016-03-28
Lost Heroes 1 and 2 - Won't boot


JohnGodgames:
Propably related to https://github.com/citra-emu/citra/issues/1556

**Lost Heroes**
[LOG](http://pasted.co/78c764b4)

**Lost Heroes 2**
[LOG](http://pasted.co/96a18b0f)

mailwl:
to pass endless loop, you have to change 
https://github.com/citra-emu/citra/blob/master/src/core/hle/svc.cpp#L862

with this line:
`*out = (process->heap_used + process->linear_heap_used + process->misc_memory_used) & 0xFFFFFFFFFFFFF000ull;`

(it works at least in LH1). I think, it's aligning stuff or something like that (last 3 bytes must be 0)

mailwl:
to pass second stuck, you hate to implement ac:u::CloseAsync function, for example, in such way:
```c++
static void CloseAsync(Service::Interface* self) {
    u32* cmd_buff = Kernel::GetCommandBuffer();
    auto evt = Kernel::g_handle_table.Get<Kernel::Event>(cmd_buff[4]);
    if (evt) {
        evt->name = "AC_U:close_event";
        evt->Signal();
    }
    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
    LOG_WARNING(Service_AC, "(STUBBED) called");
}
```

Subv:
#1647 was merged, is this still an issue?

JohnGodgames:
It's broken again with something else...break reason panic.I will update issue later

JohnGodgames:
It works on latest master so I'm closing it...

PR #1647, opened 2016-04-07
ac:u: stub CloseAsync; align memory size in svc:GetProcessInfo(type=2)

mailwl:
code from Lost Heroes 1 and 2 - Won't boot #1601

### DIFF
--- a/src/core/hle/service/ac_u.cpp
+++ b/src/core/hle/service/ac_u.cpp
@@ -3,6 +3,8 @@
 // Refer to the license.txt file included.
 
 #include "common/logging/log.h"
+
+#include "core/hle/kernel/event.h"
 #include "core/hle/service/ac_u.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


src/core/hle/svc.cpp
```diff
@@ -859,7 +859,7 @@ static ResultCode GetProcessInfo(s64* out, Handle process_handle, u32 type) {
     case 2:
         // TODO(yuriks): Type 0 returns a slightly higher number than type 2, but I'm not sure
         // what's the difference between them.
-        *out = process->heap_used + process->linear_heap_used + process->misc_memory_used;
+        *out = (process->heap_used + process->linear_heap_used + process->misc_memory_used) & ~0xFFFull;
```
Subv: This is aligning the memory to PAGE_SIZE, it shouldn't be fixed like this, IMO what you should do is make sure that heap_used, linear_heap_used and misc_memory_used all contain page-size-aligned sizes, this can be done by ensuring that the minimal allocation size is PAGE_SIZE

src/core/hle/service/ac_u.cpp
```diff
@@ -11,6 +13,26 @@
 namespace AC_U {
 
 /**
+ * AC_U::CloseAsync service function
+ *  Inputs:
+ *      4 : Event handle, should be signalled when AC connection is closed
```
Subv: signaled.
Please document the rest of the inputs

bunnei:
Looks fine to me

PR #1652, opened 2016-04-08
Debugger: implement wait tree widget

wwylele:
I just found this in the TODO list and gave a try. 
How it looks:
wt
When game is paused, it lists all threads, and show their status and what are they waiting for (recursively).
Needs more testing, because I don't have many games to test all cases.
Also please help improve the expression in the "tree view" (list view in fact)
If this is what it supposed to be, I will also try implementing other debugger.

### DIFF
--- a/src/citra_qt/debugger/wait_tree.cpp
+++ b/src/citra_qt/debugger/wait_tree.cpp
@@ -1,0 +1,417 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "citra_qt/debugger/wait_tree.h"
+#include "citra_qt/util/util.h"
+
+#include "core/hle/kernel/event.h"
+#include "core/hle/kernel/mutex.h"
+#include "core/hle/kernel/semaphore.h"
+#include "core/hle/kernel/session.h"
+#include "core/hle/kernel/thread.h"
+#include "core/hle/kernel/timer.h"
+
+WaitTreeItem::~WaitTreeItem() {}
+
+QColor WaitTreeItem::GetColor() const {
+    return QColor(Qt::GlobalColor::black);
+}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeItem::GetChildren() const {
+    return {};
+}
+
+void WaitTreeItem::Expand() {
+    if (IsExpandable() && !expanded) {
+        children = GetChildren();
+        for (std::size_t i = 0; i < children.size(); ++i) {
+            children[i]->parent = this;
+            children[i]->row = i;
+        }
+        expanded = true;
+    }
+}
+
+WaitTreeItem* WaitTreeItem::Parent() const {
+    return parent;
+}
+
+const std::vector<std::unique_ptr<WaitTreeItem>>& WaitTreeItem::Children() const {
+    return children;
+}
+
+bool WaitTreeItem::IsExpandable() const {
+    return false;
+}
+
+std::size_t WaitTreeItem::Row() const {
+    return row;
+}
+
+std::vector<std::unique_ptr<WaitTreeThread>> WaitTreeItem::MakeThreadItemList() {
+    const auto& threads = Kernel::GetThreadList();
+    std::vector<std::unique_ptr<WaitTreeThread>> item_list;
+    item_list.reserve(threads.size());
+    for (std::size_t i = 0; i < threads.size(); ++i) {
+        item_list.push_back(std::make_unique<WaitTreeThread>(*threads[i]));
+        item_list.back()->row = i;
+    }
+    return item_list;
+}
+
+WaitTreeText::WaitTreeText(const QString& t) : text(t) {}
+
+QString WaitTreeText::GetText() const {
+    return text;
+}
+
+WaitTreeWaitObject::WaitTreeWaitObject(const Kernel::WaitObject& o) : object(o) {}
+
+bool WaitTreeExpandableItem::IsExpandable() const {
+    return true;
+}
+
+QString WaitTreeWaitObject::GetText() const {
+    return tr("[%1]%2 %3")
+        .arg(object.GetObjectId())
+        .arg(QString::fromStdString(object.GetTypeName()),
+             QString::fromStdString(object.GetName()));
+}
+
+std::unique_ptr<WaitTreeWaitObject> WaitTreeWaitObject::make(const Kernel::WaitObject& object) {
+    switch (object.GetHandleType()) {
+    case Kernel::HandleType::Event:
+        return std::make_unique<WaitTreeEvent>(static_cast<const Kernel::Event&>(object));
+    case Kernel::HandleType::Mutex:
+        return std::make_unique<WaitTreeMutex>(static_cast<const Kernel::Mutex&>(object));
+    case Kernel::HandleType::Semaphore:
+        return std::make_unique<WaitTreeSemaphore>(static_cast<const Kernel::Semaphore&>(object));
+    case Kernel::HandleType::Timer:
+        return std::make_unique<WaitTreeTimer>(static_cast<const Kernel::Timer&>(object));
+    case Kernel::HandleType::Thread:
+        return std::make_unique<WaitTreeThread>(static_cast<const Kernel::Thread&>(object));
+    default:
+        return std::make_unique<WaitTreeWaitObject>(object);
+    }
+}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeWaitObject::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list;
+
+    const auto& threads = object.GetWaitingThreads();
+    if (threads.empty()) {
+        list.push_back(std::make_unique<WaitTreeText>(tr("waited by no thread")));
+    } else {
+        list.push_back(std::make_unique<WaitTreeThreadList>(threads));
+    }
+    return list;
+}
+
+QString WaitTreeWaitObject::GetResetTypeQString(Kernel::ResetType reset_type) {
+    switch (reset_type) {
+    case Kernel::ResetType::OneShot:
+        return tr("one shot");
+    case Kernel::ResetType::Sticky:
+        return tr("sticky");
+    case Kernel::ResetType::Pulse:
+        return tr("pulse");
+    }
+}
+
+WaitTreeObjectList::WaitTreeObjectList(
+    const std::vector<Kernel::SharedPtr<Kernel::WaitObject>>& list, bool w_all)
+    : object_list(list), wait_all(w_all) {}
+
+QString WaitTreeObjectList::GetText() const {
+    if (wait_all)
+        return tr("waiting for all objects");
+    return tr("waiting for one of the following objects");
+}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeObjectList::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(object_list.size());
+    std::transform(object_list.begin(), object_list.end(), list.begin(),
+                   [](const auto& t) { return WaitTreeWaitObject::make(*t); });
+    return list;
+}
+
+WaitTreeThread::WaitTreeThread(const Kernel::Thread& thread) : WaitTreeWaitObject(thread) {}
+
+QString WaitTreeThread::GetText() const {
+    const auto& thread = static_cast<const Kernel::Thread&>(object);
+    QString status;
+    switch (thread.status) {
+    case THREADSTATUS_RUNNING:
+        status = tr("running");
+        break;
+    case THREADSTATUS_READY:
+        status = tr("ready");
+        break;
+    case THREADSTATUS_WAIT_ARB:
+        status = tr("waiting for address 0x%1").arg(thread.wait_address, 8, 16, QLatin1Char('0'));
+        break;
+    case THREADSTATUS_WAIT_SLEEP:
+        status = tr("sleeping");
+        break;
+    case THREADSTATUS_WAIT_SYNCH:
+        status = tr("waiting for objects");
+        break;
+    case THREADSTATUS_DORMANT:
+        status = tr("dormant");
+        break;
+    case THREADSTATUS_DEAD:
+        status = tr("dead");
+        break;
+    }
+    QString pc_info = tr(" PC = 0x%1 LR = 0x%2")
+                          .arg(thread.context.pc, 8, 16, QLatin1Char('0'))
+                          .arg(thread.context.lr, 8, 16, QLatin1Char('0'));
+    return WaitTreeWaitObject::GetText() + pc_info + " (" + status + ") ";
+}
+
+QColor WaitTreeThread::GetColor() const {
+    const auto& thread = static_cast<const Kernel::Thread&>(object);
+    switch (thread.status) {
+    case THREADSTATUS_RUNNING:
+        return QColor(Qt::GlobalColor::darkGreen);
+    case THREADSTATUS_READY:
+        return QColor(Qt::GlobalColor::darkBlue);
+    case THREADSTATUS_WAIT_ARB:
+        return QColor(Qt::GlobalColor::darkRed);
+    case THREADSTATUS_WAIT_SLEEP:
+        return QColor(Qt::GlobalColor::darkYellow);
+    case THREADSTATUS_WAIT_SYNCH:
+        return QColor(Qt::GlobalColor::red);
+    case THREADSTATUS_DORMANT:
+        return QColor(Qt::GlobalColor::darkCyan);
+    case THREADSTATUS_DEAD:
+        return QColor(Qt::GlobalColor::gray);
+    default:
+        return WaitTreeItem::GetColor();
+    }
+}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThread::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(WaitTreeWaitObject::GetChildren());
+
+    const auto& thread = static_cast<const Kernel::Thread&>(object);
+
+    QString processor;
+    switch (thread.processor_id) {
+    case ThreadProcessorId::THREADPROCESSORID_DEFAULT:
+        processor = tr("default");
+        break;
+    case ThreadProcessorId::THREADPROCESSORID_ALL:
+        processor = tr("all");
+        break;
+    case ThreadProcessorId::THREADPROCESSORID_0:
+        processor = tr("AppCore");
+        break;
+    case ThreadProcessorId::THREADPROCESSORID_1:
+        processor = tr("SysCore");
+        break;
+    default:
+        processor = tr("Unknown processor %1").arg(thread.processor_id);
+        break;
+    }
+
+    list.push_back(std::make_unique<WaitTreeText>(tr("processor = %1").arg(processor)));
+    list.push_back(std::make_unique<WaitTreeText>(tr("thread id = %1").arg(thread.GetThreadId())));
+    list.push_back(std::make_unique<WaitTreeText>(tr("priority = %1(current) / %2(normal)")
+                                                      .arg(thread.current_priority)
+                                                      .arg(thread.nominal_priority)));
+    list.push_back(std::make_unique<WaitTreeText>(
+        tr("last running ticks = %1").arg(thread.last_running_ticks)));
+
+    if (thread.held_mutexes.empty()) {
+        list.push_back(std::make_unique<WaitTreeText>(tr("not holding mutex")));
+    } else {
+        list.push_back(std::make_unique<WaitTreeMutexList>(thread.held_mutexes));
+    }
+    if (thread.status == THREADSTATUS_WAIT_SYNCH) {
+        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, thread.wait_all));
+    }
+
+    return list;
+}
+
+WaitTreeEvent::WaitTreeEvent(const Kernel::Event& object) : WaitTreeWaitObject(object) {}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeEvent::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(WaitTreeWaitObject::GetChildren());
+
+    list.push_back(std::make_unique<WaitTreeText>(
+        tr("reset type = %1")
+            .arg(GetResetTypeQString(static_cast<const Kernel::Event&>(object).reset_type))));
+    return list;
+}
+
+WaitTreeMutex::WaitTreeMutex(const Kernel::Mutex& object) : WaitTreeWaitObject(object) {}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeMutex::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(WaitTreeWaitObject::GetChildren());
+
+    const auto& mutex = static_cast<const Kernel::Mutex&>(object);
+    if (mutex.lock_count) {
+        list.push_back(
+            std::make_unique<WaitTreeText>(tr("locked %1 times by thread:").arg(mutex.lock_count)));
+        list.push_back(std::make_unique<WaitTreeThread>(*mutex.holding_thread));
+    } else {
+        list.push_back(std::make_unique<WaitTreeText>(tr("free")));
+    }
+    return list;
+}
+
+WaitTreeSemaphore::WaitTreeSemaphore(const Kernel::Semaphore& object)
+    : WaitTreeWaitObject(object) {}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeSemaphore::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(WaitTreeWaitObject::GetChildren());
+
+    const auto& semaphore = static_cast<const Kernel::Semaphore&>(object);
+    list.push_back(
+        std::make_unique<WaitTreeText>(tr("available count = %1").arg(semaphore.available_count)));
+    list.push_back(std::make_unique<WaitTreeText>(tr("max count = %1").arg(semaphore.max_count)));
+    return list;
+}
+
+WaitTreeTimer::WaitTreeTimer(const Kernel::Timer& object) : WaitTreeWaitObject(object) {}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeTimer::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(WaitTreeWaitObject::GetChildren());
+
+    const auto& timer = static_cast<const Kernel::Timer&>(object);
+
+    list.push_back(std::make_unique<WaitTreeText>(
+        tr("reset type = %1").arg(GetResetTypeQString(timer.reset_type))));
+    list.push_back(
+        std::make_unique<WaitTreeText>(tr("initial delay = %1").arg(timer.initial_delay)));
+    list.push_back(
+        std::make_unique<WaitTreeText>(tr("interval delay = %1").arg(timer.interval_delay)));
+    return list;
+}
+
+WaitTreeMutexList::WaitTreeMutexList(
+    const boost::container::flat_set<Kernel::SharedPtr<Kernel::Mutex>>& list)
+    : mutex_list(list) {}
+
+QString WaitTreeMutexList::GetText() const {
+    return tr("holding mutexes");
+}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeMutexList::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(mutex_list.size());
+    std::transform(mutex_list.begin(), mutex_list.end(), list.begin(),
+                   [](const auto& t) { return std::make_unique<WaitTreeMutex>(*t); });
+    return list;
+}
+
+WaitTreeThreadList::WaitTreeThreadList(const std::vector<Kernel::SharedPtr<Kernel::Thread>>& list)
+    : thread_list(list) {}
+
+QString WaitTreeThreadList::GetText() const {
+    return tr("waited by thread");
+}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThreadList::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(thread_list.size());
+    std::transform(thread_list.begin(), thread_list.end(), list.begin(),
+                   [](const auto& t) { return std::make_unique<WaitTreeThread>(*t); });
+    return list;
+}
+
+WaitTreeModel::WaitTreeModel(QObject* parent) : QAbstractItemModel(parent) {}
+
+QModelIndex WaitTreeModel::index(int row, int column, const QModelIndex& parent) const {
+    if (!hasIndex(row, column, parent))
+        return {};
+
+    if (parent.isValid()) {
+        WaitTreeItem* parent_item = static_cast<WaitTreeItem*>(parent.internalPointer());
+        parent_item->Expand();
+        return createIndex(row, column, parent_item->Children()[row].get());
+    }
+
+    return createIndex(row, column, thread_items[row].get());
+}
+
+QModelIndex WaitTreeModel::parent(const QModelIndex& index) const {
+    if (!index.isValid())
+        return {};
+
+    WaitTreeItem* parent_item = static_cast<WaitTreeItem*>(index.internalPointer())->Parent();
+    if (!parent_item) {
+        return QModelIndex();
+    }
+    return createIndex(static_cast<int>(parent_item->Row()), 0, parent_item);
+}
+
+int WaitTreeModel::rowCount(const QModelIndex& parent) const {
+    if (!parent.isValid())
+        return static_cast<int>(thread_items.size());
+
+    WaitTreeItem* parent_item = static_cast<WaitTreeItem*>(parent.internalPointer());
+    parent_item->Expand();
+    return static_cast<int>(parent_item->Children().size());
+}
+
+int WaitTreeModel::columnCount(const QModelIndex&) const {
+    return 1;
+}
+
+QVariant WaitTreeModel::data(const QModelIndex& index, int role) const {
+    if (!index.isValid())
+        return {};
+
+    switch (role) {
+    case Qt::DisplayRole:
+        return static_cast<WaitTreeItem*>(index.internalPointer())->GetText();
+    case Qt::ForegroundRole:
+        return static_cast<WaitTreeItem*>(index.internalPointer())->GetColor();
+    default:
+        return {};
+    }
+}
+
+void WaitTreeModel::ClearItems() {
+    thread_items.clear();
+}
+
+void WaitTreeModel::InitItems() {
+    thread_items = WaitTreeItem::MakeThreadItemList();
+}
+
+WaitTreeWidget::WaitTreeWidget(QWidget* parent) : QDockWidget(tr("Wait Tree"), parent) {
+    setObjectName("WaitTreeWidget");
+    view = new QTreeView(this);
+    view->setHeaderHidden(true);
+    setWidget(view);
+    setEnabled(false);
+}
+
+void WaitTreeWidget::OnDebugModeEntered() {
+    if (!Core::g_app_core)
+        return;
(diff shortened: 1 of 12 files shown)


src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+
+#include "core/core.h"
+
+#include "citra_qt/util/util.h"
+#include "citra_qt/debugger/wait_tree.h"
+
+bool WaitTreeItem::IsExpandable() const {
```
lioncash: This shouldn't be before the constructor

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+
+#include "core/core.h"
+
+#include "citra_qt/util/util.h"
```
lioncash: citra_qt comes before core

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+#include "citra_qt/debugger/wait_tree.h"
+
+bool WaitTreeItem::IsExpandable() const {
+    return false;
+}
+WaitTreeItem::~WaitTreeItem() {
+
```
lioncash: unnecessary newline

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+}
+
+
+WaitTreeText::WaitTreeText(const QString& text)
+    : text(text){
```
lioncash: brace goes on the same line as the function name, not the initializer list.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+}
+QString WaitTreeText::GetText() const {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    return text;
+}
+
+bool WaitTreeExpandableItem::IsExpandable() const {
+    return true;
+}
+
+
```
lioncash: unnecessary newline

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+WaitTreeWaitObject::WaitTreeWaitObject(const Kernel::WaitObject& object)
+    : object(object) {
```
lioncash: Brace goes on the function, not the initializer list.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    } else {
+        list.push_back(std::make_unique<WaitTreeThreadList>(threads));
+    }
+    return list;
+}
+
+WaitTreeObjectList::WaitTreeObjectList(const std::vector<Kernel::SharedPtr<Kernel::WaitObject>>& list, bool wait_all)
+    : object_list(list), wait_all(wait_all) {
+}
+QString WaitTreeObjectList::GetText() const {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+        return std::make_unique<WaitTreeSemaphore>(reinterpret_cast<const Kernel::Semaphore&>(object));
+    case Kernel::HandleType::Timer:
+        return std::make_unique<WaitTreeTimer>(reinterpret_cast<const Kernel::Timer&>(object));
+    case Kernel::HandleType::Thread:
+        return std::make_unique<WaitTreeThread>(reinterpret_cast<const Kernel::Thread&>(object));
+    default:
+        return std::make_unique<WaitTreeWaitObject>(object);
+    }
+}
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeWaitObject::GetChildren() const {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+WaitTreeWaitObject::WaitTreeWaitObject(const Kernel::WaitObject& object)
+    : object(object) {
+}
+QString WaitTreeWaitObject::GetText() const {
+    return tr("[%1]%2 %3").arg(object.GetObjectId())
+        .arg(QString::fromStdString(object.GetTypeName()),
+            QString::fromStdString(object.GetName()));
+}
+std::unique_ptr<WaitTreeWaitObject> WaitTreeWaitObject::make(const Kernel::WaitObject& object) {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+bool WaitTreeExpandableItem::IsExpandable() const {
+    return true;
+}
+
+
+WaitTreeWaitObject::WaitTreeWaitObject(const Kernel::WaitObject& object)
+    : object(object) {
+}
+QString WaitTreeWaitObject::GetText() const {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    if (threads.empty()) {
+        list.push_back(std::make_unique<WaitTreeText>(tr("waited by no thread")));
+    } else {
+        list.push_back(std::make_unique<WaitTreeThreadList>(threads));
+    }
+    return list;
+}
+
+WaitTreeObjectList::WaitTreeObjectList(const std::vector<Kernel::SharedPtr<Kernel::WaitObject>>& list, bool wait_all)
+    : object_list(list), wait_all(wait_all) {
```
lioncash: brace goes on the function line, not the initializer list.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    case Kernel::HandleType::Thread:
+        return std::make_unique<WaitTreeThread>(reinterpret_cast<const Kernel::Thread&>(object));
+    default:
+        return std::make_unique<WaitTreeWaitObject>(object);
+    }
+}
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeWaitObject::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list;
+
+    const auto& threads=object.GetWaitingThreads();
```
lioncash: spaces on both sides of the `=`

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    std::transform(object_list.begin(), object_list.end(), list.begin(),
+        [](const auto& t){return WaitTreeWaitObject::make(*t); }
+    );
+    return list;
+}
+
+WaitTreeThread::WaitTreeThread(const Kernel::Thread& thread)
+    : WaitTreeWaitObject(thread) {
+}
+QString WaitTreeThread::GetText() const {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    : object_list(list), wait_all(wait_all) {
+}
+QString WaitTreeObjectList::GetText() const {
+    if (wait_all) {
+        return tr("waiting for all of objects");
+    } else {
+        return tr("waiting for one of objects");
+    }
+}
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeObjectList::GetChildren() const {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    std::vector<std::unique_ptr<WaitTreeItem>> list(object_list.size());
+    std::transform(object_list.begin(), object_list.end(), list.begin(),
```
lioncash:
```cpp
std::transform(object_list.begin(), object_list.end(), list.begin(), [](const auto& t) {
    return WaitTreeWaitObject::make(*t);
});
```

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+        [](const auto& t){return WaitTreeWaitObject::make(*t); }
+    );
+    return list;
+}
+
+WaitTreeThread::WaitTreeThread(const Kernel::Thread& thread)
+    : WaitTreeWaitObject(thread) {
```
lioncash: brace on the function line, not the initializer list.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    case THREADSTATUS_DORMANT:
+        status = tr("dormant");
+        break;
+    case THREADSTATUS_DEAD:
+        status = tr("dead");
+        break;
+    }
+    return WaitTreeWaitObject::GetText() + " (" + status + ")";
+}
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThread::GetChildren() const {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects,thread.wait_all));
+    }
+
+    return list;
+}
+
+WaitTreeEvent::WaitTreeEvent(const Kernel::Event& object)
+    : WaitTreeWaitObject(object) {
+}
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeEvent::GetChildren() const {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    }
+    if (thread.status == THREADSTATUS_WAIT_SYNCH) {
+        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects,thread.wait_all));
+    }
+
+    return list;
+}
+
+WaitTreeEvent::WaitTreeEvent(const Kernel::Event& object)
+    : WaitTreeWaitObject(object) {
```
lioncash: Brace on the function line, not the initializer list.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    }
+    list.push_back(std::make_unique<WaitTreeText>(tr("reset type = %1").arg(reset_type)));
+    return list;
+}
+
+WaitTreeMutex::WaitTreeMutex(const Kernel::Mutex& object)
+    : WaitTreeWaitObject(object) {
+
+}
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeMutex::GetChildren() const {
```
lioncash: newline between functions

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+    case Kernel::ResetType::Pulse:
+        reset_type = tr("pulse");
+        break;
+    }
+    list.push_back(std::make_unique<WaitTreeText>(tr("reset type = %1").arg(reset_type)));
+    return list;
+}
+
+WaitTreeMutex::WaitTreeMutex(const Kernel::Mutex& object)
+    : WaitTreeWaitObject(object) {
```
lioncash: brace on the function line, not this initializer list.

mailwl:
@wwylele , can you change the thead name to 
`name = Common::StringFromFormat("thread-%08x", Memory::Read32(arg+8));`
at line https://github.com/citra-emu/citra/blob/master/src/core/hle/svc.cpp#L475 ?
it's real entry point

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+class EmuThread;
+
+class WaitTreeItem : public QObject{
+    Q_OBJECT
+public:
+    virtual bool IsExpandable() const;
+    virtual QString GetText() const = 0;
+    virtual ~WaitTreeItem();
+};
+class WaitTreeText : public WaitTreeItem {
```
lioncash: newline between classes

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+#include "core/hle/kernel/thread.h"
+#include "core/hle/kernel/event.h"
+#include "core/hle/kernel/semaphore.h"
+#include "core/hle/kernel/mutex.h"
+#include "core/hle/kernel/session.h"
+#include "core/hle/kernel/timer.h"
+
+class EmuThread;
+
+class WaitTreeItem : public QObject{
```
lioncash: space before `{`

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+    virtual QString GetText() const = 0;
+    virtual ~WaitTreeItem();
+};
+class WaitTreeText : public WaitTreeItem {
+    Q_OBJECT
+private:
+    QString text;
+public:
+    WaitTreeText(const QString& text);
+    virtual QString GetText() const override;
```
lioncash: virtual isn't necessary if override is specified. Same for all the others.

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+class WaitTreeItem : public QObject{
+    Q_OBJECT
+public:
+    virtual bool IsExpandable() const;
+    virtual QString GetText() const = 0;
+    virtual ~WaitTreeItem();
+};
+class WaitTreeText : public WaitTreeItem {
+    Q_OBJECT
+private:
```
lioncash: Consider putting your private members after the public interface. People look at that first.

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+    QString text;
+public:
+    WaitTreeText(const QString& text);
+    virtual QString GetText() const override;
+};
+class WaitTreeExpandableItem : public WaitTreeItem {
```
lioncash: newline between classes (etc, etc for all other occurrences)

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+    Q_OBJECT
+public:
+    virtual bool IsExpandable() const override;
+    virtual std::vector<std::unique_ptr<WaitTreeItem>> GetChildren() const = 0;
+};
+class WaitTreeWaitObject : public WaitTreeExpandableItem {
+    Q_OBJECT
+protected:
```
lioncash: public interface should come first.

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+public:
+    WaitTreeWaitObject(const Kernel::WaitObject& object);
+    static std::unique_ptr<WaitTreeWaitObject> make(const Kernel::WaitObject& object);
+    virtual QString GetText() const override;
+    virtual std::vector<std::unique_ptr<WaitTreeItem>> GetChildren() const override;
+};
+
+class WaitTreeObjectList : public WaitTreeExpandableItem {
+    Q_OBJECT
+private:
```
lioncash: public interface should come first.

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+    Q_OBJECT
+public:
+    WaitTreeTimer(const Kernel::Timer& object);
+
+    virtual std::vector<std::unique_ptr<WaitTreeItem>> GetChildren() const override;
+};
+
+class WaitTreeMutexList : public WaitTreeExpandableItem {
+    Q_OBJECT
+private:
```
lioncash: public interface should come first.

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+    const boost::container::flat_set<Kernel::SharedPtr<Kernel::Mutex>>& mutex_list;
+public:
+    WaitTreeMutexList(const boost::container::flat_set<Kernel::SharedPtr<Kernel::Mutex>>& list);
+    virtual QString GetText() const override;
+    virtual std::vector<std::unique_ptr<WaitTreeItem>> GetChildren() const override;
+};
+
+class WaitTreeThreadList : public WaitTreeExpandableItem {
+    Q_OBJECT
+private:
```
lioncash: public interface should come first.

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+    WaitTreeThreadList(const std::vector<Kernel::SharedPtr<Kernel::Thread>>& list);
+    virtual QString GetText() const override;
+    virtual std::vector<std::unique_ptr<WaitTreeItem>> GetChildren() const override;
+};
+
+
+
+class WaitTree : public QObject {
+    Q_OBJECT
+private:
```
lioncash: public interface should come first.

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,182 @@
...
+    std::vector<std::pair<std::unique_ptr<WaitTreeItem>, int>> list;
+
+public:
+    void Init(std::vector<std::unique_ptr<WaitTreeItem>>&& root_list);
+    void Clear() {
+        list.clear();
+    }
+    void ToggleExpand(size_t index);
+    bool IsExpandable(size_t index) const {
```
lioncash: These should probably be defined in the .cpp file to stay consistent with everything else.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+}
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThreadList::GetChildren() const  {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(thread_list.size());
+    std::transform(thread_list.begin(), thread_list.end(), list.begin(),
+        [](const auto& t){return std::make_unique<WaitTreeThread>(*t); });
+    return list;
+}
+
+
+
```
lioncash: way too many newlines here

neobrain:
Have you considered using QTreeView instead?

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+WaitTreeMutexList::WaitTreeMutexList(const boost::container::flat_set<Kernel::SharedPtr<Kernel::Mutex>>& list)
+    : mutex_list(list) {
+}
+QString WaitTreeMutexList::GetText() const {
+    return tr("holding mutexes");
+}
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeMutexList::GetChildren() const {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(mutex_list.size());
+    std::transform(mutex_list.begin(), mutex_list.end(), list.begin(),
```
lioncash:
```cpp
std::transform(mutex_list.begin(), mutex_list.end(), list.begin(), [](const auto& t){
    return std::make_unique<WaitTreeMutex>(*t);
});
```

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+WaitTreeThreadList::WaitTreeThreadList(const std::vector<Kernel::SharedPtr<Kernel::Thread>>& list)
+    : thread_list(list) {
+}
+QString WaitTreeThreadList::GetText() const {
+    return tr("waited by thread");
+}
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThreadList::GetChildren() const  {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(thread_list.size());
+    std::transform(thread_list.begin(), thread_list.end(), list.begin(),
```
lioncash:
```cpp
std::transform(thread_list.begin(), thread_list.end(), list.begin(), [](const auto& t) {
    return std::make_unique<WaitTreeThread>(*t);
});
```

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+void WaitTreeModel::ClearItems() {
+    wait_tree.Clear();
+    emit layoutChanged();
+}
+void WaitTreeModel::InitItems() {
+    const auto& threads(Kernel::GetThreadList());
+
+    std::vector<std::unique_ptr<WaitTreeItem>> list(threads.size());
+    std::transform(threads.begin(), threads.end(), list.begin(),
```
lioncash: std::transform(threads.begin(), threads.end(), list.begin(), [](const auto& t) {
    return std::make_unique<WaitTreeThread>(*t);
});
```

File: src/citra_qt/debugger/wait_tree.cpp [hunk 38]
```diff
@@ -0,0 +1,349 @@
...
+}
+QString WaitTreeWaitObject::GetText() const {
+    return tr("[%1]%2 %3").arg(object.GetObjectId())
+        .arg(QString::fromStdString(object.GetTypeName()),
+            QString::fromStdString(object.GetName()));
+}
+std::unique_ptr<WaitTreeWaitObject> WaitTreeWaitObject::make(const Kernel::WaitObject& object) {
+    switch (object.GetHandleType()) {
+    case Kernel::HandleType::Event:
+        return std::make_unique<WaitTreeEvent>(reinterpret_cast<const Kernel::Event&>(object));
```
lioncash: These can be static_casts (same with most other usages of reinterpret_cast)

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,349 @@
...
+void WaitTree::ToggleExpand(size_t index) {
+    if (!list[index].first->IsExpandable()) return;
+    if (IsExpanded(index)) {
+        int level = list[index].second;
+        auto child_begin = list.begin() + index + 1;
+        auto child_end = std::find_if_not(child_begin, list.end(),
+            [level](const auto& item){return item.second>level; });
+        list.erase(child_begin, child_end);
+    }
+    else{
```
lioncash: `else {`

lioncash:
Also note that this much object casting down/up the type hierarchy usually indicates a design issue.

mailwl:
also, can this window updated on every thread switching (or every 100 switching) not only on pause?
at least at debug mode. Anyway, this window useful for games, which don't properly emulated, emulated speed does not matter

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+bool WaitTreeItem::IsExpandable() const {
+    return false;
+}
+WaitTreeItem::~WaitTreeItem() {
+}
+
+
+WaitTreeText::WaitTreeText(const QString& text) :
+    text(text){
```
lioncash: space after the `{`

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,392 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+
+#include "citra_qt/util/util.h"
+#include "citra_qt/debugger/wait_tree.h"
+
+#include "core/core.h"
+
+bool WaitTreeItem::IsExpandable() const {
```
lioncash: This should be placed after the constructors

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,392 @@
...
+    return false;
+}
+WaitTreeItem::~WaitTreeItem() {
+}
+
+
+WaitTreeText::WaitTreeText(const QString& text) :
```
lioncash: Note that this parameter will cause variable shadowing warnings because it's the same name as the member variable (this applies to other cases as well).

wwylele:
@neobrain  I am not so familiar with Qt, and get confused when reading document for TreeView. However if a TreeView is preferred, I will give a second try on it
@lioncash you are right. I didn't think calmly and it can't be well-designed . I will think it again (with TreeView way)
@mailwl I am considering add a button allowing step to switching when pausing. Is it a better idea?

mailwl:
hmm... i don't know. Sometimes need to see thread switching in action (btw, may threads be different colors for each state? for example, red for running, grey for dead?).

also, may you rename events, created by game, in such way (to see, where it comes from)
`evt->name = Common::StringFromFormat("event-%08x", Core::g_app_core->GetReg(14));`
at line https://github.com/citra-emu/citra/blob/master/src/core/hle/svc.cpp#L666

wwylele:
@mailwl events renamed, along with other three objects. Text coloring will be considered with TreeView reimplement.

wwylele:
Now I implement it as a tree view (with colored text)
ft

I still didn't find a good way to avoid casting. Making use of polymorphic kernel object classes may be a good idea, but that would be so intrusive.

mailwl:
is it possible to add subtree with reason, why threads goes to Ready state (arbitrated address, or name  of signaled event)?
just signal event not enough in some cases, need to fill other data, because of it threads again starts to wait same event, and don't get needed data again.
For example: Circle Pad Pro, game try to connection to infrared, got Connection Event from ir:u, but then check a value in shared memory, if it not equal 4, closes connection, and tryes again (similar #1613), with awake reason more chances to understand, why games just freese

wwylele:
@mailwl That's hard for me because I don't even have a full understanding of Citra's threading... I think we need an expert on it.

mailwl:
tested a lot and hard, - LGTM

wwylele:
Update and fix:
display reset type for timer; display processor for thread; fix typos

Subv:
Huh, I thought this was already merged. I'll review it later today.

src/citra_qt/debugger/wait_tree.h
```diff
@@ -0,0 +1,168 @@
...
+    WaitTreeItem();
+    virtual bool IsExpandable() const;
+    virtual std::vector<std::unique_ptr<WaitTreeItem>> GetChildren() const;
+    virtual QString GetText() const = 0;
+    virtual QColor GetColor() const;
+    virtual ~WaitTreeItem();
+    void Expand();
+    WaitTreeItem* Parent() const;
+    const std::vector<std::unique_ptr<WaitTreeItem>>& Children() const;
+    std::size_t row;
```
lioncash: Surely there's a better way than exposing this

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+// Refer to the license.txt file included.
+
+
+#include "citra_qt/util/util.h"
+#include "citra_qt/debugger/wait_tree.h"
+
+#include "core/core.h"
+
+WaitTreeItem::WaitTreeItem() :
+    expanded(false), parent(nullptr), children() {
```
lioncash: These can just be directly assigned to the class members in the class itself (`children()` isn't needed as well)

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+
+WaitTreeObjectList::WaitTreeObjectList(const std::vector<Kernel::SharedPtr<Kernel::WaitObject>>& list, bool w_all) :
+    object_list(list), wait_all(w_all) {
+}
+
+QString WaitTreeObjectList::GetText() const {
+    if (wait_all) {
+        return tr("waiting for all of objects");
+    } else {
+        return tr("waiting for one of objects");
```
lioncash: `Waiting on an object`

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+    }
+}
+
+WaitTreeObjectList::WaitTreeObjectList(const std::vector<Kernel::SharedPtr<Kernel::WaitObject>>& list, bool w_all) :
+    object_list(list), wait_all(w_all) {
+}
+
+QString WaitTreeObjectList::GetText() const {
+    if (wait_all) {
+        return tr("waiting for all of objects");
```
lioncash: `Waiting on all objects`

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+    } else {
```
lioncash: else isn't necessary here.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+    thread_list(list) {
+}
+
+QString WaitTreeThreadList::GetText() const {
+    return tr("waited by thread");
+}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThreadList::GetChildren() const  {
+    std::vector<std::unique_ptr<WaitTreeItem>> list(thread_list.size());
+    std::transform(thread_list.begin(), thread_list.end(), list.begin(), [](const auto& t){
```
lioncash: space between the `)` and brace

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+        return QModelIndex();
+    }
+    return createIndex(static_cast<int>(parent_item->row), 0, parent_item);
+
+}
+
+int WaitTreeModel::rowCount(const QModelIndex& parent) const {
+    if (!parent.isValid()) {
+        return static_cast<int>(thread_items.size());
+    } else {
```
lioncash: else isn't necessary here.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+    QDockWidget(tr("Wait Tree"), parent) {
+    setObjectName("WaitTreeWidget");
+    view = new QTreeView(this);
+    view->setHeaderHidden(true);
+    setWidget(view);
+    setEnabled(false);
+}
+
+void WaitTreeWidget::OnDebugModeEntered() {
+    if (!Core::g_app_core) return;
```
lioncash:
```cpp
if (!Core::g_app_core)
    return;
```

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+    });
+    return list;
+}
+
+WaitTreeModel::WaitTreeModel(QObject *parent) :
+    QAbstractItemModel(parent) {
+}
+
+QModelIndex WaitTreeModel::index(int row, int column, const QModelIndex& parent) const {
+    if (!hasIndex(row, column, parent)) return QModelIndex();
```
lioncash:
```cpp
if (!hasIndex(row, column, parent))
    return {};
```

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+
+    if (parent.isValid()) {
+        WaitTreeItem* parent_item = static_cast<WaitTreeItem*>(parent.internalPointer());
+        parent_item->Expand();
+        return createIndex(row, column, parent_item->Children()[row].get());
+    } else {
```
lioncash: else isn't necessary here.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+        return static_cast<int>(parent_item->Children().size());
+    }
+}
+
+int WaitTreeModel::columnCount(const QModelIndex&) const {
+    return 1;
+}
+
+QVariant WaitTreeModel::data(const QModelIndex& index, int role) const {
+    if(!index.isValid()) return QVariant();
```
lioncash:
```cpp
if (!index.isValid())
    return {};
```

lioncash:
@neobrain Seeing as you know Qt quite well, are you able to spot anything that can be slightly improved?

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+    const auto& semaphore = static_cast<const Kernel::Semaphore&>(object);
+    list.push_back(std::make_unique<WaitTreeText>(tr("available count = %1")
+        .arg(semaphore.available_count)));
+    list.push_back(std::make_unique<WaitTreeText>(tr("max count = %1")
+        .arg(semaphore.max_count)));
+    return list;
+}
+
+WaitTreeTimer::WaitTreeTimer(const Kernel::Timer& object) :
+    WaitTreeWaitObject(object){
```
mailwl: missing whitespace after ")"

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -0,0 +1,431 @@
...
+}
+
+std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeItem::GetChildren() const {
+    return std::vector<std::unique_ptr<WaitTreeItem>>();
+}
+
+void WaitTreeItem::Expand() {
+    if (IsExpandable() && !expanded) {
+        children = GetChildren();
+        for(std::size_t i = 0; i < children.size(); ++i){
```
Kloen: `for (std::size_t i = 0; i < children.size(); ++i) {`

wwylele:
Addressed all comments above

Subv:
Note to self: Add the thread creation parameters to the displayed information after this is merged.

src/core/hle/svc.cpp
```diff
@@ -473,7 +473,7 @@ static ResultCode CreateThread(Handle* out_handle, s32 priority, u32 entry_point
         TSymbol symbol = Symbols::GetSymbol(entry_point);
         name = symbol.name;
     } else {
-        name = Common::StringFromFormat("unknown-%08x", entry_point);
+        name = Common::StringFromFormat("thread-%08x", Memory::Read32(arg+8));
```
Subv: Why was this changed to [arg+8]?
AFAIK only Nintendo code puts the real entry point of the thread in the argument, whereas homebrew will almost always use the entry_point as the actual function to run.

Subv:
Call for more reviews please, it's been more than 2 months already.

src/core/hle/svc.cpp
wwylele: See the discussion above, I followed @mailwl 's suggestion (without deep thinking)

mailwl: ah, i don't hink about homebrew... In titles entrypoint completly useless (and the same for all threads)

wwylele: So I revert this change? or any better idea?

Subv: Just revert it for now, see https://github.com/citra-emu/citra/pull/1652#issuecomment-225969325

neobrain:

2016-06-28T18:48:48Z: https://github.com/citra-emu/citra/compare/9cde5cbbd39d612601df85d8792eea48c96bdc3e...wwylele:55be1ecb7c4b83420d8cd1a916dd92a4b177aca4

LittleWhite-tb:
Really good work.

Some remarks on the interface :
The wait_tree shows the thread priority as follow : "49(current) / 49(normal). I would remove "current" and "normal" and remove spaces around '/' character.
Also "Waiting for one of the following objects" is maybe too much verbose. Just write "Waiting for :"
Why do you write "waited by no thread" when there is no thread waiting for the current. Just write something only if there is one thread (or many) waiting (same for mutex) ? (and, when there is one, just write "Waited by".
I suggest to add upper case letter for "Processor", Thread in "thread id", priority and last running ticks.

Do you think it is possible to stop this recursive behavior  of the tree?

You are using lot of unique_ptr (which is really neat of course), but, I thought that Qt did care about the memory of all the child of QObject class ?


Otherwise, LGTM.

mailwl:
@wwylele i've got random nullptr access while pausing a game

wwylele:
@mailwl I would appreciate it if you can run it with a debugger to see which line cause the error.

bunnei:
This seems fine and is no doubt useful. Sorry it took so long to merge - I'd say if anyone else has feedback that hasn't been addressed, feel free to open an issue (or pull request).

PR #1657, opened 2016-04-10
Pica: Remove geometry dumper (PICA_DUMP_GEOMETRY)

### LINKED ISSUE #1635: PICA_DUMP_GEOMETRY doesn't work anymore
`PICA_DUMP_GEOMETRY` has been broken for some time now. As the vertex cache is hit it won't be informed about all vertices passing through. This will break the output data.

---

The easy fix would be to disable the vertex cache when compiling with `PICA_DUMP_GEOMETRY`.
But that would possibly break some apps which depend on the cache as it actually emulates hardware behaviour (and shader state *could* depend on it) + that's just unclean code.

I think `PICA_DUMP_GEOMETRY` should be removed entirely to remove bloat code. Dumping geometry might be a fun feature for users but it serves no use in emulation or debugging the emulator.
(I personally used the `PICA_DUMP_GEOMETRY` define before to debug an issue but ended up rewriting almost all of it to suit my needs - I'm sure others would do the same. Also I wasn't aware that the VS debugger could show register states at that point)

The alternative would be to push for inclusion of such features in tools like apitrace or use existing tools such as renderdoc.
If users just want an easy way to export models this is probably better suited for a custom fork too.

If there should be a geometry dumper in citra it should probably be implemented elsewhere in the code and work by dumping the vertices independent of the renderer loop / vertex cache. Possibly some "Primitive inspector" which is closely tied to the VS debugger with connection to a GS debugger OR a debug callback for submitted polyons ("Fragment inspector"?). That would make it possible to export all attributes (by mapping them in the UI), making this a much more powerful tool.

JayFoxRox:
Resolves #1635 

`grep` for "Dump" and "DUMP_GEOM" found nothing related to this feature anymore.

Use RenderDoc or write your own dumper if you want to dump geometry.
(I'll probably come up with a new dumper at some point as I'm working on an improved shader debugger which fits the bill for this one)

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -249,10 +249,6 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
             const u16* index_address_16 = reinterpret_cast<const u16*>(index_address_8);
             bool index_u16 = index_info.format != 0;
 
-#if PICA_DUMP_GEOMETRY
-            DebugUtils::GeometryDumper geometry_dumper;
-            PrimitiveAssembler<DebugUtils::GeometryDumper::Vertex> dumping_primitive_assembler(regs.triangle_topology.Value());
-#endif
             PrimitiveAssembler<Shader::OutputVertex>& primitive_assembler = g_state.primitive_assembler;
 
             if (g_debug_context) {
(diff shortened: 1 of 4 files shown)


yuriks:
LGTM

Subv:
LGTM

Side note: We should probably move the LookupTexture function out of the debug_utils file.

PR #1659, opened 2016-04-11
CitraQt: Apply config at startup

JayFoxRox:
This is a quick-and-dirty hack (mostly untested too).

`master` doesn't load config with CitraQt at the moment so no matter what your settings say: you are starting with SW Rasterizer (and maybe also interpreter?). It also means you can't trust the config dialog because that claims you are running with HW Rendering.

it's probably not smart to do it this way but `master` is currently broken so we need a fix.

I also ordered the GDB calls because I'm pretty sure enabling the server will open a socket on the *previously* specified port. I'm not sure if GDB can even change the port while the server is running?

### DIFF
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -93,13 +93,12 @@ int main(int argc, char **argv) {
 
     log_filter.ParseFilterString(Settings::values.log_filter);
 
-    GDBStub::ToggleServer(use_gdbstub);
-    GDBStub::SetServerPort(gdb_port);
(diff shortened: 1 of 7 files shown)


Kloen:
@JayFoxRox a quick fix it's just left 
```
VideoCore::g_hw_renderer_enabled = values.use_hw_renderer;
VideoCore::g_shader_jit_enabled = values.use_shader_jit;
```

It should work fine, at least it did on #1540

PR #1660, opened 2016-04-12
FileUtil: Missing #include, Add const to IOFile methods

MerryMage:
`Common::UTF8ToTStr` call requires `#include "common/string_util.h"`

Also spotted a few things in IOFile while I was there.

### DIFF
--- a/src/common/file_util.h
+++ b/src/common/file_util.h
@@ -14,6 +14,10 @@
 
 #include "common/common_types.h"
 
+#ifdef _MSC_VER
+#include "common/string_util.h"
+#endif
+
 // User directory indices for GetUserPath
 enum {
     D_USER_IDX,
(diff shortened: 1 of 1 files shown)


src/common/file_util.h
```diff
@@ -259,8 +263,8 @@ class IOFile : public NonCopyable
     std::FILE* m_file;
     bool m_good;
 private:
-    IOFile(IOFile&);
-    IOFile& operator=(IOFile& other);
+    IOFile(IOFile&) = delete;
+    IOFile& operator=(IOFile& other) = delete;
```
lioncash: These can just be removed entirely since IOFile inherits from NonCopyable.

MerryMage: Amended

lioncash:
LGTM

bunnei:
LGTM

PR #1666, opened 2016-04-14
Thread: Correct Common::Barrier implementation

MerryMage:
`Barrier` as currently implemented is a one-shot synchronisation barrier that is unsafe to be reused. Here we make it reusable.

### DIFF
--- a/src/common/thread.h
+++ b/src/common/thread.h
@@ -30,8 +30,7 @@
 #   endif
 #endif
 
-namespace Common
-{
+namespace Common {
 
 int CurrentThreadId();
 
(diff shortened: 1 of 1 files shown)


src/common/thread.h
```diff
@@ -43,55 +42,55 @@ class Event {
...
-    std::condition_variable m_condvar;
-    std::mutex m_mutex;
+    std::condition_variable condvar;
+    std::mutex mutex;
 };
 
 class Barrier {
 public:
-    Barrier(size_t count) : m_count(count), m_waiting(0) {}
+    Barrier(size_t count) : count(count), waiting(0), generation(0) {}
```
lioncash: `count(count)` will cause variable shadowing warnings.

src/common/thread.h
```diff
@@ -43,55 +42,55 @@ class Event {
...
 private:
-    std::condition_variable m_condvar;
-    std::mutex m_mutex;
-    const size_t m_count;
-    size_t m_waiting;
+    std::condition_variable condvar;
+    std::mutex mutex;
+    const size_t count;
+    size_t waiting;
+    size_t generation; ///< Incremented once each time the barrier is used
```
lioncash: This can likely just be a regular comment, since it's an implementation detail and not a public part of Barrier's API.

MerryMage:
Addressed the above, thanks @lioncash.

bunnei:
LGTM

PR #241, opened 2014-12-02
Improve the loader a bit

linkmauve:
Clean up the loader to open the file only once and pass the IOFile around to functions requiring it, and add a real file detection function based on its magic.

PR #1667, opened 2016-04-14
loader: only decompress code section

wwylele:
Fix for `ReadIcon`, `ReadBanner` and `ReadLogo`, though they havn't been used yet

### DIFF
--- a/src/core/loader/ncch.cpp
+++ b/src/core/loader/ncch.cpp
@@ -174,7 +174,7 @@ ResultStatus AppLoader_NCCH::LoadSectionExeFS(const char* name, std::vector<u8>&
         return ResultStatus::Error;
 
     LOG_DEBUG(Loader, "%d sections:", kMaxSections);
-    // Iterate through the ExeFs archive until we find the .code file...
+    // Iterate through the ExeFs archive until we find a section with the specified name...
     for (unsigned section_number = 0; section_number < kMaxSections; section_number++) {
         const auto& section = exefs_header.section[section_number];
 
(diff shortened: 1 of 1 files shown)


linkmauve:
I remember a discussion about the opposite change, how is the current way broken?

See https://github.com/citra-emu/citra/pull/241#discussion-diff-21252674

src/core/loader/ncch.cpp
```diff
@@ -174,7 +174,7 @@ ResultStatus AppLoader_NCCH::LoadSectionExeFS(const char* name, std::vector<u8>&
         return ResultStatus::Error;
 
     LOG_DEBUG(Loader, "%d sections:", kMaxSections);
-    // Iterate through the ExeFs archive until we find the .code file...
+    // Iterate through the ExeFs archive until we find the section...
```
lioncash: This should indicate what section it's looking for.

wwylele:
@linkmauve four functions `ReadCode`, `ReadIcon`, `ReadBanner` and `ReadBanner` call `LoadSectionExeFS` to read corresponding section. However, `LoadSectionExeFS` assumes that all sections will be compressed if `is_compressed` is set, which is not true according to [this page about ExeFS](https://www.3dbrew.org/wiki/ExeFS) on 3dbrew - the `.code` section is the only section that can be optionally compressed. Currently it doesn't break anything since only `ReadCode` is used.

I found this when try to decode icons using `ReadIcon` but failed for all game with `is_compressed` set. With this patch, `ReadIcon` (and hopefully the other two functions) works correctly.

src/core/loader/ncch.cpp
```diff
@@ -174,7 +174,7 @@ ResultStatus AppLoader_NCCH::LoadSectionExeFS(const char* name, std::vector<u8>&
         return ResultStatus::Error;
 
     LOG_DEBUG(Loader, "%d sections:", kMaxSections);
-    // Iterate through the ExeFs archive until we find the .code file...
+    // Iterate through the ExeFs archive until we find the section with specified name...
```
lioncash: `a section with the specified name`

yuriks:
LGTM

bunnei:
LGTM

PR #1671, opened 2016-04-15
debug_utils: use std::make_unique for initializing PicaTrace

lioncash:
Noticed this while doing the IOFile changes.


MerryMage:
LGTM

PR #1672, opened 2016-04-15
Fix driver root identification on Windows

wwylele:
On Windoiws, `_wstat64` doesn't work with a driver root without slash, which cause failure to try select a driver root as game list. Also fixed C-style `StripTailDirSlashes` for `string::back` to work.

### DIFF
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -69,9 +69,10 @@ static void StripTailDirSlashes(std::string &fname)
 {
     if (fname.length() > 1)
     {
-        size_t i = fname.length() - 1;
-        while (fname[i] == DIR_SEP_CHR)
-            fname[i--] = '\0';
+        size_t i = fname.length();
+        while (i > 0 && fname[i - 1] == DIR_SEP_CHR)
+            --i;
+        fname.resize(i);
     }
     return;
 }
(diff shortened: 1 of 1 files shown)


src/common/file_util.cpp
```diff
@@ -85,6 +86,7 @@ bool Exists(const std::string &filename)
     StripTailDirSlashes(copy);
 
 #ifdef _WIN32
+    if (copy.back() == ':') copy += DIR_SEP_CHR; // Windows need a slash to identify a driver root
```
MerryMage: `Windows needs a slash to identify drive root`, ditto below.

src/common/file_util.cpp
```diff
@@ -85,6 +86,7 @@ bool Exists(const std::string &filename)
     StripTailDirSlashes(copy);
 
 #ifdef _WIN32
+    if (copy.back() == ':') copy += DIR_SEP_CHR; // Windows needs a slash to identify a driver root
```
lioncash: Note that if an empty string is passed, this is undefined behavior.

src/common/file_util.cpp
```diff
@@ -102,6 +104,7 @@ bool IsDirectory(const std::string &filename)
...
```
lioncash: same here

lioncash: same here

bunnei:
LGTM

Issue #1681, opened 2016-04-15
New Super Mario Bros 2 - Graphical bugs at HW Renderer (Alpha)


JohnGodgames:
It's a regression according to this older [video](https://www.youtube.com/watch?v=HQxsONkm_9c)

bug started with **Immediate-mode improvements #1461**
https://github.com/citra-emu/citra/pull/1461/commits

![nsmb 2 1](https://cloud.githubusercontent.com/assets/11453507/14573068/f08ec1c2-0352-11e6-8616-5e0d4ea974ff.JPG)

![nsmb 2 2](https://cloud.githubusercontent.com/assets/11453507/14573073/f5f481a6-0352-11e6-9068-4bb6f91bd7b0.JPG)

![nsmb 2 3](https://cloud.githubusercontent.com/assets/11453507/14573076/faf37248-0352-11e6-8061-b428c559996f.JPG)

KillzXGaming:
It's kinda odd. Animal crossing also has this issue however in older builds it does not along with this game.
Old build (around January)
http://i.imgur.com/tkf388x.png

http://i.imgur.com/7v9Bc2n.png

JayFoxRox:
I'll self-un-assign this for now because I'm too busy with other stuff at the moment.
I'd be happy if anyone else could look into this.

B3n30:
Solved by #2921

PR #1688, opened 2016-04-17
Pica: Reduce batching errors

JayFoxRox:
This ~~gets rid of some damage~~ hides some bugs which were seen after https://github.com/citra-emu/citra/commit/81004211dda74390c02973c37e89215f5ff8829b (bisected and reported by pcmaker - Thanks!)

~~`master` currently merges way too many triangles into one draw call. In MK7 it usually combines particle effects (smoke, coin glitter, speed effects) and HUD rendering. As one batch can only share the same draw state your kart sometimes spits out menu letters from it's exhause pipe and the smoke will look like the HUD item box.
This PR fixes that by reverting some changes from the commit mentioned above.~~
Correction: This might not even be the problem, I'll have to look at the code again

Keep in mind that this is not a perfect solution. All this does is splitting draw calls into seperate batches (like you'd expect). Immediate mode might still cause bad batching - I didn't check!
These are 2 seperate issues though, so I'll create an issue (soon) about possible remaining problems to remind us.

(Also, even if we can batch draw calls for GL we shouldn't do so in debug mode and report each draw call to the debugger)

---

Fixes bugs in MK7 and NSMB2 (#1681). I didn't test much further than that.

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -409,6 +409,12 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                                                           range.second, range.first);
             }
 
+            VideoCore::g_renderer->Rasterizer()->DrawTriangles();
+
+            if (g_debug_context) {
+                g_debug_context->OnEvent(DebugContext::Event::FinishedPrimitiveBatch, nullptr);
+            }
+
             break;
         }
 


KillzXGaming:
Fixes Animal Crossing New Leaf particles/footsteps too!

yuriks:
Afaik it shouldn't be possible for applications to change GPU state without flipping the configuration flag back on. Would be interested in a csv of a command trace from these games.

JohnGodgames:
**New Super Mario Bros 2**

fix batching 1

**Animal Crossing: New Leaf**

fix batching 2

JayFoxRox:
Apparently causes crashes in Johnny's Payday Panic. I'll have a look later and also update this issue with logs as requested by yuriks.

JayFoxRox:
I'll close this and probably open a new PR or re-open if I have a better idea (Probably depending on flush when clearing the vertex cache using GPUREG_VTX_FUNC)

PR #1621, opened 2016-03-31
Implement W-buffer and fix depth-mapping

JayFoxRox:
This PR [implements W-Buffers](https://www.mvps.org/directx/articles/using_w-buffers.htm) and fixes hardware depth-mapping (scale + offset).
The formulas for the depth-mapping were guessed after testing on hardware.

The results are still about 3 bits inprecise in the lower bits due to rounding and precision issues probably. This is still somewhat better than the current `master` solution which produces complete garbage for some mappings.

This PR implements this in the SW-Rasterizer and the HW-Renderer.

---

This PR was [tested using a HW-Test](https://github.com/JayFoxRox/3ds-tests/tree/master/depth-write) ([Planar results where top.zw = bottom.zw](https://docs.google.com/spreadsheets/d/1E8xvJIbj_MVygWtr4ELPuzOJPqEv4yGeHQuKG6kFmTc/edit#gid=0) & [Perspective results](https://docs.google.com/spreadsheets/d/13_m98KE29fwTgUowAM43-eh9nLw7l2TEXwtYVhclEpA/edit#gid=0)).
There are no known regressions while this might fix some games.

The interpolations weren't tested much yet (not at all for depth-scale and offset) but it's assumed this will work. More tests could be done, feel free to add them to the sheets I linked.

(I'd like to thank everyone who tested including but not limited to @MerryMage @JohnGodGames and @LokiX)

---

~~This fixes the depth formulas but not depth clearing / buffer access. There will be another PR soon (almost complete) to fix the other bugs caused by #1462 .~~ #1624

PR #1690, opened 2016-04-18
Pica: Implement texture type 3 (Projection2D)

### LINKED ISSUE #1664: Kokuga - Graphical bugs
Nothing useful in log...

**3DS**

kokuga 2

kokuga 4

**Citra**

kokuga 1

kokuga 3

JayFoxRox:
Texture 0 (according to 3DBrew - maybe also other TUs?) supports [various texture types](https://www.3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_TEXUNITi_PARAM).
`master` only implements a 2D texturing mode so far (type 0).

This PR implements projection (type 3) which does the lookup at (u,v)/w.
There are 2 seperate commits, one for the HW renderer and another one for the SW rasterizer.

This was not HW tested but this is what's commonly referred to as texture projection. The results also look fine in Kokuga.

This fixes #1664
(SW rasterizer still partially broken (green faded background displayed as solid green area) but it's probably due to missing texture filtering in the rasterizer [Outputting `fmodf(texcoord * texsize, 1.0)*255` makes parts of the fade visible])

(Note that GL `textureProj()` will divide by the last texcoord component and might be faster than dividing manually)

---

We probably have to refactor the rasterizer texture lookup to implement some other modes. For example: shadow lookups will probably be the same as projection lookup but the read depth value will then be compared against a fourth component (and the return value is the pixel coverage instead of a pixel color).
\- However, that is something for another PR (probably one which would also implement sw texture filtering?)


JohnGodgames:
Hmm I've found a regression on Ridge Racer 3D 

rr tex type 3 1

JayFoxRox:
The log in Ridge Racer shows that this is due to the game using a cubemap texture (probably to store reflections).

Cubemap texturing is handled by a 2D texture lookup in master (which is wrong).
This PR gives a warning (both SW/HW renderer) and also creates a buggy GLSL shader (for HW renderer) if an unsupported type (such as cubemap) is used.
As the SW renderer will only modify (u,v) that should still render as before (albeit very slow due to printf per pixel).

I knew about this before and thought not handling it at all is fine. However I realize that it might have been a bad idea. I'll probably add a fallback Texture2D lookup to the HW renderer.
We have no way of a deferred or one-shot error in the SW renderer so I probably won't change that.

JayFoxRox:
I've updated this (fallback to 2D texturing + debug message in rasterizer) but it still has the same issues as #1621 (padding causes shader regeneration which causes high memory footprint). I have more branches with that problem so I'll also try to look for a solution because I don't like the proposals so far.

Subv:
LGTM

hrydgard:
This should be fine now that #1621 is merged, just needs a rebase

JayFoxRox:
I'm waiting for #1621 to go through first because we'll need to rebase this afterwards and #1621 is more important.

JayFoxRox:
I've rebased this, should be good to go. No idea why travis was ignored..?

bunnei:
LGTM

PR #1691, opened 2016-04-19
SDL2 Frontend: Use argv[0], add a --version, and reorder options

linkmauve:


src/citra/citra.cpp
```diff
@@ -34,11 +35,12 @@
 #include "video_core/video_core.h"
 
 
-static void PrintHelp()
+static void PrintHelp(const char *argv0)
 {
-    std::cout << "Usage: citra [options] <filename>" << std::endl;
-    std::cout << "--help, -h            Display this information" << std::endl;
-    std::cout << "--gdbport, -g number  Enable gdb stub on port number" << std::endl;
+    std::cout << "Usage: " << argv0 << " [options] <filename>" << std::endl;
```
yuriks: Can you make help print the version too? I don't think there's any reason not to.

src/citra/citra.cpp
```diff
@@ -34,11 +35,12 @@
...
+    std::cout << "-g, --gdbport=NUMBER  Enable gdb stub on port NUMBER" << std::endl;
```
yuriks: You can make these just a single string (either raw string literal or by concatting several string literals)

yuriks:
LGTM

src/citra/citra.cpp
linkmauve: This isn’t something most programs do, from a relatively small panel of tested programs. This help is there to give the user the most commonly used arguments, --version on the other hand just gives the version and exits.

bunnei:
LGTM

PR #1692, opened 2016-04-19
Memory: Remove most usages of GetPointer

Subv:
Most usages were replaced by the new functions Memory::ReadBlock and Memory::WriteBlock, which take into account segmentation of memory pages.

There are 14 usages of GetPointer left, down from the 54 usages we had before.
Original work by @MerryMage.

### DIFF
--- a/src/citra_qt/debugger/callstack.cpp
+++ b/src/citra_qt/debugger/callstack.cpp
@@ -37,10 +37,13 @@ void CallstackWidget::OnDebugModeEntered()
     int counter = 0;
     for (u32 addr = 0x10000000; addr >= sp; addr -= 4)
     {
+        if (!Memory::IsValidVirtualAddress(addr))
+            break;
+
         const u32 ret_addr = Memory::Read32(addr);
         const u32 call_addr = ret_addr - 4; //get call address???
 
-        if (Memory::GetPointer(call_addr) == nullptr)
+        if (!Memory::IsValidVirtualAddress(call_addr))
             break;
 
         /* TODO (mattvail) clean me, move to debugger interface */
--- a/src/core/file_sys/archive_backend.cpp
+++ b/src/core/file_sys/archive_backend.cpp
@@ -19,22 +19,22 @@ Path::Path(LowPathType type, u32 size, u32 pointer) : type(type) {
     switch (type) {
     case Binary:
     {
-        u8* data = Memory::GetPointer(pointer);
-        binary = std::vector<u8>(data, data + size);
+        binary.resize(size);
+        Memory::ReadBlock(pointer, binary.data(), binary.size());
         break;
     }
 
     case Char:
     {
-        const char* data = reinterpret_cast<const char*>(Memory::GetPointer(pointer));
-        string = std::string(data, size - 1); // Data is always null-terminated.
+        string.resize(size - 1); // Data is always null-terminated.
+        Memory::ReadBlock(pointer, &string[0], string.size());
         break;
     }
 
     case Wchar:
     {
-        const char16_t* data = reinterpret_cast<const char16_t*>(Memory::GetPointer(pointer));
-        u16str = std::u16string(data, size/2 - 1); // Data is always null-terminated.
+        u16str.resize(size / 2 - 1); // Data is always null-terminated.
+        Memory::ReadBlock(pointer, &u16str[0], u16str.size() * sizeof(char16_t));
         break;
     }
 
--- a/src/core/hle/service/apt/apt.h
+++ b/src/core/hle/service/apt/apt.h
@@ -20,16 +20,14 @@ struct MessageParameter {
     u32 sender_id = 0;
     u32 destination_id = 0;
     u32 signal = 0;
-    u32 buffer_size = 0;
     Kernel::SharedPtr<Kernel::Object> object = nullptr;
-    u8* data = nullptr;
+    std::vector<u8> buffer;
 };
 
 /// Holds information about the parameters used in StartLibraryApplet
 struct AppletStartupParameter {
-    u32 buffer_size = 0;
     Kernel::SharedPtr<Kernel::Object> object = nullptr;
-    u8* data = nullptr;
+    std::vector<u8> buffer;
 };
 
 /// Used by the application to pass information about the current framebuffer to applets.
--- a/src/core/hle/service/dsp_dsp.cpp
+++ b/src/core/hle/service/dsp_dsp.cpp
@@ -140,12 +140,15 @@ static void LoadComponent(Service::Interface* self) {
 
     // TODO(bunnei): Implement real DSP firmware loading
 
-    ASSERT(Memory::GetPointer(buffer) != nullptr);
-    ASSERT(size > 0x37C);
+    ASSERT(Memory::IsValidVirtualAddress(buffer));
 
-    LOG_INFO(Service_DSP, "Firmware hash: %#" PRIx64, Common::ComputeHash64(Memory::GetPointer(buffer), size));
+    std::vector<u8> component_data(size);
+    Memory::ReadBlock(buffer, component_data.data(), component_data.size());
+
+    LOG_INFO(Service_DSP, "Firmware hash: %#" PRIx64, Common::ComputeHash64(component_data.data(), component_data.size()));
     // Some versions of the firmware have the location of DSP structures listed here.
-    LOG_INFO(Service_DSP, "Structures hash: %#" PRIx64, Common::ComputeHash64(Memory::GetPointer(buffer) + 0x340, 60));
+    ASSERT(size > 0x37C);
+    LOG_INFO(Service_DSP, "Structures hash: %#" PRIx64, Common::ComputeHash64(component_data.data() + 0x340, 60));
 
     LOG_WARNING(Service_DSP, "(STUBBED) called size=0x%X, prog_mask=0x%08X, data_mask=0x%08X, buffer=0x%08X",
                 size, prog_mask, data_mask, buffer);
@@ -285,7 +288,7 @@ static void WriteProcessPipe(Service::Interface* self) {
         return;
     }
 
-    ASSERT_MSG(Memory::GetPointer(buffer) != nullptr, "Invalid Buffer: pipe=%u, size=0x%X, buffer=0x%08X", pipe_index, size, buffer);
+    ASSERT_MSG(Memory::IsValidVirtualAddress(buffer), "Invalid Buffer: pipe=%u, size=0x%X, buffer=0x%08X", pipe, size, buffer);
 
     std::vector<u8> message(size);
     for (u32 i = 0; i < size; i++) {
@@ -324,7 +327,7 @@ static void ReadPipeIfPossible(Service::Interface* self) {
 
     DSP::HLE::DspPipe pipe = static_cast<DSP::HLE::DspPipe>(pipe_index);
 
-    ASSERT_MSG(Memory::GetPointer(addr) != nullptr, "Invalid addr: pipe=%u, unknown=0x%08X, size=0x%X, buffer=0x%08X", pipe_index, unknown, size, addr);
+    ASSERT_MSG(Memory::IsValidVirtualAddress(addr), "Invalid addr: pipe=0x%08X, unknown=0x%08X, size=0x%X, buffer=0x%08X", pipe, unknown, size, addr);
 
     cmd_buff[0] = IPC::MakeHeader(0x10, 1, 2);
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
@@ -364,7 +367,7 @@ static void ReadPipe(Service::Interface* self) {
 
     DSP::HLE::DspPipe pipe = static_cast<DSP::HLE::DspPipe>(pipe_index);
 
-    ASSERT_MSG(Memory::GetPointer(addr) != nullptr, "Invalid addr: pipe=%u, unknown=0x%08X, size=0x%X, buffer=0x%08X", pipe_index, unknown, size, addr);
+    ASSERT_MSG(Memory::IsValidVirtualAddress(addr), "Invalid addr: pipe=0x%08X, unknown=0x%08X, size=0x%X, buffer=0x%08X", pipe, unknown, size, addr);
 
     if (DSP::HLE::GetPipeReadableSize(pipe) >= size) {
         std::vector<u8> response = DSP::HLE::PipeRead(pipe, size);
--- a/src/core/hle/service/fs/archive.cpp
+++ b/src/core/hle/service/fs/archive.cpp
@@ -116,7 +116,6 @@ ResultVal<bool> File::SyncRequest() {
             }
             Memory::WriteBlock(address, data.data(), *read);
             cmd_buff[2] = static_cast<u32>(*read);
-            Memory::RasterizerFlushAndInvalidateRegion(Memory::VirtualToPhysicalAddress(address), length);
             break;
         }
 
--- a/src/core/hle/service/fs/archive.cpp
+++ b/src/core/hle/service/fs/archive.cpp
@@ -108,11 +108,13 @@ ResultVal<bool> File::SyncRequest() {
                           offset, length, backend->GetSize());
             }
 
-            ResultVal<size_t> read = backend->Read(offset, length, Memory::GetPointer(address));
+            std::vector<u8> data(length);
+            ResultVal<size_t> read = backend->Read(offset, data.size(), data.data());
             if (read.Failed()) {
                 cmd_buff[1] = read.Code().raw;
                 return read.Code();
             }
+            Memory::WriteBlock(address, data.data(), *read);
             cmd_buff[2] = static_cast<u32>(*read);
             Memory::RasterizerFlushAndInvalidateRegion(Memory::VirtualToPhysicalAddress(address), length);
             break;
@@ -128,7 +130,9 @@ ResultVal<bool> File::SyncRequest() {
             LOG_TRACE(Service_FS, "Write %s %s: offset=0x%llx length=%d address=0x%x, flush=0x%x",
                       GetTypeName().c_str(), GetName().c_str(), offset, length, address, flush);
 
-            ResultVal<size_t> written = backend->Write(offset, length, flush != 0, Memory::GetPointer(address));
+            std::vector<u8> data(length);
+            Memory::ReadBlock(address, data.data(), data.size());
+            ResultVal<size_t> written = backend->Write(offset, data.size(), flush != 0, data.data());
             if (written.Failed()) {
                 cmd_buff[1] = written.Code().raw;
                 return written.Code();
@@ -216,12 +220,14 @@ ResultVal<bool> Directory::SyncRequest() {
         {
             u32 count = cmd_buff[1];
             u32 address = cmd_buff[3];
-            auto entries = reinterpret_cast<FileSys::Entry*>(Memory::GetPointer(address));
+            std::vector<FileSys::Entry> entries(count);
             LOG_TRACE(Service_FS, "Read %s %s: count=%d",
                 GetTypeName().c_str(), GetName().c_str(), count);
 
             // Number of entries actually read
-            cmd_buff[2] = backend->Read(count, entries);
+            u32 read = backend->Read(entries.size(), entries.data());
+            cmd_buff[2] = read;
+            Memory::WriteBlock(address, entries.data(), read * sizeof(FileSys::Entry));
             break;
         }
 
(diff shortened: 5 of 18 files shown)


src/core/memory.h
```diff
@@ -126,7 +129,10 @@ void Write16(VAddr addr, u16 data);
 void Write32(VAddr addr, u32 data);
 void Write64(VAddr addr, u64 data);
 
-void WriteBlock(VAddr addr, const u8* data, size_t size);
+void ReadBlock(const VAddr src_addr, u8* dest_buffer, size_t size);
+void WriteBlock(const VAddr dest_addr, const u8* src_buffer, size_t size);
```
mailwl: is it possible to change `const u8*` to `const void*` to let use pointer to any structs without casting?

yuriks: +1

src/core/memory.cpp
```diff
@@ -173,6 +173,24 @@ void Write(const VAddr vaddr, const T data) {
     }
 }
 
+bool IsValidVirtualAddress(const VAddr vaddr) {
+    u8* page_pointer = current_page_table->pointers[vaddr >> PAGE_BITS];
```
lioncash: This should be a const pointer.

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -66,14 +66,26 @@ static inline InterruptRelayQueue* GetInterruptRelayQueue(u32 thread_id) {
...
 
 /**
+ * Writes a single GSP GPU hardware registers with a single u32 value
+ * (For internal use.)
+ *
+ * @param base_address The address of the register in question
+ * @param data Data to be written
+ */
+static void WriteSingleHWReg(u32 base_address, u32 data) {
+    DEBUG_ASSERT_MSG(base_address & 3 == 0 && base_address < 0x420000, "Write address out of range of misaligned");
```
lioncash: `out of range or misaligned`

src/core/hle/service/apt/apt.h
```diff
@@ -19,16 +19,14 @@ struct MessageParameter {
     u32 sender_id = 0;
     u32 destination_id = 0;
     u32 signal = 0;
-    u32 buffer_size = 0;
     Kernel::SharedPtr<Kernel::Object> object = nullptr;
-    u8* data = nullptr;
```
lioncash: `= std::vector<u8>(0);` isn't necessary

src/core/hle/service/apt/apt.h
```diff
@@ -19,16 +19,14 @@ struct MessageParameter {
...
+    std::vector<u8> buffer = std::vector<u8>(0);
```
lioncash: = std::vector<u8>(0); isn't necessary

src/core/hle/service/apt/apt.h
```diff
@@ -19,16 +19,14 @@ struct MessageParameter {
...
 };
 
 /// Holds information about the parameters used in StartLibraryApplet
 struct AppletStartupParameter {
-    u32 buffer_size = 0;
     Kernel::SharedPtr<Kernel::Object> object = nullptr;
-    u8* data = nullptr;
+    std::vector<u8> buffer = std::vector<u8>(0);
```
lioncash: ditto

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -93,12 +93,15 @@ static void LoadComponent(Service::Interface* self) {
 
     // TODO(bunnei): Implement real DSP firmware loading
 
-    ASSERT(Memory::GetPointer(buffer) != nullptr);
-    ASSERT(size > 0x37C);
+    ASSERT(!Memory::IsValidVirtualAddress(buffer));
```
MerryMage: All three instances of `ASSERT(!Memory::IsValidVirtualAddress` should be `ASSERT(Memory::IsValidVirtualAddress` in this file (sorry, my fault).

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -211,7 +214,7 @@ static void WriteProcessPipe(Service::Interface* self) {
     u32 buffer = cmd_buff[4];
 
     ASSERT_MSG(IPC::StaticBufferDesc(size, 1) == cmd_buff[3], "IPC static buffer descriptor failed validation (0x%X). pipe=%u, size=0x%X, buffer=0x%08X", cmd_buff[3], pipe, size, buffer);
-    ASSERT_MSG(Memory::GetPointer(buffer) != nullptr, "Invalid Buffer: pipe=%u, size=0x%X, buffer=0x%08X", pipe, size, buffer);
+    ASSERT_MSG(!Memory::IsValidVirtualAddress(buffer), "Invalid Buffer: pipe=%u, size=0x%X, buffer=0x%08X", pipe, size, buffer);
```
MerryMage: ditto

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -248,7 +251,7 @@ static void ReadPipeIfPossible(Service::Interface* self) {
     u32 size = cmd_buff[3] & 0xFFFF; // Lower 16 bits are size
     VAddr addr = cmd_buff[0x41];
 
-    ASSERT_MSG(Memory::GetPointer(addr) != nullptr, "Invalid addr: pipe=0x%08X, unknown=0x%08X, size=0x%X, buffer=0x%08X", pipe, unknown, size, addr);
+    ASSERT_MSG(!Memory::IsValidVirtualAddress(addr), "Invalid addr: pipe=0x%08X, unknown=0x%08X, size=0x%X, buffer=0x%08X", pipe, unknown, size, addr);
```
MerryMage: ditto

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -283,7 +286,7 @@ static void ReadPipe(Service::Interface* self) {
...
```
MerryMage: ditto

src/core/memory.cpp
```diff
@@ -203,6 +222,45 @@ u64 Read64(const VAddr addr) {
...
 }
 
+void ReadBlock(const VAddr src_addr, u8* dest_buffer, const size_t size) {
+    size_t remaining_size = size;
+    size_t page_index = src_addr >> PAGE_BITS;
+    size_t page_offset = src_addr & PAGE_MASK;
+
+    while (remaining_size > 0) {
+        const size_t copy_amount = std::min(PAGE_SIZE - page_offset, remaining_size);
+        const VAddr current_vaddr = page_index << PAGE_BITS + page_offset;
```
MerryMage: Incorrect operator precedence: `(page_index << PAGE_BITS) + page_offset` (and all other instances in this file).

src/core/memory.cpp
```diff
@@ -219,9 +277,119 @@ void Write64(const VAddr addr, const u64 data) {
...
+
+        page_index++;
+        page_offset = 0;
+        remaining_size -= copy_amount;
+    }
+}
+
+void CopyBlock(VAddr dest_addr, const VAddr src_addr, const size_t size) {
+    size_t remaining_size = size;
+    size_t page_index = dest_addr >> PAGE_BITS;
```
MerryMage: `size_t page_index = src_addr >> PAGE_BITS;`

src/core/memory.cpp
```diff
@@ -219,9 +277,119 @@ void Write64(const VAddr addr, const u64 data) {
...
+    size_t page_offset = dest_addr & PAGE_MASK;
```
MerryMage: size_t page_offset = src_addr & PAGE_MASK;

src/core/memory.cpp
```diff
@@ -219,9 +277,119 @@ void Write64(const VAddr addr, const u64 data) {
...
+
+    while (remaining_size > 0) {
+        const size_t copy_amount = std::min(PAGE_SIZE - page_offset, remaining_size);
+        const VAddr current_vaddr = page_index << PAGE_BITS + page_offset;
```
MerryMage: `(page_index << PAGE_BITS)`

src/core/memory.cpp
```diff
@@ -219,9 +277,119 @@ void Write64(const VAddr addr, const u64 data) {
...
+            break;
+        }
+        default:
+            UNREACHABLE();
+        }
+
+        page_index++;
+        page_offset = 0;
+        dest_addr += copy_amount;
+        remaining_size -= copy_amount;
```
MerryMage: Missing a line: `src_addr += copy_amount;`

src/core/memory.cpp
```diff
@@ -219,9 +277,119 @@ void Write64(const VAddr addr, const u64 data) {
...
+}
+
+void ZeroBlock(const VAddr dest_addr, const size_t size) {
+    size_t remaining_size = size;
+    size_t page_index = dest_addr >> PAGE_BITS;
+    size_t page_offset = dest_addr & PAGE_MASK;
+
+    while (remaining_size > 0) {
+        const size_t copy_amount = std::min(PAGE_SIZE - page_offset, remaining_size);
+        const VAddr current_vaddr = page_index << PAGE_BITS + page_offset;
```
MerryMage: `(page_index << PAGE_BITS)`

src/core/memory.cpp
```diff
@@ -219,9 +277,119 @@ void Write64(const VAddr addr, const u64 data) {
...
-    for (u32 offset = 0; offset < size; offset++) {
-        Write8(addr + offset, data[offset]);
+void WriteBlock(const VAddr dest_addr, const u8* src_buffer, const size_t size) {
+    size_t remaining_size = size;
+    size_t page_index = dest_addr >> PAGE_BITS;
+    size_t page_offset = dest_addr & PAGE_MASK;
+
+    while (remaining_size > 0) {
+        const size_t copy_amount = std::min(PAGE_SIZE - page_offset, remaining_size);
+        const VAddr current_vaddr = page_index << PAGE_BITS + page_offset;
```
MerryMage: `(page_index << PAGE_BITS)`

JamePeng:
Is that mean，all of GetPointer should be replaced？

MerryMage:
LGTM (another pair of eyes would be good though)

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -131,14 +143,17 @@ static ResultCode WriteHWRegsWithMask(u32 base_address, u32 size_in_bytes, const
...
 
+                u32 data = Memory::Read32(data_vaddr);
+                u32 mask = Memory::Read32(masks_vaddr);
+
                 // Update the current value of the register only for set mask bits
-                reg_value = (reg_value & ~*masks) | (*data | *masks);
+                reg_value = (reg_value & ~mask) | (data | mask);
 
-                HW::Write<u32>(reg_address, reg_value);
+                WriteSingleHWReg(reg_address, reg_value);
```
wwylele: At line 141 
`const u32 reg_address = base_address + REGS_BEGIN;`
and in `WriteSingleHWReg` there is `base_address + REGS_BEGIN`
duplicated?

src/core/memory.cpp
```diff
@@ -203,6 +222,45 @@ u64 Read64(const VAddr addr) {
...
+        switch (current_page_table->attributes[page_index]) {
+        case PageType::Unmapped: {
+            LOG_ERROR(HW_Memory, "unmapped ReadBlock @ 0x%08X (start address = 0x%08X, size = %zu)", current_vaddr, src_addr, size);
+            std::memset(dest_buffer, 0, copy_amount);
+            break;
+        }
+        case PageType::Memory: {
+            DEBUG_ASSERT(current_page_table->pointers[page_index]);
+
+            u8* src_ptr = current_page_table->pointers[page_index] + page_offset;
```
lioncash: This should be a const pointer.

src/core/memory.cpp
```diff
@@ -219,9 +277,120 @@ void Write64(const VAddr addr, const u64 data) {
...
+            DEBUG_ASSERT(current_page_table->pointers[page_index]);
+
+            u8* dest_ptr = current_page_table->pointers[page_index] + page_offset;
+            std::memset(dest_ptr, 0, copy_amount);
+            break;
+        }
+        case PageType::Special: {
+            DEBUG_ASSERT(GetMMIOHandler(current_vaddr));
+
+            static std::array<u8, PAGE_SIZE> zeros = {};
```
lioncash: Is there a better way to do this that doesn't involve having static lifetimes in our memory routines?

src/core/hle/service/gsp_gpu.cpp
Subv: Indeed, good catch!

src/core/file_sys/archive_backend.cpp
```diff
@@ -19,22 +19,24 @@ Path::Path(LowPathType type, u32 size, u32 pointer) : type(type) {
...
+        string.assign(data.data(), data.size());
         break;
     }
 
     case Wchar:
     {
-        const char16_t* data = reinterpret_cast<const char16_t*>(Memory::GetPointer(pointer));
-        u16str = std::u16string(data, size/2 - 1); // Data is always null-terminated.
+        std::vector<char16_t> data(size / 2 - 1); // Data is always null-terminated.
+        Memory::ReadBlock(pointer, data.data(), data.size() * sizeof(char16_t));
```
archshift: Strings (as of C++11, IIRC) are always contiguous in memory, so you could just read right into a `u16_string`.

Subv: Keep in mind that std::string::data returns a const pointer

archshift: `&str[0]` to the rescue

wwylele:
With this PR, whenever I pause the game, citra freezes and spams log with:
`HW.Memory <Error> core\memory.cpp:Memory::Read:137: unmapped Read3
2 @ 0xFFFFFFFC`
I encountered this issue in most of my games. Some games doesn't freeze citra, but still shows several lines of log like that

src/citra_qt/debugger/callstack.cpp
```diff
@@ -40,7 +40,7 @@ void CallstackWidget::OnDebugModeEntered()
         const u32 ret_addr = Memory::Read32(addr);
         const u32 call_addr = ret_addr - 4; //get call address???
 
-        if (Memory::GetPointer(call_addr) == nullptr)
```
wwylele: `!Memory::IsValidVirtualAddress(call_addr)`

src/citra_qt/debugger/callstack.cpp
```diff
@@ -40,7 +40,7 @@ void CallstackWidget::OnDebugModeEntered()
...
+        if (!Memory::IsValidVirtualAddress(addr))
```
wwylele: `(call_addr)`

wwylele: ↑sorry that is probably not a correct fix.

src/core/memory.cpp
```diff
@@ -173,6 +173,24 @@ void Write(const VAddr vaddr, const T data) {
     }
 }
 
+bool IsValidVirtualAddress(const VAddr vaddr) {
+    const u8* page_pointer = current_page_table->pointers[vaddr >> PAGE_BITS];
+    if (page_pointer) {
+        return true;
+    }
+
+    MMIORegionPointer mmio_region = GetMMIOHandler(vaddr);
```
wwylele: See [memory.cpp#L118](https://github.com/Subv/citra/blob/rm_getpointer2/src/core/memory.cpp#L118), if `vaddr` is invalid, `GetMMIOHandler` will `ASSERT` fail and crash.

MerryMage: Ah, should probably wrap in an `if(current_page_table->attrib[vaddr >> PAGE_BITS] == Special)` in that case

JayFoxRox:
Do you have data on performance of this change? My first thought was that this could cause serious slowdown.

Also is there a guideline when to use `GetPointer` and when to use these new functions?
You said there are 14 uses of the old style - why?

(I think I'd have implemented this similarily but also extended `GetPointer` by a another output so you always know how many bytes of valid mapping still follow, then slowly checked perf for each use and replaced it if it simplifies the code)

Subv:
I did some tests replacing some code snippets with ReadBlock/WriteBlock and it turns out that compilers (or at least MSVC, Clang, and GCC) are pretty good at optimizing this pattern (Checked the before and after disassembly by hand), they won't create a temporary buffer in any of the cases I tested and will directly access the memory. That means there should be little to no performance degradation coming from this change.

Subv:
Pushed an update, I'm not sure how to address @lioncash's concern, any suggestions?

Subv:
Rebased

src/core/memory.cpp
```diff
@@ -359,9 +419,120 @@ void Write64(const VAddr addr, const u64 data) {
...
+            break;
+        }
+        case PageType::Special: {
+            DEBUG_ASSERT(GetMMIOHandler(current_vaddr));
+
+            GetMMIOHandler(current_vaddr)->WriteBlock(current_vaddr, src_buffer, copy_amount);
+            break;
+        }
+        default:
+            UNREACHABLE();
```
wwylele: My game reached here with `PageType::RasterizerCachedMemory`

tfarley: The block functions should also handle the `PageType::RasterizerCachedMemory` and `PageType::RasterizerCachedSpecial` cases.

Each case should have the same functionality as their non-cached counterpart handler, except:
- `ReadBlock` should also call `RasterizerFlushRegion()` before executing.
- `WriteBlock` and `ZeroBlock` should also call `RasterizerFlushAndInvalidateRegion()` before executing.
- All of those cases should also use `GetPointerFromVMA()` as the actual memory address, instead of `current_page_table->pointers[]` as that will be null.

Take a look at how I did those for the templated `Read()` and `Write()`.

Subv:
Updated and rebased

src/core/hle/service/fs/archive.cpp
```diff
@@ -108,11 +108,13 @@ ResultVal<bool> File::SyncRequest() {
...
-            ResultVal<size_t> read = backend->Read(offset, length, Memory::GetPointer(address));
+            std::vector<u8> data(length);
+            ResultVal<size_t> read = backend->Read(offset, data.size(), data.data());
             if (read.Failed()) {
                 cmd_buff[1] = read.Code().raw;
                 return read.Code();
             }
+            Memory::WriteBlock(address, data.data(), data.size());
             cmd_buff[2] = static_cast<u32>(*read);
             Memory::RasterizerFlushAndInvalidateRegion(Memory::VirtualToPhysicalAddress(address), length);
```
wwylele: This line can be removed because`WriteBlock` already did it.

Subv:
Addressed all comments and rebased.

wwylele:
LGTM

src/core/memory.cpp
Subv: Addressed

bunnei:
This breaks A Link Between Worlds

src/core/hle/service/fs/archive.cpp
```diff
@@ -108,13 +108,14 @@ ResultVal<bool> File::SyncRequest() {
...
             }
 
-            ResultVal<size_t> read = backend->Read(offset, length, Memory::GetPointer(address));
+            std::vector<u8> data(length);
+            ResultVal<size_t> read = backend->Read(offset, data.size(), data.data());
             if (read.Failed()) {
                 cmd_buff[1] = read.Code().raw;
                 return read.Code();
             }
+            Memory::WriteBlock(address, data.data(), data.size());
```
wwylele: `Memory::WriteBlock(address, data.data(), *read);`  This is what causes Link between worlds broken. Sorry that didn't notice this on last reviewing. Probably there are similar problems elsewhere.

src/core/hle/service/fs/archive.cpp
```diff
@@ -216,12 +219,13 @@ ResultVal<bool> Directory::SyncRequest() {
...
             u32 address = cmd_buff[3];
-            auto entries = reinterpret_cast<FileSys::Entry*>(Memory::GetPointer(address));
+            std::vector<FileSys::Entry> entries(count);
             LOG_TRACE(Service_FS, "Read %s %s: count=%d",
                 GetTypeName().c_str(), GetName().c_str(), count);
 
             // Number of entries actually read
-            cmd_buff[2] = backend->Read(count, entries);
+            cmd_buff[2] = backend->Read(entries.size(), entries.data());
+            Memory::WriteBlock(address, entries.data(), entries.size() * sizeof(FileSys::Entry));
```
wwylele: Also here. Only write data in the length that is actually readed.

Subv:
Addressed

bunnei:
Lgtm

Subv:
\o/

PR #1689, opened 2016-04-18
Kernel: Implemented shared memory.

Subv:
This PR corrects the implementation of Shared Memory thorough the codebase.
SharedMemories created by HLE processes are backed by memory located in the BASE linear heap region (Unless stated otherwise by the call to Create), as per the real kernel.

With this we can get rid of some very nasty hacks, like the APT Shared Font being located at a fixed address. This PR implements a function to automatically relocate the dumped shared font from the 0x18000000 address to whatever address the allocator places it at. This fixes an issue where we were placing the shared font in the middle of the APPLICATION memory region (Possibly breaking games that use that part of the region).

All error codes were verified by reverse engineering unless otherwise stated.
There are some sanity checks that can not be performed unless we give each HLE service its own Process, but that's a discussion for another time.

Closes #1603
Closes #1547
Closes #1591
Closes #1318
Closes #1596
Closes #1497

PR #1695, opened 2016-04-19
Kernel/Threads: Dynamically allocate the TLS region for threads.

Subv:
The memory pages are allocated from the BASE region as per the real kernel.

Each thread gets a 0x200-byte area from the 0x1000-sized page, when all 8 thread slots in a single page are used up, the kernel allocates a new page to hold another 8 entries.

This is consistent with what the real kernel does.
With this and #1689, only the VRAM mapping remains in the table of fixed legacy mappings.

### DIFF
--- a/src/core/hle/kernel/process.h
+++ b/src/core/hle/kernel/process.h
@@ -140,8 +140,11 @@ public:
 
     MemoryRegionInfo* memory_region = nullptr;
 
-    /// Bitmask of the used TLS slots
-    std::bitset<300> used_tls_slots;
+    /// The Thread Local Storage area is allocated as processes create threads,
+    /// each TLS area is 0x200 bytes, so one page (0x1000) is split up in 8 parts, and each part
+    /// holds the TLS for a specific thread. This vector contains which parts are in use for each page as a bitmask.
+    /// This vector will grow as more pages are allocated for new threads.
+    std::vector<std::bitset<8>> tls_slots;
 
     VAddr GetLinearHeapAreaAddress() const;
     VAddr GetLinearHeapBase() const;
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -117,9 +117,10 @@ void Thread::Stop() {
     }
     wait_objects.clear();
 
-    Kernel::g_current_process->used_tls_slots[tls_index] = false;
-    g_current_process->misc_memory_used -= Memory::TLS_ENTRY_SIZE;
-    g_current_process->memory_region->used -= Memory::TLS_ENTRY_SIZE;
+    // Mark the TLS slot in the thread's page as free.
+    u32 tls_page = (tls_address - Memory::TLS_AREA_VADDR) / Memory::PAGE_SIZE;
+    u32 tls_slot = ((tls_address - Memory::TLS_AREA_VADDR) % Memory::PAGE_SIZE) / Memory::TLS_ENTRY_SIZE;
+    Kernel::g_current_process->tls_slots[tls_page].reset(tls_slot);
 
     HLE::Reschedule(__func__);
 }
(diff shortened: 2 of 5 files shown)


src/core/hle/kernel/thread.cpp
```diff
@@ -403,22 +406,66 @@ ResultVal<SharedPtr<Thread>> Thread::Create(std::string name, VAddr entry_point,
...
-    auto& used_tls_slots = Kernel::g_current_process->used_tls_slots;
-    for (unsigned int i = 0; i < used_tls_slots.size(); ++i) {
-        if (used_tls_slots[i] == false) {
-            thread->tls_index = i;
-            used_tls_slots[i] = true;
-            break;
+    auto& tls_slots = Kernel::g_current_process->tls_slots;
+    bool needs_allocation = true;
+    u32 available_page; // Which allocated page has free space
+    u8 available_slot; // Which slot within the page is free
```
lioncash: This should likely be a u32, it's only used as an index.

src/core/hle/kernel/thread.cpp
```diff
@@ -403,22 +406,66 @@ ResultVal<SharedPtr<Thread>> Thread::Create(std::string name, VAddr entry_point,
...
+
+    // Iterate over all the allocated pages, and try to find one where not all slots are used.
+    for (unsigned page = 0; page < tls_slots.size(); ++page) {
```
lioncash: This should likely be its own function which returns a tuple or pair or something. This doesn't particularly have anything to do with the creation of the thread itself

src/core/hle/kernel/process.h
```diff
@@ -140,8 +140,11 @@ class Process final : public Object {
 
     MemoryRegionInfo* memory_region = nullptr;
 
-    /// Bitmask of the used TLS slots
-    std::bitset<300> used_tls_slots;
+    /// The Thread Local Storage area is allocated as processes create threads,
+    /// each TLS area is 0x200 bytes, so one page (0x1000) is split up in 8 parts, and each part
+    /// holds the TLS for a specific thread. This vector contains which parts are in use for each page as a bitmask.
+    /// This vector will grow as more pages are allocated for new threads.
+    std::vector<u8> tls_slots;
```
wwylele: use bitset<8> instead of u8 for better readability?

Subv:
Updated

src/core/hle/kernel/thread.cpp
```diff
@@ -366,6 +367,30 @@ static void DebugThreadQueue() {
...
+ * @param tls_slots The TLS page array of the thread's owner process.
+ * Returns a tuple of (page, slot, alloc_needed) where:
+ * page: The index of the first allocated TLS page that has free slots.
+ * slot: The index of the first free slot in the indicated page.
+ * alloc_needed: Whether there's a need to allocate a new TLS page (All pages are full).
+ */
+std::tuple<u32, u32, bool> GetFreeThreadLocalSlot(std::vector<std::bitset<8>>& tls_slots) {
+    // Iterate over all the allocated pages, and try to find one where not all slots are used.
+    for (unsigned page = 0; page < tls_slots.size(); ++page) {
+        if (tls_slots[page] != 0xFF) {
```
wwylele: `!tls_slots[page].all()`

src/core/hle/kernel/thread.cpp
```diff
@@ -366,6 +367,30 @@ static void DebugThreadQueue() {
...
+            // We found a page with at least one free slot, find which slot it is
+            for (unsigned slot = 0; slot < 8; ++slot) {
```
JayFoxRox: use `tls_slots[page].size()` ?

src/core/hle/kernel/thread.cpp
JayFoxRox: Maybe something like `const auto& page_tls_slots = tls_slots[page];` ?

Subv:
Updated

bunnei:
Code looks find to me. I think this has marinated long enough :)

src/core/hle/kernel/thread.cpp
```diff
@@ -117,9 +117,10 @@ void Thread::Stop() {
     }
     wait_objects.clear();
 
-    Kernel::g_current_process->used_tls_slots[tls_index] = false;
-    g_current_process->misc_memory_used -= Memory::TLS_ENTRY_SIZE;
-    g_current_process->memory_region->used -= Memory::TLS_ENTRY_SIZE;
+    // Mark the TLS slot in the thread's page as free.
+    u32 tls_page = (tls_address - Memory::TLS_AREA_VADDR) / Memory::PAGE_SIZE;
+    u32 tls_slot = ((tls_address - Memory::TLS_AREA_VADDR) % Memory::PAGE_SIZE) / Memory::TLS_ENTRY_SIZE;
+    Kernel::g_current_process->tls_slots[tls_page].reset(tls_slot);
```
JayFoxRox: Should probably have use `.at()` here?

PR #1697, opened 2016-04-22
HWRasterizer: Lighting Sync + Cleanup

tfarley:
- Adds missing sync calls for specular light uniforms for new shaders
- Reorders function declarations in header to match definitions in source


bunnei:
LGTM

PR #1699, opened 2016-04-22
gsp::Gpu: implement AcquireRight, ReleaseRight functions

mailwl:
easy PR, tested on many games, don't break compatibility (even with games, which starts library applets)

### DIFF
--- a/src/core/hle/service/gsp_gpu.cpp
+++ b/src/core/hle/service/gsp_gpu.cpp
@@ -376,7 +376,14 @@ void SignalInterrupt(InterruptId interrupt_id) {
     if (!gpu_right_acquired) {
         return;
     }
-
+    if (nullptr == g_interrupt_event) {
+        LOG_WARNING(Service_GSP, "cannot synchronize until GSP event has been created!");
+        return;
+    }
+    if (nullptr == g_shared_memory) {
+        LOG_WARNING(Service_GSP, "cannot synchronize until GSP shared memory has been created!");
+        return;
+    }
     for (int thread_id = 0; thread_id < 0x4; ++thread_id) {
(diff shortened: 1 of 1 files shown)


src/core/hle/service/gsp_gpu.cpp
```diff
@@ -371,14 +373,10 @@ static void UnregisterInterruptRelayQueue(Service::Interface* self) {
  * @todo This probably does not belong in the GSP module, instead move to video_core
  */
 void SignalInterrupt(InterruptId interrupt_id) {
-    if (nullptr == g_interrupt_event) {
-        LOG_WARNING(Service_GSP, "cannot synchronize until GSP event has been created!");
-        return;
-    }
-    if (nullptr == g_shared_memory) {
-        LOG_WARNING(Service_GSP, "cannot synchronize until GSP shared memory has been created!");
```
wwylele: I think we still need to keep these two sanitizing check just in case.

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -371,6 +373,9 @@ static void UnregisterInterruptRelayQueue(Service::Interface* self) {
  * @todo This probably does not belong in the GSP module, instead move to video_core
  */
 void SignalInterrupt(InterruptId interrupt_id) {
+    if (!gpu_right_acquired) {
```
bunnei: Where does this come from? Could use a comment explaining what this means

mailwl: Citra itself.
It spamming "cannot synchronize until GSP event has been created!" before execute AcquareRight. Whan a game want to show applet, it use pairs ReleaseRight/SaveVramSysArea before, and AcquireRight/RestoreVramSysArea after applet works

PR #1704, opened 2016-04-23
Pica: PicaShaderConfig is TC and cleared before use

JayFoxRox:
This is my proposal to solve the issue in #1621 and #1690 where the padding bytes of the PicaShaderConfig struct would get random values and therefore fail the memcmp test.
It would result in new shaders being generated for every new random pad which quickly fills up the available memory with shaders and wastes too much time compiling them.

Making PicaShaderConfig a POD apparently only means we have to avoid the initializations of members which was useless anyway. It should be safe to assume that all data members will be set by PicaShaderConfig::CurrentConfig - otherwise the whole concept of copying the pica state would be fundamentally broken anyway.
(It surprised me that ~~`std::array`~~ (thanks MSVC..) and `BitField` are both considered POD tbh)

I think the POD / memset is the best option because it's reasonably lightweight and easy to maintain.

I'd also like to remind everyone that @lioncash had [a different proposal](https://github.com/citra-emu/citra/pull/1621#discussion_r58324896).
However, I thought it was too complicated as we'd have to add new variables in multiple places.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.h
+++ b/src/video_core/renderer_opengl/gl_rasterizer.h
@@ -39,17 +39,24 @@ struct ScreenInfo;
  * directly accessing Pica registers. This should reduce the risk of bugs in shader generation where
  * Pica state is not being captured in the shader cache key, thereby resulting in (what should be)
  * two separate shaders sharing the same key.
+ *
+ * We use a union because "implicitly-defined copy/move constructor for a union X copies the object representation of X."
+ * and "implicitly-defined copy assignment operator for a union X copies the object representation (3.9) of X."
+ * = Bytewise copy instead of memberwise copy.
+ * This is important because the padding bytes are included in the hash and comparison between objects.
  */
-struct PicaShaderConfig {
+union PicaShaderConfig {
 
     /// Construct a PicaShaderConfig with the current Pica register configuration.
     static PicaShaderConfig CurrentConfig() {
         PicaShaderConfig res;
-        std::memset(&res, 0, sizeof(PicaShaderConfig));
+
+        auto& state = res.state;
+        std::memset(&state, 0, sizeof(PicaShaderConfig::State));
 
         const auto& regs = Pica::g_state.regs;
 
-        res.alpha_test_func = regs.output_merger.alpha_test.enable ?
+        state.alpha_test_func = regs.output_merger.alpha_test.enable ?
             regs.output_merger.alpha_test.func.Value() : Pica::Regs::CompareFunc::Always;
 
         // Copy relevant tev stages fields.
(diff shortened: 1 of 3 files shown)


lioncash:
> However, I thought it was too complicated as we'd have to add new variables in multiple places.

Your change also comes with its own problems. The changing of std array to a raw C array allows for arbitrary decaying to a pointer, which is often not desirable. Even more, whenever a struct instance is created, everything is now in an inconsistent uninitialized state, so if anyone forgets to call memset on it, they're operating on uninitialized memory, rather than the struct initializing by itself, which is also inconsistent with the general initialization model of a lot of our other structs and classes use..

I believe you're conflating 'complicated' as "I'm not used to this in C, so it's different". Fully defining your comparison model is typically better as it keeps the class more flexible, as well as allowing the handling of non-POD types within the struct.

wwylele:
What about not using `BitField` in `PicaShaderConfig`? Remake the problematic bit field member as a normal plain struct, and do a simple transformation in `CurrentConfig`

JayFoxRox:
Did what wwylele suggested.
I'll rebase my other commits to this for testing, but I think this should be fine now.

I've also renamed the issue as we are not aiming for POD but TC (which allows us to do `ComputeHash64`, `memset`, `memcmp` etc.). However, I didn't rename my branch.

If MSVC considers `BitField& operator=(const BitField&) = delete;` TC at some point (which it should, just like gcc and clang do) we can remove the second commit / hack.
(Which is the C++14 way to handle TC instead of private default - according to the MSDN.. yeah.. lol)

JayFoxRox:
From IRC (slighlty modified for readability), back in WIP phase:

**JayFoxRox**: is `TC x, y; x = y;` forced to do a bytewise copy from y to x? or would we have to overload the operator for that? otherwise we might loose the padding bytes again after GetCurrentConfig

**neobrain**: it only does a member-wise copy (unless the operator is overloaded, yes)
I mean, technically the compiler is free to use a memcpy, but point being that it's not specified what happens to the padding bytes

**wwylele**: the biggest problem about copy is this line I think: https://github.com/citra-emu/citra/blob/e3a8292495cf0fb4297be27c696649dc44e011f0/src/video_core/renderer_opengl/gl_rasterizer.cpp#L862

**JayFoxRox**: so yeah.. overloading it is. pretty ugly, but gets the job done; or do you have any other idea?

wwylele:
I like this solution. I think we should get it merged soon.

hrydgard:
LGTM

JamePeng:
LGTM

wwylele:
LGTM

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -56,123 +66,151 @@ struct PicaShaderConfig {
...
+                bool enable;
+                bool abs_input;
+                Pica::Regs::LightingLutInput type;
+                float scale;
+            } lut_d0, lut_d1, lut_fr, lut_rr, lut_rg, lut_rb;
+        } lighting;
+
+    } state;
 };
+#if (__GNUC__ >= 5) || defined(__clang__) || defined(_MSC_VER)
```
tfarley: It would be neat if this little block could be put into a common macro or something to be used elsewhere when necessary.

JayFoxRox: Agreed, but not in this PR. (it's already being used throughout the codebase)

tfarley:
LGTM

PR #1705, opened 2016-04-23
Protect use of std::is_trivially_copyable to compile with GCC 4.9

LittleWhite-tb:
The protection was already used in dsp.h so I just added these for other uses of std::is_trivially_copyable. Thanks to this, I can compile citra with my "old" GCC 4.9.

### DIFF
--- a/src/common/file_util.h
+++ b/src/common/file_util.h
@@ -192,7 +192,9 @@ public:
     size_t ReadArray(T* data, size_t length)
     {
         static_assert(std::is_standard_layout<T>(), "Given array does not consist of standard layout objects");
+#if (__GNUC__ >= 5) || defined(__clang__) || defined(_MSC_VER)
         static_assert(std::is_trivially_copyable<T>(), "Given array does not consist of trivially copyable objects");
+#endif
 
         if (!IsOpen()) {
             m_good = false;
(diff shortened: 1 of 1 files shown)


bunnei:
LGTM

lioncash:
@bunei We bumped the buildbots up to 5.x. Keeping everything stuck on 4.x is just going to cause problems in the future. This also lets things do the completely wrong thing on 4.x. The assert is there for a reason. Conditionally disabling it is an awful idea.

LittleWhite-tb:
I agree that we should move to newer technologies and stuff (I mean like C++14). Even if I could not compile anymore citra (because of GCC 4.9), I did not wanted to PR this change. But, while I was going to locally fix the problem (and keep it only with me), I noticed that dsp.h already did have some preprocessor condition to ignore the is_trivially_copyable stuff.
So, it make sense to protect the use everywhere and let not so old compiler to compile fine the code.

You don't lose any protection against bad code with this change since the buildbots will automatically check the code and will report it on GitHub. Then, I would be surprised if a merge will be done on code reported as not compiling in buildbots env. 

Finally, let's take the five first Linux distribution from Distrowatch :
* Mint -> GCC 4.8.4
* Debian -> GCC 4.9.2 (Stable) GCC 5.3.1 (Unstable)
* Ubuntu -> 5.3.1
* OpenSUSE ->  4.8
* Fedora -> 4.9.2

Another solution would be to use __has_trivial_copyable, which is availabled in GCC 4.X, but this is not exactly the same that the one from the specification.

lioncash:
My concern is mostly rooted with how this now allows someone using 4.9.x to make a code change, or provide an implementation of something, only to have it potentially explode on them once they make the PR, which can be a waste of time for the contributor.

openSUSE Leap 42.1 provides GCC 5.3.1 as seen [here](https://software.opensuse.org/package/gcc5)
Fedora provides 5.3.1 and 6.0 builds as seen [here](https://apps.fedoraproject.org/packages/gcc)

MerryMage:
> I noticed that dsp.h already did have some preprocessor condition to ignore the is_trivially_copyable stuff.

That was done pre-gcc 5.x and probably should be removed.

PR #1708, opened 2016-04-24
DSP Service: Cleanup

MerryMage:
Had a little dive into dsp::DSP and this is what I came back with.

* Corrected handling of interrupts.
* Added return IPC headers.

### DIFF
--- a/src/audio_core/audio_core.cpp
+++ b/src/audio_core/audio_core.cpp
@@ -20,6 +20,8 @@ static void AudioTickCallback(u64 /*userdata*/, int cycles_late) {
     if (DSP::HLE::Tick()) {
         // TODO(merry): Signal all the other interrupts as appropriate.
         DSP_DSP::SignalPipeInterrupt(DSP::HLE::DspPipe::Audio);
+        // HACK(merry): Added to prevent regressions. Will remove soon.
+        DSP_DSP::SignalPipeInterrupt(DSP::HLE::DspPipe::Binary);
     }
 
     // Reschedule recurrent event
--- a/src/audio_core/audio_core.cpp
+++ b/src/audio_core/audio_core.cpp
@@ -4,6 +4,7 @@
 
 #include "audio_core/audio_core.h"
 #include "audio_core/hle/dsp.h"
+#include "audio_core/hle/pipe.h"
 
 #include "core/core_timing.h"
 #include "core/hle/kernel/vm_manager.h"
@@ -17,10 +18,8 @@ static constexpr u64 audio_frame_ticks = 1310252ull; ///< Units: ARM11 cycles
 
 static void AudioTickCallback(u64 /*userdata*/, int cycles_late) {
     if (DSP::HLE::Tick()) {
-        // HACK: We're not signaling the interrups when they should be, but just firing them all off together.
-        // It should be only (interrupt_id = 2, channel_id = 2) that's signalled here.
-        // TODO(merry): Understand when the other interrupts are fired.
-        DSP_DSP::SignalAllInterrupts();
+        // TODO(merry): Signal all the other interrupts as appropriate.
+        DSP_DSP::SignalPipeInterrupt(DSP::HLE::DspPipe::Audio);
     }
 
     // Reschedule recurrent event
--- a/src/audio_core/hle/pipe.cpp
+++ b/src/audio_core/hle/pipe.cpp
@@ -12,6 +12,8 @@
 #include "common/common_types.h"
 #include "common/logging/log.h"
 
+#include "core/hle/service/dsp_dsp.h"
+
 namespace DSP {
 namespace HLE {
 
@@ -97,6 +99,8 @@ static void AudioPipeWriteStructAddresses() {
     for (u16 addr : struct_addresses) {
         WriteU16(DspPipe::Audio, addr);
     }
+    // Signal that we have data on this pipe.
+    DSP_DSP::SignalPipeInterrupt(DspPipe::Audio);
 }
 
 void PipeWrite(DspPipe pipe_number, const std::vector<u8>& buffer) {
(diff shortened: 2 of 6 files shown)


wwylele:
Many of my games can't boot and spams with unmapped read/write...

MerryMage:
@wwylele I forgot to invert a conditional. Fixed.

src/audio_core/audio_core.cpp
```diff
@@ -17,10 +18,10 @@ static constexpr u64 audio_frame_ticks = 1310252ull; ///< Units: ARM11 cycles
...
 static void AudioTickCallback(u64 /*userdata*/, int cycles_late) {
     if (DSP::HLE::Tick()) {
-        // HACK: We're not signaling the interrups when they should be, but just firing them all off together.
-        // It should be only (interrupt_id = 2, channel_id = 2) that's signalled here.
-        // TODO(merry): Understand when the other interrupts are fired.
-        DSP_DSP::SignalAllInterrupts();
+        // TODO(merry): Signal all the other interrupts as appropriate.
+        DSP_DSP::SignalPipeInterrupt(DSP::HLE::DspPipe::Audio);
+        // HACK(merry): Added to prevent regressions. Will remove soon.
+        DSP_DSP::SignalPipeInterrupt(DSP::HLE::DspPipe::Binary);
```
Subv: When should this interrupt be signaled?

Subv:
LGTM

src/audio_core/audio_core.cpp
MerryMage: Pipe interrupts are signals after pipe-specific events (e.g.: when data is available on a pipe.)

src/audio_core/hle/pipe.cpp
```diff
@@ -49,16 +53,20 @@ std::vector<u8> PipeRead(DspPipe pipe_number, u32 length) {
...
 
-    return pipe_data[static_cast<size_t>(pipe_number)].size();
+    return pipe_data[pipe_index].size();
 }
 
 static void WriteU16(DspPipe pipe_number, u16 value) {
-    std::vector<u8>& data = pipe_data[static_cast<size_t>(pipe_number)];
+    const size_t pipe_index = static_cast<size_t>(pipe_number);
+
+    std::vector<u8>& data = pipe_data[pipe_index];
```
JayFoxRox: Shouldn't there be an `ASSERT(pipe_index < NUM_DSP_PIPE)` or is this part of the code considered to be "overflow safe"?

MerryMage: This function is internal to the file, but I could always add in a check for sanity's sake.

lioncash: Could always just do `std::vector<u8>& data = pipe_data.at(pipe_index);`

MerryMage: Good idea.

MerryMage:
Squashed fixups.

MerryMage:
Looking at the compiler, GCC gives a few type conversion warnings with this code and I'd like to fix those before a merge.

bunnei:
Changes LGTM. I'll wait for you to fix the warnings...

MerryMage:
Done.

Fixed GCC's warnings (which all pertained to format string arguments). (MSVC warnings are done by #1708.)

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -12,37 +13,77 @@
...
+            const size_t pipe_index = static_cast<size_t>(dsp_pipe);
+            ASSERT(pipe_index < DSP::HLE::NUM_DSP_PIPE);
+            return pipe[pipe_index];
+        }
+        default:
+            UNREACHABLE_MSG("Invalid interrupt type = %zu", static_cast<size_t>(type));
+        }
+    }
+
+    bool HasTooManyEventsRegistered() {
```
lioncash: You can possibly do something like this, which doesn't put the count_if on one long line. Not really a big deal though.

```cpp
bool HasTooManyEventsRegistered() {
    // Actual service implementation only has 6 'slots' for interrupts.
    constexpr size_t max_number_of_interrupt_events = 6;

    size_t number = std::count_if(pipe.begin(), pipe.end(), [](const auto& evt) {
        return evt != nullptr;
    });

    if (zero != nullptr)
        number++;
    if (one != nullptr)
        number++;

    return number >= max_number_of_interrupt_events;
}
```

This function can also be const qualified.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -12,37 +13,77 @@
...
-    std::size_t operator()(const std::pair<T, U> &x) const {
-        // TODO(yuriks): Replace with better hash combining function.
-        return std::hash<T>()(x.first) ^ std::hash<U>()(x.second);
+/// There are three types of interrupts
+enum class InterruptType {
+    Zero, One, Pipe
+};
+constexpr size_t NUM_INTERRUPT_TYPE = 3;
+
+struct InterruptEvents {
```
lioncash: It may be preferable to make this a class, considering it has private members.

MerryMage:
Addressed @lioncash's comments above.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -12,37 +13,80 @@
...
+            const size_t pipe_index = static_cast<size_t>(dsp_pipe);
+            ASSERT(pipe_index < DSP::HLE::NUM_DSP_PIPE);
+            return pipe[pipe_index];
+        }
+        }
+
+        UNREACHABLE_MSG("Invalid interrupt type = %zu", static_cast<size_t>(type));
+    }
+
+    bool HasTooManyEventsRegistered() {
```
lioncash: This can be const qualified (was in my original comment, but I appended it a few seconds after making the comment so it might not have updated when you saw it).

MerryMage: Ah, apologies, missed that.

bunnei:
Can you squash the fixups?

MerryMage:
Squashed.

PR #1709, opened 2016-04-24
Make Citra build with MICROPROFILE_ENABLED set to 0

hrydgard:
Useful when using other profilers to get cleaner results.

Set it in src/common/microprofile.h. I also added a comment there to clarify.


mailwl:
by the way, i really don't understand, how to use microprofile.
maybe it good thing, but... well anyone need it, i think

src/citra_qt/debugger/profiler.cpp
```diff
@@ -6,16 +6,18 @@
 #include <QPainter>
 #include <QString>
 
+#include "common/microprofile.h"
 #include "citra_qt/debugger/profiler.h"
 #include "citra_qt/util/util.h"
 
-#include "common/microprofile.h"
```
MerryMage: Shouldn't be moved.

src/citra_qt/debugger/profiler.h
```diff
@@ -7,8 +7,9 @@
 #include <QAbstractItemModel>
 #include <QDockWidget>
 #include <QTimer>
-#include "ui_profiler.h"
 
+#include "common/microprofile.h"
+#include "ui_profiler.h"
```
MerryMage: Should likely be `#include "citra-qt/ut_profiler.h"`, and be above the `common` includes.

src/citra_qt/main.cpp
```diff
@@ -12,6 +12,8 @@
 #include <QMessageBox>
 #include "qhexedit.h"
 
+#include "common/microprofile.h"
```
MerryMage: Shouldn't be moved.

src/citra_qt/debugger/profiler.cpp
```diff
@@ -342,5 +349,4 @@ void MicroProfileDrawLine2D(u32 vertices_length, float* vertices, u32 hex_color)
     mp_painter->drawPolyline(point_buf.data(), vertices_length);
     point_buf.clear();
 }
-
-#endif MICROPROFILE_ENABLED
+#endif
```
MerryMage: Newline at end of file.

src/citra_qt/debugger/profiler.h
```diff
@@ -69,5 +67,3 @@ class MicroProfileDialog : public QWidget {
 private:
     QAction* toggle_view_action = nullptr;
 };
-
-#endif
```
MerryMage: Newline at end of file.

src/citra_qt/debugger/profiler.h
hrydgard: No, that doesn't compile. Is ui_compiler.h moc-generated? There certainly isn't one there..

src/citra_qt/main.cpp
hrydgard: Doesn't work if I don't move it. Can't figure out why, these generated Qt headers and stuff confuse me...

hrydgard:
Now, should I squash?

MerryMage:
LGTM

src/citra_qt/debugger/profiler.cpp
```diff
@@ -337,3 +349,4 @@ void MicroProfileDrawLine2D(u32 vertices_length, float* vertices, u32 hex_color)
     mp_painter->drawPolyline(point_buf.data(), vertices_length);
     point_buf.clear();
 }
+#endif
```
bunnei: Missing line ending

bunnei:
@hrydgard Yes you can squash after my one comment is addressed, otherwise looks fine to me.

FWIW you don't actually need to squash for small changes like this (I can squash on merge to a single commit via GitHub), I just prefer not to do that for bigger changes with multiple, well-separated commits, as the history is lost.

hrydgard:
@bunnei  OK, won't bother squashing then.

However, I tried adding a newline at the end of src/citra_qt/debugger/profiler.cpp, only to hit the following pre-commit check:

    src/citra_qt/debugger/profiler.cpp:353: new blank line at EOF.

    Error: This commit would contain trailing spaces or tabs, which is against this repo's poli   cy.
    Please correct those issues before commiting. (Use 'git diff --check' for more details)
    If you know what you are doing, you can try 'git commit --no-verify' to bypass the check

Looking at other files, they don't have an extra line ending at the end either. Please advise.

hrydgard:
Newline fixed, please squash-merge :)

src/common/microprofile.h
```diff
@@ -4,6 +4,10 @@
 
 #pragma once
 
+// Uncomment this to disable microprofile. This will get you cleaner profiles when using
+// external sampling profilers like "Very Sleepy", and will improve performance somewhat.
+// #define MICROPROFILE_ENABLED 0
```
linkmauve: It would be better to have a cmake option than to need to modify this file, imo.

JayFoxRox: Agreed. But there are too many PRs lately so we should get this merged soon.
We can still create a "cmake options" PR as Pica also has some debug defines which have to be changed in the code - I'd also rather have cmake options for that too. If we bundle all options in another PR we are more likely to get a consistent way of handling this.

hrydgard: I agree with JayFoxRox. Another option that is relevant is the other profiler, "ENABLE_PROFILER". That one eats ~5% CPU currently (mostly because it's called as often as once per vertex).

bunnei: Agreed with the above

PR #1710, opened 2016-04-24
Replace std::map with std::array for graphics event breakpoints

hrydgard:
And allow the compiler to inline by moving the actual check to the header.

Saves 1%+ in vertex heavy situations as std::map lookups are quite expensive. As Citra was doing one per Pica reg write, that can be replaced with an array lookup, I thought that might be worth fixing.


bunnei:
LGTM

PR #1714, opened 2016-04-24
Fixes for #1711 and other compilation errors.

### LINKED ISSUE #1711: Linux compilation errors
Building on Linux has started to fail with the following error:

/usr/bin/ld: ../common/libcommon.a(microprofile.cpp.o): undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/usr/lib64/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [src/citra_qt/citra-qt] Error 1
make[1]: *** [src/citra_qt/CMakeFiles/citra-qt.dir/all] Error 2
make: *** [all] Error 2

smspillaz:
The big change is moving to CMake 3.1. Other changes
fix compiler errors noticed on Clang 3.9 / OSX.

Fixes #1711

### DIFF
--- a/src/video_core/shader/shader_jit_x64.cpp
+++ b/src/video_core/shader/shader_jit_x64.cpp
@@ -795,6 +795,8 @@ void JitShader::FindReturnOffsets() {
         case OpCode::Id::CALLU:
             return_offsets.push_back(instr.flow_control.dest_offset + instr.flow_control.num_instructions);
             break;
+        default:
+            break;
         }
     }
 
--- a/src/video_core/shader/shader_jit_x64.cpp
+++ b/src/video_core/shader/shader_jit_x64.cpp
@@ -856,7 +856,7 @@ void JitShader::Compile() {
     uintptr_t size = reinterpret_cast<uintptr_t>(GetCodePtr()) - reinterpret_cast<uintptr_t>(program);
     ASSERT_MSG(size <= MAX_SHADER_SIZE, "Compiled a shader that exceeds the allocated size!");
 
-    LOG_DEBUG(HW_GPU, "Compiled shader size=%d", size);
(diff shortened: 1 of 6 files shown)


src/video_core/shader/shader_jit_x64.cpp
```diff
@@ -795,6 +795,8 @@ void JitShader::FindReturnOffsets() {
         case OpCode::Id::CALLU:
             return_offsets.push_back(instr.flow_control.dest_offset + instr.flow_control.num_instructions);
             break;
+        default:
+            UNREACHABLE_MSG("HitShader::FindReturnOffsets: Cannot handle non-CALL opcode here");
```
MerryMage: This is reachable; this should likely be `default: break;`.

smspillaz: Cheers, thanks for the tip. I'll make it `break`.

JamePeng:
travis-ci/pr :
CMake Error at CMakeLists.txt:3 (cmake_minimum_required):
CMake 3.1 or higher is required.  You are running version 2.8.11

smspillaz:
@JamePeng This was intentional - I had to bump the `cmake_minimum_required` version to support the newer `FindThreads` module.

Unfortunately travis-ci runs an Ubuntu LTS release that is now four(!!) years out of date. The solution is to either migrate to trusty or add a PPA that gets a newer CMake version.

As for OS X - I'm not sure exactly how to deal with that right now. I think there may be newer versions of CMake in homebrew/versions.

bunnei:
Code LGTM.
Merging this to fix master... If you had any other changes, please just add them as a followup PR.

PR #1718, opened 2016-04-25
fixup simple type conversions where possible

alex-laties:
This PR is a step towards fixing all warnings under the MSVC.

There are 3 major classes of warnings that need fixing:
 - implicit type casts where data can be lost
 - code paths where not all branches return
 - comparisons between signed/unsigned values

I only look to fix implicit type casts here, and only for the `citra` project directly. I make no changes to `microprofile` or `nihstro` here.

I'm looking to fix the other classes of warnings in other PRs.

### DIFF
--- a/src/audio_core/hle/pipe.cpp
+++ b/src/audio_core/hle/pipe.cpp
@@ -36,12 +36,17 @@ std::vector<u8> PipeRead(DspPipe pipe_number, u32 length) {
         return {};
     }
 
+    if (length > UINT16_MAX) { // Can only read at most UINT16_MAX from the pipe
+        LOG_ERROR(Audio_DSP, "length of %u greater than max of %u", length, UINT16_MAX);
+        return {};
+    }
+
     std::vector<u8>& data = pipe_data[pipe_index];
 
     if (length > data.size()) {
         LOG_WARNING(Audio_DSP, "pipe_number = %zu is out of data, application requested read of %u but %zu remain",
                     pipe_index, length, data.size());
-        length = data.size();
+        length = static_cast<u32>(data.size());
     }
 
     if (length == 0)
@@ -94,7 +99,7 @@ static void AudioPipeWriteStructAddresses() {
     };
 
     // Begin with a u16 denoting the number of structs.
-    WriteU16(DspPipe::Audio, struct_addresses.size());
+    WriteU16(DspPipe::Audio, static_cast<u16>(struct_addresses.size()));
     // Then write the struct addresses.
     for (u16 addr : struct_addresses) {
         WriteU16(DspPipe::Audio, addr);
--- a/src/audio_core/hle/pipe.h
+++ b/src/audio_core/hle/pipe.h
@@ -24,10 +24,14 @@ enum class DspPipe {
 constexpr size_t NUM_DSP_PIPE = 8;
 
 /**
- * Read a DSP pipe.
- * @param pipe_number The Pipe ID
- * @param length How much data to request.
- * @return The data read from the pipe. The size of this vector can be less than the length requested.
+ * Reads `length` bytes from the DSP pipe identified with `pipe_number`.
+ * @note Can read up to the maximum value of a u16 in bytes (65,535).
+ * @note IF an error is encoutered with either an invalid `pipe_number` or `length` value, an empty vector will be returned.
+ * @note IF `length` is set to 0, an empty vector will be returned.
+ * @note IF `length` is greater than the amount of data available, this function will only read the available amount.
+ * @param pipe_number a `DspPipe`
+ * @param length the number of bytes to read. The max is 65,535 (max of u16).
+ * @returns a vector of bytes from the specified pipe. On error, will be empty.
  */
 std::vector<u8> PipeRead(DspPipe pipe_number, u32 length);
 
--- a/src/citra_qt/util/util.cpp
+++ b/src/citra_qt/util/util.cpp
@@ -19,7 +19,7 @@ QString ReadableByteSize(qulonglong size) {
     static const std::array<const char*, 6> units = { "B", "KiB", "MiB", "GiB", "TiB", "PiB" };
     if (size == 0)
         return "0";
-    int digit_groups = std::min<int>((int)(std::log10(size) / std::log10(1024)), units.size());
+    int digit_groups = std::min<int>(static_cast<int>(std::log10(size) / std::log10(1024)), static_cast<int>(units.size()));
     return QString("%L1 %2").arg(size / std::pow(1024, digit_groups), 0, 'f', 1)
                             .arg(units[digit_groups]);
 }
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -374,7 +374,7 @@ static void SendReply(const char* reply) {
 
     memset(command_buffer, 0, sizeof(command_buffer));
 
-    command_length = strlen(reply);
+    command_length = static_cast<u32>(strlen(reply));
     if (command_length + 4 > sizeof(command_buffer)) {
         LOG_ERROR(Debug_GDBStub, "command_buffer overflow in SendReply");
         return;
@@ -515,7 +515,7 @@ static bool IsDataAvailable() {
         return false;
     }
 
-    return FD_ISSET(gdbserver_socket, &fd_socket);
+    return FD_ISSET(gdbserver_socket, &fd_socket) != 0;
 }
 
 /// Send requested register to gdb client.
@@ -633,10 +633,10 @@ static void ReadMemory() {
 
     auto start_offset = command_buffer+1;
     auto addr_pos = std::find(start_offset, command_buffer+command_length, ',');
-    PAddr addr = HexToInt(start_offset, addr_pos - start_offset);
+    PAddr addr = HexToInt(start_offset, static_cast<u32>(addr_pos - start_offset));
 
     start_offset = addr_pos+1;
-    u32 len = HexToInt(start_offset, (command_buffer + command_length) - start_offset);
+    u32 len = HexToInt(start_offset, static_cast<u32>((command_buffer + command_length) - start_offset));
 
     LOG_DEBUG(Debug_GDBStub, "gdb: addr: %08x len: %08x\n", addr, len);
 
@@ -658,11 +658,11 @@ static void ReadMemory() {
 static void WriteMemory() {
     auto start_offset = command_buffer+1;
     auto addr_pos = std::find(start_offset, command_buffer+command_length, ',');
-    PAddr addr = HexToInt(start_offset, addr_pos - start_offset);
+    PAddr addr = HexToInt(start_offset, static_cast<u32>(addr_pos - start_offset));
 
     start_offset = addr_pos+1;
     auto len_pos = std::find(start_offset, command_buffer+command_length, ':');
-    u32 len = HexToInt(start_offset, len_pos - start_offset);
+    u32 len = HexToInt(start_offset, static_cast<u32>(len_pos - start_offset));
 
     u8* dst = Memory::GetPointer(addr);
     if (!dst) {
@@ -752,10 +752,10 @@ static void AddBreakpoint() {
 
     auto start_offset = command_buffer+3;
     auto addr_pos = std::find(start_offset, command_buffer+command_length, ',');
-    PAddr addr = HexToInt(start_offset, addr_pos - start_offset);
+    PAddr addr = HexToInt(start_offset, static_cast<u32>(addr_pos - start_offset));
 
     start_offset = addr_pos+1;
-    u32 len = HexToInt(start_offset, (command_buffer + command_length) - start_offset);
+    u32 len = HexToInt(start_offset, static_cast<u32>((command_buffer + command_length) - start_offset));
 
     if (type == BreakpointType::Access) {
         // Access is made up of Read and Write types, so add both breakpoints
@@ -800,10 +800,10 @@ static void RemoveBreakpoint() {
 
     auto start_offset = command_buffer+3;
     auto addr_pos = std::find(start_offset, command_buffer+command_length, ',');
-    PAddr addr = HexToInt(start_offset, addr_pos - start_offset);
+    PAddr addr = HexToInt(start_offset, static_cast<u32>(addr_pos - start_offset));
 
     start_offset = addr_pos+1;
-    u32 len = HexToInt(start_offset, (command_buffer + command_length) - start_offset);
+    u32 len = HexToInt(start_offset, static_cast<u32>((command_buffer + command_length) - start_offset));
 
     if (type == BreakpointType::Access) {
         // Access is made up of Read and Write types, so add both breakpoints
--- a/src/core/hle/service/dsp_dsp.cpp
+++ b/src/core/hle/service/dsp_dsp.cpp
@@ -288,7 +288,7 @@ static void WriteProcessPipe(Service::Interface* self) {
     ASSERT_MSG(Memory::GetPointer(buffer) != nullptr, "Invalid Buffer: pipe=%u, size=0x%X, buffer=0x%08X", pipe_index, size, buffer);
 
     std::vector<u8> message(size);
-    for (size_t i = 0; i < size; i++) {
+    for (u32 i = 0; i < size; i++) {
         message[i] = Memory::Read8(buffer + i);
     }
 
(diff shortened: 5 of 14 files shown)


src/audio_core/hle/pipe.cpp
```diff
@@ -37,7 +37,7 @@ std::vector<u8> PipeRead(DspPipe pipe_number, u32 length) {
     if (length > data.size()) {
         LOG_WARNING(Audio_DSP, "pipe_number = %u is out of data, application requested read of %u but %zu remain",
                     pipe_number, length, data.size());
-        length = data.size();
+        length = static_cast<u32>(data.size());
```
linkmauve: Shouldn’t this be u16 too? (and same for the length parameter)
It seems to be only used from src/core/hle/service/dsp_dsp.cpp where size is always the lowest 16 bits of the parameter.

src/citra_qt/util/util.cpp
```diff
@@ -19,7 +19,7 @@ QString ReadableByteSize(qulonglong size) {
     static const std::array<const char*, 6> units = { "B", "KiB", "MiB", "GiB", "TiB", "PiB" };
     if (size == 0)
         return "0";
-    int digit_groups = std::min<int>((int)(std::log10(size) / std::log10(1024)), units.size());
+    int digit_groups = std::min<int>((int)(std::log10(size) / std::log10(1024)), static_cast<int>(units.size()));
```
linkmauve: Also replace the C-style (int) cast with a static_cast.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -115,7 +115,7 @@ R"(l<?xml version="1.0"?>
 
 namespace GDBStub {
 
-static int gdbserver_socket = -1;
+static SOCKET gdbserver_socket = -1;
```
MerryMage: This likely won't work for non-Windows platforms.

linkmauve: A socket is defined as an int, I don’t know where you found that SOCKET type but it’s likely wrong.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -936,7 +936,7 @@ void Init(u16 port) {
     WSAStartup(MAKEWORD(2, 2), &InitData);
 #endif
 
-    int tmpsock = socket(PF_INET, SOCK_STREAM, 0);
+    SOCKET tmpsock = socket(PF_INET, SOCK_STREAM, 0);
```
linkmauve: Same here.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -216,7 +216,7 @@ static void WriteProcessPipe(Service::Interface* self) {
     std::vector<u8> message(size);
 
     for (size_t i = 0; i < size; i++) {
-        message[i] = Memory::Read8(buffer + i);
+        message[i] = Memory::Read8(buffer + static_cast<u32>(i));
```
linkmauve: Better change the type in the loop.

src/core/loader/3dsx.cpp
```diff
@@ -182,7 +182,7 @@ static THREEDSX_Error Load3DSXFile(FileUtil::IOFile& file, u32 base_addr, Shared
                     pos += table.skip;
                     s32 num_patches = table.patch;
                     while (0 < num_patches && pos < end_pos) {
-                        u32 in_addr = (u8*)pos - program_image.data();
+                        u32 in_addr = static_cast<u32>((u8*)pos - program_image.data());
```
MerryMage: might as well `reinterpret_cast` while you're here

src/core/hle/service/dsp_dsp.cpp
alex-laties: seems reasonable, changing.

src/core/gdbstub/gdbstub.cpp
alex-laties: @MerryMage you are correct! I didn't think too much about it.

@linkmauve that's only true on POSIX platforms. On Windows, a socket is represented with a `SOCKET`, a typedef of `UINT_PTR`.

I will have to do some skullduggery to make this work sanely.

src/audio_core/hle/pipe.cpp
alex-laties: I don't know enough to be sure at the moment.

Is there way to test whether this will act correctly if I change `length` to a `u16`?

src/citra_qt/util/util.cpp
alex-laties: will do

src/audio_core/hle/pipe.cpp
MerryMage: I wrote that; length will be fine as a `u32`.

You should likely change the two instances of `u32 size = cmd_buff[3] & 0xFFFF;` in src/core/hle/service/dsp_dsp.cpp to `u16 size = static_cast<u16>(cmd_buff[3] & 0xFFFF);`

Alternatively if you wish you can leave that edit to me and I'll add it to #1708.

alex-laties: If that's all that's necessary, I'll do it here.

src/core/gdbstub/gdbstub.cpp
alex-laties: After looking at this a little more, I think the better decision is to back the changes related to `SOCKET` out for now.

There will have to be some work to make this setup windows-compatible, since there are some differences in how you handle error checking with windows sockets that are apparently currently broken.

I.E. doing `if (socket == -1)` to check if initialization failed only works correctly if `socket` is an int. if `socket` is `UINT_PTR`, `-1` will get coerced to a `max(unsigned long long)` on windows.

src/audio_core/hle/pipe.cpp
lioncash: The parameter itself should likely remain a u32. Otherwise it'll implicitly convert to `int` when used in an arithmetic expression (which can cause sign conversion warnings). It also makes it more annoying as an API, because u32 and size_t are generally more common for representing lengths than a u16 (and you also don't get casting at the callsites most of the time because of it).

src/core/gdbstub/gdbstub.cpp
alex-laties: Actually... I retract my statement that this is currently broken on windows. It's not.

The error return val for `socket` on windows is `INVALID_SOCKET`, which evaluates to the same thing as `(unsigned long long)-1`. Magical.

src/audio_core/hle/pipe.cpp
alex-laties: It sounds to me that there are two options if we want this function to be correct:
 - Set the type correctly and manage the arithmetic within the scope of the function. Up-cast as necessary.
 - Leave `length` as a `u32` is to then insert a guard in the function that checks if the value is over `max(u16)`.

I personally lean towards the first option. From the API perspective, it communicates more about the correct behavior than option 2 does.

lioncash: Your API shouldn't introduce unnecessary casts where possible. size_t was introduced strictly because it's the most adequate type for expressing a non-negative range (and generally best suited regarding pointer related things alongside other types, but that isn't relevant).

If going over 16 bits is an issue, then that should be an asserted condition or documented as clamping the value. Uses of an API shouldn't be concerned with the overall implementation details of how a function works.

alex-laties: K. Option 2 it is!

alex-laties: @lioncash @MerryMage ready for review again!

src/audio_core/hle/pipe.cpp
```diff
@@ -26,12 +26,30 @@ void ResetPipes() {
...
+* Can read up to the maximum value of a u16 in bytes (65,535).
+* IF an error is encoutered with either an invalid `pipe_number`
+* or `length` value, an empty vector will be returned.
+* IF `length` is set to 0, an empty vector will be returned.
+* IF `length` is greater than the amount of data available,
+* this function will only read the available amount.
+* \param pipe_number a `DspPipe`
+* \param length the number of bytes to read. The max is 65,535 (max of u16).
+* \returns a vector of bytes from the specified pipe. On error, will be empty.
+*/
```
JayFoxRox: We usually doc in the headers: https://github.com/citra-emu/citra/blob/master/src/audio_core/hle/pipe.h#L26

alex-laties: Ah, thank you! I'll move this.

src/audio_core/hle/pipe.cpp
```diff
@@ -27,17 +27,23 @@ void ResetPipes() {
 }
 
 std::vector<u8> PipeRead(DspPipe pipe_number, u32 length) {
+
```
lioncash: accidental newline?

src/audio_core/hle/pipe.h
```diff
@@ -24,11 +24,17 @@ enum class DspPipe {
...
 
 /**
- * Read a DSP pipe.
- * @param pipe_number The Pipe ID
- * @param length How much data to request.
- * @return The data read from the pipe. The size of this vector can be less than the length requested.
- */
+* Reads `length` bytes from the DSP pipe identified with `pipe_number`.
+* Can read up to the maximum value of a u16 in bytes (65,535).
+* IF an error is encoutered with either an invalid `pipe_number`
```
lioncash: The individual boundary conditions can probably be prefixed with an `@remark` or `@note` tag

src/audio_core/hle/pipe.cpp
alex-laties: ah, bad habit of mine. I usually leave one extra line for function sigs.
Removing.

JayFoxRox: Probably intentional to relax the look of the function a little. (I'd personally keep the space, though I know it's frowned upon to have empty lines in many conventions - do we have rules for this?)

src/audio_core/hle/pipe.h
alex-laties: Makes sense. Adding.

lioncash:
Other than those two minor nitpicks, it looks fine to me.

src/audio_core/hle/pipe.cpp
alex-laties: I'm gonna go ahead and remove it for now @JayFoxRox.

It's purely cosmetic, preferential, and it's not like removing the superfluous whitespace here makes this function significantly less readable.

I didn't see anything in the `CONTRIBUTING.md` around whitespace, other than 4-space indents.

I completely understand the push back on superfluous whitespace. Having been in codebases where people are quite liberal with it... code becomes extremely unreadable.

MerryMage:
Looks fine to me. Squash please.

src/audio_core/hle/pipe.h
```diff
@@ -24,11 +24,15 @@ enum class DspPipe {
 };
 
 /**
- * Read a DSP pipe.
- * @param pipe_number The Pipe ID
- * @param length How much data to request.
- * @return The data read from the pipe. The size of this vector can be less than the length requested.
- */
+* Reads `length` bytes from the DSP pipe identified with `pipe_number`.
```
MerryMage: Space before the `*`s, please.

alex-laties:
squashed.

There were conflicts when doing the rebase onto citra/master.

@MerryMage mind doing a quick once-over to make sure none of the resolutions I chose messed too much up? Most of it was in `PipeRead`.

src/audio_core/hle/pipe.h
```diff
@@ -24,10 +24,14 @@ enum class DspPipe {
...
- * @param length How much data to request.
- * @return The data read from the pipe. The size of this vector can be less than the length requested.
+ * Reads `length` bytes from the DSP pipe identified with `pipe_number`.
+ * @note Can read up to the maximum value of a u16 in bytes (65,535).
+ * @note IF an error is encoutered with either an invalid `pipe_number` or `length` value, an empty vector will be returned.
+ * @note IF `length` is set to 0, an empty vector will be returned.
+ * @note IF `length` is greater than the amount of data available, this function will only read the available amount.
+ * @param pipe_number a `DspPipe`
+ * @param length the number of bytes to read. The max is 65,535 (max of u16).
+ * @returns a vector of bytes from the specified pipe. On error, will be empty.
```
JayFoxRox: (nit, unless something else is bad don't change this)
`$ grep "\@return " -R ./ | wc -l -> 138` (Also used elsewhere in this file)
`$ grep "\@returns " -R ./ | wc -l -> 30`

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -476,7 +476,7 @@ bool RendererOpenGL::Init() {
 
     if (GLAD_GL_KHR_debug) {
         glEnable(GL_DEBUG_OUTPUT);
-        glDebugMessageCallback(DebugHandler, nullptr);
+        glDebugMessageCallback(reinterpret_cast<GLDEBUGPROC>(DebugHandler), nullptr);
```
JayFoxRox: This looks dangerous to me.
If there is any other way to fix warnings here (are there even any? why would there be one if the function is declared fine?) other than discarding the type info of a callback handler?

alex-laties: MSVC is a bit of a weird beast. It does complain that `DebugHandler` does not match `GLDEBUGPROC`, even though the signature does appear to match that of `GLDEBUGPROC`.

Not sure why it complains in the first place.

src/core/hle/service/dsp_dsp.cpp
```diff
@@ -287,8 +287,7 @@ static void WriteProcessPipe(Service::Interface* self) {
 
     ASSERT_MSG(Memory::GetPointer(buffer) != nullptr, "Invalid Buffer: pipe=%u, size=0x%X, buffer=0x%08X", pipe_index, size, buffer);
 
-    std::vector<u8> message(size);
```
JayFoxRox: Accidently removed this? (Travis and Appveyor fail here)

alex-laties: yes. fixing.

src/video_core/renderer_opengl/renderer_opengl.cpp
JayFoxRox: I can't see any warnings like that in the other AppVeyor logs. Could you remove the cast here or provide more details?
(Maybe MSVC++ uses your own glad includes - possibly outdated? If so that could mean we have a bug elsewhere because that should not happen afaik)

alex-laties: I'm currently traveling and do not have access to the same build
environment at the moment.

I think it's safer to back this change out and experiment when I get back.
Sound good?

JayFoxRox: Yes.

src/video_core/command_processor.cpp
```diff
@@ -132,7 +132,7 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     // sent to the primitive assembler.
 
                     auto& immediate_input = g_state.immediate.input_vertex;
-                    auto& immediate_attribute_id = g_state.immediate.current_attribute;
+                    auto immediate_attribute_id = static_cast<unsigned int>(g_state.immediate.current_attribute);
```
wwylele: Why remove reference? The behaviour will be different than before, since the next line is `immediate_input.attr[immediate_attribute_id++] = attribute;`

alex-laties: Because I'm an idiot, basically. Fixing.

alex-laties: So, the entire reason I did this was because of the `if (immediate_attribute_id >= regs.vs.num_input_attributes+1)` being a signed vs unsigned comparison, `immediate_attribute_id` being an `int` and `regs.vs.num_input_attributes+1` being a `u32`.

I think the correct fix would be to change the type of [`current_attribute`](https://github.com/citra-emu/citra/blob/master/src/video_core/pica_state.h#L54) from an `int` to a `u32`.  It is never set to a negative value anywhere in the code base as far as I can tell.

src/video_core/pica_state.h
```diff
@@ -51,7 +51,7 @@ struct State {
         // Used to buffer partial vertices for immediate-mode rendering.
         Shader::InputVertex input_vertex;
         // Index of the next attribute to be loaded into `input_vertex`.
-        int current_attribute = 0;
+        u32 current_attribute = 0;
```
alex-laties: @yuriks This fixes a signed vs unsigned comparison [here](https://github.com/citra-emu/citra/blob/master/src/video_core/command_processor.cpp#L139).

From what I can glean, this type change is harmless, as `current_attribute` is never subtracted and never assigned a negative value. However, I'd appreciate another eye on this.

Thanks.

src/audio_core/hle/pipe.cpp
```diff
@@ -36,12 +36,17 @@ std::vector<u8> PipeRead(DspPipe pipe_number, u32 length) {
         return {};
     }
 
+    if (length > UINT16_MAX) { //can only read at most UINT16_MAX from the pipe
```
Subv: // Can only read...

Subv:
LGTM other than my one comment.

@alex-laties Please address the comments and squash so we can get this merged soon, thanks!

alex-laties:
@Subv addressed and squashed. G2G!

PR #1719, opened 2016-04-26
Interpreter instruction cache

ghost:
Profiling showed that the interpreter spent around 10% time doing std::unordered_map lookups for the instruction cache . This PR reduces it to ~2% by caching into vectors instead, in a similar way to Memory::Read (look into page table, check for nullptr then get the data)
It provides a global that allows to clear the cache(s) and un/register codesets that can be useful for CRO
It also fixes MerryMage's jit that suffers the same issue, i have made a separate branch that would support this change

### DIFF
--- a/src/core/arm/cache/cache.cpp
+++ b/src/core/arm/cache/cache.cpp
@@ -87,11 +87,9 @@ void CacheBase::OnCodeLoad(u32 address, u32 size) {
 }
 
 void CacheBase::OnCodeUnload(u32 address, u32 size) {
-    const u32 end = address + size;
-
     ptr_caches.erase(std::remove_if(ptr_caches.begin(), ptr_caches.end(),
         [&](auto const& cache) {
-            if ((address < cache.addr_end) && (end > cache.addr)) {
+            if ((address < cache.addr_end) && (address + size > cache.addr)) {
                 RemoveRange(cache.addr, cache.addr_end);
                 for (u32 i = cache.addr; i < cache.addr_end; i += Memory::PAGE_SIZE) { page_pointers[i >> Memory::PAGE_BITS] = nullptr; }
                 return true;
--- a/src/core/arm/cache/cache.cpp
+++ b/src/core/arm/cache/cache.cpp
@@ -1,0 +1,155 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/assert.h"
+#include "core/arm/cache/cache.h"
+
+namespace Cache {
+
+CacheBase::CacheBase(bool index_mode, OnClearCb clearcb) : index_mode(index_mode) {
+    page_pointers.fill(nullptr);
+    Clear();
+    SetClearCallback(clearcb);
+    g_cachemanager.RegisterCache(this);
+}
+
+CacheBase::~CacheBase() {
+    g_cachemanager.UnregisterCache(this);
+}
+
+void CacheBase::Clear() {
+    if (OnClearCallback != nullptr) OnClearCallback();
+
+    for (auto& cache : ptr_caches) cache.data.assign(cache.data.size(), nullptr);
+
+    if (index_mode) {
+        blocks_pc.assign(MAX_BLOCKS, INVALID_BLOCK);
+        next_block = num_blocks = 0;
+    }
+}
+
+bool CacheBase::RemoveBlock(u32 pc) {
+    u8** ptr = page_pointers[pc >> Memory::PAGE_BITS];
+    if (ptr != nullptr) {
+        ptr = &ptr[pc & Memory::PAGE_MASK];
+        if (*ptr == nullptr) return false;
+
+        if (index_mode) {
+            const u32 id = pointer_to_id(*ptr);
+            ASSERT(blocks_pc[id] == pc);
+
+            blocks_pc[id] = INVALID_BLOCK;
+            if (id < next_block) next_block = id;
+            while (num_blocks > 0 && blocks_pc[num_blocks - 1] == INVALID_BLOCK) --num_blocks;
+        }
+        *ptr = nullptr;
+        return true;
+    }
+    return false;
+}
+
+bool CacheBase::RemoveRange(u32 start, u32 end) {
+    bool result = false;
+    for (auto& cache : ptr_caches) {
+        for (int i = std::max(start, cache.addr); i < std::min(end, cache.addr_end); ++i) {
+            u8** ptr = &cache.data[i - cache.addr];
+            if (*ptr == nullptr) continue;
+
+            if (index_mode) {
+                const u32 id = pointer_to_id(*ptr);
+                ASSERT(blocks_pc[id] == i);
+
+                blocks_pc[id] = INVALID_BLOCK;
+                if (id < next_block) next_block = id;
+                while (num_blocks > 0 && blocks_pc[num_blocks - 1] == INVALID_BLOCK) --num_blocks;
+            }
+            *ptr = nullptr;
+            result = true;
+        }
+    }
+    return result;
+}
+
+void CacheBase::OnCodeLoad(u32 address, u32 size) {
+    const u32 end = address + size;
+
+    // Check there is no overlapping
+    for (auto const& cache : ptr_caches) ASSERT((address >= cache.addr_end) || (end <= cache.addr));
+
+    ASSERT((address & Memory::PAGE_MASK) == 0 && (size & Memory::PAGE_MASK) == 0);
+
+    BlockPtrCache cache{ address, address + size };
+    cache.data.assign(size, nullptr);
+
+    for (u32 i = address; i < end; i += Memory::PAGE_SIZE) { page_pointers[i >> Memory::PAGE_BITS] = &cache.data[i - address]; }
+    ptr_caches.emplace_back(std::move(cache));
+}
+
+void CacheBase::OnCodeUnload(u32 address, u32 size) {
+    const u32 end = address + size;
+
+    ptr_caches.erase(std::remove_if(ptr_caches.begin(), ptr_caches.end(),
+        [&](auto const& cache) {
+            if ((address < cache.addr_end) && (end > cache.addr)) {
+                RemoveRange(cache.addr, cache.addr_end);
+                for (u32 i = cache.addr; i < cache.addr_end; i += Memory::PAGE_SIZE) { page_pointers[i >> Memory::PAGE_BITS] = nullptr; }
+                return true;
+            }
+            return false;
+        }),
+        ptr_caches.cend());
+}
+
+u8*& CacheBase::GetNewPtr(u32 pc) {
+    DEBUG_ASSERT(!index_mode || next_block == MAX_BLOCKS || ((next_block < MAX_BLOCKS) && blocks_pc[next_block] == INVALID_BLOCK));
+    DEBUG_ASSERT(GetPtr(pc) == nullptr);
+
+    u8** page_ptr = page_pointers[pc >> Memory::PAGE_BITS];
+    if (page_ptr == nullptr) {
+        // pc isnt within mapped code
+        OnCodeLoad(pc & ~Memory::PAGE_MASK, Memory::PAGE_SIZE);
+        page_ptr = page_pointers[pc >> Memory::PAGE_BITS];
+    }
+
+    u8** block_ptr = &page_ptr[pc & Memory::PAGE_MASK];
+
+    DEBUG_ASSERT(*block_ptr == nullptr);
+
+    if (index_mode) {
+        if (next_block == MAX_BLOCKS) Clear();
+
+        blocks_pc[next_block] = pc;
+        *block_ptr = id_to_pointer(next_block);
+
+        do ++next_block; while (next_block <= num_blocks && blocks_pc[next_block] != INVALID_BLOCK);
+        if (next_block > num_blocks) num_blocks++;
+    }
+
+    return *block_ptr;
+}
+
+
+void CacheManager::RegisterCode(u32 address, u32 size) const {
+    for (auto const& cache : caches) cache->OnCodeLoad(address, size);
+}
+
+void CacheManager::UnregisterCode(u32 address, u32 size) const {
+    for (auto const& cache : caches) cache->OnCodeUnload(address, size);
+}
+
+void CacheManager::ClearCache() const {
+    for (auto const& cache : caches) cache->Clear();
+}
+
+void CacheManager::RegisterCache(CacheBase* cache) {
+    caches.push_back(cache);
+}
+
+void CacheManager::UnregisterCache(CacheBase* cache) {
+    caches.erase(std::remove(caches.begin(), caches.end(), cache), caches.end());
+}
+
+CacheManager g_cachemanager;
+
+}
--- a/src/core/arm/cache/cache.cpp
+++ b/src/core/arm/cache/cache.cpp
@@ -2,7 +2,6 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include "common/assert.h"
 #include "core/arm/cache/cache.h"
 
 namespace Cache {
@@ -10,20 +9,20 @@ namespace Cache {
 static std::list<CacheBase*> caches;
 
 void RegisterCode(u32 address, u32 size) {
-    for (auto const& cache : caches) {
-        cache->OnCodeLoad(address, size);
+    for (const auto& cache : caches) {
+        cache->cache_fns.code_load(address, size);
     }
 }
 
 void UnregisterCode(u32 address, u32 size) {
-    for (auto const& cache : caches) {
-        cache->OnCodeUnload(address, size);
+    for (const auto& cache : caches) {
+        cache->cache_fns.code_unload(address, size);
     }
 }
 
 void ClearCache() {
-    for (auto const& cache : caches) {
-        cache->Clear();
+    for (const auto& cache : caches) {
+        cache->cache_fns.clear();
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/core/arm/cache/cache.cpp
```diff
@@ -0,0 +1,155 @@
...
+        ptr = &ptr[pc & Memory::PAGE_MASK];
+        if (*ptr == nullptr) return false;
+
+        if (index_mode) {
+            const u32 id = pointer_to_id(*ptr);
+            ASSERT(blocks_pc[id] == pc);
+
+            blocks_pc[id] = INVALID_BLOCK;
+            if (id < next_block) next_block = id;
+            while (num_blocks > 0 && blocks_pc[num_blocks - 1] == INVALID_BLOCK) --num_blocks;
```
lioncash: please put the body on a new line. Ditto for all other if statements and loops.

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,137 @@
...
+template <typename T>
+class PtrCache final : public CacheBase {
+public:
+    explicit PtrCache(OnClearCb clearcb = nullptr) : CacheBase(false, clearcb) {
+        static_assert(std::is_pointer<T>::value, "T must be a pointer");
+    }
+    ~PtrCache() {}
+
+    /// Get cached pointer for PC
+    T FindPtr(u32 pc) { return reinterpret_cast<T>(GetPtr(pc)); }
```
lioncash: These casts will break strict aliasing for anything that isn't u8*

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,137 @@
...
+
+private:
+    std::list<CacheBase*> caches;
+
+public:
+    void RegisterCache(CacheBase* cache);
+    void UnregisterCache(CacheBase* cache);
+};
+
+extern CacheManager g_cachemanager;
```
lioncash: Have you considered a non-global approach to this? If we ever get to the point of emulating the syscore, this is likely going to cause issues.

src/core/arm/cache/cache.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    for (u32 i = address; i < end; i += Memory::PAGE_SIZE) { page_pointers[i >> Memory::PAGE_BITS] = &cache.data[i - address]; }
+    ptr_caches.emplace_back(std::move(cache));
+}
+
+void CacheBase::OnCodeUnload(u32 address, u32 size) {
+    ptr_caches.erase(std::remove_if(ptr_caches.begin(), ptr_caches.end(),
+        [&](auto const& cache) {
+            if ((address < cache.addr_end) && (address + size > cache.addr)) {
+                RemoveRange(cache.addr, cache.addr_end);
```
wwylele: It seems that GCC wants `this->RemoveRange(cache.addr, cache.addr_end);`

yuriks:
Have you tried using a better hash table implementation first instead? std::unordered_map is held back by its API, so using a non-compatible hash map is likely to offer improvements. I don't have any specific library in mind, but it shouldn't be too terrible to write one to experiment with, see [this article](http://www.reedbeta.com/blog/2015/01/12/data-oriented-hash-table/) for some relevant details.

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -1144,6 +1145,8 @@ static inline void *AllocBuffer(unsigned int size) {
     return (void *)&inst_buf[start];
 }
 
+Cache::PtrCache<u8*> instr_cache([]() {top = 0;});
```
wwylele: The constructor of `instr_cache` relies on `g_cachemanager`, but their init order are not determinate (and causes a SIGSEGV here on my computer).

MerryMage:
fwiw: http://incise.org/hash-table-benchmarks.html

yuriks:
I saw that articles, but it's about 6 years old, so I don't know if the results in it are specially relevant anymore.

src/core/arm/cache/cache.h
Bonta0: How so ? the class doesn't actually access the objects, it merely stores pointers, i changed it to void* rather than u8* to make it more obvious. If that's not enough I'll have to make the base class inherit T meaning I'd have to put all the code to the header

src/core/arm/cache/cache.h
Bonta0: I ditched the "manager" class, it didn't make sense and made static functions instead. That aside, every cache has its own buffer, if you need a different one for the syscore that would cause no issue (perhaps give it a flag to differentiate the cores?). However the interpreter only has one "instruction buffer", since the cache pointers point there it would have to be moved somewhere else

Bonta0:
Regarding performance i've made my own benchmark that would match citra's behavior and the results were the same than what i could notice in citra's performance change

200M random lookups
> calculating overhead: 4977 ms
> unorderedmap lookups: 24411 ms // without overhead: 19434
> myptrcache lookups: 5144 ms // without overhead: 167
> mycache lookups: 5155 ms // without overhead: 178
> vector lookups: 4996 ms // without overhead: 19 << access as vector[pc-base] vector of IMAGE_MAX_SIZE size
> virtualalloc lookups: 4958 ms // without overhead: -19 << access as buffer[pc] buffer of 0xFFFFFFFF size
> googledensehash lookups: 17183 ms // without overhead: 12206 << google dense hash is one the fastest

Hash tables can't be efficient when the key is actually smaller than the hash or of equal size, especially when you know the range you need to access in advance. Most benchmarks focus on inserting and deleting when citra doesn't need any of that (only about ~100k inserts and 0 delete,  unless you clear the cache) but it needs a very fast "get" and you cant be faster than reading straight into in array, no matter how "optimized" the map is
I've also tried using the vector with IMAGE_MAX_SIZE approach however it showed no performance gain in citra, once you're fast enough you're gonna hit diminishing returns, it also comes with a cost: memory consumption

src/core/arm/cache/cache.cpp
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+    return result;
+}
+
+void CacheBase::OnCodeLoad(u32 address, u32 size) {
+    const u32 end = address + size;
+
+    // Check there is no overlapping
+    for (auto const& cache : ptr_caches) {
```
lioncash: This can be a std::any_of

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,136 @@
...
+void RegisterCode(u32 address, u32 size);
+void UnregisterCode(u32 address, u32 size = 1);
+
+/// Clear every cache
+void ClearCache();
+
+
+using OnClearCb = std::function<void()>;
+
+const u32 MAX_BLOCKS = 0x40000;
```
lioncash: constexpr

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,136 @@
...
+
+    std::function<void*(u32)> id_to_pointer;
+    std::function<u32(void*)> pointer_to_id;
+
+private:
+    bool index_mode;
+    OnClearCb OnClearCallback = nullptr;
+
+    std::vector<BlockPtrCache> ptr_caches;
+    std::array<void**, (1 << (32 - Memory::PAGE_BITS))> page_pointers;
```
lioncash: This can be `std::array<void**, (1 << (32 - Memory::PAGE_BITS))> page_pointers{};` which gets rid of the need to explicitly call a fill in the constructor.

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,136 @@
...
+class PtrCache final : public CacheBase {
+public:
+    explicit PtrCache(OnClearCb clearcb = nullptr) : CacheBase(false, clearcb) {
+        static_assert(std::is_pointer<T>::value, "T must be a pointer");
+    }
+    ~PtrCache() {}
+
+    /// Get cached pointer for PC
+    T GetPtr(u32 pc) {
+        return reinterpret_cast<T>(CacheBase::GetPtr(pc));
```
lioncash: These will still break strict aliasing, same as below. `void*` does *not* mean 'wildcard'

src/core/arm/cache/cache.cpp
```diff
@@ -0,0 +1,189 @@
...
+}
+
+void UnregisterCode(u32 address, u32 size) {
+    for (auto const& cache : caches) {
+        cache->OnCodeUnload(address, size);
+    }
+}
+
+void ClearCache() {
+    for (auto const& cache : caches) {
```
lioncash: `const auto&`

src/core/arm/cache/cache.cpp
```diff
@@ -0,0 +1,189 @@
...
+static std::list<CacheBase*> caches;
+
+void RegisterCode(u32 address, u32 size) {
+    for (auto const& cache : caches) {
+        cache->OnCodeLoad(address, size);
+    }
+}
+
+void UnregisterCode(u32 address, u32 size) {
+    for (auto const& cache : caches) {
```
lioncash: `const auto&`

src/core/arm/cache/cache.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+#include "common/assert.h"
+#include "core/arm/cache/cache.h"
+
+namespace Cache {
+
+static std::list<CacheBase*> caches;
+
+void RegisterCode(u32 address, u32 size) {
+    for (auto const& cache : caches) {
```
lioncash: `const auto&`

src/core/arm/cache/cache.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+    for (u32 i = address; i < end; i += Memory::PAGE_SIZE) {
+        page_pointers[i >> Memory::PAGE_BITS] = &cache.data[i - address];
+    }
+    ptr_caches.emplace_back(std::move(cache));
+}
+
+void CacheBase::OnCodeUnload(u32 address, u32 size) {
+    ptr_caches.erase(std::remove_if(ptr_caches.begin(), ptr_caches.end(),
+        [&, this](auto const& cache) {
```
lioncash: const auto&

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,136 @@
...
+template <typename T>
+class Cache final : public CacheBase {
+public:
+    explicit Cache(OnClearCb clearcb = nullptr) : CacheBase(true, clearcb) {
+        id_to_pointer = [this](u32 id) -> void* {
+            return &blocks[id];
+        };
+        pointer_to_id = [this](void* ptr) -> u32 {
+            return static_cast<u32>(std::distance(blocks.begin(),
+                std::find_if(blocks.begin(), blocks.end(), [&](auto const& block) {
```
lioncash: `const auto&`

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,136 @@
...
+                return (reinterpret_cast<T*>(ptr) == &block) ? true : false;
```
lioncash: `? true : false` is redundant.

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,136 @@
...
+    }
+    ~Cache() {}
+
+    /// Get block cached for PC
+    T* GetBlock(u32 pc) {
+        return reinterpret_cast<T*>(GetPtr(pc));
+    }
+
+    /// Allocate block for PC
+    T& GetNewBlock(u32 pc) {
```
lioncash: Inconsistent API return type, considering GetBlock returns a pointer and this one doesn't.

yuriks:
Can you explain a bit the general design here? As best as I can tell, this is simply a 2-level sparse vector. All the additional complexity must be coming from somewhere, but I can't quite tell why it's there.

Bonta0:
My bad, I probably should have explained this better.
You're right, this is simple stuff, especially when it comes to the interpreter this might seem unnecessary. The point of this was first to get an index based cache for MerryMage's JIT as it suffered from the same unordered_map slow lookups and it needs to store more than a pointer per block, that's pretty much how dolphin also handles it, so I figured I might just write a more generic class that would fit the interpreter's needs better (as there's no need to index base it)
I believe it can be useful when you keep those things in mind:
- It's a performance boost and it's free, that can't hurt. those unordered maps need to be replaced
- As you implement a JIT you will end up with something like this in some way, might as well just use what already exists. If the interpreter ever gets redesigned it could be used too I guess
- It's probably cleaner to use these helper functions if you need to clear or invalidate part of the cache for some reason (example CRO) rather than having clearcache functions scattered on different interfaces

All in all, this isnt much and it's just my opinion. I certainly do not have your hindsight, nor your experience so your input is welcome

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+                    }
+                    return true;
+                }
+                return false;
+            }),
+            ptr_caches.cend());
+    }
+
+protected:
+    T GetPtr(u32 pc) const {
```
lioncash: This should probably be `T*`

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+        if (index_mode) {
+            if (next_block == MAX_BLOCKS) {
+                Clear();
+            }
+
+            blocks_pc[next_block] = pc;
+            *block_ptr = index_to_pointer(next_block);
+
+            do {
+                ++next_block;
```
lioncash: this can be a std::count_if

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+    ~PtrCache() {}
+
+    /// Get cached pointer for PC
+    T GetPtr(u32 pc) {
+        return CacheCommon<T>::GetPtr(pc);
+    }
+
+    /// Set pointer for PC to value
+    T SetPtr(u32 pc, T value) {
+        return (*CacheCommon<T>::GetNewPtr(pc) = value);
```
lioncash: Please separate the return from the assignment.

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+    }
+    ~PtrCache() {}
+
+    /// Get cached pointer for PC
+    T GetPtr(u32 pc) {
+        return CacheCommon<T>::GetPtr(pc);
+    }
+
+    /// Set pointer for PC to value
+    T SetPtr(u32 pc, T value) {
```
lioncash: Should this be a `T*` ?

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+template <typename T>
+class PtrCache final : public CacheCommon<T> {
+public:
+    explicit PtrCache(OnClearCb clearcb = nullptr) : CacheCommon<T>(false, clearcb) {
+        static_assert(std::is_pointer<T>::value, "T must be a pointer, use IndexedCache");
+    }
+    ~PtrCache() {}
+
+    /// Get cached pointer for PC
+    T GetPtr(u32 pc) {
```
lioncash: `T*`

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+                result = true;
+            }
+        }
+        return result;
+    }
+
+    void OnCodeLoad(u32 address, u32 size) {
+        // Check there is no overlapping
+        ASSERT(std::none_of(ptr_caches.cbegin(), ptr_caches.cend(),
+            [&](const auto& cache) {
```
lioncash: This can be on the same line as above to make it flow more like a function.

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    std::function<T(u32)> index_to_pointer;
+    std::function<u32(T)> pointer_to_index;
+
+private:
+    OnClearCb OnClearCallback = nullptr;
+    const bool index_mode;
+
+    std::vector<BlockPtrCache<T>> ptr_caches;
```
lioncash: T*

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+        }
+        return false;
+    }
+
+    /// Returns true if at least one block was found, false otherwise
+    bool RemoveRange(u32 start, u32 end) {
+        bool result = false;
+        for (auto& cache : ptr_caches) {
+            for (u32 i = std::max(start, cache.addr); i < std::min(end, cache.addr_end); ++i) {
+                T* ptr = &cache.data[i - cache.addr];
```
lioncash: T**

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+};
+
+template <typename T>
+class CacheCommon : CacheBase {
+protected:
+    explicit CacheCommon(bool index_mode, OnClearCb clearcb) : index_mode(index_mode), CacheBase({
+            [this](u32 start, u32 size) { this->OnCodeLoad(start, size); },
+            [this](u32 start, u32 size) { this->OnCodeUnload(start, size); },
+            [this]() { this->Clear(); }}) {
+        static_assert(std::is_pointer<T>::value, "T must be a pointer");
```
lioncash: `std::is_pointer<T>()`

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+    u32 next_block = 0;
+    u32 num_blocks = 0;
+};
+
+/// Use this if you only need to store a pointer
+template <typename T>
+class PtrCache final : public CacheCommon<T> {
+public:
+    explicit PtrCache(OnClearCb clearcb = nullptr) : CacheCommon<T>(false, clearcb) {
+        static_assert(std::is_pointer<T>::value, "T must be a pointer, use IndexedCache");
```
lioncash: `std::is_pointer<T>()`

src/core/arm/cache/cache.h
```diff
@@ -0,0 +1,292 @@
...
+    }
+};
+
+/// Index based cache
+template <typename T>
+class IndexedCache final : public CacheCommon<T*> {
+public:
+    explicit IndexedCache(OnClearCb clearcb = nullptr) : CacheCommon<T*>(true, clearcb) {
+
+        static_assert(!std::is_pointer<T>::value, "T can't be a pointer, use PtrCache");
```
lioncash: `std::is_pointer<T>()`

lioncash:
Things to note:

There are quite a few instances where semantics are totally unclear. You use `T` but operate on it as if it's a pointer. This is better suited for `T*` which makes intent specific in design.

> It's a performance boost and it's free

Nothing is free. The fact @yuriks couldn't figure it out from first glance shows there's a lot of added complexity.

lioncash:
It seems like the PR opener deleted their account. Closing this for now.

PR #1722, opened 2016-04-27
Externals: Add soundtouch

MerryMage:
This PR just contains SoundTouch. Sumbitted in preparation for another audio-related PR.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "nihstro"]
 	path = externals/nihstro
 	url = https://github.com/neobrain/nihstro.git
+[submodule "soundtouch"]
+	path = externals/soundtouch
+	url = https://github.com/citra-emu/soundtouch.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -249,6 +249,9 @@ if(ENABLE_QT)
     include_directories(externals/qhexedit)
     add_subdirectory(externals/qhexedit)
 endif()
+
+add_subdirectory(externals/soundtouch)
+
 add_subdirectory(src)
 
 # Install freedesktop.org metadata files, following those specifications:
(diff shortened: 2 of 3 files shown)


JayFoxRox:
(direct quote from CONTRIBUTING.md:)

* Don't ever introduce new external dependencies into Core

I do realize we probably need something like this to avoid code bloat on our end, but externals can be a huge PITA to maintain (which is probably why the quote above exists - unless it means using them without adding them to the repo which would be.. well.. worse, but I'm getting off-topic here).

Especially with this being a subdirectory and not a git submodule I'm slightly worried about maintenance.
With that in mind I'd like to know if you looked at any other libraries (possibly slimmer?) and also the purpose of the library will be in Citra. Will it be used for effects by the DSP or is it purely for time-scaling of audio in case of host <=> guest sync / perf issues?

MerryMage:
> is it purely for time-scaling of audio in case of host <=> guest sync / perf issues

Purely the latter. I've looked at and considered one other option ([rubberband](http://breakfastquay.com/rubberband/)) which actually uses an algorithm I prefer and can be easily submoduled, but both audio quality and performance turn out to be abysmal for the frame-by-frame adjustment of tempo Citra requires.

> externals can be a huge PITA to maintain

Would me mirroring my own git repo be perferrable?

MerryMage:
Updated to submodule soundtouch instead.

bunnei:
LGTM

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "nihstro"]
 	path = externals/nihstro
 	url = https://github.com/neobrain/nihstro.git
+[submodule "soundtouch"]
+	path = externals/soundtouch
+	url = https://github.com/citra-emu/soundtouch.git
```
JayFoxRox: I've already said it on IRC (unless my message was dropped) before the merge: Our standard naming scheme for external submodules is `citra-emu/ext-<name>.git`.
We should rename the soundtouch module for consistency.

PR #1723, opened 2016-04-27
AudioCore: Implement interpolation

MerryMage:
Black-box tested against actual DSP firmware. Annoyingly I couldn't get it exact; that would have to wait for white-box RE. Error in values is +/- 1. `scale_factor` chosen for minimum error.

### DIFF
--- a/src/audio_core/interpolate.cpp
+++ b/src/audio_core/interpolate.cpp
@@ -1,0 +1,85 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "audio_core/interpolate.h"
+
+#include "common/assert.h"
+#include "common/math_util.h"
+
+namespace AudioInterp {
+
+// Calculations are done in fixed point with 24 fractional bits.
+// (This is not verified. This was chosen for minimal error.)
+constexpr u64 scale_factor = 1 << 24;
+constexpr u64 scale_mask = scale_factor - 1;
+
+/// Here we step over the input in steps of rate_multiplier, until we consume all of the input.
+/// Three adjacent samples are passed to fn each step.
+template <typename Function>
+static StereoBuffer16 StepOverSamples(State& state, const StereoBuffer16& input, float rate_multiplier, Function fn) {
+    ASSERT(rate_multiplier > 0);
+
+    if (input.size() < 2)
+        return {};
+
+    StereoBuffer16 output;
+    output.reserve(static_cast<size_t>(input.size() / rate_multiplier));
+
+    u64 step_size = static_cast<u64>(rate_multiplier * scale_factor);
+
+    u64 fposition = 0;
+    const u64 max_fposition = input.size() * scale_factor;
+
+    while (fposition < 1 * scale_factor) {
+        u64 fraction = fposition & scale_mask;
+
+        output.push_back(fn(fraction, state.xn2, state.xn1, input[0]));
+
+        fposition += step_size;
+    }
+
+    while (fposition < 2 * scale_factor) {
+        u64 fraction = fposition & scale_mask;
+
+        output.push_back(fn(fraction, state.xn1, input[0], input[1]));
+
+        fposition += step_size;
+    }
+
+    while (fposition < max_fposition) {
+        u64 fraction = fposition & scale_mask;
+
+        size_t index = static_cast<size_t>(fposition / scale_factor);
+        output.push_back(fn(fraction, input[index - 2], input[index - 1], input[index]));
+
+        fposition += step_size;
+    }
+
+    state.xn2 = input[input.size() - 2];
+    state.xn1 = input[input.size() - 1];
+
+    return output;
+}
+
+StereoBuffer16 None(State& state, const StereoBuffer16& input, float rate_multiplier) {
+    return StepOverSamples(state, input, rate_multiplier, [](u64 fraction, const auto& x0, const auto& x1, const auto& x2) {
+        return x0;
+    });
+}
+
+StereoBuffer16 Linear(State& state, const StereoBuffer16& input, float rate_multiplier) {
+    // Note on accuracy: Some values that this produces are +/- 1 from the actual firmware.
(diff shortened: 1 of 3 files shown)


src/audio_core/interpolate.cpp
```diff
@@ -0,0 +1,82 @@
...
+            return x0;
+
+        // This is a saturated subtraction. (Verified.)
+        s64 delta0 = MathUtil::Clamp<s64>(x1[0] - x0[0], -32768, 32767);
+        s64 delta1 = MathUtil::Clamp<s64>(x1[1] - x0[1], -32768, 32767);
+
+        std::array<s16, 2> ret;
+        ret[0] = x0[0] + fraction * delta0 / scale_factor;
+        ret[1] = x0[1] + fraction * delta1 / scale_factor;
+        return ret;
```
lioncash: This can just be:

```cpp
return {
    static_cast<s16>(x0[0] + fraction * delta0 / scale_factor),
    static_cast<s16>(x0[1] + fraction * delta1 / scale_factor)
};
```

which does the same thing (and prevents warnings).

src/audio_core/interpolate.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    while (fposition < 2 * scale_factor) {
+        u64 fraction = fposition & scale_mask;
+
+        if (fposition < 1 * scale_factor) {
+            output.push_back(fn(fraction, state.xn2, state.xn1, input[0]));
+        } else if (fposition < 2 * scale_factor) {
+            output.push_back(fn(fraction, state.xn1, input[0], input[1]));
+        } else {
+            UNREACHABLE();
```
MerryMage: ... in hind-sight this is rather silly

src/audio_core/interpolate.cpp
```diff
@@ -0,0 +1,82 @@
...
+        return x0;
+    });
+}
+
+StereoBuffer16 Linear(State& state, const StereoBuffer16& input, float rate_multiplier) {
+    return StepOverSamples(state, input, rate_multiplier, [](u64 fraction, const auto& x0, const auto& x1, const auto& x2) {
+        if (fraction == 0)
+            return x0;
+
+        // This is a saturated subtraction. (Verified.)
```
lioncash: This should probably indicate in what sense it's verified.

mailwl:
@MerryMage, do you mean RE of DSP CEVA TeakLite? I am not sure if there is instruction set for it in the internet

MerryMage:
@mailwl You don't need an instruction set to do RE, although it helps. (Also this is really off-topic.)

MerryMage:
The above commit in MerryMage/3ds-audio is a related hwtest. [Hardware output for that test](https://gist.github.com/MerryMage/f1dc00374a60032dd275e301c764d04c).

src/audio_core/interpolate.cpp
```diff
@@ -0,0 +1,82 @@
...
+StereoBuffer16 None(State& state, const StereoBuffer16& input, float rate_multiplier) {
+    return StepOverSamples(state, input, rate_multiplier, [](u64 fraction, const auto& x0, const auto& x1, const auto& x2) {
+        return x0;
+    });
+}
+
+StereoBuffer16 Linear(State& state, const StereoBuffer16& input, float rate_multiplier) {
+    return StepOverSamples(state, input, rate_multiplier, [](u64 fraction, const auto& x0, const auto& x1, const auto& x2) {
+        if (fraction == 0)
+            return x0;
```
JayFoxRox: This seems unnecessary, is it done for emulation accuracy (= fw does it), math stability or perf?

MerryMage: Good catch. That's in there because a previous model of how interpolation was done required that case to be handled differently and is no longer required. I'll remove that special case.

MerryMage:
Addressed the above comments + added some code-comments regarding accuracy.

src/audio_core/interpolate.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+/// Here we step over the input in steps of rate_multiplier, until we consume all of the input.
+/// Three adjacent samples are passed to fn each step.
+template <typename Function>
+static StereoBuffer16 StepOverSamples(State& state, const StereoBuffer16& input, float rate_multiplier, Function fn) {
+    if (input.size() < 2)
+        return {};
+
+    StereoBuffer16 output;
+    output.reserve(static_cast<size_t>(input.size() * rate_multiplier));
```
JayFoxRox: What happens if `rate_multiplier <= 0`?

src/audio_core/interpolate.h
```diff
@@ -0,0 +1,39 @@
...
+    // Two historical samples.
+    std::array<s16, 2> xn1 = {}; ///< x[n-1]
+    std::array<s16, 2> xn2 = {}; ///< x[n-2]
+};
+
+/**
+ * No interpolation. This is equivalent to a zero-order hold. There is a two-sample predelay.
+ * @param input Input buffer.
+ * @param rate_multiplier Stretch factor. Must be a positive non-zero value.
+ *                        rate_multiplier > 1.0 performs decimation and rate_multipler < 1.0 performs upsampling.
```
bunnei: Missing @return field

src/audio_core/interpolate.h
```diff
@@ -0,0 +1,39 @@
...
+ */
+StereoBuffer16 None(State& state, const StereoBuffer16& input, float rate_multiplier);
+
+/**
+ * Linear interpolation. This is equivalent to a first-order hold. There is a two-sample predelay.
+ * @param input Input buffer.
+ * @param rate_multiplier Stretch factor. Must be a positive non-zero value.
+ *                        rate_multiplier > 1.0 performs decimation and rate_multipler < 1.0 performs upsampling.
```
bunnei: Missing @return field

bunnei:
LGTM other than the nit. Please squash the fixups and this can be merged.

MerryMage:
Squashed.

src/audio_core/interpolate.h
MerryMage: Added.

src/audio_core/interpolate.h
MerryMage: Added.

PR #1724, opened 2016-04-27
OpenGL State Tracker Immediacy

tfarley:
This is one thought I had for making the state tracker a bit easier/less error prone to use by making it more immediate, and I'm wondering what people think of this as a possibility. I'm definitely open to other suggestions (even ones that involve killing the state tracker somehow) but at the least I have been finding its current incarnation a little too messy for my tastes.

**Main arguments for actually having a state tracker in the first place:**
- Automatically avoids setting redundant state.
- It allows the OpenGL frontend renderer and backend rasterizer to co-exist in the same context without polluting each others' state (less effort to keep consistent).
- It allows for chunks of code to be able to get to a clean/personal state easily without polluting the renderer/rasterizer.
  - i.e. for the rasterizer cache and OpenGL framebuffer blits in general, you don't want a lot of the state that the rasterizer has set (like color/depth/stencil mask, bound framebuffers, etc.), yet you don't want to trash the rasterizer's state as it would then need to be re-set by the rasterizer anyway. The state tracker provides an easier API to replace and restore entire states with minimal state switching calls.
- Could theoretically be helpful as a debugging aid - nice to be able to inspect the current state in-debugger without apitrace or its ilk.

**The changes are:**

Before:
- All state was expressed as public variables of the tracker object.
- Apply() was the only way to actually apply the changes to state variables, which would check each and every state variable for change and make the corresponding OpenGL call if so.

After:
- State changes are broken out into individual setter functions.
- Each setter function immediately makes the state change, so careful placement of Apply() is no longer necessary.
  - (i.e. before, had to be sure to call Apply() _after_ binding a texture but _before_ glTex*() operations)
- Each setter function checks only the state that was changed.
  - This avoids checking all state as in the old Apply(). Assumes:
  - (#state variables \* #Apply calls) > (#state changes + #state variables \* #MakeCurrent calls).
- Apply() has conceptually been transformed into MakeCurrent() which essentially just calls every state setter function with the state tracker's current value for that state - for wholesale switching between trackers.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -149,8 +149,6 @@ void RasterizerOpenGL::DrawTriangles() {
     if (vertex_batch.empty())
         return;
 
-    state.MakeCurrent();
-
     const auto& regs = Pica::g_state.regs;
 
     // Sync and bind the framebuffer surfaces
@@ -246,8 +244,6 @@ void RasterizerOpenGL::DrawTriangles() {
 
 void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
     const auto& regs = Pica::g_state.regs;
-
-    state.MakeCurrent();
 
     switch(id) {
     // Culling
@@ -612,17 +608,15 @@ bool RasterizerOpenGL::AccelerateFill(const GPU::Regs::MemoryFillConfig& config)
 
     SurfaceType dst_type = CachedSurface::GetFormatType(dst_surface->pixel_format);
 
-    OpenGLState* old_state = OpenGLState::GetCurrentState();
-    utility_state.MakeCurrent();
-
-    utility_state.SetDrawFramebuffer(framebuffer.handle);
+    OpenGLState* cur_state = OpenGLState::GetCurrentState();
+    OpenGLState old_state = OpenGLState::ApplyTransferState(0, cur_state->GetReadFramebuffer(), dst_surface->texture.handle, framebuffer.handle);
 
     if (dst_type == SurfaceType::Color || dst_type == SurfaceType::Texture) {
         glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, dst_surface->texture.handle, 0);
         glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D, 0, 0);
 
         if (OpenGLState::CheckBoundFBStatus(GL_DRAW_FRAMEBUFFER) != GL_FRAMEBUFFER_COMPLETE) {
-            old_state->MakeCurrent();
+            OpenGLState::UndoTransferState(old_state);
             return false;
         }
 
@@ -640,7 +634,7 @@ bool RasterizerOpenGL::AccelerateFill(const GPU::Regs::MemoryFillConfig& config)
                 color_values[2] = config.value_24bit_b / 255.0f;
                 break;
             default:
-                old_state->MakeCurrent();
+                OpenGLState::UndoTransferState(old_state);
                 return false;
             }
         } else if (config.fill_32bit) {
@@ -654,7 +648,7 @@ bool RasterizerOpenGL::AccelerateFill(const GPU::Regs::MemoryFillConfig& config)
                 color_values[3] = (value & 0xFF) / 255.0f;
                 break;
             default:
-                old_state->MakeCurrent();
+                OpenGLState::UndoTransferState(old_state);
                 return false;
             }
         } else {
@@ -694,12 +688,11 @@ bool RasterizerOpenGL::AccelerateFill(const GPU::Regs::MemoryFillConfig& config)
                 color_values[1] = (value_16bit & 0xFF) / 255.0f;
                 break;
             default:
-                old_state->MakeCurrent();
+                OpenGLState::UndoTransferState(old_state);
                 return false;
             }
         }
 
-        utility_state.SetColorMask(true, true, true, true);
         glClearBufferfv(GL_COLOR, 0, color_values);
     } else if (dst_type == SurfaceType::Depth) {
         glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, 0, 0);
@@ -707,7 +700,7 @@ bool RasterizerOpenGL::AccelerateFill(const GPU::Regs::MemoryFillConfig& config)
         glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D, 0, 0);
 
         if (OpenGLState::CheckBoundFBStatus(GL_DRAW_FRAMEBUFFER) != GL_FRAMEBUFFER_COMPLETE) {
-            old_state->MakeCurrent();
+            OpenGLState::UndoTransferState(old_state);
             return false;
         }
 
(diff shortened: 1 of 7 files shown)


src/video_core/renderer_opengl/gl_state.h
```diff
@@ -75,27 +131,5 @@ class OpenGLState {
...
-    static void ResetTexture(GLuint handle);
-    static void ResetSampler(GLuint handle);
-    static void ResetProgram(GLuint handle);
-    static void ResetBuffer(GLuint handle);
-    static void ResetVertexArray(GLuint handle);
-    static void ResetFramebuffer(GLuint handle);
-
-private:
-    static OpenGLState cur_state;
+    static OpenGLState* cur_state;
```
lioncash: Is there a particular reason the OpenGLState class is acting as it's own manager instead of having an OpenGLStateManager class or something? This can get rid of the use of static state while keeping everything flexible. For example, the manager can take care of all the sanitizing logic (and provide asserts if necessary, without actually touching the state code itself), while the state class deals with things explicitly only required of itself.

src/video_core/renderer_opengl/gl_state.h
```diff
@@ -66,6 +120,8 @@ class OpenGLState {
         GLuint texture_1d; // GL_TEXTURE_BINDING_1D
     } lighting_luts[6];
```
yuriks: Can you just remove all uses of "LUT" to refer to the 1D textures and call them something generic texture-y instead?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -151,6 +149,8 @@ void RasterizerOpenGL::DrawTriangles() {
     if (vertex_batch.empty())
         return;
 
+    state.MakeCurrent();
```
yuriks: Is this (and others call scattered in the PICA path) necessary? Shouldn't the other users be resetting the current state to the PICA one instead after they're done?

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -360,6 +360,7 @@ class RasterizerOpenGL : public VideoCore::RasterizerInterface {
     void SyncLightPosition(int light_index);
 
     OpenGLState state;
+    OpenGLState utility_state;
```
yuriks: Is there any particular state you want to maintain here instead of creating a new fresh OpenGLState every time?

src/video_core/renderer_opengl/gl_state.h
yuriks: (In the context of the state tracker, that is.)

yuriks:
Mostly LGTM. The only thing I'm really curious about is the random MakeCurrent's that were added.

src/video_core/renderer_opengl/gl_state.h
tfarley: A good point, and I will definitely keep it in mind as I mess with state management more. For now, I'm trying to keep my PRs smaller than they have been though, so this indeed sounds like it could be a subject in another PR.

wwylele:
With this PR citra some times prints out
`Render.OpenGL <Error> video_core/renderer_opengl/renderer_opengl.cpp:DebugHandler:451: API ERROR 13: GL_INVALID_OPERATION in glBindTexture(non-gen name)`

tfarley:
Took this out of RFC and all comments have been addressed.

Additionally, I added a registration mechanism for the states so that when a resource is deleted all states (not just the currently bound one) have any handles to that resource cleared. However, I do want to make sure that the constructor/destructor registration to the std::set is a safe/proper thing to do - any thoughts on that?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -919,23 +904,25 @@ void RasterizerOpenGL::SyncDepthModifiers() {
...
+    state.SetBlendEnabled(Pica::g_state.regs.output_merger.alphablend_enable == 1);
 }
 
 void RasterizerOpenGL::SyncBlendFuncs() {
     const auto& regs = Pica::g_state.regs;
-    state.blend.src_rgb_func = PicaToGL::BlendFunc(regs.output_merger.alpha_blending.factor_source_rgb);
-    state.blend.dst_rgb_func = PicaToGL::BlendFunc(regs.output_merger.alpha_blending.factor_dest_rgb);
-    state.blend.src_a_func = PicaToGL::BlendFunc(regs.output_merger.alpha_blending.factor_source_a);
-    state.blend.dst_a_func = PicaToGL::BlendFunc(regs.output_merger.alpha_blending.factor_dest_a);
+    state.SetBlendFunc(std::make_tuple<>(
```
Subv: I didn't know you could use templates like this, interesting

Subv:
Well, LGTM as far as I can tell, please do bear in mind that I'm no expert in OpenGL.

bunnei:
Lgtm

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -48,229 +54,472 @@ OpenGLState::OpenGLState() {
...
+}
+
+void OpenGLState::UndoTransferState(OpenGLState &old_state) {
+    cur_state->SetColorMask(old_state.GetColorMask());
+    cur_state->SetDepthWriteMask(old_state.GetDepthWriteMask());
+    cur_state->SetStencilWriteMask(old_state.GetStencilWriteMask());
+    cur_state->SetReadFramebuffer(old_state.GetReadFramebuffer());
+    cur_state->SetDrawFramebuffer(old_state.GetDrawFramebuffer());
+
+    // NOTE: Textures that were reset in ApplyTransferState() are NOT restored
```
JayFoxRox: Why? Make this more clear?
I think this is confusing

src/video_core/renderer_opengl/gl_state.h
```diff
@@ -57,13 +156,12 @@ class OpenGLState {
 
     // 3 texture units - one for each that is used in PICA fragment shader emulation
     struct {
+        GLuint texture_1d; // GL_TEXTURE_BINDING_1D
         GLuint texture_2d; // GL_TEXTURE_BINDING_2D
         GLuint sampler; // GL_SAMPLER_BINDING
-    } texture_units[3];
+    } texture_units[9];
```
JayFoxRox: Comment above is wrong then (has 3 in it).

JayFoxRox:
Missing blend equations?
I'll probably review this too, but I don't know when I'll find the time for it.
I'm a little worried that this is riddled with bugs (due to the old tracker having a lot of bugs too and much redundancy: `updateX(new_x) { if x != new_x then x = new_x }`), so this needs very good regression testing imho.

I like the general idea though.

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -48,229 +54,472 @@ OpenGLState::OpenGLState() {
...
-    if (blend.enabled != cur_state.blend.enabled) {
-        if (blend.enabled) {
+void OpenGLState::SetBlendEnabled(bool n_enabled) {
+    if (n_enabled != cur_state->blend.enabled) {
+        if (n_enabled) {
             glEnable(GL_BLEND);
 
-            cur_state.logic_op = GL_COPY;
-            glLogicOp(cur_state.logic_op);
+            SetLogicOp(GL_COPY);
```
JayFoxRox: Why does this affect the logic op in the first place?
Shouldn't the logic op stay what it was? It's disabled anyway..

src/video_core/renderer_opengl/gl_state.h
```diff
@@ -4,10 +4,109 @@
...
+    void SetCullMode(GLenum n_mode);
+    inline GLenum GetCullFrontFace() { return cull.front_face; }
+    void SetCullFrontFace(GLenum n_front_face);
+
+    inline bool GetDepthTestEnabled() { return depth.test_enabled; }
+    void SetDepthTestEnabled(bool n_test_enabled);
+    inline GLenum GetDepthFunc() { return depth.test_func; }
+    void SetDepthFunc(GLenum n_test_func);
+    inline GLboolean GetDepthWriteMask() { return depth.write_mask; }
+    void SetDepthWriteMask(GLboolean n_write_mask);
```
JayFoxRox: nit: I think these (`GLboolean`) should just take a `bool` instead.
I don't really like `GL_TRUE` / `GL_FALSE`. What do others think?

src/video_core/renderer_opengl/gl_state.h
```diff
@@ -4,10 +4,109 @@
...
+    ~OpenGLState();
+
+    /// Get a pointer to the currently bound state tracker object
+    static OpenGLState* GetCurrentState();
+
+    /// Apply this state as the current OpenGL state
+    void MakeCurrent();
+
+    /// Getter and setter functions for OpenGL state
+    inline bool GetCullEnabled() { return cull.enabled; }
```
lioncash: Note that `inline` is also not necessary for these. Any function with a definition directly inside a class body is inline by default.

wwylele:

yuriks:
Closing due to lack of activity. (Even though I think this is a good change, it'll need to be updated. We can re-open it later.)

PR #1726, opened 2016-04-28
AudioCore: CurrentRegion() -> ReadRegion(), WriteRegion()

MerryMage:
The DSP and applications read from one region while writing to another.

Replicating this is required for correct behavior.

### DIFF
--- a/src/audio_core/hle/dsp.cpp
+++ b/src/audio_core/hle/dsp.cpp
@@ -8,8 +8,32 @@
 namespace DSP {
 namespace HLE {
 
-SharedMemory g_region0;
-SharedMemory g_region1;
+std::array<SharedMemory, 2> g_regions;
+
+static size_t CurrentRegionIndex() {
+    // The region with the higher frame counter is chosen unless there is wraparound.
+    // This function only returns a 0 or 1.
+
+    if (g_regions[0].frame_counter == 0xFFFFu && g_regions[1].frame_counter != 0xFFFEu) {
+        // Wraparound has occured.
+        return 1;
+    }
+
+    if (g_regions[1].frame_counter == 0xFFFFu && g_regions[0].frame_counter != 0xFFFEu) {
+        // Wraparound has occured.
+        return 0;
+    }
+
+    return (g_regions[0].frame_counter > g_regions[1].frame_counter) ? 0 : 1;
+}
+
+static SharedMemory& ReadRegion() {
+    return g_regions[CurrentRegionIndex()];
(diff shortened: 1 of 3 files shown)


src/audio_core/hle/dsp.cpp
```diff
@@ -8,8 +8,31 @@
 namespace DSP {
 namespace HLE {
 
-SharedMemory g_region0;
-SharedMemory g_region1;
+std::array<SharedMemory, 2> g_regions;
+
+static size_t CurrentRegion() {
```
JayFoxRox: Maybe rename this to `CurrentRegionIndex` now that this is only an index? I'm not sure about it though.. - what else is used in the codebase?
(Same thing about the other functions)

MerryMage: `CurrentRegionIndex` sounds fine. I use `_index` for my variables in audio_core to indicate they're indicies.

MerryMage:
Renamed.

src/audio_core/hle/dsp.cpp
```diff
@@ -8,8 +8,31 @@
...
+    if (g_regions[1].frame_counter == 0xFFFFu && g_regions[0].frame_counter != 0xFFFEu) {
+        // Wraparound has occured.
+        return 0;
+    }
+
+    return (g_regions[0].frame_counter > g_regions[1].frame_counter) ? 0 : 1;
+}
+
+static SharedMemory& ReadRegion() {
+    return g_regions[CurrentRegionIndex() == 0 ? 0 : 1];
```
yuriks: Just drop the comparison and the ternary, imo.

src/audio_core/hle/dsp.cpp
```diff
@@ -8,8 +8,31 @@
...
+}
+
+static SharedMemory& WriteRegion() {
+    return g_regions[CurrentRegionIndex() == 0 ? 1 : 0];
```
yuriks: And then just `1 - CurrentRegionIndex()` here

yuriks:
LGTM

src/audio_core/hle/dsp.cpp
JayFoxRox: Agreed, but add a comment that CurrentRegionIndex can only return 0 or 1 to the CurrentRegionIndex function (if there isn't one yet).

MerryMage:
Squashed.

bunnei:
LGTM

PR #1727, opened 2016-04-28
AudioCore: Move samples_per_frame and num_sources into hle/common.h

MerryMage:
Very minor change.

Done so that everything under `src/audio_core/hle` doesn't have to `#include "audio_core/audio_core.h"`.

(Done as it's own PR to not clutter up the others.)


bunnei:
LGTM

PR #1729, opened 2016-04-28
Audio Config: Implement null sink and implement sink configuration

MerryMage:
SDL2Sink coming later.

This PR introduces a global list of sink types available, `g_sink_details`. This is intended to be filled in [like this](https://github.com/MerryMage/citra/blob/26092a7071e1a224516c0f094e527c6b746913d6/src/audio_core/sink_details.cpp).

`SinkDetails::id` is not automatically based on the index because that may vary depending on audio libraries available when compiling.

### DIFF
--- a/src/audio_core/audio_core.cpp
+++ b/src/audio_core/audio_core.cpp
@@ -2,9 +2,15 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <memory>
+#include <string>
+
 #include "audio_core/audio_core.h"
 #include "audio_core/hle/dsp.h"
 #include "audio_core/hle/pipe.h"
+#include "audio_core/null_sink.h"
+#include "audio_core/sink.h"
+#include "audio_core/sink_details.h"
 
 #include "core/core_timing.h"
 #include "core/hle/kernel/vm_manager.h"
@@ -28,7 +34,6 @@ static void AudioTickCallback(u64 /*userdata*/, int cycles_late) {
     CoreTiming::ScheduleEvent(audio_frame_ticks - cycles_late, tick_event);
 }
 
-/// Initialise Audio
 void Init() {
     DSP::HLE::Init();
 
@@ -36,7 +41,6 @@ void Init() {
     CoreTiming::ScheduleEvent(audio_frame_ticks, tick_event);
 }
 
-/// Add DSP address spaces to Process's address space.
 void AddAddressSpace(Kernel::VMManager& address_space) {
     auto r0_vma = address_space.MapBackingMemory(DSP::HLE::region0_base, reinterpret_cast<u8*>(&DSP::HLE::g_regions[0]), sizeof(DSP::HLE::SharedMemory), Kernel::MemoryState::IO).MoveFrom();
     address_space.Reprotect(r0_vma, Kernel::VMAPermission::ReadWrite);
@@ -45,10 +49,31 @@ void AddAddressSpace(Kernel::VMManager& address_space) {
     address_space.Reprotect(r1_vma, Kernel::VMAPermission::ReadWrite);
 }
 
-/// Shutdown Audio
+void SelectSink(std::string sink_id) {
+    if (sink_id == "auto") {
+        // Auto-select.
+        // g_sink_details is ordered in terms of desirability, with the best choice at the front.
+        const auto& sink_detail = g_sink_details.front();
+        DSP::HLE::SetSink(sink_detail.factory());
+        return;
+    }
+
+    auto iter = std::find_if(g_sink_details.begin(), g_sink_details.end(), [sink_id](const auto& sink_detail) {
+        return sink_detail.id == sink_id;
+    });
+
+    if (iter == g_sink_details.end()) {
+        LOG_ERROR(Audio, "AudioCore::SelectSink given invalid sink_id");
+        DSP::HLE::SetSink(std::make_unique<NullSink>());
+        return;
+    }
+
+    DSP::HLE::SetSink(iter->factory());
+}
+
 void Shutdown() {
     CoreTiming::UnscheduleEvent(tick_event, 0);
     DSP::HLE::Shutdown();
 }
 
-} //namespace
+} // namespace AudioCore
--- a/src/audio_core/null_sink.h
+++ b/src/audio_core/null_sink.h
@@ -1,0 +1,29 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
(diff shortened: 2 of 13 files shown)


src/audio_core/null_sink.h
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    void EnqueueSamples(const std::vector<s16>&) override {}
+
+    size_t SamplesInQueue() const override {
+        return 0;
+    }
+};
+
+}
```
MerryMage: newline + `namespace` comment

src/audio_core/audio_core.cpp
```diff
@@ -45,10 +45,28 @@ void AddAddressSpace(Kernel::VMManager& address_space) {
...
+void SelectSink(int sink_id) {
+    if (sink_id == 0) {
+        // Auto-select.
+        // g_sink_details is ordered in terms of desirability, with the best choice at the back.
+        const auto& sink_detail = g_sink_details.back();
+        DSP::HLE::SetSink(sink_detail.factory());
+        return;
+    }
+
+    for (const auto& sink_detail : g_sink_details) {
```
lioncash: This can be a find_if

src/audio_core/audio_core.cpp
```diff
@@ -45,10 +45,28 @@ void AddAddressSpace(Kernel::VMManager& address_space) {
     address_space.Reprotect(r1_vma, Kernel::VMAPermission::ReadWrite);
 }
 
-/// Shutdown Audio
+void SelectSink(int sink_id) {
+    if (sink_id == 0) {
+        // Auto-select.
+        // g_sink_details is ordered in terms of desirability, with the best choice at the back.
```
lioncash: This should likely be mentioned at the definition of g_sink_details as well.

src/audio_core/null_sink.h
linkmauve: Missing newline here.

src/citra/default_ini.h
```diff
@@ -56,6 +56,11 @@ bg_red =
 bg_blue =
 bg_green =
 
+[Audio]
+# Which audio output engine to use.
+# 0 (default): Auto-select, 1: Null (No audio output)
+output_sink =
```
yuriks: "sink" is a very unusual term and I doubt anyone will recognize what it means. I'd suggest using "output_engine", "_driver" or "_backend"  or something more common along those lines.

src/audio_core/sink_details.h
```diff
@@ -0,0 +1,30 @@
...
+
+class Sink;
+
+struct SinkDetails {
+    SinkDetails() = delete;
+    SinkDetails(int id_, const char* const name_, std::function<std::unique_ptr<Sink>()> factory_)
+        : id(id_), name(name_), factory(factory_) {}
+
+    /// id must be unique. 0 is reserved.
+    const int id;
```
yuriks: Is there any point in making all members const? The constness of the vector will already ensure they won't be modified.

src/audio_core/sink_details.cpp
```diff
@@ -0,0 +1,17 @@
...
+#include <memory>
+#include <vector>
+
+#include "audio_core/null_sink.h"
+#include "audio_core/sink_details.h"
+
+namespace AudioCore {
+
+const std::vector<SinkDetails> g_sink_details = {
+    { 1, "No Audio", []() { return std::make_unique<NullSink>(); } },
```
yuriks: You should also add a "identifier string" (`null` in this case) which can go in the config file in lieu of the index/numeric id.

src/audio_core/sink_details.h
lioncash: Keeps the data immutable even in non-const scenarios (if it were to ever occur).

src/audio_core/audio_core.cpp
```diff
@@ -45,10 +45,28 @@ void AddAddressSpace(Kernel::VMManager& address_space) {
...
+    }
+
+    for (const auto& sink_detail : g_sink_details) {
+        if (sink_id == sink_detail.id) {
+            DSP::HLE::SetSink(sink_detail.factory());
+            return;
+        }
+    }
+
+    UNREACHABLE_MSG("AudioCore::SelectSink given invalid sink_id");
```
yuriks: This isn't really unreachable: Someone simply has to put an invalid id in the config file

MerryMage: > perhaps by selecting the null sink instead

Hmm, good idea.

src/audio_core/sink_details.h
MerryMage: I like my consts.

yuriks:
Sink selection should be done by a string name/identifier, instead of a numeric one. You're already searching for the id, might as well use a string and make the config file more readable.

That's really the only general comment I have, though I have a feeling this will require some iteration for it to really work.

LGTM

src/audio_core/sink_details.cpp
lioncash: It may be beneficial to re-adapt the string used for "No Audio" into that, since it's not particularly a good idea to have user-facing strings in core.

yuriks: I'm ok with this.

src/audio_core/sink_details.h
yuriks: Yes, but is there any point? Seems like it brings no benefit while making uses of the class more restricted. (At which point someone will just come here and remove the consts, and nothing will have been achieved anyway.)

src/audio_core/sink_details.h
```diff
@@ -0,0 +1,30 @@
...
+#include <functional>
+#include <memory>
+#include <vector>
+
+namespace AudioCore {
+
+class Sink;
+
+struct SinkDetails {
+    SinkDetails() = delete;
```
yuriks: You don't need to delete this if you're already defining another constructor.

src/audio_core/sink_details.cpp
yuriks: Also, I think the numeric id isn't useful anymore with a string identifier, so you can probably drop it then.

src/audio_core/sink_details.h
MerryMage: I did that mainly because there's no real reason for any of the members of that class to be modified.

MerryMage:
Addressed comments above except the one about `const` in `SinkDetail`.

Now uses strings instead of numeric ids.

MerryMage:
Addressed @yuriks's comment about `const` and rebased against master.

bunnei:
LGTM, please rebase/squash

MerryMage:
Done!

PR #1730, opened 2016-04-28
Extract a "Vertex loader" class from command_processor.cpp

hrydgard:
These can later be optimized (JIT-ed) and cached, just like in Dolphin or PPSSPP. Turns out Citra spends quite a lot of time loading vertex data, so this will help. As a bonus, command_processor.cpp gets easier to read.

Later, it would also be nice to refactor away the primitive assembler and use GL-native indexing, then the vertex loader and vertex shader JIT can take a number of vertices to decode in one go and handle their own loops.

Not sure what the best place to keep the MemoryAccesses class is.

I also optimized it, so Citra now spends quite a bit less time on vertex loading even if no smart caching is done yet.

### DIFF
--- a/src/video_core/vertex_loader.cpp
+++ b/src/video_core/vertex_loader.cpp
@@ -1,0 +1,119 @@
+#include <cmath>
+#include <string>
+
+#include "boost/range/algorithm/fill.hpp"
+
+#include "common/assert.h"
+#include "common/alignment.h"
+#include "common/bit_field.h"
+#include "common/common_funcs.h"
+#include "common/common_types.h"
+#include "common/logging/log.h"
+
+#include "core/memory.h"
+
+#include "debug_utils/debug_utils.h"
+
+#include "pica.h"
+#include "pica_state.h"
+#include "pica_types.h"
+#include "vertex_loader.h"
+
+namespace Pica {
+
+void VertexLoader::Setup(const Pica::Regs &regs) {
+    const auto& attribute_config = regs.vertex_attributes;
+    base_address = attribute_config.GetPhysicalBaseAddress();
+    num_total_attributes = attribute_config.GetNumTotalAttributes();
+
+    boost::fill(vertex_attribute_sources, 0xdeadbeef);
+
+    for (int i = 0; i < 16; i++) {
+        vertex_attribute_is_default[i] = attribute_config.IsDefaultAttribute(i);
+    }
+
+    // Setup attribute data from loaders
+    for (int loader = 0; loader < 12; ++loader) {
+        const auto& loader_config = attribute_config.attribute_loaders[loader];
+
+        u32 offset = 0;
+
+        // TODO: What happens if a loader overwrites a previous one's data?
+        for (unsigned component = 0; component < loader_config.component_count; ++component) {
+            if (component >= 12) {
+                LOG_ERROR(HW_GPU, "Overflow in the vertex attribute loader %u trying to load component %u", loader, component);
+                continue;
+            }
+
+            u32 attribute_index = loader_config.GetComponent(component);
+            if (attribute_index < 12) {
+                int element_size = attribute_config.GetElementSizeInBytes(attribute_index);
+                offset = Common::AlignUp(offset, element_size);
+                vertex_attribute_sources[attribute_index] = base_address + loader_config.data_offset + offset;
+                vertex_attribute_strides[attribute_index] = static_cast<u32>(loader_config.byte_count);
+                vertex_attribute_formats[attribute_index] = attribute_config.GetFormat(attribute_index);
+                vertex_attribute_elements[attribute_index] = attribute_config.GetNumElements(attribute_index);
+                vertex_attribute_element_size[attribute_index] = element_size;
+                offset += attribute_config.GetStride(attribute_index);
+            } else if (attribute_index < 16) {
+                // Attribute ids 12, 13, 14 and 15 signify 4, 8, 12 and 16-byte paddings, respectively
+                offset = Common::AlignUp(offset, 4);
+                offset += (attribute_index - 11) * 4;
+            } else {
+                UNREACHABLE(); // This is truly unreachable due to the number of bits for each component
+            }
+        }
+    }
+}
+
+void VertexLoader::LoadVertex(int index, int vertex, Shader::InputVertex &input, MemoryAccesses &memory_accesses) {
+    for (int i = 0; i < num_total_attributes; ++i) {
+        if (vertex_attribute_elements[i] != 0) {
+            // Default attribute values set if array elements have < 4 components. This
+            // is *not* carried over from the default attribute settings even if they're
+            // enabled for this attribute.
+            static const float24 zero = float24::FromFloat32(0.0f);
+            static const float24 one = float24::FromFloat32(1.0f);
+            input.attr[i] = Math::Vec4<float24>(zero, zero, zero, one);
+
+            // Load per-vertex data from the loader arrays
+            for (unsigned int comp = 0; comp < vertex_attribute_elements[i]; ++comp) {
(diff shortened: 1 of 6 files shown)


src/video_core/vertex_loader.h
```diff
@@ -0,0 +1,53 @@
...
+namespace Pica {
+
+class MemoryAccesses {
+    /// Combine overlapping and close ranges
+    void SimplifyRanges() {
+        for (auto it = ranges.begin(); it != ranges.end(); ++it) {
+            // NOTE: We add 32 to the range end address to make sure "close" ranges are combined, too
+            auto it2 = std::next(it);
+            while (it2 != ranges.end() && it->first + it->second + 32 >= it2->first) {
+                it->second = std::max(it->second, it2->first + it2->second - it->first);
```
lioncash: `<iterator>` and `<algorithm>` need to be included.

src/video_core/vertex_loader.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "common/logging/log.h"
+
+#include "core/memory.h"
+
+#include "debug_utils/debug_utils.h"
+
+#include "pica.h"
+#include "pica_state.h"
+#include "pica_types.h"
+#include "vertex_loader.h"
```
lioncash: These includes should be relative like the others.

lioncash:
Making this a general comment since it's in more than one file:

The common convention we use for pointers and references is that they're against the type and not the variable

src/video_core/vertex_loader.h
lioncash: and `<map>` as well (initially missed that).

src/video_core/vertex_loader.cpp
hrydgard: wasn't sure about that since I saw several conventions in different files, but sure.

hrydgard:
All comments so far addressed.

src/video_core/vertex_loader.cpp
lioncash: Yeah, it's unfortunate, since it gets confusing. I'll need to do a pass over video_core and straighten that out.

hrydgard:
Found some low hanging fruit in the vertex loader, some simple optimization got it almost twice as fast (from 5-6% to ~3% of execution time in a vertex heavy scene).

JayFoxRox:
I'll check the code later.

I just wanted to say that we might not be able to use GL native indexing or caching (easily) because we still have to process shaders in the JIT and the Pica has 4 different shader units (SU), each with it's own register state. Vertices are scheduled to an SU (how?) and iirc the Pica has 32 cache slots (which afaik is shared across all SU). If a vertex is not in the cache we might have to run the shader again or the SU register state might be broken afterwards.
- Most of this is based on assumptions about very few Pica facts. But when this is due it should still be taken into consideration in case we are aiming at emulation accuracy.

It might not even be used in games (yet, and maybe never will) with the notable exception being geometry shaders which do depend on correct execution order as they will use the register state across invocations.

hrydgard:
As long as vertex shaders are "pure" (1 vertex in, 1 vertex out, no persistent state), which should be the majority of cases, it shouldn't matter at all how they are scheduled on the SUs. Assuming we can detect this, it should be perfectly safe to run all the shaders in a big batch and then use native indexing. Of course, we might still need to keep a more generic, slower path properly simulating the vertex cache for when these assumptions don't hold.

Either way, this issue isn't relevant to this particular pull request - except, I suppose, my comment about it.

Subv:
Side note, stuff related to CiTraces should either be fixed or removed from the code, they only bloat it up if it doesn't work.

src/video_core/vertex_loader.h
```diff
@@ -0,0 +1,54 @@
...
+
+        // Simplify ranges...
+        SimplifyRanges();
+    }
+
+    /// Map of accessed ranges (mapping start address to range size)
+    std::map<u32, u32> ranges;
+};
+
+class VertexLoader {
```
bunnei: `final`

src/video_core/vertex_loader.h
```diff
@@ -0,0 +1,54 @@
...
+public:
+    void Setup(const Pica::Regs& regs);
+    void LoadVertex(u32 base_address, int index, int vertex, Shader::InputVertex& input, MemoryAccesses& memory_accesses);
+
+    int GetNumTotalAttributes() const { return num_total_attributes; }
+
+private:
+    u32 vertex_attribute_sources[16];
```
bunnei: Why are some of these members initialized and others not?

src/video_core/vertex_loader.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <iterator>
+#include <algorithm>
+
+#include "video_core/pica.h"
+#include "video_core/shader/shader.h"
+
+namespace Pica {
+
+class MemoryAccesses {
```
bunnei: Kind of a nit, but it sort of bothers me that this is here.... either:
* It's a pretty generic utility thing, it probably could go somewhere in common. Also, it's just collapsing ranges... I feel like there should be a less-manual way to do this via Boost or something, although nothing personally comes to mind.
* Even if you wanted to leave it here (if it were something VertexLoader specific thing), it pollutes the Pica namespace with a very generic name..

bunnei:
Code mostly looks fine to me, and I think this is a good change. No need to keep this RFC, IMO.

hrydgard:
@Subv  What is wrong or broken with it? I tried to keep everything working the way it was (except that the tracer now logs per attribute instead of per component).

src/video_core/vertex_loader.h
hrydgard: I kept things the same as it was before. Could probably safely not initialize most of them.

src/video_core/vertex_loader.h
hrydgard: Previously it was an anonymous class, I had to give it some name in order to break out the vertex loader. It's only used when the debug recorder is enabled.

I don't really think it's necessary at all - it's a debug tracer anyway, why not record everything separately. But it's there so I kept it - removing it completely should be a separate change.

Tell me where you would prefer it to live and a name and I'll move it - or I can get rid of it altogether.

src/video_core/vertex_loader.h
yuriks: All the `vertex_` in the names feels redundant given the class they're in.

src/video_core/vertex_loader.h
yuriks: `MemoryAccessTracker` sounds like a fine name. We can decide what to do with it later.

yuriks: Also, it can be moved to debug_utils.h if that doesn't cause a cycle.

yuriks:
LGTM :tada:

hrydgard:
Alright, moved it to debug_utils. Should be ready now :)

src/video_core/vertex_loader.h
yuriks: poke?

JayFoxRox:
I'm not sure if the MemoryAccessTracker should be part of DebugUtils.
We potentially need something like this elsewhere in the code to sync HW renderer + CPU memory. So we can probably abuse something like this to mark pages dirty etc.
However, just saying this for the record - I think it's fine eitherway for now.

src/video_core/vertex_loader.h
hrydgard: Just kept the original names, but you're right.

Either way, the implementation of this class will change a lot later so not sure it's worth bothering. However, if you feel at all strongly about it, I can rename them.

lioncash:
LGTM

bunnei:
LGTM

Subv:
@hrydgard It's not related to this PR, specifically, but our recorded CiTraces (what the MemoryAccesses class is for) are pretty much worthless, the CiTracePlayer doesn't work on hardware nor on citra, and nobody wants to fix it, so I'm in favor of removing the relevant codepaths from the source.

Issue #1600, opened 2016-03-28
Mario vs Donkey Kong Tipping Stars - Config block 0x170000 with flags 2 and size 4 was not found


JohnGodgames:
Dupe of https://github.com/citra-emu/citra/issues/1486

[LOG](http://pasted.co/cd380882)

JohnGodgames:
Fixed by https://github.com/citra-emu/citra/pull/1732

Issue #1486, opened 2016-03-10
Chibi-Robo Zip Lash! - Config block 0x170000 with flags 2 and size 4 was not found


JohnGodgames:
Also it shows critical error with result **0xD90103FA** (Cfg:GetCountryCodeString)

https://www.3dbrew.org/wiki/Cfg:GetCountryCodeString

[LOG](http://pastebin.com/f7kqE3EH)

newretro:
Works fine with config_savegame from fw10.4

![02](https://cloud.githubusercontent.com/assets/17784684/13704597/c5d6332c-e7ad-11e5-8c1a-1c69847a7c59.jpg)

JohnGodgames:
Fixed by https://github.com/citra-emu/citra/pull/1732

Issue #1258, opened 2015-11-28
Missing Config block 0x170000 in BOXBOY!


exhalatio:
Hi!
Missing Config block 0x170000 in BOXBOY!
The following error message.

[  16.957686] Service.CFG <Error> core\hle\service\cfg\cfg.cpp:Service::CFG::GetConfigInfoBlock:236: Config block 0x170000 with flags 2 and size 4 was not found

Punk34:
You need to dump "Config Savegame" from your 3DS

Punk34:
This game probably requires the latest version, that is not working properly on Citra
https://github.com/citra-emu/citra/issues/1226

yuriks:
I'll look into adding support for generating dummy data for this block at some point, thanks.

JohnGodgames:
Fixed by https://github.com/citra-emu/citra/pull/1732

PR #1732, opened 2016-04-29
add config block 0x00170000; remove duplicated content

wwylele:
An easy patch. Fixes Kirby's Robobot Planet (and maybe some other games) that wants this block. Value (zero) and flag (0xE) is taken from my real console's config savedata.
Edit: please help test #1258, #1486 and #1600.


wwylele:
Add two minor fixes: remove a duplicated function declaration; remove a duplicated `CreateConfigInfoBlk` for username

bunnei:
LGTM

PR #1733, opened 2016-04-29
VertexLoader: Minor changes

lioncash:
Superficial things that weren't really necessary to be brought up in the initial PR that introduced it.

cc @hrydgard For one thing I wasn't really certain of. Does it make sense to allow setting up the same VertexLoader more than once (this is likely "probably no", but I just want to make sure)? If it doesn't I can modify the relevant PR to just early return  (or assert) if it's already set up.

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -199,9 +199,8 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
 
             // Processes information about internal vertex attributes to figure out how a vertex is loaded.
             // Later, these can be compiled and cached.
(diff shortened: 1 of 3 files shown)


hrydgard:
Yeah, a vertex loader is only intended to be set up once.

The exact data stored in one will probably change a little btw, like the base address for each loader may have to be read dynamically, depending on how games do things..

bunnei:
Code LGTM. Reminder to rebase.

Subv:
LGTM. Please rebase so we can merge this soon

hrydgard:
LGTM

PR #1735, opened 2016-04-30
Remove TGA dumper

JayFoxRox:
Nobody used it and we also have the more sophisticated `void DumpTexture(const Pica::Regs::TextureConfig& texture_config, u8* data)` in debug_utils which does roughly the same (exports PNG).

Even the comment was wrong - it actually requires RGB8 input, not RGBA8.

I'll probably clean up debug_utils when improving [my various debug widgets](https://github.com/JayFoxRox/citra/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+in%3Atitle+Debugger+) so `DumpTexture` will see more use.


yuriks:
LGTM

linkmauve:
LGTM.

PR #1736, opened 2016-04-30
AudioCore: SDL2 Sink

MerryMage:
A note on `SDL2_FOUND`: `find_package` and `find_package_handle_standard_args` set it as appropriate when the user isn't using bundled SDL2.


src/audio_core/sdl2_sink.cpp
```diff
@@ -0,0 +1,124 @@
...
+    SDL_UnlockAudioDevice(impl->audio_device_id);
+}
+
+size_t SDL2Sink::SamplesInQueue() const {
+    if (impl->audio_device_id <= 0)
+        return 0;
+
+    SDL_LockAudioDevice(impl->audio_device_id);
+
+    size_t total_size = std::accumulate(impl->queue.begin(), impl->queue.end(), 0, [](size_t sum, const auto& buffer) {
```
MerryMage: should probably cast `0` to `size_t` so the correct type is used.

src/audio_core/sdl2_sink.cpp
```diff
@@ -0,0 +1,125 @@
...
+    desired_audiospec.channels = 2;
+    desired_audiospec.freq = native_sample_rate;
+    desired_audiospec.samples = 1024;
+    desired_audiospec.userdata = impl.get();
+    desired_audiospec.callback = &Impl::Callback;
+
+    SDL_AudioSpec obtained_audiospec;
+    SDL_zero(obtained_audiospec);
+
+    impl->audio_device_id = SDL_OpenAudioDevice(nullptr, /*iscapture=*/false, &desired_audiospec, &obtained_audiospec, 0);
```
bunnei: Whats with the inline comment?
Would rather a one-line comment preceding this statement explaining why one might want to enable `iscapture`, if it's important.

src/audio_core/sdl2_sink.cpp
```diff
@@ -0,0 +1,125 @@
...
+    if (impl->audio_device_id <= 0) {
+        LOG_CRITICAL(Audio_Sink, "SDL_OpenAudioDevice failed");
+        return;
+    }
+
+    impl->sample_rate = obtained_audiospec.freq;
+
+    SDL_PauseAudioDevice(impl->audio_device_id, /*pause_on=*/0);
```
bunnei: Ditto

src/audio_core/sdl2_sink.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    SDL_UnlockAudioDevice(impl->audio_device_id);
+
+    return total_size;
+}
+
+void SDL2Sink::Impl::Callback(void* impl_, u8* buffer, int buffer_size_in_bytes) {
+    Impl* impl = reinterpret_cast<Impl*>(impl_);
+
+    size_t remaining_size = static_cast<size_t>(buffer_size_in_bytes) / sizeof(s16); // Convert from u8 to s16.
```
bunnei: Poor wording on the comment, IMO. "Keep track of size in 16-bit increments" (or something)

src/audio_core/sdl2_sink.cpp
```diff
@@ -0,0 +1,125 @@
...
+        return native_sample_rate;
+
+    return impl->sample_rate;
+}
+
+void SDL2Sink::EnqueueSamples(const std::vector<s16>& samples) {
+    if (impl->audio_device_id <= 0)
+        return;
+
+    ASSERT(samples.size() % 2 == 0);
```
bunnei: `ASSERT_MSG`

Perhaps pedantic, but most people probably don't know what it means if this assertion fails

src/audio_core/sdl2_sink.cpp
MerryMage: Ah, I just like to do that with boolean parameters, since it tends not to be obvious what they do. Will remove.

`iscapture` is not important for this usecase (and neither is it currently implemented in SDL2 either anyway).

src/audio_core/sdl2_sink.cpp
MerryMage: Will remove, perhaps adding a comment saying this unpauses the audio device.

MerryMage:
Addressed comments.

bunnei:
LGTM. Please squash the fixups.

MerryMage:
Squashed

src/audio_core/sdl2_sink.cpp
```diff
@@ -0,0 +1,126 @@
...
+    return impl->sample_rate;
+}
+
+void SDL2Sink::EnqueueSamples(const std::vector<s16>& samples) {
+    if (impl->audio_device_id <= 0)
+        return;
+
+    ASSERT_MSG(samples.size() % 2 == 0, "Samples must be in interleaved stereo PCM16 format (size must be a multiple of two)");
+
+    SDL_LockAudioDevice(impl->audio_device_id);
```
Subv: Why are these lock calls needed here but not in ::Callback?

Subv:
LGTM

PR #1737, opened 2016-04-30
AudioCore: Implement time stretcher

MerryMage:
ratio = wallclock time / emulated time

ratio is [exponentially smoothed](https://en.wikipedia.org/wiki/Exponential_smoothing) which acts as a crude low-pass filter. (This is something dolphin doesn't do and probably explains it's bad stretched audio.)


src/audio_core/time_stretch.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    std::vector<s16> samples = GetSamples();
+    if (samples_in_queue >= DROP_FRAMES_SAMPLE_DELAY) {
+        samples.clear();
+        LOG_DEBUG(Audio, "Dropping frames!");
+    }
+    return samples;
+}
+
+TimeStretcher::TimeStretcher() : impl(new Impl) {
```
lioncash: make_unique

src/audio_core/time_stretch.h
```diff
@@ -0,0 +1,58 @@
...
+     * @return Samples to play in interleaved stereo PCM16 format.
+     */
+    std::vector<s16> Process(size_t sample_delay);
+
+private:
+    struct Impl;
+    std::unique_ptr<Impl> impl;
+
+    /// INTERNAL: Clamp ratio within limits.
+    static double ClampRatio(double ratio);
```
lioncash: This can probably be a static local function since it doesn't rely on internal state in any way.

src/audio_core/time_stretch.cpp
```diff
@@ -0,0 +1,136 @@
...
+        ratio = ratio > 1.0 ? ratio * ratio : sqrt(ratio);
+    } else if (sample_delay > max_sample_delay) {
+        // Make the ratio smaller.
+        ratio = ratio > 1.0 ? sqrt(ratio) : ratio * ratio;
+    }
+
+    return ClampRatio(ratio);
+}
+
+std::vector<short> TimeStretcher::GetSamples() {
```
lioncash: If I'm not mistaken, this can also be const, since member is a pointer, and const only applies to the pointer itself, not the instance it points to.

MerryMage: Since `SoundTouch::receiveSamples` changes the internal state of `impl->soundtouch` (draining an internal buffer), should it really be const?

lioncash: In that case, probably not.

MerryMage:
thanks, updated

src/audio_core/time_stretch.cpp
```diff
@@ -0,0 +1,136 @@
...
+    impl->soundtouch.setSampleRate(native_sample_rate);
+}
+
+TimeStretcher::~TimeStretcher() {
+    impl->soundtouch.clear();
+}
+
+void TimeStretcher::SetOutputSampleRate(unsigned int sample_rate) {
+    impl->sample_rate = static_cast<double>(sample_rate);
+    impl->soundtouch.setRate(impl->sample_rate / static_cast<double>(native_sample_rate));
```
hrydgard: When trying this in PPSSPP, I found that division is the wrong way around, unless I made a mistake somewhere.

MerryMage: Good catch.

MerryMage:
Fixed inverted Rate.

bunnei:
Code LGTM. Call for other reviews on this one...

src/audio_core/hle/dsp.cpp
```diff
@@ -39,12 +40,20 @@ static SharedMemory& WriteRegion() {
 }
 
 static std::unique_ptr<AudioCore::Sink> sink;
+static AudioCore::TimeStretcher time_stretcher;
```
Subv: Does this keep any internal state that may have to be reset when restarting emulation without closing Citra?

src/audio_core/time_stretch.cpp
Subv: ```GetSamples``` is defined as returning ```std::vector<s16>``` in the header, but returns ```std::vector<short>``` here

src/audio_core/hle/dsp.cpp
MerryMage: Yes but it quickly self-corrects; I'll reset it anyway, gimme a sec.

MerryMage:
Addressed @Subv's comments above.

Subv:
LGTM

mailwl:
@MerryMage may you add slider to change volume to gui when you complete sound support? if it crossplatform...

PR #1741, opened 2016-04-30
Fix video_core includes (and dependencies) using include-what-you-use

linkmauve:
This has only be tested on gcc 5.3 and clang 3.7 currently.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -5,9 +5,12 @@
 #pragma once
 
 #include <array>
-#include <cmath>
 #include <cstddef>
 #include <string>
+
+#ifndef _MSC_VER
+#include <type_traits> // for std::enable_if
+#endif
 
 #include "common/assert.h"
 #include "common/bit_field.h"
@@ -15,8 +18,6 @@
 #include "common/common_types.h"
 #include "common/vector_math.h"
 #include "common/logging/log.h"
-
-#include "pica_types.h"
 
 namespace Pica {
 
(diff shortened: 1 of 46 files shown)


Subv:
It seems to pass the MSVC build test successfully

linkmauve:
I just did a test run with both compilers on my i5-3210m, the changes aren’t high but are statistically significant (2.57% and 2.43% improvement respectively for gcc and clang):

```
gcc 5.3.0:
make -j4  502.56s user 27.26s system 382% cpu 2:18.60 total

old gcc 5.3.0:
make -j4  513.56s user 27.83s system 380% cpu 2:22.25 total

clang 3.7.1:
make -j4  419.78s user 11.81s system 382% cpu 1:52.90 total

old clang 3.7.1:
make -j4  431.44s user 12.05s system 383% cpu 1:55.71 total
```

Subv:
Visual Studio can generate them, but I don't know if there's a way to do that with Clang or GCC.

MerryMage:
Include graphs, before and after:

https://gist.github.com/MerryMage/c77948b18544ac875586be32662dc826

10712 includes → 9124 includes (15% improvement)

src/video_core/pica.h
```diff
@@ -5,19 +5,20 @@
 #pragma once
 
 #include <array>
-#include <cmath>
 #include <cstddef>
 #include <string>
 
+#ifndef _MSC_VER
```
MerryMage: I don't think one should exclude MSVC here.

linkmauve: This header is only used on line 41 (and only if the includer used the PICA_REG_INDEX_WORKAROUND macro), so only on ¬MSVC.

MerryMage: Ah, makes sense, apologies.

MerryMage:
LGTM

bunnei:
LGTM

PR #1742, opened 2016-04-30
RFC: Move more code into the JIT-ed part of the vertex shaders

hrydgard:
The shader "prologue" (move from fetched vertex into input registers) and "epilogue" (move from output registers to output vertex struct according to semantics, clamp color value) are now moved into the compiled vertex shader.

Note that in the unlikely case that a game uses multiple different setups for which attribute is used for each input, or which output is used for which semantic, duplicate shaders will be generated. I don't yet have any numbers to see if that's a real issue.

In a quick test, this improves performance by 2-3% in vertex heavy scenes.

### DIFF
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -40,7 +40,8 @@ void Setup() {
         u64 cache_key =
             Common::ComputeHash64(&g_state.vs.program_code, sizeof(g_state.vs.program_code)) ^
             Common::ComputeHash64(&g_state.vs.swizzle_data, sizeof(g_state.vs.swizzle_data)) ^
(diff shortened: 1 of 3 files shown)


JayFoxRox:
I'm personally against this for now because I still have my [giant geometry shader PR](https://github.com/JayFoxRox/citra/pull/2) coming up.
\- I wouldn't want to rewrite that completly at this point.

It will be far easier to first merge GS and then work on something like this when we know all constraints (because we also have to move data from/to a GS buffer for example and we don't know how exactly Pica does it yet).

If it wasn't for that I'd be in favor of this though

hrydgard:
Right, fair enough. Seems like a good idea to merge that first indeed.

PR #1745, opened 2016-05-01
OpenGL: Copy TevStageConfig using a loop. Fixes bug: const_color not copied

JayFoxRox:
Looks like someone forgot to copy `const_color`.. Instead of adding another block of code I figured it's easier to loop over the stages.

Only tested OoT and saw no regressions.

No idea if this fixes anything.

(I originally wanted to include this fix in #1704 but I figured I should do this separately for now because I still don't have a good solution for said PR and I'm not sure when I will..)

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.h
+++ b/src/video_core/renderer_opengl/gl_rasterizer.h
@@ -39,36 +39,17 @@ struct PicaShaderConfig {
         res.alpha_test_func = regs.output_merger.alpha_test.enable ?
             regs.output_merger.alpha_test.func.Value() : Pica::Regs::CompareFunc::Always;
 
-        // Copy relevant TevStageConfig fields only. We're doing this manually (instead of calling
-        // the GetTevStages() function) because BitField explicitly disables copies.
-
-        res.tev_stages[0].sources_raw = regs.tev_stage0.sources_raw;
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -39,36 +39,17 @@ struct PicaShaderConfig {
...
-        res.tev_stages[0].scales_raw = regs.tev_stage0.scales_raw;
-        res.tev_stages[1].scales_raw = regs.tev_stage1.scales_raw;
-        res.tev_stages[2].scales_raw = regs.tev_stage2.scales_raw;
-        res.tev_stages[3].scales_raw = regs.tev_stage3.scales_raw;
-        res.tev_stages[4].scales_raw = regs.tev_stage4.scales_raw;
-        res.tev_stages[5].scales_raw = regs.tev_stage5.scales_raw;
+        // Copy tev stages
+        const auto& tev_stages = regs.GetTevStages();
+        DEBUG_ASSERT(res.tev_stages.size() == tev_stages.size());
+        for (unsigned i = 0; i < tev_stages.size(); i++) {
```
lioncash: `size_t`

tfarley:
LGTM

bunnei:
LGTM

PR #1753, opened 2016-05-03
Support additional screen layouts.

jroweboy:
Allows users to choose to show only one screen or to flip bottom and top screens. Adds a configuration option in QT to change this as well.

### DIFF
--- a/src/citra/default_ini.h
+++ b/src/citra/default_ini.h
@@ -63,6 +63,16 @@ use_scaled_resolution =
 # 0 (default): Off, 1: On
 use_vsync =
 
+[Layout]
+# Layout for the screen inside the render window.
+# 0 (default): Default Top Bottom Screen, 1: Single Screen Only, 2: Large Screen Small Screen
+layout_option =
+
+# Swaps the prominent screen with the other screen.
+# For example, if Single Screen is chosen, setting this to 1 will display the bottom screen instead of the top screen.
+# 0 (default): Top Screen is prominent, 1: Bottom Screen is prominent
+swap_screen =
+
 # The clear color for the renderer. What shows up on the sides of the bottom screen.
 # Must be in range of 0.0-1.0. Defaults to 1.0 for all.
 bg_red =
--- a/src/common/emu_window.h
+++ b/src/common/emu_window.h
@@ -7,6 +7,7 @@
 #include <tuple>
 #include <utility>
 #include "common/common_types.h"
+#include "common/framebuffer_layout.h"
 #include "common/math_util.h"
 #include "core/hle/service/hid/hid.h"
 
@@ -36,23 +37,6 @@ public:
         int res_width;
         int res_height;
         std::pair<unsigned, unsigned> min_client_area_size;
-    };
-
-    /// Describes the layout of the window framebuffer (size and top/bottom screen positions)
-    struct FramebufferLayout {
-
-        /**
-         * Factory method for constructing a default FramebufferLayout
-         * @param width Window framebuffer width in pixels
-         * @param height Window framebuffer height in pixels
-         * @return Newly created FramebufferLayout object with default screen regions initialized
-         */
-        static FramebufferLayout DefaultScreenLayout(unsigned width, unsigned height);
-
-        unsigned width;
-        unsigned height;
-        MathUtil::Rectangle<unsigned> top_screen;
-        MathUtil::Rectangle<unsigned> bottom_screen;
     };
 
     /// Swap buffers to display the next frame
@@ -211,9 +195,15 @@ public:
       * Gets the framebuffer layout (width, height, and screen regions)
       * @note This method is thread-safe
       */
-    const FramebufferLayout& GetFramebufferLayout() const {
+    const Layout::FramebufferLayout& GetFramebufferLayout() const {
         return framebuffer_layout;
     }
+
+    /**
+     * Convenience method to update the VideoCore EmuWindow
+     * Read from the current settings to determine which layout to use.
+     */
+    void UpdateCurrentFramebufferLayout(unsigned width, unsigned height);
 
 protected:
     EmuWindow() {
@@ -250,7 +240,7 @@ protected:
      * Update framebuffer layout with the given parameter.
      * @note EmuWindow implementations will usually use this in window resize event handlers.
      */
-    void NotifyFramebufferLayoutChanged(const FramebufferLayout& layout) {
+    void NotifyFramebufferLayoutChanged(const Layout::FramebufferLayout& layout) {
         framebuffer_layout = layout;
     }
 
@@ -274,7 +264,7 @@ private:
         // By default, ignore this request and do nothing.
     }
 
-    FramebufferLayout framebuffer_layout; ///< Current framebuffer layout
+    Layout::FramebufferLayout framebuffer_layout; ///< Current framebuffer layout
 
     unsigned client_area_width;  ///< Current client width, should be set by window impl.
     unsigned client_area_height; ///< Current client height, should be set by window impl.
--- a/src/common/emu_window.h
+++ b/src/common/emu_window.h
@@ -200,7 +200,7 @@ public:
     }
 
     /**
-     * Convenience method to update the VideoCore EmuWindow
+     * Convenience method to update the current frame layout
      * Read from the current settings to determine which layout to use.
      */
     void UpdateCurrentFramebufferLayout(unsigned width, unsigned height);
--- a/src/common/framebuffer_layout.cpp
+++ b/src/common/framebuffer_layout.cpp
@@ -14,8 +14,6 @@ static const float TOP_SCREEN_ASPECT_RATIO =
     static_cast<float>(VideoCore::kScreenTopHeight) / VideoCore::kScreenTopWidth;
 static const float BOT_SCREEN_ASPECT_RATIO =
     static_cast<float>(VideoCore::kScreenBottomHeight) / VideoCore::kScreenBottomWidth;
-static const float BOT_TO_TOP_SCREEN_RATIO_DIFFERENCE =
-    BOT_SCREEN_ASPECT_RATIO - TOP_SCREEN_ASPECT_RATIO;
 
 // Finds the largest size subrectangle contained in window area that is confined to the aspect ratio
 template <class T>
@@ -23,8 +21,8 @@ static MathUtil::Rectangle<T> maxRectangle(MathUtil::Rectangle<T> window_area,
                                            float screen_aspect_ratio) {
     float scale = std::min(static_cast<float>(window_area.GetWidth()),
                            window_area.GetHeight() / screen_aspect_ratio);
-    return MathUtil::Rectangle<T>{0, 0, static_cast<T>(scale),
-                                  static_cast<T>(scale * screen_aspect_ratio)};
+    return MathUtil::Rectangle<T>{0, 0, static_cast<T>(std::round(scale)),
+                                  static_cast<T>(std::round(scale * screen_aspect_ratio))};
 }
 
 FramebufferLayout DefaultFrameLayout(unsigned width, unsigned height, bool swapped) {
--- a/src/common/framebuffer_layout.cpp
+++ b/src/common/framebuffer_layout.cpp
@@ -51,11 +51,15 @@ FramebufferLayout DefaultFrameLayout(unsigned width, unsigned height, bool swapp
             bot_screen.TranslateX((screen_window_area.GetWidth() - bot_screen.GetWidth()) / 2);
     } else {
         // Window is narrower than the emulation content => apply borders to the top and bottom
-        top_screen = top_screen.TranslateY(height / 2 - top_screen.GetHeight());
         // Recalculate the bottom screen to account for the width difference between top and bottom
         screen_window_area = {0, 0, width, top_screen.GetHeight()};
         bot_screen = maxRectangle(screen_window_area, BOT_SCREEN_ASPECT_RATIO);
         bot_screen = bot_screen.TranslateX((top_screen.GetWidth() - bot_screen.GetWidth()) / 2);
+        if (swapped) {
+            bot_screen = bot_screen.TranslateY(height / 2 - bot_screen.GetHeight());
+        } else {
+            top_screen = top_screen.TranslateY(height / 2 - top_screen.GetHeight());
+        }
     }
     // Move the top screen to the bottom if we are swapped.
     res.top_screen = swapped ? top_screen.TranslateY(height / 2) : top_screen;
--- a/src/common/framebuffer_layout.cpp
+++ b/src/common/framebuffer_layout.cpp
@@ -201,47 +201,40 @@ static FramebufferLayout LargeFrameLayout(unsigned width, unsigned height) {
     ASSERT(width > 0);
     ASSERT(height > 0);
 
-    FramebufferLayout res {width, height, true, true, {}, {}};
+    FramebufferLayout res{ width, height, true, true,{},{} };
 
-    float window_aspect_ratio = static_cast<float>(height) / width;
-    float emulation_aspect_ratio = static_cast<float>(VideoCore::kScreenTopHeight * 4) /
-        (VideoCore::kScreenTopWidth * 4 + VideoCore::kScreenBottomWidth);
+    float window_aspect_ratio = static_cast<float>(width) / height;
+    float top_screen_aspect_ratio = static_cast<float>(VideoCore::kScreenTopWidth) /
+        VideoCore::kScreenTopHeight;
 
-    if (window_aspect_ratio > emulation_aspect_ratio) {
+    int viewport_height = static_cast<int>(std::round((width - VideoCore::kScreenBottomWidth) /
+        top_screen_aspect_ratio));
+    int viewport_width = static_cast<int>(std::round((height * top_screen_aspect_ratio) +
+        VideoCore::kScreenBottomWidth));
+    float emulation_aspect_ratio = static_cast<float>(width) / viewport_height;
+
+    if (window_aspect_ratio < emulation_aspect_ratio) {
         // Window is narrower than the emulation content => apply borders to the top and bottom
-        int viewport_height = static_cast<int>(std::round(emulation_aspect_ratio * width));
-
         res.top_screen.left = 0;
-        // Top screen occupies 4 / 5ths of the total width
-        res.top_screen.right = static_cast<int>(std::round(width / 5)) * 4;
+        res.top_screen.right = width - VideoCore::kScreenBottomWidth;
         res.top_screen.top = (height - viewport_height) / 2;
-        res.top_screen.bottom = res.top_screen.top + viewport_height;
-
-        int bottom_height = static_cast<int>((static_cast<float>(VideoCore::kScreenBottomHeight) /
-            VideoCore::kScreenBottomWidth) * (width - res.top_screen.right));
+        res.top_screen.bottom = viewport_height + res.top_screen.top;
 
         res.bottom_screen.left = res.top_screen.right;
         res.bottom_screen.right = width;
         res.bottom_screen.bottom = res.top_screen.bottom;
-        res.bottom_screen.top = res.bottom_screen.bottom - bottom_height;
+        res.bottom_screen.top = res.bottom_screen.bottom - VideoCore::kScreenBottomHeight;
     } else {
         // Otherwise, apply borders to the left and right sides of the window.
-        int viewport_width = static_cast<int>(std::round(height / emulation_aspect_ratio));
-        // Break the viewport into fifths and give top 4 of them
-        int fifth_width = static_cast<int>(std::round(viewport_width / 5));
-
         res.top_screen.left = (width - viewport_width) / 2;
-        res.top_screen.right = res.top_screen.left + fifth_width * 4;
+        res.top_screen.right = (top_screen_aspect_ratio * height) + res.top_screen.left;
         res.top_screen.top = 0;
         res.top_screen.bottom = height;
 
-        int bottom_height = static_cast<int>((static_cast<float>(VideoCore::kScreenBottomHeight) /
-            VideoCore::kScreenBottomWidth) * (fifth_width));
-
         res.bottom_screen.left = res.top_screen.right;
-        res.bottom_screen.right = width - (width - viewport_width) / 2;
-        res.bottom_screen.bottom = res.top_screen.bottom;
-        res.bottom_screen.top = res.bottom_screen.bottom - bottom_height;
+        res.bottom_screen.right = res.bottom_screen.left + VideoCore::kScreenBottomWidth;
+        res.bottom_screen.bottom = height;
+        res.bottom_screen.top = height - VideoCore::kScreenBottomHeight;
     }
 
     return res;
@@ -254,45 +247,38 @@ static FramebufferLayout LargeFrameLayout_Swapped(unsigned width, unsigned heigh
 
     FramebufferLayout res {width, height, true, true, {}, {}};
 
-    float window_aspect_ratio = static_cast<float>(height) / width;
-    float emulation_aspect_ratio = static_cast<float>(VideoCore::kScreenBottomHeight * 4) /
-        (VideoCore::kScreenBottomWidth * 4 + VideoCore::kScreenTopWidth);
+    float window_aspect_ratio = static_cast<float>(width) / height;
+    float bottom_screen_aspect_ratio = static_cast<float>(VideoCore::kScreenBottomWidth) /
+        VideoCore::kScreenBottomHeight;
 
-    if (window_aspect_ratio > emulation_aspect_ratio) {
+    int viewport_height = static_cast<int>(std::round((width - VideoCore::kScreenTopWidth) /
+        bottom_screen_aspect_ratio));
+    int viewport_width = static_cast<int>(std::round((height * bottom_screen_aspect_ratio) +
+        VideoCore::kScreenTopWidth));
+    float emulation_aspect_ratio = static_cast<float>(width) / viewport_height;
+
+    if (window_aspect_ratio < emulation_aspect_ratio) {
         // Window is narrower than the emulation content => apply borders to the top and bottom
-        int viewport_height = static_cast<int>(std::round(emulation_aspect_ratio * width));
-
         res.bottom_screen.left = 0;
-        // Top screen occupies 4 / 5ths of the total width
-        res.bottom_screen.right = static_cast<int>(std::round(width / 5)) * 4;
+        res.bottom_screen.right = width - VideoCore::kScreenTopWidth;
         res.bottom_screen.top = (height - viewport_height) / 2;
-        res.bottom_screen.bottom = res.bottom_screen.top + viewport_height;
-
-        int top_height = static_cast<int>((static_cast<float>(VideoCore::kScreenTopHeight) /
-            VideoCore::kScreenTopWidth) * (width - res.bottom_screen.right));
+        res.bottom_screen.bottom = viewport_height + res.bottom_screen.top;
 
         res.top_screen.left = res.bottom_screen.right;
         res.top_screen.right = width;
         res.top_screen.bottom = res.bottom_screen.bottom;
-        res.top_screen.top = res.top_screen.bottom - top_height;
+        res.top_screen.top = res.top_screen.bottom - VideoCore::kScreenTopHeight;
     } else {
         // Otherwise, apply borders to the left and right sides of the window.
-        int viewport_width = static_cast<int>(std::round(height / emulation_aspect_ratio));
-        // Break the viewport into fifths and give top 4 of them
-        int fifth_width = static_cast<int>(std::round(viewport_width / 5));
-
         res.bottom_screen.left = (width - viewport_width) / 2;
-        res.bottom_screen.right = res.bottom_screen.left + fifth_width * 4;
+        res.bottom_screen.right = (bottom_screen_aspect_ratio * height) + res.bottom_screen.left;
         res.bottom_screen.top = 0;
         res.bottom_screen.bottom = height;
 
-        int top_height = static_cast<int>((static_cast<float>(VideoCore::kScreenTopHeight) /
-            VideoCore::kScreenTopWidth) * (fifth_width));
-
         res.top_screen.left = res.bottom_screen.right;
-        res.top_screen.right = width - (width - viewport_width) / 2;
-        res.top_screen.bottom = res.bottom_screen.bottom;
-        res.top_screen.top = res.top_screen.bottom - top_height;
+        res.top_screen.right = res.top_screen.left + VideoCore::kScreenTopWidth;
+        res.top_screen.bottom = height;
+        res.top_screen.top = height - VideoCore::kScreenTopHeight;
     }
 
     return res;
--- a/src/common/framebuffer_layout.cpp
+++ b/src/common/framebuffer_layout.cpp
@@ -1,0 +1,312 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <cmath>
+
+#include "common/assert.h"
+#include "common/framebuffer_layout.h"
+#include "video_core/video_core.h"
+
+namespace Layout {
+static FramebufferLayout DefaultFrameLayout(unsigned width, unsigned height) {
+
+    ASSERT(width > 0);
+    ASSERT(height > 0);
+
+    FramebufferLayout res {width, height, true, true, {}, {}};
+
+    float window_aspect_ratio = static_cast<float>(height) / width;
+    float emulation_aspect_ratio = static_cast<float>(VideoCore::kScreenTopHeight * 2) /
+        VideoCore::kScreenTopWidth;
+
+    if (window_aspect_ratio > emulation_aspect_ratio) {
+        // Window is narrower than the emulation content => apply borders to the top and bottom
+        int viewport_height = static_cast<int>(std::round(emulation_aspect_ratio * width));
+
+        res.top_screen.left = 0;
+        res.top_screen.right = res.top_screen.left + width;
+        res.top_screen.top = (height - viewport_height) / 2;
+        res.top_screen.bottom = res.top_screen.top + viewport_height / 2;
+
+        int bottom_width = static_cast<int>((static_cast<float>(VideoCore::kScreenBottomWidth) /
+            VideoCore::kScreenTopWidth) * (res.top_screen.right - res.top_screen.left));
+        int bottom_border = ((res.top_screen.right - res.top_screen.left) - bottom_width) / 2;
+
+        res.bottom_screen.left = bottom_border;
+        res.bottom_screen.right = res.bottom_screen.left + bottom_width;
+        res.bottom_screen.top = res.top_screen.bottom;
+        res.bottom_screen.bottom = res.bottom_screen.top + viewport_height / 2;
+    } else {
+        // Otherwise, apply borders to the left and right sides of the window.
+        int viewport_width = static_cast<int>(std::round(height / emulation_aspect_ratio));
+
+        res.top_screen.left = (width - viewport_width) / 2;
+        res.top_screen.right = res.top_screen.left + viewport_width;
+        res.top_screen.top = 0;
+        res.top_screen.bottom = res.top_screen.top + height / 2;
+
+        int bottom_width = static_cast<int>((static_cast<float>(VideoCore::kScreenBottomWidth) /
+            VideoCore::kScreenTopWidth) * (res.top_screen.right - res.top_screen.left));
+        int bottom_border = ((res.top_screen.right - res.top_screen.left) - bottom_width) / 2;
+
+        res.bottom_screen.left = res.top_screen.left + bottom_border;
+        res.bottom_screen.right = res.bottom_screen.left + bottom_width;
+        res.bottom_screen.top = res.top_screen.bottom;
+        res.bottom_screen.bottom = res.bottom_screen.top + height / 2;
+    }
+
+    return res;
+}
+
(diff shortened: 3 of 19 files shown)


src/core/settings.h
```diff
@@ -38,7 +38,14 @@ static const std::array<Values, NUM_INPUTS> All = {{
     CUP, CDOWN, CLEFT, CRIGHT
 }};
 }
-
+namespace Layout {
```
lioncash:
```cpp
enum class LayoutOption {
   Default,
   TopOnly,
   BottomOnly,
   BottomFirst
}
```

JohnGodgames:
Great it seems to work.What about realtime switching with keyboard keys (F1 / F2...).It would be very helpful, than change the settings in the config...

KillzXGaming:
Changing the position of the frames would also be nice where bottom could be more on the bottom right or where ever. Example:
http://i.imgur.com/VFpE1cZ.png

bunnei:
Regarding @KillzXGaming's comments, that layout with 4X resolution for the top screen and 1X for the bottom screen fits perfectly on 1080p (making the effective screen area 1920x960).

That's how all of the HD screenshots on citra-emu.org were taken.

jroweboy:
@JohnGodgames Switching layouts by hot keys is a good idea, but I would hate to see them all mapped since it'll probably end up being 7 things in that drop down. I'll look into adding them to the hot key configuration. Also I am considering adding in a custom one as well, but I don't know what the UX for changing the sizes should be. I'm still trying to figure out the best way to get a user to change it. That should be a seperate PR I believe. For now, I'll just add that one into the config file and allow a user to change that in the ini.

@KillzXGaming I'm planning on adding in a 7th option for a custom layout, but I haven't thought of a good way to allow a person to edit that in the GUI yet. For now, I'll just add it to the ini and do a little more work on that. (Not sure if that should be a seperate PR yet, I guess i'll see what it takes to implement first and then decide)

@bunnei I'll add that in as options 5 and 6 then. Large Top and Large Bottom.

So in summary, I'll add 3 more options for now. Large Top, Large Bottom, and Custom where the two large ones are 4x : 1x size and the Custom one will be configurable through the ini. In the future, I will try to add an option to add more Custom layout configurations that you can name as well, but that might be in a different PR.

jroweboy:
Addressed @lioncash 's comment. I didn't add documentation there since it was already on the functions that create the layouts, but I can add a short blurb there too if you think its a good idea.

Added in two more options for layouts. 4x top and 1x bottom. and the opposite 4x bottom and 1x top. Still planning on looking into hotkeys for them and finishing working on the custom layout as well. 

I'm looking for ideas on how to implement a custom layout in the GUI (as in the UX for it) and I'm taking suggestions for better names for the different layouts as well. 

If anyone else wants to review this, I would love to have someone double check my math to make sure the screen ratios are correct especially the 4x ones since i wrote those at like 2AM.

slx7R4GDZM:
Here's a weird idea for the custom screen layout UI. Not sure if it's all possible.
untitled

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -481,7 +487,7 @@ bool RendererOpenGL::Init() {
 
     if (GLAD_GL_KHR_debug) {
         glEnable(GL_DEBUG_OUTPUT);
-        glDebugMessageCallback(DebugHandler, nullptr);
+        glDebugMessageCallback(reinterpret_cast<GLDEBUGPROC>(DebugHandler), nullptr);
```
JayFoxRox: Please remove this. Submit a bug report if your compiler warns here.
We should not cast pointers to functions - it's dangerous because someone can just change the DebugHandler signature and everything will break.

src/citra_qt/configure_general.ui
```diff
@@ -101,6 +101,56 @@
...
+              <item>
+               <property name="text">
+                <string notr="true">4x Top 1x Bottom</string>
+               </property>
+              </item>
+              <item>
+               <property name="text">
+                <string notr="true">4x Bottom 1x Top</string>
+               </property>
+              </item><!--
```
JayFoxRox: All of this sounds really complicated.
Can't you just let users pick the resolution freely?
Then add another checkbox to hide / show either of the screens.

We could also simply split off the renderer into widgets which can have their border turned off, then users can freely move them accross their desktop and also maximize them etc whereever they want. If we have a right click menu we can even hide the window border. We'd only need a possible cinema-mode (per host screen) so everything except the citra widgets can be hidden.

jroweboy: That was the original plan, and its still on my radar, but that made things much more complicated. We'd then have 1-2 drawing spaces depending on whether they wanted it in the citra window or if they wanted multiple windows. It would be some more work since i'd have to refactor how the EmuWindow works since a lot of things just assume one emu window. I had some more code that I wrote that added a new class that hid whether there was one emu window or two, but something went wrong and the screen flashed black every other frame. I decided to cut my losses for now and go back to working on it once I had some layouts written. (My idea is to have two EmuWindows and one uses the Top Only layout and the other uses Bottom Only so it really isn't a waste to have this code either way.)

At any rate, I still want to do multiple windows, but that work is for a separate PR.

src/video_core/renderer_opengl/renderer_opengl.cpp
jroweboy: Ah, it was just an intellisense error, not a real compile error. Fixed.

src/citra_qt/configure_general.ui
jroweboy: My dream is to remove some of this in favor of a "Edit Layout" style approach similar to OBS where you can move and resize the different screens and show and hide them. But that sounds harder than just making two windows, and I bet some people just don't care to have that much about where their screens are placed, so having a few strong defaults like 4x Top is still a good idea in my opinion.

jroweboy:
@slx7R4GDZM Thats a good start. I think it would be better to use percent of window size instead of resolution and just have a small text to the side to tell them the resolution if they cared to know if. Putting a size directly means that we can't resize when the window size changes since we have no scaling factor to determine how to stretch the screen. Also, what is the Force Screen Plane to Top option do? I'm not sure what thats supposed to do. I can probably roll with this and make a mock up of this later tonight.

slx7R4GDZM:
Great idea with the percents, that'd probably be a lot better. The idea for forcing the screen plane to top would be if the screens were overlapping it would allow for you to make one appear over the other. My idea with that was for games that don't really use the bottom screen much just for things like a map or something you could put it in the corner and still have the top screen take up most of the window. Although it would probably be best to use radio buttons with just top/bottom for the setting, I have no idea why I decided to make it a drop down list.
untitled

JohnGodgames:
@jroweboy  You can add also rotation of screens (needed by some games).It's not a priority...

jroweboy:
@JohnGodgames Thats something I can look into adding in the custom layout. I'm going to try to redo the logic for layouts to include a swap screen checkbox instead of having two similar copies of almost the same layout. From the user perspective, if you are playing with top only, you can press the swap screen hotkey and switch to playing bottom only.

raven02:
Just wonder if either set to top or bottom only , do we stop the rendering engine in behind ? so as to speed up

KillzXGaming:
@raven02 no it won't. The bottom is still in control with key bindings, removing it would likely make using menus and stuff not possible on only top screen. I doubt the bottom screen would provide a significant speed boost if removed anyways as the slow speed is caused by the interpreter.

tywald:
I would like to see an option to make the bottom screen as an overlay to the top screen that I can resize and set the opacity how I want(I would most likely make it hidden) and make it appear when I mouse over, (the opactiy changes to 100%).

It would most likely be my go-to option if I only have 1 monitor available.

Hopefully it doesn't sound like a bad idea :P

i30817:
Instead of a menu with positions i think a better option would be a very simple 'graphical layout manager' like in IDEs with only two windows. Click a button and you can drag them, either to snap to sides, or on top of others, and resize the one on top. Fullscreen would just be a resizing to fit with aspect ratio preserved.

Probably a hell of a lot of work though, i don't envy you.

jroweboy:
Updated the PR. Now there are less layouts in the drop down and there is a checkbox which switches the top and bottom screen (currently citra-qt only. I'll add in SDL support soon) I figure this is a good change for this PR since it was a little overwhelming adding all these different layout options, when to the user they are just duplicates. 

Splitting rendering into multiple windows is still on the agenda, but I think its going to require lots of work so it should be a different PR that can build on this one. I'm guessing that the multiple screen PR can still use the framebufferlayout singlescreen layout, just one reversed so that one window has only the top and the other has the bottom.

jroweboy:
Any idea why the build failed on mac?

MerryMage:
@jroweboy add `#include <cmath>` to `framebuffer_layout.cpp` probably

jroweboy:
Rebased and updated to work with sdl as well.

kreedz1:
tried to merge this for testing, conflicts with latest mater.

neobrain:

2016-06-30T22:46:07Z: https://github.com/citra-emu/citra/compare/de79b6fc4857fde1df3c8a4a2b38d231f2cc961d...jroweboy:5b20c92f67c0798eb2c635c1e91b92c4d4d9a02a

jroweboy:
@kreedz1 I just rebased it. It should merge fine against master now.

JohnGodgames:
I'm getting this with this PR now:
 `MSB6006 "cmd.exe" exited with code 1. citra-qt C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets 171`

neobrain:

2016-07-03T00:30:12Z: https://github.com/citra-emu/citra/compare/de79b6fc4857fde1df3c8a4a2b38d231f2cc961d...jroweboy:d7d531b3064b7d28a3660e142c99d22f27fd687e

jroweboy:
Fixed the merge error. Looks like I need a little work on the layout tab layout. http://i.imgur.com/fWFV1fL.png But other than that, is there any other feedback?

kreedz1:
Tested this and i really like it, i think it should be merged, here's some screens : 
85082fe3b84642dcc705ff827c74af6c
e3d61c657e6f9ddc9c6e24789da23d99

@jroweboy My feedback would be :
add option to switch main layer to the right and small layer to the left and keep perspective in addition to the swap layers option

also in the custom layout parameter you could let the user choose the resolution of each screen in addition to the ability to drag the screen with the resolution values changing.

src/citra_qt/configure_layout.ui
```diff
@@ -0,0 +1,92 @@
...
+       <property name="title">
+        <string>Custom Layout Parameters</string>
+       </property>
+       <layout class="QHBoxLayout" name="horizontalLayout_5">
+        <item>
+         <layout class="QVBoxLayout" name="verticalLayout_3"/>
+        </item>
+       </layout>
+      </widget>
+     </item>
```
wwylele: I'd prefer to add a vertical spacer, like how audio tab and debug tab do.

src/citra_qt/configure_general.cpp
```diff
@@ -35,5 +35,4 @@ void ConfigureGeneral::applyConfiguration() {
     Settings::values.use_hw_renderer = ui->toogle_hw_renderer->isChecked();
     Settings::values.use_shader_jit = ui->toogle_shader_jit->isChecked();
     Settings::values.use_scaled_resolution = ui->toogle_scaled_resolution->isChecked();
-    Settings::Apply();
```
wwylele: Why is this removed?

Leo626:
I'm not sure if this is related to Scaled Resolution but it looks worse with it when the screen is swapped in Large Screen layout. I test this with games that primarily use the bottom screen.

**Scaled Resolution On**
citra-qt_2016-07-03_12-38-15
**Scaled Resolution Off**
citra-qt_2016-07-03_12-38-25

src/common/emu_window.h
```diff
@@ -215,6 +199,12 @@ class EmuWindow
         return framebuffer_layout;
     }
 
+    /**
+     * Convenience method to update the VideoCore EmuWindow
```
wwylele: "VideoCore EmuWindow"?

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -386,6 +386,8 @@ void RendererOpenGL::DrawSingleScreenRotated(const ScreenInfo& screen_info, floa
  */
 void RendererOpenGL::DrawScreens() {
     auto layout = render_window->GetFramebufferLayout();
+    auto top_screen = layout.top_screen;
+    auto bottom_screen = layout.bottom_screen;
```
wwylele: Is these added just for reducing `layout` in the code below? If so, then should use `const auto&` I think.

neobrain:

2016-07-04T16:41:18Z: https://github.com/citra-emu/citra/compare/de79b6fc4857fde1df3c8a4a2b38d231f2cc961d...jroweboy:246af93c631532eca4630c66b585acf99d31e0a9

src/citra_qt/configure_general.cpp
jroweboy: Oh, I committed a partial change here. Thanks for the catch. Each tab was calling Settings::Apply instead of just having each tab set their configuration and applying at the end. My latest commit addresses this.

jroweboy:
Leaving this in here to remind me to test this out. Basically, in the past there were issues with the config that JayFoxRox fixed, and my changes are fiddling with the same places, so I need to make sure that its now okay to make those changes. Below is the conversation from IRC

> JayFoxRox: jroweboy: I added that when the original issue has been that starting citra wouldn't respect the config, it would load it, but afair it did it too late so it would not be applied properly (this always caused SW rendering + shader interpreter and people started complaining about perf regressions. The menu did still show that HW renderer + jit!)
JayFoxRox: https://github.com/citra-emu/citra/pull/1659 < this is the PR which fixed the issue, but it's documented like shit because we wanted to fix it asap
JayFoxRox: (which is not a good reason, but w/e)
JayFoxRox: point being: test Qt, test SDL and also test with various configs - also test the gdbstub etc (because you must apply settings again for the port to change for real)

neobrain:

2016-09-03T03:12:48Z: https://github.com/citra-emu/citra/compare/fed59e07148bbfa5517570bdfdcafde535b0da71...jroweboy:7a0a018dec6a721d5ad5cfdddffcd6922fcbbeeb

JohnGodgames:
Tested many games recently without any issues @bunnei review?Otherwise LGTM

SonofUgly:
Using the Large Screen layout doesn't quite work as expected. The top screen ends up getting squished horizontally, and the bottom screen doesn't stay at 1x. You can see this in the screenshots above, and here's a [1080p screen.](http://i.imgur.com/FqgPXa7.png) The top is 1536*960 (should be 1600*960) and the bottom is 384*288 (should be 320*240).

src/citra_qt/configure.ui
```diff
@@ -44,6 +44,11 @@
          <string>Audio</string>
        </attribute>
      </widget>
+     <widget class="ConfigureLayout" name="layoutTab">
```
bunnei: IMO, this doesn't need to be its own tab. Instead, can you move this to the Graphics tab?
Quick mockup:

src/citra_qt/configure_layout.ui
```diff
@@ -0,0 +1,92 @@
...
+          </item>
+         </layout>
+        </item>
+       </layout>
+      </widget>
+     </item>
+     <item>
+      <widget class="QGroupBox" name="groupBox">
+       <property name="title">
+        <string>Custom Layout Parameters</string>
```
bunnei: Remove this if it's unused ?

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,338 @@
...
+void FramebufferLayout::ReverseFrames(bool is_reverse) {
+    if (is_reverse != is_reverse_layout) {
+        *this = inverse(width, height);
+    }
+}
+
+FramebufferLayout::FramebufferLayout() {
+    height = width = 0;
+    top_screen_enabled = bottom_screen_enabled = 0;
+
```
bunnei: Remove extra space

src/common/framebuffer_layout.h
```diff
@@ -0,0 +1,65 @@
...
+     * Conditionally changes the layout to its inverse. For example, if a layout has a prominent top frame and a small bottom frame,
+     * this function should switch them so the bottom frame is now large.
+     * @param is_reverse If this is true, and the layout is a reverse layout, then nothing will change. This is a toggle
+     */
+    void ReverseFrames(bool is_reverse);
+
+    FramebufferLayout();
+    FramebufferLayout& operator= (const FramebufferLayout &cSource);
+
+    unsigned width;
```
bunnei: Initialize all of these here

src/common/framebuffer_layout.h
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Conditionally changes the layout to its inverse. For example, if a layout has a prominent top frame and a small bottom frame,
+     * this function should switch them so the bottom frame is now large.
+     * @param is_reverse If this is true, and the layout is a reverse layout, then nothing will change. This is a toggle
+     */
+    void ReverseFrames(bool is_reverse);
+
+    FramebufferLayout();
+    FramebufferLayout& operator= (const FramebufferLayout &cSource);
```
bunnei: Why do you need this?

src/common/framebuffer_layout.h
```diff
@@ -0,0 +1,65 @@
...
+
+    unsigned width;
+    unsigned height;
+    bool top_screen_enabled;
+    bool bottom_screen_enabled;
+    MathUtil::Rectangle<unsigned> top_screen;
+    MathUtil::Rectangle<unsigned> bottom_screen;
+private:
+
+    typedef FramebufferLayout (*InverseLayout)(unsigned, unsigned);
```
bunnei: Using "using" keyword instead of typedef

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,338 @@
...
+#include "video_core/video_core.h"
+
+void FramebufferLayout::ReverseFrames(bool is_reverse) {
+    if (is_reverse != is_reverse_layout) {
+        *this = inverse(width, height);
+    }
+}
+
+FramebufferLayout::FramebufferLayout() {
+    height = width = 0;
```
bunnei: Initialize all of these in the class definition and you can get rid of this

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+#include <cmath>
+
+#include "common/assert.h"
+#include "common/framebuffer_layout.h"
+#include "video_core/video_core.h"
+
+void FramebufferLayout::ReverseFrames(bool is_reverse) {
+    if (is_reverse != is_reverse_layout) {
+        *this = inverse(width, height);
```
bunnei: Doing it this way is weird IMO and unnecessarily complicated. You can just make FramebufferLayout be a POD struct and get rid of all this class stuff. Then just have these functions to construct it:
<pre>FramebufferLayout GetDefaultLayout(unsigned width, unsigned height, bool is_swapped);
FramebufferLayout GetSingleLayout(unsigned width, unsigned height, bool is_swapped);
FramebufferLayout GetLargeLayout(unsigned width, unsigned height, bool is_swapped);</pre>

And then the impl of these can just wrap internal functions, e.g.:
<pre>FramebufferLayout GetDefaultLayout(unsigned width, unsigned height, bool is_swapped) {
    return is_swapped ? GetDefaultLayout_Swapped(width, height) : GetDefaultLayout(width, height);
}</pre>

src/common/framebuffer_layout.h
bunnei: Apparently for the weird inverse(..) stuff... If you do my above suggestion, this can go away, as well as InverseLayout

masaki1988:
merge it finaly)))))) its work greay

neobrain:

2016-09-12T04:06:13Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...jroweboy:c85e87e51d39788bfda266703aa33328d45756bb

jroweboy:
Updated the code to be more like what bunnei suggested. Theres still a little math bug in here that someone pointed out that I need to fix. Just got a lot to do right now and not enough time to do it.

neobrain:

2016-09-12T05:56:46Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...jroweboy:5b36fb7cfd8f16945fa81fa1000849ea01d2d9d5

neobrain:

2016-09-12T06:09:38Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...jroweboy:01ebf95de65926209b691c24d4b1a147ccadf97d

neobrain:

2016-09-12T16:46:08Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...jroweboy:204ac522821e3d82d67d18d3383623b439088f0b

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+#include <cmath>
+
+#include "common/assert.h"
+#include "common/framebuffer_layout.h"
+#include "video_core/video_core.h"
+
+namespace Layout {
+static FramebufferLayout DefaultFrameLayout(unsigned width, unsigned height) {
+
```
yuriks: Blank line seems superfluous.

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,312 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <cmath>
+
+#include "common/assert.h"
+#include "common/framebuffer_layout.h"
+#include "video_core/video_core.h"
+
+namespace Layout {
```
yuriks: Add a blank line separating namespace from function

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,312 @@
...
+static FramebufferLayout DefaultFrameLayout(unsigned width, unsigned height) {
+
+    ASSERT(width > 0);
+    ASSERT(height > 0);
```
yuriks: I think we've had situations before where these were called with 0 (during e.g. window initialization). You should probably just clamp them to 1 to prevent that.

yuriks:
There seems to be a lot of redundancy in the code that calculates the screen layouts. I suspect you could improve the quality of that code by a lot by identifying and factoring out common primitives (like, say, a function that calculates the largest rectangle with given AR that fits inside the given rectangle) and then reusing them in the various functions.

Jhno591:
If you are using the Input Core #1955 with this it playing mayhem with merging, needs looking into thanks.

JayFoxRox:
How is that yaourt bug related to this @Jhno591 ?
Didn't I tell you earlier to be careful with what you write here?
Also merge conflicts are not of interest here - they'll be resolved in the order of the merges..

Jhno591:
I should have put to @jroweboy  , But don't matter Fox cheers , hope it's resolved at some point at later stage.

wwylele:

SonofUgly:
Think I've got the Large frame layout working as intended. `emulation_aspect_ratio` doesn't seem exactly right, but [it works.](https://u.pomf.is/yslpva.webm)
https://github.com/SonofUgly/citra/commit/346e2bcbe731fdff69a1aee343d08a3f9e8f01dd

jroweboy:
@SonofUgly Hey thanks for your contribution. I'm actually reworking it right now as yuriks suggested but I'll leave your commit in there so you get some credit for your help 👍 (provided no one else has any objections to doing that)

bunnei:
This is a great feature that should have been merged six months ago -- @jroweboy please get this wrapped up!

jroweboy:
Fixed up the code to have the proper scaling in the large screen layout like @SonofUgly pointed out. Reworked the code to use a common "find max size rectangle" function like @yuriks suggested. Also as a small bonus, I added a default keybinding to swap the screens at any point by using Ctrl+Tab

SonofUgly:
Now in large screen, the bottom screen scales with the large one, and is only 1x if the large screen is 4x. So if you're on a screen <1080p the bottom window is going to be <native. [Pic.](http://i.imgur.com/M3QEwFT.png) IMO the small screen should always be 1x (maybe 2x on really high resolutions, but that can be left for a future PR).
And as the rendering resolution is determined only by the top screen resolution (which should be changed, but I figured that was out of scope), if you have the screens swapped and the small screen is <native, you end up with [this.](http://i.imgur.com/Nxkyov5.png)

jroweboy:
@SonofUgly that was my original intent for the layout. I wrote it for the games that only use the bottom screen for small things like playing the ocarina or what not. So it is intentional that the size can drop below native resolution for the bottom screen, but I don't know if thats the best strategy for this layout. I'm kinda expecting this to be a stepping stone layout though, with further improvements to layout options including customizing placement entirely, or splitting into multiple windows (see the other comments)

As for the resolutions issues... yeah that'll probably need a different pr :)

SonofUgly:
Well alright, I just don't think Citra should be displaying anything at less than native, and as the bottom screen is almost always 2D, as a secondary screen, it'd be best to keep it integer scaled.

Anyway, one slight bug that happens on all the layouts, one or both of the screens are always 1-2 pixels off in both x/y.
Easiest way to see it is by using the default layout and making the window as small as it'll go. The screens should be 400x240 and 320x240, but they're 399x239 and 318x239.

jroweboy:
@SonofUgly Thats a float to integer conversion issue. I will look into that. Any other reviews or comments? After fixing the float to int casting issues it should be good to unless anyone else has any more concerns.

jroweboy:
@SonofUgly Updated to fix the size issue. It should now properly fill the screen (like in this screenshot http://i.imgur.com/S4c5fZc.png)

SonofUgly:
LGTM

src/citra/default_ini.h
```diff
@@ -63,6 +63,16 @@ use_scaled_resolution =
 # 0 (default): Off, 1: On
 use_vsync =
 
+[Layout]
+# Layout for the screen inside the render window.
+# 0 (default): Default Top Bottom Screen, 1: Single Screen Only, 2: Large Screen Small Screen
```
JayFoxRox: nit: I don't think these names are very describtive.

src/common/framebuffer_layout.cpp
JayFoxRox: Already done :)

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include "video_core/video_core.h"
+
+namespace Layout {
+
+static const float TOP_SCREEN_ASPECT_RATIO =
+    static_cast<float>(VideoCore::kScreenTopHeight) / VideoCore::kScreenTopWidth;
+static const float BOT_SCREEN_ASPECT_RATIO =
+    static_cast<float>(VideoCore::kScreenBottomHeight) / VideoCore::kScreenBottomWidth;
+static const float BOT_TO_TOP_SCREEN_RATIO_DIFFERENCE =
+    BOT_SCREEN_ASPECT_RATIO - TOP_SCREEN_ASPECT_RATIO;
```
JayFoxRox: For what kind of black magic was this ratio math / `BOT_TO_TOP_SCREEN_RATIO_DIFFERENCE` meant for?
(It seems to be unused)

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,140 @@
...
+                                  static_cast<T>(std::round(scale * screen_aspect_ratio))};
+}
+
+FramebufferLayout DefaultFrameLayout(unsigned width, unsigned height, bool swapped) {
+    ASSERT(width > 0);
+    ASSERT(height > 0);
+
+    FramebufferLayout res{width, height, true, true, {}, {}};
+    // Default layout gives equal screen sizes to the top and bottom screen
+    MathUtil::Rectangle<unsigned> screen_window_area{0, 0, width, height / 2};
```
JayFoxRox: Shouldn't it match the placement on a real 3DS, possibly also accounting for the gap between displays?
Probably beyond scope of the PR and we can fix it later.

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,140 @@
...
+        bot_screen = bot_screen.TranslateX((top_screen.GetWidth() - bot_screen.GetWidth()) / 2);
+        if (swapped) {
+            bot_screen = bot_screen.TranslateY(height / 2 - bot_screen.GetHeight());
+        } else {
+            top_screen = top_screen.TranslateY(height / 2 - top_screen.GetHeight());
+        }
+    }
+    // Move the top screen to the bottom if we are swapped.
+    res.top_screen = swapped ? top_screen.TranslateY(height / 2) : top_screen;
+    res.bottom_screen = swapped ? bot_screen : bot_screen.TranslateY(height / 2);
```
JayFoxRox: nit: This code is really ugly. With 2 lines checking for swapped, directly after checking for swapped in those if checks.
It probably would be a lot more sane to work with references here and swap those once.
We can fix it in a follow-up though.

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,140 @@
...
+            top_screen.TranslateX((screen_window_area.GetWidth() - top_screen.GetWidth()) / 2);
+        bot_screen =
+            bot_screen.TranslateX((screen_window_area.GetWidth() - bot_screen.GetWidth()) / 2);
+    } else {
+        top_screen = top_screen.TranslateY((height - top_screen.GetHeight()) / 2);
+        bot_screen = bot_screen.TranslateY((height - bot_screen.GetHeight()) / 2);
+    }
+    res.top_screen = top_screen;
+    res.bottom_screen = bot_screen;
+    return res;
```
JayFoxRox: Where does this function actually pick the screen which should be displayed?
To me it looks like, like this returns both screens and the drawing decision is made elsewhere?!

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+FramebufferLayout LargeFrameLayout(unsigned width, unsigned height, bool swapped) {
+    ASSERT(width > 0);
+    ASSERT(height > 0);
+
+    FramebufferLayout res{width, height, true, true, {}, {}};
+    // Split the window into two parts. Give 4x width to the main screen and 1x width to the small
+    // To do that, find the total emulation box and maximize that based on window size
```
JayFoxRox: Those ratios should probably be variables so we can make it easier to change them later.
Can be done in a followup though.

src/common/framebuffer_layout.cpp
JayFoxRox: After checking `FramebufferLayout` I see where this happens.
I don't like how this is implemented. But it's nitpicking.
Gets the job done.

src/common/framebuffer_layout.cpp
```diff
@@ -0,0 +1,140 @@
...
+    res.bottom_screen = swapped ? bot_screen : bot_screen.TranslateY(height / 2);
+    return res;
+}
+
+FramebufferLayout SingleFrameLayout(unsigned width, unsigned height, bool swapped) {
+    ASSERT(width > 0);
+    ASSERT(height > 0);
+    // The drawing code needs at least somewhat valid values for both screens
+    // so just calculate them both even if the other isn't showing.
+    FramebufferLayout res{width, height, !swapped, swapped, {}, {}};
```
JayFoxRox: nit: As one might not know the order of members in FramebufferLayout it would be wise to add a comment that `swapped` is resposible to control output for small and bottom screen in this line.

src/common/framebuffer_layout.h
```diff
@@ -0,0 +1,44 @@
...
+#include "common/math_util.h"
+namespace Layout {
+/// Describes the layout of the window framebuffer (size and top/bottom screen positions)
+struct FramebufferLayout {
+    unsigned width;
+    unsigned height;
+    bool top_screen_enabled;
+    bool bottom_screen_enabled;
+    MathUtil::Rectangle<unsigned> top_screen;
+    MathUtil::Rectangle<unsigned> bottom_screen;
```
JayFoxRox: We should probably just have a `std::vector<{picture, rect}>` here to render as many displays as we'd like.
So single layout would return a vector with 1 element. Most other layout 2. But if we ever decide to go fancy we might render left and right eye framebuffers elsewhere (PictureTopLeft, PictureTopRight, PictureBottom).

However this can also be done in a followup.

src/common/framebuffer_layout.h
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Factory method for constructing a default FramebufferLayout
+ * @param width Window framebuffer width in pixels
+ * @param height Window framebuffer height in pixels
+ * @return Newly created FramebufferLayout object with default screen regions initialized
+ */
+FramebufferLayout DefaultFrameLayout(unsigned width, unsigned height, bool is_swapped);
+
+/**
+ * Factory method for constructing a FramebufferLayout with only the top screen
```
JayFoxRox: top or bottom screen

src/common/framebuffer_layout.h
```diff
@@ -0,0 +1,44 @@
...
+    bool bottom_screen_enabled;
+    MathUtil::Rectangle<unsigned> top_screen;
+    MathUtil::Rectangle<unsigned> bottom_screen;
+};
+
+/**
+ * Factory method for constructing a default FramebufferLayout
+ * @param width Window framebuffer width in pixels
+ * @param height Window framebuffer height in pixels
+ * @return Newly created FramebufferLayout object with default screen regions initialized
```
JayFoxRox: `is_swapped` not explained (other instances too)

src/common/framebuffer_layout.h
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Factory method for constructing a Frame with the a 4x size Top screen with a 1x size bottom
+ * screen on the right
+ * This is useful in particular because it matches well with a 1920x1080 resolution monitor
+ * @param width Window framebuffer width in pixels
+ * @param height Window framebuffer height in pixels
+ * @return Newly created FramebufferLayout object with default screen regions initialized
+ */
+FramebufferLayout LargeFrameLayout(unsigned width, unsigned height, bool is_swapped);
+}
```
JayFoxRox: missing newline at end of file?

src/common/math_util.h
```diff
@@ -38,6 +38,16 @@ struct Rectangle {
...
     }
+    Rectangle<T> TranslateX(const T x) const {
+        return Rectangle{left + x, top, right + x, bottom};
+    }
+    Rectangle<T> TranslateY(const T y) const {
+        return Rectangle{left, top + y, right, bottom + y};
+    }
+    Rectangle<T> Scale(const float s) const {
+        return Rectangle{left, top, static_cast<T>((right + left) * s),
+                         static_cast<T>((top + bottom) * s)};
```
JayFoxRox: Shouldn't this be `left+(right-left)*s` or something?

src/common/framebuffer_layout.cpp
jroweboy: No plans to add a gap at this time.

src/common/framebuffer_layout.cpp
jroweboy: The check in the if statement is also in another if statement, so i didn't merge them as this was just easier for me to reason about. In my mind, the task was "Compute size and the x location for each screen" and then at the end "if swapped move the bottom screen down else move the top screen down" Probably better to do it more like I did in large framelayouts though where I define a "large screen" and a "small screen" and find the sizes for each of those independantly and set top and bottom at the end. 

I agree that this is somewhat clunky though.

src/common/framebuffer_layout.cpp
jroweboy: If anything, I'd rather just lock the small screen to 1x and scale the large screen to whatever fits the rest of the resolution, but I'm not sure if thats the best way to go about this.

src/common/framebuffer_layout.cpp
JayFoxRox: okay

src/common/framebuffer_layout.cpp
JayFoxRox: okay

src/common/framebuffer_layout.cpp
JayFoxRox: okay, we'll do it in a followup if necessary

JayFoxRox:
Code reviewed and tested with OoT including swaping-screens while using stylus etc.
All remaining diff comments are nits, subjective decisions or something which can be fixed in follow ups.

As @bunnei already said this should have been merged 6 months ago.
The only remaining strong critic I have is that the rendering quality is very poor if the top-screen is display small. However, that should be considered a problem with the upscaled rendering and not this PR. As such we can also fix it in a followup.

LGTM

JohnGodgames:
@jroweboy can you add a switching between layouts (default, single screen, large screen) by a TAB key?

jroweboy:
@JohnGodgames https://github.com/citra-emu/citra/pull/1753#issuecomment-258600287

I think you can rebind this in the Settings.ini  as well

JohnGodgames:
@jroweboy I'm not asking for swapping top/bottom screens, but switching a layouts...

jroweboy:
Ah my bad. I personally think that is less useful and adding it doesn't seem to make much sense (to me). But if you write a good reason to have layout switching bound to a key binding, I'll consider adding it to a different pull request. (It should really only be about 5 lines of code)

I don't think i should be a part of this one as its been stalled for far too long.

JohnGodgames:
I'll be glad for it and there is a reason https://www.youtube.com/watch?v=CkD3D1S6XTU

JayFoxRox:
Even if - we can do it in a followup. The base feature works fine as is.
Please don't delay this PR further. We probably should have a feature-freeze label or something.

PR #1754, opened 2016-05-03
OpenGL: Don't copy const_color (Reverts #1745)

### LINKED ISSUE #1752: RasterizerOpenGL::SetShader now creates much more new shader than before
This issue appears since #1745 merged. I can see there is much more "Creating new shader" log, and memory are consumed faster. I am not sure if this is the same padding issue in #1621 (I tried a quick memset/memcpy patch but seems not help). Or maybe `const_color` can vary in a large range in high frequency, so we shouldn't create shader for each value?

Edit: Now I think adding `const_color` in #1745 is a mistake. `const_color` is supposed to be handled as a  shader uniform(see [here](https://github.com/citra-emu/citra/blob/master/src/video_core/renderer_opengl/gl_rasterizer.cpp#L1007-1013) and [here](https://github.com/citra-emu/citra/blob/master/src/video_core/renderer_opengl/gl_shader_gen.cpp#L536)), not as a parameter to create shader

JayFoxRox:
Should fix #1752 by reverting erroneous change from #1745 (Sorry :disappointed: ).

This keeps the loop but removes the bad line.
I've also made the comment more clear so this doesn't happen again.

~~Untested, will test soon.~~ Just tested playing OoT, started off at 304MB RAM usage, went up to 391MB RAM after playing for 40 minutes (I've played through almost the entire deku tree). Fixed?

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.h
+++ b/src/video_core/renderer_opengl/gl_rasterizer.h
@@ -49,7 +49,9 @@ struct PicaShaderConfig {
         res.alpha_test_func = regs.output_merger.alpha_test.enable ?
             regs.output_merger.alpha_test.func.Value() : Pica::Regs::CompareFunc::Always;
 
-        // Copy tev stages
+        // Copy relevant tev stages fields.
+        // We don't sync const_color here because of the high variance, it is a
+        // shader uniform instead.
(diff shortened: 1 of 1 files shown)


bunnei:
LGTM

PR #1755, opened 2016-05-03
Implementation of a language selector to configuration widget

uberhalit:
Added a language selector to configuration widget.
Current master requires a hex-edit in `\user\nand\data\00000000000000000000000000000000\sysdata\00010017\00000000\config` to change in-game language for multi-language games as english is hardcoded as default.

I used [ISO 639-2/T](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for 3-letter codes and native language names.

Due to the current implementation of `src/core/hle/service/cfg/cfg.cpp - Init()` setting language will be enforced only if there is no config file present as writing will be skipped completely if the file is found. I noted this with TODO.

language selector


src/citra_qt/configure_general.ui
```diff
@@ -101,6 +101,76 @@
...
+                    </property>
+                  </item>
+                  <item>
+                    <property name="text">
+                      <string notr="true">ENG (English)</string>
+                    </property>
+                  </item>
+                  <item>
+                    <property name="text">
+                      <string notr="true">FRA (français)</string>
```
Kloen: `Français`

src/citra_qt/configure_general.ui
```diff
@@ -101,6 +101,76 @@
...
+                    </property>
+                  </item>
+                  <item>
+                    <property name="text">
+                      <string notr="true">DEU (Deutsch)</string>
+                    </property>
+                  </item>
+                  <item>
+                    <property name="text">
+                      <string notr="true">ITA (italiano)</string>
```
Kloen: `Italiano`

src/citra_qt/configure_general.ui
```diff
@@ -101,6 +101,76 @@
...
+                    </property>
+                  </item>
+                  <item>
+                    <property name="text">
+                      <string notr="true">SPA (español)</string>
```
Kloen: `Español`

src/citra_qt/configure_general.ui
```diff
@@ -101,6 +101,76 @@
...
+                    </property>
+                  </item>
+                  <item>
+                    <property name="text">
+                      <string notr="true">NLD (Nederlands)</string>
+                    </property>
+                  </item>
+                  <item>
+                    <property name="text">
+                      <string notr="true">POR (português)</string>
```
Kloen: `Português`

src/citra_qt/configure_general.ui
uberhalit: The Library of Congress listed these with lowercase:
https://www.loc.gov/standards/iso639-2/php/French_list.php

src/citra_qt/configure_general.ui
uberhalit: And the Unicode CLDR Project listed them with lowercase too.
http://cldr.unicode.org/

vitorsilverio:
@uberhalit I am a Portuguese native speaker and Português is the language and português is a person that born in Portugal.

Kloen:
@vitorsilverio @uberhalit 

They are all valid in this context both uppercase or lowercase. At least for what I know. I'm spanish and it's fine Spanish or spanish, I don't mind at all. But I thought that despite being politically corrects. It makes more sense to be all on uper or all on lower since they are between ( ).

Subv:
If we implement SetConfigInfoBlk4, we should be able to change the value of the config block at runtime, however I don't know how games react to this value changing dynamically.

wwylele:
IMO language select should go to a new configuration tab, with user name and birthday together, because they are not part of emulation configuration, but a configuration of 3DS itself. Also, I think leave a TODO in code and let user delete config file (even without a message about that) is not a good idea. As Subv said, it is better to implement `SetConfigInfoBlk4` first, and just call it in the configuration window (a bonus is that we can also run 3DS's system settings ROM to configure language)

uberhalit:
@wwylele I'll do a new tab next week which will include all implemented 3DS configurations but I am not able to work on `WriteConfigInfo()` as my knowledge of 3DS is pretty much zero. This should be done in a seperate commit anyway.

mailwl:
just to mention, not all regions contains all languages. For example, 3ds from USA doesn't contains Russian

wwylele:
@uberhalit I will do the `WriteConfigInfo()` thing after #1756.

uberhalit:
@wwylele great, I'll update this commit to a full blown emulation configuration afterwards.

wwylele:
@uberhalit would you mind if I take over this? I find it difficult to test and review if I implement `WriteConfigInfo` but leave the frontend unimplemented.

uberhalit:
@wwylele yes you can, I am still too busy with university

PR #1756, opened 2016-05-03
Config block: clean up

wwylele:
As I said in #1433, there shouldn't be two username blocks, so remove one of them, and clean up some format problem.
But I am still afraid of making mistake. @Romsstar or anyone else please help test to ensure this doesn't cause any regression (first remove the config file `nand/data/0000000000000000/sysdata/00010017/00000000/config` and then run games).

### DIFF
--- a/src/core/hle/service/cfg/cfg.cpp
+++ b/src/core/hle/service/cfg/cfg.cpp
@@ -47,6 +47,12 @@ struct UsernameBlock {
 };
 static_assert(sizeof(UsernameBlock) == 0x1C, "UsernameBlock must be exactly 0x1C bytes");
 
+struct BirthdayBlock {
+    u8 month; ///< The month of the birthday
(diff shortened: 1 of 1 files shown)


Subv:
LGTM

purpasmart96:
Looks good, although seeing the structs not in a header file is weird, I believe @yuriks was responsible for doing that, so although not really a complaint of this PR, but some that crap should be in the header file.

lioncash:
Structs should ideally not be in a header if they aren't intended to be used anywhere else, that needlessly puts stuff in the exposed namespace.

wwylele:
Tested with some system app and asked my friend to test _Digimon World Re: Digitize Decode_. All of them read the name block successfully and there is no regression. I think this is safe to go.

bunnei:
LGTM

PR #1757, opened 2016-05-04
Pica: Rename VertexLoaded breakpoint to VertexShaderInvocation

JayFoxRox:
Another small change to prepare for geometry shaders which will introduce "Geometry shader invocation".

1. Renames the VertexLoaded-events to VertexShaderInvocation ("Vertex shader invocation")
2. Moves the debug events closer to the actual ::Run() of the shader

(The next changes will be larger, I'm aware that these tiny PRs are annoying)


Subv:
>I'm aware that these tiny PRs are annoying

They aren't annoying.
LGTM

bunnei:
LGTM

PR #1668, opened 2016-04-14
Update the code of APT service(part2)

JamePeng:
The part2 update some stub codes 
Otherwise, I'd like to research something！

PR #1761, opened 2016-05-05
HLE/Applets: Use the correct size for the framebuffer SharedMemory

Subv:
This is preliminary work for a fix we need in #1689

### DIFF
--- a/src/core/hle/applets/swkbd.cpp
+++ b/src/core/hle/applets/swkbd.cpp
@@ -24,13 +24,6 @@
 namespace HLE {
 namespace Applets {
 
-SoftwareKeyboard::SoftwareKeyboard(Service::APT::AppletId id) : Applet(id), started(false) {
-    // Create the SharedMemory that will hold the framebuffer data
-    // TODO(Subv): What size should we use here?
-    using Kernel::MemoryPermission;
-    framebuffer_memory = Kernel::SharedMemory::Create(0x1000, MemoryPermission::ReadWrite, MemoryPermission::ReadWrite, "SoftwareKeyboard Memory");
-}
-
 ResultCode SoftwareKeyboard::ReceiveParameter(Service::APT::MessageParameter const& parameter) {
     if (parameter.signal != static_cast<u32>(Service::APT::SignalType::LibAppJustStarted)) {
         LOG_ERROR(Service_APT, "unsupported signal %u", parameter.signal);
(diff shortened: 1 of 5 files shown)


src/core/hle/applets/swkbd.h
```diff
@@ -72,8 +71,8 @@ class SoftwareKeyboard final : public Applet {
      */
     void Finalize();
 
-    /// TODO(Subv): Find out what this is actually used for.
-    /// It is believed that the application stores the current screen image here.
+    /// This SharedMemory will be created when we receive the LibAppJustStarted message.
+    /// It holds the framebuffer info retrieved by the application with GSPGPU::ImportDisplayCaptureInfo
     Kernel::SharedPtr<Kernel::SharedMemory> framebuffer_memory;
```
JayFoxRox: Is there any good way / value to initialize this?
It used to be initialized by the constructor of derived classes but this isn't the case anymore.
So if someone uses this before ReceiveParameter with LibAppJustStarted this will be bad.

src/core/hle/applets/swkbd.cpp
```diff
@@ -39,8 +32,19 @@ ResultCode SoftwareKeyboard::ReceiveParameter(Service::APT::MessageParameter con
...
+    // Create the SharedMemory that will hold the framebuffer data
+    Service::APT::CaptureBufferInfo capture_info;
+    ASSERT(sizeof(capture_info) == parameter.buffer_size);
+
+    memcpy(&capture_info, parameter.data, sizeof(capture_info));
+
+    using Kernel::MemoryPermission;
+    framebuffer_memory = Kernel::SharedMemory::Create(capture_info.size, MemoryPermission::ReadWrite,
+                                                      MemoryPermission::ReadWrite, "SoftwareKeyboard Memory");
+
```
JayFoxRox: I'm not how faithful this recreates 3DS code, but can't we forward this message to the base class instead?
(e.g. only handle ReceiveParameter here if it's actually specific to this applet, otherwise forward it to the Applet class?)

src/core/hle/service/apt/apt.h
```diff
@@ -73,6 +74,18 @@ enum class StartupArgumentType : u32 {
...
+struct CaptureBufferInfo {
+    u32_le size;
+    u8 is_3d;
+    INSERT_PADDING_BYTES(0x3); // Padding for alignment
+    u32_le top_screen_left_offset;
+    u32_le top_screen_right_offset;
+    u32_le top_screen_format;
+    u32_le bottom_screen_left_offset;
+    u32_le bottom_screen_right_offset;
+    u32_le bottom_screen_format;
```
JayFoxRox: We don't respect those kind of values yet, right? Should we? If so please note this somewhere as a todo.
Also why does the bottom screen have `left_offset` and `right_offset`? I thought only the upper screen had 3D.

src/core/hle/applets/swkbd.h
Subv: SharedPtr will automatically initialize this to nullptr, I think.

src/core/hle/applets/swkbd.cpp
Subv: Different applets may do different things in this message. I don't think applets are even supposed to send the CaptureBufferInfo like this (Considering there's APT::Send/ReceiveCaptureBufferInfo), but swkbd and MiiSelector both do it like this.
I didn't want to abstract this into the base Applet class just to find out later that not all applets do it like that.

src/core/hle/service/apt/apt.h
Subv: What do you mean by respect? See http://3dbrew.org/wiki/NS_and_APT_Services#CaptureBufferInfo

JamePeng: Why the struct element name different from 3dbrew? 
It also come into conflict with my PR #1668

Subv: We use bottom instead of sub in other parts of the code too, it's just to keep consistency.
As for your PR, IMO it is already quite big, letting this change go through here will make your PR smaller and hence easier to review.

bunnei:
LGTM

src/core/hle/service/apt/apt.h
JamePeng: ok

wwylele:
LGTM

PR #1762, opened 2016-05-06
hle: Get rid of direct global access to g_reschedule

bunnei:
Because I'm impatient.

Supersedes #1561


linkmauve:
supersedes* :D

PR #1764, opened 2016-05-06
fix loader: return proper result

wwylele:
Fixes a small bug: when opening an encrypted ROM citra says "Unknown error" instead of information about decryption.
Was introduced in my game list icon PR. Sorry!


bunnei:
LGTM

PR #1765, opened 2016-05-06
Debugger: Pica surface viewer

JayFoxRox:
This replaces the "Texture Viewer" from the "Pica Command List" debug widget and the "Pica Framebuffer" debug widget with a more generic "Pica Surface Viewer" which is more robust.

Here is some eye-candy for you (widget opened 3 times):

Surface viewer providing display of color-/depthbuffer and texture 0

*(The screenshot is outdated, "Virtual Address" is "Physical Address" now and hex-output is uppercase)*

**Pros**:

* Doesn't crash with invalid address
  Widget only crashes if texture larger than mapped memory which is probably rare (and we can fix this with something like  #1692  probably)

* Can show textures, color-, depth- and stencil-Buffers (in all formats)

* Pixel picker to show exact values for specified pixel

* PNG and memory dump

* Can open as many widgets as you want from the menu
  = show colorbuffer and depthbuffer using different widgets at the same time.

* Scrollbars so you can open large textures on small displays

**Cons (with discussion)**:

* Widget position not stored / restored
  This sucks, but I couldn't make it work. Maybe someone more knowledgeable can implement it later?

* Doesn't preview textures in command list anymore
  I think this doesn't matter because whenever you select a texture from the command list in master, it shows the current texture, not the actual texture which was used at that point in time.
We should have a seperate issue so we add a feature later where you can view textures at command submission or execution time.

* No rotate or zoom
  At least scale to fit would be nice, however, due to scrollbars this is better than what's available in master already.

* Picked pixel not visualized
  Would be nice-to-have as it would be easier to work on pixel data

* No easy way to keep widget on one specified texture
  There should be a button to clone the current window (maybe to "Custom mode") and maybe there should be a switch to disable updates to the image

* No additional display modes (= alpha or depth as grayscale)
  Should be implemented later. This feature doesn't exist in master either.

* No option for mipmaps and cubemaps
  We don't have it in master either, but we should probably have these options.

---

This PR is mostly based on old code which is already in master. You can probably diff the new surface viewer against the old framebuffer debugger.
Nibble printing ("Raw: ..." next to the pixel pickers) could still have some errors.


src/citra_qt/debugger/graphics_surface.cpp
```diff
@@ -0,0 +1,746 @@
...
+                                                    QString("%1;;%2").arg(png_filter, bin_filter), &selectedFilter);
+
+
+    if (filename.isEmpty()) {
+        // If the user canceled the dialog, don't save anything.
+        return;
+    }
+
+    if (selectedFilter == png_filter) {
+
```
lioncash: unnecessary opening newline (ditto with the endings as well)

src/citra_qt/debugger/graphics_surface.cpp
```diff
@@ -0,0 +1,746 @@
...
+        QFile file(filename);
+        file.open(QIODevice::WriteOnly);
+        int size = surface_width * surface_height * NibblesPerPixel(surface_format) / 2;
+        QByteArray data(reinterpret_cast<const char*>(buffer), size);
+        file.write(data);
+
+    } else {
+        UNREACHABLE_MSG("Unhandled filter selected");
+    }
+
```
lioncash: unnecessary newline.

src/citra_qt/debugger/graphics_surface.cpp
```diff
@@ -0,0 +1,746 @@
...
+        case Format::Unknown:
+            return QString("Unknown format");
+        default:
+            return QString("Unhandled format");
+        }
+        return QString("");
+    };
+
+    QString nibbles = "";
+    for(unsigned i = 0; i < nibbles_per_pixel; i++) {
```
lioncash: space after the `for`

src/citra_qt/debugger/graphics_surface.cpp
```diff
@@ -0,0 +1,746 @@
...
+        }
+        u8 byte = pixel[nibble_index / 2];
+        u8 nibble = (byte >> ((nibble_index % 2) ? 0 : 4)) & 0xF;
+        nibbles.append(QString::number(nibble, 16).toUpper());
+    }
+
+    surface_info_label->setText(QString("Raw: 0x%3\n(%4)").arg(nibbles).arg(GetText(surface_format, pixel)));
+    surface_info_label->setAlignment(Qt::AlignLeft | Qt::AlignVCenter);
+
+    return;
```
lioncash: This isn't necessary in a void function.

src/citra_qt/debugger/graphics_surface.cpp
```diff
@@ -0,0 +1,746 @@
...
+    // Only do something while the left mouse button is held down
+    if (!(event->buttons() & Qt::LeftButton))
+        return;
+
+    if (pixmap() == nullptr)
+        return;
+
+    if (surface_widget)
+        surface_widget->Pick(event->x() * pixmap()->width() / width(),
+                                 event->y() * pixmap()->height() / height());
```
lioncash: wonky indentation

src/citra_qt/debugger/graphics_surface.cpp
```diff
@@ -0,0 +1,746 @@
...
+
+#include "video_core/pica.h"
+#include "video_core/pica_state.h"
+#include "video_core/utils.h"
+
+SurfacePicture::SurfacePicture(QWidget* parent, GraphicsSurfaceWidget* surface_widget_) : QLabel(parent), surface_widget(surface_widget_) {}
+SurfacePicture::~SurfacePicture() {}
+
+void SurfacePicture::mousePressEvent(QMouseEvent* event) {
+
```
lioncash: unnecessary newline, etc, etc. There's other occurrences as well.

zminhquanz:
like similarly PPSSPP , add step frame and step prim , step into

bunnei:
Code looks fine to me. Call for any additional reviews on this one... If I don't hear anything, I plan on merging later this week.

JayFoxRox:
This could theoretically use the new memory interface (#1692, see first post) to fix a crash with large texture sizes.

However, I'd rather have this merged first because I'm busy with uni currently.
Also I'm working on other projects lately which means I'm not sure when and if I'd do these changes.
(I'm sure we can delegate this to a random person joining IRC too)

PR #1766, opened 2016-05-07
Kernel/Threading: Warn when a thread can be scheduled in the Syscore (Core 1)

Subv:
We do not currently implement any cores other than the AppCore (Core 0).


bunnei:
LGTM

PR #1771, opened 2016-05-09
dyncom: Reset the context into user mode correctly

lioncash:
The other mode was system mode.


Subv:
LGTM

JayFoxRox:
(Not sure if we should reset into USER or SYSTEM mode, eitherway, the current code/comment was broken and it probably won't even make a difference)

LGTM

bunnei:
LGTM

PR #1773, opened 2016-05-10
gl_rasterizer: Fix compilation for debug builds

lioncash:


src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -63,7 +63,7 @@ union PicaShaderConfig {
         // We don't sync const_color here because of the high variance, it is a
         // shader uniform instead.
         const auto& tev_stages = regs.GetTevStages();
-        DEBUG_ASSERT(res.tev_stages.size() == tev_stages.size());
+        DEBUG_ASSERT(res.state.tev_stages.size() == tev_stages.size());
```
JayFoxRox: `state = res.state` so you can probably just use `DEBUG_ASSERT(state.tev_stages.size() == tev_stages.size());`

Aside from that LGTM and sorry (was my fault / PR which broke it)

PR #1774, opened 2016-05-10
gdbstub: Silence missing prototype warnings

lioncash:


JayFoxRox:
Looks like this still compiles and those are only used in that file:

LGTM

bunnei:
LGTM

PR #1778, opened 2016-05-11
Reimplement circle pad input & add circle pad modifier

wwylele:
#### Reimplement circle pad
The way we used to handle circle pad input was:
Keyboard input ->[EmuWindow]-> circle pad direction -> [HID] -> make up circle pad position.
Now change it to:
Keyboard input ->[EmuWindow]-> make up circle pad position -> [HID] -> calculate circle pad direction.
(In the future we can also add : controller joystick -> [EmuWindow]-> real circle pad position -> ...)
The position-to-direction algorithm is confirmed by this [hw test](https://github.com/wwylele/ctrhwtest/tree/master/circle_pad). It has an error of +/- 1, but it is hard to test the edge cases on hardware, and I think it is enough for emulation.
Illustrate the circle pad direction threshold:
circle_pad

#### Add circle pad modifier
The word "modifier" is grabbed from Dolphin. (sorry for my poor English. Please help improve if I use it wrongly) When pressing the modifier key (default to D) and circle pad direction key simultaneously, the made up circle pad radius is the max radius multiplied by a scale (default to 0.5) to emulate player gently pushing the circle pad. This is useful, for example, for player to sneak in Pokemon OR/AS.

### DIFF
--- a/src/citra_qt/config.cpp
+++ b/src/citra_qt/config.cpp
@@ -27,7 +27,8 @@ static const std::array<QVariant, Settings::NativeInput::NUM_INPUTS> defaults = 
     Qt::Key_M, Qt::Key_N, Qt::Key_B,
     Qt::Key_T, Qt::Key_G, Qt::Key_F, Qt::Key_H,
     Qt::Key_Up, Qt::Key_Down, Qt::Key_Left, Qt::Key_Right,
-    Qt::Key_I, Qt::Key_K, Qt::Key_J, Qt::Key_L
+    Qt::Key_I, Qt::Key_K, Qt::Key_J, Qt::Key_L,
+    Qt::Key_D
 };
 
 void Config::ReadValues() {
@@ -36,6 +37,7 @@ void Config::ReadValues() {
         Settings::values.input_mappings[Settings::NativeInput::All[i]] =
             qt_config->value(QString::fromStdString(Settings::NativeInput::Mapping[i]), defaults[i]).toInt();
     }
+    Settings::values.circle_pad_modifier_scale = qt_config->value("circle_pad_modifier_scale", 0.5).toFloat();
     qt_config->endGroup();
 
     qt_config->beginGroup("Core");
@@ -125,6 +127,7 @@ void Config::SaveValues() {
         qt_config->setValue(QString::fromStdString(Settings::NativeInput::Mapping[i]),
             Settings::values.input_mappings[Settings::NativeInput::All[i]]);
     }
+    qt_config->setValue("circle_pad_modifier_scale", (double)Settings::values.circle_pad_modifier_scale);
     qt_config->endGroup();
 
     qt_config->beginGroup("Core");
--- a/src/common/emu_window.h
+++ b/src/common/emu_window.h
@@ -100,13 +100,45 @@ public:
     void TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y);
 
     /**
-     * Gets the current pad state (which buttons are pressed and the circle pad direction).
+     * Gets the current pad state (which buttons are pressed).
      * @note This should be called by the core emu thread to get a state set by the window thread.
+     * @note The circle pad fields of the returned state are cleared.
+     *         Circle pad should be handled from GetCirclePadState().
      * @todo Fix this function to be thread-safe.
      * @return PadState object indicating the current pad state
      */
     Service::HID::PadState GetPadState() const {
-        return pad_state;
+        auto cleared_pad_state = pad_state;
+        cleared_pad_state.circle_right.Assign(0);
+        cleared_pad_state.circle_left.Assign(0);
+        cleared_pad_state.circle_up.Assign(0);
+        cleared_pad_state.circle_down.Assign(0);
+        return cleared_pad_state;
+    }
+
+    /**
+     * Gets the current cirle pad state.
+     * @note This should be called by the core emu thread to get a state set by the window thread.
+     * @todo Fix this function to be thread-safe.
+     * @return std::tuple of (x, y), where `x` and `y` are the circle pad coordinates
+     */
+    std::tuple<s16, s16> GetCirclePadState() const {
+        const int MAX_CIRCLEPAD_POS = 0x9C; // Max radius for a circle pad position
+        const float SQRT_HALF = 0.707106781;
+        int x = 0, y = 0;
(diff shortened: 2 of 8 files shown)


src/core/hle/service/hid/hid.h
```diff
@@ -215,6 +215,8 @@ const PadState PAD_CIRCLE_LEFT  = {{1u << 29}};
 const PadState PAD_CIRCLE_UP    = {{1u << 30}};
 const PadState PAD_CIRCLE_DOWN  = {{1u << 31}};
 
+const PadState PAD_C_MODIFIER   = {{1u << 23}}; // An intermediate value only used by citra
```
JayFoxRox: We shouldn't do this. Circle pad modifier is a frontend issue. It should not touch the HLE at all.
Instead we should abstract it:
- Receive host input and store it somewhere (not in HLE but in generic frontend / input code)
- Translate the stored input to 3DS buttons / input

src/common/emu_window.h
```diff
@@ -100,13 +100,49 @@ class EmuWindow
...
+        if (pad_state.circle_left.ToBool())
+            x -= MAX_CIRCLEPAD_POS;
+        if (pad_state.circle_up.ToBool())
+            y += MAX_CIRCLEPAD_POS;
+        if (pad_state.circle_down.ToBool())
+            y -= MAX_CIRCLEPAD_POS;
+        if (x != 0)
+            y *= SQRT_HALF;
+        if (y != 0)
+            x *= SQRT_HALF;
```
JayFoxRox: I'd personally prefer doing float math here and keeping everything in a unit circle ([-1,+1] for {x,y}), then normalize the result aftwards.
After that calculate the actual circle pad radius in another step.

That will also make it easier to integrate joysticks probably

wwylele:
From #citra-dev
> [wwylele] and that "PadState PAD_C_MODIFIER" <- I now feel stupid for myself... trying removing it
> [JayFoxRox] wwylele: ideally we should do another PR before yours which cleans up that input mess
> [JayFoxRox] or we implement yours without modifier, then fixup the input subsystem and then add the modifier key
> [wwylele] JayFoxRox: so only leave the first commit in that PR, then fixup input subsystem? Not a bad idea.
> [JayFoxRox] idk, I think I'd hold back and even reimplement the first commit
> [JayFoxRox] if we refactor the input subsystem we can have it more sophisticated
> [JayFoxRox] we can have float circle_pad_x; float circle_pad_y; [-1, +1]
> [JayFoxRox] and all your current commit would have to do is to make sure it's actually radius < 1.0, then multiply by 0x9C or w/e the radius is
> [wwylele] JayFoxRox: OK so I suspend that PR and try refactoring the input system
> [JayFoxRox] basicly we need some HostInput { bool a; bool b; bool x; bool y; float circle_x; float circle_y; bool circle_modifier; } ... which would then run through something like ApplyHostInput(HostInput& input) { // set 3DS input here from `input` } 
> [JayFoxRox] wwylele: if you are fine with that that'd be awesome :)
> [wwylele] I am totally fine. actually I felt that when doing that PR. before I thought it is a easy fix but found the messy of input system
> [JayFoxRox] cool

So suspend this PR and I will work on the input system first.

src/citra_qt/config.cpp
```diff
@@ -125,6 +127,7 @@ void Config::SaveValues() {
         qt_config->setValue(QString::fromStdString(Settings::NativeInput::Mapping[i]),
             Settings::values.input_mappings[Settings::NativeInput::All[i]]);
     }
+    qt_config->setValue("circle_pad_modifier_scale", (double)Settings::values.circle_pad_modifier_scale);
```
lioncash: static_cast

src/core/hle/service/hid/hid.cpp
```diff
@@ -45,32 +43,41 @@ const std::array<Service::HID::PadState, Settings::NativeInput::NUM_INPUTS> pad_
...
 
+    PadState state = VideoCore::g_emu_window->GetPadState();
+
+    // Get current circle pad positon and update circle pad direction
+    const float TAN30 = 0.577350269, TAN60 = 1 / TAN30; // 30 degree and 60 degree are angular thresholds for directions
+    const int CIRCLE_PAD_THRESHOLD_SQUARE = 40 * 40; // a circle pad radius greater than 40 will trigger circle pad direction
+    s16 circle_pad_x, circle_pad_y;
+    std::tie(circle_pad_x, circle_pad_y) = VideoCore::g_emu_window->GetCirclePadState();
+    if (circle_pad_x * circle_pad_x + circle_pad_y * circle_pad_y > CIRCLE_PAD_THRESHOLD_SQUARE) {
+        float tan = abs((float)circle_pad_y / circle_pad_x);
```
lioncash: static_cast

src/core/hle/service/hid/hid.cpp
```diff
@@ -45,32 +43,41 @@ const std::array<Service::HID::PadState, Settings::NativeInput::NUM_INPUTS> pad_
...
 
     if (mem == nullptr) {
         LOG_DEBUG(Service_HID, "Cannot update HID prior to mapping shared memory!");
         return;
     }
 
+    PadState state = VideoCore::g_emu_window->GetPadState();
+
+    // Get current circle pad positon and update circle pad direction
+    const float TAN30 = 0.577350269, TAN60 = 1 / TAN30; // 30 degree and 60 degree are angular thresholds for directions
```
lioncash:
```cpp
constexpr float TAN30 = 0.577350269;
constexpr float TAN60 = 1 / TAN30;
```

src/core/hle/service/hid/hid.cpp
```diff
@@ -45,32 +43,41 @@ const std::array<Service::HID::PadState, Settings::NativeInput::NUM_INPUTS> pad_
...
 void Update() {
     SharedMem* mem = reinterpret_cast<SharedMem*>(shared_mem->GetPointer());
-    const PadState state = VideoCore::g_emu_window->GetPadState();
 
     if (mem == nullptr) {
         LOG_DEBUG(Service_HID, "Cannot update HID prior to mapping shared memory!");
         return;
     }
 
+    PadState state = VideoCore::g_emu_window->GetPadState();
```
lioncash: All of this code should likely be refactored into its own function (or several) so it's not putting specifics inside an otherwise generic function.

wwylele:
Superseded by #1789

PR #1780, opened 2016-05-11
Turn ShaderSetup into a class

JayFoxRox:
`Setup()` is an objectless function in master and it means that one would have to set up VS or GS before each run which is not suitable due to them being run in alternating pattern (VS is ran a couple of times to generate GS input, then GS runs. The process repeats for each group of vertices until enough GS input data is available again).

Rather than keeping the `Setup()` function in `Shader::` I moved it to a `class ShaderSetup::` (which maintains per shader type state which is not already in the Pica regs) so we can `Setup()` for VS or GS seperately.

I've also decided to rename the `Shutdown()` function to avoid ambiguity and I also kept it out of `ShaderSetup`.
This is more useful because VS and GS should be able to share the same shader cache.
If we would have called `Shutdown()` on either GS or VS it would have affected the other one too.
Also calling a `static ShaderSetup::Shutdown` function feels weird as it's not clear what will happen.

---

This doesn't add anything new. It's mostly cleanup for addition of GS which will happen in future PRs.

---

The code quality in this one is probably bad. I wasn't sure wether to use a `class` or a `struct` for example.
I'd like to remind you that this is part of a larger set of changes for GS.  So please don't ask me to refactor everything.

### DIFF
--- a/src/video_core/shader/shader.h
+++ b/src/video_core/shader/shader.h
@@ -83,23 +83,6 @@ struct OutputVertex {
 static_assert(std::is_pod<OutputVertex>::value, "Structure is not POD");
 static_assert(sizeof(OutputVertex) == 32 * sizeof(float), "OutputVertex has invalid size");
 
-/// Vertex shader memory
-struct ShaderSetup {
-    struct {
-        // The float uniforms are accessed by the shader JIT using SSE instructions, and are
-        // therefore required to be 16-byte aligned.
-        alignas(16) Math::Vec4<float24> f[96];
-
-        std::array<bool, 16> b;
-        std::array<Math::Vec4<u8>, 4> i;
-    } uniforms;
-
-    Math::Vec4<float24> default_attributes[16];
-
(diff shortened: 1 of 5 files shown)


src/video_core/shader/shader.h
```diff
@@ -342,33 +325,54 @@ struct UnitState {
...
- * Performs any shader unit setup that only needs to happen once per shader (as opposed to once per
- * vertex, which would happen within the `Run` function).
- */
-void Setup();
+/// Clears the shader cache
+void ClearCache();
 
-/// Performs any cleanup when the emulator is shutdown
-void Shutdown();
+class ShaderSetup {
```
lioncash: This should likely be a struct instead, considering you have everything exposed, but nothing hidden. Generally you use a class if there's invariants governing the data you're encapsulating.

If there actually are invariants, then the respective members they apply to should be hidden and have relevant API additions that ensure they hold (but I'm assuming that isn't the case here).

bunnei:
LGTM

PR #1782, opened 2016-05-12
DSP: dsp::DSP semaphore indicates completion of audio frame config

MerryMage:
How synchronisation between the emulated application and the DSP occurs:

1. DSP firmware signals audio-pipe interrupt. This indicates it's ready to receive a frame.
2. Application writes to shared memory region.
3. Application signals the semaphore.
4. DSP firmware does processing.
5. goto 1

On actual hardware Step 4 may take longer than ~5 ms which may result in skipped frames; this is not implemented here.


MerryMage:
Closing because of logic error.

Subv:
Side note, why is it called a semaphore if it's an event?

PR #1783, opened 2016-05-12
Cleanup ShaderSetup (Part 1)

JayFoxRox:
Another part of my GS branch.

ShaderSetup still stores the fixed per vertex attributes even if that is not directly tied to any specific shader type (VS / GS). Commit moves this into the generic Pica state.

UnitState (which is a helper struct for shader execution) contains state information which is only necessary for the interpreter. Commit moves this into respective functions as local variables.
(One could argue that it's part of the shader unit state, but it's not easily observable from the outside and JIT / Interpreter don't use the same variables, so this was just a memory hog and confusing)

The program_counter is initialized to the VS entry point now which might appear weird because it came from `config` before, however, we always use the VS as config source so this shouldn't break anything.
A later commit will change this back but it will have to change a lot of function signatures so I kept it seperated.

(ftr: There might not be a part 2, this is just broken down from a larger commit which I'll probably split into 2 more commits after this. However I might call the PR differently next time.)

### DIFF
--- a/src/video_core/pica_state.h
+++ b/src/video_core/pica_state.h
@@ -25,6 +25,8 @@ struct State {
     Shader::ShaderSetup vs;
     Shader::ShaderSetup gs;
 
+    std::array<Math::Vec4<float24>, 16> vs_default_attributes;
+
     struct {
         union LutEntry {
             // Used for raw access
--- a/src/video_core/shader/shader_interpreter.cpp
+++ b/src/video_core/shader/shader_interpreter.cpp
@@ -29,8 +29,24 @@ namespace Pica {
 
 namespace Shader {
 
+constexpr u32 INVALID_ADDRESS = 0xFFFFFFFF;
+
+struct CallStackElement {
+    u32 final_address;  // Address upon which we jump to return_address
+    u32 return_address; // Where to jump when leaving scope
+    u8 repeat_counter;  // How often to repeat until this call stack element is removed
+    u8 loop_increment;  // Which value to add to the loop counter after an iteration
+                        // TODO: Should this be a signed value? Does it even matter?
+    u32 loop_address;   // The address where we'll return to after each loop iteration
+};
+
 template<bool Debug>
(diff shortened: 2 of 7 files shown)


src/video_core/pica_state.h
```diff
@@ -25,6 +25,8 @@ struct State {
     Shader::ShaderSetup vs;
     Shader::ShaderSetup gs;
 
+    Math::Vec4<float24> vs_default_attributes[16];
```
lioncash:
```cpp
std::array<Math::Vec4<float24>, 16> vs_default_attributes;
```

src/video_core/shader/shader_interpreter.cpp
```diff
@@ -29,8 +29,27 @@ namespace Pica {
 
 namespace Shader {
 
+enum {
+        INVALID_ADDRESS = 0xFFFFFFFF
```
lioncash: overindent by about 4 spaces.

This could also likely just be a `constexpr u32` or at least be specified as

```cpp
enum : u32 {
    INVALID_ADDRESS = 0xFFFFFFFF
};
```

src/video_core/shader/shader_interpreter.cpp
```diff
@@ -29,8 +29,27 @@ namespace Pica {
...
+    u32 return_address; // Where to jump when leaving scope
+    u8 repeat_counter;  // How often to repeat until this call stack element is removed
+    u8 loop_increment;  // Which value to add to the loop counter after an iteration
+                        // TODO: Should this be a signed value? Does it even matter?
+    u32 loop_address;   // The address where we'll return to after each loop iteration
+};
+
 template<bool Debug>
 void RunInterpreter(UnitState<Debug>& state) {
+
```
lioncash: Unnecessary newline.

lioncash:
LGTM otherwise

bunnei:
LGTM

PR #1785, opened 2016-05-12
Microprofile: DPI-aware drawing

MerryMage:
Not sure if this is the best way to do this.

### DIFF
--- a/src/citra_qt/debugger/profiler.cpp
+++ b/src/citra_qt/debugger/profiler.cpp
@@ -151,6 +151,8 @@ private:
     /// This timer is used to redraw the widget's contents continuously. To save resources, it only
     /// runs while the widget is visible.
     QTimer update_timer;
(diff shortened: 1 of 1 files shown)


bunnei:
LGTM

src/citra_qt/debugger/profiler.cpp
```diff
@@ -220,11 +222,17 @@ MicroProfileWidget::MicroProfileWidget(QWidget* parent) : QWidget(parent) {
     MicroProfileInitUI();
 
     connect(&update_timer, SIGNAL(timeout()), SLOT(update()));
+
+    QPainter painter(this);
+    x_scale = qreal(painter.device()->physicalDpiX()) / qreal(painter.device()->logicalDpiX());
+    y_scale = qreal(painter.device()->physicalDpiY()) / qreal(painter.device()->logicalDpiY());
```
yuriks: Why is this being done here instead of paintEvent? Wondering if there's any reason for it.

Issue #1051, opened 2015-08-19
Race condition in Vertex Shader JIT


Subv:
There's a race condition between the UI thread enabling the VS JIT and the Shader::Setup and Shader::Run functions.

Picture this:
JIT is disabled
Game issues draw command and Shader::Setup is called, it doesn't do anything because the jit is disabled.
UI enables the shader JIT
Shader::Run now sees the JIT is enabled and tries to use jit_shader (a null pointer).
Crash

Subv:
Just a reminder that this still happens.

yuriks:
It sure does.

PR #1787, opened 2016-05-13
Refactor JIT

JayFoxRox:
*aka Cleanup ShaderSetup (Part 2)*

The shader JIT in `master` reads from a register struct (part of UnitState) and the uniform struct (part of ShaderSetup) with functions providing an offset into the register / uniform structs to access a certain member.

This commit instead gives the JIT full access to the UnitState and ShaderSetup so it's easier to see where the data is read from / written to (+ it's more extensible as we are running low on usable registers). Commit also introduces functions to get the offset into the uniform struct where calculations were inlined before.

I've also adapted the new function signature of the JIT for the interpreter.

(Also part of my GS implementation, as the ShaderSetup/Uniforms was hardcoded to VS)

---

You should enable jit before running this, I was just hit by #1051 while switching during gameplay (which I will fix in a later PR, also as part of my GS implementation)

### DIFF
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -84,9 +84,9 @@ OutputVertex ShaderSetup::Run(UnitState<false>& state, const InputVertex& input,
     if (VideoCore::g_shader_jit_enabled)
         jit_shader->Run(setup, state, config.main_offset);
     else
-        RunInterpreter(state);
+        RunInterpreter(setup, state, config.main_offset);
 #else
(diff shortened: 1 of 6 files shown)


JayFoxRox:
This seems to crash for me now if jit is enabled, I'll try to fix it (it worked fine earlier I think :confused:)

Subv:
LGTM

lioncash:
LGTM

linkmauve:
LGTM.

PR #1788, opened 2016-05-13
gitmodules: citra-emu/soundtouch → citra-emu/ext-soundtouch

MerryMage:
Repo was renamed.


JayFoxRox:
Note: Run `git submodule sync` after pulling this revision!

PR #1792, opened 2016-05-14
Avoid uninitialised in hw renderer + Only sync depth if necessary

JayFoxRox:
The latest bugs have reminded me that I should run valgrind more often.
I then noticed that pretty much every GL state caused 1 uninitalised [sic, valgrind] access.
This PR fixes this by value-initializing those variables.
It's really not that important though because the access is basically: `if (new_value != current_value) { current_value = new_value; dirty = true; }` (the logic still works, it's just not nice to do it).

Furthermore I noticed how different my depth-modifier uniform synced compared to every other uniform sync in gl_rasterizer. This PR fixes that in another commit too.


Subv:
LGTM

bunnei:
LGTM

PR #1794, opened 2016-05-14
Memory: Fixed a regression caused by #1695 and #1689.

### LINKED ISSUE #1790: Shared Memory PR Caused Black Screens
May be an issue with compatibility to recent merges as this PR once worked before when i merged it myself. Using latest master and games pop up with memory errors in the log and a black screen. Some games show fatal error. I would like to point out I have my shared font and config and every necessary file dumped and in the proper location. 

[LOG](http://pastebin.com/n4nn8Yuh)

Citra Version: 18b517e
Specs:
GTX 970
I5 4670K
8GB RAM

Subv:
Reserve enough space in the vector that holds the linear heap memory to prevent relocations of the backing memory when growing too much.

This can be considered a temporary fix until we rewrite all our service handlers to not depend on GetPointer.

Closes #1790


MerryMage:
LGTM.

bunnei:
LGTM

PR #1797, opened 2016-05-15
DSP/HLE: Implement mixer processing

MerryMage:
Includes audio output.

Only implemented well-understood portions of the firmware.

### DIFF
--- a/src/audio_core/hle/dsp.cpp
+++ b/src/audio_core/hle/dsp.cpp
@@ -88,6 +88,11 @@ static StereoFrame16 GenerateCurrentFrame() {
 static std::unique_ptr<AudioCore::Sink> sink;
 static AudioCore::TimeStretcher time_stretcher;
 
+static void OutputCurrentFrame(const StereoFrame16& frame) {
+    time_stretcher.AddSamples(&frame[0][0], frame.size());
+    sink->EnqueueSamples(time_stretcher.Process(sink->SamplesInQueue()));
+}
+
 // Public Interface
 
 void Init() {
@@ -120,6 +125,8 @@ bool Tick() {
 
     // TODO: Check dsp::DSP semaphore (which indicates emulated application has finished writing to shared memory region)
     current_frame = GenerateCurrentFrame();
+
+    OutputCurrentFrame(current_frame);
 
     return true;
 }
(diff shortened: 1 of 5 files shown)


src/audio_core/hle/mixers.cpp
```diff
@@ -0,0 +1,203 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
```
lioncash: This isn't needed in the cpp file

src/audio_core/hle/mixers.cpp
```diff
@@ -0,0 +1,203 @@
...
+    if (state.mixer2_enabled) {
+        for (size_t sample = 0; sample < samples_per_frame; sample++) {
+            for (size_t channel = 0; channel < 4; channel++) {
+                state.intermediate_mix_buffer[2][sample][channel] = read_samples.mix2.pcm32[channel][sample];
+            }
+        }
+    }
+}
+
+void Mixers::AuxSend(IntermediateMixSamples& write_samples, const std::array<QuadFrame32, 3> input) {
```
lioncash: was `input` supposed to be a const reference?

MerryMage: Yes

MerryMage:
Thanks @lioncash.

src/audio_core/hle/dsp.cpp
```diff
@@ -75,17 +121,12 @@ void Shutdown() {
 }
 
 bool Tick() {
-    SharedMemory& read = ReadRegion();
-    SharedMemory& write = WriteRegion();
+    StereoFrame16 current_frame = {};
 
-    std::array<QuadFrame32, 3> intermediate_mixes = {};
+    // TODO(merry): Only perform audio configuration when emulated application signals us to do so.
```
Subv: What does this comment mean?

MerryMage: Would something like "TODO: Check dsp::DSP semaphore" be clearer?

Subv: Not really, that still doesn't explain why you have to check it

MerryMage: Ah. "TODO: Check dsp::DSP semaphore (which indicates emulated application has finished writing to shared memory region)"

Subv: That's better

MerryMage:
Addressed @Subv's comment above.

MerryMage:
Squash

bunnei:
LGTM

PR #1799, opened 2016-05-15
DSP_DSP: Remove GetHeadphoneStatus logspam

MerryMage:
Downgrade from `WARNING` to `DEBUG` since we don't need to know it's being called x times a second.

May make headphones being inserted a config option in the future.


JayFoxRox:
LGTM

Subv:
LGTM

bunnei:
LGTM

PR #1800, opened 2016-05-15
Set fpscr for new threads

### LINKED ISSUE #1777: Bad initial fpscr
As noted on IRC yesterday we have a problem with `fpscr` (and maybe other registers such as `fpexc`).

When running homebrew compiled by Devkit Pro in citra and reading fpscr in `main()` we get `0x00000000`.
Running the same code on a real 3DS returns `0x03C00010`.
That'd be "Inexact cumulative flag" | "Round to Zero" | "Flush to Zero" | "Default NaN mode".
(Code: `context.fpscr = FPSCR_DEFAULT_NAN | FPSCR_FLUSH_TO_ZERO | FPSCR_ROUND_TOZERO | FPSCR_IXC; // 3DS returns 0x03C00010`)

Subv looked into it and said "the MPCore reference says FPSCR and FPEXC are reset to 0".
Also "looking at the kernel code that sets the fpscr in the thread saved-context page, and it initializes it with 0x3C00000 [...] a few layers inside the CreateThread SVC handler"

I assumed it might be inherited from the creating thread and Subv replied: "kernel code says it's just fixed to 0x3C00000, no inheritance going on"

Closing argument was "what you should do is create a fresh thread with svcCreateThread and check the fpscr on that" - which is exactly what I'll do if nobody else does it.
Unfortunaly we can't easily do this for `fpexc` which can only be accessed by privileged code.

JayFoxRox:
~~Improves~~ Fixes #1777 ~~(= does not fix it completly!)~~

~~Doing this in svcCreateThread doesn't work (its too late probably because the thread is already scheduled after creation in the kernel code).
This should really be reset to 0 on CPU level and only set to something else in the Kernel / Service HLE (probably also true for other regs in that function?)~~

I'm stupid, MerryMage is the best :heart: 

---

[HW-Tested using this](https://github.com/JayFoxRox/3ds-tests/commit/3eb17d9e7f4f310ebf1d989a81d11bf36ca2bac0)

Result on actual 3DS (by einstein95):
```
old_fpscr: 0x03C00010
new_fpscr: 0x03C00000
old_fpscr: 0x00000000
new_fpscr: 0x03C00000
old_fpscr: 0x03C00000
new_fpscr: 0x03C00000
old_fpscr: 0x03C00010
new_fpscr: 0x03C00000
```

Citra:
```
old_fpscr: 0x00000000
new_fpscr: 0x00000000
old_fpscr: 0x00000000
new_fpscr: 0x00000000
old_fpscr: 0x03C00000
new_fpscr: 0x00000000
old_fpscr: 0x03C00010
new_fpscr: 0x00000000
```

Citra + this PR:
```
old_fpscr: 0x03C00010
new_fpscr: 0x03C00000
old_fpscr: 0x00000000
new_fpscr: 0x03C00000
old_fpscr: 0x03C00000
new_fpscr: 0x03C00000
old_fpscr: 0x03C00010
new_fpscr: 0x03C00000
```

~~Note that there is still a bug with the first fpscr read from `main()`.
I'm assuming this means that the OS does some other stuff when creating a new process which sets the inexact math bit. We should probably emulate that behaviour but I don't know where.
This is already WAY better than what we currently have because Citra vs 3DS rounding actually causes off-by-one errors on a lot of calculations.~~

### DIFF
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -526,6 +526,8 @@ SharedPtr<Thread> SetupMainThread(u32 entry_point, s32 priority) {
 
     SharedPtr<Thread> thread = thread_res.MoveFrom();
 
+    thread->context.fpscr = FPSCR_DEFAULT_NAN | FPSCR_FLUSH_TO_ZERO | FPSCR_ROUND_TOZERO | FPSCR_IXC; // 0x03C00010
+
     // Run new "main" thread
     SwitchContext(thread.get());
 
--- a/src/core/arm/arm_interface.h
+++ b/src/core/arm/arm_interface.h
@@ -6,6 +6,7 @@
 
 #include "common/common_types.h"
 #include "core/arm/skyeye_common/arm_regformat.h"
+#include "core/arm/skyeye_common/vfp/asm_vfp.h"
 
 namespace Core {
     struct ThreadContext;
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -502,6 +502,9 @@ static ResultCode CreateThread(Handle* out_handle, s32 priority, u32 entry_point
 
     CASCADE_RESULT(SharedPtr<Thread> thread, Kernel::Thread::Create(
             name, entry_point, priority, arg, processor_id, stack_top));
+
+    thread->context.fpscr = FPSCR_DEFAULT_NAN | FPSCR_FLUSH_TO_ZERO | FPSCR_ROUND_TOZERO; // 0x03C00000
+
     CASCADE_RESULT(*out_handle, Kernel::g_handle_table.Create(std::move(thread)));
 
     LOG_TRACE(Kernel_SVC, "called entrypoint=0x%08X (%s), arg=0x%08X, stacktop=0x%08X, "


MerryMage:
Not convinced that's the best place to put it. How about:

```
diff --git a/src/core/hle/kernel/thread.cpp b/src/core/hle/kernel/thread.cpp
index 68f0269..a3abf95 100644
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -526,6 +526,8 @@ SharedPtr<Thread> SetupMainThread(u32 entry_point, s32 priority) {

     SharedPtr<Thread> thread = thread_res.MoveFrom();

+    thread->context.fpscr = 0x03C00010; // FPSCR_DEFAULT_NAN | FPSCR_FLUSH_TO_ZERO | FPSCR_ROUND_TOZERO | FPSCR_IXC
+
     // Run new "main" thread
     SwitchContext(thread.get());

diff --git a/src/core/hle/svc.cpp b/src/core/hle/svc.cpp
index 3a53126..72ef760 100644
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -504,6 +504,8 @@ static ResultCode CreateThread(Handle* out_handle, s32 priority, u32 entry_point
             name, entry_point, priority, arg, processor_id, stack_top));
     CASCADE_RESULT(*out_handle, Kernel::g_handle_table.Create(std::move(thread)));

+    thread->context.fpscr = 0x03C00000; // FPSCR_DEFAULT_NAN | FPSCR_FLUSH_TO_ZERO | FPSCR_ROUND_TOZERO
+
     LOG_TRACE(Kernel_SVC, "called entrypoint=0x%08X (%s), arg=0x%08X, stacktop=0x%08X, "
         "threadpriority=0x%08X, processorid=0x%08X : created handle=0x%08X", entry_point,
         name.c_str(), arg, stack_top, priority, processor_id, *out_handle);

```

JayFoxRox:
Doing what Merry suggested. This also fixes the other bug now. :smile:

src/core/hle/kernel/thread.cpp
```diff
@@ -526,6 +526,8 @@ SharedPtr<Thread> SetupMainThread(u32 entry_point, s32 priority) {
 
     SharedPtr<Thread> thread = thread_res.MoveFrom();
 
+    thread->context.fpscr = 0x03C00010; // FPSCR_DEFAULT_NAN | FPSCR_FLUSH_TO_ZERO | FPSCR_ROUND_TOZERO | FPSCR_IXC
```
lioncash: If you include core/arm/skyeye_common/vfp/asm_vfp.h, you can just build the mask directly instead of needing to append a comment after it.

JayFoxRox: I thought about that, but I also decided to do what Merry did because it would add a dependency on the skycom code.
We could probably move those bits to the ARM interface (which also has functions to modify the CPSR) but I wasn't sure how to name them.

JayFoxRox:
Fixed use after `std::move`. Works fine, hw-tested.

Ready to go?

JayFoxRox:
Fixed use after `std::move`. Also did what @lioncash suggested because arm_interface was already including another skycom include anyway.

Works fine, hw-tested. Ready to go?

Subv:
LGTM

bunnei:
LGTM

PR #1809, opened 2016-05-16
Update ACT:U and create ACT:A

Andrix44:
Everything is taken from [3dbrew](https://www.3dbrew.org/wiki/ACT_Services).

Sorry for making 6 commits, this is the first time I actually do something on GitHub. I'll try to figure out more, but now I simply edited the files on the webpage and unfortunately every change made a new commit.


Subv:
LGTM, you (Or whoever merges it) just need to squash the commits down to a single one.

PR #1810, opened 2016-05-16
Fix float exceptions

JayFoxRox:
This is a followup to #1568 

There is an error in master where [denormals would be flushed in `vfp_double_unpack`](https://github.com/citra-emu/citra/blob/master/src/core/arm/skyeye_common/vfp/vfp_helper.h#L360)
After flushing the denormal exception is set in fpscr.

Note how `fpscr` is a pointer. However, it's [passed *by value* to `vfp_double_cpdo`](https://github.com/citra-emu/citra/blob/master/src/core/arm/skyeye_common/vfp/vfpdouble.cpp#L1178).
\- This means that the denormal exception is never propagated back to the CPU context.
(Same problem exists for single precision float too)

---

All other exceptions are normally propagated using an `u32 exceptions` return value.
This PR fixes the issue by returning a `u32 exceptions` from the unpack functions.

Furthermore I went through the code and made sure to replace `exception =` by `exception |=` to avoid overwriting past exceptions.
I didn't test this last change though so we should probably review or even hw-test this.

---

Combined with #1568 this will cause 0 errors in the test described / linked in #1568 (which only tests ftoi!):

```
Result for Hardware (wwylele)/out-ftoi32-0x03000000.txt:
  master: 60356
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi32-0x0300009F.txt:
  master: 2716
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi32-0x03400000.txt:
  master: 59420
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi32-0x03800000.txt:
  master: 74766
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi32-0x03C00000.txt:
  master: 59420
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi64-0x03000000.txt:
  master: 57850
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi64-0x0300009F.txt:
  master: 1356
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi64-0x03400000.txt:
  master: 57850
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi64-0x03800000.txt:
  master: 72900
  sdmc: 0
Result for Hardware (wwylele)/out-ftoi64-0x03C00000.txt:
  master: 57850
  sdmc: 0
```

### DIFF
--- a/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
+++ b/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
@@ -291,8 +291,9 @@ static u32 vfp_double_fsqrt(ARMul_State* state, int dd, int unused, int dm, u32 
     LOG_TRACE(Core_ARM11, "In %s", __FUNCTION__);
     vfp_double vdm, vdd, *vdp;
     int ret, tm;
+    u32 exceptions = 0;
 
-    vfp_double_unpack(&vdm, vfp_get_double(state, dm), &fpscr);
+    exceptions |= vfp_double_unpack(&vdm, vfp_get_double(state, dm), fpscr);
 
     tm = vfp_double_type(&vdm);
     if (tm & (VFP_NAN|VFP_INFINITY)) {
@@ -369,7 +370,9 @@ sqrt_invalid:
     }
     vdd.significand = vfp_shiftright64jamming(vdd.significand, 1);
 
-    return vfp_double_normaliseround(state, dd, &vdd, fpscr, 0, "fsqrt");
+    exceptions |= vfp_double_normaliseround(state, dd, &vdd, fpscr, 0, "fsqrt");
+
+    return exceptions;
 }
 
 /*
@@ -475,7 +478,7 @@ static u32 vfp_double_fcvts(ARMul_State* state, int sd, int unused, int dm, u32 
     u32 exceptions = 0;
 
     LOG_TRACE(Core_ARM11, "In %s", __FUNCTION__);
-    vfp_double_unpack(&vdm, vfp_get_double(state, dm), &fpscr);
+    exceptions |= vfp_double_unpack(&vdm, vfp_get_double(state, dm), fpscr);
 
     tm = vfp_double_type(&vdm);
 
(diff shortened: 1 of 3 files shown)


lioncash:
Looks fine.

src/core/arm/skyeye_common/vfp/vfpdouble.cpp
```diff
@@ -1134,13 +1137,13 @@ static u32 vfp_double_fdiv(ARMul_State* state, int dd, int dn, int dm, u32 fpscr
     return vfp_double_normaliseround(state, dd, &vdd, fpscr, 0, "fdiv");
```
JayFoxRox: Missing propagation of exception here?

src/core/arm/skyeye_common/vfp/vfpsingle.cpp
```diff
@@ -1165,13 +1168,13 @@ static u32 vfp_single_fdiv(ARMul_State* state, int sd, int sn, s32 m, u32 fpscr)
     return vfp_single_normaliseround(state, sd, &vsd, fpscr, 0, "fdiv");
```
JayFoxRox: Also missing propagation of exception here?
Why does `vfp_single_normaliseround` have an exception parameter in the first place?

lioncash: To be honest, "Why?" is probably the most frequently asked question in my head whenever I need to modify/look at the VFP code.

src/core/arm/skyeye_common/vfp/vfpsingle.cpp
```diff
@@ -1180,7 +1197,7 @@ static u32 vfp_single_fdiv(ARMul_State* state, int sd, int sn, s32 m, u32 fpscr)
     goto pack;
 
 divzero:
-    exceptions = FPSCR_DZC;
+    exceptions |= FPSCR_DZC;
 infinity:
     vsd.exponent = 255;
     vsd.significand = 0;
```
JayFoxRox: There is a path in this file which *still* returns a hardcoded value.. will fix later

bunnei:
Code looks fine to me as well. Is this still waiting on anything?

JayFoxRox:
Not waiting on anything. This should be good to go.

PR #1812, opened 2016-05-16
Retrieve shader result from new OutputRegisters-type

JayFoxRox:
In `master`, running a shader automaticly generates an output vertex.
Due to the conversion (output registers --> vertex) being lossy we can't use the data to forward a vertex from the VS to the GS easily.

A shader execution only modifies the UnitState output registers anyway, so I introduce OutputRegisters instead.
These can be converted to an OutputVertex for the rasterizer or be used as input to GS later.

(This PR is part of my GS branch)

---

I don't particularly like replacing `state.registers.output` by `state.output_registers.value` but I wasn't sure how to do it nicely otherwise. At least this way the `ToVertex` function is tied to the `OutputRegisters`.

### DIFF
--- a/src/video_core/shader/shader.h
+++ b/src/video_core/shader/shader.h
@@ -84,6 +84,15 @@ struct OutputVertex {
 static_assert(std::is_pod<OutputVertex>::value, "Structure is not POD");
 static_assert(sizeof(OutputVertex) == 32 * sizeof(float), "OutputVertex has invalid size");
 
+struct OutputRegisters {
+    OutputRegisters() = default;
+
+    alignas(16) Math::Vec4<float24> value[16];
+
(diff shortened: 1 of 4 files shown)


src/video_core/shader/shader.h
```diff
@@ -84,6 +84,15 @@ struct OutputVertex {
 static_assert(std::is_pod<OutputVertex>::value, "Structure is not POD");
 static_assert(sizeof(OutputVertex) == 32 * sizeof(float), "OutputVertex has invalid size");
 
+struct OutputRegisters {
```
MerryMage: I'm assuming this is not inline in `UnitState` because gs magic is going to happen to this?

MerryMage:
Looks fine to me.

bunnei:
LGTM as well.

PR #1813, opened 2016-05-16
UXTB16: Avoid undefined behaviour (shift past width)

JayFoxRox:
`rotation` could be zero here.

Solution already [used in SXTB16](https://github.com/citra-emu/citra/blob/master/src/core/arm/dyncom/arm_dyncom_interpreter.cpp#L6259)

---

Edit: We (#citra-dev) noticed this is an ongoing problem in this file, hidden by `ROTATE_RIGHT` and `ROTATE_LEFT` macros. I have an ugly fix but I'd rather not extend this PR as I already have too many PRs to maintain.


JayFoxRox:
Closing because it's probably not important.
I'll probably create an issue soon. We should fix this using a function and possibly mix this with some cleanup.

Issue #862, opened 2015-06-14
ffmpeg fails to decode a video correctly - Wrong ARMv6 emulation ?


Lectem:
I'm writing some H264/mpeg4 player, and the video decoding works fine on HW (see https://youtu.be/x1-xWt-xyAc ).
However on Citra it seems that the frames arent decoded correctly. I can sometimes see the after effect of motions, but it's as if the decoded frames were transparent (hence the green screen).
Here is a screenshot : 

![Image of decoding problem](http://lectem.fr/dl/citra-ffmpeg-problem.png)

You can download the .3dsx here [3Damnesic.zip](https://github.com/Lectem/3Damnesic/releases/download/alpha/3Damnesic.zip).
It will try to read the file test.mp4 at the root of your sdmc.
test.mp4 has to be a mpeg4 video 400*240 (do not use another resolution).

FFmpeg uses some ARMV6 asm code to optimize the decoding process, my guess is that the problem comes from this. Also note that it is NOT using y2r. It's all software based (might use some vfp though).

Source is available at https://github.com/Lectem/3Damnesic/

Edit: this post used to say h.264, but only mpeg4 caused the issue.

Subv:
There were some fixes to the dyncom lately, @Lectem could you check if this is still a problem please? I don't have any 400 * 240 h264 video handy

Lectem:
@Subv Same problem

aroulin:
Can you check again ? A short video worked for me (although slow).

Lectem:
@aroulin I wont be able to verify it soon, I have to reinstall qt and don't have a proper connection. Are you sure your video is mpeg4 and not h264?

aroulin:
@Lectem Sorry, I did not know there was a difference. Pardon my limited knowledge of video codecs :)
I have been able to reproduce the issue, I'll try to look into it, thanks.

Subv:
Can you try with mfloat-abi=soft? It's probably the same issue as #1218

Lectem:
I managed to disable VFP usage and to use mfloat-abi=soft with devkitARM somehow, still the same issue.

linkmauve:
I just reproduced using VP8 instead, I can discern the video’s features, but unlike on the 3DS there are green blocks, magenta blocks, striped blocks, and a cyan frame every few frames.

Lectem:
Works fine thanks to #1814 !

PR #1814, opened 2016-05-18
Fix read-after-write in SMUAD, SMLAD, SMUSD, SMLSD

### LINKED ISSUE #1636: ffmpeg fails to render VP8 videos properly

JayFoxRox:
Fixes  #1636  and probably #862 .

There is a read-after-write in SMLAD (etc.) if Rd = Ra.
These [instructions are present in ffmpeg](https://github.com/FFmpeg/FFmpeg/blob/e5920425b02b0feaf45971bf1c8b8abc541f5ae1/libavcodec/arm/vp8dsp_armv6.S#L1522) where this causes bugs.

This PR fixes it.

I've also searched for `RD |=`, `RD +=` etc. in this file (and only this file!) but couldn't find anything. So hopefully it was the only bug of this kind.

### DIFF
--- a/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
@@ -5527,28 +5527,32 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
 
             // SMUAD and SMLAD
             if (BIT(op2, 1) == 0) {
-                RD = (product1 + product2);
+                u32 rd_val = (product1 + product2);
 
                 if (inst_cream->Ra != 15) {
-                    RD += cpu->Reg[inst_cream->Ra];
+                    rd_val += cpu->Reg[inst_cream->Ra];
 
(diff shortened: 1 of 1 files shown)


linkmauve:
I confirm this fixes #1636, and LGTM.

lioncash:
LGTM

PR #1816, opened 2016-05-18
Refactor Tev dump

JayFoxRox:
This will make it easier to have a GUI combiner / Tev debugger later.
It also extends the Tev dumper by new source types and operations.

I've decided to only prototype the 3 most important functions in the header:
- One to get a string for a stage alpha channel,
- one for the color channel
- and the function to dump all stages at once to `LOG_TRACE` (which existed before).

This also changes display of the functions from byte style to float style (which is more common in the graphics world). If we end up wanting to show actual GPU register values we'll have to replace said code anyway + I don't think it will happen anytime soon.

---

It should be noted that the current `DumpTevStageConfig` function is pretty useless because the `Pica::g_state.regs.GetTevStages()` call is usually inlined = we can't do it live in a debugging session with optimized builds.
\- I'll probably work on a GUI later anyway so this doesn't matter to me.

### DIFF
--- a/src/video_core/debug_utils/debug_utils.cpp
+++ b/src/video_core/debug_utils/debug_utils.cpp
@@ -708,28 +708,52 @@ void DumpTevStageConfig(const std::array<Pica::Regs::TevStageConfig,6>& stages)
         const auto& tev_stage = stages[index];
 
         static const std::map<Source, std::string> source_map = {
-            { Source::PrimaryColor, "PrimaryColor" },
-            { Source::Texture0, "Texture0" },
-            { Source::Texture1, "Texture1" },
-            { Source::Texture2, "Texture2" },
-            { Source::Constant, "Constant" },
-            { Source::Previous, "Previous" },
+            { Source::PrimaryColor,           "PrimaryColor" },
+            { Source::PrimaryFragmentColor,   "PrimaryFragmentColor" },
+            { Source::SecondaryFragmentColor, "SecondaryFragmentColor" },
+            { Source::Texture0,               "Texture0" },
+            { Source::Texture1,               "Texture1" },
+            { Source::Texture2,               "Texture2" },
+            { Source::Texture3,               "Texture3" },
+            { Source::PreviousBuffer,         "PreviousBuffer" },
+            { Source::Constant,               "Constant" },
+            { Source::Previous,               "Previous" },
         };
 
         static const std::map<ColorModifier, std::string> color_modifier_map = {
-            { ColorModifier::SourceColor, { "%source.rgb" } },
-            { ColorModifier::SourceAlpha, { "%source.aaa" } },
+            { ColorModifier::SourceColor,         "%source.rgb" },
+            { ColorModifier::OneMinusSourceColor, "(1.0 - %source.rgb)" },
+            { ColorModifier::SourceAlpha,         "%source.aaa" },
+            { ColorModifier::OneMinusSourceAlpha, "(1.0 - %source.aaa)" },
+            { ColorModifier::SourceRed,           "%source.rrr" },
+            { ColorModifier::OneMinusSourceRed,   "(1.0 - %source.rrr)" },
+            { ColorModifier::SourceGreen,         "%source.ggg" },
+            { ColorModifier::OneMinusSourceGreen, "(1.0 - %source.ggg)" },
+            { ColorModifier::SourceBlue,          "%source.bbb" },
+            { ColorModifier::OneMinusSourceBlue,  "(1.0 - %source.bbb)" },
         };
+
         static const std::map<AlphaModifier, std::string> alpha_modifier_map = {
-            { AlphaModifier::SourceAlpha, "%source.a" },
-            { AlphaModifier::OneMinusSourceAlpha, "(255 - %source.a)" },
+            { AlphaModifier::SourceAlpha,         "%source.a" },
+            { AlphaModifier::OneMinusSourceAlpha, "(1.0 - %source.a)" },
+            { AlphaModifier::SourceRed,           "%source.r" },
+            { AlphaModifier::OneMinusSourceRed,   "(1.0 - %source.r)" },
+            { AlphaModifier::SourceGreen,         "%source.g" },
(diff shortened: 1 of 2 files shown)


src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -696,106 +696,110 @@ void DumpTexture(const Pica::Regs::TextureConfig& texture_config, u8* data) {
...
-        stage_info += "Stage " + std::to_string(index) + ": " + GetColorCombinerStr(tev_stage) + "   " + GetAlphaCombinerStr(tev_stage) + "\n";
-    }
+std::string GetTevStageConfigAlphaCombinerStr(const Pica::Regs::TevStageConfig& tev_stage) {
+    auto op_str = GetTevStageConfigOperationStr(tev_stage.alpha_op);
+    op_str = ReplacePattern(op_str, "%source1", GetTevStageConfigAlphaSourceStr(tev_stage.alpha_source1, tev_stage.alpha_modifier1));
+    op_str = ReplacePattern(op_str, "%source2", GetTevStageConfigAlphaSourceStr(tev_stage.alpha_source2, tev_stage.alpha_modifier2));
+    return   ReplacePattern(op_str, "%source3", GetTevStageConfigAlphaSourceStr(tev_stage.alpha_source3, tev_stage.alpha_modifier3));
+}
 
+void DumpTevStageConfig(const std::array<Pica::Regs::TevStageConfig,6>& stages) {
```
lioncash: Space after the comma

src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -696,106 +696,110 @@ void DumpTexture(const Pica::Regs::TextureConfig& texture_config, u8* data) {
...
+        { Source::Texture0,               "Texture0" },
+        { Source::Texture1,               "Texture1" },
+        { Source::Texture2,               "Texture2" },
+        { Source::Texture3,               "Texture3" },
+        { Source::PreviousBuffer,         "PreviousBuffer" },
+        { Source::Constant,               "Constant" },
+        { Source::Previous,               "Previous" },
+    };
+
+    auto src_it = source_map.find(source);
```
lioncash:
```cpp
const auto src_it = source_map.find(source);
if (src_it == source_map.end())
    return "Unknown";

return src_it->second;
```

src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -696,106 +696,110 @@ void DumpTexture(const Pica::Regs::TextureConfig& texture_config, u8* data) {
...
+    size_t start = input.find(pattern);
+    if (start == std::string::npos)
+        return input;
+
+    std::string ret = input;
+    ret.replace(start, pattern.length(), replacement);
+    return ret;
+}
+
+static std::string GetTevStageConfigSourceStr(const Pica::Regs::TevStageConfig::Source& source) {
```
lioncash: You may as well just long-form Str into String

src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -696,106 +696,110 @@ void DumpTexture(const Pica::Regs::TextureConfig& texture_config, u8* data) {
...
+        { Operation::Modulate, "(%source1 * %source2)" },
+        { Operation::Add, "(%source1 + %source2)" },
+        { Operation::AddSigned, "(%source1 + %source2) - 0.5" },
+        { Operation::Lerp, "lerp(%source1, %source2, %source3)" },
+        { Operation::Subtract, "(%source1 - %source2)" },
+        { Operation::Dot3_RGB, "dot(%source1, %source2)" },
+        { Operation::MultiplyThenAdd, "((%source1 * %source2) + %source3)" },
+        { Operation::AddThenMultiply, "((%source1 + %source2) * %source3)" },
+    };
+    auto op_it = combiner_map.find(operation);
```
lioncash:
```cpp
const auto op_it = combiner_map.find(operation);
if (op_it == combiner_map.end())
    return "Unknown op (%source1, %source2, %source3)";
  
return op_it->second;
```

JayFoxRox:
Addressed what @lioncash said

src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -696,106 +696,111 @@ void DumpTexture(const Pica::Regs::TextureConfig& texture_config, u8* data) {
...
+        return "Unknown";
+
+    return src_it->second;
+}
+
+static std::string GetTevStageConfigColorSourceString(const Pica::Regs::TevStageConfig::Source& source, const Pica::Regs::TevStageConfig::ColorModifier modifier) {
     using ColorModifier = Pica::Regs::TevStageConfig::ColorModifier;
+    static const std::map<ColorModifier, std::string> color_modifier_map = {
+        { ColorModifier::SourceColor, { "%source.rgb" } },
+        { ColorModifier::SourceAlpha, { "%source.aaa" } },
```
JayFoxRox: Missing a couple of modes here. I'll add them tomorrow

src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -696,106 +696,111 @@ void DumpTexture(const Pica::Regs::TextureConfig& texture_config, u8* data) {
...
+        modifier_str = modifier_it->second;
+
+    return ReplacePattern(modifier_str, "%source", src_str);
+}
+
+static std::string GetTevStageConfigAlphaSourceString(const Pica::Regs::TevStageConfig::Source& source, const Pica::Regs::TevStageConfig::AlphaModifier modifier) {
     using AlphaModifier = Pica::Regs::TevStageConfig::AlphaModifier;
+    static const std::map<AlphaModifier, std::string> alpha_modifier_map = {
+        { AlphaModifier::SourceAlpha, "%source.a" },
+        { AlphaModifier::OneMinusSourceAlpha, "(1.0 - %source.a)" },
```
JayFoxRox: Missing a couple of modes here. I'll add them tomorrow

bunnei:
LGTM

PR #1817, opened 2016-05-18
Improve SMDH support in loaders and frontends

linkmauve:
This PR should be reviewed commit per commit.

The first four commits are some loader cleanup, so a loader can be created more easily~~, and the final one adds SMDH support to the SDL2 frontend~~ (pushed back to a future PR).

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -272,7 +272,15 @@ bool GMainWindow::InitializeSystem() {
 }
 
 bool GMainWindow::LoadROM(const std::string& filename) {
-    Loader::ResultStatus result = Loader::LoadFile(filename);
+    std::unique_ptr<Loader::AppLoader> app_loader = Loader::GetFileLoader(filename);
+    if (!app_loader) {
+        LOG_CRITICAL(Frontend, "Failed to obtain loader for %s!", filename.c_str());
+        QMessageBox::critical(this, tr("Error while loading ROM!"),
+                              tr("The ROM format is not supported."));
+        return false;
+    }
+
+    Loader::ResultStatus result = app_loader->Load();
     if (Loader::ResultStatus::Success != result) {
         LOG_CRITICAL(Frontend, "Failed to load ROM!");
         System::Shutdown();
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -272,7 +272,7 @@ bool GMainWindow::InitializeSystem() {
 }
 
 bool GMainWindow::LoadROM(const std::string& filename) {
-    std::unique_ptr<Loader::AppLoader> app_loader = Loader::GetFileLoader(filename);
+    std::unique_ptr<Loader::AppLoader> app_loader = Loader::GetLoader(filename);
     if (!app_loader) {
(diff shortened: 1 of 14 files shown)


src/citra_qt/main.cpp
```diff
@@ -272,7 +272,13 @@ bool GMainWindow::InitializeSystem() {
 }
 
 bool GMainWindow::LoadROM(const std::string& filename) {
-    Loader::ResultStatus result = Loader::LoadFile(filename);
+    std::unique_ptr<Loader::AppLoader> app_loader = Loader::GetLoader(filename);
+    if (!app_loader) {
+        LOG_CRITICAL(Frontend, "Failed to obtain loader for %s!", filename.c_str());
+        return false;
```
jroweboy: Looking at the other error state for the LoadROM method, they also call `System::Shutdown()` and display an error in a `QMessageBox` Shouldn't failing to get the loader do the same? (Sorry, I can't check right now to see if System Shutdown actually needs to be called :frowning:)

src/core/loader/smdh.cpp
```diff
@@ -0,0 +1,53 @@
...
+    return Loader::MakeMagic('S', 'M', 'D', 'H') == magic;
+}
+
+std::vector<u16> SMDH::GetIcon(bool large) const {
+    u32 size;
+    const u16* icon_data;
+
+    if (large) {
+        size = 48;
+        icon_data = reinterpret_cast<const u16*>(large_icon.data());
```
lioncash: You've introduced undefined behavior with this instead of keeping it as a u8 like the previous code was using.

src/citra_qt/main.cpp
wwylele: I agree to add a `QMessageBox`, however `System::Shutdown` is not needed because it hasn't trying to initialize the core system

linkmauve:
Here, fixed both comments.

src/core/loader/smdh.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "video_core/utils.h"
+
+namespace Loader {
+
+bool IsValidSMDH(const std::vector<u8>& smdh_data) {
+    if (smdh_data.size() < sizeof(Loader::SMDH))
+        return false;
+
+    u32 magic;
+    memcpy(&magic, smdh_data.data(), 4);
```
lioncash:
```cpp
memcpy(&magic, smdh_data.data(), sizeof(u32));
```

src/core/loader/smdh.cpp
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+        size = 24;
+        icon_data = small_icon.data();
+    }
+
+    std::vector<u16> icon(size * size);
+    for (u32 x = 0; x < size; ++x) {
+        for (u32 y = 0; y < size; ++y) {
+            u32 coarse_y = y & ~7;
+            const u8 *pixel = icon_data + VideoCore::GetMortonOffset(x, y, 2) + coarse_y * size * 2;
```
lioncash: asterisk against the type.

linkmauve:
Addressed these comments.

Subv:
LGTM

bunnei:
LGTM

PR #1824, opened 2016-05-20
Config: Restore previously selected audio sink option

### LINKED ISSUE #1823: Qt audio config does not load previous setting
If you open the Qt "Configuration" GUI, the audio "Output Engine" selection always resets to "auto". For example, if you have the "null" output engine enabled, then simply open the configuration and hit "OK" (even without going to the "Audio" tab), audio will be enabled.

jroweboy:
Fixes #1823

### DIFF
--- a/src/citra_qt/configure_audio.cpp
+++ b/src/citra_qt/configure_audio.cpp
@@ -28,14 +28,14 @@ ConfigureAudio::~ConfigureAudio() {
 }
 
 void ConfigureAudio::setConfiguration() {
+    int new_sink_index = 0;
     for (int index = 0; index < ui->output_sink_combo_box->count(); index++) {
         if (ui->output_sink_combo_box->itemText(index).toStdString() == Settings::values.sink_id) {
-            ui->output_sink_combo_box->setCurrentIndex(index);
+            new_sink_index = index;
             break;
         }
     }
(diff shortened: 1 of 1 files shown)


lioncash:
LGTM

src/citra_qt/configure_audio.cpp
```diff
@@ -28,14 +28,13 @@ ConfigureAudio::~ConfigureAudio() {
 }
 
 void ConfigureAudio::setConfiguration() {
+    ui->output_sink_combo_box->setCurrentIndex(0);
     for (int index = 0; index < ui->output_sink_combo_box->count(); index++) {
         if (ui->output_sink_combo_box->itemText(index).toStdString() == Settings::values.sink_id) {
             ui->output_sink_combo_box->setCurrentIndex(index);
             break;
```
JayFoxRox: Wouldn't it be easier to simply `return;` here?

jroweboy: Sure. I can make that change. I didn't use a return though because I didn't wanna make a potential bug in the future when someone adds other options.

MerryMage:
Oops. LGTM.

JayFoxRox:
LGTM

Issue #1832, opened 2016-05-20
Pokemon Super Mystery Dungeon: freeze at intro (thread locking)


wwylele:
Did some investigation, so let me explain in detail.

On real 3DS, when the game start without savegame, it goes like this:
-  Some star and flash effect
- Title screen, with camera video in the background
-  Press start
-   Game says "Finally, ..." (camera video still in the background)
-  press lower screen and hold 
-  upper screen turn to white, lower screen turn to black 
-  (camera video is no longer there) The "personality test" begins

A reference video is [here](https://www.youtube.com/watch?v=WkLRCoGV28A).

And here is what happens on citra:
- Some star and flash effect 
-  Title screen, with green background with log spamed with `Service.CAM` (because we stubbed CAM service) 
- Press start
-  Game says those welcome text (still green background and log spamming)
-  press lower screen and hold 
- upper screen turn to white, lower screen turn to black
-  freeze (still spamming `Service.CAM`)

At freezing, the wait tree shows that the main thread is waiting for another thread (thread_id = 9). That thread continues calling CAM service.

Then I did some tracing and here is the result:
before title screen, main thread creates event A, then creates thread 9 (camera thread)
The camera thread is doing the follows:
call `CAM::SetReceiving`→...→ `WaitSynchronizationN(...)` → ... → loop
The `WaitSynchronizationN(...)` is waiting for ONE OF the following 3 objects (in array order):
- "event A"
- `CAM:completion_event_cam1`
- `CAM:interrupt_error_event`

However, when the game begins to show the welcome text, the waiting strategy of the camera thread changes somehow: the order of 3 objects for `WaitSynchronizationN(...)` changes to
- `CAM:completion_event_cam1`
- `CAM:interrupt_error_event`
- "event A"

the event A is never signaled until "screen turn to white, lower screen turn to black", where the event A is signaled and the "personality test" is supposed to begin. So this event is for sure the stop signal of camera thread. The problem is: this event becomes the last one in camera threads waiting list, and [we always signal `CAM:completion_event_cam1` immediately](https://github.com/citra-emu/citra/blob/98923ab6e566f06b61d2dc671a1b8aec87f88407/src/core/hle/service/cam/cam.cpp#L84), so `WaitSynchronizationN` always catches `CAM:completion_event_cam1` but never catches "event A", which means the camera thread never knows it should stop.

[...]

PR #1837, opened 2016-05-21
SVC::WaitSynchronizationN: Reschedule at the end

wwylele:
Fixes deadlock in an extreme situation:
Say an event A is scheduled to be signaled in 1000 cycles (such event can be GSP event, HID event, etc.). Now thread B calls `WaitSynchronizationN` to wait for A. In our old `WaitSynchronizationN`, the following will happen:
 - B checks A, and finds that it is not signaled;
 - B prepares to sleep (`wait_thread = true;`);
 - calls `HLE::Reschedule()`, where `Core::g_app_core->AddTicks(4000);` steps over A's cycles, and A gets signaled (without any thread waited)
 - Back to `WaitSynchronizationN`, B register itself to wait for A, and sleeps, leaving A in signaled state.
 - If A doesn't get signaled again, B will sleep forever.

The solution is to call `HLE::Reschedule()` at the end of `WaitSynchronizationN` instead, so that event A's state won't change during synchronization.

I found this when playing with #1832, where I gave the camera event a short delay, but caused the camera thread locked.

### DIFF
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -6,6 +6,7 @@
 
 #include "common/logging/log.h"
 #include "common/microprofile.h"
+#include "common/scope_exit.h"
 #include "common/string_util.h"
 #include "common/symbols.h"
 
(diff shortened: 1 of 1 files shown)


src/core/hle/svc.cpp
```diff
@@ -326,50 +327,51 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
         }
     }
 
-    HLE::Reschedule(__func__);
+    {
```
Subv: You don't need the extra scope here

src/core/hle/svc.cpp
```diff
@@ -326,50 +327,51 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
+        SCOPE_EXIT({HLE::Reschedule("WaitSynchronizationN");});
```
Subv: ``` // Reschedule after putting the threads to sleep.```

Subv:
LGTM

bunnei:
LGTM

PR #1842, opened 2016-05-22
Kernel: Added ClientPort and ServerPort classes, along with svcCreatePort.

Subv:
This is part of an ongoing effort to implement support for multiple processes.

For now the goal is to allow applications to register their own services via srv:RegisterService.

See the sequence diagram below to get an overview of how this all fits together.

ports_sessions_3ds

### DIFF
--- a/src/core/hle/kernel/client_port.h
+++ b/src/core/hle/kernel/client_port.h
@@ -1,0 +1,34 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <string>
+
+#include "common/common_types.h"
+
+#include "core/hle/kernel/kernel.h"
+
+namespace Kernel {
+
+class ClientPort : public Object {
+public:
+    friend class ServerPort;
+    std::string GetTypeName() const override { return "ClientPort"; }
+    std::string GetName() const override { return name; }
+
+    static const HandleType HANDLE_TYPE = HandleType::ClientPort;
+    HandleType GetHandleType() const override { return HANDLE_TYPE; }
+
+    SharedPtr<ServerPort> server_port;          ///< ServerPort associated with this client port.
+    u32 max_sessions;                           ///< Maximum number of simultaneous sessions the port can have
+    u32 active_sessions;                        ///< Number of currently open sessions to this port
+    std::string name;                           ///< Name of client port (optional)
+
+protected:
+    ClientPort();
+    ~ClientPort() override;
+};
+
+} // namespace
--- a/src/core/hle/kernel/client_port.h
+++ b/src/core/hle/kernel/client_port.h
@@ -12,6 +12,8 @@
 
 namespace Kernel {
 
+class ServerPort;
+
 class ClientPort : public Object {
 public:
     friend class ServerPort;
--- a/src/core/hle/kernel/kernel.h
+++ b/src/core/hle/kernel/kernel.h
@@ -35,7 +35,7 @@ enum KernelHandle : Handle {
 
 enum class HandleType : u32 {
     Unknown         = 0,
-    ServerPort      = 1,
+
     Session         = 2,
     Event           = 3,
     Mutex           = 4,
@@ -49,6 +49,7 @@ enum class HandleType : u32 {
     ResourceLimit   = 12,
     CodeSet         = 13,
     ClientPort      = 14,
+    ServerPort      = 15,
 };
 
 enum {
--- a/src/core/hle/kernel/kernel.h
+++ b/src/core/hle/kernel/kernel.h
@@ -35,7 +35,7 @@ enum KernelHandle : Handle {
 
 enum class HandleType : u32 {
     Unknown         = 0,
-    Port            = 1,
+    ServerPort      = 1,
     Session         = 2,
     Event           = 3,
     Mutex           = 4,
@@ -48,6 +48,7 @@ enum class HandleType : u32 {
     Timer           = 11,
     ResourceLimit   = 12,
     CodeSet         = 13,
+    ClientPort      = 14,
 };
 
 enum {
@@ -72,6 +73,7 @@ public:
     bool IsWaitable() const {
         switch (GetHandleType()) {
         case HandleType::Session:
+        case HandleType::ServerPort:
         case HandleType::Event:
         case HandleType::Mutex:
         case HandleType::Thread:
@@ -80,13 +82,13 @@ public:
             return true;
 
         case HandleType::Unknown:
-        case HandleType::Port:
         case HandleType::SharedMemory:
         case HandleType::Redirection:
         case HandleType::Process:
         case HandleType::AddressArbiter:
         case HandleType::ResourceLimit:
         case HandleType::CodeSet:
+        case HandleType::ClientPort:
             return false;
         }
     }
--- a/src/core/hle/kernel/server_port.cpp
+++ b/src/core/hle/kernel/server_port.cpp
@@ -2,8 +2,11 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <tuple>
+
 #include "common/assert.h"
 
+#include "core/hle/kernel/client_port.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/server_port.h"
 #include "core/hle/kernel/thread.h"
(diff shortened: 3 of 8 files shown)


src/core/hle/kernel/server_port.h
```diff
@@ -0,0 +1,52 @@
...
+     */
+    static std::tuple<SharedPtr<ServerPort>, SharedPtr<ClientPort>> CreatePortPair(u32 max_sessions, std::string name = "UnknownPort");
+
+    std::string GetTypeName() const override { return "ServerPort"; }
+    std::string GetName() const override { return name; }
+
+    static const HandleType HANDLE_TYPE = HandleType::ServerPort;
+    HandleType GetHandleType() const override { return HANDLE_TYPE; }
+
+    std::string name;                           ///< Name of port (optional)
```
lioncash: This should likely be private if there's already a getter for it. Same with HANDLE_TYPE.

src/core/hle/kernel/client_port.h
```diff
@@ -0,0 +1,44 @@
...
+    std::string GetTypeName() const override { return "ClientPort"; }
+    std::string GetName() const override { return name; }
+
+    static const HandleType HANDLE_TYPE = HandleType::ClientPort;
+    HandleType GetHandleType() const override { return HANDLE_TYPE; }
+
+    SharedPtr<ServerPort> server_port;          ///< ServerPort associated with this client port.
+    u32 max_sessions;                           ///< Maximum number of simultaneous sessions the port can have
+    u32 active_sessions;                        ///< Number of currently open sessions to this port
+    std::string name;                           ///< Name of client port (optional)
```
lioncash: HANDLE_TYPE and name should likely be private, since they already have getters.

Subv:
@lioncash I'm following the same convention as the other kernel objects (See for example semaphore.h or mutex.h)

lioncash:
Alright, I can change everything over after this is merged, then.

Subv:
Sounds good

src/core/hle/kernel/client_port.h
yuriks: HANDLE_TYPE should be public since it's used by some templates (iirc). GetHandleType is there for polymorphism.

src/core/hle/kernel/client_port.h
```diff
@@ -0,0 +1,44 @@
...
+class ClientPort final : public Object {
+public:
+    /**
+     * Creates a client port.
+     * @param server_port ServerPort associated with this client port.
+     * @param max_sessions Maximum number of sessions to this port
+     * @param name Optional name of client port
+     * @return The created client port
+     */
+    static ResultVal<SharedPtr<ClientPort>> Create(SharedPtr<ServerPort> server_port, u32 max_sessions, std::string name = "Unknown");
```
yuriks: Since ports are always a pair, I think you should remove this Create and ServerPort's, and then make ServerPort a friend of this class so that CreatePair can construct them.

(Er, to summarize:
  - Remove ClientPort::Create
  - Remove ServerPort::Create
  - Add ServerPort as a friend of ClientPort
  - Construct ClientPort and ServerPort directly in CreatePortPair

src/core/hle/kernel/client_port.h
yuriks: Likewise, the getters here are just for polymorphism. The convention in the kernel code is to keep all fields except the constructor public.

src/core/hle/kernel/kernel.h
```diff
@@ -35,7 +35,7 @@ enum KernelHandle : Handle {
 
 enum class HandleType : u32 {
     Unknown         = 0,
-    Port            = 1,
+    ServerPort      = 1,
```
yuriks: Any particular reason to keep this up there? (Actually, why do we even have explicit values for this enum?)

src/core/hle/kernel/server_port.cpp
```diff
@@ -0,0 +1,43 @@
...
+    return MakeResult<SharedPtr<ServerPort>>(std::move(server_port));
+}
+
+bool ServerPort::ShouldWait() {
+    // If there are no pending sessions, we wait until a new one is added.
+    return pending_sessions.size() == 0;
+}
+
+void ServerPort::Acquire() {
+    ASSERT_MSG(!ShouldWait(), "object unavailable!");
```
yuriks: Add a note saying this is unimplemented.

src/core/hle/kernel/kernel.h
Subv: Excellent question, I'll probably make another PR to remove the explicit values

src/core/hle/kernel/server_port.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+std::tuple<SharedPtr<ServerPort>, SharedPtr<ClientPort>> ServerPort::CreatePortPair(u32 max_sessions, std::string name) {
+    auto server_port = ServerPort::Create(name + "Server").MoveFrom();
```
yuriks: I think you should add a _ before the Server and Client string, otherwise it'd look weird with stuff like "APT:UServer"

src/core/hle/kernel/server_port.cpp
Subv: It's not unimplemented, "acquiring" a ServerPort just does nothing, ServerSessions are accepted via svcAcceptSession

src/core/hle/kernel/server_port.cpp
```diff
@@ -0,0 +1,43 @@
...
+    auto client_port = ClientPort::Create(server_port, max_sessions, name + "Client").MoveFrom();
+
+    return std::make_tuple(server_port, client_port);
```
yuriks: std::move both (or just inline the expressions including MoveFrom())

src/core/hle/svc.cpp
```diff
@@ -833,6 +836,22 @@ static ResultCode CreateMemoryBlock(Handle* out_handle, u32 addr, u32 size, u32
...
+    // TODO(Subv): Implement named ports.
+    ASSERT_MSG(name == nullptr, "Named ports are currently unimplemented");
+
+    using Kernel::ServerPort;
+    using Kernel::ClientPort;
+    using Kernel::SharedPtr;
+
+    auto ports = ServerPort::CreatePortPair(max_sessions);
+    CASCADE_RESULT(*server_port, Kernel::g_handle_table.Create(std::move(std::get<SharedPtr<ServerPort>>(ports))));
+    CASCADE_RESULT(*client_port, Kernel::g_handle_table.Create(std::move(std::get<SharedPtr<ClientPort>>(ports))));
```
yuriks: This will leak the first handle if the creation of the second one fails.

src/core/hle/kernel/server_port.cpp
yuriks: Hm, I guess I don't remember how the waiting code works then. Shouldn't it still manage some kind of thread waitlist for when a new session arrives?

yuriks: Ah, right, it doesn't. Nevermind this then.

src/core/hle/svc.cpp
Subv: This is consistent with what the 3DS kernel does. I've fixed the write order of the handles and added a comment.
Should we fix this or stay true to the kernel quirks?

yuriks: This is actually fine. The first handle will still be set to the out variable in that case, so the caller can free it.

Subv:
Updated and rebased, I think this is good to go.

wwylele:
Code looks fine. I am not familiar with the mechanism.

src/core/hle/svc.cpp
```diff
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <map>
+#include <tuple>
```
bunnei: ?

src/core/hle/kernel/server_port.h
```diff
@@ -0,0 +1,45 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <string>
+
```
bunnei: Needs include of <tuple>

bunnei:
LGTM after the include's are fixed. I'll also wait for @yuriks to signoff before merging.

src/core/hle/kernel/kernel.h
yuriks: Can you just remove this line (keeping the numbering gap) and add a new entry in the proper place at the end?

yuriks:
[LGTM, Approval]

src/core/hle/svc.cpp
Subv: See https://github.com/citra-emu/citra/pull/1842/files#diff-4c2dcf02e3b93a7686c1362ff9f2ce5dR848

bunnei:
@Subv you lying sack of shit
<3

Subv:
Updated

PR #1846, opened 2016-05-23
OpenGL: Set shader_dirty on lighting changes

### LINKED ISSUE #1840: Yo-Kai Watch: Incorrect Hair Color
In game you pass a line and your character puts on shoes, when they put on shoes, the hair color becomes pink or really light, when they are supposed to stay brown.  5/21 build. The hair color is supposed to stay brown when the character crosses the line.

**CITRA** (correct hair color, no shoes)
capture

**CITRA** (incorrect color, with shoes on)

**3DS** (Image added by JayFoxRox, taken by JohnGodgames / pcmaker; thanks!)
capture by JohnGodgames

*Other example:*

**3DS** (left, correct coloring with shoes) vs **CITRA** (right, incorrect, with shoes)
capture

(Images reordered by JayFoxRox)

---

"Is it the hardware renderer?" It's really dark without HW renderer on, but it looks the hair doesn't change with HW renderer off.
nvidia gtx 970m
But it only happens when you have shoes on, with them off, the color is correct like in the real game.

(Posts moved here by JayFoxRox)

JayFoxRox:
This should fix #1840 .

Eventually we should get rid of that stupid notify handler thing - it's been a problem way too often.
Also I only looked at the lighting struct in Pica::Regs for this PR, it's possible that more parts of the shader are affected by stuff like this.

It would be nice if someone ( @tfarley ?) could confirm that none of these are used as shader uniform / all of them used in shader code.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -518,6 +518,58 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
         SyncLightPosition(7);
         break;
 
+    // Fragment lighting distance attenuation bias
+    case PICA_REG_INDEX_WORKAROUND(lighting.light[0].dist_atten_bias, 0x014A + 0 * 0x10):
+        SyncLightDistanceAttenuationBias(0);
+        break;
+    case PICA_REG_INDEX_WORKAROUND(lighting.light[1].dist_atten_bias, 0x014A + 1 * 0x10):
+        SyncLightDistanceAttenuationBias(1);
+        break;
+    case PICA_REG_INDEX_WORKAROUND(lighting.light[2].dist_atten_bias, 0x014A + 2 * 0x10):
+        SyncLightDistanceAttenuationBias(2);
+        break;
+    case PICA_REG_INDEX_WORKAROUND(lighting.light[3].dist_atten_bias, 0x014A + 3 * 0x10):
+        SyncLightDistanceAttenuationBias(3);
+        break;
+    case PICA_REG_INDEX_WORKAROUND(lighting.light[4].dist_atten_bias, 0x014A + 4 * 0x10):
+        SyncLightDistanceAttenuationBias(4);
+        break;
+    case PICA_REG_INDEX_WORKAROUND(lighting.light[5].dist_atten_bias, 0x014A + 5 * 0x10):
+        SyncLightDistanceAttenuationBias(5);
+        break;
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -380,6 +380,18 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
         SyncCombinerColor();
         break;
 
+    // Fragment lighting switches
+    case PICA_REG_INDEX(lighting.disable):
+    case PICA_REG_INDEX(lighting.num_lights):
+    case PICA_REG_INDEX(lighting.fresnel_selector):
```
tfarley: To make it a little more obvious that this affects a larger group of values/BitFields, you could name the union that contains fresnel_selector in pica.h, and refer to the union name instead here (3dbrew calls it "config0").

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -380,6 +380,18 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
...
+    case PICA_REG_INDEX(lighting.disable_lut_d0):
```
tfarley: Same here (3dbrew calls it "config1").

tfarley:
It seems like lighting.light[i].directional, lighting.light[i].dist_atten_bias, and lighting.light[i].dist_atten_scale should also set the shader dirty flag. The other cases that you added LGTM though (all are implicated in shader code gen).

JayFoxRox:
Good catch tfarley!

Though, shouldn't we turn those into uniforms?!
Those are float20 so imagine a game lerping them for some effects such as lights closing in..

I'll address both of these issues later or tomorrow

JayFoxRox:
This is mostly untested due to a lack of test cases. Please regression test and review

bunnei:
LGTM and tested distance attenuation, it's still appears to be working & correct.

Jobenblue:
Have you confirmed this fixes the hair issue because it still occurs on ade0889 along with new unrelated issues

Jobenblue:
Awesome! Thanks so much! Some new errors came up recently if you want to read more about them here: http://gbatemp.net/threads/citra-new-3ds-emulator.365154/page-219

JayFoxRox:
Unless this is related to this PR don't post here. It's confusing and sounds like regressions.
Use an issue on my repo to talk to me or get on IRC, send me a message on gbatemp, email me or w/e.

Your Yo-Kai issue was already being dealt with after I made wwylele aware of it. It's completly unrelated to this PR..

PR #1854, opened 2016-05-25
SVC::WaitSynchronizationN: Reschedule before checking objects (fix #1837)

wwylele:
~~This is not meant to be merged (soon), but as a reminder and I will do more investigation.~~
This was another approach (reschedule at the beginning) that came up along with #1837 (reschedule at the end) to solve the lock issue. But #1837 is causing [regression in Yo-Kai Watch](https://gbatemp.net/threads/citra-new-3ds-emulator.365154/page-219#post-6384147) now (Sorry!), and I find that the reschedule-at-the-beginning approach works fine (~~but I still don't know why~~ see the explanation below). ~~So I leave this here, looking for help to test if this also causes regressions in any games (comparing with the master before #1837)~~

Edit: This PR cannot be widely tested if it is just a PR (same as #1837). I think this should get merged into master first (this should be at least better than #1837, for #1837 is causing problem and this is not in theory), so that people can find problems (if any) by testing master.

### DIFF
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -6,7 +6,6 @@
 
 #include "common/logging/log.h"
 #include "common/microprofile.h"
-#include "common/scope_exit.h"
 #include "common/string_util.h"
 #include "common/symbols.h"
 
@@ -289,6 +288,16 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
     if (handle_count < 0)
         return ResultCode(ErrorDescription::OutOfRange, ErrorModule::OS, ErrorSummary::InvalidArgument, ErrorLevel::Usage);
 
+    std::vector<Kernel::SharedPtr<Kernel::WaitObject>> wait_objects;
+    wait_objects.reserve(handle_count);
+    for (int i = 0; i < handle_count; ++i) {
+        wait_objects.push_back(Kernel::g_handle_table.GetWaitObject(handles[i]));
+        if (wait_objects.back() == nullptr)
+            return ERR_INVALID_HANDLE;
+    }
+
+    HLE::Reschedule(__func__);
+
     // If 'handle_count' is non-zero, iterate through each handle and wait the current thread if
     // necessary
     if (handle_count != 0) {
@@ -297,12 +306,8 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
         Kernel::SharedPtr<Kernel::WaitObject> wait_object;
 
         for (int i = 0; i < handle_count; ++i) {
-            auto object = Kernel::g_handle_table.GetWaitObject(handles[i]);
-            if (object == nullptr)
-                return ERR_INVALID_HANDLE;
-
             // Check if the current thread should wait on this object...
-            if (object->ShouldWait()) {
+            if (wait_objects[i]->ShouldWait()) {
 
                 // Check we are waiting on all objects...
                 if (wait_all)
@@ -310,11 +315,11 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
                     wait_thread = true;
             } else {
                 // Do not wait on this object, check if this object should be selected...
-                if (!wait_all && (!selected || (wait_object == object && was_waiting))) {
+                if (!wait_all && (!selected || (wait_object == wait_objects[i] && was_waiting))) {
                     // Do not wait the thread
                     wait_thread = false;
                     handle_index = i;
-                    wait_object = object;
+                    wait_object = wait_objects[i];
                     selected = true;
                 }
             }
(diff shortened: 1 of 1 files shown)


src/core/hle/svc.cpp
```diff
@@ -289,6 +288,16 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
 
+    std::vector<Kernel::SharedPtr<Kernel::WaitObject>> wait_objects;
+    wait_objects.reserve(handle_count);
+    for (int i = 0; i < handle_count; ++i) {
+        wait_objects.push_back(Kernel::g_handle_table.GetWaitObject(handles[i]));
+        if(wait_objects.back() == nullptr)
+            return ERR_INVALID_HANDLE;
+    }
+
+    HLE::Reschedule("WaitSynchronizationN");
```
Subv: __func__

src/core/hle/svc.cpp
```diff
@@ -289,6 +288,16 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
     if (handle_count < 0)
         return ResultCode(ErrorDescription::OutOfRange, ErrorModule::OS, ErrorSummary::InvalidArgument, ErrorLevel::Usage);
 
+    std::vector<Kernel::SharedPtr<Kernel::WaitObject>> wait_objects;
+    wait_objects.reserve(handle_count);
+    for (int i = 0; i < handle_count; ++i) {
```
lioncash: This should be its own function, which can leverage boost::optional for the error case.

src/core/hle/svc.cpp
```diff
@@ -327,20 +332,12 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
         // Actually wait the current thread on each object if we decided to wait...
-        std::vector<SharedPtr<Kernel::WaitObject>> wait_objects;
-        wait_objects.reserve(handle_count);
-
-        for (int i = 0; i < handle_count; ++i) {
-            auto object = Kernel::g_handle_table.GetWaitObject(handles[i]);
-            object->AddWaitingThread(Kernel::GetCurrentThread());
-            wait_objects.push_back(object);
-        }
+        for (int i = 0; i < handle_count; ++i)
```
lioncash: This should likely be extracted to its own function

src/core/hle/svc.cpp
```diff
@@ -289,6 +288,16 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
     if (handle_count < 0)
         return ResultCode(ErrorDescription::OutOfRange, ErrorModule::OS, ErrorSummary::InvalidArgument, ErrorLevel::Usage);
 
+    std::vector<Kernel::SharedPtr<Kernel::WaitObject>> wait_objects;
+    wait_objects.reserve(handle_count);
+    for (int i = 0; i < handle_count; ++i) {
+        wait_objects.push_back(Kernel::g_handle_table.GetWaitObject(handles[i]));
+        if(wait_objects.back() == nullptr)
```
Subv: Whitespace after if

wwylele:
## The explanation on the regression on current master

### Normal procedure of "Thread waits for an event - Event signals - "Thread resumes":
 - the thread calls `WaitSynchronizationN`, and finds the event is not signaled;
 - the thread waits itself, registers to the event, and return `WaitSynchronizationN` with `HLE::RESULT_INVALID`.
 - then calls `Kernel::Reschedule` -> `SwitchContext` to run another thread;
 - later, the event is signaled, calls `WakeupAllWaitingThreads` -> `ResumeFromWait` on the thread -> push the thread to `ready_queue`;
 - then calls `Kernel::Reschedule` -> `SwitchContext`, where the thread is switched back, but with PC stepped back to call `WaitSynchronizationN` again - this time `WaitSynchronizationN` finds the event signaled and returns `RESULT_SUCCESS`.

### But sometimes, this can happen (for Yo-Kai Watch, it is the DSP thread)
 - the thread calls `WaitSynchronizationN`, and finds the event is not signaled;
 - the thread waits itself and registers to the event,
 - but before `WaitSynchronizationN` returns, `HLE::Reschedule` steps over the event, and the event is signaled. It calls `WakeupAllWaitingThreads` -> `ResumeFromWait` on the thread -> push the thread to `ready_queue`;
 - `WaitSynchronizationN` returns with `HLE::RESULT_INVALID`.
 -  then calls `Kernel::Reschedule`. This time, the ready thread in the queue is the thread itself, so it `// Don't bother switching to the same thread`, which means **the threads PC doesn't get stepped back to call `WaitSynchronizationN` again**!
 - The thread continues to run with `HLE::RESULT_INVALID` as the final return value. When it checks the return value, it `ThrowFatalError`.

So putting the `HLE::Reschedule` to the beginning of `WaitSynchronizationN` seems a good solution, because, this time, even if `HLE::Reschedule` signals any events, it works as if they are signaled before `WaitSynchronizationN` and any such collision won't happen. But should this approach still cause any problem, we would have to make some non-trivial fix in the scheduler. Making `Kernel::Reschedule` always calls `SwitchContext` is probably another solution, but I am not sure if this is correct.

src/core/hle/svc.cpp
wwylele: I just changed the code order in this PR to make a quick fix. We can do these refactor works with other similar functions later.

Subv:
Side note: What I want to see is a rewrite of our scheduler.

wwylele:
What's everybody's opinion on this? If this is not good, I will just revert #1837 first?

JohnGodgames:
This PR helps for most of games (random freezes at intro videos etc.)

wwylele:
Replaced by #1869

PR #1855, opened 2016-05-26
Memory: Added necessary headers and removed unnecessary header

MerryMage:
* `memory_setup.h` didn't need to include `memory.h` but did need to include `mmio.h`
* `vm_manager.cpp` was including `memory.h` via `memory_setup.h`


lioncash:
LGTM

PR #1857, opened 2016-05-27
common_funcs: Provide rotr and rotl for MSVC

MerryMage:
We currently provide them for non-MSVC platforms. Do the same for MSVC. (I'll be using them in the future.)


bunnei:
LGTM

PR #1867, opened 2016-05-28
srv: Update according 3dbrew

mailwl:


src/core/hle/service/srv.cpp
```diff
@@ -15,23 +15,86 @@ namespace SRV {
...
+static void EnableNotification(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
     // TODO(bunnei): Change to a semaphore once these have been implemented
     event_handle = Kernel::Event::Create(Kernel::ResetType::OneShot, "SRV:Event");
     event_handle->Clear();
 
-    cmd_buff[1] = 0; // No error
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    cmd_buff[2] = 0x20;
```
Subv: IPC::{Move/Copy}HandleDesc, depending on whether this moves or copies the following handle.

src/core/hle/service/srv.cpp
```diff
@@ -15,23 +15,86 @@ namespace SRV {
 
 static Kernel::SharedPtr<Kernel::Event> event_handle;
 
-static void Initialize(Service::Interface* self) {
+static const char* NotifDesc(u32 notif_id) {
```
Subv: I don't like this function at all

mailwl: i don't like it too, but have no idea how to convert most used 0x207/0x209 to human readable text

Subv: then don't convert it

src/core/hle/service/srv.cpp
mailwl: i am sorry, i have no  idea, what exactly function use instead of "magic number", 3dbrew lead to page weakly related to parameter value

mailwl: Ah, games never uses this value, may  i just leave as is?

wwylele: @mailwl 3dbrew well documented those translation descriptor [here](https://www.3dbrew.org/wiki/IPC). Just use them to match parameters for each individual service function (which is not well-documented)

src/core/hle/service/srv.cpp
```diff
@@ -15,23 +15,54 @@ namespace SRV {
...
+ *      1: ResultCode
+ *      2: Translation descriptor: 0x20
+ *      3: Handle to semaphore signaled on process notification
+ */
+static void EnableNotification(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
     // TODO(bunnei): Change to a semaphore once these have been implemented
     event_handle = Kernel::Event::Create(Kernel::ResetType::OneShot, "SRV:Event");
     event_handle->Clear();
```
JamePeng: cmd_buff[0] = IPC::MakeHeader(0x2, 0x1, 0x2);

src/core/hle/service/srv.cpp
```diff
@@ -15,23 +15,54 @@ namespace SRV {
...
+/**
+ * SRV::RegisterClient service function
+ *  Inputs:
+ *      0: 0x00010002
+ *      1: ProcessId Header (must be 0x20)
+ *  Outputs:
+ *      1: ResultCode
+ */
+static void RegisterClient(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
```
JamePeng: cmd_buff[0] = IPC::MakeHeader(0x1, 0x1, 0); //0x10040

src/core/hle/service/srv.cpp
```diff
@@ -15,23 +15,54 @@ namespace SRV {
...
 
-    cmd_buff[1] = 0; // No error
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    cmd_buff[2] = 0x20;
     cmd_buff[3] = Kernel::g_handle_table.Create(event_handle).MoveFrom();
+    LOG_WARNING(Service_SRV, "(STUBBED) called");
 }
 
+/**
+ * SRV::EnableNotification service function
```
JamePeng: SRV::GetServiceHandle service function

src/core/hle/service/srv.cpp
```diff
@@ -49,27 +80,86 @@ static void GetServiceHandle(Service::Interface* self) {
...
+ *      0: 0x00090040
+ *      1: Notification ID
+ *  Outputs:
+ *      1: ResultCode
+ */
+static void Subscribe(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 notif_id = cmd_buff[1];
+
```
JamePeng: cmd_buff[0] = IPC::MakeHeader(0x9, 0x1, 0); // 0x90040

src/core/hle/service/srv.cpp
```diff
@@ -49,27 +80,86 @@ static void GetServiceHandle(Service::Interface* self) {
...
+ *      0: 0x000A0040
+ *      1: Notification ID
+ *  Outputs:
+ *      1: ResultCode
+ */
+static void Unsubscribe(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 notif_id = cmd_buff[1];
+
```
JamePeng: cmd_buff[0] = IPC::MakeHeader(0xA, 0x1, 0); // 0xA0040

src/core/hle/service/srv.cpp
```diff
@@ -49,27 +80,86 @@ static void GetServiceHandle(Service::Interface* self) {
...
+ *      2: Flags (bit0: only fire if not fired, bit1: report errors)
+ *  Outputs:
+ *      1: ResultCode
+ */
+static void PublishToSubscriber(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 notif_id = cmd_buff[1];
+    u8 flags = cmd_buff[2] & 0xFF;
+
```
JamePeng: cmd_buff[0] = IPC::MakeHeader(0xC, 0x1, 0); // 0xC0040

src/core/hle/service/srv.cpp
```diff
@@ -49,27 +80,86 @@ static void GetServiceHandle(Service::Interface* self) {
...
+ *  Inputs:
+ *      0: 0x00090040
+ *      1: Notification ID
+ *  Outputs:
+ *      1: ResultCode
+ */
+static void Subscribe(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u32 notif_id = cmd_buff[1];
```
JamePeng: notification_id

mailwl:
addressed comments.
@JamePeng, maybe update 3dbrew with output header codes?

src/core/hle/service/srv.cpp
```diff
@@ -15,23 +15,64 @@ namespace SRV {
...
+ *  Outputs:
+ *      0: 0x00010040
+ *      1: ResultCode
+ */
+static void RegisterClient(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    cmd_buff[1] = 0; // No error
+    if (cmd_buff[1] != 0x20) {
+        cmd_buff[0] = IPC::MakeHeader(0x0, 0x1, 0); //0x40
```
Subv: Take the assign of cmd_buff[0] out of the branch.

src/core/hle/service/srv.cpp
```diff
@@ -15,23 +15,64 @@ namespace SRV {
...
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
     // TODO(bunnei): Change to a semaphore once these have been implemented
     event_handle = Kernel::Event::Create(Kernel::ResetType::OneShot, "SRV:Event");
     event_handle->Clear();
 
-    cmd_buff[1] = 0; // No error
+    cmd_buff[0] = IPC::MakeHeader(0x2, 0x1, 0x2); // 0x20042
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    cmd_buff[2] = IPC::CallingPidDesc();
```
Subv: This is weird, it doesn't make much sense for CallingPidDesc to be here, did you hwtest this response or otherwise reverse the srv module to find out if it's correct?

mailwl: no, i have no tests yet, i believe to 3dbrew, and this function most close to expected result.
i suggest leave it as is for now, i'm working on proper cecd stubbing now, and write test for this, too (games never check this value, just 1 and 3 means)

Subv: The games don't check it but the kernel uses it for parameter translation. We need this value to be absolutely correct.

mailwl: shame on me, I can't get work NTR CFW debugger running, and can't get srv: binarie to test
maybe @JamePeng may say, which value set to second cmd_buff (btw, looks like this value need to be provided for all functions, which return event)

wwylele: @mailwl For function returning handles (for event, semaphore, etc.) the translation descriptor before the handle _should_ be `MoveHanldeDesc/CopyHandleDesc` according to the IPC convention. 
@neobrain Is this value hwtested? (I saw you added it to 3dbrew)

mailwl: @wwylele i really don't understand, how to to use Move/CopyHandleDesc to get 0x20, my high education pass it

wwylele: @mailwl that's why I am asking neobrain if that's truly 0x20.

wwylele: @mailwl From IRC #citra:

> <wwylele> neobrain: https://www.3dbrew.org/wiki/SRV:EnableNotification <- is that "Translation descriptor:0x20" true? It's weird to be a PID descriptor there
> <neobrain> fyi, I didn't really add any new information in my change
> <neobrain> oh, or maybe I did
> <neobrain> https://www.3dbrew.org/w/index.php?title=SRV:EnableNotification&diff=17442&oldid=16196
> <wwylele> thanks :)
> <neobrain> it's probably just a "transfer handle without closing" descriptor fwiw
> <wwylele> yeah I guess so.

bunnei:
LGTM

Issue #1861, opened 2016-05-28
Phineas & Ferb: Quest for Cool Stuff - Hangs at opening screen


JohnGodgames:
[LOG](http://pasted.co/86ffb991)

![phineas](https://cloud.githubusercontent.com/assets/11453507/15625553/a9a64eaa-24aa-11e6-9272-9111a5e83273.JPG)

mailwl:
it second game i saw, which panic after WaitSynchronizationN on GSP::GPU event. I think @wwylele should look at this game
```
.text:001957C4                 LDR             R2, [R4,#0x2C]
.text:001957C8                 MOV             R0, #0xFFFFFFFF
.text:001957CC                 MOV             R1, R0
.text:001957D0                 STMEA           SP, {R0-R2}
.text:001957D4                 MOV             R3, #0
.text:001957D8                 MOV             R2, #1
.text:001957DC                 ADD             R1, SP, #0x30+var_28
.text:001957E0                 ADD             R0, SP, #0x30+var_24
.text:001957E4                 BL              WaitSynchronizationN
.text:001957E8                 MOVS            R1, R0,LSR#31
.text:001957EC                 NOP
.text:001957F0                 BLNE            Panic   ; <<< panic here
```

wwylele:
#1854 doesn't help?

mailwl:
no.
i don't understand why WaitSynchN returns error, if GSP::GPU event for sure exists.
looks like this code is not good:
https://github.com/wwylele/citra/blob/18a5043b7912268f7fd43358e0077ba691fc9d2d/src/core/hle/svc.cpp#L347-L348

wwylele:
@mailwl does it works before #1837 ?

mailwl:
no, the problem not in your PRs. 
I mean, if you look at WaitSynchronizationN now, the game good thing to test

wwylele:
@mailwl got it. Indeed an interesting case

wwylele:
@mailwl @JohnGodgames Try this (I don't have the game so it's just my random guess) : comment out [this two lines](https://github.com/citra-emu/citra/blob/a2c84d5779a80c61adbf07cfce6c37b0952fea84/src/core/hle/kernel/thread.cpp#L546-547) 
```c++
// Don't bother switching to the same thread
if (next == cur)
    return;
```

mailwl:
wow, it's worked!

JohnGodgames:
[...]

Issue #1866, opened 2016-05-28
True Remembrance


mailwl:
regression from #1161, now the game throws exception

```
[  13.657607] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:108: Fatal error!
[  13.657817] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:116: PID: 0x00000000_0x00000000
[  13.657930] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:117: REV: 42918
[  13.658047] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:118: AID: 0x00000000_0x00000000
[  13.658155] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:119: ADR: 0x001AF928
[  13.658251] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:121: RSL: 0xFFFFFFFF
[  13.658346] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:122:   Level: 31
[  13.658439] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:123:   Summary: 63
[  13.658533] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:124:   Module: 255
[  13.658627] Service.ERR <Critical> core\hle\service\err_f.cpp:ERR_F::ThrowFatalError:125:   Desc: 1023
```

mailwl:
ah, sorry, it's same as #1861 (and worked again with commented code)
so, there is regression between this two issues.

PR #1868, opened 2016-05-28
WaitSynchronization: Perform PC modification immediately instead of at thread wakeup

MerryMage:
A possible alternative to #1854 that hopefully makes more sense.

* Perform the PC modification *immediately* so shenanigans can't occur.
* Also added a commit that fixes a bug in `Thread::SetWaitSynchronization{Output,Result}()` I noticed as I was reading related code.

I haven't done testing to see if it fixes the same issues as #1854 does as I don't own many of the games. Ping @wwylele et al.

Thanks to @wwylele for their [investigative work](https://github.com/citra-emu/citra/pull/1854#issuecomment-221836087).

### DIFF
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -579,11 +579,19 @@ void Reschedule() {
 }
 
 void Thread::SetWaitSynchronizationResult(ResultCode result) {
-    context.cpu_registers[0] = result.raw;
+    if (this == GetCurrentThread()) {
+        Core::g_app_core->SetReg(0, result.raw);
+    } else {
+        context.cpu_registers[0] = result.raw;
+    }
 }
 
 void Thread::SetWaitSynchronizationOutput(s32 output) {
-    context.cpu_registers[1] = output;
+    if (this == GetCurrentThread()) {
+        Core::g_app_core->SetReg(1, output);
+    } else {
+        context.cpu_registers[1] = output;
+    }
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -210,17 +210,6 @@ static void SwitchContext(Thread* new_thread) {
 
         current_thread = new_thread;
 
-        // If the thread was waited by a svcWaitSynch call, step back PC by one instruction to rerun
-        // the SVC when the thread wakes up. This is necessary to ensure that the thread can acquire
-        // the requested wait object(s) before continuing.
-        if (new_thread->waitsynch_waited) {
-            // CPSR flag indicates CPU mode
-            bool thumb_mode = (new_thread->context.cpsr & TBIT) != 0;
-
-            // SVC instruction is 2 bytes for THUMB, 4 bytes for ARM
-            new_thread->context.pc -= thumb_mode ? 2 : 4;
-        }
-
         // Clean up the thread's wait_objects, they'll be restored if needed during
         // the svcWaitSynchronization call
         for (size_t i = 0; i < new_thread->wait_objects.size(); ++i) {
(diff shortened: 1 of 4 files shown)


MerryMage:
This still feels a bit *meh* to be and a scheduler rewrite would be most welcome.

MerryMage:
Update: Removed the revert commit.

src/core/hle/kernel/thread.h
```diff
@@ -203,7 +203,8 @@ Thread* GetCurrentThread();
 void WaitCurrentThread_Sleep();
 
 /**
- * Waits the current thread from a WaitSynchronization call
+ * Waits the current thread from a WaitSynchronization call.
+ * Note: Also steps back the PC one instruction so the previous SVC will be called again on thread wakeup.
```
lioncash: Not a big deal, but notes can use the `@note` tag for better markup sectioning.

MerryMage:
Updated with a commit to fix those two. Suggestion: make `context` private.

MerryMage:
Addressed @Lioncash's comment.

MerryMage:
Addressed @wwylele's comments from IRC:

    21:20 < wwylele> OK found what cause that PriorityBoostStarvedThreads crashes...
    21:21 < wwylele> It's not a bug in #1868, but a bug revealed by #1868
    21:22 < wwylele> The problem is here :https://github.com/citra-emu/citra/blob/master/src/core/hle/kernel/thread.cpp#L546 (again)
    21:23 < wwylele> even if next==cur, the thread can be in READY state
    21:24 < wwylele> (for example, the thread wait on an object, the object imediately wake up the thread, causing current_thread=next_thread but in READY state)
    21:24 < wwylele> in this case, the thread should be set in RUN state
    21:26 < wwylele> changing it to  `if(next==cur){if(next)next->state=THREADSTATUS_RUNNING; return}` would be fine

    21:47 < wwylele> merry: https://github.com/citra-emu/citra/pull/1868/files#diff-026ae30e0e7112cb315ff4a38e14f50fR316 GetPC() should be GetPC()&0xFFFFFFFE probably?
    21:48 < wwylele> oh nvm
    21:48 <@merry> I don't think it's possible for a misaligned pc to occur in that case.
    21:50 <@merry> it may be possible for it to occur in the other call if my memory of skyeye is accurate.

Subv:
Suggestion: Set Kernel::g_current_thread to nullptr when the current thread sleeps.

src/core/hle/kernel/thread.cpp
```diff
@@ -301,6 +309,19 @@ static void ThreadWakeupCallback(u64 thread_handle, int cycles_late) {
 
     thread->waitsynch_waited = false;
 
+    if (thread->status == THREADSTATUS_WAIT_SYNCH) {
+        // Step forward PC by one instruction to NOT rerun the SVC when the thread wakes up.
+        if (thread == GetCurrentThread()) {
+            bool thumb_mode = (Core::g_app_core->GetCPSR() & TBIT) != 0;
+            AssertOnWaitSynchronizationInstruction(thumb_mode, Core::g_app_core->GetPC());
```
wwylele: The case `thread == GetCurrentThread()` can happen in (probably only in) the following call stack:
```
 - Kernel::ThreadWakeupCallback
 - CoreTiming::ProcessFifoWaitEvents()
 - CoreTiming::Advance()
 - ARM_DynCom::AddTicks()
 - HLE::Reschedule()
 - SVC::WaitSynchronizationN()
 - SVC::CallSVC()
 - InterpreterMainLoop()
 - ARM_DynCom::ExecuteInstructions()
```

Note that in `InterpreterMainLoop`, it do `cpu->Reg[15] += cpu->GetInstructionSize();` after `CallSVC`, so the PC this assertion observes is an instruction behind the expected PC.

wwylele: A solution that I found is to also merge #1854 - Do `HLE::Reschedule` before the thread can register itself to wake up, so that the case `thread == GetCurrentThread()` will never happens (hopefully)

MerryMage: > PC this assertion observes is one instruction behind the expected PC
Indeed. The PC the assertion observes is not one would expect! (And is correct for the assertion.)

It sounds rare but I've had this case happen a lot, especially only one thread is free.

> Do `HLE::Reschedule` before the thread can register itself to wake up

I don't think that by itself works. `HLE::Reschedule` doesn't actually do the rescheduling, it just schedules for it to happen, so `thread == GetCurrentThread()` can still occur.

MerryMage:
@Subv Good idea.

src/core/hle/kernel/thread.cpp
MerryMage: @wwylele: That situation with the exact same callstack can still occur even with #1854. 

When the service call is called again for the **second time**, the thread has already registered itself to wakeup. The same callstack occurs.

Thus the code should stay (until someone comes up with an alternative way of advancing time...)

mailwl:
this PR fixes #1861 and #1866 (same problem as #1861)

MerryMage:
Closed in favour of #1869.

PR #1869, opened 2016-05-29
Switch context to the same thread if necessary

### LINKED ISSUE #1856: Video and Audio Regression
So I am trying out Yo-Kai Watch (it's the only game I have that works) and as recently as e054e55 (does not happen in 51ee2d2), a few new issues have arisen, seemingly stemming from an intro video/ special scenes. If I let the intro play, all sound cuts off and when I press continue at the title screen (also no audio in menu if cutscene plays), it gets stuck at a black screen. If I skip the intro video quickly enough, sound still works in the menu and I can go into the game, but when I leave the house, triggering a special "welcome to the world of yo-kai" screen (which occurs once per game), the game gets stuck on a black screen and I can't do anything, if "welcome to the world of yo-kai" has already happened, it is possible to leave the house but there is an issue with the audio looping. So basically, e054e55 and newer has issues with the intro video/special scenes/audio, leading to either no audio and cannot get in game, or has audio but cannot leave the house if "welcome to the world of yo-kai" pops up. Also happens the same way if audio is set to null.

Purple log which was not present before during intro video.
capture

Letting intro play, leading to no audio, after hitting continue
capture

Ingame with audio
capture

Ingame with audio, causes audio to stop, game still runs.
capture

wwylele:
_Yet another alternative to #1868 and #1854_
#1854 doesn't resolve all related issues (such as #1861), ~~I'd like to close it later~~ closed.
I like the thought of #1868, but it seems that it needs to fix too much stuff, and more issues tend to occur when more changes are made.
So I want a simple, temporary fix to make citra works for most cases, before we rewrite the scheduler (if it is going to be).
The common issue of #1856 and #1861 is that the scheduler doesn't bother switching to the same thread. So the simple fix here is to switch to the same thread if needed (when the thread was waiting on objects), so that it can perform PC modification (causes of `ThrowFatelError` in #1856 and #1861), ~~pop from ready queue~~ change state to RUNNING (= mark the thread popped from ready queue, was an issue in #1868), etc.

@mailwl @JohnGodgames please help test, thanks! (only on this PR, without #1868 or #1854)

Edit:
Fixes  #1856, #1861 and #1866

### DIFF
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -542,8 +542,12 @@ void Reschedule() {
 
     HLE::DoneRescheduling();
 
-    // Don't bother switching to the same thread
-    if (next == cur)
+    // Don't bother switching to the same thread.
+    // But if the thread was waiting on objects, we still need to switch it
+    // to perform PC modification, change state to RUNNING, etc.
+    // This occurs in the case when an object the thread is waiting on immediately wakes up
+    // the current thread before Reschedule() is called.
+    if (next == cur && (next == nullptr || next->waitsynch_waited == false))
         return;
 
     if (cur && next) {


mailwl:
both  #1861 and #1866 now works. other random games still works too (no regression).
i will check on CecToolBox later (this tool required a loooot unimplemented functions from cecd:s, but crash on GPU:GSP waiting).

MerryMage:
As mentioned, an eventual scheduler rewrite (or at least a cleanup) would probably be necessary. Keeping track of the invariants in my head was fun.

src/core/hle/kernel/thread.cpp
```diff
@@ -542,8 +542,12 @@ void Reschedule() {
...
     HLE::DoneRescheduling();
 
-    // Don't bother switching to the same thread
-    if (next == cur)
+    // Don't bother switching to the same thread.
+    // But if the thread was waiting on objects, we still need to switch it
+    // to perform PC modification, change state to RUNNING, etc.
+    // This occurs in the case when an object the thread is waiting on immediately wakes up
+    // the current thread before Reschedule() is called.
+    if (next == cur && (next == nullptr || next -> waitsynch_waited == false))
```
lioncash: `next->waitsynch_waited == false`

wwylele:
addressed

MerryMage:
LGTM

JamePeng:
LGTM

wwylele:
@citra-emu please clean up...

yuriks:
LGTM, but I don't really trust myself with the scheduler code anymore.

wwylele:
@bunnei ping

PR #1871, opened 2016-06-01
AddFstEntriesToGameList - prevent loading a directory as a file

LFsWang:
truth table
image

before
old

after
new

### DIFF
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -130,9 +130,7 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
         if (stop_processing)
             return false; // Breaks the callback loop.
 
-        if (recursion > 0 && FileUtil::IsDirectory(physical_name)) {
-            AddFstEntriesToGameList(physical_name, recursion - 1);
-        } else {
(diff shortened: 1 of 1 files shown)


src/citra_qt/game_list.cpp
```diff
@@ -145,6 +143,8 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
                 new GameListItem(QString::fromStdString(Loader::GetFileTypeString(loader->GetFileType()))),
                 new GameListItemSize(FileUtil::GetSize(physical_name)),
             });
+        } else if (recursion > 0) {
```
LittleWhite-tb: I would add a small comment here to specify that physical_name is a directory here. Maybe I am over commenting... so do as you like.

LittleWhite-tb:
Everything fine to me :+1:

src/citra_qt/game_list.cpp
lioncash: The if statement kind of makes it evident that it's a directory at this point.

LittleWhite-tb: Sure, maybe I'm a bit sleepy :)

bunnei:
Lgtm

PR #1789, opened 2016-05-13
Refactor input mapping & implement circle pad modifier

wwylele:
The procedure to map inputs on current master is like this:

Frontend keyboard event -> `EmuWindow::KeyPressed/KeyReleased` -> `KeyMap::GetPadKey` -> set bit field in `PadState` for each field.

This procedure assumes that each physical keyboard key one-to-one maps to a 3DS pad button, which is not true for analog input like circle pad. So I refactored it to:

Frontend keyboard event -> `KeyMap::PressKey/ReleaseKey` -> `EmuWindow::ButtonPressed/ButtonReleased/CirclePadUpdated/...` -> set button state and circle pad state

All keyboard inputs are grouped into two category: directly mapped keys and indirectly mapped ones. For directly mapped ones, `KeyMap::PressKey/ReleaseKey` will directly call `EmuWindow::ButtonPressed/ButtonReleased` with mapped buttons. For indirectly mapped ones, `KeyMap` will store the key state, translate them into analog input and call `EmuWindow::CirclePadUpdated/...` 

Now `EmuWinodw` is served as a pure common interface between input devices and the HID module. If we support real analog inputs in the future, they can be sent to `EmuWindow::CirclePadUpdated` without going through `KeyMap`.

Currently only circle-pad-related input are implemented as indirectly mapped. Later, gyroscope and accelerometer input can also go this way .Note that I leave C-stick as directly mapped inputs because I am not able to test its analog behaviour. This can be fixed in the future following the circle pad method.

About how HID module translate circle pad input to direction, see my previous closed PR #1778 

----
I also implemented the circle pad modifier in this PR (instead of another PR) to show how `KeyMap` works. The modifier is default to key D, and the default modified radius is 50%

PR #1873, opened 2016-06-02
Remove unused and bitrotted "controller config" files

archshift:
There's no use keeping this around IMO, if we want an actual useful controller config UI we'll resurrect #531 or #1119.


Subv:
Kill it with fire.

makotech222:
Just fyi, i have a controller config UI waiting in the wings. Need for #1789 to get accepted first though.

MerryMage:
@makotech222 Does that block this PR?

makotech222:
@MerryMage Nope

bunnei:
LGTM

Issue #1629, opened 2016-04-02
Games that crash when Shader JIT is enabled


JohnGodgames:
Tested on recent master and https://github.com/bunnei/citra/tree/refactor-shader-jit

Logs are from https://github.com/bunnei/citra/commit/b7bbc93dff8f7846410d98e16b12a71cc924d5ec

**Exstetra**

[LOG](http://pastebin.com/smB2tfa6)

**Sayonara Umihara Kawase**

[LOG](http://pastebin.com/BVH58Gm7)

and propably more (I will add next if I find them)

Subv:
Does this still happen?

JohnGodgames:
Yeah still with both of these...

Dragios:
#2226 Suffers from the same issue but only when loading the game itself result in crash but not during booting.

Subv:
These games are trying to perform ````movaps xmm1, [7FF66FC3A099h]````, but the manual entry for movaps says that it must operate on 16-byte aligned addresses.

yuriks:
We're not scaling the loop index by 16 before indexing into the register. (How Does Anything Even Work™-worthy for sure) The fix is just to multiply LOOPCOUNT_REG and LOOPINC by 16 when they're assigned like we do with ADDROFFS_REG_0/1.

JayFoxRox:
Proposed solution https://github.com/JayFoxRox/citra/pull/34 (untested still)

mailwl:
Cut the Rope works

PR #1874, opened 2016-06-02
[RFC] citra-qt: Print stack trace

MerryMage:
This PR implements printing of a stack trace to the console when something bad happens. On Windows and OS X it also displays a modal message box to the user.

*Example of an `ASSERT_MSG` being caught.*

## Motivation

To hopefully get more useful reports from users rather than just "it crashes".

Could *possibly* be somehow extended to dump logs to a file in that case and/or integrate with an issue tracking system.

## Known isuses

* It's in `citra-qt` because I was originally going to use `QMessageBox`, but that doesn't work at all. Perhaps I should move it to `common`.
* Untested on ~~OS X and~~  Linux at the moment.
* Consider [localisation](https://github.com/citra-emu/citra/pull/1874#discussion_r65587189) of user-facing strings.

### DIFF
--- a/src/citra_qt/crash_handler.cpp
+++ b/src/citra_qt/crash_handler.cpp
@@ -11,7 +11,13 @@
 // the crash handler is called, we're going to have to use reasonably
 // low-level methods to display things to the user.
 //
-// An advisable strategy is to get things onto the console first before trying to do fancy GUI stuff.
+// Keep it as simple as possible.
+//
+// An advisable strategy is to get things onto the console first before
+// trying to do fancy GUI stuff. TODO(merry): An even better strategy would
+// be to not do any processing at all and punt to a watchdog process, because
+// ideally one cannot assume memory allocation works (what if the crash was
+// in the allocator?) or stack allocate (what if rsp isn't in a valid state?).
 
 #ifdef _WIN32
 
@@ -19,11 +25,12 @@
 
 #include <windows.h> // Must include this first.
 #include <dbghelp.h>
-#include <tchar.h>
+
+#include "common/string_util.h"
 
 namespace CrashHandler {
 
-static LONG WINAPI MyUnhandledExceptionFilter(struct _EXCEPTION_POINTERS*);
+static LONG WINAPI MyUnhandledExceptionFilter(_EXCEPTION_POINTERS*);
 static std::string GetStackTrace(CONTEXT& c);
 
 void Register() {
@@ -36,11 +43,11 @@ void Register() {
  * @param ep The exception pointer containing exception information.
  * @return See Microsoft's documentation on SetUnhandledExceptionFilter for possible return values.
  */
-static LONG WINAPI MyUnhandledExceptionFilter(struct _EXCEPTION_POINTERS* ep) {
+static LONG WINAPI MyUnhandledExceptionFilter(_EXCEPTION_POINTERS* ep) {
     std::string stack_trace = GetStackTrace(*ep->ContextRecord);
 
-    std::string detail = std::string() +
-        "Version: " + Common::g_scm_branch + ":" + Common::g_scm_desc + "\n"
+    std::string detail =
+        std::string("Version: ") + Common::g_scm_branch + ":" + Common::g_scm_desc + "\n"
         "Commit: " + Common::g_scm_rev + "\n"
         "Stack Trace:\n" + stack_trace;
 
@@ -50,25 +57,25 @@ static LONG WINAPI MyUnhandledExceptionFilter(struct _EXCEPTION_POINTERS* ep) {
 
     std::string title = "Citra: Caught Unhandled Exception";
 
-    std::string message = std::string() +
-        "Press Ctrl+C to copy text\n" +
-        "Please also take a copy of the console window\n" +
-        "\n" +
-        detail;
+    std::string message = std::string("Press Ctrl+C to copy text\n"
+        "Please also take a copy of the console window\n\n") + detail;
 
-    // QMessageBox does not work and when it does it isn't happy. We need to use something lower-level.
+    // QMessageBox is not guaranteed to work here since:
+    // 1. We're not guaranteed to be in the GUI thread
+    // 2. Qt is not guaranteed to be in a valid state
+    // We need to use something lower-level.
     MessageBoxA(nullptr, message.c_str(), title.c_str(), MB_ICONSTOP);
-    FatalAppExit(0, _T("Terminating application"));
+    FatalAppExitA(0, "Terminating application");
 
     return EXCEPTION_CONTINUE_SEARCH;
 }
 
 /**
  * This function walks the stack of the current thread using StackWalk64.
- * @param c The context record that contains the information on the stack of interest.
+ * @param ctx The context record that contains the information on the stack of interest.
  * @return A string containing a human-readable stack trace.
  */
-static std::string GetStackTrace(CONTEXT& c) {
+static std::string GetStackTrace(CONTEXT& ctx) {
     HANDLE process = GetCurrentProcess();
     HANDLE thread = GetCurrentThread();
 
@@ -84,7 +91,7 @@ static std::string GetStackTrace(CONTEXT& c) {
         SCOPE_EXIT({ free(symbol); });
 
         // Actually get symbol info.
-        DWORD64 symbol_displacement; // The number of bytes return_address is offset from the start of the function.
+        DWORD64 symbol_displacement; // The offset of return_address from the start of the function.
         SymGetSymFromAddr64(process, return_address, &symbol_displacement, symbol);
 
         // Get undecorated name.
@@ -92,12 +99,12 @@ static std::string GetStackTrace(CONTEXT& c) {
         UnDecorateSymbolName(symbol->Name, &undecorated_name[0], SYMBOL_NAME_SIZE, UNDNAME_COMPLETE);
 
         // Get source code line information.
-        DWORD line_displacement = 0; // The number of bytes return_address is offset from the first instruction of this line.
-        IMAGEHLP_LINE64 line = { 0 };
+        DWORD line_displacement = 0; // The offset of return_address from the first instruction of this line.
+        IMAGEHLP_LINE64 line = {};
         line.SizeOfStruct = sizeof(IMAGEHLP_LINE64);
         SymGetLineFromAddr64(process, return_address, &line_displacement, &line);
 
-        // Remove unnecessary path information before the "\\src\\" directory.s
+        // Remove unnecessary path information before the "\\src\\" directory.
         std::string file_name = "(null)";
         if (line.FileName) {
             file_name = line.FileName;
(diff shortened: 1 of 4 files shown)


src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+        // Format string
+        std::string ret;
+        ret.resize(1024);
+        size_t ret_size = snprintf(&ret[0], ret.size(), "[%llx] %s+0x%llx (%s:%i)\n", return_address, undecorated_name, symbol_displacement, file_name.c_str(), line.LineNumber);
+        ret.resize(ret_size);
+        return ret;
+    };
+
+    // Initialise symbols
+    if (SymInitialize(process, NULL, TRUE) == FALSE) {
```
lioncash: nullptr

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+    s.AddrStack.Mode = AddrModeFlat;
+    s.AddrFrame.Offset = c.Rbp;
+    s.AddrFrame.Mode = AddrModeFlat;
+
+    // Return value
+    std::string stack_trace;
+
+    // Walk the stack
+    do {
+        if (!StackWalk64(IMAGE_FILE_MACHINE_AMD64, process, thread, &s, &c, NULL, SymFunctionTableAccess64, SymGetModuleBase64, NULL)) {
```
lioncash: nullptr

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+    std::string message = std::string() +
+        "Version: " + Common::g_scm_branch + "-" + Common::g_scm_desc + "\n" +
+        "Commit: " + Common::g_scm_rev + "\n" +
+        "Stack Trace:\n" + stack_trace;
+
+    OSXMessageBox(title, message);
+#endif
+
+    exit(-1);
```
lioncash:
```cpp
exit(EXIT_FAILURE);
```

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+    for (size_t i = 0; i < num_frames; i++) {
+        fprintf(stderr, "%s\n", symbols[i]);
+        ret += symbols[i];
+        ret += "\n";
+    }
+    return ret;
+}
+
+#ifdef __APPLE__
+static void OSXMessageBox(std::string title, std::string message) {
```
lioncash: const references?

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+        "Commit: " + Common::g_scm_rev + "\n"
+        "Stack Trace:\n" + stack_trace;
+
+    // Ensure we have a log of everything in the console first.
+    fprintf(stderr, "Unhandled Exception:\n%s", detail.c_str());
+    fflush(stderr);
+
+    std::string title = "Citra: Caught Unhandled Exception";
+
+    std::string message = std::string() +
```
lioncash:
```cpp
std::string message = std::string(
        "Press Ctrl+C to copy text\n"
        "Please also take a copy of the console window\n\n") +
        detail;
```

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+        DWORD64 symbol_displacement; // The number of bytes return_address is offset from the start of the function.
+        SymGetSymFromAddr64(process, return_address, &symbol_displacement, symbol);
+
+        // Get undecorated name.
+        char undecorated_name[SYMBOL_NAME_SIZE + 1];
+        UnDecorateSymbolName(symbol->Name, &undecorated_name[0], SYMBOL_NAME_SIZE, UNDNAME_COMPLETE);
+
+        // Get source code line information.
+        DWORD line_displacement = 0; // The number of bytes return_address is offset from the first instruction of this line.
+        IMAGEHLP_LINE64 line = { 0 };
```
lioncash: the zero here isn't necessary.

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+    SetUnhandledExceptionFilter(MyUnhandledExceptionFilter);
+}
+
+/**
+ * This function is called by the operating system when an unhandled exception occurs.
+ * This includes things like debug breakpoints when not connected to a debugger.
+ * @param ep The exception pointer containing exception information.
+ * @return See Microsoft's documentation on SetUnhandledExceptionFilter for possible return values.
+ */
+static LONG WINAPI MyUnhandledExceptionFilter(struct _EXCEPTION_POINTERS* ep) {
```
lioncash: `struct` isn't necessary here.

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+    return EXCEPTION_CONTINUE_SEARCH;
+}
+
+/**
+ * This function walks the stack of the current thread using StackWalk64.
+ * @param c The context record that contains the information on the stack of interest.
+ * @return A string containing a human-readable stack trace.
+ */
+static std::string GetStackTrace(CONTEXT& c) {
```
lioncash: `ctx` is likely a better shorthand here.

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+        std::string file_name = "(null)";
+        if (line.FileName) {
+            file_name = line.FileName;
+            size_t found = file_name.find("\\src\\");
+            if (found != std::string::npos)
+                file_name = file_name.substr(found + 1);
+        }
+
+        // Format string
+        std::string ret;
```
lioncash: Actually, this can just use StringFromFormat in string_util.h and just be:

```cpp
return Common::StringFromFormat("[%llx] %s+0x%llx (%s:%i)\n", return_address, undecorated_name, symbol_displacement, file_name.c_str(), line.LineNumber);
```

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+    HANDLE process = GetCurrentProcess();
+    HANDLE thread = GetCurrentThread();
+
+    // This function generates a single line of the stack trace.
+    // `return_address` is a return address found on the stack.
+    auto get_symbol_info = [&process](DWORD64 return_address) -> std::string {
+        constexpr size_t SYMBOL_NAME_SIZE = 512; // arbitrary value
+
+        // Allocate space for symbol info.
+        IMAGEHLP_SYMBOL64* symbol = static_cast<IMAGEHLP_SYMBOL64*>(calloc(sizeof(IMAGEHLP_SYMBOL64) + SYMBOL_NAME_SIZE * sizeof(char), 1));
```
lioncash: why `calloc`?

MerryMage: `calloc` instead of `malloc` because I want to zero the memory. The Win32 API (still called Win32 on x64) often requires unused members to be set to zero.

Not stack-allocated because I don't think it's possible to do so.

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+    // Ensure we have a log of everything in the console first.
+    fprintf(stderr, "Unhandled Exception:\n%s", detail.c_str());
+    fflush(stderr);
+
+    std::string title = "Citra: Caught Unhandled Exception";
+
+    std::string message = std::string() +
+        "Press Ctrl+C to copy text\n" +
+        "Please also take a copy of the console window\n" +
```
lioncash: This might want to be phrased as taking a screenshot, for better clarity.

lioncash: (it's somewhat unfortunate this can't be/isn't hooked by Qt, since this can't be localized in its current state)

src/citra_qt/crash_handler.cpp
lioncash: Hmm, looked it up and yeah, it's MS being silly/quirky with its API (technically it can be done on the stack with an instance and a trailing buffer used for the name, but it's relying on memory locations, so it's not worth it).

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,270 @@
...
+/**
+ * This function is called by the operating system when an unhandled exception occurs.
+ * This includes things like debug breakpoints when not connected to a debugger.
+ * @param ep The exception pointer containing exception information.
+ * @return See Microsoft's documentation on SetUnhandledExceptionFilter for possible return values.
+ */
+static LONG WINAPI MyUnhandledExceptionFilter(struct _EXCEPTION_POINTERS* ep) {
+    std::string stack_trace = GetStackTrace(*ep->ContextRecord);
+
+    std::string detail = std::string() +
```
Subv: Why is the std::string() needed?

MerryMage: Concatenation. I could do `std::string("Version: ")` instead if preferable.

src/citra_qt/crash_handler.cpp
MerryMage: Translated strings could be passed in if they were arguments of `Register()`. Thoughts?

MerryMage:
Addressed above comments except @lioncash's [comment on localisation](https://github.com/citra-emu/citra/pull/1874#discussion_r65587189).

Tested on OS X and added a screenshot.

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+static std::string GetStackTrace() {
+    constexpr size_t MAX_NUM_FRAMES = 64;
+    void* return_addresses[MAX_NUM_FRAMES + 1];
+
+    size_t num_frames = backtrace(return_addresses, MAX_NUM_FRAMES);
+    if (num_frames == 0)
+        return "";
+
+    char** symbols = backtrace_symbols(return_addresses, num_frames);
```
lioncash: This needs to be freed before the function ends.

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+#include <windows.h> // Must include this first.
+#include <dbghelp.h>
+
+#include "common/string_util.h"
+
+namespace CrashHandler {
+
+static LONG WINAPI MyUnhandledExceptionFilter(_EXCEPTION_POINTERS*);
+static std::string GetStackTrace(CONTEXT& c);
```
lioncash: the `c` here can be omitted for consistency with the filter function above it.

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    size_t num_frames = backtrace(return_addresses, MAX_NUM_FRAMES);
+    if (num_frames == 0)
+        return "";
+
+    char** symbols = backtrace_symbols(return_addresses, num_frames);
+
+    std::string ret;
+    for (size_t i = 0; i < num_frames; i++) {
+        fprintf(stderr, "%s\n", symbols[i]);
```
lioncash: This fprintf should likely be inside SignalHandler alongside the `fflush()` call, since it can just be one fprintf call after all the concatenation is done.

src/citra_qt/crash_handler.cpp
lioncash: One way to do it is to possibly provide the ability for the UI side to register a callback with the crash handler of the form `void CrashHandlerCallback(const std::string& stack_trace)`, which would forward the retrieved stack trace to any arbitrary front-end, and allow for localizing dialog title/user instructions. However, considering code comments say we might not be on the GUI thread when it happens, I don't know how feasible that would be.

src/citra_qt/crash_handler.cpp
lioncash: Actually, if you want to be signal-handler safe, technically you'd use POSIX `write` instead of fprintf to write to stderr.

src/citra_qt/crash_handler.cpp
lioncash: There's also [this Qt documentation](https://doc.qt.io/qt-5/unix-signals.html) and [this](https://github.com/JPNaude/dev_notes/wiki/Produce-a-stacktrace-when-something-goes-wrong-in-your-application) which may be useful.

wwylele:
Here is a test result from Ubuntu: (by intentionally trigger an ASSERT)
```
[  24.973505] Debug <Critical> core/memory.cpp:operator():146: Assertion Failed!
Unreachable code!
Oops. Now in CrashHandler::SignalHandler:
Caught signal 5 (SIGTRAP)
Git Branch: 1874 (fa25484)
Git Commit: fa25484515e2318684bc6aceae896271d9f86fb4
Stack Trace:
citra-qt() [0x4b7492]
/lib/x86_64-linux-gnu/libc.so.6(+0x354a0) [0x7f55e25184a0]
citra-qt() [0x4ec7d1]
citra-qt() [0x4eae3b]
citra-qt(_ZN21RasterizerCacheOpenGL11FlushRegionEjjPK13CachedSurfaceb+0x2b4) [0x55d624]
citra-qt() [0x4ea41b]
citra-qt(_ZN6Kernel9VMManager10UnmapRangeEjj+0x118) [0x4d6ed8]
citra-qt(_ZN6Kernel7Process10LinearFreeEjj+0xa6) [0x51af66]
citra-qt() [0x531a88]
citra-qt(_ZN3SVC7CallSVCEj+0x74) [0x5317e4]
citra-qt(_Z19InterpreterMainLoopP11ARMul_State+0xc566) [0x4fc526]
citra-qt(_ZTv0_n168_N10ARM_DynCom19ExecuteInstructionsEi+0x21) [0x4efb21]
citra-qt(_ZN4Core7RunLoopEi+0x51) [0x4cadf1]
citra-qt(_ZN9EmuThread3runEv+0x5f) [0x4b1eaf]
/usr/lib/x86_64-linux-gnu/libQt5Core.so.5(+0xa584e) [0x7f55e340f84e]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x76fa) [0x7f55e28b36fa]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7f55e25e9b5d]
QObject::~QObject: Timers cannot be stopped from another thread
QSocketNotifier: Invalid socket 28 and type 'Read', disabling...
```

lioncash:
Since Github doesn't un-collapse conversations with new comments, just leaving this here to indicate that I've added to my localization comment.

mailwl:
Unfortunally, doesn't help with Exstetra #1629 (both Debug/Release). IDE callstack doesn't help, too

extetracrash

JayFoxRox:
Isn't there a nice way to show a text box with a text field inside?

I don't like the design of the messagebox at all. It will scare users.
Also I think telling them to Ctrl+C is black magic for some users (*cough*pokemon crowd*cough*).

\- I'd even prefer just printing the backtrace to terminal, then tell the user to copy the terminal content.
That way we will at least get log + backtrace together AND we don't necessarily scare users with weird numbers and text.
If we had log *files* this would be even easier.

MerryMage:
@JayFoxRox 
I'm considering redesigning this to actually use a watchdog process like I mention in my first `TODO` comment. That way we could: (a) have a nice cross-platform Qt crash dialog, (b) also redirect stdout/stderr to that watchdog so it can save a log when necessary.

@mailwl 
Crashes within JITted code are an interesting case given that the RSP and the structure of the stack are not going to be as expected.

mailwl:
Ah, I see, generated code not belong to citra code, so, no way to link with citra code

Subv:
What's the status of this?

src/citra_qt/crash_handler.cpp
```diff
@@ -0,0 +1,277 @@
...
+#ifdef _WIN32
+
+#pragma comment(lib, "dbghelp.lib")
+
+#include <windows.h> // Must include this first.
+#include <dbghelp.h>
+
+#include "common/string_util.h"
+
+namespace CrashHandler {
```
yuriks: Whole file seems like a good addition to common instead of citra-qt

src/citra_qt/crash_handler.cpp
yuriks: This could be done with a struct:
```
struct {
  IMAGEHLP_SYMBOL64 info;
  char name[SYMBOL_NAME_SIZE];
} symbol;
```

yuriks:
ping?

MerryMage:
Not getting back to this soon, but would like to see/do similar UI work in the future.

PR #1875, opened 2016-06-02
Implement standard fog (fog mode 5)

JayFoxRox:
This implements fog mode 5.
(Known fog modes are None = 0; Fog = 5; Gas = 7. `master` treats everything like mode 0)

---

The fog mode is fully implemented in SW rasterizer and HW renderer.
[The fog was hardware tested](https://github.com/JayFoxRox/3ds-tests/tree/6aa1be178d67f115d3f9e3652affaa343dd188b8/fog) (against all logic op modes, blend modes, RGBA write masks, Z-Buffer / W-Buffer, LUT overflows / underflows, LUT index wrapping) and it's been found that this fog mode is like a final-combiner for the R, G, B channels. There is no alpha fog or dual source blending.

The overflow of the fog_lut_offset was confirmed on hardware. I didn't want to change the size of the BitField though because there might be reasons why 3dbrew lists it as 16 bit and we can't read back the actual size of the hardware register.

There are a handful of tiny accuracy problems remaining but we can't solve them until we figure out how exactly the 3DS interpolates the depth etc. I've mentioned this in the SW rasterizer code (HW is inaccurate anyway).

There is only one difference I could find between 3DS fog and this implementation: the fog LUT is cleared to zero in Citra while the 3DS seems to keep random values in it = it's not initialized for the process.
I believe the fog index is still initialized to 0 (somewhere unknown - `FINALIZE` maybe?) because Luigi's Mansion Dark Moon does fog LUT writes (in batches of 128 = fog LUT size) without ever explicitly setting the index (Note that it doesn't ever seem to use fog mode 5 either).

There are no known problems or missing functionality for this fog mode.

---

During development, I was also asked for a "Disable fog" enhancement option in the configuration. 
\- I'll leave that to someone else.
(Fog is often used for graphics effects and gives a sense of scale so I think disabling it would be pretty stupid and not an enhancement. If anybody decides to write such an enhancement they should probably just set a falloff curve for the fog instead so that it can be shifted back slightly)

---

This fixes the OoT fog and more..

Comparison screenshots:

[Citra `master`](http://imgur.com/a/AO6t9):
Zelda (Citra master)

Citra `fog`:
Zelda (Citra PR)

[Real console](http://imgur.com/a/jHhA5) (Thanks to @FenrisulfrX):
Zelda (Console)

More Screenshots by @Leo626: [Majora's Mask - Mountain Village](http://imgur.com/a/Mtw3w), [Mario Kart 7 - DK Pass](http://imgur.com/a/TrXaY), [Super Mario 3D Land - S6-5](http://imgur.com/a/ZjF1a). Also regression tested all stages of Kingdom Hearts 3D (with SW / HW renderer) and everything works.

---

Massive thanks to @Leo626 for reporting various major issues during review on my repo.
Also thanks to @wwylele, **Tohka**, @einstein95 and @FenrisulfrX (+ everybody I forgot) who ran the hardware tests for me.

### DIFF
--- a/src/video_core/rasterizer.cpp
+++ b/src/video_core/rasterizer.cpp
@@ -398,6 +398,26 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
                                                 float24::FromFloat32(static_cast<float>(w2)));
             float24 interpolated_w_inverse = float24::FromFloat32(1.0f) / Math::Dot(w_inverse, baricentric_coordinates);
 
+            // interpolated_z = z / w
+            float interpolated_z_over_w = (v0.screenpos[2].ToFloat32() * w0 +
+                                           v1.screenpos[2].ToFloat32() * w1 +
+                                           v2.screenpos[2].ToFloat32() * w2) / wsum;
+
+            // Not fully accurate. About 3 bits in precision are missing.
+            // Z-Buffer (z / w * scale + offset)
+            float depth_scale = float24::FromRaw(regs.viewport_depth_range).ToFloat32();
+            float depth_offset = float24::FromRaw(regs.viewport_depth_near_plane).ToFloat32();
+            float depth = interpolated_z_over_w * depth_scale + depth_offset;
+
+            // Potentially switch to W-Buffer
+            if (regs.depthmap_enable == Pica::Regs::DepthBuffering::WBuffering) {
+                // W-Buffer (z * scale + w * offset = (z / w * scale + offset) * w)
+                depth *= interpolated_w_inverse.ToFloat32() * wsum;
+            }
+
+            // Clamp the result
+            depth = MathUtil::Clamp(depth, 0.0f, 1.0f);
+
             // Perspective correct attribute interpolation:
             // Attribute values cannot be calculated by simple linear interpolation since
             // they are not linear in screen space. For example, when interpolating a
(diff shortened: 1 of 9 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -328,7 +363,7 @@ void RasterizerOpenGL::NotifyPicaRegisterChanged(u32 id) {
         shader_dirty = true;
         break;
 
-    // TEV stages
+    // TEV stages + fog (mode and flip; part of tev_combiner_buffer_input)
```
neobrain: uh, what?

JayFoxRox:
```
JayFoxRox: it's handling sync of tev stages but also of fog mode and fog flip (which are part of the tev_combiner_buffer_input field)
neobrain: I had literally no idea what that comment was trying to convey
JayFoxRox: I might rephrase it then
```

src/video_core/rasterizer.cpp
```diff
@@ -833,6 +854,38 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
+                if (g_state.regs.fog_flip) {
+                    fog_index = (1.0f - depth) * 128.0f;
+                } else {
+                    fog_index = depth * 128.0f;
+                }
+
+                // Generate clamped fog factor from LUT for given fog index
+                float fog_i = MathUtil::Clamp(floorf(fog_index), 0.0f, 127.0f);
+                float fog_f = fog_index - fog_i;
+                auto& fog_lut_entry = g_state.fog.lut[static_cast<unsigned int>(fog_i)];
```
lioncash: this should be a const reference.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -1040,6 +1083,30 @@ void RasterizerOpenGL::SyncBlendColor() {
...
+      regs.fog_color.g.Value() / 255.0f,
+      regs.fog_color.b.Value() / 255.0f
+    };
+    uniform_block_data.dirty = true;
+}
+
+void RasterizerOpenGL::SyncFogLUT() {
+    std::array<GLuint, 128> new_data;
+
+    for (unsigned offset = 0; offset < new_data.size(); ++offset) {
```
lioncash:
```cpp
std::transform(Pica::g_state.fog.lut.begin() Pica::g_state.fog.lut.end(), new_data.begin(), [](const auto& entry) {
    return entry.raw;
});
```

src/video_core/rasterizer.cpp
```diff
@@ -398,6 +398,27 @@ static void ProcessTriangleInternal(const Shader::OutputVertex& v0,
...
+
+            // Not fully accurate. About 3 bits in precision are missing.
+            // Z-Buffer (z / w * scale + offset)
+            float depth_scale = float24::FromRaw(regs.viewport_depth_range).ToFloat32();
+            float depth_offset = float24::FromRaw(regs.viewport_depth_near_plane).ToFloat32();
+            float depth = interpolated_z_over_w * depth_scale + depth_offset;
+
+            // Potentially switch to W-Buffer
+            if (regs.depthmap_enable == Pica::Regs::DepthBuffering::WBuffering) {
+
```
lioncash: unnecessary newline.

JayFoxRox:
I've addressed all comments

bunnei:
LGTM

PR #1877, opened 2016-06-03
Thread: update timeout when reruning WaitSynch

### LINKED ISSUE #1307: Waiting on a WaitSynchronization with timeout will reset the timeout each time the thread is awakened
If a call to WaitSynchronization1/N with timeout causes a thread to sleep, the timeout will be reset every time we repeat the WaitSynchronization call if it sends the thread to sleep again.

We should keep track of the remaining time in the Wakeup event and adjust the timeout parameter accordingly for the next call (And then restore it).

wwylele:
Fix #1307 
Here is a [small test](https://github.com/wwylele/ctrhwtest/tree/master/waitsync_timeout_test) for this.

### DIFF
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -182,6 +182,48 @@ static void PriorityBoostStarvedThreads() {
 }
 
 /**
+ * Gets the registers for timeout parameter of the next WaitSynchronization call.
+ * @param thread a pointer to the thread that is ready to call WaitSynchronization
+ * @returns a tuple of two register pointers to low and high part of the timeout parameter
+ */
+static std::tuple<u32*, u32*> GetWaitSynchTimeoutParameterRegister(Thread* thread) {
+    bool thumb_mode = (thread->context.cpsr & TBIT) != 0;
+    u16 thumb_inst = Memory::Read16(thread->context.pc & 0xFFFFFFFE);
+    u32 inst = Memory::Read32(thread->context.pc & 0xFFFFFFFC) & 0x0FFFFFFF;
+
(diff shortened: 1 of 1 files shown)


src/core/hle/kernel/thread.cpp
```diff
@@ -219,6 +219,36 @@ static void SwitchContext(Thread* new_thread) {
 
             // SVC instruction is 2 bytes for THUMB, 4 bytes for ARM
             new_thread->context.pc -= thumb_mode ? 2 : 4;
+
+            // Get the SVC call instruction
+            u16 thumb_inst = Memory::Read16(new_thread->context.pc & 0xFFFFFFFE);
```
lioncash: This should be its own function that returns a pair or a tuple containing the low or high addresses.

src/core/hle/kernel/thread.cpp
```diff
@@ -219,6 +219,36 @@ static void SwitchContext(Thread* new_thread) {
...
+            } else if ((thumb_mode && thumb_inst == 0xDF25) || (!thumb_mode && inst == 0x0F000025)) {
+                // svc #0x25 (WaitSynchronizationN)
+                reg_timeout_low = &new_thread->context.cpu_registers[0];
+                reg_timeout_high = &new_thread->context.cpu_registers[4];
+            } else {
+                UNREACHABLE();
+            }
+
+            // Update the timeout parameter
+            s64 timeout = ((s64)*reg_timeout_high << 32) | *reg_timeout_low;
```
lioncash: This should also be its own function that performs the update.

wwylele:
Addressed

Subv:
LGTM, but this seems more of a band-aid solution until we change how our scheduler works.

bunnei:
LGTM

PR #1884, opened 2016-06-04
service: Add other DLP services

lioncash:
Specifically, dlp::CLNT and dlp::FKCL

Moves all DLP services to their own folder like with other services that contain multiple sub-modules.


Subv:
LGTM

bunnei:
LGTM

PR #1890, opened 2016-06-07
Fix boot_filename encode on Windows

LFsWang:
#1620 did not deal with command line encode problem on no-qt build. This PR conuinue to fix it.

Another way  is use `wmain` ,but I think `wmain` will cause more portability issues so that I use `CommandLineToArgvW` to convert path to Unicode format.

Thanks @yami-hack 's comment!

image

### DIFF
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -17,11 +17,16 @@
 #include <getopt.h>
 #endif
 
+#ifdef _WIN32
+#include <Windows.h>
+#endif
+
 #include "common/logging/log.h"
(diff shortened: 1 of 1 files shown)


src/citra/citra.cpp
```diff
@@ -17,11 +17,16 @@
 #include <getopt.h>
 #endif
 
+#ifdef _WIN32
+#include<Windows.h>
```
wwylele: space after #include

LFsWang:
Addressed

PR #1894, opened 2016-06-10
Implement config savegame editing & clean up

wwylele:
This PR should be reviewed commit per commit.
Here is explanation for each commit.

These are independent to the rest of commits. The enum members are used nowhere (even not in the following system settings configure tab, because it matches codes to combo box indexes directly). They are added there just for completeness.

Note that `SoundOutputMode` happens to be the same as [`DSP::HLE::DspConfiguration::OutputFormat`](https://github.com/citra-emu/citra/blob/master/src/audio_core/hle/dsp.h#L355-359). I still defined a separate enum, for the following reason:
 - these two enum are too far away from each other in the code tree.
 - they are not related in program logic. It is the game that relates them.

Added the service function and refactored with `GetConfigInfoBlock`. Also added/fixed the error code for them. The error code is [hwtested](https://github.com/wwylele/ctrhwtest/tree/master/CFG_service_test)

`SetConfigInfoBlk4` can be tested by running 3DS System Settings app. (you can dump it from your 3DS NAND) However, when the app runs in Citra, it thinks Citra needs an initial system setup, then ask user to move the 3d slider (which is not implemented yet). There are ~~three~~ two ways to bypass:
 - ~~stub a value for 3d slider~~ (I failed)
 - dump a config savegame from real 3DS to citra
 - add the following code to to `cfg.cpp: FormatConfig()`, and remove the config savegame to let citra create a new one.
```c++ 
// 0x00110000 - System setup required?
u32 data = 0x00010001;
res = CreateConfigInfoBlk(0x00110000, 0x4, 0xE /*ConfigPermission::Public*/, &data);
if (!res.IsSuccess()) return res;
```

In System Settings, you can edit the username, birthday, language, etc (though the username can only be "Citra" due to stubbed software keyboard). They will be saved after changing.

mset

Allows user to change username, birthday, system language and sound output mode from front-end (i.e. edit these information in config block). System language selector is originally worked by @uberhalit  (#1755).

The profile country / location information is also editable in theory, but they are more complicated and not well-REed, so I'd like to leave it to the next PR.

To prevent data race, the tab is disabled when game is running. Editing system settings during games makes no sense because this cannot be done on real 3DS, and won't benefit anything while game can be even not aware of user's changes.

~~TODO: filter birthday (use DateTimeEdit probably)~~

Taken off. Contains too many unconfirmed guessing and distract people (at least me) from reviewing other things. Will be discussed in the future.

### DIFF
--- a/src/citra_qt/configure_dialog.cpp
+++ b/src/citra_qt/configure_dialog.cpp
@@ -9,9 +9,10 @@
 
 #include "core/settings.h"
 
-ConfigureDialog::ConfigureDialog(QWidget *parent) :
+ConfigureDialog::ConfigureDialog(QWidget *parent, bool running) :
     QDialog(parent),
-    ui(new Ui::ConfigureDialog)
+    ui(new Ui::ConfigureDialog),
+    emulation_running(running)
 {
     ui->setupUi(this);
     this->setConfiguration();
@@ -21,10 +22,14 @@ ConfigureDialog::~ConfigureDialog() {
 }
 
 void ConfigureDialog::setConfiguration() {
+    // System tab needs set manually
+    // depending on whether emulation is running
+    ui->systemTab->setConfiguration(emulation_running);
 }
 
 void ConfigureDialog::applyConfiguration() {
     ui->generalTab->applyConfiguration();
+    ui->systemTab->applyConfiguration();
     ui->audioTab->applyConfiguration();
     ui->debugTab->applyConfiguration();
 }
--- a/src/citra_qt/configure_system.cpp
+++ b/src/citra_qt/configure_system.cpp
@@ -1,0 +1,136 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "citra_qt/configure_system.h"
+#include "citra_qt/ui_settings.h"
+#include "ui_configure_system.h"
+
+#include "core/hle/service/fs/archive.h"
+#include "core/hle/service/cfg/cfg.h"
+
+static const std::array<int, 12> days_in_month = {{
+    31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31
+}};
+
+ConfigureSystem::ConfigureSystem(QWidget *parent) :
+    QWidget(parent),
+    ui(new Ui::ConfigureSystem) {
+    ui->setupUi(this);
+
+    connect(ui->combo_birthmonth, SIGNAL(currentIndexChanged(int)), SLOT(updateBirthdayComboBox(int)));
+}
+
+ConfigureSystem::~ConfigureSystem() {
+}
+
+void ConfigureSystem::setConfiguration(bool emulation_running) {
+    enabled = !emulation_running;
+
+    if (!enabled) {
+        ReadSystemSettings();
+        ui->group_system_settings->setEnabled(false);
+    } else {
+        // This tab is enabled only when game is not running (i.e. all service are not initialized).
+        // Temporarily register archive types and load the config savegame file to memory.
+        Service::FS::RegisterArchiveTypes();
+        ResultCode result = Service::CFG::LoadConfigNANDSaveFile();
+        Service::FS::UnregisterArchiveTypes();
+
+        if (result.IsError()) {
+            ui->label_disable_info->setText(tr("Failed to load system settings data."));
+            ui->group_system_settings->setEnabled(false);
+            enabled = false;
+            return;
+        }
+
+        ReadSystemSettings();
+        ui->label_disable_info->hide();
+    }
+}
+
+void ConfigureSystem::ReadSystemSettings() {
+    // set username
+    username = Service::CFG::GetUsername();
+    // ui->edit_username->setText(QString::fromStdU16String(username)); // TODO(wwylele): Use this when we move to Qt 5.5
+    ui->edit_username->setText(QString::fromUtf16(reinterpret_cast<const ushort*>(username.data())));
+
+    // set birthday
+    std::tie(birthmonth, birthday) = Service::CFG::GetBirthday();
+    ui->combo_birthmonth->setCurrentIndex(birthmonth - 1);
+    ui->combo_birthday->setCurrentIndex(birthday - 1);
+
+    // set system language
+    language_index = Service::CFG::GetSystemLanguage();
+    ui->combo_language->setCurrentIndex(language_index);
+
+    // set sound output mode
+    sound_index = Service::CFG::GetSoundOutputMode();
+    ui->combo_sound->setCurrentIndex(sound_index);
+}
+
+void ConfigureSystem::applyConfiguration() {
+    if (!enabled)
+        return;
+
+    bool modified = false;
+
+    // apply username
+    // std::u16string new_username = ui->edit_username->text().toStdU16String(); // TODO(wwylele): Use this when we move to Qt 5.5
+    std::u16string new_username(reinterpret_cast<const char16_t*>(ui->edit_username->text().utf16()));
+    if (new_username != username) {
+        Service::CFG::SetUsername(new_username);
+        modified = true;
+    }
+
+    // apply birthday
+    int new_birthmonth = ui->combo_birthmonth->currentIndex() + 1;
+    int new_birthday = ui->combo_birthday->currentIndex() + 1;
+    if (birthmonth != new_birthmonth || birthday != new_birthday) {
+        Service::CFG::SetBirthday(new_birthmonth, new_birthday);
+        modified = true;
+    }
+
+    // apply language
+    int new_language = ui->combo_language->currentIndex();
+    if (language_index != new_language) {
+        Service::CFG::SetSystemLanguage(static_cast<Service::CFG::SystemLanguage>(new_language));
+        modified = true;
+    }
+
+    // apply sound
+    int new_sound = ui->combo_sound->currentIndex();
+    if (sound_index != new_sound) {
+        Service::CFG::SetSoundOutputMode(static_cast<Service::CFG::SoundOutputMode>(new_sound));
+        modified = true;
+    }
+
+    // update the config savegame if any item is modified.
+    if (modified)
+        Service::CFG::UpdateConfigNANDSavegame();
+}
+
+void ConfigureSystem::updateBirthdayComboBox(int birthmonth_index) {
+    if (birthmonth_index < 0 || birthmonth_index >= 12)
+        return;
+
+    // store current day selection
+    int birthday_index = ui->combo_birthday->currentIndex();
+
+    // get number of days in the new selected month
+    int days = days_in_month[birthmonth_index];
+
+    // if the selected day is out of range,
+    // reset it to 1st
+    if (birthday_index < 0 || birthday_index >= days)
+        birthday_index = 0;
+
+    // update the day combo box
+    ui->combo_birthday->clear();
+    for (int i = 1; i <= days; ++i) {
+        ui->combo_birthday->addItem(QString::number(i));
+    }
+
+    // restore the day selection
+    ui->combo_birthday->setCurrentIndex(birthday_index);
+}
--- a/src/citra_qt/configure_system.ui
+++ b/src/citra_qt/configure_system.ui
@@ -1,0 +1,252 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>ConfigureSystem</class>
+ <widget class="QWidget" name="ConfigureSystem">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>360</width>
+    <height>377</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Form</string>
+  </property>
+  <layout class="QHBoxLayout" name="horizontalLayout">
+   <item>
+    <layout class="QVBoxLayout" name="verticalLayout">
+     <item>
+      <widget class="QGroupBox" name="group_system_settings">
+       <property name="title">
+        <string>System Settings</string>
+       </property>
+       <layout class="QGridLayout" name="gridLayout">
+        <item row="0" column="0">
+         <widget class="QLabel" name="label_username">
+          <property name="text">
+           <string>Username</string>
+          </property>
+         </widget>
+        </item>
+        <item row="0" column="1">
+         <widget class="QLineEdit" name="edit_username">
+          <property name="sizePolicy">
+           <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
+            <horstretch>0</horstretch>
+            <verstretch>0</verstretch>
+           </sizepolicy>
+          </property>
+          <property name="maxLength">
+           <number>10</number>
+          </property>
+         </widget>
+        </item>
+        <item row="1" column="0">
+         <widget class="QLabel" name="label_birthday">
+          <property name="text">
+           <string>Birthday</string>
+          </property>
+         </widget>
+        </item>
+        <item row="1" column="1">
+         <layout class="QHBoxLayout" name="horizontalLayout_birthday2">
+          <item>
+           <widget class="QComboBox" name="combo_birthmonth">
+            <item>
+             <property name="text">
+              <string>January</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>February</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>March</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>April</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>May</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>June</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>July</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>August</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>September</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>October</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>November</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>December</string>
+             </property>
+            </item>
+           </widget>
+          </item>
+          <item>
+           <widget class="QComboBox" name="combo_birthday"/>
+          </item>
+         </layout>
+        </item>
+        <item row="2" column="0">
+         <widget class="QLabel" name="label_language">
+          <property name="text">
+           <string>Language</string>
+          </property>
+         </widget>
+        </item>
+        <item row="2" column="1">
+         <widget class="QComboBox" name="combo_language">
+          <item>
+           <property name="text">
+            <string>Japanese (日本語)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>English</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>French (français)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>German (Deutsch)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Italian (italiano)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Spanish (español)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Simplified Chinese (简体中文)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Korean (한국어)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Dutch (Nederlands)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Portuguese (português)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Russian (Русский)</string>
+           </property>
+          </item>
+          <item>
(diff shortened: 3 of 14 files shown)


src/core/hle/service/cfg/cfg.h
```diff
@@ -35,15 +35,32 @@ enum SystemLanguage {
     LANGUAGE_KO = 7,
     LANGUAGE_NL = 8,
     LANGUAGE_PT = 9,
-    LANGUAGE_RU = 10
+    LANGUAGE_RU = 10,
+    LANGUAGE_TW = 11
```
wwylele: "Taiwan"/"Taiwanese" is not a language. I am just following 3dbrew.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -220,10 +233,26 @@ void GetConfigInfoBlk8(Service::Interface* self) {
...
 }
 
+void SetConfigInfoBlkSystem(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 block_id = cmd_buff[1];
+    u32 size = cmd_buff[2];
+    VAddr data_pointer = cmd_buff[4];
+
+    if (!Memory::IsValidVirtualAddress(data_pointer)) {
+        cmd_buff[1] = -1; // TODO(Subv): Find the right error code
```
wwylele: 3DS will actually freeze on invalid address (Kernel PANIC when mapping address?). I am not sure what to do with this code, so just copied it from `GetConfigInfoBlk8`'s

src/core/hle/service/cfg/cfg_i.cpp
```diff
@@ -21,17 +21,17 @@ const Interface::FunctionInfo FunctionTable[] = {
     {0x00090040, GetCountryCodeString,                 "GetCountryCodeString"},
     {0x000A0040, GetCountryCodeID,                     "GetCountryCodeID"},
     // cfg:i
-    {0x04010082, GetConfigInfoBlk8,                    "GetConfigInfoBlk8"},
-    {0x04020082, nullptr,                              "SetConfigInfoBlk4"},
+    {0x04010082, GetConfigInfoBlkSystem,               "GetConfigInfoBlk8"},
```
mailwl: name of service function not the same

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -40,6 +40,20 @@ struct SaveFileConfig {
 };
 static_assert(sizeof(SaveFileConfig) == 0x455C, "SaveFileConfig header must be exactly 0x455C bytes");
 
+enum ConfigBlockID {
```
lioncash: May want to vertically align these like

```cpp
enum ConfigBlockID {
    StereoCameraSettingsBlockID = 0x00050005,
    SoundOutputModeBlockID      = 0x00070001,
    ConsoleUniqueIDBlockID      = 0x00090001,
    UsernameBlockID             = 0x000A0000,
    BirthdayBlockID             = 0x000A0001,
    LanguageBlockID             = 0x000A0002,
    CountryInfoBlockID          = 0x000B0000,
    CountryNameBlockID          = 0x000B0001,
    StateNameBlockID            = 0x000B0002,
    EULAVersionBlockID          = 0x000D0000,
    ConsoleModelBlockID         = 0x000F0004,
};
```

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -431,14 +483,73 @@ void Init() {
...
+    AddService(new CFG_S_Interface);
+    AddService(new CFG_U_Interface);
+
+    LoadConfigNANDSaveFile();
 }
 
 void Shutdown() {
 }
 
+void SetUsername(std::u16string name) {
```
lioncash:
```cpp
const std::u16string&
```

src/core/hle/service/cfg/cfg.cpp
wwylele: Addresed

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -234,41 +263,68 @@ void FormatConfig(Service::Interface* self) {
...
+
     // The data is located in the block header itself if the size is less than 4 bytes
     if (itr->size <= 4)
-        memcpy(output, &itr->offset_or_data, itr->size);
+        pointer = &itr->offset_or_data;
     else
-        memcpy(output, &cfg_config_file_buffer[itr->offset_or_data], itr->size);
+        pointer = &cfg_config_file_buffer[itr->offset_or_data];
 
+    return ResultVal<void*>::WithCode(RESULT_SUCCESS, pointer);
```
lioncash: could this not be `return MakeResult<void*>(pointer);`?

wwylele: Oops, forget to push the fix

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -431,14 +483,73 @@ void Init() {
...
+    SetConfigInfoBlock(UsernameBlockID, sizeof(block), true, &block);
+}
+
+std::u16string GetUsername() {
+    UsernameBlock block;
+    GetConfigInfoBlock(UsernameBlockID, sizeof(block), true, &block);
+
+    // the username string in the block isn't null-terminated,
+    // so copy it to a buffer to add trailing null char.
+    char16_t username[ARRAY_SIZE(block.username) + 1]{};
```
lioncash: You can likely just do:

```cpp
std::u16string username(ARRAY_SIZE(block.username) + 1, '\0');
std::copy(std::begin(block.username), std::end(block.username), std::begin(username));
return username;
```
to operate on the sized string instance directly, or just do

```cpp
return std::u16string(block.username) + '\0'
```

if preserving all 20 entries isn't important.

wwylele: @lioncash I did it in another way. I don't want the string to contain null character

src/core/hle/service/cfg/cfg.cpp
Subv: Leave it, it should be fixed eventually when we implement kernel command buffer translation

src/citra_qt/configure_system.ui
```diff
@@ -0,0 +1,265 @@
...
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>Russian (Русский)</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>Traditional Chinese (正體中文)</string>
```
LFsWang: 繁體中文

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -431,14 +483,75 @@ void Init() {
...
+
+std::u16string GetUsername() {
+    UsernameBlock block;
+    GetConfigInfoBlock(UsernameBlockID, sizeof(block), true, &block);
+
+    // the username string in the block isn't null-terminated,
+    // so we need to find the end manually.
+    auto begin = std::begin(block.username);
+    auto end = std::end(block.username);
+    while (end != begin && *(end - 1) == u'\0')
```
lioncash: This is literally performing the same behavior that the single-argument constructor for u16string does.

e.g.
```cpp
return std::u16string(block.username);
```
will do a cut-off at the first '\0' it finds.

lioncash: That, or do

```cpp
std::u16string username(block.username, ARRAY_SIZE(block.username));

auto pos = username.find(char16_t('\0'));
if (pos != std::string::npos)
    username.erase(pos);
```

if you want to be safer and more straightforward.

wwylele: @lioncash It is not the same behaviour when `block.username` is full of 10 characters because it doesn't has null in the end in this case. (Well, since the next member `block.zero` is almost guaranteed to be zero, it is still safe to just past `block.username` to the constructor - but I don't want to rely on this.)

lioncash:
UI-wise you might may want to alter it so one button isn't further to the left than the other buttons.

wwylele:
Updated with aligned controls, and dynamically adding items to birthday depending on birthmonth. Didn't use `QDateEdit` for birthday because I feel bad when input date to it, and if I let it pop a calendar, the year is always shown.

wwylele:
I took of the last commit `Service::CFG: name permissions`. It is still an unconfirmed theory and distract me a lot. I will leave it to the future.

mailwl:
@wwylele may you name config functions as 3dbrew name it? and add log for calling.
Sorry if you dont want

mailwl:
@wwylele,may you add eula version agreement? both from zero to 255-255? want to see? how many games will broken at this moment

wwylele:
@mailwl editing the EULA version in config looks weird to me, and I don't want to include too many stuffs in this PR (see, the description is already that long). You can open your own PR after this.

mailwl:
OK, but I don;t think it's weird, because user shold press the button to agree (but version itself depends of firmware, for sure)

src/citra_qt/configure_dialog.cpp
```diff
@@ -21,10 +22,15 @@ ConfigureDialog::~ConfigureDialog() {
 }
 
 void ConfigureDialog::setConfiguration() {
+
```
Subv: empty line

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+ConfigureSystem::~ConfigureSystem() {
+}
+
+void ConfigureSystem::setConfiguration(bool emulation_running) {
+    this->enabled = !emulation_running;
+
+    if (!enabled) {
+        ui->group_system_settings->setEnabled(false);
+    }
```
Subv: else in the same line as }

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,140 @@
...
+        if (result.IsError()) {
+            ui->label_disable_info->setText(tr("Failed to load system settings data."));
+            ui->group_system_settings->setEnabled(false);
+            this->enabled = false;
+            return;
+        }
+
+        // set username
+        username = Service::CFG::GetUsername();
+        ui->edit_username->setText(QString::fromUtf16(reinterpret_cast<const ushort*>(username.data()))); // Should be QString::fromStdU16String(username) in Qt 5
```
Subv: Why do we still support Qt 4?

Subv:
LGTM after the comments are addressed

src/citra_qt/configure_system.cpp
wwylele: I am also asking this question

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,140 @@
...
+    ui->edit_username->setMaxLength(10);
+
+    connect(ui->combo_birthmonth, SIGNAL(currentIndexChanged(int)), SLOT(updateBirthdayComboBox(int)));
+}
+
+ConfigureSystem::~ConfigureSystem() {
+}
+
+void ConfigureSystem::setConfiguration(bool emulation_running) {
+    this->enabled = !emulation_running;
```
lioncash: the `this` isn't needed here (it's also not used in the conditional below)

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,140 @@
...
+        // This tab is enabled only when game is not running (i.e. all service are not initialized).
+        // Temporarily register archive types and load the config savegame file to memory.
+        Service::FS::RegisterArchiveTypes();
+        ResultCode result = Service::CFG::LoadConfigNANDSaveFile();
+        Service::FS::UnregisterArchiveTypes();
+
+        if (result.IsError()) {
+            ui->label_disable_info->setText(tr("Failed to load system settings data."));
+            ui->group_system_settings->setEnabled(false);
+            this->enabled = false;
```
lioncash: ditto with the `this`

src/citra_qt/configure_system.cpp
lioncash: I can move us over to Qt 5 on the CI if you want, so that we can get rid of Qt 4 stuff.

wwylele: @lioncash yes please. Also remove the qt 4 flag in cmake and update the wiki.

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+#include "citra_qt/configure_system.h"
+#include "citra_qt/ui_settings.h"
+#include "ui_configure_system.h"
+
+#include "core/hle/service/fs/archive.h"
+#include "core/hle/service/cfg/cfg.h"
+
+static const std::array<int, 12> days_in_month = {{
+    31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31
```
bunnei: 29 in February is a leap year... is this important?

bunnei:
Travis is broken... Otherwise code LGTM

wwylele:
@bunnei 
> Removed the work-around for Qt 4 so Linux CI build is expected to fail

I am waiting for the discussion whether we should drop Qt 4.

bunnei:
@wwylele #1917 was merged so you should be able to rebase and this is good to go?

lioncash:
Is there any reason why [QDateTimeEdit](https://doc.qt.io/qt-5/qdatetimeedit.html#displayFormat-prop) with a custom formatter isn't used?

wwylele:
@lioncash 
> Didn't use QDateEdit for birthday because I feel bad when input date to it, and if I let it pop a calendar, the year is always shown.

Why is the build still failed :confused:

lioncash:
@wwylele 

> Didn't use QDateEdit for birthday because I feel bad when input date to it, and if I let it pop a calendar, the year is always shown.

You can just call `setCalendarPopup(false)` to prevent the calendar view.

> Why did the build still fail

It seems that those Unicode handling functions were added in 5.5. I can see if I can import the PPA I was initially going to use prior to realizing we could just kick the distribution from Precise to Trusty.

wwylele:
@lioncash what I meant on `QDataEdit` is:
If I let the calender popup, the year will be shown, which is weird;
if I don't let the calender popup, I feel really uncomfortable to input the date.

wwylele:
@lioncash ping about Qt version? Should I still wait?

jroweboy:
@wwylele looks like the QT version changes have already been merged https://github.com/citra-emu/citra/pull/1917 Is there anything else that you need to wait for?

wwylele:
@jroweboy See the discussion above, the function I used was introduced in Qt 5.5, while Travis is using 5.2

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+ConfigureSystem::~ConfigureSystem() {
+}
+
+void ConfigureSystem::setConfiguration(bool emulation_running) {
+    enabled = !emulation_running;
+
+    if (!enabled) {
+        ui->group_system_settings->setEnabled(false);
```
yuriks: We should still read the CFG settings when emulation is running, otherwise we show bogus values in the UI.

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+ConfigureSystem::ConfigureSystem(QWidget *parent) :
+    QWidget(parent),
+    ui(new Ui::ConfigureSystem) {
+    ui->setupUi(this);
+
+    // Limits the username length.
+    // setMaxLength seems to count characters in char16_t,
+    // which is what we need because we want to limit the name
+    // to 20 bytes.
```
yuriks: IMO this comment should be removed. The call limits the text box to 10 *characters*, which is what the limit should be. It doesn't have anything to do with how it's stored.

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,138 @@
...
+    ui->edit_username->setMaxLength(10);
```
yuriks: This can be set in the .ui file.

src/citra_qt/configure_system.ui
```diff
@@ -0,0 +1,249 @@
...
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>English</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>French (français)</string>
```
yuriks: Native name should be capitalized too. (They're proper nouns, which are capitalized in all these languages afaik.)

yuriks: Incorrect, for French, at least, from googling. TIL. "Português" should definitely be capitalized though (I'm a native speaker.)

src/citra_qt/configure_system.ui
```diff
@@ -0,0 +1,249 @@
...
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Dutch (Nederlands)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Portuguese (português)</string>
```
yuriks: "Português" should be capitalized (I'm a native speaker.)

src/citra_qt/configure_system.cpp
wwylele: I left this comment originally because Qt treats a non-BMP unicode character (4 bytes in UTF-16) as __two__ characters (at least on my PC), which happens to be the same as 3DS's behavior. I should have been more explicit on this.

src/citra_qt/configure_system.cpp
wwylele: There is a small chance of race conditions to read settings when emulation is running (probably only happens when running System Settings app). Should I add a mutex or something for it?

src/citra_qt/configure_system.ui
wwylele: Sorry...  the language list was mostly inherited from #1755 where there was already some arguments about such thing, but forgot to mention these in the description. Probably needs more native speakers to review.

yuriks: After consulting a style guide, it turns out that capitalizing or not depends on the exact context where its being used, I think the correct is to leave it non-capitalized in this case. Didn't know it came from an ISO list, the names are probably correct then.

neobrain:

2016-07-02T11:33:42Z: https://github.com/citra-emu/citra/compare/45c91bf87b5419184630931fcd62b304d2c547b6...wwylele:0adf8e9271cc13b02213736ec987118d05cf40fc

wwylele:
Addressed @yuriks ' comments.
I was concerned about data race of reading config data when running, but this will happens only if user opens the configure window exactly when the app is writing to config, which is nearly impossible. So I just make it read data directly without any synchronization.
I also pushed back the work-around for low version Qt. I'd like to get this merged first. We can clean it up later.

neobrain:

2016-07-02T11:45:27Z: https://github.com/citra-emu/citra/compare/45c91bf87b5419184630931fcd62b304d2c547b6...wwylele:42866734bf1b259a863216ebb20a3e967dadd826

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+        ReadSystemSettings();
+        ui->label_disable_info->hide();
+    }
+}
+
+void ConfigureSystem::ReadSystemSettings() {
+    // set username
+    username = Service::CFG::GetUsername();
+    // ui->edit_username->setText(QString::fromStdU16String(username)); // Use this when we move to Qt 5.5
```
yuriks: TODO

src/citra_qt/configure_system.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+void ConfigureSystem::applyConfiguration() {
+    if (!enabled)
+        return;
+
+    bool modified = false;
+
+    // apply username
+    // std::u16string new_username = ui->edit_username->text().toStdU16String(); // Use this when we move to Qt 5.5
```
yuriks: TODO again

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -234,31 +263,52 @@ void FormatConfig(Service::Interface* self) {
     cmd_buff[1] = Service::CFG::FormatConfig().raw;
 }
 
-ResultCode GetConfigInfoBlock(u32 block_id, u32 size, u32 flag, u8* output) {
+static ResultVal<void*> GetConfigInfoBlockPointer(u32 block_id, u32 size, u32 flag) {
```
yuriks: I think it's better to keep this as a u8 pointer. It points to byte data always (which you can then interpret as something else), not a general type directly.

neobrain:

2016-07-03T05:30:20Z: https://github.com/citra-emu/citra/compare/45c91bf87b5419184630931fcd62b304d2c547b6...wwylele:9d0d3df05f2a3d59ae7cb48573944ad9b12abeca

src/core/hle/service/cfg/cfg.cpp
wwylele: @yuriks [`CreateConfigInfoBlk`](https://github.com/citra-emu/citra/blob/master/src/core/hle/service/cfg/cfg.cpp#L265) was using void pointer, and I want to unify them (`Create/Get/SetConfigInfoBlk`). But if I make all of them using u8 pointer, then in [`FormatConfig`](https://github.com/citra-emu/citra/blob/master/src/core/hle/service/cfg/cfg.cpp#L320) and in my utility functions there will be `reinterpret_cast<u8*>` everywhere, which seems ugly to me.

bunnei:
Still LGTM

src/core/hle/service/cfg/cfg.cpp
yuriks: Ok, that's fair I suppose.

yuriks:
LGTM. The "reading/editing cfg while game runs" thing isn't 100% bullet-proof but I think it's the best we can do right now. Since it's a modal dialog anyway the user can't stop the emulation with it open, which would be what causes more major issues.

In the future, we need to overhaul our threading to allow code such as this to run in the context of the emulation thread or at least in mutual exclusion to it, to avoid any race conditions.

yuriks:
Can you do the fixup? Then I'll merge this.

Issue #1895, opened 2016-06-10
unmapped Read32


lefta:
I just updated Citra and was surprised to see no game working.

### Symptoms

The game view widget seems to be not repainted and closing Citra freezes the whole application.

### The log : 

    $ src/citra_qt/citra-qt
    using qt5ct plugin
    [   0.000000] Common.Filesystem <Warning> common/file_util.cpp:CreateFullPath:207: path exists /home/cedric/.config/citra-emu/qt-config.ini
    [   0.000405] Audio <Error> audio_core/audio_core.cpp:SelectSink:66: AudioCore::SelectSink given invalid sink_id
    QWidget::paintEngine: Should no longer be called
    QPainter::begin: Paint device returned engine == 0, type: 1
    [   4.589359] Frontend <Info> citra_qt/main.cpp:BootGame:317: Citra starting...
    [   4.606761] Service.FS <Info> core/file_sys/archive_sdmc.cpp:ArchiveFactory_SDMC:21:  Directory /home/cedric/.local/share/citra-emu/sdmc/ set as SDMC.
    [   4.606799] Common.Filesystem <Warning> common/file_util.cpp:CreateFullPath:207: path exists /home/cedric/.local/share/citra-emu/sdmc/
    [   4.606817] Service.FS <Info> core/file_sys/archive_savedata.cpp:ArchiveFactory_SaveData:42: Directory /home/cedric/.local/share/citra-emu/sdmc/Nintendo 3DS/00000000000000000000000000000000/00000000000000000000000000000000/title/ set as SaveData.
    [   4.606838] Service.FS <Info> core/file_sys/archive_extsavedata.cpp:ArchiveFactory_ExtSaveData:62: Directory /home/cedric/.local/share/citra-emu/sdmc/Nintendo 3DS/00000000000000000000000000000000/00000000000000000000000000000000/extdata/ set as base for ExtSaveData.
    [   4.606851] Common.Filesystem <Warning> common/file_util.cpp:CreateFullPath:207: path exists /home/cedric/.local/share/citra-emu/sdmc/Nintendo 3DS/00000000000000000000000000000000/00000000000000000000000000000000/extdata/
    [   4.606863] Service.FS <Info> core/file_sys/archive_extsavedata.cpp:ArchiveFactory_ExtSaveData:62: Directory /home/cedric/.local/share/citra-emu/nand/data/00000000000000000000000000000000/extdata/ set as base for ExtSaveData.
[...]

PR #1893, opened 2016-06-09
Split huge interpreter source file into translation info and interpreter (+ some tiny misc style fixes)

archshift:
Two 2500-3000 line files are still more palatable than a 6000 line file!

PR #1898, opened 2016-06-11
Refactor arm_dyncom_interpreter into several files (take 2)

archshift:
See #1893 for take 1.

I've split this into several comments (please review them individually) so that it's no longer one monolithic change.

I've also tested this, and fixed the issue that users were previously happening ( #1895 ). Specifically, the `top` variable of the instruction buffer wasn't being updated after I changed the `AllocBuffer` function, but with this PR it is correctly changed.


archshift:
Note to self: 9da0e9f8b314682896ccc368ec658c1ff8fd2a9d doesn't build, so I need to backport the fix in the next commit
- [ ] Done

src/core/arm/dyncom/arm_dyncom_trans.cpp
```diff
@@ -1,5 +1,32 @@
...
+#include "core/arm/dyncom/arm_dyncom_interpreter.h"
+#include "core/arm/dyncom/arm_dyncom_trans.h"
+#include "core/arm/skyeye_common/armstate.h"
+#include "core/arm/skyeye_common/armsupp.h"
+#include "core/arm/skyeye_common/vfp/vfp.h"
+
+char trans_cache_buf[TRANS_CACHE_SIZE];
+size_t trans_cache_buf_top = 0;
+
+static FORCE_INLINE void* AllocBuffer(size_t size) {
```
lioncash: This doesn't need to be force inlined.

archshift:
Addressed. Anything else?

bunnei:
LGTM

PR #1906, opened 2016-06-12
Add "MinGW" precompile

yami-hack:
Compiler for mingw64,I don't seem to be able to do more.in TDM-GCC,
1,may need to turn off icov support,construct a libpng library,some errors are external module problems.
2,need Qt64(mingw64)
3,need a SDL2-devel-2.0.4-mingw(need to modify the "CMakeLists.txt" to support SDL2-devel-2.0.4-mingw)
Signed-off-by: yami-hack <yami-hack@foxmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,10 @@ if (NOT MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.1")
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse4.1")
     endif()
+
+    if (MINGW)
+        add_definitions(-DMINGW_HAS_SECURE_API)
+    endif()
 else()
     # Silence "deprecation" warnings
     add_definitions(/D_CRT_SECURE_NO_WARNINGS /D_CRT_NONSTDC_NO_DEPRECATE /D_SCL_SECURE_NO_WARNINGS)
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -9,7 +9,6 @@
 #include "common/logging/log.h"
 
 #ifdef _WIN32
-    #include <share.h>
     #include <windows.h>
     #include <shlobj.h> // for SHGetFolderPath
     #include <shellapi.h>
@@ -928,11 +927,7 @@ bool IOFile::Open(const std::string& filename, const char openmode[])
 {
     Close();
 #ifdef _WIN32
-#if defined(__MINGW64__) && !defined(MINGW_HAS_SECURE_API)
-    m_file = _wfsopen(Common::UTF8ToUTF16W(filename).c_str(), Common::UTF8ToUTF16W(openmode).c_str(), SH_DENYNO);
-#else
     _wfopen_s(&m_file, Common::UTF8ToUTF16W(filename).c_str(), Common::UTF8ToUTF16W(openmode).c_str());
-#endif
 #else
     m_file = fopen(filename.c_str(), openmode);
 #endif
@@ -985,16 +980,10 @@ bool IOFile::Flush()
 bool IOFile::Resize(u64 size)
 {
     if (!IsOpen() || 0 !=
-
 #ifdef _WIN32
-#if defined(__MINGW64__) && !defined(MINGW_HAS_SECURE_API)
-        //TDM-GCC64 does not supports _chsize_s ?in <io_s.h>
-        _chsize(_fileno(m_file), size)
-#else
         // ector: _chsize sucks, not 64-bit safe
         // F|RES: changed to _chsize_s. i think it is 64-bit safe
         _chsize_s(_fileno(m_file), size)
-#endif
 #else
         // TODO: handle 64bit and growing
         ftruncate(fileno(m_file), size)
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -928,7 +928,7 @@ bool IOFile::Open(const std::string& filename, const char openmode[])
 {
     Close();
 #ifdef _WIN32
-#if defined(__MINGW64__)
+#if defined(__MINGW64__) && !defined(MINGW_HAS_SECURE_API)
     m_file = _wfsopen(Common::UTF8ToUTF16W(filename).c_str(), Common::UTF8ToUTF16W(openmode).c_str(), SH_DENYNO);
 #else
     _wfopen_s(&m_file, Common::UTF8ToUTF16W(filename).c_str(), Common::UTF8ToUTF16W(openmode).c_str());
@@ -987,7 +987,7 @@ bool IOFile::Resize(u64 size)
     if (!IsOpen() || 0 !=
 
 #ifdef _WIN32
-#if defined(__MINGW64__)
+#if defined(__MINGW64__) && !defined(MINGW_HAS_SECURE_API)
         //TDM-GCC64 does not supports _chsize_s ?in <io_s.h>
         _chsize(_fileno(m_file), size)
 #else
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -929,7 +929,7 @@ bool IOFile::Open(const std::string& filename, const char openmode[])
     Close();
 #ifdef _WIN32
 #if defined(__MINGW64__)
-    m_file = _wfsopen(Common::UTF8ToUTF16W(filename).c_str(), Common::UTF8ToUTF16W(openmode).c_str(),SH_DENYNO);
+    m_file = _wfsopen(Common::UTF8ToUTF16W(filename).c_str(), Common::UTF8ToUTF16W(openmode).c_str(), SH_DENYNO);
 #else
     _wfopen_s(&m_file, Common::UTF8ToUTF16W(filename).c_str(), Common::UTF8ToUTF16W(openmode).c_str());
 #endif
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -9,6 +9,7 @@
 #include "common/logging/log.h"
 
 #ifdef _WIN32
+    #include <share.h>
     #include <windows.h>
     #include <shlobj.h> // for SHGetFolderPath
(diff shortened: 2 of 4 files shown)


src/common/string_util.cpp
```diff
@@ -292,11 +292,11 @@ std::string ReplaceAll(std::string result, const std::string& src, const std::st
     return result;
 }
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || !__MINGW32__ || __MINGW64__
```
JayFoxRox: Why the check for `!__MINGW32__`?

src/common/string_util.cpp
```diff
@@ -13,7 +13,7 @@
 #include "common/logging/log.h"
 #include "common/string_util.h"
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || __MINGW64__
```
JayFoxRox: why not `|| defined(__MINGW64__)` ? (Other instances too)

src/common/string_util.cpp
```diff
@@ -52,7 +52,7 @@ bool CharArrayFromFormatV(char* out, int outsize, const char* format, va_list ar
 {
     int writtenCount;
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || !defined(__MINGW64__)
```
JayFoxRox: is the negation (`!`) in `defined(__MINGW64__)` intentional? Why?
Isn't this a check for Windows platforms? Shouldn't MINGW64 qualify as such?
- doesn't this change behaviour for other / existing platforms (namely OSX and Linux)?

src/common/file_util.cpp
```diff
@@ -927,7 +928,11 @@ bool IOFile::Open(const std::string& filename, const char openmode[])
 {
     Close();
 #ifdef _WIN32
+#if defined(__MINGW64__)
+    m_file = _wfsopen(Common::UTF8ToUTF16W(filename).c_str(), Common::UTF8ToUTF16W(openmode).c_str(),SH_DENYNO);
```
JayFoxRox: space between comma and `SH_DENYNO`

src/common/string_util.cpp
```diff
@@ -323,13 +323,11 @@ std::u16string UTF8ToUTF16(const std::string& input)
 static std::wstring CPToUTF16(u32 code_page, const std::string& input)
 {
     auto const size = MultiByteToWideChar(code_page, 0, input.data(), static_cast<int>(input.size()), nullptr, 0);
-
-    std::wstring output;
+    std::wstring output(size,'\0');
```
JayFoxRox: Why this?
- output is properly resized in the line below (which is not really necessary)
- Why fill it with '\0' ?
  1. MultiByteToWideChar will fill the entire string (including zero termination if that is in the input data [which it is not I think?]). Eitherway it will get rid of all the '\0' filling
  2. Cleared if MultiByteToWideChar fails

If this is for perf then the `resize(size)` is not necessary anymore, but it's also worse than the current code probably because any modern compiler will probably optimize construction and resize

yami-hack: Well, one of the other people suggested that I do this.

mailwl:
@yami-hack, it is possible to open and compile Citra in CLion now? (i think best IDE for Win after VS)

yami-hack:
@mailwl Support generation "CDT 4 Makefiles"

src/common/string_util.cpp
lioncash: @JayFoxRox 
Constructing and then resizing one after the other is inherently worse than constructing with a known size directly (see [here](https://godbolt.org/g/SAOOuW))

also wrt using '\0', this is what resize essentially does. It fills it with null-terminators/zero.

mailwl:
@yami-hack thanks a lot, i will try it

lioncash:
By the way, the reason the `_s` functions aren't available is because the MingW build needs to define `MINGW_HAS_SECURE_API`.

src/common/string_util.cpp
```diff
@@ -324,7 +324,7 @@ static std::wstring CPToUTF16(u32 code_page, const std::string& input)
 {
     auto const size = MultiByteToWideChar(code_page, 0, input.data(), static_cast<int>(input.size()), nullptr, 0);
 
-    std::wstring output;
+	std::wstring output;
```
lioncash: This should be indented with spaces, not tabs.

lioncash:
What I had meant is that you need to define the preprocessor flag in the CMakeLists file (hence why I said CMake). Adding them to file_util is incorrect.

yami-hack:
@lioncash  I know, but I don't want to modify the CMakefileList files.

yami-hack:
@lioncash Oh, I see what you're saying,file_util.cpp does not need to be modified.CMake using the -DMINGW_HAS_SECURE_API=1

yami-hack:
@lioncash  I was wrong,CMAKE_CXX_FLAGS

lioncash:
@yami-hack 

Just change [this](https://github.com/citra-emu/citra/blob/master/CMakeLists.txt#L178) from:

```cmake
add_definitions(-D_WIN32_WINNT=0x0600)
```
to
```cmake
add_definitions(-D_WIN32_WINNT=0x0600
                -DMINGW_HAS_SECURE_API)
```
or
```cmake
add_definitions(-D_WIN32_WINNT=0x0600)
add_definitions(-DMINGW_HAS_SECURE_API)
```

lioncash:
@yami-hack Just add a 

```cmake
elseif (MINGW)
    add_definitions(-DMINGW_HAS_SECURE_API)
```

before [this](https://github.com/citra-emu/citra/blob/master/CMakeLists.txt#L75)

yami-hack:
@lioncash thank you.

lioncash:
My mistake. It can be like so:
```cmake
if (NOT MSVC)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1y -Wno-attributes")
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")

    if (ARCHITECTURE_x86_64)
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.1")
        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse4.1")
    endif()
    if (MINGW)
        add_definitions(-DMINGW_HAS_SECURE_API)
    endif()
else()
```

Sorry about that, I mistakenly thought `MINGW` also defined `MSVC` for compatibility stuff.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -14,7 +14,7 @@
 #include <map>
 #include <numeric>
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || defined(__MINGW64__)
```
jroweboy: Shouldn't these ifdefs just be `_WIN32` since `_WIN32` is used to signify that the WIN32 API is available which is really what you are trying to convey.

yami-hack: @jroweboy `#include <common/x64/abi.h>` I worked in MinGW (32bit) to build a citra,use `_win32`,I don't know what's going to happen.

jroweboy: `_WIN32` isn't related to 32bit vs 64bit, it just means that this compiler implements the windows API. Its defined for MSVC, cygwin, mingw, etc

src/core/gdbstub/gdbstub.cpp
```diff
@@ -14,7 +14,7 @@
...
 #include <WinSock2.h>
 #include <ws2tcpip.h>
 #include <common/x64/abi.h>
```
yami-hack: @jroweboy Look here,I know the definition of _win32.

yami-hack: @jroweboy  32bit using `common/x64/abi.h`?

jroweboy: Supposedly citra doesn't build on 32 bit anyway, so I didn't make a distinction about it . If you want to be correct about it why not use `_WIN64` then? At any rate, that file itself checks for `_WIN32` so i don't see what the issue with using `_WIN32` is https://github.com/citra-emu/citra/blob/master/src/common/x64/abi.h#L30

yami-hack: @jroweboy No problem, is my illusion.

src/core/gdbstub/gdbstub.cpp
LFsWang: Defined for applications for Win32 and Win64. Always defined.(MSDN)

src/common/string_util.cpp
```diff
@@ -13,7 +13,7 @@
 #include "common/logging/log.h"
 #include "common/string_util.h"
 
-#ifdef _MSC_VER
```
yuriks: Should this be `_WIN32` instead?

CMakeLists.txt
```diff
@@ -72,6 +72,10 @@ if (NOT MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.1")
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse4.1")
     endif()
+
+    if (MINGW)
+        add_definitions(-DMINGW_HAS_SECURE_API)
```
yuriks: Add a comment explaining why we're defining this

src/common/string_util.cpp
```diff
@@ -292,7 +292,7 @@ std::string ReplaceAll(std::string result, const std::string& src, const std::st
     return result;
 }
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || defined(__MINGW64__)
```
yuriks: Again, should this be `_WIN32`?

src/common/string_util.cpp
yami-hack: Some version does not support codecvt.in fact, I prefer iconv.I have little time,revocation of this PR, may make me more time to do other things.Properly modify source codes to use iconv

PR #1911, opened 2016-06-14
Kernel/IPC: Use Ports and Sessions as the fundamental building block of Inter Process Communication.

Subv:
All handles obtained via srv::GetServiceHandle or svcConnectToPort are references to ClientSessions.
Service modules will wait on the counterpart of those ClientSessions (Called ServerSessions) using svcReplyAndReceive or svcWaitSynchronization[1|N], and will be awoken when a SyncRequest is performed.

All IPC parameter translation will happen inside the implementation of svcReplyAndReceive, this is not implemented in this PR.

HLE Interfaces are now ClientPorts which override the HandleSyncRequest virtual member function to perform command handling immediately.


src/core/hle/kernel/kernel.h
```diff
@@ -36,7 +36,7 @@ enum KernelHandle : Handle {
 enum class HandleType : u32 {
     Unknown         = 0,
 
-    Session         = 2,
+
```
lioncash: Could remove the newline here (if it's not and the diff is being weird, ignore this)

src/core/hle/kernel/server_session.h
```diff
@@ -64,54 +64,65 @@ inline static u32* GetCommandBuffer(const int offset = 0) {
...
 
     /**
-     * Handles a synchronous call to this session using HLE emulation. Emulated <-> emulated calls
-     * aren't supported yet.
+     * Creates a pair of ServerSession and an associated ClientSession.
+     * @param client_port ClientPort to which the sessions are connected
+     * @param name Optional name of the ports
+     * @return The created session tuple
+     */
+    static std::tuple<SharedPtr<ServerSession>, SharedPtr<ClientSession>> CreateSessionPair(SharedPtr<ClientPort> client_port, std::string name = "Unknown");
```
lioncash: `name` can be a const reference.

src/core/hle/service/service.h
```diff
@@ -21,15 +22,25 @@
...
     // processes.
 public:
     std::string GetName() const override { return GetPortName(); }
 
+    /**
+     * Gets the maximum allowed number of sessions that can be connected to this port at the same time.
+     * It should be overwritten by each service implementation for more fine-grained control.
+     * @returns The maximum number of connections allowed.
+     */
+    virtual u32 GetMaxSessions() { return DefaultMaxSessions; }
```
lioncash: This function should likely be const qualified.

src/core/hle/kernel/kernel.h
Subv: I left the newlines there to prevent the diff from spanning the entirety of the enum.
I plan on removing the numbering (and thus the newlines) in a future PR soon.

src/core/hle/service/service.h
```diff
@@ -21,15 +22,25 @@
 namespace Service {
 
 static const int kMaxPortSize = 8; ///< Maximum size of a port name (8 characters)
+static const u32 DefaultMaxSessions = 10; ///< Arbitrary default number of maximum connections to an HLE port
 
 /// Interface to a CTROS service
-class Interface : public Kernel::Session {
+class Interface : public Kernel::ClientPort {
     // TODO(yuriks): An "Interface" being a Kernel::Object is mostly non-sense. Interface should be
```
Subv: @yuriks What direction do we want to go with this Interface class?

src/core/hle/kernel/kernel.h
yuriks: Not sure I understand why we now need *2* newlines here, but I don't care much either because of the follow up PR.

src/core/hle/kernel/kernel.h
```diff
@@ -73,7 +75,7 @@ class Object : NonCopyable {
      */
     bool IsWaitable() const {
         switch (GetHandleType()) {
-        case HandleType::Session:
+        case HandleType::ServerSession:
```
yuriks: Keep these sorted in the same order as the enum.

Subv:
I've updated the PR and decoupled Service::Interface from Kernel::ClientPort.
Interfaces are no longer kernel objects.

src/core/hle/kernel/client_port.cpp
```diff
@@ -7,10 +7,40 @@
...
+    client_port->active_sessions = 0;
+    client_port->name = hle_interface->GetPortName();
+    client_port->hle_interface = std::move(hle_interface);
+
+    return client_port;
+}
+
+void ClientPort::AddWaitingSession(SharedPtr<ServerSession> server_session) {
+    // A port that has an associated HLE interface doesn't have a server port.
+    if (hle_interface != nullptr)
```
yuriks: Did you mean == here? EIther that or the comment is wrong.

src/core/hle/kernel/client_port.cpp
```diff
@@ -7,10 +7,40 @@
...
+    // Wake the threads waiting on the ServerPort
+    server_port->WakeupAllWaitingThreads();
+}
+
+ResultCode ClientPort::HandleSyncRequest() {
+    // Forward the request to the associated HLE interface if it exists
+    if (hle_interface != nullptr)
+        return hle_interface->HandleSyncRequest();
+
+    return RESULT_SUCCESS;
```
yuriks: And if it doesn't "then return a dummy return code"

src/core/hle/kernel/client_port.h
```diff
@@ -5,30 +5,58 @@
...
+class ServerSession;
 
-class ClientPort : public Object {
+class ClientPort final : public Object {
 public:
     friend class ServerPort;
+
+    /**
+     * Creates a serverless ClientPort that represents a bridge between the HLE implementation of a service/port and the emulated application.
+     * @param max_sessions Maximum number of sessions that this port is able to handle concurrently.
```
yuriks: Add a blank line between the description and the `@param`/`@return` tags section. Same for all other comments.

src/core/hle/kernel/client_port.cpp
Subv: ?
If there's an hle_interface, then there is no server_port, so we return from the function instead of adding the session to the (non-existent) server port.

yuriks: Right, but what you're doing is returning from the function if there *isn't* a HLE interface

yuriks: Ok, you're absolutely right.

Subv:
Updated with the changes we discussed last night.

I am not sure how to handle the File and Directory classes, ideally they should inherit from the IpcHelper (currently Interface) class but i can't see any way to make that work currently, thoughts?

src/core/hle/kernel/client_session.h
```diff
@@ -41,6 +46,7 @@ class ClientSession final : public Object {
     std::string name;                           ///< Name of client port (optional)
     SharedPtr<ServerSession> server_session;    ///< The server session associated with this client session.
     SharedPtr<ClientPort> client_port;          ///< The client port which this session is connected to.
+    std::shared_ptr<Service::Interface> hle_helper = nullptr; ///< HLE implementation of this port's request handler
```
yuriks: "port"

yuriks: Also, "helper" is a strange/vague name for this, I think "hle_handler" is more obvious

src/core/hle/kernel/client_port.h
```diff
@@ -5,30 +5,52 @@
...
 
-    SharedPtr<ServerPort> server_port;          ///< ServerPort associated with this client port.
-    u32 max_sessions;                           ///< Maximum number of simultaneous sessions the port can have
-    u32 active_sessions;                        ///< Number of currently open sessions to this port
-    std::string name;                           ///< Name of client port (optional)
+    SharedPtr<ServerPort> server_port = nullptr;                 ///< ServerPort associated with this client port.
+    u32 max_sessions;                                            ///< Maximum number of simultaneous sessions the port can have
+    u32 active_sessions;                                         ///< Number of currently open sessions to this port
+    std::string name;                                            ///< Name of client port (optional)
+    std::shared_ptr<Service::Interface> hle_interface = nullptr; ///< HLE implementation of this port's request handler
```
yuriks: You don't need the nullptrs here, the default constructor already initializes to null.

src/core/hle/kernel/client_session.h
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Handle a SyncRequest from the emulated application.
+     * @return ResultCode of the operation.
+     */
+    ResultCode HandleSyncRequest();
+
+    std::string name;                           ///< Name of client port (optional)
+    SharedPtr<ServerSession> server_session;    ///< The server session associated with this client session.
+    SharedPtr<ClientPort> client_port;          ///< The client port which this session is connected to.
+    std::shared_ptr<Service::Interface> hle_helper = nullptr; ///< HLE implementation of this port's request handler
```
yuriks: hle_handler. nullptr is redundant

src/core/hle/service/fs/archive.h
```diff
@@ -46,26 +46,26 @@ enum class MediaType : u32 {
 
 typedef u64 ArchiveHandle;
 
-class File : public Kernel::Session {
+class File : public Kernel::ServerSession {
```
yuriks: These should just be HLE handlers. (ref. discussion on irc)

src/core/hle/service/service.h
```diff
@@ -21,14 +22,19 @@
...
 
 static const int kMaxPortSize = 8; ///< Maximum size of a port name (8 characters)
+static const u32 DefaultMaxSessions = 10; ///< Arbitrary default number of maximum connections to an HLE port
 
 /// Interface to a CTROS service
-class Interface : public Kernel::Session {
-    // TODO(yuriks): An "Interface" being a Kernel::Object is mostly non-sense. Interface should be
-    // just something that encapsulates a session and acts as a helper to implement service
-    // processes.
+class Interface {
```
yuriks: This type should be split in two: `HleSessionHandler`, which has only session handling related stuff, and another type inheriting from it with the service dispatch helpers.

Subv:
I'm closing this one until i feel like working on it again.

PR #1912, opened 2016-06-15
Fix AppVeyor WinSCP download

yuriks:
Seems to have been broken by some change in Sourceforge. Switched to using chocolatey instead. The downside is that the chocolatey package doesn't add `WinSCP.com` to the path. I used `WinSCP.exe` instead but that unfortunately doesn't capture the output into the terminal.


archshift:
LGTM

bunnei:
LGTM

PR #1917, opened 2016-06-17
CI: Use Qt 5 and GCC 6 on Ubuntu CI builds

lioncash:
The OSX build already uses Qt 5, so nothing needs to change here for that.

This also bumps the distribution of Ubuntu that we use up from Precise Pangolin (12.04) to Trusty Tahr (14.04). We also finally drop support for Qt 4 variants with this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ addons:
     sources:
       - ubuntu-toolchain-r-test
     packages:
-      - gcc-5
-      - g++-5
+      - gcc-6
+      - g++-6
       - qt5-default
       - libqt5opengl5-dev
       - xorg-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,16 @@
-os:
-  - linux
-  - osx
-
 language: cpp
+
+matrix:
+  include:
+    - os: linux
+      sudo: true
+      dist: trusty
+    - os: osx
+      sudo: false
 
 env:
   global:
     - secure: "AXHFIafTmbGDsHD3mUVj5a4I397DQjti/WoqAJGUp2PglxTcc04BwxZ9Z+xLuf5N2Hs5r9ojAJLT8OGxJCLBDXzneQTNSqXbFuYSLbqrEAiIRlA9eRIotWCg+wYcO+5e8MKX+cHVKwiIWasUB21AtCdq6msh6Y3pUshZp212VPg="
-
-sudo: false
 
 addons:
   apt:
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -201,16 +200,8 @@ if (ENABLE_QT)
...
-    if (NOT CITRA_FORCE_QT4)
-        find_package(Qt5 COMPONENTS Widgets OpenGL ${QT_PREFIX_HINT})
-        set(CITRA_QT_LIBS Qt5::Widgets Qt5::OpenGL)
-    endif()
-
-    if (CITRA_FORCE_QT4 OR NOT Qt5_FOUND)
-        # Try to fallback to Qt4
-        find_package(Qt4 REQUIRED COMPONENTS QtGui QtOpenGL ${QT_PREFIX_HINT})
-        set(CITRA_QT_LIBS Qt4::QtGui Qt4::QtOpenGL)
-    endif()
```
linkmauve: This should probably be in its own commit, with a more explicit summary, since some users might still be using that fallback.

.travis.yml
```diff
@@ -1,22 +1,26 @@
-os:
-  - linux
-  - osx
-
 language: cpp
 
+matrix:
+  include:
+    - os: linux
+      sudo: required
```
linkmauve: What does it require sudo for?  I don’t see any change here that would cause that change.

linkmauve:
Aside from that, LGTM!

.travis.yml
lioncash: That was a leftover from following Travis' guide for setting up the build matrix. I removed it in the revised commits.

CMakeLists.txt
lioncash: Done

bunnei:
Changes LGTM... What's with the apt-get install failed on Travis?

.travis.yml
lioncash: Actually nevermind this is required in order for some packages to be found.

lioncash:
@bunnei all good now.

.travis.yml
```diff
@@ -1,22 +1,26 @@
-os:
-  - linux
-  - osx
-
 language: cpp
 
+matrix:
+  include:
+    - os: linux
+      sudo: true
```
yuriks: Why do you need this? This puts us in Travis VMs instead of containers (at least last time I checked) which slows down builds.

jroweboy: citra is currently using the `trusty` beta environment (Ubuntu 14.04) which means you can't use the containers since they are based on ubuntu 12.04 still. Whether we need trusty is still up for debate. I don't know for sure

See https://docs.travis-ci.com/user/ci-environment/ for more info

PR #1922, opened 2016-06-19
Qt: Fix MicroProfile dpi scaling

yuriks:
This is a fix for PR #1785. The scaling was being done with the wrong value as a base, causing incorrect scale factors and unnecessary scaling in low-dpi screens. In addition, the window size wasn't being scaled, causing misalignment of screen elements.

cc @MerryMage


MerryMage:
Reply to comment on #1785: My reason for putting them in the constructor was for valid values to be in the variables after construction; but putting it in `::paintEvent` as you have done looks fine and avoids the Qt warning.

MerryMage:
Tested on my machine and LGTM.

bunnei:
LGTM

PR #1923, opened 2016-06-19
Fix recursive scanning of directories

yuriks:
ForeachDirectoryEntry didn't actually do anything with the `recursive`
parameter, and the corresponding callback parameter was shadowing the
actual recursion counters in the user functions.


bunnei:
LGTM

PR #1930, opened 2016-06-25
Remove superfluous std::move in return std::move(local_var)

scurest:
The r-value overload already gets selected, and the `std::move` technically inhibits copy elision.


yuriks:
LGTM

Issue #1694, opened 2016-04-19
Theatrhythm Final Fantasy Demo: Text scrolls outside of text border.


Leo626:
Tested with master 32dfd4b. The text is suppose to clip along the border as seen [here](https://gfycat.com/SereneEverlastingApe). I'm not sure if the log provides anything useful.

![2016-04-19_14-36-07](https://cloud.githubusercontent.com/assets/17358613/14657130/aac25dde-0659-11e6-8609-514d8d2c802e.gif)
[LOG](http://clipboard.space/clip/EByr10G6goU69be2adRv)

Something similar occurs in Fire Emblem: Awakening.
![citra-qt_2016-04-19_14-16-27](https://cloud.githubusercontent.com/assets/17358613/14657181/1cc6f066-065a-11e6-92cc-fd60b7e8653c.png)
![ffa](https://cloud.githubusercontent.com/assets/17358613/14657184/232db57a-065a-11e6-8850-009a40356c10.PNG)
[LOG](http://clipboard.space/clip/doDxB9j5J9cGJ1eZ0rqe)

Subv:
I think remember fixing this problem _somehow_, maybe it was related to #1263

Leo626:
@Subv Thanks. I tried your scissor branch and it fixed the issue.

![citra-qt_2016-04-19_19-16-28](https://cloud.githubusercontent.com/assets/17358613/14658909/6e5c02f8-0665-11e6-88b3-260531674e6c.png)
![citra-qt_2016-04-19_19-27-54](https://cloud.githubusercontent.com/assets/17358613/14658914/783b00a8-0665-11e6-8cbe-bc84243cb9c1.png)

Issue #1076, opened 2015-08-26
beyond the labyrinth - multiple graphic bugs/ untextured models


Punk34:
I just inform you that this game does not have much effect, and all models of characters and monsters have no textures:

![01](https://cloud.githubusercontent.com/assets/13983158/9496535/6390c392-4c19-11e5-807e-8f3881f6dfb5.jpg)

Subv:
Could you please test with #1049 to see if the map bug persists? thanks

Punk34:
Sorry, at the moment I am not able to compile the emulator. I'll wait for a new nightly build.
By the way, the touchscreen doesn't work well.

Punk34:
Nothing changed

![citra-qt 2015-08-27 19-22-49-59](https://cloud.githubusercontent.com/assets/13983158/9526306/3c63f37e-4cf1-11e5-87db-bb408f17d720.jpg)

yuriks:
Has #1261 fixed the missing textures?

Punk34:
no

![01](https://cloud.githubusercontent.com/assets/13983158/11588431/9c3f8c3e-9a91-11e5-8f16-21e9be98e739.jpg)

Punk34:
Oh, the аame freezes after this

![0002](https://cloud.githubusercontent.com/assets/13983158/11588586/c33d4b40-9a92-11e5-84e8-c1b11996c840.jpg)

JayFoxRox:
There have been lots of graphical fixes recently in Citra. Can you retest this game?
The map would probably be fixed by #1263

Punk34:
@JayFoxRox game is broken #1457

Subv:
Is this still an issue?

Dragios:
Please close this. It looks fine to me / fixed based on this [image](https://github.com/citra-emu/citra/issues/1457#issuecomment-214720731)

PR #1933, opened 2016-06-28
PICA: Implement scissor test

yuriks:
Nuff' said.

More efficient implementations that use glScissor or multiple rectangles in the swrast can be done later.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -115,7 +115,36 @@ struct Regs {
         BitField<24, 5, Semantic> map_w;
     } vs_output_attributes[7];
 
-    INSERT_PADDING_WORDS(0x11);
+    INSERT_PADDING_WORDS(0xe);
+
+    enum class ScissorMode : u32 {
+        Disabled = 0,
+        Exclude = 1, // Exclude pixels inside the scissor box
+
+        Include = 3 // Exclude pixels outside the scissor box
+    };
+
+    struct {
+        BitField<0, 2, ScissorMode> mode;
+
+        union {
+            BitField< 0, 16, u32> right;
+            BitField<16, 16, u32> bottom;
+        };
+
+        union {
+            BitField< 0, 16, u32> left_minus_1;
+            BitField<16, 16, u32> top_minus_1;
+        };
+
+        u32 GetTop() const {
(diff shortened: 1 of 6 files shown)


neobrain:

2016-06-28T05:18:10Z: https://github.com/citra-emu/citra/compare/9cde5cbbd39d612601df85d8792eea48c96bdc3e...yuriks:ecf6ecf32537634db15946630d62ac3bdc4fe8c9

yuriks:
I suspect shader compilation issue on AMD/Intel/Mesa, needs testing.

hrydgard:
A new OpenGL extension was just recently released that will let you fully emulate the PICA's scissor without modifying shaders.

https://www.opengl.org/registry/specs/EXT/window_rectangles.txt

It's a superset of PICA's scissor functionality.

Of course, that's nVidia only, at least for now...

linkmauve:
Mesa already has an implementation on nv50 and nvc0, and I expect other drivers to follow if users start popping around.

yuriks:
The extension looks interesting. Probably not a good idea to use it right now though, since the driver penetration will be poor for a while and it looks like supporting both it and the current fallback method would be a bit of a pain.

hrydgard:
Yeah, before supporting that it's probably worth optimizing "include mode" to use straight glScissor.

bunnei:
Code LGTM. Does this need much additional testing?

yuriks:
It should be tested on AMD/Intel GPUs to see if there's any shader errors.

yuriks:
A few games I know that use scissor that can be used to test this:
 - Fire Emblem: Awakening (and likely Fates) uses it to clip text to the text boxes
 - Labyrinth no Kanata - #1076 
 - Theathrythm Final Fantasy - #1694 
 - Apparently A Link Between Worlds also uses scissor, but I think it has no visible effect. Should still be usable for checking for a shader error though.

Leo626:
This fixes #1694 in addition to Kingdom Hearts 3D. Tested with Nvidia GPU.

fotorcreated

wwylele:
Tested Kindom Herats 3D on Intel and AMD GPU. Both looks fine :+1:

yuriks:
I guess that settles it enough.

PR #1935, opened 2016-06-30
Result: fix and update ErrorModule

wwylele:
According to https://www.3dbrew.org/wiki/Error_codes. I am curious where did the `RO_2` come from (which even broke all codes after it)


neobrain:

2016-06-30T08:30:02Z: https://github.com/citra-emu/citra/compare/de79b6fc4857fde1df3c8a4a2b38d231f2cc961d...wwylele:e023a4cfd033c9c82bf2821b51a2e83930b6048d

Subv:
LGTM

JamePeng:
LGTM

Issue #1931, opened 2016-06-26
Unable to build nightly for maybe a week


John-Gee:
Hello,

I'm using Arch Linux with GCC6 and I have not been able to build citra git lately without hacking a certain file.

By default I get errors like these:

src/citra/src/citra_qt/configure_audio.cpp: In constructor ‘ConfigureAudio::ConfigureAudio(QWidget*)’:
src/citra/src/citra_qt/configure_audio.cpp:14:12: error: ‘make_unique’ is not a member of ‘std’
         ui(std::make_unique<Ui::ConfigureAudio>();"

Removing -std=gnu++11  in src/citra_qt/CMakeFiles/citra-qt.dir/flags.make fixes that.
I am not sure what generates that file and why there is -std=c++1y and then -std=gnu++11.

relevant line:
"CXX_FLAGS = -march=native -O2 -pipe -fstack-protector-strong -fstack-check -flto=8  -std=c++1y -Wno-attributes -msse4.1 -O3 -DNDEBUG -fPIC -pthread -std=gnu++11"

I'd be happy to test anything you can think of.
In case it helps, my system CXX_FLAGS is:
"-march=native -O2 -pipe -fstack-protector-strong -fstack-check -flto=$(nproc)"

Enverex:
Ditto, the compilation worked a week ago but now fails with that error.

```
/src/citra/src/citra_qt/configure_audio.cpp: In constructor ‘ConfigureAudio::ConfigureAudio(QWidget*)’:
/src/citra/src/citra_qt/configure_audio.cpp:14:12: error: ‘make_unique’ is not a member of ‘std’
         ui(std::make_unique<Ui::ConfigureAudio>())
            ^~~
/src/citra/src/citra_qt/configure_audio.cpp:14:47: error: expected primary-expression before ‘>’ token
         ui(std::make_unique<Ui::ConfigureAudio>())
                                               ^
/src/citra/src/citra_qt/configure_audio.cpp:14:49: error: expected primary-expression before ‘)’ token
         ui(std::make_unique<Ui::ConfigureAudio>())
                                                 ^
make[2]: *** [src/citra_qt/CMakeFiles/citra-qt.dir/build.make:487: src/citra_qt/CMakeFiles/citra-qt.dir/configure_audio.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[...]

PR #1936, opened 2016-06-30
CMake: Fix for QT 5.7 overwriting -std=c++1y flag

jroweboy:
Qt 5.7 added a new check for c++11 compilers (see https://github.com/citra-emu/citra/issues/1931#issuecomment-229549577 for more information) which meant that because citra was setting the c++1y flag directly, cmake assumed that it needed to set c++11 on any target that linked to Qt. This broke the build whenever a c++14 feature was used on targets that linked against Qt. This change specifies that C++14 compilers are required, and it adds the -std=c++14 (or -std=gnu++14 for builds in gcc) flag on every build target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,5 @@
-# CMake 3.1 required for Qt5 settings to be applied automatically on
-# dependent libraries and IMPORTED targets.
-cmake_minimum_required(VERSION 3.1)
+# CMake 3.2 required for cmake to know the right flags for CXX standard on OSX
+cmake_minimum_required(VERSION 3.2)
 
 function(download_bundled_external remote_path lib_name prefix_var)
     set(prefix "${CMAKE_BINARY_DIR}/externals/${lib_name}")
@@ -63,8 +62,11 @@ if (NOT DEFINED ARCHITECTURE)
 endif()
 message(STATUS "Target architecture: ${ARCHITECTURE}")
 
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
 if (NOT MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1y -Wno-attributes")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-attributes")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
 else()
     # Silence "deprecation" warnings
--- a/.travis-deps.sh
+++ b/.travis-deps.sh
@@ -9,7 +9,7 @@ if [ "$TRAVIS_OS_NAME" = "linux" -o -z "$TRAVIS_OS_NAME" ]; then
     export CXX=g++-6
     mkdir -p $HOME/.local
 
-    curl -L http://www.cmake.org/files/v3.1/cmake-3.1.0-Linux-i386.tar.gz \
+    curl -L http://www.cmake.org/files/v3.2/cmake-3.2.0-Linux-i386.tar.gz \
         | tar -xz -C $HOME/.local --strip-components=1
 
     (
@@ -21,6 +21,6 @@ if [ "$TRAVIS_OS_NAME" = "linux" -o -z "$TRAVIS_OS_NAME" ]; then
 elif [ "$TRAVIS_OS_NAME" = "osx" ]; then
     brew update > /dev/null # silence the very verbose output
     brew unlink cmake
-    brew install cmake31 qt5 sdl2 dylibbundler
+    brew install cmake qt5 sdl2 dylibbundler
     gem install xcpretty
 fi


neobrain:

2016-06-30T18:10:19Z: https://github.com/citra-emu/citra/compare/de79b6fc4857fde1df3c8a4a2b38d231f2cc961d...jroweboy:eff9824280a79c9ba270912f6222d0daf64b219a

CMakeLists.txt
```diff
@@ -47,6 +47,10 @@ if(NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git/hooks/pre-commit)
         DESTINATION ${CMAKE_CURRENT_SOURCE_DIR}/.git/hooks)
 endif()
 
+# Force cmake to use c++14 which prevents external libraries from requesting lower versions and overwriting the version citra needs
+set(CITRA_CXX_STANDARD 14)
```
lioncash: Not a blocking comment, but have you tried specifying `-std=c++1y` through [target_compile_features](https://cmake.org/cmake/help/v3.2/command/target_compile_features.html)?

jroweboy: That function take in one of these https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html and not a direct `-std=c++1y` sadly. And I'm not sure which one we should say that citra requires on the list since I don't know which c++14 feature is getting used besides make_unique. (If you have a suggestion of which feature on the list citra uses I'd be glad to make the change)

lioncash: Ah, I had updated my comment shortly after posting it (unfortunately github doesn't update comments for others currently viewing :<), so you might not have seen the add_compile_options suggestion. Sorry about that.

lioncash: There's also [target_compile_options](https://cmake.org/cmake/help/v3.0/command/target_compile_options.html) as well.

jroweboy: So I just tested them and neither of those fix the issue. The issue is that because we aren't specifying anywhere in cmake the `CXX_STANDARD`, when cmake comes across the `target_compile_features` line in Qt5.7 that tells cmake it requires c++11, cmake adds `-std=c++11` on the end whenever `target_link_libraries` is called with the target getting linked to Qt. So despite setting `-std=c++1y` manually in many different ways, cmake still thinks we haven't set a std flag unless you explicitly set the property `CXX_STANDARD` on the target. 

Now with that said, it may be possible to try to overwrite whatever qt has cmake add in by setting CXX_FLAGS after `target_link_libraries` and that might work just fine, but that just means that the actual compile command still contains two -std= flags, it just so happens that the one we want is last. I don't know if thats possible, but it seems more hackish that just setting the `CXX_STANDARD` property on each target.

lioncash: Wouldn't setting `CMAKE_CXX_STANDARD` (as opposed to `CITRA_CXX_STANDARD`) to 14 define that as the default level for all targets, which targets can override on their own, or does Qt literally steamroll everything regarding that?

neobrain:

2016-06-30T19:59:58Z: https://github.com/citra-emu/citra/compare/de79b6fc4857fde1df3c8a4a2b38d231f2cc961d...jroweboy:227124df85e075fb7e6eaf4428021995ddda387b

jroweboy:
I knew I was doing something wrong. Thanks for the catch there @lioncash

yuriks:
LGTM

wwylele:
Why travis build it (looks like) as C++03 ? :confused:

yuriks:
Who knows. :| Can someone with OS X test, since the Travis log doesn't show the commandline flags.
I suspect it's just an older version of CMake (This sort of thing is exactly why I think this flag thing is a terrible "feature" for CMake to have.)

CMakeLists.txt
```diff
@@ -63,8 +63,11 @@ if (NOT DEFINED ARCHITECTURE)
 endif()
 message(STATUS "Target architecture: ${ARCHITECTURE}")
 
+set(CMAKE_CXX_STANDARD 14)
+set(CXX_STANDARD_REQUIRED ON)
```
MerryMage: Should likely be [`CMAKE_CXX_STANDARD_REQUIRED`](https://cmake.org/cmake/help/v3.1/variable/CMAKE_CXX_STANDARD_REQUIRED.html#variable:CMAKE_CXX_STANDARD_REQUIRED) instead.

yuriks:
If it doesn't work on 3.1.x, then we should probably bump the required CMake version at the top of the file.

jroweboy:
I'm going to try out what merry suggested tonight and see if that was the
real issue. If not I'll bump the version as well

neobrain:

2016-08-04T22:41:45Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...jroweboy:a30bb2670859923c2ab41e52960b89719a7ca7b7

neobrain:

2016-08-04T22:57:48Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...jroweboy:c35e24e76c2f1bb3bfe4e47dfac449acba3ab53f

neobrain:

2016-08-04T23:03:53Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...jroweboy:11bbe9146bed92d667bfc63bbb784af2ab67f658

neobrain:

2016-08-04T23:16:27Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...jroweboy:a976c2e3ba146df2fda9293c6fce9e11d96ac038

jroweboy:
There we go. Needs both CMAKE 3.2 and the standard set. Ready to go!

linkmauve:
Tested on ArchLinux, it fixes the build of citra-qt.
And LGTM.

PR #1947, opened 2016-07-13
[RFC] Add a V4L2-based cam:U implementation

linkmauve:
It uses the raw Video For Linux 2 API to integrate any capture device
(a camera for example) into the 3DS cam:U subsystem, so that games can
now take an actual picture of the user, use the AR system, or do
basically anything they wants with the video stream.

Conversion to RGB isn’t supported yet, so in the very likely chance
your capture device doesn’t support outputting directly in RGB you will
get a garbled but recognizable image resulting from the casting of YUYV
data into RGB565.

Only the OUT1 camera (rear left) is currently supported, every other
one will still return an empty buffer.  A possible improvement would be
to extend support to more than one /dev/videoN device, and let the user
configure which physical device corresponds to which emulated camera.
Another would be to use a computer vision library to generate a shifted
image and give the illusion of perspective when using both read
cameras.

A possible improvement would be to check for the presence of the v4l2
header instead of matching on the OS name.  A dummy implementation is
built on every other OS.

This also adds the SetOutputFormat call.

None of this has been tested on hardware yet, nor reversed from the
cam:U service, everything is pure guess work based on the camera/video
example from 3ds-examples.

### DIFF
--- a/src/core/hle/service/cam/cam.cpp
+++ b/src/core/hle/service/cam/cam.cpp
@@ -5,11 +5,13 @@
 #include "common/logging/log.h"
 
 #include "core/hle/kernel/event.h"
+#include "core/hle/result.h"
 #include "core/hle/service/cam/cam.h"
 #include "core/hle/service/cam/cam_c.h"
 #include "core/hle/service/cam/cam_q.h"
 #include "core/hle/service/cam/cam_s.h"
 #include "core/hle/service/cam/cam_u.h"
+#include "core/hle/service/cam/impl.h"
 #include "core/hle/service/service.h"
 
 namespace Service {
@@ -27,10 +29,12 @@ void StartCapture(Service::Interface* self) {
 
     u8 port = cmd_buff[1] & 0xFF;
 
-    cmd_buff[0] = IPC::MakeHeader(0x1, 1, 0);
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    ResultCode result = StartCapture(port);
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, port=%d", port);
+    cmd_buff[0] = IPC::MakeHeader(0x1, 1, 0);
+    cmd_buff[1] = result.raw;
+
+    LOG_DEBUG(Service_CAM, "called, port=%d", port);
 }
 
 void StopCapture(Service::Interface* self) {
@@ -38,10 +42,12 @@ void StopCapture(Service::Interface* self) {
 
     u8 port = cmd_buff[1] & 0xFF;
 
-    cmd_buff[0] = IPC::MakeHeader(0x2, 1, 0);
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    ResultCode result = StopCapture(port);
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, port=%d", port);
+    cmd_buff[0] = IPC::MakeHeader(0x2, 1, 0);
+    cmd_buff[1] = result.raw;
+
+    LOG_DEBUG(Service_CAM, "called, port=%d", port);
 }
 
 void GetVsyncInterruptEvent(Service::Interface* self) {
@@ -78,18 +84,20 @@ void SetReceiving(Service::Interface* self) {
     u32 image_size = cmd_buff[3];
     u16 trans_unit = cmd_buff[4] & 0xFFFF;
 
+    ResultCode result = SetReceiving(dest, port, image_size, trans_unit);
+
     Kernel::Event* completion_event = (Port)port == Port::Cam2 ?
             completion_event_cam2.get() : completion_event_cam1.get();
 
     completion_event->Signal();
 
     cmd_buff[0] = IPC::MakeHeader(0x7, 1, 2);
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    cmd_buff[1] = result.raw;
     cmd_buff[2] = IPC::MoveHandleDesc();
     cmd_buff[3] = Kernel::g_handle_table.Create(completion_event).MoveFrom();
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, addr=0x%X, port=%d, image_size=%d, trans_unit=%d",
-            dest, port, image_size, trans_unit);
+    LOG_DEBUG(Service_CAM, "called, addr=0x%X, port=%d, image_size=%d, trans_unit=%d",
+              dest, port, image_size, trans_unit);
 }
 
 void SetTransferLines(Service::Interface* self) {
@@ -166,11 +174,13 @@ void Activate(Service::Interface* self) {
 
     u8 cam_select = cmd_buff[1] & 0xFF;
 
-    cmd_buff[0] = IPC::MakeHeader(0x13, 1, 0);
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    ResultCode result = Activate(cam_select);
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, cam_select=%d",
-            cam_select);
+    cmd_buff[0] = IPC::MakeHeader(0x13, 1, 0);
+    cmd_buff[1] = result.raw;
+
+    LOG_DEBUG(Service_CAM, "called, cam_select=%d",
+              cam_select);
 }
 
 void FlipImage(Service::Interface* self) {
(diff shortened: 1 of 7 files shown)


src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+namespace Service {
+namespace CAM {
+
+// TODO(Link Mauve): make the video device configurable.
+static const char *video_device = "/dev/video0";
+static const unsigned NUM_BUFFERS = 2;
+
+static int fd;
+static enum v4l2_buf_type format_type;
+static struct v4l2_format format;
```
lioncash: `struct` and `enum` aren't necessary in C++.

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+#include "core/hle/result.h"
+#include "core/hle/service/cam/cam.h"
+#include "core/hle/service/cam/impl.h"
+
+namespace Service {
+namespace CAM {
+
+// TODO(Link Mauve): make the video device configurable.
+static const char *video_device = "/dev/video0";
+static const unsigned NUM_BUFFERS = 2;
```
lioncash: constexpr

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+    return RESULT_SUCCESS;
+}
+
+ResultCode SetReceiving(VAddr dest, u8 port, u32 image_size, u16 trans_unit) {
+    // XXX: hack stereoscopy somehow.
+    if (!(port & 1))
+        return ResultCode(ErrorDescription::NotImplemented, ErrorModule::CAM,
+                          ErrorSummary::NotSupported, ErrorLevel::Info);
+
+    struct v4l2_buffer buf;
```
lioncash: `struct` isn't necessary here. Same for other misc applicable occurrences below.

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+        // TODO(Link Mauve): find the proper error code.
+        return ResultCode(ErrorDescription::InvalidResultValue, ErrorModule::CAM,
+                          ErrorSummary::Internal, ErrorLevel::Fatal);
+    }
+
+    ASSERT(buf.flags & V4L2_BUF_FLAG_DONE);
+
+    unsigned index = buf.index;
+    ASSERT(index < NUM_BUFFERS);
+    const char *memory = mapped_buffers[index].memory;
```
lioncash: asterisk against type.

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+static int fd;
+static enum v4l2_buf_type format_type;
+static struct v4l2_format format;
+static u32 width;
+static u32 height;
+static bool crop_image_horizontally;
+
+static struct {
+    char *memory;
```
lioncash: asterisk against type.

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+#include "core/hle/result.h"
+#include "core/hle/service/cam/cam.h"
+#include "core/hle/service/cam/impl.h"
+
+namespace Service {
+namespace CAM {
+
+// TODO(Link Mauve): make the video device configurable.
+static const char *video_device = "/dev/video0";
```
lioncash: asterisk against type (this can also be made constexpr or an array of char to only store the text, rather than the text and a pointer address)

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+    ASSERT(buf.flags & V4L2_BUF_FLAG_DONE);
+
+    unsigned index = buf.index;
+    ASSERT(index < NUM_BUFFERS);
+    const char *memory = mapped_buffers[index].memory;
+
+    // Not every capture device supports the size requested.
+    if (crop_image_horizontally) {
+        // There are always two bytes per pixel, both for YUYV and for RGB565.
+        const unsigned bytes_per_pixel = 2;
```
lioncash: constexpr

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+        for (int y = 0; y < height; ++y)
+            Memory::WriteBlock(dest + (width * bytes_per_pixel * y),
+                               static_cast<const char*>(memory) + (format.fmt.pix.bytesperline * y),
```
lioncash: `memory` is already a `const char*` above; is this a leftover?

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+        LOG_WARNING(Service_CAM, "Got a higher format than requested (%d > %d), will crop",
+                    format.fmt.pix.height, height);
+
+    return RESULT_SUCCESS;
+}
+
+static inline const char * dump_format(uint32_t format, char out[4])
+{
+#if BYTE_ORDER == BIG_ENDIAN
+    format = __builtin_bswap32(format);
```
lioncash: Perhaps utilize `common/swap.h` byteswapping functions for this.

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    if (format.fmt.pix.height > height)
+        LOG_WARNING(Service_CAM, "Got a higher format than requested (%d > %d), will crop",
+                    format.fmt.pix.height, height);
+
+    return RESULT_SUCCESS;
+}
+
+static inline const char * dump_format(uint32_t format, char out[4])
```
lioncash: - inline isn't necessary here
- asterisk against type
- u32
- If you want to represent an array of fixed size and have it be enforced, it should be `char (&out)[4]`
  - Since this is only ever used by return value, it may be better to just return a `std::array` as opposed to using an out parameter for this.

src/core/hle/service/cam/impl_linux.cpp
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    // TODO(Link Mauve): find the default format on a 3DS and set it here.
+
+    return RESULT_SUCCESS;
+}
+
+ResultCode DriverFinalize() {
+    // TODO(Link Mauve): maybe this should be done in Activate(0) instead.
+    for (unsigned i = 0; i < NUM_BUFFERS; ++i) {
```
lioncash: Since this is iterating through a memory buffers, size_t is more ideal here. Or, if you give the struct representing `mapped_buffers` a type name, you can use a ranged-for loop here.

src/core/hle/service/cam/impl.h
```diff
@@ -0,0 +1,26 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+
+#include "core/hle/result.h"
```
lioncash: `ResultCode` can just be a forward declaration.

src/core/hle/service/cam/cam.cpp
```diff
@@ -78,18 +84,20 @@ void SetReceiving(Service::Interface* self) {
     u32 image_size = cmd_buff[3];
     u16 trans_unit = cmd_buff[4] & 0xFFFF;
 
+    ResultCode result = SetReceiving(dest, port, image_size, trans_unit);
+
     Kernel::Event* completion_event = (Port)port == Port::Cam2 ?
             completion_event_cam2.get() : completion_event_cam1.get();
 
     completion_event->Signal();
```
wwylele: This event shouldn't be signaled here anymore, and the data receiving should be asynchronous. Current approach causes Pokemon Super Mystery Dungeon runs at 9 seconds per frame.

wwylele:

JayFoxRox:
As this was an RFC, and you got at least some comments (unfortunately not too many about the design, which I think was your intention with the RFC?), I think we should close this. You can move this to your personal todo list and then send a new PR when it's merge ready?

This is already a couple of months old and I believe if we keep this opened it will stay here forever - even if you completed work on it, it would not get reviewed.

linkmauve:
(Damn github still failing at sending email notifications…)

Anyway, my intention was indeed to gather input about the design, especially on when to poll for transfer completion, and how to handle multiple cameras when only a single one is available (which I guess will be the case most of the time). I will rework some of that at some point and repoen this PR once I am happy with the direction.

PR #1950, opened 2016-07-14
Correct APT::0x00550040 and APT::0x00560000 function

JamePeng:
Some guy corrected the old and confusing description, so I need to fix my old wrong code :smiley: 

https://www.3dbrew.org/wiki/APT:GetScreencapPostPermission
https://www.3dbrew.org/wiki/APT:SetScreencapPostPermission

### DIFF
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -37,6 +37,8 @@ static u32 cpu_percent; ///< CPU time available to the running application
 // APT::CheckNew3DSApp will check this unknown_ns_state_field to determine processing mode
 static u8 unknown_ns_state_field;
 
+static ScreencapPostPermission screen_capture_post_permission;
+
 /// Parameter data to be returned in the next call to Glance/ReceiveParameter
 static MessageParameter next_parameter;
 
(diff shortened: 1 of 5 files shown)


src/core/hle/service/apt/apt.cpp
```diff
@@ -35,7 +35,9 @@ static Kernel::SharedPtr<Kernel::Event> parameter_event; ///< APT parameter even
 static u32 cpu_percent; ///< CPU time available to the running application
 
 // APT::CheckNew3DSApp will check this unknown_ns_state_field to determine processing mode
-static u8 unknown_ns_state_field;
+static u8 unknown_ns_state_field = 0;
```
JayFoxRox: This is already initialized in `Init`?

I'm guessing this should be set to zero on every reboot of the virtual 3DS?
If so the initialization here is probably useless and should be done elsewhere.

src/core/hle/service/apt/apt.h
```diff
@@ -94,6 +94,13 @@ enum class StartupArgumentType : u32 {
     OtherMedia = 2,
 };
 
+enum class ScreencapPostPermission : u32 {
+    CleanThePermission                 = 0,
```
Subv: What does CleanThePermission mean?

JamePeng: qq 20160715004024

The SetScreencapPostPermission first call will set to zero

Subv: Yeah but what does it mean? And how is it different than screen_capture_post_permission=1?
That needs to be either documented in the code or left as a todo.

JamePeng: Add a todo will be better to notice this enum value, but we don't know what is the inital value it is, and the zero will be more safer than [screen_capture_post_permission=1].

bunnei:
LGTM

PR #1951, opened 2016-07-14
Emulate motion sensor in frontend

wwylele:
As a continuation of #1549.

This is a relatively old branch of mine, which still contains problems. But I want to PR this before it  becomes completely stale, and discuss some questions below.

Usage: move mouse in the emulation window while holding the right button.

It emulates the accelerometer and gyroscope in a very limited range: it can only tilt in the direction of mouse movement, no rotate about the vertical (y) axis, and no linear acceleration. I also limited the tilt angle to be less than 90 degree, because I received reports saying that freeing the range to 180 degree will confuse the user (I also felt that when testing).

I don't have many games using motion sensor, so not sure if there is other common motion should be emulated.

In `EmuWindow`, gyroscope data is "stretched" according to current running speed, because in the view of slow-paced game, the motion of the "3DS" is faster. A similar time stretching should be also applied to the accelerometer, but the formula is much more complicated, and it is meaningless at current stage because only gravity are emulated (which won't be stretched) without linear acceleration(which should be stretched).

**QUESTIONS**
- Is it a good approach to create a thread to deal with sensors data?
- ~~Is PI defined somewhere in our code base? If not, where should I put it?~~ moved to `math_util.h`
- how to measure the running speed? This is needed for doing time stretching. Currently I use the profiler to get the FPS as the speed. Is it good? What's the better way?
- the [updating time interval](https://github.com/wwylele/citra/pull/2/files#diff-58315aeb8808393b7ed57620acb3b8bfR47) and the [sensitivity coefficient](https://github.com/wwylele/citra/pull/2/files#diff-58315aeb8808393b7ed57620acb3b8bfR47) are chosen arbitrarily. Is it good? Or I should put them to settings?

### DIFF
--- a/src/common/math_util.h
+++ b/src/common/math_util.h
@@ -10,6 +10,8 @@
 
 namespace MathUtil {
 
+static constexpr float PI = 3.14159265f;
+
 inline bool IntervalsIntersect(unsigned start0, unsigned length0, unsigned start1,
                                unsigned length1) {
     return (std::max(start0, start1) < std::min(start0 + length0, start1 + length1));
--- a/src/common/thread.h
+++ b/src/common/thread.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <chrono>
 #include <condition_variable>
 #include <cstddef>
 #include <mutex>
@@ -52,6 +53,15 @@ public:
         std::unique_lock<std::mutex> lk(mutex);
         condvar.wait(lk, [&] { return is_set; });
         is_set = false;
+    }
+
+    template <class Clock, class Duration>
+    bool WaitUntil(const std::chrono::time_point<Clock, Duration>& time) {
+        std::unique_lock<std::mutex> lk(mutex);
+        if (!condvar.wait_until(lk, time, [this] { return is_set; }))
+            return false;
+        is_set = false;
+        return true;
     }
 
     void Reset() {
--- a/src/core/frontend/emu_window.cpp
+++ b/src/core/frontend/emu_window.cpp
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <cmath>
 #include "common/assert.h"
+#include "common/profiler_reporting.h"
 #include "core/frontend/emu_window.h"
 #include "core/frontend/key_map.h"
 #include "video_core/video_core.h"
@@ -87,6 +88,27 @@ void EmuWindow::TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y) {
         std::tie(framebuffer_x, framebuffer_y) = ClipToTouchScreen(framebuffer_x, framebuffer_y);
 
     TouchPressed(framebuffer_x, framebuffer_y);
+}
+
+void EmuWindow::AccelerometerChanged(float x, float y, float z) {
+    constexpr float coef = 512;
+
+    // TODO(wwylele): do a time stretch as it in GyroscopeChanged
+    // The time stretch formula should be like
+    // stretched_vector = (raw_vector - gravity) * stretch_ratio + gravity
+    accel_x = x * coef;
+    accel_y = y * coef;
+    accel_z = z * coef;
+}
+
+void EmuWindow::GyroscopeChanged(float x, float y, float z) {
+    constexpr float FULL_FPS = 60;
+    float coef = GetGyroscopeRawToDpsCoefficient();
+    float stretch =
+        FULL_FPS / Common::Profiling::GetTimingResultsAggregator()->GetAggregatedResults().fps;
+    gyro_x = x * coef * stretch;
+    gyro_y = y * coef * stretch;
+    gyro_z = z * coef * stretch;
 }
 
 void EmuWindow::UpdateCurrentFramebufferLayout(unsigned width, unsigned height) {
--- a/src/core/frontend/emu_window.cpp
+++ b/src/core/frontend/emu_window.cpp
@@ -93,6 +93,8 @@ void EmuWindow::TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y) {
 void EmuWindow::AccelerometerChanged(float x, float y, float z) {
     constexpr float coef = 512;
 
+    std::lock_guard<std::mutex> lock(accel_mutex);
+
     // TODO(wwylele): do a time stretch as it in GyroscopeChanged
     // The time stretch formula should be like
     // stretched_vector = (raw_vector - gravity) * stretch_ratio + gravity
@@ -106,6 +108,7 @@ void EmuWindow::GyroscopeChanged(float x, float y, float z) {
     float coef = GetGyroscopeRawToDpsCoefficient();
     float stretch =
         FULL_FPS / Common::Profiling::GetTimingResultsAggregator()->GetAggregatedResults().fps;
+    std::lock_guard<std::mutex> lock(gyro_mutex);
     gyro_x = x * coef * stretch;
     gyro_y = y * coef * stretch;
     gyro_z = z * coef * stretch;
--- a/src/citra/emu_window/emu_window_sdl2.cpp
+++ b/src/citra/emu_window/emu_window_sdl2.cpp
@@ -19,16 +19,22 @@
 
 void EmuWindow_SDL2::OnMouseMotion(s32 x, s32 y) {
     TouchMoved((unsigned)std::max(x, 0), (unsigned)std::max(y, 0));
+    motion_emu->Tilt(x, y);
 }
 
 void EmuWindow_SDL2::OnMouseButton(u32 button, u8 state, s32 x, s32 y) {
-    if (button != SDL_BUTTON_LEFT)
-        return;
-
-    if (state == SDL_PRESSED) {
-        TouchPressed((unsigned)std::max(x, 0), (unsigned)std::max(y, 0));
-    } else {
-        TouchReleased();
+    if (button == SDL_BUTTON_LEFT) {
+        if (state == SDL_PRESSED) {
+            TouchPressed((unsigned)std::max(x, 0), (unsigned)std::max(y, 0));
+        } else {
+            TouchReleased();
+        }
+    } else if (button == SDL_BUTTON_RIGHT) {
+        if (state == SDL_PRESSED) {
+            motion_emu->BeginTilt(x, y);
+        } else {
+            motion_emu->EndTilt();
+        }
     }
 }
 
@@ -54,6 +60,7 @@ EmuWindow_SDL2::EmuWindow_SDL2() {
     keyboard_id = KeyMap::NewDeviceId();
 
     ReloadSetKeymaps();
+    motion_emu = std::make_unique<Motion::MotionEmu>(*this);
 
     SDL_SetMainReady();
 
@@ -109,6 +116,7 @@ EmuWindow_SDL2::EmuWindow_SDL2() {
 EmuWindow_SDL2::~EmuWindow_SDL2() {
     SDL_GL_DeleteContext(gl_context);
     SDL_Quit();
+    motion_emu = nullptr;
 }
 
 void EmuWindow_SDL2::SwapBuffers() {
--- a/src/citra/emu_window/emu_window_sdl2.h
+++ b/src/citra/emu_window/emu_window_sdl2.h
@@ -4,8 +4,10 @@
 
 #pragma once
 
+#include <memory>
 #include <utility>
 #include "core/frontend/emu_window.h"
+#include "core/frontend/motion_emu.h"
 
 struct SDL_Window;
 
@@ -61,4 +63,7 @@ private:
 
     /// Device id of keyboard for use with KeyMap
     int keyboard_id;
+
+    /// Motion sensors emulation
+    std::unique_ptr<Motion::MotionEmu> motion_emu;
 };
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -191,6 +191,7 @@ qreal GRenderWindow::windowPixelRatio() {
 }
 
 void GRenderWindow::closeEvent(QCloseEvent* event) {
+    motion_emu = nullptr;
     emit Closed();
     QWidget::closeEvent(event);
 }
@@ -204,11 +205,13 @@ void GRenderWindow::keyReleaseEvent(QKeyEvent* event) {
 }
 
 void GRenderWindow::mousePressEvent(QMouseEvent* event) {
+    auto pos = event->pos();
     if (event->button() == Qt::LeftButton) {
-        auto pos = event->pos();
         qreal pixelRatio = windowPixelRatio();
         this->TouchPressed(static_cast<unsigned>(pos.x() * pixelRatio),
                            static_cast<unsigned>(pos.y() * pixelRatio));
+    } else if (event->button() == Qt::RightButton) {
+        motion_emu->BeginTilt(pos.x(), pos.y());
     }
 }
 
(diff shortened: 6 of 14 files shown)


src/common/emu_window.cpp
```diff
@@ -91,6 +92,25 @@ void EmuWindow::TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y) {
...
+    // The time stretch formula should be like
+    // stretched_vector = (raw_vector - gravity) * stretch_ratio + gravity
+    accel_x = x * coef;
+    accel_y = y * coef;
+    accel_z = z * coef;
+}
+
+void EmuWindow::GyroscopeChanged(float x, float y, float z) {
+    float coef = GetGyroscopeRawToDpsCoefficient();
+    float stretch = 60 / Common::Profiling::GetTimingResultsAggregator()->GetAggregatedResults().fps;
```
lioncash: `60` should be a named constant.

src/common/thread.h
```diff
@@ -55,6 +56,14 @@ class Event {
         is_set = false;
     }
 
+    template <class Clock, class Duration>
+    bool WaitUntil(const std::chrono::time_point<Clock, Duration>& time) {
+        std::unique_lock<std::mutex> lk(mutex);
+        if (!condvar.wait_until(lk, time, [&]{ return is_set; })) return false;
```
lioncash: the last return should go on a new line. Also, the capture should be `[this]`, rather than `[&]`

src/common/motion_emu.h
```diff
@@ -0,0 +1,41 @@
...
+ */
+void Init(EmuWindow& emu_window);
+
+/**
+ * Stops the motion emulation thread.
+ */
+void Shutdown();
+
+/**
+ * Signals the user begins tilting (e.g. the right button was pressed).
```
lioncash: `Signals that a motion sensor tilt has begun`

src/common/motion_emu.h
```diff
@@ -0,0 +1,41 @@
...
+ * @param x the x-coordinate of the cursor
+ * @param y the y-coordinate of the cursor
+ */
+void BeginTilt(int x, int y);
+
+/**
+ * Signals the user tilts (e.g. mouse was moved over the emu window).
```
lioncash: `Signals that a motion sensor tilt is occurring`

src/common/motion_emu.h
```diff
@@ -0,0 +1,41 @@
...
+ * @param x the x-coordinate of the cursor
+ * @param y the y-coordinate of the cursor
+ */
+void Tilt(int x, int y);
+
+/**
+ * Signals the user stops tilting (e.g. the right button was released).
```
lioncash: `Signals that a motion sensor tilt has ended`.

src/common/motion_emu.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+static std::mutex tilt_mutex;
+static Math::Vec2<float> tilt_direction;
+static float tilt_angle;
+
+static bool is_tilting;
+
+static void MotionEmuThread(EmuWindow& emu_window) {
+    auto update_time = std::chrono::steady_clock::now();
+    Quaternion q = MakeQuaternion(Math::Vec3<float>(), 0) , old_q;
```
lioncash: `old_q` should be declared on its own line.

src/common/motion_emu.cpp
```diff
@@ -0,0 +1,135 @@
...
+    Quaternion Inverse() {
+        return Quaternion { -xyz, w };
+    }
+
+    Quaternion operator- (const Quaternion& a) {
+        return Quaternion { xyz - a.xyz, w - a.w };
+    }
+};
+
+inline Quaternion Product(const Quaternion& a, const Quaternion& b) {
```
lioncash: These should be static, rather than inline. If they were in a header, then inline would make more sense.

src/common/motion_emu.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include "common/emu_window.h"
+#include "common/math_util.h"
+#include "common/motion_emu.h"
+#include "common/thread.h"
+#include "common/vector_math.h"
+
+namespace MotionEmu {
+
+struct Quaternion {
+    Math::Vec3<float> xyz;
```
lioncash: These member variables should likely default to zero for a valid initial state. Which will still allow for treating it as an aggregate.

src/common/motion_emu.cpp
```diff
@@ -0,0 +1,135 @@
...
+    float w;
+
+    Quaternion Inverse() {
```
lioncash:
```cpp
Quaternion Inverse() const {
```

src/common/motion_emu.cpp
```diff
@@ -0,0 +1,135 @@
...
+        return Quaternion { -xyz, w };
+    }
+
+    Quaternion operator- (const Quaternion& a) {
```
lioncash:
```cpp
Quaternion operator- (const Quaternion& a) const {
```

lioncash: It's also somewhat odd that only `operator-` is provided for Quaternions. This should likely have the other basic arithmetic operators (+, /, *)

src/common/motion_emu.cpp
```diff
@@ -0,0 +1,135 @@
...
+        emu_window.AccelerometerChanged(gravity.x, gravity.y, gravity.z);
+        emu_window.GyroscopeChanged(angular_rate.x, angular_rate.y, angular_rate.z);
+    }
+}
+
+void Init(EmuWindow& emu_window) {
+    Shutdown();
+    tilt_angle = 0;
+    is_tilting = false;
+    motion_emu_thread = std::make_unique<std::thread>(MotionEmuThread, std::ref(emu_window));
```
lioncash: Why is the thread instance a unique_ptr? It can just be a regular std::thread instance that gets assigned via:

```cpp
motion_emu_thread = std::thread(MotionEmuThread, std::ref(emu_window))
```

src/common/motion_emu.cpp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+void Shutdown() {
+    if (motion_emu_thread) {
```
lioncash: Linking onto my comment above, if this made a regular std::thread, this should also check for thread joinability via:

```cpp
if (motion_emu_thread.joinable())
```

before performing a join.

src/common/emu_window.h
```diff
@@ -112,6 +112,27 @@ class EmuWindow
     void TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y);
 
     /**
+     * Signal accelerometer state has changed.
+     * @param x X-axis accelerometer value
+     * @param y Y-axis accelerometer value
+     * @param z Z-axis accelerometer value
+     * @note all values are in unit of g (gravitational acceleration).
+     *    e.g. x = 1.0 means 9.8m/s^2 in x direction.
+     *    see GetAccelerometerState for axis explanation.
```
lioncash: `see` can be `@see` to make doxygen create links to the functions.

src/common/emu_window.h
```diff
@@ -112,6 +112,27 @@ class EmuWindow
...
+     */
+    void AccelerometerChanged(float x, float y, float z);
+
+    /**
+     * Signal gyroscope state has changed.
+     * @param x X-axis accelerometer value
+     * @param y Y-axis accelerometer value
+     * @param z Z-axis accelerometer value
+     * @note all values are in deg/sec.
+     *    see GetGyroscopeState for axis explanation.
```
lioncash: ditto

src/common/emu_window.h
```diff
@@ -288,6 +313,14 @@ class EmuWindow
     s16 circle_pad_x; ///< Circle pad X-position in native 3DS pixel coordinates (-156 - 156)
     s16 circle_pad_y; ///< Circle pad Y-position in native 3DS pixel coordinates (-156 - 156)
 
+    s16 accel_x; ///< Accelerometer X-axis value in native 3DS unit
```
lioncash: `unit` → `units`

lioncash:
Is there any particular reason that the interface was designed to not be class-based? It adds more Init/Shutdown methods, when they could easily be automated through a constructor and destructor.

src/common/motion_emu.cpp
Subv: I think that's outside the scope of this PR and can be added on a future one so as to not clutter this one.

lioncash: It really isn't. It's like implementing a Vector3, but only providing accessors for x and y coordinates, because you aren't using z yet. A mathematical class should provide it's full set of very basic operations out of the box.

JayFoxRox: It still doesn't have to be part of this PR.

Using some broken partial-quaternion struct (and calling it that) or hardcoding these functions without any link to quaternions would be even worse.
The PR author suggests that these are indeed quaternions but also that we only need a tiny subset of them. I think this is fine.

We could, however, discuss the use of an external implementation of quaternions OR create an issue after the merge that our current quaternion implementation is currently incomplete (which is okay to me).
However, wether the PR author should adapt their code to the new quaternion code (either external or internal) is also up for debate. We can't just push more and more work onto PR authors. This PR is not about implementing quaternions: it is about emulating the motion sensor (which is a lot of work by itself)

wwylele:
@lioncash re: class-based
The reason is I am stupid :stuck_out_tongue: I will think about it when my brain is fresh.

src/common/motion_emu.cpp
lioncash: > It still doesn't have to be part of this PR.

Why not? This *is* the PR introducing it after all. There is literally no useful reason not to provide a very basic complete implementation that satisfies basic arithmetic use cases expected by the construct being provided.

If anything, that only makes it more telling that a Quaternion class should have initially been submitted before this.

> This PR is not about implementing quaternions: it is about emulating the motion sensor (which is a lot of work by itself)

Which requires a quaternion implementation by the looks of it, making it part of the process. These are not difficult things to provide, which is why I even bother to argue about this. This isn't even asking for much out of an initial implementation. For example: here, let me do two of them.

Addition:
```cpp
Quaternion operator+(const Quaternion& other) const {
    return { xyz + other.xyz; w + other.w };
}
```

Multiplication:

```cpp
Quaternion operator*(const Quaternion& other) const {
    const float x = xyz.x;
    const float y = xyz.y;
    const float z = xyz.z;

    const float ox = other.xyz.x;
    const float oy = other.xyz.y;
    const float oz = other.xyz.z;
    const float ow = other.w;

    Quaternion result;

    result.w     = (w * ow) - (x * ox) - (y * oy) - (z * oz);
    result.xyz.x = (w * ox) + (x * ow) + (y * oz) - (z * oy);
    result.xyz.y = (w * oy) + (y * ow) + (z * ox) - (x * oz);
    result.xyz.z = (w * oz) + (z * ow) + (x * oy) - (y * ox);

    return result;
}
```

Again, this does not require moving mountains. I understand the motion sensor is a lot of work, and I do appreciate @wwylele's time on it, but as I previously stated, a basic quaternion implementation is very easy to implement.

wwylele: OK this is my fault. I used to plan to complete the operators but forgot to do it when reviewing myself. Will add them.

JayFoxRox: The video_core also has a function (`AreQuaternionsOpposite` in src/video_core/gl_rasterizer.cpp) for dealing with lighting (quaternions) and that simply uses Vec4.

However, if you are going to complete your quaternion, you should probably make it a seperate commit. Probably also add it in src/common/ instead (we already have vector_math etc. there and this feels very generic)?

wwylele:
Now I make it a class-based interface, and here are two more questions:
 - What namespace should be `MotionEmu` in? Or just leave it outside (like `EmuWindow`)?
 - Should I move `Quaternion` to `vector_math.h` or a new header under common?

src/common/motion_emu.cpp
```diff
@@ -0,0 +1,116 @@
...
+    std::chrono::duration_cast<std::chrono::steady_clock::duration>(std::chrono::milliseconds(update_millisecond));
+static constexpr float PI = 3.14159265f;
+
+MotionEmu::MotionEmu(EmuWindow& emu_window) :
+    motion_emu_thread(&MotionEmu::MotionEmuThread, this, std::ref(emu_window)) {
+}
+
+MotionEmu::~MotionEmu() {
+    shutdown_event.Set();
+    motion_emu_thread.join();
```
lioncash: This should likely be checking

```cpp
if (motion_emu_thread.joinable())
```

before performing a join.

src/common/motion_emu.cpp
```diff
@@ -0,0 +1,116 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/emu_window.h"
+#include "common/math_util.h"
+#include "common/motion_emu.h"
+
+struct Quaternion {
+    Math::Vec3<float> xyz;
```
lioncash:
```cpp
Math::Vec3<float> xyz{0.f, 0.f, 0.f};
float w = 0.f;
```

lioncash:
@wwylele 

- Perhaps under a `Motion` namespace.
- Probably in a new header named `quaternion.h`.

wwylele:
Moved `Quaternion` to common and made it generic. I tried to follow the style of templates in `vector_math.h`, and didn't give it default values, for the following reason:
 - `Vec2/3/4` don't have default value either
 - Not sure how to make default value for such class template (is it good to just use 0 and 1?)
 - Not sure if {{0,0,0},1} should be a default value in in terms of math

wwylele:
Note to self: there is bug in thread creating and terminating

neobrain:

2016-07-28T17:29:31Z: https://github.com/citra-emu/citra/compare/5c743ddfa8e68897df6b7f52dabb6e324bce3fe1...wwylele:8cbf271992ca29a513870923d9322a4de6160d4a

wwylele:
Pause this for a while, waiting for the discussion in #1986

Subv:
>Is PI defined somewhere in our code base? If not, where should I put it?

M_PI is defined in cmath

yuriks:
> M_PI is defined in cmath
It is not. That's a GNU extension. We don't seem to have pi anywhere right now, but maybe common/math_utils.h would be a good place to put it.

wwylele:
I decided to not wait for #1995. Please review this.

src/common/thread.h
```diff
@@ -54,6 +55,15 @@ class Event {
...
 
+    template <class Clock, class Duration>
+    bool WaitUntil(const std::chrono::time_point<Clock, Duration>& time) {
+        std::unique_lock<std::mutex> lk(mutex);
+        if (!condvar.wait_until(lk, time, [this] { return is_set; }))
+            return false;
+        is_set = false;
+        return true;
+    }
+
```
JayFoxRox: Can you split this and the vector_math stuff into one or two other commits?
I feel like it's not too closely related with the motion sensor stuff.

Similar about the new quaternion code. It would also make review easier by splitting it into different units. :)

src/common/math_util.h
```diff
@@ -10,6 +10,8 @@
 
 namespace MathUtil {
 
+static constexpr float PI = 3.14159265f;
```
JayFoxRox: I find it very hard to believe that we don't have a PI elsewhere?!
Did you look for another one?

src/common/thread.h
wwylele: Split done

wwylele:
rebased and moved `motion_emu.cpp/.h` to `core/frontend/`, as they are depends on `emu_window`

src/core/frontend/emu_window.cpp
```diff
@@ -89,6 +90,27 @@ void EmuWindow::TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y) {
     TouchPressed(framebuffer_x, framebuffer_y);
 }
 
+void EmuWindow::AccelerometerChanged(float x, float y, float z) {
```
bunnei: These functions are not thread safe and probably should be

bunnei:
General comment that we should clean up EmuWindow to be just the window class, and move user input stuff to other interfaces; But since you're just implementing stubbed functions there this can be left for another change.

bunnei:
@wwylele That's fine for now I think.

Gentle reminder to the other reviewers to take a look at this - If I don't hear anything, expect to see it merged any time now...

Issue #1838, opened 2016-05-21
SDL2: Selecting an audio device


MerryMage:
We currently call [`SDL_OpenAudioDevice`](https://wiki.libsdl.org/SDL_OpenAudioDevice) with a `nullptr` as the first argument because this does (hopefully sensible) automatic selection of the most appropriate audio device. [A user reports](https://gbatemp.net/threads/citra-new-3ds-emulator.365154/page-210#post-6369199) this not to be the case.

SDL does let us list the available audio devices with [`SDL_GetAudioDeviceName`](https://wiki.libsdl.org/SDL_GetAudioDeviceName); this could be a configuration option for a user to override the automatic selection.

PR #1952, opened 2016-07-14
SDL2: Selecting an audio device

zach2good:
Everything looks stable and in line with the current coding style, but I was unable to test actual sound output through various homebrew apps.

Screenshot:
ss

### DIFF
--- a/src/audio_core/audio_core.cpp
+++ b/src/audio_core/audio_core.cpp
@@ -71,6 +71,13 @@ void SelectSink(std::string sink_id) {
     DSP::HLE::SetSink(iter->factory());
 }
 
+void SelectDevice(std::string device_id){
+    if (device_id == "") {
+        LOG_ERROR(Audio, "AudioCore::SelectDevice given invalid device_id");
+        return;
+    }
+}
+
 void Shutdown() {
     CoreTiming::UnscheduleEvent(tick_event, 0);
     DSP::HLE::Shutdown();
--- a/src/audio_core/audio_core.cpp
+++ b/src/audio_core/audio_core.cpp
@@ -71,13 +71,6 @@ void SelectSink(std::string sink_id) {
     DSP::HLE::SetSink(iter->factory());
 }
 
-void SelectDevice(std::string device_id){
-    if (device_id == "") {
-        LOG_ERROR(Audio, "AudioCore::SelectDevice given invalid device_id");
-        return;
-    }
-}
-
 void Shutdown() {
     CoreTiming::UnscheduleEvent(tick_event, 0);
     DSP::HLE::Shutdown();
--- a/src/audio_core/sdl2_sink.cpp
+++ b/src/audio_core/sdl2_sink.cpp
@@ -56,17 +56,17 @@ SDL2Sink::SDL2Sink() : impl(std::make_unique<Impl>()) {
     if (device_count < 1 ||
         Settings::values.audio_device_id == "auto" ||
         Settings::values.audio_device_id == "") {
-        impl->audio_device_id = SDL_OpenAudioDevice(nullptr, false, &desired_audiospec, &obtained_audiospec, 0);
+        impl->audio_device_id = SDL_OpenAudioDevice(nullptr, false, &desired_audiospec, &obtained_audiospec, SDL_AUDIO_ALLOW_ANY_CHANGE);
         if (impl->audio_device_id <= 0) {
-            LOG_CRITICAL(Audio_Sink, "SDL_OpenAudioDevice failed");
+            LOG_CRITICAL(Audio_Sink, "SDL_OpenAudioDevice failed with code %d for device \"auto\"", impl->audio_device_id);
             return;
         }
     }
     else
     {
-        impl->audio_device_id = SDL_OpenAudioDevice(device_map[device_id].c_str(), false, &desired_audiospec, &obtained_audiospec, 0);
+        impl->audio_device_id = SDL_OpenAudioDevice(Settings::values.audio_device_id.c_str(), false, &desired_audiospec, &obtained_audiospec, SDL_AUDIO_ALLOW_ANY_CHANGE);
         if (impl->audio_device_id <= 0) {
-            LOG_CRITICAL(Audio_Sink, "SDL_OpenAudioDevice failed");
+            LOG_CRITICAL(Audio_Sink, "SDL_OpenAudioDevice failed with code %d for device \"%s\"", impl->audio_device_id, Settings::values.audio_device_id.c_str());
             return;
         }
     }
--- a/src/audio_core/sdl2_sink.cpp
+++ b/src/audio_core/sdl2_sink.cpp
@@ -10,6 +10,8 @@
 #include "audio_core/audio_core.h"
 #include "audio_core/sdl2_sink.h"
 
+#include "core/settings.h"
+
 #include "common/assert.h"
 #include "common/logging/log.h"
 #include <numeric>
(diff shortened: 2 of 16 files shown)


JayFoxRox:
Please attach a screenshot of the new UI in your original commit description (first post).
I'll remove my post you have done this.

src/audio_core/audio_core.cpp
```diff
@@ -71,6 +71,13 @@ void SelectSink(std::string sink_id) {
     DSP::HLE::SetSink(iter->factory());
 }
 
+void SelectDevice(std::string device_id){
+    if (device_id == "") {
```
lioncash: `if (device_id.empty())`

src/audio_core/sdl2_sink.h
```diff
@@ -22,9 +23,16 @@ class SDL2Sink final : public Sink {
...
+
+
 private:
     struct Impl;
     std::unique_ptr<Impl> impl;
+    int device_id;
+    std::map<int, std::string> device_map;
 };
 
+
```
lioncash: Ditto

src/audio_core/sink.h
```diff
@@ -29,6 +30,9 @@ class Sink {
 
     /// Samples enqueued that have not been played yet.
     virtual std::size_t SamplesInQueue() const = 0;
+
+    virtual void SetDevice(int device_id) = 0;
+    virtual std::map<int, std::string>* GetDeviceMap() = 0;
```
lioncash: (forgot to clarify in my initial comment):

In the sense that the base Sink class contains the map of devices as a protected/private member (whichever is appropriate).

zach2good:
Screenshot
ss

src/audio_core/sink_details.h
```diff
@@ -7,6 +7,7 @@
 #include <functional>
 #include <memory>
 #include <vector>
+#include <map>
```
lioncash: This should come before `<memory>`

src/core/settings.h
```diff
@@ -6,6 +6,7 @@
 
 #include <string>
 #include <array>
+#include <map>
```
lioncash: Is this necessary, or is it just a leftover that crept through?

src/audio_core/sink.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <vector>
+#include <map>
```
lioncash: This should come before `<vector>`

src/audio_core/sdl2_sink.h
```diff
@@ -6,6 +6,7 @@
 
 #include <cstddef>
 #include <memory>
+#include <map>
```
lioncash: This should come before `<memory>`

JayFoxRox:
I'm wondering if this should be integrated differently into the GUI.

This one only applies to SDL2, so really it should only be visible if SDL2 is selected as an option.
So maybe we should just have an extra group frame / box with SDL2 related options which is only shown once SDL is selected?

This also raises the question wether we should have a "auto" option for the "Output Engine" ( @MerryMage what was the reason for adding that?) - however, that's not scope of this PR.

(Similarily for the video options software rendering and opengl rendering should show different options - upscaling will only work with hw rendering for example)

Note that all of this is really not scope of this PR, so I'm also fine with creating an issue about this later. But I'd still like to raise awareness about this kind of issue BEFORE merging more GUI stuff.

MerryMage:
Using a `std::string` is an okay choice for the device identifier; strings are used to identify devices in (most?) sound libraries I've seen.

---

Overall design comments: 

* Not sure why a device map is a `std::map`. A `std::set` would suffice. Non-unique device ids don't matter since `SDL_OpenAudioDevice` is unable to tell the difference anyway. (Better audio libraries like libsoundio guarantee unique device id strings.)
* `SDL2Sink::SetDevice` should likely call `SDL_CloseAudioDevice` and `SDL_OpenAudioDevice` as appropriate. `SDL2Sink::SDL2Sink` can call `SetDevice` to avoid code duplication.
* `AudioCore::SelectDevice` seems a bit useless. It should probably call `sink->SetDevice`.
* The device combobox should probably update to show a list of devices when the sink changes.
* Having a pre-populated `g_device_map` is probably not the cleverest design, since `sink` can change. Perhaps consider not pre-populating a collection, but instead generating it at run-time when needed. (This would also allow for one's audio configuration to change: e.g.: if someone connects their headphones while Citra is running.)

---

@JayFoxRox 

Regarding hiding the combobox: 
That was my thought at first, but then I realised it's not really SDL2 specific since most audio libraries allow one to specify an output device.

Regarding the "auto" sink:
We can discuss that later.

src/audio_core/sink_details.cpp
```diff
@@ -22,4 +22,11 @@ const std::vector<SinkDetails> g_sink_details = {
     { "null", []() { return std::make_unique<NullSink>(); } },
 };
 
+#ifdef HAVE_SDL2
+SDL2Sink sink;
+const std::map<int, std::string> g_device_map = *sink.GetDeviceMap();
```
MerryMage: Logic note: `g_device_map` should likely change depending on the currently selected sink. Consider generating a list of devices as-and-when-needed instead of prepopulating a list.

MerryMage:
Closes #1838.

src/audio_core/sdl2_sink.cpp
```diff
@@ -45,10 +47,28 @@ SDL2Sink::SDL2Sink() : impl(std::make_unique<Impl>()) {
...
+        Settings::values.audio_device_id.empty()) {
+        impl->audio_device_id = SDL_OpenAudioDevice(nullptr, false, &desired_audiospec, &obtained_audiospec, SDL_AUDIO_ALLOW_ANY_CHANGE);
+        if (impl->audio_device_id <= 0) {
+            LOG_CRITICAL(Audio_Sink, "SDL_OpenAudioDevice failed with code %d for device \"auto\"", impl->audio_device_id);
+            return;
+        }
+    }
+    else
+    {
+        impl->audio_device_id = SDL_OpenAudioDevice(Settings::values.audio_device_id.c_str(), false, &desired_audiospec, &obtained_audiospec, SDL_AUDIO_ALLOW_ANY_CHANGE);
```
yuriks: There don't seem to be any major differences between these two calls. You should just use variables and have them go through the same call+error handling path.

src/citra_qt/configure_audio.cpp
```diff
@@ -36,9 +41,45 @@ void ConfigureAudio::setConfiguration() {
...
+
+         // Case where the sink is pointed to a directly known sink device (NullSink, SDL2Sink).
+         // The first index (0) should be "auto", which is not in the vector list (-1).
+         quick_sink_populate_device = AudioCore::g_sink_details[sink_index - 1];
+    }
+    auto iter = quick_sink_populate_device.factory();
+
+    // Prevent accessing a null device list.
+    if (std::move(iter)->GetDeviceMap()) {
+        for (const auto& device : *(std::move(iter)->GetDeviceMap())) {
```
yuriks: You're moving from iter twice, this is pointless and a bug waiting to happen. Why are you even using std::move anyway? You're not passing ownership to anything.

src/audio_core/sink.h
```diff
@@ -29,6 +29,9 @@ class Sink {
 
     /// Samples enqueued that have not been played yet.
     virtual std::size_t SamplesInQueue() const = 0;
+
+    virtual void SetDevice(int device_id) = 0;
+    virtual std::vector<std::string>* GetDeviceMap() = 0;
```
yuriks: This should return a `const` pointer. (Or a reference, but I could conceive of a backend not supporting this somehow and wanting to return NULL to signify that.)

src/citra_qt/configure_audio.cpp
```diff
@@ -36,9 +41,45 @@ void ConfigureAudio::setConfiguration() {
...
+     if (sink_index > 0) {
+
+         // Case where the sink is pointed to a directly known sink device (NullSink, SDL2Sink).
+         // The first index (0) should be "auto", which is not in the vector list (-1).
+         quick_sink_populate_device = AudioCore::g_sink_details[sink_index - 1];
+    }
+    auto iter = quick_sink_populate_device.factory();
+
+    // Prevent accessing a null device list.
+    if (std::move(iter)->GetDeviceMap()) {
```
yuriks: You should call `GetDeviceMap` just once and store it in a variable.

src/citra_qt/configure_audio.cpp
```diff
@@ -36,9 +41,45 @@ void ConfigureAudio::setConfiguration() {
...
 }
+
+void ConfigureAudio::updateAudioDevices(int sink_index) {
+
+    ui->audio_device_combo_box->clear();
+    ui->audio_device_combo_box->addItem("auto");
+
+    // If the current sink is set to "auto", the "front" sink is selected.
+    auto quick_sink_populate_device = AudioCore::g_sink_details[0];
+     if (sink_index > 0) {
```
yuriks: Bad indentation

yuriks: On this whole block, too

bunnei:
Do you plan on addressing @yuriks' comments?

zach2good:
@bunnei  I will sort all of this out, but I'm busy until next week

wwylele:

wwylele:
@zach2good Do you plan to work on this any more?

yuriks:
Closing due to inactivity. (If it's eventually updated we can re-open it.)

PR #1963, opened 2016-07-20
Implement actual system time

wwylele:
This implemented synchronizing the system time in the `SharedPage` with actual system time. According to [the reaserch](https://github.com/wwylele/ctrhwtest/tree/master/sharedpage), it is updated every hour on real 3DS. so implemented it the same way here. Also, the name of the first field of `SharedPage` was changed, for its behaviour on real 3DS.

When translating system time to 3DS format, I uses Jan 1, 2000 as an auxiliary epoch, for the following reason:
 - it is a limit in real 3DS settings
 - `epoch = 1970` for `time_t` is not a real standard, and one needs to deal with timezone when using it
 - directly using `epoch = 1900` (real 3DS epoch) may result in invalid values, for it is before `time_t`'s epoch.

There is a known issue: game uses the formula `SharedPage.date_time + (currentSystemTick - SharedPage.update_tick) * tick_to_second_coefficient` (not exactly, needs more RE) to calculate the current time. Because of the slow pace of citra, the time presented in game will be gradually desynchronized with the actual time, and will experience a time jump at the next update. There are other ways to update system time (listed in #1641), but none of them perfectly resolves the problem. I don't think this is a big deal, and it will be automatically resolved when citra runs at full speed in the future.

----
Sorry to @ObsidianX for didn't talk to him before taking over it, but for #1641 hasn't been activated for a month, I assume that it died. Also sorry for me stealing the show in that PR. Anyway we should put an end to this. This less-than-100-line shouldn't take that long.

All research is in the old PR. Tested on Windows and Ubuntu, with different time zone, and with daylight saving on and off. The clock in System Settings App always shows the same time as computer's.

### DIFF
--- a/src/core/core_timing.h
+++ b/src/core/core_timing.h
@@ -26,7 +26,7 @@
 extern int g_clock_rate_arm11;
 
 inline s64 msToCycles(int ms) {
-    return g_clock_rate_arm11 / 1000 * ms;
+    return (s64)g_clock_rate_arm11 / 1000 * ms;
 }
 
 inline s64 msToCycles(float ms) {
--- a/src/core/hle/shared_page.cpp
+++ b/src/core/hle/shared_page.cpp
@@ -2,8 +2,11 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <chrono>
 #include <cstring>
+#include <ctime>
 
+#include "core/core_timing.h"
 #include "core/hle/shared_page.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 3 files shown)


neobrain:

2016-07-20T15:14:58Z: https://github.com/citra-emu/citra/compare/f95d119dde2addd4a83fe8dd5ca6de5139c20336...wwylele:b7ea4bba094d84ca077fa1dfee0f7e79953fb5cd

src/core/core_timing.h
```diff
@@ -26,7 +26,7 @@
 extern int g_clock_rate_arm11;
 
 inline s64 msToCycles(int ms) {
-    return g_clock_rate_arm11 / 1000 * ms;
+    return (s64)g_clock_rate_arm11 / 1000 * ms;
```
wwylele: use C-style to fit the style of this file

Subv:
LGTM

archshift:
LGTM

neobrain:
Is there any reason to emulate the system time via the host system time other than convenience? If not, I disagree with the approach taken in this PR. This change is:
* *not* sensible: Host system time and guest system time only match when an application is emulated at exactly its native speed 100% of the time. They may vastly differ on slower systems, or when emulation runs at double speed (note that speedup is a commonly supported and desired emulator feature).
* *not* deterministic: Emulation speed is not deterministic, hence making guest system time rely on host system time will inherently introduce (more) indeterministic behavior in Citra.
* *not* compatible with snapshots ("state saves"): Loading an emulation state that was saved hours ago should not magically change the guest system time. Being able to properly restore emulation state is hence limited with this change *by design*.
* *not* constructive: Any implementation aiming to be sensible and deterministic will need to write the code in this PR from scratch. In other words, this PR is a step sideways rather than forward.

JayFoxRox:
(I did not review this PR yet, and did not check how it is implemented yet)

@neobrain I think what you suggest should be in a seperate issue after the merge. We can then discuss implementing different `TimeProviders` or something (Say, host RTC, virtual RTC and based on actual emu speed).

There will certainly be games which depend on a proper RTC, there will be other games which will break because of desync. It's a problem of HLE and it will have to be made a users choice / per game setting. I don't think there's a silver-bullet.
Other emulators such as QEMU also have different RTCs available.

If I understand this PR correctly, this implements the base and additionally implements the host RTC choice.
So if the base is re-usable with different mechanisms to get the intended time, then this should be considered merge-worthy.

neobrain:
@JayFoxRox Frankly said, I disagree with the entire sentiment of your post, which is basically "Let's implement anything as long as it makes games work, regardless of whether it helps us coming up with a proper solution at all". You have to ask yourself three questions:
* What reason do we have to provide a host RTC (as you call it) when it inherently will deviate from a correct guest RTC in all but the ideal scenario of consistent fullspeed emulation? While seemingly making things work, you may be underestimating the impact on the user experience that these strong deviations may have.
* How much more effort is it to come up with any approach to more representative RTC emulation? I wouldn't expect this to be difficult at all.
* Are you just trying to defend this change because it is already written and want to "waste" time by requiring a rewrite? I hope I don't need to point out how this is a terrible attitude that will yield nothing but a broken emulator in the long run.

Lectem:
Just my 2cents as I agree with what @neobrain said. There are similar issues in game engine development for example. You rarely want to have your game (emulator) time based on your computer clock. The 1st reason is debugging, for example being able to pause a timer with a single shortcut, and then advance in time with another key is more than helpful, and it gives reproducibility.
It might be a bit more tricky in the case of an emulator, but I really think some Timer class is needed to abstract the emulated timeline and, imo, should be based on the number of cycles/frames emulated.

yuriks:
LGTM

neobrain:

2016-07-23T11:30:32Z: https://github.com/citra-emu/citra/compare/624cd9cd983e294cea77a337e8aed916f28ababe...wwylele:d63a76f4ce47962d7f191f3d4911c12344c4d249

neobrain:
It's good to know that shutting off active discussions and omitting PR author plans is how @yuriks decided this project will be ran now.

Thanks for the worthwhile discussion anyway, @wwylele .

PR #1964, opened 2016-07-21
Fixes SDL2.dll copy to bindir on windows

Lectem:
Citra uses the SDL2_DLL_DIR variable to copy the dll to the output directory on windows, however it is only set when using the bundled version of SDL2. 
Users that had CITRA_USE_BUNDLED_SDL2 set to false would have to copy the dll manually.


neobrain:

2016-07-21T08:32:05Z: https://github.com/citra-emu/citra/compare/f95d119dde2addd4a83fe8dd5ca6de5139c20336...Lectem:c7125f37aa5325f84d615eaf3b8fa2e67eb7d800

bunnei:
LGTM

PR #1966, opened 2016-07-21
CMake: Fix Info.plist template for citra_qt/OSX

dwhinham:
The Info.plist template incorrectly uses parentheses instead of curly braces,
which means that building the .app bundle using regular 'make' results in the
variable not being replaced, and hence the app bundle won't start because the
executable name is incorrect.

This commit fixes this issue.


archshift:
LGTM

LittleWhite-tb:
LGTM

Issue #1968, opened 2016-07-22
Core dumped


WyRe:
I'm trying to open Citra after compile from ArchLinux AUR repos [https://aur.archlinux.org/packages/citra-git/](url) and I get a message saying:

Ilegal instruction (core dumped)

I've got to do some modifications to PKGBUILD according to that issue:

[https://github.com/citra-emu/citra/issues/1955#issuecomment-233106336](url)

As you can see in AUR commentaries. I'm running under OpenGL 4.5.0 and NVIDIA 367.35 driver. Thank you so much.

JayFoxRox:
Contact the packager via the AUR comments (or try linkmauve1 on freenode / in \#citra).
This is not officially maintained and also not a problem with Citra itself (We should probably make that more clear).

(A proper title for this issue would have been "AUR Package 'citra-git' broken: Illegal Instruction")

---

Closed as invalid.

PR #1973, opened 2016-07-23
Remove the -msse4.1 on ¬MSVC

linkmauve:
This option makes the generated binary crash with an illegal
instruction when the target CPU doesn’t support the SSE4.1 extension
(see #1968), with no noticeable performance increase compared to a
generic build.


neobrain:

2016-07-23T16:16:16Z: https://github.com/citra-emu/citra/compare/2dee47c8277484f0fdc486ec201ed5c25d7b73a3...linkmauve:e09e6837208a5cda81c0eee233f301433201cb8b

yuriks:
LGTM

Issue #1944, opened 2016-07-09
Citra crashes when settings file is missing (or old)


LittleWhite-tb:
Hello,

I encounter a crash with citra on Linux.
```
[   0.001852] Debug <Critical> common/emu_window.cpp:operator():96: Assertion Failed!
```

![crash](https://cloud.githubusercontent.com/assets/8618754/16709049/ef6874d2-4605-11e6-825a-165f83640c82.png)

We can easely reproduce it by just removing the settings file in .config/citra-emu/qt-config.ini
It also happen if the file is old (coming from a old citra version).

yuriks:
Can you upload your qt-config.ini file which reproduces this crash? (You can edit any paths in it if you want to "anonymize" it to remove file history, etc.)

yuriks:
Very likely can be considered a dupe of #1205 (even though the crash is in a separate location).

LittleWhite-tb:
I could quick fix the issue by adding a check in GRenderWindow::OnFramebufferSizeChanged() in order to avoid a size of 0.
But, at the moment, I consider it just a quick fix (not a good fix).

About the config file. One way to reproduce it is to remove the config file (renaming/deleting it).
Otherwise, my old config file was :

[old.ini.zip](https://github.com/citra-emu/citra/files/355931/old.ini.zip)

LittleWhite-tb:
I think, this one should be closed since #1974 is merged.

PR #1974, opened 2016-07-23
Protection against a resize of size 0

LittleWhite-tb:
Fix for #1944 . I am very surprised nobody encounter this one so, I suspect this is only in Linux + Qt5.5.

When we call render_window->hide(), GGLWidgetInternal::resizeEvent() is called with a size of 0 which lead to EmuWindow::FramebufferLayout::DefaultScreenLayout having an assert on the width/height.

### DIFF
--- a/src/common/emu_window.cpp
+++ b/src/common/emu_window.cpp
@@ -51,7 +51,6 @@ static bool IsWithinTouchscreen(const EmuWindow::FramebufferLayout& layout, unsi
 }
 
 std::tuple<unsigned,unsigned> EmuWindow::ClipToTouchScreen(unsigned new_x, unsigned new_y) {
-
     new_x = std::max(new_x, framebuffer_layout.bottom_screen.left);
     new_x = std::min(new_x, framebuffer_layout.bottom_screen.right-1);
 
@@ -92,9 +91,9 @@ void EmuWindow::TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y) {
 }
 
 EmuWindow::FramebufferLayout EmuWindow::FramebufferLayout::DefaultScreenLayout(unsigned width, unsigned height) {
-
-    ASSERT(width > 0);
-    ASSERT(height > 0);
+    // When hiding the widget, the function receives a size of 0
+    if (width == 0) width = 1;
+    if (height == 0) height = 1;
 
     EmuWindow::FramebufferLayout res = { width, height, {}, {} };
 


neobrain:

2016-07-23T17:42:48Z: https://github.com/citra-emu/citra/compare/2dee47c8277484f0fdc486ec201ed5c25d7b73a3...LittleWhite-tb:3e0e92101289c720bbd47f733f76007dc0146f68

yuriks:
Can we just remove/modify the offending assert instead? Does anything bad happen otherwise?

MerryMage:
The assert has been hit in the past on start-up. (One user report.)

LittleWhite-tb:
After the assert you have :
```
float window_aspect_ratio = static_cast<float>(height) / width;
```

Which is 0 divided by 0 in this case.

neobrain:

2016-07-23T18:55:23Z: https://github.com/citra-emu/citra/compare/2117be46a8c4b4785f0411f055d81f04312955e6...LittleWhite-tb:961ae89c80404c500951b5243cd0cdadefd8a5b5

src/common/emu_window.cpp
```diff
@@ -93,8 +93,9 @@ void EmuWindow::TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y) {
 
 EmuWindow::FramebufferLayout EmuWindow::FramebufferLayout::DefaultScreenLayout(unsigned width, unsigned height) {
 
-    ASSERT(width > 0);
-    ASSERT(height > 0);
+    // When hiding the widget, the function receive a size of 0
```
yuriks: receives*

src/common/emu_window.cpp
```diff
@@ -93,8 +93,9 @@ void EmuWindow::TouchMoved(unsigned framebuffer_x, unsigned framebuffer_y) {
 
 EmuWindow::FramebufferLayout EmuWindow::FramebufferLayout::DefaultScreenLayout(unsigned width, unsigned height) {
```
yuriks: Also, get rid of this blank line while you're at it.

LittleWhite-tb:
Implemented new fix as suggested by yuriks on IRC

neobrain:

2016-07-23T19:04:35Z: https://github.com/citra-emu/citra/compare/2117be46a8c4b4785f0411f055d81f04312955e6...LittleWhite-tb:7331b790217f9a3e86e5668cb134bd8f073ff289

yuriks:
LGTM

bunnei:
LGTM

Issue #1844, opened 2016-05-22
Inform user of missing required files & document how to dump them


wwylele:
If a game wants a NAND archive that user didn't place correctly, citra just log some "failed to get a handle for archive" with unreadable IDs. What's worse, if a game wants `shared_font.bin` but it isn't there, citra [just crashes](https://discuss.citra-emu.org/d/253-rune-factory-4-not-starting).
In these cases, citra should inform the user what is missing, and give instructions on dumping those files (or a link to a instruction page, like we did for decrypting). Speak of which, we have 3dsutil, but don't have a clear instruction on using it.

JayFoxRox:
For the time being: https://github.com/citra-emu/citra/wiki/Home-Folder

---

Also keep in mind that this is a bug tracker for developers of Citra.

Please use the [Citra Forum for questions and support](https://discuss.citra-emu.org/) instead if you have questions regarding this.
Note that there is no support if you use pirated copies of the game or 3DS bios files.

FearlessTobi:
Can we close this? There are now popups to inform the user of missing files.

PR #1983, opened 2016-07-28
Shared font file: Instead of segfaulting, log an error to remind the user to dump it

hrydgard:
In case the shared font file is missing, it seems better to log an error than to crash.


neobrain:

2016-07-28T19:49:43Z: https://github.com/citra-emu/citra/compare/5c743ddfa8e68897df6b7f52dabb6e324bce3fe1...hrydgard:9a9e9dc375163090394e212a11e6f58379dc9a90

wwylele:
Although this is not what was proposed in #1844 ... well, at least better than nothing. LGTM

hrydgard:
Oh, didn't notice that, or I would have linked it. I agree that this is not fantastic but it's some sort of "MVP" of a solution.

bunnei:
LGTM, thank you

PR #1985, opened 2016-07-30
Possible solution for thread safety in emu_window files

N0r1x:
Atomic looked like a clean solution to the todo's. Don't think there's a lot to say about this.

I haven't been able to test this since I don't have any ROMs, but it looks harmless enough to my eyes.

### DIFF
--- a/src/common/emu_window.cpp
+++ b/src/common/emu_window.cpp
@@ -12,11 +12,15 @@
 #include "video_core/video_core.h"
 
 void EmuWindow::ButtonPressed(Service::HID::PadState pad) {
-    pad_state.hex |= pad.hex;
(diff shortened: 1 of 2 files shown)


neobrain:

2016-07-30T13:49:12Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...N0r1x:e2d4521a82321ec7dbd0630067d1543e35d8e94f

neobrain:
This doesn't really change much, unfortunately. GetCirclePadState will still return inconsistent results when it is called while CirclePadUpdated has already updated the x component but not yet the y component on another thread.

Issue #1883, opened 2016-06-04
Frontend <-> Backend communication is frought with race conditions


archshift:
This is mostly a problem with `citra-qt`, since that frontend uses a separate thread for emulation code and frontend code.

For example, [renderer_opengl.cpp L388](https://github.com/citra-emu/citra/blob/master/src/video_core/renderer_opengl/renderer_opengl.cpp#L388) and [emu_window.h L229](https://github.com/citra-emu/citra/blob/master/src/common/emu_window.h#L229) operate on different threads with no synchronization, and both access the `emu_window` class's `framebuffer_layout` variable.

This particular race condition actually causes a bug on OS X (release, not debug), where resizing the window does not correctly reposition the emulated screens.

Other race conditions involve configuration, where different threads might access the config singleton, and input, where one thread might trigger a keypress and another read it with no synchronization.

yuriks:
For reference, I strongly believe that to solution to this is *not* to simply do the minimum amount of work required to synchronize accesses using mutexes or atomics and calling it a day. (Though such an incremental improvement would probably be welcome anyway.) Rather, I think the communication between the frontend(s) and the emulation core should be re-structured around a message queue that centralizes passing of information (such as input state, configuration changes, emulation state changes (such as pausing/stopping), requests for HLE UIs in the future, etc.) between the core and the frontend.

Issue #1182, opened 2015-10-01
Introduce a tabbed configuration dialog


neobrain:
Tabbed configuration dialogs are great - they allow for bundling all relevant configuration modules into a single dialog so that the user can conveniently inspect and tune all settings without having to step through multiple windows.

There are some issues with traditional tabbed layouts, though: First off, text-only content makes it cognitively less straightforward to find particular functionality quickly. It would be beneficial to additionally have an icon per configuration module (= tab) hence. Secondly, a horizontal tab bar takes away (for configuration dialogs usually precious) vertical screen space. Hence a vertical tab bar should be used instead.

Both of these points (and others) have been addressed by the KDE community:
![KDE page widget](http://www.davidrevoy.com/data/images/blog/2013/01/mint14kde/linux-mint-kde-14-digital-painting_05.jpg)

I think KDE's solution is both very usable and visually pleasing, and hence would be a great candidate for adoption in Citra. The relevant code can be found at http://api.kde.org/4.x-api/kdelibs-apidocs/kdeui/html/dir_1f296a4dd1c9f1f4073f9d2032e2722f.html , and can be stripped down a bit and then included in Citra.

(Note that currently we do not actually have enough configuration dialogs to sensibly introduce such a tabbed configuration dialog - however in the future this will definitely be useful.)

Gareth422:
This is my favourite design for configuration dialogues. I would support this.

lioncash:
Same

JayFoxRox:
Partially implemented by #1368 . Still missing vertical tabs and icons so I'll keep this one opened

LittleWhite-tb:
Is someone having some ideas for the icons we should use for this tabbed configuration dialog ? It would be cool for the person willing to work on that, to have the ressources ready (even if it is a placeholder).

Megamouse:
looks like this can be closed now

Subv:
I'll leave my subjective opinion here, I don't really like how that KDE dialog looks, the vertical bar with the icons is IMO redundant.

GPUCode:
UI is now tabbed will be closing this

PR #1986, opened 2016-07-30
[RFC] Input Overhaul into InputCore

### LINKED ISSUE #1707: Citra does not recognise key pressed in foreign keyboard layout
for example, key 'A' in english looks 'Ф' in russian even if it one physical key.
Maybe change asking ascii-code to scan-codes?

makotech222:
Hi everyone! This is a request for comments on my architecture for the input overhaul. 

As is, this pull request is functional, in that Citra can handle input from keyboard and gamepads. A couple of things are broken as of right now, though. 

- Keybinding from GUI doesn't work yet, don't try to use it.
- Circle Pad Modifier key doesn't work

The code flows like so: First, settings are read from the respective settings file. The previous format for binding keys was just the QT/SDL value for each keyboard key. It now goes as follows:

pad_a=Framework/DeviceID/Device/Button
Example:
pad_a=SDL/0/Gamepad/Button A
or
pad_a=QT/0/Keyboard/1e (now a scancode)

(Note: SDL uses different scan code than QT does. Not sure what effect this will have yet)

This pretty much copies how Dolphin does keybinding. From there, inputcore is initialized upon game starting, where it reads the settings values and initializes an IDevice object for each unique mapping. After initializing, input from SDL Gamepads are polled every tick, while keyboard input is still handled from the front end on an event basis.

This pull request as is, fixes
#1707 (well i need someone to test it. I switched to using scancodes)
#1883 (partially, keyboard input is now thread safe)

So here is what I would like comments on:

1. General architecture. I tried to make it as friendly to extension as possible, allowing for more frameworks to be added (XInput?) Also tried to allow as many devices to be connected as desired by user.

2. What to do with the Circle Pad Modifier button. Its really only necessary for keyboard. Removing it as a 'game' input simplifies my code. I'd rather it be treated as more of a hotkey or some such.

3. I was thinking about ripping out the Input config tab, and putting it into it's own form. It needs quite a bit more space I imagine, as i now need configurations for things like deadzones. 

4. I was thinking about switching inputcore from a namespace into a singleton class. I don't like allowing global access to its members like pad state and circle state. I didn't originally because audiocore and videocore are namespaces, so i followed suit.

I attached a sample .ini config file (change extension to .ini from .txt) since key binding from GUI isn't working yet. This will load up the first gamepad connected to computer when used.
[qt-config.txt](https://github.com/citra-emu/citra/files/391951/qt-config.txt)

If i think of any more questions, i will add them. Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/1986)


neobrain:

2016-07-30T15:47:07Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:8aa5a2c64f93016c380a82b6719f333f075ceb7b

neobrain:

2016-07-30T16:39:48Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:788f00119cd3bd292738e7d7676c05746840cf0c

neobrain:

2016-07-30T17:19:30Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:f1979ac39d5754f0b12ac0b0a243ada6e2551c98

neobrain:

2016-07-30T18:07:39Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:2239cc9f0c1a15d5390589f2727e657a68c0ad40

lioncash:
I don't have time right now to fully review this, but our naming scheme uses `snake_case` for source file names

archshift:
Woo, thread safety! 👍

makotech222:
@lioncash Okay I will fix SDLGamepad. Is IDevice okay?

lioncash:
Given the criteria specified above, no, that's not snake_case.

You can just name it `device.h`/`device.cpp` and other classes that derive from it can use more descriptive names. e.g. This is what @MerryMage did in audio_core for the sink interface.

wwylele:
I roughly looked through it, and here are some of my thoughts and questions:

 - I agree that the modifier is mostly only used by keyboard, but this is exactly the reason it _shouldn't_ be a hotkey, for hotkeys are _program wide_ shortcut keys for emulator features or so. Since it is for keyboard, then why not just making it an additional member in `class Keyboard`?

 - (questions for those familiar with HID module) What is the period of HID updating exactly? Is 60 times per second true? (for all inputs, buttons, touch, etc.?) Should it be updated with GSP (the old implement), or just updating independently is OK (implement in this PR)?

 - Touch input is also experiencing thread safety issue. Do you plan to move them into your input core? Or if not, do you leave the space so that any one can extend it with touch input? Also, what about gyroscope and accelerometer? I won't mind if you make any changes that is conflict with #1951 since it's also an RFC, but I think you should take them into account.

 - There are many code style problem, but I think we can leave this after discussing the design.

makotech222:
Thanks wwylele. Yes, many code style problems i plan to clean up when i make official PR. I have to look into touch input and others, i haven't really considered them too much yet. Good idea on making circle pad modifier a member of keyboard class, i will look into doing that.

src/core/settings.h
```diff
@@ -6,87 +6,147 @@
...
+    enum class Device {
+        Keyboard, Gamepad
+    };
+    struct InputDeviceMapping {
+        DeviceFramework framework;
+        int number;
+        Device device;
+        std::string key;
+        InputDeviceMapping() {
+            this->framework = DeviceFramework::Qt;
```
lioncash: You can assign to member variables directly to specify initial values, so the following can be done

```cpp
struct InputDeviceMapping
{
    DeviceFramework framework = DeviceFramework::Qt;
    int number = 0;
    Device device = Device::Keyboard;
    std::string key;

    InputDeviceMapping() = default;
 
    ... // rest of the struct follows
};
```

std::string initializes to an empty string by default, so you don't need to specify it.

src/core/settings.h
```diff
@@ -6,87 +6,147 @@
...
+            this->number = 0;
+            this->device = Device::Keyboard;
+            this->key = "";
+        }
+        InputDeviceMapping(std::string input) {
+            std::vector<std::string> parts;
+            Common::SplitString(input, '/', parts);
+            if (parts.size() == 4) {
+                if (parts[0] == "Qt")
+                    this->framework = DeviceFramework::Qt;
```
lioncash: Many of the usages of `this->` in this struct aren't necessary.

src/core/settings.h
```diff
@@ -6,87 +6,147 @@
...
+                this->number = 0;
+                this->device = Device::Keyboard;
+                this->key = "";
+            }
+        }
+
+        bool operator==(const InputDeviceMapping& rhs) const {
+            return (this->device == rhs.device) && (this->framework == rhs.framework) && (this->number == rhs.number);
+        }
+        std::string toString() {
```
lioncash: The signature should be as follows:

```cpp
std::string to_string() const
```

src/core/settings.h
```diff
@@ -6,87 +6,147 @@
...
+            std::string result = "";
```
lioncash: `= ""` isn't necessary (as std::string initializes to an empty string).

src/input_core/devices/IDevice.h
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <map>
+
+#include "core/settings.h"
+#include "input_core/key_map.h"
+class IDevice {
```
lioncash: A base class should always provide a virtual destructor. In this case,

```cpp
virtual ~IDevice() = default;
```

src/input_core/devices/Keyboard.h
```diff
@@ -0,0 +1,48 @@
...
+struct KeyboardKey {
+    uint32_t key;
+    uint32_t scancode;
+    std::string character;
+    KeyboardKey(uint32_t Key, uint32_t Scancode, std::string Character) {
+        key = Key;
+        scancode = Scancode;
+        character = Character;
+    }
+    bool operator==(KeyboardKey& other) {
```
lioncash:
```cpp
bool operator==(const KeyboardKey& other) const
```

src/input_core/devices/SDLGamepad.h
```diff
@@ -0,0 +1,48 @@
...
+        { "Left X-","leftx" },
+        { "Right Y+","righty" },
+        { "Right Y-","righty" },
+        { "Right X+","rightx" },
+        { "Right X-","rightx" },
+    };
+    static bool SDLInitialized;
+    std::map<std::string, bool> keysPressed;
+    SDL_GameController* gamepad;
+public:
```
lioncash: Your public interface should generally be shown before the private one, since that's the one people usually care about when using a class.

src/input_core/devices/SDLGamepad.h
```diff
@@ -0,0 +1,48 @@
...
+    SDLGamepad();
+    ~SDLGamepad();
+
+    virtual bool InitDevice(int number, std::map<std::string, std::vector<KeyMap::KeyTarget>> keyMap) override;
```
lioncash: `virtual` isn't necessary if `override` is present.

src/input_core/devices/SDLGamepad.cpp
```diff
@@ -0,0 +1,86 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <memory>
+#include <cmath>
+
+#include "SDLGamepad.h"
+#include <SDL.h>
```
lioncash: This include should be up with the other `<>` headers as well.

src/input_core/devices/SDLGamepad.cpp
```diff
@@ -0,0 +1,86 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <memory>
+#include <cmath>
+
+#include "SDLGamepad.h"
```
lioncash:
```cpp
#include "input_core/devices/SDLGamepad.cpp"
```

Which would come after the `common/` includes.

src/input_core/devices/SDLGamepad.h
```diff
@@ -0,0 +1,48 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <SDL_gamecontroller.h>
+#include "IDevice.h"
```
lioncash:
```cpp
#include "input_core/devices/IDevice.h"
```

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <map>
+
+#include "core/core_timing.h"
+
+#include "input_core/input_core.h"
+#include "input_core/devices/Keyboard.h"
+#include "input_core/devices/SDLGamepad.h"
+
+namespace InputCore {
+    using std::vector;
```
lioncash: Please don't. Especially for trivial containers like these.

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,107 @@
...
+            input->InitDevice(mapping.number, keyMapping);
+        }
+        //init keyboard, if it hasn't already
+        if (main_keyboard == nullptr)
+            main_keyboard = std::make_shared<Keyboard>();
+
+        return devices;
+    }
+
+    bool CheckIfMappingExists(vector<Settings::InputDeviceMapping> uniqueMapping, Settings::InputDeviceMapping mappingToCheck) {
```
lioncash: You should be taking the vector by const reference here.

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,107 @@
...
+        for (auto& mapping : uniqueMapping) {
```
lioncash: The whole function body can just be:

```cpp
return std::any_of(uniqueMapping.begin(), uniqueMapping.end(), [mappingToCheck](const auto& mapping) {
    return mapping == mappingToCheck;
});
```

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,107 @@
...
+                }
+                else if (mapping.device == Settings::Device::Gamepad) {
+                    input = std::make_shared<SDLGamepad>();
+                    break;
+                }
+            }
+            }
+            devices.push_back(input);
+
+            //Build list of inputs to listen for, for this device
```
lioncash: This should be its own function.

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,107 @@
...
+            std::map<std::string, vector<KeyMap::KeyTarget>> keyMapping;
+            for (int i = 0; i < Settings::values.input_mappings.size(); i++) {
```
lioncash: You can use a range-based for loop here.

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    void Shutdown() {
+        devices.clear();
+    }
+
+    vector<shared_ptr<IDevice>> ParseSettings() {
+        vector<shared_ptr<IDevice>> devices;
+        vector<Settings::InputDeviceMapping> uniqueMappings; //unique mappings from settings file, used to init devices.
+
+        //Get Unique input mappings from settings
```
lioncash: This should be its own function

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,107 @@
...
+        for (auto& mapping : Settings::values.input_mappings) {
+            if (!CheckIfMappingExists(uniqueMappings, mapping)) {
+                uniqueMappings.push_back(mapping);
+            }
+        }
+
+        //Generate a device for each unique mapping
```
lioncash: This should be its own function.

src/input_core/key_map.cpp
```diff
@@ -0,0 +1,63 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <map>
+#include <algorithm>
```
lioncash: `<map>` should come after both of these header includes.

src/input_core/key_map.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <iterator>
+
+#include "common/emu_window.h"
+
+#include "input_core/key_map.h"
+#include "input_core/input_core.h"
```
lioncash: `i` comes before `k`

src/input_core/devices/SDLGamepad.cpp
```diff
@@ -0,0 +1,86 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <memory>
+#include <cmath>
```
lioncash: `c` comes before `m`

src/input_core/devices/Keyboard.h
```diff
@@ -0,0 +1,48 @@
...
+        return (this->scancode == other.scancode);
+    }
+    bool operator==(uint32_t other) {
```
lioncash:
```cpp
bool operator==(uint32_t other) const
```

src/input_core/devices/Keyboard.cpp
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+}
+
+bool Keyboard::CloseDevice() {
+    return true;
+}
+
+void Keyboard::KeyPressed(KeyboardKey key) {
+    m.lock();
```
lioncash: You can use a lock_guard here.

src/input_core/devices/Keyboard.cpp
```diff
@@ -0,0 +1,55 @@
...
+    keysPressed[key] = true;
+    m.unlock();
+}
+
+void Keyboard::KeyReleased(KeyboardKey key) {
+    m.lock();
```
lioncash: Same here

src/input_core/devices/Keyboard.cpp
```diff
@@ -0,0 +1,55 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "Keyboard.h"
```
lioncash:
```cpp
#include "input_core/devices/Keyboard.h"
```

src/input_core/devices/Keyboard.h
```diff
@@ -0,0 +1,48 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <map>
+#include <memory>
+#include <mutex>
+
+#include "IDevice.h"
```
lioncash:
```cpp
#include "input_core/devices/IDevice.h"
```

lioncash:
Just a suggestion, but you may want to consider a design that doesn't expose globals as part of its API.

makotech222:
Is that regarding inputcore.h? I asked if i should change it to singleton. Do you agree?

Subv:
So, say I wanted to add a Bluetooth gamepad that uses a custom protocol, i'd just have to add another "DeviceFramework" to the enum, create a CustomBluetoothGamepad class, and report inputs in ProcessInput?

makotech222:
@Subv  Yeah pretty much thats the gist of it. You will also need to set up the GUI form to handle assigning the input. It should be pretty simple, if it goes as well as I think it will, when I implement it.

src/input_core/input_core.cpp
makotech222: No i need the ith element for more than one collection.

src/input_core/input_core.cpp
makotech222: I don't see the point of doing that? It's only purpose is inside this ParseSettings function, and it will just pollute the namespace. Same for other 2 below.

makotech222:
So does anyone have any opinions on pulling the input tab gui out and putting it on its own form?

neobrain:

2016-08-06T17:10:08Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:555556bb0f7215fce25bc04dc352d4f6c2111442

wwylele:
@makotech222 As per #1182 I think we still want all configuration in one dialog. We can make it bigger, containing sub tab or so

src/input_core/input_core.cpp
lioncash: The fact you have to comment to explain *what* you're doing, and not *why* you're doing it, is indicative enough that you should be breaking them out into their own function.

It also makes it easier to refactor in the future, because you can just move the function call around instead of having to cut and past blocks of code (which is also more error prone).

> it will just pollute the namespace

It wouldn't pollute the namespace, because those functions don't need to exposed in the public interface. They can just be static functions.

lioncash:
Review status: 0 of 28 files reviewed at latest revision, 37 unresolved discussions, some commit checks failed.

---

*[src/core/settings.h, line 59 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVvA3Sbb0uZ7YbNvBi:-KOVvA3Sbb0uZ7YbNvBj:bj8o1wr) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/core/settings.h#L59)):*
> ```objc
>         Device device = Device::Keyboard;
>         std::string key;
>         InputDeviceMapping() { }
> ```

You should have a newline before the constructor. The constructor can also be defaulted:

```cpp
InputDeviceMappint() = default;
```

---

*[src/core/settings.h, line 60 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVwk3dDUINXcClf4Oh:-KOVwk3eibtcT-9v6Z8q:bm9gxla) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/core/settings.h#L60)):*
> ```objc
>         std::string key;
>         InputDeviceMapping() { }
>         InputDeviceMapping(std::string input) {
> ```

This string should be taken by const reference.

---

*[src/core/settings.h, line 77 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVvyS64_npOWJf7Mr-:-KOVvyS64_npOWJf7Mr0:bsrtktk) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/core/settings.h#L77)):*
> ```objc
>                 key = parts[3];
>             }
>             else {
> ```

The `else` case here can be eliminated, since the elements will already have their defaults assigned to them.

You can essentially change the constructor into

```cpp
InputDevice(const std::string& input) {
    std::vector<std::string> parts;
    Common::SplitString(input, '/', parts);

    if (parts.size() != 4)
        return;

    // code that was in the if-block here
}
```

---

*[src/core/settings.h, line 89 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVwuDl9-Tbg-8AhsZt:-KOVwuDnBEHugKWbWfVJ:b-ly2hx0) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/core/settings.h#L89)):*
> ```objc
>             return (this->device == rhs.device) && (this->framework == rhs.framework) && (this->number == rhs.number);
>         }
>         std::string ToString() const{
> ```

Space before the brace.

---

*[src/core/settings.h, line 91 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVx4LJWf6kOohq2Yis:-KOVx4LK5CWFBtmQeNab:blnav3e) ([raw file](https://github.com/citra-emu/citra/blob/e91327c86a863b9419fd3695b2400a52336ec3b5/src/core/settings.h#L91)):*
> ```objc
> 
> void Apply();
> ```

It's not necessary to remove the newline here.

---

*[src/core/settings.h, line 103 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVx--OJdjioI1XJbyD:-KOVx--OJdjioI1XJbyE:bhot4zx) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/core/settings.h#L103)):*
> ```objc
>                 result += "Keyboard";
>             else if (device == Device::Gamepad)
>                 result += "Gamepad";
> ```

newline after this line.

---

*[src/input_core/input_core.cpp, line 97 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVxfQD_wrzVNUVvGT8:-KOVxfQEdtQqNMMOoik-:b-ks8di8) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/input_core.cpp#L97)):*
> ```cpp
> }
> 
> bool CheckIfMappingExists(const std::vector<Settings::InputDeviceMapping> uniqueMapping, Settings::InputDeviceMapping mappingToCheck) {
> ```

You're missing the ampersand to make this a const reference.

---

*[src/input_core/input_core.h, line 6 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVxncWsJ3WEEB0OFov:-KOVxncWsJ3WEEB0OFow:b-m3y34y) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/input_core.h#L6)):*
> ```objc
> 
> #pragma once
> #include <memory>
> ```

Newline after pragma once.

---

*[src/input_core/input_core.h, line 16 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVxl204LoIilYKN92g:-KOVxl204LoIilYKN92h:bdybcyw) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/input_core.h#L16)):*
> ```objc
> 
> namespace InputCore {
>     extern Service::HID::PadState pad_state;
> ```

These should have getters or other means of using them that cement them as part of the function interface.

---

*[src/input_core/input_core.h, line 24 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVxhs57ev_1PWeqMTm:-KOVxhs6EtFXjoE5185k:bv8f4q8) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/input_core.h#L24)):*
> ```objc
>     void Shutdown();
> 
>     ///Read settings to initialize devices
> ```

Space after the forward slashes.

---

*[src/input_core/key_map.cpp, line 37 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVy0IwkUUN0ytoJH0l:-KOVy0IwkUUN0ytoJH0m:b-htw8wo) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/key_map.cpp#L37)):*
> ```cpp
> 
>     void PressKey(KeyTarget target, const float strength) {
>         if (std::find(std::begin(analog_inputs), std::end(analog_inputs), target) == std::end(analog_inputs)) { // If is digital keytarget
> ```

The comment should be on a line above the `if`.

---

*[src/input_core/key_map.cpp, line 51 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVy8O7t4FMRolOcHvy:-KOVy8O7t4FMRolOcHvz:b-89a4l7) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/key_map.cpp#L51)):*
> ```cpp
> 
>     void ReleaseKey(KeyTarget target) {
>         if (std::find(std::begin(analog_inputs), std::end(analog_inputs), target) == std::end(analog_inputs)) { // If is digital keytarget
> ```

Ditto

---

*[src/input_core/devices/IDevice.h, line 15 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW1Kdy_0-fE89nCCg3:-KOW1Kdy_0-fE89nCCg4:b3i50gj) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/IDevice.h#L15)):*
> ```objc
> public:
>     virtual ~IDevice() = default;
>     std::map<std::string, std::vector<KeyMap::KeyTarget>> keyMapping; /// Maps the string in the settings file to the HID Padstate object
> ```

This variable should likely not be exposed as a public variable.

If it's needed in this class, it should be protected with accessors.

---

*[src/input_core/devices/IDevice.h, line 17 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW1csrMI_mUji55iL7:-KOW1csrMI_mUji55iL8:bkzzp2) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/IDevice.h#L17)):*
> ```objc
>     std::map<std::string, std::vector<KeyMap::KeyTarget>> keyMapping; /// Maps the string in the settings file to the HID Padstate object
> 
>     virtual bool InitDevice(int number, std::map<std::string, std::vector<KeyMap::KeyTarget>> keyMap) = 0;
> ```

You may want to consider passing the map by const reference here.

---

*[src/input_core/devices/Keyboard.cpp, line 23 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVyLqveeWAtUyX0iRN:-KOVyLqveeWAtUyX0iRO:b1vaw8o) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.cpp#L23)):*
> ```cpp
>     std::lock_guard<std::mutex> lock(m);
>     auto keysPressedCopy = keysPressed;
>     lock.~lock_guard();
> ```

You can just use a [std::unique_lock](http://en.cppreference.com/w/cpp/thread/unique_lock) instead of calling the destructor explicitly.

---

*[src/input_core/devices/Keyboard.cpp, line 24 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVyhXS7uTv9Qvt5aHQ:-KOVyhXS7uTv9Qvt5aHR:b2w71f3) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.cpp#L24)):*
> ```cpp
>     auto keysPressedCopy = keysPressed;
>     lock.~lock_guard();
>     for (auto const &ent1 : keyMapping) {
> ```

`const auto& entry`

---

*[src/input_core/devices/Keyboard.cpp, line 28 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVyrvaFsE267eX-G6S:-KOVyrvaFsE267eX-G6T:b9m4i5e) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.cpp#L28)):*
> ```cpp
>         KeyboardKey proxy = KeyboardKey(0, scancode, "");
>         if (keysPressedCopy[proxy] == true && keysPressedLast[scancode] == false) {
>             for (auto& key : ent1.second) {
> ```

`const auto& key`

---

*[src/input_core/devices/Keyboard.cpp, line 34 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVyqBfUZFg6BQ46K0K:-KOVyqBlIc1lKmQpJ96W:b9m4i5e) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.cpp#L34)):*
> ```cpp
>         }
>         else if (keysPressedCopy[proxy] == false && keysPressedLast[scancode] == true) {
>             for (auto& key : ent1.second) {
> ```

`const auto& key`

---

*[src/input_core/devices/Keyboard.h, line 33 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVzeUKsFWn1rSKFy1F:-KOVzeUKsFWn1rSKFy1G:bi8scrd) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.h#L33)):*
> ```objc
>     uint32_t key;
>     uint32_t scancode;
>     std::string character;
> ```

newline after the members.

---

*[src/input_core/devices/Keyboard.h, line 34 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVzh0WPLC9KBdPl8JE:-KOVzh0WPLC9KBdPl8JF:bmv2yma) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.h#L34)):*
> ```objc
>     uint32_t scancode;
>     std::string character;
>     KeyboardKey(uint32_t Key, uint32_t Scancode, std::string Character) {
> ```

```cpp
KeyboardKey(KeyboardKey(uint32_t key_, uint32_t scancode_, std::string character_) 
    : key(key_), scancode(scancode_), character(std::move(character))
{
}
```

---

*[src/input_core/devices/Keyboard.h, line 40 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-290RjHSCCo0NXJj:-KOW-290RjHSCCo0NXJk:b-x2oaw7) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.h#L40)):*
> ```objc
>     }
>     bool operator==(const KeyboardKey& other) const {
>         return (this->scancode == other.scancode);
> ```

```cpp
return scancode == other.scancode;
```

---

*[src/input_core/devices/Keyboard.h, line 45 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-BM933Xj8ddx74VZ:-KOW-BMAGnCZDTvUDLXn:bckeviq) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.h#L45)):*
> ```objc
>         return (this->scancode == other);
>     }
>     bool operator<(const KeyboardKey &o)  const {
> ```

ampersand against the type. It should also be named `other` like the operator above.

---

*[src/input_core/devices/Keyboard.h, line 46 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-F5ahYLim4oUH91f:-KOW-F5ahYLim4oUH91g:b-f28emu) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.h#L46)):*
> ```objc
>     }
>     bool operator<(const KeyboardKey &o)  const {
>         return (this->scancode < o.scancode);
> ```

```cpp
 return scancode < o.scancode;
```

---

*[src/input_core/devices/SDLGamepad.cpp, line 39 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-ygDkkLYdy_n-bej:-KOW-ygDkkLYdy_n-bek:b-8h0dfk) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.cpp#L39)):*
> ```cpp
>     }
>     keyMapping = keyMap;
>     for (auto& entry : keyMapping) {
> ```

```cpp
const auto& entry
```

---

*[src/input_core/devices/SDLGamepad.cpp, line 50 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-bFzXr7_8YAoDaFc:-KOW-bG-8rBriwd3NGlF:b-8h0dfk) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.cpp#L50)):*
> ```cpp
>         return;
>     SDL_GameControllerUpdate();
>     for (auto const &ent1 : keyMapping) {
> ```

```cpp
const auto& entry
```

---

*[src/input_core/devices/SDLGamepad.cpp, line 55 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-fYWFltBd94y_Fa6:-KOW-fYXgpsBZfg0N5yy:bdcclv2) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.cpp#L55)):*
> ```cpp
>             Uint8 pressed = SDL_GameControllerGetButton(gamepad, button);
>             if (pressed == 1 && keysPressed[ent1.first] == false) {
>                 for (auto& padstate : ent1.second) {
> ```

```cpp
const auto&
```

---

*[src/input_core/devices/SDLGamepad.cpp, line 61 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-j3NbTbkcUT41F2N:-KOW-j3NbTbkcUT41F2O:bdcclv2) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.cpp#L61)):*
> ```cpp
>             }
>             else if (pressed == 0 && keysPressed[ent1.first] == true) {
>                 for (auto& padstate : ent1.second) {
> ```

```cpp
const auto&
```

---

*[src/input_core/devices/SDLGamepad.cpp, line 68 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-obXcIOCMZiP3kNZ:-KOW-obajIsbxEm5GsHY:b-gh31jx) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.cpp#L68)):*
> ```cpp
>         }
>         else {
>             //Try axis if button isn't valid
> ```

Space after `//`

---

*[src/input_core/devices/SDLGamepad.cpp, line 72 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-lxSmx0Zi2eYjHhZ:-KOW-lxSmx0Zi2eYjHh_:bdcclv2) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.cpp#L72)):*
> ```cpp
>             if (axis != SDL_GameControllerAxis::SDL_CONTROLLER_AXIS_INVALID) {
>                 Sint16 value = SDL_GameControllerGetAxis(gamepad, axis);
>                 for (auto& padstate : ent1.second) {
> ```

```cpp
const auto&
```

---

*[src/input_core/devices/SDLGamepad.cpp, line 73 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-rMPHH0XbjJWZtrp:-KOW-rMPHH0XbjJWZtrq:b-vv6h41) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.cpp#L73)):*
> ```cpp
>                 Sint16 value = SDL_GameControllerGetAxis(gamepad, axis);
>                 for (auto& padstate : ent1.second) {
>                     if (abs(value) < 0.2 * 32767.0) // dont process if in deadzone. Replace later with settings for deadzone.
> ```

The comment should go above the if.

---

*[src/input_core/devices/SDLGamepad.h, line 20 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW0DyHdGF7B5A3XwXR:-KOW0DyIwDBLD8RCKrSE:bm0nrai) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.h#L20)):*
> ```objc
>     bool CloseDevice() override;
> private:
>     std::map<std::string, std::string> friendlyNameMapping = { /// Maps the friendly name shown on GUI with the string name for getting the SDL button instance.
> ```

The comment should go above on its own line.

---

*[src/input_core/devices/SDLGamepad.h, line 48 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW0dhSFuTWxKvyIwrM:-KOW0dhTJtTsGybTPIzy:bhg5spi) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.h#L48)):*
> ```objc
>     static bool SDLInitialized;
>     std::map<std::string, bool> keysPressed;
>     SDL_GameController* gamepad;
> ```

This should likely be

```cpp
SDL_GameController* gamepad = nullptr;
```

to fully prevent cases where freeing an uninitialized variable can happen in the destructor.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/1986)*

src/input_core/devices/SDLGamepad.h
```diff
@@ -0,0 +1,48 @@
...
+
+class SDLGamepad : public IDevice {
+private:
+    std::map<std::string, std::string> friendlyNameMapping = { /// Maps the friendly name shown on GUI with the string name for getting the SDL button instance.
```
neobrain: How will you handle translation?

makotech222: Hmm i never thought about that. I didn't know we had translations. Any ideas?

wwylele: @makotech222 We haven't start translations, but we coded it in mind. As you can see, almost all strings in `ctra-qt` are wrapped by `tr(...)`, which is a function provided by Qt and which can handle translation automatically in the future.

makotech222: Okay so i will wrap them all friendly strings in tr().

makotech222:
@lioncash  regarding unique_lock:

I want to unlock the mutex immediately after the copy. How does using unique lock change that?

neobrain:

2016-08-06T20:41:28Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:743c4647b83187421464370c9a85c670d443b9aa

lioncash:
@makotech222 Oh right, that was silly of me. You can just do the following:

```cpp
std::map<KeyboardKey, bool>keysPressedCopy;
{
   std::lock_guard<std::mutex> lock(m);
   keysPressedCopy = keysPressed;
}
```

neobrain:

2016-08-06T20:47:37Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:750838581de54275d0635fe6729722f334309c67

lioncash:
Review status: 0 of 28 files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

*[src/citra/emu_window/emu_window_sdl2.cpp, line 178 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWkh9eazJM-4GbIAbO:-KOWkh9eazJM-4GbIAbP:bxonqkm) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/citra/emu_window/emu_window_sdl2.cpp#L178)):*
> ```cpp
> 
> void EmuWindow_SDL2::ReloadSetKeymaps() {
>     //KeyMap::ClearKeyMapping();
> ```

This should be removed if unused and be reintroduced whenever it's required again.

---

*[src/citra_qt/configure_input.cpp, line 39 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWfT8HGxfndUNIxImK:-KOWfT8HGxfndUNIxImL:bcxisgw) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/citra_qt/configure_input.cpp#L39)):*
> ```cpp
>         { std::make_pair(Settings::NativeInput::Values::CIRCLE_LEFT, ui->buttonCircleLeft) },
>         { std::make_pair(Settings::NativeInput::Values::CIRCLE_RIGHT, ui->buttonCircleRight) }
>         //{ std::make_pair(Settings::NativeInput::Values::CIRCLE_MODIFIER, ui->buttonCircleMod) },
> ```

If these aren't necessary they should be removed and added back in a change that makes use of them. Ditto for the rest of this file.

---

*[src/input_core/input_core.cpp, line 17 \[r6\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWcdq4tk6R6IO8-VnI:-KOWcdq5D8uzqwC6OcGg:b-jix8ip) ([raw file](https://github.com/citra-emu/citra/blob/743c4647b83187421464370c9a85c670d443b9aa/src/input_core/input_core.cpp#L17)):*
> ```cpp
> constexpr u64 frame_ticks = 268123480ull / 60;
> static int tick_event;
> Service::HID::PadState pad_state;
> ```

`pad_state`, `circle_pad`, `main_keyboard` and `devices` can all be `static` as well.

---

*[src/input_core/input_core.cpp, line 46 \[r6\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWdmsllBF6GOpWMIg2:-KOWdmsmj4Ux9bDpxYo7:b-3t57pm) ([raw file](https://github.com/citra-emu/citra/blob/743c4647b83187421464370c9a85c670d443b9aa/src/input_core/input_core.cpp#L46)):*
> ```cpp
> }
> 
> void SetPadState(Service::HID::PadState& state) {
> ```

`state` can be a const reference.

---

*[src/input_core/input_core.cpp, line 54 \[r6\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWdqVAbPkncsbtZlqv:-KOWdqVAbPkncsbtZlqw:b-ojhvpm) ([raw file](https://github.com/citra-emu/citra/blob/743c4647b83187421464370c9a85c670d443b9aa/src/input_core/input_core.cpp#L54)):*
> ```cpp
> }
> 
> void SetCirclePad(std::tuple<s16, s16>& pad) {
> ```

Taking the parameter by reference here isn't necessary, since the object size is less than that of a pointer.

---

*[src/input_core/input_core.cpp, line 66 \[r6\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWdylchXWj2tNtSoWt:-KOWdylchXWj2tNtSoWu:b-yx1ni7) ([raw file](https://github.com/citra-emu/citra/blob/743c4647b83187421464370c9a85c670d443b9aa/src/input_core/input_core.cpp#L66)):*
> ```cpp
>     std::vector<Settings::InputDeviceMapping> uniqueMappings;
> 
>     for (auto& mapping : Settings::values.input_mappings) {
> ```

`const auto&`

---

*[src/input_core/input_core.cpp, line 76 \[r6\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWbbjM8AbkW2u1o3a5:-KOWbbjNsCVwuAjkt0D7:b-d6n9pu) ([raw file](https://github.com/citra-emu/citra/blob/743c4647b83187421464370c9a85c670d443b9aa/src/input_core/input_core.cpp#L76)):*
> ```cpp
> std::map<std::string, std::vector<KeyMap::KeyTarget>> BuildKeyMapping(Settings::InputDeviceMapping mapping) {
>     std::map<std::string, std::vector<KeyMap::KeyTarget>> keyMapping;
>     for (int i = 0; i < Settings::values.input_mappings.size(); i++) {
> ```

`int` should be a `size_t` here, otherwise you get truncation warnings.

---

*[src/input_core/input_core.cpp, line 63 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWizzS2NsyY6_jtbvC:-KOWizzXiQHDrQwq_YVt:b-89fms6) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/input_core.cpp#L63)):*
> ```cpp
> 
> ///Get Unique input mappings from settings
> std::vector<Settings::InputDeviceMapping> GatherUniqueMappings() {
> ```

this should be prefixed with `static`

---

*[src/input_core/input_core.cpp, line 74 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWj4-YrtzTjwr0GrX8:-KOWj4-Z7iQ2PUhnqi0w:bg9799g) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/input_core.cpp#L74)):*
> ```cpp
> }
> 
> std::map<std::string, std::vector<KeyMap::KeyTarget>> BuildKeyMapping(Settings::InputDeviceMapping mapping) {
> ```

This should be prefixed with `static`.

---

*[src/input_core/input_core.cpp, line 86 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWjMjLx2vzAAa-4LJo:-KOWjMjLx2vzAAa-4LJp:b-dhtzf6) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/input_core.cpp#L86)):*
> ```cpp
> }
> 
> ///Generate a device for each unique mapping
> ```

Space after `///`

---

*[src/input_core/input_core.cpp, line 87 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWgRIcvcKZLlrGoWcb:-KOWgRIcvcKZLlrGoWcc:b-4af1al) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/input_core.cpp#L87)):*
> ```cpp
> 
> ///Generate a device for each unique mapping
> void GenerateUniqueDevices(std::vector<Settings::InputDeviceMapping> uniqueMappings) {
> ```

This should take `uniqueMappings` via const reference. The function signature can also be prefixed with`static`

---

*[src/input_core/input_core.cpp, line 89 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWgO1i01h6tp8UUXKu:-KOWgO1jSMQRGdRGeI9G:bdcclv2) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/input_core.cpp#L89)):*
> ```cpp
> void GenerateUniqueDevices(std::vector<Settings::InputDeviceMapping> uniqueMappings) {
>     std::shared_ptr<IDevice> input;
>     for (auto& mapping : uniqueMappings) {
> ```

```cpp
const auto&
```

---

*[src/input_core/input_core.cpp, line 112 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWjKFsZiF226w-3jw_:-KOWjKFsZiF226w-3jwa:b-gh31jx) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/input_core.cpp#L112)):*
> ```cpp
>         devices.push_back(input);
> 
>         //Build map of inputs to listen for, for this device
> ```

Space after `//`

---

*[src/input_core/input_core.cpp, line 119 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWgr12dj6u7hjUgI71:-KOWgr13KEcrotTlSY7Z:bm4t57v) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/input_core.cpp#L119)):*
> ```cpp
> }
> void ParseSettings() {
>     std::vector<std::shared_ptr<IDevice>> devices;
> ```

`devices` isn't used anywhere here.

---

*[src/input_core/input_core.h, line 16 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWhTs_Ttm_tH5V3tEh:-KOWhTs_Ttm_tH5V3tEi:beti2v0) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/input_core.h#L16)):*
> ```objc
> class Keyboard;
> 
> namespace InputCore {
> ```

We don't indent contents of namespaces.

---

*[src/input_core/input_core.h, line 22 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWjzODYVjA7SlW8BeD:-KOWjzOEt-0a5dnigRMK:bv5jzo0) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/input_core.h#L22)):*
> ```objc
>     void SetPadState(Service::HID::PadState& state);
>     std::tuple<s16, s16> GetCirclePad();
>     void SetCirclePad(std::tuple<s16, s16>& circle );
> ```

one space too many after `circle`

---

*[src/input_core/key_map.cpp, line 14 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWhA9tragniYGwo5rf:-KOWhA9urdgmVf3A0Lx4:beti2v0) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/key_map.cpp#L14)):*
> ```cpp
> #include "input_core/key_map.h"
> 
> namespace KeyMap {
> ```

We don't indent contents of namespaces.

---

*[src/input_core/key_map.h, line 14 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWhG4Ypbfuuyyfc8by:-KOWhG4ZUPmKDvCpOaU-:beti2v0) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/key_map.h#L14)):*
> ```objc
> class EmuWindow;
> 
> namespace KeyMap {
> ```

We don't indent contents of namespaces.

---

*[src/input_core/devices/IDevice.h, line 21 \[r7\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOWhdGX-MNMrygDsSGk:-KOWhdGbKQy9whs6XGGX:b-5t76iq) ([raw file](https://github.com/citra-emu/citra/blob/750838581de54275d0635fe6729722f334309c67/src/input_core/devices/IDevice.h#L21)):*
> ```objc
> protected:
>     std::map<std::string, std::vector<KeyMap::KeyTarget>> keyMapping; ///< Maps the string in the settings file to the HID Padstate object
> ```

The newline here isn't necessary.

---

*[src/input_core/devices/Keyboard.cpp, line 24 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVyhXS7uTv9Qvt5aHQ:-KOWeQd12BQAnTT6BwS7:b3qgp1k) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.cpp#L24)):*
> `const auto& entry`
</details>
ampersand against the `auto` also you can just use `entry` to be clearer, as opposed to something that doesn't need a number on it.

---

*[src/input_core/devices/Keyboard.h, line 34 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOVzh0WPLC9KBdPl8JE:-KOWhqSt9kQfqBf39SDc:b-19zur6) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/Keyboard.h#L34)):*
> ```cpp
> KeyboardKey(KeyboardKey(uint32_t key_, uint32_t scancode_, std::string character_) 
>     : key(key_), scancode(scancode_), character(std::move(character))
> {
> }
> ```
</details>
The initializer list should be on its own line, like I showed you.

---

*[src/input_core/devices/SDLGamepad.cpp, line 50 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW-bFzXr7_8YAoDaFc:-KOWf2qMTU66vFhZ0zU_:b-8pwdc3) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.cpp#L50)):*
> ```cpp
> const auto& entry
> ```
</details>
ampersand against the `auto` also `ent1` -> `entry`

---

*[src/input_core/devices/SDLGamepad.h, line 20 \[r5\]](https://reviewable.io:443/reviews/citra-emu/citra/1986#-KOW0DyHdGF7B5A3XwXR:-KOWfLAdcQ0ftegVzrgO:byqrbp3) ([raw file](https://github.com/citra-emu/citra/blob/555556bb0f7215fce25bc04dc352d4f6c2111442/src/input_core/devices/SDLGamepad.h#L20)):*
> The comment should go above on its own line.
</details>
This still applies

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/1986)*

neobrain:

2016-08-06T22:09:43Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:6ad7260471271859c7bae01f637580ef4955334d

makotech222:
@lioncash Thank you for taking the time to look through my code. Regarding any GUI or the circle pad modifier changes, they will be made when I have time to implement them fully (possibly today or tomorrow). I will make a new PR when those are complete and everything is ready to check in.

makotech222:
Okay closing this PR and opening up an official one. Work is pretty much done.

PR #1987, opened 2016-07-30
fix #1942 and add a few IPC functions for descriptors

### LINKED ISSUE #1942: IPC: Copy/MoveHandleDesc are wrong
(assuming 3dbrew is correct)
Our [`CopyHandleDesc` and `MoveHandleDesc`](https://github.com/citra-emu/citra/blob/6fa7b1f4863a2e6bde07c3f7ddca3ac52da39757/src/core/hle/kernel/session.h#L23-29) are swapped, comparing with [them on 3dbrew](https://www.3dbrew.org/wiki/IPC#Handle_Translation). We need to check all the code using them if the handle is leaked.

Lectem:
All the MoveHandleDesc and CopyHandleDesc usages were checked on hardware (client side).
TransferHandleDesc has been removed because it was wrong and duplicated CallingPidDesc.

I also added some utilities for a future IPC helper.

### DIFF
--- a/src/core/hle/kernel/session.h
+++ b/src/core/hle/kernel/session.h
@@ -16,41 +16,129 @@
 
 namespace IPC {
 
-constexpr u32 MakeHeader(u16 command_id, unsigned int regular_params, unsigned int translate_params) {
-    return ((u32)command_id << 16) | (((u32)regular_params & 0x3F) << 6) | (((u32)translate_params & 0x3F) << 0);
+enum DescriptorType : u32 {
+    // Buffer related desciptors types (mask : 0x0F)
+    StaticBuffer = 0x02,
+    PXIBuffer    = 0x04,
+    MappedBuffer = 0x08,
+    // Handle related descriptors types (mask : 0x30, but need to check for buffer related descriptors first )
+    CopyHandle   = 0x00,
+    MoveHandle   = 0x10,
+    CallingPid   = 0x20,
+};
+
+/**
+ * @brief Creates a command header to be used for IPC
+ * @param command_id            ID of the command to create a header for.
+ * @param normal_params         Size of the normal parameters in words. Up to 63.
+ * @param translate_params_size Size of the translate parameters in words. Up to 63.
+ * @return The created IPC header.
+ *
+ * Normal parameters are sent directly to the process while the translate parameters might go through modifications and checks by the kernel.
+ * The translate parameters are described by headers generated with the IPC::*Desc functions.
+ *
+ * @note While #normal_params is equivalent to the number of normal parameters, #translate_params_size includes the size occupied by the translate parameters headers.
+ */
+constexpr u32 MakeHeader(u16 command_id, unsigned int normal_params, unsigned int translate_params_size) {
+    return (u32(command_id) << 16) | ((u32(normal_params) & 0x3F) << 6) | (u32(translate_params_size) & 0x3F);
 }
 
-constexpr u32 MoveHandleDesc(unsigned int num_handles = 1) {
-    return 0x0 | ((num_handles - 1) << 26);
+union Header {
+    u32 raw;
+    BitField< 0, 6, u32> translate_params_size;
+    BitField< 6, 6, u32> normal_params;
+    BitField<16, 16, u32> command_id;
+};
+
+inline Header ParseHeader(u32 header) {
+    return{ header };
 }
 
-constexpr u32 CopyHandleDesc(unsigned int num_handles = 1) {
-    return 0x10 | ((num_handles - 1) << 26);
+constexpr u32 MoveHandleDesc(u32 num_handles = 1) {
+    return MoveHandle | ((num_handles - 1) << 26);
+}
+
+constexpr u32 CopyHandleDesc(u32 num_handles = 1) {
+    return CopyHandle | ((num_handles - 1) << 26);
 }
 
 constexpr u32 CallingPidDesc() {
-    return 0x20;
+    return CallingPid;
 }
 
-constexpr u32 TransferHandleDesc() {
-    return 0x20;
+constexpr bool isHandleDescriptor(u32 descriptor) {
+    return (descriptor & 0xF) == 0x0;
 }
 
-constexpr u32 StaticBufferDesc(u32 size, unsigned int buffer_id) {
-    return 0x2 | (size << 14) | ((buffer_id & 0xF) << 10);
+constexpr u32 HandleNumberFromDesc(u32 handle_descriptor) {
+    return (handle_descriptor >> 26) + 1;
+}
+
+constexpr u32 StaticBufferDesc(u32 size, u8 buffer_id) {
+    return StaticBuffer | (size << 14) | ((buffer_id & 0xF) << 10);
+}
+
+union StaticBufferDescInfo {
+    u32 raw;
+    BitField< 10, 4, u32> buffer_id;
+    BitField< 14, 18, u32> size;
+};
+
+inline StaticBufferDescInfo ParseStaticBufferDesc(const u32 desc) {
+    return{ desc };
+}
+
+/**
+ * @brief Creates a header describing a buffer to be sent over PXI.
+ * @param size         Size of the buffer. Max 0x00FFFFFF.
+ * @param buffer_id    The Id of the buffer. Max 0xF.
+ * @param is_read_only true if the buffer is read-only. If false, the buffer is considered to have read-write access.
+ * @return The created PXI buffer header.
+ *
+ * The next value is a phys-address of a table located in the BASE memregion.
+ */
+inline u32 PXIBufferDesc(u32 size, unsigned buffer_id, bool is_read_only) {
+    u32 type = PXIBuffer;
+    if (is_read_only) type |= 0x2;
+    return  type | (size << 8) | ((buffer_id & 0xF) << 4);
 }
 
 enum MappedBufferPermissions {
-    R = 2,
-    W = 4,
+    R = 1,
+    W = 2,
     RW = R | W,
 };
 
 constexpr u32 MappedBufferDesc(u32 size, MappedBufferPermissions perms) {
-    return 0x8 | (size << 4) | (u32)perms;
+    return MappedBuffer | (size << 4) | (u32(perms) << 1);
 }
 
+union MappedBufferDescInfo {
+    u32 raw;
+    BitField< 4, 28, u32> size;
+    BitField< 1, 2, MappedBufferPermissions> perms;
+};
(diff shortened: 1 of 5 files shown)


neobrain:

2016-07-30T17:04:50Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...Lectem:8172decc88d19db4d2b4f8f2fac7ad4b2308bc88

neobrain:

2016-07-30T17:09:05Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...Lectem:de711320d8c3e23fcf7a24c12f6a20d6bfe95ced

src/core/hle/kernel/session.h
```diff
@@ -16,28 +16,92 @@
...
+* @param translate_params_size Size of the translate parameters in words. Up to 63.
+* @return The created IPC header.
+*
+* Normal parameters are sent directly to the process while the translate parameters might go through modifications and checks by the kernel.
+* The translate parameters are described by headers generated with the IPC_Desc_* functions.
+*
+* @note While #normal_params is equivalent to the number of normal parameters, #translate_params_size includes the size occupied by the translate parameters headers.
+*/
+constexpr u32 MakeHeader(u16 command_id, unsigned int normal_params, unsigned int translate_params_size) {
+    return ((u32)command_id << 16) | (((u32)normal_params & 0x3F) << 6) | (((u32)translate_params_size & 0x3F) << 0);
```
lioncash: You can just make `normal_params` and `translate_params_size` have `u32` as their type and only static_cast `command_id`

src/core/hle/kernel/session.h
```diff
@@ -16,28 +16,92 @@
...
+}
+
+inline void ParseHeader(u32 header,u16& command_id, unsigned int& normal_params, unsigned int& translate_params_size) {
```
lioncash: This can likely just return a tuple or a struct representing a parsed header instead of using out parameters.

src/core/hle/kernel/session.h
```diff
@@ -16,28 +16,92 @@
...
+    command_id       = (u16) (header >> 16);
+    normal_params = (header >> 6) & 0x3F;
+    translate_params_size = header & 0x3F;
+}
+
+constexpr bool isHandleDescriptor(u32 descriptor) {
+    return (descriptor & 0xF) == 0x0;
+}
+
+inline static DescriptorType GetDescriptorType(u32 descriptor) {
```
lioncash: `static` shouldn't be used on functions in headers otherwise each cpp file that uses it gets its own copy of the function, which is unnecessary.

Also `else` with a sequence of returns isn't necessary, so you can just do something like:

```cpp
inline DescriptorType GetDescriptorType(u32 descriptor) {
    if (isHandleDescriptor(descriptor))
        return static_cast<DescriptorType>(descriptor & 0x30);

    // Handle the fact that the following descriptors can have rights
   if (descriptor & MappedBuffer)
        return MappedBuffer;

   if (descriptor & PXIBuffer)
        return PXIBuffer;

    return StaticBuffer;
}
```

src/core/hle/kernel/session.h
```diff
@@ -16,28 +16,92 @@
...
+/**
+* @brief Creates a header describing a buffer to be sent over PXI.
+* @param size         Size of the buffer. Max 0x00FFFFFF.
+* @param buffer_id    The Id of the buffer. Max 0xF.
+* @param is_read_only true if the buffer is read-only. If false, the buffer is considered to have read-write access.
+* @return The created PXI buffer header.
+*
+* The next value is a phys-address of a table located in the BASE memregion.
+*/
+static inline u32 IPC_Desc_PXIBuffer(u32 size, unsigned buffer_id, bool is_read_only) {
```
lioncash: `static` shouldn't be used

src/core/hle/kernel/session.h
```diff
@@ -16,28 +16,92 @@
...
+    u8 type = PXIBuffer;
```
lioncash: `u8` should be a `u32` here.

src/core/hle/kernel/session.h
```diff
@@ -47,11 +111,16 @@ enum MappedBufferPermissions {
 };
 
 constexpr u32 MappedBufferDesc(u32 size, MappedBufferPermissions perms) {
-    return 0x8 | (size << 4) | (u32)perms;
+    return MappedBuffer | (size << 4) | (u32)perms;
 }
 
+inline void ParseMappedBufferDesc(const u32 desc, u32 &size, MappedBufferPermissions &perms) {
```
lioncash: Like above, this can return a tuple or a struct.

src/core/hle/kernel/session.h
```diff
@@ -16,28 +16,92 @@
...
 
-constexpr u32 TransferHandleDesc() {
-    return 0x20;
+constexpr u32 StaticBufferDesc(u32 size, u8 buffer_id) {
+    return StaticBuffer | (size << 14) | ((buffer_id & 0xF) << 10);
 }
 
-constexpr u32 StaticBufferDesc(u32 size, unsigned int buffer_id) {
-    return 0x2 | (size << 14) | ((buffer_id & 0xF) << 10);
+inline void ParseStaticBufferDesc(const u32 desc, u32 &size, u8 &buffer_id) {
```
lioncash: Like above, this can return a tuple or a struct.

src/core/hle/kernel/session.h
Lectem: well, that's the way it was written before, I only changed the parameters names to be clearer. If you like it I can change it though.
I thought c-style casts were now preferred in case of integer conversions ?

src/core/hle/kernel/session.h
Lectem: what would we gain from it ? You will usually find yourself needing all the arguments, and I don't think std::get<x> looks better.

src/core/hle/kernel/session.h
Lectem: I thought it was better to keep those `else` in this case to convey the idea that those checks need to be in this order.
I agree that it isn't easy on the eye though, so if you insist I'll change it

neobrain:

2016-07-30T21:26:01Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...Lectem:6e9feea853067622c8d7f291da0525351aa731c7

src/core/hle/kernel/session.h
lioncash: Ah OK, I missed the fact this was just moved, disregard this then.

> I thought c-style casts were now preferred in case of integer conversions

You can either `static_cast` it or wrap it the other way like `u32(some_variable)`.

src/core/hle/kernel/session.h
lioncash: > what would we gain from it?

You'd get much more flexibility with how you can use the function.

With a tuple, you aren't only restricted to `std::get`, you can can also use `std::tie` to bind several assignments like:

```cpp
std::tie(var1, var2) = SomeFuncReturningATupleOfSizeTwo();
```

or if you only need a few arguments out of several of them, then you can leverage `std::ignore` like:

```cpp
std::tie(std::ignore, var2) = SomeFuncReturningATupleOfSizeTwo()
```

You aren't stuck being required to make a dummy variable just to get data you want if you only need a few of the total values returned. Whereas with references, you are. When you change them to pointers, then you have to pass nullptr, which introduces unnecessary checks. `std::tuple` has none of those issues.

<br/>
> I'm not really against the struct but is it really worth it to add new structs only for a function ?

You said that

> You will usually find yourself needing all the arguments

which means that those multiple pieces of data are used together as a single cohesive unit. This is something better expressed through a struct/tuple.

src/core/hle/kernel/session.h
lioncash: A comment saying something along the lines of `The ordering of these checks are important` is probably a better conveyor of that.

src/core/hle/kernel/session.h
Lectem: makes sense, what do you think of using BitField here ? or even drop the function and just use the raw member ?

    union Header {
        u32 raw;
        BitField< 0, 6,u32> translate_params_size;
        BitField< 6, 6,u32> normal_params;
        BitField<16,16,u32> command_id;
    };

neobrain:

2016-07-31T23:12:17Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...Lectem:6ec6c6e89dd395939c1e20718c1613f8f068bed8

Lectem:
I addressed the comments, and also removed the static keyword for GetCommandBuffer.
I also changed the DescriptorType enum to specify its underlying type as u32.
I'm not sure if we should keep the Parse* functions or not, as those are just aliases for the constructors of the unions.

src/core/hle/kernel/session.h
```diff
@@ -16,42 +16,130 @@
...
+    PXIBuffer    = 0x04,
+    MappedBuffer = 0x08,
+    // Handle related descriptors types (mask : 0x30, but need to check for buffer related descriptors first )
+    CopyHandle   = 0x00,
+    MoveHandle   = 0x10,
+    CallingPid   = 0x20,
+};
+
+/**
+* @brief Creates a command header to be used for IPC
```
bunnei: These lines should be indented over by one space

neobrain:

2016-08-02T14:50:43Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...Lectem:a94fb932ff8a326a2e85ab3814663af565313ec5

Lectem:
Fixed the comments indentation

Subv:
LGTM

wwylele:
LDTM

src/core/hle/service/srv.cpp
```diff
@@ -57,7 +57,7 @@ static void EnableNotification(Service::Interface* self) {
 
     cmd_buff[0] = IPC::MakeHeader(0x2, 0x1, 0x2); // 0x20042
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    cmd_buff[2] = IPC::TransferHandleDesc();
```
wwylele: I am curious what was the old code going on :joy:

Lectem: Well, it seems the author didnt really understand the IPC descriptors, and comment https://github.com/citra-emu/citra/pull/1867#discussion_r65007273 wasnt addressed. It wasnt harmful since the games dont check those, but still wrong

PR #1995, opened 2016-08-08
Overhaul Input handling into InputCore. Add Gamepad support.

makotech222:
Okay so this is pretty much the finished product. Description:

The code flows like so: First, settings are read from the respective settings file. The previous format for binding keys was just the QT/SDL value for each keyboard key. It now goes as follows:

pad_a=Framework/DeviceID/Device/Button
Example:
pad_a=SDL/0/Gamepad/Button A
or
pad_a=QT/0/Keyboard/70

Note (QT and SDL keyboard are the same)

This pretty much copies how Dolphin does keybinding. From there, inputcore is initialized upon game starting, where it reads the settings values and initializes an IDevice object for each unique mapping. After initializing, input from SDL Gamepads are polled every tick, while keyboard input is still handled from the front end on an event basis. Touch screen input is also handled on front end on event basis.

This pull request as is, fixes
~~#1707~~ (I had to reverse scan code change, due to way Qt handles scan codes. Will have to think of different solution)
#1883 (partially, keyboard and touch screen input is now thread safe)

Since the RFC, changes include:
- Fixed Qt GUI key binding.
- Keyboard now responds to Circle Pad Modifier key. Only works if bound to a keyboard key, though.
- Touch screen input is now thread safe and handled by InputCore.

I think in a future request, we may want to redo the Input GUI, since it needs to eventually handle HotKeys and a circle stick dead-zone setting.


neobrain:

2016-08-08T01:42:40Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:dd9d90c5d7227d8eedebc814af3d4c689e8a0668

src/input_core/devices/Keyboard.h
```diff
@@ -0,0 +1,49 @@
+// Copyright 2016 Citra Emulator Project
```
lioncash: This file still needs to have its name lowercased.

src/input_core/devices/device.h
```diff
@@ -0,0 +1,43 @@
...
+#include "core/hle/service/hid/hid.h"
+#include "core/settings.h"
+#include "input_core/key_map.h"
+
+class IDevice {
+public:
+    virtual ~IDevice() = default;
+
+    /**
+    * Initialize IDevice object with device's index and the map of keys that it will listen to.
```
lioncash: the asterisks should have one more space so that they line up with one another. Documentation comments for functions with parameters or return types should also have `@param` and `@return` tags describing them.

This applies to all other occurrences as well.

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "common/assert.h"
+#include "common/logging/log.h"
+
+#include "input_core/devices/sdl_gamepad.h"
+
+bool SDLGamepad::SDLInitialized = false;
+
+SDLGamepad::SDLGamepad() {
+}
+SDLGamepad::SDLGamepad(int number_, SDL_GameController* gamepad_) :number(number_), gamepad(gamepad_) {
```
lioncash: Initializer list should go on the next line.

src/core/settings.h
```diff
@@ -4,89 +4,150 @@
...
+    enum class Device {
+        Keyboard, Gamepad
+    };
+    struct InputDeviceMapping {
+        DeviceFramework framework = DeviceFramework::Qt;
+        int number = 0;
+        Device device = Device::Keyboard;
+        std::string key;
+
+        InputDeviceMapping() {}
```
lioncash: The headings for the constructor should be as follows:

```cpp
InputDeviceMapping() = default;
explicit InputDeviceMapping(int keyboardKey) {
explicit InputDeviceMapping(const std::string& input) {
```

src/input_core/devices/Keyboard.h
makotech222: Thats weird... I did rename it, somehow it turned back... Will fix.

src/input_core/devices/sdl_gamepad.h
```diff
@@ -0,0 +1,57 @@
...
+    void Clear() override;
+
+    ///Returns vector of all gamepads connected to computer. Used for keybinding setup
+    static std::vector<std::shared_ptr<IDevice>> GetAllDevices();
+private:
+    /// Maps the friendly name shown on GUI with the string name for getting the SDL button instance.
+    std::map<std::string, std::string> friendly_name_mapping = {
+        { "Button A","a" },
+        { "Button B","b" },
+        { "Button X","x" },
```
neobrain: So uh.. I'll have to ask again: How are you going to handle translations?

The point I'm trying to get at is that this has no business at all living in core code. Presentation is a frontend task.

makotech222: Its in the device code because it's not just for the front end, its also necessary when saving to settings file, specifically, for saving the analog stick mappings. On the Qt side, I believe I have used the tr(...) when displaying the strings. I'll have to double check later today.

neobrain: `tr()` has no way of magically recognizing these strings, hence all translation tools will fail to register the strings for translation.

Also note that settings file are a presentation task just as much as the GUI itself. Furthermore, there is no logical reason for using the same string for configuration entries and for GUI texts, since these are two entirely different things.

makotech222: Okay i reread how translations are handled, and I see what you mean. I will think of a different way to handle it.

Regarding the friendly name string, i don't see why they wouldn't be the same for config entry and GUI text. It seems logical to me. If they were different, what would they be?

wwylele: @makotech222 
> i don't see why they wouldn't be the same for config entry and GUI text.

What if user changes the language in GUI? If config entry and GUI text keep the same, changing language will break the existing config file.

src/citra_qt/keybinding_names.h
```diff
@@ -0,0 +1,17 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
```
lioncash: This should be `<array>`

src/core/settings.h
```diff
@@ -4,89 +4,150 @@
...
-    std::string sink_id;
-
-    // Debugging
-    bool use_gdbstub;
-    u16 gdbstub_port;
-} extern values;
-
-void Apply();
-
+    namespace NativeInput {
```
lioncash: We still don't indent on namespaces.

src/core/settings.h
```diff
@@ -4,89 +4,150 @@
...
+
+            if (parts[2] == "Keyboard")
+                device = Device::Keyboard;
+            else if (parts[2] == "Gamepad")
+                device = Device::Gamepad;
+            key = parts[3];
+        }
+
+        bool operator==(const InputDeviceMapping& rhs) const {
+            return (this->device == rhs.device) && (this->framework == rhs.framework) && (this->number == rhs.number);
```
lioncash: These `this`'s aren't necessary. Neither are the parentheses.

src/core/settings.h
```diff
@@ -4,89 +4,150 @@
...
+        }
+        bool operator==(const std::string rhs) const {
```
lioncash: This string should be a reference.

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,165 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <cmath>
+#include <memory>
+#include <SDL.h>
+
+#include "SDL_gamecontroller.h"
```
lioncash: `#include <SDL_gamecontroller.h>`

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,165 @@
...
+        return devices;
+    }
+    SDL_GameControllerEventState(SDL_IGNORE);
+    for (int i = 0; i < 8; i++) {
+        SDL_GameController* gamecontroller;
+        if (SDL_IsGameController(i)) {
+            gamecontroller = SDL_GameControllerOpen(i);
+            if (gamecontroller != nullptr) {
+                std::shared_ptr<SDLGamepad> gamepad = std::make_shared<SDLGamepad>(i,gamecontroller);
+                devices.push_back(gamepad);
```
lioncash: There's no reason to put this on two lines. Just do:

```cpp
devices.push_back(std::make_shared<SDLGamepad>(i, gamecontroller));
```

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,165 @@
...
+        auto buttonName = SDL_GameControllerGetStringForButton(SDL_GameControllerButton(i));
+        for (const auto& mapping : gamepadinput_to_sdlname_mapping) {
+            if (mapping.second == buttonName) {
+                return Settings::InputDeviceMapping("SDL/" + std::to_string(this->number) + "/" + "Gamepad/" + std::to_string(static_cast<int>(mapping.first)));
+            }
+        }
+    }
+    for (int i = 0; i < SDL_GameControllerAxis::SDL_CONTROLLER_AXIS_MAX; i++) {
+        Sint16 value = SDL_GameControllerGetAxis(gamepad, SDL_GameControllerAxis(i));
+        if (abs(value) < 0.2 * 32767.0)
```
lioncash: You should be explaining what this means.

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,165 @@
...
+}
+
+Settings::InputDeviceMapping SDLGamepad::GetInput() {
+    if (gamepad == nullptr)
+        return Settings::InputDeviceMapping("");
+    SDL_GameControllerUpdate();
+    for (int i = 0; i < SDL_GameControllerButton::SDL_CONTROLLER_BUTTON_MAX; i++)
+    {
+        Uint8 pressed = SDL_GameControllerGetButton(gamepad, SDL_GameControllerButton(i));
+        if (pressed != 1)
```
lioncash: `pressed == 0`

src/input_core/devices/sdl_gamepad.h
```diff
@@ -0,0 +1,62 @@
...
+    static enum class GamepadInputs {
+        ButtonA,ButtonB,ButtonX,ButtonY,LeftShoulder,RightShoulder,
+        Start,Back,DPadUp,DpadDown,DpadLeft,DpadRight,L3,R3,LeftTrigger,RightTrigger,
+        LeftYPlus,LeftYMinus,LeftXPlus,LeftXMinus,RightYPlus,RightYMinus,
+        RightXPlus,RightXMinus, MAX
+    };
+private:
+    /// Maps the friendly name shown on GUI with the string name for getting the SDL button instance.
+    std::map<GamepadInputs, std::string> gamepadinput_to_sdlname_mapping = {
+        { GamepadInputs::ButtonA,"a" },
```
lioncash: spaces after commas

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+std::tuple<u16, u16, bool> GetTouchState() {
+    std::lock_guard<std::mutex> lock(touch_mutex);
+    return std::make_tuple(touch_x, touch_y, touch_pressed);
+}
+
+void SetTouchState(std::tuple<u16, u16, bool> value) {
+    std::lock_guard<std::mutex> lock(touch_mutex);
+    touch_x = std::get<0>(value);
```
lioncash:
```cpp
std::tie(touch_x, touch_y, touch_pressed) = value;
```

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/// Read settings to initialize devices
+void ParseSettings() {
+    auto uniqueMappings = GatherUniqueMappings();
+    GenerateUniqueDevices(uniqueMappings);
+}
+
+void ReloadSettings() {
+    if (devices.size() == 0)
```
lioncash:
```cpp
if (devices.empty())
```

src/core/settings.h
makotech222: Ill remove the this's. I like the parenthesis :)

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,171 @@
...
+        return;
+    devices.clear();
+    ParseSettings();
+}
+
+std::vector<std::shared_ptr<IDevice>> GetAllDevices() {
+    std::vector<std::shared_ptr<IDevice>> allDevices;
+    auto sdlGamePads = SDLGamepad::GetAllDevices();
```
lioncash:
```cpp
auto all_devices = SDLGamePad::GetAllDevices();
all_devices.push_back(InputCore::GetKeyboard());

return all_devices;
```

src/input_core/devices/device.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <map>
+
+#include "core/hle/service/hid/hid.h"
+#include "core/settings.h"
+#include "input_core/key_map.h"
+
+class IDevice {
+public:
+    virtual ~IDevice() = default;
```
lioncash: You should define this in a cpp file as well, otherwise the vtable gets put in every inheriting class's translation unit, instead of just one.

makotech222: Can you explain more what to do? I don't really understand.

lioncash:
note that our code also uses `snake_case` for variables, not `camelCase`, as is shown in the coding style guidelines.

src/input_core/devices/device.h
lioncash: The header would be like so:
```cpp
virtual ~IDevice();
```

inside a cpp file, you would be defining
```cpp
IDevice::~IDevice() = default;
```

src/core/settings.h
lioncash: realistically, you can just do:

```cpp
return std::tie(device, framework, number) == std::tie(rhs.device, ths.framework, rhs.number);
```

src/input_core/devices/keyboard.h
```diff
@@ -0,0 +1,49 @@
...
+#include <map>
+#include <memory>
+#include <mutex>
+
+#include "input_core/devices/device.h"
+
+struct KeyboardKey {
+    uint32_t key;
+    uint32_t scancode;
+    std::string character;
```
wwylele: Is it necessary for `KeyboardKey` has so many members? I see it is just used as a key for a map. Did I miss something?

src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -33,16 +35,21 @@ void EmuWindow_SDL2::OnMouseButton(u32 button, u8 state, s32 x, s32 y) {
 
     if (state == SDL_PRESSED) {
         TouchPressed((unsigned)std::max(x, 0), (unsigned)std::max(y, 0));
-    } else {
+    }
+    else {
```
wwylele: keep the old format please. Also same cases below.

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include <mutex>
+
+#include "core/core_timing.h"
+
+#include "input_core/input_core.h"
+#include "input_core/devices/keyboard.h"
+#include "input_core/devices/sdl_gamepad.h"
+
+namespace InputCore {
+constexpr u64 frame_ticks = 268123480ull / 60;
```
wwylele: This frame time has been REed and should be updated (sorry I forget the value). You may want to talk about it on IRC.

wwylele:
I don't have a game pad, so this is probably a dumb question:
How do one test a game pad? Could you please give some more instruction?

src/input_core/devices/keyboard.h
makotech222: originally, i needed the scan code along with the virtual code because I was trying to use the scancode to handle keyboard input, but that didn't work out well. I can probably remove the scancode variable now.

src/citra/emu_window/emu_window_sdl2.cpp
makotech222: I don't know why it keeps changing it, I never touched it :P I will fix.

bunnei: Probably Visual Studio's auto format. You can change this to match our style guide in Preferences.

src/input_core/input_core.cpp
bunnei: This should be fixed elsewhere as well... Do you have any more information on this?

Jhno591:
Pass me on privet build, let me try, I have both xbox 360 pad for PC, and PS3 pad for PC ^

legoj15:
Tested with an XBox 360 wireless controller, works without any noticeable flaw.
Layout I used:
image
Played around in Tri-Force Heros, worked as expected.

legoj15:
Did some more thorough testing, and found that if you have any input bound to a gamepad (for instance Button A bound to the controller and the rest bound to keyboard), then close Citra, unplug/ power off the controller, then boot Citra, power the controller back on and boot a game, Citra will crash unexpectedly (with no log errors), forcing it to close.

makotech222:
@legoj15 Thank you for testing. I will fix that one ASAP

neobrain:

2016-08-17T00:04:41Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:ad83a6ec15e4458984d8d9cfc451ccbc5c6313a1

legoj15:
@makotech222 The issue is fixed, keep up the good work. 👍

src/citra_qt/configure_input.cpp
```diff
@@ -59,49 +65,77 @@ void ConfigureInput::handleClick() {
     grabKeyboard();
     grabMouse();
     changing_button = sender;
-    timer->start(5000); //Cancel after 5 seconds
+    QElapsedTimer timer = QElapsedTimer();
```
lioncash: You should probably be using a `QTimer` here. A QTimer lets you listen for the stop event using its [timeout()](https://doc.qt.io/qt-5/qtimer.html#timeout) signal.

src/citra_qt/configure_input.cpp
```diff
@@ -59,49 +65,77 @@ void ConfigureInput::handleClick() {
...
+    }
     releaseKeyboard();
     releaseMouse();
     changing_button = nullptr;
     previous_mapping = nullptr;
 }
 
-QString ConfigureInput::getKeyName(int key_code) const {
+QString ConfigureInput::getKeyName(Settings::InputDeviceMapping mapping) const {
+    if (mapping.key == "")
```
lioncash:
```cpp
if (mapping.key.empty())
```

src/citra_qt/keybinding_names.h
```diff
@@ -0,0 +1,17 @@
...
+#include <array>
+#include <QCoreApplication>
+
+#include "input_core/devices/sdl_gamepad.h"
+
+///Map translatable, user-friendly names to input device buttons for use in the Qt input configuration GUI.
+class KeyBindingNames {
+    Q_DECLARE_TR_FUNCTIONS(KeyBindingNames)
+public:
+    static const std::array <QString, static_cast<int>(SDLGamepad::GamepadInputs::MAX)> sdl_gamepad_names;
```
lioncash: one space too many after the `array` portion.

src/core/settings.h
```diff
@@ -22,41 +23,101 @@ enum Values {
...
+        // directly mapped keys
+        "pad_a", "pad_b", "pad_x", "pad_y",
+        "pad_l", "pad_r", "pad_zl", "pad_zr",
+        "pad_start", "pad_select", "pad_home",
+        "pad_dup", "pad_ddown", "pad_dleft", "pad_dright",
+        "pad_cup", "pad_cdown", "pad_cleft", "pad_cright",
+
+        // indirectly mapped keys
+        "pad_circle_up", "pad_circle_down", "pad_circle_left", "pad_circle_right"
+    } };
```
lioncash: These braces should be four indents less, like the one below.

src/core/settings.h
```diff
@@ -22,41 +23,101 @@ enum Values {
...
+
+        if (parts[2] == "Keyboard")
+            device = Device::Keyboard;
+        else if (parts[2] == "Gamepad")
+            device = Device::Gamepad;
+        key = parts[3];
+    }
+
+    bool operator==(const InputDeviceMapping& rhs) const {
+        return (device == rhs.device) && (framework == rhs.framework) && (number == rhs.number);
```
lioncash:
```cpp
return std::tie(device, framework, number) == std::tie(rhs.device, rhs.framework, rhs.number);
```

You'll need to include `<tuple>` for this.

src/input_core/devices/device.cpp
```diff
@@ -0,0 +1,7 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "input_core/devices/device.h"
+
+IDevice::~IDevice() = default;
```
lioncash: Missing a newline after this line according to the diff.

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,164 @@
...
+    for (int i = 0; i < SDL_GameControllerButton::SDL_CONTROLLER_BUTTON_MAX; i++)
+    {
+        Uint8 pressed = SDL_GameControllerGetButton(gamepad, SDL_GameControllerButton(i));
+        if (pressed == 0)
+            continue;
+
+        auto buttonName = SDL_GameControllerGetStringForButton(SDL_GameControllerButton(i));
+        for (const auto& mapping : gamepadinput_to_sdlname_mapping) {
+            if (mapping.second == buttonName) {
+                return Settings::InputDeviceMapping("SDL/" + std::to_string(this->number) + "/" + "Gamepad/" + std::to_string(static_cast<int>(mapping.first)));
```
lioncash: `this` isn't necessary.

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+    }
+    for (int i = 0; i < SDL_GameControllerAxis::SDL_CONTROLLER_AXIS_MAX; i++) {
+        Sint16 value = SDL_GameControllerGetAxis(gamepad, SDL_GameControllerAxis(i));
+        // dont process if in deadzone. Replace later with settings for deadzone.
+        if (abs(value) < 0.2 * 32767.0)
```
lioncash: Not really sure if this is relevant or if you already do this, but you may want to consider handling deadzone as a radial field rather than an axial one (examples: [1](https://twitter.com/JakubKoziol/status/671092591409160193), [2](http://www.gamasutra.com/blogs/JoshSutphin/20130416/190541/Doing_Thumbstick_Dead_Zones_Right.php)) if you don't.

neobrain:

2016-08-17T01:38:01Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:62ec2623a919a59da2eaac022ff20cca03700de2

src/citra_qt/configure_input.cpp
makotech222: I don't want to listen to the event. I want to handle it within the loop checking for inputs.

src/input_core/devices/sdl_gamepad.cpp
makotech222: well i really only know the value of one of the axis at a time. I will have to refactor a bit if i wanted to accomplish this. Save for another pull request? Ill leave a comment

neobrain:

2016-08-17T01:46:10Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:07fd1a9ec983541220f1b0488b742f1467e185cf

src/citra_qt/configure_input.cpp
lioncash: > I don't want to listen to the event

Considering you're listening on a timeout value for inputs, you do. It's what timeout() is for, literally speaking.

The workflow for this sort of thing should be something like the following:

1. User clicks the button, start the timer off (your timer is a class member in this case).
2. override [keyPressEvent](https://doc.qt.io/qt-5/qwidget.html#keyPressEvent) and handle what should happen if the user manages to press a key before the timeout (you already do this).
3. If the timeout occurred, do nothing, or restore the previously bound key to that button.

Qt already provides you with an event loop mechanism, you don't need to make one yourself. Furthermore, the fact that you're calling something like `setKey` in a function called `handleClick`, two things that are completely separate from one another, indicates there is something wrong with the design as it currently is.

src/citra_qt/configure_input.cpp
```diff
@@ -59,91 +65,115 @@ void ConfigureInput::handleClick() {
...
-    if (key_code == Qt::Key_Control)
-        return tr("Ctrl");
-    if (key_code == Qt::Key_Alt)
-        return tr("Alt");
-    if (key_code == Qt::Key_Meta)
+QString ConfigureInput::getKeyName(Settings::InputDeviceMapping mapping) const {
+    if (mapping.key == "")
         return "";
-    if (key_code == -1)
+    try {
```
lioncash: What exception or problem are you resolving with a try-catch here?

makotech222: Its mostly an issue with transitioning config files with older settings. If the InputMapping key can't be read properly, it will catch the exception and return "".

src/citra_qt/configure_input.cpp
makotech222: But sdl gamepads don't work on an event loop basis, their state is polled, thus, the while loop.

lioncash: You really shouldn't be pushing backend specifics out onto the generic codepath, especially within GUI code. If a different toolkit/frontend is ever moved to (or supported) at some point, things will just randomly break. These are behaviors that should wholly be handled by the input_core API.

Even Dolphin does not require this ([source](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/DolphinWX/InputConfigDiag.cpp#L510)), and we deal with SDL as well.

That problem is likely an underlying issue to do with something else, or a flaw in the input_core API.

src/citra_qt/configure_input.cpp
lioncash: Is there really no way to sanitize for that instead of using exceptions as control flow (instead of handling exceptional conditions)?

src/citra_qt/configure_input.cpp
makotech222: Interesting. They also have the benefit of listening to only one device for input at a time. But i see your point. I will try to move that behavior into input core.

src/citra_qt/configure_input.cpp
makotech222: Yeah i suppose I could try and sanitize it. I will try and fix today.

Jhno591:
@legoj15 , I Mearged it and tested, but no responce calibrateing buttons from my xbox360 pad, anyone know what files I'm maybe missing out of 39 of them ?

neobrain:

2016-08-21T18:24:56Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:9d1b253fed5557ba319dc45f610b05718de24bca

neobrain:

2016-08-21T18:26:07Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:e1283f4a0a2f75e06435feb30d88bd6c4cde234d

ivorynguyen:
Okay @lioncash , let me know how you feel about the new changes. They should address your concerns i hope :P

makotech222:
The annoying part about having a gf who is a software dev is that we sometimes are logged into each others github accounts haha.

lioncash:
Review status: 0 of 39 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/citra_qt/configure_input.cpp, line 122 \[r9\]](https://reviewable.io:443/reviews/citra-emu/citra/1995#-KPiTxHXyDBj7MORFEJn:-KPiTxHYJa8jJ2C-CkaU:bl7yid6) ([raw file](https://github.com/citra-emu/citra/blob/e1283f4a0a2f75e06435feb30d88bd6c4cde234d/src/citra_qt/configure_input.cpp#L122)):*
> ```cpp
>     int* key_code_temp = new int();
>     int key_code;
>     if (!Common::TryParse(mapping.key, key_code_temp))
> ```

You can just do

```cpp
if (!Common::TryParse(mapping.key, &key_code);
```

instead of allocating a pointer.

---

*[src/citra_qt/configure_input.cpp, line 147 \[r9\]](https://reviewable.io:443/reviews/citra-emu/citra/1995#-KPiV1CGiTr0BG3qs3wL:-KPiV1CHw7WodZ_0V19S:b-4ljqi1) ([raw file](https://github.com/citra-emu/citra/blob/e1283f4a0a2f75e06435feb30d88bd6c4cde234d/src/citra_qt/configure_input.cpp#L147)):*
> ```cpp
> }
> 
> void ConfigureInput::removeDuplicates(const Settings::InputDeviceMapping newValue) {
> ```

Why are you using a map, rather than a set, if duplicates are not desirable, out of curiosity?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/1995)*

makotech222:
@lioncash I don't know how to respond directly using Reviewable, so i will here:

Regarding map, i'm using it because i'm a newbie at C++ and i don't really know the best structure :D In C# I would use a dictionary, which is what i thought the map basically is. What do you suggest I change it to?

neobrain:

2016-08-21T19:36:53Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:5ec35a9fcc1246b1861c9896ebe75850690e0ea0

mailwl:
works like a charm (besides circle pad pro, but this is not related with this pr). want to be merged to master

Jhno591:
Agreed Mailwl time once last moment cleanups are done!, time to get it fully merged in to the main build.

Jhno591:
EDIT: How come only works with an xbox 360 pad..., try a ps3 usb, No calibrate with pressing buttons.

Jhno591:
Tested this pad no go - http://liangpeter79gamepad.en.ecplaza.net/main.jpg but work fine on others games,  although wireless Usb Reciver Xbox360 Pad works grate no issues at all, as I know off.

makotech222:
Unfortunately, I don't have any other controllers to test with. This method is using SDL2, so whatever SDL2 supports, should work fine.

neobrain:

2016-08-27T20:37:04Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:4b3be677d0e97071cd9bd2d1629ae5a72dcc038c

makotech222:
@Jhno591 Can you update and give another test with the PS3 controller? I pushed some changes that might fix it, but I can't test unfortunately.

src/input_core/devices/gamecontrollerdb.h
```diff
@@ -0,0 +1,167 @@
+// Copyright 2014 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+// Community sourced database of controllers, from https://raw.githubusercontent.com/gabomdq/SDL_GameControllerDB/master/gamecontrollerdb.txt
```
JayFoxRox: I somehow feel this should contain a revision info. Better yet: This should probably be some submodule and a script to convert the data?

src/input_core/devices/gamecontrollerdb.h
```diff
@@ -0,0 +1,167 @@
+// Copyright 2014 Citra Emulator Project
```
JayFoxRox: Last time I've checked it's been 2016?

src/input_core/devices/gamecontrollerdb.h
makotech222: I have no idea how to do submodule... I can add a date to it though for version info.

src/input_core/devices/gamecontrollerdb.h
makotech222: oops copy paste

src/input_core/devices/gamecontrollerdb.h
Jhno591: @makotech222 iit's not a genuine ps2/ps3 controller just a cheap job, Although all add in updated fixes and test it, But my xbox360 pad is genuine, here my Unlisted video though with pad - https://www.youtube.com/watch?v=LdcTZhn7lbY , Be easy er if some could put all the needed folders starting from src and added them in a .zip file , upload where you like so i can test, but I can do it manually so no worrys, will keep you posted on result . thanks.

neobrain:

2016-08-27T20:58:10Z: https://github.com/citra-emu/citra/compare/e91327c86a863b9419fd3695b2400a52336ec3b5...makotech222:16d1fe6e4ac5cd529ecbf26f90f6cfed6ad07459

src/input_core/devices/gamecontrollerdb.h
JayFoxRox: @Jhno591 this is not a forum (and your post is entirely unrelated to this diff discussion).

Jhno591: JayFoxRox , edited post.

Jhno591:
Still not working with the USB unofficial ps2/ps3 pad, I'll re check all files again later and re-compile them, if some one has official ps2/ps3 controller it's best to try them!, as for xbox360 pad yes well is confirmed to work full as it dose for me.

Sinxeado:
Hi guys,

I'm not a programmer, just a student trying to lear both english and pc's stuffs.

I want to help in this proyect and thought i can here. I download a software where i can replace my controller (original ps3 controller) to any controller 
[http://es.tinypic.com/view.php?pic=30cw2s8&s=9#.V8PbZq3puUk](url)

The Problem: As I said, I don't know how to compile the branch, so i need to someone give to me all I needed to testing.

Sorry for my english and programming :cry:

neobrain:

2016-09-02T23:00:56Z: https://github.com/citra-emu/citra/compare/274b925a9705ebd5c4ec242c70d44249dc9f2b0d...makotech222:ac4cdc00ad315775d82c845d1057ecd5dd787854

neobrain:

2016-09-02T23:04:58Z: https://github.com/citra-emu/citra/compare/274b925a9705ebd5c4ec242c70d44249dc9f2b0d...makotech222:10a2df5c1752f7604fd70123bd79a418dc0322f4

neobrain:

2016-09-02T23:06:26Z: https://github.com/citra-emu/citra/compare/274b925a9705ebd5c4ec242c70d44249dc9f2b0d...makotech222:c3bbbd540f797d1b34474eaf2072f4490cf3618e

Romsstar:
Tested with official PS3 and PS4 controllers.
Both need either DS3 or DS4 to XInput Mapper in order to work in Citra or they won't be recognized.
The issue with that is that the second analog stick of the DS3/DS4 is basically not recognized since Citra then interprets the controller as a 360 controller.

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,192 @@
...
+static int tick_event;
+static Service::HID::PadState pad_state;
+static std::tuple<s16, s16> circle_pad = { 0,0 };
```
jroweboy: I think this should just be 

    static std::tuple<s16, s16> circle_pad { 0,0 };

according to this http://stackoverflow.com/a/3414239/745719 (The current way breaks mingw64 builds)

wwylele:

Subv:
> Both need either DS3 or DS4 to XInput Mapper in order to work in Citra or they won't be recognized.

This is something i consider very important, you shouldn't *need* to use the XInput wrapper, new controller backends should be easy to add. I haven't reviewed the code yet, but is this the case?

ivorynguyen:
Yeah should be pretty easy to add. It needs some GUI changes though, for setting up binding to specify if you want Xinput or SDL

makotech222:
Yeah should be pretty easy to add. It needs some GUI changes though, for setting up binding to specify if you want Xinput or SDL

JayFoxRox:
I've rebased this (without the line ending fixes) to a recent bleeding-edge for the uni study room PC yesterday. My experience was rather unpleasant at first, but once it was configured it worked more or less fine ;)

(Some) problems with this PR:

- On Linux (debian testing I believe) with Xbox 360 Controller it did not correctly map buttons. It kept claiming that the "left trigger" was held, meaning it was impossible to configure any other button, except for using another axis (which was apparently checked before the trigger).
- Connecting one gamepad only and some printf magic later I figured the left trigger always returned 16384 regardless of current state (works fine in dolphin or other games).
- If it were stuck, it would be impossible to see which gamepad had the left trigger stuck when multiple gamepads were connected.
- Deadzone not in common place or configurable. Fixing the deadzone to 60% made it possible to at least configure the controller.
- The 360 "Back" button did not work (Xbox button did work) / was not detected as press.
- Left trigger was detected as "Back" button, Right trigger was detected as "Start" button (iirc)
- Configuring LZ or RZ resulted in SIGSEGV
- With 4 controllers connected: Restarting Citra suddenly detected another controller as the configured one (also not an issue in dolphin)
- Not clang-formatted yet
- Commit naming is confusing
- Does not compile out of the box as tests/ is missing input_core
- Some bad design choices / code but I'd have to review code again to comment on that

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,188 @@
...
+                    keys_pressed[entry.first] = false;
+                }
+            }
+        } else {
+            // Try axis if button isn't valid
+            SDL_GameControllerAxis axis = SDL_GameControllerGetAxisFromString(
+                gamepadinput_to_sdlname_mapping[static_cast<GamepadInputs>(stoi(entry.first))]
+                    .c_str());
+            if (axis != SDL_GameControllerAxis::SDL_CONTROLLER_AXIS_INVALID) {
+                Sint16 value = SDL_GameControllerGetAxis(gamepad, axis);
```
JayFoxRox: Why is this function called from different places in the codebase?
Shouldn't the platform abstraction be minimal and the values cached in some kind of state?

The axis read and deadzone should be independent of each other, I think:
- Read axis to common state
- When mapping to 3DS functionality: do deadzone check (could be different per 3DS button / axis)

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,188 @@
...
+        // deadzone)
+        // dont process if in deadzone. Replace later with settings for deadzone.
+        if (abs(value) < 0.2 * 32767.0)
+            continue;
+        std::string modifier;
+        if (value > 0)
+            modifier = "+";
+        else
+            modifier = "-";
+        std::string axisName = SDL_GameControllerGetStringForAxis(SDL_GameControllerAxis(i));
```
JayFoxRox: might want to cache the return value from `SDL_GameControllerAxis(i)`

src/input_core/devices/gamecontrollerdb.h
```diff
@@ -0,0 +1,169 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+// Community sourced database of controllers, from
+// https://raw.githubusercontent.com/gabomdq/SDL_GameControllerDB/master/gamecontrollerdb.txt. Taken
```
JayFoxRox: I still think this should be a submodule (generating a cpp file through CMake [configure_file?]) or included as part of the citra config (read at runtime - as recommended in the README.md at https://github.com/gabomdq/SDL_GameControllerDB ).
But in general it feels weird to have ties to input hardware in the codebase.
I'd expect a project like SDL to maintain and distribute this as part of the platform abstraction to be honest. Did you verify that there is no other way of distributing this file?

Did anyone ever ask @gabomdq about upstreaming this?

makotech222: Quick google search led me to it, and i didn't see any other official db. I don't know how to do anything with submodule, so i would appreciate someone making a PR to my fork with the change.

src/common/logging/log.h
```diff
@@ -85,7 +85,9 @@ enum class Class : ClassType {
     Audio_DSP,         ///< The HLE implementation of the DSP
     Audio_Sink,        ///< Emulator audio output backend
     Loader,            ///< ROM loader
-    Count              ///< Total number of logging classes
+    Input,             ///< Input backend
+
+    Count ///< Total number of logging classes
```
Subv: comment alignment

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <mutex>
+
+#include "core/core_timing.h"
+
+#include "input_core/devices/keyboard.h"
+#include "input_core/devices/sdl_gamepad.h"
+#include "input_core/input_core.h"
+
+namespace InputCore {
+constexpr u64 frame_ticks = 268123480ull / 60;
```
JayFoxRox: This comes from the GPU I believe? If so, document where it came from and possibly move it to a common space (not sure where exactly though) and reuse it

JayFoxRox:
You might have seen my tag for you on https://github.com/JayFoxRox/citra/pull/40 which proposes changes / gives you an idea how to add the submodule. However, as distribution for that file is still somewhat tricky: It should be in a system folder, but SDL doesn't provide such a folder yet, Citra doesn't have a system path yet either, and it really shouldn't be Citra related - it's an SDL file.

As review was also kind of slow on this one in the past, I suggest doing this differently:
I'd advise you to remove the DB and its loader code from this PR (unless absolutely necessary).
SDL will still provide its own internal DB for common controllers.

We can still add the DB and loading for it (in a followup PR) once the SDL people know how to deploy and use it. Their API seems kind of broken and unstable at this point. 

Instead try to split this code into a few smaller and nicer commits so we can review it easier.
Focus on a clean design - we can extend and improve it later.

makotech222:
Just an update. I'm working on refactoring some of the code. I will be making the devices much simpler, and reducing them to simply reporting which keys/buttons were pressed. Inputcore is going to handle determining which emulator buttons to press based on the device returns. Also cleaning up some other things and generally trying to make things simpler. If anyone has any specific things they would like to see, please let me know.

makotech222:
I'm about 80% finished with refactor. It is all implemented, i just need to finish two things:

- Cleaning up old code thats no longer needed
- Re-implement MerryMage's recent PR that fixed up input GUI on Qt. A lot of her changes don't work with new input core stuff.

Let me know how you guys feel about the refactor. I simplified all the devices, so they only care about reporting which of their buttons are pressed. Input core now handles how the emulator reacts to the physical button presses. Deadzone is now handled properly, and circle mod button works for keyboard and gamepad as an added result. There is a new setting (no gui yet) for circle pad dead zone.

JayFoxRox:
I appreciate the effort in supporting gamepads and joysticks, but I'll have to close this now. This is far from being complete, probably more like RFC or pre-RFC.
You can open a new PR if it's ready (and has been tested properly).

---

I've tested this again yesterday and had various issues still:

- SIGSEGV when pressing keyboard keys while gamepad was mapped
- SIGSEGV when changing config ingame
- Could not use circlepad left and downwards direction reliably (rendering almost anything unplayable)
- Xbox 360 gamepads were still mapped very badly (Trigger buttons shown as "Back" and "Start")
- There are still different deadzone variables / settings now which is really confusing

I could not get it to work without a lot of manual patching / private hacks 3 weeks ago and it's still in a rough shape now. If anyone accidentally hits merge on this one this would break Citra completely. That's not acceptable for a "Pull Request" (which means: "Pull my changes please!" and not "let me show you what I'm working on for 4 months to get more feedback what I could implement"). RFC PRs can be useful, but they shouldn't be opened for long and clearly marked as such.

Aside from that, this has had various conflicts and compilation issues in the past, as such it should not qualify for Bleeding-Edge anymore either.

---

If you want to continue to work on this please leave a message, however, expect that others might also want to implement this feature.
4 months is a long time and if this would have been split into smaller commits / PRs we probably would have had a working implementation by now (as code could have been improved in review more quickly + other people could have helped working on backends etc).

Closed.

makotech222:
Understood. I will continue working on my branch to get this working well. I don't know anyone else but you who is getting the SIGSEGV errors when working with the input gui, so perhaps its a linux issue which i can't really test.

makotech222:
@JayFoxRox Your issues could  be caused by driver issues on linux.

Did some research about your issue and came to these pages:
http://steamcommunity.com/app/221410/discussions/0/864960354141733830/
https://forums.libsdl.org/viewtopic.php?t=6352&sid=35bdc1b1615f9ea081671ff548a7e360

Let me know if you have any insight on it, regarding if you have latest drivers or what not. Never used linux before so I don't know much about it.

As for Dolphin, it looks like they are using the SDL_Joystick instead of SDL_Gamecontroller. I'm not sure why, and if there would be a difference on Linux. I will do more research on that end.

bunnei:
@makotech222 You'll probably make more progress on this by creating a Linux VM and debugging the issue yourself.

PR #2004, opened 2016-08-13
Dyncom: Correct implementation of STM for R15

MerryMage:
Related test: https://github.com/MerryMage/3ds-tests/tree/master/stm

`stmfd r13!, {pc}` currently stores the incorrect value to the stack (seems to always be `8`).

Correction for the ~~System~~ STM (User registers) version of the instruction (line 3231) based on manual.

### DIFF
--- a/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
@@ -3228,7 +3228,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
                     addr += 4;
                 }
                 if (BIT(inst_cream->inst, 15)) {
-                    cpu->WriteMemory32(addr, cpu->Reg_usr[1] + 8);
+                    cpu->WriteMemory32(addr, cpu->Reg[15] + 8);
                 }
             } else {
                 for (int i = 0; i < 15; i++) {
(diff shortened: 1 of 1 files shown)


neobrain:
> based on manual

Which one, though? I vaguely recall stuff like this being implementation-defined, so the description is too vague.

MerryMage:
Both ARM ARM (ARM DDI 0100) and ARMv7-A,R (ARM DDI 0406C.b) agree. The former specifies the usual implementation-defined PC+8 vs PC+12 on page A2-9, and the latter specifies PCStoreValue() which is equivalent.

The value is certainly PC+8 because, quoting the former manual, "An implementation must use the same offset for all ARM STR and STM instructions that store R15. It cannot use 8 for some of them and 12 for others."

yuriks:
LGTM, checks out with the manual

Road-Drum:
i got the following error right after pulling the latest master

`[ 83%] Linking CXX executable citra
../video_core/libvideo_core.a(debug_utils.cpp.o): In function `Pica::DebugUtils::DumpTexture(Pica::Regs::TextureConfig const&, unsigned char*)':
/home/hello/citra/src/video_core/debug_utils/debug_utils.cpp:646: undefined reference to `png_set_longjmp_fn'
collect2: error: ld returned 1 exit status
src/citra/CMakeFiles/citra.dir/build.make:158: recipe for target 'src/citra/citra' failed`

wwylele:
@Road-Drum apparently no. This kind of problem should likely go to IRC or the forum.

Issue #1495, opened 2016-03-11
pause emulation when prompted to close the emulator 


RavenHome1:
currently when the user closes the emulator the prompt statement to close citra will appear but the game in the background will continue to run and the user have not control over the game. it would be ideal if citra auto pauses the emulation once the prompt appears and auto resumes emulation if user selected no. also please add an option to disable and enable the prompt since some users may find it to be irritating.

LittleWhite-tb:
Configuration for this popup is implemented through #1368 
I will look for the pause thing ;)

RavenHome1:
alright awesome

LittleWhite-tb:
Quick question : when user click on a menu (whatever the menu is), do you think it is better to pause emulation, or to let it continue ?

LittleWhite-tb:
Additional Note : you can disable the "close popup" using configuration file. It's the item "confirmClose".

RavenHome1:
i think it should pause regardless of menu, that is how other emus do it anyway

RavenHome1:
oh i just tested Dsmume it pauses every time a menu is open so yeah pretty much it would be an ideal to pause on every menu

PR #2006, opened 2016-08-15
Set pause when focus is lost

LittleWhite-tb:
The feature is disabled by default and can be enabled through the configure widget.
Feature was asked by RavenHome1 in #1495.

pause_on_focus_lost

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -186,6 +186,8 @@ GMainWindow::GMainWindow() : config(new Config()), emu_thread(nullptr)
     connect(this, SIGNAL(EmulationStarting(EmuThread*)), graphicsTracingWidget, SLOT(OnEmulationStarting(EmuThread*)));
     connect(this, SIGNAL(EmulationStopping()), graphicsTracingWidget, SLOT(OnEmulationStopping()));
 
+    connect(render_window, SIGNAL(gotFocus()), this, SLOT(OnStartGame()));
+    connect(render_window, SIGNAL(lostFocus()), this, SLOT(OnPauseGame()));
 
     // Setup hotkeys
     RegisterHotkey("Main Window", "Load File", QKeySequence::Open);
@@ -468,6 +470,8 @@ void GMainWindow::OnStartGame() {
 
     ui.action_Pause->setEnabled(true);
     ui.action_Stop->setEnabled(true);
+
+    render_window->setFocus();
 }
 
 void GMainWindow::OnPauseGame() {
@@ -487,7 +491,7 @@ void GMainWindow::ToggleWindowMode() {
         // Render in the main window...
         render_window->BackupGeometry();
         ui.horizontalLayout->addWidget(render_window);
-        render_window->setFocusPolicy(Qt::ClickFocus);
+        render_window->setFocusPolicy(Qt::StrongFocus);
         if (emulation_running) {
             render_window->setVisible(true);
             render_window->setFocus();
@@ -505,6 +509,7 @@ void GMainWindow::ToggleWindowMode() {
             game_list->show();
         }
     }
+    render_window->setFocus();
 }
 
 void GMainWindow::OnConfigure() {
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -489,20 +489,17 @@ void GMainWindow::OnStopGame() {
     ShutdownGame();
 }
 
-void GMainWindow::OnFocusChanged(bool hasFocus) {
+void GMainWindow::OnFocusChanged(bool has_focus) {
     // The focus change does not impact actual emulator state if:
     // - the option is disabled
     // - the emulation is not started
     // - the emulation has been paused through menu
-    if(!UISettings::values.pause_onfocuslost ||
-       !emu_thread ||
-       emulation_paused)
+    if (!UISettings::values.pause_onfocuslost ||
+        !emu_thread ||
+        emulation_paused)
         return;
 
-    if (hasFocus)
-        emu_thread->SetRunning(true);
-    else // focus lost
-        emu_thread->SetRunning(false);
+    emu_thread->SetRunning(has_focus);
 }
 
 void GMainWindow::ToggleWindowMode() {
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -186,8 +186,7 @@ GMainWindow::GMainWindow() : config(new Config()), emu_thread(nullptr)
     connect(this, SIGNAL(EmulationStarting(EmuThread*)), graphicsTracingWidget, SLOT(OnEmulationStarting(EmuThread*)));
     connect(this, SIGNAL(EmulationStopping()), graphicsTracingWidget, SLOT(OnEmulationStopping()));
 
-    connect(render_window, SIGNAL(gotFocus()), this, SLOT(OnStartGame()));
-    connect(render_window, SIGNAL(lostFocus()), this, SLOT(OnPauseGame()));
+    connect(render_window, SIGNAL(focusChanged(bool)), this, SLOT(OnFocusChanged(bool)));
 
     // Setup hotkeys
     RegisterHotkey("Main Window", "Load File", QKeySequence::Open);
@@ -472,6 +471,8 @@ void GMainWindow::OnStartGame() {
     ui.action_Stop->setEnabled(true);
 
     render_window->setFocus();
+
+    emulation_paused = false;
 }
 
 void GMainWindow::OnPauseGame() {
@@ -480,10 +481,28 @@ void GMainWindow::OnPauseGame() {
     ui.action_Start->setEnabled(true);
     ui.action_Pause->setEnabled(false);
     ui.action_Stop->setEnabled(true);
+
+    emulation_paused = true;
 }
 
 void GMainWindow::OnStopGame() {
     ShutdownGame();
+}
+
+void GMainWindow::OnFocusChanged(bool hasFocus) {
+    // The focus change does not impact actual emulator state if:
+    // - the option is disabled
+    // - the emulation is not started
+    // - the emulation has been paused through menu
+    if(!UISettings::values.pause_onfocuslost ||
+       !emu_thread ||
+       emulation_paused)
+        return;
+
+    if (hasFocus)
+        emu_thread->SetRunning(true);
+    else // focus lost
+        emu_thread->SetRunning(false);
 }
 
 void GMainWindow::ToggleWindowMode() {
@@ -521,6 +540,10 @@ void GMainWindow::OnConfigure() {
         render_window->ReloadSetKeymaps();
         config->Save();
     }
+
+    // Make sure the emulation is running when all pauses are disabled
+    if (emu_thread && !UISettings::values.pause_onfocuslost && !emulation_paused)
+        emu_thread->SetRunning(true);
 }
 
 void GMainWindow::OnCreateGraphicsSurfaceViewer() {
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -150,6 +150,10 @@ int main(int argc, char **argv) {
 
     while (emu_window->IsOpen()) {
         Core::RunLoop();
+
+        // Pause on focus lost
+        if (!emu_window->HasFocus())
+            emu_window->WaitForFocus();
     }
 
     return 0;
--- a/src/citra/emu_window/emu_window_sdl2.cpp
+++ b/src/citra/emu_window/emu_window_sdl2.cpp
@@ -63,16 +63,16 @@ void EmuWindow_SDL2::OnResize() {
 }
 
 void EmuWindow_SDL2::WaitForFocus() {
-    while(!has_focus) {
+    while (!has_focus) {
         SDL_Event event;
         SDL_WaitEvent(&event);
-        if(event.type == SDL_WINDOWEVENT) {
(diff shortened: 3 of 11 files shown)


neobrain:

2016-08-15T07:23:31Z: https://github.com/citra-emu/citra/compare/c3c6ac38e0a02b6400a73e218b0f3d90e9c08af8...LittleWhite-tb:e64114f6a6a6fec5838391acfd7a17021d92586d

wwylele:
- Screenshots if UI changes
 - Equivalent feature in SDL frontend (or at least a setting item)
 - When I open the menu, the Pause button is greyed out and I can't actually pause the game as before. You should make it work well with the existing Pause and Continue buttons.
 - (A question before thinking carefully) Will it be cleaner if using the `focusObjectChanged` signal?

LittleWhite-tb:
pause_on_focus_lost

The pause menu behaviour is consistent. If the new option is enabled, just by clicking on the menu, the pause is set (so the pause menu is greyed out). When the new option is disabled, the behaviour is as before.

wwylele:
> If the new option is enabled, just by clicking on the menu, the pause is set (so the pause menu is greyed out)

This is misleading and is not friendly to users IMO. It makes it impossible to pause the game while the window _is_ focused. This can be problematic if, like, I want to bring the window to top and take a screenshot.

Bonus: if in the future the Puase in menu gains its hotkey, one can pause the game without open the menu(i.e. without making the window lose focus), and then he can unintentionally continue the game by focus out and focus in again.

RavenHome1:
:)

LittleWhite-tb:
I think it is possible to apply it to the SDL frontend (SDL_WINDOW_FOCUS_GAINED and SDL_WINDOW_FOCUS_LOST). How the user will enable/disable it ? My guess is in the INI file, but I don't want to put a new value in the struct Settings::Value since this option is only for the frontend, and not for the core.
Is it ok to create a UISettings::Value struct, similar to the one in Qt frontend ?

LittleWhite-tb:
I have improved the pause implementation : when user pauses through the menu, it can't be unset unless he is unpausing from the menu.
Also added the feature in SDL2, but there is no way to disable it. I am waiting to know how the user should configure it (new option from command line ? new entry in INI file ?).

ml-:
Just leaving my opinion on this:

**Pause on focus lost**
Nice feature that should be implemented. Just imagine losing focus when some annoying window pops up and you lose control of the game.
Should be disabled by default. (which it is)

**Resume on focus gained**
This should be optional and disabled by default.
You'd expect a fullscreen game to resume when you switch back to it but not a windows emulated game.
It's probably better (also for UX) to be required to resume the game manually.

_my two cents_

LittleWhite-tb:
Your opinion is welcomed and much appreciated.
Dolphin (ok, we don't have to follow everything made in this project) unpause when the window is focused back. To me, it looks more natural, for instance if I have IRC and an emulator running, I will do lot of Alt+TAB to talk on IRC and play. I am expecting the emulator unpause the game when I am back at it (even if playing in windowed mode).

ml-:
What do you think about adding the **sub-option "Resume on focus"** (enabled by default)
This should of course only work if "Pause on focus lost" is enabled.

Would be something like:
```
if (hasFocus &&
    UISettings::values.resume_onfocus)
      emu_thread->SetRunning(true);
```

wwylele:
@ml- my feeling is that the added option in this PR is already sort of redundant (but still ok I guess). Adding another option... tbh I don't like it.

neobrain:
I think there are a few people in this thread that should read some human interface guidelines before providing further input.

lioncash:
Review status: 0 of 11 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/citra/emu_window/emu_window_sdl2.cpp, line 66 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSb_Obw6nI8mATxbmE:-KPSb_Obw6nI8mATxbmF:b-r5gmz8) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra/emu_window/emu_window_sdl2.cpp#L66)):*
> ```cpp
> 
> void EmuWindow_SDL2::WaitForFocus() {
>     while(!has_focus) {
> ```

space after `while`

---

*[src/citra/emu_window/emu_window_sdl2.cpp, line 69 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSbY4jG7bEDCS1TqHC:-KPSbY4kYEc4uqHz-Ptn:b-2ja7h6) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra/emu_window/emu_window_sdl2.cpp#L69)):*
> ```cpp
>         SDL_Event event;
>         SDL_WaitEvent(&event);
>         if(event.type == SDL_WINDOWEVENT) {
> ```

space after `if`

---

*[src/citra/emu_window/emu_window_sdl2.cpp, line 72 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSbcwXEYCQsH6Co0ho:-KPSbcwZhWli1v_NFnpE:b88uowk) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra/emu_window/emu_window_sdl2.cpp#L72)):*
> ```cpp
>             switch (event.window.event) {
>             case SDL_WINDOWEVENT_FOCUS_GAINED:
>                 has_focus=true;
> ```

space before and after `=`

---

*[src/citra/emu_window/emu_window_sdl2.cpp, line 75 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSbfaBu6jg5YV1AYRL:-KPSbfaCsz7pps512TSX:b-7ropln) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra/emu_window/emu_window_sdl2.cpp#L75)):*
> ```cpp
>                 break;
>             case SDL_WINDOWEVENT_CLOSE:
>                 has_focus=true;
> ```

ditto

---

*[src/citra/emu_window/emu_window_sdl2.h, line 61 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSbiX4_V0lLtZD0cj6:-KPSbiX4_V0lLtZD0cj7:b-p9cnqe) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra/emu_window/emu_window_sdl2.h#L61)):*
> ```objc
>     bool is_open = true;
> 
>     /// Is the window has focus
> ```

`/// Does the window have focus`

---

*[src/citra_qt/bootmanager.h, line 139 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSdrkiDk-BW2BxcThg:-KPSdrkjBlkbYdAMQlUk:b-35topg) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra_qt/bootmanager.h#L139)):*
> ```objc
>     /// Emitted when the window is closed
>     void Closed();
>     /// Emitted when the window focus changed
> ```

`/// Emitted when the window focus changes`

---

*[src/citra_qt/bootmanager.h, line 140 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPScNgOqDhe1jOZE0jg:-KPScNgOqDhe1jOZE0jh:b-1ix2pk) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra_qt/bootmanager.h#L140)):*
> ```objc
>     void Closed();
>     /// Emitted when the window focus changed
>     void focusChanged(bool hasFocus);
> ```

`has_focus`

---

*[src/citra_qt/bootmanager.h, line 147 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSbo8KfIKUh3MayyRh:-KPSbo8KfIKUh3MayyRi:b4epqvh) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra_qt/bootmanager.h#L147)):*
> ```objc
>     /**
>      * Capture focus gain to unset pause
>      * @param event
> ```

`@param` here will cause warnings when running Doxygen, because the name, `event`,  isn't present in the function signature. 

---

*[src/citra_qt/bootmanager.h, line 153 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSbzUMiAHIq36lRr1u:-KPSbzUNwkXZ_HsGCnQ9:brhumzt) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra_qt/bootmanager.h#L153)):*
> ```objc
>     /**
>      * Capture focus lost to set pause
>      * @param event
> ```

Same here.

---

*[src/citra_qt/config.cpp, line 119 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSc0UKiuujB7HJZbwZ:-KPSc0UKiuujB7HJZbw_:b-wda1yt) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra_qt/config.cpp#L119)):*
> ```cpp
>     UISettings::values.single_window_mode = qt_config->value("singleWindowMode", true).toBool();
>     UISettings::values.display_titlebar = qt_config->value("displayTitleBars", true).toBool();
>     UISettings::values.pause_onfocuslost = qt_config->value("pauseOnFocusLost",false).toBool();
> ```

Space after the comma.

---

*[src/citra_qt/main.cpp, line 492 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPScQtsZerud6_k7lyz:-KPScQttPqkUGSPck8Ih:b-1ix2pk) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra_qt/main.cpp#L492)):*
> ```cpp
> }
> 
> void GMainWindow::OnFocusChanged(bool hasFocus) {
> ```

`has_focus`

---

*[src/citra_qt/main.cpp, line 497 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSc989oDuinsOeNxMz:-KPSc98AEpZdBpfdUk-l:bcj2xlh) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra_qt/main.cpp#L497)):*
> ```cpp
>     // - the emulation is not started
>     // - the emulation has been paused through menu
>     if(!UISettings::values.pause_onfocuslost ||
> ```

Space after the `if`

---

*[src/citra_qt/main.cpp, line 502 \[r2\]](https://reviewable.io:443/reviews/citra-emu/citra/2006#-KPSeFZXTU1qfOo_Nhp8:-KPSeFZYSHtUV99Fpoqh:b-m0rt9n) ([raw file](https://github.com/citra-emu/citra/blob/4e097e05d2eb4bde2844f4eb026a66946a08fe1e/src/citra_qt/main.cpp#L502)):*
> ```cpp
>         return;
> 
>     if (hasFocus)
> ```

This if statement can just be collapsed as:

```cpp
emu_thread->SetRunning(has_focus);
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2006)*

neobrain:

2016-08-19T12:36:28Z: https://github.com/citra-emu/citra/compare/c3c6ac38e0a02b6400a73e218b0f3d90e9c08af8...LittleWhite-tb:0df188969f7954e0df4ac4233391170fe27af5d8

LittleWhite-tb:
Thanks @lioncash . I fixed what you have reported.

baka0815:
What's with a combo box then?
Something like
```
Focus Lost Behaviour: [________________________|_v_]
                      | do nothing             | (Standard)
                      | pause                  |
                      | pause with auto resume |
					  |________________________|
```

yuriks:
IMO the setting should be kept as a checkbox, as-is. Disabling the use of the manual pause is extremely unfriendly and that is a very useful feature, it just renders this change broken if that doesn't work.

With regards to the implementation: This is the wrong way of doing this. We already have a mechanism (at least in the Qt UI/Core integration) to pause emulation, and it should be maintained. The focusing issue is purely an UI issue, and signals from both the focus state and the menu option state should be combined to determine the actual pause/unpause calls sent to the emulation core. If this doesn't exist for SDL2 then it should be implemented in a similarly general manner.

wwylele:

LittleWhite-tb:
@yuriks : I would like to change the implementation, but I have no idea how I can design it to follow your idea. I can move the FocusIn/FocusOut in GMainWindow, but that's all.

src/citra_qt/main.cpp
```diff
@@ -468,6 +469,10 @@ void GMainWindow::OnStartGame() {
 
     ui.action_Pause->setEnabled(true);
     ui.action_Stop->setEnabled(true);
+
+    render_window->setFocus();
```
JayFoxRox: I didn't test this yet, but this sounds like starting a game is trying to grab focus now? I don't think this should happen. I can imagine situations where one wants to start Citra in the background (via cli for example)

Subv:
What's the status of this PR?

Subv:

PR #2027, opened 2016-08-29
IPC helper

Lectem:
The way citra generates and parses IPC requests right now is dangerous and inaccurate. Many services do not generate answers to IPC requests correctly, while it is not a problem for games emulation (nintendo's API doesn't seem to do any kind of checks) it could become an obstacle if/when IPC request translation is implemented.
An example of such problem can be seen here : https://github.com/citra-emu/citra/blob/master/src/core/hle/service/fs/fs_user.cpp#L72

``` c++
cmd_buff[1] = file_res.Code().raw;
if (file_res.Succeeded()) {
     cmd_buff[3] = Kernel::g_handle_table.Create(*file_res).MoveFrom();
} else {
    cmd_buff[3] = 0;
    LOG_ERROR(Service_FS, "failed to get a handle for file %s", file_path.DebugStr().c_str());
}
```

We have 2 issues : 
- The command header (cmd_buff[0]) is not updated to reflect the correct number of parameters of the request
- cmd_buff[2] should hold the correct handle descriptor

This PR aims to provide a simple helper to parse and generate IPC requests and thus fix those kind of errors. The helper would also be useful when dealing with u64 or other types being passed as arguments. It is not an attempt to implement IPC translation yet, while the skeleton of a translation method is present.

Note that it had been discussed long ago on IRC that the helper should be given directly to the service function, and that the command buffer shouldn't be used directly (right now the constructor of the helper calls Kernel::GetCommandBuffer()) but instead be copied+translated before being used by the service, and then translated+copied back after the service is done with it.

As a side note, this helper could have been written to deduce the command automagically from the function parameters, and be used as a wrapper such as what was done here https://github.com/wwylele/citra/commit/c8177812fd02f0598546c6117ff369e6fd2408fa#diff-b1293674c896ef771f60b74ee6ea7529L53

I'd like to know what approach you guys prefer and to have feedback on the current implementation.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -206,6 +206,7 @@ set(HEADERS
             hle/config_mem.h
             hle/function_wrappers.h
             hle/ipc.h
+            hle/ipc_helpers.h
             hle/applets/applet.h
             hle/applets/erreula.h
             hle/applets/mii_selector.h
--- a/src/core/hle/ipc.h
+++ b/src/core/hle/ipc.h
@@ -18,11 +18,25 @@ static const int kCommandHeaderOffset = 0x80; ///< Offset into command buffer of
  * the thread's TLS to an intermediate buffer in kernel memory, and then copied again to
  * the service handler process' memory.
  * @param offset Optional offset into command buffer
+ * @param offset Optional offset into command buffer (in bytes)
  * @return Pointer to command buffer
  */
 inline u32* GetCommandBuffer(const int offset = 0) {
     return (u32*)Memory::GetPointer(GetCurrentThread()->GetTLSAddress() + kCommandHeaderOffset +
                                     offset);
+}
+
+static const int kStaticBuffersOffset =
+    0x100; ///< Offset into static buffers, relative to command buffer header
+
+/**
+ * Returns a pointer to the static buffers area in the current thread's TLS
+ * TODO(Subv): cf. GetCommandBuffer
+ * @param offset Optional offset into static buffers area (in bytes)
+ * @return Pointer to static buffers area
+ */
+inline u32* GetStaticBuffers(const int offset = 0) {
+    return GetCommandBuffer(kStaticBuffersOffset + offset);
 }
 }
 
@@ -40,35 +54,34 @@ enum DescriptorType : u32 {
     CallingPid = 0x20,
 };
 
-/**
- * @brief Creates a command header to be used for IPC
- * @param command_id            ID of the command to create a header for.
- * @param normal_params         Size of the normal parameters in words. Up to 63.
- * @param translate_params_size Size of the translate parameters in words. Up to 63.
- * @return The created IPC header.
- *
- * Normal parameters are sent directly to the process while the translate parameters might go
- * through modifications and checks by the kernel.
- * The translate parameters are described by headers generated with the IPC::*Desc functions.
- *
- * @note While #normal_params is equivalent to the number of normal parameters,
- * #translate_params_size includes the size occupied by the translate parameters headers.
- */
-constexpr u32 MakeHeader(u16 command_id, unsigned int normal_params,
-                         unsigned int translate_params_size) {
-    return (u32(command_id) << 16) | ((u32(normal_params) & 0x3F) << 6) |
-           (u32(translate_params_size) & 0x3F);
-}
-
 union Header {
     u32 raw;
     BitField<0, 6, u32> translate_params_size;
-    BitField<6, 6, u32> normal_params;
+    BitField<6, 6, u32> normal_params_size;
     BitField<16, 16, u32> command_id;
 };
 
-inline Header ParseHeader(u32 header) {
-    return {header};
+/**
+* @brief Creates a command header to be used for IPC
+* @param command_id            ID of the command to create a header for.
+* @param normal_params_size         Size of the normal parameters in words. Up to 63.
+* @param translate_params_size Size of the translate parameters in words. Up to 63.
+* @return The created IPC header.
+*
+* Normal parameters are sent directly to the process while the translate parameters might go
+* through modifications and checks by the kernel.
+* The translate parameters are described by headers generated with the IPC::*Desc functions.
+*
+* @note While #normal_params_size is equivalent to the number of normal parameters,
+* #translate_params_size includes the size occupied by the translate parameters headers.
+*/
+inline u32 MakeHeader(u16 command_id, unsigned int normal_params_size,
+                      unsigned int translate_params_size) {
+    Header header;
+    header.command_id.Assign(command_id);
+    header.normal_params_size.Assign(normal_params_size);
+    header.translate_params_size.Assign(translate_params_size);
+    return header.raw;
 }
 
 constexpr u32 MoveHandleDesc(u32 num_handles = 1) {
@@ -83,7 +96,7 @@ constexpr u32 CallingPidDesc() {
     return CallingPid;
 }
 
-constexpr bool isHandleDescriptor(u32 descriptor) {
+constexpr bool IsHandleDescriptor(u32 descriptor) {
     return (descriptor & 0xF) == 0x0;
 }
 
@@ -91,30 +104,31 @@ constexpr u32 HandleNumberFromDesc(u32 handle_descriptor) {
     return (handle_descriptor >> 26) + 1;
 }
 
-constexpr u32 StaticBufferDesc(u32 size, u8 buffer_id) {
-    return StaticBuffer | (size << 14) | ((buffer_id & 0xF) << 10);
-}
-
 union StaticBufferDescInfo {
     u32 raw;
+    BitField<0, 4, u32> descriptor_type;
     BitField<10, 4, u32> buffer_id;
     BitField<14, 18, u32> size;
 };
 
-inline StaticBufferDescInfo ParseStaticBufferDesc(const u32 desc) {
-    return {desc};
+inline u32 StaticBufferDesc(u32 size, u8 buffer_id) {
+    StaticBufferDescInfo info;
+    info.descriptor_type.Assign(StaticBuffer);
+    info.buffer_id.Assign(buffer_id);
+    info.size.Assign(size);
+    return info.raw;
 }
 
 /**
- * @brief Creates a header describing a buffer to be sent over PXI.
- * @param size         Size of the buffer. Max 0x00FFFFFF.
- * @param buffer_id    The Id of the buffer. Max 0xF.
- * @param is_read_only true if the buffer is read-only. If false, the buffer is considered to have
- * read-write access.
- * @return The created PXI buffer header.
- *
- * The next value is a phys-address of a table located in the BASE memregion.
- */
+* @brief Creates a header describing a buffer to be sent over PXI.
+* @param size         Size of the buffer. Max 0x00FFFFFF.
+* @param buffer_id    The Id of the buffer. Max 0xF.
+* @param is_read_only true if the buffer is read-only. If false, the buffer is considered to have
+* read-write access.
+* @return The created PXI buffer header.
+*
+* The next value is a phys-address of a table located in the BASE memregion.
+*/
 inline u32 PXIBufferDesc(u32 size, unsigned buffer_id, bool is_read_only) {
     u32 type = PXIBuffer;
     if (is_read_only)
@@ -122,29 +136,30 @@ inline u32 PXIBufferDesc(u32 size, unsigned buffer_id, bool is_read_only) {
     return type | (size << 8) | ((buffer_id & 0xF) << 4);
 }
 
-enum MappedBufferPermissions {
+enum MappedBufferPermissions : u32 {
     R = 1,
     W = 2,
     RW = R | W,
 };
 
-constexpr u32 MappedBufferDesc(u32 size, MappedBufferPermissions perms) {
-    return MappedBuffer | (size << 4) | (u32(perms) << 1);
-}
-
 union MappedBufferDescInfo {
     u32 raw;
-    BitField<4, 28, u32> size;
+    BitField<0, 4, u32> flags;
     BitField<1, 2, MappedBufferPermissions> perms;
+    BitField<4, 28, u32> size;
 };
 
-inline MappedBufferDescInfo ParseMappedBufferDesc(const u32 desc) {
-    return {desc};
+inline u32 MappedBufferDesc(u32 size, MappedBufferPermissions perms) {
+    MappedBufferDescInfo info;
+    info.flags.Assign(MappedBuffer);
+    info.perms.Assign(perms);
+    info.size.Assign(size);
+    return info.raw;
 }
 
 inline DescriptorType GetDescriptorType(u32 descriptor) {
     // Note: Those checks must be done in this order
-    if (isHandleDescriptor(descriptor))
+    if (IsHandleDescriptor(descriptor))
         return (DescriptorType)(descriptor & 0x30);
 
     // handle the fact that the following descriptors can have rights
--- a/src/core/hle/ipc.h
+++ b/src/core/hle/ipc.h
@@ -62,19 +62,19 @@ union Header {
 };
 
 /**
-* @brief Creates a command header to be used for IPC
-* @param command_id            ID of the command to create a header for.
-* @param normal_params_size         Size of the normal parameters in words. Up to 63.
-* @param translate_params_size Size of the translate parameters in words. Up to 63.
-* @return The created IPC header.
-*
-* Normal parameters are sent directly to the process while the translate parameters might go
-* through modifications and checks by the kernel.
-* The translate parameters are described by headers generated with the IPC::*Desc functions.
-*
-* @note While #normal_params_size is equivalent to the number of normal parameters,
-* #translate_params_size includes the size occupied by the translate parameters headers.
-*/
+ * @brief Creates a command header to be used for IPC
+ * @param command_id            ID of the command to create a header for.
+ * @param normal_params_size         Size of the normal parameters in words. Up to 63.
+ * @param translate_params_size Size of the translate parameters in words. Up to 63.
+ * @return The created IPC header.
+ *
+ * Normal parameters are sent directly to the process while the translate parameters might go
+ * through modifications and checks by the kernel.
+ * The translate parameters are described by headers generated with the IPC::*Desc functions.
+ *
+ * @note While #normal_params_size is equivalent to the number of normal parameters,
+ * #translate_params_size includes the size occupied by the translate parameters headers.
+ */
 inline u32 MakeHeader(u16 command_id, unsigned int normal_params_size,
                       unsigned int translate_params_size) {
     Header header;
@@ -120,15 +120,15 @@ inline u32 StaticBufferDesc(u32 size, u8 buffer_id) {
 }
 
 /**
-* @brief Creates a header describing a buffer to be sent over PXI.
-* @param size         Size of the buffer. Max 0x00FFFFFF.
-* @param buffer_id    The Id of the buffer. Max 0xF.
-* @param is_read_only true if the buffer is read-only. If false, the buffer is considered to have
-* read-write access.
-* @return The created PXI buffer header.
-*
-* The next value is a phys-address of a table located in the BASE memregion.
-*/
+ * @brief Creates a header describing a buffer to be sent over PXI.
+ * @param size         Size of the buffer. Max 0x00FFFFFF.
+ * @param buffer_id    The Id of the buffer. Max 0xF.
+ * @param is_read_only true if the buffer is read-only. If false, the buffer is considered to have
+ * read-write access.
+ * @return The created PXI buffer header.
+ *
+ * The next value is a phys-address of a table located in the BASE memregion.
+ */
 inline u32 PXIBufferDesc(u32 size, unsigned buffer_id, bool is_read_only) {
     u32 type = PXIBuffer;
     if (is_read_only)
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -50,9 +50,9 @@ public:
     template <typename T>
     void Push(T value);
 
-    template <>
-    void Push(u32);
-
+    void Push(u32 value) {
+        cmdbuf[index++] = value;
+    }
     template <typename First, class... Other>
     void Push(First first_value, const Other&... other_values) {
         Push(first_value);
@@ -86,7 +86,7 @@ public:
 
     void PushCurrentPIDHandle() {
         Push(CallingPidDesc());
-        Push<u32>(0);
+        Push(u32(0));
     }
 
     void PushStaticBuffer(VAddr buffer_vaddr, u32 size, u8 buffer_id) {
@@ -99,6 +99,24 @@ public:
         Push(buffer_vaddr);
     }
 };
+
+/// Push ///
+
+template <>
+inline void RequestBuilder::Push<u32>(u32 value) {
+    Push(value);
+}
+
+template <>
+inline void RequestBuilder::Push<u64>(u64 value) {
+    Push(static_cast<u32>(value));
+    Push(static_cast<u32>(value >> 32));
+}
+
+template <>
+inline void RequestBuilder::Push<ResultCode>(ResultCode value) {
+    Push(value.raw);
+}
 
 class RequestParser : public RequestHelperBase {
 public:
@@ -123,39 +141,16 @@ public:
     template <typename T>
     T Pop();
 
-    template <>
-    u32 Pop<u32>();
-
     template <typename T>
-    void Pop(T& value) {
-        value = Pop<T>();
-    }
+    void Pop(T& value);
 
     template <typename First, class... Other>
-    void Pop(First& first_value, Other&... other_values) {
-        first_value = Pop<First>();
-        Pop(other_values...);
-    }
+    void Pop(First& first_value, Other&... other_values);
 
-    Kernel::Handle PopHandle() {
-        const u32 handle_descriptor = Pop<u32>();
-        DEBUG_ASSERT_MSG(IsHandleDescriptor(handle_descriptor),
-                         "Tried to pop handle(s) but the descriptor is not a handle descriptor");
-        DEBUG_ASSERT_MSG(HandleNumberFromDesc(handle_descriptor) == 1,
-                         "Descriptor indicates that there isn't exactly one handle");
-        return Pop<Kernel::Handle>();
-    }
+    Kernel::Handle PopHandle();
 
     template <typename... H>
-    void PopHandles(H&... handles) {
-        const u32 handle_descriptor = Pop<u32>();
-        const int handles_number = sizeof...(H);
-        DEBUG_ASSERT_MSG(IsHandleDescriptor(handle_descriptor),
-                         "Tried to pop handle(s) but the descriptor is not a handle descriptor");
-        DEBUG_ASSERT_MSG(handles_number == HandleNumberFromDesc(handle_descriptor),
-                         "Number of handles doesn't match the descriptor");
-        Pop(static_cast<Kernel::Handle&>(handles)...);
-    }
+    void PopHandles(H&... handles);
 
     /**
      * @brief Pops the static buffer vaddr
@@ -171,22 +166,7 @@ public:
      * buffer information
      * Please note that the setup uses virtual addresses.
      */
-    VAddr PopStaticBuffer(size_t* data_size = nullptr, bool useStaticBuffersToGetVaddr = false) {
-        const u32 sbuffer_descriptor = Pop<u32>();
-        StaticBufferDescInfo bufferInfo{sbuffer_descriptor};
-        if (data_size != nullptr)
-            *data_size = bufferInfo.size;
-        if (!useStaticBuffersToGetVaddr)
-            return Pop<VAddr>();
-        else {
-            ASSERT_MSG(0, "remove the assert if multiprocess/IPC translation are implemented.");
-            // The buffer has already been copied to the static buffer by the kernel during
-            // translation
-            Pop<VAddr>(); // Pop the calling process buffer address
-                          // and get the vaddr from the static buffers
-            return cmdbuf[(0x100 >> 2) + bufferInfo.buffer_id * 2 + 1];
-        }
-    }
+    VAddr PopStaticBuffer(size_t* data_size = nullptr, bool useStaticBuffersToGetVaddr = false);
 
     /**
      * @brief Pops the mapped buffer vaddr
@@ -197,26 +177,14 @@ public:
      * buffer
      */
     VAddr PopMappedBuffer(size_t* data_size = nullptr,
-                          MappedBufferPermissions* buffer_perms = nullptr) {
-        const u32 sbuffer_descriptor = Pop<u32>();
-        MappedBufferDescInfo bufferInfo{sbuffer_descriptor};
-        if (data_size != nullptr)
-            *data_size = bufferInfo.size;
-        if (buffer_perms != nullptr)
-            *buffer_perms = bufferInfo.perms;
-        return Pop<VAddr>();
-    }
+                          MappedBufferPermissions* buffer_perms = nullptr);
 
     /**
      * @brief Reads the next normal parameters as a struct, by copying it
      * @note: The output class must be correctly packed/padded to fit hardware layout.
      */
     template <typename T>
-    void PopRaw(T& value) {
-        static_assert(std::is_trivially_copyable<T>(), "Raw types should be trivially copyable");
-        std::memcpy(&value, cmdbuf + index, sizeof(T));
-        index += (sizeof(T) + 3) / 4; // round up to word length
-    }
+    void PopRaw(T& value);
 };
 
 /// Pop ///
@@ -238,22 +206,70 @@ inline ResultCode RequestParser::Pop<ResultCode>() {
     return ResultCode{Pop<u32>()};
 }
 
-/// Push ///
-
-template <>
-inline void RequestBuilder::Push<u32>(u32 value) {
-    cmdbuf[index++] = value;
+template <typename T>
+void RequestParser::Pop(T& value) {
+    value = Pop<T>();
 }
 
-template <>
-inline void RequestBuilder::Push<u64>(u64 value) {
-    Push(static_cast<u32>(value));
-    Push(static_cast<u32>(value >> 32));
+template <typename First, class... Other>
+void RequestParser::Pop(First& first_value, Other&... other_values) {
+    first_value = Pop<First>();
+    Pop(other_values...);
 }
 
-template <>
-inline void RequestBuilder::Push<ResultCode>(ResultCode value) {
-    Push(value.raw);
+inline Kernel::Handle RequestParser::PopHandle() {
+    const u32 handle_descriptor = Pop<u32>();
+    DEBUG_ASSERT_MSG(IsHandleDescriptor(handle_descriptor),
+                     "Tried to pop handle(s) but the descriptor is not a handle descriptor");
+    DEBUG_ASSERT_MSG(HandleNumberFromDesc(handle_descriptor) == 1,
+                     "Descriptor indicates that there isn't exactly one handle");
+    return Pop<Kernel::Handle>();
+}
+
+template <typename... H>
+void RequestParser::PopHandles(H&... handles) {
+    const u32 handle_descriptor = Pop<u32>();
+    const int handles_number = sizeof...(H);
+    DEBUG_ASSERT_MSG(IsHandleDescriptor(handle_descriptor),
+                     "Tried to pop handle(s) but the descriptor is not a handle descriptor");
+    DEBUG_ASSERT_MSG(handles_number == HandleNumberFromDesc(handle_descriptor),
+                     "Number of handles doesn't match the descriptor");
+    Pop(static_cast<Kernel::Handle&>(handles)...);
+}
+
+inline VAddr RequestParser::PopStaticBuffer(size_t* data_size, bool useStaticBuffersToGetVaddr) {
+    const u32 sbuffer_descriptor = Pop<u32>();
+    StaticBufferDescInfo bufferInfo{sbuffer_descriptor};
+    if (data_size != nullptr)
+        *data_size = bufferInfo.size;
+    if (!useStaticBuffersToGetVaddr)
+        return Pop<VAddr>();
+    else {
+        ASSERT_MSG(0, "remove the assert if multiprocess/IPC translation are implemented.");
+        // The buffer has already been copied to the static buffer by the kernel during
+        // translation
+        Pop<VAddr>(); // Pop the calling process buffer address
+                      // and get the vaddr from the static buffers
+        return cmdbuf[(0x100 >> 2) + bufferInfo.buffer_id * 2 + 1];
+    }
+}
+
+inline VAddr RequestParser::PopMappedBuffer(size_t* data_size,
+                                            MappedBufferPermissions* buffer_perms) {
+    const u32 sbuffer_descriptor = Pop<u32>();
+    MappedBufferDescInfo bufferInfo{sbuffer_descriptor};
+    if (data_size != nullptr)
+        *data_size = bufferInfo.size;
+    if (buffer_perms != nullptr)
+        *buffer_perms = bufferInfo.perms;
+    return Pop<VAddr>();
+}
+
+template <typename T>
+void RequestParser::PopRaw(T& value) {
+    static_assert(std::is_trivially_copyable<T>(), "Raw types should be trivially copyable");
+    std::memcpy(&value, cmdbuf + index, sizeof(T));
+    index += (sizeof(T) + 3) / 4; // round up to word length
 }
 
 } // namespace IPC
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -1,0 +1,259 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+#include "core/hle/ipc.h"
+#include "core/hle/kernel/kernel.h"
+
+namespace IPC {
+
+class RequestHelperBase {
+protected:
+    u32* cmdbuf;
+    ptrdiff_t index = 1;
+    Header header;
+
+public:
+    RequestHelperBase(u32* command_buffer, Header command_header)
+        : cmdbuf(command_buffer), header(command_header) {}
+
+    /// Returns the total size of the request in words
+    size_t TotalSize() const {
+        return 1 /* command header */ + header.normal_params_size + header.translate_params_size;
+    }
+
+    void ValidateHeader() {
+        DEBUG_ASSERT_MSG(index == TotalSize(), "Operations do not match the header (cmd 0x%x)",
+                         header.raw);
+    }
+};
+
+class RequestBuilder : public RequestHelperBase {
+public:
+    RequestBuilder(u32* command_buffer, Header command_header)
+        : RequestHelperBase(command_buffer, command_header) {
+        cmdbuf[0] = header.raw;
+    }
+    explicit RequestBuilder(u32* command_buffer, u32 command_header)
+        : RequestBuilder(command_buffer, Header{command_header}) {}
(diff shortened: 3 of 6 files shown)


neobrain:

2016-08-29T12:00:45Z: https://github.com/citra-emu/citra/compare/474586bc53eb6fda40fb0db23ea1d50d32af28b6...Lectem:775665451045d22a08562a318531480d24e9da6d

neobrain:
What kind of assembly does common usage of these helpers generate?

neobrain:

2016-08-29T13:57:55Z: https://github.com/citra-emu/citra/compare/474586bc53eb6fda40fb0db23ea1d50d32af28b6...Lectem:6576af512bd76390283be4739c1ca6df094ccaf3

lioncash:
Review status: 0 of 4 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[src/core/hle/kernel/session.cpp, line 13 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMMPGRDcfLATYTxHl1:-KQMMPGRDcfLATYTxHl2:byya9ym) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.cpp#L13)):*
> ```cpp
> /// Pop ///
> 
> template<> u32 Pop_Impl<u32>(MessageHelper & rh) { return rh.cmdbuf[rh.index++]; }
> ```

Ampersand against the type (ditto below)

---

*[src/core/hle/kernel/session.cpp, line 14 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMOF0K_fC-3lzvvkJx:-KQMOF0K_fC-3lzvvkJy:b-wkqg9c) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.cpp#L14)):*
> ```cpp
> 
> template<> u32 Pop_Impl<u32>(MessageHelper & rh) { return rh.cmdbuf[rh.index++]; }
> template<> ResultCode Pop_Impl<ResultCode>(MessageHelper & rh) { return ResultCode{ rh.cmdbuf[rh.index++] }; }
> ```

Maybe consider putting linebreaks in here like with the u64 specialization for consistency wrt reading.

---

*[src/core/hle/kernel/session.cpp, line 24 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMMMMIjCbZmAhVszqY:-KQMMMMIjCbZmAhVszqZ:br4139q) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.cpp#L24)):*
> ```cpp
> /// Push ///
> 
> template<> void Push_Impl<u32>(MessageHelper & rh, u32 value) { rh.cmdbuf[rh.index++] = value; }
> ```

ampersand against the type (ditto below)

---

*[src/core/hle/kernel/session.cpp, line 27 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMO8X5a46xSXqAjTLJ:-KQMO8X6rQ_6M_acLKxs:b345e03) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.cpp#L27)):*
> ```cpp
> template<> void Push_Impl<ResultCode>(MessageHelper & rh, ResultCode value){ rh.cmdbuf[rh.index++] = value.raw; }
> 
> template<> void Push_Impl<u64>(MessageHelper & rh,u64 value){
> ```

space after the comma

---

*[src/core/hle/kernel/session.cpp, line 28 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMMJhqwvCd6nxDwEzj:-KQMMJhrATqGeae-YvpF:b-51w6ut) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.cpp#L28)):*
> ```cpp
> 
> template<> void Push_Impl<u64>(MessageHelper & rh,u64 value){
>     rh.cmdbuf[rh.index++] = (u32) (value & 0xFFFFFFFF);
> ```

static_cast

---

*[src/core/hle/kernel/session.cpp, line 45 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMMYr36w1_qplLasUt:-KQMMYr36w1_qplLasUu:b-ee33zj) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.cpp#L45)):*
> ```cpp
> 
> MessageHelper::~MessageHelper() {
>     ValidateHeader(); // Make sure the message was parsed/built correctly
> ```

Destructors should be used for teardown, not validation.

---

*[src/core/hle/kernel/session.h, line 152 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMMlVkWFpsHj-fc_3D:-KQMMlVlG6nr-Tw0k9Sl:b6e61bd) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L152)):*
> ```objc
> // Pop
> template<class T>
> T Pop_Impl(MessageHelper &);
> ```

ampersand against the type (same for below)

---

*[src/core/hle/kernel/session.h, line 179 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMMsnrPcolGGTQ7Ku6:-KQMMsnrPcolGGTQ7Ku7:b-yrch84) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L179)):*
> ```objc
> 
>     /// Returns the total size of the request in words
>     size_t totalSize() const { return 1 /* command header */ + header.normal_params + header.translate_params_size; };
> ```

`TotalSize`, also consider breaking this onto two lines.

---

*[src/core/hle/kernel/session.h, line 181 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMN21uO_vYDkPV5O-o:-KQMN21uO_vYDkPV5O-p:b-phhzy) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L181)):*
> ```objc
>     size_t totalSize() const { return 1 /* command header */ + header.normal_params + header.translate_params_size; };
> 
>     void ValidateHeader(){
> ```

Space before the brace

---

*[src/core/hle/kernel/session.h, line 182 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMN-62Nr2FanoXGQnE:-KQMN-62Nr2FanoXGQnF:bgh9ieq) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L182)):*
> ```objc
> 
>     void ValidateHeader(){
>         DEBUG_ASSERT_MSG( index == totalSize(),"Operations do not match the header (cmd 0x%x)",MakeHeader(command_id,normal_params,translate_params_size));
> ```

Spaces after commas. Also unnecessary space at the beginning.

---

*[src/core/hle/kernel/session.h, line 192 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMN45rzoUQXqea7Mdr:-KQMN45rzoUQXqea7Mds:b-phhzy) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L192)):*
> ```objc
> 
>     template<class T>
>     T Pop(){ return detail::Pop_Impl<T>(*this); }
> ```

Space before the brace

---

*[src/core/hle/kernel/session.h, line 221 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMOWyupAb5ZCRfHsoH:-KQMOWyupAb5ZCRfHsoI:b-wo6ei0) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L221)):*
> ```objc
>      * Please note that the setup uses virtual addresses.
>      */
>     void PopStaticBuffer(VAddr &buffer_vaddr, size_t* data_size = nullptr) {
> ```

ampersand against the type

---

*[src/core/hle/kernel/session.h, line 244 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMOaNpsDyMWBGTAGoU:-KQMOaNpsDyMWBGTAGoV:bop4q72) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L244)):*
> ```objc
>      * @param[out] buffer_perms If non-null, the pointed value will be set to the permissions of the buffer
>      */
>     void PopMappedBuffer(VAddr &buffer_vaddr, size_t* data_size = nullptr, MappedBufferPermissions* buffer_perms=nullptr) {
> ```

ampersand against the type. Also space before and after `=`

---

*[src/core/hle/kernel/session.h, line 264 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMOgoDJHrLB0EWwgmc:-KQMOgoE0ppfBTA5T3wa:b-wgkmbh) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L264)):*
> ```objc
> 
>     template<class T>
>     void Push(T value){ return detail::Push_Impl(*this,value); }
> ```

Space before `{`

---

*[src/core/hle/kernel/session.h, line 274 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMNV0ncSzMT5sqhClb:-KQMNV0ncSzMT5sqhClc:b-dc49xq) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L274)):*
> ```objc
>     template<class T>
>     void PushRaw(T&& value) {
>         static_assert(std::is_trivially_copyable<T>::value);
> ```

```cpp
static_assert(std::is_trivially_copyable<T>());
```

---

*[src/core/hle/kernel/session.h, line 280 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMOnqIRwLi3PvVbmtC:-KQMOnqJxpmB-WV-g_4k:bxbq0w3) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L280)):*
> ```objc
> 
>     template<class... H>
>     void PushHandles(bool copy_handles,H... handles) {
> ```

Space after the comma

---

*[src/core/hle/kernel/session.h, line 336 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMNOrmLApXDuvL3svk:-KQMNOrnjytjr4zU1Kbk:b2tv7fg) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L336)):*
> ```objc
>  * @return Pointer to static buffers area
>  */
> inline static u32* GetStaticBuffers(const int offset = 0) {
> ```

`static` shouldn't be present on functions in the header.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2027)*

neobrain:

2016-09-07T11:41:33Z: https://github.com/citra-emu/citra/compare/c3307b41db4e222ceca3860c75ccbe6300484729...Lectem:fce2fbbaf92c084673228c332e63bb01caec70a7

Lectem:
@lioncash Addressed all the formatting issues, but didn't change 

> src/core/hle/kernel/session.cpp, line 45 [r1] (raw file):
> 
>     MessageHelper::~MessageHelper() {
>         ValidateHeader(); // Make sure the message was parsed/built correctly
> 
> Destructors should be used for teardown, not validation.

How would you do it then ? Just let the user call ValidateHeader ?
Also ValidateHeader is empty in non-debug builds.

I also moved the Pop/Push_Impl specializations to the header, because VS2015 couldn't inline those, resulting in a lot of calls.

lioncash:
Review status: 0 of 4 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[src/core/hle/kernel/session.h, line 221 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMOWyupAb5ZCRfHsoH:-KR3WR-GqKV-N91aDVkO:bk1vo73) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L221)):*
> ampersand against the type
</details>
This now needs a space after the ampersand

---

*[src/core/hle/kernel/session.h, line 244 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMOaNpsDyMWBGTAGoU:-KR3WUieH9NDKASSXa4k:bk1vo73) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L244)):*
> ampersand against the type. Also space before and after `=`
</details>
This now needs a space after the ampersand

---

*[src/core/hle/kernel/session.h, line 280 \[r1\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMOnqIRwLi3PvVbmtC:-KR3Wale_4qCC1XKcseY:byqrbp3) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L280)):*
> Space after the comma
</details>
This still applies

---

*[src/core/hle/kernel/session.h, line 160 \[r4\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KR3X-wyc4UHBHAni9m1:-KR3X-wyc4UHBHAni9m2:b-cda0a6) ([raw file](https://github.com/citra-emu/citra/blob/fce2fbbaf92c084673228c332e63bb01caec70a7/src/core/hle/kernel/session.h#L160)):*
> ```objc
> } // namespace detail
> 
> class MessageHelper{
> ```

Missing space before the brace.

---

*[src/core/hle/kernel/session.h, line 185 \[r4\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KR3WLOpmfior0rNP631:-KR3WLOpmfior0rNP632:b-tfp5cq) ([raw file](https://github.com/citra-emu/citra/blob/fce2fbbaf92c084673228c332e63bb01caec70a7/src/core/hle/kernel/session.h#L185)):*
> ```objc
> 
>     void ValidateHeader() {
>         DEBUG_ASSERT_MSG(index == totalSize(), "Operations do not match the header (cmd 0x%x)",MakeHeader(command_id,normal_params,translate_params_size));
> ```

This should also be `TotalSize()`

---

*[src/core/hle/kernel/session.h, line 308 \[r4\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KR3XGbuClaCj3U4nwtr:-KR3XGbuClaCj3U4nwts:b-m17d90) ([raw file](https://github.com/citra-emu/citra/blob/fce2fbbaf92c084673228c332e63bb01caec70a7/src/core/hle/kernel/session.h#L308)):*
> ```objc
>     }
> ```

Two newlines too many?

---

*[src/core/hle/kernel/session.h, line 337 \[r4\]](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KR3Whx1oBmdtZqoBoDZ:-KR3Whx2CpTzsaR3_-BN:b-btgm4v) ([raw file](https://github.com/citra-emu/citra/blob/fce2fbbaf92c084673228c332e63bb01caec70a7/src/core/hle/kernel/session.h#L337)):*
> ```objc
> 
>     template<> inline void Push_Impl<u64>(MessageHelper& rh, u64 value) {
>         rh.cmdbuf[rh.index++] = static_cast<u32>(value & 0xFFFFFFFF);
> ```

Note that ANDing here isn't necessary with the cast.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2027)*

neobrain:

2016-09-07T12:25:26Z: https://github.com/citra-emu/citra/compare/c3307b41db4e222ceca3860c75ccbe6300484729...Lectem:7d17c944b0c5702f6d18463dabaa23f914ff955e

neobrain:

2016-09-07T12:39:01Z: https://github.com/citra-emu/citra/compare/c3307b41db4e222ceca3860c75ccbe6300484729...Lectem:cd0377af142bb1d5cdc269dec5cf8d69b461ba43

yuriks:
Review status: 0 of 4 files reviewed at latest revision, 22 unresolved discussions.

---

*[src/core/hle/kernel/session.cpp, line 45 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KQMMYr36w1_qplLasUt:-KRO3PYAi_Xaq2m2vB94:bwz926q) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.cpp#L45)):*
> Destructors should be used for teardown, not validation.
</details>
I think adding validation on the destructor is fine. Lots of things run arbitrary stuff in the destructor and it's a common idiom in C++ now. Otherwise, provide some rationale for why this should be avoided.

---

*[src/core/hle/kernel/session.h, line 145 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KR2dlVQ0dVKDOvPg2CP:-KR2dlVQ0dVKDOvPg2CQ:b-rs7kfv) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L145)):*
> ```objc
> // This define is used so that the correct code would be ready for the day when IPC translation will be done correctly.
> // Meanwhile, some behaviours are stubbed since citra uses the same TLS for both source and destination processes during an IPC request.
> #define NO_MULTIPROCESS_AND_IPC_TRANSLATION
> ```

This should be settable per-handler/helper instance to allow a gradual transition.

---

*[src/core/hle/kernel/session.h, line 159 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KR2eTRn2vY88RBVJH0D:-KR2eTRn2vY88RBVJH0E:bt1uza4) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L159)):*
> ```objc
> } // namespace detail
> 
> class MessageHelper{
> ```

Split into MessageParser and MessageBuilder, with something like MessageParser::BeginReply returning the MessageBuilder and handing-off ownership of the buffer from one to the other.

---

*[src/core/hle/kernel/session.h, line 264 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KR2dLoWtyvDlvRON_o4:-KR2dLoWtyvDlvRON_o5:b-cg4c08) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L264)):*
> ```objc
> 
>     template<class T>
>     void Push(T value){ return detail::Push_Impl(*this,value); }
> ```

Having something that accepts both u32 and u64 like this is kinda dangerous and error prone, you should find a way of disallowing it and having distinct named functions for them to avoid any accidental conversions, etc.

---

*[src/core/hle/kernel/session.h, line 280 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KR2dTQlJXrWY-BZYfEU:-KR2dTQm7pVeVDKHe5ux:b-4us26h) ([raw file](https://github.com/citra-emu/citra/blob/6576af512bd76390283be4739c1ca6df094ccaf3/src/core/hle/kernel/session.h#L280)):*
> ```objc
> 
>     template<class... H>
>     void PushHandles(bool copy_handles,H... handles) {
> ```

Split this into two PushCopyHandles and PushMoveHandles. You should boolean parameters like these, and in this case you don't have any significant sharing anyway.

---

*[src/core/hle/kernel/session.h, line 323 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KRO2Lj1EmFCD0pJ2jD5:-KRO2Lj2WJ9v5JYVPcwN:b-rqhfr) ([raw file](https://github.com/citra-emu/citra/blob/cd0377af142bb1d5cdc269dec5cf8d69b461ba43/src/core/hle/kernel/session.h#L323)):*
> ```objc
> 
>     template<> inline u64 Pop_Impl<u64>(MessageHelper& rh) {
>         u64 value = ((u64)rh.cmdbuf[rh.index + 1] << 32) | rh.cmdbuf[rh.index];
> ```

You should implement this in terms of two u32 pops.

---

*[src/core/hle/kernel/session.h, line 336 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KRO2OhmTPRwVJMNShPW:-KRO2OhmTPRwVJMNShPX:b-7w3y4g) ([raw file](https://github.com/citra-emu/citra/blob/cd0377af142bb1d5cdc269dec5cf8d69b461ba43/src/core/hle/kernel/session.h#L336)):*
> ```objc
>     template<> inline void Push_Impl<u64>(MessageHelper& rh, u64 value) {
>         rh.cmdbuf[rh.index++] = static_cast<u32>(value);
>         rh.cmdbuf[rh.index++] = static_cast<u32>((value >> 32) & 0xFFFFFFFF);
> ```

Same here, implement as two u32 pushes. And also do the same on the ResultCode implementation below. In general, try to use primitives you have already implemented.

---

*[src/core/hle/service/y2r_u.cpp, line 293 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KRO2rcgRM89_EEj5K7i:-KRO2rch5ckbwWFiyCUN:b293dyn) ([raw file](https://github.com/citra-emu/citra/blob/cd0377af142bb1d5cdc269dec5cf8d69b461ba43/src/core/hle/service/y2r_u.cpp#L293)):*
> ```cpp
>     conversion.src_Y.gap = rh.Pop<u32>();
>     Handle src_process_handle;
>     rh.PopHandles(src_process_handle);
> ```

I think that the 1-handle version of this is common enough that you can add a PopHandle that just retuns the handle value, to keep the code flowing better.

---

*[src/core/hle/service/fs/fs_user.cpp, line 60 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KRO4-756r5LeP7sJBEN:-KRO4-76hk4TrNRsI9G4:b6yzcqg) ([raw file](https://github.com/citra-emu/citra/blob/cd0377af142bb1d5cdc269dec5cf8d69b461ba43/src/core/hle/service/fs/fs_user.cpp#L60)):*
> ```cpp
>     rh.Pop<u32>(); // Always 0 ?
>     ArchiveHandle archive_handle = rh.Pop<u64>();
>     auto filename_type    = static_cast<FileSys::LowPathType>(rh.Pop<u32>());
> ```

Get rid of the alignment, it's more distracting than anything here.

---

*[src/core/hle/service/fs/fs_user.cpp, line 65 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2027#-KRO3uTTlRCARTXd4YOD:-KRO3uTUM9Ib1Ul-F8ye:b2hgrre) ([raw file](https://github.com/citra-emu/citra/blob/cd0377af142bb1d5cdc269dec5cf8d69b461ba43/src/core/hle/service/fs/fs_user.cpp#L65)):*
> ```cpp
>     u32 attributes        = rh.Pop<u32>(); // TODO(Link Mauve): do something with those attributes.
>     VAddr filename_ptr;
>     rh.PopStaticBuffer(filename_ptr);
> ```

Why doesn't this just return the VAddr?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2027)*

wwylele:

Subv:
What's the status of this PR? @Lectem

Lectem:
I will try to address @yurik's comments as soon as possible, for now I updated the PR so that it doesn't conflict anymore, and I seperated builder/parser. I've been quite busy lately and totally forgot yuriks reviewed it.

Lectem:
All @yuriks' comments are now addressed except this one : 

> Having something that accepts both u32 and u64 like this is kinda dangerous and error prone, you should find a way of disallowing it and having distinct named functions for them to avoid any accidental conversions, etc.

If you think it truly is a problem, then we could just remove the detail::push/pop_impl I suppose

Also

> I think adding validation on the destructor is fine. Lots of things run arbitrary stuff in the destructor and it's a common idiom in C++ now. Otherwise, provide some rationale for why this should be avoided.

In the end I removed it from the destructors to avoid trouble, and just put it in the MakeBuilder by default (but you can disable it with the last param)

Please forgive me if there are some formatting/naming issues even after running clang-format, I'll fix it later if you find some (it's quite late here, sorry)

Lectem:
Just added the missing #include <cstring> // for memcpy in ipc.h

src/core/hle/kernel/ipc.h
```diff
@@ -0,0 +1,412 @@
...
+        index += (sizeof(T) - 1) / 4 + 1; // round up to word length
+    }
+
+    /**
+     * Translate the parameters
+     * @return true on success, false otherwise
+     * @note The proper message will be created for the error on failure
+     * @note Will reset the index of the message to 1, as if the helper was just created
+     */
+    bool Translate(u32* dest_cmdbuffer);
```
yuriks: This does not belong here. Just get rid of this function for now.

src/core/hle/kernel/ipc.h
```diff
@@ -0,0 +1,412 @@
...
+        auto bufferInfo = ParseMappedBufferDesc(sbuffer_descriptor);
+        if (data_size != nullptr)
+            *data_size = bufferInfo.size;
+        if (buffer_perms != nullptr)
+            *buffer_perms = bufferInfo.perms;
+        return Pop<VAddr>();
+    }
+
+    template <class T>
+    void PopRaw(T&& value) {
```
yuriks: Missing documentation.

src/core/hle/kernel/ipc.h
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    template <class First, class... Other>
+    void Push(First first_value, Other&&... other_values) {
+        Push(first_value);
+        Push(std::forward<Other>(other_values)...);
+    }
+
+    template <class T>
+    void PushRaw(T&& value) {
```
yuriks: Missing documentation.

src/core/hle/kernel/ipc.cpp
```diff
@@ -0,0 +1,80 @@
+// Copyright 2016 Citra Emulator Project
```
yuriks: The parsing/builder helpers don't belong in the kernel. Move them to `core/hle/ipc_helpers.{cpp,h}` instead.

src/core/hle/service/y2r_u.cpp
```diff
@@ -283,37 +283,42 @@ static void GetTransferEndEvent(Service::Interface* self) {
...
-    conversion.src_U.image_size = cmd_buff[2];
-    conversion.src_U.transfer_unit = cmd_buff[3];
-    conversion.src_U.gap = cmd_buff[4];
+    conversion.src_U.address = rp.Pop<u32>();
+    conversion.src_U.image_size = rp.Pop<u32>();
+    conversion.src_U.transfer_unit = rp.Pop<u32>();
+    conversion.src_U.gap = rp.Pop<u32>();
+    Handle src_process_handle;
+    rp.PopHandles(src_process_handle);
+    rp.ValidateHeader();
```
yuriks: Redundant call to `ValidateHeader`

src/core/hle/service/y2r_u.cpp
```diff
@@ -283,37 +283,42 @@ static void GetTransferEndEvent(Service::Interface* self) {
...
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    IPC::RequestBuilder rb = rp.MakeBuilder(1);
+    rb.Push(RESULT_SUCCESS);
 
     LOG_DEBUG(Service_Y2R, "called image_size=0x%08X, transfer_unit=%hu, transfer_stride=%hu, "
                            "src_process_handle=0x%08X",
               conversion.src_U.image_size, conversion.src_U.transfer_unit, conversion.src_U.gap,
-              cmd_buff[6]);
+              src_process_handle);
 }
```
yuriks: The RequestBuilders need ValidateHeader calls too somehow. (Destructor, or an explicit call.)

src/core/hle/service/y2r_u.cpp
```diff
@@ -283,37 +283,42 @@ static void GetTransferEndEvent(Service::Interface* self) {
...
-
-    cmd_buff[0] = IPC::MakeHeader(0x10, 1, 0);
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    IPC::RequestParser rp({0x00100102}); // The helper should be passed by argument to the function
+    conversion.src_Y.address = rp.Pop<u32>();
+    conversion.src_Y.image_size = rp.Pop<u32>();
+    conversion.src_Y.transfer_unit = rp.Pop<u32>();
+    conversion.src_Y.gap = rp.Pop<u32>();
+    Handle src_process_handle;
+    rp.PopHandles(src_process_handle);
```
Subv: Isnt this supposed to allow you to pop multiple handles at the same time?

src/core/hle/kernel/ipc.h
```diff
@@ -0,0 +1,412 @@
...
+        return (DescriptorType)(descriptor & 0x30);
+
+    // handle the fact that the following descriptors can have rights
+    if (descriptor & MappedBuffer)
+        return MappedBuffer;
+
+    if (descriptor & PXIBuffer)
+        return PXIBuffer;
+
+    return StaticBuffer;
```
Lectem: The lower bits can contain rights, for example a descriptor of a mapped buffer can give the following :
    (descriptor & 0x0F == 0b1110 ) != MappedBuffer

src/core/hle/kernel/ipc.h
```diff
@@ -0,0 +1,412 @@
...
+class RequestHelperBase {
+protected:
+    u32* cmdbuf;
+    ptrdiff_t index = 1;
+    Header header;
+
+public:
+    RequestHelperBase(u16 command_id, unsigned normal_params, unsigned translate_params_size);
+    RequestHelperBase(Header command_header);
+    RequestHelperBase(u32* command_buffer, Header command_header);
```
Lectem: Didn't make it inline because of leftovers from the previous implementation (that required Kernel::GetCommandBuffer()), I'll change this !

src/core/hle/service/y2r_u.cpp
Lectem: @Subv it does, but we now have one version for single handles that is more coherent with the other push/pop usage, I'll update it

Lectem:
I will rebase once @yuriks reviewed the new version

src/core/hle/kernel/ipc.h
```diff
@@ -160,69 +148,70 @@ class RequestHelperBase {
...
-        cmdbuf[0] = header.raw;
-    }
-    RequestBuilder(Header command_header) : RequestHelperBase(command_header) {
-        cmdbuf[0] = header.raw;
-    }
     RequestBuilder(u32* command_buffer, Header command_header)
         : RequestHelperBase(command_buffer, command_header) {
         cmdbuf[0] = header.raw;
     }
+    explicit RequestBuilder(u32* command_buffer, u32 command_header)
```
yuriks: Add a blank line separating these constructors.

yuriks: There's no need to make a constructor taking more than one argument explicit

src/core/hle/kernel/ipc.h
```diff
@@ -245,168 +234,158 @@ class RequestBuilder : public RequestHelperBase {
...
-    friend T detail::Pop_Impl(RequestParser&);
-
 public:
-    RequestParser(const RequestHelperBase& base) : RequestHelperBase(base) {}
-    RequestParser(u16 command_id, unsigned normal_params, unsigned translate_params_size)
-        : RequestHelperBase(command_id, normal_params, translate_params_size) {}
-    RequestParser(Header command_header) : RequestHelperBase(command_header) {}
+    RequestParser(u32* command_buffer, Header command_header)
+        : RequestHelperBase(command_buffer, command_header) {}
+    explicit RequestParser(u32* command_buffer, u32 command_header)
```
yuriks: Blank lines, unnecessary `explicit`

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -53,7 +53,8 @@ static void Initialize(Service::Interface* self) {
  *      3 : File handle
  */
 static void OpenFile(Service::Interface* self) {
-    IPC::RequestParser rp({0x080201C2}); // The helper should be passed by argument to the function
+    // The helper should be passed by argument to the function
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), {0x080201C2});
```
yuriks: You don't need the `{}`

src/core/CMakeLists.txt
```diff
@@ -181,6 +180,7 @@ set(HEADERS
             hle/kernel/client_port.h
             hle/kernel/event.h
             hle/kernel/ipc.h
+            hle/kernel/ipc_helpers.h
```
yuriks: This file shouldn't be in `kernel/` I think, put it directly in `hle/`

src/core/hle/kernel/ipc.h
yuriks: Ack

src/core/hle/kernel/ipc_helpers.h
```diff
@@ -0,0 +1,257 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+#include "ipc.h"
```
yuriks: Use a fully qualified include path

src/core/hle/kernel/ipc_helpers.h
```diff
@@ -0,0 +1,257 @@
...
+    RequestParser(u32* command_buffer, Header command_header)
+        : RequestHelperBase(command_buffer, command_header) {}
+    explicit RequestParser(u32* command_buffer, u32 command_header)
+        : RequestParser(command_buffer, Header{command_header}) {}
+    RequestParser(u32* command_buffer, u16 command_id, unsigned normal_params,
+                  unsigned translate_params_size)
+        : RequestParser(command_buffer,
+                        MakeHeader(command_id, normal_params, translate_params_size)) {}
+
+    RequestBuilder MakeBuilder(u32 normal_params, u32 translate_params_size = 0,
```
yuriks: `translate_params_size` shouldn't have a default value. Also, why is it `normal_params` but `translate_params_*size*`?

Lectem: Well, I just kept the old names, normal parameters are all considered to be of the size of a u32, while translate params have descriptors and variable number of values per descriptor, so in this case nb params != size in u32.
I don't mind adding _size to normal_params for the sake of clarity though

Lectem:
Here is what APT looks like when using the helpers (with a few fixes/upate for APT) : 
https://github.com/Lectem/citra/commit/8cf74cd0e7438d25da93683299dca5edde2e521c

bunnei:
Taking this out of RFC as it has gotten a full review (and @yuriks approved it for merge), feel free to put it back if this is not the case (**or close it and re-open a new PR once you are ready**, as RFCs are not meant to be open for months).

I plan on reviewing as well later today / early tomorrow. Reminder to others to review as well (cc @Subv).

src/core/hle/ipc.h
```diff
@@ -18,12 +18,26 @@ static const int kCommandHeaderOffset = 0x80; ///< Offset into command buffer of
...
  * @param offset Optional offset into command buffer
+ * @param offset Optional offset into command buffer (in bytes)
  * @return Pointer to command buffer
  */
 inline u32* GetCommandBuffer(const int offset = 0) {
     return (u32*)Memory::GetPointer(GetCurrentThread()->GetTLSAddress() + kCommandHeaderOffset +
                                     offset);
 }
+
+static const int kStaticBuffersOffset =
```
Subv: Uuuugh clang-format, y u do dis.

src/core/hle/ipc.h
```diff
@@ -40,35 +54,34 @@ enum DescriptorType : u32 {
...
     BitField<0, 6, u32> translate_params_size;
-    BitField<6, 6, u32> normal_params;
+    BitField<6, 6, u32> normal_params_size;
     BitField<16, 16, u32> command_id;
 };
 
-inline Header ParseHeader(u32 header) {
-    return {header};
+/**
+* @brief Creates a command header to be used for IPC
```
Subv: It seems this comment block is misaligned

src/core/hle/ipc.h
```diff
@@ -83,68 +96,70 @@ constexpr u32 CallingPidDesc() {
...
- * @brief Creates a header describing a buffer to be sent over PXI.
- * @param size         Size of the buffer. Max 0x00FFFFFF.
- * @param buffer_id    The Id of the buffer. Max 0xF.
- * @param is_read_only true if the buffer is read-only. If false, the buffer is considered to have
- * read-write access.
- * @return The created PXI buffer header.
- *
- * The next value is a phys-address of a table located in the BASE memregion.
- */
+* @brief Creates a header describing a buffer to be sent over PXI.
```
Subv: Same here

src/core/hle/ipc_helpers.h
```diff
@@ -0,0 +1,275 @@
...
+        Push(MappedBufferDesc(size, perms));
+        Push(buffer_vaddr);
+    }
+};
+
+/// Push ///
+
+template <>
+inline void RequestBuilder::Push<u32>(u32 value) {
+    Push(value);
```
wwylele: I saw there is a `Push(u32)` overload but this looks like function recursively calling itself. I am not sure if one is allowed to make overload functions in this way...

wwylele:
>About Push and Pop, I wonder if anything relies on the contents of the TLS being preserved across sync requests.
For example, A service may use STRB to write only the lower byte of a response parameter, leaving the other bytes untouched.

Yes, this did happen when I was testing CAM service. All parameters with actual types of `bool`, `u8`, `s8`, `u16`and `s16` share the same issue. I see that one can currently use `PopRaw` to achieve the correct behaviour, but still better if we can have template specialization for `Pop` to explicitly write `Pop<bool>()`

Lectem:
> Yes, this did happen when I was testing CAM service. All parameters with actual types of bool, u8, s8, u16and s16 share the same issue. I see that one can currently use PopRaw to achieve the correct behaviour, but still better if we can have template specialization for Pop to explicitly write Pop<bool>()

Right, it happens in many places such as Y2R too, should I add those now or in another PR ?

Dragios:
@Lectem It's best to add those in another PR instead so that this can be merged first. Smaller PR is easier to review and other developers also agrees with that.

src/core/hle/ipc_helpers.h
```diff
@@ -0,0 +1,275 @@
...
+
+template <typename... H>
+void RequestParser::PopHandles(H&... handles) {
+    const u32 handle_descriptor = Pop<u32>();
+    const int handles_number = sizeof...(H);
+    DEBUG_ASSERT_MSG(IsHandleDescriptor(handle_descriptor),
+                     "Tried to pop handle(s) but the descriptor is not a handle descriptor");
+    DEBUG_ASSERT_MSG(handles_number == HandleNumberFromDesc(handle_descriptor),
+                     "Number of handles doesn't match the descriptor");
+    Pop(static_cast<Kernel::Handle&>(handles)...);
```
Subv: Is this static_cast needed?

Lectem: It is needed for template deduction but we could indeed use `std::is_same<H, Kernel::Handle>...` though I think it wouldn't make a difference since it is a reference cast ?

Lectem: It is needed for template deduction but we could indeed use std::is_same<H, Kernel::Handle>... though I think it wouldn't make a difference since it is a reference cast ?

wwylele: Please add `std::is_same<H, Kernel::Handle>...`. This function shouldn't be used for any other types

src/core/hle/ipc_helpers.h
wwylele: Actually I think you can just move the implementation of `Push(u32 value)` here and get rid of that overload function. I think it is more readable

src/core/hle/ipc_helpers.h
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    template <typename T>
+    void Push(T value);
+
+    void Push(u32 value) {
+        cmdbuf[index++] = value;
+    }
+    template <typename First, class... Other>
+    void Push(First first_value, const Other&... other_values) {
```
wwylele: Should the first parameter also be `const First&`? correct me if I am wrong.

mailwl:
I dislike it
Nintendo just put values in +1  +2, +64, +65. 
should we be close to devs?

mailwl:
sorry for my previous post. it can me merged for sure, but don't push to use it

wwylele:
@mailwl Please don't throw random words. If you are referring to [Server->Client static buffer transfer](https://www.3dbrew.org/wiki/IPC#Static_Buffer_Translation] [like this](https://github.com/citra-emu/citra/blob/905fc92ce1f7d99d8819a17b180a559b3a9f15de/src/core/hle/service/apt/apt.h#L413), they are not implemented here. We will improve the IPC helper (and probably the IPC part in kernel) before rewrite these service functions

wwylele:
@mailwl except that what you said is not a Nin nature either. The nature is the documentation I put above, which is not some mystery number "+1 +2, +64, +65.", and does not conflict with what is in this PR - it is just _unimplemented_ .

mailwl:
as i said, merge it, but do not push to use it.

yuriks:
@mailwl There's no reason we should be beholden to how Nintendo implements things. As long as the visible end result is the same we can implement things however is more convenient/leads to a better codebase.

(Also, it's pretty clear from looking at some service disassemblies that they also have a similar library internally, because the compile sometimes fails to inline some of the pusher methods and they can still be seen in the code.)

src/core/hle/ipc_helpers.h
Lectem: Indeed.

src/core/hle/ipc_helpers.h
Lectem: Yeah, I fixed all this on the other branch (the one for the next PR)

https://github.com/Lectem/citra/blob/effebfd4fcc3b14b16c439785d3a924c3f4b03bf/src/core/hle/ipc_helpers.h

I didn't put it in this PR because I wanted to get it merged, but if you prefer I can do it

Lectem:
@mailwl Why do you think Nintendo doesn't have wrappers ? I know for a fact that they do (you can find a few of those in most games with debug info). And what does it even have to do with how citra implements things anyway ?
Do you have another reason behind this ? The only reason one wouldn't like this is because it makes it harder to know what the offset of the values are. But if you understand how IPC command buffers work, then you shouldn't have problems with it to start with. The reason why we need those helpers is because most of the time, header codes are ignored, or descriptors are just magic values and in many cases wrong. This gave birth to ugly hacks in ctrulib for example https://github.com/smealum/ctrulib/blob/master/libctru/source/services/apt.c#L123.
I don't think we need to enforce the use of the helpers, but you need to know what you're doing when you implement services. I've seen too many copy pastes from 3dbrew that aren't complete/correct.

mailwl:
@Lectem, sorry, i dont know. Services is very stripped, and we can only view something like 
v1 = v2[1];
v3= v2 [2];

mailwl:
BTW, may we talk about 3ds SDK? it enable for all, but forbid for use in emulators (Nintendo don't want to make iur life easier)

yuriks:
@mailwl I would really rather you not, because it's very easy to taint the project's legal status with that kind of thing.

mailwl:
ah, i see, no need for violations

wwylele:
@mailwl As I already told you hundreds of times, please bring these "BTW" off-topic talks to our IRC channel. Under PR and issue there should be only talk about the PR and issue itself

mailwl:
my irc client will end trial.
may you suggest another? For Win10

wwylele:
@mailwl I uses HexChat. It is free and open source. Probably not the best one but it is easy for me to use.

PR #2032, opened 2016-08-30
Qt graphics configure & V-Sync option

bunnei:
Adds a new tab to the Qt Configuration UI for graphics settings, and moves relevant settings there. Also adds a setting to enable V-Sync, mostly just as a quick & easy frame limiter until we have a better solution. Obligatory screenshot:

### DIFF
--- a/src/citra/config.cpp
+++ b/src/citra/config.cpp
@@ -71,6 +71,7 @@ void Config::ReadValues() {
     Settings::values.use_hw_renderer = sdl2_config->GetBoolean("Renderer", "use_hw_renderer", true);
     Settings::values.use_shader_jit = sdl2_config->GetBoolean("Renderer", "use_shader_jit", true);
     Settings::values.use_scaled_resolution = sdl2_config->GetBoolean("Renderer", "use_scaled_resolution", false);
+    Settings::values.use_vsync = sdl2_config->GetBoolean("Renderer", "use_vsync", false);
 
     Settings::values.bg_red   = (float)sdl2_config->GetReal("Renderer", "bg_red",   1.0);
     Settings::values.bg_green = (float)sdl2_config->GetReal("Renderer", "bg_green", 1.0);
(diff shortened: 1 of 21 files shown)


neobrain:

2016-08-30T01:49:46Z: https://github.com/citra-emu/citra/compare/3f515c01bc68aa23f80ab5337e08cdc513a6f017...bunnei:74842116b287165d1db959437a439f13a17c724c

RavenHome1:
why not use a frame rate locker like most other emulators do.

MerryMage:
Code looks fine to me, though you could probably `s/toogle/toggle/` while you're at it.

degasus:
@RavenHome1 Please do not use frame rate locker. The hardware is able to render at various framerates, or to disable scanout. Counting and limiting CPU ticks should work by far better ;)

MerryMage:
@degasus 

    <merry> we probably will need a frame limiter in the near future
    <merry> and probably one that does it based on cycles-executed and not frames-outputted
    <yuriks_> merry: frames output is fine, we have a VSync callback that's tried to the cyclerate
    <merry> ah nice
      merry goes dig for that
    <yuriks_> I think it's in hw/gpu.cpp
    <yuriks_> that said the isolation between that and the GSP is kinda bad right now because the "interrupts" are just a direct function call, so it might need some
              tweaking
    <merry> If I'm reading this correctly the vblank event is solely dependent on a timer and not guest application-controlled?
    <yuriks_> that's how I remember it being
    <neobrain> the vblank event technically is set when a corresponding interrupt fires
    <neobrain> (which presumably happens on hw according to a particular timer)
    <neobrain> (of course, on citra it may be anything)
    <yuriks_> yeah, on the kernel you can get events that are set when an interrupt fires
    <yuriks_> but on citra we just use a regular event and reset it on that interval
    <merry> gotcha, yuriks_, neobrain
    <yuriks_> tbh the other interrupts are just barely correct, but they work fine
    <yuriks_> could use some restructuring so it's a bit more accurate to hardware in how it works under the hood

neobrain:

2016-09-01T02:12:35Z: https://github.com/citra-emu/citra/compare/3f515c01bc68aa23f80ab5337e08cdc513a6f017...bunnei:a71013cf16668a8d781a890823b6f6f203b865e9

Issue #1981, opened 2016-07-27
Super Mario 3D Land crashing


azoreseuropa:
I am using Mac OS X version: Can you test this? Super Mario 3D Land EUROPE isn't playable and crashing Citra when World 1-1 was brought up to the screen for the first time. The latest version is citra-20160726-f8f6722-osx-amd64.tar.xz at Citra Nightly because it was playable on June 8 the version - citra-20160608-ba4fb41-osx-amd64.tar.xz

JayFoxRox:
This is a known issue with the shader related to lighting. So I'm kind of surprised that you claim that it worked fine in the past.

azoreseuropa:
Yes, it was playable after World 1-1 was brought up to the screen without crashing and I played it for a while in World 1-1 only. And it was not crashing as far as I can go in World 1-1, thats all. I guess I will wait for the next version or so for more testing. :)

azoreseuropa:
Do you need log information of what caused it to crashes ?

JayFoxRox:
Is this resolved?

azoreseuropa:
Yes, it is no longer crashes.

gamedropswithpops:
hi super mario 3d land still crashes in world 8 bowsers castle after your shoot out of the second canon , is there any fix for this btw I am using https://github.com/citra-emu/citra-bleeding-edge/releases build 
thanks

freiro:
@gamedropswithpops please test on current master (bleeding edge is "unsupported") and eventually open a new ticket.

PR #2034, opened 2016-08-30
OpenGL: Avoid error on unsupported lighting LUT

JayFoxRox:
Apparently kids these days don't know how to code.
This error came up so often on the forums that I felt like it was easier to finally push my ~~fix~~ workaround for it, instead of telling them what to do again (which usually resulted in more questions).

---

Essentially all this does is add a dummy zero constant into the shader in case of an unimplemented branch (which still gives a critical log error).
This avoids generating invalid shaders which would stop execution due to the uniform block size mismatch.

IIRC there is a similar case elsewhere in that code but I don't remember exactly. If the error comes up on the forums again I'll be here with a follow-up PR.

---

This shouldn't really change anything aside from preventing a ~~crash~~ failed assertion on entering World 1-1 in Super Mario 3D Land. I've [mentioned said error here](https://github.com/citra-emu/citra/issues/1981#issuecomment-235861835). I'm not sure if #1981 is about this crash though.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_gen.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_gen.cpp
@@ -400,6 +400,7 @@ static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
         default:
             LOG_CRITICAL(HW_GPU, "Unknown lighting LUT input %d\n", (int)input);
             UNIMPLEMENTED();
+            index = "0.0";
             break;
         }
 


JayFoxRox:
@neobrain and @yuriks provided some feedback / food-for-thought on IRC about this to which I completely agree. I feel like it's worth mentioning.

---

The proper way to handle this piece of code would be to use:
```
UNIMPLEMENTED_MSG("Unknown lighting LUT input %d\n", (int)input);
```

However, this will cause an assertion which is not restricted to debug builds like the current `UNIMPLEMENTED` macro being used. We were not even sure why `UNIMPLEMENTED` is [only a debug macro to begin with](https://github.com/citra-emu/citra/blob/ef6873980eb3854d351c319514fca4116165283c/src/common/assert.h#L52). `UNIMPLEMENTED` means the emulation is tainted and we should not trust it.

This in turn means that any game like Mario Land would stop with a controlled exit instead of continuing to run as intended by this PR.

A similar situation exists for unknown shader opcodes which we [tread as a `CRITICAL` warning](https://github.com/citra-emu/citra/blob/f40fabd688576fae7ab91195547a6967178b28dd/src/video_core/shader/shader_jit_x64.cpp#L789) (even without `UNIMPLEMENTED` for sete and emit / without `UNREACHABLE` for anything else).
In the shader interpreter we don't seperate into `UNIMPLEMENTED` / `UNREACHABLE` either and weirdly [use the `ERROR` log level instead](https://github.com/citra-emu/citra/blob/552018c50a216c850c0ece02bdd9df99c5907007/src/video_core/shader/shader_interpreter.cpp#L635)

To summarize: Logging / asserting is a mess in the current codebase.

---

In my opinion we'd cause a massive outrage amongst our users if we suddenly caused `UNIMPLEMENTED` to stop emulation. It would cause exits in games which otherwise work fine and they can't easily recompile either to remove the assert.
So despite seeming illogical first, I actually think we should make both, `UNIMPLEMENTED` + `UNIMPLEMENTED_MSG` debug assertions.

* This means users will be able to play but their log will reveal that the emulation might have been tainted.
* Developers can still run in a strict mode but they are able to skip assertions with a debugger or to remove the asserts for testing. This should be discussed elsewhere in the future though. I just wanted to state my opinion on it.

---

For this issue in particular I'd like to use my reasoning above, claiming that my current workaround is fine.
Without the `index = 0.0;` we are actually depending on the driver to return a GLSL error which may give different results for different users - this means this workaround is a lot cleaner than not doing anything. Changing to `UNIMPLEMENTED_MSG` right now would make the `index = 0.0;` useless (unless we make it a debug assert) but also be shitty for our users.

However, what I probably should do, is to add a seperate case for the lut input mode which fails in Mario Land. That case would have to be handled by the proposed `index = 0.0;` workaround and an additional log entry (which level?) + `UNIMPLEMENTED` (debug assert). The default case would then be solely for buggy games or actual (serious) errors and could be switched to `UNREACHABLE` (always assert).

Something similar could be added for the shader code or assert.h could be changed, but I think it's beyond the scope of the PR and might come with side-effects.

Thoughts?

bunnei:
LGTM. Agreed with the above. In any software product ever, assertions are turned off in release except for the most critical of critical conditions (e.g. the software literally can no longer continue), and this isn't that case, neither are 99% of UNIMPLEMENTED conditions. While whether Citra being a "software product" may be up for debate, I do agree that the benefits of silencing people who just want to play games outweigh the negatives of getting feedback from these users, and I think most of us developers are pretty well aware of debug assertions.

If this really becomes any more of a debate, we can propose a better solution for UNIMPLEMENTED/UNREACHABLE (which is warranted anyway), but nonetheless it's out of scope for this PR.

yuriks:
ftr I think this change as-is LGTM too

PR #2036, opened 2016-09-01
Include the msvc runtime dlls with citra

jroweboy:
As a potential solution to the "download all the dlls" problem, just include the msvc runtime dlls in the msvc build.

There shouldn't be any issue with licensing really. https://www.visualstudio.com/en-us/downloads/2015-redistribution-vs#visual-c-runtime They are on the list of "okay to redistribute" and the license says that its fine as well https://msdn.microsoft.com/en-us/library/ms235299.aspx

Another solution is to do a silent install as part of the installer, but this patch is much simpler.

Edit: I tested this in a win 10 VM that didn't have the redistributables, and it worked. But further testing is always appreciated.

### DIFF
--- a/src/citra_qt/CMakeLists.txt
+++ b/src/citra_qt/CMakeLists.txt
@@ -121,7 +121,11 @@ if (Qt5_FOUND AND MSVC)
         Qt5Widgets$<$<CONFIG:Debug>:d>.*
     )
     windows_copy_files(citra-qt ${Qt5_PLATFORMS_DIR} ${PLATFORMS} qwindows$<$<CONFIG:Debug>:d>.*)
+    # copy the msvc redistributable dlls to the folder
+    set(MSVC_REDIST_DIR "C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/redist/x64/Microsoft.VC140.CRT")
+    windows_copy_files(citra-qt ${MSVC_REDIST_DIR} ${DLL_DEST} *.dll)
 
+    unset(MSVC_REDIST_DIR)
     unset(Qt5_DLL_DIR)
     unset(Qt5_PLATFORMS_DIR)
     unset(DLL_DEST)


neobrain:

2016-09-01T02:08:51Z: https://github.com/citra-emu/citra/compare/87893e6d68079ff4fa277688adb508d041944385...jroweboy:7fac9071ff34371e58dcebf90e357c3b75144bf0

jroweboy:
JayFoxRox mentioned that adding size to the build right now is a bad idea because of the limited bandwidth on the build servers. To counter this, I also opened #2037 which should reduce the final build size much more than this will add. Maybe that one should be merged first.

At some point we should consider builds.citra-emu to be a mirror site and host the builds on other sites as well :)

DaMan69:
Only Windows10 comes with the UCRT.  MS does support app-local deployment of that, but its a lot of dlls.

yuriks:
Last I checked, including the DLL files like this is very much discouraged, unsupported and iirc goes against the license. What you are allowed to do is include the *redistributable installer*, which would be something to perhaps do in an installer. (Though really should only be done on-demand if the user lacks the runtime already.)

jroweboy:
I provided links to the license and showed that its fine to include them.
Its not discouraged, unsupported, and as far as I can tell doesn't go
against the license (i'm hesitant to say that it doesn't for sure because i
don't wanna get legal things wrong). I do think that there are better
solutions and I'm very much open to ideas. I would like to include the
runtime installer as part of the installer, but a citra installer not a
thing (yet). If you think this isn't worth merging, thats fine by me as
well. This was just one of several PRs I wrote a while ago to try to
provide solutions for common user problems and just left them rotting on my
local machine. Thank you for the feedback :)

jroweboy:
Going to agree with Yuriks on this one. Might as well just wait and put it in as part of the installer.

PR #2037, opened 2016-09-01
MSVC: Add RelWithDebInfo and removing debugging from Release.

jroweboy:
The builds used to contain pdb information, and while it compresses well, it takes up too much bandwidth to host. Developers can use the RelWithDebInfo to get a fast build with debugging symbols. This should reduce build size on windows from roughly 12MB compressed to 7MB.

PR #2039, opened 2016-09-01
Create a separate archive for debug symbols on windows

jroweboy:
In relation to https://github.com/citra-emu/citra/pull/2037 it seemed like the consensus was RelWithDebInfo doesn't add much and it'd be better to just move the pdb files to a separate zip if they are needed for that build.

This PR uploads them to a different directory, but that dir may not exist on the build server. That may require manual intervention.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,10 +42,15 @@ on_success:
             $GITREV = $(git show -s --format='%h')
             # Where are these spaces coming from? Regardless, let's remove them
             $BUILD_NAME = "citra-${GITDATE}-${GITREV}-windows-amd64.7z" -replace " ",""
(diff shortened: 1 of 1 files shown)


neobrain:

2016-09-01T17:08:51Z: https://github.com/citra-emu/citra/compare/b9c857be819bf82ce7a5d38e18f41bb82e68a483...jroweboy:93bcd00e44f42757307cf76ef7c810637ac9dd85

MerryMage:
LGTM

bunnei:
LGTM

jroweboy:
Oh, I was expecting only this one would get merged, since they both do the same thing, just in different ways. Since the other one https://github.com/citra-emu/citra/pull/2037 was merged, VS won't output debugging info at all so this one will just upload empty files. If you still want to keep the other PR as well, I can make them upload the files from the RelWithDebugInfo folder instead. Otherwise reverting https://github.com/citra-emu/citra/pull/2037 works fine too.

PR #2045, opened 2016-09-02
travis: Update to XCode 7.3.1

MerryMage:
Xcode 7.3.1 provides a compiler with better C++14 support compared to [Xcode 6.1](https://docs.travis-ci.com/user/languages/objective-c/#Supported-OS-X-iOS-SDK-versions) which Travis uses by default.

Had to remove `xcpretty` (log prettyfier) because the version of Ruby provided on these images was segfaulting!

I've set `MACOSX_DEPLOYMENT_TARGET=10.9` so we're still producing a OS X 10.9 compatible build despite using a 10.11 build machine.

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -23,11 +23,12 @@ if [ "$TRAVIS_OS_NAME" = "linux" -o -z "$TRAVIS_OS_NAME" ]; then
 elif [ "$TRAVIS_OS_NAME" = "osx" ]; then
     set -o pipefail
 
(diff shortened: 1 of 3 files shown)


neobrain:

2016-09-02T15:53:42Z: https://github.com/citra-emu/citra/compare/274b925a9705ebd5c4ec242c70d44249dc9f2b0d...MerryMage:d93cfb32fef5b76d4181a8bf7c1e378881483a75

Issue #2005, opened 2016-08-15
pokemon mystery dungeon makes citra crush at a point


smartkaartik:
When on hazy path we are about to battle with scraggy and gudaurr citra crushes and closes.

![capture](https://cloud.githubusercontent.com/assets/14974622/17656862/4eb8e992-62dc-11e6-888b-c5aca3b8b3c1.PNG)

System specs 
Windows 10 64bit
Intel I3 3.0ghz
Nvidia geforce gt 610 2gb

wwylele:
dupe of #1682 ?

smartkaartik:
Not that as the game player in #1682 is far behind me

j-selby:
Can confirm, is dupe as the same issue has occurred (just without the same visual effect). Working on "solution" in the other thread.

wwylele:
Should be fixed by #2103. If not please leave a comment and reopen this.

Issue #1928, opened 2016-06-23
Hot Wheels: Worlds Best Driver - Crashed at ingame


JohnGodgames:
It's an old issue and it's still isn't solved. 

Unhandled exception at 0x00007FF79728BC0B in citra-qt.exe: 0xC0000094: Integer division by zero.

` size_t contiguous_input_size = config.texture_copy.size / input_width * (input_width + input_gap);
`

**Call stack:**
```
>	citra-qt.exe!GPU::Write<unsigned int>(unsigned int addr, const unsigned int data) Line 190	C++
 	citra-qt.exe!GSP_GPU::ExecuteCommand(const GSP_GPU::Command & command, unsigned int thread_id) Line 540	C++
 	citra-qt.exe!GSP_GPU::TriggerCmdReqQueue(Service::Interface * self) Line 598	C++
 	citra-qt.exe!Service::Interface::SyncRequest() Line 80	C++
 	citra-qt.exe!SVC::SendSyncRequest(unsigned int handle) Line 232	C++
 	citra-qt.exe!HLE::Wrap<&SVC::SendSyncRequest>() Line 134	C++
 	citra-qt.exe!SVC::CallSVC(unsigned int immediate) Line 1107	C++
 	citra-qt.exe!InterpreterMainLoop(ARMul_State * cpu) Line 3530	C++
 	citra-qt.exe!ARM_DynCom::ExecuteInstructions(int num_instructions) Line 87	C++
 	citra-qt.exe!Core::RunLoop(int tight_loop) Line 50	C++
 	citra-qt.exe!EmuThread::run() Line 52	C++
 	Qt5Core.dll!00000000643c7fef()	Unknown
 	[External Code]	

```

Dragios:
@JohnGodgames Could you please test the hotfix and see if it does fix the problem.

wwylele:
Should be fixed by #2103. If not please leave a comment and reopen this.

PR #2057, opened 2016-09-07
Check for invalid TextureCopy size when copying

### LINKED ISSUE #1682: Pokemon Mystery Dungeon Gates to Infinity: Unhandled exception: Integer division by zero.
Hello , sorry for my bad English. I found this exception when I was testing the battles at the end of the first dungeon of the game.

When you start the battle , the value of output_width and output_gap are not getting properly.

screenbattle

exceptiondivisionbyzero

j-selby:
Fixes #1682, #2005 and likely #1928

This commit implements a check on both the input and output width when copying a texture. While this doesn't fix any potential underlying issues, it does provide the following:
- Proper logging and handling of this behavior
- Allows games which utilize this behavior to execute with no noticeable glitch

Breaking from this new block is required as the loop which follows stalls when attempting to copy a zero-sized texture.

### DIFF
--- a/src/core/hw/gpu.cpp
+++ b/src/core/hw/gpu.cpp
@@ -187,6 +187,14 @@ inline void Write(u32 addr, const T data) {
                     u32 output_width = config.texture_copy.output_width * 16;
                     u32 output_gap = config.texture_copy.output_gap * 16;
 
+                    if (input_width == 0) {
+                        input_width = 1024 * 16;
+                    }
+
+                    if (output_width == 0) {
+                        output_width = 1024 * 16;
+                    }
+
                     size_t contiguous_input_size = config.texture_copy.size / input_width * (input_width + input_gap);
                     Memory::RasterizerFlushRegion(config.GetPhysicalInputAddress(), static_cast<u32>(contiguous_input_size));
 


neobrain:

2016-09-07T12:20:57Z: https://github.com/citra-emu/citra/compare/c3307b41db4e222ceca3860c75ccbe6300484729...j-selby:f962c4ca88f8126240c26abc631f2e6d08612012

neobrain:

2016-09-07T12:22:38Z: https://github.com/citra-emu/citra/compare/c3307b41db4e222ceca3860c75ccbe6300484729...j-selby:da0ee4904a4b8e7237788cc1a3696bf5d470d562

src/core/hw/gpu.cpp
```diff
@@ -187,6 +187,13 @@ inline void Write(u32 addr, const T data) {
                     u32 output_width = config.texture_copy.output_width * 16;
                     u32 output_gap = config.texture_copy.output_gap * 16;
 
+                    if (input_width == 0 || output_width == 0) {
+                        LOG_ERROR(HW_GPU, "TextureCopy: using texture that has zero width (input: %u, output: %u)",
```
bunnei: Probably can be a LOG_CRITICAL... LOG_ERROR is often used for "expected errors" (e.g. real error conditions caused by the game that would in fact happen on real HW), and therefore is often ignored.

neobrain:

2016-09-07T21:39:46Z: https://github.com/citra-emu/citra/compare/c3307b41db4e222ceca3860c75ccbe6300484729...j-selby:bdf980a786889b44dfb9841156a147ac4ebf41a6

j-selby:
Addressed.

src/core/hw/gpu.cpp
```diff
@@ -187,6 +187,13 @@ inline void Write(u32 addr, const T data) {
                     u32 output_width = config.texture_copy.output_width * 16;
                     u32 output_gap = config.texture_copy.output_gap * 16;
 
+                    if (input_width == 0 || output_width == 0) {
+                        LOG_CRITICAL(HW_GPU, "TextureCopy: using texture that has zero width (input: %u, output: %u)",
```
JayFoxRox: it's not zero-width. It might also be zero-height.

j-selby: Would zero size be the correct phrasing then?

Dragios: Maybe a _null value_ will be a better phrase?

j-selby: @Dragios Um, no? Null would imply that the texture is undefined (rather then simply zero-sized).

yuriks:
I did these tests a long time ago, but I seem to recall that when 0 was passed to these, the hardware actually interpreted it as a large size such as 1024. Might be worth testing this kind of behavior to see if it produces better results on the corresponding games (and then hw verifying them.)

neobrain:

2016-09-08T06:26:38Z: https://github.com/citra-emu/citra/compare/722af0703efd58d2c2de9edbac763355053601a2...j-selby:7ee875aec3ed73e0309f8d32bd8577be8f6c7cc9

j-selby:
1024 * 16 does round out nicely to 2 ^ 10, so that would make sense.

While I can't see anything in PMD:GTI that changed with that change, it has made everything happy anyway and I am sure there are cases which do abuse this behavior.

Reference:

devenv-08-09-2016_16-14-26

wwylele:
What about dumping the texture with the said size to see if it actually makes sense?

j-selby:
Yeah, sorry - I was looking to validate this behaviour and I'll get on it ASAP

j-selby:
Alrighty then. The PhysicalInputAddress maps to 0x18120000 (in all instances that I've encountered) which is the default Color Buffer. This does of course map to the actual dimensions of the screen (width=256) which would make sense within the context in which it is used. Input and output width should therefore be 256 (* 16), rather then 1024 (* 16). This does produce the required fading effect without any visible issues.

texture-0x18120000
devenv-14-09-2016_19-28-21

wwylele:
Superseded by #2103

PR #2059, opened 2016-09-07
audio_core: Tweak audio latency

MerryMage:
Fiddling with numbers.

Outputting 1024 samples and dropping samples after 1024 gave pretty much no slack to timing issues.

### DIFF
--- a/src/audio_core/hle/dsp.cpp
+++ b/src/audio_core/hle/dsp.cpp
@@ -105,7 +105,7 @@ static void OutputCurrentFrame(const StereoFrame16& frame) {
         std::vector<s16> stretched_samples = time_stretcher.Process(sink->SamplesInQueue());
         sink->EnqueueSamples(stretched_samples.data(), stretched_samples.size() / 2);
     } else {
-        constexpr size_t maximum_sample_latency = 1024; // about 32 miliseconds
+        constexpr size_t maximum_sample_latency = 2048; // about 64 miliseconds
         if (sink->SamplesInQueue() > maximum_sample_latency) {
             // This can occur if we're running too fast and samples are starting to back up.
             // Just drop the samples.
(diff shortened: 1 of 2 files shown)


neobrain:

2016-09-07T14:29:36Z: https://github.com/citra-emu/citra/compare/c3307b41db4e222ceca3860c75ccbe6300484729...MerryMage:991cdda3fc536e7d674205a4f1b54465562974e3

yuriks:
LGTM

This makes a massive difference with time stretching disabled: The omnipresent crackling is converted to very occasional glitches, and it seemed to actually have *reduced* (at least perceived) latency. I even managed to combo to 106. :P

jroweboy:
Is there any debate on whether this should be an option or something? I don't really know if this is something that should directly be accessible to the end user, but I do know of some other emulators that let you configure audio latency. (For the record, I'm fine with this change, but I was just curious if people think its a good idea to have a an audio latency config option)

bunnei:
@jroweboy, yes, this was discussed on IRC. It can be a separate change, and it might want to be something we add at some point (at least as an internal/INI configuration).

Personally, I'd like to avoid going down the route of emulator config creep where you end up with a UI with so many technical knobs/switches that the vast majority of every day users don't understand, where we also end up with lots of fine tuned game-specific configurations. Wherever possible, I would prefer to invest the little bit of extra time into finding the setting that "just works" everywhere, as that's how I'd like the emulator to be for our users (some day, obviously we're not there yet).

yuriks:
Output buffer size is a reasonable-ish option to have since the minimum safe value depends on your hardware and system, and if you want to be able to minimize audio latency you're going to want to set that as low as possible.

The other setting though should probably either be fixed or computed from the buffer size somehow. That said, I tried tweaking the values and the existing constants seems to be optimal for my system already.

PR #2060, opened 2016-09-07
Use travis cache for cmake and sdl2 on linux builds

Lectem:
CMake takes forever to download, and the same can be said for the SDL2 compilation. Caching those will give a gain from 160 up to 250 seconds on linux builds from what I experienced. (Once cached it is instant).

If a new version is needed, travis admins can delete the cache manually.


neobrain:

2016-09-07T16:17:19Z: https://github.com/citra-emu/citra/compare/c3307b41db4e222ceca3860c75ccbe6300484729...Lectem:e1f5ca92688eaff2753b53d9eb7027fb094b2137

MerryMage:
LGTM.

yuriks:
Did they change their caching implementation? I remember in the past they said to not bother caching downloads because they would just download their cache from S3 anyway.

Anyway, if this increases builds speed, LGTM.

Lectem:
They said so but downloading from the official cmake website is really slow. They also say "The cache’s purpose is to make installing language-specific dependencies easy and fast" so I guess cmake fits the description.

PR #2065, opened 2016-09-11
Add likely() and unlikely() macros in common_funcs, and use them in GetUserPath().

linkmauve:
This helps with code locality when optimising for the common path.

### DIFF
--- a/src/common/common_funcs.h
+++ b/src/common/common_funcs.h
@@ -29,6 +29,14 @@
     #define FORCE_INLINE inline __attribute__((always_inline))
 #endif
 
+#if defined(__GNUC__)
+    #define likely(x) __builtin_expect(!!(x), 1)
+    #define unlikely(x) __builtin_expect(!!(x), 0)
+#else
+    #define likely(x) (x)
+    #define unlikely(x) (x)
+#endif
+
 #ifndef _MSC_VER
 
 #ifdef ARCHITECTURE_x86_64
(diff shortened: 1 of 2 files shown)


neobrain:

2016-09-11T04:15:44Z: https://github.com/citra-emu/citra/compare/722af0703efd58d2c2de9edbac763355053601a2...linkmauve:71bb7877982c91b70a79510b02cbd32281b585e9

yuriks:
Are these functions called enough for this to matter here? If they are then we should probably fix *that* instead. You can probably get better mileage out of this by adding them in the drawing code.

lioncash:
I disagree with adding things like these because they're somewhat halfassed ways to fix performance issues inherent with code itself. If branch prediction is actually an issue, then the code itself is usually what should be looked at.

This is essentially entering micro-optimization territory.

yuriks:
I'm fine with the likely/unlikely macros, and they have their uses in hot code, but the uses in GetUserPath seem pretty superfluous (unless there's a measured improvement that can be shown here).

lioncash:
> I'm fine with the likely/unlikely macros, and they have their uses in hot code

In most cases, even in those scenarios the gains you'd get from the usage of the macro would be around the nanosecond level of difference. Mostly imperceptible and completely not worth it. It also adds another level of bikeshedding to the review process when making changes to code that makes use of those macros and then either has them removed or added to code. It raises questions along the lines of, "have you checked that removing/adding this macro has any effect?", "have you benchmarked this usage?", etc (since whenever an optimization PR is opened, people love to omit details and usually don't include this information). I can assure you not every contributor will lay out all that pertinent information off the bat.

Is it really worth adding that to the review process that already goes slow enough as it is?

PR #2044, opened 2016-09-02
Qt: unify running detection of system settings configure

wwylele:
Clean up some ugly code and replace them with the one introduced in #2032.

PR #2069, opened 2016-09-13
Qt: fix birthday combo box updating

wwylele:
Fix a bug in #2044 and a hidden bug long ago (Sorry!). ~~The birthday combo box didn't get filled properly because the signal-slot connection was performed after reading system setting data. Now the connection is moved to the ui file. the code order is changed. Also explicitly update the combo box in `ReadSystemSetting` because `setCurrentIndex` won't always emit `currentIndexChanged` - if `setCurrentIndex` sets the same index as it was (in this case, if the birthday is in January), `currentIndexChanged` won't be emitted.~~

The old bug: when the birthday is in January, `setCurrentIndex` in `ReadSystemSetting` won't emit `currentIndexChanged` because the index doesn't actually get changed, resulting birthday combo box unfilled.

Bug in #2044, the `ReadSystemSetting` call moves before the connecting, and `setCurrentIndex` will never fill the birthday combo box.

Solution: just explicitly call the updating funcion.

In addition, switched the order of connection and `ReadSystemSetting` call
>[yuriks] wwylele: I think it's good for code cleanliness to call the connect before calling the initialize function anyway

### DIFF
--- a/src/citra_qt/configure_system.cpp
+++ b/src/citra_qt/configure_system.cpp
@@ -18,9 +18,9 @@ ConfigureSystem::ConfigureSystem(QWidget *parent) :
     QWidget(parent),
     ui(new Ui::ConfigureSystem) {
     ui->setupUi(this);
-    this->setConfiguration();
-
     connect(ui->combo_birthmonth, SIGNAL(currentIndexChanged(int)), SLOT(updateBirthdayComboBox(int)));
+
+    this->setConfiguration();
 }
 
 ConfigureSystem::~ConfigureSystem() {
@@ -60,6 +60,7 @@ void ConfigureSystem::ReadSystemSettings() {
     // set birthday
     std::tie(birthmonth, birthday) = Service::CFG::GetBirthday();
     ui->combo_birthmonth->setCurrentIndex(birthmonth - 1);
+    updateBirthdayComboBox(birthmonth - 1); // explicitly update it because the signal from setCurrentIndex is not reliable
     ui->combo_birthday->setCurrentIndex(birthday - 1);
 
     // set system language


neobrain:

2016-09-13T05:16:04Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...wwylele:a21be66034ff040f117281857fdc55e0ebc9f888

Dragios:
Screenshot for this particular issue

system tab

src/citra_qt/configure_system.cpp
```diff
@@ -19,8 +19,6 @@ ConfigureSystem::ConfigureSystem(QWidget *parent) :
     ui(new Ui::ConfigureSystem) {
     ui->setupUi(this);
     this->setConfiguration();
-
-    connect(ui->combo_birthmonth, SIGNAL(currentIndexChanged(int)), SLOT(updateBirthdayComboBox(int)));
```
yuriks: Why don't you simply move this line above setConfiguration? We usually do connects in the code, and I'd like to keep everything consistent.

wwylele: I randomly picked [one file where connection is in the ui file](https://github.com/citra-emu/citra/blob/master/src/citra_qt/configure_debug.ui#L86) and thought it's a common case...

neobrain:

2016-09-13T07:57:32Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...wwylele:b64b62dd4eef414476101f414da2b1a677ce2aff

neobrain:

2016-09-13T08:00:37Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...wwylele:3a70c6d8b2558ba9abd57fdef89ac169b0dc3c97

neobrain:

2016-09-13T08:02:24Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...wwylele:aea5e9b892e984419d887b8a75ef6d1cd1fe99d0

neobrain:

2016-09-13T08:12:49Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...wwylele:80f0bf1f3149c5c166ac64957d30bf388ed114e6

neobrain:

2016-09-13T08:16:19Z: https://github.com/citra-emu/citra/compare/5cd37016639bcb1dea8a692562759ccf9d904b20...wwylele:c0262001d81df07fd07a3fd89b8c6ccee74c5b92

yuriks:
LGTM

Issue #2053, opened 2016-09-06
Video / Texture forward: assertion failure in debug build


wwylele:
This is an old issue. Open this so that we won't forget.

Some games experiencing assertion failure in debug build in texture forward code:
![bissue](https://cloud.githubusercontent.com/assets/4592895/18268708/91cede26-7457-11e6-9897-f16f78449849.PNG)

In release build there is nothing wrong.

Games with this issue known to me are _Pokemon OR/AS_ and _Paper Mario Sticker Star_. Probably more.

yuriks:
Pokemon ORAS? I wonder if this is related to the stencil clearing bug.

wwylele:
Closed by #2103

PR #2078, opened 2016-09-15
renderer_opengl: check if surface_interval is empty

bogang233:
Now the emulator won't crash in horde battles and rotation battles in Pokemon OR/AS.
I don't know how does it work, but it's indeed fixed.
Anyone has a better idea?

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -232,6 +232,7 @@ CachedSurface* RasterizerCacheOpenGL::GetSurface(const CachedSurface& params, bo
     float exact_surface_goodness = -1.f;
 
     auto surface_interval = boost::icl::interval<PAddr>::right_open(params.addr, params.addr + params_size);
+    if (boost::icl::is_empty(surface_interval)) return nullptr;
     auto range = surface_cache.equal_range(surface_interval);
     for (auto it = range.first; it != range.second; ++it) {
         for (auto it2 = it->second.begin(); it2 != it->second.end(); ++it2) {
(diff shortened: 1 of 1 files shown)


neobrain:

2016-09-15T07:12:49Z: https://github.com/citra-emu/citra/compare/81bb315839a95998fe20fdcdb5cee0f161335185...bogang233:23d3a52abeb3c04ec9711fa6075b18884fbc0436

wwylele:
Does this also fix #2053 ?

bogang233:
Maybe.

wwylele:
Can this be the same situation with #2057? That is: sometimes game inputs width=0 (causing empty interval with params_size=0) which should be interpreted as a large number?

bogang233:
I think they are not same. This PR doesn't fix bugs in #2057, and #2057 also crashed in this situation.

wwylele:
@bogang233 I am not expecting this fixing the bug in #2057, because they are in different routine: #2057 fixes the issue in the slow texture transfer, while this fixes the one in the accelerated texture transfer. The common issue of these two is that the game inputs a zero size. As in #2057, the slow routine, it turns out that the zero size probably means a large size, chances are that it also applies here, the accelerated routine.

bogang233:
I just found a way to avoid crash, I don't know well about the complicated process in it~

wwylele:
Sorry for my comments above. I was assuming these assertion failures came from DisplayTransfer, but actually they can also come from MemoryFill. Either way, I still prefer these validity checks go to hw/gpu.cpp, and give some error log. Also we should probably do some hwtest. Just "a way to avoid crash" is going to hide bugs.

I will try some hwtest these days.

wwylele:
Superseded by #2103

PR #2080, opened 2016-09-16
VideoCore: Fix dangling lambda context in shader interpreter

yuriks:
The static meant that after the first execution, these lambda context
would be pointing to a random location on the stack. Fixes a random
crash when using the interpreter.


neobrain:

2016-09-16T05:16:10Z: https://github.com/citra-emu/citra/compare/f196924dddb68f4e47ab6da36552840f82616b90...yuriks:a3afeb4687d0e531a1adc37c504aea89d8d8d71c

lioncash:
Looks good

PR #2082, opened 2016-09-16
Fix/mask crash in shader debugger in Mii Maker

yuriks:
Stopping on a GPU breakpoint at certain points of Mii Maker causes a crash in the vertex shader interpreter as it tries to produce debug info using bogus vertex data. This fixes a bug which happens to stop triggering the crash, though the root causes (the debug info is generated even without valid vertex/shader data and the interpreter doesn't do proper bounds checks of data generated by the shader) remain.

### DIFF
--- a/src/common/vector_math.h
+++ b/src/common/vector_math.h
@@ -60,7 +60,6 @@ public:
     }
 
     Vec2() = default;
-    Vec2(const T a[2]) : x(a[0]), y(a[1]) {}
     Vec2(const T& _x, const T& _y) : x(_x), y(_y) {}
 
     template <typename T2>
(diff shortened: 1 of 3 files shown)


neobrain:

2016-09-16T06:25:46Z: https://github.com/citra-emu/citra/compare/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717...yuriks:8db76cf2e5e5f9cfce5fb2f84e9fac0c167a5775

yuriks:
Rebased.

bunnei:
LGTM

AlmariJosh:
@wwylele DON'T DO THIS TO ME!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

PR #2086, opened 2016-09-17
Add clang-format as part of our {commit,travis}-time checks

linkmauve:
This removes any need for manual formatting, since the tool is always right.

Please review and merge this before any pending PR, as it will conflict with everything else.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2086)


neobrain:

2016-09-17T16:20:16Z: https://github.com/citra-emu/citra/compare/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717...linkmauve:d4d133ec62074d2f6f932ea879abbcd0281cfca9

neobrain:

2016-09-17T16:28:54Z: https://github.com/citra-emu/citra/compare/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717...linkmauve:13e6e05951e65a86230e63f9c9bc431dc86de118

src/.clang-format
```diff
@@ -0,0 +1,89 @@
...
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
+IncludeCategories:
+  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
```
MerryMage: This does not match our current include style.

`  - Regex: '^<(boost|catch|dynarmic|glad|inih|nihstro)/'`

src/.clang-format
```diff
@@ -0,0 +1,89 @@
...
+    Priority:        2
+  - Regex:           '^(<|"(gtest|isl|json)/)'
```
MerryMage: Ditto.

`- Regex: '^<'`

neobrain:

2016-09-17T16:42:46Z: https://github.com/citra-emu/citra/compare/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717...linkmauve:bff6b16c772819c2d4644024535f66f976c6da12

src/common/file_util.cpp
```diff
@@ -9,13 +9,15 @@
 #include "common/logging/log.h"
 
 #ifdef _WIN32
+// clang-format off
```
MerryMage: If you need `<windows.h>` to be above everything else that can be done with a IncludeGroup with an appropriate priority.

neobrain:

2016-09-17T17:36:19Z: https://github.com/citra-emu/citra/compare/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717...linkmauve:b8846cc4b2e0025c3ebd2ec752fb96bd1009333f

neobrain:

2016-09-17T17:47:51Z: https://github.com/citra-emu/citra/compare/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717...linkmauve:83112cf1dbb842211fb73fb61ee7cb144023aff8

linkmauve:
Fixed the include order, thanks merry!

Also switched line length to 80, which is much more sensible than 100 (and let’s use that opportunity to sanitise the coding style).

lioncash:
Review status: 0 of 410 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/audio_core/hle/dsp.h, line 271 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2086#-KRtRG1rbe1cVQrV12s8:-KRtRG1svivDo9kWlKSN:b-irkagx) ([raw file](https://github.com/citra-emu/citra/blob/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717/src/audio_core/hle/dsp.h#L271)):*
> ```objc
>         u32_dsp length;
> 
>         enum class MonoOrStereo : u16_le {
> ```

These should likely not be collapsed onto a single line.

---

*[src/core/arm/dyncom/arm_dyncom_dec.cpp, line 9 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2086#-KRtRRc3Iuvznu1lcidG:-KRtRRc3Iuvznu1lcidH:b-sn4lbl) ([raw file](https://github.com/citra-emu/citra/blob/83112cf1dbb842211fb73fb61ee7cb144023aff8/src/core/arm/dyncom/arm_dyncom_dec.cpp#L9)):*
> ```cpp
> 
> const InstructionSetEncodingItem arm_instruction[] = {
>     {"vmla",
> ```

You'll likely want to disable clang formatting for the decoding tables, same thing with anything applicable in the main interpreter cpp file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2086)*

yuriks:
> Also switched line length to 80, which is much more sensible than 100

Sure is opinions, How about we stick to enforcing the currently accepted coding style instead of inevitably starting arguments about it?

neobrain:

2016-09-18T00:33:05Z: https://github.com/citra-emu/citra/compare/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717...linkmauve:93d99e06e8adfc9876001e5b8ddec710750bec1e

neobrain:

2016-09-18T00:38:28Z: https://github.com/citra-emu/citra/compare/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717...linkmauve:dc8479928c5aee4c6ad6fe4f59006fb604cee701

linkmauve:
Review status: 0 of 389 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/audio_core/hle/dsp.h, line 271 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2086#-KRtRG1rbe1cVQrV12s8:-KRuiplN7ziPRw5ivUnT:b-jjc4q9) ([raw file](https://github.com/citra-emu/citra/blob/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717/src/audio_core/hle/dsp.h#L271)):*
> These should likely not be collapsed onto a single line.
</details>
There is no way to tell that to clang-format, I suggest to ignore it.

---

*[src/core/arm/dyncom/arm_dyncom_dec.cpp, line 9 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2086#-KRtRRc3Iuvznu1lcidG:-KRuijrckYiRdqWZPpDI:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/83112cf1dbb842211fb73fb61ee7cb144023aff8/src/core/arm/dyncom/arm_dyncom_dec.cpp#L9)):*
> You'll likely want to disable clang formatting for the decoding tables, same thing with anything applicable in the main interpreter cpp file.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2086)*

linkmauve:
I also switched back to 100 columns.

lioncash:
*[src/audio_core/hle/dsp.h, line 271 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2086#-KRtRG1rbe1cVQrV12s8:-KRuv1rlTqZ6q993eslI:bmmtopk) ([raw file](https://github.com/citra-emu/citra/blob/8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717/src/audio_core/hle/dsp.h#L271)):*
<details><summary><i>Previously, linkmauve wrote…</i></summary>
> There is no way to tell that to clang-format, I suggest to ignore it.
</details>
Surely there is, otherwise this would have presented itself in Dolphin's codebase.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2086)*

linkmauve:
It actually wouldn’t, since Dolphin’s style includes “BreakBeforeBraces: Allman”, which prevents short enums from being put on a single line.

yuriks:
I've gone and painstakingly looked through the entire diff, fixing places where the missing commas messed up formatting of initializers or enums and also re-flowing comments where I noticed them. (I also slightly tweaked the config file so it gives better results in some cases I noticed.) You can find my commits here: https://github.com/yuriks/citra/tree/clang-format

clang-format still does some utterly bizarre line-wrapping and indenting decisions, but unfortunately I couldn't find any way of changing them. Seems like it's a pretty opinionated tool. I fixed a few of them and the rest are rare enough that they don't bother me that much.

Other than adding on my commits, LGTM.

yuriks:
also, RIP git blame

linkmauve:
Thanks!  I added them to this PR.

MerryMage:
Reviewed 2 of 3 files at r1, 18 of 407 files at r2, 31 of 322 files at r3, 35 of 170 files at r5.
Review status: 86 of 386 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/audio_core/sdl2_sink.cpp, line 13 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2086#-KS2WJtaMU0j8Sv6sVTg:-KS2WJtaMU0j8Sv6sVTh:b18mfr5) ([raw file](https://github.com/citra-emu/citra/blob/396a8d91a4423d9c793eeff0798d544613647511/src/audio_core/sdl2_sink.cpp#L13)):*
> ```cpp
> #include "audio_core/sdl2_sink.h"
> 
> #include <numeric>
> ```

Needs fix.

---

*[src/citra_qt/game_list.cpp, line 44 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2086#-KS2YKgsk0y3NgT_liNd:-KS2YKgtKxaIwCbr8YAA:b-7mqots) ([raw file](https://github.com/citra-emu/citra/blob/396a8d91a4423d9c793eeff0798d544613647511/src/citra_qt/game_list.cpp#L44)):*
> ```cpp
> 
>     // We must register all custom types with the Qt Automoc system so that we are able to use it
>     // with
> ```

minor

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2086)*

MerryMage:
LGTM.

bunnei:
Awesome! LGTM

linkmauve:
I found many (194) other cases of wrong #include order, so I added a new commit removing all newlines inside of those blocks in order to both fix them all and also prevent that kind of issue in the future.

yuriks:
I was only able to get the include order correct locally by using negative priorities:

```
IncludeCategories:
  - Regex:           '^\<[^Q][^/.>]*\>'
    Priority:        -2
  - Regex:           '^\<'
    Priority:        -1
  - Regex:           '^\"'
    Priority:        0
```

These are supported since Clang 3.9. Are we okay to use that? Travis only has 3.4 to begin with (afaict, it's what comes with Trusty) so we'll probably have to install a custom binary of it anyway.

I also simplified the rule to discern stdlib includes from regular library ones which doesn't require us to maintain a list of "library" headers we're using. (As long as no other oddball like Qt shows up that has no file extension and no path separator.)

Updated commits: https://github.com/yuriks/citra/tree/clang-format

yuriks:
To be honest though, I feel like we should just drop the "stdlib includes first" rule. You can't even really discern them when they're not separated by newlines, and it'll simplify things somewhat.

yuriks:
Now also did a bunch of changes to the Travis hook:

 - It would never actually check anything before, because it was checking for files modified in the current commit (which are none since we're not commiting.)
 - Now instead check only files modified in the current PR, or if not doign a PR build (such as our master build) check all files.
 - Use clang-format 4.0 to support the new features being used. (Travis ships with 3.4. 3.9 isn't in an whitelisted PPA.)
 - Only run it on Linux, due to the 4.0 requirement.

PR #2097, opened 2016-09-21
Travis: Fix OS X build

yuriks:
Probably due to additional Travis caching, cmake is now already
installed when the script runs. This causes the unlink to remove the
symlink to the executable, which is then not re-added by the install
(since it's already installed).


yuriks:
:steam_locomotive: :train: :train: :train: :train: :train: :train:

PR #2099, opened 2016-09-22
Fix clang-format lint

wwylele:
Fixed:
 - When clang-format found a problem, `diff` would return non-zero which terminated the build, before even print the lint.
 - If a PR didn't change any source code (like this), `git diff ` would return non-zero which terminated the build.


yuriks:
Retroactive LGTM :)

PR #2100, opened 2016-09-22
Qt: shutdown system if error on loading ROM

### LINKED ISSUE #1965: crash when loading rom on osx el Capitain 10.11.5
when I load a rom I get this error:
Debug <Critical> core/hle/service/fs/archive.cpp:operator():310: Assertion Failed!
Tried to register more than one archive with same id code

can I fix it?

wwylele:
Always shutdown the system whenever `LoadROM` fails. (partially) fixes #1965.


bunnei:
LGTM

yuriks:
How does this work? It seems to me that Shutdown would be called on error before anyway. Which code path didn't do that?

yuriks:
Oh, I see, if the ROM type (and thus loader) weren't recognized that would cause a bad state.

PR #2106, opened 2016-10-02
FS: implement DeleteDirectoryRecursively

wwylele:
Found this unimplemented when writing toys for 3DS file system. So I implemented it, and tested it with my toys against real 3DS.
I plan to clean up all FS error codes in the future, so leave the `// TODO: verify description` for now.

### DIFF
--- a/src/core/file_sys/ivfc_archive.cpp
+++ b/src/core/file_sys/ivfc_archive.cpp
@@ -43,6 +43,12 @@ bool IVFCArchive::DeleteDirectory(const Path& path) const {
     return false;
 }
 
+bool IVFCArchive::DeleteDirectoryRecursively(const Path& path) const {
+    LOG_CRITICAL(Service_FS, "Attempted to delete a directory from an IVFC archive (%s).",
+                 GetName().c_str());
+    return false;
+}
+
 ResultCode IVFCArchive::CreateFile(const Path& path, u64 size) const {
     LOG_CRITICAL(Service_FS, "Attempted to create a file in an IVFC archive (%s).",
                  GetName().c_str());
--- a/src/core/hle/service/fs/fs_user.cpp
+++ b/src/core/hle/service/fs/fs_user.cpp
@@ -235,6 +235,35 @@ static void DeleteDirectory(Service::Interface* self) {
 }
 
 /*
+ * FS_User::DeleteDirectoryRecursively service function
+ *  Inputs:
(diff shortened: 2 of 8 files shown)


src/core/file_sys/ivfc_archive.cpp
```diff
@@ -43,6 +43,12 @@ bool IVFCArchive::DeleteDirectory(const Path& path) const {
     return false;
 }
 
+bool IVFCArchive::DeleteDirectoryRecursively(const Path& path) const {
+    LOG_CRITICAL(Service_FS, "Attempted to delete a directory from an IVFC archive (%s).",
```
Subv: Should these errors really be critical?

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -235,6 +235,35 @@ static void DeleteDirectory(Service::Interface* self) {
...
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    ArchiveHandle archive_handle = MakeArchiveHandle(cmd_buff[2], cmd_buff[3]);
+    auto dirname_type = static_cast<FileSys::LowPathType>(cmd_buff[4]);
+    u32 dirname_size = cmd_buff[5];
+    u32 dirname_ptr = cmd_buff[7];
+
+    FileSys::Path dir_path(dirname_type, dirname_size, dirname_ptr);
+
+    LOG_DEBUG(Service_FS, "type=%d size=%d data=%s", dirname_type, dirname_size,
```
lioncash: `%d` is for signed values, not unsigned. You also need to cast `dirname_type` to `u32` to avoid format warnings.

src/core/file_sys/ivfc_archive.cpp
wwylele: Just copied from `DeleteDirectory` above :stuck_out_tongue:

src/core/hle/service/fs/fs_user.cpp
wwylele: addressed, and cleaned up all similar cases in the file

wwylele:
A record from IRC: this allows an application easily delete the entire host file system using `../../../../something` -like path (already doable by `DeleteFile` and `DeleteDirectory`). I think this is OK to merge, but we need a proper sanitizing as soon as possible.

bunnei:
LGTM

PR #2109, opened 2016-10-04
Proper naming

Lord-Haji:


jroweboy:
Need to update the CMakeFile as well I believe. https://github.com/citra-emu/citra/blob/master/CMakeLists.txt#L261

AlmariJosh:
@wwylele don't limit me and locked me on replying ok?? Just tell me if you don't want to answer that question!!

Lord-Haji:
Chexked the whole repo therr r no other places needed to change

bunnei:
Sorry, why does this need to be changed? Please add descriptions to your PRs.

Lord-Haji:
Ehh?? For proper and convenient naming

wwylele:
@Lord-Haji It is neither proper nor convenient IMO. Although I don't know what the file is exactly (not so familiar with Linux), but this renaming gives me an even worse picture of the file.

It is not convenient, because as I see, the file contains several mime types information that citra uses, instead of encoding of citra or something. The word "encoding" tends to mislead developers and users.

It is not proper, for other similar files I found in my PC are not named as `xxx-encoding.xml` format, but most of them are bare application name or type name. I don't know if there is a rule or convention for these, but this renaming doesn't seem to fit in other files I've seen.

JayFoxRox:
If there is a convention we should follow please add it to this discussion.
I'll close this now. There is no point having 6 participants discussing a +1 -1 line change.

PR #2115, opened 2016-10-06
rename LISENCE to uppercase

Lord-Haji:
rename license.txt to LICENSE  uppercase and remove extension as normally done....

Added GNU public license template v.3.0


MerryMage:
I don't see a reason to use the v3 text, given the codebase is v2+ licensed.

yuriks:
Yeah, changing the license to GPL 3 is incorrect.

PR #2120, opened 2016-10-09
Search bar for GPU Commands

RiokuTheSlayer:
Search bar for the Pica Command List. Just adds a text area that updates the GPU command list with the input from the user. Case-sensitive. Needs optimized, pretty slow, but there's a lot of commands so that may be unavoidable. Spacing should be fixed now, I un-tabbed all my changes.
(As a side note I moved the search bar up above the start/finish/copy trace buttons. Screenshot isn't updated.)

Screenshot of the change in Citra:
screenshot 28

### DIFF
--- a/src/video_core/debug_utils/debug_utils.h
+++ b/src/video_core/debug_utils/debug_utils.h
@@ -20,282 +20,282 @@
 #include "video_core/pica.h"
 
 namespace CiTrace {
-    class Recorder;
+class Recorder;
 }
 
 namespace Pica {
 
-    namespace Shader {
-        struct ShaderSetup;
-    }
+namespace Shader {
+struct ShaderSetup;
+}
 
-    class DebugContext {
-    public:
-        enum class Event {
-            FirstEvent = 0,
+class DebugContext {
+public:
+    enum class Event {
+        FirstEvent = 0,
 
-            PicaCommandLoaded = FirstEvent,
-            PicaCommandProcessed,
-            IncomingPrimitiveBatch,
-            FinishedPrimitiveBatch,
-            VertexShaderInvocation,
-            IncomingDisplayTransfer,
-            GSPCommandProcessed,
-            BufferSwapped,
+        PicaCommandLoaded = FirstEvent,
+        PicaCommandProcessed,
+        IncomingPrimitiveBatch,
+        FinishedPrimitiveBatch,
+        VertexShaderInvocation,
+        IncomingDisplayTransfer,
+        GSPCommandProcessed,
+        BufferSwapped,
 
-            NumEvents
-        };
-
-        /**
-        * Inherit from this class to be notified of events registered to some debug context.
-        * Most importantly this is used for our debugger GUI.
-        *
-        * To implement event handling, override the OnPicaBreakPointHit and OnPicaResume methods.
-        * @warning All BreakPointObservers need to be on the same thread to guarantee thread-safe state
-        * access
-        * @todo Evaluate an alternative interface, in which there is only one managing observer and
-        * multiple child observers running (by design) on the same thread.
-        */
-        class BreakPointObserver {
-        public:
-            /// Constructs the object such that it observes events of the given DebugContext.
-            BreakPointObserver(std::shared_ptr<DebugContext> debug_context)
-                : context_weak(debug_context) {
-                std::unique_lock<std::mutex> lock(debug_context->breakpoint_mutex);
-                debug_context->breakpoint_observers.push_back(this);
-            }
-
-            virtual ~BreakPointObserver() {
-                auto context = context_weak.lock();
-                if (context) {
-                    std::unique_lock<std::mutex> lock(context->breakpoint_mutex);
-                    context->breakpoint_observers.remove(this);
-
-                    // If we are the last observer to be destroyed, tell the debugger context that
-                    // it is free to continue. In particular, this is required for a proper Citra
-                    // shutdown, when the emulation thread is waiting at a breakpoint.
-                    if (context->breakpoint_observers.empty())
-                        context->Resume();
-                }
-            }
-
-            /**
-            * Action to perform when a breakpoint was reached.
-            * @param event Type of event which triggered the breakpoint
-            * @param data Optional data pointer (if unused, this is a nullptr)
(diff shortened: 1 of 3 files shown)


neobrain:
Indentation is completely messed up with these changes, both violating our coding style and furthermore making the diff view useless. Please fix this up so that we can do a proper review!

neobrain:
Your most recent patch notes say "Spacing/indentation should be 100% valid now". In reply to that statement, would you mind checking the `git diff --color=auto` output before pushing changes?

src/video_core/debug_utils/debug_utils.h
```diff
@@ -47,15 +47,15 @@ class DebugContext {
...
-     */
+    * Inherit from this class to be notified of events registered to some debug context.
+    * Most importantly this is used for our debugger GUI.
+    *
+    * To implement event handling, override the OnPicaBreakPointHit and OnPicaResume methods.
+    * @warning All BreakPointObservers need to be on the same thread to guarantee thread-safe state
+    * access
+    * @todo Evaluate an alternative interface, in which there is only one managing observer and
+    * multiple child observers running (by design) on the same thread.
+    */
```
MerryMage: Unnecessary whitespace diff.

src/video_core/debug_utils/debug_utils.h
```diff
@@ -80,50 +80,50 @@ class DebugContext {
...
-         * Action to perform when a breakpoint was reached.
-         * @param event Type of event which triggered the breakpoint
-         * @param data Optional data pointer (if unused, this is a nullptr)
-         * @note This function will perform nothing unless it is overridden in the child class.
-         */
+        * Action to perform when a breakpoint was reached.
+        * @param event Type of event which triggered the breakpoint
+        * @param data Optional data pointer (if unused, this is a nullptr)
+        * @note This function will perform nothing unless it is overridden in the child class.
+        */
```
MerryMage: Unnecessary whitespace diff.

src/video_core/debug_utils/debug_utils.h
```diff
@@ -80,50 +80,50 @@ class DebugContext {
...
         virtual void OnPicaBreakPointHit(Event, void*) {}
 
         /**
-         * Action to perform when emulation is resumed from a breakpoint.
-         * @note This function will perform nothing unless it is overridden in the child class.
-         */
+        * Action to perform when emulation is resumed from a breakpoint.
+        * @note This function will perform nothing unless it is overridden in the child class.
+        */
```
MerryMage: Unnecessary whitespace diff; ditto for all doc comments in this file.

src/video_core/debug_utils/debug_utils.h
```diff
@@ -267,4 +298,4 @@ class MemoryAccessTracker {
 
 } // namespace
 
-} // namespace
+} // namespace
```
MerryMage: New line required at end of file.

RiokuTheSlayer:
That's so weird because I'm using clang-format to do everything, the latest Visual Studio version. So... I'll go remove them and stop using the plugin I guess

src/video_core/debug_utils/debug_utils.h
```diff
@@ -194,6 +194,37 @@ struct PicaTrace {
...
+            is_search_corrected = true;
+
+        for (int i = 0; i < writes.size(); i++) {
+            bool is_valid = true;
+            std::string cmd_name = Pica::Regs::GetCommandName(writes[i].cmd_id);
+
+            if (cmd_name.size() < search_string.size())
+                continue;
+
+            for (int j = 0; j < search_string.size(); j++) {
```
MerryMage: Consider `std::equal(cmd_name.begin(), cmd_name.end(), search_string.begin());` instead of this `for` loop.

src/video_core/debug_utils/debug_utils.h
```diff
@@ -194,6 +194,37 @@ struct PicaTrace {
...
+                    is_valid = false;
+                    break;
+                }
+            } // End of j for-loop
+
+            if (!is_valid)
+                continue;
+
+            search_corrected_writes.push_back(writes[i]);
+        } // End of i for-loop
```
MerryMage: Consider

    auto matches_search_string = [&](const Write& write) -> bool {
        std::string cmd_name = Pica::Regs::GetCommandName(write.cmd_id);
        return cmd_name.size() < search_string.size() &&
            std::equal(cmd_name.begin(), cmd_name.end(), search_string.begin());
    }
    std::copy_if(writes.begin(), write.end(), std::back_inserter(search_corrected_writes), matches_search_string);

or equivalent instead of this `for` loop.

src/video_core/debug_utils/debug_utils.h
neobrain: Even with @MerryMage's suggestion, this only looks for prefix matches of the entered string. For a search function, you'd expect any infixes to match.

Alternative suggestion for replacing the entire `for (int i = 0; i < writes.size(); ++i)` loop:
```
auto&& is_matching_write = [&search_string](auto&& write) {
    auto cmd_name = Pica::Regs::GetCommandName(write.cmd_id);
    return cmd_name.end() != boost::search(cmd_name, search_string);
};

boost::copy(writes | boost::adaptors::filtered(is_matching_write),
            std::back_inserter(search_corrected_writes));
```

src/video_core/debug_utils/debug_utils.h
```diff
@@ -194,6 +194,37 @@ struct PicaTrace {
         u32 value;
     };
     std::vector<Write> writes;
+    std::vector<Write> search_corrected_writes;
+    bool is_search_corrected = false;
+
+    void CorrectSearchWrites(std::string search_string) {
+        search_corrected_writes.clear();
+
+        if (search_string.size() == 0)
```
neobrain: `is_search_corrected = (search_string.size() > 0);`

wwylele:
What's the status of this PR?

yuriks:
I like the idea, but instead of outright filtering (which removes the context from the things you were searching for) how about making it highlight entries which match the string? That way you can easily find them in the list, but you can still see what commands are used together. Then pressing Enter in the textbox could can bring you to the next result.

Subv:

Issue #1845, opened 2016-05-23
Rewrite the HLE SOC module


Subv:
Currently our HLE SOC module has platform dependent code in it, and is overall more complicated than it should be. In my opinion this should all be abstracted away into some sort of Network handler.

Converting over to a more general approach should help out with the eventual implementation of Download Play and possibly Internet Play.

darkf:
Can a library like `boost::asio` be used for this or do we really want to invent yet-another-socket-abstraction-library?

archshift:
I don't even think it has to be such a kitchen-sink library as `boost::asio`. I'm investigating using [mongoose](https://docs.cesanta.com/mongoose/latest/) for an implementation of the `http:C` service, and I think it's likely flexible enough to use for `soc:U`.

JayFoxRox:
Is this still planned?
Did anyone work on it yet?
Does anybody volunteer to continue this? (research + implementation)

---

I'm asking because I consider taking over #1696 as it should be "good enough" with some tiny changes.
However, if it's going to be replaced soon anyway I'll just close it.
(Wether I'll actually fix it or simply close it also depends on my personal motivation)

PR #2123, opened 2016-10-10
Fix build on DragonFly and FreeBSD

jbeich:
**EDIT**: updated links

Part of my [emulators/citra](https://www.freshports.org/emulators/citra) port. Many bugs are Dolphin's legacy. I've cherry-picked dolphin-emu/dolphin@5bab6210ef26 to cleanup my own `#ifndef __GLIBC__` ugliness. Not sure about i386 due to MerryMage/dynarmic#35 but amd64 starts fine. Package build logs:
- FreeBSD 9.3 amd64, GCC 5.4: [SDL2](http://beefy2.nyi.freebsd.org/data/latest-per-pkg/citra/), [Qt5](http://beefy2.nyi.freebsd.org/data/latest-per-pkg/citra-qt5/)
- [**frequent**] FreeBSD 10.1 amd64, Clang 3.6: [SDL2](http://beefy6.nyi.freebsd.org/data/latest-per-pkg/citra/), [Qt5](http://beefy6.nyi.freebsd.org/data/latest-per-pkg/citra-qt5/)
- [**frequent**] FreeBSD 11.0 amd64, Clang 3.8: [SDL2](http://beefy9.nyi.freebsd.org/data/latest-per-pkg/citra/), [Qt5](http://beefy9.nyi.freebsd.org/data/latest-per-pkg/citra-qt5/)
- FreeBSD 12.0 amd64 (trunk), Clang 3.8/[3.9](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=212343): [SDL2](http://beefy4.nyi.freebsd.org/data/latest-per-pkg/citra/), [Qt5](http://beefy4.nyi.freebsd.org/data/latest-per-pkg/citra-qt5/)
- DragonFly 4.7 x86_64 (trunk), GCC 5.4: SDL2 ([poudriere](http://muscles.dragonflybsd.org/bulk/latest-per-pkg/citra/), [synth](http://muscles.dragonflybsd.org/synth/snap-failures/emulators___citra.log)), Qt5 ([poudriere](http://muscles.dragonflybsd.org/bulk/latest-per-pkg/citra-qt5/), [synth](http://muscles.dragonflybsd.org/synth/snap-failures/emulators___citra-qt5.log))

@bentley, can you check OpenBSD bits?


linkmauve:
Please (also) upstream the microprofile changes to https://bitbucket.org/jonasmeyer/microprofile

jroweboy:
Citra doesn't support 32 bit; is that a requirement to support 32 bit when making a port in freshports? If you can just add a 64bit version only, thats what I would suggest. In the past, citra would compile and function on 32 bit, but there was never any official support for it. It just happened to compile. And now as you see with the addition of dynarmic, citra no longer compiles on 32 bit as the JIT is very much 64bit only right now.

I'm also opposed to adding a flag to disable the JIT entirely as that would lead to more complicated code surrounding anything with the JIT. Just some things that would need to be done to add a DISABLE_DYNARMIC option to cmake that I thought of off the top of my head, you'd need to disable the checkbox in qt, add some ifdefs in the arm_dynarmic interface to make the interface still compile without dynarmic, more ifdefs on settings load to make sure it doesn't attempt to load the JIT even if the setting is set to JIT. Then any future code dealing with the the JIT will need to watch out for all three cases, JIT isn't in the build, JIT is disabled, JIT is enabled. Overall, it just sounds more troublesome to support compilation without dynarmic just to support 32 bit platforms that citra never officially supported before.

bentley:
@jbeich, I don’t see anything in the patches that would harm OpenBSD. `src/common/swap.h` is not really right, but it’s no worse than what’s already there.

Actually, unless things have changed since the last time I looked, Citra properly checks host endianness before any use of the byteswap functions, meaning the entire header can pretty much go away, to be replaced with POSIX’s `htole64`, etc., or an equivalent. But I’ll put that in another pull request (sometime) rather than complicating this one.

Disabling the JIT is probably not very useful due to performance reasons. However, it would be nice to retool the JIT to not require writable, executable memory. This is a common attack vector, and as a result the screws are tightening in [Windows 10](https://github.com/hrydgard/ppsspp/issues/7854), [iOS](https://www.youtube.com/watch?v=BLGFriOKz6U), and [OpenBSD](http://undeadly.org/cgi?action=article&sid=20160527203200). [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1215479), for example, no longer requires writeable executable memory on iOS or OpenBSD. (This also probably deserves its own issue rather than cluttering this pull request.)

jbeich:
Rebased, see [interdiff](http://pastebin.com/F6C8angs):
- fixed build on Linux (including clang-format)
- allow dogfooding clang-format on BSDs per ff6f4ad
- updated 5905ba9 to a version proposed upstream
- dropped 36abd00 for consitency with Windows and OS X
- split 136cbbe into def4dab and 285846a for clarity

> Please (also) upstream the microprofile changes to https://bitbucket.org/jonasmeyer/microprofile

https://bitbucket.org/jonasmeyer/microprofile/pull-requests/10/

*Sigh* Bitbucket embeds 1x1 GIF in HTML mail to confirm signup then spews "Unknown error" for Tor users.

> Citra doesn't support 32 bit

Is Citra x86_64-only or would it work on other 64bit architectures? I'm thinking about aarch64 for we have a [buildbot](http://thunderx1.nyi.freebsd.org/). For example, the buildbot managed to build all dependencies for PPSSPP but failed on the port itself due to lack of upstream build glue.

> is that a requirement to support 32 bit when making a port in freshports?

No, see existing examples: [xen](https://www.freshports.org/emulators/xen) (amd64-only), [xf86-video-sunffb](https://www.freshports.org/x11-drivers/xf86-video-sunffb) (sparc64-only), [raspberrypi-userland](https://www.freshports.org/misc/raspberrypi-userland) (armv6-only).

FreshPorts is just a nice (a bit dated) UI for the official FreeBSD ports tree. It has already landed there and being built as 2 packages: sdl2 (main), qt5.

> I'm also opposed to adding a flag to disable the JIT entirely as that would lead to more complicated code surrounding anything with the JIT.

Why not add unoptimized path into JIT itself (i.e. dynarmic)? PPSSPP did something similar with FakeJit. It's probably only useful for catching endianess/alignment bugs.

JayFoxRox:
This JIT discussion should probably happen elsewhere so anybody can see it. Under any circumstance it should happen in a seperate PR or issue specifically for JIT. These meta-PRs which change a ton of different things are hard to find in the history and you might catch other devs off-guard who might have valuable input, but don't care about DragonFly and/or/on FreeBSD.
My input on the CPU JIT thing is that it's just another CPU interface [which is already abstracted, afaik], so I think it would be wise to force non-jit platforms into using the interpreter [and disabling the JIT option, possibly adding a warning that Citra is not optimized for the running platform].
(The same applies to the shader JIT [which is definitely abstracted already])

jbeich:
> Citra doesn't support 32 bit

After aarch64 failed ([SDL2](http://thunderx1.nyi.freebsd.org/data/110arm64-default/424296/logs/errors/citra-s20161007_1.log), [Qt5](http://thunderx1.nyi.freebsd.org/data/110arm64-default/424296/logs/errors/citra-qt5-s20161007_1.log)) like i386, I've explicitly [disabled](https://github.com/freebsd/freebsd-ports/commit/b8f97fcb46bc976befa8b473fbdb85b91c9b22f7) package builds for powerpc64, sparc64, mips64 as well. It seems Citra doesn't support any 64bit platform beyond x86 family.

JayFoxRox:
Yes, the JITs are currently only able to generate x64 code.
You could theoretically disable the JITs for non x64 as we have interpreters as fallbacks. Seperate PRs welcome. We still won't officially create 32 bit builds.

jbeich:
Whitespace (indentation) fix for CMakeLists.txt. Mainly to trigger Travis (Linux) re-try per
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

MerryMage:
Overall looks fine to me. 

Tagging #1845 regarding the edits to socket code.

jroweboy:
Is there any more discussion to be had about this pull request? It looks like its ready to be merged to me.

PR #2124, opened 2016-10-11
[HLE/Kernel] Implement Event signaling delay

mailwl:
This PR implements Event signaling delay, mostly copy/pasted from Timer.

as example - delay for Y2R completion event, lets FIFA games shows intro video.

this will be useful for IR and CAM events, too

### DIFF
--- a/src/core/hle/kernel/event.cpp
+++ b/src/core/hle/kernel/event.cpp
@@ -3,14 +3,22 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
+#include <cinttypes>
 #include <map>
 #include <vector>
 #include "common/assert.h"
+#include "core/core_timing.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/thread.h"
 
 namespace Kernel {
+
+/// The event type of the generic event callback
+static int event_callback_event_type;
+// TODO(yuriks): This can be removed if Event objects are explicitly pooled in the future, allowing
+//               us to simply use a pool index or similar.
+static Kernel::HandleTable event_callback_handle_table;
 
 Event::Event() {}
 Event::~Event() {}
@@ -21,6 +29,7 @@ SharedPtr<Event> Event::Create(ResetType reset_type, std::string name) {
     evt->signaled = false;
     evt->reset_type = reset_type;
     evt->name = std::move(name);
+    evt->callback_handle = event_callback_handle_table.Create(evt).MoveFrom();
 
     return evt;
 }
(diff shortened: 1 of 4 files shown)


lioncash:
> mostly copy/pasted 

It's honestly incredible when someone states this as if it's an OK thing to do. Do you at least have any testing done to verify the delay in some form at least? `1000` is also used as a magic number. At least put it in a named constant.

wwylele:
I don't like this design at all. The class `Kernel::Event` is a mimic of real 3DS kernel object class `KEvent`, which doesn't have such "delay" function. Not to mention that we want the kernel as simple as possible, given that our scheduler is already a mess. If you want to emulate a delay, it shouldn't be intrusive to the kernel level like this.

Take Y2R event delay for example, I prefer these solution:

 - Y2R service manages a `CoreTiming::Event` itself, and schedule it when it wants to signal the event with a delay. This is still like a hack, but at least it keeps clean to the kernel.

 - based on this, Y2R can put its work in a separate thread and finish it before the event signal (or the main thread will wait for it if it doesn't finish the work on time). Thus we make it behave more like real console. IIRC @yuriks suggested that, with this idea, we can develop a "deadline" system to emulate all such asynchronous system task.

mailwl:
> Y2R can put its work in a separate thread and finish it before the event signal 

do you mean host thread? i don't think this will work

wwylele:
@mailwl yes, I mean host thread. A quick implementation:
```
static conversion_event_callback_type;
static void StartConversion(Service::Interface* self) {
    ...
    std::thread* task = new thread(HW::Y2R::PerformConversion, conversion);
    CoreTiming::ScheduleEvent(100, conversion_event_callback_type, (u64)task)
}

static void ConversionEventCallback(u64 user_data, int cycle_late) {
    std::thread* task = (std::thread*)user_data;
    task->join();
    completion_event->Signal();
} 

Interface::Interface() {
    ...
    conversion_event_callback_type = CoreTiming::RegisterEvent("ConversionEventCallback", ConversionEventCallback);
}
```

mailwl:
i've got the idea, so, each service with async functions must do it separately, so no common solution needed.
i close the PR

wwylele:
@mailwl Just to clarify: there can be a common solution or framework. My point is that it shouldn't be in the kernel.

mailwl:
@wwylele let me form your code in PR?
one of Nintendo games with animals will be playable, none tested by me other games not broken, it's just Y2R so capricious ATM

wwylele:
@mailwl sure, but you probably need a hwtest to get a proper delay first . An arbitrary value like 100 or 1000 is not acceptable to me

mailwl:
hm.. at my sight, exact time of video decoding differ for every frame of video, and actually, it is not gameplay (at least, i skip CG on every game i played)

Issue #2041, opened 2016-09-02
Pokemon Omega Ruby & Alpha Sapphire Issues


ybha4:
Hey guys, now that CRO has been merged, Pokemon is now officially supported by Citra (_hears sighs of developers_).

Seriously though, there are a couple of issues with Pokemon:

1) Frequently in the game, there are a lot of black lines that appear and possibly, some depth issues?

2) The game crashes frequently. Is this still missing areas within CRO implementation? Or is this a separate issue within Citra? Also, the crashes don't appear to happen at consistent points in the game. ie sometimes it will crash in the beginning. Once replayed, it will continue passed that point and might crash at another random point in the game. The crashing appears to be completely random.

To replicate, open Pokemon OR or AS and play the first 5 minutes of the game. You will see what I'm talking about.

Specs: 
Windows 7
i7 (3.4ghz)
8gb RAM
Nvidia

Note: This is different to #2026 and #2028 as those crashes occur at specific points in the game. Mine are random it seems.
![pokemon issue 1](https://cloud.githubusercontent.com/assets/21698970/18191265/718f9702-710c-11e6-8dd3-b4916934cd4c.png)

Enjoy :)

yuriks:
The issue with the wrong outlines seems to be caused by improper management of the stencil buffer in the texture cache: it'll sometimes not be cleared even though the depth buffer.

That issue and possibly other crashes might be caused by #2053.

wwylele:
I did a test about crash/freeze, and get an interesting result.

I tested Pokemon OR (JAP version, language Japanese), with a save from others (who complained to me that the save was "cursed" and freeze frequently). I am on the JIT branch, though, but I believe it doesn't matter. 

The test is: start the game, walk into grass, start a battle and finish it without any crash/freeze. This test was performed several times.

To find the cause of the freeze, I change the code for a half among all samples. The change is: stub [the result code of NWM_UDS::Initialize](https://github.com/citra-emu/citra/blob/master/src/core/hle/service/nwm_uds.cpp#L100) with 0xC9411002 instead of SUCCESS. (the error code is get from hw, which I believe to be "network disabled")

And the result of all samples:


|Code changed?|Total samples|Failed|Pass|
|---|---|---|---|
|No|10|9|1|
|Yes|10|0|10|

:smile: I guess we get a direction to work on.

ybha4:
[...]

PR #2126, opened 2016-10-12
NWM: stub Initialize with an error

wwylele:
Just make my comment in #2041 into a PR. I did more research after that comment but didn't get more useful result. Since the network isn't going to be implemented any time soon, it is better to prevent games from using it.

Here is [the hwtest to get the error code](https://github.com/wwylele/ctrhwtest/tree/master/network-init) (turn off wifi and run)

### DIFF
--- a/src/core/hle/service/nwm_uds.cpp
+++ b/src/core/hle/service/nwm_uds.cpp
@@ -97,10 +97,21 @@ static void Initialize(Service::Interface* self) {
     u32 value = cmd_buff[13];
     u32 handle = cmd_buff[14];
 
+    // Because NWM service is not implemented at all, we stub the Initialize function with an error
+    // code instead of success to prevent games from using the service and from causing more issues.
+    // The error code is from a real 3DS with wifi off, thus believed to be "network disabled".
+    /*
     cmd_buff[1] = RESULT_SUCCESS.raw;
     cmd_buff[2] = 0;
     cmd_buff[3] = Kernel::g_handle_table.Create(handle_event)
                       .MoveFrom(); // TODO(purpasmart): Verify if this is a event handle
+    */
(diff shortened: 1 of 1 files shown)


src/core/hle/service/nwm_uds.cpp
```diff
@@ -97,10 +97,21 @@ static void Initialize(Service::Interface* self) {
     u32 value = cmd_buff[13];
     u32 handle = cmd_buff[14];
 
+    // Because NWM service is not implemented at all, we stub the Initialize function with an error
+    // code instead of success, to prevent games from using the service and from causing more issue.
```
lioncash: the comma isn't necessary, and `issue` should be `issues`

src/core/hle/service/nwm_uds.cpp
```diff
@@ -97,10 +97,21 @@ static void Initialize(Service::Interface* self) {
...
+    // The error code is from real 3DS with wifi off, thus believed to be "network disabled".
```
lioncash: `The error code is from a ...`

ybha4:
Is this PR supposed to fix the crashes in the Pokemon games found in #2041 ?

I tried the master with this PR merged and the games are still crashing at the same level of frequency.

wwylele:
@ybha4 this should reduce some freeze, but not all. Aparently there are other causes of freeze.

src/core/hle/service/nwm_uds.cpp
wwylele: addressed

src/core/hle/service/nwm_uds.cpp
wwylele: addressed

bunnei:
LGTM

PR #2132, opened 2016-10-22
Correct FS error codes & add path boundary checks

### LINKED ISSUE #1607: Fantasy Life - Failed to detect SD card
fantasy

[LOG](http://pasted.co/4254db47)

### LINKED ISSUE #2094: Incorrect error code when trying to create an existing directory
Reported by gdmk:

> When the directory exists, `CreateDirectory` in `fs_user.cpp` should return 0xC82044BE instead of zero. Fixed _スライムもりもりドラゴンクエスト３ 大海賊としっぽ団_ 

We should check our `CreateX` functions (and even other fs functions) and do more hwtests on them to get proper error codes.

wwylele:
The main part of this PR is to fix error codes of archive operations. Because error codes can vary among different error types and archive types, I had to refactor the `ArchiveBackend` interface and their subclass: let all method report its own error code, and split `DiskArchive` into 3 classes `SaveDataArchive`, `ExtSaveDataArchive` and `SDMCArchive`. Sorry for not following the suggestion in the code comment "This should be subclassed by concrete archive types", but if, for example, I let `SaveDataArchive` be a subclass off `DiskArchive`, I will literally override everything due to different error check behaviour, which is effectively the same as making a brand new class.

The error codes were got from [this test program](https://github.com/wwylele/3DS-FS-Tester) ([compiled version](https://github.com/wwylele/3DS-FS-Tester/releases/download/1.0/wwylele_fs_tester.cxi)) and here is [a summary sheet](https://drive.google.com/file/d/0ByjS9HUStkSXUmtPdWdmd25CeGc/view?usp=drivesdk). However, with this PR, citra and real 3DS still don't behave exactly the same for some very edge cases that are unlikely to happen, such as `OpenFile("/")`. These cases are marked red in the sheet. Implementing those edge cases will make the code too complicated but doesn't contribute much to the compatibility, and I feel like they can be even indeterminate and are "undefined behaviour" on real 3DS. If anyone wants citra to be more accurate, they can always add more error checks in the future.

Note that I didn't test `RenameFile` and `RenameDirectory`, and I keep the old error code. I want to leave this to the next PR, because they are another large topic.

You can also run the mentioned test program on citra with this PR to test. However, the program uses software keyboard to input commands, so you may want to merge [this commit](https://github.com/wwylele/citra/commit/0baf9f8fb6f0cd160d735c59d44a43d005539e2d) so that you can input commands from stdin.

Fixes #2094, and "path boundary check" comments in #2106.

Additional commits:
- hwtested and implemented SDMC write-only archive. Should fixes #1607.
- rename `SaveDataCheck` archive to `NCCH` archive, according to [3dbrew](https://www.3dbrew.org/wiki/Filesystem_services#Archives) and its behaviour.

This PR is not widely test yet. Please help me test games to see if there is any regression! Please especially pay attention when game want to create a new save or delete the save.


mailwl:
i see 5bf45ce, it will fix Fantasy Life - Failed to detect SD card #1607 ?

wwylele:
@mailwl yup, I just updated the description

wwylele:
Note that I added a unit test for `PathParser`. I am not sure if I am allowed to create temporary files like that. Also, the OS X build seems failing on building the unit test. I don't know whether I did something wrong or there is a issue in our test framework.

src/core/file_sys/archive_extsavedata.cpp
```diff
@@ -11,13 +11,139 @@
...
+                            const u8* buffer) const override {
+        if (offset > size) {
+            return ResultCode(ErrorDescription::FS_WriteBeyondEnd, ErrorModule::FS,
+                              ErrorSummary::InvalidArgument, ErrorLevel::Usage);
+        } else if (offset == size) {
+            return MakeResult<size_t>(0);
+        }
+
+        if (offset + length > size)
+            length = size - offset;
```
bunnei: nit: Prefer using { and } even for single-line if statements

src/core/file_sys/archive_extsavedata.cpp
```diff
@@ -11,13 +11,139 @@
...
+    ErrorDescription::FS_UnexpectedFileOrDirectory, ErrorModule::FS, ErrorSummary::NotSupported,
+    ErrorLevel::Usage);
+
+/**
+ * A modified version of DiskFile for fix-size file used by ExtSaveData
+ * The file size can't be changed by SetSize or Write.
+ */
+class FixSizeDiskFile : public DiskFile {
+public:
+    FixSizeDiskFile(FileUtil::IOFile&& file, const Mode mode) : DiskFile(std::move(file), mode) {
```
bunnei: nit: It's somewhat pointless to have `mode` be `const`

bunnei: Applies here and elsewhere... I won't bother pointing out all of the occurances

bunnei: ...Actually, I thought Mode was an enum, but it's actually a union. Perhaps this should be a const reference instead?

src/core/file_sys/archive_extsavedata.cpp
```diff
@@ -11,13 +11,139 @@
...
+            LOG_ERROR(Service_FS, "Create flag is not supported");
+            return ERROR_UNSUPPORTED_OPEN_FLAGS;
+        }
+
+        auto full_path = path_parser.BuildHostPath(mount_point);
+
+        switch (path_parser.GetHostStatus(mount_point)) {
+        case PathParser::InvalidMountPoint:
+            LOG_CRITICAL(Service_FS, "Invalid mount point %s", mount_point.c_str());
+            return ERROR_FILE_NOT_FOUND;
```
bunnei: This is not a different error? Are these verified?

src/core/file_sys/archive_sdmc.cpp
```diff
@@ -8,13 +8,307 @@
...
 #include "core/file_sys/disk_archive.h"
+#include "core/file_sys/path_parser.h"
 #include "core/settings.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // FileSys namespace
 
 namespace FileSys {
 
+static const ResultCode ERROR_INVALID_PATH(ErrorDescription::FS_InvalidPath, ErrorModule::FS,
```
bunnei: Should these be duplicated?

src/core/file_sys/archive_sdmc.cpp
```diff
@@ -8,13 +8,307 @@
...
+        LOG_ERROR(Service_FS, "Empty open mode");
+        return ERROR_INVALID_OPEN_FLAGS;
+    }
+
+    if (mode.create_flag && !mode.write_flag) {
+        LOG_ERROR(Service_FS, "Create flag set but write flag not set");
+        return ERROR_INVALID_OPEN_FLAGS;
+    }
+
+    auto full_path = path_parser.BuildHostPath(mount_point);
```
bunnei: This can be `const`

src/core/file_sys/archive_sdmc.cpp
```diff
@@ -8,13 +8,307 @@
...
+    modified_mode.read_flag.Assign(1);
+
+    return OpenFileBase(path, modified_mode);
+}
+
+ResultVal<std::unique_ptr<FileBackend>> SDMCArchive::OpenFileBase(const Path& path,
+                                                                  const Mode mode) const {
+    LOG_DEBUG(Service_FS, "called path=%s mode=%01X", path.DebugStr().c_str(), mode.hex);
+
+    PathParser path_parser(path);
```
bunnei: This can be `const`

src/core/file_sys/archive_sdmc.cpp
```diff
@@ -8,13 +8,307 @@
...
+                      full_path.c_str());
+            return ERROR_NOT_FOUND;
+        } else {
+            // Create the file
+            FileUtil::CreateEmptyFile(full_path);
+        }
+        break;
+    }
+
+    std::string mode_string = "";
```
bunnei: This is already initialized as an empty string

src/core/file_sys/archive_sdmc.cpp
```diff
@@ -8,13 +8,307 @@
...
+    if (mode.write_flag)
+        mode_string += "r+";
```
bunnei: `mode_string = "r+";`

src/core/file_sys/archive_sdmc.cpp
```diff
@@ -8,13 +8,307 @@
...
+            return ERROR_NOT_FOUND;
+        } else {
+            // Create the file
+            FileUtil::CreateEmptyFile(full_path);
+        }
+        break;
+    }
+
+    std::string mode_string = "";
+    if (mode.write_flag)
```
bunnei: replace with:

std::string mode_string{ mode.write_flag ? "r+" : "r" };

src/core/file_sys/archive_sdmc.cpp
```diff
@@ -8,13 +8,307 @@
...
+        LOG_CRITICAL(Service_FS, "Unknown error opening %s", full_path.c_str());
+        return ERROR_NOT_FOUND;
+    }
+
+    auto disk_file = std::make_unique<DiskFile>(std::move(file), mode);
+    return MakeResult<std::unique_ptr<FileBackend>>(std::move(disk_file));
+}
+
+ResultCode SDMCArchive::DeleteFile(const Path& path) const {
+    PathParser path_parser(path);
```
bunnei: Can be `const`

src/core/hle/service/ptm/ptm.cpp
```diff
@@ -128,7 +128,7 @@ void Init() {
             Service::FS::OpenArchive(Service::FS::ArchiveIdCode::SharedExtSaveData, archive_path);
         ASSERT_MSG(archive_result.Succeeded(), "Could not open the PTM SharedExtSaveData archive!");
 
-        FileSys::Path gamecoin_path("gamecoin.dat");
+        FileSys::Path gamecoin_path("/gamecoin.dat");
```
bunnei: This seems crappy to have to do. Should we update `FileSys::Path` to handle this properly? (and perhaps also use the platform correct path separator?)

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -504,7 +504,7 @@ ResultCode LoadConfigNANDSaveFile() {
 
     cfg_system_save_data_archive = *archive_result;
 
-    FileSys::Path config_path("config");
+    FileSys::Path config_path("/config");
```
bunnei: Ditto (see below comment)

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -382,7 +382,7 @@ ResultCode UpdateConfigNANDSavegame() {
 ResultCode FormatConfig() {
     ResultCode res = DeleteConfigNANDSaveFile();
     // The delete command fails if the file doesn't exist, so we have to check that too
-    if (!res.IsSuccess() && res.description != ErrorDescription::FS_NotFound)
+    if (!res.IsSuccess() && res.description != ErrorDescription::FS_FileNotFound)
         return res;
```
bunnei: nit: Prefer using { and } even for single-line if statements (I know this isn't your code, but just as a future reference when changing code that already exists)

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -368,7 +368,7 @@ ResultCode UpdateConfigNANDSavegame() {
     mode.write_flag.Assign(1);
     mode.create_flag.Assign(1);
 
-    FileSys::Path path("config");
+    FileSys::Path path("/config");
```
bunnei: Ditto (see comment below)

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -359,7 +359,7 @@ ResultCode CreateConfigInfoBlk(u32 block_id, u16 size, u16 flags, const void* da
 }
 
 ResultCode DeleteConfigNANDSaveFile() {
-    FileSys::Path path("config");
+    FileSys::Path path("/config");
```
bunnei: Ditto (see comment below) ... Will stop pointing these out

src/core/hle/service/ptm/ptm.cpp
wwylele: The slash is for the guest file system path, not for host one. On real 3DS the beginning slash is mandatory and must should be "/"

src/core/file_sys/savedata_archive.h
```diff
@@ -0,0 +1,44 @@
...
+// FileSys namespace
+
+namespace FileSys {
+
+/// Archive backend for general save data archive type (SaveData and SystemSaveData)
+class SaveDataArchive : public ArchiveBackend {
+public:
+    SaveDataArchive(const std::string& mount_point_) : mount_point(mount_point_) {}
+
+    virtual std::string GetName() const override {
```
bunnei: No need for `virtual` keyword?

src/core/file_sys/archive_extsavedata.cpp
wwylele: Just copy-pasted argument list of DiskFile constructor. If it is not good, I will do a clean up

src/core/file_sys/archive_extsavedata.cpp
```diff
@@ -11,13 +11,139 @@
...
+public:
+    ExtSaveDataArchive(const std::string& mount_point) : SaveDataArchive(mount_point) {}
+
+    virtual std::string GetName() const override {
+        return "ExtSaveDataArchive: " + mount_point;
+    }
+
+    ResultVal<std::unique_ptr<FileBackend>> OpenFile(const Path& path,
+                                                     const Mode mode) const override {
+        return SaveDataArchive::OpenFile(path, mode);
```
wwylele: Oops. shouldn't be here :p

src/core/file_sys/savedata_archive.h
```diff
@@ -0,0 +1,44 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <string>
+#include <vector>
```
bunnei: This is not used in this header

src/core/file_sys/archive_sdmc.cpp
```diff
@@ -8,13 +8,307 @@
...
+    if (FileUtil::Delete(full_path))
+        return RESULT_SUCCESS;
+
+    LOG_CRITICAL(Service_FS, "Unknown error deleting %s", full_path.c_str());
+    return ERROR_NOT_FOUND;
+}
+
+ResultCode SDMCArchive::RenameFile(const Path& src_path, const Path& dest_path) const {
+    if (FileUtil::Rename(mount_point + src_path.AsString(), mount_point + dest_path.AsString()))
+        return RESULT_SUCCESS;
```
bunnei: nit: Prefer using { and } even for single-line if statements... Will stop pointing this out

src/core/file_sys/archive_extsavedata.cpp
wwylele: Actually all `LOG_CRITICAL` code path should be unreachable. As for here, if the mount point is invalid, the archive shouldn't even have been opened and the archive factory handles the error. I didn't make them `UNREACHABLE()` just in case of user edit the save directory while citra is running, or other strange cases.

wwylele: Probably I should make a comment of this?

src/core/file_sys/archive_sdmc.cpp
wwylele: I hesitated when deciding to put these error codes in to one header file... Is it OK if I put them into `archive_backend.h` or a new file?

src/core/file_sys/archive_sdmc.cpp
```diff
@@ -8,13 +8,307 @@
...
+    PathParser path_parser(path);
+
+    if (!path_parser.IsValid()) {
+        LOG_ERROR(Service_FS, "Invalid path %s", path.DebugStr().c_str());
+        return ERROR_INVALID_PATH;
+    }
+
+    auto full_path = path_parser.BuildHostPath(mount_point);
+
+    //
```
bunnei: ?

src/core/file_sys/archive_sdmc.h
```diff
@@ -14,6 +14,32 @@
 
 namespace FileSys {
 
+/// Archive backend for SDMC archive
+class SDMCArchive : public ArchiveBackend {
+public:
+    SDMCArchive(const std::string& mount_point_) : mount_point(mount_point_) {}
+
+    virtual std::string GetName() const override {
```
bunnei: No need for `virtual` keyword

src/core/file_sys/archive_sdmcwriteonly.cpp
```diff
@@ -0,0 +1,57 @@
...
+    : sdmc_directory(sdmc.sdmc_directory) {}
+
+ResultVal<std::unique_ptr<ArchiveBackend>> ArchiveFactory_SDMCWriteOnly::Open(const Path& path) {
+    auto archive = std::make_unique<SDMCWriteOnlyArchive>(sdmc_directory);
+    return MakeResult<std::unique_ptr<ArchiveBackend>>(std::move(archive));
+}
+
+ResultCode ArchiveFactory_SDMCWriteOnly::Format(const Path& path,
+                                                const FileSys::ArchiveFormatInfo& format_info) {
+    // This is kind of an undesirable operation, so let's just ignore it. :)
```
bunnei: Log something?

src/core/file_sys/path_parser.cpp
```diff
@@ -0,0 +1,98 @@
...
+    // Although some of these characters are valid on 3DS, they are unlikely to be used by games.
+    if (std::find_if(path_string.begin(), path_string.end(), [](char c) {
+            static const std::set<char> invalid_chars{'<', '>', '\\', '|', ':', '\"', '*', '?'};
+            return invalid_chars.find(c) != invalid_chars.end();
+        }) != path_string.end()) {
+        is_valid = false;
+        return;
+    }
+
+    Common::SplitString(path_string, '/', path_sequence);
```
bunnei: Seems error prone.. Should we have something that uses the platform-correct path separator?

src/core/file_sys/ivfc_archive.cpp
```diff
@@ -31,22 +31,25 @@ ResultCode IVFCArchive::DeleteFile(const Path& path) const {
                       ErrorLevel::Status);
 }
 
-bool IVFCArchive::RenameFile(const Path& src_path, const Path& dest_path) const {
+ResultCode IVFCArchive::RenameFile(const Path& src_path, const Path& dest_path) const {
     LOG_CRITICAL(Service_FS, "Attempted to rename a file within an IVFC archive (%s).",
                  GetName().c_str());
-    return false;
+    // TODO: Use correct error code
```
wwylele: This TODO is a existing one moved from `archive.cpp`... well I will added it anyways

src/core/file_sys/path_parser.h
```diff
@@ -0,0 +1,61 @@
...
+
+    /// Checks the status of the specified file / directory by the Path on the host file system.
+    HostStatus GetHostStatus(const std::string& mount_point);
+
+    /// Builds a full path on the host file system.
+    std::string BuildHostPath(const std::string& mount_point);
+
+private:
+    std::vector<std::string> path_sequence;
+    bool is_valid;
```
bunnei: These should be initialized, e.g. `bool is_valid{};`

src/core/file_sys/savedata_archive.cpp
```diff
@@ -0,0 +1,308 @@
+// Copyright 2016 Citra Emulator Project
```
bunnei: I see this file is largely a copy of the above... Will not point out all the same things above

src/core/file_sys/archive_sdmcwriteonly.cpp
wwylele: Copy-pasted from `ArchiveFactory_SDMC`... I guess I need to add more commits to clean up these existing crappy code.

src/core/file_sys/archive_extsavedata.cpp
bunnei: Sure.. I guess that makes sense. I wonder if we should have a third state similar to "UNREACHABLE"  in that it's unreachable on HW, but could in fact happen under unlikely circumstances in the emulator (in which case we'd still probably handle it gracefully with a generic error)

src/core/hle/service/ptm/ptm.cpp
bunnei: Ah.. yeah, ignore these then

src/core/file_sys/ivfc_archive.cpp
bunnei: Gotcha

wwylele:
Addressed comments above.

src/tests/core/file_sys/path_parser.cpp
```diff
@@ -0,0 +1,40 @@
...
+    REQUIRE(PathParser(Path("/a")).IsValid());
+    REQUIRE(!PathParser(Path("/a/b/../../c/../../d")).IsValid());
+    REQUIRE(PathParser(Path("/a/b/../c/../../d")).IsValid());
+    REQUIRE(PathParser(Path("/")).IsRootDirectory());
+    REQUIRE(!PathParser(Path("/a")).IsRootDirectory());
+    REQUIRE(PathParser(Path("/a/..")).IsRootDirectory());
+}
+
+TEST_CASE("PathParser - Host file system", "[core][file_sys]") {
+
```
bunnei: nit: Unnecessary space

bunnei:
LGTM

wwylele:
Added a commit to "fix" the linking error on OS X tests build. Not sure if this is the correct way to do it.

wwylele:
Removed tests for `PathParser` for now, because I can't get it built on OS X.

wwylele:
Added a work-around to resolve macOS linking error and it can build correctly. Since it is an issue unrelated to this PR, I think we can discuss and fix it later after merging this first. I don't want it to block this PR.

This has been bleeding edge build for about a month, and, aside from one bug that has been fixed, I didn't receive any regression report, so I think this is good to go.

src/core/file_sys/archive_extsavedata.cpp
```diff
@@ -11,13 +11,126 @@
...
+#include "core/file_sys/savedata_archive.h"
 #include "core/hle/service/fs/archive.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // FileSys namespace
 
 namespace FileSys {
 
+/**
+ * A modified version of DiskFile for fix-size file used by ExtSaveData
```
Subv: fixed-size

src/core/file_sys/archive_sdmc.h
```diff
@@ -33,6 +59,8 @@ class ArchiveFactory_SDMC final : public ArchiveFactory {
     ResultCode Format(const Path& path, const FileSys::ArchiveFormatInfo& format_info) override;
     ResultVal<ArchiveFormatInfo> GetFormatInfo(const Path& path) const override;
 
+    friend class ArchiveFactory_SDMCWriteOnly;
```
Subv: Is this friend declaration only so that you can access sdmc_directory from SDMCWriteOnly?

src/core/file_sys/archive_sdmcwriteonly.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+ResultVal<std::unique_ptr<DirectoryBackend>> SDMCWriteOnlyArchive::OpenDirectory(
+    const Path& path) const {
+    LOG_ERROR(Service_FS, "Not supported");
+    return ERROR_UNSUPPORTED_OPEN_FLAGS;
+}
+
+ArchiveFactory_SDMCWriteOnly::ArchiveFactory_SDMCWriteOnly(const ArchiveFactory_SDMC& sdmc)
+    : sdmc_directory(sdmc.sdmc_directory) {}
```
Subv: If you only need sdmc_directory, just pass it directly from the caller instead of taking an ArchiveFactory_SDMC

src/core/hle/service/fs/archive.cpp
```diff
@@ -543,11 +521,15 @@ void RegisterArchiveTypes() {
     std::string sdmc_directory = FileUtil::GetUserPath(D_SDMC_IDX);
     std::string nand_directory = FileUtil::GetUserPath(D_NAND_IDX);
     auto sdmc_factory = std::make_unique<FileSys::ArchiveFactory_SDMC>(sdmc_directory);
-    if (sdmc_factory->Initialize())
+    if (sdmc_factory->Initialize()) {
+        auto sdmcwriteonly_factory =
+            std::make_unique<FileSys::ArchiveFactory_SDMCWriteOnly>(*sdmc_factory);
```
Subv: Here, just pass sdmc_directory instead of the entire SDMC factory

wwylele:
@Subv (re `ArchiveFactory_SDMCWriteOnly`) my thought of design there is that `ArchiveFactory_SDMCWriteOnly` is not a stand-along factory, but somehow derived from `ArchiveFactory_SDMC`. It doesn't have `Initialize` and doesn't check the validity of the path. All these are handled by the SDMC factory. The write-only one can be created only when SDMC factory is successfully created, so I make it only accept `ArchiveFactory_SDMC` as a constructor argument.

An alternative way (that I come up with today) to implement my thought is to delete the constructor of `ArchiveFactory_SDMCWriteOnly` and it can only be created from a function of `ArchiveFactory_SDMC`. Would this be better?

Subv:
I don't think they should be dependent on each other, after all you're not checking the validity of the SDMC archive before creating the WriteOnly archive

src/core/hle/service/fs/archive.cpp
```diff
@@ -526,6 +527,13 @@ void RegisterArchiveTypes() {
         LOG_ERROR(Service_FS, "Can't instantiate SDMC archive with path %s",
                   sdmc_directory.c_str());
 
+    auto sdmcwo_factory = std::make_unique<FileSys::ArchiveFactory_SDMCWriteOnly>(sdmc_directory);
+    if (sdmcwo_factory->Initialize())
```
wwylele: Followed the code style around here that doesn't add braces. Am I supposed to fix this?

wwylele:
Addressed

Subv:
LGTM

Subv:
Can we get this merged please?

PR #2134, opened 2016-10-22
Travis: only upload for push

wwylele:
It seems that travis always upload the binary to the server even if it is a pull request build. This becomes a problem now because our server is down and it fails all PRs.

Not familiar with the script, hopefully I didn't get it wrong.


MerryMage:
looks fine to me

jroweboy:
LGTM

PR #2136, opened 2016-10-22
Fix the build with clang 3.8.

dcci:
This fixes a bunch of narrowing conversion errors, e.g.:
error: format specifies type 'unsigned long long' but the argument has type
      'u64' (aka 'unsigned long') [-Werror,-Wformat]


linkmauve:
NAK, the underlying type of `uint64_t` is platform-dependent, you should instead replace the formatter with the appropriate macro from cinttypes: http://en.cppreference.com/w/cpp/header/cinttypes#Macros

JayFoxRox:
@dcci It's been 3 weeks, what's the status of this?

Subv:

PR #1820, opened 2016-05-19
Set client SDK version to Service APIs

mailwl:

PR #2139, opened 2016-10-25
FRD: fix GetMyFriendKey

mailwl:


wwylele:
LGTM
just a note: this is split from #1820

PR #1662, opened 2016-04-12
Set accepted EULA version to maximum in config

mailwl:
to get it working, need to remove config file user\nand\data\00000000000000000000000000000000\sysdata\00010017\00000000\config

This PR helps to skip showing ErrEULA applet in some games, for example, #1606, but may be reason for huge regression in games compatibility, because of eula agree games starts to use a lot of unimplemented functions, related to network (AC, CECD).

so, this PR just for test, very uncomplete, to find out how to implement it in right way

PR #2142, opened 2016-10-26
AC_U: Stub functions, used if EULA agreed

mailwl:
This PR stubs functions, which used, if user agreed EULA (with currently closed #1662)

### DIFF
--- a/src/core/hle/service/ac_u.cpp
+++ b/src/core/hle/service/ac_u.cpp
@@ -11,7 +11,11 @@
 
 namespace AC_U {
 
-static struct AcConfig { u8 data[0x200]; } default_config = {};
+struct ACConfig {
+    std::array<u8, 0x200> data;
+};
+
+static ACConfig default_config = {};
 
 static bool ac_connected;
 
@@ -22,8 +26,8 @@ static Kernel::SharedPtr<Kernel::Event> disconnect_event;
 /**
  * AC_U::CreateDefaultConfig service function
  *  Inputs:
- *      64 : AcConfig size << 14 | 2
- *      65 : pointer to AcConfig struct
+ *      64 : ACConfig size << 14 | 2
+ *      65 : pointer to ACConfig struct
  *  Outputs:
  *      1 : Result of function, 0 on success, otherwise error code
  */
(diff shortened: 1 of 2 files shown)


wwylele:
I would call it "stub" instead of "implement"

mailwl:
i renamed commit and PR

src/core/hle/service/ac_u.cpp
```diff
@@ -11,28 +11,119 @@
...
+static void ConnectAsync(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    connect_event = Kernel::g_handle_table.Get<Kernel::Event>(cmd_buff[4]);
+    if (connect_event) {
+        connect_event->name = "AC_U:connect_event";
+        connect_event->Signal();
+        ac_connected = true;
+    }
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
```
wwylele: For the same reason as #2126, probably we want to stub this with an error that can present on real 3DS, before we actually implement AC service? So that game won't continue an AP connect before it causes more troubles.

mailwl: i don't know... without EULA accepted games never uses this function, it's underlaying level for network related services, like SpotPass.
i will test some games with local multiplayer, but at current this should not break compatibility at all.

src/core/hle/service/ac_u.cpp
```diff
@@ -11,28 +11,119 @@
 
 namespace AC_U {
 
+static struct AcConfig { u8 data[0x200]; } default_config = {};
```
lioncash: Please keep struct declarations separate from their usage. This should also be named `ACConfig`.  Also consider using std::array over C arrays.

src/core/hle/service/ac_u.cpp
```diff
@@ -52,6 +143,75 @@ static void GetWifiStatus(Service::Interface* self) {
...
+ *      64 : AcConfig size << 14 | 2
+ *      65 : Output pointer to AcConfig struct
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : Infra Priority
+ */
+static void SetRequestEulaVersion(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 major = cmd_buff[1] & 0xFF;
```
lioncash: These should be `u32` not `u8`. The result of a bitwise AND in this case is `u32`. Your printf formatting specifiers should also use `%u`, not `%d`

mailwl:
for now code looks good? i know, it may lead to maaany bugs in future, when time to simulate multiplayer will appear, but for offline playing nothing changes

src/core/hle/service/ac_u.cpp
```diff
@@ -11,28 +11,123 @@
 
 namespace AC_U {
 
+struct ACConfig {
+    std::array<u8, 0x200> data;
+};
+
+static ACConfig default_config = {};
```
bunnei: nit: No need for the " = "

src/core/hle/service/ac_u.cpp
```diff
@@ -11,28 +11,123 @@
...
+
+static bool ac_connected;
```
bunnei: Please initialize this

bunnei:
Mostly looks fine to me

bunnei:
Thanks for addressing my comments!

PR #2146, opened 2016-10-28
Small fix to let IDA see target.xml

mailwl:
IDA Pro don't add ":" to "qSupported", so, gdbstub don't send back target.xml with register descriptions, and we can't see status bits.
before PR:
ida3

after:
ida4


jroweboy:
Looks like this is a fine fix if I'm reading the docs correctly https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html#qSupported

It appears that qSupported is all thats required and the : comes from the host asking for feature support on the client instead. But the host isn't obligated to send any of those.

LGTM

bunnei:
LGTM

PR #2153, opened 2016-11-02
Crash on MacOS 10.12 because Memory::GetPhysicalPointer return NULL

sahara108:
Please review and test.

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -222,7 +222,8 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     continue;
 
                 u8* texture_data = Memory::GetPhysicalPointer(texture.config.GetPhysicalAddress());
-                if (texture_data == NULL) {
+                if (texture_data == nullptr) {
+                    LOG_ERROR(HW_GPU, "Texture's data at index %d is null", i);
                     return;
                 }
                 if (g_debug_context && Pica::g_debug_context->recorder)
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -222,6 +222,9 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     continue;
 
                 u8* texture_data = Memory::GetPhysicalPointer(texture.config.GetPhysicalAddress());
+                if (texture_data == NULL) {
+                    return;
+                }
                 if (g_debug_context && Pica::g_debug_context->recorder)
                     g_debug_context->recorder->MemoryAccessed(
                         texture_data, Pica::Regs::NibblesPerPixel(texture.format) *


src/video_core/command_processor.cpp
```diff
@@ -222,6 +222,9 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     continue;
 
                 u8* texture_data = Memory::GetPhysicalPointer(texture.config.GetPhysicalAddress());
+                if (texture_data == NULL) {
```
wwylele: use `nullptr`

wwylele:
You may want to log it when it is a null pointer. It might indicate a bug elsewhere, or the "invalid" address actually has other meaning that we haven't figured out yet.

sahara108:
Thanks. I did this quick fix to continue my game. I will try to reproduce the issue and post the log here.

wwylele:
@sahara108 By "log it" I mean something like
```c++
if (texture_data == nullptr) {
    LOG_ERROR(HW_GPU, "null pointer :( ");
    return
}

**Comment by sahara108** [comment]
Ok, miss understood  this

>  Please don't treat this as simple as just ignoring null pointer to prevent crash.

I will fix that and create another pull request

File: src/video_core/command_processor.cpp [hunk 2]
```diff
@@ -222,6 +222,10 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     continue;
 
                 u8* texture_data = Memory::GetPhysicalPointer(texture.config.GetPhysicalAddress());
+                if (texture_data == nullptr) {
+                    LOG_ERROR(HW_GPU, "Texture's data at index %d is null", i);
```
bunnei: Probably should be LOG_CRITICAL

bunnei: nit: Change "null" to "nullptr" in the log

JayFoxRox: yes

JayFoxRox: Actually this message is not true at all: The address might be too high for the GetPointer to succeed.
The nullptr only means that the texture pointer could not be retrieved.

The documentation for GetPointer / GetPhysicalPointer should probably be extended to specify what a nullptr return means [as a physical pointer could very well point to the first byte of memory (address 0) with 1:1 mapping]

src/video_core/command_processor.cpp
```diff
@@ -222,6 +222,10 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                    return;
```
JayFoxRox: This is scary, as it will skip the entire draw call.
As this is not expected to happen, this might even mess up the state.

Any code after the switch/case wouldn't be hit either.

---

A sane fallback would be to set some kind of error texture and add a log entry about stubbed behaviour, but otherwise continue as usual

sahara108: A nullptr means that the current frame is broken and I think it is better to skip the drawing. The program should re-configure the state before every draw and it shouldn't cause any problem.

JayFoxRox:
I just noticed that this is not actually the part where textures are uploaded, but only where texture access is recorded for the debug context?
This makes this less severe, but also shows that the PR description is not descriptive enough..

src/video_core/command_processor.cpp
JayFoxRox: A nullptr only means the address could not be mapped for one of the textures (which has no impact on the entire frame, but only that draw call). This might even be intended behaviour on the GPU  (needs more testing?) to cause specific behaviour.
Furthermore your hack won't skip drawing this frame, but only the current draw call, possibly the batch (needs a more careful review).

Only because a single tex might be bad (and could not be recorded for the debug context!), we should not skip drawing / mess up the state, it could break a lot of otherwise working games or hide additional errors.

We already have a hidden state machine for batching afaik, and we might add more in the future if necessary. Simply leaving the switch/case unexpectedly is absolutely non-standard and might break such state machines.
Example: https://github.com/sahara108/citra/blob/1b5cd4d363a5891ccaef36be95f81f875765b7d7/src/video_core/command_processor.cpp#L452 is not called in your code

We also shouldn't reconfigure the GPU each time because the 3DS itself might contain additional state machines + the 3DS has the concept of a configuration and drawing state.
This is also unrelated because your hack only prevents recording the texture access in the debug context.

So really: Test what happens on real hardware, then implement that behaviour; for additional invalid pointers caused by other missing logic in Citra, keep the draw call and log it as an error so we can look into it in the future.

Breaking the behaviour to avoid crashes should never be done. So if you don't want to do it properly yet, add a log only to avoid the crash here (debug context), then add `assert`ions in the renderers to avoid actual crashing (as the assertion gives us the option to debug + it can cleanly quit).

wwylele:
Since #2195 is merged, this is conflicted out. I didn't carefully read what these two PR do, but I think this is superseded and can be closed?

JayFoxRox:
This code only handles CiTrace now (which wasn't part of this PR).
We should probably find a good solution how CiTrace handles invalid memory addresses in the future.

The original crash @sahara108 was experiencing might occur in the actual renderer now.
I agree that this PR can't fulfill its original goal anymore. Please create a new PR for the moved crashes if te renderers don't already handle it. Closed.

Issue #2030, opened 2016-08-29
Inazuma Eleven 3: won't boot 


wwylele:
More specifically, these three game won't boot, freezing in a quite similar way:

Inazuma Eleven 3 - Bomb Blast
Inazuma Eleven 3 - Team Ogre Attacks
Inazuma Eleven 3 - Lightning Bolt

This is an issue separated from #834. Though CRO has been merged, these three game, using CRO, still cannot boot. I still don't have a clear idea whether this is CRO-related. Here is the log from Lighting Bolt:

```
[  21.043652] Service.SRV <Warning> core/hle/service/srv.cpp:RegisterClient:38: (STUBBED) called
[  21.043703] Service.SRV <Warning> core/hle/service/srv.cpp:EnableNotification:62: (STUBBED) called
[  21.043862] Service.APT <Warning> core/hle/service/apt/apt.cpp:GetLockHandle:124: (STUBBED) called handle=0x00038009 applet_attributes=0x00000000
[  21.044108] Service.NDM <Warning> core/hle/service/ndm/ndm.cpp:OverrideDefaultDaemons:198: (STUBBED) default_daemon_bit_mask=0x0000000F 
[  21.044132] Service.NDM <Warning> core/hle/service/ndm/ndm.cpp:SuspendDaemons:84: (STUBBED) daemon_bit_mask=0x00000009 
[  21.065775] HW.Memory <Error> core/memory.cpp:Read:181: unmapped Read32 @ 0x00000000
[  21.065799] Service.APT <Warning> core/hle/service/apt/apt.cpp:GetStartupArgument:442: (stubbed) called startup_argument_type=1 , parameter_size=0x00000030 , parameter_value=0x00000000
[  21.069209] Service.APT <Warning> core/hle/service/apt/apt.cpp:Enable:132: (STUBBED) called attributes=0x00000000
[  21.069326] Service.APT <Warning> core/hle/service/apt/apt.cpp:GlanceParameter:247: called app_id=0x00000300, buffer_size=0x00001000
[  21.069411] Service.APT <Warning> core/hle/service/apt/apt.cpp:NotifyToWait:107: (STUBBED) app_id=768
[  21.069489] Service.APT <Warning> core/hle/service/apt/apt.cpp:ReceiveParameter:225: called app_id=0x00000300, buffer_size=0x00001000
[...]

PR #2154, opened 2016-11-02
APT::GetStartupArgument: force clear startup argument

mailwl:
This PR do not provide startup argument for applications at all instead of garbage
Partially fix **Inazuma Eleven 3: won't boot** #2030 (Now it shows LEVEL 5 logo before stuck)

### DIFF
--- a/src/core/hle/service/apt/apt.h
+++ b/src/core/hle/service/apt/apt.h
@@ -407,9 +407,11 @@ void CancelLibraryApplet(Service::Interface* self);
  *  Inputs:
  *      1 : Parameter Size (capped to 0x300)
  *      2 : StartupArgumentType
+ *      65 : Output buffer for startup argument
  *  Outputs:
  *      0 : Return header
- *      1 : u8, Exists (0 = does not exist, 1 = exists)
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : u8, Exists (0 = does not exist, 1 = exists)
  */
 void GetStartupArgument(Service::Interface* self);
 
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -458,12 +458,16 @@ void GetStartupArgument(Service::Interface* self) {
         return;
     }
 
-    LOG_WARNING(Service_APT, "(stubbed) called startup_argument_type=%u , parameter_size=0x%08x , "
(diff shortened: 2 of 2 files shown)


wwylele:
Is this service function for debug use? If so, then this looks fine to me and it is probably the actual implementation on retail console (does real 3DS zero the buffer, though?). I would appreciate it if there is a test result from real 3DS.

src/core/hle/service/apt/apt.h
```diff
@@ -385,6 +385,7 @@ void StartLibraryApplet(Service::Interface* self);
  *  Inputs:
  *      1 : Parameter Size (capped to 0x300)
  *      2 : StartupArgumentType
+ *      65 : Output buffer for startup argument
  *  Outputs:
  *      0 : Return header
  *      1 : u8, Exists (0 = does not exist, 1 = exists)
```
wwylele: Just found this is wrong. You can take this chance to correct it against [3dbrew](https://www.3dbrew.org/wiki/APT:GetStartupArgument)

mailwl:
@wwylele, i am not sure, it's debug function. The game ask for command line for Restart - does games on 3DS restart after update?

src/core/hle/service/apt/apt.h
mailwl: maybe on real hw 3DS provide some command lines for apps?

wwylele:
> does games on 3DS restart after update? 

I am not sure which update you mean, and don't understand how this is related to the function...

mailwl:
@wwylele, game update. Now i remember, if game has update, the message about it on Home Menu, before starting game.

src/core/hle/service/apt/apt.h
wwylele: I mean, the documentation here is surely wrong. It should be 
```
0 : Return header
1 : Result of function, 0 on success, otherwise error code
2 : u8, Exists (0 = does not exist, 1 = exists)
```

wwylele:
Sorry for forgetting this for weeks. Did a hwtest with a CIA application and wit the exact same call as Inazuma Eleven 3. The returned result matches and the buffer does get cleared. So LGTM

src/core/hle/service/apt/apt.h
mailwl: fixed

Issue #2155, opened 2016-11-02
Pre-commit hook does not check for clang-format


jphalimi:
Commit Hash: 958e81404b85d60c71626796812be300a07cc41a
OS: Linux jphalimi-desktop 4.4.0-45-generic #66-Ubuntu SMP x86_64 x86_64 x86_64 GNU/Linux (Ubuntu 16.04)

The git hook that is responsible to check code style does not check that clang-format is installed, although it is required.

Consequently, the hook will reject any commit, asking developers to remove the content of the files that were modified or added in the commit.

Here is an example:

```
.git/hooks/pre-commit: 35: .git/hooks/pre-commit: clang-format: not found
!!! src/video_core/shader/shader.cpp not compliant to coding style, here is the fix:
--- src/video_core/shader/shader.cpp	2016-11-02 15:45:55.592484422 -0700
+++ -	2016-11-02 16:05:29.050207149 -0700
@@ -1,160 +0,0 @@
-// Copyright 2015 Citra Emulator Project
-// Licensed under GPLv2 or any later version
-// Refer to the license.txt file included.
-
-#include <atomic>
-#include <cmath>
-#include <cstring>
-#include <unordered_map>
-#include <utility>
-#include <boost/range/algorithm/fill.hpp>
-#include "common/bit_field.h"
-#include "common/hash.h"
-#include "common/logging/log.h"
-#include "common/microprofile.h"
-#include "video_core/pica.h"
```

jphalimi:
I have submitted a patch to fix this issue here: https://github.com/jphalimi/citra/commit/d03c1c6367338ec6708af920aa71dad0427ccd93

JayFoxRox:
About your patch:
Are you planning to send a PR?
What needs to be done to complete this / current problems?

wwylele:
I didn't know we have format check in the hook, for it doesn't work on my PC at all for some reason - it always passes the check even if there is a format error. I have clang-format installed on my computer.

Now I am for completely removing this check: the format check shouldn't be mandatory at pre-commit stage, because one can make temporary / testing / for-fun commits that doesn't strictly follow the  rule, and they just need to clang-format everything eventually when they think it is ready to push it to the master branch. After all, we have another format check in our CI, which is enough for a contributor to learn how to correct the code style.

mailwl:
on MacOS X git checks stylem and do not allow to commit

jphalimi:
Ok, so there's a bunch of things to say here:
[...]

PR #2164, opened 2016-11-10
Check clang-format during pre-commit

jphalimi:
Users who do not have clang-format installed were not able to push commits. This patch gives
better feedback to the developer by explicitely telling the user to install the package.

Tracked-On: https://github.com/citra-emu/citra/issues/2155

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -25,6 +25,13 @@ END
     exit 1
 fi
 
+if ! hash "clang-format" 2> /dev/null; then
+    cat<<END
+Error: You must have clang-format installed on your system for codestyle checking purposes.
+END
+    exit 1
+fi
+
 for f in $(git diff --name-only --diff-filter=ACMRTUXB --cached); do
     if ! echo "$f" | egrep -q "[.](cpp|h)$"; then
         continue


bunnei:
LGTM

Subv:
What will happen to developers who try to make commits from Windows? do we also have to install clang-format? Does it have a simple installation method or will I have to fiddle with the %PATH%?

jphalimi:
@Subv That is an excellent point. It seems there is a consensus saying that clang-format checking is ill-formed... Should we go ahead and remove it entirely?

yuriks:
clang-format comes in the LLVM installer, so we can rely on the path for that, but it's yet another huge thing for people to download. I think that if you checked if it's present on the path first before attempting to run it it would be fine (and just ignore it otherwise). At that point though you might as well make an optional script that people can add to their hooks themselves if they wanted though.

yuriks:
>At that point though you might as well make an optional script that people can add to their hooks themselves if they wanted though.

And to be honest, this should probably also be the case for the rest of the stuff in the hook. Just move everything to a script that people can call from their hook if they want to.

freiro:
I think this PR should be merged because right now it's useful, we can then discuss about moving everything to a script.

emmauss:
@freiro currently, i am trying to clean up the algorithm. when there is no fault there, then we discuss design issues like identifier name, position of  functions and even the data types. not to mention me fixing clang issues.

bunnei:
Closing, as the the suggestions made above can/should just be another PR. Feel free to submit a followup.

PR #2165, opened 2016-11-10
Better Pica Vertex caching

jphalimi:
We can improve the vertex caching of Pica, by:
- Using a better caching data structure, having O(1) access time.
- Caching the output vertices, instead of the output registers

Consequence of the patch:
- Increases memory footprint slightly, since we increase
the size of the vertex cache.
- Performance increase, since more values are cached.
- Performance increase, since cache lookup time is now O(1).

From my performance analysis of Pokemon OR, I see that:
- ToVertex is a big hotspot (7% during intro scene)
- WritePicaReg is also a big hotspot (5% during the intro scene)

**Here are the results from my perf analysis, based on the intro scene of Pokemon OR:**
Without patch: Intro lasts 88 seconds
With patch: Intro lasts 84 seconds (-4.5%)

**Without patch:**
without_patch
**With patch:**
with_patch

**The way the patch works:**
Instead of a circular buffer, we use an unordered_map which has the benefit of having a O(1) access time. This removes the circular buffer lookup loop. We use a separate vector to avoid the allocation overhead of unordered_map (otherwise, for each map emplace call, there would be a call to malloc). Finally, instead of caching the output registers, we cache the output vertex, because ToiVertex is an expensive (and very hot) operation (see pictures attached).

Since I am new to the community, I am unsure whether there are performance suites that are used for performance improvements. I haven't seen any. I am looking for a deeper understanding of this patch. Perhaps someone can help me figuring how much other games benefit from the patch?


wwylele:
@JayFoxRox does this conflict with your gs branch?

yuriks:
IRC discussion:
```
[22:52:58] <yuriks> jphalimi: looking at your PR... did you try any other approaches before?
[22:53:37] <yuriks> specifically, how much of this difference is because of caching OutputVertex instead of OutputRegisters?
[22:54:50] <yuriks> when I implemented the cache I tried a few approaches (unordered_map, direct-mapped cache, and the N-way cache that's there right now) along with several cache sizes, and this was the faster combination
[22:56:13] <yuriks> things probably changed in the meantime which shifted the balance of the cost of lookup vs. processing a vertex, but I'm surprised unordered_map is faster
[22:56:35] <jphalimi> Hm that's interesting. Actually using unordered_map comes with a cost
[22:57:04] <yuriks> furthermore, increasing the cache size past a certain point should have almost no effect on hit rates
[22:57:55] <jphalimi> If you look at the stack traces, it adds an std::list lookup hotspot, but combined, it generates less overhead than computing ToVertex all the time
[22:58:23] <yuriks> a cache of 64 (instead of 32) had negligible increase on hit rates, but was slower to scan (I think because it spills to another CPU cacheline), so ended up slower overall, increasing the cache further didn't improve hit rates at all on my test cases
[22:58:53] <yuriks> jphalimi: can you do the change first changing it so that it simply caches outputvertex instead of the registers, and measure that separately from the datastructure change?
[22:59:15] <jphalimi> I can do the comparison analysis
[22:59:37] <jphalimi> But first: why do you think it would be less efficient to cache registers instead of verrtex?
[23:00:05] <yuriks> I don't, you said caching vertices was faster, so I'm believing you :)
[23:00:50] <yuriks> I'm just skeptical of the change in data structure/cache size
[23:05:19] <jphalimi> @yuriks: Gotcha. I think, as I said in the PR, we should have more data from different configurations to decide what to do
[23:05:35] <jphalimi> That's the reason why I asked people if they could help try the patch and tell me if they see similar improvement
[23:06:15] <yuriks> yeah, I'm not sure, I used to just test my changes in fire emblem...
[23:06:23] <yuriks> but that was before we even had a vertex shader JIT
[23:06:46] <yuriks> I think that testing in pokemon is fine. The relative gains might vary per game
[23:07:07] <yuriks> but as long as it's faster it's faster, unless you have an unusually long vertex shader
[23:07:47] <yuriks> (which might tip the scale in favor of caching more at the expense of less executions of the shader)
[23:08:05] <yuriks> but given vertex caching being what it is, I don't think that'll ever be the case with a reasonable cache size
[23:08:33] <yuriks> GPUs themselves only have a cache of a handful vertices, because there's no benefit past a certain point
[23:08:47] <jphalimi> I thought measuring the opening scene duration was a good way to do, since it does not go full speed (at least on my computer) and it is a stable workload
[23:09:26] <jphalimi> Actually, a cache size of 1024 is reached easily during some scenes of the opening in Pokemon OR
[23:09:49] <yuriks> well, you're going to hit several vertices
[23:09:59] <yuriks> but what matters if is you're going to actually re-use them
[23:10:09] <yuriks> is if*
[23:10:16] <jphalimi> That is true. I could also get some stats about the usage of the cache itself
[23:10:23] <yuriks> I encourage you to do that
[23:10:32] <jphalimi> Let me gather some stats tomorrow to convince you
[23:10:50] <yuriks> (to prove me wrong, or just to convince yourself of what I'm saying if nothing else)
[23:11:07] <jphalimi> But again, numbers don't lie: I really see some improvement. Is it possible for you to try out this patch in your config so that we can compare the benefits with 2 kinds of hardware?
[23:11:35] <yuriks> I could, but that would require dusting off my citra dev env ;P
[23:11:43] <yuriks> maybe some other day
[23:12:29] <jphalimi> Ahah :) 
```

src/video_core/shader/shader.cpp
```diff
@@ -30,16 +30,15 @@ OutputVertex OutputRegisters::ToVertex(const Regs::ShaderConfig& config) {
     OutputVertex ret;
     // TODO(neobrain): Under some circumstances, up to 16 attributes may be output. We need to
     // figure out what those circumstances are and enable the remaining outputs then.
-    unsigned index = 0;
```
j-selby: It seems that this pattern of counting a separate index is actually required - Pokémon Super Mystery Dungeon goes all "party mode" if we rely solely on the for loop iterator: 



Specifically, the behavior in the first if statement, where we are comparing index against vs_output_total is the killer here.

JayFoxRox: Yes, this is intended. Only active attributes are counted (probably).

JayFoxRox:
As @wwylele pointed out, this is changing code which is also affected by my GS implementation (which I'll start PR'ing again very soon - I only want a new notebook first).

I'll try rebasing my GS stuff onto this for testing (hopefully this week) and can then report wether this change is actually compatible or not. I can't remember from the top of my head. As it's not as invasive as ectors attempt (if I remember correctly) this could actually work.
Even if it's not compatible, we could possibly do something like this in the future, but we should still make basic GPU features such as GS working first, before optimizing.

jphalimi:
@j-selby I really thought I removed the modifications I made in shader.cpp, it seems that I did not :-)

This was actually trying to get rid of useless code, although the compiler might be able to figure it out itself. I agree we can remove this code as it is not what I wanted to showcase in this PR. However, I don't see how your argument is valid, since it seems index === i in this loop. There is no difference in type, no difference in the iteration space. Am I missing something obvious here?

I really can't see how removing index can break something. Otherwise, it should be carefully documented, no?

@JayFoxRox we can work together on your refactoring work if you want, I can try on my side to use your patch with mine as well. :)

Jp

src/video_core/shader/shader.cpp
```diff
@@ -30,16 +30,15 @@ OutputVertex OutputRegisters::ToVertex(const Regs::ShaderConfig& config) {
...
-    for (unsigned i = 0; i < 7; ++i) {
 
-        if (index >= g_state.regs.vs_output_total)
+    for (unsigned i = 0; i < 7; ++i) {
+        if (i >= g_state.regs.vs_output_total)
             break;
 
         if ((config.output_mask & (1 << i)) == 0)
             continue;
```
jphalimi: Gotcha here, it is obvious to me now. :-)

jphalimi:
Following the discussion with @yuriks , I can say now that the cache is indeed very hot. I get around more than 50% cache hits on average on a part of the opening of Pokemon OR (Yes, I could not add the whole file in Google Docs, since there are so many frames to compute). So that brings another good point for this patch. Here are the numbers, in case you are interested:

https://docs.google.com/spreadsheets/d/13uW4It9goEp_T2Rhr3lka4oJi4XARs6G5IcykJCE6xM/edit#gid=0

jphalimi:
Following our discussion with @yuriks , we agreed to revise the patch as currently proposed. While we still believe OutputVertex instances should be cached instead of OutputRegisters, because they are smaller and reduce the overhead of ToVertex() calls, using unordered_map has not shown any performance benefit. Although theoritically, they allow O(1) access, we actually do not require a bigger cache than what is already implemented. As a result, hash cost + additional memory access is not necessarily better than linear search through the original circular buffer.

Thus, I will post a new patch set to OutputVertex caching only.

jphalimi:
I have just updated the patch to reflect the changes mentioned previously. Please let me know your thoughts.

Subv:
With this new approach, what are the relative performance gains?

yuriks:
@Subv about 3% (by my measure). This will probably have to be at least partially reverted for GS support, but I'm fine with that and it's trivial to do.

@jphalimi LGTM, but can you rebase to fix whatever weirdness you did with the merge commit? (The other commit doesn't seem to have correct commiter info either since it's not linking with your github.)

PR #2166, opened 2016-11-10
CMakeLists: Autodetect clang and only then use libc++

endrift:
This removes the need for manually setting the CXX flags (cf. https://github.com/citra-emu/citra/wiki/Building-For-Linux#building-citra) as well as allowing macOS users to build with gcc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,8 +183,10 @@ IF (APPLE)
     FIND_LIBRARY(COCOA_LIBRARY Cocoa)           # Umbrella framework for everything GUI-related
     set(PLATFORM_LIBRARIES ${COCOA_LIBRARY} ${IOKIT_LIBRARY} ${COREVIDEO_LIBRARY})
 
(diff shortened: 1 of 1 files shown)


yuriks:
LGTM

jroweboy:
Pretty straightforward changes. LGTM

endrift:
This doesn't build properly with clang on Arch Linux, and the wiki instructions for clang appear to be out of date. Will update this PR; don't merge it in the meantime.

bunnei:
@endrift, what's the status of this?

endrift:
Got sidetracked and forgot about it. I'm away from my Linux box until Friday, so I won't be able to finish this up until then, but I can probably update it without testing it on Linux.

endrift:
fwiw, to get it building with libc++ instead of libstdc++ on Arch, you needed to link against `-lc++abi`, otherwise it wouldn't work. However, libc++ seems to be a WIP on Linux and not needed for Citra to work when built with Clang.

Issue #2138, opened 2016-10-23
Cut the Rope: Triple Treat ‒ Emulation stop with fatal error at loading screen


Dragios:
Happens with both Shader JIT on/off & hardware/software renderer.
_Does the game works before?_ I don't think so.

Nothing special before the loading screen.
Upon loading the rom it will show a warning not to close the game if you see the candy logo loading as it saves the game automatically, followed by the copyrights notice and then comes the **Activision** logo animation screen.
Next, it will show "The game is best played with a stylus" before it jumps into the loading screen as shown below.

![Cut the Rope - Triple Treat](https://cloud.githubusercontent.com/assets/18496868/19630024/a806ba60-99b4-11e6-8449-658d98edd18b.png)

Log:
![Cmd Log](https://cloud.githubusercontent.com/assets/18496868/19630025/b0a9e746-99b4-11e6-98f6-a6e2ad67a827.png)

wwylele:
> RSL: 0xD8E007F7

```
Module:
  Kernel (1)
Description:
  Invalid handle (1015)
Summary:
  Invalid argument (7)
Level: 
  Permanent (27)
```

So a candidate is unimplemented `mic:u:GetEventHandle`, or in other words, the whole microphone service.

Subv:
See #2168

PR #2168, opened 2016-11-11
MIC_U: Stub service funcions

mailwl:
It's just stub to avoid errors, related with not mapped memory and not created event.

after update to SDL 2.0.5 (current release) will be possible to implement audio capturing

### DIFF
--- a/src/core/hle/service/mic_u.cpp
+++ b/src/core/hle/service/mic_u.cpp
@@ -98,7 +98,9 @@ static void StartSampling(Service::Interface* self) {
 
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
     is_sampling = true;
-    LOG_WARNING(Service_MIC, "(STUBBED) called");
+    LOG_WARNING(Service_MIC, "(STUBBED) called, encoding=%u, sample_rate=%u, "
+                             "audio_buffer_offset=%d, audio_buffer_size=%u, audio_buffer_loop=%u",
+                encoding, sample_rate, audio_buffer_offset, audio_buffer_size, audio_buffer_loop);
 }
 
 /**
@@ -113,7 +115,7 @@ static void AdjustSampling(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
     sample_rate = static_cast<SampleRate>(cmd_buff[1] & 0xFF);
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    LOG_WARNING(Service_MIC, "(STUBBED) called");
+    LOG_WARNING(Service_MIC, "(STUBBED) called, sample_rate=%u", sample_rate);
 }
 
 /**
@@ -172,7 +174,7 @@ static void SetGain(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
     mic_gain = cmd_buff[1] & 0xFF;
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    LOG_WARNING(Service_MIC, "(STUBBED) called, gain = %d", mic_gain);
+    LOG_WARNING(Service_MIC, "(STUBBED) called, mic_gain=%u", mic_gain);
 }
 
 /**
@@ -202,7 +204,7 @@ static void SetPower(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
     mic_power = static_cast<bool>(cmd_buff[1] & 0xFF);
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    LOG_WARNING(Service_MIC, "(STUBBED) called, power = %u", mic_power);
+    LOG_WARNING(Service_MIC, "(STUBBED) called, mic_power=%u", mic_power);
 }
 
 /**
--- a/src/core/hle/service/mic_u.cpp
+++ b/src/core/hle/service/mic_u.cpp
@@ -2,6 +2,9 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include "common/logging/log.h"
+#include "core/hle/kernel/event.h"
+#include "core/hle/kernel/shared_memory.h"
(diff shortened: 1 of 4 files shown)


src/core/hle/service/mic_u.cpp
```diff
@@ -2,37 +2,331 @@
...
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Namespace MIC_U
 
 namespace MIC_U {
 
+enum class Encoding : u8 {
+    ENCODING_PCM8 = 0,
+    ENCODING_PCM16 = 1,
+    ENCODING_PCM8_SIGNED = 2,
+    ENCODING_PCM16_SIGNED = 3
```
wwylele: Use `CamelCase` for enum members

src/core/hle/service/mic_u.cpp
```diff
@@ -2,37 +2,331 @@
...
+static Kernel::SharedPtr<Kernel::SharedMemory> shared_memory;
+static u8 gain = 0;
+static bool power = false;
+static bool is_sampling = false;
+static bool allow_shell_closed;
+static bool clamp = false;
+static Encoding encoding;
+static SampleRate sample_rate;
+static s32 offset;
+static u32 size;
```
wwylele: `offset`, `size` are really not good names for static global variables IMO. They are not descriptive and can be easily hide by local variable, which tends to cause bugs.

src/core/hle/service/mic_u.cpp
```diff
@@ -2,37 +2,331 @@
...
+ */
+static void MapSharedMem(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 size = cmd_buff[1];
+    Handle mem_handle = cmd_buff[3];
+    shared_memory = Kernel::g_handle_table.Get<Kernel::SharedMemory>(mem_handle);
+    if (shared_memory) {
+        shared_memory->name = "MIC_U:shared_memory";
+        shared_memory->Map(Kernel::g_current_process.get(), 0, shared_memory->permissions,
+                           shared_memory->other_permissions);
```
wwylele: I feel like this is a wrong use of shared memory. My first impression to this service funtion is that, the shared memory is created by application from its memory space, and MIC system module maps the memory to its space using the input shared memory object. If what I think is true, then our HLE service don't need to map the memory at all, but just to operate the memory directly using shared memory interface.

src/core/hle/service/mic_u.cpp
```diff
@@ -2,37 +2,331 @@
...
+ *  Inputs:
+ *      0 : Header Code[0x00020000]
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void UnmapSharedMem(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    if (shared_memory) {
+        shared_memory->Unmap(Kernel::g_current_process.get(), shared_memory->base_address);
```
wwylele: ditto

mailwl:
addressed the comments.

i remove memory map&unmap, but i think, it's not rights, because Nintendogs don't shows i lot of unmapped read errors without it. But, for Nintendogs games not enough just mic:u service, Y2R completion must be signaled async.

Anyway, as is, PR fixes #2138, with Shader JIT disabled game works.

mailwl:
hmm... mic sample from devkitPro don't show any memory errors...

src/core/hle/service/mic_u.cpp
```diff
@@ -2,37 +2,320 @@
...
+#include "core/hle/kernel/event.h"
+#include "core/hle/kernel/shared_memory.h"
 #include "core/hle/service/mic_u.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Namespace MIC_U
 
 namespace MIC_U {
 
+enum class Encoding : u8 { PCM8 = 0, PCM16 = 1, PCM8Signed = 2, PCM16Signed = 3 };
```
wwylele: You can add a trailing comma to let clang-format know you want to expand it
```c++
enum class Encoding : u8 {
    PCM8 = 0,
    PCM16 = 1,
    PCM8Signed = 2,
    PCM16Signed = 3,
};

Comment (mailwl) [inline src/core/hle/service/mic_u.cpp]: thanks!

**Comment by Subv** [comment]
Please fix the build and squash the commits.

**Comment by mailwl** [comment]
done

**Comment by Subv** [comment]
LGTM

File: src/core/hle/service/mic_u.cpp [hunk 6]
```diff
@@ -2,37 +2,325 @@
...
+
+    encoding = static_cast<Encoding>(cmd_buff[1] & 0xFF);
+    sample_rate = static_cast<SampleRate>(cmd_buff[2] & 0xFF);
+    audio_buffer_offset = cmd_buff[3];
+    audio_buffer_size = cmd_buff[4];
+    audio_buffer_loop = static_cast<bool>(cmd_buff[5] & 0xFF);
+
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    is_sampling = true;
+    LOG_WARNING(Service_MIC, "(STUBBED) called");
```
wwylele: Log the inputs here?

src/core/hle/service/mic_u.cpp
```diff
@@ -2,37 +2,325 @@
...
+ *      0 : Header Code[0x00040040]
+ *      1 : SampleRate
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void AdjustSampling(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    sample_rate = static_cast<SampleRate>(cmd_buff[1] & 0xFF);
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    LOG_WARNING(Service_MIC, "(STUBBED) called");
```
wwylele: ditto. will stop commenting on same things

mailwl:
done

PR #2171, opened 2016-11-11
Fix Travis build issue on mac

jroweboy:
Removes cmake from the install list. Cmake now up to date on the new travis mac image meaning brew install cmake would throw an error. Doing `--force` means force reinstalling it every build, and ignoring error codes means that you potentially will miss other build breaking errors that could occur at this phase.


bunnei:
LGTM. Thanks for fixing this!

PR #2175, opened 2016-11-14
Change "OS X" to "macOS" in the Readme

PEmu1:
Change "OS X" to "macOS" as Apple no longer uses the OS X branding as of 10.12. Also update the wiki page link as it broke with the new title.


jroweboy:
Please get in contact with us first before changing the name of a wiki article. Its linked in many places and changing the name of the page changes the link to it. If you are going to move the page, at least remake the old page with just a link to the new page. "Moved: *new link*"

Pringo:
@jroweboy Sure. I didn't realize that editing the page would cause issues with linking to it. Sorry!

JayFoxRox:
Due to the many many changes in the naming by Apple I recommend calling it simply "Building for Mac" btw.

Pringo:
I would prefer to keep it as macOS. Mac is the hardware lineup which isn't consistent with the other OS listings. Also I highly doubt that Apple will change the name again any time in the near future.

bunnei:
I can go either way..

JayFoxRox:
I'm also fine eitherway, but expect a "told you so" if it happens :P

LGTM

Issue #2177, opened 2016-11-16
Move user folder on windows to the user's home folder


jroweboy:
Currently on macOS and linux, the user directory for citra (which contains configuration and other things) is located in the current user's home directory/.local On windows, the directory is contained inside the current citra folder, which makes updating more painful than it needs to be. When a user downloads the new nightly build, they need to copy old information from their user folder over to the new download directory, something thats "Yet Another Thing To Train People To Do." I suggest moving the user directory to the home folder on windows as well, so that its unified across all platforms. Potentially we can move it to AppData/Local/citra where it should go, but that might make it more painful for people to find as windows hides the AppData folder.

JayFoxRox:
How does other (recent) Windows software do it?
Is there a guide on the MSDN or something?

(I remember GTA3 and Vice City (15 years ago; feel the age!) storing all savegames etc. in a "My Documents" subfolder or something :P )

SonofUgly:
>How does other (recent) Windows software do it?

Either in Appdata or My Documents. Appdata seems to be becoming the more popular option.

Either way I think there needs to be a way to keep it portable like it is now. Dolphin has it so that if you have a file called portable.txt in the same directory as dolphin.exe then it'll run as a portable install, otherwise it'll put everything in My Documents.

linkmauve:
In Citra it’s the presence of a directory named “user” in the cwd that will trigger that portable mode.

freiro:
> In Citra it’s the presence of a directory named “user” in the cwd that will trigger that portable mode, on every OS.
When Citra is in this non-portable mode, where should the user folder be located (on Windows)?
Because I am pretty sure that if there isn't one, Citra will create it (while not looking elsewhere).

I agree with the OP we should fix this somehow while keeping in consideration that using the same configuration file between git, nightly, bleeding edge might break something.

mailwl:
`char* appdata = getenv("APPDATA");`
return something like "c:\Users\%username%\AppData\Roaming"
don't work on XP :)

linkmauve:
We don’t support Windows XP, do we?  Even Microsoft doesn’t support it anymore.

JayFoxRox:
@freiro we could simply have a citra folder for the main release and patch it for bleeding edge to use a citra-bleeding-edge folder if it ever became a problem

freiro:
this can be closed 👍

PR #2185, opened 2016-11-17
Change "user" folder default location to AppData/Roaming/ on Windows systems

freiro:
Hello, this is my first PR, I tried to address #2177 and move the default location of the "user" folder to AppData/Roaming/Citra Emulator/(*).
If there is a "user" folder next to citra executable, that folder will take priority (portable mode).

Configuration changes should be handled with care to avoid problems between different versions.

Also note that SHGetKnownFolderPath works only on Windows Vista and later.

Thanks to @linkmauve for the suggestions he gave me yesterday.

Let me know what you think.

(*) After a bit of discussion we decided for Roaming instead of Local.

### DIFF
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -26,6 +26,9 @@
 #define stat _stat64
 #define fstat _fstat64
 #define fileno _fileno
+// Windows version, at least Vista is required to obtain AppData Path
+#define WINVER 0x0600
+#define _WIN32_WINNT 0x0600
 #else
 #ifdef __APPLE__
 #include <sys/param.h>
@@ -594,6 +597,21 @@ std::string& GetExeDirectory() {
     }
     return exe_path;
 }
+
+std::string& AppDataLocalDirectory() {
+    // Windows Vista or later only
+    static std::string local_path;
+    if (local_path.empty()) {
+        PWSTR pw_local_path = 0;
+        wchar_t* wchar_local_path;
+        SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, NULL, &pw_local_path);
+        wchar_local_path = pw_local_path;
+        local_path = Common::UTF16ToUTF8(wchar_local_path);
+        // Freeing memory
+        CoTaskMemFree(static_cast<void*>(pw_local_path));
+    }
+    return local_path;
+}
 #else
 /**
  * @return The user’s home directory on POSIX systems
@@ -671,6 +689,11 @@ const std::string& GetUserPath(const unsigned int DirIDX, const std::string& new
     if (paths[D_USER_IDX].empty()) {
 #ifdef _WIN32
         paths[D_USER_IDX] = GetExeDirectory() + DIR_SEP USERDATA_DIR DIR_SEP;
+        if (!FileUtil::IsDirectory(paths[D_USER_IDX])) {
+            paths[D_USER_IDX] =
+                AppDataLocalDirectory() + DIR_SEP + EMU_DATA_DIR DIR_SEP USERDATA_DIR DIR_SEP;
+        }
+
         paths[D_CONFIG_IDX] = paths[D_USER_IDX] + CONFIG_DIR DIR_SEP;
         paths[D_CACHE_IDX] = paths[D_USER_IDX] + CACHE_DIR DIR_SEP;
 #else
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -684,8 +684,7 @@ const std::string& GetUserPath(const unsigned int DirIDX, const std::string& new
 #ifdef _WIN32
         paths[D_USER_IDX] = GetExeDirectory() + DIR_SEP USERDATA_DIR DIR_SEP;
         if (!FileUtil::IsDirectory(paths[D_USER_IDX])) {
-            paths[D_USER_IDX] =
-                AppDataRoamingDirectory() + DIR_SEP EMU_DATA_DIR DIR_SEP USERDATA_DIR DIR_SEP;
+            paths[D_USER_IDX] = AppDataRoamingDirectory() + DIR_SEP EMU_DATA_DIR DIR_SEP;
         }
 
         paths[D_CONFIG_IDX] = paths[D_USER_IDX] + CONFIG_DIR DIR_SEP;
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -598,10 +598,10 @@ std::string& GetExeDirectory() {
     return exe_path;
 }
 
-std::string AppDataLocalDirectory() {
+std::string AppDataRoamingDirectory() {
     PWSTR pw_local_path = nullptr;
     // Only supported by Windows Vista or later
-    SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, nullptr, &pw_local_path);
+    SHGetKnownFolderPath(FOLDERID_RoamingAppData, 0, nullptr, &pw_local_path);
     std::string local_path = Common::UTF16ToUTF8(pw_local_path);
     CoTaskMemFree(pw_local_path);
     return local_path;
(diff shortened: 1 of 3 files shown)


src/common/file_util.cpp
```diff
@@ -594,6 +597,21 @@ std::string& GetExeDirectory() {
     }
     return exe_path;
 }
+
+std::string& AppDataLocalDirectory() {
+    // Windows Vista or later only
+    static std::string local_path;
```
lioncash: Hi, please don't use static lifetime variables for things like this. Also, this function should ideally return the string by value (the fact GetExeDirectory does this is something that should be corrected, but isn't within the scope of this PR).

You should be able to just do:

```cpp
PWSTR path = nullptr;
SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, nullptr, &path);

std::string local_path = Common::UTF16ToUTF8(path);
CoTaskMemFree(path);

return local_path;
```

j-selby:
Would FOLDERID_RoamingAppData be more appropriate here? Citra save data would not be considered machine dependent, and Roaming AppData is generally considered best-practice for this kind of data:

> [User Name\AppData\Roaming] is used for User Profile specific data, while the AppData\Local folder structure is used for Machine Specific data
https://social.technet.microsoft.com/Forums/windows/en-US/b4fd7ac2-8ad9-4a06-8c28-3f7993aa0272/appdata-local-vs-roaming?forum=w7itprogeneral

As you mentioned, this would mean that configuration data ([qt+sdl2]config.ini) would exist between instances. Would this be preferable when a Debug Citra might need different settings then a Release Citra? Of course, as you said, the portable configuration option would suffice, but I would think that would have to be clearly documented.

src/common/file_util.cpp
freiro: For some reason yesterday compilation was failing, fixed 👍

freiro:
> Would FOLDERID_RoamingAppData be more appropriate here? Citra save data would not be considered machine dependent, and Roaming AppData is generally considered best-practice for this kind of data

@j-selby  True, I choose Local AppData since the bleeding edge version does the same. It would be useful to have more feedback on this.

> As you mentioned, this would mean that configuration data ([qt+sdl2]config.ini) would exist between instances. Would this be preferable when a Debug Citra might need different settings then a Release Citra? Of course, as you said, the portable configuration option would suffice, but I would think that would have to be clearly documented.
Documentation would be the next step :)
Regarding "Debug Citra" version, I think that it shouldn't matter too much since it's enough to create the "user" folder next to citra(-qt).exe to use it.
Actually this might allow to switch quickly between two different configurations (release and debug possibly).

j-selby:
@freiro Bleeding Edge also contains binary/library files in its installation, which would, by my above definition, be suited for Local I would guess (though, its own 'user' directory could live in Roaming). I'm likely overthinking this though.

src/common/file_util.cpp
```diff
@@ -594,6 +597,19 @@ std::string& GetExeDirectory() {
     }
     return exe_path;
 }
+
+std::string AppDataLocalDirectory() {
+    std::string local_path;
+    if (local_path.empty()) {
```
freiro: Yes it will, since I removed static, same modifications would apply to GetExeDirectory.
Fixed.

src/common/file_util.cpp
```diff
@@ -594,6 +597,18 @@ std::string& GetExeDirectory() {
     }
     return exe_path;
 }
+
+std::string AppDataLocalDirectory() {
+    std::string local_path;
+    PWSTR pw_local_path = nullptr;
+    // Only supported by Windows Vista or later
+    SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, nullptr, &pw_local_path);
+    local_path = Common::UTF16ToUTF8(pw_local_path);
```
lioncash: local_path's declaration can be joined with the assignment here.

`// Freeing memory` isn't necessary either, it's self-evident that is what's being done.

src/common/file_util.cpp
lioncash: std::string constructs by default to an empty string; it wouldn't be undefined. However, it's correct that the if statement wasn't needed since the static specifier was dropped.

src/common/file_util.cpp
freiro: ok, what about std::string& GetExeDirectory(), shall we leave it to a different PR?

lioncash: yes

src/common/file_util.cpp
```diff
@@ -671,6 +683,11 @@ const std::string& GetUserPath(const unsigned int DirIDX, const std::string& new
     if (paths[D_USER_IDX].empty()) {
 #ifdef _WIN32
         paths[D_USER_IDX] = GetExeDirectory() + DIR_SEP USERDATA_DIR DIR_SEP;
+        if (!FileUtil::IsDirectory(paths[D_USER_IDX])) {
+            paths[D_USER_IDX] =
+                AppDataLocalDirectory() + DIR_SEP + EMU_DATA_DIR DIR_SEP USERDATA_DIR DIR_SEP;
```
linkmauve: You still have one + too much here, you can concatenate `DIR_SEP` and `EMU_DATA_DIR` just by separating them with a space.

Other than that, LGTM.

freiro: Missed it, thanks :)

emmauss:
wouldn't be better to let the user folder be in Documents\Citra and not appdata, like dolphin.some guys are gonna mess up their systems if they want to make changes to the user folder

jroweboy:
@emmauss https://blogs.msdn.microsoft.com/patricka/2010/03/18/where-should-i-store-my-data-and-configuration-files-if-i-target-multiple-os-versions/

In short, configuration files for a single user thats not synced with active directory should go in AppData/Local

emmauss:
@jroweboy but that would only apply to the config folder, but what about the nand files and sdmc files.users would regularly move things to and from there, so an easily accessible location would be good.

jroweboy:
@emmauss 

    regularly move things to and from there

I disagree that people will regularly access these locations. There is a one time setup with a home folder dump, and after that, they wouldn't need to access the folder again for most use cases. The only other time I can see people wanting to access the folder is for syncing files between their real 3ds and citra such as save files. I personally don't have a strong opinion about where it goes, but I prefer to have it where Windows says it should go.

yuriks:
Really disagree with the rationales here... the user directory should go in Roaming. It contains things like save files, sd contents, etc. which are not disposable data. Local should be used for things like caches and application installs.

Also, I think people want to access the directory all the time. Changing hidden config settings, transfering save games, adding files to the SD for use with homebrew, etc. etc.

yuriks:
I would also prefer that this kind of decision be moved to the frontend (we could then just use Qt for example to get this) because I think this makes more sense in the front end. (As well as the filesystem layer in general) but that would be a large undertaking, and this isn't too bad.

mailwl:
summon @hrydgard to this thread, he know all about multiplatform

hrydgard:
Few people will run this on setups where it matters whether you put it in Roaming or Local , but yuriks is right. As long as portable mode still works for those who prefer it, I don't see why it matters. Don't know what other input I'm supposed to provide, heh.

freiro:
Since users are currently having problems about this [(here)](https://discuss.citra-emu.org/d/985-save-files-gone ), I just committed a patch that selects **AppData/Roaming** as default folder.

Please change the PR title when you can.

yuriks:
LGTM

jroweboy:
Has this been tested in mingw builds yet? I looked up and saw that they were in the headers, but it'd be nice to know that it was tested in mingw64 first

https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-headers/include/combaseapi.h#L363
https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-headers/include/shlobj.h#L755

Other than that LGTM

freiro:
Currently works on mingw without problems.

linkmauve:
LGTM.

yuriks:
Can we rename "Citra Emulator" to just "Citra" while we're breaking the directory location?

freiro:
ok, now the folder is C:\Users\username\AppData\Roaming\Citra\.

yuriks:
LGTM

src/common/file_util.cpp
```diff
@@ -26,6 +26,9 @@
 #define stat _stat64
 #define fstat _fstat64
 #define fileno _fileno
+// Windows version, at least Vista is required to obtain AppData Path
+#define WINVER 0x0600
+#define _WIN32_WINNT 0x0600
```
JayFoxRox: Shouldn't this happen before the first include or via command line?
(And verified where necessary?)

PR #2186, opened 2016-11-17
cfg: add config block 0x00090000

wwylele:
This is found needed by Pokemon games, which also causes a freeze at intro in Pokemon Sun/Moon:
```
[ 723.289812] Service.CFG <Error> core/hle/service/cfg/cfg.cpp:GetConfigInfoBlockPointer:285: Config block 0x90000 with flags 2 and size 8 was not found
```

From observation of a config file dumped from my 3DS, 0x90000 and 0x90001 blocks has exactly the same size(8), flags(0xE) and contents, so my theory is these two blocks are duplicated from the same data. Needs to confirm.

One need to remove the pre-existing config file to test this.

### DIFF
--- a/src/core/hle/service/cfg/cfg.cpp
+++ b/src/core/hle/service/cfg/cfg.cpp
@@ -45,7 +45,8 @@ static_assert(sizeof(SaveFileConfig) == 0x455C,
 enum ConfigBlockID {
     StereoCameraSettingsBlockID = 0x00050005,
     SoundOutputModeBlockID = 0x00070001,
-    ConsoleUniqueIDBlockID = 0x00090001,
+    ConsoleUniqueID1BlockID = 0x00090000,
+    ConsoleUniqueID2BlockID = 0x00090001,
     UsernameBlockID = 0x000A0000,
     BirthdayBlockID = 0x000A0001,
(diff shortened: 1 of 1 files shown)


ybha4:
Nice! Does this fix any of the issues in #2041 ?

wwylele:
@ybha4 I don't expect it to fix those issues, but it might helps some of those. Haven't actually tested

JayFoxRox:
I haven't done anything with config savegame yet, how did you find that flags value?
https://www.3dbrew.org/wiki/Config_Savegame lists is as 0x2 (with a questionmark) for 0x00090000.
(I assume it's outdated though; their wording also doesn't make clear that it's the same content)

wwylele:
@JayFoxRox I found the flag and contents of the block by parsing my config savegame through [this](https://gist.github.com/wwylele/830f5046a0304c083e94d455bcda23a2). This is also confirmed by the disassembly of CFG module.

Also: game read this config block through `GetConfigInfoBlk2`, which requires the block has _at least_ the second bit of flag (0x2) set, which is probably why 3dbrew mark it as "0x2?" because it was tested through the same function.

yuriks:
LGTM

Issue #1910, opened 2016-06-14
Bravely Second End Layer - Stop emulation when leaving status menu


whymsical210:
While testing out Bravely Second on Citra, I came across the fact that exiting the status screen seems to just stop the emulation. Not really crash, but just stops, like it's been paused. I have pictures of the console log and the game screen itself for your viewing. 

console log:
![log](http://i.imgur.com/544F1BJ.png)

screenshot of actual game when stuck:
![screenshot](http://i.imgur.com/nNFgGEV.png)

This was on a build that's about 2 days old (The commit used was 63557e3).
I tested on a current nightly (6/14/2016) and as well does it there.

(Markup and style fixes by JayFoxRox)

whymsical210:
The commit used  was 63557e3

SteveSsSsS:
Same problem here, only crashes when i press menu key.
![screenshot_37](https://cloud.githubusercontent.com/assets/21972033/18225443/dc43392e-71f1-11e6-8f67-eb92b14b720e.png)

Marborao:
same here

Marborao:
to bad, becouse the rest of game seems to be working perfectly

Marborao:
Same here. a shame because the rest of the game seems to work perfectly

Strange09:
Been trying on the Bleeding Edge and it wont even open the game.  Shame since it looked like it was going to work perfectly as the Default.

JayFoxRox:
(Removed comments by @Marborao and @Strange09 which didn't contribute anything to the discussion - note that we don't support Bleeding Edge either)

wwylele:
Closed by #2188

PR #2188, opened 2016-11-18
Bravely Default/Second stuck #1822

### LINKED ISSUE #1822: Bravely Default/Second stuck
Got 3 similar crashes in the Bravely games:
Default:
1) Chapter 1 when saving the merchant introduces the friend bot
   Stuck when pressing X to exit the window
2) When closing the special move menu
Second:
Black screen when closing the characters menu

screen shot 2016-05-19 at 3 38 11 pm
[err_log_default_friend _bot.txt](https://github.com/citra-emu/citra/files/273614/err_log_default_friend._bot.txt)

screen shot 2016-05-19 at 3 51 59 pm
[err_log_special.txt](https://github.com/citra-emu/citra/files/273615/err_log_special.txt)

screen shot 2016-05-19 at 3 53 00 pm
[err_log_second.txt](https://github.com/citra-emu/citra/files/273617/err_log_second.txt)

pippo2931:
From shrds work:
Fixes #1822 #1910 
CancelLibraryApplet stub.

### DIFF
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -396,6 +396,12 @@ void StartLibraryApplet(Service::Interface* self) {
     cmd_buff[1] = applet->Start(parameter).raw;
 }
 
+void CancelLibraryApplet(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    LOG_WARNING(Service_APT, "(STUBBED) called");
+    cmd_buff[1] = 1;
+}
+
 void SetScreenCapPostPermission(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -398,7 +398,9 @@ void StartLibraryApplet(Service::Interface* self) {
 
 void CancelLibraryApplet(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
-    LOG_WARNING(Service_APT, "(STUBBED) called");
+    u32 exists = cmd_buff[1];
+
+    LOG_WARNING(Service_APT, "(STUBBED) called Exists %d", exists);
     cmd_buff[1] = 1;
 }
 
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -398,11 +398,11 @@ void StartLibraryApplet(Service::Interface* self) {
 
 void CancelLibraryApplet(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 exists = cmd_buff[1];
+    u32 exiting = cmd_buff[1] & 0XFF;
 
     cmd_buff[1] = 1; // TODO: Find the return code meaning
 
-    LOG_WARNING(Service_APT, "(STUBBED) called exists %u", exists);
+    LOG_WARNING(Service_APT, "(STUBBED) called exiting=%u", exiting);
 }
 
 void SetScreenCapPostPermission(Service::Interface* self) {
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -398,7 +398,7 @@ void StartLibraryApplet(Service::Interface* self) {
 
 void CancelLibraryApplet(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 exiting = cmd_buff[1] & 0XFF;
+    u32 exiting = cmd_buff[1] & 0xFF;
 
     cmd_buff[1] = 1; // TODO: Find the return code meaning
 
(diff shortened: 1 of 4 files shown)


src/core/hle/service/apt/apt.h
```diff
@@ -381,6 +381,17 @@ void PreloadLibraryApplet(Service::Interface* self);
 void StartLibraryApplet(Service::Interface* self);
 
 /**
+ * APT::CancelLibraryApplet service function
+ *  Inputs:
+ *      0 : Command header [0x003B0040]
+ *      1 : u8, Exists (0 = does not exist, 1 = exists)
```
mailwl: maybe log this parameter?

src/core/hle/service/apt/apt.cpp
```diff
@@ -396,6 +396,14 @@ void StartLibraryApplet(Service::Interface* self) {
     cmd_buff[1] = applet->Start(parameter).raw;
 }
 
+void CancelLibraryApplet(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 exists = cmd_buff[1];
+
+    LOG_WARNING(Service_APT, "(STUBBED) called Exists %d", exists);
+    cmd_buff[1] = 1;
```
JayFoxRox: What does this mean? Can we return a symbolic name instead?
Did anyone test if this is a value a 3DS could even return?

I'm fine with stubbing but there should at least be some explanation as to why we return someting or we might cause unintended behaviour. (similar to our network SUCCESS bug which seems to break pokemon)

---

Check some other functions in this file for how we usually format these stubs. I feel like this should be reordered and some blank lines added to seperate logging from actions (ignore me if other instances do it same as you).

mailwl: i dislike this magic value too, but game got stuck on non-zero value (i think need to look closer why). There is another place in citra, where 1 return as result: https://github.com/citra-emu/citra/blob/master/src/core/hle/service/csnd_snd.cpp#L73

mailwl:
well, for me looks good, except bool value should be "and"-ed with &0xFF, because ARM don't care garbage in other 3 bytes of whole u32 word

src/core/hle/service/apt/apt.cpp
```diff
@@ -398,7 +398,9 @@ void StartLibraryApplet(Service::Interface* self) {
 
 void CancelLibraryApplet(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
-    LOG_WARNING(Service_APT, "(STUBBED) called");
+    u32 exists = cmd_buff[1];
+
+    LOG_WARNING(Service_APT, "(STUBBED) called Exists %d", exists);
```
wwylele: nit: `exists=%d`

src/core/hle/service/apt/apt.cpp
```diff
@@ -396,6 +396,15 @@ void StartLibraryApplet(Service::Interface* self) {
     cmd_buff[1] = applet->Start(parameter).raw;
 }
 
+void CancelLibraryApplet(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 exists = cmd_buff[1];
+
+    cmd_buff[1] = 1; // TODO: Find the return code meaning
+
+    LOG_WARNING(Service_APT, "(STUBBED) called exists %u", exists);
```
JayFoxRox: "exists=%u"

please change this, might be confusing in logs otherwise

Subv:
I'm against this change, we have most (all?) the facilities to implement the function properly. If all it does is stop the currently running library applet then we may check if a library applet is running and stop it, see ````HLE::Applets::IsLibraryAppletRunning()````.

My bet is that the non-zero error code is because there's no library applet running when the function is called, I'll check the NS module and report back later.

mailwl:
wah wah, @Subv let users play BD/BS, it's really good games

Subv:
The response code (```cmd_buff[0]```) is ````0x3B0040````.

If no library applet has been preloaded (Either with ````PrepareToStartLibraryApplet```` or ````PreloadLibraryApplet````), the return code is ````0xC8A0CC04````

src/core/hle/service/apt/apt.cpp
```diff
@@ -398,11 +398,11 @@ void StartLibraryApplet(Service::Interface* self) {
 
 void CancelLibraryApplet(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 exists = cmd_buff[1];
+    u32 exiting = cmd_buff[1] & 0XFF;
```
JayFoxRox: `0xFF`

src/core/hle/service/apt/apt.h
```diff
@@ -384,10 +384,10 @@ void StartLibraryApplet(Service::Interface* self);
  * APT::CancelLibraryApplet service function
  *  Inputs:
  *      0 : Command header [0x003B0040]
- *      1 : u8, Exists (0 = does not exist, 1 = exists)
+ *      1 : u8, Application Exiting (0 = not exiting, 1 = exiting)
```
JayFoxRox: nit: `Application exiting`

bunnei:
I'm OK with this - it's obviously just a stub, and this project has approved many other similar stubs. The argument that "this could just be implemented" applies to almost all of these cases, and this is of course welcome as a followup change.

pippo2931:
This closes #1910 too.

PR #2195, opened 2016-11-20
GPU/CiTrace: Avoid calling GetTextures() when not necessary.

Subv:
We don't need to perform this loop if the CiTrace recorder is not enabled.

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -215,18 +215,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
 
         PrimitiveAssembler<Shader::OutputVertex>& primitive_assembler = g_state.primitive_assembler;
 
-        if (g_debug_context) {
+        if (g_debug_context && g_debug_context->recorder) {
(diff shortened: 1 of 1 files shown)


JayFoxRox:
Note that this also affects #2153 . This change feels a lot better though.

However, note that this moves a possible crash into the GPU backends. That's fine in my opinion, but we should still be aware that this influences crash behaviour as that might affect some issues.

bunnei:
LGTM

wwylele:
This is merged with a clang-format error. Please do a quick fix.

Road-Drum:
does this in any way fix the following error?

`[  52.602885] HW.GPU <Critical> video_core/renderer_opengl/gl_shader_gen.cpp:AppendSource:58: Unhandled texture type 1`

which caused this:

kakusei

JayFoxRox:
No it doesn't. If no issue exists for it so far, create a new one with more information (which game, which version of citra [must be one from this repo and NOT bleedinge-edge]). Please don't post in merged PRs unless there is something wrong (unintended merge, regressions etc.)

PR #2196, opened 2016-11-20
Kernel/Loader: Grab the system mode from the NCCH ExHeader.

### LINKED ISSUE #1849: Kernel memory type initialization
Currently we always initialize the kernel with MemoryType = 2 because that gives the biggest allocation to the APPLICATION memory region in FCRAM. However this behavior breaks some sysmodules that expect to be loaded with MemoryType = 0 in the SYSTEM region (Like for example Home Menu), because the SharedFont takes up a lot of space from there.

The solution to this is to peek into the ExHeader and use the MemoryType stored there when calling Kernel::MemoryInit.

It is currently not possible to peek into a file before loading the kernel, but #1817 should make things easier.

This took me way longer to find out than it should have.

Subv:
3dsx and elf files default to system mode 2 (96MB allocated to the application).

This allows Home Menu to boot without modifications.
Closes #1849

### DIFF
--- a/src/core/loader/ncch.cpp
+++ b/src/core/loader/ncch.cpp
@@ -117,6 +117,14 @@ FileType AppLoader_NCCH::IdentifyType(FileUtil::IOFile& file) {
     return FileType::Error;
 }
 
+ResultStatus AppLoader_NCCH::LoadKernelSystemMode(u32& memory_type) {
+    ResultStatus result = LoadExeFS();
+    if (result != ResultStatus::Success)
+        return result;
+    memory_type = exheader_header.arm11_system_local_caps.system_mode;
+    return ResultStatus::Success;
+}
+
 ResultStatus AppLoader_NCCH::LoadExec() {
     using Kernel::SharedPtr;
     using Kernel::CodeSet;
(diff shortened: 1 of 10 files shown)


src/core/loader/ncch.cpp
```diff
@@ -117,6 +117,14 @@ FileType AppLoader_NCCH::IdentifyType(FileUtil::IOFile& file) {
     return FileType::Error;
 }
 
+ResultStatus AppLoader_NCCH::LoadKernelSystemMode(u32& memory_type) {
```
bunnei: Argument name is not the same as in the header

src/core/loader/ncch.h
```diff
@@ -186,6 +186,13 @@ class AppLoader_NCCH final : public AppLoader {
     ResultStatus Load() override;
 
     /**
+     * Loads the Exheader and returns the system mode for this application.
+     * @param system_mode Out variable where the system mode will be stored.
+     * @return ResultStatus result of the operation
+     */
+    ResultStatus LoadKernelSystemMode(u32& system_mode);
```
bunnei: IMO prefer not to use out arguments, instead return a custom return type (either just return the system mode, or if we  need the ResultStatus, return a custom struct or tuple)

yuriks: This is pretty much the ideal use case for ResultVal

yuriks:
LGTM after @bunnei's comments are addressed.

Dragios:
Build error with MinGW
```
C:\projects\citra\src\core\loader\ncch.cpp: In member function 'virtual boost::optional<unsigned int> Loader::AppLoader_NCCH::LoadKernelSystemMode()':
C:\projects\citra\src\core\loader\ncch.cpp:125:52: error: could not convert '((Loader::AppLoader_NCCH*)this)->Loader::AppLoader_NCCH::exheader_header.ExHeader_Header::arm11_system_local_caps.ExHeader_ARM11_SystemLocalCaps::<anonymous>.ExHeader_ARM11_SystemLocalCaps::<anonymous union>::system_mode' from 'BitField<4ull, 4ull, unsigned char>' to 'boost::optional<unsigned int>'
     return exheader_header.arm11_system_local_caps.system_mode;
                                                    ^
src\core\CMakeFiles\core.dir\build.make:3412: recipe for target 'src/core/CMakeFiles/core.dir/loader/ncch.cpp.obj' failed
```

Subv:
Updated.

wwylele:
Not a big deal, but now load a encrypted game will fall into "Could not determine the system mode" error

PR #2197, opened 2016-11-20
implement frame rate limiter

emmauss:
This is a simple frame rate limiter, an alternative to vsync but not meant to be a replacement. Sleep time is based on the time between vblanks, as opposed to a previous implement using average frame time from profile.
No issues in execution, though it is a little bit slower by a few ms than vsync.
Current maximum frame rate is locked to 60fps. i have not experienced any case where a game must run at 30fps, though i hear that some games run natively at 30fps, though that is just due to 3ds hardware and may not be applicable to citra.


src/core/hw/gpu.cpp
```diff
@@ -547,6 +550,18 @@ static void VBlankCallback(u64 userdata, int cycles_late) {
 
     // Reschedule recurrent event
     CoreTiming::ScheduleEvent(frame_ticks - cycles_late, vblank_event);
+
+    // Frame limiting
+    if (Settings::values.toggle_framelimit) {
+        uint32_t time_diff = std::chrono::duration_cast<std::chrono::milliseconds>(
+                                 std::chrono::high_resolution_clock::now() - timer)
+                                 .count();
+        static const float ex = 1000.0f / 60.0f;
```
wwylele: `ex` can be a `std::chrono::duration` object to avoid casting?

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -6,6 +6,7 @@
 #include <cstddef>
 #include <cstdlib>
 #include <memory>
+#include <thread>
```
wwylele: unused include?

src/video_core/renderer_opengl/renderer_opengl.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <array>
+#include <thread>
```
wwylele: ditto

src/core/hw/gpu.cpp
```diff
@@ -547,6 +550,18 @@ static void VBlankCallback(u64 userdata, int cycles_late) {
 
     // Reschedule recurrent event
     CoreTiming::ScheduleEvent(frame_ticks - cycles_late, vblank_event);
+
+    // Frame limiting
+    if (Settings::values.toggle_framelimit) {
```
lioncash: The code within this if statement should be in its own function.

src/core/hw/gpu.cpp
lioncash: `ex` also isn't a very explanatory name. `60.0f` and `1000.0f` are also magic numbers that should be constants with a name.

Subv:
Me having zero experience in this topic, is this how frame limiters are normally implemented? I take it this won't slow down the emulation thread, yes?

RavenHome1:
how about allowing the user to select at what to lock the fps at?

emmauss:
@RavenHome1 currently the aim is to limit the frame rate to 60fps, adjustable fps can be added later. reason being that, a game would run at full 100% speed if it is rendering at 60fps. a fraction of that changes the game speed.

src/citra/config.cpp
```diff
@@ -67,6 +67,8 @@ void Config::ReadValues() {
     Settings::values.use_scaled_resolution =
         sdl2_config->GetBoolean("Renderer", "use_scaled_resolution", false);
     Settings::values.use_vsync = sdl2_config->GetBoolean("Renderer", "use_vsync", false);
+    Settings::values.toggle_framelimit =
+        sdl2_config->GetBoolean("Renderer", "toggle_framelimit", false);
```
jroweboy: Why does SDL default to false when qt defaults to true?

jroweboy: Also you forgot to add the setting to the default sdl config.ini that's in the repo

emmauss: fixed.

src/core/hw/gpu.cpp
```diff
@@ -589,6 +599,18 @@ void Init() {
     LOG_DEBUG(HW_GPU, "initialized OK");
 }
 
+// Frame Limiter
```
lioncash: This comment doesn't explain anything that the title currently can't.

src/core/hw/gpu.cpp
```diff
@@ -589,6 +599,18 @@ void Init() {
...
+void FrameLimiter() {
+    std::chrono::duration<long, std::nano> time_difference =
```
lioncash: why does this specifically use `long` and `std::nano` here instead of just doing.

```cpp
const auto time_difference = std::chrono::high_resolution_clock::now() - update_time_point;
```
?

If you specifically need nanoseconds, there's `std::chrono::nanoseconds`, in which case you can do:

```cpp
const auto difference = std::chrono::nanoseconds(std::chrono::high_resolution_clock::now() - update_time_point);

File: src/core/hw/gpu.cpp [hunk 8]
```diff
@@ -589,6 +599,18 @@ void Init() {
...
+        (std::chrono::high_resolution_clock::now() - update_time_point);
+    uint32_t frame_limit = 60;
```
lioncash:
```cpp
constexpr u32 frames_per_second = 60;
```

src/core/hw/gpu.cpp
```diff
@@ -589,6 +599,18 @@ void Init() {
...
+    const std::chrono::milliseconds expected_frame_time =
```
lioncash:
```cpp
constexpr auto milliseconds_per_frame = std::chrono::milliseconds(1000 / frames_per_second);
```

src/core/hw/gpu.h
```diff
@@ -327,6 +327,9 @@ void Write(u32 addr, const T data);
 /// Initialize hardware
 void Init();
 
+/// Frame Limiter
+void FrameLimiter();
```
lioncash: This doesn't need to be exposed in the header if it's only used in the cpp file. Simply make the function static and move it above VBlankCallback

src/core/hw/gpu.cpp
```diff
@@ -547,6 +551,12 @@ static void VBlankCallback(u64 userdata, int cycles_late) {
 
     // Reschedule recurrent event
     CoreTiming::ScheduleEvent(frame_ticks - cycles_late, vblank_event);
+
+    // Frame limiting
```
lioncash: This comment is somewhat redundant—it doesn't explain anything that the if statement itself can't.

emmauss:
addressed @lioncash requests

kemenaran:
> I take it this won't slow down the emulation thread, yes?

Strangely enough it seems to slow down the emulation thread.

For instance, without the frame limiter my machine can run Ocarina of Time Link's house much faster than normal speed (avg frame duration: 7ms). However, when I activate the frame limiter, the avg frame duration is indeed stable around 16ms (i.e. 60 FPS) – but the game is sluggish, and all actions and animations are slowed down.

_(For what is worth, I also get the same slow-down behavior with VSync.)_

freiro:
@kemenaran From a computational pov there should be no noticeable slow-down.
What actually happens is that Zelda OOT is a 30fps game, it generates 1 "normal frame" followed by 1 "fast frame" and this is messin with the algorithm.

src/core/hw/gpu.cpp
```diff
@@ -39,6 +40,10 @@ static int vblank_event;
 static u64 frame_count;
 /// True if the last frame was skipped
 static bool last_skip_frame;
+/// Start clock for frame limiter
+static std::chrono::high_resolution_clock::time_point update_time_point =
+    std::chrono::high_resolution_clock::now();
+std::chrono::duration<long , std::nano> previous_difference = std::chrono::milliseconds(16);
```
lioncash: I sorta pointed this out in a different place in the previous review, but the type that should be used on the left-hand side should either be `std::chrono::milliseconds` or `auto`, or `std::chrono::nanoseconds` if you specifically need nanoseconds.

This would technically be a bug on Windows if nanosecond precision is required and the expectation is that long is always 64-bits wide. This would be a bug in that case because `long` is 32-bits wide on Windows, unlike Linux, in which long is generally 64-bits wide. Using `std::chrono::nanoseconds` will abstract that away so you don't need to care about integral size requirements.

emmauss:
fixed my branch. current changes include an improvised frame rate mode detection, 30fps games support for frame limit.

yuriks:
Nay for this approach to handling 30 fps games. It doesn't generalize well and relies on an arbitrary heuristic.

If you can run the game fullspeed then just keeping 60 vblanks/sec is fine, there's no need to do anything more complex because the game is 30 fps. If you can't then that means you're going exceed the frametime on the first frame. You can keep track of that (and how much time you exceeded) and sleep less while you are behind, which should cause you to catch up on the next short frame.

Frame 1 (slow): takes 22ms out of 16.6ms: now -5.4ms ahead (behind), don't sleep
Frame 2 (fast): takes 4ms out of 16.6ms: now 12.6-5.4ms ahead, sleep for 7.2ms, now ~0ms ahead
Frame 3 (slow): etc.

yuriks:
(Note that the amount of time you're behind should be capped to something like 3-4 frames, otherwise a very long frame such as a loading screen will fast forward the game for many frames afterwards.)

emmauss:
built upon what @yuriks said. no longer need to skip a frame, but this time sleeps the short frame to 16.6. msvc builds never go above 60fps. i need to test mingw builds, so @jroweboy rebuild BE.

Issue #2173, opened 2016-11-12
Update microprofile


jroweboy:
As part of the changes in https://github.com/citra-emu/citra/pull/2123 we will need to update microprofile code to the new version on github. (Apparently microprofile had a rewrite happen recently) Updating to the latest version of microprofile is blocked on this PR. https://github.com/jonasmr/microprofile/pull/1 Also we should update the readme in the externals and other changes when updating microprofile. It looks like the UI version of microprofile is no more as well (I could be wrong, I just don't see it anywhere), so it might be worth writing up something on how to use the microprofile web version.

JayFoxRox:
I just checked the new codebase. I don't like what I'm seeing..

Changed from: "MicroProfile is a embeddable profiler in a single file, written in C++"
Changed to: "Microprofile is a embeddable profiler in a few files, written in C++"

---

I've also noticed the commit history is generally unclean and documentation is out of date or bad:

* Sample directory README does list a "UI" sample still (which doesn't exist)
* The Makefile for noui still explicitly disables the SDL2 UI (which doesn't seem to exist anymore)
* Project README only mentions "chrome" but probably means "browser"
* ...

Most of us probably only ever used the - now removed - SDL2 UI and not the HTML view, too. So we'll have to review if that even works for us.
Then there is also the problem that it doesn't ship its own cmake yet but contains a source file.

Not only is this a lack in direction, but also feels like a switch from a feather to a 100 ton weight with some rough edges.

While we should consider switching to a new version, we should stick to what works, unless there are issues (or someone really wants to do this). I think this update could be a relatively large change for zero gain. Also the "rewrite" doesn't look nearly as usable as the old version did - Just my 2cts.

PR #2200, opened 2016-11-21
Disable Microprofile on Mingw builds

j-selby:
Microprofile currently has a number of issues on Mingw that (at least, to my eye) would require a major reworking of how it internally works. This PR disables Microprofile on Mingw builds so that they are actually usable.

As of now, Citra (Mingw) crashes when attempting to load a game, with an assertion error @ [externals/microprofile/microprofile.h, line 1177](/citra-emu/citra/tree/master/externals/microprofile/microprofile.h#L1177). Furthermore, much of the timing components of Microprofile do not actually work regardless of this assertion. The rewrite of Microprofile (assuming it fixes this bug) isn't an option at this point due to its lack of a non-web UI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,9 @@ if (NOT MSVC)
 
     if (MINGW)
         add_definitions(-DMINGW_HAS_SECURE_API)
+        # Microprofile causes crashes when launching titles on MinGW
+        add_definitions(-DMICROPROFILE_ENABLED=0)
+
         if (MINGW_STATIC_BUILD)
             add_definitions(-DQT_STATICPLUGIN)
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static")


mailwl:
what about menu in qt-version?

j-selby:
@mailwl This was tested in QT. Microprofile opens and renders happily, but still kills everything when loading a game. No timer bars are ever shown, even after the assertion was disabled.

j-selby:
Please ignore my previous comment - I am an idiot - Microprofile does function correctly assuming that assertion is removed. However, my point still stands - the assertion was there for a reason, and weird stuff is going on here. Occasionally Citra also crashes at other assertions in the Microprofiler, such as in MicroProfileFlip() (specifically I've hit line 1675 so far).

yuriks:
What are the actual reasons it doesn't work on MinGW? Can you get stack traces for the asserts? Is it something related to TLS?

j-selby:
@yuriks 

I could only guess some weird threading issue, but I'm not overly familiar with that.

Sorry I can only produce screenshots, CLion doesn't give much help in terms of text copying them:

Loading ROM:
st1
Occasionally, when ending emulation (Pokemon: SMD):
st2
clion64-21-11-2016_19-59-59

During normal execution at no specific time, with SDL Microprofile open with all timers and timing options enabled:
clion64-21-11-2016_20-03-22

darkf:
Seems like it would be preferable to fix the issues with MinGW (preferably upstream) than gimp the build just because, IMO.

j-selby:
@darkf That would be awesome, though frankly Microprofile is already a mess in terms of its codebase.

In terms of upstreaming it however, there are a number of issues (#2173) that would have to be addressed before this happens, such as the current Microprofile included in Citra being updated, and finding a means around the current lack of a SDL frontend in the latest upstream version.

freiro:
As a temporary fix you can use this cmake command to disable microprofile without editing CMakeFiles.txt
cmake -G "MSYS Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_MAKE_PROGRAM=mingw32-make -DCMAKE_CXX_FLAGS="-DMICROPROFILE_ENABLED=0" ..

JayFoxRox:
Not sure how this still isn't merged.. mingw crashing by default is not acceptable.
However, I think that this patch should also stop depending on microprofile if we set `MICROPROFILE_ENABLED=0` (meaning not using the microprofile directory at all). It's still added as an include directory from what I can tell.

---

A seperate issue to this one is the naming. We use `ENABLE_XYZ` etc. elsewhere, so this should be renamed to `ENABLE_MICROPROFILE`. It should also be made an option. Someone should open an issue if this PR doesn't handle it (in a seperate commit).

CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ if (NOT MSVC)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
 
     if (MINGW)
-        add_definitions(-DMINGW_HAS_SECURE_API)
+        add_definitions(-DMINGW_HAS_SECURE_API -DMICROPROFILE_ENABLED=0)
```
JayFoxRox: Please add this as a seperate line and add a comment why it was disabled

j-selby:
@JayFoxRox Should I stub MICROPROFILE_* methods then, if we are not including it?

JayFoxRox:
Ignore what I said, I didn't consider that we have a rather deep integration with it.

MerryMage:
@JayFoxRox `MICROPROFILE_ENABLED` is a macro internal to microprofile.

j-selby:
Issues addressed.

jroweboy:
LGTM

PR #2202, opened 2016-11-22
Add man documentation

j-selby:
Includes documentation for both both SDL and QT frontends, detailing basic command-line usage and where to find important files/information. This uses mdoc macros to generate its overall layout. Section 6 [(games)](https://linux.die.net/man/) is used for both documentation files, representing games.

This is correctly copied and compressed by man-db when installing using Ubuntu 16.04 when installing Citra via apt/dpkg. It should work on other distributions.

Rendered (for those who don't have access to man):
[citra.6](https://gist.github.com/j-selby/4c4e4c062607a0f50f677ed6500c0ea6)
[citra-qt.6](https://gist.github.com/j-selby/47577d258d616799040b1ed393c05a9b)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,3 +286,15 @@ if(ENABLE_QT AND UNIX AND NOT APPLE)
     install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.xml"
             DESTINATION "${CMAKE_INSTALL_PREFIX}/share/mime/packages")
 endif()
+
+if(UNIX)
+    if(ENABLE_SDL2)
+        install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.6"
+                DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man6")
+    endif()
+
+    if (ENABLE_QT)
+        install(FILES "${CMAKE_SOURCE_DIR}/dist/citra-qt.6"
+                DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man6")
+    endif()
+endif()
--- a/dist/citra-qt.6
+++ b/dist/citra-qt.6
@@ -1,0 +1,40 @@
+.Dd November 22 2016
+.Dt citra-qt 6
+.Os
+.Sh NAME
+.Nm Citra-Qt
+.Nd Nintendo 3DS Emulator/Debugger (Qt)
+.Sh SYNOPSIS
+.Nm citra-qt
+.Op Ar file
+.Sh DESCRIPTION
+Citra is an experimental open-source Nintendo 3DS emulator/debugger.
+.Pp
+.Nm citra-qt
+is the Qt implementation.
+.Sh FILES
+.Bl -tag -width Ds
+.It Pa $XDG_DATA_HOME/citra-emu
+Emulator storage.
+.It Pa $XDG_CONFIG_HOME/citra-emu
+Configuration files.
+.El
+.Sh AUTHORS
+This document is made available to you under the CC-BY license.
+.Pp
+Citra is made by a team of volunteers. These contributors are listed
+ at <\fBhttps://github.com/citra-emu/citra/contributors\fR>.
+.Pp
+.Sh SEE ALSO
+.Bl -tag -width Ds
+.It Xr citra 6
+The SDL frontend of the application
+.El
+.Pp
+Resources are available for this project:
+.Bl -tag -width Ds
+.It <\fBhttps://citra-emu.org\fR>
+The main homepage of the project.
+.It <\fBhttps://github.com/citra-emu/citra\fR>
+The main source code repository for the Citra emulator.
+.Pp
--- a/dist/citra.6
+++ b/dist/citra.6
@@ -1,0 +1,49 @@
+.Dd November 22 2016
+.Dt citra 6
+.Os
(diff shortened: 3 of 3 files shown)


linkmauve:
The actual location of the configuration file is actually `$XDG_CONFIG_HOME/citra-emu/{sdl2,qt}-config.ini`, and the rest of the data isn’t necessarily in `$HOME/.local/share` but is more accurately described as `$XDG_DATA_HOME/citra-emu/`, see https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

CMakeLists.txt
```diff
@@ -285,4 +285,9 @@ if(ENABLE_QT AND UNIX AND NOT APPLE)
             DESTINATION "${CMAKE_INSTALL_PREFIX}/share/pixmaps")
     install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.xml"
             DESTINATION "${CMAKE_INSTALL_PREFIX}/share/mime/packages")
+
+    install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.6"
+            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man6")
+    install(FILES "${CMAKE_SOURCE_DIR}/dist/citra-qt.6"
+            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man6")
```
linkmauve: These two should be installed conditionally depending on ENABLE_SDL2 and ENABLE_QT.

dist/citra.6
```diff
@@ -0,0 +1,39 @@
...
+is the Simple DirectMedia Layer (SDL) implementation.
+.Sh SEE ALSO
+.Xr citra-qt 6
+.Sh FILES
+.Bl -tag -width Ds
+.It Pa $HOME/.local/share/citra-emu
+Save data and configuration files.
+.El
+.Sh AUTHORS
+This manual was produced by jselby@jselby.net.
```
JayFoxRox: I feel like it would be better to have a generic address or link to our homepage here?

dist/citra.6
```diff
@@ -0,0 +1,39 @@
...
+.Pp
+.Nm
+is produced by a team of volunteers on Github at
+ https://github.com/citra-emu/citra. Bug tracking occurs on this
+ platform.
```
JayFoxRox: `Github` -> `GitHub`

However, I'm thinking that linking the homepage is better (in case we ever switch platforms, the website will probably stay and point to our host / bugtracker). Other than that, I like this section :+1:

dist/citra-qt.6
```diff
@@ -0,0 +1,30 @@
...
+ written in C++.
+.Pp
+.Nm citra-qt
+is the Qt implementation.
+.Sh SEE ALSO
+.Xr citra 6
+.Sh FILES
+.Bl -tag -width Ds
+.It Pa $HOME/.local/share/citra-emu
+Save data and configuration files.
```
JayFoxRox: This is used for emulated storage (for a lack of a better term)
There is also `$HOME/.config/citra-emu/` for the frontend configurations.

dist/citra.6
```diff
@@ -0,0 +1,39 @@
...
+.It Fl g Ar port , Fl Fl gdbport Ar port
+Starts the GDB stub on the specified port
+.It Fl h , Fl Fl help
+Shows syntax help and exits
+.It Fl v , Fl Fl version
+Describes the installed version and exits
+.Sh DESCRIPTION
+.Nm
+is an experimental open-source Nintendo 3DS emulator/debugger
+ written in C++.
```
JayFoxRox: Is the license state and language of interest for this man page?
Do other projects do it?

CMakeLists.txt
j-selby: Agreed. I'll fix that up in a sec.

dist/citra.6
j-selby: A 'see also' section would be better for this, I would think. I will tweak this.

dist/citra.6
j-selby: I've seen conflicting reasoning behind this. Dolphin (linked below) includes a manual creators details, as does leafpad (text editor). Firefox includes none and cmake includes a mailing list. It is not that I want to have that there myself, but I'm totally happy to generic-ify that.

j-selby:
I've pushed several changes which address the issues above. Should we keep the 'experimental' tag in this documentation, considering that in the future this may be dropped? The status of the application in this context should be conveyed via the distribution's packaging system (e.g testing/main).

linkmauve:
You should also mention each other in the SEE ALSO section, so that someone reading the man of citra would also know that citra-qt exists.

bunnei:
Looks fine to me, but I'm not a Linux user - I'd appreciate a few signoffs from Citra devs who are before merging. Thanks!

linkmauve:
I just noticed the CC-BY license, any reason for that choice instead of the GPLv2+ of the rest of the project?

j-selby:
@linkmauve From what I hear, licensing documentation/assets under code licenses is bad practice. I was following the precedent set by Dolphin (which is also otherwise GPL licensed) for its man documentation.

linkmauve:
LGTM.

Subv:
Soooo, can we merge this?

PR #2210, opened 2016-11-24
Expose page table to dynarmic for optimized reads and writes to the JIT

jroweboy:
Subv commented that this was something left out from the original jit pr. Dynarmic has a fast path for memory access if you expose the memory backing the page table, and will gracefully fall back to calling citra's memory read and writes if the pointer at that addess is null. See here for the implementation: https://github.com/MerryMage/dynarmic/blob/master/src/backend_x64/emit_x64.cpp#L1863

This is just a simple fix that adds a method to return the array that backs the page table's memory.

### DIFF
--- a/src/core/memory.h
+++ b/src/core/memory.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <array>
 #include <cstddef>
 #include <string>
 #include "common/common_types.h"
@@ -17,6 +18,7 @@ namespace Memory {
 const u32 PAGE_SIZE = 0x1000;
 const u32 PAGE_MASK = PAGE_SIZE - 1;
 const int PAGE_BITS = 12;
+const size_t PAGE_TABLE_NUM_ENTRIES = 1 << (32 - PAGE_BITS);
 
 /// Physical memory regions as seen from the ARM11
 enum : PAddr {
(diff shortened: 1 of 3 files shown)


MerryMage:
`#include <array>` in memory.h

src/core/memory.h
```diff
@@ -17,6 +18,7 @@ namespace Memory {
 const u32 PAGE_SIZE = 0x1000;
 const u32 PAGE_MASK = PAGE_SIZE - 1;
 const int PAGE_BITS = 12;
+const size_t NUM_ENTRIES = 1 << (32 - PAGE_BITS);
```
yuriks: This needs a way less generic name now that's it's been pulled out of the struct.

src/core/memory.h
```diff
@@ -166,4 +168,11 @@ void RasterizerFlushRegion(PAddr start, u32 size);
  * Flushes and invalidates any externally cached rasterizer resources touching the given region.
  */
 void RasterizerFlushAndInvalidateRegion(PAddr start, u32 size);
+
+/**
+ * Dynarmic has an optimization to memory accesses when the pointer to the page exists that
+ * can be used by setting up the current page table as a callback. This function is used to
+ * retrieve the current page table for that purpose.
+ */
+std::array<u8*, NUM_ENTRIES>* GetCurrentPageTableBackingMemory();
```
yuriks: `GetCurrentPageTablePointerArray`. It's not "backing memory" for anything. Also, this should return a const pointer.

lioncash: could just make it `GetCurrentPageTablePointers()` so the name is agnostic from the container used to store them.

yuriks: Well, "array" there is talking about "a contiguous sequence of value in memory", not std::array per se. Since that's the contract expected by Dynarmic for the data I think it's fine either way.

jroweboy: @yuriks the name came from this comment. https://github.com/citra-emu/citra/blob/master/src/core/memory.cpp#L51 Is that comment correct? If not I can change it with this PR as well.

PR #2211, opened 2016-11-24
Travis: Remove build uploading

yuriks:
@archshift has been MIA for a while, so our builds server is currently
down. Furthermore, uploading is broken in both Linux and macOS:

- The version of lftp in Trusty doesn't support the sftp:auto-confirm
  option, which means that uploading would hang, even if the server
  weren't down. It currently tries to connect forever and fails the
  build with a timeout.
- lftp was removed from Homebrew, so it wasn't being installed on macOS
  and invoking it just failed.


jroweboy:
LGTM

PR #2216, opened 2016-11-26
Add files via upload

unholywinter:
Custom appveyor.yml to generate artifacts


jroweboy:
I already have a patch for this hopefully arriving in the next couple of hours or so, that should include uploading artifacts to a remote server for both travis and appveyor. I don't think its worth the time to clean this one up :)

PR #2217, opened 2016-11-27
[Screen layout] Set small screen at 1x resolution

TTFH:
The large screen scales to fit the window and the small screen remains at its native resolution on "Large Frame Layout".


JayFoxRox:
I actually liked that it was smaller than native resolution.
I think we should consider making it scalable instead (that could possibly even replace the single screen layout by simply scaling the smaller screen to 0%).

TTFH:
Solved Clang-Format error, I'll close it for now.

PR #2218, opened 2016-11-27
Rasterizer/Memfill: Set the correct stencil write mask when clearing the stencil buffer.

Subv:
This fixes the lingering outlines in Pokemon and other games.


freiro:
Good work, looks fixed.
Tested under MinGW, Qt static build, OR/AS.

RavenHome1:
@jroweboy  any chance having this in bleeding edge builds

yuriks:
LGTM. Nice catch

PR #2223, opened 2016-11-27
Implement Frame rate limiter

emmauss:
This is my implementation of a framelimiter. the framelimiter works by keeping track of the lag caused by slow frames, and sleeping the difference between the elapsed frame time and the total delay. the delay is reduced by the total frame time after each frame

Thanks goes to @freiro for testing and feedback , and the basic implementation was gotten from @yuriks 's https://github.com/yuriks/super-match-5-dx/blob/master/src/main.cpp#L224

 msvc, mingw and linux gcc builds shows very little difference in frame rate.


emmauss:
removed the switch.

freiro:
This works on both MinGW and MSVC without problems, tested with multiple games.

src/core/hw/gpu.cpp
```diff
@@ -35,6 +37,16 @@ const u64 frame_ticks = 268123480ull / 60;
...
+/// Start clock for frame limiter
+static u32 time_point;
+/// Frame time for current frame
+static u32 frame_time;
+/// Total delay caused by slow frames
+static float time_delay;
+const float fixed_frame_time = 1000.0f / 60;
+// Max lag caused by slow frames. Can be adjusted to compensate for too many slow frames. Higher
+// values increases time needed to limit frame rate after spikes
+static const float MAX_LAG_TIME = 20;
```
lioncash: this can just be `constexpr` without the static. `fixed_frame_time` can also be constexpr, and should also be upper-cased like `MAX_LAG_TIME`.

src/core/hw/gpu.cpp
```diff
@@ -35,6 +37,16 @@ const u64 frame_ticks = 268123480ull / 60;
 static int vblank_event;
 /// Total number of frames drawn
 static u64 frame_count;
+/// Start clock for frame limiter
+static u32 time_point;
+/// Frame time for current frame
+static u32 frame_time;
+/// Total delay caused by slow frames
+static float time_delay;
```
freiro: You can move `static u32 time_point;` to `FrameLimiter() `and you should completely remove `static u32 frame_time;` because it's redefined below.

Something like this: 
``` 
static void FrameLimiter() {
    /// Total delay caused by slow frames
    static float time_delay = 0;
```

src/core/hw/gpu.cpp
```diff
@@ -563,6 +602,8 @@ void Init() {
     framebuffer_sub.active_fb = 0;
 
     frame_count = 0;
+    time_point = Common::Timer::GetTimeMs();
+    time_delay = 0;
```
freiro: You can remove this according to my previous comment.

src/core/hw/gpu.cpp
```diff
@@ -512,6 +524,25 @@ template void Write<u32>(u32 addr, const u32 data);
...
+        time_delay = -MAX_LAG_TIME;
+    }
+    int32_t desired_time = static_cast<int32_t>(time_delay);
+    int32_t elapsed_time = Common::Timer::GetTimeMs() - time_point;
+
+    if (elapsed_time < desired_time) {
+        Common::SleepCurrentThread(desired_time - elapsed_time);
+    }
+
+    int32_t frame_time = Common::Timer::GetTimeMs() - time_point;
```
lioncash: The previous comment applies here as well.

This variable name shadows the variable of the same name declared in the outer file scope. This will cause a compiler warning.

src/core/hw/gpu.cpp
```diff
@@ -512,6 +524,25 @@ template void Write<u32>(u32 addr, const u32 data);
...
 
+static void FrameLimiter() {
+    time_delay += fixed_frame_time;
+    if (time_delay > MAX_LAG_TIME) {
+        time_delay = MAX_LAG_TIME;
+    } else if (time_delay < -MAX_LAG_TIME) {
+        time_delay = -MAX_LAG_TIME;
+    }
+    int32_t desired_time = static_cast<int32_t>(time_delay);
+    int32_t elapsed_time = Common::Timer::GetTimeMs() - time_point;
```
lioncash: the expression `Common::Timer::GetTimeMs() - time_point` results in a `u32`, not an `s32`. You're converting signedness without an explicit cast.

src/core/hw/gpu.cpp
```diff
@@ -528,8 +559,16 @@ static void VBlankCallback(u64 userdata, int cycles_late) {
     // Check for user input updates
     Service::HID::Update();
 
+    if (!Settings::values.use_vsync) {
```
lioncash:
```cpp
if (!Settings::values.use_vsync && Settings::values.toggle_framelimit) {
    FrameLimiter();
}
```

src/core/hw/gpu.cpp
```diff
@@ -512,6 +524,25 @@ template void Write<u32>(u32 addr, const u32 data);
 template void Write<u16>(u32 addr, const u16 data);
 template void Write<u8>(u32 addr, const u8 data);
 
+static void FrameLimiter() {
+    time_delay += fixed_frame_time;
+    if (time_delay > MAX_LAG_TIME) {
```
emmauss: addressed

src/core/hw/gpu.cpp
emmauss: addressed

src/core/hw/gpu.cpp
emmauss: addressed

src/core/hw/gpu.cpp
emmauss: addressed

src/core/hw/gpu.cpp
emmauss: addressed

src/core/hw/gpu.cpp
emmauss: addressed

src/core/hw/gpu.cpp
```diff
@@ -512,6 +523,21 @@ template void Write<u32>(u32 addr, const u32 data);
...
+    time_delay += FIXED_FRAME_TIME;
+    time_delay = MathUtil::Clamp(time_delay, -MAX_LAG_TIME, MAX_LAG_TIME);
+    int32_t desired_time = static_cast<int32_t>(time_delay);
+    int32_t elapsed_time = s32(Common::Timer::GetTimeMs() - time_point);
+
+    if (elapsed_time < desired_time) {
+        Common::SleepCurrentThread(desired_time - elapsed_time);
+    }
+
+    uint32_t frame_time = Common::Timer::GetTimeMs() - time_point;
```
freiro: u32

src/core/hw/gpu.cpp
```diff
@@ -512,6 +523,21 @@ template void Write<u32>(u32 addr, const u32 data);
 template void Write<u16>(u32 addr, const u16 data);
 template void Write<u8>(u32 addr, const u8 data);
 
+static void FrameLimiter() {
+    time_delay += FIXED_FRAME_TIME;
+    time_delay = MathUtil::Clamp(time_delay, -MAX_LAG_TIME, MAX_LAG_TIME);
+    int32_t desired_time = static_cast<int32_t>(time_delay);
+    int32_t elapsed_time = s32(Common::Timer::GetTimeMs() - time_point);
```
freiro: `s32 desired_time`, same for `elapsed_time`

src/core/hw/gpu.cpp
lioncash: `FIXED_FRAME_TIME` is still not constexpr.

src/core/hw/gpu.cpp
```diff
@@ -35,6 +38,14 @@ const u64 frame_ticks = 268123480ull / 60;
 static int vblank_event;
 /// Total number of frames drawn
 static u64 frame_count;
+/// Start clock for frame limiter
+static u32 time_point;
+/// Total delay caused by slow frames
+static float time_delay;
+const float FIXED_FRAME_TIME = 1000.0f / 60;
```
wwylele: `constexpr`, as `MAX_LAG_TIME` is

src/core/hw/gpu.cpp
```diff
@@ -528,8 +554,14 @@ static void VBlankCallback(u64 userdata, int cycles_late) {
...
     Service::HID::Update();
 
+    if (!Settings::values.use_vsync && Settings::values.toggle_framelimit) {
+        FrameLimiter();
+    }
+
     // Reschedule recurrent event
     CoreTiming::ScheduleEvent(frame_ticks - cycles_late, vblank_event);
+
+    time_point = Common::Timer::GetTimeMs();
```
wwylele: It's better to move this line up to before `ScheduleEvent` so that frame limiter codes are together. It will be even more accurate if you count the time delay from `ScheduleEvent`.

freiro: Since you're there move it to Framelimiter().

src/core/hw/gpu.cpp
emmauss: addressed

src/core/hw/gpu.cpp
freiro: Well, true then I agree with keeping outside.

Just in case it will ever be needed: even if is inside, the time value is clamped and the game would stabilize after a few frames.

emmauss: brought it up before the scheduler, but outside the framelimiter.

bunnei:
LGTM

yuriks:
"FrameLimiter" is kind of a bad function name, but other than that, post-merge LGTM

emmauss:
would it be good to create a fixup PR @yuriks ?

yuriks:
If you really want to, you don't need to bother though.

emmauss:
Okay, will leave it at that and think of what to do next.

PR #2240, opened 2016-11-30
Config: auto-select region and language

wwylele:
Added an option to auto-select region based on game region lockout info. System language will also be override if the region is auto-selected and the chosen language is not available. Also made this option as default value. This aims to reduce false positive caused by selecting a wrong region.

~~I'd like to add a prompt that the system language setting may be overrided. but not sure how to add it (add a tooltip?or make a description bar on the bottom of the configuration dialog?)~~ Added a tooltip

### DIFF
--- a/src/citra/config.cpp
+++ b/src/citra/config.cpp
@@ -89,7 +89,8 @@ void Config::ReadValues() {
 
     // System
     Settings::values.is_new_3ds = sdl2_config->GetBoolean("System", "is_new_3ds", false);
-    Settings::values.region_value = sdl2_config->GetInteger("System", "region_value", 1);
+    Settings::values.region_value =
+        sdl2_config->GetInteger("System", "region_value", Settings::REGION_VALUE_AUTO_SELECT);
 
     // Miscellaneous
     Settings::values.log_filter = sdl2_config->Get("Miscellaneous", "log_filter", "*:Info");
--- a/src/citra_qt/configure_system.ui
+++ b/src/citra_qt/configure_system.ui
@@ -129,6 +129,9 @@
         </item>
         <item row="2" column="1">
          <widget class="QComboBox" name="combo_language">
+          <property name="toolTip">
+           <string>Note: this can be overridden when region setting is auto-select</string>
+          </property>
           <item>
            <property name="text">
             <string>Japanese (日本語)</string>
--- a/src/core/hle/service/cfg/cfg.cpp
+++ b/src/core/hle/service/cfg/cfg.cpp
@@ -114,6 +114,8 @@ static const std::vector<u8> cfg_system_savedata_id = {
     0x00, 0x00, 0x00, 0x00, 0x17, 0x00, 0x01, 0x00,
 };
 
+static u32 preferred_region_code = 0;
+
 void GetCountryCodeString(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
     u32 country_code_id = cmd_buff[1];
@@ -159,11 +161,18 @@ void GetCountryCodeID(Service::Interface* self) {
     cmd_buff[2] = country_code_id;
 }
 
+static u32 GetRegionValue() {
+    if (Settings::values.region_value == Settings::REGION_VALUE_AUTO_SELECT)
+        return preferred_region_code;
+
+    return Settings::values.region_value;
+}
+
 void SecureInfoGetRegion(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
     cmd_buff[1] = RESULT_SUCCESS.raw;
-    cmd_buff[2] = Settings::values.region_value;
+    cmd_buff[2] = GetRegionValue();
 }
 
 void GenHashConsoleUnique(Service::Interface* self) {
@@ -183,7 +192,7 @@ void GetRegionCanadaUSA(Service::Interface* self) {
     cmd_buff[1] = RESULT_SUCCESS.raw;
 
     u8 canada_or_usa = 1;
-    if (canada_or_usa == Settings::values.region_value) {
+    if (canada_or_usa == GetRegionValue()) {
         cmd_buff[2] = 1;
     } else {
         cmd_buff[2] = 0;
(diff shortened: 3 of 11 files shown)


src/core/loader/ncch.cpp
```diff
@@ -325,6 +327,21 @@ ResultStatus AppLoader_NCCH::Load() {
 
     Service::FS::RegisterArchiveType(std::make_unique<FileSys::ArchiveFactory_RomFS>(*this),
                                      Service::FS::ArchiveIdCode::RomFS);
+
+    // Set the preferred region from the region lockout in the SMDH
+    std::vector<u8> smdh;
```
lioncash: This code can be in its own function.

src/citra/config.cpp
```diff
@@ -87,7 +87,7 @@ void Config::ReadValues() {
 
     // System
     Settings::values.is_new_3ds = sdl2_config->GetBoolean("System", "is_new_3ds", false);
-    Settings::values.region_value = sdl2_config->GetInteger("System", "region_value", 1);
+    Settings::values.region_value = sdl2_config->GetInteger("System", "region_value", -1);
```
lioncash: You should probably make a named constant for `-1`, so there isn't a magic number all over the place.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -317,6 +331,35 @@ ResultCode GetConfigInfoBlock(u32 block_id, u32 size, u32 flag, void* output) {
     void* pointer;
     CASCADE_RESULT(pointer, GetConfigInfoBlockPointer(block_id, size, flag));
     memcpy(output, pointer, size);
+
+    // override the language setting if the region setting is auto
+    if (block_id == LanguageBlockID && Settings::values.region_value == -1) {
```
Subv: Can you factor this out into its own function, or something? it has a lot of reinterpret_casts, is there no better way to do it?

wwylele: Factored into a function. I tried to reduce casting but at least one reinterpret_cast is unavoidable.

src/core/settings.h
```diff
@@ -109,5 +109,7 @@ struct Values {
     u16 gdbstub_port;
 } extern values;
 
+constexpr int REGION_VALUE_AUTO_SELECT = -1;
```
lioncash: This should have documentation indicating what it represents or it's intended use. A random constant sticking out doesn't say much.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -163,7 +165,12 @@ void SecureInfoGetRegion(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
     cmd_buff[1] = RESULT_SUCCESS.raw;
-    cmd_buff[2] = Settings::values.region_value;
+
+    if (Settings::values.region_value == -1) {
```
lioncash: This and the line below are still using `-1` as a magic value. You could also probably put this into a function since they're essentially duplicated.

```cpp
u32 GetRegionValue() {
    if (Settings::values.region_value == Settings::REGION_VALUE_AUTO_SELECT)
        return preferred_region_code;

    return Settings::values.region_value;
}

File: src/core/hle/service/cfg/cfg.cpp [hunk 6]
```diff
@@ -313,10 +327,42 @@ static ResultVal<void*> GetConfigInfoBlockPointer(u32 block_id, u32 size, u32 fl
...
+        {LANGUAGE_EN, LANGUAGE_FR, LANGUAGE_DE, LANGUAGE_IT, LANGUAGE_ES, LANGUAGE_NL, LANGUAGE_PT,
+         LANGUAGE_RU},
+        // CHN
+        {LANGUAGE_ZH},
+        // KOR
+        {LANGUAGE_KO},
+        // TWN
+        {LANGUAGE_TW},
+    }};
+    auto& available = region_languages[region];
```
lioncash: This can be a const reference.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -313,10 +327,42 @@ static ResultVal<void*> GetConfigInfoBlockPointer(u32 block_id, u32 size, u32 fl
     return MakeResult<void*>(pointer);
 }
 
+/// Checks if the language is available in the chosen region, and overrides it if not fits
+static void AdjustLanguageInfoBlock(u32 region, u8& language) {
```
lioncash: in/out references for this sort of thing is kind of unnecessary. This can return by value; just memcpy the returned value if you need to.

src/core/loader/ncch.cpp
```diff
@@ -309,6 +311,21 @@ ResultStatus AppLoader_NCCH::LoadExeFS() {
     return ResultStatus::Success;
 }
 
+void AppLoader_NCCH::ParseRegionLockoutInfo() {
```
lioncash: This can be a local function, since its not dependent on the class (assuming I've read the code correctly).

wwylele: It is dependent on the class. It calls `ReadIcon`.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -313,10 +322,48 @@ static ResultVal<void*> GetConfigInfoBlockPointer(u32 block_id, u32 size, u32 fl
...
 ResultCode GetConfigInfoBlock(u32 block_id, u32 size, u32 flag, void* output) {
     void* pointer;
     CASCADE_RESULT(pointer, GetConfigInfoBlockPointer(block_id, size, flag));
     memcpy(output, pointer, size);
+
+    // override the language setting if the region setting is auto
+    if (block_id == LanguageBlockID &&
+        Settings::values.region_value == Settings::REGION_VALUE_AUTO_SELECT) {
+        u8 language;
+        memcpy(&language, output, 1);
```
lioncash: `sizeof(u8)` instead of `1`.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -313,10 +322,48 @@ static ResultVal<void*> GetConfigInfoBlockPointer(u32 block_id, u32 size, u32 fl
...
+        {LANGUAGE_ZH},
+        // KOR
+        {LANGUAGE_KO},
+        // TWN
+        {LANGUAGE_TW},
+    }};
+    const auto& available = region_languages[region];
+    if (std::find(available.begin(), available.end(), language) == available.end()) {
+        return available[0];
+    } else {
```
lioncash: `else` isn't necessary, since it's a return. below another return.

wwylele:
@Subv I added a tooltip with language setting combobox. I still feel not good enough but can't find a better place

Dragios:
Maybe something like this?
`Hover over an item above for more details`
Cryptoprevent

src/citra_qt/configure_system.ui
```diff
@@ -129,6 +129,9 @@
         </item>
         <item row="2" column="1">
          <widget class="QComboBox" name="combo_language">
+          <property name="toolTip">
+           <string>Note: this can be overrided when region setting is auto-select</string>
```
bunnei: nit: overridden (overrided is not a word)

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -313,10 +322,47 @@ static ResultVal<void*> GetConfigInfoBlockPointer(u32 block_id, u32 size, u32 fl
     return MakeResult<void*>(pointer);
 }
 
+/// Checks if the language is available in the chosen region, and returns a proper one
+static u8 AdjustLanguageInfoBlock(u32 region, u8 language) {
+    const std::array<std::vector<u8>, 7> region_languages{{
```
bunnei: Can be static

bunnei:
Mostly LGTM

wwylele:
Addressed

@Dragios That looks good, but I prefer doing it in another PR since we need to add descriptions for all options, which is beyond the scope of this PR

mailwl:
don't know if it right or not, but if i set in config `region_value=` without value, then start and stop Citra, then in config i will see `region_value=0` (JPN), not -1 (Auto-select)

mailwl:
another issue: Auto-select don't work MiiSelector (appletEd) from USA region don't work - we get FatalError file not found, looks like default value set to JPN

wwylele:
@mailwl 

> don't know if it right or not, but if i set in config region_value= without value, then start and stop Citra, then in config i will see region_value=0 (JPN), not -1 (Auto-select)

This is already an issue in current master (default is USA but leaving a empty field results in JPN). This seems an inssue in Qt that it treats empty field as integer 0, instead of using the default value. I am not going to address this here (rare case + beyond the PR scope)

> another issue: Auto-select don't work with MiiSelector (appletEd) from USA region - we get FatalError file not found, looks like default value set to JPN

The region detector only works with NCCH file containing a SMDH data, so it won't works for most system modules and applets. I choose JPN as the default value only because it is with code 0, and I don't really have good reasons to choose any of them as default value

PR #2245, opened 2016-11-30
Clang-cl support Citra

freiro:
I currently got clang-cl to fully compile a working version of citra, this PR works without problems and it's a first step to fully support clang-cl as a compiler.
(I actually wanted to PR this to my own repo, but is ready)

I am interested in feedback from Linux (using clang) and macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ if (NOT DEFINED ARCHITECTURE)
 endif()
 message(STATUS "Target architecture: ${ARCHITECTURE}")
 
-IF (WIN32 AND (${CMAKE_CXX_COMPILER_ID}  MATCHES "Clang"))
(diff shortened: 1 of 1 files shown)


MerryMage:
This has never been necessary in my experience. 

For the record I just did a build with cmake 3.4.1 and clang-cl 3.8.0.

freiro:
@MerryMage that's the point of the discussion we were having.

JayFoxRox:
Which discussion? There is no discussion on record here at github.

freiro:
Well, to sum this up:
a few days ago Citra wasn't able to compile on Windows using cmake (3.5.2 iirc) + clang-cl 3.9.0, the error was: 
> clang-cl.exe : error : unknown argument ignored in clang-cl: '-std=c++14'
And this is the error the PR was targeting.

In the attempt to solve the other errors related to dynarmic (broken cpu jit and missing compiler flags) I updated to Cmake 3.7.1 and clang-cl 4.0.0 (r284979). 

Merry told me she didn't encounter this error before, I double checked on the latest master and it works without patchs.
I tried with an older cmake and still works, this might have been a bug in that specific version I was using.

Issue #2125, opened 2016-10-12
Asynchronous file reading softlocks Brunswick Pro Bowling


neobrain:
# Observation

After making two small modifications (see below) to the Citra source that _technically_ should improve Citra's emulation accuracy of the CTROS scheduler, Brunswick Pro Bowling softlocks upon startup in a function called FileRead. This is surprising, since an asynchronous file reading operation seems like a very elementary operation, and it's one of the first things the game is doing to load its resources.
# Analysis

The following outlines the behavior of my (modified) Citra build, which reflects our current understanding of the CTR scheduler:
- The game ships with an `.axf` file in its RomFS that includes full symbol information (function names and parameter list, type information), making analysis of this issue very easy
- FileRead requests a worker thread to read data from a RomFS file by appending a work item to a task queue. The addition of the new item is signaled using a LightEvent (implemented via svcArbitrateAddress), which causes a deschedule of the main thread.
- The worker thread gets scheduled in and finds a new item appended to the task queue, and hence eventually calls FS::File::Read (using svcSendSyncRequest) on a RomFS FS file handle. In (modified) Citra, this causes a deschedule of the worker thread (see the list of modifications to see why; for reference, the file reading operation is executed before the deschedule).
- The main thread is scheduled back in and enters a stationary loop (presumably waiting for the asynchronous file reading operation to complete) without ever calling a system call. It hence never gets scheduled out and hence indefinitely blocks the AppCore, preventing the worker thread from ever getting scheduled back in.
- Note that the worker thread technically has higher priority than the main thread, but this is useless since on the real system the svcSendSyncRequest wouldn't return immediately (since file reading takes some time) but rather block the current thread until the service process replies.

As far as I can tell, this behavior is perfectly in line with what's publicly documented about the cooperative CTROS scheduler, which suggests we are missing a piece of the bigger picture.
# Theories

I considered the following theories to resolve the issue. Unfortunately, most of these seem to have already been proven wrong, and the rest are plain guesses.
- Maybe BPB runs threads on the SysCore? If the worker thread were allowed to run on the SysCore, it could be scheduled in even when the AppCore is blocked. However, the game's exheader suggests it's using the AppCore for the main thread, and the parameters to svcCreateThread suggest it doesn't use SysCore on the helper threads either.
[...]

PR #2249, opened 2016-12-01
Kernel/IPC: Use Ports and Sessions as the fundamental building block of Inter Process Communication.

Subv:
This is an overhaul of #1911 and a continuation of #1842

This PR is part of an ongoing effort to implement multiprocess support in Citra.

All handles obtained via srv::GetServiceHandle or svcConnectToPort are references to ClientSessions.
Service modules will wait on the counterpart of those ClientSessions (Called ServerSessions) using svcReplyAndReceive or svcWaitSynchronization[1|N], and will be awoken when a SyncRequest is performed.

All IPC parameter translation for LLE modules will happen inside the implementation of svcReplyAndReceive, this is not implemented in this PR.

All IPC parameter translation for HLE modules should happen in the SessionRequestHandler class, this is not implemented in this PR.

HLE Interfaces are no longer kernel objects, and now derive from SessionRequestHandler, they are bound to a Session pair when a connection request is initiated by the emulated application (See svcConnectToPort and srv:GetServiceHandle).

There is currently a very important unimplemented feature related to #2125. SyncRequests should put the calling thread to sleep until the server counterpart of the session handles and answers the request (by means of svcReplyAndReceive), but this is all instantaneous in our current HLE modules.
It is also speculated that the real kernel preempts the threads running in the AppCore and reschedules to another thread if said thread was waiting on a SyncRequest and the request was completed. (This speculation comes from the aforementioned #2125 and reverse engineering of the svcReplyAndReceive handler in the ARM11 kernel)

# Notable changes
* HLE Interfaces are no longer kernel objects
* HLE Interfaces are now attached to ServerSessions, and are called during ServerSession::HandleSyncRequest
* HLE Interfaces now all inherit from SessionRequestHandler.
* GetServiceHandle and ConnectToPort now properly return ClientSession handles to the emulated application.

# Future changes
- [ ] Support session disconnection.
- [ ] Implement svcAcceptSession, it would only be used by the LLE services.
- [ ] Implement svcReplyAndReceive, it would only be used by the LLE services. This call works pretty much like WaitSynchronizationN.
- [ ] Rewrite how WaitSynchronizationN works. The current code is confusing and may lead to errors. (See #2260)
- [ ] Implement command buffer translation inside svcReplyAndReceive.
- [ ] Copy the command buffer to a local buffer before calling HandleSyncRequest in SessionRequestHandler.

### DIFF
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -7,16 +7,39 @@
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/server_port.h"
 #include "core/hle/kernel/server_session.h"
+#include "core/hle/service/service.h"
 
 namespace Kernel {
 
 ClientPort::ClientPort() {}
 ClientPort::~ClientPort() {}
 
+Kernel::SharedPtr<ClientPort> ClientPort::CreateForHLE(u32 max_sessions, std::unique_ptr<Service::Interface> hle_interface) {
+    SharedPtr<ClientPort> client_port(new ClientPort);
+    client_port->max_sessions = max_sessions;
+    client_port->active_sessions = 0;
+    client_port->name = hle_interface->GetPortName();
+    client_port->hle_interface = std::move(hle_interface);
+
+    return client_port;
+}
+
 void ClientPort::AddWaitingSession(SharedPtr<ServerSession> server_session) {
+    // A port that has an associated HLE interface doesn't have a server port.
+    if (hle_interface != nullptr)
+        return;
+
     server_port->pending_sessions.push_back(server_session);
     // Wake the threads waiting on the ServerPort
     server_port->WakeupAllWaitingThreads();
+}
+
+ResultCode ClientPort::HandleSyncRequest() {
+    // Forward the request to the associated HLE interface if it exists
+    if (hle_interface != nullptr)
+        return hle_interface->HandleSyncRequest();
+
+    return RESULT_SUCCESS;
 }
 
 } // namespace
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -13,10 +13,18 @@ namespace Kernel {
 ClientPort::ClientPort() {}
 ClientPort::~ClientPort() {}
 
-void ClientPort::AddWaitingSession(SharedPtr<ServerSession> server_session) {
+SharedPtr<ClientSession> ClientPort::Connect() {
+    // Create a new session pair, let the created sessions inherit the parent port's HLE handler.
+    auto sessions = ServerSession::CreateSessionPair(server_port->GetName(), server_port->hle_handler);
+    auto client_session = std::get<SharedPtr<ClientSession>>(sessions);
+    auto server_session = std::get<SharedPtr<ServerSession>>(sessions);
+
     server_port->pending_sessions.push_back(server_session);
+
     // Wake the threads waiting on the ServerPort
     server_port->WakeupAllWaitingThreads();
+
+    return std::move(client_session);
 }
 
 } // namespace
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -19,7 +19,8 @@ ResultVal<SharedPtr<ClientSession>> ClientPort::Connect() {
     // AcceptSession before returning from this call.
 
     if (active_sessions >= max_sessions) {
-        // TODO(Subv): Return an error code in this situation after session disconnection is implemented.
+        // TODO(Subv): Return an error code in this situation after session disconnection is
+        // implemented.
         /*return ResultCode(ErrorDescription::MaxConnectionsReached,
                           ErrorModule::OS, ErrorSummary::WouldBlock,
                           ErrorLevel::Temporary);*/
@@ -27,7 +28,8 @@ ResultVal<SharedPtr<ClientSession>> ClientPort::Connect() {
     active_sessions++;
 
     // Create a new session pair, let the created sessions inherit the parent port's HLE handler.
-    auto sessions = ServerSession::CreateSessionPair(server_port->GetName(), server_port->hle_handler);
+    auto sessions =
+        ServerSession::CreateSessionPair(server_port->GetName(), server_port->hle_handler);
     auto client_session = std::get<SharedPtr<ClientSession>>(sessions);
     auto server_session = std::get<SharedPtr<ServerSession>>(sessions);
 
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -14,7 +14,14 @@ namespace Kernel {
 ClientPort::ClientPort() {}
 ClientPort::~ClientPort() {}
 
-SharedPtr<ClientSession> ClientPort::Connect() {
+ResultVal<SharedPtr<ClientSession>> ClientPort::Connect() {
+    if (active_sessions >= max_sessions) {
+        return ResultCode(ErrorDescription::MaxConnectionsReached,
+                          ErrorModule::OS, ErrorSummary::WouldBlock,
+                          ErrorLevel::Temporary);
+    }
+    active_sessions++;
+
     // Create a new session pair, let the created sessions inherit the parent port's HLE handler.
     auto sessions = ServerSession::CreateSessionPair(server_port->GetName(), server_port->hle_handler);
     auto client_session = std::get<SharedPtr<ClientSession>>(sessions);
@@ -25,7 +32,7 @@ SharedPtr<ClientSession> ClientPort::Connect() {
     // Wake the threads waiting on the ServerPort
     server_port->WakeupAllWaitingThreads();
 
-    return std::move(client_session);
+    return MakeResult<SharedPtr<ClientSession>>(std::move(client_session));
 }
 
 } // namespace
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -19,9 +19,10 @@ ResultVal<SharedPtr<ClientSession>> ClientPort::Connect() {
     // AcceptSession before returning from this call.
 
     if (active_sessions >= max_sessions) {
-        return ResultCode(ErrorDescription::MaxConnectionsReached,
+        // TODO(Subv): Return an error code in this situation after session disconnection is implemented.
+        /*return ResultCode(ErrorDescription::MaxConnectionsReached,
                           ErrorModule::OS, ErrorSummary::WouldBlock,
-                          ErrorLevel::Temporary);
+                          ErrorLevel::Temporary);*/
     }
     active_sessions++;
 
@@ -29,6 +30,9 @@ ResultVal<SharedPtr<ClientSession>> ClientPort::Connect() {
     auto sessions = ServerSession::CreateSessionPair(server_port->GetName(), server_port->hle_handler);
     auto client_session = std::get<SharedPtr<ClientSession>>(sessions);
     auto server_session = std::get<SharedPtr<ServerSession>>(sessions);
+
+    if (server_port->hle_handler)
+        server_port->hle_handler->ClientConnected(server_session);
 
     server_port->pending_sessions.push_back(std::move(server_session));
 
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -15,6 +15,9 @@ ClientPort::ClientPort() {}
 ClientPort::~ClientPort() {}
 
 ResultVal<SharedPtr<ClientSession>> ClientPort::Connect() {
+    // Note: Threads do not wait for the server endpoint to call
+    // AcceptSession before returning from this call.
+
     if (active_sessions >= max_sessions) {
         return ResultCode(ErrorDescription::MaxConnectionsReached,
                           ErrorModule::OS, ErrorSummary::WouldBlock,
@@ -27,7 +30,7 @@ ResultVal<SharedPtr<ClientSession>> ClientPort::Connect() {
     auto client_session = std::get<SharedPtr<ClientSession>>(sessions);
     auto server_session = std::get<SharedPtr<ServerSession>>(sessions);
 
-    server_port->pending_sessions.push_back(server_session);
+    server_port->pending_sessions.push_back(std::move(server_session));
 
     // Wake the threads waiting on the ServerPort
     server_port->WakeupAllWaitingThreads();
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -6,10 +6,17 @@
 #include "core/hle/kernel/client_port.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/server_port.h"
+#include "core/hle/kernel/server_session.h"
 
 namespace Kernel {
 
 ClientPort::ClientPort() {}
 ClientPort::~ClientPort() {}
+
+void ClientPort::AddWaitingSession(SharedPtr<ServerSession> server_session) {
+    server_port->pending_sessions.push_back(server_session);
+    // Wake the threads waiting on the ServerPort
+    server_port->WakeupAllWaitingThreads();
+}
 
 } // namespace
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -14,21 +14,7 @@ namespace Kernel {
 ClientPort::ClientPort() {}
 ClientPort::~ClientPort() {}
 
-Kernel::SharedPtr<ClientPort> ClientPort::CreateForHLE(u32 max_sessions, std::shared_ptr<Service::Interface> hle_interface) {
-    SharedPtr<ClientPort> client_port(new ClientPort);
-    client_port->max_sessions = max_sessions;
-    client_port->active_sessions = 0;
-    client_port->name = hle_interface->GetPortName();
-    client_port->hle_interface = std::move(hle_interface);
-
-    return client_port;
-}
-
 void ClientPort::AddWaitingSession(SharedPtr<ServerSession> server_session) {
-    // A port that has an associated HLE interface doesn't have a server port.
-    if (hle_interface != nullptr)
-        return;
-
     server_port->pending_sessions.push_back(server_session);
     // Wake the threads waiting on the ServerPort
     server_port->WakeupAllWaitingThreads();
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -7,7 +7,6 @@
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/server_port.h"
 #include "core/hle/kernel/server_session.h"
-#include "core/hle/service/service.h"
 
 namespace Kernel {
 
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -14,7 +14,7 @@ namespace Kernel {
 ClientPort::ClientPort() {}
 ClientPort::~ClientPort() {}
 
-Kernel::SharedPtr<ClientPort> ClientPort::CreateForHLE(u32 max_sessions, std::unique_ptr<Service::Interface> hle_interface) {
+Kernel::SharedPtr<ClientPort> ClientPort::CreateForHLE(u32 max_sessions, std::shared_ptr<Service::Interface> hle_interface) {
     SharedPtr<ClientPort> client_port(new ClientPort);
     client_port->max_sessions = max_sessions;
     client_port->active_sessions = 0;
@@ -32,14 +32,6 @@ void ClientPort::AddWaitingSession(SharedPtr<ServerSession> server_session) {
     server_port->pending_sessions.push_back(server_session);
     // Wake the threads waiting on the ServerPort
     server_port->WakeupAllWaitingThreads();
-}
-
-ResultCode ClientPort::HandleSyncRequest() {
-    // Forward the request to the associated HLE interface if it exists
-    if (hle_interface != nullptr)
-        return hle_interface->HandleSyncRequest();
-
-    return RESULT_SUCCESS;
 }
 
 } // namespace
--- a/src/core/hle/kernel/client_port.cpp
+++ b/src/core/hle/kernel/client_port.cpp
@@ -4,6 +4,7 @@
 
 #include "common/assert.h"
 #include "core/hle/kernel/client_port.h"
+#include "core/hle/kernel/client_session.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/server_port.h"
 #include "core/hle/kernel/server_session.h"
--- a/src/core/hle/kernel/client_session.cpp
+++ b/src/core/hle/kernel/client_session.cpp
@@ -10,13 +10,22 @@
 namespace Kernel {
 
 ClientSession::ClientSession() = default;
-ClientSession::~ClientSession() = default;
+ClientSession::~ClientSession() {
+    // This destructor will be called automatically when the last ClientSession handle is closed by the emulated application.
 
-ResultVal<SharedPtr<ClientSession>> ClientSession::Create(SharedPtr<ServerSession> server_session, std::string name) {
+    if (server_session->hle_handler)
+        server_session->hle_handler->ClientDisconnected(server_session);
+
+    // TODO(Subv): If the session is still open, set the connection status to 2 (Closed by client),
+    // wake up all the ServerSession's waiting threads and set the WaitSynchronization result to 0xC920181A.
+}
+
+ResultVal<SharedPtr<ClientSession>> ClientSession::Create(ServerSession* server_session, std::string name) {
     SharedPtr<ClientSession> client_session(new ClientSession);
 
     client_session->name = std::move(name);
-    client_session->server_session = std::move(server_session);
+    client_session->server_session = server_session;
+    client_session->session_status = SessionStatus::Open;
     return MakeResult<SharedPtr<ClientSession>>(std::move(client_session));
 }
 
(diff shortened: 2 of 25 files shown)


Subv:
cc @Lectem

src/core/hle/kernel/server_session.h
```diff
@@ -0,0 +1,77 @@
...
+ * to the various system services.
+ *
+ * To make a service call, the client must write the command header and parameters to the buffer
+ * located at offset 0x80 of the TLS (Thread-Local Storage) area, then execute a SendSyncRequest
+ * SVC call with its ClientSession handle. The kernel will read the command header, using it to marshall
+ * the parameters to the process at the server endpoint of the session. After the server replies to
+ * the request, the response is marshalled back to the caller's TLS buffer and control is
+ * transferred back to it.
+ */
+class ServerSession : public WaitObject {
```
Subv: This class should contain a reference to the last thread that initiated a SyncRequest for the associated ClientSession so we can access it's TLS.

src/core/hle/service/service.cpp
```diff
@@ -61,7 +64,9 @@ static std::string MakeFunctionString(const char* name, const char* port_name,
     return function_string;
 }
 
-ResultVal<bool> Interface::SyncRequest() {
+ResultCode Interface::HandleSyncRequest(Kernel::SharedPtr<Kernel::ServerSession> server_session) {
+    // TODO(Subv): Make use of the server_session in the HLE service handlers to distinguish which session triggered each command.
```
Subv: How should we approach this TODO?

src/core/hle/service/fs/archive.cpp
```diff
@@ -236,10 +237,10 @@ ResultVal<bool> Directory::SyncRequest() {
         LOG_ERROR(Service_FS, "Unknown command=0x%08X!", cmd);
         ResultCode error = UnimplementedFunction(ErrorModule::FS);
         cmd_buff[1] = error.raw; // TODO(Link Mauve): use the correct error code for that.
-        return MakeResult<bool>(false);
```
Subv: Is this a  bug? in the File::SyncRequest default case we return the error instead of RESULT_SUCCESS.

src/core/hle/service/fs/archive.cpp
```diff
@@ -172,7 +173,8 @@ ResultVal<bool> File::SyncRequest() {
 
     case FileCommand::OpenLinkFile: {
         LOG_WARNING(Service_FS, "(STUBBED) File command OpenLinkFile %s", GetName().c_str());
```
Subv: I don't think this is a stub anymore, after this change

src/core/hle/service/service.cpp
```diff
@@ -75,14 +80,14 @@ ResultVal<bool> Interface::SyncRequest() {
...
-        return MakeResult<bool>(false);
+        return RESULT_SUCCESS;
     }
     LOG_TRACE(Service, "%s",
               MakeFunctionString(itr->second.name, GetPortName().c_str(), cmd_buff).c_str());
 
     itr->second.func(this);
 
-    return MakeResult<bool>(false); // TODO: Implement return from actual function
+    return RESULT_SUCCESS; // TODO: Implement return from actual function, it should fail if the parameter translation fails
```
Subv: The second part of this TODO should be handled by the Translate() call.

src/core/hle/service/service.h
```diff
@@ -9,29 +9,205 @@
...
+#include "core/hle/kernel/client_port.h"
+#include "core/hle/kernel/thread.h"
 #include "core/hle/result.h"
+#include "core/memory.h"
+
+namespace Kernel {
+class ServerSession;
+
+// TODO(Subv): Move these declarations out of here
+static const int kCommandHeaderOffset = 0x80; ///< Offset into command buffer of header
```
Subv: Where would be a good place to put these?

src/core/hle/service/service.h
```diff
@@ -9,29 +9,205 @@
...
+ * @return Pointer to command buffer
+ */
+inline u32* GetCommandBuffer(const int offset = 0) {
+    return (u32*)Memory::GetPointer(GetCurrentThread()->GetTLSAddress() + kCommandHeaderOffset +
+                                    offset);
+}
+}
+
+// TODO(Subv): Move this namespace out of here
+namespace IPC {
```
Subv: Where would be a good place to put these?

src/core/hle/service/service.h
```diff
@@ -9,29 +9,205 @@
...
+static const u32 DefaultMaxSessions = 10; ///< Arbitrary default number of maximum connections to an HLE port
+
+/// TODO(Subv): Write documentation for this class
+class SessionRequestHandler {
+public:
+    /**
+     * Dispatches and handles a sync request from the emulated application.
+     * @param server_session The ServerSession that was triggered for this sync request,
+     * it should be used to differentiate which client (As in ClientSession) we're answering to.
+     * TODO(Subv): Make a HandleSyncRequestParent function that is called from the outside and does { ReturnIfError(Translate()); HandleSyncRequest(); }
```
Subv: Thoughts on this? @Lectem @yuriks

src/core/hle/service/service.h
```diff
@@ -9,29 +9,205 @@
...
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Namespace Service
 
 namespace Service {
 
 static const int kMaxPortSize = 8; ///< Maximum size of a port name (8 characters)
+static const u32 DefaultMaxSessions = 10; ///< Arbitrary default number of maximum connections to an HLE port
+
+/// TODO(Subv): Write documentation for this class
```
Subv: I'm not sure what to write here.

src/core/hle/service/service.h
```diff
@@ -9,29 +9,205 @@
...
+     * @param server_session The ServerSession that was triggered for this sync request,
+     * it should be used to differentiate which client (As in ClientSession) we're answering to.
+     * TODO(Subv): Make a HandleSyncRequestParent function that is called from the outside and does { ReturnIfError(Translate()); HandleSyncRequest(); }
+     * The Translate() function would copy the command buffer from the ServerSession thread's TLS into a temporary buffer, and pass it to HandleSyncRequest.
+     * TODO(Subv): HandleSyncRequest's return type should be void.
+     */
+    virtual ResultCode HandleSyncRequest(Kernel::SharedPtr<Kernel::ServerSession> server_session) = 0;
+};
 
 /// Interface to a CTROS service
```
Subv: This comment is not descriptive at all, what would be a better docstring for this class?

src/core/hle/service/srv.cpp
```diff
@@ -81,7 +85,23 @@ static void GetServiceHandle(Service::Interface* self) {
...
+        // The hle_handler will be nullptr if this port was registered by the emulated
+        // application by means of srv:RegisterService.
+        auto hle_handler = std::get<std::shared_ptr<Service::Interface>>(it->second);
+
+        // Create a new session pair
+        auto sessions = Kernel::ServerSession::CreateSessionPair(port_name, hle_handler);
+        auto client_session = std::get<Kernel::SharedPtr<Kernel::ClientSession>>(sessions);
+        auto server_session = std::get<Kernel::SharedPtr<Kernel::ServerSession>>(sessions);
+
+        // TODO(Subv): Wait the current thread until the ServerPort calls AcceptSession.
```
Subv: After further inspecting of the ARM11 kernel code, it appears this call should not put the thread to sleep, more investigation is needed.

src/core/hle/service/fs/archive.cpp
wwylele: Not really sure what bug you are referring to... does real 3DS return RESULT_SUCCESS on unknown command?

Subv: No idea, but we do different things for unknown File commands and unknown Directory commands.

Subv:
It has been verified with a hardware test that GetServiceHandle and ConnectToPort do not put the calling thread to sleep until AcceptSession has been called.

src/core/hle/kernel/client_session.cpp
```diff
@@ -0,0 +1,28 @@
...
+// Refer to the license.txt file included.
+
+#include "common/assert.h"
+
+#include "core/hle/kernel/client_session.h"
+#include "core/hle/kernel/server_session.h"
+
+namespace Kernel {
+
+ClientSession::ClientSession() {}
```
lioncash: You can default these as:

```cpp
ClientSession::ClientSession() = default;
...
```

src/core/hle/kernel/server_session.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <tuple>
+
+#include "core/hle/kernel/client_session.h"
+#include "core/hle/kernel/server_session.h"
+#include "core/hle/kernel/thread.h"
+
+namespace Kernel {
+
+ServerSession::ServerSession() {}
```
lioncash: ditto on defaulting.

src/core/hle/service/fs/archive.cpp
```diff
@@ -92,7 +93,7 @@ File::File(std::unique_ptr<FileSys::FileBackend>&& backend, const FileSys::Path&
 
 File::~File() {}
 
-ResultVal<bool> File::SyncRequest() {
+ResultCode File::HandleSyncRequest(Kernel::SharedPtr<Kernel::ServerSession> server_session) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
```
Lectem: We should use the IPC helper from #2027 to wrap the command buffer, and make it accessible from the ServerSession, or (better I think?) simply replace the pointer to the session by the helper

src/core/hle/service/service.cpp
Lectem: Is it important to know what session triggered each command ? I thought that it was handled by creating a new thread and storing session related information in the TLS (I might be entirely off the mark for this one)

src/core/hle/service/service.h
Lectem: if #2027 gets merged, it will already be moved to ipc.h

src/core/hle/service/service.h
Lectem: I think it should be an internal of the IPC command helper

src/core/hle/service/service.h
Lectem: > it should be used to differentiate which client (As in ClientSession) we're answering to.

I think each server session should map to a TLS, and the service is responsible for choosing which one (by creating a thread) if it is needed.

src/core/hle/service/service.cpp
Subv: Multiple applications could be doing requests to the same service at the same time, we need to know which thread to awake and who we're responding to.
Real services are allowed to do whatever they want, most do spawn a new thread for each request, however others just handle them all in the same thread.

src/core/hle/service/service.h
Subv: See my previous response.

src/core/hle/kernel/client_port.cpp
```diff
@@ -6,10 +6,17 @@
...
 #include "core/hle/kernel/server_port.h"
+#include "core/hle/kernel/server_session.h"
 
 namespace Kernel {
 
 ClientPort::ClientPort() {}
 ClientPort::~ClientPort() {}
 
+void ClientPort::AddWaitingSession(SharedPtr<ServerSession> server_session) {
+    server_port->pending_sessions.push_back(server_session);
```
Subv: Should we check if the ServerSession has an associated hle_handler and not add it to the pending list? Or just add it anyways? (Will we ever use this list in HLE services?)

src/core/hle/service/service.cpp
yuriks: You can just stick the session data in a map (keyed by the pointer) to maintain state.

src/core/hle/service/service.h
yuriks: > TODO(Subv): Make a HandleSyncRequestParent function that is called from the outside and does { ReturnIfError(Translate()); HandleSyncRequest(); }

Nay. This class only applies to HLE stuff, and at least for the moment we won't be doing any translation on them. Real calls won't even go through here, so you'll need some other translation call else where. I think this should be handled on ServerSession::HandleSyncRequest or a similar place.

src/core/hle/service/service.h
```diff
@@ -9,29 +9,205 @@
...
+/// TODO(Subv): Write documentation for this class
+class SessionRequestHandler {
+public:
+    /**
+     * Dispatches and handles a sync request from the emulated application.
+     * @param server_session The ServerSession that was triggered for this sync request,
+     * it should be used to differentiate which client (As in ClientSession) we're answering to.
+     * TODO(Subv): Make a HandleSyncRequestParent function that is called from the outside and does { ReturnIfError(Translate()); HandleSyncRequest(); }
+     * The Translate() function would copy the command buffer from the ServerSession thread's TLS into a temporary buffer, and pass it to HandleSyncRequest.
+     * TODO(Subv): HandleSyncRequest's return type should be void.
```
yuriks: Is it too difficult to just do this right now? It's mostly just deleting lines from a few places you're already modifying in this PR anyway.

src/core/hle/service/service.h
```diff
@@ -9,29 +9,205 @@
...
+     */
+    virtual ResultCode HandleSyncRequest(Kernel::SharedPtr<Kernel::ServerSession> server_session) = 0;
```
yuriks: I recommend making a struct (`RequestParams` or so) to hold the parameters. That way we can expand it in the future without having to go back and fix every single handler to handle the new parameters. (This can in the future hold stuff such as the translated cmdbuf, the originating thread, etc.)

src/core/hle/service/service.h
yuriks: > Interface implemented by HLE Session handlers. This can be provided to a Session in order to hook into several relevant events (such as a new connection or a SyncRequest) so they can be implemented in the emulator.

...or something like that?

src/core/hle/service/service.h
```diff
@@ -9,29 +9,205 @@
...
         version.raw = raw_version;
     }
+    virtual ~Interface() {}
+
+    /**
+     * Gets the maximum allowed number of sessions that can be connected to this port at the same time.
+     * It should be overwritten by each service implementation for more fine-grained control.
+     * @returns The maximum number of connections allowed.
+     */
+    virtual u32 GetMaxSessions() const { return DefaultMaxSessions; }
```
yuriks: No reason to make this a virtual function, just make it a constructor parameter so subclasses can set it.

src/core/hle/service/service.h
yuriks: > Framework for implementing HLE service handlers which dispatch incoming SyncRequests based on a table mapping header ids to handler functions. *etc.* *etc.*

src/core/hle/service/srv.cpp
```diff
@@ -81,7 +85,24 @@ static void GetServiceHandle(Service::Interface* self) {
...
+        // Create a new session pair
+        auto sessions = Kernel::ServerSession::CreateSessionPair(port_name, hle_handler);
+        auto client_session = std::get<Kernel::SharedPtr<Kernel::ClientSession>>(sessions);
+        auto server_session = std::get<Kernel::SharedPtr<Kernel::ServerSession>>(sessions);
+
+        // Note: Threads do not wait for the server endpoint to call
+        // AcceptSession before returning from this call.
+
+        // Add the server session to the port's queue
+        client_port->AddWaitingSession(server_session);
```
yuriks: Lot's of code duplicated between here and the ConnectToPort svc. See comment in ClientPort::AddWaitingSession.

src/core/hle/service/service.h
Subv: The translation (both client-server and server-client) has to happen during svcReplyAndReceive for non-HLE stuff.

I think the translation codepaths for HLE and non-HLE are different enough to warrant separating them and calling the HLE version from here

src/core/hle/kernel/server_session.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+void ServerSession::Acquire() {
+    ASSERT_MSG(!ShouldWait(), "object unavailable!");
+    signaled = false;
+}
+
+ResultCode ServerSession::HandleSyncRequest() {
+    // The ServerSession received a sync request, this means that there's new data available
+    // from its ClientSession, so wake up any threads that may be waiting on a svcReplyAndReceive or similar.
```
Subv: It's just a description of the function.
We received data, so wake the waiting threads

src/core/hle/kernel/server_session.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    // If this ServerSession has an associated HLE handler, forward the request to it.
+    if (hle_handler != nullptr)
+        return hle_handler->HandleSyncRequest(SharedPtr<ServerSession>(this));
+
+    // If this ServerSession does not have an HLE implementation, just wake up the threads waiting on it.
+    signaled = true;
+    WakeupAllWaitingThreads();
```
Subv: Yep

src/core/hle/kernel/server_session.cpp
```diff
@@ -0,0 +1,56 @@
...
+    return RESULT_SUCCESS;
+}
```
Subv: This function doesnt have any missing functionality that i can think of right now, what do you mean?

Subv:
Addressed some comments.
The Session disconnection logic is still todo, but I think that can be done in a follow up PR so as to not make this one bigger than it already is.

src/core/hle/kernel/server_port.cpp
```diff
@@ -24,12 +24,13 @@ void ServerPort::Acquire() {
...
 
 std::tuple<SharedPtr<ServerPort>, SharedPtr<ClientPort>> ServerPort::CreatePortPair(
-    u32 max_sessions, std::string name) {
+    u32 max_sessions, std::string name, std::shared_ptr<Service::SessionRequestHandler> hle_handler) {
 
     SharedPtr<ServerPort> server_port(new ServerPort);
     SharedPtr<ClientPort> client_port(new ClientPort);
 
     server_port->name = name + "_Server";
+    server_port->hle_handler = hle_handler;
```
yuriks: ~ std::move ~

src/core/hle/kernel/server_session.cpp
yuriks: Hm, yeah, not sure what I was thinking before, since you do wake up the threads. nvm

src/core/hle/kernel/server_session.cpp
```diff
@@ -0,0 +1,56 @@
...
+    signaled = false;
+}
+
+ResultCode ServerSession::HandleSyncRequest() {
+    // The ServerSession received a sync request, this means that there's new data available
+    // from its ClientSession, so wake up any threads that may be waiting on a svcReplyAndReceive or similar.
+
+    // If this ServerSession has an associated HLE handler, forward the request to it.
+    if (hle_handler != nullptr)
+        return hle_handler->HandleSyncRequest(SharedPtr<ServerSession>(this));
```
yuriks: Is there any reason why you shouldn't also signal this session if it's HLE?

src/core/hle/kernel/server_session.cpp
yuriks: nvm

src/core/hle/kernel/server_session.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    // If this ServerSession does not have an HLE implementation, just wake up the threads waiting on it.
+    signaled = true;
+    WakeupAllWaitingThreads();
+    return RESULT_SUCCESS;
+}
+
+std::tuple<SharedPtr<ServerSession>, SharedPtr<ClientSession>> ServerSession::CreateSessionPair(const std::string& name, std::shared_ptr<Service::SessionRequestHandler> hle_handler) {
```
yuriks: Overlong line

src/core/hle/kernel/server_session.cpp
```diff
@@ -0,0 +1,56 @@
...
+    auto server_session = ServerSession::Create(name + "_Server", hle_handler).MoveFrom();
```
yuriks: move hle_handler

src/core/hle/service/service.h
yuriks: Are you keeping these in this file after all?

src/core/hle/service/fs/archive.cpp
yuriks: Are you removing it?

src/core/hle/service/fs/archive.h
```diff
@@ -41,33 +41,37 @@ enum class MediaType : u32 { NAND = 0, SDMC = 1 };
 
 typedef u64 ArchiveHandle;
 
-class File : public Kernel::Session {
+class File : public SessionRequestHandler, public std::enable_shared_from_this<File> {
```
yuriks: Mark as final

src/core/hle/service/fs/archive.h
```diff
@@ -41,33 +41,37 @@ enum class MediaType : u32 { NAND = 0, SDMC = 1 };
...
     FileSys::Path path; ///< Path of the file
     u32 priority;       ///< Priority of the file. TODO(Subv): Find out what this means
     std::unique_ptr<FileSys::FileBackend> backend; ///< File backend interface
+
+protected:
+    void HandleSyncRequestImpl(Kernel::SharedPtr<Kernel::ServerSession> server_session) override;
 };
 
-class Directory : public Kernel::Session {
+class Directory : public SessionRequestHandler {
```
yuriks: Mark as final

src/core/hle/service/service.h
yuriks: This will be used by the kernel translation code too, so it should stay in a more generic file.

src/core/hle/service/srv.cpp
```diff
@@ -81,7 +85,20 @@ static void GetServiceHandle(Service::Interface* self) {
     auto it = Service::g_srv_services.find(port_name);
 
     if (it != Service::g_srv_services.end()) {
-        cmd_buff[3] = Kernel::g_handle_table.Create(it->second).MoveFrom();
+        auto client_port = it->second;
+
+        // Note: Threads do not wait for the server endpoint to call
+        // AcceptSession before returning from this call.
```
yuriks: Noticed that this comment is duplicated in two places. Maybe move it to the Connect function.

src/core/hle/service/srv.cpp
```diff
@@ -81,7 +85,20 @@ static void GetServiceHandle(Service::Interface* self) {
...
+
+        // Connect to the port and retrieve the client endpoint of the connection Session.
```
yuriks: Seems like a kinda superfluous comment, it's just re-stating the code line.

src/core/hle/svc.cpp
```diff
@@ -222,20 +224,31 @@ static ResultCode ConnectToPort(Handle* out_handle, const char* port_name) {
...
+
+    // Note: Threads do not wait for the server endpoint to call
+    // AcceptSession before returning from this call.
+
+    // Return the client session
+    CASCADE_RESULT(*out_handle, Kernel::g_handle_table.Create(client_session));
     return RESULT_SUCCESS;
 }
 
 /// Synchronize to an OS service
```
yuriks: "Makes a blocking IPC call to an OS service."

src/core/hle/service/service.cpp
```diff
@@ -61,7 +64,29 @@ static std::string MakeFunctionString(const char* name, const char* port_name,
...
+    return RESULT_SUCCESS;
+}
+
+ResultCode SessionRequestHandler::TranslateRequest(Kernel::SharedPtr<Kernel::ServerSession> server_session) {
+    // TODO(Subv): Implement this function once multiple concurrent processes are supported.
+    return RESULT_SUCCESS;
+}
+
+void Interface::HandleSyncRequestImpl(Kernel::SharedPtr<Kernel::ServerSession> server_session) {
+    // TODO(Subv): Make use of the server_session in the HLE service handlers to distinguish which session triggered each command.
```
yuriks: Wouldn't this be more of a per-service level concern? Each one might want to do different things, so this doesn't seem like an appropriate place for this todo.

src/core/hle/service/fs/archive.cpp
Subv: I'd rather wait until we actually reverse the function to see what it does.

src/core/hle/service/service.h
Subv: No, I'll move them to an ipc.h header

src/core/hle/kernel/server_session.cpp
Subv: Nothing would be waiting on the ServerSession if it's HLE, it wouldn't even be exposed to the running application, there's no point in doing that (I think?)

src/core/hle/service/service.h
```diff
@@ -9,30 +9,235 @@
...
+ */
+class SessionRequestHandler {
+public:
+    /**
+     * Dispatches and handles a sync request from the emulated application.
+     * @param server_session The ServerSession that was triggered for this sync request,
+     * it should be used to differentiate which client (As in ClientSession) we're answering to.
+     * @returns ResultCode the result code of the translate operation.
+     */
+    ResultCode HandleSyncRequest(Kernel::SharedPtr<Kernel::ServerSession> server_session);
```
Subv: I don't agree, where would we put it? IMO this class should take care of the translation for all HLE IPC requests, and just hand a translated buffer to the implementation.

src/core/hle/kernel/server_session.cpp
yuriks: But is there any reason not to do it? That's one less difference in case we ever move HLE to be more process-based instead of callback based.

src/core/hle/service/service.cpp
```diff
@@ -84,6 +84,9 @@ ResultCode SessionRequestHandler::TranslateRequest(Kernel::SharedPtr<Kernel::Ser
     return RESULT_SUCCESS;
 }
 
+Interface::~Interface() = default;
+Interface::Interface(u32 max_sessions) : max_sessions(max_sessions) { }
```
yuriks: Constructor goes before the destructor

Subv:
@yuriks Do you have anything against the ClientConnected / ClientDisconnected callbacks? I thought you were in favor of this idea.

src/core/hle/kernel/server_session.cpp
Subv: Done~~

Subv:
A @yuriks LGTM means this thing is out of RFC state \o/

Subv:
Formatting errors have been fixed, can this be merged?

PR #2253, opened 2016-12-02
moved Region option to System tab

NicolaiVdS:
The Region option is now in the system tab 
Region option in System tab


src/citra_qt/configure_system.cpp
```diff
@@ -46,6 +47,7 @@ void ConfigureSystem::setConfiguration() {
         ReadSystemSettings();
         ui->label_disable_info->hide();
     }
+    ui->combo_region-setCurrentIndex(Settings::values.region_value);
```
wwylele: ->setCurrentIndex

src/citra_qt/configure_general.ui
```diff
@@ -43,83 +43,20 @@
...
-            </item>
-          </layout>
-        </widget>
-      </item>
      <item>
-      <widget class="QGroupBox" name="groupBox_4">
+      <widget class="QGroupBox" name="groupBox_2">
        <property name="title">
-        <string>Emulation</string>
+        <string>Performance</string>
```
Subv: Why was this renamed to Performance?

wwylele: Is this intended? Should mention this change in the PR description.

src/citra_qt/configure_system.cpp
```diff
@@ -46,6 +47,7 @@ void ConfigureSystem::setConfiguration() {
         ReadSystemSettings();
         ui->label_disable_info->hide();
     }
+    ui->combo_region->setCurrentIndex(Settings::values.region_value);
```
Subv: Weren't we updating the combobox and the region_value in the Settings already?

wwylele:
I don't have strong opinion about this change. However I am worried about this: the system settings itself also has "region/location" settings (countries and subregions), which is just not implemented yet. Moving the console firm region to system settings leads to ambiguity, especially when we add user location setting in the future.

src/citra_qt/configure_general.ui
NicolaiVdS: @wwylele and @Subv  that Emulation/Performance was made by QT Creator when i removed the Emulation section in the general tab, if you take a look at the master the CPU JIT option is called Performance

yuriks:
Kinda have the same concern as @wwylele. The "User Profile" region and the console region are two different things and we should make that clear. We should add some separation between the user profile data and other misc. data that's stored in the config file. Does anyone know where the console region is stored in the 3DS? Is it a hardware fuse somewhere or is it just a software thing?

Subv:
@yuriks It's stored in the [SecureInfo_A file](https://www.3dbrew.org/wiki/Nandrw/sys/SecureInfo_A)

lefta:
I agree with @NicolaiVdS that the console region is not at its place in the `general` tab, the settings found here are general emulator settings. But I agree with @yuriks and @wwylele that it does not have its place in the `system settings` part as the console region is not part of the real console available settings.

A new frame could be added in the `system` tab, like `console settings`, where the console region would be configured. This frame would also contain the other console built-in settings, if any and restrict the `system settings` frame for the settings configurable on the real console.

NicolaiVdS:
if every one else agrees with @lefta that there should be a a new frame in the `system` tab called `console settings` i can add that it in

yuriks:
I would rename the current frame to "User Profile" or something similar, and then make the new frame be "System Settings". (I think you could move the headphone setting there too.) other than this, that approach SGTM

NicolaiVdS:
added the second frame and moved the 2 options to it 2 frames

mailwl:
sound mode in system, when Audio tab already exists... not obviously, at my sight

NicolaiVdS:
@mailwl there is a audio option in the `system` frame to configure how the system should play the sound ` (mono, stereo or surround) ` and then there is the audio tab that determents what `audio engine` `citra` is going to use, the one is handled by the `emulated 3DS` and the other by `citra`

wwylele:
@NicolaiVdS @mailwl Here is my original thought why I put "Sound output mode" along with "User name" etc: sound output mode is an changeable option in real 3DS System Settings application and is stored in config, which is unlike firmwire region which is not changeable and is not in config, and is unlike options in audio tab which are options for audio frontend instead. I know it is probably not the best layout for these options, but this is the info you can take into consideration.

Dragios:
*User Name/Birthday/Region* is under one category which is Profile in real 3DS
Sound output mode is under *Sound* section in real 3DS.
Language is no where to be found as it is setup initially when you first purchased the 3DS.

I think it will be best to categorised them according to the real console.

lefta:
@Dragios: In this case, we are not talking about the 3DS region setting, but the console built-in region (EUR/U/J) which may not be changed. But I agree with you that settings should keep the real console categories, with a new category for BIOS settings (which obviously do not exist on the real console settings).

Dragios:
Ok, looks like I was confused.
I suggest this instead
```
[Console Type]
Region:

[Profile]
User Name:
Birthday:

[Sound]
Output mode:

[Mic Test]
Sensitivity: //Unimplemented

lefta:
Some modding:

    [User profile]
    User name:
    Birthday:
    
    [Sound]
    Output mode:
    
    [Mic test]
    Sensitivity:
    
    [Console/System settings] // Whatever you think fits better
    Region:
    Language:

Dragios:
@lefta I prefer the Console settings to be at the top but yours can also be a good alternative.

lefta:
Just a matter of preference, personally I don't care. Why not a setting to set this? (just joking...)
My mod was to add the missing `language` setting and reword some sections.

NicolaiVdS:
well i can be wrong but does the 3ds has diferent Lang options depending on the region ?

mailwl:
sure, but Citra is not 1:1 program clone of 3DS

lefta:
I don't think it matters. Why not having French language in an USA console, even if the USA 3DS don't have it?

JayFoxRox:
I think the data should be in order of hierarchy. Meaning system settings set at time of fabrication first. All of those in the same box unless they are strongly tied to either Audio, Graphics or Input. Then user profile and all other settings configurable by users.

Subv:
Can we get over with the bikeshedding and finally decide on this please?

lefta:
Let me update my proposal based on comments so far (with only already implemented settings):

    [Console settings]
    Region:
    Language:

    [User profile]
    Name:
    Birthday:
    
    [Sound]
    Output mode:

Does everybody agree on it?

Dragios:
Preferably if you use `User Name` instead as it follows the naming scheme in the actual console.
Other than that, LGTM

lefta:
My rationale was that `user name` in `user profile` was redundant, but let's keep original 3DS settings wording:
    [Console settings]
    Region:
    Language:
    
    [User profile]
    User name:
    Birthday:
    
    [Sound]
    Output mode:

Dragios:
If you think is redundant you can name the section as `Profile` only instead as it follows the actual console naming scheme.

lefta:
[Console]
    Region:
    Language:
    
    [Profile]
    User name:
    Birthday:
    
    [Sound]
    Output mode:

Dragios:
Should be good to go. Anyway this are all small nitpicks that can be change in future PR, if needed.

NicolaiVdS:
Il update the PR acording to that latest scheme

bunnei:
Reminder to update this / remove merge conflicts.

wwylele:
What's the status of this PR?

yuriks:
Closing due to inactivity. If you want to continue working on this post a message here and we'll re-open it.

Issue #2226, opened 2016-11-28
Cut the Rope: Triple Treat ‒ Crash with Shader JIT on


Dragios:
Now that #2168 has been merged the game will only load when *Shader JIT* is disabled. Enabling *Shader JIT* will result in a crash.

Related to #2138 

*Crash Log*
![crash log](https://cloud.githubusercontent.com/assets/18496868/20666561/bf2da8c8-b59f-11e6-803a-79c5d10aa2c0.png)

Tested on the latest commit as of 3174bfd using my own build.

Subv:
That's not a crash log, what's your OS and compiler? Please provide a backtrace.

Dragios:
OS: Windows 10 build 14393.447
Compiler: Both MSVC and MinGW

![Output](https://cloud.githubusercontent.com/assets/18496868/20671774/a9adc8e2-b5b9-11e6-9656-7b93b57b7007.png)
![Call stack](https://cloud.githubusercontent.com/assets/18496868/20671775/aa8cd8d4-b5b9-11e6-89cc-e693c8c9ea22.png)
![Disassembly](https://cloud.githubusercontent.com/assets/18496868/20671778/ac1bc642-b5b9-11e6-9315-461b8f54bb04.png)

JayFoxRox:
Fixed by #2255

Issue #2209, opened 2016-11-24
Pokemon sun graphical errors related to shaderJIT


wakka1:
Specs : 
amd HD 6790
Phenom ii X4 955 3.20 ghz
6gb ram
windows 7 64bit
Tested with master d3b7f57
Edit : Forgot to mention in first post that the issue is most likely related to certain hardware as i saw some yotube footage working fine for other people.

With ShaderJIT: broken text box and visual effects
![citra-qt 2016-11-27 14-52-37-44](https://cloud.githubusercontent.com/assets/22716011/20648842/dee82be6-b4b1-11e6-9076-a3ae82f6c92b.jpg)

![citra-qt 2016-11-24 13-44-44-45](https://cloud.githubusercontent.com/assets/22716011/20599435/6d57edf0-b24e-11e6-8bf4-c9d28a1675e1.jpg)
![citra-qt 2016-11-24 13-45-04-78](https://cloud.githubusercontent.com/assets/22716011/20599466/90ac77ee-b24e-11e6-85fc-d20e872c5d4e.jpg)

Without ShaderJIT: works fine
![citra-qt 2016-11-27 14-52-46-94](https://cloud.githubusercontent.com/assets/22716011/20648843/e7375344-b4b1-11e6-9f4a-0160f7f0465b.jpg)

![citra-qt 2016-11-24 13-47-45-31](https://cloud.githubusercontent.com/assets/22716011/20599481/a173da68-b24e-11e6-85cb-4d844592aaa8.jpg)
![citra-qt 2016-11-24 13-48-31-26](https://cloud.githubusercontent.com/assets/22716011/20599497/a7e4ca56-b24e-11e6-8368-c9b612935082.jpg)

Log: 
[error log.txt](https://github.com/citra-emu/citra/files/614729/error.log.txt)

JayFoxRox:
Please re-test on current master. There is no support for bleeding edge and there are shader related features in bleeding edge.

wakka1:
Sorry for that, i do not have any nightly build and never built  one, i'll see what  i can do.

wwylele:
Tested on current master with shader JIT on and doesn't encounter such problem. Probably an issue in GS.

wakka1:
Citra crashes when i load a game.  Looks like i couldn't build it right.
[...]

Issue #1743, opened 2016-04-30
Harvest Moon The Lost Valley Texture and Freeze Problem


amirlou:
hello so after testing this game with current master i found some problems
1-game got bad texture and buggy screen with jit-Enable
2-game Freeze after first time sleep in game and show only black screen

here is some pic with jit-enable

![xx2 2016-05-01 01-27-25-71](https://cloud.githubusercontent.com/assets/16171668/14938850/04f25d40-0f47-11e6-8edd-635a435afeb5.jpg)

this one with Jit-Disable

![xx2 2016-05-01 01-27-35-33](https://cloud.githubusercontent.com/assets/16171668/14938852/137a9620-0f47-11e6-9460-9103859815b9.jpg)

logs here
https://jpst.it/I1oa

amirlou:
just tested this game with current Master (DSP HLE)  game just got graphics problem with jit rest is ok i play it for an hour

JayFoxRox:
Unable to reproduce on 86ecbdf + JIT:
![master](http://i.imgur.com/DYaY62F.png)

What CPU do you use?
ftr: I'm on 'model name	: Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz' (linux users can use `cat /proc/cpuinfo`)

The panel is drawn in 4 draw calls. I've dumped the shader just-in case but won't upload it unless necessary

amirlou:
im using a old cpu AMD Athlone(tm) II x2 240 processor 3.50ghz

JohnGodgames:
It's ok on my side (JIT enabled):

![harvest valley](https://cloud.githubusercontent.com/assets/11453507/15248236/655f9bda-1919-11e6-8246-727606fa812f.JPG)

CPU - Intel Core i7 6700K @ 4.60 Ghz
RAM - 16 GB DDR4
GPU - GTX 970 GAMING 100ME 4GB
OS  - Windows 10 64bit

JayFoxRox:
Please re-test now. #2255 is merged

PR #2255, opened 2016-12-03
shader_jit: Load LOOPCOUNT_REG and LOOPINC 4 bit left-shifted

### LINKED ISSUE #1629: Games that crash when Shader JIT is enabled

JayFoxRox:
As noticed by @Subv and pointed out by @yuriks in citra-emu#1629 (comment) , our shader-JIT had some serious issue where the LOOP address register was used directly in offset calculations.
As each addressed vector is 16-bytes in size, this PR changes the JIT registers by keeping their values shifted 4 bits to the left (effectively keeping these regs multiplied by 16 at all times).

The offset used in address calculations is stored in LOOPCOUNT_REG. It's incremented by LOOPINC.
Hence both registers were shifted.

This fixes #1629 and #2226.

Additionally the indexed lookup is performed using MOVAPS which requires the memory address to be 16 byte aligned. As we didn't satisfy that condition with the old code, we probably read garbage. This might have been responsible for #1743 and #2209 (Issues on certain AMD Phenom CPUs). I did not get any confirmation for this yet though, so I'd appreciate further tests by people affected by those 2 issues.

### DIFF
--- a/src/video_core/shader/shader_jit_x64.cpp
+++ b/src/video_core/shader/shader_jit_x64.cpp
@@ -102,11 +102,11 @@ static const X64Reg SETUP = R9;
 /// The two 32-bit VS address offset registers set by the MOVA instruction
 static const X64Reg ADDROFFS_REG_0 = R10;
 static const X64Reg ADDROFFS_REG_1 = R11;
-/// VS loop count register
+/// VS loop count register (Multiplied by 16)
 static const X64Reg LOOPCOUNT_REG = R12;
 /// Current VS loop iteration number (we could probably use LOOPCOUNT_REG, but this quicker)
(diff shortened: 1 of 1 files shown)


Subv:
LGTM

yuriks:
🚢 it

PR #2257, opened 2016-12-03
Travis: Use a stable version of clang-format

yuriks:
We were using a SVN version of clang-format (due to the stable repo not being whitelisted by travis) which may change randomly and break our formatting/build. Changed to manually install a stable version using sudo.

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -12,7 +12,7 @@ fi
 # Only run clang-format on Linux because we don't have 4.0 on OS X images
 if [ "$TRAVIS_OS_NAME" = "linux" ]; then
     # Default clang-format points to default 3.5 version one
-    CLANG_FORMAT=clang-format-4.0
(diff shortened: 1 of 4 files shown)


JayFoxRox:
Does this put us on slower build machines, or were we limited to slow machines before anyway? (due to `sudo: true` or something?)

jroweboy:
@JayFoxRox We already were on the full VM instead of the container. This does add some time to the build, but its worth it to not have new clang format changes breaking things that were previously formatted by clang format. (For an example see https://github.com/citra-emu/citra/pull/2194 for the commit that this PR also reverts)

LGTM

yuriks:
Yeah, I'm pretty sure we were already on the VM infrastructure before. And anyway, it's still a bit faster than the OS X build, so whatever. (As an aside, how did we get to the point where a Windows VM is faster to start up and finish the build than the Linux ones?)

Issue #1904, opened 2016-06-12
Unimplemented pulse event/timer


wwylele:
There are three reset type for event and timer: `OneShot`, `Sticky` and `Pulse`. The first two types are well-implemented in citra, but the third one is not.

[Here is a hwtest](https://github.com/wwylele/ctrhwtest/blob/master/event_timer_type/source/main.cpp) for different behavior of each types. Generally speaking, the behavior for three types are (take event for instance)
- `OneShot` event will not be reset once signaled until one thread acquires it. If an event is signaled when there are several threads waiting on it, exactly one thread will resume and reset the event.  If an event is signaled when there is no thread waiting on it, it will remain signaled until one thread try waiting on it and reset itself then.
- `Sticky` event will not be reset until `ClearEvent` is called on it. If an event is signaled when there are several threads waiting on it, all of these threads will resume and remain the event signaled.
- `Pulse` event will be "immediately reset" after signaled. If an event is signaled when there are several threads waiting on it, **all** of these threads will resume and the event **will be reset**. If an event is signaled when there is no thread waiting on it, the event **will also reset itself as if nothing happens** 

Timer is basically an event that is scheduled to be signaled in a certain time, and follows the similar reset behavior as event.

Currently citra treats `Pulse` event as `OneShot`, and `Pulse` timer as `Sticky`.

Issue #1308, opened 2015-12-30
WaitSynchronizationN should (probably) not acquire Semaphores more than once


Subv:
If you call WaitSynchronizationN with N handles all pointing to the same Semaphore object, Citra will currently acquire the semaphore N times, which may result in the Semaphore being locked.

We should verify if this is indeed hardware behavior.

wwylele:
(If I understand this correctly)
Hardware behaviour verified: is the same as citra's behaviour. This can be closed.
[Here is the test](https://github.com/wwylele/ctrhwtest/tree/master/semaphore_acquire_n)

Subv:
You set the semaphore to have 10 free slots, but only acquire it 3 times, the problem manifests when you acquire it more times than are available in the semaphore.

wwylele:
@Subv updated the test ([3ef30a5](https://github.com/wwylele/ctrhwtest/tree/3ef30a59b792122bbfc844f3486ec879c82b5b0e/semaphore_acquire_n)). The last test is acquire the semaphore 3 times when the available count is 2. The behaviour of 3DS and citra is still the same: didn't time out, decrease the count to zero. Citra's code just magically works.

Subv:
Why does it work?

wwylele:
citra's code logic for WaitSynchN-wait-all is like: check all objects, if all of them are available, then acquire *each object that is available* ; otherwise wait the thread. So it works

Subv:
Could you upload a 3dsx of this test please?

wwylele:
@Subv https://www.dropbox.com/s/2blq0qptizagi08/semaphore_acquire_n.3dsx?dl=1

Subv:
I see what happens, for the last Acquire() call, ShouldWait() returns true, thus the call is not performed, leaving the available_count of the semaphore in 0 instead of underflowing.
This does not make much sense, but then again, passing the same semaphore multiple times to WaitSynchN doesn't make much sense either, and this is apparently the same behavior as the hardware.

I'll just close this.

PR #2260, opened 2016-12-04
Threading: Reworked the way our scheduler works.

### LINKED ISSUE #2026: Pokemon Alpha Sapphire-Black screen depending on the save file language
So as i downloaded the latest nightly that has CRO implemented i put a save file for Pokemon Alpha Sapphire in the right folder and started the game.The game starts to the main menu(where i see my save) and when i click to load my game save screen goes black.So i found the fix to change save file language to German and i did.And the crash is fixed i can now play the game.My question is what is the difference between English and German version,i mean that's just the translation,right?And the Log is a little different for both German and English.

German log: http://pastebin.com/0byUsgYC
English log: http://pastebin.com/8PgqhJYG

My specs are:
i5 4200U
4Gb DDr3
GT740M
Windows 10 64-bit

Maybe there is an issue in shared_font.bin or something like that.What do you guys think what is the issue?
P.S. The issue doesn't appear on the Pokemon ROM hacks like star saphhire only on the original game.

[0011c500(eng).zip](https://github.com/citra-emu/citra/files/442277/0011c500.eng.zip)
[0011c500(german).zip](https://github.com/citra-emu/citra/files/442278/0011c500.german.zip)

### LINKED ISSUE #2028: Pokemon alpha sapphire pokecenter freeze
i have been having this problem with the unofficial build and now that it is official i am posting it here. as i have only experience this freeze at the pokemon center, it may be related to other freezes found this game and omega ruby.
thee freeze occurs anything you try to enter a pokecenter with a female character, ie when playing as May.screen shots here

screenshot 11

for console log 
http://pastebin.com/MUQUUtxN

i am using master build , git commit 3f515c0 

for steps ;
create or load a save with the female trainers.
go to the nearest pokecenter 
try entering.
it would freeze during the blacking out sequence, when the pokeball inner circle almost touches the head of your trainer.
PC specs
Intel core i5 4210u @ 2.4ghz ~ 2.7ghz
nvidia geforce 840m 2gb
6 gb ddr3 ram

i dont know how i can get more info for the issue report, it does not crash so no backtrace.

### LINKED ISSUE #2041: Pokemon Omega Ruby & Alpha Sapphire Issues

### LINKED ISSUE #2091: Zelda: ALBW - Savegame doesn't load properly on interpreter
(Reported by PPLToast on IRC, verified by myself.)
Version tested: 8bf09b5302dc7717477f6c3b5c1f0ea8f2c80717

This save game [000ec400.zip](https://github.com/citra-emu/citra/files/479228/000ec400.zip) loads normally when using the CPU JIT. When using the interpreter, it loads as if you had 999 rupees, all inventory items and no discovered weather vanes or map markers. Even more strangely, ~~if you save the game none of the changes will be made permanent~~ (EDIT: It wasn't actually saving, see comment below.), if you reload using the JIT, you'll be back to your old items and story position.

PPLToast speculates that this might be a debug feature being activated somehow.

### LINKED ISSUE #2119: Digimon World Re:Digitize Decode - Moflex Video Loading Black screen
This bug seems to be rather a regression but I was unable to pinpoint it exactly since I can't seem to find the build where it used to work:

With the current build of Citra, no additional merges, the game will freeze to a black screen when trying to play the intro of the game or when starting a new game. 

This happens in both cases because a .moflex video is about to be loaded. 

In past builds (I really can't tell which one, it's been months)) it was at least possible to skip the intros by pressing start, in spite of the black screen. But now, pressing start would do nothing and the game would simply freeze on starting a new game or starting the game without having a savegame present (Since having a savegame present will skip the first moflex video).

Now I'm not sure what caused that, or whether that is relevant at all, but I wanted to report it.

Unfortunately the log contains no useful information at all, literally nothing, so there's hardly any point pasting it or the black screen. 

Perhaps someone with a deeper understanding of the root of this problem might look into this.
I guess simplied the problem could be summarized to: Moflex Video Loading is freezing the game/emulator to a black screen.

### LINKED ISSUE #2125: Asynchronous file reading softlocks Brunswick Pro Bowling

Subv:
Threads will now be awakened when the objects they're waiting on are signaled, instead of repeating the WaitSynchronization call every now and then.

The scheduler is now called once after every SVC call, and once after a thread is awakened from sleep by its timeout callback.

This new implementation is based off reverse-engineering of the real kernel.

See [here](https://gist.github.com/Subv/02f29bd9f1e5deb7aceea1e8f019c8f4) for a more detailed description of how the real kernel handles rescheduling.

Fixes #2125 even with @neobrain's changes.
Fixes #2091
Fixes #2119 
Fixes #2028 
Fixes #2026
Fixes #2041

These changes provide a more stable framework for implementing #1904. I actually have an implementation in mind already, following what the real kernel does.

This PR will make it much easier to implement svcReplyAndReceive after #2249 is merged.

### DIFF
--- a/src/citra_qt/debugger/wait_tree.cpp
+++ b/src/citra_qt/debugger/wait_tree.cpp
@@ -230,7 +230,7 @@ std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThread::GetChildren() const {
         list.push_back(std::make_unique<WaitTreeMutexList>(thread.held_mutexes));
     }
     if (thread.status == THREADSTATUS_WAIT_SYNCH) {
-        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, thread.IsWaitingAll()));
+        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, thread.IsSleepingOnWaitAll()));
     }
 
     return list;
--- a/src/citra_qt/debugger/wait_tree.cpp
+++ b/src/citra_qt/debugger/wait_tree.cpp
@@ -230,7 +230,7 @@ std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThread::GetChildren() const {
         list.push_back(std::make_unique<WaitTreeMutexList>(thread.held_mutexes));
     }
     if (thread.status == THREADSTATUS_WAIT_SYNCH) {
-        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, thread.wait_all));
+        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, !thread.wait_objects.empty()));
     }
 
     return list;
--- a/src/citra_qt/debugger/wait_tree.cpp
+++ b/src/citra_qt/debugger/wait_tree.cpp
@@ -230,7 +230,8 @@ std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThread::GetChildren() const {
         list.push_back(std::make_unique<WaitTreeMutexList>(thread.held_mutexes));
     }
     if (thread.status == THREADSTATUS_WAIT_SYNCH) {
-        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, thread.IsSleepingOnWaitAll()));
+        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects,
+                                                            thread.IsSleepingOnWaitAll()));
     }
 
     return list;
--- a/src/citra_qt/debugger/wait_tree.cpp
+++ b/src/citra_qt/debugger/wait_tree.cpp
@@ -230,7 +230,7 @@ std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThread::GetChildren() const {
         list.push_back(std::make_unique<WaitTreeMutexList>(thread.held_mutexes));
     }
     if (thread.status == THREADSTATUS_WAIT_SYNCH) {
-        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, !thread.wait_objects.empty()));
+        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, thread.IsWaitingAll()));
     }
 
     return list;
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -50,9 +50,9 @@ SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
         if (thread->current_priority >= candidate_priority)
             continue;
 
-        bool ready_to_run = std::none_of(thread->wait_objects.begin(), thread->wait_objects.end(), [](const SharedPtr<WaitObject>& object) {
-            return object->ShouldWait();
-        });
+        bool ready_to_run =
+            std::none_of(thread->wait_objects.begin(), thread->wait_objects.end(),
+                         [](const SharedPtr<WaitObject>& object) { return object->ShouldWait(); });
         if (ready_to_run) {
             candidate = thread.get();
             candidate_priority = thread->current_priority;
@@ -83,7 +83,8 @@ void WaitObject::WakeupAllWaitingThreads() {
 
         thread->SetWaitSynchronizationResult(RESULT_SUCCESS);
         thread->ResumeFromWait();
-        // Note: Removing the thread from the object's waitlist will be done by GetHighestPriorityReadyThread
+        // Note: Removing the thread from the object's waitlist will be
+        // done by GetHighestPriorityReadyThread.
     }
 }
 
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -40,24 +40,23 @@ SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
     if (waiting_threads.empty())
         return nullptr;
 
-    auto candidate_threads = waiting_threads;
+    SharedPtr<Thread> candidate = nullptr;
+    s32 candidate_priority = THREADPRIO_LOWEST + 1;
 
-    // Eliminate all threads that are waiting on more than one object, and not all of said objects are ready
-    candidate_threads.erase(std::remove_if(candidate_threads.begin(), candidate_threads.end(), [](const SharedPtr<Thread>& thread) -> bool {
-        return std::any_of(thread->wait_objects.begin(), thread->wait_objects.end(), [](const SharedPtr<WaitObject>& object) -> bool {
+    for (const auto& thread : waiting_threads) {
+        if (thread->current_priority >= candidate_priority)
+            continue;
+
+        bool ready_to_run = std::none_of(thread->wait_objects.begin(), thread->wait_objects.end(), [](const SharedPtr<WaitObject>& object) {
             return object->ShouldWait();
         });
-    }), candidate_threads.end());
+        if (ready_to_run) {
+            candidate = thread;
+            candidate_priority = thread->current_priority;
+        }
+    }
 
-    // Return the thread with the lowest priority value (The one with the highest priority)
-    auto thread_itr = std::min_element(candidate_threads.begin(), candidate_threads.end(), [](const SharedPtr<Thread>& lhs, const SharedPtr<Thread>& rhs) {
-        return lhs->current_priority < rhs->current_priority;
-    });
-
-    if (thread_itr == candidate_threads.end())
-        return nullptr;
-
-    return *thread_itr;
+    return candidate;
 }
 
 void WaitObject::WakeupAllWaitingThreads() {
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -31,13 +31,62 @@ void WaitObject::RemoveWaitingThread(Thread* thread) {
         waiting_threads.erase(itr);
 }
 
+SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
+    // Remove the threads that are ready or already running from our waitlist
+    waiting_threads.erase(std::remove_if(waiting_threads.begin(), waiting_threads.end(), [](SharedPtr<Thread> thread) -> bool {
+        return thread->status == THREADSTATUS_RUNNING || thread->status == THREADSTATUS_READY;
+    }), waiting_threads.end());
+
+    if (waiting_threads.empty())
+        return nullptr;
+
+    auto candidate_threads = waiting_threads;
+
+    // Eliminate all threads that are waiting on more than one object, and not all of them are ready
+    candidate_threads.erase(std::remove_if(candidate_threads.begin(), candidate_threads.end(), [](SharedPtr<Thread> thread) -> bool {
+        for (auto object : thread->wait_objects)
+            if (object->ShouldWait())
+                return true;
+        return false;
+    }), candidate_threads.end());
+
+    // Return the thread with the lowest priority value (The one with the highest priority)
+    auto thread_itr = std::min_element(candidate_threads.begin(), candidate_threads.end(), [](const SharedPtr<Thread>& lhs, const SharedPtr<Thread>& rhs) {
+        return lhs->current_priority < rhs->current_priority;
+    });
+
+    if (thread_itr == candidate_threads.end())
+        return nullptr;
+
+    return *thread_itr;
+}
+
 void WaitObject::WakeupAllWaitingThreads() {
-    for (auto thread : waiting_threads)
+    // Wake up all threads that can be awoken, in priority order
+    while (auto thread = GetHighestPriorityReadyThread()) {
+        if (thread->wait_objects.empty()) {
+            Acquire();
+            // Set the output index of the WaitSynchronizationN call to the index of this object.
+            if (thread->wait_set_output) {
+                thread->SetWaitSynchronizationOutput(thread->GetWaitObjectIndex(this));
+                thread->wait_set_output = false;
+            }
+        } else {
+            for (auto object : thread->wait_objects) {
+                object->Acquire();
+                // Remove the thread from the object's waitlist
+                object->RemoveWaitingThread(thread.get());
+            }
+            // Note: This case doesn't update the output index of WaitSynchronizationN.
+            // Clear the thread's waitlist
+            thread->wait_objects.clear();
+        }
+
+        // Set the result of the call to WaitSynchronization to RESULT_SUCCESS
+        thread->SetWaitSynchronizationResult(RESULT_SUCCESS);
         thread->ResumeFromWait();
-
-    waiting_threads.clear();
-
-    HLE::Reschedule(__func__);
+        // Note: Removing the thread from the object's waitlist will be done by GetHighestPriorityReadyThread
+    }
 }
 
 const std::vector<SharedPtr<Thread>>& WaitObject::GetWaitingThreads() const {
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
+#include <boost/range/algorithm_ext/erase.hpp>
 #include "common/assert.h"
 #include "common/logging/log.h"
 #include "core/hle/config_mem.h"
@@ -33,9 +34,9 @@ void WaitObject::RemoveWaitingThread(Thread* thread) {
 
 SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
     // Remove the threads that are ready or already running from our waitlist
-    waiting_threads.erase(std::remove_if(waiting_threads.begin(), waiting_threads.end(), [](const SharedPtr<Thread>& thread) -> bool {
+    boost::range::remove_erase_if(waiting_threads, [](const SharedPtr<Thread>& thread) -> bool {
         return thread->status == THREADSTATUS_RUNNING || thread->status == THREADSTATUS_READY;
-    }), waiting_threads.end());
+    });
 
     if (waiting_threads.empty())
         return nullptr;
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -38,6 +38,11 @@ SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
         return thread->status == THREADSTATUS_RUNNING || thread->status == THREADSTATUS_READY;
     });
 
+    // TODO(Subv): This call should be performed inside the loop below to check if an object can be
+    // acquired by a particular thread. This is useful for things like recursive locking of Mutexes.
+    if (ShouldWait())
+        return nullptr;
+
     Thread* candidate = nullptr;
     s32 candidate_priority = THREADPRIO_LOWEST + 1;
 
@@ -67,7 +72,7 @@ void WaitObject::WakeupAllWaitingThreads() {
                 thread->wait_set_output = false;
             }
         } else {
-            for (auto object : thread->wait_objects) {
+            for (auto& object : thread->wait_objects) {
                 object->Acquire();
                 object->RemoveWaitingThread(thread.get());
             }
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -34,14 +34,11 @@ void WaitObject::RemoveWaitingThread(Thread* thread) {
 
 SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
     // Remove the threads that are ready or already running from our waitlist
-    boost::range::remove_erase_if(waiting_threads, [](const SharedPtr<Thread>& thread) -> bool {
+    boost::range::remove_erase_if(waiting_threads, [](const SharedPtr<Thread>& thread) {
         return thread->status == THREADSTATUS_RUNNING || thread->status == THREADSTATUS_READY;
     });
 
-    if (waiting_threads.empty())
-        return nullptr;
-
-    SharedPtr<Thread> candidate = nullptr;
+    Thread* candidate = nullptr;
     s32 candidate_priority = THREADPRIO_LOWEST + 1;
 
     for (const auto& thread : waiting_threads) {
@@ -52,7 +49,7 @@ SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
             return object->ShouldWait();
         });
         if (ready_to_run) {
-            candidate = thread;
+            candidate = thread.get();
             candidate_priority = thread->current_priority;
         }
     }
@@ -61,9 +58,8 @@ SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
 }
 
 void WaitObject::WakeupAllWaitingThreads() {
-    // Wake up all threads that can be awoken, in priority order
     while (auto thread = GetHighestPriorityReadyThread()) {
-        if (thread->wait_objects.empty()) {
+        if (!thread->IsSleepingOnWaitAll()) {
             Acquire();
             // Set the output index of the WaitSynchronizationN call to the index of this object.
             if (thread->wait_set_output) {
@@ -73,7 +69,6 @@ void WaitObject::WakeupAllWaitingThreads() {
         } else {
             for (auto object : thread->wait_objects) {
                 object->Acquire();
-                // Remove the thread from the object's waitlist
                 object->RemoveWaitingThread(thread.get());
             }
             // Note: This case doesn't update the output index of WaitSynchronizationN.
@@ -81,7 +76,6 @@ void WaitObject::WakeupAllWaitingThreads() {
             thread->wait_objects.clear();
         }
 
-        // Set the result of the call to WaitSynchronization to RESULT_SUCCESS
         thread->SetWaitSynchronizationResult(RESULT_SUCCESS);
         thread->ResumeFromWait();
         // Note: Removing the thread from the object's waitlist will be done by GetHighestPriorityReadyThread
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -33,7 +33,7 @@ void WaitObject::RemoveWaitingThread(Thread* thread) {
 
 SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
     // Remove the threads that are ready or already running from our waitlist
-    waiting_threads.erase(std::remove_if(waiting_threads.begin(), waiting_threads.end(), [](SharedPtr<Thread> thread) -> bool {
+    waiting_threads.erase(std::remove_if(waiting_threads.begin(), waiting_threads.end(), [](const SharedPtr<Thread>& thread) -> bool {
         return thread->status == THREADSTATUS_RUNNING || thread->status == THREADSTATUS_READY;
     }), waiting_threads.end());
 
@@ -42,12 +42,11 @@ SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
 
     auto candidate_threads = waiting_threads;
 
-    // Eliminate all threads that are waiting on more than one object, and not all of them are ready
-    candidate_threads.erase(std::remove_if(candidate_threads.begin(), candidate_threads.end(), [](SharedPtr<Thread> thread) -> bool {
-        for (auto object : thread->wait_objects)
-            if (object->ShouldWait())
-                return true;
-        return false;
+    // Eliminate all threads that are waiting on more than one object, and not all of said objects are ready
+    candidate_threads.erase(std::remove_if(candidate_threads.begin(), candidate_threads.end(), [](const SharedPtr<Thread>& thread) -> bool {
+        return std::any_of(thread->wait_objects.begin(), thread->wait_objects.end(), [](const SharedPtr<WaitObject>& object) -> bool {
+            return object->ShouldWait();
+        });
     }), candidate_threads.end());
 
     // Return the thread with the lowest priority value (The one with the highest priority)
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -120,8 +120,6 @@ void Thread::Stop() {
     u32 tls_slot =
         ((tls_address - Memory::TLS_AREA_VADDR) % Memory::PAGE_SIZE) / Memory::TLS_ENTRY_SIZE;
     Kernel::g_current_process->tls_slots[tls_page].reset(tls_slot);
-
-    HLE::Reschedule(__func__);
 }
 
 Thread* ArbitrateHighestPriorityThread(u32 address) {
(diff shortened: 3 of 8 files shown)


src/core/hle/svc.cpp
```diff
@@ -300,90 +301,113 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
-            auto object = Kernel::g_handle_table.GetWaitObject(handles[i]);
-            object->AddWaitingThread(Kernel::GetCurrentThread());
-            wait_objects.push_back(object);
+        // Add the thread to each of the objects' waiting threads.
+        for (int i = 0; i < objects.size(); ++i) {
+            ObjectPtr object = objects[i];
+            // Set the index of this object in the mapping of Objects -> index for this thread.
+            thread->wait_objects_index[object->GetObjectId()] = i;
+            object->AddWaitingThread(thread);
+            // TODO(Subv): Perform things like update the mutex lock owner's priority to prevent priority inversion.
```
Subv: Currently this is done in Kernel::Mutex::ShouldWait, but we might want to move it to a dedicated function that is called from here.

Subv:
I've tested this on a few games (Quell Reflect, Pokemon X, Brunswick Pro Bowling, Xevious, Zelda: A Link Between Worlds) and haven't found any regressions or freezes.

Zelda: Majora's Mask still doesn't let you go past the CirclePadPro screen, but I guess that's related to the IR:USER service instead of a thread scheduling error.

src/core/hle/kernel/kernel.cpp
```diff
@@ -31,13 +31,62 @@ void WaitObject::RemoveWaitingThread(Thread* thread) {
...
+    if (waiting_threads.empty())
+        return nullptr;
 
-    waiting_threads.clear();
+    auto candidate_threads = waiting_threads;
 
-    HLE::Reschedule(__func__);
+    // Eliminate all threads that are waiting on more than one object, and not all of them are ready
+    candidate_threads.erase(std::remove_if(candidate_threads.begin(), candidate_threads.end(), [](SharedPtr<Thread> thread) -> bool {
+        for (auto object : thread->wait_objects)
```
lioncash: You can just `return std::any_of(...` here

src/core/hle/kernel/thread.h
```diff
@@ -125,6 +126,16 @@ class Thread final : public WaitObject {
     void SetWaitSynchronizationOutput(s32 output);
 
     /**
+     * Retrieves the index that this particular object occupies in the list of objects
+     * that the thread passed to WaitSynchronizationN.
+     * It is used to set the output value of WaitSynchronizationN when the thread is awakened.
+     * @param object Object to query the index of.
+     */
+    s32 GetWaitObjectIndex(WaitObject* object) {
```
lioncash: `object` and the function itself can be const-qualified.

src/core/hle/svc.cpp
```diff
@@ -300,90 +301,113 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
+        auto object = Kernel::g_handle_table.GetWaitObject(handles[i]);
+        if (object == nullptr)
+            return ERR_INVALID_HANDLE;
+        objects[i] = object;
     }
 
-    SCOPE_EXIT({
-        HLE::Reschedule("WaitSynchronizationN");
-    }); // Reschedule after putting the threads to sleep.
+    // Clear the mapping of wait object indices
```
lioncash: This doesn't really elaborate on anything that wouldn't be obvious with the line below this.

src/core/hle/svc.cpp
```diff
@@ -300,90 +301,113 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
+    thread->wait_objects_index.clear();
+
+    if (!wait_all) {
+        // Find the first object that is acquireable in the provided list of objects
+        auto itr = std::find_if(objects.begin(), objects.end(), [](const ObjectPtr& object) {
```
lioncash: This and the below if statement can probably be its own function.

src/core/hle/svc.cpp
```diff
@@ -300,90 +301,113 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
-        wait_objects.reserve(handle_count);
+        // Clear the thread's waitlist, we won't use it for wait_all = false
+        thread->wait_objects.clear();
 
-        for (int i = 0; i < handle_count; ++i) {
-            auto object = Kernel::g_handle_table.GetWaitObject(handles[i]);
-            object->AddWaitingThread(Kernel::GetCurrentThread());
-            wait_objects.push_back(object);
+        // Add the thread to each of the objects' waiting threads.
+        for (int i = 0; i < objects.size(); ++i) {
```
lioncash: `size_t` (the usage can be casted).

src/core/hle/svc.cpp
```diff
@@ -300,90 +301,113 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
-        // If no handles were passed in, put the thread to sleep only when 'wait_all' is false
-        // NOTE: This should deadlock the current thread if no timeout was specified
-        if (!wait_all) {
-            wait_thread = true;
-        }
+    using ObjectPtr = Kernel::SharedPtr<Kernel::WaitObject>;
+
+    std::vector<ObjectPtr> objects(handle_count);
+
+    for (int i = 0; i < handle_count; ++i) {
```
lioncash: This can probably just be a function that returns a vector.

src/core/hle/svc.cpp
```diff
@@ -249,27 +249,30 @@ static ResultCode WaitSynchronization1(Handle handle, s64 nano_seconds) {
...
     LOG_TRACE(Kernel_SVC, "called handle=0x%08X(%s:%s), nanoseconds=%lld", handle,
               object->GetTypeName().c_str(), object->GetName().c_str(), nano_seconds);
 
-    HLE::Reschedule(__func__);
-
-    // Check for next thread to schedule
     if (object->ShouldWait()) {
 
+        if (nano_seconds == 0)
+            return ResultCode(ErrorDescription::Timeout, ErrorModule::OS,
```
lioncash: Should probably brace this, since the body traverses multiple lines to prevent the potential for a dead if-statement if refactoring happens in the future.

yuriks:
This does not fix #2091. That issue is not about the deadlock when creating a new game, it's about the savegame loading with the wrong player data.

yuriks:
We'll, I'll be fucked. This actually *does* fix the savegame loading incorrectly in #2091.

src/core/hle/svc.cpp
```diff
@@ -249,27 +249,30 @@ static ResultCode WaitSynchronization1(Handle handle, s64 nano_seconds) {
...
         object->AddWaitingThread(thread);
-        Kernel::WaitCurrentThread_WaitSynchronization({object}, false, false);
+        thread->status = THREADSTATUS_WAIT_SYNCH;
 
         // Create an event to wake the thread up after the specified nanosecond delay has passed
         thread->WakeAfterDelay(nano_seconds);
 
-        // NOTE: output of this SVC will be set later depending on how the thread resumes
-        return HLE::RESULT_INVALID;
+        // Note: The output of this SVC will be set to RESULT_SUCCESS if the thread resumes due to a signal in one of its wait objects.
```
wwylele: "one of its wait objects." should be just "its wait object"

yuriks:
Tested on all my other games, no regressions. Apart from the ALBW fix, Pokemon OR does not lock up after exiting the truck after the intro anymore. I can play the game normally. Pokemon Y still freezes after the intro and Bravely Default still freezes if you change the audio language.

src/core/hle/svc.cpp
wwylele: ^ This comment should go into the TODO, otherwise we will forget it is in `Kernel::Mutex::ShouldWait` later

src/core/hle/svc.cpp
```diff
@@ -300,90 +301,113 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
+                          ErrorSummary::StatusChanged,
+                          ErrorLevel::Info);
+    } else {
+        bool all_available = std::all_of(objects.begin(), objects.end(), [](const ObjectPtr& object) {
+            return !object->ShouldWait();
+        });
+        if (all_available) {
+            // We can acquire all objects right now, do so.
+            for (auto object : objects)
+                object->Acquire();
```
wwylele: Probably a regression on https://github.com/citra-emu/citra/issues/1308 ?

src/core/hle/svc.cpp
```diff
@@ -1148,6 +1172,7 @@ void CallSVC(u32 immediate) {
     if (info) {
         if (info->func) {
             info->func();
+            HLE::Reschedule(__func__);
```
wwylele: According to https://gist.github.com/Subv/02f29bd9f1e5deb7aceea1e8f019c8f4, not all `WaitSynchronization1` will cause Reschedule, no?

src/core/hle/kernel/thread.h
```diff
@@ -154,16 +165,16 @@ class Thread final : public WaitObject {
 
     VAddr tls_address; ///< Virtual address of the Thread Local Storage of the thread
 
-    bool waitsynch_waited; ///< Set to true if the last svcWaitSynch call caused the thread to wait
-
     /// Mutexes currently held by this thread, which will be released when it exits.
     boost::container::flat_set<SharedPtr<Mutex>> held_mutexes;
 
     SharedPtr<Process> owner_process;                ///< Process that owns this thread
     std::vector<SharedPtr<WaitObject>> wait_objects; ///< Objects that the thread is waiting on
```
wwylele: Needs to document about the usage when `wait_all = false`. Also Needs a small function like `bool IsWaitingAll() { return !wait_objects.empty();}`. A single `wait_objects.empty()` floating around is really hard to read.

src/core/hle/svc.cpp
Subv: No, the behavior here didnt Change much.

It will check ShouldWait for all objects first (And it will be true for all of the semaphores in that test during the last WaitSynchN call), then acquire them all in sequence

src/core/hle/svc.cpp
Subv: Indeed, but that is a big change in itself, I'll add a TODO and fix the reschedule conditions in my next PR.

Romsstar:
This fixed #2119 ,
thanks, much appreciated :)

Subv:
There appears to be a bug in our thread priority boosting code, the priority of a thread will be reset to its nominal priority when it is scheduled back, however, this conflicts with manual calls to svcSetThreadPriority.

We should make a test to verify whether the thread priority is indeed reset after each schedule.
Are we sure we should be doing priority boosting?
Pinging @bunnei

src/core/hle/kernel/kernel.cpp
```diff
@@ -31,13 +31,61 @@ void WaitObject::RemoveWaitingThread(Thread* thread) {
...
+    // Remove the threads that are ready or already running from our waitlist
+    waiting_threads.erase(std::remove_if(waiting_threads.begin(), waiting_threads.end(), [](const SharedPtr<Thread>& thread) -> bool {
+        return thread->status == THREADSTATUS_RUNNING || thread->status == THREADSTATUS_READY;
+    }), waiting_threads.end());
+
+    if (waiting_threads.empty())
+        return nullptr;
 
-    waiting_threads.clear();
+    auto candidate_threads = waiting_threads;
```
yuriks: This algorithm will cause frequent allocations, does a lot of unnecessary checks. The "manual" version is shorter, faster and imo easier to understand:

```
Thread* candidate = nullptr;
int candidate_priority = 0x40; // Do we have a MAX_PRIORITY constant?

for (Thread* t : waiting_threads) {
    if (t->current_priority > candidate_priority) {
        continue;
    }
    bool ready_to_run = std::none_of(t->wait_objects.begin(), t->wait_object.end(), [](const WaitObject* object) {
        return object->ShouldWait();
    };
    if (ready_to_run) {
        candidate = t;
        candidate_priority = t->current_priority;
    }
}

return candidate;
```

src/core/hle/kernel/kernel.cpp
```diff
@@ -40,24 +40,23 @@ SharedPtr<Thread> WaitObject::GetHighestPriorityReadyThread() {
     if (waiting_threads.empty())
         return nullptr;
 
-    auto candidate_threads = waiting_threads;
+    SharedPtr<Thread> candidate = nullptr;
```
yuriks: It's fine to use a plain pointer here since I think you'd be holding the scheduler lock while in here, right? Saves on a bunch of refcount increment/decrements.

src/core/hle/kernel/thread.cpp
```diff
@@ -619,14 +536,6 @@ void Reschedule() {
 
     HLE::DoneRescheduling();
 
-    // Don't bother switching to the same thread.
-    // But if the thread was waiting on objects, we still need to switch it
-    // to perform PC modification, change state to RUNNING, etc.
-    // This occurs in the case when an object the thread is waiting on immediately wakes up
-    // the current thread before Reschedule() is called.
-    if (next == cur && (next == nullptr || next->waitsynch_waited == false))
```
yuriks: Can `next == cur` never happen anymore, or why did you remove it? If it's impossible then add an assert.

src/core/hle/kernel/thread.h
```diff
@@ -154,16 +174,20 @@ class Thread final : public WaitObject {
...
     boost::container::flat_set<SharedPtr<Mutex>> held_mutexes;
 
     SharedPtr<Process> owner_process;                ///< Process that owns this thread
-    std::vector<SharedPtr<WaitObject>> wait_objects; ///< Objects that the thread is waiting on
+
+    /// Objects that the thread is waiting on.
+    /// This is only populated when the thread should wait for all the objects to become ready.
+    std::vector<SharedPtr<WaitObject>> wait_objects;
+
+    std::unordered_map<int, s32> wait_objects_index; ///< Mapping of Object ids to their position in the last waitlist that this object waited on.
```
yuriks: flat_map probably a better choice

src/core/hle/svc.cpp
```diff
@@ -249,27 +249,33 @@ static ResultCode WaitSynchronization1(Handle handle, s64 nano_seconds) {
...
 
     LOG_TRACE(Kernel_SVC, "called handle=0x%08X(%s:%s), nanoseconds=%lld", handle,
               object->GetTypeName().c_str(), object->GetName().c_str(), nano_seconds);
 
-    HLE::Reschedule(__func__);
-
-    // Check for next thread to schedule
     if (object->ShouldWait()) {
 
+        if (nano_seconds == 0) {
```
yuriks: Flip this to `nano_seconds != 0` so you don't need to duplicate the return.

src/core/hle/svc.cpp
```diff
@@ -283,8 +289,6 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
     bool wait_thread = !wait_all;
```
yuriks: I think this is unused now

src/core/hle/svc.cpp
```diff
@@ -283,8 +289,6 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
     int handle_index = 0;
```
yuriks: Also unused

src/core/hle/svc.cpp
```diff
@@ -300,90 +304,117 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
-            }
-        }
-    } else {
-        // If no handles were passed in, put the thread to sleep only when 'wait_all' is false
-        // NOTE: This should deadlock the current thread if no timeout was specified
-        if (!wait_all) {
-            wait_thread = true;
-        }
+    using ObjectPtr = Kernel::SharedPtr<Kernel::WaitObject>;
+
```
yuriks: Unnecessary newline

src/core/hle/svc.cpp
```diff
@@ -300,90 +304,117 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
-        std::vector<SharedPtr<Kernel::WaitObject>> wait_objects;
-        wait_objects.reserve(handle_count);
+        // No objects were ready to be acquired, prepare to suspend the thread.
 
-        for (int i = 0; i < handle_count; ++i) {
-            auto object = Kernel::g_handle_table.GetWaitObject(handles[i]);
-            object->AddWaitingThread(Kernel::GetCurrentThread());
-            wait_objects.push_back(object);
+        // If a timeout value of 0 was provided, just return the Timeout error code instead of suspending the thread.
+        if (nano_seconds == 0) {
```
yuriks: Same deal, flip the condition to avoid duplicating the timeout.

src/core/hle/svc.cpp
```diff
@@ -300,90 +304,117 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
-        Kernel::WaitCurrentThread_WaitSynchronization(std::move(wait_objects), true, wait_all);
+        // Put the thread to sleep
+        thread->status = THREADSTATUS_WAIT_SYNCH;
+
+        // Clear the thread's waitlist, we won't use it for wait_all = false
+        thread->wait_objects.clear();
+
+        // Add the thread to each of the objects' waiting threads.
+        for (size_t i = 0; i < objects.size(); ++i) {
+            ObjectPtr object = objects[i];
```
yuriks: This can be just a raw pointer

src/core/hle/svc.cpp
```diff
@@ -300,90 +304,117 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
-        HLE::Reschedule("WaitSynchronizationN");
-    }); // Reschedule after putting the threads to sleep.
+    // Clear the mapping of wait object indices.
+    // We don't want any lingering state in this map.
+    // It will be repopulated later in the wait_all = false case.
+    thread->wait_objects_index.clear();
 
-    // If thread should wait, then set its state to waiting
-    if (wait_thread) {
+    if (!wait_all) {
```
yuriks: Can you switch the two sides of the branch and remove the negation? They're both about the same size so I don't think there's any reason to keep the not.

src/core/hle/svc.cpp
```diff
@@ -300,90 +304,117 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
-    // Acquire objects if we did not wait...
-    for (int i = 0; i < handle_count; ++i) {
-        auto object = Kernel::g_handle_table.GetWaitObject(handles[i]);
+        // Not all objects were available right now, prepare to suspend the thread.
 
-        // Acquire the object if it is not waiting...
-        if (!object->ShouldWait()) {
-            object->Acquire();
+        // If a timeout value of 0 was provided, just return the Timeout error code instead of suspending the thread.
+        if (nano_seconds == 0) {
```
yuriks: You know the drill.

src/core/hle/svc.cpp
```diff
@@ -300,90 +304,117 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
+
+        // Put the thread to sleep
+        thread->status = THREADSTATUS_WAIT_SYNCH;
 
-            // If this was the first non-waiting object and 'wait_all' is false, don't acquire
-            // any other objects
-            if (!wait_all)
-                break;
+        // Set the thread's waitlist to the list of objects passed to WaitSynchronizationN
+        thread->wait_objects = objects;
```
yuriks: Move this to after the for loop and you can then std::move the vector

emmauss:
This fixes #2028.

src/citra_qt/debugger/wait_tree.cpp
```diff
@@ -230,7 +230,7 @@ std::vector<std::unique_ptr<WaitTreeItem>> WaitTreeThread::GetChildren() const {
         list.push_back(std::make_unique<WaitTreeMutexList>(thread.held_mutexes));
     }
     if (thread.status == THREADSTATUS_WAIT_SYNCH) {
-        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, thread.wait_all));
+        list.push_back(std::make_unique<WaitTreeObjectList>(thread.wait_objects, thread.IsWaitingAll()));
```
Subv: Any suggestions?

src/core/hle/kernel/thread.cpp
Subv: It is not possible *right now*, but it will be when we make SendSyncRequest pause the running thread. we want to cause a context Switch in that case too.

src/core/hle/kernel/thread.h
Subv: why?

src/core/hle/svc.cpp
Subv: Done

Dragios:
Fixes #2026 as well

src/citra_qt/debugger/wait_tree.cpp
yuriks: "IsSleepingOnWaitAll"? I don't have any either. This is kind of a weird one because it doesn't even makes sense in all states either. For example, if the thread isn't sleeping on a WaitSynchN at all it mayb still return true if the list is non-empty.

src/core/hle/kernel/thread.cpp
yuriks: That still doesn't really answer why you removed it. If next == cur and you don't need to do the special wait synch setup then you can skip switching the context (this also avoids pointless null -> null reschedules.)

src/core/hle/kernel/thread.h
yuriks: The map holds only small number of elements and the keys and values are very small

src/core/hle/svc.cpp
yuriks: At least factor the error code out to a constant then, because it's repeated 4 times in this file.

src/core/hle/svc.cpp
yuriks: ...locally?

src/core/hle/svc.cpp
```diff
@@ -300,90 +302,116 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
 
-        for (int i = 0; i < handle_count; ++i) {
-            auto object = Kernel::g_handle_table.GetWaitObject(handles[i]);
-            object->AddWaitingThread(Kernel::GetCurrentThread());
-            wait_objects.push_back(object);
+        // Put the thread to sleep
+        thread->status = THREADSTATUS_WAIT_SYNCH;
+
+        // Add the thread to each of the objects' waiting threads.
+        for (auto object : objects) {
```
yuriks: auto& (otherwise you're making a copy of the SharedPtr on every iteration)

src/core/hle/kernel/thread.cpp
Subv: When applying the changes outlined in #2125 (Which are more or less correct, and will be part of my next PR), not performing a context switch in the ```cur == next``` case where ```cur != nullptr && cur->status == THREADSTATUS_READY``` causes a SIGSEGV in ```PriorityBoostStarvedThreads``` because we don't re-add the thread to the ```ready_queue``` and it ends up being empty.

src/citra_qt/debugger/wait_tree.cpp
Subv: I'll rename it to IsSleepingOnWaitAll for now, my next PR will add more thread statuses anyways, so we can check for  THREADSTATUS_WAITSYNCH && !wait.empty() inside this function.

src/core/hle/kernel/thread.cpp
Subv: The case above can happen if the thread was put to sleep for a very short period of time, and awakened before another thread was made to run.

src/core/hle/svc.cpp
```diff
@@ -320,100 +315,90 @@ static ResultCode WaitSynchronizationN(s32* out, Handle* handles, s32 handle_cou
...
         });
-
-        if (itr != objects.end()) {
-            // We found a ready object, acquire it and set the result value
-            ObjectPtr object = *itr;
-            object->Acquire();
-            *out = std::distance(objects.begin(), itr);
+        if (all_available) {
+            // We can acquire all objects right now, do so.
+            for (auto object : objects)
```
yuriks: oops, auto& here too

src/core/hle/svc.cpp
wwylele: (Sorry for a late test) The test in #1308 ([this](https://github.com/wwylele/ctrhwtest/tree/master/semaphore_acquire_n/source)) hits the ASSERT
```
[   0.345067] Debug <Critical> core/hle/kernel/semaphore.cpp:operator():38: Assertion Failed!
object unavailable!
```

wwylele:
I haven't got time to write and do tests, so I try to do some thought experiment with the code. However, the logic in some circumstances seems unclear (incorrect?) to me:

- say a thread calls `WaitSynchN` with `wait_all=true`, `thread->wait_objects` gets populate. Then it times out and the thread wakes up, with `wait_objects` not cleared (I can't find any code in this path clearing it?). After a while the thread calls `WaitSynch1` and goes into a waiting state again, with `wait_objects` still populated (`WaitSynch1` doesn't clear it either), now the thread is actually in a "waiting for all N objects" state

 - say a thread, A, is holding a mutex, and another two theads, B and C, call `WaitSynch1` on that mutex and go into waiting state. Now thread A releases the mutex, where the code path is: `ReleaseMutex` -> `Mutex::Release` -> `(mutex.cpp) ResumeWaitingThread` -> `WaitObject::WakeupAllWaitingThread` ->
{  `GetHighestPriorityReadyThread` -> returns thread B because of an empty `wait_objects`;  `Mutex::Acquire()`; B resumes; `GetHighestPriorityReadyThread` again -> returns thread C because of an empty `wait_objects`; `Mutex::Acquire()` again -> hits `ASSERT` because B is holding the mutex }

Subv:
@wwylele Both issues should be fixed now with the latest commit.

Subv:
A @yuriks LGTM means this thing is out of RFC state \o/

Now I'd love to get a @wwylele LGTM

src/core/hle/kernel/thread.h
```diff
@@ -215,10 +240,9 @@ void WaitCurrentThread_Sleep();
  * @param wait_objects Kernel objects that we are waiting on
  * @param wait_set_output If true, set the output parameter on thread wakeup (for
  * WaitSynchronizationN only)
- * @param wait_all If true, wait on all objects before resuming (for WaitSynchronizationN only)
  */
 void WaitCurrentThread_WaitSynchronization(std::vector<SharedPtr<WaitObject>> wait_objects,
-                                           bool wait_set_output, bool wait_all);
+                                           bool wait_set_output);
```
wwylele: This function is actually not used any more. We can remove it

yuriks:
You should fix the formatting errors in the build. (Are we testing this on BE for regressions for a while or just merging when it's ready?)

wwylele:
Please testing this on BE for regressions for some days. Actually I have already received potential regression report.

bunnei:
I don't think it hurts to let it marinate on BE for a few days...

PR #2264, opened 2016-12-04
Print shaders in case of error

JayFoxRox:
We recently had a lot of issues with people reporting "Uniform block size did not match!" which usually has one of 3 reasons:

- A broken shader (= compilation fails due to us not having implemented a certain codepath). The block size will be 0.
- Broken code / Alignment issues. This can happen if the last element is not 16 bytes in size (as the spec is unclear wether the block size will be aligned to the next 16 byte block or truncated). The block size will vary a little.
- A broken driver. The block size could vary greatly depending how it intepreted our shader. This should rarely happen.

To show these kind of errors better, the `ASSERT_MSG` was extended to show the expected and received block size.
Additionally the shader source code will be `LOG_ERROR`'d in case of program linking (both shaders will be printed) or shader compilation errors (that particular shader will be printed).

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_util.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_util.cpp
@@ -4,6 +4,7 @@
 
 #include <vector>
 #include <glad/glad.h>
+#include "common/assert.h"
 #include "common/logging/log.h"
 #include "video_core/renderer_opengl/gl_shader_util.h"
 
@@ -87,6 +88,7 @@ GLuint LoadProgram(const char* vertex_shader, const char* fragment_shader) {
         LOG_ERROR(Render_OpenGL, "Vertex shader:\n%s", vertex_shader);
         LOG_ERROR(Render_OpenGL, "Fragment shader:\n%s", fragment_shader);
     }
+    ASSERT_MSG(result == GL_TRUE, "Shader not linked");
 
     glDeleteShader(vertex_shader_id);
     glDeleteShader(fragment_shader_id);
--- a/src/video_core/renderer_opengl/gl_shader_util.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_util.cpp
@@ -31,7 +31,7 @@ GLuint LoadProgram(const char* vertex_shader, const char* fragment_shader) {
     if (info_log_length > 1) {
         std::vector<char> vertex_shader_error(info_log_length);
         glGetShaderInfoLog(vertex_shader_id, info_log_length, nullptr, &vertex_shader_error[0]);
-        if (result) {
+        if (result == GL_TRUE) {
             LOG_DEBUG(Render_OpenGL, "%s", &vertex_shader_error[0]);
         } else {
             LOG_ERROR(Render_OpenGL, "Error compiling vertex shader:\n%s", &vertex_shader_error[0]);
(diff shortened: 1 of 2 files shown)


JayFoxRox:
Interestingly this just gave me "Uniform block size did not match! Got -1406844112, expected 960" when I removed the `main` function in the shader. I'll check wether this is a GL error or a bug in my cast soon

JayFoxRox:
I've corrected the clang-format error (hopefully) and I've checked the weird block size using apitrace now:
My driver will trigger a handful of errors if the shader is not linked because we still try to access its uniforms. Once indidivual uniforms already gave errors, we're getting the uniform block index which is returning -1. Getting the size of that invalid block returns garbage (without any further GL error).

I think it's safest to just assert that the shader was linked properly.
In the future, as we progress towards a stable release, we should probably add more GL error checking and gracefully close if a shader did not compile / link.

src/video_core/renderer_opengl/gl_shader_util.cpp
```diff
@@ -69,19 +72,28 @@ GLuint LoadProgram(const char* vertex_shader, const char* fragment_shader) {
...
         glGetProgramInfoLog(program_id, info_log_length, nullptr, &program_error[0]);
-        if (result) {
+        if (program_result == GL_TRUE) {
             LOG_DEBUG(Render_OpenGL, "%s", &program_error[0]);
         } else {
             LOG_ERROR(Render_OpenGL, "Error linking shader:\n%s", &program_error[0]);
         }
     }
 
+    if (vertex_shader_result == GL_FALSE || program_result == GL_FALSE) {
```
yuriks: Isn't it sufficient to check just the linking result? If one shader fails to link that implies that the linking will also fail.

src/video_core/renderer_opengl/gl_shader_util.cpp
```diff
@@ -15,7 +16,9 @@ GLuint LoadProgram(const char* vertex_shader, const char* fragment_shader) {
     GLuint vertex_shader_id = glCreateShader(GL_VERTEX_SHADER);
     GLuint fragment_shader_id = glCreateShader(GL_FRAGMENT_SHADER);
 
-    GLint result = GL_FALSE;
+    GLint vertex_shader_result = GL_FALSE;
+    GLint fragment_shader_result = GL_FALSE;
+    GLint program_result = GL_FALSE;
```
yuriks: You can declare these just before they're used.

src/video_core/renderer_opengl/gl_shader_util.cpp
wwylele: I feel like the logic is wrong. Looks like you want to only print the corrupted shader, but now it will print both shaders because program_result is false (?) even if only one shader fails.

JayFoxRox: @wwylele yes, that was intended.
As shader varyings are send from the VS to the FS and a mismatch would cause linking issues, I decided to print both shaders. Similarily things like a missing `main()` function would also only occur at the linking stage without giving any (non-string) hints what actually went wrong.
But yuriks is also right: Both shaders are necessary for a working program. So if either of them fails compilation, the linking will also likely fail. So my logic is overcomplicated.

Issue #2261, opened 2016-12-04
Pokémon X/Y ‒ Both screen garbling bug


Dragios:
This happens when you open the menu on clothing customization session and it shows this transition effect for a second before entering the menu. Occurs only on *hardware renderer*
Related to #2246
![Pixelated screens](https://cloud.githubusercontent.com/assets/18496868/20865328/aa1ea91a-ba49-11e6-97f0-a4a5e47b0d59.png)

Nothing useful in log:
`HW.GPU <Critical> core\hw\gpu.cpp:TextureCopy:332: zero input width`
`HW.GPU <Critical> core\hw\gpu.cpp:TextureCopy:332: zero input width`

Switching to *software renderer* will playback the animation exactly like the actual hardware behaviour.

Tested this on commit 4f0f88b

**OS:** _Windows 10 build 14393.447 (codename Redstone 1)_
**GPU:** _nVidia GT 610 2GB DDR3 (Asus)_
**Driver:** _Version 375.95_
**OpenGL:** _Version 4.5_

yuriks:
Can you upload a save file and instructions to reproduce this?

Dragios:
[Pokémon X save file](https://github.com/Dragios/Cloud-Drive/blob/master/Pok%C3%A9mon%20X%20save%20file.zip?raw=true)

Once you loaded the game you will land in the Pokémon Center.

**Instruction to reproduce the issue:**
1. Move to the left and head straight up until you reach the location as shown in the screenshot.
![pokemon center](https://cloud.githubusercontent.com/assets/18496868/20865418/13894282-ba4c-11e6-9135-e3cedf7fb9c8.png)
2. *Press A* to enter to the clothing customization section.
3. *Press X* to open the menu

JayFoxRox:
I'm assuming this is the same as #2246 ?
Eitherway, we should label this [gl-renderer] unless confirmed to be a [tex-fwd] issue.
As the other issue was confirmed we should probably re-label it (or add both labels so we can find it more easily)

yuriks:
Please verify it's fixed with #2266

Dragios:
Confirm fixed.

Dragios:
Confirm fixed but it blinked once before enter the menu, weird.

yuriks:
What do you mean with "it blinked once before the menu"? What blinked?

Dragios:
[...]

PR #2266, opened 2016-12-04
OpenGL: Fix DisplayTransfer accel when input width != output width

### LINKED ISSUE #2246: Pokémon Omega Ruby/Alpha Saphire ‒ Top screen garbling bug
This happens when you found a wild pokemon and it shows the transition effect before starting the battle
Using *Hardware Renderer* displays this during transition:
Pixelated graphic

Nothing useful in log:
```
Service.APT <Warning> core\hle\service\apt\apt.cpp:SetScreenCapPostPermission:416: (STUBBED) screen_capture_post_permission=3
Service.CFG <Warning> core\hle\service\cfg\cfg.cpp:GenHashConsoleUnique:177: (STUBBED) called app_id_salt=0x55D
Service.FRD <Warning> core\hle\service\frd\frd.cpp:GetMyFriendKey:91: (STUBBED) called
```
Switching to *Sofware Renderer* will playback the animation exactly like the actual hardware behaviour.

Tested this on commit 3174bfd

**OS:** _Windows 10 build 14393.447 (codename Redstone 1)_
**GPU:** _nVidia GT 610 2GB DDR3 (Asus)_
**Driver:** _Version 375.95_
**OpenGL:** _Version 4.5_

yuriks:
First commit is the actual fix. Others are small cleanups/ticking-time-bomb bugfixes.
Fixes #2246, #2261.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -158,24 +158,21 @@ bool RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
         buffers = GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT;
     }
 
-    if (OpenGLState::CheckFBStatus(GL_READ_FRAMEBUFFER) != GL_FRAMEBUFFER_COMPLETE) {
-        return false;
-    }
+    bool can_blit = OpenGLState::CheckFBStatus(GL_READ_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE &&
+                    OpenGLState::CheckFBStatus(GL_DRAW_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE;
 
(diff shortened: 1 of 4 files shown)


yuriks:
cc @tfarley in case he wants to give some input on how I missed a crucial details in the implementation here and this is actually the wrong way to fix it. :)

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -291,6 +289,10 @@ CachedSurface* RasterizerCacheOpenGL::GetSurface(const CachedSurface& params, bo
 
     MICROPROFILE_SCOPE(OpenGL_SurfaceUpload);
 
+    if (params.is_tiled) {
+        ASSERT(params.stride == 0);
+    }
```
Subv: A comment about this would be nice.
Why?

yuriks: This *is* the comment. ;) (I spent some time chasing this lead before realizing that stride only makes sense for linear images.)

Subv:
LGTM

PR #2267, opened 2016-12-04
Support mingw cross-compilation

JayFoxRox:
This involves 3 kind of changes:

- Lowercase spelling for includes (`Windows.h` and `WinSock2.h`)
- Use of gcc intrinsics (Old check was for `_WIN32`, not the compiler)
- Use of forward slashes as path seperator (present in *.rc files, used for exe icon path)

Additionally I decided to update the icon name in the rc files.
One was kind of arbritrary, the other one referenced GLFW which we aren't even using anymore.

---

This currently does not pack SDL2 or Qt5 dependencies properly (which is done conditionally for MSVC builds). I intend to do this in a follow-up PR as we still have various Qt version checks which I'd like to remove anyway (as chances are very slim we'll ever support Qt4 or below again).
It's also not a cross-compile issue, but rather a lack in the original mingw support.

### DIFF
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -18,7 +18,7 @@
 #endif
 
 #ifdef _WIN32
-#include <Windows.h>
+#include <windows.h>
 #endif
 
 #include "citra/config.h"
--- a/src/citra/citra.rc
+++ b/src/citra/citra.rc
@@ -5,5 +5,5 @@
 
 // Icon with lowest ID value placed first to ensure application icon
 // remains consistent on all systems.
-GLFW_ICON               ICON                    "..\\..\\dist\\citra.ico"
+GLFW_ICON               ICON                    "../../dist/citra.ico"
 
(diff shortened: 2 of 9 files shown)


MerryMage:
LGTM

MerryMage:
Due to f120e78, clang-format complains about header ordering.

Suggestion is to do something akin to this: https://github.com/citra-emu/citra/blob/f120e78b5692e1c710f2335d36fd0593c447a613/src/common/file_util.cpp#L13

lioncash:
You can just do

```cpp
// clang-format off
#include <windows.h>
// clang-format on
```

yuriks:
Adding a comment between blocks of includes already instructs clang-format that it shouldn't reorder includes across that comment, so I think it's not necessary to fully disable it.

JayFoxRox:
I've done what merry suggested and based my reasoning on http://www.cplusplus.com/forum/windows/112014/ and http://stackoverflow.com/questions/5971332/redefinition-errors-in-winsock2-h .

I've also noticed that there was a `common/abi/` include specifically for windows and decided to remove it. It still compiles fine for me in mingw. I believe this was a copy/paste error or leftover?

linkmauve:
LGTM.

PR #2274, opened 2016-12-06
Drop framebuffer completeness check.

degasus:
This OpenGL call synchronize the worker thread of the nvidia blob.
It can be verified on linux with the __GL_THREADED_OPTIMIZATIONS=1 environment variable.
Those errors should not happen on tested drivers.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -103,7 +103,7 @@ static void MortonCopyPixels(CachedSurface::PixelFormat pixel_format, u32 width,
     }
 }
 
-bool RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
+void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
                                          CachedSurface::SurfaceType type,
                                          const MathUtil::Rectangle<int>& src_rect,
                                          const MathUtil::Rectangle<int>& dst_rect) {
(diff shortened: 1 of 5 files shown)


linkmauve:
The original reason for these checks was that a Mesa bug was preventing complete framebuffers from being used if there wasn’t a `glCheckFramebufferStatus()` call before the draw call.  It would manifest with a black screen on e.g. the Stock Pot Inn scene in MM3D’s intro.

I can’t find either the Mesa bug number or our tracking issue anymore though, it’s probably from long ago, and we don’t have any driver bug workaround system like in Dolphin, so LGTM (or even better, totally removing the checks since `GL_KHR_debug` should already give us that information).

JayFoxRox:
The original patch for the [issue mentioned / reported](https://bugs.freedesktop.org/show_bug.cgi?id=94148) by linkmauve ended up in [Mesa 12.0.2](http://www.mesa3d.org/relnotes/12.0.2.html) .

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -158,21 +158,14 @@ bool RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
     }
 
-    bool can_blit = OpenGLState::CheckFBStatus(GL_READ_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE &&
-                    OpenGLState::CheckFBStatus(GL_DRAW_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE;
-
-    if (can_blit) {
-        glBlitFramebuffer(src_rect.left, src_rect.top, src_rect.right, src_rect.bottom,
-                          dst_rect.left, dst_rect.top, dst_rect.right, dst_rect.bottom, buffers,
-                          buffers == GL_COLOR_BUFFER_BIT ? GL_LINEAR : GL_NEAREST);
-    }
```
yuriks: This test might be necessary, because we don't have a whitelist of valid blit sources/destinations. We use RGBA for all textures though (which is something we may or may not want to change later) so you'd only have to check against color<->depth/stencil surfaces I think.

degasus: Are you talking about TryBlitSurfaces? This function here will always return true on conform drivers.

yuriks: Ah, yes I was, sorry.

PR #2280, opened 2016-12-07
Fixed the gpu command list size when creating CiTraces.

Subv:
We were writing 4 times as much data, and thus reading (and writing) garbage to the linear heap when replaying the traces with the citrace player.


linkmauve:
LGTM.

JayFoxRox:
LGTM

PR #2282, opened 2016-12-08
Update AM service function tables

lioncash:
Updated based off information from 3dbrew.


Subv:
Do you plan on doing this for the other services? That would be nice

lioncash:
Yeah, I do, I just wasn't sure what would be preferable: a single PR with individual service updates per commit, or one service per PR (so each change is compact), so I just went with whatever would be smallest (I can do the all of the rest in one follow-up PR, if that's preferable).

PR #2290, opened 2016-12-09
Fixed Storing Input Configuration - QT Implementation

gyroninja:
I'm assuming this only affected a subset of all citra users but to users it did effect it rendered the qt implementation of Citra's UI useless since they could not keys. (Some keys could be bound to but they were limited and seemingly random)

After storing text in a push button QT will add an accelerator key to the text. For example if you stored "H" in the QPushButton after control got passed back to QT it would turn into "&H" when trying to find what key "&H" represented it would fail and just return Key_unknown. This & is known as an accelerator key.

My PR makes sure that after it grabs the text from the QPushButton it `.remove('&')` the result. I also added some code to handle when ampersands get thrown into it but like previous versions of Citra you can not bind to ampersand. Accelerators aren't always just a single ampersand but those cases won't show up automatically.

### DIFF
--- a/src/citra_qt/configure_input.cpp
+++ b/src/citra_qt/configure_input.cpp
@@ -57,7 +57,8 @@ ConfigureInput::ConfigureInput(QWidget* parent)
 
 void ConfigureInput::handleClick() {
     QPushButton* sender = qobject_cast<QPushButton*>(QObject::sender());
-    previous_mapping = sender->text();
+    QVariant key_variant = sender->property("key");
+    previous_mapping = key_variant.value<QString>();
     sender->setText(tr("[waiting]"));
     sender->setFocus();
     grabKeyboard();
@@ -68,7 +69,8 @@ void ConfigureInput::handleClick() {
 
 void ConfigureInput::applyConfiguration() {
     for (int i = 0; i < Settings::NativeInput::NUM_INPUTS; ++i) {
-        int value = getKeyValue(input_mapping[Settings::NativeInput::Values(i)]->text());
+        QVariant key_variant = input_mapping[Settings::NativeInput::Values(i)]->property("key");
+        int value = getKeyValue(key_variant.value<QString>());
         Settings::values.input_mappings[Settings::NativeInput::All[i]] = value;
     }
     Settings::Apply();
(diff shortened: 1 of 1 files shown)


src/citra_qt/configure_input.h
```diff
@@ -55,6 +55,9 @@ class ConfigureInput : public QWidget {
     /// Set button text to name of key pressed.
     void setKey();
 
+    /// Strip the accelerator character from the key
+    QString stripAccelerator(const QString& text);
```
wwylele: This doesn't need to be a member function.

Subv:
Doesnt QPushButton have a setData or setTag or something we can use to store custom values?

gyroninja:
Just checked and it looks like there is a `setProperty` for assigning metadata to them.

makotech222:
I think there were reports of people having problems binding keys. I never got to look into it myself though, so maybe this was the cause.

gyroninja:
Ok, added a property to store the key that was binded instead of the janky manipulation.

NOTE: I did not check if the variant returned by `property("key")` was invalid or not (or even if it was a QString. This shouldn't be an issue because properties for each button are set during the widgets constructor. We can also assume that no one else would set our "key" property to some random object.

If you do want to me to add some checking in to make sure 100% that what we are getting back is valid (which right now it will 100% always be valid) feel free to ask.

@makotech222 I'm pretty sure this is what other people were experiencing since it doesn't allow you to save the key you just bound (unless it's one of the special ones). Even by the time you are binding a second key the first key can already be incorrect (ex. H->&H). This PR makes sure that none of this  funny business takes place.

src/citra_qt/configure_input.cpp
```diff
@@ -68,7 +68,8 @@ void ConfigureInput::handleClick() {
 
 void ConfigureInput::applyConfiguration() {
     for (int i = 0; i < Settings::NativeInput::NUM_INPUTS; ++i) {
-        int value = getKeyValue(input_mapping[Settings::NativeInput::Values(i)]->text());
+        int value = getKeyValue(
+            input_mapping[Settings::NativeInput::Values(i)]->property("key").value<QString>());
```
Subv: Perhaps store the variant value in a variable to shorten this line

gyroninja: Created a QVariant to store it for each instance that we read this property.

yuriks:
We appreciate the contribution, but I think #2295 solves this in a more straightforward way, so I'll close it in favor of that PR.

PR #2291, opened 2016-12-09
service: Add the cfg:nor service

lioncash:


src/core/hle/service/cfg/cfg_nor.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include "core/hle/service/cfg/cfg.h"
+#include "core/hle/service/cfg/cfg_nor.h"
+
+namespace Service {
+namespace CFG {
+
+const Interface::FunctionInfo FunctionTable[] = {
+    // cfg common
+    {0x00010082, GetConfigInfoBlk2, "GetConfigInfoBlk2"},
```
mailwl: i am not sure cfg:nor has common block ( see: 10082 and 10040 - one index for different functions )

wwylele: Agree, according to [3dbrew](https://www.3dbrew.org/wiki/Config_Services) cfg:nor doesn't contain common cfg functions

lioncash: @wwylele At the top of the page where the common functions are, it says: `These commands are common to all Config services.`

wwylele: @lioncash Then the page is conflicting to itself, because "cfg:nor" also contains `0x00010040` function with same main ID 1 as `GetConfigInfoBlk2`, which is almost impossible because most real system modules dispatch the function call only by main ID

lioncash: Just want to apologize if the ending half of my previous comment came off as standoff-ish or abrasive. I've been on painkillers for a recent operation (wisdom teeth removal), so I'm clearly not in my usual mindset. So yeah, sorry about that.

bunnei:
LGTM

Issue #427, opened 2015-01-06
Qt: Disassembler view *always* loads 125000 lines (500000 bytes) no matter how big the .code is. 


archshift:
Probably due to [this crazy function](https://github.com/citra-emu/citra/blob/master/src/citra_qt/debugger/disassembler.cpp#L102-L136) which populates the disasm with 125000 lines, with no concern for how large the .code actually is.

mattvail:
Quick and dirty solution we had... Maybe @Kingcom can add this too his to-do.
Loading everything is still a bad idea for performance. Ideally this would be dynamic and load only x number of lines before and after the current op being displayed.

JayFoxRox:
Referencing for #1480 (meta issue about improving debugger)

PR #2294, opened 2016-12-09
Reduced size of chunk_size and removed if conditions

tinaq:


yuriks:
The files have trailing whitespace somewhere, please find that and remove it so that CI will run. Also, try shortening the commit message a bit so it fits in one 80-character line

MerryMage:
Related to #427.

JayFoxRox:
The original code already looks pretty bad (it shouldn't have to check overlap, but only add lines in front and in the end which may or may not be zero), but how does this new logic even work?!
What's the rationale for this change?
Why isn't the old logic necessary (anymore)?

Aside from that, while the PR title was shortened, why wasn't the commit title shortened?
Also why is the PR by @tinaq but the commit by @rukuluc ?

tinaq:
@JayFoxRox sorry I had two github accounts and I messed up using the other account for the commit.
As for the new logic, I tested and realized that the code_size is always zero because it was never given a value so the if-statements are pretty much not useful anymore.
I've asked several people, apparently chunk_size was given a magic number and it seems like it is made very big just in case num_rows becomes negative. Since base_address equals address anyways, the final num_rows will equal to chunk_size/4, so I just reduced the chunk_size.

JayFoxRox:
How can it be always zero if it's modified within the code you remove?
It probably only used to be zero because the existing code loaded far too many lines. If you load less lines it will (hopefully) hit that function more often with code_size growing each time. So I can't really follow your logic.

If `code_size` really is always zero and has no other references. Shouldn't we remove that variable altogether?

The old logic, while shitty style wise, looks far more reliable and better designed than the proposed change.

tinaq:
Correct me if I'm wrong but that function is in init() and as far as I can see it is always called just once, and when it's called the code_size is zero.

JayFoxRox:
Then that should be the bug which should be fixed probably..
Code could change at runtime with CRO etc. right?

That said, I'm absolutely unfamiliar with this piece of code.

tinaq:
Shouldn't init() be only called once?
init() is called once a game is loaded, if it's restarted, a new model is initialized.
I think I will see what others say then.

wwylele:
What's the status of this PR?

tinaq:
People stopped replying to it so it's just sitting there now I guess

yuriks:
This has been obsoleted by #2689

PR #2295, opened 2016-12-10
configure_input: Modernize and cleanup input configuration tab

### LINKED ISSUE #2289: configure_input: Setting strings should not be translated
Poor design in `configure_input`.

[`ConfigureInput::getKeyName`](https://github.com/citra-emu/citra/blob/84fbbe26297652d994d203bde543ec252c2d801a/src/citra_qt/configure_input.cpp#L108-L121) returns translated strings. This is not an issue in itself, but these English versions of these strings are subsequently used retrieve the appropriate `Qt::Key` value with [`ConfigureInput::getKeyValue`](https://github.com/citra-emu/citra/blob/84fbbe26297652d994d203bde543ec252c2d801a/src/citra_qt/configure_input.cpp#L71). This is not a round-trip conversion in the general case.

MerryMage:
* Removed use of raw QTimer* pointer.
* Update to use type-safe QObject::connect.
* getKeyName can be a static local function.
* Prefer to use function arguments instead of member variables.
* Store Qt::Key instead of converting string back into keycode.

This file was in need of some spring-cleaning.

Closes #2289.
Closes #2290.

While this is a UI code change, no screenshot is provided as there is no modification to the appearance of the UI.


gyroninja:
I can confirm that it fixes the same issue #2290 addressed.

src/citra_qt/configure_input.cpp
```diff
@@ -7,13 +7,29 @@
...
+    if (key_code == -1)
+        return "";
+
+    return QKeySequence(key_code).toString();
+}
+
 ConfigureInput::ConfigureInput(QWidget* parent)
-    : QWidget(parent), ui(std::make_unique<Ui::ConfigureInput>()) {
+    : QWidget(parent), ui(std::make_unique<Ui::ConfigureInput>()),
+      timer(std::make_unique<QTimer>(this)) {
```
yuriks: Per Qt's object ownership model, since this is the parent of the timer, the timer will automatically be destroyed with the dialog. I think that this is removed if it's destructed manually, so this won't crash, but it still makes either the unique_ptr or the parent relationship redundant.

src/citra_qt/configure_input.h
```diff
@@ -30,35 +31,28 @@ class ConfigureInput : public QWidget {
...
-private slots:
-    /// Event handler for all button released() event.
-    void handleClick();
+    /// Each input is represented by a QPushButton.
+    std::map<Settings::NativeInput::Values, QPushButton*> button_map;
+    /// Each input is configured to respond to a press of this Qt::Key.
+    std::map<Settings::NativeInput::Values, Qt::Key> key_map;
 
+    /// Load configuration settings.
+    void setConfiguration();
```
yuriks: `loadConfiguration` would be a better name for this

src/citra_qt/configure_input.cpp
```diff
@@ -40,114 +56,87 @@ ConfigureInput::ConfigureInput(QWidget* parent)
...
 }
 
-void ConfigureInput::restoreDefaults() {
-    for (int i = 0; i < Settings::NativeInput::NUM_INPUTS; ++i) {
-        const QString keyValue = getKeyName(Config::defaults[i].toInt());
-        input_mapping[Settings::NativeInput::Values(i)]->setText(keyValue);
-    }
+void ConfigureInput::setInput(Settings::NativeInput::Values input_id, Qt::Key key_pressed) {
+    // Remove duplicates
+    for (auto& pair : key_map)
```
yuriks: Add braces on the for scope

src/citra_qt/configure_input.cpp
```diff
@@ -40,114 +56,87 @@ ConfigureInput::ConfigureInput(QWidget* parent)
...
-        return tr("Ctrl");
-    if (key_code == Qt::Key_Alt)
-        return tr("Alt");
-    if (key_code == Qt::Key_Meta)
-        return "";
-    if (key_code == -1)
-        return "";
+void ConfigureInput::handleClick(Settings::NativeInput::Values input_id) {
+    QPushButton* button = button_map[input_id];
+    button->setText(tr("[waiting]"));
```
yuriks: "[press key]"? ("waiting" for what?)

src/citra_qt/configure_input.h
MerryMage: Done.

src/citra_qt/configure_input.cpp
MerryMage: Done.

src/citra_qt/configure_input.cpp
MerryMage: Reworded to "[press key]".

src/citra_qt/configure_input.cpp
MerryMage: Addressed by removing parent argument to QTimer's constructor.

MerryMage:
Addressed the above.

linkmauve:
LGTM.

PR #2296, opened 2016-12-10
audio_core: SelectSink should default to auto if sink_id is invalid

### LINKED ISSUE #2279: citra (not qt) empty value audio in config not select default 'auto'
in sdl2-config.ini:

```
[Audio]
# Which audio output engine to use.
# auto (default): Auto-select, null: No audio output, sdl2: SDL2 (if available)
output_engine = 
```

error in log:
`[   0.007132] Audio <Error> audio_core\audio_core.cpp:AudioCore::SelectSink:72: AudioCore::SelectSink given invalid sink_id`

MerryMage:
Closes #2279.

### DIFF
--- a/src/audio_core/audio_core.cpp
+++ b/src/audio_core/audio_core.cpp
@@ -56,22 +56,17 @@ void AddAddressSpace(Kernel::VMManager& address_space) {
 }
 
 void SelectSink(std::string sink_id) {
-    if (sink_id == "auto") {
-        // Auto-select.
-        // g_sink_details is ordered in terms of desirability, with the best choice at the front.
-        const auto& sink_detail = g_sink_details.front();
-        DSP::HLE::SetSink(sink_detail.factory());
-        return;
-    }
-
     auto iter =
         std::find_if(g_sink_details.begin(), g_sink_details.end(),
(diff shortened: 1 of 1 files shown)


src/audio_core/audio_core.cpp
```diff
@@ -56,21 +56,18 @@ void AddAddressSpace(Kernel::VMManager& address_space) {
...
-    if (iter == g_sink_details.end()) {
-        LOG_ERROR(Audio, "AudioCore::SelectSink given invalid sink_id");
-        DSP::HLE::SetSink(std::make_unique<NullSink>());
+    if (sink_id == "auto" || iter == g_sink_details.end()) {
+        if (sink_id != "auto") {
+            LOG_ERROR(Audio, "AudioCore::SelectSink given invalid sink_id %s", sink_id.c_str());
+        }
+        // Auto-select.
+        // g_sink_details is ordered in terms of desirability, with the best choice at the front.
+        const auto& sink_detail = g_sink_details.front();
```
Subv: can't you set the iterator and just let the outer part of the if handle the rest?

MerryMage: Yes, that's a thing that can be done.

MerryMage: Done.

PR #2298, opened 2016-12-10
Add all services to the Service namespace

lioncash:
Previously there was a split where some of the services were in the Service namespace and others were not.


src/core/hle/service/ac_u.cpp
```diff
@@ -2,14 +2,14 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <array>
```
InusualZ: This is used?

lioncash: Line 15

InusualZ: Ops, my bad.

src/core/hle/service/gsp_gpu.cpp
```diff
@@ -18,13 +18,11 @@
...
-const static u32 REGS_BEGIN = 0x1EB00000;
-
-////////////////////////////////////////////////////////////////////////////////////////////////////
-// Namespace GSP_GPU
+namespace Service {
+namespace GSP {
 
-namespace GSP_GPU {
+// Beginning address of HW regs
+const u32 REGS_BEGIN = 0x1EB00000;
```
linkmauve: What is the reason for dropping the static specifier here?  This symbol doesn’t seem to be used anywhere else.

lioncash: const internally links by default; static isn't necessary.

src/core/hle/service/soc_u.cpp
```diff
@@ -53,12 +53,10 @@
...
 
-static const s32 SOCKET_ERROR_VALUE = -1;
-
-////////////////////////////////////////////////////////////////////////////////////////////////////
-// Namespace SOC_U
+namespace Service {
+namespace SOC {
 
-namespace SOC_U {
+static const s32 SOCKET_ERROR_VALUE = -1;
```
linkmauve: Then this one should drop static, right?

lioncash: Done

PR #2302, opened 2016-12-11
game_list: Minor changes

lioncash:
Each commit should be fairly self-explanatory.


src/citra_qt/game_list.h
```diff
@@ -34,7 +34,7 @@ class GameList : public QWidget {
     void LoadInterfaceLayout();
 
 public slots:
```
yuriks: Do we still need these to be slots if we're using the new connection syntax?

lioncash: Nah, the new syntax lets you bind regular functions as a slot. I can make them all private functions if that's preferable.

yuriks: Might as well.

lioncash: Done

PR #2303, opened 2016-12-11
Copy SDL2.dll when compiling citra-qt with msvc

freiro:
CMake now properly copies SDL2.dll when compiling citra with citra-qt as a target
on visual studio.


src/citra_qt/CMakeLists.txt
```diff
@@ -107,27 +107,27 @@ if(UNIX AND NOT APPLE)
...
-        Qt5Core$<$<CONFIG:Debug>:d>.*
-        Qt5Gui$<$<CONFIG:Debug>:d>.*
-        Qt5OpenGL$<$<CONFIG:Debug>:d>.*
-        Qt5Widgets$<$<CONFIG:Debug>:d>.*
-    )
-    windows_copy_files(citra-qt ${Qt5_PLATFORMS_DIR} ${PLATFORMS} qwindows$<$<CONFIG:Debug>:d>.*)
-
-    unset(Qt5_DLL_DIR)
-    unset(Qt5_PLATFORMS_DIR)
+    windows_copy_files(citra-qt ${SDL2_DLL_DIR} ${DLL_DEST} SDL2.dll)
```
lioncash: You might want to consider extracting this code out into a new function. Especially if manually unsetting variables is being done. functions open a new scope, you won't need to care about doing that.

CMakeModules/CopyMSVCLibsQt.cmake
```diff
@@ -0,0 +1,19 @@
+function(copy_msvc_libraries)
+    include(WindowsCopyFiles)
+    set(DLL_DEST "${CMAKE_BINARY_DIR}/bin/$<CONFIG>/")
+    windows_copy_files(citra-qt ${SDL2_DLL_DIR} ${DLL_DEST} SDL2.dll)
```
JayFoxRox: Why is SDL stuff in a file called "CopyMSVCLibsQt" ?

CMakeModules/CopyMSVCLibsQt.cmake
```diff
@@ -0,0 +1,19 @@
...
+
+    set(Qt5_DLL_DIR "${Qt5_DIR}/../../../bin")
+    set(Qt5_PLATFORMS_DIR "${Qt5_DIR}/../../../plugins/platforms/")
+    set(PLATFORMS ${DLL_DEST}platforms/)
+    windows_copy_files(citra-qt ${Qt5_DLL_DIR} ${DLL_DEST}
```
JayFoxRox: I've also discussed this in the past: This is really ugly as CMake has it's own copy tool.
However, I failed to write this properly because it's a nightmare to debug.

Here is my lousy attempt: https://github.com/JayFoxRox/citra/pull/35/commits/ab39b2ef911a4a96f57aa56171e27f008ab6eb8b

CMakeModules/CopyMSVCLibsQt.cmake
freiro: I added "Qt" because this is realted to citra-qt and also to differentiate it from the future "SDL" companion (a similar work needs to be done for "citra" target).
Suggestions for the name are welcome,

CMakeModules/CopyMSVCLibsQt.cmake
```diff
@@ -0,0 +1,19 @@
...
+        icudt*.dll
+        icuin*.dll
+        icuuc*.dll
+        Qt5Core$<$<CONFIG:Debug>:d>.*
+        Qt5Gui$<$<CONFIG:Debug>:d>.*
+        Qt5OpenGL$<$<CONFIG:Debug>:d>.*
+        Qt5Widgets$<$<CONFIG:Debug>:d>.*
```
JayFoxRox: This shouldn't be a module imo. This is very (too) specific to our needs.
We should check if Qt5 has a tool / script for this which is more generic or give it a Citra specific name?
(others might not agree though, so wait for more feedback on this)

JayFoxRox:
I like the idea of the PR but for a small hotfix there are too many changes (but it still feels incomplete).
For a large change there are too many issues with the new design.

Just my 2 cents though.

---

(I think the whole packaging / getting deps is really horrible in Citra anyway.
CMake provides it's own tool to do dependency collection and packing (CPack?) apparently.
So a lot of what we do with travis-upload (platform dependent packing scripts) or very deeply hidden in the citra-qt cmake file (such as our own platform dependent copy file) looks very.. weird. We should fix it properly instead of adding to it)

freiro:
@JayFoxRox if the first/second commits are acceptable we can split this PR in two and discuss modules and improvments later.

freiro:
@JayFoxRox 

> <me> JayFoxRox, to conclude, let's leave it as a module, and let's make the code less ugly in a different PR but let's change now the name of the module
> <me> correct?

JayFoxRox:
Yes and please have seperate functions or even modules for SDL2 and Qt DLL copy so it's more modular.

freiro:
addresses all suggestions, let me know.

src/citra_qt/CMakeLists.txt
```diff
@@ -107,27 +108,10 @@ if(UNIX AND NOT APPLE)
     install(TARGETS citra-qt RUNTIME DESTINATION "${CMAKE_INSTALL_PREFIX}/bin")
 endif()
 
-if (Qt5_FOUND AND MSVC)
-    include(WindowsCopyFiles)
+include(CopyCitraQt5Deps)
+include(CopyCitraSDLDeps)
```
linkmauve: You should probably only include these in the if block, since you don’t use them outside.

src/citra/CMakeLists.txt
```diff
@@ -27,12 +29,7 @@ if(UNIX AND NOT APPLE)
     install(TARGETS citra RUNTIME DESTINATION "${CMAKE_INSTALL_PREFIX}/bin")
 endif()
 
+include(CopyCitraSDLDeps)
```
linkmauve: Same here.

freiro:
@linkmauve done

PR #2307, opened 2016-12-11
Cmake config flexible

FernandoS27:
Added flexibility to cmake configuration. You can now tune builds for specific optimization levels as well as specific instruction sets. Tested on MINGW and MSVC. Should not cause troubles on Linux or CLANG (will add tuning options to CLANG later)

Default options remain the same. Tuning ONLY Affects RELEASE Builds, they do not affect Debug options. Also I suggest adding the /Zl flag to debug in order to be able to profile code better in Visual Studio.

MINGW builds now build with O2 MMX, SSE and SSE2 by default (since there's no debug targets). All other options are off by default and shouldn't be used on official builds


CMakeLists.txt
```diff
@@ -40,6 +40,56 @@ option(CITRA_USE_BUNDLED_SDL2 "Download bundled SDL2 binaries" OFF)
...
+    set(CC_CPU_FLAGS "/arch:SSE2 /arch:AVX2")
+  endif()
+elseif (MINGW)
+  set(CC_EXTRA "${CC_EXTRA} -pthread")
+  set(CC_OP_LEVEL "-O2")
+  if (ENABLE_HIGH_OP)
+    #Citra benefits a lot from this option as my testing have conducted
+    set(CC_OP_LEVEL "-O3")
+  endif()
+  if (CPU_TARGET MATCHES "skylake")
```
freiro: This looks over-complicated, you should use `-march=native`.

JayFoxRox: that wouldn't work on a build-server though ;)

FernandoS27: that could be an option but native means tuning to your own cpu. I wanted options to be used by knowledgeable developers since tuning improves the effectiveness of the in-built debugging tools.

JayFoxRox:
I don't really see the benefit over using `cmake -D CMAKE_BUILD_TYPE=Release -D CMAKE_CXX_FLAGS_RELEASE="-O3 -march=native -mtune=native` etc.

A majority of our users will never use or see these flags anyway.
Most people building from source will know how to tune their compiler and we can't (shouldn't) tune our binary builds as [explained earlier on the forums](https://discuss.citra-emu.org/d/1108-question-on-pull-request/2).

yuriks:
Nay to this. Significantly bloats up the CMake file, and setting most of these flags probably doesn't even actually help with performance in a significant way. Given what our bottlenecks are, enabling AVX, for example, won't help at all.

FernandoS27:
The idea of the variables is to have multicompiler tuning as you know tuning options are not the same in MSVC, GCC and CLANG. By providing options, we can abstract those and save time to many on research and knowledge. Also many people don't know about the intrinsics of optimizations and which are beneficial and which aren't. I feel more comfortable by providing such options to most users.

yuriks:
@FernandoS27 If people don't know what they're doing when they're messing around with flags, then *maybe* they just shouldn't?

JayFoxRox:
That's what toolchain files and CMake build types (e.g. Release / RelWithDebInfo etc.) are for though.
The CMakeLists.txt per project should do the bare minimum to set which files / deps / flags are necessary for compilation etc.

By setting those flags in the CMake file you actually take away the flexibility offered by a toolchain.
(Making it significantly harder to cross-compile etc.)

CMakeLists.txt
JayFoxRox: What build-in debugging tools of what? Do you have any data to back this claim?

FernandoS27:
Hmn I see, I find it beneficial since I like testing builds on all 3 OS and I hate having to go through which flags are which for each compiler. I designed this in such way you can still add new flags to cmake through -D, I feel this adds instead of removes.

lioncash:
Agreeing with @yuriks on this one, as well as @JayFoxRox . If you're looking to tune your compiler build options, you should probably reasonably know what you're doing. Having a blanket option labeled to as 'highest optimization' is kind of a misnomer as it isn't guaranteed to produced the best optimizations, especially given the wide set of flags GCC and Clang provide.

Also look at this from the side of maintainability; no one (where no one means none of the citra core devs) will actively maintain this massive list of compiler flags.

CMakeLists.txt
FernandoS27: Well, I've been trying to use the debug PICA surfaces dissambler and other tools with much ease. Tuning to haswell (my cpu) increases these tools responsiveness by twice the speed compared to O2 , it feels a lot better to work with them in this state.

I do, however, agree with yuriks below. There's too much code bloating and the cmake building could definetly be better refactored. I would prefer to discard my pull request until the cmakelists be refactored into a more elegant manner.

PR #2308, opened 2016-12-12
Service/AC: add ac:i service

mailwl:
ac:i service is used by System Settings application for network configuring


src/core/hle/service/ac/ac.cpp
```diff
@@ -0,0 +1,181 @@
...
+    const u32 version = cmd_buff[1];
+    self->SetVersion(version);
+
+    LOG_WARNING(Service_AC, "(STUBBED) called, version: 0x%08X", version);
+
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+}
+
+void Init() {
+    using namespace Kernel;
```
lioncash: Nothing in the Kernel namespace is used here.

src/core/hle/service/ac/ac.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    ac_connected = false;
+
+    close_event = nullptr;
+    connect_event = nullptr;
+    disconnect_event = nullptr;
+}
+
+void Shutdown() {
+    close_event = nullptr;
```
lioncash: `ac_connected` should probably also be set to false here.

mailwl:
like some other services which has user/system variants, both have common part, and system service has additional functions, like:
`unknown / unimplemented function '0x046A0002': port=ac:i, cmd_buff[1]=0x83C02, cmd_buff[2]=0x2AF454`
i may add this functions without description to 3dbrew

lioncash:
> like some other services which has user/system variants, both have common part, and system service...

Yes, I know. I was bringing up the fact that they weren't documented, which is the issue in question.

src/core/hle/service/ac/ac.cpp
```diff
@@ -0,0 +1,182 @@
...
+    const u32 version = cmd_buff[1];
+    self->SetVersion(version);
+
+    LOG_WARNING(Service_AC, "(STUBBED) called, version: 0x%08X", version);
+
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+}
+
+void Init() {
+
```
lioncash: Unnecessary newline.

lioncash:
@mailwl Note that you can use `git commit --amend` to update a commit instead of making several new ones.

mailwl:
ah, thanks

bunnei:
Rebase please

mailwl:
rebased

bunnei:
Comment from IRC:

<bunnei> Lioncash, thanks. That excludes your general comment regarding ac:i not being on 3dbrew ofc, but IMO it's probably fine.. Regardless, if we want to be really conservative about it, we still want the restructure to match the other services (and according to 3dbrew ac:i exists), so we could just leave the ac:i FunctionTable empty

Will let @lioncash (and/or others) weigh in on this before merging.

mailwl:
fixed dates

PR #2309, opened 2016-12-12
Convert shader JIT to Xbyak

yuriks:
~Currently this crashes because the Citra and Dynarmic versions of Xybak conflict and cause ODR violations. Use merrymage/dynarmic#62 to fix this.~ Since merged and added to the PR.

Unified diff is probably more pleasant than Split for this one. I tried to keep this code a straight semantic port from the previous one. I'll make some cleanups in following PRs.

### DIFF
--- a/src/common/x64/xbyak_abi.h
+++ b/src/common/x64/xbyak_abi.h
@@ -1,0 +1,178 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <initializer_list>
+#include <xbyak.h>
+#include "common/assert.h"
+#include "common/bit_set.h"
+
+namespace Common {
+namespace X64 {
+
+int RegToIndex(const Xbyak::Reg& reg) {
+    using Kind = Xbyak::Reg::Kind;
+    ASSERT_MSG((reg.getKind() & (Kind::REG | Kind::XMM)) != 0,
+               "RegSet only support GPRs and XMM registers.");
+    ASSERT_MSG(reg.getIdx() < 16, "RegSet only supports XXM0-15.");
+    return reg.getIdx() + (reg.getKind() == Kind::REG ? 0 : 16);
+}
+
+inline Xbyak::Reg64 IndexToReg64(int reg_index) {
+    ASSERT(reg_index < 16);
+    return Xbyak::Reg64(reg_index);
+}
+
+inline Xbyak::Xmm IndexToXmm(int reg_index) {
+    ASSERT(reg_index >= 16 && reg_index < 32);
+    return Xbyak::Xmm(reg_index - 16);
+}
+
+inline Xbyak::Reg IndexToReg(int reg_index) {
+    if (reg_index < 16) {
+        return IndexToReg64(reg_index);
+    } else {
+        return IndexToXmm(reg_index);
+    }
+}
+
+inline BitSet32 BuildRegSet(std::initializer_list<Xbyak::Reg> regs) {
+    BitSet32 bits;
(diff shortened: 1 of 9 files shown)


src/common/x64/xbyak_abi.h
```diff
@@ -0,0 +1,213 @@
...
+
+inline Xbyak::Xmm IndexToXmm(int reg_index) {
+    ASSERT(reg_index >= 16 && reg_index < 32);
+    return Xbyak::Xmm(reg_index - 16);
+}
+
+inline Xbyak::Reg IndexToReg(int reg_index) {
+    if (reg_index < 16) {
+        return IndexToReg64(reg_index);
+    } else {
```
lioncash: else isn't necessary here, since the bodies are stacked returns.

yuriks: I like the symmetry of the else here. There's two cases and they're both equally important.

yuriks:
Xbyak isn't a module because I was originally going to use a different file structure where it'd get it's own include directory (so you'd use paths like <xbyak/xbyak.h>). But then I realized all files have a xbyak_ prefix anyway, so that would be kinda pointless. I'll change it.

src/common/x64/xbyak_abi.h
lioncash: I don't really understand the notion that not including redundant control specifiers means a particular branch is less important, but OK.

I get the symmetry one though, I suppose.

src/common/x64/xbyak_util.h
```diff
@@ -0,0 +1,49 @@
...
+    return !(distance >= 0x8000'0000ULL && distance <= ~0x8000'0000ULL);
+}
+
+inline bool IsWithin2G(const Xbyak::CodeGenerator& code, uintptr_t target) {
+    return IsWithin2G(reinterpret_cast<uintptr_t>(code.getCurr()), target);
+}
+
+template <typename T>
+inline void CallFarFunction(Xbyak::CodeGenerator& code, const T f) {
+    static_assert(std::is_pointer<T>(), "Argument must be a (function) pointer.");
```
lioncash: I missed this when initially reviewing the first time, but you can do:

```
std::is_pointer<T>() && std::is_function<std::remove_pointer_t<T>>()
```

as your static assert to ensure the given type is a function pointer. If the JIT just uses opaque pointers to functions (ew), disregard this.

yuriks: I didn't add that other check in case the caller *does* have an opaque pointer (which is not our case but I could totally see it happening), so they can call this without having to cast it to some dummy function type. (Since this function doesn't care about the signature or calling convention of the pointer at all.)

src/video_core/shader/shader_jit_x64.h
```diff
@@ -103,17 +104,16 @@ class JitShader : public Gen::XCodeBlock {
     void FindReturnOffsets();
 
     /// Mapping of Pica VS instructions to pointers in the emitted code
+    // TODO(yuriks): instruction_labels already keeps track of this, but xbyak doesn't let us it
```
yuriks: Fixed

yuriks:
Updated:
 - Small cleanup with the register types
 - Fixed formatting errors
 - Made xbyak a git submodule
 - Use `-fno-operator-names` on non-MSVC

bunnei:
LGTM, other than the build issues ofc on GCC/clang

yuriks:
Fixed build. Also, since herumi added a feature I requested (herumi/xbyak#43) I've taken care of that TODO, if anyone cares enough to review it. Otherwise, I'll merge this tomorrow.

bunnei:
Reviewed the recent changes. Still LGTM

MerryMage:
Looks fine to me.

PR #2310, opened 2016-12-12
Dynarmic: Update submodule commit to 34e19f135c0dd2feac4f77660f51aa4ea28a7386

Dragios:
Dynarmic Changelog:
```
[Commits on Dec 12, 2016]
34e19f1   CMake: Re-use external xbyak target if present (#62)
5bea2e1   block_of_code: Support stack unwinding on Windows
4962d92   block_of_code: Do not regenerate prelude when clearing cache
2a1cf94   CMakeLists: Include backend_x64 only if we're targeting x86_64
dcc880a   assert: _a_ expression string shouldn't be part of the format string
179a338   block_of_code: Provide an alloc function to allocate space in the code block

[Commits on Dec 6, 2016]
f467589   emit_x64: Remove unnecessary casts
a37631c   emit_x64: Use reinterpret_cast for pointer casts

[Commits on Dec 5, 2016]
fafa845   emit_x64: Make GetBasicBlock() const qualified
6a16edc   emit_x64: Move implementations into the cpp file
282029f   emit_x64: Forward declare BlockOfCode
6898b74   emit_x64: Get rid of indirect includes
1dc1eaa   CMakeLists: Enable more warnings for MSVC (#58) 
```


Dragios:
Don't merge yet, let me update it 1st.

JayFoxRox:
I think the amount of update PRs for dynarmic is getting a little out of hand.
We should probably also leave updating dynarmic to Merry or dynarmic contributors.
If they are aware of important changes we can update it once in a while.

Also keep in mind that dynarmic doesn't get as much review as the main repo (I think?) and each submodule update has a lot of changes. So if there was broken code (or possibly even dangerous / malicious code) it would be far easier to sneak changes in through an external like dynarmic.

The last dynarmic PRs already got barely tested / reviewed.

While this might work at the moment, with few contributors to dynarmic (who we trust), this is a horrible trend for the future.
dynarmic should get the same kind of regression testing and (albeit shorter) review like most other PRs.
However, if that happens we can't be doing dynarmic updates 2-3 times a week.
So please do not get used to updating it that often.

Dragios:
There is one dynarmic update which is a good thing for discovering new bug I guess?
https://github.com/MerryMage/dynarmic/pull/58

MerryMage:
Thanks but no thanks.

There's not really any reason for citra to track dynarmic updates closely. imo cases that would require a update would be (a) bug fixes, (b) major behavioural updates and (c) interface changes.

If a PR requires a submodule update that update should be part of that PR and not a separate PR.

PR #2314, opened 2016-12-13
Service/ACT: move ACT services to folder

mailwl:


src/core/hle/service/act/act.h
```diff
@@ -0,0 +1,20 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+namespace Service {
+
+class Interface;
```
bunnei: Is this forward declaration used at all?

mailwl: ATM no, but later i want to stub some functions in next PR, there is needed
https://github.com/mailwl/citra/blob/account-stub/src/core/hle/service/act/act.h

src/core/hle/service/act/act.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Service {
+namespace ACT {
+
+void Init() {
+    AddService(new ACT_A);
+    AddService(new ACT_U);
+}
+
+void Shutdown() {}
```
yuriks: Leave the shutdown function out if it does nothing.

src/core/hle/service/act/act.h
yuriks: Can you just add it in the next PR? This makes git blames, etc. kinda confusing.

mailwl:
done

PR #2316, opened 2016-12-13
Common: Fix gcc build on macOS

endrift:
A regrettably ugly fix for an asinine but seemingly unavoidable issue.


endrift:
As I said in IRC earlier, if this is a configuration you don't care about supporting, feel free to close this PR.

PR #2318, opened 2016-12-15
VideoCore: Inline IsPicaTracing

yuriks:
Speeds up ALBW main menu slightly (~3%)

### DIFF
--- a/src/video_core/debug_utils/debug_utils.cpp
+++ b/src/video_core/debug_utils/debug_utils.cpp
@@ -276,10 +276,10 @@ void DumpShader(const std::string& filename, const Regs::ShaderConfig& config,
 
 static std::unique_ptr<PicaTrace> pica_trace;
 static std::mutex pica_trace_mutex;
-static int is_pica_tracing = false;
+bool g_is_pica_tracing = false;
 
 void StartPicaTracing() {
-    if (is_pica_tracing) {
+    if (g_is_pica_tracing) {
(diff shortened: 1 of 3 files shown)


src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -276,45 +276,37 @@ void DumpShader(const std::string& filename, const Regs::ShaderConfig& config,
...
-    return is_pica_tracing != 0;
+    g_is_pica_tracing = true;
 }
 
 void OnPicaRegWrite(PicaTrace::Write write) {
-    // Double check for is_pica_tracing to avoid pointless locking overhead
-    if (!is_pica_tracing)
-        return;
-
     std::lock_guard<std::mutex> lock(pica_trace_mutex);
```
JayFoxRox: scope around this one so it only gets created when tracing is activated?

JayFoxRox: nvm, just noticed that there is only one user now and the following if check will always be true..
Makes me wonder if we should just seperate `is_pica_tracing` from this fuction entirely, then document that caller is responsible for only calling it during tracing.

yuriks: I'm not sure what you mean? (This only gets called when tracing is active in the first place.)

yuriks: The second check is there to ensure that the value is correct while the lock is held. It's a crappy implementation of double-checked locking.

JayFoxRox:
LGTM

PR #2322, opened 2016-12-15
game_list: Add a context menu with "Open Save Location"  option

MerryMage:
What changed:
* Added `Loader::GetProgramId`.
* Added static function `ArchiveSource_SDSaveData::GetSaveDataPathFor`.
* Added signal `GameList::OpenSaveFolderRequested`.

Issues that arose:
* The least messy way I could think of to expose a way to get the save data path was to implement a static function.
* I really really wanted to use a `boost::optional<u64>` to represent the `program_id` since not all formats have a program id.

Screenshot:

### DIFF
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <QHeaderView>
+#include <QMenu>
 #include <QThreadPool>
 #include <QVBoxLayout>
 #include "common/common_paths.h"
@@ -28,6 +29,7 @@ GameList::GameList(QWidget* parent) : QWidget{parent} {
     tree_view->setSortingEnabled(true);
     tree_view->setEditTriggers(QHeaderView::NoEditTriggers);
     tree_view->setUniformRowHeights(true);
+    tree_view->setContextMenuPolicy(Qt::CustomContextMenu);
 
     item_model->insertColumns(0, COLUMN_COUNT);
     item_model->setHeaderData(COLUMN_NAME, Qt::Horizontal, "Name");
@@ -35,10 +37,10 @@ GameList::GameList(QWidget* parent) : QWidget{parent} {
     item_model->setHeaderData(COLUMN_SIZE, Qt::Horizontal, "Size");
 
     connect(tree_view, &QTreeView::activated, this, &GameList::ValidateEntry);
+    connect(tree_view, &QTreeView::customContextMenuRequested, this, &GameList::PopupContextMenu);
 
     // We must register all custom types with the Qt Automoc system so that we are able to use it
-    // with
-    // signals/slots. In this case, QList falls under the umbrells of custom types.
+    // with signals/slots. In this case, QList falls under the umbrells of custom types.
     qRegisterMetaType<QList<QStandardItem*>>("QList<QStandardItem*>");
 
     layout->addWidget(tree_view);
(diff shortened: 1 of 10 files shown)


src/citra_qt/game_list.cpp
```diff
@@ -71,6 +73,23 @@ void GameList::DonePopulating() {
...
+    QModelIndex item = tree_view->indexAt(menu_location);
+    if (!item.isValid())
+        return;
+
+    int row = item_model->itemFromIndex(item)->row();
+    QStandardItem* child_file = item_model->invisibleRootItem()->child(row, COLUMN_NAME);
+    u64 program_id = child_file->data(GameListItemPath::ProgramIdRole).toULongLong();
+
+    QMenu context_menu;
+    QAction* open_save_location = context_menu.addAction(tr("Open Save Location"));
```
bunnei: nit: For some reason "Open Save Location" sounds weird to me ...  But perhaps that's just me

src/citra_qt/main.cpp
```diff
@@ -460,6 +464,21 @@ void GMainWindow::OnGameListLoadFile(QString game_path) {
...
+    QString qpath = QString::fromStdString(path);
+
+    QDir dir(qpath);
+    if (!dir.exists()) {
+        QMessageBox::critical(this, tr("Error Opening Save Folder"), tr("Folder does not exist!"));
+        return;
+    }
+
+    LOG_INFO(Frontend, "Opening save data path for program_id=%" PRIu64, program_id);
+    QDesktopServices::openUrl(QUrl::fromLocalFile(qpath));
```
bunnei: This should work on all three major platforms, right?

src/citra_qt/game_list.cpp
lioncash: I've usually seen `Open Folder Location` used.

src/core/loader/loader.h
```diff
@@ -144,6 +144,15 @@ class AppLoader : NonCopyable {
     }
 
     /**
+     * Get the program id of the application
+     * @param out_program_id Reference to store program id into
+     * @return ResultStatus result of function
+     */
+    virtual ResultStatus ReadProgramId(u64& out_program_id) {
```
bunnei: So why didn't you use boost::optional? (or std::optional, is that not available on all compilers yet?)
I see you're just following the pattern already used in this file, so I guess it's fine for now, but the loader interface could use some improvements

src/citra_qt/main.cpp
MerryMage: Yes.

src/core/loader/loader.h
MerryMage: `std::optional` is part of C++17.

I imitated the pattern already in the file. I could do a follow up PR that improves on the interface.

src/citra_qt/game_list.cpp
MerryMage: Reworded to `Open Save Data Location`.

MerryMage:
Should update [the guide on the wiki](https://github.com/citra-emu/citra/wiki/Dumping-Save-Data-from-a-3DS-Console) if/when this is merged.

MerryMage:
Should the folder be created if it doesn't exist?

legoj15:
I think that would be a good idea

yuriks:
I think directory creation should be in a separate PR. (I'm not sure if it should even *be* done, since that'll create problems with the lack of save metadata. @Subv says many things actually rely on that being correct.)

yuriks:
@MerryMage feel free to merge this if you're done with it

Issue #2313, opened 2016-12-13
Rendering text in the SDL frontend


lynn:
We currently can’t render text in the SDL frontend: we’re not linking `SDL_ttf` (which is imperfect anyway) and we don’t depend on any libraries that replace it (such as [Pango](http://sdlpango.sourceforge.net/), which is better, but Linux-centric, and a much heavier dependency, according to some people on IRC.)

This would, for example, be useful for #111: the SDL frontend could then have an interactive, SDL-based key config editor. But which library should we use?

Here’s the relevant discussion from `#citra`:

    <linkmauve1> lynn, you shouldn’t use SDL_ttf, it’s a bad wrapper around Freetype that doesn’t support anything useful.
    <linkmauve1> I made that mistake in PyTouhou.
    <linkmauve1> Better use something like pango.
    ...
    <endrift> it's kinda true
    <endrift> SDL_TTF is a really restrictive API that's complete garbage for anything other than the most basic of text rendering
    <endrift> but then
    <endrift> what are you going to be doing other than basic text rendering?
    <neobrain> yeah, precisely that lol
    <neobrain> I used SDL_ttf in the past, and it's certainly not bad enough to warrant blocking useful work.. plus, outputting text on the screen is trivially turned into an interface with flexible implementation
    <linkmauve1> endrift, at some point Citra may want to be translated, for example.
    <linkmauve1> At which case, all of your assumptions will stop holding true.
    <endrift> yep, pango is way better at that
    <merry> you'd probably want to include fontconfig too if you want to use a system font rather than bundling a font
    <merry> SDL_ttf has no mechanism of locating a system font iirc
    <endrift> I think it does now, but it's rather garbage
    <endrift> pango might handle fontconfig stuff
    <neobrain> merry: once you go that route you'll also want to make sure to follow the system's configuration wrt antialiasing/subpixel hinting, at which point you might as well just use Qt already
    <endrift> but idk
    <endrift> I think pango handles all of that for you
    <endrift> or maybe harfbuzz
    <endrift> but pango might depend on harfbuzz
    <endrift> I've never looked too deeply at it
    <WinterMute> pango depends on a load of shit you probably don't want
    <endrift> ^ that's the problem with pango
    <endrift> it's a heavyweight dependency
    <WinterMute> 2.x glib
    <endrift> ughhhh glib
    <endrift> I forgot about that
    * WinterMute attempted a cube pango port once
    <endrift> glib is honestly one of the worst base libraries in existence
[...]

PR #2324, opened 2016-12-15
Add interactive key rebind tool to SDL2 frontend

lynn:
Solves #111.

This involved writing a slightly messy function that updates an .ini file while keeping its comments and whitespace intact. It seems to do a good job.

This tool isn't *great*, right now -- but issue #2313 prevents me from rendering any text in the SDL window that pops up to prompt the user for a single keypress.

Try running `citra -l` to view the current bindings, `src/citra/citra -r pad_x` to update one of them.

### DIFF
--- a/src/citra/rebind_window/rebind_window_sdl2.cpp
+++ b/src/citra/rebind_window/rebind_window_sdl2.cpp
@@ -1,0 +1,130 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+// #include <algorithm>
+#include <cstdlib>
+#include <iostream>
+#include <string>
+#define SDL_MAIN_HANDLED
+#include <SDL.h>
+
+#include "citra/rebind_window/rebind_window_sdl2.h"
+// #include "common/key_map.h"
+#include "common/file_util.h"
+#include "common/logging/log.h"
+#include "common/scm_rev.h"
+#include "common/string_util.h"
+// #include "core/hle/service/hid/hid.h"
+#include "citra/config.h"
+#include "core/settings.h"
+// #include "video_core/video_core.h"
+
+const u32 fps = 60;
+const u32 minimum_frame_time = 1000 / fps;
+const int REBIND_WINDOW_WIDTH = 640;
+const int REBIND_WINDOW_HEIGHT = 480;
+
+void RebindWindow_SDL2::OnKeyDown(int key) {
+    const char* key_name = Settings::NativeInput::Mapping[modifying_index];
+    std::string scancode_name = SDL_GetScancodeName(static_cast<SDL_Scancode>(key));
+    std::cout << "Rebound " << key_name << " to " << scancode_name << "." << std::endl;
+    FileUtil::SetINIKey(config_filename, "Controls", key_name, std::to_string(key));
+    is_open = false;
+}
+
+bool RebindWindow_SDL2::IsOpen() const {
+    return is_open;
+}
+
+RebindWindow_SDL2::RebindWindow_SDL2(int modifying_index, const std::string& config_filename)
+    : modifying_index(modifying_index), config_filename(config_filename) {
+    SDL_SetMainReady();
+
+    // Initialize the window
+    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
+        LOG_CRITICAL(Frontend, "Failed to initialize SDL2! Exiting...");
+        exit(1);
+    }
+
+    std::string window_title = Common::StringFromFormat("Citra | %s-%s | Rebinding keys",
+                                                        Common::g_scm_branch, Common::g_scm_desc);
+
+    render_window = SDL_CreateWindow(window_title.c_str(),
+                                     SDL_WINDOWPOS_UNDEFINED, // x position
+                                     SDL_WINDOWPOS_UNDEFINED, // y position
+                                     REBIND_WINDOW_WIDTH, REBIND_WINDOW_HEIGHT, SDL_WINDOW_SHOWN);
+
+    if (render_window == nullptr) {
+        LOG_CRITICAL(Frontend, "Failed to create SDL2 window! Exiting...");
+        exit(1);
+    }
+
+    renderer = SDL_CreateRenderer(render_window, -1, SDL_RENDERER_ACCELERATED);
+
+    if (renderer == nullptr) {
+        LOG_CRITICAL(Frontend, "Failed to create SDL2 renderer! Exiting...");
+        exit(1);
+    }
+}
+
+RebindWindow_SDL2::~RebindWindow_SDL2() {
+    SDL_DestroyRenderer(renderer);
+    SDL_DestroyWindow(render_window);
+    SDL_Quit();
+}
+
+void RebindWindow_SDL2::PollEvents() {
+    SDL_Event event;
+
+    while (SDL_PollEvent(&event) && is_open) {
+        switch (event.type) {
+        case SDL_WINDOWEVENT:
+            if (event.window.event == SDL_WINDOWEVENT_CLOSE) {
+                is_open = false;
+            }
+            break;
+        case SDL_KEYDOWN:
+            OnKeyDown(static_cast<int>(event.key.keysym.scancode));
+            break;
+        case SDL_QUIT:
+            is_open = false;
+            break;
+        }
+    }
+}
+
+void RebindWindow_SDL2::Render() {
+    SDL_SetRenderDrawColor(renderer, 0xff, 0xff, 0xff, 0xff);
+    SDL_RenderClear(renderer);
+    // TODO: Ideally, this window should be an interactive editor, but at the very least,
+    // display some sort of "press a key" text here. This requires a text rendering library
+    // though, see issue #2313.
+}
+
+void RebindWindow_SDL2::Run() {
+    Uint32 last_frame_time = 0;
+    Uint32 frame_time;
+    Uint32 time_delta;
+
+    int current = Settings::values.input_mappings[Settings::NativeInput::All[modifying_index]];
+    std::cout << "Press a key to bind to " << Settings::NativeInput::Mapping[modifying_index]
+              << " (currently " << SDL_GetScancodeName(static_cast<SDL_Scancode>(current)) << ")..."
+              << std::endl;
+
+    while (is_open) {
+        frame_time = SDL_GetTicks();
+
+        PollEvents();
+
+        time_delta = frame_time - last_frame_time;
+        last_frame_time = frame_time;
+
+        Render();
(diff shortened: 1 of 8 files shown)


src/common/file_util.cpp
```diff
@@ -836,6 +838,83 @@ void SplitFilename83(const std::string& filename, std::array<char, 9>& short_nam
     }
 }
 
+bool SetINIKey(const std::string& filename, const std::string& section, const std::string& key,
```
bunnei: Should we really have to write this function? It seems that INIH is just a parser, perhaps we should use a different library that allows for writing as well?

src/citra/rebind_window/rebind_window_sdl2.h
```diff
@@ -0,0 +1,50 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <utility>
+#include "common/emu_window.h"
+
+struct SDL_Window;
+
+class RebindWindow_SDL2 {
```
bunnei: Why does this class exist? Does SDL require a window to even be created for key input? Even if so, we shouldn't have a whole separate code path for creating a render window just for key binding - it should just use the regular emulator window. We'll probably also want to support rebinding keys in game (once it is overlayed text) triggered by a hot key.

src/common/file_util.cpp
```diff
@@ -836,6 +838,83 @@ void SplitFilename83(const std::string& filename, std::array<char, 9>& short_nam
...
+                continue;
+            } else {
+                // Invalid section header.
+                return false;
+            }
+        }
+
+        // Handle key-value pairs.
+        // We want to keep indentation, here, so work with `original`.
+        size_t equals = original.find("=");
```
lioncash: You can use a character literal here.

src/common/file_util.cpp
```diff
@@ -836,6 +838,83 @@ void SplitFilename83(const std::string& filename, std::array<char, 9>& short_nam
...
+        if (equals != std::string::npos) {
```
lioncash: You can test against the non-negated case first which would allow unindenting the body of the if.

e.g. You can just do.

```cpp
if (equals == std::string::npos)
    return false;

...
```

src/common/file_util.cpp
```diff
@@ -836,6 +838,83 @@ void SplitFilename83(const std::string& filename, std::array<char, 9>& short_nam
...
+            std::string pre = original.substr(0, equals);
+            std::string this_key = Common::ToLower(Common::StripSpaces(pre));
+            if (this_section == target_section && this_key == target_key) {
+                new_contents += pre + "=" + value + "\n";
```
lioncash: character literals can be used here.

src/common/file_util.cpp
```diff
@@ -836,6 +838,83 @@ void SplitFilename83(const std::string& filename, std::array<char, 9>& short_nam
...
+            }
+        } else {
+            // Invalid key-value pair.
+            return false;
+        }
+    }
+
+    if (!updated) {
+        // Create a new section.
+        new_contents += "\n[" + section + "]\n" + key + "=" + value + "\n";
```
lioncash: Character literals can be used for `=` and `\n`.

bunnei:
I don't like how this was written... it feels like you're making an entire app just for configuring keys, and then packing it into Citra. I think this can be done much cleaner/more generically, with much less code. Some overall suggestions:

* We shouldn't have to write our own code to write INI, just like we didn't write our own parser. Find a better library that does both and replace INIH with it. This can probably be a PR of its own.
* We shouldn't have to duplicate all of the SDL boilerplate to get key input, such as creating a window, polling events, etc. Duplicate code just pollutes our code base and is hard to maintain. This also makes it so we'll never be able to rebind keys in-game. 
* This can be done generically (i.e. not specific to SDL frontend), so why wouldn't we want to do this? The CLI version of Citra (the "citra" sub-project that builds with SDL) is meant to be a very thin wrapper around the emulator components, having as little code as possible. One of the benefits of this is it helps to ensure proper design elsewhere, such that we don't become tied too deeply to a frontend.
* In regards to doing this generically, as a prerequisite to this, we probably want a way to get input generically from the frontend (rather than just for game input) - a step 2 (after the INI stuff) might be to change EmuWindow class to have something like an OnKeyEvent(Key key, State state) function that gets called on any key press, allowing us to have generic input handling regardless of the frontend/platform.
* Step 3 - use steps 1 and 2 to implement key rebinding in a common place in like 5 lines of code

src/common/file_util.cpp
```diff
@@ -836,6 +838,83 @@ void SplitFilename83(const std::string& filename, std::array<char, 9>& short_nam
...
+            continue;
+        }
+
+        if (line[0] == '[') {
+            if (line.back() == ']') {
+                // Valid section header.
+
+                if (this_section == target_section && !updated) {
+                    // Create a new key.
+                    new_contents += key + "=" + value + "\n\n";
```
lioncash: `=` can be a character literal.

src/common/file_util.cpp
```diff
@@ -836,6 +838,83 @@ void SplitFilename83(const std::string& filename, std::array<char, 9>& short_nam
...
+    std::istringstream iss(contents);
+
+    std::string new_contents;
+    std::string this_section;
+    std::string target_section = Common::ToLower(section);
+    std::string target_key = Common::ToLower(key);
+
+    bool updated = false;
+    for (std::string line; std::getline(iss, line);) {
+        std::string original = line + "\n";
```
lioncash: `\n` can be a character literal.

src/common/file_util.cpp
```diff
@@ -836,6 +838,83 @@ void SplitFilename83(const std::string& filename, std::array<char, 9>& short_nam
...
+
+        // Strip whitespace, too.
+        line = Common::StripSpaces(line);
+        if (line.empty() || line[0] == '#' || line[0] == ';') {
+            new_contents += original;
+            continue;
+        }
+
+        if (line[0] == '[') {
+            if (line.back() == ']') {
```
lioncash: You can test for the invalid case first here which would let you unindent the valid case.

```cpp
// Invalid section header
if (line.back() != ']')
    return false;

...
```

src/citra/rebind_window/rebind_window_sdl2.h
```diff
@@ -0,0 +1,50 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <utility>
+#include "common/emu_window.h"
+
+struct SDL_Window;
```
lioncash: SDL_Renderer can be forward declared as well.

src/citra/rebind_window/rebind_window_sdl2.h
```diff
@@ -0,0 +1,50 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <utility>
```
lioncash: `<string>` needs to be included as well

src/citra/rebind_window/rebind_window_sdl2.h
```diff
@@ -0,0 +1,50 @@
...
+#include "common/emu_window.h"
```
lioncash: It seems that emu_window.h is unused here and can be removed.

src/citra/rebind_window/rebind_window_sdl2.cpp
```diff
@@ -0,0 +1,130 @@
...
+                                                        Common::g_scm_branch, Common::g_scm_desc);
+
+    render_window = SDL_CreateWindow(window_title.c_str(),
+                                     SDL_WINDOWPOS_UNDEFINED, // x position
+                                     SDL_WINDOWPOS_UNDEFINED, // y position
+                                     REBIND_WINDOW_WIDTH, REBIND_WINDOW_HEIGHT, SDL_WINDOW_SHOWN);
+
+    if (render_window == nullptr) {
+        LOG_CRITICAL(Frontend, "Failed to create SDL2 window! Exiting...");
+        exit(1);
```
lioncash:
```cpp
exit(EXIT_FAILURE);
```

src/citra/rebind_window/rebind_window_sdl2.cpp
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    renderer = SDL_CreateRenderer(render_window, -1, SDL_RENDERER_ACCELERATED);
+
+    if (renderer == nullptr) {
+        LOG_CRITICAL(Frontend, "Failed to create SDL2 renderer! Exiting...");
+        exit(1);
```
lioncash:
```cpp
exit(EXIT_FAILURE);
```

src/citra/rebind_window/rebind_window_sdl2.cpp
```diff
@@ -0,0 +1,130 @@
...
+}
+
+RebindWindow_SDL2::RebindWindow_SDL2(int modifying_index, const std::string& config_filename)
+    : modifying_index(modifying_index), config_filename(config_filename) {
+    SDL_SetMainReady();
+
+    // Initialize the window
+    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
+        LOG_CRITICAL(Frontend, "Failed to initialize SDL2! Exiting...");
+        exit(1);
```
lioncash:
```cpp
exit(EXIT_FAILURE);
```

src/citra/rebind_window/rebind_window_sdl2.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include "citra/rebind_window/rebind_window_sdl2.h"
+// #include "common/key_map.h"
+#include "common/file_util.h"
+#include "common/logging/log.h"
+#include "common/scm_rev.h"
+#include "common/string_util.h"
+// #include "core/hle/service/hid/hid.h"
+#include "citra/config.h"
+#include "core/settings.h"
+// #include "video_core/video_core.h"
```
lioncash: Should likely remove the unused headers here.

src/citra/rebind_window/rebind_window_sdl2.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+const u32 fps = 60;
```
lioncash: These constants can be `constexpr`

lynn:
@bunnei brings up really good points, and I’m not very happy with this solution in light of them. I’m going to think harder about a more generic solution — but `SetINIKey` is going to be replaced by the solution to #2329, and the remaining code is going to be replaced by, well, ten lines of code or something. So I’m closing this first attempt. (Thank you both very much for the review! I’ll definitely keep your pointers in mind for future PRs, @lioncash.)

PR #2328, opened 2016-12-15
Fix debug build from #2249

wwylele:
`GetTypeName()` is not a member anymore.
`LOG_TRACE` often hides mistakes ...


mailwl:
i fix it by `server_session->GetTypeName().c_str()`

wwylele:
@mailwl since `File` and `Directory` are not `Session` anymore and printing it's type "Session" is meaningless, so I decided to remove it.

mailwl:
heh, now i understand, why Release don't show compile error, LOG_TRACE just empty

PR #2331, opened 2016-12-15
hid: Get rid of a double -> float truncation warning

lioncash:
float literals need to have the 'f' prefix.


src/core/hle/service/hid/hid.cpp
```diff
@@ -37,7 +37,8 @@ static int enable_gyroscope_count = 0;     // positive means enabled
 
 static PadState GetCirclePadDirectionState(s16 circle_pad_x, s16 circle_pad_y) {
     // 30 degree and 60 degree are angular thresholds for directions
-    constexpr float TAN30 = 0.577350269, TAN60 = 1 / TAN30;
+    constexpr float TAN30 = 0.577350269f;
+    constexpr float TAN60 = 1 / TAN30;
```
JayFoxRox: Wouldn't we want `1.0f / TAN30` here?

PR #2332, opened 2016-12-15
gdbstub: Remove global variable from public interface

lioncash:
Currently, this is only ever queried, so adding a function to check if the server is enabled is more sensible.

If directly modifying this externally is ever desirable, it should be done by adding a function to the interface, rather than exposing implementation details directly.


yuriks:
I think the reason this was a global was to make it slightly cheaper to check, since it's in the core loop. Probably fine anyway.

PR #2334, opened 2016-12-16
Stdin software keyboard implementation

jroweboy:
Allows users to type into the console for text input, runs several of the validation checks, and returns the correct result codes. Information about the structs comes from ctrulib with minor changes to make more sense in citras codebase.


src/core/hle/service/apt/apt.cpp
```diff
@@ -208,6 +212,11 @@ void ReceiveParameter(Service::Interface* self) {
     u32 buffer_size = cmd_buff[2];
     VAddr buffer = cmd_buff[0x104 >> 2];
 
+    if (cancelled) {
+        cmd_buff[1] = -1;
```
wwylele: `// TODO: find the correct code`
Also I feel like these cancel-related stuff is taken from my hacking code? I didn't do any hwtest and research when coding them...

src/core/hle/applets/swkbd.h
```diff
@@ -15,33 +15,153 @@
...
+    bool predictive_input;
+    bool multiline;
+    bool fixed_width;
+    bool allow_home;
+    bool allow_reset;
+    bool allow_power;
+    bool unknown; // XX: what is this supposed to do? "communicateWithOtherRegions"
+    bool default_qwerty;
+    bool button_submits_text[4];
+    u16 language; // XX: not working? supposedly 0 = use system language, CFG_Language+1 = pick
```
wwylele: I know these comments are taken from ctrulib, but we should reword it when applying them to citra ("not working" is not a meaningful comment to an emulator)

src/core/hle/applets/swkbd.h
```diff
@@ -15,33 +15,153 @@
...
+/// Maximum filter callback error message length, in UTF-16 code units.
+#define SWKBD_MAX_CALLBACK_MSG_LEN 256
+
+/// Keyboard types
+enum class SwkbdType : u32 {
+    NORMAL = 0, ///< Normal keyboard with several pages (QWERTY/accents/symbol/mobile)
+    QWERTY,     ///< QWERTY keyboard only.
+    NUMPAD,     ///< Number pad.
+    WESTERN,    ///< On JPN systems, a text keyboard without Japanese input capabilities,
+                /// otherwise same as SWKBD_TYPE_NORMAL.
```
wwylele: Our code style is to use `CamelCase` for enum members.

src/core/hle/applets/swkbd.h
```diff
@@ -15,33 +15,153 @@
...
-struct SoftwareKeyboardConfig {
-    INSERT_PADDING_WORDS(0x8);
+/// Maximum number of buttons that can be in the keyboard.
+#define SWKBD_MAX_BUTTON 3
+/// Maximum button text length, in UTF-16 code units.
+#define SWKBD_MAX_BUTTON_TEXT_LEN 16
+/// Maximum hint text length, in UTF-16 code units.
+#define SWKBD_MAX_HINT_TEXT_LEN 64
+/// Maximum filter callback error message length, in UTF-16 code units.
+#define SWKBD_MAX_CALLBACK_MSG_LEN 256
```
wwylele: I prefer using `const` (`constexpr`?) instead of `#define`

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+    std::string input;
+    std::cout << "SOFTWARE KEYBOARD" << std::endl;
+    // Display hint text
+    std::u16string hint(reinterpret_cast<char16_t*>(config.hint_text));
+    if (!hint.empty()) {
+        std::cout << "Hint text: " << Common::UTF16ToUTF8(hint) << std::endl;
+    }
+    do {
+        std::cout << "Enter the text you will send to the application:" << std::endl;
+        std::getline(std::cin, input);
```
wwylele: What if the user wants to input multiple lines (while the game does accept it) ?

src/core/hle/applets/swkbd.cpp
```diff
@@ -21,6 +25,10 @@
 namespace HLE {
 namespace Applets {
 
+const static std::array<std::string, 1> swkbd_default_1_button = {"Ok"};
```
lioncash: const internally links by default, static isn't necessary.

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
     return RESULT_SUCCESS;
 }
 
+static bool ValidateFilters(const u32 filters, const std::string& input) {
+    bool valid = true;
+    bool local_filter = true;
+    if ((filters & SWKBDFILTER_DIGITS) == SWKBDFILTER_DIGITS) {
+        valid &= local_filter =
+            std::all_of(input.begin(), input.end(), [](const char c) { return !std::isdigit(c); });
```
lioncash: You can just use `auto` instead of `const char`

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+    } while (!ValidateInput(config, input));
 
-    // TODO(Subv): Remove this hardcoded text
-    std::u16string text = Common::UTF8ToUTF16("Citra");
-    memcpy(text_memory->GetPointer(), text.c_str(), text.length() * sizeof(char16_t));
+    std::string option_text;
+    // convert all of the button texts into something we can output
+    // num_buttons is in the range of 0-2 so use <= instead of <
+    u32 num_buttons = static_cast<u32>(config.num_buttons_m1);
+    for (u32 i = 0; i <= num_buttons; ++i) {
```
lioncash: This loop should probably be its own function.

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+    std::cout << "SOFTWARE KEYBOARD" << std::endl;
+    // Display hint text
+    std::u16string hint(reinterpret_cast<char16_t*>(config.hint_text));
+    if (!hint.empty()) {
+        std::cout << "Hint text: " << Common::UTF16ToUTF8(hint) << std::endl;
+    }
+    do {
+        std::cout << "Enter the text you will send to the application:" << std::endl;
+        std::getline(std::cin, input);
+    } while (!ValidateInput(config, input));
```
lioncash: This do-while can likely be its own function.

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+            } else {
+                final_text = swkbd_default_3_button[i];
+            }
+        } else {
+            final_text = Common::UTF16ToUTF8(custom_button_text);
+        }
+        option_text += Common::StringFromFormat("\t(%u) %s\t", i, final_text);
+    }
+    std::string option;
+    do {
```
lioncash: This loop can likely be its own function

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+        std::cout << "\nPlease type the number of the button you will press: \n"
+                  << option_text << std::endl;
+        std::getline(std::cin, option);
+    } while (!ValidateButton(static_cast<u32>(config.num_buttons_m1), option));
+
+    s32 button = std::stol(option);
+    switch (config.num_buttons_m1) {
```
lioncash: This switch can likely be its own function.

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+    bool valid = false;
+    try {
+        u32 num = std::stoul(input);
+        valid = num <= num_buttons;
+        if (!valid) {
+            std::cout << Common::StringFromFormat("Please choose a number between 0 and %u",
+                                                  num_buttons)
+                      << std::endl;
+        }
+    } catch (const std::invalid_argument& e) {
```
lioncash: You can just omit the `e` instead of `(void)` casting.

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+        (void)e;
+        std::cout << "Unable to parse input as a number." << std::endl;
+    } catch (const std::out_of_range& e) {
```
lioncash: ditto

src/core/hle/applets/swkbd.cpp
```diff
@@ -2,7 +2,11 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
```
lioncash: `<array>` also needs to be included.

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+    bool local_filter = true;
+    if ((filters & SWKBDFILTER_DIGITS) == SWKBDFILTER_DIGITS) {
+        valid &= local_filter =
+            std::all_of(input.begin(), input.end(), [](const char c) { return !std::isdigit(c); });
+        if (!local_filter) {
+            std::cout << "Input must not contain any digits" << std::endl;
+        }
+    }
+    if ((filters & SWKBDFILTER_AT) == SWKBDFILTER_AT) {
+        valid &= local_filter = input.find("@") == std::string::npos;
```
lioncash: `@` can be a character literal.

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+        if (!local_filter) {
+            std::cout << "Input must not contain the @ symbol" << std::endl;
+        }
+    }
+    if ((filters & SWKBDFILTER_PERCENT) == SWKBDFILTER_PERCENT) {
+        valid &= local_filter = input.find("%") == std::string::npos;
```
lioncash: `%` can be a character literal.

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+        if (!local_filter) {
+            std::cout << "Input must not contain the % symbol" << std::endl;
+        }
+    }
+    if ((filters & SWKBDFILTER_BACKSLASH) == SWKBDFILTER_BACKSLASH) {
+        valid &= local_filter = input.find("\\") == std::string::npos;
```
lioncash: `\\` can be a character literal.

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+                                                  config.max_text_length)
+                      << std::endl;
+        }
+        break;
+    case SwkbdValidInput::NOTEMPTY_NOTBLANK:
+    case SwkbdValidInput::NOTBLANK:
+        valid =
+            std::any_of(input.begin(), input.end(), [](const char c) { return !std::isspace(c); });
+        if (!valid) {
+            std::cout << "Input must not be blank." << std::endl;
```
lioncash: This could probably be clarified better, especially considering `std::isspace` is being used here, and not [std::isblank](http://en.cppreference.com/w/cpp/string/byte/isblank)

wwylele:
What's the status of this PR?

jroweboy:
@wwylele this is blocked on the logging rewrite, since that will probably include removing the console, stdin won't be a viable option to receive text input. I recall someone (i think it was @bunnei ?) was working on a simple gui package to draw on the screen, which would be what this should be based off in the future.

bunnei:
@jroweboy Correct, no idea when I'll get around to finishing it though, I haven't to much time to code lately - and there is still a lot of work left to do.

If there is a lot of demand for this feature, as a potential stop-gap solution, we could expose a static text input field in the config that always gets entered on a swkbd prompt (e.g. instead of the "citra" text that usually gets used for user names and such)

JayFoxRox:
I didn't follow the technical implementation of this, but why don't we just pause emulation and forward the event to the GUI? (which could be a popup-qt dialog ["Enter text here:"], stdin or just a function which returns such a config key)

jroweboy:
This is a really old PR that has some major outstanding issues.

1. Relies heavily on stdin which is planned to be removed
2. Hacks around lack of proper sleep support

A future PR to implement the software keyboard will need some form of GUI callback, or potentially draw to the screen. But most importantly, they will need to properly add sleep support to citra. As such, I don't see myself finishing the PR in its current incarnation, and those two things will involve a major rewrite, so I'll just close this for now. If I rewrite this someday, I'll just open a new PR.

src/core/hle/applets/swkbd.cpp
jroweboy: neobrain on irc: the condition isn't even correct

jroweboy: neobrain on irc: wrapping is_digit is entirely unnecessary in that context (because instead of using std::all_of with a negated function object you can just use std::none_of), but then I noticed the ridiculous "x &= y = " chaining, and then I finally noticed that the whole thing doesn't even do what it's supposed to do

src/core/hle/applets/swkbd.cpp
```diff
@@ -74,18 +82,192 @@ ResultCode SoftwareKeyboard::StartImpl(Service::APT::AppletStartupParameter cons
...
+                final_text = swkbd_default_1_button[i];
+            } else if (num_buttons == 1) {
+                final_text = swkbd_default_2_button[i];
+            } else {
+                final_text = swkbd_default_3_button[i];
+            }
+        } else {
+            final_text = Common::UTF16ToUTF8(custom_button_text);
+        }
+        option_text += Common::StringFromFormat("\t(%u) %s\t", i, final_text);
```
valentinvanelslande: Corrupt text with this, use:
`option_text += "\t(" + std::to_string(i) + ") " + final_text + "\t";`

PR #2335, opened 2016-12-16
Misc. Shader refactors

yuriks:
See individual commits


yuriks:
Updated with a bunch of larger refactors. (I'll stop force-pushing now.)

PR #2343, opened 2016-12-17
Core: Top-level consolidate & misc cleanup

bunnei:
This consolidates some functionality/state in Core, removing some legacy files that don't really need to exist/make sense, as well as some other misc changes/cleanups related to some other things I'm working on (sorry for the hodgepodge of changes, but it made sense at the time...) Also removed some cruft that I don't think does anything. This should probably be tested a bit in BE just to be sure.


src/core/hle/service/fs/archive.h
```diff
@@ -16,10 +16,11 @@ class DirectoryBackend;
 class FileBackend;
 }
 
+/// TODO(Subv): Confirm length of these strings
```
yuriks: I think you can just remove this, lengths are correct. (And if they weren't it wouldn't really matter.)

src/core/system.h
```diff
@@ -4,18 +4,80 @@
 
 #pragma once
 
+#include <string>
+
+#include "core/loader/loader.h"
```
lioncash: This can be a forward declaration.

PR #2345, opened 2016-12-17
Thread: remove the thread from the thread list when exiting

wwylele:
Currently we leave all dead thread in the `thread_list` and never remove them until exiting the game. This has a significant impact on game that loves to create small threads frequently (such as Pokemon Sun, which creates thousands of threads in 10 minutes). In this case, the `thread_list` grows up quickly and slows down all functions that iterating over the list, as users reported "games slow down after an hour".

I did a profiling on Pokemon running for 40 minutes, with the character standing there without any input. The frame rate drops from 25 FPS at the beginning to 20 FPS at the end. The profiling shows a time difference spending on a function between the beginning and the end:

The first 5 minutes:
fast

The last 5 minutes:
slow

The profiling results after the change shows no difference in the function between the first and the last 5 minutes.

To avoid use-after-free, I turn `current_thread` into a `SharedPtr`, ~~as well as the return type of `GetCurrentThread`~~. But I am not confident with these changes. Please help me check if the memory management is ok.

### DIFF
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -46,7 +46,7 @@ static std::vector<SharedPtr<Thread>> thread_list;
 // Lists only ready thread ids.
 static Common::ThreadQueueList<Thread*, THREADPRIO_LOWEST + 1> ready_queue;
 
-static Thread* current_thread;
+static SharedPtr<Thread> current_thread;
 
 // The first available thread id at startup
 static u32 next_thread_id;
@@ -63,7 +63,7 @@ Thread::Thread() {}
 Thread::~Thread() {}
 
 Thread* GetCurrentThread() {
-    return current_thread;
+    return current_thread.get();
 }
 
 /**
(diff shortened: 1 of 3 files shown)


wwylele:
Temporary closed. I am checking the memory management.

src/core/hle/kernel/thread.cpp
```diff
@@ -120,6 +120,9 @@ void Thread::Stop() {
     u32 tls_slot =
         ((tls_address - Memory::TLS_AREA_VADDR) % Memory::PAGE_SIZE) / Memory::TLS_ENTRY_SIZE;
     Kernel::g_current_process->tls_slots[tls_page].reset(tls_slot);
+
+    // Remove the thread from the thread list
+    thread_list.erase(std::remove(thread_list.begin(), thread_list.end(), this), thread_list.end());
```
TTFH: This line makes emulator to crash when click in "Emulation => Close" in TLZ A Link Between Worlds.

Segmentation fault in ReleaseThreadMutexes()

wwylele:
@TTFH I pushed a fix. Does it fix the crash?

TTFH:
@wwylele Yes, It does.

PR #2346, opened 2016-12-17
More shader refactoring

### LINKED ISSUE #1051: Race condition in Vertex Shader JIT

yuriks:
Act 2: In which I give @JayFoxRox a heart attack.

I strongly recommend reviewing each commit individually. There's a lot of stuff here, but the common theme is removing direct dependencies on global state.

@JayFoxRox: Most of this seems to me like it should make your job easier, since it makes it trivial to have several independent shader units. If you think it's too onerous to rebase on top of this, I can split some of these changes out to a separate PR I can rebase on top of your stuff whenever you eventually submit that. (Though I'd of course prefer to merge my cleanups now...)

(As a bonus, this also fixes #1051)


JayFoxRox:
It will take some time to review this. I'll try to rebase my GS branch onto it to see if there are any design issues.

I believe this fixes #1051 by moving the check for the enabled JIT to the GetEngine() which will remain good even if the setting is changed shortly after.

src/video_core/shader/shader_jit_x64_compiler.cpp
```diff
@@ -833,7 +824,11 @@ void JitShader::FindReturnOffsets() {
     std::sort(return_offsets.begin(), return_offsets.end());
 }
 
-void JitShader::Compile() {
+void JitShader::Compile(const std::array<u32, 1024>* program_code_,
+                        const std::array<u32, 1024>* swizzle_data_) {
```
JayFoxRox: I'm not sure wether VS and GS uniforms are different (per shader type), but if they are, this should just get passed the ShaderSetup which only consists of Uniforms, Program Code and Swizzle Data. (so if uniforms are also different this would essentially just unpack and repack a ShaderSetup).
Even if the uniforms are different per shader unit this could be passed a ShaderSetup as we'd move the uniforms to the unit state instead. (so program_code and swizzle_data would be everything the ShaderSetup consists of).

src/video_core/shader/shader.cpp
```diff
@@ -76,84 +70,42 @@ OutputVertex OutputRegisters::ToVertex(const Regs::ShaderConfig& config) const {
...
-        } else {
-            auto shader = std::make_unique<JitShader>();
-            shader->Compile();
-            jit_shader = shader.get();
-            shader_map[cache_key] = std::move(shader);
-        }
-    }
-#endif // ARCHITECTURE_x86_64
+    conditional_code[0] = false;
+    conditional_code[1] = false;
```
JayFoxRox: Why is this part of vertex loading?

src/video_core/shader/shader.cpp
```diff
@@ -76,84 +70,42 @@ OutputVertex OutputRegisters::ToVertex(const Regs::ShaderConfig& config) const {
...
-    state.conditional_code[0] = false;
-    state.conditional_code[1] = false;
+#ifdef ARCHITECTURE_x86_64
+static std::unique_ptr<JitX64Engine> jit_engine;
+#endif // ARCHITECTURE_x86_64
+static InterpreterEngine interpreter_engine;
 
+ShaderEngine* GetEngine() {
 #ifdef ARCHITECTURE_x86_64
+    // TODO(yuriks): Re-initialize on each change rather than being persistent
```
JayFoxRox: We need at least 2 engines.

When GS is enabled, the VS will be run a couple of times until a buffer is filled up.
Once the buffer has filled up to a certain point the GS will process the generated data.
After that the VS will continue its thing.

As we currently only have 1 engine which is modified by SetupBatch its not possible to switch code back and forth (between VS / GS)

JayFoxRox:
I've rebased my GS branch https://github.com/JayFoxRox/citra/pull/49

Problems with design:

1. We need 2 engines (/ cached shader code) at the same time
2. Input vertex loading for GS is done from a buffer, not InputVertex, so the LoadInputVertex function is kind of useless here (we could overload it maybe?)
3. Initialization of the conditional vars must be done manually currently as GS doesn't use LoadInputVertex

Aside from that it's fine I think. I only tested 2 games though.
I also did not rebase my shader debugger yet, but I guess it's not that important.

src/video_core/shader/shader_jit_x64_compiler.h
```diff
@@ -0,0 +1,125 @@
...
+    std::vector<unsigned> return_offsets;
+
+    unsigned program_counter = 0; ///< Offset of the next instruction to decode
+    bool looping = false;         ///< True if compiling a loop, used to check for nested loops
+
+    using CompiledShader = void(const void* setup, void* state, const u8* start_addr);
+    CompiledShader* program = nullptr;
+};
+
+} // Shader
```
lioncash: Isn't the general convention to do `// namespace Shader` as opposed to just the name of the namespace?

yuriks: I guess? I didn't change this line.

src/video_core/shader/shader_jit_x64_compiler.cpp
yuriks: Different in the sense that they have different structure, or different values? They definitely have different values (otherwise it'd be impossible to decouple shader programs from each other because, for example, all shader constants are actually uniforms.) The compiler doesn't need to should know about uniform values, so I only pass to it the things it's actually compiling (the program/swizzles.)

src/video_core/shader/shader.cpp
yuriks: I envisioned engines as being singletons, because that way they can share cache. I'll try to think of an alternative way of allowing multiple concurrent setups.

src/video_core/shader/shader_jit_x64_compiler.h
lioncash: Ah OK, I misread the diff then, my bad.

yuriks:
@JayFoxRox 

> We need 2 engines (/ cached shader code) at the same time

I'm trying to avoid that. Will think about how to solve this.

> Input vertex loading for GS is done from a buffer, not InputVertex, so the LoadInputVertex function is kind of useless here (we could overload it maybe?)

LoadInputVertex doesn't apply then, just make another function to transfer VSH outputs to GSH inputs. We'll probably end up with a bit of code duplication. For example, I think the way we're handling the VSH outputs right now merges two concepts that should be separate (shader engine appending output regs into a buffer/queue according to the outmask, and the geometry unit reading them from that buffer), but that can be refactored after we have the two shader types in place.

> Initialization of the conditional vars must be done manually currently as GS doesn't use LoadInputVertex

Fixed that.

yuriks:
@JayFoxRox Pushed a commit which should address the issue with using 2 different shaders at the same time.

src/video_core/shader/shader_jit_x64_compiler.h
yuriks: This was part of a file rename, and then I created another file with the same name on a later commit. So it doesn't get detected by the full diff, but if you look at each individual commit then it doesn't show up as a change.

yuriks:
Assuming the last changes address @JayFoxRox's concerns, I'll cut this PR off here. I have more refactors I want to make but I want to avoid the PR growing too large.

JayFoxRox:
LGTM

yuriks:
FYI I'm waiting on @JayFoxRox to do another review on this one (and possibly someone else if they feel like they know the code enough) because that last one had been when he'd been up for more than 24 hours or something... After that I'll rebase and merge.

yuriks:
Rebased. Since @JayFoxRox seems to still be MIA, I'd love if someone else could review this. Otherwise I'll just merge it in say, a week.

src/video_core/shader/shader_jit_x64_compiler.cpp
```diff
@@ -833,7 +824,11 @@ void JitShader::FindReturnOffsets() {
...
+    program_code = program_code_;
```
Subv: What are these members for? You assign them here but set them to nullptr at the end of the function

src/video_core/shader/shader.h
```diff
@@ -177,13 +183,14 @@ class ShaderEngine {
      * Performs any shader unit setup that only needs to happen once per shader (as opposed to once
      * per vertex, which would happen within the `Run` function).
      */
-    virtual void SetupBatch(const ShaderSetup* setup) = 0;
+    virtual void SetupBatch(ShaderSetup& setup) = 0;
 
     /**
      * Runs the currently setup shader
      * @param state Shader unit state, must be setup per shader and per shader unit
```
Subv: This comment is outdated now

src/video_core/shader/shader_jit_x64_compiler.cpp
JayFoxRox: They are used in the compilation functions (one per instruction) and are not passed by parameter. So in order to make them available they are temporarily stored in members during compilation.

As we don't need them afterwards anymore we can set them to `nullptr` afterwards.

yuriks: It's as he says. Not the best design, but unfortunately doing otherwise would require passing those parameters around on every function call.

src/video_core/shader/shader.h
yuriks: Updated.

src/video_core/shader/shader.cpp
jroweboy: Just following up on this conversation (since its the last call for reviews) @yuriks any thoughts on how to have multiple engines yet?

yuriks: Yes, this has been addressed by "VideoCore/Shader: Move per-batch ShaderEngine state into ShaderSetup"

PR #2350, opened 2016-12-18
Updated main.cpp

CaptV0rt3x:
*Fixed issue on render_window / game_list focus.
previously if we started a game and then immediately pressed down+enter, the game_list would load the second game from list, due to focus issue. By setting focus on render_window, issue is fixed.

P.s ignore the second commit. It was a accidental mistake.


CaptV0rt3x:
Creating a new pr. 2 commits were a mistake.

PR #2351, opened 2016-12-18
Fixed game_list focus issue.

CaptV0rt3x:
*Fixed game_list focus issue.
previously, if we pressed down/up +enter after game started, the game list would load the previous/next game from list. setting focus on render_window fixes this issue.


wwylele:
Please rename the commit message so that it indicates what is the change exactly

PR #2355, opened 2016-12-19
[RFC] citra-qt: Print stack trace redux

MerryMage:
Original PR: #1874 

Things changed from original PR:

* Rewritten to use `setjmp`/`longjmp` so that Qt can be used.
* Windows-only for now.
* Requires a pdb file to be present for symbols to work on Windows.
* Design of crash dialog TBD, it's just a QMessageBox but can be anything.

Screenshot (real design tbd):


src/common/crash_handler.cpp
```diff
@@ -0,0 +1,161 @@
...
+static jmp_buf unhandled_exception_jmp_buf;
+static std::vector<std::string> unhandled_exception_stack_trace;
+
+static LONG WINAPI MyUnhandledExceptionFilter(_EXCEPTION_POINTERS*);
+static void GetStackTrace(CONTEXT& c);
+
+void CrashHandler(std::function<void()> try_,
+                  std::function<void(const Common::CrashInformation&)> catch_) {
+    unhandled_exception_called = false;
+    SetUnhandledExceptionFilter(MyUnhandledExceptionFilter);
```
MerryMage: Unset when done.

Subv:
I think instead of showing a dialog, we should dump the relevant information to a file, perhaps a Visual Studio dmp file.

FernandoS27:
Must agree with Subv. A crash report generated with date, version and additional info would be cooler.

Nice work, though.

chris062689:
A crash report would be much more useful attaching to the forums or discord. If it could include the git hash that would be useful too....

MerryMage:
@chris062689 Regarding the crash dialog: I was thinking of having a copy-pastable textbox. git hash and branch name are things that can be added quite easily.

Subv:
As a side note, a dmp file would also contain information about the registers / variables at crash time

MerryMage:
@Subv A minidump would be reasonable starting point, though I think we'd want slightly more information that just the minidump. Not sure how that would work with the gcc builds (but those aren't supported anyway).

yuriks:
Stack traces are pretty easy to do if it's just windows, but if we're going the minidump route then it's probably better to just use something off-the-shelf like Breakpad.

MerryMage:
Breakpad is certainly an option for a library to integrate.

Breakpad does provide the ability to send minidumps to a server and a server-side interface for parsing minidumps. It does not however provide a user-side UI, though this should be rather straightfoward to implement.

Can I consider this approach agreeable with everyone?

chris062689:
I'd prefer user-uploadable minidumps instead of automatically sending them to a server. We don't have the infrastructure setup for that and that may take time.

MerryMage:
Closed in favour of eventual implementation of alternative suggestion(s) above.

PR #2367, opened 2016-12-22
Lighting LUT Quickfix

JayFoxRox:
This fixes the broken lighting in Zelda: ALBW (As reported by Migue via Discord).

**Broken (master):**
Broken

**Fixed (this PR):**
Fixed

---

The error is caused by a wrapping of the texture in `LUT[4]` (`GL_TEXTURE_7`) where the red channel (so `texture(LUT[4], index)[0])` will be a ramp from 0.0 to 1.0 accross its width of 256 pixels.
Hence turning samples at texcoord 0.0 into the value 0.5 (average of 0.0 (left border) and 1.0 (right border)).

Note that this also removes the scaling with `255.0/256.0` (`FLOAT_255`) because I'm not sure what it was supposed to do (as offsetting from 0.0 wasn't done and scaling would have to be done using: *stupidme* check what yuriks wrote below for the proper formula).
All of the clamping has been moved into the texture object now which means we can simplify the shader slightly.

Aside from that, this PR removes some ugly float suffixes and implicit conversions in GLSL.
The whole shader_gen deserves a rewrite, but that is a seperate issue.

---

@degasus pointed out that glTexSubImage1D is probably a lot worse than buffer textures.
So in the future, we should be rewriting the whole LUT (lighting and fog) upload code.
This would also give us the option to adapt a better interpolation like the fog LUT already does (albeit being very slow / stupid in GLSL in the case of the fog LUT).
However, I consider that a seperate issue. So this quickfix will have to do for now.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_gen.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_gen.cpp
@@ -423,15 +423,12 @@ static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
             // LUT index is in the range of (0.0, 1.0)
             index = lighting.light[light_num].two_sided_diffuse ? "abs(" + index + ")"
                                                                 : "max(" + index + ", 0.f)";
-            return "(FLOAT_255 * clamp(" + index + ", 0.0, 1.0))";
         } else {
             // LUT index is in the range of (-1.0, 1.0)
-            index = "clamp(" + index + ", -1.0, 1.0)";
-            return "(FLOAT_255 * ((" + index + " < 0) ? " + index + " + 2.0 : " + index +
-                   ") / 2.0)";
+            index = "((" + index + " < 0) ? " + index + " + 2.0 : " + index + ") / 2.0";
         }
 
-        return std::string();
+        return "(OFFSET_256 + SCALE_256 * clamp(" + index + ", 0.0, 1.0))";
     };
 
     // Gets the lighting lookup table value given the specified sampler and index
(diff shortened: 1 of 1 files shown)


JayFoxRox:
I've tested this on a handful of games. I'm not aware of any regressions.

The texcoords are still bad (as they were originally because it did not respect GL texcoords) but the clamping is far better than before (faster and actually works).
This is a temporary measure so I don't care about the small texcoord issues. We'll have to rewrite the lighting LUT code anyway.

JayFoxRox:
I just came accross a gbatemp post by accident (thanks for not reporting it.. idiots) and this might break lighting in Majoras Mask: https://gbatemp.net/threads/citra-new-3ds-emulator.365154/page-310#post-6945375

I'll look into it soon.

yuriks:
I'm pretty sure the intend with FLOAT_255 was to remap it so that it wouldn't sample from the edges, however, the correct way of doing it would be `u' = u * ((W-1) / W) + (0.5 / W)`, we forgot the 0.5 texel offset before, which is probably why this didn't work properly.

With regards to the PR, I'm not sure if there's much point in doing this: Without remapping it properly the 0 and 1 values still won't be aligned properly in the texels, it regresses ZMM, and it won't be correct anyway. I'd just skip ahead to the full implementation with the deltas. (Or materialize a 257th entry in the texture so that the last element is correct.)

JayFoxRox:
Apparently there is a bug with older AMD legacy drivers where texture1D wrapmode does not work properly (that's an assumption, I couldn't find other reports about it but I've talked to the person who has the bug and they ran a July 2015 stable / Nov 2015 beta driver for Radeon HD 5870).

To avoid such issues and for more correct (still far from perfect!) emulation I chose what yuriks suggested (clamping in shader and offset texture by 0.5/w, re-introduce scaling).
Tested in Zelda: MM and Zelda: ALBW only.

Dragios:
I have discovered a regression in Pokémon Main Series game #2551

Issue #1886, opened 2016-06-05
Events with wrong type


wwylele:
A example confirmed by hwtest is [CAM completion event](https://github.com/citra-emu/citra/blob/master/src/core/hle/service/cam/cam.cpp#L296), which should be `Sticky` instead of  `OneShot`. 
There are also other events that are probably with wrong type (such as Y2R completion event, which is functioning similarly to the CAM one), but I haven't test yet.

mailwl:
if Y2r sticky, then all FIFA games shows intro videos

Subv:
There are 19 HLE events in the codebase, here's a list of what's verified and what's not, to be modified by contributors as they are checked.

***Unverified events***

- [ ] HID DebugPad event
- [ ] HID Gyroscope event
- [ ] HID Accelerometer event
- [ ] HID Pad or Touch 1 event
- [ ] HID Pad or Touch 2 event
- [ ] DSP Semaphore event
- [ ] IR Connection status event
- [ ] IR Handle event
- [ ] CECD Change state event
- [ ] CECD CEC info event
- [ ] CAM VSync interrupt event
- [ ] CAM Interrupt error event
- [ ] CAM Completion event CAM 1
- [ ] CAM Completion event CAM 2
- [ ] APT Notification event
- [ ] APT Parameter event
- [ ] SRV event
- [ ] NWM event


***Verified events: OneShot***
- [x] Y2R Completion event

***Verified events: Sticky***

***Verified events: Pulse***

mailwl:
long time ago i tryed to pulse cam-completion events, here: https://github.com/mailwl/citra/blob/threads/src/core/hle/service/cam/cam.cpp#L79
no success with camera-powered apps

PR #2368, opened 2016-12-22
CAM: build the service framework with a dummy implementation

### LINKED ISSUE #1832: Pokemon Super Mystery Dungeon: freeze at intro (thread locking)

wwylele:
What is included in this PR:
 - CAM service cleanup, and abstracted all essential configuration functions into `Camera::CameraInterface`. Implemented asynchronous image receiving from the interface (fixes #1832).
 - Made camera configurable from INI file. 
 - Added three implementation of `Camera::Interface`:
   - `BlankCamera` (default) a dummy camera that always returns black image
   - ~~`V4L2Camera` (Linux only) a V4L2 based camera. The original code is taken from #1947 (credit to @linkmauve). I am not so familiar with linux, tho, so I can't tell if this part of code is good enough.~~
   - ~~`StillImageCamera` (Qt frontend only) a camera that always returns the same image the user selected.~~ (removed as Lioncash suggested. will be in follow-up PRs)

What is not included in this PR and can be done in the future:
 - Non-essential configuration functions (white balance, exposure, etc.) implementation.
 - Configuration tab for camera in Qt frontend
 - ~~Other real camera implementation to support all three major platforms. A candidate is Qt's `QCamera`. This will add a new dependency QtMultiMedia, though.~~ real camera implementation

This main part of this PR builds the camera service framework. The framework is based on my test and theory. The test program can be found [here](https://github.com/wwylele/ctrhwtest/tree/master/cam_service_test).

I put two actual implementations in my repo to show how `CameraInterface` works. I will PR them if this PR get merged:

 - [V4L2 camera](https://github.com/wwylele/citra/tree/v4l2-camera). To use it you need to assign `V4L2` to `camera_*_name` (the `out_right` one is the most used one) in the ini file, and to assign the device ID (`0` for example) to `camera_*_config`
 - [Still image camera](https://github.com/wwylele/citra/tree/image-camera). To use it, you need to assign `image` to `camera_*_name` in the ini file, and to assign the image file name to `camera_*_config`

Here are some points to note in my code:

 - I made CAM service function return SUCCESS even if there is an error in the actual implementation. I made all functions of `Camera::Interface` returns void. I did this because "returning SUCCESS" is an expected result on real 3DS, and games may not know how to deal with device errors.

 - (Will add here if I come up with more)

----

Tested games/apps:
 - devKitPro camera example: works fine.
 - System camera app: not showing anything. This app calls some undocumented functions.
 - Mii Maker (QR reader): capturing works fine. QR reading fails due to unimplemented crypto functions.
 - Pokemon Super Mystery Dungeon (title screen without a save): works fine.
 - Pokemon Sun / Moon (QR reader): works fine.


lioncash:
@wwylele It might be more beneficial to split these up into three PRs, since they're logically three separate changes being done.

src/core/hle/service/cam/cam.h
```diff
@@ -183,36 +178,82 @@ struct PackageParameterCameraSelect {
...
+ *      1: u8 port select
  *  Outputs:
  *      0: 0x00010040
  *      1: ResultCode
  */
 void StartCapture(Service::Interface* self);
 
 /**
- * Unknown
+ * Stops the select ports capturing.
```
lioncash: `Stops capturing from a camera identified by a selected port`

src/core/hle/service/cam/cam.h
```diff
@@ -183,36 +178,82 @@ struct PackageParameterCameraSelect {
...
  *  Inputs:
  *      0: 0x00020040
- *      1: u8 Camera port (`Port` enum)
+ *      1: u8 port select
```
lioncash: `selected port?`

src/core/hle/service/cam/cam.h
```diff
@@ -183,36 +178,82 @@ struct PackageParameterCameraSelect {
...
+    Effect effect;
+    Resolution resolution;
+};
+
+static_assert(sizeof(PackageParameterWithContextDetail) == 16,
+              "PackageParameterWithContextDetail structure size is wrong");
 
 /**
- * Unknown
+ * Start the select ports capturing.
```
lioncash: `Starts capturing from a camera identified by a selected port.`

src/core/hle/service/cam/cam.h
```diff
@@ -183,36 +178,82 @@ struct PackageParameterCameraSelect {
...
  *  Inputs:
  *      0: 0x00010040
- *      1: u8 Camera port (`Port` enum)
+ *      1: u8 port select
```
lioncash: `selected port`?

src/common/camera/interface.h
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Sets the output format of the frame.
+     * @param format Output format of the frame
+     */
+    virtual void SetFormat(Service::CAM::OutputFormat format) = 0;
+
+    /**
+     * Receives a frame from the camera.
+     * This function should be only called between a StartCapture call and a StopCapture call.
+     * @returns a std::vector of u16, with one pixel per each, and with totally (width * height)
```
lioncash: `A std::vector<u16> containing pixels. The total size of the vector is width * height where width and height are set by a call to SetResolution.`

src/common/camera/interface.h
```diff
@@ -0,0 +1,61 @@
...
+     * For the meaning of the parameters, please refer to Service::CAM::Resolution. Note that the
+     * actual camera implementation doesn't need to respect all the parameters. However, the width
+     * and the height parameters must be respected and be used to determine the size of output
+     * frames.
+     * @param resolution The resolution parameters to set
+     */
+    virtual void SetResolution(const Service::CAM::Resolution& resolution) = 0;
+
+    /**
+     * Sets the vertical and horizontal flip applying on the frame.
```
lioncash: If this only applies to the next frame received, this would be better phrased as `Sets the vertical and horizontal flip applies to the next received frame.` If it applies to all frames received until a SetFlip with different parameters is provided, then this would be better phrased as:

`Sets the vertical and horizontal flip applied to all frames received after this function is called.`

This applis similarly to the other setters below.

src/core/hle/service/cam/cam.h
```diff
@@ -183,36 +178,82 @@ struct PackageParameterCameraSelect {
...
+ *      1: ResultCode
+ *      2: 0 if not capturing, 1 if capturing
+ */
+void IsBusy(Service::Interface* self);
+
+/**
+ * Clears the buffer of select ports.
+ *  Inputs:
+ *      0: 0x00040040
+ *      1: u8 port select
```
lioncash: `selected port`?

src/common/camera/interface.h
```diff
@@ -0,0 +1,61 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <vector>
+#include "core/hle/service/cam/cam.h"
```
wwylele: I saw there is already [an example of dependency on core](https://github.com/citra-emu/citra/blob/master/src/common/emu_window.h#L12) so I thought it's ok. I don't use any functions from the core, but just reuse some `enum`s so that I don't need to create my own `enum` and do redundant translation. I also thought of putting these into the core, but something like `V4L2Camera` that must be in common will still include this from the core, which is not better than the current.

lioncash: > I saw there is already an example of dependency on core 

This is something that should be fixed. [Circular dependencies](https://en.wikipedia.org/wiki/Circular_dependency) are not a good thing to have.

> V4L2Camera that must be in common 

Why would `V4L2Camera` be required to be in common? This is starting to sound like a design/organizational flaw if it's a requirement.

wwylele: Oh sorry, I was probably thinking of something wrong. Yeah `V4L2Camera` doesn't need to be common. So I am going to move the whole folder `common/camera` to `core/hle/service/cam`. Is this good?

wwylele:
Addressed comments above. I moved the interface stuff to `core/frontend/camera`, the same place as the one proposed in #2369 .

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+}
+
+// verifies a "*_select" parameter represents an unique ID
+static bool VerifyBitSetUnique(u32 bit_set, unsigned count) {
+    return bit_set < (1 << count) && bit_set != 0 && !(bit_set & (bit_set - 1));
+}
+
+static void CompletionEventCallBack(u64 port_id, int) {
+    auto& port = ports[port_id];
+    auto& camera = cameras[port.camera_id];
```
lioncash: This can be a const reference as far as I can tell.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+    auto buffer = port.capture_result.get();
+
+    if (port.is_trimming) {
+        u32 trim_size = (port.x1 - port.x0) * (port.y1 - port.y0) * 2;
+        if (port.dest_size != trim_size) {
+            LOG_ERROR(Service_CAM, "The destination size (%d) doesn't match the source (%d)!",
```
lioncash: `%u` is the correct specifier for an unsigned value.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+                      port.dest_size, trim_size);
+        }
+
+        VAddr dest_ptr = port.dest;
+        int dest_size = port.dest_size;
```
lioncash: You may want to consider making variables not intended to change `const` here.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+        int trim_width = port.x1 - port.x0;
+        int line_bytes = trim_width * sizeof(u16);
```
lioncash: `sizeof` returns a `size_t`, so this would be truncating without a cast.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+        int trim_height = port.y1 - port.y0;
+        int original_width = camera.contexts[camera.current_context].resolution.width;
+        int src_offset = port.y0 * original_width + port.x0;
+        const u16* src_ptr = buffer.data() + src_offset;
+        int src_size = (buffer.size() - src_offset) * sizeof(u16);
```
lioncash: `src_size` should be a size_t.

src/core/hle/service/cam/cam.cpp
lioncash: `dest_size` should also be either a `u32` (port.dest_size's type) or a `size_t`.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
 #include "core/hle/service/cam/cam_s.h"
 #include "core/hle/service/cam/cam_u.h"
 #include "core/hle/service/service.h"
+#include "core/settings.h"
 
 namespace Service {
 namespace CAM {
 
-static const u32 TRANSFER_BYTES = 5 * 1024;
+struct ContextConfig {
```
lioncash: You can place all of the file-local stuff into an anonymous namespace, that way you don't need to type `static` over and over.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+        std::async(std::launch::async, &Camera::CameraInterface::ReceiveFrame, camera.impl.get());
+
+    // schedules an completion event according to the frame rate. The event will block on the
+    // capture task if it is not finished in the expected time.
+    CoreTiming::ScheduleEvent(
+        msToCycles(LATENCY_BY_FRAME_RATE[static_cast<int>(camera.frame_rate)]),
+        completion_event_callback, port_id);
+}
+
+// Force to cancel any ongoing receiving process on the specified port. This is used by functions
```
lioncash:
```
Cancels any ongoing receiving processes at the specified port. This is used by functions that stop capturing.
```

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+    auto& port = ports[port_id];
+    port.is_receiving = true;
+
+    // launches a capture task asynchronously
+    auto& camera = cameras[port.camera_id];
+    port.capture_result =
+        std::async(std::launch::async, &Camera::CameraInterface::ReceiveFrame, camera.impl.get());
+
+    // schedules an completion event according to the frame rate. The event will block on the
+    // capture task if it is not finished in the expected time.
```
lioncash: `within the expected time`

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+static void StartReceiving(int port_id) {
+    auto& port = ports[port_id];
+    port.is_receiving = true;
+
+    // launches a capture task asynchronously
+    auto& camera = cameras[port.camera_id];
+    port.capture_result =
+        std::async(std::launch::async, &Camera::CameraInterface::ReceiveFrame, camera.impl.get());
+
+    // schedules an completion event according to the frame rate. The event will block on the
```
lioncash: `a completion event`

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+}
+
+// Starts a receiving process on the specified port. This can only be called when is_busy = true and
+// is_receiving = false.
+static void StartReceiving(int port_id) {
+    auto& port = ports[port_id];
+    port.is_receiving = true;
+
+    // launches a capture task asynchronously
+    auto& camera = cameras[port.camera_id];
```
lioncash: I believe camera can be `const` here.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+                break;
+            }
+            Memory::WriteBlock(dest_ptr, src_ptr, copy_length);
+            dest_ptr += copy_length;
+            dest_size -= copy_length;
+            src_ptr += original_width;
+            src_size -= original_width * sizeof(u16);
+        }
+    } else {
+        u32 buffer_size = buffer.size() * sizeof(u16);
```
lioncash: This should be a `size_t`

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+        if (port.dest_size != buffer_size) {
+            LOG_ERROR(Service_CAM, "The destination size (%d) doesn't match the source (%d)!",
```
lioncash: `%u` and `%zu` (for size_t), respectively.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
-    LOG_WARNING(Service_CAM, "(STUBBED) called, addr=0x%X, port=%d, image_size=%d, trans_unit=%d",
-                dest, port, image_size, trans_unit);
+    LOG_DEBUG(Service_CAM, "called, addr=0x%X, port_select=%d, image_size=%d, trans_unit=%d", dest,
+              port_select, image_size, trans_unit);
+}
+
+void IsFinishedReceiving(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 port_select = cmd_buff[1] & 0xFF;
```
lioncash: This should probably be a u32 otherwise this is truncating from u32 to u8 without a cast.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
...
+
+    if (VerifyBitSetUnique(port_select, 2)) {
+        int port = BitSetToIDs(port_select).back();
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+        cmd_buff[2] = (ports[port].is_receiving || ports[port].is_pending_receiving) ? 0 : 1;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid port_select=%d", port_select);
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
     u16 width = cmd_buff[1] & 0xFFFF;
     u16 height = cmd_buff[2] & 0xFFFF;
 
+    // Note: the result of the algorithm below are hwtested with width < 640 and with height < 480
+    if (width * height * 2 % 256 != 0) {
```
lioncash: You may want to expand your comment indicate what some of these magic constants are (or otherwise make them named constants).

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, width=%d, height=%d, lines = %d", width, height,
-                cmd_buff[2]);
+    LOG_DEBUG(Service_CAM, "called, width=%d, height=%d", width, height);
+}
+
+void SetTransferBytes(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 port_select = cmd_buff[1] & 0xFF;
```
lioncash: These should all probably be `const u32` otherwise these all truncate to smaller values without a cast.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+    u16 width = cmd_buff[3] & 0xFFFF;
+    u16 height = cmd_buff[4] & 0xFFFF;
+
+    if (VerifyBitSet(port_select, 2)) {
+        for (int i : BitSetToIDs(port_select)) {
+            ports[i].transfer_bytes = transfer_bytes;
+        }
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid port_select=%d", port_select);
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+        cmd_buff[1] = ERROR_INVALID_ENUM_VALUE.raw;
+    }
+
+    cmd_buff[0] = IPC::MakeHeader(0xB, 1, 0);
+
+    LOG_WARNING(Service_CAM, "(STUBBED)called, port_select=%d, bytes=%d, width=%d, height=%d",
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+}
+
+void GetMaxBytes(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u16 width = cmd_buff[1] & 0xFFFF;
+    u16 height = cmd_buff[2] & 0xFFFF;
+
+    // Note: the result of the algorithm below are hwtested with width < 640 and with height < 480
+    if (width * height * 2 % 256 != 0) {
```
lioncash: ditto about magic constants

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
-    cmd_buff[2] = TRANSFER_BYTES;
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, port=%d", port);
+    LOG_WARNING(Service_CAM, "(STUBBED)called, port_select=%d", port_select);
+}
+
+void GetMaxBytes(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u16 width = cmd_buff[1] & 0xFFFF;
```
lioncash: These should probably be `const u32`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+        while (width * height * 2 % bytes != 0) {
+            bytes -= 256;
+        }
+
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+        cmd_buff[2] = bytes;
+    }
+    cmd_buff[0] = IPC::MakeHeader(0xD, 2, 0);
+
+    LOG_DEBUG(Service_CAM, "called, width=%d, height=%d", width, height);
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
-    cmd_buff[1] = RESULT_SUCCESS.raw;
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, port=%d, trim=%d", port, trim);
+    LOG_DEBUG(Service_CAM, "called, port_select=%d, trim=%d", port_select, trim);
+}
+
+void IsTrimming(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 port_select = cmd_buff[1] & 0xFF;
```
lioncash: This should be a const u32

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+
+    if (VerifyBitSetUnique(port_select, 2)) {
+        int port = BitSetToIDs(port_select).back();
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+        cmd_buff[2] = ports[port].is_trimming;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid port_select=%d", port_select);
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+        cmd_buff[1] = ERROR_INVALID_ENUM_VALUE.raw;
+    }
+
+    cmd_buff[0] = IPC::MakeHeader(0xF, 2, 0);
+
+    LOG_DEBUG(Service_CAM, "called, port_select=%d", port_select);
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+}
+
+void SetTrimmingParams(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 port_select = cmd_buff[1] & 0xFF;
```
lioncash: these should likely all be `const u32` or cast where appropriate.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+    if (VerifyBitSet(port_select, 2)) {
+        for (int i : BitSetToIDs(port_select)) {
+            ports[i].x0 = x0;
+            ports[i].y0 = y0;
+            ports[i].x1 = x1;
+            ports[i].y1 = y1;
+        }
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid port_select=%d", port_select);
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+        cmd_buff[1] = ERROR_INVALID_ENUM_VALUE.raw;
+    }
+
+    cmd_buff[0] = IPC::MakeHeader(0x10, 1, 0);
+
+    LOG_DEBUG(Service_CAM, "called, port_select=%d, x0=%d, y0=%d, x1=%d, y1=%d", port_select, x0,
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+              y0, x1, y1);
+}
+
+void GetTrimmingParams(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 port_select = cmd_buff[1] & 0xFF;
```
lioncash: This should probably be a const u32.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+
+    if (VerifyBitSetUnique(port_select, 2)) {
+        int port = BitSetToIDs(port_select).back();
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+        cmd_buff[2] = ports[port].x0;
+        cmd_buff[3] = ports[port].y0;
+        cmd_buff[4] = ports[port].x1;
+        cmd_buff[5] = ports[port].y1;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid port_select=%d", port_select);
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+        cmd_buff[1] = ERROR_INVALID_ENUM_VALUE.raw;
+    }
+
+    cmd_buff[0] = IPC::MakeHeader(0x11, 5, 0);
+
+    LOG_DEBUG(Service_CAM, "called, port_select=%d", port_select);
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
 
 void SetTrimmingParamsCenter(Service::Interface* self) {
     u32* cmd_buff = Kernel::GetCommandBuffer();
 
-    u8 port = cmd_buff[1] & 0xFF;
-    s16 trimW = cmd_buff[2] & 0xFFFF;
-    s16 trimH = cmd_buff[3] & 0xFFFF;
-    s16 camW = cmd_buff[4] & 0xFFFF;
-    s16 camH = cmd_buff[5] & 0xFFFF;
+    u8 port_select = cmd_buff[1] & 0xFF;
```
lioncash: These should probably be const u32s which or explicitly casted.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+    if (VerifyBitSet(port_select, 2)) {
+        for (int i : BitSetToIDs(port_select)) {
+            ports[i].x0 = (cam_w - trim_w) / 2;
+            ports[i].y0 = (cam_h - trim_h) / 2;
+            ports[i].x1 = ports[i].x0 + trim_w;
+            ports[i].y1 = ports[i].y0 + trim_h;
+        }
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid port_select=%d", port_select);
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+        cmd_buff[1] = ERROR_INVALID_ENUM_VALUE.raw;
+    }
 
     cmd_buff[0] = IPC::MakeHeader(0x12, 1, 0);
-    cmd_buff[1] = RESULT_SUCCESS.raw;
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, port=%d, trimW=%d, trimH=%d, camW=%d, camH=%d",
-                port, trimW, trimH, camW, camH);
+    LOG_DEBUG(Service_CAM, "called, port_select=%d, trim_w=%d, trim_h=%d, cam_w=%d, cam_h=%d",
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
-    cmd_buff[1] = RESULT_SUCCESS.raw;
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, cam_select=%d", cam_select);
+    LOG_DEBUG(Service_CAM, "called, camera_select=%d", camera_select);
+}
+
+void SwitchContext(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 camera_select = cmd_buff[1] & 0xFF;
```
lioncash: These should probably be const u32s.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+            cameras[camera].current_context = context;
+            const auto& context_config = cameras[camera].contexts[context];
+            cameras[camera].impl->SetFlip(context_config.flip);
+            cameras[camera].impl->SetEffect(context_config.effect);
+            cameras[camera].impl->SetFormat(context_config.format);
+            cameras[camera].impl->SetResolution(context_config.resolution);
+        }
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid camera_select=%d, context_select=%d", camera_select,
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+                  context_select);
+        cmd_buff[1] = ERROR_INVALID_ENUM_VALUE.raw;
+    }
+
+    cmd_buff[0] = IPC::MakeHeader(0x14, 1, 0);
+
+    LOG_DEBUG(Service_CAM, "called, camera_select=%d, context_select=%d", camera_select,
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
-    LOG_WARNING(Service_CAM, "(STUBBED) called, cam_select=%d, flip=%d, context=%d", cam_select,
-                flip, context);
+    LOG_DEBUG(Service_CAM, "called, camera_select=%d, flip=%d, context_select=%d", camera_select,
+              static_cast<int>(flip), context_select);
+}
+
+void SetDetailSize(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 camera_select = cmd_buff[1] & 0xFF;
```
lioncash: This should probably be a `const u32`.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid camera_select=%d, context_select=%d", camera_select,
+                  context_select);
+        cmd_buff[1] = ERROR_INVALID_ENUM_VALUE.raw;
+    }
+
+    cmd_buff[0] = IPC::MakeHeader(0x1E, 1, 0);
+
+    LOG_DEBUG(Service_CAM, "called, camera_select=%d, width=%d, height=%d, crop_x0=%d, crop_y0=%d, "
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+    LOG_DEBUG(Service_CAM, "called, camera_select=%d, flip=%d, context_select=%d", camera_select,
+              static_cast<int>(flip), context_select);
+}
+
+void SetDetailSize(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 camera_select = cmd_buff[1] & 0xFF;
+    Resolution resolution;
+    resolution.width = cmd_buff[2] & 0xFFFF;
```
lioncash: These need explicit casts to `u16`.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+    resolution.height = cmd_buff[3] & 0xFFFF;
+    resolution.crop_x0 = cmd_buff[4] & 0xFFFF;
+    resolution.crop_y0 = cmd_buff[5] & 0xFFFF;
+    resolution.crop_x1 = cmd_buff[6] & 0xFFFF;
+    resolution.crop_y1 = cmd_buff[7] & 0xFFFF;
+    u8 context_select = cmd_buff[8] & 0xFF;
```
lioncash: This should probably be a `const u32`.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+            for (int context : BitSetToIDs(context_select)) {
+                cameras[camera].contexts[context].resolution = resolution;
+                if (cameras[camera].current_context == context) {
+                    cameras[camera].impl->SetResolution(resolution);
+                }
+            }
+        }
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid camera_select=%d, context_select=%d", camera_select,
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
     cmd_buff[0] = IPC::MakeHeader(0x20, 1, 0);
+
+    LOG_WARNING(Service_CAM, "(STUBBED) called, camera_select=%d, frame_rate=%d", camera_select,
+                static_cast<int>(frame_rate));
+}
+
+void SetEffect(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 camera_select = cmd_buff[1] & 0xFF;
```
lioncash: This an context_select can be `const u32`.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+            for (int context : BitSetToIDs(context_select)) {
+                cameras[camera].contexts[context].effect = effect;
+                if (cameras[camera].current_context == context) {
+                    cameras[camera].impl->SetEffect(effect);
+                }
+            }
+        }
+        cmd_buff[1] = RESULT_SUCCESS.raw;
+    } else {
+        LOG_ERROR(Service_CAM, "invalid camera_select=%d, context_select=%d", camera_select,
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+                  context_select);
+        cmd_buff[1] = ERROR_INVALID_ENUM_VALUE.raw;
+    }
+
+    cmd_buff[0] = IPC::MakeHeader(0x22, 1, 0);
+
+    LOG_DEBUG(Service_CAM, "called, camera_select=%d, effect=%d, context_select=%d", camera_select,
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+              static_cast<int>(effect), context_select);
+}
+
+void SetOutputFormat(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 camera_select = cmd_buff[1] & 0xFF;
```
lioncash: This and context_select can be `const u32s`.

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+    cmd_buff[0] = IPC::MakeHeader(0x25, 1, 0);
+
+    LOG_DEBUG(Service_CAM, "called, camera_select=%d, format=%d, context_select=%d", camera_select,
+              static_cast<int>(format), context_select);
+}
+
+void SynchronizeVsyncTiming(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    u8 camera_select1 = cmd_buff[1] & 0xFF;
```
lioncash: These can just be `const u32`

src/core/hle/service/cam/cam.cpp
```diff
@@ -112,104 +419,482 @@ void GetMaxLines(Service::Interface* self) {
...
+    u8 camera_select2 = cmd_buff[2] & 0xFF;
+
+    cmd_buff[0] = IPC::MakeHeader(0x29, 1, 0);
     cmd_buff[1] = RESULT_SUCCESS.raw;
 
-    LOG_WARNING(Service_CAM, "(STUBBED) called, cam_select=%d, frame_rate=%d", cam_select,
-                frame_rate);
+    LOG_WARNING(Service_CAM, "(STUBBED) called, camera_select1=%d, camera_select2=%d",
```
lioncash: `%u`

src/core/hle/service/cam/cam.cpp
```diff
@@ -263,24 +1026,62 @@ void PlayShutterSound(Service::Interface* self) {
...
+        camera.impl = Camera::CreateCamera(Settings::values.camera_name[camera_id],
+                                           Settings::values.camera_config[camera_id]);
+        camera.impl->SetFlip(camera.contexts[0].flip);
+        camera.impl->SetEffect(camera.contexts[0].effect);
+        camera.impl->SetFormat(camera.contexts[0].format);
+        camera.impl->SetResolution(camera.contexts[0].resolution);
+    }
+
+    for (auto& port : ports) {
+        port.completion_event->Clear();
```
lioncash: You may just want to give `PortConfig` a `Clear` member function, which you can call here.

src/core/hle/service/cam/cam.h
```diff
@@ -183,36 +178,82 @@ struct PackageParameterCameraSelect {
...
+    Effect effect;
+    Resolution resolution;
+};
+
+static_assert(sizeof(PackageParameterWithContextDetail) == 16,
+              "PackageParameterWithContextDetail structure size is wrong");
 
 /**
- * Unknown
+ * Start the selected ports capturing.
```
lioncash: `Starts capturing at the selected port`

src/core/hle/service/cam/cam.h
```diff
@@ -183,36 +178,82 @@ struct PackageParameterCameraSelect {
...
+ *      1: u8 selected port
  *  Outputs:
  *      0: 0x00010040
  *      1: ResultCode
  */
 void StartCapture(Service::Interface* self);
 
 /**
- * Unknown
+ * Stops the selected ports capturing.
```
lioncash: `Stops capturing from the selected port`

src/core/hle/service/cam/cam.h
```diff
@@ -222,10 +263,10 @@ void StopCapture(Service::Interface* self);
 void GetVsyncInterruptEvent(Service::Interface* self);
 
 /**
- * Unknown
+ * Gets the event associated with buffer error interrupts of the selected port.
```
lioncash: This probably needs to be phrased better (I can't offer a better suggestion because I can't parse/understand it).

src/core/frontend/camera/interface.h
```diff
@@ -0,0 +1,63 @@
...
+#include "common/common_types.h"
+#include "core/hle/service/cam/cam.h"
+#include "core/hle/service/y2r_u.h"
+
+namespace Camera {
+
+/// An abstract class standing for a camera. All camera implementation should inherit from this.
+class CameraInterface {
+public:
+    virtual ~CameraInterface(); // !!!
```
wwylele: sorry, forgot to remove it. it was a save point when I was working on it

src/core/hle/service/cam/cam.h
wwylele: I actually don't understand it very much either... I will change it back to "Unknown"

src/core/hle/service/cam/cam.cpp
wwylele: I made this int because it can beecome negative below. I will make an explicit cast.
`dest_size` is also `int` to match the type of `src_size` so that they can be used in `std::min` without cast there

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,108 +2,415 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
```
wwylele: fixed

wwylele:
Addressed comments above (hopefully, I probably missed one or two because there are so many...). Thanks @lioncash for reviewing!

wwylele:
I should have used `Common::BitSet`. Will fix it.

src/citra/citra.cpp
```diff
@@ -126,6 +127,8 @@ int main(int argc, char** argv) {
     Settings::values.use_gdbstub = use_gdbstub;
     Settings::Apply();
 
+    Camera::Init();
```
bunnei: Why is this called directly by the frontend, instead of somewhere from Core::System::Init?

wwylele: My thought is that the Qt frontend can configure cameras by retrieving a list of registered camera factories (not implemented yet). This requires that the camera factory system should be still alive when no game is running, so it shouldn't be initialized with the core system, but be initialized on citra booting.

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,214 +2,941 @@
...
+        x0 = 0;
+        y0 = 0;
+        x1 = 0;
+        y1 = 0;
+        transfer_bytes = 256;
+    }
+};
+
+// built-in resolution parameters
+constexpr std::array<Resolution, 8> PRESET_RESOLUTION{{
```
bunnei: These and the below can be `static`

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,214 +2,941 @@
...
+std::array<CameraConfig, 3> cameras;
+std::array<PortConfig, 2> ports;
+int completion_event_callback;
+
+const ResultCode ERROR_INVALID_ENUM_VALUE(ErrorDescription::InvalidEnumValue, ErrorModule::CAM,
+                                          ErrorSummary::InvalidArgument, ErrorLevel::Usage);
+const ResultCode ERROR_OUT_OF_RANGE(ErrorDescription::OutOfRange, ErrorModule::CAM,
+                                    ErrorSummary::InvalidArgument, ErrorLevel::Usage);
+
+void CompletionEventCallBack(u64 port_id, int) {
```
bunnei: `static`

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,214 +2,941 @@
...
+        Memory::WriteBlock(port.dest, buffer.data(), std::min<u32>(port.dest_size, buffer_size));
+    }
+
+    port.is_receiving = false;
+    port.completion_event->Signal();
+}
+
+// Starts a receiving process on the specified port. This can only be called when is_busy = true and
+// is_receiving = false.
+void StartReceiving(int port_id) {
```
bunnei: `static`

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,214 +2,941 @@
...
+    CoreTiming::ScheduleEvent(
+        msToCycles(LATENCY_BY_FRAME_RATE[static_cast<int>(camera.frame_rate)]),
+        completion_event_callback, port_id);
+}
+
+// Cancels any ongoing receiving processes at the specified port. This is used by functions that
+// stop capturing.
+// TODO: what is the exact behaviour on real 3DS when stopping capture during an ongoing process?
+//       Will the completion event still be signaled?
+void CancelReceiving(int port_id) {
```
bunnei: `static`

src/core/frontend/camera/interface.cpp
```diff
@@ -0,0 +1,11 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/frontend/camera/interface.h"
+
+namespace Camera {
+
+CameraInterface::~CameraInterface() = default;
```
wwylele: I was asked by @lioncash  to move the = default to a cpp file 

> The destructor's implementation should also be placed in the cpp file, otherwise the vtable will be created in every translation unit that uses this class.

src/core/hle/service/cam/cam.cpp
wwylele: As proposed by @lioncash, these variables and functions are in a anonymous namespace, thus `static` is not needed.

src/core/hle/service/cam/cam.cpp
```diff
@@ -239,6 +966,90 @@ void GetStereoCameraCalibrationData(Service::Interface* self) {
...
+void SetPackageParameterWithoutContext(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    PackageParameterWithoutContext package;
+    std::memcpy(&package, cmd_buff + 1, sizeof(package));
+
+    cmd_buff[0] = IPC::MakeHeader(0x33, 1, 0);
+    cmd_buff[1] = RESULT_SUCCESS.raw;
+
+    LOG_DEBUG(Service_CAM, "(STUBBED) called");
```
wwylele: yeah should be WARNING. The reason this is stubbed now is the same reason as I left those `SetWhiteBalance`-like function unimplemented: they are not that important and I don't want to make a bunch of functions to increase the diff (= I am lazy). Anyone else wants them in the future can easily write them. I made this function stubbed instead of leaving it unimplemented, though, because there was an unused `struct PackageParameterCameraSelect` in `cam.h`, and I want to correct the name and show how it is used, so others can continue the work based on this.

src/core/frontend/camera/interface.cpp
bunnei: Fair enough

src/core/hle/service/cam/cam.cpp
bunnei: Fair enough

Subv:
After this is merged, https://github.com/citra-emu/citra/issues/1886#issuecomment-261732936 has to be updated.

src/citra/config.cpp
```diff
@@ -91,6 +91,21 @@ void Config::ReadValues() {
     Settings::values.is_new_3ds = sdl2_config->GetBoolean("System", "is_new_3ds", false);
     Settings::values.region_value = sdl2_config->GetInteger("System", "region_value", 1);
 
+    // Camera
+    using namespace Service::CAM;
+    Settings::values.camera_name[OuterRightCamera] =
+        sdl2_config->Get("Camera", "camera_outer_right_name", "Blank camera");
```
Subv: Why are the defaults different between SDL and Qt? ("Blank camera" vs "blank")

src/core/frontend/camera/interface.h
```diff
@@ -0,0 +1,62 @@
...
+     * For the meaning of the parameters, please refer to Service::CAM::Resolution. Note that the
+     * actual camera implementation doesn't need to respect all the parameters. However, the width
+     * and the height parameters must be respected and be used to determine the size of output
+     * frames.
+     * @param resolution The resolution parameters to set
+     */
+    virtual void SetResolution(const Service::CAM::Resolution& resolution) = 0;
+
+    /**
+     * Sets the vertical and horizontal flip applied to all frames received after this function is
```
Subv: This (and SetEffect's) docstring is not entirely clear to me, who is supposed to apply the flip? the caller of this function, or the CameraInterface itself?

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,214 +2,941 @@
...
+    int current_context;
+    FrameRate frame_rate;
+};
+
+struct PortConfig {
+    int camera_id;
+
+    bool is_active;            // set when the port is activated by an Activate call.
+    bool is_pending_receiving; // set if SetReceiving is called when is_busy = false. When
+                               // StartCapture is called then, this will trigger an receiving
```
Subv: a receiving process

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,214 +2,941 @@
...
+                               // process and reset itself.
+    bool is_busy;      // set when StartCapture is called and reset when StopCapture is called.
+    bool is_receiving; // set when there is an ongoing receiving process.
+
+    // trimming related
```
Subv: This comment doesn't add anything productive

src/core/hle/service/cam/cam.cpp
```diff
@@ -2,214 +2,941 @@
...
+    {160, 120, 0, 0, 639, 479},  // QQVGA
+    {352, 288, 26, 0, 613, 479}, // CIF
+    {176, 144, 26, 0, 613, 479}, // QCIF
+    {256, 192, 0, 0, 639, 479},  // DS_LCD
+    {512, 384, 0, 0, 639, 479},  // DS_LCDx4
+    {400, 240, 0, 48, 639, 431}, // CTR_TOP_LCD
+}};
+
+// latency in ms for each frame rate option
+constexpr std::array<int, 13> LATENCY_BY_FRAME_RATE{{
```
Subv: Where do these values come from?

src/core/frontend/camera/interface.h
wwylele: The `CameraInterface` itself. I am not good at wording these... any suggestions?

src/core/hle/service/cam/cam.cpp
wwylele: The values are taken from 1000/frame_rate where frame_rate are the largest number documented on each `enum FrameRate` option

src/core/frontend/camera/interface.h
Subv:
````
Sets the vertical and horizontal flip to be applied to all frames.
````

Comment (wwylele) [inline src/core/frontend/camera/interface.h]: From IRC:
```
<wwylele> Subv: (regarding to the Camera interface documentation) you know what, the original documentation I wrote was "Sets the vertical and horizontal flip applying on the frame."
<wwylele> and lioncash asked me to change it to what it is now
<Subv> lol
<Subv> well, i'm not a native speaker so *shrug*
<wwylele> me neither *shrug*
```
So I am not going to change this until some native speakers give better idea

Dragios: `All frames are set according to their respective vertical and horizontal flip`

Is that better? I am a native speaker btw.

yuriks: @Dragios No, that doesn't mean the same thing, as far as I can tell? Subv's sounds fine to me but I'm not a native speaker either. Maybe "Configures whether received frames should be flipped vertically or horizontally by the camera."

yuriks: Actually, "Configures how received frames should be flipped by the camera." since it's an enum.

wwylele:
Updated the documentation for `SetFlip` and `SetEffect`

wwylele:
Just moved around the added content in default_ini.h so that it won't conflict with other PR in bleeding edge merge

Subv:
Anyone else wants to review this? If not, expect to see it merged anytime now.

yuriks:
@Subv looks like you forgot to merge it ;)

Issue #2371, opened 2016-12-24
Severe performance regression encountered on a few games


Dragios:
Performance regression due to PR #2343. Some games are less likely affected but a few are hit severely.
Since no one has reported this issue during *Bleeding Edge* testing, I assumed that it only affects older generation processors.
Left & right image is tested on build commit 17d7402 & aa47af7 respectively

There are other games with varying performance within a certain degree.
I used this 3 games as the example.

Nano Assault is least affected
![Nano Assault](https://cloud.githubusercontent.com/assets/18496868/21465980/8177a6b6-c9f2-11e6-9efb-e76e59d9ab12.png)

Performance hit is moderately affected but at least it is still playable
![Pokémon Mystery Dungeon – Gates to Infinity](https://cloud.githubusercontent.com/assets/18496868/21465981/82d1e788-c9f2-11e6-97e4-19c871475d6e.png)

Performance hit is severely affected for this particular game making it not playable at all
![Pokémon Super Mystery Dungeon](https://cloud.githubusercontent.com/assets/18496868/21465982/83e3c948-c9f2-11e6-9552-d16df23709ea.png)
With the log filter set to Info level the game slow down until becoming like a slide show. <1 FPS

**OS:** _Windows 10 build 14393.576 (codename Redstone 1)_
**Processor:** _Intel(R) Core(TM)2 Duo CPU E7500  @ 2.93GHz, 2926 Mhz, 2 Core(s), 2 Logical Processor(s) [Overclocked to 3.7 GHz]_
**GPU:** _nVidia GT 610 2GB DDR3 (Asus)_
**Driver:** _Version 376.33_
**OpenGL:** _Version 4.5_

wwylele:
#2343 has several commits in it. Could you please bisect the regression to one commit?

Dragios:
I will try. That will take a lot of time

wwylele:
@Dragios I didn't record the FPS these days on my side, nor paid attention to the speed change, so I can't tell...

Dragios:
I have successfully bisected the commit that causes the regression.
It was commit 4fc8b8229ed1d9ea9d20faee7059c898265db6cf

wwylele:
@Dragios does #2372 fix it?

Dragios:
It is. Thanks for the quick fix @wwylele

Dragios:
Yes it does, Thanks for the quick fix @wwylele

bunnei:
[...]

PR #2372, opened 2016-12-24
core: fix removed AddTicks

wwylele:
Another regression from #2343 (4fc8b82, issue #2371). The `AddTicks(4000);` hack didn't get moved to the new function. Plus the old `hle.cpp` wasn't removed (it even get edited in the next commit...).


Dragios:
Although this fix speed up game but I encounter a problem with a game.
*Nano Assault EX* will occasionally stuck at black screen upon starting the game mode with the `AddTicks(4000);` hack.

Although a restart will be able to bypass it at certain time and will play a very short video after that.
Left side shows the black screen freezes while right side shows that it managed to load properly

Nano Assault EX
I have to slow down the FPS to 10 in order to screenshot "loading" as it disappear too quick.

wwylele:
@Dragios before I look into it, I want to make sure that this issue didn't happen before#2343, right?

Dragios:
Actually it is the opposite. It happens before #2343 is merged. After #2343 is merged it does not happen.

wwylele:
@Dragios then filing a new issue after this is merged. #2343 didn't mean to change the behaviour (= fix anything). This PR is mean to restore the old behaviour that is changed by mistake.

Dragios:
Will do.

bunnei:
Sorry, but this wasn't a "regression" ... It was intentionally removed. I'm not sure this does anything useful anymore. It was added at the time (as a hack) because it seemed to fix hangs in several games. That does not seem to be the case anymore. This is effectively just resulting in a vsync happening more frequently in areas of games that aren't GPU bound .. effectively the same as underclocking the CPU, but even more hacky IMO.

Issue #2373, opened 2016-12-24
unlock locked option once emulation has been stopped 


RavenHome1:
bleeding 108

currently options which are locked when emulation is running do not get unlocked when user stops the emulation from the drop down menu (complete stop and not pause), not sure if intended or a minor bug but decided to report it anyway.

setting that remain locked after emulation has been stopped are

-cpu jit option in the general tab

-the entire system tab

-Vsync option in the graphics tab

RavenHome1:
fixed by https://github.com/citra-emu/citra/pull/2374

PR #2374, opened 2016-12-24
Core: reset cpu_core in Shutdown to make IsPoweredOn work properly

wwylele:
Yet another regression from #2343 (232ef55c1a13552e5ba8b72d61d1d072f5851598, issue #2373). `cpu_core` doesn't get reset when shutting down, causing `IsPoweredOn(){return cpu_core != nullptr;}` not working properly. I kind of want to restore the old `bool is_powered_on;` method, but not sure.

_Three regressions from one PR is ridiculous._

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -159,6 +159,7 @@ void System::Shutdown() {
     Kernel::Shutdown();
     HW::Shutdown();
     CoreTiming::Shutdown();
+    cpu_core.reset();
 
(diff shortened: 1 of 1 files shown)


Subv:
>Three regressions from one PR is ridiculous.

Just par for the course

bunnei:
> I kind of want to restore the old bool is_powered_on; method, but not sure.
Please don't unless there is a legitimate reason (which, there isn't)

PR #2376, opened 2016-12-27
Core: remove unused hle.cpp

wwylele:
Sorry for misunderstanding #2343 and opening #2372. I wish next time the PR description can be more specific rather than "removed some cruft that I don't think does anything" so that I can catch up any discussion I missed.

Anyway, `hle.cpp` is unused now and should be removed, right?


bunnei:
Indeed - sorry for the lack of description, I thought that I had separated that change into its own specific commit, but it seems that I forgot to. Thanks.

PR #2382, opened 2016-12-29
Service/NFC: stub GetTagInRangeEvent

mailwl:
This PR mostly for pass Fatal Error in game **Mini-Mario & Friends - amiibo Challenge** (no amiibo emulating, so, no way to pass amiibo scanning)

### DIFF
--- a/src/core/hle/service/nfc/nfc.cpp
+++ b/src/core/hle/service/nfc/nfc.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include "core/hle/kernel/event.h"
 #include "core/hle/service/nfc/nfc.h"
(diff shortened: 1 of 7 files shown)


lioncash:
You should probably include the what games this primarily applied to in your commit message as well.

src/core/hle/service/nfc/nfc.cpp
```diff
@@ -2,16 +2,34 @@
...
 
 namespace Service {
 namespace NFC {
 
+static Kernel::SharedPtr<Kernel::Event> tag_in_range_event;
+
+void GetTagInRangeEvent(Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
```
Subv: Please add the proper command response header and the correct descriptor in cmd_buff[2]

PR #2384, opened 2016-12-30
config: Add option for specifying screen resolution scale factor.

bunnei:
For those of us with hardware fast enough to upscale, but not fast enough to upscale to our native resolutions...

Obligatory UI change screenshot:

image

### DIFF
--- a/src/citra/default_ini.h
+++ b/src/citra/default_ini.h
@@ -51,9 +51,10 @@ use_hw_renderer =
 # 0: Interpreter (slow), 1 (default): JIT (fast)
 use_shader_jit =
 
-# Whether to use native 3DS screen resolution or to scale rendering resolution to the displayed screen size.
-# 0 (default): Native, 1: Scaled
-use_scaled_resolution =
+# Resolution scale factor
+# 0: Auto (scales resolution to window size), 1: Native 3DS screen resolution, Otherwise a scale
+# factor for the 3DS resolution
+resolution_factor =
 
 # Whether to enable V-Sync (caps the framerate at 60FPS) or not.
 # 0 (default): Off, 1: On
--- a/src/citra_qt/configure_graphics.cpp
+++ b/src/citra_qt/configure_graphics.cpp
@@ -18,10 +18,81 @@ ConfigureGraphics::ConfigureGraphics(QWidget* parent)
 
 ConfigureGraphics::~ConfigureGraphics() {}
 
+enum class Resolution : int {
+    Auto,
+    Scale1x,
+    Scale2x,
+    Scale3x,
+    Scale4x,
+    Scale5x,
+    Scale6x,
+    Scale7x,
+    Scale8x,
+    Scale9x,
+    Scale10x,
+};
+
+float ToResolutionFactor(Resolution option) {
+    switch (option) {
+    case Resolution::Auto:
+        return 0.f;
+    case Resolution::Scale1x:
+        return 1.f;
+    case Resolution::Scale2x:
+        return 2.f;
+    case Resolution::Scale3x:
+        return 3.f;
+    case Resolution::Scale4x:
+        return 4.f;
+    case Resolution::Scale5x:
+        return 5.f;
+    case Resolution::Scale6x:
+        return 6.f;
+    case Resolution::Scale7x:
+        return 7.f;
+    case Resolution::Scale8x:
+        return 8.f;
+    case Resolution::Scale9x:
+        return 9.f;
+    case Resolution::Scale10x:
+        return 10.f;
+    }
+    return 0.f;
+}
+
+Resolution FromResolutionFactor(float factor) {
+    if (factor == 0.f) {
+        return Resolution::Auto;
+    } else if (factor == 1.f) {
+        return Resolution::Scale1x;
+    } else if (factor == 2.f) {
+        return Resolution::Scale2x;
+    } else if (factor == 3.f) {
+        return Resolution::Scale3x;
+    } else if (factor == 4.f) {
+        return Resolution::Scale4x;
+    } else if (factor == 5.f) {
+        return Resolution::Scale5x;
+    } else if (factor == 6.f) {
+        return Resolution::Scale6x;
+    } else if (factor == 7.f) {
+        return Resolution::Scale7x;
+    } else if (factor == 8.f) {
+        return Resolution::Scale8x;
+    } else if (factor == 9.f) {
+        return Resolution::Scale9x;
+    } else if (factor == 10.f) {
+        return Resolution::Scale10x;
+    }
+    return Resolution::Auto;
+}
+
 void ConfigureGraphics::setConfiguration() {
     ui->toggle_hw_renderer->setChecked(Settings::values.use_hw_renderer);
+    ui->resolution_factor_combobox->setEnabled(Settings::values.use_hw_renderer);
     ui->toggle_shader_jit->setChecked(Settings::values.use_shader_jit);
-    ui->toggle_scaled_resolution->setChecked(Settings::values.use_scaled_resolution);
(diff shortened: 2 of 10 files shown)


src/citra_qt/configure_graphics.cpp
```diff
@@ -18,10 +18,82 @@ ConfigureGraphics::ConfigureGraphics(QWidget* parent)
...
+    Scale2_5x,
+    Scale3x,
+    Scale4x,
+    Scale5x,
+    Scale6x,
+    Scale7x,
+    Scale8x,
+};
+
+float ToResolutionFactor(Resolution option) {
```
Subv: All cases here have narrowing conversions from double to float.

src/citra_qt/configure_graphics.cpp
```diff
@@ -18,10 +18,82 @@ ConfigureGraphics::ConfigureGraphics(QWidget* parent)
...
+        return 6.0;
+    case Resolution::Scale7x:
+        return 7.0;
+    case Resolution::Scale8x:
+        return 8.0;
+    }
+    return 0.0;
+}
+
+Resolution FromResolutionFactor(float factor) {
```
Subv: Comparing floats to doubles, does that not generate a warning?

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -560,10 +560,18 @@ RasterizerCacheOpenGL::GetFramebufferSurfaces(const Pica::Regs::FramebufferConfi
         auto layout = VideoCore::g_emu_window->GetFramebufferLayout();
 
         // Assume same scaling factor for top and bottom screens
-        color_params.res_scale_width = depth_params.res_scale_width =
-            (float)layout.top_screen.GetWidth() / VideoCore::kScreenTopWidth;
-        color_params.res_scale_height = depth_params.res_scale_height =
-            (float)layout.top_screen.GetHeight() / VideoCore::kScreenTopHeight;
+        if (Settings::values.resolution_factor == 0.0) {
```
Subv: Same

PEmu1:
I think it would be best to not introduce 1.5x and 2.5x scaling options. Integer scaling factors generally seem to introduce fewer issues (not sure if this is completely true with Citra but 1.5x and 2.5x have caused some headaches for Dolphin) and they make the dropdown list look like a bit of a mess. Given the option I think it would be better longterm to not add them in the first place.

src/citra_qt/configure_graphics.cpp
bunnei: Not on Visual Studio 2015

bunnei:
@PEmu1 What sort of headaches has this caused on Dolphin? I'm having a hard time imagining why supporting them on Citra would introduce any issues that would be distinct from other upscaled sizes. 

I'm not sure how things work with Dolphin, but Citra can do *lots* of copies between system memory and VRAM with certain games for things like framebuffers and textures. Increasing the internal resolution increases the size of these copies exponentially, quickly causing even really strong hardware to crawl. In my own testing (and playing games), I've found that 1.5x and 2.5x are pretty useful, as they are often the highest I can go (and still maintain ~60fps) and yet they still provide a pretty obvious graphical improvement over 1x.

Will add 10x though, although it'll probably be a while before Citra can run anything reasonably on that 😛

PEmu1:
@bunnei I don't think there's a complete list of issues that it causes but some info on the problems can be found here: https://bugs.dolphin-emu.org/issues/8747 This may be partly due to the GameCube's architecture so I am not sure if similar issues would impact Citra or not but it is perhaps worth keeping in mind. Personally I would like to kill off those two options in Dolphin so seeing them being introduced in Citra years later is a bit painful...

I also think it's an ugly convention for the UI (having a partial list of half steps) but I find others often don't necessarily seem to care as much about minor UI details as I apparently do. If you do decide to go along with half step options it may make more sense to utilize a slider UI element or something else rather than a dropdown box.

> Will add 10x though, although it'll probably be a while before Citra can run anything reasonably on that 😛

I was considering asking for up to 5K support but thought that may be asking too much. 😛

PS: I noticed that the resolution listings are specifically related to the top screen only. Is that something that should be addressed?

bunnei:
@PEmu1 You are correct in that the resolution listed in the dropdown box is that of the top screen, however the scale factor is applied to both screens' internal resolutions. However, I think listing just the top screen resolution is both useful and what most people care about. Also, we do not currently support separate resolution scaling factors for both screens.

SonofUgly:
@bunnei 
>I've found that 1.5x and 2.5x are pretty useful, as they are often the highest I can go (and still maintain ~60fps) 

Really, what hardware/games was that framerate tested on? On a 750ti it isn't until 7x that OoT3D and ALBW start being bottlenecked by rendering (150fps at 1x-6x, 145fps at 7x, 75fps at 10x). And in the one scene I tested with Kirby Robobot, it wasn't until 9x before framerate started to go down (103fps to 98fps, and still 83fps at 10x). Would've thought the only use 1.5x and 2.5x would have are for games that have emulation bugs that hammer the GPU unnecessarily or something like that.

Also, out of scope for this PR, but the downscaling filter could be better. Guessing it's just doing bilinear right now.

bunnei:
@SonofUgly the games you listed actually aren't particularly demanding with regards to copies, *especially* OoT. Nonetheless, I have to assume you have pretty top (and perhaps overclocked) hardware to have ALBW running at ~145fps at 7x. I'm guessing that isn't indicative of our typical userbase...

I'm testing on an i7-4710HQ/GTX 860M. Most games (other than OoT) run fullspeed anywhere between 1x and 3x.

tabnk:
Been try to build this PR with GCC C++ complier but it cause a lot of fps drop in game. However, Microsoft Visual C++ Compiler is fine and no speed loss.

src/citra/default_ini.h
```diff
@@ -51,10 +51,14 @@ use_hw_renderer =
 # 0: Interpreter (slow), 1 (default): JIT (fast)
 use_shader_jit =
 
-# Whether to use native 3DS screen resolution or to scale rendering resolution to the displayed screen size.
+# Whether to use native 3DS screen resolution or to scale rendering resolution by `resolution_factor`
```
yuriks: Afaict this parameter is now redundant, it has exactly the same effect as setting resolution_factor to 1. If we can remove it I think that would be better.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -560,10 +560,18 @@ RasterizerCacheOpenGL::GetFramebufferSurfaces(const Pica::Regs::FramebufferConfi
...
+            // Auto - scale resolution to the window size
+            color_params.res_scale_width = depth_params.res_scale_width =
+                (float)layout.top_screen.GetWidth() / VideoCore::kScreenTopWidth;
+            color_params.res_scale_height = depth_params.res_scale_height =
+                (float)layout.top_screen.GetHeight() / VideoCore::kScreenTopHeight;
+        } else {
+            // Otherwise, scale the resolution by the specified factor
+            color_params.res_scale_width = depth_params.res_scale_width =
+                color_params.res_scale_height = depth_params.res_scale_height =
+                    Settings::values.resolution_factor;
```
yuriks: Nit, but this looks kinda weird. Maybe split into 4 separate assignments instead.

bunnei:
Updated screenshot:

image

baka0815:
It's integer scaling only now, but the value is stored as a floating point value - why?

In ``FromResolutionFactor()`` the check is only on an equals basis, so if the config has values like 1.5 you will  get 0 (auto scaled) instead of 1.0 or 2.0 which would be nearer to the setting.
But as I stated above: why allow floating point values if the scaling is integer only?

PR #2399, opened 2017-01-03
Fix some warnings

johnathan79717:
I fixed some warnings including unused variable, format, override, initialization order.

### DIFF
--- a/src/core/hle/service/boss/boss.cpp
+++ b/src/core/hle/service/boss/boss.cpp
@@ -33,7 +33,7 @@ void InitializeSession(Service::Interface* self) {
     cmd_buff[0] = IPC::MakeHeader(0x1, 0x1, 0);
     cmd_buff[1] = RESULT_SUCCESS.raw;
 
-    LOG_WARNING(Service_BOSS, "(STUBBED) unk_param=0x%016X, translation=0x%08X, unk_param4=0x%08X",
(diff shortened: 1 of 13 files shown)


src/core/hle/service/boss/boss.cpp
```diff
@@ -33,7 +33,7 @@ void InitializeSession(Service::Interface* self) {
     cmd_buff[0] = IPC::MakeHeader(0x1, 0x1, 0);
     cmd_buff[1] = RESULT_SUCCESS.raw;
 
-    LOG_WARNING(Service_BOSS, "(STUBBED) unk_param=0x%016X, translation=0x%08X, unk_param4=0x%08X",
+    LOG_WARNING(Service_BOSS, "(STUBBED) unk_param=0x%016llX, translation=0x%08X, unk_param4=0x%08X",
```
Subv: This one should use the macro we have for 64-bit values

src/core/hle/service/mic_u.cpp
```diff
@@ -97,7 +97,7 @@ static void StartSampling(Interface* self) {
 
     cmd_buff[1] = RESULT_SUCCESS.raw; // No error
     is_sampling = true;
-    LOG_WARNING(Service_MIC, "(STUBBED) called, encoding=%u, sample_rate=%u, "
```
lioncash: You need to static_cast enums to an integral type, you can't printf them directly, otherwise they cause warnings on higher warning levels.

johnathan79717:
Thanks @Subv and @lioncash.  I've made the changes you suggested.

PR #2402, opened 2017-01-04
QT: GameList don't attemp to load non-game files

Kloen:
When loading game_list, we are loading all files inside the directory, despite they are any kind of file. This fills the log with a few unnecesary lines regarding the file load of X file, which are of no use at all. With this, we only attemp to load the files we commonly use (The ones we can load through File->Load File...)

### DIFF
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <QFileInfo>
 #include <QHeaderView>
 #include <QMenu>
 #include <QThreadPool>
@@ -131,6 +132,14 @@ void GameList::LoadInterfaceLayout() {
     item_model->sort(header->sortIndicatorSection(), header->sortIndicatorOrder());
 }
 
+const QStringList GameList::supported_file_extensions = {"3ds", "3dsx", "elf", "axf",
+                                                         "cci", "cxi",  "app"};
+
+static bool HasSupportedFileExtension(const std::string& file_name) {
+    QFileInfo file = QFileInfo(file_name.c_str());
+    return GameList::supported_file_extensions.contains(file.completeSuffix(), Qt::CaseInsensitive);
+}
+
 void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsigned int recursion) {
     const auto callback = [this, recursion](unsigned* num_entries_out, const std::string& directory,
                                             const std::string& virtual_name) -> bool {
@@ -139,7 +148,7 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
         if (stop_processing)
             return false; // Breaks the callback loop.
 
-        if (!FileUtil::IsDirectory(physical_name)) {
+        if (!FileUtil::IsDirectory(physical_name) && HasSupportedFileExtension(physical_name)) {
             std::unique_ptr<Loader::AppLoader> loader = Loader::GetLoader(physical_name);
             if (!loader)
                 return true;
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -466,9 +466,14 @@ void GMainWindow::OnGameListOpenSaveFolder(u64 program_id) {
 }
 
 void GMainWindow::OnMenuLoadFile() {
-    QString filename =
-        QFileDialog::getOpenFileName(this, tr("Load File"), UISettings::values.roms_path,
-                                     tr("3DS executable (*.3ds *.3dsx *.elf *.axf *.cci *.cxi)"));
+    QString extensions;
+    for (const auto& piece : game_list->supported_file_extensions)
+        extensions += "*." + piece + " ";
+
+    QString file_filter = tr("3DS executable") + " (" + extensions + ")";
+
+    QString filename = QFileDialog::getOpenFileName(this, tr("Load File"),
+                                                    UISettings::values.roms_path, file_filter);
     if (!filename.isEmpty()) {
         UISettings::values.roms_path = QFileInfo(filename).path();
 
--- a/src/citra_qt/game_list.h
+++ b/src/citra_qt/game_list.h
@@ -33,6 +33,8 @@ public:
     void SaveInterfaceLayout();
     void LoadInterfaceLayout();
 
+    static const QStringList supported_file_extensions;
+
 signals:
     void GameChosen(QString game_path);
     void ShouldCancelWorker();


src/citra_qt/game_list.cpp
```diff
@@ -131,6 +131,15 @@ void GameList::LoadInterfaceLayout() {
     item_model->sort(header->sortIndicatorSection(), header->sortIndicatorOrder());
 }
 
+bool GameListWorker::IsGameFile(std::string const& fileName) {
```
wwylele: This can be just a local `static` function

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +131,15 @@ void GameList::LoadInterfaceLayout() {
...
+    std::string extension;
+    Common::SplitPath(fileName, nullptr, nullptr, &extension);
+
+    return (extension.compare(".3ds") == 0 || extension.compare(".3dsx") == 0 ||
+            extension.compare(".elf") == 0 || extension.compare(".axf") == 0 ||
+            extension.compare(".cci") == 0 || extension.compare(".cxi") == 0);
```
wwylele: I would prefer using `std::unordered_set` to achieve this. It is more readable and easy to extend

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +132,14 @@ void GameList::LoadInterfaceLayout() {
     item_model->sort(header->sortIndicatorSection(), header->sortIndicatorOrder());
 }
 
+static std::unordered_set<std::string> extensions = { ".3ds", ".3dsx", ".elf", ".axf", ".cci", ".cxi" };
```
Kloen: I really think that this should be somewhere else, but don't know what could be a suitable place for this. Suggestions are accepted.

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +132,14 @@ void GameList::LoadInterfaceLayout() {
...
+
+static bool IsGameFile(const std::string& fileName) {
+    std::string extension;
+    Common::SplitPath(fileName, nullptr, nullptr, &extension);
+    return (extensions.find(extension) != extensions.end());
```
lioncash: the outer set of parentheses are unnecessary here.

src/citra_qt/game_list.cpp
lioncash: This can probably be a `static const` variable inside the function itself.

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +132,14 @@ void GameList::LoadInterfaceLayout() {
     item_model->sort(header->sortIndicatorSection(), header->sortIndicatorOrder());
 }
 
+static std::unordered_set<std::string> extensions = { ".3ds", ".3dsx", ".elf", ".axf", ".cci", ".cxi" };
+
+static bool IsGameFile(const std::string& fileName) {
```
lioncash: `filename` or `file_name`

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +132,15 @@ void GameList::LoadInterfaceLayout() {
     item_model->sort(header->sortIndicatorSection(), header->sortIndicatorOrder());
 }
 
+static bool IsGameFile(const std::string& file_name) {
+    static const std::unordered_set<std::string> extensions = {".3ds", ".3dsx", ".elf",
```
jroweboy: Should this list somehow be shared with the list in the file open dialog found here? https://github.com/citra-emu/citra/blob/master/src/citra_qt/main.cpp#L472

PPLToast:
The extension line only works for uncapitalized names, so if the file is for example .3DS, it won't be recognized.

Kloen:
@PPLToast nice catch. Should be fixed by now.

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +132,16 @@ void GameList::LoadInterfaceLayout() {
...
+                                                               ".axf", ".cci",  ".cxi"};
+
+    std::string extension;
+    Common::SplitPath(file_name, nullptr, nullptr, &extension);
+    std::transform(extension.begin(), extension.end(), extension.begin(), tolower);
```
wwylele: `std::tolower`

src/citra_qt/game_list.cpp
Kloen: I have an approach in mind to do this, will push it later today.

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +132,15 @@ void GameList::LoadInterfaceLayout() {
     item_model->sort(header->sortIndicatorSection(), header->sortIndicatorOrder());
 }
 
+static bool IsGameFile(const std::string& file_name) {
+    std::string extension;
+    Common::SplitPath(file_name, nullptr, nullptr, &extension);
+    std::transform(extension.begin(), extension.end(), extension.begin(),
+                   (int (*)(int))std::tolower);
```
Subv: Just leave it as ::tolower instead of resorting to ugly casts to hint the correct overload.

src/citra_qt/ui_settings.h
```diff
@@ -42,4 +43,7 @@ struct Values {
 };
 
 extern Values values;
+
+static const std::unordered_set<std::string> allowed_file_extensions = {".3ds", ".3dsx", ".elf",
```
Subv: Drop the static, though I would suggest defining this in the cpp and just declaring it here.

src/citra_qt/ui_settings.cpp
```diff
@@ -7,4 +7,7 @@
 namespace UISettings {
 
 Values values = {};
+
+const std::unordered_set<std::string> allowed_file_extensions = {".3ds", ".3dsx", ".elf",
+                                                                 ".axf", ".cci",  ".cxi"};
```
wwylele: This looks weird to me to have this constant in `ui_settings.cpp/.h` (which is supposed to contain configurable data). I am not sure where is the good place to put this, though... probably just put them in `main.cpp/.h`?

Kloen: Well, it's here, because it's used on both `main.cpp` and `game_list.cpp`, this could be on `game_list,h/cpp`, but I don't think it fits there due to the use on `main.cpp` which is not the in the same context entirely, it's here because I think on it as a ui related kind of setting. So I put it here. If I put it on `main.h/cpp` it would end in a circular dependency between `main` and `game_list`. Could put it on `game_list.h/cpp` but it's out of the file context too, so ¯\\_(ツ)_/¯

yuriks: Also, you should add `.app` (Alternative extension for NCCH files.) to this list while you're at it.

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +132,14 @@ void GameList::LoadInterfaceLayout() {
     item_model->sort(header->sortIndicatorSection(), header->sortIndicatorOrder());
 }
 
+static bool IsGameFile(const std::string& file_name) {
+    std::string extension;
+    Common::SplitPath(file_name, nullptr, nullptr, &extension);
+    std::transform(extension.begin(), extension.end(), extension.begin(), ::tolower);
+    return UISettings::allowed_file_extensions.find(extension) !=
+           UISettings::allowed_file_extensions.end();
+}
```
yuriks: This whole function can be significantly simplified if you just use QString, QStringList and QFileInfo instead of the stdlib/common types.

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +132,14 @@ void GameList::LoadInterfaceLayout() {
     item_model->sort(header->sortIndicatorSection(), header->sortIndicatorOrder());
 }
 
+static bool IsGameFile(const std::string& file_name) {
```
yuriks: `HasSupportedFileExtension` is a better name

Kloen:
@yuriks Everything addressed.

src/citra_qt/main.cpp
```diff
@@ -466,9 +466,14 @@ void GMainWindow::OnGameListOpenSaveFolder(u64 program_id) {
...
 
 void GMainWindow::OnMenuLoadFile() {
-    QString filename =
-        QFileDialog::getOpenFileName(this, tr("Load File"), UISettings::values.roms_path,
-                                     tr("3DS executable (*.3ds *.3dsx *.elf *.axf *.cci *.cxi)"));
+    QString extensions;
+    for (const auto& piece : game_list->supported_file_extensions)
+        extensions += "*." + piece + " ";
+
+    QString file_filter = "3DS executable (" + extensions + ")";
```
yuriks: "3DS executable" should be in a `tr()` (because it's a translatable string)

src/citra_qt/main.cpp
```diff
@@ -466,9 +466,14 @@ void GMainWindow::OnGameListOpenSaveFolder(u64 program_id) {
...
 
 void GMainWindow::OnMenuLoadFile() {
-    QString filename =
-        QFileDialog::getOpenFileName(this, tr("Load File"), UISettings::values.roms_path,
-                                     tr("3DS executable (*.3ds *.3dsx *.elf *.axf *.cci *.cxi)"));
+    QString extensions;
+    for (const auto& piece : game_list->supported_file_extensions)
+        extensions += "*." + piece + " ";
+
+    QString file_filter = tr("3DS executable") + "(" + extensions + ")";
```
yuriks: You forgot a space here before the `(`. I'll just fix it myself before merging.

PR #2404, opened 2017-01-05
Kernel: Fix SharedMemory objects always returning error when addr = 0

### LINKED ISSUE #2400: Creating SharedMemory objects with addr = 0 always returns an error
Currently applications can not create shared memory objects with addr = 0 due to this check https://github.com/citra-emu/citra/blob/master/src/core/hle/svc.cpp#L893

Using addr = 0 means that the kernel will take care of allocating the backing space for the memory block.

The fix is to just add ````addr != 0```` to the if.

hyperdrive3:
Closes #2400


yuriks:
...you messed it up. This will fail 100% of the time, because either addr is 0 (in which case `addr < PROCESS_IMAGE_VADDR` is true) or isn't 0 (so `addr != 0`) is true, resulting in the error path always being taken.

hyperdrive3:
.. yep haha. Have fixed it. Will open a new PR with less retardation.

yuriks:
You can re-open this PR and then push on the branch again, no need to create a new one.

src/core/hle/svc.cpp
```diff
@@ -890,7 +890,8 @@ static ResultCode CreateMemoryBlock(Kernel::Handle* out_handle, u32 addr, u32 si
         return ResultCode(ErrorDescription::InvalidCombination, ErrorModule::OS,
                           ErrorSummary::InvalidArgument, ErrorLevel::Usage);
 
-    if (addr < Memory::PROCESS_IMAGE_VADDR || addr + size > Memory::SHARED_MEMORY_VADDR_END) {
+    if ((addr < Memory::PROCESS_IMAGE_VADDR || addr + size > Memory::SHARED_MEMORY_VADDR_END) &&
```
Subv: Please add a TODO here saying that processes with the APPLICATION memory type are not allowed to create memory blocks with addr = 0, any attempts to do so should return error 0xD92007EA

PR #2406, opened 2017-01-05
Kernel: Properly update port counters on session disconnection.

Subv:
This PR is supposed to fix some of the TODOs introduced by #2249.

Namely properly managing the current connection count of the kernel ports.
The current connection count is increased when a session is created to the port, and it is decremented when said session is destroyed.
The port will be kept alive so long as the session is still alive.

There are still a few things left, like awaking all the waiting threads of a ClientPort when its connection count goes below the max, but that's something that will be implemented later when ClientPort inherits from WaitObject.

This implementation depends deeply on not keeping any extra owning references to ServerSessions, there should only be one for HLE services: the one stored in SessionRequestHandler::connected_sessions.
If this invariant is violated, ServerSessions will not be properly destroyed and the ClientPort's connection count will not be updated.

We don't have to care about non-HLE services because we won't be keeping any references to ServerSessions in this case, except in the pending_sessions list of the ServerPort.
The server application will take care of managing the lifetime of the ServerSessions for us (via AcceptSession and CloseHandle)

# Future changes
- [ ] Add a way to determine when a thread should sleep during a call to svcSendSyncRequest
- [ ] Add a way to wake up a thread that is sleeping on svcSendSyncRequest and set the synchronization result (tentative name for the function: ReturnFromSyncRequest(ResultCode) )
- [ ] Wake up the affected threads when one of the synchronization primitives triggers an error (When any of the endpoints of a session closes while the client is waiting for the synchronization result)
 Implement svcAcceptSession, it would only be used by the LLE services.
- [ ] Implement svcReplyAndReceive, it would only be used by the LLE services. This call works pretty much like WaitSynchronizationN.
- [ ] Implement command buffer translation inside svcReplyAndReceive.
- [ ] Copy the command buffer to a local buffer before calling HandleSyncRequest in SessionRequestHandler.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2406)

### DIFF
--- a/src/core/hle/kernel/client_session.cpp
+++ b/src/core/hle/kernel/client_session.cpp
@@ -30,7 +30,6 @@ ResultVal<SharedPtr<ClientSession>> ClientSession::Create(std::string name) {
 
     client_session->name = std::move(name);
     client_session->parent = nullptr;
-    client_session->session_status = SessionStatus::Open;
     return MakeResult<SharedPtr<ClientSession>>(std::move(client_session));
 }
 
--- a/src/core/hle/kernel/client_session.cpp
+++ b/src/core/hle/kernel/client_session.cpp
@@ -14,27 +14,33 @@ ClientSession::~ClientSession() {
     // This destructor will be called automatically when the last ClientSession handle is closed by
     // the emulated application.
 
-    if (server_session->hle_handler)
-        server_session->hle_handler->ClientDisconnected(server_session);
+    if (parent->server) {
+        if (parent->server->hle_handler)
+            parent->server->hle_handler->ClientDisconnected(parent->server);
 
-    // TODO(Subv): If the session is still open, set the connection status to 2 (Closed by client),
-    // wake up all the ServerSession's waiting threads and set the WaitSynchronization result to
-    // 0xC920181A.
+        // TODO(Subv): Force a wake up of all the ServerSession's waiting threads and set
+        // their WaitSynchronization result to 0xC920181A.
+    }
+
+    parent->client = nullptr;
 }
 
-ResultVal<SharedPtr<ClientSession>> ClientSession::Create(ServerSession* server_session,
-                                                          std::string name) {
+ResultVal<SharedPtr<ClientSession>> ClientSession::Create(std::string name) {
     SharedPtr<ClientSession> client_session(new ClientSession);
 
     client_session->name = std::move(name);
-    client_session->server_session = server_session;
+    client_session->parent = nullptr;
     client_session->session_status = SessionStatus::Open;
     return MakeResult<SharedPtr<ClientSession>>(std::move(client_session));
 }
 
 ResultCode ClientSession::SendSyncRequest() {
     // Signal the server session that new data is available
-    return server_session->HandleSyncRequest();
+    if (parent->server)
+        return parent->server->HandleSyncRequest();
+
+    return ResultCode(ErrorDescription::SessionClosedByRemote, ErrorModule::OS,
+                      ErrorSummary::Canceled, ErrorLevel::Status);
 }
 
 } // namespace
--- a/src/core/hle/kernel/client_session.cpp
+++ b/src/core/hle/kernel/client_session.cpp
@@ -25,14 +25,6 @@ ClientSession::~ClientSession() {
     parent->client = nullptr;
 }
 
-ResultVal<SharedPtr<ClientSession>> ClientSession::Create(std::string name) {
-    SharedPtr<ClientSession> client_session(new ClientSession);
-
-    client_session->name = std::move(name);
-    client_session->parent = nullptr;
-    return MakeResult<SharedPtr<ClientSession>>(std::move(client_session));
-}
-
 ResultCode ClientSession::SendSyncRequest() {
     // Signal the server session that new data is available
     if (parent->server)
--- a/src/core/hle/kernel/server_session.cpp
+++ b/src/core/hle/kernel/server_session.cpp
@@ -14,8 +14,15 @@ ServerSession::ServerSession() = default;
 ServerSession::~ServerSession() {
     // This destructor will be called automatically when the last ServerSession handle is closed by
     // the emulated application.
-    // TODO(Subv): Reduce the ClientPort's connection count,
-    // if the session is still open, set the connection status to 3 (Closed by server),
+
+    // Decrease the port's connection count.
+    if (parent->port)
+        parent->port->active_sessions--;
+
+    // TODO(Subv): Wake up all the ClientSession's waiting threads and set
+    // the SendSyncRequest result to 0xC920181A.
+
+    parent->server = nullptr;
 }
 
 ResultVal<SharedPtr<ServerSession>> ServerSession::Create(
@@ -25,6 +32,7 @@ ResultVal<SharedPtr<ServerSession>> ServerSession::Create(
     server_session->name = std::move(name);
     server_session->signaled = false;
     server_session->hle_handler = std::move(hle_handler);
+    server_session->parent = nullptr;
 
     return MakeResult<SharedPtr<ServerSession>>(std::move(server_session));
 }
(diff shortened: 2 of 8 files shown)


src/core/hle/kernel/client_session.cpp
```diff
@@ -14,27 +14,31 @@ ClientSession::~ClientSession() {
...
     return MakeResult<SharedPtr<ClientSession>>(std::move(client_session));
 }
 
 ResultCode ClientSession::SendSyncRequest() {
     // Signal the server session that new data is available
-    return server_session->HandleSyncRequest();
+    if (parent->server)
+        return parent->server->HandleSyncRequest();
+
+    return ResultCode(ErrorDescription(26), ErrorModule::OS, ErrorSummary::Canceled, ErrorLevel::Status);
```
Subv: This one is supposed to be SessionClosedByRemote

src/core/hle/kernel/client_session.cpp
```diff
@@ -14,27 +14,31 @@ ClientSession::~ClientSession() {
...
-    if (server_session->hle_handler)
-        server_session->hle_handler->ClientDisconnected(server_session);
+    if (parent->server) {
+        if (parent->server->hle_handler)
+            parent->server->hle_handler->ClientDisconnected(parent->server);
 
-    // TODO(Subv): If the session is still open, set the connection status to 2 (Closed by client),
-    // wake up all the ServerSession's waiting threads and set the WaitSynchronization result to
-    // 0xC920181A.
+        // TODO(Subv): Wake up all the ServerSession's waiting threads and set
```
Subv: What would be a good way to do this?

src/core/hle/kernel/server_session.cpp
```diff
@@ -14,8 +14,11 @@ ServerSession::ServerSession() = default;
 ServerSession::~ServerSession() {
     // This destructor will be called automatically when the last ServerSession handle is closed by
     // the emulated application.
-    // TODO(Subv): Reduce the ClientPort's connection count,
-    // if the session is still open, set the connection status to 3 (Closed by server),
+
+    // Decrease the port's connection count.
+    if (parent->port)
+        parent->port->active_sessions--;
+    parent->server = nullptr;
```
Subv: We have to wake the threads that are waiting for the SendSyncRequest call to complete.

src/core/hle/kernel/client_session.cpp
Subv: We also have to wake the threads that are waiting for the SendSyncRequest call to complete and set the result of SendSyncRequest to 0xC920181A.

src/core/hle/kernel/server_session.cpp
```diff
@@ -14,8 +14,11 @@ ServerSession::ServerSession() = default;
 ServerSession::~ServerSession() {
     // This destructor will be called automatically when the last ServerSession handle is closed by
     // the emulated application.
-    // TODO(Subv): Reduce the ClientPort's connection count,
-    // if the session is still open, set the connection status to 3 (Closed by server),
+
+    // Decrease the port's connection count.
+    if (parent->port)
```
wwylele: Is there any code that set `port` to `nullptr`? Probably in `~ClientPort` this pointer should be reset?

yuriks: I think the port pointer should just be a regular SharedPtr, unless ports also have similar "sensing" behavior like sessions endpoints do?

src/core/hle/kernel/client_session.cpp
```diff
@@ -14,27 +14,31 @@ ClientSession::~ClientSession() {
     // This destructor will be called automatically when the last ClientSession handle is closed by
     // the emulated application.
 
-    if (server_session->hle_handler)
-        server_session->hle_handler->ClientDisconnected(server_session);
+    if (parent->server) {
+        if (parent->server->hle_handler)
+            parent->server->hle_handler->ClientDisconnected(parent->server);
```
JayFoxRox:
```cpp
if (parent->server && parent->server->hle_handler)
  parent->server->hle_handler->ClientDisconnected(parent->server);
```

Maybe? (Both look ugly imo) #bikeshed

JayFoxRox:
What's the state of this? It's been "rotting" here for more than 3 months, it is also in BE.

It really sucks that this didn't get any review because it was explained in the first post. You also had questions in your code.
I can only speak for myself to explain the lack of review / feedback: I feel this might have to do with the lack of documentation about the feature in question or a lack of links to such documentation. We only have few people experienced with the HLE and especially these more low-level parts of the HLE. I'm not one of those and honestly have almost no idea what I'm looking at and how it should be or what it currently does.
I think the red X about the clang-format also scares people from reviewing.

Maybe @Lectem @wwylele or @yuriks can review with a focus on those questions proposed by @Subv ?
No PR should have to wait this long.. (especially not design feedback)

Lectem:
Sorry but I'm not familiar enough with the low level session system to review this, though it does look ok at first glance

src/core/hle/kernel/server_session.cpp
Subv: @wwylele There can exist Sessions without any associated Ports

src/core/hle/kernel/client_session.cpp
yuriks: The version with the collapsed conditions looks better, but the intent is that more code would be added into the outer if below.

Subv:
I'll refactor this (move Session to its own file) and take it out of RFC state tomorrow so that it can have a chance of being merged.

Subv:
This is now out of RFC state, please review

src/core/hle/kernel/session.h
```diff
@@ -0,0 +1,29 @@
...
+namespace Kernel {
+
+class ClientSession;
+class ClientPort;
+class ServerSession;
+
+/**
+ * Parent structure to link the client and server endpoints of a session with their associated
+ * client port.
+ * The client port need not exist, as is the case for portless sessions like the FS File and
```
yuriks: What's up with the weird jagged wrapping here?

src/core/hle/kernel/client_session.cpp
```diff
@@ -14,27 +14,35 @@ ClientSession::~ClientSession() {
...
 
-    // TODO(Subv): If the session is still open, set the connection status to 2 (Closed by client),
-    // wake up all the ServerSession's waiting threads and set the WaitSynchronization result to
-    // 0xC920181A.
+        // TODO(Subv): Force a wake up of all the ServerSession's waiting threads and set
+        // their WaitSynchronization result to 0xC920181A.
+
+        // TODO(Subv): Wake up all the ClientSession's waiting threads and set
+        // the SendSyncRequest result to 0xC920181A.
```
yuriks: How would this be even possible? If ClientSession is being destroyed then it means there are no more references to it.

src/core/hle/kernel/client_session.cpp
```diff
@@ -14,27 +14,35 @@ ClientSession::~ClientSession() {
...
+    }
+
+    parent->client = nullptr;
 }
 
-ResultVal<SharedPtr<ClientSession>> ClientSession::Create(ServerSession* server_session,
-                                                          std::string name) {
+ResultVal<SharedPtr<ClientSession>> ClientSession::Create(std::string name) {
```
yuriks: Does this need to be exposed? Are ClientSessions ever created stand-alone? Otherwise, can ClientSession be made a friend of ServerSession so that everything gets done in CreateSessionPair?

src/core/hle/kernel/server_session.h
```diff
@@ -51,7 +54,8 @@ class ServerSession final : public WaitObject {
      */
     static SessionPair CreateSessionPair(
         const std::string& name = "Unknown",
-        std::shared_ptr<Service::SessionRequestHandler> hle_handler = nullptr);
+        std::shared_ptr<Service::SessionRequestHandler> hle_handler = nullptr,
+        ClientPort* client_port = nullptr);
```
yuriks: Should be a SharedPtr

src/core/hle/kernel/server_session.cpp
```diff
@@ -61,13 +69,19 @@ ResultCode ServerSession::HandleSyncRequest() {
 }
 
 ServerSession::SessionPair ServerSession::CreateSessionPair(
-    const std::string& name, std::shared_ptr<Service::SessionRequestHandler> hle_handler) {
+    const std::string& name, std::shared_ptr<Service::SessionRequestHandler> hle_handler,
+    ClientPort* port) {
```
yuriks: Add a blank line below this.

src/core/hle/kernel/client_session.cpp
Subv: This is already the case, ClientSession::Create is private

yuriks: My point is that, in that case, there isn't much of a point in having this function around instead of CreateSessionPair construction the ClientSession directly.

yuriks:
LGTM!

PR #2408, opened 2017-01-05
Kernel: Removed the priority boost code for starved threads.

Subv:
After hwtesting and reverse engineering the kernel, it was found that the CTROS scheduler performs no priority boosting for threads like this, although some other forms of scheduling priority-starved threads might take place.

For example, it was found that hardware interrupts might cause low-priority threads to run if the CPU is preempted in the middle of an SVC handler that deschedules the current (high priority) thread before scheduling it again.

### DIFF
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -156,28 +156,6 @@ void ArbitrateAllThreads(u32 address) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


JayFoxRox:
Just for reference, can you link your hardware test?

Subv:
This is the test I used https://gist.github.com/Subv/fe6d4abdb69fb349c22f69204e3e6529

PR #2410, opened 2017-01-05
Don't yield execution in SleepThread(0) if there are no available threads to run

### LINKED ISSUE #2383: Severe Kernel Rescheduling stalling on Super Smash Bros: Smash Run
I've been running some tests to measure different performance bottlenecks on different games. This particular game has scheduling issues as you can see from the images:
issue1

With CPU Jit Off:
issue2

how to reproduce the problem?
Enter super smash, go into smash run, you'll notice a massive lag all over the place.

tested on:
3f6c82b, 14f8065 and personal branch.

related Pull Requests::
#2260

Subv:
With this we avoid an useless temporary deschedule of the current thread.

Closes #2383

### DIFF
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -849,6 +849,11 @@ static ResultCode CancelTimer(Kernel::Handle handle) {
 static void SleepThread(s64 nanoseconds) {
     LOG_TRACE(Kernel_SVC, "called nanoseconds=%lld", nanoseconds);
 
+    // Don't attempt to yield execution if there are no available threads to run,
+    // this way we avoid a useless reschedule to the idle thread.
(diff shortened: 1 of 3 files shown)


src/core/hle/svc.cpp
```diff
@@ -849,6 +849,11 @@ static ResultCode CancelTimer(Kernel::Handle handle) {
 static void SleepThread(s64 nanoseconds) {
     LOG_TRACE(Kernel_SVC, "called nanoseconds=%lld", nanoseconds);
 
+    // Don't attempt to yield execution if there are no available threads to run,
+    // this way we avoid an useless reschedule to the idle thread.
```
Dragios: Nit: `a useless reschedule ...`

Issue #254, opened 2014-12-04
[RFC] Crypto library


Subv:
A crypto library will be needed for AES support, I propose PolarSSL because it is modular and somewhat lightweight, we could only import the Cipher-AES module.

Requesting suggestions. Please keep in mind that the AES library needs to support 128-bit keys, along with CBC mode and CTR mode. Other key sizes and modes might be needed in the future but for now these are the requirements.

vgturtle127:
There is also OpenAES, Crypto++, GnuTLS, WolfSSL, and obviously OpenSSL.

PolarSSL seems good though. I have heard good things.

I have personally used OpenSSL and GnuTLS and they are pretty good. Obviously everybody remembers Heartbleed, and thinks that automatically means everybody should avoid OpenSSL. The bug is fixed now though, so I see no issues for using it. Some people might be against it though.

Not sure about WolfSSL and OpenAES.

SSLeay is still used by many applications, but I think it is really old and outdated.

Subv:
Resolved as of #2412

PR #2412, opened 2017-01-06
Added Crypto++ as crypto library

wwylele:
I am looking into implementing APT:Wrap/Unwrap these days. However, a crypto library is needed for it. So this is an RFC to import a crypto library before I PR the Wrap/Unwrap implementation. This is also an attempt to address #254. I didn't check all the crypto libraries mentioned there, but I find [Crypto++ library](https://www.cryptopp.com/) good, for the following reasons:
 - It contains all crypto feature supported by 3DS afaik: AES (with ECB, CTR, CBC and CCM mode), SHA, and RSA.
 - It is a class-based C++ library. What's more, its class interface allows me to do the trick to implement 3DS non-standard crypto algorithm (such as [non-standard AES-CCM used by Wrap/Unwrap](https://github.com/wwylele/citra/commit/6c2b3c8ceeac5053c624b0219f7ed4d489eb8e67#diff-bc0e955915363089200e77a484b6abf5R20))

How do we integrate Crypto++ into citra? Here are some options:
 - Use the precompiled package of library. Users need to install them first before compiling citra
   - Or add it to our repo and automatically download it in cmake, like we did for Qt and SDL
 - Import it as a submodule in citra repo and compile it with citra.
   - Or we manage our own fork with only the needed module imported (I have no luck separating out the modules, though. It seems that Crypto++ is not that modularized) 

I currently import the whole library as a submodule. However, I encountered some problems: the `CMakeLists.txt` included in the library is bad. It pollutes the cmake option list (something like `DISABLE_SSE3` with no prefix), and pollutes the installation list (`make install` command will install both citra and Crypto++). To resolve this, I have to create my own `CMakeLists.txt` outside the submodule. Everything seems working fine from now, but another issue appears in MSVC, as I write in `CMakeLists.txt`:
> ...(rdrand.asm in the submodule is causing problem) Because there is also a rdrand.S file in the submodule, CMake will specify the target path for rdrand.asm as "/crytopp.dir/{Debug|Release}/cryptopp/rdrand.asm.obj". The additional target folder "cryptopp" being specified is because the file rdrand.asm is in the source folder "cryptopp". But MSVC assembler can't build target file to an non-existing folder("cryptopp"). 

Currently the workaround is to exclude the file `rdrand.asm` (it is used for hardware random number generator, which will probably not be used by citra). This will not be a problem if we manage our own fork, so that we can replace the CMakeLists in the submodule.


mailwl:
do we need some keys? or plain text enough?

wwylele:
@mailwl That will be the talk in my next PR. Please focus on the library use here.
Just a quick answer to your question: yes, user needs to provide keys to get correct behaviour.

Subv:
Is there any possibility of upstreaming the cmake changes to cryptopp?

wwylele:
@Subv I think it is possible, but I don't want to spend much time on it for now. The cmake change is not a bug fix or something trivial, but related to how they define their product. If they treat crypto++ as a standalone library that should be build with only itself (which seems to be the case from their building system), then it is more likely that they don't need to change anything. I expect a long talk if I am going to upstream the change, but I don't have that time for it these days.

mailwl:
without it PR would be compile?













'

mailwl:
some problem with travis

mailwl:
it good idea to add asm to clean citra c++ code?

wwylele:
@mailwl I don't care it much if the asm code is in just the library but not in our core code, plus crypto++ allows you to disable asm code by `CRYPTOPP_DISABLE_ASM`.  Probably others have different idea, but anyway, 

>maybe we write it itself? in c++

no, I would say if crypto++ is not as good as we want, we just go to find another library. It is really wasting time to create the wheel again when there are already *a lot of* crypto library

src/core/CMakeLists.txt
```diff
@@ -342,10 +342,10 @@ set(HEADERS
             settings.h
             )
 
-include_directories(../../externals/dynarmic/include)
+include_directories(../../externals/dynarmic/include ../../externals/cryptopp)
```
mailwl: maybe split into 2 include_directories()?
these things not related

wwylele:
Since only a few people commented, I assume no one is against this...? So I took off [RFC] and this is ready for review (and merge?)

mailwl:
Yes! need to merge

yuriks:
Is Crypto++ readily available as a package in Linux distros and on OS X? In that case I'm ok just including the binaries like we do with Qt/SDL. Otherwise I think it's worth making the CMake option work well for us. Can't you just do whatever it is that the library is doing to resolve the compilation conflict?

Other than that, SGTM

wwylele:
It is available on OS X, and [generally in Linux distros](https://www.cryptopp.com/wiki/Linux#Linux_Distributions). I am fine with both options, but iirc @lioncash said to include the source in citra.

> Can't you just do whatever it is that the library is doing to resolve the compilation conflict?
The problem only presents when I make the CMakeLists outside cryptopp source folder, even if using the original CMakeLists file (with path modified of course). This is more like a bug in CMake or MSVC.

CMakeLists.txt
```diff
@@ -283,6 +283,8 @@ endif()
 
 add_subdirectory(externals/soundtouch)
 
+add_subdirectory(externals/cryptopp)
```
yuriks: Can you move this to the CMakeLists in externals/ instead?

externals/cryptopp/CMakeLists.txt
```diff
@@ -0,0 +1,209 @@
...
+#============================================================================
+# Third-party libraries
+#============================================================================
+
+if(WIN32)
+    target_link_libraries(cryptopp ws2_32)
+endif()
+
+# This may need to be expanded to "Solaris"
+if ("${UNAME_SYSTEM}" STREQUAL "SunOS")
```
yuriks: We don't really need support for this.

externals/cryptopp/CMakeLists.txt
```diff
@@ -0,0 +1,209 @@
...
+
+# Library sources. You can use the GNUmakefile to generate the list: `make sources`.
+file(GLOB cryptopp_SOURCES cryptopp/*.cpp)
+list(REMOVE_ITEM cryptopp_SOURCES
+        # These are removed in the original CMakeLists.txt
+        ${CMAKE_CURRENT_SOURCE_DIR}/cryptopp/pch.cpp
+        ${CMAKE_CURRENT_SOURCE_DIR}/cryptopp/simple.cpp
+        ${CMAKE_CURRENT_SOURCE_DIR}/cryptopp/winpipes.cpp
+        ${CMAKE_CURRENT_SOURCE_DIR}/cryptopp/cryptlib_bds.cpp
+        ${cryptopp_SOURCES_TEST}
```
yuriks: Do you still need the tests here?

externals/cryptopp/CMakeLists.txt
```diff
@@ -0,0 +1,209 @@
...
+# -march=native for GCC, Clang and ICC in any version that does support it.
+if ((NOT CRYPTOPP_DISABLE_CXXFLAGS_OPTIMIZATIONS) AND (NOT CRYPTOPP_CROSS_COMPILE) AND (CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU|Intel"))
+    CHECK_CXX_COMPILER_FLAG("-march=native" COMPILER_OPT_ARCH_NATIVE_SUPPORTED)
+    if (COMPILER_OPT_ARCH_NATIVE_SUPPORTED AND NOT CMAKE_CXX_FLAGS MATCHES "-march=")
+        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
+    endif()
+endif()
+
+# Solaris specific 
+if ((NOT CRYPTOPP_CROSS_COMPILE) AND "${UNAME_SYSTEM}" STREQUAL "SunOS")
```
yuriks: We don't need this platform.

externals/cryptopp/CMakeLists.txt
wwylele: No I don't. This is the `REMOVE_ITEM` block, not the source adding block. I probably worded bad here. By "These are removed in the original CMakeLists.txt" I mean: these files are removed in the original CMakeLists, so I do the same thing here without thinking much

CMakeLists.txt
wwylele: I thought this is preferred? (See above for those soundtouch etc. subdirectory)

wwylele:
(Hided comment in diff)

>Can you move this to the CMakeLists in externals/ instead?

I thought this is preferred? (See above for those soundtouch etc. subdirectory)

Or we want to begin moving all subdirectory include to externals? If it's this case I will do it

CMakeLists.txt
yuriks: We should be adding all new externals there. And eventually if the file is cleaned up the existing ones too, because they clutter up the main file too much, and have no sensible ordering or anything right now.

externals/cryptopp/CMakeLists.txt
yuriks: Oh, derp, nevermind. I didn't pay enough attention to what this was actually doing. Not explicitly listing files is an anit-pattern, but it doesn't matter much for an external library.

PR #2414, opened 2017-01-07
Texture decoding cleanups

yuriks:
Moves the texture lookup function out of DebugUtils and cleans up the addressing code to explicitly work with image tiles, which makes the addressing code much more straightforward.

**NOTE:** I pushed this because @Subv told me to. :< GPU emulation works fine, but the various debug widgets are untested and probably broken. Thus this is an RFC until I test and fix those.


Subv:
I've added this PR to the bleeding edge build so that we can crowd-test it

TTFH:
Compilation fail, need to include  "video_core/texture/texture_decode.h"
in video_core/debug_utils/debug_utils.cpp and in _void DumpTexture(TextureConfig&, u8*)_ use:
Pica::Texture::TextureInfo info;

yuriks:
Rebased and fixed the build error. Will try to do some testing of the debug widgets soon.

yuriks:
Apparently the debug widgets just... work? Including depth textures. ¯\_(ツ)_/¯

Taking this out of RFC then.

yuriks:
Ping for reviews?

MerryMage:
Could it be documented / `DEBUG_ASSERT`-ed that the `x`, `y` parameters in  `LookupTexelInTile` must be `< 8`?

yuriks:
Rebased/addressed.

@MerryMage Added both!

PR #2421, opened 2017-01-09
Timers: Immediately signal the timer if it was started with an initial value of 0

Subv:
Also added some error checking in the SetTimer svc.

Huge thanks to @TuxSH for spotting these deficiencies.

Please note that the timer_callback_handle_table HandleTable is currently a source of memory leaks due to Timer objects never reaching 0 references. (ping @yuriks )

### DIFF
--- a/src/core/hle/kernel/timer.h
+++ b/src/core/hle/kernel/timer.h
@@ -54,6 +54,14 @@ public:
     void Cancel();
     void Clear();
 
+    /**
+     * Signals the timer, waking up any waiting threads and rescheduling it
+     * for the next interval.
+     * This method should not be called from outside the timer callback handler,
+     * lest multiple callback events get scheduled.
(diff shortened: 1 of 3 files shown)


src/core/hle/kernel/timer.h
```diff
@@ -52,6 +52,12 @@ class Timer final : public WaitObject {
     void Cancel();
     void Clear();
 
+    /**
+     * Signals the timer, waking up any waiting threads and rescheduling it
+     * for the next interval.
+     */
+    void Signal(int cycles_late);
```
yuriks: Add a warning saying that this shouldn't be called from outside of the timer callback, because otherwise you'll get multiple scheduled events.

bunnei:
This is breaking BE builds. Please rebase!

Issue #2093, opened 2016-09-19
Implement Looped Buffers


WishfulThinking:
Many games seem to use looped audio buffers and produce the following log message.

`Audio.DSP <Error> audio_core/hle/source.cpp:DequeueBuffer:249: Looped buffers are unimplemented at the moment`

We should implement them. 

This is an issue to track the progress of the implementation.

MerryMage:
The work would likely require: 

1. Either writing a few tests to confirm how looped buffers work or reverse engineering the DSP binary blob.
2. Implementation.

A buffer is a looped buffer if `Buffer::is_looped` is set. `SourceConfiguration::loop_related` indicates where the first sample position of the loop is. There is a second loop boolean `SourceConfiguration::looping`.

My own work on this is waiting on some DSP RE I'm doing; but I don't mind someone jumping in in the meantime.

jmerdich:
Chiming in here after doing a bunch of fiddling based on @MerryMage 's test suite on my o3ds and a few games that use loops in the emulator... 

- Simple loops are the only thing I've seen in a commercial game. Pretty much, having a buffer in the queue flagged as looping the entire, usually for a reverb effect (found this in Zelda OOT's background music, Mario Kart's button noises, and Mario 3D Land's "?" blocks).  A patch to support this would be trivial, and I can clean it up and toss one out sometime tomorrow. 
- I haven't been able to get `loop_related` to do anything on my hardware. The `play_position` register affects the start of the first playback of the embedded buffer. All following loops start from the beginning of the buffer regardless of the value of `loop_related`. Commercial games seem to be fond of setting the dirty bit for `play_position=0` when looping, though that register does not seem to be updated to any other value by either the game or the DSP engine. Further testing is definitely necessary.
- Complex loops don't seem to work with queued buffers, only the embedded ones. This might just be a timing or caching issue though (Mario Kart 7 always invalidates the cache to the DSP command buffer when playing loops, so there might be some errata).

I'd be happy to make a PR for any of this, and any additional information can help speed the latter 2 bullet points.

MerryMage:
It's been a while since I've taken a look at audio. The above statements were based on what I remembered of my previous work before I started writing everything down; apologies for what appears to be my poor memory and sending you on what seems to be a wild goose chase.

Feel free to submit a PR for what you do have; I'm currently (very slowly) reverse engineering the binary blob.

MerryMage:
Mostly implemented afaik in #2422.

[...]

PR #2422, opened 2017-01-10
Support looping HLE audio

jmerdich:
Simple loops to start fixing #2093. This handles all the loops I've seen in commercial games, though the DSP engine can in theory manipulate the way it loops beyond replaying the entire buffer (hence, warnings).

### DIFF
--- a/src/audio_core/hle/source.cpp
+++ b/src/audio_core/hle/source.cpp
@@ -176,6 +176,16 @@ void Source::ParseConfig(SourceConfiguration::Configuration& config,
                   config.physical_address, config.length, config.buffer_id);
     }
 
+    if (config.loop_related_dirty && config.loop_related != 0) {
+        LOG_WARNING(Audio_DSP, "Unhandled complex loop with loop_related=0x%08x",
+                    static_cast<u32>(config.loop_related));
+    }
+
+    if (config.play_position_dirty && config.play_position != 0) {
+        LOG_WARNING(Audio_DSP, "Unhandled complex loop with play_position=0x%08x",
+                    static_cast<u32>(config.play_position));
+    }
+
     if (config.buffer_queue_dirty) {
         config.buffer_queue_dirty.Assign(0);
         for (size_t i = 0; i < 4; i++) {
@@ -248,15 +258,14 @@ bool Source::DequeueBuffer() {
         return false;
 
     const Buffer buf = state.input_queue.top();
-    state.input_queue.pop();
+
+    if (!buf.is_looping) {
+        state.input_queue.pop();
+    }
 
     if (buf.adpcm_dirty) {
         state.adpcm_state.yn1 = buf.adpcm_yn[0];
         state.adpcm_state.yn2 = buf.adpcm_yn[1];
-    }
-
-    if (buf.is_looping) {
-        LOG_ERROR(Audio_DSP, "Looped buffers are unimplemented at the moment");
     }
 
     const u8* const memory = Memory::GetPhysicalPointer(buf.physical_address);
@@ -307,8 +316,8 @@ bool Source::DequeueBuffer() {
 
     state.current_sample_number = 0;
     state.next_sample_number = 0;
+    state.buffer_update = buf.from_queue && (state.current_buffer_id != buf.buffer_id);
     state.current_buffer_id = buf.buffer_id;
-    state.buffer_update = buf.from_queue;
 
     LOG_TRACE(Audio_DSP, "source_id=%zu buffer_id=%hu from_queue=%s current_buffer.size()=%zu",
               source_id, buf.buffer_id, buf.from_queue ? "true" : "false",
--- a/src/audio_core/hle/source.cpp
+++ b/src/audio_core/hle/source.cpp
@@ -325,9 +325,9 @@ bool Source::DequeueBuffer() {
     }
 
     // the first playthrough starts at play_position, loops start at the beginning of the buffer
-    state.current_buffer_id = buf.buffer_id;
     state.current_sample_number = (!buf.has_played) ? buf.play_position : 0;
     state.next_sample_number = state.current_sample_number;
+    state.current_buffer_id = buf.buffer_id;
     state.buffer_update = buf.from_queue && !buf.has_played;
 
     buf.has_played = true;
--- a/src/audio_core/hle/source.cpp
+++ b/src/audio_core/hle/source.cpp
@@ -180,11 +180,11 @@ void Source::ParseConfig(SourceConfiguration::Configuration& config,
                 state.mono_or_stereo,
                 state.format,
                 false,
-                config.play_position,
+                play_position,
             });
             LOG_TRACE(Audio_DSP, "enqueuing embedded addr=0x%08x len=%u id=%hu start=%u",
                       config.physical_address, config.length, config.buffer_id,
-                      config.play_position.u32());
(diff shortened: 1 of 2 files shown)


src/audio_core/hle/source.cpp
```diff
@@ -176,6 +176,16 @@ void Source::ParseConfig(SourceConfiguration::Configuration& config,
                   config.physical_address, config.length, config.buffer_id);
     }
 
+    if (config.loop_related_dirty && config.loop_related != 0) {
+        LOG_WARNING(Audio_DSP, "Unhandled complex loop with loop_related=0x%08x",
```
MerryMage: While this doesn't really do anything, 

    config.loop_related_dirty.Assign(0);

in case someone forgets in the future.

src/audio_core/hle/source.cpp
```diff
@@ -176,6 +176,16 @@ void Source::ParseConfig(SourceConfiguration::Configuration& config,
...
+                    static_cast<u32>(config.loop_related));
+    }
+
+    if (config.play_position_dirty && config.play_position != 0) {
+        LOG_WARNING(Audio_DSP, "Unhandled complex loop with play_position=0x%08x",
```
MerryMage: While this doesn't really do anything, 

    config.play_position_dirty.Assign(0);

in case someone forgets in the future.

src/audio_core/hle/source.cpp
```diff
@@ -248,17 +258,16 @@ bool Source::DequeueBuffer() {
         return false;
 
     const Buffer buf = state.input_queue.top();
-    state.input_queue.pop();
+
+    if (!buf.is_looping) {
+        state.input_queue.pop();
```
MerryMage: A comment here stating that this loops the entirety of the current buffer or something to that effect would be nice.

wwylele:
Tried this PR on _Nintendo 3DS Sound_. On current master, this app can only play the first second or so of the music. With this PR, it can play the full music, but with a problem: the music "frames" are disordered. I don't know much about HLE audio, just want to report a problem I encountered. If it is due to some unimplemented feature, then this is not blocking this PR.

(I will probably upload a sample when I get time)

jmerdich:
I've been having trouble replicating that on my machine; still showing the one-second intro of a recording then nothing. Is this a recording copied from a 3ds or a music file?

wwylele:
@jmerdich The music I tried are some regular MP3 files (which, of course, can play normally on PC and on my 3DS)

Schplee:
I think this breaks Ocarina of Time. When clicking the Ocarina icon, the game will immediately softlock. Tested with the Ocarina and Fairy Ocarina. Citra freezes, the console doesn't move, but the music in the background continues to play. 

However, when first loading your save file, when your hearts/magic bar fill, if you click the Ocarina and it freezes, that sound will infinitely play (unlike the music in the background that plays normally, but just infinitely loops).

jmerdich:
Ok, it appears that 3DS Sound's music decoder uses a single looping buffer as a FIFO, with either precise timing interaction or some unknown mechanism to keep it in sync. The reordering you hear is the buffer underflowing. 

Audio recordings (copied from an actual 3DS) don't exhibit this behavior in citra. It only plays the first chunk, just like the music decoder did before this patch.

yuriks:
So, if I understand it correctly, that isn't an issue with this PR that would block merging it, right?

src/audio_core/hle/source.cpp
```diff
@@ -158,22 +158,40 @@ void Source::ParseConfig(SourceConfiguration::Configuration& config,
...
-            config.is_looping.ToBool(),
-            config.buffer_id,
-            state.mono_or_stereo,
-            state.format,
-            false,
-        });
-        LOG_TRACE(Audio_DSP, "enqueuing embedded addr=0x%08x len=%u id=%hu",
-                  config.physical_address, config.length, config.buffer_id);
+    {
+        u32_dsp play_position = u32_dsp();
```
yuriks: This doesn't currently get initialized to 0 afaik, so if the dirty bit isn't set it'll have an undefined value, unlike what the comment below claims.

jmerdich:
No issues that I know of within the scope of this PR, anyways.

src/audio_core/hle/source.cpp
```diff
@@ -158,22 +158,40 @@ void Source::ParseConfig(SourceConfiguration::Configuration& config,
...
+                config.length,
+                static_cast<u8>(config.adpcm_ps),
+                {config.adpcm_yn[0], config.adpcm_yn[1]},
+                config.adpcm_dirty.ToBool(),
+                config.is_looping.ToBool(),
+                config.buffer_id,
+                state.mono_or_stereo,
+                state.format,
+                false,
+                play_position,
```
yuriks: Please explicitly initialize the last member here.

src/audio_core/hle/source.cpp
```diff
@@ -305,10 +323,14 @@ bool Source::DequeueBuffer() {
         break;
     }
 
-    state.current_sample_number = 0;
-    state.next_sample_number = 0;
+    // the first playthrough starts at play_position, loops start at the beginning of the buffer
+    state.current_sample_number = (!buf.has_played) ? buf.play_position : 0;
+    state.next_sample_number = state.current_sample_number;
+
+    state.buffer_update = buf.from_queue && !buf.has_played;
```
yuriks: Any reason this line got moved up from it's previous position?

src/audio_core/hle/source.cpp
```diff
@@ -158,22 +158,40 @@ void Source::ParseConfig(SourceConfiguration::Configuration& config,
...
-            config.adpcm_dirty.ToBool(),
-            config.is_looping.ToBool(),
-            config.buffer_id,
-            state.mono_or_stereo,
-            state.format,
-            false,
-        });
-        LOG_TRACE(Audio_DSP, "enqueuing embedded addr=0x%08x len=%u id=%hu",
-                  config.physical_address, config.length, config.buffer_id);
+    {
```
yuriks: Seems unnecessary to add an entire indent and scope just to limit the scope a simple integer variable. I'd prefer not doing that

jmerdich:
Addressed your comments in the latest patch set, @yuriks. If there's anything else, let me know.

src/audio_core/hle/source.cpp
```diff
@@ -305,10 +324,13 @@ bool Source::DequeueBuffer() {
         break;
     }
 
-    state.current_sample_number = 0;
-    state.next_sample_number = 0;
+    // the first playthrough starts at play_position, loops start at the beginning of the buffer
+    state.current_sample_number = (!buf.has_played) ? buf.play_position : 0;
+    state.next_sample_number = state.current_sample_number;
+    state.buffer_update = buf.from_queue && !buf.has_played;
```
yuriks: Oh, what I mean with that comment was that `state.buffer_update` was below `state.current_buffer_id` before, but your patch moves it above it. It just makes the diffs a bit weirder to follow.

jmerdich: Ah. There was once a valid reason, but no longer. Fixing...

jmerdich:
Sorry. Typo on the command line.

yuriks:
And me re-opening this seems to have done something funny. Probably should re-push on your end, sorry.

jmerdich:
Should be fixed now. That was weird.

jmerdich:
Likewise, and please do.

PR #2423, opened 2017-01-11
SDL2: Config, fix double to float warning

Kloen:


baka0815:
As I commented in the original PR:

It's integer scaling only now, but the value is stored as a floating point value - why?

In ``FromResolutionFactor()`` the check is only on an equals basis, so if the config has values like 1.5 you will  get 0 (auto scaled) instead of 1.0 or 2.0 which would be nearer to the setting.
But as I stated above: why allow floating point values if the scaling is integer only?

bunnei:
@baka0815 because the core emulator supports scaling the IR by a float. E.g. 0.25 or 1.5 in the config setting are perfectly valid and will do the right thing - this is entirely decoupled from the Qt frontend.

With that, the "cast to int" you are referring to only happens to convert the value in the config file to a selected index on the Qt drop down box. If you don't use the config UI, you can still use a floating point scale factor with the Qt frontend by setting the value manually.

baka0815:
@bunnei thanks for the classification, I didn't see that the scaling really uses float values. Nontheless, shouldn't the float values be rounded so opening and closing the GUI would leave the settings at least near the configured values (1.5 -> 2.0) instead of defaulting to 0 0f (auto)?

bunnei:
@baka0815 not sure - I didn't really consider it much. Afterall, if you're setting the config manually, its because you _don't_ want one of the predefined settings. But if someone wants to change the behavior, I don't really care much.

PR #2428, opened 2017-01-13
GSP_GPU::StoreDataCache stubbed

mailwl:
This avoid endless spam of "unimplemented StoreDataCache..." in some games

### DIFF
--- a/src/core/hle/service/gsp_gpu.cpp
+++ b/src/core/hle/service/gsp_gpu.cpp
@@ -705,6 +705,33 @@ static void ReleaseRight(Interface* self) {
     LOG_WARNING(Service_GSP, "called");
 }
 
+/**
+ * GSP_GPU::StoreDataCache service function
+ *
+ * This Function is a no-op, We aren't emulating the CPU cache any time soon.
+ *
+ *  Inputs:
+ *      0 : Header code [0x001F0082]
+ *      1 : Address
+ *      2 : Size
+ *      3 : Value 0, some descriptor for the KProcess Handle
+ *      4 : KProcess handle
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void StoreDataCache(Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 address = cmd_buff[1];
+    u32 size = cmd_buff[2];
+    u32 process = cmd_buff[4];
+
+    cmd_buff[0] = IPC::MakeHeader(0x1F, 0x1, 0);
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+
+    LOG_DEBUG(Service_GSP, "(STUBBED) called address=0x%08X, size=0x%08X, process=0x%08X", address,
+              size, process);
+}
+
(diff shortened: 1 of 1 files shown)


src/core/hle/service/gsp_gpu.cpp
```diff
@@ -705,6 +705,33 @@ static void ReleaseRight(Interface* self) {
...
+static void StoreDataCache(Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+    u32 address = cmd_buff[1];
+    u32 size = cmd_buff[2];
+    u32 process = cmd_buff[4];
+
+    cmd_buff[0] = IPC::MakeHeader(0x1F, 0x1, 0);
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+
+    LOG_DEBUG(Service_GSP, "(STUBBED) called address=0x%08X, size=0x%08X, process=0x%08X", address,
```
bunnei: Stubbed functions are supposed to be WARNING, but then I guess it still spams unless we change the default log level :P

yuriks:
Tbh I'd even be fine wuth removing the STUBBED tag and converting the
log to TRACE. This function is now implemented, it's just a no-op in our
implementation.


--yuriks





On Fri, Jan 13, 2017, at 09:04 AM, bunnei wrote:

> *@bunnei* commented on this pull request.



> 

> In src/core/hle/service/gsp_gpu.cpp[1]:


> > + *      2 : Size + *      3 : Value 0, some descriptor for the
> >   KProcess Handle + *      4 : KProcess handle + *  Outputs: + *
> >   1 : Result of function, 0 on success, otherwise error code + */
> >   +static void StoreDataCache(Interface* self) { +    u32* cmd_buff
> >   = Kernel::GetCommandBuffer(); +    u32 address = cmd_buff[1]; +
> >   u32 size = cmd_buff[2]; +    u32 process = cmd_buff[4]; + +
> >   cmd_buff[0] = IPC::MakeHeader(0x1F, 0x1, 0); +    cmd_buff[1] =
> >   RESULT_SUCCESS.raw; // No error + +    LOG_DEBUG(Service_GSP,
> >   "(STUBBED) called address=0x%08X, size=0x%08X, process=0x%08X",
> >   address,
>
> Stubbed functions are supposed to be WARNING, but then I guess it
> still spams unless we change the default log level :P
> — You are receiving this because you commented. Reply to this email
> directly, view it on GitHub[2], or mute the thread[3].
> 


Links:

  1. https://github.com/citra-emu/citra/pull/2428#pullrequestreview-16606779
  2. https://github.com/citra-emu/citra/pull/2428#pullrequestreview-16606779

src/core/hle/service/gsp_gpu.cpp
yuriks: TBH I'd even be fine with removing the STUBBED tag and converting the
log to TRACE. This function is now implemented, it's just a no-op in our
implementation.

mailwl: all these cache functions additionally checks address ranges, then call appropriate SVC call.
we may fully implement service functions, but stub SVC calls in other PR. nothing changes, but code will be more clean

PR #2429, opened 2017-01-13
CFG: move language override to the boot process

wwylele:
Previously in #2240 I added a language overriding in `GetConfigInfoBlock`, which happens when the region setting is auto. This causes a small problem because `GetConfigInfoBlock` is also used by the Qt frontend to checkout the current settings. The symptom is that the configuration dialog will always shows "Japanese" as language if the region is auto, in spite of the actual language settings. ~~Now I moved this overriding into the service functions, so that it only happens when the game reads it.~~ Now I moved this overriding to the boot process, so that it won't happens when users edit the system settings. It will still affect if users open the configuration window while a game is running, but it doesn't matter and actually tells users that the language is changed.

### DIFF
--- a/src/core/hle/service/cfg/cfg.cpp
+++ b/src/core/hle/service/cfg/cfg.cpp
@@ -322,46 +322,10 @@ static ResultVal<void*> GetConfigInfoBlockPointer(u32 block_id, u32 size, u32 fl
     return MakeResult<void*>(pointer);
 }
 
-/// Checks if the language is available in the chosen region, and returns a proper one
-static u8 AdjustLanguageInfoBlock(u32 region, u8 language) {
-    static const std::array<std::vector<u8>, 7> region_languages{{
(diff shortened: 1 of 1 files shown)


yuriks:
Hm, rather than keep doing this, I think it'd be cleaner to have the boot process set the language if automatic region detection is enabled. That way the functions don't need to be modified at all. Are there any problems with that approach?

wwylele:
Did what @yuriks suggested.

yuriks:
Doesn't the current version always override the language, even if auto-region isn't enabled? I think that we should probably warn about this in the UI, but should let the user do it anyway, in case someone wants to experiment or find unused content in games, etc.

wwylele:
>Doesn't the current version always override the language, even if auto-region isn't enabled?

My fault. I forgot to wrap it with the condition. Fixed

PR #2439, opened 2017-01-14
InputCore Overhaul

### LINKED ISSUE #1883: Frontend <-> Backend communication is frought with race conditions

makotech222:
This is second attempt at InputCore PR. Previous #1995 

This PR consolidates all input handling during games into Input_Core. It also adds support for SDL Gamepads to work with Citra. It also partially fixes #1883, relating to thread safety on keyboard inputs.

Since first PR, following changes have been made:

1. Circle pad Deadzone setting added to config files. Defaults to 0.3. No GUI support for editing yet.
2. Cleanup of Input GUI code.
3. GPU Tick rate moved to be static constexpr value, and now referenced in Input Core
4. Circle Pad Modifier Key now works when bound to any input device, not just keyboard.
5. An input device threshold value of 0.45 is used when binding analog sticks to citra button presses. Mimics Dolphin.

After this PR is accepted, the following changes will eventually need to be made:
- Redesign input gui to allow setting deadzone and setting hotkeys during game
- Add support for hotkeys
- Add support for Xinput

Note: There is a report of 360 game pad not working properly on Linux. I personally don't have access to a Linux device to test, so please let me know if you run into this issue. I theorize it is due to either the Linux drivers, or to the gamecontrollerdb file. Need more testing to conclude.


src/citra/emu_window/emu_window_sdl2.cpp
```diff
@@ -8,13 +8,17 @@
...
 #include "common/scm_rev.h"
 #include "common/string_util.h"
-#include "core/frontend/key_map.h"
 #include "core/hle/service/hid/hid.h"
 #include "core/settings.h"
+
+#include "input_core/devices/keyboard.h"
+#include "input_core/input_core.h"
+
+#include "citra/emu_window/emu_window_sdl2.h"
```
Kloen: ??? Do these import order even pass clang-format? 😮

src/citra_qt/config.h
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright 2014 Citra Emulator Project
+// Copyright 2014 Citra Emulator Project
```
Kloen: Dafuq with the diff?

src/citra_qt/configure_input.cpp
```diff
@@ -1,146 +1,168 @@
-﻿// Copyright 2016 Citra Emulator Project
+// Copyright 2016 Citra Emulator Project
```
Kloen: Dafuq with the diff v2?

src/citra_qt/keybinding_names.cpp
```diff
@@ -0,0 +1,31 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: It's 2017 already!

src/citra_qt/keybinding_names.h
```diff
@@ -0,0 +1,19 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: Ditto.

src/core/frontend/emu_window.h
```diff
@@ -301,11 +241,7 @@ class EmuWindow {
 
     bool touch_pressed; ///< True if touchpad area is currently pressed, otherwise false
 
-    u16 touch_x; ///< Touchpad X-position in native 3DS pixel coordinates (0-320)
-    u16 touch_y; ///< Touchpad Y-position in native 3DS pixel coordinates (0-240)
-
-    s16 circle_pad_x; ///< Circle pad X-position in native 3DS pixel coordinates (-156 - 156)
-    s16 circle_pad_y; ///< Circle pad Y-position in native 3DS pixel coordinates (-156 - 156)
+
```
Kloen: Unnecesary blank lines here and there.

src/input_core/devices/device.cpp
```diff
@@ -0,0 +1,7 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/devices/device.h
```diff
@@ -0,0 +1,49 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/devices/keyboard.cpp
```diff
@@ -0,0 +1,60 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/devices/keyboard.h
```diff
@@ -0,0 +1,49 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,122 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/devices/sdl_gamepad.h
```diff
@@ -0,0 +1,113 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,262 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/input_core.h
```diff
@@ -0,0 +1,125 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/key_map.cpp
```diff
@@ -0,0 +1,27 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/key_map.h
```diff
@@ -0,0 +1,15 @@
+// Copyright 2016 Citra Emulator Project
```
Kloen: 2017

src/input_core/input_core.h
```diff
@@ -0,0 +1,125 @@
...
+
+#include "core/hle/service/hid/hid.h"
+#include "core/settings.h"
+#include "input_core/devices/device.h"
+
+class Keyboard;
+
+class InputCore {
+public:
+    static void Init();
```
lioncash: This class shouldn't be entirely composed of static state.

legoj15:
I decided to do some testing on Linux (Ubuntu 16.10) with a wired Xbox 1 controller. All buttons were mapped properly, but I had some issues with the analogue sticks and triggers

vokoscreen-2017-01-14_19-15-50_1
After binding the left trigger, I go to bind the right, but it immediately binds right to left, as if I was still holding down the left. I eventually got right trigger bound to right bumper, and left to left. This persists with the left and right stick too.
If I had to guess, the current implementation is too "sensitive" to inputs, so it will bind anything that is slightly pressed.
Can't seem to reproduce the issue on Windows 10, so maybe it's a Linux driver issue(?)

legoj15:
Did some testing with a Xbox 360 wireless controller on Ubuntu, and had no issues at all, so maybe it's a Xinput driver issue with the Xbox 1 controller?

src/input_core/devices/keyboard.cpp
```diff
@@ -0,0 +1,60 @@
...
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include <SDL_keyboard.h>
+
+#include "input_core/devices/keyboard.h"
+
+Keyboard::Keyboard() {}
+
+Keyboard::~Keyboard() {}
```
lioncash: `Keyboard::~Keyboard() = default;`

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include "common/assert.h"
+#include "common/logging/log.h"
+#include "common/string_util.h"
+
+#include "input_core/devices/gamecontrollerdb.h"
+#include "input_core/devices/sdl_gamepad.h"
+
+bool SDLGamepad::SDLInitialized = false;
+
+SDLGamepad::SDLGamepad() {}
```
lioncash: `SDLGamepad::SDLGamepad() = default;`

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,122 @@
...
+        SDL_GameControllerAddMapping(s.c_str());
+    }
+}
+
+Settings::InputDeviceMapping SDLGamepad::GetInput() {
+    if (gamepad == nullptr)
+        return Settings::InputDeviceMapping("");
+
+    auto results = ProcessInput();
+    for (auto& input : results) {
```
lioncash: This can likely be `const auto&`

lioncash: This can likely be `const auto&`

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,122 @@
...
+void SDLGamepad::LoadGameControllerDB() {
+    std::vector<std::string> lines1, lines2, lines3, lines4;
+    Common::SplitString(SDLGameControllerDB::db_file1, '\n', lines1);
+    Common::SplitString(SDLGameControllerDB::db_file2, '\n', lines2);
+    Common::SplitString(SDLGameControllerDB::db_file3, '\n', lines3);
+    Common::SplitString(SDLGameControllerDB::db_file4, '\n', lines4);
+    lines1.insert(lines1.end(), lines2.begin(), lines2.end());
+    lines1.insert(lines1.end(), lines3.begin(), lines3.end());
+    lines1.insert(lines1.end(), lines4.begin(), lines4.end());
+    for (std::string s : lines1) {
```
lioncash: This should be a const reference `const std::string&` otherwise this makes unnecessary copies.

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,122 @@
...
+        bool success = gamepad->InitDevice(
+            i, Settings::InputDeviceMapping("SDL/" + std::to_string(i) + "/Gamepad/-1"));
+        if (success)
+            devices.push_back(gamepad);
+    }
+    return devices;
+}
+
+void SDLGamepad::LoadGameControllerDB() {
+    std::vector<std::string> lines1, lines2, lines3, lines4;
```
lioncash: These are extremely vague variable names.

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,262 @@
...
+std::vector<std::shared_ptr<IDevice>> InputCore::GetAllDevices() {
+    auto all_devices = SDLGamepad::GetAllDevices();
+    auto keyboard = InputCore::GetKeyboard();
+    all_devices.push_back(keyboard);
+
+    return all_devices;
+}
+
+Settings::InputDeviceMapping InputCore::DetectInput(int max_time,
+                                                    std::function<void(void)> update_GUI) {
```
lioncash: `update_gui`

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,262 @@
...
+    std::lock_guard<std::mutex> lock(touch_mutex);
+    return std::make_tuple(touch_x, touch_y, touch_pressed);
+}
+
+void InputCore::SetTouchState(std::tuple<u16, u16, bool> value) {
+    std::lock_guard<std::mutex> lock(touch_mutex);
+    std::tie(touch_x, touch_y, touch_pressed) = value;
+}
+
+bool InputCore::CheckIfMappingExists(const std::vector<Settings::InputDeviceMapping>& uniqueMapping,
```
lioncash: `unique_mapping`

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,262 @@
...
+                                     Settings::InputDeviceMapping mappingToCheck) {
```
lioncash: `mapping_to_check`

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,262 @@
...
+    return std::any_of(uniqueMapping.begin(), uniqueMapping.end(),
+                       [mappingToCheck](const auto& mapping) { return mapping == mappingToCheck; });
+}
+
+std::vector<Settings::InputDeviceMapping> InputCore::GatherUniqueMappings() {
+    std::vector<Settings::InputDeviceMapping> uniqueMappings;
```
lioncash: `unique_mappings`

Note that if these are unique in the sense that only one of each mapping is allowed to exist, then a `std::set` or `std::unordered_set` is more applicable.

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,262 @@
...
+    // Reschedule recurrent event
+    CoreTiming::ScheduleEvent(GPU::frame_ticks - cycles_late, tick_event);
+}
+
+void InputCore::UpdateEmulatorInputs(
+    std::vector<std::map<Settings::InputDeviceMapping, float>> inputs) {
+    std::lock_guard<std::mutex> lock(pad_state_mutex);
+
+    // Apply deadzone for circle pad
+    float leftx = 0, lefty = 0;
```
lioncash: `left_x` and `left_y`

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,122 @@
...
+    for (int i = 0; i < SDL_GameControllerButton::SDL_CONTROLLER_BUTTON_MAX; i++) {
+        SDL_GameControllerButton button = static_cast<SDL_GameControllerButton>(i);
+        Uint8 pressed = SDL_GameControllerGetButton(gamepad, button);
+        input_device_mapping.key = static_cast<int>(
+            gamepadinput_to_sdlname_mapping2[SDL_GameControllerGetStringForButton(button)]);
+        button_status.emplace(input_device_mapping, pressed);
+    }
+    for (int i = 0; i < SDL_GameControllerAxis::SDL_CONTROLLER_AXIS_MAX; i++) {
+        SDL_GameControllerAxis axis = static_cast<SDL_GameControllerAxis>(i);
+        float strength = fmaxf(-1, (float)SDL_GameControllerGetAxis(gamepad, axis) / 32767.0);
```
lioncash: `static_cast`

src/input_core/devices/keyboard.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+std::map<Settings::InputDeviceMapping, float> Keyboard::ProcessInput() {
+    std::map<KeyboardKey, bool> keysPressedCopy;
+    {
+        std::lock_guard<std::mutex> lock(m);
+        keysPressedCopy = keys_pressed;
+    }
+    std::map<Settings::InputDeviceMapping, float> button_status;
+    bool circlePadModPressed = keysPressedCopy[circle_pad_modifier];
```
lioncash: `circle_pad_mod_pressed`

src/input_core/devices/keyboard.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+Keyboard::~Keyboard() {}
+
+bool Keyboard::InitDevice(int number, Settings::InputDeviceMapping device_mapping) {
+    input_device_mapping = device_mapping;
+    return true;
+}
+
+std::map<Settings::InputDeviceMapping, float> Keyboard::ProcessInput() {
+    std::map<KeyboardKey, bool> keysPressedCopy;
```
lioncash: `keys_pressed_copy`

src/core/hle/service/hid/hid.cpp
```diff
@@ -3,9 +3,16 @@
 // Refer to the license.txt file included.
 
 #include <cmath>
-#include "common/logging/log.h"
-#include "core/core_timing.h"
 #include "core/frontend/emu_window.h"
+
+#include "core/hle/service/hid/hid.h"
```
lioncash: These includes are already present below.

src/core/hle/service/hid/hid.cpp
```diff
@@ -3,9 +3,16 @@
...
+#include "core/hle/service/hid/hid_spvr.h"
+#include "core/hle/service/hid/hid_user.h"
+#include "core/hle/service/service.h"
+
+#include "input_core/input_core.h"
```
lioncash: This should be above the `video_core/video_core.h` include.

src/citra_qt/configure_input.h
```diff
@@ -30,28 +30,33 @@ class ConfigureInput : public QWidget {
...
+    QPushButton* changing_button = nullptr; ///< button currently waiting for key press.
+    QString previous_mapping;
 
-    /// Load configuration settings.
+    /// Load configuration settings into button text
     void loadConfiguration();
+
+    /// Check all inputs for duplicate keys. Clears out any other button with the same value as this
+    /// button's new value.
+    void removeDuplicates(const Settings::InputDeviceMapping newValue);
```
lioncash: `new_value`

src/citra_qt/configure_input.h
```diff
@@ -30,28 +30,33 @@ class ConfigureInput : public QWidget {
...
+
+    /// Handle keykoard key press event for input tab when a button is 'waiting'.
+    void keyPressEvent(QKeyEvent* event) override;
+
+    /// Convert key ASCII value to its' letter/name
+    static QString getKeyName(Settings::InputDeviceMapping mapping);
+
+    /// Set button text to name of key pressed.
+    void setKey(Settings::InputDeviceMapping keyPressed);
```
lioncash: `key_pressed`

src/citra_qt/configure_input.cpp
```diff
@@ -1,146 +1,168 @@
...
-void ConfigureInput::keyPressEvent(QKeyEvent* event) {
-    releaseKeyboard();
-    releaseMouse();
-
-    if (!current_input_id || !event)
-        return;
-
-    if (event->key() != Qt::Key_Escape)
-        setInput(*current_input_id, static_cast<Qt::Key>(event->key()));
+void ConfigureInput::removeDuplicates(const Settings::InputDeviceMapping newValue) {
```
lioncash: `new_value`

src/citra_qt/configure_input.cpp
```diff
@@ -1,146 +1,168 @@
...
     }
+    key_map[ui->buttonCircleMod] = Settings::values.pad_circle_modifier;
     updateButtonLabels();
 }
 
-void ConfigureInput::restoreDefaults() {
-    for (const auto& input_id : Settings::NativeInput::All) {
-        const size_t index = static_cast<size_t>(input_id);
-        key_map[input_id] = static_cast<Qt::Key>(Config::defaults[index].toInt());
+void ConfigureInput::setKey(Settings::InputDeviceMapping keyPressed) {
```
lioncash: `key_pressed`

src/input_core/input_core.h
makotech222: Its supposed to be a static class. What do you suggest?

src/core/frontend/emu_window.cpp
```diff
@@ -7,32 +7,11 @@
...
-    if (r > 1) {
-        r = std::sqrt(r);
-        x /= r;
-        y /= r;
-    }
-
-    circle_pad_x = static_cast<s16>(x * MAX_CIRCLEPAD_POS);
-    circle_pad_y = static_cast<s16>(y * MAX_CIRCLEPAD_POS);
-}
+#include "emu_window.h"
```
makotech222: Lot of these must've been from rebasing. Will fix

src/citra/emu_window/emu_window_sdl2.cpp
makotech222: My clang formatter didn't have a problem with it. Weird.

src/input_core/devices/keyboard.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+void Keyboard::Clear() {
+    std::lock_guard<std::mutex> lock(m);
+    keys_pressed.clear();
+}
+
+Settings::InputDeviceMapping Keyboard::GetInput() {
+    auto result = ProcessInput();
+    for (const auto& entry : result) {
+        if (entry.second > 0.5)
```
makotech222: Need to replace this with a constant variable.

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+}
+
+Settings::InputDeviceMapping SDLGamepad::GetInput() {
+    if (gamepad == nullptr)
+        return Settings::InputDeviceMapping("");
+
+    auto results = ProcessInput();
+    for (const auto& input : results) {
+        if (input.second > 0.5)
```
makotech222: constant variable here too

makotech222:
@legoj15 Interesting. Can you explain how you eventually got it to work? Right now the threshold for being accepted as input is 50% of maximum of input, so it shouldn't be too sensitive really.

legoj15:
@makotech222 I simply re-selected the button I had just bound

makotech222:
Yeah that's pretty interesting. Can you try with the latest changes?

legoj15:
Still doesn't seem to be working, unless I'm doing something wrong :V

makotech222:
Hmm, can you try bind inputs on the dolphin emulator? I wonder if you have problems with that emulator.

src/input_core/devices/keyboard.h
```diff
@@ -0,0 +1,49 @@
...
+        return key == other.key;
+    }
+    bool operator==(uint32_t other) const {
+        return key == other;
+    }
+    bool operator<(const KeyboardKey& other) const {
+        return key < other.key;
+    }
+};
+class Keyboard : public IDevice {
```
wwylele: I thought this have been addressed in the original PR... `IDevice` is not our nameing style (nor most c++ style). You can use something like `InputDeviceInterface`

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,263 @@
...
+u16 InputCore::touch_x;        ///< Touchpad X-position in native 3DS pixel coordinates (0-320)
+u16 InputCore::touch_y;        ///< Touchpad Y-position in native 3DS pixel coordinates (0-240)
+bool InputCore::touch_pressed; ///< True if touchpad area is currently pressed, otherwise false
+const float InputCore::input_detect_threshold =
+    0.45; ///< Applies to analog controls being used for digital 3ds inputs.
+
+void InputCore::Init() {
+    ParseSettings();
+    tick_event = CoreTiming::RegisterEvent("InputCore::tick_event", InputTickCallback);
+    CoreTiming::ScheduleEvent(GPU::frame_ticks, tick_event);
```
wwylele: Using `CoreTiming` looks bad to me. Input core should be a pure frontend thing and ideally it doesn't depend on core at all. `CoreTiming` is for the emulated virtual time line, only for scheduling emulated software/hardware event. It's pace is not uniform with the real word time. On the other hand, a frontend thing shouldn't intrude the core and let the core to process the frontend event. Input core should manage its own timing method

src/input_core/input_core.h
wwylele: @makotech222 turn the class in to a namespace OR a singleton

lioncash: A class being solely used as a container static behavior is a design issue. At least when it comes to C++. This isn't like C# or Java, or languages that disallow functions to be defined outside a class. If you ever need that sort of thing, a namespace is much more suitable.

However, in this case, what I'd suggest is making all the behavior of the input core non-static where applicable, and making the [System](https://github.com/citra-emu/citra/blob/master/src/core/core.h#L22) class contain a private instance of it. and expose getter functions.

src/input_core/input_core.cpp
makotech222: Input processing was originally inside the GPU tick event, i just pulled it out into its own core. I personally think it belongs there. Would like other opinions before changing.

makotech222:
I think given the problems that SDL_Gamecontroller has on Linux, I am going to switch to using SDL_Joypad, which is what dolphin uses. I think i will leave PR open to receive more feedback, and push changes when ready.

src/input_core/input_core.h
```diff
@@ -0,0 +1,78 @@
...
+#include <mutex>
+#include <set>
+#include <tuple>
+#include "core/hle/service/hid/hid.h"
+#include "core/settings.h"
+#include "input_core/devices/device.h"
+
+class Keyboard;
+
+namespace InputCore {
```
lioncash: See the comment [here](https://github.com/citra-emu/citra/pull/2439#discussion_r96132311). 

The input subsystem should interface through `System` (i.e. the input core would be a class which System contains an instance of). The way this currently is makes the entirety of the input subsystem single state. If you hypothetically wanted to spin up two system instances in the future for some form of testing, it wouldn't be possible with this.

makotech222: Do any other 'Cores' do this? seems like all others are namespaces too. Feels like a case of YAGNI imo.

makotech222:
@Subv Apologies for the single commit. I didn't know peoples preferences for how commits are done. How do i break up a commit?

src/input_core/key_map.h
```diff
@@ -0,0 +1,14 @@
...
+
+#pragma once
+
+#include <array>
+#include "core/hle/service/hid/hid.h"
+
+namespace KeyMap {
+extern const std::array<Service::HID::PadState, Settings::NativeInput::NUM_INPUTS> mapping_targets;
+extern const std::array<Service::HID::PadState, 4> analog_inputs;
+constexpr int MAX_CIRCLEPAD_POS = 0x9C; /// Max value for a circle pad position
```
j-selby: Says who? 3DS?

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+/**
+* Loops through all unique input devices, and all bound inputs to update the emulator's input
+* status.
+*/
+void UpdateEmulatorInputs(std::vector<std::map<Settings::InputDeviceMapping, float>> inputs) {
+    std::lock_guard<std::mutex> lock(pad_state_mutex);
+
+    // Apply deadzone for circle pad
```
j-selby: But, you aren't applying the deadzone. You are reading  values from your input...

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,272 @@
...
+std::mutex pad_state_mutex;
+std::mutex touch_mutex;
+u16 touch_x;        ///< Touchpad X-position in native 3DS pixel coordinates (0-320)
+u16 touch_y;        ///< Touchpad Y-position in native 3DS pixel coordinates (0-240)
+bool touch_pressed; ///< True if touchpad area is currently pressed, otherwise false
+const float input_detect_threshold =
+    0.45; ///< Applies to analog controls being used for digital 3ds inputs.
+
+/**
+* Takes two floats and the deadzone and applies formula to
```
j-selby: "applies formula" is not a very useful description.

src/input_core/devices/sdl_gamepad.cpp
```diff
@@ -0,0 +1,122 @@
...
+            gamepadinput_to_sdlname_mapping[SDL_GameControllerGetStringForButton(button)]);
+        button_status.emplace(input_device_mapping, pressed);
+    }
+    for (int i = 0; i < SDL_GameControllerAxis::SDL_CONTROLLER_AXIS_MAX; i++) {
+        SDL_GameControllerAxis axis = static_cast<SDL_GameControllerAxis>(i);
+        float strength =
+            fmaxf(-1, static_cast<float>(SDL_GameControllerGetAxis(gamepad, axis) / 32767.0));
+        input_device_mapping.key = static_cast<int>(
+            gamepadinput_to_sdlname_mapping[SDL_GameControllerGetStringForAxis(axis)]);
+        if (strength < 0 && i < 4) {
```
j-selby: What the heck is going on here? What does `i < 4` mean, in the context of SDL controllers (relating to analog axis I would presume)? At the very least, comment this completely.

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,272 @@
...
+#include <chrono>
+#include <cmath>
+#include <map>
+#include "core/core_timing.h"
+#include "core/hw/gpu.h"
+#include "input_core/devices/keyboard.h"
+#include "input_core/devices/sdl_gamepad.h"
+#include "input_core/input_core.h"
+
+int tick_event;
```
j-selby: All of these variables tucked together in this big block isn't very useful. At the very least, comment it, or better yet, sort them out into groups or something.

makotech222:
@j-selby I always run my clang formatter on all my files. I'm using the Manobit extension on VS2015. Do you recommend another?

j-selby:
@makotech222 I use CLion, so I am not sure what the Visual Studio selection is. I normally just use clang-format on the command line, but generally plugins which just act as a frontend are fine - e.g https://plugins.jetbrains.com/clion/plugin/8396-clangformatij

Regardless, newlines would do this codebase wonders (which is what I was hoping clang-format would do).

src/input_core/input_core.cpp
```diff
@@ -0,0 +1,272 @@
...
+    // Apply deadzone for circle pad
+    float left_x = 0, left_y = 0;
+    float circle_pad_modifier = 1.0;
+    auto circle_pad_modifier_mapping = Settings::values.pad_circle_modifier;
+    for (auto& input_device : inputs) {
+        for (auto& button_states : input_device) { // Loop through all buttons from input device
+            float strength = button_states.second;
+            auto emulator_inputs = key_mappings[button_states.first];
+            for (auto& emulator_input : emulator_inputs) {
+                if (emulator_input == Service::HID::PAD_CIRCLE_UP && abs(strength) > 0) {
```
j-selby: You do not need these absolute function calls here - it appears to be converting strength to a integer, and you already are doing `abs` in the input handlers themselves (SDL).

This fixes the non-floating-point circle pad issue I described above.

src/citra_qt/CMakeLists.txt
```diff
@@ -27,8 +27,10 @@ set(SRCS
             configure_graphics.cpp
             configure_system.cpp
             configure_input.cpp
+            configure_system.cpp
```
mailwl: this includes twice (see line 28)

src/citra_qt/CMakeLists.txt
```diff
@@ -61,9 +63,11 @@ set(HEADERS
             configure_graphics.h
             configure_system.h
             configure_input.h
+            configure_system.h
```
mailwl: include twice (line 64)

src/citra_qt/CMakeLists.txt
```diff
@@ -80,6 +84,7 @@ set(UIS
             configure_graphics.ui
             configure_system.ui
             configure_input.ui
+            configure_system.ui
```
mailwl: include twice (line 85)

src/input_core/input_core.h
lioncash: Just because another core does a certain thing doesn't mean it's a good idea to do the same thing.

The `System` class is intended to act as a central point of unification for major behaviors that a hypothetical 3DS system would need in order to function. The only reason it's not housing the rest of the 'cores' yet is because they haven't been converted from namespaces yet.

Ideally the end goal is that you can just declare a new `System` variable instance to spin up a new instance of 3DS emulation.

makotech222: Okay sounds reasonable enough. Will implement

makotech222:
Okay got a lot of input on the PR. I'm going to close this due to number of issues that need to be fixed. Before i make another PR attempt, I will fix the following:
1. Break commits into smaller pieces.
2. Turn inputcore into a class again and have System hold a ref
3. Replace SDL_Gamecontroller with SDL_Joystick to fix linux issues
4. Add more comments to clarify some code
5. Other Misc fixes.

Thanks everyone for the help with reviewing.

src/citra_qt/configure_input.cpp
```diff
@@ -1,146 +1,168 @@
-﻿// Copyright 2016 Citra Emulator Project
+// Copyright 2016 Citra Emulator Project
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include <memory>
 #include <utility>
-#include <QTimer>
+
```
makotech222: @wwylele This contradicts what is in the style guide for citra. Specifically says

// each major module is separated

https://github.com/citra-emu/citra/blob/master/CONTRIBUTING.md#reporting-issues

wwylele: @makotech222 sorry, that one should be updated since #2086.

>(clang-format include sorting only really works properly on #include blocks without spaces in between afaik)

makotech222: Got it. Thanks!

PR #2441, opened 2017-01-15
Gui: Change title bar to include build name

jroweboy:
A highly requested feature from the citra community moderators. This will hopefully give an "at-a-glance" look at what build each person asking for help is using, so the moderators don't have to search for commits to see if the person they are helping are actually using citra nightly builds. 

Nightly builds now have "Citra nightly" in the titlebar
Bleeding edge builds now have "Citra bleeding edge" in the titlebar
Uploads from CI on the master repo have "Citra master" in the titlebar
Any other build just has "Citra unofficial" in the titlebar

### DIFF
--- a/src/citra/emu_window/emu_window_sdl2.cpp
+++ b/src/citra/emu_window/emu_window_sdl2.cpp
@@ -79,8 +79,8 @@ EmuWindow_SDL2::EmuWindow_SDL2() {
     SDL_GL_SetAttribute(SDL_GL_BLUE_SIZE, 8);
     SDL_GL_SetAttribute(SDL_GL_ALPHA_SIZE, 0);
 
-    std::string window_title =
-        Common::StringFromFormat("Citra | %s-%s", Common::g_scm_branch, Common::g_scm_desc);
+    std::string window_title = Common::StringFromFormat("Citra %s| %s-%s ", Common::g_build_name,
+                                                        Common::g_scm_branch, Common::g_scm_desc);
     render_window = SDL_CreateWindow(
         window_title.c_str(),
         SDL_WINDOWPOS_UNDEFINED, // x position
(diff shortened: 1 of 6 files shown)


bunnei:
Not sure I like the "master" and "unofficial" in the title, but "nightly" and "bleeding edge" are fine for those releases. What do we really gain from the former (other than additional UI cruft) if we only official support nightlies? (ie. regardless of whether it says, "master", "unofficial", or nothing, we're still not going to assist people with random builds)... Furthermore, we also get git commit (which will tell us if it's in master) and the "dirty" tag (which also is an indicator of a build we can ignore).

jroweboy:
> What do we really gain from the former (other than additional UI cruft) if we only official support nightlies?

It makes it easier to distinguish between builds that were made before the commit and builds after since if its not there, then we can easily tell that a build was made before this commit. Otherwise is a guessing game to see if the help request is using a older nightly build or some rando speed h4x build they found online. That said, I'm not attached to having them, so I don't mind removing them.

>  Furthermore, we also get git commit (which will tell us if it's in master) and the "dirty" tag (which also is an indicator of a build we can ignore).

We shouldn't be expect people in the community to understand how to look up the hashes. There are plenty of people that help out on the forums without understanding what a hash is. And dirty only says if the current working directory is dirty, which is hardly ever the case for any of the builds people will ask for help with.

chris062689:
LGTM

MerryMage:
Have any decisions been made regarding this PR?

yuriks:
I'm confused about what the labels actually mean. What's the difference between master and nightly? *Isn't* nightly master?

I would be more in favor of including the branch name in the title, or replacing it with nightly/BE if it came from an official CI build and that's it. The "unofficial" tag gives this presumption of authority which I don't really like.

jroweboy:
Updated the PR according to the feedback. Master/unofficial have been removed entirely. Nightly and Bleeding Edge are also now properly capitalized in the title bar.

PR #2442, opened 2017-01-15
HID: manages updating itself using correct ticks

wwylele:
We should correct this before #2439.

Moved HID updating event from GPU to its own module. Separated three devices updating. Corrected the updating ticks using empirical values.

The ticks values are got from [this test](https://github.com/wwylele/ctrhwtest/tree/master/hid_time_test). Usage: press A to test the updating frequency for each HID event; press X/Y to enable/disable gyroscope/accelerometer
Test result (on a JAP old 3DS):
 - PAD0/1 event updates ~235 times per second with accelerometer disabled, or ~233 times per second with accelerometer enabled.
 - Accelerometer updates ~104 times per second.
 - Gyroscope updates ~101 times per second.

Fixes touch pad not working issue in some games (tested Nintendo Video and Nintendo 3DS Sound)

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -35,6 +35,15 @@ static u32 next_gyroscope_index;
 static int enable_accelerometer_count = 0; // positive means enabled
 static int enable_gyroscope_count = 0;     // positive means enabled
 
+static int pad_update_event;
+static int accelerometer_update_event;
+static int gyroscope_update_event;
+
+// Updating period for each HID device. These empirical values are measured from a 11.2 3DS.
+constexpr u64 pad_update_ticks = 268123480ull / 234;
+constexpr u64 accelerometer_update_ticks = 268123480ull / 104;
+constexpr u64 gyroscope_update_ticks = 268123480ull / 101;
+
 static PadState GetCirclePadDirectionState(s16 circle_pad_x, s16 circle_pad_y) {
     // 30 degree and 60 degree are angular thresholds for directions
     constexpr float TAN30 = 0.577350269f;
@@ -65,13 +74,8 @@ static PadState GetCirclePadDirectionState(s16 circle_pad_x, s16 circle_pad_y) {
     return state;
 }
 
-void Update() {
+static void UpdatePadCallback(u64 userdata, int cycles_late) {
     SharedMem* mem = reinterpret_cast<SharedMem*>(shared_mem->GetPointer());
-
-    if (mem == nullptr) {
-        LOG_DEBUG(Service_HID, "Cannot update HID prior to mapping shared memory!");
-        return;
-    }
 
     PadState state = VideoCore::g_emu_window->GetPadState();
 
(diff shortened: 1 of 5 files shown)


src/core/hle/service/hid/hid.cpp
```diff
@@ -35,6 +35,15 @@ static u32 next_gyroscope_index;
 static int enable_accelerometer_count = 0; // positive means enabled
 static int enable_gyroscope_count = 0;     // positive means enabled
 
+static int pad_update_event;
+static int accelerometer_update_event;
+static int gyroscope_update_event;
+
+// Updating period for each HID device. These value are empirical.
+constexpr u64 pad_update_ticks = 268123480ull / 234;
```
yuriks: Use a named constant for the clock rate instead of duplicating it. There's `g_clock_rate_arm11`, but that's actually a variable. I think for now you should add a constant with the same value, and use it to initialize that variable and replace other instances of this magic number in the codebase. (Do that in a separate commit.)

src/core/hle/service/hid/hid.cpp
```diff
@@ -65,7 +74,7 @@ static PadState GetCirclePadDirectionState(s16 circle_pad_x, s16 circle_pad_y) {
     return state;
 }
 
-void Update() {
+static void UpdatePadCallBack(u64 userdata, int cycles_late) {
```
yuriks: The "CallBack"s in these fuction names are strange. They turn the name into a noun, when long function names are usually verbs. If nothing else at least makes it `Callback` (the B isn't capitalized).

src/core/hle/service/hid/hid.cpp
```diff
@@ -131,7 +140,18 @@ void Update() {
...
-    // Update accelerometer
+    // Reschedule recurrent event
+    CoreTiming::ScheduleEvent(pad_update_ticks - cycles_late, pad_update_event);
+}
+
+static void UpdateAccelerometerCallBack(u64 userdata, int cycles_late) {
+    SharedMem* mem = reinterpret_cast<SharedMem*>(shared_mem->GetPointer());
+
+    if (mem == nullptr) {
+        LOG_DEBUG(Service_HID, "Cannot update HID prior to mapping shared memory!");
```
yuriks: I'm pretty sure this doesn't need to be a thing anymore with our new shared memory implementation: you can always just modify the backing memory of the object directly, whether it's mapped or not.

src/core/hle/service/hid/hid.cpp
```diff
@@ -35,6 +35,15 @@ static u32 next_gyroscope_index;
 static int enable_accelerometer_count = 0; // positive means enabled
 static int enable_gyroscope_count = 0;     // positive means enabled
 
+static int pad_update_event;
+static int accelerometer_update_event;
+static int gyroscope_update_event;
+
+// Updating period for each HID device. These value are empirical.
```
yuriks: Mention what they were measured from (old/new 3DS vX.XX)

wwylele:
Addressed comments above

yuriks:
@wwylele The idea was to add the commit with the constant *before* the HID commit that uses it... but w/e.

PR #2451, opened 2017-01-18
[RFC] Texture overhaul

FernandoS27:
Well it's finaly here. The texture decoding refactor. It's quite a big PR since a lot of copy pasted code was all over the place. There's still a lot of cleanup potential.

So Refactoring Texture Decoding, why?
* old texture decoding was fragmented all over the place since only the old function https://github.com/citra-emu/citra/blob/master/src/video_core/debug_utils/debug_utils.cpp#L318 did was decode color-texture formats.
* This function had 3 problems: first it did untiling, unswizzling and decoding all at once. Second, it only supported Texture formats. Third, it did decoding per texel which is highly inefficient.
* Since that function only decoded texture formats, many "user modules" were implementing depth formats decoding by hand and, in many cases, with a few hacky unmatainable code here and there. example: https://github.com/citra-emu/citra/blob/master/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp#L57
* The old decoder was using a lot of temporal heap memory which could be bounded in 99% of the cases. This brings consequences such as a lot of heap fragmentation.

This refactor has been written with the next things in mind:
* Highest possible efficiency when uploading/downloading textures from/to the GPU.
* Use static caches whenever it's possible.
* Simple interface that hides texture decoding from the user, through an abstract class called Codec.
* Centralized all code related to texture decoding whatsoever.
* More dynamism: user modules can manage how much decoding/encoding should be done. (This helps if the back-end can handle a given format natively).
* Write efficient understandable code without using intrinsics or hacks (ahem Duff's devices). Except compiler options and attributes.

Inmediate effects while running citra:
* Smoother gameplay due to less time uploading and downloading surfaces.
* Improved speed while playing games that download and upload the depth buffer every frame. (Example: Super Smash Bros 3DS).

EDIT: Conflicts with pull #2414 
EDIT2: this PR will be applied after #2414 , please review it first.


RavenHome1:
any chance of having this in BE?

FernandoS27:
@RavenHome1 it first needs to be reviewed, which can take quite sometime since the PR is quite big. Afterwards, I'll probably need to clean up some code and do changes. Once that's done it may be sent to BE for global testing.

tabnk:
Dragon Quest VII - Fragments of the Forgotten Past (USA) 
Master: **50+ fps**
Older-version PR: **65 fps**
This-version PR: **71 fps**

**Great :)**

MarcoEstevez:
Awesome!

PR #2457, opened 2017-01-20
Fixed microprofile on mingw

FernandoS27:
Fixes microprofile on Mingw.

Don't mind the big diffs, it's just auto clang-format.

the fix is stupid: replace __WIN32 macro with _MSC_VER


linkmauve:
We don’t use clang-formats on externals so you can put your modifications as-is.

Also whenever you do such a large amount of automatic changes, always make it two commits at least, so we can see what actually changed.

FernandoS27:
It's an abandoned external, the mantainance is our responsability now anyways.

yuriks:
Uh, yeah, I don't think anyone is going to review this. I certainly won't. We don't run the clang-format check on externals, so don't re-format it.

FernandoS27:
@yuriks there you go.

linkmauve:
LGTM.

yuriks:
Not entirely convinced by this. Why does switching all of these fix a *crash*? If it were compilation issues it'd make sense, but a crash not really. Is there a specific one that actually fixes it? Even on mingw you still want to use Win32 codepaths for many things.

FernandoS27:
@yuriks ¯\_(ツ)_/¯. It just fixes it. Mingw should be more stable using gcc libraries than microsoft's it seems.

freiro:
Are we just delaying the inevitable?
We should have a clear path for microprofile if we want to keep including it.

darkf:
@FernandoS27 For one thing, this is wrong: all Windows toolchains still have to use winsock2 and the related functions to use sockets. You appear to just be blindly replacing everything with `_MSC_VER` without understanding the context.

I suggest using gdb and debugging the actual crash instead of this.

darkf:
@freiro IMO the working version should be forked and maintained as part of Citra. The Citra team appears uninterested in upgrading microprofile after the author kind of scrapped it for a newer, much different version.

PR #2459, opened 2017-01-20
Updated Citra forum link

3ds-emu:
The Citra forum (discuss.citra-emu.org) has recently been down. The link that has been added 
in this file now leads to an error pages. I update updated the link so it now leads to Citra's new
forum.


jroweboy:
Thanks for catching this. Are there any places in the wiki with the same issue?

3ds-emu:
I think so. Should I fix it?

bunnei:
LGTM, thanks

PR #2464, opened 2017-01-21
[RFC] Improved Vertex Caching

FernandoS27:
This is a basic PR for making vertex caching never fail at the expense of 8mb + 64kb of static data. This PR improves performance on games that have high poly models or which models don't have their triangles ordered.

Basic testing saw improvements on average drawing time of 5% for games like Super Mario 3D Land, Super Mario Bros 2 and Pokemon S/M on open areas.

Normaly, I would test it further but I don't own too many games with high poly models. So I'm publishing it for others to test.


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -53,6 +53,8 @@ RasterizerOpenGL::RasterizerOpenGL() : shader_dirty(true) {
     state.draw.uniform_buffer = uniform_buffer.handle;
     state.Apply();
 
+    vertex_batch.reserve(256 * 256); // 8MB for worst case scenario
```
InusualZ: Why you don't use `VERTEX_CACHE_MAX_SIZE ` here too?

FernandoS27: That's a different file, I could redefine it.

literalmente-game:
I've tested this pr with Mario Kart 7 and... well... just look.

citra-qt 2017-01-21 21-19-16-36

FernandoS27:
that shouldn't happen. Hmn

literalmente-game:
In Resident Evil The Mercenaries 3D:
citra-qt 2017-01-21 21-51-55-89

citra-qt 2017-01-21 21-54-24-47

In Sonic Generations we have some visual glitches as well (not THAT obvious, just little ones).
A Link Between Worlds had a good performance gain, but the glitches are even worse than in Mario Kart 7.

FernandoS27:
I can't seem to reproduce your problems.

literalmente-game:
I'm using the build from your own artifact (https://ci.appveyor.com/api/buildjobs/o6ibnm3u4d6tkj6w/artifacts/citra-windows-msvc-20170121-3006f84.zip).

FernandoS27:
Seems to be a problem specific to Microsoft Visual Studio.

TTFH:
> Seems to be a problem specific to Microsoft Visual Studio

Tested on mingw and also happen

tabnk:
Super Mario 3D Land (U) - Gain extra 5 fps. :)

Before 
image

After
image

tabnk:
Graphic Issue with this PR (Merge with Citra master 25c5aa)

The Legend of Zelda - A Link Between Worlds [USA]
image

Mario Kart 7 [USA]
image

src/video_core/command_processor.cpp
```diff
@@ -225,23 +230,18 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
 
                 u8* texture_data = Memory::GetPhysicalPointer(texture.config.GetPhysicalAddress());
                 g_debug_context->recorder->MemoryAccessed(
-                    texture_data, Pica::Regs::NibblesPerPixel(texture.format) *
-                                      texture.config.width / 2 * texture.config.height,
+                    texture_data,
```
mailwl: Code not clang-formatted (travis fails)

dampih:
wow, it sure brings huge fps in many game, 
some game works perfectly, but some dont such as DQVIII it can reach 60fps in huge area, but broken graphic.. I hope you can fix it 
tessss

FernandoS27:
Fixed, it looks like indices not necessarily start on 0.

tabnk:
With new Fixed PR

Mario Kart 7 [USA]
image

The Legend of Zelda - A Link Between Worlds [USA}
image

Super Smash Bros. for Nintendo 3DS
image

LTGM :)

src/video_core/renderer_opengl/gl_rasterizer.cpp
wwylele: Could you split this into another commit if it is not in the same context?

src/video_core/command_processor.cpp
```diff
@@ -44,6 +44,12 @@ static const u32 expand_bits_to_bytes[] = {
     0xff000000, 0xff0000ff, 0xff00ff00, 0xff00ffff, 0xffff0000, 0xffff00ff, 0xffffff00, 0xffffffff,
 };
 
+constexpr u32 VERTEX_CACHE_MAX_SIZE = (256 * 256);
+constexpr u32 VERTEX_CACHE_MIN_SIZE = 256;
+
+static bool vertex_cache_ids[VERTEX_CACHE_MAX_SIZE] = {false};               // 64 Kb index cache
```
wwylele: Please change this name. Using the same name for different purpose is confusing. Also why this is not an `std::array`?

src/video_core/command_processor.cpp
```diff
@@ -44,6 +44,12 @@ static const u32 expand_bits_to_bytes[] = {
...
+static std::array<Shader::OutputVertex, VERTEX_CACHE_MAX_SIZE> vertex_cache; // 8MB Cache
```
wwylele: Move these stuff ~~into the minimal scope that uses them (under `case PICA_REG_INDEX(trigger_draw_indexed)`)~~ to where the original cache is.  Also, add documentation as the original caching method does

yuriks:
I just tested this on a few games (Zelda: ALBW, Fire Emblem, Project Mirai DX) and in all of them this PR regressed performance by 0-1.5ms. (So, pretty much imperceptible, but never faster.) Can anyone reproduce the perf improvements (and make sure they're not just due to noise)?

We did an analysis of hit rates when #2165 came around, and found that they did not improve significantly when using an unlimited size cache (that PR used a unordered_map, initially). The approach used here has cheaper lookups, but that still wouldn't change the hit rate. It's possible that the games you tested behave differently for some reason, but I'd want to have some data about that to compare then.

tabnk:
Gain 5 FPS at Super Mario 3D Land at world 1-1.
Gain 25-30 FPS at Super Mario Bro 2 Gold Edition in game.
Gain 3 FPS for Zelda: ALBW.

wwylele:
I am wondering if this is related to CPU cache? Say, if the CPU cache size is too small, accessing such big memory block would actually regress the performance.

yuriks:
@wwylele My CPU caches are 32KB/256KB/6MB. These sizes (except for the L3) are standard across all recent Intel CPUs. I don't think it's caused by cache size necessarily, probably varies more with the particular workload of each game (and/or benchmarking methodology.)

yuriks:
An alternative caching method is to use a cache with less ways. The current cache is equivalent to a 32-way cache. When implementing this I tried a 1-way cache but that was worse in general than the current one. Maybe a 2 or 4-way cache would give more competitive results without using so much extra memory (and thus cache). The big issue with this PR's approach is the 64kb memset that has to be done at the start of every batch.

FernandoS27:
@yuriks benefits Super Mario Bros 2, Super Mario Land 3d, Hyrule Warriors and Dragon Quest VII from my testing

yeah the memset is the issue you got.

In theory all the indices should be between [0,Num_vertices) but if you do memsets based on that info. You'll see the triangle issues that appear above.

The thing is that all that section of the code is inaccurate and is wrong. That section should be called "trigger primitive drawing" instead of vertex drawing. The "vertex caching" should work on every pass until it's "finished"

wwylele:
Here are test results on some of my games. The data is in ms/frame reported by citra profiler (smaller number means better performance). The deviation is ±2ms/frame

|  Test case | on current master ea67a89   |  this branch rebased on current master |  
|---|---|---|
| Kirby Planet Robobot 1-1 beginning  | 27  | 27  | 
| Pokemon Sun open area | 56  | 55  | 
| Pokemon Omega Ruby open area | 36  | 36  |
| Super Smash Bros smash at the same stage  | 96 | 96|
|Super Mario 3D Land 1-1 beginning | 39| 37|
| Zelda OoT | 10 | 10|
|New Super Mario Bros 2 1-1 thorough | 30 | 29|

Info of my PC and build:
CPU: Intel® Core™ i7-4500U CPU @ 1.80GHz × 4 
System: Ubuntu 16.04
Citra built by GCC with RelWithDebInfo
Will test on Windows later

src/video_core/command_processor.cpp
```diff
@@ -232,16 +238,12 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
-        const size_t VERTEX_CACHE_SIZE = 32;
-        std::array<u16, VERTEX_CACHE_SIZE> vertex_cache_ids;
-        std::array<Shader::OutputVertex, VERTEX_CACHE_SIZE> vertex_cache;
         Shader::OutputVertex output_vertex;
 
-        unsigned int vertex_cache_pos = 0;
-        vertex_cache_ids.fill(-1);
+        if (is_indexed) {
+            const u32 num_indices = index_u16 ? VERTEX_CACHE_MAX_SIZE : VERTEX_CACHE_MIN_SIZE;
+            std::memset(vertex_cache_ids, false, sizeof(bool) * num_indices);
```
lioncash: You can just use 

```cpp
std::fill(vertex_cache_ids.begin(), vertex_cache_ids.begin() + num_indices, false);
```

to fill an array without needing to care about the underlying type size. It will resolve to the equivalent memset.

src/video_core/command_processor.cpp
```diff
@@ -44,6 +44,12 @@ static const u32 expand_bits_to_bytes[] = {
     0xff000000, 0xff0000ff, 0xff00ff00, 0xff00ffff, 0xffff0000, 0xffff00ff, 0xffffff00, 0xffffffff,
 };
 
+constexpr u32 VERTEX_CACHE_MAX_SIZE = (256 * 256); // 16-bit indices max
+constexpr u32 VERTEX_CACHE_MIN_SIZE = 256;         // 8-bit indices max
+
+static bool vertex_cache_ids[VERTEX_CACHE_MAX_SIZE] = {false};               // 64 Kb index cache
```
lioncash: This should be a `std::array` as well. You also don't need to provide a value in an initializer—this is a C-ism. C++ allows you to just do `= {};`.

src/video_core/command_processor.cpp
```diff
@@ -44,6 +44,12 @@ static const u32 expand_bits_to_bytes[] = {
     0xff000000, 0xff0000ff, 0xff00ff00, 0xff00ffff, 0xffff0000, 0xffff00ff, 0xffffff00, 0xffffffff,
 };
 
+constexpr u32 VERTEX_CACHE_MAX_SIZE = (256 * 256); // 16-bit indices max
```
lioncash: These should have the type `size_t`.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -53,6 +53,11 @@ RasterizerOpenGL::RasterizerOpenGL() : shader_dirty(true) {
     state.draw.uniform_buffer = uniform_buffer.handle;
     state.Apply();
 
+    // reserve 8MB for worst case scenario.
+    // this will avoid vector resizing as we send vertices.
+    constexpr u32 VERTEX_ARRAY_MAX_SIZE = 256 * 256;
```
lioncash: `size_t`

FernandoS27:
Well, let's try out now. Instead of full memsets. It will now calculate the proper min/max indices and patch the cache appropriately.

I'll do thereview changes once I'm sure it works perfectly.

yuriks:
> EDIT: Now vertex caching is global and we only invalidate the cache partialy when:

This is literally useless. Nothing will ever do two draws without changing any of those because it would do nothing.

In addition, you went back into compiler pragma la-la land again.

tabnk:
In Super Mario 3D Land, new version of PR doesn't improve speed like the old version. About same speed as master now.

tabnk:
New Super Mario Bros. 2 Gold Edition - In Game 1-1 

Master: **44 fps**
New Version of this PR: **44 fps**

**No Improvement.** 

image

FernandoS27:
Well, I stopped experimenting and refactored it to be clean now. Thanks for the reviews. Also this goes back to the original idea.

src/video_core/command_processor.cpp
```diff
@@ -215,6 +221,29 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+        if (is_indexed) {
+            u32 min_index;
+            u32 max_index;
+            const size_t last_index = regs.num_vertices;
+            // TODO: is there any easier way to get the min/max indices?
+            // FIX: also regs.num_vertices is actualy the number of indices
+            // in the element being rendered.
+            if (index_u16) {
+                auto result =
+                    std::minmax_element(&index_address_16[0], &index_address_16[last_index]);
```
wwylele: I think `std::minmax_element(index_address_16, index_address_16 + reg.num_vertices);` is easier to read.

src/video_core/command_processor.cpp
```diff
@@ -215,6 +221,29 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                min_index = *result.first;
+                max_index = *result.second;
+            } else {
+                auto result =
+                    std::minmax_element(&index_address_8[0], &index_address_8[last_index]);
+                min_index = *result.first;
+                max_index = *result.second;
+            }
+            const u32 num_indices = max_index - min_index;
```
wwylele: This isn't used?

src/video_core/command_processor.cpp
```diff
@@ -316,7 +338,6 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
     case PICA_REG_INDEX(vs.bool_uniforms):
         for (unsigned i = 0; i < 16; ++i)
             g_state.vs.uniforms.b[i] = (regs.vs.bool_uniforms.Value() & (1 << i)) != 0;
-
```
wwylele: clean up the diff, please

src/video_core/command_processor.cpp
```diff
@@ -44,6 +44,11 @@ static const u32 expand_bits_to_bytes[] = {
     0xff000000, 0xff0000ff, 0xff00ff00, 0xff00ffff, 0xffff0000, 0xffff00ff, 0xffffff00, 0xffffffff,
 };
 
+constexpr size_t VERTEX_CACHE_MAX_SIZE = (256 * 256); // 16-bit indices max
+
+static std::array<bool, VERTEX_CACHE_MAX_SIZE> vertex_cache_ids = {};        // 64 Kb index cache
+static std::array<Shader::OutputVertex, VERTEX_CACHE_MAX_SIZE> vertex_cache; // 8MB Cache
```
wwylele: I still prefer moving these to the local scope, as close as possible to where they are used. Do you have any particular reason not to do that?

src/video_core/command_processor.cpp
```diff
@@ -54,6 +59,7 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
 
     // TODO: Figure out how register masking acts on e.g. vs.uniform_setup.set_value
     u32 old_value = regs[id];
+    u32 old_vsms = regs.vs.main_offset;
```
wwylele: Is this used?

wwylele:
Tested again on Windows with the same machine. Built by MSVC. The data is in ms/frame reported by citra profiler (smaller number means better performance). The deviation is ±2ms/frame. The result:
|  Test case | on current master ea67a89   |  this branch rebased on current master |  
|---|---|---|
|New Super Mario Bros 2 1-1 thorough | 39 | 40|

@tabnk what was your configuration that get a speed boost in your original test?

tabnk:
My configuration: Build on GCC 6.3.0. Intel Ivy-bridge CPU. On Windows 10. 

Super Mario 3D Land - Opening 
Master: 59 fps
This PR: 62 fps

Mario Kart 7 - Mushroom Cup
Master: 28 fps
This PR: 30 fps

Only New Super Mario Bros 2 got big gain. Maybe going test again.

FernandoS27:
It seems the difference in performance is compiler related. I use MINGW64 with GCC 6.2 with options "-O2 -msse3". Anyways, the changes were made.

wwylele:
If the performance improvement does only occur with specified compiler and flags, then IMO it doesn't worth to spend such memory cost on a uncertain optimization. Note that our nightly build is compiled by MSVC, and the bleeding edge build is compiled by MinGW without any flags like -msse3

FernandoS27:
bleeding edge does use -msse2 -msse, it's on it's cmake source code. I use -msse3 because that's the minimum requirement for dynarmic and shader JIT.

In the case of msvc, have you tried testing it with disabling microprofile? microprofile adds some overhead when running each shader jit call, thus rendering this PR unsignificant in effect.

tabnk:
Pokemon Moon 

Master: 48 fps
This PR: 51 fps

image

tabnk:
Bravely Second: End Layer 

Master: 139 fps
This PR: 145 fps

Shinmegumi:
So this my sound silly because I'm new to GitHub and I've been using Citra 3DS as a tool to futher my knowledge in coding, but how would I pull some of these requests to test the commits and see if there is any noticable improvement?  I know how to compile, just not how to use GitHub very well.

wwylele:
(Refer to a deleted comment) @Shinmegumi Please use our [forum](https://community.citra-emu.org/), discord channel or IRC channel for such support question.

FernandoS27:
Well I'm closing this, since I've made a different PR which fixes the problems of both versions.

PR #2468, opened 2017-01-23
appveyor: workaround for unnecesary Xamarin log spam

Kloen:
There is an issue with AppVeyor and Xamarin log spam since long ago: http://help.appveyor.com/discussions/problems/4569-the-target-_convertpdbfiles-listed-in-a-beforetargets-attribute-at-c-does-not-exist-in-the-project-and-will-be-ignored

We simply don't need that lines (About 56 lines in total) in the log either.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,6 +20,8 @@ install:
   - git submodule update --init --recursive
 
 before_build:
+  # Xamarin log spam workaround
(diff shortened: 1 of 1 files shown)


bunnei:
I don't really have a strong opinion on this one -- anyone else care to weigh in before this is merged?

PR #2476, opened 2017-01-27
Oh No! More shader changes!

yuriks:
Oh No! More Lemmings!

These include small functional changes, so not strictly refactors, but nothing major.

Again, recommend reviewing each commit individually.

### DIFF
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -1176,7 +1176,12 @@ struct Regs {
         }
     } command_buffer;
 
-    INSERT_PADDING_WORDS(0x07);
+    INSERT_PADDING_WORDS(4);
+
+    /// Number of input attributes to the vertex shader minus 1
+    BitField<0, 4, u32> max_input_attrib_index;
+
+    INSERT_PADDING_WORDS(2);
 
     enum class GPUMode : u32 {
         Drawing = 0,
--- a/src/video_core/pica.h
+++ b/src/video_core/pica.h
@@ -1225,36 +1225,15 @@ struct Regs {
         // Offset to shader program entry point (in words)
         BitField<0, 16, u32> main_offset;
 
-        union {
-            BitField<0, 4, u64> attribute0_register;
-            BitField<4, 4, u64> attribute1_register;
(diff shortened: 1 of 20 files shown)


src/video_core/pica.h
```diff
@@ -1176,7 +1176,12 @@ struct Regs {
         }
     } command_buffer;
 
-    INSERT_PADDING_WORDS(0x07);
+    INSERT_PADDING_WORDS(4);
+
+    /// Number of input attributes to the vertex shader minus 1
+    BitField<0, 4, u32> max_input_attrib_index;
```
Subv: Does this change break/fix anything obvious?

src/video_core/shader/shader.cpp
```diff
@@ -22,23 +22,28 @@ namespace Shader {
 
 OutputVertex OutputVertex::FromAttributeBuffer(const Regs& regs, AttributeBuffer& input) {
     // Setup output data
-    OutputVertex ret;
+    union {
+        OutputVertex ret{};
+        std::array<float24, 24> vertex_slots;
+    };
+    static_assert(sizeof(vertex_slots) <= sizeof(ret), "Struct and array have different sizes.");
```
Subv: Why <= instead of == ?

src/video_core/pica.h
yuriks: Not really, for correct functioning (under the current VS-only setup) these two have to be set to the same value, this is just for better isolation between the funcional units of the hardware. Ths is the actual configuration of the vetex frontend, whereas the other one was fr the shader input (which the frontend feeds to). Afaik for GS support the previous code would need to switch between the correct input unit, whereas with this reg this will always have the appropriate setting.

src/video_core/shader/shader.cpp
yuriks: OutputVertex has additional members and padding which are used internally. (I do remove those on the next commit.)

yuriks:
Updated with a commit to fix the build error on Linux.

yuriks:
Rebased/autosquashed

yuriks:
I plan on merging this sometimes this weekend.

PR #2479, opened 2017-01-27
Qt: Release all pressed buttons when window focus is lost

### LINKED ISSUE #2392: QT - keypress event gets stuck  when emuwindow looses focus.
this occurs only in the qt frontend, and has existed for as long as i can remember. when playing any game, while pressing a input key, make the main emulation window loose focus( by pressing ALT, window key, or by causing an external window to gain focus). the input key gets 'stuck'. It behaves as if the key is still being pressed, even when the window has lost focus, and can only be fixed when you press the same input key after you switch to the emuwindow.
Tested it on Windows 10.

Hawkheart:
Currently, if the Citra window loses focus for any reason while a button is being held, no key release will be registered until the user presses and releases the key again.

This commit solves this problem by releasing all keys when window focus is lost.

An alternate solution (if this is not acceptable) could be for the Qt window to keep track of which keys are being held and specifically release those.

(Should fix #2392)

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -230,6 +230,10 @@ void GRenderWindow::mouseReleaseEvent(QMouseEvent* event) {
         motion_emu->EndTilt();
 }
 
+void GRenderWindow::focusOutEvent(QFocusEvent* event) {
+    KeyMap::ReleaseAllKeys(*this);
+}
+
 void GRenderWindow::ReloadSetKeymaps() {
     KeyMap::ClearKeyMapping(keyboard_id);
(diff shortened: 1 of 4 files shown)


jroweboy:
@wwylele does this conflict with your input PR? If not I don't see any reason not to merge this. It seems like a pretty simple change and sounds like expected behavior.

wwylele:
@jroweboy as I already meantioned there, yes this does conflict with the input PR. I could cover this with similar simple code in my PR but I readlly don't want to add any new code to expand the reviewing process...

wwylele:
Thanks for your contributing! This is implemented by #2630 with the new input system. Sorry that I took over it without asking...

PR #2483, opened 2017-01-28
http svc moved to its own sub directory

jfmherokiller:
I believe this pull request will help in the organization of the code


wwylele:
I think the current convention is to make sub directories only for service that contain multiple ports. http module only contains one port `http:C` (according to 3dbrew) so no need to make a sub directory?

Even if a sub directory is preferred, why not make sub directories for all services?

mailwl:
why? http:C is alone service

jfmherokiller:
so that if it get split into multiple files during implementation it wont dirty the service directory

mailwl:
implemetation will go to /externals

wwylele:
@jfmherokiller A service implementation getting split into multiple files only happens in the following cases:
 - it contains multiple ports. This happened to most services in sub directories. But it won't happen to http service unless someone find out it has multiple ports.
 - the implementation is so huge that one file can't hold it. This happened to `fs:user` and `ldr:ro`. But This should be decided by who actually implement the service, and the split should be go with the implementation PR

darkf:
This was done in WIP implementations of http:C which had more than one implementation file. If someone plans on picking one of those up, they should probably do it.

jfmherokiller:
so is this pull request a good idea or bad idea? ill close it if it was a bad idea

mailwl:
mostly bad.
but there is another services, like soc:U & soc:P, which nobody cares, because privileged ports not used in games

jfmherokiller:
ok closing it then

PR #2471, opened 2017-01-25
Fix some MSVC warnings V2 + nihstro update

Kloen:
This PR could be bigger, but I think it's big enough.

Fixed about 30 cast warnings ([C4267](https://msdn.microsoft.com/en-us/library/6kck0s93.aspx) and [C4244](https://msdn.microsoft.com/en-us/library/th7a07tz.aspx))
```
C:\projects\citra\src\core\memory.cpp(395): warning C4267: 'initializing': conversion from 'size_t' to 'VAddr', possible loss of data 
C:\projects\citra\src\core\memory.cpp(395): warning C4267: 'initializing': conversion from 'size_t' to 'const VAddr', possible loss of data
C:\projects\citra\src\core\memory.cpp(418): warning C4267: 'argument': conversion from 'size_t' to 'u32', possible loss of data 
C:\projects\citra\src\core\memory.cpp(426): warning C4267: 'argument': conversion from 'size_t' to 'u32', possible loss of data 
C:\projects\citra\src\core\memory.cpp(465): warning C4267: 'initializing': conversion from 'size_t' to 'VAddr', possible loss of data 
C:\projects\citra\src\core\memory.cpp(465): warning C4267: 'initializing': conversion from 'size_t' to 'const VAddr', possible loss of data
C:\projects\citra\src\core\memory.cpp(489): warning C4267: 'argument': conversion from 'size_t' to 'u32', possible loss of data 
C:\projects\citra\src\core\memory.cpp(498): warning C4267: 'argument': conversion from 'size_t' to 'u32', possible loss of data 
C:\projects\citra\src\core\memory.cpp(523): warning C4267: 'initializing': conversion from 'size_t' to 'VAddr', possible loss of data 
C:\projects\citra\src\core\memory.cpp(523): warning C4267: 'initializing': conversion from 'size_t' to 'const VAddr', possible loss of data
C:\projects\citra\src\core\memory.cpp(546): warning C4267: 'argument': conversion from 'size_t' to 'u32', possible loss of data 
C:\projects\citra\src\core\memory.cpp(555): warning C4267: 'argument': conversion from 'size_t' to 'u32', possible loss of data 
C:\projects\citra\src\core\memory.cpp(577): warning C4267: 'initializing': conversion from 'size_t' to 'VAddr', possible loss of data 
C:\projects\citra\src\core\memory.cpp(577): warning C4267: 'initializing': conversion from 'size_t' to 'const VAddr', possible loss of data
[...]

PR #2485, opened 2017-01-28
Switch ComputeHash64 len param to size_t instead of int

Kloen:
Bisected from #2471 

Switch ComputeHash64 len param to size_t instead of int.

Fixes following warning on MSVC:

```
C:\projects\citra\src\core\hle\service\dsp_dsp.cpp(142): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
```

### DIFF
--- a/src/common/hash.h
+++ b/src/common/hash.h
@@ -8,7 +8,7 @@
 
 namespace Common {
 
-void MurmurHash3_128(const void* key, int len, u32 seed, void* out);
+void MurmurHash3_128(const void* key, size_t len, u32 seed, void* out);
 
 /**
  * Computes a 64-bit hash over the specified block of data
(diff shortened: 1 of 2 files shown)


src/common/hash.h
```diff
@@ -8,15 +8,15 @@
 
 namespace Common {
```
lioncash: This header just needs to have `<cstdint>` included above `common_types.h` and it should fix the build issue.

mailwl:
Never understand, why @MerryMage takes hash from dsp fw hash (already unique).
sorry for offtop

wwylele:
(refer to a deleted comment) @mailwl Please go to our IRC channel for such question. "sorry for offtop" is not an excuse to go offtopic

PR #2494, opened 2017-01-29
core: inline CPU, 132 warnings fixed on GCC

Kloen:
This, should, fix, 132, warnings, on, GCC,  ¯\\_(ツ)_/¯.

They happen to exist due to CPU function being declared static on a `.h` file but don't being used in it's own `.cpp` file.

```
/Users/travis/build/citra-emu/citra/src/core/core.h:118:23: warning: unused function 'CPU' [-Wunused-function]
```

PR by guess & docs all around the net, couldn't find nothing official, open to alternative solutions.

Shortened the build log a bit lol:
image

### DIFF
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -115,7 +115,7 @@ private:
     static System s_instance;
 };
 
-static ARM_Interface& CPU() {
+inline ARM_Interface& CPU() {
     return System::GetInstance().CPU();
 }
 


src/core/core.h
```diff
@@ -115,7 +115,7 @@ class System {
     static System s_instance;
 };
 
-static ARM_Interface& CPU() {
+static inline ARM_Interface& CPU() {
```
Subv: Does this really need to be static?

yuriks: Yeah I think this should be just inline

Kloen:
Addresed comments.

Issue #965, opened 2015-07-21
CirclePadPro not functional causes game crashes


RavenHome1:
this is a recreation of the issue https://github.com/citra-emu/citra/issues/954 as requested by @bunnei 

in a nutshell due to not receiving info from the Circle Pad pro majora's mask game just gets confused and straight out crash itself

here is a full console log 

http://pastebin.com/PhLYUxHE

notice: it was not possible to obtain a log from start of the boot due to the fact that an error spam replaces the initial startup logs but the main errors that caused the crash are in there

mailwl:
bug still exists? The game still throws error?
i tryed to implement CirclePadPro, but i don't have one, so, don't know, which exactly data should be return to game

RavenHome1:
@mailwl the issue no longer exists but another issue with the game is that i can't select the "don't use" circle pad option. the game won't respond to me clicking the button and i tried all of the buttons on my keyboard and still nothing.

![capture](https://cloud.githubusercontent.com/assets/12820322/15469468/1113013e-20f3-11e6-9459-06531b92e0f8.PNG)

RavenHome1:
an update, using bleeding 194 the "Don't use" button is now functional  and can bypass the screen, however the ZR key is still not functional

PR #2497, opened 2017-01-30
Refactor input emulation & add SDL gamepad support

wwylele:
First of all, sorry to @makotech222 for interrupting your attempt. Also thanks a lot for the inspiration.

Recommend reviewing commit-by-commit

---- 

### What is changed in this PR
 - Remake the button and analog input using factory pattern
 - ini configuration files are now using a new format to store input configuration
 - SDL gamepad supported. But only configurable through ini files
 - Stop supporting ZL, ZR, Home button and C-Stick for now. They were in the wrong places.
 
### What is not done in this PR, which can be done in the followup
 - Remake the touch and motion input as well. Completely decouple input from EmuWindow
 - Gamepad configuration through GUI
 - Handle ZL, ZR, Home button and C-Stick in correct services

### Concept
An `InputDevice` object is an atomic unit of one input function in 3DS. `InputDevice` exposes one function `GetStatus` to 3DS system side, and expose several functions to the user side for changing status. The return type of `GetStatus` depends on the category. A 3DS consists of these `InputDevice` in 4 categories:
 - `ButtonDevice` (`GetStatus` returns `bool`): 
   - handled by HID: face buttons A/B/X/Y, direction buttons, shoulder buttons L/R, Start, Select
   - handled by IR/IRRST: extra shoulder buttons ZL/ZR (not implemented in this PR)
   - handled by NS?: Home button (not implemented in this PR)
 - `AnalogDevice` (`GetStatus` returns a 2D vector)
   - handled by HID: circle pad
   - handled by IR/IRRST: circle pad pro, c-stick (not implemented)
 - Touch screen device (`GetStatus` returns a 2D vector and a bool): the touch screen handled by HID (not refactored in this PR)
 - Motion sensor device (`GetStatus` returns two 3D vector for accelerometer and gyroscope): one motion sensor handled by HID (not refactored in this PR)
 
A `Factory` produces `InputDevice`s in the same category and using the same input emulation engine. It can also help frontend to communicate with produced `InputDevice`s.

### `ParamPackage`

This is a small implementation of string-based key-value container, which supports serializing to and deserializing from a single string for storing in ini file. This class will be used as parameter for creating input devices. Apart from a mandatory parameter "engine" for the name of the factory, the factory implementation will decide the name (key) of parameters and their meanings. This class will serialize the data in the following format: "key0:value0,key1:value1,...,keyN:valueN", and it support escape charactor: "$0", "$1" and "$2" stand for ":", "," and "$" that appear in keys/values, respectively.

I am not sure if I should implement this by hand, or should use some existing structure like INI, JSON or XML.

### `input_common`

This is a collection of implemented input devices/factories. As a module, Its dependency level is between `core` and `citra`/`citra-qt`. In other words, `input_common` depends on `core` (specifically `core/frontend/input.h`), and is depended by the frontend.

### Implemented input device/factory
 - `Keyboard` (provides `ButtonDevice`): a reimplementation of current keyboard input. Replaces the "direct keys" part of `key_map.cpp`. Parameters required in `ParamPackage` are `"engine:keyboard,code:[key_code]"`, where `[key_code]` is a integer for key code
 - `AnalogFromButton` (provides `AnalogDevice`): a reimplementation of current circle pad emulation. Replaces the "indirect keys" part of `key_map.cpp`. Parameters required in `ParamPackage` are `"engine:analog_from_button,up:[button],down:[button],left:[button],right:[button],modifier:[button],modifier_scale:[scale]"`, where `[button]`s are second level `ParamPackage`s to create `ButtonDevice`s, and `[scale]` is a float number;
 - SDL gamepad input (provides both `ButtonDevice` and `AnalogDevice`): an implementation of SDL input. Parameters required in `ParamPackage` are `"engine:sdl,joystick:[gamepad_id],button:[button_id](for ButtonDevice),hat:[hat_id],direction=[direction_name](for ButtonDevice bound with hat as direction button),axis_x=[axis_id],axis_y=[axis_id](for Analog Device)"`

### Example configuration string in ini file:
The default configuration
```
button_a="code:65,engine:keyboard"
button_b="code:83,engine:keyboard"
circle_pad="engine:analog_from_button,modifier_scale:0.500000,up:code$016777235$1engine$0keyboard,down:code$016777237$1engine$0keyboard,left:code$016777234$1engine$0keyboard,right:code$016777236$1engine$0keyboard,modifier:code$068$1engine$0keyboard"
```
Example configuration to a gamepad
```
button_a="engine:sdl,joystick:0,button:0"
button_b="engine:sdl,joystick:0,button:1"
button_up="engine:sdl,joystick:0,hat:0,direction:up"
circle_pad="engine:sdl,joystick:0,axis_x:0,axis_y:1"
```

Note: input using different engine can mix together.

### Reworking Qt configuration window
In this PR, I reworked Qt input configuration code, but kept the its function and appearance as is (i.e. can only configure input from keyboard). Enabling it to configure gamepad is another big topic and I don't want to introduce it here. For any one want to test gamepad, please manually edit the ini file following the format above.

Note: When reworking input configuration window, I replaced all `Qt::Key` by `int`. This is due to a Qt convention that `Qt::Key` is just a constant group, and use `int` to store key code (take [`QKeyEvent::key()`](http://doc.qt.io/qt-5/qkeyevent.html#key) for example, which returns `int` for code). This also reduces a lot of type casting


src/citra_qt/configure_input.cpp
```diff
@@ -70,77 +94,110 @@ ConfigureInput::ConfigureInput(QWidget* parent)
...
+        }
     }
     updateButtonLabels();
     applyConfiguration();
 }
 
 void ConfigureInput::updateButtonLabels() {
-    for (const auto& input_id : Settings::NativeInput::All) {
-        button_map[input_id]->setText(getKeyName(key_map[input_id]));
+    QString non_keyboard(tr("(non-keybaord)"));
```
j-selby: Typo (non-keyboard)?

src/citra_qt/configure_input.cpp
```diff
@@ -70,77 +94,110 @@ ConfigureInput::ConfigureInput(QWidget* parent)
...
     button->setText(tr("[press key]"));
     button->setFocus();
 
-    current_input_id = input_id;
+    key_setter = new_key_setter;
 
     grabKeyboard();
     grabMouse();
-    timer->start(5000); // Cancel after 5 seconds
+    timer->start(5000); // Cancel after 5 seconds*/
```
j-selby: Am I missing the opening /* for this?

wwylele:
Should fix the input part of #1883. I tried making it thread-safe in each input devices.

This PR conflict with #2479 and doesn't cover the issue there... I can follow the idea and do a similar fix. Should I include it in this PR?

src/common/param_package.cpp
```diff
@@ -0,0 +1,121 @@
...
+// Refer to the license.txt file included.
+
+#include <vector>
+#include "common/logging/log.h"
+#include "common/param_package.h"
+#include "common/string_util.h"
+
+namespace Common {
+
+constexpr char KEY_VALUE_SEPARATER = ':';
```
j-selby: `SEPARATOR`

src/common/param_package.cpp
```diff
@@ -0,0 +1,121 @@
...
+constexpr char PARAM_SEPARATER = ',';
```
j-selby: Ditto, and all below

src/citra_qt/configure_input.cpp
```diff
@@ -70,77 +94,110 @@ ConfigureInput::ConfigureInput(QWidget* parent)
...
+                    analog_map[analog_id][sub_button_id]->setText(KeyToText(param));
+            }
+        }
     }
 }
 
-void ConfigureInput::handleClick(Settings::NativeInput::Values input_id) {
-    QPushButton* button = button_map[input_id];
+void ConfigureInput::handleClick(QPushButton* button, std::function<void(int)> new_key_setter) {
     button->setText(tr("[press key]"));
```
j-selby: Is there a reason why we are using two different brackets here (`(` above, versus `[` here)?

src/citra_qt/configure_input.cpp
wwylele: This */ is a leftover. will remove

src/citra_qt/configure_input.cpp
wwylele: no reason. Will fix it :p

Shinmegumi:
Great work. Will this end up in bleeding edge sooner or later?

src/input_common/main.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "input_common/main.h"
+#ifdef HAVE_SDL2
+#include "input_common/sdl.h"
+#endif
+
+namespace InputCommon {
+
+static std::shared_ptr<Keyboard> keyboard;
+
+void Init() {
```
lioncash: This is kind of the same criticism that I brought up in the original PR, but all interfacing with the input backend should be done through the System class. There should ideally be no need to explicitly require initialization or shutdown. These are what constructors and destructors are for (at least as far as initial initialization/teardown is concerned in class-based interfaces).

This kind of limiting to one input subsystem at a time should not be necessary. i.e. it should be possible to start many system instances using different backends if available or a user chooses without anything preventing that. This also makes using citra as a lib, or moving core libraries towards being used as a library less cumbersome (the only reason System is a singleton at the moment is because namespaced-based interfaces like these still exist). If retrieving the emulated CPU instance is situated within the `System` class, then it also follows that getting the current input backend instance should be possible through that as well (and ideally the same with the video backend and audio backend, but those aren't relevant right now).

Also how well will this handle several different input interfaces on a system? Say someone made an XInput backend and also had the SDL one around - how well would this be able to handle that? What if a user wanted XInput over SDL, or SDL over XInput, how well would it handle that scenario?

src/input_common/sdl.cpp
```diff
@@ -0,0 +1,150 @@
...
+class SDLJoystick;
+class SDLButtonFactory;
+class SDLAnalogFactory;
+static std::unordered_map<int, std::weak_ptr<SDLJoystick>> joystick_list;
+static std::shared_ptr<SDLButtonFactory> button_factory;
+static std::shared_ptr<SDLAnalogFactory> analog_factory;
+
+class SDLJoystick {
+public:
+    SDLJoystick(int joystick_index) {
```
lioncash: `explicit SDLJoystick`

src/input_common/keyboard.cpp
```diff
@@ -0,0 +1,81 @@
...
+// Refer to the license.txt file included.
+
+#include <atomic>
+#include "input_common/keyboard.h"
+
+namespace InputCommon {
+
+class KeyButton : public Input::ButtonDevice {
+public:
+    KeyButton(std::shared_ptr<KeyButtonList> key_button_list_)
```
lioncash: `explicit KeyButton`

src/common/param_package.h
```diff
@@ -0,0 +1,38 @@
...
+    using DataType = std::unordered_map<std::string, std::string>;
+
+    ParamPackage() = default;
+    ParamPackage(const std::string& serialized);
+    ParamPackage(std::initializer_list<DataType::value_type> list);
+    ParamPackage(const ParamPackage& other) = default;
+    ParamPackage(ParamPackage&& other);
+
+    ParamPackage& operator=(const ParamPackage& other) = default;
+    ParamPackage& operator=(ParamPackage&& other);
```
lioncash: You can default this for the same behavior.

src/common/param_package.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+int ParamPackage::Get(const std::string& key, int default_value) const {
+    auto pair = data.find(key);
+    if (pair == data.end()) {
+        LOG_DEBUG(Common, "key %s not found", key.c_str());
+        return default_value;
+    } else {
+        try {
+            return std::stoi(pair->second);
+        } catch (...) {
```
lioncash: This should catch `const std::logic_error&` rather than sinking all exception types. Ditto for below in the other `Get()`

src/common/param_package.cpp
```diff
@@ -0,0 +1,121 @@
...
+    result.resize(result.size() - 1); // discard the trailing PARAM_SEPARATOR
+    return result;
+}
+
+std::string ParamPackage::Get(const std::string& key, const std::string& default_value) const {
+    auto pair = data.find(key);
+    if (pair == data.end()) {
+        LOG_DEBUG(Common, "key %s not found", key.c_str());
+        return default_value;
+    } else {
```
lioncash: `else` isn't necessary since the `if` terminates with a return.

src/common/param_package.cpp
```diff
@@ -0,0 +1,121 @@
...
+    data = std::move(other.data);
+    return *this;
+}
+
+std::string ParamPackage::Serialize() const {
+    std::string result;
+    if (data.empty())
+        return "";
+
+    for (auto pair : data) {
```
lioncash: `const auto&`

src/common/param_package.cpp
```diff
@@ -0,0 +1,121 @@
...
+        std::string key_value[2]{pair.first, pair.second};
```
lioncash: `std::array`

src/common/param_package.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Common {
+
+/// A string-based key-value container supporting serializing to and deserializing from a string
+class ParamPackage {
+public:
+    using DataType = std::unordered_map<std::string, std::string>;
+
+    ParamPackage() = default;
+    ParamPackage(const std::string& serialized);
```
lioncash: This should likely be an `explicit` constructor.

src/input_common/main.cpp
wwylele: The point is input_common is _not a input subsystem_. ( If you like, the `core/frontend/input.h' is the actual input subsystem). core can run perfectly without input_common (just no input method). The original idea is, the frontend provides the input method and plugs in the factory via (core)`Input::RegisterFactory`. The input_common is here only because, our two frontends have the same factory to register (keyboard, analog_from_key and sdl). To avoid the repeating code, I move these three engines into a common module and provide a shortcut to register all of them.

If one want to make a new interface, they can just make a subclass of `Factory`, put the code anywhere they want, and call `RegisterFactory` in frontend booting.

lioncash:
Also with regards to:

> In other words, input_common depends on core

This should be the other way around. Core should depend on InputCommon. It's point of existence should be to bridge inputs so they can be used in conjunction with core. It's somewhat nonsensical to have a `common` lib depending on the core -- in the same way that it would be  odd for the regular `common` to depend on core.

wwylele:
>This should be the other way around. Core should depend on InputCommon. It's point of existence should be to bridge inputs so they can be used in conjunction with core. It's somewhat nonsensical to have a common lib depending on the core -- in the same way that it would be odd for the regular common to depend on core.

Probably because I made a bad name, but `input_common` mean to be a common library for _frontend_ (not for core!). The existence of this is to collect the code that both two frontends will use.

lioncash:
It's still a design issue though. `System` *should* be what houses or provides the reference to the input backend. (via `Input()`, an analog to the CPU backend's `CPU()`, therefore input_common *cannot* depend on core, otherwise you get a circular dependency.

wwylele:
First of all, there is no circular dependency as I already pointed out core doesn't depend on input_common. Second, [core is housing the input backend](https://github.com/citra-emu/citra/pull/2497/commits/51272d2237d77d548c84969a9121580e71d809d4#diff-02f4e9a6e71304cf5e8b7d69c7e7d844R44). The role of input_core is just providing the implementation and call the core to receive them.

I probably still don't fully under stand what you says. So let's try this:
If I remove input_common, move everything in input_common to citra_qt (ignore citra to simplify the problem), and in citra_qt's booting it calls four `Input::RegisterFactories`. Is there any design issue in this case?

lioncash:
> Is there any design issue in this case?

Excluding the cases that it's still:

1. A namespace based "interface" (something that should gradually be gotten rid of in the codebase, at least for the 'main 4' - audio, input, graphics, cpu)
2. Input not being a fixture in `System`

There would be a new third issue that it's now dependent on a specific UI frontend.

Excluding the above three, then no, there is no design issue.

wwylele:
>  1. A namespace based "interface" (something that should gradually be gotten rid of in the codebase, at least for the 'main 4' - audio, input, graphics, cpu)
>  2. Input not being a fixture in System

If this is an issue, I can wrap `core/front_end/input.h` to a class and move it into `System`. I didn't do this because `core/front_end/input.h` is so simple that it doesn't need an initialize function. I will do this later

> There would be a new third issue that it's now dependent on a specific UI frontend.
Yes, I noticed this too. the SDL frontend (and any other PC frontend) should share these factories. So I move them out of citra_qt, to a common frontend library called input_common, so that all frontend can use it. In this step what I did is just to extract the common part of frontend. Does this generate new design issue?

lioncash:
> If this is an issue

Well, yeah, it kind of is. The entire *point* of `System` is to encapsulate all system state. Input is one of the things that comprise said state. You should ideally be able to just create a new instance of System in order to instantiate most of the necessary scaffolding to kick up a new core instance—that's the end goal. That's why it was created in the first place.

> Does this generate new design issue?

No, in that case, it's fine.

wwylele:
> No, in that case, it's fine.
So, that case is what it is now.

src/input_common/main.cpp
lioncash: OK, that makes sense.

However, it looks like there's some backends that would be initialized but not shut down (e.g. `SDL_Init` is called via `SDL::Init`, but `SDL_Quit` would never be called by anything). SDL would also be initialized, even if it hypothetically wasn't an active input backend, but still built into citra.

wwylele:
Also about making `Input` into a class and as a component, I still find it wierd, because:
The state `Input` contains is just some device factory list (currently two). Factories shouldn't be a state of 3DS system (in the same way that real 3DS factory doesn't package itself into a 3DS). Actually, two `System` object can share one factory without any problem (just remember to do synchronization for thread safety). What needs to be a state of `System` is [the devices that factory created](https://github.com/citra-emu/citra/pull/2497/commits/868467f37010da50d118210ed25b86e41727e2d5#diff-18f61ae048ffbd249059d59125522468R49) - yes, they are not `System` members now, but this involves that we should move all system module variables into the `System` object, which is another big topic that I don't want to talk about here

wwylele:
Addressed comments above except for the SDL init/quit one. Give me a while to think

archshift:
Does this address threading at all? (wrt #1883)

wwylele:
@archshift tbh I don't really understand the thread issue of input. I think as long as the input status is atomic-read/write, it has no problems. Probably I missed something. Anyway, I am not confident enough so I erased that sentence.

makotech222:
@wwylele Keyboard inputs are gathered from the frontend on a separate thread than what citra is running on.

wwylele:
@makotech222 sure I knew that. What I don't know is what else I should do, after I made the status that two threads read/write atomic.

wwylele:
Update:
 - Add `Shutdown` function in input_common/main.cpp. Do quit SDL joystick subsystem when shutting down. 
 - Change `HID::ReloadInputDevices` to just make a notification. The actual reloading will happen on the next `UpdatePadCallback` cycle. This avoid a potential race condition if the device gets reload when `UpdatePadCallback` is reading the status.

Also about what @lioncash said:
>SDL would also be initialized, even if it hypothetically wasn't an active input backend, but still built into citra.

I do recognize this issue. However, I want to put this to another PR, where the SDL engine can be initialized conditionally depends on user settings. It should be as trivial as just break up `InputCommon::Init()` into small pieces. Currently leaving SDL initialized doesn't hurt anything.

phire:
I tested this branch for a few hours; Configuring through ini files is a bit of a pain.

One minor flaw I found is that SDL doesn't expose the d-pad buttons over it's button API, instead it converts them to a hat input (which this PR doesn't support) and exposes that.

Fixing this would require adding a second ButtonDevice class to SDL which binds to hat inputs. Could be accessed with a `engine:sdl,joystick:0,hat:up` parameter pack.

wwylele:
Update:

 - added support for hat in SDL. usage example: `button_up="engine:sdl,joystick:0,hat:0,direction:up"`
 - fixed variable shadowing and missing `override` in the SDL part

phire:
Tested the dpad mapping, it works.

Here is my config for an xinput controller, if anyone wants to test: 

    button_a="engine:sdl,joystick:0,button:1"
    button_b="engine:sdl,joystick:0,button:0"
    button_x="engine:sdl,joystick:0,button:3"
    button_y="engine:sdl,joystick:0,button:2"
    button_up="engine:sdl,joystick:0,hat:0,direction:up"
    button_down="engine:sdl,joystick:0,hat:0,direction:down"
    button_left="engine:sdl,joystick:0,hat:0,direction:left"
    button_right="engine:sdl,joystick:0,hat:0,direction:right"
    button_l="engine:sdl,joystick:0,button:4"
    button_r="engine:sdl,joystick:0,button:5"
    button_start="engine:sdl,joystick:0,button:7"
    button_select="engine:sdl,joystick:0,button:6"
    button_zl="engine:keyboard,code:49"
    button_zr="engine:keyboard,code:50"
    button_home="engine:keyboard,code:66"
    circle_pad="axis_x:0,engine:sdl,joystick:0,axis_y:1"
    c_stick="axis_x:2,engine:sdl,joystick:0,axis_y:3"

The mapping works with any xinput compatible controller, I personally tested xbox 360, xbox one and the dual shock 4 through a xinput wrapper. Note, X Y A B buttons match the 3DS's face layout, not the labels on the 360 controller. L and R are mapped to the bumpers. 

One extra deficiency I've just noticed. You are going to need the reverse `button_from_analog` mapping at some point, otherwise you can't bind ZL and ZR to the triggers. I guess it's out of the scope of this PR.

src/input_common/sdl.cpp
```diff
@@ -0,0 +1,204 @@
...
+    std::shared_ptr<SDLJoystick> joystick;
+    int button;
+};
+
+class SDLDirectionButton : public Input::ButtonDevice {
+public:
+    explicit SDLDirectionButton(std::shared_ptr<SDLJoystick> joystick_, int hat_, Uint8 direction_)
+        : joystick(joystick_), hat(hat_), direction(direction_) {}
+
+    bool GetStatus() const {
```
phire: Needs to be marked `override`

src/input_common/sdl.cpp
```diff
@@ -0,0 +1,204 @@
...
+            return {};
+        SDL_JoystickUpdate();
+        float x = SDL_JoystickGetAxis(joystick, axis_x) / 32767.0f;
+        float y = SDL_JoystickGetAxis(joystick, axis_y) / 32767.0f;
+        y = -y; // 3DS uses an y-axis inverse from SDL
+
+        // Make sure the coordinates are in the unit circle,
+        // otherwise normalize it.
+        float r = x * x + y * y;
+        if (r > 1) {
```
phire: Really should be `r > 1.0f`

mailwl: interesting... compiler casts r to int, or 1 to float?

phire: Compiler casts 1 to float.

https://godbolt.org/g/E6ZycE

mailwl: ah, i see. so casting int(1) to (float)1 just for clarity or unusual compiler in the future?
make sense

phire: Type promotion behaviour is defined by the c++ standard, so all compilers *should* do the same thing.

It's mainly for readability. Some people do have the type promotion rules memorized and will instantly know it will do the correct thing. But for the rest of us, we just waste time wondering what will actually happen.

Also, I have been burned by the type promotion behaviour [in the past.](https://github.com/dolphin-emu/dolphin/pull/3774)

mailwl:
just a question. ZL/ZR+circle pad is really IR responsibility?
i mean new 3ds still doubles on IR keypressed events?

wwylele:
@mailwl yes if 3dbrew documents it correctly: https://www.3dbrew.org/wiki/IRRST_Shared_Memory

wwylele:
Update:
 - addressed comments above
 - InputCommon: add AnalogFromButton
```
- Input::UnregisterFactory<Input::ButtonDevice>("analog_from_button");
+ Input::UnregisterFactory<Input::AnalogDevice>("analog_from_button");
```

src/input_common/analog_from_button.h
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * An analog device factory that takes direction button devices and combines them into a analog
+ * device.
+ */
+class AnalogFromButton : public Input::Factory<Input::AnalogDevice> {
+public:
+    /**
+     * Creates an analog device from direction button devices
+     * @param params contains parameters for creating the device:
+     *     "up": a serialized ParamPackage for creating a button device for up direction
```
lioncash: These should have hyphens before them (e.g. `- "up" ...`) so that doxygen would generate a bulleted list.

src/input_common/sdl.cpp
```diff
@@ -0,0 +1,204 @@
...
+        if (r > 1.0f) {
+            r = std::sqrt(r);
+            x /= r;
+            y /= r;
+        }
+
+        return std::make_tuple(x, y);
+    }
+
+    bool GetHatDirection(int hat, Uint8 direction) {
```
lioncash: This can be a const member function

src/input_common/sdl.cpp
```diff
@@ -0,0 +1,204 @@
...
+        return (SDL_JoystickGetHat(joystick, hat) & direction) != 0;
+    }
+
+private:
+    SDL_Joystick* joystick;
+};
+
+class SDLButton : public Input::ButtonDevice {
```
lioncash: These classes can likely be marked `final`.

src/input_common/keyboard.h
```diff
@@ -0,0 +1,46 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <list>
```
lioncash: `<memory>` should also be included

src/input_common/keyboard.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include <atomic>
+#include "input_common/keyboard.h"
+
+namespace InputCommon {
+
+class KeyButton : public Input::ButtonDevice {
+public:
+    explicit KeyButton(std::shared_ptr<KeyButtonList> key_button_list_)
+        : key_button_list(key_button_list_), status(false) {}
```
lioncash: You can directly initialize `status` as false like so in the class body:

```cpp
std::atomic<bool> status{false};
```

src/input_common/keyboard.cpp
```diff
@@ -0,0 +1,81 @@
...
+class KeyButtonList {
+public:
+    void AddKeyButton(int key_code, KeyButton* key_button) {
+        std::lock_guard<std::mutex> guard(mutex);
+        list.push_back(KeyButtonPair{key_code, key_button});
+    }
+
+    void RemoveKeyButton(KeyButton* key_button) {
+        std::lock_guard<std::mutex> guard(mutex);
+        list.remove_if([key_button](KeyButtonPair& pair) { return pair.key_button == key_button; });
```
lioncash: `pair` can be `const`. The parameter,`key_button`, can also be const, since it's only used in the query and isn't modified.

src/input_common/keyboard.cpp
```diff
@@ -0,0 +1,81 @@
...
+    std::mutex mutex;
+    std::list<KeyButtonPair> list;
+};
+
+KeyButton::~KeyButton() {
+    key_button_list->RemoveKeyButton(this);
+}
+
+Keyboard::Keyboard() {
+    key_button_list = std::make_shared<KeyButtonList>();
```
lioncash: This can go in the initializer list:

```cpp
Keyboard::Keyboard() : key_button_list{std::make_shared<KeyButtonList>()} {
}

File: src/input_common/keyboard.cpp [hunk 24]
```diff
@@ -0,0 +1,81 @@
...
+}
+
+std::unique_ptr<Input::ButtonDevice> Keyboard::Create(const Common::ParamPackage& params) {
+    int key_code = params.Get("code", 0);
+    std::unique_ptr<KeyButton> button = std::make_unique<KeyButton>(key_button_list);
+    key_button_list->AddKeyButton(key_code, button.get());
+    return std::move(button);
```
lioncash: You don't need to explicitly move here.

src/input_common/sdl.cpp
```diff
@@ -0,0 +1,204 @@
...
+static std::unordered_map<int, std::weak_ptr<SDLJoystick>> joystick_list;
+static std::shared_ptr<SDLButtonFactory> button_factory;
+static std::shared_ptr<SDLAnalogFactory> analog_factory;
+
+static bool initialized = false;
+
+class SDLJoystick {
+public:
+    explicit SDLJoystick(int joystick_index) {
+        joystick = SDL_JoystickOpen(joystick_index);
```
lioncash: You can use a `unique_ptr` here to get rid of the need to close it in the destructor.

You declare your class member like:

```cpp
std::unique_ptr<SDL_Joystick, decltype(&SDL_JoystickClose)> joystick;
```

then you'd make the constructor:

```cpp
explicit SDLJoystick(int joystick_index) : joystick{SDL_JoystickOpen(joystick_index), SDL_JoystickClose} {
   if (!joystick) {
       LOG_ERROR(Input, "failed to open joystick %d", joystick_index);
   }
}
```

which will automatically call `SDL_JoystickClose` whenever an SDLJoystick goes out of scope.

src/core/hle/service/hid/hid.cpp
```diff
@@ -2,10 +2,12 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <atomic>
```
lioncash: `<memory>` and `<algorithm>` needs to be included.

src/core/frontend/input.h
```diff
@@ -0,0 +1,108 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
+#include <tuple>
```
lioncash: `<string>` and `<utility>` need to be included.

src/common/param_package.h
```diff
@@ -0,0 +1,39 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <string>
```
lioncash: `<initializer_list>` needs to be included.

src/citra_qt/configure_input.h
```diff
@@ -4,10 +4,13 @@
 
 #pragma once
 
+#include <array>
+#include <functional>
```
lioncash: `<string>` needs to be included.

src/citra_qt/config.h
```diff
@@ -23,5 +23,7 @@ class Config {
 
     void Reload();
     void Save();
-    static const std::array<QVariant, Settings::NativeInput::NUM_INPUTS> defaults;
+
+    static const std::array<int, Settings::NativeButton::NumButtons> default_buttons;
```
lioncash: `<array>` needs to be included.

src/input_common/sdl.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+};
+
+/// An analog device factory that creates analog devices from SDL joystick
+class SDLAnalogFactory : public Input::Factory<Input::AnalogDevice> {
+public:
+    /**
+     * Creates analog device from joystick axes
+     * @param params contains parameters for creating the device:
+     *     "joystick": the index of the joystick to bind
```
lioncash: These should have a hyphen before them so that doxygen will make a bulleted list.

src/input_common/sdl.cpp
```diff
@@ -0,0 +1,204 @@
...
+    return joystick;
+}
+
+/// A button device factory that creates button devices from SDL joystick
+class SDLButtonFactory : public Input::Factory<Input::ButtonDevice> {
+public:
+    /**
+     * Creates a button device from a joystick button
+     * @param params contains parameters for creating the device:
+     *     "joystick": the index of the joystick to bind
```
lioncash: These should have a hyphen before them so that doxygen creates a bulleted list.

src/input_common/sdl.cpp
```diff
@@ -0,0 +1,204 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
```
lioncash: `<cmath>`, `<memory>`, `<string>`, and `<tuple>` need to be included.

src/input_common/main.h
```diff
@@ -0,0 +1,28 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <string>
+#include "input_common/keyboard.h"
```
lioncash: this can be a forward declaration.

src/input_common/main.cpp
```diff
@@ -0,0 +1,61 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/param_package.h"
```
lioncash: `<memory>` needs to be included.

src/input_common/keyboard.h
```diff
@@ -0,0 +1,46 @@
...
+#include <mutex>
```
lioncash: `<list>` and `<mutex>` should be in the cpp file, as they're not used anywhere in the header.

src/input_common/analog_from_button.h
```diff
@@ -0,0 +1,33 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <list>
```
lioncash: `<list>` and `<mutex>` aren't used anywhere in the header.

src/input_common/main.h
wwylele: `Keyboard` needs to expose its `KeyPressed/Released` interface to frontend

wwylele:
Addressed comments above and rebased. I will split out the sdl code on another day.

src/input_common/main.h
lioncash: You can just include this where it's needed then. The only thing this immediate header needs defined is `Keyboard*` which a forward declaration will suffice in doing.

wwylele:
Update: I added back the `std::move` in `Keyboard::Create`. Removing it causes error on clang: 
```
/Users/travis/build/citra-emu/citra/src/input_common/keyboard.cpp:71:12: error: no viable conversion from returned value of type 'unique_ptr<InputCommon::KeyButton>' to function return type 'unique_ptr<Input::ButtonDevice>'
```
This is due to the return type is not the same of the returned object. It need to do derived->base conversion and an implicit move is not allowed.

wwylele:
moved sdl.cpp/sdl.h into input_common/sdl folder

bunnei:
Fixes #965

wwylele:
Signing off from bunnei means this is out of RFC state.
more reviews are welcome.

wwylele:
Update: fixed connecting to nullptr in configure_input.

disk64:
Hey all,
While trying out this pull request, I've found that in games such as Majora's Mask and Ocarina of Time, the character often cannot run at full speed with the joystick fully pressed in a direction (depending on the direction pressed).
Looking at jstest-gtk, the axis values I'm getting from my joystick range from -32767 to 32124 for the X axis and -32458 to 32451 for Y axis, which are very close to the actual min/max range of these axes, so I suspect something in citra's code is the cause.

Digging into the code, the variable MAX_CIRCLEPAD_POS looked suspicious (defined [here](https://github.com/citra-emu/citra/pull/2497/files#diff-18f61ae048ffbd249059d59125522468R125) )

I tried changing the value of this variable from 0x9C to 0xFF for kicks to see what would happen.
Sure enough, I was then able to run at full speed in OoT3D/MM3D!

I'm not sure why a max analog cap was made for the circlepad, but I would recommend it be removed or revised, as best as I can tell it is resulting in inaccurate input emulation.
As a sidenote, this should also translate over when working on re-implementing C-stick support.

wwylele:
@disk64 the value `0x9C` is inherited from [the old code](https://github.com/citra-emu/citra/blob/85ba60d5ecc6fbd809223318d76b6e7fb3c07d27/src/core/frontend/emu_window.cpp#L22) (= not something I made up in this PR), and was originally observed as a maximum value that a real 3DS can get (= we were trying to get accurate input emulation). 

Could you try if keyboard input shares the same issue (Try both master branch and this PR)?  Note that keyboard input also uses this maximum value. If issue also presents in keyboard input, we need to revise the value; otherwise, it is more like the games are very sensitive to the little difference between actual value and maximum one. Then we need joystick calibration feature in citra (will be in another PR)

disk64:
I initially didn't think this was an issue joystick calibration will solve- looking at the min/max values for my joystick that would be used for calibration, my joystick is hitting very close to the limits of joystick input values (-32768 to 32767), so calibration would do practically nothing for the controller I'm using compared to just using the raw input values as-is.

I tested keyboard input and it does not have this issue- OK, so the value 0x9C should stay.
Given 0x9C is the correct max value for the circle pad, and that calibration would not significantly change the input coming from my controller (as I've described), I figured there must be a bug hiding somewhere else, so I took some time to dig deeper.
Eventually I found that the joystick values coming from SDL_JoystickGetAxis() were not matching the values I was seeing in jstest-gtk, instead ranging from -27628 to 25571 on the X-axis and -26857 to 26856 on the Y-axis.
jstest-gtk was adjusting the joystick values based on calibration settings, which I didn't think it was doing, my apologies.

On Linux SDL2 uses the newer evdev, which has a separate set of calibration settings from the older joydev (which jstest-gtk uses). Like joydev, evdev's calibration settings are applied system-wide (so SDL2's joystick values will be affected by this evdev joystick calibration).
Unfortunately there isn't a GUI program for working with evdev joystick calibration settings like jstest-gtk for joydev, but there is a command-line tool, evdev-joystick, so I used this to configure my joystick properly, after which SDL2 returned proper, calibrated values when SDL_JoystickGetAxis() was called.

At least in the case of Linux, this is less an issue of citra needing a calibration feature and more an issue of Linux needing a GUI tool for system-wide joystick calibration (evdev-joystick is a relatively new program, and at one point [some websites](http://atomicinf.tumblr.com/post/79545628906/calibrating-evdev-joysticks) were mistakenly reporting that evdev had no calibration feature at all because there was no tool that could modify the calibration settings that were there yet).

Well, this was a fun rabbit-hole to poke my head in, I learned a lot of things about my operating system I'm surprised I didn't know already ;)

InusualZ:
Maybe someone could write that "error" in an FAQ or something. People may encounter it in the future.

disk64:
It's also worth noting that currently evdev lacks the same toolset that joydev has, which includes a graphical joystick calibration tool (jstest-gtk), a tool for storing and restoring calibration data (jscal-store, jscal-restore) your device), and a udev rule installed with the "joystick" debian package that calls jscal-restore when your device shows up.

Without these tools, users must re-calibrate their joystick using the command-line tool evdev-joystick for use with evdev every time they connect their controller. (How on earth has no one developed equivalent tools for evdev yet? I guess most people get by with uncalibrated joysticks and never notice anything.)

src/common/param_package.cpp
```diff
@@ -0,0 +1,120 @@
...
+            part = Common::ReplaceAll(part, {KEY_VALUE_SEPARATOR}, KEY_VALUE_SEPARATOR_ESCAPE);
+        }
+        result += key_value[0] + KEY_VALUE_SEPARATOR + key_value[1] + PARAM_SEPARATOR;
+    }
+
+    result.pop_back(); // discard the trailing PARAM_SEPARATOR
+    return result;
+}
+
+std::string ParamPackage::Get(const std::string& key, const std::string& default_value) const {
```
jroweboy: Should this be made more generic? You can template this method using stringstream so we don't have to have 3 different copy pasted setters/getters This method should automatically infer the type from the default value provided.

```cpp
#include <sstream>
...

template <typename T>
T ParamPackage::Get(const std::string& key, const T default_value) const {
    auto pair = data.find(key);
    if (pair == data.end()) {
        LOG_DEBUG(Common, "key %s not found", key.c_str());
        return default_value;
    }

    T value;
    std::stringstream stream(pair->second);
    stream >> value;
    if (stream.fail()) {
        LOG_ERROR(Common, "failed to convert %s to the expected type", pair->second.c_str());
        return default_value;
    }
    return value;
}
```

and I'm guessing the setter doesn't need to do anything special since `to_string` has overloads for the most common types

```cpp
template <typename T>
void ParamPackage::Set(const std::string& key, const T value) {
    data[key] = std::to_string(value);
}
```

wwylele: This can lead to misuse for strings. Users of `ParamPackage` tends to write code like `param_package.Get("some_key", "default_string")`, but with templates this will be deducted as `T = const char*` and instantiate an ill-formed function. So I guess I had better not make templates until we introduce c++17 deduction guides.

Curnivore:
People redirected here might be interested in this method I wrote about for Citra: https://community.citra-emu.org/t/how-to-use-a-mouse-as-a-virtual-joystick-in-citra-in-order-to-control-a-camera-view-or-for-custom-mouse-input/1885

It practically allows full capturing of the mouse for camera viewing (depended on game), and even makes it possible to bind the main mouse buttons on demand.

The main technology is Virtual Joystick via vJoy -> SDL -> Citra. An autohotkey script talks to vJoy. The use of Autohotkey can have very powerful effects.

src/citra_qt/configure_input.cpp
TAD3V: missing a third bracket at the end

Issue #2326, opened 2016-12-15
Apps that try to close themselves causes the emulator to crash


deadair3210:
version = HEAD-b3c53b6
Any application that calles 'PrepareToCloseApplication' or 'CloseApplication' will cause the emulator to crash, the last thing in the log is 'Unreachable Code'

mailwl:
link here related #1663

deadair3210:
Didn't know it already existed, sorry.
The only thing I noticed was that the result was different, will add that info to the other issue

mailwl:
no, it different issues, one from user applications, another from system/homebrew

deadair3210:
So the issue still stands?

Issue #1663, opened 2016-04-13
SVC::CallSVC:1055: unimplemented SVC function ExitProcess(..)


JamePeng:
Exit the homebrew application will cause this error!
![qq 20160413143221](https://cloud.githubusercontent.com/assets/17095606/14484777/3eaf3ab4-0185-11e6-8b78-898a243ac922.png)

I think it should return to homebrew launcher(which launch the homebrew) or this window
![qq 20160413144126](https://cloud.githubusercontent.com/assets/17095606/14484859/d8a28c8e-0185-11e6-8f21-550871be7a2a.png)

deadair3210:
Emulator now crashes with an 'Unreachable Code!' error as of version = HEAD-b3c53b6

Subv:
This function has already been implemented, please create a new issue with a log if the crashing persists.

PR #2506, opened 2017-01-31
SVC::ExitProcess: implement via callback

mailwl:
Should fix: 
SVC::CallSVC:1055: unimplemented SVC function ExitProcess(..) #1663
Apps that try to close themselves causes the emulator to crash #2326

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -147,6 +147,8 @@ static u8* GetPointerFromVMA(VAddr vaddr) {
     case Kernel::VMAType::BackingMemory:
         direct_pointer = vma.backing_memory;
         break;
+    case Kernel::VMAType::Free:
+        return nullptr;
     default:
         UNREACHABLE();
     }
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -163,6 +163,8 @@ int main(int argc, char** argv) {
         break; // Expected case
     }
 
+    system.SetExitCallback([&]{ emu_window->CloseWindow(); });
+
     while (emu_window->IsOpen()) {
         system.RunLoop();
     }
--- a/src/citra/emu_window/emu_window_sdl2.h
+++ b/src/citra/emu_window/emu_window_sdl2.h
@@ -31,6 +31,10 @@ public:
     /// Whether the window is still open, and a close request hasn't yet been sent
     bool IsOpen() const;
 
+    void CloseWindow() {
+        is_open = false;
+    }
+
     /// Load keymap from configuration
     void ReloadSetKeymaps() override;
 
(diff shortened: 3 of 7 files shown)


wwylele:
But commercial game, such as Pokemon OR/AS (L+R+Start), calls this function to reboot the game

src/core/memory.cpp
```diff
@@ -147,6 +147,8 @@ static u8* GetPointerFromVMA(VAddr vaddr) {
     case Kernel::VMAType::BackingMemory:
         direct_pointer = vma.backing_memory;
         break;
+    case Kernel::VMAType::Free:
+        return nullptr;
```
wwylele: what's this?

mailwl: video_core try to flush cache on unmapped memory after exit, and Citra crash on UNIMPLEMENTED()

mailwl:
Pokemon OR call Exit Process twice, an crash Citra.
callbacks is wrong, because SVC 3 must be last emulated opcode in process.
Don't know, how to fix it...

mailwl:
About rebooting back to game, I think, it somewhere in umimplemented functions (ExitProcess has no parameters):
```
[  95.191735] Service <Error> core\hle\service\service.cpp:Service::Interface::HandleSyncRequest:95: unknown / unimplemented function 'PrepareToDoApplicationJump': port=APT:A, cmd_buff[1]=0x2, cmd_buff[2]=0xFFFFFFFF, cmd_buff[3]=0xFFFFFFFF, cmd_buff[4]=0x0
[  95.193145] Service.GSP <Warning> core\hle\service\gsp_gpu.cpp:Service::GSP::ReleaseRight:705: called
[  95.193988] Service <Error> core\hle\service\service.cpp:Service::Interface::HandleSyncRequest:95: unknown / unimplemented function 'DoApplicationJump': port=APT:A, cmd_buff[1]=0x0, cmd_buff[2]=0x0, cmd_buff[3]=0x2, cmd_buff[4]=0x0, cmd_buff[5]=0x802, cmd_buff[6]=0x0
```

wwylele:
So I assume `ExitProcess` does mean to terminate the process instead of reboot or something. To implement this, in the function you should immediately abort the current CPU emulation, unschedule all threads, stop everything (to prevent video cache from flushing data or something) and then finally calls a callback to notify the frontend to stop. You can't rely on the frontend to terminate the core from outside. As you can tell from the code, `ShutdownGame` doesn't terminate the emulation thread immediately, but only send a signal to it, and the emulation thread will still execute (beyond `ExitProcess`) for a while.

mailwl:
video caching flush before call ExitProcess, as part of unmapping memory, so, if remove return nullptr to unmapped memory, citra crash before call this function. I think, this situation for other PR, to do it in right way, but for tests i leave it as is (this PR not ready even for bleeding edge without changes).
I close it for now, before implementing stopping CPU

PR #2511, opened 2017-01-31
added comments the applet ids and added missing app ids

jfmherokiller:
i updated the applet ids in the apt.h file so they are equal to the ones on [this]
(https://3dbrew.org/wiki/NS_and_APT_Services) page. 
I also added a comment above each applet id containing the info from the description column.

 However, applet id 400 does not seem to be on that page and applet id 300 did not have a better description so those have question marks on them for possible later review. 

The developer ids were added because they could be useful for testing certain pieces of homebrew which either override an applet or make use of the developer applet slots to provide additional functionality.(maybe an ingame screenshot twitter/imgur uploader?)


src/core/hle/service/apt/apt.h
```diff
@@ -70,35 +70,84 @@ enum class SignalType : u32 {
...
     InstructionManual = 0x115,
+    /// Notifications applet (newslist)
     Notifications = 0x116,
+    /// Miiverse applet (olv)
     Miiverse = 0x117,
+    /// Miiverse posting applet (solv3)
     MiiversePost = 0x118,
+    /// Amiibo settings (cabinet)
     AmiiboSettings = 0x119,
+    /// Software Keyboard (swkbd) (?)
```
Subv: Why do these have question marks?

src/core/hle/service/apt/apt.h
```diff
@@ -70,35 +70,84 @@ enum class SignalType : u32 {
...
+    /// Circle Pad Pro Calibrator (extrapad) (?)
     Extrapad = 0x208,
+    /// Notepad (memolib) (?)
     Memolib = 0x209,
+    /// Application?
     Application = 0x300,
+    /// eShop (tiger)
+    Tiger = 0x301,
+    /// Library applet?
     AnyLibraryApplet = 0x400,
```
Subv: This one is supposed to mean the currently running library applet.

src/core/hle/service/apt/apt.h
```diff
@@ -70,35 +70,84 @@ enum class SignalType : u32 {
...
+    /// Error Display (error) (?)
     Error = 0x206,
+    /// eShop applet (mint) (?)
     Mint = 0x207,
+    /// Circle Pad Pro Calibrator (extrapad) (?)
     Extrapad = 0x208,
+    /// Notepad (memolib) (?)
     Memolib = 0x209,
+    /// Application?
     Application = 0x300,
```
Subv: This one is the currently running application.

mailwl:
maybe better move comments to one line with enum items, like here:
https://github.com/citra-emu/citra/blob/master/src/core/core.h#L34

src/core/hle/service/apt/apt.h
jfmherokiller: because i copied them from the description column

jfmherokiller:
@mailwl i thought of doing that but i wasnt sure if that was the correct style or not + it just looks weird to me

src/core/hle/service/apt/apt.h
jfmherokiller: updated it in the latest version

src/core/hle/service/apt/apt.h
Subv: Please remove the question marks

jfmherokiller:
closed because of disinterest

PR #2512, opened 2017-02-01
Add custom layout settings.

SonofUgly:
Allows users to scale and place both screens how they want. This takes care of most frame layout related feature requests for now, until simpler solutions are made.

It's not the most user-friendly thing - setting where the edges of the screens are within the render window may not be that intuitive for some, and then there's having to change 8 settings when you change the window size - so perhaps this would be best as just an .ini setting for advanced users?
Regardless, here's what the settings currently look like:

### DIFF
--- a/src/citra_qt/configure_graphics.cpp
+++ b/src/citra_qt/configure_graphics.cpp
@@ -14,6 +14,8 @@ ConfigureGraphics::ConfigureGraphics(QWidget* parent)
     this->setConfiguration();
 
     ui->toggle_vsync->setEnabled(!Core::System::GetInstance().IsPoweredOn());
+
+    ui->layoutBox->setEnabled(!Settings::values.custom_layout);
 }
 
 ConfigureGraphics::~ConfigureGraphics() {}
--- a/src/citra_qt/configure_graphics.ui
+++ b/src/citra_qt/configure_graphics.ui
@@ -126,7 +126,7 @@
     </layout>
    </item>
    <item>
-    <widget class="QGroupBox" name="groupBox2">
+    <widget class="QGroupBox" name="layoutBox">
      <property name="title">
       <string>Layout</string>
      </property>
--- a/src/citra/config.cpp
+++ b/src/citra/config.cpp
@@ -77,6 +77,23 @@ void Config::ReadValues() {
     Settings::values.layout_option =
         static_cast<Settings::LayoutOption>(sdl2_config->GetInteger("Layout", "layout_option", 0));
     Settings::values.swap_screen = sdl2_config->GetBoolean("Layout", "swap_screen", false);
+    Settings::values.custom_layout = sdl2_config->GetBoolean("Layout", "custom_layout", false);
+    Settings::values.custom_top_left =
+        static_cast<u16>(sdl2_config->GetInteger("Layout", "custom_top_left", 0));
+    Settings::values.custom_top_top =
+        static_cast<u16>(sdl2_config->GetInteger("Layout", "custom_top_top", 0));
+    Settings::values.custom_top_right =
+        static_cast<u16>(sdl2_config->GetInteger("Layout", "custom_top_right", 400));
+    Settings::values.custom_top_bottom =
+        static_cast<u16>(sdl2_config->GetInteger("Layout", "custom_top_bottom", 240));
+    Settings::values.custom_bottom_left =
+        static_cast<u16>(sdl2_config->GetInteger("Layout", "custom_bottom_left", 40));
+    Settings::values.custom_bottom_top =
+        static_cast<u16>(sdl2_config->GetInteger("Layout", "custom_bottom_top", 240));
+    Settings::values.custom_bottom_right =
+        static_cast<u16>(sdl2_config->GetInteger("Layout", "custom_bottom_right", 360));
+    Settings::values.custom_bottom_bottom =
+        static_cast<u16>(sdl2_config->GetInteger("Layout", "custom_bottom_bottom", 480));
 
     // Audio
     Settings::values.sink_id = sdl2_config->Get("Audio", "output_engine", "auto");
--- a/src/citra/default_ini.h
+++ b/src/citra/default_ini.h
@@ -71,6 +71,21 @@ bg_green =
 # 0 (default): Default Top Bottom Screen, 1: Single Screen Only, 2: Large Screen Small Screen
 layout_option =
 
+# Toggle custom layout (using the settings below) on or off.
+# 0 (default): Off , 1: On
+custom_layout =
+
+# Screen placement when using Custom layout option
+# 0x, 0y is the top left corner of the render window.
+custom_top_left =
+custom_top_top =
+custom_top_right =
+custom_top_bottom =
+custom_bottom_left =
+custom_bottom_top =
+custom_bottom_right =
+custom_bottom_bottom =
+
 #Whether to toggle frame limiter on or off.
 # 0: Off , 1  (default): On
 toggle_framelimit =
--- a/src/citra_qt/config.cpp
+++ b/src/citra_qt/config.cpp
@@ -57,6 +57,15 @@ void Config::ReadValues() {
     Settings::values.layout_option =
         static_cast<Settings::LayoutOption>(qt_config->value("layout_option").toInt());
     Settings::values.swap_screen = qt_config->value("swap_screen", false).toBool();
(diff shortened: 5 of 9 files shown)


src/citra_qt/configure_graphics.ui
```diff
@@ -176,6 +181,153 @@
     </widget>
    </item>
    <item>
+    <widget class="QGroupBox" name="custom_groupBox">
+     <property name="title">
+      <string>Custom Layout</string>
+     </property>
+      <layout class="QHBoxLayout" name="horizontalLayout_3">
```
Kloen: C:\projects\citra\src\citra_qt\configure_graphics.ui : warning : The name 'horizontalLayout_3' (QHBoxLayout) is already in use, defaulting to 'horizontalLayout_31'. [C:\projects\citra\build\src\citra_qt\citra-qt.vcxproj]

src/citra_qt/configure_graphics.ui
```diff
@@ -176,6 +181,153 @@
...
+       <item>
+        <layout class="QVBoxLayout" name="verticalLayout_4">
```
Kloen: C:\projects\citra\src\citra_qt\configure_graphics.ui : warning : The name 'verticalLayout_4' (QVBoxLayout) is already in use, defaulting to 'verticalLayout_41'. [C:\projects\citra\build\src\citra_qt\citra-qt.vcxproj]

src/citra_qt/configure_graphics.ui
```diff
@@ -176,6 +181,153 @@
...
+         <item>
+          <widget class="QLabel" name="label">
```
Kloen: C:\projects\citra\src\citra_qt\configure_graphics.ui : warning : The name 'label' (QLabel) is already in use, defaulting to 'label2'. [C:\projects\citra\build\src\citra_qt\citra-qt.vcxproj]

src/citra_qt/configure_graphics.ui
```diff
@@ -176,6 +181,153 @@
...
+           <property name="text">
+            <string>Top Screen:</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <layout class="QHBoxLayout" name="horizontalLayout_7">
+           <item>
+            <widget class="QLabel" name="label">
```
Kloen: Ditto on label warnings.

src/citra_qt/configure_graphics.ui
```diff
@@ -176,6 +181,153 @@
...
+        <layout class="QVBoxLayout" name="verticalLayout_4">
+         <item>
+          <widget class="QLabel" name="label">
+           <property name="text">
+            <string>Top Screen:</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <layout class="QHBoxLayout" name="horizontalLayout_7">
```
Kloen: C:\projects\citra\src\citra_qt\configure_graphics.ui : warning : The name 'horizontalLayout_7' (QHBoxLayout) is already in use, defaulting to 'horizontalLayout_71'. [C:\projects\citra\build\src\citra_qt\citra-qt.vcxproj

src/citra_qt/configure_graphics.ui
jfmherokiller: that onw should be easy to fix

SonofUgly:
Sorry about that, thought I checked the warnings. Fixed and squashed.

wwylele:
Suggestions on the configuration window:

Since the custom layout parameters only works for custom layout, disable these widgets when the screen layout is not custom. Also, it seems that "Swap Screens" doesn't work with custom layout, so either make it work, or disable the checkbox when the layout is custom

SonofUgly:
> Since the custom layout parameters only works for custom layout, disable these widgets when the screen layout is not custom.

How do I set the connection signal for specific options in comboboxes in the .ui file?
> Also, it seems that "Swap Screens" doesn't work with custom layout, so either make it work, or disable the checkbox when the layout is custom

I was going to follow this up with https://github.com/SonofUgly/citra/commit/0854c90b9ca08c28fb402a6b6eb52ed827816183 (wanted to keep this PR as simple as possible), so you can use swap screens to switch which screen gets overlapped ([example](https://my.mixtape.moe/zqhujf.mp4)). Because of the different aspect ratios, using swap screen normally wouldn't work right without making things much more complex.
So should I add that commit to this PR, or just disable the checkbox?

wwylele:
> How do I set the connection signal for specific options in comboboxes?

I don't think there is a way to do this. You can connect to the general option-changing signal (`currentIndexChanged`) and check the selected option there.

SonofUgly:
Addressed above.

src/citra_qt/configure_graphics.cpp
```diff
@@ -14,10 +14,24 @@ ConfigureGraphics::ConfigureGraphics(QWidget* parent)
...
+
+    connect(ui->layout_combobox, SIGNAL(currentIndexChanged(int)), this,
+            SLOT(layoutComboboxChanged(int)));
+    layoutComboboxChanged(ui->layout_combobox->currentIndex());
 }
 
 ConfigureGraphics::~ConfigureGraphics() {}
 
+void ConfigureGraphics::layoutComboboxChanged(int index) {
+    if (index == 3) {
```
vonchenplus: what's the 3 mean? please define macro.

wwylele: For readability, I prefer an `enum` for the items in this combobox

SonofUgly: Changed `currentIndexChanged` to `currentTextChanged`, should make that more clear.

src/citra_qt/configure_graphics.cpp
```diff
@@ -14,10 +14,24 @@ ConfigureGraphics::ConfigureGraphics(QWidget* parent)
...
+
+    connect(ui->layout_combobox, SIGNAL(currentTextChanged(const QString&)), this,
+            SLOT(layoutComboboxChanged(const QString&)));
+    layoutComboboxChanged(ui->layout_combobox->currentText());
 }
 
 ConfigureGraphics::~ConfigureGraphics() {}
 
+void ConfigureGraphics::layoutComboboxChanged(const QString& text) {
+    if (text == "Custom") {
```
wwylele: Using the text is actually worse than before... what if the text get translated into other languages? Note that Qt can do auto translation to the string in ui file

SonofUgly: That's a good point. Having a hard time with making an enum and connecting it, will going back to how it was and adding a comment suffice?

jroweboy:
Okay. So I've finally decided to review this after putting it off for a long time. I understand the motivation for this PR and like the idea behind it, but I personally dislike the execution. Having users define their own boundaries means that they will almost never get the aspect ratio correct, despite probably wanting to. Additionally, having 8 individual boxes means that to customize it to the way you want it, a user needs to "guess and check" until they finally tweak it jussst right. And its not easy to get it right as each time you change one field, you will need to go back and update all of the other affected fields. For example, in a configuration with the two screens side by side and bottom aligned, making one screen slightly taller (moving the top value for the top screen) involves changing the top and left fields for the top screen and then the top and left fields for the bottom screen. To actually maintain the correct aspect ratio means you'll need to recalculate the left value for the top screen and then the top value for the bottom screen every time you want to make the top screen slightly larger. Definitely something thats very easy to get wrong! 

In spite of my objections, I am not opposed to this PR entirely. I personally would vote to remove the options in the GUI and just make it an INI only setting until a better way to set those values in the GUI is written. I would love to hear some more feedback about my suggestion to remove it from the gui though, as I am not sure what others think about this.

yuriks:
fwiw, I agree with @jroweboy's rationale that this is too error-prone to configure. Exposing it in the ini sounds like a good compromise. Don't really have an opinion other than that.

devarion1:
Hello,

Iam very new here and I have no idea how to make this work.
Can anyone help me? I realy want this custom layout feature in my Citra but I don't understand what to download or how to install? If I download this whole Citra Folder here there is no ".exe" in it like in the other citra I downloaded.

I was trying to find tutorials or anything else which shows me how to install this custom layout feature but I can find NOTHING in the whole internet.

Can anyone explain this to a total noob? :(

jroweboy:
@devarion1 please don't spam the issue tracker with help requests. Use the forums for support requests. https://community.citra-emu.org/

I will be deleting both yours and my comments as they are off topic. Open a support request on the forum if you would like further assistance.

PR #2520, opened 2017-02-03
ShaderJIT: add 16 dummy bytes at the bottom of the stack

### LINKED ISSUE #2498: Pokemon Sun/Moon crashes at intro with Shader JIT on
screenshot from 2017-01-30 15-27-49

^ Exactly at this point

OS: Ubuntu 16.04
version: master-ff28080

This should be a regression because IIRC I passed this point with shader JIT on. I am going to bisect it.

Log is just StoreDataCache spam and is useless

wwylele:
Fixes #2498, and potentially other shader JIT crash

Here is a recap of what happens in #2498. It is basically many small things happen at the same time:

First, the shader in question is loaded with some leftover code from a previous shader. When compiling the shader, `FindReturnOffsets` scans the code including the leftover, and generates some garbage return offsets, which are function calls in the leftover to the true part of the shader. This is still OK-ish, because we have [the address check on the potential return](https://github.com/citra-emu/citra/blob/master/src/video_core/shader/shader_jit_x64_compiler.cpp#L774), which should prevent the code from returning on a garbage return offsets.

But the bad thing is, the garbage return address is in shader main() routine. During main() execution, when it meets a garbage return, it tries to access `qword[rsp + 8]` check the address, which, because it is in main(), is outside the shader stack, and is in [the callee-saved register we pushed before shader execution](https://github.com/citra-emu/citra/blob/master/src/video_core/shader/shader_jit_x64_compiler.cpp#L844). So now what it will do all depends on the unknown stack value there. If the value doesn't match the address, shader will continue and everything is fine; if the value unfortunately matches the address, shader will do an unexpected return and crash.

In the case of mine, that stack value is caller's r15, according to system-V ABI. What's worse, after yuriks' shader refactor, gcc decides to put [this loop variable](https://github.com/citra-emu/citra/blob/master/src/video_core/command_processor.cpp#L253) in r15. so the crash will be almost for sure to happen because that loop variable iterates over a wide range, and makes it look like a regression.

The solution here is pretty simple: reserve 16 bytes more spaces at the bottom of the stacks, so that any return address check in main() won't access outside shader stack. In the long term we should fix the control flow analysis, but having this fix first is better than nothing.

### DIFF
--- a/src/video_core/shader/shader_jit_x64_compiler.cpp
+++ b/src/video_core/shader/shader_jit_x64_compiler.cpp
@@ -589,7 +589,7 @@ void JitShader::Compile_RSQ(Instruction instr) {
 void JitShader::Compile_NOP(Instruction instr) {}
 
 void JitShader::Compile_END(Instruction instr) {
-    ABI_PopRegistersAndAdjustStack(*this, ABI_ALL_CALLEE_SAVED, 8);
+    ABI_PopRegistersAndAdjustStack(*this, ABI_ALL_CALLEE_SAVED, 8, 16);
     ret();
 }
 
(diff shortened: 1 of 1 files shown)


jfmherokiller:
is this the reason why it falls through and ends up going into a read mem loop on some applications?

wwylele:
@jfmherokiller if you mean the "Unmapped Read" log spam, then no, because that's guest application that reads wrong memory, not the same as this case where it is the host that reads wrong memory

PR #2530, opened 2017-02-04
Fixed microprofile for MINGW

FernandoS27:


CMakeLists.txt
```diff
@@ -73,7 +73,7 @@ if (NOT MSVC)
     if (MINGW)
         add_definitions(-DMINGW_HAS_SECURE_API)
         # Microprofile causes crashes when launching titles on MinGW
-        add_definitions(-DMICROPROFILE_ENABLED=0)
+        # add_definitions(-DMICROPROFILE_ENABLED=0)
```
yuriks: Just remove this line and the comment now.

jroweboy:
i can't believe it was so simple >.> good catch

Issue #2531, opened 2017-02-05
Use IPC helper & ServiceFramework on our services


wwylele:
Since #2027 is merged, we should get started to apply it to our services. This issue is to record the assignment of each service and the process of them. 

Update: we also need to convert them into new `ServiceFramework` and ditch some old interface of IPCHelpers. Please do not use those `[[deprecated]]` functions (including `Push/PopHandle`) in the new `ServiceFramework`.

We have the following services, neglecting services without any actual implemented/stubbed functions: (bold ones are service that are already implemented a lot and should be taken more care)
- [x] ac (#2992)
- [x] **am**
- [x] **apt** (#2533) (need new service framework) (#3421)
- [x] boss
- [x] **cam** (#2722) (need new service framework) (#3239)
- [x] cecd
- [x] **cfg**
- [x] csnd
- [x] **dsp**
- [x] **err:f** (#3190)
- [x] **frd** (#3469)
- [x] **gsp** (#3307)
- [x] **fs** (#3441)
- [x] **hid** (#3304)
- [x] **ir** (#3248)
- [x] **ldr:ro** (#2532) (need new service framework) (#3192)
- [x] mic (#3253)
- [x] **ndm**
- [x] nfc (#3476)
- [x] nim
- [x] ns
- [x] **nwm** (partially done)
- [x] **ptm** (#2710) (need new service framework)(#3361)
- [x] **soc**
- [x] **srv** (#2774)
- [x] ssl
- [x] **y2r** (#3367)

Please comment below if you are going to take over one from this list.

~~Also note that @Lectem will have follow up PRs to IPC helper itself. Currently these functions are planed to add:~~ Already done

JohnnyPony:
I am currently working on the ac service.

wwylele:
Unresolved problems presented in #2532 : 
Most real system model checks param translation descriptor, and returns `ERROR_INVALID_DESCRIPTOR` on failure. Currently the check is handled by IPCHelper with `DEBUG_ASSERT_MSG`. We need to consider the importance of this (usually game won't hit this error) and think of a proper way to handle it.

tgsm:
#2532 is merged, so I am assuming ldr:ro can be checked off now.

wwylele:
Probably we can wait for #2646, and do IPC helper and log rewrite for each module at the same time

yuriks:
[...]

PR #2532, opened 2017-02-05
ldr_ro: use IPC helper

wwylele:
My first attempt on using IPC helper. I hope I get it right :smile:

### DIFF
--- a/src/core/hle/service/ldr_ro/ldr_ro.cpp
+++ b/src/core/hle/service/ldr_ro/ldr_ro.cpp
@@ -40,9 +40,6 @@ static const ResultCode ERROR_INVALID_MEMORY_STATE = // 0xD8A12C08
 static const ResultCode ERROR_NOT_LOADED = // 0xD8A12C0D
     ResultCode(static_cast<ErrorDescription>(13), ErrorModule::RO, ErrorSummary::InvalidState,
                ErrorLevel::Permanent);
-static const ResultCode ERROR_INVALID_DESCRIPTOR = // 0xD9001830
-    ResultCode(ErrorDescription::OS_InvalidBufferDescriptor, ErrorModule::OS,
-               ErrorSummary::WrongArgument, ErrorLevel::Permanent);
 
 static MemorySynchronizer memory_synchronizer;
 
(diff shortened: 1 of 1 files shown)


src/core/hle/service/ldr_ro/ldr_ro.cpp
```diff
@@ -275,87 +247,78 @@ static void UnloadCRR(Interface* self) {
...
-    }
+              auto_link ? "true" : "false", fix_level, crr_address, process);
 
-    cmd_buff[0] = IPC::MakeHeader(link_on_load_bug_fix ? 9 : 4, 2, 0);
+    IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
 
     if (loaded_crs == 0) {
         LOG_ERROR(Service_LDR, "Not initialized");
-        cmd_buff[1] = ERROR_NOT_INITIALIZED.raw;
+        rb.Push(ERROR_NOT_INITIALIZED, u32(0));
```
lioncash: These can just be `0U` instead of casting. Ditto for all occurrences below.

Lectem: It could be a problem since the U suffix can be int or long or long long, and we don't want to trigger the u64 Push

src/core/hle/service/ldr_ro/ldr_ro.cpp
```diff
@@ -275,87 +247,78 @@ static void UnloadCRR(Interface* self) {
...
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), link_on_load_bug_fix ? 0x09 : 0x04, 11, 2);
+    VAddr cro_buffer_ptr = rp.Pop<u32>();
+    VAddr cro_address = rp.Pop<u32>();
+    u32 cro_size = rp.Pop<u32>();
+    VAddr data_segment_address = rp.Pop<u32>();
+    u32 zero = rp.Pop<u32>();
+    u32 data_segment_size = rp.Pop<u32>();
+    u32 bss_segment_address = rp.Pop<u32>();
+    u32 bss_segment_size = rp.Pop<u32>();
+    bool auto_link = (rp.Pop<u32>() & 0xFF) != 0;
```
Lectem: Should be able to use Pop<bool> once #2533 gets merged

mailwl:
@wwylele you should check it with pokemon, for sure it was the first, users will check

src/core/hle/service/ldr_ro/ldr_ro.cpp
```diff
@@ -579,40 +533,31 @@ static void UnloadCRO(Interface* self) {
...
-    u32 descriptor = cmd_buff[2];
-    u32 process = cmd_buff[3];
-
-    LOG_DEBUG(Service_LDR, "called, cro_address=0x%08X, descriptor=0x%08X, process=0x%08X",
-              cro_address, descriptor, process);
-
-    if (descriptor != 0) {
-        LOG_ERROR(Service_LDR, "IPC handle descriptor failed validation (0x%X)", descriptor);
-        cmd_buff[0] = IPC::MakeHeader(0, 1, 0);
-        cmd_buff[1] = ERROR_INVALID_DESCRIPTOR.raw;
```
Lectem: I didn't know that services could return such errors, it might be better if we changed (or added a new) PopHandle to return a boolean when descriptors don't match ? I don't feel good throwing a debug_assert if the real behaviour should be to ignore it and return an error.

Subv: I didn't know that either, I thought the error would come from the SendSyncRequest call, before it even reached the service code.

wwylele: In the morning I thought I wrote some redundant code so I just removed this lol. But now I recheck the service, this error check does present in the ldr:ro service (and many other services). SendSyncRequest doesn't check this because it doesn't know what kind of translate descriptor the service want - it only errors on an invalid translate descriptor that kernel can't even parse. 

So my idea: add an function to `RequestParser` that checks whether all translate parameter have been successfully parsed, and automatically return this error code when failed. The code is like:
```c++
IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x01, 3, 2);
VAddr crs_buffer_ptr = rp.Pop<u32>();
u32 crs_size = rp.Pop<u32>();
VAddr crs_address = rp.Pop<u32>();
Kernel::Handle process = rp.PopHandle();
if (rp.Failed()) return; // Failed() will do RequestBuilder::push(ERROR_INVALID_DESCRIPTOR) inside on failure
...
```

Lectem: Are there any cases where a service would give more data than just this error ?
Building the error in the parser is a good idea because we can skip validation that way but I'm not sure if we always want to stop parsing on the first error

wwylele: @Lectem I looked about 6 services and all of them do the same thing, and some of them even have some explicit structure like `if(!ValidateHandleDescriptor())return ERROR_INVALID_DESCRIPTOR;` So I think it is safe to assume this applies to all service.

wwylele:
Rebased onto current master. Please review. I leave the descriptor check removed and going to leave a comment in #2531 about this

yuriks:
Agreed with adding TODOs for the missing descriptor verification. Otherwise LGTM

wwylele:
Added comments about the descriptor

PR #2533, opened 2017-02-05
IpcHelper enhancement and APT refactor

Lectem:
This is a follow-up of #2027 and part of #2531.
The IPC helpers code was cleaned and specializations for u8/u16/bool were added, and use Push/PopRaw(memcpy) as an implementation.
The u64 specializations still use bitshifts, I wasn't sure if I should keep it that way or use Push/PopRaw too.
A PeekStaticBuffer method was also added so that we can read values and descriptions of the static buffers.

APT was fully rewritten with the IPC helpers, and ctrulib should be able to remove its hack once this is merged @fincs .
I had to modify PTM a bit too so that I could rewrite APT:CheckNew3DSApp.

### DIFF
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -27,6 +27,18 @@ public:
         DEBUG_ASSERT_MSG(index == TotalSize(), "Operations do not match the header (cmd 0x%x)",
                          header.raw);
     }
+
+    /**
+     * @brief Retrieves the address of a static buffer, used when a buffer is needed for output
+     * @param buffer_id The index of the static buffer
+     * @param data_size If non-null, will store the size of the buffer
+     */
+    VAddr PeekStaticBuffer(u8 buffer_id, size_t* data_size = nullptr) const {
+        u32* static_buffer = cmdbuf + Kernel::kStaticBuffersOffset / 4 + buffer_id * 2;
+        if (data_size)
+            *data_size = StaticBufferDescInfo{static_buffer[0]}.size;
+        return static_buffer[1];
+    }
 };
 
 class RequestBuilder : public RequestHelperBase {
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -119,7 +119,7 @@ inline void RequestBuilder::Push(u64 value) {
 
 template <>
 inline void RequestBuilder::Push(bool value) {
-    Push(u8(value));
+    Push(static_cast<u8>(value));
 }
 
 template <>
@@ -277,7 +277,7 @@ inline u64 RequestParser::Pop() {
 
 template <>
 inline bool RequestParser::Pop() {
-    return Pop<u8>() != 0; // != 0 to remove warning C4800
+    return Pop<u8>() != 0;
 }
 
 template <>
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -62,14 +62,8 @@ public:
     template <typename T>
     void Push(T value);
 
-    void Push(u32 value) {
-        cmdbuf[index++] = value;
-    }
     template <typename First, typename... Other>
-    void Push(const First& first_value, const Other&... other_values) {
-        Push(first_value);
-        Push(other_values...);
-    }
+    void Push(const First& first_value, const Other&... other_values);
 
     /**
      * @brief Copies the content of the given trivially copyable class to the buffer as a normal
@@ -77,57 +71,94 @@ public:
      * @note: The input class must be correctly packed/padded to fit hardware layout.
      */
     template <typename T>
-    void PushRaw(const T& value) {
-        static_assert(std::is_trivially_copyable<T>(), "Raw types should be trivially copyable");
-        std::memcpy(cmdbuf + index, &value, sizeof(T));
-        index += (sizeof(T) + 3) / 4; // round up to word length
-    }
+    void PushRaw(const T& value);
 
     // TODO : ensure that translate params are added after all regular params
     template <typename... H>
-    void PushCopyHandles(H... handles) {
-        Push(CopyHandleDesc(sizeof...(H)));
-        Push(static_cast<Kernel::Handle>(handles)...);
-    }
+    void PushCopyHandles(H... handles);
 
     template <typename... H>
-    void PushMoveHandles(H... handles) {
-        Push(MoveHandleDesc(sizeof...(H)));
-        Push(static_cast<Kernel::Handle>(handles)...);
-    }
+    void PushMoveHandles(H... handles);
 
-    void PushCurrentPIDHandle() {
-        Push(CallingPidDesc());
-        Push(u32(0));
-    }
+    void PushCurrentPIDHandle();
 
-    void PushStaticBuffer(VAddr buffer_vaddr, u32 size, u8 buffer_id) {
-        Push(StaticBufferDesc(size, buffer_id));
-        Push(buffer_vaddr);
-    }
+    void PushStaticBuffer(VAddr buffer_vaddr, u32 size, u8 buffer_id);
 
-    void PushMappedBuffer(VAddr buffer_vaddr, u32 size, MappedBufferPermissions perms) {
-        Push(MappedBufferDesc(size, perms));
-        Push(buffer_vaddr);
-    }
+    void PushMappedBuffer(VAddr buffer_vaddr, u32 size, MappedBufferPermissions perms);
 };
 
 /// Push ///
 
 template <>
-inline void RequestBuilder::Push<u32>(u32 value) {
-    Push(value);
+inline void RequestBuilder::Push(u32 value) {
+    cmdbuf[index++] = value;
+}
+
+template <typename T>
+void RequestBuilder::PushRaw(const T& value) {
+    static_assert(std::is_trivially_copyable<T>(), "Raw types should be trivially copyable");
+    std::memcpy(cmdbuf + index, &value, sizeof(T));
+    index += (sizeof(T) + 3) / 4; // round up to word length
 }
 
 template <>
-inline void RequestBuilder::Push<u64>(u64 value) {
+inline void RequestBuilder::Push(u8 value) {
+    PushRaw(value);
+}
+
+template <>
+inline void RequestBuilder::Push(u16 value) {
+    PushRaw(value);
+}
+
+template <>
+inline void RequestBuilder::Push(u64 value) {
     Push(static_cast<u32>(value));
     Push(static_cast<u32>(value >> 32));
 }
 
 template <>
-inline void RequestBuilder::Push<ResultCode>(ResultCode value) {
+inline void RequestBuilder::Push(bool value) {
+    Push(u8(value));
+}
+
+template <>
+inline void RequestBuilder::Push(ResultCode value) {
     Push(value.raw);
+}
+
+template <typename First, typename... Other>
+void RequestBuilder::Push(const First& first_value, const Other&... other_values) {
+    Push(first_value);
+    Push(other_values...);
+}
+
+template <typename... H>
+inline void RequestBuilder::PushCopyHandles(H... handles) {
+    Push(CopyHandleDesc(sizeof...(H)));
+    Push(static_cast<Kernel::Handle>(handles)...);
+}
+
+template <typename... H>
+inline void RequestBuilder::PushMoveHandles(H... handles) {
+    Push(MoveHandleDesc(sizeof...(H)));
+    Push(static_cast<Kernel::Handle>(handles)...);
+}
+
+inline void RequestBuilder::PushCurrentPIDHandle() {
+    Push(CallingPidDesc());
+    Push(u32(0));
+}
+
+inline void RequestBuilder::PushStaticBuffer(VAddr buffer_vaddr, u32 size, u8 buffer_id) {
+    Push(StaticBufferDesc(size, buffer_id));
+    Push(buffer_vaddr);
+}
+
+inline void RequestBuilder::PushMappedBuffer(VAddr buffer_vaddr, u32 size,
+                                             MappedBufferPermissions perms) {
+    Push(MappedBufferDesc(size, perms));
+    Push(buffer_vaddr);
 }
 
 class RequestParser : public RequestHelperBase {
@@ -197,24 +228,60 @@ public:
      */
     template <typename T>
     void PopRaw(T& value);
+
+    /**
+     * @brief Reads the next normal parameters as a struct, by copying it into a new value
+     * @note: The output class must be correctly packed/padded to fit hardware layout.
+     */
+    template <typename T>
+    T PopRaw();
 };
 
 /// Pop ///
 
 template <>
-inline u32 RequestParser::Pop<u32>() {
+inline u32 RequestParser::Pop() {
     return cmdbuf[index++];
 }
 
+template <typename T>
+void RequestParser::PopRaw(T& value) {
+    static_assert(std::is_trivially_copyable<T>(), "Raw types should be trivially copyable");
+    std::memcpy(&value, cmdbuf + index, sizeof(T));
+    index += (sizeof(T) + 3) / 4; // round up to word length
+}
+
+template <typename T>
+T RequestParser::PopRaw() {
+    T value;
+    PopRaw(value);
+    return value;
+}
+
 template <>
-inline u64 RequestParser::Pop<u64>() {
+inline u8 RequestParser::Pop() {
+    return PopRaw<u8>();
+}
+
+template <>
+inline u16 RequestParser::Pop() {
+    return PopRaw<u16>();
+}
+
+template <>
+inline u64 RequestParser::Pop() {
     const u64 lsw = Pop<u32>();
     const u64 msw = Pop<u32>();
     return msw << 32 | lsw;
 }
 
 template <>
-inline ResultCode RequestParser::Pop<ResultCode>() {
+inline bool RequestParser::Pop() {
+    return Pop<u8>() != 0; // != 0 to remove warning C4800
+}
+
+template <>
+inline ResultCode RequestParser::Pop() {
     return ResultCode{Pop<u32>()};
 }
 
@@ -275,13 +342,6 @@ inline VAddr RequestParser::PopMappedBuffer(size_t* data_size,
     if (buffer_perms != nullptr)
         *buffer_perms = bufferInfo.perms;
     return Pop<VAddr>();
-}
-
-template <typename T>
-void RequestParser::PopRaw(T& value) {
-    static_assert(std::is_trivially_copyable<T>(), "Raw types should be trivially copyable");
-    std::memcpy(&value, cmdbuf + index, sizeof(T));
-    index += (sizeof(T) + 3) / 4; // round up to word length
 }
 
 } // namespace IPC
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -28,13 +28,19 @@ public:
                          header.raw);
     }
 
+    void Skip(unsigned size_in_words, bool set_to_null) {
+        if (set_to_null)
+            memset(cmdbuf + index, 0, size_in_words * sizeof(u32));
+        index += size_in_words;
+    }
+
     /**
      * @brief Retrieves the address of a static buffer, used when a buffer is needed for output
      * @param buffer_id The index of the static buffer
      * @param data_size If non-null, will store the size of the buffer
      */
     VAddr PeekStaticBuffer(u8 buffer_id, size_t* data_size = nullptr) const {
-        u32* static_buffer = cmdbuf + Kernel::kStaticBuffersOffset / 4 + buffer_id * 2;
+        u32* static_buffer = cmdbuf + Kernel::kStaticBuffersOffset / sizeof(u32) + buffer_id * 2;
         if (data_size)
             *data_size = StaticBufferDescInfo{static_buffer[0]}.size;
         return static_buffer[1];
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -279,9 +279,8 @@ void CancelParameter(Service::Interface* self) {
     rb.Push(RESULT_SUCCESS); // No error
     rb.Push(true);           // Set to Success
 
-    LOG_WARNING(Service_APT,
-                "(STUBBED) called check_sender=0x%08X, sender_appid=0x%08X, "
-                "check_receiver=0x%08X, receiver_appid=0x%08X",
+    LOG_WARNING(Service_APT, "(STUBBED) called check_sender=0x%08X, sender_appid=0x%08X, "
+                             "check_receiver=0x%08X, receiver_appid=0x%08X",
                 check_sender, sender_appid, check_receiver, receiver_appid);
 }
 
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -231,7 +231,7 @@ void ReceiveParameter(Service::Interface* self) {
     rb.PushMoveHandles((next_parameter.object != nullptr)
                            ? Kernel::g_handle_table.Create(next_parameter.object).MoveFrom()
                            : 0);
-    rb.PushStaticBuffer(buffer, next_parameter.buffer.size(), 0);
+    rb.PushStaticBuffer(buffer, static_cast<u32>(next_parameter.buffer.size()), 0);
 
     Memory::WriteBlock(buffer, next_parameter.buffer.data(), next_parameter.buffer.size());
 
@@ -261,7 +261,7 @@ void GlanceParameter(Service::Interface* self) {
     rb.PushCopyHandles((next_parameter.object != nullptr)
                            ? Kernel::g_handle_table.Create(next_parameter.object).MoveFrom()
                            : 0);
-    rb.PushStaticBuffer(buffer, next_parameter.buffer.size(), 0);
+    rb.PushStaticBuffer(buffer, static_cast<u32>(next_parameter.buffer.size()), 0);
 
     Memory::WriteBlock(buffer, next_parameter.buffer.data(), next_parameter.buffer.size());
 
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -69,10 +69,11 @@ void Initialize(Service::Interface* self) {
 
 void GetSharedFont(Service::Interface* self) {
     IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x44, 0, 0); // 0x00440000
+    IPC::RequestBuilder rb = rp.MakeBuilder(2, 2);
     if (!shared_font_mem) {
         LOG_ERROR(Service_APT, "shared font file missing - go dump it from your 3ds");
-        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
         rb.Push<u32>(-1); // TODO: Find the right error code
+        rb.Skip(1 + 2, true);
         return;
     }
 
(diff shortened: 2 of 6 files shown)


mailwl:
i have many questions about APT, but it's not a place where refactor?

mailwl:
Lectem, may you at least check output parameters, which don't provided already?

src/core/hle/ipc_helpers.h
```diff
@@ -197,24 +228,60 @@ class RequestParser : public RequestHelperBase {
...
+inline u64 RequestParser::Pop() {
     const u64 lsw = Pop<u32>();
     const u64 msw = Pop<u32>();
     return msw << 32 | lsw;
 }
 
 template <>
-inline ResultCode RequestParser::Pop<ResultCode>() {
+inline bool RequestParser::Pop() {
+    return Pop<u8>() != 0; // != 0 to remove warning C4800
```
lioncash: This comment isn't necessary. It's how the type system works. A function with the return type `bool` expects a `bool` return value (makes sense, right?). A `u8` is not a `bool`, hence the warning. `!= 0` makes it a bool as expected.

src/core/hle/ipc_helpers.h
```diff
@@ -62,74 +62,105 @@ class RequestBuilder : public RequestHelperBase {
...
-inline void RequestBuilder::Push<u64>(u64 value) {
+inline void RequestBuilder::Push(u64 value) {
     Push(static_cast<u32>(value));
     Push(static_cast<u32>(value >> 32));
 }
 
 template <>
-inline void RequestBuilder::Push<ResultCode>(ResultCode value) {
+inline void RequestBuilder::Push(bool value) {
+    Push(u8(value));
```
lioncash: `static_cast`

src/core/hle/service/y2r_u.cpp
```diff
@@ -189,11 +189,9 @@ static void SetSpacialDithering(Interface* self) {
...
  */
 static void GetSpacialDithering(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-
-    cmd_buff[0] = IPC::MakeHeader(0xA, 2, 0);
-    cmd_buff[1] = RESULT_SUCCESS.raw;
-    cmd_buff[2] = spacial_dithering_enabled;
+    IPC::RequestBuilder rb(Kernel::GetCommandBuffer(), 0xA, 2, 0);
+    rb.Push(RESULT_SUCCESS);
+    rb.Push(bool(spacial_dithering_enabled));
```
lioncash: `spacial_dithering_enabled != 0`

Lectem:
> you at least check output parameters, which don't provided already?

@mailwl What output parameters are you talking about exactly ?

mailwl:
https://github.com/citra-emu/citra/blob/master/src/core/hle/service/apt/apt.cpp#L205 dont't have 7 and 8 parameters in current firmware
it's just an example, the whole APT need to be revieved again

Lectem:
Where did you find this information ? 3dbrew still has those parameters, but anyway this indeed wasn't the place to discuss this as it is beyond this PR scope. This PR only updates the current code to use the helpers and fix a few things, perhaps you should add an issue about your problem.

mailwl:
my 3ds dont do this (real hardware)
you right, this PR not a reason to fix other problems

src/core/hle/service/apt/apt.cpp
```diff
@@ -62,18 +62,15 @@ void Initialize(Service::Interface* self) {
...
 void GetSharedFont(Service::Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x44, 0, 0); // 0x00440000
     if (!shared_font_mem) {
         LOG_ERROR(Service_APT, "shared font file missing - go dump it from your 3ds");
-        cmd_buff[0] = IPC::MakeHeader(0x44, 2, 2);
-        cmd_buff[1] = -1; // TODO: Find the right error code
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(u32(-1)); // Failure
```
Subv: Please keep the TODO, returning -1 here is not correct. Likewise if you removed any other similar TODOs

src/core/hle/service/apt/apt.cpp
```diff
@@ -203,78 +209,93 @@ void SendParameter(Service::Interface* self) {
...
-    if (next_parameter.object != nullptr)
-        cmd_buff[6] = Kernel::g_handle_table.Create(next_parameter.object).MoveFrom();
-    cmd_buff[7] = (next_parameter.buffer.size() << 14) | 2;
-    cmd_buff[8] = buffer;
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0xD, 2, 0); // 0xD0080
+    u32 app_id = rp.Pop<u32>();
+    u32 buffer_size = rp.Pop<u32>();
+
+    size_t static_buff_size;
+    VAddr buffer = rp.PeekStaticBuffer(0, &static_buff_size);
```
wwylele: IMO this `PeekStaticBuffer`, alone with `PushStaticBuffer and `Memory::WriteBlock` below, should be something like
```
std::vector buffer_data;
rb.SendStaticBuffer(data, 0/*index*/);
```
The static buffer reply routine in real 3DS is:
The system module (equivalent to apt.cpp) put the reply data to a own buffer (equivalent to `std::vector buffer_data`), and push static buffer descriptor with the address (in the service memory space) to command buffer (equivalent to `PushStaticBuffer`), then call `svcReplyAndReceive`. Then the kernel matches the static buffer descriptor of client (equivalent to `PeekStaticBuffer`) with the descriptor in the reply, and transfer the data (equivalent to `Memory::WriteBlock`). Note that the kernel part is common for all such static buffer transfer, and currently our IPC helper is somehow at the level of kernel, so it should integrate the same thing into itself

src/core/hle/service/apt/apt.cpp
```diff
@@ -85,103 +82,112 @@ void GetSharedFont(Service::Interface* self) {
...
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    cmd_buff[2] = 0;
-    cmd_buff[3] = 0;
-    cmd_buff[4] = static_cast<u32>(AppletId::HomeMenu);    // Home menu AppID
-    cmd_buff[5] = static_cast<u32>(AppletId::Application); // TODO(purpasmart96): Do this correctly
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x5, 1, 0); // 0x50040
+    u32 unk = rp.Pop<u32>();
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+    rb.Push(RESULT_SUCCESS); // No error
+    rb.Push(u32(0));
```
wwylele: tbh I begin to dislike this kind of `u32(...)` thing. Probably its better to have some wrap functions like `Push32` ?

src/core/hle/service/apt/apt.cpp
```diff
@@ -85,103 +82,112 @@ void GetSharedFont(Service::Interface* self) {
...
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x9, 1, 0); // 0x90040
+    u32 app_id = rp.Pop<u32>();
+    IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
+    rb.Push(RESULT_SUCCESS); // No error
 
     // TODO(Subv): An application is considered "registered" if it has already called APT::Enable
     // handle this properly once we implement multiprocess support.
-    cmd_buff[2] = 0; // Set to not registered by default
+    u32 isRegistered = 0; // Set to not registered by default
```
wwylele: probably should be a bool

yuriks:
(From #2560 by @Schplee)

When the player selects a save file to load/create, the game soft locks indefinitely in BE-196, 194, and 192.
BE-190 works normally, and the only difference between the two builds is #2533 (apart from changes to #2497 which I highly doubt cause this regression). #2533 isn't merged with BE-190.

Here's a screenshot of what it looks like (the Applet messages repeat):

mgcx3pc

src/core/hle/service/apt/apt.cpp
Lectem: Perhaps use Push<u32>(...) then ? there's no need to duplicate code imo

src/core/hle/service/apt/apt.cpp
Lectem: I like the idea, that way we can also move the Memory::WriteBlock out of the helper once we get multiprocess and IPC translation done by the kernel.
It will also mean less work on the services that already use it.
How should we handle errors though ? are there generic errors when the size of the input is too big for the allocated static buffer ? Is it the only error a service needs to check when sending data through static buffers ?

Lectem:
@Schplee @yuriks Found the issue, I somehow pushed some values as u64 while it should have been bool/u32, I'll double check for other errors

yuriks:
These issues and other things mentioned in the PR just make me think "I told you so" about that request to make the integer Pushes explicitly named instead of overloads/specializations. :P

Lectem:
Well, in this case I explicitly casted to u64, probably from a bad copy paste that I forgot to edit :/

src/core/hle/service/apt/apt.cpp
wwylele: >are there generic errors when the size of the input is too big for the allocated static buffer ? Is it the only error a service needs to check when sending data through static buffers ?

This kind of error is not checked by the service (because actual service doesn't read client command_buffer+0x100 at all). The error is checked by the kernel, and yes the error code is generic (I don't know the value though). There are other kind of errors like invalid address, which are also checked by the kernel and produce generic errors. Service doesn't check any thing here, so it is pretty safe to say any possible error code is generic

src/core/hle/service/apt/apt.cpp
```diff
@@ -62,18 +62,15 @@ void Initialize(Service::Interface* self) {
...
 
 void GetSharedFont(Service::Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x44, 0, 0); // 0x00440000
     if (!shared_font_mem) {
         LOG_ERROR(Service_APT, "shared font file missing - go dump it from your 3ds");
-        cmd_buff[0] = IPC::MakeHeader(0x44, 2, 2);
-        cmd_buff[1] = -1; // TODO: Find the right error code
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
```
wwylele: Actually the original code with (2,2) parameters looks more correct to me. Though I didn't looked into APT service, I have encountered it in many other services that the header is made to a full parameters length even when only an error code is returned and the rest of response is garbage value. This is not that important, though, because the header is rarely checked. If you want the ultimate accuracy, you need specify a (2,2) parameter here and implement a `Skip` function to `RequestBuilder`. If you don't want to bother to do that, probably you can just leave a comment in the code.

src/core/hle/service/apt/apt.cpp
```diff
@@ -85,103 +82,112 @@ void GetSharedFont(Service::Interface* self) {
...
-    cmd_buff[4] = IPC::CopyHandleDesc();
-    cmd_buff[5] = Kernel::g_handle_table.Create(lock).MoveFrom();
-
-    LOG_WARNING(Service_APT, "(STUBBED) called handle=0x%08X applet_attributes=0x%08X", cmd_buff[5],
+    u32 applet_attributes = rp.Pop<u32>();
+    IPC::RequestBuilder rb = rp.MakeBuilder(3, 2);
+    rb.Push(RESULT_SUCCESS);    // No error
+    rb.Push(applet_attributes); // Applet Attributes, this value is passed to Enable.
+    rb.Push<u32>(0);            // Least significant bit = power button state
+    Kernel::Handle handleCopy = Kernel::g_handle_table.Create(lock).MoveFrom();
```
wwylele: `handle_copy`

src/core/hle/service/apt/apt.cpp
```diff
@@ -85,103 +82,112 @@ void GetSharedFont(Service::Interface* self) {
...
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 unk = cmd_buff[1];
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    cmd_buff[2] = 0;
-    cmd_buff[3] = 0;
-    cmd_buff[4] = static_cast<u32>(AppletId::HomeMenu);    // Home menu AppID
-    cmd_buff[5] = static_cast<u32>(AppletId::Application); // TODO(purpasmart96): Do this correctly
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x5, 1, 0); // 0x50040
+    u32 unk = rp.Pop<u32>();
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
```
wwylele: `MakeBuilder(5,0)`

src/core/hle/service/apt/apt.cpp
```diff
@@ -85,103 +82,112 @@ void GetSharedFont(Service::Interface* self) {
...
+    rb.Push(RESULT_SUCCESS); // No error
 
     // TODO(Subv): An application is considered "registered" if it has already called APT::Enable
     // handle this properly once we implement multiprocess support.
-    cmd_buff[2] = 0; // Set to not registered by default
+    bool isRegistered = false; // Set to not registered by default
 
     if (app_id == static_cast<u32>(AppletId::AnyLibraryApplet)) {
-        cmd_buff[2] = HLE::Applets::IsLibraryAppletRunning() ? 1 : 0;
+        isRegistered = HLE::Applets::IsLibraryAppletRunning() ? true : false;
```
wwylele: just `isRegistered = HLE::Applets::IsLibraryAppletRunning();`

src/core/hle/service/apt/apt.cpp
```diff
@@ -85,103 +82,112 @@ void GetSharedFont(Service::Interface* self) {
...
-    u32 handle = cmd_buff[6];
-    u32 size = cmd_buff[7];
-    u32 buffer = cmd_buff[8];
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0xC, 4, 4); // 0xC0104
+    u32 src_app_id = rp.Pop<u32>();
+    u32 dst_app_id = rp.Pop<u32>();
+    u32 signal_type = rp.Pop<u32>();
+    u32 buffer_size = rp.Pop<u32>();
+    u32 value = rp.Pop<u32>();
+    u32 handle = rp.Pop<u32>();
```
wwylele: These two should be `PopHandle`

src/core/hle/service/apt/apt.cpp
```diff
@@ -85,103 +82,112 @@ void GetSharedFont(Service::Interface* self) {
...
+    u32 size = rp.Pop<u32>();
+    u32 buffer = rp.Pop<u32>();
```
wwylele: These two should be `PopStaticBuffer`

src/core/hle/service/apt/apt.cpp
```diff
@@ -203,78 +209,93 @@ void SendParameter(Service::Interface* self) {
...
 }
 
 void PrepareToStartApplication(Service::Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 title_info1 = cmd_buff[1];
-    u32 title_info2 = cmd_buff[2];
-    u32 title_info3 = cmd_buff[3];
-    u32 title_info4 = cmd_buff[4];
-    u32 flags = cmd_buff[5];
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x00150140);
```
wwylele: `IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x15, 5, 0);`

src/core/hle/service/apt/apt.cpp
```diff
@@ -283,16 +304,17 @@ void PrepareToStartApplication(Service::Interface* self) {
...
-    u32 buffer1_size = cmd_buff[1];
-    u32 buffer2_size = cmd_buff[2];
-    u32 flag = cmd_buff[3];
-    u32 size1 = cmd_buff[4];
-    u32 buffer1_ptr = cmd_buff[5];
-    u32 size2 = cmd_buff[6];
-    u32 buffer2_ptr = cmd_buff[7];
-
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x001B00C4);
```
wwylele: `IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1B, 3, 4);`

src/core/hle/service/apt/apt.cpp
```diff
@@ -283,16 +304,17 @@ void PrepareToStartApplication(Service::Interface* self) {
...
+    u32 buffer1_size = rp.Pop<u32>();
+    u32 buffer2_size = rp.Pop<u32>();
+    u32 flag = rp.Pop<u32>();
+    u32 size1 = rp.Pop<u32>();
+    u32 buffer1_ptr = rp.Pop<u32>();
```
wwylele: These two should be `PopStaticBuffer`

src/core/hle/service/apt/apt.cpp
```diff
@@ -283,16 +304,17 @@ void PrepareToStartApplication(Service::Interface* self) {
...
+    u32 size2 = rp.Pop<u32>();
+    u32 buffer2_ptr = rp.Pop<u32>();
```
wwylele: ditto

src/core/hle/service/apt/apt.cpp
```diff
@@ -301,154 +323,170 @@ void StartApplication(Service::Interface* self) {
 }
 
 void AppletUtility(Service::Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x004B00C2);
```
wwylele: IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x4B, 3, 2);

src/core/hle/service/apt/apt.cpp
```diff
@@ -301,154 +323,170 @@ void StartApplication(Service::Interface* self) {
...
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 exiting = cmd_buff[1] & 0xFF;
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x3B, 1, 0); // 0x003B0040
+    bool exiting = rp.Pop<bool>();
 
-    cmd_buff[1] = 1; // TODO: Find the return code meaning
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+    rb.Push<u32>(1);
 
     LOG_WARNING(Service_APT, "(STUBBED) called exiting=%u", exiting);
```
wwylele: since the type of exiting is changed, this log entry need to update. (I forgot to check this for previous functions. Please check yourself if there are similar cases)

src/core/hle/service/apt/apt.cpp
```diff
@@ -458,38 +496,45 @@ void GetStartupArgument(Service::Interface* self) {
...
     if (addr && parameter_size) {
         Memory::ZeroBlock(addr, parameter_size);
     }
 
     LOG_WARNING(Service_APT, "(stubbed) called startup_argument_type=%u , parameter_size=0x%08x",
                 startup_argument_type, parameter_size);
 
-    cmd_buff[1] = RESULT_SUCCESS.raw;
-    cmd_buff[2] = 0;
+    IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
```
wwylele: Should be `(2,2)`. There is a static buffer transfer from service to client, so service should `PushStaticBuffer`.

src/core/hle/service/ptm/ptm.cpp
```diff
@@ -92,21 +92,25 @@ void GetSoftwareClosedFlag(Service::Interface* self) {
...
-    cmd_buff[1] = RESULT_SUCCESS.raw;
-    cmd_buff[2] = is_new_3ds ? 1 : 0;
+    rb.Push(RESULT_SUCCESS);
+    rb.Push(is_new_3ds);
 
     LOG_WARNING(Service_PTM, "(STUBBED) called isNew3DS = 0x%08x", static_cast<u32>(is_new_3ds));
 }
 
+void CheckNew3DS(Service::Interface* self) {
+    IPC::RequestBuilder rb(Kernel::GetCommandBuffer(), 0x040A0000);
```
wwylele: IPC::RequestBuilder rb(Kernel::GetCommandBuffer(), '0x40A', 0, 0);

src/core/hle/ipc_helpers.h
```diff
@@ -27,6 +27,18 @@ class RequestHelperBase {
...
                          header.raw);
     }
+
+    /**
+     * @brief Retrieves the address of a static buffer, used when a buffer is needed for output
+     * @param buffer_id The index of the static buffer
+     * @param data_size If non-null, will store the size of the buffer
+     */
+    VAddr PeekStaticBuffer(u8 buffer_id, size_t* data_size = nullptr) const {
+        u32* static_buffer = cmdbuf + Kernel::kStaticBuffersOffset / 4 + buffer_id * 2;
```
mailwl: magic numbers

src/core/hle/service/apt/apt.cpp
```diff
@@ -85,103 +82,112 @@ void GetSharedFont(Service::Interface* self) {
...
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x9, 1, 0); // 0x90040
+    u32 app_id = rp.Pop<u32>();
+    IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
+    rb.Push(RESULT_SUCCESS); // No error
 
     // TODO(Subv): An application is considered "registered" if it has already called APT::Enable
     // handle this properly once we implement multiprocess support.
-    cmd_buff[2] = 0; // Set to not registered by default
+    bool isRegistered = false; // Set to not registered by default
```
wwylele: `is_registered`

src/core/hle/service/apt/apt.cpp
```diff
@@ -195,88 +200,104 @@ void SendParameter(Service::Interface* self) {
...
-    cmd_buff[8] = buffer;
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0xD, 2, 0); // 0xD0080
+    u32 app_id = rp.Pop<u32>();
+    u32 buffer_size = rp.Pop<u32>();
+
+    size_t static_buff_size;
+    VAddr buffer = rp.PeekStaticBuffer(0, &static_buff_size);
+    if (buffer_size > static_buff_size)
+        LOG_WARNING(Service_APT,
+                    "ReceiveParameter: buffer_size is bigger than the size in the "
```
wwylele: nit: no need to log "ReceiveParameter" because it is already logged by the logging system.

src/core/hle/service/apt/apt.cpp
```diff
@@ -195,88 +200,104 @@ void SendParameter(Service::Interface* self) {
...
-    cmd_buff[8] = buffer;
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0xE, 2, 0); // 0xE0080
+    u32 app_id = rp.Pop<u32>();
+    u32 buffer_size = rp.Pop<u32>();
+
+    size_t static_buff_size;
+    VAddr buffer = rp.PeekStaticBuffer(0, &static_buff_size);
+    if (buffer_size > static_buff_size)
+        LOG_WARNING(Service_APT,
+                    "ReceiveParameter: buffer_size is bigger than the size in the "
```
wwylele: ditto

src/core/hle/service/apt/apt.cpp
```diff
@@ -285,172 +306,189 @@ void PrepareToStartApplication(Service::Interface* self) {
...
 
 void CancelLibraryApplet(Service::Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 exiting = cmd_buff[1] & 0xFF;
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x3B, 1, 0); // 0x003B0040
+    bool exiting = rp.Pop<bool>();
 
-    cmd_buff[1] = 1; // TODO: Find the return code meaning
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+    rb.Push<u32>(1);
```
wwylele: nit: keep the TODO

src/core/hle/service/apt/apt.cpp
```diff
@@ -285,172 +306,189 @@ void PrepareToStartApplication(Service::Interface* self) {
...
     if (auto applet = HLE::Applets::Applet::Get(app_id)) {
         // TODO(Subv): Get the title id for the current applet and write it in the response[2-3]
-        cmd_buff[1] = RESULT_SUCCESS.raw;
-        cmd_buff[4] = static_cast<u32>(Service::FS::MediaType::NAND);
-        cmd_buff[5] = 1; // Registered
-        cmd_buff[6] = 1; // Loaded
-        cmd_buff[7] = 0; // Applet Attributes
+        IPC::RequestBuilder rb = rp.MakeBuilder(7, 0);
+        rb.Push(RESULT_SUCCESS);
+        u64 titileId = 0;
```
wwylele: nit: `title_id`

src/core/hle/service/apt/apt.cpp
```diff
@@ -285,172 +306,189 @@ void PrepareToStartApplication(Service::Interface* self) {
...
 }
 
 void GetStartupArgument(Service::Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 parameter_size = cmd_buff[1];
-    StartupArgumentType startup_argument_type = static_cast<StartupArgumentType>(cmd_buff[2]);
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x51, 2, 0); // 0x00510080
+    u32 parameter_size = rp.Pop<u32>();
+    StartupArgumentType startup_argument_type =
+        static_cast<StartupArgumentType>(rp.Pop<u32>() & 0xF);
```
wwylele: any particular reason to `& 0xF` instead of `& 0xFF` (in which case it can be just `Pop<u8>()`) ?

src/core/hle/service/apt/apt.cpp
```diff
@@ -64,18 +64,16 @@ void Initialize(Service::Interface* self) {
...
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x44, 0, 0); // 0x00440000
+    IPC::RequestBuilder rb = rp.MakeBuilder(2, 2);
     if (!shared_font_mem) {
         LOG_ERROR(Service_APT, "shared font file missing - go dump it from your 3ds");
-        cmd_buff[0] = IPC::MakeHeader(0x44, 2, 2);
-        cmd_buff[1] = -1; // TODO: Find the right error code
+        rb.Push<u32>(-1); // TODO: Find the right error code
+        rb.Skip(1 + 2, true);
```
Lectem: Skip is used here

src/core/hle/service/apt/apt.cpp
Lectem: whoops

src/core/hle/service/apt/apt.cpp
Lectem: right, forgot this one

src/core/hle/service/apt/apt.cpp
Lectem: Not really, I'll change it to `Pop<u8>`

src/core/hle/service/apt/apt.cpp
Lectem: Perhaps create an issue for this and implement it later ?

Issue #2238, opened 2016-11-30
Kid Icarus Uprising - Hangs after saving


JohnGodgames:
Tested on https://github.com/citra-emu/citra/commit/cf5b8483fc53ca3b4919d231045348625a507d05 .

First boot [LOG](http://pastebin.com/9ZJ67fTv)

bottom screen is clickable:
![kid 1](https://cloud.githubusercontent.com/assets/11453507/20746658/f8ba6e58-b6e6-11e6-9cf7-79509cd00ff8.PNG)

Second boot [LOG](http://pastebin.com/MqP7ji78)

screen:
![kid 2](https://cloud.githubusercontent.com/assets/11453507/20746662/fdd03a08-b6e6-11e6-8125-7fdce35d989e.PNG)

After saving spams with:
```
[ 138.420488] Service <Error> core\hle\service\service.cpp:Service::Interface::SyncRequest:74: unknown / unimplemented function 'ReadSavedData': port=cecd:u, cmd_buff[1]=0x60, cmd_buff[2]=0x30000, cmd_buff[3]=0x2, cmd_buff[4]=0x12, cmd_buff[5]=0x20, cmd_buff[6]=0x0, cmd_buff[7]=0x60C, cmd_buff[8]=0xE014D28
[ 138.420979] Service.NDM <Warning> core\hle\service\ndm\ndm.cpp:Service::NDM::ResumeDaemons:102: (STUBBED) daemon_bit_mask=0x0000000F
```

Subv:
Another game that loops trying to access the CECD savedata, just like Tomodachi Life

mailwl:
to pass this loop, just stub ReadSavedData, which return error (-1) for example

SigmaVirus:
@mailwl What do you mean by stubbing ReadSavedData?

I can't get this game to work at all and experiencing the same as @JohnGodgames 

I have tried the latest nightly build of Citra and the bleeding-edge-190 build and it's still the same.

Have seen videos on YouTube demonstrating this game on Citra so it must work somehow.

SigmaVirus:
OK, I created some pull requests for this (#2535, #2526 and #2537)
Sorry for the clutter, but I didn't know how to change multiple files at once.

SigmaVirus:
Created a proper pull request for this #2538 according to what @mailwl said about stubbing.

mailwl:
i dislike workaround in codebase, @wwylele what do you think about it? maybe as temporary hack is allowed,

[...]

PR #2538, opened 2017-02-05
Potential fix for #2238 Kid Icarus Uprising

SigmaVirus:
According to @mailwl ReadSavedData needs to be stubbed in order to pass the loop in Kid Icarus.


src/core/hle/service/cecd/cecd.cpp
```diff
@@ -42,6 +42,13 @@ void GetChangeStateEventHandle(Service::Interface* self) {
 
     LOG_WARNING(Service_CECD, "(STUBBED) called");
 }
+    
+void OpenAndRead(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = -1;
+    LOG_WARNING(Service_CECD, "(STUBBED) called");
+}
```
jroweboy: Trailing whitespace

mailwl:
it's mostly the same as @wwylele hack with https://github.com/citra-emu/citra/blob/master/src/core/hle/service/nwm/nwm_uds.cpp#L91
not implemented yet, so return error. I dislike it too, but this will be fixed, when @wwylele implement cecd

Subv:
That one at least has documentation and a sensible error code.

mailwl:
true :)
CEC will not work, if WiFi is off, but "-1" error includes "Fatal" i think, it will be good to return proper code, not to break other games

SigmaVirus:
OK I get it, it's a bad and awful cursed hack to get a game running.

But I just followed what @mailwl said in issue #2238 which he implemented in his own fork (https://github.com/mailwl/citra/tree/kid/src/core/hle/service/cecd) and thought it would solve the issue.

If you don't like this and feel it breaks more than it solves I could close the issue, but I just wanted to help out instead of just complaining about it.

mailwl:
it's not bad itself i think, because Kid Icarus really cool to be emulated by Citra, but incompleted. i did just workaround, but Citra needs proper code

wwylele:
Just want to correct what @mailwl said: CEC does work even if wifi is off. It allows game to check the mailbox offline.

mailwl:
SigmaVirus, did you change error core to prorer one? looks like stop PR from merging

SigmaVirus:
@mailwl 
While it's true that I'm an educated systems developer and database designer I got literally 0 experience in coding emulators, let alone C++

I really don't know how to change the error code to a proper one and I'm sorry but I didn't think this would be such a big problem when I made this PR.

I have been able to try these changes out practically so I know they work.
The game starts, can be played, saved etc. and I haven't seen any other games suffer from these changes (at least not from my collection)

It would be nice to solve this properly however, and I would in a heartbeat if I could.

Kid Icarus Uprising is a really cool game and it would be nice having all my games in the collection bootable and playable but since I don't have what it takes to make the changes I see three options here:
1. Close this PR
2. Merge it and see it as a temporary workaround that needs to be corrected
3. Someone else more experienced must help me and take over and finish it properly

mailwl:
i have a loader to 3ds https://github.com/mailwl/IDA_plugin_3ds for IDA, but it's without signatures. it can do some help, write me on my mail mailwl@gmail.com and i send this strongly profibited file

mailwl:
ah, the games itself dont have  proper error codes, it just does left SHL to 31 bits, to leave only -1

SigmaVirus:
It's unlikely this is going somewhere right now so I'll just close this because I don't have what it takes to fix it.
Maybe someone with more expertise will pick this up and solve it.

PR #2540, opened 2017-02-06
cleardnscache function in httpc implemented

jfmherokiller:
I stubbed the cleardnscache function in HTTP:C
I also added the http service enumeration value to the log.h file


mailwl:
+1, this stub absolutely equal for unrealized functions (except red string in log)

Subv:
And what would be the point of stubbing this? I see no benefit to it, and it just clutters the codebase.

jfmherokiller:
well i wasnt sure if i should implement it or not since clearing the dns cache could be bad

jfmherokiller:
I now provided clear dns cache code for all platforms

mailwl:
wow, wow! Please don't include this in Citra code! Emulator not allowed to change host

wwylele:
...nope, you probably missed the point of emulation. It is most likely that this function will be implemented in a container that does not interfere the host system. And what's more important, how this is going to be implemented should be decided by who implement the http service (or at least the main part of it). Just picking this single function and implement it with any method doesn't make sense to me.

jfmherokiller:
ok ill remove the changes i was just trying to provide an implementation

mailwl:
@jfmherokiller why do you create whis PR? some game disturbing with log errors?

jfmherokiller:
@mailwl no i just wanted to provide a possible implementation

mailwl:
well, good reason, but why exactly this function? Initialize goes first?

mailwl:
httpc need to be realized, but it tons of work(for all 3 OSes)

PR #2546, opened 2017-02-09
added a refresh game directory option to the file menu

jfmherokiller:


src/citra_qt/main.ui
```diff
@@ -45,7 +45,7 @@
      <x>0</x>
      <y>0</y>
      <width>1081</width>
-     <height>19</height>
+     <height>22</height>
```
jfmherokiller: I dont know why this changed

yuriks:
Is there any particular demand for this feature? In all cases you can simply re-open Citra. I don't think it's worth the UI clutter.

A better way of doing something like this would be to use [QFileSystemWatcher](http://doc.qt.io/qt-5/qfilesystemwatcher.html) to automatically detect changes to game directories and update the list.

jfmherokiller:
@yuriks well i thought it was just a good idea. ill see if i can implement it using what you described instead of a ui option and add it to this pull request

jfmherokiller:
@yuriks i was successfully able to implement it im just not sure if my organization is very good

jroweboy:
Maybe I'm misreading the code, but it doesn't look like it handles the following case

Open citra -> set game folder directory -> close citra
Open citra -> add a new game to the folder -> no refresh?

I don't see where the watcher adds the games list directory onload. It looks like it only adds it when the Select Game Directory is clicked.

yuriks:
Also, AFAIK watching directories isn't recursive, so you need to actually watch all it's sub-diretories and keep track of them as they're added and removed. (It's probably easier to just clear the watch list and do a full re-scan instead of attempting to scan only changed files and dirs.)

mailwl:
or, as a variant - just leave a refresh button

jroweboy:
@mailwl it seems kinda silly to have a menu item to refresh the game list
when QT has as built in file watcher to handle the hard part of
automatically watching the folder in a cross platform way.

Is there any compelling reason *to* have it as a menu option besides being
marginally easier to code it that way? Having a menu option means its more
clutter in the UI and its more surprising to users ("wait i have to click
refresh?"). Adding a file watcher is much more user friendly without being
much more complicated to code.

mailwl:
@jroweboy as you wish :)

jfmherokiller:
jesus this thing is more in depth then i thought lol

jroweboy:
@jfmherokiller any plans on finishing this? if you need some help you can hop on IRC or discord and I wouldn't mind offering some advice. If you don't want to let me know because i think its a great idea and would love to see it in citra.

jfmherokiller:
@jroweboy the only reason i stopped is because i dont really know enough about qt and c++ in general i kinda make up stuff on the fly and its not very good

PR #2549, opened 2017-02-09
[HACK - Do Not Merge] Loader/NCCH: Ignore compression flag in ExHeader

yuriks:
There seem to be some bad NCCHs around that have this flag unset even
though the .code is compressed. This adds a warning message in these
cases and forces the flag on.

(Context: https://community.citra-emu.org/t/terrys-wonderland-3d-does-not-boot/508)


Salijour:
https://gbatemp.net/threads/translation-dragon-quest-monsters-terrys-wonderland-3d.431204/
I don't know if this is known, but there is an English translation which might be screwing things up. Not like translation patches have been messing up header reading for emulators since ever.

yuriks:
@wwylele confirmed that LDR does respect this flag when loading a NCCH, so this change is indeed a hack. I'm going to assume that NCCHs that fail because of this are bad dumps.

PR #2550, opened 2017-02-09
Small VideoCore cleanups

yuriks:


src/video_core/regs.h
```diff
@@ -45,46 +45,31 @@ namespace Pica {
...
+            TexturingRegs texturing;
+            FramebufferRegs framebuffer;
+            LightingRegs lighting;
+            PipelineRegs pipeline;
+            ShaderRegs gs;
+            ShaderRegs vs;
+            INSERT_PADDING_WORDS(0x20);
+        };
+        std::array<u32, NUM_REGS> reg_array;
+    };
```
yuriks: @lioncash This should be fine, right? (Or at least equivalent to the previous implementation.)

src/video_core/regs.cpp
```diff
@@ -474,19 +474,14 @@ static const std::pair<u16, const char*> register_names[] = {
...
-    if (map.empty()) {
-        map.insert(std::begin(register_names), std::end(register_names));
-    }
-
-    // Return empty string if no match is found
-    auto it = map.find(index);
-    if (it != map.end()) {
-        return it->second;
+const char* Regs::GetRegisterName(u16 index) {
+    auto found = std::lower_bound(std::begin(register_names), std::end(register_names), index,
```
lioncash: `<algorithm>` needs to be included.

src/video_core/regs.h
lioncash: It would still be undefined behavior for some member accesses due to the common initial sequence differing in some places. `LightingRegs` would need to have some of its enum classes explicitly marked with `: u32` to be OK for all members. `PipelineRegs` uses `u64` in some places, so accessing up until the first non-u32 member within it is OK.

Other than those two cases, it's fine.

yuriks: Ah, I wasn't aware there were more u64 members (the ones I knew about were already fixed). I'll add a commit correcting those.

yuriks: Fixed (both enums and u64's)

PR #2553, opened 2017-02-10
video_core: Fix benign out-of-bounds indexing of array

yuriks:
The resulting pointer wasn't written to unless the index was verified as
valid, but that's still UB and triggered debug checks in MSVC.

Reported by garrettboast on IRC


yuriks:
@wwylele Code used to be different, but then I didn't realize this wasn't needed anymore when I changed it in one of my recent refactors.

PR #2555, opened 2017-02-11
Use QFileSystemWatcher to reload the game list when a change is detected.

jroweboy:
A continuation of https://github.com/citra-emu/citra/pull/2546 This makes the watcher recursive if the setting is set to watch the directories recursively and also will start watching the games directory starting from the first load. Special thanks to @jfmherokiller for the idea and the initial implementation (i tried to credit them by leaving their commits in the PR)

### DIFF
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -15,7 +15,7 @@
 #include "game_list_p.h"
 #include "ui_settings.h"
 
-GameList::GameList(QWidget* parent) : QWidget{parent}, watcher{} {
+GameList::GameList(QWidget* parent) : QWidget{parent} {
     QVBoxLayout* layout = new QVBoxLayout;
 
     tree_view = new QTreeView;
@@ -161,7 +161,7 @@ void GameList::RefreshGameDirectory() {
  * list folder.
  *
  * Notice: This method is run on the UI thread because QFileSystemWatcher is not thread safe and
- * this function is fast enough to not stall the UI thread.  If performance is an issue, it should
+ * this function is fast enough to not stall the UI thread. If performance is an issue, it should
  * be moved to another thread and properly locked to prevent concurrency issues.
  *
  * @param dir folder to check for changes in
@@ -179,10 +179,11 @@ void GameList::UpdateWatcherList(const std::string& dir, unsigned int recursion)
         }
         return true;
     };
-    if (FileUtil::IsDirectory(dir)) {
-        watcher.addPath(QString::fromStdString(dir));
+
+    watcher.addPath(QString::fromStdString(dir));
+    if (recursion > 0) {
+        FileUtil::ForeachDirectoryEntry(nullptr, dir, callback);
     }
-    FileUtil::ForeachDirectoryEntry(nullptr, dir, callback);
 }
 
 void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsigned int recursion) {
@@ -216,6 +217,7 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
 
         return true;
     };
+
     FileUtil::ForeachDirectoryEntry(nullptr, dir_path, callback);
 }
 
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -91,7 +91,8 @@ void GameList::PopupContextMenu(const QPoint& menu_location) {
     context_menu.exec(tree_view->viewport()->mapToGlobal(menu_location));
 }
 
-void GameList::PopulateAsync(const QString& dir_path, bool deep_scan) {
+void GameList::PopulateAsync(const QString& dir_path, bool deep_scan,
+                             std::shared_ptr<QFileSystemWatcher> watcher) {
     if (!FileUtil::Exists(dir_path.toStdString()) ||
         !FileUtil::IsDirectory(dir_path.toStdString())) {
         LOG_ERROR(Frontend, "Could not find game list folder at %s", dir_path.toLocal8Bit().data());
@@ -103,7 +104,7 @@ void GameList::PopulateAsync(const QString& dir_path, bool deep_scan) {
     item_model->removeRows(0, item_model->rowCount());
 
     emit ShouldCancelWorker();
-    GameListWorker* worker = new GameListWorker(dir_path, deep_scan);
+    GameListWorker* worker = new GameListWorker(dir_path, deep_scan, watcher);
 
     connect(worker, &GameListWorker::EntryReady, this, &GameList::AddEntry, Qt::QueuedConnection);
     connect(worker, &GameListWorker::Finished, this, &GameList::DonePopulating,
@@ -166,22 +167,28 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
                 new GameListItemSize(FileUtil::GetSize(physical_name)),
             });
         } else if (recursion > 0) {
+            watcher->addPath(QString::fromStdString(physical_name));
             AddFstEntriesToGameList(physical_name, recursion - 1);
         }
 
         return true;
     };
-
+    // add the base directory to the watch path as well
+    watcher->addPath(QString::fromStdString(dir_path));
     FileUtil::ForeachDirectoryEntry(nullptr, dir_path, callback);
 }
 
 void GameListWorker::run() {
     stop_processing = false;
+    auto watch_dirs = watcher->directories();
+    if (!watch_dirs.isEmpty()) {
+        watcher->removePaths(watch_dirs);
+    }
     AddFstEntriesToGameList(dir_path.toStdString(), deep_scan ? 256 : 0);
     emit Finished();
 }
 
 void GameListWorker::Cancel() {
-    disconnect(this, nullptr, nullptr, nullptr);
+    this->disconnect();
     stop_processing = true;
 }
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -17,6 +17,7 @@
 
 GameList::GameList(QWidget* parent) : QWidget{parent} {
     QVBoxLayout* layout = new QVBoxLayout;
+    watcher = std::make_unique<QFileSystemWatcher>();
 
     tree_view = new QTreeView;
     item_model = new QStandardItemModel(tree_view);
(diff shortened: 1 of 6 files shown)


src/citra_qt/game_list.cpp
```diff
@@ -38,6 +39,7 @@ GameList::GameList(QWidget* parent) : QWidget{parent} {
 
     connect(tree_view, &QTreeView::activated, this, &GameList::ValidateEntry);
     connect(tree_view, &QTreeView::customContextMenuRequested, this, &GameList::PopupContextMenu);
+    connect(watcher.get(), SIGNAL(directoryChanged(QString)), SLOT(RefreshGameDirectory()));
```
yuriks: Use the newer pointer-based syntax:

```
connect(watcher.get(), &QFileSystemWatcher::directoryChanged, this, &GameList::RefreshGameDirectory);
```

src/citra_qt/game_list.h
```diff
@@ -38,14 +39,19 @@ class GameList : public QWidget {
     void ShouldCancelWorker();
     void OpenSaveFolderRequested(u64 program_id);
 
+private slots:
+    void RefreshGameDirectory();
```
yuriks: I think if you can move this to just be a private function instead of a slot if you change the signal syntax above.

src/citra_qt/game_list.h
```diff
@@ -38,14 +39,19 @@ class GameList : public QWidget {
...
     void ValidateEntry(const QModelIndex& item);
     void DonePopulating();
 
     void PopupContextMenu(const QPoint& menu_location);
+    void UpdateWatcherList(const std::string& path, unsigned int recursion);
 
     QTreeView* tree_view = nullptr;
     QStandardItemModel* item_model = nullptr;
     GameListWorker* current_worker = nullptr;
+    std::unique_ptr<QFileSystemWatcher> watcher;
```
yuriks: Instead of a unique_ptr just set the widget to be the watcher's parent in the constructor, and that will handle destroying it.

src/citra_qt/main.cpp
```diff
@@ -67,7 +66,6 @@ GMainWindow::GMainWindow() : config(new Config()), emu_thread(nullptr) {
     RestoreUIState();
 
     ConnectWidgetEvents();
-
```
yuriks: Kinda random

src/citra_qt/main.h
```diff
@@ -6,6 +6,7 @@
 #define _CITRA_QT_MAIN_HXX_
 
 #include <memory>
+#include <QFileSystemWatcher>
```
yuriks: Not necessary anymore

src/citra_qt/game_list.cpp
```diff
@@ -131,6 +139,29 @@ void GameList::LoadInterfaceLayout() {
...
 }
 
+void GameList::RefreshGameDirectory() {
+    if (!UISettings::values.gamedir.isEmpty() && current_worker != nullptr) {
+        LOG_INFO(Frontend, "Change detected in the games directory. Reloading game list.");
+        PopulateAsync(UISettings::values.gamedir, UISettings::values.gamedir_deepscan);
+    }
+}
+
+void GameList::UpdateWatcherList(const std::string& dir, unsigned int recursion) {
```
yuriks: Add a comment explaining why this was moved to the main thread instead of being populated from the thread.

yuriks:
Also needs a rebase.

src/citra_qt/main.cpp
jfmherokiller: gotta love those randome whitespaces that slip through

src/core/loader/loader.cpp
```diff
@@ -139,7 +139,9 @@ std::unique_ptr<AppLoader> GetLoader(const std::string& filename) {
             type = filename_type;
     }
 
-    LOG_INFO(Loader, "Loading file %s as %s...", filename.c_str(), GetFileTypeString(type));
+    if (type != FileType::Unknown) {
```
jroweboy: This doesn't need to be here anymore since the changes in https://github.com/citra-emu/citra/pull/2402

jfmherokiller:
so if i push a commit to my thing will it show up here?

jroweboy:
No. If you have changes you think I should make, leave them as a comment
and I'll do that when I get a chance to update the code. I wanted to give
you credit for your work on the initial code so I went and squashed your
commits and kept you as the author. But this Pr isn't connected to your
repo or branches or anything.

src/citra_qt/game_list.cpp
```diff
@@ -140,6 +147,41 @@ static bool HasSupportedFileExtension(const std::string& file_name) {
...
+ * The file watcher will fire off an update to the game list when a change is detected in the
+ * game list folder.
+ * Notice: This method is run on the UI thread because QFileSystemWatcher is not thread safe
+ * and this function is fast enough to not stall the UI thread. If performance is an issue, it
+ * should be moved to another thread and properly locked to prevent concurrency issues.
+ * @param dir folder to check for changes in
+ * @param recursion 0 if recursion is disabled. Any positive number passed to this will add each
+ * directory recursively to the watcher and will update the file list if any of the folders change.
+ * The number determines how deep the recursion should traverse.
+ */
```
yuriks: Can you format this better? These run-on paragraphs are very hard to read.

```
/**
 * Adds the game list folder to the QFileSystemWatcher to check for updates.
 *
 * The file watcher will fire off an update to the game list when a change is detected in the
 * game list folder.
 *
 * Notice: This method is run on the UI thread because QFileSystemWatcher is not thread safe
 * and this function is fast enough to not stall the UI thread. If performance is an issue, it
 * should be moved to another thread and properly locked to prevent concurrency issues.
 *
 * @param dir folder to check for changes in
 * @param recursion 0 if recursion is disabled. Any positive number passed to this will add each
 *        directory recursively to the watcher and will update the file list if any of the folders
 *        change. The number determines how deep the recursion should traverse.
 */
```

src/citra_qt/game_list.cpp
```diff
@@ -15,7 +15,7 @@
 #include "game_list_p.h"
 #include "ui_settings.h"
 
-GameList::GameList(QWidget* parent) : QWidget{parent} {
+GameList::GameList(QWidget* parent) : QWidget{parent}, watcher{parent} {
```
yuriks: This will result in a double free. Either dynamically allocate the watcher, or don't set a parent. QObject deletes its children when it's destructed.

jroweboy:
Addressed the comments

src/citra_qt/game_list.cpp
```diff
@@ -15,7 +15,7 @@
 #include "game_list_p.h"
 #include "ui_settings.h"
 
-GameList::GameList(QWidget* parent) : QWidget{parent} {
+GameList::GameList(QWidget* parent) : QWidget{parent}, watcher{} {
```
yuriks: You're just calling the default constructor, so this `watcher{}` is useless.

src/citra_qt/game_list.cpp
```diff
@@ -140,6 +147,44 @@ static bool HasSupportedFileExtension(const std::string& file_name) {
...
+}
+
+/**
+ * Adds the game list folder to the QFileSystemWatcher to check for updates.
+ *
+ * The file watcher will fire off an update to the game list when a change is detected in the game
+ * list folder.
+ *
+ * Notice: This method is run on the UI thread because QFileSystemWatcher is not thread safe and
+ * this function is fast enough to not stall the UI thread.  If performance is an issue, it should
```
bunnei: nit: Never use two spaces between a period and a new sentence :P
(even in real writing, it's a bad habit!)

src/citra_qt/game_list.cpp
```diff
@@ -140,6 +147,44 @@ static bool HasSupportedFileExtension(const std::string& file_name) {
...
+ * be moved to another thread and properly locked to prevent concurrency issues.
+ *
+ * @param dir folder to check for changes in
+ * @param recursion 0 if recursion is disabled. Any positive number passed to this will add each
```
bunnei: nit: Not a great name, fwiw I've seen "max_depth" used elsewhere for this type of things (as it's really the max directory depth searched)

wwylele:
This caused a big lag, about 5 seconds, on citra booting up, with the following line spammed in the log:
`QInotifyFileSystemWatcherEngine::addPaths: inotify_add_watch failed: No space left on device`

OS: Ubuntu 16.04

src/citra_qt/game_list.cpp
```diff
@@ -140,6 +147,44 @@ static bool HasSupportedFileExtension(const std::string& file_name) {
...
+ *        directory recursively to the watcher and will update the file list if any of the folders
+ *        change. The number determines how deep the recursion should traverse.
+ */
+void GameList::UpdateWatcherList(const std::string& dir, unsigned int recursion) {
```
wwylele: `recursion` is unused

bunnei:
Regarding @wwylele's comment... 80 files isn't really that big or uncommon, IMO. @jroweboy, can you look into repro'ing this lag and seeing if it can be eliminated?

yuriks:
That's weird, because afaik when not in recursive mode, it should only place 1 watch on the directory and that's it, we don't try to watch individual files.

yuriks:
`recursion` is never actually checked in `UpdateWatcherList`, so it'll always recurse (infinitely) no matter if you've selected recursive scan or not.

yuriks:
Patch:
```patch
diff --git a/src/citra_qt/game_list.cpp b/src/citra_qt/game_list.cpp
index b551ac0..3bf95cf 100644
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -179,10 +179,11 @@ void GameList::UpdateWatcherList(const std::string& dir, unsigned int recursion)
         }
         return true;
     };
-    if (FileUtil::IsDirectory(dir)) {
-        watcher.addPath(QString::fromStdString(dir));
+
+    watcher.addPath(QString::fromStdString(dir));
+    if (recursion > 0) {
+        FileUtil::ForeachDirectoryEntry(nullptr, dir, callback);
     }
-    FileUtil::ForeachDirectoryEntry(nullptr, dir, callback);
 }

 void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsigned int recursion) {
```

src/citra_qt/game_list.cpp
jroweboy: the word recursion was chosen as the function was a copy -> paste -> modify from another function in the same file. See AddFstEntriesToGameList https://github.com/citra-emu/citra/blob/f75e964d494b9870d0991a33f1981a6485696fda/src/citra_qt/game_list.cpp#L143

Issue #2519, opened 2017-02-03
Crash loading Yuki-FM homebrew (Lua Player based)


voltagex:
- You have provided the version (commit hash) of Citra you are using.
d0bf7df

- You have provided sufficient detail for the issue to be reproduced.
1. git clone https://github.com/coc4tm/Yuki-FM

2. chmod +x the Linux build tools in the repo

3. `make`

4. Run `build/Yuki FM/Yuki FM.3dsx`

- You have provided system specs (if relevant).
Fedora 25 using a nightly binary from Citra site

- Please also provide:
  - For crashes, a backtrace.
```
[voltagex@badger Yuki FM]$ ~/Downloads/citra-linux-20170131-d0bf7df/citra Yuki\ FM.3dsx 
[   0.000000] Config <Info> citra/config.cpp:LoadINI:36: Successfully loaded /home/voltagex/.config/citra-emu/sdl2-config.ini
[   0.005169] Audio <Error> audio_core/sink_details.cpp:GetSinkDetails:32: AudioCore::SelectSink given invalid sink_id 
[   0.140395] Loader <Info> core/loader/loader.cpp:GetLoader:142: Loading file Yuki FM.3dsx as 3DSX...
[   0.151050] Service.FS <Info> core/file_sys/archive_sdmc.cpp:ArchiveFactory_SDMC:309: Directory /home/voltagex/.local/share/citra-emu/sdmc/ set as SDMC.
[   0.151161] Common.Filesystem <Warning> common/file_util.cpp:CreateFullPath:200: path exists /home/voltagex/.local/share/citra-emu/sdmc/
[   0.151217] Service.FS <Info> core/file_sys/archive_sdmcwriteonly.cpp:ArchiveFactory_SDMCWriteOnly:35: Directory /home/voltagex/.local/share/citra-emu/sdmc/ set as SDMCWriteOnly.
[   0.151241] Common.Filesystem <Warning> common/file_util.cpp:CreateFullPath:200: path exists /home/voltagex/.local/share/citra-emu/sdmc/
[   0.151266] Service.FS <Info> core/file_sys/archive_source_sd_savedata.cpp:ArchiveSource_SDSaveData:42: Directory /home/voltagex/.local/share/citra-emu/sdmc/Nintendo 3DS/00000000000000000000000000000000/00000000000000000000000000000000/title/ set as SaveData.
[...]

Issue #2515, opened 2017-02-02
ArchiveId=3 it's Card, not just RomFS


mailwl:
When AC is connecting, it compares EULA from config, and from SMDH icon. To load SMDH, it opens archive id 3, then opens file with Path=[Binary: 0200000069636f6e00000000]
looks like 2 - section number, and in the middle - name ("icon").

```
[   3.155553] Service.FS <Debug> core\hle\service\fs\fs_user.cpp:Service::FS::OpenArchive:443: archive_id=0x00000003 archive_path=[Empty]
[   3.156306] Service.FS <Trace> core\hle\service\fs\archive.cpp:Service::FS::OpenArchive:266: Opening archive with id code 0x00000003
[   3.192393] Service.FS <Debug> core\hle\service\fs\fs_user.cpp:Service::FS::OpenFile:69: path=[Binary: 0200000069636f6e00000000], mode=1 attrs=0
[   3.231121] Service.FS <Trace> core\hle\service\fs\archive.cpp:Service::FS::File::HandleSyncRequest:105: Read Path: [Binary: 0200000069636f6e00000000]: offset=0x0 length=14016 address=0x121000
[   3.233528] Service.FS <Error> core\hle\service\fs\archive.cpp:Service::FS::File::HandleSyncRequest:110: Reading from out of bounds offset=0x0 length=0x000036C0 file_size=0x3000
[   3.234085] Service.FS <Trace> core\file_sys\ivfc_archive.cpp:FileSys::IVFCFile::Read:90: called offset=0x0, length=0x36C0
[   3.248890] Service.FS <Trace> core\hle\service\fs\archive.cpp:Service::FS::File::HandleSyncRequest:160: Close Path: [Binary: 0200000069636f6e00000000]
```

wwylele:
This is [partially documented on 3dbrew](https://www.3dbrew.org/wiki/Filesystem_services#Binary_LowPath)

>The 0x3 archive is an interface for the 0x2345678E archive with the current process programID+mediatype. The file lowpath is 3-words. These words are written to 0x2345678E-archive file_lowpath+0, with the rest of that lowpath set to all-zero(lowpath is different from archive 0x2345678A). 

(Though I think that `file_lowpath+0` should be `file_lowpath+8`)

File low path for 0x2345678E

|Index word 	|Description|
|-------------|-------|
[...]

Issue #1862, opened 2016-05-28
Eucliod - Crash while booting


Andrix44:
[LOG](http://pasted.co/8db82bf4)
This is a homebrew game, and it is not that easy to find on Google, so you can check it out [here](https://gbatemp.net/threads/eucliod-a-homebrew-bullet-hell-for-the-3ds.419553/).
At first I tought that it crashes at `[   1.989495] Service.FS <Error> core\hle\service\fs\archive.cpp:Service::FS::File::SyncRequest:108: Reading from out of bounds offset=0x0 length=0x00000028 file_size=0x0` , but it still crashes after trying it with https://github.com/citra-emu/citra/pull/1634 .
![_j_bitk_p_b](https://cloud.githubusercontent.com/assets/13806656/15626094/c6f9261a-24bb-11e6-907b-834b5704f385.jpg)

mailwl:
FYI, cia to 3ds converted works fine. (it's no reason to close this issue)

Andrix44:
Yep, it does.

Subv:
The 3dsx has a header size of 0x20 and it's trying to open the RomFS.
````
[20:05:10] <Cruel`> subv, 0x20 is the size of the 3dsx header when it's missing the romfs entries (the old header size)
[20:05:22] <Cruel`> I think it's user error not making proper 3dsx
[20:05:41] <Cruel`> maybe using old 3dsxtool
````

I'm closing this as invalid.

yuriks:
Is this really invalid? Shouldn't we support 3dsx files with the old header format?

Subv:
@yuriks The thing is that it's not really the old format, since it's trying to use the RomFS, it's a badly-generated 3dsx

yuriks:
Reopening since old format or not this is still our bug: If opening RomFS fails, we just log to the console and then happily use the nullptr file object to read. We should report and error so that the application knows it failed to open RomFS.

- Log when opening RomFS: https://github.com/citra-emu/citra/blob/e5a7f2ef29a8dc715ab9670dedb632a02b835372/src/core/file_sys/archive_romfs.cpp#L19
- Same file, on line 25, no error checking is done before creating the archive in Open.

wwylele:
With #2561 merged, on current master the 3dsx version can boot now, but it hangs when loading

[...]

PR #2561, opened 2017-02-11
file_sys: change RomFS archive to Self NCCH archive

wwylele:
aka "Correct FS error codes Part 2"

Addressed #2515. I did a hardware test and confirmed the use of archive 0x00000003 to access ExeFS sections. 

Also fixed the crash when a game tries to access non-exist RomFS (#2519, probably #1862, #1412). Many homebrew applications depend on the RomFS open result to choose whether load from RomFS or SDMC. Now citra returns a proper error code and they should work fine.

Note that ArchiveNCCH (0x2345678A) also have similar ability to access ExeFS section, but its file path format is different from SelfNCCH one (means they cannot share the same `ArchiveBackend` implementation), and it is much more complicated, so I keep it as-is, alone with the legacy `IVFCArchive`, and leave it to a future PR.

All added error codes are verified against real hardware. However I cannot provide the testing program because it is a big mess... if it is need, I will clean it up and upload it later.

A small diagram of the component changes: (green - added; red - removed)
rom_fs


src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,239 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// FileSys namespace
+
+namespace FileSys {
+
+enum class SelfNCCHFilePathType : u32 {
+    RomFS = 0,
+    Code = 1, // This is not supported by SelfNCCHArchive but by archive 0x2345678E
+    ExeFS = 2,
+    UpdateRomFS = 5,
```
Subv: What does the UpdateRomFS do?

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+ArchiveFactory_SelfNCCH::ArchiveFactory_SelfNCCH(Loader::AppLoader& app_loader) {
+    std::shared_ptr<FileUtil::IOFile> romfs_file_;
+    if (Loader::ResultStatus::Success ==
+        app_loader.ReadRomFS(romfs_file_, ncch_data.romfs_offset, ncch_data.romfs_size))
+        ncch_data.romfs_file = romfs_file_;
+
+    std::vector<u8> buffer;
+
+    if (Loader::ResultStatus::Success == app_loader.ReadIcon(buffer))
```
Subv: Should we log an error in the failure case of each of these?

mailwl:
Now proper EULA version checking from icon may broke some games (mostly which wants ErrEULA applet before).
But we tune it later.

src/core/file_sys/archive_selfncch.cpp
wwylele: Presumably for accessing the RomFS of the update patch. 

On 3dbrew it says:
>The update RomFS can be accessed with the first u32 in the binary file-lowpath being set to 0x5.

(though this part of documentation is a big mess)

 On my self-made CIA-installed program without update patch, this gives the same data as RomFS (0)

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,239 @@
...
+            } else {
+                LOG_ERROR(Service_FS, "Unable to read RomFS!");
+                return ERROR_ROMFS_NOT_FOUND;
+            }
+
+        case SelfNCCHFilePathType::Code:
+            LOG_ERROR(Service_FS, "Reading code section is not supported!");
+            return ERROR_COMMAND_NOT_ALLOWED;
+
+        case SelfNCCHFilePathType::ExeFS: {
```
lioncash: Given the conditionals contain nested returns, `else` can be omitted in quite a few places and simply be:

```cpp
case SelfNCCHFilePathType::ExeFS: {
    std::string filename(file_path.exefs_filename.begin(), file_path.exefs_filename.end());
    const size_t pos = filename.find('\0');
    if (pos != std::string::npos)
        filename.erase(pos);

    if (filename == "icon") {
        if (ncch_data.icon) {
            return MakeResult<std::unique_ptr<FileBackend>>(
                std::make_unique<ExeFSSectionFile>(*ncch_data.icon));
        }

        LOG_ERROR(Service_FS, "Unable to read icon!");
        return ERROR_EXEFS_SECTION_NOT_FOUND;
    }

    if (filename == "logo") {
        if (ncch_data.logo) {
            return MakeResult<std::unique_ptr<FileBackend>>(
                std::make_unique<ExeFSSectionFile>(*ncch_data.logo));
        }

        LOG_ERROR(Service_FS, "Unable to read logo!");
        return ERROR_EXEFS_SECTION_NOT_FOUND;
    }

    if (filename == "banner") {
        if (ncch_data.banner) {
            return MakeResult<std::unique_ptr<FileBackend>>(
                std::make_unique<ExeFSSectionFile>(*ncch_data.banner));
        }

        LOG_ERROR(Service_FS, "Unable to read banner!");
        return ERROR_EXEFS_SECTION_NOT_FOUND;
    }

    LOG_ERROR(Service_FS, "Unknown ExeFS section %s!", filename.c_str());
    return ERROR_INVALID_PATH;
}
```

Another suggestion to consider might be also moving it to a private `OpenExeFS()` function, like with the RomFS code suggested above, so no fallthrough can be introduced by accident if the code ever gets changed.

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,239 @@
...
+            LOG_ERROR(Service_FS, "Wrong path size %zu", binary.size());
+            return ERROR_INVALID_PATH;
+        }
+
+        SelfNCCHFilePath file_path;
+        std::memcpy(&file_path, binary.data(), sizeof(SelfNCCHFilePath));
+
+        switch (file_path.type) {
+        case SelfNCCHFilePathType::UpdateRomFS:
+            LOG_WARNING(Service_FS, "(STUBBED) open update RomFS");
```
wwylele: Yes, it's intentional. Will separate

src/core/file_sys/archive_selfncch.cpp
wwylele: Done

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,254 @@
...
+        return true;
+    }
+
+    void Flush() const override {}
+
+private:
+    std::shared_ptr<std::vector<u8>> data;
+};
+
+class SelfNCCHArchive final : public ArchiveBackend {
```
yuriks: Document what the SelfNCCHArchive does in general and what kinds of "files" it lets you open.

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,254 @@
...
+        case SelfNCCHFilePathType::UpdateRomFS:
+            LOG_WARNING(Service_FS, "(STUBBED) open update RomFS");
+            return OpenRomFS();
+
+        case SelfNCCHFilePathType::RomFS:
+            return OpenRomFS();
+
+        case SelfNCCHFilePathType::Code:
+            LOG_ERROR(Service_FS, "Reading code section is not supported!");
+            return ERROR_COMMAND_NOT_ALLOWED;
```
yuriks: This feels like a misnomer, but I don't have much additional context into the structure of the service. Seems like what's currently called `ERROR_UNSUPPORTED_OPEN_FLAGS` should be comething like `ERROR_UNSUPPORTED_OPERATION`, and then this should be `ERROR_UNSUPPORTED_PATH_TYPE`, but maybe it makes more sense when you take into account other usages.

src/core/file_sys/archive_selfncch.cpp
yuriks: Document this in the code.

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,254 @@
...
+        switch (file_path.type) {
+        case SelfNCCHFilePathType::UpdateRomFS:
+            LOG_WARNING(Service_FS, "(STUBBED) open update RomFS");
+            return OpenRomFS();
+
+        case SelfNCCHFilePathType::RomFS:
+            return OpenRomFS();
+
+        case SelfNCCHFilePathType::Code:
+            LOG_ERROR(Service_FS, "Reading code section is not supported!");
```
yuriks: Reading *the* code section [...]

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+    u64 GetFreeBytes() const override {
+        return 0;
+    }
+
+private:
+    ResultVal<std::unique_ptr<FileBackend>> OpenRomFS() const {
+        if (ncch_data.romfs_file) {
+            return MakeResult<std::unique_ptr<FileBackend>>(std::make_unique<IVFCFile>(
+                *ncch_data.romfs_file, ncch_data.romfs_offset, ncch_data.romfs_size));
```
yuriks: You should be able to omit the template parameter to `MakeResult` and let it be auto-deduced.

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,254 @@
...
+        } else {
+            LOG_ERROR(Service_FS, "Unable to read RomFS!");
```
yuriks: Isn't this a pretty common thing for homebrew to try to do? (If I understand the issue reports we've had correctly.) Maybe we should turn down these log levels if this is expected.

src/core/file_sys/archive_selfncch.h
```diff
@@ -17,22 +18,29 @@
 
 namespace FileSys {
 
-/// File system interface to the RomFS archive
-class ArchiveFactory_RomFS final : public ArchiveFactory {
+struct NCCHData {
+    boost::optional<std::shared_ptr<std::vector<u8>>> icon;
+    boost::optional<std::shared_ptr<std::vector<u8>>> logo;
+    boost::optional<std::shared_ptr<std::vector<u8>>> banner;
+    boost::optional<std::shared_ptr<FileUtil::IOFile>> romfs_file;
```
yuriks: I don't think the `optional`s are adding much value here to be honest. shared_ptr is already nullable so this is just introducing an extra level of optionalness. If you remove them I think the rest of the code actually stays exactly the same.

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,254 @@
...
+    std::shared_ptr<FileUtil::IOFile> romfs_file_;
+    if (Loader::ResultStatus::Success ==
+        app_loader.ReadRomFS(romfs_file_, ncch_data.romfs_offset, ncch_data.romfs_size)) {
+        ncch_data.romfs_file = std::move(romfs_file_);
+    }
+
+    std::vector<u8> buffer;
+
+    if (Loader::ResultStatus::Success == app_loader.ReadIcon(buffer))
+        ncch_data.icon = std::make_shared<std::vector<u8>>(std::move(buffer));
```
yuriks: While I realize that this works properly, I had to go look it up in the std::vector reference and in the implementation of ReadIcon/Logo/Banner to verify that re-using the moved-from vector was ok. I think a `buffer.clear()` in-between each of these would help isolate them from the loader implementation, or the fact that previous contents are completely discarded should be made explicit in the AppLoader documentation.

src/core/file_sys/archive_selfncch.h
```diff
@@ -17,22 +18,29 @@
...
     }
     ResultVal<std::unique_ptr<ArchiveBackend>> Open(const Path& path) override;
     ResultCode Format(const Path& path, const FileSys::ArchiveFormatInfo& format_info) override;
     ResultVal<ArchiveFormatInfo> GetFormatInfo(const Path& path) const override;
 
 private:
-    std::shared_ptr<FileUtil::IOFile> romfs_file;
-    u64 data_offset;
-    u64 data_size;
+    NCCHData ncch_data;
```
yuriks: Is there any particular reason why this was split up to a separate struct?

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    NCCHData ncch_data;
+};
+
+ArchiveFactory_SelfNCCH::ArchiveFactory_SelfNCCH(Loader::AppLoader& app_loader) {
+    std::shared_ptr<FileUtil::IOFile> romfs_file_;
+    if (Loader::ResultStatus::Success ==
+        app_loader.ReadRomFS(romfs_file_, ncch_data.romfs_offset, ncch_data.romfs_size)) {
+        ncch_data.romfs_file = std::move(romfs_file_);
```
yuriks: Add a blank line between the condition and the statement.

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+        case SelfNCCHFilePathType::Code:
+            LOG_ERROR(Service_FS, "Reading code section is not supported!");
+            return ERROR_COMMAND_NOT_ALLOWED;
+
+        case SelfNCCHFilePathType::ExeFS: {
+            std::string filename(file_path.exefs_filename.begin(), file_path.exefs_filename.end());
+            const size_t pos = filename.find('\0');
+            if (pos != std::string::npos)
+                filename.erase(pos);
```
wwylele: This is the same idiom as [here](https://github.com/citra-emu/citra/blob/master/src/core/hle/service/cfg/cfg.cpp#L607) which was originally suggested by @lioncash 

Yeah I will add comments to it

src/core/file_sys/archive_selfncch.cpp
wwylele: No, there is a `IVFCFile`->`FileBackend` pointer cast, so the type need to be there

src/core/file_sys/archive_selfncch.h
wwylele: So that `Archive_SelfNCCH` and `SelfNCCHArchive` can share the same struct and copying data can be simpler.

src/core/file_sys/archive_selfncch.cpp
yuriks: I mean you can just do this:

```
auto end = std::find(file_path.exefs_filename.begin(), file_path.exefs_filename.end(), '\0');
std::string filename(file_path.exefs_filename.begin(), end);
```

I think this is better, anyway. (Duplicating the `file_path_exefs_filename` 3 times is a bit annoying because of how long it is though.) Up to you, I guess.

src/core/file_sys/archive_selfncch.cpp
wwylele: The name `ERROR_COMMAND_NOT_ALLOWED` comes from 3dbrew documentation:
> 630 	Command not allowed / missing permissions? (fs) 

and yeah, I agree that currently not all error codes are named correctly. I prefer using the existing name for a while, and after collecting enough usage of these codes, we then do a name clean up

src/core/file_sys/archive_selfncch.cpp
wwylele: added clear

wwylele:
Addressed comments above. Update:

Lowered the log level of missing RomFS from ERROR to INFO (because it is common in homebrew application), and the level of missing ExeFS section from ERROR to WARNING (less common but we still handle it correctly)

PR #2563, opened 2017-02-13
Add asio as a submodule

jroweboy:
Fixes #Subv pesters random people on IRC to get them to make this extremely minor change instead of writing the 2 lines of code.

As far as I know Subv wants asio added so that soc:u and maybe some other things can use that instead.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2563)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "cryptopp"]
     path = externals/cryptopp/cryptopp
     url = https://github.com/weidai11/cryptopp.git
+[submodule "asio"]
+    path = externals/asio
+    url = https://github.com/chriskohlhoff/asio/
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -9,3 +9,6 @@ endif()
 
 add_subdirectory(cryptopp)
 
+add_library(asio INTERFACE)
+target_include_directories(asio INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/asio/asio/include)
+target_compile_definitions(asio INTERFACE -DASIO_STANDALONE=1)


CMakeLists.txt
```diff
@@ -271,6 +271,9 @@ endif()
 # process subdirectories
 add_subdirectory(externals/soundtouch)
 
+add_definitions(-DASIO_STANDALONE=1)
+include_directories(externals/asio/asio/include)
```
yuriks: Add this to the CMakeFiles in externals/ instead.

wwylele: I think this should be added to the one in the subdirectory that uses asio? (In this case, probably  `src/core/CMakeLists.txt`). Or just remove this and let the one add this who are going to write code using asio

yuriks: For some reason I thought that ASIO had a CMake file that we could just import.
The proper way of doing it is using the CMake support for library interfaces. See what I did for xbyak for example.

Saphiresurf:
This will be nice!

jfmherokiller:
i think this change would be better on the https://github.com/citra-emu/ext-boost submodule since it is built into boost if added by mcp.

MerryMage:
@jfmherokiller Doing so (minorly) complicates the build, because the boost version of asio requires boost::system.

jfmherokiller:
oh i see and boost system is already included in c++11 so it would add duplicates.

yuriks:
We'd need to create a CMake-based build system for the boost version anyway. The standalone version is updated more often and I'd prefer using as many native C++ features as possible instead of the older boost alternatives.

yuriks:
Just to be sure, can you actually link asio into one of our libraries with this target? (Just to prevent a following PR to fix it if that ends up failing.) If so, then LGTM

bunnei:
@Subv do we still want this? I don't see the point of merging this if we're not using it. Do you plan on making the changes to use it in the near term? Otherwise, this should be closed and re-added when we are ready for it.

mailwl:
+1 for this, it's "стильно, модно, молодёжно". Sorry, don't know how to proper translate this meme to english

Subv:
That depends on how you define "near term", i plan on rewriting the soc module with it, but that's blocked until a few more of my kernel and threading patches make it in.

mailwl:
yes, Youtube app still not working

jroweboy:
As this is really old and not necessary at this moment, I'm just going to close this and if someone wants to use asio they can add it themselves (its really straightforward).

Also @yuriks I forgot to mention, but I did have a commit after this one on my local copy that tests asio to make sure it works.

PR #2569, opened 2017-02-14
APT: implemented Wrap and Unwrap

wwylele:
So this is the first use of the crypto library. APT:Wrap/Unwrap are functions for encrypting/decrypting data using AES-CCM with built-in keys. 3DS Mii Maker uses these functions to encrypt its QR code. With these function implemented and correct keys provided, Mii Maker running in citra can now generate QR code and export mii to real 3DS. Importing can also be done theoretically, but it needs implemented camera feature.

AES key slot management is additionally implemented, to mimic the way a real 3DS does. APT:Wrap/Unwrap will use key slot 0x31. You need provide the actual key for this slot. I put the key storage to `<user-directory>/sysdata/aes_keys.txt`. The format of this file is
```
generator=<32-digit hex key generator/scrambler constant>
slot0x31KeyX=<32-digit hex key x>
slot0x31KeyY=<32-digit hex key y>
```
or you can provide the normal key directly
```
slot0x31KeyN=<32-digit hex normal key>
```

If you don't provide the key, it will use the zeros normal key. This still partially works for game using Wrap/Unwrap stores and loads data within citra, but disables the communication between citra and real 3DS.

[Here is a test for the functions](https://github.com/wwylele/ctrhwtest/tree/master/wrap_test) with 727 test cases consisting of different data size, nonce size and offset. To run the test, you need to put the file `wrap_test_case.bin` to the working directory (on citra it is SD root directory `sdmc`, on real 3DS it is determined by hblauncher and usually is the directory the the 3dsx is in), and run the compiled 3dsx file. On citra it passes only when you provide the correct key.


wwylele:
btw here is a mii QR code generated by citra :smile: 
hni_0022

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,222 @@
...
+            continue;
+        }
+
+        int slot_id;
+        char key_type;
+        if (std::sscanf(name.c_str(), "slot0x%XKey%c", &slot_id, &key_type) != 2) {
+            continue;
+        }
+
+        if (slot_id < 0 || slot_id >= 0x40) {
```
mailwl: MaxKeySlotID ?

src/core/hle/service/apt/apt.cpp
```diff
@@ -470,6 +472,107 @@ void GetStartupArgument(Service::Interface* self) {
...
+    u32 pdata_size = input_size - nonce_size;
+    std::vector<u8> pdata(pdata_size);
+    Memory::ReadBlock(input, pdata.data(), nonce_offset);
+    Memory::ReadBlock(input + nonce_offset + nonce_size, pdata.data() + nonce_offset,
+                      pdata_size - nonce_offset);
+
+    // encrypts the plain text using AES-CCM
+    auto cipher = HW::AES::EncryptSignCCM(pdata, nonce, HW::AES::KeySlotID::APTWrap);
+
+    // Puts the nonce to the begining of the output, with cipher text data followed
```
mailwl: beginning

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,222 @@
...
+}
+
+AESKey HexToKey(const std::string& hex) {
+    if (hex.size() < 32) {
+        return {};
+    }
+
+    AESKey key;
+    try {
+        for (int i = 0; i < 16; ++i) {
```
lioncash: Reminder, indexing with size_t or ptrdiff_t, when an API supports it, is generally [mildly more performant](https://godbolt.org/g/yKGqGo)

src/core/hw/aes/key.h
```diff
@@ -0,0 +1,32 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <algorithm>
```
lioncash: There's nothing using algorithm in this header

src/core/hw/aes/key.h
```diff
@@ -0,0 +1,32 @@
...
+namespace HW {
+namespace AES {
+
+enum KeySlotID {
+    APTWrap = 0x31,
+
+    MaxKeySlotID = 0x40,
+};
+
+constexpr int AES_BLOCK_SIZE = 16;
```
lioncash: This should be a `size_t`

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,222 @@
...
+        carry = sum >> 8;
+        out[i] = static_cast<u8>(sum & 0xff);
+    }
+
+    return out;
+}
+
+AESKey Xor128(const AESKey& a, const AESKey& b) {
+    AESKey out;
+    for (int i = 0; i < 16; i++) {
```
lioncash: This can just be

```cpp
AESKey out;
std::transform(a.cbegin(), a.cend(), b.cbegin(), out.begin(), std::bit_xor<>());
return out;
```

src/core/hw/aes/ccm.h
```diff
@@ -0,0 +1,38 @@
...
+#pragma once
+
+#include <array>
+#include <vector>
+#include "common/common_types.h"
+
+namespace HW {
+namespace AES {
+
+constexpr int CCM_NONCE_SIZE = 12;
```
lioncash: These should be of type `size_t`

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,222 @@
...
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// 128-bit unsigned wraparound arithmetic
+
+int WrapIndex128(int i) {
+    return i < 0 ? ((i % 16) + 16) % 16 : (i > 15 ? i % 16 : i);
+}
+
+AESKey Lrot128(const AESKey& in, u32 rot) {
+    AESKey out;
+    u32 bit_shift, byte_shift;
```
lioncash: These can be moved to their assignment and be const variables.

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+    rot = rot % 128;
```
lioncash: `rot %= 128;`

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,222 @@
...
+        available = false;
+    }
+
+    void Set(const AESKey& new_key) {
+        key = new_key;
+        available = true;
+    }
+};
+
+KeyResource generator_constant;
```
lioncash: This should be static

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    void Clear() {
+        x.Reset();
+        y.Reset();
+        normal.Reset();
+    }
+};
+
+std::array<KeySlot, KeySlotID::MaxKeySlotID> key_slots;
```
lioncash: Ditto

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,222 @@
...
+AESKey Xor128(const AESKey& a, const AESKey& b) {
+    AESKey out;
+    for (int i = 0; i < 16; i++) {
+        out[i] = a[i] ^ b[i];
+    }
+    return out;
+}
+
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// key slot management
```
lioncash: If you have to use comments as a separator, it usually means your organization is too convoluted or things should be split into another source file.

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,222 @@
...
+    for (int i = 0; i < 16; i++) {
+        out[i] = (in[WrapIndex128(i + byte_shift)] << bit_shift) |
+                 (in[WrapIndex128(i + byte_shift + 1)] >> (8 - bit_shift));
+    }
+    return out;
+}
+
+AESKey Add128(const AESKey& a, const AESKey& b) {
+    AESKey out;
+    unsigned carry = 0;
```
lioncash: `u32`

src/core/hw/aes/key.cpp
wwylele: It is in the anonymous namespace.

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,176 @@
...
+    ClearAllKeys();
+    LoadPresetKeys();
+}
+
+void SetGeneratorConstant(const AESKey& key) {
+    generator_constant.Set(key);
+}
+
+void SetKeyX(int slot_id, const AESKey& key) {
+    key_slots[slot_id].SetKeyX(key);
```
lioncash: Missed this in my initial review, but should these not be using `.at()` to prevent potential out-of-bounds accesses from silently working? `slot_id` should also be a size_t here, otherwise `operator[]`/`at()` does implicit sign conversion to it. Ditto below.

src/core/hw/aes/arithmetic128.h
```diff
@@ -0,0 +1,16 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "core/hw/aes/key.h"
```
lioncash: `common/common_types.h` needs to be included here.

src/core/hw/aes/key.h
```diff
@@ -0,0 +1,34 @@
...
+namespace HW {
+namespace AES {
+
+enum KeySlotID {
+    APTWrap = 0x31,
+
+    MaxKeySlotID = 0x40,
+};
+
+constexpr size_t AES_BLOCK_SIZE = 16;
```
lioncash: `<cstddef>` needs to be included for `size_t`

src/core/hw/aes/ccm.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include "core/hw/aes/key.h"
+
+namespace HW {
+namespace AES {
+
+namespace {
+
+// 3DS uses a non-standard AES-CCM algorithm, so we need to derive a sub class from the standard one
+// and override with the non-standard part.
+using namespace CryptoPP;
```
lioncash: There's likely no need to dump the whole namespace in here.

src/core/hw/aes/ccm.cpp
```diff
@@ -0,0 +1,91 @@
...
+template <bool T_IsEncryption>
+class CCM_3DSVariant_Final : public CCM_Final<CryptoPP::AES, CCM_MAC_SIZE, T_IsEncryption> {
+public:
+    void UncheckedSpecifyDataLengths(lword headerLength, lword messageLength,
+                                     lword footerLength) override {
+        // 3DS uses the aligned size to generate B0 for authentication, instead of the original size
+        lword aligned_messageLength = Common::AlignUp(messageLength, AES_BLOCK_SIZE);
```
lioncash: `aligned_message_length`

src/core/hw/aes/ccm.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+namespace {
+
+// 3DS uses a non-standard AES-CCM algorithm, so we need to derive a sub class from the standard one
+// and override with the non-standard part.
+using namespace CryptoPP;
+template <bool T_IsEncryption>
+class CCM_3DSVariant_Final : public CCM_Final<CryptoPP::AES, CCM_MAC_SIZE, T_IsEncryption> {
+public:
+    void UncheckedSpecifyDataLengths(lword headerLength, lword messageLength,
```
lioncash: `header_length`, `message_length`, `foorter_length`

src/core/hw/aes/ccm.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+class CCM_3DSVariant {
+public:
+    using Encryption = CCM_3DSVariant_Final<true>;
+    using Decryption = CCM_3DSVariant_Final<false>;
+};
+
+} // namespace
+
+std::vector<u8> EncryptSignCCM(const std::vector<u8>& pdata, const CCMNonce& nonce, int slot_id) {
```
lioncash: Why `pdata` as opposed to just `data`?

src/core/hw/aes/arithmetic128.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+AESKey Xor128(const AESKey& a, const AESKey& b) {
+    AESKey out;
+    std::transform(a.cbegin(), a.cend(), b.cbegin(), out.begin(), std::bit_xor<>());
+    return out;
+}
+
+} // namespace AES
+} // namespace HW
```
lioncash: Missing newline?

src/core/hw/aes/arithmetic128.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+AESKey Lrot128(const AESKey& in, u32 rot) {
+    AESKey out;
+    rot %= 128;
+    const u32 byte_shift = rot / 8;
+    const u32 bit_shift = rot % 8;
+
+    for (int i = 0; i < 16; i++) {
+        out[i] = (in[WrapIndex128(i + byte_shift)] << bit_shift) |
```
lioncash: I missed this in my first review, but note this is mixing two differently signed types here. What you're passing to WrapIndex128 is a u32 (once integral promotions are performed), yet it expects an `int`. There likely needs to be a cast of the final result.

src/core/hw/aes/ccm.cpp
wwylele: PDATA is a generic term to indicate the default or plaint text which is usually encountered in discussions of cipher modes such as CCM and GCM.

src/core/hw/aes/ccm.cpp
wwylele: I followed crypto++ code style and uses the original parameter name in this class. I dit it in this way so that others can get the idea quickly what I modified.

src/core/hw/aes/arithmetic128.cpp
wwylele: screw this. I didn't check it carefully when I copied the code from elsewhere. I will clean it up.

src/core/hw/aes/arithmetic128.h
wwylele: removed the use of `u32`

src/core/hw/aes/ccm.h
```diff
@@ -0,0 +1,38 @@
...
+#pragma once
+
+#include <array>
+#include <vector>
+#include "common/common_types.h"
+
+namespace HW {
+namespace AES {
+
+constexpr size_t CCM_NONCE_SIZE = 12;
```
wwylele: added

src/core/hw/aes/arithmetic128.cpp
wwylele: fixed

src/core/hw/aes/key.h
wwylele: fixed

src/core/hw/aes/ccm.cpp
wwylele: ditto. I treated `messageLength` as an atomic name coming from crypto++ and added `aligned_` suggesting the modification.

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+AESKey HexToKey(const std::string& hex) {
+    if (hex.size() < 32) {
+        return {};
+    }
+
+    AESKey key;
+    try {
+        for (size_t i = 0; i < 16; ++i) {
+            key.at(i) = static_cast<u8>(std::stoi(hex.substr(i * 2, 2), 0, 16));
```
lioncash: This doesn't need to use `at()` it's guaranteed to be in range, given that AESKey defines itself as `std::array<u8, AES_BLOCK_SIZE>;` where `AES_BLOCK_SIZE` is defined as `16`. If anything, the loop header should just be: 

`for (size_t i = 0; i < key.size(); ++i)`

src/core/hw/aes/arithmetic128.h
lioncash: You may as well just include it and use `u32`. The cpp file uses those type aliases as well. `u32` is also shorter/simpler to read. (pedantically, `unsigned int` is not required to be 32 bits wide, `u32` aliases `uint32_t`, which on the other hand, is, and as such, documents an API better).

src/core/hw/aes/ccm.cpp
```diff
@@ -0,0 +1,94 @@
...
+using CryptoPP::CCM_Base;
+template <bool T_IsEncryption>
+class CCM_3DSVariant_Final : public CCM_Final<AES, CCM_MAC_SIZE, T_IsEncryption> {
+public:
+    void UncheckedSpecifyDataLengths(lword headerLength, lword messageLength,
+                                     lword footerLength) override {
+        // 3DS uses the aligned size to generate B0 for authentication, instead of the original size
+        lword aligned_messageLength = Common::AlignUp(messageLength, AES_BLOCK_SIZE);
+        CCM_Base::UncheckedSpecifyDataLengths(headerLength, aligned_messageLength, footerLength);
+        CCM_Base::m_messageLength = messageLength; // restore the actuall message size
```
lioncash: one `l` too many

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include "common/string_util.h"
+#include "core/hw/aes/arithmetic128.h"
+#include "core/hw/aes/key.h"
+
+namespace HW {
+namespace AES {
+
+namespace {
+
+struct KeyResource {
```
MerryMage: `boost::optional<AESKey>`?

src/core/hw/aes/key.cpp
```diff
@@ -0,0 +1,176 @@
...
+    if (hex.size() < 32) {
+        return {};
+    }
+
+    AESKey key;
+    try {
+        for (size_t i = 0; i < key.size(); ++i) {
+            key[i] = static_cast<u8>(std::stoi(hex.substr(i * 2, 2), 0, 16));
+        }
+    } catch (const std::logic_error&) {
```
Subv: Shouldn't this log an error?

wwylele:
Update: 
 - use `boost::optional` instead of `KeyResource`
 - Log errors when parsing aes_keys.txt. Move exception catching outside `HexToKey`

jfmherokiller:
would aeskeydb.bin also work for supplying the keys?

wwylele:
@jfmherokiller no... I didn't look into aeskeydb.bin. We can do it in the future

yuriks:
(This may be causing crashes in BE, untagging so we can confirm or disprove.)

wwylele:
The crash is a SIGILL in crypto++ that only happens in mingw build on some computers. it might be a bug in crypto++. I am trying to figure it out...

Dragios:
I can't apitrace the MinGW build as I don't have the tools required. This is the best I can do.
`Unhandled exception at 0x0000000000713641 in citra-qt.exe: 0xC000001D: Illegal Instruction.`

Disassembly
Call Stack

jroweboy:
The issue is with this line https://github.com/citra-emu/citra/blob/master/externals/cryptopp/CMakeLists.txt#L103 which turns on arch=native. The builds created on appveyor have at least Haswell, which means the compiler optimizes the shift on this line https://github.com/weidai11/cryptopp/blob/master/cpu.cpp#L260 to a shrx which is not supported on Sandy Bridge or lower. The solution would be to only compile with arch=native when the user passes that as a option to cmake, and not use it by default.

wwylele:
Pushed a patch to remove the -march=native flag (yay for copy paste error)
Added back to bleeding edge to test. Hopefully it won't cause problems any more...

RavenHome1:
@wwylele works again on my amd cpu

src/core/hw/aes/key.h
```diff
@@ -0,0 +1,35 @@
...
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include "common/common_types.h"
+
+namespace HW {
+namespace AES {
+
+enum KeySlotID {
```
bunnei: enum class?

wwylele: The actual slot ID are all of `size_t` type because they are index for array elements. This enum is used as a constant group to document the usage of each index. So I don't use `enum class` to avoid casting.

btw I just found there are some type inconsistency... let me fix it

wwylele:
Hidden conversation:
`+enum KeySlotID {`
bunnei: enum class?
wwylele: The actual slot ID are all of size_t type because they are index for array elements. This enum is used as a constant group to document the usage of each index. So I don't use enum class to avoid casting.

small update:
fix `slot_id` type in ccm to `size_t`
```
-enum KeySlotID {
+enum KeySlotID : size_t {
```

PR #2573, opened 2017-02-16
Added drag and drop feature to the code

jfmherokiller:


src/citra_qt/game_list.cpp
```diff
@@ -36,7 +36,7 @@ GameList::GameList(QWidget* parent) : QWidget{parent} {
     item_model->setHeaderData(COLUMN_NAME, Qt::Horizontal, "Name");
     item_model->setHeaderData(COLUMN_FILE_TYPE, Qt::Horizontal, "File type");
     item_model->setHeaderData(COLUMN_SIZE, Qt::Horizontal, "Size");
-
+    this->setAcceptDrops(true);
```
yuriks: Don't use `this->` unless you need to

src/citra_qt/game_list.cpp
```diff
@@ -185,3 +185,17 @@ void GameListWorker::Cancel() {
...
+    const QMimeData* mimeData = event->mimeData();
+    QString filename = mimeData->urls().at(0).toLocalFile();
+    if (HasSupportedFileExtension(filename.toStdString())) {
+        emit GameChosen(filename);
+    }
+}
+void GameList::dragEnterEvent(QDragEnterEvent* event) {
+    event->acceptProposedAction();
+}
+void GameList::dragMoveEvent(QDragMoveEvent* event) {
```
jfmherokiller: yes because without it the file isnt accepted at all

src/citra_qt/game_list.cpp
```diff
@@ -185,3 +185,17 @@ void GameListWorker::Cancel() {
...
+
+void GameList::dropEvent(QDropEvent* event) {
+    const QMimeData* mimeData = event->mimeData();
+    QString filename = mimeData->urls().at(0).toLocalFile();
+    if (HasSupportedFileExtension(filename.toStdString())) {
+        emit GameChosen(filename);
+    }
+}
+void GameList::dragEnterEvent(QDragEnterEvent* event) {
+    event->acceptProposedAction();
```
jfmherokiller:
```
void GameList::dragEnterEvent(QDragEnterEvent* event) { 
  const QMimeData* mimeData = event->mimeData();
  QString filename = mimeData->urls().at(0).toLocalFile();
  if (HasSupportedFileExtension(filename.toStdString())) {
  }
}
```
would this be better?

src/citra_qt/game_list.cpp
```diff
@@ -185,3 +185,17 @@ void GameListWorker::Cancel() {
     disconnect(this, nullptr, nullptr, nullptr);
     stop_processing = true;
 }
+
+void GameList::dropEvent(QDropEvent* event) {
+    const QMimeData* mimeData = event->mimeData();
+    QString filename = mimeData->urls().at(0).toLocalFile();
```
jfmherokiller: well should i put up an error message or deny the drop event if there exists more than one file or if the file is not valid?

yuriks: Just deny/don't accept the event. That'll cause the OS to show an "invalid" icon and reject drop attempt.

src/citra_qt/game_list.cpp
yuriks: `if (mimeData->hasUrls() && mimeData->urls().length() == 1)`

src/citra_qt/game_list.cpp
yuriks: Ok.

src/citra_qt/main.cpp
```diff
@@ -654,3 +654,29 @@ int main(int argc, char* argv[]) {
     main_window.show();
     return app.exec();
 }
+void GMainWindow::dropEvent(QDropEvent* event) {
+    if (ConfirmChangeGame()) {
+        const QMimeData* mimeData = event->mimeData();
+        QString filename = mimeData->urls().at(0).toLocalFile();
```
yuriks: You should still do the check to see if it's a valid event here. (Preferably factor that out into a small function so it's not duplicated between dragEventEnter and dropEvent.)

src/citra_qt/main.cpp
```diff
@@ -654,3 +654,29 @@ int main(int argc, char* argv[]) {
...
+        BootGame(filename.toStdString());
+    }
+}
+void GMainWindow::dragEnterEvent(QDragEnterEvent* event) {
+    const QMimeData* mimeData = event->mimeData();
+    if (mimeData->hasUrls() && (mimeData->urls().length() == 1)) {
```
yuriks: You don't need the extra parenthesis around the second part of the condition.

src/citra_qt/main.cpp
```diff
@@ -654,3 +654,32 @@ int main(int argc, char* argv[]) {
     main_window.show();
     return app.exec();
 }
+bool checkeventvalid(QDropEvent* event) {
```
yuriks: Name this something more descriptive, like `IsSingleFileDropEvent`.

src/citra_qt/main.cpp
```diff
@@ -654,3 +654,32 @@ int main(int argc, char* argv[]) {
     main_window.show();
     return app.exec();
 }
```
yuriks: You should leave a blank line between each function definition. Same for all the other functions you added.

src/citra_qt/main.cpp
yuriks: Also, move your function above the the `main` function (and the ifdefs above it) so it stays with the rest of the GMainWindow code.

src/citra_qt/main.cpp
```diff
@@ -624,6 +624,39 @@ void GMainWindow::closeEvent(QCloseEvent* event) {
 
     QWidget::closeEvent(event);
 }
+bool IsSingleFileDropEvent(QDropEvent* event) {
```
yuriks: You forgot a newline here...

src/citra_qt/main.cpp
```diff
@@ -624,6 +624,39 @@ void GMainWindow::closeEvent(QCloseEvent* event) {
...
+    if (IsSingleFileDropEvent(event)) {
+        event->acceptProposedAction();
+    }
+}
+
+void GMainWindow::dragMoveEvent(QDragMoveEvent* event) {
+    event->acceptProposedAction();
+}
+
+bool GMainWindow::ConfirmChangeGame() {
```
jroweboy: Maybe in a follow up PR, you could reuse ConfirmChangeGame to ask when they click "Open File" while a game is running if they want to change games

src/citra_qt/main.cpp
jfmherokiller: oh cmon lol

src/citra_qt/main.cpp
jfmherokiller: not a bad idea ill see if i can do it

jfmherokiller: @jroweboy ill begin working on that pull request idea now

PR #2574, opened 2017-02-16
HID: move enable_accelerometer/gyroscope_count initialization into Init() function

### LINKED ISSUE #2556: The Legend of Zelda: Ocarina of Time - Can't play the Ocarina!!!
Well, I'm sorry if I mess up with this post, I'm not American... 

The emulator runs Zelda Smoothly, but the problem is when I try to play the ocarina. The game simply freezes, and I have to restart the emulator to it go back to normal...

The Citra Version is Nightly Build 2889372

wwylele:
Should fix #2556


src/core/hle/service/hid/hid.cpp
```diff
@@ -32,8 +32,8 @@ static u32 next_touch_index;
 static u32 next_accelerometer_index;
 static u32 next_gyroscope_index;
 
-static int enable_accelerometer_count = 0; // positive means enabled
-static int enable_gyroscope_count = 0;     // positive means enabled
+static int enable_accelerometer_count; // positive means enabled
```
bunnei: nit: There is no reason to change these, IMO... everything should always be initialized.

yuriks: These are initialized to 0 either way (as all globals are), this change is a no-op.

PR #2577, opened 2017-02-17
Qt cleanup

yuriks:


mailwl:
btw, Qt is upated to 5.8...

src/citra_qt/main.cpp
```diff
@@ -290,12 +290,12 @@ bool GMainWindow::LoadROM(const std::string& filename) {
...
     Core::System& system{Core::System::GetInstance()};
 
-    const Core::System::ResultStatus result{system.Load(render_window, filename)};
+    const Core::System::ResultStatus result{system.Load(render_window, filename.toStdString())};
 
     if (result != Core::System::ResultStatus::Success) {
         switch (result) {
         case Core::System::ResultStatus::ErrorGetLoader:
-            LOG_CRITICAL(Frontend, "Failed to obtain loader for %s!", filename.c_str());
+            LOG_CRITICAL(Frontend, "Failed to obtain loader for %s!", filename.toUtf8().data());
```
wwylele: I had to look for documentation to confirm `data()`is null-terminated... can be just `filename.toStdString().c_str()` which is clearer and does the same thing.

paulconrad12345:
The latest bleeding build won't open :(? It crash

jfmherokiller:
sorry for not making IsSingleFileDropEvent static in the first place

src/citra_qt/main.cpp
yuriks: Fair enough, I'll change it.

Issue #2510, opened 2017-01-31
Clang-format bypass control on git hook


Kloen:
Actually, there is a way to bypass the clang-format. I open this issue because I think that someone willing to contribute could attempt to fix this.

Steps to repeat:

1. Write something non compilant with clang-format rules.
2. git add -A
3. git commit -m "Test"
4. Enjoy the clang-format error
5. Fix clang-format on file and save it
6. Don't do git add, and directly use git commit -m "X"
7. Enjoy your non-compilant code on your branches, or PR's

jroweboy:
This sounds like a non issue to me personally. Running the pre commit hook is for convenience, but its not meant to force you to comply. I mean, a git commit --no-verify would just not run the commit hook as well, and there are several times where I don't want to struggle to get my clang format working just to make a commit to a local branch before switching branches or something. At the end of the day, when the code is PRed, the linux build will fail if its not formatted correctly and you just don't merge the code to master until its properly formatted. In theory thats what should happen at least :)

jfmherokiller:
well i hopefully made it compatable with github desktop for linux and mac as long as clang-format is installed.

MerryMage:
pre-commit hook no longer runs clang-format locally as of #2579.

PR #2579, opened 2017-02-17
hook: remove clang-format check

wwylele:
This is constantly causing problems (#2155, #2510 and so on) and is not that useful. We already have format check in CI.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -24,20 +24,3 @@ If you know what you are doing, you can try 'git commit --no-verify' to bypass t
 END
     exit 1
 fi
-
-for f in $(git diff --name-only --diff-filter=ACMRTUXB --cached); do
-    if ! echo "$f" | egrep -q "[.](cpp|h)$"; then
-        continue
-    fi
-    if ! echo "$f" | egrep -q "^src/"; then
-        continue
-    fi
-    d=$(clang-format "$f" | diff -u "$f" -)
-    if ! [ -z "$d" ]; then
-        echo "!!! $f not compliant to coding style, here is the fix:"
-        echo "$d"
-        fail=1
-    fi
-done
-
-exit "${fail-0}"


paulconrad12345:
The latest bleeding edge stopped working

Kloen:
Mmh... Could we simply run clang-format and update required files on hook, instead of fully drop it to CI? If it's installed on our environments obviously:

image
_FlowChart by Paint_

Main reasons to suggest this:

- Sometimes we just forgot about running it.
- New users could still contribute without having it installed at first.
- Not a hard requirement for short / long term contributors.

jfmherokiller:
we could also have it edit the files in place

wwylele:
I don't like it that a pre-commit hook _edits_ the code. I myself often intentionally commit some middle-stage code that is not format-compliant, and only do clang-format at the final rebasing. We should just give advices (already done in CONTRIBUING.md), and allow users to choose their own time and tools to do formatting. Speaking of tools, this is another issue that clang-format can be installed in user's computer with different name and form - it can be a plug-in in an IDE which cannot be called directly. I don't think we can handle this well in a pre-commit hook.

Also, I treat the current pre-commit hook as broken, so I want to remove it asap, and after that we can sit down to discuss how to make a better one. Sorry for being not constructive, but I want to avoid  bikesheding that was already happened in some previous issues and PRs

yuriks:
I think something that would be more useful to do would be to factor the CI-time formatting check into a script, then people can run it locally to check first if they want. (And maybe even add it to the hook themselves!)

jroweboy:
> I don't like it that a pre-commit hook edits the code.

Wait, I thought it doesn't edit the code? It should just print out what the fix is, not actually change the files. 

That said, I totally think this is a good change. (I would even go as far to remove the rest of the whitespace checks but I don't care either way)

wwylele:
@jroweboy 

>Wait, I thought it doesn't edit the code? It should just print out what the fix is, not actually change the files.

I was replying to @Kloen and @jfmherokiller

jfmherokiller:
@wwylele i recently just found the -i command option and started using it to shut up the prehook lol

Issue #2447, opened 2017-01-17
Zelda Ocarina of Time crashing when the deku tree is talking.


Ncapron:
When you leave the deku tree after beating Ghoma you got the heart, that gives you a 4th heart for your HP. And then you go back outside the tree and he says something like "do you want me to tell you more?" you have the choice between yes or no. and Then the game crash with this on the console. 

Btw : I'm playing on mac with Nightly Build - 55c91ae version

''
> ped buffers are unimplemented at the moment
> [ 125.866441] Audio.DSP <Error> audio_core/hle/source.cpp:DequeueBuffer:259: Looped buffers are unimplemented at the moment
> [ 126.094656] Audio.DSP <Error> audio_core/hle/source.cpp:DequeueBuffer:259: Looped buffers are unimplemented at the moment
> [ 126.628569] Audio.DSP <Error> audio_core/hle/source.cpp:DequeueBuffer:259: Looped buffers are unimplemented at the moment
> [ 126.700078] Audio.DSP <Error> audio_core/hle/source.cpp:DequeueBuffer:259: Looped buffers are unimplemented at the moment
> [ 126.894703] Audio.DSP <Error> audio_core/hle/source.cpp:DequeueBuffer:259: Looped buffers are unimplemented at the moment
> [ 127.785783] Debug <Critical> video_core/renderer_opengl/gl_rasterizer.cpp:operator():1084: Assertion Failed!
> Uniform block size did not match! Got 0, expected 960
> Trace/BPT trap: 5
> logout
> Saving session...
> ...copying shared history...
> ...saving history...truncating history files...
> ...completed.
> Deleting expired sessions...none found.
> 
> [Opération terminée]
''

prasoc:
Could you provide me with the saved session and detail what version/region of OoT you are using so I can reproduce this issue

Thanks,
James

Ncapron:
what do you mean by saved session?

I'm playing with the 3DS OoT the Europe one with Nightly Build - 55c91ae version for citra.

Hope it will be usefull,

Nicolas

RavenHome1:
i think he means the game save file (upload it here), also this issue was posted a lot on forums but it was gone with the forum updates

prasoc:
yup, just to be able to quickly test and debug the issue. I can get the right ROM version, but a quick save that I can load will do wonders for fixing

Ncapron:
you'll think i'm dumb :D but the save is with the .3DS or there is a folder somewhere just for the save fild

[...]

PR #2581, opened 2017-02-18
OpenGL: Check if uniform block exists before updating it

JayFoxRox:
Proposed solution for #2447 where `shader_data` is optimized out by the shader compiler for trivial shaders. It would be nice if someone who can reproduce the issue tests this ( @TTFH ? ).

I wasn't sure wether to update the uniforms even if there was no valid `shader_data` variable. I decided it would be a waste of processing time, given that we don't access the updated values. This might cause issues if the sync functions do more than updating the uniform block (which they shouldn't in this code path).

I also couldn't remember why we even sync the uniforms in that part of the code. We also do it in `NotifyPicaRegisterChanged` so that might be something to look into :man_shrugging:


TTFH:
This fix the crash, but why is `block_index` invalid in some hardware? also you could add a `LOG_CRITICAL()` when it happen.

JayFoxRox:
I've explained this in https://github.com/citra-emu/citra/issues/2447#issuecomment-280751709 .
If alpha-testing is set to NEVER the shader is reduced to `main() { discard; }`. So it will never actually access `shader_data`. As the variable is not necessary the shader compiler (which is part of the driver) can optimize it away. `glGetUniformBlockIndex` will look for a non-existing `shader_data` variable and will get `GL_INVALID_INDEX` [= -1] (as described in the man page).

We used to ignore that return value and queried the block size for `GL_INVALID_INDEX` which would in turn return a garbage value. Comparing that garbage value against the expected value would fail most of the time.

Dumb shader compilers will not optimize away the `shader_data` variable and will therefore still return the proper size of the uniform block.

So this is the proper fix for the issue at hand. It's comparable to a `nullptr` check for an optional object.

---

This should not be logged because it's just an optimization in GL drivers. GL was designed with this in mind so checking for `GL_INVALID_INDEX` is enough.
If you check the lines slightly above this one you'll also find that we do the same for uniform variables (which have a location of -1 if they don't exist). Only if the texture units are used the respective variable will have a location and be updated.

However, I've also created issue #2582 to figure out if we can ignore some useless draw calls altogether. So we should probably warn if alpha-test is NEVER (instead of panic'ing when the `shader_data` is optimized away as a result of it) long before passing this down to the GL renderer. There are probably other (valid) rendering situations which could cause a missing `shader_data` (in the future).
So this is a different issue.

PR #2587, opened 2017-02-20
Replace built-in Profiler with indicators in status bar

yuriks:
Removes the old vestigial profiler and replaces it with a more useful set of indicators in a new status bar.

Screenshots: [1](https://cloud.githubusercontent.com/assets/341401/23110547/0d17af84-f6d7-11e6-94c8-ee1c50ea7960.png) [2](https://cloud.githubusercontent.com/assets/341401/23110549/0d19d26e-f6d7-11e6-8d9c-2352eb78ea20.png) [3](https://cloud.githubusercontent.com/assets/341401/23110551/0d1beedc-f6d7-11e6-8361-db9cec8e19c1.png) [4](https://cloud.githubusercontent.com/assets/341401/23110548/0d190992-f6d7-11e6-8a7d-55165f8320fc.png) [5](https://cloud.githubusercontent.com/assets/341401/23110550/0d1abeb8-f6d7-11e6-8bb0-66ccce90e3e9.png)


Dragios:
The spacing between the different variable is kind of too packed. Can you increase for easier readability?

yuriks:
@Dragios I can try to play around with the spacing, but I'd like to see screenshots of how it looks in other platforms first so I can have a better idea of how each system treats those.

legoj15:
This is how it looks on Ubuntu 16.10
screenshot from 2017-02-20 01-15-25

src/citra_qt/main.ui
```diff
@@ -101,7 +102,6 @@
    <addaction name="menu_View"/>
    <addaction name="menu_Help"/>
   </widget>
-  <widget class="QStatusBar" name="statusbar"/>
```
wwylele: Are there any particular reasons not to define the status bar layout in the ui file?

src/core/hle/kernel/server_session.h
```diff
@@ -5,6 +5,8 @@
 #pragma once
 
 #include <string>
+#include <memory>
+
```
wwylele: We don't add new line between include groups since clang-format

src/core/hle/kernel/thread.h
```diff
@@ -9,9 +9,9 @@
 #include <vector>
 #include <boost/container/flat_map.hpp>
 #include <boost/container/flat_set.hpp>
+
```
wwylele: ditto

src/citra_qt/main.cpp
```diff
@@ -614,6 +624,23 @@ void GMainWindow::OnCreateGraphicsSurfaceViewer() {
...
+
+    auto results = Core::System::GetInstance().GetAndResetPerfStats();
+
+    emu_speed_label->setText(tr("Speed: %1%").arg(results.emulation_speed * 100.0, 0, 'f', 2));
+    game_fps_label->setText(tr("Game: %1 FPS").arg(results.game_fps, 0, 'f', 1));
+    emu_frametime_label->setText(tr("Frame: %1 ms").arg(results.frametime * 1000.0, 0, 'f', 2));
+
+    emu_speed_label->setVisible(true);
+    game_fps_label->setVisible(true);
+    emu_frametime_label->setVisible(true);
```
wwylele: Why not move these `setVisible` to `BootGame` ?

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -7,11 +7,13 @@
 #include <cstdlib>
 #include <memory>
 #include <glad/glad.h>
+
```
wwylele: ditto. no new line

src/core/perf_stats.cpp
```diff
@@ -0,0 +1,55 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <chrono>
+
```
wwylele: ditto. Will stop pointing out the same case

src/core/perf_stats.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "core/core_timing.h"
```
wwylele: I probably missed something but is this ever used?

wwylele:
The displayed Frame is still counting the frame limiter. `FrameLimiter` lives in `VBlankCallBack`, outside `SwapBuffers`, thus included in the system frame time

src/citra_qt/main.ui
yuriks: As far as I can tell, this is impossible to define in the .ui file. You can add a statusbar, but not any contents to it. Since calling `statusBar()` already creates one for you it's pointless to have it in the .ui file.

src/citra_qt/main.cpp
yuriks: I don't want them to be visible before they have content in them.

src/core/hle/kernel/server_session.h
yuriks: I'm a dummy. Gonna fix all of those.

src/core/perf_stats.cpp
yuriks: Not anymore. Removing.

yuriks:
@wwylele I noticed that but forgot to fix it. Thanks for the reminder, I'll see if I can move the frame limiter.

Dragios:
Shouldn't 60 FPS = 16.67ms frame time?
What I see so far is it displays between 16.60ms - 16.63ms.
Luxor
And another thing, in this game it is showing 31-33FPS. Is this some kind of CPU timing bug?

yuriks:
@Dragios It's normal for the measurement to have a bit of error, since the stats are polled by the UI asynchronously (and so you'll often get "partial" frames not considered in the stats.)

Do you have the frame limiter on? Try the latest commit I just pushed.

@wwylele Updated with new implementation of the frame limiter.

@legoj15 Can you get me a new screenshot with the latest version?

Dragios:
Yup, I have the framelimiter on previously. This time the same game have a fixed 32FPS. Isn't all console games are locked in 30 or 60FPS only? If this game shows 32 FPS which means it is rendering faster than it should (which is why I suspect earlier could be some CPU timing bug)?

Other games I tested so far shows either 30/60 FPS provided if full speed.

Also the frames now are counted based on max FPS? But it seems like it did not work that way. The frame time increase when I uncheck lock frame rate.

legoj15:
I'd like to send a screenshot of the latest revision, but I can't seem to get it to build:

    citra/src/core/perf_stats.cpp: In member function ‘void Core::FrameLimiter::DoFrameLimiting(u64)’:
    citra/src/core/perf_stats.cpp:93:14: error: ‘std::this_thread’ has not been declared
         std::this_thread::sleep_for(frame_limiting_delta_err);
              ^~~~~~~~~~~
(This is the same error for the Linux Travis CI build)

yuriks:
@legoj15 Should be fixed now.

legoj15:
This is what the latest revision looks like:
screenshot from 2017-02-22 16-06-06

yuriks:
Thanks. I'm not sure how to get rid of the frames around the labels, I already tried explicitly disabling them in the code. Someone on Linux should try to tweak it so it looks better if possible.

wwylele:
I tried different frame style settings but the frame is always there on my Ubuntu...
though I am personally fine with the frame.

bunnei:
The borders look really silly on Ubuntu... but this is still a much nicer replacement for the old profiler, so LGTM

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -145,21 +145,16 @@ void RendererOpenGL::SwapBuffers() {
...
 
     // Swap buffers
     render_window->PollEvents();
     render_window->SwapBuffers();
 
-    prev_state.Apply();
-
-    profiler.BeginFrame();
+    Core::System::GetInstance().frame_limiter.DoFrameLimiting(CoreTiming::GetGlobalTimeUs());
+    Core::System::GetInstance().perf_stats.BeginSystemFrame();
```
yuriks: I could, but that would leave a much larger amount of code "untimed" between Begin and EndSystemFrame (i.e. all the OpenGL state switching and code to draw the framebuffers to the screen), and I'd like to keep that as restricted to just v-sync and the frame limiter as possible so that the measurement is more accurate.

bunnei: Fair enough

Dragios:
Any chance this can be included in BE builds?

yuriks:
@Dragios I'm planning to merge it soon anyways.

PR #2589, opened 2017-02-23
[UI] Modify recursive scanning label

Helios747:
I have no idea how translations are handled. Correct me if I didn't do it right.

Anyways, the old string's wording is a bit complex. Use simple words for users.


Helios747:
That was something I wanted to address in a later PR.

I am a very technical person. I could not even figure out how to add a directory for a few minutes of browsing the UI. That should take seconds.

This was mostly a "I was browsing the UI to see what needed work and this took a literal 5 seconds"

PR #2590, opened 2017-02-23
Use gzip instead of lzma on macOS releases

jroweboy:
A common report from macOS users is they can't figure out how to unzip the
program. This will allow them to double click to unzip the archive which
is what users on macOS expect.


linkmauve:
How much bigger will these archives become?  Could you change that only for OS X and let Linux users get the better compression rate?

jroweboy:
Here are the numbers (all of the numbers provided are in KB)

Release | LZMA | GZIP | GZIP -9
------- | ------ | ----- | -----
Linux | 3417 | 6058 | 5963
macOS | 9234 | 12374 | 12332

So it looks like about a flat 3MB increase on each release. If this is significant enough, I can easily change it to use lzma on linux still.

yuriks:
LZMA is very common in Linux, so I think we should keep it. Changing it on mac is a bummer too, but I'm not the one who has to support people trying to use the files.

jroweboy:
Updated. macOS now uses gzip and linux still uses xz

PR #2593, opened 2017-02-25
Doxygen: Amend minor issues

lioncash:
Corrects a few issues with regards to Doxygen documentation, for example:

- Incorrect parameter referencing.
- Missing param tags.
- Typos in param tags.

and a few minor other issues.


src/citra_qt/hotkeys.h
```diff
@@ -29,6 +29,8 @@ void RegisterHotkey(const QString& group, const QString& action,
 /**
  * Returns a QShortcut object whose activated() signal can be connected to other QObjects' slots.
  *
+ * @param group  General group this hotkey belongs to (e.g. "Main Window", "Debugger").
```
bunnei: nit: Extra space

src/core/gdbstub/gdbstub.cpp
```diff
@@ -230,6 +230,7 @@ static void GdbHexToMem(u8* dest, const u8* src, size_t len) {
  * Convert a u32 into a gdb-formatted hex string.
  *
  * @param dest Pointer to buffer to store output hex string characters.
+ * @param v    Value to convert.
```
bunnei: Should we be lining up the text like this? Just curious because in most places I don't think we are...

lioncash: I only did that because it's what Doxygen's output itself does. I can change it back if it's preferable.

PR #2594, opened 2017-02-25
IR: separate functions of each port to their own files

wwylele:
I am going to make circle pad pro IR implementation, and this is the first step. Just moving code around but not changing anything

Unlike those of other services, The three ports of IR services, "ir:u", "ir:USER" and "ir:rst", expose totally different set of functions and are served for different works: "ir:u" is for low level IR communication; "ir:rst" is for low level handling of new 3DS additional buttons and c-stick (not actually IR related); "ir:USER" is the high level interface above the other two. They don't "share" functions and status, so it makes no sense to put all functions into one file.

### DIFF
--- a/src/core/hle/service/ir/ir.cpp
+++ b/src/core/hle/service/ir/ir.cpp
@@ -2,9 +2,6 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include "core/hle/kernel/event.h"
-#include "core/hle/kernel/kernel.h"
-#include "core/hle/kernel/shared_memory.h"
 #include "core/hle/service/ir/ir.h"
 #include "core/hle/service/ir/ir_rst.h"
 #include "core/hle/service/ir/ir_u.h"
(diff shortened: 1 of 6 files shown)


src/core/hle/service/ir/ir_rst.cpp
```diff
@@ -2,12 +2,34 @@
...
+
+/**
+ * IR::GetHandles service function
+ *  Outputs:
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : Translate header, used by the ARM11-kernel
+ *      3 : Shared memory handle
+ *      4 : Event handle
+ */
+void GetHandles(Service::Interface* self) {
```
mailwl: no need `Service::`

src/core/hle/service/ir/ir.cpp
```diff
@@ -14,101 +11,18 @@
...
-
-    cmd_buff[1] = RESULT_SUCCESS.raw;
-
-    LOG_WARNING(Service_IR, "(STUBBED) called");
-}
-
 void Init() {
-    using namespace Kernel;
-
     AddService(new IR_RST_Interface);
```
mailwl: no need `_Interface`

wwylele: Is removing `_Interface` prefered? Could you link me the original discussion?
And anyway this is out of scope of the PR, so I will probably do this next time

vonchenplus:
LGTM

PR #2600, opened 2017-02-25
fix log entry in timer::signal

B3n30:
This will fix an error caused by 
LOG_TRACE(Kernel, "Timer %08" PRIx64 " fired", timer_handle); in timer.cpp


wwylele:
If adding a parameter is just for logging, then I think we should just remove that log or log something else.

@Subv probably you know the right way to fix this.

B3n30:
I first thought about changing the log entry but I think without the timer_handle the log entry is pointless. But since Subv programmed a log entry there I just didn't want to remove it. Thus I went for the way with the additional parameter. But I'll change it in any way you want to.

yuriks:
Uh, there's something very weird with this change. If the parameter didn't exist before, then by definition it'll always have the same default value of 0 every time now. So the log is really pointless. The log should probably use the object id of the timer, not the scheduling handle, since the object id is stable.

B3n30:
Timer::Signal was for now only called from Timer::Set when initial = 0 and from Timer::TimerCallback. So I changed this calls in my PR, too. 
But you are right: To log the handle dosen't bring a lot information. I changed the command to log the name of the timer.

src/core/hle/kernel/timer.cpp
```diff
@@ -78,7 +78,7 @@ void Timer::WakeupAllWaitingThreads() {
 }
 
 void Timer::Signal(int cycles_late) {
-    LOG_TRACE(Kernel, "Timer %08" PRIx64 " fired", timer_handle);
+    LOG_TRACE(Kernel, "Timer %s fired", GetName().c_str());
```
Subv: Use GetObjectId() and the proper format specifier for its type

B3n30: Fixed to only log ObjectId and excluded the GetName. I hope that is what you had in mind

Issue #2522, opened 2017-02-03
Optimize log spam


mailwl:
Main idea:
if previous log string equal current one, do not show current, but add counter, and then log string changes, log previous + number of counter. i.e:
```
[  11.423697] HW.GPU <Error> video_core\command_processor.cpp:Pica::CommandProcessor::WritePicaReg:364: Invalid VS uniform index 127
[  12.246473] HW.GPU <Error> video_core\command_processor.cpp:Pica::CommandProcessor::WritePicaReg:364: Invalid VS uniform index 127 (repeats 238 times)
```

FernandoS27:
uhh maybe generate a unique id depending on where it was called and compare that id to last id. Comparing strings is a no go for logs unless you want to exchange lag for span.

mailwl:
comparing strings in mooore faster console output

FernandoS27:
yeah but you are forgetting of the massive trace and debug logs that are already filtered.

mailwl:
check after filtering? don't remember, is it possible.
when i need to log all service logs, except spamming, i did such dirty hack:
https://github.com/mailwl/citra/blob/sys-services/src/core/hle/service/service.cpp#L132
but it do not work in common

FernandoS27:
you could add new filtering option to config.ini and apply it on the log filter in main.cpp

log/filter.h allows you to filter based on class (module which is writing to log).

mailwl:
for sure i can do what i want with opensource (if you look at code i link, i log very useful register LR to see where in code was service call).
Idea for Citra codebase - get rid spam

Hexagon12:
Is this for #2473 ?

mailwl:
No, i hate pokemon games, it was just a random game to introduce the problem

Hexagon12:
Still, it will fix that issue also. Don't it?

mailwl:
no. in this case it just hide issue a little.
it was bad example

PR #2605, opened 2017-02-28
Common/Log: get rid repeated log entries

mailwl:
idea from #2522

example of optimized log:

> [   8.244176] Service.Y2R <Warning> core\hle\service\y2r_u.cpp:Service::Y2R::SetTransferEndInterrupt:248: (STUBBED) called
> [   8.244176] Service.Y2R <Warning> core\hle\service\y2r_u.cpp:Service::Y2R::SetTransferEndInterrupt:248: (STUBBED) called (repeats 287 times)
...
> [  17.753575] Service <Error> core\hle\service\service.cpp:Service::Interface::HandleSyncRequest:95: unknown / unimplemented function 'InvalidateDCache': port=dsp::DSP, cmd_buff[1]=0x94AC80, cmd_buff[2]=0x4000, cmd_buff[3]=0x0, cmd_buff[4]=0xFFFF8001
> [  17.753575] Service <Error> core\hle\service\service.cpp:Service::Interface::HandleSyncRequest:95: unknown / unimplemented function 'InvalidateDCache': port=dsp::DSP, cmd_buff[1]=0x94AC80, cmd_buff[2]=0x4000, cmd_buff[3]=0x0, cmd_buff[4]=0xFFFF8001 (repeats 10 times)

### DIFF
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -147,6 +147,15 @@ void LogMessage(Class log_class, Level log_level, const char* filename, unsigned
     Entry entry = CreateEntry(log_class, log_level, filename, line_nr, function, format, args);
     va_end(args);
 
+    static Entry prev_entry{};
+    if (entry == prev_entry) {
+        prev_entry.count++;
+        prev_entry.timestamp = entry.timestamp;
+        return;
+    }
+    if (prev_entry.count != 0)
+        PrintColoredMessage(prev_entry);
     PrintColoredMessage(entry);
+    prev_entry = std::move(entry);
 }
-}
+} // namespace Log
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -151,6 +151,10 @@ void LogMessage(Class log_class, Level log_level, const char* filename, unsigned
     if (entry == prev_entry) {
         prev_entry.count++;
         prev_entry.timestamp = entry.timestamp;
+        if (prev_entry.count == 100) {
+            PrintColoredMessage(prev_entry);
+            prev_entry.count = 0;
+        }
         return;
     }
     if (prev_entry.count != 0)
--- a/src/common/logging/backend.h
+++ b/src/common/logging/backend.h
@@ -7,7 +7,6 @@
 #include <chrono>
 #include <cstdarg>
 #include <string>
-#include <utility>
 #include "common/logging/log.h"
 
 namespace Log {
@@ -24,12 +23,18 @@ struct Entry {
     Level log_level;
     std::string location;
     std::string message;
+    u32 count{0};
 
     Entry() = default;
     Entry(Entry&& o) = default;
 
     Entry& operator=(Entry&& o) = default;
 };
+
+inline bool operator==(const Entry& lhs, const Entry& rhs) {
+    return (lhs.log_class == rhs.log_class) && (lhs.log_level == rhs.log_level) &&
+           (lhs.location == rhs.location) && (lhs.message == rhs.message);
+}
 
 /**
  * Returns the name of the passed log class as a C-string. Subclasses are separated by periods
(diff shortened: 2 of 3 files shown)


wwylele:
If we fold log like this, such situation might occur:

A bug cause game run into an infinite loop, in which a service function is called repeatedly. Due to the log folding, only the first line of the service function call is logged, and the second line with `(repeats N times)` is never outputted because no other log triggers that before the user stop the game. Then we cannot tell that the function is actually called repeatedly from the log user provided, which means we lose a chance to find the problem.

mailwl:
log every 100th line, now log looks like this:

> [   5.282268] Service.Y2R <Warning> core\hle\service\y2r_u.cpp:Service::Y2R::SetTransferEndInterrupt:248: (STUBBED) called
> [   5.967325] Service.Y2R <Warning> core\hle\service\y2r_u.cpp:Service::Y2R::SetTransferEndInterrupt:248: (STUBBED) called (repeats 100 times)
> [   6.774516] Service.Y2R <Warning> core\hle\service\y2r_u.cpp:Service::Y2R::SetTransferEndInterrupt:248: (STUBBED) called (repeats 100 times)
> [   7.275288] Service.Y2R <Warning> core\hle\service\y2r_u.cpp:Service::Y2R::SetTransferEndInterrupt:248: (STUBBED) called (repeats 87 times)
...
> [  19.955374] Service <Error> core\hle\service\service.cpp:Service::Interface::HandleSyncRequest:95: unknown / unimplemented function 'InvalidateDCache': port=dsp::DSP, cmd_buff[1]=0x94AC80, cmd_buff[2]=0x4000, cmd_buff[3]=0x0, cmd_buff[4]=0xFFFF8001
> [  30.423245] Service <Error> core\hle\service\service.cpp:Service::Interface::HandleSyncRequest:95: unknown / unimplemented function 'InvalidateDCache': port=dsp::DSP, cmd_buff[1]=0x94AC80, cmd_buff[2]=0x4000, cmd_buff[3]=0x0, cmd_buff[4]=0xFFFF8001 (repeats 100 times)
> [  40.798378] Service <Error> core\hle\service\service.cpp:Service::Interface::HandleSyncRequest:95: unknown / unimplemented function 'InvalidateDCache': port=dsp::DSP, cmd_buff[1]=0x94AC80, cmd_buff[2]=0x4000, cmd_buff[3]=0x0, cmd_buff[4]=0xFFFF8001 (repeats 100 times)

jmerdich:
I toyed around with another idea based off this PR: not hiding dupes, but 'live-updating' them using terminal features, like CLI progress bars would do.

Proof of concept GIF

[Here's the branch.](https://github.com/jmerdich/citra/tree/terminal-folding) Tested on Windows and Linux, should work on Mac too.

mailwl:
@jmerdich looks nice, crash with int division by zero when log to file with:
`./citra-qt 1> console.log 2>&1`

jmerdich:
That's odd; I'm not getting that on my machines (Arch Linux and Windows 10). Can you describe your environment?

JayFoxRox:
Maybe it's just me, but I'd rather have easy to parse, plain and simple logs.
This folding etc. could easily be put in external tools or happen on piped output.

Ideally we'll only have to use the text log for debugging new platforms and in the plain frontend (SDL) / if no GUI is available.

Rather than messing with the text log (which is slow on Windows anyway), I'd rather like to see a log for the GUI which could do all sorts of fancy things, not limited by the console.

Just my two cents on the whole topic.

jmerdich:
I agree that adding complexity to the current behavior should not be a decision made lightly, especially if a GUI log will pop up in the future. I'd personally take a convenient log to stderr, though, and being able to actually read fast-moving repeated messages is a major boost for users and devs alike.

As a side note this does fail gracefully. Piped output is completely identical to current behavior, except that duped lines have "[Repeated Nx]" in front of the log message, as in the last line of the gif.

LittleWhite-tb:
I have seen some time ago that updating the line of the terminal like you did (\r if I am right) comes with some performance issues. Can you check that your new log behaviour is negligible in terms of performances.

PR #2606, opened 2017-03-01
ir: implement circle pad pro

wwylele:
Circle pad pro (ZL, ZR and right circle pad) is implemented, along with the minimum set of required ir:USER service functions. For documentation of them please refer to [3dbrew:Circle Pad Pro](https://www.3dbrew.org/wiki/Circle_Pad_Pro) and [3dbrew:IRUSER Shared Memory](https://www.3dbrew.org/wiki/IRUSER_Shared_Memory). 

Some trivial service functions, such as `ReceiveIrnop` and `GetConnectionStatus`, are not implemented, because, from my observation, games usually read data/status directly from the shared memory instead of calling these functions. To minimize the diff, I left them unimplemented. If any games are found using them, I will implement them.

Note on new 3DS input: new 3DS additional HID hardware is mainly handled by ir:rst service, which is not implemented here. However, new 3DS can redirect ir:USER function call to access the additional HID, and most commercial games still uses ir:USER service to access new 3DS additional HID, so this PR should work for them. ir:rst is mostly used by homebrew apps, and is actually more trivial to implement, so I leave it to another PR (not necessary by me)

~~This branch is currently based on my input refactor branch (PR #2497) in order to get a clean input interface to start with (In the current master ZL, ZR and c-stick is mixed into HID service). **Please only review the last commit!**. If this PR get reviewed faster than the input PR, I will rebase this to the master branch. The only expected change at that time is in `ExtraHID::SendHIDStatus`~~

### DIFF
--- a/src/core/hle/service/ir/extra_hid.cpp
+++ b/src/core/hle/service/ir/extra_hid.cpp
@@ -1,0 +1,231 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/alignment.h"
+#include "common/bit_field.h"
+#include "common/string_util.h"
+#include "core/core_timing.h"
+#include "core/hle/service/ir/extra_hid.h"
+#include "core/settings.h"
+
+namespace Service {
+namespace IR {
+
+enum class RequestID : u8 {
+    /**
+     * ConfigureHIDPolling request
+     * Starts HID input polling, or changes the polling interval if it is already started.
+     *  Inputs:
+     *     byte 0: request ID
+     *     byte 1: polling interval in ms
+     *     byte 2: unknown
+     */
+    ConfigureHIDPolling = 1,
+
+    /**
+     * ReadCalibrationData request
+     * Reads the calibration data stored in circle pad pro.
+     *  Inputs:
+     *     byte 0: request ID
+     *     byte 1: expected response time in ms?
+     *     byte 2-3: data offset (aligned to 0x10)
+     *     byte 4-5: data size (aligned to 0x10)
+     */
+    ReadCalibrationData = 2,
+
+    // TODO(wwylele): there are three more request types (id = 3, 4 and 5)
+};
+
+enum class ResponseID : u8 {
+
+    /**
+     * PollHID response
+     * Sends current HID status
+     *  Output:
+     *     byte 0: response ID
+     *     byte 1-3: Right circle pad position. This three bytes are two little-endian 12-bit
+     *         fields. The first one is for x-axis and the second one is for y-axis.
+     *     byte 4: bit[0:4] battery level; bit[5] ZL button; bit[6] ZR button; bit[7] R button
+     *         Note that for the three button fields, the bit is set when the button is NOT pressed.
+     *     byte 5: unknown
+     */
+    PollHID = 0x10,
+
+    /**
+     * ReadCalibrationData response
+     * Sends the calibration data reads from circle pad pro.
+     *  Output:
+     *     byte 0: resonse ID
+     *     byte 1-2: data offset (aligned to 0x10)
+     *     byte 3-4: data size (aligned to 0x10)
+     *     byte 5-...: calibration data
+     */
+    ReadCalibrationData = 0x11,
+};
+
+ExtraHID::ExtraHID(SendFunc send_func) : IRDevice(send_func) {
+    LoadInputDevices();
+
+    // The data below was retrieved from a New 3DS
+    // TODO(wwylele): this data is probably writable (via request 3?) and thus should be saved to
+    // and loaded from somewhere.
+    calibration_data = std::array<u8, 0x40>{{
+        // 0x00
+        0x00, 0x00, 0x08, 0x80, 0x85, 0xEB, 0x11, 0x3F,
+        // 0x08
+        0x85, 0xEB, 0x11, 0x3F, 0xFF, 0xFF, 0xFF, 0xF5,
+        // 0x10
+        0xFF, 0x00, 0x08, 0x80, 0x85, 0xEB, 0x11, 0x3F,
+        // 0x18
+        0x85, 0xEB, 0x11, 0x3F, 0xFF, 0xFF, 0xFF, 0x65,
+        // 0x20
+        0xFF, 0x00, 0x08, 0x80, 0x85, 0xEB, 0x11, 0x3F,
+        // 0x28
+        0x85, 0xEB, 0x11, 0x3F, 0xFF, 0xFF, 0xFF, 0x65,
+        // 0x30
+        0xFF, 0x00, 0x08, 0x80, 0x85, 0xEB, 0x11, 0x3F,
+        // 0x38
+        0x85, 0xEB, 0x11, 0x3F, 0xFF, 0xFF, 0xFF, 0x65,
+    }};
+
+    hid_polling_callback_id =
+        CoreTiming::RegisterEvent("ExtraHID::SendHIDStatus", [this](u64, int cycles_late) {
+            SendHIDStatus();
+            CoreTiming::ScheduleEvent(msToCycles(hid_period) - cycles_late,
+                                      hid_polling_callback_id);
+        });
+}
+
+ExtraHID::~ExtraHID() {
+    OnDisconnect();
+}
+
+void ExtraHID::OnConnect() {}
+
+void ExtraHID::OnDisconnect() {
+    CoreTiming::UnscheduleEvent(hid_polling_callback_id, 0);
+}
+
+void ExtraHID::HandleConfigureHIDPollingRequest(const std::vector<u8>& request) {
+    if (request.size() != 3) {
+        LOG_ERROR(Service_IR, "Wrong request size (%zu): %s", request.size(),
+                  Common::ArrayToString(request.data(), request.size()).c_str());
+        return;
+    }
+
+    // Change HID input polling interval
+    CoreTiming::UnscheduleEvent(hid_polling_callback_id, 0);
+    hid_period = request[1];
+    CoreTiming::ScheduleEvent(msToCycles(hid_period), hid_polling_callback_id);
+}
+
+void ExtraHID::HandleReadCalibrationDataRequest(const std::vector<u8>& request_buf) {
+    struct ReadCalibrationDataRequest {
+        RequestID request_id;
+        u8 expected_response_time;
+        u16_le offset;
+        u16_le size;
+    };
+    static_assert(sizeof(ReadCalibrationDataRequest) == 6,
+                  "ReadCalibrationDataRequest has wrong size");
+
+    if (request_buf.size() != sizeof(ReadCalibrationDataRequest)) {
+        LOG_ERROR(Service_IR, "Wrong request size (%zu): %s", request_buf.size(),
+                  Common::ArrayToString(request_buf.data(), request_buf.size()).c_str());
+        return;
+    }
+
+    ReadCalibrationDataRequest request;
+    std::memcpy(&request, request_buf.data(), sizeof(request));
+
+    const u16 offset = Common::AlignDown(request.offset, 16);
+    const u16 size = Common::AlignDown(request.size, 16);
+
+    if (offset + size > calibration_data.size()) {
+        LOG_ERROR(Service_IR, "Read beyond the end of calibration data! (offset=%u, size=%u)",
+                  offset, size);
+        return;
+    }
+
+    std::vector<u8> response(5);
+    response[0] = static_cast<u8>(ResponseID::ReadCalibrationData);
+    std::memcpy(&response[1], &request.offset, sizeof(request.offset));
+    std::memcpy(&response[3], &request.size, sizeof(request.size));
+    response.insert(response.end(), calibration_data.begin() + offset,
+                    calibration_data.begin() + offset + size);
+    Send(response);
+}
+
+void ExtraHID::OnReceive(const std::vector<u8>& data) {
+    switch (static_cast<RequestID>(data[0])) {
+    case RequestID::ConfigureHIDPolling:
+        HandleConfigureHIDPollingRequest(data);
+        break;
+    case RequestID::ReadCalibrationData:
+        HandleReadCalibrationDataRequest(data);
+        break;
+    default:
+        LOG_ERROR(Service_IR, "Unknown request: %s",
+                  Common::ArrayToString(data.data(), data.size()).c_str());
+        break;
+    }
+}
+
+void ExtraHID::SendHIDStatus() {
+    if (is_device_reload_pending.exchange(false))
+        LoadInputDevices();
+
+    struct {
+        union {
+            BitField<0, 8, u32_le> header;
+            BitField<8, 12, u32_le> c_stick_x;
+            BitField<20, 12, u32_le> c_stick_y;
+        } c_stick;
+        union {
+            BitField<0, 5, u8> battery_level;
+            BitField<5, 1, u8> zl_not_held;
+            BitField<6, 1, u8> zr_not_held;
+            BitField<7, 1, u8> r_not_held;
+        } buttons;
+        u8 unknown;
+    } response;
+    static_assert(sizeof(response) == 6, "HID status response has wrong size!");
+
+    constexpr int C_STICK_CENTER = 0x800;
+    // TODO(wwylele): this value is not accurately measured. We currently assume that the axis can
+    // take values in the whole range of a 12-bit integer.
+    constexpr int C_STICK_RADIUS = 0x7FF;
+
+    float x, y;
+    std::tie(x, y) = c_stick->GetStatus();
+
+    response.c_stick.header.Assign(static_cast<u8>(ResponseID::PollHID));
+    response.c_stick.c_stick_x.Assign(static_cast<u32>(C_STICK_CENTER + C_STICK_RADIUS * x));
+    response.c_stick.c_stick_y.Assign(static_cast<u32>(C_STICK_CENTER + C_STICK_RADIUS * y));
+    response.buttons.battery_level.Assign(0x1F);
+    response.buttons.zl_not_held.Assign(!zl->GetStatus());
+    response.buttons.zr_not_held.Assign(!zr->GetStatus());
+    response.buttons.r_not_held.Assign(1);
+    response.unknown = 0;
+
+    std::vector<u8> response_buffer(sizeof(response));
+    memcpy(response_buffer.data(), &response, sizeof(response));
+    Send(response_buffer);
+}
+
+void ExtraHID::RequestInputDevicesReload() {
+    is_device_reload_pending.store(true);
+}
+
+void ExtraHID::LoadInputDevices() {
+    zl = Input::CreateDevice<Input::ButtonDevice>(
+        Settings::values.buttons[Settings::NativeButton::ZL]);
+    zr = Input::CreateDevice<Input::ButtonDevice>(
+        Settings::values.buttons[Settings::NativeButton::ZR]);
+    c_stick = Input::CreateDevice<Input::AnalogDevice>(
+        Settings::values.analogs[Settings::NativeAnalog::CStick]);
+}
+
+} // namespace IR
+} // namespace Service
--- a/src/core/hle/service/ir/extra_hid.h
+++ b/src/core/hle/service/ir/extra_hid.h
@@ -1,0 +1,48 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <atomic>
+#include "core/frontend/input.h"
+#include "core/hle/service/ir/ir_user.h"
+
+namespace Service {
+namespace IR {
+
+/**
+ * An IRDevice emulating Circle Pad Pro or New 3DS additional HID hardware.
+ * This device sends periodic udates at a rate configured by the 3DS, and sends calibration data if
+ * requested.
+ */
+class ExtraHID final : public IRDevice {
+public:
+    explicit ExtraHID(SendFunc send_func);
+    ~ExtraHID();
+
+    void OnConnect() override;
+    void OnDisconnect() override;
+    void OnReceive(const std::vector<u8>& data) override;
+
+    /// Requests input devices reload from current settings. Called when the input settings change.
+    void RequestInputDevicesReload();
+
+private:
+    void SendHIDStatus();
+    void HandleConfigureHIDPollingRequest(const std::vector<u8>& request);
+    void HandleReadCalibrationDataRequest(const std::vector<u8>& request);
+    void LoadInputDevices();
+
+    u8 hid_period;
+    int hid_polling_callback_id;
+    std::array<u8, 0x40> calibration_data;
+    std::unique_ptr<Input::ButtonDevice> zl;
+    std::unique_ptr<Input::ButtonDevice> zr;
+    std::unique_ptr<Input::AnalogDevice> c_stick;
+    std::atomic<bool> is_device_reload_pending;
+};
+
+} // namespace IR
+} // namespace Service
--- a/src/core/hle/service/ir/ir_user.cpp
+++ b/src/core/hle/service/ir/ir_user.cpp
@@ -2,110 +2,481 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <memory>
+#include <boost/crc.hpp>
+#include <boost/optional.hpp>
+#include "common/string_util.h"
+#include "common/swap.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/kernel/shared_memory.h"
+#include "core/hle/service/ir/extra_hid.h"
 #include "core/hle/service/ir/ir.h"
 #include "core/hle/service/ir/ir_user.h"
 
 namespace Service {
 namespace IR {
 
-static Kernel::SharedPtr<Kernel::Event> conn_status_event;
-static Kernel::SharedPtr<Kernel::SharedMemory> transfer_shared_memory;
+// This is a header that will present in the ir:USER shared memory if it is initialized with
+// InitializeIrNopShared service function. Otherwise the shared memory doesn't have this header if
+// it is initialized with InitializeIrNop service function.
+struct SharedMemoryHeader {
+    u32_le latest_receive_error_result;
+    u32_le latest_send_error_result;
+    // TODO(wwylele): for these fields below, make them enum when the meaning of values is known.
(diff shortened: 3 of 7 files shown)


src/core/hle/service/ir/crc8.cpp
```diff
@@ -0,0 +1,39 @@
+// Copyright 2017 Citra Emulator Project
```
mailwl: is IR the only place where crc8 used? maybe move it to... common, for example

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include "core/settings.h"
+
+namespace Service {
+namespace IR {
+
+ExtraHID::ExtraHID(SendFunc send_func) : IRDevice(send_func) {
+    LoadInputDevices();
+
+    // The data below is got from a New 3DS
+    calibration_data = std::array<u8, 0x40>{{
```
mailwl: to the future: this should be in config (cfg:u), to be able to tune via System Settings

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,164 @@
...
+    size = Common::AlignDown(size, 16);
+
+    if (offset + size > calibration_data.size()) {
+        LOG_ERROR(Service_IR, "Read beyond the end of calibration data! (offset=%u, size=%u)",
+                  offset, size);
+        return;
+    }
+
+    std::vector<u8> response;
+    response.push_back(0x11); // Fixed header value
```
mailwl: magic number. Enum with Request/response IDs

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,308 @@
...
 namespace IR {
 
-static Kernel::SharedPtr<Kernel::Event> conn_status_event;
-static Kernel::SharedPtr<Kernel::SharedMemory> transfer_shared_memory;
+/// Structure of data stored in ir:USER shared memory
+struct SharedMemory {
+    u32 latest_receive_error_result;
+    u32 latest_send_error_result;
+    // TODO(wwylele): for these fields below, make them enum when the meaning of values is known.
+    u8 connection_status;
```
mailwl: it's enum, looks like (after reversing 0004013000003302.app):
1: TryingToConnect
2: Connected

src/core/hle/service/ir/crc8.cpp
wwylele: Currently IR is the only user of crc8 known to me... I hesitated when moving it to common because there are lots of crc8 standard, and putting this specific one in common will probably confuse others if they want to use a different standard.

Anyway we can move this out any time if we find another user of crc8 in 3DS system.

src/core/hle/service/ir/extra_hid.cpp
wwylele: I don't know. At least for old 3DS, I believe this data is stored in circle pad pro internal memory (otherwise it doesn't need to do IR communication to read this data).

src/core/hle/service/ir/crc8.h
```diff
@@ -0,0 +1,22 @@
...
+#include "common/common_types.h"
+
+namespace Service {
+namespace IR {
+
+/**
+ * Calculates CRC-8-CCITT for the given data.
+ * @param data The pointer to the begining of the data.
+ * @param size Size of the data.
+ * @params the CRC value of the data.
```
bunnei: returns

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,175 @@
...
+                  Common::ArrayToString(data.data(), data.size()).c_str());
+        return;
+    }
+
+    u16 offset;
+    u16 size;
+    std::memcpy(&offset, data.data() + 2, sizeof(u16));
+    std::memcpy(&size, data.data() + 4, sizeof(u16));
+
+    // align
```
bunnei: nit: Useless comment

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,175 @@
...
+        u8 unknown;
+    } response;
+    static_assert(sizeof(response) == 6, "HID status response has wrong size!");
+
+    response.c_stick.header.Assign(static_cast<u8>(ResponseID::ReadHIDStatus));
+    float x, y;
+    std::tie(x, y) = c_stick->GetStatus();
+    response.c_stick.c_stick_x.Assign(0x800 + 0x7FF * x);
+    response.c_stick.c_stick_y.Assign(0x800 + 0x7FF * y);
+    response.buttons.battery.Assign(0x1F);
```
bunnei: Where are the above constants from? Can we define these somewhere for better readability?

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -144,14 +456,40 @@ IR_User_Interface::IR_User_Interface() {
...
 
 void ShutdownUser() {
-    transfer_shared_memory = nullptr;
+    if (connected_device) {
+        connected_device->Disconnect();
+        connected_device = nullptr;
+    }
+
+    extra_hid = nullptr;
+
```
bunnei: nit: Unnecessary extra space

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -101,11 +351,73 @@ static void GetConnectionStatusEvent(Interface* self) {
...
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    IPC::RequestBuilder rb(Kernel::GetCommandBuffer(), 0x02, 1, 0);
+    rb.Push(RESULT_SUCCESS);
 
-    LOG_WARNING(Service_IR, "(STUBBED) called");
+    LOG_INFO(Service_IR, "called");
+}
+
+/**
+ * IR::SendIrNopservice function
```
bunnei: SendIrNop service function

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,310 @@
...
+static std::unique_ptr<ExtraHID> extra_hid;
+static IRDevice* connected_device;
+static std::unique_ptr<CircularBuffer> receive_buffer;
+
+/// Wraps the payload into packet and puts it to the receive buffer
+static void PutToReceive(const std::vector<u8>& payload) {
+    LOG_TRACE(Service_IR, "called, data=%s",
+              Common::ArrayToString(payload.data(), payload.size()).c_str());
+    size_t size = payload.size();
+    DEBUG_ASSERT(size < 0x4000);
```
bunnei: Whats the significance of this size? Can it be sizeof() some struct? Or constant?

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,310 @@
...
+
+    std::vector<u8> packet;
+
+    // Builds packet header. For the format info:
+    // https://www.3dbrew.org/wiki/IRUSER_Shared_Memory#Packet_structure
+    packet.push_back(0xA5);
+    packet.push_back(*(shared_memory->GetPointer(offsetof(SharedMemory, network_id))));
+    if (size < 0x40) {
```
bunnei: ditto with the magic numbers

bunnei: (fwiw and/or add comments explaining what this means, what if someone needs to update this code in a few years and 3dbew no longer exists?)

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,310 @@
...
-    u32 send_buff_size = cmd_buff[4];
-    u32 unk2 = cmd_buff[5];
-    u8 baud_rate = cmd_buff[6] & 0xFF;
-    Kernel::Handle handle = cmd_buff[8];
-
-    if (Kernel::g_handle_table.IsValid(handle)) {
-        transfer_shared_memory = Kernel::g_handle_table.Get<Kernel::SharedMemory>(handle);
-        transfer_shared_memory->name = "IR:TransferSharedMemory";
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x18, 6, 2);
+    u32 shared_buff_size = rp.Pop<u32>();
```
bunnei: These can all be const

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,310 @@
...
+    u32 info_offset;
+    u32 buffer_offset;
+    u32 max_packet_count;
+    u32 max_data_size;
+};
+
+static Kernel::SharedPtr<Kernel::Event> conn_status_event, send_event, receive_event;
+static Kernel::SharedPtr<Kernel::SharedMemory> shared_memory;
+static std::unique_ptr<ExtraHID> extra_hid;
+static IRDevice* connected_device;
```
Subv: Why is this one not an unique_ptr?

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,175 @@
...
+enum class ResponseID : u8 {
+    ReadHIDStatus = 0x10,
+    ReadCalibrationData = 0x11,
+};
+
+ExtraHID::ExtraHID(SendFunc send_func) : IRDevice(send_func) {
+    LoadInputDevices();
+
+    // The data below is got from a New 3DS
+    calibration_data = std::array<u8, 0x40>{{
```
wwylele: The reason why it is a non-const member if `ExtraHID` is that it can be modified by the circle pad pro calibrating applet (probably via request 3, not tested and implemented here). I will add more comment about this in the code.

src/core/hle/service/ir/ir_user.cpp
wwylele: This pointer points to a device only when it is connect, and doesn't take the ownership of it, because the device itself doesn't create/destroy by connecting/disconnecting. An IR device may still want to persist some state when not connected. Take `ExtraHID` for example, it want to persist the calibration data (explained above), and also it registers an event in `CoreTiming`, which is hard to un- and re-register if the device get destroyed by disconnecting.

wwylele:
Addressed @bunnei 's comments

makotech222:
I'm getting weird behavior with this on Majoras mask. When C-stick is bound to my xbox 360 gamepad right stick, the camera in game goes a little wild and constantly rotates around player. Binding the cstick to keyboard fixes it.

wwylele:
@makotech222 That's because the config is wrong. Previously I (or someone else) recommend using `c_stick=engine:sdl,joystick:0,axis_x=2,axis_y=3` as configuration for c-stick, but axis 2 is actually an invalid axis and the correct one should be `c_stick=engine:sdl,joystick:0,axis_x=3,axis_y=4`. Anyway, this problem is in the input device interface, and has nothing to do with this PR.

wwylele:
Rebased onto current master

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+enum class ResponseID : u8 {
+    ReadHIDStatus = 0x10,
+    ReadCalibrationData = 0x11,
+};
+
+ExtraHID::ExtraHID(SendFunc send_func) : IRDevice(send_func) {
+    LoadInputDevices();
+
+    // The data below is got from a New 3DS
```
JayFoxRox: "The data below was retrieved from a New 3DS"

wwylele:
addressed comments, rebased and replaced `rp.Pop<u32>() & 0xFF` with `rp.Pop<u8>()`

JayFoxRox:
This is lacking a hardware test in my opinion. I tried testing it some days ago but only found 1 homebrew for accessing the circle pad pro. However, that did not work as it used the non-implemented interface.
We should have an issue about this as it can be very confusing for homebrew authors (not strictly necessary for Citra as we should worry more about commercial games = low priority).

I was mostly wondering how to configure this and how analogs translate to buttons in the current input PR.
If we currently can not map an axis [like xinput LT / RT] to buttons [like ZL, ZR] we should have a follow-up PR for such things or at least an issue.

---

Also: Reminding @bunnei about their review (changes requested but never accepted those - hence github still shows "changes requested").

wwylele:
Yes I admit that lack of hwtest is a problem, which I can't fix well myself because I don't have a circle pad pro / a new 3DS. I sent my friends some misc hwtests, asking them to describe what they observed to confirm my theories. These hwtests are very fragmentary, hacky and not self-descriptive. [Here is one of them that shows most ir:USER functions](https://gist.github.com/wwylele/a1667da027399e283d7ca752e8e97935): it monitors the shared memory while user presses A to execute the service function. However, this test program has several issues:
 - it fails to connect to the circle pad pro on old 3DS for unknown reason. Probably has something to do with waking up circle pad pro from sleep mode.
 - when testing on new 3DS, it needs to be built with modified ctrulib, with irrstInit commented out in hidInit, because ir:rst services blocks ir:USER service to connect to new 3DS HID.
 - for the same reason above, the program needs to be built as CIA, or be executed through a homebrew entry point that doesn't automatically grab ir:rst service (Hint: Hombrew launcher loader does)

For these annoying issues and the fact I don't have a required hardware to play around with them myself, I am not able to make an integrated hwtest. I would appreciate it if someone can help me make a good one.

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+void ExtraHID::SendHIDStatus(int cycles_late) {
+    if (is_device_reload_pending.exchange(false))
+        LoadInputDevices();
+
+    struct {
+        union {
+            BitField<0, 8, u32> header;
+            BitField<8, 12, u32> c_stick_x;
+            BitField<20, 12, u32> c_stick_y;
```
yuriks: u32_le

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,181 @@
...
+    static_assert(sizeof(response) == 6, "HID status response has wrong size!");
+
+    constexpr int C_STICK_CENTER = 0x800;
+    // TODO(wwylele): this value is not accurately measured. We currently assume that the axis can
+    // take values in the whole range of a 12-bit integer.
+    constexpr int C_STICK_RADIUS = 0x7FF;
+
+    response.c_stick.header.Assign(static_cast<u8>(ResponseID::ReadHIDStatus));
+    float x, y;
+    std::tie(x, y) = c_stick->GetStatus();
```
yuriks: Move this before all the assigns, looks a bit better when they're not broken up by this.

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,181 @@
...
+        LoadInputDevices();
+
+    struct {
+        union {
+            BitField<0, 8, u32> header;
+            BitField<8, 12, u32> c_stick_x;
+            BitField<20, 12, u32> c_stick_y;
+        } c_stick;
+        union {
+            BitField<0, 5, u8> battery;
```
yuriks: `battery_level`

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,181 @@
...
+            BitField<5, 1, u8> zl;
```
yuriks: `zl_held` (or `zl_not_held`?), etc.

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,181 @@
...
+    response.buttons.zl.Assign(!zl->GetStatus());
+    response.buttons.zr.Assign(!zr->GetStatus());
+    response.buttons.r.Assign(1);
+    response.unknown = 0;
+
+    std::vector<u8> response_buffer(sizeof(response));
+    memcpy(response_buffer.data(), &response, sizeof(response));
+    send_func(response_buffer);
+
+    CoreTiming::ScheduleEvent(msToCycles(hid_period) - cycles_late, send_callback);
```
yuriks: This periodic scheduling should go into the lambda that calls this function. This decouples this function ("Send one hid status update") from the timing.

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,181 @@
...
+}
+
+void ExtraHID::ReloadInputDevices() {
```
yuriks: Confusing naming, it's very similar to "LoadInputDevices". Name it "RequestInputDevicesReload".

src/core/hle/service/ir/extra_hid.h
```diff
@@ -0,0 +1,42 @@
...
+/// An IRDevice emulating Circle Pad Pro or New 3DS additional HID hardware.
+class ExtraHID final : public IRDevice {
+public:
+    explicit ExtraHID(SendFunc send_func);
+    ~ExtraHID();
+
+    void Connect() override;
+    void Disconnect() override;
+    void Receive(const std::vector<u8>& data) override;
+    void ReloadInputDevices();
```
yuriks: Add a blank line before to separate this one from the overrides and document what it does.

src/core/hle/service/ir/extra_hid.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <array>
+#include <atomic>
+#include "core/frontend/input.h"
+#include "core/hle/service/ir/ir_user.h"
+
+namespace Service {
+namespace IR {
+
+/// An IRDevice emulating Circle Pad Pro or New 3DS additional HID hardware.
```
yuriks: Expand this comment. Explain in general terms how this device works ("sends periodic updates at a rate configured by the 3DS")

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,323 @@
...
+    u8 trying_to_connect_status;
+    u8 connection_role;
+    u8 machine_id;
+    u8 connected;
+    u8 network_id;
+    u8 initialized;
+    u8 unknown;
+
+    // This is not the end of the shared memory. It is followed by a receive buffer and a send
+    // buffer. We handle receive buffer by the CircularBuffer class. For the send buffer, because
```
yuriks: *in* the CircularBuffer class

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,323 @@
...
 #include "core/hle/service/ir/ir.h"
 #include "core/hle/service/ir/ir_user.h"
 
 namespace Service {
 namespace IR {
 
-static Kernel::SharedPtr<Kernel::Event> conn_status_event;
-static Kernel::SharedPtr<Kernel::SharedMemory> transfer_shared_memory;
+/// Structure of data stored in ir:USER shared memory
+struct SharedMemory {
```
yuriks: SharedMemoryHeader, perhaps?

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,323 @@
...
+
+/// A manager of the circular buffer in the shared memory.
+class CircularBuffer {
+public:
+    CircularBuffer(Kernel::SharedPtr<Kernel::SharedMemory> shared_memory_, u32 info_offset_,
+                   u32 buffer_offset_, u32 max_packet_count_, u32 buffer_size)
+        : shared_memory(shared_memory_), info_offset(info_offset_), buffer_offset(buffer_offset_),
+          max_packet_count(max_packet_count_),
+          max_data_size(buffer_size - sizeof(PacketInfo) * max_packet_count_) {
+
```
yuriks: I think the newline can be ommited here. The indents won't align anyway, (the statement is indented 2 spaces less than `max_data_size...`) so the start of the block will be visible.

src/core/hle/service/ir/ir_user.cpp
wwylele: I am going to rename this as `SharedStatusInfo`, which was found to be an optional part in the shared memory in later RE.

src/core/hle/service/ir/extra_hid.cpp
```diff
@@ -0,0 +1,181 @@
...
+    ReadHIDStatus = 0x10,
+    ReadCalibrationData = 0x11,
+};
+
+ExtraHID::ExtraHID(SendFunc send_func) : IRDevice(send_func) {
+    LoadInputDevices();
+
+    // The data below was retrieved from a New 3DS
+    // TODO(wwylele): this data is probably writable (via request 3?) and thus should be saved to
+    // and loaded from somewhere.
```
wwylele: Probably no. 
 - For o3ds, this data is more likely stored in CPP internal memory, otherwise this calibration data protocol doesn't need to be implemented in CPP (`ExtraHID` as equivalent) - IR service can just be implemented as reading data from cfg directly.
 - For n3ds, as far as I RE-d, I can't find any code that loads config save to this data. It is more like that the data is always initialized to the same value (as here) every time the system module boots up.

src/core/hle/service/ir/ir_user.h
```diff
@@ -4,11 +4,33 @@
...
 #pragma once
 
+#include <functional>
 #include "core/hle/service/service.h"
 
 namespace Service {
 namespace IR {
 
+/// An interface representing a device that can communicate with 3DS via ir:USER service
+class IRDevice {
```
wwylele: This `IRDevice` is used only for ir:USER interfaces. ir:u uses a completely different interfaces, and ir:rst doesn't do IR stuff at all

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,323 @@
...
+    if (receive_buffer->Put(packet)) {
+        receive_event->Signal();
+    } else {
+        LOG_ERROR(Service_IR, "receive buffer is full!");
+    }
+}
 
 /**
  * IR::InitializeIrNopShared service function
```
wwylele: documented

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,323 @@
...
-                transfer_buff_size, recv_buff_size, unk1, send_buff_size, unk2, baud_rate, handle);
+    LOG_INFO(Service_IR, "called, shared_buff_size=%u, recv_buff_size=%u, "
+                         "recv_buff_packet_count=%u, send_buff_size=%u, "
+                         "send_buff_packet_count=%u, baud_rate=%u, handle=0x%08X",
+             shared_buff_size, recv_buff_size, recv_buff_packet_count, send_buff_size,
+             send_buff_packet_count, baud_rate, handle);
 }
 
 /**
  * IR::RequireConnection service function
```
wwylele: documented

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,323 @@
...
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    rb.Push(RESULT_SUCCESS);
+    rb.PushCopyHandles(Kernel::g_handle_table.Create(Service::IR::receive_event).MoveFrom());
 
-    LOG_WARNING(Service_IR, "(STUBBED) called");
+    LOG_INFO(Service_IR, "called");
+}
+
+/**
+ * IR::GetSendEvent service function
```
wwylele: documented

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -73,26 +327,35 @@ static void RequireConnection(Interface* self) {
...
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    IPC::RequestBuilder rb(Kernel::GetCommandBuffer(), 0x09, 1, 0);
+    rb.Push(RESULT_SUCCESS);
 
-    LOG_WARNING(Service_IR, "(STUBBED) called");
+    LOG_INFO(Service_IR, "called");
 }
 
 /**
  * IR::GetConnectionStatusEvent service function
```
wwylele: documented

src/core/hle/service/ir/crc8.h
```diff
@@ -0,0 +1,22 @@
...
+#pragma once
+
+#include <cstddef>
+#include "common/common_types.h"
+
+namespace Service {
+namespace IR {
+
+/**
+ * Calculates CRC-8-CCITT for the given data.
```
wwylele: using boost.crc now

src/core/hle/service/ir/extra_hid.cpp
yuriks: Ah, didn't realize at the time that this is on the CPP side, yeah, disregard.

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,323 @@
...
+        LOG_CRITICAL(Service_IR, "invalid shared memory handle 0x%08X", handle);
+        rb.Push(ResultCode(ErrorDescription::InvalidHandle, ErrorModule::OS,
+                           ErrorSummary::WrongArgument, ErrorLevel::Permanent));
+        return;
     }
+    shared_memory->name = "IR_USER: shared memory";
+
+    receive_buffer = std::make_unique<CircularBuffer>(shared_memory, 0x10, 0x20,
+                                                      recv_buff_packet_count, recv_buff_size);
+    SharedMemory shared_memory_init;
```
wwylele: fixed

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,69 +2,323 @@
...
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+    rb.Push(RESULT_SUCCESS);
+
+    LOG_INFO(Service_IR, "called, device_id = %u", device_id);
+}
 
-    conn_status_event->Signal();
+/**
+ * IR::GetReceiveEvent service function
```
wwylele: documented

wwylele:
Addressed @yuriks' comments. Changes are:
 - more documentation
 - renaming
 - removing crc8 code and using boost.crc instead
...

Note that in `HandleReadCalibrationDataRequest` I still keep the old vector pushing way for response, because the struct would need `#pragma pack` to work

src/core/hle/service/ir/ir_user.cpp
```diff
@@ -2,110 +2,481 @@
...
+    if (!shared_memory) {
+        LOG_CRITICAL(Service_IR, "invalid shared memory handle 0x%08X", handle);
+        rb.Push(ResultCode(ErrorDescription::InvalidHandle, ErrorModule::OS,
+                           ErrorSummary::WrongArgument, ErrorLevel::Permanent));
+        return;
     }
+    shared_memory->name = "IR_USER: shared memory";
 
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    receive_buffer = boost::make_optional(
```
yuriks: The `make_optional` isn't necessary, you can just directly assign the value.

wwylele:
Fixed and squashed

Issue #2456, opened 2017-01-20
implement warning signs to notify users for common issues.


RavenHome1:
currently the forum gets flooded with people having the same exact issues like

-rom encryption

-gpu not supporting open GL 3.3

-the need to dump system archives and fonts

-lack of knowledge about citra's ability to emulate tilting motion (gyro sensor)

my suggestion would be to

-add warning to warn the user about ROM encryption (already implemented apparently but is broken ATM according to https://github.com/citra-emu/citra/issues/2268)

-warn user if their gpu does not support openGL 3.3 (same as above implemented but is broken)

-warn user about the need to dump the system home folder

this can be done by looking for  specific log errors 

[  11.818456] Debug.Emulated <Critical> core\hle\svc.cpp:SVC::Break:466: Break r
eason: PANIC

[  11.834715] Debug.Emulated <Critical> core\hle\svc.cpp:SVC::Break:450: Emulate
d program broke execution!

and 

[ 4.095480] Service.FS core/hle/service/fs/fs_user.cpp:OpenFileDirectly:130: failed to get a handle for archive archive_id=0x2345678A archive_path=[Binary: 02030100db00040000ffffff01000000]

-notify the user about the tilting support with a one time only notification (a check box that includes "do not show this again" would be nice) when a game attempts to access the gyroscope function.

this would reduce the amount of repeated posts on the forums and help find real bugs in the software

mailwl:
easy, but bored... developers never see this kind of errors :)

Subv:
Here's a fun fact: We already have logs that warn about most of these, people just ignore them.

There's a pretty yellow log that tells you that the shared font is missing.
The emulator reports the OpenGL version on startup.
The emulator shows a pink-ish error when a ROM is encrypted.

Users just don't like to read.

wwylele:
@mailwl This is not as easy as you think. Take this for example:

>warn user about the need to dump the system home folder
...
[ 4.095480] Service.FS core/hle/service/fs/fs_user.cpp:OpenFileDirectly:130: failed to get a handle for archive archive_id=0x2345678A archive_path=[Binary: 02030100db00040000ffffff01000000]

This requires a proper interface between core and frontend so that core can report a generic message via frontend-defined message box, which is not as trivial as just calling `MessageBox` where ever you want.

mailwl:
wellm frontend can provide some set of callbacks, which core just call in some situations.
For example - one for file not exists, another for swkbd need enter text...

[...]

Issue #2398, opened 2017-01-02
Puyo Puyo Chronicle fails to boot and crashes Citra


Electricstar:
Self-explanatory; Puyo Puyo Chronicle, a recently released entry in the _Puyo Puyo_ series, will immediately crash Citra when loaded.

The debug window yields

"<Critical> core/memory.cpp:operator():319: Assertion Failed!", followed by:
"Rasterizer resource cache counter overflow!".

This occurs in the most recent build I got via update, version 0.1.116

Edit: added the hash of my build.

Hash: HEAD-b79ad3b

mailwl:
The game want to open itself as archive:
`[   3.898988] Service.FS <Error> core\hle\service\fs\fs_user.cpp:Service::FS::OpenFileDirectly:130: failed to get a handle for archive archive_id=0x2345678A archive_path=[Binary: 007519008c00040001fd800001000000]`

wwylele:
@mailwl @Electricstar Could you please provide a full log?

mailwl:
http://pastebin.com/y1eJ2H5y (sorry, not master, i add logs fo log hwregs only)
same error in game Skylanders - Spyros Adventure

mailwl:
as i can see, the game can't read any file from ROM (maybe it looks for it in update or something...)

wwylele:
Rechecked this log:
`[ 3.898988] Service.FS <Error> core\hle\service\fs\fs_user.cpp:Service::FS::OpenFileDirectly:130: failed to get a handle for archive archive_id=0x2345678A archive_path=[Binary: 007519008c00040001fd800001000000]`

Rearrange the archive path:
`007519008c00040001fd800001000000` -> `0x00197500, 0x0004008C, 0x0000D81F, 0x00000001`

Notice this higher word of title ID:
`0x0004008C` <- note this, this is DLC content according to 3dbrew. It is not the game itself (in which case it should be `0x00040000`)

benderscruffy:
using the latest nightly goes ingame now
can be closed

PR #2611, opened 2017-03-05
Display QMessageBox Dialogs For Errors

TheKoopaKingdom:
As a forum/Discord moderator, we see many people ask for support when their solution is simply to dump their system files and shared font. So, what this PR does is it displays a QMessageBox dialog when one of these errors happens.

A dialog box will be shown when the following happens:
 - The shared font is missing.
 - The system/home files are missing.
 - An encrypted ROM is loaded (This was already in, but I had to fix it because something was breaking it from being shown.).
 - OpenGL requirements not being met (Again, this was already in, but I'm not sure if it was really working.).
 - Any other unknown fatal error.

Here's how it works:
 - When the error happens within core, Core::System::GetInstance().SetStatus() is called.
 - In the main loop within bootmanager.cpp,  the Core Loop returns the status. It checks if it was successful or not, and if the latter, it emits the ErrorThrown signal.
 - At this point ErrorThrown has already been connected to the OnCoreError slot.
 - OnCoreError shuts down the game, checks the error type, and shows the dialog acoordingly.

With this system in place, adding new error messages is pretty easy, add a new ResultStatus entry, set it somewhere within core, and add the handler to OnCoreError.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2611)

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -39,7 +39,7 @@ void EmuThread::run() {
 
             Core::System::ResultStatus result = Core::System::GetInstance().RunLoop();
             if (result != Core::System::ResultStatus::Success) {
-                emit ErrorThrown(result);
+                emit ErrorThrown(result, Core::System::GetInstance().GetStatusDetails());
             }
 
             was_active = running || exec_step;
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -37,7 +37,11 @@ void EmuThread::run() {
             if (!was_active)
                 emit DebugModeLeft();
 
-            Core::System::GetInstance().RunLoop();
+            Core::System::ResultStatus result = Core::System::GetInstance().RunLoop();
+            if (result != Core::System::ResultStatus::Success) {
+                emit ErrorThrown(result);
+                break;
+            }
 
             was_active = running || exec_step;
             if (!was_active && !stop_run)
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -40,7 +40,6 @@ void EmuThread::run() {
             Core::System::ResultStatus result = Core::System::GetInstance().RunLoop();
             if (result != Core::System::ResultStatus::Success) {
                 emit ErrorThrown(result);
-                break;
             }
 
             was_active = running || exec_step;
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -99,7 +99,7 @@ signals:
      */
     void DebugModeLeft();
 
-    void ErrorThrown(Core::System::ResultStatus, boost::optional<std::string>);
+    void ErrorThrown(Core::System::ResultStatus, std::string);
 };
 
 class GRenderWindow : public QWidget, public EmuWindow {
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -99,7 +99,7 @@ signals:
      */
     void DebugModeLeft();
 
-    void ErrorThrown(Core::System::ResultStatus);
+    void ErrorThrown(Core::System::ResultStatus, boost::optional<std::string>);
 };
 
 class GRenderWindow : public QWidget, public EmuWindow {
--- a/src/citra_qt/bootmanager.h
+++ b/src/citra_qt/bootmanager.h
@@ -10,6 +10,7 @@
 #include <QGLWidget>
 #include <QThread>
 #include "common/thread.h"
+#include "core/core.h"
 #include "core/frontend/emu_window.h"
 #include "core/frontend/motion_emu.h"
 
@@ -97,6 +98,8 @@ signals:
      * Qt::BlockingQueuedConnection (additionally block source thread until slot returns)
      */
     void DebugModeLeft();
+
+    void ErrorThrown(Core::System::ResultStatus);
 };
 
 class GRenderWindow : public QWidget, public EmuWindow {
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -340,9 +340,9 @@ bool GMainWindow::LoadROM(const QString& filename) {
                    "Citra. A real 3DS is required.<br/><br/>"
                    "For more information on dumping and decrypting games, please see the following "
                    "wiki pages: <ul>"
-                   "<li><a href='https://citra-emu.org/wiki/Dumping-Game-Cartridges/'>Dumping Game "
+                   "<li><a href='https://citra-emu.org/wiki/dumping-game-cartridges/'>Dumping Game "
                    "Cartridges</a></li>"
-                   "<li><a href='https://citra-emu.org/wiki/Dumping-Installed-Titles/'>Dumping "
+                   "<li><a href='https://citra-emu.org/wiki/dumping-installed-titles/'>Dumping "
                    "Installed Titles</a></li>"
                    "</ul>"));
             break;
@@ -355,7 +355,7 @@ bool GMainWindow::LoadROM(const QString& filename) {
         case Core::System::ResultStatus::ErrorVideoCore:
             QMessageBox::critical(
                 this, tr("An error occured in the video core."),
-                tr("Citra has encountered an error while running the video core,  please see the "
+                tr("Citra has encountered an error while running the video core, please see the "
                    "log for more details."
                    "For more information on accessing the log, please see the following page: "
                    "<a href='https://community.citra-emu.org/t/how-to-upload-the-log-file/296'>How "
@@ -656,9 +656,10 @@ void GMainWindow::OnCoreError(Core::System::ResultStatus result, std::string det
            "before playing.<br/><br/>For more information on dumping these files, please see the "
            "following wiki page: <a "
            "href='https://citra-emu.org/wiki/"
-           "Dumping-System-Archives-and-the-Shared-Fonts-from-a-3DS-Console/'>Dumping System "
+           "dumping-system-archives-and-the-shared-fonts-from-a-3ds-console/'>Dumping System "
            "Archives and the Shared Fonts from a 3DS Console</a>.<br/><br/>Would you like to quit "
-           "back to the game list?");
+           "back to the game list? Continuing emulation may result in crashes, corrupted save "
+           "data, or other bugs.");
     switch (result) {
     case Core::System::ResultStatus::ErrorSystemFiles: {
         QString message = "Citra was unable to locate a 3DS system archive";
@@ -689,7 +690,8 @@ void GMainWindow::OnCoreError(Core::System::ResultStatus result, std::string det
             tr("Citra has encountered a fatal error, please see the log for more details. "
                "For more information on accessing the log, please see the following page: "
                "<a href='https://community.citra-emu.org/t/how-to-upload-the-log-file/296'>How to "
-               "Upload the Log File</a>.<br/><br/>Would you like to quit back to the game list?"),
+               "Upload the Log File</a>.<br/><br/>Would you like to quit back to the game list? "
+               "Continuing emulation may result in crashes, corrupted save data, or other bugs."),
             QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
         status_message = "Fatal Error encountered";
         break;
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -301,8 +301,7 @@ bool GMainWindow::LoadROM(const QString& filename) {
 
     if (!gladLoadGL()) {
         QMessageBox::critical(this, tr("Error while starting Citra!"),
-                              tr("Failed to initialize the video core!\n\n"
-                                 "Please ensure that your GPU supports OpenGL 3.3 and that you "
+                              tr("Your GPU may not support OpenGL 3.3, or you do not"
                                  "have the latest graphics driver."));
         return false;
     }
@@ -327,18 +326,17 @@ bool GMainWindow::LoadROM(const QString& filename) {
             break;
 
         case Core::System::ResultStatus::ErrorLoader_ErrorEncrypted: {
-            // Build the MessageBox ourselves to have clickable link
-            QMessageBox popup_error;
-            popup_error.setTextFormat(Qt::RichText);
-            popup_error.setWindowTitle(tr("Error while loading ROM!"));
-            popup_error.setText(
+            QMessageBox::critical(
+                this, tr("Error while loading ROM!"),
                 tr("The game that you are trying to load must be decrypted before being used with "
                    "Citra.<br/><br/>"
-                   "For more information on dumping and decrypting games, please see: <a "
-                   "href='https://citra-emu.org/wiki/Dumping-Game-Cartridges'>https://"
-                   "citra-emu.org/wiki/Dumping-Game-Cartridges</a>"));
-            popup_error.setIcon(QMessageBox::Critical);
-            popup_error.exec();
+                   "For more information on dumping and decrypting games, please see the following "
+                   "wiki pages: <ul>"
+                   "<li><a href='https://citra-emu.org/wiki/Dumping-Game-Cartridges/'>Dumping Game "
+                   "Cartridges</a></li>"
+                   "<li><a href='https://citra-emu.org/wiki/Dumping-Installed-Titles/'>Dumping "
+                   "Installed Titles</a></li>"
+                   "</ul>"));
             break;
         }
         case Core::System::ResultStatus::ErrorLoader_ErrorInvalidFormat:
@@ -346,8 +344,16 @@ bool GMainWindow::LoadROM(const QString& filename) {
                                   tr("The ROM format is not supported."));
             break;
 
+        case Core::System::ResultStatus::ErrorOpenGL:
+            QMessageBox::critical(this, tr("Error while loading OpenGL!"),
+                                  tr("Your GPU may not support OpenGL 3.3, or you do not "
+                                     "have the latest graphics driver."));
+            break;
+
         default:
-            QMessageBox::critical(this, tr("Error while loading ROM!"), tr("Unknown error!"));
+            QMessageBox::critical(
+                this, tr("Error while loading ROM!"),
+                tr("An unknown error occured. Please see the log for more details."));
             break;
         }
         return false;
@@ -530,6 +536,9 @@ void GMainWindow::OnMenuRecentFile() {
 
 void GMainWindow::OnStartGame() {
     emu_thread->SetRunning(true);
+    qRegisterMetaType<Core::System::ResultStatus>("Core::System::ResultStatus");
+    connect(emu_thread.get(), SIGNAL(ErrorThrown(Core::System::ResultStatus)), this,
+            SLOT(OnCoreError(Core::System::ResultStatus)));
 
     ui.action_Start->setEnabled(false);
     ui.action_Start->setText(tr("Continue"));
@@ -622,14 +631,57 @@ void GMainWindow::UpdateStatusBar() {
     emu_frametime_label->setVisible(true);
 }
 
+void GMainWindow::OnCoreError(Core::System::ResultStatus result) {
+    // Waiting for the dialog to be closed before shutting down causes a segfault, maybe because of
+    // the profiler
+    ShutdownGame();
+    switch (result) {
+    case Core::System::ResultStatus::ErrorSystemFiles:
+        QMessageBox::critical(
+            this, "System Archive Not Found",
+            "Citra was unable to locate the 3DS system archive.<br/><br/>"
+            "The game you are trying to load requires additional files from your 3DS to be dumped "
+            "before playing.<br/><br/>"
+            "For more information on dumping these files, please see the following wiki page: "
+            "<a "
+            "href='https://citra-emu.org/wiki/"
+            "Dumping-System-Archives-and-the-Shared-Fonts-from-a-3DS-Console/'>Dumping System "
+            "Archives and the Shared Fonts from a 3DS Console</a>"
+            ".");
+        break;
+
+    case Core::System::ResultStatus::ErrorSharedFont:
+        QMessageBox::critical(
+            this, "Shared Fonts Not Found",
+            "Citra was unable to locate the 3DS shared fonts.<br/><br/>"
+            "The game you are trying to load requires additional files from your 3DS to be dumped "
+            "before playing.<br/><br/>"
+            "For more information on dumping these files, please see the following wiki page: "
+            "<a "
+            "href='https://citra-emu.org/wiki/"
+            "Dumping-System-Archives-and-the-Shared-Fonts-from-a-3DS-Console/'>Dumping System "
+            "Archives and the Shared Fonts from a 3DS Console</a>"
+            ".");
+        break;
+
+    case Core::System::ResultStatus::ErrorUnknown:
+        QMessageBox::critical(
+            this, "Fatal Error",
+            "Citra has encountered a fatal error, please see the log for more details.");
+        break;
+
+    default:
+        break;
+    }
+}
+
 bool GMainWindow::ConfirmClose() {
     if (emu_thread == nullptr || !UISettings::values.confirm_before_closing)
         return true;
 
-    auto answer =
-        QMessageBox::question(this, tr("Citra"), tr("Are you sure you want to close Citra?"),
-                              QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
-    return answer != QMessageBox::No;
+    return QMessageBox::question(this, tr("Citra"), tr("Are you sure you want to close Citra?"),
+                                 QMessageBox::Yes | QMessageBox::No,
+                                 QMessageBox::No) != QMessageBox::No;
 }
 
 void GMainWindow::closeEvent(QCloseEvent* event) {
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -93,6 +93,14 @@ void GMainWindow::InitializeWidgets() {
     ui.horizontalLayout->addWidget(game_list);
 
     // Create status bar
+    message_label = new QLabel();
+    // Configured separately for left alignment
+    message_label->setVisible(false);
+    message_label->setFrameStyle(QFrame::NoFrame);
+    message_label->setContentsMargins(4, 0, 4, 0);
+    message_label->setAlignment(Qt::AlignLeft);
+    statusBar()->addPermanentWidget(message_label, 1);
+
     emu_speed_label = new QLabel();
     emu_speed_label->setToolTip(tr("Current emulation speed. Values higher or lower than 100% "
                                    "indicate emulation is running faster or slower than a 3DS."));
@@ -108,7 +116,7 @@ void GMainWindow::InitializeWidgets() {
         label->setVisible(false);
         label->setFrameStyle(QFrame::NoFrame);
         label->setContentsMargins(4, 0, 4, 0);
-        statusBar()->addPermanentWidget(label);
+        statusBar()->addPermanentWidget(label, 0);
     }
     statusBar()->setVisible(true);
     setStyleSheet("QStatusBar::item{border: none;}");
@@ -437,6 +445,7 @@ void GMainWindow::ShutdownGame() {
 
     // Disable status bar updates
     status_bar_update_timer.stop();
+    message_label->setVisible(false);
     emu_speed_label->setVisible(false);
     game_fps_label->setVisible(false);
     emu_frametime_label->setVisible(false);
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -553,10 +553,9 @@ void GMainWindow::OnMenuRecentFile() {
 void GMainWindow::OnStartGame() {
     emu_thread->SetRunning(true);
     qRegisterMetaType<Core::System::ResultStatus>("Core::System::ResultStatus");
-    qRegisterMetaType<boost::optional<std::string>>("boost::optional<std::string>");
-    connect(emu_thread.get(),
-            SIGNAL(ErrorThrown(Core::System::ResultStatus, boost::optional<std::string>)), this,
-            SLOT(OnCoreError(Core::System::ResultStatus, boost::optional<std::string>)));
+    qRegisterMetaType<std::string>("std::string");
+    connect(emu_thread.get(), SIGNAL(ErrorThrown(Core::System::ResultStatus, std::string)), this,
+            SLOT(OnCoreError(Core::System::ResultStatus, std::string)));
 
     ui.action_Start->setEnabled(false);
     ui.action_Start->setText(tr("Continue"));
@@ -649,8 +648,7 @@ void GMainWindow::UpdateStatusBar() {
     emu_frametime_label->setVisible(true);
 }
 
-void GMainWindow::OnCoreError(Core::System::ResultStatus result,
-                              boost::optional<std::string> details) {
+void GMainWindow::OnCoreError(Core::System::ResultStatus result, std::string details) {
     QMessageBox::StandardButton answer;
     QString status_message;
     const QString common_message =
@@ -664,8 +662,8 @@ void GMainWindow::OnCoreError(Core::System::ResultStatus result,
     switch (result) {
     case Core::System::ResultStatus::ErrorSystemFiles: {
         QString message = "Citra was unable to locate a 3DS system archive";
-        if (details)
-            message.append(tr(": %1. ").arg(details.get().c_str()));
+        if (details != std::string())
+            message.append(tr(": %1. ").arg(details.c_str()));
         else
             message.append(". ");
         message.append(common_message);
@@ -693,7 +691,7 @@ void GMainWindow::OnCoreError(Core::System::ResultStatus result,
                "<a href='https://community.citra-emu.org/t/how-to-upload-the-log-file/296'>How to "
                "Upload the Log File</a>.<br/><br/>Would you like to quit back to the game list?"),
             QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
-        status_message = "Fatal Error encountered.";
+        status_message = "Fatal Error encountered";
         break;
     }
 
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -663,10 +663,11 @@ void GMainWindow::OnCoreError(Core::System::ResultStatus result, std::string det
     switch (result) {
     case Core::System::ResultStatus::ErrorSystemFiles: {
         QString message = "Citra was unable to locate a 3DS system archive";
-        if (details != std::string())
+        if (!details.empty()) {
             message.append(tr(": %1. ").arg(details.c_str()));
-        else
+        } else {
             message.append(". ");
+        }
         message.append(common_message);
 
         answer = QMessageBox::question(this, tr("System Archive Not Found"), message,
@@ -698,11 +699,15 @@ void GMainWindow::OnCoreError(Core::System::ResultStatus result, std::string det
     }
 
     if (answer == QMessageBox::Yes) {
-        if (emu_thread != nullptr)
+        if (emu_thread) {
             ShutdownGame();
+        }
     } else {
-        message_label->setText(status_message);
-        message_label->setVisible(true);
+        // Only show the message if the game is still running.
+        if (emu_thread) {
+            message_label->setText(status_message);
+            message_label->setVisible(true);
+        }
     }
 }
 
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -300,7 +300,7 @@ bool GMainWindow::LoadROM(const QString& filename) {
     render_window->MakeCurrent();
 
     if (!gladLoadGL()) {
-        QMessageBox::critical(this, tr("Error while starting Citra!"),
+        QMessageBox::critical(this, tr("Error while initializing OpenGL 3.3 Core!"),
                               tr("Your GPU may not support OpenGL 3.3, or you do not"
                                  "have the latest graphics driver."));
         return false;
@@ -329,7 +329,7 @@ bool GMainWindow::LoadROM(const QString& filename) {
             QMessageBox::critical(
                 this, tr("Error while loading ROM!"),
                 tr("The game that you are trying to load must be decrypted before being used with "
-                   "Citra.<br/><br/>"
+                   "Citra. A real 3DS is required.<br/><br/>"
                    "For more information on dumping and decrypting games, please see the following "
                    "wiki pages: <ul>"
                    "<li><a href='https://citra-emu.org/wiki/Dumping-Game-Cartridges/'>Dumping Game "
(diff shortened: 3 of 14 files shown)


src/core/loader/loader.h
```diff
@@ -100,11 +100,13 @@ class AppLoader : NonCopyable {
      * Loads the system mode that this application needs.
      * This function defaults to 2 (96MB allocated to the application) if it can't read the
      * information.
-     * @returns Optional with the kernel system mode
+     * @ param boost::optional<u32>* Pointer to Boost optional to store system mode.
+     * @ return Result of operation.
      */
-    virtual boost::optional<u32> LoadKernelSystemMode() {
+    virtual ResultStatus LoadKernelSystemMode(boost::optional<u32>* system_mode) {
```
wwylele: If you are outputting `system_mode` via the param, it should be just a raw pointer/reference.

Eventually we should change the return type of this function (and other functions in similar situation) to `variant<ResultStatus, u32>`. Can be in another PR, though

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -128,6 +128,11 @@ static void OpenFileDirectly(Service::Interface* self) {
 
     ResultVal<ArchiveHandle> archive_handle = OpenArchive(archive_id, archive_path);
     if (archive_handle.Failed()) {
+        if (static_cast<u32>(archive_id) == 0x2345678A) {
+            LOG_ERROR(Service_FS, "oh noes");
+            Core::System::GetInstance().throwErr(-1);
+            return;
```
wwylele: You should do either of the two here:
 - immediately stop the core, or
 - assign proper error code to `cmd_buff` so that the core won't run with unexpected response.

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -128,6 +128,11 @@ static void OpenFileDirectly(Service::Interface* self) {
 
     ResultVal<ArchiveHandle> archive_handle = OpenArchive(archive_id, archive_path);
     if (archive_handle.Failed()) {
+        if (static_cast<u32>(archive_id) == 0x2345678A) {
```
wwylele: Use `ArchiveIdCode::NCCH`. Also a note: as already suggested in the name, this archive is not only for system archive, but for general archive (including installed games etc.). For 99% of games I believe they only uses this for system archive, so this condition is fine, but don't forget suspicious cases like this: #2398

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -131,6 +132,10 @@ static void OpenFileDirectly(Service::Interface* self) {
         LOG_ERROR(Service_FS,
                   "failed to get a handle for archive archive_id=0x%08X archive_path=%s",
                   static_cast<u32>(archive_id), archive_path.DebugStr().c_str());
+        if (static_cast<u32>(archive_id) == 0x2345678A) {
```
JayFoxRox: What is this magic? Could it be a named constant maybe?

src/citra_qt/main.cpp
```diff
@@ -646,14 +654,57 @@ void GMainWindow::UpdateStatusBar() {
...
     if (emu_thread == nullptr || !UISettings::values.confirm_before_closing)
         return true;
 
-    auto answer =
-        QMessageBox::question(this, tr("Citra"), tr("Are you sure you want to close Citra?"),
-                              QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
-    return answer != QMessageBox::No;
+    return QMessageBox::question(this, tr("Citra"), tr("Are you sure you want to close Citra?"),
+                                 QMessageBox::Yes | QMessageBox::No,
+                                 QMessageBox::No) != QMessageBox::No;
```
JayFoxRox: This is harder to read than the previous code imo.

src/core/loader/loader.h
```diff
@@ -100,11 +100,13 @@ class AppLoader : NonCopyable {
...
-     * @returns Optional with the kernel system mode
+     * @param boost::optional<u32> Reference to Boost optional to store system mode.
+     * @ return Result of operation.
      */
-    virtual boost::optional<u32> LoadKernelSystemMode() {
+    virtual ResultStatus LoadKernelSystemMode(boost::optional<u32>& system_mode) {
         // 96MB allocated to the application.
-        return 2;
+        system_mode = 2;
+        return ResultStatus::Success;
```
JayFoxRox: nit: I believe it's better to return a pair here instead of modifying `system_mode` by reference (where it's easy to miss what's going on).

src/citra_qt/main.cpp
```diff
@@ -646,14 +654,57 @@ void GMainWindow::UpdateStatusBar() {
...
+            "href='https://citra-emu.org/wiki/"
+            "Dumping-System-Archives-and-the-Shared-Fonts-from-a-3DS-Console/'>Dumping System "
+            "Archives and the Shared Fonts from a 3DS Console</a>"
+            ".");
+        break;
+
+    case Core::System::ResultStatus::ErrorUnknown:
+        QMessageBox::critical(
+            this, "Fatal Error",
+            "Citra has encountered a fatal error, please see the log for more details.");
```
JayFoxRox: I believe we have an article (wiki or forums?) about dumping logs

src/citra_qt/main.cpp
```diff
@@ -341,27 +340,33 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
         }
         case Core::System::ResultStatus::ErrorLoader_ErrorInvalidFormat:
             QMessageBox::critical(this, tr("Error while loading ROM!"),
                                   tr("The ROM format is not supported."));
             break;
 
+        case Core::System::ResultStatus::ErrorOpenGL:
+            QMessageBox::critical(this, tr("Error while loading OpenGL!"),
+                                  tr("Your GPU may not support OpenGL 3.3, or you do not "
+                                     "have the latest graphics driver."));
```
JayFoxRox: Can be stricter imo. Something about "Citra not working until GPU driver or hardware is upgraded."

src/citra_qt/main.cpp
```diff
@@ -341,27 +340,33 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
-                   "citra-emu.org/wiki/Dumping-Game-Cartridges</a>"));
-            popup_error.setIcon(QMessageBox::Critical);
-            popup_error.exec();
+                   "For more information on dumping and decrypting games, please see the following "
+                   "wiki pages: <ul>"
+                   "<li><a href='https://citra-emu.org/wiki/Dumping-Game-Cartridges/'>Dumping Game "
+                   "Cartridges</a></li>"
+                   "<li><a href='https://citra-emu.org/wiki/Dumping-Installed-Titles/'>Dumping "
+                   "Installed Titles</a></li>"
+                   "</ul>"));
```
JayFoxRox: Can mention that a 3DS is required in the process so we prevent idiots from asking questions.

src/citra_qt/main.cpp
```diff
@@ -315,8 +315,7 @@ bool GMainWindow::LoadROM(const QString& filename) {
 
     if (!gladLoadGL()) {
         QMessageBox::critical(this, tr("Error while starting Citra!"),
-                              tr("Failed to initialize the video core!\n\n"
-                                 "Please ensure that your GPU supports OpenGL 3.3 and that you "
+                              tr("Your GPU may not support OpenGL 3.3, or you do not"
                                  "have the latest graphics driver."));
```
JayFoxRox: Could be more precise: "Could not initialize OpenGL 3.3 Core." or something.
(Mostly so we don't have to check what went wrong when someone asks for that error message)

src/core/core.cpp
```diff
@@ -139,7 +151,7 @@ System::ResultStatus System::Init(EmuWindow* emu_window, u32 system_mode) {
     GDBStub::Init();
 
     if (!VideoCore::Init(emu_window)) {
-        return ResultStatus::ErrorVideoCore;
+        return ResultStatus::ErrorOpenGL;
```
JayFoxRox: The old constant made more sense: Video Core not being able to init must not be related to OpenGL at all.

src/citra_qt/main.cpp
```diff
@@ -335,7 +335,7 @@ bool GMainWindow::LoadROM(const QString& filename) {
             break;
 
         case Core::System::ResultStatus::ErrorSystemMode:
-            LOG_CRITICAL(Frontend, "Failed to load ROM!");
+            LOG_CRITICAL(Frontend, "Failed to load ROM!!!");
```
JayFoxRox: Not really necessary + doesn't fit to the rest of this commit.

src/citra_qt/bootmanager.cpp
```diff
@@ -37,9 +37,17 @@ void EmuThread::run() {
...
-                emit ErrorThrown(result);
-                break;
+            if (result != Core::System::ResultStatus::Success)
+            {
+                ready = false;
+                emit ErrorThrown(result, &stop_run, &ready);
+                // TODO: there's probably a better way to wait for the dialog to be closed
+                while (!ready);
+                if (stop_run)
+                    break;
```
JayFoxRox: yes, there is a better way and this could isn't going to fly.

src/citra_qt/bootmanager.cpp
```diff
@@ -37,9 +37,17 @@ void EmuThread::run() {
...
+
+                // Reset status to success so we dont run into the same error every time
+                Core::System::GetInstance().SetStatus(Core::System::ResultStatus::Success);
```
JayFoxRox: Shouldn't this happen at the start [or every error-free exit] of `RunLoop` instead?

src/citra_qt/bootmanager.h
```diff
@@ -4,11 +4,11 @@
 
 #pragma once
 
-#include <atomic>
```
JayFoxRox: nit: Why was this removed? std::atomic is still used below

src/citra_qt/main.cpp
```diff
@@ -646,14 +655,57 @@ void GMainWindow::UpdateStatusBar() {
     emu_frametime_label->setVisible(true);
 }
 
+void GMainWindow::OnCoreError(Core::System::ResultStatus result) {
+    // Waiting for the dialog to be closed before shutting down causes a segfault, maybe because of
+    // the profiler
```
JayFoxRox: This feels like it should be a github issue, not just a comment somewhere in the code?

src/citra_qt/main.cpp
```diff
@@ -646,14 +655,57 @@ void GMainWindow::UpdateStatusBar() {
...
+        break;
+
+    case Core::System::ResultStatus::ErrorUnknown:
+        QMessageBox::critical(
+            this, "Fatal Error",
+            "Citra has encountered a fatal error, please see the log for more details.");
+        break;
+
+    default:
+        break;
```
JayFoxRox: With the ability to add new Errors now, shouldn't this be handled similar to `ErrorUnknown`?

src/citra_qt/bootmanager.cpp
TheKoopaKingdom: This was from before Jrowe looked at the code for me, it's already much more improved in a later commit.

src/core/hle/service/fs/fs_user.cpp
TheKoopaKingdom: In my testing, games that required the system files always failed with that archive handle. I intended on waiting for input from a reviewer familiar with that, not sure exactly what that is/means.

RavenHome1:
closes https://github.com/citra-emu/citra/issues/2456

src/core/hle/service/fs/fs_user.cpp
wwylele: Use `ArchiveIdCode::NCCH`. Also a note: as already suggested in the name, this archive is not only for system archive, but for general archive (including installed games etc.). For 99% of games I believe they only uses this for system archive, so this condition is fine, but don't forget suspicious cases like this: #2398

src/core/loader/loader.h
```diff
@@ -100,12 +100,12 @@ class AppLoader : NonCopyable {
      * Loads the system mode that this application needs.
      * This function defaults to 2 (96MB allocated to the application) if it can't read the
      * information.
-     * @ param boost::optional<u32>* Pointer to Boost optional to store system mode.
+     * @param boost::optional<u32> Reference to Boost optional to store system mode.
      * @ return Result of operation.
      */
-    virtual ResultStatus LoadKernelSystemMode(boost::optional<u32>* system_mode) {
+    virtual ResultStatus LoadKernelSystemMode(boost::optional<u32>& system_mode) {
```
wwylele: If you are outputting `system_mode` via the param, it should be just a raw pointer/reference.

Eventually we should change the return type of this function (and other functions in similar situation) to `variant<ResultStatus, u32>`. Can be in another PR, though

src/core/loader/loader.h
wwylele: Also, if you are outputting `system_mode` via the param, it should be just a raw pointer/reference.

Eventually we should change the return type of this function (and other functions in similar situation) to `variant<ResultStatus, u32>`. Can be in another PR, though

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -131,6 +132,10 @@ static void OpenFileDirectly(Service::Interface* self) {
...
+            Core::System::GetInstance().SetStatus(Core::System::ResultStatus::ErrorSystemFiles);
+            return;
```
wwylele: You should do either of the two here:
 - immediately terminate the core, or
 - assign proper error code to `cmd_buff` so that the core won't run with garbage response.

src/core/loader/loader.h
TheKoopaKingdom: Alright, for now I'll use std::pair, not sure if boost::variant would be better, but the header failed to compile when I included it.

jroweboy:
@RavenHome1 I suggest opening another issue about tilt controls. I agree that it is a recurring problem and it'd be nice to inform the user somehow, but thats way out of scope for this PR.

src/citra_qt/main.cpp
```diff
@@ -341,27 +340,34 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
         }
         case Core::System::ResultStatus::ErrorLoader_ErrorInvalidFormat:
             QMessageBox::critical(this, tr("Error while loading ROM!"),
                                   tr("The ROM format is not supported."));
             break;
 
+        case Core::System::ResultStatus::ErrorVideoCore:
+            QMessageBox::critical(this, tr("Error while initializing OpenGL 3.3 Core!"),
+                                  tr("Your GPU may not support OpenGL 3.3, or you do not "
+                                     "have the latest graphics driver."));
```
JayFoxRox: This is still misleading.

Reminder: VideoCore is just the name for our graphics emulation portion. It does not necessarily use OpenGL 3.3 Core (especially in the future).
There may be other errors in the future by the VideoCore subsystem which are not related to OpenGL.

All you can say at this point is: "An unknown error occured in the VideoCore subsystem" (which is probably not something we should say to users). It's probably best to suggest the user to ask for help and mention that the emulator reported an error in VideoCore.

(The other OpenGL error message in case of glad init failure is fine as it's strictly GL related)

uberhalit:
This breaks my Fire Emblem Fates and Awakening ROMs in BDE which both were working fine before 238.

jroweboy:
I've gotten several other reports from people that this caused issues so I'm going to remove this from bleeding edge for now. I couldn't cause the error locally, or I would have looked deeper into this problem. (If I had to take a guess, I wonder if some of the errors were previously non fatal for some games, and this PR makes them fatal errors.)

wwylele:
Tested some of my games and found the following issues:

 - Kirby Planet Robobot asking for a non-critical system archives: `failed to get a handle for archive archive_id=0x2345678A archive_path=[Binary: 020601009b00040000ffffff04000000]` and the dialog pops up. According to 3dbrew, this contains "Non-Nintendo TLS Root-CA Certificates". This file is currently not dumped by our archive dumper 3dsutil (can be easily added, though), and missing of this file doesn't prevent the game from running.

 - Some games may call `ThrowFatelError` on non-critical situation. For example, Pokemon asks for config block 0x00090000, which is just added in citra recently. If the game uses citra with save data inherited from an old version, the block doesn't exist and the game will `ThrowFatelError`, which doesn't prevent the game from continuing running before this PR. I am not sure about this... `ThrowFatelError` is still something that is not supposed happen in "normal" cases, but it can happen more frequently than we think.

TheKoopaKingdom:
@uberhalit What exactly happened when you tried running it?

(Will look into the false positive issues tomorrow.)

src/citra_qt/main.cpp
```diff
@@ -663,8 +663,6 @@ void GMainWindow::UpdateStatusBar() {
 }
 
 void GMainWindow::OnCoreError(Core::System::ResultStatus result) {
-    if (emu_thread != nullptr)
-        ShutdownGame();
```
JayFoxRox: You'll probably want to reflect this in the error messages. Otherwise we'll get support requests: "It said fatal error but then worked, is it safe to save?!".

Also I'm not sure if this might open a whole new can of worms: say an error in a tight loop - would that generate thousands of error messages?

---

I think a good way to solve this would be to utilize the new status bar for this and append warning messages if the game tried to do something like opening the font. This would signal that the emulation is tainted.
You could still show the error messages, but only the first time per run of a game?

(only my idea though and I feel there'll be different opinions about this)

TheKoopaKingdom: >Also I'm not sure if this might open a whole new can of worms: say an error in a tight loop - would that generate thousands of error messages?

I had Core::RunLoop() reset the status every time.

>I think a good way to solve this would be to utilize the new status bar for this and append warning messages if the game tried to do something like opening the font. This would signal that the emulation is tainted.

Sounds like a good idea.

uberhalit:
@TheKoopaKingdom I get: 

> Citra was unable to locate the 3DS system archive.

wwylele:
What's the status of this PR? @TheKoopaKingdom are you still working on this?

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -129,14 +129,61 @@ static void OpenFileDirectly(Service::Interface* self) {
...
+                if (path == ng_word_list) {
+                    LOG_ERROR(Service_FS,
+                              "Failed to get a handle for system data archive: NG bad word list. "
+                              "Archive ID=0x%08X Archive Path=%s",
+                              static_cast<u32>(archive_id), archive_path.DebugStr().c_str());
+                    Core::System::GetInstance().SetStatus(
+                        Core::System::ResultStatus::ErrorSystemFiles, "NG bad word list");
+                    return;
+                }
+            }
```
wwylele: I feel like these checks should be placed [in `Archive_NCCH` handler, to which all 0x2345678A archives are dispatched](https://github.com/citra-emu/citra/blob/master/src/core/file_sys/archive_ncch.cpp#L40). Not sure if it is good to intrude so much, though.

TheKoopaKingdom: I think I'll go with that, because that way, the error reporting will be covered by both OpenFileDirectly() and OpenArchive() (As in, the service function in FS_User), as opposed to just the former.

wwylele: @TheKoopaKingdom you can still check the TID there. `data[0]` and `data[1]` are the TID

TheKoopaKingdom: Added in [4c180e8](https://github.com/citra-emu/citra/pull/2611/commits/4c180e88904156620fe3545d9283999837c7c01b). Currently, situations in which there's a white-listed system file missing, there will be 2 messages logged: 
```
[   3.544855] Service.FS <Error> core/file_sys/archive_ncch.cpp:Open:68: Failed to get a handle for system data archive: NG bad word list. 
[   3.544870] Service.FS <Error> core/hle/service/fs/fs_user.cpp:OpenFileDirectly:136: Failed to get a handle for archive archive_id=0x2345678A archive_path=[Binary: 02030100db00040000ffffff01000000]
```
Not sure if this is too repetitive? I'm not really sure how to go about only having one message be printed, because of `Open()` returning a `ResultVal`, which `OpenFileDirectly()` can't really use to tell if the TID was white-listed or not.

wwylele: I think two messages in log are fine, since log is meant to be verbose in order to report all status, and it is unnecessary to mess up with logic around log IMO.

TheKoopaKingdom:
Currently, aside from some commit squishing, this PR is about ready to be LGTM'd. The core functionality described in the PR description is still intact, but now has a few improvements:
 - The correct error code is assigned to core internally (Verified against HW.).
 - An option to quit back to the game list is provided within the message box.
 - The last error, if ignored by the user, is shown in the status bar.
 - Added a whitelist for the system title IDs that are dumped by 3dsutils to prevent dialogs being opened when they're unnecessary.

src/core/core.cpp
```diff
@@ -73,22 +77,32 @@ System::ResultStatus System::Load(EmuWindow* emu_window, const std::string& file
...
     }
+    std::pair<boost::optional<u32>, Loader::ResultStatus> system_mode =
+        app_loader->LoadKernelSystemMode();
 
-    boost::optional<u32> system_mode{app_loader->LoadKernelSystemMode()};
-    if (!system_mode) {
-        LOG_CRITICAL(Core, "Failed to determine system mode!");
-        return ResultStatus::ErrorSystemMode;
+    if (system_mode.second != Loader::ResultStatus::Success) {
+        LOG_CRITICAL(Core, "Failed to determine system mode (Error %i)!", system_mode.second);
```
wwylele: `static_cast<int>(system_mode.second)`

src/core/core.cpp
```diff
@@ -3,6 +3,9 @@
 // Refer to the license.txt file included.
 
 #include <memory>
+#include <utility>
+
```
wwylele: no blank line between includes

src/core/core.cpp
```diff
@@ -73,22 +77,32 @@ System::ResultStatus System::Load(EmuWindow* emu_window, const std::string& file
...
     if (init_result != ResultStatus::Success) {
         LOG_CRITICAL(Core, "Failed to initialize system (Error %i)!", init_result);
         System::Shutdown();
         return init_result;
     }
 
-    const Loader::ResultStatus load_result{app_loader->Load()};
-    if (Loader::ResultStatus::Success != load_result) {
+    Loader::ResultStatus load_result = app_loader->Load();
+    if (load_result != Loader::ResultStatus::Success) {
```
wwylele: What's the point of this change? I understand that `Loader::ResultStatus::Success != load_result` looks bad but... IMO the diff should be kept as clean as possible

src/core/core.h
```diff
@@ -6,6 +6,9 @@
 
 #include <memory>
 #include <string>
+
```
wwylele: ditto, no blank line between includes

src/core/core.h
```diff
@@ -118,6 +140,9 @@ class System {
     bool reschedule_pending{};
 
     static System s_instance;
+
+    ResultStatus status;
+    boost::optional<std::string> status_details;
```
wwylele: why not just store this as plain `std::string` and represent "no details" as empty string, as `SetStatus` do?

src/core/hle/service/fs/archive.cpp
```diff
@@ -266,11 +266,9 @@ ResultVal<ArchiveHandle> OpenArchive(ArchiveIdCode id_code, FileSys::Path& archi
...
 
     auto itr = id_code_map.find(id_code);
-    if (itr == id_code_map.end()) {
-        // TODO: Verify error against hardware
-        return ResultCode(ErrorDescription::NotFound, ErrorModule::FS, ErrorSummary::NotFound,
-                          ErrorLevel::Permanent);
-    }
+    if (itr == id_code_map.end())
+        return ResultCode(ErrorDescription::FS_NotFound, ErrorModule::FS, ErrorSummary::NotFound,
+                          ErrorLevel::Status);
```
wwylele: clean this diff

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -128,11 +128,12 @@ static void OpenFileDirectly(Service::Interface* self) {
...
     if (archive_handle.Failed()) {
-        LOG_ERROR(Service_FS,
-                  "failed to get a handle for archive archive_id=0x%08X archive_path=%s",
-                  static_cast<u32>(archive_id), archive_path.DebugStr().c_str());
         cmd_buff[1] = archive_handle.Code().raw;
         cmd_buff[3] = 0;
+
+        LOG_ERROR(Service_FS,
+                  "Failed to get a handle for archive archive_id=0x%08X archive_path=%s",
+                  static_cast<u32>(archive_id), archive_path.DebugStr().c_str());
```
wwylele: clean this diff

src/core/loader/loader.h
```diff
@@ -8,8 +8,11 @@
 #include <initializer_list>
 #include <memory>
 #include <string>
+#include <utility>
 #include <vector>
+
```
wwylele: no blank lines

src/core/loader/loader.h
```diff
@@ -100,11 +103,11 @@ class AppLoader : NonCopyable {
      * Loads the system mode that this application needs.
      * This function defaults to 2 (96MB allocated to the application) if it can't read the
      * information.
-     * @returns Optional with the kernel system mode
+     * @return A pair with the system mode (If found) and the result.
```
wwylele: `@returns a pair of Optional with the kernel system mode and ResultStatus`

src/core/loader/ncch.h
```diff
@@ -179,9 +179,9 @@ class AppLoader_NCCH final : public AppLoader {
 
     /**
      * Loads the Exheader and returns the system mode for this application.
-     * @return Optional with the kernel system mode
+     * @return A pair with the system mode (If found) and the result.
```
wwylele: ditto

chris062689:
Has there been any comments on this since it hit Bleeding Edge? I'm kind of surprised we haven't seen any support topics asking for what the new alert messages mean.

bunnei:
I think this broke BE, untagging.

@TheKoopaKingdom can you address the above comments so that we can get this wrapped up?

src/core/core.h
TheKoopaKingdom: Aren't STD includes separated from library includes?

wwylele:
Please resolve conflicts and squash commits

src/citra_qt/main.cpp
TheKoopaKingdom: (This has been addressed.)

src/core/hle/service/fs/archive.cpp
wwylele: This can be replaced with the constant `ERROR_NOT_FOUND` defined in `file_sys/errors.h

src/core/file_sys/archive_ncch.cpp
```diff
@@ -33,11 +34,45 @@ ArchiveFactory_NCCH::ArchiveFactory_NCCH(const std::string& nand_directory)
...
+            }
+        } else if (high == system_data_archive) {
+            if (low == ng_word_list) {
+                LOG_ERROR(Service_FS,
+                          "Failed to get a handle for system data archive: NG bad word list.");
+                Core::System::GetInstance().SetStatus(Core::System::ResultStatus::ErrorSystemFiles,
+                                                      "NG bad word list");
+            }
+        }
+        return ResultCode(ErrorDescription::FS_NotFound, ErrorModule::FS, ErrorSummary::NotFound,
```
wwylele: This can be replaced with the constant ERROR_NOT_FOUND defined in `file_sys/errors.h

src/citra_qt/main.cpp
```diff
@@ -328,27 +335,41 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
-                   "Citra.<br/><br/>"
-                   "For more information on dumping and decrypting games, please see: <a "
-                   "href='https://citra-emu.org/wiki/Dumping-Game-Cartridges'>https://"
-                   "citra-emu.org/wiki/Dumping-Game-Cartridges</a>"));
-            popup_error.setIcon(QMessageBox::Critical);
-            popup_error.exec();
+                   "Citra. A real 3DS is required.<br/><br/>"
+                   "For more information on dumping and decrypting games, please see the following "
+                   "wiki pages: <ul>"
+                   "<li><a href='https://citra-emu.org/wiki/Dumping-Game-Cartridges/'>Dumping Game "
```
jroweboy: This link is a 404

src/citra_qt/main.cpp
```diff
@@ -328,27 +335,41 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
+                   "Cartridges</a></li>"
+                   "<li><a href='https://citra-emu.org/wiki/Dumping-Installed-Titles/'>Dumping "
```
jroweboy: This link is a 404

src/citra_qt/main.cpp
```diff
@@ -623,11 +649,67 @@ void GMainWindow::UpdateStatusBar() {
...
 }
 
+void GMainWindow::OnCoreError(Core::System::ResultStatus result, std::string details) {
+    QMessageBox::StandardButton answer;
+    QString status_message;
+    const QString common_message =
+        tr("The game you are trying to load requires additional files from your 3DS to be dumped "
+           "before playing.<br/><br/>For more information on dumping these files, please see the "
+           "following wiki page: <a "
+           "href='https://citra-emu.org/wiki/"
```
jroweboy: This link is a 404

chris062689:
This breakage was due to a recent update to the website that forces all wiki article names to lowercase.

chris062689:
While we're editing the dialog box, can we modify the text at the bottom to say, in addition to "Would you like to quit back to the game list?" to additionally say "If you continue, your may experience bugs or corrupted save data.", basically an additional note that if they continue they're going to have a bad time...

src/citra_qt/main.cpp
TheKoopaKingdom: Fixed in [dbbd3f6f0c7b62c107f9afc737e14279ab2241ea](https://github.com/citra-emu/citra/pull/2611/commits/dbbd3f6f0c7b62c107f9afc737e14279ab2241ea#diff-5e8bf22f5c3b3579c402308dd96b7563R660).

TheKoopaKingdom:
> While we're editing the dialog box, can we modify the text at the bottom to say, in addition to "Would you like to quit back to the game list?" to additionally say "If you continue, your may experience bugs or corrupted save data.", basically an additional note that if they continue they're going to have a bad time...

@chris062689 Sure. Added in [dbbd3f6f0c7b62c107f9afc737e14279ab2241ea](https://github.com/citra-emu/citra/pull/2611/commits/dbbd3f6f0c7b62c107f9afc737e14279ab2241ea#diff-5e8bf22f5c3b3579c402308dd96b7563R662).

I said this on Discord, but it's worth mentioning that when a user does ignore the message box, a message is shown in the status bar, added in [63c870196ca8f81ec8acc823522e12a2ef0e5855](https://github.com/citra-emu/citra/pull/2611/commits/63c870196ca8f81ec8acc823522e12a2ef0e5855) and implemented in [32f5d4da9485ba68a23e52fcc6bef352c4e14a25](https://github.com/citra-emu/citra/pull/2611/commits/32f5d4da9485ba68a23e52fcc6bef352c4e14a25#diff-5e8bf22f5c3b3579c402308dd96b7563R704).

chris062689:
LGTM

src/core/file_sys/archive_ncch.cpp
```diff
@@ -33,11 +35,44 @@ ArchiveFactory_NCCH::ArchiveFactory_NCCH(const std::string& nand_directory)
...
+                  high, low);
+
+        if (high == shared_data_archive) {
+            if (low == mii_data) {
+                LOG_ERROR(Service_FS, "Failed to get a handle for shared data archive: Mii data. ");
+                Core::System::GetInstance().SetStatus(Core::System::ResultStatus::ErrorSystemFiles,
+                                                      "Mii data");
+            } else if (low == region_manifest) {
+                LOG_ERROR(Service_FS,
+                          "Failed to get a handle for shared data archive: region manifes");
```
bunnei: nit: manifest?

src/core/file_sys/archive_ncch.cpp
```diff
@@ -33,11 +35,44 @@ ArchiveFactory_NCCH::ArchiveFactory_NCCH(const std::string& nand_directory)
...
-    std::string file_path = GetNCCHPath(mount_point, data[1], data[0]);
+    u32 high = data[1];
+    u32 low = data[0];
+    std::string file_path = GetNCCHPath(mount_point, high, low);
     auto file = std::make_shared<FileUtil::IOFile>(file_path, "rb");
 
     if (!file->IsOpen()) {
-        return ResultCode(-1); // TODO(Subv): Find the right error code
+        // High Title ID of the archive: The category (https://3dbrew.org/wiki/Title_list).
+        const u32 shared_data_archive = 0x0004009B;
```
bunnei: These can all be constexpr

src/core/loader/ncch.h
```diff
@@ -179,9 +179,9 @@ class AppLoader_NCCH final : public AppLoader {
 
     /**
      * Loads the Exheader and returns the system mode for this application.
-     * @return Optional with the kernel system mode
+     * @returns a pair of Optional with the kernel system mode and ResultStatus
```
bunnei: This comment reads weird now... (it kind of did before)

src/core/core.cpp
```diff
@@ -101,6 +113,8 @@ System::ResultStatus System::Load(EmuWindow* emu_window, const std::string& file
             return ResultStatus::ErrorLoader;
         }
     }
+    // this->status will be used for errors while actually running the game
+    status = ResultStatus::Success;
     return ResultStatus::Success;
```
bunnei: return status?

src/core/loader/loader.h
```diff
@@ -100,11 +101,11 @@ class AppLoader : NonCopyable {
      * Loads the system mode that this application needs.
      * This function defaults to 2 (96MB allocated to the application) if it can't read the
      * information.
-     * @returns Optional with the kernel system mode
+     * @returns a pair of Optional with the kernel system mode and ResultStatus.
```
bunnei: Ditto with weird comment

src/core/core.h
```diff
@@ -105,6 +108,19 @@ class System {
...
+    ResultStatus GetStatus() {
+        return status;
+    }
+
+    void SetStatus(ResultStatus new_status, std::string details = std::string()) {
+        status = new_status;
+        status_details = details;
+    }
+
+    std::string GetStatusDetails() {
```
bunnei: This can return `const std::string&`. Should also be a `const` method. E.g.
`const std::string GetStatusDetails)() const`

src/core/core.h
```diff
@@ -105,6 +108,19 @@ class System {
     PerfStats perf_stats;
     FrameLimiter frame_limiter;
 
+    ResultStatus GetStatus() {
+        return status;
+    }
+
+    void SetStatus(ResultStatus new_status, std::string details = std::string()) {
```
bunnei: The std::string arg here is not particularly performant. If we really need this, consider passing this by const reference (e.g. `const std::string&` and making it required. As an alternative, this can also just be a C-string if you need it to be optional, e.g. `const char* details = nullptr`, then `if (details) status_details = details` - prevents unnecessary construction/copies of std::string

src/core/core.h
```diff
@@ -105,6 +108,19 @@ class System {
     PerfStats perf_stats;
     FrameLimiter frame_limiter;
 
+    ResultStatus GetStatus() {
```
bunnei: Should be a `const` method

src/core/core.h
```diff
@@ -130,6 +146,9 @@ class System {
     std::unique_ptr<Core::TelemetrySession> telemetry_session;
 
     static System s_instance;
+
+    ResultStatus status;
```
bunnei: Initialize this to something? (otherwise it will be garbage and can be hard to debug issues)

src/core/core.cpp
```diff
@@ -26,6 +26,7 @@ namespace Core {
 /*static*/ System System::s_instance;
 
 System::ResultStatus System::RunLoop(int tight_loop) {
+    this->status = ResultStatus::Success;
```
bunnei: No need for `this->`

src/citra_qt/main.cpp
```diff
@@ -623,11 +649,69 @@ void GMainWindow::UpdateStatusBar() {
...
+               "<a href='https://community.citra-emu.org/t/how-to-upload-the-log-file/296'>How to "
+               "Upload the Log File</a>.<br/><br/>Would you like to quit back to the game list? "
+               "Continuing emulation may result in crashes, corrupted save data, or other bugs."),
+            QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
+        status_message = "Fatal Error encountered";
+        break;
+    }
+
+    if (answer == QMessageBox::Yes) {
+        if (emu_thread != nullptr)
```
bunnei: Can be just `if (emu_thread)`

src/core/loader/ncch.cpp
```diff
@@ -121,12 +121,16 @@ FileType AppLoader_NCCH::IdentifyType(FileUtil::IOFile& file) {
     return FileType::Error;
 }
 
-boost::optional<u32> AppLoader_NCCH::LoadKernelSystemMode() {
+std::pair<boost::optional<u32>, ResultStatus> AppLoader_NCCH::LoadKernelSystemMode() {
     if (!is_loaded) {
-        if (LoadExeFS() != ResultStatus::Success)
-            return boost::none;
+        ResultStatus res = LoadExeFS();
```
bunnei: No need for this var, just call the load function in the if-statement?

src/core/core.cpp
```diff
@@ -73,22 +74,33 @@ System::ResultStatus System::Load(EmuWindow* emu_window, const std::string& file
...
+    ResultStatus init_result{Init(emu_window, system_mode.first.get())};
     if (init_result != ResultStatus::Success) {
         LOG_CRITICAL(Core, "Failed to initialize system (Error %i)!", init_result);
         System::Shutdown();
         return init_result;
     }
 
-    const Loader::ResultStatus load_result{app_loader->Load()};
-    if (Loader::ResultStatus::Success != load_result) {
+    Loader::ResultStatus load_result = app_loader->Load();
```
bunnei: No need for this var, just call the load function in the if-statement?

src/citra_qt/bootmanager.cpp
```diff
@@ -38,7 +38,10 @@ void EmuThread::run() {
             if (!was_active)
                 emit DebugModeLeft();
 
-            Core::System::GetInstance().RunLoop();
+            Core::System::ResultStatus result = Core::System::GetInstance().RunLoop();
```
bunnei: No need for this var, just call the function in the if-statement?

src/citra_qt/main.cpp
```diff
@@ -623,11 +649,69 @@ void GMainWindow::UpdateStatusBar() {
...
+           "following wiki page: <a "
+           "href='https://citra-emu.org/wiki/"
+           "dumping-system-archives-and-the-shared-fonts-from-a-3ds-console/'>Dumping System "
+           "Archives and the Shared Fonts from a 3DS Console</a>.<br/><br/>Would you like to quit "
+           "back to the game list? Continuing emulation may result in crashes, corrupted save "
+           "data, or other bugs.");
+    switch (result) {
+    case Core::System::ResultStatus::ErrorSystemFiles: {
+        QString message = "Citra was unable to locate a 3DS system archive";
+        if (details != std::string())
```
bunnei: nit: Prefer `{` and `}` even on one-line if-statements. While pedantic, it can prevent some nasty white space bugs

bunnei: (applies here and elsewhere in this change)

src/core/loader/ncch.cpp
TheKoopaKingdom: `res` is used when returning the status on [line 128](https://github.com/citra-emu/citra/pull/2611/files#diff-97ddda51413fc4c2c643428d97920640R128).

src/citra_qt/bootmanager.cpp
TheKoopaKingdom: `result` is used when passing the status to `ErrorThrown` on [line 43](https://github.com/citra-emu/citra/pull/2611/files/dbbd3f6f0c7b62c107f9afc737e14279ab2241ea#diff-38ca3f21a73cc5c54e8fe31685b11829R43).

Alternatively, subbing something like `Core::System::GetInstance().GetStatus()` in for that parameter would be possible.

PR #2612, opened 2017-03-05
Fix small UI bugs

jmerdich:
Three small UI-related things fixed here. 

 - Sort order is now preserved after auto-refresh
 - Changing settings now triggers a list refresh (for recursive checked/unchecked)
 - The UI dialog now applies its settings exactly once.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2612)

### DIFF
--- a/src/citra_qt/configure_audio.cpp
+++ b/src/citra_qt/configure_audio.cpp
@@ -53,7 +53,7 @@ void ConfigureAudio::setConfiguration() {
     ui->audio_device_combo_box->setCurrentIndex(new_device_index);
 }
 
-void ConfigureAudio::applyConfiguration() {
+void ConfigureAudio::applyConfiguration(bool commit_settings_file) {
     Settings::values.sink_id =
         ui->output_sink_combo_box->itemText(ui->output_sink_combo_box->currentIndex())
             .toStdString();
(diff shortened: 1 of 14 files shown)


wwylele:
> Changing settings now triggers a list refresh (for recursive checked/unchecked)

Does it refresh for any settings change? I am worried about this because I have a relatively large game directory and it can take a second to load the game list. If every change will cause a reload it can be annoying.

src/citra_qt/config.cpp
```diff
@@ -263,6 +263,7 @@ void Config::SaveValues() {
 void Config::Reload() {
     ReadValues();
     Settings::Apply();
+    UISettings::values.apply();
```
vonchenplus: UISetings namespace add apply method.

bunnei:
@jmerdich can you update this PR?

jmerdich:
Latest patchset does away with the qt signals/slots (which made this commit relatively intrusive) and only causes a rescan if the recursion setting is actually changed.

JayFoxRox:
Please resolve the conflicts @jmerdich

bunnei:
@jmerdich do you have any plans to finish this PR? Please resolve merge conflicts if so - It would be great to get these fixes merged! Thanks.

bunnei:
Closing since it looks like #2705 is more likely to get completed

PR #2618, opened 2017-03-08
Reduce host file name and path logging

wwylele:
Logging file name and path have several problems:
 - Too verbose and can distract one from finding the actual problem
 - Can contain sensitive personal information (e.g. username) and user can upload the log with it without knowing it.
 - The way a user to name their game file can sometimes lead to misunderstanding about piracy or something.

The aim of this PR is to lower the level of logs that contains file name and path and that is often logged. (generally all <=Warning log are now <=Debug).


jroweboy:
One thing that would be nice to know is if citra is running in portable mode or if its using the AppData\Roaming location on windows. That information was usually logged as part of the old logging, so it would be nice to print a log if its running as portable, and probably just don't log anything if citra is just using the standard location.

(If you feel its out of scope even though its only a one line change, thats fine. Its just something thats useful to know from the logs that is no longer going to be there after this change.)

wwylele:
Took @jroweboy 's advise. Now it logs when it detects that the local user directory is used

PR #2619, opened 2017-03-08
added support for loading aeskeydb.bin to the aes code

jfmherokiller:
This is an ammendment to pull request #2569 to allow it to both load from the text file and load from aeskeydb.bin. Checks have been added

Please let me know how to organize it better if it needs it


src/core/hw/aes/key.cpp
```diff
@@ -137,6 +137,49 @@ void LoadPresetKeys() {
         }
     }
 }
+void LoadKeysFromDB() {
+    // aeskeydb.bin struct taken from https://git.io/vyzHF
+    typedef struct {
```
bunnei: No need for C-style struct

src/core/hw/aes/key.cpp
```diff
@@ -137,6 +137,49 @@ void LoadPresetKeys() {
...
+        u8 slot{};        // keyslot, 0x00...0x3F
+        char type{};      // type 'X' / 'Y' / 'N' for normalKey
+        char id[10]{};    // key ID for special keys, all zero for standard keys
+        u8 reserved[2]{}; // reserved space
+        u8 isDevkitKey{}; // 0 for retail units / 1 for DevKit units
```
bunnei: Incorrect style. Member variables should be `lowercase_with_underscores`. Please see our [Contributing Guide](https://github.com/citra-emu/citra/blob/master/CONTRIBUTING.md) for more information.

src/core/hw/aes/key.cpp
```diff
@@ -137,6 +137,49 @@ void LoadPresetKeys() {
...
+    } __attribute__((packed)) AesKeyInfo;
+
+    const std::string filepath = FileUtil::GetUserPath(D_SYSDATA_IDX) + AES_KEYS_DB;
+    FileUtil::CreateFullPath(filepath); // Create path if not already created
+    std::ifstream file;
+    OpenFStream(file, filepath, std::ios_base::in);
+    if (!file) {
+        return;
+    }
+    std::vector<AesKeyInfo> keys(0);
```
bunnei: This is unused?

jfmherokiller:
learned the hard way that visual studio doesnt like __attribute__((packed))

src/core/hw/aes/key.cpp
```diff
@@ -137,6 +137,49 @@ void LoadPresetKeys() {
...
+            break;
+        default:
+            LOG_ERROR(HW_AES, "Invalid key type %c", keyinfo.type);
+            break;
+        }
+    }
+}
+void LoadPresetKeys() {
+    LoadKeysFromText();
+    LoadKeysFromDB();
```
bunnei: Do we want to only load one of these if the other does not succeed? What if both exist? Right now, the text keys will be overwritten. That might be fine, but if so, we should at least log it, and/or notify the user some how.

src/core/hw/aes/key.cpp
```diff
@@ -137,6 +137,50 @@ void LoadPresetKeys() {
         }
     }
 }
+void LoadKeysFromDB() {
+    // aeskeydb.bin struct taken from https://git.io/vyzHF
```
bunnei: Could use a little more context IMO, that this is the format from Decrypt9, how one might generate it, etc.

src/core/hw/aes/key.cpp
```diff
@@ -137,6 +137,50 @@ void LoadPresetKeys() {
...
+#pragma pack(push, 1)
```
bunnei: Do we really need these?

src/core/hw/aes/key.cpp
```diff
@@ -137,6 +137,50 @@ void LoadPresetKeys() {
...
+#pragma pack(pop)
+    const std::string filepath = FileUtil::GetUserPath(D_SYSDATA_IDX) + AES_KEYS_DB;
+    FileUtil::CreateFullPath(filepath); // Create path if not already created
+    std::ifstream file;
+    OpenFStream(file, filepath, std::ios_base::in);
+    if (!file) {
+        return;
+    }
+    std::vector<AesKeyInfo> keys(0);
+    AesKeyInfo keyinfo;
```
bunnei: key_info

jfmherokiller:
@bunnei thanks for the changes ill begin working on a few of them now

src/core/hw/aes/key.cpp
jfmherokiller: yea i forgot to remove that since before i had implemented a forloop

src/core/hw/aes/key.cpp
jfmherokiller: should i add a secondary link or copy info from the decrypt9 wiki?
eg this info 

* __Build Key Database__: This is used to build the `aeskeydb.bin` file from all available `slot0x??key?.bin` files. It will also process files that are not used by Decrypt9. With the `aeskeydb.bin`available, Decrypt9 can load keys from it and doesn't need the `slot0x??key?.bin` files anymore.

src/core/hw/aes/key.cpp
jfmherokiller: @bunnei i do not know if it would mess up being able to read the file format or not so thats why i added them

jfmherokiller:
btw just so you know i do plan to squash all the commits before this gets accepted

src/core/hw/aes/key.cpp
jfmherokiller: removed in the latest commit

src/core/hw/aes/key.cpp
```diff
@@ -121,22 +143,45 @@ void LoadPresetKeys() {
...
+ * It requires bin files containing the key data following the format slot0x??key?.bin.
+ * The wiki describes that this database can be encrypted or decrypted however as of right now only
+ * the decrypted format is supported.
+ */
+void LoadKeysFromDB() {
+    // aeskeydb.bin struct taken from https://git.io/vyzHF
+    typedef struct AesKeyInfo {
+        u8 slot;         // keyslot, 0x00...0x3F
+        char type;       // type 'X' / 'Y' / 'N' for normalKey
+        char id[10];     // key ID for special keys, all zero for standard keys
```
wwylele: `std::array<char, 10> id;`

jfmherokiller: @wwylele thanks gonna put that in now

src/core/hw/aes/key.cpp
jfmherokiller: @bunnei i was going to put an info log that would output the key to the console however i didnt want to mess with trying to convert an array to a hex string

jfmherokiller:
going to squash now before this gets out of control

src/core/hw/aes/key.cpp
jfmherokiller: Added a doc style comment containing an explanation of the format

jfmherokiller:
Should i add a boolean argument to the SetKeySlot function to allow them to be overridden if necessary? I think the 3ds overwrites them without a second thought

mailwl:
this data not allowed to use, why wylele allow it?

jfmherokiller:
@mailwl why cant we use the data? its available in the 3ds.guide

mailwl:
via torrent

jfmherokiller:
@mailwl well im not going to include the file itself im just adding code to allow citra to parse it

jfmherokiller:
closed because the legality of using a resource available on the 3ds.guide is not confirmed and i dont wish to put this project in hot water

dogtopus:
I don't really understand why this PR got closed. IMO anyone with some knowledge about the format of aeskeydb.bin (which is freely available on the Internet) could just write a converter that converts the file on 3ds.guide to aeskeys.txt/whatever in just 5 minutes. 
It's not wrong to just support a file format, otherwise all the text editors should be banned because one can store illegal content in a text file.

jfmherokiller:
i closed this pull request because i didn't want to put citra into legal issues but if there is enough want for it ill reopen it

PR #2620, opened 2017-03-09
Refined thread launch on syscore error messages

FernandoS27:
Just refined the error for each case, this way we can know better under which circumstances syscore emulation is really needed. It can also help isolate certain issues from each other.

### DIFF
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -556,11 +556,21 @@ static ResultCode CreateThread(Kernel::Handle* out_handle, s32 priority, u32 ent
         break;
     }
 
-    if (processor_id == THREADPROCESSORID_1 || processor_id == THREADPROCESSORID_ALL ||
-        (processor_id == THREADPROCESSORID_DEFAULT &&
-         Kernel::g_current_process->ideal_processor == THREADPROCESSORID_1)) {
-        LOG_WARNING(Kernel_SVC,
-                    "Newly created thread is allowed to be run in the SysCore, unimplemented.");
+    if (processor_id == THREADPROCESSORID_ALL) {
+        LOG_INFO(Kernel_SVC,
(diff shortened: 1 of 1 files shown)


src/core/hle/svc.cpp
```diff
@@ -556,11 +556,20 @@ static ResultCode CreateThread(Kernel::Handle* out_handle, s32 priority, u32 ent
...
     }
 
-    if (processor_id == THREADPROCESSORID_1 || processor_id == THREADPROCESSORID_ALL ||
-        (processor_id == THREADPROCESSORID_DEFAULT &&
+    if (processor_id == THREADPROCESSORID_ALL) {
+        LOG_INFO(Kernel_SVC,
+                    "Newly created thread is allowed to be run in any Core, unimplemented.");
+    }
+
+    if ((processor_id == THREADPROCESSORID_DEFAULT &&
```
vonchenplus: double brackets?

src/core/hle/svc.cpp
```diff
@@ -556,11 +556,20 @@ static ResultCode CreateThread(Kernel::Handle* out_handle, s32 priority, u32 ent
...
          Kernel::g_current_process->ideal_processor == THREADPROCESSORID_1)) {
         LOG_WARNING(Kernel_SVC,
-                    "Newly created thread is allowed to be run in the SysCore, unimplemented.");
+                    "Newly created thread is allowed to be run in the SysCore (Core1), unimplemented.");
+    }
+
+    if (processor_id == THREADPROCESSORID_1) {
+        LOG_ERROR(Kernel_SVC,
+                    "Newly created thread must be allowed to be run in the SysCore, unimplemented.");
```
wwylele: Should also be `SysCore (Core1)` as above. Also, "must be allowed" is kind of hard to read for me...

PR #2622, opened 2017-03-11
error conversion fixes for soc_u

jfmherokiller:
this commit fixes the errors in the soc_u implementation involving converting and sending the correct error code to the 3ds software
the current implementation had a fault where it would set cmd_buffer[1] to the value of the converted ERRNO message.
This implementation is incorrect because cmd_buffer[1] should be zero if the functions return an error message.
The converted error message should be contained in cmd_buffer[2] instead of cmd_buffer[1].
A working example of this pull request can be observed if you run the 3ds sockets example on linux/osx with citra as root and then access the ip address displayed in the window.

As of rn the windows portion of this code is broken where it returns an invalid address family

### DIFF
--- a/src/core/hle/service/soc_u.cpp
+++ b/src/core/hle/service/soc_u.cpp
@@ -362,18 +362,18 @@ static void Socket(Interface* self) {
         return;
     }
 
-    u32 socket_handle = static_cast<u32>(::socket(domain, type, protocol));
+    u32 ret = static_cast<u32>(::socket(domain, type, protocol));
 
-    if ((s32)socket_handle != SOCKET_ERROR_VALUE)
-        open_sockets[socket_handle] = {socket_handle, true};
+    if ((s32)ret != SOCKET_ERROR_VALUE)
+        open_sockets[ret] = {ret, true};
 
     int result = 0;
-    if ((s32)socket_handle == SOCKET_ERROR_VALUE)
-        result = TranslateError(GET_ERRNO);
+    if ((s32)ret == SOCKET_ERROR_VALUE)
+        ret = TranslateError(GET_ERRNO);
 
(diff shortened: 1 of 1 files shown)


jfmherokiller:
this does not fully make it work but it does get it farther along

lioncash:
Your commit message should also include what it is that you've fixed in particular.

The current equivalent of "fixed some stuff" is quite vague.

jfmherokiller:
woops let me fix that

JayFoxRox:
There are a lot of other instances / functions in this file which must be fixed.
I feel only changing 3 of these issues will actually make the problem worse because it will be very hard to find the broken ones - you are turning a pile of crap into a haystack with a needle of shit.

jfmherokiller:
@JayFoxRox which other ones do i need to fix?

src/core/hle/service/soc_u.cpp
```diff
@@ -369,7 +369,7 @@ static void Socket(Interface* self) {
 
     int result = 0;
     if ((s32)socket_handle == SOCKET_ERROR_VALUE)
-        result = TranslateError(GET_ERRNO);
+        socket_handle = TranslateError(GET_ERRNO);
```
JayFoxRox: I think in this case it makes more sense to add a temporary variable or rename the `socket_handle` variable.

src/core/hle/service/soc_u.cpp
```diff
@@ -397,7 +397,7 @@ static void Bind(Interface* self) {
 
     int result = 0;
     if (res != 0)
-        result = TranslateError(GET_ERRNO);
+        res = TranslateError(GET_ERRNO);
```
JayFoxRox: Should probably be renamed to `ret`

PR #2623, opened 2017-03-11
fix compiling problem with vs2017

yhnmj6666:
VS 2017 will report something like " '__cpuid': identifier not found " if <intrin.h> isn't included.


wwylele:
Is this the only error we get on VS 2017?

jfmherokiller:
I dont believe so but rn i think visual studio 2017 is still in the release candidate stage

legoj15:
As of March 9th, VS 2017 is released. 

Or that's what this says at least https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes

jfmherokiller:
well can somone test that you can select the visual studio 2015 toolset from the cmake gui if you install it via vs2017?

bunnei:
This change was just merged with #2626, thanks anyway!

PR #2625, opened 2017-03-12
cfg: correctly implement GenHashConsoleUnique

wwylele:
see [3dbrew](https://www.3dbrew.org/wiki/Cfg:GenHashConsoleUnique) for documentation. Results are verified against real 3DS.

### DIFF
--- a/src/core/hle/service/cfg/cfg.cpp
+++ b/src/core/hle/service/cfg/cfg.cpp
@@ -3,6 +3,8 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
+#include <array>
+#include <cryptopp/sha.h>
 #include "common/file_util.h"
 #include "common/logging/log.h"
 #include "common/string_util.h"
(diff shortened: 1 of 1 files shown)


mailwl:
maybe rename to GetTransferableId?

wwylele:
@mailwl I see both names on 3dbrew, and can't tell where they come and which is better. I want to make my PR as clean as possible, so until I find a good reason to change it, I won't touch it. You can submit a PR later if you have a good reason.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -176,14 +177,29 @@ void SecureInfoGetRegion(Service::Interface* self) {
...
-
-    cmd_buff[1] = RESULT_SUCCESS.raw;
-    cmd_buff[2] = 0x33646D6F ^ (app_id_salt & 0xFFFFF); // 3dmoo hash
-    cmd_buff[3] = 0x6F534841 ^ (app_id_salt & 0xFFFFF);
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x03, 1, 0);
+    u32 app_id_salt = rp.Pop<u32>() & 0x000FFFFF;
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(3, 0);
+
+    u8 buffer[12];
```
lioncash: `std::array`

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -176,14 +177,29 @@ void SecureInfoGetRegion(Service::Interface* self) {
...
+    ResultCode result = GetConfigInfoBlock(ConsoleUniqueID2BlockID, 8, 2, buffer);
+    rb.Push(result);
+    if (result.IsSuccess()) {
+        std::memcpy(buffer + 8, &app_id_salt, 4);
```
lioncash: magic constants (`4` can be `sizeof(u32)`)

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -176,14 +177,29 @@ void SecureInfoGetRegion(Service::Interface* self) {
...
+        u8 hash[CryptoPP::SHA256::DIGESTSIZE];
```
lioncash: `std::array`

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -176,14 +177,29 @@ void SecureInfoGetRegion(Service::Interface* self) {
...
 
 void GenHashConsoleUnique(Service::Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 app_id_salt = cmd_buff[1];
-
-    cmd_buff[1] = RESULT_SUCCESS.raw;
-    cmd_buff[2] = 0x33646D6F ^ (app_id_salt & 0xFFFFF); // 3dmoo hash
-    cmd_buff[3] = 0x6F534841 ^ (app_id_salt & 0xFFFFF);
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x03, 1, 0);
+    u32 app_id_salt = rp.Pop<u32>() & 0x000FFFFF;
```
lioncash: `const u32`

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -176,14 +177,29 @@ void SecureInfoGetRegion(Service::Interface* self) {
...
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(3, 0);
+
+    u8 buffer[12];
+    ResultCode result = GetConfigInfoBlock(ConsoleUniqueID2BlockID, 8, 2, buffer);
```
lioncash: `const ResultCode`

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -176,14 +178,29 @@ void SecureInfoGetRegion(Service::Interface* self) {
...
+
+    std::array<u8, 12> buffer;
+    const ResultCode result = GetConfigInfoBlock(ConsoleUniqueID2BlockID, 8, 2, buffer.data());
+    rb.Push(result);
+    if (result.IsSuccess()) {
+        std::memcpy(buffer.data() + 8, &app_id_salt, sizeof(u32));
+        std::array<u8, CryptoPP::SHA256::DIGESTSIZE> hash;
+        CryptoPP::SHA256().CalculateDigest(hash.data(), buffer.data(), sizeof(buffer));
+        u32 low, high;
+        memcpy(&low, hash.data() + CryptoPP::SHA256::DIGESTSIZE - 8, sizeof(u32));
```
lioncash: You can just do `&hash[CryptoPP::SHA256::DIGESTSIZE - 8]` instead of calling `data()` and using it like an offset point. Ditto below.

PR #2626, opened 2017-03-13
Fix building with MSVC 2017

yuriks:


jroweboy:
see also #2623

Subv:
Do we have to update the Qt binaries in our Qt repo?

yuriks:
No, 2017 (vc14.0) is binary compatible with 2015 (vc14.1)

PR #2628, opened 2017-03-13
Services/UDS: Initial support for hosting local-wlan networks.

Subv:
Currently it will let games create a network as hosts, but will not broadcast it anywhere and will not allow clients to connect.

### DIFF
--- a/src/core/hle/service/nwm/nwm_uds.cpp
+++ b/src/core/hle/service/nwm/nwm_uds.cpp
@@ -297,6 +297,7 @@ static void DestroyNetwork(Interface* self) {
     CoreTiming::UnscheduleEvent(beacon_broadcast_event, 0);
 
     connection_status.status = static_cast<u8>(NetworkStatus::NotConnected);
+    connection_status_event->Signal();
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
 
--- a/src/core/hle/service/nwm/nwm_uds.cpp
+++ b/src/core/hle/service/nwm/nwm_uds.cpp
@@ -247,6 +247,9 @@ static void BeginHostingNetwork(Interface* self) {
 
     Memory::ReadBlock(network_info_address, &network_info, sizeof(NetworkInfo));
 
+    // The real UDS module throws a fatal error if this assert fails.
+    ASSERT_MSG(network_info.max_nodes > 1, "Trying to host a network of only one member.");
+
     connection_status.status = static_cast<u32>(NetworkStatus::ConnectedAsHost);
     connection_status.max_nodes = network_info.max_nodes;
 
--- a/src/core/hle/service/nwm/nwm_uds.cpp
+++ b/src/core/hle/service/nwm/nwm_uds.cpp
@@ -35,7 +35,7 @@ static NodeInfo node_info;
 // Mapping of bind node ids to their respective events.
 static std::unordered_map<u32, Kernel::SharedPtr<Kernel::Event>> bind_node_events;
 
-// The wifi network channel that the network is currently on.
+// The WiFi network channel that the network is currently on.
 // Since we're not actually interacting with physical radio waves, this is just a dummy value.
 static u8 network_channel = DefaultNetworkChannel;
 
@@ -320,10 +320,10 @@ static void GetChannel(Interface* self) {
     IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1A, 0, 0);
     IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
 
-    u8 channel = network_channel;
+    bool is_connected = connection_status.status !=
+        static_cast<u32>(NetworkStatus::NotConnected);
 
-    if (connection_status.status == static_cast<u32>(NetworkStatus::NotConnected))
-        channel = 0;
+    u8 channel = is_connected ? network_channel : 0;
 
     rb.Push(RESULT_SUCCESS);
     rb.PushRaw(channel);
--- a/src/core/hle/service/nwm/nwm_uds.cpp
+++ b/src/core/hle/service/nwm/nwm_uds.cpp
@@ -38,11 +38,8 @@ static std::unordered_map<u32, Kernel::SharedPtr<Kernel::Event>> bind_node_event
 // Since we're not actually interacting with physical radio waves, this is just a dummy value.
 static u8 network_channel = DefaultNetworkChannel;
 
-// The identifier of the network kind, this is used to filter away networks that we're not interested in.
-static u32 wlan_comm_id = 0;
-
-// Application data that is sent when broadcasting the beacon frames.
-static std::vector<u8> application_data;
+// Information about the network that we're currently connected to.
+static NetworkInfo network_info;
 
 /**
  * NWM_UDS::Shutdown service function
@@ -244,12 +241,10 @@ static void BeginHostingNetwork(Interface* self) {
 
     LOG_DEBUG(Service_NWM, "called");
 
-    NetworkInfo network_info;
     Memory::ReadBlock(network_info_address, &network_info, sizeof(NetworkInfo));
 
     connection_status.status = static_cast<u32>(NetworkStatus::ConnectedAsHost);
     connection_status.max_nodes = network_info.max_nodes;
-    wlan_comm_id = network_info.wlan_comm_id;
 
     // There's currently only one node in the network (the host).
     connection_status.total_nodes = 1;
@@ -262,9 +257,6 @@ static void BeginHostingNetwork(Interface* self) {
     // If the game has a preferred channel, use that instead.
     if (network_info.channel != 0)
         network_channel = network_info.channel;
-
-    // Clear the pre-existing application data.
-    application_data.clear();
 
     connection_status_event->Signal();
 
@@ -333,8 +325,8 @@ static void SetApplicationData(Interface* self) {
         return;
     }
 
-    application_data.resize(size);
-    Memory::ReadBlock(address, application_data.data(), size);
+    network_info.application_data_size = size;
+    Memory::ReadBlock(address, network_info.application_data.data(), size);
 
     rb.Push(RESULT_SUCCESS);
 }
@@ -379,7 +371,7 @@ NWM_UDS::NWM_UDS() {
 }
 
 NWM_UDS::~NWM_UDS() {
-    application_data.clear();
+    network_info = {};
     bind_node_events.clear();
     connection_status_event = nullptr;
     recv_buffer_memory = nullptr;
--- a/src/core/hle/service/nwm/nwm_uds.cpp
+++ b/src/core/hle/service/nwm/nwm_uds.cpp
@@ -7,6 +7,7 @@
 #include <vector>
 #include "common/common_types.h"
 #include "common/logging/log.h"
+#include "core/core_timing.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/kernel/shared_memory.h"
 #include "core/hle/result.h"
@@ -40,6 +41,9 @@ static u8 network_channel = DefaultNetworkChannel;
 
 // Information about the network that we're currently connected to.
 static NetworkInfo network_info;
+
+// Event that will generate and send the 802.11 beacon frames.
+static int beacon_broadcast_event;
 
 /**
  * NWM_UDS::Shutdown service function
@@ -260,13 +264,42 @@ static void BeginHostingNetwork(Interface* self) {
 
     connection_status_event->Signal();
 
-    // TODO(Subv): Start broadcasting the network, send a beacon frame every 102.4ms.
+    // Start broadcasting the network, send a beacon frame every 102.4ms.
+    CoreTiming::ScheduleEvent(msToCycles(DefaultBeaconInterval * MillisecondsPerTU),
+                              beacon_broadcast_event, 0);
 
     LOG_WARNING(Service_NWM,
                 "An UDS network has been created, but broadcasting it is unimplemented.");
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
     rb.Push(RESULT_SUCCESS);
+}
+
+/**
+ * NWM_UDS::DestroyNetwork service function.
+ * Closes the network that we're currently hosting.
+ *  Inputs:
+ *      0 : Command header.
+ *  Outputs:
+ *      0 : Return header
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void DestroyNetwork(Interface* self) {
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x08, 0, 0);
+
+    // TODO(Subv): Find out what happens if this is called while
+    // no network is being hosted.
+
+    // Unschedule the beacon broadcast event.
+    CoreTiming::UnscheduleEvent(beacon_broadcast_event, 0);
+
+    connection_status.status = static_cast<u8>(NetworkStatus::NotConnected);
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+
+    rb.Push(RESULT_SUCCESS);
+
+    LOG_WARNING(Service_NWM, "called");
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/core/hle/service/nwm/nwm_uds.h
```diff
@@ -4,13 +4,68 @@
...
+    u64_le friend_code_seed;
+    u16_le username[10];
+    INSERT_PADDING_BYTES(4);
+    u16_le network_node_id;
+    INSERT_PADDING_BYTES(6);
+};
+
+static_assert(sizeof(NodeInfo) == 40, "NodeInfo has incorrect size.");
+
+enum NetworkStatus {
```
lioncash: This should likely be an `enum class`

src/core/hle/service/nwm/nwm_uds.h
```diff
@@ -4,13 +4,68 @@
...
 
+#include "common/swap.h"
 #include "core/hle/service/service.h"
 
 // Local-WLAN service
 
 namespace Service {
 namespace NWM {
 
+const u32 ApplicationDataSize = 0xC8;
```
lioncash: `size_t`

src/core/hle/service/nwm/nwm_uds.h
```diff
@@ -4,13 +4,68 @@
 
 #pragma once
 
+#include "common/swap.h"
```
lioncash: `common/common_types.h` also needs to be included.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -5,12 +5,39 @@
 #include "common/common_types.h"
 #include "common/logging/log.h"
 #include "core/hle/kernel/event.h"
+#include "core/hle/kernel/shared_memory.h"
```
lioncash: `<cstring>`, `<unordered_map>`, `<vector>`, `core/hle/result.h` and `core/memory.h`  need to be included.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +106,153 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
-    cmd_buff[3] = Kernel::g_handle_table.Create(uds_handle_event)
-                      .MoveFrom(); // TODO(purpasmart): Verify if this is a event handle
-    */
-    cmd_buff[0] = IPC::MakeHeader(0x1B, 1, 2);
-    cmd_buff[1] = ResultCode(static_cast<ErrorDescription>(2), ErrorModule::UDS,
-                             ErrorSummary::StatusChanged, ErrorLevel::Status)
-                      .raw;
+    cmd_buff[3] = Kernel::g_handle_table.Create(connection_status_event).MoveFrom();
+
+    LOG_DEBUG(Service_NWM, "called sharedmem_size=0x%08X, version=0x%08X, value=%u, handle=0x%08X",
```
wwylele: `"called sharedmem_size=0x%08X, version=0x%08X,  sharedmem_handle=0x%08X"`

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +106,153 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+              sharedmem_size, version, sharedmem_handle);
+}
+
+static void GetConnectionStatus(Interface* self) {
```
wwylele: Need documentation for these new functions. Better to have some comment about what they do and when will game call them - I can't get these idea quickly by just seeing the implementation.

src/core/hle/service/nwm/nwm_uds.h
```diff
@@ -4,13 +4,70 @@
...
 
 namespace Service {
 namespace NWM {
 
+const size_t ApplicationDataSize = 0xC8;
+const u8 DefaultNetworkChannel = 11;
+
+struct NodeInfo {
+    u64_le friend_code_seed;
+    u16_le username[10];
```
wwylele: `std::array`

wwylele:
Add this to BE to test if Pokemon ORAS still randomly crashes with NWM initialized. The original issue was #2041 and it was temporarily fixed by #2126 .

src/core/hle/service/nwm/nwm_uds.cpp
Subv: Addressed

Subv:
I've addressed all the current comments.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +110,231 @@ static void RecvBeaconBroadcastData(Interface* self) {
 /**
  * NWM_UDS::Initialize service function
  *  Inputs:
- *      1 : Unknown
- *   2-11 : Input Structure
- *     12 : Unknown u16
+ *      1 : Shared memory size
+ *   2-11 : Input NodeInfo Structure
+ *     12 : Version
```
wwylele: Still need to keep the comment that it is a `u16`. (or it was wrong?)

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +110,231 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
-    LOG_WARNING(Service_NWM, "(STUBBED) called unk1=0x%08X, unk2=0x%08X, value=%u, handle=0x%08X",
-                unk1, unk2, value, handle);
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1B, 12, 2);
+
+    u32 sharedmem_size = rp.Pop<u32>();
+
+    // Update the node information with the data the game gave us.
+    rp.PopRaw(node_info);
+
+    u32 version;
```
wwylele: related to the comment above. Should be `u16`?

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +110,231 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0xB, 0, 0);
+    IPC::RequestBuilder rb = rp.MakeBuilder(13, 0);
+
+    rb.Push(RESULT_SUCCESS);
+    rb.PushRaw(connection_status);
+
+    LOG_DEBUG(Service_NWM, "called");
+}
+
+static void Bind(Interface* self) {
```
wwylele: lack of documentation

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +110,231 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 2);
+
+    // TODO(Subv): Store the data channel and verify it when receiving data frames.
+
+    LOG_DEBUG(Service_NWM, "called");
+
+    if (data_channel == 0) {
+        rb.Push(ResultCode(ErrorDescription::NotAuthorized, ErrorModule::UDS,
+                           ErrorSummary::WrongArgument, ErrorLevel::Usage));
```
wwylele: You need either make a different header `rp.MakeBuilder(1, 0);` or use the `Skip` function implemented in #2533

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +110,231 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+
+    // If the game has a preferred channel, use that instead.
+    if (network_info.channel != 0)
+        network_channel = network_info.channel;
+
+    connection_status_event->Signal();
+
+    // Start broadcasting the network, send a beacon frame every 102.4ms.
+    CoreTiming::ScheduleEvent(msToCycles(DefaultBeaconInterval * MillisecondsPerTU),
+                              beacon_broadcast_event, 0);
```
wwylele: This event need to be unscheduled in `DestroyNetwork`. iirc games like Pokemon likes to do BeginHosting/Destroy look in background, and this will cause event piling

Subv:
Addressed all comments.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +110,273 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+static void BeginHostingNetwork(Interface* self) {
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1D, 1, 4);
+
+    const u32 passphrase_size = rp.Pop<u32>();
+
+    size_t desc_size;
+    IPC::MappedBufferPermissions desc_permission;
+    const VAddr network_info_address = rp.PopMappedBuffer(&desc_size, &desc_permission);
+    ASSERT(desc_size == sizeof(NetworkInfo));
+    const VAddr passphrase_address = rp.PopMappedBuffer(&desc_size, &desc_permission);
```
wwylele: These two should be `PopStaticBuffer`

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +110,273 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+ *      2 : Channel of the current WiFi network connection.
+ */
+static void SetApplicationData(Interface* self) {
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1A, 1, 2);
+
+    u32 size = rp.Pop<u32>();
+
+    size_t desc_size;
+    IPC::MappedBufferPermissions desc_permission;
+    const VAddr address = rp.PopMappedBuffer(&desc_size, &desc_permission);
```
wwylele: Should be `PopStaticBuffer` according to 3dbrew

Subv:
Call for reviews on this.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +110,271 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+static void DestroyNetwork(Interface* self) {
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x08, 0, 0);
+
+    // TODO(Subv): Find out what happens if this is called while
+    // no network is being hosted.
+
+    // Unschedule the beacon broadcast event.
+    CoreTiming::UnscheduleEvent(beacon_broadcast_event, 0);
+
+    connection_status.status = static_cast<u8>(NetworkStatus::NotConnected);
```
B3n30: I don't know if connection_status_event should signal here but for what i know it should be closed here. According to @wwylele some games frequently call BeginHostingNetwork and DestroyNetwork. Wouldn't this cause a lot of dead handles?

Subv: It should probably be signaled here, but we don't have to manage the lifetime of kernel objects ourselves due to the magic of Kernel::SharedPtr

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +110,271 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+    ASSERT(desc_size == passphrase_size);
+
+    // TODO(Subv): Store the passphrase and verify it when attempting a connection.
+
+    LOG_DEBUG(Service_NWM, "called");
+
+    Memory::ReadBlock(network_info_address, &network_info, sizeof(NetworkInfo));
+
+    connection_status.status = static_cast<u32>(NetworkStatus::ConnectedAsHost);
+    connection_status.max_nodes = network_info.max_nodes;
```
Subv: Indeed, I'll add a log

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -32,14 +64,14 @@ static void Shutdown(Interface* self) {
...
- *      2 : Unknown
- *      3 : Unknown
- *      4 : MAC address?
- *   6-14 : Unknown, usually zero / uninitialized?
+ *    2-3 : Unknown
+ *    4-5 : Host MAC address.
+ *   6-14 : Unused
  *     15 : WLan Comm ID
- *     16 : This is the ID also located at offset 0xE in the CTR-generation structure.
+ *     16 : Id
```
JayFoxRox: nit: Spelling of ID differs here and there
Also is there any reason why the comment was removed?

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +109,234 @@ static void RecvBeaconBroadcastData(Interface* self) {
 /**
  * NWM_UDS::Initialize service function
  *  Inputs:
- *      1 : Unknown
- *   2-11 : Input Structure
- *     12 : Unknown u16
+ *      1 : Shared memory size
+ *   2-11 : Input NodeInfo Structure
+ *     12 : 2-byte Version
```
JayFoxRox: nit: `2-byte Version` of what? -> wording is slightly misleading

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +109,234 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+    connection_status.total_nodes = 1;
+    // The host is always the first node
+    connection_status.network_node_id = 1;
+    node_info.network_node_id = 1;
+    // Set the bit 0 in the nodes bitmask to indicate that node 1 is already taken.
+    connection_status.node_bitmask |= 1;
+
+    // If the game has a preferred channel, use that instead.
+    if (network_info.channel != 0)
+        network_channel = network_info.channel;
```
JayFoxRox: not sure how the rest of the HLE handles it, but I prefer explicit scopes to avoid `goto fail`

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -1,16 +1,48 @@
...
+
+// Node information about the current 3DS.
+// TODO(Subv): Keep an array of all nodes connected to the network,
+// that data has to be retransmitted in every beacon frame.
+static NodeInfo node_info;
+
+// Mapping of bind node ids to their respective events.
+static std::unordered_map<u32, Kernel::SharedPtr<Kernel::Event>> bind_node_events;
+
+// The wifi network channel that the network is currently on.
```
JayFoxRox: nit: Just noticed you sometimes switch between `WLan`, `wifi` and `WiFi`

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -77,42 +109,234 @@ static void RecvBeaconBroadcastData(Interface* self) {
...
+ *      2 : Channel of the current WiFi network connection.
+ */
+static void GetChannel(Interface* self) {
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1A, 0, 0);
+    IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
+
+    u8 channel = network_channel;
+
+    if (connection_status.status == static_cast<u32>(NetworkStatus::NotConnected))
+        channel = 0;
```
JayFoxRox: nit: I think creating a `bool is_connected` and then doing a ternary bool op / select (`is_connected ? network_channel : 0;`) would be nicer.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -148,13 +372,20 @@ const Interface::FunctionInfo FunctionTable[] = {
...
 
 NWM_UDS::~NWM_UDS() {
-    uds_handle_event = nullptr;
+    application_data.clear();
+    bind_node_events.clear();
+    connection_status_event = nullptr;
+    recv_buffer_memory = nullptr;
+
+    connection_status = {};
+    connection_status.status = static_cast<u32>(NetworkStatus::NotConnected);
```
JayFoxRox: If these variables are supposed to represent a virtual network device or something I'd recommend putting them into a seperate struct / class and have a `reset()` function or something there. It took me too long that you had to do all of this because you are working with static variables.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -333,8 +325,8 @@ static void SetApplicationData(Interface* self) {
         return;
```
JayFoxRox: Should we log this case (looks like a case which will be rarely hit)? The first comment made it seem like application_data can resize however it wants, now it looks like it's fixed in max size. Clarify please: where does the current max size come from?

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -41,6 +42,9 @@ static u8 network_channel = DefaultNetworkChannel;
 // Information about the network that we're currently connected to.
 static NetworkInfo network_info;
 
+// Event that will generate and send the 802.11 beacon frames.
+static int beacon_broadcast_event;
```
JayFoxRox: Just double-checking: event handles are stored as `int` in the HLE?

src/core/hle/service/nwm/nwm_uds.cpp
Subv: The general codestyle is to not use explicit scope for one-line statements

src/core/hle/service/nwm/nwm_uds.cpp
Subv: The version of the service, this wording is used throughout the entire HLE layer

src/core/hle/service/nwm/nwm_uds.cpp
Subv: The previous comment was pretty much useless, i've settled for just calling this member "id" everywhere.
Where is the spelling different than here?

src/core/hle/service/nwm/nwm_uds.cpp
Subv: Do you mean the entirety of this destructor? That's the convention we use for this kind of services.

Ideally all this would be inside a map in the Interface class, keyed by the Session that called the function, but alas, that's not ready yet.

PR #2633, opened 2017-03-20
implemented basic cia loader

jfmherokiller:
A very raw modification to the ncch loader to allow it to load cias.

This is an addition to the ncch loader(ncch.cpp/.h) to allow it find the ncch header inside of the cia files by using the data stored in the cia header. It performs 64byte aligned adds using the sizes of the header itself, the certificate chain, the Ticket, and the TMD file data. 

Through the proccess of adding up these sizes it is able to reach the offset of the ncch archive stored in the content section.

I plan to follow this up with pull request to refactor the loader setup to be hopefully more modular. this is more or less a bleeding edge enhancement right now.


src/core/loader/ncch.h
```diff
@@ -9,6 +9,34 @@
 #include "common/common_types.h"
 #include "common/swap.h"
 #include "core/loader/loader.h"
+////////////////////////////////////////////////////////////////////////////////////////////////////
+/// CIA header
+struct CIA_Header {
```
lioncash: `CIAHeader`

src/core/loader/ncch.h
```diff
@@ -9,6 +9,34 @@
...
+    u8 version[2];
+    u32_le certsize;
+    u32_le ticketsize;
+    u32_le tmdsize;
+    u32_le metasize;
+    u64_le contentsize;
+    u8 contentindex[0x2000];
+};
+
+struct cia_context {
```
lioncash: `CIAContext`

src/core/loader/ncch.h
```diff
@@ -9,6 +9,34 @@
 #include "common/common_types.h"
 #include "common/swap.h"
 #include "core/loader/loader.h"
+////////////////////////////////////////////////////////////////////////////////////////////////////
+/// CIA header
+struct CIA_Header {
+    u32_le headersize;
```
lioncash: almost all of these are missing an underscore between separate words.

src/core/loader/ncch.h
```diff
@@ -9,6 +9,34 @@
...
+    u32_le ticketsize;
+    u32_le tmdsize;
+    u32_le metasize;
+    u64_le contentsize;
+    u8 contentindex[0x2000];
+};
+
+struct cia_context {
+    CIA_Header header;
+    u32 sizeheader;
```
lioncash: ditto about underscores. This should also likely not be inverted (e.g. header_size is likely more appropriate).

src/core/loader/ncch.cpp
```diff
@@ -411,4 +428,30 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_
     return ResultStatus::ErrorNotUsed;
 }
 
+u32 align(u32 offset, u32 alignment) {
```
lioncash: `common/alignment.h` likely already has a function that does this.

src/core/loader/ncch.cpp
```diff
@@ -240,26 +244,39 @@ ResultStatus AppLoader_NCCH::LoadSectionExeFS(const char* name, std::vector<u8>&
...
+    u32 tempmagic;
     if (is_exefs_loaded)
         return ResultStatus::Success;
 
     if (!file.IsOpen())
         return ResultStatus::Error;
-
+    file.Seek(0, SEEK_SET);
+    file.ReadBytes(&tempmagic, 4);
+    if (0x2020 == tempmagic) {
```
lioncash: This should be swapped in terms of placement. It's 2017, compilers warn about this sort of thing.

src/core/loader/ncch.cpp
```diff
@@ -411,4 +428,30 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_
     return ResultStatus::ErrorNotUsed;
 }
 
+u32 align(u32 offset, u32 alignment) {
+    u32 mask = ~(alignment - 1);
+
+    return (offset + (alignment - 1)) & mask;
+}
+void AppLoader_NCCH::ReadCiaFile(cia_context& returnme) {
```
lioncash: Just return the instance by value like:

```cpp
cia_context AppLoader_NCCH::ReadCiaFile()
```

This is a pre-c++11ism the way it currently is.

src/core/loader/ncch.cpp
```diff
@@ -411,4 +428,30 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_
...
+    // Reset read pointer in case this file has been read before.
+    file.Seek(0, SEEK_SET);
+    cia_context* ctx = new cia_context;
```
lioncash: This can just be allocated on the stack. The way it currently is results in a memory leak.

src/core/loader/ncch.h
```diff
@@ -9,6 +9,34 @@
 #include "common/common_types.h"
 #include "common/swap.h"
 #include "core/loader/loader.h"
+////////////////////////////////////////////////////////////////////////////////////////////////////
+/// CIA header
```
lioncash: This is self-evident by the struct being called `CIA_Header`

src/core/loader/ncch.cpp
jfmherokiller: i was afraid of that but id didn't know the correct way to do it

src/core/loader/ncch.cpp
```diff
@@ -240,26 +244,38 @@ ResultStatus AppLoader_NCCH::LoadSectionExeFS(const char* name, std::vector<u8>&
...
+        // Skip NCSD header and load first NCCH (NCSD is just a container of NCCH files)...
+        if (MakeMagic('N', 'C', 'S', 'D') == ncch_header.magic) {
+            LOG_DEBUG(Loader, "Only loading the first (bootable) NCCH within the NCSD file!");
+            ncch_offset = 0x4000;
+            file.Seek(ncch_offset, SEEK_SET);
+            file.ReadBytes(&ncch_header, sizeof(NCCH_Header));
+        }
+    } else {
+        CIAContext ctx = ReadCiaFile();
+        LOG_DEBUG(Loader, "Attempting to load CIA contents hold onto your hats");
```
JayFoxRox: Yo surferboy, what's poppin?

- I guess this could need some more seriousness

src/core/loader/ncch.cpp
```diff
@@ -411,4 +427,24 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_
...
+    ctx.cert_size = ctx.header.cert_size;
+    ctx.tik_size = ctx.header.ticket_size;
+    ctx.tmd_size = ctx.header.tmd_size;
+    ctx.content_size = ctx.header.content_size;
+    ctx.meta_size = ctx.header.meta_size;
+    ctx.certs_offset = Common::AlignUp(ctx.header_size, 64);
+    ctx.tik_offset = Common::AlignUp(ctx.certs_offset + ctx.cert_size, 64);
+    ctx.tmd_offset = Common::AlignUp(ctx.tik_offset + ctx.tik_size, 64);
+    ctx.content_offset = Common::AlignUp(ctx.tmd_offset + ctx.tmd_size, 64);
+  return ctx;
```
JayFoxRox: indentation looks bad

src/core/loader/loader.cpp
```diff
@@ -113,8 +113,8 @@ static std::unique_ptr<AppLoader> GetFileLoader(FileUtil::IOFile&& file, FileTyp
     // NCCH/NCSD container formats.
     case FileType::CXI:
     case FileType::CCI:
+    case FileType::CIA:
         return std::make_unique<AppLoader_NCCH>(std::move(file), filepath);
-
```
JayFoxRox: why was this removed?

src/core/loader/ncch.h
```diff
@@ -157,7 +184,6 @@ static_assert(sizeof(ExHeader_Header) == 0x800, "ExHeader structure size is wron
 // Loader namespace
 
 namespace Loader {
-
```
JayFoxRox: Why are these randomly removed?

src/core/loader/ncch.cpp
```diff
@@ -107,6 +107,10 @@ static bool LZSS_Decompress(const u8* compressed, u32 compressed_size, u8* decom
 
 FileType AppLoader_NCCH::IdentifyType(FileUtil::IOFile& file) {
     u32 magic;
+    file.Seek(0, SEEK_SET);
+    file.ReadBytes(&magic, 4);
+    if (0x2020 == magic)
```
Subv: Magic headers don't mean you have to use magic numbers for them

src/core/loader/ncch.cpp
```diff
@@ -411,4 +427,24 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_
     return ResultStatus::ErrorNotUsed;
 }
 
+CIAContext AppLoader_NCCH::ReadCiaFile() {
+    // Reset read pointer in case this file has been read before.
+    file.Seek(0, SEEK_SET);
+    CIAContext ctx;
```
Subv: Just...Why reassign all those fields?

src/core/loader/ncch.cpp
```diff
@@ -240,26 +244,38 @@ ResultStatus AppLoader_NCCH::LoadSectionExeFS(const char* name, std::vector<u8>&
...
 ResultStatus AppLoader_NCCH::LoadExeFS() {
+    bool is_cia = false;
+    u32 tempmagic;
     if (is_exefs_loaded)
         return ResultStatus::Success;
 
     if (!file.IsOpen())
         return ResultStatus::Error;
-
+    file.Seek(0, SEEK_SET);
```
jroweboy: The previously mentioned "IsValidCia" method would work wonders here. With that you can do something like

```cpp
if IsValidCIA:
  seek to ncch header location
else:
  seek to the start
```

Then you don't need to special case `is_cia`

src/core/loader/ncch.cpp
jfmherokiller: that was the original implementation as sourced from ctrtool

src/core/loader/ncch.h
```diff
@@ -8,7 +8,27 @@
 #include "common/bit_field.h"
 #include "common/common_types.h"
 #include "common/swap.h"
+#include "common/alignment.h"
 #include "core/loader/loader.h"
+struct CIAHeader {
+    u32_le header_size;
+    u8 type[2];
```
jroweboy: why are type and version u8 arrays instead of u16_le?

src/core/loader/ncch.h
```diff
@@ -250,6 +270,7 @@ class AppLoader_NCCH final : public AppLoader {
     ExHeader_Header exheader_header;
 
     std::string filepath;
+    CIAContext ReadCiaFile();
```
jroweboy: I'd personally rather see the CIA related methods be just a static method in the file rather part of the class

src/core/loader/ncch.cpp
```diff
@@ -411,4 +427,17 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_
...
+    // Reset read pointer in case this file has been read before.
+    file.Seek(0, SEEK_SET);
+    CIAContext ctx;
+
+    file.ReadBytes(&ctx.header, sizeof(CIAHeader));
+    ctx.certs_offset = Common::AlignUp(ctx.header.header_size, 64);
+    ctx.tik_offset = Common::AlignUp(ctx.certs_offset + ctx.header.cert_size, 64);
+    ctx.tmd_offset = Common::AlignUp(ctx.tik_offset + ctx.header.ticket_size, 64);
+    ctx.content_offset = Common::AlignUp(ctx.tmd_offset + ctx.header.tmd_size, 64);
+    return ctx;
```
jroweboy: Is the CIA context useful for anything if all you are going to use from it is the content_offset? Maybe its better to have this return the content_offset and rename the method to something like CIAFindNCCHOffset (note: I'm bad at names) This "find offset" function can be used as part of the validate process to check that the file contains a NCCH in it.

src/core/loader/ncch.cpp
```diff
@@ -120,6 +116,11 @@ FileType AppLoader_NCCH::IdentifyType(FileUtil::IOFile& file) {
 
     if (MakeMagic('N', 'C', 'C', 'H') == magic)
         return FileType::CXI;
+    file.Seek(0, SEEK_SET);
+    if (1 != file.ReadArray<u32>(&magic, 1))
```
InusualZ: You did it again. Should be `file.ReadArray<u32>(&magic, 1) != 1`

jfmherokiller: the reason i had it that way was because of line number 111 which does it this way
`if (1 != file.ReadArray<u32>(&magic, 1))`

jroweboy:
The current code causes issues in some games and prevents them from loading. For example: http://i.imgur.com/Z02jzRe.png those two games are missing the icon because citra can't load them when using a build with just this PR added to it. Because of this I'm removing this from bleeding edge until the issues are resolved.

jfmherokiller:
@jroweboy i removed my uh lazy hack i used to check if a file was a cia and correctly implemented a check based up calculating the offset of the ncch magic including an additional 256 byte padding that i observed in some cias

wwylele:
is #2644 planned to go before this?

jfmherokiller:
@wwylele yes and once #2644 gets merged i plan to close this pull request. This is because I already did the work of separating the cia code into its own class. However, i do wish to keep this pull request open as an alternative if #2644 does not get merged.

jfmherokiller:
closing because im working on a rewrite that will be opened once #2644 gets merged

PR #2634, opened 2017-03-21
shared_page: stub battery state

wwylele:
We already stubbed the battery state in the PTM service, but there is another API that is more common used by games, which is the shared page. This PR stubbed the battery state in the shared page with the same value as the one in PTM service.

### DIFF
--- a/src/core/hle/shared_page.cpp
+++ b/src/core/hle/shared_page.cpp
@@ -6,6 +6,7 @@
 #include <cstring>
 #include <ctime>
 #include "core/core_timing.h"
+#include "core/hle/service/ptm/ptm.h"
 #include "core/hle/shared_page.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


Aceofzeroz:
Looking great! It's definitely a welcomed PR for games like Sm4sh. :)
2017-03-21 08_13_06-citra _ head-34ba31b

baka0815:
Is there a reason one would somehow "emulate" the battery state instead of just stubbing it at 100%?

wwylele:
@baka0815 this is stubbing it at 100%

JayFoxRox:
Is this using the same variable to represent the battery state, or are you just setting the same value?
In the feature this might end up in the config menu or something so I fear if it's not the same variable someone might patch it poorly then and we'll have forgotten about this by then.

(Unless there is a good reason to keep them seperated)

wwylele:
@JayFoxRox good point. So I am going to add a "battery" entry in the config and let both PTM and shared_page use it. If no one comes up with a better idea, I am going to do this tomorrow.

mailwl:
why? every app/game (even sysupdater) likes battery 100% and charger plugged. IMHO, it's enough for emulating.
well, in some rare cases plug/unplug notifications may interesting... no, this is no debugger, anyway

baka0815:
@wwylele I know that this stubs it at 100%, I'm just curious if there would ever be a reason to not just stub this but emulate the battery state (however it may be done).

wwylele:
wel, since Subv signed off, I will just leave this PR as is. I will open an issue regarding to the PTM stuff and emulating battery state change etc...

Subv:
IMO that's better left as a separate PR.

PR #2637, opened 2017-03-24
 Added 3DS controls in main window

RedTrainer97:
immagine

You can set the volume from 0% to 100%.

Coming soon: keyboard shortcuts?

Bye and thanks a lot to all contributors ^.^

### DIFF
--- a/src/audio_core/hle/dsp.cpp
+++ b/src/audio_core/hle/dsp.cpp
@@ -10,6 +10,7 @@
 #include "audio_core/hle/source.h"
 #include "audio_core/sink.h"
 #include "audio_core/time_stretch.h"
+#include "core/settings.h"
 
 namespace DSP {
 namespace HLE {
@@ -89,6 +90,15 @@ static bool perform_time_stretching = true;
 static std::unique_ptr<AudioCore::Sink> sink;
 static AudioCore::TimeStretcher time_stretcher;
 
+static StereoFrame16 ChangeVolume(const StereoFrame16& frame) {
+    StereoFrame16 new_frame = StereoFrame16();
+    for (int i = 0; i < frame.size(); i++) {
+        new_frame[i][0] = frame[i][0] * Settings::values.volume;
+        new_frame[i][1] = frame[i][1] * Settings::values.volume;
+    }
+    return new_frame;
+}
+
 static void FlushResidualStretcherAudio() {
     time_stretcher.Flush();
     while (true) {
@@ -100,8 +110,9 @@ static void FlushResidualStretcherAudio() {
 }
 
 static void OutputCurrentFrame(const StereoFrame16& frame) {
+    StereoFrame16 new_frame = ChangeVolume(frame);
     if (perform_time_stretching) {
-        time_stretcher.AddSamples(&frame[0][0], frame.size());
+        time_stretcher.AddSamples(&new_frame[0][0], frame.size());
         std::vector<s16> stretched_samples = time_stretcher.Process(sink->SamplesInQueue());
         sink->EnqueueSamples(stretched_samples.data(), stretched_samples.size() / 2);
     } else {
@@ -112,7 +123,7 @@ static void OutputCurrentFrame(const StereoFrame16& frame) {
             return;
         }
 
-        sink->EnqueueSamples(&frame[0][0], frame.size());
+        sink->EnqueueSamples(&new_frame[0][0], frame.size());
     }
 }
 
--- a/src/audio_core/hle/dsp.cpp
+++ b/src/audio_core/hle/dsp.cpp
@@ -10,7 +10,6 @@
 #include "audio_core/hle/source.h"
 #include "audio_core/sink.h"
 #include "audio_core/time_stretch.h"
-#include "core/settings.h"
 
 namespace DSP {
 namespace HLE {
@@ -90,15 +89,6 @@ static bool perform_time_stretching = true;
 static std::unique_ptr<AudioCore::Sink> sink;
 static AudioCore::TimeStretcher time_stretcher;
 
-static StereoFrame16 ChangeVolume(const StereoFrame16& frame) {
-    StereoFrame16 new_frame = StereoFrame16();
-    for (int i = 0; i < frame.size(); i++) {
-        new_frame[i][0] = frame[i][0] * Settings::values.volume;
-        new_frame[i][1] = frame[i][1] * Settings::values.volume;
-    }
-    return new_frame;
-}
-
 static void FlushResidualStretcherAudio() {
     time_stretcher.Flush();
     while (true) {
@@ -110,9 +100,8 @@ static void FlushResidualStretcherAudio() {
 }
 
 static void OutputCurrentFrame(const StereoFrame16& frame) {
-    StereoFrame16 new_frame = ChangeVolume(frame);
     if (perform_time_stretching) {
-        time_stretcher.AddSamples(&new_frame[0][0], frame.size());
+        time_stretcher.AddSamples(&frame[0][0], frame.size());
         std::vector<s16> stretched_samples = time_stretcher.Process(sink->SamplesInQueue());
         sink->EnqueueSamples(stretched_samples.data(), stretched_samples.size() / 2);
     } else {
@@ -123,7 +112,7 @@ static void OutputCurrentFrame(const StereoFrame16& frame) {
             return;
         }
 
-        sink->EnqueueSamples(&new_frame[0][0], frame.size());
+        sink->EnqueueSamples(&frame[0][0], frame.size());
     }
 }
 
--- a/src/citra/config.cpp
+++ b/src/citra/config.cpp
@@ -100,6 +100,7 @@ void Config::ReadValues() {
     Settings::values.enable_audio_stretching =
         sdl2_config->GetBoolean("Audio", "enable_audio_stretching", true);
     Settings::values.audio_device_id = sdl2_config->Get("Audio", "output_device", "auto");
+    Settings::values.volume = sdl2_config->GetReal("Audio", "volume", 1);
 
     // Data Storage
     Settings::values.use_virtual_sd =
--- a/src/citra/config.cpp
+++ b/src/citra/config.cpp
@@ -100,7 +100,6 @@ void Config::ReadValues() {
     Settings::values.enable_audio_stretching =
         sdl2_config->GetBoolean("Audio", "enable_audio_stretching", true);
     Settings::values.audio_device_id = sdl2_config->Get("Audio", "output_device", "auto");
-    Settings::values.volume = sdl2_config->GetReal("Audio", "volume", 1);
 
     // Data Storage
     Settings::values.use_virtual_sd =
--- a/src/citra/default_ini.h
+++ b/src/citra/default_ini.h
@@ -108,6 +108,10 @@ enable_audio_stretching =
 # auto (default): Auto-select
 output_device =
 
+# Output volume.
+# 1.0 (default): 100%, 0.0; mute
+volume =
+
 [Data Storage]
 # Whether to create a virtual SD card.
 # 1 (default): Yes, 0: No
--- a/src/citra/default_ini.h
+++ b/src/citra/default_ini.h
@@ -108,10 +108,6 @@ enable_audio_stretching =
 # auto (default): Auto-select
 output_device =
 
-# Output volume.
-# 1.0 (default): 100%, 0.0; mute
-volume =
-
 [Data Storage]
 # Whether to create a virtual SD card.
 # 1 (default): Yes, 0: No
--- a/src/citra_qt/config.cpp
+++ b/src/citra_qt/config.cpp
@@ -87,6 +87,7 @@ void Config::ReadValues() {
         qt_config->value("enable_audio_stretching", true).toBool();
     Settings::values.audio_device_id =
         qt_config->value("output_device", "auto").toString().toStdString();
+    Settings::values.volume = qt_config->value("volume", 1).toFloat();
     qt_config->endGroup();
 
     using namespace Service::CAM;
@@ -213,6 +214,7 @@ void Config::SaveValues() {
     qt_config->setValue("output_engine", QString::fromStdString(Settings::values.sink_id));
     qt_config->setValue("enable_audio_stretching", Settings::values.enable_audio_stretching);
     qt_config->setValue("output_device", QString::fromStdString(Settings::values.audio_device_id));
+    qt_config->setValue("volume", Settings::values.volume);
     qt_config->endGroup();
 
     using namespace Service::CAM;
--- a/src/citra_qt/config.cpp
+++ b/src/citra_qt/config.cpp
@@ -168,6 +168,7 @@ void Config::ReadValues() {
     qt_config->endGroup();
 
     UISettings::values.single_window_mode = qt_config->value("singleWindowMode", true).toBool();
+    UISettings::values.show_3DS_controls = qt_config->value("show3dsControls", true).toBool();
     UISettings::values.display_titlebar = qt_config->value("displayTitleBars", true).toBool();
     UISettings::values.show_status_bar = qt_config->value("showStatusBar", true).toBool();
     UISettings::values.confirm_before_closing = qt_config->value("confirmClose", true).toBool();
@@ -278,6 +279,7 @@ void Config::SaveValues() {
     qt_config->endGroup();
 
     qt_config->setValue("singleWindowMode", UISettings::values.single_window_mode);
+    qt_config->setValue("show3dsControls", UISettings::values.single_window_mode);
     qt_config->setValue("displayTitleBars", UISettings::values.display_titlebar);
     qt_config->setValue("showStatusBar", UISettings::values.show_status_bar);
     qt_config->setValue("confirmClose", UISettings::values.confirm_before_closing);
--- a/src/citra_qt/config.cpp
+++ b/src/citra_qt/config.cpp
@@ -87,7 +87,6 @@ void Config::ReadValues() {
         qt_config->value("enable_audio_stretching", true).toBool();
     Settings::values.audio_device_id =
         qt_config->value("output_device", "auto").toString().toStdString();
-    Settings::values.volume = qt_config->value("volume", 1).toFloat();
     qt_config->endGroup();
 
     using namespace Service::CAM;
@@ -214,7 +213,6 @@ void Config::SaveValues() {
     qt_config->setValue("output_engine", QString::fromStdString(Settings::values.sink_id));
     qt_config->setValue("enable_audio_stretching", Settings::values.enable_audio_stretching);
     qt_config->setValue("output_device", QString::fromStdString(Settings::values.audio_device_id));
-    qt_config->setValue("volume", Settings::values.volume);
     qt_config->endGroup();
 
     using namespace Service::CAM;
--- a/src/citra_qt/configure_audio.cpp
+++ b/src/citra_qt/configure_audio.cpp
@@ -51,6 +51,7 @@ void ConfigureAudio::setConfiguration() {
         }
     }
     ui->audio_device_combo_box->setCurrentIndex(new_device_index);
+    ui->volume_slider->setValue(Settings::values.volume * ui->volume_slider->maximum());
 }
 
 void ConfigureAudio::applyConfiguration() {
@@ -61,6 +62,8 @@ void ConfigureAudio::applyConfiguration() {
     Settings::values.audio_device_id =
         ui->audio_device_combo_box->itemText(ui->audio_device_combo_box->currentIndex())
             .toStdString();
+    Settings::values.volume = (float)ui->volume_slider->value() / ui->volume_slider->maximum();
+
     Settings::Apply();
 }
 
--- a/src/citra_qt/configure_audio.cpp
+++ b/src/citra_qt/configure_audio.cpp
@@ -51,7 +51,6 @@ void ConfigureAudio::setConfiguration() {
         }
     }
     ui->audio_device_combo_box->setCurrentIndex(new_device_index);
-    ui->volume_slider->setValue(Settings::values.volume * ui->volume_slider->maximum());
 }
 
 void ConfigureAudio::applyConfiguration() {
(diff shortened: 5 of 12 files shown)


LittleWhite-tb:
I don't see how it can be useful. You already can change the volume of a specific app in the "volume control" stuff in Windows and Linux. How is your volume slider is different ?

MerryMage:
Volume control would be a post-processing effect, and should probably go [here](https://github.com/citra-emu/citra/blob/93e70a840b5f60c2bb572346a66e2c8774bd1eb2/src/audio_core/hle/dsp.cpp#L158) or in `OutputCurrentFrame` instead. Currently this PR places it in code that's emulating what the DSP firmware is doing.

Minor point from a UX point of view: Linear volume controls are bad. Volume perception is logarithmic.

Helios747:
Modern OSes already handle this.

Why does Citra need to?

jroweboy:
I think a volume control slider is a good idea.

* Unlike Dolphin, the 3ds has a physical volume slider. Having a volume slider that matches how the 3ds volume slider works would be a positive thing in my opinion.
* The claim that the system already has a volume mixer isn't a good argument in my opinion. At least on windows, the system volume mixer is painfully bad. I know you can write this off as "Use a better OS" and "we shouldn't work around bad windows design" but I personally use a per app volume slider for everything that has it. In my book, saying we shouldn't have one because the OS has one has as much weight as saying we should have one because the OS ones tend to suck.

I'm not attached to having a volume control slider, but considering the response so far has been so negative, I thought I would offer a second opinion.

src/citra/default_ini.h
```diff
@@ -123,6 +123,10 @@ enable_audio_stretching =
 # auto (default): Auto-select
 output_device =
 
+# Setup the volume
+# 5
```
wwylele: what is this 5? please be more specific

src/citra/config.cpp
```diff
@@ -117,6 +117,7 @@ void Config::ReadValues() {
     Settings::values.enable_audio_stretching =
         sdl2_config->GetBoolean("Audio", "enable_audio_stretching", true);
     Settings::values.audio_device_id = sdl2_config->Get("Audio", "output_device", "auto");
+    Settings::values.volume = sdl2_config->GetInteger("Audio", "volume", 5);
```
wwylele: I think default to maximum volume makes more sense

src/audio_core/hle/mixers.cpp
```diff
@@ -32,6 +33,7 @@ DspStatus Mixers::Tick(DspConfiguration& config, const IntermediateMixSamples& r
     return GetCurrentStatus();
 }
 
+int currentVolume = 5;
```
wwylele: Please follow our variable naming style. Use `current_volume`.

src/citra_qt/configure_audio.cpp
```diff
@@ -75,3 +79,8 @@ void ConfigureAudio::updateAudioDevices(int sink_index) {
         ui->audio_device_combo_box->addItem(device.c_str());
     }
 }
+
+void ConfigureAudio::updateVolume(int volume)
+{
+    Settings::values.volume = ui->volume_slider->value();
```
wwylele: the volume shouldn't be updated immediately. If the user click "cancel" after changing the volume, the volume is not expected to change

src/core/settings.h
```diff
@@ -118,6 +118,7 @@ struct Values {
     std::string sink_id;
     bool enable_audio_stretching;
     std::string audio_device_id;
+    int volume;
```
wwylele: I was expecting this to be a `float`/`double` with a natural range of [0, 1]. Could you use `float`/`double` instead? or at least, document the range of volume here (is it [0, 10]?).

RavenHome1:
i would love to also see an option to mute audio completely , having audio always running in the background while attempting to do something such as watch a video or maybe listen to a song kinda gets in the way. maybe add a key bind for that.

JayFoxRox:
First a disclaimer: I'm very opposed to the idea of changing the volume of the emu with our own controls. The reason for me is that people should use their audio / sound server mixers anyway. Most OS / sound servers support setting the volume per application anyway (and if they don't they probably will before this emulator is very stable). Adding another layer of indirection will probably degrade sound quality or at the very least be very confusing (enjoy figuring out if your overall volume, per application OS, Citra overall volume or ingame value is wrong and the best values to transmit good audio data).

However, there is one reason why I can agree putting it in: https://www.3dbrew.org/wiki/I2C_Registers
The volume slider is an input which can be used by the software.

So I'm asking to extend the scope of this PR and implement these registers. Alternatively create an issue once this PR is complete. The available number of "steps" should at least reflect the amount of steps representable by the software. This is at least "0x00 - 0x3F" = 64 steps (which the current proposal does not handle). You might also want to check the resolution of register 0x27.

Also I'm not sure how we should control volume. I agree with @MerryMage that the current position in the code is wrong, on the other hand I think mixing SDL buffers overall might not be ideal.
Given what I said initially it might make sense to look wether there is a cross-platform way of synchronizing volume with the sound servers (I've seen some video players do this in the past I believe?).
I'm actually surprised to find out that SDL does not seem to allow this yet.

wwylele:
Addition to the register JayFoxRox mentioned: that register is usually not exposed to games, but its value is forwarded by 3DS system to the service function [HID:GetSoundVolume](https://github.com/citra-emu/citra/blob/master/src/core/hle/service/hid/hid.cpp#L342), which is exposed to games. So if one want to implement this feature, they may want to implement it in this service function directly, or make a complete register system and several service layers to mimic the 3DS way.

RedTrainer97:
I think this feature could be useful for some users, and if someone doesn't want it, he/she can just leave it as is (I'll fix the default volume to 100%).

This is my first commit for Citra, I don't know the whole code, so I've tried to edit it where I think it's more suitable.

To mute Citra, I usually set the Output engine to null, but I'd like too to have a proper option (and maybe shortcuts to do it and to increase and decrease the volume).

src/citra_qt/configure_audio.cpp
```diff
@@ -23,9 +23,14 @@ ConfigureAudio::ConfigureAudio(QWidget* parent)
     this->setConfiguration();
     connect(ui->output_sink_combo_box, SIGNAL(currentIndexChanged(int)), this,
             SLOT(updateAudioDevices(int)));
+    connect(ui->volume_slider, SIGNAL(valueChanged(int)), this,
+        SLOT(updateVolume(int)));
 }
 
-ConfigureAudio::~ConfigureAudio() {}
+float initial_volume = Settings::values.volume; //Save initial volume to restore it if user clicks Cancel.
```
wwylele: Make this as a member variable of `ConfigureAudio`. Also I can't find where this is used? I guess this should be used in a "Cancel" function?

src/audio_core/hle/mixers.cpp
```diff
@@ -49,24 +51,28 @@ void Mixers::ParseConfig(DspConfiguration& config) {
         LOG_TRACE(Audio_DSP, "mixers mixer2_enabled = %hu", config.mixer2_enabled);
     }
 
-    if (config.volume_0_dirty) {
+    bool volumeChanged = current_volume != Settings::values.volume && config.volume[0] != 0; //TODO: move to proper section
+    if (config.volume_0_dirty || volumeChanged) {
         config.volume_0_dirty.Assign(0);
-        state.intermediate_mixer_volume[0] = config.volume[0];
+        state.intermediate_mixer_volume[0] = config.volume[0] * Settings::values.volume; //TODO: move to proper section
```
wwylele: Please be more specific on what to move. In the context of this PR I can tell it is the volume multiplier, but later when people read this code again won't get what this TODO is about. Leaving your name here, like `// TODO (RedTrainer97):...` is also prefered, so that people will know who to ask when they get confused.

Moreover, I think what MerryMage suggested about "proper section" is pretty trivial to do now. I don't like "not knowing the whole code" as an excuse to throw code wherever you want and to leave fixing works to others - it is more likely that this TODO code will be there for a long time and people won't remember it any more. It would be better for you to get more familiar with citra code base now, and to code it properly.

RedTrainer97: I've added the TODO just to underline that my commit is partial, I want to try to fix it.

src/citra_qt/configure_audio.cpp
```diff
@@ -53,7 +56,7 @@ void ConfigureAudio::setConfiguration() {
         }
     }
     ui->audio_device_combo_box->setCurrentIndex(new_device_index);
-    ui->volume_slider->setValue(Settings::values.volume);
+    ui->volume_slider->setValue(Settings::values.volume * 10);
```
JayFoxRox: As I said in my previous post: 10 steps is still not enough to represent the hardware slider.

src/citra_qt/configure_audio.cpp
```diff
@@ -64,7 +67,7 @@ void ConfigureAudio::applyConfiguration() {
     Settings::values.audio_device_id =
         ui->audio_device_combo_box->itemText(ui->audio_device_combo_box->currentIndex())
             .toStdString();
-    Settings::values.volume = ui->volume_slider->value();
+	initial_volume = (float)ui->volume_slider->value() / 10;
```
JayFoxRox: Seeing the use of 10 everywhere makes me thing this should be either a constant or be retrieved from the max value of the volume_slider (which is probably the better idea)

src/audio_core/hle/mixers.cpp
```diff
@@ -33,7 +33,7 @@ DspStatus Mixers::Tick(DspConfiguration& config, const IntermediateMixSamples& r
     return GetCurrentStatus();
 }
 
-int currentVolume = 5;
+float current_volume = 1; //TODO: move to proper section
```
JayFoxRox: We don't really want TODOs in our codebase unless they are tagged with a name who is responsible for that mess (and fixing thereof).
If this is just a temporary TODO for yourself you shouldn't have pushed yet.

src/citra_qt/configure_audio.cpp
RedTrainer97: I've changed the slider to 16-steps [0...15], which is a more "programming" number.
Do you think it's enough or should I increase?

src/citra_qt/configure_audio.ui
```diff
@@ -62,13 +62,13 @@
...
           <property name="maximum">
-           <number>10</number>
+           <number>15</number>
           </property>
           <property name="pageStep">
            <number>1</number>
           </property>
           <property name="value">
-           <number>5</number>
+           <number>15</number>
```
JayFoxRox: This *still* doesn't reflect the minimum amount of hardware states.

src/citra_qt/configure_audio.cpp
```diff
@@ -20,6 +20,7 @@ ConfigureAudio::ConfigureAudio(QWidget* parent)
         ui->output_sink_combo_box->addItem(sink_detail.id);
     }
 
+    ConfigureAudio::volume_slider_steps = ui->volume_slider->maximum();
```
JayFoxRox: I'd expect this the other way around.. = Use `ui->volume_slider->maximum()` directly to avoid using outdated values but set it once from a constant which defines how many steps there are (or just set the number of steps in the .ui file like you do now).

src/audio_core/hle/mixers.cpp
```diff
@@ -49,24 +50,28 @@ void Mixers::ParseConfig(DspConfiguration& config) {
...
-    if (config.volume_2_dirty) {
+    if (config.volume_2_dirty || volumeChanged) {
         config.volume_2_dirty.Assign(0);
-        state.intermediate_mixer_volume[2] = config.volume[2];
+        state.intermediate_mixer_volume[2] = config.volume[2] * Settings::values.volume; //TODO(RedTrainer97): move '* Settings::values.volume' to proper section
         LOG_TRACE(Audio_DSP, "mixers volume[2] = %f", config.volume[2]);
     }
 
+    if (volumeChanged) //TODO(RedTrainer97): move to proper section
+        Mixers::current_volume = Settings::values.volume; //TODO(RedTrainer97): move to proper section
```
JayFoxRox: Explicit scope? (Not sure if we have code guidelines about that but I personally don't want another `goto fail`)

src/audio_core/hle/mixers.cpp
```diff
@@ -49,24 +50,28 @@ void Mixers::ParseConfig(DspConfiguration& config) {
         LOG_TRACE(Audio_DSP, "mixers mixer2_enabled = %hu", config.mixer2_enabled);
     }
 
-    if (config.volume_0_dirty) {
+    bool volumeChanged = Mixers::current_volume != Settings::values.volume && config.volume[0] != 0; //TODO(RedTrainer97): move to proper section
+    if (config.volume_0_dirty || volumeChanged) {
```
JayFoxRox: Why do you keep working on this code bit?
I thought @MerryMage said it was the wrong place to do the volume adjustment. Unless the real 3DS does audio volume settings in the DSP we shouldn't do it here either.

src/citra_qt/config.cpp
```diff
@@ -231,7 +232,8 @@ void Config::SaveValues() {
     qt_config->setValue("output_engine", QString::fromStdString(Settings::values.sink_id));
     qt_config->setValue("enable_audio_stretching", Settings::values.enable_audio_stretching);
     qt_config->setValue("output_device", QString::fromStdString(Settings::values.audio_device_id));
-    qt_config->endGroup();
+	qt_config->setValue("volume", Settings::values.volume);
+	qt_config->endGroup();
```
JayFoxRox: Bad indentation?

src/citra_qt/configure_audio.cpp
```diff
@@ -75,3 +88,9 @@ void ConfigureAudio::updateAudioDevices(int sink_index) {
         ui->audio_device_combo_box->addItem(device.c_str());
     }
 }
+
+void ConfigureAudio::updateVolume(int volume)
+{
```
JayFoxRox: Other functions seem to have braces on same line, did you run / check clang-format?

src/citra_qt/configure_audio.ui
RedTrainer97: How much should I set it to? I set 15 because I think bigger numbers  would be less useful for humans.

JayFoxRox:
Please squash up once in a while - it gets annoying to comment on this. Whenever I comment there's new commits which replace the code already and sometimes github is just confused

jroweboy:
If i may intervene, it seems like you are editing the files through the github editor. I hope you are aware that you can edit the files locally and use git to push them to your fork so you can edit multiple files locally and push all the changes in a single commit.

RedTrainer97:
@jroweboy I'm editing the files on Visual Studio, but when I commit it has a different coding style, so I'm trying to push them manually to avoid this problem (sometime with bad luck...).

jroweboy:
@RedTrainer97 before you commit, you should use clang-format to make sure that the code matches citra code guidelines. see something like this https://marketplace.visualstudio.com/items?itemName=HansWennborg.ClangFormat to add the ability to clang format the code from within vs

RedTrainer97:
Sorry for double commit.

src/audio_core/hle/mixers.h
```diff
@@ -58,4 +58,4 @@ class Mixers final {
 };
 
 } // namespace HLE
-} // namespace DSP
+} // namespace DSP
```
JayFoxRox: Unintentional changes

src/citra_qt/configure_audio.cpp
```diff
@@ -20,12 +20,20 @@ ConfigureAudio::ConfigureAudio(QWidget* parent)
...
     connect(ui->output_sink_combo_box, SIGNAL(currentIndexChanged(int)), this,
             SLOT(updateAudioDevices(int)));
+    connect(ui->volume_slider, SIGNAL(valueChanged(int)), this,
+        SLOT(updateVolume(int)));
 }
 
-ConfigureAudio::~ConfigureAudio() {}
+ConfigureAudio::~ConfigureAudio() {
+    //Restore initial volume or apply change (value edited in ConfigureAudio::applyConfiguration()).
+    Settings::values.volume = ConfigureAudio::initial_volume;
```
JayFoxRox: This feels like a poor design choice. How do our other config dialogs do this?

src/audio_core/hle/mixers.h
RedTrainer97: It is, I leave it right now in order to avoid three commits at the same time

JayFoxRox: just squash finally.. this is a nightmare to review as is anyway.

src/citra_qt/configure_audio.cpp
RedTrainer97: I've added this because I want the volume to update immediately, so user has an immediate feedback and can set it more easily.

JayFoxRox: Which is probably a good thing, but I feel this is inconsistent. So I *personally* would rather have it working like the rest and then updated all at once for the others too (I often felt we should have more direct config feedback with the graphics options too).

RedTrainer97: So should I remove this?

src/audio_core/hle/dsp.cpp
```diff
@@ -11,6 +11,8 @@
 #include "audio_core/sink.h"
 #include "audio_core/time_stretch.h"
```
RedTrainer97: TODO(RedTrainer97): remove this line if the volume change is ok in this file.

src/citra_qt/configure_audio.ui
```diff
@@ -50,6 +50,33 @@
         </item>
        </layout>
       </item>
+      <item>
+       <layout class="QHBoxLayout" name="horizontalLayout">
+        <item>
+         <widget class="QLabel" name="label_2">
```
j-selby: "label_2"?

Be consistent with your naming scheme.

src/citra_qt/configure_audio.ui
```diff
@@ -50,6 +50,33 @@
         </item>
        </layout>
       </item>
+      <item>
+       <layout class="QHBoxLayout" name="horizontalLayout">
```
j-selby: "horizontalLayout"?

src/citra_qt/configure_audio.ui
j-selby: "label_2"?

Be consistent with your naming scheme. Name them something easy to identify.

src/citra/default_ini.h
```diff
@@ -123,6 +123,10 @@ enable_audio_stretching =
 # auto (default): Auto-select
 output_device =
 
+# Setup the volume
```
j-selby: "Setup the volume"? 
I would suggest "Output volume".

Setup means nothing. Does this mean only at the start, or throughout?

j-selby:
I disagree with the contents of this PR for the reasons that others have mentioned - you are duplicating functionality against the OS. wwylele's suggestion about implementing HLE GetSoundVolume would be a better fit.

I would suggest a simple toggle box in the top toolbar to enable/disable. Leave the rest to the OS.

RedTrainer97:
You are right that I'm duplicating an OS function, but 3DS has a volume slider, so even the emulator should have it. And if the user won't use it, he/she can leave it to 100% as default and change it on OS.
The HLE GetSoundVolume function seems to be never reached, so I preferred to work on Audio output.
For the inconsistent naming in .ui, there are already generic "label"s and unnamed QHBoxLayouts, but I'll fix my code in next commit.

src/citra_qt/configure_audio.ui
```diff
@@ -50,6 +50,33 @@
...
+         <widget class="QLabel" name="label_2">
+          <property name="text">
+           <string>Volume:</string>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <widget class="QSlider" name="volume_slider">
+          <property name="maximum">
+           <number>15</number>
```
JayFoxRox: This is still too small. It has to be at very least 64. I'd recommend setting it to 1000 or even higher so we have fine control and don't have to worry about it again

wwylele:
> The HLE GetSoundVolume function seems to be never reached

fyi the system app 3DS Sound uses this function to show a volume icon on the screen. Not saying because of this you should implement this here immediately, just letting you know.

RedTrainer97:
Added Mute checkbox.

JayFoxRox:
We've had at least 4 people (me included) saying that this is kind of stretching the bounds of what should be Citras responsibility. While the slider is present on an actual 3DS, a mute button isn't.

Overall it was pointed out that the design is bad as it conflicts with how other controls within the config menu work. Also see #2642 to resolve this.

I feel we should put the input slider in the status bar which is also where you'd expect it on media players (bottom of the screen). This way it also doesn't conflict with the rest of the configs.
Even if you insist on putting it in the config window (which I completely disagree with) it still shouldn't be on the audio panel but in something related to the 3DS or even the input panel. It's not related to the audio backend of the emulator but a feature of the 3DS.

IMHO: If you feel like adding a mute button, please do so in a fork.

RedTrainer97:
@JayFoxRox Maybe I can move the volume slider there (that is more or less where the volume slider is on the 3DS).
immagine

JayFoxRox:
I think these changes are premature and there is still an ongoing discussion / debate in #2642 
I'm completly opposed to the current control placement.

I think you should focus on the technical implementation (HLE, hotkey and audio backend integration) and configure it using the config for now, we can then add a UI at a later point once these design issues have been resolved.
- There is no word what happens if we split windows for screens in the future (which has been requested numerous times), what happens in fullscreen mod, how does it look like on other platforms or with a non-standard background color (e.g. black or gray), how does it look like if the window is squished, how does it interact with other widgets such as the debugger, ...

I feel this change is basically a hotfix for a very case-specific problem of very few users, and it's a poor one at that. Nobody prevents you from using such an (ugly) GUI in a fork for now.

MerryMage:
(You're trying to do too many things at once, and quite radically changing your PR as you do. Please break any changes down. Future PRs can be submitted as required.)

bunnei:
Closing this as it's clear it's not going to be merged as-is. I agree with the others that the UI design is not well though-out and (if anything) these controls should be buried in the config UI.

Feel free to address the above feedback (specifically, correct and separate HLE implementations of these features, set by the config files, not UI) and resubmit it as followup PRs.

PR #2638, opened 2017-03-24
updated the implementation of GetAppletManInfo to include the latest information from 3dbrew

jfmherokiller:


jfmherokiller:
closed because it conflicts with #2533

PR #2644, opened 2017-03-29
refactored the ncch class into 2 classes

jfmherokiller:
the original implementation of the ncch class contained the logic for both cci files and cxi files.

this refactoring moves the logic for the cci files to its own subclass for both readability and possibly maintainability. however it does not modify the original intended function of the code. This is shown in the following screenshot
screen shot 2017-03-28 at 2 21 33 pm

if this happens to get merged i will rework my cia loader to use it as well

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2644)

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -175,6 +175,7 @@ set(SRCS
             loader/elf.cpp
             loader/loader.cpp
             loader/ncch.cpp
+            loader/ncsd.cpp
             loader/smdh.cpp
             tracer/recorder.cpp
             memory.cpp
@@ -373,6 +374,7 @@ set(HEADERS
             loader/elf.h
             loader/loader.h
             loader/ncch.h
+            loader/ncsd.h
             loader/smdh.h
             tracer/recorder.h
             tracer/citrace.h
--- a/src/core/loader/loader.cpp
+++ b/src/core/loader/loader.cpp
@@ -10,6 +10,7 @@
 #include "core/loader/3dsx.h"
 #include "core/loader/elf.h"
 #include "core/loader/ncch.h"
+#include "core/loader/ncsd.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
@@ -32,6 +33,7 @@ FileType IdentifyFile(FileUtil::IOFile& file) {
     CHECK_TYPE(THREEDSX)
     CHECK_TYPE(ELF)
     CHECK_TYPE(NCCH)
+    CHECK_TYPE(NCSD)
 
 #undef CHECK_TYPE
 
@@ -110,10 +112,13 @@ static std::unique_ptr<AppLoader> GetFileLoader(FileUtil::IOFile&& file, FileTyp
     case FileType::ELF:
         return std::make_unique<AppLoader_ELF>(std::move(file), filename);
 
-    // NCCH/NCSD container formats.
+    // NCCH container format.
     case FileType::CXI:
-    case FileType::CCI:
         return std::make_unique<AppLoader_NCCH>(std::move(file), filepath);
+
+    // NCSD container format.
+    case FileType::CCI:
+        return std::make_unique<AppLoader_NCSD>(std::move(file), filepath);
 
     default:
         return nullptr;
--- a/src/core/loader/ncch.h
+++ b/src/core/loader/ncch.h
@@ -161,9 +161,10 @@ namespace Loader {
 /// Loads an NCCH file (e.g. from a CCI, or the first NCCH in a CXI)
 class AppLoader_NCCH final : public AppLoader {
 public:
+    AppLoader_NCCH(FileUtil::IOFile&& file, const std::string& filepath, u32 ncch_offset)
+        : AppLoader(std::move(file)), filepath(filepath), ncch_offset(ncch_offset) {}
     AppLoader_NCCH(FileUtil::IOFile&& file, const std::string& filepath)
-        : AppLoader(std::move(file)), filepath(filepath) {}
-
+        : AppLoader(std::move(file)), filepath(filepath), ncch_offset(0) {}
     /**
      * Returns the type of the file
      * @param file FileUtil::IOFile open file
@@ -218,16 +219,16 @@ private:
     ResultStatus LoadSectionExeFS(const char* name, std::vector<u8>& buffer);
 
     /**
-     * Loads .code section into memory for booting
-     * @return ResultStatus result of function
-     */
-    ResultStatus LoadExec();
-
-    /**
      * Ensure ExeFS is loaded and ready for reading sections
      * @return ResultStatus result of function
      */
     ResultStatus LoadExeFS();
+
+    /**
+     * Loads .code section into memory for booting
+     * @return ResultStatus result of function
+     */
+    ResultStatus LoadExec();
 
     /// Reads the region lockout info in the SMDH and send it to CFG service
     void ParseRegionLockoutInfo();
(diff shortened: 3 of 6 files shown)


src/core/loader/ncsd.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include <common/logging/log.h>
+#include "core/loader/ncsd.h"
+namespace Loader {
+FileType AppLoader_NCSD::IdentifyType(FileUtil::IOFile& file) {
+    u32 magic;
+    file.Seek(0x100, SEEK_SET);
+    if (file.ReadArray<u32>(&magic, 1) != 1)
+        return FileType::Error;
+
+    if (MakeMagic('N', 'C', 'S', 'D') == magic)
```
JayFoxRox: the one below is not yoda. Pick one.
(I prefer non yoda and I consider `MakeMagic` to be a constant generator)

src/core/loader/ncch.h
```diff
@@ -242,14 +236,20 @@ class AppLoader_NCCH final : public AppLoader {
...
     u8 priority = 0;
     u8 resource_limit_category = 0;
-    u32 ncch_offset = 0; // Offset to NCCH header, can be 0 or after NCSD header
     u32 exefs_offset = 0;
 
-    NCCH_Header ncch_header;
     ExeFs_Header exefs_header;
     ExHeader_Header exheader_header;
 
+protected:
```
lioncash: protected members should come before private members.

src/core/loader/ncsd.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
+ResultStatus AppLoader_NCSD::LoadExeFS() {
+    // Reset read pointer in case this file has been read before.
+    file.Seek(0, SEEK_SET);
+
+    if (file.ReadBytes(&ncch_header, sizeof(NCCH_Header)) != sizeof(NCCH_Header))
+        return ResultStatus::Error;
+
+    // Skip NCSD header and load first NCCH (NCSD is just a container of NCCH files)...
+    if (ncch_header.magic == MakeMagic('N', 'C', 'S', 'D')) {
```
wwylele: I know the `ncch_header` comes from the old code, but that's a shitty workaround for NCCH and NCSD being handled together. Now since the NCSD loader is separated out, you should make and `NCSD_Header` for its own structure.

src/core/CMakeLists.txt
```diff
@@ -174,6 +174,7 @@ set(SRCS
             memory.cpp
             perf_stats.cpp
             settings.cpp
+            loader/ncsd.cpp
```
wwylele: place file name in alphabetical order

src/core/loader/ncsd.cpp
jfmherokiller: completed however i wish i knew of a better way to handle the media units so i could remove the 0x200 part from this `ncch_offset = ncsd_header.partition_table[0].partition_offset * 0x200;`

src/core/loader/ncsd.cpp
```diff
@@ -0,0 +1,35 @@
+// Copyright 2014 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <common/logging/log.h>
+#include "core/loader/ncsd.h"
```
jfmherokiller: fixed in latest version

src/core/loader/ncch.h
```diff
@@ -208,6 +208,16 @@ class AppLoader_NCCH final : public AppLoader {
     ResultStatus ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_file, u64& offset,
                            u64& size) override;
 
+protected:
+    std::string filepath;
```
wwylele: I can't find where this is used in NCSD loader? why is this moved to `protected`?

src/core/loader/ncsd.cpp
```diff
@@ -0,0 +1,31 @@
...
+    file.Seek(0x100, SEEK_SET);
+    if (file.ReadArray<u32>(&magic, 1) != 1)
+        return FileType::Error;
+
+    if (magic == MakeMagic('N', 'C', 'S', 'D'))
+        return FileType::CCI;
+
+    return FileType::Error;
+}
+ResultStatus AppLoader_NCSD::LoadExeFS() {
```
wwylele: in NCCH loader `LoadExeFS` can be called several times when loading (it becomes no-op if ExeFS is already loaded). So if it get override by this, although it doesn't hurt, the NCSD header will be read several times, which looks redundant. 

A probably better approach is that NCCH Loader exposes a virtual function `TryGetNCCHOffset`, whose default implementation is returning 0, and which gets override by NCSD loader to return the actual offset. Thus you don't need to expose `LoadExeFS` and `ncch_offset`, which should be only for internal use. I still find this approach not a really good design... you probably want to listen to others about the overall design of this class and inheritance.

src/core/loader/ncch.h
jfmherokiller: it not used exactly but as far as i could tell it was needed to be able to correctly perform the class inheritance setup

jfmherokiller: it was during the time i spent trying to work out how to correctly perform class inheritance

src/core/loader/ncch.h
```diff
@@ -218,16 +227,16 @@ class AppLoader_NCCH final : public AppLoader {
...
-    ResultStatus LoadExec();
+    virtual ResultStatus LoadExeFS();
 
     /**
-     * Ensure ExeFS is loaded and ready for reading sections
+     * Loads .code section into memory for booting
      * @return ResultStatus result of function
      */
-    ResultStatus LoadExeFS();
+    ResultStatus LoadExec();
```
wwylele: ^ garbage diff. also `LoadExeFS` doesn't need to be `virtual` now

src/core/loader/ncch.h
```diff
@@ -242,14 +251,12 @@ class AppLoader_NCCH final : public AppLoader {
     u32 core_version = 0;
     u8 priority = 0;
     u8 resource_limit_category = 0;
-    u32 ncch_offset = 0; // Offset to NCCH header, can be 0 or after NCSD header
+    std::string filepath;
```
wwylele: ^ garbage diff

src/core/loader/ncch.h
```diff
@@ -208,6 +208,15 @@ class AppLoader_NCCH final : public AppLoader {
     ResultStatus ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_file, u64& offset,
                            u64& size) override;
 
+protected:
+    /**
+     * get the offset of the ncch header
+     * @return offset of ncch header
+     */
+    virtual u32 TryGetNCCHOffset() {
```
wwylele: I explicitly said `Try` in my suggestion because this function can fail in the case of an invalid file. Thus, the return type should be a nullable object or something with `ResultStatus`

src/core/loader/loader.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2014 Citra Emulator Project
+// Copyright 2017 Citra Emulator Project
```
yuriks: Don't change copyright dates unless you're significantly re-writing the file.

src/core/loader/ncsd.h
```diff
@@ -0,0 +1,52 @@
+
```
yuriks: Leading blank line

src/core/loader/ncsd.h
```diff
@@ -0,0 +1,52 @@
...
+    std::array<u8, 8>
+        partitions_fs_type; //< Partitions FS type (0=None, 1=Normal, 3=FIRM, 4=AGB_FIRM save)
+    std::array<u8, 8> partitions_crypt_type; //< Partitions crypt type (each byte corresponds to a
+                                             // partition in the partition table)
+    std::array<Partition_Table_Entry, 8> partition_table;
+};
+
+namespace Loader {
+
+class AppLoader_NCSD final : public AppLoader_NCCH {
```
yuriks: It's not really appropriate to use inheritance like this. Rather use composition and delegate functions to to a wrapper NCCH loader instead.

jfmherokiller: thanks for the design insight i mainly went with the inheritance method because i thought it would be much easier

jfmherokiller:
i had to do the wrapper class refactor as a seporate commit because there are some bits im not sure of

B3n30:
What's the status of this PR?

B3n30:
Closed, due to inactivity.

PR #2645, opened 2017-03-30
Fix various warnings

marcoc2:


src/core/hle/service/y2r_u.cpp
```diff
@@ -110,7 +110,8 @@ static void SetOutputFormat(Interface* self) {
     cmd_buff[0] = IPC::MakeHeader(0x3, 1, 0);
     cmd_buff[1] = RESULT_SUCCESS.raw;
 
-    LOG_DEBUG(Service_Y2R, "called output_format=%hhu", conversion.output_format);
+    LOG_DEBUG(Service_Y2R, "called output_format=%hhu",
+              static_cast<int>(conversion.output_format));
```
mailwl: u32?

src/video_core/swrasterizer/rasterizer.cpp
```diff
@@ -309,8 +309,10 @@ static void ProcessTriangleInternal(const Vertex& v0, const Vertex& v1, const Ve
                     (texture.config.wrap_t == TexturingRegs::TextureConfig::ClampToBorder &&
                      (t < 0 || static_cast<u32>(t) >= texture.config.height))) {
                     auto border_color = texture.config.border_color;
-                    texture_color[i] = {border_color.r, border_color.g, border_color.b,
-                                        border_color.a};
+                    texture_color[i] = {static_cast<u8>(border_color.r),
+                                        static_cast<u8>(border_color.r),
+                                        static_cast<u8>(border_color.r),
+                                        static_cast<u8>(border_color.r)};
```
Dragios:
```
texture_color[i] = {static_cast<u8>(border_color.r),
                   static_cast<u8>(border_color.g),
                   static_cast<u8>(border_color.b),
                   static_cast<u8>(border_color.a)};
```

src/core/hle/kernel/process.cpp
```diff
@@ -90,6 +90,9 @@ void Process::ParseKernelCaps(const u32* kernel_caps, size_t len) {
             mapping.size = Memory::PAGE_SIZE;
             mapping.writable = true; // TODO: Not sure if correct
             mapping.unk_flag = false;
+
+            // Missing line??
```
Dragios: What is this comment for? Add a review comment in this PR instead of commenting inside the code.

marcoc2: "mapping" is created inside this if statement but it is not used

src/video_core/swrasterizer/rasterizer.cpp
marcoc2: Yeah, paste and copy mistaken

src/core/core.cpp
```diff
@@ -82,14 +82,15 @@ System::ResultStatus System::Load(EmuWindow* emu_window, const std::string& file
 
     ResultStatus init_result{Init(emu_window, system_mode.get())};
     if (init_result != ResultStatus::Success) {
-        LOG_CRITICAL(Core, "Failed to initialize system (Error %i)!", init_result);
+        LOG_CRITICAL(Core, "Failed to initialize system (Error %i)!",
+                     static_cast<int>(init_result));
```
JayFoxRox: In general re LOG fixes: @jroweboy is working on a log rewrite which is probably better than hotfixing these issues.

src/core/hle/kernel/process.cpp
JayFoxRox: >Subv: JayFoxRox: that's the IO mappings, which are not implemented yet
JayFoxRox: shouldn't there be a comment to document this then Subv?
***Subv shrugd

JayFoxRox:
In general re LOG fixes: @jroweboy is working on a log rewrite which is probably better than hotfixing these issues.

> jroweboy: i'm working on a spdlog branch actually :) all thats left is actually regex replacing all the existing logs ...

---

I'll also remind about the IRC discussion we had yesterday @marcoc2 : 

> JayFoxRox: also don't use casts everywhere for printf. I'm not sure if we decided to use it, but use inttypes printf strings? yuriks , Lioncash et al?
Lioncash: static_cast is necessary for enum classes, but yeah, some of those other casts can be made unnecessary with the correct printf specifier.

> JayFoxRox: Lioncash: do we use PRIX16 etc? because I don't remember seeing them anywhere
Lioncash: JayFoxRox: We mostly use the 64-bit ones (PRId64, PRIu64, PRIx64) afaik.

> JayFoxRox: should we also use the other ones? also is there a guidelines for hex (upper or lowercase? - prefixed with "0x"?) I feel this is one of those things which varies a lot over the codebase
JayFoxRox: (and even the frontend / debug messages)
Lioncash: imo we should just use something like spdlog, which uses fmtlib internally which is type agnostic and doesn't need explicit datatype specifiers (and is extensible for custom types). So you can just do fmt::format("{}{}{}", "print my crap", 1, 2.3); or fmt::format("{0}{1}{0}", "Positional", "args"); :p
Lioncash: makes dealing with all of that way less of a pain.
neobrain: ^
yuriks: JayFoxRox: the reason we don't have many is because using PRI* is a pain in the ass xD
yuriks: and yeah, agreed with using something fmt-based
***yuriks had plans to do that but lost motivation
yuriks: I was just gonna replace the formatting in our existing log stuff though

So overall the impression I got was to only fix the non logging related things and do a proper log rewrite like @jroweboy is doing now. If you take time to hotfix the LOG code I'd *personally* use inttypes when using fixed width types.

marcoc2:
Ok. I tried to fix many warnings as possible because

marcoc2:
Ok, I'll revert casts for the log. I was just trying to take away many warnings as possible.

JayFoxRox:
Please clang-format?

JayFoxRox:
@marcoc2 do you still intend to work on this? It's been a week. Not pushing, just asking.

wwylele:
Closing this because of inactivity

PR #2650, opened 2017-04-03
added comments the applet ids and added missing app ids

jfmherokiller:
i updated the applet ids in the apt.h file so they are equal to the ones on [this](https://3dbrew.org/wiki/NS_and_APT_Services) page. 
I also added a doxygen comment next to each applet id containing the info from the description column if the applet id did not already have a comment. 

The developer ids were added because without them i felt the 3ds software wasn't fully documented.
this is a resubmission of pull request #2511


JayFoxRox:
What happened to "Code is documentation" ?

I feel most of these enums are already very descriptive. Just copy/pasting possibly creates ambiguity.
If anything I'd just link to the respective 3dbrew page and say: "For explanations see: http..."; but I feel it's implied that people know where to look up this kind of stuff - it's easy to find probably.

MerryMage:
Previous PR: #2511

jfmherokiller:
@JayFoxRox i will remove most of the stuff in parentheses but i wish to keep some of the names because they correspond to the process name

src/core/hle/service/apt/apt.h
```diff
@@ -70,35 +70,46 @@ enum class SignalType : u32 {
...
+    GameNotes = 0x113,         ///< Game Notes applet (Cherry)
+    InternetBrowser = 0x114,   ///< Internet Browser (spider/SKATER)
+    InstructionManual = 0x115, ///< Instruction Manual applet
+    Notifications = 0x116,     ///< Notifications applet (newslist)
+    Miiverse = 0x117,          ///< Miiverse applet (olv)
+    MiiversePost = 0x118,      ///< Miiverse posting applet (solv3)
+    AmiiboSettings = 0x119,    ///< Amiibo settings (cabinet)
+    SoftwareKeyboard1 = 0x201, ///< Software Keyboard (swkbd)
+    Ed1 = 0x202,               ///< Mii Selector (appletEd)
+    PnoteApp = 0x204,          ///< Photo Selector (PNOTE_AP)
```
JayFoxRox: This is already described by the name.
So ideally we'd avoid any sort of redundancy and the comments would just add information.

Also a lot of our enums are named by descriptive names, others are called by nintendo codenames / applet names? (what are those in braces?)

I think we should make our enums more descriptive and only include the codename as the comment.
Say.. `PhotoSelector = 0x204, ///< PNOTE_AP`

Maybe also the other way around, I'm not sure where / how they are used, maybe @wwylele can shed some light on it.

---

Overall I feel this is a very unimportant issue and we shouldn't waste time with this sort of thing.
There are more important issues we have to fix.

jfmherokiller: i think ill rework it so that the code names are in the comment and the actual process name is the enum

jfmherokiller:
im going to close this pull request so i can put more effort into it.

PR #2655, opened 2017-04-07
basic translation support

jfmherokiller:
this adds basic translation support to citra however the LinguistTools seem to be unable to be found by travis and appveyor. 

i can confirm that when it does find it the tools it builds the language files correctly.


wwylele:
> however the LinguistTools seem to be unable to be found by travis and appveyor.

Because they are not installed by default. You need add installation to the script.

----

Some previous discussion on Discord:

> (context: complained about shitty qt linguistic tools)
wwylele: I think we need a separate repo dealing with translation scanning/scripting/dirty_hack/etc + translation workflow. and just update the translated patch file (called .qm iirc?) to the main repo periodically
Noah: not a bad idea
JayFoxRox:imo that's a bad idea
we should instead fix the upstream or find a better solution
we can't depend on shitty and broken software
wwylele: well, yeah... _ideally_ we should fix the upstream
...
wwylele:  reminds me another reason I proposed to separate translation into another repo: no matter whether LinguistTools is good or bad, it is a dependency. For main repo CI and users that only want to compile citra, they should be able to use published translation file and don't need this dependency. And I don't want to see "update fr and jp translation" PR mixed with regular PR

src/citra_qt/configuration/configure_general.cpp
```diff
@@ -15,6 +17,21 @@ ConfigureGeneral::ConfigureGeneral(QWidget* parent)
...
+    QStringList fileNames = dir.entryList(QStringList("*.qm"));
+
+    for (int i = 0; i < fileNames.size(); ++i) {
+        // get locale extracted by filename
+        QString locale;
+        locale = fileNames[i];                    // "de.qm"
+        locale.truncate(locale.lastIndexOf('.')); // "de"
+        QString lang = locale.right(2);
+        ui->language_combobox->addItem(lang, locale);
+    }
```
JayFoxRox: Just saying: This was naively stolen (by me) from https://wiki.qt.io/How_to_create_a_multi_language_application

JayFoxRox:
I suggest making the Linguist Tools optional.

Also: This PR was clearly rushed.. again. [You should have asked me what else might be necessary, checked my todo list](https://github.com/JayFoxRox/citra/pull/39).

Now we are discussing design-issues again: Do we expect linguist tools? Do we ship Citra with translations? Are they in a seperate repo? No issue / TODO for the hack in the cmake to work around Qt bugs etc.?
I expected some of these things to be common sense that I did not even mention them in my branch. If the code was fine already I'd have PR'd it myself.

jfmherokiller:
closed because i meant to open this pull request on my own repo instead of the main repo and passed out before i could close it last night

PR #2659, opened 2017-04-08
dsp_dsp: Messages are modified by service before being sent to DSP

MerryMage:
Confirmed by RE.

Code has been a part of BE for a few months.


wwylele:
These message modification looks pretty random to me. Could you explain what they are actually? Or at least add a comment "This is what DSP service does confirmed by RE. The actual meaning is not clear" so that others will know what's going on there when reading the code.

MerryMage:
Will leave a comment.

I can guess why the modification happens though: Games pass in garbage at these bytes because they read random bytes off the stack (passing in a short then declaring to send 4 bytes, for example), so they fixed it here.

PR #2661, opened 2017-04-12
Services/UDS: Generate 802.11 beacon frames when a network is open.

Subv:
These beacon frames are verified to be correct by comparing them to a real wifi pcap capture of a Smash Bros 4 on a real 3DS.

The NWM key is needed for these to be exactly the same as on a real 3DS. For Citra's purposes, an all-zero key is enough.

We use the 802.11 beacon frame structure because that's what games expect to receive when calling ReceiveBroadcastBeaconData

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2661)

### DIFF
--- a/src/core/hle/service/nwm/nwm_uds.cpp
+++ b/src/core/hle/service/nwm/nwm_uds.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <array>
 #include <cstring>
 #include <unordered_map>
 #include <vector>
@@ -12,6 +13,7 @@
 #include "core/hle/kernel/shared_memory.h"
 #include "core/hle/result.h"
 #include "core/hle/service/nwm/nwm_uds.h"
+#include "core/hle/service/nwm/uds_beacon.h"
 #include "core/memory.h"
 
 namespace Service {
@@ -27,10 +29,12 @@ static Kernel::SharedPtr<Kernel::SharedMemory> recv_buffer_memory;
 // Connection status of this 3DS.
 static ConnectionStatus connection_status{};
 
-// Node information about the current 3DS.
-// TODO(Subv): Keep an array of all nodes connected to the network,
-// that data has to be retransmitted in every beacon frame.
-static NodeInfo node_info;
+/* Node information about the current network.
+ * The amount of elements in this vector is always the maximum number
+ * of nodes specified in the network configuration.
+ * The first node is always the host, so this always contains at least 1 entry.
+ */
+static NodeList node_info(1);
 
 // Mapping of bind node ids to their respective events.
 static std::unordered_map<u32, Kernel::SharedPtr<Kernel::Event>> bind_node_events;
@@ -127,7 +131,7 @@ static void InitializeWithVersion(Interface* self) {
     u32 sharedmem_size = rp.Pop<u32>();
 
     // Update the node information with the data the game gave us.
-    rp.PopRaw(node_info);
+    rp.PopRaw(node_info[0]);
 
     u16 version;
     rp.PopRaw(version);
@@ -251,13 +255,25 @@ static void BeginHostingNetwork(Interface* self) {
     ASSERT_MSG(network_info.max_nodes > 1, "Trying to host a network of only one member.");
 
     connection_status.status = static_cast<u32>(NetworkStatus::ConnectedAsHost);
+
+    // Ensure the application data size is less than the maximum value.
+    ASSERT_MSG(network_info.application_data_size <= ApplicationDataSize, "Data size is too big.");
+
+    // Set up basic information for this network.
+    network_info.oui_value = NintendoOUI;
+    network_info.oui_type = static_cast<u8>(NintendoTagId::NetworkInfo);
+
     connection_status.max_nodes = network_info.max_nodes;
+
+    // Resize the nodes list to hold max_nodes.
+    node_info.resize(network_info.max_nodes);
 
     // There's currently only one node in the network (the host).
     connection_status.total_nodes = 1;
+    network_info.total_nodes = 1;
     // The host is always the first node
     connection_status.network_node_id = 1;
-    node_info.network_node_id = 1;
+    node_info[0].network_node_id = 1;
     // Set the bit 0 in the nodes bitmask to indicate that node 1 is already taken.
     connection_status.node_bitmask |= 1;
 
@@ -373,7 +389,8 @@ static void BeaconBroadcastCallback(u64 userdata, int cycles_late) {
     if (connection_status.status != static_cast<u32>(NetworkStatus::ConnectedAsHost))
         return;
 
-    // TODO(Subv): Actually generate the beacon and send it.
+    // TODO(Subv): Actually send the beacon.
+    std::vector<u8> frame = GenerateBeaconFrame(network_info, node_info);
 
     // Start broadcasting the network, send a beacon frame every 102.4ms.
     CoreTiming::ScheduleEvent(msToCycles(DefaultBeaconInterval * MillisecondsPerTU) - cycles_late,
--- a/src/core/hle/service/nwm/nwm_uds.cpp
+++ b/src/core/hle/service/nwm/nwm_uds.cpp
@@ -86,29 +86,79 @@ static void Shutdown(Interface* self) {
  *      1 : Result of function, 0 on success, otherwise error code
  */
 static void RecvBeaconBroadcastData(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 out_buffer_size = cmd_buff[1];
-    u32 unk1 = cmd_buff[2];
-    u32 unk2 = cmd_buff[3];
-    u32 mac_address = cmd_buff[4];
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x0F, 16, 4);
 
-    u32 unk3 = cmd_buff[6];
+    u32 out_buffer_size = rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
 
-    u32 wlan_comm_id = cmd_buff[15];
-    u32 ctr_gen_id = cmd_buff[16];
-    u32 value = cmd_buff[17];
-    u32 input_handle = cmd_buff[18];
-    u32 new_buffer_size = cmd_buff[19];
-    u32 out_buffer_ptr = cmd_buff[20];
+    MacAddress mac_address;
+    rp.PopRaw(mac_address);
 
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    // TODO(Subv): Use IPC::RequestParser::Skip when possible.
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
 
-    LOG_WARNING(Service_NWM,
-                "(STUBBED) called out_buffer_size=0x%08X, unk1=0x%08X, unk2=0x%08X,"
-                "mac_address=0x%08X, unk3=0x%08X, wlan_comm_id=0x%08X, ctr_gen_id=0x%08X,"
-                "value=%u, input_handle=0x%08X, new_buffer_size=0x%08X, out_buffer_ptr=0x%08X",
-                out_buffer_size, unk1, unk2, mac_address, unk3, wlan_comm_id, ctr_gen_id, value,
-                input_handle, new_buffer_size, out_buffer_ptr);
+    u32 wlan_comm_id = rp.Pop<u32>();
+    u32 id = rp.Pop<u32>();
+    Kernel::Handle input_handle = rp.PopHandle();
+
+    size_t desc_size;
+    const VAddr out_buffer_ptr = rp.PopMappedBuffer(&desc_size);
+    ASSERT(desc_size == out_buffer_size);
+
+    VAddr current_buffer_pos = out_buffer_ptr;
+    u32 total_size = sizeof(BeaconDataReplyHeader);
+
+    // Retrieve all beacon frames that were received from the desired mac address.
+    std::deque<WifiPacket> beacons =
+        GetReceivedPackets(WifiPacket::PacketType::Beacon, mac_address);
+
+    BeaconDataReplyHeader data_reply_header{};
+    data_reply_header.total_entries = beacons.size();
+    data_reply_header.max_output_size = out_buffer_size;
+
+    Memory::WriteBlock(current_buffer_pos, &data_reply_header, sizeof(BeaconDataReplyHeader));
+    current_buffer_pos += sizeof(BeaconDataReplyHeader);
+
+    // Write each of the received beacons into the buffer
+    for (const auto& beacon : beacons) {
+        BeaconEntryHeader entry{};
+        // TODO(Subv): Figure out what this size is used for.
+        entry.unk_size = sizeof(BeaconEntryHeader) + beacon.data.size();
+        entry.total_size = sizeof(BeaconEntryHeader) + beacon.data.size();
+        entry.wifi_channel = beacon.channel;
+        entry.header_size = sizeof(BeaconEntryHeader);
+        entry.mac_address = beacon.transmitter_address;
+
+        ASSERT(current_buffer_pos < out_buffer_ptr + out_buffer_size);
+
+        Memory::WriteBlock(current_buffer_pos, &entry, sizeof(BeaconEntryHeader));
+        current_buffer_pos += sizeof(BeaconEntryHeader);
+
+        Memory::WriteBlock(current_buffer_pos, beacon.data.data(), beacon.data.size());
+        current_buffer_pos += beacon.data.size();
+
+        total_size += sizeof(BeaconEntryHeader) + beacon.data.size();
+    }
+
+    // Update the total size in the structure and write it to the buffer again.
+    data_reply_header.total_size = total_size;
+    Memory::WriteBlock(out_buffer_ptr, &data_reply_header, sizeof(BeaconDataReplyHeader));
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+    rb.Push(RESULT_SUCCESS);
+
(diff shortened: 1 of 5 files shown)


src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,173 @@
...
+struct BeaconDataReplyHeader {
+    u32_le max_output_size;
+    u32_le total_size;
+    u32_le total_entries;
+};
+
+static_assert(sizeof(BeaconDataReplyHeader) == 12, "BeaconDataReplyHeader has incorrect size.");
+
+#pragma pack(push, 1)
+struct BeaconFrameHeader {
```
Subv: I'm not a fan of packing structures, but we need these to have a specific size, and the padding messes that up.

src/core/hle/service/nwm/uds_beacon.cpp
```diff
@@ -0,0 +1,314 @@
...
+std::vector<u8> GenerateNintendoSecondEncryptedDataTag(const NetworkInfo& network_info, const NodeList& nodes) {
+    // This tag is only present if the payload is larger than EncryptedDataSizeCutoff (0xFA).
+    if (nodes.size() * sizeof(NodeInfo) <= EncryptedDataSizeCutoff)
+        return {};
+
+    const size_t payload_size = nodes.size() * sizeof(NodeInfo) - EncryptedDataSizeCutoff;
+
+    const size_t tag_length = sizeof(EncryptedDataTag) - sizeof(TagHeader) + payload_size;
+
+    ASSERT_MSG(tag_length <= 255, "Data is too big.");
```
Subv: This tag length is only 8 bits wide, how do we handle this? The uds homebrew demo uses a very large amount of nodes (13), and it doesn't fit here.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -82,29 +86,78 @@ static void Shutdown(Interface* self) {
...
+    // TODO(Subv): Use IPC::RequestParser::Skip when possible.
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
+    rp.Pop<u32>();
```
JayFoxRox: Why isn't `IPC::RequestParser::Skip` used here?

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -251,13 +304,25 @@ static void BeginHostingNetwork(Interface* self) {
...
+    network_info.oui_type = static_cast<u8>(NintendoTagId::NetworkInfo);
+
     connection_status.max_nodes = network_info.max_nodes;
 
+    // Resize the nodes list to hold max_nodes.
+    node_info.resize(network_info.max_nodes);
+
     // There's currently only one node in the network (the host).
     connection_status.total_nodes = 1;
+    network_info.total_nodes = 1;
```
JayFoxRox: I kind of expected `node_info.size()` here?

src/core/hle/service/nwm/uds_beacon.cpp
JayFoxRox: How does a 3DS handle it?

src/core/hle/service/nwm/uds_beacon.h
JayFoxRox: How about functions to generate a buffer from / to this? (Serialize / Unserialize?)

src/core/hle/service/nwm/uds_beacon.cpp
Subv: I'm not sure, but i think it just truncates the data

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -82,29 +86,69 @@ static void Shutdown(Interface* self) {
...
+
+    BeaconDataReplyHeader data_reply_header{};
+    data_reply_header.total_entries = beacons.size();
+    data_reply_header.max_output_size = out_buffer_size;
+
+    Memory::WriteBlock(current_buffer_pos, &data_reply_header, sizeof(BeaconDataReplyHeader));
+    current_buffer_pos += sizeof(BeaconDataReplyHeader);
+
+    // Write each of the received beacons into the buffer
+    for (auto& beacon : beacons) {
```
lioncash: This can be a const reference as far as I can tell.

src/core/hle/service/nwm/uds_beacon.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+#include <cryptopp/aes.h>
+#include <cryptopp/md5.h>
+#include <cryptopp/modes.h>
+#include <cryptopp/sha.h>
+
+namespace Service {
+namespace NWM {
+
+static const u64 DefaultNetworkUptime = 900000000; // 15 minutes in microseconds.
```
lioncash: Numeric constants can just be constexpr without the `static` (const internally links by default).

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,172 @@
...
+ * @returns The generated frame.
+ */
+std::vector<u8> GenerateBeaconFrame(const NetworkInfo& network_info, const NodeList& nodes);
+
+/**
+ * Returns a list of received 802.11 frames from the specified sender
+ * matching the type since the last call.
+ */
+std::deque<WifiPacket> GetReceivedPackets(WifiPacket::PacketType type, const MacAddress& sender);
+}
```
lioncash:
```cpp
} // namespace NWM
} // namespace Service
```

src/core/hle/service/nwm/uds_beacon.cpp
```diff
@@ -0,0 +1,321 @@
...
+    buffer.insert(buffer.end(), basic_tags.begin(), basic_tags.end());
+    buffer.insert(buffer.end(), nintendo_tags.begin(), nintendo_tags.end());
+
+    return buffer;
+}
+
+std::deque<WifiPacket> GetReceivedPackets(WifiPacket::PacketType type, const MacAddress& sender) {
+    return {};
+}
+}
```
lioncash:
```cpp
} // namespace NWM
} // namespace Service
```

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,172 @@
...
+struct BeaconNodeInfo {
+    u64_be friend_code_seed;
+    std::array<u16_be, 10> username;
+    u16_be network_node_id;
+};
+
+static_assert(sizeof(BeaconNodeInfo) == 0x1E, "BeaconNodeInfo has incorrect size.");
+
+struct BeaconData {
+    u8 md5_hash[0x10];
```
lioncash: ditto wrt std::array.

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,172 @@
...
+};
+
+static_assert(sizeof(TagHeader) == 2, "TagHeader has incorrect size.");
+
+struct DummyTag {
+    static const u8 TagId = 20;
+    TagHeader header;
+    std::array<u8, 3> oui;
+    u8 oui_type;
+    u8 data[3];
```
lioncash: any reason for this one to not be a std::array?

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,172 @@
...
+};
+
+static_assert(sizeof(DummyTag) == 9, "DummyTag has incorrect size.");
+
+struct NetworkInfoTag {
+    static const u8 TagId = 21;
+
+    TagHeader header;
+    u8 network_info[0x1F];
```
lioncash: Ditto wrt std::array.

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,172 @@
...
+    Dummy = 20,
+    NetworkInfo = 21,
+    EncryptedData0 = 24,
+    EncryptedData1 = 25,
+};
+
+using MacAddress = std::array<u8, 6>;
+
+const u32 UDSMaxNodes = 16;
+const std::array<u8, 3> NintendoOUI = {0x00, 0x1F, 0x32};
```
lioncash: constexpr

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,172 @@
...
+enum class NintendoTagId : u8 {
+    Dummy = 20,
+    NetworkInfo = 21,
+    EncryptedData0 = 24,
+    EncryptedData1 = 25,
+};
+
+using MacAddress = std::array<u8, 6>;
+
+const u32 UDSMaxNodes = 16;
```
lioncash: constexpr

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,172 @@
...
+    TagHeader header;
+    std::array<u8, 3> oui;
+    u8 oui_type;
+    u8 data[3];
+};
+
+static_assert(sizeof(DummyTag) == 9, "DummyTag has incorrect size.");
+
+struct NetworkInfoTag {
+    static const u8 TagId = 21;
```
lioncash: `static constexpr`

src/core/hle/service/nwm/uds_beacon.cpp
```diff
@@ -0,0 +1,321 @@
...
+    for (const NodeInfo& node : nodes) {
+        // Serialize each node and convert the data from
+        // host byte-order to Big Endian.
+        BeaconNodeInfo info{};
+        info.friend_code_seed = node.friend_code_seed;
+        info.network_node_id = node.network_node_id;
+        for (int i = 0; i < info.username.size(); ++i)
+            info.username[i] = node.username[i];
+
+        buffer.insert(buffer.end(), (u8*)&info, (u8*)&info + sizeof(info));
```
lioncash: `reinterpret_cast`

src/core/hle/service/nwm/uds_beacon.cpp
```diff
@@ -0,0 +1,321 @@
...
+    std::memcpy(buffer.data(), &data, sizeof(BeaconData));
+
+    for (const NodeInfo& node : nodes) {
+        // Serialize each node and convert the data from
+        // host byte-order to Big Endian.
+        BeaconNodeInfo info{};
+        info.friend_code_seed = node.friend_code_seed;
+        info.network_node_id = node.network_node_id;
+        for (int i = 0; i < info.username.size(); ++i)
+            info.username[i] = node.username[i];
```
lioncash: Because both `username`s here are std::arrays containing the same type of data, and have the same length, the loop can just be replaced with:

```cpp
info.username = node.username;
```

src/core/hle/service/nwm/uds_beacon.cpp
```diff
@@ -0,0 +1,321 @@
...
+/**
+ * Generates a buffer with the Dummy Nintendo tag.
+ * It is currently unknown what this tag does.
+ * TODO(Subv): Figure out if this is needed and what it does.
+ * @returns A buffer with the Nintendo tagged parameters of the beacon frame.
+ */
+std::vector<u8> GenerateNintendoDummyTag() {
+    // Note: These values were taken from a packet capture of an o3DS XL
+    // broadcasting a Super Smash Bros. 4 lobby.
+    static const std::vector<u8> dummy_data = {0x0A, 0x00, 0x00};
```
lioncash: This can just be a static constexpr std::array instead of heap allocating, can it not?

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,172 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <deque>
+#include <vector>
+#include "common/swap.h"
```
lioncash: `common/common_types.h` should be included as well.

src/core/hle/service/nwm/uds_beacon.cpp
```diff
@@ -0,0 +1,321 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/hle/service/nwm/nwm_uds.h"
+#include "core/hle/service/nwm/uds_beacon.h"
+
```
lioncash: `<cstring>` and `"common/common_types.h", should be included.

src/core/hle/service/nwm/uds_beacon.cpp
Subv: They're not the same, one is an array of u16_be, and the other is an array of u16_le

src/core/hle/service/nwm/uds_beacon.cpp
JayFoxRox: As it is a duration, maybe use std::chrono or somehow calculate it? `15 * 60 * 1000 * 1000` (maybe with more constants or some function to hide the ugliness?)

src/core/hle/service/nwm/nwm_uds.cpp
Subv: I feel just using the 1 is more explicit, besides, there's a comment about it

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -82,29 +86,69 @@ static void Shutdown(Interface* self) {
...
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 out_buffer_size = cmd_buff[1];
-    u32 unk1 = cmd_buff[2];
-    u32 unk2 = cmd_buff[3];
-    u32 mac_address = cmd_buff[4];
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x0F, 16, 4);
 
-    u32 unk3 = cmd_buff[6];
+    u32 out_buffer_size = rp.Pop<u32>();
+    rp.Skip(2, false);
```
wwylele: I see these are marked as "unknown" (instead of uninitialized) on 3dbrew, so I think we should still log them?

src/core/hle/service/nwm/uds_beacon.cpp
```diff
@@ -0,0 +1,325 @@
...
+// broadcasting a Super Smash Bros. 4 lobby.
+constexpr u16 DefaultExtraCapabilities = 0x0431;
+
+// Size of the SSID broadcast by an UDS beacon frame.
+constexpr u8 UDSBeaconSSIDSize = 8;
+
+// The maximum size of the data stored in the EncryptedData0 tag (24).
+constexpr u32 EncryptedDataSizeCutoff = 0xFA;
+
+// NWM Beacon data encryption key, taken from the NWM module code.
```
wwylele: Should add a comment that we stub it with zeros key here (instead of implying the actual key is all zeros)

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,170 @@
...
+#include <deque>
+#include <vector>
+#include "common/common_types.h"
+#include "common/swap.h"
+#include "core/hle/service/service.h"
+
+namespace Service {
+namespace NWM {
+
+enum class TagId : u8 {
```
wwylele: Are these `enum`/`struct` needed to expose in the header file?

src/core/hle/service/nwm/uds_beacon.cpp
wwylele: I think `std::copy` should work? (not sure if it is guaranteed to do type casting properly)

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -82,29 +86,69 @@ static void Shutdown(Interface* self) {
...
+    for (const auto& beacon : beacons) {
+        BeaconEntryHeader entry{};
+        // TODO(Subv): Figure out what this size is used for.
+        entry.unk_size = sizeof(BeaconEntryHeader) + beacon.data.size();
+        entry.total_size = sizeof(BeaconEntryHeader) + beacon.data.size();
+        entry.wifi_channel = beacon.channel;
+        entry.header_size = sizeof(BeaconEntryHeader);
+        entry.mac_address = beacon.transmitter_address;
+
+        ASSERT(current_buffer_pos < out_buffer_ptr + out_buffer_size);
```
wwylele: Should be `current_buffer_pos + sizeof(BeaconEntryHeader) + beacon.data.size() <= out_buffer_ptr + out_buffer_size`?
Also, what prevents this ASSERT from failure? What if the beacon gets piled up to excess the buffer before the game calls this function? What does real uds service do in this case?

src/core/hle/service/nwm/uds_beacon.h
```diff
@@ -0,0 +1,170 @@
...
+    std::vector<u8> data;
+    MacAddress transmitter_address; ///< Mac address of the transmitter.
+    MacAddress destination_address; ///< Mac address of the receiver.
+    u8 channel;                     ///< WiFi channel where this frame was transmitted.
+};
+
+/**
+ * Decrypts the beacon data buffer for the network described by `network_info`.
+ */
+void DecryptBeaconData(const NetworkInfo& network_info, std::vector<u8>& buffer);
```
wwylele: I can't find the user of this function. Is this meant to be the implementation of `DecryptBeaconData` service function?

Subv: yes, it's coming in the next pr. ive found that dropping everything on the same pr just causes it to sit unreviewed for several months

src/core/hle/service/nwm/nwm_uds.cpp
wwylele: Just leave a record here: Subv said real uds service will crash in this case

Subv: It would _probably_ crash, or just discard the extra beacons, i'll have to perform more reverse engineering to find out for sure

Subv:
Updated, can we get this merged?

Subv:
I think this has marinated enough, I'll merge it when CI finishes

src/core/hle/service/nwm/uds_beacon.h
Subv: yes, nwm_uds uses them too

PR #2662, opened 2017-04-12
Better looking status bar under Linux Ubuntu

Cereal-Killa:
This commit will remove the inner border on the status bar cells under Linux.
I undid the first commit to replace the tab with spaces.

screenshot from 2017-04-12 16-06-49


jroweboy:
Thanks for the PR! (I went ahead and squashed it down to one commit and merged it)

PR #2664, opened 2017-04-15
netCore stubb

B3n30:


JayFoxRox:
I don't think this is supposed to be here?
Lacking PR description, shitty title etc.

Closed.

PR #2667, opened 2017-04-17
input_common/sdl: add support for binding button to axis

wwylele:
Controllers like Xbox have left/right triggers that are ideal for binding with 3DS ZL/ZR, but they are exposed via axis API. Also, one may want to bind controller stick (axis API as well) to 3DS D-pad. This resolves the problem of not being able to bind axis to button.

Usage: `button_*="engine:sdl,joystick:*,axis:*,threshold:*,direction:+/-"`. refer to the comment in the code for detail

Here are some example:
Binding controller stick to 3DS D-pad:
```
button_down="engine:sdl,joystick:0,axis:1,threshold:0.5,direction:+"
button_left="engine:sdl,joystick:0,axis:0,threshold:-0.5,direction:-"
button_right="engine:sdl,joystick:0,axis:0,threshold:0.5,direction:+"
button_up="engine:sdl,joystick:0,axis:1,threshold:-0.5,direction:-"
```
Binding Xbox trigger to 3DS ZL/ZR (not confirmed working):
```
button_zl="engine:sdl,joystick:0,axis:8,threshold:0.5,direction:+"
button_zr="engine:sdl,joystick:0,axis:9,threshold:0.5,direction:+"
```

### DIFF
--- a/src/input_common/sdl/sdl.cpp
+++ b/src/input_common/sdl/sdl.cpp
@@ -8,6 +8,7 @@
 #include <tuple>
 #include <unordered_map>
 #include <SDL.h>
+#include "common/logging/log.h"
 #include "common/math_util.h"
 #include "input_common/sdl/sdl.h"
 
@@ -40,12 +41,16 @@ public:
         return SDL_JoystickGetButton(joystick.get(), button) == 1;
     }
 
-    std::tuple<float, float> GetAnalog(int axis_x, int axis_y) const {
+    float GetAxis(int axis) const {
         if (!joystick)
             return {};
         SDL_JoystickUpdate();
-        float x = SDL_JoystickGetAxis(joystick.get(), axis_x) / 32767.0f;
-        float y = SDL_JoystickGetAxis(joystick.get(), axis_y) / 32767.0f;
+        return SDL_JoystickGetAxis(joystick.get(), axis) / 32767.0f;
+    }
+
+    std::tuple<float, float> GetAnalog(int axis_x, int axis_y) const {
+        float x = GetAxis(axis_x);
+        float y = GetAxis(axis_y);
         y = -y; // 3DS uses an y-axis inverse from SDL
 
         // Make sure the coordinates are in the unit circle,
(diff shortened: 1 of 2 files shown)


src/input_common/sdl/sdl.cpp
```diff
@@ -153,6 +185,21 @@ class SDLButtonFactory final : public Input::Factory<Input::ButtonDevice> {
             }
             return std::make_unique<SDLDirectionButton>(GetJoystick(joystick_index), hat,
                                                         direction);
+        } else if (params.Has("axis")) {
```
lioncash: `else` isn't necessary here because of the return within the initial if statement.

src/input_common/sdl/sdl.cpp
```diff
@@ -97,6 +101,28 @@ class SDLDirectionButton final : public Input::ButtonDevice {
...
+    explicit SDLAxisButton(std::shared_ptr<SDLJoystick> joystick_, int axis_, float threshold_,
+                           bool trigger_if_greater_)
+        : joystick(joystick_), axis(axis_), threshold(threshold_),
+          trigger_if_greater(trigger_if_greater_) {}
+
+    bool GetStatus() const override {
+        float axis_value = joystick->GetAxis(axis);
+        if (trigger_if_greater)
+            return axis_value > threshold;
+        else
```
lioncash: `else` isn't necessary here.

src/input_common/sdl/sdl.cpp
```diff
@@ -153,6 +185,21 @@ class SDLButtonFactory final : public Input::Factory<Input::ButtonDevice> {
...
+            const float threshold = params.Get("threshold", 0.5f);
+            const std::string direction_name = params.Get("direction", "");
+            bool trigger_if_greater;
+            if (direction_name == "+") {
+                trigger_if_greater = true;
+            } else if (direction_name == "-") {
+                trigger_if_greater = false;
+            } else {
+                trigger_if_greater = true;
+                LOG_ERROR(Input, "Unknown direction %s", direction_name.c_str());
```
lioncash: `common/logging/log.h` needs to be included.

wwylele: fixed

src/input_common/sdl/sdl.cpp
```diff
@@ -97,6 +102,27 @@ class SDLDirectionButton final : public Input::ButtonDevice {
...
+
+    bool GetStatus() const override {
+        float axis_value = joystick->GetAxis(axis);
+        if (trigger_if_greater)
+            return axis_value > threshold;
+        return axis_value < threshold;
+    }
+
+private:
+    std::shared_ptr<SDLJoystick> joystick;
```
bunnei: unique_ptr?

src/input_common/sdl/sdl.cpp
```diff
@@ -97,6 +102,27 @@ class SDLDirectionButton final : public Input::ButtonDevice {
     Uint8 direction;
 };
 
+class SDLAxisButton final : public Input::ButtonDevice {
+public:
+    explicit SDLAxisButton(std::shared_ptr<SDLJoystick> joystick_, int axis_, float threshold_,
```
bunnei: nit: Why do all of these end in underscore? Is this a style we use anywhere else?

src/input_common/sdl/sdl.cpp
wwylele: It is a shared pointer because it shares with other devices binding with the same joystick, and works together with weak pointers in the joystick list to do automatic resource acquiring and releasing. See the existing code for `SDLButton`, `SDLAnalog` etc

src/input_common/sdl/sdl.cpp
wwylele: Because they have the same names as the member variables. I know that using the same name won't cause any issue. I don't know if we have a general code style for this, but I have been using this style since [this comment](https://github.com/citra-emu/citra/pull/1652#r59067940):
> Note that this parameter will cause variable shadowing warnings because it's the same name as the member variable.

PR #2668, opened 2017-04-17
Create a random console_unique_id

B3n30:
This will allow that each citra gets its own console_unique_id and gets rid of the hardcoded 0xDEADC0DE.

It will create a random console_unique_id when the config savegame is created.
To also have to ability to change the console_unique_id without formatting the complete config savegame a button was added to the system config dialog which allows to generate a new random console_unique_id.
system config dialog with new refresh console id button
One minor draw back is that this won't allow to set the console_unique_id manually. But most users won't ever need to set it manually, and those i need to set it, should know how to edit the savegame. On the other side adding a textbox for that in the ui, seemed way to much for a feature used that rarely.

### DIFF
--- a/src/citra_qt/configuration/configure_system.cpp
+++ b/src/citra_qt/configuration/configure_system.cpp
@@ -77,7 +77,7 @@ void ConfigureSystem::ReadSystemSettings() {
     ui->combo_sound->setCurrentIndex(sound_index);
 
     // set the console id
-    u64_le console_id = Service::CFG::GetConsoleUniqueId();
+    u64 console_id = Service::CFG::GetConsoleUniqueId();
     ui->label_console_id->setText("Console ID: 0x" + QString::number(console_id, 16).toUpper());
 }
 
@@ -159,8 +159,8 @@ void ConfigureSystem::refreshConsoleID() {
                                   QMessageBox::No | QMessageBox::Yes);
     if (reply == QMessageBox::No)
         return;
-    u32_le random_number;
-    u64_le console_id;
+    u32 random_number;
+    u64 console_id;
     Service::CFG::GenerateConsoleUniqueId(random_number, console_id);
     Service::CFG::SetConsoleUniqueId(random_number, console_id);
     Service::CFG::UpdateConfigNANDSavegame();
--- a/src/citra_qt/configuration/configure_system.cpp
+++ b/src/citra_qt/configuration/configure_system.cpp
@@ -75,6 +75,10 @@ void ConfigureSystem::ReadSystemSettings() {
     // set sound output mode
     sound_index = Service::CFG::GetSoundOutputMode();
     ui->combo_sound->setCurrentIndex(sound_index);
+
+    // set the console id
+    u64_le console_id = Service::CFG::GetConsoleUniqueId();
+    ui->label_console_id->setText("Console ID: 0x" + QString::number(console_id, 16).toUpper());
 }
 
 void ConfigureSystem::applyConfiguration() {
@@ -149,7 +153,8 @@ void ConfigureSystem::refreshConsoleID() {
     QMessageBox::StandardButton reply;
     QString warning_text = tr("This will replace your current virtual 3DS with a new one. "
                               "Your current virtual 3DS will not be recoverable. "
-                              "This might have unexpected effects in games. Continue?");
+                              "This might have unexpected effects in games. This might fail, "
+                              "if you use an outdated config savegame. Continue?");
     reply = QMessageBox::critical(this, tr("Warning"), warning_text,
                                   QMessageBox::No | QMessageBox::Yes);
     if (reply == QMessageBox::No)
@@ -157,15 +162,7 @@ void ConfigureSystem::refreshConsoleID() {
     u32_le random_number;
     u64_le console_id;
     Service::CFG::GenerateConsoleUniqueId(random_number, console_id);
-    ResultCode res = Service::CFG::SetConsoleUniqueId(random_number, console_id);
-    if (res != RESULT_SUCCESS) {
-        QMessageBox::critical(this, tr("Failure"), tr("Failed to set a new console ID"),
-                              QMessageBox::Ok);
-        return;
-    }
+    Service::CFG::SetConsoleUniqueId(random_number, console_id);
     Service::CFG::UpdateConfigNANDSavegame();
-    QMessageBox::information(this, tr("New Console ID"),
-                             tr("Your new unique console ID is 0x") +
-                                 QString::number(console_id, 16).toUpper(),
-                             QMessageBox::Ok);
+    ui->label_console_id->setText("Console ID: 0x" + QString::number(console_id, 16).toUpper());
 }
--- a/src/citra_qt/configuration/configure_system.cpp
+++ b/src/citra_qt/configuration/configure_system.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <QMessageBox>
 #include "citra_qt/configuration/configure_system.h"
 #include "citra_qt/ui_settings.h"
 #include "core/core.h"
@@ -15,9 +16,11 @@ static const std::array<int, 12> days_in_month = {{
 
 ConfigureSystem::ConfigureSystem(QWidget* parent) : QWidget(parent), ui(new Ui::ConfigureSystem) {
     ui->setupUi(this);
-    connect(ui->combo_birthmonth, SIGNAL(currentIndexChanged(int)),
-            SLOT(updateBirthdayComboBox(int)));
-    connect(ui->button_refresh_console_id, SIGNAL(clicked()), this, SLOT(refreshConsoleID()));
+    connect(ui->combo_birthmonth,
+            static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this,
+            &ConfigureSystem::updateBirthdayComboBox);
+    connect(ui->button_regenerate_console_id, &QPushButton::clicked, this,
+            &ConfigureSystem::refreshConsoleID);
 
     this->setConfiguration();
 }
@@ -143,6 +146,19 @@ void ConfigureSystem::updateBirthdayComboBox(int birthmonth_index) {
 }
 
 void ConfigureSystem::refreshConsoleID() {
+    QMessageBox::StandardButton reply;
+    reply = QMessageBox::critical(
+        this, tr("Warning"),
+        tr("This is an advanced feature. Some of your games might lose their saves. Only "
+           "continue, if you know what you are doing. Do you want to continue?"),
+        QMessageBox::Abort | QMessageBox::Yes);
+    if (reply == QMessageBox::Abort)
+        return;
     Service::CFG::GenerateConsoleUniqueId();
     Service::CFG::UpdateConfigNANDSavegame();
+    const u64 console_id = Service::CFG::GetConsoleUniqueId();
+    QMessageBox::information(this, tr("New Console ID"),
+                             tr("Your new unique console ID is 0x") +
+                                 QString::number(console_id, 16).toUpper(),
+                             QMessageBox::Ok);
 }
--- a/src/citra_qt/configuration/configure_system.cpp
+++ b/src/citra_qt/configuration/configure_system.cpp
@@ -147,16 +147,23 @@ void ConfigureSystem::updateBirthdayComboBox(int birthmonth_index) {
 
 void ConfigureSystem::refreshConsoleID() {
     QMessageBox::StandardButton reply;
-    reply = QMessageBox::critical(this, tr("Warning"),
-                                  tr("This will replace your virtual 3DS with a new one. Your "
-                                     "current 3DS will not be recoverable. "
-                                     "This might have unexpected effects in games. Continue?"),
-                                  QMessageBox::Abort | QMessageBox::Yes);
-    if (reply == QMessageBox::Abort)
+    QString warning_text = tr("This will replace your current virtual 3DS with a new one. "
+                              "Your current virtual 3DS will not be recoverable. "
+                              "This might have unexpected effects in games. Continue?");
+    reply = QMessageBox::critical(this, tr("Warning"), warning_text,
+                                  QMessageBox::No | QMessageBox::Yes);
+    if (reply == QMessageBox::No)
         return;
-    Service::CFG::GenerateConsoleUniqueId();
+    u32_le random_number;
+    u64_le console_id;
+    Service::CFG::GenerateConsoleUniqueId(random_number, console_id);
+    ResultCode res = Service::CFG::SetConsoleUniqueId(random_number, console_id);
+    if (res != RESULT_SUCCESS) {
+        QMessageBox::critical(this, tr("Failure"), tr("Failed to set a new console ID"),
+                              QMessageBox::Ok);
+        return;
+    }
     Service::CFG::UpdateConfigNANDSavegame();
-    const u64 console_id = Service::CFG::GetConsoleUniqueId();
     QMessageBox::information(this, tr("New Console ID"),
                              tr("Your new unique console ID is 0x") +
                                  QString::number(console_id, 16).toUpper(),
--- a/src/citra_qt/configuration/configure_system.cpp
+++ b/src/citra_qt/configuration/configure_system.cpp
@@ -17,6 +17,7 @@ ConfigureSystem::ConfigureSystem(QWidget* parent) : QWidget(parent), ui(new Ui::
     ui->setupUi(this);
     connect(ui->combo_birthmonth, SIGNAL(currentIndexChanged(int)),
             SLOT(updateBirthdayComboBox(int)));
+    connect(ui->button_refresh_console_id, SIGNAL(clicked()), this, SLOT(refreshConsoleID()));
 
     this->setConfiguration();
 }
--- a/src/citra_qt/configuration/configure_system.cpp
+++ b/src/citra_qt/configuration/configure_system.cpp
@@ -140,3 +140,8 @@ void ConfigureSystem::updateBirthdayComboBox(int birthmonth_index) {
     // restore the day selection
     ui->combo_birthday->setCurrentIndex(birthday_index);
 }
+
+void ConfigureSystem::refreshConsoleID() {
+    Service::CFG::GenerateConsoleUniqueId();
+    Service::CFG::UpdateConfigNANDSavegame();
+}
--- a/src/citra_qt/configuration/configure_system.cpp
+++ b/src/citra_qt/configuration/configure_system.cpp
@@ -147,11 +147,11 @@ void ConfigureSystem::updateBirthdayComboBox(int birthmonth_index) {
 
 void ConfigureSystem::refreshConsoleID() {
     QMessageBox::StandardButton reply;
-    reply = QMessageBox::critical(
-        this, tr("Warning"),
-        tr("This is an advanced feature. Some of your games might lose their saves. Only "
-           "continue, if you know what you are doing. Do you want to continue?"),
-        QMessageBox::Abort | QMessageBox::Yes);
+    reply = QMessageBox::critical(this, tr("Warning"),
+                                  tr("This will replace your virtual 3DS with a new one. Your "
+                                     "current 3DS will not be recoverable. "
+                                     "This might have unexpected effects in games. Continue?"),
+                                  QMessageBox::Abort | QMessageBox::Yes);
     if (reply == QMessageBox::Abort)
         return;
     Service::CFG::GenerateConsoleUniqueId();
--- a/src/citra_qt/configuration/configure_system.ui
+++ b/src/citra_qt/configuration/configure_system.ui
@@ -267,12 +267,5 @@
   </layout>
  </widget>
  <resources/>
- <connections>
-  <connection>
-   <sender>button_refresh_console_id</sender>
-   <signal>clicked()</signal>
-   <receiver>ConfigureSystem</receiver>
-   <slot>refreshConsoleID()</slot>
-  </connection>
- </connections>
+ <connections/>
 </ui>
--- a/src/citra_qt/configuration/configure_system.ui
+++ b/src/citra_qt/configuration/configure_system.ui
@@ -221,7 +221,7 @@
          </widget>
         </item>
         <item row="4" column="1">
-         <widget class="QPushButton" name="button_refresh_console_id">
+         <widget class="QPushButton" name="button_regenerate_console_id">
           <property name="sizePolicy">
            <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
             <horstretch>0</horstretch>
@@ -232,7 +232,7 @@
            <enum>Qt::RightToLeft</enum>
           </property>
           <property name="text">
-           <string>Refresh Console ID</string>
+           <string>Generate new console ID</string>
           </property>
          </widget>
         </item>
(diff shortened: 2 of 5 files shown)


src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +667,13 @@ SoundOutputMode GetSoundOutputMode() {
     return static_cast<SoundOutputMode>(block);
 }
 
+void GenerateConsoleUniqueId() {
+        const u64 console_id = dis(rng);
```
Subv: There are too many whitespaces here

RavenHome1:
is there any advantage to this?

B3n30:
In my opinion that should be included as soon as possible. Every user using Citra now with the same unique id. If at some stage a unique id is required every user is forced to generate it manually. This might cause a lot of support requests and issues. Therefore adding a unique id to each Citra, shouldn't be delayed, cause right now there is no use for it.

src/citra_qt/configuration/configure_system.ui
```diff
@@ -251,5 +267,12 @@
   </layout>
  </widget>
  <resources/>
- <connections/>
+ <connections>
```
jroweboy: I don't remember anywhere else in the program that we set up connections for GUI elements in the ui file. If we want to move to that it might be worth it to make a new issue about it.

B3n30: Moved the connect form .ui to the constructor, which is probably the better way. As i recently discussed with @JayFoxRox would it be best, if we get rid of the .ui files and implement everything in the cpp-files.

src/citra_qt/configuration/configure_system.ui
```diff
@@ -220,6 +220,22 @@
...
+           <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+            <horstretch>0</horstretch>
+            <verstretch>0</verstretch>
+           </sizepolicy>
+          </property>
+          <property name="layoutDirection">
+           <enum>Qt::RightToLeft</enum>
+          </property>
+          <property name="text">
+           <string>Refresh Console ID</string>
```
MerryMage: s/Refresh/Regenerate/ may be better wording.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +665,11 @@ SoundOutputMode GetSoundOutputMode() {
     return static_cast<SoundOutputMode>(block);
 }
 
+void GenerateConsoleUniqueId() {
+    const u64 console_id = dis(rng);
+    SetConfigInfoBlock(ConsoleUniqueID1BlockID, sizeof(console_id), 0xE, &console_id);
+    SetConfigInfoBlock(ConsoleUniqueID2BlockID, sizeof(console_id), 0xE, &console_id);
```
MerryMage: Would it be possible to return an error if these calls fail?

If this did, one could replace the code in `FormatConfig()` with a call to this function to avoid code duplication.

src/citra_qt/configuration/configure_system.cpp
```diff
@@ -17,6 +17,7 @@ ConfigureSystem::ConfigureSystem(QWidget* parent) : QWidget(parent), ui(new Ui::
     ui->setupUi(this);
     connect(ui->combo_birthmonth, SIGNAL(currentIndexChanged(int)),
             SLOT(updateBirthdayComboBox(int)));
+    connect(ui->button_refresh_console_id, SIGNAL(clicked()), this, SLOT(refreshConsoleID()));
```
jroweboy: Minor nitpick, but we are slowly moving over to qt5 signals and slots syntax (as the qt4 stuff is so out of date and we don't really support it anymore) http://doc.qt.io/qt-5/signalsandslots.html It'd be nice to change that in the entire function, but if you don't feel like it, it'd be a simple task for a beginner to pick up.

src/citra_qt/configuration/configure_system.ui
JayFoxRox: Agreed. Refresh is very confusing and users might accidentally hit it.
"Generate new console ID" would be better + there should probably be a warning (as this could potentially break games if they are all of sudden running on a different 3DS)

B3n30: These settings can't be changed while a game is running. So hitting this button will behave as if you plug in your game into another 3ds. I don't think a warning is needed

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -118,6 +118,9 @@ static const std::vector<u8> cfg_system_savedata_id = {
 
 static u32 preferred_region_code = 0;
 
+static std::mt19937_64 rng{};
```
lioncash: These can just be placed within the two scopes they're needed in, rather than being file-scope, or put into an internally linked function to generate the value.

yuriks: - This needs to be seeded somehow, otherwise it will always generate the same value.
- Rather than using a prng, use something like [random_device](http://en.cppreference.com/w/cpp/numeric/random/random_device) or a random source from cryptocpp. (iirc implementations of random_device tend to be very spotty.)

src/citra_qt/configuration/configure_system.ui
yuriks: I agree that "Regenerate" is also a better word to use. "Refresh" sounds like something that should yield the same result if I click the button a bunch of times.

B3n30:
Addressed all of the issues.
@JayFoxRox requested a warning dialog and a feedback that it actually changed the console_id
This is how it looks now:
bildschirmfoto 2017-04-18 um 11 27 02
Clicking the button will lead to:
bildschirmfoto 2017-04-18 um 11 25 29
And only if you click ok, the console ID will be changed and this message will appear:
bildschirmfoto 2017-04-18 um 11 25 44

src/core/hle/service/cfg/cfg.cpp
B3n30: std::random_device shouldn't be used, at least not if it should work properly on different platforms. 
http://en.cppreference.com/w/cpp/numeric/random/random_device:
> std::random_device may be implemented in terms of an implementation-defined pseudo-random number engine if a non-deterministic source (e.g. a hardware device) is not available to the implementation. In this case each std::random_device object may generate the same number sequence.
Which in this case would mean, hitting the button the first time would always result in the same console_id.
Thus I used the rng from cryptopp.

src/citra_qt/configuration/configure_system.ui
B3n30: It's now called "Generate new console ID"

wwylele:
I doubt the complete randomness of this block, for
 - The config savegame dump from my 3DS has random-looking value for the first four bytes of the console unique ID block , but the second four bytes are "00 00 0f 80", which looks not random at all.
 - As 3dbrew says (and confirmed on my 3DS): The first word is the same as LocalFriendCodeSeed, while the latter is a separate word. 

I will try to find out what the second word is.

src/citra_qt/configuration/configure_system.cpp
```diff
@@ -140,3 +144,21 @@ void ConfigureSystem::updateBirthdayComboBox(int birthmonth_index) {
...
     ui->combo_birthday->setCurrentIndex(birthday_index);
 }
+
+void ConfigureSystem::refreshConsoleID() {
+    QMessageBox::StandardButton reply;
+    reply = QMessageBox::critical(this, tr("Warning"),
+                                  tr("This will replace your virtual 3DS with a new one. Your "
+                                     "current 3DS will not be recoverable. "
+                                     "This might have unexpected effects in games. Continue?"),
+                                  QMessageBox::Abort | QMessageBox::Yes);
```
lioncash: This should likely be `QMessageBox::No` (reads nicer with the question in the message box instance) or `QMessageBox::Cancel`. Abort implies a currently started operation (which isn't the case here).

src/citra_qt/configuration/configure_system.cpp
```diff
@@ -147,11 +147,11 @@ void ConfigureSystem::updateBirthdayComboBox(int birthmonth_index) {
...
 void ConfigureSystem::refreshConsoleID() {
     QMessageBox::StandardButton reply;
-    reply = QMessageBox::critical(
-        this, tr("Warning"),
-        tr("This is an advanced feature. Some of your games might lose their saves. Only "
-           "continue, if you know what you are doing. Do you want to continue?"),
-        QMessageBox::Abort | QMessageBox::Yes);
+    reply = QMessageBox::critical(this, tr("Warning"),
+                                  tr("This will replace your virtual 3DS with a new one. Your "
+                                     "current 3DS will not be recoverable. "
```
JayFoxRox: "current virtual 3DS"

(Otherwise we'll get support requests like: "will this break my 3DS?!?!" from idiots)

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +658,30 @@ SoundOutputMode GetSoundOutputMode() {
     return static_cast<SoundOutputMode>(block);
 }
 
+ResultCode GenerateConsoleUniqueId() {
+    CryptoPP::AutoSeededRandomPool rng;
+    const u32 random_number = static_cast<u16>((CryptoPP::Integer(rng, 16).ConvertToLong()));
```
wwylele: casting to u16 and assign to u32?

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +658,30 @@ SoundOutputMode GetSoundOutputMode() {
...
+    const u32 local_friend_code_seed = rng.GenerateWord32();
+    const u64 console_id =
+        ((0x3FFFFFFFF) & local_friend_code_seed) | (static_cast<u64>(random_number) << 48);
+    ResultCode res =
+        SetConfigInfoBlock(ConsoleUniqueID1BlockID, sizeof(console_id), 0xE, &console_id);
```
wwylele: The block need to be created by `CreateConfigInfoBlk` before set

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +658,30 @@ SoundOutputMode GetSoundOutputMode() {
     return static_cast<SoundOutputMode>(block);
 }
 
+ResultCode GenerateConsoleUniqueId() {
+    CryptoPP::AutoSeededRandomPool rng;
+    const u32 random_number = static_cast<u16>((CryptoPP::Integer(rng, 16).ConvertToLong()));
+    const u32 local_friend_code_seed = rng.GenerateWord32();
```
wwylele: Shouldn't this be u64? Although yeah from my config savegame, the higher word seems not random but zero. However in the case that it is a u32, the next line `0x3FFFFFFFF &` (note there are eight F there) makes no sense be cause it won't mask anything out.

src/core/hle/service/cfg/cfg.cpp
B3n30: This was just to assure that only the lower 16bit are actually random while the upper 16bit are fixed to zero. This is the format this number is stored in block 0x90002. But the cast is probably unnecessary.

src/core/hle/service/cfg/cfg.cpp
B3n30: According to https://3dbrew.org/wiki/Nand/private/movable.sed the upper 4 byte are usually zero. Also If u8[1] is zero, then u8 [0], [2], and [3] must be zero as well, with u8 being the upper 4byte. This can't be assured if they are randomly generated.

src/core/hle/service/cfg/cfg.cpp
B3n30: This is caused by an idea @MerryMage suggested. The format should have the create. The next commit will have a fix for it. There is still the problem that the system only creates new entries if the config savegame is written from scratch. Thus adding any block to the config savegame will require every user to delete the existing savegame. So i could add a create if the set fails, but a better solution would probably that a set should atomatically create it if it doesn't exist. This shouldn't be a separate PR.

src/core/hle/service/cfg/cfg.cpp
wwylele: @B3n30 First of all, the sentence "LocalFriendCodeSeed, this is the same as the first 0x8-bytes from the movable.sed keyY. " in [this page](https://3dbrew.org/wiki/Nandrw/sys/LocalFriendCodeSeed_B) doesn't refer to offset 0 of movable.sed. It is more likely it refers to offset 0x118 instead (although this doesn't match the data I got from my 3ds at all). Second, let's assume this is a u32, then `0x3FFFFFFFF &` in the next line should be removed (which is there assuming a u64 seed)

IMO at this stage we should pause our work on citra and clean up the documentation on 3dbrew first. It looks to me like a mess and potentially containing mistakes.

B3n30:
@lioncash I think that adding a button to the ui, is (almost) to much for a feature most users won't ever use. Adding an additional text seems - at least in my opinion - unnecessary, since the value itself is of no use for the user. The only important thing is, that its different, to the one it was before and that is different to any other console id used.

jfmherokiller:
the only issue i see with this commit is that it could make saves break. but i do recognize its purpose for increasing accuracy. I do have a suggestion though and it's to add an option to allow the user to import the console_id from a nand dump

B3n30:
There is no save I know off, that will be get lost through this PR. This would only be saves tied to one specific console. If a user has dumped his config savegame from a 3DS, the best way (and also the easiest) is to just use that file as the config savegame. That way the citra console id will be the same as the real 3DS one. This behavior is the only reason the button exists: To get a unique id while you can still use your dumped savegame. Every user that didn't dump the savegame will automatically get a unique id, when the config savegame is created.

src/core/hle/service/cfg/cfg.cpp
yuriks: Masking with 0xFFFF would be clearer. In this case, the returned value will only have a range of [0, 2^16-1] anyway because you're passing 16 on the second param.

However, rather than using `Integer`, you should just call `rng.GenerateWord32(0, 0xFFFF)`.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +658,30 @@ SoundOutputMode GetSoundOutputMode() {
...
+    const u64 console_id =
```
yuriks: This should be a `u64_le` to be endian-safe.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +658,30 @@ SoundOutputMode GetSoundOutputMode() {
...
+    res = SetConfigInfoBlock(ConsoleUniqueID2BlockID, sizeof(console_id), 0xE, &console_id);
+    if (!res.IsSuccess())
+        return res;
+
+    res = SetConfigInfoBlock(ConsoleUniqueID3BlockID, sizeof(random_number), 0xE, &random_number);
+    return RESULT_SUCCESS;
+}
+
+u64 GetConsoleUniqueId() {
+    u64 console_id;
```
yuriks: Should also be a `u64_le`.

src/core/hle/service/cfg/cfg.cpp
yuriks: This should also be a `u32_le` since you're writing it out as bytes to the block.

src/citra_qt/configuration/configure_system.cpp
```diff
@@ -140,3 +144,21 @@ void ConfigureSystem::updateBirthdayComboBox(int birthmonth_index) {
     // restore the day selection
     ui->combo_birthday->setCurrentIndex(birthday_index);
 }
+
+void ConfigureSystem::refreshConsoleID() {
+    QMessageBox::StandardButton reply;
+    reply = QMessageBox::critical(this, tr("Warning"),
```
yuriks: I assume you're splitting this in 2 lines because we align parameters to the opening parenthesis and it looks ugly like that? *barf* :(

src/citra_qt/configuration/configure_system.cpp
```diff
@@ -71,6 +75,10 @@ void ConfigureSystem::ReadSystemSettings() {
     // set sound output mode
     sound_index = Service::CFG::GetSoundOutputMode();
     ui->combo_sound->setCurrentIndex(sound_index);
+
+    // set the console id
+    u64_le console_id = Service::CFG::GetConsoleUniqueId();
```
yuriks: Doesn't need to be `u64_le`. The *only* variables that need to be `_le` are one you're memcpying to or from, or otherwise interpreting as raw bytes.

src/citra_qt/configuration/configure_system.cpp
```diff
@@ -140,3 +148,21 @@ void ConfigureSystem::updateBirthdayComboBox(int birthmonth_index) {
...
+    QString warning_text = tr("This will replace your current virtual 3DS with a new one. "
+                              "Your current virtual 3DS will not be recoverable. "
+                              "This might have unexpected effects in games. This might fail, "
+                              "if you use an outdated config savegame. Continue?");
+    reply = QMessageBox::critical(this, tr("Warning"), warning_text,
+                                  QMessageBox::No | QMessageBox::Yes);
+    if (reply == QMessageBox::No)
+        return;
+    u32_le random_number;
+    u64_le console_id;
```
yuriks: These don't need to be `_le`

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +669,38 @@ SoundOutputMode GetSoundOutputMode() {
     return static_cast<SoundOutputMode>(block);
 }
 
+void GenerateConsoleUniqueId(u32_le& random_number, u64_le& console_id) {
```
yuriks: Make the param types regular `u32` and `u64`, and use intermediary local `_le` variables to copy to.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +669,38 @@ SoundOutputMode GetSoundOutputMode() {
...
+    CryptoPP::AutoSeededRandomPool rng;
+    random_number = rng.GenerateWord32(0, 0xFFFF);
+    u64_le local_friend_code_seed;
+    rng.GenerateBlock(reinterpret_cast<byte*>(&local_friend_code_seed),
+                      sizeof(local_friend_code_seed));
+    console_id =
+        (local_friend_code_seed & 0x3FFFFFFFF) | (static_cast<u64_le>(random_number) << 48);
+}
+
+ResultCode SetConsoleUniqueId(u32_le random_number, u64_le console_id) {
```
yuriks: Same, params should be non-le and copied to le local vars.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +669,38 @@ SoundOutputMode GetSoundOutputMode() {
...
+
+    res = SetConfigInfoBlock(ConsoleUniqueID3BlockID, sizeof(random_number), 0xE, &random_number);
+    if (!res.IsSuccess())
+        return res;
+
+    return RESULT_SUCCESS;
+}
+
+u64_le GetConsoleUniqueId() {
+    u64 console_id;
```
yuriks: This variable should be `_le`.

B3n30:
fixed the `_le` usage. BTW, i would love to learn more about when to us _le and when to avoid it.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -669,37 +669,43 @@ SoundOutputMode GetSoundOutputMode() {
     return static_cast<SoundOutputMode>(block);
 }
 
-void GenerateConsoleUniqueId(u32_le& random_number, u64_le& console_id) {
+void GenerateConsoleUniqueId(u32& random_number, u64& console_id) {
     CryptoPP::AutoSeededRandomPool rng;
-    random_number = rng.GenerateWord32(0, 0xFFFF);
+    u32_le random_number_le = rng.GenerateWord32(0, 0xFFFF);
```
yuriks: This doesn't need to be `_le` since you don't copy it as bytes anywhere.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -669,37 +669,43 @@ SoundOutputMode GetSoundOutputMode() {
...
     u64_le local_friend_code_seed;
     rng.GenerateBlock(reinterpret_cast<byte*>(&local_friend_code_seed),
                       sizeof(local_friend_code_seed));
-    console_id =
-        (local_friend_code_seed & 0x3FFFFFFFF) | (static_cast<u64_le>(random_number) << 48);
+    u64_le console_id_le =
```
yuriks: Doesn't need to be `_le`.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -669,37 +669,43 @@ SoundOutputMode GetSoundOutputMode() {
...
+        (local_friend_code_seed & 0x3FFFFFFFF) | (static_cast<u64_le>(random_number_le) << 48);
+
+    random_number = random_number_le;
+    console_id = console_id_le;
 }
 
-ResultCode SetConsoleUniqueId(u32_le random_number, u64_le console_id) {
+ResultCode SetConsoleUniqueId(u32 random_number, u64 console_id) {
+    u32_le random_number_le = random_number;
```
yuriks: Move this var down closer to where it's used.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -671,19 +671,15 @@ SoundOutputMode GetSoundOutputMode() {
...
     u64_le local_friend_code_seed;
     rng.GenerateBlock(reinterpret_cast<byte*>(&local_friend_code_seed),
                       sizeof(local_friend_code_seed));
-    u64_le console_id_le =
-        (local_friend_code_seed & 0x3FFFFFFFF) | (static_cast<u64_le>(random_number_le) << 48);
-
-    random_number = random_number_le;
-    console_id = console_id_le;
+    console_id =
+        (local_friend_code_seed & 0x3FFFFFFFF) | (static_cast<u64_le>(random_number) << 48);
```
wwylele: `static_cast<u64>`, so that endianess casting only happens at the assignment.

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -663,5 +669,41 @@ SoundOutputMode GetSoundOutputMode() {
...
 }
 
+void GenerateConsoleUniqueId(u32& random_number, u64& console_id) {
+    CryptoPP::AutoSeededRandomPool rng;
+    random_number = rng.GenerateWord32(0, 0xFFFF);
+    u64_le local_friend_code_seed;
+    rng.GenerateBlock(reinterpret_cast<byte*>(&local_friend_code_seed),
+                      sizeof(local_friend_code_seed));
+    console_id =
+        (local_friend_code_seed & 0x3FFFFFFFF) | (static_cast<u64_le>(random_number) << 48);
```
wwylele: static_cast<u64>

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -438,13 +439,18 @@ ResultCode FormatConfig() {
     if (!res.IsSuccess())
         return res;
 
-    res = CreateConfigInfoBlk(ConsoleUniqueID1BlockID, sizeof(CONSOLE_UNIQUE_ID), 0xE,
-                              &CONSOLE_UNIQUE_ID);
+    u32_le random_number;
+    u64_le console_id;
+    GenerateConsoleUniqueId(random_number, console_id);
```
wwylele:
```
u32 random_number;
 u64 console_id;
GenerateConsoleUniqueId(random_number, console_id);
u32_le random_number_le = random_number; // use these two le variables to initialize config block
u64_le  console_id_le =  console_id;
```
Another solution is to let `GenerateConsoleUniqueId` returns a `tuple` and use `std::tie` for auto type conversion

bunnei:
Looks good. Thanks for submitting this!

PR #2669, opened 2017-04-18
Frontend: Prevent FileSystemWatcher from blocking UI thread

jroweboy:
Instead of tying the QFileSystemWatcher to the GameList and updating in the UI thread, this change moves it to the worker thread. Since it gets deleted and recreated as part of the worker thread, this prevents it from
ever getting used from multiple threads (which is why it was originally done on the UI thread)

macOS users have had issues recently where for some reason their default game list is to some directory where its too large and appears to permanently lock up the main thread.

### Context

The original reason that the `QFileSystemWatcher` was placed on the main thread was because of an odd behavior with sharing a watcher to a different thread. When the watcher was shared to a new thread, and the thread would end, the watcher would emit the `directoryChanged` signal for each of the directories thats its watching. This showed up in the original pull request as a game list refresh on a 30 second interval because that is the default `QThreadPool` timeout after which it would terminate the worker thread. Properly terminating the worker thread after completion caused an endless cycle of `thread creation -> watch dirs -> thread termination -> directoryChanged emitted -> populateAsync called -> thread creation -> ...` 

I have no idea what exactly causes the file watcher to emit the signal for all the directories it was watching on thread exit, but this pull request works around this issue by never moving the watcher to a worker thread in the first place. Instead it always remains a child of the `GameList` and the worker thread will emit a list directories to watch when the worker thread is done.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2669)

### DIFF
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <QApplication>
 #include <QFileInfo>
 #include <QHeaderView>
 #include <QKeyEvent>
@@ -194,6 +195,9 @@ void GameList::onFilterCloseClicked() {
 }
 
 GameList::GameList(GMainWindow* parent) : QWidget{parent} {
+    watcher = new QFileSystemWatcher(this);
+    connect(watcher, &QFileSystemWatcher::directoryChanged, this, &GameList::RefreshGameDirectory);
+
     this->main_window = parent;
     layout = new QVBoxLayout;
     tree_view = new QTreeView;
@@ -218,7 +222,6 @@ GameList::GameList(GMainWindow* parent) : QWidget{parent} {
 
     connect(tree_view, &QTreeView::activated, this, &GameList::ValidateEntry);
     connect(tree_view, &QTreeView::customContextMenuRequested, this, &GameList::PopupContextMenu);
-    connect(&watcher, &QFileSystemWatcher::directoryChanged, this, &GameList::RefreshGameDirectory);
 
     // We must register all custom types with the Qt Automoc system so that we are able to use it
     // with signals/slots. In this case, QList falls under the umbrells of custom types.
@@ -269,7 +272,22 @@ void GameList::ValidateEntry(const QModelIndex& item) {
     emit GameChosen(file_path);
 }
 
-void GameList::DonePopulating() {
+void GameList::DonePopulating(QStringList watch_list) {
+    // Clear out the old directories to watch for changes and add the new ones
+    auto watch_dirs = watcher->directories();
+    if (!watch_dirs.isEmpty()) {
+        watcher->removePaths(watch_dirs);
+    }
+    // Workaround: Add the watch paths in chunks to allow the gui to refresh
+    // This prevents the UI from stalling when a large number of watch paths are added
+    // Also artificially caps the watcher to a certain number of directories
+    constexpr int LIMIT_WATCH_DIRECTORIES = 5000;
+    constexpr int SLICE_SIZE = 25;
+    int len = std::min(watch_list.length(), LIMIT_WATCH_DIRECTORIES);
+    for (int i = 0; i < len; i += SLICE_SIZE) {
+        watcher->addPaths(watch_list.mid(i, i + SLICE_SIZE));
+        QCoreApplication::processEvents();
+    }
     tree_view->setEnabled(true);
     int rowCount = tree_view->model()->rowCount();
     search_field->setFilterResult(rowCount, rowCount);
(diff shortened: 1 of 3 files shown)


jroweboy:
> so this couldn't have possibly worked in the first place...

My mistake. I've tried so many different ways to fix this problem, that I got lazy in my testing.

I've updated the code to actually work this time, while still fixing the original issue. I'm also updating the original PR to provide context for anyone else that wants to review this.

src/citra_qt/game_list.h
```diff
@@ -44,14 +44,13 @@ class GameList : public QWidget {
...
 
     void PopupContextMenu(const QPoint& menu_location);
-    void UpdateWatcherList(const std::string& path, unsigned int recursion);
     void RefreshGameDirectory();
 
     QTreeView* tree_view = nullptr;
     QStandardItemModel* item_model = nullptr;
     GameListWorker* current_worker = nullptr;
-    QFileSystemWatcher watcher;
+    std::unique_ptr<QFileSystemWatcher> watcher;
```
yuriks: Is there any reason this was made a unique_ptr? Accident?

src/citra_qt/game_list.cpp
```diff
@@ -16,6 +16,10 @@
 #include "ui_settings.h"
 
 GameList::GameList(QWidget* parent) : QWidget{parent} {
+    watcher = std::make_unique<QFileSystemWatcher>(this);
```
yuriks: Either make it a unique_ptr, or make GameList the parent (passing `this` to the constructor), not both.

src/citra_qt/game_list.cpp
```diff
@@ -155,47 +159,15 @@ void GameList::RefreshGameDirectory() {
...
 void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsigned int recursion) {
     const auto callback = [this, recursion](unsigned* num_entries_out, const std::string& directory,
                                             const std::string& virtual_name) -> bool {
         std::string physical_name = directory + DIR_SEP + virtual_name;
 
         if (stop_processing)
             return false; // Breaks the callback loop.
-
-        if (!FileUtil::IsDirectory(physical_name) && HasSupportedFileExtension(physical_name)) {
+        bool is_dir = FileUtil::IsDirectory(physical_name);
```
yuriks: nit: You ate a blank line here.

yuriks:
ping?

jroweboy:
Updated to call `processEvents` every so often while trying to `addPaths`. This is not the most elegant solution, but its sufficient for now.

yuriks:
LGTM. I did notice this PR adds a noticeable delay of ~1 second for the list enumeration to finish compared to master, but it's not a big deal, and I think we should make this properly async eventually so it's temporary issue.

---

Reviewed 1 of 3 files at r2, 2 of 2 files at r3.

---

*[src/citra_qt/game_list.cpp, line 285 at r3](https://reviewable.io:443/reviews/citra-emu/citra/2669#-Kjke5I-IdVQL6himY0_:-Kjke5I-IdVQL6himY0a:b-397al7) ([raw file](https://github.com/citra-emu/citra/blob/b6dc1a21172f62298f1368842ddd2f2db15689d3/src/citra_qt/game_list.cpp#L285)):*
> ```cpp
>     // Also artificially caps the watcher to a certain number of directories
>     constexpr int LIMIT_WATCH_DIRECTORIES = 5000;
>     constexpr int SLICE_SIZE = 10;
> ```

I'd consider raising this to something like 25 or 50. It makes the process take much less time to eventually complete, while still leaving the UI at least somewhat usable.

---

*[src/citra_qt/game_list.h, line 100 at r3](https://reviewable.io:443/reviews/citra-emu/citra/2669#-Kjka4EtwNxIlHVkNyON:-Kjka4EulX4fxlo4S_0D:bd6ugfj) ([raw file](https://github.com/citra-emu/citra/blob/b6dc1a21172f62298f1368842ddd2f2db15689d3/src/citra_qt/game_list.h#L100)):*
> ```objc
>     QStandardItemModel* item_model = nullptr;
>     GameListWorker* current_worker = nullptr;
>     QFileSystemWatcher* watcher;
> ```

Initialize this field with nullptr too, just for consistency.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2669#-:-KjkeEPrUZ6kv_hC1tWJ:b-jal8eo)*

jroweboy:
Fixed those issues. Moved up from 10 to 25.

---

Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/citra_qt/game_list.cpp, line 285 at r3](https://reviewable.io:443/reviews/citra-emu/citra/2669#-Kjke5I-IdVQL6himY0_:-KjmzayaQO6pZco2x8eG:bsh0zx1) ([raw file](https://github.com/citra-emu/citra/blob/b6dc1a21172f62298f1368842ddd2f2db15689d3/src/citra_qt/game_list.cpp#L285)):*

I'd consider raising this to something like 25 or 50. It makes the process take much less time to eventually complete, while still leaving the UI at least somewhat usable.
</blockquote></details>

There is a noticeable lag when you do too many at once, but I'm fine moving it up to 25.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2669#-:-Kjn-sSHRU01ZFonaQOq:bf733y1)*

yuriks:
:lgtm:

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2669#-:-KjpAJ5cZIArCJ2RTvIi:bnfp4nl)*

PR #2673, opened 2017-04-20
citra-qt: game list search function

nicoboss:
A search filter that allows users to search the game list for a specific game. This helps to quickly find the game you want to play even if there is a huge amount of games in the game list.

The search is word order- and case insensitive and also accepts word parts. For example, even a search for “Oca 3D TiMe” would be able to find the correct game “Ocarina of Time 3D”.

As search data it uses the filename, title name and the exact program ID. This allows to find the game if the file isn’t named correctly.

The filter field can be turned on and off over Ctrl+F or under “view => Show Filter Bar”. Its state gets saved in qt-config.ini. During gameplay, it’s always hidden because it’s part of the game list.

Nice to know: If there is only one result left it will also launch by pressing enter in the filter field.
The filter field will be cleared by pressing ESC or launching any game.

This is my first pull request and also the first time I ever worked with Qt so tell if there is anything to correct or to improve.

Screen capture of the latest version:
gamelist_filter_layton


LittleWhite-tb:
Hello,

Nice stuff, can you add some (at least one) screen capture of the new feature, please :)

nicoboss:
Hello,
Here the screen capture

gamelist_search_mario

nicoboss:
And here some more screen captures

gamelist_search_layton_phoenixwright

gamelist_search_oot

src/citra_qt/game_list.cpp
```diff
@@ -15,10 +16,150 @@
...
+                emit gamelist->GameChosen(file_path);
+            }
+            else {
+                return QObject::eventFilter(obj, event);
+            }
+            break;
+        }
+        default: {
+            return QObject::eventFilter(obj, event);
+            break;
```
jroweboy: break after return?

src/citra_qt/game_list.cpp
```diff
@@ -15,10 +16,150 @@
...
+                gamelist->search_field->edit_search->setText("");
+                edit_search_text = "";
+                emit gamelist->GameChosen(file_path);
+            }
+            else {
+                return QObject::eventFilter(obj, event);
+            }
+            break;
+        }
+        default: {
```
jroweboy: no need to start a block here

src/citra_qt/game_list.cpp
```diff
@@ -15,10 +16,150 @@
...
+        }
+        }
+    }
 
+    edit_search_text_old = edit_search_text;
+    QStandardItemModel* item_model = new QStandardItemModel(gamelist->tree_view);
+
+    // If the searchfield is empty every item is visible
+    // Otherwise the filter gets applied
+    if (edit_search_text == "") {
```
jroweboy: `.empty()`

src/citra_qt/game_list.cpp
```diff
@@ -15,10 +16,150 @@
...
+            // Only items which filename in combination with its title contains all words
+            // that are in the searchfiel will be visible in the gamelist
+            // The search is case insensitive because of toLower()
+            // I decided not to use Qt::CaseInsensitive in containsAllWords
+            // to prevent multiple conversions of the same string
+            // or for each substring instead one for the whole string
+            if (containsAllWords(file_name + " " + file_title, edit_search_text)
+                || (file_programmid.count() == 16 && edit_search_text.contains(file_programmid))) {
+                gamelist->tree_view->setRowHidden(i, item_model->invisibleRootItem()->index(), false);
+            }
```
jroweboy: more msvc formatting (please use clang format instead 😄 )

src/citra_qt/game_list.cpp
```diff
@@ -15,10 +16,150 @@
...
+    int rowCount = gamelist->tree_view->model()->rowCount();
+    QString edit_search_text = gamelist->search_field->edit_search->text().toLower();
+
+    // If the searchfield's text hasn't changed special function keys gets checked
+    // If no function key changes the searchfield's text  the filter doesn't need to get reloaded
+    if (edit_search_text == edit_search_text_old) {
+        switch (keyEvent->key()) {
+        // Escape: Resets the searchfield
+        case Qt::Key_Escape: {
+            if (edit_search_text_old == "") {
```
jroweboy: `.empty()`

src/citra_qt/game_list.cpp
```diff
@@ -15,10 +16,150 @@
...
+/**
+* Checks if all words separated by spaces are contained in another string
+* This offers a word order insensitive search function
+*
+* @param String that gets checked if it contains all words of the userinput string
+* @param String containing all words getting checked
+* @return true if containsstring contains all words of userinput
+*/
+bool GameList::SearchField::KeyReleaseEater::containsAllWords(QString containsstring, QString userinput) {
+    QStringList userinput_split = userinput.split(" ", QString::SplitBehavior::SkipEmptyParts);
```
jroweboy: something like this would be better

```cpp
return std::all_of(userinput_split.begin(), userinput_split.end(), [haystack](QString s){ return haystack.contains(s); });
```

JayFoxRox:
I don't think launching a game should clear it, but that might be subjective.

However, the menu option is kind of unintuitive. Can you bind a Ctrl+F hotkey to it?
If I compare it to the find feature in firefox or pidgin, it should also contain a close button on the search bar and maybe also show the number of found results. It should also probably be called "Filter" and not "Search" as I'd expect a search to highlight one match at a time, whereas filter would condense the list.
Ctrl+F should also refocus the filter bar.

Additionally there doesn't seem to be an active focus on any of the games, what happens when pressing enter in the search bar with a single found entry? Should it launch the game?

nicoboss:
Thanks for your recommandations. I'm working on them.
If there is only one result it's possible to launch it by pressing enter but also by clicking on it as usual.

src/citra_qt/game_list.h
```diff
@@ -26,21 +32,53 @@ class GameList : public QWidget {
...
+
+        private:
+            GameList* gamelist = nullptr;
+            QString edit_filter_text_old;
+
+        protected:
+            bool eventFilter(QObject* obj, QEvent* event);
+        };
+        void setFilterResult(int visable, int total);
+        void clear(void);
```
lioncash: `void` in the parameter list doesn't need to be specified in C++.

src/citra_qt/game_list.h
```diff
@@ -26,21 +32,53 @@ class GameList : public QWidget {
         COLUMN_COUNT, // Number of columns
     };
 
-    explicit GameList(QWidget* parent = nullptr);
+    class SearchField : public QWidget {
+    public:
+        class KeyReleaseEater : public QObject {
+        public:
+            KeyReleaseEater(GameList* gamelist);
```
lioncash: this should be a constructor marked `explicit`

src/citra_qt/game_list.h
```diff
@@ -26,21 +32,53 @@ class GameList : public QWidget {
...
     static const QStringList supported_file_extensions;
 
 signals:
     void GameChosen(QString game_path);
     void ShouldCancelWorker();
     void OpenSaveFolderRequested(u64 program_id);
 
+private slots:
+    void onTextChanged(const QString& newText);
+    void onFilterCloseClicked(void);
```
lioncash: Ditto on not needing void in the parameter list.

src/citra_qt/game_list.h
```diff
@@ -26,21 +32,53 @@ class GameList : public QWidget {
...
+
+    explicit GameList(GMainWindow* parent = nullptr);
     ~GameList() override;
 
     void PopulateAsync(const QString& dir_path, bool deep_scan);
 
     void SaveInterfaceLayout();
     void LoadInterfaceLayout();
 
+    SearchField* search_field;
```
lioncash: This should be hidden behind a getter, since it's not intended to be directly modified externally.

src/citra_qt/game_list.h
```diff
@@ -26,21 +32,53 @@ class GameList : public QWidget {
         COLUMN_COUNT, // Number of columns
     };
 
-    explicit GameList(QWidget* parent = nullptr);
+    class SearchField : public QWidget {
+    public:
+        class KeyReleaseEater : public QObject {
```
lioncash: This can be private

monkeys257:
I can't even get the emulator to work with certain roms it gives me a error saying this game isn't compatible or it gives me a loading error saying unable to load rom

src/citra_qt/game_list.cpp
```diff
@@ -15,10 +15,189 @@
...
+        }
+        default:
+            return QObject::eventFilter(obj, event);
+        }
+    }
+    edit_filter_text_old = edit_filter_text;
+    return QObject::eventFilter(obj, event);
+}
+
+void GameList::SearchField::setFilterResult(int visable, int total) {
```
baka0815: visible?

src/citra_qt/game_list.cpp
```diff
@@ -151,25 +349,26 @@ static bool HasSupportedFileExtension(const std::string& file_name) {
...
+*
+* Notice: This method is run on the UI thread because QFileSystemWatcher is not thread safe and
+* this function is fast enough to not stall the UI thread. If performance is an issue, it should
+* be moved to another thread and properly locked to prevent concurrency issues.
+*
+* @param dir folder to check for changes in
+* @param recursion 0 if recursion is disabled. Any positive number passed to this will add each
+*        directory recursively to the watcher and will update the file list if any of the folders
+*        change. The number determines how deep the recursion should traverse.
+*/
```
baka0815: Unintentional remove of spaces?

wwylele:
@baka0815 for merged PR it is better to open a new fixup PR instead of comment here. People tends to ignore comments on merged PR.

nicoboss:
I'll fix this typo and formatting mistake in a new pull request this evening.

PR #2677, opened 2017-04-26
Add BINDIR and MANDIR

orbea:
This add `BINDIR` and `MANDIR` to the `CMakeLists.txt`. This can be useful on some distros, for example in Slackware man pages are installed in `/usr/man` instead of `/usr/share/man`.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2677)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,8 @@ ELSEIF (WIN32)
     ENDIF (MINGW)
 ELSEIF (CMAKE_SYSTEM_NAME MATCHES "^(Linux|kFreeBSD|GNU|SunOS)$")
     set(PLATFORM_LIBRARIES rt)
+    set(BINDIR ${CMAKE_INSTALL_PREFIX}/bin CACHE PATH "BINDIR")
+    set(MANDIR ${CMAKE_INSTALL_PREFIX}/share/man CACHE PATH "MANDIR")
 ENDIF (APPLE)
 
 # MINGW: GCC does not support codecvt, so use iconv instead
@@ -291,11 +293,11 @@ endif()
 if(UNIX)
     if(ENABLE_SDL2)
         install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.6"
-                DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man6")
+                DESTINATION "${MANDIR}/man6")
     endif()
 
     if (ENABLE_QT)
         install(FILES "${CMAKE_SOURCE_DIR}/dist/citra-qt.6"
-                DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man6")
+                DESTINATION "${MANDIR}/man6")
     endif()
 endif()
--- a/src/citra/CMakeLists.txt
+++ b/src/citra/CMakeLists.txt
@@ -26,7 +26,7 @@ endif()
 target_link_libraries(citra ${PLATFORM_LIBRARIES} Threads::Threads)
 
 if(UNIX AND NOT APPLE)
-    install(TARGETS citra RUNTIME DESTINATION "${CMAKE_INSTALL_PREFIX}/bin")
+    install(TARGETS citra RUNTIME DESTINATION "${BINDIR}")
 endif()
 
 if (MSVC)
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -196,6 +196,8 @@ ELSEIF (WIN32)
     ENDIF (MINGW)
 ELSEIF (CMAKE_SYSTEM_NAME MATCHES "^(Linux|kFreeBSD|GNU|SunOS)$")
     set(PLATFORM_LIBRARIES rt)
+    set(BINDIR ${CMAKE_INSTALL_PREFIX}/bin CACHE PATH "BINDIR")
+    set(MANDIR ${CMAKE_INSTALL_PREFIX}/share/man CACHE PATH "MANDIR")
```
JayFoxRox: [Shouldn't this be done by `GNUInstallDirs`](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) ?

JayFoxRox: What happens for other platforms? `BINDIR` an `MANDIR` would not be set. This would install to `"/man6"` and `""`

linkmauve: We don’t install the man on other systems, but `BINDIR` would indeed be an issue, please define it in the general case too.

orbea: If they are not set:
```
usr/bin/
usr/bin/citra
usr/bin/citra-qt
usr/share/
usr/share/man/
usr/share/man/man6/
usr/share/man/man6/citra-qt.6
usr/share/man/man6/citra.6
```
If I set them like this.
```
-DBINDIR=/usr/games \
-DMANDIR=/usr/man \
```
```
usr/games/
usr/games/citra
usr/games/citra-qt
usr/man/
usr/man/man6/
usr/man/man6/citra.6
usr/man/man6/citra-qt.6
```
The behavior should not change unless they are set.

orbea: If I understand correctly `BINDIR` will only be needed for unix systems? Its already wrapped in:
```
if(UNIX AND NOT APPLE)
```
And it will be set if:
```
ELSEIF (CMAKE_SYSTEM_NAME MATCHES "^(Linux|kFreeBSD|GNU|SunOS)$")
```

orbea: To be honest, this is pretty much copy/pasted from dolphin-emu. It works fine there...

orbea: I should say I am fine changing the variable names, but last time I tried to add `CMAKE_INSTALL_DOCDIR` to a project I was told "it's better to reserve `CMAKE_xxx` variables to official CMake variables".

https://github.com/apitrace/apitrace/pull/434

But your link seems to contradict that?

yuriks: I think the point is that CMake already defines these variables itself if you use that module. So yeah, that looks like what we should be doing if that's a module that ships with CMake.

jbeich: `-lrt` conditional is limited to Solaris and glibc/uclibc, other Unix systems have `clock_gettime()` as part of `-lc` (default unless `-nostdlib`). To avoid breaking BSDs (except GNU/kFreeBSD) move `*DIR` assignments under`if(UNIX)` as well.
```
$ cmake /path/to/citra
[...]
CMake Error at src/citra/CMakeLists.txt:29 (install):
  install TARGETS given no RUNTIME DESTINATION for executable target "citra".


CMake Error at src/citra_qt/CMakeLists.txt:105 (install):
  install TARGETS given no RUNTIME DESTINATION for executable target
  "citra-qt".


-- Configuring incomplete, errors occurred!
```

yuriks:
@JayFoxRox: Do you have any issue with this that still need to be addressed?
@linkmauve: If JFR has no concerns and you're also ok with the latest version of the code feel free to merge afterwards.

JayFoxRox:
I'm fine with this (better than master), but I'd rather have it with my proposed changes.
Note that I did not test this.

bunnei:
@JayFoxRox please sign off if you're fine with this being merged, as a non-Linux user, I have limited to no preference on this change

JayFoxRox:
I don't care. I'd have to look into it again and I won't do that at the moment.

chris062689:
@JayFoxRox Has there been an update to this?  Thanks.

JayFoxRox:
No, and I don't intend to review it either. There should be enough users in the citra or cmake community who are more knowledgeable about this.

B3n30:
What's the status of this PR?

B3n30:
Closed, due to inactivity.

PR #2679, opened 2017-05-01
DynCom VFP tests

MerryMage:
This tests only vadd.f32.

Tests do not pass.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2679)


Subv:
See #2681 and #2680

yuriks:
`vfp_vadd_f32.inc` is (estimated) about 800 KB by itself... I think we're gonna need some other plan for those or store them separately, if you're planning on having one for every different instruction.

MerryMage:
They could be compressed, they compress really well, but... no.

Pareing them down to a representative sample of cases is probably the best.

Subv:
#2690 + #2691 + #2692 allow all those tests to pass in the dyncom.

src/tests/core/arm/dyncom/arm_dyncom_vfp_tests.cpp
```diff
@@ -0,0 +1,48 @@
...
+    std::vector<VfpTestCase> test_cases {{
+#include "vfp_vadd_f32.inc"
+    }};
+
+    for (const auto& test_case : test_cases) {
+        dyncom.SetPC(0);
+        dyncom.SetVFPSystemReg(VFP_FPSCR, test_case.initial_fpscr);
+        dyncom.SetVFPReg(4, test_case.a);
+        dyncom.SetVFPReg(6, test_case.b);
+        if (dyncom.GetVFPReg(2) != test_case.result || dyncom.GetVFPSystemReg(VFP_FPSCR) != test_case.final_fpscr)
```
Subv: Need to call dyncom.ExecuteInstructions(1);

src/tests/core/arm/dyncom/arm_dyncom_vfp_tests.cpp
```diff
@@ -0,0 +1,48 @@
...
+    test_env.SetMemory32(4, 0xEAFFFFFE); // b +#0
+
+    ARM_DynCom dyncom(USER32MODE);
+
+    std::vector<VfpTestCase> test_cases {{
+#include "vfp_vadd_f32.inc"
+    }};
+
+    for (const auto& test_case : test_cases) {
+        dyncom.SetPC(0);
```
Subv: You can set dyncom.down_count = 1000; here to ensure that CoreTiming will not be called

yuriks:
What's the point of the MMIO handler thing? It seems like it's trying to dynamically feed the values to the program on-demand, but why not just dump the whole thing to an array and map that into memory?

MerryMage:
@yuriks To determine if writes to memory occur and their values to any arbitrary address, which can be useful in some tests one may want to write.

wwylele:
Are all these tests passing now?

Subv:
What's the status of this? The tests should now all pass.

MerryMage:
Fixed up, tests pass.

wwylele:
Are we going to merge this as is? or as said 
>Pareing them down to a representative sample of cases is probably the best.

PR #2682, opened 2017-05-02
citra-qt: game list search function fixed minor mistakes

nicoboss:
I fixed some minor mistakes of the search filter
- Bug that made the filter result label invisible if the game list folder couldn't be found
- Formatting mistakes on some multiline comments
- Corrected typos

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2682)

### DIFF
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -239,8 +239,8 @@ void GameList::setFilterFocus() {
     search_field->setFocus();
 }
 
-void GameList::setFilterVisible(bool visablility) {
-    search_field->setVisible(visablility);
+void GameList::setFilterVisible(bool visibility) {
+    search_field->setVisible(visibility);
 }
 
 void GameList::clearFilter() {
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -45,7 +45,7 @@ bool GameList::SearchField::KeyReleaseEater::eventFilter(QObject* obj, QEvent* e
             break;
         }
         // Return and Enter
-        // If the enter key gets pressed first checks how many and which entry is visable
+        // If the enter key gets pressed first checks how many and which entry is visible
         // If there is only one result launch this game
         case Qt::Key_Return:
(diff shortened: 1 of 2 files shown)


src/citra_qt/game_list.cpp
```diff
@@ -295,6 +295,8 @@ void GameList::PopulateAsync(const QString& dir_path, bool deep_scan) {
     if (!FileUtil::Exists(dir_path.toStdString()) ||
         !FileUtil::IsDirectory(dir_path.toStdString())) {
         LOG_ERROR(Frontend, "Could not find game list folder at %s", dir_path.toLocal8Bit().data());
+        search_field->setFilterResult(0, 0);
+        search_field->setFocus();
```
yuriks: Why are you setting focus here? Won't this steal focus from elsewhere in the UI unpredictably?

nicoboss: The setFocus function only has an effect if the search field is visible and the PopulateAsync function only gets triggered on startup, changing the game directory or by modifying the dictionary externally. User usually wants to search something if the search field is visible on startup or after the game directory get changed. The reason why it also gets focused if an error occurs is that these events shouldn’t affect each other’s and so even if there is an error the search field look the same as if the game folder doesn’t contain any games. In addition, there’s nothing else that would make sense to focus while no game is running. So in my opinion it wouldn’t make much sense to remove the autofocus.

yuriks: PopulateAsync is also called when changes are detected in the directory (and you can be doing things like browsing the menus, etc. while this happens.) I don't see why someone would want to filter on an empty list? Changing the game directory also doesn't steal focus from the bar here, so I'm very confused as to what use-case this is supposed to be addressing.

nicoboss: The setFocus() command will be automatically ignored if any other thing like the settings window or the menu bar is open. Qt uses a very intelligent focus system. But I also see your point that nobody ever wants to filter an empty list and so in this case autofocus the search field doesn't make sense so I'll remove that.

PR #2683, opened 2017-05-03
Telemetry framework Part 1

bunnei:
This is the first part of a set of changes to add a telemetry framework to Citra. In the near term, this will be used for user-submitted compatibility data to our own custom service endpoint (services.citra-emu.org). My goal, however, is to make this very generic, and allow us to easily log useful/interesting data fields throughout the emulator, and then upload them to our service at a later time. In the long term, we may use this for anonymous telemetry, critical errors and crash reports, measuring how often features are used, etc.

As I mentioned above, this will be a multi-part submission. Here is what is included in this submission, as well as what is to come next:

- [x] Common interface for submitting data fields to the telemetry log
- [x] Dummy backend that does nothing with the telemetry log
- [x] Core instantiation of the telemetry interface with a single global log used for game compatibility
- [x] A few example fields are logged in core
- [ ] Implement a backend that does an authenticated log submission to our service
- [ ] Implement remaining fields for game compatibility submission

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2683)

### DIFF
--- a/src/common/telemetry.cpp
+++ b/src/common/telemetry.cpp
@@ -1,0 +1,40 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include "common/telemetry.h"
+
+namespace Telemetry {
+
+void FieldCollection::Accept(VisitorInterface& visitor) const {
+    for (const auto& field : fields) {
+        field.second->Accept(visitor);
+    }
+}
+
+void FieldCollection::AddField(std::unique_ptr<FieldInterface> field) {
+    fields[field->GetName()] = std::move(field);
+}
+
+template <class T>
+void Field<T>::Accept(VisitorInterface& visitor) const {
+    visitor.Visit(*this);
+}
+
+template class Field<bool>;
+template class Field<double>;
+template class Field<float>;
+template class Field<u8>;
+template class Field<u16>;
+template class Field<u32>;
+template class Field<u64>;
+template class Field<s8>;
+template class Field<s16>;
+template class Field<s32>;
+template class Field<s64>;
+template class Field<std::string>;
+template class Field<const char*>;
+template class Field<std::chrono::microseconds>;
+
+} // namespace Telemetry
--- a/src/common/telemetry.h
+++ b/src/common/telemetry.h
@@ -1,0 +1,196 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <chrono>
+#include <map>
+#include <memory>
+#include <string>
+#include "common/common_types.h"
+
+namespace Telemetry {
+
+/// Field type, used for grouping fields together in the final submitted telemetry log
+enum class FieldType : u8 {
+    None = 0,     ///< No specified field group
+    App,          ///< Citra application fields (e.g. version, branch, etc.)
+    Session,      ///< Emulated session fields (e.g. title ID, log, etc.)
+    Performance,  ///< Emulated performance (e.g. fps, emulated CPU speed, etc.)
+    UserFeedback, ///< User submitted feedback (e.g. star rating, user notes, etc.)
+    UserConfig,   ///< User configuration fields (e.g. emulated CPU core, renderer, etc.)
+    UserSystem,   ///< User system information (e.g. host CPU type, RAM, etc.)
+};
+
+struct VisitorInterface;
+
+/**
+ * Interface class for telemetry data fields.
+ */
+class FieldInterface : NonCopyable {
+public:
+    virtual ~FieldInterface() = default;
+
+    /**
+     * Accept method for the visitor pattern.
+     * @param visitor Reference to the visitor that will visit this field.
+     */
+    virtual void Accept(VisitorInterface& visitor) const = 0;
+
+    /**
+     * Gets the name of this field.
+     * @returns Name of this field as a string.
+     */
+    virtual const std::string& GetName() const = 0;
+};
+
+/**
+ * Represents a telemetry data field, i.e. a unit of data that gets logged and submitted to our
+ * telemetry web service.
+ */
+template <typename T>
+class Field : public FieldInterface {
+public:
+    Field(FieldType type, std::string name, const T& value)
+        : type(type), name(std::move(name)), value(value) {}
+
+    Field(FieldType type, std::string name, T&& value)
+        : type(type), name(std::move(name)), value(std::move(value)) {}
+
+    Field(const Field& other) : Field(other.type, other.name, other.value) {}
+
+    Field& operator=(const Field& other) {
+        type = other.type;
+        name = other.name;
+        value = other.value;
+        return *this;
+    }
+
+    Field& operator=(Field&& other) {
+        type = other.type;
+        name = std::move(other.name);
+        value = std::move(other.value);
+        return *this;
+    }
+
+    void Accept(VisitorInterface& visitor) const override;
+
+    const std::string& GetName() const override {
+        return name;
+    }
+
+    /**
+     * Returns the type of the field.
+     */
+    FieldType GetType() const {
+        return type;
+    }
+
+    /**
+     * Returns the value of the field.
+     */
+    const T& GetValue() const {
+        return value;
+    }
+
+    inline bool operator==(const Field<T>& other) {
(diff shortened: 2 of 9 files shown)


src/common/telemetry.h
```diff
@@ -0,0 +1,214 @@
...
+inline bool operator==(const Field<T>& left, const Field<T>& right) {
+    return left.IsEqual(right);
+}
+
+template<typename T>
+inline bool operator!=(const Field<T>& left, const Field<T>& right) {
+    return !(left == right);
+}
+
+class FieldCollection final : FieldInterface {
```
bunnei: Missing docstring

src/common/telemetry.h
```diff
@@ -0,0 +1,214 @@
...
+    virtual void Visit(const Field<u64>& field) = 0;
+    virtual void Visit(const Field<s8>& field) = 0;
+    virtual void Visit(const Field<s16>& field) = 0;
+    virtual void Visit(const Field<s32>& field) = 0;
+    virtual void Visit(const Field<s64>& field) = 0;
+    virtual void Visit(const Field<std::string>& field) = 0;
+    virtual void Visit(const Field<const char*>& field) = 0;
+    virtual void Visit(const Field<std::chrono::microseconds>& field) = 0;
+
+    /// Completion method, called once all fields have been visited,
```
bunnei: Trailing comma

src/common/telemetry.h
```diff
@@ -0,0 +1,213 @@
...
+    const T& GetValue() const {
+        return value;
+    }
+
+    /**
+     * Checks if a field is equal to another field.
+     * @param other Other field to compare this field to.
+     * @returns True if this field is equal to the specified field, otherwise false.
+     */
+    bool IsEqual(const Field<T>& other) const {
```
yuriks: What's the point of this over just defining operator==?

src/common/telemetry.h
```diff
@@ -0,0 +1,213 @@
...
+        value = other.value;
+        return *this;
+    }
+
+    Field(Field&& other) : Field(other.type, other.name, std::move(other.value)) {}
+
+    Field& operator=(Field&& other) {
+        type = other.type;
+        std::swap(name, other.name);
+        std::swap(value, other.value);
```
yuriks: Why not a std::move assignment for these instead? At worst the impl may use the swap trick underneath, and you'll avoid the pointless actual swap for the other types (which are most of the ones you use except std::string).

src/common/telemetry.h
```diff
@@ -0,0 +1,213 @@
...
+ * Represents a telemetry data field, i.e. a unit of data that gets logged and submitted to our
+ * telemetry web service.
+ */
+template <typename T>
+class Field : public FieldInterface {
+public:
+    Field(FieldType type, const std::string& name, const T& value)
+        : type(type), name(name), value(value) {}
+
+    Field(FieldType type, const std::string& name, T&& value)
```
yuriks: Why not make the name also movable? (Or pass-by-value, and then move out of it.)

src/common/telemetry.h
```diff
@@ -0,0 +1,213 @@
...
+    Field(const Field& other) : Field(other.type, other.name, other.value) {}
+
+    Field& operator=(const Field& other) {
+        type = other.type;
+        name = other.name;
+        value = other.value;
+        return *this;
+    }
+
+    Field(Field&& other) : Field(other.type, other.name, std::move(other.value)) {}
```
yuriks: Did VC 2015 support defaulted move constructors/assignment? If not, can we drop it already? :D

src/core/telemetry_session.h
```diff
@@ -0,0 +1,38 @@
...
+     * @param name Name of the field to add.
+     * @param value Value for the field to add.
+     */
+    template <typename T>
+    void AddField(Telemetry::FieldType type, const char* name, T value) {
+        field_collection.AddField(type, name, std::move(value));
+    }
+
+private:
+    Telemetry::FieldCollection field_collection; ///< Field collection, tracks all added fields
```
yuriks: `FieldCollection field_collection; /// Field collection`

(What I'm saying is, that part of the comment is probably unnecessary.)

src/core/telemetry_session.h
```diff
@@ -0,0 +1,38 @@
...
+    std::unique_ptr<Telemetry::VisitorInterface> backend; ///< Backend interface that logs fields
```
yuriks: Just a design thought: Are going to be streaming fields out? (I can't think of a good reason for us to do that.) If not, can you leave this out of this class and have something else trigger the send of a session (which then becomes kinda redundant.)

src/core/telemetry_session.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+TelemetrySession::~TelemetrySession() {
+    // Log one-time session end information
+    const auto duration{std::chrono::steady_clock::now().time_since_epoch()};
+    const auto end_time{std::chrono::duration_cast<std::chrono::microseconds>(duration).count()};
+    AddField(Telemetry::FieldType::Session, "EndTime", end_time);
+
+    // Complete the session, submitting to web service if necessary
+    field_collection.Accept(*backend);
+    backend->Complete();
```
yuriks: For a backend that does IO, this doesn't seem like the sort of thing we want to block on in a destructor. Any plan on how to make this async?

src/core/loader/ncch.cpp
```diff
@@ -339,6 +340,9 @@ ResultStatus AppLoader_NCCH::Load() {
 
     LOG_INFO(Loader, "Program ID: %016" PRIX64, ncch_header.program_id);
 
+    auto& telemetry{Core::System::GetInstance().TelemetrySession()};
```
bunnei: Yes, this log is supposed to just be for the current (core) emulation session

src/core/telemetry_session.cpp
bunnei: Probably, this isn't going to go here anyways once this actually does IO - it's just a placeholder spot to wrap up the session once the core is stopped. In the first working implementation, this will probably be a user-initiated async task kicked off from the UI. Will put a comment here.

src/core/telemetry_session.h
bunnei: The core impl is really just a stub for now - this will change in subsequent PRs. Would prefer to leave as-is for now.

src/common/telemetry.h
```diff
@@ -0,0 +1,198 @@
...
+private:
+    std::string name; ///< Field name, must be unique
+    FieldType type{}; ///< Field type, used for grouping fields together
+    T value;          ///< Field value
+};
+
+/**
+ * Collection of data fields that have been logged.
+ */
+class FieldCollection final : NonCopyable {
```
yuriks: Any particular reason why this is non-copyable?

src/common/telemetry.cpp
```diff
@@ -0,0 +1,40 @@
...
+        field.second->Accept(visitor);
+    }
+}
+
+void FieldCollection::AddField(std::unique_ptr<FieldInterface> field) {
+    fields[field->GetName()] = std::move(field);
+}
+
+template <class T>
+void Field<T>::Accept(VisitorInterface& visitor) const {
```
yuriks: Not clear to me why *this* method specifically is implemented here in the cpp, while everything else is right in the header. 
Was your intention to restrict instantiation of Field to be only with T as the types below? If so, wouldn't it be better to do this for the constructors, preventing the objects from being constructed in the first place, instead of giving a potentially confusing error message where Accept is called.

src/common/telemetry.h
```diff
@@ -0,0 +1,198 @@
...
+
+    void Accept(VisitorInterface& visitor) const override;
+
+    const std::string& GetName() const override {
+        return name;
+    }
+
+    /**
+     * Gets the type of the field.
+     * @returns Returns the type of the field.
```
yuriks: Just make the text in the @return the main summary. Same below

src/common/telemetry.cpp
bunnei: I believe we chatted about this already - but this was deliberate because `Accept` (which uses `VisitorInterface`) is a virtual member of `Field<T>`'s base class `FieldInterface`, and `VisitorInterface` uses `Field<T>`... mutual dependencies and stuff.

src/common/telemetry.h
bunnei: Because I didn't envision it being copyable. If someone wants to copy it, they can get a compiler error and determine if this should be removed

PR #2686, opened 2017-05-03
pica: use correct coordinates for texture 2

wwylele:
See [here](https://www.3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_TEXUNIT_CONFIG) for documentation of the register in question. We currently only respect the low 3 bits. Texture 2 coordinates should be chosen according to bit 13.

[Here is a hardware test](https://gist.github.com/wwylele/fb8306a1facbff5fe77f5db7142f78ef) adapted from the "graphics/gpu/textured_cube" example program. In the test, the texture is bind as texture 2, and the shader outputs the original texture coordinates to coord2, and the negated coordinates to coord1. Press A to flip bit 13 of the register, and texture flipping, as the result of switching coordinates, can be observed.

Don't know if this fix any games.

### DIFF
--- a/src/video_core/regs_texturing.h
+++ b/src/video_core/regs_texturing.h
@@ -126,7 +126,7 @@ struct TexturingRegs {
         BitField<10, 1, u32> texture3_enable;     // TODO: unimplemented
         BitField<13, 1, u32> texture2_use_coord1;
         BitField<16, 1, u32> clear_texture_cache; // TODO: unimplemented
-    };
+    } main_config;
     TextureConfig texture0;
     INSERT_PADDING_WORDS(0x8);
     BitField<0, 4, TextureFormat> texture0_format;
@@ -146,9 +146,9 @@ struct TexturingRegs {
     };
     const std::array<FullTextureConfig, 3> GetTextures() const {
         return {{
-            {texture0_enable.ToBool(), texture0, texture0_format},
-            {texture1_enable.ToBool(), texture1, texture1_format},
-            {texture2_enable.ToBool(), texture2, texture2_format},
+            {main_config.texture0_enable.ToBool(), texture0, texture0_format},
+            {main_config.texture1_enable.ToBool(), texture1, texture1_format},
+            {main_config.texture2_enable.ToBool(), texture2, texture2_format},
         }};
     }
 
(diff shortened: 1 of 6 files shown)


wwylele:
btw, we should take some time to consolidate our pica register layout according to 3dbrew documentation. Would be great if we can have some easy way to detect if a game is accessing an unimplemented register

src/video_core/regs_texturing.h
```diff
@@ -122,6 +122,10 @@ struct TexturingRegs {
         BitField<0, 1, u32> texture0_enable;
         BitField<1, 1, u32> texture1_enable;
         BitField<2, 1, u32> texture2_enable;
+        BitField<8, 2, u32> texture3_coordinates;
+        BitField<10, 1, u32> texture3_enable;
+        BitField<13, 1, u32> texture2_use_coord1;
+        BitField<16, 1, u32> clear_texture_cache;
```
JayFoxRox: This is dangerous: adding unused registers will probably cause lot of confusion when finding remaining broken features

src/video_core/regs_texturing.h
```diff
@@ -122,6 +122,10 @@ struct TexturingRegs {
         BitField<0, 1, u32> texture0_enable;
         BitField<1, 1, u32> texture1_enable;
         BitField<2, 1, u32> texture2_enable;
+        BitField<8, 2, u32> texture3_coordinates;
```
JayFoxRox: Please also add support for this while you are at it

JayFoxRox:
This feature is necessary for Kid Icarus. I've a blog article lined up for this.

My implementation from December (rebasing it right now) can be found at https://github.com/JayFoxRox/citra/pull/47

wwylele:
Good point, I am trying to figure out the bump mapping stuff. My first guess is yes.
Just want to make sure: there is not other place using textures, right?

wwylele:
Fixed bump map selector in hw renderer (sw one doesn't implement fragment lighting). Still left texcoord3 unimplemented because procedural texture sampling might go through a complete different routine in the future implementation.

wwylele:
hold on, I forgot to make `shader_dirty = true`

wwylele:
Update: added `shader_dirty=true` for texture2 coordinates change

Issue #2681, opened 2017-05-01
VFP: Invalid sentinel FPSCR values are propagated to guest registers.


Subv:
vfp_raise_exceptions is not removing the sentinel VFP_NAN_FLAG value before updating the FPSCR, resulting in the flag FPSCR_IOE being set when propagating quiet nans (as the flag and the sentinel have the same value).

I propose we get rid of VFP_NAN_FLAG altogether and just return 0 in the cases where we don't need to raise an FP exception. (https://github.com/citra-emu/citra/blob/master/src/core/arm/skyeye_common/vfp/vfpsingle.cpp#L260 vfp_propagate_nan)

This, coupled with what's proposed in #2680 should be enough to prevent erroneous values in the fpscr.

PR #2690, opened 2017-05-08
Dyncom/VFP: Properly pack QNaN values.

### LINKED ISSUE #2680: VFP: Incorrect writing of QNaNs in Dyncom
After https://github.com/citra-emu/citra/commit/edf30d84cc0e8299d61c98f5bb40a6428d1576bc , qnans are not being properly packed into registers in this check in vfp_single_normaliseround: https://github.com/citra-emu/citra/blob/master/src/core/arm/skyeye_common/vfp/vfpsingle.cpp#L96 .

This results in qnans being converted into numbers, ie, 7fc00000 (qnan) becomes 7f000000 which is 1.7014118e+38.

A proposed fix is to replace that check with this:

````
int ts = vfp_single_type(vs);

    /*
     * Infinities and NaNs are a special case.
     */
    if (ts & (VFP_NAN | VFP_INFINITY))
        goto pack;
````

After performing this change, the relevant tests from #2679 pass.

I did not check if the double precision code also has this problem, i suspect it does.

Subv:
vfp_*_normaliseround was previously converting qnans into numbers, ie, 7fc00000 (qnan) into 7f000000 (1.7014118e+38).

Closes #2680

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2690)

### DIFF
--- a/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
+++ b/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
@@ -90,10 +90,12 @@ u32 vfp_double_normaliseround(ARMul_State* state, int dd, struct vfp_double* vd,
 
     vfp_double_dump("pack: in", vd);
 
+    int td = vfp_double_type(vd);
+
     /*
      * Infinities and NaNs are a special case.
      */
-    if (vd->exponent == 2047 && (vd->significand == 0 || exceptions))
+    if (td & (VFP_NAN | VFP_INFINITY))
         goto pack;
 
     /*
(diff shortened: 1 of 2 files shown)


Subv:
This creates a bug in the vfp_single_multiply function in the case where the sum of the exponents of both operands becomes 0xFF, the number is thus interpreted as NaN.

A possible fix is to perform the normalization inside vfp_single_multiply, before calling vfp_single_normaliseround.

Subv:
At this point i'm thinking it's better if we just revert edf30d84cc0e8299d61c98f5bb40a6428d1576bc and fix #2681 by masking out the VFP_NAN_FLAG in vfp_single_cpdo.

Any thoughts?

Subv:
Superseded by #2696

PR #2691, opened 2017-05-08
Dyncom/VFP: Do not raise sentinel values as FPSCR exceptions.

### LINKED ISSUE #2681: VFP: Invalid sentinel FPSCR values are propagated to guest registers.

Subv:
NaNs are handled inside vfp_*_normaliseround.
This was causing a trap enable bit in the FPSCR to be set when a QNaN was encountered.

Closes #2681

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2691)

### DIFF
--- a/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
+++ b/src/core/arm/skyeye_common/vfp/vfpdouble.cpp
@@ -253,7 +253,7 @@ static u32 vfp_propagate_nan(struct vfp_double* vdd, struct vfp_double* vdn, str
     /*
      * If one was a signalling NAN, raise invalid operation.
      */
-    return tn == VFP_SNAN || tm == VFP_SNAN ? FPSCR_IOC : VFP_NAN_FLAG;
+    return tn == VFP_SNAN || tm == VFP_SNAN ? FPSCR_IOC : 0;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


yuriks:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2691#-:-KjfGzIJSUa-mQz-1vwG:bnfp4nl)*

yuriks:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2691)*

Subv:
Superseded by #2696

PR #2695, opened 2017-05-08
Prepare Pica registers for Geometry Shaders

JayFoxRox:
This code is part of the geometry shader (GS) implementation written by me which is based on work by @ds84182 .

This PR shouldn't change any behaviour currently (except maybe for cases where GS is active, but those are bugged anyway). It merely prepares the registers for GS.
I'd recommend inclusion in bleeding-edge to figure out if there are any issues.
However, this went through a larger refactor shortly before creating this PR, so review / testing should be done first.

Review per commit; see commit description too.

---

### Notes:

* The limit for VS (Vertex Shaders) appears to be 512 program words (compared to 4096 for GS). We should probably introduce a template parameter or something to handle these differences between VS / GS structure. However, that's beyond the scope of this PR. I can create an issue for this.

* [`vsh_com_mode`](https://3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_VSH_COM_MODE) will later (next PR) be indicating if GS will use the same shader data (uniform / program / swizzle data) as VS.
It is not yet known if this influences writes (the functions introduced in this commit) OR reads (in the interpreter and jit).
We also don't know wether the shader data is per shader mode (VS / GS) or per shader-unit (4 units, only one of which is able to run GS).
However, this should be easy to change in the future if the current implementation is wrong (handling it on read = not doing extra work in functions introduced by this PR).
*We need a couple of hardware tests to understand how exactly the shader data is uploaded. However, I vote for accepting this for now, and refactoring later (once hardware tests have been done).*

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2695)

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -422,6 +422,58 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
         break;
     }
 
+    case PICA_REG_INDEX(gs.bool_uniforms):
+        WriteUniformBoolReg(g_state.gs, value);
+        break;
+
+    case PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[0], 0x281):
+    case PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[1], 0x282):
+    case PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[2], 0x283):
+    case PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[3], 0x284): {
+        unsigned index = (id - PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[0], 0x281));
+        auto values = regs.gs.int_uniforms[index];
+        WriteUniformIntReg(g_state.gs, index,
+                           Math::Vec4<u8>(values.x, values.y, values.z, values.w));
+        break;
+    }
+
+    case PICA_REG_INDEX_WORKAROUND(gs.uniform_setup.set_value[0], 0x291):
+    case PICA_REG_INDEX_WORKAROUND(gs.uniform_setup.set_value[1], 0x292):
+    case PICA_REG_INDEX_WORKAROUND(gs.uniform_setup.set_value[2], 0x293):
+    case PICA_REG_INDEX_WORKAROUND(gs.uniform_setup.set_value[3], 0x294):
+    case PICA_REG_INDEX_WORKAROUND(gs.uniform_setup.set_value[4], 0x295):
+    case PICA_REG_INDEX_WORKAROUND(gs.uniform_setup.set_value[5], 0x296):
+    case PICA_REG_INDEX_WORKAROUND(gs.uniform_setup.set_value[6], 0x297):
+    case PICA_REG_INDEX_WORKAROUND(gs.uniform_setup.set_value[7], 0x298): {
+        WriteUniformFloatReg(g_state.regs.gs, g_state.gs, gs_float_regs_counter,
+                             gs_uniform_write_buffer, value);
+        break;
+    }
+
+    case PICA_REG_INDEX_WORKAROUND(gs.program.set_word[0], 0x29c):
+    case PICA_REG_INDEX_WORKAROUND(gs.program.set_word[1], 0x29d):
+    case PICA_REG_INDEX_WORKAROUND(gs.program.set_word[2], 0x29e):
+    case PICA_REG_INDEX_WORKAROUND(gs.program.set_word[3], 0x29f):
+    case PICA_REG_INDEX_WORKAROUND(gs.program.set_word[4], 0x2a0):
+    case PICA_REG_INDEX_WORKAROUND(gs.program.set_word[5], 0x2a1):
+    case PICA_REG_INDEX_WORKAROUND(gs.program.set_word[6], 0x2a2):
+    case PICA_REG_INDEX_WORKAROUND(gs.program.set_word[7], 0x2a3): {
+        WriteProgramCode(g_state.regs.gs, g_state.gs, 4096, value);
+        break;
+    }
+
+    case PICA_REG_INDEX_WORKAROUND(gs.swizzle_patterns.set_word[0], 0x2a6):
+    case PICA_REG_INDEX_WORKAROUND(gs.swizzle_patterns.set_word[1], 0x2a7):
+    case PICA_REG_INDEX_WORKAROUND(gs.swizzle_patterns.set_word[2], 0x2a8):
+    case PICA_REG_INDEX_WORKAROUND(gs.swizzle_patterns.set_word[3], 0x2a9):
+    case PICA_REG_INDEX_WORKAROUND(gs.swizzle_patterns.set_word[4], 0x2aa):
+    case PICA_REG_INDEX_WORKAROUND(gs.swizzle_patterns.set_word[5], 0x2ab):
+    case PICA_REG_INDEX_WORKAROUND(gs.swizzle_patterns.set_word[6], 0x2ac):
+    case PICA_REG_INDEX_WORKAROUND(gs.swizzle_patterns.set_word[7], 0x2ad): {
+        WriteSwizzlePatterns(g_state.regs.gs, g_state.gs, value);
+        break;
+    }
+
     case PICA_REG_INDEX(vs.bool_uniforms):
         WriteUniformBoolReg(g_state.vs, value);
         break;
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -32,12 +32,13 @@ namespace Pica {
 
 namespace CommandProcessor {
 
-static int float_regs_counter = 0;
+static int vs_float_regs_counter = 0;
+static u32 vs_uniform_write_buffer[4];
 
-static u32 uniform_write_buffer[4];
+static int gs_float_regs_counter = 0;
+static u32 gs_uniform_write_buffer[4];
 
 static int default_attr_counter = 0;
-
 static u32 default_attr_write_buffer[3];
 
 // Expand a 4-bit mask to 4-byte mask, e.g. 0b0101 -> 0x00FF00FF
@@ -47,6 +48,97 @@ static const u32 expand_bits_to_bytes[] = {
 };
 
 MICROPROFILE_DEFINE(GPU_Drawing, "GPU", "Drawing", MP_RGB(50, 50, 240));
+
+static const char* GetShaderSetupTypeName(Shader::ShaderSetup& setup) {
+    if (&setup == &g_state.vs) {
+        return "vertex shader";
+    }
+    if (&setup == &g_state.gs) {
+        return "geometry shader";
+    }
+    return "unknown shader";
+}
+
+static void WriteUniformBoolReg(Shader::ShaderSetup& setup, u32 value) {
+    for (unsigned i = 0; i < setup.uniforms.b.size(); ++i)
+        setup.uniforms.b[i] = (value & (1 << i)) != 0;
+}
+
+static void WriteUniformIntReg(Shader::ShaderSetup& setup, unsigned index,
+                               const Math::Vec4<u8>& values) {
+    ASSERT(index < setup.uniforms.i.size());
+    setup.uniforms.i[index] = values;
+    LOG_TRACE(HW_GPU, "Set %s integer uniform %d to %02x %02x %02x %02x",
+              GetShaderSetupTypeName(setup), index, values.x, values.y, values.z, values.w);
+}
+
+static void WriteUniformFloatReg(ShaderRegs& config, Shader::ShaderSetup& setup,
+                                 int& float_regs_counter, u32 uniform_write_buffer[4], u32 value) {
+    auto& uniform_setup = config.uniform_setup;
+
+    // TODO: Does actual hardware indeed keep an intermediate buffer or does
+    //       it directly write the values?
+    uniform_write_buffer[float_regs_counter++] = value;
+
+    // Uniforms are written in a packed format such that four float24 values are encoded in
+    // three 32-bit numbers. We write to internal memory once a full such vector is
+    // written.
+    if ((float_regs_counter >= 4 && uniform_setup.IsFloat32()) ||
+        (float_regs_counter >= 3 && !uniform_setup.IsFloat32())) {
+        float_regs_counter = 0;
+
+        auto& uniform = setup.uniforms.f[uniform_setup.index];
+
+        if (uniform_setup.index >= 96) {
+            LOG_ERROR(HW_GPU, "Invalid %s float uniform index %d", GetShaderSetupTypeName(setup),
+                      (int)uniform_setup.index);
+        } else {
+
+            // NOTE: The destination component order indeed is "backwards"
+            if (uniform_setup.IsFloat32()) {
+                for (auto i : {0, 1, 2, 3})
+                    uniform[3 - i] = float24::FromFloat32(*(float*)(&uniform_write_buffer[i]));
+            } else {
+                // TODO: Untested
+                uniform.w = float24::FromRaw(uniform_write_buffer[0] >> 8);
+                uniform.z = float24::FromRaw(((uniform_write_buffer[0] & 0xFF) << 16) |
+                                             ((uniform_write_buffer[1] >> 16) & 0xFFFF));
+                uniform.y = float24::FromRaw(((uniform_write_buffer[1] & 0xFFFF) << 8) |
+                                             ((uniform_write_buffer[2] >> 24) & 0xFF));
+                uniform.x = float24::FromRaw(uniform_write_buffer[2] & 0xFFFFFF);
+            }
+
+            LOG_TRACE(HW_GPU, "Set %s float uniform %x to (%f %f %f %f)",
+                      GetShaderSetupTypeName(setup), (int)uniform_setup.index,
+                      uniform.x.ToFloat32(), uniform.y.ToFloat32(), uniform.z.ToFloat32(),
+                      uniform.w.ToFloat32());
+
+            // TODO: Verify that this actually modifies the register!
+            uniform_setup.index.Assign(uniform_setup.index + 1);
+        }
+    }
+}
+
+static void WriteProgramCode(ShaderRegs& config, Shader::ShaderSetup& setup,
+                             unsigned max_program_code_length, u32 value) {
+    if (config.program.offset >= max_program_code_length) {
+        LOG_ERROR(HW_GPU, "Invalid %s program offset %d", GetShaderSetupTypeName(setup),
+                  (int)config.program.offset);
+    } else {
+        setup.program_code[config.program.offset] = value;
+        config.program.offset++;
+    }
+}
+
+static void WriteSwizzlePatterns(ShaderRegs& config, Shader::ShaderSetup& setup, u32 value) {
+    if (config.swizzle_patterns.offset >= setup.swizzle_data.size()) {
+        LOG_ERROR(HW_GPU, "Invalid %s swizzle pattern offset %d", GetShaderSetupTypeName(setup),
+                  (int)config.swizzle_patterns.offset);
+    } else {
+        setup.swizzle_data[config.swizzle_patterns.offset] = value;
+        config.swizzle_patterns.offset++;
+    }
+}
 
 static void WritePicaReg(u32 id, u32 value, u32 mask) {
     auto& regs = g_state.regs;
@@ -331,20 +423,17 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
     }
 
     case PICA_REG_INDEX(vs.bool_uniforms):
-        for (unsigned i = 0; i < 16; ++i)
-            g_state.vs.uniforms.b[i] = (regs.vs.bool_uniforms.Value() & (1 << i)) != 0;
-
+        WriteUniformBoolReg(g_state.vs, value);
         break;
 
     case PICA_REG_INDEX_WORKAROUND(vs.int_uniforms[0], 0x2b1):
     case PICA_REG_INDEX_WORKAROUND(vs.int_uniforms[1], 0x2b2):
     case PICA_REG_INDEX_WORKAROUND(vs.int_uniforms[2], 0x2b3):
     case PICA_REG_INDEX_WORKAROUND(vs.int_uniforms[3], 0x2b4): {
-        int index = (id - PICA_REG_INDEX_WORKAROUND(vs.int_uniforms[0], 0x2b1));
+        unsigned index = (id - PICA_REG_INDEX_WORKAROUND(vs.int_uniforms[0], 0x2b1));
         auto values = regs.vs.int_uniforms[index];
-        g_state.vs.uniforms.i[index] = Math::Vec4<u8>(values.x, values.y, values.z, values.w);
-        LOG_TRACE(HW_GPU, "Set integer uniform %d to %02x %02x %02x %02x", index, values.x.Value(),
-                  values.y.Value(), values.z.Value(), values.w.Value());
+        WriteUniformIntReg(g_state.vs, index,
+                           Math::Vec4<u8>(values.x, values.y, values.z, values.w));
         break;
     }
 
@@ -356,51 +445,11 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
     case PICA_REG_INDEX_WORKAROUND(vs.uniform_setup.set_value[5], 0x2c6):
     case PICA_REG_INDEX_WORKAROUND(vs.uniform_setup.set_value[6], 0x2c7):
     case PICA_REG_INDEX_WORKAROUND(vs.uniform_setup.set_value[7], 0x2c8): {
-        auto& uniform_setup = regs.vs.uniform_setup;
-
-        // TODO: Does actual hardware indeed keep an intermediate buffer or does
-        //       it directly write the values?
-        uniform_write_buffer[float_regs_counter++] = value;
-
-        // Uniforms are written in a packed format such that four float24 values are encoded in
-        // three 32-bit numbers. We write to internal memory once a full such vector is
-        // written.
-        if ((float_regs_counter >= 4 && uniform_setup.IsFloat32()) ||
-            (float_regs_counter >= 3 && !uniform_setup.IsFloat32())) {
-            float_regs_counter = 0;
-
-            auto& uniform = g_state.vs.uniforms.f[uniform_setup.index];
-
-            if (uniform_setup.index > 95) {
-                LOG_ERROR(HW_GPU, "Invalid VS uniform index %d", (int)uniform_setup.index);
-                break;
-            }
-
-            // NOTE: The destination component order indeed is "backwards"
-            if (uniform_setup.IsFloat32()) {
-                for (auto i : {0, 1, 2, 3})
-                    uniform[3 - i] = float24::FromFloat32(*(float*)(&uniform_write_buffer[i]));
-            } else {
-                // TODO: Untested
-                uniform.w = float24::FromRaw(uniform_write_buffer[0] >> 8);
-                uniform.z = float24::FromRaw(((uniform_write_buffer[0] & 0xFF) << 16) |
-                                             ((uniform_write_buffer[1] >> 16) & 0xFFFF));
-                uniform.y = float24::FromRaw(((uniform_write_buffer[1] & 0xFFFF) << 8) |
-                                             ((uniform_write_buffer[2] >> 24) & 0xFF));
-                uniform.x = float24::FromRaw(uniform_write_buffer[2] & 0xFFFFFF);
-            }
-
-            LOG_TRACE(HW_GPU, "Set uniform %x to (%f %f %f %f)", (int)uniform_setup.index,
-                      uniform.x.ToFloat32(), uniform.y.ToFloat32(), uniform.z.ToFloat32(),
-                      uniform.w.ToFloat32());
-
-            // TODO: Verify that this actually modifies the register!
-            uniform_setup.index.Assign(uniform_setup.index + 1);
-        }
+        WriteUniformFloatReg(g_state.regs.vs, g_state.vs, vs_float_regs_counter,
+                             vs_uniform_write_buffer, value);
         break;
     }
 
-    // Load shader program code
     case PICA_REG_INDEX_WORKAROUND(vs.program.set_word[0], 0x2cc):
     case PICA_REG_INDEX_WORKAROUND(vs.program.set_word[1], 0x2cd):
     case PICA_REG_INDEX_WORKAROUND(vs.program.set_word[2], 0x2ce):
@@ -409,12 +458,10 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
     case PICA_REG_INDEX_WORKAROUND(vs.program.set_word[5], 0x2d1):
     case PICA_REG_INDEX_WORKAROUND(vs.program.set_word[6], 0x2d2):
     case PICA_REG_INDEX_WORKAROUND(vs.program.set_word[7], 0x2d3): {
-        g_state.vs.program_code[regs.vs.program.offset] = value;
-        regs.vs.program.offset++;
+        WriteProgramCode(g_state.regs.vs, g_state.vs, 512, value);
         break;
     }
 
-    // Load swizzle pattern data
     case PICA_REG_INDEX_WORKAROUND(vs.swizzle_patterns.set_word[0], 0x2d6):
     case PICA_REG_INDEX_WORKAROUND(vs.swizzle_patterns.set_word[1], 0x2d7):
     case PICA_REG_INDEX_WORKAROUND(vs.swizzle_patterns.set_word[2], 0x2d8):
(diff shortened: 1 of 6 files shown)


Subv:
```
Note that some of the registers will be unused for a while (until the next PR at least). Shall I mark them with // Unused or something?
```
IMO, no

src/video_core/command_processor.cpp
```diff
@@ -330,24 +326,90 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+    case PICA_REG_INDEX(gs.bool_uniforms):
+        Shader::WriteUniformBoolReg(true, value);
+        break;
+
+    case PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[0], 0x281):
+    case PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[1], 0x282):
+    case PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[2], 0x283):
+    case PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[3], 0x284): {
+        unsigned index = (id - PICA_REG_INDEX_WORKAROUND(gs.int_uniforms[0], 0x281));
+        auto values = regs.gs.int_uniforms[index];
```
Subv: How is this different from the "value" passed as parameter to this function?

src/video_core/shader/shader.cpp
```diff
@@ -109,6 +109,124 @@ void Shutdown() {
 #endif // ARCHITECTURE_x86_64
 }
 
+bool SharedGS() {
```
Subv: Can we give this a better name and some documentation?

src/video_core/shader/shader.cpp
```diff
@@ -109,6 +109,124 @@ void Shutdown() {
...
+    const char* shader_type = gs ? "GS" : "VS";
+    auto& setup = gs ? g_state.gs : g_state.vs;
+
+    ASSERT(index < setup.uniforms.i.size());
+    setup.uniforms.i[index] = values;
+    LOG_TRACE(HW_GPU, "Set %s integer uniform %d to %02x %02x %02x %02x", shader_type, index,
+              values.x.Value(), values.y.Value(), values.z.Value(), values.w.Value());
+}
+
+void WriteUniformFloatSetupReg(bool gs, u32 value) {
```
Subv: Is this function really needed? afaik the WritePicaReg function will write the value anyways.

src/video_core/shader/shader.cpp
```diff
@@ -109,6 +109,124 @@ void Shutdown() {
...
+    auto& uniform_setup = config.uniform_setup;
+    auto& uniform_write_buffer = setup.uniform_write_buffer;
+    auto& float_regs_counter = setup.float_regs_counter;
+
+    // TODO: Does actual hardware indeed keep an intermediate buffer or does
+    //       it directly write the values?
+    uniform_write_buffer[float_regs_counter++] = value;
+
+    // Uniforms are written in a packed format such that four float24 values are
+    // encoded in
```
Subv: what's with this break?

src/video_core/shader/shader.cpp
```diff
@@ -109,6 +109,124 @@ void Shutdown() {
...
+                      (int)uniform_setup.index, uniform.x.ToFloat32(), uniform.y.ToFloat32(),
+                      uniform.z.ToFloat32(), uniform.w.ToFloat32());
+
+            // TODO: Verify that this actually modifies the register!
+            uniform_setup.index.Assign(uniform_setup.index + 1);
+        }
+    }
+}
+
+void WriteProgramCodeOffset(bool gs, u32 value) {
```
Subv: same as WriteUniformFloatSetupReg

j-selby:
> The limit for VS (Vertex Shaders) appears to be 512 program words. We should probably introduce a template parameter or something to handle these differences between VS / GS structure. However, that's beyond the scope of this PR. I can create an issue for this.

Imo, this should definitely be recorded. It might be better to add a `TODO: Handle difference between GS and VS`. That might be a good introduction project for someone as well, so yeah, create a issue.

> Another little annoyance with this commit is that the functions take a bool as first argument for false = VS / true = GS. Shall I change that to an enum class; something else; or leave as-is?

enum - `bool gs` really doesn't look great on paper, though that's the Java programmer in me speaking :P

src/video_core/shader/shader.cpp
JayFoxRox: Until some hours ago these functions did a write-through / copy to GS registers on VS writes.
Depending on how the hardware-tests turns out this will also be re-added later.

src/video_core/shader/shader.cpp
JayFoxRox: good question.. I assume clang-format broke it. - I'll fix this

src/video_core/regs_pipeline.h
```diff
@@ -200,7 +200,14 @@ struct PipelineRegs {
     /// Number of input attributes to the vertex shader minus 1
     BitField<0, 4, u32> max_input_attrib_index;
 
-    INSERT_PADDING_WORDS(2);
+    INSERT_PADDING_WORDS(1);
+
+    enum class VSComMode : u32 {
+        Shared = 0,
+        Exclusive = 1
+    };
```
JayFoxRox: Clang format proposes `enum class VSComMode : u32 { Shared = 0, Exclusive = 1 };` but I don't really like that (it's less readable imho). I'm also thinking that I've seen code like this in the past (and even copied it from elsewhere in Citra).

What should I change it into?

yuriks: Add a comma after the last item, which will preserve the multi-line formatting.

yuriks:
You should add your commit descriptions to the actual commit messages.

---

Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[src/video_core/regs_pipeline.h, line 205 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjfc7FYmYkx97lCTAiI:-Kjfc7FZU-UfuxYQ7zit:b61d1vr) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/regs_pipeline.h#L205)):*
> ```objc
>     INSERT_PADDING_WORDS(1);
> 
>     enum class VSComMode : u32 {
> ```

What's a "com mode"? "Command mode"? Make this name clearer.

---

*[src/video_core/regs_pipeline.h, line 207 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfcPSq2WyUtey7Lb7s:-KjfcPSq2WyUtey7Lb7t:b-3fttpi) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/regs_pipeline.h#L207)):*
> ```objc
>     enum class VSComMode : u32 {
>         Shared = 0,
>         Exclusive = 1
> ```

Also document what each of these values means. This whole reg needs some documentation.

---

*[src/video_core/shader/shader.cpp, line 112 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kje9SuJ0eryT4JCV7Fb-r1-112:-KjfguDKnrg01tRfnX1l:b-v6sxld) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L112)):*

Can we give this a better name and some documentation?
</blockquote></details>

This doesn't even seem to be used?

---

*[src/video_core/shader/shader.cpp, line 116 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfeYcg6igOSIzb_vGo:-KjfeYcg6igOSIzb_vGp:b8xh5n9) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L116)):*
> ```cpp
> }
> 
> void WriteUniformBoolReg(bool gs, u32 value) {
> ```

50000% makes this an enum and not a bool. The bool is pretty much impossible to read on the call-site.

Or even better: in most cases where you just use the parameter to decide between which state or regs struct to use, just pass those in as parameters.

---

*[src/video_core/shader/shader.cpp, line 119 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjffOLo50lKrDO7TOsV:-KjffOLo50lKrDO7TOsW:bpctyv3) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L119)):*
> ```cpp
>     auto& setup = gs ? g_state.gs : g_state.vs;
> 
>     ASSERT(setup.uniforms.b.size() == 16);
> ```

Shouldn't this be a static_assert?

---

*[src/video_core/shader/shader.cpp, line 134 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kje9SuJ0eryT4JCV7Fb-r1-134:-KjfflKKPnFl4dDaPEGF:b-oa3djr) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L134)):*
<details><summary><i>Previously, JayFoxRox (Jannik Vogel) wrote…</i></summary><blockquote>

Until some hours ago these functions did a write-through / copy to GS registers on VS writes.
Depending on how the hardware-tests turns out this will also be re-added later.
</blockquote></details>

I say leave this function out until there comes a time when it becomes necessary, since right now it's not actually saving any code or work.

---

*[src/video_core/shader/shader.cpp, line 159 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfgGSHCALC1U_l7rSl:-KjfgGSI0U4uhMq9lJ-z:b-lqgcvz) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L159)):*
> ```cpp
>     // written.
>     if ((float_regs_counter >= 4 && uniform_setup.IsFloat32()) ||
>         (float_regs_counter >= 3 && !uniform_setup.IsFloat32())) {
> ```

`if (float_regs_counter >= (uniform_setup.IsFloat32() ? 4 : 3)) {`, maybe make the latter part a separate variable.

---

*[src/video_core/shader/shader.cpp, line 211 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjfh4A9doctW9mdxomv:-Kjfh4AAZneXiRzwsxf1:bw47sfw) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L211)):*
> ```cpp
> }
> 
> void WriteSwizzlePatternsOffset(bool gs, u32 value) {
> ```

Same as the other trivial funcs pointed out by Subv

---

*[src/video_core/shader/shader.h, line 148 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfiqORCxfRmxK0MzHU:-KjfXm92V2viUXhEaIU_:b-4w0xof) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.h#L148)):*
> ```objc
> 
>     int float_regs_counter = 0;
>     u32 uniform_write_buffer[4];
> ```

This doesn't seem like a good place to put these. Have you verified that both units have a separate counter/buffer for this? If not, then this seems like a premature change to make.

---

*[src/video_core/shader/shader.h, line 192 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjfj9dl3_CQdjYrBKo_:-Kjfj9dl3_CQdjYrBKoa:b-akp0ct) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.h#L192)):*
> ```objc
> void WriteProgramCode(bool gs, u32 value);
> void WriteSwizzlePatternsOffset(bool gs, u32 value);
> void WriteSwizzlePatterns(bool gs, u32 value);
> ```

IMO these should go into command_processor.cpp, or be split between that (incoming value parsing) and member vars in the relevant structs (applying the value, iff it's not trivial)

---

*[src/video_core/shader/shader_jit_x64_compiler.h, line 26 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfhcJeKCKx996eW3ji:-KjfhcJeKCKx996eW3jj:b15xzek) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader_jit_x64_compiler.h#L26)):*
> ```objc
> 
> /// Memory allocated for each compiled shader (256Kb)
> constexpr size_t MAX_SHADER_SIZE = 1024 * 256;
> ```

Probably `4096 * 64` makes more sense? Also, maybe you should introduce a constant for the maximum shader size so it's just one place to change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695#-:-KjfkkoC35nXey3I_tW6:bv3833y)*

src/video_core/command_processor.cpp
JayFoxRox: No idea. This was naively copied from the `vs.int_uniforms` case.

JayFoxRox:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[src/video_core/regs_pipeline.h, line 205 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjfc7FYmYkx97lCTAiI:-Kjg22vIGVvfjau4r-mg:b-vv7tla) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/regs_pipeline.h#L205)):*

What's a "com mode"? "Command mode"? Make this name clearer.
</blockquote></details>

This is the name from https://3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_VSH_COM_MODE .

The existing documentation is very vague about this and I'm not sure what the com stands for myself tbh..

---

*[src/video_core/shader/shader.cpp, line 112 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kje9SuJ0eryT4JCV7Fb-r1-112:-Kjg2YsTc9_eIGuKMeDg:bwbp4bg) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L112)):*

This doesn't even seem to be used?
</blockquote></details>

I can remove `VSComMode` and `SharedGS()` temporarily and add them later as they are currently unused.

---

*[src/video_core/shader/shader.cpp, line 116 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfeYcg6igOSIzb_vGo:-Kjg2sTxHQ8v6Mq_BCxw:b-g92sez) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L116)):*

50000% makes this an enum and not a bool. The bool is pretty much impossible to read on the call-site.

Or even better: in most cases where you just use the parameter to decide between which state or regs struct to use, just pass those in as parameters.
</blockquote></details>

- Will do.
- Will do.

---

*[src/video_core/shader/shader.cpp, line 119 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjffOLo50lKrDO7TOsV:-Kjg33_42O3wic9qK0Gr:b-i2ii) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L119)):*

Shouldn't this be a static_assert?
</blockquote></details>

Probably.

---

*[src/video_core/shader/shader.cpp, line 159 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfgGSHCALC1U_l7rSl:-Kjg3FGIBEn4Obdq6ynn:b5p0jno) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L159)):*

`if (float_regs_counter >= (uniform_setup.IsFloat32() ? 4 : 3)) {`, maybe make the latter part a separate variable.
</blockquote></details>

This is existing code which was just moved.
I don't want to refactor too much at this point.

---

*[src/video_core/shader/shader.h, line 148 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfiqORCxfRmxK0MzHU:-Kjg3ZnEwyirVt3GHtEc:b4lrts3) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.h#L148)):*

This doesn't seem like a good place to put these. Have you verified that both units have a separate counter/buffer for this? If not, then this seems like a premature change to make.
</blockquote></details>

Agreed. Again: Just moved existing code. The place to put them is not ideal.
There is a TODO in `WriteUniformFloatReg()`.
I can create an issue about this after merge?

---

*[src/video_core/shader/shader.h, line 192 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjfj9dl3_CQdjYrBKo_:-Kjg4_Q0XtALxUv8TlB5:bdbsyfq) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.h#L192)):*

IMO these should go into command_processor.cpp, or be split between that (incoming value parsing) and member vars in the relevant structs (applying the value, iff it's not trivial)
</blockquote></details>

This is more or less a temporary solution until we know wether there is some write-through from VS to GS and wether this information is stored per shader-unit or shader type.
(At which point we can move them into the appropriate objects)

I'm not sure how you'd want to split this.

---

*[src/video_core/shader/shader_jit_x64_compiler.h, line 26 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfhcJeKCKx996eW3ji:-Kjg42vvh19V7AaTZI-0:brkw54x) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader_jit_x64_compiler.h#L26)):*

Probably `4096 * 64` makes more sense? Also, maybe you should introduce a constant for the maximum shader size so it's just one place to change.
</blockquote></details>

Yeah, I just wasn't sure wether the 1024 was used to get the value in Kb or it was being used to represent the number of instructions.
It was more or less a quickfix as the current size will result in xbyak running out of memory.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695)*

yuriks:
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/video_core/shader/shader.h, line 148 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfiqORCxfRmxK0MzHU:-Kjida2QQhecJM3vdUVN:b-e7a6rx) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.h#L148)):*
<details><summary><i>Previously, JayFoxRox (Jannik Vogel) wrote…</i></summary><blockquote>

Agreed. Again: Just moved existing code. The place to put them is not ideal.
There is a TODO in `WriteUniformFloatReg()`.
I can create an issue about this after merge?
</blockquote></details>

You moved this code yourself as a separate change, it's not related to the moved code below? I think it doesn't make sense to make this part of the struct. Use separate static counters instead if necessary. These are more part of the command interface, not the shader engine state.

---

*[src/video_core/shader/shader.h, line 192 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjfj9dl3_CQdjYrBKo_:-KjihQvYn6uzWU3BvEiV:bua1pv7) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.h#L192)):*
<details><summary><i>Previously, JayFoxRox (Jannik Vogel) wrote…</i></summary><blockquote>

This is more or less a temporary solution until we know wether there is some write-through from VS to GS and wether this information is stored per shader-unit or shader type.
(At which point we can move them into the appropriate objects)

I'm not sure how you'd want to split this.
</blockquote></details>

As mentioned in IRC, I think they're better off in command_processor.cpp

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695)*

JayFoxRox:
Commit descriptions have been added, the first post has been converted into purely notes.

I've tried to make the requested changes.
I've decided to drop one of the commits - it will be added to the next PR when the introduced variables will actually be used.

---

Review status: 3 of 6 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/video_core/shader/shader.h, line 148 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfiqORCxfRmxK0MzHU:-KjiwGjInjcLsgRSMW7d:b-85oa8p) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.h#L148)):*

You moved this code yourself as a separate change, it's not related to the moved code below? I think it doesn't make sense to make this part of the struct. Use separate static counters instead if necessary. These are more part of the command interface, not the shader engine state.
</blockquote></details>

Moved back to where it was.
I've introduced a seperate static buffer.

---

*[src/video_core/regs_pipeline.h, line 207 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjfcPSq2WyUtey7Lb7s:-KjivMsaQbjMIJkZ99xB:b-4hzvdi) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/regs_pipeline.h#L207)):*

Also document what each of these values means. This whole reg needs some documentation.
</blockquote></details>

I'll remove this commit from the PR.
It's only necessary when implementing the actual GS logic (it was necessary until like 10 minutes before I created the PR, when SharedGS() etc were still being used in Write*() functions).

---

*[src/video_core/shader/shader.cpp, line 112 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kje9SuJ0eryT4JCV7Fb-r1-112:-Kjivlj-lGbp0Ik-i9YR:b-wssd43) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L112)):*
<details><summary><i>Previously, JayFoxRox (Jannik Vogel) wrote…</i></summary><blockquote>

I can remove `VSComMode` and `SharedGS()` temporarily and add them later as they are currently unused.
</blockquote></details>

Dropped this commit altogether.

---

*[src/video_core/shader/shader.cpp, line 134 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kje9SuJ0eryT4JCV7Fb-r1-134:-KjivrTIk1TWEAl3Ln-W:bbox8rh) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L134)):*

I say leave this function out until there comes a time when it becomes necessary, since right now it's not actually saving any code or work.
</blockquote></details>

Function dropped

---

*[src/video_core/shader/shader.cpp, line 211 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjfh4A9doctW9mdxomv:-KjiympWJa0L4eKqK3sw:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L211)):*

Same as the other trivial funcs pointed out by Subv
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695#-:-Kjivckak-1-f0H2pL1Z:b-6nac5)*

yuriks:
Reviewed 2 of 7 files at r2, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/video_core/shader/shader.h, line 147 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjkSY9YAWGaDHTMutvI:-KjkSY9YAWGaDHTMutvJ:b-6ijrhi) ([raw file](https://github.com/citra-emu/citra/blob/7cb79e3e9d04874ba5ec33b2c48173831c55d60b/src/video_core/shader/shader.h#L147)):*
> ```objc
>     }
> 
>     std::array<u32, 4096> program_code;
> ```

I think it's worth taking the opportunity to make a separate named constant for the maximum shader size, since you're already touching all places where it's repeated in this commit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695)*

JayFoxRox:
Review status: 0 of 6 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/video_core/shader/shader.h, line 147 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjkSY9YAWGaDHTMutvI:-Kjn6bty9GpWVTbLEEe4:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/7cb79e3e9d04874ba5ec33b2c48173831c55d60b/src/video_core/shader/shader.h#L147)):*

I think it's worth taking the opportunity to make a separate named constant for the maximum shader size, since you're already touching all places where it's repeated in this commit.
</blockquote></details>

Done.

---

*[src/video_core/shader/shader.cpp, line 193 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kje9SuJ0eryT4JCV7Fb-r1-193:-Kjn6fuLyZiA0csqHmc-:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/f0577160510c043f9f3ea2a4dad7b7adcdcc54e8/src/video_core/shader/shader.cpp#L193)):*

same as WriteUniformFloatSetupReg
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695)*

yuriks:
:lgtm:

---

Reviewed 4 of 7 files at r2, 6 of 6 files at r5, 1 of 1 files at r6, 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/video_core/shader/shader.h, line 28 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjp8tAGySYtyBTiP-7P:-Kjp8tAHJxoXNnz0AgR0:b-qc5h91) ([raw file](https://github.com/citra-emu/citra/blob/9500b5d9f6472c4953437fbd9ce47d135131b22b/src/video_core/shader/shader.h#L28)):*
> ```objc
> 
> constexpr unsigned MaxProgramCodeLength = 4096;
> constexpr unsigned MaxSwizzleDataLength = 4096;
> ```

These should be either `max_swizzle_data_length` or `MAX_SWIZZLE_DATA_LENGTH` (we're inconsistent on which of those two we use, but CamelCase is reserved for functions and classes.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695#-:-KjpA2fHfRvmaZWg5e8f:bnfp4nl)*

yuriks:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjpAzUbWiBvy14wEcjx:-KjpAzUbWiBvy14wEcjy:b-qr1kvt) (no related file):*
Both builds are failing, you should look into that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695)*

src/video_core/command_processor.cpp
```diff
@@ -48,6 +49,99 @@ static const u32 expand_bits_to_bytes[] = {
...
+        return "vertex shader";
+    }
+    if (&setup == &g_state.gs) {
+        return "geometry shader";
+    }
+    return "unknown shader";
+}
+
+static void WriteUniformBoolReg(Shader::ShaderSetup& setup, u32 value) {
+    static_assert(setup.uniforms.b.size() == 16);
```
wwylele: `static_assert` needs a string as an error message (the no-message version is only in c++17 iirc)

wwylele:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/video_core/command_processor.cpp, line 63 at r7](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjq1u9_frQiGZmHBcu-:-Kjq1u9_frQiGZmHBcu0:b-yx3o1r) ([raw file](https://github.com/citra-emu/citra/blob/5d82c9aa63276b0ddd3b785b6c13cbbe562affa0/src/video_core/command_processor.cpp#L63)):*
> ```cpp
> 
> static void WriteUniformBoolReg(Shader::ShaderSetup& setup, u32 value) {
>     static_assert(setup.uniforms.b.size() == 16);
> ```

`static_assert` needs a string as an error message (the no-message version is only in c++17 iirc)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695)*

JayFoxRox:
Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjpAzUbWiBvy14wEcjx:-KjrbCCcmqVGlpjbrz8Y:b-q3pqso) (no related file):*

Both builds are failing, you should look into that.
</blockquote></details>

Builds failed because of the `static_assert`

---

*[src/video_core/command_processor.cpp, line 63 at r7](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjq1u9_frQiGZmHBcu-:-Kjr_81tTRvpp5ELRIjS:b-2j2mt3) ([raw file](https://github.com/citra-emu/citra/blob/5d82c9aa63276b0ddd3b785b6c13cbbe562affa0/src/video_core/command_processor.cpp#L63)):*

`static_assert` needs a string as an error message (the no-message version is only in c++17 iirc)
</blockquote></details>

Dropped the assert altogether (it used to be `ASSERT()`, then `static_assert`, but the MacOS version didn't like that [with or without strings])

---

*[src/video_core/shader/shader.h, line 28 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2695#-Kjp8tAGySYtyBTiP-7P:-KjraRJ4m_FcbVUpmpAe:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/9500b5d9f6472c4953437fbd9ce47d135131b22b/src/video_core/shader/shader.h#L28)):*

These should be either `max_swizzle_data_length` or `MAX_SWIZZLE_DATA_LENGTH` (we're inconsistent on which of those two we use, but CamelCase is reserved for functions and classes.)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695)*

yuriks:
Reviewed 5 of 6 files at r8, 1 of 1 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695)*

wwylele:
Mostly LGTM. Please make sure all `LOG_TRACE` still works

---

---

*[src/video_core/command_processor.cpp, line 72 at r10](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjsQZtTqxhJUbFuHS03:-KjsQZtTqxhJUbFuHS04:b-5805au) ([raw file](https://github.com/citra-emu/citra/blob/8af6ba4a86611675b9020f2780d0fb68f6f903e4/src/video_core/command_processor.cpp#L72)):*
> ```cpp
>     setup.uniforms.i[index] = values;
>     LOG_TRACE(HW_GPU, "Set %s integer uniform %d to %02x %02x %02x %02x",
>               GetShaderSetupTypeName(setup), index, values.x.Value(), values.y.Value(),
> ```

Remove `Value()`. You should also check if debug build also compile.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695#-:-KjsQwavIr9-7YpocpUs:b-z5t7jx)*

Subv:
:lgtm:

---

Reviewed 6 of 6 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695#-:-Kjua2-WAZE9n5okrgsm:bnfp4nl)*

JayFoxRox:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/video_core/command_processor.cpp, line 72 at r10](https://reviewable.io:443/reviews/citra-emu/citra/2695#-KjsQZtTqxhJUbFuHS03:-Kjx2InklAIcAOx6jLos:bt43niw) ([raw file](https://github.com/citra-emu/citra/blob/8af6ba4a86611675b9020f2780d0fb68f6f903e4/src/video_core/command_processor.cpp#L72)):*

Remove `Value()`. You should also check if debug build also compile.
</blockquote></details>

Done.

Compiled fine with `-DCMAKE_BUILD_TYPE=Debug`.
MK logs ints "01 00 00 00" to trace at startup (as it does on master), so the log seems to work fine too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695)*

wwylele:
:lgtm: will merge after CI check

---

Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2695#-:-Kjx50cB1-st6mW4s70E:b-o50knb)*

Issue #1858, opened 2016-05-28
Combat of Giants: Dinosaurs 3D - Graphical issues (missing text)


JohnGodgames:
Tested also with GS https://github.com/JayFoxRox/citra/tree/gs-new-refactor and it's the same

[LOG](http://pasted.co/35382c33)

**3DS:**

![combat of giant dinosaur 3ds 1](https://cloud.githubusercontent.com/assets/11453507/15625425/baab63e2-24a6-11e6-9ba0-7408378d643f.JPG)

![combat of giant dinosaur 3ds 2](https://cloud.githubusercontent.com/assets/11453507/15625426/ce2ff234-24a6-11e6-8b4f-3bcfb6ffe628.JPG)

![combat of giant dinosaur 3ds 3](https://cloud.githubusercontent.com/assets/11453507/15625431/d41e7922-24a6-11e6-9012-5fc416dc3810.JPG)

![combat of giant dinosaur 3ds 4](https://cloud.githubusercontent.com/assets/11453507/15625433/d9e18836-24a6-11e6-8aca-c73d6205f548.JPG)

**Citra:**

![combat of giant dinosaur 1](https://cloud.githubusercontent.com/assets/11453507/15625436/e024639e-24a6-11e6-8d03-4ab3e37cb13c.JPG)

![combat of giant dinosaur 2](https://cloud.githubusercontent.com/assets/11453507/15625438/e4e7bd86-24a6-11e6-9fd0-776f3d4eecf9.JPG)

![combat of giant dinosaur 3](https://cloud.githubusercontent.com/assets/11453507/15625440/e96366f8-24a6-11e6-9368-76786580e78e.JPG)

![combat of giant dinosaur 4](https://cloud.githubusercontent.com/assets/11453507/15625441/edf8b182-24a6-11e6-8a35-9c3d33f5ccdc.JPG)

![combat of giant dinosaur 5](https://cloud.githubusercontent.com/assets/11453507/15625442/f2aeef66-24a6-11e6-965a-d0cabef2e2c8.JPG)

JayFoxRox:
This is caused by `Source::Texture3` ("source op 6") not being implemented.
We return 0.0 as default, but that means the game will multiply away all color.

I propose we stub the `Source::Texture3` with another color. I noticed this just by messing with the Texture3 value. I'm not sure if the game actually uses Texture3 or wether this is a bug (accidentally using Texture3 or using it as some kind of "always-white-texture").

[...]

Issue #2213, opened 2016-11-26
Mario & Luigi Paper Jam - Uniform block size did not match!


pippo2931:
Mario & Luigi Paper Jam crashes when loading a save game. OS X latest local build.

Driver is the vanilla driver of macOS Sierra (10.12.1) for the Intel Iris Pro.
Yes, size is zero and the shader is failing.

```
^[[0m^[[1;35m[  16.574664] Render.OpenGL <Critical> video_core/renderer_opengl/gl_rasterizer_cache.cpp:GetFramebufferSurfaces:556: Color and depth framebuffer memory regions overlap; overlapping framebuffers not supported!
^[[0m^[[1;35m[  16.574687] Render.OpenGL <Critical> video_core/renderer_opengl/gl_rasterizer_cache.cpp:GetFramebufferSurfaces:556: Color and depth framebuffer memory regions overlap; overlapping framebuffers not supported!
^[[0m^[[1;35m[  16.574707] Render.OpenGL <Critical> video_core/renderer_opengl/gl_rasterizer_cache.cpp:GetFramebufferSurfaces:556: Color and depth framebuffer memory regions overlap; overlapping framebuffers not supported!
^[[0m^[[1;35m[  16.755412] Render.OpenGL <Critical> video_core/renderer_opengl/gl_shader_gen.cpp:AppendSource:80: Unknown source op 6
^[[0m^[[1;31m[  16.760042] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadProgram:58: Error compiling fragment shader:
ERROR: 0:63: Index 3 beyond bounds (size 3)
ERROR: 0:63: Index 3 beyond bounds (size 3)
ERROR: 0:64: Use of undeclared identifier 'surface_normal'
ERROR: 0:69: Use of undeclared identifier 'normal'
ERROR: 0:70: Use of undeclared identifier 'normal'

^[[0m^[[1;31m[  16.760082] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.cpp:LoadProgram:81: Error linking shader:
ERROR: One or more attached shaders not successfully compiled

^[[0m^[[1;35m[  16.760095] Debug <Critical> video_core/renderer_opengl/gl_rasterizer.cpp:operator():1073: Assertion Failed!
Uniform block size did not match!
 ```

*Posts re-arranged and re-formated by JayFoxRox*

wwylele:
cc @JayFoxRox

[...]

Issue #1521, opened 2016-03-14
Ace Attorney Dual Destinies - Missing 2D graphics / 3d models


Punk34:
Good news, the game now is playable. Bad: a lot of 2d / 3d graphics is missing. The console says: **shader_jit: unhandled instructuon 0x2a (0xae800000)** and **Unknown source op 6**

![01](https://cloud.githubusercontent.com/assets/13983158/13755966/ba31907e-ea2e-11e5-88c7-c995fc9bb345.jpg)
![02](https://cloud.githubusercontent.com/assets/13983158/13755967/ba32932a-ea2e-11e5-9ca7-f3fe9ded249a.jpg)

KillzXGaming:
May have #1498

MoochMcGee:
Sounds like geo shaders m8.

anirudh8860:
How did it became playable? Can anyone tell me?

Subv:
@bunnei Weren't you looking for a game that used Texture3? Here you have it

wwylele:
The `Unknown source op 6`(Texture3-proctex) part should have been fixed, but the graphics is still broken for other reason.

wwylele:
Fixed by #3281 as reported so closing. Feel free to reopen if it is not the case.

PR #2697, opened 2017-05-11
Implemented Procedural Texture (Texture Unit 3)

wwylele:
Credits to @fincs and @MerryMage 

Implementation based on [fincs' notes](https://gist.github.com/fincs/543f7e1375815f495f10020a053f14e9).

**Hardware test**
[Here is a test written by fincs](https://gist.github.com/wwylele/cb0088d7ab628e9b711f22d212ba276b) ([binary](https://feos.mtheall.com/proctex-test.3dsx)). Note that this program need to be compiled using citro3d's "next" branch. You can change the parameters in the source code to test different modes and LUT. I will probably make a fully configurable testing program if requested.

**Problems**
 - There exists a mode involving 1D mipmaps for the final color LUT, however it hasn't been tested yet. Considering the low significance of it, I left it as TODO in citra.
 - PICA has a broken sqrt, which produces very inaccurate result for small numbers, and causes a little different output between 3ds and citra, for `SqrtAdd2` and `RMax` combiner mode (can be seen at the most inner ring of the "wood" example above). I am not going to "fix" this because the difference is not easy to notice and it is giant PITA to emulate "a bug".

**Fixed games:**
Mario & Luigi: Paper Jam (issue #2213)
screenshot from 2017-05-05 15-55-04

The sea is rendered by procedural texture, which previously causes shader compilation failure because the bump map uses proctex. 

Other test results are welcome. This PR is going to fix games that log `Unknown op source 6`, or have shader compilation failure due to using texture3 in bump map.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2697)


Subv:
Reviewed 8 of 8 files at r1.
Review status: 8 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/video_core/command_processor.cpp, line 473 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk6v2iRwY-wXtvihDc-:-Kk6v2iRwY-wXtvihDc0:b-51u7n4) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/command_processor.cpp#L473)):*
> ```cpp
>         switch (regs.texturing.proctex_lut_config.ref_table.Value()) {
>         case TexturingRegs::ProcTexLutTable::Noise:
>             pt.noise_table[index % pt.noise_table.size()].raw = value;
> ```

Is it verified that this index wraps around?

---

*[src/video_core/swrasterizer/proctex.cpp, line 106 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk6whcJXkDdPUAntk8U:-Kk6whcKfe_BEA1V_rXT:b6wmyza) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L106)):*
> ```cpp
>     }
>     case ProcTexClamp::Pulse:
>         if (coord < 0.5f)
> ```

Is 0.5 also treated as 1?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

Subv:
Reviewed 6 of 6 files at r2.

---

*[src/video_core/renderer_opengl/gl_rasterizer.cpp, line 1200 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk6yOwouTaxv6GgHoHD:-Kk6yOwouTaxv6GgHoHE:bgqvokf) ([raw file](https://github.com/citra-emu/citra/blob/4ab2c72ded31122e5550f95ae997a28684a3017b/src/video_core/renderer_opengl/gl_rasterizer.cpp#L1200)):*
> ```cpp
>             glGetUniformLocation(shader->shader.handle, "proctex_noise_lut");
>         if (uniform_proctex_noise_lut != -1) {
>             glUniform1i(uniform_proctex_noise_lut, 10);
> ```

Mabye we should start using an enum for these texture ids, that'd be a different PR, though

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

yuriks:
Partial review, I haven't looked at the OpenGL bits yet.

---

Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 22 unresolved discussions.

---

*[src/video_core/command_processor.cpp, line 471 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk80pwJ4MBfziiypMV5:-Kk80pwJ4MBfziiypMV6:b-1ztab) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/command_processor.cpp#L471)):*
> ```cpp
>         auto& index = regs.texturing.proctex_lut_config.index;
>         auto& pt = g_state.proctex;
>         switch (regs.texturing.proctex_lut_config.ref_table.Value()) {
> ```

Add a blank line before this to make it more visible.

---

*[src/video_core/pica_state.h, line 31 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk80Ew2Ui2VOCyxauU0:-Kk80Ew2Ui2VOCyxauU1:bqud0ru) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/pica_state.h#L31)):*
> ```objc
>     struct ProcTex {
>         union Lut128Entry {
>             // Used for raw access
> ```

This comment seem unnecessary. A `u32 raw;` entry in unions is a common idiom in the codebase, and the comment doesn't really explain anything over the variable name.

---

*[src/video_core/pica_state.h, line 37 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7wcqVCQDb4_Y3Yovi:-Kk7wcqVCQDb4_Y3Yovj:b-c11xab) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/pica_state.h#L37)):*
> ```objc
>             BitField<0, 12, u32> value; // 0.0.12 fixed point
> 
>             // Difference between two entry value. Used for efficient interpolation.
> ```

"entry *values*."

---

*[src/video_core/pica_state.h, line 40 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7xSrg39MsaEB7lkjE:-Kk7xSrg39MsaEB7lkjF:b-5nwly0) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/pica_state.h#L40)):*
> difference
Opinion: I think "delta" is a better/less awkward word than "difference". But "difference" is simpler vocabulary so you might want to keep using it regardless.

---

*[src/video_core/pica_state.h, line 51 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7wrR7PFFZPZj4TlhH:-Kk7wrR7PFFZPZj4TlhI:bw3onc9) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/pica_state.h#L51)):*
> ```objc
>         };
> 
>         using Lut128 = std::array<Lut128Entry, 128>;
> ```

The name "Lut128" is kinda arbitrary... I think I'd get rid of this typedef and use arrays directly (like for the color LUTs) and rename the union to "ValueEntry", to mirror "ColorEntry".

---

*[src/video_core/regs_texturing.h, line 205 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7m4TKYOVI0QQpBk1h:-Kk7m4TLFjlXT1jG9PRn:b-m31y23) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/regs_texturing.h#L205)):*
> float1.5.10
Is this the only format used for 16-bit floats? We can probably just say it's a "float16" and then have a class to manipulate those similar to our float24 one.

---

*[src/video_core/swrasterizer/proctex.cpp, line 19 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7pZqQa9S9CkEOPk9_:-Kk7pZqQa9S9CkEOPk9a:b-rm0jac) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L19)):*
> index
I think (someone correct me if I'm wrong) "index" is always used to refer to integers. A value from 0.0 to 1.0 would be a coordinate

---

*[src/video_core/swrasterizer/proctex.cpp, line 31 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk83IM6h3_Opk884TXW:-Kk83IM6h3_Opk884TXX:bcmrnp9) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L31)):*
> UnknownPRNG
Is it really unknown now, if it has been fully reversed already?

---

*[src/video_core/swrasterizer/proctex.cpp, line 31 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk9JRI1MKme_2FEJ5Pb:-Kk9JRI1MKme_2FEJ5Pc:b-p59fno) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L31)):*
> ```cpp
> // These function are used to generate random noise for procedural texture. Their results are
> // verified against real hardware.
> static int UnknownPRNG(int v) {
> ```

Change the parameter, return and table type to `unsigned int`. (This actually makes the divisions somewhat faster since they can be simplified to less bitwise operations.)

---

*[src/video_core/swrasterizer/proctex.cpp, line 32 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk83WxzlnvUWV0NKbap:-Kk83WxzlnvUWV0NKbaq:bwpl7ur) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L32)):*
> ```cpp
> // verified against real hardware.
> static int UnknownPRNG(int v) {
>     const std::array<int, 16> table{{0, 4, 10, 8, 4, 9, 7, 12, 5, 15, 13, 14, 11, 15, 2, 11}};
> ```

Make this static and constexpr (if you can)

---

*[src/video_core/swrasterizer/proctex.cpp, line 33 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk9JsjAhfCWlUcDcXtR:-Kk9JsjAhfCWlUcDcXtS:bl1tc6u) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L33)):*
> ```cpp
> static int UnknownPRNG(int v) {
>     const std::array<int, 16> table{{0, 4, 10, 8, 4, 9, 7, 12, 5, 15, 13, 14, 11, 15, 2, 11}};
>     return ((v % 9 + 2) * 3 & 0xF) ^ table[(v / 9) & 0xF];
> ```

You could replace the first half with a table of length 9 (`{6, 9, 12, 15, 2, 5, 8, 11, 14}`), unless to preserve the somewhat arbitrary looking formula.

---

*[src/video_core/swrasterizer/proctex.cpp, line 36 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk9JdePTU8YkMMUPice:-Kk9JdePTU8YkMMUPicf:b-tdpil4) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L36)):*
> ```cpp
> }
> 
> static float NoiseRand(int x, int y) {
> ```

Also change the types to unsigned

---

*[src/video_core/swrasterizer/proctex.cpp, line 37 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk83Y6fHwifz_0Ykqz8:-Kk83Y6fHwifz_0Ykqz9:bwpl7ur) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L37)):*
> ```cpp
> 
> static float NoiseRand(int x, int y) {
>     const std::array<int, 16> table{{10, 2, 15, 8, 0, 7, 4, 5, 5, 13, 2, 6, 13, 9, 3, 14}};
> ```

Make this static and constexpr (if you can)

---

*[src/video_core/swrasterizer/proctex.cpp, line 40 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk84ZvoVI6UXfWq6Wm8:-Kk84Zvp-kLTGpb5I_ON:b-5m99tz) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L40)):*
> ```cpp
>     int u2 = UnknownPRNG(x);
>     int v2 = UnknownPRNG(y);
>     v2 += ((u2 & 3) == 1) ? 4 : 0;
> ```

You could express this without the ternary as just `((u2 & 3) == 1) * 4`. Up to you.

---

*[src/video_core/swrasterizer/proctex.cpp, line 69 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk8B9wVuIrUfKjT0Z96:-Kk8B9wWkepHuH69IUuo:b5kyrt6) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L69)):*
> ```cpp
>     const float x0 = g0 * (1 - x_noise) + g1 * x_noise;
>     const float x1 = g2 * (1 - x_noise) + g3 * x_noise;
>     return x0 * (1 - y_noise) + x1 * y_noise;
> ```

Last 3 lines are bilinear interpolation. Factor this out into a function. There's also the Math::Lerp function you can use, which results in:

```c++
static float BilinearInterp(float s00, float s01, float s10, float s11, float s, float t) {
    float s0 = Lerp(s00, s01, s);
    float s1 = Lerp(s10, s11, s);
    return Lerp(s0, s1, t);
}
```

---

*[src/video_core/swrasterizer/proctex.cpp, line 177 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkA1icCnCD0I-Wy0c6o:-KkA1icDd6GUHPx83RxI:b-7ik9uw) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L177)):*
> ```cpp
>     // Shift
>     const float u_shift = ShiftCoord(u, v, regs.proctex.u_shift, regs.proctex.u_clamp);
>     const float v_shift = ShiftCoord(v, u, regs.proctex.v_shift, regs.proctex.v_clamp);
> ```

This seems to be incorrect somehow. [This modification of the test](https://gist.github.com/yuriks/9d43d5f610a8af6f3fb579143b234db2/revisions) gives slightly different results between Citra and hardware.

---

*[src/video_core/swrasterizer/proctex.cpp, line 193 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkA2vcq1acEDOzN9XYz:-KkA2vcq1acEDOzN9XZ-:b-1d7uas) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L193)):*
> ```cpp
>     const u32 offset = regs.proctex_lut_offset;
>     const u32 width = regs.proctex_lut.width;
>     const float index = offset + (lut_index * (width - 1));
> ```

I feel like "index" is especially confusing here. `color_coord`?

---

*[src/video_core/swrasterizer/proctex.h, line 12 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkA3SBPbn91T1jEPuOn:-KkA3SBPbn91T1jEPuOo:b2ssk0h) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.h#L12)):*
> ```objc
> 
> /// Generates procedural texture color for the given coordinates
> Math::Vec4<u8> ProcTex(float u, float v);
> ```

Pass in `g_state.regs.texturing` (or a more specific sub-struct if one exists) and `g_state.proctex` as parameters.

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 339 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkA3rqxCb-TrG7PaNDM:-KkA3rqyP2yoyfams2XT:b955c3t) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/rasterizer.cpp#L339)):*
> ```cpp
> 
>             // sample procedural texture
>             if (regs.texturing.main_config.texture3_enable) {
> ```

Did you check what's in the buffer if you try to use `Source::Texture3` while this is disabled?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697#-:-KkA43PmT-vgt8UZW2iA:b-ijsdj7)*

yuriks:
Took a look at the OpenGL commit too.

---

Reviewed 6 of 6 files at r2.
Review status: all files reviewed at latest revision, 23 unresolved discussions.

---

*[src/video_core/renderer_opengl/gl_rasterizer.cpp, line 1200 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk6yOwouTaxv6GgHoHD:-KkF2fKY7T6riIpEPoOz:bxtqjed) ([raw file](https://github.com/citra-emu/citra/blob/4ab2c72ded31122e5550f95ae997a28684a3017b/src/video_core/renderer_opengl/gl_rasterizer.cpp#L1200)):*

Mabye we should start using an enum for these texture ids, that'd be a different PR, though
</blockquote></details>

+1, though the LUT textures might be replaced by something else (uniforms, etc.) soon™ (as soon as someone gets around to it).

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 833 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkF4M0G-hfeFFNHkN2G:-KkF4M0G-hfeFFNHkN2H:b-gl9tkc) ([raw file](https://github.com/citra-emu/citra/blob/4ab2c72ded31122e5550f95ae997a28684a3017b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L833)):*
> ```cpp
>                                      // extracted as index_i = 127.0 and index_f = 1.0
>     uint entry = texelFetch(lut, int(index_i), 0).r;
>     float entry_difference = float(int(entry << 8U) >> 20U); // extract the signed difference
> ```

You should do this on the CPU and upload the LUT with 2 components instead. That way you're also free to use another data type for this instead of integer and skip some of the conversion in the return below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697#-:-KkF6GMGG3Id7NWOCI9L:bvp1zqo)*

wwylele:
Review status: all files reviewed at latest revision, 23 unresolved discussions.

---

*[src/video_core/command_processor.cpp, line 471 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk80pwJ4MBfziiypMV5:-KkRabFyAg49ZOsJbctE:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/command_processor.cpp#L471)):*

Add a blank line before this to make it more visible.
</blockquote></details>

Done.

---

*[src/video_core/command_processor.cpp, line 473 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk6v2iRwY-wXtvihDc-:-KkR_ZrxlK3boTaKt4zY:b8dhkev) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/command_processor.cpp#L473)):*

Is it verified that this index wraps around?
</blockquote></details>

Not verified. This was following  the fog lut register implementation. I will change it to an assert instead (like the lighting lut one)

---

*[src/video_core/pica_state.h, line 31 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk80Ew2Ui2VOCyxauU0:-KkRaiyqZEzC5RJ4B4K8:b-w6ujxo) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/pica_state.h#L31)):*

This comment seem unnecessary. A `u32 raw;` entry in unions is a common idiom in the codebase, and the comment doesn't really explain anything over the variable name.
</blockquote></details>

Removed. Was a copy pasta from lighting lut

---

*[src/video_core/pica_state.h, line 37 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7wcqVCQDb4_Y3Yovi:-KkRawX53df-hShaULvZ:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/pica_state.h#L37)):*

"entry *values*."
</blockquote></details>

Done.

---

*[src/video_core/pica_state.h, line 40 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7xSrg39MsaEB7lkjE:-KkRay_qh8dCLhJT3qIR:b-kuh4qd) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/pica_state.h#L40)):*

> difference
Opinion: I think "delta" is a better/less awkward word than "difference". But "difference" is simpler vocabulary so you might want to keep using it regardless.
</blockquote></details>

"Delta" was actually the first word coming to my mind, but "difference" is already used for fog and lighting lut, so I don't want to break the convention.

---

*[src/video_core/pica_state.h, line 51 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7wrR7PFFZPZj4TlhH:-KkRdW17YqQY9iwxgyUW:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/pica_state.h#L51)):*

The name "Lut128" is kinda arbitrary... I think I'd get rid of this typedef and use arrays directly (like for the color LUTs) and rename the union to "ValueEntry", to mirror "ColorEntry".
</blockquote></details>

Done.

---

*[src/video_core/regs_texturing.h, line 205 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7m4TKYOVI0QQpBk1h:-KkRe0gEspvPZdViQY25:bidpo24) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/regs_texturing.h#L205)):*

> float1.5.10
Is this the only format used for 16-bit floats? We can probably just say it's a "float16" and then have a class to manipulate those similar to our float24 one.
</blockquote></details>

Should be the only format and we already had the class for it. Fixed the comment

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 833 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkF4M0G-hfeFFNHkN2G:-KkReGfwtbE7M39ESFUq:bwys2rm) ([raw file](https://github.com/citra-emu/citra/blob/4ab2c72ded31122e5550f95ae997a28684a3017b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L833)):*

You should do this on the CPU and upload the LUT with 2 components instead. That way you're also free to use another data type for this instead of integer and skip some of the conversion in the return below.
</blockquote></details>

This was following the fog lut style. I agree to decode on CPU, but since we are going to do #2704, should I keep the same style so that we won't get confused when we change them all together? (Or this PR should wait for a final solution to the issue and I change it directly?)

---

*[src/video_core/swrasterizer/proctex.cpp, line 19 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk7pZqQa9S9CkEOPk9_:-KkRicQI_XuLbhelwrhJ:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L19)):*

> index
I think (someone correct me if I'm wrong) "index" is always used to refer to integers. A value from 0.0 to 1.0 would be a coordinate
</blockquote></details>

Done.

---

*[src/video_core/swrasterizer/proctex.cpp, line 31 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk83IM6h3_Opk884TXW:-KkRimCp_N2IHC2cpxVj:brs6a5l) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L31)):*

> UnknownPRNG
Is it really unknown now, if it has been fully reversed already?
</blockquote></details>

The result is proved fully correct, but the algorithm is not necessary the one used in 3ds, so I marked it as "unknown if this is the 3ds way". Do you have any suggestions on naming this?

---

*[src/video_core/swrasterizer/proctex.cpp, line 31 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk9JRI1MKme_2FEJ5Pb:-KkRjPFXoSOt_EFHeF06:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L31)):*

Change the parameter, return and table type to `unsigned int`. (This actually makes the divisions somewhat faster since they can be simplified to less bitwise operations.)
</blockquote></details>

Done.

---

*[src/video_core/swrasterizer/proctex.cpp, line 32 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk83WxzlnvUWV0NKbap:-KkRjklUid_nIJhZNZ41:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L32)):*

Make this static and constexpr (if you can)
</blockquote></details>

Done.

---

*[src/video_core/swrasterizer/proctex.cpp, line 33 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk9JsjAhfCWlUcDcXtR:-KkRjxZr7FGOwKBI0zly:b-fzr4pe) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L33)):*

You could replace the first half with a table of length 9 (`{6, 9, 12, 15, 2, 5, 8, 11, 14}`), unless to preserve the somewhat arbitrary looking formula.
</blockquote></details>

This is not arbitrary looking to me... it is a simple linear formula which looks good to me

---

*[src/video_core/swrasterizer/proctex.cpp, line 36 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk9JdePTU8YkMMUPice:-KkRkctBZLxQC7cW--_v:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L36)):*

Also change the types to unsigned
</blockquote></details>

Done.

---

*[src/video_core/swrasterizer/proctex.cpp, line 37 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk83Y6fHwifz_0Ykqz8:-KkRkd_3E8qkJjYuzmhh:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L37)):*

Make this static and constexpr (if you can)
</blockquote></details>

Done.

---

*[src/video_core/swrasterizer/proctex.cpp, line 40 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk84ZvoVI6UXfWq6Wm8:-KkRkg93FCzHSw8_hQsm:bv93bj5) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L40)):*

You could express this without the ternary as just `((u2 & 3) == 1) * 4`. Up to you.
</blockquote></details>

I find the `true is 1` trick hard to read, so I prefer the current one.

---

*[src/video_core/swrasterizer/proctex.cpp, line 69 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk8B9wVuIrUfKjT0Z96:-KkRvOCleSBbi-5XPjcG:b3jimzq) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L69)):*

Last 3 lines are bilinear interpolation. Factor this out into a function. There's also the Math::Lerp function you can use, which results in:

```c++
static float BilinearInterp(float s00, float s01, float s10, float s11, float s, float t) {
    float s0 = Lerp(s00, s01, s);
    float s1 = Lerp(s10, s11, s);
    return Lerp(s0, s1, t);
}
```
</blockquote></details>

Done. Add the function to vector_math.h, where the `Lerp` lives

---

*[src/video_core/swrasterizer/proctex.cpp, line 193 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkA2vcq1acEDOzN9XYz:-KkRnFpSBTRw3gp6G7VK:b-obqfnl) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L193)):*

I feel like "index" is especially confusing here. `color_coord`?
</blockquote></details>

The `lut_index` should be actually `coord`. The `index` is an actual index that has its integer part representing an index into an array.

---

*[src/video_core/swrasterizer/proctex.h, line 12 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkA3SBPbn91T1jEPuOn:-KkRrbLuXlVDNyctegr2:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.h#L12)):*

Pass in `g_state.regs.texturing` (or a more specific sub-struct if one exists) and `g_state.proctex` as parameters.
</blockquote></details>

Done.

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 339 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkA3rqxCb-TrG7PaNDM:-KkRnhuJ9Fh-2IYt2Eqj:b8hce5g) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/rasterizer.cpp#L339)):*

Did you check what's in the buffer if you try to use `Source::Texture3` while this is disabled?
</blockquote></details>

Checked, the color is pretty random. I am not going to emulate that weird color

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

wwylele:
Addressed most trivial issues

---

Review status: 6 of 15 files reviewed at latest revision, 23 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697#-:-KkRvuOkLT77gnnnDeyI:b-3wspk4)*

wwylele:
Review status: 6 of 15 files reviewed at latest revision, 23 unresolved discussions.

---

*[src/video_core/swrasterizer/proctex.cpp, line 106 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk6whcJXkDdPUAntk8U:-KkS4ga6w5DsBEDN_anK:b-3ivk1z) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L106)):*

Is 0.5 also treated as 1?
</blockquote></details>

Just did hardware test. 0.5 is treated as 0, so I fixed here. The gl renderer already has 0.5->0. Sorry for inconsistency.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

wwylele:
Review status: 6 of 15 files reviewed at latest revision, 23 unresolved discussions.

---

*[src/video_core/command_processor.cpp, line 473 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk6v2iRwY-wXtvihDc-:-KkS6ybPHt3W7yxTbxqe:b-l4u1ad) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/command_processor.cpp#L473)):*

Not verified. This was following  the fog lut register implementation. I will change it to an assert instead (like the lighting lut one)
</blockquote></details>

Update: wrap around _verified_.  Changed back to what it was.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

yuriks:
Reviewed 2 of 2 files at r3, 7 of 7 files at r4.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 833 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkF4M0G-hfeFFNHkN2G:-KkTspo4cNGv7wHP1Kh1:b-eqj29a) ([raw file](https://github.com/citra-emu/citra/blob/4ab2c72ded31122e5550f95ae997a28684a3017b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L833)):*

This was following the fog lut style. I agree to decode on CPU, but since we are going to do #2704, should I keep the same style so that we won't get confused when we change them all together? (Or this PR should wait for a final solution to the issue and I change it directly?)
</blockquote></details>

IMO do it now. It's not that big of a change anyway.

---

*[src/video_core/swrasterizer/proctex.cpp, line 31 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk83IM6h3_Opk884TXW:-KkTtcnAK-ayf-vbKOEP:b-we3yfj) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L31)):*

The result is proved fully correct, but the algorithm is not necessary the one used in 3ds, so I marked it as "unknown if this is the 3ds way". Do you have any suggestions on naming this?
</blockquote></details>

If that's the intent with the name then I'm spell that out in the comment above (saying that the results are equivalent, but it's not known if the algorithm is the same). For the function name, maybe name this one `NoiseRand1D`, and the other one `NoiseRand2D`?

---

*[src/video_core/swrasterizer/proctex.cpp, line 33 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk9JsjAhfCWlUcDcXtR:-KkTtSs8YtP7AujBl7U7:b2nhll9) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L33)):*

This is not arbitrary looking to me... it is a simple linear formula which looks good to me
</blockquote></details>

Ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

wwylele:
Review status: 11 of 15 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 833 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkF4M0G-hfeFFNHkN2G:-KkUI_tzUplB2zpG2F9O:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/4ab2c72ded31122e5550f95ae997a28684a3017b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L833)):*

IMO do it now. It's not that big of a change anyway.
</blockquote></details>

Done.

---

*[src/video_core/swrasterizer/proctex.cpp, line 31 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-Kk83IM6h3_Opk884TXW:-KkUIbrKCG1DG66YLh9c:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L31)):*

If that's the intent with the name then I'm spell that out in the comment above (saying that the results are equivalent, but it's not known if the algorithm is the same). For the function name, maybe name this one `NoiseRand1D`, and the other one `NoiseRand2D`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

yuriks:
Reviewed 1 of 1 files at r5, 3 of 3 files at r6.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 832 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkUJbxhpdGW1nh9EMCY:-KkUJbxhpdGW1nh9EMCZ:b-vce25t) ([raw file](https://github.com/citra-emu/citra/blob/e60be355b52539b7392f1a6e496a2a610eb28a9c/src/video_core/renderer_opengl/gl_shader_gen.cpp#L832)):*
> ```cpp
>     float index_f = coord - index_i; // fract() cannot be used here because 128.0 needs to be
>                                      // extracted as index_i = 127.0 and index_f = 1.0
>     vec4 entry = texelFetch(lut, int(index_i), 0);
> ```

You can use a ".rg" swizzle similar to before, I think it makes it clearer you're only using 2 values.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

wwylele:
Review status: 14 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 832 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkUJbxhpdGW1nh9EMCY:-KkUtwFtok5kcoxoqZWv:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/e60be355b52539b7392f1a6e496a2a610eb28a9c/src/video_core/renderer_opengl/gl_shader_gen.cpp#L832)):*

You can use a ".rg" swizzle similar to before, I think it makes it clearer you're only using 2 values.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

yuriks:
Reviewed 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

wwylele:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/video_core/swrasterizer/proctex.cpp, line 177 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkA1icCnCD0I-Wy0c6o:-KkV4JryEZP11wfL0CA5:bfnalkt) ([raw file](https://github.com/citra-emu/citra/blob/5ad20dd1c90126c7a5c564e4b5d044df17f021f5/src/video_core/swrasterizer/proctex.cpp#L177)):*

This seems to be incorrect somehow. [This modification of the test](https://gist.github.com/yuriks/9d43d5f610a8af6f3fb579143b234db2/revisions) gives slightly different results between Citra and hardware.
</blockquote></details>

Fixed. The auxiliary coordinate should be without noise applied.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

yuriks:
Reviewed 1 of 7 files at r8, 5 of 6 files at r9.
Review status: 14 of 15 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/video_core/swrasterizer/proctex.cpp, line 81 at r9](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkZowU7npNRANe5dlVs:-KkZowU7npNRANe5dlVt:bsidbbp) ([raw file](https://github.com/citra-emu/citra/blob/9bd00cb300875776d936c5c9e60871b86eb1fa13/src/video_core/swrasterizer/proctex.cpp#L81)):*
> ```cpp
>     default:
>         LOG_CRITICAL(HW_GPU, "Unknown shift mode %u", static_cast<u32>(mode));
>         return u;
> ```

Instead of always summing u here and returning that, you can instead return the *amount* of shift to apply (or 0). Then below you can calculate the u_shift/v_shift before the noise, and just apply it afterwards. Makes the code look somewhat nicer imo.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

yuriks:
Yup, output seems to match now!

---

Reviewed 1 of 6 files at r9.

---

*[src/video_core/swrasterizer/proctex.cpp, line 81 at r9](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkZowU7npNRANe5dlVs:-KkZpbXe1zT6otBRPJ-A:bmleagz) ([raw file](https://github.com/citra-emu/citra/blob/9bd00cb300875776d936c5c9e60871b86eb1fa13/src/video_core/swrasterizer/proctex.cpp#L81)):*

Instead of always summing u here and returning that, you can instead return the *amount* of shift to apply (or 0). Then below you can calculate the u_shift/v_shift before the noise, and just apply it afterwards. Makes the code look somewhat nicer imo.
</blockquote></details>

But well, the code also looks good enough as-is, so up to you really.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697#-:-KkZphamMsSBCWZdMhLL:b-mb3ibc)*

wwylele:
Review status: 8 of 15 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/video_core/swrasterizer/proctex.cpp, line 81 at r9](https://reviewable.io:443/reviews/citra-emu/citra/2697#-KkZowU7npNRANe5dlVs:-Kk_TzUxmEwgcUfbO_Es:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/9bd00cb300875776d936c5c9e60871b86eb1fa13/src/video_core/swrasterizer/proctex.cpp#L81)):*

But well, the code also looks good enough as-is, so up to you really.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

yuriks:
Reviewed 1 of 7 files at r10, 6 of 6 files at r11.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

Subv:
Reviewed 4 of 7 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2697)*

yuriks:
Is this waiting on anything? Can we merge it?

wwylele:
I am just waiting for merge

yuriks:
Guess you're not waiting anymore then.

turbedi:
Does this fix #1858 and #1521 ?

Issue #2698, opened 2017-05-12
LOG_TRACE statements should be verified during CI builds


yuriks:
Right now it happens semi-often that a LOG_TRACE statement has its parameters broken somehow and this gets checked in because they're not built in Release builds. Since developers rarely use Debug builds due to how slow they are, we should check this in CI somehow. There are several approaches to this:

1. Make Travis CI Linux (which is generally the fastest to build) also do Debug builds, using Travis' matrix feature. I'm not sure how easy this is to implement but I don't imagine it's very hard. It might potentially increase the flakiness of the builds or make them take more time, decreasing agility.

2. Remove LOG_TRACE from very hot loops, and then enable it as a regular conditional statement. This decreases the amount of magic in the build and makes these problems be detected during regular development. However, the LOG_TRACE statements are occasionally useful for debugging issues or collecting data for simulations/benchmarks.

3. Add a dedicated defined symbols to enable LOG_TRACE, and enable it for CI builds. This could be a good option since afaik we don't currently distribute the presubmit builds anyway, so their runtime performance doesn't matter.

Does the new logging framework (#2646) offer anything to address this that we might be able to simply wait for, @jroweboy?

Personally, alternative 3 seems like the best for me.

jroweboy:
RE option 1: Both travis and appveyor support matrix builds, and both of them only build one at a time for open source applications. So if we use that feature, then it will increase build time.

RE option 3: CI environments (both travis and appveyor) have a different flag for push to master and pull request. We can use this enable `LOG_TRACE` for builds that are checking pull requests while still keeping them out of release builds.

I'm fine for both option 1 and 3. I like option 1 because it seems weird to have an option to build debug builds, without actually testing the build. But I also like option 3 because it won't increase build times while still testing LOG_TRACE statements. If I had to choose, I'd probably go with option 1 and just deal with the increased build times.

If we come to a consensus about which option to choose, I can add it to the logging rewrite task

vvanelslande:
Fixed by #4602

Issue #2646, opened 2017-03-30
Rework logging in citra


jroweboy:
# Rationale

* Citra has a large number of warnings caused by improper casts when printf formatting.
* Logging only writes to stderr, which makes copying logs on windows more painful than it should be (because the windows console is still inadequate) and requires a work around in macOS to even display the logs. (For the mac workaround, see https://github.com/citra-emu/citra/blob/master/.travis-upload.sh#L104)
* Logging to disk isn't a hard change to make, but it is still work to roll your own sinks, meaning any future sinks you want to add will take even more work to fit it into the existing framework (not a big issue, but still one as it could add problems for cross platform sinks).

With these issues in mind, I wrote a proof of concept rework of logging to use the excellent spdlog library. spdlog uses fmtlib for log formatting which, in addition to being almost as fast as printf formatting, has a similar interface to python formatting. This allows developers to use {} as a flexible, typesafe format specifier, with a host of other options for fine tuned control of the output. See http://fmtlib.net/latest/syntax.html for more on what can be done with fmtlib. spdlog also has many other benefits such as several easy-to-configure logging sinks, support for async logging, and safe multithread logging. The proof of concept can be found here. Its not ready for PR yet, but it gives a little insight into what changes will be in phase 1 https://github.com/citra-emu/citra/compare/master...jroweboy:great-logging-rewrite-of-2017

# Design

Number one goal is to keep the logging framework as close as possible to the function of the current logging system. This way the contributors don't need to learn anything new to add more logs; they will still be able to use the familiar LOG_LEVEL("") macros. Additionally the log filters should continue to work as they currently do. Luckily enough, spdlog supports all of the same logging levels that citra uses, and filtering is also supported, so migrating is not too complicated

After talking with @wwylele about the changes, he suggested to make this issue for some feedback, and also to split the rework into 3 separate phases to help it get merged faster.

1. Add a spdlog backend that co-exists with the current logging system, and a new macro SPD_LEVEL that uses the new logger.
2. After phase 1 is merged, start porting each of the existing logging statements to use the new SPD_LEVEL macros and fmtlib formatting. This can be done on a module by module basis so its easier to review and faster to merge than a full application find/replace
[...]

PR #2707, opened 2017-05-17
Logging backend rewrite to use spdlog

jroweboy:
## Updates since RFC:

* Class based logging is back in. 
* Filter support for the new backend has been added.
* Loggers are stored in an array for fast lookup.
* Macros renamed from `SPDLOG_*` to `SLOG_*` in order to prevent a conflict since `spdlog` already defines `SPDLOG_TRACE` and `SPDLOG_DEBUG` macros... (it doesn't matter really since these names are temporary anyway)
* Several style and bad code fixes.

Thanks for the excellent feedback so far. Please review once more so we can get this merged and get to work on part 2 :)

## Original Description
This is a implementation of the logging rewrite to use spdlog. This was written in conjunction with the feedback in #2646 and tries to address the comments in that proposal.

## TODO List:
- [x] Functional spdlog backend that writes to console and disk.
- [x] Helper macros to enable easy conversion from the old macros.
- [x] Ability to set and change filters. I want to make sure the interface is okay before I start writing glue for filters.
- [ ] Better handling for LOG_TRACE (see #2698)

## Plan of Action

1. Comments, Review Fixes, and Merge this pull request
2. Migrate existing logging statements from the old macros to the new ones. (Can be done in segments when people get bored and are looking for easy contributions)
2.5 Merge any PRs using the old macros in the interim
3. One last sweep for any missing logging that need to be converted. Change them all and remove the old logging system
3.5 Any PRs adding old logging code will need to be rebased.

## Why RFC?

I would like to get a few comments about the interface before I pull this out of RFC in order to make sure that we are comfortable with the new logging code. After I get some feedback, I'll rebase to make the commit messages better. The logging section of this PR includes some sample code, and I've taken a little time to change the few logging statements 

#### (OBSOLETE) Register a new logger

At the top of each file you intend to use logging in, you will need to use the `REGISTER_LOGGER(name)` macro which will setup a new `spdlog::logger` with that name. This name is similar in function to how the old ClassName works, but since its just a string, its much more flexible. The macro expands to instantiate a new logger and stores a value that will be used to look up the logger in the other logging macros. 

Since the macro registers a global logger, if you need to use the same logger in different files, you can use the `spdlog::get` method to return a handle to the logger you want. (In order to do so, you will need to include the spdlog header in your file, as log.h does not include that)

#### Logging

The new logging macros are almost exactly the same as the old macros, but no longer require the class as that's automatically inferred from the global variable that is created with `REGISTER_LOGGER`. Examples:

```cpp
// All of the old macros are still around, and can be used during the migration process. The new macros are prefixed with S and function the same way.
// The following are valid with the new macros:
SLOG_DEBUG(Frontend, "Testing argument order:");
SLOG_INFO(Renderer, "{1} {0}", "world!", "Hello"); // Prints Hello world!
SLOG_INFO(Frontend, "Hex digit: {0:#x}", 16); // Prints Hex digit: 0x10

SLOG_WARNING(Frontend, "Hi", 't', 15.67); // Compiles and runs, but only prints Hi since its the other arguments are considered part of the formatting arguments (and there are no {} to format into)
SLOG_ERROR(); // doesn't compile as at least 1 argument is required. Fixing this didn't seem important enough
```

The specifiers for formatting can be found on the website for fmtlib http://fmtlib.net/latest/syntax.html

#### Filtering

Filtering is currently implemented, and it works in much the same manner as it does currently.

#### Other details

As an implementation detail, `log.h` now includes `fmt/format.h`, which means if you want to use `fmtlib` directly, its probably already included.

Also this change depends on https://github.com/MerryMage/dynarmic/pull/106 as I'm adding fmt as a static library which is incompatible with fmt-header-only used in dynarmic.

#### Goal #1 - Don't increase the compile time 

spdlog is a header only library, and according to the numbers here https://github.com/gabime/spdlog/issues/120#issuecomment-282995233 spdlog greatly increases the compile time. In order to hide spdlog from the rest of the code base, I took care to make sure there were no references to it outside of the backend logging code. In order to handle variadic arguments to the logging call, I generated a `fmt::ArgList` of all of the arguments, but as pointed out in the review, this is an internal method. 

Additionally, the file level static logger (see goal #2) is stored as an index into the vector of loggers, so as to, once again, not include spdlog everywhere. (This could potentially be solved by forward declaring spdlog::logger, or just not storing it on the class and looking up the logger using spdlog's global logger registry, or something i haven't considered yet.)

#### Goal #2 - Remove Logging Classes

This idea came from a conversation with yuriks, where we discussed what could a rewrite of logging provide (and perhaps I misinterpreted how to code the idea up). See the last bullet point here: https://github.com/citra-emu/citra/issues/2646#issuecomment-294379564 and I mentioned this on IRC and had the following conversation: https://pastebin.com/vgw4vbXk I think it would be a good idea to have more flexible logging classes, in order to have more flexible filtering. You currently _can_ add a new logging class for more narrow filtering, but people don't. One eventual goal I have (and was brought up in #2646) is to write a "console" logger for citra-qt (sdl can continue to use stderr) which would allow runtime filtering of logs. Having more flexible class names would allow more specific log filtering.

I agree that my current implementation could use a lot of work, and I will gladly put in the effort to make it work.

(I apologize for not including this rationale from the start.)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2707)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,6 +28,3 @@
 [submodule "spdlog"]
     path = externals/spdlog
     url = https://github.com/gabime/spdlog.git
-[submodule "fmt"]
-    path = externals/fmt
-    url = https://github.com/fmtlib/fmt
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,3 +25,6 @@
 [submodule "fmt"]
     path = externals/fmt
     url = https://github.com/fmtlib/fmt.git
+[submodule "spdlog"]
+    path = externals/spdlog
+    url = https://github.com/gabime/spdlog.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "spdlog"]
     path = externals/spdlog
     url = https://github.com/gabime/spdlog.git
+[submodule "fmt"]
+    path = externals/fmt
+    url = https://github.com/fmtlib/fmt
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,7 +1,5 @@
 # Enable modules to include each other's files
 include_directories(.)
-# Include fmtlib so it can be used across the application for logging
-include_directories(../externals/fmt)
 
 add_subdirectory(common)
 add_subdirectory(core)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,7 @@
 # Enable modules to include each other's files
 include_directories(.)
+# Include fmtlib so it can be used across the application for logging
+include_directories(../externals/fmt)
 
 add_subdirectory(common)
 add_subdirectory(core)
--- a/src/common/logging/backend_spdlog.cpp
+++ b/src/common/logging/backend_spdlog.cpp
@@ -1,14 +1,42 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <array>
+#include <memory>
+#include <vector>
 #include <spdlog/spdlog.h>
 
 #include "common/assert.h"
 #include "common/logging/backend.h"
 #include "common/logging/backend_spdlog.h"
+#include "common/logging/filter.h"
 #include "common/logging/formatter.h"
 #include "common/string_util.h"
 
 namespace Log {
 
-static spdlog::level::level_enum GetLevel(Log::Level log_level) {
+class SpdLogBackend {
+public:
+    SpdLogBackend();
+    ~SpdLogBackend();
+
+    static std::shared_ptr<SpdLogBackend> Instance();
+
+    SpdLogBackend(SpdLogBackend const&) = delete;
+    const SpdLogBackend& operator=(SpdLogBackend const&) = delete;
+
+    using LogArray =
+        std::array<std::shared_ptr<spdlog::logger>, static_cast<u8>(Log::Class::Count)>;
+    const LogArray& GetLoggers() {
+        return loggers;
+    }
+
+private:
+    LogArray loggers;
+};
+
+static spdlog::level::level_enum GetSpdLogLevel(Log::Level log_level) {
     switch (log_level) {
     case Log::Level::Trace:
         return spdlog::level::trace;
@@ -27,8 +55,8 @@ static spdlog::level::level_enum GetLevel(Log::Level log_level) {
     }
 }
 
-SpdLogBackend& SpdLogBackend::Instance() {
-    static SpdLogBackend instance;
+std::shared_ptr<SpdLogBackend> SpdLogBackend::Instance() {
+    static auto instance = std::make_shared<SpdLogBackend>();
     return instance;
 }
 
@@ -36,23 +64,22 @@ SpdLogBackend::SpdLogBackend() {
     // setup the custom citra formatter
     spdlog::set_formatter(std::make_shared<Formatter>());
 
-    std::vector<spdlog::sink_ptr> sinks;
     // Define the sinks to be passed to the loggers
     // true means truncate file
     auto file_sink = std::make_shared<spdlog::sinks::simple_file_sink_mt>("citra_log.txt", true);
 #ifdef _WIN32
     auto color_sink = std::make_shared<spdlog::sinks::wincolor_stderr_sink_mt>();
 #else
-    auto stderr_sink = spdlog::sinks::stderr_sink_mt::instance();
-    auto color_sink = std::make_shared<spdlog::sinks::ansicolor_sink>(stderr_sink);
+    auto color_sink = std::make_shared<spdlog::sinks::ansicolor_stdout_sink_mt>();
 #endif
+    std::vector<spdlog::sink_ptr> sinks;
     sinks.push_back(std::move(file_sink));
     sinks.push_back(std::move(color_sink));
 
     // register all of loggers with spdlog
     for (u8 log_class = 0; log_class != static_cast<u8>(Log::Class::Count); ++log_class) {
-        spdlog::create(GetLogClassName(static_cast<Log::Class>(log_class)), begin(sinks),
-                       end(sinks));
+        loggers[log_class] = spdlog::create(GetLogClassName(static_cast<Log::Class>(log_class)),
+                                            begin(sinks), end(sinks));
     }
 }
 
@@ -62,11 +89,18 @@ SpdLogBackend::~SpdLogBackend() {
 
 void SpdLogImpl(Class log_class, Level log_level, const char* file, int line_num,
                 const char* function, const char* format, fmt::ArgList args) {
-    SpdLogBackend::Instance();
+    auto logger = SpdLogBackend::Instance()->GetLoggers()[static_cast<u8>(log_class)];
     fmt::MemoryWriter formatting_buffer;
     formatting_buffer << Common::TrimSourcePath(file) << ':' << function << ':' << line_num << ": "
                       << format;
-    auto& logger = spdlog::get(GetLogClassName(log_class));
-    logger->log(GetLevel(log_level), formatting_buffer.c_str(), args);
+    logger->log(GetSpdLogLevel(log_level), formatting_buffer.c_str(), args);
+}
+
+void SpdLogSetFilter(Filter* filter) {
+    auto loggers = SpdLogBackend::Instance()->GetLoggers();
+    auto class_level = filter->GetClassLevel();
+    for (u8 log_class = 0; log_class != static_cast<u8>(Log::Class::Count); ++log_class) {
+        loggers[log_class]->set_level(GetSpdLogLevel(class_level[log_class]));
+    }
 }
 }; // namespace Log
--- a/src/common/logging/backend_spdlog.cpp
+++ b/src/common/logging/backend_spdlog.cpp
@@ -1,9 +1,12 @@
 #include <spdlog/spdlog.h>
 
 #include "common/assert.h"
+#include "common/logging/backend.h"
 #include "common/logging/backend_spdlog.h"
 #include "common/logging/formatter.h"
 #include "common/string_util.h"
+
+namespace Log {
 
 static spdlog::level::level_enum GetLevel(Log::Level log_level) {
     switch (log_level) {
@@ -21,14 +24,10 @@ static spdlog::level::level_enum GetLevel(Log::Level log_level) {
         return spdlog::level::critical;
     default:
         UNREACHABLE();
-        break;
     }
-    return spdlog::level::off;
 }
 
-namespace Log {
-
-SpdLogBackend& SpdLogBackend::instance() {
+SpdLogBackend& SpdLogBackend::Instance() {
     static SpdLogBackend instance;
     return instance;
 }
@@ -37,6 +36,7 @@ SpdLogBackend::SpdLogBackend() {
     // setup the custom citra formatter
     spdlog::set_formatter(std::make_shared<Formatter>());
 
+    std::vector<spdlog::sink_ptr> sinks;
     // Define the sinks to be passed to the loggers
     // true means truncate file
     auto file_sink = std::make_shared<spdlog::sinks::simple_file_sink_mt>("citra_log.txt", true);
@@ -46,29 +46,27 @@ SpdLogBackend::SpdLogBackend() {
     auto stderr_sink = spdlog::sinks::stderr_sink_mt::instance();
     auto color_sink = std::make_shared<spdlog::sinks::ansicolor_sink>(stderr_sink);
 #endif
-    sinks.push_back(file_sink);
-    sinks.push_back(color_sink);
+    sinks.push_back(std::move(file_sink));
+    sinks.push_back(std::move(color_sink));
+
+    // register all of loggers with spdlog
+    for (u8 log_class = 0; log_class != static_cast<u8>(Log::Class::Count); ++log_class) {
+        spdlog::create(GetLogClassName(static_cast<Log::Class>(log_class)), begin(sinks),
+                       end(sinks));
+    }
 }
 
 SpdLogBackend::~SpdLogBackend() {
     spdlog::drop_all();
 }
 
-const std::shared_ptr<spdlog::logger>& SpdLogBackend::GetLogger(u32 logger) const {
-    return loggers[logger];
-}
-
-u32 SpdLogBackend::RegisterLogger(const char* class_name) {
-    loggers.push_back(spdlog::create(class_name, sinks.begin(), sinks.end()));
-    return loggers.size() - 1;
-}
-
-void SpdLogImpl(u32 logger, Level log_level, const char* format, fmt::ArgList& args) {
-    auto log = SpdLogBackend::instance().GetLogger(logger);
-    log->log(GetLevel(log_level), format, args);
-}
-
-u32 RegisterLogger(const char* class_name) {
-    return SpdLogBackend::instance().RegisterLogger(class_name);
+void SpdLogImpl(Class log_class, Level log_level, const char* file, int line_num,
+                const char* function, const char* format, fmt::ArgList args) {
+    SpdLogBackend::Instance();
+    fmt::MemoryWriter formatting_buffer;
+    formatting_buffer << Common::TrimSourcePath(file) << ':' << function << ':' << line_num << ": "
+                      << format;
+    auto& logger = spdlog::get(GetLogClassName(log_class));
+    logger->log(GetLevel(log_level), formatting_buffer.c_str(), args);
 }
 }; // namespace Log
--- a/src/common/logging/backend_spdlog.cpp
+++ b/src/common/logging/backend_spdlog.cpp
@@ -8,6 +8,7 @@
 #include <spdlog/spdlog.h>
 
 #include "common/assert.h"
+#include "common/file_util.h"
 #include "common/logging/backend.h"
 #include "common/logging/backend_spdlog.h"
 #include "common/logging/filter.h"
@@ -66,7 +67,8 @@ SpdLogBackend::SpdLogBackend() {
 
     // Define the sinks to be passed to the loggers
     // true means truncate file
-    auto file_sink = std::make_shared<spdlog::sinks::simple_file_sink_mt>("citra_log.txt", true);
+    auto file_sink = std::make_shared<spdlog::sinks::simple_file_sink_mt>(
+        FileUtil::GetUserPath(D_USER_IDX) + "citra_log.txt", true);
 #ifdef _WIN32
     auto color_sink = std::make_shared<spdlog::sinks::wincolor_stderr_sink_mt>();
 #else
--- a/src/common/logging/backend_spdlog.cpp
+++ b/src/common/logging/backend_spdlog.cpp
@@ -1,7 +1,30 @@
+#include <spdlog/spdlog.h>
+
 #include "common/assert.h"
 #include "common/logging/backend_spdlog.h"
 #include "common/logging/formatter.h"
 #include "common/string_util.h"
+
+static spdlog::level::level_enum GetLevel(Log::Level log_level) {
+    switch (log_level) {
+    case Log::Level::Trace:
+        return spdlog::level::trace;
+    case Log::Level::Debug:
+        return spdlog::level::debug;
+    case Log::Level::Info:
+        return spdlog::level::info;
+    case Log::Level::Warning:
+        return spdlog::level::warn;
+    case Log::Level::Error:
+        return spdlog::level::err;
+    case Log::Level::Critical:
+        return spdlog::level::critical;
+    default:
+        UNREACHABLE();
+        break;
+    }
+    return spdlog::level::off;
+}
 
 namespace Log {
 
@@ -31,54 +54,18 @@ SpdLogBackend::~SpdLogBackend() {
     spdlog::drop_all();
 }
 
-const std::shared_ptr<spdlog::logger> SpdLogBackend::GetLogger(u32 logger) const {
+const std::shared_ptr<spdlog::logger>& SpdLogBackend::GetLogger(u32 logger) const {
     return loggers[logger];
 }
 
 u32 SpdLogBackend::RegisterLogger(const char* class_name) {
-    loggers.push_back(std::make_shared<spdlog::logger>(class_name, sinks.begin(), sinks.end()));
+    loggers.push_back(spdlog::create(class_name, sinks.begin(), sinks.end()));
     return loggers.size() - 1;
 }
 
-spdlog::level_t GetLevel(Level log_level) {
-    switch (log_level) {
-    case Level::Trace:
-        return spdlog::level::trace;
-    case Level::Debug:
-        return spdlog::level::debug;
-    case Level::Info:
-        return spdlog::level::info;
-    case Level::Warning:
-        return spdlog::level::warn;
-    case Level::Error:
-        return spdlog::level::err;
-    case Level::Critical:
-        return spdlog::level::critical;
-    default:
-        UNREACHABLE();
-        break;
-    }
-    return spdlog::level::off;
-}
-
-template <typename Arg1, typename... Args>
-void SpdLogMessage(u32 logger, Level log_level, const char* filename, unsigned int line_nr,
-                   const char* function, const char* format, const Arg1& arg, const Args&... args) {
+void SpdLogImpl(u32 logger, Level log_level, const char* format, fmt::ArgList& args) {
     auto log = SpdLogBackend::instance().GetLogger(logger);
-    fmt::MemoryWriter formatting_buffer;
-    formatting_buffer << Common::TrimSourcePath(filename) << ':' << function << ':' << line_nr
-                      << ": " << format;
-    log->log(GetLevel(log_level), formatting_buffer.c_str(), arg, args...);
-}
-
-template <typename T>
-void SpdLogMessage(u32 logger, Level log_level, const char* filename, unsigned int line_nr,
-                   const char* function, const T& msg) {
-    auto log = SpdLogBackend::instance().GetLogger(logger);
-    fmt::MemoryWriter formatting_buffer;
-    formatting_buffer << Common::TrimSourcePath(filename) << ':' << function << ':' << line_nr
-                      << ": " << msg;
-    log->log(GetLevel(log_level), formatting_buffer.c_str());
+    log->log(GetLevel(log_level), format, args);
 }
 
 u32 RegisterLogger(const char* class_name) {
--- a/src/common/logging/backend_spdlog.cpp
+++ b/src/common/logging/backend_spdlog.cpp
@@ -79,7 +79,8 @@ SpdLogBackend::SpdLogBackend() {
     sinks.push_back(std::move(color_sink));
 
     // register all of loggers with spdlog
-    for (u8 log_class = 0; log_class != static_cast<u8>(Log::Class::Count); ++log_class) {
+    for (ClassType log_class = 0; log_class != static_cast<ClassType>(Log::Class::Count);
+         ++log_class) {
         loggers[log_class] = spdlog::create(GetLogClassName(static_cast<Log::Class>(log_class)),
                                             begin(sinks), end(sinks));
     }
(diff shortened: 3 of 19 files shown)


src/common/logging/log.h
```diff
@@ -106,6 +107,22 @@ void LogMessage(Class log_class, Level log_level, const char* filename, unsigned
 #endif
     ;
 
+void SpdLogImpl(u32 logger, Level log_level, const char* format, fmt::ArgList& args);
+
+template <typename Arg1, typename... Args>
+void SpdLogMessage(u32 logger, Level log_level, const char* filename, unsigned int line_nr,
+                   const char* function, const Arg1& format, const Args&... args) {
+    typedef fmt::internal::ArgArray<sizeof...(Args)> ArgArray;
```
lioncash:
```cpp
using ArgArray = fmt::internal::ArgArray<sizeof...(Args)>;
```

also are you certain it's even OK to use stuff that's present in the internal namespace of fmt?

src/common/logging/formatter.cpp
```diff
@@ -0,0 +1,46 @@
+#include <chrono>
+#include <string>
+#include "common/assert.h"
+#include "common/logging/formatter.h"
+
+static const char* GetLevelName(spdlog::level_t log_level) {
```
lioncash: Why is this outside the Log namespace?

src/common/logging/log.h
```diff
@@ -128,3 +145,31 @@ void LogMessage(Class log_class, Level log_level, const char* filename, unsigned
     LOG_GENERIC(::Log::Class::log_class, ::Log::Level::Error, __VA_ARGS__)
 #define LOG_CRITICAL(log_class, ...)                                                               \
     LOG_GENERIC(::Log::Class::log_class, ::Log::Level::Critical, __VA_ARGS__)
+
+// Define the spdlog level macros
+
+#define REGISTER_LOGGER(class_name) static u32 _logger = ::Log::RegisterLogger(class_name)
```
lioncash: `static const`, also names with prefixed underscores (at all) are reserved (in this case by an implementation for use in the global namespace) according to the C++ standard. This also makes it infeasible to use more than one file-scoped logger.

src/common/logging/backend_spdlog.h
```diff
@@ -0,0 +1,27 @@
+#include <memory>
+#include <vector>
+#include <spdlog/spdlog.h>
+#include "common/logging/log.h"
+
+namespace Log {
+
+class SpdLogBackend {
+public:
+    static SpdLogBackend& instance();
```
wwylele: This could be just a static variable instead of a function.

src/common/logging/formatter.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+    unsigned int time_seconds = static_cast<unsigned int>(timestamp.count() / 1000000);
+    unsigned int time_fractional = static_cast<unsigned int>(timestamp.count() % 1000000);
+
+    msg.formatted << '[' << fmt::pad(time_seconds, 4, ' ') << '.'
+                  << fmt::pad(time_fractional, 6, '0') << "] ";
+    msg.formatted << *msg.logger_name << " <" << GetLevelName(msg.level) << "> ";
+
+    msg.formatted << fmt::StringRef(msg.raw.data(), msg.raw.size());
+    msg.formatted.write(spdlog::details::os::eol, spdlog::details::os::eol_size);
```
wwylele: like the `fmt::internal` namespace, are you sure you can use stuff in `spdlog::details` namespace?

bunnei:
I didn't review this yet, but just wanted to say I don't like the "SPDLOG_*" naming convention (including the name of the dependency in the new logging API). Perhaps we'll want another logging backend in the future, but not need/want to change all the callers...?

jroweboy:
@bunnei See https://github.com/citra-emu/citra/issues/2646 The SPDLOG name is only temporary. The plan is to name them something different during the migration so that existing logging code will continue to function and other PRs can be merged without rebasing against this. Then after all of the migration is complete, remove the old logging code entirely, and just do a s/SPDLOG/LOG replace everywhere. If you don't like this (and have a better migration plan) lemme know.

@lioncash Thanks for the great feedback. I've edited the original post to put a justification for some of the design decisions (because I stupidly forgot to include them) The feedback I've gotten seems to be that class based logging should stay and I have no issue with doing that. I'll have a little more time tomorrow to update the PR to use class based logging again.

bunnei:
@jroweboy ah got it - Nope, that sounds good to me!

yuriks:
I responded to some of the feedback to the file-level logger idea at #2646

jroweboy:
Okay, according to the feedback I've read, I think class based logging is here to stay, so I've updated the PR to use the original logging classes. I've also addressed the other issues that were raised about the code as well.

I'd like to look into adding a function to create a new logging class on the fly if there is interest. I'm worried that it'll make it more confusing in the future if there are two ways to define loggers so I'm not bothering with it right now.

src/common/logging/backend_spdlog.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto color_sink = std::make_shared<spdlog::sinks::wincolor_stderr_sink_mt>();
+#else
+    auto stderr_sink = spdlog::sinks::stderr_sink_mt::instance();
+    auto color_sink = std::make_shared<spdlog::sinks::ansicolor_sink>(stderr_sink);
+#endif
+    sinks.push_back(std::move(file_sink));
+    sinks.push_back(std::move(color_sink));
+
+    // register all of loggers with spdlog
+    for (u8 log_class = 0; log_class != static_cast<u8>(Log::Class::Count); ++log_class) {
```
jroweboy: I couldn't find a better way to iterate over all of the log classes without defining an array of log classes (which would mean a 3rd location to add your logger to when you add a new class) Any thoughts?

src/common/logging/backend_spdlog.cpp
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+SpdLogBackend::~SpdLogBackend() {
+    spdlog::drop_all();
+}
+
+void SpdLogImpl(Class log_class, Level log_level, const char* file, int line_num,
+                const char* function, const char* format, fmt::ArgList args) {
+    SpdLogBackend::Instance();
```
jroweboy: I don't like this, but I wasn't sure where else I should put the global? I figured here was the best spot to guarantee that spdlog is initialized on the first logging logging call

src/common/logging/backend_spdlog.cpp
```diff
@@ -0,0 +1,72 @@
...
+    fmt::MemoryWriter formatting_buffer;
+    formatting_buffer << Common::TrimSourcePath(file) << ':' << function << ':' << line_num << ": "
+                      << format;
+    auto& logger = spdlog::get(GetLogClassName(log_class));
```
wwylele: Is string-based query fast enough?
Also, this shouldn't be a reference.

.gitmodules
```diff
@@ -25,3 +25,9 @@
 [submodule "fmt"]
     path = externals/fmt
     url = https://github.com/fmtlib/fmt.git
+[submodule "spdlog"]
+    path = externals/spdlog
+    url = https://github.com/gabime/spdlog.git
+[submodule "fmt"]
```
wwylele: remove this block, or we will have two fmt blocks...

src/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 # Enable modules to include each other's files
 include_directories(.)
+# Include fmtlib so it can be used across the application for logging
+include_directories(../externals/fmt)
```
wwylele: See #2734, I am assuming @yuriks already prepared the including directories in recent PRs?

src/common/logging/backend_spdlog.h
```diff
@@ -0,0 +1,19 @@
+#include <memory>
+#include <vector>
+#include <spdlog/spdlog.h>
+#include "common/logging/log.h"
+
+namespace Log {
+
+class SpdLogBackend {
```
wwylele: This class doesn't need expose in a header

src/common/logging/backend_spdlog.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include "common/assert.h"
+#include "common/logging/backend.h"
+#include "common/logging/backend_spdlog.h"
+#include "common/logging/formatter.h"
+#include "common/string_util.h"
+
+namespace Log {
+
+static spdlog::level::level_enum GetLevel(Log::Level log_level) {
```
wwylele: I would prefer a more specific name like `GetSpdLogLevel`

src/common/logging/backend_spdlog.cpp
wwylele: I don't find anything wrong with this?

src/common/logging/backend_spdlog.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+SpdLogBackend::SpdLogBackend() {
+    // setup the custom citra formatter
+    spdlog::set_formatter(std::make_shared<Formatter>());
+
+    std::vector<spdlog::sink_ptr> sinks;
+    // Define the sinks to be passed to the loggers
+    // true means truncate file
+    auto file_sink = std::make_shared<spdlog::sinks::simple_file_sink_mt>("citra_log.txt", true);
```
wwylele: Does this mean all log will also output to the file? (and where is this file located, exactly? the current working directory?)

And why is `stderr_sink` not included in `sinks`? is that automatically registered?

(Sorry for not reading spdlog documentation, just want a quick answer)

src/common/logging/log.h
```diff
@@ -128,3 +134,28 @@ void LogMessage(Class log_class, Level log_level, const char* filename, unsigned
     LOG_GENERIC(::Log::Class::log_class, ::Log::Level::Error, __VA_ARGS__)
 #define LOG_CRITICAL(log_class, ...)                                                               \
     LOG_GENERIC(::Log::Class::log_class, ::Log::Level::Critical, __VA_ARGS__)
+
+// Define the spdlog level macros
+#ifdef _DEBUG
+#define SPDLOG_TRACE(log_class, fmt, ...)                                                          \
+    ::Log::SpdLogMessage(log_class, ::Log::Level::Trace, __FILE__, __LINE__, __func__, fmt,        \
```
wwylele: `log_class`->`::Log::Class::log_class`?

src/common/logging/backend_spdlog.cpp
wwylele: I find this really weird to read and not sure what you actually want to do...

To reduce some weirdness (but probably still not good), you can just define the `static SpdLogBackend backend` here. Since the backend object won't be used anywhere else, it hasn't to follow a full singleton style.

Or just define a plain `Initialize` function and let it called by the frontend or called on the first log.

jroweboy: The main reason I did it this way is to avoid any potential future problems where someone adds a logging statement in a static initializer, which would be playing russian roulette to find out if the logging has loaded first. https://isocpp.org/wiki/faq/ctors#static-init-order

I can't decide if its still a good idea to keep it named `Instance` or should I switch to `Initialize` ... i'm bikeshedding my own PR right now :/

src/common/logging/backend_spdlog.cpp
jroweboy: > Does this mean all log will also output to the file? 

Yes. Its in the current working directory. I don't see a reason to make this configurable or turn-off-able either, but if someone has a good reason why they should. It truncates at the start of program so its not going to balloon in size or anything.

> And why is stderr_sink not included in sinks? is that automatically registered?

the `stderr_sink` is passed into the constructor to the color sink where it will colorize the output. when reading the docs though i just found a shortcut for this idiom, so i'll update it.

src/common/logging/backend_spdlog.cpp
jroweboy: It hasn't really affected speed. My idea to store the loggers in a vector last time didn't go over as well, but that was probably because I removed the logging classes. Now that its back to using classes again, I can store them in a array

src/CMakeLists.txt
jroweboy: yes. but we'll have to link fmt against everywhere now as well

src/common/logging/backend_spdlog.cpp
Subv: why is log_class an u8? if we end up adding more log classes this will come back to haunt us in the future

src/common/logging/backend_spdlog.h
```diff
@@ -0,0 +1,11 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+namespace Log {
+class Filter;
+
+void SpdLogSetFilter(Filter* filter);
+} // namespace Log
```
Subv: missing newline at end of file.

yuriks:
Bunch of smaller stylistic comment, but my one main concern is that we're doing expensive string formatting even for disabled logs atm. I really think that one needs to be addressed.

---

Reviewed 1 of 8 files at r4, 13 of 13 files at r11, 2 of 2 files at r12, 2 of 11 files at r14, 12 of 15 files at r16, 1 of 1 files at r19, 2 of 2 files at r20.
Review status: all files reviewed at latest revision, 46 unresolved discussions.

---

*[externals/CMakeLists.txt, line 54 at r11](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNrxd3KrAvnPQO8CEA:-KnNrxd3KrAvnPQO8CEB:b-vt13ki) ([raw file](https://github.com/citra-emu/citra/blob/50efaebbae2f211df29e1f35e9ad99b2d73445ac/externals/CMakeLists.txt#L54)):*
> ```CMake
> # Spdlog
> add_library(spdlog INTERFACE)
> target_include_directories(spdlog INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/spdlog/include)
> ```

Leave the entries in this file sorted in alphabetical order

---

*[src/citra_qt/game_list.cpp, line 320 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNvROAwZl-Zg6lJs_q:-KnNvROBXS3F-ykTwrHS:b4pc2gw) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/citra_qt/game_list.cpp#L320)):*
> ```cpp
>         !FileUtil::IsDirectory(dir_path.toStdString())) {
>         SLOG_ERROR(Frontend, "Could not find game list folder at {}",
>                    dir_path.toLocal8Bit().data());
> ```

Does spdlog have a defined encoding for the logs (e.g. UTF-8) or is `toLocal8Bit()` still the correct thing to use? Also, we should probably define a formatting overload for QString to avoid manually doing this conversion.

---

*[src/citra_qt/main.cpp, line 325 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNw06j1iA7mjEeLg1c:-KnNw06kRvfuoHsUZem0:b-1kn9qx) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/citra_qt/main.cpp#L325)):*
> ```cpp
>         switch (result) {
>         case Core::System::ResultStatus::ErrorGetLoader:
>             SLOG_CRITICAL(Frontend, "Failed to obtain loader for {}!", filename.toStdString());
> ```

Another QString usage (just ear-marking these so they can be updated.)

---

*[src/common/CMakeLists.txt, line 100 at r11](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNs7OVgQfq_Yip_dYT:-KnNs7OVgQfq_Yip_dYU:b1xzxvj) ([raw file](https://github.com/citra-emu/citra/blob/50efaebbae2f211df29e1f35e9ad99b2d73445ac/src/common/CMakeLists.txt#L100)):*
> ```CMake
> 
> target_link_libraries(common PUBLIC Boost::boost microprofile spdlog)
> ```

Did you intentionally introduce these newlines?

---

*[src/common/string_util.h, line 148 at r12](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNssCg7z5E4qu2hhOj:-KnNssChy2t1KtgVkDGM:b-aunbh3) ([raw file](https://github.com/citra-emu/citra/blob/73658a7629e618f779112949e653ca04a0627441/src/common/string_util.h#L148)):*
> ```objc
>  * @return A pointer to the same string passed as `path`, but starting at the trimmed portion
>  */
> const char* TrimSourcePath(const char* path, const char* root = "src");
> ```

You're copying, not moving, this function in this commit. Either remove from the source and fix references or just update the commit description.

---

*[src/common/logging/backend.cpp, line 91 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNwoaEyNzzNwTYkq7r:-KnNwoaEyNzzNwTYkq7s:b-gzd8zf) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend.cpp#L91)):*
> ```cpp
>     case Class::Count:
>     default:
>         UNREACHABLE();
> ```

Instead of adding the `default:` case, add another UNREACHABLE outside after the switch. This should lets us take advantage of compiler warnings in either case.

---

*[src/common/logging/backend_spdlog.cpp, line 31 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNy-kIk139mBnW5oS3:-KnNy-kJxF1okt9ipmeB:b-yeajsx) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend_spdlog.cpp#L31)):*
> ```cpp
> 
>     using LogArray =
>         std::array<std::shared_ptr<spdlog::logger>, static_cast<u8>(Log::Class::Count)>;
> ```

size_t, not u8

---

*[src/common/logging/backend_spdlog.cpp, line 55 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNxGuZEavdZ6z71IZr:-KnNxGuZEavdZ6z71IZs:b-wzilat) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend_spdlog.cpp#L55)):*
> ```cpp
>         return spdlog::level::critical;
>     default:
>         UNREACHABLE();
> ```

Same principle here, move the UNREACHABLE outside the switch

---

*[src/common/logging/backend_spdlog.cpp, line 59 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNxZoJJJM5Jk2CLkyA:-KnNxZoJJJM5Jk2CLkyB:bwoumyh) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend_spdlog.cpp#L59)):*
> Instance
This is a static instance, there's no reason to return a shared_ptr to it.

---

*[src/common/logging/backend_spdlog.cpp, line 60 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNxiISlFlfjxodP18g:-KnNxiISlFlfjxodP18h:b-xkcp2q) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend_spdlog.cpp#L60)):*
> ```cpp
> 
> std::shared_ptr<SpdLogBackend> SpdLogBackend::Instance() {
>     static auto instance = std::make_shared<SpdLogBackend>();
> ```

Just make the variable type SpdLogBackend directly, you don't need to heap-allocate it.

---

*[src/common/logging/backend_spdlog.cpp, line 77 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNyRdMM6zc7VPK1wWS:-KnNyRdMM6zc7VPK1wWT:b-oak9yb) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend_spdlog.cpp#L77)):*
> ```cpp
>     auto color_sink = std::make_shared<spdlog::sinks::ansicolor_stderr_sink_mt>();
> #endif
>     std::vector<spdlog::sink_ptr> sinks;
> ```

This can be an initializer list or array instead of a vector (and even if you keep the vector you can initialize it with an initializer list instead of the pushes). It's also shorter to just define the sinks inline instead of going through variables I think.

---

*[src/common/logging/backend_spdlog.cpp, line 82 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNyomzlzzQube6mmYY:-KnNyomzlzzQube6mmYZ:b7oaodb) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend_spdlog.cpp#L82)):*
> ```cpp
> 
>     // register all of loggers with spdlog
>     for (ClassType log_class = 0; log_class != static_cast<ClassType>(Log::Class::Count);
> ```

Since you're casting everywhere, might as well make the loop counter an int

---

*[src/common/logging/backend_spdlog.cpp, line 95 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNzNJpkmmi9l_SiHhf:-KnNzNJpkmmi9l_SiHhg:b-hvh4pt) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend_spdlog.cpp#L95)):*
> ```cpp
> void SpdLogImpl(Class log_class, Level log_level, const char* file, int line_num,
>                 const char* function, const char* format, fmt::ArgList args) {
>     auto logger = SpdLogBackend::Instance()->GetLoggers()[static_cast<u8>(log_class)];
> ```

int/size_t, not u8

---

*[src/common/logging/backend_spdlog.cpp, line 99 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnNznKn2vvW3N1SvR6N:-KnNznKn2vvW3N1SvR6O:bknxdgr) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend_spdlog.cpp#L99)):*
> ```cpp
>     formatting_buffer << Common::TrimSourcePath(file) << ':' << function << ':' << line_num << ": "
>                       << format;
>     logger->log(GetSpdLogLevel(log_level), formatting_buffer.c_str(), args);
> ```

Honestly, I think this is pretty much unacceptable from a performance standpoint. You're going to incur an allocation, trimming the path, concatenating/formatting into the buffer, etc. for *every* log, even disabled ones. We need to find a way around this so it's only done for active logs, or move to some other logging backend that natively supports these features, or just stick with our own code. (Worst case, if you want to keep using spdlog's async sinks we can just pre-filter and pre-format the messages with our own mechanism, then just pass the raw string to spdlog. We're not doing any kind of structure logging anyway so this doesn't hurt us.)

---

*[src/common/logging/backend_spdlog.cpp, line 102 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO1AWiEAmaOt3dzE1y:-KnO1AWiEAmaOt3dzE1z:b-h4c84x) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/backend_spdlog.cpp#L102)):*
> ```cpp
> }
> 
> void SpdLogSetFilter(Filter* filter) {
> ```

filter isn't modified, so it should probably be a const&

---

*[src/common/logging/filter.cpp, line 26 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO4sA0iokd9Qis0co0:-KnO4sA0iokd9Qis0co1:bp8orhy) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/filter.cpp#L26)):*
> ```cpp
> const std::array<Level, static_cast<size_t>(Class::Count)>& Filter::GetClassLevel() {
>     return class_levels;
> }
> ```

Just inline this in the class.

---

*[src/common/logging/filter.h, line 29 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO38vzxx1G-zs5iu19:-KnO38vzxx1G-zs5iu1A:by02dqb) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/filter.h#L29)):*
> ```objc
>     void SetClassLevel(Class log_class, Level level);
>     /// Returns the list of levels that each logger is filtered to
>     const std::array<Level, static_cast<size_t>(Class::Count)>& GetClassLevel();
> ```

This sort of break the whole abstraction this class is based of and exposes exactly how it's implementing filters, but I guess that fine for what we're using it for and not a huge burden...

---

*[src/common/logging/formatter.cpp, line 28 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO3SAY3-Kiy8HdWTFT:-KnO3SAY3-Kiy8HdWTFU:b-iieo9g) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/formatter.cpp#L28)):*
> ```cpp
>     default:
>         UNREACHABLE();
>         return "UNREACHABLE";
> ```

Place this outside the switch.

---

*[src/common/logging/formatter.cpp, line 42 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO3eK7I0rJe_zkn7CI:-KnO3eK8obaqEVbHxn5p:b-yts46k) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/formatter.cpp#L42)):*
> ```cpp
>     const auto time_fractional = timestamp.count() % 1000000;
> 
>     msg.formatted << '[' << fmt::pad(time_seconds, 4, ' ') << '.'
> ```

I think you can use `msg.formatted.format("[...]", a, b, c)` to format this using the regular format-string syntax.

---

*[src/common/logging/formatter.cpp, line 44 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO4UrOiZQ-0OOSaXdE:-KnO4UrOiZQ-0OOSaXdF:b-km5075) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/formatter.cpp#L44)):*
> ```cpp
>     msg.formatted << '[' << fmt::pad(time_seconds, 4, ' ') << '.'
>                   << fmt::pad(time_fractional, 6, '0') << "] ";
>     msg.formatted << *msg.logger_name << " <" << GetLevelName(msg.level) << "> ";
> ```

You seem to be splitting these in different statements pretty arbitrarily, just make everything a big `<<` chain.

---

*[src/common/logging/formatter.cpp, line 50 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO4zdMMHX3YkJeX2bQ:-KnO4zdMMHX3YkJeX2bR:b-i3venf) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/formatter.cpp#L50)):*
> ```cpp
> }
> 
> } // namespace Log
> ```

Missing newline

---

*[src/common/logging/formatter.h, line 12 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO4CM1E2HyFTkBNbIj:-KnO4CM1E2HyFTkBNbIk:bn2myjp) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/formatter.h#L12)):*
> ```objc
> 
> class Formatter : public spdlog::formatter {
> ```

Remove this newline.

---

*[src/common/logging/formatter.h, line 14 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO5Cy6Q6jOGRI5OQU2:-KnO5Cy7pwlhRhIjgb2O:boso8oo) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/formatter.h#L14)):*
> ```objc
> 
> public:
>     explicit Formatter() = default;
> ```

no such thing as an explicit no-arg constructor (and I think defaulting it is redundant too)

---

*[src/common/logging/formatter.h, line 20 at r20](https://reviewable.io:443/reviews/citra-emu/citra/2707#-KnO4DrGa4WqXtOUOjhZ:-KnO4DrGa4WqXtOUOjh_:b64evrz) ([raw file](https://github.com/citra-emu/citra/blob/b4fab4bd64aedec80ef635a9e1d06d71e8447522/src/common/logging/formatter.h#L20)):*
> ```objc
> };
> 
> } // namespace Log
> ```

Missing newline at the end of file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2707#-:-KnO6oF5f1O_1gCXCnmf:b-g8azo5)*

Dragios:
Any update on this @jroweboy

jroweboy:
I still think that citra could use some logging changes as outlined in the original issue. I feel like the idea behind this PR is good, but the implementation is not. When I get some time to work on this, I'd like to revisit it and see if I can solve the main issues with the implementation.

PR #2708, opened 2017-05-17
add GetPedometerState stub

emmauss:
stubbing this service, though i dont know if anything uses it when citra is "awake".

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2708)


wwylele:
Since we have IPCHelper now, I think we should use it for every new added service function?

emmauss:
@wwylele could you explain? i am just following the format of related services in the ptm module that have already been stubbed/implemented.

emmauss:
closed for #2710

PR #2709, opened 2017-05-17
pica: use correct register value for shader bool_uniforms

wwylele:
This fixes a regression from #2695 (bug can be seen in Monster Hunter). The `value` variable is a pitfall， which doesn't contain the current register value if the command comes with a mask.

I also check other use of `value` in command processor, but all of them are used for FIFO registers (lut, shader program code, etc.), which are not likely used with a mask. It would be better if we can figure out if the mask actually have any effects to these registers, but I think we can do this after this quick fix.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2709)

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -423,7 +423,7 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
     }
 
     case PICA_REG_INDEX(gs.bool_uniforms):
-        WriteUniformBoolReg(g_state.gs, value);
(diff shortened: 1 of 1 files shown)


yuriks:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2709#-:-KkOX3MtnB2sCj2qFpMA:bnfp4nl)*

PR #2717, opened 2017-05-21
added support for loading aeskeydb.bin to the aes code

jfmherokiller:
This is an amendment to the encryption code to allow it to both load from the text file and load from aeskeydb.bin.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2717)

### DIFF
--- a/src/core/hw/aes/key.cpp
+++ b/src/core/hw/aes/key.cpp
@@ -76,7 +76,30 @@ AESKey HexToKey(const std::string& hex) {
     return key;
 }
 
-void LoadPresetKeys() {
+void SetKeySlot(const AESKey& key, size_t slot_id, char key_type) {
+    switch (key_type) {
+    case 'X':
+        if (!key_slots.at(slot_id).x.is_initialized()) {
+            SetKeyX(slot_id, key);
+        }
+        break;
+    case 'Y':
+        if (!key_slots.at(slot_id).y.is_initialized()) {
+            SetKeyY(slot_id, key);
+        }
+        break;
+    case 'N':
+        if (!key_slots.at(slot_id).normal.is_initialized()) {
+            SetNormalKey(slot_id, key);
+        }
+        break;
+    default:
+        LOG_ERROR(HW_AES, "Invalid key type %c", key_type);
+        break;
+    }
+}
+
+void LoadKeysFromText() {
     const std::string filepath = FileUtil::GetUserPath(D_SYSDATA_IDX) + AES_KEYS;
     FileUtil::CreateFullPath(filepath); // Create path if not already created
     std::ifstream file;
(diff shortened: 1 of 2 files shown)


wwylele:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/core/hw/aes/key.cpp, line 100 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2717#-KkgLxFL4w-5XezZVUgU:-KkgLxFL4w-5XezZVUgV:bignpyj) ([raw file](https://github.com/citra-emu/citra/blob/46111d305f433019cc5e3781580f4f166331c283/src/core/hw/aes/key.cpp#L100)):*
> ```cpp
>         break;
>     }
> }
> ```

Needs blank line between functions

---

*[src/core/hw/aes/key.cpp, line 155 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2717#-KkgMJvv2PuP7jnX1MHB:-KkgMJvv2PuP7jnX1MHC:bjqfxfw) ([raw file](https://github.com/citra-emu/citra/blob/46111d305f433019cc5e3781580f4f166331c283/src/core/hw/aes/key.cpp#L155)):*
> ```cpp
>  * The wiki describes that this database can be encrypted or decrypted however as of right now only
>  * the decrypted format is supported.
>  */
> ```

If this is a doc-comments, start it with `/**`; other wise use `//` instead.

---

*[src/core/hw/aes/key.cpp, line 171 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2717#-KkgNJh9WLuJ3c2PHguF:-KkgNJh9WLuJ3c2PHguG:b-omp604) ([raw file](https://github.com/citra-emu/citra/blob/46111d305f433019cc5e3781580f4f166331c283/src/core/hw/aes/key.cpp#L171)):*
> ```cpp
>     FileUtil::CreateFullPath(filepath); // Create path if not already created
>     std::ifstream file;
>     OpenFStream(file, filepath, std::ios_base::in);
> ```

needs `ios::binary` since this is a binary file unlike the text one. Probably you want to use `FileUtil::IOFile`

---

*[src/core/hw/aes/key.cpp, line 178 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2717#-KkgMnTar20WOf53nqWI:-KkgMnTar20WOf53nqWJ:bvfwk07) ([raw file](https://github.com/citra-emu/citra/blob/46111d305f433019cc5e3781580f4f166331c283/src/core/hw/aes/key.cpp#L178)):*
> ```cpp
>     while (!file.eof()) {
>         file.read((char*)&key_info, sizeof(key_info));
>         SetKeySlot(key_info.key, key_info.slot, key_info.type);
> ```

I am not familiar with keydb format standard, but you should probably respect to other fields lie `id`, `is_encrypted` etc. Should just skip keys that doesn't match any slot in citra

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2717)*

src/core/hw/aes/key.cpp
```diff
@@ -121,22 +143,45 @@ void LoadPresetKeys() {
...
+    const std::string filepath = FileUtil::GetUserPath(D_SYSDATA_IDX) + AES_KEYS_DB;
+    FileUtil::CreateFullPath(filepath); // Create path if not already created
+    std::ifstream file;
+    OpenFStream(file, filepath, std::ios_base::in);
+    if (!file) {
+        return;
+    }
+    AesKeyInfo key_info;
+    while (!file.eof()) {
+        file.read((char*)&key_info, sizeof(key_info));
```
lioncash: `reinterpret_cast`

src/core/hw/aes/key.cpp
```diff
@@ -121,22 +143,45 @@ void LoadPresetKeys() {
...
+
+    const std::string filepath = FileUtil::GetUserPath(D_SYSDATA_IDX) + AES_KEYS_DB;
+    FileUtil::CreateFullPath(filepath); // Create path if not already created
+    std::ifstream file;
+    OpenFStream(file, filepath, std::ios_base::in);
+    if (!file) {
+        return;
+    }
+    AesKeyInfo key_info;
+    while (!file.eof()) {
```
lioncash: This should likely be:

```cpp
while (file.read(reinterpret_cast<char*>(&key_info), sizeof(key_info)))
{
  SetKeySlot(key_info.key, key_info.slot, key_info.type);
}
```

so that the body is only ever entered if the full read succeeds. That or have some other form of error checking.

Currently, if a read fails (either due to a corrupt file or something else),  junk data ends up being set into the key slot.

i.e.

```cpp
while (!file.eof()) // Assume this condition is true
{
  // Assume this fails, the eof/status bit(s) are set, but all data didn't get read into key_info
  file.read(reinterpret_cast<char*>(&key_info), sizeof(key_info));

  // Junk data gets set into the keyslot.
  SetKeySlot(key_info.key, key_info.slot, key_info.type);
}
```

src/core/hw/aes/key.cpp
```diff
@@ -121,22 +143,45 @@ void LoadPresetKeys() {
...
+        char type;               // type 'X' / 'Y' / 'N' for normalKey
+        std::array<char, 10> id; // key ID for special keys, all zero for standard keys
+        u8 reserved[2];          // reserved space
+        u8 is_devkitkey;         // 0 for retail units / 1 for DevKit units
+        u8 is_encrypted;         // 0 if not / anything else if it is
+        AESKey key;
+    };
+
+    const std::string filepath = FileUtil::GetUserPath(D_SYSDATA_IDX) + AES_KEYS_DB;
+    FileUtil::CreateFullPath(filepath); // Create path if not already created
```
lioncash: This isn't always guaranteed to succeed and should be checked against. Also comments that don't document a struct/class member are preferred on top of code.

fox8091:
Perhaps also allow key loading from boot9 dumps now?

B3n30:
What's the status of this PR?

B3n30:
Closed, due to inactivity.

PR #2721, opened 2017-05-23
swrasterizer: implemented TextureCube

wwylele:
Currently only implemented in the software renderer, not in gl one. I have some questions regarding to this implementation and incoming gl implementation:
 - `// TODO: Refactor so cubemaps and shadowmaps can be handled` what kind of refactor is preferred here?
 - When implementing in gl, should gl's cube mapping be used, or we store them in 6 texture2D and manually convert the coordinates?

And [here](https://www.dropbox.com/s/l6zzmkiw4gh1eoq/texture_box.zip?dl=1) is a testing program

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2721)

### DIFF
--- a/src/video_core/regs_texturing.h
+++ b/src/video_core/regs_texturing.h
@@ -133,7 +133,32 @@ struct TexturingRegs {
         BitField<16, 1, u32> clear_texture_cache; // TODO: unimplemented
     } main_config;
     TextureConfig texture0;
-    INSERT_PADDING_WORDS(0x8);
+
+    enum class CubeFace {
+        PositiveX = 0,
+        NegativeX = 1,
+        PositiveY = 2,
+        NegativeY = 3,
+        PositiveZ = 4,
+        NegativeZ = 5,
+    };
+
+    BitField<0, 22, u32> cube_address[5];
+
+    PAddr GetCubePhysicalAddress(CubeFace face) const {
+        PAddr address = texture0.address;
+        if (face != CubeFace::PositiveX) {
+            // Bits [22:27] from the main texture address is shared with all cubemap additional
+            // addresses.
+            auto& face_addr = cube_address[static_cast<size_t>(face) - 1];
+            address &= ~face_addr.mask;
+            address |= face_addr;
+        }
+        // A multiplier of 8 is also needed in the same way as the main address.
+        return address * 8;
+    }
+
+    INSERT_PADDING_WORDS(0x3);
     BitField<0, 4, TextureFormat> texture0_format;
     BitField<0, 1, u32> fragment_lighting_enable;
     INSERT_PADDING_WORDS(0x1);
--- a/src/video_core/swrasterizer/rasterizer.cpp
+++ b/src/video_core/swrasterizer/rasterizer.cpp
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <array>
 #include <cmath>
+#include <tuple>
 #include "common/assert.h"
 #include "common/bit_field.h"
 #include "common/color.h"
@@ -69,6 +70,49 @@ static int SignedArea(const Math::Vec2<Fix12P4>& vtx1, const Math::Vec2<Fix12P4>
     // TODO: There is a very small chance this will overflow for sizeof(int) == 4
     return Math::Cross(vec1, vec2).z;
 };
+
+/// Convert a 3D vector for cube map coordinates to 2D texture coordinates along with the face name
+static std::tuple<float24, float24, PAddr> ConvertCubeCoord(float24 u, float24 v, float24 w,
+                                                            const TexturingRegs& regs) {
+    const float abs_u = std::abs(u.ToFloat32());
+    const float abs_v = std::abs(v.ToFloat32());
+    const float abs_w = std::abs(w.ToFloat32());
+    float24 x, y, z;
+    PAddr addr;
+    if (abs_u > abs_v && abs_u > abs_w) {
+        if (u > float24::FromFloat32(0)) {
+            addr = regs.GetCubePhysicalAddress(TexturingRegs::CubeFace::PositiveX);
+            y = -v;
+        } else {
+            addr = regs.GetCubePhysicalAddress(TexturingRegs::CubeFace::NegativeX);
+            y = v;
+        }
+        x = -w;
+        z = u;
+    } else if (abs_v > abs_w) {
+        if (v > float24::FromFloat32(0)) {
+            addr = regs.GetCubePhysicalAddress(TexturingRegs::CubeFace::PositiveY);
+            x = u;
+        } else {
+            addr = regs.GetCubePhysicalAddress(TexturingRegs::CubeFace::NegativeY);
+            x = -u;
+        }
+        y = w;
+        z = v;
+    } else {
+        if (w > float24::FromFloat32(0)) {
+            addr = regs.GetCubePhysicalAddress(TexturingRegs::CubeFace::PositiveZ);
+            y = -v;
+        } else {
+            addr = regs.GetCubePhysicalAddress(TexturingRegs::CubeFace::NegativeZ);
+            y = v;
+        }
+        x = u;
+        z = w;
+    }
+    const float24 half = float24::FromFloat32(0.5f);
+    return std::make_tuple(x / z * half + half, y / z * half + half, addr);
+}
 
 MICROPROFILE_DEFINE(GPU_Rasterization, "GPU", "Rasterization", MP_RGB(50, 50, 240));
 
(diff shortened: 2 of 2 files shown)


src/video_core/regs_texturing.h
```diff
@@ -133,7 +133,25 @@ struct TexturingRegs {
...
+    enum class CubeFace {
+        PositiveX = 0,
+        NegativeX = 1,
+        PositiveY = 2,
+        NegativeY = 3,
+        PositiveZ = 4,
+        NegativeZ = 5,
+    };
+
+    BitField<0, 22, u32> cube_address[5];
```
wwylele: I am not sure if I should use `std::array` here. As far as I see all existing arrays defined in regs are raw array, so I followed the convention here

bunnei: Consistency here is fine IMO... although they probably can all be updated to std::array in a separate change

src/video_core/regs_texturing.h
```diff
@@ -133,7 +133,25 @@ struct TexturingRegs {
...
+
+    PAddr GetCubePhysicalAddress(CubeFace face) const {
+        if (face == CubeFace::PositiveX)
+            return texture0.GetPhysicalAddress();
+        return ((texture0.address & 0xFC00000) | cube_address[static_cast<size_t>(face) - 1]) * 8;
```
bunnei: * What is 0xFC0000? Should it be a constant? Ditto with 8
* Perhaps add a comment explaining what's going on here?
* Should we do any bounds checking/assert/etc. on `face`?

src/video_core/swrasterizer/rasterizer.cpp
```diff
@@ -69,6 +70,49 @@ static int SignedArea(const Math::Vec2<Fix12P4>& vtx1, const Math::Vec2<Fix12P4>
     return Math::Cross(vec1, vec2).z;
 };
 
+/// Convert a 3D vector for cube map coordinates to 2D texture coordinates along with the face name
+static std::tuple<float24, float24, TexturingRegs::CubeFace> ConvertCubeCoord(float24 u, float24 v,
+                                                                              float24 w) {
+    float abs_u = std::abs(u.ToFloat32());
```
bunnei: These can be `const`

src/video_core/swrasterizer/rasterizer.cpp
```diff
@@ -69,6 +70,49 @@ static int SignedArea(const Math::Vec2<Fix12P4>& vtx1, const Math::Vec2<Fix12P4>
...
+    float abs_v = std::abs(v.ToFloat32());
+    float abs_w = std::abs(w.ToFloat32());
+    float24 x, y, z;
+    TexturingRegs::CubeFace face;
```
bunnei: nit: Always prefer to initialize variables to prevent unpredictable behavior, although in this case these will always get set

src/video_core/regs_texturing.h
wwylele: - The five additional address share some higher bits with the main address. `0xFC00000` is the mask of the shared part. I will change this to be generated by the bitfield length of addresses.
 - `8` is the common address multiplier. See `GetPhysicalAddress` above. 
 - Will add some comments, of course
 - since face is already a `enum class`, which has a natural boundary, I think the checking is unnecessary because one shouldn't use a value out from the enum anyway.

wwylele:
Addressed comments above. For the address registers, I made a separate bitfield for the shared part of the main address and made it easier to read

yuriks:
Reviewed 2 of 2 files at r3, 2 of 2 files at r4.

---

*[src/video_core/regs_texturing.h, line 151 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2721#-KkpOeX5P-d5Wh0jQlX5-r2-151:-KlFwIh5JR5Tx-otn7nY:b-8boboi) ([raw file](https://github.com/citra-emu/citra/blob/375ae9c2068c6712b9f9a34e3977175eff62da76/src/video_core/regs_texturing.h#L151)):*

 - The five additional address share some higher bits with the main address. `0xFC00000` is the mask of the shared part. I will change this to be generated by the bitfield length of addresses.
 - `8` is the common address multiplier. See `GetPhysicalAddress` above. 
 - Will add some comments, of course
 - since face is already a `enum class`, which has a natural boundary, I think the checking is unnecessary because one shouldn't use a value out from the enum anyway.
</blockquote></details>

I think that not using `texture0.GetPhysicalAddress` and instead calculating face 0 in here might make it more obvious how the addresses relate to each other, though your current version isn't bad either:
```c++
PAddr GetCubePhysicalAddress(CubeFace face) const {
    PAddr address = texture0.address;
    if (face != CubeFace::PositiveX) {
        auto& face_addr = cube_address[static_cast<size_t>(face) - 1];
        address &= ~face_addr.mask;
        address |= face_addr;
    }
    return address * 8;
}
```

---

*[src/video_core/regs_texturing.h, line 151 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2721#-KlFy7dqzgAuYS-AEOz-:-KlFy7dqzgAuYS-AEOz0:bud27ss) ([raw file](https://github.com/citra-emu/citra/blob/26908ab39d3e8e66e1ec13f7c45057f4dfd75fe8/src/video_core/regs_texturing.h#L151)):*
> ```objc
> 
>     BitField<0, 22, u32> cube_address[5];
> ```

Roll this change (trimming whitespace) into the previous commit.

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 80 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2721#-KkpOeX5P-d5Wh0jQlX6-r2-80:-KlG-ex6F-dwxkXiq4FY:bbgdui5) ([raw file](https://github.com/citra-emu/citra/blob/375ae9c2068c6712b9f9a34e3977175eff62da76/src/video_core/swrasterizer/rasterizer.cpp#L80)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

nit: Always prefer to initialize variables to prevent unpredictable behavior, although in this case these will always get set
</blockquote></details>

If you expect to set a variable in every branch, then isn't it better to leave it un-initialized, allowing the compiler to give you warnings if you forget to do it in some branch?

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 384 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2721#-KlFzdV1zZy05viZ3cik:-KlFzdV2Kx7iEGQN2ktO:b-cfc9tk) ([raw file](https://github.com/citra-emu/citra/blob/26908ab39d3e8e66e1ec13f7c45057f4dfd75fe8/src/video_core/swrasterizer/rasterizer.cpp#L384)):*
> ```cpp
>                         texture_data =
>                             Memory::GetPhysicalPointer(texture.config.GetPhysicalAddress());
>                     }
> ```

Instead adding this `is_cube` variable, add a texture_address variable, initialize it with `texture.config.GetPhysicalAddress()` before the `if (i == 0)`. In the case of cube texturing simply replace the address with the correct one inside the switch, removing the need to have two cases here. Might also be worth moving the computation of the address inside `ConvertCubeCoord`, because then you'll always be passing a constant value into `GetCubePhysicalAddress`, making the branching effectively free since it'll be inlined.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2721)*

wwylele:
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/video_core/regs_texturing.h, line 151 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2721#-KkpOeX5P-d5Wh0jQlX5-r2-151:-KlKVbEAChbEwe1bsJEB:b-hiuolz) ([raw file](https://github.com/citra-emu/citra/blob/375ae9c2068c6712b9f9a34e3977175eff62da76/src/video_core/regs_texturing.h#L151)):*

I think that not using `texture0.GetPhysicalAddress` and instead calculating face 0 in here might make it more obvious how the addresses relate to each other, though your current version isn't bad either:
```c++
PAddr GetCubePhysicalAddress(CubeFace face) const {
    PAddr address = texture0.address;
    if (face != CubeFace::PositiveX) {
        auto& face_addr = cube_address[static_cast<size_t>(face) - 1];
        address &= ~face_addr.mask;
        address |= face_addr;
    }
    return address * 8;
}
```
</blockquote></details>

Done. This PR was before BitField revise so didn't make use of the `mask` field.

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 384 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2721#-KlFzdV1zZy05viZ3cik:-KlKZowwWXSto0GsLk7z:b-zeom22) ([raw file](https://github.com/citra-emu/citra/blob/26908ab39d3e8e66e1ec13f7c45057f4dfd75fe8/src/video_core/swrasterizer/rasterizer.cpp#L384)):*

Instead adding this `is_cube` variable, add a texture_address variable, initialize it with `texture.config.GetPhysicalAddress()` before the `if (i == 0)`. In the case of cube texturing simply replace the address with the correct one inside the switch, removing the need to have two cases here. Might also be worth moving the computation of the address inside `ConvertCubeCoord`, because then you'll always be passing a constant value into `GetCubePhysicalAddress`, making the branching effectively free since it'll be inlined.
</blockquote></details>

Done, but didn't move address computation here, because it might also be shared with gl renderer (or, at least as a documentation, can give the idea when one want to implement it in the gl renderer)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2721)*

wwylele:
oops, something wrong

wwylele:
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 384 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2721#-KlFzdV1zZy05viZ3cik:-KlKf06FiH9UXXyzA6zo:b-63vqb7) ([raw file](https://github.com/citra-emu/citra/blob/26908ab39d3e8e66e1ec13f7c45057f4dfd75fe8/src/video_core/swrasterizer/rasterizer.cpp#L384)):*

Done, but didn't move address computation here, because it might also be shared with gl renderer (or, at least as a documentation, can give the idea when one want to implement it in the gl renderer)
</blockquote></details>

nvm about the comment of address computation. fixed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2721)*

yuriks:
Reviewed 2 of 2 files at r5, 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2721)*

yuriks:
LGTM

PR #2725, opened 2017-05-26
gl_shader: refactor texture sampler into its own function

wwylele:
It is confirmed that bump map respects to all "special features" of texture units: 
 - different texture type for unit 0 ([here is a testing program](https://gist.github.com/wwylele/c939cc90551b0c28f1298bf08a026583) that uses projection texture for bump map)
 - different coordinates for unit 2 (tested before, see #2686)
 - proctex for unit 3. . (tested when implementing proctex)

So it should use the same sampler function as combiner's

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2725)

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_gen.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_gen.cpp
@@ -144,12 +144,40 @@ static bool IsPassThroughTevStage(const TevStageConfig& stage) {
             stage.GetColorMultiplier() == 1 && stage.GetAlphaMultiplier() == 1);
 }
 
-static std::string TexCoord(const PicaShaderConfig& config, int texture_unit) {
-    if (texture_unit == 2 && config.state.texture2_use_coord1) {
-        return "texcoord[1]";
+static std::string SampleTexture(const PicaShaderConfig& config, unsigned texture_unit) {
+    const auto& state = config.state;
+    switch (texture_unit) {
+    case 0:
+        // Only unit 0 respects the texturing type
+        switch (state.texture0_type) {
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -144,12 +144,40 @@ static bool IsPassThroughTevStage(const TevStageConfig& stage) {
...
 }
 
-static std::string TexCoord(const PicaShaderConfig& config, int texture_unit) {
-    if (texture_unit == 2 && config.state.texture2_use_coord1) {
-        return "texcoord[1]";
+static std::string SampleTexture(const PicaShaderConfig& config, unsigned texture_unit) {
+    const auto& state = config.state;
+    switch (texture_unit) {
+    case 0:
+        // Only unit 0 respects the texturing type (according to 3DBrew)
```
bunnei: nit: If you're gonna source 3Dbrew, include link to relevant article?

wwylele: this is an already existing comment moved from below. I am not sure which page it was referring to. Or can I just remove it because this is now a well-known info?

bunnei: Sure, feel free to just remove it then.

Subv:
Lgtm

yuriks:
Random thought: It would be nice if instead of generating these fetches on-demand, we'd pre-sample every active texture to a local variable, so the values can just be re-used if they're used more than once.

PR #2732, opened 2017-05-27
Add the fmt string formatting library

yuriks:
More info at http://fmtlib.net/

This was based on @jroweboy's work on his spdlog branch, but with modifications. Namely, I don't manually add the include directory for fmt. Rather, individual libraries should add it using `target_link_libraries` instead.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2732)


yuriks:
To clarify, this is splitting the fmt inclusion from #2707 so that it can be merge separately without being held up by that PR.

PR #2734, opened 2017-05-28
CMake: Use CMake target properties for all libraries

yuriks:
My goal was to get rid of every one-off `include_directories` scattered around. 🗡 

Each commits makes sense individually, but there's a lot of them, so YMMV on how to review it.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2734)


yuriks:
FYI: The reason for the Travis build failure is that the old version of CMake is still in the cache. I'm going to clear the cache before merging this so that the next builds will suceed. (Appveyor seems like just a spurious failure.)

Issue #2742, opened 2017-06-01
Metal Gear Solid 3D: Snake Eater - Crashes when emulating Title Screen


Dragios:
It boots to the title screen without any problem before this. PR #2703 changes its behaviour and causes it to crash instead.

Log that appear when crash:
```
Render.OpenGL <Critical> video_core/renderer_opengl/pica_to_gl.h:PicaToGL::WrapMode:62: Unknown texture wrap mode 6
Debug <Critical> video_core/renderer_opengl/pica_to_gl.h:PicaToGL::WrapMode::<lambda_4475911c1a7164258d047613d97648c2>::operator ():63: Assertion Failed!
Unreachable code!
```
![Disassembly](https://cloud.githubusercontent.com/assets/18496868/26683311/f02f8cb8-4715-11e7-901f-a00c56313c12.png)
![Call Stack](https://cloud.githubusercontent.com/assets/18496868/26683346/13982c32-4716-11e7-97e3-c6cd94a33301.png)

Dragios:
Closed in #2743

Issue #2740, opened 2017-05-31
Emulator crash in Ocarina of Time 3D


marouf33:
Hi, I have encountered a crash in the emulator . It is reproduced consistently in 	"Nightly Build - 54ea95c".
The crash happens after Sheik is revealed as Zelda in the temple of time. 
I'm sorry I can't provide a backtrace, but I will attach the save file to right before the scene of the crash happens.

[00000001.zip](https://github.com/citra-emu/citra/files/1042014/00000001.zip)

jroweboy:
Here's a summary of the issue:

There's an assert here https://github.com/citra-emu/citra/blob/54ea95cca74f76777aa156f6d750c9d1ba5b6f11/src/video_core/renderer_opengl/pica_to_gl.h#L32 which prevents any texture modes from getting used that aren't in the enum. The problem is there's a messed up config being passed in http://i.imgur.com/bayduef.png with some clearly bogus values in it.

wwylele:
Ref #2703. This is a regression due to length expansion of the wrap mode fields, and the game seems using invalid wrap modes 6 and 7.

Abyss2381:
I was able to work around this by downloading bleeding-edge-300, v 0.1.300. I tried several between it and the current build, but they all failed at the same spot. If you download and run this one, continue your save, you should be able to get past the point and save again. I will test the rest of OoT 3D this afternoon to see if any more crashes occur with the newest build.

Abyss2381:
Okay, it will crash again in the final credits, when young Link is approaching Zelda at the window in Hyrule Castle. Same error, so only those two spots are affected.

jroweboy:
Fixed in https://github.com/citra-emu/citra/pull/2743 (Please retest on nightly if you wish)

PR #2743, opened 2017-06-01
pica/rasterizer: implement/stub texture wrap mode 4-7

wwylele:
This is intended to avoid crash in #2740 and #2742.

(Remade after bunnei's comment)
Based on previous hardware test, mode 4-7 is implemented in the software rasterizer, and partially implemented in gl rasterizer. Mode 4 and 5 doesn't have their direct counterpart in opengl, so they are not fully implemented. Also added warning log for these mode in opengl (not added for sw one because it would be too verbose)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2743)

### DIFF
--- a/src/video_core/renderer_opengl/pica_to_gl.h
+++ b/src/video_core/renderer_opengl/pica_to_gl.h
@@ -55,6 +55,12 @@ inline GLenum WrapMode(Pica::TexturingRegs::TextureConfig::WrapMode mode) {
         GL_CLAMP_TO_BORDER, // WrapMode::ClampToBorder
         GL_REPEAT,          // WrapMode::Repeat
         GL_MIRRORED_REPEAT, // WrapMode::MirroredRepeat
+        // TODO(wwylele): ClampToEdge2 and ClampToBorder2 are not properly implemented here. See the
+        // comments in enum WrapMode.
+        GL_CLAMP_TO_EDGE,   // WrapMode::ClampToEdge2
+        GL_CLAMP_TO_BORDER, // WrapMode::ClampToBorder2
+        GL_REPEAT,          // WrapMode::Repeat2
+        GL_REPEAT,          // WrapMode::Repeat3
     };
 
     // Range check table for input
@@ -63,6 +69,13 @@ inline GLenum WrapMode(Pica::TexturingRegs::TextureConfig::WrapMode mode) {
         UNREACHABLE();
 
         return GL_CLAMP_TO_EDGE;
+    }
+
+    if (static_cast<u32>(mode) > 3) {
+        // It is still unclear whether mode 4-7 are valid, so log it if a game uses them.
+        // TODO(wwylele): telemetry should be added here so we can collect more info about which
+        // game uses this.
+        LOG_WARNING(Render_OpenGL, "Using texture wrap mode %u", static_cast<u32>(mode));
     }
 
     GLenum gl_mode = wrap_mode_table[mode];
--- a/src/video_core/regs_texturing.h
+++ b/src/video_core/regs_texturing.h
@@ -30,10 +30,10 @@ struct TexturingRegs {
             Repeat = 2,
             MirroredRepeat = 3,
             // Mode 4-7 produces some weird result and may be just invalid:
-            // 4: Positive coord: clamp to edge; negative coord: repeat
-            // 5: Positive coord: clamp to border; negative coord: repeat
-            // 6: Repeat
-            // 7: Repeat
+            ClampToEdge2 = 4,   // Positive coord: clamp to edge; negative coord: repeat
+            ClampToBorder2 = 5, // Positive coord: clamp to border; negative coord: repeat
+            Repeat2 = 6,        // Same as Repeat
+            Repeat3 = 7,        // Same as Repeat
         };
 
         enum TextureFilter : u32 {
--- a/src/video_core/swrasterizer/rasterizer.cpp
+++ b/src/video_core/swrasterizer/rasterizer.cpp
@@ -307,10 +307,22 @@ static void ProcessTriangleInternal(const Vertex& v0, const Vertex& v1, const Ve
                 int t = (int)(v * float24::FromFloat32(static_cast<float>(texture.config.height)))
                             .ToFloat32();
 
-                if ((texture.config.wrap_s == TexturingRegs::TextureConfig::ClampToBorder &&
-                     (s < 0 || static_cast<u32>(s) >= texture.config.width)) ||
-                    (texture.config.wrap_t == TexturingRegs::TextureConfig::ClampToBorder &&
-                     (t < 0 || static_cast<u32>(t) >= texture.config.height))) {
+                bool use_border_s = false;
+                bool use_border_t = false;
+
+                if (texture.config.wrap_s == TexturingRegs::TextureConfig::ClampToBorder) {
+                    use_border_s = s < 0 || s >= static_cast<int>(texture.config.width);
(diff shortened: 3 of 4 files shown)


src/video_core/renderer_opengl/pica_to_gl.h
```diff
@@ -55,6 +55,12 @@ inline GLenum WrapMode(Pica::TexturingRegs::TextureConfig::WrapMode mode) {
         GL_CLAMP_TO_BORDER, // WrapMode::ClampToBorder
         GL_REPEAT,          // WrapMode::Repeat
         GL_MIRRORED_REPEAT, // WrapMode::MirroredRepeat
+        // The four modes below shouldn't be used in normal situation, so assign dummy values to
+        // them and report the error later.
+        0, // Positive coord: clamp to edge; negative coord: repeat
```
bunnei: What is the point of these comments if these are unknown?

wwylele: The comments are observed behaviour of these modes (copy-pasted from the comment of the registers). It is unknown whether they are intended or just glitched out.

bunnei: Observed behavior on hardware? Do we expect lots of graphical issues now that they are being emulated as GL_CLAMP_TO_EDGE?

wwylele: Yes they were hardware-tested. I actually want to see graphical issues in future reports, which can help us identify the cause of using this mode (intentional, by game bug, or by citra bug)

bunnei: IMO, if we have already defined how these behave on hardware, then we should just implement them correctly... Instead of waiting for bugs to be reported, then fixing them re-actively. This could potentially happen years later when no one remembers about this, and end up getting fixed way later than they should (or wasting someones time debugging). Even if it's potentially masking some other Citra bug, there is nothing stopping a buggy game (or homebrew) from using these legitimately.

Wanting to know when games use these to debug potential Citra bugs seems like it can/should be separate from actually implementing the feature as tested on HW. Right now, we don't have a *great* way to do this, since things like ASSERT or LOG_CRITICAL will mostly be ignored by users if the game works and looks fine. However, this type of thing is an excellent use of our new telemetry framework (once it's ready)... i.e. we will be able to log this condition as a user test case, and automatically see any games that hit it (and debug further if we'd like).

All that being said, I don't feel *super* strongly about this, so will sign off. However, I suggest doing one of the following:
* Implementing these properly, put a comment in the code discussing that these shouldn't normally be hit but are tested on HW or whatever, putting a ASSERT and/or LOG_FOO if you'd like, and creating an issue that we should add telemetry here (and putting it in a TODO comment), which can then be assigned to me if you'd like
* Create an issue to revisit this (regardless of if we get error reports in), put a "TODO(wwylele): Issue XYZ" in the code, and document more thoroughly that we know the HW behavior of these but that they are intentionally left unimplemented so that we can get better test data

wwylele: Sounds good to me. Will do tomorrow

yuriks:
:lgtm:

---

Reviewed 4 of 4 files at r2.

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 316 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2743#-KlloBYPmycpVHVT9iWP:-KlloBYPmycpVHVT9iWQ:b5l6v5n) ([raw file](https://github.com/citra-emu/citra/blob/4800a7ff889548ca3bf1f2546868ecaa5dde17e4/src/video_core/swrasterizer/rasterizer.cpp#L316)):*
> ```cpp
>                     use_border_s = s < 0 || static_cast<u32>(s) >= texture.config.width;
>                 } else if (texture.config.wrap_s == TexturingRegs::TextureConfig::ClampToBorder2) {
>                     use_border_s = s > 0 && static_cast<u32>(s) >= texture.config.width;
> ```

Isn't the `s > 0` part redundant?

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 322 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2743#-KlloJgOJyQwW-U7ts_X:-KlloJgOJyQwW-U7ts_Y:b-6gcu0u) ([raw file](https://github.com/citra-emu/citra/blob/4800a7ff889548ca3bf1f2546868ecaa5dde17e4/src/video_core/swrasterizer/rasterizer.cpp#L322)):*
> ```cpp
>                     use_border_t = t < 0 || static_cast<u32>(t) >= texture.config.height;
>                 } else if (texture.config.wrap_t == TexturingRegs::TextureConfig::ClampToBorder2) {
>                     use_border_t = t > 0 && static_cast<u32>(t) >= texture.config.height;
> ```

Same, isn't `t > 0` redundant?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2743#-:-Kllp6BKj-a9IX4-bdIs:bnfp4nl)*

wwylele:

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 316 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2743#-KlloBYPmycpVHVT9iWP:-Kllqx8WwuCTPRCdB9vo:b-x7bkwt) ([raw file](https://github.com/citra-emu/citra/blob/4800a7ff889548ca3bf1f2546868ecaa5dde17e4/src/video_core/swrasterizer/rasterizer.cpp#L316)):*

Isn't the `s > 0` part redundant?
</blockquote></details>

cleaned. was being dumb about that `static_cast<u32>`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2743)*

yuriks:
Reviewed 1 of 1 files at r3.

---

*[src/video_core/swrasterizer/rasterizer.cpp, line 322 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2743#-KlloJgOJyQwW-U7ts_X:-Klm07m-TqR3m8cCfe7u:b32w0x) ([raw file](https://github.com/citra-emu/citra/blob/4800a7ff889548ca3bf1f2546868ecaa5dde17e4/src/video_core/swrasterizer/rasterizer.cpp#L322)):*

Same, isn't `t > 0` redundant?
</blockquote></details>

OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2743)*

src/video_core/renderer_opengl/pica_to_gl.h
```diff
@@ -55,6 +55,12 @@ inline GLenum WrapMode(Pica::TexturingRegs::TextureConfig::WrapMode mode) {
         GL_CLAMP_TO_BORDER, // WrapMode::ClampToBorder
         GL_REPEAT,          // WrapMode::Repeat
         GL_MIRRORED_REPEAT, // WrapMode::MirroredRepeat
+        // TODO(wwylele): ClampToEdge2 and ClampToBorder2 are not properly implemented here. See the
+        // comments in enum WrapMode.
+        GL_CLAMP_TO_EDGE,   // WrapMode::ClampToEdge2
```
Subv: Can we handle the clamping for these modes manually in the frag shader?

wwylele: yes, but I want to leave this to next PR, as I want to do a quick fix here to a crash regressed from my previous PR.

wwylele: And I still doubt if these modes are valid modes. I don't feel like to bloat the code logic to implement them before we can confirm games use them purposely.

PR #2754, opened 2017-06-06
Handle service registrations using sm/srv

yuriks:
Depends on:
 - [x] #2752 HLE: Move SessionRequestHandler from Service:: to Kernel::
 - [ ] #2753 Add SetHleHandler to ServerPort/ServerSession

Moves the srv/sm implementation to a sub-directory and tasks it with handling service registrations.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2754)

### DIFF
--- a/src/core/hle/service/sm/srv.cpp
+++ b/src/core/hle/service/sm/srv.cpp
@@ -9,10 +9,10 @@
 #include "core/hle/kernel/client_session.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/kernel/server_session.h"
-#include "core/hle/service/srv.h"
+#include "core/hle/service/sm/srv.h"
 
 namespace Service {
-namespace SRV {
+namespace SM {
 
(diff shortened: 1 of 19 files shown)


src/core/hle/service/sm/srv.cpp
```diff
@@ -7,14 +7,16 @@
...
 #include "core/hle/kernel/server_session.h"
 #include "core/hle/service/sm/srv.h"
 
 namespace Service {
 namespace SM {
 
-static Kernel::SharedPtr<Kernel::Event> event_handle;
+constexpr int MAX_PENDING_NOTIFICATIONS = 16;
+
+static Kernel::SharedPtr<Kernel::Semaphore> notification_semaphore;
```
Subv: Is this really a semaphore, though?

yuriks: Yes (checked in the code, it's where the 16 comes from too)

PR #2756, opened 2017-06-07
New service framework

yuriks:
From the main commit's description:

> The old "Interface" class had a few problems such as using free
> functions (Which didn't allow you to write the service handler as if it
> were a regular class.) which weren't very extensible. (Only received one
> parameter with a pointer to the Interface object.)
> 
> The new ServiceFramework aims to solve these problems by working with
> member functions and passing a generic context struct as parameter. This
> struct can be extended in the future without having to update all
> existing service implementations.

Please hold off on actually migrating stuff to this new framework for now. I have some other API changes planned/half-implemented which, while backwards compatible, should save a lot of work when writing handlers, so it's probably better to wait for them.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2756)

### DIFF
--- a/src/core/hle/service/service.h
+++ b/src/core/hle/service/service.h
@@ -18,10 +18,15 @@
 
 namespace Kernel {
 class ClientPort;
+class ServerPort;
 class ServerSession;
 }
 
 namespace Service {
+
+namespace SM {
+class ServiceManager;
+}
 
 static const int kMaxPortSize = 8; ///< Maximum size of a port name (8 characters)
 /// Arbitrary default number of maximum connections to an HLE service.
@@ -30,6 +35,9 @@ static const u32 DefaultMaxSessions = 10;
 /**
  * Framework for implementing HLE service handlers which dispatch incoming SyncRequests based on a
  * table mapping header ids to handler functions.
+ *
+ * @deprecated Use ServiceFramework for new services instead. It allows services to be stateful and
+ *     is more extensible going forward.
  */
 class Interface : public Kernel::SessionRequestHandler {
 public:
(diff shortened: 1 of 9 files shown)


src/core/hle/service/service.h
```diff
@@ -101,6 +109,147 @@ class Interface : public Kernel::SessionRequestHandler {
...
+    struct FunctionInfoBase {
+        u32 expected_header;
+        HandlerFnP<ServiceFrameworkBase> handler_callback;
+        const char* name;
+    };
+
+    using InvokerFn = void(ServiceFrameworkBase* object, HandlerFnP<ServiceFrameworkBase> member,
+                           Kernel::HLERequestContext& ctx);
+
+    ServiceFrameworkBase(const char* service_name, unsigned int max_sessions,
```
lioncash: Should these this and the class member be using `u32` instead to be consistent with the other functions?

src/core/hle/service/service.h
```diff
@@ -101,6 +109,147 @@ class Interface : public Kernel::SessionRequestHandler {
...
+                  // Type-erase member function pointer by casting it down to the base class.
+                  static_cast<HandlerFnP<ServiceFrameworkBase>>(handler_callback), name} {}
+    };
+
+    /**
+     * Initializes the handler with no functions installed.
+     * @param max_sessions Maximum number of sessions that can be
+     * connected to this service at the same time.
+     */
+    ServiceFramework(const char* service_name, unsigned int max_sessions = DefaultMaxSessions)
```
lioncash: ditto

src/core/hle/service/sm/srv.h
```diff
@@ -4,21 +4,33 @@
...
 namespace Service {
 namespace SM {
 
 /// Interface to "srv:" service
-class SRV final : public Interface {
+class SRV final : public ServiceFramework<SRV> {
 public:
-    SRV();
-    ~SRV() override;
+    SRV(std::shared_ptr<ServiceManager> service_manager);
```
lioncash: Should this be an explicit constructor?

src/core/hle/service/service.h
yuriks: Done.

src/core/hle/service/service.h
yuriks: Done.

PR #2758, opened 2017-06-07
Disable warning for Cryptopp

BastienIstil:
Add CXX_FLAGS /W0 or -w, to suppress compiler warning in cryptopp component

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2758)

### DIFF
--- a/externals/cryptopp/CMakeLists.txt
+++ b/externals/cryptopp/CMakeLists.txt
@@ -44,6 +44,13 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Intel")
     add_definitions(-wd68 -wd186 -wd279 -wd327 -wd161 -wd3180)
 endif()
 
+#disable warning for external libraries
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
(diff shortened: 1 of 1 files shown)


yuriks:
Can we disable just the specific warnings that are being generated right now? On MSVC I only see one right now ("empty controlled statement") so I'd prefer that instead of just lowering to /W0.

BastienIstil:
Since it's a linker warning, I don't know how to pass /ignore:4221 to MSVC with CMAKE.
I try  
set_target_properties(cryptopp PROPERTIES LINK_FLAGS "/ignore:4221")
or 
SET( CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} /ignore:4221" )
but there is no change.
Do you have any idea ? 

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2758#-:-KmRgHjtntN8gYtqre1Z:bgxbutc)*

PR #2762, opened 2017-06-08
Fragment lighting: implement lut input 5 (CP) and tangent mapping

wwylele:
[Here](https://gist.github.com/wwylele/f61d6653620bb763ab9dc7a60a9831ce) is my notes when testing these features on 3DS.

[Here](https://github.com/wwylele/ctrhwtest/tree/master/lighting-cp) is a testing program with different bumping modes and lut inputs combined. The bump map this program uses consists of vectors of different directions, length and z-components. With this PR, the only difference between 3DS and citra is in the middle of the cube faces, where the bump map value is exactly (0.0, 0.0, 0.0), using normal map mode, CP input and "renorm" off. In this situation there is actually a divide-by-zero when projecting the vector, and game usually shouldn't have an zero normal vector, so I am not going to fix this.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2762)

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_gen.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_gen.cpp
@@ -534,18 +534,24 @@ static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
                 "(1.0 - (surface_normal.x*surface_normal.x + surface_normal.y*surface_normal.y))";
             out += "surface_normal.z = sqrt(max(" + val + ", 0.0));\n";
         }
+
+        // The tangent vector is not perturbed by the normal map and is just a unit vector.
+        out += "vec3 surface_tangent = vec3(1.0, 0.0, 0.0);\n";
     } else if (lighting.bump_mode == LightingRegs::LightingBumpMode::TangentMap) {
         // Bump mapping is enabled using a tangent map
         LOG_CRITICAL(HW_GPU, "unimplemented bump mapping mode (tangent mapping)");
         UNIMPLEMENTED();
     } else {
-        // No bump mapping - surface local normal is just a unit normal
+        // No bump mapping - surface local normal and tangent are just unit vectors
         out += "vec3 surface_normal = vec3(0.0, 0.0, 1.0);\n";
+        out += "vec3 surface_tangent = vec3(1.0, 0.0, 0.0);\n";
     }
 
     // Rotate the surface-local normal by the interpolated normal quaternion to convert it to
     // eyespace.
-    out += "vec3 normal = quaternion_rotate(normalize(normquat), surface_normal);\n";
+    out += "vec4 normalized_normquat = normalize(normquat);\n";
+    out += "vec3 normal = quaternion_rotate(normalized_normquat, surface_normal);\n";
+    out += "vec3 tangent = quaternion_rotate(normalized_normquat, surface_tangent);\n";
 
     // Gets the index into the specified lookup table for specular lighting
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -534,18 +534,23 @@ static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
...
-        // No bump mapping - surface local normal is just a unit normal
+        // No bump mapping - surface local normal and tangent are just unit vectors
         out += "vec3 surface_normal = vec3(0.0, 0.0, 1.0);\n";
+        out += "vec3 surface_tangent = vec3(1.0, 0.0, 0.0);\n";
     }
 
     // Rotate the surface-local normal by the interpolated normal quaternion to convert it to
     // eyespace.
     out += "vec3 normal = quaternion_rotate(normalize(normquat), surface_normal);\n";
+    out += "vec3 tangent = quaternion_rotate(normalize(normquat), surface_tangent);\n";
```
wwylele: The tangent vector is rarely used, so this computation is redundant in most cases. Can I rely on glsl optimization, or I should wrap it with `if` detecting if any LUT input is CP?

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -534,18 +534,24 @@ static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
...
         out += "vec3 surface_normal = vec3(0.0, 0.0, 1.0);\n";
+        out += "vec3 surface_tangent = vec3(1.0, 0.0, 0.0);\n";
     }
 
     // Rotate the surface-local normal by the interpolated normal quaternion to convert it to
     // eyespace.
-    out += "vec3 normal = quaternion_rotate(normalize(normquat), surface_normal);\n";
+    out += "vec4 normalized_normquat = normalize(normquat);\n";
+    out += "vec3 normal = quaternion_rotate(normalized_normquat, surface_normal);\n";
+    out += "vec3 tangent = quaternion_rotate(normalized_normquat, surface_tangent);\n";
```
wwylele: The tangent vector is rarely used, so this computation is redundant in most cases. Can I rely on glsl optimization, or I should wrap it with `if` detecting if any LUT input is CP?

yuriks:
:lgtm: Good job researching and implementing this stuff.

---

Reviewed 1 of 2 files at r4, 1 of 1 files at r5, 1 of 1 files at r7.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 554 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2762#-Km7vPWsXMMasBOyHJbh-r5-554:-Km_DJpayA0eAwvU9LFr:b-annyu0) ([raw file](https://github.com/citra-emu/citra/blob/cb063a70a7a6caf1e2d43e1b9f06ede3f870f5a9/src/video_core/renderer_opengl/gl_shader_gen.cpp#L554)):*

The tangent vector is rarely used, so this computation is redundant in most cases. Can I rely on glsl optimization, or I should wrap it with `if` detecting if any LUT input is CP?
</blockquote></details>

GLSL compilers are usually pretty good at eliminating dead code, I don't think we need to worry about it for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2762#-:-Km_HT3IMbXKf_zX5OF2:b-f9sqp8)*

PR #2764, opened 2017-06-09
Video Core: Add vertex loader cache

jroweboy:
## Description

Adds a vertex loader cache. The code is similar to the shader JIT cache.

## Rationale

Before adding a vertex loader JIT, we should have a cache to store the vertex loaders so we don't need to regenerate them all the time.

## Performance

In my testing, this has no affect on performance. If anything, I'd imagine it might be slower if the cache lookup is more than the penalty for generating new vertex loaders, but I tested all of my games and the frametime didn't change.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2764)

### DIFF
--- a/src/video_core/vertex_loader.cpp
+++ b/src/video_core/vertex_loader.cpp
@@ -16,10 +16,7 @@
 
 namespace Pica {
 
-void VertexLoader::Setup(const PipelineRegs& regs) {
-    ASSERT_MSG(!is_setup, "VertexLoader is not intended to be setup more than once.");
-
-    const auto& attribute_config = regs.vertex_attributes;
+VertexLoader::VertexLoader(const PipelineRegs::VertexAttributes& attribute_config) {
     num_total_attributes = attribute_config.GetNumTotalAttributes();
 
     boost::fill(vertex_attribute_sources, 0xdeadbeef);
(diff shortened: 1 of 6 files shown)


src/video_core/vertex_loader.cpp
```diff
@@ -16,10 +16,7 @@
 
 namespace Pica {
 
-void VertexLoader::Setup(const PipelineRegs& regs) {
```
jroweboy: I removed `Setup` in favor of just adding the setup code to the constructor. Should I instead leave it as a virtual function on the base class instead?

src/video_core/vertex_loader_base.h
```diff
@@ -0,0 +1,32 @@
...
+struct AttributeBuffer;
+}
+
+class VertexLoaderBase {
+public:
+    virtual void LoadVertex(u32 base_address, int index, int vertex, Shader::AttributeBuffer& input,
+                            DebugUtils::MemoryAccessTracker& memory_accesses) = 0;
+};
+
+using VertexLoaderCache = std::unordered_map<u64, std::unique_ptr<VertexLoaderBase>>;
```
InusualZ: nitpick: you are making an alias that you only use once?

jroweboy: Yes. I figured it would be nicer to be explicit about what the type was for, but I can remove it if thats what people want (esp considering its probably very obvious what its for).

jroweboy:
@yuriks suggested on IRC that I just close this pull request until the vertex loader JIT is complete, as the change is really useless without the JIT. I will either reopen this one later when I submit the second part as well, or combine it with the vertex loader JIT PR. (I opened this one now to make it easier to review just the cache without the rest of the code)

PR #2767, opened 2017-06-10
OpenGL: Update comment on AreQuaternionsOpposite with new information

yuriks:
While debugging the software renderer implementation, it was noticed
that this is actually exactly what the hardware does, upgrading the
status of this "hack" to being a proper implementation. And there was
much rejoicing.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2767)

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -182,19 +182,22 @@ RasterizerOpenGL::RasterizerOpenGL() : shader_dirty(true) {
 RasterizerOpenGL::~RasterizerOpenGL() {}
 
 /**
- * This is a helper function to resolve an issue with opposite quaternions being interpolated by
- * OpenGL. See below for a detailed description of this issue (yuriks):
+ * This is a helper function to resolve an issue when interpolating opposite quaternions. See below
+ * for a detailed description of this issue (yuriks):
  *
  * For any rotation, there are two quaternions Q, and -Q, that represent the same rotation. If you
  * interpolate two quaternions that are opposite, instead of going from one rotation to another
  * using the shortest path, you'll go around the longest path. You can test if two quaternions are
- * opposite by checking if Dot(Q1, W2) < 0. In that case, you can flip either of them, therefore
- * making Dot(-Q1, W2) positive.
+ * opposite by checking if Dot(Q1, Q2) < 0. In that case, you can flip either of them, therefore
+ * making Dot(Q1, -Q2) positive.
  *
- * NOTE: This solution corrects this issue per-vertex before passing the quaternions to OpenGL. This
- * should be correct for nearly all cases, however a more correct implementation (but less trivial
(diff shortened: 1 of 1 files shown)


yuriks:
For posteriority this is how this came about:
 1. While thinking about this in the context of the software renderer I thought of a possible way to optimize this into a sign check instead of a dot product: "If this implementation works for all sets of vertices, then I could take a quaternion such as (1,0,0,0) or (0,0,0,1) and use it as a reference against the other 3 quaternions. This reduces the dot product to a simple sign comparison."
 2. The optimization produced wrong results. Conclusion: (0,0,0,1) in v0 is in fact a counter-example that breaks the workaround and which can be used to check against hardware.
 3. Homebrew confirms that hardware works for the testcase (as does Citra with the workaround.)
 4. However, using a vertex other than v0 (such as v1 or v2) breaks in those test cases. This is equivalent to rotating the triangle so that that vertex is v0. Doing this in the homebrew program also breaks hardware for that case.
 5. Checked all 3 x 4 permutations of the triangle and of "axis-aligned" unit quaternions and confirmed Citra matches output for all of them. Implemented "proper solution" in software renderer and confirmed that test-cases are stable under triangle rotation as expected, but differ from hardware.

wwylele:
We probably want to move this quaternion correction to somewhere before the rasterizer (such as [here](https://github.com/citra-emu/citra/blob/master/src/video_core/command_processor.cpp#L262)?), since it is confirmed as the hardware accurate way and should be shared by all rasterizer implementation

yuriks:
@wwylele I considered whether we should do that, but decided against it because you can do it much more easily/with less code in the software renderer, as well as skipping these checks entirely if the triangle is clipped out after primitive assembly or if fragment lighting is disabled (and thus we don't need to use the normal).

wwylele:
>\<yuriks\> wwylele: also, there isn't a very natural place to put the code, we don't have much space between primitive assembly and the rasterization backends
\<yuriks\> moving the checks/comment to an agnostic helper function would probably be a good one, but I'll leave that to Subv or do it after his PR :P
\<wwylele\> sounds fine to me
\<wwylele\> Though, I don't see how we can do it with less code in sw...
\<yuriks\> you can just mutate the quaternion XD
\<yuriks\> (unlike that, where you just have const refs to the vertices)
\<yuriks\> and we want to avoid making a redundant copy of the vertex there because it's quite a hot function
\<yuriks\> so you end up with that akward "flip_quaternion" parameter
\<yuriks\> maybe I missed a good way to do it though. Certainly not going to protest if a good PR shows up :)
\<wwylele\> but this quaternion correction needs to happen before clipping
\<wwylele\> (because interpolation happens in clipping)
\<yuriks\> oh, that's an excellent point
\* yuriks decides to actually check that
\<yuriks\> sw renderer has some weird depth issues: http://i.imgur.com/ChDwzaA.png
\<yuriks\> (also, yes, check needs to happen before clipping)

I'll leave this to Subv to decide

PR #2772, opened 2017-06-10
[RFC] Separate parts of the sw renderer into their own functions

Subv:
This helps the readability of the sw renderer since it is no longer a monolithic function.

I want to hear opinions about this change, is it a good idea? is it completely misled?
Hopefully in the future, this will help optimizing the rasterizer into a series of precomputed function calls, doing away with all the redundant switches.

Currently only the tev stages, the alpha test and the fog combiner are factored out.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2772)

### DIFF
--- a/src/video_core/swrasterizer/rasterizer.cpp
+++ b/src/video_core/swrasterizer/rasterizer.cpp
@@ -114,6 +114,70 @@ static std::tuple<float24, float24, PAddr> ConvertCubeCoord(float24 u, float24 v
     return std::make_tuple(x / z * half + half, y / z * half + half, addr);
 }
 
+Math::Vec4<u8> GetNextCombinerBuffer(unsigned tev_stage_index, const Math::Vec4<u8>& combiner_output,
+    const Math::Vec4<u8>& previous_buffer) {
+    Math::Vec4<u8> next_combiner_buffer = previous_buffer;
+
+    if (g_state.regs.texturing.tev_combiner_buffer_input.TevStageUpdatesCombinerBufferColor(
+        tev_stage_index)) {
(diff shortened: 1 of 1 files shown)


src/video_core/swrasterizer/rasterizer.cpp
```diff
@@ -114,6 +114,146 @@ static std::tuple<float24, float24, PAddr> ConvertCubeCoord(float24 u, float24 v
     return std::make_tuple(x / z * half + half, y / z * half + half, addr);
 }
 
+Math::Vec4<u8> GetNextCombinerBuffer(unsigned tev_stage_index, const Math::Vec4<u8>& combiner_output,
```
lioncash: Most of these new functions should be static, no?

yuriks:
Move these to texturing.cpp (I chose a bad name for that one... should be combiners.cpp or something) where some of this code already is.

yuriks:
(And re. if this is a good idea or not: it totally is)

src/video_core/swrasterizer/rasterizer.cpp
```diff
@@ -410,6 +550,7 @@ static void ProcessTriangleInternal(const Vertex& v0, const Vertex& v1, const Ve
             for (unsigned tev_stage_index = 0; tev_stage_index < tev_stages.size();
                  ++tev_stage_index) {
                 const auto& tev_stage = tev_stages[tev_stage_index];
+
```
jroweboy: random space added?

Subv:
I'm sorry but I have lost motivation to work on this anymore, I'm closing it so if anyone wants to take it over, feel free.

PR #2773, opened 2017-06-11
[RFC] citra-qt: Rework HotKeys & Add new ones

Kloen:
## [RFC]

This is a preliminar version, most of it works as expected, but I'm aware that isn't on the perfect way we want to. So I want to hear your thoughts about it.

If you want a hotkey to be added to trigger X button/action, I'll add it too.

In addition, if someone with a OSX or linux could try this, it would be good, because I can't try on that OS at the moment, so I really don't know if it works as intended or not.

### Current Issues:
- [x] ~~You can set a Input Key as HotKey~~ (Ugly workaround, 1st reason of this being RFC)
- [ ] You can set a HotKey as Input Key
- [ ] Tab Key can't be binded manually.
- [ ] Hotkeys not working on Ubuntu... Not even on master?

### TODO:
- [ ] Show an error or something if the user tries to bind an already binded key.
- [ ] Allow to use Alt / Shift as hotkeys

### Not going to be in this PR (just yet):
- [ ] Reload Hotkeys on the fly.

#### Debug Tools
image

#### Emulation
image

#### Config Tab & HotKey list
image

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2773)


wwylele:
Did some test on Ubuntu:
 - It _silently_ refuses to bind an already bound key, which makes me think it just doesn't work at the first time
 - It only accept Ctrl as combination key, not for Shift and Alt, which kind of goes against usual experience.
 - :joy_cat: the hotkeys just don't work for me. They do show in the menu and buttons as hot key but they don't response when I press the key... Actually, it doesn't work on the master branch either, so I don't know where the actual problem is.

Will do code wise review the other day

B3n30:
What's the status of this PR. Was it just a RFC and it's done, or do you plan to actually PR it?
@wwylele You promised to review this on 12 Jun...

jroweboy:
Sorry I haven't chimed in about this yet. Been really busy and I kept putting reviewing this one off :/ I've only skimmed the PR, so I'm not going to comment on the code yet, but I like the idea of having all these bindings be customizable. If you'd like to bring this one back to life, let me know and I'll try to make time to get it reviewed.

thenobletheif:
Could you add a hotkey for taking screenshots (if you haven't already)? Maybe with a window pop up that saves the screenshot to a location? sometimes I just want to capture what i'm seeing with a single button press.

B3n30:
Closed, due to inactivity.

PR #2774, opened 2017-06-11
Add basic support for IPC translation for HLE services

yuriks:
Only on ServiceFramework™-based services (aka, just sm/`srv:` for now.) Changes include:
  - Use a separate command buffer copy instead of working directly with the caller's.
  - Make handle lookup/creation the responsibility of the Kernel/Framework

Commits are probably reviewable individually.

Depends on citra-emu/ext-boost#11 for `small_vector`.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2774)

### DIFF
--- a/src/core/hle/ipc.h
+++ b/src/core/hle/ipc.h
@@ -44,6 +44,9 @@ inline u32* GetStaticBuffers(const int offset = 0) {
 
 namespace IPC {
 
+/// Size of the command buffer area, in 32-bit words.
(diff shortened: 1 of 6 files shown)


yuriks:
[15:56:17] <Subv> well yuriks, merge it so i can start blindly converting stuff and breaking everything in the meantime

wwylele:
Retroactive LGTM, but I am a little bit worried that in this fast pace we might make hidden bugs that are hard to find later. Can we have a dummy testing service, together with a service testing program, to make sure everything works fine?

Issue #2477, opened 2017-01-27
Lighting issues with ACNL, Smash Bros. 4, and ACHHD


KillzXGaming:
This has been around since fragment lighting was first implemented, where you get shiny or over bright lighting on certain models or characters. I have no idea if this is something unimplemented so please correct me if this is the case on what causes this. I've tried searching issues and couldn't find any related issues on this , no idea if this is related to LUT lighting or anything or even just some mip map or rim lighting thing. These are currently the only titles i have so i do not currently know which other titles are affected by this. 

Here are Citra screenshots:
![Alt Text](http://i.imgur.com/3yxSWnY.png)

![Alt Text](http://i.imgur.com/GR4q8n7.png )

 ![Alt Text](http://i.imgur.com/Mbh5C6J.png )

In Game
 ![Alt Text](http://i.imgur.com/Uyk3o4Z.png)

 ![Alt Text](http://i.imgur.com/s8RYmjw.png)  

![Alt Text](http://i.imgur.com/0QqHmZp.png )

Screenshots taken from an older master build from a few weeks ago but i can assure you i tested again just now on this one here:
https://github.com/citra-emu/citra/commit/bf14f4be2263b4769e97800b35951717192c2d1c

PC Specs:
i5 4670k
gtx 970
8gb RAM
Win 10

khang06:
Also happens on some Sun/Moon NPCs.

wwylele:
For Smash Bros 4, it is likely caused by unimplemented geometric factor (see [here](http://mathb.in/64694) for G0 and G1). When rendering characters, the game have the geometric factor register bit enabled. I tried plugging in some geometric factor formula got from google, and the characters do look much more better

![screenshot from 2017-06-07 23-03-01](https://user-images.githubusercontent.com/4592895/26898852-f51e55be-4bd5-11e7-9b92-d9472650b1ed.png)
![screenshot from 2017-06-07 22-58-09](https://user-images.githubusercontent.com/4592895/26898857-f8c8b5c4-4bd5-11e7-95ed-f4fdfd4cc7b1.png)

KillzXGaming:
All games now have proper lighting!
![2](https://user-images.githubusercontent.com/13475262/27265706-562a4518-5467-11e7-86e6-c16896d911de.jpg)

![1](https://user-images.githubusercontent.com/13475262/27265713-6ab49222-5467-11e7-9648-f192da6ccf0a.png)

PR #2776, opened 2017-06-13
Fragment lighting: implement geometric factor

wwylele:
Reference formula on 3dbrew: http://mathb.in/64694. G0 and G1 are geometric factors

The test result shows that the formula for geometric factors in 3DS is not [the standard Cook-Torrance one](http://ruh.li/images/CookTorranceGeometricAttenuation.png). To get the correct formula, [a testing program](https://github.com/wwylele/ctrhwtest/tree/master/geo_factor) is used. Some assumptions are made when finding the formula:
 - the formula on 3dbrew is correct
 - G only depends on normalized vector N (normal), V (view) and L (light), and can contain half-angle vector H (still depends on V and L)
 - spherical symmetry, as in, rotating N, V and L by the same angle won't change G value 

Based on these assumptions, G value is sampled for different combinations of N, V and L (with two-sided-diffuse on). Because of the symmetry, V and L is confined on the xy plane and are symmetrical about x-axis (so that the potentially used vector H is always (+/-1, 0, 0)). A result of ~65600 samples is [here](https://www.dropbox.com/s/j7cycr7hn9wpuch/geo_factor_big.txt?dl=1). 

The nearest formula found is `G = min(1.0, 0.5 * abs(L * N) / (1 + L * V))`. For the case `1 + L * V == 0` G is set to 0. The overall error of this formula is +/- 0.01, which is negligible considering the measurement is taken by dividing two u8 integer. It is also shown in later tests that G does respect to the two-sided-diffuse bit, which means `abs(L * N)` part is actually `dot_product`.

Fixes the lighting issue of Smash Bros. Probably also fixes other games in #2477 but not tested yet.

Side note: after this PR and the tangent map one, the only known unimplemented feature in lighting is the shadow mapping.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2776)

### DIFF
--- a/src/video_core/regs_lighting.h
+++ b/src/video_core/regs_lighting.h
@@ -168,6 +168,8 @@ struct LightingRegs {
         union {
             BitField<0, 1, u32> directional;
             BitField<1, 1, u32> two_sided_diffuse; // When disabled, clamp dot-product to 0
+            BitField<2, 1, u32> geometric_factor_0;
+            BitField<3, 1, u32> geometric_factor_1;
         } config;
 
         BitField<0, 20, u32> dist_atten_bias;
--- a/src/video_core/renderer_opengl/gl_shader_gen.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_gen.cpp
@@ -73,6 +73,8 @@ PicaShaderConfig PicaShaderConfig::BuildFromRegs(const Pica::Regs& regs) {
         state.lighting.light[light_index].num = num;
         state.lighting.light[light_index].directional = light.config.directional != 0;
         state.lighting.light[light_index].two_sided_diffuse = light.config.two_sided_diffuse != 0;
+        state.lighting.light[light_index].geometric_factor_0 = light.config.geometric_factor_0 != 0;
+        state.lighting.light[light_index].geometric_factor_1 = light.config.geometric_factor_1 != 0;
         state.lighting.light[light_index].dist_atten_enable =
             !regs.lighting.IsDistAttenDisabled(num);
         state.lighting.light[light_index].spot_atten_enable =
@@ -518,7 +520,8 @@ static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
            "vec4 specular_sum = vec4(0.0, 0.0, 0.0, 1.0);\n"
            "vec3 light_vector = vec3(0.0);\n"
            "vec3 refl_value = vec3(0.0);\n"
-           "vec3 spot_dir = vec3(0.0);\n;";
+           "vec3 spot_dir = vec3(0.0);\n"
+           "float geo_factor = 1.0;\n";
 
     // Compute fragment normals and tangents
     const std::string pertubation =
@@ -671,6 +674,12 @@ static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
         std::string clamp_highlights =
             lighting.clamp_highlights ? "(dot(light_vector, normal) <= 0.0 ? 0.0 : 1.0)" : "1.0";
 
+        if (light_config.geometric_factor_0 || light_config.geometric_factor_1) {
+            out += "geo_factor = 1 + dot(light_vector, normalize(view));\n"
+                   "geo_factor = geo_factor == 0.0 ? 0.0 : min(0.5 * " +
+                   dot_product + " / geo_factor, 1.0);\n";
+        }
+
         // Specular 0 component
         std::string d0_lut_value = "1.0";
         if (lighting.lut_d0.enable &&
@@ -683,6 +692,9 @@ static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
                            GetLutValue(LightingRegs::LightingSampler::Distribution0, index) + ")";
         }
         std::string specular_0 = "(" + d0_lut_value + " * " + light_src + ".specular_0)";
+        if (light_config.geometric_factor_0) {
+            specular_0 = "(" + specular_0 + " * geo_factor)";
+        }
 
         // If enabled, lookup ReflectRed value, otherwise, 1.0 is used
         if (lighting.lut_rr.enable &&
@@ -738,6 +750,9 @@ static void WriteLighting(std::string& out, const PicaShaderConfig& config) {
         }
         std::string specular_1 =
             "(" + d1_lut_value + " * refl_value * " + light_src + ".specular_1)";
+        if (light_config.geometric_factor_1) {
+            specular_1 = "(" + specular_1 + " * geo_factor)";
+        }
 
         // Fresnel
         if (lighting.lut_fr.enable &&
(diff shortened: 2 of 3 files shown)


yuriks:
The paper "Primitive Processing and Advanced Shading Architecture for Embedded Space" (both actual paper as well as [presentation slides](http://highperformancegraphics.org/previous/www_2011/media/Papers/HPG2011_Papers_Kazakov.pdf)) have the formula `(L * N) / |L + V|^2` (aka `dot(L, N) / dot(H, H)`) for their approximation. This produces exactly the same results as the formula you found, but we should probably use that since it's easier to find a source for and slightly cheaper to compute.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2776#-:-Kmt96fKiGnfYtnY1kpP:b77mxzz)*

wwylele:
Changed to what @yuriks suggested, and slightly refactored out the half-angle vector computation.

yuriks:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 524 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2776#-Kmu3oBVNJ7bg8DdtDkK:-Kmu3oBVNJ7bg8DdtDkL:b-eku8ri) ([raw file](https://github.com/citra-emu/citra/blob/0136b9a1e72a4086dc7345941d38799f9360d96b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L524)):*
> ```cpp
>            "vec3 refl_value = vec3(0.0);\n"
>            "vec3 spot_dir = vec3(0.0);\n"
>            "vec3 half_angle = vec3(0.0);\n"
> ```

`half_vector` is a more common name for it.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 568 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2776#-Kmu3VGcCkXrd97c2Kw3:-Kmu3VGcCkXrd97c2Kw4:bsghwr3) ([raw file](https://github.com/citra-emu/citra/blob/0136b9a1e72a4086dc7345941d38799f9360d96b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L568)):*
> ```cpp
>     auto GetLutIndex = [&lighting](unsigned light_num, LightingRegs::LightingLutInput input,
>                                    bool abs) {
>         const std::string half_angle = "normalize(half_angle)";
> ```

For `view` we repeat the `normalize(view)` on every usage below. I think it would make sense to do the the same for `half_angle` which would turn all the strings below into constant strings with no concatenation (easier to read).

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 645 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2776#-Kmu3vV2gzsYh-qHYNRZ:-Kmu3vV3VVMJJhFXpSaG:box5m8b) ([raw file](https://github.com/citra-emu/citra/blob/0136b9a1e72a4086dc7345941d38799f9360d96b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L645)):*
> ```cpp
> 
>         out += "spot_dir = " + light_src + ".spot_direction;\n";
>         out += "half_angle = normalize(view) + light_vector;\n";
> ```

You should declare the variable here instead of at the top so we don't accidentally use it before it has a value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2776#-:-Kmu43Z1eX_tyFsULFMj:bnfp4nl)*

wwylele:

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 524 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2776#-Kmu3oBVNJ7bg8DdtDkK:-Kmu5jsfBNb47ysbObUo:bb7f2mf) ([raw file](https://github.com/citra-emu/citra/blob/0136b9a1e72a4086dc7345941d38799f9360d96b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L524)):*

`half_vector` is a more common name for it.
</blockquote></details>

done

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 568 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2776#-Kmu3VGcCkXrd97c2Kw3:-Kmu5lbVob8DgkPhRON1:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/0136b9a1e72a4086dc7345941d38799f9360d96b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L568)):*

For `view` we repeat the `normalize(view)` on every usage below. I think it would make sense to do the the same for `half_angle` which would turn all the strings below into constant strings with no concatenation (easier to read).
</blockquote></details>

Done.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 645 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2776#-Kmu3vV2gzsYh-qHYNRZ:-Kmu5m9UBlRivCksUnPQ:bdrqnw2) ([raw file](https://github.com/citra-emu/citra/blob/0136b9a1e72a4086dc7345941d38799f9360d96b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L645)):*

You should declare the variable here instead of at the top so we don't accidentally use it before it has a value.
</blockquote></details>

We declare all these variable (`light_vector`, `spot_dir`, etc) above because these lines are in a c++ loop but not in a GLSL loop. If they get declared here, they will be declared multiple times. Is this allowed in GLSL? or does GLSL support plain scope `{}` so we can move them into individual scope?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2776)*

yuriks:
Reviewed 1 of 1 files at r3.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 645 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2776#-Kmu3vV2gzsYh-qHYNRZ:-Kmu7wdGmhZjYVCsc2TB:b-xprzu) ([raw file](https://github.com/citra-emu/citra/blob/0136b9a1e72a4086dc7345941d38799f9360d96b/src/video_core/renderer_opengl/gl_shader_gen.cpp#L645)):*

We declare all these variable (`light_vector`, `spot_dir`, etc) above because these lines are in a c++ loop but not in a GLSL loop. If they get declared here, they will be declared multiple times. Is this allowed in GLSL? or does GLSL support plain scope `{}` so we can move them into individual scope?
</blockquote></details>

Ah, I didn't realize that. From a quick test in shadertoy.com it seems that at least my drivers are fine with a stand-alone scope, but I don't know if that'll work with all vendors. I think it's fine to leave this as is then.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2776)*

PR #2784, opened 2017-06-15
load shared font from system archive

wwylele:
(This is me having OCD...)

**Just in case of misunderstanding, this PR doesn't eliminate the need of shared font. It only changes the loading format of the shared font**

This mimics the way how the real APT service loads the shared font archive. The result is verified against the data from real hardware, the data from citra running NS sysmodule, and shared_font.bin itself. With this we can get rid of the weird shared_font.bin file and have all system data uniformly stored in system archives. A fallback to the old shared_font.bin is provided for compatibility.

To use this, 3dsutils needs to be modified to dump the font archive. Add this to the source code:
```
    // savedatacheck/0004009b00014002.bin
    u64 binary_lowpath_0004009b00014002[] = { 0x0004009b00014002, 0x00000001ffffff00 };
    DumpSharedRomFS((u32*)binary_lowpath_0004009b00014002);
```
If this PR is accepted, I will also push the 3dsutils change.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2784)

### DIFF
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -6,11 +6,13 @@
 #include "common/file_util.h"
 #include "common/logging/log.h"
 #include "core/core.h"
+#include "core/file_sys/file_backend.h"
 #include "core/hle/applets/applet.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/kernel/mutex.h"
 #include "core/hle/kernel/process.h"
 #include "core/hle/kernel/shared_memory.h"
+#include "core/hle/romfs.h"
 #include "core/hle/service/apt/apt.h"
 #include "core/hle/service/apt/apt_a.h"
 #include "core/hle/service/apt/apt_s.h"
@@ -27,6 +29,7 @@ namespace APT {
 
 /// Handle to shared memory region designated to for shared system font
 static Kernel::SharedPtr<Kernel::SharedMemory> shared_font_mem;
+static bool shared_font_loaded = false;
 static bool shared_font_relocated = false;
 
 static Kernel::SharedPtr<Kernel::Mutex> lock;
@@ -71,7 +74,7 @@ void Initialize(Service::Interface* self) {
 void GetSharedFont(Service::Interface* self) {
     IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x44, 0, 0); // 0x00440000
     IPC::RequestBuilder rb = rp.MakeBuilder(2, 2);
-    if (!shared_font_mem) {
+    if (!shared_font_loaded) {
         LOG_ERROR(Service_APT, "shared font file missing - go dump it from your 3ds");
         rb.Push<u32>(-1); // TODO: Find the right error code
         rb.Skip(1 + 2, true);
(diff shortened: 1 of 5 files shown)


src/common/string_util.h
```diff
@@ -74,7 +75,17 @@ bool AsciiToHex(const char* _szValue, u32& result);
 
 std::string TabsToSpaces(int tab_size, const std::string& in);
 
-void SplitString(const std::string& str, char delim, std::vector<std::string>& output);
+template <typename CharT>
+void SplitString(const std::basic_string<CharT>& str, CharT delim,
+                 std::vector<std::basic_string<CharT>>& output) {
+    std::basic_istringstream<CharT> iss(str);
```
wwylele: It seems that `std::basic_istringstream` only supports `char` and `wchar_t` in the standard. I need to find a work around or reimplement this function

MerryMage: Was bored:

    template <typename CharT>
    std::vector<std::basic_string<CharT>> SplitString(const std::basic_string<CharT>& str, CharT delim) {
        std::vector<std::basic_string<CharT>> output;
        auto curr = str.cbegin();
        auto last = str.cend();
        while (curr != last) {
            const auto pos = std::find(curr, last, delim);
            output.emplace_back(curr, pos);
            if (pos == last)
                break;
            curr = ++pos;
        }
        return output;
    }

wwylele: thanks. However I just decided not to use this function because it is not necessary for share font archive implementation. I will leave this to others who wants to use it :stuck_out_tongue:

src/core/hle/service/apt/apt.cpp
```diff
@@ -644,36 +646,236 @@ void CheckNew3DS(Service::Interface* self) {
...
+        u64_le data_length;
+        u32_le same_hash_next_file_offset;
+        u32_le name_length;
+        // followed by file name
+    } file;
+    static_assert(sizeof(file) == 0x20, "file has incorrect size");
+
+    // Find the file
+    u32 file_offset = dir.first_file_offset;
+    while (1) {
```
jroweboy:
```cpp
while (file_offset != INVALID_FIELD) {
    ...
}
return nullptr;
```

src/core/hle/service/apt/apt.cpp
```diff
@@ -644,36 +646,234 @@ void CheckNew3DS(Service::Interface* self) {
...
+    } dir;
+    static_assert(sizeof(dir) == 0x18, "dir has incorrect size");
+
+    // Find directories of each level
+    const u8* current_dir = romfs + header.dir_table_offset;
+    std::memcpy(&dir, current_dir, sizeof(dir));
+    for (const std::u16string& dir_name : dir_names) {
+        u32 child_dir_offset;
+        child_dir_offset = dir.first_child_dir_offset;
+        while (1) {
```
lioncash: `while (true)`

wwylele:
As no one expresses their opinion on the change, I am assuming no one is against it? Taking out of RFC then

B3n30:
Reviewed 1 of 1 files at r1, 2 of 3 files at r3, 1 of 1 files at r5.

---

*[src/core/hle/service/apt/apt.cpp, line 846 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2784#-KnU-MLB8CDEB1EtEFsw:-KnU-MLB8CDEB1EtEFsx:b-fzla23) ([raw file](https://github.com/citra-emu/citra/blob/4baac0c305462d828eebc878d76c5be349fce823/src/core/hle/service/apt/apt.cpp#L846)):*
> ```cpp
>     // "shared_font.bin" in the Citra "sysdata" directory.
> 
>     LOG_WARNING(Service_APT, "LoadSharedFont failed. Trying legacy method");
> ```

I find the location of this log entry strange. It only is correct if LoadLegacySharedFont is called in the way you programmed it with the if clause below. Although the way you've made the if clause is very nice, I think it would be better to do:
`if (LoadSharedFont()) {`
`    shared_font_loaded = true;`
`} else if (LoadLegacySharedFont) {`
`    LOG_WARNING(Service_APT, "LoadSharedFont failed. Trying legacy method");`
`    shared_font_loaded = true;`
`} else {`
`    LOG_WARNING(Service_APT, "Unable to load shared font");`
`    shared_font_loaded = false;`
`}`

besides that: LGTM

---

*[src/core/hle/service/apt/apt.cpp, line 872 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2784#-KnU0wzHs4CPRl2j3hh1:-KnU0wzIu09Mi1N3UDbg:b-33b6k6) ([raw file](https://github.com/citra-emu/citra/blob/4baac0c305462d828eebc878d76c5be349fce823/src/core/hle/service/apt/apt.cpp#L872)):*
> ```cpp
>                                      Kernel::MemoryRegion::SYSTEM, "APT:SharedFont");
> 
>     if (!LoadSharedFont() && !LoadLegacySharedFont()) {
> ```

See comment above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2784)*

wwylele:
Review status: 1 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/core/hle/service/apt/apt.cpp, line 846 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2784#-KnU-MLB8CDEB1EtEFsw:-KnUqThpJriFL7hHZBFe:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/4baac0c305462d828eebc878d76c5be349fce823/src/core/hle/service/apt/apt.cpp#L846)):*

I find the location of this log entry strange. It only is correct if LoadLegacySharedFont is called in the way you programmed it with the if clause below. Although the way you've made the if clause is very nice, I think it would be better to do:
`if (LoadSharedFont()) {`
`    shared_font_loaded = true;`
`} else if (LoadLegacySharedFont) {`
`    LOG_WARNING(Service_APT, "LoadSharedFont failed. Trying legacy method");`
`    shared_font_loaded = true;`
`} else {`
`    LOG_WARNING(Service_APT, "Unable to load shared font");`
`    shared_font_loaded = false;`
`}`

besides that: LGTM
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2784)*

B3n30:
:lgtm:

---

Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2784#-:-KnUqzDxvKDBJ3Uy9NFk:bnfp4nl)*

wwylele:
Moved RomFS stuff to its own file as @yuriks suggested

wwylele:
Last call for reviews... if there is no issue I will merge both this and the one for 3dsutils(https://github.com/citra-emu/3dsutils/pull/14), and update the wiki guide

PR #2787, opened 2017-06-18
Kernel/IPC: Add tests for HLERequestContext buffer translation

yuriks:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2787)

### DIFF
--- a/src/tests/core/hle/kernel/hle_ipc.cpp
+++ b/src/tests/core/hle/kernel/hle_ipc.cpp
@@ -1,0 +1,193 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <catch.hpp>
+#include "core/hle/ipc.h"
+#include "core/hle/kernel/client_port.h"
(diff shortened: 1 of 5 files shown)


yuriks:
@wwylele Ah, indeed, I think that's true and I did forget to handle that case. Will write a test on the 3DS to confirm and then update the implementation.

src/tests/core/hle/kernel/hle_ipc.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "core/hle/kernel/process.h"
+#include "core/hle/kernel/server_session.h"
+
+namespace Kernel {
+
+static SharedPtr<Object> MakeObject() {
+    return Event::Create(ResetType::OneShot);
+}
+
+TEST_CASE("HLERequestContext::PopoulateFromIncomingCommandBuffer", "[core][kernel]") {
```
baka0815: Typo: Popoulate should be Populate I guess?

yuriks: Indeed. Will fix on my next PR, thanks.

Issue #2551, opened 2017-02-09
Pokémon Main Series (X/Y, OR/AS, S/M) – Lighting/Shadowing Issue


Dragios:
I just discovered this regression is due to **Lighting LUT Quickfix #2367**.
Exact commit will be 88f409aec9a490138012cde35789d1f21f20818f

**Before** (commit 14f8065843bf865cbd0f2f34eabf54c37e52df81) / **After** (commit 2f746e9946f78a2e283dfdcbeda9cf332e44d099) / **Console**

**[Pokémon X/Y]**
![Pokémon X](https://cloud.githubusercontent.com/assets/18496868/25071192/7ee350a4-22e3-11e7-9051-6616b7ce6a5c.png)
Source: [Before](https://cloud.githubusercontent.com/assets/18496868/25092820/ef8cc680-23c1-11e7-835e-d4ee8d9abd8c.png) | [After](https://cloud.githubusercontent.com/assets/18496868/25092822/f12a442c-23c1-11e7-9fc9-950f71d833ce.png) | [Console](https://cloud.githubusercontent.com/assets/18496868/25092824/f254f7ca-23c1-11e7-9d8d-a03b4ffdbdc0.png) (Darker background unrelated)
<blockquote class="imgur-embed-pub" lang="en" data-id="a/cohbg"><a href="//imgur.com/a/cohbg">[Gallery] New 3DS XL – Pokémon X</a></blockquote>

**[Pokémon Alpha Saphire]**
![Pokémon Alpha Saphire](https://cloud.githubusercontent.com/assets/18496868/25071054/90528600-22e0-11e7-8969-6e4517a5d1f3.png)
Source: [Before](https://cloud.githubusercontent.com/assets/18496868/25092409/1ea4041c-23c0-11e7-90b8-e70b8ec12a16.png) | [After](https://cloud.githubusercontent.com/assets/18496868/25092411/206a621e-23c0-11e7-818a-b19531583f9d.png) | [Console](https://cloud.githubusercontent.com/assets/18496868/25092414/22e5a15c-23c0-11e7-96ae-8c845e2aa3c6.png) (Bubble effect unrelated)
`//Gallery of images to be added later`

**[Pokémon Sun & Moon]**
![Pokémon Sun](https://cloud.githubusercontent.com/assets/18496868/25071568/73fe5dc8-22ed-11e7-8fbc-4496d156e670.png)
[...]

PR #2792, opened 2017-06-20
gl_rasterizer: fix lighting LUT interpolation

wwylele:
## gl_rasterizer/lighting: fix LUT interpolation

Currently on the GL renderer, we use the texture linear filter for LUT interpolation. This approach have the following problems:
 - if the delta entry the game assigns is not the difference between two entries, linear filter will produce different output from real 3DS. (less likely)
 - hard to handle sampling at the edge due to different coordinates convention between OpenGL and PICA (see #2367 and #2551). Note that the [-1.0, 1.0] sampling mode in PICA, which works in a two's complement way, is the root of most lighting issue, because GL will interpolate between LUT(-1.0) and LUT(1.0) and generate wrong result at the adge.

Solution: just do what PICA do. Both LUT value and delta are uploaded to the shader and proper interpolation is calculated there. Since the filter is not used any more, texture can be also replaced 
by cheaper mechanism. Two alternative, uniform buffer and texture buffer, are considered.

---

### texture buffer
(this was originally suggested by @degasus)
Buffer texture is still sampled as texture units, but uploaded as a buffer. Because we are going to upload delta as well, the current "packing 4 LUT into one texture" method must change. Here are some options:
 - break them into 24 textures. But this is not feasible because the total texture unit limit is only guaranteed at least 16.
 - upload them into one texture buffer. The total size off all LUT is `24*256*8 = 49152` (assuming an entry is stored as two float), while the texture buffer size limit is at least 65536, meaning that it can hold all LUT data. Partially uploading the texture buffer is much cheaper than normal texture, so this is feasible
 - keep 4 LUT in one texture and change the structure of packing. This should still work, but the number  4 looks pretty arbitrary now. IMO it reduces the code readability. Also, we are already close to the 16 textures limit (15 currently), so we should take this opportunity to reduce it.

### uniform buffer
Recall that the total size off all LUT 49152, while the size limit of a uniform buffer is 16384, so one uniform buffer can't hold it. The number limits of uniform buffer is 12, meaning using one buffer for each LUT is not possible either. So a similar "packing" is needed, which has the same readability issue as packed textures.

---

There is another different between the two alternatives: texture buffer prefers random memory access, while uniform buffer prefers coherent memory access (source [here](http://rastergrid.com/blog/2010/01/uniform-buffers-vs-texture-buffers/)). For the case of LUT, the pattern is random access, so texture buffer is better.

Considering these, I chose to use one texture buffer to store all LUT. The texture buffer is used as a `vec2[256 * 24]` array, where the two components of the vector are for entry value and delta, and all LUT are concatenated to each other.

I didn't touch proctex and fog LUT. They are already properly interpolated using delta, but still using 1D textures. To avoid making a giant PR, I decided to leave them for the next one.

---

### gl_rasterizer: manage texture ids in one place
Originally proposed in proctex PR. Though I didn't use `enum`, but constants of a helper class instead. In this way, it can be easily used for both texture unit enum and uniform value.

### gl_rasterizer: fix glGetUniformLocation type 
:expressionless:

### gl_state: reset 1D textures
This fixes the log spam when switching from hw renderer to sw renderer. Originally the reset function only resets the 2D texture units, but not for 1D texture units.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2792)

### DIFF
--- a/src/video_core/renderer_opengl/gl_state.h
+++ b/src/video_core/renderer_opengl/gl_state.h
@@ -64,8 +64,8 @@ public:
     } texture_units[3];
 
     struct {
-        GLuint texture_1d; // GL_TEXTURE_BINDING_1D
-    } lighting_luts[6];
+        GLuint texture_buffer; // GL_TEXTURE_BINDING_BUFFER
+    } lighting_lut;
 
     struct {
         GLuint texture_1d; // GL_TEXTURE_BINDING_1D
--- a/src/video_core/renderer_opengl/gl_state.h
+++ b/src/video_core/renderer_opengl/gl_state.h
@@ -6,6 +6,29 @@
 
 #include <glad/glad.h>
 
+namespace TextureUnits {
+
+struct TextureUnit {
+    GLint id;
+    constexpr GLenum Enum() const {
+        return static_cast<GLenum>(GL_TEXTURE0 + id);
+    }
+};
+
+constexpr TextureUnit PicaTexture(int unit) {
+    return TextureUnit{unit};
+}
+
+constexpr TextureUnit LightingLUT{3};
+constexpr TextureUnit FogLUT{4};
+constexpr TextureUnit ProcTexNoiseLUT{5};
+constexpr TextureUnit ProcTexColorMap{6};
+constexpr TextureUnit ProcTexAlphaMap{7};
+constexpr TextureUnit ProcTexLUT{8};
+constexpr TextureUnit ProcTexDiffLUT{9};
+
+} // namespace TextureUnits
+
 class OpenGLState {
 public:
     struct {
--- a/src/video_core/pica_state.h
+++ b/src/video_core/pica_state.h
@@ -87,11 +87,17 @@ struct State {
             // LUT value, encoded as 12-bit fixed point, with 12 fraction bits
             BitField<0, 12, u32> value; // 0.0.12 fixed point
 
-            // Used by HW for efficient interpolation, Citra does not use these
-            BitField<12, 12, s32> difference; // 1.0.11 fixed point
+            // Used for efficient interpolation.
+            BitField<12, 11, u32> difference; // 0.0.11 fixed point
+            BitField<23, 1, u32> neg_difference;
 
-            float ToFloat() {
+            float ToFloat() const {
                 return static_cast<float>(value) / 4095.f;
+            }
+
+            float DiffToFloat() const {
+                float diff = static_cast<float>(difference) / 2047.f;
+                return neg_difference ? -diff : diff;
             }
         };
 
--- a/src/video_core/regs_lighting.h
+++ b/src/video_core/regs_lighting.h
@@ -26,6 +26,8 @@ struct LightingRegs {
         DistanceAttenuation = 16,
     };
 
+    static constexpr unsigned NumLightingSampler = 24;
+
     static LightingSampler SpotlightAttenuationSampler(unsigned index) {
         return static_cast<LightingSampler>(
             static_cast<unsigned>(LightingSampler::SpotlightAttenuation) + index);
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -1183,7 +1183,7 @@ void RasterizerOpenGL::SetShader() {
         state.Apply();
 
         // Set the texture samplers to correspond to different texture units
-        GLuint uniform_tex = glGetUniformLocation(shader->shader.handle, "tex[0]");
+        GLint uniform_tex = glGetUniformLocation(shader->shader.handle, "tex[0]");
         if (uniform_tex != -1) {
             glUniform1i(uniform_tex, TextureUnits::PicaTexture(0).id);
         }
@@ -1197,40 +1197,40 @@ void RasterizerOpenGL::SetShader() {
         }
 
         // Set the texture samplers to correspond to different lookup table texture units
-        GLuint uniform_lut = glGetUniformLocation(shader->shader.handle, "lighting_lut");
+        GLint uniform_lut = glGetUniformLocation(shader->shader.handle, "lighting_lut");
         if (uniform_lut != -1) {
             glUniform1i(uniform_lut, TextureUnits::LightingLUT.id);
         }
 
-        GLuint uniform_fog_lut = glGetUniformLocation(shader->shader.handle, "fog_lut");
+        GLint uniform_fog_lut = glGetUniformLocation(shader->shader.handle, "fog_lut");
         if (uniform_fog_lut != -1) {
             glUniform1i(uniform_fog_lut, TextureUnits::FogLUT.id);
         }
 
-        GLuint uniform_proctex_noise_lut =
+        GLint uniform_proctex_noise_lut =
             glGetUniformLocation(shader->shader.handle, "proctex_noise_lut");
         if (uniform_proctex_noise_lut != -1) {
             glUniform1i(uniform_proctex_noise_lut, TextureUnits::ProcTexNoiseLUT.id);
         }
 
-        GLuint uniform_proctex_color_map =
+        GLint uniform_proctex_color_map =
             glGetUniformLocation(shader->shader.handle, "proctex_color_map");
         if (uniform_proctex_color_map != -1) {
             glUniform1i(uniform_proctex_color_map, TextureUnits::ProcTexColorMap.id);
         }
 
-        GLuint uniform_proctex_alpha_map =
+        GLint uniform_proctex_alpha_map =
             glGetUniformLocation(shader->shader.handle, "proctex_alpha_map");
         if (uniform_proctex_alpha_map != -1) {
             glUniform1i(uniform_proctex_alpha_map, TextureUnits::ProcTexAlphaMap.id);
         }
 
-        GLuint uniform_proctex_lut = glGetUniformLocation(shader->shader.handle, "proctex_lut");
+        GLint uniform_proctex_lut = glGetUniformLocation(shader->shader.handle, "proctex_lut");
         if (uniform_proctex_lut != -1) {
             glUniform1i(uniform_proctex_lut, TextureUnits::ProcTexLUT.id);
         }
 
-        GLuint uniform_proctex_diff_lut =
+        GLint uniform_proctex_diff_lut =
             glGetUniformLocation(shader->shader.handle, "proctex_diff_lut");
         if (uniform_proctex_diff_lut != -1) {
             glUniform1i(uniform_proctex_diff_lut, TextureUnits::ProcTexDiffLUT.id);
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -49,9 +49,7 @@ RasterizerOpenGL::RasterizerOpenGL() : shader_dirty(true) {
 
     uniform_block_data.dirty = true;
 
-    for (unsigned index = 0; index < lighting_luts.size(); index++) {
-        uniform_block_data.lut_dirty[index] = true;
-    }
+    uniform_block_data.lut_dirty.fill(true);
 
     uniform_block_data.fog_lut_dirty = true;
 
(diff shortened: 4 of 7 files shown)


degasus:
The limits you've shown are indeed correct, but every GPU has bigger limits. There is no GPU with less than 64k bytes for a UBO block. So all LUTs will fit within one UBO. And this is not a limitation for the buffer object, they can be bigger, so you could store many combination of LUTs within one buffer object.

The texure buffer objects are indeed guaranteed to be at least 64k bytes, but they may usually be as big as the device memory of the GPU.

wwylele:
@degasus yes, I am aware that the actual limit is usually larger than required by the standard. I just want to play safe here :p .

src/video_core/renderer_opengl/gl_state.h
```diff
@@ -6,6 +6,33 @@
...
+    return TextureUnit{unit};
+}
+
+constexpr TextureUnit lighting_lut{3};
+constexpr TextureUnit fog_lut{4};
+constexpr TextureUnit proctex_noise_lut{5};
+constexpr TextureUnit proctex_color_map{6};
+constexpr TextureUnit proctex_alpha_map{7};
+constexpr TextureUnit proctex_lut{8};
+constexpr TextureUnit proctex_diff_lut{9};
```
wwylele: I probably should name these as `LightingLUT`, `FogLUT` etc?

yuriks:
:lgtm: wwylele for Citra president 2017

---

Reviewed 6 of 6 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[src/video_core/renderer_opengl/gl_rasterizer.cpp, line 53 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8ByU0OYCChpNTGt39:-Kn8ByU0OYCChpNTGt3A:b-9enn9l) ([raw file](https://github.com/citra-emu/citra/blob/809ecb88d3ab9f932bd55a7b3ec1ac968fafb48e/src/video_core/renderer_opengl/gl_rasterizer.cpp#L53)):*
> ```cpp
>     uniform_block_data.dirty = true;
> 
>     std::fill(uniform_block_data.lut_dirty.begin(), uniform_block_data.lut_dirty.end(), true);
> ```

`uniform_block_data.lut_dirty.fill(true);`

---

*[src/video_core/renderer_opengl/gl_rasterizer.cpp, line 103 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8DeVQpJhMfXRrcRrh:-Kn8DeVQpJhMfXRrcRri:b-ytk9fq) ([raw file](https://github.com/citra-emu/citra/blob/809ecb88d3ab9f932bd55a7b3ec1ac968fafb48e/src/video_core/renderer_opengl/gl_rasterizer.cpp#L103)):*
> ```cpp
>     lighting_lut.Create();
>     glBindBuffer(GL_TEXTURE_BUFFER, lighting_lut_buffer.handle);
>     glBufferData(GL_TEXTURE_BUFFER, 256 * 8 * 24, nullptr, GL_DYNAMIC_DRAW);
> ```

`sizeof(GLfloat) * 2 * 256 * 24` (add more named constants as needed)

---

*[src/video_core/renderer_opengl/gl_rasterizer.h, line 266 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8DNYIPdGBZvA_uk8b:-Kn8DNYIPdGBZvA_uk8c:b9diq98) ([raw file](https://github.com/citra-emu/citra/blob/809ecb88d3ab9f932bd55a7b3ec1ac968fafb48e/src/video_core/renderer_opengl/gl_rasterizer.h#L266)):*
> 24
There are a bunch of references to this number now. I think it would be good to add a named constant for it, possibly in regs_lighting.h.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 1068 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8G2E3GqRi2kRv_uVx:-Kn8G2E4XT-MHVHi0YxE:bt27ahc) ([raw file](https://github.com/citra-emu/citra/blob/809ecb88d3ab9f932bd55a7b3ec1ac968fafb48e/src/video_core/renderer_opengl/gl_shader_gen.cpp#L1068)):*
> ```cpp
> 
> float LookupLightingLUT(int lut_index, int index, float delta) {
>     return dot(texelFetch(lighting_lut, lut_index * 256 + index).rg, vec2(1.0, delta));
> ```

The `dot` thing is cute but I think this would be clearer if done manually.

---

*[src/video_core/renderer_opengl/gl_state.h, line 12 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8M1QhOUaFPwNItMBA:-Kn8M1QhOUaFPwNItMBB:bctqvi4) ([raw file](https://github.com/citra-emu/citra/blob/035542040f506bcdfdc704e1e5086976577c2383/src/video_core/renderer_opengl/gl_state.h#L12)):*
> ```objc
> 
> class TextureUnit {
> public:
> ```

Should just make this a struct, I don't think it's ever going to be anything more than a simple aggregate.

---

*[src/video_core/renderer_opengl/gl_state.h, line 14 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8MSSI4VebGxPL3XB7:-Kn8MSSI4VebGxPL3XB8:bggn4ul) ([raw file](https://github.com/citra-emu/citra/blob/035542040f506bcdfdc704e1e5086976577c2383/src/video_core/renderer_opengl/gl_state.h#L14)):*
> ```objc
> public:
>     int id;
>     constexpr operator GLint() const {
> ```

I dislike these conversion operators. For one, they rely on GLint and GLenum being distinct integral types (I think GLint is signed int and GLenum is unsigned int). If you pass this to a function accepting unsigned int then you'll silently get the GLenum conversion instead, which will be hard to spot. I'd much rather these be explicit functions. (`id` can probably be accessed directly, and then add a `Enum()` member to get the enum version of it.)

Alternatively, drop the type altogether, and have a enum + a free function to do the id -> enum conversion.

---

*[src/video_core/renderer_opengl/gl_state.h, line 32 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn3PRWWPoeMhjkV1xSP-r2-32:-Kn8Llz0R0x21kaO0JIa:b-eb6bv0) ([raw file](https://github.com/citra-emu/citra/blob/035542040f506bcdfdc704e1e5086976577c2383/src/video_core/renderer_opengl/gl_state.h#L32)):*

I probably should name these as `LightingLUT`, `FogLUT` etc?
</blockquote></details>

We don't really have a convention for constants. (We should though.) So, *shrug*

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2792#-:-Kn8NxAzWCNUxwr99yfW:beusjuq)*

wwylele:

---

*[src/video_core/renderer_opengl/gl_rasterizer.cpp, line 53 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8ByU0OYCChpNTGt39:-KnAvo44gvDalz6fG779:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/809ecb88d3ab9f932bd55a7b3ec1ac968fafb48e/src/video_core/renderer_opengl/gl_rasterizer.cpp#L53)):*

`uniform_block_data.lut_dirty.fill(true);`
</blockquote></details>

Done.

---

*[src/video_core/renderer_opengl/gl_rasterizer.cpp, line 103 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8DeVQpJhMfXRrcRrh:-KnAvoJA6gK2iJKuOHN2:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/809ecb88d3ab9f932bd55a7b3ec1ac968fafb48e/src/video_core/renderer_opengl/gl_rasterizer.cpp#L103)):*

`sizeof(GLfloat) * 2 * 256 * 24` (add more named constants as needed)
</blockquote></details>

Done.

---

*[src/video_core/renderer_opengl/gl_rasterizer.h, line 266 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8DNYIPdGBZvA_uk8b:-KnAvooaUiwtKuBGLMED:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/809ecb88d3ab9f932bd55a7b3ec1ac968fafb48e/src/video_core/renderer_opengl/gl_rasterizer.h#L266)):*

> 24
There are a bunch of references to this number now. I think it would be good to add a named constant for it, possibly in regs_lighting.h.
</blockquote></details>

Done.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 1068 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8G2E3GqRi2kRv_uVx:-KnAvpIJZ8cQPgZLwR3P:bb6un1j) ([raw file](https://github.com/citra-emu/citra/blob/809ecb88d3ab9f932bd55a7b3ec1ac968fafb48e/src/video_core/renderer_opengl/gl_shader_gen.cpp#L1068)):*

The `dot` thing is cute but I think this would be clearer if done manually.
</blockquote></details>

Done

---

*[src/video_core/renderer_opengl/gl_state.h, line 12 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8M1QhOUaFPwNItMBA:-KnAvqbnmVR7jg_44KQK:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/035542040f506bcdfdc704e1e5086976577c2383/src/video_core/renderer_opengl/gl_state.h#L12)):*

Should just make this a struct, I don't think it's ever going to be anything more than a simple aggregate.
</blockquote></details>

Done.

---

*[src/video_core/renderer_opengl/gl_state.h, line 14 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2792#-Kn8MSSI4VebGxPL3XB7:-KnAvqyKnnRwgLBsC734:b-3y2uj9) ([raw file](https://github.com/citra-emu/citra/blob/035542040f506bcdfdc704e1e5086976577c2383/src/video_core/renderer_opengl/gl_state.h#L14)):*

I dislike these conversion operators. For one, they rely on GLint and GLenum being distinct integral types (I think GLint is signed int and GLenum is unsigned int). If you pass this to a function accepting unsigned int then you'll silently get the GLenum conversion instead, which will be hard to spot. I'd much rather these be explicit functions. (`id` can probably be accessed directly, and then add a `Enum()` member to get the enum version of it.)

Alternatively, drop the type altogether, and have a enum + a free function to do the id -> enum conversion.
</blockquote></details>

Done. I still keep the struct way because using enum + free function means two namespace name in one call. They should be in a namespace, but on the other hand it is too long.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2792)*

yuriks:
Reviewed 3 of 6 files at r5, 1 of 3 files at r6, 1 of 1 files at r7, 1 of 1 files at r8.

---

*[src/video_core/regs_lighting.h, line 29 at r8](https://reviewable.io:443/reviews/citra-emu/citra/2792#-KnB9PH9gh0NYLbEPCS6:-KnB9PH9gh0NYLbEPCS7:b5rxthj) ([raw file](https://github.com/citra-emu/citra/blob/f0722eb64da2a58703d71ded40ec80e284013a21/src/video_core/regs_lighting.h#L29)):*
> ```objc
>     };
> 
>     static constexpr unsigned MaxLightingSampler = 24;
> ```

`NumLightingSamplers`, since 24 isn't a valid sampler itself

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2792)*

wwylele:

---

*[src/video_core/regs_lighting.h, line 29 at r8](https://reviewable.io:443/reviews/citra-emu/citra/2792#-KnB9PH9gh0NYLbEPCS6:-KnBBhjGUqXB5dWVJUcY:b-2znpdp) ([raw file](https://github.com/citra-emu/citra/blob/f0722eb64da2a58703d71ded40ec80e284013a21/src/video_core/regs_lighting.h#L29)):*

`NumLightingSamplers`, since 24 isn't a valid sampler itself
</blockquote></details>

I am not sure about this, because sampler 2 and 7 are likely invalid either, and the total number of samplers is actually 22. We just use 24 with two dummy samplers to simplify the code logic

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2792)*

yuriks:
Reviewed 2 of 5 files at r9, 1 of 3 files at r10, 1 of 1 files at r11, 1 of 1 files at r12.

---

*[src/video_core/regs_lighting.h, line 29 at r8](https://reviewable.io:443/reviews/citra-emu/citra/2792#-KnB9PH9gh0NYLbEPCS6:-KnBLK-Xb-FrtKwhXOEk:b-syvlen) ([raw file](https://github.com/citra-emu/citra/blob/f0722eb64da2a58703d71ded40ec80e284013a21/src/video_core/regs_lighting.h#L29)):*

I am not sure about this, because sampler 2 and 7 are likely invalid either, and the total number of samplers is actually 22. We just use 24 with two dummy samplers to simplify the code logic
</blockquote></details>

OK, Resolved.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2792)*

yuriks:
:shipit:
image

PR #2795, opened 2017-06-21
Change default UI background from white to black.

chris062689:
Changing the default game UI screen from white (1.0) to black (0.0) allows for easier viewing of the game window.

When taking screenshots for the compatibility database, we agreed on a black background would be easier. 

Making this the default background color.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2795)


LittleWhite-tb:
Well why not.
Nothing special to check -> LGTM

Lord-Haji:
i think it would be better if you could implement this as a dark/night mode option not default @bunnei @chris062689

LittleWhite-tb:
@Lord-Haji See #2804

PR #2798, opened 2017-06-22
Kernel: Implement CreateSession SVC

yuriks:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2798)

### DIFF
--- a/src/core/hle/function_wrappers.h
+++ b/src/core/hle/function_wrappers.h
@@ -232,6 +232,16 @@ void Wrap() {
     FuncReturn(retval);
 }
 
+template <ResultCode func(Kernel::Handle*, Kernel::Handle*)>
+void Wrap() {
+    Kernel::Handle param_1 = 0;
+    Kernel::Handle param_2 = 0;
+    u32 retval = func(&param_1, &param_2).raw;
+    Core::CPU().SetReg(1, param_1);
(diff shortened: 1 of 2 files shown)


src/core/hle/function_wrappers.h
```diff
@@ -232,6 +232,17 @@ void Wrap() {
     FuncReturn(retval);
 }
 
+template <ResultCode func(Kernel::Handle*, Kernel::Handle*)>
+void Wrap() {
+    Kernel::Handle param_1 = 0;
+    Kernel::Handle param_2 = 0;
+    u32 retval = func(&param_1, &param_2).raw;
+    // The first out parameter is moved into R2 and the second is moved into R1.
```
Subv: Are you sure this order is correct? From looking at the kernel disassembly, the client session is stored in R2

yuriks: Classic case of two wrongs make a right: I'd flipped the assignment to the output params in CreateSession. Fixed.

However, I'm now pretty sure that CreatePort is incorrect. Both functions have identical stubs in ctrulib/the kernel, and since I just verified this one is correct, CreatePort must be wrong. I can't easily verify that right now however.

Subv: I checked the assembly, CreatePort indeed seems to be incorrect

PR #2801, opened 2017-06-23
Implement CreateSessionToPort and AcceptSession; fix CreatePort

yuriks:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2801)

### DIFF
--- a/src/core/hle/kernel/server_port.h
+++ b/src/core/hle/kernel/server_port.h
@@ -14,6 +14,7 @@
 namespace Kernel {
 
 class ClientPort;
+class ServerSession;
 class SessionRequestHandler;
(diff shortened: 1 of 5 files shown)


src/core/hle/kernel/server_port.h
```diff
@@ -40,6 +41,8 @@ class ServerPort final : public WaitObject {
         return HANDLE_TYPE;
     }
 
+    ResultVal<SharedPtr<ServerSession>> Accept();
```
Subv: docstring

yuriks: Done.

PR #2803, opened 2017-06-23
Implement basic virtual Room support based on enet

B3n30:
# Description of this PR
This will allow to create rooms (aka server) and room_members (aka clients), and connect them to each other.
There is still a UI missing (which will come in a later PR), so establish connections isn't possible yet.
This is just for including the needed external lib ENet, which we rely on. This will allow to send data over UDP.
There is for now no support for actually sending and receiving any data. This will come in a later PR.

There was the idea to have this as an external. We can later change it to an external lib. But for now we want to rely on the citra review process as well as get it partially into citra. That way we can use and test the actually implemented uds/connecting code. The design is to be able to convert most of the stuff to an external if this is wished/needed

# General description
Scope of this design:
    - We are not a chat program
            - Text chat only, no formatting, ...
    - We are not a matchmaking service
    - We are not ...
    
- Concept of rooms
      - Someone creates a room
      - Rooms can be created from the main menu and are independent of the game / UDS service.
- Rooms will act as a virtual neighborhood that simulates being physically near the other 3DSs (Or in this case, Citras).
      - Others join that room
      - All data in the room is forwarded amongst connected members, no checking on data is made
      - If you create a room, it's a dedicated server of sorts
        - So you start a Room but also join "localhost" [in the Citra frontend this will be coupled, but code wise it's 2 seprate things]

 - Each connected member..
      - ..has a nickname
                - The nickname is checked for collisions on connect, it can't be changed afterwards
     - ...has a MAC address
               - The room can create a free random MAC for each member
                - The MAC is checked for collisions on connect, it can't be changed afterwards

There might be an dedicated room app later
It will be possible to write a program that captures real 3ds WLAN data and transmit them to our citra network

# Roadmap
This is the first of a view PRs: 
1. This PR
2. Implement the receiving threats for room and room_member
3. Implement the sending/receiving the join request and the chat messages
3. Implement the sending/receiving of game data
4. Callbacks
5. UI to create a room/connect/chat window

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2803)


mailwl:
is it work with Mario Kart? (can't check right now)

B3n30:
@mailwl Not now. If we reach the end of this roadmap, and Subv finished all of the uds stuff... 
And Mario Kart also needs some Download Play stuff, thats not in citra now. 
But when that all is done, then: yes

tywald:
Are there any limitations similar to the ad hoc implementation that PPSSPP uses?(only one client per IP address can connect even if the MAC address is different). Or does it connect clients solely on the MAC address?

I can't wait to test Monster Hunter games when everything is finished with the UDS implementations etc.

B3n30:
@tywald You can connect multiple room_members(clients) from the same ip. But there will still be some limitations, like a maximum number of members connected to one room and of course the limitations of each game itself remain.

src/network/network.cpp
```diff
@@ -0,0 +1,46 @@
...
+    g_room_member = std::make_shared<RoomMember>();
+    LOG_DEBUG(Network, "initialized OK");
+    return true;
+}
+
+std::weak_ptr<Room> GetRoom() {
+    return g_room;
+}
+
+std::weak_ptr<RoomMember> getRoomMember() {
```
tgsm: Looks like this is supposed to be capitalized

B3n30: done

src/network/network.cpp
```diff
@@ -0,0 +1,46 @@
...
+// Refer to the license.txt file included.
+
+#include "common/assert.h"
+#include "common/logging/log.h"
+#include "enet/enet.h"
+#include "network/network.h"
+
+namespace Network {
+
+std::shared_ptr<RoomMember> g_room_member; ///< RoomMember (Client) for network games
```
bunnei: These should be `static`

src/network/room.h
```diff
@@ -0,0 +1,65 @@
...
+
+#pragma once
+
+#include <atomic>
+#include <string>
+#include "enet/enet.h"
+
+namespace Network {
+
+const uint16_t DefaultRoomPort = 1234;
```
bunnei: nit: `u16`? (and `u32` below)
This is just what we use everywhere in the emulator...

src/network/room.h
```diff
@@ -0,0 +1,65 @@
...
+struct RoomInformation {
+    std::string name;      // Name of the server
+    uint32_t member_slots; // Maximum number of members in this room
+};
+
+// This is what a server [person creating a server] would use.
+class Room final {
+public:
+    enum class State {
+        Open,   // The room is open and ready to accept connections.
```
bunnei: nit: use `///<` here to denote docstrings

src/network/room.h
```diff
@@ -0,0 +1,65 @@
...
+
+const uint16_t DefaultRoomPort = 1234;
+const size_t NumChannels = 1; // Number of channels used for the connection
+
+struct RoomInformation {
+    std::string name;      // Name of the server
+    uint32_t member_slots; // Maximum number of members in this room
+};
+
+// This is what a server [person creating a server] would use.
```
bunnei: nit: Should always be a docstring for class comment

src/network/room.h
```diff
@@ -0,0 +1,65 @@
...
+#include <string>
+#include "enet/enet.h"
+
+namespace Network {
+
+const uint16_t DefaultRoomPort = 1234;
+const size_t NumChannels = 1; // Number of channels used for the connection
+
+struct RoomInformation {
+    std::string name;      // Name of the server
```
bunnei: nit: use `///<` here to denote docstrings

src/network/room.h
bunnei: These can be `constexpr`

src/network/room_member.cpp
```diff
@@ -0,0 +1,58 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/assert.h"
+#include "network/room_member.h"
+
+namespace Network {
+
+const uint32_t ConnectionTimeout = 5000; // ms
```
bunnei: `constexpr`

src/network/room_member.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+RoomMember::~RoomMember() {
+    ASSERT_MSG(!IsConnected(), "RoomMember is being destroyed while connected");
+    enet_host_destroy(client);
+}
+
+void RoomMember::Join(const std::string& nickname, const std::string& server, uint16_t server_port,
+                      uint16_t client_port) {
+    ENetAddress address;
```
bunnei: Should this be initialized in case the line below fails? e.g. `ENetAddress address{};`

src/network/room_member.cpp
```diff
@@ -0,0 +1,58 @@
...
+    enet_address_set_host(&address, server.c_str());
+    address.port = server_port;
+
+    this->server = enet_host_connect(client, &address, NumChannels, 0);
```
bunnei: nit: No need for `this->`, applies here and below

src/network/room_member.cpp
```diff
@@ -0,0 +1,58 @@
...
+    client = enet_host_create(nullptr, 1, NumChannels, 0, 0);
+    ASSERT_MSG(client != nullptr, "Could not create client");
+}
+
+RoomMember::~RoomMember() {
+    ASSERT_MSG(!IsConnected(), "RoomMember is being destroyed while connected");
+    enet_host_destroy(client);
+}
+
+void RoomMember::Join(const std::string& nickname, const std::string& server, uint16_t server_port,
```
bunnei: nit: server_addr instead of server, to prevent name duplication/confusion below?

bunnei: Should these take in `const char*` instead of std::string? (since that's what the enet api uses, and may prevent unnecessary construction of `std::string` - Just a thought, if this makes it much easier for callers than leave it

src/network/room_member.cpp
```diff
@@ -0,0 +1,58 @@
...
+                      uint16_t client_port) {
+    ENetAddress address;
+    enet_address_set_host(&address, server.c_str());
+    address.port = server_port;
+
+    this->server = enet_host_connect(client, &address, NumChannels, 0);
+
+    if (this->server == nullptr) {
```
bunnei: nit: `if (!server) {` (less code is better)

src/network/room_member.cpp
```diff
@@ -0,0 +1,58 @@
...
+        state = State::Error;
+        return;
+    }
+
+    ENetEvent event;
```
bunnei: Should this be initialized in case the line below fails? e.g. `ENetEvent event{};`

src/network/room_member.h
```diff
@@ -0,0 +1,66 @@
...
+
+#pragma once
+
+#include <atomic>
+#include <string>
+#include "network/room.h"
+
+namespace Network {
+
+// This is what a client [person joining a server] would use.
```
bunnei: Docstring

src/network/room_member.cpp
```diff
@@ -0,0 +1,58 @@
...
+    int net = enet_host_service(client, &event, ConnectionTimeout);
+    if (net > 0 && event.type == ENET_EVENT_TYPE_CONNECT) {
+        this->nickname = nickname;
```
bunnei: No `this`

src/network/room_member.h
```diff
@@ -0,0 +1,66 @@
...
+// It also has to be used if you host a game yourself (You'd create both, a Room and a
+// RoomMembership for yourself)
+class RoomMember final {
+public:
+    enum class State {
```
bunnei: define underlying type? e.g.
`enum class State : u32 {`

src/network/room.h
```diff
@@ -0,0 +1,65 @@
...
+    uint32_t member_slots; // Maximum number of members in this room
+};
+
+// This is what a server [person creating a server] would use.
+class Room final {
+public:
+    enum class State {
```
bunnei: Define underlying type?
e.g. `enum class State : u32 {`

src/network/room_member.h
```diff
@@ -0,0 +1,66 @@
...
+    void Join(const std::string& nickname, const std::string& server = "127.0.0.1",
+              const uint16_t serverPort = DefaultRoomPort, const uint16_t clientPort = 0);
+
+    /**
+     * Leaves the current room.
+     */
+    void Leave();
+
+private:
+    ENetHost* client;                      ///< ENet network interface.
```
bunnei: Initialize these? e.g.

`ENetHost* client{};`

src/network/room_member.h
```diff
@@ -0,0 +1,66 @@
...
+        Idle,    ///< Default state
+        Error,   ///< Some error [permissions to network device missing or something]
+        Joining, ///< The client is attempting to join a room.
+        Joined,  ///< The client is connected to the room and is ready to send/receive packets.
+        LostConnection, ///< Connection closed
+
+        // Reasons why connection was rejected
+        NameCollision,  ///< Somebody is already using this name
+        MacCollision,   ///< Somebody is already using that mac-address
+        CouldNotConnect ///< The room is not responding to a connection attempt
```
bunnei: Should we have a way to kick users? (would that just be `LostConnection`?)

src/network/network.cpp
bunnei: More global state == evil

(I realize you're just following the same pattern we use everywhere)

src/network/room_member.cpp
Subv: There's a "server" parameter too

src/network/room_member.h
B3n30: There might be a way later. But this will require a lot more stuff (remote commanding the room, etc.) which isn't necessary for a working netplay.
Same for password protected rooms, etc.

src/network/network.cpp
B3n30: For now i will stay with the globals. But it's on the todo list and will be in a separate PR.

src/network/room.h
```diff
@@ -0,0 +1,66 @@
...
+
+    Room() = default;
+    ~Room() = default;
+
+    /**
+     * Gets the current state of the room.
+     */
+    State GetState() const {
+        return state;
+    };
```
lioncash: The semicolon here isn't necessary

src/network/room.h
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Gets the room information of the room.
+     */
+    const RoomInformation& GetRoomInformation() const {
+        return room_information;
+    };
```
lioncash: Ditto

src/network/room.h
```diff
@@ -0,0 +1,66 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <atomic>
+#include <string>
+#include "common/common_types.h"
+#include "enet/enet.h"
```
lioncash: You may want to consider the [PImpl idiom](http://en.cppreference.com/w/cpp/language/pimpl) here which prevents enet includes propagating via this header.

src/network/network.cpp
```diff
@@ -0,0 +1,47 @@
...
+    if (g_room) {
+        if (g_room->GetState() == Room::State::Open)
+            g_room->Destroy();
+        g_room.reset();
+    }
+    enet_deinitialize();
+    LOG_DEBUG(Network, "shutdown OK");
+}
+
+} // namespace
```
lioncash: `// namespace Network`

src/network/network.h
```diff
@@ -0,0 +1,24 @@
...
+/// Returns a pointer to the room handle
+std::weak_ptr<Room> GetRoom();
+
+/// Returns a pointer to the room member handle
+std::weak_ptr<RoomMember> GetRoomMember();
+
+/// Unregisters the network device, the room, and the room member and shut them down.
+void Shutdown();
+
+} // namespace
```
lioncash: `// namespace Network`

src/network/room.cpp
```diff
@@ -0,0 +1,39 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "network/room.h"
+
+namespace Network {
+
+/// Maximum number of concurrent connections allowed to this room.
+static const u32 MaxConcurrentConnections = 10;
```
lioncash: This can just be `constexpr`

src/network/room.h
```diff
@@ -0,0 +1,66 @@
...
+    void Destroy();
+
+private:
+    ENetHost* server = nullptr;              ///< network interface.
+    std::atomic<State> state{State::Closed}; ///< Current state of the room.
+
+    RoomInformation room_information; ///< Information about this room.
+};
+
+} // namespace
```
lioncash: `// namespace Network`

src/network/room_member.cpp
```diff
@@ -0,0 +1,58 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/assert.h"
+#include "network/room_member.h"
+
+namespace Network {
+
+constexpr u32 ConnectionTimeout = 5000; // ms
```
lioncash: You can just name this `ConnectionTimeoutMs` to get rid of the need for the comment.

src/network/room_member.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+void RoomMember::Leave() {
+    enet_peer_disconnect(server, 0);
+    state = State::Idle;
+    // TODO(B3N30): Close the receive thread
+    enet_peer_reset(server);
+}
+
+} // namespace
```
lioncash: `namespace Network`

src/network/room_member.h
```diff
@@ -0,0 +1,68 @@
...
+
+    RoomMember();
+    ~RoomMember();
+
+    /**
+     * Returns the status of our connection to the room.
+     */
+    State GetState() const {
+        return state;
+    };
```
lioncash: The semicolon here isn't necessary.

src/network/room_member.h
```diff
@@ -0,0 +1,68 @@
...
+
+private:
+    ENetHost* client = nullptr;            ///< ENet network interface.
+    ENetPeer* server = nullptr;            ///< The server peer the client is connected to
+    std::atomic<State> state{State::Idle}; ///< Current state of the RoomMember.
+
+    std::string nickname; ///< The nickname of this member.
+};
+
+} // namespace
```
lioncash: `// namespace Network`

src/citra_qt/CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ if (APPLE)
 else()
     add_executable(citra-qt ${SRCS} ${HEADERS} ${UI_HDRS})
 endif()
-target_link_libraries(citra-qt PRIVATE audio_core common core input_common video_core)
+target_link_libraries(citra-qt PRIVATE audio_core common core input_common network video_core)
```
wwylele: Just want to confirm the overall design here: so the `core` doesn't depends on `network`, but only exposes some interfaces/callbacks, and let the front end link them with `network`, right?

src/network/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+            room.h
+            room_member.h
+            )
+
+include_directories(../../externals/enet/include)
+
+create_directory_groups(${SRCS} ${HEADERS})
+
+add_library(network STATIC ${SRCS} ${HEADERS})
+target_link_libraries(network enet)
```
wwylele: Also needs to link against `common`

src/network/room.h
```diff
@@ -0,0 +1,66 @@
...
+    void Create(const std::string& name, const std::string& server = "",
+                u16 server_port = DefaultRoomPort);
+
+    /**
+     * Destroys the socket
+     */
+    void Destroy();
+
+private:
+    ENetHost* server = nullptr;              ///< network interface.
```
lioncash: First char should be capitalized like the rest.

src/network/network.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "network/network.h"
+
+namespace Network {
+
+static std::shared_ptr<RoomMember> g_room_member; ///< RoomMember (Client) for network games
+static std::shared_ptr<Room> g_room;              ///< Room (Server) for network games
+// TODO(B3N30): Put these globals into a networking class
+
+bool Init() {
+    ASSERT_MSG(enet_initialize() == 0, "Error initalizing network");
```
wwylele: not sure in which case `enet_initialize` would fail, but IMO this failure shouldn't just crash the whole program. Can just log a message and keep the network module uninitialized and unusable

src/network/room.h
wwylele: I only see a `ENetHost` being used as a pointer, so I guess a forward declaration can work?

src/citra_qt/CMakeLists.txt
B3n30: nope. core will indeed depend on network. But since there is no need for it in this PR it network in't linked to core, yet.

src/network/room_member.h
```diff
@@ -0,0 +1,68 @@
...
+    void Join(const std::string& nickname, const char* server_addr = "127.0.0.1",
+              const u16 serverPort = DefaultRoomPort, const u16 clientPort = 0);
+
+    /**
+     * Leaves the current room.
+     */
+    void Leave();
+
+private:
+    ENetHost* client = nullptr;            ///< ENet network interface.
```
B3n30: I don't think pimp'ing here is needed. Instead I forward declared ENetHost and ENetPeer

src/network/room.h
lioncash: Right, but my reason for suggesting the PImpl idiom here is it doesn't introduce anything external into the surrounding code. It keeps implementation details fully hidden (which is usually desirable with external non-template libraries).

You also can't forward-declare typedefs (i.e. ENetHost is a typedef of `_ENetHost`). So this would require something like:

```cpp
using ENetHost = _ENetHost;
```

which is bad because this makes the library depend on an internal library name.

src/network/room_member.h
lioncash: See [this comment](#discussion_r124628594) on why that's likely not a good idea in this case (tl;dr: because you're re-declaring an internal library data structure name outside of the library itself).

B3n30:
Review status: 0 of 15 files reviewed at latest revision, 36 unresolved discussions.

---

*[src/citra_qt/CMakeLists.txt, line 94 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSU-r6-94:-KnkMQRmFSKaL5RvRWcD:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/citra_qt/CMakeLists.txt#L94)):*

nope. core will indeed depend on network. But since there is no need for it in this PR it network in't linked to core, yet.
</blockquote></details>

Done.

---

*[src/network/CMakeLists.txt, line 18 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSY-r6-18:-KnkMWWdEtAfJTwjp8Ge:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/CMakeLists.txt#L18)):*

Also needs to link against `common`
</blockquote></details>

Done.

---

*[src/network/network.cpp, line 12 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSZ-r4-12:-KnkMUbHmD476OtCaHGF:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/network.cpp#L12)):*

For now i will stay with the globals. But it's on the todo list and will be in a separate PR.
</blockquote></details>

Done.

---

*[src/network/network.cpp, line 17 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSZ-r6-17:-KnkMY21ViRUlObj2VNB:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/network.cpp#L17)):*

not sure in which case `enet_initialize` would fail, but IMO this failure shouldn't just crash the whole program. Can just log a message and keep the network module uninitialized and unusable
</blockquote></details>

Done.

---

*[src/network/network.cpp, line 47 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSZ-r6-47:-KnkMbP_lnhs2hdzbA5p:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/network.cpp#L47)):*

`// namespace Network`
</blockquote></details>

Done.

---

*[src/network/network.h, line 24 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeS_-r6-24:-KnkMcFPG9K9DNZKAhaz:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/network.h#L24)):*

`// namespace Network`
</blockquote></details>

Done.

---

*[src/network/room.cpp, line 10 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSc-r6-10:-KnkMd2AOcfyUtBFmkNE:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room.cpp#L10)):*

This can just be `constexpr`
</blockquote></details>

Done.

---

*[src/network/room.cpp, line 39 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSc-r6-39:-KnkMehzb7wVphF2mbRk:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room.cpp#L39)):*

`// namespace Network`
</blockquote></details>

Done.

---

*[src/network/room.h, line 13 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r4-13:-KnkMiDlCf7HLggj9M9M:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room.h#L13)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

These can be `constexpr`
</blockquote></details>

Done.

---

*[src/network/room.h, line 17 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r4-17:-KnkMjdHQkl8ewyxe-dJ:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room.h#L17)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

nit: use `///<` here to denote docstrings
</blockquote></details>

Done.

---

*[src/network/room.h, line 21 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r4-21:-KnkMkKmCq67RfoHAvsJ:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room.h#L21)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

nit: Should always be a docstring for class comment
</blockquote></details>

Done.

---

*[src/network/room.h, line 24 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r4-24:-KnkMmrK2-e0y-H5GCaR:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room.h#L24)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

Define underlying type?
e.g. `enum class State : u32 {`
</blockquote></details>

Done.

---

*[src/network/room.h, line 25 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r4-25:-KnkMneJh2eE4A8tsY5l:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room.h#L25)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

nit: use `///<` here to denote docstrings
</blockquote></details>

Done.

---

*[src/network/room.h, line 10 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r6-10:-KnkMgUZD4FAHusmrjvT:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room.h#L10)):*

Right, but my reason for suggesting the PImpl idiom here is it doesn't introduce anything external into the surrounding code. It keeps implementation details fully hidden (which is usually desirable with external non-template libraries).

You also can't forward-declare typedefs (i.e. ENetHost itself is a typedef of the struct `_ENetHost`). So this would require something like:

```cpp
using ENetHost = _ENetHost;
```

which is bad because this makes our library depend on an external library's internal name directly.
</blockquote></details>

Done.

---

*[src/network/room.h, line 38 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r6-38:-KnkMorzrPfd3aoocJmv:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room.h#L38)):*

The semicolon here isn't necessary
</blockquote></details>

Done.

---

*[src/network/room.h, line 45 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r6-45:-KnkMqstCkBSovuJBXNR:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room.h#L45)):*

Ditto
</blockquote></details>

Done.

---

*[src/network/room.h, line 60 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r6-60:-KnkMs8w-xM-LL0BlwMm:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room.h#L60)):*

First char should be capitalized like the rest.
</blockquote></details>

Done.

---

*[src/network/room.h, line 66 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSd-r6-66:-KnkMt4mzgvPcaJT65vH:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room.h#L66)):*

`// namespace Network`
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 10 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSe-r4-10:-KnkMuIR95nrXGjbuVa0:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.cpp#L10)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

`constexpr`
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 22 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSe-r4-22:-KnkMwyV7iohPXEu2Du9:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.cpp#L22)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

Should these take in `const char*` instead of std::string? (since that's what the enet api uses, and may prevent unnecessary construction of `std::string` - Just a thought, if this makes it much easier for callers than leave it
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 24 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSe-r4-24:-KnkMyngDqbt8g0nQTRS:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.cpp#L24)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

Should this be initialized in case the line below fails? e.g. `ENetAddress address{};`
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 28 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSe-r4-28:-KnkN-3EydE-sPfYaJkz:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.cpp#L28)):*

There's a "server" parameter too
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 30 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSe-r4-30:-KnkN04fUGhX_VK4eppM:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.cpp#L30)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

nit: `if (!server) {` (less code is better)
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 35 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSe-r4-35:-KnkNDq6oyMOjGZHG6FQ:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.cpp#L35)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

Should this be initialized in case the line below fails? e.g. `ENetEvent event{};`
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 38 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSe-r4-38:-KnkN1FIv1kipKrw3rb1:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.cpp#L38)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

No `this`
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 10 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSe-r6-10:-KnkMvHsGLaYWXlyIe_n:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room_member.cpp#L10)):*

You can just name this `ConnectionTimeoutMs` to get rid of the need for the comment.
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 58 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSe-r6-58:-KnkN22nWd4czuQf317v:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room_member.cpp#L58)):*

`namespace Network`
</blockquote></details>

Done.

---

*[src/network/room_member.h, line 13 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSf-r4-13:-KnkN4H4pgtp51dz0niP:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.h#L13)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

Docstring
</blockquote></details>

Done.

---

*[src/network/room_member.h, line 18 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSf-r4-18:-KnkN5bQWgi3kSctc7KW:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.h#L18)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

define underlying type? e.g.
`enum class State : u32 {`
</blockquote></details>

Done.

---

*[src/network/room_member.h, line 28 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSf-r4-28:-KnkN85miCptewjt4NBD:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.h#L28)):*

There might be a way later. But this will require a lot more stuff (remote commanding the room, etc.) which isn't necessary for a working netplay.
Same for password protected rooms, etc.
</blockquote></details>

Done.

---

*[src/network/room_member.h, line 59 at r4](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSf-r4-59:-KnkN9mMF4HU3xt18oru:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/a86d94cb406551ce764604b7be5a9348432f593d/src/network/room_member.h#L59)):*
<details><summary><i>Previously, bunnei wrote…</i></summary><blockquote>

Initialize these? e.g.

`ENetHost* client{};`
</blockquote></details>

Done.

---

*[src/network/room_member.h, line 9 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSf-r6-9:-KnkN3NUDb7ELROiYZm7:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room_member.h#L9)):*

`common/common_types.h` needs to be included.
</blockquote></details>

Done.

---

*[src/network/room_member.h, line 41 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSf-r6-41:-KnkN7TPEAZC3jgiIkS7:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room_member.h#L41)):*

The semicolon here isn't necessary.
</blockquote></details>

Done.

---

*[src/network/room_member.h, line 61 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSf-r6-61:-KnkNANs0wgt36RMDK7i:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room_member.h#L61)):*

See [this comment](#discussion_r124628594) on why that's likely not a good idea in this case (tl;dr: because you're re-declaring an internal library data structure name outside of the library itself).
</blockquote></details>

Done.

---

*[src/network/room_member.h, line 68 at r6](https://reviewable.io:443/reviews/citra-emu/citra/2803#-KnKooKoPb_890PVmeSf-r6-68:-KnkNB2GKZ-4qRG9Ix6j:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/c635565d1f8b9d6f4479e2fcbc9b7493048ed411/src/network/room_member.h#L68)):*

`// namespace Network`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2803)*

src/network/room_member.h
```diff
@@ -60,11 +58,8 @@ class RoomMember final {
     void Leave();
 
 private:
-    class RoomMemberPimpl;
-    std::unique_ptr<RoomMemberPimpl> room_member_pimpl;
-    std::atomic<State> state{State::Idle}; ///< Current state of the RoomMember.
-
-    std::string nickname; ///< The nickname of this member.
+    class RoomMemberImpl;
+    std::unique_ptr<RoomMemberImpl> room_member_impl_p;
```
Subv: Just name this ```impl```, same for the other one.

Issue #1181, opened 2015-10-01
Introduce a color theme management framework for the Qt GUI


neobrain:
Colors in debug widgets (or within the UI in general) shouldn't be hardcoded, but the UI designer quite often still needs some way to highlight text in some way. Hardcoding colors in this case often works for "standard" system themes, but causes poor readability whenever the program user uses a non-standard color theme (e.g. dark ones). A solution is to have a central code framework to request different types of "highlighting" colors. The concrete values of these colors could either be inferred from the system theme (which will be hard to implement) or could be user-adjustable (which would be easy and flexible enough). 

An example of such a framework is KDE's KColorScheme (cf http://api.kde.org/frameworks-api/frameworks5-apidocs/kconfigwidgets/html/classKColorScheme.html). This need not be the same interface that we would use, but it's a good example of what idea I have in mind.

To get things going, we could start with a dummy implementation which only supports a single (hardcoded) color theme. The "hardcoding" is then localized to a central place rather than needing to fixup the whole source tree later on.

BreadFish64:
This has been implemented

PR #2804, opened 2017-06-24
citra-qt: UI Themes

Kloen:
Add Themes to citra-qt, yeah... themes. Related to #1181

Themes are packed inside citra and loaded as Qt resources.

Current theme added: https://github.com/ColinDuquesnoy/QDarkStyleSheet (as-is)

## As always, mandatory GIF!
themesandstuff

image

## Size VS!
Nightly Build - 6d436d7 ///// VS ////// Themes
image

There are currently some issues with the Widgets that have forced colors to their componets, I plan to address that later, in another PR, as well as glClearColor.

## Next PR
- Allow to change glClearColor based on the current theme.
- Allow to change Widgets hardcoded colors based on the theme.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2804)


src/citra_qt/CMakeLists.txt
```diff
@@ -75,7 +78,52 @@ set(UIS
...
+            themes/qdarkstyle/rc/stylesheet-vline.png
+            themes/qdarkstyle/rc/transparent.png
+            themes/qdarkstyle/rc/undock.png
+            themes/qdarkstyle/rc/up_arrow.png
+            themes/qdarkstyle/rc/up_arrow_disabled.png
+            themes/qdarkstyle/rc/Vmovetoolbar.png
+            themes/qdarkstyle/rc/Vsepartoolbar.png
+            )
+
+create_directory_groups(${SRCS} ${HEADERS} ${UIS} ${THEMES})
```
jroweboy: why are we adding the `qdarkstyle` folder to visual studio? Remove `${THEMES}` from the `create_directory_groups` call. Since we are getting rid of `theme.h` and `theme.cpp` theres no point in keeping the `themes` folder around as well.

src/citra_qt/CMakeLists.txt
```diff
@@ -21,6 +22,7 @@ set(SRCS
             debugger/profiler.cpp
             debugger/registers.cpp
             debugger/wait_tree.cpp
+            themes/themes.cpp
```
jroweboy: `themes.h` and `themes.cpp` are extremely empty files. I'd personally would put them in `ui_settings` instead since they really are only used to hold the list of possible themes.

src/citra_qt/configuration/config.cpp
```diff
@@ -134,6 +134,7 @@ void Config::ReadValues() {
     qt_config->endGroup();
 
     qt_config->beginGroup("UI");
+    UISettings::values.theme = qt_config->value("theme", "qt-default").toString();
```
jroweboy: Nit: After moving the list of themes to `UISettings` you can change this to so you don't have to hard code what the default is

```cpp
UISettings::values.theme = qt_config->value("theme", UISettings::themes[0].second).toString();
```

src/citra_qt/themes/themes.cpp
```diff
@@ -0,0 +1,13 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <QString>
+#include "citra_qt/themes/themes.h"
+
+std::vector<Theme> themes;
```
jroweboy: since this is really only used for populating the list of settings, it should be moved to `UISettings` instead. Then we can get rid of the `themes.h` and `themes.cpp` files

src/citra_qt/themes/themes.h
```diff
@@ -0,0 +1,18 @@
...
+
+#pragma once
+
+#include <vector>
+
+class QString;
+
+typedef std::pair<QString, QString> Theme;
+
+extern std::vector<Theme> themes;
```
jroweboy: (see https://github.com/citra-emu/citra/blob/master/src/core/settings.h#L55 for an example of where we do that)

src/citra_qt/main.cpp
```diff
@@ -791,6 +794,24 @@ void GMainWindow::filterBarSetChecked(bool state) {
     emit(OnToggleFilterBar());
 }
 
+void GMainWindow::UpdateUITheme() {
+    if (UISettings::values.theme != QString("qt-default")) {
+        QString theme_uri(":" + UISettings::values.theme + "/style.qss");
```
Kloen: Done

slashiee:
Should probably mention that QDarkStyleSheet uses the MIT license while Citra uses the GPL-2.0 license.

bunnei:
Yeah, I thought the idea was silly (inb4 Pokemon theme or something), but I think Dark looks really slick, and considering it's such a minimal change, I'm OK with it.

neico:
I feel like the disabled color seems a bit too dark in some parts, in most this can be overlooked because there's a mouseover highlight for them, but the port dropdown text on the Debug / GDB settings dialog for example doesn't have that which makes it pretty hard to read the port.

This of course isn't a deal breaker either but I believe that this might be a issue somewhere else in the future if left unattended.

PR #2807, opened 2017-06-24
Set global definition WIN32_LEAN_AND_MEAN

B3n30:
There occurred a problem in #2803 that needed to include winsocks2.h but windows.h included winsocks.h which caused a lot of redefinition errors.

With a global definition of WIN32_LEAN_AND_MEAN windows.h skips a lot of includes that are usually not used. But it might be necessary at some points in the code to add the missing includes manually. There is only one spot in citra (citra.cpp) where this is necessary. 

Defining WIN32_LEAN_AND_MEAN globally causes a lot of redefinition macro warnings, since many externals already use it. One solution would be to define WIN32_LEAN_AND_MEAN just everywhere windows.h is included but I think defining it manually is way less error-prone. 

It might also increase the build speed on windows.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2807)


CMakeLists.txt
```diff
@@ -79,6 +79,8 @@ else()
     add_definitions(/D_CRT_SECURE_NO_WARNINGS /D_CRT_NONSTDC_NO_DEPRECATE /D_SCL_SECURE_NO_WARNINGS)
     # Avoid windows.h junk
     add_definitions(/DNOMINMAX)
+    # Avoid windows.h from including some usally unused libs like winsocks.h, since this might cause some redifinition errors.
```
Subv: usally -> usually
redifinition -> redefinition

src/citra/citra.cpp
```diff
@@ -18,7 +18,10 @@
 #endif
 
 #ifdef _WIN32
+// windows.h needs to be included before shellapi.h
 #include <windows.h>
+
```
Subv: extra blank line

B3n30: The additional blank line is to get through the clang checks

PR #2811, opened 2017-06-28
configure_debug: Add label warning that CPU JIT needs to be disabled …

MerryMage:
…for gdbstub to work

screen shot 2017-06-28 at 08 56 44

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2811)

### DIFF
--- a/src/citra_qt/configuration/configure_debug.ui
+++ b/src/citra_qt/configuration/configure_debug.ui
@@ -23,6 +23,13 @@
        </property>
        <layout class="QVBoxLayout" name="verticalLayout_2">
         <item>
+         <widget class="QLabel">
+          <property name="text">
+           <string>The GDB Stub only works correctly when the CPU JIT is off.</string>
+          </property>
+         </widget>
+        </item>
+        <item>
          <layout class="QHBoxLayout" name="horizontalLayout_3">
           <item>
            <widget class="QCheckBox" name="toggle_gdbstub">


j-selby:
I do feel like this should be an active thing - e.g. warning prompt when both are enabled at once, rather than some text which people may or may not read/remember.

chris062689:
I agree with @j-selby, we could leverage @TheKoopaKingdom's recent PR about popup error messages for displaying it?

TheKoopaKingdom:
Hm, I don't think so. The [OnCoreError() slot](https://github.com/citra-emu/citra/blob/7af6931fb5bcb73640c5c2223eadbe72367a9b6b/src/citra_qt/main.cpp#L651) is meant more so for errors in the core, and not the frontend. Not only is it named for this use, but it uses the ResultStatus enum in the core, which would make adding errors for the Qt frontend a bit dirty. [OnCoreError()](https://github.com/citra-emu/citra/blob/7af6931fb5bcb73640c5c2223eadbe72367a9b6b/src/citra_qt/main.cpp#L651) also makes the assumption that a game is running, which doesn't necessarily seem to be the case here.

jroweboy:
at the very tangible risk of bikeshedding such a small change, i personally think disabling the gdb stub when cpu jit is turned on is the right thing to do (and disabling the checkbox while the cpu jit is enabled)

BUT... this message (or something similar) should be there whether we add any of the other suggestions or not, so I'm just going to merge this.

PR #2800, opened 2017-06-22
gl_rasterizer: use texture buffer for fog LUT

wwylele:
Follow up of #2792. Use texture buffer for fog LUT, and move entry decoding from shader to pica_state

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2800)

PR #2816, opened 2017-07-01
gl_rasterizer: use texture buffer for proctex LUT

wwylele:
Follow up of #2800 . I still kept these textures in separate unites because they don't have good numerical relationship to pack them into one texture like the lighting LUT one. We can pack them whenever it is more unit space is needed, anyway

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2816)


wwylele:
Last call for reviews...

PR #2818, opened 2017-07-09
Enable data transfer over ENet

B3n30:
This PR is a continuation of #2803 

It implements:

- Sending/handling JoinRequests
- Sending/receiving WifiPackets
- Disconnect handling
- Send/receive chat messages
- Send a game title

Things still to come:

1. A way to register and invoke callbacks
2. A way to register a room to our server api
3. Make use of the network implementation inside nwm::uds
4. The UI

It's probably best to review each commit separately.

To test this PR use [https://github.com/B3n30/citra-1/tree/network_test](https://github.com/B3n30/citra-1/tree/network_test)
It's currently possible to test:
- create room
- destroy room
- join room
- leave room
- send/receive chat messages
- receive RoomInformation

What's currently missing in test:
- send/receive WifiPackets
- send game title 

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2818)

### DIFF
--- a/src/network/packet.cpp
+++ b/src/network/packet.cpp
@@ -1,0 +1,229 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#ifdef _WIN32
+#include <winsock2.h>
+#else
+#include <arpa/inet.h>
+#endif
+#include <cstring>
+#include <string>
+#include "network/packet.h"
+
+namespace Network {
+
+Packet::Packet() : read_pos(0), is_valid(true) {}
+
+Packet::~Packet() {}
+
+void Packet::Append(const void* in_data, std::size_t size_in_bytes) {
+    if (in_data && (size_in_bytes > 0)) {
+        std::size_t start = data.size();
+        data.resize(start + size_in_bytes);
+        std::memcpy(&data[start], in_data, size_in_bytes);
+    }
+}
+
+void Packet::Read(void* out_data, std::size_t size_in_bytes) {
+    if (out_data && CheckSize(size_in_bytes)) {
+        std::memcpy(out_data, &data[read_pos], size_in_bytes);
+        read_pos += size_in_bytes;
+    }
+}
+
+void Packet::Clear() {
+    data.clear();
+    read_pos = 0;
+    is_valid = true;
+}
+
+const void* Packet::GetData() const {
+    return !data.empty() ? &data[0] : NULL;
+}
+
+void Packet::IgnoreBytes(u32 length) {
+    read_pos += length;
+}
+
+std::size_t Packet::GetDataSize() const {
+    return data.size();
+}
+
+bool Packet::EndOfPacket() const {
+    return read_pos >= data.size();
+}
+
+Packet::operator BoolType() const {
+    return is_valid ? &Packet::CheckSize : NULL;
+}
+
+Packet& Packet::operator>>(bool& out_data) {
+    u8 value;
+    if (*this >> value) {
+        out_data = (value != 0);
+    }
+    return *this;
+}
+
+Packet& Packet::operator>>(s8& out_data) {
+    Read(&out_data, sizeof(out_data));
+    return *this;
+}
+
+Packet& Packet::operator>>(u8& out_data) {
+    Read(&out_data, sizeof(out_data));
+    return *this;
+}
+
+Packet& Packet::operator>>(s16& out_data) {
+    s16 value;
+    Read(&value, sizeof(value));
+    out_data = ntohs(value);
+    return *this;
+}
+
+Packet& Packet::operator>>(u16& out_data) {
+    u16 value;
+    Read(&value, sizeof(value));
+    out_data = ntohs(value);
+    return *this;
+}
+
+Packet& Packet::operator>>(s32& out_data) {
+    s32 value;
+    Read(&value, sizeof(value));
+    out_data = ntohl(value);
+    return *this;
+}
+
+Packet& Packet::operator>>(u32& out_data) {
+    u32 value;
+    Read(&value, sizeof(value));
+    out_data = ntohl(value);
+    return *this;
+}
+
+Packet& Packet::operator>>(float& out_data) {
+    Read(&out_data, sizeof(out_data));
+    return *this;
+}
+
+Packet& Packet::operator>>(double& out_data) {
+    Read(&out_data, sizeof(out_data));
+    return *this;
+}
+
+Packet& Packet::operator>>(char* out_data) {
+    // First extract string length
+    u32 length = 0;
+    *this >> length;
+
+    if ((length > 0) && CheckSize(length)) {
+        // Then extract characters
+        std::memcpy(out_data, &data[read_pos], length);
+        out_data[length] = '\0';
+
+        // Update reading position
+        read_pos += length;
+    }
+
+    return *this;
+}
+
+Packet& Packet::operator>>(std::string& out_data) {
+    // First extract string length
+    u32 length = 0;
+    *this >> length;
+
+    out_data.clear();
+    if ((length > 0) && CheckSize(length)) {
+        // Then extract characters
+        out_data.assign(&data[read_pos], length);
+
+        // Update reading position
+        read_pos += length;
+    }
+
+    return *this;
+}
+
+Packet& Packet::operator<<(bool in_data) {
+    *this << static_cast<u8>(in_data);
+    return *this;
+}
+
+Packet& Packet::operator<<(s8 in_data) {
+    Append(&in_data, sizeof(in_data));
+    return *this;
+}
+
+Packet& Packet::operator<<(u8 in_data) {
+    Append(&in_data, sizeof(in_data));
+    return *this;
+}
+
+Packet& Packet::operator<<(s16 in_data) {
+    s16 toWrite = htons(in_data);
+    Append(&toWrite, sizeof(toWrite));
+    return *this;
+}
+
+Packet& Packet::operator<<(u16 in_data) {
+    u16 toWrite = htons(in_data);
+    Append(&toWrite, sizeof(toWrite));
+    return *this;
+}
+
+Packet& Packet::operator<<(s32 in_data) {
+    s32 toWrite = htonl(in_data);
+    Append(&toWrite, sizeof(toWrite));
+    return *this;
+}
+
+Packet& Packet::operator<<(u32 in_data) {
+    u32 toWrite = htonl(in_data);
+    Append(&toWrite, sizeof(toWrite));
+    return *this;
+}
+
+Packet& Packet::operator<<(float in_data) {
+    Append(&in_data, sizeof(in_data));
+    return *this;
+}
+
+Packet& Packet::operator<<(double in_data) {
+    Append(&in_data, sizeof(in_data));
+    return *this;
+}
+
+Packet& Packet::operator<<(const char* in_data) {
+    // First insert string length
+    u32 length = std::strlen(in_data);
+    *this << length;
+
+    // Then insert characters
+    Append(in_data, length * sizeof(char));
+
+    return *this;
+}
+
+Packet& Packet::operator<<(const std::string& in_data) {
+    // First insert string length
+    u32 length = static_cast<u32>(in_data.size());
+    *this << length;
+
+    // Then insert characters
+    if (length > 0)
+        Append(in_data.c_str(), length * sizeof(std::string::value_type));
+
+    return *this;
+}
+
+bool Packet::CheckSize(std::size_t size) {
+    is_valid = is_valid && (read_pos + size <= data.size());
+
+    return is_valid;
+}
+
+} // namespace Network
--- a/src/network/packet.cpp
+++ b/src/network/packet.cpp
@@ -13,10 +13,6 @@
 
 namespace Network {
 
-Packet::Packet() : read_pos(0), is_valid(true) {}
-
-Packet::~Packet() {}
-
 void Packet::Append(const void* in_data, std::size_t size_in_bytes) {
     if (in_data && (size_in_bytes > 0)) {
         std::size_t start = data.size();
@@ -39,7 +35,7 @@ void Packet::Clear() {
 }
 
 const void* Packet::GetData() const {
-    return !data.empty() ? &data[0] : NULL;
+    return !data.empty() ? &data[0] : nullptr;
 }
 
 void Packet::IgnoreBytes(u32 length) {
@@ -54,8 +50,8 @@ bool Packet::EndOfPacket() const {
     return read_pos >= data.size();
 }
 
-Packet::operator BoolType() const {
-    return is_valid ? &Packet::CheckSize : NULL;
+Packet::operator bool() const {
+    return is_valid ? &Packet::CheckSize : nullptr;
 }
 
 Packet& Packet::operator>>(bool& out_data) {
--- a/src/network/packet.h
+++ b/src/network/packet.h
@@ -115,6 +115,12 @@ private:
 
 template <typename T>
 Packet& Packet::operator>>(std::vector<T>& out_data) {
+    // First extract the size
+    u32 size = 0;
+    *this >> size;
+    out_data.resize(size);
+
+    // Then extract the data
     for (std::size_t i = 0; i < out_data.size(); ++i) {
         T character = 0;
         *this >> character;
@@ -135,6 +141,10 @@ Packet& Packet::operator>>(std::array<T, S>& out_data) {
 
 template <typename T>
 Packet& Packet::operator<<(const std::vector<T>& in_data) {
+    // First insert the size
+    *this << static_cast<u32>(in_data.size());
+
+    // Then insert the data
     for (std::size_t i = 0; i < in_data.size(); ++i) {
         *this << in_data[i];
     }
--- a/src/network/packet.h
+++ b/src/network/packet.h
@@ -1,0 +1,192 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <vector>
+#include "common/common_types.h"
+
+namespace Network {
+
+/// A class for serialize data for network transfer. It also handles endianess
+class Packet {
+    /// A bool-like type that cannot be converted to integer or pointer types
+    typedef bool (Packet::*BoolType)(std::size_t);
+
+public:
+    Packet();
+    ~Packet();
+
+    /**
+     * Append data to the end of the packet
+     * @param data        Pointer to the sequence of bytes to append
+     * @param size_in_bytes Number of bytes to append
+     */
+    void Append(const void* data, std::size_t size_in_bytes);
+
+    /**
+     * Reads data from the current read position of the packet
+     * @param out_data        Pointer where the data should get written to
+     * @param size_in_bytes Number of bytes to read
+     */
+    void Read(void* out_data, std::size_t size_in_bytes);
+
+    /**
+     * Clear the packet
+     * After calling Clear, the packet is empty.
+     */
+    void Clear();
+
+    /**
+     * Ignores bytes while reading
+     * @param length THe number of bytes to ignore
+     */
+    void IgnoreBytes(u32 length);
+
+    /**
+     * Get a pointer to the data contained in the packet
+     * @return Pointer to the data
+     */
+    const void* GetData() const;
+
+    /**
+     * This function returns the number of bytes pointed to by
+     * what getData returns.
+     * @return Data size, in bytes
+     */
+    std::size_t GetDataSize() const;
+
+    /**
+     * This function is useful to know if there is some data
+     * left to be read, without actually reading it.
+     * @return True if all data was read, false otherwise
+     */
+    bool EndOfPacket() const;
+    /**
+     * Test the validity of the packet, for reading
+     * This operator allows to test the packet as a boolean
+     * variable, to check if a reading operation was successful.
+     *
+     * A packet will be in an invalid state if it has no more
+     * data to read.
+     *
+     * This behaviour is the same as standard C++ streams.
+     *
+     * Usage example:
+     * @code
+     * float x;
+     * packet >> x;
+     * if (packet)
+     * {
+     *    // ok, x was extracted successfully
+     * }
+     *
+     * // -- or --
+     *
+     * float x;
+     * if (packet >> x)
+     * {
+     *    // ok, x was extracted successfully
+     * }
+     * @endcode
+     *
+     * Don't focus on the return type, it's equivalent to bool but
+     * it disallows unwanted implicit conversions to integer or
+     * pointer types.
+     *
+     * @return True if last data extraction from packet was successful
+     */
+    operator BoolType() const;
+
+    /// Overloads of operator >> to read data from the packet
+    Packet& operator>>(bool& out_data);
+    Packet& operator>>(s8& out_data);
+    Packet& operator>>(u8& out_data);
+    Packet& operator>>(s16& out_data);
+    Packet& operator>>(u16& out_data);
+    Packet& operator>>(s32& out_data);
+    Packet& operator>>(u32& out_data);
+    Packet& operator>>(float& out_data);
+    Packet& operator>>(double& out_data);
+    Packet& operator>>(char* out_data);
+    Packet& operator>>(std::string& out_data);
+    template <typename T>
+    Packet& operator>>(std::vector<T>& out_data);
+    template <typename T, std::size_t S>
+    Packet& operator>>(std::array<T, S>& out_data);
+
+    /// Overloads of operator << to write data into the packet
+    Packet& operator<<(bool in_data);
+    Packet& operator<<(s8 in_data);
+    Packet& operator<<(u8 in_data);
+    Packet& operator<<(s16 in_data);
+    Packet& operator<<(u16 in_data);
+    Packet& operator<<(s32 in_data);
+    Packet& operator<<(u32 in_data);
+    Packet& operator<<(float in_data);
+    Packet& operator<<(double in_data);
+    Packet& operator<<(const char* in_data);
+    Packet& operator<<(const std::string& in_data);
+    template <typename T>
+    Packet& operator<<(const std::vector<T>& in_data);
+    template <typename T, std::size_t S>
+    Packet& operator<<(const std::array<T, S>& data);
+
+private:
+    /// Disallow comparisons between packets
+    bool operator==(const Packet& right) const;
+    bool operator!=(const Packet& right) const;
+
+    /**
+     * Check if the packet can extract a given number of bytes
+     * This function updates accordingly the state of the packet.
+     * @param size Size to check
+     * @return True if size bytes can be read from the packet
+     */
+    bool CheckSize(std::size_t size);
+
+    // Member data
+    std::vector<char> data; ///< Data stored in the packet
+    std::size_t read_pos;   ///< Current reading position in the packet
+    bool is_valid;          ///< Reading state of the packet
+};
+
+template <typename T>
+Packet& Packet::operator>>(std::vector<T>& out_data) {
+    for (u32 i = 0; i < out_data.size(); ++i) {
+        T character = 0;
+        *this >> character;
+        out_data[i] = character;
+    }
+    return *this;
+}
+
+template <typename T, std::size_t S>
+Packet& Packet::operator>>(std::array<T, S>& out_data) {
+    for (u32 i = 0; i < out_data.size(); ++i) {
+        T character = 0;
+        *this >> character;
+        out_data[i] = character;
+    }
+    return *this;
+}
+
+template <typename T>
+Packet& Packet::operator<<(const std::vector<T>& in_data) {
+    for (u32 i = 0; i < in_data.size(); ++i) {
+        *this << in_data[i];
+    }
+    return *this;
+}
+
+template <typename T, std::size_t S>
+Packet& Packet::operator<<(const std::array<T, S>& in_data) {
+    for (u32 i = 0; i < in_data.size(); ++i) {
+        *this << in_data[i];
+    }
+    return *this;
+}
+
+} // namespace Network
--- a/src/network/packet.h
+++ b/src/network/packet.h
@@ -10,14 +10,11 @@
 
 namespace Network {
 
-/// A class for serialize data for network transfer. It also handles endianess
+/// A class that serializes data for network transfer. It also handles endianess
 class Packet {
-    /// A bool-like type that cannot be converted to integer or pointer types
-    typedef bool (Packet::*BoolType)(std::size_t);
-
 public:
-    Packet();
-    ~Packet();
+    Packet() = default;
+    ~Packet() = default;
 
     /**
      * Append data to the end of the packet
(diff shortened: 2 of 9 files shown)


wwylele:
(Just log here some conversation from discord)

> wwylele - Today at 4:49 PM
> B3N30 can those network stuff be unit-tested?
> B3N30 - Today at 4:49 PM
> not easy.
> you have to hard_code a connecting sequence and then manually check the status
> and obviously you have to run 2 citras at a time
> wwylele - Today at 4:51 PM
> ...like, these code currently is pretty much "dead code" that doesn't get run at all. It is hard to review it without ability to test it
> B3N30 - Today at 4:52 PM
> Yeah i know.  Problem is: It is easy to test with the ui
> But at the current stage the ui will just confuse the users
> But I'm open to suggestions on how to test it
> wwylele - Today at 4:56 PM
> I guess someone will be get crazy by your reinterpret_cast
> B3N30 - Today at 4:58 PM
> It could be solved with memcpy, too.  But i think that's exactly the way reinterpret_cast is meant to be used :stuck_out_tongue:(edited)
> wwylele - Today at 4:59 PM
> I don't think so...
> \> reinterpret_cast only guarantees that if you cast a pointer to a different type, and then reinterpret_cast it back to the original type, you get the original value.
> but your counterpart is memcpy in Append
> B3N30 - Today at 5:01 PM
> hmm... :confused:
> wwylele - Today at 5:08 PM
> also, could you template them?
> B3N30 - Today at 5:09 PM
> I couldn't come up with a good idea how to do it...
> Any suggestions?
> wwylele - Today at 5:10 PM
> for example, make a hton template and specialize it for some data type, then template <<(edited)
> also, just use u32 and so on
> B3N30 - Today at 5:12 PM
> damn it. I'mm not used to u32...
> wwylele - Today at 5:16 PM
> is there any existing Packet implementation we can use? like, in boost?
> B3N30 - Today at 5:17 PM
> I dont't know. This is just a adaptation of the packet class used in dolphin
> wwylele - Today at 5:17 PM
> oh it's from dolphin
> B3N30 - Today at 5:18 PM
> yupp. I used dolphin as my inspiration

src/network/room.h
```diff
@@ -19,6 +19,24 @@ struct RoomInformation {
     u32 member_slots; ///< Maximum number of members in this room
 };
 
+using MacAddress = std::array<uint8_t, 6>;
+/// A special MAC address that tells the room we're joining to assign us a MAC address
+/// automatically.
+const MacAddress NoPreferredMac = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
+
+// The different types of messages that can be sent. The first byte of each packet defines the type
+typedef uint8_t MessageID;
```
Subv: using MessageID = uint8_t

src/network/room_member.cpp
```diff
@@ -15,11 +19,218 @@ class RoomMember::RoomMemberImpl {
...
+                    break;
+                }
+                enet_packet_destroy(event.packet);
+            }
+        }
+    }
 };
 
+void RoomMember::RoomMemberImpl::StartLoop() {
+    receive_thread = std::make_unique<std::thread>(&RoomMember::RoomMemberImpl::ReceiveLoop, this);
```
Subv: don't you have to call receive_thread->Start() here?

src/network/room.cpp
```diff
@@ -12,12 +18,329 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+    do {
+        for (int i = 3; i < result_mac.size(); ++i) {
+            result_mac[i] = dis(random_gen);
+        }
+    } while (!IsValidMacAddress(result_mac));
+    return result_mac;
+}
+
+void Room::RoomImpl::HandleWifiPacket(const ENetEvent* event) {
+    for (auto it = members.begin(); it != members.end(); ++it) {
```
Subv: This function should only retransmit the wifi packet to the peer with mac_address == wifipacket.destination_address

src/network/room_member.h
```diff
@@ -4,14 +4,32 @@
...
+#include <vector>
 #include "common/common_types.h"
 #include "network/room.h"
 
 namespace Network {
 
+/// Information about the received WiFi packets.
+/// Acts as our own 802.11 header.
+struct WifiPacket {
+    enum class PacketType { Beacon, Data, Management };
```
Subv: Remove the 'Management' type, it'll be superceded by specific frame types

src/network/room_member.h
```diff
@@ -4,14 +4,32 @@
...
+    PacketType type;           ///< The type of 802.11 frame, Beacon / Data.
```
Subv: Remove the 'Beacon / Data' part

src/network/room_member.cpp
MerryMage: `std::thread` begins execution immediately after construction.

src/network/packet.h
```diff
@@ -0,0 +1,192 @@
...
+
+#pragma once
+
+#include <array>
+#include <vector>
+#include "common/common_types.h"
+
+namespace Network {
+
+/// A class for serialize data for network transfer. It also handles endianess
```
Subv: ```A class that serializes data for network transfer. It also handles endianess```

src/network/room.cpp
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+    }
+
+    ENetPacket* enet_packet =
+        enet_packet_create(packet.GetData(), packet.GetDataSize(), ENET_PACKET_FLAG_RELIABLE);
+    enet_host_broadcast(server, 0, enet_packet);
+    enet_host_flush(server);
+}
+
+MacAddress Room::RoomImpl::GenerateMacAddress() {
+    MacAddress result_mac = NintendoOUI;
```
Subv: Add a comment here about using the OUI as the first 3 bytes of the MAC

src/network/room.cpp
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+
+    if (destination_address == BroadcastMac) { // Send the data to everyone except the sender
+        for (auto it = members.begin(); it != members.end(); ++it) {
+            if (it->peer != event->peer)
+                enet_peer_send(it->peer, 0, event->packet);
+        }
+    } else { // Send the data only to the destination client
+        for (auto it = members.begin(); it != members.end(); ++it) {
+            if (it->mac_address == destination_address)
+                enet_peer_send(it->peer, 0, event->packet);
```
Subv: break out of the loop here

src/network/packet.cpp
```diff
@@ -0,0 +1,229 @@
...
+std::size_t Packet::GetDataSize() const {
+    return data.size();
+}
+
+bool Packet::EndOfPacket() const {
+    return read_pos >= data.size();
+}
+
+Packet::operator BoolType() const {
+    return is_valid ? &Packet::CheckSize : NULL;
```
lioncash: `nullptr`

src/network/packet.h
```diff
@@ -0,0 +1,192 @@
...
+     * }
+     * @endcode
+     *
+     * Don't focus on the return type, it's equivalent to bool but
+     * it disallows unwanted implicit conversions to integer or
+     * pointer types.
+     *
+     * @return True if last data extraction from packet was successful
+     */
+    operator BoolType() const;
```
lioncash: Is there any reason:

```cpp
explicit operator bool() const;
```

isn't just implemented instead?

src/network/room.cpp
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+}
+
+void Room::RoomImpl::HandleGameNamePacket(const ENetEvent* event) {
+    Packet in_packet;
+    in_packet.Append(event->packet->data, event->packet->dataLength);
+
+    in_packet.IgnoreBytes(sizeof(MessageID));
+    std::string game_name;
+    in_packet >> game_name;
+    for (auto it = members.begin(); it != members.end(); ++it) {
```
lioncash:
```cpp
for (auto& member : members) {
    if (member.peer == event->peer) 
        member.game_name = game_name;
}

```

src/network/room.cpp
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+    }
+
+    Packet out_packet;
+    out_packet << static_cast<MessageID>(IdChatMessage);
+    out_packet << sending_member->nickname;
+    out_packet << message;
+
+    ENetPacket* enet_packet = enet_packet_create(out_packet.GetData(), out_packet.GetDataSize(),
+                                                 ENET_PACKET_FLAG_RELIABLE);
+    for (auto it = members.begin(); it != members.end(); ++it) {
```
lioncash:
```cpp
for (const auto& member : members) {
    if (member.peer != event->peer)
        enet_peer_send(member.peer, 0, enet_packet);
}

File: src/network/packet.cpp [hunk 13]
```diff
@@ -0,0 +1,229 @@
...
+#else
+#include <arpa/inet.h>
+#endif
+#include <cstring>
+#include <string>
+#include "network/packet.h"
+
+namespace Network {
+
+Packet::Packet() : read_pos(0), is_valid(true) {}
```
lioncash: These can just be assigned in the class body.

i.e.
```cpp
class cls
{
private:
  int one = 0;
  bool thing = true;
};
```

The constructor can then just be:

```cpp
Packet::Packet() = default;
```

src/network/packet.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+Packet::~Packet() {}
```
lioncash:
```cpp
Packet::~Packet() = default;
```

src/network/packet.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+void Packet::Clear() {
+    data.clear();
+    read_pos = 0;
+    is_valid = true;
+}
+
+const void* Packet::GetData() const {
+    return !data.empty() ? &data[0] : NULL;
```
lioncash: `nullptr`

src/network/packet.h
```diff
@@ -0,0 +1,192 @@
...
+    Packet& operator<<(const char* in_data);
+    Packet& operator<<(const std::string& in_data);
+    template <typename T>
+    Packet& operator<<(const std::vector<T>& in_data);
+    template <typename T, std::size_t S>
+    Packet& operator<<(const std::array<T, S>& data);
+
+private:
+    /// Disallow comparisons between packets
+    bool operator==(const Packet& right) const;
```
lioncash: Comparison operators aren't generated by default, so these aren't needed.

src/network/packet.h
```diff
@@ -0,0 +1,192 @@
...
+
+    // Member data
+    std::vector<char> data; ///< Data stored in the packet
+    std::size_t read_pos;   ///< Current reading position in the packet
+    bool is_valid;          ///< Reading state of the packet
+};
+
+template <typename T>
+Packet& Packet::operator>>(std::vector<T>& out_data) {
+    for (u32 i = 0; i < out_data.size(); ++i) {
```
lioncash: `std::size_t i`

src/network/packet.h
```diff
@@ -0,0 +1,192 @@
...
+        T character = 0;
+        *this >> character;
+        out_data[i] = character;
+    }
+    return *this;
+}
+
+template <typename T, std::size_t S>
+Packet& Packet::operator>>(std::array<T, S>& out_data) {
+    for (u32 i = 0; i < out_data.size(); ++i) {
```
lioncash: `std::size_t i`

src/network/packet.h
```diff
@@ -0,0 +1,192 @@
...
+        T character = 0;
+        *this >> character;
+        out_data[i] = character;
+    }
+    return *this;
+}
+
+template <typename T>
+Packet& Packet::operator<<(const std::vector<T>& in_data) {
+    for (u32 i = 0; i < in_data.size(); ++i) {
```
lioncash: `std::size_t i`

src/network/packet.h
```diff
@@ -0,0 +1,192 @@
...
+        *this << in_data[i];
+    }
+    return *this;
+}
+
+template <typename T, std::size_t S>
+Packet& Packet::operator<<(const std::array<T, S>& in_data) {
+    for (u32 i = 0; i < in_data.size(); ++i) {
```
lioncash: `std::size_t i`

src/network/room.cpp
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+
+    // Notify everyone that the room information has changed.
+    BroadcastRoomInformation();
+    SendJoinSuccess(event->peer, preferred_mac);
+}
+
+bool Room::RoomImpl::IsValidNickname(const std::string& nickname) const {
+    // A nickname is valid if it is not already taken by anybody else in the room.
+    // TODO(B3N30): Check for empty names, spaces, etc.
+    for (const Member& member : members) {
```
lioncash:
```cpp
return std::all_of(members.begin(), members.end(),
                   [&nickname](const auto& member) { return member.nickname != nickname; });

File: src/network/room.cpp [hunk 22]
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+        if (member.nickname == nickname) {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool Room::RoomImpl::IsValidMacAddress(const MacAddress& address) const {
+    // A MAC address is valid if it is not already taken by anybody else in the room.
+    for (const Member& member : members) {
```
lioncash:
```cpp
return std::all_of(members.begin(), members.end(),
                   [&address](const auto& member) { return member.mac_address != address; });

File: src/network/room.cpp [hunk 23]
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+    in_packet.Append(event->packet->data, event->packet->dataLength);
+    in_packet.IgnoreBytes(sizeof(MessageID));
+    in_packet.IgnoreBytes(sizeof(u8));         // WifiPacket Type
+    in_packet.IgnoreBytes(sizeof(u8));         // WifiPacket Channel
+    in_packet.IgnoreBytes(sizeof(MacAddress)); // WifiPacket Transmitter Address
+    MacAddress destination_address;
+    in_packet >> destination_address;
+
+    if (destination_address == BroadcastMac) { // Send the data to everyone except the sender
+        for (auto it = members.begin(); it != members.end(); ++it) {
```
lioncash:
```cpp
for (const auto& member : members) {
    if (member.peer != event->peer)
        enet_peer_send(member.peer, 0, event->packet);
}

File: src/network/room.cpp (hunk 8 above)
Comment (lioncash) [inline src/network/room.cpp]: This can utilize `std::find_if`.

File: src/network/room.cpp [hunk 24]
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+        if (it->peer == event->peer)
+            it->game_name = game_name;
+    }
+    BroadcastRoomInformation();
+}
+
+void Room::RoomImpl::HandleClientDisconnection(ENetPeer* client) {
+    // Remove the client from the members list.
+    members.erase(std::remove_if(members.begin(), members.end(),
+                                 [&](const Member& member) { return member.peer == client; }),
```
lioncash: you can just capture the single variable instead of everything by reference here.

src/network/room.cpp
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+}
+
+void Room::RoomImpl::HandleChatPacket(const ENetEvent* event) {
+    Packet in_packet;
+    in_packet.Append(event->packet->data, event->packet->dataLength);
+
+    in_packet.IgnoreBytes(sizeof(MessageID));
+    std::string message;
+    in_packet >> message;
+    auto CompareNetworkAddress = [&](const Member member) -> bool {
```
lioncash: You can just capture `event` instead of allowing everything to be captured by reference.

src/network/room.cpp
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+        preferred_mac = GenerateMacAddress();
+    }
+
+    // At this point the client is ready to be added to the room.
+    Member member{};
+    member.mac_address = preferred_mac;
+    member.nickname = nickname;
+    member.peer = event->peer;
+
+    members.push_back(member);
```
lioncash:
```cpp
members.push_back(std::move(member));
```

src/network/room_member.h
```diff
@@ -41,6 +67,21 @@ class RoomMember final {
     State GetState() const;
 
     /**
+     * Returns information about the members in the room we're currently connected to.
+     */
+    const MemberList& GetMemberInformation() const;
+
+    /**
+     * Returns information about the members in the room we're currently connected to.
```
Subv: This comment is wrong

src/network/room.cpp
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+        ENetEvent event;
+        if (enet_host_service(server, &event, 1000) > 0) {
+            switch (event.type) {
+            case ENET_EVENT_TYPE_RECEIVE:
+                switch (event.packet->data[0]) {
+                case IdJoinRequest:
+                    HandleJoinRequest(&event);
+                    break;
+                case IdSetGameName:
+                    HandleGameNamePacket(&event);
```
Subv: missing a break here

src/network/room.cpp
```diff
@@ -12,12 +18,345 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+    in_packet.IgnoreBytes(sizeof(u8));         // WifiPacket Type
+    in_packet.IgnoreBytes(sizeof(u8));         // WifiPacket Channel
+    in_packet.IgnoreBytes(sizeof(MacAddress)); // WifiPacket Transmitter Address
+    MacAddress destination_address;
+    in_packet >> destination_address;
+
+    if (destination_address == BroadcastMac) { // Send the data to everyone except the sender
+        for (auto it = members.begin(); it != members.end(); ++it) {
+            if (it->peer != event->peer)
+                enet_peer_send(it->peer, 0, event->packet);
```
Subv: A new packet has to be created using enet_packet_create, otherwise you'll invoke use-after-free when the acknowledgement packet comes back.

src/network/room_member.cpp
```diff
@@ -44,31 +267,65 @@ void RoomMember::Join(const std::string& nick, const char* server_addr, u16 serv
...
 }
 
 void RoomMember::Leave() {
-    enet_peer_disconnect(room_member_impl->server, 0);
-    room_member_impl->state = State::Idle;
-    // TODO(B3N30): Close the receive thread
+    ASSERT_MSG(room_member_impl->receive_thread != nullptr, "Must be in a room to leave it.");
+    {
+        std::lock_guard<std::mutex> lock(room_member_impl->network_mutex);
+        enet_peer_disconnect(room_member_impl->server, 0);
```
Subv: For disconnection i recommend checking how Dolphin does it:
https://github.com/dolphin-emu/dolphin/blob/a3e9aba7b08b2ffef348cf2c9cc9e637ea4efa7f/Source/Core/Core/NetPlayClient.cpp#L593

They have a 3-second timeout on the disconnection acknowledgement, after which they just forcefully stop the connection.

src/network/room_member.cpp
```diff
@@ -15,11 +19,218 @@ class RoomMember::RoomMemberImpl {
...
+    return state == State::Joining || state == State::Joined;
+}
+
+void RoomMember::RoomMemberImpl::ReceiveLoop() {
+    // Receive packets while the connection is open
+    while (IsConnected()) {
+        std::lock_guard<std::mutex> lock(network_mutex);
+        ENetEvent event;
+        if (enet_host_service(client, &event, 1000) > 0) {
+            if (event.type == ENET_EVENT_TYPE_RECEIVE) {
```
Subv: You need to handle ENET_EVENT_TYPE_DISCONNECT so we can check for server timeouts.

src/network/room_member.cpp
```diff
@@ -15,11 +19,218 @@ class RoomMember::RoomMemberImpl {
...
+            }
+        }
+    }
 };
 
+void RoomMember::RoomMemberImpl::StartLoop() {
+    receive_thread = std::make_unique<std::thread>(&RoomMember::RoomMemberImpl::ReceiveLoop, this);
+}
+
+void RoomMember::RoomMemberImpl::Send(Packet& packet) {
```
Subv: Can we make it so this function doesn't block the caller thread until the packet is sent?

src/network/room_member.cpp
```diff
@@ -15,11 +19,218 @@ class RoomMember::RoomMemberImpl {
...
+    MemberList member_information;
+    /// Information about the room we're connected to.
+    RoomInformation room_information;
+
     std::atomic<State> state{State::Idle}; ///< Current state of the RoomMember.
+    void SetState(const State new_state);
+    bool IsConnected() const;
+
+    std::string nickname;   ///< The nickname of this member.
+    MacAddress mac_address; ///< The mac_address of this member.
```
Subv: Can you expose this via a GetMacAddress please?

src/network/room_member.h
```diff
@@ -4,14 +4,32 @@
...
+#include <vector>
 #include "common/common_types.h"
 #include "network/room.h"
 
 namespace Network {
 
+/// Information about the received WiFi packets.
+/// Acts as our own 802.11 header.
+struct WifiPacket {
+    enum class PacketType { Beacon, Data };
```
Subv: This should have Beacon, Data, Authentication, AssociationResponse

src/network/room_member.cpp
B3n30: It's possible but i didn't know it was necessary

src/network/packet.h
lioncash: still applies

src/network/room.cpp
lioncash: still applies

src/network/packet.h
B3n30: reason is:     
```
/// A bool-like type that cannot be converted to integer or pointer types
    typedef bool (Packet::*BoolType)(std::size_t);
```

src/network/room.cpp
B3n30: sorry. missed this one. Will be fixed in the next commit

src/network/packet.h
lioncash: That still doesn't explain the purpose of it though.

lioncash: This seems like the [safe bool idiom](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Safe_bool) which is made unnecessary from C++11 and onwards by explicit conversion operators.

B3n30: Yeah you are right, will be fixed in the next commit. Probably as soon as I talked with @Subv about the async send

src/network/room.cpp
```diff
@@ -12,20 +18,357 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+
+    Packet out_packet;
+    out_packet.Append(event->packet->data, event->packet->dataLength);
+    ENetPacket* enet_packet = enet_packet_create(out_packet.GetData(), out_packet.GetDataSize(),
+                                                 ENET_PACKET_FLAG_RELIABLE);
+
+    if (destination_address == BroadcastMac) { // Send the data to everyone except the sender
+        for (const auto& member : members) {
+            if (member.peer != event->peer)
+                enet_peer_send(member.peer, 0, event->packet);
```
Subv: Send the enet_packet instead of event->packet

src/network/room.cpp
```diff
@@ -12,20 +18,357 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+
+// Room
 Room::Room() : room_impl{std::make_unique<RoomImpl>()} {}
 
 Room::~Room() = default;
 
 void Room::Create(const std::string& name, const std::string& server_address, u16 server_port) {
     ENetAddress address;
     address.host = ENET_HOST_ANY;
-    enet_address_set_host(&address, server_address.c_str());
```
Subv: This should probably still happen but only if the name is not empty

src/network/room_member.cpp
```diff
@@ -15,11 +20,256 @@ class RoomMember::RoomMemberImpl {
...
+    packet >> chat_entry.message;
+    // TODO(B3N30): Invoke callbacks
+}
+
+void RoomMember::RoomMemberImpl::Disconnect() {
+    member_information.clear();
+    room_information.member_slots = 0;
+    room_information.name.clear();
+
+    if (server) {
```
Subv: if (!server) return; enet_peer_disconnect(server, 0);

src/network/room_member.cpp
```diff
@@ -15,11 +20,256 @@ class RoomMember::RoomMemberImpl {
...
+                case IdMacCollision:
+                    SetState(State::MacCollision);
+                    break;
+                default:
+                    break;
+                }
+                enet_packet_destroy(event.packet);
+                break;
+            case ENET_EVENT_TYPE_DISCONNECT:
+                SetState(State::LostConnection);
```
Subv: break

src/network/room_member.cpp
```diff
@@ -15,11 +20,254 @@ class RoomMember::RoomMemberImpl {
...
+    }
+    Disconnect();
+};
+
+void RoomMember::RoomMemberImpl::StartLoop() {
+    loop_thread = std::make_unique<std::thread>(&RoomMember::RoomMemberImpl::MemberLoop, this);
+}
+
+void RoomMember::RoomMemberImpl::Send(Packet& packet) {
+    send_list.push_back(std::move(packet));
```
Subv: You need to lock the mutex here

src/network/room.cpp
```diff
@@ -186,6 +194,10 @@ void Room::RoomImpl::HandleJoinRequest(const ENetEvent* event) {
         preferred_mac = GenerateMacAddress();
     }
 
+    if (client_version != network_version) {
+        SendVersionMissmatch(event->peer);
```
Subv: return

src/network/room.cpp
```diff
@@ -12,20 +18,405 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+     */
+    void SendNameCollision(ENetPeer* client);
+
+    /**
+     * Sends a ID_ROOM_MAC_COLLISION message telling the client that the MAC is invalid.
+     */
+    void SendMacCollision(ENetPeer* client);
+
+    /**
+     * Sends a ID_ROOM_VERSION_MISSMATCH message telling the client that the MAC is invalid.
```
Subv: MISMATCH, a single s

src/network/room.cpp
```diff
@@ -12,20 +18,405 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+     */
+    void SendVersionMissmatch(ENetPeer* client);
```
Subv: same here

src/network/room.cpp
```diff
@@ -12,20 +18,405 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+
+    ENetPacket* enet_packet =
+        enet_packet_create(packet.GetData(), packet.GetDataSize(), ENET_PACKET_FLAG_RELIABLE);
+    enet_peer_send(client, 0, enet_packet);
+    enet_host_flush(server);
+}
+
+void Room::RoomImpl::SendVersionMissmatch(ENetPeer* client) {
+    Packet packet;
+    packet << static_cast<MessageID>(IdVersionMissmatch);
```
Subv: IdVersionMismatch

src/network/room_member.cpp
```diff
@@ -34,41 +291,94 @@ RoomMember::State RoomMember::GetState() const {
...
+}
+
+const std::string& RoomMember::GetNickname() const {
+    return room_member_impl->nickname;
+}
+
+const MacAddress& RoomMember::GetMacAddress() const {
+    if (GetState() == State::Joined)
+        return room_member_impl->mac_address;
+    return MacAddress{};
```
Subv: warning: returning reference to temporary [-Wreturn-local-addr]

src/network/room.cpp
```diff
@@ -12,20 +18,405 @@ static constexpr u32 MaxConcurrentConnections = 10;
...
+     * to all other clients.
+     */
+    void HandleClientDisconnection(ENetPeer* client);
 };
 
+// RoomImpl
+void Room::RoomImpl::ServerLoop() {
+    while (state != State::Closed) {
+        ENetEvent event;
+        if (enet_host_service(server, &event, 1000) > 0) {
```
Subv: Since we're now sending packets in this thread, we need to reduce that third parameter.
Right now this function will wait up to 1 second before returning, delaying the sending of any queued packets.
Dolphin uses [250ms](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/NetPlayClient.cpp#L640) for this value, maybe we can do the same?

Subv: Scratch that, I've been testing various values and found 100ms to be a good timeout for chat purposes on a loopback connection

src/network/room.h
```diff
@@ -19,6 +21,29 @@ struct RoomInformation {
...
+/// A special MAC address that tells the room we're joining to assign us a MAC address
+/// automatically.
+const MacAddress NoPreferredMac = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
+
+// 802.11 broadcast MAC address
+constexpr MacAddress BroadcastMac = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
+
+// The different types of messages that can be sent. The first byte of each packet defines the type
+using MessageID = u8;
+enum RoomMessageTypes {
```
wwylele: `enum RoomMessageTypes: u8 {`. Also I think that `using MessageID = u8` is useless and cause more confusion: when I read the code `packet << static_cast<MessageID>(something)` I wondered for a while if the `Packet` class support custom type, only to find it is actually `u8`

src/network/room_member.cpp
```diff
@@ -15,11 +20,263 @@ class RoomMember::RoomMemberImpl {
...
+    std::list<Packet> send_list; ///< A list that stores all packets to send the async
+    void MemberLoop();
+    void StartLoop();
+
+    /**
+     * Sends data to the room. It will be send on channel 0 with flag RELIABLE
+     * @param packet The data to send
+     */
+    void Send(Packet& packet);
+    /**
```
wwylele: nit: new line between functions. Same for below

src/network/room_member.cpp
```diff
@@ -15,11 +20,263 @@ class RoomMember::RoomMemberImpl {
...
+    Disconnect();
+};
+
+void RoomMember::RoomMemberImpl::StartLoop() {
+    loop_thread = std::make_unique<std::thread>(&RoomMember::RoomMemberImpl::MemberLoop, this);
+}
+
+void RoomMember::RoomMemberImpl::Send(Packet& packet) {
+    std::lock_guard<std::mutex> lock(send_list_mutex);
+    send_list.push_back(std::move(packet));
```
wwylele: since you are moving the packet in the implementation, the parameter should be also marked as rvalue reference (`Packet&& packet`).

src/network/room_member.cpp
```diff
@@ -15,11 +20,263 @@ class RoomMember::RoomMemberImpl {
...
+}
+
+void RoomMember::RoomMemberImpl::HandleJoinPacket(const ENetEvent* event) {
+    Packet packet;
+    packet.Append(event->packet->data, event->packet->dataLength);
+
+    // Ignore the first byte, which is the message id.
+    packet.IgnoreBytes(sizeof(MessageID));
+
+    // Parse the MAC Address from the BitStream
```
wwylele: `BitStream`
if it is not referring to some class/object in code, just `bit stream` or `bitstream`. Or do you actually mean `Packet`?
Same for below

src/network/room_member.h
```diff
@@ -4,14 +4,32 @@
...
+#include <vector>
 #include "common/common_types.h"
 #include "network/room.h"
 
 namespace Network {
 
+/// Information about the received WiFi packets.
+/// Acts as our own 802.11 header.
+struct WifiPacket {
+    enum class PacketType { Beacon, Data, Authentication, AssociationResponse };
```
wwylele: `enum class PacketType:u8{...`

src/network/room_member.cpp
```diff
@@ -34,41 +291,93 @@ RoomMember::State RoomMember::GetState() const {
...
+RoomInformation RoomMember::GetRoomInformation() const {
+    return room_member_impl->room_information;
+}
+
 void RoomMember::Join(const std::string& nick, const char* server_addr, u16 server_port,
-                      u16 client_port) {
+                      u16 client_port, const MacAddress& preferred_mac) {
+    // If the member is connected, kill the connection first
+    if (room_member_impl->loop_thread && room_member_impl->loop_thread->joinable()) {
+        room_member_impl->SetState(State::Error);
```
wwylele: Since this set it to an error state, should it return here?

B3n30: No, it sets the error state just to get the loop_thread to finish, thus it could get joined and reset. Another way would be to just ASSERT here, since this shouldn't happen. It would mean you try to connect while you are still in a room. Depending on how the ui is designed, this could still happen. The current test ui has it that way.

B3n30:
Did the following test a the current stage with a test ui:

- Creating and joining the room
- Sending some chat messages
- Leaving a room
- Reconnecting
- Closing the room while another member is joined
- Recreating a room
- Try to join a room with the same nickname as another connected member
- Joining the room

All tests were done on OSX with two Citras at the same machine using loopback.

PR #2819, opened 2017-07-10
Telemetry: Submit logged data to the Citra service

bunnei:
This is the second part of a set of changes to add a telemetry framework to Citra. In the near term, this will be used for user-submitted compatibility data to our own custom service endpoint (services.citra-emu.org). My goal, however, is to make this very generic, and allow us to easily log useful/interesting data fields throughout the emulator, and then upload them to our service at a later time. In the long term, we may use this for anonymous telemetry, critical errors and crash reports, measuring how often features are used, etc.

With this change, I've added support for submitting logged telemetry to the service. I've introduced two new dependencies for doing this:
* [CPR](https://github.com/whoshuu/cpr) - A modern C++ wrapper around libcurl. Besides a nicer interface, the primary reason for using this (instead of curl directly) is that it works seamlessly with our build system. This is used for authenticated web requests to the Citra service.
* [JSON](https://github.com/nlohmann/json) - Just a very simple header-only JSON serialization library. Nothing particularly special here, except that it worked nicely for our implementation.

- [x] Implement a backend that does an authenticated log submission to our service
- [ ] Implement remaining fields for game compatibility submission

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2819)


src/web_service/web_backend.h
```diff
@@ -0,0 +1,40 @@
...
+     * @returns Token as a string, empty if not set.
+     */
+    static const std::string& GetToken();
+
+    /**
+     * Posts JSON to services.citra-emu.org.
+     * @param url URL of the services.citra-emu.org endpoint to post data to.
+     * @param data String of JSON data to use for the body of the POST request.
+     */
+    void PostJson(const std::string& url, const std::string& data);
```
bunnei: I suppose this class can just go away and/or this can be a static function

src/web_service/web_backend.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    if (GetUsername().empty() || GetToken().empty()) {
+        LOG_ERROR(WebService, "Environment variables %s and %s must be set to POST JSON",
+                  ENV_VAR_USERNAME, ENV_VAR_TOKEN);
+        return;
+    }
+
+    auto result = cpr::Post(cpr::Url{url}, cpr::Body{data},
+                            cpr::Header{{"Content-Type", "application/json"},
+                                        {"x-username", GetUsername()},
```
Subv: [RFC 6648](https://tools.ietf.org/html/rfc6648) recomments not using "x-" as a prefix for nonstandard HTTP headers. I don't particularly care, but I thought I'd let you know

src/web_service/web_backend.cpp
```diff
@@ -0,0 +1,59 @@
...
+                                        {"x-token", GetToken()},
+                                        {"api-version", API_VERSION}});
```
j-selby: Pedantic:  For consistencies sake, they either should all be prefixed with `x-`, or preferably, no prefix at all (RFC6648). We might also want to make this part of the URL scheme, potentially (`https://services.citra-emu.org/testcases/v1/`).

src/web_service/web_backend.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    if (result.error.code != cpr::ErrorCode::OK) {
+        LOG_ERROR(WebService, "POST JSON failed, error code: %d, text: %s", result.error.code,
+                  result.text.c_str());
```
j-selby: No way of returning success/failure to a frontend? Is this out of scope, or is it not needed?

bunnei: Out of scope for this change - this is still all "backend work" - frontend stuff to follow

src/web_service/telemetry_json.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <array>
+#include <string>
+#include <json.hpp>
+#include "common/telemetry.h"
+
+namespace WebService {
+
+/**
+ * Empty implementation of VisitorInterface that drops all fields. Used when a functional
```
Subv: This comment seems to be wrong

src/web_service/web_backend.cpp
```diff
@@ -0,0 +1,59 @@
...
+    static const std::string username{GetEnvironmentVariable(ENV_VAR_USERNAME)};
+    return username;
+}
+
+const std::string& GetToken() {
+    static const std::string token{GetEnvironmentVariable(ENV_VAR_TOKEN)};
+    return token;
+}
+
+void PostJson(const std::string& url, const std::string& data) {
```
Subv: Does this happen on a separate thread?

bunnei: Will change to be asynchronous

src/web_service/telemetry_json.h
```diff
@@ -12,8 +12,8 @@
 namespace WebService {
 
 /**
- * Empty implementation of VisitorInterface that drops all fields. Used when a functional
- * backend implementation is not available.
+ * Implementation of VisitorInterface that serialized telemetry into JSON, and submits it to the
```
Subv: ```serializes telemetry```

bunnei:
Pinging reviewers... Would like to get this merged this week. Thanks.

src/web_service/telemetry_json.h
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Implementation of VisitorInterface that serialized telemetry into JSON, and submits it to the
+ * Citra web service
+ */
+class TelemetryJson : public Telemetry::VisitorInterface {
+public:
+    TelemetryJson();
+    ~TelemetryJson() = default;
+
+    void Visit(const Telemetry::Field<bool>& field) override;
```
Subv: Can't you have a single template Visit function and use explicit template specialization to avoid repeating the implementation so many times in the source file? like so http://ideone.com/ho1Q77

src/web_service/telemetry_json.cpp
```diff
@@ -0,0 +1,89 @@
...
+// Refer to the license.txt file included.
+
+#include "common/assert.h"
+#include "core/settings.h"
+#include "web_service/telemetry_json.h"
+#include "web_service/web_backend.h"
+
+namespace WebService {
+
+TelemetryJson::TelemetryJson() {}
```
Subv: This constructor is empty, make it default in the header file

src/web_service/telemetry_json.h
wwylele: I guess the fact that they are virtual function override makes them not able to be templated.

bunnei: This ^

bunnei:
Last call for reviews...

PR #2823, opened 2017-07-13
telemetry: Log performance, configuration, and system data.

bunnei:
This is the third part of a set of changes to add a telemetry framework to Citra. In the near term, this will be used for user-submitted compatibility data to our own custom service endpoint (services.citra-emu.org). My goal, however, is to make this very generic, and allow us to easily log useful/interesting data fields throughout the emulator, and then upload them to our service at a later time. In the long term, we may use this for anonymous telemetry, critical errors and crash reports, measuring how often features are used, etc.

With this change, I've added additional telemetry logging throughout the emulator. These include:
* User configuration settings of interest (just things that affect how games run / performance, if I'm missing anything of interest - feel free to let me know)
* User system information (just CPU and GPU for now)
* Simple performance metrics (just emulation speed and frame rate when shutdown is initiated)

- [x] [#2819] Implement a backend that does an authenticated log submission to our service
- [x] Implement remaining fields for game compatibility submission

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2823)

### DIFF
--- a/src/core/telemetry_session.cpp
+++ b/src/core/telemetry_session.cpp
@@ -4,7 +4,10 @@
 
 #include <cstring>
 
+#include "common/assert.h"
 #include "common/scm_rev.h"
+#include "common/x64/cpu_detect.h"
+#include "core/settings.h"
 #include "core/telemetry_session.h"
 
 #ifdef ENABLE_WEB_SERVICE
@@ -13,6 +16,18 @@
 
 namespace Core {
 
+static const char* CpuVendorToStr(Common::CPUVendor vendor) {
+    switch (vendor) {
+    case Common::CPUVendor::INTEL:
+        return "Intel";
+    case Common::CPUVendor::AMD:
+        return "Amd";
+    case Common::CPUVendor::OTHER:
+        return "Other";
+    }
+    UNREACHABLE();
+}
+
 TelemetrySession::TelemetrySession() {
 #ifdef ENABLE_WEB_SERVICE
     backend = std::make_unique<WebService::TelemetryJson>();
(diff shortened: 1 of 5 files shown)


src/core/telemetry_session.cpp
```diff
@@ -22,20 +37,59 @@ TelemetrySession::TelemetrySession() {
...
+    AddField(Telemetry::FieldType::App, "Git_Branch", Common::g_scm_branch);
+    AddField(Telemetry::FieldType::App, "Git_Revision", Common::g_scm_rev);
+
+    // Log user system information
+    AddField(Telemetry::FieldType::UserSystem, "CPU_Model", Common::GetCPUCaps().cpu_string);
+    AddField(Telemetry::FieldType::UserSystem, "CPU_BrandString",
+             Common::GetCPUCaps().brand_string);
+    AddField(Telemetry::FieldType::UserSystem, "CPU_Vendor",
+             CpuVendorToStr(Common::GetCPUCaps().vendor));
+    AddField(Telemetry::FieldType::UserSystem, "CPU_Extention_x64_AES", Common::GetCPUCaps().aes);
```
j-selby: `Extention`?

bunnei: derp

src/video_core/renderer_opengl/pica_to_gl.h
```diff
@@ -72,9 +73,9 @@ inline GLenum WrapMode(Pica::TexturingRegs::TextureConfig::WrapMode mode) {
     }
 
     if (static_cast<u32>(mode) > 3) {
-        // It is still unclear whether mode 4-7 are valid, so log it if a game uses them.
-        // TODO(wwylele): telemetry should be added here so we can collect more info about which
-        // game uses this.
+        Core::Telemetry().AddField(Telemetry::FieldType::Session,
```
Subv: Is this call thread safe?

bunnei: Nope.. but right now, we're only calling this from the emu thread

bunnei:
Ping for additional reviews on this.

PR #2828, opened 2017-07-14
citra-qt: add compatibility to the game_list

Kloen:
Add the Game Compatibility information to the citra-qt game list.

The game list it's downloaded by CMake, that way each nightly version will have the latest version of it ;)

Mandatory Screenshot:
image

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2828)

### DIFF
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -2,9 +2,12 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <cinttypes>
 #include <QApplication>
 #include <QFileInfo>
 #include <QHeaderView>
+#include <QJsonDocument>
+#include <QJsonObject>
 #include <QKeyEvent>
 #include <QMenu>
 #include <QThreadPool>
@@ -147,6 +150,33 @@ bool GameList::containsAllWords(QString haystack, QString userinput) {
                        [haystack](QString s) { return haystack.contains(s); });
 }
 
+QString GameList::GetReadableCompatibility(QString compat) {
+    int c = compat.toInt();
+    switch (c) {
+    case 0:
+        return QString("Perfect");
+        break;
+    case 1:
+        return QString("Great");
+        break;
+    case 2:
+        return QString("Okay");
+        break;
+    case 3:
+        return QString("Bad");
+        break;
+    case 4:
+        return QString("Intro/Menu");
+        break;
+    case 5:
+        return QString("Won't Boot");
+        break;
+    default:
+        return QString("Not Tested");
+        break;
+    }
+}
+
 // Event in order to filter the gamelist after editing the searchfield
 void GameList::onTextChanged(const QString& newText) {
     int rowCount = tree_view->model()->rowCount();
@@ -217,6 +247,7 @@ GameList::GameList(GMainWindow* parent) : QWidget{parent} {
 
     item_model->insertColumns(0, COLUMN_COUNT);
     item_model->setHeaderData(COLUMN_NAME, Qt::Horizontal, "Name");
+    item_model->setHeaderData(COLUMN_COMPATIBILITY, Qt::Horizontal, "Compatibility");
     item_model->setHeaderData(COLUMN_FILE_TYPE, Qt::Horizontal, "File type");
     item_model->setHeaderData(COLUMN_SIZE, Qt::Horizontal, "Size");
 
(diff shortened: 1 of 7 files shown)


mailwl:
good. don't want to waste time for great comp. games

jroweboy:
I love the idea, but I do have a few suggestions/concerns

* This doesn't play well with changes that fix the issues with the game. Say for instance you have a pull request that fixes the issue that just got merged and put into this build. The compatibility won't reflect this since it would not have been updated until after the PR is merged and build.
* The previous issue is compounded in bleeding edge since it will use the latest nightly compat, but many fixes end up in bleeding edge weeks before they end up in nightly.
* Words need to be translated but stars conveys the meaning without words :)

mailwl:
yes, need to add date, something like "last checked 100 days ago"

chris062689:
I disagree with star ratings, they don't convey anything. Our compatibility ratings were carefully chosen to convey (with their description) to show the state of the game.

B3n30:
Since web-services in citra are a thing now, wouldn't it be better to just check on a daily( or whatever) interval, if the compatibility list has changed and update it directly in citra?

chris062689:
The problem I see with that is if they don't update, they won't see valid results if they're running an old version.

bunnei:
Would like to see the little colored square in there. A little color makes the UI look nicer.
Also, I'm not convinced that having these statically compiled into the emulator is best. Would probably be better to download/cache them somewhere on first run of the emulator, and periodically grab updates. I'm sure @chris062689 would be happy to expose these as JSON in a new endpoint :)

chris062689:
The new endpoint exists here: https://citra-emu.org/game/index.json

Kloen:
IMHO, by now, we should statically build it into the emulator. The points that made me think that statically compile it into the build was better are:

* You **need** to use a single compatibility list, that it's more or less updated to the time your build was made. Otherwise it may lead to points where, if I use a build that's 3 months old, and a compatibility list that's 1 day old, I'm likely to get false information.

* We could request it once a day to the server, or store it on a path, and let Citra update it whenever you want, but that would make a little and **unnecesary** load on the servers, unnecesary? See the point above.

So, imo, there is no such reason to make citra update it, due to the false information it may deliver sometimes. At least, I don't want to see "Your compatibility list says that Pokemon X/Y works, but isn't working on my 3 months old Citra version" posts.

Now, about the current information displayed, well, I think, as well as you do, that it can be improved. Color boxes are a thing, I'll add them this weekend. As well as a tooltip to define each compatibility type based on (https://github.com/citra-emu/citra-web/blob/master/site/data/compatibility.json).

In addition, we could also display information like "Last version tested" or "Last date tested", but that isn't on the current exposed endpoint I'm using to get the list, so we may add that information, or put it somewhere else. In my opinion, it would be a good addition that may prevent "You say this game don't work, but it actually does!!!11" posts.

Actually, since this part isn't going to work on Bleeding Edge as pointed out before, we could just... don't know... disable it for BE builds? Since well, there is no such point of have a compatibility list with BE to begin with, because, you know, it's BE, a lot of strange things happen there.

bunnei:
@Kloen to your first point above, both nightlies and BE will eventually be updated to a new installer system that provides automatic updates. No one should be using Citra with old builds. Furthermore, this will eliminate scenarios that @jroweboy described above, where for example a major issue is fixed, but the compatibility DB lags behind a little bit (which will almost always happen), or simply new games are tested. Furthermore, we will soon be getting very rich information on game usage/compatibility. We will be able to do things like only pull compatibility data for versions up to the users current version of Citra.

I realize that all of this is a little ways out, so I'm fine with this as a step in that direction - but just giving you an idea of what I would like to see happen in the future.

Kloen:
@bunnei yes, definetly. Most of the reasons that stand behind this implementation will change in the future, short term, or long term, that's a must, but at the time being, this is imo the best choice, and can be easily changed into a "load from X file" implementation later on (Remove the CMake stuff, and make qt load from file, not a big deal for a D day patch). That way it will work for the current use case we have, and can be implemented the way the future use case, with updaters and stuff, will work.

So, if we are fine with that point, I'll probably add the colored boxes and the tooltip with the description and see if we can merge this before all that happens.

src/citra_qt/game_list.cpp
```diff
@@ -147,6 +150,33 @@ bool GameList::containsAllWords(QString haystack, QString userinput) {
                        [haystack](QString s) { return haystack.contains(s); });
 }
 
+QString GameList::GetReadableCompatibility(QString compat) {
+    int c = compat.toInt();
+    switch (c) {
+    case 0:
+        return QString("Perfect");
```
Megamouse: I'd recommend the use of tr("") here and on the other instances

bunnei:
@Kloen do you plan on wrapping this one up?

Kloen:
@bunnei yup, sure, I plan to wrap everything up this weekend, I couldn't yet due to IRL and work stuff.

jroweboy:
@chris062689 star ratings and word ratings don't have to be mutually exclusive. you can have both. for example, a start rating can have the word for its rating below it, or as part of a label for people to see when they hover the mouse.

> I disagree with star ratings, they don't convey anything. Our compatibility ratings were carefully chosen to convey (with their description) to show the state of the game.

in my opinion, the star ratings show exactly the same information that a word does. "Perfect" doesn't add much more than "5 stars" as people will launch the game and say "OMG HOW IS IT PERFECT WHEN ITS SO SLOW," just as "Bad" doesn't add anything that "2 stars" doesn't convey. The full description gives more meaning of course, but lets not say that stars don't convey anything. As I already mentioned, I think its fine to include the word rating below the star rating if you really want the word rating.

chris062689:
I just feel that we should be consistent in the way we display it across the website and UI. 

_Sent from my OnePlus ONEPLUS A3000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

chris062689:
I just feel that we should be consistent in the way we display it across the website and UI.

_Sent from my OnePlus ONEPLUS A3000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

chris062689:
I just feel that we should be consistent in the way we display it across the website and UI.

_Sent from my OnePlus ONEPLUS A3000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

chris062689:
@Kloen Has there been an update to this?

_Sent from my OnePlus ONEPLUS A3000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

B3n30:
Closed, due to inactivity.

PR #2829, opened 2017-07-16
CMakeLists: Check that all submodules are present

MerryMage:
Does this by looking in `.gitmodules` and grepping the paths to all the submodules out of it, and checking for the presence of the `.git` folder in each of those paths.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2829)


jroweboy:
@wwylele SEND_ERROR will stop the build file from getting generated, but it will continue doing configuration. I think thats good enough since when they send us the cmake log it'll have "Please run command: " in it

PR #2832, opened 2017-07-18
Disable web services by default

j-selby:
Telemetry reporting should not be enabled by default. Branches in development or unofficial builds should not be submitting telemetry data by default unless it is explicitly required, as this data could potentially be mixed with data from valid, prebuilt nightly distributions.

This also enables web services for Travis and AppVeyor builds by default.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2832)


B3n30:
After that PR gets merged, we should change the wiki pages for building thus users building citra will enable web_services.

chris062689:
We probably need separate flags for telemetry vs netplay.

yuriks:
:lgtm:, but maybe should be reviewed by someone more familiar with this subsystem.

---

Reviewed 1 of 1 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2832#-:-KqQwJlBMg2z-RQC-7Eb:bpngtip)*

yuriks:
Also, needs a rebase.

j-selby:
Rebased.

yuriks:
Reviewed 1 of 3 files at r3, 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2832)*

chris062689:
@B3n30 isn't the point that anyone building Citra shouldn't build with web_services, as we don't want them submitting telemetry? 

_Sent from my OnePlus ONEPLUS A3000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

B3n30:
I'm still not sure if disabling web_services is the right way to disable telemetry. Currently it's only telemetry using web_service, but we will add other stuff that needs web_services. I think it's better to have a flag that gets checked when sending telemetry, thus this can be disabled by default and travis/appveyor will enable it by default. The same flag is probably used for settings

chris062689:
@bunnei What is your opinion on this related to telemetry? 

Should we enable it by default (anyone that builds Citra will have it enabled) or only on our official builds? 

We already have it setup where you can easily filter out custom builds from the rest of the data.

j-selby:
Closing. Not willing to maintain this.

PR #2833, opened 2017-07-18
Don't pull in entire JSON repo for single header file

j-selby:
It is nonsensical to pull in a 100mb+ repository, with tests, documentation and so on for a single C++ header. This PR removes the link to the upstream repository and substitutes it for the aforementioned header.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2833)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,6 +31,3 @@
 [submodule "cpr"]
     path = externals/cpr
     url = https://github.com/whoshuu/cpr.git
-[submodule "json"]
-    path = externals/json
-    url = https://github.com/nlohmann/json.git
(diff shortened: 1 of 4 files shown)


wwylele:
Would be better if you add a README in that directory saying the source of the repo and the version

j-selby:
@wwylele We aren't doing this for getopt and only a URL for microprofile. Should that happen here?

wwylele:
@j-selby IMO they should also have, but they are out of the scope of this PR so you don't need to mind it :p

j-selby:
Fixed.

bunnei:
With fast internet and lots of storage space, I'm personally indifferent to this (and generally prefer submoduling the source repo wherever possible instead of copying things into our repository)... but I know others care about size.

PR #2838, opened 2017-07-20
Added missing parts in libnetwork

B3n30:
This PR is the third regarding the ENet inclusion.
With this PR the libnetwork is fully functional. There might still be some features that could be added, though.

The features added in this PR:
- Set the GameInformation of the RoomMember the moment they change
- The callback interface for RoomMember
here's an example of how to hook a function to one of the events:
``` // Callback identifier for the OnStateChanged event.
static Network::RoomMember::Connection<Network::RoomMember::State> state_changed;

void OnStateChanged(const Network::RoomMember::State) {
 ...
}

if (auto room_member = Network::GetRoomMember().lock())
        wifi_packet_received = room_member->ConnectOnWifiPacketReceived(OnWifiPacketReceived);

...
//Every connected function must be disconnected
if (auto room_member = Network::GetRoomMember().lock())
        room_member->Disconnect(state_changed);
```
To use this with class members:
``` 
room_member->ConnectOnStateChanged(std::bind(&SomeClass::OnStateChanged, 
                                             this, std::placeholders::_1));
```

- The GetMemberList function for Room, which is needed for the announce web_service

Things to come, regarding this topic:

- [ ] UI
- [ ] Announce web_service
- [ ] Integrating libnetwork into uds

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2838)

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -388,7 +388,7 @@ set(HEADERS
 
 create_directory_groups(${SRCS} ${HEADERS})
 add_library(core STATIC ${SRCS} ${HEADERS})
-target_link_libraries(core PUBLIC common PRIVATE audio_core video_core)
+target_link_libraries(core PUBLIC common PRIVATE audio_core network video_core)
 target_link_libraries(core PUBLIC Boost::boost PRIVATE cryptopp dynarmic fmt)
 if (ENABLE_WEB_SERVICE)
     target_link_libraries(core PUBLIC json-headers web_service)
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -19,6 +19,7 @@
 #include "core/loader/loader.h"
 #include "core/memory_setup.h"
 #include "core/settings.h"
+#include "network/network.h"
 #include "video_core/video_core.h"
 
 namespace Core {
@@ -188,6 +189,10 @@ void System::Shutdown() {
     cpu_core = nullptr;
     app_loader = nullptr;
     telemetry_session = nullptr;
+    if (auto room_member = Network::GetRoomMember().lock()) {
+        Network::GameInfo game_info{};
+        room_member->SendGameInfo(game_info);
+    }
 
     LOG_DEBUG(Core, "Shutdown OK");
 }
--- a/src/core/loader/ncch.cpp
+++ b/src/core/loader/ncch.cpp
@@ -358,9 +358,10 @@ ResultStatus AppLoader_NCCH::Load() {
         SMDH smdh;
         std::memcpy(&smdh, smdh_data.data(), sizeof(SMDH));
         auto title = smdh.GetShortTitle(SMDH::TitleLanguage::English);
-        game_info.name.assign(Common::UTF16ToUTF8(std::u16string{title.begin(), title.end()}));
+        auto result = std::find_if(title.begin(), title.end(),
+                                   [](const auto& text) -> bool { return text == '\u0000'; });
+        game_info.name = Common::UTF16ToUTF8(std::u16string{title.begin(), result});
         game_info.id = ncch_header.program_id;
-        game_info.version = ncch_header.version;
         room_member->SendGameInfo(game_info);
     }
 
--- a/src/core/loader/ncch.cpp
+++ b/src/core/loader/ncch.cpp
@@ -358,9 +358,8 @@ ResultStatus AppLoader_NCCH::Load() {
         SMDH smdh;
         std::memcpy(&smdh, smdh_data.data(), sizeof(SMDH));
         auto title = smdh.GetShortTitle(SMDH::TitleLanguage::English);
-        auto result = std::find_if(title.begin(), title.end(),
-                                   [](const auto& text) -> bool { return text == '\u0000'; });
-        game_info.name = Common::UTF16ToUTF8(std::u16string{title.begin(), result});
+        auto title_end = std::find(title.begin(), title.end(), u'\0');
+        game_info.name = Common::UTF16ToUTF8(std::u16string{title.begin(), title_end});
         game_info.id = ncch_header.program_id;
         room_member->SendGameInfo(game_info);
     }
--- a/src/core/loader/ncch.cpp
+++ b/src/core/loader/ncch.cpp
@@ -353,13 +353,7 @@ ResultStatus AppLoader_NCCH::Load() {
 
     if (auto room_member = Network::GetRoomMember().lock()) {
         Network::GameInfo game_info;
-        std::vector<u8> smdh_data;
-        ReadIcon(smdh_data);
-        SMDH smdh;
-        std::memcpy(&smdh, smdh_data.data(), sizeof(SMDH));
-        auto title = smdh.GetShortTitle(SMDH::TitleLanguage::English);
-        auto title_end = std::find(title.begin(), title.end(), u'\0');
-        game_info.name = Common::UTF16ToUTF8(std::u16string{title.begin(), title_end});
+        ReadTitle(game_info.name);
         game_info.id = ncch_header.program_id;
         room_member->SendGameInfo(game_info);
     }
--- a/src/core/loader/ncch.cpp
+++ b/src/core/loader/ncch.cpp
@@ -20,6 +20,7 @@
 #include "core/loader/ncch.h"
 #include "core/loader/smdh.h"
 #include "core/memory.h"
+#include "network/network.h"
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Loader namespace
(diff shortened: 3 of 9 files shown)


wwylele:
Reviewed 9 of 9 files at r1, 2 of 2 files at r2.
Review status: 7 of 9 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/core/loader/ncch.cpp, line 359 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqCb2ZqnHuWCSRWKC64:-KqCb2ZrwHEfG3f9F-y5:bup8bdm) ([raw file](https://github.com/citra-emu/citra/blob/5daf44c6775e8c926ce08b7b9cd5ffe5179f34bc/src/core/loader/ncch.cpp#L359)):*
> ```cpp
>         std::memcpy(&smdh, smdh_data.data(), sizeof(SMDH));
>         auto title = smdh.GetShortTitle(SMDH::TitleLanguage::English);
>         game_info.name.assign(Common::UTF16ToUTF8(std::u16string{title.begin(), title.end()}));
> ```

`game_info.name = Common::UTF16ToUTF8(std::u16string{title.data()});`
`GetShortTitle` returns a fix-size array which includes all zero padding, so you may assume it is null-terminated and parse it as c-style string (your current code will make all paddings included in the string). And just use a plain `=` instead of `assign` to reduce confusion.

---

*[src/core/loader/ncch.cpp, line 361 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqC_M_rotCZMem6ZdF3:-KqC_M_rotCZMem6ZdF4:b-chftfm) ([raw file](https://github.com/citra-emu/citra/blob/5daf44c6775e8c926ce08b7b9cd5ffe5179f34bc/src/core/loader/ncch.cpp#L361)):*
> ```cpp
>         game_info.name.assign(Common::UTF16ToUTF8(std::u16string{title.begin(), title.end()}));
>         game_info.id = ncch_header.program_id;
>         game_info.version = ncch_header.version;
> ```

I remember we had mentioned this on discord before: `ncch_header.version` is not the game version. It is the version of the ncch container format and always has a value of 0, 1 or 2. I am not sure which field in the header is the actual game version. Actually I doubt if there is such reliable field

---

*[src/network/room.cpp, line 477 at r3](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqCkUQPHXDMCkoyIVOI:-KqCkUQPHXDMCkoyIVOJ:bdlw0g0) ([raw file](https://github.com/citra-emu/citra/blob/e8e2205e4fe0aec92803d269d6d84d410cd88929/src/network/room.cpp#L477)):*
> ```cpp
>         member_list.push_back(member);
>     }
>     return std::move(member_list);
> ```

`return member_list;`as the return type should be changed to a plain value type, and copy elision is guaranteed here.

---

*[src/network/room.h, line 83 at r3](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqCj_v8o4zaLZpVYoj_:-KqCj_v9gETjzFR5vkds:bh7jr9y) ([raw file](https://github.com/citra-emu/citra/blob/e8e2205e4fe0aec92803d269d6d84d410cd88929/src/network/room.h#L83)):*
> ```objc
>      * Gets a list of the mbmers connected to the room.
>      */
>     const std::vector<Member>&& GetRoomMemberList() const;
> ```

returning a const rvalue reference makes little sense to me. Probably you just want to return by value.

---

*[src/network/room_member.cpp, line 334 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqChfA27A0wpIsNxLnn:-KqChfA27A0wpIsNxLno:b-ampa58) ([raw file](https://github.com/citra-emu/citra/blob/23c37b4a61817fbf9491c43873d79de78ad17fda/src/network/room_member.cpp#L334)):*
> ```cpp
> 
> template <typename T>
> void RoomMember::RoomMemberImpl::Invoke(const T& data) {
> ```

Consider the following situation:
 - `RoomMember` enters `Invoke`. It gets the mutex, copies the current callback list, and releases the mutex.
 - Now another thread calls `Disconnect` to remove a callback from the list that `Invoke` just copied, and destroys all the related resources.
 - Then `Invoke` calls the already-disconnected callback, which access destroyed resources. bang

---

*[src/network/room_member.h, line 65 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqCf1V3z9jvfY1rZIyg:-KqCf1V4lRpRiTOC1dZ2:bg33yv9) ([raw file](https://github.com/citra-emu/citra/blob/23c37b4a61817fbf9491c43873d79de78ad17fda/src/network/room_member.h#L65)):*
> ```objc
>     // The handle for the callback functions
>     template <typename T>
>     using Connection = std::shared_ptr<std::function<void(const T&)>>;
> ```

I am not a fan of name `Connection`. It is really easy to get confused with network connection. Just call it `CallbackHandle` as comment suggested. Same as `Disconnect` (I really thought it was for disconnecting from the network when seeing the code draft...)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2838)*

B3n30:
Review status: 4 of 9 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/core/loader/ncch.cpp, line 359 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqCb2ZqnHuWCSRWKC64:-KqDaEknVBV2bneuMBy3:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/5daf44c6775e8c926ce08b7b9cd5ffe5179f34bc/src/core/loader/ncch.cpp#L359)):*

`game_info.name = Common::UTF16ToUTF8(std::u16string{title.data()});`
`GetShortTitle` returns a fix-size array which includes all zero padding, so you may assume it is null-terminated and parse it as c-style string (your current code will make all paddings included in the string). And just use a plain `=` instead of `assign` to reduce confusion.
</blockquote></details>

Done.

---

*[src/core/loader/ncch.cpp, line 361 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqC_M_rotCZMem6ZdF3:-KqDa1wtT1724vlQ04ET:b5r0up8) ([raw file](https://github.com/citra-emu/citra/blob/5daf44c6775e8c926ce08b7b9cd5ffe5179f34bc/src/core/loader/ncch.cpp#L361)):*

I remember we had mentioned this on discord before: `ncch_header.version` is not the game version. It is the version of the ncch container format and always has a value of 0, 1 or 2. I am not sure which field in the header is the actual game version. Actually I doubt if there is such reliable field
</blockquote></details>

Removed the game version, since there is no way to get it

---

*[src/network/room.cpp, line 477 at r3](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqCkUQPHXDMCkoyIVOI:-KqDaFEn8mfEYMkhd340:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/e8e2205e4fe0aec92803d269d6d84d410cd88929/src/network/room.cpp#L477)):*

`return member_list;`as the return type should be changed to a plain value type, and copy elision is guaranteed here.
</blockquote></details>

Done.

---

*[src/network/room.h, line 83 at r3](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqCj_v8o4zaLZpVYoj_:-KqDaGwznj7QX0yfvEDQ:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/e8e2205e4fe0aec92803d269d6d84d410cd88929/src/network/room.h#L83)):*

returning a const rvalue reference makes little sense to me. Probably you just want to return by value.
</blockquote></details>

Done.

---

*[src/network/room_member.cpp, line 334 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqChfA27A0wpIsNxLnn:-KqDaI_QuQ4L4Zthl40Y:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/23c37b4a61817fbf9491c43873d79de78ad17fda/src/network/room_member.cpp#L334)):*

Consider the following situation:
 - `RoomMember` enters `Invoke`. It gets the mutex, copies the current callback list, and releases the mutex.
 - Now another thread calls `Disconnect` to remove a callback from the list that `Invoke` just copied, and destroys all the related resources.
 - Then `Invoke` calls the already-disconnected callback, which access destroyed resources. bang
</blockquote></details>

Done.

---

*[src/network/room_member.h, line 65 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2838#-KqCf1V3z9jvfY1rZIyg:-KqDaJuw1JYRI_BLfzVa:b-setuk1) ([raw file](https://github.com/citra-emu/citra/blob/23c37b4a61817fbf9491c43873d79de78ad17fda/src/network/room_member.h#L65)):*

I am not a fan of name `Connection`. It is really easy to get confused with network connection. Just call it `CallbackHandle` as comment suggested. Same as `Disconnect` (I really thought it was for disconnecting from the network when seeing the code draft...)
</blockquote></details>

Done. Called them similar to c++: Bind() and Unbind()

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2838)*

PR #2839, opened 2017-07-21
Kernel/HLE: Use a mutex  to synchronize access to the HLE kernel state between the cpu thread and any other possible threads that might touch the kernel (network thread, etc).

Subv:
This mutex is acquired in SVC::CallSVC, ie, as soon as the guest application enters the HLE kernel, and on non-fast paths of the memory access routines. It should be acquired by the aforementioned threads before modifying kernel structures.

```
<@yuriks> "slap a lock on that sucker" < you shouldn't need anything more than this
```
~yuriks, 2017

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2839)

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -60,6 +60,7 @@ set(SRCS
             hle/kernel/timer.cpp
             hle/kernel/vm_manager.cpp
             hle/kernel/wait_object.cpp
+            hle/lock.cpp
             hle/romfs.cpp
             hle/service/ac/ac.cpp
             hle/service/ac/ac_i.cpp
@@ -258,6 +259,7 @@ set(HEADERS
             hle/kernel/timer.h
             hle/kernel/vm_manager.h
             hle/kernel/wait_object.h
+            hle/lock.h
             hle/result.h
             hle/romfs.h
             hle/service/ac/ac.h
--- a/src/core/hle/kernel/kernel.h
+++ b/src/core/hle/kernel/kernel.h
@@ -129,4 +129,4 @@ void Init(u32 system_mode);
 /// Shutdown the kernel
 void Shutdown();
 
-} // namespace
+} // namespace Kernel
--- a/src/core/hle/lock.cpp
+++ b/src/core/hle/lock.cpp
@@ -1,0 +1,11 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <core/hle/lock.h>
+
+namespace HLE {
+std::mutex g_hle_lock;
+}
--- a/src/core/hle/lock.h
+++ b/src/core/hle/lock.h
@@ -1,0 +1,18 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <mutex>
+
+namespace HLE {
+/*
+ * Synchronizes access to the internal HLE kernel structures, it is acquired when a guest
+ * application thread performs a syscall. It should be acquired by any host threads that read or
+ * modify the HLE kernel state. Note: Any operation that directly or indirectly reads from or writes
+ * to the emulated memory is not protected by this mutex, and should be avoided in any threads other
+ * than the CPU thread.
+ */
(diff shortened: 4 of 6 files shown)


wwylele:
log from discord:

> btw is your mutex pr for uds?
> Subv - Today at 5:14 PM
> it's related, but that's not its only usecase
> the idea is to be able to manipulate HLE state from any thread
> (this includes the gui thread)
> combined with https://github.com/Subv/citra/commit/25ddfd9c42b68c42150db0a4de2b37abcee6a3c6 , it'll allow us to pause threads in HLE using a KEvent in a similar fashion to WaitSynch, and then signal the KEvent when we want to awake the thread
> also, jroweboy there goes the swkbd roadblock :smiley:
> jroweboy - Today at 5:35 PM
> @Subv while you're at it, mind implementing the sleep stuff? kthxbai
> Subv - Today at 5:35 PM
> what sleep stuff?
> ...
> wwylele / 白疾風 - Today at 5:44 PM
> @Subv Did you consider other options? such as other threads scheduling an operation into CoreTiming?
> Subv - Today at 5:44 PM
> that's what i'm currently using
> however that leads to very delayed events, since events scheduled from other threads have to wait until the end of the timeslice to be moved to the emu thread
> and using CoreTiming means you have to register events beforehand
> the global lock could also be useful if we ever want to emulate the syscorein another thread
> wwylele / 白疾風 - Today at 5:51 PM
> ...ok. I just have a feeling that not all problems are going to be resolved just by a mutex, but I can't tell
> I probably want to see a complete usecase
> also, does this affect performance? Because every memory Read/Write do mutex operation (which can do OS calls)
> Subv - Today at 5:57 PM
> what
> no, Read/Writes are explicitly not protected by this mutex
> (they should be protected by an actual LDREX/STREX/CLREX implementation)
> wwylele / 白疾風 - Today at 5:58 PM
> oh sorry github didn't expand the code and I misread it
> however that leads to very delayed events, since events scheduled from other threads have to wait until the end of the timeslice to be moved to the emu thread
> But even if the other thread does some operation immediately, say, signaling a kernel event, it still needs to wait for the end of the timeslice to take effect, where the waiting thread can be rescheduled in.
> Subv - Today at 6:01 PM
> remember that we don't execute events on every kernel call
> we wait until the entire timeslice is over, which might have consisted of one or many kernel calls
> anyways, the lock is mostly for convenience, i'm not expecting it to have any real effect on performance
> locking and doing what i have to do is infinitely more convenient than registering an event and scheduling it
> wwylele / 白疾風 - Today at 6:11 PM
> ok I am not against it. I just can't decide before seeing a real usecase, and I have to save my brain power before the exam :p

yuriks:
:lgtm: With great power comes great responsibility...

---

Reviewed 6 of 6 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/core/hle/kernel/kernel.cpp, line 19 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2839#-KqQxh1sERK8zUJV2RNY:-KqQxh1sERK8zUJV2RNZ:bwjmin2) ([raw file](https://github.com/citra-emu/citra/blob/f0e9e4fdb01ac5cc11070c5c2b3cb5f71dfad789/src/core/hle/kernel/kernel.cpp#L19)):*
> ```cpp
> unsigned int Object::next_object_id;
> 
> std::mutex g_hle_lock;
> ```

This shouldn't be defined here anymore.

---

*[src/core/hle/kernel/kernel.h, line 8 at r2](https://reviewable.io:443/reviews/citra-emu/citra/2839#-KqQxpV7wGnlyOHhF42Z:-KqQxpV7wGnlyOHhF42_:b9kbl80) ([raw file](https://github.com/citra-emu/citra/blob/99e1d4837e778ac32b9f8090c08f8facfa3272b2/src/core/hle/kernel/kernel.h#L8)):*
> ```objc
> 
> #include <cstddef>
> #include <mutex>
> ```

This file can also be reverted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2839#-:-KqQy-7a0cLy1yehN3Af:bkbq1b7)*

jroweboy:
Last call for reviews. Please leave a comment if you would like a chance to review it later, otherwise it will be merged in 24 hours.

PR #2842, opened 2017-07-22
Kernel/Memory: Give each process its own page table and allow switching the current page table upon reschedule

Subv:
This PR also implements a proper Memory::GetPhysicalPointer that doesn't depend on the current process' memory map.

This is just one step towards the ultimate goal of being able to load multiple processes simultaneously in citra.

This PR has a performance regression regarding dynarmic and fast page-table access. It should be reimplemented once dynarmic supports runtime pagetable switches.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2842)

### DIFF
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -178,8 +178,18 @@ static void SwitchContext(Thread* new_thread) {
 
         Core::CPU().LoadContext(new_thread->context);
         Core::CPU().SetCP15Register(CP15_THREAD_URO, new_thread->GetTLSAddress());
+
+        if (!previous_thread || previous_thread->owner_process != current_thread->owner_process) {
+            Kernel::g_current_process = current_thread->owner_process;
+            Memory::current_page_table = &Kernel::g_current_process->vm_manager.page_table;
+            // We have switched processes and thus, page tables, clear the instruction cache so we
+            // don't keep stale data from the previous process.
+            Core::CPU().ClearInstructionCache();
+        }
     } else {
         current_thread = nullptr;
+        // Note: We do not reset the current process and current page table when idling because
+        // technically we haven't changed processes, our threads are just paused.
     }
 }
 
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -171,6 +171,8 @@ static void SwitchContext(Thread* new_thread) {
         // Cancel any outstanding wakeup events for this thread
         CoreTiming::UnscheduleEvent(ThreadWakeupEventType, new_thread->callback_handle);
 
+        auto previous_process = Kernel::g_current_process;
+
         current_thread = new_thread;
 
         ready_queue.remove(new_thread->current_priority, new_thread);
@@ -179,7 +181,7 @@ static void SwitchContext(Thread* new_thread) {
         Core::CPU().LoadContext(new_thread->context);
         Core::CPU().SetCP15Register(CP15_THREAD_URO, new_thread->GetTLSAddress());
 
-        if (!previous_thread || previous_thread->owner_process != current_thread->owner_process) {
+        if (previous_process != current_thread->owner_process) {
             Kernel::g_current_process = current_thread->owner_process;
             Memory::current_page_table = &Kernel::g_current_process->vm_manager.page_table;
             // We have switched processes and thus, page tables, clear the instruction cache so we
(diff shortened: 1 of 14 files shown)


Subv:
I am not happy with how the rasterizer cache depends on there being an active page table, but I am not knowledgeable enough about it to change it. I am concerned about resource duplication/leakage once we start loading multiple processes.

src/core/hle/kernel/thread.cpp
```diff
@@ -178,8 +178,18 @@ static void SwitchContext(Thread* new_thread) {
 
         Core::CPU().LoadContext(new_thread->context);
         Core::CPU().SetCP15Register(CP15_THREAD_URO, new_thread->GetTLSAddress());
+
+        if (!previous_thread || previous_thread->owner_process != current_thread->owner_process) {
+            Kernel::g_current_process = current_thread->owner_process;
+            Memory::current_page_table = &Kernel::g_current_process->vm_manager.page_table;
+            // We have switched processes and thus, page tables, clear the instruction cache so we
+            // don't keep stale data from the previous process.
+            Core::CPU().ClearInstructionCache();
```
Subv: TODO: Don't clear the cache if coming from an idle state on the same process.

jroweboy:
With this PR in bleeding edge, performance plummeted. We'll need someway to be smarter about clearing the instruction cache :)

purpasmart96:
Nice, I have been experimenting with something like this for a while, although never managed to fully materialize in a working state.

src/core/hle/kernel/thread.cpp
Subv: Done

wwylele:
Could you also reimplement `IsValidPhysicalAddress`?

wwylele:
Please also fix the memory function usage in unit test introduced by the ARM intepreter test PR

B3n30:
:lgtm:

---

Reviewed 7 of 9 files at r1, 2 of 3 files at r3, 1 of 2 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2842#-:-Kqr3c5AqD7QK3Fwa5AV:bnfp4nl)*

src/core/memory.cpp
```diff
@@ -266,8 +208,23 @@ bool IsValidVirtualAddress(const VAddr vaddr) {
...
 
 bool IsValidPhysicalAddress(const PAddr paddr) {
-    boost::optional<VAddr> vaddr = PhysicalToVirtualAddress(paddr);
-    return vaddr && IsValidVirtualAddress(*vaddr);
+    struct PhysicalRegion {
+        PAddr start;
+        PAddr end;
+    };
+
+    static constexpr PhysicalRegion physical_regions[] = {
```
wwylele: can this array be merged with the one in `GetPhysicalPointer`? or can this function be just implemented as `return GetPhysicalPointer(paddr) != nullptr`?

B3n30:
Reviewed 1 of 7 files at r9, 1 of 3 files at r11, 1 of 2 files at r13, 1 of 1 files at r14, 1 of 1 files at r15, 1 of 1 files at r16, 1 of 1 files at r17, 1 of 1 files at r18.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2842)*

src/core/memory.cpp
Subv: Yes it can, I've updated the PR.

wwylele:
Bug report from bleeding edge test:
If any games are boot successively with CPU JIT on, the second one will crash/hang. The crash happens inside JIT call stack and the error is pretty random. This doesn't happen if one or both games are boot with the CPU interpreter.

src/core/hle/kernel/thread.cpp
MerryMage: Future implementations of this should likely have a `Core::CPU().ProcessSwap()` or `Core::CPU().PageTableSwap()` call here instead.

MerryMage:
https://github.com/citra-emu/citra/blob/a630fa287e178695b16887a56f7726ff97c91108/src/core/arm/dynarmic/arm_dynarmic.cpp#L59

Changing this line to `user_callbacks.page_table = nullptr;` should fix that.

yuriks:
dibs

yuriks:
Undibs actually. I'm not gonna have time to review this in the near future.

wwylele:
Since the performance regression is expected (and I did get some report), are we going to merge this anyway, or wait for some better solution?

Subv:
What should we do with this?

B3n30:
LGTM?

purpasmart96:
Awesome

PR #2844, opened 2017-07-24
Use WinSSPI instead of OpenSSL

jroweboy:
Nightly on Windows is currently bringing up error messages about missing SSLEAY.dll and LIBEAY.dll because Appveyor has OpenSSL installed, so cURL will attempt to use that. We would rather have it use the native windows ssl (SSPI) instead. Simply turning off "Use OpenSSL" makes everything work again.

Fixes nightly missing DLLs

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2844)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,7 @@ install:
 before_build:
   - mkdir build
   - cd build
-  - cmake -G "Visual Studio 15 2017 Win64" -DCITRA_USE_BUNDLED_QT=1 -DCITRA_USE_BUNDLED_SDL2=1 ..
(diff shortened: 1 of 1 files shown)


j-selby:
Now builds + runs fine (separated from source) on MSVC (Community 2017, Windows 10.14393) - libcurl now links against `crypt32.dll` fine.

externals/CMakeLists.txt
```diff
@@ -57,6 +57,11 @@ if (ENABLE_WEB_SERVICE)
     # CPR
     option(BUILD_TESTING OFF)
     option(BUILD_CPR_TESTS OFF)
+    # Appveyor has OpenSSL installed, so cURL will attempt to use that.
```
wwylele: If you comment it like this, this configuration should go into Appveyor script. If you want to put it in CMakeLists, you need to explain here why it is better to default it to SSPI on Windows in general instead.

PR #2855, opened 2017-08-02
Telemetry: Add several more useful fields

bunnei:
Adds a few more fields to telemetry that we'd like before going live:
- ProgramName (title string stored in games)
- BuildDate (of the application)
- RequiresSharedFont (if a game tries to use the shared font)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2855)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,11 +295,18 @@ function(create_directory_groups)
     endforeach()
 endfunction()
 
-# generate git revision information
+# Gets a UTC timstamp and sets the provided variable to it
+function(get_timestamp _var)
+    string(TIMESTAMP timestamp UTC)
+    set(${_var} "${timestamp}" PARENT_SCOPE)
+endfunction()
+
+# generate git/build information
 include(GetGitRevisionDescription)
 get_git_head_revision(GIT_REF_SPEC GIT_REV)
 git_describe(GIT_DESC --always --long --dirty)
 git_branch_name(GIT_BRANCH)
+get_timestamp(BUILD_DATE)
 
 enable_testing()
 add_subdirectory(externals)
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -75,6 +75,10 @@ void Initialize(Service::Interface* self) {
 void GetSharedFont(Service::Interface* self) {
     IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x44, 0, 0); // 0x00440000
     IPC::RequestBuilder rb = rp.MakeBuilder(2, 2);
+
+    // Log in telemetry if the game uses the shared font
+    Core::Telemetry().AddField(Telemetry::FieldType::Session, "RequiresSharedFont", true);
+
     if (!shared_font_loaded) {
         LOG_ERROR(Service_APT, "shared font file missing - go dump it from your 3ds");
         rb.Push<u32>(-1); // TODO: Find the right error code
--- a/src/core/loader/ncch.cpp
+++ b/src/core/loader/ncch.cpp
@@ -4,7 +4,9 @@
 
 #include <algorithm>
 #include <cinttypes>
+#include <codecvt>
 #include <cstring>
+#include <locale>
 #include <memory>
 #include "common/logging/log.h"
 #include "common/string_util.h"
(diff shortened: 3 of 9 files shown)


chris062689:
LGTM

CMakeLists.txt
```diff
@@ -295,11 +295,17 @@ function(create_directory_groups)
     endforeach()
 endfunction()
 
-# generate git revision information
+function(get_build_time _var)
```
j-selby: `_var` could be better named - is this the target variable?

src/core/hle/service/apt/apt.cpp
```diff
@@ -75,6 +75,7 @@ void Initialize(Service::Interface* self) {
 void GetSharedFont(Service::Interface* self) {
     IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x44, 0, 0); // 0x00440000
     IPC::RequestBuilder rb = rp.MakeBuilder(2, 2);
+    Core::Telemetry().AddField(Telemetry::FieldType::Session, "RequiresSharedFont", true);
```
j-selby: This could be broken apart a bit.

bunnei: Hmm? You mean add some spacing?

j-selby: Yeah - just nitpicking - makes it seem as if it is part of the emulation process itself.

CMakeLists.txt
```diff
@@ -295,11 +295,18 @@ function(create_directory_groups)
     endforeach()
 endfunction()
 
-# generate git revision information
+# Gets a UTC timstamp and sets the provided variable to it
```
jroweboy: This should be moved into a CMakeModule file... but since its common that we *don't* follow that practice (and even have two locations for cmake functions???) .... i'll let it slide :p I'll just open an issue for it instead.

src/core/loader/ncch.cpp
```diff
@@ -420,4 +422,18 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_
...
 }
 
+ResultStatus AppLoader_NCCH::ReadTitle(std::string& title) {
+    std::vector<u8> icon_data;
+    Loader::SMDH smdh;
+    ReadIcon(icon_data);
+    memcpy(&smdh, icon_data.data(), sizeof(Loader::SMDH));
+    const auto& short_title{smdh.GetShortTitle(Loader::SMDH::TitleLanguage::English)};
+
+    using convert_type = std::codecvt_utf8<wchar_t>;
```
jroweboy: `string_util.h` has a `UTF16ToUTF8` that should be used here instead. codecvt doesn't work on mingw and we have that solved with `UTF16ToUTF8` already (see https://github.com/citra-emu/citra/blob/master/CMakeLists.txt#L270)

also this is what the qt file list does (just with the qt specific utf16 to utf8 function)  https://github.com/citra-emu/citra/blob/master/src/citra_qt/game_list_p.h#L50

src/core/loader/ncch.h
```diff
@@ -191,23 +191,13 @@ class AppLoader_NCCH final : public AppLoader {
 
     ResultStatus ReadLogo(std::vector<u8>& buffer) override;
 
-    /**
-     * Get the program id of the application
-     * @param out_program_id Reference to store program id into
-     * @return ResultStatus result of function
-     */
     ResultStatus ReadProgramId(u64& out_program_id) override;
```
jroweboy: why are we removing these doc strings? even if its really obvious, its not bad to leave them imo

src/core/loader/ncch.cpp
```diff
@@ -420,4 +422,18 @@ ResultStatus AppLoader_NCCH::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_
     return ResultStatus::ErrorNotUsed;
 }
 
+ResultStatus AppLoader_NCCH::ReadTitle(std::string& title) {
+    std::vector<u8> icon_data;
+    Loader::SMDH smdh;
+    ReadIcon(icon_data);
+    memcpy(&smdh, icon_data.data(), sizeof(Loader::SMDH));
```
jroweboy: consider checking that this is a valid smdh (it might not be valid in the case of homebrew) 

https://github.com/citra-emu/citra/blob/a7ddec88d452cc5a9c8af1adf837d5466db793d1/src/citra_qt/game_list_p.h#L80

src/core/loader/ncch.cpp
jroweboy: here is how @B3n30 does it in the latest pr as well 

https://github.com/citra-emu/citra/pull/2838/files#diff-97ddda51413fc4c2c643428d97920640R358 (which seems to be more of what i was expecting here)

bunnei: Will fix - We should also just have this in one place (here)

src/core/loader/ncch.h
bunnei: They are copy+pasted from the base class. We don't duplicate docstrings in this scenario

PR #2856, opened 2017-08-02
pica: upload shared shader code & swizzle to both unit

wwylele:
One step toward GS. This is the most tricky finding about GS until now, so I want to put it here for discussion first.

I didn't hwtest the exact behaviour of the register `gs_unit_exclusive_configuration` (0x0244, `GPUREG_VSH_COM_MODE` on [3dbrew](https://www.3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_VSH_COM_MODE)), but as the theory suggested

> This register sets whether to use the geometry shader configuration or reuse the vertex shader configuration for the geometry shader shading unit. 

This means everything uploaded to VS regs should also be uploaded to the special GS unit.
A real use case of this, _Professor Layton and the Miracle Mask_, is found in recent test of GS experimental implementation. A simplified PICA command list the game sends is [here](https://pastebin.com/i0cvfhPC). As can be seen, the game sends both VS & GS program & swizzle via VS regs. Only part of GS program, whose offset excesses the maximum VS program size, is sent via GS regs (which also shows that, even in shared configuration mode, GS regs can still upload code to GS unit)

However, the game doesn't set uniforms in this way, so due to lack of evidence, I leave TODOs for uniform regs. But anyway, even if they behave the same way as program & swizzle, it is much less likely that a game will use it in this way, as the uniform is hard to share the same space between VS and GS.

Questions
 - Is there anyway to easily test this? ctrulib is managing this register depending on shader type, so I can't make program in a standard way to test this.
 - Should I use the register name "vs(h)_com_mode" instead? or some other better name?

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2856)

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -119,27 +119,6 @@ static void WriteUniformFloatReg(ShaderRegs& config, Shader::ShaderSetup& setup,
     }
 }
 
-static void WriteProgramCode(ShaderRegs& config, Shader::ShaderSetup& setup,
-                             unsigned max_program_code_length, u32 value) {
-    if (config.program.offset >= max_program_code_length) {
-        LOG_ERROR(HW_GPU, "Invalid %s program offset %d", GetShaderSetupTypeName(setup),
-                  (int)config.program.offset);
-    } else {
-        setup.program_code[config.program.offset] = value;
-        config.program.offset++;
-    }
-}
-
-static void WriteSwizzlePatterns(ShaderRegs& config, Shader::ShaderSetup& setup, u32 value) {
-    if (config.swizzle_patterns.offset >= setup.swizzle_data.size()) {
-        LOG_ERROR(HW_GPU, "Invalid %s swizzle pattern offset %d", GetShaderSetupTypeName(setup),
-                  (int)config.swizzle_patterns.offset);
-    } else {
-        setup.swizzle_data[config.swizzle_patterns.offset] = value;
-        config.swizzle_patterns.offset++;
-    }
-}
-
 static void WritePicaReg(u32 id, u32 value, u32 mask) {
     auto& regs = g_state.regs;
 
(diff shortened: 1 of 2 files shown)


B3n30:
LGTM

yuriks:
:lgtm: I don't like the structure of the code that much (the functions end up feeling kinda pointless now...) but it's good enough.

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/video_core/command_processor.cpp, line 122 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2856#-KqtX29HXAOxKS6edjC_:-KqtX29HXAOxKS6edjCa:b7k6n0c) ([raw file](https://github.com/citra-emu/citra/blob/c0852314e85fe89930645237b3acc6c833e172c3/src/video_core/command_processor.cpp#L122)):*
> const ShaderRegs& config

The only thing you use from this is the program offset, so I think you should just pass that directly. (This is particularly very confusing when you write to both units at once, where you pass the vs config, which ends up looking like a bug.)

---

*[src/video_core/command_processor.cpp, line 132 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2856#-KqtXCi7cQVBoz2czlnI:-KqtXCi7cQVBoz2czlnJ:b6i2k83) ([raw file](https://github.com/citra-emu/citra/blob/c0852314e85fe89930645237b3acc6c833e172c3/src/video_core/command_processor.cpp#L132)):*
> const ShaderRegs& config

Same, only the offset is used.

---

*[src/video_core/regs_pipeline.h, line 210 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2856#-KqtYQa6EaqoyZbiDGzh:-KqtYQa7AhUFFuL5YTck:b4a53am) ([raw file](https://github.com/citra-emu/citra/blob/c0852314e85fe89930645237b3acc6c833e172c3/src/video_core/regs_pipeline.h#L210)):*
> ```objc
>     // configuration depending on this register. If this is not set, the unit will share the
>     // configuration in regs.vs with other units; otherwise it will use the exclusive configuration
>     // in regs.gs
> ```

Doesn't this comment contradict the implementation? Your implementation implies that only *uploading* is shared, not that flipping this one will automatically make the GS unit immediately assume all values of the GS unit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2856#-:-KqtYk9cByg-g5M706qZ:b-92dwo0)*

wwylele:
I don't like the functions either. It looks even more useless after changing the parameter to offset. So eventually I break it up

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/video_core/regs_pipeline.h, line 210 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2856#-KqtYQa6EaqoyZbiDGzh:-Kqv_XlW_VzBSqI4dAOh:b-imech2) ([raw file](https://github.com/citra-emu/citra/blob/c0852314e85fe89930645237b3acc6c833e172c3/src/video_core/regs_pipeline.h#L210)):*

Doesn't this comment contradict the implementation? Your implementation implies that only *uploading* is shared, not that flipping this one will automatically make the GS unit immediately assume all values of the GS unit.
</blockquote></details>

The comment was made before the discover in this PR, so yeah, kind of outdated.
Changed to make it clear

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2856#-:-Kqv_Gd-Z6K1cSiQtMa5:bukb0gs)*

yuriks:
LGTM

PR #2857, opened 2017-08-03
Travis/AppVeyor: Deploy based upon tags

j-selby:
Previously, Citra would deploy when it was on `citra-emu/citra-nightly`. This hardcoded branch meant that building bleeding-edge/canary required a dedicated patchset, and that references to tags could not be used (both AppVeyor and Travis only support `branch` filtering or `tags`, not both).

This fixes this, by deploying only when a tag has been specified. This has a host of advantages:
- Nightly and canary no longer need to have their own independent build scripts.
- Tags can be created as required (which includes `citra-emu/citra` itself, if "stable" builds wanted to be created), and individual pushes won't cause the entire system to collapse.
- This fixes the race condition between Travis and AppVeyor.

This introduces a few issues:
- Webhooks need to be fixed (nightly is ready, be/canary is coming)
- Both AppVeyor and Travis build both for a commit, and for a tag. I haven't disabled this by default, as Travis/etc is also used for testing. Travis builds fine in parallel, but AppVeyor builds are now delayed by ~10 minutes, depending on if the commit builds first, or if the tag does.

This disables checking for the kind of task that we are building in the Travis upload task - this doesn't actually do any uploading for itself, and uploads are filtered by Travis itself. This doesn't really hurt normal builds.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2857)

### DIFF
--- a/.travis-upload.sh
+++ b/.travis-upload.sh
@@ -1,134 +1,132 @@
-if [ "$TRAVIS_EVENT_TYPE" = "push" ]&&[ "$TRAVIS_BRANCH" = "master" ]; then
-    GITDATE="`git show -s --date=short --format='%ad' | sed 's/-//g'`"
-    GITREV="`git show -s --format='%h'`"
-    mkdir -p artifacts
+GITDATE="`git show -s --date=short --format='%ad' | sed 's/-//g'`"
+GITREV="`git show -s --format='%h'`"
+mkdir -p artifacts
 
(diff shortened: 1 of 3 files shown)


j-selby:
Removed `all_branches` - not required after further testing.

jroweboy:
LGTM. Is anyone else planning on reviewing this? Speak up in the next 24 hours or else I'll just merge it. (its already been rotting here a week for a small change. don't let the +/- trick you. its mostly just a formatting change)

PR #2861, opened 2017-08-07
Refactor MotionEmu into a InputDevice

wwylele:
This is a follow-up of #2497. The motion device (accelerometer and gyroscope) is now coded in the same way as button and analog devices. This allows adding other motion devices in the future.

`core/frontend/motion_emu` is now effectively moved to `input_common/motion_emu`. The algorithm is unchanged. Unit conversion and time stretching are moved from `EmuWindow` to `Service::HID`. The whole motion input pipeline is decoupled from `EmuWindow` 

In addition, the update period and sensitivity coefficient in `MotionEmu`, previously hard-coded, are now configurable in ini.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2861)


src/core/hle/service/hid/hid.cpp
```diff
@@ -193,10 +199,19 @@ static void UpdateAccelerometerCallback(u64 userdata, int cycles_late) {
     mem->accelerometer.index = next_accelerometer_index;
     next_accelerometer_index = (next_accelerometer_index + 1) % mem->accelerometer.entries.size();
 
+    Math::Vec3<float> accel;
+    std::tie(accel, std::ignore) = motion_device->GetStatus();
+    accel *= 512; // coefficient measured by hw test
```
Subv: This should be a constant like gyroscope_coef

src/input_common/motion_emu.cpp
```diff
@@ -0,0 +1,159 @@
+// Copyright 2016 Citra Emulator Project
```
Subv: 2017

src/input_common/motion_emu.cpp
```diff
@@ -0,0 +1,159 @@
...
+    ~MotionEmuDevice() {
+        if (motion_emu_thread.joinable()) {
+            shutdown_event.Set();
+            motion_emu_thread.join();
+        }
+    }
+
+    void BeginTilt(int x, int y) {
+        mouse_origin = Math::MakeVec(x, y);
+        is_tilting = true;
```
Subv: Why is the tilt mutex not locked here?

src/input_common/motion_emu.cpp
wwylele: The file was moved from an old one... will change anyway

src/input_common/motion_emu.cpp
wwylele: because `is_tilting` is only read by `Tilt`, not by `MotionEmuThread`. It is not supposed to be operated by multiple thread.

wwylele:
> Is there any particular reason why the accelerometer and gyroscope are a single device instead of two separate ones?

I considered that the implementation of these two always come together: if they are emulated, they must be calculated together like this; if they are input from real device, having two separate devices for each doesn't make sense either (motion of which device are you sensing?)

src/core/hle/service/hid/hid.cpp
wwylele: done

src/input_common/motion_emu.cpp
wwylele: done

B3n30:
Reviewed 1 of 3 files at r1, 1 of 19 files at r3, 16 of 18 files at r4, 2 of 2 files at r5.

---

*[src/citra/default_ini.h, line 60 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2861#-KrqHfm0Oo3cIkjcQ1Ja:-KrqHfm0Oo3cIkjcQ1Jb:b-z9d798) ([raw file](https://github.com/citra-emu/citra/blob/a6d7b7178d094b1fdc2fa80e8233dcdf65ae40c5/src/citra/default_ini.h#L60)):*
> ```objc
> 
> # for motion input, the following devices are available:
> #  - "motion_emu" (default) for emulating motion input from mouse input. Required parameters:
> ```

I think motion_emu is very generic when it is actually something like motion_mouse_right. Same for the name of the class, etc.

---

*[src/input_common/motion_emu.cpp, line 22 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2861#-KrqMWatCpbaA99Imavc:-KrqMWatCpbaA99Imavd:b-ia21ck) ([raw file](https://github.com/citra-emu/citra/blob/a6d7b7178d094b1fdc2fa80e8233dcdf65ae40c5/src/input_common/motion_emu.cpp#L22)):*
> ```cpp
> 
>     ~MotionEmuDevice() {
>         if (motion_emu_thread.joinable()) {
> ```

NIT: Switching MotionEmuDevices during runtime isn't supported yet, but is still think resetting the position to 0 should happen here

---

*[src/input_common/motion_emu.cpp, line 42 at r5](https://reviewable.io:443/reviews/citra-emu/citra/2861#-KrqLHiN6ZP_lm0BwBkz:-KrqLHiN6ZP_lm0BwBl-:bkpn26m) ([raw file](https://github.com/citra-emu/citra/blob/a6d7b7178d094b1fdc2fa80e8233dcdf65ae40c5/src/input_common/motion_emu.cpp#L42)):*
> ```cpp
>                 tilt_direction = mouse_move.Cast<float>();
>                 tilt_angle = MathUtil::Clamp(tilt_direction.Normalize() * sensitivity, 0.0f,
>                                              MathUtil::PI * 0.5f);
> ```

Shouldn't the tilt angle be between 0 and PI at least in one direction?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2861)*

src/input_common/motion_emu.h
```diff
@@ -0,0 +1,46 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
```
lioncash: There should be a new line after the `#pragma once`

src/input_common/motion_emu.cpp
```diff
@@ -0,0 +1,160 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/math_util.h"
+#include "common/quaternion.h"
+#include "common/thread.h"
```
lioncash: `<chrono>`, `<mutex>`, `<thread>`, and `<tuple>` should still be included directly to avoid indirect inclusions.

src/input_common/motion_emu.h
```diff
@@ -0,0 +1,46 @@
...
+#include "common/vector_math.h"
```
lioncash: Why does `vector_math` need to be here?

wwylele: copy-paste leftover. will remove

wwylele:
(sorry, can't open Reviewable right now)
@B3n30 
>I think motion_emu is very generic when it is actually something like motion_mouse_right. Same for the name of the class, etc.

The name was inherited from previous code so I didn't rename it to avoid confusion. I agree that it is not the best name, but  `motion_mouse_right` sounds too fragmentary to me... given there is no other implemented devices and the name is not that exposed to user, I prefer to keep a short name for now.

>NIT: Switching MotionEmuDevices during runtime isn't supported yet, but is still think resetting the position to 0 should happen here

But this is in the destructor. Assigning any value to the member makes no sense. `MotionEmuDevice` is owned by `MotionEmuDeviceWrapper`, which keeps alive until the core decide to unload the device. The core should manage its own reset position if the device gets unloaded (destroyed)

wwylele:
> Shouldn't the tilt angle be between 0 and PI at least in one direction?

The code was inherited from the old one. IIRC I discussed this before. The reason I confine the range to 0 to pi/2 is that the user can easily goes to pi angle by moving mouse, at which angle every motion gets inverted and may cause confusion. This isn't supposed to be a full-motion emulation, but an easy one intuitive to use, anyway

jroweboy:
LGTM

PR #2863, opened 2017-08-09
HID: zero unused PadState bits

wwylele:
Unused PadState bits usually means nothing to commercial games, but some homebrew programs (such as devkitPro-3ds-examples/input/read-controls) assume they are all zero.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2863)

### DIFF
--- a/src/core/hle/service/hid/hid.h
+++ b/src/core/hle/service/hid/hid.h
@@ -24,7 +24,7 @@ namespace HID {
  */
 struct PadState {
     union {
-        u32 hex;
+        u32 hex{};
 
         BitField<0, 1, u32> a;
         BitField<1, 1, u32> b;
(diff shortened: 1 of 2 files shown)


mailwl:
maybe better zero hex right in https://github.com/citra-emu/citra/blob/master/src/core/hle/service/hid/hid.h#L27 ?

i don't like split constructor and initialization

wwylele:
Changed. But I am still open to other opinions.
Btw add a similar fix to ir:rst

src/core/hle/service/hid/hid.h
```diff
@@ -24,7 +24,7 @@ namespace HID {
  */
 struct PadState {
     union {
-        u32 hex;
+        u32 hex = 0;
```
bunnei: nit: `u32 hex{};` (less code)

wwylele: not a fan of `{}`... I think `= 0` indicates setting to zero in a clearer way, and there is really no need to save one or two characters in the code

wwylele: alright, B3n30 also prefer `{}` so I changed it

PR #2865, opened 2017-08-09
PICA: implemented geometry shader

wwylele:
Credit to @ds84182 and @JayFoxRox  for their original implementation, and to @fincs for their research on GS

Sorry for putting all these stuff into one PR. These commits only work when come together, and I didn't find a good way to split them.

Probably better to review this commit by commit. Detailed description is in each commit message.

Most unknowns from the original implementation were resolved in recent research. There are still a few unknowns remained to test, for which I made assumptions and put ASSERT for them. This branch has been tested in bleeding edge and no ASSERT crash has been reported.

----

For anyone who has reviewed [JayFoxRox's GS branch](https://github.com/jayfoxrox/citra/pull/49), implementation difference in this branch are listed here for reference:
 - respects the topology type in assembler even with GS (HW tested), and move winding (inverting) into the specific `Shader` topology type (HW tested)
 - keeps the vertex cache, and instead change the cached vertex type. The cache happens before GS
 - wraps the GS input buffer stuff into `GeometryPipeline` and move it to global state, for supporting immediate mode (HW tested) (with buffering in this mode, also HW tested) and the other GS modes
 - adds support for subdivision mode and particle mode
 - leaves the four shader units and the full scheduler unimplemented. This is considered out of scope of this PR.

----

Some test result:
 - devkitPro/3ds-examples/graphics/gpu/geoshader: works fine
 - devkitPro/3ds-examples/graphics/gpu/loop_subdivision: works fine
 - devkitPro/3ds-examples/graphics/gpu/particles: works fine. The output is not exactly the same as real 3DS. This is believed to be caused by float24 precision error accumulation in a PRNG in the shader program. Resolving it is out of scope of this PR.
 - Pokemon OR/AS: works fine
 - Monster Hunter XX: works fine
 - Professor Layton and the Miracle Mask: works fine
 - snes9x (a snes emulator): display nothing in game. If all shader registers are initialized to 1, it works fine then. The source code shows that its geometry shader never sets the output position.w. This presumably relies on register preservation for the special GS unit when switching between VS and GS, so this will only work when the 4 shader units and the scheduler is implemented.
 - other games tested by players using bleeding edge build: only one crash found, which is later fixed in this branch

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2865)

### DIFF
--- a/src/video_core/geometry_pipeline.cpp
+++ b/src/video_core/geometry_pipeline.cpp
@@ -1,0 +1,274 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "video_core/geometry_pipeline.h"
+#include "video_core/pica_state.h"
+#include "video_core/regs.h"
+#include "video_core/renderer_base.h"
+#include "video_core/video_core.h"
+
+namespace Pica {
+
+/// An attribute buffering interface for different pipeline modes
+class GeometryPipelineBackend {
+public:
+    virtual ~GeometryPipelineBackend() = default;
+
+    /// Checks if there is no incomplete data transfer
+    virtual bool IsEmpty() const = 0;
+
+    /// Checks if the pipeline needs a direct input from index buffer
+    virtual bool NeedIndexInput() const = 0;
+
+    /// Submits an index from index buffer
+    virtual void SubmitIndex(unsigned int val) = 0;
+
+    /**
+     * Submits vertex attributes
+     * @param input attributes of a vertex output from vertex shader
+     * @return if the buffer is full and the geometry shader should be invoked
+     */
+    virtual bool SubmitVertex(const Shader::AttributeBuffer& input) = 0;
+};
+
+// In the Point mode, vertex attributes are sent to the input registers in the geometry shader unit.
+// The size of vertex shader outputs and geometry shader inputs are constants. Geometry shader is
+// invoked upon inputs buffer filled up by vertex shader outputs. For example, if we have a geometry
+// shader that takes 6 inputs, and the vertex shader outputs 2 attributes, it would take 3 vertices
+// for one geometry shader invocation.
+// TODO: what happens when the input size is not divisible by the output size?
+class GeometryPipeline_Point : public GeometryPipelineBackend {
+public:
+    GeometryPipeline_Point(const Regs& regs, Shader::GSUnitState& unit) : regs(regs), unit(unit) {
+        ASSERT(regs.pipeline.variable_primitive == 0);
+        ASSERT(regs.gs.input_to_uniform == 0);
+        vs_output_num = regs.pipeline.vs_outmap_total_minus_1_a + 1;
+        size_t gs_input_num = regs.gs.max_input_attribute_index + 1;
+        ASSERT(gs_input_num % vs_output_num == 0);
+        buffer_cur = attribute_buffer.attr;
+        buffer_end = attribute_buffer.attr + gs_input_num;
+    }
+
+    bool IsEmpty() const override {
+        return buffer_cur == attribute_buffer.attr;
+    }
+
+    bool NeedIndexInput() const override {
+        return false;
+    }
+
+    void SubmitIndex(unsigned int val) override {
+        UNREACHABLE();
+    }
+
+    bool SubmitVertex(const Shader::AttributeBuffer& input) override {
+        buffer_cur = std::copy(input.attr, input.attr + vs_output_num, buffer_cur);
+        if (buffer_cur == buffer_end) {
+            buffer_cur = attribute_buffer.attr;
+            unit.LoadInput(regs.gs, attribute_buffer);
+            return true;
+        }
+        return false;
+    }
+
+private:
+    const Regs& regs;
+    Shader::GSUnitState& unit;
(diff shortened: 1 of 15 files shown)


RavenHome1:
so this is the final and complete version of GS?

Dragios:
@RavenHome1 
- leaves the four shader units and the full scheduler unimplemented. This is considered out of scope of this PR.

Which means is not complete.

Dragios:
@JayFoxRox [test mention]

src/video_core/geometry_pipeline.h
```diff
@@ -0,0 +1,49 @@
...
+namespace Pica {
+
+struct State;
+
+class GeometryPipelineBackend;
+
+/// A pipeline receiving from vertex shader and sending to geometry shader and primitive assembler
+class GeometryPipeline {
+public:
+    GeometryPipeline(State& state);
```
lioncash: `explicit GeometryPipeline(State& state);`

src/video_core/shader/shader.h
```diff
@@ -61,12 +69,36 @@ static_assert(std::is_pod<OutputVertex>::value, "Structure is not POD");
...
+ */
+struct GSEmitter {
+    std::array<std::array<Math::Vec4<float24>, 16>, 3> buffer;
+    u8 vertex_id;
+    bool prim_emit;
+    bool winding;
+    u32 output_mask;
+
+    // Function objects are hidden behind a raw pointer to make the structure standard layout type,
+    // for JIT to use offsetof to access other members.
```
lioncash: This is indicative that the two sets of data should be logically separate from one another (either separating the main struct, or putting the JITable data into its own struct).

src/video_core/shader/shader.cpp
```diff
@@ -82,6 +83,44 @@ void UnitState::WriteOutput(const ShaderRegs& config, AttributeBuffer& output) {
...
+    delete handlers;
+}
+
+void GSEmitter::Emit(Math::Vec4<float24> (&vertex)[16]) {
+    ASSERT(vertex_id < 3);
+    std::copy(std::begin(vertex), std::end(vertex), buffer[vertex_id].begin());
+    if (prim_emit) {
+        if (winding)
+            handlers->winding_setter();
+        for (unsigned int i = 0; i < 3; ++i) {
```
lioncash: `for (size_t i = 0; i < buffer.size(); i++)`

src/video_core/geometry_pipeline.cpp
```diff
@@ -0,0 +1,272 @@
...
+#include "video_core/geometry_pipeline.h"
+#include "video_core/pica_state.h"
+#include "video_core/regs.h"
+#include "video_core/renderer_base.h"
+#include "video_core/video_core.h"
+
+namespace Pica {
+
+/// An attribute buffering interface for different pipeline modes
+class GeometryPipelineBackend {
```
lioncash: This should probably have a virtual destructor.

src/video_core/shader/shader.h
```diff
@@ -61,12 +69,36 @@ static_assert(std::is_pod<OutputVertex>::value, "Structure is not POD");
...
+static_assert(std::is_standard_layout<GSEmitter>::value, "GSEmitter is not standard layout type");
+
+/**
  * This structure contains the state information that needs to be unique for a shader unit. The 3DS
  * has four shader units that process shaders in parallel. At the present, Citra only implements a
  * single shader unit that processes all shaders serially. Putting the state information in a struct
  * here will make it easier for us to parallelize the shader processing later.
  */
 struct UnitState {
+    UnitState(GSEmitter* emitter = nullptr);
```
lioncash: `explicit UnitState(GSEmitter* emitter = nullptr);`

src/video_core/shader/shader.cpp
```diff
@@ -82,6 +83,44 @@ void UnitState::WriteOutput(const ShaderRegs& config, AttributeBuffer& output) {
...
+            }
+            handlers->vertex_handler(output);
+        }
+    }
+}
+
+GSUnitState::GSUnitState() : UnitState(&emitter) {}
+
+void GSUnitState::SetVertexHandler(VertexHandler vertex_handler, WindingSetter winding_setter) {
+    emitter.handlers->vertex_handler = vertex_handler;
```
lioncash: These can both be `std::move`d

src/video_core/shader/shader.h
wwylele: This doesn't resolve the essential problem though. To be able to be used by JIT, `GSUnitState` must be standard layout type, which means it can't have any function object as direct or indirect member. But on the other hand, these two function objects must belong to `GSUnitState` or `ShaderSetup` (both reference by JIT by `offsetof`) in order for shader program to invoke at runtime. Thus I can't find a good way better than using plain pointers to hide them.

src/video_core/geometry_pipeline.cpp
wwylele: done

wwylele:
Calling for review. This is opened for nearly a month and it begins to fade from my memory.

bunnei:
We don't really have any GPU experts actively reviewing... so I'm fine YOLO'ing this one. Code looks fine, and it's been in BE/canary for a while without any reported regressions. As always, if anyone has any issues with this, they can submit a followup PR improving this code.

Issue #2791, opened 2017-06-19
Super Mario 3D Land - Missing water effect in W4-1 level


Leo626:
Apart from the GS related lines (tested with BE-86 and the water is still the same), log (Critical) shows:

```HW.GPU <Critical> video_core\renderer_opengl\gl_shader_gen.cpp:GLShader::SampleTexture:169: Unhandled texture type 5```

[LOG](https://hastebin.com/daqorohage.pas)
[LOG](https://hastebin.com/ijakinoqot.rb) (Non-shining stars)

![sm3dland](https://user-images.githubusercontent.com/17358613/27290756-8c5ac5ae-54e4-11e7-9cac-217a29c0d74e.png)

[Save file](https://github.com/citra-emu/citra/files/737215/SM3DL.-.Regular.World.zip)
[GameData.zip](https://github.com/citra-emu/citra/files/1086330/GameData.zip) (Non-shining stars)

What it looks like in on 3DS in [video](https://youtu.be/NvXMg_H8ByQ?t=1m20s). 

Commit: d0888f8
Windows 10
i5 4670k
GTX 970 (382.05)
8GB RAM

wwylele:
not necessarily related, but `Commandlist tried to write to invalid register` error looks pretty broken. Does it shows up in the same area or somewhere else?

Leo626:
It only shows when I press A to start on the start screen. It doesn't appear when I start that level.

B3n30:
solve by #2921

Leo626:
This isn't fixed. What I had meant to say was the water shader effects were missing hence the water looked bland in Citra. The small black thing (ripples) by Mario is batching related and is fixed though.

Hardware
![ntr_0041](https://user-images.githubusercontent.com/17358613/30820156-7728fbdc-a1f7-11e7-9409-449632a17a1a.png)
Hardware with annotations
![ntr_0041](https://user-images.githubusercontent.com/17358613/30820174-8421e9ca-a1f7-11e7-82f0-fbab8b55cfe9.png)
Nightly-d881dee
![citra-qt_2017-09-25_09-14-04](https://user-images.githubusercontent.com/17358613/30820200-9a761f84-a1f7-11e7-93eb-7a47e0128acf.png)

PR #2866, opened 2017-08-10
Added batch fix in command_processor; Make it optional (opt-out)

B3n30:
This fix resides inside BE since Oct. 2016 (credit to @JayFoxRox ). It fixes some graphical glitches (e.g. #2791, #1681)
This fix should also be in nightly and canary.

Since for most games it adds unnecessary draws, I added an option in the settings to disable the fix. By default batch_fix is enabled, thus BE users won't even recognize the change. 
I tested it with some games and couldn't see any slowdowns (maybe 1 fps). 

With batch_fix:
bildschirmfoto 2017-08-10 um 08 15 40

Without batch_fix:
bildschirmfoto 2017-08-10 um 08 17 01

The option for enable/disable the batch fix in qt is located together with the other video settings:
bildschirmfoto 2017-08-10 um 08 17 24

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2866)

### DIFF
--- a/src/citra/config.cpp
+++ b/src/citra/config.cpp
@@ -87,6 +87,7 @@ void Config::ReadValues() {
     Settings::values.use_vsync = sdl2_config->GetBoolean("Renderer", "use_vsync", false);
     Settings::values.toggle_framelimit =
         sdl2_config->GetBoolean("Renderer", "toggle_framelimit", true);
+    Settings::values.use_batch_fix = sdl2_config->GetBoolean("Renderer", "use_batch_fix", true);
 
     Settings::values.bg_red = (float)sdl2_config->GetReal("Renderer", "bg_red", 0.0);
     Settings::values.bg_green = (float)sdl2_config->GetReal("Renderer", "bg_green", 0.0);
--- a/src/citra/default_ini.h
+++ b/src/citra/default_ini.h
@@ -109,6 +109,10 @@ custom_bottom_bottom =
 # 0: Off , 1  (default): On
 toggle_framelimit =
 
+#Enable batch fix.
+# 0: Off , 1  (default): On
+use_batch_fix =
+
 # Swaps the prominent screen with the other screen.
 # For example, if Single Screen is chosen, setting this to 1 will display the bottom screen instead of the top screen.
 # 0 (default): Top Screen is prominent, 1: Bottom Screen is prominent
--- a/src/citra_qt/configuration/config.cpp
+++ b/src/citra_qt/configuration/config.cpp
@@ -69,6 +69,7 @@ void Config::ReadValues() {
     Settings::values.resolution_factor = qt_config->value("resolution_factor", 1.0).toFloat();
     Settings::values.use_vsync = qt_config->value("use_vsync", false).toBool();
     Settings::values.toggle_framelimit = qt_config->value("toggle_framelimit", true).toBool();
+    Settings::values.use_batch_fix = qt_config->value("use_batch_fix", true).toBool();
 
     Settings::values.bg_red = qt_config->value("bg_red", 0.0).toFloat();
     Settings::values.bg_green = qt_config->value("bg_green", 0.0).toFloat();
@@ -215,6 +216,7 @@ void Config::SaveValues() {
     qt_config->setValue("resolution_factor", (double)Settings::values.resolution_factor);
     qt_config->setValue("use_vsync", Settings::values.use_vsync);
     qt_config->setValue("toggle_framelimit", Settings::values.toggle_framelimit);
+    qt_config->setValue("use_batch_fix", Settings::values.use_batch_fix);
 
     // Cast to double because Qt's written float values are not human-readable
     qt_config->setValue("bg_red", (double)Settings::values.bg_red);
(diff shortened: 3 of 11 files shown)


wwylele:
I am not sure if this is a good idea, but I think at least you should put a TODO there for this is a hack for unknown thing

jroweboy:
I'm very much against making it configurable. Citra has gone this far without adding configurable "hacks" to the emulator, and I think it should stay that way. It should either be always enabled or just not merged at all, imo.

chris062689:
Is this considered a hack or is it intended functionality of the 3DS?

JayFoxRox:
This is a hack. Therefore I'll veto it.
Keeping the hack in BE is acceptable, but still not good.

We should research how this actually works and fix the emulator.

chris062689:
@JayFoxRox I was under the assumption this isn't supposed to be merged into the main source tree.

I believe it's only here so we can have feature parity between canary and bleeding edge.

JayFoxRox:
Even then, an opt-out / menu option makes no sense for such a hack.

Basically Citra makes an assumption so it can optimize better. My hack (if I remember correctly) just removed / hackily disabled the optimization as the assumption turned out to be wrong.

With the hack, it's more accurate behaviour, but probably still wrong.

Users also don't understand what this feature does, so the menu option will just lead to confusion.
There is absolutely no reason to not have it permanently allowed. The performance impact is minimal.
If anybody is bothered by the performance impact, they should figure out how this actually works, then fix it.

---

There is also no hint suggesting that this should not be merged / is a temporary solution. A maintainer might see this in 2 months and just merge it, thinking it was just forgotten.

B3n30:
It is 100% the copy of @JayFoxRox hack from BE. My intention was since there might be temporary 3 release types an a movement to canary it is a bad idea to have the hacks/fixes in the repos. Also i see no reason why this shouldn't be in nightly. Of course it isn't the correct way of doing it, but the current way is wrong, too.

I agree that i missed a clear documentation. Question is if this PR should get merged or just get's the tags to add it to the corresponding and an additional tag "hack" or so.

I added the setting since there was a discussion about performance loss due to the additional draws, but if the common sense is, that it shouldn't be optional I'll remove the setting.

JayFoxRox:
Also see #1688 which was my attempt. If this was a good solution, it would already be in nightly.

B3n30:
So to get this clear: We won't merge hacks, but it's fine to have them in BE? If this is the case, I'm fine with it, but I wouldn't hide them as a PR in the BE repo. I don't think such hacks should stay in theBE repo for about a year. Then my suggestion would be to have those hacks as a PR in the citra repo tagged with BE and maybe a tag "NotForMerge" or "Hack".

JayFoxRox:
This is getting really offtopic. But: Yes, we don't merge hacks.
Why did this hack stay for more than a year? Because of historic reasons.

We initially introduced Bleeding Edge to combat unofficial builds popping up. The first-party titles (most Mario games for example) were considered important and when they broke, I've tried to fix it. While working on the fix, I noticed that my understanding of the issue was bad, so I closed my PR ( #1688 ). As unofficial builds did not care about the hacky nature, we also had to follow and keep this hack in BE.
Otherwise unofficial builds would have run Mario games, but Bleeding Edge wouldn't.

I was going to look into it again but was busy with other things. So the temporary hack is still present today.


This also shows why unofficial builds or hacks are bad: If there is no visible issue, people are not motivated to fix it. If there wasn't a hack around this, this would have been fixed a long time ago probably.

B3n30:
Closing, because it seems that my approach to centralize all PRs (even hacks) in the main repo isn't wanted

JayFoxRox:
I think having it here is a good idea. But turning this hack into a permanent or mergeable solution isn't.
It can be PR'd here with a clear warning that it's a hack, and it should not be extended (like the new menu option).

However, I think it will get more attention when it's elsewhere, or if we just drop this hack from BE / someone works on an actual solution.

mailwl:
maybe just leave it on config file (like new3ds option)?

ds84182:
Alright, I've made some discoveries regarding the batch fix.

When the GPU is in "Drawing Mode", GPU registers that configure drawing related things (like textures and lighting) can still be set.

GPUREG_START_DRAW_FUNC0 _probably_ just enables vertex processing in the PICA, making "gpu_mode", "PipelineRegs::GPUMode::Configuring", and "PipelineRegs::GPUMode::Drawing", misnomers. [The some of the oddities recorded with GPUREG_START_DRAW_FUNC0 are related to vertex processing.](https://www.3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_START_DRAW_FUNC0)

I've verified [with this hardware test.](https://github.com/ds84182/3ds-gpu-stuff/commit/98b576496c8ba0fc9179355e86ed356c47c6fb36) On real hardware two triangles are rendered, one with a kitten and one without. On Citra, two black triangles are rendered.

A proper fix would be calling DrawTriangles after every AddTriangle in command_processor.cpp, however, this will affect the speed of OpenGL rendering when immediate mode rendering is used (since OpenGL wouldn't be able to batch any immediate mode triangles, it would submit them to the host GPU one by one).

Another fix would be telling the rasterizer (in this case OpenGL) that a drawing configuration register has changed so the previous batch can be flushed, but this depends on any speed regressions caused by the proper fix.

This PR (and the PRs that came before it) implement half of the proper fix. The other half is calling DrawTriangles after adding an immediate mode triangle.

PR #2868, opened 2017-08-10
SwRasterizer/Lighting: use make_tuple instead of constructor

wwylele:
implicit tuple constructor is c++17 thing, which is not supported by some not-so-old libraries (for example, the one with gcc5). Play safe for now.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2868)

### DIFF
--- a/src/video_core/swrasterizer/lighting.cpp
+++ b/src/video_core/swrasterizer/lighting.cpp
@@ -244,7 +244,7 @@ std::tuple<Math::Vec4<u8>, Math::Vec4<u8>> ComputeFragmentsColors(
                                          MathUtil::Clamp(specular_sum.z, 0.0f, 1.0f) * 255,
                                          MathUtil::Clamp(specular_sum.w, 0.0f, 1.0f) * 255)
                         .Cast<u8>();
-    return {diffuse, specular};
+    return std::make_tuple(diffuse, specular);
 }
 
 } // namespace Pica


B3n30:
LGTM

wwylele:
For reference:

> Selby - Today at 11:31 AM
> sw lighting broke ubuntu, I think
> https://github.com/citra-emu/citra/blame/master/src/video_core/swrasterizer/lighting.cpp#L247
> Log: https://travis-ci.org/j-selby/citra-nightly/jobs/262989781#L5977 - ubuntu 16.04 via docker
> @wwylele - when you get a sec
> wwylele / 白疾風 - Today at 11:38 AM
> ... why it didin't error on previous ci and my local ubuntu 16.04(edited)
> Selby - Today at 11:38 AM
> have absolutely no clue.
> https://github.com/j-selby/citra-nightly/blob/3681270810687f2960fa4f457bb27f32696377c1/.travis-build-docker.sh - build script
> you haven't got a testing version of gcc/g++?
> the ci obviously does
> wwylele / 白疾風 - Today at 11:43 AM
> g++ (Ubuntu/Linaro 6.3.0-18ubuntu2~16.04) 6.3.0 20170519 
> this is my g++ version
> I see it uses g++5
> CI on master is also g++6
> Selby - Today at 11:45 AM
> yeah, ci pulls from ubuntu-toolchain-test (or similar)
> wwylele / 白疾風 - Today at 11:50 AM
> alright, looks like it is the fault in my code
> (or actually, subv's)
> It just happened to be an extension to all three compiler updates
> I'll make a quick fix

and c++ standard reference: 
http://en.cppreference.com/w/cpp/language/list_initialization
> If the previous stage does not produce a match, all constructors of T participate in overload resolution against the set of arguments that consists of the elements of the braced-init-list, with the restriction that only non-narrowing conversions are allowed. __If this stage produces an explicit constructor as the best match for a copy-list-initialization, compilation fails__ (note, in simple copy-initialization, explicit constructors are not considered at all). 

http://en.cppreference.com/w/cpp/utility/tuple/tuple (see how the `explicit` specifier get erased in C++17)

PR #2869, opened 2017-08-10
Travis: Use Docker to build for Linux

j-selby:
Travis seems to be sticking to Ubuntu 14.04 for the next century, so this PR implements Docker as a means to get around that. A standard Ubuntu 16.04 image from the Docker repository is used, and there is no need for hacked up PPAs/etc. This adds a minute or two of additional build time.

This fixes Linux builds not finding the correct versions of linked libraries, and ensures that we are using `libpng16` as well. This has been tested on Arch (02/08/17) and Ubuntu 16.04.

This also has the additional benefit of allowing Docker images of different distributions to be used, if that was a requirement (Debian, Ubuntu 17.04, etc).

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2869)

### DIFF
--- a/.travis-build-docker.sh
+++ b/.travis-build-docker.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+set -e
+set -x
+
+cd /citra
+
+apt-get update
+apt-get install -y build-essential libsdl2-dev qtbase5-dev libqt5opengl5-dev libcurl4-openssl-dev libssl-dev wget git
+
+# Get a recent version of CMake
+wget https://cmake.org/files/v3.9/cmake-3.9.0-Linux-x86_64.sh
+echo y | sh cmake-3.9.0-Linux-x86_64.sh --prefix=cmake
+export PATH=/citra/cmake/cmake-3.9.0-Linux-x86_64/bin:$PATH
+
+mkdir build && cd build
+cmake .. -DCMAKE_BUILD_TYPE=Release
+make -j4
+
+ctest -VV -C Release
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -44,15 +44,7 @@ fi
 
 #if OS is linux or is not set
 if [ "$TRAVIS_OS_NAME" = "linux" -o -z "$TRAVIS_OS_NAME" ]; then
-    export CC=gcc-6
-    export CXX=g++-6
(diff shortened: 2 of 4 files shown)


jroweboy:
In review, I brought up the fact that this removes the cache that we added in #2060 Its likely not going to be a real issue though, so its probably not worth the effort to download cmake, store it in the cache, load it on the volume mount that docker has and set everything up that way. But I figure that its worth it to mention it here in case we do need to go back and add a cache again.

LGTM

probonopd:
Why do projects want to build against the latest versions of distributions? This has the unintended consequence that the build artifacts will only run on the latest versions of distributions. It would be much appreciated if building on 14.04 was still supported, and tested.

j-selby:
@probonopd Several reasons. 

- Citra is *experimental*, and we don't want to be targeting older distributions simply because it is more work for us. We are already stretched thin on keeping everything working on the platforms which we *do* support (Windows MSVC/MinGW, macOS, Linux). We are supporting the latest Ubuntu LTS, which is more then suitable.
- We are targeting a newer version of g++ then Ubuntu 14.04 provides. We moved off 14.04 because using a 3rd party repository that broke compatibility with everything was, quite frankly, not in our best interests.

If users want Citra to be on different distros, they can, either:

- Build it from source
- Get the package into official package management systems (which would be the best result for those users).

And yeah, I get this breaks compatibility with your AppImages, but we aren't going to hold back development for those who love using out of date systems.

probonopd:
> If users want Citra to be on different distros, they can, either:
> Build it from source

That is hard to impossible (without rewriting the software) if application developers "blindly" assume newer functionality (e.g., in the compiler or in libraries) to be available. Some projects (like LibreOffice and Firefox) deliberately make sure their latest versions are still available for the oldest still-supported distributions. 

> And yeah, I get this breaks compatibility with your AppImages, but we aren't going to hold back development for those who love using out of date systems.

It doesn't. They just won't run on systems older than your build system, e.g., the oldest still-supported CentOS and Ubuntu LTS releases. If that's what you want, then well, your decision.

B3n30:
1. We all agreed that this was the way to go. You had time to review this PR when it was open. If you aren't happy with it, make a PR that "fixes" it. Citra is still open source
2. And just because a Version is still supported doesn't mean Citra have to support that release, too.
3. Linux is the OS where Citra can be build very is by the useres them self, so I don't think it is necessary to support many distributions. Supporting the distribution most users use (the most recent one) seems sufficient.

PR #2872, opened 2017-08-10
SwRasterizer/Lighting: implement geometric factor

wwylele:
This is the swrasterizer counterpart of #2776 . Same test program there can be used

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2872)

### DIFF
--- a/src/video_core/swrasterizer/lighting.cpp
+++ b/src/video_core/swrasterizer/lighting.cpp
@@ -55,6 +55,9 @@ std::tuple<Math::Vec4<u8>, Math::Vec4<u8>> ComputeFragmentsColors(
 
         light_vector.Normalize();
 
+        Math::Vec3<float> norm_view = view.Normalized();
+        Math::Vec3<float> half_vector = norm_view + light_vector;
+
         float dist_atten = 1.0f;
         if (!lighting.IsDistAttenDisabled(num)) {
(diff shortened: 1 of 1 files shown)


src/video_core/swrasterizer/lighting.cpp
```diff
@@ -55,6 +55,9 @@ std::tuple<Math::Vec4<u8>, Math::Vec4<u8>> ComputeFragmentsColors(
 
         light_vector.Normalize();
 
+        Math::Vec3<float> norm_view = view.Normalized();
+        Math::Vec3<float> half_vector = norm_view + light_vector;
```
Dragios: Any reason why `.Normalized()` is not in the formula itself?

wwylele: Because the one used in the `geo_factor` below is not the normalized one - it actually needs its length there

B3n30:
:lgtm: I couldn't spot any diff to HW

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2872#-:-KrqSot_zBTKJO8Td8q0:b-exa2ph)*

yuriks:
LGTM

PR #2873, opened 2017-08-11
[RFC, WIP] Handle the Float<M,E> with the correct precision

B3n30:
With the PR #2865 the homebrew example particles shows a different behavior than a real 3ds. But this isn't caused by wrong implementation of GS, but instead is caused by our treatment of float24. We use a normal float (32bit) to represent it. The difference arises from the fact, that the particles example uses floats for rng. With our wrong treatment of float24, the bigger precision of float32 causes this difference. This PR treats float24 with the needed smaller precision and thus fixes it. (Not in JIT!!!, We would need a similar approach in JIT).

The values are still stored as float32. All the operations are done with the more precise float on HW, to don't slow the system down. But after each calculation the precision is reduced to the needed one. This is done in FromFloat32(), a function each calculation uses.

This calculation gives exactly the same result as calculating with Float<M,E> in the first place. Only exception is, if during the float32 calculation the rounding (done one the 24 bit of the mantissa) cases (through carry) the M+1 bit to change from 0 to 1. This would cause the rounding algorithm to falsely round up instead of down. For float24 (aka Float<16,7>) the probability for that is 1/2^(24-16) = 0.4%.

There might still be optimizations to speed the process up.

The particles homebrew might be the only app that shows a wrong behavior due to (un)-precise floats, but there might be other graphical glitches that happen due to to precise float calculation. Usually it is not obvious that the cause for that glitches is wrong float calculation and thus very hard to debug.

Question is: Is this precision necessary? Should a similar approach be made in JIT?
Is even more precision necessary? Note that that would mean we have to do float calculations in SW and cant rely on any float-calculation-on-HW, which would mean a huge performance drop.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2873)

### DIFF
--- a/src/video_core/pica_types.h
+++ b/src/video_core/pica_types.h
@@ -28,24 +28,22 @@ public:
         u32 hex;
         std::memcpy(&hex, &val, sizeof(u32));
 
-        if (val == 0.f)
-            return Float<M, E>::Zero(); // Pica200 has no -0
-        const int bias = 127 - Float<M, E>::EXPONENT_BIAS;
-        u32 sign = hex >> 31;
-        u32 exponent = ((hex >> 23) & ((1 << 8) - 1)) - bias;
-        u32 mantissa = (hex & ((1 << 23) - 1)) >> (23 - M - 1);
+        // Take care of special case 0.0
+        if (val == 0.f) {
+            // Since Pica has no -0, we don't have to check for the sign
+            return Float<M, E>::Zero();
+        }
+        const s32 bias = 127 - EXPONENT_BIAS;
+        const s32 sign = hex & 1 << 31;
+        const s32 exponent = ((hex >> 23) & ((1 << 8) - 1)) - bias;
 
-        // calculate with a M+1-bit mantissa and round down to M bit
-        if (mantissa << 31)
-            mantissa = ((mantissa >> 1) + 1);
-        else
-            mantissa = (mantissa >> 1);
-
+        // Take care of the special cases NAN and INF
         if (std::isnan(val)) {
             Float<M, E> res;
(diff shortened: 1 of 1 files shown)


wwylele:
Test result (sorry for a late test):
Although the overall appearance does get closer to real 3DS (the distribution is more uniform), it still doesn't get good accuracy down to individual particles. The PRNG begins to diverge from the third or forth iteration. For comparison, in shader JIT, where the precision correction doesn't take effect, it begins to diverge around the same iteration.

So my conclusion is: either there are some other bugs causing the error, or this correction is not accurate enough. In either cases, I would say this correction doesn't worth it

B3n30:
Anyone else having an opinion about that? Otherwise I will close it.

B3n30:
Since there seems to be no interest in precise pica floats, I'm closing this PR

yuriks:
You already closed this but, my thoughts on this are: At this point this representation stops making sense. You're converting to a lower precision float and then converting back to a 32-bit one, which is stored in the value. This does not really provide for accurate math as if done in the lower precision, and also makes the type pretty misleading. If you do this then the value should be stored internally in the target precision, and with some kind of soft-float impl (or at least convert to float before doing arithmetic, not when storing).

B3n30:
Since during any float calculation the precision is extended anyway, there is no difference in doing the math with float32 or float24 or any other. The differences arise solely, due to rounding after calculation and clamping to min/max exponent. I also did a implementation with sw-calculation but that is far to slow to be actually implemented. And as stated in my first sentence it has no real benefit since the calculation result will still be the same. Storing it internally as a float32 is just to don't break jit

PR #2876, opened 2017-08-12
Mii Selector Applet: update Mii structures

mailwl:
According to https://www.3dbrew.org/wiki/Mii_Selector

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2876)

### DIFF
--- a/src/core/hle/applets/mii_selector.h
+++ b/src/core/hle/applets/mii_selector.h
@@ -16,51 +16,46 @@ namespace HLE {
 namespace Applets {
 
 struct MiiConfig {
-    u8 unk_000;
-    u8 unk_001;
-    u8 unk_002;
-    u8 unk_003;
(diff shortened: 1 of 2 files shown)


src/core/hle/applets/mii_selector.h
```diff
@@ -16,51 +16,46 @@ namespace HLE {
 namespace Applets {
 
 struct MiiConfig {
-    u8 unk_000;
-    u8 unk_001;
-    u8 unk_002;
-    u8 unk_003;
-    u8 unk_004;
+    u8 cancel_button_flag;
```
wwylele: `_flag` is probably not our naming style... I guess `enable_cancel_button` and so on works better

mailwl: if you have another мнение готов изменить.

Issue #2133, opened 2016-10-22
Pokemon X/Y - Black spot on male character's face


Dragios:
At first I thought this is related to this graphical bug #1515 but it seems like that one is fixed but not this one. Turning on/off shader JIT & using both hardware or software renderer does not change the result.

Citra (as of cc7f115)
![Black Spot Face](https://cloud.githubusercontent.com/assets/18496868/19618958/d7db784a-988b-11e6-8fae-c78f8e93f1cc.png)

3DS
![3DS reference](https://cloud.githubusercontent.com/assets/18496868/19619023/5ec6be08-988e-11e6-8877-e51377223154.png)

wwylele:
Should mention that when the male character is on the screen, log spam with
```
HW.Memory <Error> core/memory.cpp:GetPointer:282: unknown GetPointer @ 0x00000000
```

Subv:
The game renders the characters fine to the Color Buffer, but then enables and loads a NULL texture (Param 0x01002200: Address 0, Width 0, Height 0, Type 0, WrapT Repeat, WrapS Repeat, Filtering Linear) and draws the two black rectangles with it, one after the other.

Subv:
Here's a command dump of the game drawing both spots:
https://gist.github.com/Subv/7bbfa381138e74fa5b5729dee5ac6764

The command list responsible for setting the texture addresses to 0 lives in physical address 0x2080A180. It doesn't appear to be a timing issue as the texture addresses are never replaced once the command list is stored in memory.

Dragios:
Partially fixed by #3097 
Only software rasterizer is fixed.

bamikhan:
Guys,has a solution to this been found out yet?

Issue #2010, opened 2016-08-18
Add a custom RTC feature to begin a game from a specific date/time


LCarvalhoUIS:
What I have in mind is a custom RTC setting, which lets you start a game from a specific date/time. That'd be extremely useful for clock manipulation in games. It'd be great for any game with some sort of clock related mechanic, and eventually make RNG manipulation easier in TAS/speedruns, as one'd be able to hit the same seed over and over.

I spoke to Dolphin devs on the same feature, and it helped immensely with seeding control once it was implemented. This is what their version looks like.

![image](https://cloud.githubusercontent.com/assets/1724823/17759322/e37067a0-64bb-11e6-9379-28e11abc9054.png)

valentinvanelslande:
Added in https://github.com/citra-emu/citra/pull/4043

PR #2878, opened 2017-08-13
Adding a system for providing Scripted Input

danzel:
Currently citra has a few small unit tests, and manual testing is also performed when changes are made.

It would be cool if we had a way to find regressions automatically, and keep a history of them long term.

My idea for this is that we can record a users inputs when they play a game, up until an issue occurs (bad textures, crash, etc).
The user can provide this log, and citra can replay it, saving screenshots as it goes.
This will provide a shorter fix and test cycle to developers, and save the manual work of retesting every game.

As a first step towards this, I've added a replacement for the buttons input that replays inputs from a script, resulting in exactly the same behaviour and resulting screenshots every time.

The screenshots will probably vary slightly by GPU, and between software/hardware rendering (based on device/driver differences and unimplemented/differently implemented features).

I've created a script [here](https://gist.github.com/danzel/2a5ad8895c81ede1aec9fb6cdea54c71) for Pokemon X (eu) that goes through the intro up until the black cheeks bug appears (#2133).

Follow up tasks:
Add a way to record user inputs.
Random number fixing. (#2010)
Allow scripting analog stick and touch screen.
Create a CI system like [FifoCI](https://dolphin-emu.org/blog/2015/01/25/making-developers-more-productive-dolphin-development-infrastructure/)

I'm on discord (danzel) and am happy to discuss this change :)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2878)


Dragios:
Review status: 0 of 18 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/scripted_input/script_runner.cpp, line 180 at r9](https://reviewable.io:443/reviews/citra-emu/citra/2878#-KrO5o7aQMnb0x3JCQVh:-KrO5o7aQMnb0x3JCQVi:bqxasb4) ([raw file](https://github.com/citra-emu/citra/blob/8c58f30d79d30a80ea2002f7e1f7d9e3e63d042b/src/scripted_input/script_runner.cpp#L180)):*
> unsigned char file_header[14] = {'B',
<details><summary><i alt="summary">Quoted 52 lines of code…</i></summary>
>                                      'M',
>                                      static_cast<unsigned char>(size),
>                                      static_cast<unsigned char>(size >> 8),
>                                      static_cast<unsigned char>(size >> 16),
>                                      static_cast<unsigned char>(size >> 24),
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      14 + 40,
>                                      0,
>                                      0,
>                                      0};
>     unsigned char info_header[40] = {40,
>                                      0,
>                                      0,
>                                      0,
>                                      static_cast<unsigned char>(w),
>                                      static_cast<unsigned char>(w >> 8),
>                                      static_cast<unsigned char>(w >> 16),
>                                      static_cast<unsigned char>(w >> 24),
>                                      static_cast<unsigned char>(h),
>                                      static_cast<unsigned char>(h >> 8),
>                                      static_cast<unsigned char>(h >> 16),
>                                      static_cast<unsigned char>(h >> 24),
>                                      1,
>                                      0,
>                                      24,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      static_cast<unsigned char>(image_bytes),
>                                      static_cast<unsigned char>(image_bytes >> 8),
>                                      static_cast<unsigned char>(image_bytes >> 16),
>                                      static_cast<unsigned char>(image_bytes >> 24),
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
>                                      0,
</details>
What's up with this weird formatting?

---

*[src/scripted_input/scripted_buttons.cpp, line 12 at r9](https://reviewable.io:443/reviews/citra-emu/citra/2878#-KrO5_N9JAGn84gaux-W:-KrO5_N9JAGn84gaux-X:b-2b2uo0) ([raw file](https://github.com/citra-emu/citra/blob/8c58f30d79d30a80ea2002f7e1f7d9e3e63d042b/src/scripted_input/scripted_buttons.cpp#L12)):*
> std::string button_name_to_index[] = {"a", "b", "x",     "y",      "up", "down", "left", "right",
>                                                                  "l", "r", "start", "select", "zl", "zr",   "home"};
Unnecessary spaces

---

*[src/scripted_input/scripted_input.h, line 16 at r9](https://reviewable.io:443/reviews/citra-emu/citra/2878#-KrO3TAqYMzQ-CnFXkFb:-KrO3TAqYMzQ-CnFXkFc:b-ke3toy) ([raw file](https://github.com/citra-emu/citra/blob/8c58f30d79d30a80ea2002f7e1f7d9e3e63d042b/src/scripted_input/scripted_input.h#L16)):*
> ```objc
> void LoadScript(std::string script_name);
> 
> /// Unresisters all build-in input device factories and shut them down.
> ```

Did you mean `Unregister`? Also it is built-in not build-in.
nit: Deregister sounds more suitable in this context
See https://english.stackexchange.com/a/154691

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2878)*

danzel:
Have fixed the scripted_input English, that was copy/paste from `input_common/main.h`, so I've fixed that up too.

Haven't touched the white space, that is what clang-format in the CI told me to change it to. If you can suggest a better way to lay those out that clang is happy with I'll change them.

danzel:
Just pushed one last little thing, an option to make the emulator close once the script has finished. That should be everything for now.
Adding touch support will need a bit of refactoring (right now we just reach over to the window to ask it for touch status), and I wanna save analog inputs for a future commit so this doesn't keep growing in size!

wwylele:
Haven't got time to thoroughly review this, but I don't like that it goes into the video_core. 3DS input system is not video-frame-based, but has its own modules (mostly in HID) and different updating period. So the input script system should likely hooks into HID module, or the core timing

danzel:
Thanks for the early feedback. I thought it was a bit ugly to put that hook in video_core, but I wasn't sure what else I could do.
I'll take a deeper look in to the HID/timing systems tomorrow.

src/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ add_subdirectory(video_core)
 add_subdirectory(audio_core)
 add_subdirectory(network)
 add_subdirectory(input_common)
+add_subdirectory(scripted_input)
```
j-selby: Nit: Don't like this name - doesn't fit well with the others. If other forms of automation are wanted, it might be better to call it something along those lines.

src/core/settings.cpp
```diff
@@ -34,6 +35,8 @@ void Apply() {
 
     Service::HID::ReloadInputDevices();
     Service::IR::ReloadInputDevices();
+
+    ScriptedInput::LoadScript(values.script_name, values.close_at_end);
```
j-selby: This doesn't handle null scripts?

src/input_common/main.h
```diff
@@ -11,7 +11,7 @@ namespace InputCommon {
 /// Initializes and registers all built-in input device factories.
 void Init();
 
-/// Unresisters all build-in input device factories and shut them down.
+/// Deregisters all built-in input device factories and shuts them down.
```
j-selby: Unrelated change. When you squash this, leave it to it's own commit.

src/scripted_input/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+            scripted_input.h
+            scripted_buttons.h
+            script_runner.h
+            )
+
+create_directory_groups(${SRCS} ${HEADERS})
+
+add_library(scripted_input STATIC ${SRCS} ${HEADERS})
+target_link_libraries(scripted_input PUBLIC common core)
+target_link_libraries(scripted_input PRIVATE glad SDL2)
```
j-selby: Missing endline.

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,198 @@
...
+// Refer to the license.txt file included.
+
+#include <SDL.h>
+#include <glad/glad.h>
+#include <string.h>
+#include "common/logging/log.h"
+#include "scripted_input/script_runner.h"
+#include "scripted_input/scripted_buttons.h"
+
+#define SCRIPT_MAX_LINE 200
```
j-selby: Seems a little low.

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,198 @@
...
+        if (close_at_end) {
+            SDL_Event sdlevent;
+            sdlevent.type = SDL_QUIT;
+            SDL_PushEvent(&sdlevent);
+        }
+    }
+}
+
+void ScriptRunner::SaveScreenshot() {
+    char buf[12];
```
j-selby: Citra comes with `libpng` - much nicer then `.bmp`s.

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,198 @@
...
+    sprintf(buf, "%i.bmp", frame_number);
+
+    const int w = 400; // render_window->GetActiveConfig().min_client_area_size.first;
```
j-selby: Is this hardcoded for a reason?

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+    FILE* fp = fopen(buf, "wb");
+    fwrite(file_header, sizeof(file_header), 1, fp);
+    fwrite(info_header, sizeof(info_header), 1, fp);
+    fwrite(pixels, w * h * 3, 1, fp);
+    fclose(fp);
+
+    delete[] pixels;
+}
+}
```
j-selby: Missing newline.

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,198 @@
...
+    }
+}
+
+void ScriptRunner::SaveScreenshot() {
+    char buf[12];
+    sprintf(buf, "%i.bmp", frame_number);
+
+    const int w = 400; // render_window->GetActiveConfig().min_client_area_size.first;
+    const int h = 480; // render_window->GetActiveConfig().min_client_area_size.second;
+    unsigned char* pixels = new unsigned char[w * h * 3];
```
j-selby: `std::vector`

src/scripted_input/README.md
```diff
@@ -0,0 +1,35 @@
...
+screenshot # screenshot of second menu fading out
+```
+
+Each line starts with a number, this is how many frames to run. It is then followed by a space separated list of buttons.
+
+Optionally a line maybe a have a comment, started with a `#` character, everything after that is ignored.
+
+If a line starts with `screenshot` then a screenshot will be saved at that time, the file is named `{framenumber}.bmp` and saved in the current working directory.
+
+**THE FOLLOWING ARE NOT YET IMPLEMENTED**
```
j-selby: This shouldn't exist here. `// TODO` it somewhere, and make a issue for it if you are going to leave it out.

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,198 @@
...
+namespace ScriptedInput {
+
+void ScriptRunner::SetButtons(std::shared_ptr<ScriptedButtons> buttons) {
+    scripted_buttons = buttons;
+}
+
+void ScriptRunner::LoadScript(std::string script_name, bool close) {
+    close_at_end = close;
+
+    FILE* file = fopen(script_name.c_str(), "r");
```
j-selby: Use https://github.com/citra-emu/citra/blob/master/src/common/file_util.cpp#L804

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+#define SCRIPT_MAX_LINE 200
+
+namespace ScriptedInput {
+
+void ScriptRunner::SetButtons(std::shared_ptr<ScriptedButtons> buttons) {
+    scripted_buttons = buttons;
+}
+
+void ScriptRunner::LoadScript(std::string script_name, bool close) {
```
j-selby: `close` is ambiguous - `close_on_exit`?

src/scripted_input/script_runner.h
```diff
@@ -0,0 +1,47 @@
...
+private:
+    std::vector<ScriptItem> script;
+    int frame_number{0};
+    int script_index{0};
+    int script_frame{0};
+
+    std::shared_ptr<ScriptedButtons> scripted_buttons;
+    bool close_at_end;
+
+    void SaveScreenshot();
```
j-selby: Does this need to be private?

src/scripted_input/scripted_buttons.cpp
```diff
@@ -0,0 +1,85 @@
...
+            button->status.store(true);
+        } else {
+            LOG_ERROR(
+                ScriptedInput,
+                "Button %s isn't mapped but is scripted, it should have engine:scripted,button:%s",
+                button_name_to_index[i].c_str(), button_name_to_index[i].c_str());
+        }
+    }
+}
+}
```
j-selby: Missing newline.

src/citra/config.cpp
```diff
@@ -152,6 +152,10 @@ void Config::ReadValues() {
     Settings::values.gdbstub_port =
         static_cast<u16>(sdl2_config->GetInteger("Debugging", "gdbstub_port", 24689));
 
+    // Scripted Input
```
j-selby: This should also be a CLI option, me thinks - config files are too hard to edit when quickly changing these around.

j-selby:
There is no implementation of this in the Qt frontend from what I can see as well?

src/scripted_input/script_runner.cpp
danzel: LIB_PNG seemed to be optional?
I'm building on windows and I don't have it.

src/citra/config.cpp
danzel: Yep, good plan.

src/core/settings.cpp
danzel: script_name defaults to "" in the config parser, so it shouldn't be null.

src/input_common/main.h
danzel: Yeah my bad. Removing now.

src/scripted_input/script_runner.cpp
danzel: IOFile doesn't have a method for reading lines (`fgets`), ok to add one?

src/scripted_input/script_runner.cpp
danzel: Hadn't figured out how to get access to that from there, fixing now.

src/scripted_input/script_runner.h
danzel: Doesn't need to be, however at the moment it uses frame_number for generating the file name, so it is only valid when ScriptRunner is enabled. (Otherwise all screenshots will be `0.bmp`)

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include "scripted_input/script_runner.h"
+#include "scripted_input/scripted_buttons.h"
+#include "video_core/video_core.h"
+
+#define SCRIPT_MAX_LINE 200
+
+namespace ScriptedInput {
+
+void ScriptRunner::SetButtons(std::shared_ptr<ScriptedButtons> buttons) {
+    scripted_buttons = buttons;
```
lioncash: `std::move(buttons);`

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include <SDL.h>
+#include <glad/glad.h>
+#include <string.h>
+#include "common/logging/log.h"
+#include "core/frontend/emu_window.h"
+#include "scripted_input/script_runner.h"
+#include "scripted_input/scripted_buttons.h"
+#include "video_core/video_core.h"
+
+#define SCRIPT_MAX_LINE 200
```
lioncash: Use concrete types instead of defines for constants.

src/scripted_input/scripted_buttons.cpp
```diff
@@ -0,0 +1,85 @@
...
+// Refer to the license.txt file included.
+
+#include <atomic>
+#include <vector>
+#include "core/settings.h"
+#include "scripted_input/scripted_buttons.h"
+
+const int button_count = 15;
+
+std::string button_name_to_index[] = {"a", "b", "x",     "y",      "up", "down", "left", "right",
```
lioncash: `std::array`, this could also be `const` as well.

src/scripted_input/scripted_input.cpp
```diff
@@ -0,0 +1,42 @@
...
+static ScriptRunner script_runner;
+
+void Init() {
+    scripted_buttons = std::make_shared<ScriptedInput::ScriptedButtons>();
+    Input::RegisterFactory<Input::ButtonDevice>("scripted", scripted_buttons);
+    script_runner.SetButtons(scripted_buttons);
+}
+
+void LoadScript(std::string script_name, bool close_at_end) {
+    if (script_name.length() > 0) {
```
lioncash:
```cpp
if (script_name.empty())
    return;
```

danzel:
This has gone through `clang-format`, it suggested the horrible array formatting :(
https://travis-ci.org/danzel/citra/jobs/263928275#L1982
I need a human suggestion for what it should look like please.

Agree ScriptedInput isn't a good name, need a suggestion on this too.

When a script is enabled, we override the configured buttons [ref](https://github.com/danzel/citra/commit/196a09fefa60f3c5c9f8b0a7286cd3e9a25e3f6d#diff-1880aac48bb9d9c28be07c57d88fd5e2R44)
So you only need to set the script in the config (As suggested this should become a command line option instead, then we'd require no config changes)

There is no implementation for the Qt frontend ATM, this was designed for automated testing rather than human use.

We could use something like the Movie system and [DTM file format](http://tasvideos.org/DTM.html) from dolphin. The files are not human editable however, so we'll need to implement a recorder too to make them usable.
If you want me to go down this route I can, I'll need a bit of help defining the format.

I haven't fixed a few of the bigger items above, I need some feedback on whether to progress with the system in this way, or look at implementing it like DTM/Movie in dolphin before I progress.
I feel like DTM/Movie would be better, but I'll need to touch a lot more of the emulator to implement it.
Thanks!

src/scripted_input/scripted_input.cpp
```diff
@@ -0,0 +1,42 @@
...
+// TODO: static std::shared_ptr<ScriptedAnalog> scripted_analog;
+static ScriptRunner script_runner;
+
+void Init() {
+    scripted_buttons = std::make_shared<ScriptedInput::ScriptedButtons>();
+    Input::RegisterFactory<Input::ButtonDevice>("scripted", scripted_buttons);
+    script_runner.SetButtons(scripted_buttons);
+}
+
+void LoadScript(std::string script_name, bool close_at_end) {
```
lioncash: `script_name` can be a const reference as far as I can tell. Also, you may want to consider using an enum class instead of a bool here, so call sites are more explicit as opposed to just passing `true` or `false` which is more ambiguous as to what the parameter does at first glance (and it prevents implicit integral conversions so you also guard against potentially passing in the wrong var as well).

src/scripted_input/script_runner.cpp
```diff
@@ -0,0 +1,200 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <SDL.h>
+#include <glad/glad.h>
+#include <string.h>
```
lioncash: `<cstring>`

src/scripted_input/scripted_buttons.cpp
```diff
@@ -0,0 +1,85 @@
...
+void ScriptedButtons::SetActiveButtons(const std::vector<int>& buttons_active) {
+
+    for (int i = 0; i < button_count; i++) {
+        auto button = scripted_button_list.get()->buttons[i];
+        if (button) {
+            button->status.store(false);
+        }
+    }
+
+    for (int i = 0; i < buttons_active.size(); i++) {
```
lioncash: `i` should be a `size_t` here.

danzel:
Thanks @lioncash! Haven't worked on a such a modern c++ codebase, so still getting used to all the new things.
Will definitely fix all those if we go ahead with this implementation.

src/core/settings.cpp
j-selby: The question was more does it handle those blank/empty values?

danzel: Yep, it checks if it is given a script_name before it initializes itself:
https://github.com/citra-emu/citra/pull/2878/files#diff-3b095790f1aa07be477bc9adb5434632R23

src/scripted_input/script_runner.cpp
j-selby: https://github.com/citra-emu/citra/blob/3e7459bbf9efa80b2d2dd45b892f7f3cef8ab751/src/common/file_util.cpp#L750

danzel: Yeah that'd work!

src/scripted_input/script_runner.cpp
j-selby: Then `#ifdef` it in/out depending on if it is available - I don't like the idea of hardcoding a image format here.

src/scripted_input/script_runner.cpp
j-selby: `ReadArray` (as used by `ReadFileToString`) is also a option.

src/scripted_input/script_runner.cpp
danzel: Or could I make libpng non-optional?
If you are making something that consumes these screenshots it'll be simpler if you don't have to handle two different file formats.

danzel:
I'm going to close this.
The more I think about it (and review the dolphin code), the more I think we should just go straight to a TAS capable system like their Movie system.
I'll take on board the comments above, thank you all for your feedback :)

PR #2880, opened 2017-08-15
Create shared font system archive if it is missing

B3n30:
With this PR it isn't necessary any more to dump the shared font archive. It will only write the shared font data if there isn't any data available (regarding the legacy method as well).

The font data are from [https://github.com/B3n30/citra_system_archives](https://github.com/B3n30/citra_system_archives). The generated shared font data isn't perfect. And might never really be perfect. The Nintendo specific symbols are still missing, due to copyright. It might be necessary to create our own symbols. Also the Chinese/Japanese/Korean symbols aren't tested, yet. I would love if someone could test these languages. Besides that some characters are slightly misplaced, caused by different sizes or placement of the used fonts. 

We probably should decide of a way to automatically update the shared_font.app.romfs.h on changes in the [https://github.com/B3n30/citra_system_archives](https://github.com/B3n30/citra_system_archives) (and maybe move that repo to citra-emu).

In a next step we could probably also provide the special shared_fonts for CHN/JPN/KOR, too.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2880)


FearlessTobi:
It would be nice to inform the user with pop-up that he should dump the original shared fonts from his 3ds. I think many users of Citra don't even know that they exist. Otherwise we could get many bug reports on missing chars and things like that.

src/core/file_sys/archive_ncch.cpp
```diff
@@ -82,6 +84,36 @@ ResultVal<std::unique_ptr<ArchiveBackend>> ArchiveFactory_NCCH::Open(const Path&
...
+                                  ErrorSummary::NotSupported, ErrorLevel::Permanent);
+            }
+            file->WriteBytes(SHARED_FONT_DATA, SHARED_FONT_DATA_len);
+            file->Close();
+            file.reset();
+            LOG_DEBUG(Service_FS, "Created shared font file.");
+            return RESULT_SUCCESS;
+        }
+    }
+
```
wwylele: Creating font archive shouldn't have anything to do with the _emulated 3DS file system API backend_ `Format`. You do realize that `Format` is a virtual function, right? Modifying it will change the behaviour when a game calls the frontend `Format` function. The archive format `ArchiveFactory_NCCH` (aka `0x2345678A`) doesn't support formatting on real 3DS and should always returns an error. In fact, 3DS system never format/create font archive, so this process in the emulator should not penetrate into any actual emulation stuff. It deserve its own function for special use.

B3n30:
This commit allows to download from another repo and thus makes it unnecessary to store the binary in the citra repo. It also sets the use of the custom system archives to off. Only the builds from travis and appveyor will be with use custom system archives enabled

bunnei:
@wwylele don't necessarily agree - I think the long term goal should be to allow Citra to be a complete replacement for the 3DS without requiring any hardware/technical knowledge. This is a common goal for many emulators in this scenario (HLE BIOS/etc. for ease of use for casual users, and LLE/more accurate implementation for users that demand more accuracy). With that in mind, if there are bugs in our shared font that cause users to complain, we should hear those complaints and improve it. I think actively impeding them from using this feature defeats the purpose of having it altogether.
IMO, a good solution would just be a one-time dialog strongly encouraging them to use a real shared font, and then just default to this one. Users that care about accuracy will do what they need, users that don't will ignore it - and any new bug reports we'll file bugs as appropriate.

wwylele:
@bunnei fair enough. I was probably concerned too much.

B3n30:
This warning will be shown, when the Shared Font is created: bildschirmfoto 2017-08-30 um 15 00 53. It will only be shown once. 
I know the code for creating the Shared Font isn't in good shape (especially the bcfnt.py), but I don't think it needs to be. It is mainly just there to prove, that the shared font has no copyright or license issues. 

I will soon release a new version of the shared font, that fixes the misplaced glyphes. I think we should delay the merge of this Pr until then. Maybe until then custom Nintendo Glyphes are added as well

From the telemetry data, it is visible that a missing shared font is the main reason for crashes. And people don't seem to fix this.

bunnei:
>From the telemetry data, it is visible that a missing shared font is the main reason for crashes. And people don't seem to fix this.

This just sounds like nonsense to me. I'm sure devs aren't sitting around going "hmm now what?". Collecting telemetry for this purpose sounds like a solution in search of a problem.

src/core/core.h
```diff
@@ -43,6 +43,10 @@ class System {
         ErrorUnknown                    ///< Any other error
     };
 
+    enum Integrity : u32 {
+        SharedFont = 1 << 0,
+    };
```
wwylele: If you are not going to serialize this to any external storage, you can just use a bunch of `bool`s instead of bit flags

src/core/core.cpp
```diff
@@ -197,4 +207,40 @@ void System::Shutdown() {
     LOG_DEBUG(Core, "Shutdown OK");
 }
 
+void System::SystemIntegrityCheck() {
+    system_integrity = static_cast<Integrity>(0);
+    // Shared Font
+    // TODO(B3N30): check/create font archive for region CHN/KOR/TWN
+    std::string nand_directory = FileUtil::GetUserPath(D_NAND_IDX);
+    const u32 shared_font_archive_id_low = 0x00014002;
+    const u32_le shared_font_archive_id_high = 0x0004009b;
```
wwylele: no need of `_le`

src/core/core.cpp
```diff
@@ -197,4 +207,40 @@ void System::Shutdown() {
...
+    std::string shared_font_path = Common::StringFromFormat(
+        "%s%s/title/%08x/%08x/content/00000000.app.romfs", nand_directory.c_str(), SYSTEM_ID,
+        shared_font_archive_id_high, shared_font_archive_id_low);
+    LOG_DEBUG(Core, "%s", shared_font_path.c_str());
```
wwylele: `"Checking shared font at %s"`

src/core/core.cpp
```diff
@@ -197,4 +207,40 @@ void System::Shutdown() {
...
+    auto file = std::make_shared<FileUtil::IOFile>(shared_font_path, "rb");
```
wwylele: You aren't sharing this object with others, so just make it a plan stack object.

src/core/core.cpp
```diff
@@ -197,4 +207,40 @@ void System::Shutdown() {
...
+    if (file->IsOpen()) {
+        file->Close();
+        LOG_INFO(Core, "SystemCheck: Shared Font exists.");
```
wwylele: We have `FileUtil::Exists` for checking existance

src/core/core.cpp
```diff
@@ -197,4 +207,40 @@ void System::Shutdown() {
...
+        std::string shared_font_legacy_path = FileUtil::GetUserPath(D_SYSDATA_IDX) + SHARED_FONT;
+        file = std::make_shared<FileUtil::IOFile>(shared_font_legacy_path, "rb");
+        if (file->IsOpen()) {
+            file->Close();
+            LOG_INFO(Core, "SystemCheck: Shared Font(legacy) exists.");
+        } else {
+            LOG_ERROR(Core, "SystemCheck: Shared Font missing.");
+            system_integrity = static_cast<Integrity>(system_integrity | Integrity::SharedFont);
+#ifdef ENABLE_CUSTOM_SYSTEM_ARCHIVES
+            file = std::make_shared<FileUtil::IOFile>(shared_font_path, "w+b");
```
wwylele: you are not reading the file, so just `"wb"`

src/citra_qt/main.cpp
```diff
@@ -421,6 +421,22 @@ bool GMainWindow::LoadROM(const QString& filename) {
         }
         return false;
     }
+    Core::System::Integrity integrity = system.GetSystemIntegrity();
+    if (integrity && Core::System::Integrity::SharedFont) {
```
wwylele: I am assuming you are doing bit-and operation, then it should be single `&`

BreadFish64:
[Hand crafted the relevant Nintendo symbols](https://github.com/B3n30/citra_system_archives/pull/2)
_feel the pressure B3n30_

src/citra_qt/main.cpp
```diff
@@ -421,6 +421,22 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
+#ifdef ENABLE_CUSTOM_SYSTEM_ARCHIVES
+        QMessageBox::critical(
+            this, tr("Shared Font Created."),
+            tr("The Shared Font is missing. A custom Shared Font was created. This could result in "
+               "some ugly or wrong glyphs. It is recomended to dump the Shared Font from your 3ds. "
```
atouchet: "3ds" should be capitalized here. I think this entire paragraph could be revised. I don't think it is entirely clear to the average user what this all means.

B3n30:
With the recent changes it will be necessary to add a line to the wiki explaining that the line `shared_font_jeu_version = dumped` is needed to not overwrite the dumped version with a custom one.

I will rebase and fix the conflicts when this PR is approved

src/core/core.cpp
```diff
@@ -197,4 +207,52 @@ void System::Shutdown() {
...
+#ifdef ENABLE_CUSTOM_SYSTEM_ARCHIVES
+        } else {
+            FileUtil::IOFile file(shared_font_path, "wb");
+            file.WriteBytes(SHARED_FONT_DATA, SHARED_FONT_DATA_len);
+            file.Close();
+            system_integrity.shared_font_JEU = Integrity::CreatedCustom;
+            Settings::values.shared_font_jeu_version = CUSTOM_ARCHIVES_VERSION;
+            LOG_WARNING(Core, "SystemCheck: Updated Shared Font.");
+        }
+#endif
```
baka0815: The #endif should be one line up - the closing bracket is wrong this way when not specifying ENABLE_CUSTOM_SYSTEM_ARCHIVES.

src/citra_qt/main.cpp
```diff
@@ -421,6 +421,29 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
         return false;
     }
+    Core::System::Integrity integrity = system.GetSystemIntegrity();
+    switch (integrity.shared_font_JEU) {
+    case Core::System::Integrity::CreatedCustom:
+#ifdef ENABLE_CUSTOM_SYSTEM_ARCHIVES
+        QMessageBox::critical(
+            this, tr("Shared Font Created."),
+            tr("A custom Shared Font was created. This could result in some ugly or wrong glyphs. "
+               "It is recomended to dump the Shared Font from your 3DS. "
```
BreadFish64: recommended

src/core/core.cpp
```diff
@@ -196,4 +206,68 @@ void System::Shutdown() {
...
+                                                  const void* custom_data,
+                                                  const size_t custom_data_length,
+                                                  const std::string& legacy_path = "") {
+    std::string nand_directory = FileUtil::GetUserPath(D_NAND_IDX);
+    std::string shared_font_path =
+        Common::StringFromFormat("%s%s/title/%08x/%08x/content/00000000.app.romfs",
+                                 nand_directory.c_str(), SYSTEM_ID, id_high, id_low);
+
+    if (FileUtil::Exists(shared_font_path)) {
+        std::string version = Settings::values.shared_font_jeu_version;
```
wwylele: I am assuming the function is going to be reused by other archives? Then the version string shouldn't be hardcoded to the font one. Is `settings_version` supposed to use here?

src/core/core.cpp
```diff
@@ -196,4 +206,68 @@ void System::Shutdown() {
     LOG_DEBUG(Core, "Shutdown OK");
 }
 
-} // namespace
+static System::Integrity::Status GetArchiveStatus(const std::string& name, const u32 id_high,
```
wwylele: Since the function also performs file writing operation in some cases, a `Get***` name is not suitable...

src/core/core.h
```diff
@@ -43,6 +43,18 @@ class System {
         ErrorUnknown                    ///< Any other error
     };
 
+    struct Integrity {
+        enum Status : u32 {
+            Dumped,
+            Custom,
+            CreatedCustom,
+            Missing,
```
wwylele: Better to have a little explanation to these, including what do they mean, in which case they will return and so on.

B3n30:
I'll close this PR for now, since it was a common opinion, that this should be part of the installer/updater and not part of citra. I'll create a new PR when/if the installer is merged

FearlessTobi:
Can this be now reopened/reworked because the installer is merged?

jroweboy:
@FearlessTobi The current idea is to add an "Install CIA" option in the UI and we can add the FOSS fonts and other system archives in a CIA to get them installed. That way the user can use the FOSS version but also can replace it with the actual archives dumped from a 3ds by the same method. One issue with shipping it with the installer is we would have two sources of truth for the "User Folder" location since this cant be shared. Any updates to this location (including allowing the user to select a new SD card location) would mess up adding it to the main installer.

FearlessTobi:
@jroweboy Okay, I think thats a pretty neat solution. But while we are at the topic of shipped files, could that be also used to ship some pre-defined Miis so you dont have to dump Mii Maker? (And regardless if it's possible to do that, is it even legal to ship them?)

PR #2882, opened 2017-08-15
Movie (Game Inputs) recording and playback

danzel:
Heavily inspired by the [Dolphin implementation](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/Movie.h).

Adds command lines to the SDL2 exe to control playback and recording. We should add something to the QT UI to match in the future.
(Is there somewhere these should be documented other than just in the `--help` response?)

Every time the hardware reads an input device (buttons, touch screen, accelerometer, gyroscope, c-stick) we copy (when recording) or replace (when playing) the values.
Once playback has finished, we release control of inputs so that the user can take control.

It is possible (likely even) that replays will desync between versions of Citra as behaviours change. This is expected.

File format will need documenting in the wiki. I've hopefully left enough padding room to populate the header with other fields we want in the future, I've brought over a minimal set to start with.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2882)

### DIFF
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -241,6 +241,7 @@ int main(int argc, char** argv) {
 
     if (!movie_record.empty() && !movie_play.empty()) {
         LOG_CRITICAL(Frontend, "Cannot both play and record a movie");
+        return -1;
     }
 
     log_filter.ParseFilterString(Settings::values.log_filter);
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -43,13 +43,16 @@
 #include "network/network.h"
 
 static void PrintHelp(const char* argv0) {
-    std::cout << "Usage: " << argv0 << " [options] <filename>\n"
-                                       "-g, --gdbport=NUMBER Enable gdb stub on port NUMBER\n"
-                                       "-i, --install=FILE    Installs a specified CIA file\n"
-                                       "-m, --multiplayer=nick:password@address:port"
-                                       " Nickname, password, address and port for multiplayer\n"
-                                       "-h, --help           Display this help and exit\n"
-                                       "-v, --version        Output version information and exit\n";
+    std::cout << "Usage: " << argv0
+              << " [options] <filename>\n"
+                 "-g, --gdbport=NUMBER Enable gdb stub on port NUMBER\n"
+                 "-i, --install=FILE    Installs a specified CIA file\n"
+                 "-m, --multiplayer=nick:password@address:port"
+                 " Nickname, password, address and port for multiplayer\n"
+                 "-r, --movie-record=[file]  Record a movie (game inputs) to the given file\n"
+                 "-p, --movie-play=[file]    Playback the movie (game inputs) from the given file\n"
+                 "-h, --help           Display this help and exit\n"
+                 "-v, --version        Output version information and exit\n";
 }
 
 static void PrintVersion() {
@@ -109,6 +112,9 @@ int main(int argc, char** argv) {
     int option_index = 0;
     bool use_gdbstub = Settings::values.use_gdbstub;
     u32 gdb_port = static_cast<u32>(Settings::values.gdbstub_port);
+    std::string movie_record;
+    std::string movie_play;
+
     char* endarg;
 #ifdef _WIN32
     int argc_w;
@@ -129,12 +135,13 @@ int main(int argc, char** argv) {
 
     static struct option long_options[] = {
         {"gdbport", required_argument, 0, 'g'},     {"install", required_argument, 0, 'i'},
-        {"multiplayer", required_argument, 0, 'm'}, {"help", no_argument, 0, 'h'},
+        {"multiplayer", required_argument, 0, 'm'}, {"movie-record", required_argument, 0, 'r'},
+        {"movie-play", required_argument, 0, 'p'},  {"help", no_argument, 0, 'h'},
         {"version", no_argument, 0, 'v'},           {0, 0, 0, 0},
     };
 
     while (optind < argc) {
-        char arg = getopt_long(argc, argv, "g:i:m:hv", long_options, &option_index);
+        char arg = getopt_long(argc, argv, "g:i:m:r:p:hv", long_options, &option_index);
         if (arg != -1) {
             switch (arg) {
             case 'g':
@@ -194,6 +201,12 @@ int main(int argc, char** argv) {
                 }
                 break;
             }
+            case 'r':
+                movie_record = optarg;
+                break;
+            case 'p':
+                movie_play = optarg;
+                break;
             case 'h':
                 PrintHelp(argv[0]);
                 return 0;
@@ -226,11 +239,17 @@ int main(int argc, char** argv) {
         return -1;
     }
 
+    if (!movie_record.empty() && !movie_play.empty()) {
+        LOG_CRITICAL(Frontend, "Cannot both play and record a movie");
+    }
+
     log_filter.ParseFilterString(Settings::values.log_filter);
 
     // Apply the command line arguments
     Settings::values.gdbstub_port = gdb_port;
     Settings::values.use_gdbstub = use_gdbstub;
+    Settings::values.movie_play = std::move(movie_play);
+    Settings::values.movie_record = std::move(movie_record);
     Settings::Apply();
 
     std::unique_ptr<EmuWindow_SDL2> emu_window{std::make_unique<EmuWindow_SDL2>()};
--- a/src/core/hle/service/ir/extra_hid.h
+++ b/src/core/hle/service/ir/extra_hid.h
@@ -6,6 +6,8 @@
 
 #include <array>
 #include <atomic>
+#include "common/bit_field.h"
+#include "common/swap.h"
 #include "core/frontend/input.h"
 #include "core/hle/service/ir/ir_user.h"
 
@@ -15,6 +17,22 @@ struct EventType;
 
 namespace Service {
 namespace IR {
+
+struct ExtraHIDResponse {
+    union {
+        BitField<0, 8, u32_le> header;
+        BitField<8, 12, u32_le> c_stick_x;
+        BitField<20, 12, u32_le> c_stick_y;
+    } c_stick;
+    union {
+        BitField<0, 5, u8> battery_level;
+        BitField<5, 1, u8> zl_not_held;
+        BitField<6, 1, u8> zr_not_held;
+        BitField<7, 1, u8> r_not_held;
+    } buttons;
+    u8 unknown;
+};
+static_assert(sizeof(ExtraHIDResponse) == 6, "HID status response has wrong size!");
 
 /**
  * An IRDevice emulating Circle Pad Pro or New 3DS additional HID hardware.
--- a/src/core/hle/service/ir/ir_rst.cpp
+++ b/src/core/hle/service/ir/ir_rst.cpp
@@ -2,29 +2,17 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include "common/bit_field.h"
 #include "core/core_timing.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/kernel/shared_memory.h"
 #include "core/hle/service/hid/hid.h"
 #include "core/hle/service/ir/ir_rst.h"
+#include "core/movie.h"
 #include "core/settings.h"
 
 namespace Service {
 namespace IR {
-
-union PadState {
-    u32_le hex{};
-
-    BitField<14, 1, u32_le> zl;
-    BitField<15, 1, u32_le> zr;
-
-    BitField<24, 1, u32_le> c_stick_right;
-    BitField<25, 1, u32_le> c_stick_left;
-    BitField<26, 1, u32_le> c_stick_up;
-    BitField<27, 1, u32_le> c_stick_down;
-};
 
 struct PadDataEntry {
     PadState current_state;
@@ -74,8 +62,10 @@ void IR_RST::UpdateCallback(u64 userdata, int cycles_late) {
     float c_stick_x_f, c_stick_y_f;
     std::tie(c_stick_x_f, c_stick_y_f) = c_stick->GetStatus();
     constexpr int MAX_CSTICK_RADIUS = 0x9C; // Max value for a c-stick radius
-    const s16 c_stick_x = static_cast<s16>(c_stick_x_f * MAX_CSTICK_RADIUS);
-    const s16 c_stick_y = static_cast<s16>(c_stick_y_f * MAX_CSTICK_RADIUS);
+    s16 c_stick_x = static_cast<s16>(c_stick_x_f * MAX_CSTICK_RADIUS);
+    s16 c_stick_y = static_cast<s16>(c_stick_y_f * MAX_CSTICK_RADIUS);
+
+    Movie::HandleIrRst(state, c_stick_x, c_stick_y);
 
     if (!raw_c_stick) {
         const HID::DirectionState direction = HID::GetStickDirectionState(c_stick_x, c_stick_y);
--- a/src/core/hle/service/ir/ir_rst.cpp
+++ b/src/core/hle/service/ir/ir_rst.cpp
@@ -65,7 +65,7 @@ void IR_RST::UpdateCallback(u64 userdata, int cycles_late) {
     s16 c_stick_x = static_cast<s16>(c_stick_x_f * MAX_CSTICK_RADIUS);
     s16 c_stick_y = static_cast<s16>(c_stick_y_f * MAX_CSTICK_RADIUS);
 
-    Movie::HandleIrRst(state, c_stick_x, c_stick_y);
+    Core::Movie::GetInstance().HandleIrRst(state, c_stick_x, c_stick_y);
 
     if (!raw_c_stick) {
         const HID::DirectionState direction = HID::GetStickDirectionState(c_stick_x, c_stick_y);
--- a/src/core/hle/service/ir/ir_rst.h
+++ b/src/core/hle/service/ir/ir_rst.h
@@ -6,6 +6,9 @@
 
 #include <atomic>
 #include <memory>
+#include "common/bit_field.h"
+#include "common/common_types.h"
+#include "common/swap.h"
 #include "core/frontend/input.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/service/service.h"
@@ -21,6 +24,18 @@ class EventType;
 
 namespace Service {
 namespace IR {
+
+union PadState {
+    u32_le hex{};
+
+    BitField<14, 1, u32_le> zl;
+    BitField<15, 1, u32_le> zr;
+
+    BitField<24, 1, u32_le> c_stick_right;
+    BitField<25, 1, u32_le> c_stick_left;
+    BitField<26, 1, u32_le> c_stick_up;
+    BitField<27, 1, u32_le> c_stick_down;
+};
 
 /// Interface to "ir:rst" service
 class IR_RST final : public ServiceFramework<IR_RST> {
--- a/src/core/movie.cpp
+++ b/src/core/movie.cpp
@@ -19,7 +19,9 @@
 #include "core/hle/service/ir/ir_rst.h"
 #include "core/movie.h"
 
-namespace Movie {
+namespace Core {
+
+/*static*/ Movie Movie::s_instance;
 
 enum class PlayMode { None, Recording, Playing };
 
@@ -116,25 +118,21 @@ struct CTMHeader {
 static_assert(sizeof(CTMHeader) == 256, "CTMHeader should be 256 bytes");
 #pragma pack(pop)
 
-static PlayMode play_mode = PlayMode::None;
-static std::vector<u8> recorded_input;
-static size_t current_byte = 0;
-
-static bool IsPlayingInput() {
+bool Movie::IsPlayingInput() {
     return play_mode == PlayMode::Playing;
 }
-static bool IsRecordingInput() {
+bool Movie::IsRecordingInput() {
     return play_mode == PlayMode::Recording;
 }
 
-static void CheckInputEnd() {
+void Movie::CheckInputEnd() {
     if (current_byte + sizeof(ControllerState) > recorded_input.size()) {
         LOG_INFO(Movie, "Playback finished");
         play_mode = PlayMode::None;
     }
 }
 
-static void Play(Service::HID::PadState& pad_state, s16& circle_pad_x, s16& circle_pad_y) {
+void Movie::Play(Service::HID::PadState& pad_state, s16& circle_pad_x, s16& circle_pad_y) {
     ControllerState s;
     std::memcpy(&s, &recorded_input[current_byte], sizeof(ControllerState));
     current_byte += sizeof(ControllerState);
@@ -163,7 +161,7 @@ static void Play(Service::HID::PadState& pad_state, s16& circle_pad_x, s16& circ
     circle_pad_y = s.pad_and_circle.circle_pad_y;
 }
 
-static void Play(Service::HID::TouchDataEntry& touch_data) {
+void Movie::Play(Service::HID::TouchDataEntry& touch_data) {
     ControllerState s;
     std::memcpy(&s, &recorded_input[current_byte], sizeof(ControllerState));
     current_byte += sizeof(ControllerState);
@@ -180,7 +178,7 @@ static void Play(Service::HID::TouchDataEntry& touch_data) {
     touch_data.valid.Assign(s.touch.valid);
 }
 
-static void Play(Service::HID::AccelerometerDataEntry& accelerometer_data) {
+void Movie::Play(Service::HID::AccelerometerDataEntry& accelerometer_data) {
     ControllerState s;
     std::memcpy(&s, &recorded_input[current_byte], sizeof(ControllerState));
     current_byte += sizeof(ControllerState);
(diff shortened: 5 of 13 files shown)


src/core/movie.cpp
```diff
@@ -0,0 +1,421 @@
...
+    }
+}
+
+void Shutdown() {
+    if (play_mode == PlayMode::Recording) {
+        LOG_INFO(Movie, "Saving movie");
+        FileUtil::IOFile save_record(Settings::values.movie_record, "wb");
+
+        CTMHeader header;
+        memset(&header, 0, sizeof(CTMHeader));
```
lioncash: You can just do `CTMHeader header = {};` to zero it out.

src/core/movie.cpp
```diff
@@ -0,0 +1,421 @@
...
+        header.filetype[1] = 'T';
+        header.filetype[2] = 'M';
+        header.filetype[3] = 0x1B;
+
+        Core::System::GetInstance().GetAppLoader().ReadProgramId(header.program_id);
+
+        std::string rev_bytes;
+        CryptoPP::StringSource(Common::g_scm_rev, true,
+                               new CryptoPP::HexDecoder(new CryptoPP::StringSink(rev_bytes)));
+        memcpy(header.revision, rev_bytes.data(), 20);
```
lioncash:
```cpp
memcpy(header.revision, rev_bytes.data(), sizeof(CTMHeader::revision));
```

src/core/movie.cpp
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    if (!Settings::values.movie_record.empty()) {
+        LOG_INFO(Movie, "Enabling Movie recording");
+        play_mode = PlayMode::Recording;
+    }
+}
+
+void Shutdown() {
+    if (play_mode == PlayMode::Recording) {
```
lioncash: You can do:

```cpp
if (!IsRecordingInput()) {
    return;
}
```

to unindent this block of code.

src/core/movie.cpp
```diff
@@ -0,0 +1,421 @@
...
+                // MAX_CIRCLEPAD_POS in hid.cpp fits in one byte
+                BitField<16, 8, u32> circle_pad_x;
+                BitField<24, 8, u32> circle_pad_y;
+            };
+        } PadAndCircle;
+
+        struct {
+            u16 x;
+            u16 y;
+            bool valid;
```
lioncash: Note that `bool` isn't required to have a fixed bit size in the standard, so you may want to use a different type if you're writing it out to disk.

src/core/movie.cpp
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    return true;
+}
+
+void Init() {
+    if (!Settings::values.movie_play.empty()) {
+        LOG_INFO(Movie, "Loading Movie for playback");
+        FileUtil::IOFile save_record(Settings::values.movie_play, "rb");
+        u64 size = {save_record.GetSize()};
```
lioncash: Why the braces?

src/citra/citra.cpp
```diff
@@ -55,6 +57,9 @@ int main(int argc, char** argv) {
     int option_index = 0;
     bool use_gdbstub = Settings::values.use_gdbstub;
     u32 gdb_port = static_cast<u32>(Settings::values.gdbstub_port);
+    std::string movie_record = "";
```
lioncash: `= "";` isn't necessary. `std::string` constructs as empty by default.

src/citra/citra.cpp
```diff
@@ -121,11 +131,17 @@ int main(int argc, char** argv) {
...
+    if (!movie_record.empty() && !movie_play.empty()) {
+        LOG_CRITICAL(Frontend, "Cannot both play and record a movie");
+    }
+
     log_filter.ParseFilterString(Settings::values.log_filter);
 
     // Apply the command line arguments
     Settings::values.gdbstub_port = gdb_port;
     Settings::values.use_gdbstub = use_gdbstub;
+    Settings::values.movie_play = movie_play;
```
lioncash: These strings can be `std::move`d.

src/core/hle/service/ir/ir_rst.h
```diff
@@ -9,6 +9,18 @@
 namespace Service {
 namespace IR {
 
+union PadState {
+    u32_le hex{};
+
+    BitField<14, 1, u32_le> zl;
```
lioncash: `common/bit_field.h` needs to be included. As does `common/common_types.h`

src/core/movie.cpp
```diff
@@ -0,0 +1,421 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <vector>
+#include <cryptopp/hex.h>
+#include "common/bit_field.h"
+#include "common/file_util.h"
```
lioncash: `<cstring>`, `<string>`, and `common/common_types.h` needs to be included.

src/core/hle/service/ir/ir_rst.cpp
```diff
@@ -12,23 +12,12 @@
...
+#include "core/movie.h"
 #include "core/settings.h"
 
 namespace Service {
 namespace IR {
 
-union PadState {
-    u32_le hex{};
-
-    BitField<14, 1, u32_le> zl;
```
lioncash: The `common/bit_field.h` header include can be removed.

src/core/movie.cpp
wwylele: and `std::memcpy`

danzel:
Thanks @lioncash, done everything except moving Movie in to a class.
I can do, just need a bit of advice on how it should look.
I think the public API (movie.h) stays as it does now, then inside we move all of the static variables and methods in to a new class that has a singleton instance created when `Init()` is called.
Sound right?

src/core/movie.cpp
```diff
@@ -0,0 +1,474 @@
...
+static void CheckInputEnd() {
+    if (current_byte + sizeof(ControllerState) > temp_input.size()) {
+        LOG_INFO(Movie, "Playback finished");
+        play_mode = PlayMode::None;
+    }
+}
+
+static void Play(Service::HID::PadState& pad_state, s16& circle_pad_x, s16& circle_pad_y) {
+    ControllerState s;
+    memcpy(&s, &temp_input[current_byte], sizeof(ControllerState));
```
wwylele: `std::memcpy`. Same for below

src/core/movie.cpp
```diff
@@ -0,0 +1,474 @@
...
+
+    LOG_INFO(Movie, "Saving movie");
+    FileUtil::IOFile save_record(Settings::values.movie_record, "wb");
+
+    CTMHeader header = {};
+
+    header.filetype[0] = 'C';
+    header.filetype[1] = 'T';
+    header.filetype[2] = 'M';
+    header.filetype[3] = 0x1B;
```
wwylele: "CTM\x1B` should be defined as a constant, shared with the one used in `ValidateHeader`

src/core/movie.h
```diff
@@ -0,0 +1,42 @@
...
+void HandleAccelerometerStatus(Service::HID::AccelerometerDataEntry& accelerometer_data);
+
+void HandleGyroscopeStatus(Service::HID::GyroscopeDataEntry& gyroscope_data);
+
+void HandleCStick(Service::IR::PadState& pad_state, s16& c_stick_x, s16& c_stick_y);
+
+void HandleCirclePad(Service::IR::CirclePadResponse& circle_pad);
+
+// TODO: Handle C-Stick
+}
```
wwylele: new line at the end

src/core/movie.h
```diff
@@ -0,0 +1,42 @@
...
+}
+
+namespace Movie {
+
+void Init();
+
+void Shutdown();
+
+void HandlePadAndCircleStatus(Service::HID::PadState& pad_state, s16& circle_pad_x,
```
wwylele: Need some documentation of what these `HandleSomething` do

danzel:
Thank you wwylele. Have implemented all of your feedback hopefully.

I renamed the fields to `ir_rst` and `extra_hid_response` to line them up with what the structs they are copying are called.

When making the `GetStickDirectionState` related change I discovered I was not storing the `circle_x/y` and `ir_rst.x/y` values correctly. So that was fixed also.


In dolphin they have a flag that controls if the system is trying to run in a deterministic mode.
ref: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/Core.cpp#L959

We could look to implement something similar to control parts of the engine that may be non-deterministic for performance.

lioncash:
> I think the public API (movie.h) stays as it does now

@danzel Note that by keeping it wholly namespace-based externally makes it more difficult to get to the point where more than one instance of the emulator can run in the same process (which is something that should be allowable from a library point of view, IMO). Ideally you'd be able to just create a new System object and be fine (the only reason it's a singleton right now is because other code and globals haven't been amended yet).

src/core/movie.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+    return true;
+}
+
+void Init() {
+    if (!Settings::values.movie_play.empty()) {
+        LOG_INFO(Movie, "Loading Movie for playback");
+        FileUtil::IOFile save_record(Settings::values.movie_play, "rb");
+        u64 size = save_record.GetSize();
+        save_record.Seek(0, SEEK_SET);
```
lioncash: `GetSize()` seeks back to the original position it was at before the call, in this case, so this Seek() is likely not required.

src/core/movie.cpp
```diff
@@ -0,0 +1,461 @@
...
+    if (!IsRecordingInput()) {
+        return;
+    }
+
+    LOG_INFO(Movie, "Saving movie");
+    FileUtil::IOFile save_record(Settings::values.movie_record, "wb");
+
+    CTMHeader header = {};
+
+    std::memcpy(header.filetype, header_magic_bytes, 4);
```
lioncash: You can make the filetype member initialize to the magic bits automatically as part of the initialization if you utilize `std::array`, like so:

```cpp
constexpr std::array<u8, 4> header_magic_bytes{{'C', 'T', 'M', 0x1B}};

#pragma pack(push, 1)
struct CTMHeader {
    std::array<u8, 4> filetype = header_magic_bytes;  // Unique Identifier (always "CTM"0x1B)
    u64 program_id;  // Also called title_id
    std::array<u8, 20> revision; // Git hash

    std::array<u8, 224> reserved; // Make heading 256 bytes, just because we can
};
static_assert(sizeof(CTMHeader) == 256, "CTMHeader should be 256 bytes");
#pragma pack(pop)
```

or alternatively the same as above can be done, minus the assignment in the struct and instead just do the assignment at the site it's needed at. like:

```cpp
...

CTMHeader header = {};
header.filetype = header_magic_bytes;

...
```

src/core/movie.cpp
```diff
@@ -0,0 +1,461 @@
...
+        play_mode = PlayMode::Recording;
+    }
+}
+
+void Shutdown() {
+    if (!IsRecordingInput()) {
+        return;
+    }
+
+    LOG_INFO(Movie, "Saving movie");
```
lioncash: You may want to move the saving code to a `SaveMovie()` function to keep the logic separate from the shutdown logic if it ever needs to change. You'll likely also want to check if the file status is still good (in case the attempt to open the file fails) before continuing onwards attempting to save data.

src/core/movie.cpp
```diff
@@ -0,0 +1,461 @@
...
+    s.extra_hid_response.zl_not_held.Assign(extra_hid_response.buttons.zl_not_held);
+    s.extra_hid_response.zr_not_held.Assign(extra_hid_response.buttons.zr_not_held);
+
+    temp_input.resize(current_byte + sizeof(ControllerState));
+    memcpy(&temp_input[current_byte], &s, sizeof(ControllerState));
+    current_byte += sizeof(ControllerState);
+}
+
+static bool ValidateHeader(const CTMHeader& header) {
+    if (std::memcmp(header_magic_bytes, header.filetype, 4) != 0) {
```
lioncash: this can be simplified to `if (header_magic_bits != header.filetype)` if `std::array` is used as mentioned in my other comments.

src/core/movie.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+    s.extra_hid_response.battery_level.Assign(extra_hid_response.buttons.battery_level);
+    s.extra_hid_response.c_stick_x.Assign(extra_hid_response.c_stick.c_stick_x);
+    s.extra_hid_response.c_stick_y.Assign(extra_hid_response.c_stick.c_stick_y);
+    s.extra_hid_response.r_not_held.Assign(extra_hid_response.buttons.r_not_held);
+    s.extra_hid_response.zl_not_held.Assign(extra_hid_response.buttons.zl_not_held);
+    s.extra_hid_response.zr_not_held.Assign(extra_hid_response.buttons.zr_not_held);
+
+    temp_input.resize(current_byte + sizeof(ControllerState));
+    memcpy(&temp_input[current_byte], &s, sizeof(ControllerState));
```
lioncash: This one's missing the `std::` prefix.

src/core/movie.cpp
```diff
@@ -0,0 +1,461 @@
...
+    IrRst,
+    ExtraHidResponse
+};
+
+#pragma pack(push, 1)
+struct ControllerState {
+    ControllerStateType type;
+
+    union {
+        struct {
```
lioncash: Should some of the `u16`/`s16` members have `_le` prefixes as well?

src/core/movie.cpp
```diff
@@ -0,0 +1,461 @@
...
+};
+static_assert(sizeof(ControllerState) == 7, "ControllerState should be 7 bytes");
+#pragma pack(pop)
+
+const u8 header_magic_bytes[4] = {'C', 'T', 'M', 0x1B};
+
+#pragma pack(push, 1)
+struct CTMHeader {
+    u8 filetype[4];  // Unique Identifier (always "CTM"0x1B)
+    u64 program_id;  // Also called title_id
```
lioncash: Should this be a `u64_le`?

src/core/movie.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+#include <cstring>
+#include <string>
+#include <vector>
+#include <cryptopp/hex.h>
+#include "common/bit_field.h"
+#include "common/common_types.h"
+#include "common/file_util.h"
+#include "common/logging/log.h"
+#include "common/scm_rev.h"
```
lioncash: `common/swap.h` needs to be included for the `_le`/`_be` types.

src/core/hle/service/ir/ir_rst.h
```diff
@@ -4,11 +4,25 @@
 
 #pragma once
 
+#include "common/bit_field.h"
+#include "common/common_types.h"
```
lioncash: `common/swap.h` needs to be included for the `_le`/`_be` types.

src/core/hle/service/ir/extra_hid.h
```diff
@@ -6,12 +6,29 @@
 
 #include <array>
 #include <atomic>
+#include "common/bit_field.h"
```
lioncash: `common/swap.h` needs to be included for the `_le`/`_be` types.

src/core/movie.cpp
```diff
@@ -0,0 +1,461 @@
...
+        }
+    }
+
+    if (!Settings::values.movie_record.empty()) {
+        LOG_INFO(Movie, "Enabling Movie recording");
+        play_mode = PlayMode::Recording;
+    }
+}
+
+void Shutdown() {
```
lioncash: shouldn't this function, as the API currently is, reset `current_byte` to `0`?

src/core/movie.cpp
danzel: ReadProgramId is: `virtual ResultStatus ReadProgramId(u64& out_program_id)`

danzel:
Checked over the s16/u16, they don't need `_le`, only `extra_hid_response` has it in its source.

Thanks for the `std::array` advice, that's much nicer, used them all :)

I agree this should be an object living inside system for the reasons you've mentioned.
Right now most of the parts of the system are violating this though I think.
https://github.com/danzel/citra/blob/912b3cbdbe99053956228751437f553ca0e381a6/src/core/core.cpp#L151-L157
Happy to work on this in the future as a separate item.

danzel:
Moved the duplicated code from the end of the Record methods into another method to get rid of the duplication and Squashed the commits.

Can see the (mostly original) history here:
https://github.com/danzel/citra/commits/movie-squash-no

Tried making it flush the inputs to disk every time we record one, but (as expected) it completely kills framerate. This would be nice for recording lockups, may need revisiting in the future.

B3n30:
What's the status of this PR?

danzel:
Rebased to fix merge issue, should be good to go again once the builds run.

src/core/movie.cpp
wwylele: > Checked over the s16/u16, they don't need _le, only extra_hid_response has it in its source.

This is not how it works. All members specified for a file format should be tagged with endianess _for platform compatibility_ (as you wrote in the comment), which is independent from the endianess of the data source.

danzel:
Oh I see! ok will fix that shortly, thanks.

danzel:
Ok, endianness specified :+1:

Subv:
@wwylele Async file IO doesn't have to be completely non-deterministic, we can provide a fixed (and deterministic for each file operation) time window to perform the IO operation and block the emulation thread if it takes longer than that.

I'll give this a review either today or tomorrow.

Subv:
@danzel Does the recording and playback happen per-frame? Could it be possible to implement a playback system in a real 3DS using the same movie files this PR accepts?

danzel:
Recording and playback happens at the hardware read rate, which is more than once per frame.
https://github.com/danzel/citra/blob/a6cd3a9e5ddc537c519120ebe8418659a057ff3f/src/core/hle/service/hid/hid.cpp#L50-L53

We could change it to be once per frame, but that would require changing all of the controller input to be updated once per frame. SDL does this currently for keyboard/mouse inputs (not joystick). QT does not do this at all.
I considered changing all user controller inputs to be updated once per frame, but wylele didn't like it, so I've left it as it is. I don't think it is important to change.

As for playing these back on Hardware... I've done close to zero 3ds stuff, so I'm not sure.
Assuming we can hook the relevant sections that read the hardware and redirect them to get the values from this file, then yeah it should be possible.
I can look at this in the future if someone can point me in the right direction.

Fixing the clang-format now...

src/core/movie.cpp
```diff
@@ -0,0 +1,470 @@
...
+    pad_state.y.Assign(s.pad_and_circle.y);
+
+    circle_pad_x = s.pad_and_circle.circle_pad_x;
+    circle_pad_y = s.pad_and_circle.circle_pad_y;
+}
+
+static void Play(Service::HID::TouchDataEntry& touch_data) {
+    ControllerState s;
+    std::memcpy(&s, &temp_input[current_byte], sizeof(ControllerState));
+    current_byte += sizeof(ControllerState);
```
Subv: Why does this copy and skip the entirety of ControllerState if only TouchData was requested?

Is the current system dependent on the update order of the different devices?

src/core/movie.cpp
```diff
@@ -0,0 +1,470 @@
...
+}
+
+static bool ValidateHeader(const CTMHeader& header) {
+    if (header_magic_bytes != header.filetype) {
+        LOG_ERROR(Movie, "Playback file does not have valid header");
+        return false;
+    }
+
+    std::string revision;
+    CryptoPP::StringSource ss(header.revision.data(), sizeof(header.revision), true,
```
Subv: There's an ArrayToString function that doesn't depend on CryptoPP in Common.

src/core/movie.cpp
```diff
@@ -0,0 +1,470 @@
...
+    u64_le program_id;           // Also called title_id
+    std::array<u8, 20> revision; // Git hash
+
+    std::array<u8, 224> reserved; // Make heading 256 bytes, just because we can
+};
+static_assert(sizeof(CTMHeader) == 256, "CTMHeader should be 256 bytes");
+#pragma pack(pop)
+
+static PlayMode play_mode = PlayMode::None;
+static std::vector<u8> temp_input;
```
Subv: Surely you can come up with a better name for this variable

src/core/movie.cpp
```diff
@@ -0,0 +1,470 @@
...
+        } extra_hid_response;
+    };
+};
+static_assert(sizeof(ControllerState) == 7, "ControllerState should be 7 bytes");
+#pragma pack(pop)
+
+constexpr std::array<u8, 4> header_magic_bytes{{'C', 'T', 'M', 0x1B}};
+
+#pragma pack(push, 1)
+struct CTMHeader {
```
jroweboy: Is there any advantage to having a fixed size header instead of a variable sized header with a fixed "start" location. So for example:

```cpp
std::array<u8, 20> revision;
u32 start_offset;
```

then we shouldn't have to worry as much about running out of header space since each movie file could potentially have a different start location.

(Note this isn't much of a review concern as it is a review comment)

src/core/movie.cpp
```diff
@@ -0,0 +1,470 @@
...
+    std::array<u8, 4> filetype;  // Unique Identifier (always "CTM"0x1B)
```
jroweboy: triple slash to make it a doc comment (and maybe make them a little less silly :D)

src/core/movie.cpp
danzel: ControllerState is a union of all of the different inputs recordings we take, so it will be a `touch` or a `pad_and_circle` etc, it doesn't have both (if I got my `union`ing correct).

The system is dependent on the update order of the different devices. It will break if this ever changes.
However this should only change if we find the update rate is different than we have specified (the update rate is based on hardware tests)

src/core/movie.cpp
danzel: But dolphin.... https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/Movie.cpp#L73
hahaha will rename :)

src/core/movie.cpp
danzel: Nah, could change this. No big advantage either way IMHO.
Variable is a bit more complicated, fixed wastes a little space.

danzel:
Implemented latest feedback and resquashed.
Individual latest commit is here https://github.com/danzel/citra/commit/06f98fe2c6b5244a25cc93d9878c633fa6597194

bunnei:
@danzel thanks for proposing this change. This PR is getting a bit stale - we need to decide if we'd like to get this merged, or if it's not something that is of value to us and should be abandoned. Personally, I'm not really familiar with Dolphin's (or other emulator's) ability to record game inputs, and what it might be useful for. Can other folks on the team comment on this, and/or review this PR if this feature is something we should adopt?

danzel:
It's has been suggested I copy over some propoganda from discord, so...

image

image
The implementation is as close to dolphin as I could get it

image
Absolutely something we will need to think about once we get to that part, a bit outside of the scope of this PR however.

image

image
Currently we record based on how often hardware is polled.
We can change this in the future if it turns out to not be a good way to do it (We can't do it frame-based as we can update inputs in the middle of a frame, but we could record timestamps instead).

image

I'll look at rebasing this again-again this weekend.

danzel:
Some notes for comparison with dolphin:

[Movie.cpp](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/Movie.cpp)
Our Movie.cpp is based on the same idea of dolphins. We have a big buffer we write to (recording) or read from (playback).
Dolphins file is quite a bit bigger because it has methods in their for showing the current inputs and a bunch of stuff we don't need yet (they probably shouldn't be in that file imho)

Every time the emulator updates the state of the controller inputs (GC / Wiimote) it does the normal update (ask SDL for the buttons held etc), then calls in to the movie system so it can (when enabled) replace the state with the recorded one, or copy it in to the recording.

These are hooked in each place an input is read:
https://github.com/dolphin-emu/dolphin/blob/d89698898abeda018dc2b8896f00bc46f036f36b/Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp#L810
https://github.com/dolphin-emu/dolphin/blob/38f44b2839d1cb3792f1bf517efe4a092933c43e/Source/Core/Core/HW/SI/SI_DeviceGCController.cpp#L163

They support network inputs and some other thigs which we don't, so again theirs is a bit bigger.
However we have multiple input types (Pad, Touch Screen, C-Pad, Circle Pad, Accel/Gyro iirc), so we need to handle recording all of these different types of inputs.

src/core/movie.cpp
```diff
@@ -0,0 +1,469 @@
...
+static void CheckInputEnd() {
+    if (current_byte + sizeof(ControllerState) > recorded_input.size()) {
+        LOG_INFO(Movie, "Playback finished");
+        play_mode = PlayMode::None;
+    }
+}
+
+static void Play(Service::HID::PadState& pad_state, s16& circle_pad_x, s16& circle_pad_y) {
+    ControllerState s;
+    std::memcpy(&s, &recorded_input[current_byte], sizeof(ControllerState));
```
B3n30: Assert or check for CheckInputEnd before copy?

src/core/movie.cpp
```diff
@@ -0,0 +1,469 @@
...
+}
+
+void Shutdown() {
+    if (!IsRecordingInput()) {
+        return;
+    }
+
+    SaveMovie();
+
+    play_mode = PlayMode::None;
```
B3n30: Those values should probably get reseted independently of IsRecordingInput(). So move that in front of the check

danzel: :+1:

src/core/movie.cpp
danzel: We check and update after every Play call (see Handle template), but adding an assert before calling Play too. (Could happen if given an invalid file to playback)

src/citra/citra.cpp
```diff
@@ -121,11 +131,17 @@ int main(int argc, char** argv) {
         return -1;
     }
 
+    if (!movie_record.empty() && !movie_play.empty()) {
+        LOG_CRITICAL(Frontend, "Cannot both play and record a movie");
```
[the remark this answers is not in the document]
danzel: Yep!

danzel:
Rebased and fixed things based on B3n30s feedback above.

jroweboy:
After several months, more than enough reviews, no strong opposition, many changes made in the name of bikeshedding, lets just click the merge button already!

PR #2884, opened 2017-08-17
gl_rasterizer: add clipping plane z<=0 defined in PICA

### LINKED ISSUE #2248: Pokémon Omega Ruby/Alpha Saphire ‒ Objects close to camera block view
Only when using *Hardware Rendering* displays the wrong behaviour

Left side is from **Nintendo 3DS** and right side is using the **Citra** emulator
May's room

Professor Birch's Lab 1

Professor Birch's Lab 2

Nothing useful in log (continuous spamming with this):
```
Service.GSP <Warning> core\hle\service\gsp_gpu.cpp:SetAxiConfigQoSMode:349: (STUBBED) called mode=0x00000203
Service.GSP <Warning> core\hle\service\gsp_gpu.cpp:SetAxiConfigQoSMode:349: (STUBBED) called mode=0x00000205
```

Switching to *Software Renderer* the view is corrected but the text is garbled somehow (that is another issue by the way).
Correct view

Again nothing useful in log for software renderer:
```
Service.GSP <Warning> core\hle\service\gsp_gpu.cpp:SetAxiConfigQoSMode:349: (STUBBED) called mode=0x00000203
Service.GSP <Warning> core\hle\service\gsp_gpu.cpp:SetAxiConfigQoSMode:349: (STUBBED) called mode=0x00000205
Render.OpenGL <Error> video_core\renderer_opengl\renderer_opengl.cpp:DebugHandler:483: API ERROR 1282: GL_INVALID_OPERATION error generated. Texture name does not refer to a texture object generated by OpenGL.
```
`Render.OpenGL` log shown 12 times before restart back to `Service.GSP`
Not sure if that is useful in any manner but I mention it anyway

Tested this on commit 3174bfd
**OS:** _Windows 10 build 14393.447 (codename Redstone 1)_
**GPU:** _nVidia GT 610 2GB DDR3 (Asus)_
**Driver:** _Version 375.95_
**OpenGL:** _Version 4.5_
_________
Save file included to reach the scene quickly
[Scene 1.zip](https://github.com/citra-emu/citra/files/1219365/Scene.1.zip) - *Just walk up the stairs to proceed*
[Scene 2.zip](https://github.com/citra-emu/citra/files/1219366/Scene.2.zip) - *Just enter the Professor's Lab to proceed*

wwylele:
Fixes #2248. A related hwtest can be found there below

OpenGL convention for depth range in the normalized device coordinate is [-1,1], which is different from the one of PICA [-1, 0]. We do correct transformation in the fragment shader, but didn't take the clipping plane difference into account. In the clip space, OpenGL clips vertices against -w<=z<=w, while PICA clips them against -w<=z<=0. This is correctly implemented in the swrasterizer/clipper. In order to achieve the same result in GL rasterizer, a user-defined clipping plane is added at z<=0. 

I reserved another GL user-defined clipping plane, for future implementation for PICA user-defined clipping plane.

The other three commits are some no-op clean-up. See commit message for detail

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2884)

### DIFF
--- a/src/video_core/renderer_opengl/gl_state.cpp
+++ b/src/video_core/renderer_opengl/gl_state.cpp
@@ -68,6 +68,8 @@ OpenGLState::OpenGLState() {
     draw.vertex_buffer = 0;
     draw.uniform_buffer = 0;
     draw.shader_program = 0;
+
+    clip_distance = {};
 }
 
 void OpenGLState::Apply() const {
@@ -259,6 +261,17 @@ void OpenGLState::Apply() const {
     // Shader program
     if (draw.shader_program != cur_state.draw.shader_program) {
         glUseProgram(draw.shader_program);
+    }
+
+    // Clip distance
+    for (size_t i = 0; i < clip_distance.size(); ++i) {
+        if (clip_distance[i] != cur_state.clip_distance[i]) {
+            if (clip_distance[i]) {
+                glEnable(GL_CLIP_DISTANCE0 + i);
+            } else {
+                glDisable(GL_CLIP_DISTANCE0 + i);
+            }
+        }
     }
 
     cur_state = *this;
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -28,6 +28,9 @@ MICROPROFILE_DEFINE(OpenGL_Blits, "OpenGL", "Blits", MP_RGB(100, 100, 255));
 MICROPROFILE_DEFINE(OpenGL_CacheManagement, "OpenGL", "Cache Mgmt", MP_RGB(100, 255, 100));
 
 RasterizerOpenGL::RasterizerOpenGL() : shader_dirty(true) {
+    // Clipping plane 0 is always enabled for PICA fixed clip plane z <= 0
+    state.clip_distance[0] = true;
+
     // Create sampler objects
     for (size_t i = 0; i < texture_samplers.size(); ++i) {
         texture_samplers[i].Create();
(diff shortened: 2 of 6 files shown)


danzel:
Have built and verified the incorrect and correct behavior based on your hwtest. :+1:

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -68,6 +68,9 @@ OpenGLState::OpenGLState() {
     draw.vertex_buffer = 0;
     draw.uniform_buffer = 0;
     draw.shader_program = 0;
+
+    for (bool& clip : clip_distance)
```
lioncash: `clip_distance.fill(false);` or `clip_distance = {};`, or just declaring it in the body of the class as: `std::array<bool, 2> clip_distance{};`

wwylele: Fixed. I chose to keep the initialization here to keep it together with the comment above:
`// These all match default OpenGL values`

yuriks:
(Note: Zc = clip space, Zd = NDC (post-divide), Zw = window space)

So, regarding the -z that was in the shader, I did some archeology, and the reason that's there is that before depth scale support was implemented in #1621, this was necessary to get correct depth values. The PICA behavior is that clipping is done as -1 <= Zd <= 0, and the window mapping is Zw = scale * Zd + offset. This means that scale usually needs to be set to a negative value, so that the Zd in range [-1, 0] is flipped and mapped to [0, 1]. In OpenGL, we still received Zc as a negative value, and since the depth scale wasn't applied, it had to be negated so that the depth functions worked properly.

However, if we want to optionally use ARB_clip_control instead of the user clipping plane, we could still keep negating it to get the proper clipping "for free". We may or may not do that, so it's up to you if you want to care about it or not.

PR #2888, opened 2017-08-19
Fixed some warnings in the core project.

Subv:
Visual Studio 2017 15.3 reports 137 warnings in total (many of which come from externals) and 67 warnings in the ```core``` project as of 21204ba488fd00be4cc6e44998c5d7c8bd0fec91, this PR fixes 13 of those warnings, bringing down the number of ```core``` warnings to 54. This PR does not touch the other projects or the warning settings in cmake.

ref #136

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2888)

### DIFF
--- a/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
@@ -759,7 +759,7 @@ static ThumbDecodeStatus DecodeThumbInstruction(u32 inst, u32 addr, u32* arm_ins
     ThumbDecodeStatus ret = TranslateThumbInstruction(addr, inst, arm_inst, inst_size);
     if (ret == ThumbDecodeStatus::BRANCH) {
         int inst_index;
-        int table_length = arm_instruction_trans_len;
+        int table_length = static_cast<int>(arm_instruction_trans_len);
         u32 tinstr = GetThumbInstruction(inst, addr);
 
         switch ((tinstr & 0xF800) >> 11) {
--- a/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
+++ b/src/core/arm/dyncom/arm_dyncom_interpreter.cpp
@@ -838,7 +838,7 @@ static unsigned int InterpreterTranslateInstruction(const ARMul_State* cpu, cons
     return inst_size;
 }
 
-static int InterpreterTranslateBlock(ARMul_State* cpu, int& bb_start, u32 addr) {
+static int InterpreterTranslateBlock(ARMul_State* cpu, std::size_t& bb_start, u32 addr) {
     MICROPROFILE_SCOPE(DynCom_Decode);
 
     // Decode instruction, get index
@@ -871,7 +871,7 @@ static int InterpreterTranslateBlock(ARMul_State* cpu, int& bb_start, u32 addr) 
     return KEEP_GOING;
 }
 
-static int InterpreterTranslateSingle(ARMul_State* cpu, int& bb_start, u32 addr) {
+static int InterpreterTranslateSingle(ARMul_State* cpu, std::size_t& bb_start, u32 addr) {
     MICROPROFILE_SCOPE(DynCom_Decode);
 
     ARM_INST_PTR inst_base = nullptr;
@@ -1620,7 +1620,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     unsigned int addr;
     unsigned int num_instrs = 0;
 
-    int ptr;
+    std::size_t ptr;
 
     LOAD_NZCVT;
 DISPATCH : {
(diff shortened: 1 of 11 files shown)


src/core/hw/gpu.h
```diff
@@ -74,9 +74,9 @@ struct Regs {
         case PixelFormat::RGB5A1:
         case PixelFormat::RGBA4:
             return 2;
-        default:
-            UNIMPLEMENTED();
         }
+
```
jroweboy: neobrain mentioned this on IRC (and I agree) that we should range check the inputs. Its probably better to range change and assert and not just throw `UNREACHABLE` there in case of invalid values.

src/core/hle/kernel/kernel.h
```diff
@@ -84,6 +85,8 @@ class Object : NonCopyable {
         case HandleType::ClientSession:
             return false;
         }
+
```
jroweboy: ditto

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -759,7 +759,7 @@ static ThumbDecodeStatus DecodeThumbInstruction(u32 inst, u32 addr, u32* arm_ins
     ThumbDecodeStatus ret = TranslateThumbInstruction(addr, inst, arm_inst, inst_size);
     if (ret == ThumbDecodeStatus::BRANCH) {
         int inst_index;
```
Subv: I tried this before, but it generated even more warnings when calling the callbacks because they take int instead of size_t, and changing all the callbacks doesn't sound like a good idea.

jroweboy: Thats fine then.

src/core/hw/gpu.h
jroweboy: after a lengthy discussion on IRC, `UNREACHABLE` seems to be good enough for this use case as we've already checked all enumeration values.

src/core/frontend/emu_window.cpp
```diff
@@ -78,7 +78,7 @@ void EmuWindow::AccelerometerChanged(float x, float y, float z) {
 void EmuWindow::GyroscopeChanged(float x, float y, float z) {
     constexpr float FULL_FPS = 60;
     float coef = GetGyroscopeRawToDpsCoefficient();
-    float stretch = Core::System::GetInstance().perf_stats.GetLastFrameTimeScale();
```
wwylele: Covered by #2861. Please fix the conflict

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -838,7 +838,7 @@ static unsigned int InterpreterTranslateInstruction(const ARMul_State* cpu, cons
     return inst_size;
 }
 
-static int InterpreterTranslateBlock(ARMul_State* cpu, int& bb_start, u32 addr) {
+static int InterpreterTranslateBlock(ARMul_State* cpu, size_t& bb_start, u32 addr) {
```
wwylele: `std::size_t` (since you get this correct for the first one).

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -871,7 +871,7 @@ static int InterpreterTranslateBlock(ARMul_State* cpu, int& bb_start, u32 addr)
     return KEEP_GOING;
 }
 
-static int InterpreterTranslateSingle(ARMul_State* cpu, int& bb_start, u32 addr) {
+static int InterpreterTranslateSingle(ARMul_State* cpu, size_t& bb_start, u32 addr) {
```
wwylele: ditto

src/core/arm/dyncom/arm_dyncom_interpreter.cpp
```diff
@@ -1620,7 +1620,7 @@ unsigned InterpreterMainLoop(ARMul_State* cpu) {
     unsigned int addr;
     unsigned int num_instrs = 0;
 
-    int ptr;
+    size_t ptr;
```
wwylele: ditto

src/core/arm/skyeye_common/armstate.h
```diff
@@ -230,7 +230,7 @@ struct ARMul_State final {
 
     // TODO(bunnei): Move this cache to a better place - it should be per codeset (likely per
     // process for our purposes), not per ARMul_State (which tracks CPU core state).
-    std::unordered_map<u32, int> instruction_cache;
+    std::unordered_map<u32, size_t> instruction_cache;
```
wwylele: ditto

jroweboy:
Last call for reviews. If you would like to review this, please leave a comment, otherwise I'll merge this tomorrow.

mailwl:
@jroweboy, looks good. +1 for merge

PR #2889, opened 2017-08-20
Updated master logo to new logo svg

Schplee:
This will finally updated the outdated Nightly logo to coincide with the new logo. @bunnei signed off on it.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2889)


dist/citra.svg
```diff
@@ -1,80 +1,2 @@
...
...092-0.45-0.369-1.8-0.736-3.599-0.185-0.9-0.216-0.915-0.436-1.79c-2.367-9.534-5.355-18.932-9.077-28.033-3.717-9.105-8.196-17.911-13.474-26.221-5.276-8.31-11.358-16.118-18.197-23.206-6.832-7.094-14.416-13.459-22.563-18.989-8.15-5.528-16.852-10.229-25.9-14.096-9.054-3.857-18.44-6.901-27.983-9.247-9.54-2.354-19.27-3.997-29.075-4.737-9.804-0.741-19.683-0.576-29.444 0.61-9.761 1.181-19.397 3.381-28.694 6.565-9.3 3.177-18.256 7.329-26.737 12.275-8.488 4.941-16.474 10.71-23.972 17.028-7.497 6.325-14.5 13.22-21.044 20.512z" fill="url(#linearGradient7248-9)"/></g></g></g><path d="m853.83 340.91z" fill="none" stroke="#000" stroke-width="1px"/><path d="m512.59 583.25c24.074 19.047 61.695 18.896 102.99 7.7902 41.299-11.105 86.278-33.164 124.63-62.438 43.462-33.178 74.436-61.06 97.049-98.852 10.211-17.065 16.881-32.679 20.405-47.4 3.5243-14.721 2.1355-27.046-3.4171-40.626 0 0 22.052 31.822 28.025 81.607 2.9864 24.893 1.4477 59.407-11.26 93.337-11.843 31.622-27.743 61.223-54.874 85.55-1.0301 0.92357-12.207 12.478-32.938 24.374-25.596 14.688-46.884 22.5-75.534 27.634-34.926 6.2583-59.469 4.4522-85.833-1.8111-36.815-8.7462-71.794-28.542-110.24-69.041" fill="url(#radialGradient7556)"/></g></g></svg>
```
Dragios: Single line long string of text?
You need to format the text allignment better.

mailwl: don't thing svg need to be "read"

Schplee:
I agree with @mailwl. The XML doesn't need to be easy to read, and it's simply generated from an inkscape export.

j-selby:
Strongly disagree. Compress, if you need to, with cmake or whatever if this is going into binaries, but text files in a source code repository shouldn't be stored like that.

jroweboy:
Just for reference, I did some research into using CMake to generate the icons. I found https://github.com/arx/ArxLibertatis/blob/7854561fe7444955f8184c22a2393b5e970c69ff/cmake/Icons.cmake which can generate all of the 3 major OS platform icons, but I think its overkill for the following reasons:

* The icon svg will hardly be updated, so theres no need to make everyone who downloads citra from source to build it.
* It requires convert.exe from imagemagick on windows to generate the .ico file so we'd need to redistribute yet another binary.
* Its not that hard to make icons from an svg anyway.

I just don't think its worth it to generate the icons with cmake, but i'm leaving this research here as a comment in case we decide to revisit this in the future.

bunnei:
As others have said, there is pretty much no reason why the SVG should be read by humans. IMO it's preferable for it to be version controlled as the file is generated. Requiring the submitter to reformat it on edits is silly.

PR #2891, opened 2017-08-21
SwRasterizer/Lighting: implement bump mapping

wwylele:
This is the last part of SwRasterizer/Lighting to reach feature parity with the GL one, including normal map, tangent map and LUT input CP. See #2762 for the GL counterpart and related hwtest.

----
This also includes a revision of the CP formula: if we assume that `normal` is a unit vector, then the term `/ dot(normal, normal)` is a no-op. This assumption doesn't hold true for the edge case that `bump_renorm` is off and the map contains unnormalized vectors. Further hwtest shows that PICA probably makes the same assumption, which makes sense because thus it doesn't need to do expensive division. The detailed test result is shown below, using the program in #2762, in normal map mode, with CP input and `bump_renorm` off

|3DS|GL(master) with old formula |SW or GL(this branch)* with revised formula|
|-|-|-|
|3ds|hw|sw

_* The result of SW and GL in this branch are almost identical, except for two or three pixels_

The normal map used in the program is this
normal1

It can be seen that many areas that fall in the edge case get fixed with the revised formula. However, there is a regression at the center, where the normal vector is set to (0,0,0), which is a very edge case that might trigger some errors in PICA. The (0,0,0) normal vector already has some issues before, and given that it is unlikely to be used in games, I think it is OK to leave it as is.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2891)

### DIFF
--- a/src/video_core/swrasterizer/lighting.cpp
+++ b/src/video_core/swrasterizer/lighting.cpp
@@ -22,14 +22,32 @@ static float LookupLightingLut(const Pica::State::Lighting& lighting, size_t lut
 
 std::tuple<Math::Vec4<u8>, Math::Vec4<u8>> ComputeFragmentsColors(
     const Pica::LightingRegs& lighting, const Pica::State::Lighting& lighting_state,
-    const Math::Quaternion<float>& normquat, const Math::Vec3<float>& view) {
+    const Math::Quaternion<float>& normquat, const Math::Vec3<float>& view,
+    const Math::Vec4<u8> (&texture_color)[4]) {
 
-    // TODO(Subv): Bump mapping
-    Math::Vec3<float> surface_normal = {0.0f, 0.0f, 1.0f};
+    Math::Vec3<float> surface_normal;
+    Math::Vec3<float> surface_tangent;
 
     if (lighting.config0.bump_mode != LightingRegs::LightingBumpMode::None) {
-        LOG_CRITICAL(HW_GPU, "unimplemented bump mapping");
-        UNIMPLEMENTED();
+        Math::Vec3<float> perturbation =
+            texture_color[lighting.config0.bump_selector].xyz().Cast<float>() / 127.5f -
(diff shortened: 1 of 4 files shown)


src/video_core/swrasterizer/lighting.cpp
```diff
@@ -22,18 +22,37 @@ static float LookupLightingLut(const Pica::State::Lighting& lighting, size_t lut
...
 
-    // TODO(Subv): Bump mapping
-    Math::Vec3<float> surface_normal = {0.0f, 0.0f, 1.0f};
+    Math::Vec3<float> surface_normal;
+    Math::Vec3<float> surface_tangent;
 
     if (lighting.config0.bump_mode != LightingRegs::LightingBumpMode::None) {
-        LOG_CRITICAL(HW_GPU, "unimplemented bump mapping");
-        UNIMPLEMENTED();
+        Math::Vec3<float> pertubation =
```
bunnei: nit: `const`

src/video_core/swrasterizer/lighting.cpp
```diff
@@ -22,18 +22,37 @@ static float LookupLightingLut(const Pica::State::Lighting& lighting, size_t lut
...
+            texture_color[lighting.config0.bump_selector].xyz().Cast<float>() / 127.5f -
+            Math::MakeVec(1.0f, 1.0f, 1.0f);
+        if (lighting.config0.bump_mode == LightingRegs::LightingBumpMode::NormalMap) {
+            if (!lighting.config0.disable_bump_renorm) {
+                float z_square = 1 - pertubation.xy().Length2();
```
bunnei: nit: `const`

src/video_core/swrasterizer/lighting.cpp
wwylele: This variable is actually modified below...

PR #2892, opened 2017-08-21
Warnings: Fixed a few missing-return warnings in video_core.

Subv:
The reason i'm using ASSERT(false) in some places instead of UNREACHABLE() is that I can not be sure that the code is actually unreachable, there are gaps in the depth format values.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2892)

### DIFF
--- a/src/video_core/regs_framebuffer.h
+++ b/src/video_core/regs_framebuffer.h
@@ -256,10 +256,9 @@ struct FramebufferRegs {
             return 3;
         case DepthFormat::D24S8:
             return 4;
-        default:
-            LOG_CRITICAL(HW_GPU, "Unknown depth format %u", format);
-            UNIMPLEMENTED();
         }
+
+        ASSERT_MSG(false, "Unknown depth format %u", format);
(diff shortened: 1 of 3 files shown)


src/video_core/regs_framebuffer.h
```diff
@@ -256,10 +256,10 @@ struct FramebufferRegs {
             return 3;
         case DepthFormat::D24S8:
             return 4;
-        default:
-            LOG_CRITICAL(HW_GPU, "Unknown depth format %u", format);
-            UNIMPLEMENTED();
         }
+
+        LOG_CRITICAL(HW_GPU, "Unknown depth format %u", format);
```
jroweboy: `ASSERT_MSG(false, "Unknown depth format %u", format);`

(we have a macro just for this sorta thing https://github.com/citra-emu/citra/blob/0c8df1046f789c9f965300c54d03004fc0b8721e/src/common/assert.h#L37 )

PR #2894, opened 2017-08-22
motion_emu: fix initialization order

wwylele:
Sorry, this is a flaw from #2861 . The thread is initialized before `status_mutex` and can run fast enough to access the uninitialized mutex.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2894)


danzel:
LGTM, I've tested it and it fixes the issue.

PR #2895, opened 2017-08-22
Groundwork for Screenshot functionality

danzel:
Adds functionality to take a screenshot from the framebuffer memory.
The EmuWindow requests a screenshot by returning true from `IsScreenshotRequested`, which `RendererOpenGL` checks once per frame.
When set, the renderer copies the framebuffer in to a `ScreenshotData` object and passes it through to `emu_window->ReceiveScreenshot`.

Currently the frontends just contain sample implementations of saving an image. QT will need some UI for it, but SDL can probably stay as it is (or get a keyboard shortcut?).

When using increased Internal Resolution rendering we still produce screenshots in normal 3ds resolution.
I think we could read back the rasterizer framebuffer textures instead if capturing this is desirable.

TBH I'm not sure if any of this is done the right way, so feedback is desired :)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2895)

### DIFF
--- a/src/video_core/renderer_base.cpp
+++ b/src/video_core/renderer_base.cpp
@@ -4,6 +4,9 @@
 
 #include <atomic>
 #include <memory>
+#include "common/color.h"
+#include "core/frontend/emu_window.h"
+#include "core/frontend/screenshot_data.h"
 #include "core/hw/hw.h"
 #include "core/hw/lcd.h"
 #include "video_core/renderer_base.h"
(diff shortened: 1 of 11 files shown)


src/video_core/renderer_base.cpp
```diff
@@ -21,3 +26,79 @@ void RendererBase::RefreshRasterizerSetting() {
...
+    // 3ds framebuffers are stored rotate 90 degrees
+    std::unique_ptr<ScreenshotData> screenshot = std::make_unique<ScreenshotData>(
+        GPU::g_regs.framebuffer_config[0].height, GPU::g_regs.framebuffer_config[0].width,
+        GPU::g_regs.framebuffer_config[1].height, GPU::g_regs.framebuffer_config[1].width);
+
+    for (int i : {0, 1}) {
+        const auto& framebuffer = GPU::g_regs.framebuffer_config[i];
+        LCD::Regs::ColorFill color_fill{GetColorFillForFramebuffer(i)};
+
+        u8* dest = screenshot.get()->screens[i].data.data();
```
Subv: why get()?.

dest can have a more descriptive name

danzel:
Renamed `dest` to `destBuffer` and updated the comment below related to it.
Removed the unnecessary `.get()` calls.
Thanks!

danzel:
Fixed a few variables I'd named with the wrong name format. (thanks wwylele)

JayFoxRox:
I think this is a bad feature and it should not exist.

There are various ways to grab screenshots. Even mobile phones / OS allow to take screenshots these days. We also have debug features to inspect the framebuffer memory.

In my opinion, a feature for taking screenshots is just code bloat and hardly anyone will actually use it.
The only purpose I can see is to grab screenshots for the compatibility DB  or at very high resolutions.
However, this PR seems to have a different goal (which I personally don't like).

danzel:
This PR is definitely tied to #2898 and if that isn't going to go ahead then this shouldn't either.
For that proposal we need a method of taking a screenshot at a specific frame. Capturing that externally is not going to be possible.

This feature does not need to be as generic as it is implemented in this PR, it could be implemented specifically inside of the test related code instead (when that eventually exists).

I submitted it like this as I thought it may be useful to be able to get exact pixel screenshots for comparison with rendering behaviour of real hardware. But I'm happy to hide it away instead.

wwylele:
I was requested to review this, but after trying it, I found that I am probably not able to do it. Sorry.

The reason is that I don't fully understand the original design idea of the part of the old code this PR touches. Questions such as "why was `SetWindow` pure virtual" and "is `EmuWindow` designed for handling screenshot request" leave unanswered to me. I also have an impression that this part of code is less maintained and need some refactor before adding new feature. We have also brought up questions like "why is swrenderer still tied to OpenGL (3.3)", which more or less indicates the structure and the interface around the renderer is not good.

danzel:
Thanks for the feedback @wwylele, I'll close this off for now.

PR #2897, opened 2017-08-23
Telemetry UI and final touches

bunnei:
This is the fourth part of a set of changes to add a telemetry framework to Citra. In the near term, this will be used for user-submitted compatibility data to our own custom service endpoint (services.citra-emu.org). My goal, however, is to make this very generic, and allow us to easily log useful/interesting data fields throughout the emulator, and then upload them to our service at a later time. In the long term, we may use this for anonymous telemetry, critical errors and crash reports, measuring how often features are used, etc.

With this change, I've added the following:
* Settings UI for configuring telemetry related fields
* "Callouts", one-time important messages to the user, which is currently used to indicate the addition of telemetry
* TelemetryId, a user-unique random number used for grouping telemetry data by user
* Support for unauthenticated/anonymous telemetry
* Logging of the emulation frontend type (Qt or SDL)

- [x] [#2819] Implement a backend that does an authenticated log submission to our service
- [x] [#2823] Implement remaining fields for game compatibility submission
- [x] Implement Qt UI for authenticating users and user-submitted data

Obligatory UI screenshots:

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2897)

### DIFF
--- a/src/citra_qt/configuration/configure_web.ui
+++ b/src/citra_qt/configuration/configure_web.ui
@@ -1,0 +1,126 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>ConfigureWeb</class>
+ <widget class="QWidget" name="ConfigureWeb">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>400</width>
+    <height>300</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Form</string>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <layout class="QVBoxLayout" name="verticalLayout_3">
+     <item>
+      <widget class="QGroupBox" name="groupBox">
+       <property name="title">
+        <string>Citra Web Service</string>
+       </property>
+       <layout class="QVBoxLayout" name="verticalLayout_2">
+        <item>
+         <widget class="QLabel" name="web_credentials_disclaimer">
+          <property name="text">
+           <string>By providing your username and token, you agree to allow Citra to collect additional usage data, which may include user identifying information.</string>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <layout class="QGridLayout" name="gridLayout">
+          <item row="0" column="0">
+           <widget class="QLabel" name="label_username">
+            <property name="text">
+             <string>Username: </string>
+            </property>
+           </widget>
+          </item>
+          <item row="0" column="1">
+           <widget class="QLineEdit" name="edit_username">
+            <property name="maxLength">
+             <number>36</number>
+            </property>
+           </widget>
+          </item>
+          <item row="1" column="0">
+           <widget class="QLabel" name="label_token">
+            <property name="text">
+             <string>Token: </string>
+            </property>
+           </widget>
+          </item>
+          <item row="1" column="1">
+           <widget class="QLineEdit" name="edit_token">
+            <property name="maxLength">
+             <number>36</number>
+            </property>
+            <property name="echoMode">
+             <enum>QLineEdit::Password</enum>
+            </property>
+           </widget>
+          </item>
+          <item row="2" column="0">
+           <widget class="QLabel" name="web_signup_link">
+            <property name="text">
+             <string>Sign up</string>
+            </property>
+           </widget>
+          </item>
+          <item row="2" column="1">
+           <widget class="QLabel" name="web_token_info_link">
+            <property name="text">
+             <string>What is my token?</string>
+            </property>
+           </widget>
+          </item>
+         </layout>
+        </item>
+       </layout>
+      </widget>
+     </item>
+     <item>
+      <widget class="QGroupBox" name="groupBox">
+       <property name="title">
+        <string>Telemetry</string>
+       </property>
+       <layout class="QVBoxLayout" name="verticalLayout_2">
+        <item>
+         <widget class="QCheckBox" name="toggle_telemetry">
+          <property name="text">
+           <string>Share anonymous usage data with the Citra team</string>
+          </property>
+         </widget>
+        </item>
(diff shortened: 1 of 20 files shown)


mailwl:
omg! so long password

src/web_service/web_backend.cpp
```diff
@@ -5,48 +5,37 @@
...
 static constexpr char API_VERSION[]{"1"};
-static constexpr char ENV_VAR_USERNAME[]{"CITRA_WEB_SERVICES_USERNAME"};
-static constexpr char ENV_VAR_TOKEN[]{"CITRA_WEB_SERVICES_TOKEN"};
 
-static std::string GetEnvironmentVariable(const char* name) {
-    const char* value{getenv(name)};
-    if (value) {
-        return value;
+void PostJson(const std::string& url, const std::string& data) {
+    if (!Settings::values.enable_telemetry) {
```
B3n30: This check should be in TelemetryJson::Complete() since web_backend isn't for telemetry only, right?

src/web_service/web_backend.cpp
```diff
@@ -5,48 +5,37 @@
...
 
 static constexpr char API_VERSION[]{"1"};
-static constexpr char ENV_VAR_USERNAME[]{"CITRA_WEB_SERVICES_USERNAME"};
-static constexpr char ENV_VAR_TOKEN[]{"CITRA_WEB_SERVICES_TOKEN"};
 
-static std::string GetEnvironmentVariable(const char* name) {
-    const char* value{getenv(name)};
-    if (value) {
-        return value;
+void PostJson(const std::string& url, const std::string& data) {
```
B3n30: I think the web_service calling PostJson should also send a flag if the data should be sent anonymous if no lgin data is available, e.g.
`void PostJson(const std::string& url, const std::string& data, const bool enable_anonymous=true)`

src/web_service/web_backend.cpp
bunnei: Derp - yes

src/web_service/web_backend.cpp
bunnei: Yes - this will make it easier for other callers. I wasn't thinking about you guys when I wrote it :)

jroweboy:
Calling dibs on review, but i might not be able to review it till next week since i'm outta town this weekend. sorry!

src/web_service/web_backend.cpp
```diff
@@ -5,48 +5,37 @@
 #include <cpr/cpr.h>
 #include <stdlib.h>
 #include "common/logging/log.h"
+#include "core/settings.h"
```
bunnei: This is not new with this PR, but will fix...

src/core/telemetry_session.cpp
```diff
@@ -3,8 +3,10 @@
 // Refer to the license.txt file included.
 
 #include <cstring>
+#include <cryptopp/osrng.h>
```
bunnei: Huh? This is already used in Core..?

wwylele: oops nvm, was looking at the wrong file :p

wwylele: oops nvm, was looking at the wrong file :p

src/citra_qt/main.cpp
```diff
@@ -48,6 +48,47 @@
...
+    msg.setText(message);
+    msg.setStandardButtons(QMessageBox::Ok);
+    msg.setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
+    msg.setStyleSheet("QLabel{min-width: 900px;}");
+    msg.exec();
+}
+
+void GMainWindow::ShowCallouts() {
+    static const QString telemetry_message =
+        tr("To improve Citra, we have begun collecting anonymous usage data. No private or "
```
danzel: Are we planning to come back and change this text again when the feature isn't new?
Maybe "To improve Citra, we collect anonymous usage data." instead?

src/web_service/web_backend.cpp
```diff
@@ -10,43 +10,33 @@
...
         return;
     }
 
-    cpr::PostAsync(cpr::Url{url}, cpr::Body{data}, cpr::Header{{"Content-Type", "application/json"},
-                                                               {"x-username", GetUsername()},
-                                                               {"x-token", GetToken()},
-                                                               {"api-version", API_VERSION}});
+    if (are_credentials_provided) {
+        // Authenticated request if credentials are provided
+        cpr::PostAsync(cpr::Url{url}, cpr::Body{data},
```
wwylele: I found that `PostAsync` doesn't actually go async when I tested it. The reason is that the function returns a future, which we doesn't hold. The future gets destructed immediately after the function call, and it gets blocked in its destructor to wait for the Post to finish.

Probably we can do `std::thread(cpr::Post, ...).detach()` instead

src/citra_qt/main.cpp
bunnei: Sure, I will just make the above change now

src/citra_qt/configuration/configure_web.ui
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
bunnei: addressed

bunnei:
@B3n30 while that would probably be slightly better, I wasn't sure how to do that and it's not really important. Disabling web_services won't be an officially supported configuration, it's just for faster dev builds and unofficial stuff. For these scenarios, the extra tab doesn't really harm anything.

src/web_service/web_backend.cpp
```diff
@@ -2,15 +2,31 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <thread>
 #include <cpr/cpr.h>
 #include <stdlib.h>
+
```
wwylele: no need this new line

PR #2899, opened 2017-08-24
Refactor touch input into a TouchDevice

wwylele:
This is a follow-up of #2497 and #2861 , and it is the last one remained to integrate into the input system.

Unlike the other inputs, touch input is highly bound to the window, so instead of decoupling it from `EmuWindow`, I let `EmuWindow` becomes the factory provider, and wrapped the plain touch state previously in `EmuWindow` into the `TouchState`, which implements the device factory.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2899)

### DIFF
--- a/src/core/frontend/emu_window.h
+++ b/src/core/frontend/emu_window.h
@@ -4,11 +4,10 @@
 
 #pragma once
 
-#include <mutex>
+#include <memory>
 #include <tuple>
 #include <utility>
 #include "common/common_types.h"
-#include "common/math_util.h"
 #include "core/frontend/framebuffer_layout.h"
 
 /**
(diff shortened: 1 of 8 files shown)


src/core/frontend/emu_window.h
```diff
@@ -177,10 +158,8 @@ class EmuWindow {
                                 /// ProcessConfigurationChanges)
     WindowConfig active_config; ///< Internal active configuration
 
-    bool touch_pressed; ///< True if touchpad area is currently pressed, otherwise false
-
-    u16 touch_x; ///< Touchpad X-position in native 3DS pixel coordinates (0-320)
-    u16 touch_y; ///< Touchpad Y-position in native 3DS pixel coordinates (0-240)
+    class TouchState;
```
danzel: Other places I've seen we forward-declare classes outside of the class, not right before use. Not sure if this is a rule or not.

danzel: OIC, this is an internal class, my bad this is probably fine :)

PR #2900, opened 2017-08-25
PICA: implement custom clip plane

wwylele:
Not sure if this is used by any games, but since it was also tested in my previous PR, it can be just implemented now so that we won't forget it later.

[Here is a test program](https://github.com/wwylele/ctrhwtest/tree/master/clip-plane)

In GL implementation, the clipping calculation is in the vertex shader, which now needs uniform input. So I let vertex shader and fragment shader share the same uniform block and put the clip coefficient there.

I also flipped the sign convention of the sw clipper to use the same rule that GL and PICA use.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2900)

### DIFF
--- a/src/video_core/regs_rasterizer.h
+++ b/src/video_core/regs_rasterizer.h
@@ -5,10 +5,10 @@
 #pragma once
 
 #include <array>
-
 #include "common/bit_field.h"
 #include "common/common_funcs.h"
 #include "common/common_types.h"
+#include "video_core/pica_types.h"
 
 namespace Pica {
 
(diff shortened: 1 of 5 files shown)


yuriks:
:lgtm:

---

Reviewed 2 of 2 files at r1, 3 of 3 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2900#-:-Ku8m301-S02-DWIspWY:bnfp4nl)*

PR #2903, opened 2017-08-27
externals: Use system curl on Unix/Apple.

bunnei:
Curl doesn't seem to be working with nightlies on Apple... trying this

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2903)


bunnei:
Does not resolve issue

PR #2906, opened 2017-08-29
Services/NS: Port ns:s to the new service framework.

Subv:
There is now an NS::InstallInterfaces function, all future NS-provided services should be registered there.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2906)


src/core/hle/service/ns/ns.h
```diff
@@ -0,0 +1,19 @@
...
+
+#pragma once
+
+#include <memory>
+#include "core/hle/service/service.h"
+
+namespace Service {
+namespace NS {
+
+using SM::ServiceManager;
```
lioncash: Is there really a benefit of having a `using` here just to avoid typing `SM::`?

src/core/hle/service/ns/ns_s.h
```diff
@@ -1,21 +1,22 @@
-// Copyright 2015 Citra Emulator Project
+// Copyright 2017 Citra Emulator Project
```
lioncash: This doesn't need to be updated. It's to establish year of creation.

src/core/hle/service/ns/ns.cpp
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/hle/service/ns/ns.h"
+#include "core/hle/service/ns/ns_s.h"
+
+namespace Service {
+namespace NS {
+
+void InstallInterfaces(std::shared_ptr<ServiceManager> service_manager) {
```
lioncash: Should this not take `service_manager` as a reference (or take a reference to type `ServiceManager&` directly) to avoid the atomic reference count increase/decrease?

src/core/hle/service/ns/ns.cpp
```diff
@@ -0,0 +1,16 @@
...
+    std::make_shared<NS_S>()->InstallAsService(*service_manager);
```
lioncash: This seems like an odd API use (constructing on the heap just to toss the object away), that aside, wouldn't `std::make_unique<NS_S>()` be more appropriate here?

src/core/hle/service/ns/ns_s.h
```diff
@@ -1,21 +1,22 @@
...
 #pragma once
 
+#include "core/hle/kernel/kernel.h"
 #include "core/hle/service/service.h"
 
 namespace Service {
 namespace NS {
 
-class NS_S final : public Interface {
+using SM::ServiceManager;
```
lioncash: Ditto regarding the `using`. Especially considering it's not even used in the header anywhere.

src/core/hle/service/ns/ns.cpp
Subv: Somewhere inside InstallAsService, the NS_S instance is stored inside the ServerPort via shared_from_this()

lioncash: Ah OK, that makes sense. Missed that.

vonchenplus:
:lgtm:

---

Review status: 0 of 7 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/core/hle/service/ns/ns.cpp, line 11 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2906#-Ksir7Z0gPF6YJzJ3DL8-r1-11:-KsvAekEXOTVPC9vqrMQ:bnfp4nl) ([raw file](https://github.com/citra-emu/citra/blob/0c8eb6e40e7ed8fcc007b874a880973a668c1938/src/core/hle/service/ns/ns.cpp#L11)):*

Should this not take `service_manager` as a reference (or take a reference to type `ServiceManager&` directly) to avoid the atomic reference count increase/decrease?
</blockquote></details>

:lgtm:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2906#-:-KsvB306eKCMxioQyVuk:bnfp4nl)*

vonchenplus:
Review status: 0 of 7 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/core/hle/service/ns/ns.cpp, line 12 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2906#-Ksir7Z0gPF6YJzJ3DL8-r1-12:-KsvBpZp5jwVNsN5S0cO:bnfp4nl) ([raw file](https://github.com/citra-emu/citra/blob/0c8eb6e40e7ed8fcc007b874a880973a668c1938/src/core/hle/service/ns/ns.cpp#L12)):*

Ah OK, that makes sense. Missed that.
</blockquote></details>

:lgtm:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2906)*

yuriks:
:lgtm:, but fix Lioncash's comment first.

---

Reviewed 8 of 8 files at r1.

---

*[src/core/hle/service/ns/ns.cpp, line 11 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2906#-Ksir7Z0gPF6YJzJ3DL8-r1-11:-Ku8uXOpUDXTBFe92hkQ:b-yz73a) ([raw file](https://github.com/citra-emu/citra/blob/0c8eb6e40e7ed8fcc007b874a880973a668c1938/src/core/hle/service/ns/ns.cpp#L11)):*
<details><summary><i>Previously, vonchenplus (Feng Chen) wrote…</i></summary><blockquote>

:lgtm:
</blockquote></details>

Yeah, this should probably be a reference to ServiceManager instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2906#-:-Ku8uskIcV0N_LusFOzw:b-tkt70r)*

yuriks:
Reviewed 4 of 4 files at r2.

---

*[src/core/hle/service/ns/ns.cpp, line 11 at r1](https://reviewable.io:443/reviews/citra-emu/citra/2906#-Ksir7Z0gPF6YJzJ3DL8-r1-11:-KuB5bkaOGyRr0n551LX:b32w0x) ([raw file](https://github.com/citra-emu/citra/blob/0c8eb6e40e7ed8fcc007b874a880973a668c1938/src/core/hle/service/ns/ns.cpp#L11)):*

Yeah, this should probably be a reference to ServiceManager instead.
</blockquote></details>

OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2906)*

PR #2907, opened 2017-08-29
Disable unary operator- on Math::Vec2/Vec3/Vec4 for unsigned types.

Subv:
It is unlikely we will ever use this without first doing a Cast to a signed type.
Fixes 9 "unary minus operator applied to unsigned type, result still unsigned" warnings on MSVC2017.3.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2907)

### DIFF
--- a/src/common/vector_math.h
+++ b/src/common/vector_math.h
@@ -31,6 +31,7 @@
 #pragma once
 
 #include <cmath>
+#include <type_traits>
 
 namespace Math {
 
@@ -90,7 +91,8 @@ public:
         y -= other.y;
     }
 
-    Vec2<decltype(-T{})> operator-() const {
+    template <typename U = T>
+    Vec2<std::enable_if_t<std::is_signed<U>::value, U>> operator-() const {
         return MakeVec(-x, -y);
     }
     Vec2<decltype(T{} * T{})> operator*(const Vec2& other) const {
@@ -247,7 +249,8 @@ public:
         z -= other.z;
     }
 
-    Vec3<decltype(-T{})> operator-() const {
+    template <typename U = T>
+    Vec3<std::enable_if_t<std::is_signed<U>::value, U>> operator-() const {
         return MakeVec(-x, -y, -z);
     }
     Vec3<decltype(T{} * T{})> operator*(const Vec3& other) const {
@@ -462,7 +465,8 @@ public:
         w -= other.w;
     }
 
-    Vec4<decltype(-T{})> operator-() const {
+    template <typename U = T>
+    Vec4<std::enable_if_t<std::is_signed<U>::value, U>> operator-() const {
         return MakeVec(-x, -y, -z, -w);
     }
     Vec4<decltype(T{} * T{})> operator*(const Vec4& other) const {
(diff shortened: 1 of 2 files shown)


src/video_core/pica_types.h
```diff
@@ -145,3 +146,23 @@ using float20 = Float<12, 7>;
 using float16 = Float<10, 5>;
 
 } // namespace Pica
+
+/// std::numeric_limits specializations for our PICA float types
```
lioncash: this should likely be a regular comment, otherwise this is documenting over `namespace std`

src/video_core/pica_types.h
```diff
@@ -145,3 +146,23 @@ using float20 = Float<12, 7>;
...
+namespace std {
+template <>
+struct numeric_limits<Pica::float24> {
```
lioncash: Should these not also have the other constants specified? In particular, the ones relating to floating-point arithmetic?

yuriks:
A similar check was very recently removed because it didn't work properly: f3660ba9dd13f342c591aaa9901e94b6caee8d9a

Make sure you're not re-introducing a problem.

Subv:
@yuriks the warning occurs because MSVC substitutes decltype(-T{}) to decltype(-some_unsigned_type{}), this version of operator- does away with the decltype entirely.

As for the problem with custom types, that's what the numeric_limits<> specializations are for.

src/video_core/pica_types.h
Subv: We are not using those anywhere so I didn't add them, they can be added later when needed so as to not clutter this specialization with potentially unneeded members.

wwylele: http://en.cppreference.com/w/cpp/language/extending_std says
> Specializations of std::numeric_limits must define all members declared static constexpr in the primary template, in such a way that they are usable as integral constant expressions. 

looks like defining all constants is required?

lioncash: Yeah, that's what I had figured. It also leads to potentially unexpected failures otherwise (if numeric_limits on a type is done for a certain attribute, it's expected that the other ones are present as well from an interface point of view).

Subv:
I have added the rest of the std::numeric_limits members, except for 4 that I wasn't sure how to define.

B3n30:
I think since the underlying structure is a float and we decided to not take the small differences between float<24>, etc. and float into account those missing limits should just be the same as the float limits.

Subv:
I'm starting to think it might be better to just disable these operators for vectors of PICA Float types, what is your opinion on that?

Subv:
I have removed the Pica::Float specializations since they are a lot of boilerplate code for a comparatively small benefit.

src/common/vector_math.h
```diff
@@ -90,7 +92,8 @@ class Vec2 {
         y -= other.y;
     }
 
-    Vec2<decltype(-T{})> operator-() const {
+    template <typename U = T>
+    Vec2<std::enable_if_t<std::numeric_limits<U>::is_signed, U>> operator-() const {
```
lioncash: would these be able to use the non-member [std::is_signed](http://en.cppreference.com/w/cpp/types/is_signed) instead inside `<type_traits>`? (allowing removal of the `<limits>` include), or no? If not, then LGTM

Subv: I've updated it to use std::is_signed

PR #2909, opened 2017-08-31
video_core: report telemetry for gas mode

wwylele:
>>These developers have found it difficult to prioritize their efforts
>
>This just sounds like nonsense to me. I'm sure devs aren't sitting around going "hmm now what?". This sounds like a solution in search of a problem.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2909)


Subv:
Should this be added to the sw renderer too?

wwylele:
@Subv as users usually don't use sw renderer, there is little point to add telemetry there, and adding log there will just cause log spaming

PR #2911, opened 2017-08-31
Add manifest

DaMan69:
This tells Windows the program supports 7/8/10, is DPI aware, and enables long filename support on W10.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2911)


src/citra/citra.rc
```diff
@@ -6,4 +6,4 @@
 // Icon with lowest ID value placed first to ensure application icon
 // remains consistent on all systems.
 CITRA_ICON              ICON                    "../../dist/citra.ico"
-
+1                       24                      "../../dist/citra.manifest"
```
Subv: Aren't there any constants for this?

jroweboy: they are defined in a different header so you'd need to include it (`WinUser.h`) but the named constant is `RT_MANIFEST`

DaMan69:
Updated with Subv's feedback.

yuriks:
Is disabling the linker manifest and the manual embedding necessary? My understanding is that you can give a manifest to the build system, and it'll merge it with the linker-generated one. CMake supports this by [supplying the manifest file as a source file](https://cmake.org/cmake/help/v3.4/release/3.4.html?highlight=manifest#other).

DaMan69:
You are correct for MSVC, but it is necessary for MinGW.

PR #2912, opened 2017-09-01
Build: Add mingw64 compile support to appveyor

jroweboy:
Releases will be built with both mingw and msvc and the binaries of both
builds will be uploaded to github releases

Tests are run against both builds.

Since this doubles the build time for appveyor, I've also taken steps to reduce the build time by having bunnei turn on a feature in appveyor and travis that cancels builds if a newer build is push. This fixes an issue where a PR is merged to master, then the release bot will push a tag for it, and the tag release build will have to wait for the master build to finish first. With the change to turn on rolling builds, the tag build will cancel the master build so the release won't be held up by the useless build for master.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2912)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,15 @@ cache:
 
 os: Visual Studio 2017
 
+environment:
+  # Tell msys2 to add mingw64 to the path
+  MSYSTEM: MINGW64
+  # Tell msys2 to inherit the current directory when starting the shell
+  CHERE_INVOKING: 1
+  matrix:
+    - BUILD_TYPE: mingw
+    - BUILD_TYPE: msvc
+
 platform:
   - x64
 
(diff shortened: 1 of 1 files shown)


jroweboy:
Note that this PR is one of a few blockers for the new installer/updater, so I would appreciate timely reviews :)

appveyor.yml
```diff
@@ -15,72 +24,142 @@ configuration:
...
 test_script:
-  - cd build && ctest -VV -C Release && cd ..
+  - cd %BUILD_TYPE%build
+  - ps: |
+        if ($env:BUILD_TYPE -eq 'msvc') {
+          ctest -VV -C Release
+        } else {
+          C:\msys64\usr\bin\bash.exe -lc "ctest -VV -C Release"
+        }
+  - cd ..
```
j-selby: This needed?

appveyor.yml
```diff
@@ -15,72 +24,142 @@ configuration:
...
         }
 
-        mkdir $RELEASE_DIST
-        Copy-Item .\build\bin\release\* -Destination $RELEASE_DIST -Recurse
-        Copy-Item .\license.txt -Destination $RELEASE_DIST
-        Copy-Item .\README.md -Destination $RELEASE_DIST
-
-        7z a -tzip $MSVC_BUILD_NAME $RELEASE_DIST\*
-        7z a $MSVC_SEVENZIP $RELEASE_DIST
+        if ($env:BUILD_TYPE -eq 'mingw') {
```
j-selby: You are matching different conditions here vs. in `test_script` - might want to flip around one of the statements.

appveyor.yml
```diff
@@ -15,72 +24,142 @@ configuration:
...
+          # workaround to prevent pacman from failing due to cyclical dependencies
+          C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S mingw64/mingw-w64-x86_64-freetype mingw64/mingw-w64-x86_64-fontconfig" 2> $null
+          C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S $dependencies" 2> $null
+        }
 
 before_build:
-  - mkdir build
-  - cd build
-  - cmake -G "Visual Studio 15 2017 Win64" -DCITRA_USE_BUNDLED_QT=1 -DCITRA_USE_BUNDLED_SDL2=1 -DCMAKE_USE_OPENSSL=0 ..
+  - mkdir %BUILD_TYPE%build
```
j-selby: Nit: Think there should be a separator here - `-`, `_`.

appveyor.yml
```diff
@@ -15,72 +24,142 @@ configuration:
...
+          $env:BUILD_ZIP = $MINGW_BUILD_ZIP
+          $env:BUILD_SYMBOLS = $MINGW_NO_DEBUG_SYMBOLS
+          $env:BUILD_UPDATE = $MINGW_SEVENZIP
+
+          $CMAKE_SOURCE_DIR = "$env:APPVEYOR_BUILD_FOLDER"
+          $CMAKE_BINARY_DIR = "$CMAKE_SOURCE_DIR/mingwbuild"
+          $RELEASE_DIST = $RELEASE_DIST + "-mingw"
+          mkdir $RELEASE_DIST
+          mkdir $RELEASE_DIST/platforms
+          # copy the compiled binaries and other release files to the release folder
```
j-selby: Some newlines in this block would be nice.

appveyor.yml
jroweboy: Its probably for the best. If at some point we need to add after_test scripts, we'll need to add it back in anyway.

PR #2831, opened 2017-07-17
Services/UDS: Handle beacon frames and the basic AP connection sequence frames.

Subv:
There is currently no connection stage tracking, a client is considered "Connected" when it receives the EAPoL Logoff packet from the server, this is not yet implemented.

The received beacon frames are stored in a circular buffer of 15 elements, where new elements displace the older ones.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2831)

PR #2920, opened 2017-09-07
Services/UDS: Added a function to send EAPoL-Start packets

B3n30:
This is a continuation of #2831. It is the 2nd of 3 PRs to complete the connection sequence.
It also adds the filters missing in #2831 I promised.

EAPoL-Start packets are sent in response to an AssociationResponse.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2920)


src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -91,6 +120,28 @@ void HandleBeaconFrame(const Network::WifiPacket& packet) {
         received_beacons.pop_front();
 }
 
+void HandleAssociationResponseFrame(const Network::WifiPacket& packet) {
+    auto assoc_result = GetAssociationResult(packet.data);
+
+    ASSERT_MSG(std::get<AssocStatus>(assoc_result) == AssocStatus::Successful,
+               "Could not join network");
+    {
```
InusualZ: Since you are creating `sub-scope` just to avoid deadlock(I guess?). Why not create a method to avoid code duplication

B3n30: I don#t think moving ASSERTS in a method is a good way, because you loose track of which function was responsible for the assert in the first place. Maybe add a compiler definition, but I think thats a lot of effort, just for some asserts, that are mainly to keep track for development.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -70,7 +78,19 @@ static std::deque<Network::WifiPacket> received_beacons;
  */
 std::deque<Network::WifiPacket> GetReceivedBeacons(const MacAddress& sender) {
     std::lock_guard<std::mutex> lock(beacon_mutex);
-    // TODO(Subv): Filter by sender.
+    if (sender != Network::BroadcastMac) {
+        std::deque<Network::WifiPacket> filtered_list;
+        const auto beacon = std::find_if(received_beacons.begin(), received_beacons.end(),
```
Subv: I think instead of doing this, we could have the received_beacons list be a set of found networks containing only the latest beacon for that network instead of a deque of all beacons.

When we receive a beacon we'd update the data for that specific network.
If a network hasn't received any new beacons after certain time, we remove it from the lst.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -37,10 +38,13 @@ static ConnectionStatus connection_status{};
 /* Node information about the current network.
  * The amount of elements in this vector is always the maximum number
  * of nodes specified in the network configuration.
- * The first node is always the host, so this always contains at least 1 entry.
+ * The first node is always the host.
  */
 static NodeList node_info(1);
 
+// Node information about our own system.
+static NodeInfo current_node;
```
Subv: You're not using this anywhere

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -37,10 +38,13 @@ static ConnectionStatus connection_status{};
 /* Node information about the current network.
  * The amount of elements in this vector is always the maximum number
  * of nodes specified in the network configuration.
- * The first node is always the host, so this always contains at least 1 entry.
```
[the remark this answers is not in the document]
B3n30: The first node is always the host, but nevertheless it's empty until a host is available

src/core/hle/service/nwm/nwm_uds.cpp
B3n30: Since this is for storing beacons a set isn't practical, since beacons aren't unique even if they come from the same sender. Thus I choose to do it with a list

PR #2921, opened 2017-09-08
GPU: Add draw for immediate and batch modes

jroweboy:
PR #1461 introduced a regression where some games would change configuration
even while in the poorly named "drawing" mode, which broke the heuristic
citra was using to determine when to draw the batch. This change adds
back in a draw call for batching, and also adds in a draw call in
immediate mode each time it adds a triangle.

This is a continuation of https://github.com/citra-emu/citra/pull/2866 and follows the advice that @ds84182 laid out in the following comment https://github.com/citra-emu/citra/pull/2866#issuecomment-327011550 This change was discussed with @yuriks on IRC as well, so I hope I got it right this time.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2921)

### DIFF
--- a/src/video_core/command_processor.cpp
+++ b/src/video_core/command_processor.cpp
@@ -243,6 +243,15 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     ASSERT(!g_state.geometry_pipeline.NeedIndexInput());
                     g_state.geometry_pipeline.Setup(shader_engine);
                     g_state.geometry_pipeline.SubmitVertex(output);
+
+                    // TODO: If drawing after every immediate mode triangle kills performance,
+                    // change it to flush triangles whenever a draing config register changes
+                    // See: https://github.com/citra-emu/citra/pull/2866#issuecomment-327011550
+                    VideoCore::g_renderer->Rasterizer()->DrawTriangles();
+                    if (g_debug_context) {
+                        g_debug_context->OnEvent(DebugContext::Event::FinishedPrimitiveBatch,
+                                                 nullptr);
+                    }
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


wwylele:
I think the `gpu_mode` block can be removed
https://github.com/citra-emu/citra/blob/11baa40d75d8a479bd738a05e91bba6f09856fe5/src/video_core/command_processor.cpp#L252-L263

Also as seen in this block, you may want to add `MICROPROFILE_SCOPE(GPU_Drawing);` to `DrawTriangles`

Do we know which game uses immediate mode the most? It would be nice to see how much performance regression is there. If this kills the performance, please add a TODO there about the proposed second method.

Reminder: fix format issue.

jroweboy:
@wwylele what part of the `gpu_mode` block? All of the contents? Or just the `if` surrounding the contents. 

Also I've addressed your other comments.

wwylele:
> what part of the gpu_mode block? All of the contents? Or just the if surrounding the contents.
The entire `case` block

I was wrong about `MICROPROFILE_SCOPE(GPU_Drawing);`. They were already there (several lines above) so you don't need to add for either of two

src/video_core/command_processor.cpp
```diff
@@ -243,6 +243,15 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                     ASSERT(!g_state.geometry_pipeline.NeedIndexInput());
                     g_state.geometry_pipeline.Setup(shader_engine);
                     g_state.geometry_pipeline.SubmitVertex(output);
+
+                    // TODO: If drawing after every immediate mode triangle kills performance,
+                    // change it to flush triangles whenever a draing config register changes
```
Subv: `draing`

src/video_core/command_processor.cpp
```diff
@@ -398,6 +407,12 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
                                                       range.second, range.first);
         }
 
+        MICROPROFILE_SCOPE(GPU_Drawing);
```
Subv: There is already a call to MICROPROFILE_SCOPE(GPU_Drawing); in this case block

jroweboy:
Updated pls review and merge kthxbai

PR #2922, opened 2017-09-09
Build: Enable SSL in mingw by linking against WinSSL

jroweboy:
The mingw builds aren't submitting telemetry because the curl library
they are linked against is configured to use openSSL and openSSL looks
for the certificates in the users home folder. This keeps it from
contacting web services because it can't communicate over SSL.

This commit adds a download in mingw builds that will download a
precompiled curl for mingw linked against winssl and sspi.

This change is dependent on https://github.com/citra-emu/ext-windows-bin/pull/7 so **do not merge before that one** This PR won't even build properly without it. I'm submitting this PR now so we can get the ext-win-bin one merged sooner than later, so we can get this one merged as well. After that one is merged, we will need to do some more testing on different machines to verify that everything is still working in both mingw and msvc builds.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2922)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@
 cmake_minimum_required(VERSION 3.6)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules")
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/externals/cmake-modules")
+include(DownloadExternals)
 
 project(citra)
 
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -12,6 +13,17 @@ option(ENABLE_QT "Enable the Qt frontend" ON)
 option(CITRA_USE_BUNDLED_QT "Download bundled Qt binaries" OFF)
 
 option(ENABLE_WEB_SERVICE "Enable web services (telemetry, etc.)" ON)
+option(CITRA_USE_BUNDLED_CURL "FOR MINGW ONLY: Download curl configured against winssl instead of openssl" OFF)
+if (ENABLE_WEB_SERVICE AND CITRA_USE_BUNDLED_CURL AND MSVC)
```
wwylele: probably `AND MSVC` -> `AND NOT MINGW` ? So that Linux/MacOS user will also get the message when turning on.
We can also follow the style of bundled SDL/Qt like
```
        if (MSVC14 AND ARCHITECTURE_x86_64)
            set(SDL2_VER "SDL2-2.0.5")
        else()
            message(FATAL_ERROR "No bundled SDL2 binaries for your toolchain. Disable CITRA_USE_BUNDLED_SDL2 and provide your own.")
        endif()
```

jroweboy: Thanks for the review. First for `MSVC` -> `AND NOT MINGW` this change is also useful for people compiling in `CYGWIN` or `CLANG` so i didn't want to say `AND NOT MINGW` but to be honest, I don't think anyone will be compiling in cygwin or clang except maybe pcmaker... I'll change it to `WINDOWS AND MSVC` then. 

What I'm thinking the long term and proper solution should be is submit a mingw64-curl-winssl-dev package to msys2 and just use msys2 to install that dependency.

PR #2923, opened 2017-09-10
travis_OSX: build with system curl

B3n30:
This fixes that osx builds won't send any telemetry

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2923)


bunnei:
ty

PR #2925, opened 2017-09-11
Allow for partial invalidation of instruction cache

MerryMage:
Add the necessary interface to `ARM_Interface`, then uses it in the ro service.

Ping @wwylele.

This probably breaks things, I'm not 100% certain I'm invaidating everything that's necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2925)


src/core/hle/service/ldr_ro/memory_synchronizer.cpp
```diff
@@ -35,6 +37,7 @@ void MemorySynchronizer::RemoveMemoryBlock(VAddr mapping, VAddr original) {
 void MemorySynchronizer::SynchronizeOriginalMemory() {
     for (auto& block : memory_blocks) {
         Memory::CopyBlock(block.original, block.mapping, block.size);
+        Core::CPU().InvalidateCacheRange(block.mapping, block.size);
```
wwylele: `MemorySynchronizer` is only a work around for unimplemented memory aliasing. The entire file will be removed once memory aliasing is implemented, so any other stuff shouldn't be added here.

src/core/hle/service/ldr_ro/ldr_ro.cpp
```diff
@@ -439,8 +437,6 @@ static void LoadCRO(Interface* self, bool link_on_load_bug_fix) {
         }
     }
 
-    Core::CPU().ClearInstructionCache();
```
wwylele: Instead of doing it in `MemorySynchronizer`, `Core::CPU().InvalidateCacheRange(cro_address, cro_size);` can be done here.

src/core/hle/service/ldr_ro/ldr_ro.cpp
```diff
@@ -535,8 +531,6 @@ static void UnloadCRO(Interface* self) {
         memory_synchronizer.RemoveMemoryBlock(cro_address, cro_buffer_ptr);
     }
 
-    Core::CPU().ClearInstructionCache();
```
wwylele: Ditto. Note that this is not needed in `LoadCRO/UnloadCRO` though, because all invalidation there is done by the one in `ApplyRelocation/ClearRelocation`

src/core/hle/service/ldr_ro/ldr_ro.cpp
MerryMage: Done.

src/core/hle/service/ldr_ro/ldr_ro.cpp
MerryMage: Did `Core::CPU().InvalidateCacheRange(cro_address, fixed_size);`

src/core/hle/service/ldr_ro/memory_synchronizer.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include <algorithm>
 #include "common/assert.h"
+#include "core/arm/arm_interface.h"
+#include "core/core.h"
```
wwylele: Unused leftover

wwylele:
Pokemon AS test result:
Can boot, but freeze with black screen when going in game after loading save.

acnleditor2:
CPU usage increases a lot.

MerryMage:
Waiting on performance improvements.

PR #2927, opened 2017-09-16
Loader/NCCH: Add support for loading application updates

shinyquagsire23:
NCCH parsing was split off into `FileSys::NCCHContainer` to allow for multiple NCCHs to be loaded in the NCCH loader. If an update NCCH is found in `sdmc:/Nintendo 3DS/.../title/<TID High>/<TID Low>/content/00000000.app` its ExeFS and ExHeader will override the base game's ExeFS and ExHeader, and the UpdateRomFS in SelfNCCH will load from the update NCCH instead of the base game NCCH.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2927)

### DIFF
--- a/src/core/file_sys/archive_selfncch.cpp
+++ b/src/core/file_sys/archive_selfncch.cpp
@@ -102,8 +102,7 @@ public:
 
         switch (static_cast<SelfNCCHFilePathType>(file_path.type)) {
         case SelfNCCHFilePathType::UpdateRomFS:
-            LOG_WARNING(Service_FS, "(STUBBED) open update RomFS");
-            return OpenRomFS();
+            return OpenUpdateRomFS();
 
         case SelfNCCHFilePathType::RomFS:
             return OpenRomFS();
@@ -179,6 +178,16 @@ private:
         }
     }
 
+    ResultVal<std::unique_ptr<FileBackend>> OpenUpdateRomFS() const {
+        if (ncch_data.update_romfs_file) {
+            return MakeResult<std::unique_ptr<FileBackend>>(std::make_unique<IVFCFile>(
+                ncch_data.update_romfs_file, ncch_data.update_romfs_offset, ncch_data.update_romfs_size));
+        } else {
+            LOG_INFO(Service_FS, "Unable to read update RomFS");
+            return ERROR_ROMFS_NOT_FOUND;
+        }
+    }
+
     ResultVal<std::unique_ptr<FileBackend>> OpenExeFS(const std::string& filename) const {
         if (filename == "icon") {
             if (ncch_data.icon) {
@@ -223,6 +232,13 @@ ArchiveFactory_SelfNCCH::ArchiveFactory_SelfNCCH(Loader::AppLoader& app_loader) 
         app_loader.ReadRomFS(romfs_file_, ncch_data.romfs_offset, ncch_data.romfs_size)) {
 
         ncch_data.romfs_file = std::move(romfs_file_);
+    }
+
+    std::shared_ptr<FileUtil::IOFile> update_romfs_file_;
+    if (Loader::ResultStatus::Success ==
+        app_loader.ReadUpdateRomFS(update_romfs_file_, ncch_data.update_romfs_offset, ncch_data.update_romfs_size)) {
+
+        ncch_data.update_romfs_file = std::move(update_romfs_file_);
     }
 
     std::vector<u8> buffer;
--- a/src/core/file_sys/archive_selfncch.cpp
+++ b/src/core/file_sys/archive_selfncch.cpp
@@ -228,19 +228,19 @@ private:
 };
 
 ArchiveFactory_SelfNCCH::ArchiveFactory_SelfNCCH(Loader::AppLoader& app_loader) {
-    std::shared_ptr<FileUtil::IOFile> romfs_file_;
+    std::shared_ptr<FileUtil::IOFile> romfs_file;
     if (Loader::ResultStatus::Success ==
-        app_loader.ReadRomFS(romfs_file_, ncch_data.romfs_offset, ncch_data.romfs_size)) {
+        app_loader.ReadRomFS(romfs_file, ncch_data.romfs_offset, ncch_data.romfs_size)) {
 
-        ncch_data.romfs_file = std::move(romfs_file_);
+        ncch_data.romfs_file = std::move(romfs_file);
     }
 
-    std::shared_ptr<FileUtil::IOFile> update_romfs_file_;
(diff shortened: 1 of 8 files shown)


B3n30:
First of all, you should fix the build errors.

Secondly I don't like that the game updates needed to be stored in sdmc. Citra is a HLE and we don't require the games to be in a specific location, thus the updates shouldn't, too. IMO it would be better to store the updates in the same folder as the game itself. I don't know enough about the structure of the update files, but I think the best way would be, to add a second path for the update-file in the open file window, as well as automatically search for updates in the game list.

shinyquagsire23:
Seems some random keystrokes slipped into `ncch_container.cpp`, compiles on my machine again so hopefully the buildbot is good.

Yeah that seems to be big debate on this one, having a second loading dialogue of some sort or mimicking 3DS behavior. My reasoning on this is that extdata, savedata, and system titles are all currently installed the same as or similarly to an actual 3DS, and other emulators like rpcs3 also keep a sort of virtual filesystem which system updates and extra game data which can be installed to.

As far as how updates are handled, I suppose someone will have to test this so I'll do a quick rundown. On the SD card, each title has a `content/` directory which has a `00000000.tmd` file, and then whatever `.app` files are listed in that TMD. For Smash 4, I have `0000000d.app` which has some manuals and stuff and `00000017.app` which has the update's ExeFS, exheader, and RomFS. I'm not certain on the launch procedure for updates, but the update's ExeFS and exheader are used in place of the main app's. If you strip off the SD card encryption the file should be identical to the contents from CDN with CDN encryption removed, and from there you can strip the NCCH encryption to get something that can be read by Citra (at least until Citra gets crypto support proper).

That's also kinda why I went for the approach of mimicking the 3DS, once encryption is added properly to Citra you could, in theory, just use your actual 3DS SD card contents outright in Citra, or install CDN contents directly. Though right now it always looks for `00000000.app` since it's not parsing TMD contents, but system titles on Citra's NAND aren't using TMDs either so maybe that's something for later?

src/core/file_sys/archive_selfncch.cpp
```diff
@@ -225,6 +235,14 @@ ArchiveFactory_SelfNCCH::ArchiveFactory_SelfNCCH(Loader::AppLoader& app_loader)
         ncch_data.romfs_file = std::move(romfs_file_);
     }
 
+    std::shared_ptr<FileUtil::IOFile> update_romfs_file_;
```
wwylele: The ending underscore of `romfs_file_` is probably a leftover during code change and is not needed. You don't need to follow the style here and can also fix the `romfs_file_` one

src/core/loader/ncch.cpp
```diff
@@ -27,87 +28,7 @@
...
-            }
-            control <<= 1;
-        }
-    }
-    return true;
-}
-
-////////////////////////////////////////////////////////////////////////////////////////////////////
-// AppLoader_NCCH class
+static const u64_le updateMask = 0x0000000e00000000;
```
wwylele: no need to use `_le` since there is no direct operation on bytes. And I think our code style is to use `UpdateMast` or `UPDATE_MASK` or `update_mask` for a constant?

src/core/loader/ncch.cpp
```diff
@@ -124,15 +45,27 @@ FileType AppLoader_NCCH::IdentifyType(FileUtil::IOFile& file) {
     return FileType::Error;
 }
 
+static std::string GetUpdateNCCHPath(u64_le program_id) {
+    u32 high = (program_id | updateMask) >> 32;
+    u32 low = (program_id | updateMask);
```
wwylele: `u32 low = (program_id | updateMask) & 0xFFFFFFFF;`
and `static_cast<u32>` for both.

src/core/loader/ncch.cpp
```diff
@@ -373,61 +200,41 @@ ResultStatus AppLoader_NCCH::Load() {
...
-        size = romfs_size;
+ResultStatus AppLoader_NCCH::ReadUpdateRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_file,
+                                             u64& offset, u64& size) {
+    ResultStatus result = update_ncch.ReadRomFS(romfs_file, offset, size);
 
-        return ResultStatus::Success;
-    }
-    LOG_DEBUG(Loader, "NCCH has no RomFS");
-    return ResultStatus::ErrorNotUsed;
+    // TODO(shinyquagsire23): Validate that this is the default behavior on hardware
```
wwylele: I did verify this is the correct behaviour when testing #2561 (and therefore I made that stub)

src/core/loader/ncch.cpp
```diff
@@ -338,23 +153,35 @@ void AppLoader_NCCH::ParseRegionLockoutInfo() {
...
     if (is_loaded)
         return ResultStatus::ErrorAlreadyLoaded;
 
-    ResultStatus result = LoadExeFS();
+    ResultStatus result = base_ncch.Load();
     if (result != ResultStatus::Success)
         return result;
 
-    std::string program_id{Common::StringFromFormat("%016" PRIX64, ncch_header.program_id)};
+    // TODO(shinyquagsire23): swapping between NCCH containers for update overlay
```
wwylele: I don't really understand this TODO. Isn't it swapped below if update exists?

wwylele:
This is currently causing crash when loading the game list. The reason is that the game list calls `loader->ReadIcon(smdh);` without calling `Load` first, which leaves `overlay_ncch` uninitialized and causes a invalid pointer dereference.

legoj15:
Just tested this on the latest canary build using SDL, and I can finally boot into Sm4sh 3DS with the savefile from my actual 3DS, everything (except citra-qt as stated above by wwylele) appears to be working from what I see.

Subv:
I actually prefer the updates to live in the same folder as they would on a real 3DS, the argument about being able to use your own SD card is also appealing.

src/core/loader/ncch.cpp
shinyquagsire23: yeah I just forgot to remove that

shinyquagsire23:
Sorted out the game list issue, seems to be working fine now on my end. Just ended up instantiating `overlay_ncch` in the constructor so that it's pointing to `base_ncch` by default and the only assignment that changes it is the assignment to `update_ncch`.

shinyquagsire23:
Apparently my update masking broke demos due to overlap with that mask, so I ended up fixing the check. I also went ahead and fixed NCCHContainer::Load so that it is able to load NCCHs which might not have an exheader or an exefs but can have a RomFS, ie some system archives will only have a romFS, DLC can have a romFS and an exeFS with only an icon.

PR #2928, opened 2017-09-16
Fixed framebuffer warning

huwpascoe:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2928)

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -237,13 +237,24 @@ void RasterizerOpenGL::DrawTriangles() {
 
     glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D,
                            color_surface != nullptr ? color_surface->texture.handle : 0, 0);
-    glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D,
-                           depth_surface != nullptr ? depth_surface->texture.handle : 0, 0);
-    bool has_stencil =
-        regs.framebuffer.framebuffer.depth_format == Pica::FramebufferRegs::DepthFormat::D24S8;
-    glFramebufferTexture2D(
-        GL_DRAW_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D,
-        (has_stencil && depth_surface != nullptr) ? depth_surface->texture.handle : 0, 0);
+    if (depth_surface != nullptr) {
+        if (regs.framebuffer.framebuffer.depth_format ==
+            Pica::FramebufferRegs::DepthFormat::D24S8) {
+            // attach both depth and stencil
+            glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D,
+                                   depth_surface->texture.handle, 0);
+        } else {
+            // attach depth
+            glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D,
+                                   depth_surface->texture.handle, 0);
+            // clear stencil attachment
+            glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D, 0, 0);
+        }
+    } else {
+        // clear both depth and stencil attachment
+        glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D, 0,
+                               0);
+    }
 
     // Sync the viewport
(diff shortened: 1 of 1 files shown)


huwpascoe:
(The warning was
``
Render.OpenGL <Warning> video_core\renderer_opengl\renderer_opengl.cpp:DebugHandler:472: API OTHER 65536: Framebuffer unsupported. Framebuffer object 3 is unsupported because the depth and stencil attachments are mismatched.
``)

yuriks:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2928#-:-Ku7tXxLnFY7mlOMwjvy:bnfp4nl)*

wwylele:
I am not a opengl expert, but I think this is a potential flaw in the new code: the game sets the buffer type to D24S8 and attaches a depth+stencil buffer. Then it sets the buffer type to D24 and disabled depths buffer. Now the function here calls `glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, 0, 0)`, leaving the stencil buffer still attached(?)

So to avoid such ambiguity and to enhance readability, I prefer the following code:
```
if (depth_surface != nullptr) {
    if (regs.framebuffer.framebuffer.depth_format == Pica::FramebufferRegs::DepthFormat::D24S8) {
        glFramebufferTexture2D(
            GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D,
            depth_surface->texture.handle, 0); // attach both depth and stencil
    } else {
        glFramebufferTexture2D(
            GL_DRAW_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D,
            depth_surface->texture.handle, 0); // attach depth
        glFramebufferTexture2D(
            GL_DRAW_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D,
            0, 0); // clear stencil attachment
    }
} else {
    glFramebufferTexture2D(
        GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D,
        0, 0); // clear both depth and stencil attachment
}
```

----

Also, I think the warning comes from mis-implementation in some drivers (thus I can't reproduce on my side). The related section in the spec I found is 9.4.3 (glspec4.3):

> The depth attachment may be in any of the required depth or combined depth+stencil formats described in those sections, and the stencil attachment may be in any of the required stencil or combined depth+stencil formats. However, when both depth and stencil attachments are present, implementations are only required to support framebuffer objects where both attachments refer to the same image.

The driver might implement "refer to the same image" as "not calling with `GL_DEPTH_ATTACHMENT` and `GL_STENCIL_ATTACHMENT` separately", despite that the attachment is actually the same object. But anyway, using `GL_DEPTH_STENCIL_ATTACHMENT` is still a more intuitive way and the change here is better.

huwpascoe:
@wwylele I wasn't aware that this was called repeatedly. There could be a bigger performance problem, in that case. I'll research this before more changes.

huwpascoe:
https://www.reddit.com/r/opengl/comments/1rsnhy/most_efficient_fbo_usage_in_multipass_pipeline/

> Do not create a single FBO and then swap out attachments on it. This causes lots of validation in the driver, which in turn leads to poor performance.

huwpascoe:
I'm going to go ahead with this commit, using wwylele's suggestion. I ran clang format so hopefully should build now.

huwpascoe:
Cancel that. It broke the rasterizer. Reverting back to the original changes.

wwylele:
@huwpascoe right... I didn't test my code so it can be broken, but I hope you can address my concern (improve the code, or explain that my concern is invalid)

huwpascoe:
Got it working now, I must have done something stupid...

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -237,13 +237,22 @@ void RasterizerOpenGL::DrawTriangles() {
...
-        regs.framebuffer.framebuffer.depth_format == Pica::FramebufferRegs::DepthFormat::D24S8;
-    glFramebufferTexture2D(
-        GL_DRAW_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D,
-        (has_stencil && depth_surface != nullptr) ? depth_surface->texture.handle : 0, 0);
+    if (depth_surface != nullptr) {
+        if (regs.framebuffer.framebuffer.depth_format ==
+            Pica::FramebufferRegs::DepthFormat::D24S8) {
+            glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D,
+                                   depth_surface->texture.handle,
+                                   0); // attach both depth and stencil
```
wwylele: You can rearrange the comment to above the code to have prettier clang-format result

Dragios:
Pokédex 3D Pro still shows the log even with this PR. I wonder why.

yuriks:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2928)*

PR #2929, opened 2017-09-16
Add QtIFW configuration and build files

j-selby:
Blocked by https://github.com/citra-emu/ext-windows-bin/pull/9

This adds configuration and build files for Qt Installer Framework binaries.

Icons are included within the configuration folder, because the binary creator is dumb and doesn't understand relative paths. The config files *could* be moved into the main `dist/` folder, but that is already disorganised and bloated.

Windows binaries can be built using `build.ps1`. No build tools are required at all, except Powershell 2.0+ (comes with Windows 10, iirc can be installed on older versions of Windows manually). 

Linux/Mac binaries can be built using Bash + Curl using `build.sh`.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2929)


dist/installer/config/config_windows.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Installer>
+    <Name>Citra</Name>
+    <Version>1.0.0</Version>
+    <Title>Citra Updater</Title>
+    <Publisher>Citra team</Publisher>
```
jroweboy: This sounds like it was ..borrowed... from what i put down for publisher in bleeding edge. Anyone care enough about it to bikeshed it before "citra team" becomes the publisher again?

j-selby: Wasn't intentionally borrowed, but I didn't think of anything better. `Team Citra` :P ?

jroweboy:
This was discussed briefly on discord, and @B3n30 and @Subv both suggest that the powershell and bash scripts should be ported to cmake. cmake should be able to support downloading the right zip, extracting it, and running the right build command

> B3n30
I personally don't like the build system for the installer in the current PR.
I think it should just be part of the cmake structure and adds another target e.g installer

>jroweboy
@B3N30 i'm not sure if this is clear, but we only build the installer once. we don't really ever need to build it again unless we need to update to a newer version
the installer reads package metadata
we don't ship a new installer with every release

>B3N30
I'm not saying we should release the installer every time we have a new release for citra
but it could still be part of the normal cmake structure
with the setting that it isn't built on default

>Subv
i agree with @B3N30

PR #2930, opened 2017-09-16
WebService: Verify username and token

B3n30:
This adds a button to the config ui, to verify the username and token. This seems necessary since most users enter their password instead of the token.  This PR adds also some error logging to the Post in WebServices.

The obligatory screenshots:

The verify button:
bildschirmfoto 2017-09-16 um 14 15 27
If verification succeeds:
bildschirmfoto 2017-09-16 um 14 15 43
If it fails:
bildschirmfoto 2017-09-16 um 14 15 53
If a user applies the settings without verification:
bildschirmfoto 2017-09-16 um 14 16 02

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2930)


src/web_service/web_backend.cpp
```diff
@@ -56,8 +56,81 @@ void PostJson(const std::string& url, const std::string& data, bool allow_anonym
...
+            if (r.status_code >= 400) {
+                LOG_ERROR(WebService, "POST returned error code: %u", r.status_code);
+                return;
+            }
+            if (r.header["content-type"].find("application/json") == std::string::npos) {
+                LOG_ERROR(WebService, "POST returned wrong content: %s",
+                          r.header["content-type"].c_str());
+                return;
+            }
+            return;
```
wwylele: this return is not necessary.

Also, some errors, for example, the ssl cert failure issue on my side (which is in the `error` field iirc), can't be caught by these error check. Same for the one in `GetJson`

src/web_service/web_backend.cpp
```diff
@@ -56,8 +56,81 @@ void PostJson(const std::string& url, const std::string& data, bool allow_anonym
...
+        header = {{"Content-Type", "application/json"},
+                  {"x-username", username.c_str()},
+                  {"x-token", token.c_str()},
+                  {"api-version", API_VERSION}};
+    } else {
+        // Otherwise, anonymous request
+        header = cpr::Header{{"Content-Type", "application/json"}, {"api-version", API_VERSION}};
+    }
+
+    // Post JSON asynchronously
```
wwylele: `GET`

src/web_service/web_backend.h
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <future>
```
wwylele: needs `<functional>`

src/citra_qt/configuration/configure_web.ui
```diff
@@ -76,6 +76,22 @@
             </property>
            </widget>
           </item>
+          <item row="3" column="1">
```
jroweboy: nitpick: i don't like using grid layout here since its not much of a grid. would rather see the two input fields above be put in a formlayout, and then put these links and verify button below it in a horizontal layout with a horizontal spacer between the links and the button, so you can have them all on the same line, but still have the button right aligned. its probably not worth the effort.

src/citra_qt/configuration/configure_web.cpp
```diff
@@ -50,3 +63,33 @@ void ConfigureWeb::refreshTelemetryID() {
     ui->label_telemetry_id->setText("Telemetry ID: 0x" +
                                     QString::number(new_telemetry_id, 16).toUpper());
 }
+
+void ConfigureWeb::loginChanged() {
```
jroweboy: bikeshed: I don't like the whole "disable the verify button if they are verified" thing. It sounds weird to have since we already have a dialog box popping up, telling the user if they validated successfully or not. Additionally, its not persisted in anyway, so closing and reopening the dialog will let you verify again.

suggestion: If you want a symbol to show that they are verified, you can get a green checkmark icon or something and stick it in dist/icons/ and put it off to the right of the `QLineEdit`s. I can give you some specific code to do this if you need some help using qt resource system (but its pretty straight forward)

src/citra_qt/configuration/configure_web.ui
jroweboy: On second thought, you *could* still stick with a grid layout. Just add a 3rd column on the right and make the `QLineEdit`s in row 0 and 1 have a `colspan="2"` Then just stick a horizontal spacer and a button in row 2 column 2 and it should look about the same as what I suggested in the previous comment.

src/citra_qt/configuration/configure_web.h
```diff
@@ -22,9 +23,18 @@ class ConfigureWeb : public QWidget {
 
 public slots:
     void refreshTelemetryID();
+    void loginChanged();
```
jroweboy: nitpick: In general, qt slots are present tense verbs, and signals are past tense. we are kinda mixing them together (but probably should be a little less lenient about it. For a naming suggestion i think this should be changed `loginChanged` => `EnableVerify` 

style: method names are `CamelCase`

src/citra_qt/configuration/configure_web.cpp
```diff
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <QMessageBox>
```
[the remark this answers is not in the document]
wwylele: @jroweboy that's old style before clang-format, which prevents clang-format from doing proper ordering. Our current style is not to add the space

src/citra_qt/configuration/configure_web.h
wwylele: > style: method names are `CamelCase`

This doesn't necessarily apply to qt related functions since their style is `camelCase`

lioncash: Unless the case is overriding a Qt API's member function, it should generally still be `CamelCase`.

src/web_service/web_backend.cpp
```diff
@@ -56,8 +56,81 @@ void PostJson(const std::string& url, const std::string& data, bool allow_anonym
...
+        return std::async(std::launch::async, [func{std::move(func)}]() { return func(""); });
+    }
+
+    const bool are_credentials_provided{!token.empty() && !username.empty()};
+    if (!allow_anonymous && !are_credentials_provided) {
+        LOG_ERROR(WebService, "Credentials must be provided for authenticated requests");
+        return std::async(std::launch::async, [func{std::move(func)}]() { return func(""); });
+    }
+
+#ifdef _WIN32
```
lioncash: This is basically a copy/paste of the same code from above. This should likely be put into its own function.

B3n30:
Improved the ui:
bildschirmfoto 2017-09-17 um 17 39 28
bildschirmfoto 2017-09-17 um 17 39 42
The message box on succeed was removed since it seemed redundant

PR #2933, opened 2017-09-17
Improved performance of FromAttributeBuffer

huwpascoe:
Ternary operator is optimized by the compiler
whereas std::min() is meant to return a value.

I've noticed a 5%-10% emulation speed increase*.

*In Pokemon Moon. Compiled with Visual Studio 2017

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2933)

### DIFF
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -52,7 +52,8 @@ OutputVertex OutputVertex::FromAttributeBuffer(const RasterizerRegs& regs,
     // The hardware takes the absolute and saturates vertex colors like this, *before* doing
     // interpolation
     for (unsigned i = 0; i < 4; ++i) {
-        ret.color[i] = float24::FromFloat32(std::fmin(std::fabs(ret.color[i].ToFloat32()), 1.0f));
+        float c = std::fabs(ret.color[i].ToFloat32());
+        ret.color[i] = float24::FromFloat32(c < 1.0f ? c : 1.0f);
     }
 
     LOG_TRACE(HW_GPU, "Output vertex: pos(%.2f, %.2f, %.2f, %.2f), quat(%.2f, %.2f, %.2f, %.2f), "


huwpascoe:
perf-1

chris062689:
What are the performance differences in MinGW vs MVSC?

Subv:
I compiled this and compared it with the previous version using MSVC2017 and here's the result:

https://gist.github.com/Subv/53c5aca853fa974988394367f0708410

We can see that the previous version emitted a call to fminf and kept on looping. The new version unrolls the loop and uses the MINSS instruction.

I do not understand why MSVC decided not to vectorize the instructions, I reckon we can process the entire array in 4 SSE vector instructions.

I also do not know why MSVC decided to use cvtps2pd on each value just to apply the mask and then convert back to single using cvtpd2ps. It would be interesting to see what code GCC generates for both versions of the C++ code.

Subv:
Some more insight!

This is a godbolt snippet with output for Clang, GCC and MSVC. Clang's output seems to be the best one, packing everything into 4 SSE instructions. GCC seems to generate a loop and MSVC packs the values in groups of 2 but still does the Single->Double->Single conversion.

https://godbolt.org/g/AEvUaL

Using std::fmin instead of the ternary operator in godbolt didn't change GCC's output. MSVC output got much closer to the gist i pasted in my previous comment, making use of the minss instruction (but doing it one at a time (no vectorization)), which is weird because in my local machine, using std::fmin just emits a call to the fmin implementation.

Clang's output got *much* more convoluted, so I'd say this change is appreciated and i give it my LGTM.

:+1:

B3n30:
I'm fine with this speed improvement since it isn't a big change a doesn't make followup PRs more complex. 

Since all the compilers treat this part differently, I would love to see a profiling for each build we have, before and after that PR, to see if we don't make it worse for any of our builds.

PR #2936, opened 2017-09-19
WebService: Set USE_SYSTEM_CURL for travis linux builds

B3n30:
Fixes issue that linux builds can't send telemetry.

It could be necessary that users have to install libcurl on their system for citra to work:
`apt-get install libcurl4-openssl-dev`

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2936)


wwylele:
Tested canary build on ubuntu 16.04 and it works

Issue #1250, opened 2015-11-23
Fix GDB stub usability


neobrain:
Currently, having the GDB stub enabled will cause the emulator to wait for a GDB connection whenever a game is being started. A notification message will be printed to the console, but for all purposes the console should not be considered a valid method of user notifications like this.

The current behavior is unacceptable because

* End users will enable the GDB stub unintentionally, and will notice no immediate change until they try to run a game. At that point, they will be completely lost as to what's happening, and will not be able to figure out how to resolve the issue. Instead, they will probably blame the emulator. You may be inclined to think this would only affect "dumb" users, but that's simply not true, and even then it is unacceptable for a proper GUI to lead users into such situations.
* Since console messages are unlikely to be caught be the user, there is essentially no visual feedback at all. This adds to the confusion that an end user (or even developer!) would experience when waiting for the game to launch.

As a solution, I propose a modal dialog to be spawned while waiting for the GDB stub. The dialog would provide the following functionality:

* Inform the user about what's happening
* Provide a way to start emulation without waiting for the GDB stub (leaving the GDB stub enabled in future runs)
* Provide a way to cancel emulation and return to the main application window (leaving the GDB stub enabled in future runs)
* Disable the GDB stub (this could be done by adding a checkbox to the dialog saying something like "Leave GDB stub disabled")

An alternative option (which is highly preferable, but probably requires a better architecture than we have currently) is to wait for GDB connections in a thread rather than blocking emulation until a connection is received. Having GDB connected to the emulator from the very start of emulation would be achieved by providing the user with a way to initialize the emulator without actually starting emulation (so they could initialize, connect GDB, and then start emulation).

RavenHome1:
i can relate https://github.com/citra-emu/citra/issues/1249

neobrain:
Testing stuff, ignore this post.

shinespeciall:
If it is necessary to help those who don't know about the GDB stub but accidentally select the checkbox, why not simply show a messagebox to explain and ask for another confirmation before start the GDB stub?
   Besides, It seems more function need to be added for multi-thread debugging or just initial the debugger.

PR #2942, opened 2017-09-21
A simple solution to Issue #1250

shinespeciall:
####    I just disable the gdb stub record in config file and disable it every time citra starts to stop user from enable the GDB stub unintentionally, debugging needs time and we don't need to configure it frequently. a toolTip will give user some useful information to end user as a referrence.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2942)


Subv:
I'm not particularly against this change, in fact, I've been burned by this several times before.

It is worth noting that this PR only affects the Qt frontend.

JayFoxRox:
I'd rather have this "expert"-option hidden (in the GUI, possibly disabled entirely) behind a config-file-only option ("export_mode=1").

I'd prefer if I can configure my Qt GUI to startup with gdb easily connectable.
Because *if* I have to debug, I'll want to restart the session quite often (between code changes).
So going through the GUI on every restart, just to re-configure gdb, is not an option as it would hinder workflow.

jroweboy:
@shinespeciall I apologize for taking so long to get around to this. I appreciate the code change and agree that we should fix the problem, but I don't think this solution is the right way to fix it. There are game modders who pretty much exclusively use the gdb stub, and developers who are fixing game issues that will use the gdb stub, and forcing them to enable it on every load is very time consuming and will just be annoying.

I honestly can't think of a solution that fixes every issue, but this PR would make it harder for the people that actually want to use it, so I'm just going to close it. Perhaps the better solution is to have the gdb stub listen on a separate thread that doesn't block the emulation. This could cause threading issues so I'm not sure if this is a good idea or not, but it would at least solve the problem where a user turns it on and it makes "citra freeze" Anyway, I don't see this getting merged, so I'll close it for now. Thanks again for the contribution!

PR #2943, opened 2017-09-21
Extract functions on Command Processor

huwpascoe:
Using the fastest method, which turns out to be an array of c-style function pointers (after testing with std::function, etc.)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2943)


src/video_core/command_map.h
```diff
@@ -0,0 +1,526 @@
...
+void UniformFloatVs(const Command&);
+void SetWordGs(const Command&);
+void SetWordVs(const Command&);
+void SetWordSwizzleGs(const Command&);
+void SetWordSwizzleVs(const Command&);
+void LightingLut(const Command&);
+void FogLut(const Command&);
+void ProcTexLut(const Command&);
+
+typedef void (*CommandAction)(const Command&);
```
lioncash:
```cpp
using CommandAction = void (*)(const Command&);
```

src/video_core/command_map.h
```diff
@@ -0,0 +1,526 @@
...
+void nop(const Command&) {}
+
+#define _(id) nop
```
lioncash: Identifiers beginning with underscores are reserved by the standard. Also, don't use [macros](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es30-dont-use-macros-for-program-text-manipulation) for program text manipulation.

src/video_core/command_processor.cpp
```diff
@@ -32,24 +33,26 @@ namespace Pica {
 
 namespace CommandProcessor {
 
-static int vs_float_regs_counter = 0;
-static u32 vs_uniform_write_buffer[4];
+int vs_float_regs_counter = 0;
```
lioncash: Why do these variables need to be non-static? They're still only used internally.

src/video_core/command_processor.cpp
```diff
@@ -6,6 +6,7 @@
 #include <cstddef>
 #include <memory>
 #include <utility>
+#include "command_map.h"
```
lioncash: This should be included as `"video_core/command_map.h"`

src/video_core/command_processor.cpp
huwpascoe: Why do they need to be static?

src/video_core/command_map.h
huwpascoe: I was basing it on the current code which does exactly the same.
#define PICA_REG_INDEX_WORKAROUND(field_name, backup_workaround_index) (backup_workaround_index)

huwpascoe:
> Just saying "it's faster" alone doesn't really justify the maintenance burden IMO.

A "maintenance burden"? Compared to a massive switch statement? Don't talk BS. : P

chris062689:
That still doesn't address what type of systems you've tested this on, games, before and after statistics. 
We need proof. :)

huwpascoe:
@chris062689 Don't worry, you can [clone this PR locally](https://stackoverflow.com/questions/14947789/github-clone-from-pull-request) to test it.

src/video_core/command_processor.cpp
lioncash: If they're only ever used in the same translation unit, they're marked static to signify internal linkage. When applied to functions, it also enables compilers to potentially inline said functions within the translation unit.

lioncash:
Generally the onus is on the one opening the PR to provide the performance information

src/video_core/command_map.h
lioncash: That exists only as a workaround to a bug in the Visual Studio 2015 compiler (said bug may already be fixed in VS 2017). It's the only reason the constants were hardcoded.

huwpascoe:
Amendments:
```
static, static everywhere
Removed the macros although command_map.h is now a lot longer with those unknown regs expanded.
Removed changes to command_processor.h
Included new file in cmakelists.txt because that would help if it could actually compile.
```
I think there's a misunderstanding though. I don't claim this will improve performance, I don't know if it will. I was assured in discord that this refactor was the right step towards future performance increases however.

B3n30:
Basically, your code shouldn't really change anything on the compiled version. The compiler should compile the switch statement as well as your function array into a branch table. So there is no `right step towards future performance increases`. But I haven't checked if our compilers really do that. That is probably something that should get checked before merging this PR.

If it is compiled to a branch table than I think that PR is an improvement in case of readability and we use function arrays in other places inside citra as well. So it's just consistency.

Subv:
@B3n30 MSVC won't compile the switch into a jump table because it is _very_ sparse, I don't know what the generated code looks like for this PR, but i doubt it'd be much worse than a bunch of jumps strung together

src/video_core/command_processor.cpp
```diff
@@ -119,495 +126,403 @@ static void WriteUniformFloatReg(ShaderRegs& config, Shader::ShaderSetup& setup,
...
-                g_state.gs.swizzle_data[offset] = value;
-            }
-            offset++;
-        }
-        break;
+static void WritePicaReg(const Command& cmd) {
+    auto& regs = g_state.regs;
+
+    if (cmd.ID >= Regs::NUM_REGS) {
+        LOG_ERROR(HW_GPU,
```
B3n30: Is this how the HW reacts, just ignore the write, or does it crash?

src/video_core/command_map.h
```diff
@@ -0,0 +1,810 @@
...
+static void SetWordGs(const Command&);
+static void SetWordVs(const Command&);
+static void SetWordSwizzleGs(const Command&);
+static void SetWordSwizzleVs(const Command&);
+static void LightingLut(const Command&);
+static void FogLut(const Command&);
+static void ProcTexLut(const Command&);
+
+using CommandAction = void (*)(const Command&);
+static const CommandAction CommandMap[] = {
```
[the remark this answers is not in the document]
huwpascoe: This might make for an extremely long .cpp though?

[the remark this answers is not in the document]
Subv: Currently it is a static variable in a header file, every translation unit that includes this file will get its own CommandMap array, potentially bloating the generated code.

huwpascoe:
I can't see this getting merged, maybe too big to handle. Will rethink.

PR #2945, opened 2017-09-23
citra-qt: basic translation support

wwylele:
The idea of this PR is to setup a minimal platform for citra translation development, so that we can get people involved in the translation work early and gather more ideas and feedbacks from them, from which we can decide our official translation workflow.

The workflow in this PR is:
 - install Qt linguist tool (not required for general citra devs)
 - config cmake with `GENERATE_QT_TRANSLATION` ON
   - Note: for windows/msvc, because the linguist tool is not bundled in our Qt package, one need to download it and manually set the cmake file path when being prompted error
 - build target `translation`. This generates `translation.ts` file in `src/citra_qt/` and `translations.qm` in `build` directory.
   - command `make translation` for linux. 
   - for Visual Studio, a new project `translation` will be added to the solution. Just build the project.
   - MacOS not tested yet
 
 - do translation work in `translation.ts`, and build target `translation` again to update `translation.qm`
 - in case of upstream code updated, pull the repo and build target `translation` to update  `translation.ts`
 - eventually release `translation.qm` as a translation patch file
 - citra user downloads the patch file and loads it into citra.

citra-tran
Here is an example with one menu item translated

[Here](https://cdn.discordapp.com/attachments/242442830486700045/361483002078953482/citra-zh.qm) is an near-complete chinese translation patch. You can use this to test if you don't want to do the translation yourself.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2945)


RavenHome1:
i like this idea but is there really a need to have citra support other languages?

einstein95:
Given the amount of posts made on Chinese forums, I'd say so. Also, it's always nice to have a program in your native language rather than English.

jroweboy:
Some thoughts:

* Translations are shipped separately from citra and will need to be `Loaded` in order to be chooseable as a language
  * Which implies that updating a translation means users will then need to redownload the updated translation file from the web
  * We should do a little research to see if the installer can handle this
* Translators will need to put the qm files... where exactly?
* No easy way to vet or review translation changes right now?

But for a start, its more than welcome. We should probably start talking about the final design so we can break up the work into mergable chunks :) Thanks for the PR!

wwylele:
@jroweboy 
> Translators will need to put the qm files... where exactly?

Anywhere, and the user can choose the path with a file open dialog

> No easy way to vet or review translation changes right now?

This is what I don't know how to do. My thought is to let the translation developer figure out their own method to do source control and review (maybe a bot fetching ts file, a server hosting translation work, etc.), and release their "unofficial" patches right now. We can then find the best solution with idea from all people and make it "official".

Subv:
> No easy way to vet or review translation changes right now?

Maybe we can use a service like Transifex to make the reviewing and packaging easier

B3n30:
LGTM

Just some thoughts about it: For now I think it is ok, that there is a file_select dialog for changing language, but if we ship the translation with the installer/updater we should probably just list all the files in a specific folder where all the qm files are placed.

For the review/release part: What I think would be great, is to have the ci build the ts file, then let the user edit this files (maybe on our website?) and if they submit their changes this creates a PR in a github repo where we collect the ts files. If the PR is merged the ci builds the qm file for that language. And then the installer/updater can access that repo for downloading the qm files.

Though I'm not sure if every language would need their own repo (which is something I don't like).

wwylele:
> Just some thoughts about it: For now I think it is ok, that there is a file_select dialog for changing language, but if we ship the translation with the installer/updater we should probably just list all the files in a specific folder where all the qm files are placed.

Yes. As I planned, once we start to accept some official translation files, they can be bundled with citra and listed in the menu directly.

huwpascoe:
:lgtm:

---

Reviewed 5 of 5 files at r1, 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2945#-:-KvJdsrUEPz4qYg8o1j3:bnfp4nl)*

wwylele:
Since there is no opinion against this, I am taking off RFC.

MerryMage:
LGTM assuming:

* We will have a defined location for *.qm files we're going to bundle in the future
* Other bundled languages will go in the UI where the English menu item is now

wwylele:
>We will have a defined location for *.qm files we're going to bundle in the future
>Other bundled languages will go in the UI where the English menu item is now

Yes, that's what I want to achieve eventually

lioncash:
FWIW, like @Subv mentioned, Transifex is a good place to centralize and review translations. It's also what Dolphin uses for its [own](https://www.transifex.com/delroth/dolphin-emu/) translations as well. Essentially you let translation work happen for some amount of time until the work done is substantial enough, then you use Transifex's tool [tx](https://docs.transifex.com/client/introduction) to synchronize the changes by running [tx pull](https://docs.transifex.com/client/pull). To update modified resources, you just use [tx push](https://docs.transifex.com/client/push).

This is something that could also be build infrastructure-driven. e.g. After N days do a `tx pull` and if any files are changed, automatically commit the updates.

wwylele:
Thanks for the advise. I am very busy this month and don't have time to think about this, so I will leave this open and anyone comes up with a better plan can take over it.

> Wouldn't a better way of doing this be to just have a dedicated languages directory right now and iterate over it on program load and add each available language to the languages menu?

Yes, this sounds good to me. However, I am not sure which directory should I put. Should I put it to the user directory, or the "executable directory"? The former one sounds a little bit weird to me, since this is not save data or configuration file. For the latter one, I am not familiar with the executable directory on Linux or MacOS, and don't know what is the correct way to put them. 

> This way you don't need to rewrite any loading code in the future, and people can just drop translation files into the directory and it works out of the box without any needed interaction.

I envision that the official translations will be eventually packaged as qt resources, which means the code will be modified anyway...

einstein95:
Ideally the files should be placed in the configuration directory as the executable directory requires root/administrator access to place files in. This also makes it easier to tweak files when it comes to updates or to test changes.

lioncash:
> Yes, this sounds good to me. However, I am not sure which directory should I put. Should I put it to the user directory, or the "executable directory"?

The executable directory (at least on Windows and macOS), since languages are global assets to the application regardless of the user configuration (which would just store which of those languages to use).

> For the latter one, I am not familiar with the executable directory on Linux or MacOS, and don't know what is the correct way to put them.

If we bundle our application on the macOS build, `QApplication::applicationDirPath()` will give the directory to the application within the bundle (e.g. `/Applications/citra-emu.app/Contents/MacOS`)

The general form a macOS application is bundled is like so:

```
NameOfApp.app/
╚ Contents/
  ╚ Info.plist
  ╚ MacOS/
    ╚ NameOfApp <- This is the actual executable
  ╚ Resources/ <- This is where assets generally go (languages, graphics, etc.)
```

If it's not bundled, it goes just points to the directory the executable is in, which can still be used to find the resource dir.

I'm not sure what the preferred organization is on Linux, but @linkmauve, or someone else who knows Linux better than me can tell you.

So you'd just need a function like:

```cpp
static QString GetLanguagesPath()
{
#if defined(Q_OS_WIN)
    return QApplication::applicationDirPath() + "/Languages/";
#elif defined(Q_OS_OSX)
    return QApplication::applicationDirPath() + "/../Resources/Languages";
#elif defined(Q_OS_LINUX)
    return QApplication::applicationDirPath() + "Whatever path out of the 42069 paths *nix uses to organize its files)";
#else
    // Some default
    return QApplication::applicationDirPath() + "/Languages";
#endif
}
```

acnleditor2:
If I add this to a Citra with cheats, how I can translate the cheats option?

acnleditor2:
filter bar dont translates.
not found system archives messages boxes dont translates.
it is necessary to translate messages from the status bar. I tried but it does not translate
and not everything translates after loading the translation, you must restart Citra.

translation (spanish, yes, with a lot of things): 
[Change extension to .qm](https://github.com/citra-emu/citra/files/1396918/Espanol.-.Change.extension.to.qm.zip)

wwylele:
I am currently trying Transifex and it looks promising. I will update this / open a new one when I get a solid setup

wwylele:
Closing as I am going to open a new one

PR #2843, opened 2017-07-22
 Services/APT: Use an array to hold data about the 4 possible concurrent applet types (Application, Library, HomeMenu, System)

Subv:
This gives each applet type its own set of events as per the real NS module.

This depends on #2840 , reviewers should focus on the last commit only.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2843)

PR #2946, opened 2017-09-23
Implement PrepareToStartNewestHomeMenu and fixed an APT regression.

Subv:
This PR fixes a regression from #2843 and implements the trivial PrepareToStartNewestHomeMenu so that the Home Menu can boot.

With this PR, the Home Menu boots but crashes a few seconds later due to it calling WaitSynchronization1 on a ClientSession handle leftover in the command buffer from an unimplemented nim:u operation.

This PR can be reviewed on a per-commit basis.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2946)


src/core/hle/service/apt/apt.cpp
```diff
@@ -197,6 +203,19 @@ void Initialize(Service::Interface* self) {
...
+        slot_data->applet_id == AppletId::HomeMenu) {
+        // Initialize the APT parameter to wake up the application.
+        next_parameter.emplace();
+        next_parameter->signal = static_cast<u32>(SignalType::Wakeup);
+        next_parameter->sender_id = static_cast<u32>(AppletId::None);
+        next_parameter->destination_id = app_id;
+        // Not signaling the parameter event will cause the application (or Home Menu) to hang
+        // during startup. In the real console, it is usually the Kernel and Home Menu who cause NS
+        // to signal the HomeMenu and Application parameter events, respectively.
+        slot_data->parameter_event->Signal();
```
B3n30: This wasn't exactly what I had in mind when I suggested the solution in discord. I'm not sure this is the correct place to signal. Is this really the place where 3DS signals that event?

src/core/hle/service/apt/apt.cpp
```diff
@@ -158,6 +158,12 @@ static AppletSlotData* GetAppletSlotData(AppletAttributes attributes) {
     if (slot == AppletSlot::Error)
         return nullptr;
 
+    // The Home Menu is a system applet, however, it has its own applet slot so that it can run
+    // concurrently with other system applets.
+    static constexpr u32 HomeMenuAttributeMask = 0x20000000;
```
wwylele: Can't this be defined as a `BitField` of attributes?

src/core/hle/service/apt/apt.cpp
Subv: @B3n30 The real NS module sends the Wakeup message and signals the event inside the APT:StartApplication / APT:StartNewestHomeMenu function, but that's never going to get called when loading a game via the Citra interface.

Subv:
Addressed @wwylele 's comment

src/core/hle/service/apt/apt.cpp
B3n30: Can't we just call APT:StartApplication when citra loads a game? And signal it from there?

Subv: There's an entire state machine dedicated to doing this in the 3DS, and not all applications are started in the same way (The Home Menu won't need StartApplication to be called, for example. It uses a different function which does different things). IMO that's overengineering it for now. We'll eventually need to implement those functions if we plan on allowing the Home Menu to boot up games though

B3n30:
LGTM

PR #2947, opened 2017-09-24
HLE/Archives: Allow multiple loaded applications to access their SelfNCCH archive independently.

Subv:
The loaders now register each loaded ROM with the SelfNCCH factory, which keeps the data around for the duration of the emulation session.

When opening the SelfNCCH archive, the factory queries the current program's programid and uses that as a key to the map that contains the NCCHData structure (RomFS, Icon, Banner, etc).

This is another PR on the Road To Multiprocess (TM)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2947)

### DIFF
--- a/src/core/file_sys/archive_selfncch.cpp
+++ b/src/core/file_sys/archive_selfncch.cpp
@@ -3,12 +3,14 @@
 // Refer to the license.txt file included.
 
 #include <array>
+#include <cinttypes>
 #include "common/common_types.h"
 #include "common/logging/log.h"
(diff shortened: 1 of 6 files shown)


Subv:
I have updated the PR to fix a regression when loading 3dsx files with RomFS.

Since 3dsx files do not have a programid, they'll use a default value of 0, thus only one 3dsx can be loaded at the same time.

B3n30:
Can you fix the conflict, so that we can merge this?

PR #2949, opened 2017-09-24
citra-qt: fix some untranslated strings

wwylele:
This doesn't fix all untranslated strings, though. There are some strange problem with game_list.cpp where adding missing tr() generates `tr cannot be called without context` warnings. Qt might gets confused with multiple/nested class there.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2949)


jroweboy:
> tr cannot be called without context

Is it referring to http://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments where you need to add some disambiguation comments? Hmm.. never seen that error before...

PR #2950, opened 2017-09-24
[Qt] toolbar

Styleoshin:
I added a toolbar for citra-qt

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2950)


jroweboy:
Please include some screen shots of what the tool bar looks like in the PR description. Also some tips on what icons and features you expect the toolbar to be used for would be wonderful as well.

MerryMage:
Hi! Thanks for your contribution.

Are the images self-created?

Simonx22:
Windows Screenshots
Dark: screenshot_1
Default: 
screenshot_3

Styleoshin:
@jroweboy : I wanted to add the toolbar for ease of use, it has features that are used most often :
1. Load game.
2. Toggle screen if the user would like to play in fullscreen/norma.l
3. Play/pause emulation.
4. Stop emulation.
5. Open setting.
The toolbar is movable on the sides if the current position disturb.

@MerryMage : no, i downloaded in http://www.iconsplace.com/

@Simonx22  thank's for screens

jroweboy:
Did a little chatting with some of the other contributors, and the general feelings about adding a toolbar are... mixed. I'm pretty happy with the idea of a toolbar, and i think it'll be great to put the most common ones on there as citra gets more and more options over time. Others that I talked to are against having a tool bar because citra has so few options anyway.

So before I review this, I'd like to be sure that the idea is welcome. Otherwise, we'll just be wasting everyone's time.

src/citra_qt/CMakeLists.txt
```diff
@@ -79,6 +79,7 @@ set(UIS
             main.ui
             )
 
+file(GLOB_RECURSE TOOLBAR_ICONS ${CMAKE_SOURCE_DIR}/dist/toolbar_icons/*)
```
B3n30: We have the icon folder. I don't think adding a special folder for toolbar_icons is necessary.
Better in a subfolder dist/icons/toolbar

DerProGamer2000:
i would be interested in a toolbar :)

Luffykun007:
me too

superheri:
For me having a toolbar like that is a bit like the dolphin. Useful to get most used commands easily and the dolphin's one is not really customizable and works well. So why not adding it?

jroweboy:
I personally am fine with a toolbar, but I don't think citra has enough options to warrant one right now. I appreciate the effort that went into this though, and maybe when multiplayer is merged, and we get some more things to put on the toolbar, it might be good to revisit this.

PR #2951, opened 2017-09-24
Optimized Morton

huwpascoe:
It is now better. The modulo also make it clearer (which are optimized to &7 anyway)
https://godbolt.org/g/W6VQt1

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2951)

### DIFF
--- a/src/video_core/utils.h
+++ b/src/video_core/utils.h
@@ -8,17 +8,11 @@
 
 namespace VideoCore {
 
-/**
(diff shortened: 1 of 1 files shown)


chris062689:
What effect does this have on existing test cases? Was this hardware tested?

What type of speed performance did you experience with both MVSC and MinGW before / after?

PR #2952, opened 2017-09-24
Switchable Page Tables

MerryMage:
Restores the page table optimization in the CPU JIT.

Completes [this card](https://github.com/citra-emu/citra/projects/3#card-4172636).

* `Core::CPU().LoadContext` must be called after `Memory::SetCurrentPageTable`, so the CPU state is set to the correct instance.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2952)

### DIFF
--- a/src/core/arm/dynarmic/arm_dynarmic.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic.cpp
@@ -41,7 +41,7 @@ static bool IsReadOnlyMemory(u32 vaddr) {
 }
 
 static Dynarmic::UserCallbacks GetUserCallbacks(
-    const std::shared_ptr<ARMul_State>& interpeter_state) {
+    const std::shared_ptr<ARMul_State>& interpeter_state, Memory::PageTable* current_page_table) {
     Dynarmic::UserCallbacks user_callbacks{};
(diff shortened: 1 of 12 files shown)


src/core/arm/dynarmic/arm_dynarmic.cpp
```diff
@@ -178,3 +178,16 @@ void ARM_Dynarmic::PrepareReschedule() {
...
+    current_page_table = Memory::GetCurrentPageTable();
+
+    auto iter = jits.find(current_page_table);
+    if (iter != jits.end()) {
+        jit = iter->second.get();
+        return;
+    }
+
+    jit = new Dynarmic::Jit(GetUserCallbacks(interpreter_state, current_page_table));
+    jits.insert(std::make_pair(current_page_table, std::unique_ptr<Dynarmic::Jit>(jit)));
```
lioncash: This can utilize std::map's emplace(), no?

```cpp
 jits.emplace(current_page_table, std::unique_ptr<Dynarmic::Jit>(jit));
```

MerryMage: Thanks, LC.

PR #2953, opened 2017-09-25
HLE/APT: Always set up the APT parameter when starting a library applet.

Subv:
Only use the HLE interface if an HLE applet with the desired id was started.

This commit reorganizes the APT code surrounding parameter creation and delivery to make it easier to support LLE applets in the future.

As future work, the HLE applet interface can be reworked to utilize the same facilities as the LLE interface.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2953)


src/core/hle/service/apt/apt.cpp
```diff
@@ -746,57 +744,72 @@ void PrepareToStartLibraryApplet(Service::Interface* self) {
...
+    Memory::ReadBlock(buffer_addr, param.buffer.data(), param.buffer.size());
+    SendParameter(param);
+
+    // In case the applet is being HLEd, attempt to communicate with it.
+    if (auto applet = HLE::Applets::Applet::Get(applet_id)) {
+        AppletStartupParameter parameter;
+        parameter.object = Kernel::g_handle_table.GetGeneric(handle);
+        parameter.buffer.resize(buffer_size);
+        Memory::ReadBlock(buffer_addr, parameter.buffer.data(), parameter.buffer.size());
+        rb.Push(applet->Start(parameter));
```
wwylele: Can't you just send `param` here instead of setting up another `parameter`?

Subv: They are different types, a rewrite of the HLE interface to make it more similar to the LLE one is in my TODO list

Luffykun007:
@Subv  Works again thanks!

PR #2954, opened 2017-09-25
Memory/RasterizerCache: Ignore unmapped memory regions when caching physical regions

Subv:
Not all physical regions need to be mapped into the address space of every process, for example, system modules do not have a VRAM mapping.

This fixes a crash when loading applets and system modules.

In the future we'll have to figure out a way to update the page tables of other processes when caching / uncaching a region.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2954)

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -316,8 +316,15 @@ void RasterizerMarkRegionCached(PAddr start, u32 size, int count_delta) {
 
     for (unsigned i = 0; i < num_pages; ++i, paddr += PAGE_SIZE) {
         boost::optional<VAddr> maybe_vaddr = PhysicalToVirtualAddress(paddr);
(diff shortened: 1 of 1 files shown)


B3n30:
How does a 3ds behave if a app tries to access some unmapped VRAM?
Shouldn't it normally fail and only ignore it for system modules?

Maybe at least add some log, to make a search for errors possible?

Subv:
Trying to access an unmapped memory region from the CPU will trigger an exception in the 3DS, the exception handler will then just close the app

huwpascoe:
👍

huwpascoe:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2954)*

PR #2955, opened 2017-09-25
Multi thread proof of concept (Do not merge)

huwpascoe:
Has the dependency boost::lockfree which is not part of the externals.

This shows that the GPU core is able to run separately from the CPU core with little modification. I haven't made any attempt to optimize or understand the hardware, but hopefully this can give some inspiration.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2955)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <memory>
+#include <thread>
 #include <utility>
 #include "audio_core/audio_core.h"
 #include "common/logging/log.h"
@@ -26,6 +27,11 @@ namespace Core {
 
 /*static*/ System System::s_instance;
 
+std::thread CpuThread;
+std::atomic<bool> CpuThreadRunning;
+std::atomic<bool> RunCPU;
+std::atomic<int> TightLoop;
+
 System::ResultStatus System::RunLoop(int tight_loop) {
     status = ResultStatus::Success;
     if (!cpu_core) {
@@ -47,6 +53,15 @@ System::ResultStatus System::RunLoop(int tight_loop) {
         }
     }
 
+    TightLoop += tight_loop;
+    RunCPU = true;
+
+    HW::Update();
+
+    return status;
+}
+
+void System::RunLoopInner() {
     // If we don't have a currently active thread then don't execute instructions,
     // instead advance to the next event and try to yield to the next thread
     if (Kernel::GetCurrentThread() == nullptr) {
@@ -55,13 +70,20 @@ System::ResultStatus System::RunLoop(int tight_loop) {
         CoreTiming::Advance();
         PrepareReschedule();
     } else {
-        cpu_core->Run(tight_loop);
+        cpu_core->Run(TightLoop);
     }
 
-    HW::Update();
+    TightLoop = 0;
     Reschedule();
+}
 
-    return status;
+void System::RunLoopInThread() {
+    while (CpuThreadRunning) {
+        if (RunCPU) {
+            RunLoopInner();
+            RunCPU = false;
+        }
+    }
 }
 
 System::ResultStatus System::SingleStep() {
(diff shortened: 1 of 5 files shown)


wwylele:
I think the main concern about separate GPU into its own thread is that it can introduce indeterminacy, as it does in this easy implementation: the cycle of GPU thread to complete one task is not strongly tied to `CoreTiming` anymore, and depends on different hardware performance, the same game may behave different on different computes.

I don't have solid opinion to this yet, so I am just throwing here some questions with which we can discuss:

 - do we want to keep determinism? 
 - do we want accuracy? i.e. emulate the actual latency of a real 3DS GPU. Is it even possible to accurately measure the latency?
 - is undetermined latency considered more accurate than no latency?
 - can we have an option to switch between single thread/multithread?

lioncash:
> the same game may behave different on different computes.

If this is the case, then the implementation should likely be improved to mitigate that, or be avoided for the time being. I can see this being kind of annoying to build a FifoCI analog around, if one is desirable (it should be), since consistent behavior is a necessity for accurate comparisons between changes. 

> do we want to keep determinism?

If we plan to support TASing in the future (or if someone else would like to add that feature), we'd likely want determinism.

huwpascoe:
>do we want to keep determinism?
The VBlank Interrupt is probably important in this case. But are there other situations where the CPU has to wait for the GPU to send data back? If the model is entirely producer-consumer then the CPU thread decides what's what.

>can we have an option to switch between single thread/multithread?
That would be easy enough to do.

wwylele:
> But are there other situations where the CPU has to wait for the GPU to send data back?

Yes, CPU can read back what GPU has rendered. For example, as I observed it is likely that in Monster Hunter games fetch back the rendered framebuffer and do an overall color balance on CPU side before displaying. Pokemon games might has CPU and GPU working together to draw the outlines as well.

BreadFish64:
after testing Pokemon sun
At best, this implementation is (rarely) 10%ish faster.
Usually it's slower by up to 40-50%

emmauss:
@BreadFish64  this isnt a PR for test. it is a proof of concept, showing that it is possible, but there havve not been anything implemented that takes full advantage of multithreading.

BreadFish64:
@emmauss I'm aware, I just tried it for fun and @wwylele told me to drop the results here

src/core/core.cpp
```diff
@@ -55,13 +70,20 @@ System::ResultStatus System::RunLoop(int tight_loop) {
         CoreTiming::Advance();
         PrepareReschedule();
     } else {
-        cpu_core->Run(tight_loop);
+        cpu_core->Run(TightLoop);
```
MerryMage: Race condition: `TightLoop` may be advanced between this statement and the next. Setting `TightLoop = 0;` thus seems like incorrect behaviour.

src/core/hw/gpu.cpp
```diff
@@ -28,6 +29,14 @@ namespace GPU {
 
 Regs g_regs;
 
+struct WriteThing {
+    u32 addr;
+    u32 data;
+};
+
+std::atomic<bool> VBlankGo = false;
+boost::lockfree::queue<WriteThing> WriteQueue(128);
```
MerryMage: Performance note: Prefer `boost::lockfree::spsc_queue` in this case.

src/core/hw/gpu.cpp
```diff
@@ -565,6 +578,17 @@ void Init() {
     LOG_DEBUG(HW_GPU, "initialized OK");
 }
 
+void Update() {
+    WriteThing thing;
+    while (WriteQueue.pop(thing)) {
```
MerryMage: Performance note: Prefer `spsc_queue::consume_all`.

src/core/hw/gpu.cpp
```diff
@@ -565,6 +578,17 @@ void Init() {
...
+        WriteE<u32>(thing.addr, thing.data);
+    }
+    if (VBlankGo) {
```
MerryMage: Race condition: It is possible for there to be an elements in the write queue that should have been executed after the VBlank.

Prefer to use the queue for all communication: i.e.: place a sentinel value on the queue.

src/core/core.cpp
```diff
@@ -47,6 +53,15 @@ System::ResultStatus System::RunLoop(int tight_loop) {
         }
     }
 
+    TightLoop += tight_loop;
```
MerryMage: Note: It may be possible for `TightLoop` to grow unboundedly.

Subv:
Signaling the GPU interrupts from the GPU thread can cause race conditions, we have an HLE mutex for cases like this.

B3n30:
Closed, since it is only POC

PR #2958, opened 2017-09-25
Audio: Use std::deque instead of std::vector for the audio buffer type (StereoBuffer16)

Subv:
The current code inserts and deletes elements from the beginning of the audio buffer, which is very inefficient in an std::vector.

Profiling was done using VisualStudio2017's Performance Analyzer in Super Mario 3D Land.

Before this change: AudioInterp::Linear had 14.14% of the runtime (inclusive) and most of that time was spent in std::vector's insert implementation.
After this change: AudioInterp::Linear has 0.36% of the runtime (inclusive)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2958)

### DIFF
--- a/src/audio_core/codec.cpp
+++ b/src/audio_core/codec.cpp
@@ -117,7 +117,9 @@ StereoBuffer16 DecodePCM16(const unsigned num_channels, const u8* const data,
             ret[i].fill(sample);
         }
     } else {
-        std::memcpy(ret.data(), data, sample_count * 2 * sizeof(u16));
+        for (size_t i = 0; i < sample_count; ++i) {
+            std::memcpy(&ret[i], data + i * sizeof(s16) * 2, 2 * sizeof(s16));
+        }
     }
 
     return ret;
--- a/src/audio_core/interpolate.cpp
+++ b/src/audio_core/interpolate.cpp
@@ -47,7 +47,7 @@ static void StepOverSamples(State& state, StereoBuffer16& input, float rate,
     state.xn1 = input[inputi + 1];
     state.fposition = fposition - inputi * scale_factor;
 
-    input.erase(input.begin(), input.begin() + inputi + 2);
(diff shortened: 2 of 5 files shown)


src/audio_core/interpolate.cpp
```diff
@@ -47,7 +47,9 @@ static void StepOverSamples(State& state, StereoBuffer16& input, float rate,
     state.xn1 = input[inputi + 1];
     state.fposition = fposition - inputi * scale_factor;
 
-    input.erase(input.begin(), input.begin() + inputi + 2);
+    auto end = input.begin();
+    std::advance(end, inputi + 2);
+    input.erase(input.begin(), end);
```
MerryMage: `input.erase(input.begin(), std::next(input.begin(), inputi + 2));`

src/audio_core/codec.cpp
```diff
@@ -117,7 +117,9 @@ StereoBuffer16 DecodePCM16(const unsigned num_channels, const u8* const data,
             ret[i].fill(sample);
         }
     } else {
-        std::memcpy(ret.data(), data, sample_count * 2 * sizeof(u16));
+        for (size_t i = 0; i < sample_count; ++i) {
+            memcpy(&ret[i], data + i * sizeof(s16) * 2, 2 * sizeof(s16));
```
MerryMage: `std::memcpy` matches rest of file, though we have an [annoying lack of consistency](https://github.com/citra-emu/citra/search?p=1&q=memcpy&type=&utf8=%E2%9C%93).

lioncash: There's also a relatively recent proposal to [deprecate](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0657r0.pdf) functions in the \<c*name*\> C++ headers from dumping their declarations into the global namespace.

PR #2961, opened 2017-09-26
Loaders: Don't automatically set the current process every time we load an application.

Subv:
The loaders will now just create a Kernel::Process, construct it and return it to the caller, which is responsible for setting it as the current process and configuring the global page table.

This behavior is needed for loading extra NCCHs while a game is already running (applets, system modules, etc).

This PR can be reviewed commit by commit.

Ultimately I want to make it so `current_page_table` is no longer needed, and all operations happen on the currently scheduled process by default, with overloads to operate on any arbitrary process:

```cpp
IsValidVirtualAddress(VAddr) = IsValidVirtualAddress(g_current_process, VAddr);
GetPointer(VAddr) = GetPointer(g_current_process, VAddr);
...
```
and so on

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2961)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -13,6 +13,7 @@
 #include "core/core_timing.h"
 #include "core/gdbstub/gdbstub.h"
 #include "core/hle/kernel/kernel.h"
+#include "core/hle/kernel/process.h"
 #include "core/hle/kernel/thread.h"
 #include "core/hle/service/service.h"
 #include "core/hw/hw.h"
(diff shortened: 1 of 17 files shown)


shinyquagsire23:
LGTM

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, 4 of 4 files at r4, 8 of 8 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2961#-:-Kv2U-vTlvRVFDVaKnM6:bb6ywmx)*

PR #2967, opened 2017-09-28
Kernel/Threads: When putting a thread to wait, specify a function to execute when it is awoken

Subv:
This change makes for a clearer (less confusing) path of execution in the scheduler, now the code to execute when a thread awakes is closer to the code that puts the thread to sleep (WaitSynch1, WaitSynchN). It also allows us to implement the special wake up behavior of ReplyAndReceive without hacking up WaitObject::WakeupAllWaitingThreads.

If savestates are desired in the future, we can change this implementation to one similar to the CoreTiming event system, where we first register the callback functions at startup and assign their identifiers to the Thread callback variable instead of directly assigning a lambda to the wake up callback variable.

I wanted to add [some tests](https://gist.github.com/Subv/7d83131e571f34ffcbf0ac8c4edc7182) for this new code but that's currently blocked by our inability to start and tear down the emulated system from within a test.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2967)

### DIFF
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -247,12 +247,15 @@ static void ThreadWakeupCallback(u64 thread_handle, int cycles_late) {
 
     if (thread->status == THREADSTATUS_WAIT_SYNCH_ANY ||
         thread->status == THREADSTATUS_WAIT_SYNCH_ALL || thread->status == THREADSTATUS_WAIT_ARB) {
-        thread->wait_set_output = false;
+
+        // Invoke the wakeup callback before clearing the wait objects
+        if (thread->wakeup_callback)
+            thread->wakeup_callback(ThreadWakeupReason::Timeout, thread, nullptr);
+
         // Remove the thread from each of its waiting objects' waitlists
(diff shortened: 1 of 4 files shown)


src/core/hle/kernel/thread.cpp
```diff
@@ -278,6 +281,9 @@ void Thread::ResumeFromWait() {
         break;
 
     case THREADSTATUS_READY:
+        // The thread's wakeup callback must have already been cleared when the thread was first
+        // awoken.
+        ASSERT(wakeup_callback == nullptr);
```
B3n30: ASSERT(wakeup_callback)

Subv: That's not the same thing, though, I want to make sure that wakeup_callback doesn't exist when this path is taken

mailwl:
nice! please implement this for y2r for Fifa games, when we told about broken semaphore implementation

Subv:
@mailwl I have no idea what you're talking about

mailwl:
Subv i told about FIFA 11 12 13 -this ported games dislike emulated Citra

PR #2968, opened 2017-09-28
Kernel/Threads: Add a new thread status that will allow using a Kernel::Event to put a guest thread to sleep inside an HLE handler until said event is signaled

Subv:
Some service functions should block the caller guest thread on SendSyncRequest until an async response can be calculated and returned (Most soc:U functions when dealing with blocking sockets, some UDS functions like ConnectToNetwork, most http:C functions, etc).

This change lets us put the threads to wait in the HLE handler and wake them up when the response comes back from another host thread, without having to neither spinlock (and freezing the entirety of the emulator thread) nor ignore the blocking nature of the function.

It is used like so:
```cpp
// In service function
auto future = PerformAsyncOperation();
auto token = SleepClientThread("soc::sendto", [future](SharedPtr<Thread> thread) {
    // Write response to thread's command buffer
    auto response = future.get();
});
...
// Somewhere else, when the async operation has finished
std::lock_guard<std::mutex> lock(HLE::g_hle_lock);
ContinueClientThread(token);
// Now the thread is awake and back into the scheduler queue.
```

This change, alongside #2967 will allow for much better control over the wait and wakeup behaviors of the guest threads.

# Concerns
* How should we handle the command buffer translation in HLERequestContext-based services? WriteToOutgoingCommandBuffer might be called with an incomplete response buffer.
* How should we write to the response buffer inside the callback? Tentative code:
```cpp
VAddr address = thread->GetCommandBufferAddress();
std::array<u8, IPC::COMMAND_BUFFER_LENGTH> buffer;
RequestBuilder rb(buffer.data(), params);
rb.Push(RESULT_SUCCESS);
Memory::WriteBlock(address, *thread->owner_process, buffer.data());
```
but we'd have to repeat that boilerplate in every callback.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2968)

### DIFF
--- a/src/citra_qt/debugger/wait_tree.cpp
+++ b/src/citra_qt/debugger/wait_tree.cpp
@@ -158,6 +158,9 @@ QString WaitTreeThread::GetText() const {
     case THREADSTATUS_WAIT_SYNCH_ANY:
         status = tr("waiting for objects");
         break;
+    case THREADSTATUS_WAIT_HLE_EVENT:
+        status = tr("waiting for HLE return");
+        break;
     case THREADSTATUS_DORMANT:
         status = tr("dormant");
         break;
@@ -184,6 +187,7 @@ QColor WaitTreeThread::GetColor() const {
         return QColor(Qt::GlobalColor::darkYellow);
     case THREADSTATUS_WAIT_SYNCH_ALL:
     case THREADSTATUS_WAIT_SYNCH_ANY:
+    case THREADSTATUS_WAIT_HLE_EVENT:
         return QColor(Qt::GlobalColor::red);
     case THREADSTATUS_DORMANT:
         return QColor(Qt::GlobalColor::darkCyan);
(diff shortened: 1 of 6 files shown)


wwylele:
I am assuming that the said HLE event is mimicking the behaviour of `svcRelyAndReceive` waking up the client thread. Then I have two questions:

 - does this work well when we introduce LLE sysmodule and implement the true `svcRelyAndReceive`?
 - iirc as we tested/theorized, `svcRelyAndReceive` has special preempting behaviour. Can the same behaviour be achieved by an event?

And this looks like a hack to me, which mixes itself into non-hack kernel stuff. So I suggest putting most of the PR description into the code comment, emphersizing how it is intended to be used

Subv:
I have updated the PR with a follow up commit that actually implements the interface functions i would like service implementers to use, and updated the PR description accordingly.

huwpascoe:
:lgtm:

---

Reviewed 4 of 4 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2968#-:-KvJYGLC0_Ov1R2Aq1tK:bnfp4nl)*

PR #2969, opened 2017-09-29
Service/PTM: Stub GetStepHistory function

mailwl:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2969)

### DIFF
--- a/src/core/hle/service/ptm/ptm.cpp
+++ b/src/core/hle/service/ptm/ptm.cpp
@@ -30,7 +30,7 @@ static bool battery_is_charging;
 
 static bool pedometer_is_counting;
 
-void GetAdapterState(Service::Interface* self) {
+void GetAdapterState(Interface* self) {
     IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x5, 0, 0);
 
     IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
@@ -40,7 +40,7 @@ void GetAdapterState(Service::Interface* self) {
     LOG_WARNING(Service_PTM, "(STUBBED) called");
 }
 
-void GetShellState(Service::Interface* self) {
+void GetShellState(Interface* self) {
     IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x6, 0, 0);
 
     IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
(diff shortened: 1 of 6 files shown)


B3n30:
I generally don't like it (and probably some devs agree), to just stub functions. So I would love to hear what the plans with that PR are? How should step counting be added? If there is no plan for it, I don't think this should be merged

mailwl:
@B3n30, if you look here: https://github.com/citra-emu/citra/blob/master/src/core/hle/service/ptm/ptm.cpp#L76
and here: https://github.com/citra-emu/citra/blob/master/src/core/hle/service/ptm/ptm.cpp#L130

you will see, emulated pedometer doesn't count steps. So, stubbed function works right

B3n30:
@mailwl That was not my question, if there is no way to count steps, stubbing that function is useless and is just dead code.

Thats why I asked if you plan to add a way to count steps. And if yes, how?

mailwl:
no. it's not dead code. If it function not stubbed, output data will contain random garbage.
If stubbed - always 0 steps for every period, like 3ds always at one place (btw, my real 3ds always with zero steps counts, i play at home)

B3n30:
Thats wrong. If this function isn't stubbed it will just return an error. So everything is fine. Implementing that function is only required if something with the step count is planned. Since it seems that you don't have any intend to add some sort of counting steps in any way, I'm against merging this PR

mailwl:
What??? Not stubbing functions alwas return success:
https://github.com/citra-emu/citra/blob/master/src/core/hle/service/service.cpp#L98

mailwl:
I don't have an idea, how to implement stp counting in right way atm. It should be consistent with other functions, like GetTotalStepCount or GetStepHistoryEntry, and Coin calculations. so, stubbing to zero looks fine for me

B3n30:
I think a way to start would be to provide a file, that has one number per line, each line represents the step count of that hour.

It's probably not the best way, but at least something useable. 
Even better would be an integration into the ui (but this might be as well to much for that feature).
Of course all functions you mentioned should be consistent.

DerProGamer2000:
or just make it a option or something where you can put in how many steps it should show like a cheat or something

mailwl:
@DerProGamer2000, very good point!  Citra is emulator, not cheat engine

acnleditor2:
nintendogs + cats hang on loading screen

wwylele:
@acnleditor2 could you be more specifically on how the game behaves differently on master and on this pr?

mailwl:
@wwylele just... ignore the troll

acnleditor2:
@wwylele 
master: crash
this pr: hang

mailwl:
достал бля.
what game crashed because of this PR? log please

Jhon591:
The toll found some think in mario maker at lest, must be in one of his commits so a mess, best to check the cmd line, what was in original citra and his..

acnleditor2:
@mailwl Nintendogs + Cats

PR #2970, opened 2017-09-29
[WIP] Implement Shader Scheduler

B3n30:
This splits the vertex rendering into the 3 VS Units. 
It also adds a common ThreadPool that can be used for the execution of multiple parallel instructions (like the 3 VS Units) .

Sorry for the ugly diff of command_processor.cpp. It's mainly just moving lot's of `case PICA_REG_INDEX(pipeline.trigger_draw_indexed)` into a lambda UnitLoop 

Shoutouts to Phanto-m for the initial design.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2970)

### DIFF
--- a/src/common/thread_pool.h
+++ b/src/common/thread_pool.h
@@ -1,0 +1,121 @@
+// Copyright 2016 Citra Emulator Project / PPSSPP Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <condition_variable>
+#include <deque>
+#include <functional>
+#include <future>
+#include <mutex>
+#include <thread>
+#include <vector>
+
+#include "common/assert.h"
+
+namespace Common {
+
+class ThreadPool {
+private:
+    explicit ThreadPool(unsigned int num_threads) : num_threads(num_threads), workers(num_threads) {
+        ASSERT(num_threads);
+    }
+
+public:
+    static ThreadPool& GetPool() {
+        static ThreadPool thread_pool(std::thread::hardware_concurrency());
+        return thread_pool;
+    }
+
+    template <typename F, typename... Args>
+    auto push(F&& f, Args&&... args) {
+        auto ret = workers[next_worker].push(std::forward<F>(f), std::forward<Args>(args)...);
+        next_worker = (next_worker + 1) % num_threads;
+        return ret;
+    }
+
+    unsigned int total_threads() {
+        return num_threads;
+    }
+
+private:
+    template <typename T>
+    class ThreadsafeQueue {
+    private:
+        const size_t capacity;
+        std::vector<T> queue_storage;
+        std::mutex mutex;
+        std::condition_variable queue_changed;
+
+    public:
+        explicit ThreadsafeQueue(const size_t capacity) : capacity(capacity) {
+            queue_storage.reserve(capacity);
+        }
+
+        void push(const T& element) {
+            std::unique_lock<std::mutex> lock(mutex);
+            while (queue_storage.size() >= capacity) {
+                queue_changed.wait(lock);
+            }
+            queue_storage.push_back(element);
+            queue_changed.notify_one();
+        }
+
+        T pop() {
+            std::unique_lock<std::mutex> lock(mutex);
+            while (queue_storage.empty()) {
+                queue_changed.wait(lock);
+            }
+            T element(std::move(queue_storage.back()));
+            queue_storage.pop_back();
+            queue_changed.notify_one();
+            return element;
+        }
+
+        void push(T&& element) {
+            std::unique_lock<std::mutex> lock(mutex);
+            while (queue_storage.size() >= capacity) {
+                queue_changed.wait(lock);
+            }
+            queue_storage.emplace_back(element);
+            queue_changed.notify_one();
+        }
+    };
+
+    class Worker {
+    private:
+        ThreadsafeQueue<std::function<void()>> queue;
+        std::thread thread;
+        static constexpr size_t MAX_QUEUE_CAPACITY = 100;
+
+    public:
+        Worker() : queue(MAX_QUEUE_CAPACITY), thread([this] { loop(); }) {}
+
+        ~Worker() {
+            queue.push(nullptr); // Exit the loop
+            thread.join();
+        }
+
+        void loop() {
+            for (;;) {
+                std::function<void()> fn(queue.pop());
+                if (!fn) // a nullptr function is the signal to exit the loop
+                    break;
+                fn();
+            }
+        }
+
+        template <typename F, typename... Args>
+        auto push(F&& f, Args&&... args) {
+            auto task = std::make_shared<std::packaged_task<decltype(f(args...))()>>(
+                std::bind(std::forward<F>(f), std::forward<Args>(args)...));
+            queue.push([task]() { (*task)(); });
+            return task->get_future();
+        }
+    };
+
+    const unsigned int num_threads;
+    int next_worker = 0;
+    std::vector<Worker> workers;
+};
+
+} // namespace Common
--- a/src/common/thread_pool.h
+++ b/src/common/thread_pool.h
@@ -1,9 +1,8 @@
-// Copyright 2016 Citra Emulator Project / PPSSPP Project
+// Copyright 2017 Citra Emulator Project
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
 #include <condition_variable>
-#include <deque>
 #include <functional>
 #include <future>
 #include <mutex>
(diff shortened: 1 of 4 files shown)


src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+public:
+    static ThreadPool& GetPool() {
+        static ThreadPool thread_pool(std::thread::hardware_concurrency());
+        return thread_pool;
+    }
+
+    template <typename F, typename... Args>
+    auto push(F&& f, Args&&... args) {
```
danzel: Should we document that this uses a LIFO queue internally?
I wasn't expecting it to, queue usually means FIFO.

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
+// Copyright 2016 Citra Emulator Project / PPSSPP Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <condition_variable>
+#include <deque>
```
danzel: unused?

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+    public:
+        Worker() : queue(MAX_QUEUE_CAPACITY), thread([this] { loop(); }) {}
+
+        ~Worker() {
+            queue.push(nullptr); // Exit the loop
+            thread.join();
+        }
+
+        void loop() {
+            for (;;) {
```
lioncash: `while (true)`

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+                    break;
+                fn();
+            }
+        }
+
+        template <typename F, typename... Args>
+        auto push(F&& f, Args&&... args) {
+            auto task = std::make_shared<std::packaged_task<decltype(f(args...))()>>(
+                std::bind(std::forward<F>(f), std::forward<Args>(args)...));
+            queue.push([task]() { (*task)(); });
```
lioncash: The first set of `()` can be omitted like the other lambda

src/video_core/command_processor.cpp
```diff
@@ -298,6 +299,39 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
         const u16* index_address_16 = reinterpret_cast<const u16*>(index_address_8);
         bool index_u16 = index_info.format != 0;
 
+        struct CacheEntry {
+            Shader::AttributeBuffer output_attr;
+            Shader::OutputVertex output_vertex;
+            std::atomic<u32> id;
+            std::atomic_flag writing{
```
lioncash: Maybe place the comment above this variable so it doesn't need to wrap like this?

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+#include <thread>
+#include <vector>
+
+#include "common/assert.h"
+
+namespace Common {
+
+class ThreadPool {
+private:
+    explicit ThreadPool(unsigned int num_threads) : num_threads(num_threads), workers(num_threads) {
```
lioncash: `size_t`

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    template <typename F, typename... Args>
+    auto push(F&& f, Args&&... args) {
+        auto ret = workers[next_worker].push(std::forward<F>(f), std::forward<Args>(args)...);
+        next_worker = (next_worker + 1) % num_threads;
+        return ret;
+    }
+
+    unsigned int total_threads() {
```
lioncash: This can be a const member function.

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+        template <typename F, typename... Args>
+        auto push(F&& f, Args&&... args) {
+            auto task = std::make_shared<std::packaged_task<decltype(f(args...))()>>(
+                std::bind(std::forward<F>(f), std::forward<Args>(args)...));
+            queue.push([task]() { (*task)(); });
+            return task->get_future();
+        }
+    };
+
+    const unsigned int num_threads;
```
lioncash: `size_t`

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+            queue_changed.notify_one();
+            return element;
+        }
+
+        void push(T&& element) {
+            std::unique_lock<std::mutex> lock(mutex);
+            while (queue_storage.size() >= capacity) {
+                queue_changed.wait(lock);
+            }
+            queue_storage.emplace_back(element);
```
lioncash: This wont perform a move, since `element` is being used as an lvalue, not an xvalue.

```cpp
queue_storage.push_back(std::move(element));
```

is likely what you want instead.

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+        std::vector<T> queue_storage;
+        std::mutex mutex;
+        std::condition_variable queue_changed;
+
+    public:
+        explicit ThreadsafeQueue(const size_t capacity) : capacity(capacity) {
+            queue_storage.reserve(capacity);
+        }
+
+        void push(const T& element) {
```
lioncash: The member function names should be cased the same as the type name. Ditto elsewhere.

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+        auto push(F&& f, Args&&... args) {
+            auto task = std::make_shared<std::packaged_task<decltype(f(args...))()>>(
+                std::bind(std::forward<F>(f), std::forward<Args>(args)...));
+            queue.push([task]() { (*task)(); });
+            return task->get_future();
+        }
+    };
+
+    const unsigned int num_threads;
+    int next_worker = 0;
```
lioncash: `size_t`

src/video_core/command_processor.cpp
```diff
@@ -298,6 +299,39 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+                ATOMIC_FLAG_INIT}; // Set when a thread is writing into this entry
+        };
+        static std::array<CacheEntry, 0x10000> cache;
+
+        // used as a mean to invalidate data from the previous batch without clearing it
+        static u32 cache_batch_id = std::numeric_limits<u32>::max();
+
+        ++cache_batch_id;
+        if (cache_batch_id == 0) { // reset cache if the emu ever runs long enough to overflow id
```
lioncash: Prefer placing comments like these above rather than beside it.

src/video_core/command_processor.cpp
```diff
@@ -298,6 +299,39 @@ static void WritePicaReg(u32 id, u32 value, u32 mask) {
...
+        struct VsOutput {
+            explicit VsOutput() = default;
+            VsOutput(VsOutput&& other) {
+                batch_id = 0;
+            }
+
+            Pica::Shader::OutputVertex vertex;
+            std::atomic<u32> batch_id;
+        };
+        static std::vector<VsOutput> vs_output;
```
lioncash: Should these *really* live directly in the function instead of being extracting them into a class or turning the command processor into one? This seems kind of like a design smell (and a site for data races).

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
+// Copyright 2016 Citra Emulator Project / PPSSPP Project
```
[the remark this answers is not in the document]
B3n30: Yeah I made a major rework so we could probably remove the `PPSSPP Project`

src/common/thread_pool.h
```diff
@@ -0,0 +1,121 @@
...
+#include <future>
+#include <mutex>
+#include <thread>
+#include <vector>
+
+#include "common/assert.h"
+
+namespace Common {
+
+class ThreadPool {
```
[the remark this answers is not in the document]
B3n30: Since this has only 100 lines of code, I don't think it is necessary to split it into .h and .cpp but if it is common sense to have pure header files, I'll split it

tabnk:
Tested Mario Kart and Super Mario 3D. About 8 to 12 fps improvement in most case.

acnleditor2:
Breaks Pokémon ORAS animations

Jhon591:
Speed is good, I'm no programmer but any change to fix wwylele GS++ up with your branch ?, and rebase the conflict issue.

Thanks, and just to say very nice work hope you keep it up.

Jhon591:
why close ?, put line 330 back and it will prevent crashing upon pokemon xy startup in fmv gs++

RavenHome1:
@Jhon591 
https://community.citra-emu.org/t/progress-report-for-september-october/4153/5?u=ravenhome

Jhon591:
@RavenHome1

thanks, yes best not to use until fully completed.

PR #2973, opened 2017-09-30
Moved down_count to CoreTiming

huwpascoe:
It's needed for testing.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2973)

### DIFF
--- a/src/core/core_timing.cpp
+++ b/src/core/core_timing.cpp
@@ -57,6 +57,9 @@ static s64 idled_cycles;
 static s64 last_global_time_ticks;
 static s64 last_global_time_us;
 
+static s64 down_count = 0; ///< A decreasing counter of remaining cycles before the next event,
+                           /// decreased by the cpu run loop
+
 static std::recursive_mutex external_event_section;
 
 // Warning: not included in save state.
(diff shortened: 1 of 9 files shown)


src/tests/core/arm/dyncom/arm_dyncom_vfp_tests.cpp
```diff
@@ -29,7 +30,7 @@ TEST_CASE("ARM_DynCom (vfp): vadd", "[arm_dyncom]") {
     }};
 
     for (const auto& test_case : test_cases) {
-        dyncom.down_count = 1000; // Ensure that CoreTimeing will not be called.
+        CoreTiming::SetDownCount(1000); // Ensure that CoreTimeing will not be called.
```
Subv: Remove this call, it's no longer needed

src/core/core_timing.cpp
```diff
@@ -185,8 +188,19 @@ void Shutdown() {
     }
 }
 
+void SetDownCount(u64 value) {
```
Subv: This function is not needed

src/tests/core/arm/dyncom/arm_dyncom_vfp_tests.cpp
huwpascoe: Good

huwpascoe:
amended

tywald:
What's been said on Discord should probably be copy-pasted in the description so the information is tied here.

huwpascoe:
Um, all that was said was it's needed for testing. Should I add that?

tywald:
Better than "as mentioned in discord" if you will look back at this PR one year from now.

MerryMage:
Note for posterity: Decouples CoreTiming from CPU emulation, making it easier for unit tests to be written.

PR #2975, opened 2017-09-30
file_sys/archive_ncch: use NCCHs/.apps instead of .romfs files, NCCH section override

### LINKED ISSUE #2735: Allow RomFS/ExeFS patch files to be applied to games
I'm currently refactoring/rewriting the `pk3DS` Pokemon randomizer, which operates on an extracted 3DS ROM filesystem, with the typical format of:

```
ExeFS
--code.bin
RomFS
--<rom FS files>
```

The latest versions of Luma allow live-patching running ROMs with a structure identical to this, meaning I can test randomizations or ROM edits without having to wait for an entire ROM repack operation to complete (which takes a while). However, it's not easy to constantly swap test saves on and off of a game cartridge, and it's obnoxious having to constantly move my SD card between my 3DS and computer.

It would be nice if Citra had a feature similar to Luma's, where a directory could be specified for a specific title to patch individual RomFS files, or the ExeFS `code.bin`.

I'm not familiar with the inner-workings of Citra specifically, so I'm not sure if this is possible without major modifications, but I thought I'd bring it up in case it's something that could be done.

shinyquagsire23:
- archive_ncch now uses NCCHContainer instead of loading a .romfs
- Backward compatibility with existing is maintained through RomFS override, ExeFS override has also been added which fixes #2735 
- A class for reading from and writing to TMDs has been added, which paves the way for title install via CIAs
- archive_ncch will use title metadata to determine the content ID for the main executable. This allows for decrypted NAND contents to be used without altering the file structure.
- The NCCH loader will also use title metadata to determine the content ID for update NCCHs, given 

RomFS and ExeFS override is per-NCCH. For example, if I have a update of Smash installed as `sdmc:/Nintendo 3DS/00000000000000000000000000000000/00000000000000000000000000000000/title/0004000e/000ee000/content/00000017.app`, the RomFS override would be `sdmc:/Nintendo 3DS/00000000000000000000000000000000/00000000000000000000000000000000/title/0004000e/000ee000/content/00000017.app.romfs`. If I wanted to override an exeFS file (ie .code), I can place the file at `sdmc:/Nintendo 3DS/00000000000000000000000000000000/00000000000000000000000000000000/title/0004000e/000ee000/content/00000017.app.exefsdir/.code`. (For Windows-based systems, .code can be overridden as code.bin instead).

If a TMD is not found for an application, the path will default to `.../00000000.app`, and if `.../00000000.app` is not found but `.../00000000.app.romfs` is, the RomFS will stand-in as an NCCH even though the full contents of the NCCH are not available. For system archives, this generally works fine and archive_ncch will still work with older .romfs-based dumps.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2975)

### DIFF
--- a/src/core/file_sys/ncch_container.cpp
+++ b/src/core/file_sys/ncch_container.cpp
@@ -116,92 +116,143 @@ Loader::ResultStatus NCCHContainer::Load() {
     if (is_loaded)
         return Loader::ResultStatus::Success;
 
-    // Reset read pointer in case this file has been read before.
-    file.Seek(0, SEEK_SET);
+    if (file.IsOpen()) {
+        // Reset read pointer in case this file has been read before.
+        file.Seek(0, SEEK_SET);
 
-    if (file.ReadBytes(&ncch_header, sizeof(NCCH_Header)) != sizeof(NCCH_Header))
-        return Loader::ResultStatus::Error;
-
-    // Skip NCSD header and load first NCCH (NCSD is just a container of NCCH files)...
-    if (Loader::MakeMagic('N', 'C', 'S', 'D') == ncch_header.magic) {
-        LOG_DEBUG(Service_FS, "Only loading the first (bootable) NCCH within the NCSD file!");
-        ncch_offset = 0x4000;
-        file.Seek(ncch_offset, SEEK_SET);
-        file.ReadBytes(&ncch_header, sizeof(NCCH_Header));
-    }
-
-    // Verify we are loading the correct file type...
-    if (Loader::MakeMagic('N', 'C', 'C', 'H') != ncch_header.magic)
-        return Loader::ResultStatus::ErrorInvalidFormat;
-
-    // System archives and DLC don't have an extended header but have RomFS
-    if (ncch_header.extended_header_size) {
-        if (file.ReadBytes(&exheader_header, sizeof(ExHeader_Header)) != sizeof(ExHeader_Header))
+        if (file.ReadBytes(&ncch_header, sizeof(NCCH_Header)) != sizeof(NCCH_Header))
             return Loader::ResultStatus::Error;
 
-        is_compressed = (exheader_header.codeset_info.flags.flag & 1) == 1;
-        u32 entry_point = exheader_header.codeset_info.text.address;
-        u32 code_size = exheader_header.codeset_info.text.code_size;
-        u32 stack_size = exheader_header.codeset_info.stack_size;
-        u32 bss_size = exheader_header.codeset_info.bss_size;
-        u32 core_version = exheader_header.arm11_system_local_caps.core_version;
-        u8 priority = exheader_header.arm11_system_local_caps.priority;
-        u8 resource_limit_category =
-            exheader_header.arm11_system_local_caps.resource_limit_category;
-
-        LOG_DEBUG(Service_FS, "Name:                        %s", exheader_header.codeset_info.name);
-        LOG_DEBUG(Service_FS, "Program ID:                  %016" PRIX64, ncch_header.program_id);
-        LOG_DEBUG(Service_FS, "Code compressed:             %s", is_compressed ? "yes" : "no");
-        LOG_DEBUG(Service_FS, "Entry point:                 0x%08X", entry_point);
-        LOG_DEBUG(Service_FS, "Code size:                   0x%08X", code_size);
-        LOG_DEBUG(Service_FS, "Stack size:                  0x%08X", stack_size);
-        LOG_DEBUG(Service_FS, "Bss size:                    0x%08X", bss_size);
-        LOG_DEBUG(Service_FS, "Core version:                %d", core_version);
-        LOG_DEBUG(Service_FS, "Thread priority:             0x%X", priority);
-        LOG_DEBUG(Service_FS, "Resource limit category:     %d", resource_limit_category);
-        LOG_DEBUG(Service_FS, "System Mode:                 %d",
-                  static_cast<int>(exheader_header.arm11_system_local_caps.system_mode));
-
-        if (exheader_header.system_info.jump_id != ncch_header.program_id) {
-            LOG_ERROR(Service_FS, "ExHeader Program ID mismatch: the ROM is probably encrypted.");
-            return Loader::ResultStatus::ErrorEncrypted;
+        // Skip NCSD header and load first NCCH (NCSD is just a container of NCCH files)...
+        if (Loader::MakeMagic('N', 'C', 'S', 'D') == ncch_header.magic) {
+            LOG_DEBUG(Service_FS, "Only loading the first (bootable) NCCH within the NCSD file!");
+            ncch_offset = 0x4000;
+            file.Seek(ncch_offset, SEEK_SET);
+            file.ReadBytes(&ncch_header, sizeof(NCCH_Header));
         }
 
-        has_exheader = true;
+        // Verify we are loading the correct file type...
+        if (Loader::MakeMagic('N', 'C', 'C', 'H') != ncch_header.magic)
+            return Loader::ResultStatus::ErrorInvalidFormat;
+
+        has_header = true;
+
+        // System archives and DLC don't have an extended header but have RomFS
+        if (ncch_header.extended_header_size) {
+            if (file.ReadBytes(&exheader_header, sizeof(ExHeader_Header)) !=
+                sizeof(ExHeader_Header))
+                return Loader::ResultStatus::Error;
+
+            is_compressed = (exheader_header.codeset_info.flags.flag & 1) == 1;
+            u32 entry_point = exheader_header.codeset_info.text.address;
+            u32 code_size = exheader_header.codeset_info.text.code_size;
+            u32 stack_size = exheader_header.codeset_info.stack_size;
+            u32 bss_size = exheader_header.codeset_info.bss_size;
+            u32 core_version = exheader_header.arm11_system_local_caps.core_version;
+            u8 priority = exheader_header.arm11_system_local_caps.priority;
+            u8 resource_limit_category =
+                exheader_header.arm11_system_local_caps.resource_limit_category;
+
+            LOG_DEBUG(Service_FS, "Name:                        %s",
+                      exheader_header.codeset_info.name);
+            LOG_DEBUG(Service_FS, "Program ID:                  %016" PRIX64,
+                      ncch_header.program_id);
+            LOG_DEBUG(Service_FS, "Code compressed:             %s", is_compressed ? "yes" : "no");
+            LOG_DEBUG(Service_FS, "Entry point:                 0x%08X", entry_point);
+            LOG_DEBUG(Service_FS, "Code size:                   0x%08X", code_size);
+            LOG_DEBUG(Service_FS, "Stack size:                  0x%08X", stack_size);
+            LOG_DEBUG(Service_FS, "Bss size:                    0x%08X", bss_size);
+            LOG_DEBUG(Service_FS, "Core version:                %d", core_version);
+            LOG_DEBUG(Service_FS, "Thread priority:             0x%X", priority);
+            LOG_DEBUG(Service_FS, "Resource limit category:     %d", resource_limit_category);
+            LOG_DEBUG(Service_FS, "System Mode:                 %d",
+                      static_cast<int>(exheader_header.arm11_system_local_caps.system_mode));
+
+            if (exheader_header.system_info.jump_id != ncch_header.program_id) {
+                LOG_ERROR(Service_FS,
+                          "ExHeader Program ID mismatch: the ROM is probably encrypted.");
+                return Loader::ResultStatus::ErrorEncrypted;
+            }
+
+            has_exheader = true;
+        }
+
+        // DLC can have an ExeFS and a RomFS but no extended header
+        if (ncch_header.exefs_size) {
+            exefs_offset = ncch_header.exefs_offset * kBlockSize;
+            u32 exefs_size = ncch_header.exefs_size * kBlockSize;
+
+            LOG_DEBUG(Service_FS, "ExeFS offset:                0x%08X", exefs_offset);
+            LOG_DEBUG(Service_FS, "ExeFS size:                  0x%08X", exefs_size);
+
+            file.Seek(exefs_offset + ncch_offset, SEEK_SET);
+            if (file.ReadBytes(&exefs_header, sizeof(ExeFs_Header)) != sizeof(ExeFs_Header))
+                return Loader::ResultStatus::Error;
+
+            exefs_file = FileUtil::IOFile(filepath, "rb");
+            has_exefs = true;
+        }
+
+        if (ncch_header.romfs_offset != 0 && ncch_header.romfs_size != 0)
+            has_romfs = true;
     }
 
-    // DLC can have an ExeFS and a RomFS but no extended header
-    if (ncch_header.exefs_size) {
-        exefs_offset = ncch_header.exefs_offset * kBlockSize;
-        u32 exefs_size = ncch_header.exefs_size * kBlockSize;
+    LoadOverrides();
 
-        LOG_DEBUG(Service_FS, "ExeFS offset:                0x%08X", exefs_offset);
-        LOG_DEBUG(Service_FS, "ExeFS size:                  0x%08X", exefs_size);
-
-        file.Seek(exefs_offset + ncch_offset, SEEK_SET);
-        if (file.ReadBytes(&exefs_header, sizeof(ExeFs_Header)) != sizeof(ExeFs_Header))
-            return Loader::ResultStatus::Error;
-
-        has_exefs = true;
-    }
-
-    if (ncch_header.romfs_offset != 0 && ncch_header.romfs_size != 0)
-        has_romfs = true;
+    // We need at least one of these or overrides, practically
+    if (!(has_exefs || has_romfs || is_tainted))
+        return Loader::ResultStatus::Error;
 
     is_loaded = true;
     return Loader::ResultStatus::Success;
 }
 
-Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vector<u8>& buffer) {
-    if (!file.IsOpen())
-        return Loader::ResultStatus::Error;
+Loader::ResultStatus NCCHContainer::LoadOverrides() {
+    // Check for split-off files, mark the archive as tainted if we will use them
+    std::string romfs_override = filepath + ".romfs";
+    if (FileUtil::Exists(romfs_override)) {
+        is_tainted = true;
+    }
 
+    // If we have a split-off exefs file/folder, it takes priority
+    std::string exefs_override = filepath + ".exefs";
+    std::string exefsdir_override = filepath + ".exefsdir/";
+    if (FileUtil::Exists(exefs_override)) {
+        exefs_file = FileUtil::IOFile(exefs_override, "rb");
+
+        if (exefs_file.ReadBytes(&exefs_header, sizeof(ExeFs_Header)) == sizeof(ExeFs_Header)) {
+            LOG_DEBUG(Service_FS, "Loading ExeFS section from %s", exefs_override.c_str());
+            exefs_offset = 0;
+            is_tainted = true;
+            has_exefs = true;
+        } else {
+            exefs_file = FileUtil::IOFile(filepath, "rb");
+        }
+    } else if (FileUtil::Exists(exefsdir_override) && FileUtil::IsDirectory(exefsdir_override)) {
+        is_tainted = true;
+    }
+
+    if (is_tainted)
+        LOG_WARNING(Service_FS,
+                    "Loaded NCCH %s is tainted, application behavior may not be as expected!",
+                    filepath.c_str());
+
+    return Loader::ResultStatus::Success;
+}
+
+Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vector<u8>& buffer) {
     Loader::ResultStatus result = Load();
     if (result != Loader::ResultStatus::Success)
         return result;
 
-    if (!has_exefs)
-        return Loader::ResultStatus::ErrorNotUsed;
+    // Check if we have files that can drop-in and replace
+    result = LoadOverrideExeFSSection(name, buffer);
+    if (result == Loader::ResultStatus::Success || !has_exefs)
+        return result;
+
+    // If we don't have any separate files, we'll need a full ExeFS
+    if (!exefs_file.IsOpen())
+        return Loader::ResultStatus::Error;
 
     LOG_DEBUG(Service_FS, "%d sections:", kMaxSections);
     // Iterate through the ExeFs archive until we find a section with the specified name...
@@ -215,7 +266,7 @@ Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vect
 
             s64 section_offset =
                 (section.offset + exefs_offset + sizeof(ExeFs_Header) + ncch_offset);
-            file.Seek(section_offset, SEEK_SET);
+            exefs_file.Seek(section_offset, SEEK_SET);
 
             if (strcmp(section.name, ".code") == 0 && is_compressed) {
                 // Section is compressed, read compressed .code section...
@@ -226,7 +277,7 @@ Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vect
                     return Loader::ResultStatus::ErrorMemoryAllocationFailed;
                 }
 
-                if (file.ReadBytes(&temp_buffer[0], section.size) != section.size)
+                if (exefs_file.ReadBytes(&temp_buffer[0], section.size) != section.size)
                     return Loader::ResultStatus::Error;
 
                 // Decompress .code section...
@@ -237,7 +288,7 @@ Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vect
             } else {
                 // Section is uncompressed...
                 buffer.resize(section.size);
-                if (file.ReadBytes(&buffer[0], section.size) != section.size)
+                if (exefs_file.ReadBytes(&buffer[0], section.size) != section.size)
                     return Loader::ResultStatus::Error;
             }
             return Loader::ResultStatus::Success;
@@ -246,19 +297,55 @@ Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vect
     return Loader::ResultStatus::ErrorNotUsed;
 }
 
-Loader::ResultStatus NCCHContainer::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_file,
-                                              u64& offset, u64& size) {
-    if (!file.IsOpen())
+Loader::ResultStatus NCCHContainer::LoadOverrideExeFSSection(const char* name,
+                                                             std::vector<u8>& buffer) {
+    std::string override_name;
+
+    // Map our section name to the extracted equivalent
+    if (!strcmp(name, ".code"))
+        override_name = "code.bin";
+    else if (!strcmp(name, "icon"))
+        override_name = "code.bin";
+    else if (!strcmp(name, "banner"))
+        override_name = "banner.bnr";
+    else if (!strcmp(name, "logo"))
+        override_name = "logo.bcma.lz";
+    else
         return Loader::ResultStatus::Error;
 
+    std::string section_override = filepath + ".exefsdir/" + override_name;
+    FileUtil::IOFile section_file(section_override, "rb");
+
+    if (section_file.IsOpen()) {
+        auto section_size = section_file.GetSize();
+        buffer.resize(section_size);
+
+        section_file.Seek(0, SEEK_SET);
+        if (section_file.ReadBytes(&buffer[0], section_size) == section_size) {
+            LOG_WARNING(Service_FS, "File %s overriding built-in ExeFS file",
+                        section_override.c_str());
+            return Loader::ResultStatus::Success;
+        }
+    }
+    return Loader::ResultStatus::ErrorNotUsed;
+}
+
+Loader::ResultStatus NCCHContainer::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_file,
+                                              u64& offset, u64& size) {
     Loader::ResultStatus result = Load();
     if (result != Loader::ResultStatus::Success)
         return result;
+
+    if (ReadOverrideRomFS(romfs_file, offset, size) == Loader::ResultStatus::Success)
+        return Loader::ResultStatus::Success;
 
     if (!has_romfs) {
         LOG_DEBUG(Service_FS, "RomFS requested from NCCH which has no RomFS");
         return Loader::ResultStatus::ErrorNotUsed;
     }
+
+    if (!file.IsOpen())
+        return Loader::ResultStatus::Error;
 
     u32 romfs_offset = ncch_offset + (ncch_header.romfs_offset * kBlockSize) + 0x1000;
     u32 romfs_size = (ncch_header.romfs_size * kBlockSize) - 0x1000;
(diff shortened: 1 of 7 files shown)


src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+    int GetContentCount();
+    u32 GetBootContentID();
+    u32 GetManualContentID();
+    u32 GetDLPContentID();
+
+    void SetTitleID(u64 title_id);
+    void SetTitleType(u32 type);
+    void SetTitleVersion(u16 version);
+    void SetSystemVersion(u64 version);
+    void AddContentChunk(TMD_ContentChunk& chunk);
```
lioncash:
```cpp
void AddContentChunk(const TMD_ContentChunk& chunk);
```

src/core/file_sys/ncch_container.cpp
```diff
@@ -116,93 +116,154 @@ Loader::ResultStatus NCCHContainer::Load() {
...
 
-        has_exefs = true;
+        if (ncch_header.romfs_offset != 0 && ncch_header.romfs_size != 0)
+            has_romfs = true;
     }
 
-    if (ncch_header.romfs_offset != 0 && ncch_header.romfs_size != 0)
-        has_romfs = true;
+    // Check for split-off files, mark the archive as tainted if we will use them
+    bool has_split_romfs = false;
```
Subv: IMO this entire section could be factored into a function, something like `LoadOverrides` but with a better name.

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+ * If a file path is provided and the file exists, it will be parsed and used, otherwise
+ * it will be created. The TMD file can then be interpreted, modified and/or saved.
+ */
+class TitleMetadata {
+public:
+    TitleMetadata(std::string& path) : filepath(path) {}
+    Loader::ResultStatus Load();
+    Loader::ResultStatus Save();
+
+    u64 GetTitleID();
```
lioncash:
```cpp
u64 GetTitleID() const;
```

etc for all the other getters

src/core/file_sys/ncch_container.cpp
```diff
@@ -116,93 +116,154 @@ Loader::ResultStatus NCCHContainer::Load() {
...
 Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vector<u8>& buffer) {
-    if (!file.IsOpen())
-        return Loader::ResultStatus::Error;
-
     Loader::ResultStatus result = Load();
     if (result != Loader::ResultStatus::Success)
         return result;
 
-    if (!has_exefs)
+    // Check if we have files that can drop-in and replace
```
Subv: This too could be made into its own function, `LoadOverrideExeFSSection`

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+u16 TitleMetadata::GetTitleVersion() {
+    return static_cast<u16>(tmd_body.title_version);
+}
+
+u64 TitleMetadata::GetSystemVersion() {
+    return static_cast<u64>(tmd_body.system_version);
+}
+
+int TitleMetadata::GetContentCount() {
```
lioncash: This should return a `size_t`, since this truncates otherwise.

src/core/file_sys/ncch_container.cpp
```diff
@@ -248,18 +309,30 @@ Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vect
...
 Loader::ResultStatus NCCHContainer::ReadRomFS(std::shared_ptr<FileUtil::IOFile>& romfs_file,
                                               u64& offset, u64& size) {
-    if (!file.IsOpen())
-        return Loader::ResultStatus::Error;
-
     Loader::ResultStatus result = Load();
     if (result != Loader::ResultStatus::Success)
         return result;
 
+    // Check for RomFS overrides
```
Subv: Same

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// FileSys namespace
+
+namespace FileSys {
+
+static const int TMD_INDEX_MAIN = 0;
+static const int TMD_INDEX_MANUAL = 1;
+static const int TMD_INDEX_DLP = 2;
+
+static u32 GetSignatureSize(u32_be signature_type) {
```
Subv: The parameter can be a normal u32

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    // Signature lengths are variable, and the body follows the signature
+    u32 signature_size = GetSignatureSize(signature_type);
+
+    tmd_signature.resize(signature_size);
+    if (!file.ReadBytes(&tmd_signature[0], signature_size))
+        return Loader::ResultStatus::Error;
+
+    // The TMD body start position is rounded to the nearest 0x40 after the signature
+    u32 body_start = (signature_size + sizeof(u32_be) + 0x40) & ~0x3F;
```
Subv: We have AlignUp/AlignDown functions in the Common namespace

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+    LOG_WARNING(Service_FS, "Seeking to %x", body_start);
```
Subv: Why is this a warning?

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+    file.Seek(body_start, SEEK_SET);
+
+    // Read our TMD body, then load the amount of ContentChunks specified
+    if (file.ReadBytes(&tmd_body, sizeof(TMD_Body)) != sizeof(TMD_Body))
+        return Loader::ResultStatus::Error;
+
+    for (int i = 0; i < tmd_body.content_count; i++) {
```
Subv: int -> unsigned int

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+        TMD_ContentChunk chunk;
+        if (file.ReadBytes(&chunk, sizeof(TMD_ContentChunk)) == sizeof(TMD_ContentChunk)) {
+            tmd_chunks.push_back(chunk);
+        }
```
Subv: If the read fails for whatever reason, this will just silently not add the chunk to the list, shouldn't it report an error about the malformed file?

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (!file.IsOpen())
+        return Loader::ResultStatus::Error;
+
+    if (!file.WriteBytes(&signature_type, sizeof(u32_be)))
+        return Loader::ResultStatus::Error;
+
+    // Signature lengths are variable, and the body follows the signature
+    u32 signature_size = GetSignatureSize(signature_type);
+
+    if (!file.WriteBytes(&tmd_signature[0], signature_size))
```
Subv: tmd_signature.data()

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+        return Loader::ResultStatus::Error;
+
+    // The TMD body start position is rounded to the nearest 0x40 after the signature
+    u32 body_start = (signature_size + sizeof(u32_be) + 0x40) & ~0x3F;
```
Subv: Common::Align[Up|Down]

src/core/file_sys/title_metadata.cpp
lioncash: This should be a `u16` or `u32`, otherwise implicit sign conversion is being done.

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+    // Update our TMD body values and hashes
+    tmd_body.content_count = tmd_chunks.size();
+
+    // TODO(shinyquagsire23): Do TMDs with more than one contentinfo exist?
+    tmd_body.contentinfo[0].index = 0;
+    tmd_body.contentinfo[0].command_count = tmd_chunks.size();
+    tmd_body.contentinfo[1].command_count = 0;
+
+    CryptoPP::SHA256 chunk_hash;
+    for (int i = 0; i < tmd_body.content_count; i++) {
```
lioncash: ditto wrt implicit sign conversion (size_t is more appropriate here)

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    // The TMD body start position is rounded to the nearest 0x40 after the signature
+    u32 body_start = (signature_size + sizeof(u32_be) + 0x40) & ~0x3F;
+    file.Seek(body_start, SEEK_SET);
+
+    // Update our TMD body values and hashes
+    tmd_body.content_count = tmd_chunks.size();
+
+    // TODO(shinyquagsire23): Do TMDs with more than one contentinfo exist?
+    tmd_body.contentinfo[0].index = 0;
```
Subv: What is this code doing? please add a more descriptive comment

src/core/file_sys/title_metadata.cpp
Subv: unsigned, same for all the other loop variables. This is so we can avoid signed/unsigned comparison warnings

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+        u8* contentinfo_data = static_cast<u8*>(static_cast<void*>(&tmd_body.contentinfo[i]));
+        chunk_hash.Update(contentinfo_data, sizeof(TMD_ContentInfo));
+    }
+    chunk_hash.Final(tmd_body.contentinfo_hash);
+
+    // Write our TMD body, then write each of our ContentChunks
+    if (file.WriteBytes(&tmd_body, sizeof(TMD_Body)) != sizeof(TMD_Body))
+        return Loader::ResultStatus::Error;
+
+    for (int i = 0; i < tmd_body.content_count; i++) {
```
lioncash: ditto wrt sign conversion (size_t is more appropriate here since memory is being indexed).

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    // TODO(shinyquagsire23): Do TMDs with more than one contentinfo exist?
+    tmd_body.contentinfo[0].index = 0;
+    tmd_body.contentinfo[0].command_count = tmd_chunks.size();
+    tmd_body.contentinfo[1].command_count = 0;
+
+    CryptoPP::SHA256 chunk_hash;
+    for (int i = 0; i < tmd_body.content_count; i++) {
+        u8* chunk_data = static_cast<u8*>(static_cast<void*>(&tmd_chunks[i]));
+        chunk_hash.Update(chunk_data, sizeof(TMD_ContentChunk));
```
Subv: you can use reinterpret_cast<u8*>(&tmd_chunks[i]) directly here.

This is imo a disadvantage of the CryptoPP API, but 🤷‍♂️

src/core/loader/ncch.cpp
```diff
@@ -46,12 +47,22 @@ FileType AppLoader_NCCH::IdentifyType(FileUtil::IOFile& file) {
 }
 
 static std::string GetUpdateNCCHPath(u64_le program_id) {
+    u32 content_id = 0;
```
lioncash: This should probably be moved closer to where it's actually used.

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+    chunk_hash.Final(tmd_body.contentinfo[0].hash);
+
+    CryptoPP::SHA256 contentinfo_hash;
+    for (int i = 0; i < 64; i++) {
+        u8* contentinfo_data = static_cast<u8*>(static_cast<void*>(&tmd_body.contentinfo[i]));
+        chunk_hash.Update(contentinfo_data, sizeof(TMD_ContentInfo));
```
Subv: same

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
```
lioncash: `<string>` and `<vector>` need to be included

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+        if (file.WriteBytes(&chunk, sizeof(TMD_ContentChunk)) != sizeof(TMD_ContentChunk))
+            return Loader::ResultStatus::Error;
+    }
+
+    file.Close();
+    return Loader::ResultStatus::Success;
+}
+
+u64 TitleMetadata::GetTitleID() {
+    return static_cast<u64>(tmd_body.title_id);
```
Subv: There's no need for the casts in any of these functions, there's an implicit conversion from T_be to T

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+u32 TitleMetadata::GetManualContentID() {
+    return tmd_chunks[TMD_INDEX_MANUAL].id;
+}
+
+u32 TitleMetadata::GetDLPContentID() {
+    return tmd_chunks[TMD_INDEX_DLP].id;
+}
+
+void TitleMetadata::SetTitleID(u64 title_id) {
+    tmd_body.title_id = static_cast<u64_be>(title_id);
```
Subv: same in these functions

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <cryptopp/sha.h>
+#include "common/logging/log.h"
+#include "core/file_sys/title_metadata.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// FileSys namespace
+
+namespace FileSys {
+
+static const int TMD_INDEX_MAIN = 0;
```
lioncash: These constants can likely be an enum

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+void TitleMetadata::SetSystemVersion(u64 version) {
+    tmd_body.system_version = static_cast<u64_be>(version);
+}
+
+void TitleMetadata::AddContentChunk(TMD_ContentChunk& chunk) {
+    tmd_chunks.push_back(chunk);
+}
+
+void TitleMetadata::Print() {
+    LOG_WARNING(Service_FS, "%s - %u chunks", filepath.c_str(),
```
Subv: IMO these should be LOG_DEBUG instead of warnings

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+                static_cast<u16>(tmd_body.content_count));
```
Subv: u32

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    // Content info describes ranges of content chunks
+    LOG_WARNING(Service_FS, "Content info:");
+    for (int i = 0; i < 64; i++) {
```
Subv: You should define a constexpr variable somewhere and use that instead of using 64 as a magic constant

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+            tmd_chunks.push_back(chunk);
+        }
+    }
+
+    file.Close();
+    return Loader::ResultStatus::Success;
+}
+
+Loader::ResultStatus TitleMetadata::Save() {
+    FileUtil::IOFile file = FileUtil::IOFile(filepath, "wb");
```
lioncash: `auto file = FileUtil::IOFile(filepath, "wb");` or `FileUtil::IOFile file(filepath, "wb");`

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+        return 0x100;
+
+    case EllipticSha1:
+    case EcdsaSha256:
+        return 0x3C;
+    }
+}
+
+Loader::ResultStatus TitleMetadata::Load() {
+    FileUtil::IOFile file = FileUtil::IOFile(filepath, "rb");
```
lioncash: `auto file = FileUtil::IOFile(filepath, "rb");` or `FileUtil::IOFile file(filepath, "rb");`

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+#include "common/file_util.h"
+#include "common/swap.h"
+#include "core/loader/loader.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+/// TMD structures
+
+#pragma pack(1)
+
+struct TMD_ContentChunk {
```
Subv: Doesn't this structure have natural alignment? there's no need for it to be packed if so

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+    u32_be id;
+    u16_be index;
+    u16_be type;
+    u64_be size;
+    u8 hash[0x20];
+};
+
+static_assert(sizeof(TMD_ContentChunk) == 0x30, "TMD ContentChunk structure size is wrong");
+
+struct TMD_ContentInfo {
```
Subv: Same with this one

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+    u16_be index;
+    u16_be command_count;
+    u8 hash[0x20];
+};
+
+static_assert(sizeof(TMD_ContentInfo) == 0x24, "TMD ContentInfo structure size is wrong");
+
+struct TMD_Body {
+    char issuer[0x40];
```
Subv: can this be an u8 array instead of a char array? also, use std::array, i'll stop pointing this out for the rest of the arrays

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+/// TMD structures
+
+#pragma pack(1)
+
+struct TMD_ContentChunk {
+    u32_be id;
+    u16_be index;
+    u16_be type;
+    u64_be size;
+    u8 hash[0x20];
```
Subv: std::array

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+};
+
+static_assert(sizeof(TMD_ContentChunk) == 0x30, "TMD ContentChunk structure size is wrong");
+
+struct TMD_ContentInfo {
+    u16_be index;
+    u16_be command_count;
+    u8 hash[0x20];
```
Subv: std::array

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+    u32_be access_rights;
+    u16_be title_version;
+    u16_be content_count;
+    u16_be boot_content;
+    u8 reserved_4[2];
+    u8 contentinfo_hash[0x20];
+    TMD_ContentInfo contentinfo[64];
+};
+
+static_assert(sizeof(TMD_Body) == 0x9C4, "TMD body structure size is wrong");
```
Subv: I'm not a fan of the naming convention in these structures, can't you make them members of the TitleMetadata class and get rid of the TMD_ prefix?

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+#include "common/file_util.h"
```
lioncash: This include seems to be able to be moved to the cpp file.

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+#include "common/swap.h"
+#include "core/loader/loader.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+/// TMD structures
+
+#pragma pack(1)
```
Subv: #pragma pack(push, 1)
and #pragma pack(pop)

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+#include "common/file_util.h"
+#include "common/swap.h"
+#include "core/loader/loader.h"
```
lioncash: Ditto. This just needs a forward declaration:

```cpp
namespace Loader {
enum class ResultStatus;
}
```

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,122 @@
...
+    u32 GetManualContentID();
+    u32 GetDLPContentID();
+
+    void SetTitleID(u64 title_id);
+    void SetTitleType(u32 type);
+    void SetTitleVersion(u16 version);
+    void SetSystemVersion(u64 version);
+    void AddContentChunk(TMD_ContentChunk& chunk);
+
+    void Print();
```
Subv: this can be const

src/core/file_sys/title_metadata.h
Subv: These Get functions can be const

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    if (!file.WriteBytes(&tmd_signature[0], signature_size))
+        return Loader::ResultStatus::Error;
+
+    // The TMD body start position is rounded to the nearest 0x40 after the signature
+    u32 body_start = (signature_size + sizeof(u32_be) + 0x40) & ~0x3F;
+    file.Seek(body_start, SEEK_SET);
+
+    // Update our TMD body values and hashes
+    tmd_body.content_count = tmd_chunks.size();
```
lioncash: `tmd_chunks.size()` needs to be `static_cast`ed to a u16, otherwise this implicitly truncates a 64-bit size_t.

src/core/file_sys/title_metadata.cpp
lioncash: This should likely be a `size_t`, otherwise this truncates to a u32 implicitly (sizeof returns a size_t)

src/core/file_sys/title_metadata.cpp
lioncash: This should likely be a `size_t`, otherwise this truncates to a u32 implicitly (sizeof returns a size_t). The below log formatting specifier would need to be changed to `%zu` as well.

src/core/file_sys/ncch_container.cpp
```diff
@@ -247,6 +247,12 @@ Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vect
     std::string section_override = filepath + ".exefsdir/" + name;
     FileUtil::IOFile section_file = FileUtil::IOFile(section_override, "rb");
 
+    // Special override case for .code -> code.bin
+    if (!section_file.IsOpen() && !strcmp(name, ".code")) {
```
Subv: Should we really support both naming conventions?

src/core/file_sys/ncch_container.cpp
```diff
@@ -116,93 +116,160 @@ Loader::ResultStatus NCCHContainer::Load() {
...
-        return Loader::ResultStatus::Error;
-
     Loader::ResultStatus result = Load();
     if (result != Loader::ResultStatus::Success)
         return result;
 
-    if (!has_exefs)
+    // Check if we have files that can drop-in and replace
+    std::string section_override = filepath + ".exefsdir/" + name;
+    FileUtil::IOFile section_file = FileUtil::IOFile(section_override, "rb");
```
lioncash:
```cpp
auto section_file = FileUtil::IOFile(section_override, "rb");
```

or 

```cpp
FileUtil::IOFile section_file(section_override, "rb");
```

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+        return Loader::ResultStatus::Error;
+
+    for (int i = 0; i < tmd_body.content_count; i++) {
+        TMD_ContentChunk chunk;
+        if (file.ReadBytes(&chunk, sizeof(TMD_ContentChunk)) == sizeof(TMD_ContentChunk)) {
+            tmd_chunks.push_back(chunk);
+        }
+    }
+
+    file.Close();
```
lioncash: `file` will automatically close when it goes out of scope, so this isn't necessary.

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (file.WriteBytes(&tmd_body, sizeof(TMD_Body)) != sizeof(TMD_Body))
+        return Loader::ResultStatus::Error;
+
+    for (int i = 0; i < tmd_body.content_count; i++) {
+        TMD_ContentChunk chunk = tmd_chunks[i];
+        if (file.WriteBytes(&chunk, sizeof(TMD_ContentChunk)) != sizeof(TMD_ContentChunk))
+            return Loader::ResultStatus::Error;
+    }
+
+    file.Close();
```
lioncash: `file` will automatically close when it goes out of scope

src/core/file_sys/ncch_container.cpp
shinyquagsire23: Hmm, probably not I suppose. Would it just be better to map all files to their 3dstool counterpart, ie .code -> code.bin, icon -> icon.icn, banner -> banner.bnr, logo -> logo.bcma.lz?

Subv: That sounds like a good idea

shinyquagsire23:
Review status: 0 of 7 files reviewed at latest revision, 47 unresolved discussions.

---

*[src/core/file_sys/title_metadata.cpp, line 51 at r3](https://reviewable.io:443/reviews/citra-emu/citra/2975#-KvJKEcjnNhh5CTo8FRS-r3-51:-KvK1IgeFnEoMOMz-_kv:bnqp2c1) ([raw file](https://github.com/citra-emu/citra/blob/829b020aa398a1de89dc028fa175f3c8d3818cea/src/core/file_sys/title_metadata.cpp#L51)):*

Why is this a warning?
</blockquote></details>

Yeah that was just a test print for myself, removed now.

---

*[src/core/file_sys/title_metadata.h, line 36 at r3](https://reviewable.io:443/reviews/citra-emu/citra/2975#-KvJKEcjnNhh5CTo8FRT-r3-36:-KvK2APG4FOrL-uHZAi9:b9e8rtw) ([raw file](https://github.com/citra-emu/citra/blob/829b020aa398a1de89dc028fa175f3c8d3818cea/src/core/file_sys/title_metadata.h#L36)):*

can this be an u8 array instead of a char array? also, use std::array, i'll stop pointing this out for the rest of the arrays
</blockquote></details>

Had it as a char array because it's actually a readable/possibly printable string, but I doubt we'll ever really be checking certs

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2975)*

shinyquagsire23:
Fixed nits, there was some conflicting opinions on the for loops but I ended up using a u16 since the struct field was u16, though for the Print() for I used size_t since I checked against .size().

src/core/file_sys/title_metadata.cpp
```diff
@@ -77,134 +78,133 @@ Loader::ResultStatus TitleMetadata::Save() {
...
+        chunk_hash.Update(reinterpret_cast<u8*>(&tmd_chunks[i]), sizeof(ContentChunk));
     }
-    chunk_hash.Final(tmd_body.contentinfo[0].hash);
+    chunk_hash.Final(tmd_body.contentinfo[0].hash.data());
 
     CryptoPP::SHA256 contentinfo_hash;
-    for (int i = 0; i < 64; i++) {
-        u8* contentinfo_data = static_cast<u8*>(static_cast<void*>(&tmd_body.contentinfo[i]));
-        chunk_hash.Update(contentinfo_data, sizeof(TMD_ContentInfo));
+    for (int i = 0; i < tmd_body.contentinfo.size(); i++) {
```
Subv: size_t

src/core/file_sys/title_metadata.cpp
```diff
@@ -77,134 +78,133 @@ Loader::ResultStatus TitleMetadata::Save() {
...
+    LOG_DEBUG(Service_FS, "Content info:");
+    for (size_t i = 0; i < tmd_body.contentinfo.size(); i++) {
         if (tmd_body.contentinfo[i].command_count == 0)
             break;
 
-        LOG_WARNING(Service_FS, "    Index %04X, Command Count %04X",
-                    static_cast<u16>(tmd_body.contentinfo[i].index),
-                    static_cast<u16>(tmd_body.contentinfo[i].command_count));
+        LOG_DEBUG(Service_FS, "    Index %04X, Command Count %04X",
+                  static_cast<u16>(tmd_body.contentinfo[i].index),
```
Subv: These will all be promoted to int automatically, but i prefer it be explicit, `static_cast<u32>`

src/core/file_sys/title_metadata.cpp
```diff
@@ -77,134 +78,133 @@ Loader::ResultStatus TitleMetadata::Save() {
...
+    for (size_t i = 0; i < tmd_body.contentinfo.size(); i++) {
         u16 index = static_cast<u16>(tmd_body.contentinfo[i].index);
         u16 count = static_cast<u16>(tmd_body.contentinfo[i].command_count);
 
         if (count == 0)
             continue;
 
-        LOG_WARNING(Service_FS, "Content chunks for content info index %u:", i);
-        for (int j = index; j < index + count; j++) {
+        LOG_DEBUG(Service_FS, "Content chunks for content info index %u:", i);
```
Subv: %zu

src/core/file_sys/title_metadata.cpp
```diff
@@ -77,134 +78,133 @@ Loader::ResultStatus TitleMetadata::Save() {
...
                 break;
 
-            TMD_ContentChunk chunk = tmd_chunks[j];
-            LOG_WARNING(Service_FS, "    ID %08X, Index %04X, Type %04x, Size %016llx",
-                        static_cast<u32>(chunk.id), static_cast<u16>(chunk.index),
-                        static_cast<u16>(chunk.type), static_cast<u64>(chunk.size));
+            ContentChunk chunk = tmd_chunks[j];
+            LOG_DEBUG(Service_FS, "    ID %08X, Index %04X, Type %04x, Size %016llx",
+                      static_cast<u32>(chunk.id), static_cast<u16>(chunk.index),
+                      static_cast<u16>(chunk.type), static_cast<u64>(chunk.size));
```
Subv: Cast to u32.

As for the u64 cast, it won't work as you expect, you'll have to use an 64 bit format specifier like PRIu64

src/core/file_sys/title_metadata.cpp
```diff
@@ -0,0 +1,211 @@
...
+        if (count == 0)
+            continue;
+
+        LOG_DEBUG(Service_FS, "Content chunks for content info index %zu:", i);
+        for (u16 j = index; j < index + count; j++) {
+            // Don't attempt to print content we don't have
+            if (j > tmd_body.content_count)
+                break;
+
+            ContentChunk chunk = tmd_chunks[j];
```
lioncash: Should this be a `const ContentChunk&` so you aren't making copies of the structs instances in a loop?

src/core/file_sys/title_metadata.h
```diff
@@ -0,0 +1,125 @@
...
+
+enum TMDContentTypeFlag : u16 {
+    Encrypted = 1 << 1,
+    Disc = 1 << 2,
+    CFM = 1 << 3,
+    Optional = 1 << 14,
+    Shared = 1 << 15
+};
+
+enum TMDContentIndex { Main = 0, Manual = 1, DLP = 2 };
```
lioncash: This can be private within the TitleMetadata class since it's only ever used there.

lioncash:
> u16 stuff

FWIW, my reasoning for `size_t` is it results in slightly better codegen for no work, since `size_t` generally scales with the memory architecture on a platform (as it's required to be able to hold the maximum size of any theoretical possible object on a platform) so it removes the need for an extra zero or sign extension. Reasoning behind that being since it generally scales with memory arch, the compiler assumption can be made that the extra upper bits of the register won't contain junk data. [Example](https://godbolt.org/g/ifmPPU).

That said, it doesn't really matter much here.

src/core/file_sys/ncch_container.cpp
```diff
@@ -116,92 +116,143 @@ Loader::ResultStatus NCCHContainer::Load() {
...
-    // DLC can have an ExeFS and a RomFS but no extended header
-    if (ncch_header.exefs_size) {
-        exefs_offset = ncch_header.exefs_offset * kBlockSize;
-        u32 exefs_size = ncch_header.exefs_size * kBlockSize;
+    LoadOverrides();
 
-        LOG_DEBUG(Service_FS, "ExeFS offset:                0x%08X", exefs_offset);
-        LOG_DEBUG(Service_FS, "ExeFS size:                  0x%08X", exefs_size);
+    // We need at least one of these or overrides, practically
+    if (!(has_exefs | has_romfs | is_tainted))
```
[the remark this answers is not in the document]
shinyquagsire23: It was kinda intentional, but in retrospect || probably compiles nicer.

shinyquagsire23:
Got those changes sorted out, and yeah I don't think the slight differences in compiler output will matter too much in this case at least. Though, there will be cases later where title scanning will have to load several TMDs but even then I don't think this code will end up being in any hot code, mostly just loading stuff.

---

Review status: 0 of 7 files reviewed at latest revision, 55 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/2975#-:-KvL36euYi3xMGh4DY7f:b-y709ud)*

Dragios:
Your PR introduce 1 new warning. If you can fix it in upcoming PR, that will be great.

'Loader::AppLoader_NCCH::ReadUpdateRomFS': not all control paths return a value
`src\core\loader\ncch.cpp` -------------------------------------------------------------- _Line 242_ [core]

PR #2977, opened 2017-10-01
SharedMemory: Don't take over and unmap the source memory block when creating a shared memory, just reference it

Subv:
Fixes a potential issue where two shared memories pointing to the same memory address might not have the same backing memory.

# Bug scenario
Consider the following,
Processes = [A, B, C]
SharedMemories = [S, T] pointing to address K

1. Process A creates SharedMemory S pointing to address K, thus A's memory mapping for K now points to a block owned by S.
2. Process B maps S into its address space.
3. Process C maps S into its address space.
-- At this point all 3 processes have the memory pointed to by S mapped into their address space --
4. Process A creates SharedMemory T pointing to address K, thus A's memory mapping for K now points to a block owned by T, which is different than the one owned by S.
5. A writes some data at address K.
6. Processes B and C do not see the changes in the data because their memory mapping still points to a memory block owned by S.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2977)

### DIFF
--- a/src/core/hle/kernel/shared_memory.cpp
+++ b/src/core/hle/kernel/shared_memory.cpp
@@ -55,22 +55,19 @@ SharedPtr<SharedMemory> SharedMemory::Create(SharedPtr<Process> owner_process, u
             Kernel::g_current_process->vm_manager.RefreshMemoryBlockMappings(linheap_memory.get());
         }
     } else {
-        // TODO(Subv): What happens if an application tries to create multiple memory blocks
-        // pointing to the same address?
         auto& vm_manager = shared_memory->owner_process->vm_manager;
         // The memory is already available and mapped in the owner process.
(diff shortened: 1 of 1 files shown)


src/core/hle/kernel/shared_memory.cpp
```diff
@@ -60,17 +60,10 @@ SharedPtr<SharedMemory> SharedMemory::Create(SharedPtr<Process> owner_process, u
         auto& vm_manager = shared_memory->owner_process->vm_manager;
         // The memory is already available and mapped in the owner process.
         auto vma = vm_manager.FindVMA(address)->second;
```
Subv: This should verify that the returned VMA actually refers to the address where we want to map the SharedMemory, since FindVMA might return a VMA for a more general region

PR #2982, opened 2017-10-02
macOS: Build x86_64h slice

MerryMage:
This commit produces a fat-binary with two slices. The x86_64 slice
is for all x64 systems, and the x86_64h slice targets x64 systems
starting with Haswell. The latter allows the compiler to use newer
instructions that are not available on older microarchitectures.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2982)

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -52,8 +52,8 @@ elif [ "$TRAVIS_OS_NAME" = "osx" ]; then
     export Qt5_DIR=$(brew --prefix)/opt/qt5
 
     mkdir build && cd build
-    cmake .. -DUSE_SYSTEM_CURL=ON -GXcode
-    xcodebuild -configuration Release
+    cmake .. -DUSE_SYSTEM_CURL=ON -DCMAKE_OSX_ARCHITECTURES="x86_64;x86_64h" -DCMAKE_BUILD_TYPE=Release
+    make -j4
 
     ctest -VV -C Release
 fi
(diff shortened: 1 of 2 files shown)


B3n30:
I didn't test this PR but with my latest test xcode compiled the code a lot better compared to make(gcc?). Isn't it possible to build with xcode for haswell?

I also can't test the compile for haswell, since my OSX hasn't got one

jroweboy:
Couple of quick questions before I click that green merge button. Whats the size difference after this change? (Should be no more than a few extra MB right?) Have you double checked that this doesn't break anything with our custom post build deploy steps? (I'm guessing this isn't an issue, but its worth checking before clicking merge and I don't have a mac or I woulda done it myself)

MerryMage:
> xcode compiled the code a lot better compared to make(gcc?).

They should be using the same compiler. You can see this by running `/Applications/Xcode.app/Contents/Developer/usr/bin/g++ --version`, which would show it's really just clang and not gcc.

> Whats the size difference after this change? (Should be no more than a few extra MB right?)

Double binary size. Libraries remain the same size. A few extra MB.

> Have you double checked that this doesn't break anything with our custom post build deploy steps?

Tried it, paths changed, fixed that.

B3n30:
without GXcode:
`/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++`
with GXcode:
`/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++`

MerryMage:
@B3n30 

Sure, but c++ is just an alias of clang++.

    $ ls -la c++
    lrwxr-xr-x  1 root  wheel  5 20 Mar  2017 c++ -> clang

PR #2983, opened 2017-10-02
[Qt] Adding thumbnail toolbar for windows

Styleoshin:
I added the thumbnail toolbar for ui, and for shortcut. She contains a :
- Button play/pause emulation.
- Button stop emulation.
- Button restart game.

The icons are free (create by a friend)
For build in windows with option CITRA_USE_BUNDLED_QT, you need to add Qt5WinExtras ?

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2983)


jroweboy:
Thanks for the second first time contribution <3

Heres my quick thoughts on it. I haven't looked at the code yet since I know i'm probably going to have a lotta feedback on that, and I'd like to get the design settled before we talk code :)

* I like the idea; its the simple things like this that make citra look more like a native windows application.
* Don't like having a "Restart Game" button. We don't have it anywhere else in the emulator right now, and I'm not confident its something that would get used even if we did have one.
* As you mentioned, this needs an update to the bundled QT, so please submit a PR to https://github.com/citra-emu/ext-windows-bin with the updated QT libs

Styleoshin:
Thank's, I do my best to contribute well.

- A design problem?

- Yes, i added the functionality restart game in a button, because in many emulators use this function on menu action, it's still a shortcut.

- For PR in : https://github.com/citra-emu/ext-windows-bin , it's no necessary if this PR is not accepted ?

src/citra_qt/main.h
```diff
@@ -134,6 +144,16 @@ private slots:
 private:
     void UpdateStatusBar();
 
+#ifdef _WIN32
```
jroweboy: Instead of putting `#ifdef windows` everywhere, I'd much rather have everything windows specific in a private header thats is conditionally included so we don't have several functions that are `ifdef`

jroweboy: Feel free to ask a little on IRC/Discord in case i need to be more clear about what I'm meaning here (sorry, feeling sick today but I did promise a review :P)

Styleoshin:
I implemented the WindowsExtras class to group native Windows elements, so if you want to add native elements, this will be done in this class

B3n30:
@Styleoshin Are you going to address the issues and fix the conflicts or should some other dev take over on this task? I think I would be beneficial for citra to have this feature.

B3n30:
Closed due to inactivity

B3n30:
Sorry. I missed the activity, so I'll reopen it.

adityaruplaha:
Is this one dead? Its been inactive for a month...

PR #2990, opened 2017-10-04
swkbd: basic implementation

acnleditor2:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2990)


jroweboy:
I appreciate the effort, but you did not read the comment on the last PR or you did not understand the reason it was originally closed. Please reread the comments on the original swkbd PR and see why it was closed. Using stdin is not an acceptable way to get input from the user. End of story. A true HLE swkbd will need to draw to framebuffer and read the input from HID like a service should do, and if you do those things, then please open a PR. But please do not open another PR for any swkbd implementation that reads from stdin.

PR #2991, opened 2017-10-04
Remove more usages of GetPointer.

Subv:
Memory::GetPointer is an unsafe function that will only bring us potential out-of-bounds access and buffer overruns if the emulated address space stops being always contiguous for whatever reason.

After this PR there are only 7 remaining references to GetPointer, 2 of which are the declaration and definition.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2991)

### DIFF
--- a/src/core/hle/function_wrappers.h
+++ b/src/core/hle/function_wrappers.h
@@ -69,11 +69,10 @@ void Wrap() {
     FuncReturn(retval);
 }
 
-template <ResultCode func(s32*, u32*, s32, u32)>
+template <ResultCode func(s32*, VAddr, s32, u32)>
 void Wrap() {
     s32 param_1 = 0;
-    u32 retval =
-        func(&param_1, (Kernel::Handle*)Memory::GetPointer(PARAM(1)), (s32)PARAM(2), PARAM(3)).raw;
+    u32 retval = func(&param_1, PARAM(1), (s32)PARAM(2), PARAM(3)).raw;
 
     Core::CPU().SetReg(1, (u32)param_1);
     FuncReturn(retval);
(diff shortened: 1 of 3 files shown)


src/core/hle/function_wrappers.h
```diff
@@ -58,22 +58,21 @@ void Wrap() {
...
 
-template <ResultCode func(s32*, u32*, s32, bool, s64)>
+template <ResultCode func(s32*, VAddr, s32, bool, s64)>
 void Wrap() {
     s32 param_1 = 0;
-    s32 retval = func(&param_1, (Kernel::Handle*)Memory::GetPointer(PARAM(1)), (s32)PARAM(2),
-                      (PARAM(3) != 0), (((s64)PARAM(4) << 32) | PARAM(0)))
-                     .raw;
+    s32 retval =
+        func(&param_1, PARAM(1), (s32)PARAM(2), (PARAM(3) != 0), (((s64)PARAM(4) << 32) | PARAM(0)))
```
MerryMage: C-style cast?

src/core/hle/function_wrappers.h
```diff
@@ -58,22 +58,21 @@ void Wrap() {
...
     FuncReturn(retval);
 }
 
-template <ResultCode func(s32*, u32*, s32, u32)>
+template <ResultCode func(s32*, VAddr, s32, u32)>
 void Wrap() {
     s32 param_1 = 0;
-    u32 retval =
-        func(&param_1, (Kernel::Handle*)Memory::GetPointer(PARAM(1)), (s32)PARAM(2), PARAM(3)).raw;
+    u32 retval = func(&param_1, PARAM(1), (s32)PARAM(2), PARAM(3)).raw;
```
MerryMage: C-style cast?

src/core/hle/function_wrappers.h
Subv: The previous code already did this, should i change it?

MerryMage: *shrugs* I'm personally fine with it as-is for this PR

PR #2992, opened 2017-10-04
Services/AC: Converted the ac:i and ac:u services to the new service framework.

Subv:
# ~Concerns~

~There are a few service functions whose command header doesn't seem to match their parameters, why does `CreateDefaultConfig`, with a command header of `0x00010000`, work on a static buffer?~

~3dbrew has no information on that function~

I've reverse engineered a bit of the AC module and fixed some missing/wrong parameters in the response command buffers.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2992)

### DIFF
--- a/src/core/hle/service/ac/ac.cpp
+++ b/src/core/hle/service/ac/ac.cpp
@@ -2,11 +2,10 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include <array>
-
 #include "common/common_types.h"
 #include "common/logging/log.h"
 #include "core/hle/ipc.h"
+#include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/kernel/handle_table.h"
 #include "core/hle/result.h"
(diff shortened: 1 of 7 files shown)


acnleditor2:
error: Debug <Critical> core/hle/service/ac/ac.cpp:operator():26: Assertion Failed!
Output buffer size not equal ACConfig size

Subv:
@acnleditor2 What game is that?

acnleditor2:
@Subv Tomodachi Life, with wwylele/cec merged. Repo: acnleditor2/citra

wwylele:
> why does CreateDefaultConfig, with a command header of 0x00010000, work on a static buffer?

Because it is an service->client static buffer, which only requires the client setup a static buffer (a "mailbox") descriptor in TLS+0x180 (not the command buffer TLS+0x80). This is never counted as an parameter, so it leaves command buffer with zero parameter. In turn, the service is supposed to put a static buffer translation descriptor in the reply, so that the kernel will transfer the data from the service to the client "mailbox" according to the static buffer descriptor. Another implication of this is that the client mailbox doesn't necessarily matches the size of actual transferred data. The mailbox size is only required large enough to hold data of any expected size.

src/core/hle/service/ac/ac.h
```diff
@@ -4,131 +4,153 @@
...
+         * AC::SetClientVersion service function
+         *  Inputs:
+         *      1 : Used SDK Version
+         *  Outputs:
+         *      1 : Result of function, 0 on success, otherwise error code
+         */
+        void SetClientVersion(Kernel::HLERequestContext& ctx);
+
+    protected:
+        std::shared_ptr<Module> ac;
```
lioncash: `<array>` and `<memory>` need to be included.

src/core/hle/service/ac/ac_i.h
```diff
@@ -4,18 +4,15 @@
...
 
-class AC_I final : public Interface {
+class AC_I final : public Module::Interface {
 public:
-    AC_I();
-
-    std::string GetPortName() const override {
-        return "ac:i";
-    }
+    AC_I(std::shared_ptr<Module> ac);
```
lioncash: This should likely be an explicit constructor,

src/core/hle/service/ac/ac_u.h
```diff
@@ -4,18 +4,15 @@
...
 
-class AC_U final : public Interface {
+class AC_U final : public Module::Interface {
 public:
-    AC_U();
-
-    std::string GetPortName() const override {
-        return "ac:u";
-    }
+    AC_U(std::shared_ptr<Module> ac);
```
lioncash: ditto

src/core/hle/service/ac/ac.h
```diff
@@ -4,131 +4,153 @@
...
- */
-void SetClientVersion(Interface* self);
-
-/// Initialize AC service
-void Init();
-
-/// Shutdown AC service
-void Shutdown();
+class Module {
+public:
```
lioncash: Doesn't this only need to be protected?

src/core/hle/service/ac/ac.cpp
```diff
@@ -17,169 +17,170 @@
...
 }
 
-void SetClientVersion(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
+void Module::Interface::SetClientVersion(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp(ctx, 0x40, 1, 2);
 
-    const u32 version = cmd_buff[1];
-    self->SetVersion(version);
+    u32 version = rp.Pop<u32>();
```
jroweboy: Why was `const` removed?

src/core/hle/service/ac/ac.cpp
```diff
@@ -3,10 +3,10 @@
 // Refer to the license.txt file included.
 
 #include <array>
```
jroweboy: don't need array anymore

src/core/hle/service/ac/ac.cpp
Subv: No particular reason, it didn't do anything

acnleditor2:
[with wwylele/cec merged] tomodachi life crash assertion on click on settings in town hall.

PR #2966, opened 2017-09-28
Qtifw build installer

jroweboy:
Depends on: https://github.com/citra-emu/ext-windows-bin/pull/10

Supercedes: https://github.com/citra-emu/citra/pull/2929

(Got tired of waiting for @j-selby to get internet back)

This reworks the installer to be built through cmake instead of powershell/bash scripts. I tested windows, but if someone would be so kind as to test macOS and linux.

After this PR, there is one more followup PR for QT gui integration to allow users to update through the UI. When that follow up PR is merged, the download page will need to be reworked as well to point directly to the new installer instead. (We will need lots of testing of course!)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2966)

PR #2997, opened 2017-10-08
Updater QT Integration

j-selby:
~~Blocked by https://github.com/citra-emu/ext-windows-bin/pull/11, #2966~~. Related to https://github.com/citra-emu/citra-web/pull/56.

Original implementation by @jroweboy. Based on https://github.com/Skycoder42/QtAutoUpdater.

This implements support for Qt integration into the main UI. The installer binary is searched for in the parent directory, labelled "maintenancetool[.exe]".

Autoupdating can be configured via Citra's configuration menu, as well as Help->Open Updater/Check for Updates. These settings/options are hidden when the installer binary does not exist.

Two separate updating mechanisms are implemented:
- Check on startup (default). Popup is shown at startup which launches the main updater UI in updating mode if Yes is selected.
- Silent updating. Updater binary is launched at shutdown, which silently updates Citra (if required).

Relatively untested on Linux, and untested on Macs.

UI Screenshots:

image
image
image (if menubar option manually clicked)
image

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/2997)


src/citra_qt/main.cpp
```diff
@@ -345,6 +368,45 @@ void GMainWindow::OnDisplayTitleBars(bool show) {
...
+    }
+}
+
+void GMainWindow::OnUpdateFound(bool found, bool error) {
+    if (error) {
+        LOG_WARNING(Frontend, "Update check failed.");
+        return;
+    }
+
+    if (found) {
```
lioncash: It's possible to unindent much of this by comparing the no update case first via

```cpp
if (!found) {
    LOG_INFO(Frontend, "No updates found");
    return;
}

...
```

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+#endif
+
+Updater::Updater(QObject* parent) : Updater(DEFAULT_TOOL_PATH, parent) {}
+
+Updater::Updater(const QString& maintenance_tool_path, QObject* parent)
+    : QObject(parent), backend(new UpdaterPrivate(this)) {
+    backend->tool_path = UpdaterPrivate::toSystemExe(maintenance_tool_path);
+}
+
+Updater::~Updater() {}
```
lioncash:
```cpp
Updater::~Updater() = default;
```

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+}
+
+UpdaterPrivate::~UpdaterPrivate() {
+    if (main_process && main_process->state() != QProcess::NotRunning) {
+        main_process->kill();
+        main_process->waitForFinished(1000);
+    }
+}
+
+const QString UpdaterPrivate::toSystemExe(QString basePath) {
```
lioncash: `const` isn't really necessary here. Also `base_path`

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+bool UpdaterPrivate::HasUpdater() {
+    QFileInfo toolInfo(QCoreApplication::applicationDirPath(), tool_path);
+    return toolInfo.exists();
+}
+
+bool UpdaterPrivate::StartUpdateCheck() {
+    if (running || !HasUpdater()) {
+        return false;
+    } else {
```
lioncash: This `else` can be omitted and the indentation reduced, since the if statement body contains a return.

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+        running = true;
+
+        emit q->UpdateInfoChanged(update_info);
+        emit q->RunningChanged(true);
+        return true;
+    }
+}
+
+void UpdaterPrivate::StopUpdateCheck(int delay, bool async) {
+    if (main_process && main_process->state() != QProcess::NotRunning) {
```
lioncash:
```cpp
if (!main_process || main_process->state() == QProcess::NotRunning) {
    return;
}
```
can be used to unindent the rest of the code.

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+        LOG_ERROR(Frontend, "Cannot read xml for update: %s", reader.errorString());
+        return XMLParseResult::INVALID_XML;
+    }
+
+    out = updates;
+    return XMLParseResult::SUCCESS;
+}
+
+void UpdaterPrivate::UpdaterReady(int exitCode, QProcess::ExitStatus exitStatus) {
+    if (main_process) {
```
lioncash: You can invert this like:

```cpp
if (!main_process) {
    return;
}
```

and make the inner if like

```cpp
if (exitStatus != QProcess::NormalExit) {
    UpdaterError(QProcess::Crashed);
    return;
}
```

to unindent the normal exit path quite a bit.

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+            }
+            emit q->CheckUpdatesDone(!update_info.isEmpty(), hasError);
+        } else {
+            UpdaterError(QProcess::Crashed);
+        }
+    }
+}
+
+void UpdaterPrivate::UpdaterError(QProcess::ProcessError error) {
+    if (main_process) {
```
lioncash:
```cpp
if (!main_process) {
    return;
}
```

src/citra_qt/updater/updater_p.h
```diff
@@ -0,0 +1,67 @@
...
+public slots:
+    void UpdaterReady(int exit_code, QProcess::ExitStatus exit_status);
+    void UpdaterError(QProcess::ProcessError error);
+
+    void AboutToExit();
+
+private:
+    XMLParseResult ParseResult(const QByteArray& output, QList<Updater::UpdateInfo>& out);
+
+    Updater* q;
```
lioncash: Should this likely be more informatively named?

src/citra_qt/updater/updater_p.h
```diff
@@ -0,0 +1,67 @@
...
+
+    QString tool_path;
+    QList<Updater::UpdateInfo> update_info;
+    bool normal_exit;
```
lioncash: Shouldn't these bools (and ints) have a sane default value for safety?

src/citra_qt/updater/updater_p.h
```diff
@@ -0,0 +1,67 @@
...
+//
+// Based on the original work by Felix Barx
+// Copyright (c) 2015, Felix Barz
+// All rights reserved.
+
+#pragma once
+
+#include <QtCore/QProcess>
+
+#include "updater.h"
```
lioncash: Includes should be relative, like: `"citra-qt/updater/updater.h"`

src/citra_qt/updater/updater_p.h
```diff
@@ -0,0 +1,67 @@
...
+};
+
+class UpdaterPrivate : public QObject {
+    Q_OBJECT;
+
+public:
+    explicit UpdaterPrivate(Updater* q_ptr);
+    ~UpdaterPrivate();
+
+    static const QString toSystemExe(QString basePath);
```
lioncash: `base_path`, also this should likely be named `ToSystemExe` to be consistent with the other function names.

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+        main_process->deleteLater();
+        main_process = nullptr;
+
+        running = false;
+        emit q->RunningChanged(false);
+        emit q->CheckUpdatesDone(false, true);
+    }
+}
+
+void UpdaterPrivate::LaunchWithArguments(QStringList &args) {
```
lioncash: As far as I can tell, args can be a const reference (ampersand against the type as well)

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+    if (!HasUpdater()) {
+        return;
+    }
+
+    QFileInfo toolInfo(QCoreApplication::applicationDirPath(), tool_path);
```
lioncash: `tool_info`

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+}
+
+void Updater::LaunchUIOnExit() {
+    backend->LaunchUIOnExit();
+}
+
+Updater::UpdateInfo::UpdateInfo() : name(), version(), size(0ull) {}
+
+Updater::UpdateInfo::UpdateInfo(const Updater::UpdateInfo& other)
+    : name(other.name), version(other.version), size(other.size) {}
```
lioncash: Isn't this just doing what the regular copy constructor does? As in, `UpdateInfo(const UpdateInfo&) = default;`

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+Updater::UpdateInfo::UpdateInfo(QString name, QVersionNumber version, quint64 size)
+    : name(name), version(version), size(size) {}
```
lioncash: The name string can be `std::move`d

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+UpdaterPrivate::UpdaterPrivate(Updater* q_ptr)
+    : QObject(nullptr), q(q_ptr), tool_path(), update_info(), normal_exit(true),
+      last_error_code(EXIT_SUCCESS), last_error_log(), running(false), main_process(nullptr),
```
lioncash: many of these arguments can just be assigned to member variables directly in the class body.

e.g.

```cpp
class Thing
{
private:
    bool some_thing = false;
};

File: src/citra_qt/updater/updater.h [hunk 17]
```diff
@@ -0,0 +1,97 @@
...
+
+public slots:
+    //! Starts checking for updates
+    bool CheckForUpdates();
+    //! Aborts checking for updates
+    void AbortUpdateCheck(int maxDelay = 5000, bool async = false);
+
+signals:
+    //! Will be emitted as soon as the updater finished checking for updates
+    void CheckUpdatesDone(bool hasUpdates, bool hasError);
```
lioncash: `has_updates`, `has_error`

src/citra_qt/updater/updater.h
```diff
@@ -0,0 +1,97 @@
...
+    bool HasUpdater();
+
+    //! Instead of silently updating, explictly open the UI on shutdown
+    void LaunchUIOnExit();
+
+public slots:
+    //! Starts checking for updates
+    bool CheckForUpdates();
+    //! Aborts checking for updates
+    void AbortUpdateCheck(int maxDelay = 5000, bool async = false);
```
lioncash: `max_delay`

src/citra_qt/updater/updater.h
```diff
@@ -0,0 +1,97 @@
...
+
+signals:
+    //! Will be emitted as soon as the updater finished checking for updates
+    void CheckUpdatesDone(bool hasUpdates, bool hasError);
+
+    //! notifyAcFn{Updater::running}
+    void RunningChanged(bool running);
+    //! notifyAcFn{Updater::updateInfo}
+    void UpdateInfoChanged(QList<Updater::UpdateInfo> UpdateInfo);
```
lioncash: `update_info`

src/citra_qt/updater/updater_p.h
```diff
@@ -0,0 +1,67 @@
...
+// Copyright (c) 2015, Felix Barz
+// All rights reserved.
+
+#pragma once
+
+#include <QtCore/QProcess>
+
+#include "updater.h"
+
+enum class XMLParseResult {
```
lioncash: Our general style for `enum class`es is to case them like regular identifiers (ISO C++ Guidelines also [advise](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Renum-caps) against this). So in this case:

```cpp
enum class XMLParseResult {
    Success,
    NoUpdate,
    InvalidXML,
};

File: src/citra_qt/updater/updater.cpp [hunk 21]
```diff
@@ -0,0 +1,292 @@
...
+#ifdef Q_OS_OSX
+#define DEFAULT_TOOL_PATH QStringLiteral("../../../maintenancetool")
+#else
+#define DEFAULT_TOOL_PATH QStringLiteral("../maintenancetool")
+#endif
+
+Updater::Updater(QObject* parent) : Updater(DEFAULT_TOOL_PATH, parent) {}
+
+Updater::Updater(const QString& maintenance_tool_path, QObject* parent)
+    : QObject(parent), backend(new UpdaterPrivate(this)) {
```
lioncash: `make_unique`?

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+void Updater::SilentlyUpdate() {
+    backend->SilentlyUpdate();
+}
+
+void Updater::LaunchUIOnExit() {
+    backend->LaunchUIOnExit();
+}
+
+Updater::UpdateInfo::UpdateInfo() : name(), version(), size(0ull) {}
```
lioncash: You can just make the struct body like so to avoid the need for initializing these directly in the initializer list:

```cpp
//! Provides information about updates for components
struct UpdateInfo {
    //! The name of the component that has an update
    QString name;
    //! The new version for that compontent
    QVersionNumber version;
    //! The update download size (in Bytes)
    quint64 size = 0;

    //! Default Constructor
    UpdateInfo();
    //! Copy Constructor
    UpdateInfo(const UpdateInfo& other);
    //! Constructor that takes name, version and size
    UpdateInfo(QString name, QVersionNumber version, quint64 size);
};
```

This can then be defaulted as:

```cpp
Updater::UpdateInfo::UpdateInfo() = default;
```

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,292 @@
...
+    if (basePath.endsWith(QStringLiteral(".app")))
+        basePath.truncate(basePath.lastIndexOf(QStringLiteral(".")));
+    return basePath + QStringLiteral(".app/Contents/MacOS/") + QFileInfo(basePath).fileName();
+#elif defined(Q_OS_UNIX)
+    return basePath;
+#endif
+}
+
+bool UpdaterPrivate::HasUpdater() {
+    QFileInfo toolInfo(QCoreApplication::applicationDirPath(), tool_path);
```
lioncash: `tool_info`

src/citra_qt/updater/updater.h
```diff
@@ -0,0 +1,97 @@
...
+    QList<UpdateInfo> LatestUpdateInfo() const;
+
+    //! Launches the updater UI formally
+    void LaunchUI();
+
+    //! Sliently updates the application in the background
+    void SilentlyUpdate();
+
+    //! Checks to see if a updater application is available
+    bool HasUpdater();
```
lioncash: As far as I'm aware, this can be a const member function.

src/citra_qt/updater/updater_p.h
```diff
@@ -0,0 +1,67 @@
...
+class UpdaterPrivate : public QObject {
+    Q_OBJECT;
+
+public:
+    explicit UpdaterPrivate(Updater* q_ptr);
+    ~UpdaterPrivate();
+
+    static const QString toSystemExe(QString basePath);
+
+    bool HasUpdater();
```
lioncash: As far as I'm aware, this can be a const member function.

src/citra_qt/configuration/configure_general.ui
```diff
@@ -25,16 +25,44 @@
...
              <string>Search sub-directories for games</string>
             </property>
            </widget>
           </item>
+         </layout>
+        </item>
+       </layout>
+      </widget>
+     </item>
+
```
B3n30: Usually a ui file doesn't have blank lines like this

src/citra_qt/configuration/configure_general.ui
```diff
@@ -149,8 +178,7 @@
              </widget>
             </item>
             <item>
-             <widget class="QComboBox" name="theme_combobox">
-             </widget>
+             <widget class="QComboBox" name="theme_combobox"/>
```
B3n30: (Unintentional change?, but the change is fine)

src/citra_qt/main.cpp
```diff
@@ -345,6 +368,46 @@ void GMainWindow::OnDisplayTitleBars(bool show) {
     }
 }
 
+void GMainWindow::OnCheckForUpdates() {
+    this->CheckForUpdates();
```
B3n30: is `this->` necessary here?

src/citra_qt/main.cpp
```diff
@@ -345,6 +368,46 @@ void GMainWindow::OnDisplayTitleBars(bool show) {
...
+
+    if (!found) {
+        LOG_INFO(Frontend, "No updates found");
+        return;
+    }
+
+    LOG_INFO(Frontend, "Update found!");
+    auto result = QMessageBox::question(
+        this, tr("Update available!"),
+        tr("A update has been found for Citra. Do you wish to install it now?<br /><br />"
```
B3n30: Nit: "A update for citra is available. Do you wish to install it now?"
I somehow don't like to use found here

src/citra_qt/main.ui
```diff
@@ -45,7 +45,7 @@
      <x>0</x>
      <y>0</y>
      <width>1081</width>
-     <height>19</height>
+     <height>21</height>
```
B3n30: Nit: Unintentional change?

src/citra_qt/main.ui
```diff
@@ -196,6 +199,17 @@
     <string>Create Pica Surface Viewer</string>
    </property>
   </action>
+  <action name="action_Open_Maintenance_Tool">
+   <property name="text">
+    <string>Open Updater</string>
```
B3n30: I don't like that label. For a user it isn't clear what is the difference between "Check for  Updates" and "Open Updater". I have no good idea for it but maybe something like "Install Citra components"

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,294 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+//
+// Based on the original work by Felix Barx
+// Copyright (c) 2015, Felix Barz
+// All rights reserved.
```
B3n30: The license of QtAutoUpdater states: 

* Redistributions of source code must retain the above copyright notice, this
  list of conditions and the following disclaimer.

* Redistributions in binary form must reproduce the above copyright notice,
  this list of conditions and the following disclaimer in the documentation
  and/or other materials provided with the distribution.

How does this PR handle this?

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+#include <QCoreApplication>
+#include <QDir>
+#include <QFileInfo>
+#include <QTimer>
+#include <QXmlStreamReader>
+
```
B3n30: No blank line between includes

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,294 @@
...
+        }
+    } else {
+        main_process->kill();
+        main_process->waitForFinished(100);
+    }
+}
+
+XMLParseResult UpdaterPrivate::ParseResult(const QByteArray& output,
+                                           QList<Updater::UpdateInfo>& out) {
+    const auto outString = QString::fromUtf8(output);
```
B3n30: out_string

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,294 @@
...
+    const auto xmlBegin = outString.indexOf(QStringLiteral("<updates>"));
```
B3n30: xml_begin

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,294 @@
...
+    if (xmlBegin < 0)
+        return XMLParseResult::NoUpdate;
+    const auto xmlEnd = outString.indexOf(QStringLiteral("</updates>"), xmlBegin);
```
B3n30: xml_end

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,294 @@
...
+    if (reader.hasError()) {
+        LOG_ERROR(Frontend, "Cannot read xml for update: %s", reader.errorString().toStdString().c_str());
+        return XMLParseResult::InvalidXML;
+    }
+
+    out = updates;
+    return XMLParseResult::Success;
+}
+
+void UpdaterPrivate::UpdaterReady(int exitCode, QProcess::ExitStatus exitStatus) {
```
B3n30: exit_code, exit_status

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    if (exitStatus != QProcess::NormalExit) {
+        UpdaterError(QProcess::Crashed);
+    }
+
+    normal_exit = true;
+    last_error_code = exitCode;
+    last_error_log = main_process->readAllStandardError();
+    const auto updateOut = main_process->readAllStandardOutput();
```
B3n30: update_out

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,294 @@
...
+    return XMLParseResult::Success;
+}
+
+void UpdaterPrivate::UpdaterReady(int exitCode, QProcess::ExitStatus exitStatus) {
+    if (main_process == nullptr) {
+        return;
+    }
+
+    if (exitStatus != QProcess::NormalExit) {
+        UpdaterError(QProcess::Crashed);
```
B3n30: return; ?

src/citra_qt/updater/updater.cpp
```diff
@@ -0,0 +1,294 @@
...
+    const auto updateOut = main_process->readAllStandardOutput();
+    main_process->deleteLater();
+    main_process = nullptr;
+
+    running = false;
+    emit parent->RunningChanged(false);
+
+    QList<Updater::UpdateInfo> update_info;
+    auto err = ParseResult(updateOut, update_info);
+    bool hasError = false;
```
B3n30: has_error

src/citra_qt/updater/updater.h
```diff
@@ -0,0 +1,97 @@
...
+#include <QDateTime>
+#include <QList>
+#include <QScopedPointer>
+#include <QString>
+#include <QStringList>
+#include <QVersionNumber>
+
+class UpdaterPrivate;
+
+//! The main updater. Can check for updates and run the maintenancetool as updater
```
B3n30: `//!` isn't the comment style we use here: `///` or `//<`

src/citra_qt/main.cpp
Dragios: It should be "An update".

src/citra_qt/main.ui
j-selby: "Modify Citra Install" maybe?

src/citra_qt/updater/updater.cpp
j-selby: Good point. At the same time, are we also including the licenses for other stuff somewhere? Soundtouch etc?

j-selby: Added the full license in there.

src/citra_qt/main.ui
B3n30: sounds fine to me

j-selby:
From Discord:
```
wwylele / 白疾風 - Today at 8:31 PM
> a normal user isn't going to have loaded a game
cci/cxi file format is linked to citra on my machine, and I can double click the file to start it up in citra immediately
CitraBotBOT - Today at 8:32 PM
<neobrain> you're just adding UI verbosity to the update dialog for a scenario that could easily be avoided (e.g. by not allowing games to be started until update is ran through, by aborting the update automatically [because let's be honest is it going to be critical enough to annoy the user about it], etc)
Selby - Today at 8:33 PM
yeah, it could be suppressed, true
```

RavenHome1:
i am more supportive of keeping it simple

user opens citra -> citra checks for update and asks the user if they want to update -> if user selects yes then citra updates and restarts and do what ever it needs to do.

if user selected no then continue running citra normally -> user can select check for updates from the help drop down menu and citra will check and ask the user to update.

if emulation was already running then grey out the check for updates button and prevent any update operations until the user brings emulation to a full stop (basically check that no emulation is happening)

adding the mentioned updates and silent run mode just creates superfluous and confusing options.

also perhaps look into hiding maintenancetool[.exe] so the user won't run it out of curiosity and cause some problems (if there is any chance of that happening)

B3n30:
@RavenHome1 
- Your suggestions are mainly what this PR does (except for my tiny popup wish and the greyed out menu during emulation).
- But I still think a silent update mode is necessary and not to confusing.
- Hiding the maintenancetool isn't necessary since it is basically the deinstaller and provides the option to install additional stuff. So it is fine if the user can see it/ use it. Thats to tool that gets opened when you select "Open Updater", which is now renamed to "Modify Citra Install". From here you can select if you want to use canary or nightly and might also be able to install custom system archives, ...

RavenHome1:
@B3n30 

- i guess i was not paying much attention, thanks for clearing that up

- i still don't find a need for the silent updater, giving the user the option at citra startup is quite enough.

_ "the modify citra install" now seems much better when you cleared it up and named it a better name but will it still be needed in the future. will canary and nightly remain forever or will they be merged together at some point in a more stable release?.

jroweboy:
* we didn't choose the name of the maintenence tool. its part of the library for installing and updating
* obtrusive popups on almost every launch is very annoying. Most "modern" applications like chrome/firefox have moved to a background update on close because its a better user experience.

adityaruplaha:
When is this getting merged ? Looks quite ready.

src/citra_qt/main.ui
```diff
@@ -207,4 +223,5 @@
   </action>
  </widget>
  <resources/>
+ <connections/>
```
B3n30: NIT: unintentional change, but I think it can stay

src/citra_qt/updater/updater.h
```diff
@@ -0,0 +1,120 @@
...
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice, this
+// list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
```
B3n30: I don't think we have this anywhere in our license file, but we should

src/citra_qt/updater/updater.h
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    /// Default constructor
+    explicit Updater(QObject* parent = nullptr);
+    /// Constructor with an explicitly set path
+    explicit Updater(const QString& maintenance_tool_path, QObject* parent = nullptr);
+    /// Destroys the updater and kills the update check (if running)
+    ~Updater();
+
+    /// Returns `true`, if the updater exited normally
```
B3n30: The docstrings for functions in citra are usually
```
/**
 * returns true if the updater exited normally 
 */
```

Some for the other public functions

src/citra_qt/updater/updater.h
```diff
@@ -0,0 +1,120 @@
...
+    bool ExitedNormally() const;
+    /// Returns the mainetancetools error code of the last update
+    int ErrorCode() const;
+    /// returns the error output (stderr) of the last update
+    QByteArray ErrorLog() const;
+
+    /// readAcFn{Updater::running}
```
B3n30: What is readAcFn? Same for LatestUpdateInfo, RunningChanged and UpdateInfoChanged

src/citra_qt/updater/updater.h
j-selby: Not sure how you think we should handle this.

wwylele:
As reported, this fails the frozen dependencies build because of the usage of Qt 5.4 overload of `QTimer::singleShot` (and probably more). Our frozen dependencies uses Qt 5.2 (= current minimal requirement to compile citra), so please consider using functions in this version, otherwise we need to bump the version requirement and document it.

j-selby:
Fair enough. I don't have the skills to do that kind of backporting (nor the time to fuss over the million other things wrong with this), so I will likely close this PR shortly.

wwylele:
@j-selby hey I am not rejecting this. I just skimmed over the code, and found that it is pretty easy to fix `QTimer::singleShot` issue: you just need to make what is a lambda now an actual slot/member function and use the old version of `singleShot` to call it.

jroweboy:
I fixed the qtimer issue and conflicts and manually pushed to master

PR #3001, opened 2017-10-08
citra-qt : Adding fullscreen mode

Styleoshin:
I  added mode fullscreen for play with more comfort.
For display fullscreen,  just click on action bar or the shortcut : alt+return

For exit fullscreen just click in alt+return or escape

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3001)

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -244,6 +244,8 @@ void GMainWindow::InitializeHotkeys() {
     RegisterHotkey("Main Window", "Load File", QKeySequence::Open);
     RegisterHotkey("Main Window", "Swap Screens", QKeySequence::NextChild);
     RegisterHotkey("Main Window", "Start Emulation");
+    RegisterHotkey("Main Window", "Fullscreen", QKeySequence::FullScreen);
+    RegisterHotkey("Main Window", "Exit Fullscreen", QKeySequence::Cancel, Qt::ApplicationShortcut);
     LoadHotkeys();
 
     connect(GetHotkey("Main Window", "Load File", this), SIGNAL(activated()), this,
@@ -252,6 +254,16 @@ void GMainWindow::InitializeHotkeys() {
             SLOT(OnStartGame()));
     connect(GetHotkey("Main Window", "Swap Screens", render_window), SIGNAL(activated()), this,
             SLOT(OnSwapScreens()));
+    connect(GetHotkey("Main Window", "Fullscreen", render_window), &QShortcut::activated,
+            ui.action_Fullscreen, &QAction::trigger);
+    connect(GetHotkey("Main Window", "Fullscreen", render_window), &QShortcut::activatedAmbiguously,
+            ui.action_Fullscreen, &QAction::trigger);
+    connect(GetHotkey("Main Window", "Exit Fullscreen", this), &QShortcut::activated, this, [&] {
+        if (emulation_running) {
+            ui.action_Fullscreen->setChecked(false);
+            ToggleFullscreen();
+        }
+    });
 }
 
 void GMainWindow::SetDefaultUIGeometry() {
@@ -279,6 +291,8 @@ void GMainWindow::RestoreUIState() {
 
     ui.action_Single_Window_Mode->setChecked(UISettings::values.single_window_mode);
     ToggleWindowMode();
+
+    ui.action_Fullscreen->setChecked(UISettings::values.fullscreen);
 
     ui.action_Display_Dock_Widget_Headers->setChecked(UISettings::values.display_titlebar);
     OnDisplayTitleBars(ui.action_Display_Dock_Widget_Headers->isChecked());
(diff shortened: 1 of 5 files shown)


src/citra_qt/main.ui
```diff
@@ -45,7 +45,7 @@
      <x>0</x>
      <y>0</y>
      <width>1081</width>
-     <height>19</height>
+     <height>21</height>
```
j-selby: Not sure why Qt Designer keeps adding this. Unintentional change?

acnleditor2:
Submodule citra?

src/citra_qt/main.cpp
```diff
@@ -824,6 +840,17 @@ void GMainWindow::dragMoveEvent(QDragMoveEvent* event) {
     event->acceptProposedAction();
 }
 
+void GMainWindow::keyPressEvent(QKeyEvent* event) {
+
+    if (event->key() == Qt::Key_Return) {
+        if (event->modifiers() && Qt::AltModifier) {
```
j-selby: This isn't doing what you think it is. `&` is the bitwise operator (which is likely what you are looking for).

Example fix:
`event->modifiers() & Qt::AltModifier != 0`

src/citra_qt/main.ui
```diff
@@ -196,6 +197,18 @@
...
+   <property name="text">
+    <string>Fullscreen</string>
+   </property>
+   <property name="shortcut">
+    <string>Alt+Return</string>
+   </property>
+  </action>
  </widget>
  <resources/>
+ <connections/>
```
j-selby: Is this needed?

src/citra_qt/bootmanager.cpp
```diff
@@ -208,6 +208,15 @@ void GRenderWindow::closeEvent(QCloseEvent* event) {
 
 void GRenderWindow::keyPressEvent(QKeyEvent* event) {
     InputCommon::GetKeyboard()->PressKey(event->key());
+
+    // Exit fullscreen in GMainWindow
+    if (event->key() == Qt::Key_Return) {
+        if (event->modifiers() && Qt::AltModifier) {
```
j-selby: See below comment on this.

j-selby:
"canno't enter in fullscreen if single window mode is not checked"

This shouldn't be a hidden note in this PR. Is there any reason why this is not the case?

src/citra_qt/main.cpp
```diff
@@ -624,6 +626,20 @@ void GMainWindow::OnStopGame() {
     ShutdownGame();
 }
 
+void GMainWindow::OnDisplayFullscreen() {
+    if (ui.action_Single_Window_Mode->isChecked()) {
```
lioncash: It's possible to unindent this by comparing in terms of the early-exit case first.

```cpp
void GMainWindow::OnDisplayFullscreen() {
    if (!ui.action_Single_Window_Mode->isChecked()) {
        return;
    }

    ui.menubar->hide();
    statusBar()->hide();
    showFullScreen();
}
```

src/citra_qt/main.ui
Styleoshin: Yes, Qt Designer addins this

Styleoshin:
Now it's possible to play fullscreen on single or separate window

jroweboy:
@Styleoshin i went ahead and added some more fixes to my branch. This will register the Fullscreen hotkey in the menu as well, so that way people will know what the shortcut is. Its also registered with the hotkey system in citra, so people can change the hotkey through the settings if they want. https://github.com/jroweboy/citra/tree/fullscreen

Styleoshin:
ok, i see, that's a good idea :)

jroweboy:
@Styleoshin go ahead and add what I wrote to your PR however you want. I just didn't want to have a back and forth about "Thats not quite what I meant" and just wrote out what I was expecting. Also when I sat down to work on this, I didn't realize how tricky having fullscreen as a menu action made it :P

Styleoshin:
You're right, it's better

jroweboy:
Last call for reviews before I merge it in 24 hours

FearlessTobi:
In builds with this PR merged, I get an annoying bug. When I am on the Game List and my Citra window is maximized and I now start a game, my Citra window will minimize. Here´s a GIF for better explanation: 

ezgif com-video-to-gif

jroweboy:
Ah this is a regression because `GMainWindow::ToggleFullscreen` doesn't track what state it was in (Maximized, Minimized, etc) and just returns the window to "Normal" if the window isn't fullscreened. The fix shouldn't be that hard to do if anyone has some free time :)

RavenHome1:
actually while we are on the topic of full screen mode

-the status bar (the thing that shows speed % and frame rate) completely vanishes and it would be nice if you can maybe move it to the corner so it can keep displaying these value.

-using full screen mode causes a lot of tearing to happen, not sure if  something that can't be helped without V-sync since i have it off but maybe look into it

Yogarpg:
Why my emulation background is white, even in fullscreen?
Black is so much better
captura de tela 267

jroweboy:
@Yogarpg this was already change from white to black back in june https://github.com/citra-emu/citra/pull/2795 the background color is a customizable setting in the config ini. Please ask on the forums for more help if you need it, thanks!

PR #3004, opened 2017-10-10
citra-qt: add continue/pause hotkey

acnleditor2:
the key is f4

This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3004)


acnleditor2:
video https://www.youtube.com/watch?v=cQ7sZa43Zh4

src/citra_qt/main.cpp
```diff
@@ -824,6 +824,12 @@ void GMainWindow::dragMoveEvent(QDragMoveEvent* event) {
     event->acceptProposedAction();
 }
 
+void GMainWindow::keyPressEvent(QKeyEvent* event) {
+if (event->key() == Qt::Key_F4 && emulation_running) {
```
bunnei: This key should be user configurable, right?

acnleditor2: I do not know how to make it configurable...

j-selby: This should be done using the pre-existing hotkey infrastructure - https://github.com/acnleditor3/citra/blob/d56f908e5cc95e9d1cd7e3e52503da3264a76b4a/src/citra_qt/main.cpp#L249

e.g. 
```cpp
connect(GetHotkey("Main Window", "Play/Pause", this), SIGNAL(activated()), this,
            SLOT(OnRunningToggle()));

[...]

void GMainWindow::OnRunningToggle() {
    if (emulation_running) {
        [...]
    }
}
``` 

And add a binding for that in the class definition, here:

https://github.com/acnleditor3/citra/blob/d56f908e5cc95e9d1cd7e3e52503da3264a76b4a/src/citra_qt/main.h#L115

e.g.

```cpp
void OnRunningToggle();
```

acnleditor2: I do not understand...

j-selby: That is pretty much all you need to know - what are you confused about?

acnleditor2: I'll try
I do not know much about c ++

src/citra_qt/main.cpp
```diff
@@ -252,6 +252,14 @@ void GMainWindow::InitializeHotkeys() {
             SLOT(OnStartGame()));
     connect(GetHotkey("Main Window", "Swap Screens", render_window), SIGNAL(activated()), this,
             SLOT(OnSwapScreens()));
+    connect(GetHotkey("Main Window", "Play/Pause", this), SIGNAL(activated()), this,
+            SLOT(OnRunningToggle()));
+}
+
+void GMainWindow::OnRunningToggle() {
+    if (emulation_running) {
+        OnPauseGame();
```
j-selby: I don't like the idea of calling a slot as part of another method directly. The contents of the `OnPauseGame` method should be refactored out into `PauseGame` or similar.

src/citra_qt/main.cpp
```diff
@@ -252,6 +252,14 @@ void GMainWindow::InitializeHotkeys() {
...
+    }
```
j-selby: What if emulation is *not* running? It is not possible to resume the game?

src/citra_qt/main.cpp
```diff
@@ -823,7 +831,7 @@ void GMainWindow::dragEnterEvent(QDragEnterEvent* event) {
 void GMainWindow::dragMoveEvent(QDragMoveEvent* event) {
     event->acceptProposedAction();
 }
-
+
```
j-selby: This shouldn't be here.

src/citra_qt/main.ui
```diff
@@ -124,6 +124,9 @@
    <property name="text">
     <string>&amp;Start</string>
    </property>
+     <property name="shortcut">
```
j-selby: These are incorrectly indented, and do not reflect the actual hotkey pressed (if it was to be re-bound).

acnleditor2:
@j-selby I'm new to C ++, I do not know how to do that :(

Dragios:
Then learn how to do it the proper way rather than complaining that your c++ knowledge is limited.
If that is the case this, then no point in opening this PR in the 1st place.

yuriks:
@Dragios Please refrain from making gratuitously abrasive comments that don't really add to the conversation. :/

jroweboy:
Thank you for the contribution. As @j-selby mentioned though, this doesn't use the existing hotkey framework that citra already has, and as such it won't get merged until that is fixed. For an example of how the hotkeys work, look at the existing hotkeys in https://github.com/citra-emu/citra/blob/f0027e9996866a612b35c6cfcd1f01b2eda220ac/src/citra_qt/main.cpp#L254 Please open a new PR when you get this worked out. Feel free to ask for help on IRC or discord as always. Closing this one in the meantime

PR #3007, opened 2017-10-15
Install Linux icon in hicolor instead of pixmaps

KAMiKAZOW:
hicolor is the preferred location for applications. See https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#directory_layout

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3007)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,7 +317,7 @@ if(ENABLE_QT AND UNIX AND NOT APPLE)
     install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.desktop"
             DESTINATION "${CMAKE_INSTALL_PREFIX}/share/applications")
     install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.svg"
-            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/pixmaps")
+            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/apps")
     install(FILES "${CMAKE_SOURCE_DIR}/dist/citra.xml"
             DESTINATION "${CMAKE_INSTALL_PREFIX}/share/mime/packages")
 endif()


j-selby:
I disagree. I read that as "hicolor" should always exist, not that applications should specifically support it. Citra does not specifically support any particular theme, so I don't believe this is correct. Furthermore, not sure where you are getting `/usr/share/icons` from - I don't see that there. The theme itself should supply a Citra icon if it wants to theme it specifically - we shouldn't be dealing with it ourself.

`By default, apps should look in $HOME/.icons (for backwards compatibility), in $XDG_DATA_DIRS/icons and in /usr/share/pixmaps (in that order). Applications may further add their own icon directories to this list, and users may extend or change the list (in application/desktop specific ways)`.

KAMiKAZOW:
I'm a bit confused. You don't have /usr/share/icons on your systems? What? Sure you're using Linux? That is the canonical location to store icon themes. "hicolor" is not an icon theme in itself. Those would be KDE's Breeze or Gnome's Adwaita. Linux desktops need to look in hicolor for 3rd party application icons. pixmaps is the legacy icon location. It's legacy because it does only support placing a single resolution icon – which brings me to my next argument:

Using hicolor has the added advantage that multiple resolutions can be placed there. Currently there is only the SVG icon but especially tweaked low res icons for application launcher menus would now be possible as well – e.g. /usr/share/icons/hicolor/16x16/apps/citra.png.

Btw, for reference I'm a somewhat experienced Linux packager (various distributions), so I like to believe that I know a thing or two about file locations under Linux. Feel free to look some of my work up on e.g. https://build.opensuse.org/project/show/home:KAMiKAZOW:Emulators

j-selby:
I most certainly do have `/usr/share/cions/hiccolor`, it just seems really weird to be dropping files in such a specific path, though if that is the way to go, sure.

jroweboy:
@linkmauve if you have a second to weigh in on this, I would appreciate your input :)

KAMiKAZOW:
@j-selby What's more specific about /usr/share/icons than /usr/share/pixmaps? You may also probably want to have a look at https://github.com/dolphin-emu/dolphin/blob/master/CMakeLists.txt#L800 how another project handles the same task.

j-selby:
@KAMiKAZOW `/usr/share/icons/hicolor/scalable/apps` is a hell of a lot more specific then `/usr/share/pixmaps`, which was my argument there.

PR #3011, opened 2017-10-17
citra-qt: Add New 3DS Settings

acnleditor2:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3011)


archshift:
Why add a functionless toggle to the settings?

acnleditor2:
can be useful later

jroweboy:
Hello! I really appreciate your effort and this contribution, but this change will only make users more confused about why when they turn on the setting their game still thinks its a old 3ds. The is_new_3ds setting was intentionally added to the ini file only (and not the UI) because it doesn't actually change anything. There are a few services in PTM that will use the variable, but it turns out that most games just check config info for it, and since the config info is dumped from your 3ds devices, we don't use the setting to overwrite the config block.

As such, the setting is useless and shouldn't be presented to users.

PR #3020, opened 2017-10-20
[Qt] Add an animation for the menubar

Styleoshin:
Hello, I added an opacity animation feature in the menu bar.  When the menu bar is active and the cursor isn't at the top (but in the application), a timer will be triggered after 3 secsto hide the menu. If the cursor is at the top, the menu bar will be displayed.
For display in :
Windows :  the cursor will be at top of centralWidget or in title bar
Mac OS :  the cursor will be at top of centralWidget or ?
Linux :  the cursor will be at top of centralWidget. 

I can use mouseMoveEvent but the tracking will be activated only in the centralWidget.

[Animation menubar.zip](https://github.com/citra-emu/citra/files/1403416/Animation.menubar.zip)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3020)

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -93,12 +93,14 @@ GMainWindow::GMainWindow() : config(new Config()), emu_thread(nullptr) {
     Pica::g_debug_context = Pica::DebugContext::Construct();
     setAcceptDrops(true);
     ui.setupUi(this);
+    installEventFilter(this);
     statusBar()->hide();
 
     InitializeWidgets();
     InitializeDebugWidgets();
     InitializeRecentFileMenuActions();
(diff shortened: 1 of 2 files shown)


jroweboy:
Two major concerns right now:

* This animation is happening outside of fullscreen mode?
* Its way way too loong

Hiding the menu bar goes against most every style guide out there. When I originally suggested it, I was proposing that it would only be used to allow people in fullscreen mode to bring up the menu bar while still in fullscreen mode.

jroweboy:
Closing due to the concerns I mentioned in the previous comment. If you would like to revive this, let me know on IRC or discord and we can discuss what a good design for this should be. Thanks again for your contributions!

PR #3026, opened 2017-10-21
Cached sequential access of GetPhysicalPointer

huwpascoe:
Often memory is accessed in sequence, checking the
previous area before searching other areas reduces
the cost of lookup.

Tested running Pokemon Moon with
Visual Studio 2017 Profiler (100,000~ samples):

BaseLine:
Memory::GetPhysicalPointer    1.8% CPU samples (exclusive)
SVC::CallSVC                  56.71% CPU samples (inclusive)

With AreaCache:
Memory::GetPhysicalPointer    0.4% CPU samples (exclusive)
SVC::CallSVC                  55.01% CPU samples (inclusive)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3026)

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -24,6 +24,39 @@ namespace Memory {
 static std::array<u8, Memory::VRAM_SIZE> vram;
 static std::array<u8, Memory::N3DS_EXTRA_RAM_SIZE> n3ds_extra_ram;
 
+class AreaCache {
+public:
+    void CacheNewArea(u8* p_ptr, u32 base, u32 start, u32 size) {
+        physical_base = p_ptr;
+        address_base = base;
+        region_start = start;
+        region_end = start + size;
+    }
+
+    void CacheNewArea(u8* p_ptr, u32 base, u32 size) {
+        CacheNewArea(p_ptr, base, base, size);
+    }
+
+    u8* GetCachedAreaPointer(PAddr address) {
+        if (address < region_end && address >= region_start) {
+            return physical_base + (address - address_base);
+        }
+        Invalidate();
+        return nullptr;
+    }
+
+    void Invalidate() {
+        region_start = region_end = 0;
+    }
+
+private:
+    u8* physical_base;
+    PAddr address_base = 0;
+    PAddr region_start = 0;
+    PAddr region_end = 0;
+};
+
+static AreaCache area_cache;
 static PageTable* current_page_table = nullptr;
 
 void SetCurrentPageTable(PageTable* page_table) {
@@ -40,6 +73,8 @@ PageTable* GetCurrentPageTable() {
 static void MapPages(PageTable& page_table, u32 base, u32 size, u8* memory, PageType type) {
     LOG_DEBUG(HW_Memory, "Mapping %p onto %08X-%08X", memory, base * PAGE_SIZE,
               (base + size) * PAGE_SIZE);
+
+    area_cache.Invalidate();
 
     RasterizerFlushVirtualRegion(base << PAGE_BITS, size * PAGE_SIZE,
                                  FlushMode::FlushAndInvalidate);
@@ -275,6 +310,9 @@ std::string ReadCString(VAddr vaddr, std::size_t max_length) {
 }
 
 u8* GetPhysicalPointer(PAddr address) {
+    if (u8* cached_ptr = area_cache.GetCachedAreaPointer(address))
+        return cached_ptr;
+
     struct MemoryArea {
         PAddr paddr_base;
         u32 size;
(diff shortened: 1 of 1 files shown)


src/core/memory.cpp
```diff
@@ -274,7 +274,37 @@ std::string ReadCString(VAddr vaddr, std::size_t max_length) {
...
 }
 
+class {
+private:
+    PAddr previous_address;
+    u8* cached_target;
+    PAddr start = 0, end = 0;
+
+public:
+    inline void NewChain(u8* target, u32 area_base, u32 area_size) {
```
yuriks: `inline` keywords are redundant if you're already defining a function inline in the class declaration.

yuriks: Why is this called a "chain"? It allows non-sequential accesses, so the name doesn't seem to fit. "CacheNewArea"/"GetCachedAreaPointer"?

src/core/memory.cpp
```diff
@@ -274,7 +274,37 @@ std::string ReadCString(VAddr vaddr, std::size_t max_length) {
...
+        cached_target = target;
+        start = area_base;
+        end = area_base + area_size;
+    }
+
+    inline u8* NextInChain(PAddr address) {
+        const int diff = (address - previous_address);
+        previous_address = address;
```
yuriks: What is the point of keeping track of this? You could just store the base address of the area and offset into that.

I think the code will also be shorter with some extra integration into GetPhysicalPointer: If you separate the determination of the area (in the switch) from offsetting into it, then you can store the area information and keep the offsetting code common.

src/core/memory.cpp
```diff
@@ -274,7 +274,37 @@ std::string ReadCString(VAddr vaddr, std::size_t max_length) {
     return string;
 }
 
+class {
```
yuriks: Give this class a name.

src/core/memory.cpp
```diff
@@ -274,7 +274,37 @@ std::string ReadCString(VAddr vaddr, std::size_t max_length) {
...
+
+        if (address >= start && address < end) {
+            cached_target += diff;
+            return cached_target;
+        }
+
+        return nullptr;
+    }
+
+} AreaCache;
```
yuriks: (Global/File-level) variables should use foo_bar naming convention.

yuriks:
Also forgot to mention: Add a comment explaining the what and why of the optimization.

Echoing some questions from IRC: What effect does this have on frametimes in games, etc.?

src/core/memory.cpp
huwpascoe: Oh yeah...deep. I'll fix this tomorrow

src/core/memory.cpp
```diff
@@ -274,7 +274,37 @@ std::string ReadCString(VAddr vaddr, std::size_t max_length) {
     return string;
 }
 
+class {
+private:
```
lioncash: Public interface should go above the private interface.

src/core/memory.cpp
```diff
@@ -274,7 +274,37 @@ std::string ReadCString(VAddr vaddr, std::size_t max_length) {
...
+    PAddr previous_address;
+    u8* cached_target;
+    PAddr start = 0, end = 0;
```
lioncash: Please put variables on their own line.

src/core/memory.cpp
lioncash: This should also be a `static` variable and not bound to the class body itself.

src/core/memory.cpp
```diff
@@ -274,7 +274,37 @@ std::string ReadCString(VAddr vaddr, std::size_t max_length) {
...
+
+public:
+    inline void NewChain(u8* target, u32 area_base, u32 area_size) {
+        cached_target = target;
+        start = area_base;
+        end = area_base + area_size;
+    }
+
+    inline u8* NextInChain(PAddr address) {
+        const int diff = (address - previous_address);
```
lioncash: parentheses aren't necessary here.

acnleditor2:
some games cause citra to crash when starting / after loading

tabnk:
Issues: Mario Kart and Mario 3D Land failed to launch

huwpascoe:
@acnleditor2 @tabnk Hopefully it should work now.

src/core/memory.cpp
```diff
@@ -24,6 +24,39 @@ namespace Memory {
 static std::array<u8, Memory::VRAM_SIZE> vram;
 static std::array<u8, Memory::N3DS_EXTRA_RAM_SIZE> n3ds_extra_ram;
 
+class AreaCache {
+private:
```
lioncash: My comment about public interface above private interface still applies here.  Private vars below the public interface will still work correctly here.

huwpascoe: Oh, woops missed that one. I'll change it tomorrow

acnleditor2:
nintendogs + cats crash

acnleditor2:
pokemon sun & moon crash on opening qr reader

Subv:
I don't like the idea of adding _yet another cache and static state_ to the emulator for something that will probably not benefit speed much if at all in the grand scheme of things. This imo falls in the realm of microoptimization, there are bigger targets out there.

B3n30:
Closed because there is reportedly no improvements out of this PR and it just adds complexity to the code

tabnk:
Notice Super Mario 3D Land slower after this PR was removed from canary build so I’m decided test some Mario games.

Master Build on Haswell CPU
Super Mario 3D = 64 fps 
Mario Kart - 68 fps 

This PR on Haswell CPU
Super Mario 3D = 67 fps 
Mario Kart - 72 fps 

Overall still a nice performance improvement.

Issue #3023, opened 2017-10-21
QKeySequence::Cancel requires Qt 5.6


wwylele:
[This](https://github.com/citra-emu/citra/blob/master/src/citra_qt/main.cpp#L248) was introduced recently by the fullscreen feature. [The version requirement of this enum value is undocumented.](https://bugreports.qt.io/browse/QTBUG-54189)

This version requirement might be too high for some linux distros, where the shipped version are still around 5.2~5.5. Our CI, running Ubuntu 16.04, also uses Qt 5.5 (See #3022 for why it didn't report the error in the PR)

We can consider using other `QKeySequence` for this. Otherwise we need to explicitly note the version requirement in the building guide and upgrade the CI

jroweboy:
Should be easy to just hardcode a value for escape instead of using the key sequence for it

PR #3027, opened 2017-10-21
travis: Add build with frozen versions for dependencies

MerryMage:
Related to #3023.

Adds a build which pulls specific versions of packages so we can ensure new code is compatible with those specific library versions. Unfortunately Ubuntu's apt repositories remove older versions regularly, so we can't use apt-get's built-in version specifiers and have to pull the packages directly from Ubuntu's archives.

Justification for each version:

* Qt 5.2: 5.2 is the newest version available in repositories of some Linux distributions
* SDL 2.0.4: No particular reason
* libicu52: Required for pulling Qt 5.2 because the Xenial repositories no longer have this package

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3027)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,12 @@ matrix:
       install: "./.travis/macos/deps.sh"
       script: "./.travis/macos/build.sh"
       after_success: "./.travis/macos/upload.sh"
+    - os: linux
+      env: NAME="linux build (frozen versions of dependencies)"
+      sudo: required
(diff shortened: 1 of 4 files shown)


MerryMage:
> Shouldn't this replace the other linux build?

The reason why I didn't remove the other build was so that we don't miss something that's in a compilation failure in a more recent version, either.

PR #3029, opened 2017-10-22
Services/AM: Add CIA title installation support.

shinyquagsire23:
This PR can be reviewed commit-by-commit.

This PR adds support for am:u and am:net CIA information retrieval and installation commands. It does *not* add any Qt UI methods for installation, that will probably get handled in its own PR. The CIAContainer class is designed to allow for CIAs to be loaded from a filepath, FileBackend, or buffered CIA data, or for instances where more control is needed, CIAContainer can load portions of data as it is retrieved.

This PR was tested with GodMode9-dumped CIAs and homebrew CIAs using https://github.com/Steveice10/FBI/releases/tag/2.4.2. This can be run by extracting the main CXI from the CIA using `ctrtool -x --contents=fbi FBI.cia`.

A video demonstration of this PR can be found [here](https://www.youtube.com/watch?v=yQp-sPKfho4).

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3029)


src/core/file_sys/title_metadata.cpp
```diff
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <cinttypes>
+#include <streambuf>
```
mailwl: not alphabetical order of include

Jhon591:
@acnleditor2
http://www.boost.org/users/history/version_1_65_1.html

You could fork and update you're own branch, with files from the .7zip archive

also not what is in the array.hpp file open with notepad, can lead to other files needed, for included, make note.

Jhon591:
edit: did a local compile, compiled fine anyway without the latest boost full.

src/core/hle/service/am/am.cpp
```diff
@@ -57,7 +76,119 @@ struct TicketInfo {
...
+                cia_installing.Print();
+                cia_installing_header_loaded = true;
+            }
+        }
+
+        if (cia_installing_header_loaded) {
+            if (offset < cia_installing.GetContentOffset()) {
+                cia_installing_data.resize(
+                    std::min(offset + length, cia_installing.GetContentOffset()));
+                memcpy(cia_installing_data.data() + offset, buffer,
```
Subv: will this actually work if you write the entire CIA in one single Write operation?
If the buffer contains the CIA header you'll be copying it to the CIAContainer header properly, but then you'll be copying it *again* to the cia_installing_data vector.
Shouldn't this block deal only with the remaining data after extracting the header?

src/core/hle/service/am/am.cpp
```diff
@@ -57,7 +76,119 @@ struct TicketInfo {
...
+class CIAFile : public FileSys::FileBackend {
+public:
+    explicit CIAFile(Service::FS::MediaType media_type) : media_type(media_type) {}
+
+    ResultVal<size_t> Read(u64 offset, size_t length, u8* buffer) const override {
+        LOG_WARNING(Service_AM, "(STUBBED) ReadCIA");
+        return MakeResult<size_t>(length);
+    }
+
+    ResultVal<size_t> Write(u64 offset, size_t length, bool flush,
```
Subv: I didn't finish reviewing this function, it is rather large and very complex, is there no way we can alleviate some of that complexity by breaking it into smaller functions?

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+    if (cia_is_installing) {
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(ResultCode(ErrCodes::CIACurrentlyInstalling, ErrorModule::AM,
+                           ErrorSummary::InvalidState, ErrorLevel::Permanent));
+        return;
+    }
+
+    // Create our CIAFile handle for the app to write to, and while the app writes
+    // Citra will store contents out to sdmc/nand
+    auto file = std::shared_ptr<Service::FS::File>(
```
Subv: std::make_shared(std::make_unique())

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+        FileSys::TitleMetadata old_tmd;
+        FileSys::TitleMetadata new_tmd = cia_installing.GetTitleMetadata();
+
+        std::string old_tmd_path =
+            GetTitleMetadataPath(cia_installing_media_type, new_tmd.GetTitleID(), false);
+        std::string new_tmd_path =
+            GetTitleMetadataPath(cia_installing_media_type, new_tmd.GetTitleID(), true);
+        old_tmd.LoadFromFile(old_tmd_path);
+
+        for (u16 i = 0; i < old_tmd.GetContentCount(); i++) {
```
Subv: Please add a comment about what this loop is doing, and use more meaningful variable names than `i` and `j`

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+            bool abort = false;
+            for (u16 j = 0; j < new_tmd.GetContentCount(); j++) {
+                if (old_tmd.GetContentIDByIndex(i) == new_tmd.GetContentIDByIndex(j))
```
Subv: What if the content changed but the id did not change?

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+            FileUtil::Delete(
+                GetTitleContentPath(cia_installing_media_type, old_tmd.GetTitleID(), i));
+        }
+
+        FileUtil::Delete(old_tmd_path);
+    }
+    ScanForAllTitles();
+
+    cia_is_installing = false;
+    Kernel::g_handle_table.Close(cia_handle);
```
Subv: This is not necessary

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+            rb.PushRaw<TitleInfo>(title_info);
+            return;
+        }
+
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(ResultCode(ErrCodes::InvalidCIAHeader, ErrorModule::AM,
+                           ErrorSummary::InvalidArgument, ErrorLevel::Permanent));
+        return;
+    } else {
+        // TODO(shinyquagsire23): LLE file handling?
```
Subv: Add an ASSERT(false) or UNIMPLEMENTED here, along with a log.

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+    Kernel::SharedPtr<Kernel::ServerSession> server = file_session->parent->server;
+    if (server == nullptr) {
+        LOG_WARNING(Service_AM, "File handle ServerSession disconnected!");
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(Kernel::ERR_SESSION_CLOSED_BY_REMOTE);
+        return;
+    }
+
+    if (server->hle_handler != nullptr) {
+        auto file = std::dynamic_pointer_cast<Service::FS::File>(server->hle_handler);
```
Subv: you should check if file is not nullptr, since you're doing a dynamic_cast.

As an aside, this may break horribly if you pass a handle that is not a File handle to this function.

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+
+        FileSys::CIAContainer container;
+        if (container.LoadFromFileBackend(*file->backend) == Loader::ResultStatus::Success) {
```
Subv: IMO, invert the condition and handle the error in the innermost indentation level, instead of the opposite

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+
+    Kernel::SharedPtr<Kernel::ServerSession> server = file_session->parent->server;
+    if (server == nullptr) {
+        LOG_WARNING(Service_AM, "File handle ServerSession disconnected!");
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(Kernel::ERR_SESSION_CLOSED_BY_REMOTE);
+        return;
+    }
+
+    if (server->hle_handler != nullptr) {
```
Subv: IMO, invert the condition and handle the error in the innermost indentation level, instead of the opposite

src/core/hle/service/am/am.cpp
```diff
@@ -27,6 +35,17 @@ namespace AM {
 static bool lists_initialized = false;
 static std::array<std::vector<u64_le>, 3> am_title_list;
 
+// Current CIA install context
+static bool cia_is_installing = false;
```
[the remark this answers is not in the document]
Subv: If only one CIA can be installed at the same time, why don't you make these members of CIAFile and just keep a reference to the file you're installing in the global scope?

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+
+    Kernel::SharedPtr<Kernel::ServerSession> server = file_session->parent->server;
+    if (server == nullptr) {
+        LOG_WARNING(Service_AM, "File handle ServerSession disconnected!");
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(Kernel::ERR_SESSION_CLOSED_BY_REMOTE);
+        return;
+    }
+
+    if (server->hle_handler != nullptr) {
```
Subv: Always try to keep the normal function flow in the least possible indentation level.

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+        auto file = std::dynamic_pointer_cast<Service::FS::File>(server->hle_handler);
```
Subv: same as before, this may break horribly if the handle isn't a File handle

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+
+        FileSys::CIAContainer container;
+        if (container.LoadFromFileBackend(*file->backend) == Loader::ResultStatus::Success) {
+            std::vector<u8> temp(output_buffer_size);
+
+            //  Read from the Meta offset + 0x400 for the 0x36C0-large SMDH
+            auto read_result = file->backend->Read(container.GetMetadataOffset() + 0x400,
```
Subv: What is this 0x400?

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+
+    // We need our FileBackend to get info, so we've gotta step up the chain from
+    // Handle->ClientSession->ServerSession->Backend
+    auto file_session = Kernel::g_handle_table.Get<Kernel::ClientSession>(rp.PopHandle());
+
+    size_t output_buffer_size;
+    VAddr output_buffer = rp.PopMappedBuffer(&output_buffer_size);
+    output_buffer_size = std::min(
+        output_buffer_size,
+        static_cast<size_t>(0x36C0)); // TODO(shinyquagsire23): is this size defined anywhere?
```
Subv: This is sizeof(SMDH)

src/core/hle/service/am/am.cpp
```diff
@@ -464,6 +597,495 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+        rb.Push(ResultCode(ErrCodes::InvalidCIAHeader, ErrorModule::AM,
+                           ErrorSummary::InvalidArgument, ErrorLevel::Permanent));
+        return;
+    } else {
+        // TODO(shinyquagsire23): LLE file handling?
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(Kernel::ERR_NOT_IMPLEMENTED);
+        return;
+    }
+}
```
Subv: add a blank line here

Subv:
Github hid some of my comments because you moved some code around, but they are still valid and should be addressed.

acnleditor2:
@Jhon591 https://github.com/acnleditor2/citra/invitations

src/core/file_sys/cia_container.cpp
```diff
@@ -0,0 +1,212 @@
...
+    file.Seek(GetTitleMetadataOffset(), SEEK_SET);
+    if (!file.ReadBytes(tmd_data.data(), cia_header.tmd_size) != cia_header.tmd_size)
+        return Loader::ResultStatus::Error;
+
+    result = LoadTitleMetadata(tmd_data);
+    if (result != Loader::ResultStatus::Success)
+        return result;
+
+    // Load CIA Metadata
+    std::vector<u8> meta_data(sizeof(Metadata));
```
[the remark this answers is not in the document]
shinyquagsire23: Good catch

src/core/file_sys/cia_container.h
```diff
@@ -0,0 +1,90 @@
...
+    struct Header {
+        u32 header_size;
+        u16 type;
+        u16 version;
+        u32 cert_size;
+        u32 tik_size;
+        u32 tmd_size;
+        u32 meta_size;
+        u64 content_size;
+        u8 content_index[0x2000];
```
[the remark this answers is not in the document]
shinyquagsire23: Mayyyyybe? If I'm being honest I don't think it's significant other than there's 0x2000 indexes and a define would be a bit redundant. Though, I think it should be an std::array since that was the sentiment w/ TMD so I'll do that much at least.

src/core/hle/service/am/am.cpp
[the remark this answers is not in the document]
shinyquagsire23: I wanted to make them members of CIAFile, only thing stopping that is that FileBackend is NonCopyable, so all member variables are read-only. Kinda annoying having a big hunk of state stuff sitting around though so I'll try and describe things better.

src/core/hle/service/am/am.cpp
shinyquagsire23: I can try and do that I suppose, at the very least I suppose I can try and describe the process of things a bit better?

src/core/hle/service/am/am.cpp
shinyquagsire23: If `offset` is greater than the Content offset, then the data in `buffer` won't contain TMD data which is what's being buffered here.

src/core/hle/service/am/am.cpp
shinyquagsire23: I wasn't sure if there was a way to check if the cast was successful, good to know. AM breaks horribly on hardware when you pass invalid handles so... Citra's actually being a bit more gracious by handing back errors.

src/core/file_sys/title_metadata.cpp
shinyquagsire23: Yeah clang-format really wants it there for some reason so I'm not going to question it.

shinyquagsire23:
@Subv addressed all of your comments I think, ended up finding a bug in how I got update TMD paths so that's fixed now. I think I need to do some research on the CIA content index because it just occurred to me that CIAs can possibly install less content than is specified in the TMD (ie a single theme CIA, where the TMD will list *all* themes but the only content it has is the one theme), so that can change things I think.

src/core/hle/service/am/am.cpp
Subv: Keep in mind that checking the result of a dynamic_cast will only work if RTTI is enabled.

Subv:
Reviewed 1 of 3 files at r1, 2 of 4 files at r4, 1 of 5 files at r11, 5 of 5 files at r12.
Review status: all files reviewed at latest revision, 40 unresolved discussions, some commit checks failed.

---

*[src/core/file_sys/cia_container.cpp, line 18 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP23ck8O2qBmlZh7-r1-18:-Kx8swc0cVE9bprEd_OT:bb6ywmx) ([raw file](https://github.com/citra-emu/citra/blob/690d5a37f1cbbd3d8fcd4eae6607700b78098939/src/core/file_sys/cia_container.cpp#L18)):*

I don't know how to feel about these LoadFromX functions, we could use function overloading here and have a single Load method with different parameters
</blockquote></details>

LGTM

---

*[src/core/file_sys/cia_container.cpp, line 110 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP23ck8O2qBmlZh7-r1-110:-Kx8tIRhz_UeGCa4HNQx:bb6ywmx) ([raw file](https://github.com/citra-emu/citra/blob/690d5a37f1cbbd3d8fcd4eae6607700b78098939/src/core/file_sys/cia_container.cpp#L110)):*

the offset parameter of all of these functions should be size_t
</blockquote></details>

LGTM

---

*[src/core/file_sys/cia_container.cpp, line 140 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP23ck8O2qBmlZh7-r1-140:-Kx8tPHt08613HnUua0H:b32w0x) ([raw file](https://github.com/citra-emu/citra/blob/690d5a37f1cbbd3d8fcd4eae6607700b78098939/src/core/file_sys/cia_container.cpp#L140)):*

const method
</blockquote></details>

OK

---

*[src/core/file_sys/cia_container.cpp, line 145 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP23ck8O2qBmlZh7-r1-145:-Kx8tTqy81SWAhokkBa3:b32w0x) ([raw file](https://github.com/citra-emu/citra/blob/690d5a37f1cbbd3d8fcd4eae6607700b78098939/src/core/file_sys/cia_container.cpp#L145)):*

please define this 0x40 as a named constant somewhere
</blockquote></details>

OK

---

*[src/core/file_sys/cia_container.cpp, line 199 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP23ck8O2qBmlZh7-r1-199:-Kx8ul_WgGqMBjCmE1I5:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/690d5a37f1cbbd3d8fcd4eae6607700b78098939/src/core/file_sys/cia_container.cpp#L199)):*

similarly, all these logs use a 64-bit format specifier, but you're only printing the lower 32 bits (%08)?
</blockquote></details>

ok

---

*[src/core/file_sys/cia_container.cpp, line 181 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8uaCo9maPmsjmke8v:-Kx8uaCpcRpxAe0Yzmzj:br09jn5) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.cpp#L181)):*
> ```cpp
>         // is included in the CIA, so check the bit for this index and add if set.
>         // The bits in the content index are arranged w/ index 0 as the MSB, 7 as the LSB, etc.
>         if (cia_header.content_present[i >> 3] & 0x80 >> (i & 7))
> ```

Can you use Common::BitSet for this?

---

*[src/core/file_sys/cia_container.cpp, line 205 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8uexRl9kLoyIZDX-X:-Kx8uexRl9kLoyIZDX-Y:b-xhy64n) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.cpp#L205)):*
> ```cpp
> u64 CIAContainer::GetContentSize(u16 index) const {
>     // If the content doesn't exist in the CIA, it doesn't have a size.
>     if (!(cia_header.content_present[index >> 3] & 0x80 >> (index & 7)))
> ```

Can you use Common::BitSet for this?

---

*[src/core/file_sys/cia_container.h, line 36 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8v-waWwgwfg7sTZiI:-Kx8v-waWwgwfg7sTZiJ:bnv6tsw) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.h#L36)):*
> ```objc
> public:
>     // Load whole CIAs outright
>     Loader::ResultStatus Load(FileBackend& backend);
> ```

Can't these be const parameters?

---

*[src/core/file_sys/cia_container.h, line 41 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8v2OoKWb99pqVjwz9:-Kx8v2OoKWb99pqVjwzA:b-clqh1j) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.h#L41)):*
> ```objc
> 
>     // Load parts of CIAs (for CIAs streamed in)
>     Loader::ResultStatus LoadHeader(std::vector<u8>& header_data, size_t offset = 0);
> ```

The vectors can all be const

---

*[src/core/file_sys/cia_container.h, line 45 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8y6PLDQ1pREFH-5Hn:-Kx8y6PLDQ1pREFH-5Ho:bdp29kl) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.h#L45)):*
> ```objc
>     Loader::ResultStatus LoadMetadata(std::vector<u8>& meta_data, size_t offset = 0);
> 
>     TitleMetadata& GetTitleMetadata();
> ```

This can be const and return a const reference

---

*[src/core/file_sys/title_metadata.cpp, line 6 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanM8-r8-6:-Kx8vdnlLCNddi_0vPuw:b-m9c2sl) ([raw file](https://github.com/citra-emu/citra/blob/f0b140b3eef83ffca980a46633de38f1158b740a/src/core/file_sys/title_metadata.cpp#L6)):*
<details><summary><i>Previously, shinyquagsire23 (Max Thomas) wrote…</i></summary><blockquote>

Yeah clang-format really wants it there for some reason so I'm not going to question it.
</blockquote></details>

You don't need this include anymore

---

*[src/core/file_sys/title_metadata.cpp, line 79 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8vZ6N06tpMc7tTKsg:-Kx8vZ6OjgiT7SjdgXNL:b-ufqkhy) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/title_metadata.cpp#L79)):*
> ```cpp
>     if (file_data.size() - offset < expected_size) {
>         LOG_ERROR(Service_FS, "Malformed TMD, expected size 0x%zx, got 0x%zx!", expected_size,
>                   file_data.size() - offset);
> ```

Does this subtraction return size_t instead of int? You're using the size_t format specifier

---

*[src/core/file_sys/title_metadata.h, line 95 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8vjl34quRO6MBR97d:-Kx8vjl34quRO6MBR97e:brwxadg) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/title_metadata.h#L95)):*
> ```objc
> #pragma pack(pop)
> 
>     Loader::ResultStatus Load(std::string& file_path);
> ```

Make these ref parameters const

---

*[src/core/hle/service/am/am.cpp, line 39 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-39:-Kx8vpCLEyoNxAoeGf3e:b-orne0z) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L39)):*
<details><summary><i>Previously, shinyquagsire23 (Max Thomas) wrote…</i></summary><blockquote>

I wanted to make them members of CIAFile, only thing stopping that is that FileBackend is NonCopyable, so all member variables are read-only. Kinda annoying having a big hunk of state stuff sitting around though so I'll try and describe things better.
</blockquote></details>

I don't understand the problem here, why does NonCopyable prevent you from assigning member variables?

---

*[src/core/hle/service/am/am.cpp, line 86 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-86:-Kx8wGNWqx78dgpwT051:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L86)):*

what is this supposed to do? add an ASSERT(false) / UNIMPLEMENTED() here if unimplemented
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 89 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-89:-Kx8zSYEeHmsR4YHvICi:b-q7wz5o) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L89)):*
<details><summary><i>Previously, shinyquagsire23 (Max Thomas) wrote…</i></summary><blockquote>

I can try and do that I suppose, at the very least I suppose I can try and describe the process of things a bit better?
</blockquote></details>

it's much better now, but i still think it should be split in some functions

---

*[src/core/hle/service/am/am.cpp, line 95 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-95:-Kx8wLHYglyHoZE80ZxO:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L95)):*

constexpr static
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 100 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-100:-Kx8yCo-1-BTkbpNySmV:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L100)):*

use cia_header_size here instead of the magic number
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 109 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-109:-Kx8x2RmytKsRDO0L4A0:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L109)):*

break up this line using variables
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 111 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-111:-Kx8xkC2YeEWbpR7XLsk:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L111)):*
<details><summary><i>Previously, shinyquagsire23 (Max Thomas) wrote…</i></summary><blockquote>

If `offset` is greater than the Content offset, then the data in `buffer` won't contain TMD data which is what's being buffered here.

Edit: Oh, wait I totally misread what you said, yeah let me fix the unneeded copying bit.
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 656 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-656:-Kx9-5w8ayEPtI4ir6zD:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L656)):*

Please add a comment about what this loop is doing, and use more meaningful variable names than `i` and `j`
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 674 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-674:-Kx9-rEPgPHX6CV9-Lkf:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L674)):*

This is not necessary
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 699 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-699:-Kx9-vBfwz_CzgDOt4yv:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L699)):*

IMO, invert the condition and handle the error in the innermost indentation level, instead of the opposite
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 700 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-700:-Kx9-uYsYz3U0NP5IMGB:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L700)):*

Keep in mind that checking the result of a dynamic_cast will only work if RTTI is enabled.
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 727 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-727:-Kx9059TEEmiVSgxoqpC:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L727)):*

Add an ASSERT(false) or UNIMPLEMENTED here, along with a log.
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 744 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-744:-Kx90ABDALX8YP9SmErs:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L744)):*

This is sizeof(SMDH)
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 762 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-762:-Kx90C1HLCNRmp0HdAtu:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L762)):*

same as before, this may break horribly if the handle isn't a File handle
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 769 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-769:-Kx90Ds9KGRjwHnr0Q6R:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L769)):*

What is this 0x400?
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 105 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8wVD2TN0KQXlGQNF6:-Kx8wVD4SvWUi8nckA00:b-7j125d) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L105)):*
> ```cpp
>         // content sizes so it ends up becoming a problem of keeping track of how  much has been
>         // written and what we have been able to pick up.
>         if (cia_install_state < HeaderLoaded) {
> ```

Please don't use enums like this

---

*[src/core/hle/service/am/am.cpp, line 255 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8zd8wsZIxABh9Hgol:-Kx8zd8xLWGCnZvdwefB:b-pr5gp4) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L255)):*
> ```cpp
>     // For now, just scan for any .tmd files which exist, the smallest will be the
>     // base ID and the largest will be the (currently installing) update ID.
>     u32 base_id = 0xFFFFFFFF;
> ```

use an inline static constexpr named constant for this guard value

---

*[src/core/hle/service/am/am.cpp, line 760 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx9-ItuNbOEOz7ULxWB:-Kx9-ItuNbOEOz7ULxWC:beritdy) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L760)):*
> ```cpp
>     if (server->hle_handler != nullptr) {
>         auto file = std::dynamic_pointer_cast<Service::FS::File>(server->hle_handler);
>         if (file != nullptr)
> ```

add a comment and a log about this, silently failing is not a good behavior, note that this check will only work with RTTI enabled. Is there really no other way to do this? it may hang up the real system but crashing citra is not an acceptable behavior

---

*[src/core/hle/service/am/am.cpp, line 767 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx9-j3M4LhpzkJjYpJZ:-Kx9-j3M4LhpzkJjYpJ_:b-n9uz75) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L767)):*
> ```cpp
>         // Probably the best bet if someone is LLEing the fs service is to just have them LLE AM
>         // while they're at it, so not implemented.
>         return Kernel::ERR_NOT_IMPLEMENTED;
> ```

add a log here

---

*[src/core/hle/service/am/am.cpp, line 811 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx906MTLum_DXRFPqL-:-Kx906MU4Npv6WgQ1rdj:b-hy30lv) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L811)):*
> ```cpp
>                        ErrorLevel::Permanent));
>     return;
> }
> ```

add a blank line between each service function

---

*[src/core/hle/service/am/am.cpp, line 873 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx90KImUYVwCxbUgiec:-Kx90KImUYVwCxbUgied:bof5wq1) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L873)):*
> ```cpp
> 
>     FileSys::CIAContainer container;
>     if (container.Load(*file->backend) == Loader::ResultStatus::Success) {
> ```

Invert the condition so that the error case is higher up and the normal flow is unindented, same everywhere else, i'll stop pointing this out

---

*[src/core/hle/service/am/am.cpp, line 885 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx90QAGqBbcCJaENq82:-Kx90QAGqBbcCJaENq83:b-c08i55) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L885)):*
> ```cpp
>                        ErrorLevel::Permanent));
> }
> void GetTransferSizeFromCia(Service::Interface* self) {
> ```

add a blank line, i'll stop pointing this out

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3029)*

src/core/file_sys/cia_container.cpp
```diff
@@ -0,0 +1,228 @@
...
+#include "common/logging/log.h"
+#include "core/file_sys/cia_container.h"
+#include "core/file_sys/title_metadata.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// FileSys namespace
+
+namespace FileSys {
+
+static constexpr u32 CIA_SECTION_ALIGNMENT = 0x40;
```
lioncash: Note that `const` and `constexpr` at file-scope mean `static`/internal linkage by default in C++, so you don't need to write `static` explicitly in these cases.

e.g. In the compatibility section of the standard:

> **Change**: A name of file scope that is explicitly declared `const`, and not explicitly declared `extern`, has internal linkage, while in C it would have external linkage
> **Rationale**: Because const objects can be used as compile-time values in C++, this feature urges programmers to provide explicit initializer values for each `const`. This feature allows the user to put `const` objects in header files that are included in many compilation units.

src/core/file_sys/cia_container.h
```diff
@@ -0,0 +1,96 @@
...
+#include "core/file_sys/title_metadata.h"
+#include "core/hle/service/fs/archive.h"
+#include "core/loader/loader.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// FileSys namespace
+
+namespace FileSys {
+
+static constexpr size_t CIA_CONTENT_MAX_COUNT = 0x10000;
```
lioncash: Ditto about statics

src/core/file_sys/cia_container.h
```diff
@@ -0,0 +1,96 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <string>
+#include <vector>
+#include "common/common_types.h"
+#include "common/swap.h"
+#include "core/file_sys/file_backend.h"
```
lioncash: A few of these seem like they could be forward declarations.

src/core/hle/service/am/am.h
```diff
@@ -19,22 +19,43 @@ class Interface;
...
+namespace ErrCodes {
+enum {
+    CIACurrentlyInstalling = 4,
+    EmptyCIA = 32,
+    InvalidCIAHeader = 104,
+};
+} // namespace ErrCodes
+
+enum CIAInstallState : u32 {
+    NotInstalling = 0,
```
lioncash: The `= 0` is kind of superfluous here.

src/core/hle/service/am/am.h
```diff
@@ -19,22 +19,43 @@ class Interface;
...
 
+namespace ErrCodes {
+enum {
+    CIACurrentlyInstalling = 4,
+    EmptyCIA = 32,
+    InvalidCIAHeader = 104,
+};
+} // namespace ErrCodes
+
+enum CIAInstallState : u32 {
```
lioncash: enum class?

src/core/hle/service/am/am.cpp
```diff
@@ -20,13 +28,28 @@
...
+static CIAInstallState cia_install_state = NotInstalling;
+
+// Writing state variables: How much has been written total, CIAContainer for the installing CIA,
+// buffer of all data prior to content data, how much of each content index has been written, and
+// where is the CIA being installed to?
+static u64 cia_installing_written = 0;
+static FileSys::CIAContainer cia_installing;
+static std::vector<u8> cia_installing_data;
+static std::vector<u64> cia_installing_content_written;
+Service::FS::MediaType cia_installing_media_type;
```
lioncash: Should this be static as well?

src/core/hle/service/am/am.cpp
```diff
@@ -57,7 +80,168 @@ struct TicketInfo {
...
+        // TODO(shinyquagsire23): Can we assume that things will only be written in sequence?
+        // Does AM send an error if we write to things out of order?
+        // Or does it just ignore offsets and assume a set sequence of incoming data?
+
+        // The data in CIAs is always stored CIA Header > Cert > Ticket > TMD > Content > Meta.
+        // The CIA Header describes Cert, Ticket, TMD, total content sizes, and TMD is needed for
+        // content sizes so it ends up becoming a problem of keeping track of how  much has been
+        // written and what we have been able to pick up.
+        if (cia_install_state < HeaderLoaded) {
+            size_t buf_copy_size = std::min(length, static_cast<size_t>(FileSys::CIA_HEADER_SIZE));
```
lioncash: `<algorithm>` and `<cstddef>` should be included.

src/core/hle/service/am/am.cpp
```diff
@@ -57,7 +80,168 @@ struct TicketInfo {
...
+            size_t buf_max_size = std::min(offset + length, FileSys::CIA_HEADER_SIZE);
+            cia_installing_data.resize(buf_max_size);
+            memcpy(cia_installing_data.data() + offset, buffer, buf_copy_size);
```
lioncash: `<cstring>` should be included.

src/core/hle/service/am/am.cpp
```diff
@@ -461,7 +658,377 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+        return Kernel::ERR_SESSION_CLOSED_BY_REMOTE;
+    }
+
+    if (server->hle_handler != nullptr) {
+        auto file = std::dynamic_pointer_cast<Service::FS::File>(server->hle_handler);
+        if (file != nullptr)
+            return MakeResult<std::shared_ptr<Service::FS::File>>(file);
+
+        return Kernel::ERR_INVALID_HANDLE;
+    } else {
```
lioncash: This else block can be removed and the comment can be placed on the return outside of the if-else here. Otherwise the outermost return is effectively dead code.

src/core/hle/service/am/am.cpp
```diff
@@ -461,7 +658,377 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+
+        // For each content ID in the old TMD, check if there is a matching ID in the new
+        // TMD. If a CIA contains (and wrote to) an identical ID, it should be kept while
+        // IDs which only existed for the old TMD should be deleted.
+        for (u16 old_index = 0; old_index < old_tmd.GetContentCount(); old_index++) {
+            bool abort = false;
+            for (u16 new_index = 0; new_index < new_tmd.GetContentCount(); new_index++) {
+                if (old_tmd.GetContentIDByIndex(old_index) ==
+                    new_tmd.GetContentIDByIndex(new_index))
+                    abort = true;
```
lioncash: conditionals that have the condition portion travel more than one line should be braced.

src/core/hle/service/am/am.cpp
```diff
@@ -461,7 +658,377 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+        IPC::RequestBuilder rb = rp.MakeBuilder(8, 0);
+        rb.Push(RESULT_SUCCESS);
+        rb.PushRaw<TitleInfo>(title_info);
+        return;
+    }
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+    rb.Push(ResultCode(ErrCodes::InvalidCIAHeader, ErrorModule::AM, ErrorSummary::InvalidArgument,
+                       ErrorLevel::Permanent));
+    return;
```
lioncash: This return isn't necessary, since it's a void function (and it being top-level and at the end is redundant).

shinyquagsire23:
Review status: all files reviewed at latest revision, 49 unresolved discussions, some commit checks failed.

---

*[src/core/file_sys/cia_container.cpp, line 18 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP23ck8O2qBmlZh7-r12-18:-Kx9lcdQLlme6trgv7HE:bgzdldr) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.cpp#L18)):*

Note that `const` and `constexpr` at file-scope mean `static`/internal linkage by default in C++, so you don't need to write `static` explicitly in these cases.

e.g. In the compatibility section of the standard:

> **Change**: A name of file scope that is explicitly declared `const`, and not explicitly declared `extern`, has internal linkage, while in C it would have external linkage
> **Rationale**: Because const objects can be used as compile-time values in C++, this feature urges programmers to provide explicit initializer values for each `const`. This feature allows the user to put `const` objects in header files that are included in many compilation units.
</blockquote></details>

Good to know, thanks.

---

*[src/core/file_sys/cia_container.cpp, line 181 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8uaCo9maPmsjmke8v:-Kx9Hy0tT9Ke-UBKQxBe:bcbgf7g) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.cpp#L181)):*

Can you use Common::BitSet for this?
</blockquote></details>

Not directly, Common::BitSet only seems to support the least significant bit as index 0 and this uses the most significant bit as index 0, so I'd have to do some weird (7 - i) thing to check indexes.

---

*[src/core/file_sys/cia_container.cpp, line 205 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8uexRl9kLoyIZDX-X:-Kx9ly0Os7VO8CHreUKM:b-z3p5ge) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.cpp#L205)):*

Can you use Common::BitSet for this?
</blockquote></details>

ditto above

---

*[src/core/hle/service/am/am.cpp, line 659 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-659:-Kx9RxWG-pAmULrKbN6h:b-dchhea) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L659)):*

What if the content changed but the id did not change?
</blockquote></details>

In this case the old content ID will have been overwritten with new content anyhow, but it shouldn't be deleted if the same content ID is present in the new TMD.

---

*[src/core/hle/service/am/am.cpp, line 760 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx9-ItuNbOEOz7ULxWB:-Kx9PzWXq8PQnwkuzCYw:bbphu2p) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L760)):*

add a comment and a log about this, silently failing is not a good behavior, note that this check will only work with RTTI enabled. Is there really no other way to do this? it may hang up the real system but crashing citra is not an acceptable behavior
</blockquote></details>

The only other way I'm seeing is to implement service calls instead, which I guess would sort out LLE issues but I'm not sure what the logistics of doing that would entail. But it's totally an option if we really want to avoid RTTI.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3029)*

Subv:
Review status: 3 of 10 files reviewed at latest revision, 47 unresolved discussions.

---

*[src/core/file_sys/cia_container.h, line 36 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8v-waWwgwfg7sTZiI:-KxA7_4PUv8yT22wJaSJ:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.h#L36)):*

Can't these be const parameters?
</blockquote></details>

ok

---

*[src/core/file_sys/cia_container.h, line 41 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8v2OoKWb99pqVjwz9:-KxA7ZDMzJs-Qyfn440q:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.h#L41)):*

The vectors can all be const
</blockquote></details>

ok

---

*[src/core/file_sys/cia_container.h, line 45 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8y6PLDQ1pREFH-5Hn:-KxA7aKA07zWr48_7RxN:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/cia_container.h#L45)):*

This can be const and return a const reference
</blockquote></details>

ok

---

*[src/core/file_sys/title_metadata.cpp, line 61 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanM8-r2-61:-KxA7gZawCgdN5nlx7Jc:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/28b162d9c21f4912c41f2dab3a553e42fd3447a5/src/core/file_sys/title_metadata.cpp#L61)):*

why? just use std::memcpy as necessary
</blockquote></details>

ok

---

*[src/core/file_sys/title_metadata.cpp, line 6 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanM8-r8-6:-KxA7e5UAlM57OQKfvOH:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/f0b140b3eef83ffca980a46633de38f1158b740a/src/core/file_sys/title_metadata.cpp#L6)):*

You don't need this include anymore
</blockquote></details>

ok

---

*[src/core/file_sys/title_metadata.cpp, line 79 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8vZ6N06tpMc7tTKsg:-KxA7mb6XxgnwDKKkhOe:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/title_metadata.cpp#L79)):*

Does this subtraction return size_t instead of int? You're using the size_t format specifier
</blockquote></details>

ok

---

*[src/core/file_sys/title_metadata.h, line 95 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8vjl34quRO6MBR97d:-KxA7ptrfOm1sboHR6Ab:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/file_sys/title_metadata.h#L95)):*

Make these ref parameters const
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 80 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-80:-KxA7wbDwdGShmiPAVVd:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L80)):*

mark this class as final
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 613 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-613:-KxA8eZbnkVWsOnkMfG_:b-de1mmm) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L613)):*

std::make_shared(std::make_unique())
</blockquote></details>

I don't understand what this code is doing anymore, you're constructing a shared_ptr from an unique_ptr?

---

*[src/core/hle/service/am/am.cpp, line 703 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-703:-KxA8sMuyn0T__NeHSAN:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L703)):*

IMO, invert the condition and handle the error in the innermost indentation level, instead of the opposite
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 795 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-795:-KxA8yJaTh-KqAGtYPVJ:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L795)):*

add a blank line here
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 255 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx8zd8wsZIxABh9Hgol:-KxA8L8frPR9S1LqkG-W:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L255)):*

use an inline static constexpr named constant for this guard value
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 767 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx9-j3M4LhpzkJjYpJZ:-KxA8qFm1EuAalLS7ZdG:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L767)):*

add a log here
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 873 at r12](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx90KImUYVwCxbUgiec:-KxA9-N_RySdccuaxDWc:b32wtd) ([raw file](https://github.com/citra-emu/citra/blob/9fe5e517f3dc5dea479aec437bffda4a04a63b12/src/core/hle/service/am/am.cpp#L873)):*

Invert the condition so that the error case is higher up and the normal flow is unindented, same everywhere else, i'll stop pointing this out
</blockquote></details>

ok

---

*[src/core/hle/service/am/am.cpp, line 229 at r13](https://reviewable.io:443/reviews/citra-emu/citra/3029#-KxA8BtU1zEAJ-M0RTTa:-KxA8BtU1zEAJ-M0RTTb:b5e8bg8) ([raw file](https://github.com/citra-emu/citra/blob/ce965372d19dc42acfbe1cd700ade7f1fd918236/src/core/hle/service/am/am.cpp#L229)):*
> ```cpp
>         if (cia_installing_written >= cia_installing.GetContentOffset() &&
>             cia_install_state != TMDLoaded) {
>             auto result = WriteTitleMetadata(offset, length, buffer);
> ```

what are you using this variable for?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3029)*

shinyquagsire23:
Review status: 3 of 10 files reviewed at latest revision, 47 unresolved discussions.

---

*[src/core/hle/service/am/am.cpp, line 613 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-613:-KxAnB2jUGH4MJDxbAtk:b-lgb0lq) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L613)):*

I don't understand what this code is doing anymore, you're constructing a shared_ptr from an unique_ptr?
</blockquote></details>

OK I think for now I'm going to revert back because I'm not 100% here, I think you might have to spell this one out a bit. My understanding is that file should be a shared_ptr and the CIAFile a unique_ptr here.

---

*[src/core/hle/service/am/am.cpp, line 229 at r13](https://reviewable.io:443/reviews/citra-emu/citra/3029#-KxA8BtU1zEAJ-M0RTTa:-KxAdZHkL4hZF1d4AmCe:b-xew4wd) ([raw file](https://github.com/citra-emu/citra/blob/ce965372d19dc42acfbe1cd700ade7f1fd918236/src/core/hle/service/am/am.cpp#L229)):*

what are you using this variable for?
</blockquote></details>

Should be returned if failed like the one below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3029)*

Subv:
Reviewed 4 of 7 files at r13, 3 of 3 files at r14.
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[src/core/hle/service/am/am.cpp, line 39 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-39:-KxKZDcOFtkpZa_s1vN0:b-oi7lra) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L39)):*

I don't understand the problem here, why does NonCopyable prevent you from assigning member variables?
</blockquote></details>

As discussed on IRC, feel free to make the FileBackend::Write method nonconst and move all these variables into the CIAFile class.

---

*[src/core/hle/service/am/am.cpp, line 613 at r4](https://reviewable.io:443/reviews/citra-emu/citra/3029#-Kx0SP24sREkAzcAanMA-r4-613:-KxK_P-zfmDl6YXYg8wJ:bt0pgii) ([raw file](https://github.com/citra-emu/citra/blob/b29e892f04702f9ec7380b7b88b640d74bab178a/src/core/hle/service/am/am.cpp#L613)):*
<details><summary><i>Previously, shinyquagsire23 (Max Thomas) wrote…</i></summary><blockquote>

OK I think for now I'm going to revert back because I'm not 100% here, I think you might have to spell this one out a bit. My understanding is that file should be a shared_ptr and the CIAFile a unique_ptr here.
</blockquote></details>

this *has* to be a shared_ptr, since it has std::enable_shared_from_this in its inheritance graph.
`auto file = std::make_shared<Service::FS::File>(std::make_unique<CIAFile>(media_type), cia_path);`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3029)*

Subv:
Reviewed 12 of 12 files at r15.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3029)*

Subv:
:lgtm:

---

Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3029#-:-KxZcYrxe7GnBDWLse9J:bnfp4nl)*

src/core/file_sys/cia_container.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+u64 CIAContainer::GetContentOffset(u16 index) const {
+    u64 offset =
+        Common::AlignUp(GetTitleMetadataOffset() + cia_header.tmd_size, CIA_SECTION_ALIGNMENT);
+    for (u16 i = 0; i < index; i++) {
+        // The content_present is a bit array which defines which content in the TMD
+        // is included in the CIA, so check the bit for this index and add if set.
+        // The bits in the content index are arranged w/ index 0 as the MSB, 7 as the LSB, etc.
+        if (cia_header.content_present[i >> 3] & 0x80 >> (i & 7))
```
wwylele: This bit checking pattern, along with the comment, deserves its own member function in `cia_header`.

btw, this isn't documented on 3dbrew. Could you also add it to 3dbrew?

src/core/file_sys/cia_container.h
```diff
@@ -0,0 +1,100 @@
...
+private:
+    struct Header {
+        u32 header_size;
+        u16 type;
+        u16 version;
+        u32 cert_size;
+        u32 tik_size;
+        u32 tmd_size;
+        u32 meta_size;
+        u64 content_size;
```
wwylele: All these members above should be marked as `_le`

src/core/file_sys/cia_container.h
```diff
@@ -0,0 +1,100 @@
...
+        std::array<u8, CIA_CONTENT_BITS_SIZE> content_present;
+    };
+
+    static_assert(sizeof(Header) == CIA_HEADER_SIZE, "CIA Header structure size is wrong");
+
+    struct Metadata {
+        std::array<u64, 0x30> dependencies;
+        std::array<u8, 0x180> reserved;
+        u32 core_version;
```
wwylele: `u32_le`

src/core/file_sys/cia_container.h
```diff
@@ -0,0 +1,100 @@
...
+        std::array<u8, 0xfc> reserved_2;
+    };
+
+    static_assert(sizeof(Metadata) == CIA_METADATA_SIZE, "CIA Metadata structure size is wrong");
+
+    bool loaded = false;
+    std::string filepath;
+    std::unique_ptr<FileBackend> backend;
```
wwylele: I can't find any where `filepath` or `backend` being used.

src/core/file_sys/cia_container.cpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    return Loader::ResultStatus::Success;
+}
+
+Loader::ResultStatus CIAContainer::LoadHeader(const std::vector<u8>& header_data, size_t offset) {
+    if (header_data.size() - offset < sizeof(Header))
+        return Loader::ResultStatus::Error;
+
+    memcpy(&cia_header, header_data.data(), sizeof(Header));
```
wwylele: `std::memcpy`. Applies for all usage below.

src/core/hle/service/am/am.cpp
```diff
@@ -60,7 +78,192 @@ struct TicketInfo {
...
+    ResultVal<size_t> Write(u64 offset, size_t length, bool flush, const u8* buffer) override {
+        written += length;
+
+        // TODO(shinyquagsire23): Can we assume that things will only be written in sequence?
+        // Does AM send an error if we write to things out of order?
+        // Or does it just ignore offsets and assume a set sequence of incoming data?
+
+        // The data in CIAs is always stored CIA Header > Cert > Ticket > TMD > Content > Meta.
+        // The CIA Header describes Cert, Ticket, TMD, total content sizes, and TMD is needed for
+        // content sizes so it ends up becoming a problem of keeping track of how  much has been
```
wwylele: `how  much`
double space

src/core/hle/service/am/am.cpp
```diff
@@ -545,7 +762,369 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+    // Get a File from our Handle
+    auto file_res = GetFileFromHandle(rp.PopHandle());
+    if (!file_res.Succeeded()) {
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(file_res.Code());
+        return;
+    }
+
+    size_t output_buffer_size;
+    VAddr output_buffer = rp.PopStaticBuffer(&output_buffer_size);
```
wwylele: Should use `PeekStaticBuffer` because this is the output buffer descriptor (the "mailbox" descriptor) in TLS+0x180 area (so-called command buffer [64-65]), not the input buffer descriptor (the "mail" descriptor) in TLS+0x80 area (the command buffer)

src/core/hle/service/am/am.cpp
```diff
@@ -545,7 +762,369 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+    if (container.Load(*file->backend) != Loader::ResultStatus::Success) {
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(ResultCode(ErrCodes::InvalidCIAHeader, ErrorModule::AM,
+                           ErrorSummary::InvalidArgument, ErrorLevel::Permanent));
+        return;
+    }
+
+    Memory::WriteBlock(output_buffer, container.GetDependencies().data(), output_buffer_size);
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
```
wwylele: The header should be likely `MakeBuilder(1, 2);`, because the response need to push a input buffer descriptor ("a mail") and let the kernel to post it to the output buffer ("the mailbox") set by client. This is unimplemented in citra, and the mail posting is done by the service itself (`Memory::WriteBlock`), but better to leave a comment here.

src/core/hle/service/am/am.cpp
```diff
@@ -545,7 +762,369 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+
+    // Get a File from our Handle
+    auto file_res = GetFileFromHandle(rp.PopHandle());
+    if (!file_res.Succeeded()) {
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(file_res.Code());
+        return;
+    }
+
+    VAddr output_buffer = rp.PopStaticBuffer();
```
wwylele: ditto about `PeekStaticBuffer`

src/core/hle/service/am/am.cpp
```diff
@@ -545,7 +762,369 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+        file->backend->Read(container.GetMetadataOffset(), output_buffer_size, temp.data());
+    if (read_result.Failed() || *read_result != output_buffer_size) {
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(ResultCode(ErrCodes::InvalidCIAHeader, ErrorModule::AM,
+                           ErrorSummary::InvalidArgument, ErrorLevel::Permanent));
+        return;
+    }
+
+    Memory::WriteBlock(output_buffer, temp.data(), output_buffer_size);
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
```
wwylele: ditto about header with static buffer posting

src/core/hle/service/am/am.cpp
```diff
@@ -545,7 +762,369 @@ void CheckContentRightsIgnorePlatform(Service::Interface* self) {
...
+    if (read_result.Failed() || *read_result != output_buffer_size) {
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(ResultCode(ErrCodes::InvalidCIAHeader, ErrorModule::AM,
+                           ErrorSummary::InvalidArgument, ErrorLevel::Permanent));
+        return;
+    }
+
+    Memory::WriteBlock(output_buffer, temp.data(), output_buffer_size);
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
```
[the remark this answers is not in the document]
shinyquagsire23: Huh, I had to go back and check that but yeah that's consistent. Never knew.

PR #3035, opened 2017-10-23
Fix typo in travis clang-format comment version name

superheri:
Correcting a typo in the comment of the travis clang-format about the version number to actually represent the code.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3035)


wwylele:
:thinking: I thought that comment actually meant:
"The default clang-format (in Travis CI) points to default 3.5 version one, (so we have to explicitly set it to 3.9 here)" ?

superheri:
It feels more to me that it is an old comment left there when the version was changed.

Dragios:
I think @wwylele is correct here.
See #2086 and #2257

superheri:
If it is for the Travis CI default version, I think we should just clost this pull request then. Sorry for the misunderstanding.

Issue #3036, opened 2017-10-23
Missing libpcre2-16-0.dll from both latest Nightly and Canary builds


Hexagon12:
Using both Nightly-aee8bf6 and Canary-c77463f

The program instantly quits after requiring libpcre2-16-0.dll.

Related to MinGW builds, MSVC builds don't provide the error.
![citra-qt_2017-10-23_19-40-34](https://user-images.githubusercontent.com/25362478/31903365-94a75fd8-b830-11e7-942f-39b4933cf152.png)

jroweboy:
@Hexagon12 thanks for noticing this. this error tends to happen when appveyor updates msys2 on their machine and since we don't do a static build, we have to hardcode dependencies, meaning citra may randomly break like this. The solution is to add the missing dll to the copy step. Please double check that this is the only missing dll (add it locally) and submit a pull request to add any of the missing dlls to build. 

See: https://github.com/citra-emu/citra/blob/master/appveyor.yml#L121 for the current list of dlls

Hexagon12:
Fixed by #3037, waiting for the builds to arrive.

Dragios:
Should be related to the missing audio as I noticed the updated MinGW dependency no longer generates this error message.
```
C:/projects/citra/src/audio_core/hle/source.cpp: In member function 'void DSP::HLE::Source::Reset()':
C:/projects/citra/src/audio_core/hle/source.cpp:45:14: warning: converting to 'std::priority_queue<DSP::HLE::Source::Buffer, std::vector<DSP::HLE::Source::Buffer>, DSP::HLE::Source::BufferOrder>' from initializer list would use explicit constructor 'std::priority_queue<_Tp, _Sequence, _Compare>::priority_queue(const _Compare&, _Sequence&&) [with _Tp = DSP::HLE::Source::Buffer; _Sequence = std::vector<DSP::HLE::Source::Buffer>; _Compare = DSP::HLE::Source::BufferOrder]'
     state = {};
              ^
```

MerryMage:
Closed by #3037.

PR #3037, opened 2017-10-23
Fix Missing libpcre2-16-0.dll from both latest Nightly and Canary builds #3036

acnleditor2:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3037)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -122,7 +122,7 @@ after_build:
                        # QT dll dependencies
                        "libbz2-*.dll","libicudt*.dll","libicuin*.dll","libicuuc*.dll","libffi-*.dll",
                        "libfreetype-*.dll","libglib-*.dll","libgobject-*.dll","libgraphite2.dll","libiconv-*.dll",
-                       "libharfbuzz-*.dll","libintl-*.dll","libpcre-*.dll","libpcre16-*.dll","libpng16-*.dll",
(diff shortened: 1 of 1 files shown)


wwylele:
Question: out of all "pcre" dlls
```
"libpcre-*.dll",
"libpcre16-*.dll",
"libpcre2-16-*.dll", (new added)
"libpng16-*.dll"
```
are these still all relevant? Or some of them can be removed due to the update?

Hexagon12:
@wwylele Those DLLs are still relevant.

Jhon591:
try using it on me
https://gbatemp.net/threads/citra-unofficial-chinese-builds-discussion.431974/page-306#post-7649940

jroweboy:
@wwylele yes, they are all dependcies for qt5 in case you are wondering where they came from.

@MerryMage a proper fix would be much appreciated, but in the meantime, lets just merge this one so at least its not completely non functional

Also re sound issues: someone will have to investigate them separately.

acnleditor2:
@jroweboy I can create a folder called dlls in the root, add SDL2.0.5 there and make AppVeyor download it instead of copying it from the mingw folder.

Jhon591:
I replaced the New SDL2.dll with the Old one and it worked fine, so is new SDL2.dll causing No Audio in Citra.

PR #3043, opened 2017-10-23
appveyor: Determine dlls to include in package programmatically

MerryMage:
This is done by recursively calling bindump to determine the dependencies
of each binary that is required from each of the executables. Doing this
allows us to avoid hard-coding a list of required DLL files to copy into
the release archives.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3043)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -116,23 +116,16 @@ after_build:
           Get-ChildItem "$CMAKE_BINARY_DIR" -Recurse -Filter "libcurl.dll" | Copy-Item -destination $RELEASE_DIST
           Copy-Item -path "$CMAKE_SOURCE_DIR/license.txt" -destination $RELEASE_DIST
           Copy-Item -path "$CMAKE_SOURCE_DIR/README.md" -destination $RELEASE_DIST
+
           # copy all the dll dependencies to the release folder
-          # hardcoded list because we don't build static and determining the list of dlls from the binary is a pain.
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -54,7 +54,7 @@ build_script:
           # https://www.appveyor.com/docs/build-phase
           msbuild msvc_build/citra.sln /maxcpucount /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
         } else {
-          C:\msys64\usr\bin\bash.exe -lc 'mingw32-make -C mingw_build/ 2>&1'
+          C:\msys64\usr\bin\bash.exe -lc 'mingw32-make -j4 -C mingw_build/ 2>&1'
```
jroweboy: adding -j4 wouldn't help much since appveyor free tier builds are single threaded

MerryMage:
> We have a CopyDeps cmake step for windows development, and I'm wondering if we should rebuild that cmake step off this instead :) Would remove the need for another hardcoded list of DLLs (which is always welcome!)

That's an interesting idea. Might look into that.

jroweboy:
@MerryMage In the meantime, do you have any objections to shipping what you have here? If you feel like unifying copy dependencies steps, that can come in a later PR. This would be really nice as I might have to drag in some more libraries in order to fix the curl issues.

MerryMage:
No, go ahead. Rebased.

Jhon591:
Looks like it needed, latest dlls are casing mingw builds to crash again on exit citra while playing a game, it will work fine if this box is unchecked though.. same issue from some time ago.

PR #3047, opened 2017-10-24
Workaround for SDL2 2.0.6 audio issues

Hexagon12:
This PR is a workaround for SDL2 2.0.6 having a bug with audio output, the bug report is right here: https://bugzilla.libsdl.org/show_bug.cgi?id=3858 - Thanks @MerryMage !
This PR also downgrades the SDL2 version to 2.0.5.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3047)


jroweboy:
Reminder that we need to revert this when SDL2.0.7 lands in msys2

debugzxcv:
This pr should be revert.

PR #3048, opened 2017-10-25
Services/AM: Implement GetPatchTitleInfos, Misc Cleanup

### LINKED ISSUE #2980: Flipnote Studio 3D US: Crash on boot
Log: https://hastebin.com/icugikuyoy.xml
Commit: 30fabc4
Reproduction Steps: Run the application, see a crash.

shinyquagsire23:
Fixes #2980 (thanks to @mailwl for finding the actual issue there)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3048)

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -24,6 +24,9 @@
 namespace Service {
 namespace AM {
 
+constexpr u32 TID_HIGH_UPDATE = 0x0004000E;
+constexpr u32 TID_HIGH_DLC = 0x0004008C;
+
 static bool lists_initialized = false;
 static std::array<std::vector<u64_le>, 3> am_title_list;
 
@@ -182,7 +185,7 @@ void GetNumPrograms(Service::Interface* self) {
 }
 
 void FindContentInfos(Service::Interface* self) {
-    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1002, 4, 2); // 0x10020104
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1002, 4, 4); // 0x10020104
 
     auto media_type = static_cast<Service::FS::MediaType>(rp.Pop<u8>());
     u64 title_id = rp.Pop<u64>();
@@ -225,7 +228,7 @@ void FindContentInfos(Service::Interface* self) {
 }
 
 void ListContentInfos(Service::Interface* self) {
-    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1003, 5, 1); // 0x10030142
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1003, 5, 2); // 0x10030142
     u32 content_count = rp.Pop<u32>();
     auto media_type = static_cast<Service::FS::MediaType>(rp.Pop<u8>());
     u64 title_id = rp.Pop<u64>();
@@ -264,7 +267,7 @@ void ListContentInfos(Service::Interface* self) {
 }
 
 void DeleteContents(Service::Interface* self) {
-    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1004, 4, 1); // 0x10040102
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x1004, 4, 2); // 0x10040102
     u8 media_type = rp.Pop<u8>();
     u64 title_id = rp.Pop<u64>();
     u32 content_count = rp.Pop<u32>();
@@ -278,7 +281,7 @@ void DeleteContents(Service::Interface* self) {
 }
 
 void GetProgramList(Service::Interface* self) {
-    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 2, 2, 1); // 0x00020082
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 2, 2, 2); // 0x00020082
 
     u32 count = rp.Pop<u32>();
     u8 media_type = rp.Pop<u8>();
@@ -302,18 +305,9 @@ void GetProgramList(Service::Interface* self) {
     rb.Push(copied);
 }
 
-void GetProgramInfos(Service::Interface* self) {
-    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 3, 2, 2); // 0x00030084
-
-    auto media_type = static_cast<Service::FS::MediaType>(rp.Pop<u8>());
-    u32 title_count = rp.Pop<u32>();
-    VAddr title_id_list_pointer = rp.PopMappedBuffer();
-    VAddr title_info_out = rp.PopMappedBuffer();
-
-    std::vector<u64> title_id_list(title_count);
-    Memory::ReadBlock(title_id_list_pointer, title_id_list.data(), title_count * sizeof(u64));
-
-    for (u32 i = 0; i < title_count; i++) {
+ResultCode GetTitleInfoFromList(const std::vector<u64>& title_id_list,
+                                Service::FS::MediaType media_type, VAddr title_info_out) {
+    for (u32 i = 0; i < title_id_list.size(); i++) {
         std::string tmd_path = GetTitleMetadataPath(media_type, title_id_list[i]);
 
         TitleInfo title_info = {};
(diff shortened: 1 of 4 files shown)


src/core/hle/service/am/am.cpp
```diff
@@ -341,6 +346,12 @@ void GetDataTitleInfos(Service::Interface* self) {
     LOG_WARNING(Service_AM, "(STUBBED) called");
 }
 
+void GetPatchTitleInfos(Service::Interface* self) {
+    GetProgramInfos(self);
```
Subv: You can't just call GetProgramInfos, the command headers are different. If needed, factor out the body of GetProgramInfos and call that from both functions after extracting the input parameters

Subv: On that note, GetProgramInfos has ` IPC::RequestParser rp(Kernel::GetCommandBuffer(), 3, 2, 2); // 0x00030084` which is wrong, it should be ` IPC::RequestParser rp(Kernel::GetCommandBuffer(), 3, 2, 4); // 0x00030084`

Subv: According to RE, GetPatchTitleInfos has 4 translate parameters that you're not sending in the response.

Subv: The only difference between the GetPatchTitleInfos and GetProgramInfos implementation is that GetPatchTitleInfos will first iterate over the input title ids and return 0xE0E0803C if any of them satisfies `high >> 14 != 16 || high & 0xFFFF == 14`

shinyquagsire23: Ah, yeah you're right, on hardware I'm seeing a 0x100D0044 in the [0] spot for the response. Looks like it's just spitting out the same buffers passed in, at least judging by the sizes. Weird.

shinyquagsire23: Actually, shouldn't it be high & 0xFFFF != 0xE? Since the TID high should always be 0x0004000E.

Subv: > Ah, yeah you're right, on hardware I'm seeing a 0x100D0044 in the [0] spot for the response. Looks like it's just spitting out the same buffers passed in, at least judging by the sizes. Weird.

Yeah that's precisely what it's doing, it is indeed weird but 🤷‍♂️ 

> Actually, shouldn't it be high & 0xFFFF != 0xE? Since the TID high should always be 0x0004000E.

Yeah you're right, sorry, i misread the condition.

src/core/hle/service/am/am.cpp
```diff
@@ -326,23 +320,84 @@ void GetProgramInfos(Service::Interface* self) {
...
+
+    ResultCode result = RESULT_SUCCESS;
+
+    // Validate that update TIDs were passed in
+    std::vector<u64> title_id_list(title_count);
+    Memory::ReadBlock(title_id_list_pointer, title_id_list.data(), title_count * sizeof(u64));
+    for (u32 i = 0; i < title_count; i++) {
+        u32 tid_high = static_cast<u32>(title_id_list[i] >> 32);
+        if (tid_high >> 14 != 0x10 || tid_high & 0xFFFF != 0xE) {
+            result = ResultCode(ErrCodes::InvalidTIDInList, ErrorModule::AM,
```
Subv: break

src/core/hle/service/am/am.cpp
```diff
@@ -326,23 +320,84 @@ void GetProgramInfos(Service::Interface* self) {
...
+
+    if (result.IsSuccess()) {
+        result =
+            GetTitleInfoFromList(title_count, media_type, title_id_list_pointer, title_info_out);
+    }
 
-    LOG_WARNING(Service_AM, "(STUBBED) called");
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 4);
+    rb.Push(result);
+    rb.PushMappedBuffer(title_id_list_pointer, title_id_list_size, title_id_list_perms);
```
Subv: GetProgramInfos also pushes these buffers, i am unaware of whether GetDataTitleInfos does as well, but it should be checked

src/core/hle/service/am/am.cpp
```diff
@@ -326,23 +320,84 @@ void GetProgramInfos(Service::Interface* self) {
...
+    VAddr title_info_out = rp.PopMappedBuffer(&title_info_size, &title_info_perms);
+
+    ResultCode result = RESULT_SUCCESS;
+
+    // Validate that update TIDs were passed in
+    std::vector<u64> title_id_list(title_count);
+    Memory::ReadBlock(title_id_list_pointer, title_id_list.data(), title_count * sizeof(u64));
+    for (u32 i = 0; i < title_count; i++) {
+        u32 tid_high = static_cast<u32>(title_id_list[i] >> 32);
+        if (tid_high >> 14 != 0x10 || tid_high & 0xFFFF != 0xE) {
```
Subv: add some parentheses to this expression, operator precedence will screw you over otherwise.

src/core/hle/service/am/am.cpp
shinyquagsire23: Just checked, it does seem to be the case so I might be able to merge things a bit nicer then.

src/core/hle/service/am/am.cpp
```diff
@@ -302,11 +302,39 @@ void GetProgramList(Service::Interface* self) {
...
+    ResultCode result = RESULT_SUCCESS;
+
     std::vector<u64> title_id_list(title_count);
     Memory::ReadBlock(title_id_list_pointer, title_id_list.data(), title_count * sizeof(u64));
 
+    // Validate that update TIDs were passed in
+    if (must_be_patch) {
+        for (u32 i = 0; i < title_count; i++) {
+            u32 tid_high = static_cast<u32>(title_id_list[i] >> 32);
+            if ((tid_high >> 14 != 0x10) || (tid_high & 0xFFFF != 0xE)) {
```
Subv: that's not what i meant, this still has incorrect behavior due to operator precedence, use
`if ((tid_high >> 14) != 0x10 || (tid_high & 0xFFFF) != 0xE)`

shinyquagsire23: Derp, fixed

Subv:
I know this is pretty much the definition of scope creep, but GetDataTitleInfos also checks that every tid meets  `(high >> 14) == 16 && (high & 0xFFFF) == 0x8C`, if any of them doesn't meet that check, then error 0xE0E0803C is returned

src/core/hle/service/am/am.cpp
```diff
@@ -302,40 +302,9 @@ void GetProgramList(Service::Interface* self) {
...
-        if (result.IsError()) {
-            IPC::RequestBuilder rb = rp.MakeBuilder(1, 4);
-            rb.Push(result);
-            rb.PushMappedBuffer(title_id_list_pointer, title_id_list_size, title_id_list_perms);
-            rb.PushMappedBuffer(title_info_out, title_info_size, title_info_perms);
-        }
-    }
-
-    for (u32 i = 0; i < title_count; i++) {
+ResultCode GetTitleInfoFromList(std::vector<u64>& title_id_list, Service::FS::MediaType media_type,
```
Subv: make the title id list const

Subv:
Make the std::vector parameter a const reference and squash, LGTM

src/core/hle/service/am/am.cpp
```diff
@@ -326,23 +317,113 @@ void GetProgramInfos(Service::Interface* self) {
...
+
+    std::vector<u64> title_id_list(title_count);
+    Memory::ReadBlock(title_id_list_pointer, title_id_list.data(), title_count * sizeof(u64));
+
+    ResultCode result = RESULT_SUCCESS;
+
+    // Validate that DLC TIDs were passed in
+    for (u32 i = 0; i < title_count; i++) {
+        u32 tid_high = static_cast<u32>(title_id_list[i] >> 32);
+        if ((tid_high >> 14) != 0x10 || (tid_high & 0xFFFF) != 0x8C) {
```
Subv: If this is checking for the high tid to be some specific value, define that value as an inline static constexpr variable and compare it to that instead of doing this bit fiddling

src/core/hle/service/am/am.cpp
```diff
@@ -326,23 +317,113 @@ void GetProgramInfos(Service::Interface* self) {
...
+
+    std::vector<u64> title_id_list(title_count);
+    Memory::ReadBlock(title_id_list_pointer, title_id_list.data(), title_count * sizeof(u64));
+
+    ResultCode result = RESULT_SUCCESS;
+
+    // Validate that update TIDs were passed in
+    for (u32 i = 0; i < title_count; i++) {
+        u32 tid_high = static_cast<u32>(title_id_list[i] >> 32);
+        if ((tid_high >> 14) != 0x10 || (tid_high & 0xFFFF) != 0xE) {
```
Subv: same here

Issue #3051, opened 2017-10-26
All Games - Crash when opening Camera (Only on Windows, only with MinGW compiler)


acnleditor2:
Commit: citra-emu/citra@79852d3707843a31f56b8f2dbc50fa91b4851be8
Stopped working on Nightly 373.
With only hw renderer, with hw renderer+shader jit, only shader jit, hw renderer+shader jit disabled.
Specs i5-7200U, Intel HD Graphics 620, Windows 10 15063.
Log (nightly 372, works):
https://pastebin.com/3mRE92d4

Log (new, dont works):
https://pastebin.com/4fWSUVtR

wwylele:
I cannot reproduce. Same commit 26065fd, Ubuntu system, tested games:
 - Pokemon Sun (QR code)
 - Pokemon Super Mystery Dungeon (Intro without a save)
 - Mii Maker (QR code)
Black image as usual, without crashing

adityaruplaha:
Reproduced by me.

https://community.citra-emu.org/t/psmd-crash-on-strtup-i-worked-well-earlier/4541

adityaruplaha:
Happens on #b36f748 as well.

wwylele:
Alright, here is a summery of some researching:

 - The crash comes from the `std::future`/`std::async` usage in the camera module. This is confirmed by commenting out all the `std::future` usage to see no crash anymore.
 - The stack is broken upon the crash
 - The crash only happens in the Qt frontend. Built in the same environment on the same commit, and running the same game, the SDL frontend doesn't crash
 - To test if `std::future` itself is the problem, an experiment was conducted:
the example code from http://en.cppreference.com/w/cpp/thread/future was put into
   - a standalone source file
   - the beginning of the `main()` in citra Qt frontend
   - the beginning of the `main()` in citra SDL frontend

  and compile them under the same msys environment. The result is that only the Qt frontend one crashes on boot, and the other two give the correct result
 - The theory is that `std::future` gets broken somehow by linking against Qt library
 - Also, the telemetry crash (#3039) has a very similar broken stack. Telemetry is also the other one of the only two user of `std::future` in citra. This indicates that it might be the crash comes from the same source.

acnleditor2:
![sdl](https://user-images.githubusercontent.com/22228082/32138905-362b24ce-bc01-11e7-9ee8-24ba83dd7d66.PNG)
on sdl

mailwl:
maybe use boost future in this case?

[...]

PR #3063, opened 2017-10-30
Appveyor: Use Previous version of appveyor build environment

B3n30:
This is a temporary fix for the issues caused by the most recent appveyor update (issue #3051). This will only work until the next update of appyevor, since then the current (broken) version will get the previous. So we still need to find a solution, but at least until then, we have a non broken version.

P.S. @chris062689 forced me to push this

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3063)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ cache:
   - C:\ProgramData\chocolatey\bin -> appveyor.yml
   - C:\ProgramData\chocolatey\lib -> appveyor.yml
 
-os: Visual Studio 2017
+os: Previous Visual Studio 2017
 
 environment:
   # Tell msys2 to add mingw64 to the path


acnleditor2:
fixes #3051!!!

superheri:
Can it be linked to the fact that it is impossible for me to build on windows with CMake 3.9.4 and Visual Studio 2017 Entreprise 15.4.1 (the latest Visual Studio Update)?

wwylele:
**To make it clear: unlike the name in the code suggests, this has nothing to do with Visual Studio 2017, but for the workaround of MinGW update bug. Appveyor updated both VS and MSYS the same time (of which the latter one is undocumented), so rolling back to previous VS also gives us previous version of MSYS.**

@superheri yours is a separate issue. You can ask on forum or Discord for .

B3n30:
Changed the title to make clear what this PR does

All of the issues this PR fixes, those issues will come up again as soon as appveyor releases a new update. So those issues shouldn't get closed, even if this gets merged.

appveyor.yml
```diff
@@ -5,7 +5,7 @@ cache:
   - C:\ProgramData\chocolatey\bin -> appveyor.yml
   - C:\ProgramData\chocolatey\lib -> appveyor.yml
 
-os: Visual Studio 2017
+os: Previous Visual Studio 2017
```
bunnei: Should we add a comment in here explaining why?

B3n30:
Closing since appveyor updated again, so this will no longer be useful

PR #3067, opened 2017-11-04
Removed duplicate calls of WSACleanup in soc_u.

Hexagon12:
Soc_u is the reason for most of the WSAStartup issues. It calls WSAStartup only on SOC_U::InitializeSockets,
but calls WSACleanup every time it gets destructed in addition to SOC_U::ShutdownSockets. Thus calling WSACleanup way more often than WSAStartup and thus close it although other parts of the code still need the sockets.

With that, we should test if removing WSAStartup in web_service is possible.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3067)


wwylele:
LGTM please fix clang-format problem

PR #3068, opened 2017-11-04
Libnetwork: add password protected rooms, guid, and error fixes

B3n30:
This PR adds the possibility the set a password for rooms
It also adds a unique (randomly generated) GUID used to announce rooms to our web-service.
It adds the possibility to set a preferred game for a room.

The default Port is changed to "CITRA" too ;-)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3068)

### DIFF
--- a/src/network/room.cpp
+++ b/src/network/room.cpp
@@ -4,17 +4,16 @@
 
 #include <algorithm>
 #include <atomic>
+#include <iomanip>
 #include <mutex>
 #include <random>
+#include <sstream>
 #include <thread>
 #include "enet/enet.h"
 #include "network/packet.h"
 #include "network/room.h"
 
 namespace Network {
-
-/// Maximum number of concurrent connections allowed to this room.
-static constexpr u32 MaxConcurrentConnections = 10;
 
 class Room::RoomImpl {
 public:
@@ -27,6 +26,8 @@ public:
     std::atomic<State> state{State::Closed}; ///< Current state of the room.
     RoomInformation room_information;        ///< Information about this room.
 
+    std::string password; ///< The password required to connect to this room.
+
     struct Member {
         std::string nickname;   ///< The nickname of the member.
         GameInfo game_info;     ///< The current game of the member
@@ -38,8 +39,12 @@ public:
     mutable std::mutex member_mutex; ///< Mutex for locking the members list
     /// This should be a std::shared_mutex as soon as C++17 is supported
 
+    std::string guid; /// The GUID of the room
+
     RoomImpl()
-        : random_gen(std::random_device()()), NintendoOUI{0x00, 0x1F, 0x32, 0x00, 0x00, 0x00} {}
+        : random_gen(std::random_device()()), NintendoOUI{0x00, 0x1F, 0x32, 0x00, 0x00, 0x00} {
+        CreateGUID();
+    }
 
     /// Thread that receives and dispatches network packets
     std::unique_ptr<std::thread> room_thread;
@@ -82,6 +87,11 @@ public:
     void SendVersionMismatch(ENetPeer* client);
 
     /**
+     * Sends a ID_ROOM_WRONG_PASSWORD message telling the client that the password is wrong.
+     */
+    void SendWrongPassword(ENetPeer* client);
+
+    /**
      * Notifies the member that its connection attempt was successful,
      * and it is now part of the room.
      */
@@ -99,6 +109,8 @@ public:
      * <MessageID>ID_ROOM_INFORMATION
      * <String> room_name
      * <u32> member_slots: The max number of clients allowed in this room
+     * <String> guid
+     * <u16> port
      * <u32> num_members: the number of currently joined clients
      * This is followed by the following three values for each member:
      * <String> nickname of that member
@@ -136,13 +148,18 @@ public:
      * to all other clients.
      */
     void HandleClientDisconnection(ENetPeer* client);
+
+    /**
+     * Creates a random ID in the form 12345678-1234-1234-1234-123456789012
+     */
+    void CreateGUID();
 };
 
 // RoomImpl
 void Room::RoomImpl::ServerLoop() {
     while (state != State::Closed) {
         ENetEvent event;
-        if (enet_host_service(server, &event, 100) > 0) {
+        if (enet_host_service(server, &event, 50) > 0) {
             switch (event.type) {
             case ENET_EVENT_TYPE_RECEIVE:
                 switch (event.packet->data[0]) {
@@ -187,6 +204,14 @@ void Room::RoomImpl::HandleJoinRequest(const ENetEvent* event) {
 
     u32 client_version;
     packet >> client_version;
+
+    std::string pass;
+    packet >> pass;
+
+    if (!(pass == password)) {
+        SendWrongPassword(event->peer);
+        return;
+    }
 
     if (!IsValidNickname(nickname)) {
         SendNameCollision(event->peer);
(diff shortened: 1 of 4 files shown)


src/network/room.cpp
```diff
@@ -188,6 +205,14 @@ void Room::RoomImpl::HandleJoinRequest(const ENetEvent* event) {
     u32 client_version;
     packet >> client_version;
 
+    std::string pass;
+    packet >> pass;
+
+    if (!(pass == password)) {
```
Subv: why

src/network/room_member.cpp
```diff
@@ -427,6 +450,7 @@ void RoomMember::SendWifiPacket(const WifiPacket& wifi_packet) {
     packet << wifi_packet.channel;
     packet << wifi_packet.transmitter_address;
     packet << wifi_packet.destination_address;
+    packet << (u32)wifi_packet.data.size();
```
Subv: static_cast

src/network/room_member.cpp
```diff
@@ -380,18 +391,29 @@ RoomInformation RoomMember::GetRoomInformation() const {
...
+                      u16 client_port, const MacAddress& preferred_mac,
+                      const std::string password) {
     // If the member is connected, kill the connection first
-    if (room_member_impl->loop_thread && room_member_impl->loop_thread->joinable()) {
-        room_member_impl->SetState(State::Error);
-        room_member_impl->loop_thread->join();
-        room_member_impl->loop_thread.reset();
+    if (room_member_impl->loop_thread) {
+        if (room_member_impl->loop_thread->joinable()) {
+            room_member_impl->SetState(State::Error);
```
Subv: Why can't this call Leave() ?

src/network/room.cpp
```diff
@@ -38,8 +39,12 @@ class Room::RoomImpl {
     mutable std::mutex member_mutex; ///< Mutex for locking the members list
     /// This should be a std::shared_mutex as soon as C++17 is supported
 
+    std::string guid; /// The GUID of the room
```
wwylele: Is this member necessary? `CreateGUID()` could just fill `room_information->guid` instead

src/network/room.cpp
```diff
@@ -433,26 +474,51 @@ void Room::RoomImpl::HandleClientDisconnection(ENetPeer* client) {
     BroadcastRoomInformation();
 }
 
+void Room::RoomImpl::CreateGUID() {
```
wwylele: wwylele / 白疾風 - Today at 10:51 AM
B3N30 why is it called "GUID"? I think the name is given to a specific ID format, not just a random number sequence
B3N30 - Today at 10:51 AM
thats exactly the id format we reconstruct
wwylele / 白疾風 - Today at 10:52 AM
but... according to http://www.ietf.org/rfc/rfc4122.txt, it is not supposed to be generated like that :p
B3N30 - Today at 10:55 AM
hmm. We should ask @Flame Sage. he requested the GUID and said it should be that format with all numbers just randomly generated
wwylele / 白疾風 - Today at 10:55 AM
at least I am expecting some hex digits
http://www.boost.org/doc/libs/1_65_1/libs/uuid/ we can use boost for this
B3N30 - Today at 11:01 AM
I suggested that but Subv and jroewboy were against using boost
And thinking about it i kind of agree
Since the random numbers are probably sufficient i don't think it's necessary to add boost to libnetwork just for that number
wwylele / 白疾風 - Today at 11:02 AM
:shrug: then just don't call it GUID/UUID
B3N30 - Today at 11:02 AM
Since I think we still have the plan/idea to make libnetwork an external???
wwylele / 白疾風 - Today at 11:02 AM
is it required on the server side?
B3N30 - Today at 11:03 AM
you mean web-service server side?
wwylele / 白疾風 - Today at 11:03 AM
yes
B3N30 - Today at 11:03 AM
i think it's part of node.js and they request that GUID, somehow
wwylele / 白疾風 - Today at 11:04 AM
What did we do before when we didn't send the "GUID"?
B3N30 - Today at 11:05 AM
what do you mean with before?
We didn't have the lobby/announce/web-service stuff
wwylele / 白疾風 - Today at 11:05 AM
oh ok
anyway, I don't like the current implementation to forge a fake GUID. Either make a true GUID, or don't use it at all (and make a workaround for node.js somehow)
B3N30 - Today at 11:13 AM
k
we should ask Flame Sage about the web-service side.

src/network/room_member.cpp
```diff
@@ -380,18 +391,29 @@ RoomInformation RoomMember::GetRoomInformation() const {
 }
 
 void RoomMember::Join(const std::string& nick, const char* server_addr, u16 server_port,
-                      u16 client_port, const MacAddress& preferred_mac) {
+                      u16 client_port, const MacAddress& preferred_mac,
+                      const std::string password) {
     // If the member is connected, kill the connection first
-    if (room_member_impl->loop_thread && room_member_impl->loop_thread->joinable()) {
```
wwylele: wwylele / 白疾風 - Today at 11:30 AM
B3N30 is splitting the condition into to two ifs intentional?
B3N30 - Today at 11:30 AM
yeah, otherwise it would segfault
wwylele / 白疾風 - Today at 11:30 AM
why?
B3N30 - Today at 11:30 AM
i don't know
from my understanding it shoudln't try to check the second part of && if the first one fails but it does
and thus segfault
wwylele / 白疾風 - Today at 11:32 AM
Which compilers did you check?
B3N30 - Today at 11:33 AM
i think this issue was with xcode
so clang
wwylele / 白疾風 - Today at 11:37 AM
It would be a huge bug if a compiler messes up with short-circuit evaluation. I think it is something else
B3N30 - Today at 11:38 AM
i double checked it tried to call joinable although loop_thread was nullptr
And splitting it up fixed it
wwylele / 白疾風 - Today at 11:42 AM
wait, what the...
if (room_member_impl->loop_thread) {...} else if (room_member_impl->loop_thread) {...}
this is your new code
and what's worse is that you use a comment `// If the thread isn't running but the ptr still exists, reset it` to hide the connection of if and else if
B3N30 - Today at 11:44 AM
oh, this is clearly wrong
it should be part of the inner if and be just else
But tbh we probably could remove that part completely, I just added that to ensure that citra doesn't crash even if something really went wrong
wwylele / 白疾風 - Today at 11:45 AM
there is only one thread operating on the pointer loop_thread, right?
B3N30 - Today at 11:46 AM
yes
wwylele / 白疾風 - Today at 11:50 AM
I still don't think the segfault is caused by that. Especially now we know your code splitting if actually changes the behaviour in a way you were not aware
B3N30 - Today at 11:51 AM
So you are suggesting to revert it and investigate it more...
wwylele / 白疾風 - Today at 11:51 AM
yes

src/network/room_member.cpp
```diff
@@ -427,6 +443,7 @@ void RoomMember::SendWifiPacket(const WifiPacket& wifi_packet) {
     packet << wifi_packet.channel;
     packet << wifi_packet.transmitter_address;
     packet << wifi_packet.destination_address;
+    packet << static_cast<u32>(wifi_packet.data.size());
```
wwylele: we do we need to send the size (again) here? isn't the << overload for std::vector already sending it?

Same for the receiving part.

src/network/room_member.h
```diff
@@ -109,8 +116,8 @@ class RoomMember final {
      * This may fail if the username is already taken.
      */
     void Join(const std::string& nickname, const char* server_addr = "127.0.0.1",
-              const u16 serverPort = DefaultRoomPort, const u16 clientPort = 0,
-              const MacAddress& preferred_mac = NoPreferredMac);
+              const u16 server_port = DefaultRoomPort, const u16 client_port = 0,
+              const MacAddress& preferred_mac = NoPreferredMac, const std::string password = "");
```
lioncash: Shouldn't `password` be a `const` reference?

src/network/room_member.cpp
```diff
@@ -215,8 +222,13 @@ void RoomMember::RoomMemberImpl::HandleRoomInformationPacket(const ENetEvent* ev
     RoomInformation info{};
     packet >> info.name;
     packet >> info.member_slots;
+    packet >> info.uid;
```
Subv: Don't we have to bump the network version for this?

PR #3069, opened 2017-11-04
Announce room webservice

B3n30:
This will introduce the web-service to announce rooms to our web-service.
It also allows to fetch a list of announce rooms from the web-service.

For that the possibility to send DELETE msg to web-service was introduced, as well as some error handling in web-service backend

This PR requires #3068 to get merged first

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3069)

### DIFF
--- a/src/common/announce_multiplayer_room.h
+++ b/src/common/announce_multiplayer_room.h
@@ -1,0 +1,105 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <functional>
+#include <future>
+#include <string>
+#include <vector>
+#include "common/common_types.h"
+
+namespace Common {
+struct WebResult {
+    enum Code : u32 {
+        Success,
+        InvalidURL,
+        CredentialsMissing,
+        CprError,
+        HttpError,
+        WrongContent,
+        NoWebservice,
+    };
+    Code result_code;
+    std::string result_string;
+};
+} // namespace Common
+
+namespace AnnounceMultiplayerRoom {
+
+using MacAddress = std::array<u8, 6>;
+
+struct Room {
+    struct Member {
+        std::string name;
+        MacAddress mac_address;
+        std::string game_name;
+        u64 game_id;
+    };
+    std::string name;
+    std::string GUID;
+    std::string owner;
+    std::string ip;
+    u16 port;
+    u32 max_player;
+    u32 net_version;
+    bool has_password;
+    std::string preferred_game;
+    u64 preferred_game_id;
+
+    std::vector<Member> members;
+};
+using RoomList = std::vector<Room>;
+
+/**
+ * A AnnounceMultiplayerRoom interface class. A backend to submit/get to/from a web service should
+ * implement this interface.
+ */
+class Backend : NonCopyable {
+public:
+    virtual ~Backend() = default;
+    virtual void SetRoomInformation(const std::string& guid, const std::string& name,
+                                    const u16 port, const u32 max_player, const u32 net_version,
+                                    const bool has_password, const std::string& preferred_game,
+                                    const u64 preferred_game_id) = 0;
+    virtual void AddPlayer(const std::string& nickname, const MacAddress& mac_address,
+                           const u64 game_id, const std::string& game_name) = 0;
+    virtual std::future<Common::WebResult> Announce() = 0;
+    virtual void ClearPlayers() = 0;
+    virtual std::future<RoomList> GetRoomList(std::function<void()> func) = 0;
+    virtual void Delete() = 0;
+};
+
+/**
+ * Empty implementation of AnnounceMultiplayerRoom interface that drops all data. Used when a
+ * functional backend implementation is not available.
+ */
+class NullBackend : public Backend {
+public:
+    ~NullBackend() = default;
+    void SetRoomInformation(const std::string& /*guid*/, const std::string& /*name*/,
+                            const u16 /*port*/, const u32 /*max_player*/, const u32 /*net_version*/,
+                            const bool /*has_password*/, const std::string& /*preferred_game*/,
+                            const u64 /*preferred_game_id*/) override {}
+    void AddPlayer(const std::string& /*nickname*/, const MacAddress& /*mac_address*/,
+                   const u64 /*game_id*/, const std::string& /*game_name*/) override {}
+    std::future<Common::WebResult> Announce() override {
+        return std::async(std::launch::async, []() {
+            return Common::WebResult{Common::WebResult::Code::NoWebservice,
+                                     "WebService is missing"};
+        });
+    }
+    void ClearPlayers() override {}
+    std::future<RoomList> GetRoomList(std::function<void()> func) override {
+        return std::async(std::launch::async, [func]() {
+            func();
+            return RoomList{};
+        });
+    }
+
+    void Delete() override {}
+};
+
+} // namespace AnnounceMultiplayerRoom
--- a/src/common/announce_multiplayer_room.h
+++ b/src/common/announce_multiplayer_room.h
@@ -60,15 +60,53 @@ using RoomList = std::vector<Room>;
 class Backend : NonCopyable {
 public:
     virtual ~Backend() = default;
+
+    /**
+     * Sets the Information that gets used for the announce
+     * @param uid The Id of the room
+     * @param name The name of the room
+     * @param port The port of the room
+     * @param net_version The version of the libNetwork that gets used
+     * @param has_password True if the room is passowrd protected
+     * @param preferred_game The preferred game of the room
+     * @param preferred_game_id The title id of the preferred game
+     */
     virtual void SetRoomInformation(const std::string& uid, const std::string& name, const u16 port,
                                     const u32 max_player, const u32 net_version,
                                     const bool has_password, const std::string& preferred_game,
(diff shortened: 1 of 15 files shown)


JayFoxRox:
Is there any good reason as to why announce is in core instead of being in web_service as the PR title and description suggest?
This makes [awkward hacks like this](https://github.com/JayFoxRox/citra/pull/52) necessary to make the dedicated room more portable.

I also think that the announce could itself should not read the `Settings::`.
That only causes [ugly code like this](https://github.com/JayFoxRox/citra/pull/52/files#diff-2a7a5898ce7fc13b5ee55f55a2819014R195). If the settings were just passed to the constructor this would look somewhat better probably.

src/common/announce_multiplayer_room.h
```diff
@@ -0,0 +1,105 @@
...
+#include <array>
+#include <functional>
+#include <future>
+#include <string>
+#include <vector>
+#include "common/common_types.h"
+
+namespace Common {
+struct WebResult {
+    enum Code : u32 {
```
Subv: enum class

src/core/announce_multiplayer_session.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void AnnounceMultiplayerSession::AnnounceMultiplayerLoop() {
+    while (!finished) {
+        std::this_thread::sleep_for(announce_time_interval / 10);
+    }
+    announce = true;
+    finished = false;
+    std::future<Common::WebResult> future;
+    while (announce) {
+        if (std::shared_ptr<Network::Room> room = Network::GetRoom().lock()) {
```
Subv: This whole while block can flow better if you move the `std::shared_ptr<Network::Room> room = Network::GetRoom().lock()` to a normal statement instead of an if-assignment

src/core/announce_multiplayer_session.cpp
```diff
@@ -0,0 +1,119 @@
...
+    std::lock_guard<std::mutex> lock(callback_mutex);
+    error_callbacks.erase(handle);
+}
+
+AnnounceMultiplayerSession::~AnnounceMultiplayerSession() {
+    Stop();
+}
+
+void AnnounceMultiplayerSession::AnnounceMultiplayerLoop() {
+    while (!finished) {
```
Subv: What is this for?

src/core/announce_multiplayer_session.cpp
```diff
@@ -0,0 +1,119 @@
...
+        std::make_unique<std::thread>(&AnnounceMultiplayerSession::AnnounceMultiplayerLoop, this);
+}
+
+void AnnounceMultiplayerSession::Stop() {
+    if (!announce && finished)
+        return;
+    announce = false;
+    // Detaching the loop, to not wait for the sleep to finish. The loop thread will finish soon.
+    if (announce_multiplayer_thread) {
+        announce_multiplayer_thread->detach();
```
Subv: Are you sure this will not crash?

src/web_service/telemetry_json.cpp
```diff
@@ -80,7 +80,10 @@ void TelemetryJson::Complete() {
     SerializeSection(Telemetry::FieldType::UserFeedback, "UserFeedback");
     SerializeSection(Telemetry::FieldType::UserConfig, "UserConfig");
     SerializeSection(Telemetry::FieldType::UserSystem, "UserSystem");
-    PostJson(endpoint_url, TopSection().dump(), true, username, token);
+
+    // Send the telemetry async but don't handle the errors since the were written to the log
```
Subv: they

src/web_service/telemetry_json.cpp
```diff
@@ -80,7 +80,10 @@ void TelemetryJson::Complete() {
...
+    static std::future<Common::WebResult> future =
+        PostJson(endpoint_url, TopSection().dump(), true, username, token);
```
Subv: Won't this only be called once in the lifetime of the application if you assign it to a static variable like this?

B3n30:
@JayFoxRox I struggled a lot with where to place the announce session part. Problem is if you pass the Settings stuff to the constructor it won't changed when you apply new settings, thus you would have to restart citra to apply the new settings, or call some setters when applying the settings, which would cause some ring dependency. 

A similar problem occures if you place this stuff in web_service. You can look at bunnei's telemetry PR where he struggled with that and decided to put the telemetry session in.

But I'm really open for suggestions on how to change this, since I totally agree with the issue it causes e.g. for dedicated_room

src/core/announce_multiplayer_session.cpp
B3n30: It was the best way to make the stop non-blocking.
Thats the reason the std::atomic<bool> finished is there.
The current loop gets detached but probably live for a few more seconds till the sleep is done.
If you start a new announce the new loop will just wait in the while (!finished) loop till the old announce loop is done and then it can start announcing. Thus there is only one announce_loop running and there won't be any race conditions

src/core/announce_multiplayer_session.cpp
B3n30: explained above

src/web_service/telemetry_json.cpp
B3n30: no. If the code reaches this point a second time, and the async is finished, it will just write the new return value to the static. If the async isn't finished yet (which shouldn't happen), it will just wait for the async to finish, before performing a new one.

src/core/announce_multiplayer_session.cpp
Subv: And does detaching allow you to call the thread's destructor via reset?

B3n30: yes it does, see http://en.cppreference.com/w/cpp/thread/thread/detach

src/common/announce_multiplayer_room.h
```diff
@@ -0,0 +1,105 @@
...
+public:
+    ~NullBackend() = default;
+    void SetRoomInformation(const std::string& /*guid*/, const std::string& /*name*/,
+                            const u16 /*port*/, const u32 /*max_player*/, const u32 /*net_version*/,
+                            const bool /*has_password*/, const std::string& /*preferred_game*/,
+                            const u64 /*preferred_game_id*/) override {}
+    void AddPlayer(const std::string& /*nickname*/, const MacAddress& /*mac_address*/,
+                   const u64 /*game_id*/, const std::string& /*game_name*/) override {}
+    std::future<Common::WebResult> Announce() override {
+        return std::async(std::launch::async, []() {
```
wwylele: use `std::launch::deferred` since it is jut putting a ready result

src/common/announce_multiplayer_room.h
```diff
@@ -0,0 +1,105 @@
...
+    virtual ~Backend() = default;
+    virtual void SetRoomInformation(const std::string& guid, const std::string& name,
+                                    const u16 port, const u32 max_player, const u32 net_version,
+                                    const bool has_password, const std::string& preferred_game,
+                                    const u64 preferred_game_id) = 0;
+    virtual void AddPlayer(const std::string& nickname, const MacAddress& mac_address,
+                           const u64 game_id, const std::string& game_name) = 0;
+    virtual std::future<Common::WebResult> Announce() = 0;
+    virtual void ClearPlayers() = 0;
+    virtual std::future<RoomList> GetRoomList(std::function<void()> func) = 0;
```
wwylele: document what is `func` (I presume it is a callback when the result is ready?)

src/core/announce_multiplayer_session.h
```diff
@@ -0,0 +1,71 @@
...
+public:
+    AnnounceMultiplayerSession();
+    ~AnnounceMultiplayerSession();
+
+    /**
+     * Allows to bind a function that will get called if the announce encounters an error
+     * @param function The function that gets called
+     * @return A handle that can be used the unbind the function
+     */
+    std::shared_ptr<std::function<void(const Common::WebResult&)>> BindErrorCallback(
```
wwylele: better to alias the return type as a `CallbackHandle` type to make clear what it means

src/core/announce_multiplayer_session.h
```diff
@@ -0,0 +1,71 @@
...
+
+private:
+    std::atomic<bool> announce{false};
+    std::atomic<bool> finished{true};
+    std::mutex callback_mutex;
+    std::set<std::shared_ptr<std::function<void(const Common::WebResult&)>>> error_callbacks;
+    std::unique_ptr<std::thread> announce_multiplayer_thread;
+
+    std::unique_ptr<AnnounceMultiplayerRoom::Backend>
+        backend; ///< Backend interface that logs fields
```
wwylele: move the comment above the code to make it pretty?

src/core/announce_multiplayer_session.cpp
wwylele: I have two concerns about this:
 - `backend->Delete();` is called immediately after this, when the thread might still uses the `backend` for a while. Is this OK, from both lifetime perspective, and data race perspective?
 - `Stop();` is called by the destructor, which immediately free the object without waiting for the thread to finish. The thread, as a member function of this object, can use-after-free object members.

wwylele:
@Subv would you like to give another review?

B3n30:
I have a plan to address the issues JayFoxRox mentioned, but since it will include changes in other open PRs I would prefer to get the current status merged first and make the necessary changes to move the announce_session to the web_service_lib in a new PR.

@Subv Can you give this another review so we can merge it?

PR #3070, opened 2017-11-04
NWM_UDS: change to Service Framework

B3n30:
This changes uds to a service framework.

Functional changes:
- add NWM_UDS::ConnectToNetwork and NWM_UDS::DisconnectNetwork
- ties UDS to libnetwork

This PR requires #2968 to get merged first

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3070)

### DIFF
--- a/src/core/hle/service/nwm/nwm_uds.cpp
+++ b/src/core/hle/service/nwm/nwm_uds.cpp
@@ -444,7 +444,6 @@ void HandleDeauthenticationFrame(const Network::WifiPacket& packet) {
     connection_status.total_nodes--;
 
     network_info.total_nodes--;
-    node_info.erase(node);
     connection_status_event->Signal();
 }
 
--- a/src/core/hle/service/nwm/nwm_uds.cpp
+++ b/src/core/hle/service/nwm/nwm_uds.cpp
@@ -7,6 +7,7 @@
 #include <atomic>
 #include <cstring>
 #include <list>
+#include <map>
 #include <mutex>
 #include <unordered_map>
 #include <vector>
@@ -75,12 +76,21 @@ static u8 network_channel = DefaultNetworkChannel;
 // Information about the network that we're currently connected to.
 static NetworkInfo network_info;
 
+// Mapping of mac addresses to their respective node_ids.
+static std::map<MacAddress, u32> node_map;
+
 // Event that will generate and send the 802.11 beacon frames.
 static CoreTiming::EventType* beacon_broadcast_event;
+
+// Callback identifier for the OnWifiPacketReceived event.
+static Network::RoomMember::CallbackHandle<Network::WifiPacket> wifi_packet_received;
 
 // Mutex to synchronize access to the connection status between the emulation thread and the
 // network thread.
 static std::mutex connection_status_mutex;
+
+// token for the blocking ConnectToNetwork
+static ThreadContinuationToken connection_token;
 
 // Mutex to synchronize access to the list of received beacons between the emulation thread and the
 // network thread.
@@ -119,7 +129,12 @@ std::list<Network::WifiPacket> GetReceivedBeacons(const MacAddress& sender) {
 
 /// Sends a WifiPacket to the room we're currently connected to.
 void SendPacket(Network::WifiPacket& packet) {
-    // TODO(Subv): Implement.
+    if (auto room_member = Network::GetRoomMember().lock()) {
+        if (room_member->GetState() == Network::RoomMember::State::Joined) {
+            packet.transmitter_address = room_member->GetMacAddress();
+            room_member->SendWifiPacket(packet);
+        }
+    }
 }
 
 /*
@@ -214,6 +229,8 @@ static void HandleEAPoLPacket(const Network::WifiPacket& packet) {
 
         network_info.total_nodes++;
 
+        node_map[packet.transmitter_address] = node_id;
+
         // Send the EAPoL-Logoff packet.
         using Network::WifiPacket;
         WifiPacket eapol_logoff;
@@ -237,6 +254,7 @@ static void HandleEAPoLPacket(const Network::WifiPacket& packet) {
         }
         auto logoff = ParseEAPoLLogoffFrame(packet.data);
 
+        network_info.host_mac_address = packet.transmitter_address;
         network_info.total_nodes = logoff.connected_nodes;
         network_info.max_nodes = logoff.max_nodes;
 
@@ -260,6 +278,9 @@ static void HandleEAPoLPacket(const Network::WifiPacket& packet) {
         // If blocking is implemented this lock needs to be changed,
         // otherwise it might cause deadlocks
         connection_status_event->Signal();
+        if (connection_token.IsValid()) {
+            ContinueClientThread(connection_token);
+        }
     }
 }
 
@@ -397,6 +418,36 @@ void HandleAuthenticationFrame(const Network::WifiPacket& packet) {
     }
 }
 
+/// Handles the deauthentication frames sent from clients to hosts, when they leave a session
+void HandleDeauthenticationFrame(const Network::WifiPacket& packet) {
+    LOG_DEBUG(Service_NWM, "called");
+    std::unique_lock<std::recursive_mutex> hle_lock(HLE::g_hle_lock, std::defer_lock);
+    std::unique_lock<std::mutex> lock(connection_status_mutex, std::defer_lock);
+    std::lock(hle_lock, lock);
+    if (connection_status.status != static_cast<u32>(NetworkStatus::ConnectedAsHost)) {
+        LOG_ERROR(Service_NWM, "Got deauthentication frame but we are not the host");
+        return;
+    }
+    if (node_map.find(packet.transmitter_address) == node_map.end()) {
+        LOG_ERROR(Service_NWM, "Got deauthentication frame from unknown node");
+        return;
+    }
+
+    u16 node_id = node_map[packet.transmitter_address];
+    auto node = std::find_if(node_info.begin(), node_info.end(), [&node_id](const NodeInfo& info) {
+        return info.network_node_id == node_id + 1;
+    });
+    ASSERT(node != node_info.end());
+
+    connection_status.node_bitmask &= ~(1 << node_id);
+    connection_status.changed_nodes |= 1 << node_id;
+    connection_status.total_nodes--;
+
+    network_info.total_nodes--;
+    node_info.erase(node);
+    connection_status_event->Signal();
+}
+
 static void HandleDataFrame(const Network::WifiPacket& packet) {
     switch (GetFrameEtherType(packet.data)) {
     case EtherType::EAPoL:
@@ -423,6 +474,9 @@ void OnWifiPacketReceived(const Network::WifiPacket& packet) {
     case Network::WifiPacket::PacketType::Data:
         HandleDataFrame(packet);
         break;
+    case Network::WifiPacket::PacketType::Deauthentication:
+        HandleDeauthenticationFrame(packet);
+        break;
     }
 }
 
@@ -435,13 +489,22 @@ void OnWifiPacketReceived(const Network::WifiPacket& packet) {
  *      1 : Result of function, 0 on success, otherwise error code
  */
 static void Shutdown(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x03, 0, 0);
 
-    // TODO(purpasmart): Verify return header on HW
+    if (auto room_member = Network::GetRoomMember().lock())
+        room_member->Unbind(wifi_packet_received);
 
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    // TODO(B3N30): Check on HW if Shutdown signals those events
+    for (auto bind_node : channel_data) {
+        bind_node.second.event->Signal();
+    }
+    channel_data.clear();
 
-    LOG_WARNING(Service_NWM, "(STUBBED) called");
+    recv_buffer_memory.reset();
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+    rb.Push(RESULT_SUCCESS);
+    LOG_DEBUG(Service_NWM, "called");
 }
 
 /**
@@ -494,7 +557,6 @@ static void RecvBeaconBroadcastData(Interface* self) {
 
     Memory::WriteBlock(current_buffer_pos, &data_reply_header, sizeof(BeaconDataReplyHeader));
     current_buffer_pos += sizeof(BeaconDataReplyHeader);
-
     // Write each of the received beacons into the buffer
     for (const auto& beacon : beacons) {
         BeaconEntryHeader entry{};
@@ -560,6 +622,16 @@ static void InitializeWithVersion(Interface* self) {
 
     ASSERT_MSG(recv_buffer_memory->size == sharedmem_size, "Invalid shared memory size.");
 
+    if (auto room_member = Network::GetRoomMember().lock()) {
+        wifi_packet_received = room_member->BindOnWifiPacketReceived(OnWifiPacketReceived);
+    } else {
+        LOG_ERROR(Service_NWM, "Network isn't initalized");
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        // TODO(B3N30): Find the correct error code and return it;
+        rb.Push<u32>(-1);
+        return;
+    }
+
     {
         std::lock_guard<std::mutex> lock(connection_status_mutex);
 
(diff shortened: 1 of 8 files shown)


src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -237,8 +255,10 @@ static void HandleEAPoLPacket(const Network::WifiPacket& packet) {
         // Some games require ConnectToNetwork to block, for now it doesn't
         // If blocking is implemented this lock needs to be changed,
         // otherwise it might cause deadlocks
-        std::lock_guard<std::recursive_mutex> lock(HLE::g_hle_lock);
         connection_status_event->Signal();
+        if (connection_token.IsValid()) {
```
Subv: Why wouldn't this token be valid here?

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -361,13 +413,22 @@ void OnWifiPacketReceived(const Network::WifiPacket& packet) {
...
  */
 static void Shutdown(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x03, 0, 0);
+
+    if (auto room_member = Network::GetRoomMember().lock())
+        room_member->Unbind(wifi_packet_received);
 
-    // TODO(purpasmart): Verify return header on HW
+    // TODO(B3N30): Check on HW if Shutdown signals those events
```
Subv: This should be checked before merging.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -484,22 +544,31 @@ static void InitializeWithVersion(Interface* self) {
 
     ASSERT_MSG(recv_buffer_memory->size == sharedmem_size, "Invalid shared memory size.");
 
+    if (auto room_member = Network::GetRoomMember().lock()) {
+        wifi_packet_received = room_member->BindOnWifiPacketReceived(OnWifiPacketReceived);
+    } else {
+        LOG_ERROR(Service_NWM, "Network isn't initalized");
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        // TODO(B3N30): Find the correct error code and return it;
```
Subv: There is no "correct error code" here, just log it, and ignore the error

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -484,22 +544,31 @@ static void InitializeWithVersion(Interface* self) {
...
+
     {
         std::lock_guard<std::mutex> lock(connection_status_mutex);
 
         // Reset the connection status, it contains all zeros after initialization,
         // except for the actual status value.
         connection_status = {};
         connection_status.status = static_cast<u32>(NetworkStatus::NotConnected);
+        node_info.clear();
+        node_info.push_back(NodeInfo{});
```
Subv: why?

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -671,26 +739,79 @@ static void BeginHostingNetwork(Interface* self) {
...
-        std::lock_guard<std::mutex> lock(connection_status_mutex);
-
-        // TODO(Subv): Check if connection_status is indeed reset after this call.
-        connection_status = {};
-        connection_status.status = static_cast<u8>(NetworkStatus::NotConnected);
+    // Only a host can destroy
+    std::lock_guard<std::mutex> lock(connection_status_mutex);
+    if (connection_status.status != static_cast<u8>(NetworkStatus::ConnectedAsHost)) {
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(RESULT_SUCCESS);
```
Subv: This returns error 0xC8A113EA if status != ConnectedAsHost

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -671,26 +739,79 @@ static void BeginHostingNetwork(Interface* self) {
...
+        LOG_WARNING(Service_NWM, "called with status %u", connection_status.status);
+        return;
     }
+
+    // TODO(B3N30): Send 3 Deauth packets
+
+    u16_le tmp_node_id = connection_status.network_node_id;
```
Subv: This is unnecessary, the host is always network node id 1.
Was it verified on hardware that the network_node_id is persisted after the network is destroyed?

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -671,26 +739,79 @@ static void BeginHostingNetwork(Interface* self) {
...
+    connection_status = {};
+    connection_status.status = static_cast<u32>(NetworkStatus::NotConnected);
+    connection_status.network_node_id = tmp_node_id;
     connection_status_event->Signal();
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
 
+    // TODO(B3N30): HW test if events get signaled here.
+    for (auto bind_node : channel_data) {
```
Subv: This must be hw tested before merging.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -671,26 +739,79 @@ static void BeginHostingNetwork(Interface* self) {
...
     rb.Push(RESULT_SUCCESS);
 
-    LOG_WARNING(Service_NWM, "called");
+    using Network::WifiPacket;
+    WifiPacket deauth;
+    {
+        std::lock_guard<std::mutex> lock(connection_status_mutex);
+        if (connection_status.status == static_cast<u32>(NetworkStatus::ConnectedAsHost)) {
+            // A real 3ds makes strange things here. We do the same
+            u16_le tmp_node_id = connection_status.network_node_id;
```
Subv: This path should return 0xC8A113EA, the real UDS service completely mangles the internal state when DisconnectNetwork is called when connected as a host (It memsets the list of connected clients to 0 among other things).

Also, this tmp_node_id thing is unnecessary, the host is always network node id 1.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -671,26 +739,79 @@ static void BeginHostingNetwork(Interface* self) {
...
+            connection_status = {};
+            connection_status.status = static_cast<u32>(NetworkStatus::ConnectedAsHost);
+            connection_status.network_node_id = tmp_node_id;
+            LOG_DEBUG(Service_NWM, "called as a host");
+            return;
+        }
+        u16_le tmp_node_id = connection_status.network_node_id;
```
Subv: Is it hwtested that the connection_status retains the network node id after disconnecting?

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -671,26 +739,79 @@ static void BeginHostingNetwork(Interface* self) {
...
+        connection_status = {};
+        connection_status.status = static_cast<u32>(NetworkStatus::NotConnected);
```
Subv: Are you sure this is the state that DisconnectNetwork sets? from RE i see state 8 is set, this should be confirmed by hwtest.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -671,26 +739,79 @@ static void BeginHostingNetwork(Interface* self) {
...
+        // TODO(B3N30): Add disconnect reason
+        deauth.data = {};
+        deauth.destination_address = network_info.host_mac_address;
+        deauth.type = WifiPacket::PacketType::Deauthentication;
+    }
+
+    SendPacket(deauth);
+
+    // TODO(B3N30): Check on HW if Shutdown signals those events
+    for (auto bind_node : channel_data) {
```
Subv: This must be hwtested before merging.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -800,6 +921,55 @@ static void GetChannel(Interface* self) {
...
+    StartConnectionSequence(network_info.host_mac_address);
+
+    connection_token =
+        SleepClientThread("uds::ConnectToNetwork", [](Kernel::SharedPtr<Kernel::Thread> thread) {
+            VAddr address = thread->GetCommandBufferAddress();
+            std::array<u32, IPC::COMMAND_BUFFER_LENGTH> buffer;
+            IPC::RequestBuilder rb(buffer.data(), 0x1E, 1, 0);
+            // TODO(B3N30): Add error handling for host full and timeout
+            rb.Push(RESULT_SUCCESS);
+            Memory::WriteBlock(address, &*thread->owner_process, *buffer.data());
```
Subv: I'm pretty sure you're using the wrong overload of WriteBlock here.
You should be using 
`void WriteBlock(const Kernel::Process& process, const VAddr dest_addr, const void* src_buffer,
                size_t size);`

src/core/hle/service/nwm/nwm_uds.cpp
B3n30: It's not implemented yet, but ContinueClientThread(connection_token); could also get called if a timeout happens, and thus the token can be invalid which will cause an ASSERT.

I didn't add a timeout yet, since i couldn't find the time a 3ds has for that.

src/core/hle/service/nwm/nwm_uds.cpp
B3n30: Yes i check with a homebrew how connection_status looks after that

src/core/hle/service/nwm/nwm_uds.cpp
B3n30: Yes

src/core/hle/service/nwm/nwm_uds.cpp
B3n30: thats what i got from my HW-Test: 
`-- Connected to network.
constatus:
status=0x9
-- Attempting tp disconnect.
-- Disconnected.
constatus:
status=0x3
`

src/core/hle/service/nwm/nwm_uds.cpp
B3n30: Because node_info always holds one default entry in node_info[0] (at least according to my research)

Subv: The self node? Then add it directly (current_node) instead of adding a dummy NodeInfo

wwylele:
Just to keep the status update: the whole NWM_UDS service is under transformation to service framework, and this PR is going to be replaced by a new one.

src/core/hle/service/nwm/nwm_uds.cpp
```diff
@@ -540,7 +540,8 @@ void NWM_UDS::RecvBeaconBroadcastData(Kernel::HLERequestContext& ctx) {
         out_buffer.Write(&entry, offset, sizeof(BeaconEntryHeader));
         offset += sizeof(BeaconEntryHeader);
         const unsigned char* beacon_data = beacon.data.data();
-        out_buffer.Write(const_cast<void*>(static_cast<const void*>(beacon_data)), offset, beacon.data.size());
+        out_buffer.Write(const_cast<void*>(static_cast<const void*>(beacon_data)), offset,
```
Subv: Please change Write to accept a const pointer as input and remove the const_cast

wwylele:
Review status: 0 of 7 files reviewed at latest revision, 18 unresolved discussions, some commit checks failed.

---

*[src/core/hle/service/nwm/nwm_uds.cpp, line 543 at r7](https://reviewable.io:443/reviews/citra-emu/citra/3070#-Ky7Y6wEZwxPAJL5dVCs-r7-543:-L42w7P6pgfyK6cvTxoc:bv8b9ex) ([raw file](https://github.com/citra-emu/citra/blob/f88fca471fda62f98885eab9c35c4a1cc87c47e2/src/core/hle/service/nwm/nwm_uds.cpp#L543)):*

Please change Write to accept a const pointer as input and remove the const_cast
</blockquote></details>

Write is already accepting const ptr now

---

*[src/core/hle/service/nwm/nwm_uds.cpp, line 513 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3070#-L42v6vKBlcUXfOxyfBg:-L42v6vLDaK0MQAnG0XE:b2u7omv) ([raw file](https://github.com/citra-emu/citra/blob/1941e645d7dbaf532a97d08b5b2bac2be4b7dde8/src/core/hle/service/nwm/nwm_uds.cpp#L513)):*
> ```cpp
>     u32 wlan_comm_id = rp.Pop<u32>();
>     u32 id = rp.Pop<u32>();
>     Kernel::Handle input_handle = rp.PopHandle();
> ```

`PopObject();`. Also it would be nice to add some comment that this is an event handle that is never used.

---

*[src/core/hle/service/nwm/nwm_uds.cpp, line 555 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3070#-L42vnOot00LrUMYPdBw:-L42vnOot00LrUMYPdBx:bqwzgb8) ([raw file](https://github.com/citra-emu/citra/blob/1941e645d7dbaf532a97d08b5b2bac2be4b7dde8/src/core/hle/service/nwm/nwm_uds.cpp#L555)):*
> ```cpp
> 
>     IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
>     rb.Push(RESULT_SUCCESS);
> ```

Need to Push the mapped buffer in response

---

*[src/core/hle/service/nwm/nwm_uds.cpp, line 666 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3070#-L42wgJ6uzgLZjaGRPXh:-L42wgJ7pQpcJgVD8rbm:bp0yk0w) ([raw file](https://github.com/citra-emu/citra/blob/1941e645d7dbaf532a97d08b5b2bac2be4b7dde8/src/core/hle/service/nwm/nwm_uds.cpp#L666)):*
> ```cpp
>         rb.Push(ResultCode(ErrorDescription::NotAuthorized, ErrorModule::UDS,
>                            ErrorSummary::WrongArgument, ErrorLevel::Usage));
>         LOG_DEBUG(Service_NWM, "data_channel = %d, bind_node_id = %d", data_channel, bind_node_id);
> ```

We usually use at least WARNING log level for error cases. Same for cases below

---

*[src/core/hle/service/nwm/nwm_uds.cpp, line 970 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3070#-L42xTbLIVU7xeGDqIOZ:-L42xTbMbo-62zUb-2ZT:b9ej82n) ([raw file](https://github.com/citra-emu/citra/blob/1941e645d7dbaf532a97d08b5b2bac2be4b7dde8/src/core/hle/service/nwm/nwm_uds.cpp#L970)):*
> ```cpp
>     u32 max_out_buff_size = rp.Pop<u32>();
> 
>     u32 buff_size = std::min<u32>(max_out_buff_size_aligned, 0x172) << 2;
> ```

Better to add a comment stating that the magic number 0x172 is hard-coded in NWM module and we don't know much meaning of it.

---

*[src/core/hle/service/nwm/nwm_uds.cpp, line 1130 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3070#-L42yg12XIuLCy6axCE_:-L42yg12XIuLCy6axCEa:b-8t3uqy) ([raw file](https://github.com/citra-emu/citra/blob/1941e645d7dbaf532a97d08b5b2bac2be4b7dde8/src/core/hle/service/nwm/nwm_uds.cpp#L1130)):*
> ```cpp
>                 encrypted_data0_buffer.size());
>     std::memcpy(beacon_data.data() + encrypted_data0_buffer.size(),
>                 encrypted_data1_buffer.data() + 4, encrypted_data1_buffer.size());
> ```

Read out-of-range of `encrypted_data0_buffer` and `encrypted_data1_buffer`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3070)*

wwylele:
Here is my second review pass and I guess these are all I can review. 

All file scope static variables in nwm_uds should be moved into class member, but I guess we can do this in a follow up PR.

---

Review status: 0 of 7 files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---

*[src/core/hle/service/nwm/nwm_uds.cpp, line 519 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3070#-L4ShrxB82xGQD-i7gs4:-L4ShrxCnIzdefLHBQ94:b-7eaqcc) ([raw file](https://github.com/citra-emu/citra/blob/1941e645d7dbaf532a97d08b5b2bac2be4b7dde8/src/core/hle/service/nwm/nwm_uds.cpp#L519)):*
> ```cpp
> 
>     size_t offset = sizeof(BeaconDataReplyHeader);
>     u32 total_size = sizeof(BeaconDataReplyHeader);
> ```

I wonder if variable `offset` and `total_size` always have the same value below and can be merged into one variable to produce cleaner code.

---

*[src/core/hle/service/nwm/nwm_uds.cpp, line 1084 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3070#-L4SjHH5gRHMyGAg7w91:-L4SjHH5gRHMyGAg7w92:b-442au2) ([raw file](https://github.com/citra-emu/citra/blob/1941e645d7dbaf532a97d08b5b2bac2be4b7dde8/src/core/hle/service/nwm/nwm_uds.cpp#L1084)):*
> ```cpp
>     u32 size = rp.Pop<u32>();
> 
>     const std::vector<u8> address = rp.PopStaticBuffer();
> ```

probably should rename `address` to something like `application_data`

---

*[src/core/hle/service/nwm/nwm_uds.h, line 339 at r8](https://reviewable.io:443/reviews/citra-emu/citra/3070#-L4SjoJbMKusqYQ9tgxY:-L4SjoJc0_smwGUbOknM:b9nsxtc) ([raw file](https://github.com/citra-emu/citra/blob/1941e645d7dbaf532a97d08b5b2bac2be4b7dde8/src/core/hle/service/nwm/nwm_uds.h#L339)):*
> ```objc
>      *  Outputs:
>      *      0 : Return header
>      *      1 : Result of function, 0 on success, otherwise error code
> ```

2, 3: output buffer return descriptor & ptr

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3070#-:-L4SkCcUIoTlpBdIBISd:b-y26sr5)*

wwylele:
LGTM the part I commented. What is the status of @Subv's comments?

---

Review status: 0 of 7 files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3070#-:-L5QVPHqFLRcPWPbmA3r:b-rll1ax)*

PR #3073, opened 2017-11-04
Citra-qt: Add multiplayer ui

jroweboy:
## Design Goals

Multiplayer works by emulating local wireless connections and serving the packets to others across the internet. Since most games aren't designed to support large groups of people in the same location, the goal of this design is to focus on smaller rooms of people that are playing a single game instead of larger rooms where people play whatever game they feel like. The design is meant to encourage the idea of "I want to play X game with others" and to help guide them to others that also want to play that game.

With this in mind, the room is limited to 16 people in order to encourage many smaller rooms instead of a few larger rooms. The lobby also allows users to filter by games that I own, hide full rooms from the lobby, and other filtering options. Inside the room, there is a chat system which is only meant to be simple enough to provide bare bones communication between players (useful for things like helping users determine game mode to play and other coordination tasks)

Due to limitations on time, I decided that it would be best to focus on the use case listed above, of random people playing together. Its easier to get a few dedicated rooms set up that people can join, rather than having people host their own rooms and having them deal with port forwarding and so on. There wasn't enough time to add support for NAT traversal for the first release, so prioritizing players using dedicated rooms through the lobby was the plan.

In summary, the primary goal of the first release of the UI is meant to help someone find a group to play a specific game with. From there, secondary goals are to let people host their own rooms, direct connect, and so on.

## Screenshots

Lobby: 
lobby

Features of lobby:
 * Easy to change the nickname you want to connect the room with.
 * Filter rooms by search (searches on room name, preferred game and host name), games I own (by game id so it doesn't handle region differences currently), and Hide full games (max players)
 * Sortable columns
 * Doesn't hammer citra services by automatically refreshing (intended)

Rooms:
connected as client

Features of Room:
  * Simple. Has only player listing and chat.
  * Disconnect is different from close window.

Create Room:
create room

Features of create room:
 * Simple. Designed to put the "Player list and Chat" below it once the room is created, but that had to be scrapped from this release due to technical issues with updating options after the room hosting is started.
 * Fields like nickname and room name are stored in uisettings.

Direct Connect:
direct connect

Features of direct connect:
 * Ready to support NAT traversal (TODO: remove text about traversal server)
 * Fields are stored in uisettings.

## Code Design

**TODO** (sorry ran outta time writing this again!)

## TODO

- [x] Remove text about traversal server in direct connect (since that didn't make the first release)
- [x] Add "Connected" text to the status bar next to the connected icon (along with a tool tip telling people that they can double click it to open the room)
- [x] Fix `QObject::connect: signal not found in Lobby` error
- [ ] ~~Make the multiplayer windows not parented so you can minimize them independently~~ Won't do

## Follow up changes

- [ ] NAT traversal support to make self hosting more useful
- [ ] Integrated NAT feedback to help people diagnose if they can host
- [x] ~~make citra faster~~

Also major apologies for the single commit. I will split it up if reviewing file by file is proving to be too much

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3073)


BynariStar:
As discussed on Discord:

- After closing the multiplayer window (but not disconnecting) and reopening the lobby, it prompts disconnection while trying to reopen the room you're currently still in. Should probably change this to just open the room.
- Should probably change the "Chat" button to something more along the lines of "Current Room" to avoid confusion
- The connection status icon IS TOO DAMN SMALL (scaling issue with HiDPI screens)

esQmo:
Please add a back button to the chat client beacause when you select a room and go in, there is no way to go back to the room browser, you have to close the chat client windows and then re-open it again from the Multiplayer tab. to be able to, say, select another rom. Or you can make it open in a new window instead of that same.

src/citra_qt/main.cpp
```diff
@@ -714,6 +758,27 @@ void GMainWindow::OnMenuRecentFile() {
...
+    ui.action_Chat->setDisabled(true);
+
+    ChangeRoomState();
+}
+
+void GMainWindow::OnAnnounceFailed(const Common::WebResult& result) {
+    announce_multiplayer_session->Stop();
+    QMessageBox::warning(
+        this, tr("Error"),
+        tr("Announcing the room failed. \nThe room will not get listed publicy. \nError: ") +
```
BreadFish64: listed publicly

dist/icons/license.md
```diff
@@ -0,0 +1,7 @@
+The icons in this folder have the following licenses:
+checked.png       Free for non-commercial use
+connect.png       CC BY-ND 3.0 (aquired from https://icons8.com)
+disconnect.png    CC BY-ND 3.0 (aquired from https://icons8.com)
+failed.png        Free for non-commercial use
+lock.png         CC BY-ND 3.0 (aquired from https://icons8.com)
```
baka0815: Indentation

src/citra_qt/main.cpp
baka0815: Also the spaces before the \n are not really necessary.

src/citra_qt/main.ui
```diff
@@ -206,6 +221,43 @@
...
+     <property name="enabled">
+       <bool>false</bool>
+     </property>
+     <property name="text">
+       <string>Close Room</string>
+     </property>
+   </action>
+  <action name="action_Connect_To_Room">
+   <property name="text">
+    <string>Direct Connect to Room</string>
```
baka0815: What's the "Direct" here for? Isn't "Connect to Room" enough?

Is the capitalization correct the way it is? Shouldn't "Room" be lower case?
So "Direct[ly] connect to room" or "Connect to room"?
The same comment applies to "Create Room" ("Create room") and "Browse Public Game Lobby" ("Browse public game lobby").

src/citra_qt/main.cpp
```diff
@@ -714,6 +758,27 @@ void GMainWindow::OnMenuRecentFile() {
...
 }
 
+void GMainWindow::OnNetworkStateChanged(const Network::RoomMember::State& state) {
+    if (state == Network::RoomMember::State::Joined) {
+        network_status->setPixmap(QPixmap(":/icons/connected.png"));
+        ui.action_Chat->setEnabled(true);
+        return;
+    }
+    network_status->setPixmap(QPixmap(":/icons/disconnected.png"));
+    ui.action_Chat->setDisabled(true);
```
baka0815: I'm not into Qt enough to know, but you're using "setDisabled(true)" here and "setEnabled(false)" in another place.
Shouldn't that all be using "setEnabled(true)" or "setEnabled(false)" respectively?

At least I find it confusing.

src/citra_qt/multiplayer/lobby.cpp
```diff
@@ -0,0 +1,314 @@
...
+    }
+    ui->chat->setDisabled(true);
+}
+
+Lobby::~Lobby() {}
+
+const QString Lobby::PasswordPrompt() {
+    bool ok;
+    const QString text =
+        QInputDialog::getText(this, tr("Password Required to Join"), tr("Password:"),
```
baka0815: Password **r**equired to **j**oin

src/citra_qt/multiplayer/lobby.h
```diff
@@ -0,0 +1,118 @@
...
+     * returned from web services, `LobbyRefreshed` will be signalled
+     */
+    void RefreshLobby();
+
+private slots:
+    /**
+     * Pulls the list of rooms from network and fills out the lobby model with the results
+     */
+    void OnRefreshLobby();
+    /**
```
baka0815: Missing newline?

src/citra_qt/multiplayer/lobby.h
```diff
@@ -0,0 +1,118 @@
...
+    void StateChanged(const Network::RoomMember::State&);
+
+private:
+    /**
+     * Removes all entries in the Lobby before refreshing.
+     */
+    void ResetModel();
+
+    /**
+     * Prompts for a password. Returns an empty QString in the user either did not provide a
```
baka0815: QString **if** the user

src/citra_qt/multiplayer/validation.h
```diff
@@ -0,0 +1,28 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <QRegExp>
+#include <QValidator>
+
+namespace Validation {
+/// room name can be alphanumeric and " " "_" "." and "-"
+static const QRegExp room_name_regex("^[a-zA-Z0-9._- ]+$");
```
baka0815: Is "." correct here? Shouldn't it be "\\."?

src/citra_qt/multiplayer/validation.h
```diff
@@ -0,0 +1,28 @@
...
+static const QValidator* room_name = new QRegExpValidator(room_name_regex);
+
+/// nickname can be alphanumeric and " " "_" "." and "-"
+static const QRegExp nickname_regex("^[a-zA-Z0-9._- ]+$");
```
baka0815: Same as room name.

jroweboy: Thanks for the review so far, but I would love it if you batched the comments into a single review so i don't get flooded with emails about it :)

(Also, yes the regex is correct. Its a character class, so you don't need to escape the period https://www.regular-expressions.info/refcharclass.html)

baka0815: Sorry about the noise, I'll keep that in mind for the next time.

src/citra_qt/main.h
```diff
@@ -48,6 +62,9 @@ class GMainWindow : public QMainWindow {
 public:
     void filterBarSetChecked(bool state);
     void UpdateUITheme();
+    void ChangeRoomState();
+
+    GameList* game_list;
```
MerryMage: Duplicate `game_list`.

MerryMage:
RE CI: Ignore continuous-integration/travis-ci/push and continuous-integration/appveyor/branch, that was my error.

B3n30:
Reviewed 8 of 35 files at r1, 1 of 3 files at r2, 8 of 8 files at r3.
Review status: 17 of 35 files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

*[src/citra_qt/multiplayer/chat_room.cpp, line 154 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3073#-L3YF3bKARxMRT7XW5io:-L3YF3bKARxMRT7XW5ip:by27tax) ([raw file](https://github.com/citra-emu/citra/blob/9c482ee1fa9e373d166b00fede5ad13bf5402827/src/citra_qt/multiplayer/chat_room.cpp#L154)):*
> ```cpp
>             return;
>         }
>         auto player = std::distance(members.begin(), it);
> ```

That way the color of the chat messages will change as soon as a player disconnects

---

*[src/citra_qt/multiplayer/chat_room.cpp, line 178 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3073#-L3YFHd2i33dShHsYfdV:-L3YFHd2i33dShHsYfdW:b2ss7xm) ([raw file](https://github.com/citra-emu/citra/blob/9c482ee1fa9e373d166b00fede5ad13bf5402827/src/citra_qt/multiplayer/chat_room.cpp#L178)):*
> ```cpp
>                 LOG_INFO(Network, "Chat message received from unknown player");
>             }
>             auto player = std::distance(members.begin(), it);
> ```

see comment above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3073)*

B3n30:
Reviewed 26 of 35 files at r1.
Review status: all files reviewed at latest revision, 19 unresolved discussions, some commit checks broke.

---

*[src/citra_qt/multiplayer/client_room.cpp, line 13 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3073#-L3e0MwbFbKhP34KMmJV:-L3e0MwbFbKhP34KMmJW:bqx6ynr) ([raw file](https://github.com/citra-emu/citra/blob/bbf148ffde8b29c2b265c2e2c17c85d297841f44/src/citra_qt/multiplayer/client_room.cpp#L13)):*
> ```cpp
> #include <QTime>
> #include <QtConcurrent/QtConcurrentRun>
> ```

remove blank line

---

*[src/citra_qt/multiplayer/direct_connect.ui, line 85 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3073#-L3e2BgNP4pXYv5EYGpt:-L3e2BgNP4pXYv5EYGpu:bvss8fd) ([raw file](https://github.com/citra-emu/citra/blob/bbf148ffde8b29c2b265c2e2c17c85d297841f44/src/citra_qt/multiplayer/direct_connect.ui#L85)):*
> ```XML
>              <widget class="QLineEdit" name="ip">
>               <property name="toolTip">
>                <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;IPv4 address of the host&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
> ```

Is it necessary to put these strings into that ugly html code?

---

*[src/citra_qt/multiplayer/host_room.cpp, line 51 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3073#-L3e2khcTR31NPnQRht9:-L3e2khcTR31NPnQRhtA:bs8pe6y) ([raw file](https://github.com/citra-emu/citra/blob/bbf148ffde8b29c2b265c2e2c17c85d297841f44/src/citra_qt/multiplayer/host_room.cpp#L51)):*
> ```cpp
>     ui->port->setText(UISettings::values.room_port);
>     ui->max_player->setValue(UISettings::values.max_player);
>     int index = ui->host_type->findData(UISettings::values.host_type);
> ```

host type isn't restored properly. Not sure why though

---

*[src/citra_qt/multiplayer/lobby.cpp, line 175 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3073#-L3e3YyY8Od_9YZyLDKm:-L3e3YyY8Od_9YZyLDKn:bmdshz2) ([raw file](https://github.com/citra-emu/citra/blob/bbf148ffde8b29c2b265c2e2c17c85d297841f44/src/citra_qt/multiplayer/lobby.cpp#L175)):*
> ```cpp
>         ui->refresh_list->setText(tr("Refreshing"));
>     } else {
>         // TODO(jroweboy): Display an error box about network couldn't be started
> ```

about announce couldn't get started.

---

*[src/citra_qt/util/clickable_label.cpp, line 1 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3073#-L3e4GqaWDXIr2Y6MJe_:-L3e4GqaWDXIr2Y6MJea:b-nzk1ak) ([raw file](https://github.com/citra-emu/citra/blob/bbf148ffde8b29c2b265c2e2c17c85d297841f44/src/citra_qt/util/clickable_label.cpp#L1)):*
> ```cpp
> #include "util/clickable_label.h"
> ```

Copyright header?

---

*[src/citra_qt/util/clickable_label.h, line 1 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3073#-L3e4L-yCMGykAYrL1QL:-L3e4L-yCMGykAYrL1QM:bxo2i1) ([raw file](https://github.com/citra-emu/citra/blob/bbf148ffde8b29c2b265c2e2c17c85d297841f44/src/citra_qt/util/clickable_label.h#L1)):*
> ```objc
> #ifndef CLICKABLELABEL_H
> ```

#pragma once
Copyright header

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3073)*

adityaruplaha:
When will this be merged into nightly? Its been in canary for quite some time.

BreadFish64:
@adityalruplaha this is the last piece of the puzzle for multiplayer, all the other parts of multiplayer need to merged before this

CaptV0rt3x:
@jroweboy  Please check out #3432 . According to @B3n30 , it is caused by this PR.

FearlessTobi:
Just noticed that you are able to chose a name with spaces at the end. They should be trimmed away imo.

jroweboy:
A long time ago, I had pestered neobrain for some feedback about the UI that I planned to apply in some shape or form to this before merging. Here was the conversation from the logs so i can reference later: https://pastebin.com/d6y3WB2f

src/citra_qt/multiplayer/lobby.cpp
jroweboy: This is the title of the window so the capitalization is correct.

src/citra_qt/main.ui
jroweboy: The current wording is consistent with the existing menu items. All of the menu items capitalize the first letter of each word.

dist/icons/license.md
```diff
@@ -0,0 +1,7 @@
+The icons in this folder have the following licenses:
+checked.png       Free for non-commercial use
+connect.png       CC BY-ND 3.0 (aquired from https://icons8.com)
```
baka0815: The filename is "connected.png"

dist/icons/license.md
```diff
@@ -0,0 +1,7 @@
...
+disconnect.png    CC BY-ND 3.0 (aquired from https://icons8.com)
```
baka0815: The filename is "disconnected.png"

chris062689:
What are we planning on doing from a UI perspective with multiple netVersions?

We will likely run into this scenario when this gets merged into nightly, and we push out a new net version for Canary builds. 

Should we grey out servers that run a different net version? 
What should we do from a UI perspective?

chris062689:
The version mismatch error is a bit cryptic. 
`you are using a different version of Citra-Local-Wifi???`

Perhaps a better dialog message would be
`ERROR: VERSION_MISMATCH You are attempting to connect to a room running a different multiplayer version. Please ensure that you are running the latest version of Citra.`

adityaruplaha:
Yes. That would also discourage unofficial builds, as they will most likely
be outdated very fast.

Adityarup Laha

Schplee:
*Recommended by Flame Sage to post here*

* Perhaps the lobby should default sort by player count, so that users can have an easier time seeing which lobbies are populated. With how many lobbies exist now, it can become a hassle to sort without pressing a button.

* I'm not sure if it's already planned, but currently the tabs' size can't be changed: https://i.imgur.com/m1Raj4p.png

jroweboy:
@Schplee Don't really like that as the default sort. Sort by name seems so much more intuitive as you want to try to find lobbies that are close to you so you have the best latency usually.

wasn't planning on making the tabs resizable, but i can do that

also note to self: add lobby host username in the list of things that can be filtered by. no reason not to add that imo

jroweboy:
Conflict fixed. Added a little "Connected" status bar message as well as migrated to using the new themed icons so if someone wants to make me some dark themed versions for the connection/disconnected icons, go ahead, and we can stick them in there. I plan to fix up the UI state stuff soon (TM) so we can merge it so others feel more comfortable fixing issues with this...

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    /// Format the message using the players color
+    QString GetPlayerChatMessage(u16 player) const {
+        auto color = player_color[player % 16];
+        return QString("[%1] <font color='%2'>&lt;%3&gt;</font> %4")
+            .arg(timestamp, color, nickname.toHtmlEscaped(), message.toHtmlEscaped());
+    }
+
+private:
+    ChatMessage() {}
```
lioncash: This can likely just be removed if the intention is to not allow it to be used, since another constructor is already provided.

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+    const QList<QString> player_color = {
```
lioncash: This should probably be `static const`. It can likely also just be a std::array of const char*'s. There's no real need to allocate it all on the heap in Qt strings, when it's just being printed out.

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include "citra_qt/multiplayer/message.h"
+#include "common/logging/log.h"
+#include "core/announce_multiplayer_session.h"
+#include "ui_chat_room.h"
+
+class ChatMessage {
+public:
+    explicit ChatMessage(const Network::ChatEntry& chat, QTime ts = {}) {
+        /// Convert the time to their default locale defined format
+        static QLocale locale;
```
lioncash: Is there a reason this just doesn't construct the default instance on each invocation? Is it really that performance intensive?

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+        message = msg;
+    }
+
+    QString GetSystemChatMessage() const {
+        return QString("[%1] <font color='%2'><i>%3</i></font>")
+            .arg(timestamp, system_color, message);
+    }
+
+private:
+    const QString system_color = "#888888";
```
lioncash: This can be: `static constexpr const char system_color[] = "#888888"`

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+    QString timestamp;
+    QString message;
+};
+
+ChatRoom::ChatRoom(QWidget* parent) : ui(new Ui::ChatRoom) {
```
lioncash: make_unique

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+        }
+        auto player = std::distance(members.begin(), it);
+        ChatMessage m(chat);
+        AppendChatMessage(m.GetPlayerChatMessage(player));
+    }
+}
+
+void ChatRoom::OnSendChat() {
+    if (auto room = Network::GetRoomMember().lock()) {
+        if (room->GetState() == Network::RoomMember::State::Joined) {
```
lioncash: This if statement can be an early return, unindenting all the code within it.

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+            ui->chat_message->clear();
+        }
+    }
+}
+
+void ChatRoom::SetPlayerList(const Network::RoomMember::MemberList& member_list) {
+    // TODO(B3N30): Remember which row is selected
+    player_list->removeRows(0, player_list->rowCount());
+    for (const auto& member : member_list) {
+        if (member.nickname == "")
```
lioncash: `if (member.nickname.empty())`

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+            continue;
+        QList<QStandardItem*> l;
+        std::vector<std::string> elements = {member.nickname, member.game_info.name};
+        for (auto& item : elements) {
```
lioncash: `const auto&`

src/citra_qt/multiplayer/client_room.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include "citra_qt/game_list_p.h"
+#include "citra_qt/multiplayer/client_room.h"
+#include "citra_qt/multiplayer/message.h"
+#include "common/logging/log.h"
+#include "core/announce_multiplayer_session.h"
+#include "ui_client_room.h"
+
+ClientRoomWindow::ClientRoomWindow(QWidget* parent)
+    : QDialog(parent, Qt::WindowTitleHint | Qt::WindowCloseButtonHint | Qt::WindowSystemMenuHint),
+      ui(new Ui::ClientRoom) {
```
lioncash: make_unique

src/citra_qt/multiplayer/chat_room.h
```diff
@@ -0,0 +1,57 @@
...
+
+signals:
+    void ChatReceived(const Network::ChatEntry&);
+
+private:
+    const u32 max_chat_lines = 1000;
+    void AppendChatMessage(const QString&);
+    bool ValidateMessage(const std::string&);
+    QStandardItemModel* player_list;
+    Ui::ChatRoom* ui;
```
lioncash: This should be a unique_ptr, otherwise you leak resources.

src/citra_qt/multiplayer/client_room.h
```diff
@@ -0,0 +1,38 @@
...
+    void Closed();
+    void RoomInformationChanged(const Network::RoomInformation&);
+    void StateChanged(const Network::RoomMember::State&);
+
+private:
+    void Disconnect();
+    void UpdateView();
+
+    QStandardItemModel* player_list;
+    Ui::ClientRoom* ui;
```
lioncash: This should be a unique_ptr otherwise you leak resources.

src/citra_qt/multiplayer/direct_connect.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include "citra_qt/ui_settings.h"
+#include "core/settings.h"
+#include "network/network.h"
+#include "ui_direct_connect.h"
+
+enum class ConnectionType : u8 { TRAVERSAL_SERVER, IP };
+
+DirectConnectWindow::DirectConnectWindow(QWidget* parent)
+    : QDialog(parent, Qt::WindowTitleHint | Qt::WindowCloseButtonHint | Qt::WindowSystemMenuHint),
+      ui(new Ui::DirectConnect) {
```
lioncash: make_unique

src/citra_qt/multiplayer/direct_connect.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include "citra_qt/multiplayer/client_room.h"
+#include "citra_qt/multiplayer/direct_connect.h"
+#include "citra_qt/multiplayer/message.h"
+#include "citra_qt/multiplayer/validation.h"
+#include "citra_qt/ui_settings.h"
+#include "core/settings.h"
+#include "network/network.h"
+#include "ui_direct_connect.h"
+
+enum class ConnectionType : u8 { TRAVERSAL_SERVER, IP };
```
lioncash: enum class formatting would be `TraversalServer, IP`

src/citra_qt/multiplayer/direct_connect.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+void DirectConnectWindow::Connect() {
+    ClearAllError();
+    bool isValid = true;
+    if (!ui->nickname->hasAcceptableInput()) {
+        isValid = false;
+        ShowError(NetworkMessage::USERNAME_NOT_VALID);
+    }
+    if (const auto member = Network::GetRoomMember().lock()) {
+        if (member->IsConnected()) {
```
lioncash:
```cpp
if (member->IsConnected() && !NetworkMessage::WarnDisconnect()) {
    return;
}
```

src/citra_qt/multiplayer/direct_connect.cpp
```diff
@@ -0,0 +1,132 @@
...
+    // box. Add this back in when the traversal server support is added.
+    connect(ui->connect, &QPushButton::pressed, this, &DirectConnectWindow::Connect);
+}
+
+void DirectConnectWindow::Connect() {
+    ClearAllError();
+    bool isValid = true;
+    if (!ui->nickname->hasAcceptableInput()) {
+        isValid = false;
+        ShowError(NetworkMessage::USERNAME_NOT_VALID);
```
lioncash: Should this not just return here rather than waiting until it hits the check further below?

src/citra_qt/multiplayer/direct_connect.h
```diff
@@ -0,0 +1,39 @@
...
+    void OnConnection();
+
+private:
+    void Connect();
+    void ClearAllError();
+    void BeginConnecting();
+    void EndConnecting();
+
+    QFutureWatcher<void>* watcher;
+    Ui::DirectConnect* ui;
```
lioncash: This should be a unique_ptr otherwise you leak resources

src/citra_qt/multiplayer/host_room.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include "common/logging/log.h"
+#include "core/announce_multiplayer_session.h"
+#include "core/settings.h"
+#include "ui_chat_room.h"
+#include "ui_host_room.h"
+
+HostRoomWindow::HostRoomWindow(QWidget* parent, QStandardItemModel* list,
+                               std::shared_ptr<Core::AnnounceMultiplayerSession> session)
+    : QDialog(parent, Qt::WindowTitleHint | Qt::WindowCloseButtonHint | Qt::WindowSystemMenuHint),
+      ui(new Ui::HostRoom), announce_multiplayer_session(session), game_list(list) {
```
lioncash: make_unique

src/citra_qt/multiplayer/host_room.h
```diff
@@ -0,0 +1,73 @@
...
+     */
+    void OnConnection();
+
+private:
+    void Host();
+
+    std::weak_ptr<Core::AnnounceMultiplayerSession> announce_multiplayer_session;
+    QStandardItemModel* game_list;
+    ComboBoxProxyModel* proxy;
+    Ui::HostRoom* ui;
```
lioncash: This should be a unique_ptr otherwise you leak resources.

src/citra_qt/util/clickable_label.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <QLabel>
+#include <QWidget>
+#include <Qt>
+
+class ClickableLabel : public QLabel {
+    Q_OBJECT
+
+public:
+    explicit ClickableLabel(QWidget* parent = Q_NULLPTR, Qt::WindowFlags f = Qt::WindowFlags());
```
lioncash: = nullptr

src/citra_qt/util/clickable_label.h
```diff
@@ -0,0 +1,23 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <QLabel>
+#include <QWidget>
+#include <Qt>
```
lioncash: I would remove including the main Qt header in another header file.

src/citra_qt/util/clickable_label.cpp
```diff
@@ -0,0 +1,13 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "citra_qt/util/clickable_label.h"
+
+ClickableLabel::ClickableLabel(QWidget* parent, Qt::WindowFlags f) : QLabel(parent) {}
+
+ClickableLabel::~ClickableLabel() {}
```
lioncash: This can be a defaulted destructor

src/citra_qt/multiplayer/validation.h
lioncash: static can be omitted from all of these. `const` internally links by default.

src/citra_qt/multiplayer/message.h
```diff
@@ -0,0 +1,53 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+namespace NetworkMessage {
+
+class ConnectionError {
+
+public:
+    explicit ConnectionError(const std::string& str) : err(str) {}
```
lioncash: std::string can be taken by value here and std::moved in the initializer list.

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+};
+
+Q_DECLARE_METATYPE(LobbyMember);
+
+class LobbyItemMemberList : public LobbyItem {
+public:
+    static const int MemberListRole = Qt::UserRole + 1;
+    static const int MaxPlayerRole = Qt::UserRole + 2;
+
+    LobbyItemMemberList() : LobbyItem() {}
```
lioncash: This can just be:

```cpp
LobbyItemMemberList() = default;
```

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    LobbyItemMemberList(QList<QVariant> members, u32 max_players) : LobbyItem() {
```
lioncash: `: LobbyItem()` is redundant.

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    bool operator<(const QStandardItem& other) const override {
+        return data(HostUsernameRole)
+                   .toString()
+                   .localeAwareCompare(other.data(HostUsernameRole).toString()) < 0;
+    }
+};
+
+class LobbyMember {
+public:
+    LobbyMember() {}
```
lioncash: This can be a defaulted constructor

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    LobbyMember(const LobbyMember& other) {
```
lioncash: This can just be

```cpp
LobbyMember(const LobbyMember& other) = default;
```

the copy-assignment operator should likely be defaulted as well.

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+        username = other.username;
+        title_id = other.title_id;
+        game_name = other.game_name;
+    }
+    LobbyMember(const QString username, u64 title_id, const QString game_name)
+        : username(username), title_id(title_id), game_name(game_name) {}
+    ~LobbyMember() {}
```
lioncash: This can be a defaulted destructor

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+
+class LobbyMember {
+public:
+    LobbyMember() {}
+    LobbyMember(const LobbyMember& other) {
+        username = other.username;
+        title_id = other.title_id;
+        game_name = other.game_name;
+    }
+    LobbyMember(const QString username, u64 title_id, const QString game_name)
```
lioncash: These can just be non-const and std::moved in the initializer lists.

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    }
+};
+
+class LobbyItemHost : public LobbyItem {
+public:
+    static const int HostUsernameRole = Qt::UserRole + 1;
+    static const int HostIPRole = Qt::UserRole + 2;
+    static const int HostPortRole = Qt::UserRole + 3;
+
+    LobbyItemHost() : LobbyItem() {}
```
lioncash: This can just be:

```cpp
LobbyItemHost() = default;
```

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    LobbyItemHost(QString username, QString ip, u16 port) : LobbyItem() {
```
lioncash:
```cpp
: LobbyItem()
```
is redundant.

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    }
+};
+
+class LobbyItemGame : public LobbyItem {
+public:
+    static const int TitleIDRole = Qt::UserRole + 1;
+    static const int GameNameRole = Qt::UserRole + 2;
+    static const int GameIconRole = Qt::UserRole + 3;
+
+    LobbyItemGame() : LobbyItem() {}
```
lioncash: This can just be:

```cpp
LobbyItemGame() = default;
```

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    LobbyItemGame(u64 title_id, QString game_name, QPixmap smdh_icon) : LobbyItem() {
```
lioncash:
```cpp
: LobbyItem()
```
is redundant.

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+
+    bool operator<(const QStandardItem& other) const override {
+        return data(PasswordRole).toBool() < other.data(PasswordRole).toBool();
+    }
+};
+
+class LobbyItemName : public LobbyItem {
+public:
+    static const int NameRole = Qt::UserRole + 1;
+    LobbyItemName() : LobbyItem() {}
```
lioncash: This can just be:

```cpp
LobbyItemName() = default;
```

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    LobbyItemName(QString name) : LobbyItem() {
```
lioncash: `explicit LobbyItemName`

also

```cpp
: LobbyItem()
```
is redundant.

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    GAME_NAME,
+    HOST,
+    MEMBER,
+    TOTAL,
+};
+}
+
+class LobbyItem : public QStandardItem {
+public:
+    LobbyItem() : QStandardItem() {}
```
lioncash:
```cpp
LobbyItem() = default;
```

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    LobbyItem(const QString& string) : QStandardItem(string) {}
```
lioncash:
```cpp
explicit LobbyItem
```

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    virtual ~LobbyItem() override {}
```
lioncash:
```cpp
virtual ~LobbyItem() override = default;
```

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+};
+
+class LobbyItemPassword : public LobbyItem {
+public:
+    static const int PasswordRole = Qt::UserRole + 1;
+    LobbyItemPassword() : LobbyItem() {}
```
lioncash:
```cpp
LobbyItemPassword() = default;
```

src/citra_qt/multiplayer/lobby_p.h
```diff
@@ -0,0 +1,189 @@
...
+    LobbyItemPassword(const bool has_password) : LobbyItem() {
```
lioncash:
```cpp
explicit LobbyItemPassword
```
also `: LobbyItem()` is redundant.

src/citra_qt/multiplayer/lobby.cpp
```diff
@@ -0,0 +1,314 @@
...
+#include "citra_qt/ui_settings.h"
+#include "common/logging/log.h"
+#include "core/settings.h"
+#include "network/network.h"
+#include "ui_lobby.h"
+
+Lobby::Lobby(QWidget* parent, QStandardItemModel* list,
+             std::shared_ptr<Core::AnnounceMultiplayerSession> session)
+    : QDialog(parent, Qt::WindowTitleHint | Qt::WindowCloseButtonHint | Qt::WindowSystemMenuHint),
+      ui(new Ui::Lobby), announce_multiplayer_session(session), game_list(list) {
```
lioncash: make_unique

src/citra_qt/multiplayer/lobby.cpp
```diff
@@ -0,0 +1,314 @@
...
+    if (auto member = Network::GetRoomMember().lock()) {
+        if (member->IsConnected()) {
+            ui->chat->setEnabled(true);
+            return;
+        }
+    }
+    ui->chat->setDisabled(true);
+}
+
+Lobby::~Lobby() {}
```
lioncash: This can be a defaulted destructor

src/citra_qt/multiplayer/chat_room.h
```diff
@@ -0,0 +1,57 @@
...
+    void OnSendChat();
+    void OnChatTextChanged();
+    void Disable();
+    void Enable();
+
+signals:
+    void ChatReceived(const Network::ChatEntry&);
+
+private:
+    const u32 max_chat_lines = 1000;
```
lioncash: `static constexpr` or move this into the cpp file.

src/citra_qt/main.cpp
```diff
@@ -1184,6 +1333,16 @@ void GMainWindow::closeEvent(QCloseEvent* event) {
 
     render_window->close();
 
+    // Close Multiplayer windows
```
lioncash: This should probably go within its own function.

src/citra_qt/main.cpp
```diff
@@ -1053,6 +1128,80 @@ void GMainWindow::OnCreateGraphicsSurfaceViewer() {
...
+            LOG_INFO(Frontend, "Destroying host room");
+            host_room = nullptr;
+        });
+    }
+    BringWidgetToFront(host_room);
+}
+
+void GMainWindow::OnCloseRoom() {
+    if (auto room = Network::GetRoom().lock()) {
+        if (room->GetState() == Network::Room::State::Open) {
```
lioncash: These can be inverted to unindent contained code.

jroweboy:
Multiplayer UI status:

* Moved almost all the multiplayer handling stuff from main to its own class
* Added a new feature to allow people to see who is playing what in the lobby. (Screenshot below)
* Fixed an issue where you couldn't host a room, and then close it and host another
* Found a new issue where if you try to join a full room, theres no error message. From there, you can't join the next room (you get an error message) and then the next room you try to join makes citra crash :P 
* Haven't address @lioncash 's comments yet. Will be doing that before merge though.
* Blank names seem to be allowed somehow. Need to look more into that

Light theme
citra-qt_2018-04-05_12-13-16

Dark theme
citra-qt_2018-04-05_12-14-08

FearlessTobi:
I personally don't think the "Current Players in the room" text is needed. Imo it's fairly self-explanatory what's displayed there.

daniellimws:
Why is the first column called "Password"?

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+                               [&chat](const Network::RoomMember::MemberInformation& member) {
+                                   return member.nickname == chat.nickname;
+                               });
+        if (it == members.end()) {
+            LOG_INFO(Network, "Chat message received from unknown player. Ignoring it.");
+            return;
+        }
+        auto player = std::distance(members.begin(), it);
+        ChatMessage m(chat);
+        AppendChatMessage(m.GetPlayerChatMessage(player));
```
baka0815: Doesn't that change the color of all players if one who joined earlier leaves?

src/citra_qt/multiplayer/chat_room.cpp
```diff
@@ -0,0 +1,208 @@
...
+                                   [&chat](const Network::RoomMember::MemberInformation& member) {
+                                       return member.nickname == chat.nickname;
+                                   });
+            if (it == members.end()) {
+                LOG_INFO(Network, "Chat message received from unknown player");
+            }
+            auto player = std::distance(members.begin(), it);
+            ChatMessage m(chat);
+            room->SendChatMessage(message);
+            AppendChatMessage(m.GetPlayerChatMessage(player));
```
baka0815: Same as above.
Also can't this duplicated code (``get the id of the player``) be put into it's own function?

tywald:
It seems commit 1acbea9 broke a couple of things:
1) The 'File' and 'Emulation' menus are unclickable, doesn't react when you mouseover them. Need to click on 'View', 'Multiplayer' or 'Help' menus then move the mouse towards 'Emulation' and 'File' to access them.

2) Filtering doesn't work in the public room browser.

Tested with Appveyor builds 3b170be2 and eba8953c with a fresh config.

src/citra_qt/multiplayer/chat_room.cpp
jroweboy: yes. its a hack to add *some* color consistency since theres not a persistent player id assigned by the room and i couldn't convince anyone to add one either.

jroweboy:
@daniellimws the password column shows an icon if a room has a password on it. the columns are ordered by things that affect whether or not people want to join a room. If someone is just looking to play pokemon, they need to know first if its password protected, as they can't join it if it is. From there, I somewhat arbitrarily chose the ordering of the other columns, but it seemed to be a pretty good ordering.

daniellimws:
@jroweboy The first column looks more like a button to expand the room to see more details about the room members instead of an indicator on whether it is password-protected, that's why I asked. Or is that still under work?

src/citra_qt/multiplayer/chat_room.cpp
jroweboy: We don't use unique ptrs for UI files anywhere in the application. I'm not sure if we *should* but I'm just following the practice that the rest of the code uses.

src/citra_qt/multiplayer/client_room.cpp
jroweboy: in setupUi, qt parents all of the ui elements to the parent widget, so it should be the one responsible for managing the memory as I understand it.

src/citra_qt/multiplayer/direct_connect.cpp
jroweboy: The original idea was to show form errors inline, and an early return doesn't let all validators run. Currently its just a little more annoying as you get multiple popups (one for each error), so i'm fine changing it.

jroweboy:
One other issue that probably should be investigated before merge is the search and filter options no longer seem to work. I don't know what broke them, but all the code is still there so it shouldn't be too much to fix it.

jl91569:
I'm still getting multiple error windows opening when trying to connect to a nonexistent server.

1. Connect to a server.
2. Disconnect from the server
3. Fail to connect to a server

Expected behaviour: One error window opens
Actual behaviour: 2 error windows open

Tested on citra-windows-mingw-20180409-e8ac99e

jroweboy:
I think that i've fixed everything thats holding this back.

* Lobby Filters are working again.
* The menu items for Show Current Room and Leave Room are functional
* Couldn't replicate @jl91569 issue. Maybe my leave room changes fixed it :P
* Removed the misleading instructions in Direct Connect
* Minor stylistic changes in a few of the UIs

pls do one last review and merge @wwylele

src/citra_qt/multiplayer/message.cpp
```diff
@@ -0,0 +1,59 @@
...
+    QT_TR_NOOP("Unable to find an internet connection. Check your internet settings."));
+const ConnectionError UNABLE_TO_CONNECT(
+    QT_TR_NOOP("Unable to connect to the host. Verify that the connection settings are correct."));
+const ConnectionError COULD_NOT_CREATE_ROOM(
+    QT_TR_NOOP("Creating a room failed. Please retry. Restarting Citra might be necessary."));
+const ConnectionError HOST_BANNED(
+    QT_TR_NOOP("The host of the room has banned you. Speak with the host to unban you "
+               "or try a different room."));
+const ConnectionError WRONG_VERSION(
+    QT_TR_NOOP("You are using a different version of Citra-Local-Wifi(tm) then the room "
```
BreadFish64: *than* the room, the message could be more professional too

src/citra_qt/multiplayer/message.cpp
```diff
@@ -0,0 +1,59 @@
...
+const ConnectionError ROOMNAME_NOT_VALID(
+    QT_TR_NOOP("Room name is not valid. Must be 4 to 20 alphanumeric characters."));
+const ConnectionError USERNAME_IN_USE(
+    QT_TR_NOOP("Username is already in use. Please choose another."));
+const ConnectionError IP_ADDRESS_NOT_VALID(QT_TR_NOOP("IP is not a valid IPv4 address."));
+const ConnectionError PORT_NOT_VALID(QT_TR_NOOP("Port must be a number between 0 to 65535."));
+const ConnectionError NO_INTERNET(
+    QT_TR_NOOP("Unable to find an internet connection. Check your internet settings."));
+const ConnectionError UNABLE_TO_CONNECT(
+    QT_TR_NOOP("Unable to connect to the host. Verify that the connection settings are correct."));
```
BreadFish64: this usually happens when the host doesn't have the proper port forwarded, maybe mention that?

src/citra_qt/multiplayer/message.cpp
```diff
@@ -0,0 +1,59 @@
...
+const ConnectionError COULD_NOT_CREATE_ROOM(
+    QT_TR_NOOP("Creating a room failed. Please retry. Restarting Citra might be necessary."));
+const ConnectionError HOST_BANNED(
+    QT_TR_NOOP("The host of the room has banned you. Speak with the host to unban you "
+               "or try a different room."));
+const ConnectionError WRONG_VERSION(
+    QT_TR_NOOP("You are using a different version of Citra-Local-Wifi(tm) then the room "
+               "you are trying to connect to."));
+const ConnectionError WRONG_PASSWORD(QT_TR_NOOP("Wrong password."));
```
BreadFish64: "Incorrect Password"

src/citra_qt/multiplayer/state.cpp
```diff
@@ -0,0 +1,150 @@
...
+    status_icon->setPixmap(QIcon::fromTheme("disconnected").pixmap(16));
+    status_text->setText(tr("Not Connected"));
+    leave_room->setEnabled(false);
+    show_room->setEnabled(false);
+}
+
+void MultiplayerState::OnAnnounceFailed(const Common::WebResult& result) {
+    announce_multiplayer_session->Stop();
+    QMessageBox::warning(this, tr("Error"),
+                         tr("Failed to announce the room to the public lobby.\nThe room will not "
```
BreadFish64: We get quite a few users who don't know what to do about the "credentials needed" error, I'd hate to make a special case for it, but there should be some way to let them know they need to log in with their Citra account or make a private room

src/citra_qt/multiplayer/validation.h
```diff
@@ -0,0 +1,28 @@
...
+static const QValidator* nickname = new QRegExpValidator(nickname_regex);
+
+/// ipv4 address only
+// TODO remove this when we support hostnames in direct connect
+static const QRegExp ip_regex(
+    "(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|"
+    "2[0-4][0-9]|25[0-5])");
+static const QValidator* ip = new QRegExpValidator(ip_regex);
```
wwylele: Are you sure you are going to:
 - `new` a global object and not `delete` it anywhere (or somewhere I can't see) (I know this is harmless in general, but would produce noise if we run citra in memory leak detection tools)
 - and `new` objects for every translation units, one for each, that include this header file?

src/citra_qt/multiplayer/client_room.cpp
jroweboy: @wwylele if thats the case, then we should make this change for the other places that have the issue.

src/citra_qt/game_list_p.h
```diff
@@ -7,6 +7,7 @@
 #include <atomic>
 #include <map>
 #include <unordered_map>
+#include <QCoreApplication>
```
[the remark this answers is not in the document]
jroweboy: https://github.com/jroweboy/citra/blob/180568e5dd88d0c0f078f34e493564dd630e26e7/src/citra_qt/game_list_p.h#L161 (i missed adding this header in the quick fix)

src/citra_qt/multiplayer/client_room.cpp
wwylele: @jroweboy AFAIK most if not all existing UI in our code base is using `std::unique_ptr`

jroweboy: Nope

* https://github.com/citra-emu/citra/blob/2d372bae1409cd8088c9c6b03fb7334a1e4357f0/src/citra_qt/configuration/configure_debug.cpp#L9
* https://github.com/citra-emu/citra/blob/2d372bae1409cd8088c9c6b03fb7334a1e4357f0/src/citra_qt/configuration/configure_dialog.cpp#L10
* https://github.com/citra-emu/citra/blob/2d372bae1409cd8088c9c6b03fb7334a1e4357f0/src/citra_qt/configuration/configure_general.cpp#L13
* https://github.com/citra-emu/citra/blob/2d372bae1409cd8088c9c6b03fb7334a1e4357f0/src/citra_qt/configuration/configure_graphics.cpp#L11
* https://github.com/citra-emu/citra/blob/2d372bae1409cd8088c9c6b03fb7334a1e4357f0/src/citra_qt/configuration/configure_system.cpp#L28
* https://github.com/citra-emu/citra/blob/2d372bae1409cd8088c9c6b03fb7334a1e4357f0/src/citra_qt/aboutdialog.cpp#L12 (This one deletes it appropriately though)

wwylele: @jroweboy for all examples you gave they are declared as `unique_ptr` in the header file, although constructed by `new`, which is, though not preferable, still fine. However yours are just simply declared as raw pointer.

jroweboy: wasn't aware that you can pass a raw pointer to the constructor of unique ptr like that. (sorry for not checking the header) will fix, thanks.

jroweboy:
Addressed the review comments.

tywald:
One issue I encountered on Appveyor build eac98297.
Trying to expand to view players inside the room shows blank:
image

By opening the Public Room Browser the log says:

jroweboy:
Last call for issue reports. If you find any bugs, report them here soon because I plan to have this merged very very soon.

Issues fixed since the last update:
* Host Room and Direct Connect are working again
* Trying to host a room while in a room only gives one warning pop up (and no longer crashes)
* Player list is working again

tywald:
Tested with Appveyor build: HEAD-c7fee2c

If you filter anything then remove it, it will break the player list.

src/citra_qt/multiplayer/message.cpp
```diff
@@ -0,0 +1,61 @@
...
+    QT_TR_NOOP("Unable to connect to the host. Verify that the connection settings are correct. If "
+               "you still cannot connect, contact the room host and verify that the host is "
+               "properly configured with the external port forwarded."));
+const ConnectionError COULD_NOT_CREATE_ROOM(
+    QT_TR_NOOP("Creating a room failed. Please retry. Restarting Citra might be necessary."));
+const ConnectionError HOST_BANNED(
+    QT_TR_NOOP("The host of the room has banned you. Speak with the host to unban you "
+               "or try a different room."));
+const ConnectionError WRONG_VERSION(
+    QT_TR_NOOP("Version mismatch! Please update to the latest version of citra. If the problem "
```
BreadFish64: capital Citra

PR #3082, opened 2017-11-06
Kernel/Thread: Run the main thread in the CPU specified by the process' exheader.

Subv:
System services usually have Core1 in this field, but citra was running them in Core0 regardless.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3082)


Subv:
@wwylele Yes, but I consider that should be another PR because it's not directly related.

Issue #3054, opened 2017-10-27
Happy Home Designer freezes during opening


Chrisatsinnoh:
I tried sending a support ticket to the Citra Community, but then they tell me to submit it here.
Here's what I said regarding my problem:

When I try to run Happy Home Designer in the latest Citra, it always causes the game to never continue during the opening, as the office chair appears to be empty and clicking on the Manual Option causes the game to freeze. How can I solve this?

System Information

Operating System: Windows 8.1
CPU: ??
GPU: ??
Citra Version (found in title bar): Latest Version
Game: Animal Crossing Happy Home Designer
Screenshot of Issue (include the full Citra window including titlebar):
https://i.imgur.com/vlnnlun.jpg3
Log: ??
<???>
Here's the log file of the problem (The [ 0.838167] Render.OpenGL video_core/renderer_opengl/renderer_opengl.cpp:DebugHandler:472: API PERFORMANCE 131218: Program/shader state performance warning: Fragment Shader is going to be recompiled because the shader key based on GL state mismatches. code seems to be causing the issue, but it's currently unknown. I did not use any 3rd parties for Citra.)
[log.txt](https://github.com/citra-emu/citra/files/1422655/log.txt)

chris062689:
In order for your bug report to be useful, we need the full output of your log, along with the full game name and title ID. Along with CPU and GPU.

Chrisatsinnoh:
Animal Crossing Happy Home Designer with a Title ID of 000400000014F100

My computer's CPU is an Intel Core i7-4800MQ CPU @2.70GHz 2.70 GHz and its GPU is an NVIDIA GeForce, I believe. A GTX 770M with 4095 MB.

The full output of my log in is the text file above.

mailwl:
no matter. it's common bug for Animal Crossing, related with not accurate threading emulation, i think...

Chrisatsinnoh:
Did anyone fix the common Animal Crossing bug, though? As for the emulation used, it was a nightly Citra build.

mailwl:
i don't remember details, but bug is easy fixed by inserting pause in WaitSyncronization SVC, but game will looks like slide show, can't find discussion about it, sorry

Chrisatsinnoh:
then the game is simply incompatible? That's a shame.

acnleditor2:
@mailwl i need the files, i want the fix on my fork, send the files to valentinvanelslandeacnl@gmail.com
#2446

[...]

Issue #2749, opened 2017-06-04
Star Fox - Stuck Initializing Game Data


Schplee:
Nightly - 8f0a23b

Star Fox won't continue without a save file.

![xjvxs0o](https://cloud.githubusercontent.com/assets/24275329/26764961/729118ea-4926-11e7-833c-0e27a4dee94c.png)

[EDIT] Updated Log file to trace log
[log.txt](https://github.com/citra-emu/citra/files/1056122/log.txt)

Subv:
Log?

Schplee:
@Subv Yes, sorry. I forgot to include a log file when I rushed to get this issue posted for the compatibility list.

arkhamsaiyan:
The only time this happens is when you do not have a previous save from your actual system import that and it will go away

eadmaster:
do you have a savegame for the European version?
I've tried loading the USA savegame but it says it is corrupted...

einstein95:
No premade savedata is needed, only 0 byte files in the save folder called:
- menu0.sav
- n640.sav
- rmain00.sav
- rmain01.sav
- rmain02.sav
- rscore00.sav
- rscore01.sav
- rscore02.sav

arkhamsaiyan:
well technically, you need certain files from the main save to get past the initializing... screen then

einstein95:
Technically those files just need to be present, not valid files. The game
will then properly format them.

arkhamsaiyan:
but you get those files from the actual system or do you get them from the game data?

eadmaster:
@einstein95: i've tried what you suggested with the latest nightly build, but it still freeze on the initializing screen...
the save files  for the EUR version should go in `C:\Users\[your-user-name]\AppData\Roaming\Citra\sysdata\00040000\00049100\data\00000001`, right?

einstein95:
@arkhamsaiyan Neither. They can be just 0 byte files. In Windows, a renamed new text document should do the trick (making sure extensions are shown).
@eadmaster No.  `%APPDATA%\Citra\sdmc\Nintendo 3DS\00000000000000000000000000000000\00000000000000000000000000000000\title\00040000\00049100\data\00000001`.

eadmaster:
@einstein95: thank you, it works now!

benderscruffy:
still doesn't work in the latest nightly but works in the canary version

lino11:
[...]

Issue #2118, opened 2016-10-09
Pokemon Super Mystery Dungeon: Softlock when opening Meowth Theater in multiple cases


j-selby:
Commit: cc7f115
Save Game (right at the correct spot): [00174600.zip](https://github.com/citra-emu/citra/files/518130/00174600.zip)

Accessing Meowth's Theater initially writes "Meowth's Theater data" immediately, and dumps you to a menu. Selecting a option from this menu freezes the game with either of the below screenshots depending on what you select. After the first opening of Meowth's Theater you are always thrown to the screenshot as well without seeing the menu. Meowth's Theater data is saved even if you don't choose the main save option from the main menu. During all softlocks, no keys work and graphics is frozen, but audio continues to play. This occurs with both CPU JIT on and off.

Update: Deleting Meowth's Theater data from extdata (00001746) allows us access to the main menu again. It now produces the second screenshot, as attached when attempting to access a dungeon, with another softlock.

_Insert witty Team Rocket joke here._

Log during initial write:

```
[  10.219106] Service.FS <Error> core\hle\service\fs\fs_user.cpp:Service::FS::OpenArchive:435: failed to get a handle for archive archive_id=0x00000006 archive_path=[Binary: 010000004617000000000000]
[  10.220596] Service.FS <Warning> core\hle\service\fs\fs_user.cpp:Service::FS::CreateExtSaveData:628: (STUBBED) savedata_high=00000000 savedata_low=00001746 cmd_buff[3]=00000000 cmd_buff[4]=00030002 cmd_buff[5]=00000064 cmd_buff[6]=00001388 cmd_buff[7]=FFFFFFFF cmd_buff[8]=FFFFFFFF icon_size=000036C0 icon_descriptor=00036C0A icon_buffer=31DF2800
[  10.221065] Common.Filesystem <Warning> common\file_util.cpp:FileUtil::IsDirectory:116: stat failed on C:\Users\James\Downloads\citra-latest-windows-amd64/user/sdmc/Nintendo 3DS/00000000000000000000000000000000/00000000000000000000000000000000/extdata/00000000/00001746/: The system cannot find the file specified.
[...]

Issue #3061, opened 2017-10-28
Pokédex 3D stuck at black screen


superheri:
There is the log with DEBUG filter :
[logPokedex3d.txt](https://github.com/citra-emu/citra/files/1424527/logPokedex3d.txt)

Tested on commit : 26065fd

Critical errors popping in core\hle\svc.cpp

Subv:
The hell? why is this thing trying to open `pxi:dev`?

Is this a game or a homebrew application?

BreadFish64:
@Subv it's a 3D Pokemon viewer application, not Homebrew, not really a game either

superheri:
Corrected by #3091 

There is now the possibility to see something interact with a bit of the menu but there is a bug while writing th the SD card :

[logPokedex3dAfterBoot.txt](https://github.com/citra-emu/citra/files/1503415/logPokedex3dAfterBoot.txt)


![pkdex3d1](https://user-images.githubusercontent.com/30275781/33234943-e885bd3a-d1fc-11e7-98eb-0e32caf5fdb3.PNG)
![pkdex3d2](https://user-images.githubusercontent.com/30275781/33234944-e9e653e2-d1fc-11e7-8107-5cde992b15a3.PNG)

FearlessTobi:
This should be closed because of #3091.

PR #3091, opened 2017-11-08
Kernel/IPC: Add a small delay after each SyncRequest to prevent thread starvation.

### LINKED ISSUE #2794: Moco Moco Friends - softlock on chapter screen
Citra version: Nightly - ecd332b

The game had a solflock at the chapter screen. The music was still playing, but inputs doesn't do anything. 

screen shot 06-20-17 at 01 28 pm

[log.txt](https://github.com/citra-emu/citra/files/1090086/log.txt)

System Information
* Operating System: Windows 7 Ultimate Service Pack 1
* CPU: Intel(R) Core(TM) i7-3610QM CPU @ 2.3GHz
* GPU: Geforce GTX 660M
* Citra Version (found in title bar): citra-windows-msvc-20170619-ecd332b
* Game: Moco Moco Friends (USA)

Subv:
In a real 3DS, threads that call svcSyncRequest are put to sleep until the server responds via svcReplyAndReceive. Our HLE services don't implement this mechanism and are effectively immediate from the 3DS's point of view. This commit makes it so that we at least simulate the IPC delay.

Specific HLE handlers might need to put their callers to sleep for a longer period of time to simulate IO timings. This is their responsibility but doing so is currently not implemented.

See https://gist.github.com/ds84182/4a7690c5376e045cab9129ca4185bbeb for a test that was not passing before this commit.

Thanks @ds84182 for pointing this out!

Closes #2794 #1105 #3054 #2647 #2802 #2608 #2749 #2118 #2526 #2749 

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3091)

### DIFF
--- a/src/core/hle/kernel/server_session.cpp
+++ b/src/core/hle/kernel/server_session.cpp
@@ -67,13 +67,28 @@ ResultCode ServerSession::HandleSyncRequest(SharedPtr<Thread> thread) {
     // If this ServerSession has an associated HLE handler, forward the request to it.
     if (hle_handler != nullptr) {
         hle_handler->HandleSyncRequest(SharedPtr<ServerSession>(this));
-    } else {
+    }
+
+    if (thread->status == THREADSTATUS_RUNNING) {
         // Put the thread to sleep until the server replies, it will be awoken in
-        // svcReplyAndReceive.
+        // svcReplyAndReceive for LLE servers.
         thread->status = THREADSTATUS_WAIT_IPC;
-        // Add the thread to the list of threads that have issued a sync request with this
-        // server.
-        pending_requesting_threads.push_back(std::move(thread));
+
+        if (hle_handler != nullptr) {
+            // For HLE services, we put the request threads to sleep for a short duration to
+            // simulate IPC overhead, but only if the HLE handler didn't put the thread to sleep for
+            // other reasons like an async callback. The IPC overhead is needed to prevent
+            // starvation when a thread only does sync requests to HLE services while a
+            // lower-priority thread is waiting to run.
+
+            // TODO(Subv): Figure out a good value for this.
+            static constexpr u64 IPCDelayNanoseconds = 100;
+            thread->WakeAfterDelay(IPCDelayNanoseconds);
+        } else {
+            // Add the thread to the list of threads that have issued a sync request with this
+            // server.
+            pending_requesting_threads.push_back(std::move(thread));
+        }
     }
 
     // If this ServerSession does not have an HLE implementation, just wake up the threads waiting
--- a/src/core/hle/kernel/server_session.cpp
+++ b/src/core/hle/kernel/server_session.cpp
@@ -81,8 +81,11 @@ ResultCode ServerSession::HandleSyncRequest(SharedPtr<Thread> thread) {
             // starvation when a thread only does sync requests to HLE services while a
             // lower-priority thread is waiting to run.
 
-            // TODO(Subv): Figure out a good value for this.
-            static constexpr u64 IPCDelayNanoseconds = 100;
+            // This delay was approximated in a homebrew application by measuring the average time
+            // it takes for svcSendSyncRequest to return when performing the SetLcdForceBlack IPC
+            // request to the GSP:GPU service in a n3DS with firmware 11.6. The measured values have
+            // a high variance and vary between models.
+            static constexpr u64 IPCDelayNanoseconds = 39000;
             thread->WakeAfterDelay(IPCDelayNanoseconds);
         } else {
             // Add the thread to the list of threads that have issued a sync request with this


mailwl:
looks like fixes Animal Crossing #1105

mailwl:
don't think each SyncRequest needs own delay, because games works well on differ variants of (new/old)3ds/2ds and firmwares

mailwl:
also fixes FIFA (at least 12), but first need we need to signal y2r::completion_event ASYNC

Dragios:
Does it also fix the the missing models on the character / hand in ANCL?

mailwl:
@Dragios, yes. i don't play enough yet, but at start missing chars visible

dampih:
This PR also fixes #3054  .My friend said it also fixes **Dragon Quest XI** stuck after choosing New Game and became playable, but unfortunately it also make **Attack of the Friday Monster! A Tokyo tale** stuck at intro logo. Sorry I don't have those game so I can't test it out.
citra-qt 2017-11-08 16-07-06-14

acnleditor2:
OMG

wwylele:
IMO the default delay time should be the minimal one a `SendSyncRequest` can take. i.e. it is the time taken by a no-op service function. Is this measurable or calculable?

acnleditor2:
this pr causes more problems with mario maker, pigeon dont eats with priority boost hack

wwylele:
@acnleditor2 no other hack talk is allowed here, thanks.

Subv:
@wwylele Agreed. We could write a custom service that has a single function that takes an input ticks value, calculates the difference with the tick value when it is executed, and returns that in the command response. Then we just call that a lot of times and average it out.

mailwl:
just check random softlock game: Moco Moco Friends - softlock on chapter screen #2794
it works.
looks like maaaany issues will be closed

dampih:
Changing initial value `100 `to `90000` give additional bonus fixes. Fixes #2446  #2802  #2745

Subv:
That delay really has to be measured on real hardware, but I believe an IPC overhead of `0.09ms` (`90000ns`) isn't out of the realm of possibility.

Dragios:
Fixes #3061 as well. I have set the delay to 0.1ms.
Pokédex 3D Pro now boots up instantly rather than stuck at black screen.

src/core/hle/kernel/server_session.cpp
```diff
@@ -62,13 +62,28 @@ ResultCode ServerSession::HandleSyncRequest(SharedPtr<Thread> thread) {
...
+        if (hle_handler != nullptr) {
+            // For HLE services, we put the request threads to sleep for a short duration to
+            // simulate IPC overhead, but only if the HLE handler didn't put the thread to sleep for
+            // other reasons like an async callback. The IPC overhead is needed to prevent
+            // starvation when a thread only does sync requests to HLE services while a
+            // lower-priority thread is waiting to run.
+
+            // TODO(Subv): Figure out a good value for this.
+            static constexpr u64 IPCDelayNanoseconds = 100;
+            thread->WakeAfterDelay(IPCDelayNanoseconds);
```
Subv: I just noticed that due to the way WakeAfterDelay is implemented, we can't have delays of less than a microsecond, this is essentially the same as SleepThread(0)

Subv:
I have updated the delay to 54.14 microseconds as measured by @jroweboy. Could you all please re-test everything and report here what was fixed / what was broken by this?

bakugo:
Can confirm that https://github.com/citra-emu/citra/issues/2802 is fixed with this value.

acnleditor2:
Mario Maker: Pigeon stops here. needs a higher value
super mario maker 3ds

acnleditor2:
Pokémon Sun & Moon: freeze after changing the form of Lycanroc in the Pokédex

acnleditor2:
Pokémon Sun & Moon: freeze exiting festival plaza
Pokémon Sun & Moon: freeze after exiting pokedex, entering in festival plaza.

Hexagon12:
Can confirm that #2608 is also fixed.

acnleditor2:
AC:HHD freeze doing a house (except for lottie and goldie)

RavenHome1:
causes pokemon moon to show a black screen instead of the intro

acnleditor2:
achhd freeze on search items by name/color? with some items (only with shader jit enabled)?
achhd freeze on enter in building with shader jit disabled
achhd...................................... is unplayable

Chrisatsinnoh:
Where can I download the fixes, though?

bakugo:
Update on https://github.com/citra-emu/citra/issues/2802: The freeze that always happens when starting the game OR loading a save is indeed fixed with `IPCDelayNanoseconds = 51140`, however it still happens in other situations: if you create a character, let it save, then quit and start again, it will happen after the intro cutscene. Increasing the value further (to `851140`) makes the game not freeze but get stuck on a white screen instead with sound playing.

Seems like this is a pretty complex issue.

einstein95:
This fixes #2749. Star Fox 64 3D now properly initialises save data without needing blank files or premade save files.

j-selby:
Fixes #2118. Woop!

AuraOfTheDawn:
Megaman Legacy collection no longer hangs after the intro. ( #2526 )

SigmaVirus:
I saw a mention of https://github.com/citra-emu/citra/issues/2526, this particular issue has a reference to https://github.com/citra-emu/citra/issues/2238

Might https://github.com/citra-emu/citra/issues/2238 be solved by this PR too by any chance? Any one knows?

acnleditor2:
#3119 breaks ever oasis fix

wwylele:
I am going to temporarily pull this from canary and push #3119 first. We need to make sure it doesn't break anything from master. After everything settle down, we will revisit this fix

einstein95:
@SigmaVirus No, it doesn't

bunnei:
Can this be merged? (I have hesitation since I don't see any more approvals...)

Subv:
This is waiting on someone actually measuring the delay properly

mailwl:
hmm... i suppose New 3ds faster, than 2ds, sorry for trolling

B3n30:
@mailwl nope 2ds is faster.

HW tests with https://gist.github.com/ds84182/ecdbbd25b56a29bd4e5b32a7544b8e92 compared with testing in citra showed that a delay of 39ms will give the same results as the average on new3ds. Those were also values observed on 2ds. Thus a delay of 39ms is a good start. But I also would add a comment on how we obtained that value. Besides changing the delay: LGTM

src/core/hle/kernel/server_session.cpp
tokumeiwokiboushimasu: Is this really 51140 instead of 54140?

PR #3096, opened 2017-11-08
Kernel/Arbiters: When doing ArbitrateAddress(Signal), always pick the highest priority thread, using the first one that was put to sleep if more than one thread with the same highest priority exists.

Subv:
This is consistent with hardware behavior as shown by this test https://gist.github.com/ds84182/40e46129bd38b46a5100f15f96ba5eaf

Many thanks to @ds84182 for writing these tests!

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3096)

### DIFF
--- a/src/core/hle/kernel/address_arbiter.h
+++ b/src/core/hle/kernel/address_arbiter.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <vector>
 #include "common/common_types.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/result.h"
@@ -11,12 +12,14 @@
 // Address arbiters are an underlying kernel synchronization object that can be created/used via
 // supervisor calls (SVCs). They function as sort of a global lock. Typically, games/other CTR
 // applications use them as an underlying mechanism to implement thread-safe barriers, events, and
-// semphores.
+// semaphores.
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Kernel namespace
 
 namespace Kernel {
+
+class Thread;
 
 enum class ArbitrationType : u32 {
     Signal,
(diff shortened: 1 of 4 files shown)


src/core/hle/kernel/address_arbiter.h
```diff
@@ -50,11 +53,25 @@ class AddressArbiter final : public Object {
...
+
+    /// Resume all threads found to be waiting on the address under this address arbiter
+    void ResumeAllThreads(VAddr address);
+
+    /// Resume one thread found to be waiting on the address under this address arbiter and return
+    /// the resumed thread.
+    SharedPtr<Thread> ResumeHighestPriorityThread(VAddr address);
+
+    /// Threads waiting for the address arbiter to be signaled.
+    std::vector<SharedPtr<Thread>> waiting_threads;
```
mtheall: I'd use a std::list since you're using erase(iterator)

Subv: I'm only removing from the end though

mtheall: In all cases it iterates the whole list. It is just a list of pointers though, so swapping/shifting is not a performance concern. I'm good with a vector. I can't imagine there'd be thousands of threads anyway to make a measurable difference.

Signal(num) can probably be better optimized since it currently has O(n²) complexity since you iterate the entire list each time to remove one. But again, I'm not expecting thousands of threads here.

dampih:
This PR make Citra crash when booting **Kirby Battle Royale** reverting this temporary fix it.
32990573-5cfd7016-cd5e-11e7-89c6-0ae07d19fd7a

BreadFish64:
@dampih is that Fraps? Fraps is known to cause issues with Citra, try disabling it

dampih:
@BreadFish64  disable fraps, still happening. Like I said reverting this PR fix the issue 
same

wwylele:
@dampih could you open an issue for this?

Issue #2440, opened 2017-01-15
The Denpa Men 2: Beyond the Waves Crashes Citra on SW Backend


LunaDook:
Tested on 16a3f9e.

To reproduce, simply turn on the software renderer and launch the game.

Log: http://pastebin.com/iyqTrsF3

Stacktrace: http://i.imgur.com/UxodcVO.png

yuriks:
The stack trace isn't very useful unfortunately... Have you built it yourself? It should have debug info then and show function names.

jroweboy:
@Vectrex33 you can download the debug symbols from appveyor as well https://ci.appveyor.com/api/buildjobs/38hpp9sxt44vexdd/artifacts/citra-windows-msvc-20170114-16a3f9e-debugsymbols.zip

simply unzip the file, and put the pdb files next to the citra exe and then the stack trace will actually show useful information

Vectrex33:
Sorry, here's a proper stack trace: http://i.imgur.com/IZIUjXN.png

Still from the same commit, I'm going to retest this on latest master and see if anything has changed.

Vectrex33:
Just built ebe01506, log shows more information.

Log: http://pastebin.com/2wYrW056

Also here's another stack trace: http://i.imgur.com/2hP7GQS.png

Vectrex33:
This still happens on a4f88c7.

benderscruffy:
nightly HEAD-a66e458
still happens

benderscruffy:
still happens in the latest nightly
[citra_log.zip](https://github.com/citra-emu/citra/files/2729036/citra_log.zip)

benderscruffy:
Has this been fixed ??
Please update

PR #3097, opened 2017-11-09
Round primary color in swrast

ds84182:
Also includes a fix for logic ops not being enabled in OpenGL (depending on circumstances).

Also ports the hack from the hardware renderer to software rasterizer where null texture addresses are ignored. (Otherwise the software rasterizer will crash doing a null deref).

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3097)

### DIFF
--- a/src/video_core/swrasterizer/rasterizer.cpp
+++ b/src/video_core/swrasterizer/rasterizer.cpp
@@ -293,18 +293,18 @@ static void ProcessTriangleInternal(const Vertex& v0, const Vertex& v1, const Ve
             };
 
             Math::Vec4<u8> primary_color{
-                (u8)(
+                static_cast<u8>(round(
                     GetInterpolatedAttribute(v0.color.r(), v1.color.r(), v2.color.r()).ToFloat32() *
-                    255),
-                (u8)(
+                    255)),
+                static_cast<u8>(round(
                     GetInterpolatedAttribute(v0.color.g(), v1.color.g(), v2.color.g()).ToFloat32() *
-                    255),
-                (u8)(
+                    255)),
+                static_cast<u8>(round(
                     GetInterpolatedAttribute(v0.color.b(), v1.color.b(), v2.color.b()).ToFloat32() *
-                    255),
-                (u8)(
+                    255)),
+                static_cast<u8>(round(
                     GetInterpolatedAttribute(v0.color.a(), v1.color.a(), v2.color.a()).ToFloat32() *
-                    255),
+                    255)),
             };
 
             Math::Vec2<float24> uv[3];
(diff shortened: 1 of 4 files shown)


Subv:
Could you please include a hwtest for this?
Also, in the future, please don't include unrelated changes in the same PR, they make it much harder to review.

fincs:
The test case from last night which resulted in this PR: https://feos.mtheall.com/citra-logicopbug.zip (originally intended just for the logic op bug, ended up discovering the rounding bug too)

FearlessTobi:
https://github.com/citra-emu/citra/issues/2440 This should likely be fixed by skipping the null textures. I will test it later.

src/video_core/swrasterizer/rasterizer.cpp
```diff
@@ -293,16 +293,16 @@ static void ProcessTriangleInternal(const Vertex& v0, const Vertex& v1, const Ve
             };
 
             Math::Vec4<u8> primary_color{
-                (u8)(
+                (u8)round(
```
B3n30: is there a reason static_cast<u8>() isn't used?

src/video_core/swrasterizer/rasterizer.cpp
```diff
@@ -331,6 +331,7 @@ static void ProcessTriangleInternal(const Vertex& v0, const Vertex& v1, const Ve
                 // Only unit 0 respects the texturing type (according to 3DBrew)
                 // TODO: Refactor so cubemaps and shadowmaps can be handled
                 PAddr texture_address = texture.config.GetPhysicalAddress();
+                if (texture_address == 0) continue;
```
B3n30: Should there be a log if that case happens? Maybe debug

B3n30:
What's the status of that PR? It needs clang-format fixing. Also my comments weren't addressed yet.

ds84182:
I'll try to get back to this today but I discovered that this fixes Pokemon X&Y because textures with a null address get skipped with one of my changes. I did some GPU research and I concluded that the color output from a null address texture depends on the previous texture used and it varies if the types of the texture are different. The color is deterministic since it depends on what was drawn before it, but I'd conclude that **this is a bug in Pokemon X & Y**. The only reason why it wasn't fixed was because the developers _could not_ see what was happening with the null texture, since they always draw the male character's body before the two cheeks. In my tests I get slightly non-deterministic results when opening the home menu while using the null address. If a game depends on this weird hardware behavior for correct rendering we'd be in trouble though.

Anyways the correct thing to do is to warn in the log when a null texture is used and emit a transparent black. I've already argued that using the null texture is a bug in the game and is undefined (could break if Nintendo ever decides to do a new GPU revision, although I don't think they will), and we should provide a sensible default.

I'll remove the null address check out of software rasterizer for now, and I'll address the comments.

MakerMayker:
I hope this works.

PR #3077, opened 2017-11-05
Kernel/IPC: Implement StaticBuffer translation for HLE services that use the HLERequestContext architecture.

Subv:
The real kernel requires services to set up their static buffer targets ahead of time. This implementation does not require that and will simply create the storage for the buffers as they are processed in the incoming IPC request.

Static buffers are kept in an unordered_map keyed by their buffer id, and are written into the already-setup area of the request thread when responding an IPC request.

This fixes a regression (crash) introduced in #2992.

This PR introduces more warnings due to the [[deprecated]] attribute being added to void PushStaticBuffer(VAddr buffer_vaddr, size_t size, u8 buffer_id); and VAddr PopStaticBuffer(size_t* data_size);

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3077)

PR #3101, opened 2017-11-09
Kernel/Threads: Add a new thread status that will allow using a Kernel::Event to put a guest thread to sleep inside an HLE handler until said event is signaled

Subv:
This is an improved take on #2968

Some service functions should block the caller guest thread on SendSyncRequest until an async response can be calculated and returned (Most soc:U functions when dealing with blocking sockets, some UDS functions like ConnectToNetwork, most http:C functions, etc).

This change lets us put the threads to wait in the HLE handler and wake them up when the response comes back from another host thread, without having to neither spinlock (and freezing the entirety of the emulator thread) nor ignore the blocking nature of the function.

It is used like so:
```cpp
// In service function
auto promise = PerformAsyncOperation();
auto event = ctx.SleepClientThread(
        Kernel::GetCurrentThread(), "soc::sendto", SOCSendToTimeout,
        [&promise](Kernel::SharedPtr<Kernel::Thread> thread, Kernel::HLERequestContext& ctx,
           ThreadWakeupReason reason) {
        if (reason == ThreadWakeupReason::Timeout) {
            // Construct RequestBuilder and write Timeout response.
        } else {
            // Construct RequestBuilder and write Success response.
        }
    });
...
// Somewhere else, when the async operation has finished
promise.set_value(ResultOfOperation);
std::lock_guard<std::mutex> lock(HLE::g_hle_lock);
event->Signal();
// Now the thread is awake and back into the scheduler queue.
```

# Concerns
The timeout value is specified in nanoseconds elapsed in the guest CPU, but if no threads are eligible to run then CoreTiming might decide to advance the time to the next queued event. This means we have to carefully choose the timeout values when implementing such logic.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3101)

### DIFF
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -39,7 +39,7 @@ SharedPtr<Event> HLERequestContext::SleepClientThread(SharedPtr<Thread> thread,
         // We must copy the entire command buffer *plus* the entire static buffers area, since
         // the translation might need to read from it in order to retrieve the StaticBuffer
         // target addresses.
-        std::array<u32, IPC::COMMAND_BUFFER_LENGTH + 2 * IPC::MAX_STATIC_BUFFERS> cmd_buff;
+        std::array<u32_le, IPC::COMMAND_BUFFER_LENGTH + 2 * IPC::MAX_STATIC_BUFFERS> cmd_buff;
         Memory::ReadBlock(*process, thread->GetCommandBufferAddress(), cmd_buff.data(),
                           cmd_buff.size() * sizeof(u32));
         context.WriteToOutgoingCommandBuffer(cmd_buff.data(), *process, Kernel::g_handle_table);
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -29,9 +29,8 @@ SharedPtr<Event> HLERequestContext::SleepClientThread(SharedPtr<Thread> thread,
                                                       const std::string& reason, u64 timeout,
                                                       WakeupCallback&& callback) {
     // Put the client thread to sleep until the wait event is signaled or the timeout expires.
-    thread->wakeup_callback = [context = *this, callback](ThreadWakeupReason reason,
-                                                          SharedPtr<Thread> thread,
-                                                          SharedPtr<WaitObject> object) mutable {
+    thread->wakeup_callback = [ context = *this, callback ](
+        ThreadWakeupReason reason, SharedPtr<Thread> thread, SharedPtr<WaitObject> object) mutable {
         ASSERT(thread->status == THREADSTATUS_WAIT_HLE_EVENT);
         callback(thread, context, reason);
 
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -26,7 +26,8 @@ void SessionRequestHandler::ClientDisconnected(SharedPtr<ServerSession> server_s
 }
 
 SharedPtr<Event> HLERequestContext::SleepClientThread(SharedPtr<Thread> thread,
-                                                      const std::string& reason, u64 timeout,
+                                                      const std::string& reason,
+                                                      std::chrono::nanoseconds timeout,
                                                       WakeupCallback&& callback) {
     // Put the client thread to sleep until the wait event is signaled or the timeout expires.
     thread->wakeup_callback = [ context = *this, callback ](
@@ -52,8 +53,8 @@ SharedPtr<Event> HLERequestContext::SleepClientThread(SharedPtr<Thread> thread,
     thread->wait_objects = {event};
     event->AddWaitingThread(thread);
 
-    if (timeout > 0)
-        thread->WakeAfterDelay(timeout);
+    if (timeout.count() > 0)
+        thread->WakeAfterDelay(timeout.count());
 
     return event;
 }
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -6,6 +6,7 @@
 #include <boost/range/algorithm_ext/erase.hpp>
 #include "common/assert.h"
 #include "common/common_types.h"
+#include "core/hle/kernel/event.h"
 #include "core/hle/kernel/handle_table.h"
 #include "core/hle/kernel/hle_ipc.h"
 #include "core/hle/kernel/kernel.h"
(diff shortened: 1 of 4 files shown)


src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -23,6 +24,40 @@ void SessionRequestHandler::ClientDisconnected(SharedPtr<ServerSession> server_s
...
+                                                          SharedPtr<Thread> thread,
+                                                          SharedPtr<WaitObject> object) mutable {
+        ASSERT(thread->status == THREADSTATUS_WAIT_HLE_EVENT);
+        callback(thread, context, reason);
+
+        auto& process = thread->owner_process;
+        // We must copy the entire command buffer *plus* the entire static buffers area, since
+        // the translation might need to read from it in order to retrieve the StaticBuffer
+        // target addresses.
+        std::array<u32, IPC::COMMAND_BUFFER_LENGTH + 2 * 16> cmd_buff;
```
Subv: change this to MAX_STATIC_BUFFERS when #3077 is merged

src/core/hle/kernel/hle_ipc.h
```diff
@@ -100,6 +103,22 @@ class HLERequestContext {
...
+     * specified timeout expires.
+     * @param thread Thread to be put to sleep.
+     * @param reason Reason for pausing the thread, to be used for debugging purposes.
+     * @param callback Callback to be invoked when the thread is resumed. This callback must write
+     * the entire command response once again, regardless of the state of it before this function
+     * was called.
+     * @returns Event that when signaled will resume the thread and call the callback function.
+     */
+    SharedPtr<Event> SleepClientThread(SharedPtr<Thread> thread, const std::string& reason,
+                                       u64 timeout, WakeupCallback callback);
```
danzel: The units of timeout are not specified

src/core/hle/kernel/hle_ipc.cpp
MerryMage: plsfix

B3n30:
don't merge, since it will break canary. We first have to rewrite nwm:uds to a service framework or remove multiplayer from canary

Subv:
#2968 should be reverted in favor of this PR once it's reviewed.

src/core/hle/kernel/hle_ipc.cpp
CDAGaming: Clang Format Fixes Needed...

CDAGaming:
Cherry Pick [this commit](https://github.com/Subv/citra/pull/8/commits/2e8cb54a4bc95a780c825e2c028aa27d9abfe17a) for Clang Format Fixes, @Subv .

src/core/hle/kernel/hle_ipc.h
```diff
@@ -140,6 +143,24 @@ class HLERequestContext {
...
+     * @param reason Reason for pausing the thread, to be used for debugging purposes.
+     * @param timeout Timeout in nanoseconds after which the thread will be awoken and the callback
+     * invoked with a Timeout reason.
+     * @param callback Callback to be invoked when the thread is resumed. This callback must write
+     * the entire command response once again, regardless of the state of it before this function
+     * was called.
+     * @returns Event that when signaled will resume the thread and call the callback function.
+     */
+    SharedPtr<Event> SleepClientThread(SharedPtr<Thread> thread, const std::string& reason,
+                                       u64 timeout, WakeupCallback&& callback);
```
lioncash: Would it not be better to use `std::chrono::nanoseconds` here, to enforce the time unit?

lioncash:
The clang format stuff could also just be amended into the previous commit, since it's fairly trivial stuff.

src/core/hle/kernel/hle_ipc.h
B3n30: I think it depends on the use case. If the AsyncOperation is 100% inside the emulation process (no call to a server, etc) citras core_timing should be used. But in other cases  a timeout outside of core_timing should be used, to assure the timeout is really what you want. In the second case you probably have to write your own timeout function.

lioncash: This would still work with coretiming. `std::chrono::nanoseconds` just makes the precondition (the value represents a nanosecond value, as stated in the doc comment) part of the enforced API. If the raw value is needed, then it's as simple as calling `.count()` on the object internally. Whether or not coretiming is used under the hood doesn't matter.

B3n30: Sorry. I got your first statement wrong. I thought you were suggesting to use `std:chrono` to determine the timeout

lioncash: Yeah, to be clearer I'm only suggesting:

```cpp
SharedPtr<Event> SleepClientThread(SharedPtr<Thread> thread, const std::string& reason,
                                   std::chrono::nanoseconds timeout, WakeupCallback&& callback);
```

which would make the time unit more evident at call sites, as opposed to only having a doc comment about it

CDAGaming:
True, why not just Rebase it back into just the one commit?

src/core/hle/kernel/hle_ipc.h
```diff
@@ -140,6 +143,24 @@ class HLERequestContext {
         return session;
     }
 
+    using WakeupCallback = std::function<void(SharedPtr<Thread> thread, HLERequestContext& context,
+                                              ThreadWakeupReason reason)>;
+
+    /*
```
B3n30: /**

src/core/hle/kernel/hle_ipc.h
B3n30: Why not just std::chrono::duration? Besides that nit: LGTM

wwylele: @B3n30 because you want to specify the unit here

src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -24,6 +25,39 @@ void SessionRequestHandler::ClientDisconnected(SharedPtr<ServerSession> server_s
...
+    thread->wakeup_callback = [ context = *this, callback ](
+        ThreadWakeupReason reason, SharedPtr<Thread> thread, SharedPtr<WaitObject> object) mutable {
+        ASSERT(thread->status == THREADSTATUS_WAIT_HLE_EVENT);
+        callback(thread, context, reason);
+
+        auto& process = thread->owner_process;
+        // We must copy the entire command buffer *plus* the entire static buffers area, since
+        // the translation might need to read from it in order to retrieve the StaticBuffer
+        // target addresses.
+        std::array<u32, IPC::COMMAND_BUFFER_LENGTH + 2 * IPC::MAX_STATIC_BUFFERS> cmd_buff;
```
wwylele: ` std::array<u32_le, ...>`

PR #3113, opened 2017-11-14
SDL CIA Installation

shinyquagsire23:
This PR can be reviewed commit-by-commit.

This PR adds command-line CIA installs with the `-i` or `--install=` argument. Installation is done using the `CIAContainer` and `CIAFile` classes, `CIAContainer` to initially check if a CIA is valid, and `CIAFile` to handle the actual installation.

Example CIA installation command:
`citra -i scummvm.cia`

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3113)

### DIFF
--- a/src/citra/citra.cpp
+++ b/src/citra/citra.cpp
@@ -26,6 +26,7 @@
 
 #include "citra/config.h"
 #include "citra/emu_window/emu_window_sdl2.h"
+#include "common/file_util.h"
 #include "common/logging/backend.h"
 #include "common/logging/filter.h"
 #include "common/logging/log.h"
@@ -33,7 +34,9 @@
 #include "common/scope_exit.h"
 #include "common/string_util.h"
 #include "core/core.h"
+#include "core/file_sys/cia_container.h"
 #include "core/gdbstub/gdbstub.h"
+#include "core/hle/service/am/am.h"
 #include "core/loader/loader.h"
 #include "core/settings.h"
 
@@ -41,6 +44,7 @@ static void PrintHelp(const char* argv0) {
     std::cout << "Usage: " << argv0
               << " [options] <filename>\n"
                  "-g, --gdbport=NUMBER  Enable gdb stub on port NUMBER\n"
+                 "-i, --install=FILE    Installs a specified CIA file\n"
                  "-h, --help            Display this help and exit\n"
                  "-v, --version         Output version information and exit\n";
 }
@@ -69,13 +73,14 @@ int main(int argc, char** argv) {
 
     static struct option long_options[] = {
         {"gdbport", required_argument, 0, 'g'},
+        {"install", required_argument, 0, 'i'},
         {"help", no_argument, 0, 'h'},
         {"version", no_argument, 0, 'v'},
         {0, 0, 0, 0},
     };
 
     while (optind < argc) {
-        char arg = getopt_long(argc, argv, "g:hv", long_options, &option_index);
+        char arg = getopt_long(argc, argv, "g:i:hv", long_options, &option_index);
         if (arg != -1) {
             switch (arg) {
             case 'g':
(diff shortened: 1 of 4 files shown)


jroweboy:
Just skimmed it, awesome! Wanted to bring up that a lot of that logic for CIA install in the SDL frontend looks like it should be something thats in core *somewhere* since its basically going to be duplicated for every frontend that wants to install a CIA file. Does anyone have any thoughts on this?

Subv:
Agreed, the Install function itself should be in core

src/citra/citra.cpp
```diff
@@ -26,21 +26,25 @@
...
 #include "core/gdbstub/gdbstub.h"
+#include "core/hle/service/am/am.h"
 #include "core/loader/loader.h"
 #include "core/settings.h"
 
 static void PrintHelp(const char* argv0) {
     std::cout << "Usage: " << argv0
               << " [options] <filename>\n"
                  "-g, --gdbport=NUMBER  Enable gdb stub on port NUMBER\n"
+                 "-i, --install=FILE    Installs a specified CIA file\n"
```
mailwl: there is no "=" it TS...

BreadFish64:
can confirm it works

src/citra/citra.cpp
shinyquagsire23: It seems to accept either `--install scummvm.cia` or `--install=scummvm.cia`, I think `--something=somethingelse` is pretty standard (was just going off --gdbport though, to be honest).

src/core/hle/service/am/am.h
```diff
@@ -38,6 +42,53 @@ enum class CIAInstallState : u32 {
...
+    u64 written = 0;
+    FileSys::CIAContainer container;
+    std::vector<u8> data;
+    std::vector<u64> content_written;
+    Service::FS::MediaType media_type;
+};
+
+/**
+ * Installs a CIA file from a specified file path.
+ * @param path file path of the CIA file to install
```
Subv: Missing parameter docs for the callback

src/core/hle/service/am/am.h
```diff
@@ -38,6 +42,53 @@ enum class CIAInstallState : u32 {
...
+ * @returns bool whether the install was successful
+ */
+bool InstallCIA(const std::string& path,
+                const std::function<void(size_t, size_t)>&& update_callback = {});
```
Subv: remove the const. Use nullptr instead of {}

Alias void(size_t, size_t) to something more meaningful like `using ProgressCallback = void(size_t, size_t);`

src/core/hle/service/am/am.cpp
```diff
@@ -78,190 +77,268 @@ struct TicketInfo {
...
+    return written;
+}
+
+bool CIAFile::SetSize(u64 size) const {
+    return false;
+}
+
+bool CIAFile::Close() const {
+    bool complete = true;
+    for (int i = 0; i < container.GetTitleMetadata().GetContentCount(); i++) {
```
lioncash: This is a signed/unsigned mismatch, `GetContentCount()` returns a `size_t`

src/citra/citra.cpp
```diff
@@ -89,6 +98,12 @@ int main(int argc, char** argv) {
                     exit(1);
                 }
                 break;
+            case 'i':
+                if (!Service::AM::InstallCIA(std::string(optarg), CIAProgress))
```
MerryMage: Personally I prefer:

            case 'i': {
                const auto cia_progress = [](size_t written, size_t total) {
                    LOG_INFO(Frontend, "%02zu%%", (written * 100 / total));
                };
                if (!Service::AM::InstallCIA(std::string(optarg), cia_progress))
                    errno = EINVAL;
                if (errno != 0)
                    exit(1);
                break;
            }

to limit the scope of CIAProgress.

shinyquagsire23: Yeah that's way cleaner, thanks.

src/core/hle/service/am/am.cpp
```diff
@@ -78,190 +77,268 @@ struct TicketInfo {
...
+
+        FileUtil::Delete(old_tmd_path);
     }
+    return true;
+}
+
+void CIAFile::Flush() const {}
 
-    bool SetSize(u64 size) const override {
+bool InstallCIA(const std::string& path, std::function<ProgressCallback>&& update_callback) {
```
BreadFish64: Would it be possible to return a more detailed result than true/false so that GUI frontends can give the user a more specific error?

shinyquagsire23: Done

PR #3116, opened 2017-11-15
Allow input configuration with SDL joysticks

muemart:
This PR adds the ability to configure controls using SDL gamepads (or joysticks, I guess). I created some scaffolding so that the settings dialog can ask for inputs from input backends. Basically, there is now an abstract DevicePoller class (if anyone has a better name, let me know) which can be implemented by different backends to return inputs.
There are a few ugly corners, which might or might not be a problem:
- The poller for SDL devices drops all SDL events it doesn't need. The Qt frontend doesn't use them, so it should be fine, but there might be other problems I'm not aware of.
- Configuring the analog sticks might be a bit confusing, because now you can configure them with 5 buttons, or an actual analog stick. Might need a better GUI for this.
- The buttons still have placeholder labels for SDL mappings, but that's easy to do once everything else is done.

There are two new buttons under the input tab in the configuration window to set analog stick controls (labeled "Set Analog Stick"):

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3116)

### DIFF
--- a/src/citra_qt/configuration/configure_input.cpp
+++ b/src/citra_qt/configuration/configure_input.cpp
@@ -71,9 +71,10 @@ ConfigureInput::ConfigureInput(QWidget* parent)
     for (int button_id = 0; button_id < Settings::NativeButton::NumButtons; button_id++) {
         if (button_map[button_id])
             connect(button_map[button_id], &QPushButton::released, [=]() {
-                handleClick(button_map[button_id],
-                            [=](const Common::ParamPackage& params) { buttons_param[button_id] = params; },
-                            InputCommon::Polling::DeviceType::Button);
+                handleClick(
+                    button_map[button_id],
+                    [=](const Common::ParamPackage& params) { buttons_param[button_id] = params; },
+                    InputCommon::Polling::DeviceType::Button);
             });
     }
 
@@ -95,9 +96,10 @@ ConfigureInput::ConfigureInput(QWidget* parent)
             QMessageBox::information(
                 this, "Information",
                 "After pressing OK, first move your joystick horizontally, and then vertically.");
-            handleClick(analog_map_stick[analog_id],
-                        [=](const Common::ParamPackage& params) { analogs_param[analog_id] = params; },
-                        InputCommon::Polling::DeviceType::Analog);
+            handleClick(
+                analog_map_stick[analog_id],
+                [=](const Common::ParamPackage& params) { analogs_param[analog_id] = params; },
+                InputCommon::Polling::DeviceType::Analog);
         });
     }
 
--- a/src/citra_qt/configuration/configure_input.cpp
+++ b/src/citra_qt/configuration/configure_input.cpp
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <memory>
 #include <utility>
+#include <QMessageBox>
 #include <QTimer>
 #include "citra_qt/configuration/config.h"
 #include "citra_qt/configuration/configure_input.h"
@@ -43,7 +44,8 @@ static void SetAnalogButton(const Common::ParamPackage& input_param, Common::Par
 
 ConfigureInput::ConfigureInput(QWidget* parent)
     : QWidget(parent), ui(std::make_unique<Ui::ConfigureInput>()),
-      timer(std::make_unique<QTimer>()) {
+      timeout_timer(std::make_unique<QTimer>()), poll_timer(std::make_unique<QTimer>()),
+      want_keyboard_keys(false) {
 
     ui->setupUi(this);
     setFocusPolicy(Qt::ClickFocus);
@@ -91,22 +93,29 @@ ConfigureInput::ConfigureInput(QWidget* parent)
             }
         }
         connect(analog_map_stick[analog_id], &QPushButton::released, [=]() {
+            QMessageBox::information(
+                this, "Information",
+                "After pressing OK, first move your joystick horizontally, and then vertically.");
             handleClick(analog_map_stick[analog_id],
-                        [](Common::ParamPackage params) {
-
-                        },
+                        [=](Common::ParamPackage params) { analogs_param[analog_id] = params; },
                         InputCommon::Polling::DeviceType::Analog);
         });
     }
 
     connect(ui->buttonRestoreDefaults, &QPushButton::released, [this]() { restoreDefaults(); });
 
-    timer->setSingleShot(true);
-    connect(timer.get(), &QTimer::timeout, [this]() {
-        releaseKeyboard();
-        releaseMouse();
-        input_setter = boost::none;
-        updateButtonLabels();
+    timeout_timer->setSingleShot(true);
+    connect(timeout_timer.get(), &QTimer::timeout, [this]() { setPollingResult({}, true); });
+
+    connect(poll_timer.get(), &QTimer::timeout, [this]() {
+        Common::ParamPackage params;
+        for (auto& poller : device_pollers) {
+            params = poller->GetNextInput();
+            if (params.Has("engine")) {
+                setPollingResult(params, false);
+                return;
+            }
+        }
     });
 
     this->loadConfiguration();
@@ -155,7 +164,9 @@ void ConfigureInput::updateButtonLabels() {
     QString non_keyboard(tr("[non-keyboard]"));
 
     auto KeyToText = [&non_keyboard](const Common::ParamPackage& param) {
-        if (param.Get("engine", "") != "keyboard") {
+        if (!param.Has("engine")) {
+            return QString("[not set]");
+        } else if (param.Get("engine", "") != "keyboard") {
             return non_keyboard;
         } else {
             return getKeyName(param.Get("code", 0));
@@ -180,6 +191,7 @@ void ConfigureInput::updateButtonLabels() {
                     analog_map_buttons[analog_id][sub_button_id]->setText(KeyToText(param));
             }
         }
+        analog_map_stick[analog_id]->setText("Set Analog Stick");
     }
 }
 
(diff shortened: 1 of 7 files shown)


jroweboy:
Thanks for the contribution; its something we definitely would like to see it get merged! I'll be doing a full review later, but in the meantime, I'm going to test it locally, and if its all fine, I'll add it to canary. 

Also, I see that you added some new buttons https://github.com/citra-emu/citra/pull/3116/files#diff-282077f5f81f30cefcf9b34f15400054R292 Please include a screenshot for any UI changes that you make so people don't have to build just to give UI feedback. (Bonus points if you have before and after, and include a screenshot for each of the themes)

src/input_common/sdl/sdl.cpp
```diff
@@ -7,6 +7,7 @@
 #include <string>
 #include <tuple>
 #include <unordered_map>
+#include <utility>
```
jroweboy: Need to also include vector (which is keeping it from being in canary for now)

muemart:
Added picture and hopefully made CI happy.

BreadFish64:
My tests with emulated dInput and xInput controllers went swimmingly, however the physical XBone controller was trigger-happy about mapping everything to an analog stick, suggesting there needs to be a bigger dead-zone.

src/input_common/sdl/sdl.cpp
```diff
@@ -247,5 +247,158 @@ void Shutdown() {
...
+
+    ~SDLPoller() = default;
+
+    void Start() override {
+        // SDL joysticks must be opened, otherwise they don't generate events
+        int num_joysticks = SDL_NumJoysticks();
+        for (int i = 0; i < num_joysticks; i++) {
+            joysticks_opened.emplace_back(GetJoystick(i));
+        }
+        // Empty event queue to get rid of old events. citra-qt doesn't use the queue
```
wwylele: Our SDL frontent do use event queue, though it doesn't use the poller. Better to be more specific that the Qt frontend doesn't use the queue, so that in the future if anyone want to use the poller in SDL frontend, they will know this and deal with it

src/input_common/sdl/sdl.cpp
```diff
@@ -247,5 +247,158 @@ void Shutdown() {
...
+            params.Set("axis_y", analog_axes.second);
+            analog_axes = std::make_pair(-1, -1);
+            analog_axes_joystick = -1;
+            return params;
+        }
+        return params;
+    }
+
+private:
+    std::pair<int, int> analog_axes;
```
wwylele: IMO just using plain `int analog_axis_x, analog_axis_y;` is cleaner and more readable.

muemart:
I added a deadzone of 50% now, that should be plenty to ignore unintentional movement.

src/input_common/sdl/sdl.h
```diff
@@ -4,7 +4,13 @@
 
 #pragma once
 
+#include <memory>
+#include <vector>
+#include "common/param_package.h"
 #include "core/frontend/input.h"
+#include "input_common/main.h"
```
lioncash: I'm fairly sure this can be replaced with forward declarations

src/input_common/sdl/sdl.h
```diff
@@ -4,7 +4,13 @@
 
 #pragma once
 
+#include <memory>
+#include <vector>
+#include "common/param_package.h"
```
lioncash: This can be replaced with a forward declaration

src/input_common/main.h
```diff
@@ -5,6 +5,8 @@
 #pragma once
 
 #include <string>
+#include <vector>
+#include "common/param_package.h"
```
lioncash: This can be replaced with a forward declaration.

src/input_common/main.h
```diff
@@ -31,4 +33,30 @@ std::string GenerateKeyboardParam(int key_code);
...
+    /**
+     * Every call to this function returns the next input recorded since calling Start
+     * @return A ParamPackage of the recorded input, which can be used to create an InputDevice.
+     *         If there has been no input, the package is empty
+     */
+    virtual Common::ParamPackage GetNextInput() = 0;
+};
+
+// Get all DevicePoller from all backends for a specific device type
+std::vector<std::unique_ptr<DevicePoller>> getPollers(DeviceType type);
```
lioncash: This should be cased as `GetPollers`

src/input_common/sdl/sdl.h
```diff
@@ -15,5 +21,15 @@ void Init();
 /// Unresisters SDL device factories and shut them down.
 void Shutdown();
 
+/// Creates a ParamPackage from an SDL_Event that can directly be used to create a ButtonDevice
+Common::ParamPackage SDLEventToButtonParamPackage(const SDL_Event& event);
+
+namespace Polling {
+
+/// Get all DevicePoller that use the SDL backend for a specific device type
+std::vector<std::unique_ptr<InputCommon::Polling::DevicePoller>> getPollers(
```
lioncash: This should be cased as `GetPollers`

src/citra_qt/configuration/configure_input.cpp
```diff
@@ -31,23 +31,20 @@ static QString getKeyName(int key_code) {
...
     }
-    analog_param.Set(button_name, InputCommon::GenerateKeyboardParam(key));
+    analog_param.Set(button_name, input_param.Serialize());
 }
 
 ConfigureInput::ConfigureInput(QWidget* parent)
     : QWidget(parent), ui(std::make_unique<Ui::ConfigureInput>()),
-      timer(std::make_unique<QTimer>()) {
+      timeout_timer(std::make_unique<QTimer>()), poll_timer(std::make_unique<QTimer>()),
+      want_keyboard_keys(false) {
```
lioncash: `want_keyboard_keys` can be initialized directly in the class definition.

e.g.

```cpp
class Thing {
...
private:
    bool want_keyboard_keys = false;
};
```

src/input_common/sdl/sdl.cpp
```diff
@@ -247,5 +247,164 @@ void Shutdown() {
...
+                return SDLEventToButtonParamPackage(event);
+            }
+        }
+        return {};
+    }
+};
+
+class SDLAnalogPoller final : public SDLPoller {
+public:
+    SDLAnalogPoller() : analog_xaxis(-1), analog_yaxis(-1), analog_axes_joystick(-1) {}
```
lioncash: These can all be initialized directly at their definition site:

```cpp
class SDLAnalogPoller final : public SDLPoller {
...
private:
    int analog_xaxis = -1;
    int analog_yaxis = -1;
    SDL_JoystickID analog_axes_joystick = -1;
};

Subv:
I'll be testing this locally either tonight or tomorrow morning (GMT-5) with a few of my controllers

Subv:
Sorry for the delay, I didn't review the code but performed a few tests, here's my thoughts on the matter:

* Xbox One Controller wasn't detected when I plugged it in while Citra was already started. This only happened with the normal buttons, the Analog bindings did detect the controller even if it was plugged in after the fact. After binding an analog stick I was then able to bind the normal buttons.
* Binding a gamepad button turns the label into "[non-keyboard]". I think this can be improved, at the very least it should include the gamepad id that the key was bound to. As a first implementation we could just show numbers like in those cheap controllers that have keys labelled 1, 2, 3, etc.

I consider the first point to be very important for this PR. The second observation can be left to a separate PR.

muemart:
Thanks for testing. I didn't do anything with the labels because that's sort of independent of mapping inputs, and as you said this can be done in a separate PR which I'll probably do after this.
I couldn't reproduce your issue, though I only have a Logitech XInput controller, not an Xbox One controller. I did notice however that it takes a few seconds after plugging in until the controller works (I guess this depends on SDL/the OS?). The SDL joysticks get opened every time you press a button, so plugging in a controller while it's waiting for input won't work either, and you have to press the button again.

Subv:
I tested again and the problem is indeed fixed, thanks!

nicoboss:
I’ve tested it using my N64 controller. Unfortunately, it always uses the wrong joystick number and so when configuring joystick 0 it configures it as joystick:1 in the config file but Citra expects it to be joystick:0 and when configuring joystick 1 it put joystick:0 in the config file but Citra expect to be joystick:1.
Configuring joystick 0:
button_a="engine:sdl,joystick:1,button:1" but Citra expect it to be button_a="engine:sdl,joystick:0,button:1"
For connecting real N64 controllers to my PC I’m using the N64 Dual Controller Gamepad to USB Adapter which works perfectly fine.

Furthermore, there seems to be a problem with the analog stick. The game always thinks that it has only moved a little when pushing it to the maximum and so at example in Ocarina of Time link is unable to walk in normal speed.

I’ve tested it using the latest version from your repository and in the Canary build but had the same problems on booth of them.

muemart:
I just fixed a bug with the joystick numbers, but I'm not sure it's the problem you had. What do you mean by "Citra expects it to be joystick:0"? Do the buttons not work after mapping? How do you know what Citra expects?
And for the analog stick, make sure you use the "Set Analog Stick" button and don't map up/down/left/right separately.

nicoboss:
Thank you very much the joystick numbers now working fine!

For the analog stick the "Set Analog Stick" button is not configure the sensitivity properly (It's far to insensitive so that you can't roll in Ocarina of Time and walking is way too slow and it's the same for all other games I have). To configure I just pressed the “Set Analog Stick” button and then moved the joystick horizontally to the right until it reached the maximum, let it go back to the center and then moved it vertically to the bottom and let it go back to the center and it gets accepted. Configure each direction separately would work but it's not how it should be done.
It saves the following into the configuration file:
circle_pad="engine:sdl,axis_x:0,joystick:0,axis_y:1"
I haven’t found any way to change the sensitivity in the configuration file manually.

muemart:
Hm, the configuration looks correct to me. There's no sensitivity or deadzone in the emulator, so if anything it should be *too* sensitive. I assume your controller works fine in other software? Do you usually have to change sensitivity or similar things there?
There might be problems in Citra's input handling for your controller, but as far as configuration is concerned, it looks like it works as it should.

nicoboss:
After doing some research I found a solution how to fix the sensitivity issue: Open Systems Control Center => Hardware and Sound => Right click on “USB Gamepad” => Gamecontrollersettings => double-click on the joystick you use => Go to Settings tap => Press “Calibrate…” => follow the calibration assistant and now the sensitivity should work properly in Citra. This is because unlike most other Software (at example the Dolphin Emulator) Citra uses the Calibration from the OS/Driver instead of an own one. It’s no problem to configure it properly but users must know how so maybe we should add this to the FAQ. It’s funny I haven’t noticed this miss calibration for like two years and haven’t even known that you should calibrate them before first use because every software supported my N64 controller just handled this by his own… Sorry for wasting your time with something haven’t much to do with this PR. At least we know now hot to help users having the same issue.

muemart:
No problem. Cool to see that the Windows calibration thing can actually be useful!

bunnei:
@muemart Thanks for your contribution! Based on this, I look forward to seeing more submissions from you.

Getting this feature right is pretty difficult - but it seems like this is a great start and things are pretty solid. I'm going to do something *somewhat* unconventional with this project and merge as-is. I'm sure there are a few bugs here and there, but they will be reported and we will work them out - this is a hugely requested feature, and there have been some tragically unsuccessful attempts getting it done, so I think we'll do things a little differently this time 😄

Issue #2391, opened 2017-01-01
Pokémon Omega Ruby/Alpha Saphire – Frame fast forwarding issue


Dragios:
With an in-game save this issue can be reproduced all the time.
However, without an in-game save this issue does not occur.
There is a very tiny chance that it might still happen when starting a *New Game* but I can't reproduce it most of the time.

![Honeycam 2017-01-01 18-58-15](https://cloud.githubusercontent.com/assets/18496868/21581028/607b07d4-d058-11e6-9592-dbdb2a26ebef.gif)
I literally have no idea how come with an in-game save existance will trigger such issue in the first place.
Tested this on commit f7dd8d5e816c4157ae909600fd2a7d848a1c9712

Dragios:
Fixed by PR #3119

Issue #2445, opened 2017-01-16
Fix ARM11 clock rate


TuxSH:
Kernel11 uses in places such as SleepThread, to convert a time in ns to a number of ticks (then add the current tick count to the result), code like:

```cpp
      v7 = __PAIR__(
             (0x44A2FA85 * (unsigned __int64)(unsigned int)ns_high >> 32) + 0x44A2FA85 * (ns_high >> 31),
             0x44A2FA85 * ns_high)
         + (0x44A2FA85 * (unsigned __int64)(unsigned int)ns_low >> 32)// ns -> tick
         + KTimerAndWDTManager::GetSystemTick(&TimerAndWDTManager);
```

This is a multiply-by-constant division:

1 tick -> 2^32/0x44A2FA85  ns

Hence:
1/(2^32/0x44A2FA85) ticks -> 1 s

Therefore the theoritical clockrate of the 3DS's MPCore ARM11 processor is 268,111,855.956 Hz.

SijmenSchoon:
Is this issue still relevant, or has #336 already fixed it?

wwylele:
The fraction number is not necessarily an accurate number. It could be resulted from optimization from an integer division, in a similar way like this: https://godbolt.org/g/4KKs75. In this example, by doing the same math calculation purely from assembly code `input * 1459166399 / (2 ^54)`, one would conclude that the original divisor is `12345677.999320`, instead of the actual integer `12345678`. In the same way, the original kernel source code probably wrote an integer there, and due to the optimization, what we can calculate from the assembly is not the integer.

FearlessTobi:
This is fixed by #3119

Issue #2339, opened 2016-12-16
Nano Assault/Nano Assault EX – Inconsistent FPS despite running at full emulation speed


Dragios:
The video and in game emulation suffer from jittery playback even when Dxtory shows the in game FPS at almost 60 FPS unlike the real console which renders it smoothly.

_Citra_
[bandicam 2016-12-15 21-09-40-380](https://github.com/Dragios/Cloud-Drive/blob/master/bandicam%202016-12-15%2021-09-40-380.mp4?raw=true)

_Nintendo 3DS_
https://www.youtube.com/watch?v=7IXE0n4ARZU

Log for Nano Assault
http://pastebin.com/W3atjGxM

Log for Nano Assault EX
http://pastebin.com/NwfisjUv

Dragios:
So with the built in FPS profiler reported by Citra (assuming it is accurate) the game speed remains the same but the FPS keeps dropping unlike the 3DS which maintains smooth 60 FPS all the time.

Dragios:
Need PR #3091 + #3119 to fix this issue.

PR #3119, opened 2017-11-15
CoreTiming: Reworked CoreTiming

B3n30:
The current CoreTiming has an error that Events scheduled with a cycles_into_future > 0 are called way to late. 
Since thee current implementation is very convoluted and not well documented we decided to just rewrite the whole CoreTiming (adopt the current version from Dolphin).
It was tested that this version and the error isn't there anymore.

This also fixes a wrong BASE_CLOCK_RATE_ARM11 (#2445) and add the ability to jump cycles by ns (in addition to ms and us).

It also adds some overflow handling when calculation cycles out of ns/us/ms.

In a last step it adds Test for checking CoreTiming.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3119)

### DIFF
--- a/src/core/core_timing.cpp
+++ b/src/core/core_timing.cpp
@@ -12,9 +12,9 @@
 #include <unordered_map>
 #include <vector>
 #include "common/assert.h"
-#include "common/fifo_queue.h"
 #include "common/logging/log.h"
 #include "common/thread.h"
+#include "common/threadsafe_queue.h"
 
 namespace CoreTiming {
 
@@ -53,14 +53,17 @@ static std::unordered_map<std::string, EventType> event_types;
 // by the standard adaptor class.
 static std::vector<Event> event_queue;
 static u64 event_fifo_id;
-static std::mutex ts_write_lock;
-static Common::FifoQueue<Event, false> ts_queue;
+// the queue for storing the events from other threads threadsafe until they will be added
+// to the event_queue by the emu thread
+static Common::MPSCQueue<Event, false> ts_queue;
 
 static constexpr int MAX_SLICE_LENGTH = 20000;
 
 static s64 idled_cycles;
 
 // Are we in a function that has been called from Advance()
+// If events are sheduled from a function that gets called from Advance(),
+// don't change slice_length and downcount.
 static bool is_global_timer_sane;
 
 static EventType* ev_lost = nullptr;
@@ -103,7 +106,6 @@ void Init() {
 }
 
 void Shutdown() {
-    std::lock_guard<std::mutex> lock(ts_write_lock);
     MoveEvents();
     ClearPendingEvents();
     UnregisterAllEvents();
@@ -132,7 +134,7 @@ void ClearPendingEvents() {
 }
 
 void ScheduleEvent(s64 cycles_into_future, const EventType* event_type, u64 userdata) {
-    ASSERT_MSG(event_type, "Event type is nullptr, will crash now.");
+    ASSERT(event_type != nullptr);
     s64 timeout = GetTicks() + cycles_into_future;
 
     // If this event needs to be scheduled before the next advance(), force one early
@@ -144,7 +146,6 @@ void ScheduleEvent(s64 cycles_into_future, const EventType* event_type, u64 user
 }
 
 void ScheduleEventThreadsafe(s64 cycles_into_future, const EventType* event_type, u64 userdata) {
-    std::lock_guard<std::mutex> lock(ts_write_lock);
     ts_queue.Push(Event{global_timer + cycles_into_future, 0, userdata, event_type});
 }
 
@@ -171,7 +172,7 @@ void RemoveEvent(const EventType* event_type) {
     }
 }
 
-void RemoveAllEvents(const EventType* event_type) {
+void RemoveNormalAndThreadsafeEvent(const EventType* event_type) {
     MoveEvents();
     RemoveEvent(event_type);
 }
--- a/src/core/core_timing.cpp
+++ b/src/core/core_timing.cpp
@@ -8,6 +8,7 @@
 #include <cinttypes>
 #include <mutex>
 #include <string>
+#include <tuple>
 #include <unordered_map>
 #include <vector>
 #include "common/assert.h"
@@ -30,7 +31,7 @@ struct Event {
     s64 time;
     u64 fifo_order;
     u64 userdata;
-    EventType* type;
+    const EventType* type;
 };
 
 // Sort by time, unless the times are the same, in which case sort by the order added to the queue
@@ -93,7 +94,7 @@ void Init() {
 
     // The time between CoreTiming being intialized and the first call to Advance() is considered
     // the slice boundary between slice -1 and slice 0. Dispatcher loops must call Advance() before
-    // executing the first PPC cycle of each slice to prepare the slice length and downcount for
+    // executing the first cycle of each slice to prepare the slice length and downcount for
     // that slice.
     is_global_timer_sane = true;
 
@@ -130,7 +131,7 @@ void ClearPendingEvents() {
     event_queue.clear();
 }
 
-void ScheduleEvent(s64 cycles_into_future, EventType* event_type, u64 userdata) {
+void ScheduleEvent(s64 cycles_into_future, const EventType* event_type, u64 userdata) {
     ASSERT_MSG(event_type, "Event type is nullptr, will crash now.");
     s64 timeout = GetTicks() + cycles_into_future;
 
@@ -142,12 +143,12 @@ void ScheduleEvent(s64 cycles_into_future, EventType* event_type, u64 userdata) 
     std::push_heap(event_queue.begin(), event_queue.end(), std::greater<Event>());
 }
 
-void ScheduleEventThreadsafe(s64 cycles_into_future, EventType* event_type, u64 userdata) {
+void ScheduleEventThreadsafe(s64 cycles_into_future, const EventType* event_type, u64 userdata) {
     std::lock_guard<std::mutex> lock(ts_write_lock);
     ts_queue.Push(Event{global_timer + cycles_into_future, 0, userdata, event_type});
 }
 
-void UnscheduleEvent(EventType* event_type, u64 userdata) {
+void UnscheduleEvent(const EventType* event_type, u64 userdata) {
     auto itr = std::remove_if(event_queue.begin(), event_queue.end(), [&](const Event& e) {
         return e.type == event_type && e.userdata == userdata;
     });
@@ -159,7 +160,7 @@ void UnscheduleEvent(EventType* event_type, u64 userdata) {
     }
 }
 
-void RemoveEvent(EventType* event_type) {
+void RemoveEvent(const EventType* event_type) {
     auto itr = std::remove_if(event_queue.begin(), event_queue.end(),
                               [&](const Event& e) { return e.type == event_type; });
 
@@ -170,7 +171,7 @@ void RemoveEvent(EventType* event_type) {
     }
 }
 
-void RemoveAllEvents(EventType* event_type) {
+void RemoveAllEvents(const EventType* event_type) {
     MoveEvents();
     RemoveEvent(event_type);
 }
@@ -196,8 +197,8 @@ void MoveEvents() {
 void Advance() {
     MoveEvents();
 
-    int cyclesExecuted = slice_length - downcount;
-    global_timer += cyclesExecuted;
+    int cycles_executed = slice_length - downcount;
+    global_timer += cycles_executed;
     slice_length = MAX_SLICE_LENGTH;
 
     is_global_timer_sane = true;
--- a/src/core/core_timing.cpp
+++ b/src/core/core_timing.cpp
@@ -1,562 +1,236 @@
-// Copyright (c) 2012- PPSSPP Project / Dolphin Project.
-// Licensed under GPLv2 or any later version
+// Copyright 2008 Dolphin Emulator Project / 2017 Citra Emulator Project
+// Licensed under GPLv2+
 // Refer to the license.txt file included.
 
-#include <atomic>
-#include <cinttypes>
-#include <mutex>
-#include <vector>
-#include "common/chunk_file.h"
-#include "common/logging/log.h"
-#include "common/string_util.h"
-#include "core/arm/arm_interface.h"
-#include "core/core.h"
 #include "core/core_timing.h"
 
-int g_clock_rate_arm11 = BASE_CLOCK_RATE_ARM11;
-
-// is this really necessary?
-#define INITIAL_SLICE_LENGTH 20000
-#define MAX_SLICE_LENGTH 100000000
+#include <algorithm>
+#include <cinttypes>
+#include <mutex>
+#include <string>
+#include <unordered_map>
+#include <vector>
+#include "common/assert.h"
+#include "common/fifo_queue.h"
+#include "common/logging/log.h"
+#include "common/thread.h"
 
 namespace CoreTiming {
-struct EventType {
-    EventType() {}
-
-    EventType(TimedCallback cb, const char* n) : callback(cb), name(n) {}
-
-    TimedCallback callback;
-    const char* name;
-};
-
-static std::vector<EventType> event_types;
-
-struct BaseEvent {
-    s64 time;
-    u64 userdata;
-    int type;
-};
-
-typedef LinkedListItem<BaseEvent> Event;
-
-static Event* first;
-static Event* ts_first;
-static Event* ts_last;
-
-// event pools
-static Event* event_pool = nullptr;
-static Event* event_ts_pool = nullptr;
-static int allocated_ts_events = 0;
-// Optimization to skip MoveEvents when possible.
-static std::atomic<bool> has_ts_events(false);
-
-int g_slice_length;
 
 static s64 global_timer;
+static int slice_length;
+static int downcount;
+
+struct EventType {
+    TimedCallback callback;
+    const std::string* name;
+};
+
+struct Event {
+    s64 time;
+    u64 fifo_order;
+    u64 userdata;
+    EventType* type;
+};
+
+// Sort by time, unless the times are the same, in which case sort by the order added to the queue
+static bool operator>(const Event& left, const Event& right) {
+    return std::tie(left.time, left.fifo_order) > std::tie(right.time, right.fifo_order);
+}
+
+static bool operator<(const Event& left, const Event& right) {
+    return std::tie(left.time, left.fifo_order) < std::tie(right.time, right.fifo_order);
+}
+
+// unordered_map stores each element separately as a linked list node so pointers to elements
+// remain stable regardless of rehashes/resizing.
+static std::unordered_map<std::string, EventType> event_types;
+
+// The queue is a min-heap using std::make_heap/push_heap/pop_heap.
+// We don't use std::priority_queue because we need to be able to serialize, unserialize and
+// erase arbitrary events (RemoveEvent()) regardless of the queue order. These aren't accomodated
+// by the standard adaptor class.
+static std::vector<Event> event_queue;
+static u64 event_fifo_id;
+static std::mutex ts_write_lock;
+static Common::FifoQueue<Event, false> ts_queue;
+
+static constexpr int MAX_SLICE_LENGTH = 20000;
+
 static s64 idled_cycles;
-static s64 last_global_time_ticks;
-static s64 last_global_time_us;
 
-static s64 down_count = 0; ///< A decreasing counter of remaining cycles before the next event,
-                           /// decreased by the cpu run loop
+// Are we in a function that has been called from Advance()
+static bool is_global_timer_sane;
 
-static std::recursive_mutex external_event_section;
+static EventType* ev_lost = nullptr;
 
-// Warning: not included in save state.
-using AdvanceCallback = void(int cycles_executed);
-static AdvanceCallback* advance_callback = nullptr;
-static std::vector<MHzChangeCallback> mhz_change_callbacks;
+static void EmptyTimedCallback(u64 userdata, s64 cyclesLate) {}
 
-static void FireMhzChange() {
-    for (auto callback : mhz_change_callbacks)
-        callback();
-}
+EventType* RegisterEvent(const std::string& name, TimedCallback callback) {
+    // check for existing type with same name.
+    // we want event type names to remain unique so that we can use them for serialization.
+    ASSERT_MSG(event_types.find(name) == event_types.end(),
+               "CoreTiming Event \"%s\" is already registered. Events should only be registered "
+               "during Init to avoid breaking save states.",
+               name.c_str());
 
-void SetClockFrequencyMHz(int cpu_mhz) {
-    // When the mhz changes, we keep track of what "time" it was before hand.
-    // This way, time always moves forward, even if mhz is changed.
-    last_global_time_us = GetGlobalTimeUs();
-    last_global_time_ticks = GetTicks();
-
-    g_clock_rate_arm11 = cpu_mhz * 1000000;
-    // TODO: Rescale times of scheduled events?
-
-    FireMhzChange();
-}
-
-int GetClockFrequencyMHz() {
-    return g_clock_rate_arm11 / 1000000;
-}
-
-u64 GetGlobalTimeUs() {
-    s64 ticks_since_last = GetTicks() - last_global_time_ticks;
-    int freq = GetClockFrequencyMHz();
-    s64 us_since_last = ticks_since_last / freq;
-    return last_global_time_us + us_since_last;
-}
-
-static Event* GetNewEvent() {
-    if (!event_pool)
-        return new Event;
-
-    Event* event = event_pool;
-    event_pool = event->next;
-    return event;
-}
-
-static Event* GetNewTsEvent() {
-    allocated_ts_events++;
-
-    if (!event_ts_pool)
-        return new Event;
-
-    Event* event = event_ts_pool;
-    event_ts_pool = event->next;
-    return event;
-}
-
-static void FreeEvent(Event* event) {
-    event->next = event_pool;
-    event_pool = event;
-}
-
-static void FreeTsEvent(Event* event) {
-    event->next = event_ts_pool;
-    event_ts_pool = event;
-    allocated_ts_events--;
-}
-
-int RegisterEvent(const char* name, TimedCallback callback) {
-    event_types.emplace_back(callback, name);
-    return (int)event_types.size() - 1;
-}
-
-static void AntiCrashCallback(u64 userdata, int cycles_late) {
-    LOG_CRITICAL(Core_Timing, "Savestate broken: an unregistered event was called.");
-}
-
-void RestoreRegisterEvent(int event_type, const char* name, TimedCallback callback) {
-    if (event_type >= (int)event_types.size())
-        event_types.resize(event_type + 1, EventType(AntiCrashCallback, "INVALID EVENT"));
-
-    event_types[event_type] = EventType(callback, name);
+    auto info = event_types.emplace(name, EventType{callback, nullptr});
+    EventType* event_type = &info.first->second;
+    event_type->name = &info.first->first;
+    return event_type;
 }
 
 void UnregisterAllEvents() {
-    if (first)
-        LOG_ERROR(Core_Timing, "Cannot unregister events with events pending");
+    ASSERT_MSG(event_queue.empty(), "Cannot unregister events with events pending");
     event_types.clear();
 }
 
 void Init() {
-    down_count = INITIAL_SLICE_LENGTH;
-    g_slice_length = INITIAL_SLICE_LENGTH;
+    downcount = MAX_SLICE_LENGTH;
+    slice_length = MAX_SLICE_LENGTH;
     global_timer = 0;
     idled_cycles = 0;
-    last_global_time_ticks = 0;
-    last_global_time_us = 0;
-    has_ts_events = 0;
-    mhz_change_callbacks.clear();
 
-    first = nullptr;
-    ts_first = nullptr;
-    ts_last = nullptr;
+    // The time between CoreTiming being intialized and the first call to Advance() is considered
+    // the slice boundary between slice -1 and slice 0. Dispatcher loops must call Advance() before
+    // executing the first PPC cycle of each slice to prepare the slice length and downcount for
+    // that slice.
+    is_global_timer_sane = true;
 
-    event_pool = nullptr;
-    event_ts_pool = nullptr;
-    allocated_ts_events = 0;
-
-    advance_callback = nullptr;
+    event_fifo_id = 0;
+    ev_lost = RegisterEvent("_lost_event", &EmptyTimedCallback);
 }
 
 void Shutdown() {
+    std::lock_guard<std::mutex> lock(ts_write_lock);
     MoveEvents();
     ClearPendingEvents();
     UnregisterAllEvents();
+}
 
-    while (event_pool) {
-        Event* event = event_pool;
-        event_pool = event->next;
-        delete event;
+// This should only be called from the CPU thread. If you are calling
+// it from any other thread, you are doing something evil
+u64 GetTicks() {
+    u64 ticks = static_cast<u64>(global_timer);
+    if (!is_global_timer_sane) {
+        ticks += slice_length - downcount;
     }
-
-    std::lock_guard<std::recursive_mutex> lock(external_event_section);
-    while (event_ts_pool) {
-        Event* event = event_ts_pool;
-        event_ts_pool = event->next;
-        delete event;
-    }
+    return ticks;
 }
 
 void AddTicks(u64 ticks) {
-    down_count -= ticks;
-    if (down_count < 0) {
-        Advance();
-    }
-}
-
-u64 GetTicks() {
-    return (u64)global_timer + g_slice_length - down_count;
+    downcount -= ticks;
 }
 
 u64 GetIdleTicks() {
-    return (u64)idled_cycles;
-}
-
-// This is to be called when outside threads, such as the graphics thread, wants to
-// schedule things to be executed on the main thread.
-void ScheduleEvent_Threadsafe(s64 cycles_into_future, int event_type, u64 userdata) {
-    std::lock_guard<std::recursive_mutex> lock(external_event_section);
-    Event* new_event = GetNewTsEvent();
-    new_event->time = GetTicks() + cycles_into_future;
-    new_event->type = event_type;
-    new_event->next = nullptr;
-    new_event->userdata = userdata;
-    if (!ts_first)
-        ts_first = new_event;
-    if (ts_last)
-        ts_last->next = new_event;
-    ts_last = new_event;
-
-    has_ts_events = true;
-}
-
-// Same as ScheduleEvent_Threadsafe(0, ...) EXCEPT if we are already on the CPU thread
-// in which case the event will get handled immediately, before returning.
-void ScheduleEvent_Threadsafe_Immediate(int event_type, u64 userdata) {
-    if (false) // Core::IsCPUThread())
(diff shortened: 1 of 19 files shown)


src/core/hle/service/ir/extra_hid.h
```diff
@@ -6,6 +6,7 @@
 
 #include <array>
 #include <atomic>
+#include "core/core_timing.h"
```
lioncash: This can be a forward declaration.

src/core/core_timing.cpp
```diff
@@ -1,562 +1,236 @@
...
-        const int diff = target - g_slice_length;
-        g_slice_length += diff;
-        down_count += diff;
-    }
-    if (advance_callback)
-        advance_callback(static_cast<int>(cycles_executed));
-}
+    MoveEvents();
+
+    int cyclesExecuted = slice_length - downcount;
```
lioncash: `cycles_executed`

src/core/core_timing.cpp
```diff
@@ -1,562 +1,236 @@
...
+// Licensed under GPLv2+
 // Refer to the license.txt file included.
 
-#include <atomic>
+#include "core/core_timing.h"
+
+#include <algorithm>
 #include <cinttypes>
 #include <mutex>
+#include <string>
```
lioncash: `<tuple>` needs to be included for `std::tie`

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
-    return (g_clock_rate_arm11 / 1000000 * (s64)us);
+    return (BASE_CLOCK_RATE_ARM11 * static_cast<s64>(us) / 1000000);
 }
 
 inline s64 usToCycles(s64 us) {
-    return (g_clock_rate_arm11 / 1000000 * us);
+    if ((us / 1000000) > MAX_VALUE_TO_MULTIPLY) {
+        LOG_ERROR(Core_Timing, "Integer overflow, use max value");
+        return std::numeric_limits<s64>::max();
+    } else if (us > MAX_VALUE_TO_MULTIPLY) {
```
lioncash: `else` isn't necessary here, since the if portion contains a return.

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+        LOG_DEBUG(Core_Timing, "Time very big, do rounding");
+        return (BASE_CLOCK_RATE_ARM11 * (us / 1000000));
```
lioncash: `outer parentheses aren't necessary, the inner ones are also inconsistent with the return statement below.

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+    }
+    return (BASE_CLOCK_RATE_ARM11 * us / 1000000);
```
lioncash: outer parentheses aren't necessary.

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
 }
 
 inline s64 usToCycles(u64 us) {
-    return (s64)(g_clock_rate_arm11 / 1000000 * us);
+    if ((us / 1000000) > MAX_VALUE_TO_MULTIPLY) {
+        LOG_ERROR(Core_Timing, "Integer overflow, use max value");
+        return std::numeric_limits<s64>::max();
+    } else if (us > MAX_VALUE_TO_MULTIPLY) {
```
lioncash: Ditto about `else`

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+        LOG_DEBUG(Core_Timing, "Time very big, do rounding");
+        return (BASE_CLOCK_RATE_ARM11 * static_cast<s64>(us / 1000000));
```
lioncash: Ditto oabout outer parentheses and inconsistencies with the below return statement.

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+    }
+    return (BASE_CLOCK_RATE_ARM11 * static_cast<s64>(us) / 1000000);
```
lioncash: Ditto about outer parentheses

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+}
+
+inline s64 nsToCycles(float ns) {
+    return static_cast<s64>(BASE_CLOCK_RATE_ARM11 * ns * (0.000000001f));
```
lioncash: `0.000000001f` doesn't need to be parenthesized

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+}
+
+inline s64 nsToCycles(int ns) {
+    return (BASE_CLOCK_RATE_ARM11 * static_cast<s64>(ns) / 1000000000);
```
lioncash: The outermost parentheses are unnecessary

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+}
+
+inline s64 nsToCycles(s64 ns) {
+    if ((ns / 1000000000) > MAX_VALUE_TO_MULTIPLY) {
+        LOG_ERROR(Core_Timing, "Integer overflow, use max value");
+        return std::numeric_limits<s64>::max();
+    } else if (ns > MAX_VALUE_TO_MULTIPLY) {
```
lioncash: `else` isn't necessary here, since the `if` portion returns

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+        LOG_DEBUG(Core_Timing, "Time very big, do rounding");
+        return (BASE_CLOCK_RATE_ARM11 * (ns / 1000000000));
```
lioncash: Outermost parentheses aren't necessary. Innermost parentheses are inconsistent with the below return statement too.

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+    }
+    return (BASE_CLOCK_RATE_ARM11 * ns / 1000000000);
```
lioncash: Outer parentheses aren't necessary.

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+}
+
+inline s64 nsToCycles(u64 ns) {
+    if ((ns / 1000000000) > MAX_VALUE_TO_MULTIPLY) {
+        LOG_ERROR(Core_Timing, "Integer overflow, use max value");
+        return std::numeric_limits<s64>::max();
+    } else if (ns > MAX_VALUE_TO_MULTIPLY) {
```
lioncash: Ditto about `else` being unnecessary

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+        LOG_DEBUG(Core_Timing, "Time very big, do rounding");
+        return (BASE_CLOCK_RATE_ARM11 * static_cast<s64>(ns / 1000000000));
```
lioncash: Ditto about outer parentheses being unnecessary

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
+    }
+    return (BASE_CLOCK_RATE_ARM11 * static_cast<s64>(ns) / 1000000000);
```
lioncash: Ditto about outer parentheses being unnecessary

src/core/core_timing.h
```diff
@@ -1,144 +1,190 @@
...
 }
 
 inline s64 usToCycles(int us) {
-    return (g_clock_rate_arm11 / 1000000 * (s64)us);
+    return (BASE_CLOCK_RATE_ARM11 * static_cast<s64>(us) / 1000000);
 }
 
 inline s64 usToCycles(s64 us) {
-    return (g_clock_rate_arm11 / 1000000 * us);
+    if ((us / 1000000) > MAX_VALUE_TO_MULTIPLY) {
```
lioncash: The inner parentheses here are unnecessary, ditto for similar occurrences below.

src/core/hle/kernel/thread.cpp
```diff
@@ -194,6 +194,7 @@ static void ThreadWakeupCallback(u64 thread_handle, int cycles_late) {
         LOG_CRITICAL(Kernel, "Callback fired for invalid thread %08X", (Handle)thread_handle);
         return;
     }
+    LOG_ERROR(Kernel, "%d", cycles_late);
```
lioncash: Why is this logging as an error? This should also probably provide more context than just printing out a number (like saying what the number actually means or is).

src/core/core_timing.cpp
```diff
@@ -1,562 +1,236 @@
...
-    while (ts_first) {
-        if (ts_first->type == event_type) {
-            Event* next = ts_first->next;
-            FreeTsEvent(ts_first);
-            ts_first = next;
-        } else {
-            break;
-        }
-    }
+void RemoveEvent(EventType* event_type) {
```
lioncash: `event_type` can be made `const EventType*`

src/core/core_timing.cpp
```diff
@@ -1,562 +1,236 @@
...
+    // Removing random items breaks the invariant so we have to re-establish it.
+    if (itr != event_queue.end()) {
+        event_queue.erase(itr, event_queue.end());
+        std::make_heap(event_queue.begin(), event_queue.end(), std::greater<Event>());
     }
 }
 
-void RemoveAllEvents(int event_type) {
-    RemoveThreadsafeEvent(event_type);
+void RemoveAllEvents(EventType* event_type) {
```
lioncash: If `RemoveEvent`'s parameter is made `const`, this can also be `const EventType*`

src/core/core_timing.cpp
```diff
@@ -1,562 +1,236 @@
...
-    advance_callback = callback;
+void ScheduleEventThreadsafe(s64 cycles_into_future, EventType* event_type, u64 userdata) {
+    std::lock_guard<std::mutex> lock(ts_write_lock);
+    ts_queue.Push(Event{global_timer + cycles_into_future, 0, userdata, event_type});
 }
 
-void RegisterMHzChangeCallback(MHzChangeCallback callback) {
-    mhz_change_callbacks.push_back(callback);
-}
+void UnscheduleEvent(EventType* event_type, u64 userdata) {
```
lioncash: `event_type` can be `const EventType*`

src/core/core_timing.cpp
```diff
@@ -1,562 +1,236 @@
...
 
-static std::vector<EventType> event_types;
-
-struct BaseEvent {
+struct Event {
     s64 time;
+    u64 fifo_order;
     u64 userdata;
-    int type;
+    EventType* type;
```
lioncash: I believe this can be made `const`

src/core/core_timing.cpp
```diff
@@ -1,562 +1,236 @@
...
-            result = ts_first->time - GetTicks();
-
-            Event* next = ts_first->next;
-            FreeTsEvent(ts_first);
-            ts_first = next;
-        } else {
-            break;
-        }
-    }
+void ScheduleEvent(s64 cycles_into_future, EventType* event_type, u64 userdata) {
```
lioncash: `event_type` can be made const here if the struct member is `const`

src/core/core_timing.h
B3n30: this inconsistent parenthesis is intended to avoid the integer overflow

Soban44444:
Fixes the Pokemon intro ,didn't find errors in any games.

Dragios:
Fixes #2391 and #2339 (it requires this PR + #3091 to work)

dampih:
This PR make Citra crash when booting **Kirby Battle Royale** 
canary_error

yuriks:
:lgtm: I didn't look *too* closely at queue code and the test code, since I assumed those were just copied from Dolphin with few modifications.

---

Reviewed 14 of 18 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 34 unresolved discussions.

---

*[src/common/fifo_queue.h, line 18 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSKyQ70Ap5CRtgj2BD:-KzSKyQ70Ap5CRtgj2BE:b-o18kw4) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/common/fifo_queue.h#L18)):*
> ```objc
> namespace Common {
> template <typename T, bool NeedSize = true>
> class FifoQueue {
> ```

Isn't a queue by definition FIFO? Saying that the queue is SPSC (Single-Producer, Single-Consumer) in the name instead seems more informative.

---

*[src/core/core_timing.cpp, line 57 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSLgi2Mn6PiKE9sW36:-KzSLgi3IqzH0xaJ7x59:b-3178rz) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.cpp#L57)):*
> ```cpp
> static u64 event_fifo_id;
> static std::mutex ts_write_lock;
> static Common::FifoQueue<Event, false> ts_queue;
> ```

What does "ts" stand for? Also, the combination of the SPSC FifoQueue + the write mutex makes it effectively a MPSC queue, so maybe wrapping these into a class that enforces this would prevent accidental mis-use.

---

*[src/core/core_timing.cpp, line 64 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSM4l-nmUcLUEzlbQj:-KzSM4l-nmUcLUEzlbQk:bqa1i7v) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.cpp#L64)):*
> ```cpp
> 
> // Are we in a function that has been called from Advance()
> static bool is_global_timer_sane;
> ```

This needs more explanation. What does it actually mean, and what is it used for?

---

*[src/core/core_timing.cpp, line 135 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSMa-RR5G7yJwzhMOg:-KzSMa-TNE-eRMICfvBI:bsoxcy7) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.cpp#L135)):*
> ```cpp
> 
> void ScheduleEvent(s64 cycles_into_future, const EventType* event_type, u64 userdata) {
>     ASSERT_MSG(event_type, "Event type is nullptr, will crash now.");
> ```

This assert message is quite pointless... I'd just replace it with `ASSERT(event_type != nullptr)`.

---

*[src/core/core_timing.cpp, line 209 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSPU2l4fZkVajpshHI:-KzSPU2mOezaruh7JLz_:b-uf3vve) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.cpp#L209)):*
> ```cpp
>         Event evt = std::move(event_queue.front());
>         std::pop_heap(event_queue.begin(), event_queue.end(), std::greater<Event>());
>         event_queue.pop_back();
> ```

nit: Could also do pop_heap(...) first before moving from event_queue.back() and then pop_back() immediately afterwards. It's a bit clearer for me, but probably doesn't matter much.

---

*[src/core/core_timing.h, line 60 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-60:-KzSSYje7oVYCK51cSBX:b-7wdetk) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L60)):*

this inconsistent parenthesis is intended to avoid the integer overflow
</blockquote></details>

In this case I'd make the parenthesis explicit on the bottom return too to emphasize it.

---

*[src/core/core_timing.h, line 25 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSRgVfMJNU9VtLnMap:-KzSRgVgWCWzgHajiH20:bqvqsrh) ([raw file](https://github.com/citra-emu/citra/blob/f55f4fc6841c40a616dd771b57ab1005650d7b37/src/core/core_timing.h#L25)):*
> ```objc
> 
> constexpr int BASE_CLOCK_RATE_ARM11 = 268123480;
> extern int g_clock_rate_arm11;
> ```

Since we're dropping support for changing clock rates, do we have any plan on how to support n3DS clock-rate switching in the future? (I suspect we can probably just always run at the n3DS clock rate, it shouldn't really matter, but needs to be given some thought I guess.)

---

*[src/core/core_timing.h, line 39 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSSA-0RMaOJ2eikxaP:-KzSSA-0RMaOJ2eikxaQ:b-9xr6yq) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.h#L39)):*
> ```objc
> 
> inline s64 msToCycles(float ms) {
>     return static_cast<s64>(BASE_CLOCK_RATE_ARM11 * ms * (0.001f));
> ```

Multiplying the float suffixes on the left should let the compiler per-compute BASE_BLOCK_RATE_ARM11 * 0.001f (etc.) and make it marginally faster and more precise.

---

*[src/core/core_timing.h, line 144 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSTF2xEBUp0JCTNtiC:-KzSTF2xEBUp0JCTNtiD:b-sfhg9p) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.h#L144)):*
> if name is not unique, an existing event_type will be

This is not true, it asserts that there aren't duplicate names.

---

*[src/core/core_timing.h, line 167 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSTvWPEW3kvTfD1UqW:-KzSTvWPEW3kvTfD1UqX:b-i9ml7z) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.h#L167)):*
> ```objc
> void UnscheduleEvent(const EventType* event_type, u64 userdata);
> 
> /// We only permit one event of each type in the queue at a time.
> ```

If that's the case, then what's the difference between RemoveEvent and RemoveAllEvents? (answer: RemoveAllEvents also looks at the multithread queue. Should be renamed.)

---

*[src/core/hle/applets/applet.cpp, line 41 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSUmhRccPATRqyxjaU:-KzSUmhRccPATRqyxjaV:b-kj97h6) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/hle/applets/applet.cpp#L41)):*
> ```cpp
> 
> static std::unordered_map<Service::APT::AppletId, std::shared_ptr<Applet>> applets;
> static CoreTiming::EventType* applet_update_event = nullptr;
> ```

You nixed the comment (I think it's still relevant.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3119#-:-KzSb-4B2c4NST3BajWh:bwc4zou)*

B3n30:
Review status: all files reviewed at latest revision, 34 unresolved discussions.

---

*[src/common/fifo_queue.h, line 18 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSKyQ70Ap5CRtgj2BD:-KzTyzLVk2uLPNWHg9Tr:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/common/fifo_queue.h#L18)):*

Isn't a queue by definition FIFO? Saying that the queue is SPSC (Single-Producer, Single-Consumer) in the name instead seems more informative.
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 10 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70Q-r1-10:-KzTdIq2nYEPu69vB8St:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.cpp#L10)):*

`<tuple>` needs to be included for `std::tie`
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 33 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70Q-r1-33:-KzTdK5QJbpTPavr36vE:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.cpp#L33)):*

I believe this can be made `const`
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 133 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70Q-r1-133:-KzTsssi4Q3KefT-82ke:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.cpp#L133)):*

`event_type` can be made const here if the struct member is `const`
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 150 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70Q-r1-150:-KzTswbQIyd-QcCagw-_:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.cpp#L150)):*

`event_type` can be `const EventType*`
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 162 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70Q-r1-162:-KzTsvtAbTP-ML3iweAO:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.cpp#L162)):*

`event_type` can be made `const EventType*`
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 173 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70Q-r1-173:-KzTsyUykMih0PSyBxtF:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.cpp#L173)):*

If `RemoveEvent`'s parameter is made `const`, this can also be `const EventType*`
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 199 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70Q-r1-199:-KzTszH0ygM_IufE8YPO:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.cpp#L199)):*

`cycles_executed`
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 57 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSLgi2Mn6PiKE9sW36:-KzTzGmeHzUnHD33kW8Y:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.cpp#L57)):*

What does "ts" stand for? Also, the combination of the SPSC FifoQueue + the write mutex makes it effectively a MPSC queue, so maybe wrapping these into a class that enforces this would prevent accidental mis-use.
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 64 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSM4l-nmUcLUEzlbQj:-KzTzxfxo1myRD8C_yU8:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.cpp#L64)):*

This needs more explanation. What does it actually mean, and what is it used for?
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 135 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSMa-RR5G7yJwzhMOg:-KzU-bHZFNU5sP1S83u7:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.cpp#L135)):*

This assert message is quite pointless... I'd just replace it with `ASSERT(event_type != nullptr)`.
</blockquote></details>

Done.

---

*[src/core/core_timing.cpp, line 209 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSPU2l4fZkVajpshHI:-KzTtCO51JdIRY4nhtMx:b-lra58h) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.cpp#L209)):*

nit: Could also do pop_heap(...) first before moving from event_queue.back() and then pop_back() immediately afterwards. It's a bit clearer for me, but probably doesn't matter much.
</blockquote></details>

it's moving from the front and not from the back, thus the pop_heap is after the move, to sort the moved event to the end and then you can do the pop_back

---

*[src/core/core_timing.h, line 55 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-55:-KzU0qsT0l-KncayeV-r:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L55)):*

The inner parentheses here are unnecessary, ditto for similar occurrences below.
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 58 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-58:-KzU0rh0fLSzoGn5cgjI:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L58)):*

`else` isn't necessary here, since the if portion contains a return.
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 60 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-60:-KzU18SOh7WzIq6db5I1:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L60)):*

In this case I'd make the parenthesis explicit on the bottom return too to emphasize it.
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 62 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-62:-KzU193I7rQUw_ctyTdm:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L62)):*

outer parentheses aren't necessary.
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 69 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-69:-KzU1B_44g0VrnKc2YT0:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L69)):*

Ditto about `else`
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 71 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-71:-KzU1Cpca-pxCVYJnZzK:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L71)):*

Ditto oabout outer parentheses and inconsistencies with the below return statement.
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 73 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-73:-KzU1DjXqHxfYf6aHE-z:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L73)):*

Ditto about outer parentheses
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 77 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-77:-KzU1ECTyALUxyinK4kg:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L77)):*

`0.000000001f` doesn't need to be parenthesized
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 81 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-81:-KzU1E_IAiGon2sHUluy:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L81)):*

The outermost parentheses are unnecessary
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 88 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-88:-KzU1F9VzkdHqydknf1U:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L88)):*

`else` isn't necessary here, since the `if` portion returns
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 90 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-90:-KzU1FaKlQWCaSCqAYiP:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L90)):*

Outermost parentheses aren't necessary. Innermost parentheses are inconsistent with the below return statement too.
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 92 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-92:-KzU1G99WGgNx-AxOIey:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L92)):*

Outer parentheses aren't necessary.
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 99 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-99:-KzU1Gr5GbBgPCASGwXd:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L99)):*

Ditto about `else` being unnecessary
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 101 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-101:-KzU1HJnRDfWRvMSdwKf:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L101)):*

Ditto about outer parentheses being unnecessary
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 103 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70R-r1-103:-KzU1Hmik2imw-FpKvzk:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/core_timing.h#L103)):*

Ditto about outer parentheses being unnecessary
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 25 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSRgVfMJNU9VtLnMap:-KzU0YAVzs5dZ4yswa7n:b-xf2cwx) ([raw file](https://github.com/citra-emu/citra/blob/f55f4fc6841c40a616dd771b57ab1005650d7b37/src/core/core_timing.h#L25)):*

Since we're dropping support for changing clock rates, do we have any plan on how to support n3DS clock-rate switching in the future? (I suspect we can probably just always run at the n3DS clock rate, it shouldn't really matter, but needs to be given some thought I guess.)
</blockquote></details>

Adding the support for switching clock_rates can be done very easily since the original dolphin code has a similar approach to support overclocking. We can adopt that code when we need it.

---

*[src/core/core_timing.h, line 39 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSSA-0RMaOJ2eikxaP:-KzU02jRFI6vtCfnPwZJ:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.h#L39)):*

Multiplying the float suffixes on the left should let the compiler per-compute BASE_BLOCK_RATE_ARM11 * 0.001f (etc.) and make it marginally faster and more precise.
</blockquote></details>

Done.

---

*[src/core/core_timing.h, line 144 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSTF2xEBUp0JCTNtiC:-KzU1YvHxeRvO0y9UNZd:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/core_timing.h#L144)):*

> if name is not unique, an existing event_type will be

This is not true, it asserts that there aren't duplicate names.
</blockquote></details>

Done.

---

*[src/core/hle/applets/applet.cpp, line 41 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3119#-KzSUmhRccPATRqyxjaU:-KzU832VWKpOl6fQAyeb:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/cc51c51992920ce00bdd0b5c028738d627fcf66e/src/core/hle/applets/applet.cpp#L41)):*

You nixed the comment (I think it's still relevant.)
</blockquote></details>

Done.

---

*[src/core/hle/kernel/thread.cpp, line 197 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70T-r1-197:-KzU83s6Qi8g92tzKcq9:b-eej8w6) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/hle/kernel/thread.cpp#L197)):*

Why is this logging as an error? This should also probably provide more context than just printing out a number (like saying what the number actually means or is).
</blockquote></details>

Was a remainder from my debugging

---

*[src/core/hle/service/ir/extra_hid.h, line 9 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3119#-Kz0Bi4U0RssYFa2r70X-r1-9:-KzU8KrUNO-wrhjLN7MG:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/337a5834318e4ede112e0d4ecdaa05292f8582be/src/core/hle/service/ir/extra_hid.h#L9)):*

This can be a forward declaration.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3119)*

PR #3120, opened 2017-11-16
file_sys/ncch_container: Expose ncch_offset during instantiation and in OpenFile

shinyquagsire23:
Paves the way for proper NCSD loading, NCSD partition parsing in NCCH archives (for the manual applet, DLP, maybe update partitions), parsing of NCCHs held within CIAs (for frontend parsing, possible loaders). Pretty much anything with an NCCH (or several) not at 0.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3120)

### DIFF
--- a/src/core/file_sys/ncch_container.cpp
+++ b/src/core/file_sys/ncch_container.cpp
@@ -95,12 +95,14 @@ static bool LZSS_Decompress(const u8* compressed, u32 compressed_size, u8* decom
     return true;
 }
 
-NCCHContainer::NCCHContainer(const std::string& filepath) : filepath(filepath) {
+NCCHContainer::NCCHContainer(const std::string& filepath, u32 ncch_offset)
+    : filepath(filepath), ncch_offset(ncch_offset) {
     file = FileUtil::IOFile(filepath, "rb");
 }
 
-Loader::ResultStatus NCCHContainer::OpenFile(const std::string& filepath) {
+Loader::ResultStatus NCCHContainer::OpenFile(const std::string& filepath, u32 ncch_offset) {
     this->filepath = filepath;
+    this->ncch_offset = ncch_offset;
     file = FileUtil::IOFile(filepath, "rb");
 
     if (!file.IsOpen()) {
@@ -118,7 +120,7 @@ Loader::ResultStatus NCCHContainer::Load() {
 
     if (file.IsOpen()) {
         // Reset read pointer in case this file has been read before.
-        file.Seek(0, SEEK_SET);
+        file.Seek(ncch_offset, SEEK_SET);
 
         if (file.ReadBytes(&ncch_header, sizeof(NCCH_Header)) != sizeof(NCCH_Header))
             return Loader::ResultStatus::Error;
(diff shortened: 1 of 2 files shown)


src/core/file_sys/ncch_container.cpp
```diff
@@ -118,15 +120,15 @@ Loader::ResultStatus NCCHContainer::Load() {
...
     if (file.IsOpen()) {
         // Reset read pointer in case this file has been read before.
-        file.Seek(0, SEEK_SET);
+        file.Seek(ncch_offset, SEEK_SET);
 
         if (file.ReadBytes(&ncch_header, sizeof(NCCH_Header)) != sizeof(NCCH_Header))
             return Loader::ResultStatus::Error;
 
         // Skip NCSD header and load first NCCH (NCSD is just a container of NCCH files)...
         if (Loader::MakeMagic('N', 'C', 'S', 'D') == ncch_header.magic) {
```
Subv: I suppose this part will be removed in a future PR

shinyquagsire23: Yeah, once there's an NCSD loader proper it will get removed.

src/core/file_sys/ncch_container.cpp
```diff
@@ -118,15 +120,15 @@ Loader::ResultStatus NCCHContainer::Load() {
...
             LOG_DEBUG(Service_FS, "Only loading the first (bootable) NCCH within the NCSD file!");
-            ncch_offset = 0x4000;
+            ncch_offset += 0x4000;
```
MerryMage: What happens if `Load()` is called more than once? (Should that even be allowed?)

src/core/file_sys/ncch_container.cpp
B3n30: Can this even happen for ncch_offset != 0?

src/core/file_sys/ncch_container.h
```diff
@@ -168,10 +168,10 @@ namespace FileSys {
  */
 class NCCHContainer {
 public:
-    NCCHContainer(const std::string& filepath);
+    NCCHContainer(const std::string& filepath, u32 ncch_offset = 0);
     NCCHContainer() {}
 
-    Loader::ResultStatus OpenFile(const std::string& filepath);
+    Loader::ResultStatus OpenFile(const std::string& filepath, u32 ncch_offset = 0);
```
B3n30: I'm not sure this scenario is even possible, but: I create a Container (with offset) but later when i OpenFile i have to call it agin with the offset or it will get defaulted to zero. So isn't it pointless to set the offset in the ctor?

src/core/file_sys/ncch_container.cpp
shinyquagsire23: Not reeaaaallly but I'm planning on moving this code out eventually anyhow.

src/core/file_sys/ncch_container.h
shinyquagsire23: Well when you're calling OpenFile you're overriding everything that the constructor is doing. 

I had to check on this but the only usage of that is in `core/loader/ncch.cpp` for the update NCCH, where it has the class instantiated without a path (w/ `NCCHContainer()`) and OpenFile provides loading paths later.

PR #3125, opened 2017-11-17
Remove vsync checkbox from ui

valentinvanelslande:
UI Change:

config
------------------------------

This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3125)


MerryMage:
While vsync doesn't work properly, there's probably some context I'm missing here.

acnleditor2:
what???

RavenHome1:
are we sure we want to remove vsync?. while limit frame rate works greatly but there is a matter of screen tearing that happens when you go to full screen mode (F11) which currently can only be solved by Vsync

baka0815:
What's the reason for removing this?
As far as I can tell from the commit it doesn't remove logic, just the setting from the GUI. So does the setting have no effect because it never changes anything?
Or is the commit just incomplete?

CaptV0rt3x:
IIRC, this was suggested by @jroweboy and @B3n30 , as this led to an increased number of people in discord with issues like " My Citra runs slow" - which is caused by enabling Vsync (on low end hardware). It was their view that the implementation of vsync was not that good. Hence a small work around - removing Qt option to enable Vsync.

B3n30:
Since the current v-sync has some flaws, i think that option shouldn't be easy accessible to users until we fix it. The option can still be set in the config file. We can add the button at a later stage when the v-sync code is fixed. 

Alternatively we could disable that option, but I prefer removing it for now.

moben:
Doesn't this leave people who turned vsync on unable to turn it off through the UI? They will continue complaining "My Citra runs slow" until someone manually digs through their config and notices vsync being on.

Maybe the name of the config option should change to make sure everyone has the same state.

baka0815:
I think so too. The code isn't removed, so the set setting (vsync=on) will still trigger the same vsync logic, but it's now "impossible" to turn it off.

I think we're talking about non technical end-users here, because everyone else should know what vsync is. So just removing the setting leaves those people who were previously complaining with no (easy) possibility to unset vsync.

bunnei:
@valentinvanelslande Thanks for submitting this change! I think that we all agree that there are problems with the vsync option that need to be addressed. That being said (as others have expressed in this thread), I'm not sure that simply removing the option from the Qt UI is the best approach. It might be better to simply remove the option altogether, and perhaps hide it behind a new setting so folks who specifically need it can turn it on.
Since there have been no updates on this in a few weeks, I'm going to close the PR - but feel free to submit a new one with the above issues addressed. Thank you!

PR #3136, opened 2017-11-19
Kernel/Memory: Added a function to change the memory state of an address range

Subv:
This function will be useful when implementing memory aliasing operations, see https://github.com/citra-emu/citra/commit/e55583d4f9f89b2e60da5fdda710e447ab7dabe9 for a draft of what's to come.

The function will enforce that the permissions and memory state of each block in the range match the input parameters before carving and updating the VMA.

Also added some VMManager tests, more to come.

This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3136)

### DIFF
--- a/src/core/hle/kernel/vm_manager.cpp
+++ b/src/core/hle/kernel/vm_manager.cpp
@@ -129,6 +129,39 @@ ResultVal<VMManager::VMAHandle> VMManager::MapMMIO(VAddr target, PAddr paddr, u3
     return MakeResult<VMAHandle>(MergeAdjacent(vma_handle));
 }
 
+ResultCode VMManager::ChangeMemoryState(VAddr target, u32 size, MemoryState expected_state,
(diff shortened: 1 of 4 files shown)


src/core/hle/kernel/vm_manager.cpp
```diff
@@ -129,6 +129,39 @@ ResultVal<VMManager::VMAHandle> VMManager::MapMMIO(VAddr target, PAddr paddr, u3
...
+    }
+
+    CASCADE_RESULT(auto vma, CarveVMARange(target, size));
+    ASSERT(vma->second.size == size);
+
+    vma->second.permissions = new_perms;
+    vma->second.meminfo_state = new_state;
+    UpdatePageTableForVMA(vma->second);
+
+    vma = std::next(MergeAdjacent(vma));
```
wwylele: dead code

B3n30:
Same OSX bullshit

PR #3138, opened 2017-11-19
Kernel/IPC: Implement MappedBuffer translation for HLE services that use the HLERequestContext architecture.

wwylele:
The last missing block of `HLERequestContext`. A `MappedBuffer` object is used to represent the memory block the kernel sets up for the service.

Note that this kind of translation is asymmetrical: for client->service request, the kernel sets up the mapped buffer; for service->client response, the kernel unmaps the buffer instead. Because `MappedBuffer`'s lifetime is managed by the context, we don't need to do anything in the response other than modifying the command buffer.

In the test, the reading buffer (with permission R) is tested in command buffer request, while writing buffer (with permission W) is tested after command buffer response. This is because the real kernel can copy the buffer data only after unmapping when handling response, and the HLE implementation can be in the same way (although it is not in this code).

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3138)

### DIFF
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -107,6 +107,12 @@ ResultCode HLERequestContext::PopulateFromIncomingCommandBuffer(const u32_le* sr
             cmd_buf[i++] = source_address;
             break;
         }
+        case IPC::DescriptorType::MappedBuffer: {
+            u32 next_id = static_cast<u32>(request_mapped_buffers.size());
+            request_mapped_buffers.emplace_back(src_process, descriptor, src_cmdbuf[i], next_id);
+            cmd_buf[i++] = next_id;
+            break;
+        }
         default:
             UNIMPLEMENTED_MSG("Unsupported handle translation: 0x%08X", descriptor);
         }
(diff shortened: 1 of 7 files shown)


src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -174,4 +185,31 @@ ResultCode HLERequestContext::WriteToOutgoingCommandBuffer(u32_le* dst_cmdbuf, P
     return RESULT_SUCCESS;
 }
 
+MappedBuffer& HLERequestContext::GetMappedBuffer(u32 id_from_cmdbuf) {
+    return request_mapped_buffers[id_from_cmdbuf];
```
Subv: maybe assert that the buffer id is within bounds

src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -174,4 +185,31 @@ ResultCode HLERequestContext::WriteToOutgoingCommandBuffer(u32_le* dst_cmdbuf, P
...
+}
+
+MappedBuffer::MappedBuffer(Process& process, u32 descriptor, VAddr address, u32 id)
```
Subv: Process can be const

src/core/hle/kernel/hle_ipc.h
```diff
@@ -63,6 +63,26 @@ class SessionRequestHandler : public std::enable_shared_from_this<SessionRequest
...
+    void Write(void* src_buffer, size_t offset, size_t size);
+
+    // interface for ipc helper
+    u32 GenerateDescriptor() const;
+    const u32 id;
+
+private:
+    friend class HLERequestContext;
+    const VAddr address;
+    Process& process;
```
Subv: this can be const

wwylele:
Update: removed `const` qualifier on `MappedBuffer` members to allow default copy assignment (used in #3101 for lambda capture)

wwylele:
Update 2:

wwylele:
Squashed commits

PR #3141, opened 2017-11-20
Appveyor: on mingw set cmake version to 3.9.6

### LINKED ISSUE #3051: All Games - Crash when opening Camera (Only on Windows, only with MinGW compiler)

B3n30:
This fixes #3051. The main cause is described here https://gitlab.kitware.com/cmake/cmake/issues/17436. Appveyor currently uses cmake 3.9.2 which has this issue. In 3.9.6 the issue is fixed. So we set mingw to use that version. This will also prevent future updates of cmake (which can be positive and negative). The other solution described in the link would be to insert `list(REMOVE_ITEM CMAKE_C_IMPLICIT_LINK_LIBRARIES gcc_eh)`.

Since i can't go in-game with my Windows VM I only tested that citra doesn't crash on verifying the token. Would be good to test if the other crashes are fixed, too.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3141)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,14 +26,15 @@ install:
   - git submodule update --init --recursive
   - ps: |
         if ($env:BUILD_TYPE -eq 'mingw') {
-          $dependencies = "mingw64/mingw-w64-x86_64-cmake",
-                          "mingw64/mingw-w64-x86_64-qt5",
+          $dependencies = "mingw64/mingw-w64-x86_64-qt5",
                           "mingw64/mingw-w64-x86_64-curl"
           # redirect err to null to prevent warnings from becoming errors
           # workaround to prevent pacman from failing due to cyclical dependencies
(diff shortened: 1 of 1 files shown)


FearlessTobi:
I can confirm the crash that occured when stopping emulation to be fixed.

CDAGaming:
Oh wait THATS fixed
Consider this thing approved

CDAGaming:
(And I'm merging this ASAP into my fork) xD

chris062689:
Has two approvals. Merging.

baka0815:
Is this somehow related to #3124?

jroweboy:
@baka0815 it shouldn't be. This is just related to a regression in cmake where it was adding an additional library to link against on mingw which broke thread local storage in citra.

PR #3145, opened 2017-11-25
shader_jit_x64_compiler: Remove ABI overhead of LG2 and EX2

MerryMage:
We reimplement log2f and exp2f in assembly.

Algorithms were inspired by those found in Mesa I've used a higher order approximation for exp2 than they did for more accuracy.

Performance numbers (for 1.2 billion sequential calls of the same instruction on my machine):

| Instruction | Old time | New time |
| ------------- | ------------- |  ------------- |
| `ex2` | 21.5s | 8.8s |
| `lg2` | 13.3s | 7.3s |

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3145)

### DIFF
--- a/src/video_core/shader/shader_jit_x64_compiler.cpp
+++ b/src/video_core/shader/shader_jit_x64_compiler.cpp
@@ -432,27 +432,13 @@ void JitShader::Compile_DPH(Instruction instr) {
 
 void JitShader::Compile_EX2(Instruction instr) {
     Compile_SwizzleSrc(instr, 1, instr.common.src1, SRC1);
-    movss(xmm0, SRC1); // ABI_PARAM1
-
-    ABI_PushRegistersAndAdjustStack(*this, PersistentCallerSavedRegs(), 0);
-    CallFarFunction(*this, exp2f);
-    ABI_PopRegistersAndAdjustStack(*this, PersistentCallerSavedRegs(), 0);
-
-    shufps(xmm0, xmm0, _MM_SHUFFLE(0, 0, 0, 0)); // ABI_RETURN
-    movaps(SRC1, xmm0);
+    call(exp2_subroutine);
     Compile_DestEnable(instr, SRC1);
 }
 
 void JitShader::Compile_LG2(Instruction instr) {
     Compile_SwizzleSrc(instr, 1, instr.common.src1, SRC1);
-    movss(xmm0, SRC1); // ABI_PARAM1
-
-    ABI_PushRegistersAndAdjustStack(*this, PersistentCallerSavedRegs(), 0);
-    CallFarFunction(*this, log2f);
-    ABI_PopRegistersAndAdjustStack(*this, PersistentCallerSavedRegs(), 0);
-
-    shufps(xmm0, xmm0, _MM_SHUFFLE(0, 0, 0, 0)); // ABI_RETURN
-    movaps(SRC1, xmm0);
+    call(log2_subroutine);
     Compile_DestEnable(instr, SRC1);
 }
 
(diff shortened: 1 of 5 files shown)


yuriks:
Nice perf wins and nice, tests!

---

Reviewed 2 of 2 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/tests/CMakeLists.txt, line 24 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp2ZnPZg5mXpWlkcvZ:-Kzp2ZnQP-e20YGIa1AN:buerzos) ([raw file](https://github.com/citra-emu/citra/blob/e2b22c721a353e8b954fa5d123ee215f9afb8a55/src/tests/CMakeLists.txt#L24)):*
> ```CMake
> target_link_libraries(tests PRIVATE ${PLATFORM_LIBRARIES} catch-single-include nihstro-headers Threads::Threads)
> if (ARCHITECTURE_x86_64)
>     target_link_libraries(tests PRIVATE xbyak)
> ```

That's weird, I would've expected this to be pulled transitively via video_core, so you shouldn't need to specify it here, since you're not using any xbyak functions directly.

Ah, looks like we're specifying xbyak as a PRIVATE dependency in common and video_core's CMakeLists. I think this is wrong, since they're used in public headers from those libraries, so they should be changed to PUBLIC.

---

*[src/tests/video_core/shader/shader_jit_x64_compiler.cpp, line 5 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp3tLV8mdB3tf5GeFW:-Kzp3tLW9aU9-TWzUFdS:bnq1e51) ([raw file](https://github.com/citra-emu/citra/blob/e2b22c721a353e8b954fa5d123ee215f9afb8a55/src/tests/video_core/shader/shader_jit_x64_compiler.cpp#L5)):*
> ```cpp
> // Refer to the license.txt file included.
> 
> #ifdef ARCHITECTURE_x86_64
> ```

Can you conditionally include the entire via on CMakeLists instead (like is done in the main codebase). Or do you have a particular reason to do it this way?

---

*[src/tests/video_core/shader/shader_jit_x64_compiler.cpp, line 49 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp5OCYSwkYdQVJw5Bf:-Kzp5OCYSwkYdQVJw5Bg:bw1hl0h) ([raw file](https://github.com/citra-emu/citra/blob/e2b22c721a353e8b954fa5d123ee215f9afb8a55/src/tests/video_core/shader/shader_jit_x64_compiler.cpp#L49)):*
> ```cpp
>     });
> 
>     const auto run = [&](float input) {
> ```

You could encapsulate this (compiling the shader, setting up N inputs and the output, and then the `run` function receiving N params and returning the output) into a standalone class to avoid duplicating this setup twice, and since it seems to be widely applicable to other shader tests.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 930 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3145#-KzplIGKEQoGqujW-X-R:-KzplIGL7lsWBTY5_sPC:b-43bd4e) ([raw file](https://github.com/citra-emu/citra/blob/d9383f7c78a5d9d3f5a634a78369a1c4a74a783e/src/video_core/shader/shader_jit_x64_compiler.cpp#L930)):*
> ```cpp
> void JitShader::CompilePrelude() {
>     CompilePrelude_Log2();
>     CompilePrelude_Exp2();
> ```

Optional: Have these two functions return the Labels to the compiled function, which are then assigned here, making it a bit more obvious that's where their results are going.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 969 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp8OGF2U5S3UnBvOE7:-Kzp8OGF2U5S3UnBvOE8:bl96ih2) ([raw file](https://github.com/citra-emu/citra/blob/d9383f7c78a5d9d3f5a634a78369a1c4a74a783e/src/video_core/shader/shader_jit_x64_compiler.cpp#L969)):*
> ```cpp
> 
>     align(16);
>     L(input_is_nan);
> ```

Any reason not to merge this block with one of the other two ret from the other two cases?

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 980 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3145#-KzphS9H9lZQf1XVCnnX:-KzphS9IHFrpJ1MHZOdO:bnih6mq) ([raw file](https://github.com/citra-emu/citra/blob/d9383f7c78a5d9d3f5a634a78369a1c4a74a783e/src/video_core/shader/shader_jit_x64_compiler.cpp#L980)):*
> ```cpp
> 
>     align(16);
>     log2_subroutine = getCurr();
> ```

You should be able to also use labels for the routine entry points and then use call with them, letting xbyak possibly emit a shorter relative call instead.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 993 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp8tBeZn9pFRr5tmXb:-Kzp8tBfTDU3g-Z5Kgq3:btsc1vp) ([raw file](https://github.com/citra-emu/citra/blob/d9383f7c78a5d9d3f5a634a78369a1c4a74a783e/src/video_core/shader/shader_jit_x64_compiler.cpp#L993)):*
> ```cpp
>     and_(eax, 0x7f800000);
>     and_(edx, 0x007fffff);
>     movss(SCRATCH, xword[rip + c0]); // Preload c0.
> ```

Curious: How did you determine placement of this load? (i.e. why here, instead of at the beginning of this block?)

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 1050 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp8f3mTJY_aIqrHOAG:-Kzp8f3mTJY_aIqrHOAH:bnpxnua) ([raw file](https://github.com/citra-emu/citra/blob/d9383f7c78a5d9d3f5a634a78369a1c4a74a783e/src/video_core/shader/shader_jit_x64_compiler.cpp#L1050)):*
> ```cpp
> 
>     align(16);
>     L(ret_label);
> ```

Isn't it better to just put ret_label before the final ret in the subroutine, instead of dedicating 16 bytes for it?

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 783 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3145#-KzqWNrmiavMQu2_D_fZ:-KzqWNrmiavMQu2_D_f_:b-ec4zgo) ([raw file](https://github.com/citra-emu/citra/blob/af6d0acd23145dc79aef002c8080263230c3ea13/src/video_core/shader/shader_jit_x64_compiler.cpp#L783)):*
> ```cpp
>     add(ABI_PARAM2, static_cast<Xbyak::uint32>(offsetof(UnitState, registers.output)));
>     CallFarFunction(*this, Emit);
>     if (Common::GetCPUCaps().avx) {
> ```

After some research, it seems that the latest versions of gcc, clang and MSVC all generate vzeroupper before calls or rets in functions in which AVX was used. Tests: https://godbolt.org/g/K7Eeun (Note that at the time of writing the MSVC in godbolt is out of date and has a bug.)

Given that, do we still need to emit this? And in any case, would it make more sense to gate this behind __AVX__ (the compiler define) instead of the runtime cap bit? Presumably if __AVX__ isn't set then we don't need to worry about AVX code in the rest of the application.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3145#-:-KzqXJM5Kglu3Riip7nu:bmvymp6)*

MerryMage:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/tests/video_core/shader/shader_jit_x64_compiler.cpp, line 5 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp3tLV8mdB3tf5GeFW:-L-1aztlhW6XbYij786F:b-gpfqt) ([raw file](https://github.com/citra-emu/citra/blob/e2b22c721a353e8b954fa5d123ee215f9afb8a55/src/tests/video_core/shader/shader_jit_x64_compiler.cpp#L5)):*

Can you conditionally include the entire via on CMakeLists instead (like is done in the main codebase). Or do you have a particular reason to do it this way?
</blockquote></details>

No reason, will do it in CMakeLists.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 969 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp8OGF2U5S3UnBvOE7:-L-1cg8AmWF0o4OVzRpM:bhy9ot6) ([raw file](https://github.com/citra-emu/citra/blob/d9383f7c78a5d9d3f5a634a78369a1c4a74a783e/src/video_core/shader/shader_jit_x64_compiler.cpp#L969)):*

Any reason not to merge this block with one of the other two ret from the other two cases?
</blockquote></details>

Will do.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 993 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp8tBeZn9pFRr5tmXb:-L-1ci7E_46eSt6mY2HJ:bx6h70g) ([raw file](https://github.com/citra-emu/citra/blob/d9383f7c78a5d9d3f5a634a78369a1c4a74a783e/src/video_core/shader/shader_jit_x64_compiler.cpp#L993)):*

Curious: How did you determine placement of this load? (i.e. why here, instead of at the beginning of this block?)
</blockquote></details>

Decision of ordering of instructions was based on execution port contention, done on paper. The difference in practice is minor - a couple of cycles - and is unlikely to be noticeable, but I like doing it anyway.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 1050 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3145#-Kzp8f3mTJY_aIqrHOAG:-L-1dOilIWxx45RW8wt-:bxdjc3q) ([raw file](https://github.com/citra-emu/citra/blob/d9383f7c78a5d9d3f5a634a78369a1c4a74a783e/src/video_core/shader/shader_jit_x64_compiler.cpp#L1050)):*

Isn't it better to just put ret_label before the final ret in the subroutine, instead of dedicating 16 bytes for it?
</blockquote></details>

Smaller jump instruction in the function itself.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 783 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3145#-KzqWNrmiavMQu2_D_fZ:-L-1cbgAj_gAK0hDWjMo:bub8dtj) ([raw file](https://github.com/citra-emu/citra/blob/af6d0acd23145dc79aef002c8080263230c3ea13/src/video_core/shader/shader_jit_x64_compiler.cpp#L783)):*

After some research, it seems that the latest versions of gcc, clang and MSVC all generate vzeroupper before calls or rets in functions in which AVX was used. Tests: https://godbolt.org/g/K7Eeun (Note that at the time of writing the MSVC in godbolt is out of date and has a bug.)

Given that, do we still need to emit this? And in any case, would it make more sense to gate this behind __AVX__ (the compiler define) instead of the runtime cap bit? Presumably if __AVX__ isn't set then we don't need to worry about AVX code in the rest of the application.
</blockquote></details>

Paranoia on my part, essentially. I can drop that commit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3145)*

MerryMage:
Added a `ShaderTest` class, leaving different future requirements for `Run` inputs and outputs for any future tests to figure out.

Thanks for the review @yuriks.

yuriks:
:lgtm:

---

Reviewed 2 of 5 files at r4, 2 of 2 files at r5, 1 of 1 files at r6.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 1018 at r6](https://reviewable.io:443/reviews/citra-emu/citra/3145#-L-AiTOzfPZGHIzDDLGS:-L-AiTOzfPZGHIzDDLGT:b-7g2cyc) ([raw file](https://github.com/citra-emu/citra/blob/9927c195d10fde6f6d06fea17bdbf1c687a4d6ed/src/video_core/shader/shader_jit_x64_compiler.cpp#L1018)):*
> ```cpp
>     xorps(SRC1, SRC1); // break dependency chain
>     movss(SRC1, SCRATCH2);
>     L(input_is_nan);
> ```

Just to confirm: Including the shufps in this case (after moving the label here) is intentional, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3145#-:-L-AqLuB4EBbSdnRLohl:bnfp4nl)*

MerryMage:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 1018 at r6](https://reviewable.io:443/reviews/citra-emu/citra/3145#-L-AiTOzfPZGHIzDDLGS:-L-D8Gp1_S0wAkCJBs4G:b1hz18i) ([raw file](https://github.com/citra-emu/citra/blob/9927c195d10fde6f6d06fea17bdbf1c687a4d6ed/src/video_core/shader/shader_jit_x64_compiler.cpp#L1018)):*

Just to confirm: Including the shufps in this case (after moving the label here) is intentional, right?
</blockquote></details>

Indeed, mistake the first time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3145)*

MerryMage:
Rebased.

src/video_core/shader/shader_jit_x64_compiler.cpp
```diff
@@ -935,7 +921,179 @@ void JitShader::Compile(const std::array<u32, MAX_PROGRAM_CODE_LENGTH>* program_
...
+    mulss(SCRATCH, SRC1);
+    addss(SCRATCH, xword[rip + c3]);
+    mulss(SCRATCH, SRC1);
+    subss(SRC1, ONE);
+    addss(SCRATCH, xword[rip + c4]);
+    mulss(SCRATCH, SRC1);
+    addss(SCRATCH2, SCRATCH);
+
+    // Duplicate result across vector
+    xorps(SRC1, SRC1); // break dependency chain
```
hrydgard: Curious, does this really help? The mov to it in the next instruction should have the same effect..

MerryMage: My logic was that shufps depends on the entire xmm register, while the movss register only affects the lowest 32 bits. I could be incorrect.

hrydgard: Oh, doh. I forgot about that.. Yeah, this is probably right.

Issue #3148, opened 2017-11-25
Flip 3DS implementation [enhancement]


adityaruplaha:
Planned implementation of preset tilt 90, 180, 270 degrees. This will allow flipping of the 3DS. Considering tilt code is already present, thanks to @wwylele , it shouldn't be too hard. Planned code:
<code>
// Stuff done by wwylele's code here.
if (preset_tilt != 0) // checks the settings for the value
{
tilt = preset_tilt; // Changes whatever values wwlele's code calculated to the preset value
}
else
{
// does nothing to anything. 
}
</code>

Requires a modification to settings. UI implementation currently not planned.

adityaruplaha:
Can anyone point out @wwylele 's code?

wwylele:
The current motion control implementation is here: https://github.com/citra-emu/citra/blob/master/src/input_common/analog_from_button.cpp

adityaruplaha:
Thanks ! Sorry for editing your code though.

adityaruplaha:
Huh? That's not what I want exactly. I have got a `std::tuple<float, float>`, but I want the tilt as an integer, to set it to a value. Where is that computed?

adityaruplaha:
OK, I found it.

Dragios:
Partially implemented with PR #3366 & #3590

PR #3154, opened 2017-11-26
3DS Preset Tilt for flipping console

adityaruplaha:
This will allow setting tilt to any preset value including 180 degrees, which will flip the 3DS. It can also be set to a garbage value if the user doesn't want to use tilt. ~I am still working on closing the loopholes, like a value beyond 360 degrees.~

Relevant issue: #3148 
Relevant community questions:

1. [Pokemon help turning 3ds upside down](https://community.citra-emu.org/t/pokemon-help-turning-3ds-upside-down/5786)
2.  [How to Evolve Inkay in Pokemon?](https://community.citra-emu.org/t/how-to-evolve-inkay-in-pokemon/2768/4)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3154)


src/core/settings.cpp
```diff
@@ -31,6 +31,8 @@ void Apply() {
 
     AudioCore::SelectSink(values.sink_id);
     AudioCore::EnableStretching(values.enable_audio_stretching);
+    
+    MotionEmu::preset_tilt = values.preset_tilt % 360;
```
Subv: Is the modulo (%) operator defined for floating point types?

RavenHome1:
if the only concern is a 180 flip then why not just bind a certain key in which the user holds and the emulator will feed the game that the console is flipped upside down.

src/core/settings.cpp
```diff
@@ -8,7 +8,7 @@
 #include "core/hle/service/ir/ir.h"
 #include "core/settings.h"
 #include "video_core/video_core.h"
-
+#include "input_common/motion_emu.h"
```
wwylele: You are introducing a circular dependency here. `input_common` already depends on `core`

src/input_common/motion_emu.h
```diff
@@ -18,6 +18,8 @@ class MotionEmu : public Input::Factory<Input::MotionDevice> {
      *     - "update_period": update period in milliseconds
      *     - "sensitivity": the coefficient converting mouse movement to tilting angle
      */
+    float preset_tilt;
```
wwylele: Don't break up between the comment and the function it belongs to

src/input_common/motion_emu.cpp
```diff
@@ -37,7 +37,10 @@ class MotionEmuDevice {
 
     void Tilt(int x, int y) {
         auto mouse_move = Math::MakeVec(x, y) - mouse_origin;
-        if (is_tilting) {
+        if (preset_tilt != 0) {
+                    tilt_angle = preset_tilt; // Changes whatever values wwylele's code calculated to the preset value
+                }
+        elseif (is_tilting) {
```
wwylele: `else if`

src/input_common/motion_emu.cpp
```diff
@@ -37,7 +37,10 @@ class MotionEmuDevice {
 
     void Tilt(int x, int y) {
         auto mouse_move = Math::MakeVec(x, y) - mouse_origin;
-        if (is_tilting) {
+        if (preset_tilt != 0) {
+                    tilt_angle = preset_tilt; // Changes whatever values wwylele's code calculated to the preset value
```
wwylele: This leaves `tilt_direction` undefined, so the you can get any tilt direction.

src/core/settings.cpp
adityaruplaha: Ok, I will address that

src/input_common/motion_emu.h
adityaruplaha: Addressed

src/core/settings.cpp
adityaruplaha: Addressed. ( I'm a absolute fool)

src/core/settings.cpp
adityaruplaha: OK, but I'm modifying values stored in that file, so I need to include it right?

src/input_common/motion_emu.cpp
adityaruplaha: setting this to `Math::MakeVec(0.0f ,1.0f ,0.0f)`.

src/input_common/motion_emu.cpp
adityaruplaha: Addressed. ( I'm an absolute fool)

src/input_common/motion_emu.cpp
adityaruplaha: Addressed. Though I'm not sure if that vector is right.

adityaruplaha:
Briefing up:
Addressed all requests except:
- Circular dependency ( @wwylele ) >> I don't know how to circumvent it. 😢 
- `tilt_direction` undefined ( @wwylele ) >> Approval & background research required.

Please add `pr:needs-work`

adityaruplaha:
Now, @RavenHome1 
I wanted this pr to be as flexible & as performance-friendly as possible.
My concern is not a 180 flip, but allowing presets, so that values beyond 90 can be accessed, atleast partially.

Actually, I thought of adding a hotkey binding, but then decided against it. I think you have noticed, that if anyone wants disable (say, he/she has a game that doesn't use it), only has to garbage that value. I'm sure you know all that computation is expensive, so disabling tilt might give +3 or +4 fps. Win-win! The emulator doesn't need to do all that stuff that just goes to waste, & the user gets better speed.

adityaruplaha:
Ok, fixed the error.
Now, please help me with the vectors.
I don't think the circular dependency is causing a problem right now. (Or is it? 🤔 )

jroweboy:
First off, thanks for the contribution! I hope what I explain here will help you understand what wwylele are trying to explain :)

The 3ds has two different measurements that could apply to "tilt," accelerometer and gyroscope. the accelerometer measures the velocity that the 3ds is moving in a direction in 3d space (up/down, left/right, front/back) and the gyroscope measures the 3d rotation that the console is doing (roll, yaw, pitch) Simply saying that there is a single number "rotation" that you want to add to the code doesn't really explain what you are trying to flip here. Is it sending data to the accelerometer or the gyroscope? if its the gyroscope, is it roll, yaw, or pitch? 

What the function you are changing does is convert a 2d mapping of the point from where you started right clicking and dragging the mouse to the current point and calculates a direction and angle from it. There is a background thread that reads this information and does some math to convert this into the accelerometer and gryoscope values that I was talking about before. Elsewhere in the code, the game will read from these values. By editing this value, you are inadvertently affecting the reading from both the accelerometer and the gryoscope, which may or may not be intended.

This is what wwylele is pointing out when he said "remember that rotation is in 3d space" The part of the code you are touching is for converting the 2d mouse input into values that can be calculated into a 3d acceleration and rotation, so its probably not where you want your changes to go :) I'm not the right person to ask where it should go, as I only have a cursory understanding of the code.

jroweboy:
Also, yes, a circular dependency is a very bad thing (TM) Its something that will prevent the PR from getting merged completely. (But its also something that shouldn't be too hard to fix) It means that InputCore has includes from Core (and thats fine), but your PR does it the other way around, where Core now has an include from InputCore (now they depend on each other!)

lioncash:
We also likely shouldn't arbitrarily expose members of a class as public. It should be done by adding an API to the class to do that.

adityaruplaha:
Many many thanks @jroweboy ! I happen to know a bit of vector math ( thanks to opengl tutorials), so i'll clean it up asap. I wanna set roll (rotation along y axis according to @wwylele 's code) to -180 degrees, can you point out the exact snippet? Also, I have found a workaround for the circ dep, which i'll explain here:

> I'll add a function to the settings.cpp, which will return the value, & I'll call it from the `MotionEmu` thread. Simple & quick.

adityaruplaha:
Also, Its likely that I'll open a pr 3 months later (coz ill be busy with school exams) that'll allow setting all gyro values to any value from the ini, will that be encouraged?

adityaruplaha:
Omg! @wwylele you sure have a super firm math base. Very nice work on the quaternions. It will take me some time to figure all this stuff out, so it's currently 'on hold'.

adityaruplaha:
That should do it I guess.
I just put the code in the wrong place.
Note, `tilt_direction` remains defined as usual.

Any suggestions are welcome.

src/input_common/motion_emu.cpp
```diff
@@ -91,7 +91,13 @@ class MotionEmuDevice {
         while (!shutdown_event.WaitUntil(update_time)) {
             update_time += update_duration;
             old_q = q;
-
+
```
CDAGaming: Whitespace...

wwylele:
Revisiting this PR, I come up with some questions:

First of all, what is the correct angle of "turning 3ds upside down" ? I played Pokemon, but haven't try to evolve Inkay (lets just call this event E). As far as I heard from others, the "upside down" should be the angle such that, the screen is vertical (perpendicular to the ground), and the "up" direction in the screen points downward to the ground. Assuming that the 3DS is fully opened - that is, the two screens share the same "up" direction, this "upside down" direction, described as the gravitiy vector in 3DS space (in which the accelerometer measures), is `v1 = (x = 0, y = 0, z = 1)` (See [here](https://github.com/citra-emu/citra/blob/master/src/core/frontend/input.h#L111) about the 3DS space axis definition). The default orientation of 3DS in citra is that the screens are parallel to the ground, which means the gravity vector of accelerometer is `v0 = (x = 0, y = -1, z = 0)`. Notice how `v0` and `v1` are actually perpendicular. This means that a 90-degree tilt should be able to made it to `v1`. In fact, in the current implementation, this can be achieved by moving the mouse upward to the maximum. Would this actually trigger event E?

So this lead me to wonder whether you have actually tested your code to find the required angle to trigger event E, and whether you understand what 180-degree rotation means. Also, in the case that you do understand it, would user also understand it? People might also just put the number "180" in the config, which is actually wrong if my theory above is true. How do you tell user what the angle actually means?

adityaruplaha:
@wwylele you are right. i didnt even bother to research this stuff first before trying to implement this.
I just assumed that i y = 180 was the correct angle. hence proved, im an as****e.
But those vectors are exactly what i need!
I can't test this as I dont have the game.

adityaruplaha:
Wait a sec, do you mean that if i can modify the mouse tilt controls, then i can make the top of the screen correspond to 180 degrees? If so, please tell me.

wwylele:
>  then i can make the top of the screen correspond to 180 degrees?

Technically yes. But isn't it even less intuitive to users?

adityaruplaha:
Nah, we can do that. i was a newbie to all this just a year ago, i picked it up quite well. Not to mention my first dump was diastrous tho.

wwylele:
@adityaruplaha

You probably misunderstood something. I didn't say it is hard for us to implement. I said it is hard for user to use.

The reason I capped the angle to 90 instead of 180 is: it is pretty hard for a user to properly perform a "1/2 of screen mouse movement". When a user wants to tile, the first thing they would try is to move the mouse to the maximum. See #1951 for the original PR, and the report I received there about 180 degree being confusing.

adityaruplaha:
I understand this, but i believe even if a majority of users want it to be capped at 90, if a full (180 degree) implementation is easy, then it should at least be provided as an option. I think at least a few users will agree with me. Do you?
If so, please point out the snippet, so I can implement this in a new PR.

wwylele:
@adityaruplaha 

The angle is capped [here](https://github.com/citra-emu/citra/blob/9b647d459bbb2ac45f821273e8d3a85ab12b60e6/src/input_common/motion_emu.cpp#L46-L47). Changing `MathUtil::PI * 0.5f` to `MathUtil::PI` will give you a maximum of 180 degree. But before you start off coding, please do some tests to check what is the required angle for pokemon, and decide if a 180 degree maximum is actually necessary.

I have some advice for coding:
 - no need to explicitly mention my name in the comment. That makes me a little bit embarrassed :sweat_smile: . Whoever want to know the code author can just check `git blame`.
 - format your code style. You can read the style suggestion by clang-format in one of our CIs (or just run you code through clang-format-3.8 locally)
 - respect to the existing code structure and dependency hierarchy. Limit the code change in their scope. For example, for this input feature, you want little to zero change to `core`, and probably most changes in `input_common`. 
 - makes it intuitive to use. I am not particular good at user experience stuff. Perhaps you can ask @jroweboy for more info.

Looking forward to your future contribution :D Though, remember that citra is not just for you, but for everyone ;)

adityaruplaha:
- Ok , i wont mention you.
- How do i get clang format?
- yes, i made a gigantic blunder in this pr
- i'm just focusing on implementing this now, i'll worry about ui integration (whin im no good at) later.

adityaruplaha:
My new PR #3366.

PR #3157, opened 2017-11-26
citra-qt: Print stack trace

MerryMage:
Third iteration of this concept.

* Windows-only for now
* Requires a pdb file to be present for symbols to work.
* ~~**TODO:** Actually dump minidumps, which users can then upload.~~
* **TODO:** Generate pdbs for mingw using cv2pdb or similar

screen shot 2017-11-26 at 18 42 10

Original PRs: #1874, #2355

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3157)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,7 +253,7 @@ IF (APPLE)
 ELSEIF (WIN32)
     # WSAPoll and SHGetKnownFolderPath (AppData/Roaming) didn't exist before WinNT 6.x (Vista)
     add_definitions(-D_WIN32_WINNT=0x0600 -DWINVER=0x0600)
-    set(PLATFORM_LIBRARIES dbghelp winmm ws2_32)
+    set(PLATFORM_LIBRARIES dbghelp ole32 shell32 winmm ws2_32)
     IF (MINGW)
         # PSAPI is the Process Status API
         set(PLATFORM_LIBRARIES ${PLATFORM_LIBRARIES} psapi imm32 version)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,7 +253,7 @@ IF (APPLE)
 ELSEIF (WIN32)
     # WSAPoll and SHGetKnownFolderPath (AppData/Roaming) didn't exist before WinNT 6.x (Vista)
     add_definitions(-D_WIN32_WINNT=0x0600 -DWINVER=0x0600)
-    set(PLATFORM_LIBRARIES winmm ws2_32)
+    set(PLATFORM_LIBRARIES dbghelp winmm ws2_32)
     IF (MINGW)
         # PSAPI is the Process Status API
         set(PLATFORM_LIBRARIES ${PLATFORM_LIBRARIES} psapi imm32 version)
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -1,6 +1,9 @@
 #include <QApplication>
+#include <QDateTime>
+#include <QDir>
 #include <QHBoxLayout>
 #include <QKeyEvent>
+#include <QStandardPaths>
 #include <QScreen>
 #include <QWindow>
 
@@ -17,7 +20,14 @@
 #include "input_common/motion_emu.h"
 #include "network/network.h"
 
+static std::string GenerateMinidumpFilename() {
+    QString dir = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
+    QString file = QDateTime::currentDateTime().toString("'minidump-citra-'yyyyMMdd-HHmmss'.dmp'");
+    return QDir::toNativeSeparators(dir + "/" + file).toStdString();
+}
+
 EmuThread::EmuThread(GRenderWindow* render_window) : render_window(render_window) {
+    // Custom types must be registered to be used with Qt signals/slots.
     qRegisterMetaType<Common::CrashInformation>("Common::CrashInformation");
 }
 
@@ -64,7 +74,8 @@ void EmuThread::run() {
                 }
             }
         },
-        [&](const Common::CrashInformation& crash_info) { emit Crashed(crash_info); });
+        [&](const Common::CrashInformation& crash_info) { emit Crashed(crash_info); },
+        GenerateMinidumpFilename());
 
     // Shutdown the core emulation
     Core::System::GetInstance().Shutdown();
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -3,12 +3,12 @@
 #include <QDir>
 #include <QHBoxLayout>
 #include <QKeyEvent>
-#include <QStandardPaths>
 #include <QScreen>
 #include <QWindow>
 
 #include "citra_qt/bootmanager.h"
 #include "common/crash_handler.h"
+#include "common/file_util.h"
 #include "common/microprofile.h"
 #include "common/scm_rev.h"
 #include "common/string_util.h"
@@ -21,7 +21,7 @@
 #include "network/network.h"
 
 static std::string GenerateMinidumpFilename() {
-    QString dir = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
+    QString dir = QString::fromStdString(FileUtil::GetUserPath(D_LOGS_IDX));
     QString file = QDateTime::currentDateTime().toString("'minidump-citra-'yyyyMMdd-HHmmss'.dmp'");
     return QDir::toNativeSeparators(dir + "/" + file).toStdString();
 }
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -5,6 +5,7 @@
 #include <QWindow>
 
 #include "citra_qt/bootmanager.h"
+#include "common/crash_handler.h"
 #include "common/microprofile.h"
 #include "common/scm_rev.h"
 #include "common/string_util.h"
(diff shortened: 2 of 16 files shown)


chris062689:
Not sure if it's possible, but can the stackdump include additionally information such as CPU, GPU, OpenGL version. Etc? That may be useful for support

src/citra_qt/main.cpp
```diff
@@ -836,6 +841,12 @@ void GMainWindow::UpdateStatusBar() {
     emu_frametime_label->setVisible(true);
 }
 
+void GMainWindow::OnCrashed(const Common::CrashInformation& crash_info) {
+    CrashDialog crashDialog(this, crash_info);
+    crashDialog.exec();
+    QCoreApplication::exit(EXIT_FAILURE);
```
lioncash: `<cstdlib>` needs to be included for `EXIT_FAILURE`

src/citra_qt/crash_dialog/crash_dialog.cpp
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/crash_handler.h"
+#include "common/scm_rev.h"
+#include "crash_dialog.h"
+#include "ui_crash_dialog.h"
+
+CrashDialog::CrashDialog(QWidget* parent, const Common::CrashInformation& crash_info)
+    : QDialog(parent), ui(new Ui::CrashDialog) {
```
lioncash: `make_unique`

src/common/crash_handler.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+#ifdef _WIN32
+
+#pragma comment(lib, "dbghelp.lib")
+
+#include <windows.h>
+// windows.h must be included first.
+#include <dbghelp.h>
+
+#include <csetjmp>
```
lioncash: `<cstdio>` needs to be included for `fprintf`

src/common/crash_handler.cpp
```diff
@@ -0,0 +1,173 @@
...
+static LONG WINAPI UnhandledExceptionFilter(_EXCEPTION_POINTERS* ep) {
+    // Prevent re-entry.
+    if (unhandled_exception_called)
+        return EXCEPTION_CONTINUE_SEARCH;
+
+    unhandled_exception_called = true;
+    GetStackTrace(*ep->ContextRecord);
+
+    // Ensure we have a log of everything in the console first.
+    fprintf(stderr, "Unhandled Exception:\n");
```
lioncash: This should probably have the `std::` qualifier as well to be consistent with `std::longjmp`

src/citra_qt/main.cpp
MerryMage: done.

src/common/crash_handler.cpp
MerryMage: done.

MerryMage: Forgot `<cstdlib>` here.

MerryMage:
Implemented dumping of minidumps. There's a lot of flexibility in what one can save in a minidump. With the settings I've picked they'd probably be less than 1 MiB in size.

@chris062689 Will do. I'll just do it whatever way telemetry is doing it.

src/common/crash_handler.cpp
MerryMage: done.

MerryMage:
Added CPU/GPU information.

Will generate mingw PDBs in another PR.

src/citra_qt/bootmanager.cpp
```diff
@@ -20,8 +24,17 @@
 #include "input_common/motion_emu.h"
 #include "network/network.h"
 
+static std::string GenerateMinidumpFilename() {
+    QString dir = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
```
jroweboy: When the logging backend changes are merged, this should likely use the `FileUtil::GetUserPath(D_LOGS_IDX)` instead.

src/common/crash_handler.cpp
```diff
@@ -0,0 +1,266 @@
...
+    if (unhandled_exception_called)
+        return EXCEPTION_CONTINUE_SEARCH;
+    unhandled_exception_called = true;
+
+    if (ctx) {
+        // A copy is required as getting the stack trace modifies the context.
+        CopyContext(ctx, CONTEXT_ALL, ep->ContextRecord);
+        GetStackTrace(*ctx);
+
+        // Ensure we have a log of everything in the console.
```
jroweboy: Whats the point of writing this to the console? If we want to write this to the log file (after logging backends is merged) is it possible to use the logging macros at this point?

src/common/crash_handler.cpp
```diff
@@ -0,0 +1,266 @@
...
+                                        undecorated_name, symbol_displacement, file_name.c_str(),
+                                        line.LineNumber);
+    };
+
+    // NOTE: SymFunctionTableAccess64 doesn't work with the non-standard stack frames our JIT
+    // produces. Thus we elect to not use StackWalk64, but instead manually use the Rtl* functions.
+
+    // Initialise symbols
+    if (SymInitialize(process, nullptr, TRUE) == FALSE) {
+        std::fprintf(stderr, "Failed to get symbols. Continuing anyway...\n");
```
jroweboy: ditto about using the logging macros

src/citra_qt/crash_dialog/crash_dialog.cpp
```diff
@@ -0,0 +1,59 @@
...
+    if (crash_info.minidump_filename) {
+        minidump_filename = QString::fromStdString(*crash_info.minidump_filename);
+        ui->view_minidump_button->setEnabled(QFile::exists(minidump_filename));
+    }
+}
+
+CrashDialog::~CrashDialog() = default;
+
+void CrashDialog::on_view_minidump_button_released() {
+    Common::ShowInFileBrowser(minidump_filename.toStdString());
```
jroweboy: If this is only ever called from the QT crash dialog, we should just use `QDesktopServices` instead of in housing our own method like we did for `Open Save Data Location`. If you think we should have this in common, then maybe we should consider moving the other usages to use the one we wrote.

src/citra_qt/crash_dialog/crash_dialog.cpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+CrashDialog::CrashDialog(QWidget* parent, const Common::CrashInformation& crash_info)
+    : QDialog(parent), ui(std::make_unique<Ui::CrashDialog>()) {
+    ui->setupUi(this);
+
+    ui->informational_box->clear();
+    AddLine("Citra Crash Information");
+    AddLine("===========================");
+    AddLine(fmt::format("Build information: {} {}", Common::g_build_date, Common::g_build_name));
```
jroweboy: This should likely just use QString formatting, but honestly, i prefer fmtlib formatting so I don't mind if you keep it this way ;D

CMakeLists.txt
```diff
@@ -250,7 +250,7 @@ IF (APPLE)
 ELSEIF (WIN32)
     # WSAPoll and SHGetKnownFolderPath (AppData/Roaming) didn't exist before WinNT 6.x (Vista)
     add_definitions(-D_WIN32_WINNT=0x0600 -DWINVER=0x0600)
-    set(PLATFORM_LIBRARIES winmm ws2_32)
+    set(PLATFORM_LIBRARIES dbghelp ole32 shell32 winmm ws2_32)
```
jroweboy: Was shell32 only used for running the system call to open the folder in explorer?

src/common/crash_handler.cpp
MerryMage: The use of `fprintf` is intentional, since we cannot really assume anything about application state at this point. This is a last ditch effort to get something useful somewhere, before we attempt to display UI (which may not actually work since the program may not be in a well defined state).

src/citra_qt/crash_dialog/crash_dialog.cpp
MerryMage: The reason I didn't use QDesktopServices was because I couldn't find a way to open the containing folder *and* select the file. tbh this is probably not such a big deal and I could switch to using those if you don't mind dropping the file-select.

CMakeLists.txt
MerryMage: Yes.

src/citra_qt/bootmanager.cpp
MerryMage: Noted.

src/citra_qt/crash_dialog/crash_dialog.cpp
jroweboy: i'm fine with either. i personally lean towards qdesktopservices being cleaner, but i could see an argument for having this in common

MerryMage:
Rebased onto current master.

jroweboy:
I went and played around with the appveyor builds and I couldn't ever get this to work. For the mingw builds, I just got the typical vs popup with "the program has crashed. would you like to debug or close?" and with the msvc builds... i couldn't get the build to crash lol.

So I'll try putting it in canary and see if it has something to do with the fact that I have all these build tools set up before I merge this.

wwylele:
Removing canary-merge tag due to conflict

MerryMage:
@jroweboy Here are builds that assert when you start a game: https://ci.appveyor.com/project/MerryMage/citra/build/1.0.298

jroweboy:
GetGLString only returns information from the current GL context. If Citra is crashed at this point, I don't think we have the GL context anymore, correct? When I tested your force crash build, this information was always null. Is it fine to quickly make a 3.3 core context here to get this information?

Also, i wasn't able to click the minidump button because the logs directory doesn't exist until the logging backend changes are merged (meaning it doesn't generate a minidump). I don't think this is a blocker of course :)

jroweboy:
I've taken a bit to clean this up and rebase it, but I'm not sure I have the time or energy to finish this myself. I would like to have a framework for crash watching, so please let me know what you want done with this, and feel free to reopen it if we want to continue. Otherwise, please open an issue about crash reporting, and link this PR in there so it doesn't get lost. In the meantime, I need to close this PR so we don't feel so overwhelmed about everything in the back log. I hope you can understand, and thanks again so much for this contribution!

Issue #3040, opened 2017-10-23
Duplicated menu bar on Unity graphical shell (Ubuntu)


wwylele:
![double menu](https://user-images.githubusercontent.com/4592895/31908617-2f34ee58-b840-11e7-990a-efe1c242b556.png)

As designed, [Unity](https://en.wikipedia.org/wiki/Unity_(Ubuntu)) put all menu bars from any application **to the top bar** (instead of in the window). This worked fine for citra, until recently. Now there is another menu bar shown in the window, causing two menu bars shown. I didn't bisect exactly which commit caused this yet, but I am sure it is within a month, and as the fullscreen PR is the only recent UI change I am aware of, I guess it is related.

wwylele:
Note: the duplicated menu bar won't show on citra startup. It will shows once a you start a game

PR #3159, opened 2017-11-27
citra-qt : Fix a bug in our fullscreen implementation

FearlessTobi:
Fixes the bug described [here](https://github.com/citra-emu/citra/pull/3001#issuecomment-338401048). (Could also fix https://github.com/citra-emu/citra/issues/3040 but I don´t have a Ubuntu VM set up to test it.)
Therefore, the contents of ToggleFullscreen() have been refactored into ShowFullscreen() and HideFullscreen() (feel free to suggest better names for them) to make it easier to fix the bug. 
Also tracking of the window geometry before it was fullscreened has been added.

Thanks @CDAGaming for helping me with clang format.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3159)

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -583,7 +583,9 @@ void GMainWindow::BootGame(const QString& filename) {
     render_window->setFocus();
 
     emulation_running = true;
-    ToggleFullscreen();
+    if (ui.action_Fullscreen->isChecked()) {
+        ShowFullscreen();
+    }
     OnStartGame();
 }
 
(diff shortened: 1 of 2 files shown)


wwylele:
Yes, this fixes #3040

src/citra_qt/main.cpp
```diff
@@ -745,21 +747,33 @@ void GMainWindow::ToggleFullscreen() {
...
-        } else {
-            render_window->showNormal();
-        }
+        HideFullscreen();
+    }
+}
+
+void GMainWindow::ShowFullscreen() {
+    if (ui.action_Single_Window_Mode->isChecked()) {
+        UISettings::values.geometry = saveGeometry();
```
jroweboy: I didn't think fullscreen mode affects the window geometry. Is it really required to save the geometry?

Instead shouldn't you be storing the window state and then just if (state == normal) showNormal() else if state == maximized showMaximized() or something like that

FearlessTobi: I actually tested this when I tried to figure out a solution. But I had two concerns, so I went with saving the Geometry:

* When I tested just saving the state, when escaping fullscreen everthing showed up, except the Windows Program bar and the Taskbar. Citra essentially was still in fullscreen.
* I couldn´t find a way to properly store the state of the render window when not in single window mode.

And another minor advantage of saving the geometry is, that not just state of the window but also it´s screen position before it was fullscreen is stored and will get restored.
Also i can´t really find any cases where would be better to just store the state. Imo the code even looks cleaner without another if-statement or switch.

FearlessTobi:
Rebased and squashed commits.

PR #3168, opened 2017-11-30
[External] Travis CI Quick Fix + Minor Submodule Updates

CDAGaming:
## Info
This PR fixes several Travis-Related Warnings/Errors/Deprecation's in MacOS (Now running on Sierra instead of El Capitan, as 7.x Versions of XCode have now been marked obsolete by Travis CI). These Warnings included Path Not Set Warnings, shell_session warnings, and a few other types.

This Commit also Updates the following Submodules:
- enet (Includes Minor CMakeList Changes for cleaner Solutions) ([Changelog](https://github.com/lsalzman/enet/compare/a84c120eff13d2fa3eadb41ef7afe0f7819f4d6c...39a72ab1990014eb399cee9d538fd529df99c6a0))
- Nihstro (Includes a bugfix on compiling with Boost v1.63) ([Changelog](https://github.com/neobrain/nihstro/compare/7e24743af21a7c2e3cef21ef174ae4269d0cfdac...28ac07a5273b6250cfafb4630bee8da820ad33d7))

## Tests Info
Both the MINGW and MSVC Builds have been tested for compilation success, and in-app performance to ensure nothing isn;t broken by this small PR.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3168)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
     - os: osx
       env: NAME="macos build"
       sudo: false
-      osx_image: xcode7.3
+      osx_image: xcode9.2
       install: "./.travis/macos/deps.sh"
       script: "./.travis/macos/build.sh"
       after_success: "./.travis/macos/upload.sh"


.travis/linux/deps.sh
```diff
@@ -1,3 +1,10 @@
 #!/bin/sh -ex
 
 docker pull ubuntu:16.04
+sudo apt-get remove -y postgresql-9.1
+sudo apt-get remove -y postgresql-9.2
+sudo apt-get remove -y postgresql-9.3
+sudo apt-get remove -y postgresql-9.4
+sudo apt-get remove -y postgresql-9.5
```
MerryMage: Why? We do not use Postgres.

.travis/linux/deps.sh
```diff
@@ -1,3 +1,10 @@
...
+sudo apt-get update -q
+sudo apt-get upgrade -q -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"
```
MerryMage: These two lines are unnecessary. Host packages don't matter to us. This just unnecessarily lengthens build times.

.travis/macos/deps.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh -ex
 
+rvm get head --auto-dotfiles
```
MerryMage: We do not use ruby.

CDAGaming: Citra Itself doesn't, though Travis Does. Without this line, the following occurs on each Travis Build in this Repo:
`$ rvm use
Warning! PATH is not properly set up, '/Users/travis/.rvm/gems/ruby-2.0.0-p648/bin' is not at first place,
         usually this is caused by shell initialization files - check them for 'PATH=...' entries,
         it might also help to re-add RVM to your dotfiles: 'rvm get stable --auto-dotfiles',
         to fix temporarily in this shell session run: 'rvm use ruby-2.0.0-p648'.`

.travis/linux/deps.sh
CDAGaming: When running the added sudo apt-get upgrade command, Travis would fail as previous versions of postgresql installed already on it were not ending their instances. These lines Force-Remove any previous versions of this so that they can install no problem.

.travis/linux/deps.sh
CDAGaming: As for the apt-get update and upgrade commands, these do not really extend build time (Even though it takes an extra 3 minutes, there is seemingly a delay on when the MacOS Build starts. From this, the build times end up almost being the exact same, either longer or shorter by a few seconds). The commands were added as some libraries relevant to Citra are also Updated, and thus keep everything Up-To-Date.

jroweboy: > keep everything Up-To-Date

This is not a goal for CI. Can you imagine the horror when you wake up one morning to find citra is now broken on mac and linux because an external library update broke things? Having frozen dependencies is a feature not a bug.

True story: a recent appveyor update broke mingw builds (our most popular download with something like 95% usage) and we scrambled to find the fix. Several things broke when appveyor silently updated cmake and sdl and it took many many hours to find out what the issues were. Auto updates are a maintenance nightmare when it goes wrong.

Please please please do not just update things because you think newer is better. Stability trumps minor updates everytime. That said, I agree that we should bump the osx image since the other is deprecated, but I strongly oppose doing a brew upgrade and a apt-get upgrade.

MerryMage: > The commands were added as some libraries relevant to Citra are also Updated 

No libraries relevant to citra will be updated with this command. We do not use host libraries in the build process. The build process is contained within the docker image.

The host environment does not matter, and is irrelevant.

.travis/linux/deps.sh
MerryMage: See below on why this is irrelevant.

.travis/macos/deps.sh
MerryMage: And? We do not use ruby.

Furthermore, adding this command would not even remove that warning.

CDAGaming: https://travis-ci.org/CDAGaming/citra/jobs/309775109
(Build from my Fork running this PR, that warning does not show.)

wwylele:
> Both the MINGW and MSVC Builds have been tested for approx. 1 Week prior to the posting of this PR with no noticeable issues.

Honestly speaking, I don't believe your "test" any more. Every time you propose an external update (catch, cryptopp, curl etc.) and I ask what you test exactly, you response are always "I **play games** with the update for a long time without issue", without realizing that many of these externals do not have direct effect in game play at all.

CDAGaming:
Well @wwylele Considering enet is just an Update to it's CMakeLists and nihstro is just a bugfix for **compiling** on Boost 1.66.0, the only possible test I could perhaps do would just be compilation (And using Boost 1.66.0 for a "full" test. That's why I only did minor submodule changes this time around. You can feel free to suggest what i could test these specific submodules by.

wwylele:
@CDAGaming then why would you even touch Travis CI?

CDAGaming:
@wwylele The whole point of this PR, as mentioned it above, almost entirely focuses just on Travis, to just clear out some warnings I've seen for awhile on it and to also clear out deprecation issues (Such as XCode)

wwylele:
1. always separate your PR for unrelated topics even it would be extremely small
2. your "Test Info" section in the description mad it as if you have tested all the stuff including Travis CI, but did you really do sufficient test?

CDAGaming:
If i had to split it up, then i can split this this into 2: One for the minor Submodules, the other for the Travis CI Warning Fixes. All I've really tested was just what it effected (Which, to me, was just compiling, and under specific conditions like Boost 1.66.0)

.travis/macos/deps.sh
MerryMage: That’s because the image has a correct ruby environment, not because of this command. Time travel is not feasible at this point in time, unfortunately.

.travis/linux/deps.sh
j-selby: The only host applications we are even touching with Travis are:
- `bash`, for which there are no real reasons why we should be updating (we don't need any new features, and has existed since 1989, so is likely bug free for our purposes).
- `tar`/`7z` for compression. Again, Unless there was specific security/compression reasons why we should be updating, this is not needed.
- `docker`. Changes to this could potentially break our build, as docker moves (relatively) quickly.
- `git`. We don't need fancy stuff to work out commit information, so updating is totally ineffectual, and could introduce unexpected changes (what if a new major version comes out with a different output format, or renamed commands?).
- `core-utils`. Again, no real reason to update.

Updating the entire build environment therefore is a waste of build time, and literally won't do *anything* except break the build.

I get it - it is tempting to want to update code like you might a rolling operating system, such as Arch, or the Windows insider stuff, but when it is only going to break stuff, it is only additional maintenance when a bug that was introduced has to be fixed later down the line by someone else.

.travis/macos/deps.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh -ex
 
 brew update
+brew upgrade
```
j-selby: See above comment.

CDAGaming:
Rebased Again, and Edited Accordingly from Reviews.
(_Off to bed for me now_)

yuriks:
I see that most of the PR ended up being reverted but... are there objections to merging the two submodule updates and the Xcode version bump? The Xcode bump seems beneficial to us as far as getting a newer clang goes, and the submodule updates seem harmless enough.

CDAGaming:
Once I Merge the Ubuntu Update as discussed In Discord, This'll be ready to merge :)

wwylele:
@CDAGaming what is it? could you provide a reference of the discussion?

CDAGaming:
All the Ubuntu Update of this PR, and Yuzu's, is as follows:
Update to 17.10 for Citra
Downgrade to 17.10 for Yuzu (As 18.04 is not fully released yet)

Once 18.04 is released both can be put to pull from that version again.

wwylele:
@CDAGaming could you open a new PR for it please? Don't put unrelated stuff together

CDAGaming:
Sure, I can do that. In that case, then this PR would be ready to merge.

PR #3171, opened 2017-12-01
fixed AVX

KIYOX:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3171)


jroweboy:
Compilers already emit vzeroupper before this, so this does nothing. See https://github.com/citra-emu/citra/pull/3145#issuecomment-346982923

yuriks:
That said, if there *is* some situation where there's evidence that a compiler doesn't properly emit it, it would be interesting to hear about it.

PR #3173, opened 2017-12-01
Services/NFC: Stub StartTagScanning as it should be for o3DS

shinyquagsire23:
Fixes an issue in Smash 4's version 1.1.7 where a Kirby amiibo (or an invalid Amiibo) would constantly be scanned in.

Before:
Smash 4 Kirby Amiibo

After:
Smash 4 Fixed Kirby Amiibo

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3173)

### DIFF
--- a/src/core/hle/service/nfc/nfc.cpp
+++ b/src/core/hle/service/nfc/nfc.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include "core/hle/ipc.h"
+#include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/kernel/handle_table.h"
 #include "core/hle/service/nfc/nfc.h"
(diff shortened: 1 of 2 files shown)


src/core/hle/service/nfc/nfc.cpp
```diff
@@ -58,7 +58,9 @@ void StartTagScanning(Interface* self) {
     nfc_tag_state = TagState::TagInRange;
     tag_in_range_event->Signal();
 
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    cmd_buff[1] = ResultCode(ErrCodes::CommandInvalidForState, ErrorModule::NFC,
```
B3n30: please us ipc helper

src/core/hle/service/nfc/nfc.cpp
```diff
@@ -58,7 +58,9 @@ void StartTagScanning(Interface* self) {
     nfc_tag_state = TagState::TagInRange;
```
Subv: We're returning an error but still setting these values as if a Tag were found, what gives?

src/core/hle/service/nfc/nfc.cpp
```diff
@@ -58,7 +58,9 @@ void StartTagScanning(Interface* self) {
...
     tag_in_range_event->Signal();
```
B3n30: This should probably use #3101 as soon as it gets merged with a timeout

mailwl:
doesn't it break mini-mario?

src/core/hle/service/nfc/nfc.cpp
MerryMage: We're not though, it's dead code.

Subv: For some reason Github didn't hide my comment after @shinyquagsire23 changed the code. It's an outdated comment from when the TagInRange status was set unconditionally.

PR #3181, opened 2017-12-03
ipc_helper: split PushMoveObjects and PushCopyObjects

wwylele:
Although the move/copy bit in the handle descriptor doesn't mean anything to HLE service, it does make a different from client's point of view. A game can specifically check this bit if it matches expected result. IIRC I observed a similar checking code in service, which can be reused by games. So we should allow HLE service be able to specify this move/copy bit.

Among the four current user of `PushObjects`, `SRV::EnableNotification` was specifically marked as copying handles before `HLERequestContext` implemented, so this behaviour is restored here. ~~The other three don't have clear documentation anywhere whether they were move or copy, so I left them as-is.~~ According to Subv, they are indeed all `Move`

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3181)

### DIFF
--- a/src/core/hle/service/fs/archive.cpp
+++ b/src/core/hle/service/fs/archive.cpp
@@ -206,7 +206,7 @@ void File::OpenLinkFile(Kernel::HLERequestContext& ctx) {
     ClientConnected(std::get<SharedPtr<ServerSession>>(sessions));
 
     rb.Push(RESULT_SUCCESS);
-    rb.PushObjects(std::get<SharedPtr<ClientSession>>(sessions));
+    rb.PushMoveObjects(std::get<SharedPtr<ClientSession>>(sessions));
 }
 
 File::~File() {}
--- a/src/core/hle/service/sm/srv.cpp
+++ b/src/core/hle/service/sm/srv.cpp
@@ -66,7 +66,7 @@ void SRV::EnableNotification(Kernel::HLERequestContext& ctx) {
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 2);
     rb.Push(RESULT_SUCCESS);
-    rb.PushObjects(notification_semaphore);
+    rb.PushCopyObjects(notification_semaphore);
     LOG_WARNING(Service_SRV, "(STUBBED) called");
 }
 
@@ -114,7 +114,7 @@ void SRV::GetServiceHandle(Kernel::HLERequestContext& ctx) {
                   (*session)->GetObjectId());
         IPC::RequestBuilder rb = rp.MakeBuilder(1, 2);
         rb.Push(session.Code());
-        rb.PushObjects(std::move(session).Unwrap());
+        rb.PushMoveObjects(std::move(session).Unwrap());
     } else if (session.Code() == Kernel::ERR_MAX_CONNECTIONS_REACHED && wait_until_available) {
         LOG_WARNING(Service_SRV, "called service=%s -> ERR_MAX_CONNECTIONS_REACHED", name.c_str());
         // TODO(Subv): Put the caller guest thread to sleep until this port becomes available again.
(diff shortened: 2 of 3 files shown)


src/core/hle/service/fs/archive.cpp
```diff
@@ -206,7 +206,8 @@ void File::OpenLinkFile(Kernel::HLERequestContext& ctx) {
     ClientConnected(std::get<SharedPtr<ServerSession>>(sessions));
 
     rb.Push(RESULT_SUCCESS);
-    rb.PushObjects(std::get<SharedPtr<ClientSession>>(sessions));
+    // TODO: confirm that this is Move, not Copy
```
B3n30: nit: TODO (wwylele): 
same for the other todo's
or are you going to figure it out before we merge it.
Besides that: LGTM

src/core/hle/service/sm/srv.cpp
```diff
@@ -66,7 +66,7 @@ void SRV::EnableNotification(Kernel::HLERequestContext& ctx) {
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 2);
     rb.Push(RESULT_SUCCESS);
-    rb.PushObjects(notification_semaphore);
+    rb.PushCopyObjects(notification_semaphore);
```
B3n30: I think notification_semaphore could also be std:move()'d

src/core/hle/service/fs/archive.cpp
```diff
@@ -206,7 +206,8 @@ void File::OpenLinkFile(Kernel::HLERequestContext& ctx) {
...
+    rb.PushMoveObjects(std::get<SharedPtr<ClientSession>>(sessions));
```
Subv: The descriptor at cmd_buff[4] for this call is 16

src/core/hle/service/sm/srv.cpp
```diff
@@ -114,7 +114,8 @@ void SRV::GetServiceHandle(Kernel::HLERequestContext& ctx) {
                   (*session)->GetObjectId());
         IPC::RequestBuilder rb = rp.MakeBuilder(1, 2);
         rb.Push(session.Code());
-        rb.PushObjects(std::move(session).Unwrap());
+        // TODO: confirm that this is Move, not Copy
+        rb.PushMoveObjects(std::move(session).Unwrap());
```
Subv: The descriptor for this one is also 16

src/core/hle/service/sm/srv.cpp
```diff
@@ -204,7 +205,8 @@ void SRV::RegisterService(Kernel::HLERequestContext& ctx) {
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 2);
     rb.Push(RESULT_SUCCESS);
-    rb.PushObjects(port.Unwrap());
+    // TODO: confirm that this is Move, not Copy
+    rb.PushMoveObjects(port.Unwrap());
```
Subv: Descriptor is also 16

src/core/hle/service/sm/srv.cpp
wwylele: No, `notification_semaphore` is a member of `SRV`, and we want to keep a handle(pointer) to the semaphore so we can change its state when needed. 

This is actually akin to what `Move/Copy` does on real hardware. `Move` means the sender won't keep the handle of the object, which in HLE services means not to keep the pointer of it, so you would expect a `std::move` for `PushMoveObjects`. `Copy` means the sender still keep the handle/pointer of the object, so there shouldn't be a `std::move`, but a pointer copy instead. This actually a good example how distinguish between `PushMoveObjects` and `PushCopyObjects` will help writing HLE services

wwylele:
Updated

B3n30:
@wwylele How about you remove the `TODO`s and we get this merged?

PR #3187, opened 2017-12-04
Update Compact

joaozin23:
Compct game Citra

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3187)


Hexagon12:
Requesting to close.

CDAGaming:
Requesting to Close.
joaozin, Stop Spamming the Citra Network with your PRs, as they do nothing to improve Citra(Even if they would, you just change one line and that's not even giving an effort).

jroweboy:
i do think we should probably mention the compatibility wiki in the readme tho

PR #3192, opened 2017-12-05
ldr_ro: convert to ServiceFramework

wwylele:
Also fixes all `process` usage, and creates storage for each client.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3192)

### DIFF
--- a/src/core/hle/service/ldr_ro/ldr_ro.cpp
+++ b/src/core/hle/service/ldr_ro/ldr_ro.cpp
@@ -9,10 +9,8 @@
 #include "core/core.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/process.h"
-#include "core/hle/kernel/vm_manager.h"
 #include "core/hle/service/ldr_ro/cro_helper.h"
 #include "core/hle/service/ldr_ro/ldr_ro.h"
-#include "core/hle/service/ldr_ro/memory_synchronizer.h"
 
 namespace Service {
 namespace LDR {
@@ -42,11 +40,6 @@ static const ResultCode ERROR_NOT_LOADED = // 0xD8A12C0D
     ResultCode(static_cast<ErrorDescription>(13), ErrorModule::RO, ErrorSummary::InvalidState,
                ErrorLevel::Permanent);
 
-static MemorySynchronizer memory_synchronizer;
-
-// TODO(wwylele): this should be in the per-client storage when we implement multi-process
-static VAddr loaded_crs; ///< the virtual address of the static module
-
 static bool VerifyBufferState(VAddr buffer_ptr, u32 size) {
     auto vma = Kernel::g_current_process->vm_manager.FindVMA(buffer_ptr);
     return vma != Kernel::g_current_process->vm_manager.vma_map.end() &&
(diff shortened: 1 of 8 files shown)


Subv:
Reviewed 3 of 3 files at r1, 2 of 2 files at r2, 4 of 4 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3192)*

src/core/hle/service/ldr_ro/memory_synchronizer.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <vector>
+#include "core/hle/kernel/process.h"
```
lioncash: Forward declaration?

src/core/hle/service/ldr_ro/ldr_ro.h
```diff
@@ -4,19 +4,156 @@
...
+     *      1 : CRS buffer pointer
+     *      2 : CRS Size
+     *      3 : Process memory address where the CRS will be mapped
+     *      4 : handle translation descriptor (zero)
+     *      5 : KProcess handle
+     *  Outputs:
+     *      0 : Return header
+     *      1 : Result of function, 0 on success, otherwise error code
+     */
+    void Initialize(Kernel::HLERequestContext& ctx);
```
lioncash: As far as I can tell, the functions can also be made private, unless I'm mistaken, as everything goes through the internal function table.

src/core/hle/service/ldr_ro/ldr_ro.h
```diff
@@ -4,19 +4,156 @@
 
 #pragma once
 
+#include <unordered_map>
+#include "core/hle/kernel/kernel.h"
```
lioncash: Unless I'm mistaken, this can be a forward declaration, despite the LoadCRO template, as it's passing a reference to another function taking a reference (so the reference isn't dereferenced).

wwylele: actually I don't even need the declaration because it must have been declared by the service framework... will just remove this

wwylele:
addressed

wwylele:
Removed canary merge tag and let #3227 go into test first.

src/core/hle/service/ldr_ro/ldr_ro.cpp
```diff
@@ -582,44 +490,36 @@ static void LinkCRO(Interface* self) {
...
     CROHelper cro(cro_address);
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
 
-    if (loaded_crs == 0) {
+    auto find_result = slots.find(process->process_id);
+    if (find_result == slots.end()) {
         LOG_ERROR(Service_LDR, "Not initialized");
         rb.Push(ERROR_NOT_INITIALIZED);
+        rb.Push<u32>(0);
```
B3n30: Adding an additional normal parameter, so IPC::RequestBuilder rb = rp.MakeBuilder(1, 0); must be different for that case
Why is this parameter even necessary?

src/core/hle/service/ldr_ro/ldr_ro.cpp
```diff
@@ -635,83 +535,72 @@ static void UnlinkCRO(Interface* self) {
...
+    LOG_DEBUG(Service_LDR, "called, crs_buffer_ptr=0x%08X", crs_buffer_ptr);
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
 
-    if (loaded_crs == 0) {
+    auto find_result = slots.find(process->process_id);
+    if (find_result == slots.end()) {
         LOG_ERROR(Service_LDR, "Not initialized");
         rb.Push(ERROR_NOT_INITIALIZED);
+        rb.Push<u32>(0);
```
B3n30: Same as above

src/core/hle/service/ldr_ro/ldr_ro.cpp
wwylele: oops, copy-paste error

wwylele:
(Hold on. need to fix some stuff)

wwylele:
Rebased and addressed comments. Also added a TODO about passing process handle to CRO helper. I didn't finish this because CRO helper uses lots of Read/Write 8/16/32 which haven't got their overload with process parameter. I don't want to put too many change in one PR

PR #3225, opened 2017-12-06
Applet: Print unknown AppletId on ASSERT

B3n30:
This adds the Applet ID to the message on Assert when a unknown id is encountered.

We recently get some messages that games crash on that point. The games usually crash later when citra tries to emulate the applet.

This will help to identify which applets are missing. Dumping that applet and adding the id to the list of applets could fix those games.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3225)

### DIFF
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -114,7 +114,7 @@ static u64 GetTitleIdForApplet(AppletId id) {
                                 return data.applet_ids[0] == id || data.applet_ids[1] == id;
                             });
 
-    ASSERT_MSG(itr != applet_titleids.end(), "Unknown applet id %u", id);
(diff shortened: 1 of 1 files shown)


LogPod:
LGTMINSTAMERGE :shipit:

src/core/hle/service/apt/apt.cpp
```diff
@@ -114,7 +114,7 @@ static u64 GetTitleIdForApplet(AppletId id) {
                                 return data.applet_ids[0] == id || data.applet_ids[1] == id;
                             });
 
-    ASSERT_MSG(itr != applet_titleids.end(), "Unknown applet id");
+    ASSERT_MSG(itr != applet_titleids.end(), "Unknown applet id %u", id);
```
wwylele: `static_cast<u32>(id)`

B3n30:
Fixed. Also changed the output to hex

PR #3153, opened 2017-11-26
[WIP] dynarmic updates

MerryMage:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3153)

PR #3227, opened 2017-12-06
Allow for partial invalidation of instruction cache

MerryMage:
Add the necessary interface to ARM_Interface, then uses it in the ro service. Reduces stuttering during loading and unloading of CROs.

Based on @Phanto-m's suggestion [here](https://github.com/Phanto-m/dynarmic/commit/0b522d462a153ca0c09a4d10fb7dadad3cecee0b#commitcomment-26017040).

Previous PR: #2925

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3227)

### DIFF
--- a/src/core/arm/arm_interface.h
+++ b/src/core/arm/arm_interface.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <cstddef>
 #include "common/common_types.h"
 #include "core/arm/skyeye_common/arm_regformat.h"
 #include "core/arm/skyeye_common/vfp/asm_vfp.h"
@@ -32,6 +33,13 @@ public:
 
     /// Clear all instruction cache
     virtual void ClearInstructionCache() = 0;
+
+    /**
+     * Invalidate the code cache at a range of addresses.
+     * @param start_address The starting address of the range to invalidate.
+     * @param length The length (in bytes) of the range to invalidate.
+     */
+    virtual void InvalidateCacheRange(u32 start_address, size_t length) = 0;
 
     /// Notify CPU emulation that page tables have changed
     virtual void PageTableChanged() = 0;
--- a/src/core/arm/dynarmic/arm_dynarmic.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic.cpp
@@ -187,6 +187,10 @@ void ARM_Dynarmic::ClearInstructionCache() {
     }
 }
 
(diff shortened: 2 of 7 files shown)


chris062689:
This PR supercedes and replaces #2925, correct?

MerryMage:
Correct.

jroweboy:
The only thing i noticed that was strange was moving the cache invalidation from link/unlink to load/unload, but I saw that wwylele suggested it here https://github.com/citra-emu/citra/pull/2925#pullrequestreview-61809683 so I suppose that is fine. This also updates dynarmic, meaning it grabs a few other changes that were already tested with https://github.com/citra-emu/citra/pull/3153 so :shipit:

src/core/arm/arm_interface.h
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <cstddef>
```
B3n30: why is this included here?

wwylele:
Just to clarify what @jroweboy said: the invalidation wasn't "moved" from link/unlink to load/unload; it is removed because it is unnecessary after we adding fine-grain invalidation in `Apply/ClearRelocation` - they are the only functions that link/unlink can call to cause change in code region.

src/core/hle/service/ldr_ro/memory_synchronizer.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include <algorithm>
 #include "common/assert.h"
+#include "core/arm/arm_interface.h"
+#include "core/core.h"
```
B3n30: Why are those includes necessary?

MerryMage: Nice catch.

src/core/arm/arm_interface.h
MerryMage: for `size_t`.

MerryMage:
Isolated dynarmic update to one commit.

src/core/arm/arm_interface.h
B3n30: since this includes common_types.h which includes cstddef it isn't really necessary

jroweboy: @B3n30 I'm of the opinion that its better to have the explicit include instead of the indirect include, but i don't really care in this circumstance

PR #3234, opened 2017-12-07
HLE/FS: Implemented FS::GetProgramLaunchInfo.

Subv:
This function is used by the DLP system module during the DLPSRVR initialization.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3234)

### DIFF
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
 #include <memory>
 #include "common/assert.h"
 #include "common/common_funcs.h"
@@ -15,6 +16,9 @@
 #include "core/memory.h"
 
 namespace Kernel {
+
+// Lists all processes that exist in the current session.
+static std::vector<SharedPtr<Process>> process_list;
 
 SharedPtr<CodeSet> CodeSet::Create(std::string name, u64 program_id) {
     SharedPtr<CodeSet> codeset(new CodeSet);
@@ -37,6 +41,7 @@ SharedPtr<Process> Process::Create(SharedPtr<CodeSet> code_set) {
     process->flags.raw = 0;
     process->flags.memory_region.Assign(MemoryRegion::APPLICATION);
 
+    process_list.push_back(process);
     return process;
 }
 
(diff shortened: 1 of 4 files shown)


src/core/hle/kernel/process.cpp
```diff
@@ -299,5 +302,16 @@ ResultCode Process::LinearFree(VAddr target, u32 size) {
 Kernel::Process::Process() {}
 Kernel::Process::~Process() {}
 
-SharedPtr<Process> g_current_process;
+SharedPtr<Process> GetProcessById(u32 process_id) {
+    auto itr = std::find_if(
```
lioncash: `<algorithm>` needs to be included.

src/core/hle/kernel/process.h
```diff
@@ -175,5 +176,11 @@ class Process final : public Object {
     ~Process() override;
 };
 
+/// Retrieves a process from the current list of processes.
+SharedPtr<Process> GetProcessById(u32 process_id);
+
+// Lists all processes that exist in the current session.
+extern std::vector<SharedPtr<Process>> process_list;
```
lioncash: Given this is only exposed so it can be cleared, would it be better to just make a function a part of the external interface that just does that?

src/core/hle/kernel/process.cpp
```diff
@@ -16,6 +17,9 @@
 
 namespace Kernel {
 
+// Lists all processes that exist in the current session.
+static std::vector<SharedPtr<Process>> process_list;
```
bunnei: ugh at introducing additional static state

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -920,6 +922,51 @@ static void GetFormatInfo(Service::Interface* self) {
...
+    auto process = Kernel::GetProcessById(process_id);
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(5, 0);
+
+    if (process == nullptr) {
+        // Note: In this case, the rest of the parameters are not changed but the command header
+        // remains the same.
+        rb.Push(ResultCode(FileSys::ErrCodes::ArchiveNotMounted, ErrorModule::FS,
+                           ErrorSummary::NotFound, ErrorLevel::Status));
+        return;
```
wwylele: Since `rp.MakeBuilder(5, 0);`, you need to skip 4 words of output here, or define a different param number here

Subv:
Addressed

PR #3239, opened 2017-12-08
cam: convert to ServiceFramework

wwylele:
RE shows that three services "cam:u", "cam:c" and "cam:s" are handled by the same function, and they share the same handler table. It is possible that "cam:s" or "cam:c" has some additional function that "cam:u" can't access (blocked by permission checking inside the handler), but this isn't confirmed yet.

The services "cam:q" is strange. 3dbrew reports that it is added to new 3DS (for QTM probably), but I can't find the `RegisterService("cam:q")` call in camera module code, although there is a string "cam:q" in the module. In the initialization code, the services registration code is hard coded as a loop of 4 iterations, over the array "cam:u", "cam:c", "cam:s" and "y2r:u" (Yes, Y2R is in the camera module). Probably "cam:q" isn't a real service, or it is in another module. For now I just keep it blank.

Besides filling cam:c and cam:s handler table, the only functional change in this PR is that the camera service now respects to the process handle, and transfers data accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3239)


src/core/hle/service/cam/cam.cpp
```diff
@@ -3,89 +3,25 @@
...
-        is_pending_receiving = false;
-        is_busy = false;
-        is_trimming = false;
-        x0 = 0;
-        y0 = 0;
-        x1 = 0;
-        y1 = 0;
-        transfer_bytes = 256;
-    }
-};
```
wwylele: review guide： these structs are moved to cam.h, inside `CAM::Module`. The definition of `PortConfig::Clear` is moved below

src/core/hle/service/cam/cam.cpp
```diff
@@ -115,16 +51,28 @@ constexpr std::array<int, 13> LATENCY_BY_FRAME_RATE{{
     33,  // Rate_30_To_10
 }};
 
-std::array<CameraConfig, NumCameras> cameras;
-std::array<PortConfig, 2> ports;
-CoreTiming::EventType* completion_event_callback;
```
wwylele: review guide: these are moved to cam.h, becoming members of `CAM::Module`

src/core/hle/service/cam/cam.cpp
```diff
@@ -177,16 +125,15 @@ void CompletionEventCallBack(u64 port_id, int) {
...
+        Memory::WriteBlock(*port.dest_process, port.dest, buffer.data(),
+                           std::min<size_t>(port.dest_size, buffer_size));
     }
 
     port.is_receiving = false;
     port.completion_event->Signal();
 }
 
-// Starts a receiving process on the specified port. This can only be called when is_busy = true and
-// is_receiving = false.
```
wwylele: review guide: comments are moved to header file

PR #3240, opened 2017-12-08
Add button labels for sdl joystick mappings

muemart:
This PR replaces the [non-keyboard] placeholders for gamepads in the input configuration dialog by labels actually telling you the mappings. Here's how it looks now:
inputs

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3240)

### DIFF
--- a/src/citra_qt/configuration/configure_input.cpp
+++ b/src/citra_qt/configuration/configure_input.cpp
@@ -161,34 +161,63 @@ void ConfigureInput::restoreDefaults() {
 }
 
 void ConfigureInput::updateButtonLabels() {
-    QString non_keyboard(tr("[non-keyboard]"));
+    QString unknown_mapping(tr("[unknown]"));
 
-    auto KeyToText = [&non_keyboard](const Common::ParamPackage& param) {
+    auto ButtonToText = [&unknown_mapping](const Common::ParamPackage& param) {
         if (!param.Has("engine")) {
             return QString("[not set]");
-        } else if (param.Get("engine", "") != "keyboard") {
-            return non_keyboard;
-        } else {
+        } else if (param.Get("engine", "") == "keyboard") {
             return getKeyName(param.Get("code", 0));
+        } else if (param.Get("engine", "") == "sdl") {
+            QString text = QString("Joystick %1").arg(param.Get("joystick", "").c_str());
+            if (param.Has("hat")) {
+                text += QString(" Hat %1 %2")
+                            .arg(param.Get("hat", "").c_str(), param.Get("direction", "").c_str());
+            }
+            if (param.Has("axis")) {
+                text += QString(" Axis %1%2")
+                            .arg(param.Get("axis", "").c_str(), param.Get("direction", "").c_str());
+            }
+            if (param.Has("button")) {
+                text += QString(" Button %1").arg(param.Get("button", "").c_str());
+            }
+            return text;
+        } else {
+            return unknown_mapping;
+        }
+    };
+    auto AnalogToText = [&unknown_mapping, &ButtonToText](const Common::ParamPackage& param,
+                                                          const std::string& dir) {
+        if (!param.Has("engine")) {
+            return QString("[not set]");
+        } else if (param.Get("engine", "") == "analog_from_button") {
+            return ButtonToText(Common::ParamPackage{param.Get(dir, "")});
+        } else if (param.Get("engine", "") == "sdl") {
+            if (dir == "modifier") {
+                return QString("[unused]");
+            }
+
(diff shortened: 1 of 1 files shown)


src/citra_qt/configuration/configure_input.cpp
```diff
@@ -161,34 +161,63 @@ void ConfigureInput::restoreDefaults() {
 }
 
 void ConfigureInput::updateButtonLabels() {
-    QString non_keyboard(tr("[non-keyboard]"));
+    QString unknown_mapping(tr("[unknown]"));
 
-    auto KeyToText = [&non_keyboard](const Common::ParamPackage& param) {
+    auto ButtonToText = [&unknown_mapping](const Common::ParamPackage& param) {
         if (!param.Has("engine")) {
             return QString("[not set]");
```
wwylele: Since you are on this, could you change it to `tr(...)`?

src/citra_qt/configuration/configure_input.cpp
```diff
@@ -161,34 +161,63 @@ void ConfigureInput::restoreDefaults() {
...
-        } else if (param.Get("engine", "") != "keyboard") {
-            return non_keyboard;
-        } else {
+        } else if (param.Get("engine", "") == "keyboard") {
             return getKeyName(param.Get("code", 0));
+        } else if (param.Get("engine", "") == "sdl") {
+            QString text = QString("Joystick %1").arg(param.Get("joystick", "").c_str());
```
wwylele: `tr("Joystick %1")`

src/citra_qt/configuration/configure_input.cpp
```diff
@@ -161,34 +161,63 @@ void ConfigureInput::restoreDefaults() {
...
+            if (param.Has("hat")) {
+                text += QString(" Hat %1 %2")
```
wwylele: `tr(" Hat %1 %2")`

src/citra_qt/configuration/configure_input.cpp
```diff
@@ -161,34 +161,63 @@ void ConfigureInput::restoreDefaults() {
...
+                            .arg(param.Get("hat", "").c_str(), param.Get("direction", "").c_str());
+            }
+            if (param.Has("axis")) {
+                text += QString(" Axis %1%2")
```
wwylele: `tr(" Axis %1%2")`

src/citra_qt/configuration/configure_input.cpp
```diff
@@ -161,34 +161,63 @@ void ConfigureInput::restoreDefaults() {
...
+                            .arg(param.Get("axis", "").c_str(), param.Get("direction", "").c_str());
+            }
+            if (param.Has("button")) {
+                text += QString(" Button %1").arg(param.Get("button", "").c_str());
```
wwylele: `tr(" Button %1")`. I will stop commenting on the similar thing

muemart:
Added translation support.

src/citra_qt/configuration/configure_input.cpp
```diff
@@ -161,37 +161,67 @@ void ConfigureInput::restoreDefaults() {
 }
 
 void ConfigureInput::updateButtonLabels() {
-    QString non_keyboard(tr("[non-keyboard]"));
+    QString unknown_mapping(tr("[unknown]"));
+    QString mapping_not_set(tr("[not set]"));
 
-    auto KeyToText = [&non_keyboard](const Common::ParamPackage& param) {
+    auto ButtonToText = [&unknown_mapping, &mapping_not_set](const Common::ParamPackage& param) {
```
Subv: I would prefer if these two lambdas were static functions instead. They are quite big and detract from the readability of the updateButtonLabels function

src/citra_qt/configuration/configure_input.cpp
```diff
@@ -41,6 +41,52 @@ static void SetAnalogButton(const Common::ParamPackage& input_param,
     analog_param.Set(button_name, input_param.Serialize());
 }
 
+static QString ButtonToText(const Common::ParamPackage& param) {
+    if (!param.Has("engine")) {
+        return QObject::tr("[not set]");
```
wwylele: IIRC `QObject::tr` shouldn't be used like this. You can still make these functions as member function and use the non-static version of tr instead.

muemart: The function getKeyName above does it the same way, I just copied it from there. I can't find anything in the docs explicitly forbidding this use in a quick search, but I guess we'll be on the safe side by moving the functions into the class.

wwylele: nvm I was confused with other problems

PR #3241, opened 2017-12-08
svc: correct GetProcessInfo(20)

wwylele:
This return value is meant to be used as a simple conversion between linear heap PAddr<->VAddr, thus it only very depends on where the entire FCRAM is mapped in VAddr (can be only 0x30000000 / 0x14000000). It shouldn't use `GetLinearHeapBase` which is for the APPLICATION/SYSTEM/BASE region base address.

Fixes new version swkbd not rendering issue

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3241)

### DIFF
--- a/src/core/hle/svc.cpp
+++ b/src/core/hle/svc.cpp
@@ -1268,7 +1268,7 @@ static ResultCode GetProcessInfo(s64* out, Kernel::Handle process_handle, u32 ty
         LOG_ERROR(Kernel_SVC, "unimplemented GetProcessInfo type=%u", type);
         break;
     case 20:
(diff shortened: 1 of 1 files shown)


valentinvanelslande:
[wwylele](https://user-images.githubusercontent.com/22228082/33782092-e3a07b32-dc1c-11e7-8143-6a281d8b3014.PNG)
[me](https://user-images.githubusercontent.com/22228082/33782091-e389f588-dc1c-11e7-841d-121d2dc4412a.PNG)

:(

Subv:
lgtminstamerge

This is verified kernel behavior

PR #3249, opened 2017-12-09
NCCH: Load the "logo" file from the specified offset if it's present instead of the ExeFS.

Subv:
This is "new" behavior as of firmware 5.0.0-11, older titles have the logo offset and size set to 0, indicating that the logo is stored in the ExeFS.

The Home Menu tries to open this file when launching a title.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3249)

### DIFF
--- a/src/core/file_sys/ncch_container.cpp
+++ b/src/core/file_sys/ncch_container.cpp
@@ -252,6 +252,26 @@ Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vect
     if (result == Loader::ResultStatus::Success || !has_exefs)
         return result;
 
+    // As of firmware 5.0.0-11 the logo is stored between the access descriptor and the plain region
+    // instead of the ExeFS.
+    if (std::strcmp(name, "logo") == 0) {
+        if (ncch_header.logo_region_offset && ncch_header.logo_region_size) {
+            size_t logo_offset = ncch_header.logo_region_offset * kBlockSize;
+            size_t logo_size = ncch_header.logo_region_size * kBlockSize;
+
+            buffer.resize(logo_size);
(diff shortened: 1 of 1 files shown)


src/core/file_sys/ncch_container.cpp
```diff
@@ -252,6 +252,26 @@ Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vect
     if (result == Loader::ResultStatus::Success || !has_exefs)
         return result;
 
+    // As of firmware 5.0.0-11 the logo is stored between the access descriptor and the plain region
+    // instead of the ExeFS.
+    if (!strcmp(name, "logo")) {
```
wwylele: `std::strcmp`
And since it doesn't returns a bool, the usage of `!` is confusing IMO. I prefer `== 0` instead.

src/core/file_sys/ncch_container.cpp
```diff
@@ -252,6 +252,26 @@ Loader::ResultStatus NCCHContainer::LoadSectionExeFS(const char* name, std::vect
...
+            buffer.resize(logo_size);
+            file.Seek(ncch_offset + logo_offset, SEEK_SET);
+
+            if (file.ReadBytes(buffer.data(), logo_size) != logo_size) {
+                LOG_ERROR(Service_FS, "Could not read NCCH logo");
+                return Loader::ResultStatus::Error;
+            }
+            return Loader::ResultStatus::Success;
+        } else {
+            LOG_WARNING(Service_FS, "Attempting to load logo from the ExeFS");
```
wwylele: This is a common case for pre-5.0.0-11 ROMs, so the log level should be just `INFO`

PR #3251, opened 2017-12-09
HLE/FS: Stubbed GetNumSeeds to always return 0 seeds.

Subv:
We don't yet implement the SEEDDB store.
This number is supposed to be modified by the AddSeed function (0x087A0180) but there's too little info about it on 3dbrew.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3251)

### DIFF
--- a/src/core/hle/service/fs/fs_user.cpp
+++ b/src/core/hle/service/fs/fs_user.cpp
@@ -968,6 +968,26 @@ static void GetProgramLaunchInfo(Service::Interface* self) {
     rb.Push<u32>(0);
 }
 
+/**
+ * FS_User::GetNumSeeds service function.
+ *  Inputs:
+ *      0 : 0x087D0000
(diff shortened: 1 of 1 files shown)


src/core/hle/service/fs/fs_user.cpp
```diff
@@ -920,6 +920,26 @@ static void GetFormatInfo(Service::Interface* self) {
...
 }
 
+/**
+ * FS_User::GetNumSeeds service function.
+ *  Inputs:
+ *      0 : 0x087D0000
+ *  Outputs:
+ *      0 : 0x087D0080
+ *      1 : Result of function, 0 on success, otherwise error code
+ *      2 : Number of seeds in the SEEDB
```
valentinvanelslande: SEEDDB, no SEEDB

PR #3253, opened 2017-12-09
mic_u: Migrate to the new service framework

lioncash:
So, this is my first time using the new IPC helpers and the service framework (which is why I didn't pick a relatively important module to migrate), so if I've made any glaringly wrong issues, please feel free to point them out.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3253)

### DIFF
--- a/src/core/hle/service/mic_u.cpp
+++ b/src/core/hle/service/mic_u.cpp
@@ -4,6 +4,7 @@
 
 #include "common/logging/log.h"
 #include "core/hle/ipc.h"
+#include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/event.h"
 #include "core/hle/kernel/handle_table.h"
 #include "core/hle/kernel/kernel.h"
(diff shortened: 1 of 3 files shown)


src/core/hle/service/mic_u.cpp
```diff
@@ -51,14 +52,18 @@ static bool audio_buffer_loop;
  *      1 : Result of function, 0 on success, otherwise error code
  */
 static void MapSharedMem(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 size = cmd_buff[1];
-    Kernel::Handle mem_handle = cmd_buff[3];
+    IPC::RequestParser rp{Kernel::GetCommandBuffer(), 0x01, 1, 2};
+    const u32 size = rp.Pop<u32>();
+    const Kernel::Handle mem_handle = rp.PopHandle();
```
Subv: you have to pop the SharedMemory with rp.PopObject<SharedMemory>() to let the RequestParser eat the descriptor

src/core/hle/service/mic_u.cpp
```diff
@@ -89,15 +93,16 @@ static void UnmapSharedMem(Interface* self) {
...
+    IPC::RequestParser rp{Kernel::GetCommandBuffer(), 0x03, 5, 0};
 
-    encoding = static_cast<Encoding>(cmd_buff[1] & 0xFF);
-    sample_rate = static_cast<SampleRate>(cmd_buff[2] & 0xFF);
-    audio_buffer_offset = cmd_buff[3];
-    audio_buffer_size = cmd_buff[4];
-    audio_buffer_loop = (cmd_buff[5] & 0xFF) != 0;
+    encoding = static_cast<Encoding>(rp.Pop<u8>());
+    sample_rate = static_cast<SampleRate>(rp.Pop<u8>());
+    audio_buffer_offset = static_cast<s32>(rp.Pop<u32>());
```
Subv: you can use PopRaw<s32>();

src/core/hle/service/mic_u.cpp
```diff
@@ -128,8 +135,8 @@ static void AdjustSampling(Interface* self) {
  *      1 : Result of function, 0 on success, otherwise error code
  */
 static void StopSampling(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    IPC::RequestBuilder rb{Kernel::GetCommandBuffer(), 0x05, 1, 0};
```
Subv: It is preferable to construct the RequestParser and then get the builder from it, even if the function takes no parameters.

src/core/hle/service/mic_u.cpp
```diff
@@ -143,9 +150,9 @@ static void StopSampling(Interface* self) {
  *      2 : 0 = sampling, non-zero = sampling
  */
 static void IsSampling(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    cmd_buff[2] = is_sampling;
+    IPC::RequestBuilder rb{Kernel::GetCommandBuffer(), 0x06, 2, 0};
```
Subv: same

src/core/hle/service/mic_u.cpp
```diff
@@ -158,9 +165,9 @@ static void IsSampling(Interface* self) {
  *      3 : Event handle
  */
 static void GetBufferFullEvent(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    cmd_buff[3] = Kernel::g_handle_table.Create(buffer_full_event).Unwrap();
+    IPC::RequestBuilder rb{Kernel::GetCommandBuffer(), 0x07, 1, 2};
```
Subv: same

src/core/hle/service/mic_u.cpp
```diff
@@ -203,9 +214,11 @@ static void GetGain(Interface* self) {
  *      1 : Result of function, 0 on success, otherwise error code
  */
 static void SetPower(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    mic_power = (cmd_buff[1] & 0xFF) != 0;
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    IPC::RequestParser rp{ Kernel::GetCommandBuffer(), 0x0A, 1, 0};
```
Subv: extra whitespace

src/core/hle/service/mic_u.cpp
```diff
@@ -229,19 +242,19 @@ static void GetPower(Interface* self) {
...
-    u32 size = cmd_buff[1];
-    VAddr buffer = cmd_buff[3];
+    IPC::RequestParser rp{Kernel::GetCommandBuffer(), 0x0C, 1, 2};
+    const u32 size = rp.Pop<u32>();
+    const Kernel::MappedBuffer& buffer = rp.PopMappedBuffer();
 
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    LOG_WARNING(Service_MIC, "(STUBBED) called, size=0x%X, buffer=0x%08X", size, buffer);
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+    rb.Push(RESULT_SUCCESS);
```
Subv: If you pop a MappedBuffer, you have to push it again in the response

src/core/hle/service/mic_u.cpp
```diff
@@ -229,19 +242,19 @@ static void GetPower(Interface* self) {
...
+    LOG_WARNING(Service_MIC, "(STUBBED) called, size=0x%X, buffer=0x%08X", size);
```
Subv: you're missing a parameter here

src/core/hle/service/mic_u.cpp
```diff
@@ -268,43 +283,47 @@ static void SetClamp(Interface* self) {
  *      2 : Clamp (0 = don't clamp, non-zero = clamp)
  */
 static void GetClamp(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
-    cmd_buff[2] = clamp;
+    IPC::RequestBuilder rb{Kernel::GetCommandBuffer(), 0x0E, 2, 0};
```
Subv: construct the parser

src/core/hle/service/mic_u.h
```diff
@@ -4,20 +4,190 @@
...
+     *  Inputs:
+     *      0 : Header Code[0x00100040]
+     *      1 : Used SDK Version
+     *  Outputs:
+     *      1 : Result of function, 0 on success, otherwise error code
+     */
+    void SetClientVersion(Kernel::HLERequestContext& ctx);
+
+    struct Impl;
+    std::unique_ptr<Impl> impl;
```
Subv: what's the purpose of using the pImpl idiom here?

Subv:
I've left some comments but Github hid them for some reason

src/core/hle/service/mic_u.h
lioncash: Avoid the need to expose implementation details in the header, such as members, etc (and other headers as well), which lessens the amount of files that need to be recompiled if an implementation detail changes, and improves compilation speed due the includes only being in the compiled translation unit as opposed to being included into everything else that also includes this header.

Given the services aren't intended to be continually instantiated over and over, it's a more ideal place to utilize it.

lioncash:
I believe it must have been an older iteration of the commit prior to me opening the PR somehow, as the diff already corrects most of the issues you've pointed out (the missing parser ones)

Subv:
About the pImpl idiom, I've never been a fan of it, but I won't block this PR over it

lioncash:
Hopefully addressed the lingering comments that were still valid.

wrt pimpl, if it's not preferable despite those benefits (and the tradeoffs they imply), I can get rid of it, it's not that much work to remove

lioncash:
@CDAGaming I appreciate that you're pointing out clang-format issues, however, in the future you don't need to do that (at least for me), I'll notice if the clang-format check fails. Depending on the discussion I may opt to not resolve the failure right away because I can resolve that along with other stuff, or to prevent annoying someone doing a review.

For example:

- If someone is in the middle of reviewing a PR and I know about it, I won't synchronize my changes immediately so I don't potentially invalidate a large swath of their review that they took the time do.

- If I'm waiting on an answer to a question where, depending on said answer, more changes may be necessary, I usually won't resolve it immediately because if it does require more work, then I can just resolve the clang-fomat warnings over the course of the work.

By all means, if I (or anyone else) has a PR open where almost everything is resolved, but the only thing left to do is resolve clang-format warnings, and the PR has been open for several days with no activity or no comment explaining why (life happens, sometimes people are just really busy), then do reply with a reminder.

src/core/hle/service/mic_u.cpp
```diff
@@ -27,319 +28,270 @@ enum class SampleRate : u8 {
...
-                             "audio_buffer_offset=%d, audio_buffer_size=%u, audio_buffer_loop=%u",
-                static_cast<u32>(encoding), static_cast<u32>(sample_rate), audio_buffer_offset,
-                audio_buffer_size, audio_buffer_loop);
-}
+    void StartSampling(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx, 0x03, 5, 0};
+
+        encoding = static_cast<Encoding>(rp.Pop<u8>());
+        sample_rate = static_cast<SampleRate>(rp.Pop<u8>());
+        rp.PopRaw(audio_buffer_offset);
```
Subv: imo this should be `audio_buffer_offset = rp.PopRaw<s32>()` to be consistent with the rest of the function

src/core/hle/service/mic_u.cpp
```diff
@@ -27,319 +28,270 @@ enum class SampleRate : u8 {
...
+        LOG_WARNING(Service_MIC, "(STUBBED) called");
+    }
+
+    void SetIirFilterMic(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx, 0x0C, 1, 2};
+        const u32 size = rp.Pop<u32>();
+        const Kernel::MappedBuffer& buffer = rp.PopMappedBuffer();
+
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+        rb.Push(RESULT_SUCCESS);
```
Subv: Popping a mapped buffer usually means you have to push it back so that the kernel can unmap it from the server process

src/core/hle/service/mic_u.cpp
```diff
@@ -27,319 +28,270 @@ enum class SampleRate : u8 {
...
-    is_sampling = true;
-    LOG_WARNING(Service_MIC, "(STUBBED) called, encoding=%u, sample_rate=%u, "
-                             "audio_buffer_offset=%d, audio_buffer_size=%u, audio_buffer_loop=%u",
-                static_cast<u32>(encoding), static_cast<u32>(sample_rate), audio_buffer_offset,
-                audio_buffer_size, audio_buffer_loop);
-}
+    void StartSampling(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx, 0x03, 5, 0};
+
+        encoding = static_cast<Encoding>(rp.Pop<u8>());
```
Subv: should probably use rp.PopEnum for this

lioncash:
Should be all addressed

src/core/hle/service/mic_u.cpp
```diff
@@ -27,319 +28,271 @@ enum class SampleRate : u8 {
...
+        rb.Push(RESULT_SUCCESS);
+        rb.Push<bool>(is_sampling);
+        LOG_WARNING(Service_MIC, "(STUBBED) called");
+    }
+
+    void GetBufferFullEvent(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx, 0x07, 0, 0};
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 2);
+        rb.Push(RESULT_SUCCESS);
+        rb.PushCopyHandles(Kernel::g_handle_table.Create(buffer_full_event).Unwrap());
```
wwylele: Use the new function for service framework:
`rb.PushCopyObjects(buffer_full_event);`

src/core/hle/service/mic_u.cpp
```diff
@@ -27,319 +28,271 @@ enum class SampleRate : u8 {
...
+        rb.Push<u8>(mic_power);
+        LOG_WARNING(Service_MIC, "(STUBBED) called");
+    }
+
+    void SetIirFilterMic(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx, 0x0C, 1, 2};
+        const u32 size = rp.Pop<u32>();
+        const Kernel::MappedBuffer& buffer = rp.PopMappedBuffer();
+
+        IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
```
wwylele: The header should be `rp.MakeBuilder(1, 2)`. The 2 is for the `PushMappedBuffer`

lioncash:
Amended

PR #3255, opened 2017-12-10
cryptopp: Build only required files

yuriks:
Instead of globbing all .cpp files that ship with CryptoPP, use an
explicit list of files to compile. This cuts out compilation of a lot of
stuff we don't use (reducing build times) and also works around
https://github.com/weidai11/cryptopp/issues/527

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3255)

### DIFF
--- a/externals/cryptopp/CMakeLists.txt
+++ b/externals/cryptopp/CMakeLists.txt
@@ -116,20 +116,44 @@ endif()
 # Library headers
 file(GLOB cryptopp_HEADERS cryptopp/*.h)
 
-# Library sources. You can use the GNUmakefile to generate the list: `make sources`.
-file(GLOB cryptopp_SOURCES cryptopp/*.cpp)
-list(REMOVE_ITEM cryptopp_SOURCES
-        # These are removed in the original CMakeLists.txt
-        ${CMAKE_CURRENT_SOURCE_DIR}/cryptopp/pch.cpp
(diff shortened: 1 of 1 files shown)


yuriks:
Can someone benchmark compilation times before and after this change? (I can't since the previous version doesn't build on VS 15.5.) I'm curious how much it'll help.

CDAGaming:
I'll post times as soon as I can.
 (Or someone else if they beat me to it)

Subv:
Compilation results on an i7-4710MQ laptop running Windows 10 with Visual Studio 15.3. Build times are as reported by Visual Studio after a successful Clean All + Build All.

Before (25afbe57075e8560e62f10224ad9439b60263472):
Build Time: 6m 59.30s
citra-qt.exe Size: 4,634,624 bytes
citra.exe Size: 3,852,800 bytes


Before (664963d0a4ac3b34e1e5da2e37551fdf1c1bec02):
Build Time: 5m 33.57s
citra-qt.exe Size: 4,635,136 bytes
citra.exe Size: 3,853,312 bytes

noloader:
@yuriks, @CDAGaming, @Subv,

I aw the cross-reference in the Crypto++ BLAKE2 bug.

Ping me directly if you encounter problems or need a hand with Crypto++. *noloader, gmail* account.

PR #3264, opened 2017-12-11
CMakeLists: Derive the source directory grouping from targets themselves

lioncash:
Removes the need to store to separate SRC and HEADER variables, and then construct the target in most cases.


src/audio_core/CMakeLists.txt
```diff
@@ -1,40 +1,42 @@
...
-            hle/mixers.h
-            hle/pipe.h
-            hle/source.h
-            interpolate.h
-            null_sink.h
-            sink.h
-            sink_details.h
-            time_stretch.h
-            )
+    # Header files
```
yuriks: If we're going to stop having the two separate vars, then I'd rather just co-mingle all files so I have to stop scrolling 5 pages down every time to find the .h corresponding to the .cpp I edited.

src/audio_core/CMakeLists.txt
```diff
@@ -1,40 +1,42 @@
...
+)
 
 if(SDL2_FOUND)
-    set(SRCS ${SRCS} sdl2_sink.cpp)
-    set(HEADERS ${HEADERS} sdl2_sink.h)
+    target_sources(audio_core
+        PRIVATE
+            sdl2_sink.cpp
+            sdl2_sink.h
+    )
```
yuriks: Could use generator expressions to put these in the main lists of sources instead of appending separately.

src/citra_qt/CMakeLists.txt
```diff
@@ -3,107 +3,117 @@ set(CMAKE_AUTORCC ON)
...
+        ${UI_HDRS}
+        ${UIS}
+)
+
 if (APPLE)
     set(MACOSX_ICON "../../dist/citra.icns")
     set_source_files_properties(${MACOSX_ICON} PROPERTIES MACOSX_PACKAGE_LOCATION Resources)
-    add_executable(citra-qt MACOSX_BUNDLE ${SRCS} ${HEADERS} ${UI_HDRS} ${ICONS} ${THEMES} ${MACOSX_ICON})
+    target_sources(citra-qt PRIVATE ${MACOSX_ICON})
+    set_target_properties(citra-qt PROPERTIES MACOSX_BUNDLE TRUE)
```
yuriks: What does this do?

yuriks: nvm I'm blind (it replaces the `MACOSX_BUNDLE` in `add_executable`)

PR #3265, opened 2017-12-11
HLE/APT: Implement GetAppletInfo for LLE applets.

Subv:
Calling this function for AppletId::Application is not yet implemented because we don't support launching applications from APT.

The HLE applet machinery is showing its age, it's getting awkward to still support it alongside LLE applets. I want to rewrite it one day to make it use an interface more similar to real applets.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3265)

### DIFF
--- a/src/core/hle/service/apt/apt.cpp
+++ b/src/core/hle/service/apt/apt.cpp
@@ -971,7 +971,20 @@ void GetAppletInfo(Service::Interface* self) {
     IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x6, 1, 0); // 0x60040
     auto app_id = static_cast<AppletId>(rp.Pop<u32>());
 
-    if (auto applet = HLE::Applets::Applet::Get(app_id)) {
+    LOG_DEBUG(Service_APT, "called appid=%u", static_cast<u32>(app_id));
+
+    const auto* slot = GetAppletSlotData(app_id);
+
+    if (slot == nullptr || !slot->registered) {
+        // See if there's an HLE applet and try to use it before erroring out.
+        auto hle_applet = HLE::Applets::Applet::Get(app_id);
+        if (hle_applet == nullptr) {
+            IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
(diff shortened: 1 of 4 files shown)


src/core/hle/service/apt/apt.cpp
```diff
@@ -887,6 +888,20 @@ void PreloadLibraryApplet(Service::Interface* self) {
     }
 }
 
+void FinishPreloadingLibraryApplet(Service::Interface* self) {
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x17, 1, 0); // 0x00170040
+    u32 applet_id = rp.Pop<u32>();
```
B3n30: in all other functions `AppletId applet_id = static_cast<AppletId>(rp.Pop<u32>());` is used

src/core/hle/service/apt/apt.cpp
```diff
@@ -966,12 +994,27 @@ void GetAppletInfo(Service::Interface* self) {
         rb.Push(true);   // Registered
         rb.Push(true);   // Loaded
         rb.Push<u32>(0); // Applet Attributes
-    } else {
+        LOG_WARNING(Service_APT, "Using HLE applet info for applet %03X", static_cast<u32>(app_id));
+        return;
```
B3n30: In this case there is nothing written to the command_buffer?

src/core/hle/service/apt/apt.cpp
```diff
@@ -76,6 +76,7 @@ struct AppletSlotData {
     AppletId applet_id;
     AppletSlot slot;
     bool registered;
+    bool loaded;
```
B3n30: Shouldn't this get initialized for every applet in Init() and also reset in Shutdown()

src/core/hle/service/apt/apt.cpp
Subv: It is written, the else branch was removed.

src/core/hle/service/apt/apt.cpp
Subv: Done

PR #3268, opened 2017-12-11
ipc_helpers: Add member functions for pushing and popping strongly typed enums

lioncash:
Alleviates the need to static_cast to an enum type at the call sites of the Push and Pop calls.

We only allow strongly typed enums, as they have a defined type of int by default if an underlying type isn't explicitly specified, whereas with regular enums, if an underlying type isn't specified, an implementation-defined  type is used that can fit all the enumeration values. So this opts for a "do the default consistent thing" approach.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3268)

### DIFF
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -98,6 +98,15 @@ public:
     template <typename First, typename... Other>
     void Push(const First& first_value, const Other&... other_values);
 
+    template <typename T>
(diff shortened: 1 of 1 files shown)


lioncash:
Whoops, assumed we were on C++17-capable compilers (some builders failed because variable templates were used, I'll amend this a little later, need to step out for a few minutes).

PR #3270, opened 2017-12-11
[RFC] ShaderJIT: Use host regs xmm5-xmm13 to store the first 9 temporary PICA regs (r0-r8).

Subv:
Operations on temporary registers are by far the most common ones, not having to load and store memory after each individual instruction is probably good for performance.

Note: Untested, may break horribly. No performance benchmarks were actually performed on this.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3270)

### DIFF
--- a/src/tests/video_core/shader/shader_jit_x64_compiler.cpp
+++ b/src/tests/video_core/shader/shader_jit_x64_compiler.cpp
@@ -89,3 +89,44 @@ TEST_CASE("EX2", "[video_core][shader][shader_jit]") {
     REQUIRE(shader.Run(79.7262742773f) == Approx(1.e24f));
     REQUIRE(std::isinf(shader.Run(800.f)));
 }
+
+TEST_CASE("Temp Registers", "[video_core][shader][shader_jit]") {
+    const auto sh_input = SourceRegister::MakeInput(0);
+    const auto sh_temp = SourceRegister::MakeTemporary(0);
+    const auto sh_output = DestRegister::MakeOutput(0);
+
+    auto shader = ShaderTest({
+        // clang-format off
+        {OpCode::Id::MOV, sh_temp, sh_input},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
+        {OpCode::Id::MOV, sh_temp, sh_temp},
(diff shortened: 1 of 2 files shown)


src/tests/video_core/shader/shader_jit_x64_compiler.cpp
```diff
@@ -89,3 +89,44 @@ TEST_CASE("EX2", "[video_core][shader][shader_jit]") {
...
+        {OpCode::Id::ADD, sh_temp, sh_temp, "", sh_temp},
+        {OpCode::Id::MOV, sh_output, sh_temp},
+        {OpCode::Id::END},
+        // clang-format on
+    });
+
+    REQUIRE(shader.Run(2.f) == Approx(4.f));
+    REQUIRE(shader.Run(20.f) == Approx(40.f));
+    REQUIRE(shader.Run(7.f) == Approx(14.f));
+}
```
Hexagon12: Needs a newline.

yuriks:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 321 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3270#-L03NGrAWQ-Fz0LdQp2E:-L03NGrAWQ-Fz0LdQp2F:b-tspcti) ([raw file](https://github.com/citra-emu/citra/blob/8b71af3482ec8accd97a860dbb35b619015abdc9/src/video_core/shader/shader_jit_x64_compiler.cpp#L321)):*
> ```cpp
>         // Not all components are enabled, so mask the result when storing to the destination
>         // register...
>         movaps(SCRATCH, *dest_loc);
> ```

For the SSE4.1 case with blendps, you can directly use the cached register as the destination for the blenps, avoiding any moves whatsoever:

blendps(*dest_loc, src, mask);

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 326 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3270#-L03NkY34TCgxv0BiyUb:-L03NkY4CAzGOCU8cRgo:b-yzi0j5) ([raw file](https://github.com/citra-emu/citra/blob/8b71af3482ec8accd97a860dbb35b619015abdc9/src/video_core/shader/shader_jit_x64_compiler.cpp#L326)):*
> ```cpp
>             u8 mask = ((swiz.dest_mask & 1) << 3) | ((swiz.dest_mask & 8) >> 3) |
>                       ((swiz.dest_mask & 2) << 1) | ((swiz.dest_mask & 4) >> 1);
>             blendps(SCRATCH, src, mask);
> ```

Looks like blendps supports a memory operand for the source, so you could also skip the movaps above in the memory case by inverting the mask and doing this:

blendps(src, *dest_loc, ~mask);
movaps(*dest_loc, src);

---

*[src/video_core/shader/shader_jit_x64_compiler.cpp, line 338 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3270#-L03RaMvnQBDiA1Wkgwx:-L03RaMvnQBDiA1Wkgwy:bv75z3k) ([raw file](https://github.com/citra-emu/citra/blob/8b71af3482ec8accd97a860dbb35b619015abdc9/src/video_core/shader/shader_jit_x64_compiler.cpp#L338)):*
> ```cpp
>                      ((swiz.DestComponentEnabled(2) ? 0 : 1) << 4) |
>                      ((swiz.DestComponentEnabled(3) ? 2 : 3) << 6);
>             shufps(SCRATCH, SCRATCH2, sel);
> ```

And in the above you can replace all uses of `SCRATCH` by `*dest_loc` for the register case to skip the move at the end.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3270)*

Subv:
I'll close this as it doesn't actually seem to provide any improvement for the special cases it handles.

PR #3272, opened 2017-12-11
core/arm: Backend-specific context implementations

MerryMage:
~~WIP as dynarmic side changes are in flux.~~

Advantage is this avoids needing to invalidate cache-like structures on the CPU emulator like the Return Stack Buffer when doing a context-switch. This also avoids the cost of seralizing and deseralizing the CPSR and FPSCR when the CPU backend has a non-canonical representation of these things.

On dynarmic's side we also have the following changes for speed:

* Split CPSR for slightly faster GE flags access and slightly faster runtime location-hash calculation.
* Added prediction for where the next basic block would be when exiting due to exhaustion of time slice to avoid an expensive std::unordered_map lookup on re-entry.
* Turn RSB back into a stack since it offers better predictions with the context changes.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3272)

### DIFF
--- a/src/core/arm/arm_interface.h
+++ b/src/core/arm/arm_interface.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <cstddef>
+#include <memory>
 #include "common/common_types.h"
 #include "core/arm/skyeye_common/arm_regformat.h"
 #include "core/arm/skyeye_common/vfp/asm_vfp.h"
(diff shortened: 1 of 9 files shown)


src/core/arm/dynarmic/arm_dynarmic.cpp
```diff
@@ -148,30 +202,24 @@ void ARM_Dynarmic::SetCP15Register(CP15Register reg, u32 value) {
...
+std::unique_ptr<ARM_Interface::ThreadContext> ARM_Dynarmic::NewContext() {
+    return std::make_unique<DynarmicThreadContext>();
+}
 
-    ctx.sp = jit->Regs()[13];
-    ctx.lr = jit->Regs()[14];
-    ctx.pc = jit->Regs()[15];
-    ctx.cpsr = jit->Cpsr();
+void ARM_Dynarmic::SaveContext(const std::unique_ptr<ThreadContext>& arg) {
+    DynarmicThreadContext* ctx = dynamic_cast<DynarmicThreadContext*>(arg.get());
```
Subv: Do we build with RTTI?

src/core/arm/arm_interface.h
```diff
@@ -15,14 +16,41 @@ class ARM_Interface : NonCopyable {
     virtual ~ARM_Interface() {}
 
     struct ThreadContext {
```
lioncash: This should likely be turned into a `class` if its interface is becoming pure virtual.

src/core/arm/arm_interface.h
```diff
@@ -124,17 +152,23 @@ class ARM_Interface : NonCopyable {
      */
     virtual void SetCP15Register(CP15Register reg, u32 value) = 0;
 
+    /**
+     * Creates a CPU context
+     * @param ctx Thread context to save
```
lioncash: This doesn't have a context parameter

src/core/arm/arm_interface.h
```diff
@@ -124,17 +152,23 @@ class ARM_Interface : NonCopyable {
...
+     */
+    virtual std::unique_ptr<ThreadContext> NewContext() = 0;
```
lioncash: This can likely be const qualified as well, no? Given its task is to create a new context, it'd be kind of questionable to modify the current instance itself directly.

src/core/arm/dynarmic/arm_dynarmic.cpp
wwylele: @Subv #1238

MerryMage: I guess since this is casting to a known child cast it could be a static_cast instead, though getting that wrong is UB.

PR #3273, opened 2017-12-11
Announce room webservice

B3n30:
Moved AnnounceMultiplayerSession from core to WebService, so the namespace and the include path changed.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3273)


B3n30:
Close, since this should be a PR at jroweboy's fork

PR #3280, opened 2017-12-12
[RFC] Assert: Don't crash citra completely on assert, but stop the emu_thread

B3n30:
This will allow to do things like log storing from within the ui when citra crashes.
The exception should only be thrown from within emu_thread.
Since the exception isn't handled in JIT we have to circumvent it.
Future work will include:

- [ ] Add separate macros for this type of assert (Something like EMU_ASSERT)
- [ ] Close the renderer window
- [ ] Show an error dialog on that error 
- [ ] Add the stack trace to the log

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3280)

### DIFF
--- a/src/common/assert.h
+++ b/src/common/assert.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <cstdlib>
+#include <exception>
 #include "common/common_funcs.h"
 #include "common/logging/log.h"
 
@@ -15,6 +16,9 @@
 // template that calls the lambda. This seems to generate an extra instruction at the call-site
 // compared to the ideal implementation (which wouldn't support ASSERT_MSG parameters), but is good
 // enough for our purposes.
+
+class AssertException : std::exception {};
+
 template <typename Fn>
 #if defined(_MSC_VER)
 __declspec(noinline, noreturn)
@@ -23,8 +27,7 @@ __declspec(noinline, noreturn)
 #endif
     static void assert_noinline_call(const Fn& fn) {
     fn();
-    Crash();
-    exit(1); // Keeps GCC's mouth shut about this actually returning
+    throw AssertException();
 }
 
 #define ASSERT(_a_)                                                                                \
--- a/src/core/arm/dynarmic/arm_dynarmic.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <csetjmp>
 #include <cstring>
 #include <dynarmic/dynarmic.h>
 #include "common/assert.h"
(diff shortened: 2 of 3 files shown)


MerryMage:
I would probably prefer the `try` block to wrap the entire `while (!stop_run)` loop.

src/core/arm/dynarmic/arm_dynarmic.cpp
```diff
@@ -14,6 +15,27 @@
 #include "core/hle/kernel/svc.h"
 #include "core/memory.h"
 
+jmp_buf jit_exit_point;
```
lioncash: These should likely not be file scope, but be instance based. This would be problematic if we ever get around to running both core instances, since it would allow one core to trash the others' data.

src/core/arm/dynarmic/arm_dynarmic.cpp
```diff
@@ -82,8 +104,12 @@ MICROPROFILE_DEFINE(ARM_Jit, "ARM JIT", "ARM JIT", MP_RGB(255, 64, 64));
 void ARM_Dynarmic::Run() {
     ASSERT(Memory::GetCurrentPageTable() == current_page_table);
     MICROPROFILE_SCOPE(ARM_Jit);
-
-    jit->Run(GetTicksRemaining());
+    int asserted = setjmp(jit_exit_point);
```
lioncash: We should avoid using `setjmp` and `longjmp` here if at all possible (aside from the fact it's [discouraged](https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=88046492) in multiple safety enforcement manuals), but also because if a longjmp occurs in any code path where non-trivial objects would have otherwise had their destructor called if the `setjmp` and `longjmp` were replaced with `catch` and `throw`, then it invokes undefined behavior.

From C++17 `[csetjmp.syn] ¶ 2"

> The function signature `longjmp(jmp_buf jbuf, int val)` has more restricted behavior in this International Standard. A `setjmp`/`longjmp` call pair has undefined behavior if replacing the `setjmp` and `longjmp` by `catch` and `throw` would invoke any non-trivial destructors for any automatic objects.

It also strikes me as somewhat extremely odd to use *both* setjmp/longjmp and exceptions in the same context.

src/common/assert.h
```diff
@@ -15,6 +16,9 @@
 // template that calls the lambda. This seems to generate an extra instruction at the call-site
 // compared to the ideal implementation (which wouldn't support ASSERT_MSG parameters), but is good
 // enough for our purposes.
+
+class AssertException : std::exception {};
```
lioncash: This should likely inherit from `std::runtime_error` or `std::logic_error` rather than the top-level exception class.

B3n30: I think it's neither. In my first design I was just throwing a std::exception, since  the assert can be any exception. But I wanted to just catch exceptions thrown by the assert, thus the class. But since the assert can be any kind of error, I don't think it should be inherited from another sub class

src/core/arm/dynarmic/arm_dynarmic.cpp
B3n30: The main goal is to use exceptions instead of just crashing. But the jit part don't handle stack unwinding. So setjmp/longjmp is the only way to circumvent the jit (besides hadnling stack unwinding in jit). But your argument is the reason setjmp/longjmp are as close as possible to the jit entry and exit points. This inside jit there shouldn't be any ctor/dtor

src/common/assert.h
lioncash: An assert is inherently when a precondition or requirement has failed (which would be a logic error).

From C++17 `[logic.error] ¶ 1`

> The class `logic_error` defines the type of objects thrown as exceptions to report errors presumably detectable before the program executes, such as violations of logical preconditions or class invariants.

src/core/arm/dynarmic/arm_dynarmic.cpp
MerryMage: > This inside jit there shouldn't be any ctor/dtor

I wouldn't rely on this. I'd wait for non-windows unwinding support.

B3n30:
I'll close this for now until we have stack unwinding in our JIT (which is something I'm not capable of implementing). Overall, I think the feedback that exception handling instead of crashing on assert is a good idea, so I will reopen that as soon as our JIT is prepared for it

Issue #3260, opened 2017-12-10
Saving twice in XY causes the game to crash.


LogPod:
Oh wow another XY issue
tested on Head-e8209ea(latest canary)
[log.txt](https://github.com/citra-emu/citra/files/1546108/log.txt)

legoj15:
Could you also test on latest nightly for posterity? Canary builds can have some funky side effects.

LogPod:
Head-2146311 Latest Nightly
[Nightly log.txt](https://github.com/citra-emu/citra/files/1546135/Nightly.log.txt)
Same error @legoj15

LogPod:
Or open-gl errors are not causes by any programs. i have literally nothing open so i don't why it gives those errors.

CDAGaming:
Tested on Latest Canary, Master, and 'Sandbox':

Stacktrace (From MerryMage's crash-trace PR):
https://hastebin.com/ojocumolez.scala

CDAGaming:
@LogPod I've only seen those OpenGL errors on PC's using Integrated Graphics that don't support the proper version for OpenGL.

LogPod:
My Graphics card supports up to open gl 4.4 so that isn't the problem

legoj15:
The log mentions a NVIDA Quadro card, but it's running on Qudro driver version 377.61. The latest one I have is Quadro 386.01. Is there a way that you could try to update your drivers, just incase?

LogPod:
This is the latest one for me 
![capture](https://user-images.githubusercontent.com/30522087/33828788-41a865d6-de8f-11e7-8b5d-148217d48795.PNG)

MerryMage:
Hi guys. Thanks for the enthusiasm. From the log it looks like an assertion in the texture forwarding code, hence my label.

Either cached_res_count is too small or there's a logic error somewhere else.

LogPod:
Seems like this is fixed in the latest canary.

LogPod:
Issues are closed when they are fixed in nightly builds @adityaruplaha

adityaruplaha:
Oh, I see. Sorry, I didnt know that.

wwylele:
Fixed by #3281 as reported so closing. Feel free to reopen if it is not the case.

Issue #3102, opened 2017-11-10
Phoenix Wright -  Dual Destinies Both Screen Malfunction


Dragios:
I have found that PR #3048 cause this regression. This is kind of weird, not sure how this PR can cause such bugs. But I bisected it till here. The upper display no longer works as it should anymore for some reason. It just show that static image.
![glitch](https://user-images.githubusercontent.com/18496868/32637405-8f64388a-c5f4-11e7-95f9-999c0d079ff5.png)

Switching renderer from HW to SW just shows black screen for both screens.
Switching it back to HW again results the same as SW now.

When it black screen it will show this log:
`Render.OpenGL <Error> video_core/renderer_opengl/renderer_opengl.cpp:DebugHandler:472: API ERROR 1282: GL_INVALID_OPERATION error generated. Texture name does not refer to a texture object generated by OpenGL.`

Full log [here](https://pastebin.com/fbqrsXax)

**Separate issue:**
There is another funky thing going horibly wrong on the bottom screen (also introduced by this PR).
Only triggers by auto internal resolution setting. Can be explain by viewing this [video](https://streamable.com/j9qia).

---
**Update:** The glitch will be gone if you have a save file to begin with.

Hexagon12:
The latest Nightly doesn't produce the issue.
![citra-qt_2017-11-13_12-49-23](https://user-images.githubusercontent.com/25362478/32722020-32b8f042-c871-11e7-834b-045045c60034.png)

If this issue is still present for you, it means that your GPU drivers produce artifacts with Citra and I would suggest either updating to the latest or previous drivers.

Dragios:
I know why @Hexagon12 does not have the issue. It's because he have a save file to begin with. This glitch will only occur if you start a **New Game**

Dragios:
Fixed by #3307

Issue #3055, opened 2017-10-27
Poochy & Yoshi's Wooly World - Crash with Intel GPU


valentinvanelslande:
Commit: edfc8d1
Start the game, configure the game and on the yoshis scene citra crashes.
Specs: i5-7200U, Intel HD Graphics 620, Windows 10 15063
[Log](https://raw.githubusercontent.com/acnleditor2/citra-logs/master/Poochy%20%26%20Yoshi's%20Wooly%20World.txt)

acnleditor2:
workaround: disable hw renderer in level selection screen.

RavenHome1:
what is up with this massive regression, most of these games worked greatly for so long

Hexagon12:
People on the Discord reported that this game worked before just fine, so this issue is a regression.

Could you please find out on what version it first started to crash?

Schplee:
In Nightly - `79852d3`, the game doesn't crash. Followed your steps: Created new save, configured the settings, and successfully finished the intro cut scene and started a level.

wwylele:
System: Ubuntu
Commit: 79852d3 (compiled myself using GCC 7.2)
Game: Poochy & Yoshi's Wooly World (JAP) (00040000 001a3200)
Result: Failed to reproduce

More specifically, I can go in game with hw renderer without any issue. In the log, I don't get those craze log entries that indicates broken gpu command list such as
 - `Invalid VS program offset 512`, 
 - `Unhandled instruction: 0x20 (0x80000000)`,
 - `Unknown source op 8`, 
 - `Commandlist tried to write to invalid register`

Here are some possibilities I came up with:
 - Your ROM is somehow broken, or at least we are not using the same version of the game
 - A log entry that can be seen in both of your and my log: `Emulated program broke execution! Break reason: PANIC`, which might be caused by unimplemented cecd functions, probably leads the program execution into undefined state and results in inconsistent behaviour

acnleditor2:
using jpn version. log updated.

acnleditor2:
@wwylele no broken graphics in my fork... and with your cec branch not fixed

acnleditor2:
@Schplee crash.

acnleditor2:
in eur version broken graphics on all citras

Jhon591:
You're braking your own build by adding to much at once.

wwylele:
> and with your cec branch not fixed

Just for others seeing this, my cec branch is still WIP and I don't expect it to fix anything

[...]

Issue #2741, opened 2017-06-01
Phoenix Wright Ace Attorney: Spirit of Justice – Flashing Graphical Glitches on Hardware Renderer


Dragios:
This only happens on the integrated __**Intel**__ GPU and I have no idea why. This issue is not present when using the dedicated __**nVidia**__ GPU. Tested on latest master based on commit 4857eb441bbe7ba1637838c9e5158f0801f40730
Altering *Shader JIT* on/off yield the same result.
Partial mirroring from top screen shown on bottom screen indicates a separate issue.

*Hardware renderer* with [full log](https://pastebin.com/K9YzPTz3) attached:
![Flashing Graphical Glitch](https://cloud.githubusercontent.com/assets/18496868/26680851/8a031318-470d-11e7-8ce8-aaedc7711a1d.png)

*Software renderer*:
![Software Renderer](https://cloud.githubusercontent.com/assets/18496868/26681064/57486aa8-470e-11e7-8517-759f06e77391.png)

Log spams with this in *Software Renderer*:
`Render.OpenGL <Error> video_core\renderer_opengl\renderer_opengl.cpp:DebugHandler:472: API ERROR 1282: Error has been generated. GL error GL_INVALID_OPERATION in (null): (ID: 173538523) Generic error`

I have include a [video](https://streamable.com/6f47m) for a better picture.

RavenHome1:
there is also some glitchy effects right as the game boots from the video you provided

Dragios:
@RavenHome1 That one is not related and has been explained [here](https://github.com/citra-emu/citra/issues/2715#issuecomment-302906860)

Dragios:
@RavenHome1 Well it could be this as explained [here](https://github.com/citra-emu/citra/issues/2715#issuecomment-302906860) or not.

yuriks:
Seems like some state tracking bug from us. We have a few of these that show up pretty often depending on what you're playing...

@Dragios that issue isn't the same as this one. That one is only about the glitched video showing for a frame at boot.

wwylele:
Fixed by #3281

PR #3281, opened 2017-12-12
Texture Cache Rework

jroweboy:
(Please note this code was written by @phanto-m and not me. I've studied it extensively so I can be ready to answer questions in review, but I am still noob and may not know everything/got something wrong. Please bear with me)

### Goal:

Eliminate as many superfluous CPU <-> GPU flushes as possible. This is primarily accomplished by adding an Invalidate (without flush) method on the texture cache that is preferred to use instead of the InvalidateAndFlush that was in use everywhere before. This also adds support for hw texture copy which fixes upscaling in many games and makes several others full speed (and removes the need to use the CPU implementation.)

### Whats new:

* The entire texture cache interface got a face lift. Within the texture cache, lookup for surfaces is now done with easier to use helper methods to find the most appropriate surface. The parameters used to find/create a surface is now split from the actual cached surface as well, which should make it a bit easier to get a surface from the cache now.
* Optimized morton copy to copy in tiles
* Resolution scaling is now locked to integer sizes
* GL_SCISSOR and viewport tests to try to avoid drawing out of the draw area if not needed
* Invalidate without flushing which will attempt to revalidate the surface by copying from compatible surfaces if it can.
* HW texture copy (yay!)

### General Notes:

* The coordinate system in the texture cache was changed to match GL_ORIGIN instead of the PICA coordinate system. I didn't bother to split this into its own commit. Please be aware of this when reviewing. Heres what phantom said when i asked them about it: "getsurfacerect would return top-left origin coords when dealing with nontiled textures which made was a real PITA, now it's consistent and matches gl's origin. the flipping is now done in displaytransfer since it's the only place where that conversion is required. i kept the kinda broken texcoords in acceleratedisplay, it's weird how it expects rotated coors or i just didnt understand how the rotating is done in the rendered, didnt bother looking too much into it"
* The final commit is a hacky workaround that I'm going to leave up the reviewers to decide what to do with it. Its put at the end there so we can drop it if we decide its too much. It works by ignoring any regions that are dirty when revalidating (and being in the dirty region implies that the region was created only on the GPU), and if every place we are trying to validate in the interval is either copyable from another surface, or in the dirty regions, then it gets ignored. This makes things like the after match screen in smash 4 much faster with no visible differences, and also removes pokemon sun/moon outlines as well.

I did my best to split all the changes into separate commits, so the reviewers can just read them one at a time, but as they can probably tell, I got a little tired of doing this towards the end. "Add the rest of the cache methods" "Update the rasterizer to use the new cache" are probably the most "[draw the rest of the owl](http://i0.kym-cdn.com/photos/images/facebook/000/572/078/d6d.jpg)" that this pr gets. I tried to at least make the diffs line up so they are easier to read :)

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3281)


valentinvanelslande:
miitopia very long loading time

Subv:
I see a memory write will now only Invalidate the region in the GPU instead of Flushing + Invalidating.

What happens if the CPU does a Write, and then a Read? The comment in InvalidateRegion says that it removes the surface from the cache, if this is the case then how will it be flushed to memory during the Read?

Subv:
Reviewed 9 of 9 files at r1, 8 of 8 files at r2.
Review status: 16 of 24 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/core/hw/gpu.cpp, line 360 at r2](https://reviewable.io:443/reviews/citra-emu/citra/3281#-L0AwmyvsAy8gs4ThJPy:-L0AwmywDeCZOy5Ae8sv:bwcb0np) ([raw file](https://github.com/citra-emu/citra/blob/0396f7417e3fed5ed7247e7fb2f5871d8cf5f64e/src/core/hw/gpu.cpp#L360)):*
> ```cpp
>     const auto FlushInvalidate_fn = (output_gap != 0) ? Memory::RasterizerFlushAndInvalidateRegion
>                                                       : Memory::RasterizerInvalidateRegion;
>     FlushInvalidate_fn(config.GetPhysicalOutputAddress(), static_cast<u32>(contiguous_output_size));
> ```

I'd much prefer that you use a normal if for this

---

*[src/video_core/renderer_opengl/gl_rasterizer_cache.cpp, line 528 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3281#-L0AvRJBgSA-S9XzJWEA:-L0AvRJBgSA-S9XzJWEB:b3kre) ([raw file](https://github.com/citra-emu/citra/blob/bc204dce5529bea0af5b835e8a8c7b7e0d8c927d/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp#L528)):*
> ```cpp
> }
> 
> // If the resolution
> ```

?

---

*[src/video_core/renderer_opengl/gl_rasterizer_cache.cpp, line 530 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3281#-L0AvT0mvUTRLbL8U7rn:-L0AvT0mvUTRLbL8U7ro:b-loic1a) ([raw file](https://github.com/citra-emu/citra/blob/bc204dce5529bea0af5b835e8a8c7b7e0d8c927d/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp#L530)):*
> ```cpp
> // If the resolution
> static u16 GetResolutionScaleFactor() {
>     return !Settings::values.resolution_factor
> ```

Settings::values.resolution_factor == 0

---

*[src/video_core/renderer_opengl/gl_rasterizer_cache.cpp, line 543 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3281#-L0AvbwRVsAJgIL3_etb:-L0AvbwS-KDrlp8MZLSe:b-jwy2yk) ([raw file](https://github.com/citra-emu/citra/blob/bc204dce5529bea0af5b835e8a8c7b7e0d8c927d/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp#L543)):*
> ```cpp
>     // update resolution_scale_factor and reset cache if changed
>     static u16 resolution_scale_factor = GetResolutionScaleFactor();
>     if (resolution_scale_factor != GetResolutionScaleFactor()) {
> ```

this is never true?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3281)*

valentinvanelslande:
Fixes #3055 and #2938

jroweboy:
Review status: 16 of 24 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/video_core/renderer_opengl/gl_rasterizer_cache.cpp, line 528 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3281#-L0AvRJBgSA-S9XzJWEA:-L0B1QgTSN5lUbIfbHUX:b-t5p25a) ([raw file](https://github.com/citra-emu/citra/blob/bc204dce5529bea0af5b835e8a8c7b7e0d8c927d/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp#L528)):*

?
</blockquote></details>

Oops :D

---

*[src/video_core/renderer_opengl/gl_rasterizer_cache.cpp, line 543 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3281#-L0AvbwRVsAJgIL3_etb:-L0B0mqCGGHAbcUDG--7:brmiymw) ([raw file](https://github.com/citra-emu/citra/blob/bc204dce5529bea0af5b835e8a8c7b7e0d8c927d/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp#L543)):*

this is never true?
</blockquote></details>

resolution_scale_factor is a static variable, meaning it'll get set to the current scale factor only on the first initialization.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3281)*

LexLight:
In "Paper Mario: Sticker Star", most of the low screen is white during the title, file select screens, and ingame, too, making the stickers unseeable.
https://cdn.discordapp.com/attachments/242442830486700045/390224478312988672/unknown.png
imagen
In "Professor Layton and the Miracles Mask", in Chapter 1, after they read the letter, the background becomes black, through after the scene, the background is fine.
imagen
I will test more if I can.

Hexagon12:
Oh boy.
Fixes #3260, #2846, #2640, #2505, #2375, #2262, #2229, #2220 (the outlines are fully removed), #1830, #1521.

However, reintroduces #1682 and #2110.

New bugs: 

- Virtual Console games have no display at all.
- Ultimate NES Remix now doesn't boot.

B3n30:
Also surprisingly fixes #2416

Subv:
How does this fix  #2262, a crash in the software renderer?

wwylele:
Just want to emphasizers a principle here:

Removing outlines in Pokemon games would be considered as **a bug**, not a feature.

jroweboy:
@wwylele Yes, I apologize that I didn't make it clear that it was a bug. There is a follow PR that does d24s8 to rgba8 conversion with a PBO that I was debating on including or not with this, but I felt like it should be in a separate pull request entirely so I didn't add it. That code will be run before the skip dirty regions check, so they are compatible and with it outlines will be drawn.

ds84182:
The first screen transition that Art of Balance TOUCH! does is broken with this PR. Instead of having a copy of the framebuffer it just shows black. Subsequent screen transitions work properly.

This can be reproduced by opening the game, selecting a save file, then clicking either "Arcade", "Endurance", or "Credits". Going back and selecting those options a second time shows the transition properly though.

jroweboy:
Review status: 16 of 24 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/common/math_util.h, line 32 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3281#-L0B5YTy-D7qHrJCHAp3:-L0BW_eov7WN73zRNPlz:bj5in0o) ([raw file](https://github.com/citra-emu/citra/blob/bc204dce5529bea0af5b835e8a8c7b7e0d8c927d/src/common/math_util.h#L32)):*

unrelated (but fine)
</blockquote></details>

It had to go somewhere (and I didn't think it was anything special enough to get into a different PR)

---

*[src/core/frontend/framebuffer_layout.cpp, line 20 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3281#-L0B6SwId-BoWf5-8cz1:-L0BWGTMOvPZUQh6E9nh:byc6fcl) ([raw file](https://github.com/citra-emu/citra/blob/bc204dce5529bea0af5b835e8a8c7b7e0d8c927d/src/core/frontend/framebuffer_layout.cpp#L20)):*

Can the width get 0? If so this will fail.
</blockquote></details>

It might be possible in some weird case when the game is just booting, so I probably should handle that. But we do have a minimum size for the window so it doesn't happen in practice right now.

---

*[src/video_core/renderer_opengl/gl_shader_gen.cpp, line 44 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3281#-L0B7GfBAoNoSc5C0hxh:-L0BW19u6Ak3XcArl5fb:bk484ru) ([raw file](https://github.com/citra-emu/citra/blob/bc204dce5529bea0af5b835e8a8c7b7e0d8c927d/src/video_core/renderer_opengl/gl_shader_gen.cpp#L44)):*

Should probably belong to another commit?
</blockquote></details>

no, this is part of changing the resolution scale from a float to an int. originally it was actually two floats, one for horizontal and one for vertical hence the vec2, but now its just a constant int.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3281)*

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -159,641 +286,1182 @@ void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
-        return;
-    }
+    BlitSurfaces(src_surface, src_surface->GetScaledRect(), dest_surface,
+                 dest_surface->GetScaledSubRect(*src_surface));
 
-    MICROPROFILE_SCOPE(OpenGL_SurfaceDownload);
+    dest_surface->invalid_regions -= src_surface->GetInterval();
+    dest_surface->invalid_regions += src_surface->invalid_regions;
 
-    u8* dst_bu
```
Subv: Please add a more descriptive comment here, this is what phantom said about it: `true, that comment doesn't tell much. the point is to avoid thousands of small writes every frame if the cpu decides to access that region, anything higher than 8 you're guaranteed it comes from a service (dma, failed texcopy ...)
`

AuraOfTheDawn:
Breaks color on enemy hp meteres in Puzzle And Dragons Mario Bros Edition. Previously enemy hp meters were colored after their element, now all hp meters are black (and empty to grey)

Dragios:
There are a few things I wanna share.

What @Hexagon12 said was partially correct regarding *Virtual Console*. It does not display anything, true. But if you upscale, it will show the previous image buffer. Same goes with downscaling back. It just shows nothing initially so I assume the texture is not forwarding properly. Same issue occurs with **Tales of the Abyss.**
Pokémon Yellow

Kind of confused that this fixes #2262 as the crash involve purely software renderer. Mind explain how you get the result?

---
Most of the bugs I found have been mentioned already so I'm kind of late to the party. I just highlight the new bugs I discovered.

- I think **Cooking Mama 5 - Bon Appetit!** has similar issue as the Virtual Console but instead it shows  turquoise screen instead initially. Once I upscale it will show the previous image buffer and are the same for both screen. This only occurs when there is no save file like #2235 but at a much worse state.
Cooking Mama 5
Looks similar to #3102 before this PR

- **Metal Gear Solid 3D** have a few spot of flickering when on native resolution and occurs on _nVidia_ GPU only. _Intel_ GPU is unaffected. Not so sure with _AMD_ GPU.

Fixes #2741 & #3102 

I notice this log appears in some games on _Intel_ GPU. Will occur on _nVidia_ GPU if I quit the game during the right moment.
```
Render.OpenGL <Warning> video_core/renderer_opengl/renderer_opengl.cpp:DebugHandler:472:
API PERFORMANCE 131154: Pixel-path performance warning: Pixel transfer is synchronized with
3D rendering.
```

src/core/memory.cpp
```diff
@@ -419,6 +409,12 @@ void RasterizerFlushRegion(PAddr start, u32 size) {
     }
 }
 
+void RasterizerInvalidateRegion(PAddr start, u32 size) {
+    if (VideoCore::g_renderer != nullptr) {
```
lioncash: This can be inverted, which is usually what's done (early out on the error case).

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -893,227 +947,143 @@ void RasterizerOpenGL::FlushAll() {
...
+    SurfaceParams dst_params = *src_surface;
+    dst_params.addr = config.GetPhysicalOutputAddress();
+    dst_params.width = src_rect.GetWidth() / src_surface->res_scale;
+    dst_params.stride = dst_params.width + src_surface->PixelsInBytes(
+                                               src_surface->is_tiled ? output_gap / 8 : output_gap);
+    dst_params.height = src_rect.GetHeight() / src_surface->res_scale;
+    dst_params.res_scale = src_surface->res_scale;
+    dst_params.UpdateParams();
+
+    const bool load_gap = output_gap != 0; // Since we are going to invalidate the gap if there is
```
lioncash: This comment should likely go above the variable, not to the side of it.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -25,103 +31,224 @@
 #include "video_core/utils.h"
 #include "video_core/video_core.h"
 
+using SurfaceType = SurfaceParams::SurfaceType;
+using PixelFormat = SurfaceParams::PixelFormat;
+
+static std::array<OGLFramebuffer, 2> transfer_framebuffers;
```
lioncash: ugh @ non-const static state

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -25,103 +31,224 @@
...
+        std::array<u8, tile_size> tmp_buf;
+        MortonCopyTile<morton_to_gl, format>(stride, &tmp_buf[0], gl_buffer);
+        std::memcpy(tile_buffer, &tmp_buf[start - aligned_down_start],
+                    std::min(aligned_start, end) - start);
+
+        tile_buffer += aligned_start - start;
+        glbuf_next_tile();
+    }
+
+    u8* const buffer_end = tile_buffer + aligned_end - aligned_start;
```
lioncash: This can also be a pointer to const `const u8*` as well as a const pointer (i.e. `const u8* const`)

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -159,641 +286,1182 @@ void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
+void CachedSurface::LoadGLBuffer(PAddr load_start, PAddr load_end) {
+    ASSERT(type != SurfaceType::Fill);
+
+    const u8* const texture_src_data = Memory::GetPhysicalPointer(addr);
+    if (texture_src_data == nullptr)
+        return;
+
+    if (gl_buffer == nullptr) {
+        gl_buffer_size = width * height * GetGLBytesPerPixel(pixel_format);
+        gl_buffer.reset(new u8[gl_buffer_size]);
```
lioncash:
```cpp
gl_buffer = std::make_unique<u8[]>(gl_buffer_size);
```

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -159,641 +286,1182 @@ void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
+void CachedSurface::DownloadGLTexture(const MathUtil::Rectangle<u32>& rect) {
+    if (type == SurfaceType::Fill)
+        return;
 
-                MortonCopyPixels(params.pixel_format, params.width, params.height, bytes_per_pixel,
-                                 gl_bytes_per_pixel, texture_src_data, temp_fb_depth_buffer_ptr,
-                                 true);
+    if (gl_buffer == nullptr) {
+        gl_buffer_size = width * height * GetGLBytesPerPixel(pixel_format);
+        gl_buffer.reset(new u8[gl_buffer_size]);
```
lioncash:
```cpp
gl_buffer = std::make_unique<u8[]>(gl_buffer_size);
```

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
lioncash: `while (true)`

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -162,78 +183,180 @@ struct CachedSurface {
...
+    MathUtil::Rectangle<u32> GetRect() const {
+        return {0, height, width, 0};
+    }
+
+    MathUtil::Rectangle<u32> GetScaledRect() const {
+        return {0, GetScaledHeight(), GetScaledWidth(), 0};
+    }
+
+    u32 PixelsInBytes(u32 size) const {
+        return size * 8 / GetFormatBpp(pixel_format);
```
lioncash: `size * CHAR_BIT`

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -162,78 +183,180 @@ struct CachedSurface {
...
+    }
+
+    u32 BytesInPixels(u32 pixels) const {
+        return pixels * GetFormatBpp(pixel_format) / 8;
```
lioncash: `/ CHAR_BIT`

src/video_core/utils.h
```diff
@@ -9,9 +9,9 @@
 namespace VideoCore {
 
 // 8x8 Z-Order coordinate from 2D coordinates
-static inline u32 MortonInterleave(u32 x, u32 y) {
-    static const u32 xlut[] = {0x00, 0x01, 0x04, 0x05, 0x10, 0x11, 0x14, 0x15};
-    static const u32 ylut[] = {0x00, 0x02, 0x08, 0x0a, 0x20, 0x22, 0x28, 0x2a};
+static constexpr u32 MortonInterleave(u32 x, u32 y) {
+    constexpr u32 xlut[] = {0x00, 0x01, 0x04, 0x05, 0x10, 0x11, 0x14, 0x15};
```
lioncash: From Discord:

> Does MortonInterleave's arrays being non-static constexpr in that PR really improve anything? (as in, if the compiler optimizes all its calls despite the loop its in can't be fully compile-time optimized away).
>
> I know GCC will always load those array values onto the stack every time now if it can't run the expression at compile-time (clang does better in this regard and would still store the values fixed without the static keyword).

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
Phanto-m: You don't want to pay extra for default initialization here, which was the point of using unique_ptr in the first place

lioncash: Ah, fair point.

jroweboy:
Okay, I removed the re interpretation hack since there seems to be more problems with it than originally noticed. With that gone, please retest the games and see if they still have the same errors.

I also (hopefully) fixed compilation for linux and mac on the build bots.

emmauss:
Comparing Canary 253(previous build)  and 256(first texture cache build), using Pokemon Sun,
Area tested is the Pokecenter on route 8
253
Integrated Intel HD 4400 : Average Speed = 55%
Nvidia Geforce 840m: Average Speed = 43% 

256
Integrated Intel HD 4400 : Average Speed = 47%
Nvidia Geforce 840m: Average Speed = 60% 

Looks like caching is faster on discrete than integrated.

jroweboy:
@Phanto-m added in some more checks for accelerated texture copy which should fix pokemon super mystery dungeon and also fixed a mistake i made when removing the reinterpretation hack

Dragios:
Ok, here is the latest result based on my testing.
- Virtual Console's display works as it should now
- Pokémon outline is still missing
- Pokémon Mystery Dungeon series no longer crashes
- Metal Gear Solid 3D's few spot flickering still exist (nothing changed)
- Cooking Mama 5's display issue remains as shown [here](https://github.com/citra-emu/citra/pull/3281#issuecomment-351277925) (nothing changed)
- Performance warning log remains

jroweboy:
Rebased off master to fix a conflict so it can go back on canary. Thanks for all the testing so far, I will try to address everyones feedback when I get time.

tywald:
A portion of the top screen(on the far right) isn't displayed correctly in Dragon Quest VII and Monster Hunter 4 Ultimate. In addition, the bottom screen is mirrored on the top screen in Dragon Quest Monsters: Terry's Wonderland 3D. Introduced in the last two commits by @Phanto-m.

Tested on the appveyor build, both msvc and mingw.

Hexagon12:
Fire Emblem Echoes: Shadows of Valentia is also broken by those two Phanto-m commits.
citra-qt_2017-12-14_17-34-52

jroweboy:
@tywald @Hexagon12 It seems that I was accidentally removing the wrong interval when validating a surface. This is fixed now, which should fix those few regressions from the last few commits.

Hexagon12:
Can confirm that the graphical bugs are gone with Shadows of Valentia and Monster Hunter Stories (from @tywald's picture).

LexLight:
In lastest Canary, in "Paper Mario: Sticker Star" part of the right part of the low screen is white and unseeable:
imagen
In the map, there's also a black bold line, through it dissappears when you go into a phase.
imagen

As for "Professor Layton and the Miracle Mask", now the background of the Chapter 1 when they finish reading the letter displays like normal.
imagen

ds84182:
On the latest Canary build the first transition almost works, but the top screen is cropped. Subsequent transitions work still. 
aob

jroweboy:
Found an issue which could be the cause of the cropping bug in those games. Please retest when you get a chance :)

urdaneta09:
NICE! in Kingdom Hearts DDD the motionblur effect are broken even in the first version of this PR
Now with the last update is fixed!
tested in MinGW-appveyor
imagen1

Other thing is the big regression in speed after of Canary build #256 (first version of this pr) 20-30% less only this game, others only 2%-5% less, after of the "Revert OpenGL Cache: Ignore format reinterpretation hack" updates.

jroweboy:
@urdaneta09 speed regressions in some games compared to earlier versions of this PR is expected since there was a hack that I was testing to see if it broke any games. It made some games faster with no issues, but it completely broke some games like paper mario, so it will be removed permanently. The hack is a placeholder for hardware texture decoding and encoding, so if someone gets around to adding that, we'd get that speed boost back. I don't see any reason it has to be a part of this change though

LexLight:
Lastest Canary, and finally the "Paper Mario" glitches are gone!
imagen
imagen
Through, hopefully that someone will be able to bring that speed boost back.

Dragios:
Looks like I missed some bugs during testing. Tested as of commit e48a61272c6153064e622fb65fea0bdfb13353b2

- This PR actually fix #3102 partially. On native it looks fine but once I auto scale you can see all kinds of artifacts like in the issue I reported previously.
PW:AA-DD
There is a chance the screen will swap itself. It only happens when save file is absent.

- **Cooking Mama 5** display issue remains (nothing changed). Now I highly believe both #2235 & #3102 are related somehow.

- **Metal Gear Solid 3D** flickering on some spot issue now occurs on both native and scaled resolution but at a lesser degree.
MGS3D
Notice that some spots on the sky is brighter on native while dimmer in scaled. The one in red circle shows that an artifact that looks like triangle appears out of no where (which shouldn't happen. The red arrow shows the small flickering area.

- **Tales of the Abyss** screen works now. Screen offset is still wrong but not related to this PR.

- **Pokémon outlines** are back but pixelated when on scaled resolution (same as nightly now)

- Performance warning log remains

LogPod:
Outlines are broken in Pokemon
Nightly
capture
Canary

Robz8:
Just wanted to let you all know.
Style Savvy: Styling Star is one of those games that look pixelated when scaled on the nightly builds, along with Metroid: Samus Returns.

Here are some screenshots.

ss4 title
ss4 character
ss4 abigail

Looks good and runs full speed!

I hope this PR gets merged if/when everything works properly. 👍

ds84182:
Just retested Art of Balance. Cropping bug seems to be fixed.

jroweboy:
Rebased off the latest master since someone said there was a conflict, and also addressed some review comments in the latest commit.

Dragios:
Nice work @jroweboy 
Here are my feedbacks based on 1c4d1d1aceed4a0be369b1969587e9c710fc083e
- Fixes #2105, #2235 & #3102 completely. The frameskipping issue when save file is absent for **Phoenix Wright: Ace Attorney - Dual Destinies** and **The Legend of Zelda: Ocarina of time** has been fixed. No more weird graphical artifacts when continuously switching scaling anymore. Mirrored screen issue no longer exist.

- #2741 is fixed as usual
---
Not solve yet:
- **Metal Gear Solid 3D** flickering on some spot issue now occurs on both native and scaled resolution but at a lesser degree.

- **Pokémon** outlines is still pixelated when on scaled resolution (as shown on image above my comment)

- This log still appears in some games on _Intel_ GPU. Will occur on _nVidia_ GPU if terminate emulation.
```
Render.OpenGL <Warning> video_core/renderer_opengl/renderer_opengl.cpp:DebugHandler:472:
API PERFORMANCE 131154: Pixel-path performance warning: Pixel transfer is synchronized with
3D rendering.
```

Senjosei:
Tested on Canary 268
The legend of legacy crash after main menu, before i tested with Canary 253 (build before texcache rework merge) it boots up fine
screenshot 559

jroweboy:
@Senjosei as a quick test, could you compare against a previous version of the texture cache code (preferably one without the very latest commit https://github.com/citra-emu/citra/pull/3281/commits/1c4d1d1aceed4a0be369b1969587e9c710fc083e which would help me determine if i messed up when removing some static state when i get some time to work on this again)

Senjosei:
My previous test was run on Canary 268 which i believe was a build from 13 days ago that doesnt have the [`1c4d1d1`](https://github.com/citra-emu/citra/commit/1c4d1d1aceed4a0be369b1969587e9c710fc083e). I also found another assert failure (both on Canary 268 and on latest without [`1c4d1d1`](https://github.com/citra-emu/citra/commit/1c4d1d1aceed4a0be369b1969587e9c710fc083e))on bravely default its a different assert. (Im using a custom build with MerryMage stack trace for debuging but the assert still fails even if i dont merge the stack trace) Any other commit u want me to test?
image

Dragios:
| Nightly  | This PR |
| ------------- | ------------- |
| image | image |

There is a large accumulation of black translucent pixels flashing in latest nightly on both native and scaled resolution on all scenes (except the video). On the other hand with this PR only small spots of black translucent pixels is visible with scaled resolution on this scene only. On native resolution, a few pixels flicker for a short moment during animation before it vanishes.
The game is **Monster Hunter 4 Ultimate**
Can this be fixed in your PR?

Senjosei:
@Dragios are u sure ur testing the right version? I tested MH4 on the previous build jroweboy asked me to do (master + texcache-pt2 without 1c4d1d1aceed4a0be369b1969587e9c710fc083e) and there are no black pixels.
screenshot 566
I also tested on another build with 1c4d1d1aceed4a0be369b1969587e9c710fc083e , canary 287 (4e053220a826676d0765ef9eee313a4b13e3cd0e) and i didnt have black pixels (tested on all combination of native res, 4x res, nvidia gpu, intel gpu)

Dragios:
@Senjosei Sorry for the misunderstanding. I should have rephrase my sentence. It does not occur all the time with this PR. I say its a 20% chance occuring. It happens when I constantly toggling between native and scaled resolution. This is tested on my nVidia GTX 1050 Ti

Senjosei:
@Dragios can confirm it may occur 
screenshot 569
Although i remembered this issue and i think its because of nvidia's tile based rasterizer added on maxwell and later, not because of this PR. I turned mine off and ive been sitting here spamming the switch resolution and it hasnt shown up again. I also tested this on nightly which doesnt have the `texcache-pt2` merged and the artifacts do show up but after i turned of the tile rasterizer, it doesnt show up (Links below for more info) https://forums.geforce.com/default/topic/1020411/geforce-drivers/tile-based-rasterizer-causing-artifacts https://www.reddit.com/r/cemu/comments/6qwmon/got_blocky_artifacts_in_lava_water_and_other/

Dragios:
From what I read it affects older games mostly. ~But 3DS is not really that old so it shouldn't get affected with tile based rasterizer. I believe it is due to the openGL rendering code itself.~ PICA200 is a very old GPU so I think you're correct in this case. Nevertheless @jroweboy and @Phanto-m should have a better idea on what is happening. I leave the issue note as it is and let them figure out what went wrong.

wwylele:
@Dragios just to clarify: citra is not a game, plus this has nothing to do with the age of the guest GPU (PICA200 in our case), because we are not fowarding guest GPU API directly to host GPU API; we translate it using a specific host GPU API subset, which might be not a common use comparing with modern PC games, and might be broken by some specific drivers.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -25,103 +31,223 @@
...
+    SCOPE_EXIT({ prev_state.Apply(); });
+
     // Make sure textures aren't bound to texture units, since going to bind them to framebuffer
     // components
-    OpenGLState::ResetTexture(src_tex);
-    OpenGLState::ResetTexture(dst_tex);
+    state.ResetTexture(src_tex);
+    state.ResetTexture(dst_tex);
 
     // Keep track of previous framebuffer bindings
```
wwylele: This comment is not relevant anymore

jroweboy:
Thanks for the initial read through @wwylele I will add some more comments on those parts hopefully later this week.

For everyone else that has reported issues, @Phanto-m has made some changes that should have fixed all of the reported issues up till now. Please retest and let me know if anything is wrong. Thanks!

tabnk:
Dragon Quest VII: Fragments of the Forgotten Past failed (black screen) to run with latest update.

jroweboy:
@tabnk iirc the dragon quest issue is actually related to https://github.com/citra-emu/citra/pull/3324 For now, can you test with builds straight from appveyor? https://ci.appveyor.com/api/buildjobs/d5gutqwjch97dpa1/artifacts/citra-windows-mingw-20180104-bc72485.zip

tabnk:
Dragon Quest VII: Fragments of the Forgotten Past run fine with https://ci.appveyor.com/api/buildjobs/d5gutqwjch97dpa1/artifacts/citra-windows-mingw-20180104-bc72485.zip.

Dragios:
Still not solve yet:
- **Metal Gear Solid 3D**

| Intel GPU | nVidia GPU |
| -------- | -------- |
| Referenced image as it is similar with real hardware MGS3D | Flickering on some spot issue https://streamable.com/q59t4 |

- Minor black translucent pixels still flashing in **Monster Hunter 4 Ultimate** when on scaled resolution
https://streamable.com/kqeyi

*Note: Both issue are tested on GTX 745. I will test with my more modern GPU (GTX 1050 Ti) once I reach home. Will update the note after that.*

- Pixel-path performance warning still appear as of latest commit

Hexagon12:
@Dragios The flashing black pixels in Monster Hunter games aren't related to Citra at all. It's a bug in Nvidia's tile based rasterizer, implemented in Maxwell and Pascal series of Nvidia GPUs.

Dragios:
@Hexagon12 if you did read my comment carefully you will realized this. GTX745 is a Kepler based card. So it is not related to tile based rasterizer.

tywald:
@Dragios I don't have this problem with GTX 680, so it's not Citra. I would try check anyway if that fix would solve the problem, I assume you know what it is.

wwylele:
@Dragios Since master branch also has this problem, and it is GPU-specified, I would suggest to not try to take effort to fix this in this PR. It can be a totally different issue and out of the scope of this PR, which only adds unnecessary work on this already huge code change.

Dragios:
Nvm I was wrong. GTX 745 was a Maxwell card. Anyway the reason I report this is because this PR managed to eliminate most of the problem that exist in master. Also a secret build by @jroweboy actually fixed all the issue in both of the games reported above. Don't ask where to get that build as it is for internal testing only.

urdaneta09:
Pixel-path performance warning appear too in Kingdom Hearts DDD

tabnk:
Dragon Quest VII: Fragments of the Forgotten Past. Haswell CPU.

Master build : 50+ fps
This PR : over 200+ fps

Really amazing speed improvements

whymsical:
Not sure if this is the right PR. Was told to add this to the PR for the texture cache. I can only assume this is the correct one. 

7th Dragon III Code VFD
If upscaling the game to anything above native, the game will look nice and good til after the first battle victory screen where it will then pixellate and then not recover. Performance also appears to be heavily crippled by any form of upscaling above native for this game.

image

x4 Native. Normal Looking. 

image

Combat, 4x Native. Looks Normal

image

Post Combat, x4 Native. Bug triggered. Performance Loss

image

Back on Map. Still bugged. Performance degrading

image

Next battle. Still bugged. (Perfomance returned, but battle isn't graphically heavy.)

tabnk:
Newer update from 

https://github.com/Phanto-m/citra/commit/fef0b5c8d90f07ec4317b2993d86fd83bcaa5799

seen to fixed 7th Dragon III Code VFD scalling issue.  👍

Lemiru:
Project x zone crashes emulator during certain attacks and cutscenes:
screen shot 01-24-18 at 10 49 am

LogPod:
New commit from Phanto-m 
https://github.com/Phanto-m/citra/commit/7c933fd01e70c006fcf15201a4b349abddf5a936

jroweboy:
Whats changed:

* Phantom: Removed an assert that was no longer correct. This should some crashes
* Added pokemon outline fix (d24s8 -> abgr conversion with PBOs)

I originally was planning on putting the outline fix in a follow up PR, but if someone wants to review that, i've added the changes here instead. Its pretty related to the rest of the changes here, as its essentially another corner case that the old texture cache didn't support.

Lemiru:
Project X zone doesn't crash anymore with canary build

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -159,641 +285,1326 @@ void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
+    // Attempt to find encompassing surface
+    Surface surface = FindMatch<MatchFlags::SubRect | MatchFlags::Invalid>(surface_cache, params,
+                                                                           match_res_scale);
+
+    // Check if FindMatch failed because of res scaling
+    // If that's the case create a new surface with
+    // the dimensions of the lower res_scale surface
+    // to suggest it should not be used again
+    if (surface == nullptr && match_res_scale != ScaleMatch::Ignore) {
+        surface = FindMatch<MatchFlags::SubRect | MatchFlags::Invalid>(surface_cache, params,
```
lioncash: The body here seems like something should be its own function.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -159,641 +285,1326 @@ void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
+        aligned_params.height = Common::AlignUp(params.height, 8);
+        aligned_params.width = Common::AlignUp(params.width, 8);
+        aligned_params.stride = Common::AlignUp(params.stride, 8);
+        aligned_params.UpdateParams();
+    }
 
-        return best_subrect_surface;
+    // Check for a surface we can expand before creating a new one
+    if (surface == nullptr) {
+        surface = FindMatch<MatchFlags::Expand | MatchFlags::Invalid>(surface_cache, aligned_params,
```
lioncash: ditto

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -159,641 +285,1326 @@ void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
-            color_surface = GetSurface(color_params, true, true);
-        }
-
-        if (depth_rect.left != 0 || depth_rect.top != 0) {
-            depth_surface = GetSurface(depth_params, true, true);
-        }
-
-        if (!color_surface->is_tiled) {
-            rect = MathUtil::Rectangle<int>(
-                0, 0, (int)(color_params.width * color_
```
lioncash: `sizeof(u32)` instead of `4`

lioncash: `(match_surface->is_tiled ? 8 : 1)` is repeated quite a bit. Perhaps make it its own `const`-qualified variable instead.

lioncash: This seems like it could be a const reference.

lioncash: This seems pretty sketchy, even if well-defined.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -159,641 +285,1326 @@ void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
+                  ScaleMatch match_scale_type,
+                  boost::optional<SurfaceInterval> validate_interval = boost::none) {
+    Surface match_surface = nullptr;
+    bool match_valid = false;
+    u32 match_scale = 0;
+    SurfaceInterval match_interval{};
+
+    for (auto& pair : RangeFromInterval(surface_cache, params.GetInterval())) {
+        for (auto& surface : pair.second) {
+            bool res_scale_matched = match_scale_type == ScaleMatch::Exact
```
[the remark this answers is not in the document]
jroweboy: i removed const as it caused an ICE on gcc5: https://github.com/citra-emu/citra/pull/3281/commits/ac4c589ab56af1ce44899868506b0c3e12b7f644

TTFH:
**Hyrule Warriors Legends**
citra-emu:master            17 fps
jroweboy:texcache-pt2   20 fps

Skylake 3.6 GHz, small modifications to avoid crash.

jroweboy:
TODO List before merging:

* I've addressed most of @lioncash's comments. I haven't done much research into the `const` vs `constexpr` for `MortonInterleave` so I'm iffy about reverting it. I also did not split those functions as recommended as it seemed like a stylistic choice, and I personally don't enjoy having to hunt down an extra function to find what I'm looking for. If you insist, I will gladly make the change though.
* wwylele suggested adding more comments to the header to describe the expected usage. I still need to do this.

Last call for reviews! Will be looking to merge this in the next couple of weeks 👍

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -159,641 +283,1324 @@ void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
+        glReadPixels(static_cast<GLint>(rect.left), static_cast<GLint>(rect.bottom),
+                     static_cast<GLsizei>(rect.GetWidth()), static_cast<GLsizei>(rect.GetHeight()),
+                     tuple.format, tuple.type, &gl_buffer[buffer_offset]);
+    }
 
-                std::vector<Math::Vec4<u8>> tex_buffer(params.width * params.height);
+    glPixelStorei(GL_PACK_ROW_LENGTH, 0);
+}
+
+enum MatchFlags {
```
wwylele: `enum class`

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -159,641 +283,1324 @@ void RasterizerCacheOpenGL::BlitTextures(GLuint src_tex, GLuint dst_tex,
...
-                MathUtil::Rectangle<int>(x0, best_subrect_surface->height - y0, x0 + params.width,
-                                         best_subrect_surface->height - (y0 + params.height));
+    // Attempt to find encompassing surface
+    Surface surface = FindMatch<MatchFlags::SubRect | MatchFlags::Invalid>(surface_cache, params,
+                                                                           match_res_scale);
+
+    // Check if FindMatch failed because of res scaling
+    // If that's the case create a new surface with
+    // the dimensions of the lower res_scale surface
+    // to suggest it should not be used again
```
wwylele: The line break is somewhat off.

wwylele:
When you fixing stuff, please squash the current code first (as a version freeze) and add more comments on top of it so I can easily read it, thanks!

Senjosei:
I dont remember if uve taken a look at this but theres this [https://community.citra-emu.org/t/cf-vanguard-stride-to-victory-canary-version-crash/10366](https://community.citra-emu.org/t/cf-vanguard-stride-to-victory-canary-version-crash/10366)

wwylele:
Here is some detail information about the issue above (thanks god I have a friend having this game on hand):

The assertion failure is on
```
void RasterizerCacheOpenGL::InvalidateRegion(...) {
    ....
    if (region_owner != nullptr) {
        ASSERT(region_owner->type != SurfaceType::Texture); // here
        ....
    }
}
```

The call stack is 
`GPU::TextureCopy`->`RasterizerOpenGL::AccelerateTextureCopy`->`InvaidateRegion`

The copy dst has `pixel_format = SurfaceParams::PixelFormat::ETC1` (inherited from src format?). I feel like this is an oversight of texture-only formats handling in TextureCopy, but I am not sure.

cc @jroweboy & @Phanto-m

jroweboy:
@wwylele pushed a quick fix for the issue. Let me know if it resolves it and we can merge this 👍

wwylele:
The issue is resolved. I pushed a clang-format fix commit. Will merge after CI check.

PR #3283, opened 2017-12-13
Optimize AttributeBuffer to OutputVertex conversion

ds84182:
First I unrolled the inner loop, then I pushed semantics validation
outside of the hotloop.

I also added overflow slots to avoid conditional branches.

Super Mario 3D Land's intro runs at almost full speed when compiled with
Clang, and theres a noticible speed increase in MSVC. GCC hasn't been
tested but I'm confident in its ability to optimize this code.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3283)

### DIFF
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -21,32 +21,39 @@ namespace Pica {
 
 namespace Shader {
 
+void OutputVertex::ValidateSemantics(const RasterizerRegs& regs) {
+    unsigned int num_attributes = regs.vs_output_total;
+    ASSERT(num_attributes <= 7);
+    for (unsigned int i = 0; i < num_attributes; ++i) {
+        u32 output_register_map = regs.vs_output_attributes[i].raw;
+        for (unsigned int j = 0; j < 4; j++) {
+            auto semantic = output_register_map & 0x1F;
+            ASSERT_MSG(semantic < 24 ||
+                semantic == RasterizerRegs::VSOutputAttributes::INVALID,
+                "Invalid/unknown semantic id: %u", (unsigned int)semantic);
+            output_register_map >>= 8;
+        }
+    }
+}
+
 OutputVertex OutputVertex::FromAttributeBuffer(const RasterizerRegs& regs,
                                                const AttributeBuffer& input) {
     // Setup output data
     union {
-        OutputVertex ret{};
+        OutputVertex ret {};
         std::array<float24, 24> vertex_slots;
+        // Allow us to overflow vertex_slots to avoid branches
+        std::array<float24, 32> vertex_slots_overflow;
     };
     static_assert(sizeof(vertex_slots) == sizeof(ret), "Struct and array have different sizes.");
 
-    unsigned int num_attributes = regs.vs_output_total;
-    ASSERT(num_attributes <= 7);
+    unsigned int num_attributes = regs.vs_output_total & 7;
     for (unsigned int i = 0; i < num_attributes; ++i) {
-        const auto& output_register_map = regs.vs_output_attributes[i];
-
-        RasterizerRegs::VSOutputAttributes::Semantic semantics[4] = {
-            output_register_map.map_x, output_register_map.map_y, output_register_map.map_z,
-            output_register_map.map_w};
-
-        for (unsigned comp = 0; comp < 4; ++comp) {
-            RasterizerRegs::VSOutputAttributes::Semantic semantic = semantics[comp];
-            if (semantic < vertex_slots.size()) {
-                vertex_slots[semantic] = input.attr[i][comp];
-            } else if (semantic != RasterizerRegs::VSOutputAttributes::INVALID) {
-                LOG_ERROR(HW_GPU, "Invalid/unknown semantic id: %u", (unsigned int)semantic);
-            }
-        }
+        u32 output_register_map = regs.vs_output_attributes[i].raw;
+        vertex_slots_overflow[output_register_map & 0x1F] = input.attr[i][0];
+        vertex_slots_overflow[(output_register_map >> 8) & 0x1F] = input.attr[i][1];
+        vertex_slots_overflow[(output_register_map >> 16) & 0x1F] = input.attr[i][2];
+        vertex_slots_overflow[(output_register_map >> 24) & 0x1F] = input.attr[i][3];
     }
 
     // The hardware takes the absolute and saturates vertex colors like this, *before* doing
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -24,9 +24,9 @@ namespace Shader {
 void OutputVertex::ValidateSemantics(const RasterizerRegs& regs) {
     unsigned int num_attributes = regs.vs_output_total;
     ASSERT(num_attributes <= 7);
-    for (size_t i = 0; i < num_attributes; ++i) {
-        u32 output_register_map = regs.vs_output_attributes[i].raw;
-        for (size_t j = 0; j < 4; j++) {
+    for (size_t attrib = 0; attrib < num_attributes; attrib++) {
+        u32 output_register_map = regs.vs_output_attributes[attrib].raw;
+        for (size_t comp = 0; comp < 4; comp++) {
             u32 semantic = output_register_map & 0x1F;
             ASSERT_MSG(semantic < 24 || semantic == RasterizerRegs::VSOutputAttributes::INVALID,
                        "Invalid/unknown semantic id: %u", semantic);
@@ -41,18 +41,20 @@ OutputVertex OutputVertex::FromAttributeBuffer(const RasterizerRegs& regs,
     union {
         OutputVertex ret{};
         std::array<float24, 24> vertex_slots;
-        // Allow us to overflow vertex_slots to avoid branches
+        // Allow us to overflow vertex_slots to avoid branches, since
+        // RasterizerRegs::VSOutputAttributes::INVALID would write to slot 31, which
+        // would be out of bounds otherwise.
         std::array<float24, 32> vertex_slots_overflow;
     };
     static_assert(sizeof(vertex_slots) == sizeof(ret), "Struct and array have different sizes.");
 
     unsigned int num_attributes = regs.vs_output_total & 7;
-    for (size_t i = 0; i < num_attributes; ++i) {
-        u32 output_register_map = regs.vs_output_attributes[i].raw;
-        vertex_slots_overflow[output_register_map & 0x1F] = input.attr[i][0];
-        vertex_slots_overflow[(output_register_map >> 8) & 0x1F] = input.attr[i][1];
-        vertex_slots_overflow[(output_register_map >> 16) & 0x1F] = input.attr[i][2];
-        vertex_slots_overflow[(output_register_map >> 24) & 0x1F] = input.attr[i][3];
+    for (size_t attrib = 0; attrib < num_attributes; attrib++) {
+        u32 output_register_map = regs.vs_output_attributes[attrib].raw;
+        vertex_slots_overflow[output_register_map & 0x1F] = input.attr[attrib][0];
+        vertex_slots_overflow[(output_register_map >> 8) & 0x1F] = input.attr[attrib][1];
+        vertex_slots_overflow[(output_register_map >> 16) & 0x1F] = input.attr[attrib][2];
+        vertex_slots_overflow[(output_register_map >> 24) & 0x1F] = input.attr[attrib][3];
     }
 
     // The hardware takes the absolute and saturates vertex colors like this, *before* doing
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -28,10 +28,9 @@ void OutputVertex::ValidateSemantics(const RasterizerRegs& regs) {
     for (size_t attrib = 0; attrib < num_attributes; ++attrib) {
         u32 output_register_map = regs.vs_output_attributes[attrib].raw;
         for (size_t comp = 0; comp < 4; ++comp) {
-            u32 semantic = output_register_map & 0x1F;
+            u32 semantic = (output_register_map >> 8 * comp) & 0x1F;
             ASSERT_MSG(semantic < 24 || semantic == RasterizerRegs::VSOutputAttributes::INVALID,
                        "Invalid/unknown semantic id: %" PRIu32, semantic);
-            output_register_map >>= 8;
         }
     }
 }
@@ -48,15 +47,16 @@ OutputVertex OutputVertex::FromAttributeBuffer(const RasterizerRegs& regs,
     };
 
     // Assert that OutputVertex has enough space for 24 semantic registers
-    static_assert(sizeof(std::array<float24, 24>) == sizeof(ret), "Struct and array have different sizes.");
+    static_assert(sizeof(std::array<float24, 24>) == sizeof(ret),
+                  "Struct and array have different sizes.");
 
     unsigned int num_attributes = regs.vs_output_total & 7;
     for (size_t attrib = 0; attrib < num_attributes; ++attrib) {
-        u32 output_register_map = regs.vs_output_attributes[attrib].raw;
-        vertex_slots_overflow[output_register_map & 0x1F] = input.attr[attrib][0];
-        vertex_slots_overflow[(output_register_map >> 8) & 0x1F] = input.attr[attrib][1];
-        vertex_slots_overflow[(output_register_map >> 16) & 0x1F] = input.attr[attrib][2];
-        vertex_slots_overflow[(output_register_map >> 24) & 0x1F] = input.attr[attrib][3];
+        const auto &output_register_map = regs.vs_output_attributes[attrib];
+        vertex_slots_overflow[output_register_map.map_x] = input.attr[attrib][0];
+        vertex_slots_overflow[output_register_map.map_y] = input.attr[attrib][1];
+        vertex_slots_overflow[output_register_map.map_z] = input.attr[attrib][2];
+        vertex_slots_overflow[output_register_map.map_w] = input.attr[attrib][3];
     }
 
     // The hardware takes the absolute and saturates vertex colors like this, *before* doing
--- a/src/video_core/shader/shader.cpp
+++ b/src/video_core/shader/shader.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <cinttypes>
 #include <cmath>
 #include <cstring>
 #include "common/bit_set.h"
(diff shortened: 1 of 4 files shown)


src/video_core/shader/shader.cpp
```diff
@@ -21,32 +21,39 @@ namespace Pica {
 
 namespace Shader {
 
+void OutputVertex::ValidateSemantics(const RasterizerRegs& regs) {
+    unsigned int num_attributes = regs.vs_output_total;
+    ASSERT(num_attributes <= 7);
+    for (unsigned int i = 0; i < num_attributes; ++i) {
+        u32 output_register_map = regs.vs_output_attributes[i].raw;
+        for (unsigned int j = 0; j < 4; j++) {
+            auto semantic = output_register_map & 0x1F;
```
lioncash: `u32` can just be specified here.

src/video_core/shader/shader.cpp
```diff
@@ -21,32 +21,39 @@ namespace Pica {
...
+            ASSERT_MSG(semantic < 24 ||
+                semantic == RasterizerRegs::VSOutputAttributes::INVALID,
+                "Invalid/unknown semantic id: %u", (unsigned int)semantic);
```
lioncash: This was pre-existing so feel free to ignore this, but the cast isn't necessary, given it's already a `u32`

src/video_core/shader/shader.cpp
```diff
@@ -21,32 +21,39 @@ namespace Pica {
 
 namespace Shader {
 
+void OutputVertex::ValidateSemantics(const RasterizerRegs& regs) {
+    unsigned int num_attributes = regs.vs_output_total;
+    ASSERT(num_attributes <= 7);
+    for (unsigned int i = 0; i < num_attributes; ++i) {
```
lioncash: Making `i` a `size_t` here (as well as `j` for consistency I guess, but it won't have any codegen difference since it's not used to index memory) can also avoid a zero extension of a register on every loop iteration ([example](https://godbolt.org/g/YY1SpH))

src/video_core/shader/shader.cpp
```diff
@@ -21,32 +21,39 @@ namespace Pica {
...
         std::array<float24, 24> vertex_slots;
+        // Allow us to overflow vertex_slots to avoid branches
+        std::array<float24, 32> vertex_slots_overflow;
     };
     static_assert(sizeof(vertex_slots) == sizeof(ret), "Struct and array have different sizes.");
 
-    unsigned int num_attributes = regs.vs_output_total;
-    ASSERT(num_attributes <= 7);
+    unsigned int num_attributes = regs.vs_output_total & 7;
     for (unsigned int i = 0; i < num_attributes; ++i) {
```
lioncash: ditto, this can also be made a size_t for above reason in a comment

src/video_core/shader/shader.cpp
```diff
@@ -21,32 +21,39 @@ namespace Pica {
...
+
 OutputVertex OutputVertex::FromAttributeBuffer(const RasterizerRegs& regs,
                                                const AttributeBuffer& input) {
     // Setup output data
     union {
-        OutputVertex ret{};
+        OutputVertex ret {};
         std::array<float24, 24> vertex_slots;
+        // Allow us to overflow vertex_slots to avoid branches
+        std::array<float24, 32> vertex_slots_overflow;
```
MerryMage: `std::array::operator[]` doesn't perform bounds checking; is this necessary instead of accessing `vertex_slots` directly?

ds84182: I used `vertex_slots_overflow` instead of `vertex_slots` because `RasterizerRegs::VSOutputAttributes::INVALID` (the no op semantic) would set vertex slot 31.

src/video_core/shader/shader.cpp
```diff
@@ -21,32 +21,38 @@ namespace Pica {
 
 namespace Shader {
 
+void OutputVertex::ValidateSemantics(const RasterizerRegs& regs) {
+    unsigned int num_attributes = regs.vs_output_total;
+    ASSERT(num_attributes <= 7);
+    for (size_t i = 0; i < num_attributes; ++i) {
```
Subv: Please use more meaningful names for these loop vars (attribute and comp)

src/video_core/shader/shader.cpp
```diff
@@ -21,32 +21,38 @@ namespace Pica {
...
+    }
+}
+
 OutputVertex OutputVertex::FromAttributeBuffer(const RasterizerRegs& regs,
                                                const AttributeBuffer& input) {
     // Setup output data
     union {
         OutputVertex ret{};
         std::array<float24, 24> vertex_slots;
+        // Allow us to overflow vertex_slots to avoid branches
```
Subv: This comment should explain that it's related to VSOutputAttributes::INVALID

tabnk:
Master (nightly-488 Mingw):
Super Mario 3D Land - 74 fps
Mario Kart 7 - 24 fps 

This PR:
Super Mario 3D Land - 78 fps
Mario Kart 7 - 27 fps 

Nice :)

wwylele:
@Hexagon12 reports a potential performance regression due to this PR. Please hold on before confirmation.

Hexagon12:
Game: Ace Attorney - Dual Destinies, in a court, resolution 5x native.
Latest master, MSVC
Speed: 12%
FPS: 7 FPS 
Frametime: 144.94 ms 

Latest master + #3281 latest commit, MSVC 
Speed: 28% 
FPS: 17 FPS 
Frametime: 59.48 ms 

Latest master + #3283 latest commit, MSVC 
Speed: 10% 
FPS: 5 FPS 
Frametime: 151.56 ms

This regression is seen with Capcom games. (Monster Hunter, Ace Attorney)

src/video_core/shader/shader.cpp
```diff
@@ -21,32 +21,40 @@ namespace Pica {
...
-                LOG_ERROR(HW_GPU, "Invalid/unknown semantic id: %u", (unsigned int)semantic);
-            }
-        }
+    unsigned int num_attributes = regs.vs_output_total & 7;
+    for (size_t attrib = 0; attrib < num_attributes; attrib++) {
+        u32 output_register_map = regs.vs_output_attributes[attrib].raw;
+        vertex_slots_overflow[output_register_map & 0x1F] = input.attr[attrib][0];
+        vertex_slots_overflow[(output_register_map >> 8) & 0x1F] = input.attr[attrib][1];
+        vertex_slots_overflow[(output_register_map >> 16) & 0x1F] = input.attr[attrib][2];
+        vertex_slots_overflow[(output_register_map >> 24) & 0x1F] = input.attr[attrib][3];
```
wwylele: What's the benefit to do raw bit operation than using BitField member functions(which is more readable)? These two should generate the same code after inline unfolding, unless the compiler can't eliminate the const if

Subv:
It's weird that this would cause a performance regression, what's the generated assembly for the new code?

yuriks:
It doesn't look like there's anything in this PR that could make it *slower* (except if the overhead of validating the semantics separately makes it slower than doing it in the same loop, but that should only be the case if the game is rendering, like, 2 vertices per batch). The reported frametime difference could be just the result of noise. Are those games causing a lot of texture download/uploads?

---

Reviewed 3 of 4 files at r1, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/video_core/command_processor.cpp, line 292 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0NkI3IYMnrxyqzE1II:-L0NkI3IYMnrxyqzE1IJ:b1s2zpf) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/command_processor.cpp#L292)):*
> ```cpp
>         const u32 base_address = regs.pipeline.vertex_attributes.GetPhysicalBaseAddress();
>         VertexLoader loader(regs.pipeline);
>         Pica::Shader::OutputVertex::ValidateSemantics(regs.rasterizer);
> ```

You don't need to specify the `Pica::` part.

---

*[src/video_core/command_processor.cpp, line 292 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0NkaBAZnoqJJ_QEGZg:-L0NkaBAZnoqJJ_QEGZh:b-rsdqke) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/command_processor.cpp#L292)):*
> ```cpp
>         const u32 base_address = regs.pipeline.vertex_attributes.GetPhysicalBaseAddress();
>         VertexLoader loader(regs.pipeline);
>         Pica::Shader::OutputVertex::ValidateSemantics(regs.rasterizer);
> ```

You need to run validation for immediate mode drawing too.

---

*[src/video_core/shader/shader.cpp, line 33 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0FZ4nk8y9Kch_1fmk0-r1-33:-L0NbCzFOQFCBItaPCax:bk4nymp) ([raw file](https://github.com/citra-emu/citra/blob/bf4b18d8dc5801064b66a7badfda253d88342eef/src/video_core/shader/shader.cpp#L33)):*

This was pre-existing so feel free to ignore this, but the cast isn't necessary, given it's already a `u32`
</blockquote></details>

IIRC this generates a warning on some compilers (where u32 is a unsigned long, not int), which is why the cast is there. Maybe use `PRIu32` instead (from `<cinttypes>`).

---

*[src/video_core/shader/shader.cpp, line 27 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0Nh_uu5zzMcHvWXEhP:-L0Nh_uu5zzMcHvWXEhQ:be020pl) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L27)):*
>  attrib++

`++attrib`

---

*[src/video_core/shader/shader.cpp, line 30 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0NcGJs70A5WRxuj0V4:-L0NcGJs70A5WRxuj0V5:b-t32hme) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L30)):*
> ```cpp
>         u32 output_register_map = regs.vs_output_attributes[attrib].raw;
>         for (size_t comp = 0; comp < 4; comp++) {
>             u32 semantic = output_register_map & 0x1F;
> ```

IMO just shift by `8 * comp` here in this expression, instead of mutating the variable.

---

*[src/video_core/shader/shader.cpp, line 49 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0NjJhztxaEdcPP_GUi:-L0NjJhztxaEdcPP_GUj:b-ph0rbz) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L49)):*
> ```cpp
>         std::array<float24, 32> vertex_slots_overflow;
>     };
>     static_assert(sizeof(vertex_slots) == sizeof(ret), "Struct and array have different sizes.");
> ```

This seems to be the only place where vertex_slots is still used. You can probably change it to do sizeof against the array type directly instead and then get rid of it.

---

*[src/video_core/shader/shader.cpp, line 52 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0NhdO-GjP01ElkdIO4:-L0NhdO-GjP01ElkdIO5:b-49bo13) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L52)):*
>  attrib++

`++attrib` again

---

*[src/video_core/shader/shader.cpp, line 57 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0FZ4nk8y9Kch_1fmk0-r3-57:-L0NjkMrQIBVM1d8Zoxd:b32v57) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L57)):*

What's the benefit to do raw bit operation than using BitField member functions(which is more readable)? These two should generate the same code after inline unfolding, unless the compiler can't eliminate the (constexpr) if
</blockquote></details>

+1

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3283#-:-L0Njuix-qJZCFWRLkwW:b-hlx69x)*

yuriks:
What's the status of this, can we wrap this up and merge it?

ds84182:
Yeah, I'll address comments today. Kinda had a busy week unfortunately

src/video_core/shader/shader.cpp
ds84182: MSVC doesn't alias bitfield accesses so it ends up loading the value from memory over and over again https://godbolt.org/g/P4QNpB

wwylele: IMO if this is just MSVC being dumb I think we can ignore it, since our release uses mingw as the main one on windows.

yuriks: You can just store it in a local variable so that it does not alias against the other array: https://godbolt.org/g/3s6uoZ

(The reason MSVC doesn't optimize this is because it doesn't do TBAA, so it considers vertex_slots and the attributes as potentially aliasing.)

yuriks:
Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/video_core/shader/shader.cpp, line 30 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0NcGJs70A5WRxuj0V4:-L1oSt3tm-hvbp6HYhpV:b-jkwgi1) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L30)):*

IMO just shift by `8 * comp` here in this expression, instead of mutating the variable.
</blockquote></details>

Seems like this is still un-addressed?

---

*[src/video_core/shader/shader.cpp, line 57 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0FZ4nk8y9Kch_1fmk0-r3-57:-L1oSwTtv0hCniiEDe1H:b-jkwgi1) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L57)):*

You can just store it in a local variable so that it does not alias against the other array: https://godbolt.org/g/3s6uoZ

(The reason MSVC doesn't optimize this is because it doesn't do TBAA, so it considers vertex_slots and the attributes as potentially aliasing.)
</blockquote></details>

Seems like this is still un-addressed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3283)*

ds84182:
Review status: 3 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/video_core/shader/shader.cpp, line 33 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0FZ4nk8y9Kch_1fmk0-r1-33:-L1oveUifNt3hvLsqAud:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/bf4b18d8dc5801064b66a7badfda253d88342eef/src/video_core/shader/shader.cpp#L33)):*

IIRC this generates a warning on some compilers (where u32 is a unsigned long, not int), which is why the cast is there. Maybe use `PRIu32` instead (from `<cinttypes>`).
</blockquote></details>

Done.

---

*[src/video_core/shader/shader.cpp, line 46 at r1](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0FZ4nk8y9Kch_1fmk0-r1-46:-L1ovkIbdWTTEc5PgTDL:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/bf4b18d8dc5801064b66a7badfda253d88342eef/src/video_core/shader/shader.cpp#L46)):*
<details><summary><i>Previously, ds84182 (Dwayne Slater) wrote…</i></summary><blockquote>

I used `vertex_slots_overflow` instead of `vertex_slots` because `RasterizerRegs::VSOutputAttributes::INVALID` (the no op semantic) would set vertex slot 31.
</blockquote></details>

Done.

---

*[src/video_core/shader/shader.cpp, line 30 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0NcGJs70A5WRxuj0V4:-L1ovUkzmMAm9QC_pR9A:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L30)):*

Seems like this is still un-addressed?
</blockquote></details>

Done.

---

*[src/video_core/shader/shader.cpp, line 57 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0FZ4nk8y9Kch_1fmk0-r3-57:-L1ovgh1K65iDdsgg7UZ:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L57)):*

Seems like this is still un-addressed?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3283)*

yuriks:
Review status: 3 of 4 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/video_core/shader/shader.cpp, line 30 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0NcGJs70A5WRxuj0V4:-L1pHdYxVuMlvOLwRpTj:b-4werdk) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L30)):*
<details><summary><i>Previously, ds84182 (Dwayne Slater) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

SGTM but I'd recommend adding parens around `8 * comp` just for clarity.

---

*[src/video_core/shader/shader.cpp, line 57 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0FZ4nk8y9Kch_1fmk0-r3-57:-L1pGk6ibWnBMiPFIfYI:b-ssdkfd) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L57)):*
<details><summary><i>Previously, ds84182 (Dwayne Slater) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

The way you're doing it here now will still imply aliasing, which caused the original problem. You want to make a local *copy* of the value, not take it by reference. Removing the `&` should be enough.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3283)*

ds84182:
Review status: 3 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/video_core/shader/shader.cpp, line 30 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0NcGJs70A5WRxuj0V4:-L1rj2N6UzNgDDydRLuq:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L30)):*

SGTM but I'd recommend adding parens around `8 * comp` just for clarity.
</blockquote></details>

Done.

---

*[src/video_core/shader/shader.cpp, line 57 at r3](https://reviewable.io:443/reviews/citra-emu/citra/3283#-L0FZ4nk8y9Kch_1fmk0-r3-57:-L1riwW5JZ4mhb0faTgF:b-896fix) ([raw file](https://github.com/citra-emu/citra/blob/2c9fae242be18d4c2d18c732c3eab8c52695897d/src/video_core/shader/shader.cpp#L57)):*

The way you're doing it here now will still imply aliasing, which caused the original problem. You want to make a local *copy* of the value, not take it by reference. Removing the `&` should be enough.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3283)*

yuriks:
Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/citra-emu/citra/3283)*

PR #3284, opened 2017-12-13
 HLE/Services: Allow specifying a SessionData template parameter to ServiceFramework.

Subv:
Some services can have multiple clients at the same time, and they identify the different clients using the server session as a key.
This parameter (if present) should be a structure that contains the per-session data for each service.
The data can be retrieved using ServiceFramework::GetSessionData(session)

cc @wwylele

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3284)

### DIFF
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -2,26 +2,32 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
 #include <vector>
-#include <boost/range/algorithm_ext/erase.hpp>
 #include "common/assert.h"
 #include "common/common_types.h"
 #include "core/hle/kernel/handle_table.h"
 #include "core/hle/kernel/hle_ipc.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/process.h"
-#include "core/hle/kernel/server_session.h"
 
 namespace Kernel {
 
+SessionRequestHandler::SessionInfo::SessionInfo(SharedPtr<ServerSession> session,
+                                                std::unique_ptr<SessionDataBase> data)
+    : session(std::move(session)), data(std::move(data)) {}
+
 void SessionRequestHandler::ClientConnected(SharedPtr<ServerSession> server_session) {
     server_session->SetHleHandler(shared_from_this());
-    connected_sessions.push_back(server_session);
+    connected_sessions.emplace_back(std::move(server_session), MakeSessionData());
 }
 
 void SessionRequestHandler::ClientDisconnected(SharedPtr<ServerSession> server_session) {
     server_session->SetHleHandler(nullptr);
-    boost::range::remove_erase(connected_sessions, server_session);
+    connected_sessions.erase(
+        std::remove_if(connected_sessions.begin(), connected_sessions.end(),
+                       [&](const SessionInfo& info) { return info.session == server_session; }),
+        connected_sessions.end());
 }
 
 HLERequestContext::HLERequestContext(SharedPtr<ServerSession> session)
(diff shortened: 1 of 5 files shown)


src/core/hle/kernel/hle_ipc.h
```diff
@@ -56,11 +57,34 @@ class SessionRequestHandler : public std::enable_shared_from_this<SessionRequest
...
-    /// A ServerSession whose server endpoint is an HLE implementation is kept alive by this list
-    // for the duration of the connection.
-    std::vector<SharedPtr<ServerSession>> connected_sessions;
+    /// Creates the storage for the session data of the service.
+    virtual std::unique_ptr<SessionDataBase> MakeSessionData() const = 0;
+
+    /// Returns the session data associated with the server session.
+    template <typename T>
+    T* GetSessionData(SharedPtr<ServerSession> session) {
+        static_assert(std::is_base_of<SessionDataBase, T>::value,
```
lioncash:
```cpp
std::is_base_of<SessionDatabase, T>()
```

src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -2,26 +2,34 @@
...
+void SessionRequestHandler::ClientConnected(
+    Kernel::SharedPtr<Kernel::ServerSession> server_session) {
     server_session->SetHleHandler(shared_from_this());
-    connected_sessions.push_back(server_session);
+    connected_sessions.emplace_back(std::move(server_session), MakeSessionData());
 }
 
-void SessionRequestHandler::ClientDisconnected(SharedPtr<ServerSession> server_session) {
+void SessionRequestHandler::ClientDisconnected(
+    Kernel::SharedPtr<Kernel::ServerSession> server_session) {
```
wwylele: No need for `Kernel::`

src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -2,26 +2,34 @@
...
 
 namespace Kernel {
 
-void SessionRequestHandler::ClientConnected(SharedPtr<ServerSession> server_session) {
+SessionRequestHandler::SessionInfo::SessionInfo(SharedPtr<ServerSession> session,
+                                                std::unique_ptr<SessionDataBase> data)
+    : session(std::move(session)), data(std::move(data)) {}
+
+void SessionRequestHandler::ClientConnected(
+    Kernel::SharedPtr<Kernel::ServerSession> server_session) {
```
wwylele: No need for `Kernel::`

src/core/hle/service/fs/archive.h
```diff
@@ -108,6 +101,10 @@ class Directory final : public Kernel::SessionRequestHandler {
 
 protected:
     void HandleSyncRequest(Kernel::SharedPtr<Kernel::ServerSession> server_session) override;
+
+    std::unique_ptr<SessionDataBase> MakeSessionData() const override {
+        return nullptr;
```
wwylele: Shouldn't this return a pointer to a new `FileSessionSlot` isntance?

src/core/hle/kernel/hle_ipc.h
```diff
@@ -47,7 +48,7 @@ class SessionRequestHandler : public std::enable_shared_from_this<SessionRequest
      * associated ServerSession alive for the duration of the connection.
      * @param server_session Owning pointer to the ServerSession associated with the connection.
      */
-    void ClientConnected(SharedPtr<ServerSession> server_session);
+    virtual void ClientConnected(SharedPtr<ServerSession> server_session);
```
wwylele: What's the reason to make this virtual?

src/core/hle/service/fs/archive.h
Subv: This is the Directory class

wwylele: oops, but in this case why is this needed?

PR #3286, opened 2017-12-14
citra_qt: 3DS Installation File -> CTR Importable Archive

valentinvanelslande:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3286)


wwylele:
I am aware that the name you proposed is the "official" name we inferred from the abbreviation. However, from a user's point of view:
 - The two have the same meaning
 - The old one is far more understandable (what the hell is "CTR"?)

BreadFish64:
I think this is **bad** from an end user standpoint, I deliberately chose to word it the way I did to avoid confusion. In the file selection dialog for running games the file type is "3DS executable" not "CTR executable image"

Issue #3289, opened 2017-12-14
Citra-SDL: telemetry silently on by default


Tilka:
The SDL version of Citra never asks the user for permission to use telemetry but enables it by default.

Tested with 4b8a7eb1ca19a9ebbf81886fa6f35d6d6fd491e6.

PR #3290, opened 2017-12-14
Disable telemetry by default in SDL frontend

BreadFish64:
As noted in issue #3289, SDL frontend users are never notified about telemetry at startup. This should not be enabled without explicit permission from the user.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3290)

### DIFF
--- a/src/citra/config.cpp
+++ b/src/citra/config.cpp
@@ -159,7 +159,7 @@ void Config::ReadValues() {
 
     // Web Service
     Settings::values.enable_telemetry =
-        sdl2_config->GetBoolean("WebService", "enable_telemetry", true);
+        sdl2_config->GetBoolean("WebService", "enable_telemetry", false);
     Settings::values.telemetry_endpoint_url = sdl2_config->Get(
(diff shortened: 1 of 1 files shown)


FearlessTobi:
Wait a minute... 

"To help improve Citra, the Citra Team collects anonymous usage data. [...] **By default, this feature is enabled**. [...]"

Doesn´t that mean it **should** als be opt-out in SDL2?

BreadFish64:
@FearlessTobi but SDL users are not told that it is enabled by default, which is bad.

TTFH:
Finally someone cares about it.

bunnei:
We could just add a pop up....

bunnei:
In fact, yeah, I don't want to merge this change - I think we want a message like the Qt frontend, if folks care enough. We could also just put a message in the installer. Otherwise, we will effectively be losing all telemetry for SDL build.

BreadFish64:
mmk, I'll come up with something else

PR #3293, opened 2017-12-14
[WIP] Allow user-configurable analog input deadzones

### LINKED ISSUE #3291: Analog deadzone issues with left and down.
Citra Version: Canary Build - 938c990
Game: Majora's Mask (probably happens on others)

When controlling the camera with the C-Stick (right analog on my Xbox One wireless controller), the camera still turns ever-so-slightly if my last input was left or down. Right and Up work as expected.

brebory:
Still working on the solution for this, but I think a good approach to fix the problem with strange dead zones for analog inputs is to allow users to configure the dead zones in the application. Currently, a hard coded threshold value of 40 is used in the HID analog input handler. This could be exposed to the user through the application settings and allow them to find a dead zone value that works for their hard ware.

There's also a possible integer overflow error in `src/core/hle/hid/hid.cpp l:71`, so perhaps addressing that will also fix the problems outlined in the linked issue.

I'm still setting up my development environment for debugging, so I've only arrived at these conclusions from speculation from looking at the code.

Fixes #3291

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3293)

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -65,7 +65,7 @@ DirectionState GetStickDirectionState(s16 circle_pad_x, s16 circle_pad_y) {
     constexpr float TAN30 = 0.577350269f;
     constexpr float TAN60 = 1 / TAN30;
     // a circle pad radius greater than 40 will trigger circle pad direction
-    constexpr int CIRCLE_PAD_THRESHOLD_SQUARE = 40 * 40;
+    int CIRCLE_PAD_THRESHOLD_SQUARE = Settings::Values::input_threshold * 0x9c;
     DirectionState state{false, false, false, false};
 
     if (circle_pad_x * circle_pad_x + circle_pad_y * circle_pad_y > CIRCLE_PAD_THRESHOLD_SQUARE) {
--- a/src/core/settings.h
+++ b/src/core/settings.h
@@ -82,6 +82,7 @@ struct Values {
     std::array<std::string, NativeAnalog::NumAnalogs> analogs;
     std::string motion_device;
     std::string touch_device;
+    float input_threshold;
 
     // Core
     bool use_cpu_jit;


B3n30:
As stated in discord, changing the deadzone in the service isn't the way to fix that issue, since the deadzone of 40 is hardcoded in the 3ds service. The deadzone you want to look for is probably in input_common. So I will close this for now.

MerryMage:
Hi and welcome to Citra thanks for taking your time to make a contribution!

Citra is a high-level emulator and (currently) emulates most of the 3DS’s operating system and system modules as such. HID is a module that runs on the 3DS; this part of Citra represents code that would actually have run on the real machine.

A change in the way input is handled should be done in input_common, before input is translated into the 3DS format. Separation of emulation code from what is a frontend problem is ideal.

Feel free to poke me on discord if I’m available in the future.

wwylele:
in case you didn't see my comment on discord, I'll copy it here:

The threshold 40 in HID/IRRST is hardcoded in real 3DS as well, and as a service emulation, we shouldn't change that. Also, that threshold is to only used to generate the direction bits which in fact not many games are using (they calculate it themselves from the raw -156~+156 data instead), so this isn't the "deadzone" setting you are looking for.
When addressing the deadzone issue, please limit the area of code change before it passes to the core. i.e. Ideally anything that need to be changed should be in input_common(edited)
A built-in deadzone configuration in citra would be a layer between SDL data and Input::AnalogDevice::GetStatus data, which rescales the radius from [deadzone_radius, 1.0] to [0.0, 1.0]

But I am not a fan of this. I agree with Subv that this should be handled by the operating system

>circle_pad_x and circle_pad_y could possibly each be 156, squared = 24336, and added together overflows s16 max
hmm doesn't s16 get converted to int before multiplication? (though I do agree that int > s16 is an assumption)

PR #3295, opened 2017-12-15
citra_qt: CMakeLists: Drop leftover handling code for Qt 4 UI files

lioncash:
We don't support Qt 4 anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3295)


src/citra_qt/CMakeLists.txt
```diff
@@ -87,8 +87,6 @@ file(GLOB_RECURSE THEMES ${CMAKE_SOURCE_DIR}/dist/qt_themes/*)
 
 if (Qt5_FOUND)
```
yuriks: You can just get rid of the if entirely, since this module if Qt-only in the first place.

B3n30:
If I merge this now, does it break Canary again?

lioncash:
@B3n30 I've no idea how Canary is supposed to operate, so I wouldn't be able to say

PR #3297, opened 2017-12-15
citra-qt: add translation support; link translation with transifex

wwylele:
Thanks @lioncash, @Subv, @B3N30, @jroweboy, @MerryMage and many others for discussing the idea!

This sets up the framework for supporting translations. The translation work is centralized on [citra project on transifex](https://www.transifex.com/citra/citra). The project there is still being tested and the exact management plan is not decided yet. However, I plan to only manually fetch the translation there to the main repo, so we can set up the github side first and discuss the transfex side later.

Contents in this PR:
 - Qt frontend code change: scan the resource folder ":/languages/" and list all available QM files as languages in the menu.
 - Two CMake change. Both of them needs dependency Qt Linguist Tools, can be turned on/off by options
   - Compile TS files in the language folder (default to `dist/languages/`) to QM files and bundle them in to Qt resource folder ":/languages/" so that Qt frontend can find them
   - Generate source TS file (en.ts) from the source code to the language folder. This file is for translator to use as the base file.
 - folder added: `dist/languages/`, currently contains only transifex configuration. In the future, this will hold all translation TS files. The folder is linked with citra project on transifex, and one can use command `tx pull -a` to fetch the newest translation files.
 - Two CI changes:
   - Travis CI has an additional job on the main repo master branch only. On every master branch push, it will execute en.ts generating command and push this file to transifex. A secret token will be later added only to the main repo to ensure this won't be executed on other forks.
   - All CI jobs enable building with translations, except for
     - Linux frozen version, because it is not use for release and including translation or not doesn't affect test result.
     - Windows MSVC build, because this requires an update to the Qt binary in our repo. Since this is not the main release version, we can do it later.

Translation workflow:
 - on every master branch update, translatable strings is pushed to transifex
 - translator do their work on transifex.
 - when enough new translation is made on transifex, or at a specific period like every week/month, a person or a bot executes `tx pull -a` in `dist/languages/`, commits the change and open a PR to the main repo, and thus update the translation in the official build. I will do this for some initial translation updates.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3297)


wwylele:
Info for tester: I don't include any actual translation in this PR, so by directly building citra, the only language showing up in the menu is "English". Please manually download ts files or use `tx pull -a` from the transifex site to directory `dist/languages`, and configure cmake with `ENABLE_QT_TRANSLATION` on to use unofficial translation to test it.

About the obligatory screenshot for UI change, I found it hard to screenshot the menu on my system, so I am afraid I can't provide it. All the UI change is added a "Language" submenu, which contains a list of languages, to the menu "File".

Subv:
Are we planning on shipping all translated languages by default? I could see the list of languages growing very large if so. In this case, I don't think that a popup menu is the best way to present the available languages. What about a selectbox in the Configuration dialog?

wwylele:
> Are we planning on shipping all translated languages by default?

Yes this is my plan

> What about a selectbox in the Configuration dialog?

hmm this was an option I considered. I decided to make it into menu because I think it is easier to access for a person who doesn't understand English very well. And if I make it in the configuration dialog, it would be either a combo box or a list box. IMO A combo box effectively have the same problem as a menu when the list is long; a list box for choosing language just sounds... weird. I am not sure about this. If more people agree that moving to configuration dialog is better, I will change it.

jroweboy:
@wwylele I think the best solution for that would be to use QLocale::system
to get their OS language is on start up, and if we have a matching
translation, then use that by default. Then in the 80% case that we have a
translation for their language and we can properly detect it and show their
preferred language, then its best to have it as a config option. On the
other hand, in the 20% case that we somehow show the wrong language (ie: we
mistakenly grabbed region instead of os language, or they have their os
language set to something else and they didn't know that, or we don't have
that language supported) then it would be hard to find it in the config
menu, but maybe not harder than finding it in the drop down, so I don't
really see this as a difference between menu or drop down.

I guess my whole point boils down to we should really try to get their
language from the system and use it as the default instead of english if we
can :)

src/citra_qt/CMakeLists.txt
```diff
@@ -91,12 +93,43 @@ else()
...
+
+    # Find all TS files except en.ts
+    file(GLOB_RECURSE LANGUAGES_TS ${CITRA_QT_LANGUAGES}/*.ts)
+    list(REMOVE_ITEM LANGUAGES_TS ${CITRA_QT_LANGUAGES}/en.ts)
+
+    # Compile TS files to QM files
+    qt5_add_translation(LANGUAGES_QM ${LANGUAGES_TS})
+
+    # Build a QRC file from the QM file list
+    SET(LANGUAGES_QRC ${CMAKE_CURRENT_BINARY_DIR}/languages.qrc)
```
lioncash: this `SET` and the following `FILE`'s should be lower-cased.

src/citra_qt/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ set(CMAKE_AUTORCC ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/CMakeModules)
 
-add_executable(citra-qt
+set(SRCS
```
lioncash: Sources can be retrieved from a target with

```cmake
get_target_property([name of output variable] citra-qt SOURCES)
```

there's no need to make a file-scoped variable of all the sources.

src/citra_qt/CMakeLists.txt
```diff
@@ -91,12 +93,43 @@ else()
     qt4_wrap_ui(UI_HDRS ${UIS})
 endif()
 
+if (ENABLE_QT_TRANSLATION)
+
```
lioncash: Unnecessary newline

src/citra_qt/main.cpp
```diff
@@ -1084,6 +1118,37 @@ void GMainWindow::UpdateUITheme() {
...
+    }
+
+    UISettings::values.translation_file = action->data().toString();
+    LoadTranslation();
+    ui.retranslateUi(this);
+    SetupUIStrings();
+}
+
+void GMainWindow::SetupUIStrings() {
+    setWindowTitle(QString("Citra %1| %2-%3")
```
lioncash: This should likely be a translatable string, considering ordering can differ between languages.

src/citra_qt/CMakeLists.txt
```diff
@@ -91,12 +93,43 @@ else()
     qt4_wrap_ui(UI_HDRS ${UIS})
 endif()
 
+if (ENABLE_QT_TRANSLATION)
```
lioncash: Is there a case where we would not want all languages to be provided by default? I'm fairly sure the expectation for people building the software is that their language translation gets included if they're not a native speaker/reader of English.

CMakeLists.txt
```diff
@@ -10,10 +10,15 @@ option(ENABLE_SDL2 "Enable the SDL2 frontend" ON)
 option(CITRA_USE_BUNDLED_SDL2 "Download bundled SDL2 binaries" OFF)
 
 option(ENABLE_QT "Enable the Qt frontend" ON)
+option(ENABLE_QT_TRANSLATION "Enable translation for the Qt frontend" OFF)
```
lioncash: `Enable translations ...`

CMakeLists.txt
```diff
@@ -10,10 +10,15 @@ option(ENABLE_SDL2 "Enable the SDL2 frontend" ON)
...
 option(CITRA_USE_BUNDLED_QT "Download bundled Qt binaries" OFF)
+option(GENERATE_QT_TRANSLATION "Generate en.ts as the translation source file" OFF)
 
 option(ENABLE_WEB_SERVICE "Enable web services (telemetry, etc.)" ON)
 option(CITRA_USE_BUNDLED_CURL "FOR MINGW ONLY: Download curl configured against winssl instead of openssl" OFF)
+
+set(CITRA_QT_LANGUAGES "${CMAKE_SOURCE_DIR}/dist/languages" CACHE PATH "Path to the translation bundle for the Qt frontend")
```
lioncash: Why is this in the top-level CMakeLists?

CMakeLists.txt
```diff
@@ -10,10 +10,15 @@ option(ENABLE_SDL2 "Enable the SDL2 frontend" ON)
 option(CITRA_USE_BUNDLED_SDL2 "Download bundled SDL2 binaries" OFF)
 
 option(ENABLE_QT "Enable the Qt frontend" ON)
+option(ENABLE_QT_TRANSLATION "Enable translation for the Qt frontend" OFF)
 option(CITRA_USE_BUNDLED_QT "Download bundled Qt binaries" OFF)
+option(GENERATE_QT_TRANSLATION "Generate en.ts as the translation source file" OFF)
```
lioncash: What would be the use case of not wanting to use English as the base translation source file for all the others, considering the strings are English by default in the UI?

src/citra_qt/CMakeLists.txt
wwylele: Because this requires an extra dependency (Qt Linguist Tools). For developers that want to test other parts of the program and don't care about the translation when testing, they can turn off this option and don't download the dependency.

CMakeLists.txt
wwylele: No there isn't such use case. I probably worded it poorly here. This option here just means whether to generate the base file or not, and it is hard-coded to English. I make it an option because 
 - the generated file is not in the build directory despite being a build artifact. Some people might not like polluting the directory with artifact.
 - It is not a dependency to build citra itself.

CMakeLists.txt
wwylele: Because I am bad and forget I can put them in subdirectories :p 

Should I put all new added options to citra-qt since they only make sense when Qt frontend is enabled?

wwylele:
image

Addressed comments, and moved the language configuration to the configuration dialog. Also added the ability to detect system language.

MerryMage:
I'm fine with this as-is. I would suggest pushing to a branch of this repo first (not the master branch) to verify that the transifex travis script works. My gut tells me it probably won't first try.

wwylele:
I tested the branch in this repo: https://github.com/citra-emu/citra/tree/translation-tx-test . To test it, I also added my transifex token to the CI environment var.

With current commits, the transifex push job is skipped as expected because the unmatched branch name. Then I added a new commit to change the branch name in travis script to "ttranslation-tx-test", it executed the job successfully and I get the push notification on transifex. So I think everything is right.

chris062689:
What would the process be for including in canary? Tagging, changing the branch name, and securing the secret token in Appveyor / Travis?

wwylele:
@chris062689 There is nothing to do for canary: for the two CI changes made in this PR:
 - transifex push: This is designed only to happen on the main repo master branch. It is coded as only happens on a branch named "master", and in the repo that has the token. So for nightly/canary/any other forks, they don't need, and shouldn't include the token in CI
 - build with translation: This has nothing to do with transifex, because it just uses the translation file stored in the repo. No special care needed for any repo.

baka0815:
Is it possible to add comments to the translatable strings to know where in the code they are from?
Currently it's not always clear if they are place-holder values or strings to be translated.

"ConfigureInput" for example is either a spelling error or some part which shouldn't be translated.
"free" is another one unclear where to be found and I'm therefore unable to translate it without context.

MerryMage:
Qt allows one to add translator comments: https://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments

.travis.yml
```diff
@@ -21,6 +21,22 @@ matrix:
       install: "./.travis/linux/deps.sh"
       script: "./.travis/linux/build.sh"
       after_success: "./.travis/linux/upload.sh"
+    - if: branch = master AND type = push
```
jroweboy: This will trigger on nightly push as well as some unofficial builds. I'm guessing it won't be an issue because they won't have the secret key though right?

src/citra_qt/configuration/configure_general.cpp
```diff
@@ -12,6 +13,23 @@ ConfigureGeneral::ConfigureGeneral(QWidget* parent)
...
+    QDirIterator it(":/languages", QDirIterator::NoIteratorFlags);
+    while (it.hasNext()) {
+        QString locale = it.next();
+        locale.truncate(locale.lastIndexOf('.'));
+        locale.remove(0, locale.lastIndexOf('/') + 1);
+        QString lang = QLocale::languageToString(QLocale(locale).language());
+        ui->language_combobox->addItem(lang, locale);
+    }
+
+    // Unlike other configuration, interface language change need to be reflect on the interface
```
jroweboy: // Unlike other configuration changes, interface language changes need to be reflected

src/citra_qt/configuration/configure_general.cpp
```diff
@@ -12,6 +13,23 @@ ConfigureGeneral::ConfigureGeneral(QWidget* parent)
...
+    // immediately. This is done by passing a signal to the main window, and then retranslating when
+    // passing back.
+    connect(ui->language_combobox,
+            static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this,
```
jroweboy: https://doc-snapshots.qt.io/qt5-dev/qtglobal.html#qOverload

`qOverload<int>(&QComboBox::currentIndexChanged)`

src/citra_qt/configuration/configure_general.cpp
```diff
@@ -52,3 +72,14 @@ void ConfigureGeneral::applyConfiguration() {
     Settings::values.use_cpu_jit = ui->toggle_cpu_jit->isChecked();
     Settings::Apply();
 }
+
+void ConfigureGeneral::onLanguageChanged(int index) {
```
jroweboy: this should be a lambda in the connect function since its a private slot and really short imo

wwylele:
In addition to what MerryMage says, transifex shows the text context and source file location on the web page for each text entry (which comes from parsing the Qt context strings), in the "More info" box below the translation editor box. Maintainers can also upload images as context reference to transifex if needed

.travis.yml
wwylele: Yes, when a fork doesn't have the key, the job will be executed and fails with no permission error.

src/citra_qt/configuration/configure_general.cpp
wwylele: The version requirement for this is Qt 5.7, which is beyond our current version requirement (5.0 or 5.2? I can't recall)

wwylele:
> who is running the transifex project? Who does new language approvals? Who accepts people to join the project? What will we do with an unmaintained language?

Yeah， these questions are to be discussed.  At the beginning I might be the "who" to do these (sorry!). Though these are completely transifex side and won't affect the code on github. Want to talk about this on irc/Discord later?

> Also this will likely cause lots of problems with the multiplayer UI as they both change citra_qt/cmakelists.txt and citra_qt/main.cpp Shouldn't be too hard to fix, but before we decide to merge one or the other we'll want to be careful. 

I skimmed over the code change in the both PRs, and I think that conflicts, if any, would be trivial to fix. The only flaw might occur is that the translation configuration cannot reflect on the multiplayer UI immediately, but I guess that's something we can fix later after merging both PRs.

> Also the multiplayer UI adds a bunch of new strings to translate. Should we just wait for the PR to be merged before pushing the strings to transifex then? (That sounds good to me)

The transifex push on every master branch (after merging this PR, of course), so it doesn't matter whether multiplayer UI PR gets merged first or after. The string will be automatically updated to transifex. However, if you want to have the translation for multiplayer UI in canary in advance, I need to do some temporary workaround instead, like removing the token from main repo and give it to the canary repo, so that new strings to canary are also pushed to transifex.

baka0815:
@wwylele sorry, I didn't see the "more info" panel. That helps a lot translating, thanks!

src/citra_qt/configuration/configure_general.cpp
wwylele: I am not sure about this. The function itself is still three lines, and the corresponding connection also takes three lines. Putting them together using lambda would make it really long

wwylele:
Added a blurb in main readme as @jroweboy requested.

wwylele:
I am going to merge this by the weekend if no additional discussion.

PR #3301, opened 2017-12-15
Kernel/SVC: Partially implemented svcExitProcess.

Subv:
Terminating processes with ready threads is not currently implemented and will assert. It is currently unknown how the 3DS kernel stops ready threads or threads running in another core.

Please note that this is still not enough to allow applets to exit. Some APT functions still need to be implemented, and the GSP module needs to be changed to account for multiple clients.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3301)

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -143,6 +143,36 @@ static ResultCode ControlMemory(u32* out_addr, u32 operation, u32 addr0, u32 add
     return RESULT_SUCCESS;
 }
 
+static void ExitProcess() {
+    LOG_INFO(Kernel_SVC, "Process %u exiting", g_current_process->process_id);
+
+    ASSERT_MSG(g_current_process->status == ProcessStatus::Running, "Process has already exited");
+
+    g_current_process->status = ProcessStatus::Exited;
+
+    // Stop all the process threads that are currently waiting for objects.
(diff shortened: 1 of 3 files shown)


src/core/hle/kernel/svc.cpp
```diff
@@ -143,6 +143,36 @@ static ResultCode ControlMemory(u32* out_addr, u32 operation, u32 addr0, u32 add
...
+    for (auto& thread : thread_list) {
+        if (thread->owner_process != g_current_process)
+            continue;
+
+        if (thread == GetCurrentThread())
+            continue;
+
+        // TODO(Subv): When are the other running/ready threads terminated?
+        ASSERT_MSG(thread->status == THREADSTATUS_WAIT_SYNCH_ANY ||
+                       thread->status == THREADSTATUS_WAIT_SYNCH_ALL,
```
wwylele: Should we also allow `WAIT_ARB`, `WAIT_SLEEP` here?

Subv: threads waiting on ArbitrateAddress do some extra processing after waking up (removing them from the arbiter's waitlist) and i'm not sure what happens with SleepThread yet, so i decided against allowing them until more research is done

purpasmart96:
Yeah, there is currently a whole bunch of apt function that need implementing...

PR #3302, opened 2017-12-15
citra_qt: Add console model setting

valentinvanelslande:
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3302)

### DIFF
--- a/src/citra_qt/configuration/configure_system.ui
+++ b/src/citra_qt/configuration/configure_system.ui
@@ -251,7 +251,7 @@
          </widget>
         </item>
         <item row="4" column="1">
-         <widget class="QComboBox" name="comboBox">
+         <widget class="QComboBox" name="combo_model">
           <item>
            <property name="text">
             <string>Nintendo 3DS</string>
--- a/src/citra_qt/configuration/configure_system.ui
+++ b/src/citra_qt/configuration/configure_system.ui
@@ -219,6 +219,9 @@
         </item>
         <item row="4" column="1">
          <widget class="QComboBox" name="combo_model">
+          <property name="toolTip">
+           <string>Changing this does not improve the speed.</string>
+          </property>
           <item>
            <property name="text">
             <string>Nintendo 3DS</string>
--- a/src/citra_qt/configuration/configure_system.ui
+++ b/src/citra_qt/configuration/configure_system.ui
@@ -22,11 +22,85 @@
         <string>System Settings</string>
        </property>
        <layout class="QGridLayout" name="gridLayout">
+        <item row="1" column="0">
+         <widget class="QLabel" name="label_birthday">
+          <property name="text">
+           <string>Birthday</string>
+          </property>
+         </widget>
+        </item>
         <item row="0" column="0">
          <widget class="QLabel" name="label_username">
           <property name="text">
            <string>Username</string>
           </property>
+         </widget>
+        </item>
+        <item row="2" column="1">
+         <widget class="QComboBox" name="combo_language">
+          <property name="toolTip">
+           <string>Note: this can be overridden when region setting is auto-select</string>
+          </property>
+          <item>
+           <property name="text">
+            <string>Japanese (日本語)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>English</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>French (français)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>German (Deutsch)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Italian (italiano)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Spanish (español)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Simplified Chinese (简体中文)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Korean (한국어)</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>Dutch (Nederlands)</string>
+           </property>
+          </item>
+          <item>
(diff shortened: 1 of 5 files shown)


src/core/hle/service/cfg/cfg.h
```diff
@@ -137,6 +138,20 @@ void GenHashConsoleUnique(Service::Interface* self);
...
+ *      Model of the console
+ */
+void SetSystemModel(u32 model);
+
+/**
+ * CFG::GetSystemModelID service function
+ * Output:
+ *      Model of the console
+ */
+u32 GetSystemModelID();
```
wwylele: These two are not service functions. Please use standard function doc string for them instead. You can refer to those `SetUsername` and  `GetUsername` functions below.

Also, aren't these two functions supposed to accept/return `SystemModel` type?

valentinvanelslande:
model

valentinvanelslande:
better I do not change the type

src/citra_qt/configuration/configure_system.ui
```diff
@@ -219,6 +219,9 @@
         </item>
         <item row="4" column="1">
          <widget class="QComboBox" name="combo_model">
+          <property name="toolTip">
+           <string>Changing this does not improve the speed.</string>
+          </property>
```
FearlessTobi: You should also mention that it doesn't improve game compatibility or adds features.

valentinvanelslande:
[   1.960537] Service.CFG <Error> core/hle/service/cfg/cfg.cpp:GetConfigInfoBlockPointer:331: Config block 0xF0004 with flags 8 and size 4 was not found
[   1.961213] Service.CFG <Debug> core/hle/service/cfg/cfg.cpp:GetSystemModelID:232: called model=289712712
WTF, changed the size?

valentinvanelslande:
changes the size of the block :(
how i can fix that?

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -221,6 +221,18 @@ void GetRegionCanadaUSA(Service::Interface* self) {
     }
 }
 
+void SetSystemModel(u32 model) {
+    LOG_DEBUG(Service_CFG, "called model=%u", model);
+    SetConfigInfoBlock(ConsoleModelBlockID, 4, 0x8, reinterpret_cast<u8*>(&model));
```
wwylele: This conflicts with [this default definition of the ConsoleModelBlock](https://github.com/citra-emu/citra/blob/80025792d60a76b1c73a46ec2cf073897e3941cd/src/core/hle/service/cfg/cfg.cpp#L520). (See also [here](https://github.com/citra-emu/citra/blob/80025792d60a76b1c73a46ec2cf073897e3941cd/src/core/hle/service/cfg/cfg.cpp#L82) for the struct defination). Only the first byte is the model ID.

src/core/hle/service/cfg/cfg.h
```diff
@@ -137,6 +138,18 @@ void GenHashConsoleUnique(Service::Interface* self);
...
+ * Sets the model in config savegame.
+ * @param model the model to set.
+ */
+void SetSystemModel(u32 model);
+
+/**
+ * Gets the model from config savegame.
+ * @returns the model
+ */
+u32 GetSystemModelID();
```
wwylele: To pair with `SetSystemModel`, you probably want to omit `ID` in this function name.

src/citra_qt/configuration/configure_system.ui
```diff
@@ -22,13 +22,87 @@
         <string>System Settings</string>
        </property>
        <layout class="QGridLayout" name="gridLayout">
+        <item row="1" column="0">
```
wwylele: The change of this file looks excessive. It looks like the layout items are shuffled some how. Could you fix it?

wwylele:
>changes the size of the block :(
how i can fix that?

What do you mean?

valentinvanelslande:
@wwylele in my unoficial build PTM, the Kernel, etc. read the model to support New 3DS games, leave errors that the block does not exist and the model change is temporary, it is lost after closing Citra

src/citra_qt/configuration/configure_system.ui
valentinvanelslande: fixed

src/core/hle/service/cfg/cfg.cpp
```diff
@@ -221,6 +221,18 @@ void GetRegionCanadaUSA(Service::Interface* self) {
     }
 }
 
+void SetSystemModel(ConsoleModelInfo model) {
+    LOG_DEBUG(Service_CFG, "called model=%u", model.model);
+    SetConfigInfoBlock(ConsoleModelBlockID, sizeof(model), 0x8, &model);
+}
+
+u32 GetSystemModel() {
+    u32 model;
```
wwylele: Same type issue here. You should use `ConsoleModelInfo` to receive the data

wwylele:
I partially agree with @B3n30 that this feature is rather misleading than useful before we fully implement all new 3DS mode. However, I do find changing model can be useful for homebrew testing or other kind of experiment. But again, this kind of usefulness is only after we actually implement it.

jroweboy:
I would be more than fine if it was an ini only setting, so homebrew developers can use it for testing. I also like the fact that this PR tries to add the service check for it. It would be nice if that was fleshed out a little better. So, I'm in favor of changing the scope of the PR instead of flat out closing it, if @valentinvanelslande is fine with that

wwylele:
@jroweboy The problem about ini is that, the field this PR trying to configure is stored in the config savegame, not in ini. Adding an overriding ayer of ini configuration is doable, but I guess this PR has to be reworked :p

valentinvanelslande:
How do I fix the errors?
When someone tells me how, I'll make PTM CheckNew3DS function return true if the model is n2ds/n3ds, or false if not

valentinvanelslande:
Closing because the value of New 2DS XL was added and the build fails.

PR #3307, opened 2017-12-16
HLE: Convert GSP_GPU to ServiceFramework.

Subv:
The only functional change is the error handling of GSP_GPU::ReadHWRegs function. We previously didn't return error codes (not even for success). The new returns were found by reverse engineering the GSP module.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3307)


src/core/hle/service/gsp/gsp_gpu.cpp
```diff
@@ -300,115 +275,77 @@ ResultCode SetBufferSwap(u32 screen_id, const FrameBufferInfo& info) {
...
     }
-    cmd_buff[2] = g_thread_id++;                                           // Thread ID
-    cmd_buff[4] = Kernel::g_handle_table.Create(g_shared_memory).Unwrap(); // GSP shared memory
 
-    g_interrupt_event->Signal(); // TODO(bunnei): Is this correct?
+    rb.Push(thread_id);
+    rb.PushCopyObjects(shared_memory);
+
+    thread_id++;
+    interrupt_event->Signal(); // TODO(bunnei): Is this correct?
```
Subv: hint: it's not

B3n30: Then why is this here?

src/core/hle/service/gsp/gsp_gpu.cpp
```diff
@@ -181,110 +180,86 @@ static ResultCode WriteHWRegsWithMask(u32 base_address, u32 size_in_bytes, VAddr
...
 
-/// Read a GSP GPU hardware register
-static void ReadHWRegs(Interface* self) {
-    u32* cmd_buff = Kernel::GetCommandBuffer();
-    u32 reg_addr = cmd_buff[1];
-    u32 size = cmd_buff[2];
+void GSP_GPU::ReadHWRegs(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp(ctx, 0x4, 2, 0);
+    u32 reg_addr = rp.Pop<u32>();
+    u32 input_size = rp.Pop<u32>();
```
wwylele: should this be named `output_size` instead?

src/core/hle/service/gsp/gsp_gpu.cpp
Subv: It's been there since the beginning, I'll probably make another PR later to correct some things about our GSP implementation

src/core/hle/service/gsp/gsp_gpu.cpp
```diff
@@ -671,133 +582,120 @@ static void ImportDisplayCaptureInfo(Interface* self) {
     FrameBufferUpdate* top_screen = GetFrameBufferInfo(thread_id, 0);
     FrameBufferUpdate* bottom_screen = GetFrameBufferInfo(thread_id, 1);
 
-    cmd_buff[0] = IPC::MakeHeader(0x18, 0x9, 0);
-    cmd_buff[1] = RESULT_SUCCESS.raw;
+    struct CaptureInfoEntry {
+        u32 address_left;
+        u32 address_right;
+        u32 format;
+        u32 stride;
```
wwylele: all these should be `u32_le`

src/core/hle/service/gsp/gsp.h
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Signals that the specified interrupt type has occurred to userland code
+ * @param interrupt_id ID of interrupt that is being signalled
+ */
+void SignalInterrupt(InterruptId interrupt_id);
+
+void InstallInterfaces(SM::ServiceManager& service_manager);
+} // namespace GSP
+} // namespace Service
```
wwylele: New line at the end of the file

src/core/hle/service/gsp/gsp_gpu.cpp
Subv: I named it input as in it's an input from the service function, the real size is clamped to 0x80, should i perform the clamping directly when calling Pop and just name it size?

wwylele: Ah in this case it is fine to me. No need to change it.

src/core/hle/service/gsp/gsp.h
Subv: done

Subv:
@wwylele Yes the idea is to move the gsplcd service into the subfolder as well, but this PR is already quite a big refactor so i thought of doing it in another one

purpasmart96:
LGTM

PR #3309, opened 2017-12-16
ldr_ro: use ServiceFramework's session slot for client slot

wwylele:
Make use of #3284. This is closer to the real RO service as it use session handle as key to the slot, not the process handle/ID.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3309)


wwylele:
@MerryMage the code originally came from real RO service which also check `loaded_crs == 0` in the code. I think keeping this can be a reference for others who look into RO service code.

MerryMage:
Thanks, I thought that was the case, just wanted to confirm.

PR #3315, opened 2017-12-19
[RFC] Add framework for save states

B3n30:
This adds a basic framework for saving and loading save states.
It adds an additional dependency to serialize the data: https://uscilab.github.io/cereal/index.html
cereal supports almost all types available in c++14 and c++17 types are currently under development.
The only thing cereal doesn't support are raw pointers.

The first commit is the framework, the second commit is just an example on how to save one part of citra. In this case I choose CoreTiming. If this is going to get merged I'll split it into two PRs.

The main goal is tho have the framework and then adding the serialization of different parts in small PRs. AS soon as every part is serialized we can add the UI part to actually use this feature.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3315)

### DIFF
--- a/src/core/save_state.cpp
+++ b/src/core/save_state.cpp
@@ -1,0 +1,41 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <fstream>
+#include <cereal/archives/portable_binary.hpp>
+#include "common/logging/log.h"
+#include "core/core.h"
+#include "core/save_state.h"
+
+namespace State {
+
+static void WriteHeader(std::ofstream& os, StateHeader header) {
+    os << header.version;
+}
+
+static StateHeader ReadHeader(std::ifstream& is) {
+    StateHeader header;
+    is >> header.version;
+    return header;
+}
+
+void SaveState(std::ofstream os) {
+    WriteHeader(os, {SAVE_STATE_VERSION});
+    cereal::PortableBinaryOutputArchive archive(os);
+    Core::System::GetInstance().SerializeState(archive);
+}
+
+LoadStateError LoadState(std::ifstream is) {
+    StateHeader header = ReadHeader(is);
+    if (header.version != SAVE_STATE_VERSION) {
+        LOG_ERROR(Core, "Wrong version of save state: version=%d", header.version);
+        return LoadStateError::IncorrectVersion;
+    }
+
+    cereal::PortableBinaryInputArchive archive(is);
+    Core::System::GetInstance().SerializeState(archive);
+    return LoadStateError::None;
+}
+
+} // namespace State
--- a/src/core/save_state.h
+++ b/src/core/save_state.h
@@ -1,0 +1,27 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <fstream>
+#include "common/common_types.h"
+
+namespace State {
+
+constexpr u32 SAVE_STATE_VERSION = 1;
+
+struct StateHeader {
+    u32 version;
+};
+
+void SaveState(std::ofstream);
+
+enum class LoadStateError {
+    None,
+    IncorrectVersion,
+};
+
+LoadStateError LoadState(std::ifstream);
+
+} // namespace State
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,6 @@
 [submodule "inih"]
     path = externals/inih/inih
     url = https://github.com/benhoyt/inih.git
+[submodule "externals/cereal"]
+    path = externals/cereal
+    url = https://github.com/USCiLab/cereal
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -7,6 +7,10 @@ include(DownloadExternals)
 add_library(catch-single-include INTERFACE)
 target_include_directories(catch-single-include INTERFACE catch/single_include)
 
+# cereal
+add_library(cereal INTERFACE)
+target_include_directories(cereal INTERFACE cereal/include)
+
 # Crypto++
 add_subdirectory(cryptopp)
 
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -10,6 +10,7 @@
 #include "common/string_util.h"
 #include "core/3ds.h"
 #include "core/core.h"
+#include "core/save_state.h"
 #include "core/settings.h"
 #include "input_common/keyboard.h"
 #include "input_common/main.h"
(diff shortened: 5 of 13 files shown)


MerryMage:
Only thing to note about choice of library: [hidden state would require explicit template specialisation](https://uscilab.github.io/cereal/pimpl.html).

Currently save_state is a bit too inflexible for my liking, I'll write some suggestions when I have time. The frontend should have more control.

Regarding the header: Version should come first.

Otherwise, fine.

src/core/save_state.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "core/save_state.h"
+
+namespace State {
+
+enum SaveStateAction : u8 {
+    None,
+    Save,
+    Load,
+};
+static std::atomic<SaveStateAction> save_state_action;
```
lioncash: Can we *please* introduce this in a way that doesn't introduce file-static or global state from the get-go? This is totally a doable thing, given the `System` class exists (and we're essentially storing its state)

src/core/save_state.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+void SaveState() {
+    std::ofstream os(CreateFileName(), std::ofstream::binary);
+    WriteHeader(os);
+    cereal::PortableBinaryOutputArchive archive(os);
+    DoState(archive);
+}
+
+void ScheduleLoad(u8 slot) {
+    if (save_state_action == SaveStateAction::None) {
```
lioncash:
```cpp
if (save_state_action != SaveStateAction::None) {
    return;
}
```

src/core/save_state.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+void LoadState() {
+    std::ifstream is(CreateFileName(), std::ifstream::binary);
+    StateHeader header = ReadHeader(is);
+    if (header.version != SAVE_STATE_VERSION) {
+        LOG_ERROR(Core, "Wrong version of save state: slot=%d, version=%d", current_slot,
+                  header.version);
+        return;
+    } else if (header.title_id != current_title_id) {
```
lioncash: `else` isn't needed here, since the above if has a return in it.

src/core/save_state.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+void SaveState() {
+    std::ofstream os(CreateFileName(), std::ofstream::binary);
+    WriteHeader(os);
+    cereal::PortableBinaryOutputArchive archive(os);
+    DoState(archive);
+}
+
+void ScheduleLoad(u8 slot) {
```
lioncash: this and the slot storage variable should likely just be an `int`/`s32` or a `u32`. This will cause more truncation warnings than it's likely worth.

src/core/save_state.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+void Init() {
+    save_state_action = SaveStateAction::None;
+    current_slot = 1;
+    current_title_id = 0;
+}
+
+void ScheduleSave(u8 slot) {
+    if (save_state_action == SaveStateAction::None) {
```
lioncash:
```cpp
if (save_state_action == SaveStateAction::None) {
    return;
}
```

MerryMage:
Pushed changes.

imo if there is effort going towards making savestates happen, some cleaning up of global state should be made at the same time.

I've not done that in the above; but ideally the globals in CoreTiming should be removed at the same time.

B3n30:
@bunnei suggested that we could also use the 3ds's sleep mode to implement save states, which would allow for save states without touching every part of citra and adding some serialization code. But It will be a bit more limited. 

On the other hand touching every part of citra and removing globals and file statics as much as possible is probably not bad, too. Also for debugging purposes a sleep mode proposed in this PR is probably better.

src/core/save_state.h
```diff
@@ -0,0 +1,27 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <fstream>
+#include "common/common_types.h"
+
+namespace State {
+
+constexpr u32 SAVE_STATE_VERSION = 1;
```
jroweboy: Is it better to have a global save state version instead of a local version? Reading here: http://uscilab.github.io/cereal/serialization_functions.html we can actually have a version serialized class, which would let us support upgrades. if you have to add something new to the class that needs some computation (ie something that isn't something you can just default init) then you can add an if inside the serializer for that class to generate a default one.

This approach has some limitations of course since you can't expect every version change to be compatible, so we'd need some way to check for errors if the (de)serialization failed, but it let the programmers add an upgrade path in the case that a class changed and it would affect the save state. (Also since it defaults to 0 for the version number, it won't add any code to the save and load functions except when its time to bump the version for a class)

B3n30:
For me save-states are usually a temporary save not something that is used for month's or years. Since TAS is usually also bound to one version of the emulator, I don't see any issue with a global version for save states.

Also I think save states at the current stage of development of citra are more a debug feature then a user feature (but of course also usable for users). I was thinking about using cereal to not save the state in binary but in xml or similar. So that you can manually change the save state and investigate issues that way.

adityaruplaha:
Hmmmmmm, this is a highly requested feature. Also, as a personal note, its reaaalllyyy helpful in PSMD. Also, most people use save states as temporary saves (like in middle of a very lengthy mission). I'm saying this from my own experience, so other views are welcome.

jroweboy:
Since my last comment, we spoke with neobrain about this on irc, and he suggested that save states should only be tied to one particular version of citra, which means we can and probably should use the git hash as the version. From a UX standpoint, I think a popup warning if the save state version doesn't match should be fine. Or maybe an ini setting that would allow the user to load different versions of save states or something. At any rate, we shouldn't be encouraging people to save state across versions of citra.

(Maybe we should make save states the final 1.0 feature so people can have a stable version?)

MerryMage:
> Since my last comment, we spoke with neobrain about this on irc, and he suggested that save states should only be tied to one particular version of citra, which means we can and probably should use the git hash as the version.

This sounds reasonable to me. We should encourage save states for short-term saves (on the order of minutes to single digit days), or for bug reports. An adjuvant for in-game saving, not to replace them. 

Pop-up is a 👍 . ini setting is a maybe (command-line flag would be more appropriate?), not having it wouldn't be a blocker.

> (Maybe we should make save states the final 1.0 feature so people can have a stable version?)

1.0 hype.

MerryMage:
I would like to see this merged at some point this year if at all feasible.

afaik remaining issues are as follows:

1. Having current commit-hash in the save-state (and verifying this):
    This should hopefully be straightforward.

2. Global vs Local version numbers:
    I'm personally not convinced there is much value in local version numbers since we're intending save-states to be primarily used with a single commit-version, which makes the issue kind of moot.
    Why should the global version number exist, then? A global version number is useful in case we do decide to change policy and have backwards-compatible savestates in a future version, we can easily reject earlier savestates that don't conform to that (or perform conversion for those), for example. Or, if we do decided to go with local version numbers in the future, we can increment the global version. Of, if we decide to switch to an entirely new format.

3. Having a command-line option for force loading saves of the incorrect version:
    imo not a blocker

jroweboy:
@B3n30 I think we've had all the discussion on the matter that will be had at this stage. When you get some time, please update this to add the current commit hash as the version and validate it and lets get this merged 👍 

I forgot to link the conversation about saves states that I mentioned in my previous comment, so here it is https://pastebin.com/DwnLV7hM

jroweboy:
Its been two months since the last update. I really would like this, but its not on the top of my priority list. If you would be so kind to reopen this when you are ready to work on it again ❤️

PR #3324, opened 2017-12-21
Services: Keep track of per-session data in GSPGPU

Subv:
Only signal the GSP interrupts for the currently active thread (set by GSP_GPU::AcquireRight).

This PR makes it so that applets won't mess up the GSP state for the running application via UnregisterInterruptRelayQueue after they're closed.

Full disclosure: I don't know if this will break any games that use the GSP in funny ways (acquiring the gsp right multiple times without releasing it, other weird but probably valid behaviors). I don't have a way to verify this since my IDB for the GSP module was recently corrupted resulting in a total data loss.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3324)

### DIFF
--- a/src/core/hle/service/gsp/gsp_gpu.cpp
+++ b/src/core/hle/service/gsp/gsp_gpu.cpp
@@ -370,9 +370,6 @@ void GSP_GPU::UnregisterInterruptRelayQueue(Kernel::HLERequestContext& ctx) {
  * @todo This probably does not belong in the GSP module, instead move to video_core
  */
 void GSP_GPU::SignalInterrupt(InterruptId interrupt_id) {
-    if (!gpu_right_acquired) {
-        return;
-    }
     if (nullptr == shared_memory) {
         LOG_WARNING(Service_GSP, "cannot synchronize until GSP shared memory has been created!");
         return;
@@ -634,18 +631,26 @@ void GSP_GPU::AcquireRight(Kernel::HLERequestContext& ctx) {
     u32 flag = rp.Pop<u32>();
     auto process = rp.PopObject<Kernel::Process>();
 
-    gpu_right_acquired = true;
+    // TODO(Subv): This case should put the caller thread to sleep until the right is released.
+    ASSERT_MSG(active_thread_id == -1, "GPU right has already been acquired");
+
+    SessionData* session_data = GetSessionData(ctx.Session());
+    active_thread_id = session_data->thread_id;
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
     rb.Push(RESULT_SUCCESS);
 
-    LOG_WARNING(Service_GSP, "called flag=%08X process=%u", flag, process->process_id);
+    LOG_WARNING(Service_GSP, "called flag=%08X process=%u thread_id=%u", flag, process->process_id,
+                active_thread_id);
 }
 
 void GSP_GPU::ReleaseRight(Kernel::HLERequestContext& ctx) {
     IPC::RequestParser rp(ctx, 0x17, 0, 0);
 
-    gpu_right_acquired = false;
+    SessionData* session_data = GetSessionData(ctx.Session());
+    ASSERT_MSG(active_thread_id == session_data->thread_id,
+               "Wrong thread tried to release GPU right");
+    active_thread_id = -1;
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
     rb.Push(RESULT_SUCCESS);
@@ -669,7 +674,7 @@ void GSP_GPU::StoreDataCache(Kernel::HLERequestContext& ctx) {
 
 SessionData* GSP_GPU::FindRegisteredThreadData(u32 thread_id) {
     for (auto& session_info : connected_sessions) {
-        SessionData* data = static_cast<SessionData*>(session_info.data.get())
+        SessionData* data = static_cast<SessionData*>(session_info.data.get());
         if (!data->registered)
             continue;
         if (data->thread_id == thread_id)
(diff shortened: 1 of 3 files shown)


src/core/hle/service/gsp/gsp_gpu.h
```diff
@@ -351,14 +360,18 @@ class GSP_GPU final : public ServiceFramework<GSP_GPU> {
      */
     void StoreDataCache(Kernel::HLERequestContext& ctx);
 
-    /// Event triggered when GSP interrupt has been signalled
-    Kernel::SharedPtr<Kernel::Event> interrupt_event;
-    /// GSP shared memoryings
+    /// Returns the session data for the specified registered thread id, or nullptr if not found.
+    SessionData* FindRegisteredThreadData(u32 thread_id);
```
lioncash: Shouldn't this use an optional<SessionData> as the return type and use an empty optional to signify the "not found" case?

MerryMage: imo SessionData* is carries the same meaning as optional<SessionData&>, but since std::optional doesn't support ref-optionals...

LogPod:
Home Menu doesn't boot now 
[   3.632565] Debug <Critical> core/hle/service/gsp/gsp_gpu.cpp:operator():654: Assertion Failed!
GPU right has already been acquired****

wwylele:
@LogPod could you test against the latest update of this PR? Because there was a fix to this just now. The canary is probably not updated yet, in which case you need to build citra yourself.

src/core/hle/service/gsp/gsp_gpu.cpp
```diff
@@ -622,18 +650,34 @@ void GSP_GPU::AcquireRight(Kernel::HLERequestContext& ctx) {
...
+
+    LOG_WARNING(Service_GSP, "called flag=%08X process=%u thread_id=%u", flag, process->process_id,
+                session_data->thread_id);
 
     IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
-    rb.Push(RESULT_SUCCESS);
 
-    LOG_WARNING(Service_GSP, "called flag=%08X process=%u", flag, process->process_id);
+    if (active_thread_id == session_data->thread_id) {
+        rb.Push(ResultCode(ErrorDescription::AlreadyDone, ErrorModule::GSP, ErrorSummary::Success,
```
valentinvanelslande: GX

src/core/hle/service/gsp/gsp_gpu.h
```diff
@@ -179,7 +179,16 @@ struct CommandBuffer {
 };
 static_assert(sizeof(CommandBuffer) == 0x200, "CommandBuffer struct has incorrect size");
 
-class GSP_GPU final : public ServiceFramework<GSP_GPU> {
+struct SessionData : public Kernel::SessionRequestHandler::SessionDataBase {
```
wwylele: `SessionData` probably needs an explicit default destructor (`~SessionData();` here and `SessionData::~SessionData()=default;` in the cpp file), otherwise other compilation units have to generate default destructor for `Kernel::SharedPtr` with an incomplete type `Kernel::Event`, which might be the reason why msvc is screaming errors everywhere.

tabnk:
Dragon Quest VII: Fragments of the Forgotten Past Failed to run when this PR and #3281 are together

wwylele:
@tabnk could you test them separately (and make sure to use the last update)? The two PR has little influence on each other.

tabnk:
Tested https://ci.appveyor.com/api/buildjobs/wxcwwwyx1360llnc/artifacts/citra-windows-mingw-20180103-de85890.zip with Dragon Quest VII: Fragments of the Forgotten Past run fine. 

However, combine latest #3281 it just black screen.

wwylele:
@tabnk which build that combined this and #3281 you tested? If it was canary it probably merges an old version of this PR, which is known causing black screen of DQ VII

tabnk:
@jroweboy suggected that dragon quest issue is actually related to #3324 if you read here https://github.com/citra-emu/citra/pull/3281#issuecomment-355185833

wwylele:
@tabnk yes but what he said is the issue is related *only* to this PR (of probably an old version), not *combined*.

tabnk:
Based on canary-302 and merged with latest #3281.

wwylele:
@tabnk that canary is behind the last update of this PR. I triggered a new canary build just now. Please retest it when it comes out

tabnk:
Alright will test again when new build (canary-303) it out.

tabnk:
Tested (Canary-303) but run fine with Dragon Quest VII: Fragments of the Forgotten Past. :D

wwylele:
cleaned noise

src/core/hle/service/gsp/gsp_gpu.cpp
```diff
@@ -363,18 +378,32 @@ void GSP_GPU::UnregisterInterruptRelayQueue(Kernel::HLERequestContext& ctx) {
  * @todo This probably does not belong in the GSP module, instead move to video_core
  */
 void GSP_GPU::SignalInterrupt(InterruptId interrupt_id) {
-    if (!gpu_right_acquired) {
+    // Don't do anything if no process has acquired the GPU right.
```
wwylele: Should it still signal PDC0 and PDC1?

Subv: It didn't signal them if no process held the GPU right in the previous code.

neobrain:

2018-01-06T18:51:43Z: https://github.com/citra-emu/citra/compare/c3c684cd2b282e4aceb5feadf8434f104d02c821...Subv:34685f48dce91a12d0c7b0e014c6ce9ada7177e7

PR #3325, opened 2017-12-22
Add ability to send test cases through telemetry

BreadFish64:
Adds a UI dialog that allows users to submit test cases through the QT frontend using a telemetry field.
@chris062689 has plans to integrate it with the compatibility list.
Pretty Pictures:
image
image
image
image

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3325)


FearlessTobi:
I would prefer having the Next Button greyed out until something is selected, instead of another message box. I don't know if its worth the effort to implement though.

BreadFish64:
@FearlessTobi the buttons at the bottom are rather inflexible, I investigated making the radio buttons a mandatory field, but it involved making a subclass of QWizardPage, more trouble than it's worth

src/citra_qt/compatdb.h
```diff
@@ -0,0 +1,28 @@
...
+
+class CompatDB : public QWizard {
+    Q_OBJECT
+
+public:
+    explicit CompatDB(QWidget* parent = 0);
+    ~CompatDB();
+
+private:
+    Ui::CompatDB* ui;
```
lioncash: This can be `std::unique_ptr`, which would remove the need to explicitly `delete` this in the destructor.

src/citra_qt/compatdb.h
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Ui {
+class CompatDB;
+}
+
+class CompatDB : public QWizard {
+    Q_OBJECT
+
+public:
+    explicit CompatDB(QWidget* parent = 0);
```
lioncash: `= nullptr`

src/citra_qt/compatdb.h
```diff
@@ -0,0 +1,28 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#ifndef COMPATDB_H
```
lioncash: We use `#pragma once` as our header guard

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,56 @@
...
+        compatibility->addButton(ui->radioButton_Bad, 3);
+        compatibility->addButton(ui->radioButton_IntroMenu, 4);
+        compatibility->addButton(ui->radioButton_WontBoot, 5);
+        LOG_DEBUG(
+            Frontend,
+            tr("Compatibility Rating: %1").arg(compatibility->checkedId()).toStdString().c_str());
+        if (compatibility->checkedId() != -1) {
+            Core::Telemetry().AddField(Telemetry::FieldType::UserFeedback, "Compatibility",
+                                       compatibility->checkedId());
+            delete compatibility;
```
lioncash: Qt should handle the automatic deletion of a `QButtonGroup`, since the wizard is being passed as a parent to it. the delete here and below seem unnecessary.

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <QMessageBox>
+#include <QPushButton>
+#include "common/telemetry.h"
+#include "compatdb.h"
+#include "core/core.h"
+#include "ui_compatdb.h"
+
+CompatDB::CompatDB(QWidget* parent)
+    : QWizard(parent, Qt::WindowTitleHint | Qt::WindowCloseButtonHint | Qt::WindowSystemMenuHint),
+      ui(new Ui::CompatDB) {
```
lioncash: `ui{std::make_unique<Ui::CompatDB>()}`

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <QMessageBox>
+#include <QPushButton>
+#include "common/telemetry.h"
+#include "compatdb.h"
+#include "core/core.h"
+#include "ui_compatdb.h"
```
lioncash: `#include "citra_qt/ui_compatdb.h"`

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+CompatDB::CompatDB(QWidget* parent)
+    : QWizard(parent, Qt::WindowTitleHint | Qt::WindowCloseButtonHint | Qt::WindowSystemMenuHint),
+      ui(new Ui::CompatDB) {
+    ui->setupUi(this);
+    QList<QWizard::WizardButton> layout1;
+    layout1 << QWizard::Stretch << QWizard::BackButton << QWizard::CustomButton1
+            << QWizard::CancelButton << QWizard::FinishButton;
+    this->setButtonLayout(layout1);
```
lioncash: `this->` can be omitted here and below.

src/citra_qt/compatdb.ui
```diff
@@ -0,0 +1,215 @@
...
+    </item>
+    <item row="0" column="0" colspan="2">
+     <widget class="QLabel" name="lbl_Independent">
+      <property name="font">
+       <font>
+        <pointsize>10</pointsize>
+       </font>
+      </property>
+      <property name="text">
+       <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Independant of speed / performance, how well does the game play from start to finish on this version of Citra?&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
lioncash: `Independant of speed or performance, ...`

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright 2017 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <QMessageBox>
+#include <QPushButton>
+#include "common/telemetry.h"
+#include "compatdb.h"
```
lioncash: `#include "citra_qt/compatdb.h"`

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+CompatDB::CompatDB(QWidget* parent)
+    : QWizard(parent, Qt::WindowTitleHint | Qt::WindowCloseButtonHint | Qt::WindowSystemMenuHint),
+      ui(new Ui::CompatDB) {
+    ui->setupUi(this);
+    QList<QWizard::WizardButton> layout1;
+    layout1 << QWizard::Stretch << QWizard::BackButton << QWizard::CustomButton1
+            << QWizard::CancelButton << QWizard::FinishButton;
+    this->setButtonLayout(layout1);
+    this->setButtonText(QWizard::CustomButton1, tr("Next"));
```
lioncash: Why not just use `QWizard::NextButton` as opposed to `QWizard::CustomButton1` with `layout1` above? This would also alleviate the need for translators to translate one more string.

chris062689:
This functionality should be hidden from Canary builds, FYI. 
Only Nightly builds should have this option.

src/citra_qt/compatdb.cpp
BreadFish64: You sure about that? that file isn't in that directory

src/citra_qt/compatdb.cpp
BreadFish64: I guess you're probably more knowledgeable about it than I am, I just didn't know if would cause a memory leak if the user went back and forth between pages

src/citra_qt/compatdb.h
```diff
@@ -0,0 +1,27 @@
...
+
+class CompatDB : public QWizard {
+    Q_OBJECT
+
+public:
+    explicit CompatDB(QWidget* parent = nullptr);
+    ~CompatDB();
+
+private:
+    std::unique_ptr<Ui::CompatDB, std::default_delete<Ui::CompatDB>> ui;
```
lioncash: You don't need to specify default delete here. It's automatically chosen (hence the name signifying it's the default deleter ;))

src/citra_qt/compatdb.ui
```diff
@@ -0,0 +1,215 @@
...
+    </item>
+    <item row="0" column="0" colspan="2">
+     <widget class="QLabel" name="lbl_Independent">
+      <property name="font">
+       <font>
+        <pointsize>10</pointsize>
+       </font>
+      </property>
+      <property name="text">
+       <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Independant of speed or performance, how well does the game play from start to finish on this version of Citra?&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
lioncash: I missed this the first time around, but 'Independant" should be "Independent"

src/citra_qt/main.cpp
```diff
@@ -795,6 +799,7 @@ void GMainWindow::OnStartGame() {
 
     ui.action_Pause->setEnabled(true);
     ui.action_Stop->setEnabled(true);
+    ui.action_Report_Compatibility->setEnabled(true);
```
baka0815: Should probably only be enabled when the collection of telemetry-data is configured completely.

BreadFish64: Right, I'll make sure the user is verified

CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ option(ENABLE_QT "Enable the Qt frontend" ON)
 option(CITRA_USE_BUNDLED_QT "Download bundled Qt binaries" OFF)
 
 option(ENABLE_WEB_SERVICE "Enable web services (telemetry, etc.)" ON)
+option(ENABLE_COMPATIBILITY_REPORTING "Enable compatibility reporting" OFF)
```
lioncash: You likely want to prefix this with `CITRA_` so it doesn't potentially clash with an external in the future.

What is the benefit of making this toggleable, by the way?

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,52 @@
...
+        Core::Telemetry().AddField(Telemetry::FieldType::UserFeedback, "Compatibility",
+                                   compatibility->checkedId());
+        break;
+    }
+}
+void CompatDB::EnableNext() {
+    button(NextButton)->setEnabled(true);
+}
+
+CompatDB::~CompatDB() {}
```
lioncash: The expectation is a constructor is followed by a destructor if it's necessary to specify one, so this should be moved up. This can also be defaulted like so:

```cpp
CompatDB::~CompatDB() = default;
```

src/citra_qt/CMakeLists.txt
```diff
@@ -82,6 +82,16 @@ set(UIS
     main.ui
 )
 
+if (ENABLE_COMPATIBILITY_REPORTING)
+    set(citra-qt ${citra-qt}
```
lioncash: This can be replaced with generator expressions in the respective lists.

I'd also be surprised if this worked as intended on a clean source tree and reporting is enabled. `citra-qt` is a target, not a list that can be set.

e.g.

```cmake
set(UIS
     configuration/configure.ui
     configuration/configure_audio.ui
     configuration/configure_debug.ui
     configuration/configure_general.ui
     configuration/configure_graphics.ui
     configuration/configure_input.ui
     configuration/configure_system.ui
     configuration/configure_web.ui
     debugger/registers.ui
     aboutdialog.ui
     hotkeys.ui
     main.ui
     $<$<BOOL:${ENABLE_COMPATIBILITY_REPORTING}>:compatdb.ui>
)
```

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,52 @@
...
+    compatibility->addButton(ui->radioButton_Okay, 2);
+    compatibility->addButton(ui->radioButton_Bad, 3);
+    compatibility->addButton(ui->radioButton_IntroMenu, 4);
+    compatibility->addButton(ui->radioButton_WontBoot, 5);
+    switch (currentId()) {
+    case 1:
+        if (compatibility->checkedId() == -1) {
+            button(NextButton)->setEnabled(false);
+            break;
+        }
```
lioncash: You likely want a break here as well, otherwise this falls through to the case below when it's not `-1`. If the fallthrough is intended, it should probably be restructured so that it isn't.

CMakeLists.txt
chris062689: We want to specifically enable this flag for nightly, but not make it the default.
 
The reasoning behind this is this feature shouldn't be used by canary or any unofficial build. 
This is functionality specifically for nightly.

BreadFish64:
Oh, missed your comment before making changes, I'll get back to that

CMakeLists.txt
BreadFish64: ... which is why I set it as off by default

BreadFish64: ... which is why I set it as off by default

BreadFish64: @lioncash I just followed the pattern set by ENABLE_WEB_SERVICE but I can certainly change it

src/citra_qt/compatdb.cpp
BreadFish64: Woops, just put the break on the wrong side of the brackets, totally unintentional

src/citra_qt/main.ui
```diff
@@ -73,6 +73,7 @@
     <addaction name="action_Pause"/>
     <addaction name="action_Stop"/>
     <addaction name="separator"/>
+    <addaction name="action_Report_Compatibility"/>
```
j-selby: Not sure about the placement in the Emulation menu.

CMakeLists.txt
```diff
@@ -235,6 +236,9 @@ endif()
 if (ENABLE_WEB_SERVICE)
     add_definitions(-DENABLE_WEB_SERVICE)
 endif()
+if (CITRA_ENABLE_COMPATIBILITY_REPORTING)
+    add_definitions(-DCITRA_ENABLE_COMPATIBILITY_REPORTING)
```
j-selby: Does this need to be prefixed with `CITRA_`? See the above statement on the web service module for consistency.

chris062689: Lioncash commented up above that we should be prefixing CITRA_ to these.

src/citra_qt/main.ui
chris062689: I figure it's a good placement only because that's close to the Start / Stop options, where else would be a better place to put it?

BreadFish64: It's a place where users will see it, and since it's disabled when emulation isn't running, I think it makes the most sense

CMakeLists.txt
j-selby: Sure!

src/citra_qt/main.ui
j-selby: I would personally think that it would be better in the Help menu, where we can also have, say, direct links to the compat db, and where the other web service stuff is currently located. Just my 2c though.

LogPod:
I would personally like the Report Compatibility option below the Configure option because i pressed it accidentally so many time.

FearlessTobi:
Imo the compatibility report option doesn't fit into the Emulation submenu. The other entries all have to do with the Emulation in Citra itself. I would also prefer it in the Help Tab or at least below the configure button. Just my opinion.

neobrain:

2018-01-06T17:56:17Z: https://github.com/citra-emu/citra/compare/e203c10cc24ab57165e0b6f019e9b5552e5d5e49...BreadFish64:68959823e92260cfc095d27a2faebcc65676fc45

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,55 @@
...
+    switch (currentId()) {
+    case 1:
+        if (compatibility->checkedId() == -1) {
+            button(NextButton)->setEnabled(false);
+        }
+        break;
+    case 2:
+        LOG_DEBUG(
+            Frontend,
+            tr("Compatibility Rating: %1").arg(compatibility->checkedId()).toStdString().c_str());
```
wwylele: I don't think a logging message should be translated. Also, could just use LOG_* 's format string? I think our convention is, to use Qt's utility for user interface string, and to use logging formatter for log.

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,55 @@
...
+        Core::Telemetry().AddField(Telemetry::FieldType::UserFeedback, "Compatibility",
+                                   compatibility->checkedId());
+        // the frozen dependency Linux build does not support the "NoCancelButtonOnLastPage" option,
```
wwylele: Instead of mention the frozen dependency Linux build, which is specifically about our CI system, could just mention that old Qt library doesn't support it, "so this is a workaround for those who build citra using old a Qt library.

src/citra_qt/compatdb.cpp
```diff
@@ -39,17 +39,16 @@ void CompatDB::Submit() {
         }
         break;
     case 2:
-        LOG_DEBUG(
-            Frontend,
-            tr("Compatibility Rating: %1").arg(compatibility->checkedId()).toStdString().c_str());
+        LOG_DEBUG(Frontend, "Compatibility Rating: %1", compatibility->checkedId());
```
wwylele: `%1` should be `%d` for logging system

wwylele:
@chris062689 / @jroweboy would you like to give an additional review?

.travis/macos/build.sh
```diff
@@ -6,7 +6,7 @@ export MACOSX_DEPLOYMENT_TARGET=10.9
 export Qt5_DIR=$(brew --prefix)/opt/qt5
 
 mkdir build && cd build
-cmake .. -DUSE_SYSTEM_CURL=ON -DCMAKE_OSX_ARCHITECTURES="x86_64;x86_64h" -DCMAKE_BUILD_TYPE=Release
+cmake .. -DUSE_SYSTEM_CURL=ON -DCMAKE_OSX_ARCHITECTURES="x86_64;x86_64h" -DCMAKE_BUILD_TYPE=Release -DCITRA_ENABLE_COMPATIBILITY_REPORTING=${ENABLE_COMPATIBILITY_REPORTING:-"OFF"}
```
jroweboy: I see `-DCITRA_ENABLE_COMPATIBILITY_REPORTING=${ENABLE_COMPATIBILITY_REPORTING:-"OFF"}` but enable compat reporting isn't set for mac anywhere.

CMakeLists.txt
```diff
@@ -14,14 +14,14 @@ option(CITRA_USE_BUNDLED_QT "Download bundled Qt binaries" OFF)
...
-if (ENABLE_WEB_SERVICE AND CITRA_USE_BUNDLED_CURL AND WINDOWS AND MSVC)
-    message("Turning off use bundled curl as msvc can compile curl on cpr")
-    SET(CITRA_USE_BUNDLED_CURL OFF CACHE BOOL "" FORCE)
-endif()
 if (ENABLE_WEB_SERVICE AND NOT CITRA_USE_BUNDLED_CURL AND MINGW)
     message(AUTHOR_WARNING "Turning on CITRA_USE_BUNDLED_CURL. Override it only if you know what you are doing.")
     SET(CITRA_USE_BUNDLED_CURL ON CACHE BOOL "" FORCE)
 endif()
+if (CITRA_ENABLE_COMPATIBILITY_REPORTING AND NOT ENABLE_WEB_SERVICE)
+    message("Compatibility reporting requires web services to be enabled")
```
jroweboy: `message(AUTHOR_WARNING "..")` instead of default message type

CMakeLists.txt
```diff
@@ -14,14 +14,14 @@ option(CITRA_USE_BUNDLED_QT "Download bundled Qt binaries" OFF)
 
 option(ENABLE_WEB_SERVICE "Enable web services (telemetry, etc.)" ON)
 option(CITRA_USE_BUNDLED_CURL "FOR MINGW ONLY: Download curl configured against winssl instead of openssl" OFF)
-if (ENABLE_WEB_SERVICE AND CITRA_USE_BUNDLED_CURL AND WINDOWS AND MSVC)
```
jroweboy: Why was this removed?

src/citra_qt/compatdb.cpp
```diff
@@ -0,0 +1,54 @@
...
+void CompatDB::Submit() {
+    QButtonGroup* compatibility = new QButtonGroup(this);
+    compatibility->addButton(ui->radioButton_Perfect, 0);
+    compatibility->addButton(ui->radioButton_Great, 1);
+    compatibility->addButton(ui->radioButton_Okay, 2);
+    compatibility->addButton(ui->radioButton_Bad, 3);
+    compatibility->addButton(ui->radioButton_IntroMenu, 4);
+    compatibility->addButton(ui->radioButton_WontBoot, 5);
+    switch (currentId()) {
+    case 1:
```
jroweboy: what is `1` and `2` in this switch? What page they are on? Also it should have a default with some error logging in there i think.

CMakeLists.txt
BreadFish64: oops

.travis/macos/build.sh
BreadFish64: It takes it directly from the CI environment variable and defaults to off if not set

jroweboy: who set the CI env var for mac and where?

src/citra_qt/compatdb.cpp
BreadFish64: yes, that's the ID of current page

src/citra_qt/compatdb.h
```diff
@@ -25,3 +25,5 @@ private slots:
     void Submit();
     void EnableNext();
 };
+
+enum CompatDBPage { IntroPage = 0, SelectionPage = 1, FinalPage = 2 };
```
jroweboy: nit: the name of the enum already includes Page, so adding it to the end of each item is really redundant <3

src/citra_qt/compatdb.cpp
```diff
@@ -33,19 +33,21 @@ void CompatDB::Submit() {
     compatibility->addButton(ui->radioButton_IntroMenu, 4);
     compatibility->addButton(ui->radioButton_WontBoot, 5);
     switch (currentId()) {
-    case 1:
+    case CompatDBPage::SelectionPage:
```
jroweboy: still need a case for `CompatDBPage::Intro` (which can just `break;`)

BreadFish64: this is executed after the page id is changed (but before the UI updates), so technically `CompatDBPage::Intro` should never be encountered as case, but I can still add it if you want

B3n30:
@jroweboy, @chris062689 You requested changes on that PR. Are those fixed? Can it get merged? Do you want to give it a final review?

BreadFish64:
@jroweboy, @chris062689 bump^

Issue #3331, opened 2017-12-28
Popups for missing system files repeat and spam the user. 


chris062689:
This is a requested enhancement for the pop-up dialog messages when users are missing system files or shared fonts. 

In Pokemon USUM when a user goes to the festival plaza and are missing system files, the game attempts to load these files repeatedly, causing the dialog box to open multiple times, spam the user, and eventually crash the emulator. 

For each type of error, could we keep track of if it's already appeared to the user, and only show the dialogue box once? (We'd still want to log it though in the console)

![](https://i.imgur.com/HK1MZUH.png)

Attached screenshot submitted by a user, notice the black outline indicating multiple popups.

einstein95:
For one, try dumping the required files from a 3DS like it's telling you to do.

Secondly, activate Windows.

Subv:
We could pause the emulation when the error is triggered and resume it when the dialog is closed.

chris062689:
@Subv seems a PR was open for that. Is that the best approach?

chris062689:
An issue I see with pausing the main thread in #3333 is it would cause multiplayer desync if the emulation thread is paused.

hyarsan:
This happened one time.... it nearly crashed my computer!

adityaruplaha:
Erm, it doesn't look at all easy to me, so why is it labelled so? Afaik, 
![image](https://user-images.githubusercontent.com/30696515/34907920-86216b18-f8ac-11e7-891a-e03ca6f04178.png)
No disrespect meant tho.

jroweboy:
@adityaruplaha It is an easy task; was two lines to fix: https://github.com/citra-emu/citra/pull/3333/files It doesn't touch anything but the qt frontend and doesn't require any citra specific knowledge. Also I'm closing this as the fix was merged :)

adityaruplaha:
Oh, I was thinking of writing to a shared location to check if that has appeared.... etc & what not. Also this is  **not** a sarcastic comment.

adityaruplaha:
Anyways, @jroweboy , if u r free, can you please review #3366 , I really need some guidance (also some advice about how to use clang-format).

PR #3333, opened 2017-12-28
citra_qt: Pause emulation on CoreError

Senjosei:
Just a small commit to pause the emulation thread on CoreError because some games will spam the dialog boxes when the users are missing system files or shared fonts #3331

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3333)

### DIFF
--- a/src/citra_qt/bootmanager.cpp
+++ b/src/citra_qt/bootmanager.cpp
@@ -36,6 +36,7 @@ void EmuThread::run() {
 
             Core::System::ResultStatus result = Core::System::GetInstance().RunLoop();
             if (result != Core::System::ResultStatus::Success) {
+                this->SetRunning(false);
                 emit ErrorThrown(result, Core::System::GetInstance().GetStatusDetails());
             }
 
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -918,6 +918,7 @@ void GMainWindow::UpdateStatusBar() {
 }
 
 void GMainWindow::OnCoreError(Core::System::ResultStatus result, std::string details) {
+    emu_thread->SetRunning(false);
     QMessageBox::StandardButton answer;
     QString status_message;
     const QString common_message =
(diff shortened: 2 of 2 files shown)


wwylele:
We probably had some reason to avoid pausing the game when introducing the CoreError message, but I can't recall. @TheKoopaKingdom Do you remember something?

chris062689:
An issue I see with pausing the main thread is it would cause multiplayer desync if the emulation thread is paused.

TheKoopaKingdom:
>We probably had some reason to avoid pausing the game when introducing the CoreError message, but I can't recall. @TheKoopaKingdom Do you remember something?

There were 2 main reasons for making the message boxes let core keep running:
- [False positives](https://github.com/citra-emu/citra/pull/2611#issuecomment-285876905). Pausing the game for something that's non-critical is undesirable.
- Unintrusive. I forget where this was said, but someone (I believe JFR?) brought up the fact that running a game should be straightforward, and pausing it for an error would be too intrusive.

However, do keep in mind these reasons were to ditch the [original implementation](https://github.com/citra-emu/citra/pull/2611/commits/0409bdfea5ea046e3d040ab494b8a0764fd35424#diff-5e8bf22f5c3b3579c402308dd96b7563R689), which kept the game running but offered the ability to shutdown the game.

chris062689:
@TheKoopaKingdom
I don't necessarily agree on the thought that we shouldn't be unobtrusive when they're clearly kissing files that would be required for proper 3DS emulation. 

Perhaps it's time to take another look at implementing our own open-source system files for these specific types of files, and provide the user an option to install them automatically if we determine they are missing on start. 

That would solve this issue of getting spammed due to missing system files, or at least make it less of an issue?

yuriks:
The false positive argument seems compelling to me, you really can just steam ahead in many situations. Maybe an alternative would be to add a "Don't show this again for this emulation session." checkmark on the dialog?

chris062689:
That's what I was suggesting as the default. Once they get that specific error once, don't show it again.

PR #3334, opened 2017-12-29
[WIP] Service/HTTP_C: stub methods

mailwl:
Initial commit: devkitPro/examples/3ds/network/http [pass]
- Only GET
- Not sure, if CPR is good for this

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3334)


src/core/hle/service/http_c.cpp
```diff
@@ -2,70 +2,337 @@
...
-    {0x003800C0, nullptr, "SetPostDataTypeSize"},
-    {0x00390000, nullptr, "Finalize"},
-};
-
-HTTP_C::HTTP_C() {
-    Register(FunctionTable);
+static const ResultCode ERROR_CONTEXT_ERROR(0xD8A0A066);
+static const ResultCode RESULT_DOWNLOADPENDING(0xd840a02b);
+
+bool HTTP_C::ContextExists(s32 context_id, IPC::RequestParser& rp) const {
```
Subv: I don't like this pattern, it hides the error code behind this function call

src/core/hle/service/http_c.cpp
```diff
@@ -2,70 +2,337 @@
...
+    }
+    return true;
+}
+
+void HTTP_C::Initialize(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp(ctx, 0x1, 1, 4);
+    const u32 shmem_size = rp.Pop<u32>();
+    const u32 process_header = rp.Pop<u32>();
+    rp.Pop<u32>(); // unused
+    const auto sharedmem_handle = rp.PopHandle();
```
Subv: PopObject

src/core/hle/service/http_c.cpp
```diff
@@ -2,70 +2,337 @@
...
+    const s32 context_id = rp.Pop<s32>();
+
+    if (!ContextExists(context_id, rp))
+        return;
+
+    Context& context = contexts[context_id];
+
+    switch (context.method) {
+    case RequestMethod::Get: {
+        context.response = cpr::Get(context.url, context.request_headers);
```
Subv: Operations that can block the host thread should be performed async while the calling guest thread sleeps (via SleepClientThread) if the functions are supposed to block in a real 3DS

src/core/hle/service/http_c.cpp
```diff
@@ -2,70 +2,337 @@
...
-    {0x00360000, nullptr, "ClearDNSCache"},
-    {0x00370080, nullptr, "SetKeepAlive"},
-    {0x003800C0, nullptr, "SetPostDataTypeSize"},
-    {0x00390000, nullptr, "Finalize"},
-};
-
-HTTP_C::HTTP_C() {
-    Register(FunctionTable);
+static const ResultCode ERROR_CONTEXT_ERROR(0xD8A0A066);
+static const ResultCode RESULT_DOWNLOADPENDING(0xd840a02b);
```
wwylele: Use `ResultCode` constructor that accepts each components of the code. Fill up each components so that we can understand the code in a clearer way

src/core/hle/service/http_c.cpp
```diff
@@ -2,70 +2,337 @@
...
+    IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
+    rb.Push(RESULT_SUCCESS);
+    rb.Push<u32>(context_counter);
+
+    LOG_WARNING(Service_HTTP, "called, size=%u, url=%s", size, context.url.c_str());
+}
+
+void HTTP_C::CloseContext(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp(ctx, 0x3, 1, 0);
+    const s32 context_id = rp.Pop<s32>();
```
wwylele: Do you have any source stating that this parameter is signed? To me, an unsigned type makes more sense for an "ID".

src/core/hle/service/http_c.cpp
mailwl: i dislike it too, but i want to avoid duplicating code, and don't know, how to make it better. There is many context-based functions

src/core/hle/service/http_c.cpp
```diff
@@ -2,70 +2,337 @@
...
+    context.url.resize(size);
+    context.method = rp.PopEnum<RequestMethod>();
+
+    Kernel::MappedBuffer& buffer = rp.PopMappedBuffer();
+    buffer.Read(&context.url[0], 0, size);
+    contexts[++context_counter] = context;
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(2, 0);
+    rb.Push(RESULT_SUCCESS);
+    rb.Push<u32>(context_counter);
```
wwylele: `PopMappedBuffer` always has a corresponding `PushMappedBuffer` to the IPC response (and thus the response header should be `(2, 2)`. I know this is not documented on 3dbrew, which is more like a oversight there. Unless you can prove that it is not the case by hwtest or system module code, please always add `PushMappedBuffer` for mapped buffer to the IPC response.

src/core/hle/service/http_c.h
```diff
@@ -4,19 +4,235 @@
...
 namespace Service {
 namespace HTTP {
 
-class HTTP_C final : public Interface {
+class HTTP_C final : public ServiceFramework<HTTP_C> {
 public:
     HTTP_C();
+    ~HTTP_C();
+
+private:
```
lioncash: I would suggest using the PImpl idiom here to avoid requiring the dependency on the cpr include to be present within this header, preventing it from being included in any other file that includes it.

src/core/hle/service/http_c.h
```diff
@@ -4,19 +4,235 @@
...
+        Put = 0x4,
+        Delete = 0x5,
+    };
+    struct Context {
+        cpr::Url url;
+        cpr::Header request_headers;
+        RequestMethod method{RequestMethod::Get};
+        bool initialized = false;
+        bool proxy_default = false;
+        bool keep_alive;
```
lioncash: This should probably have a default initializer as well.

src/core/hle/service/http_c.h
```diff
@@ -4,19 +4,235 @@
 
 #pragma once
 
+#include <map>
```
lioncash: This should be `<unordered_map>`

neobrain:

2018-01-09T08:15:09Z: https://github.com/citra-emu/citra/compare/80025792d60a76b1c73a46ec2cf073897e3941cd...mailwl:988b2d5099b82bc6992ef62b9ad0e405dd524900

neobrain:

2018-01-09T14:54:34Z: https://github.com/citra-emu/citra/compare/80025792d60a76b1c73a46ec2cf073897e3941cd...mailwl:a84a5c3e7e405295bea3e9a064dde1e3004e3787

wwylele:
Since the title says "[WIP]", I assume that you are going to push more commits to this branch. What features are planned to be included in the final version of this branch? For example, are you going to implement SET, POST etc. as well?

mailwl:
yes. Now i am trying to get work imgr3ds (it uses POST). If i success, i think, this is enough for one PR

src/core/hle/service/http_c.cpp
```diff
@@ -2,70 +2,488 @@
...
+    contexts[context_id].ssl_options = ssl_options;
+
+    IPC::RequestBuilder rb = rp.MakeBuilder(1, 0);
+    rb.Push(RESULT_SUCCESS);
+
+    LOG_WARNING(Service_HTTP, "called, context_id=%u, ssl_options=0x%X", context_id, ssl_options);
+}
+
+void HTTP_C::Impl::SetKeepAlive(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp(ctx, 0x2B, 2, 0);
```
tgsm: `IPC::RequestParser rp(ctx, 0x37, 2, 0);`

B3n30:
I'm closing this since there is a follow up PR and we are going to remove CPR as a submodule

PR #3335, opened 2017-12-29
citra(SDL) small fixes: fix fall-through

mailwl:
+ remove u16 'always false' compare, add const to const strings

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3335)


mailwl:
thanks. it's Resharper C++ suggests to add const by checking function formal parameters

PR #3336, opened 2017-12-30
citra_qt: Resized main config window

PizzicatoWolf:
Hello there! This is my first pull request. I thought I could do some work on the front-end so I decided to resize the default configuration layout width. It looks too wide and It looks much nicer when it's not so stretched out. Let me know if this is all good. :)

Original
original

Fixed
fixed

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3336)

### DIFF
--- a/src/citra_qt/configuration/configure.ui
+++ b/src/citra_qt/configuration/configure.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>740</width>
+    <width>461</width>
     <height>500</height>
    </rect>
   </property>
@@ -17,7 +17,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>4</number>
      </property>
      <widget class="ConfigureGeneral" name="generalTab">
       <attribute name="title">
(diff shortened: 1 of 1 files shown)


src/citra_qt/configuration/configure.ui
```diff
@@ -34,15 +34,15 @@
...
+     <widget class="ConfigureGraphics" name="graphicsTab">
+      <attribute name="title">
+       <string>Graphics</string>
+      </attribute>
+     </widget>
      <widget class="ConfigureAudio" name="audioTab">
-       <attribute name="title">
-         <string>Audio</string>
-       </attribute>
+      <attribute name="title">
```
BreadFish64: Why is this changed?

src/citra_qt/configuration/configure.ui
```diff
@@ -34,15 +34,15 @@
...
       </attribute>
      </widget>
-      <widget class="ConfigureGraphics" name="graphicsTab">
-        <attribute name="title">
-          <string>Graphics</string>
-        </attribute>
-      </widget>
+     <widget class="ConfigureGraphics" name="graphicsTab">
+      <attribute name="title">
+       <string>Graphics</string>
```
BreadFish64: Why is this changed?

src/citra_qt/configuration/configure.ui
```diff
@@ -17,7 +17,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
```
BreadFish64: Why is this changed?

src/citra_qt/configuration/configure.ui
PizzicatoWolf: This is unintentionally changed by Qt Designer. I'll submit a new commit with this fixed.

PizzicatoWolf: Just took a look at the original file and found that the rest was using one space indents except for this section which Qt Designer fixed.

src/citra_qt/configuration/configure.ui
wwylele: indentation fix, I guess

PR #3361, opened 2018-01-10
Service/PTM: convert to ServiceFramework

wwylele:
Functional change:
 - the maximum session for each port is RE-ed from PTM code. All are 26, except for the one for `ptm:sets`, which is 1.
 - Push mapped buffer in response in `GetStepHistory`.
 - Fixed response header in `CheckNew3DS`.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3361)


adityaruplaha:
Can you please explain what are the major changes made in this, from a user point of view?

wwylele:
@adityaruplaha there is basically no changes from a user point of view. It is part of ongoing project (#2531) to convert deprecated service interface to the new framework, which would help more accurate service implementation in the future.

PR #3389, opened 2018-01-15
Common: Replace MurmurHash3 with CityHash64

yuriks:
CityHash64 is faster than Murmur3 at all sizes, but especially for short keys. I plan on using this to improve hashing of small keys in the vertex shader cache, where we currently just assume no collisions.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3389)


src/video_core/renderer_opengl/gl_shader_gen.h
```diff
@@ -131,10 +131,6 @@ union PicaShaderConfig {
 
     } state;
 };
-#if (__GNUC__ >= 5) || defined(__clang__) || defined(_MSC_VER)
-static_assert(std::is_trivially_copyable<PicaShaderConfig::State>::value,
```
linkmauve: Are you sure you want to remove this assert, and not just the guards around it?

yuriks: Yeah, I removed it because the check was moved to the struct function. Afaik the only reason we had that requirement was because of the hashing, so imo it's better to let that function verify it. That way the asserts are attached to the particular uses that impose them, rather then the type.

If there is something else imposing this requirement then I'm ok with restoring this.

src/common/hash.h
```diff
@@ -5,22 +5,30 @@
...
+}
+
+/**
+ * Computes a 64-bit hash of a struct. In addition to being POD (trivially copyable and having
+ * standard layout), it is also critical that either the struct includes no padding, or that any
+ * padding is initialized to a known value by memsetting the struct to 0 before filling it in.
+ */
+template <typename T>
+static inline u64 ComputeStructHash64(const T& data) {
+    static_assert(std::is_pod<T>::value, "Type passed to ComputeStructHash64 must be POD");
```
lioncash: Note that the notion of PODs will be [deprecated](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0767r1.html) in C++20 in favor of standard layout or trivially copyable types.

lioncash: To add to this, unfortunately, since we don't use C++17 yet, we can't just use [std::has_unique_object_representations](http://en.cppreference.com/w/cpp/types/has_unique_object_representations) and call it a day. Since this is literally the case it was introduced for.

yuriks: Eh, I had both separate checks there before, then read on cppreference that is_pod was defined as (is_trivially_copyable && is_standard_layout). They don't seem to mention that it's deprecated, will change.

`has_unique_object_representation` seems useful since it also detects padding bits. Do you think it's a good idea to conditionally add the check for C++17 compilers if possible? I'm kinda scared of buggy implementations though.

PR #3394, opened 2018-01-17
Added string to display Compile date in the titlebar

zoltx23:
Shows the Build date for when the Application is compiled. Pretty straight forward. 

If you don't want the text `Build date:` it can just be easily removed.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3394)


zoltx23:
Oops. I did that wrong. 

I need to check it before I PR.

PR #3402, opened 2018-01-20
Allow Installed Apps to Display in the Game List

BreadFish64:
- Allows installed games to display in the game list
- Allows installed (executable) system applications to display in the game list
- Replaces the game icon with the icon from an installed update if available
image

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3402)

### DIFF
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -403,11 +403,11 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
                 std::vector<u8> original_smdh;
                 loader->ReadIcon(original_smdh);
 
-                if (program_id >= 0x4000000000000 && program_id <= 0x40000FFFFFFFF)
+                if (program_id < 0x00040000'00000000 || program_id > 0x00040000'FFFFFFFF)
                     return original_smdh;
 
                 std::string update_path = Service::AM::GetTitleContentPath(
-                    Service::FS::MediaType::SDMC, program_id + 0xe00000000);
+                    Service::FS::MediaType::SDMC, program_id + 0x0000000E'00000000);
 
                 if (!FileUtil::Exists(update_path))
                     return original_smdh;
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -449,6 +449,10 @@ void GameListWorker::run() {
         std::string(FileUtil::GetUserPath(D_SDMC_IDX).c_str()) +
         "Nintendo "
         "3DS/00000000000000000000000000000000/00000000000000000000000000000000/title/00040000"));
+    watch_list.append(QString::fromStdString(
+        std::string(FileUtil::GetUserPath(D_SDMC_IDX).c_str()) +
+        "Nintendo "
+        "3DS/00000000000000000000000000000000/00000000000000000000000000000000/title/0004000e"));
     watch_list.append(
         QString::fromStdString(std::string(FileUtil::GetUserPath(D_NAND_IDX).c_str()) +
                                "00000000000000000000000000000000/title/00040010"));
@@ -457,10 +461,10 @@ void GameListWorker::run() {
         std::string(FileUtil::GetUserPath(D_SDMC_IDX).c_str()) +
             "Nintendo "
             "3DS/00000000000000000000000000000000/00000000000000000000000000000000/title/00040000",
-        true ? 256 : 0);
+        2);
     AddFstEntriesToGameList(std::string(FileUtil::GetUserPath(D_NAND_IDX).c_str()) +
                                 "00000000000000000000000000000000/title/00040010",
-                            true ? 256 : 0);
+                            2);
     emit Finished(watch_list);
 }
 
--- a/src/citra_qt/game_list.cpp
+++ b/src/citra_qt/game_list.cpp
@@ -403,7 +403,7 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
                 std::vector<u8> original_smdh;
                 loader->ReadIcon(original_smdh);
 
-                if (0x4000000000000 > program_id && program_id > 0x40000FFFFFFFF)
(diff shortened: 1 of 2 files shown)


valentinvanelslande:
the home menu does not appear

BreadFish64:
@valentinvanelslande the home menu is technically an applet so it's not in the same directory as the other runnable system apps, I could add the six possible paths to detect it but I would want someone else's opinion on that

chris062689:
I don't see the need to launch applets in the games list that's nonsensical.   @BreadFish64 I'd leave the home menu launching to another PR.

src/citra_qt/game_list.cpp
```diff
@@ -401,8 +402,27 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
             u64 program_id = 0;
             loader->ReadProgramId(program_id);
 
+            std::vector<u8> update_smdh;
+            std::string update_path;
+            u64 update_id = 0;
+            std::unique_ptr<Loader::AppLoader> update_loader = nullptr;
```
MerryMage: Reduce the scope of these variables.

At least do:

```
            std::vector<u8> update_smdh;

            if (0x4000000000000 <= program_id && program_id <= 0x40000FFFFFFFF) {
                u64 update_id = program_id + 0xe00000000;
                std::string update_path =
                    Service::AM::GetTitleContentPath(Service::FS::MediaType::SDMC, update_id);
                if (FileUtil::Exists(update_path)) {
                    std::unique_ptr<Loader::AppLoader> update_loader = 
                        Loader::GetLoader(update_path);
                    if (update_loader) {
                        update_loader->ReadIcon(update_smdh);
                        update_loader->ReadProgramId(update_id);
                    }
                }
            }
```

Personally I would prefer the IIFE-style:

```
            std::vector<u8> update_smdh = [program_id]() -> std::vector<u8> {
                if (0x4000000000000 > program_id && program_id > 0x40000FFFFFFFF)
                    return {};

                u64 update_id = program_id + 0xe00000000;
                std::string update_path =
                    Service::AM::GetTitleContentPath(Service::FS::MediaType::SDMC, update_id);

                if (!FileUtil::Exists(update_path))
                    return {};

                std::unique_ptr<Loader::AppLoader> update_loader = Loader::GetLoader(update_path);

                if (!update_loader)
                    return {};

                std::vector<u8> update_smdh;
                update_loader->ReadIcon(update_smdh);
                update_loader->ReadProgramId(update_id);
                return update_smdh;
            }();
```

src/citra_qt/game_list.cpp
```diff
@@ -421,7 +445,22 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
...
+                               "00000000000000000000000000000000/title/00040010"));
     AddFstEntriesToGameList(dir_path.toStdString(), deep_scan ? 256 : 0);
+    AddFstEntriesToGameList(
+        std::string(FileUtil::GetUserPath(D_SDMC_IDX).c_str()) +
+            "Nintendo "
+            "3DS/00000000000000000000000000000000/00000000000000000000000000000000/title/00040000",
+        true ? 256 : 0);
+    AddFstEntriesToGameList(std::string(FileUtil::GetUserPath(D_NAND_IDX).c_str()) +
+                                "00000000000000000000000000000000/title/00040010",
+                            true ? 256 : 0);
```
wwylele: what's up with the `true ? 256 : 0`? isn't it just `256`? Also, the directory hierarchy under sdmc or nand is well-defined and the recursion doesn't need to be greater than 5 or something.

src/citra_qt/game_list.cpp
```diff
@@ -401,8 +402,31 @@ void GameListWorker::AddFstEntriesToGameList(const std::string& dir_path, unsign
...
+                std::vector<u8> update_smdh;
+                update_loader->ReadIcon(update_smdh);
+                update_loader->ReadProgramId(update_id);
+                return update_smdh;
+            }();
+
             emit EntryReady({
-                new GameListItemPath(QString::fromStdString(physical_name), smdh, program_id),
+                new GameListItemPath(QString::fromStdString(physical_name), smdh, program_id,
+                                     update_smdh),
```
wwylele: Is it necessary to let `GameListItemPath` object handle the two smdh? why not just replacing the `smdh` with `update_smdh` here (when update is found)?

BreadFish64: The SDMH is also where the program name is taken from... I might want to check what happens if I take that from update_sdmh as well

src/citra_qt/game_list.cpp
BreadFish64: Tbh, I just put that there because the original had it, I could try actually figuring out what it does

FearlessTobi: IIRC it originally was used for recursive GameList search. When you enable it in the Citra menu, it searches Games in up to 256 subdirectories. I think you just copied it and replaced the boolean which enables recursion with true.

BreadFish64: Wwylele explain it in discord, I get what the problem is now :)

LogPod:
Would be nice if we could have an option like say open the location where the game/app is installed.

src/citra_qt/game_list_p.h
```diff
@@ -77,19 +77,19 @@ class GameListItemPath : public GameListItem {
...
         setData(qulonglong(program_id), ProgramIdRole);
 
+        Loader::SMDH smdh;
         if (!Loader::IsValidSMDH(smdh_data)) {
             // SMDH is not valid, set a default icon
             setData(GetDefaultIcon(true), Qt::DecorationRole);
             return;
         }
 
-        Loader::SMDH smdh;
```
wwylele: unnecessary change

MerryMage:
Gonna merge this tmr if there are no complains.

jroweboy:
:shipit:  Would like a follow up PR for what Logpod suggested as well if you get the chance.

BreadFish64:
@jroweboy already started on it ;)

PR #3403, opened 2018-01-21
[RFC] VFS Interfaces and utility functions

yuriks:
The intent of this is to introduce a layerable file abstraction. This can then be used to implement file formats in a way that's transparent to whatever's using the file, e.g. HostFile -> NCSD container -> NCCH container -> decryption layer -> RomFS.

It's an explicit non-goal for this to handle any emulation behavior from the 3DS' FS module, so archives and stuff will still be their own thing, that can use these libraries to implement whatever they need if necessary.

I'm marking this as RFC because I don't actually have anything that uses this yet. I'll probably work on porting the loaders to it to start with.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3403)

### DIFF
--- a/src/core/hle/result.h
+++ b/src/core/hle/result.h
@@ -152,6 +152,7 @@ enum class ErrorModule : u32 {
     QTM = 96,
     NFP = 97,
 
+    Citra = 253,
     Application = 254,
     InvalidResult = 255
 };
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,13 +1,15 @@
 # Enable modules to include each other's files
 include_directories(.)
 
+add_subdirectory(audio_core)
 add_subdirectory(common)
 add_subdirectory(core)
-add_subdirectory(video_core)
-add_subdirectory(audio_core)
-add_subdirectory(network)
 add_subdirectory(input_common)
+add_subdirectory(network)
 add_subdirectory(tests)
+add_subdirectory(vfs)
+add_subdirectory(video_core)
+
 if (ENABLE_SDL2)
     add_subdirectory(citra)
 endif()
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -62,6 +62,7 @@ add_library(common STATIC
     platform.h
     printable.h
     quaternion.h
+    result.h
     scm_rev.cpp
     scm_rev.h
     scope_exit.h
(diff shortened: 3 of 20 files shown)


src/core/hle/result.h
```diff
@@ -152,6 +152,7 @@ enum class ErrorModule : u32 {
     QTM = 96,
     NFP = 97,
 
+    Citra = 253,
```
wwylele: I don't like putting this into the hle error code. There are several disadvantages:
 - this forms an unnecessary dependency of `vfs` on `core` (and `core obviously also depends on `vfs`, yay for circular dependency).
 - you bind yourself to the complicated and inflexible 3DS error code format.
 - it breaks the purity of `ErrorModule` which was a documentation only for real 3DS codes.

I do see some potential advantages. However, they are arguable to me:
 - VFS can utilize the `ResultVal` class by using `ResultCode`. However, IMO this can be achieved cleaner by either
   - using existing library template such as `std::/boost::variant`, or
   - abstracting `ResultVal`into a more generalized template for any types of error code.
 - core/file_sys can potentially forward the error code from VFS to HLE services. However, as the existing implementation shows, different archive types can have different error codes for the same error. Also, you are still defining your own error codes in vfs/vfs.h. In both cases, the error codes are not forwardable and a big switch table or code map is unavoidable.

My suggestion is: let VFS define its own error type (can be a simple wrapper around a single enum).

src/common/common_types.h
```diff
@@ -56,6 +56,9 @@ class NonCopyable {
     constexpr NonCopyable() = default;
     ~NonCopyable() = default;
 
+    constexpr NonCopyable(NonCopyable&&) = default;
+    constexpr NonCopyable& operator=(NonCopyable&&) = default;
+
```
wwylele: I personally treat the `NonCopyable` idiom as pre-c++11 stuff, and it is really confusing when involving the move construction. Maybe one class wants to disallow both copy and move by inheriting `NonCopyable`? I think it is clean enough to just `delete` the constructor and assignment in the target class.

yuriks: Done

src/core/hle/result.h
yuriks: I created a new Result type based on boost::variant, which can hold any result or error types. Take a look and see if you think it's reasonable.

yuriks:
Closing as I don't plan on continuing work on this anytime soon... Someone else feel free to pick it up if wanted.

PR #3409, opened 2018-01-26
 Service/FRD: Stub HasLoggedIn, IsOnline, Login and Logout

valentinvanelslande:
I do not know if the errors are the right ones...

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3409)


src/core/hle/service/frd/frd_u.cpp
```diff
@@ -9,10 +9,10 @@ namespace Service {
 namespace FRD {
 
 const Interface::FunctionInfo FunctionTable[] = {
-    {0x00010000, nullptr, "HasLoggedIn"},
-    {0x00020000, nullptr, "IsOnline"},
-    {0x00030000, nullptr, "Login"},
-    {0x00040000, nullptr, "Logout"},
+    {0x00010000, HasLoggedIn, "HasLoggedIn"},
+    {0x00020000, IsOnline, "IsOnline"},
+    {0x00030000, Login, "Login"},
```
mailwl: change please code to 0x00030002, here is error

src/core/hle/service/frd/frd.cpp
```diff
@@ -89,6 +90,51 @@ void GetFriendAttributeFlags(Service::Interface* self) {
...
+void IsOnline(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
+
+    cmd_buff[1] = RESULT_SUCCESS.raw; // No error
+    cmd_buff[2] = 0;                  // Not online
+    LOG_WARNING(Service_FRD, "(STUBBED) called");
+}
+
+void Login(Service::Interface* self) {
+    u32* cmd_buff = Kernel::GetCommandBuffer();
```
mailwl: according https://www.3dbrew.org/wiki/FRDU:Login, second parameter is completion event handler

valentinvanelslande: How i can get that from cmd_buff?

mailwl: for example, https://github.com/mailwl/citra/blob/friends/src/core/hle/service/frd/frd.cpp#L46
but, this method is obsoleted, i am not sure it has to be used for new stubbing...

valentinvanelslande:
@wwylele what error i can use?

mailwl:
i saw the game which uses this functions, but don't remember the name (AFAIK, it don't crash anyway, just red lines in log)

valentinvanelslande:
test for checking if returns error:
https://gist.github.com/valentinvanelslande/c2b165cef72eded8733b9006d4fbee72
build: [.3dsx](https://github.com/citra-emu/citra/files/1670497/frd_test.zip)

Issue #2642, opened 2017-03-28
Make a control panel widget


wwylele:
This is an idea I came up with when thinking about #2634, #2637 and other stuff. I think citra-qt frontend should have a control panel containing miscellaneous inputs/outputs, which are not very important but useful in some cases, such as the volume slider, the 3D slider, the network switch, battery/AC state emulation, headphone state emulation, and notification LEDs. These stuff deserve their own widget because they don't belong to emulation configuration, but are parts of real 3DS interface which should be emulated.

RedTrainer97:
In my opinion, things like volume slider, 3D slider and network switch should stay in Configuration; LEDs, emulated battery state and other "feedbacks" instead may be putted in the Status bar.

JayFoxRox:
@RedTrainer97 the system configuration panel is an emulation of the 3DS OS / software settings dialogs.

The proposed widget / panel / status bar changes by wwylele reflect actual hardware present on the 3DS which is visible / can be interacted with at all times.

I think we should move these hardware features into something like a list of trayicons, similar to how it's done in VirtualBox or other virtualizers. (And the ability to map these features to hotkeys in the settings for fullscreen users)

RedTrainer97:
I think this layout will be useful for sliders and other inputs.
![immagine](https://cloud.githubusercontent.com/assets/25330179/24414047/899de80c-13dd-11e7-817a-c12834b2789b.jpg)
And the statusbar will show LEDs.

B3n30:
Sorry, but I don't like it. These are all functions you normally won't use EVER. Placing them so prominently makes no sense at all.

jroweboy:
@B3n30 i thought that is for a configuration dialog. its still awkward for a config dialog, but it would at least make more sense to me.

LittleWhite-tb:
If you want to place these like this, I think the sliders should retract/disappear automatically. To make these appear again, just move the mouse cursor on the left/right of the window.

RedTrainer97:
Of course they are not permanent, this is just an example of how to position them in order to make them similar of 3DS positions.

RedTrainer97:
I've added the controls in the main window (you see them only in play mode and if you check the option).

![immagine](https://cloud.githubusercontent.com/assets/25330179/24509130/ddee8194-1565-11e7-8f91-46adb00b53cc.jpg)

I've renamed my PR (https://github.com/citra-emu/citra/pull/2637) for this new feature.

yuriks:
[...]

PR #3415, opened 2018-01-31
 citra_qt: Add Control Panel (#2642)

valentinvanelslande:
Added Power tab
screenshot

- Level:
Battery level.

- Adapter Connected:
Adapter connected.

- Battery Charging:
Battery charging.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3415)

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -17,6 +17,7 @@
 #include <QtWidgets>
 #include "citra_qt/aboutdialog.h"
 #include "citra_qt/bootmanager.h"
+#include "citra_qt/control_panel.h"
 #include "citra_qt/compatdb.h"
 #include "citra_qt/configuration/config.h"
 #include "citra_qt/configuration/configure_dialog.h"
@@ -387,6 +388,7 @@ void GMainWindow::ConnectMenuEvents() {
     connect(ui.action_Report_Compatibility, &QAction::triggered, this,
             &GMainWindow::OnMenuReportCompatibility);
     connect(ui.action_Configure, &QAction::triggered, this, &GMainWindow::OnConfigure);
+    connect(ui.action_Control_Panel, &QAction::triggered, this, &GMainWindow::OnControlPanel);
 
     // View
     connect(ui.action_Single_Window_Mode, &QAction::triggered, this,
@@ -935,6 +937,11 @@ void GMainWindow::OnToggleFilterBar() {
 void GMainWindow::OnSwapScreens() {
     Settings::values.swap_screen = !Settings::values.swap_screen;
     Settings::Apply();
+}
+
+void GMainWindow::OnControlPanel() {
+    ControlPanel control_panel;
+    control_panel.exec();
 }
 
 void GMainWindow::OnCreateGraphicsSurfaceViewer() {
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -1030,11 +1030,6 @@ void GMainWindow::OnToggleFilterBar() {
     }
 }
 
-void GMainWindow::OnSwapScreens() {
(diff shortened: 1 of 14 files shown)


chris062689:
All UI changes need a screenshot attached and a detailed description of what each new setting does.

wwylele:
I am assuming that this is related to #2642 ?

valentinvanelslande:
Screenshot added

src/core/settings.h
```diff
@@ -141,6 +147,9 @@ struct Values {
     std::string verify_endpoint_url;
     std::string citra_username;
     std::string citra_token;
+
+    // Control Panel
+    PTMInfo ptm_values;
```
wwylele: The "setting" is conventionally used used for savable user configurations (i.e. those stuff saved in ini file), so the control panel stuff doesn't belong here IMO. You can just let PTM module expose some interface to frontend directly

src/citra_qt/main.cpp
```diff
@@ -937,6 +939,11 @@ void GMainWindow::OnSwapScreens() {
     Settings::Apply();
 }
 
+void GMainWindow::OnControlPanel() {
+    ControlPanel control_panel;
+    control_panel.exec();
```
wwylele: You open a [modal dialog](http://doc.qt.io/archives/qt-4.8/qdialog.html#exec) here which would block any input to the main window, and user can't play the game while the panel is open. I think the design goal of the panel is that user can let the panel be there and continue the game, and occasionally view some information and change some settings in it, considering that we might add more stuff including 3d slider, volume slider and motion control viewer etc.

src/core/settings.h
valentinvanelslande: I added it there because in other places the linker gives error

src/citra_qt/main.cpp
MerryMage: Instead of `QDialog`, subclassing `QDockWidget` would be preferable. This would allow the window to stay on top of the parent window and be minimised when the parent window is minimised &c. Of course you'll have to call `setAllowedAreas(Qt::NoDockWidgetArea)` so it's not actually dockable, if docking is undesireable.

valentinvanelslande: with show() the window closes

adityaruplaha:
Clang fixes required.

valentinvanelslande:
@adityaruplaha where is it?

adityaruplaha:
Go to edit, you will see it. GitHub truncates trailing whitespaces while viewing.

valentinvanelslande:
@adityaruplaha there is no whitespace!

valentinvanelslande:
Does anyone know how to fix this?
https://user-images.githubusercontent.com/32885846/35893067-46d5c798-0b7a-11e8-94eb-b22461b0b903.png

wwylele:
@valentinvanelslande could you join discord and ask all your questions there? Because
1. I feel that you don't like to provide sufficient information at once (for example, what linker error and what's your code); 
2. Your questions are more about the programming language itself rather the feature you implement.

valentinvanelslande:
I'm already in discord (Valentin Vanelslande#3318), there's no linker error

adityaruplaha:
Please remove trailing whitespace.

valentinvanelslande:
THERE IS NO WHITESPACE!!!

adityaruplaha:
I dont think the Travis diff is wrong...
https://travis-ci.org/citra-emu/citra/jobs/338643236

Also, see this:
image
https://github.com/valentinvanelslande/citra/edit/master/src/citra_qt/control_panel.h

wwylele:
I don't see this going anywhere as the modal dialog issue is still not fixed. Closing.

valentinvanelslande:
@wwylele fixed

PR #3421, opened 2018-02-03
Service/APT: convert to ServiceFramework

wwylele:
Follow up of #3408.

Small functional change: Receive/GlanceParameter always push a buffer with the maximum size specified in the request, regardless of the actual data size. This is RE-d from the NS sysmodule/APT service code, ~~and also prevent ASSERT on zero size buffer in hle_ipc.cpp~~removed the ASSERT anyway because it still need to push an empty buffer in some cases.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3421)


src/core/hle/service/apt/apt.h
```diff
@@ -46,416 +51,472 @@ enum class ScreencapPostPermission : u32 {
...
+         *     7 : (Size<<14) | 2
+         *     8 : Input parameter buffer ptr
+         * Outputs:
+         *     0 : Return Header
+         *     1 : Result of function, 0 on success, otherwise error code
+        */
+        void SendParameter(Kernel::HLERequestContext& ctx);
+
+        /**
+         * APT::ReceiveParameter service function. This returns the current parameter data from NS
```
Subv: What's with the formatting of this comment block?

wwylele: forgot, they were reformatted by clang-format because of additional indentation, but I actually have to re-insert line breaks by hand.

wwylele:
@Subv  Regarding to the issue above ^

[The parameter from erreula is posted with an empty buffer](https://github.com/citra-emu/citra/blob/3c03da12c83281d185957eb3c80494ad487fec97/src/core/hle/applets/erreula.cpp#L38-L45). The game first calls `GlanceParameter` with buffer_size = 0x1000 on this parameter, then calls `ReceiveParameter` with **buffer_size = 0** (the size is either hard-coded or from the result of GlanceParameter). Either way, if the implementation of erreula posting an empty buffer is correct, real 3DS should indeed push an empty static buffer in this case. I am going to remove the assert on empty buffer. Do you have any objection on this?

Subv:
I don't have any objections, the kernel supports empty static buffers, that assert was an attempt to catch code errors when copy/pasting but i suppose it is not needed anymore.

PR #3429, opened 2018-02-07
Service/FS_User: Stub Secure Value Functions

Cuyler36:
Stubbing these functions results in saves loading correctly in games which require a secure value.

An example is shown here with Animal Crossing: New Leaf:
Before:
citra-qt_2018-02-06_18-54-07

After:
citra-qt_2018-02-06_19-30-27

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3429)

### DIFF
--- a/src/core/hle/service/fs/fs_user.cpp
+++ b/src/core/hle/service/fs/fs_user.cpp
@@ -990,6 +990,60 @@ static void GetNumSeeds(Service::Interface* self) {
     rb.Push<u32>(0);
 }
 
+/**
+ * FS_User::SetSaveDataSecureValue service function.
+ *  Inputs:
+ *      0 : 0x08650140
+ *      1-2 : Secure Value
+ *      3 : Secure Value Slot
+ *      4 : Title Id
+ *      5 : Title Variation
+ *  Outputs:
+ *      0 : 0x08650140
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void SetSaveDataSecureValue(Service::Interface* self) {
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x865, 5, 0);
+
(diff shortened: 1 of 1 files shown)


LogPod:
So,this fixes the save bug in animal crossing right?

BreadFish64:
@LogPod yes, in the future please ask your questions elsewhere like discord or the forums to avoid cluttering development

adityaruplaha:
@LogPod this is for development only, so please dont ask stuff like this here. Preferably, ask in a community topic with a link to the relevant pr

src/core/hle/service/fs/fs_user.cpp
```diff
@@ -990,6 +990,60 @@ static void GetNumSeeds(Service::Interface* self) {
...
+ *  Outputs:
+ *      0 : 0x08650140
+ *      1 : Result of function, 0 on success, otherwise error code
+ */
+static void SetSaveDataSecureValue(Service::Interface* self) {
+    IPC::RequestParser rp(Kernel::GetCommandBuffer(), 0x865, 5, 0);
+
+    // TODO: Generate and Save the Secure Value
+
+    LOG_WARNING(Service_FS, "(STUBBED) called");
```
superheri: Shouldn't it be in the Debug log instead of warning? Because of warning would trigger messages like it is not something normal, but it should be normal to access the save file am I right?

Cuyler36: Some of the other FS functions that are stubbed use warning (such as FormatSaveData), while others use debug. The save file isn't directly accessed by the function. I can change it to debug if that's preferred.

wwylele: Our convention is to log on warning level for stubbed functions.

PR #3430, opened 2018-02-07
Bump to C++17

MerryMage:
This also updates the docker machines we use on our build servers to Ubuntu 18.04.

Note: Upstream CMake file can be found at: https://github.com/noloader/cryptopp-cmake/blob/7ef1dbe0db095ea480500dd8e10c9ece8975974f/CMakeLists.txt

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3430)

### DIFF
--- a/.travis/linux-frozen/install_package.py
+++ b/.travis/linux-frozen/install_package.py
@@ -35,4 +35,4 @@ def get_package(pkg, distro):
 for i in xrange(1, len(sys.argv), 3):
     get_package([sys.argv[i], sys.argv[i + 1]], sys.argv[i + 2])
 
-subprocess.check_call(['apt-get', 'install', '-y'] + deb_file_list)
+subprocess.check_call(['apt-get', 'install', '-y', '--force-yes'] + deb_file_list)
--- a/.travis/linux-frozen/build.sh
+++ b/.travis/linux-frozen/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash -ex
 
-docker pull ubuntu:16.04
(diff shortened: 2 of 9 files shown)


.travis/linux-frozen/install_package.py
```diff
@@ -35,4 +35,4 @@ def get_package(pkg, distro):
 for i in xrange(1, len(sys.argv), 3):
     get_package([sys.argv[i], sys.argv[i + 1]], sys.argv[i + 2])
 
-subprocess.check_call(['apt-get', 'install', '-y'] + deb_file_list)
+subprocess.check_call(['apt-get', 'install', '-y', '--force-yes'] + deb_file_list)
```
baka0815: Why is the ``--force-yes`` needed?

danzel:
Do we need to go to ubuntu 18.04? It's not a stable version yet, so might be a bit wonky.
Probably ok, only 2 months to release so there are unlikely to be any big changes that might affect us.

wwylele:
@danzel on 16.04 you can also install gcc 7 which supports c++17

MerryMage:
@danzel Main reason for picking 18.04 instead of 16.04 is because it's an LTS release. Feature freeze happens 1st March, beta freeze happens 8th March, and the final release is on 26th April.

danzel:
Cool, sounds fine to me then :)

adityaruplaha:
I guess this is being reviewed? I would personally like to see this merged, as I'm more familiar with the C++17 standard.

PR #3410, opened 2018-01-26
Clang format as a build target

jroweboy:
See https://github.com/yuzu-emu/yuzu/pull/117 for more details.

Please review it commit by commit. This also depends on https://github.com/citra-emu/ext-windows-bin/pull/15 for letting windows download a clang-format binary that they can use directly.

## What is all here

* Generates an optional build target when cmake can find the right version of clang-format. For MSVC, you can right click the clang-format project and have it format all source files. For Makefiles, you can run `make clang-format`.

* Each platform has a custom command that is run to gather all the files when the clang-format is run and not when the project is generated. This means that untracked files will also be formatted. 

* Bumps up the version to clang-format 6.0 which is the version installed when installing the MSVC plugin.

* For a bonus change, I added conditional options to automatically turn on `USE_BUNDLED_*` when MSVC is the toolset and the `ENABLE_*` option is ON. It always seemed silly to me that we make you click those check boxes ... After this is merged, I'll remove those steps from the wiki.

## If you would like to test this:

I've updated the instructions on the contributing guide with how to run clang format. Please try it out and let me know if anything is unclear.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3410)

PR #3434, opened 2018-02-14
QT: allow installation of multiple CIAs 

BreadFish64:
Lets the user select multiple CIAs to be installed at once.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3434)

### DIFF
--- a/src/citra_qt/main.cpp
+++ b/src/citra_qt/main.cpp
@@ -96,8 +96,9 @@ void GMainWindow::ShowCallouts() {
 }
 
 GMainWindow::GMainWindow() : config(new Config()), emu_thread(nullptr) {
-    // register size_t to use in slots and signals
+    // register types to use in slots and signals
     qRegisterMetaType<size_t>("size_t");
+    qRegisterMetaType<Service::AM::InstallStatus>("Service::AM::InstallStatus");
 
     LoadTranslation();
 
@@ -382,6 +383,8 @@ void GMainWindow::ConnectWidgetEvents() {
     connect(&status_bar_update_timer, &QTimer::timeout, this, &GMainWindow::UpdateStatusBar);
 
     connect(this, &GMainWindow::UpdateProgress, this, &GMainWindow::OnUpdateProgress);
+    connect(this, &GMainWindow::CIAInstallReport, this, &GMainWindow::OnCIAInstallReport);
+    connect(this, &GMainWindow::CIAInstallFinished, this, &GMainWindow::OnCIAInstallFinished);
 }
 
 void GMainWindow::ConnectMenuEvents() {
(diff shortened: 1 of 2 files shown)


TTFH:
Too late, i install more than 100 today.
for %f in (*.cia) do ..\citra -i %f

src/citra_qt/main.cpp
```diff
@@ -791,57 +794,67 @@ void GMainWindow::OnMenuSelectGameListRoot() {
...
-        const auto cia_progress = [&](size_t written, size_t total) {
-            emit UpdateProgress(written, total);
-        };
-        return Service::AM::InstallCIA(filepath.toStdString(), cia_progress);
+
+    QtConcurrent::run([&, filepaths] {
+        QString current_path;
+        Service::AM::InstallStatus status;
+        for (int current_index = 0; current_index < filepaths.size(); current_index++) {
+            const auto cia_progress = [&](size_t written, size_t total) {
```
baka0815: Does this have to be created every time in the loop? Shouldn't it be fine if created before the loop once?

src/citra_qt/main.cpp
```diff
@@ -897,7 +910,8 @@ void GMainWindow::OnMenuReportCompatibility() {
         QMessageBox::critical(
             this, tr("Missing Citra Account"),
             tr("In order to submit a game compatibility test case, you must link your Citra "
-               "account.<br><br/>To link your Citra account, go to Emulation &gt; Configuration &gt; "
+               "account.<br><br/>To link your Citra account, go to Emulation &gt; Configuration "
+               "&gt; "
```
baka0815: Unintentional change?

src/citra_qt/main.cpp
```diff
@@ -1045,11 +1059,10 @@ void GMainWindow::UpdateStatusBar() {
...
     if (Settings::values.use_frame_limit) {
         emu_speed_label->setText(tr("Speed: %1% / %2%")
-                                    .arg(results.emulation_speed * 100.0, 0, 'f', 0)
-                                    .arg(Settings::values.frame_limit));
+                                     .arg(results.emulation_speed * 100.0, 0, 'f', 0)
+                                     .arg(Settings::values.frame_limit));
     } else {
-        emu_speed_label->setText(tr("Speed: %1%")
-                                    .arg(results.emulation_speed * 100.0, 0, 'f', 0));
+        emu_speed_label->setText(tr("Speed: %1%").arg(results.emulation_speed * 100.0, 0, 'f', 0));
```
baka0815: Those changes are only format changes - are they necessary in this pr?

src/citra_qt/main.cpp
BreadFish64: good catch

src/citra_qt/main.cpp
BreadFish64: ah, the vs plugin probably formatted it automatically because it was "wrong" before, I don't suppose it hurts anything until #3410 is merged

PR #3436, opened 2018-02-15
Unbreak build on FreeBSD in C++17 mode

jbeich:
cryptopp master branch is currently 51 commits ahead of 6.0.0. To avoid regressions let's cherry-pick just the fix.

---
This change is [Reviewable](https://reviewable.io/reviews/citra-emu/citra/3436)


jbeich:
Hmm, PRs (unlike archives) don't support virtual backports. Closing as I don't want to touch `.gitmodules`.

wwylele:
we usually just update submodule commit like this. What's the reason you think.it is not good?

